PR #6, opened 2018-03-15
Specify nginx ingress version and allow upgrade by default

Zempashi:


roles/kube_nginx_ingress/tasks/main.yml
```diff
@@ -1,16 +1,29 @@
...
+  copy:
+    src: nginx_ingress_values.yml
+    dest: '{{ nginx_ingress_value_file.path }}'
+
 - name: 'install nginx ingress'
   command: >-
-    helm install stable/nginx-ingress
-    --name nginx-ingress
+    helm upgrade --install nginx-ingress stable/nginx-ingress
+    --version 0.11.1
```
NicolasT: Any chance we can store versions of Helm Charts we deploy in some Ansible `vars` file as well, such that they can be overridden for testing purposes and such?

roles/kube_nginx_ingress/tasks/main.yml
```diff
@@ -1,16 +1,29 @@
...
   run_once: True
   delegate_to: "{{ groups['kube-master'][0] }}"
-  register: helm_nginx_ingress
-  failed_when: >-
-    helm_nginx_ingress.rc != 0 and
-    'already exists' not in helm_nginx_ingress.stderr|default('')
-  changed_when: >-
-    'already exists' not in helm_nginx_ingress.stderr|default('')
+
+- name: 'remove uneeded nginx-ingress values file'
```
NicolasT: `unneeded`

roles/kube_nginx_ingress/tasks/main.yml
NicolasT: This may require to run `helm repo update` at some point on the host, at least when upgrading.

Also, we may want to use `--repo`  to point at a specific repository (URL) rather than depending on repositories to be pre-configured on the host system.

roles/kube_nginx_ingress/files/nginx_ingress_values.yml
```diff
@@ -0,0 +1,6 @@
+controller:
+    hostNetwork: True
+    kind: 'DaemonSet'
+
```
NicolasT: Even though we currently don't use them yet, setting `controller.stats.enabled=true` may be useful.

roles/kube_nginx_ingress/tasks/main.yml
```diff
@@ -1,16 +1,29 @@
...
+- name: 'copy nginx-ingress values into temporary file'
+  copy:
+    src: nginx_ingress_values.yml
+    dest: '{{ nginx_ingress_value_file.path }}'
+
 - name: 'install nginx ingress'
   command: >-
-    helm install stable/nginx-ingress
-    --name nginx-ingress
+    helm upgrade --install nginx-ingress stable/nginx-ingress
```
NicolasT: When using `--repo`, this'll be `nginx-ingress` without `stable/`.

roles/kube_nginx_ingress/tasks/main.yml
Zempashi: I would like to find a proper path on the system to allow futur admin to edit and test modification directly (without ansible), but now it's a temporary file in /tmp/, it's commonly admitted that you need to cleanup your file when you finished with it ?

roles/kube_nginx_ingress/tasks/main.yml
Zempashi: Yes ^^

roles/kube_nginx_ingress/tasks/main.yml
NicolasT: No no, I meant, there's a typo :smiley:

PR #8, opened 2018-03-16
Ping as the first task of the playbook

Zempashi:


metal-k8s.yml
```diff
@@ -1,6 +1,11 @@
 - hosts: k8s-cluster
   gather_facts: False
-  roles:
+  tasks:
+    - ping:
+
+- hosts: k8s-cluster
+  gather_facts: False
+  tasks:
```
Zempashi: Hey ! thats "roles" here ...

NicolasT: Hmh, and same above I guess?

PR #9, opened 2018-03-16
Deploy Heapster

NicolasT:


NicolasT:
I'll wait for the Prometheus stuff to land, then remodel this one based on it, such that it includes `wait_for_helm` etc.

metal-k8s.yml
```diff
@@ -39,6 +39,9 @@
       tags: ['ingress']
     - role: kube_prometheus
       tags: ['kube-prometheus']
+      tags: ingress
```
NicolasT: This seems like a merge gone wrong...

PR #10, opened 2018-03-16
Install kube_prometheus

Zempashi:


NicolasT:
I believe this misses the setting of `kubelet_authentication_token_webhook: true` on `k8s-cluster`?

PR #11, opened 2018-03-19
Import and run `ansible-hardening`

NicolasT:
This commit imports the OpenStack `ansible-hardening` project, and runs it on `k8s-cluster` hosts when deploying.

See https://github.com/scality/metal-k8s/commit/130e337f922c429b20ef0aa56f6bf1ad7ace5daa for more information (other commits simply subtree the upstream repository).


metal-k8s.yml
```diff
@@ -8,6 +8,11 @@
   roles:
     - role: python2_install
 
+- name: Harden all systems
+  hosts: k8s-cluster
+  roles:
+    - ansible-hardening
```
Zempashi: I would see something like:
`
roles:
    - role: ansible-hardening
      when: enable_hardening|default(True)|bool
`

To optionnaly opt-out hardening if needed (for developement purpose)
Note: for non-ansible-speaking person: It's enabled by default

NicolasT:
@Zempashi What do you think?

ansible.cfg
```diff
@@ -0,0 +1,2 @@
+[defaults]
+roles_path = ansible-hardening
```
Zempashi: Why not puting ansible-hardening in the roles directory ?

NicolasT:
Complete rework compared to the old version of this PR.

NicolasT:
@ballot-scality @Zempashi One thing I noticed is this role enables SELinux, and Kubespray disables it again. I may look to not let this role enable SELinux in the first place.

PR #14, opened 2018-03-20
Install kube_prometheus

Zempashi:


roles/kube_prometheus/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+debug: False
+
+remove_metal_k8s_temporary_file: True
+
+prometheus_operator_version: '0.0.15'
+prometheus_operator_namespace: 'kube-ops'
+prometheus_operator_repo: 'https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/'
+prometheus_operator_release_name: 'prometheus-operator'
+
+kube_prometheus_version: '0.0.30'
```
NicolasT: There's a 0.0.33 already in `index.yaml` now... So may remove the need to override things?

NicolasT:
@Zempashi Looks like you pushed a `kube_prometheus` branch for 0.0.33, but this PR is about another branch :)

PR #15, opened 2018-03-21
Ft/zenko 160 add lvm storage

ballot-scality:
This PR brings the setup of an local LVM based storage for K8S PVs.

You need to setup either group_vars file or host_vars files containing the attributes found into roles/setup_lvm/defaults/main.yml and roles/kube_lvm_storage/defaults/main.yml

```
# ################ #
# LVM confguration #
# ################ #

# Specify which VG and which drive to use in host_vars for each node
metal_k8s_lvm:
  vgs:
    kubevg:
      drives: []


# Set the storage class setup on kubernetes node
metal_k8s_storage_class:
  storage_classes:
    local-lvm:
      is_default: true
  lvm_conf:
    default_fstype: 'ext4'
    default_fs_force: False
    default_fs_opts: '-m 0'
    default_mount_opts: 'defaults'
    vgs:
      kubevg:
        host_path: '/mnt/kubevg'
        storage_class: 'local-lvm'
        volumes:
          - name: lv01
            size: 10G
          - name: lv02
            size: 10G
          - name: lv03
            size: 10G
          - name: lv04
            size: 10G
          - name: lv05
            size: 10G

#
# You can also customize a specific volume like this
#
#    vgs:
#      kubevg:
#        volumes:
#          - name: lv01
#            size: 10G
#            fstype: xfs
#            fs_force: True
#            fs_opts: '-m 0 -cc'
#            mount_opts: 'defaults,noatime'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /*.retry
 /venv
+/host_vars
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /*.retry
 /venv
 /host_vars
+/inventories
--- a/metal-k8s.yml
+++ b/metal-k8s.yml
@@ -14,6 +14,14 @@
   roles:
     - role: setup_lvm
 
+- hosts: k8s-cluster
+  gather_facts: False
+  tags:
+    - kubespray
+  roles:
+    - role: kubespray/roles/kubespray-defaults/
+    - role: kubespray_var_patcher
+
 - import_playbook: 'kubespray/cluster.yml'
   tags: kubespray
 
--- a/metal-k8s.yml
+++ b/metal-k8s.yml
@@ -8,8 +8,20 @@
   roles:
     - role: python2_install
 
+- hosts: kube-node
+  tags:
+    - lvm-storage
+  roles:
+    - role: setup_lvm
+
 - import_playbook: 'kubespray/cluster.yml'
   tags: kubespray
+
+- hosts: kube-node
+  tags:
+    - lvm-storage
+  roles:
+    - role: kube_lvm_storage
 
 - hosts: kube-master
   gather_facts: False
--- a/roles/kube_lvm_storage/defaults/main.yml
+++ b/roles/kube_lvm_storage/defaults/main.yml
@@ -1,6 +1,51 @@
 debug: False
 
-lvm_mount_prefix: '/mnt/lvm'
-
 local_volume_provisioner_image_repo: quay.io/external_storage/local-volume-provisioner
 local_volume_provisioner_image_tag: v2.0.0
+
+# ################ #
+# LVM confguration #
+# ################ #
+#
+# Specify which VG and which drive to use in host_vars for each node
+metal_k8s_lvm:
+  vgs:
+    kubevg:
+      drives: []
+
+# Set the storage class setup on kubernetes node
+# possible values are "lvm" or "local-storage"
+metal_k8s_storage_class:
+  type: "lvm"
+  lvm:
+    default_fstype: 'ext4'
+    default_fs_force: False
+    default_fs_opts: '-m 0'
+    default_mount_opts: 'defaults'
+    vgs:
+      kubevg:
+        volumes:
+          - name: lv01
+            size: 10G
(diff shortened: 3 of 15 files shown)


NicolasT:
```
metal_k8s_storage_class:
  storage_classes:
    local-lvm:
      is_default: true
  lvm_conf:
    default_fstype: 'ext4'
    default_fs_force: False
    default_fs_opts: '-m 0'
    default_mount_opts: 'defaults'
    vgs:
      kubevg:
        host_path: '/mnt/kubevg'
        storage_class: 'local-lvm'
        volumes:
          - name: lv01
            size: 10G
          - name: lv02
            size: 10G
          - name: lv03
            size: 10G
          - name: lv04
            size: 10G
          - name: lv05
            size: 10G
```
I don't think this is the right approach.

In general, one will want to create one LVM VG per 'storage type', e.g. over a bunch of SSD-backed LVM PVs ('vg-ssd'), or some spindles ('vg-hdd')

Now, this means we should have 2 Kubernetes `StorageClass`es, and LVM LVs with Kubernetes PVs for said `StorageClass`es.

E.g.
```
metal_k8s_lvm:
  vgs:
    vg-ssd:
      drives: ['/dev/sdb', '/dev/sdc']
    vg-hdd:
      drives: ['/dev/sdd', '/dev/sde']

metal_k8s_storageClasses:
    - name: local-fast
      default: True
      lvmVG: vg-ssd
      hostPath: /mnt/local-fast
      fs:
        type: ext4
        options: '-m 0 --optimize-for-ssd' #Haha
        mountOptions: 'defaults'
      
    - name: local-slow
      lvmVG: vg-hdd

metal_k8s_volumes:
    - name: lv01
      storageClassName: local-fast
      size: 10Gi
    - name: lv02
      storageClassName: local-slow
      size: 128Ti
```

Exact naming of fields etc. bikeshed'able.

xaf-scality:
I agree with @NicolasT, this approach makes more sense in a TS perspective.
Moreover I would like to know what is the current state of art in Eng for specifying volume space. Is it:

- GB
- G
- Gi
- GiB
- All of them ?

ballot-scality:
Okay.

@xaf-scalitywhatever is supported by the "lvol" module : http://docs.ansible.com/ansible/latest/lvol_module.html
So in this case "G|g" is the only valid choice

@NicolasT Okay, although this model allow precisely what you describe (this was the intention)

Now I Agree I could have some defaults for the filesystem and mount options and a way to override that per VG and by LV compared to only LV right now, so they would make sense for a type of storage.

However I disagree on the data modeling of the storage classes. I found this a recurrent "issue" in the ansible world and discussed it with @Zempashi , if you use a list, you have to manually check for the name collision as opposite of a dict, where the key is by design unique.

So I would rather do something like this:

```
metal_k8s_lvm:
  vgs:
    vg-ssd:
      drives: ['/dev/sdb', '/dev/sdc']
    vg-hdd:
      drives: ['/dev/sdd', '/dev/sde']

metal_k8s_storageClasses:
  local-fast:
    default: True
    lvmVG: vg-ssd
    hostPath: /mnt/local-fast
     fs:
       type: ext4
       options: '-m 0 --optimize-for-ssd' #Haha (<- I see what you did there !)
       mountOptions: 'defaults,nattime'
  local-slow
    lvmVG: vg-hdd

metal_k8s_volumes:
  default_fs:
    type: ext4
    options: '-m 0 --barrier=0' #IWantToLoseMyData
    mountOptions: 'defaults'
  volumes:
    lv01:
      storageClassName: local-fast
      size: 10G
    lv02:
      storageClassName: local-slow
      size: 128T
      fs:
        type: xfs
        options: '-m 0'
```
What do you think ?

xaf-scality:
It's ok for me, what I will challenge is the #Haha and the #IWantToLoseMyData stuff ;-)

Zempashi:
Tested, works for me and Zenko.

Note: according to https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume

> Note that with manual PV creation, the volume has to be manually reclaimed when deleted
Confirm, I experienced this as well.

PR #22, opened 2018-03-22
Ft/zenko 160 fix lvm storage

Zempashi:


group_vars/k8s-cluster/10-metal-k8s.yml
```diff
@@ -5,6 +5,9 @@ kube_basic_auth: True
 kubeconfig_localhost: True
 kubelet_authentication_token_webhook: True
 kube_api_pwd: "{{ lookup('password', playbook_dir + '/credentials/kube_user length=15 chars=ascii_letters,digits') }}"
+kube_feature_gates:
```
NicolasT: I decided to not get rid of var-patcher, because now we overwrite `kube_feature_gates`, even if Kubespray would enable some gate by default, or whatever. Sucks we can't 'append' to a default list in Ansible.

Zempashi: But that's the all point no ? To handle the case of kubespray upgrade ?

PR #23, opened 2018-03-22
feat(scale): add `scale-metal-k8s` playbook

NicolasT:


Zempashi:
Isn't it the same as 
`ansible-playbook metal-k8s.yml -l kube-node` ?

NicolasT:
Don't think so... it delegates to `kubespray/scale.yml`, not `kubespray/cluster.yml`.

PR #24, opened 2018-03-23
grafana suburl

Zempashi:


roles/kube_prometheus/files/grafana_ingress.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  namespace: 'kube-ops'
```
NicolasT: Hmh, maybe don't put `namespace` here, but pass it to the `kube` step in the playbook.

Zempashi: But I don't like having same piece of information in differente places. If an admin go to the server and want to just modify something and apply, it will spawn a Ingress in default or something like that.

I can put a template, and a variable if you want

NicolasT: Good point. On the other hand, I was wondering why we're leaving these deployment files on the server anyway? Why not remove them as we do for the Helm values files?

roles/kube_prometheus/files/grafana_ingress.yml
```diff
@@ -0,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: "kube-prometheus-grafana-ingress"
```
NicolasT: Could make sense to add labels here as well over time, cfr https://github.com/scality/metal-k8s/pull/30#discussion_r176879125

```yaml
app: kube-prometheus-grafana
release: kube-prometheus
```

PR #25, opened 2018-03-23
Deploy `metrics-server`

NicolasT:


roles/kube_metrics_server/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    - resource-reader.yaml
+  run_once: true
+  delegate_to: "{{ groups['kube-master'][0] }}"
+
+- name: 'Deploy metrics-server'
+  kube:
+    kubectl: '{{ bin_dir }}/kubectl'
+    filename: '{{ ms_addon_dir }}/{{ item }}'
+    state: 'latest'
+  with_items:
```
Zempashi: Avoiding repetition of files, could be done via `register: ms_manifests` on template task, 
and `with_items: '{{ ms_manifests.results|default([])|map(attribute='item')|list }}'` on this one

PR #26, opened 2018-03-23
cert-manager with self-signed ca

Zempashi:
`kubectl describe clusterissuer`
```
[...]
Events:
  Type     Reason         Age               From                     Message
  ----     ------         ----              ----                     -------
  Warning  ErrGetKeyPair  1s (x10 over 4s)  cert-manager-controller  Error getting keypair for CA issuer: error parsing private key: asn1: structure error: tags don't match (2 vs {class:0 tag:16 length:13 isComp
ound:true}) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false}  @5
  Warning  ErrInitIssuer  1s (x10 over 4s)  cert-manager-controller  Error initializing issuer: error parsing private key: asn1: structure error: tags don't match (2 vs {class:0 tag:16 length:13 isCompound:true}
) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false}  @5
```
If someone has an idea ;)

### DIFF
--- a/roles/kube_cert_manager/defaults/main.yml
+++ b/roles/kube_cert_manager/defaults/main.yml
@@ -1,0 +1,12 @@
+debug: False
+
+cert_manager_version: '0.2.4'
+cert_manager_namespace: 'kube-ingress'
+cert_manager_repo: 'https://kubernetes-charts.storage.googleapis.com'
+cert_manager_release_name: 'cert-manager'
+
+ca_key_localpath: '{{ playbook_dir }}/group_vars/k8s-cluster/11-zenko-ca-private-key.yml'
+ca_cert_localpath: '{{ playbook_dir }}/group_vars/k8s-cluster/11-zenko-ca-cert.yml'
+kube_ca_cn: example.com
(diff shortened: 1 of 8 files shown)


roles/kube_cert_manager/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+debug: False
+
+cert_manager_version: '0.2.4'
+cert_manager_namespace: 'kube-ops'
```
NicolasT: Maybe `kube-ingress` is more appropriate?

Zempashi: yes, indeed...

Zempashi:
Problem above: fixed (openssl_private key generate private key with ` -----BEGIN PRIVATE KEY-----` instead of `-----BEGIN RSA PRIVATE KEY-----` header)

PR #30, opened 2018-03-23
Deploy ingresses for Prometheus and AlertManager

Zempashi:


roles/kube_prometheus/files/prometheus_ingress.yml
```diff
@@ -0,0 +1,19 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: "kube-prometheus-prometheus-ingress"
```
NicolasT: Could you add labels matching the Helm chart?

```yaml
app: prometheus
release: kube-prometheus
```
Or wouldn't that make sense?

Zempashi: For the demo ?

NicolasT: Not necessary for the demo, but since you're on this anyway, why not add it now. Easy to forget later.

NicolasT: Could it  make sense to have the `/_/grafana/` definition in the very same file? We don't need 2 Ingress objects, right?

roles/kube_prometheus/files/prometheus_ingress.yml
```diff
@@ -0,0 +1,19 @@
...
+    nginx.ingress.kubernetes.io/ssl-redirect: "false"
+    nginx.ingress.kubernetes.io/rewrite-target: "/"
+spec:
+  rules:
+    - http:
+        paths:
+          - path: '/_/prometheus'
+            backend:
+              serviceName: "kube-prometheus-prometheus"
+              servicePort: 9090
```
NicolasT: Consider using named ports (twice `http`) instead of numerical ones.

PR #32, opened 2018-03-26
Fix the mismatch format between lvm size format and kubernetes one.

Zempashi:


roles/kube_lvm_storage/filter_plugins/size_lvm_to_k8s.py
```diff
@@ -0,0 +1,13 @@
+
+def size_lvm_to_k8s(lvm_size):
+    try:
+        return int(lvm_size)
+    except ValueError:
+        if not lvm_size.lower().endswith('b'):
```
NicolasT: Bit on the edge here: what if it already ends with `i`? (Though in that case `lvcreate` likely bailed out before).

We may want to have a somewhat more resilient approach here later on.

Zempashi: That's why I end up in pure python. Handle all cases. Should do the trick right now, no ?

PR #38, opened 2018-04-16
ZENKO-240: Add lvm2 package

ballot-scality:


roles/setup_lvm/tasks/main.yml
```diff
@@ -4,16 +4,21 @@
...
 
+- name: "LVM Setup: Check LVM packages"
+  package:
+    name: lvm2
+    state: present
+
 - name: start lvm storage setup
   debug:
-    msg: "Start LVM storage setup"
+    msg: "LVM Setup: Start LVM storage setup"
```
Zempashi: Is this tasks needed ? The name os the task is printed as well as the debug output: You're displaying basically (n+1)  times "start lvm storage setup" (one for the name of the task, one for each server you run on).

ballot-scality: Oh ok, removing it then

PR #41, opened 2018-04-20
Ft/zenko 255 support centos7

Zempashi:


roles/helm_common/defaults/main.yml
```diff
@@ -0,0 +1 @@
+bin_dir: /usr/local/bin
```
NicolasT: This is already set elsewhere, no?

roles/check_os/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+
+- name: 'Assertion on OS'
```
NicolasT: Is there a way to 'override' this assertion?

roles/helm_common/defaults/main.yml
Zempashi: Yes, but considering that:
- bin_dir is misnamed (bin_dir="/usr/bin" ?) and the name of the variable could change
- /usr/local/bin/ is the preferable destination to install helm without package
- we will detect rapidly if helm has been changed of location
- kubespray-defaults is slow

I find reasonnable to have the duplication of the variable

roles/check_os/tasks/main.yml
Zempashi: Done

roles/helm_common/defaults/main.yml
Zempashi: But I don't care, so I could change I you want (to be more opensource friendly or whatever)

PR #43, opened 2018-04-24
Add a framework for pre-flight checks, and implement some

NicolasT:
This commit adds an Ansible 'action plugin' which is used to run various
pre-flight checks against a given inventory and configuration, to ensure
consistency or guard against common errors.

It also implements a couple of basic validations.

### DIFF
--- a/metal-k8s.yml
+++ b/metal-k8s.yml
@@ -1,3 +1,11 @@
+- hosts: localhost
+  gather_facts: False
+  any_errors_fatal: True
+  roles:
+    - role: preflight_checks
+  tags:
+    - preflight-checks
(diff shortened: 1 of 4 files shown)


NicolasT:
Example:
screenshot from 2018-04-24 12-47-42

NicolasT:
I guess, in theory, many of these could be (declaratively) expressed as `assert` tasks with `when` clauses. Some, however, could be... difficult to code up this way (thinking of the 'address reuse' check here).

As such I decided to put everything in one scheme instead. Maybe not the 'cleanest' solution, but arguably the most effective and versatile.

PR #47, opened 2018-04-25
Add a `shell` target

NicolasT:
These commits add a `shell` target to `Makefile`, which sets up an environment containing a supported Ansible version and all its dependencies (in a managed virtualenv), as well as `kubectl` and `helm`.

It also adds a `requirements.txt` lockfile, and tooling to re-generate this file when desired.

This is very much related to #34, but doesn't require Ansible (where we rely on a specific version) to be pre-installed on the system, rather `tox` which we don't really need at runtime at all.


NicolasT:
Tested on my laptop (`tox 2.7.0`, as shipped with Fedora) as well as CentOS 7.4, by
```
$ sudo yum install epel-release
$ sudo yum install python2-pip
$ pip install --user tox
$ tox -e shell
```

NicolasT:
For future reference: https://asciinema.org/a/cXlgjUVSlaWut1iUEerdVUtbT

Zempashi:
I'm a bit confused, we don't want makefile but tox is ok ?

Zempashi:
I'm considering removing the need of ansible in https://github.com/scality/metal-k8s/pull/34

NicolasT:
This PR has now been reworked to use a `Makefile`, relying on nothing but `make`, `python` and some very standard tools (`tar`, `sha256sum`) to be available on the system.

Tested on F26 and CentOS 7.4.

Zempashi:
Update of the README to tell about "make shell" (usage section ?) could be nice.

NicolasT:
Yes. I intend to tackle documentation more holistically in another branch/PR, hence didn't spend time on that (yet).

Makefile
```diff
@@ -0,0 +1,103 @@
...
+
+KUBECTL_VERSION := 1.10.1
+KUBECTL_SRC_SHA256SUM := 1bb4d3793fb0f9e1cfee86599e0f43ae5f15578a01b61011fe7c9488e114a00b
+
+HELM_VERSION := 2.8.2
+HELM_SRC_SHA256SUM := 614b5ac79de4336b37c9b26d528c6f2b94ee6ccacb94b0f4b8d9583a8dd122d3
+HELM_BIN_SHA256SUM := 0521956fa22be33189cc825bb27b3f4178c5ce9a448368b5c81508d446472715
+
+UNAME_KERNEL = $(shell uname -s | tr '[:upper:]' '[:lower:]')
+UNAME_MACHINE = $(shell if [ "x$$(uname -m)" == "xx86_64" ]; then echo amd64; else uname -m; fi)
```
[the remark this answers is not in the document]
NicolasT: Hmh, good point. I'll apply that change.

Zempashi:
"kubectl get pods" doesn't work for me, but there is no export of admin.conf so...Is this intended ?

NicolasT:
@Zempashi Yes, that's intended. As I said, in newer Kubespray versions, `admin.conf` is copied to a directory within the inventory, so there's no way to easily know it from the `Makefile`.

Documenting how to export `KUBECONFIG` and educating people to do so, seems more suited.

Makefile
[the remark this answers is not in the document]
NicolasT: Done

Makefile
```diff
@@ -0,0 +1,103 @@
...
+$(VENV_EXISTS): $(PYTHON_VIRTUALENV_EXISTS)
+	@rm -rf $(VENV)
+	@echo "Creating virtualenv..."
+	@PYTHONPATH=$(PYTHON_VIRTUALENV) $(PYTHON) -m virtualenv $(VENV) > /dev/null
+	@touch $@
+
+REQUIREMENTS_INSTALLED = $(SHELL_ENV)/.requirements_installed
+$(REQUIREMENTS_INSTALLED): $(VENV_EXISTS) requirements.txt
+	@echo "Installing Python dependencies..."
+	@source $(VENV_ACTIVATE) && pip install -r requirements.txt > /dev/null
```
ballot-scality: Because the ($shell ) directive use "sh" and this is "dash" by default on ubuntu, this fails.

Putting a "." fixes the issue

Makefile
```diff
@@ -0,0 +1,103 @@
...
+	@ln -sf $(shell realpath $(HELM_BIN)) $@
+
+BASHRC = $(SHELL_ENV)/bashrc
+$(BASHRC): $(SHELL_ENV_EXISTS) hack/shell-bashrc
+	@rm -f $@
+	@sed s:@VENV_ACTIVATE@:$(VENV_ACTIVATE):g < hack/shell-bashrc > $@ || (rm -f $@; exit 1)
+
+shell: $(VENV_EXISTS) $(REQUIREMENTS_INSTALLED) $(KUBECTL) $(HELM) $(BASHRC)
+	@echo "Launching metal-k8s shell environment. Run 'exit' to quit."
+	@bash --rcfile $(BASHRC)
```
ballot-scality: Need to add a ||: so if the last command has an error exit code, it doesn't display 

```
(metal-k8s) [ballot@OnePower metal-k8s (tox-shell *=)]
$ exit
exit
Makefile:97: recipe for target 'shell' failed
make: *** [shell] Error 2
```

NicolasT: I considered that before, but wasn't sure 'hiding' shell failures was a good approach. Fair enough, done.

PR #48, opened 2018-04-27
Initial documentation

NicolasT:
Add some initial documentation, based on Sphinx.


NicolasT:
Would :heart: approval from @thomasdanan once completed the Quickstart guide.

README.md
```diff
@@ -1,13 +1,22 @@
...
+For the really impatient, there's a [quickstart] available.
 
-## Test config
-If you want  to test metal-k8s you just need one simple VM running ubuntu
- 16.04
+[quickstart]: https://metal-k8s.readthedocs.io/en/latest/usage/quickstart.html
 
-## Production Config
+## Documentation
+The project documentation is available at https://metal-k8s.readthedocs.io.
```
LaureVergeron: Shouldn't this be a link?

README.md
```diff
@@ -1,13 +1,22 @@
...
+Sources can be found in the `docs` folder of this repository. Pull-requests
+welcome!
+
+## Hardware Requirements
+
+### Test config
+If you want  to test metal-k8s you just need one simple VM running CentOS 7.4.
```
LaureVergeron: double space

README.md
```diff
@@ -17,56 +26,5 @@ For production-ready environment, you need:
...
-## Usage
-You also need `kubectl` and `helm` binary to play with your cluster from your
-local computer.
-
-Please refer to the official documentation to install those tools:
-
-- https://kubernetes.io/docs/tasks/tools/install-kubectl/
-- https://docs.helm.sh/using_helm/#installing-helm
+## Software Requirements
+To get started, all you need is `python` and `make` on a Unix-like system.
```
LaureVergeron: which versions?

NicolasT: Doesn't matter, which is exactly the point.

README.md
NicolasT: It is.

README.md
NicolasT: @wabernatScality Feedback welcome at any point in time as far as I'm concerned :smile: Though maybe a bit too early now, I'll add more content first. Thanks for the :+1:!

README.md
NicolasT: Fixed

NicolasT:
Change since previous reviews: addition of the *Cluster Services* document, see https://github.com/scality/metal-k8s/blob/docs/docs/architecture/cluster-services.rst

NicolasT:
Agree we should try to get patterns set up in Kibana by default, though that's out of scope for this PR :) I guess something should be feasible through a Helm `init job`.

PR #54, opened 2018-05-17
Update Helm version

NicolasT:
Updating to Helm 2.8.2 fixes a timeout issue several users ran into,
especially on slow networks. When doing so, why not simply upgrade to
the latest current release.

See: https://github.com/kubernetes/helm/releases/tag/v2.8.2
See: https://github.com/kubernetes/helm/releases/tag/v2.9.0
See: https://github.com/kubernetes/helm/releases/tag/v2.9.1

PR #55, opened 2018-05-17
Update third party

NicolasT:
Update
- deployed Helm chart versions
- ElasticSearch deployment (updating ES version)
- Kubespray (among other things, upgrade to Kubernetes 1.10.2)

Tested by upgrading a deployment (caveat: setting `helm_version` to `v2.9.1`, because the cluster was used to test #54)


Zempashi:
Upgrade from kubernetes 1.9 works ?

NicolasT:
> Upgrade from kubernetes 1.9 works ?

I had a 1.9.5, after running the playbook I had a 1.10.2, and all pods were in `Running` state, so I guess so... We don't have any more conformance checks right now...

Issue #37, opened 2018-04-12
etcd3-only nodes are handled incorrectly


NicolasT:
If one uses separate nodes for the etcd3 cluster, they're not pinged at the start of the playbook, and `docker_dns_servers_strict=no` is not set for them.

PR #57, opened 2018-05-18
Ping all nodes at playbook start, and re-align `scale` playbook

NicolasT:
See #37 

Also, we should find a way to easier keep the `main` playbook and the `scale` playbook in sync. Using tags, maybe? Pinging @Zempashi @alxf @ballot-scality

### DIFF
--- a/metal-k8s.yml
+++ b/metal-k8s.yml
@@ -6,7 +6,7 @@
   tags:
     - preflight-checks
 
-- hosts: k8s-cluster
+- hosts: k8s-cluster:etcd
   gather_facts: False
   tasks:
     - ping:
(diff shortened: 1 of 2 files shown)


NicolasT:
@Zempashi @alxf In order to merge playbook manifests into one (and support deploy/upgrade/scale from one entry-point), could we use something like https://github.com/ansible/ansible/issues/31747, then

```shell
$ ansible-playbook metal-k8s.yml # Runs 'deploy' mode
$ ansible-playbook -e metalk8s_mode=scale metal-k8s.yml # Runs 'scale' mode
$ ansible-playbook -e metalk8s_mode=upgrade metal-k8s.yml # Runs 'upgrade' mode
```
All of the above would correctly import the relevant Kubespray playbook file, and e.g. run or skip the deployment of the Helm packages we use, etc.

Issue #36, opened 2018-03-29
Review need for unauthenticated read-only API port and Heapster


NicolasT:
When Heapster was added in https://github.com/scality/metal-k8s/commit/71206ef027347d4386f0109aa88848929c10a42c#diff-384fbcda7fecbabd124cf2ccae3167cdR9, an unauthenticated API server access mechanism was opened as well to make things work.

This is not desirable. Luckily there appears to be a way to get Heapster (and maybe metrics-server?) to work without it: https://github.com/kubernetes-incubator/kubespray/issues/2550#issuecomment-377145148

We may want to adopt this.

PR #59, opened 2018-05-18
Disable anonymous read-only API port

NicolasT:
This commit disables the anonymous (non-authenticated) read-only API
port, which is insecure, and adjusts the Heapster and `metrics-server`
configuration to take this into account.

Fixes: #36
Fixes: https://github.com/scality/metal-k8s/issues/36
See: https://github.com/kubernetes-incubator/kubespray/issues/2550#issuecomment-377145148
See: 71206ef027347d4386f0109aa88848929c10a42c
See: https://github.com/scality/metal-k8s/commit/71206ef027347d4386f0109aa88848929c10a42c#diff-384fbcda7fecbabd124cf2ccae3167cdR9

### DIFF
--- a/group_vars/k8s-cluster/10-metal-k8s.yml
+++ b/group_vars/k8s-cluster/10-metal-k8s.yml
@@ -6,9 +6,6 @@ kubeconfig_localhost: True
 kubelet_authentication_token_webhook: True
 kube_api_pwd: "{{ lookup('password', playbook_dir + '/credentials/kube_user length=15 chars=ascii_letters,digits') }}"
 
-# Currently required for Heapster kubelet access
-kube_read_only_port: 10255
-
 # Request usage of the `overlay2` storage driver, even on pre-18.03 Docker
 # installs.
 # Whilst this is not guaranteed to work on 'old' kernels, we check whether we're
--- a/roles/kube_heapster/files/values.yml
+++ b/roles/kube_heapster/files/values.yml
@@ -1,2 +1,9 @@
 rbac:
   create: true
+
+command:
+- "/heapster"
+# Chart default:
+# - "--source=kubernetes.summary_api:''"
+# For https://github.com/scality/metal-k8s/issues/36:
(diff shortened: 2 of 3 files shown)


ssalaues:
I tried testing this branch with a fresh 3 node install the below in Firefox when trying to connect to the ui. Perhaps this is an unintended side effect of disabling the read only port?
```
  |  
-- | --
paths |  
0 | "/apis"
1 | "/apis/"
2 | "/apis/apiextensions.k8s.io"
3 | "/apis/apiextensions.k8s.io/v1beta1"
4 | "/healthz"
5 | "/healthz/etcd"
6 | "/healthz/ping"
7 | "/healthz/poststarthook/generic-apiserver-start-informers"
8 | "/healthz/poststarthook/start-apiextensions-controllers"
9 | "/healthz/poststarthook/start-apiextensions-informers"
10 | "/metrics"
11 | "/openapi/v2"
12 | "/swagger-2.0.0.json"
13 | "/swagger-2.0.0.pb-v1"
14 | "/swagger-2.0.0.pb-v1.gz"
15 | "/swagger.json"
16 | "/swaggerapi"
17 | "/version"
```

NicolasT:
Ugh, you're right! Seems like a blocker to me :smile: (and shows once again we need some kind of conformance tests...)

NicolasT:
Looks like this may be caused by #55 or something else, not sure current `master` works...

NicolasT:
So, well, no, it makes sense. The change is not due to this PR, but due to the switch to Kubernetes 1.10 through #55. In 1.10, the `/ui` shortcut no longer exists, and one needs to use the full service URL, see the first *Note* in https://github.com/kubernetes/dashboard/blob/9485b641b0cc6008c2fbd5f5e1a927a3b40ef255/README.md#getting-started

I guess this should be documented now, and maybe we should add the required annotations to the Dashboard service such that the URL is displayed in `kubectl cluster-info`.

ssalaues:
Ahh I see and yeah it works just fine with that URL! But yeah it would be a good idea to get it documented and added to the required annotations.

I agree with the conformance tests aswell. We could do a single node deployment with EVE and an open stack worker to test deployment then run at least some basic conformance tests.

PR #61, opened 2018-05-23
light copyedit

wabernatScality:
I got rid of a lot of passive constructions, which made agency unclear to (this) lay reader.


docs/index.rst
```diff
@@ -18,13 +18,13 @@ customer datacenters.
 .. _Scality: https://www.scality.com
 .. _Zenko: https://www.zenko.io
 
-It is based on the Kubespray_ project to reliably install a base Kubernetes
+MetalK8s, based on the Kubespray_ project, reliably installs a base Kubernetes
```
NicolasT: No. The intention is: it's based on Kubespray to reliably install a base Kubernetes cluster (which is Kubespray's job). The intent is not "MetalK8s reliably installs a base Kubernetes cluster".

docs/index.rst
```diff
@@ -18,13 +18,13 @@ customer datacenters.
...
 cluster, including all dependencies (like etcd_), using the Ansible_
-provisioning tool. This installation is further augmented with operational
-tools for monitoring and metering, including Prometheus_, Grafana_,
-ElasticSearch_ and Kibana_. Furthermore, an "ingress controller" is deployed
-by default, based on Nginx_. All of these are managed as Helm_ packages. See
-:doc:`architecture/cluster-services` for a whole listing.
+provisioning tool. MetalK8s augments this installation with operational
+monitoring and metering tools, including Prometheus_, Grafana_, ElasticSearch_
```
NicolasT: No, they're not "operational monitoring and metering tools". They're tool for operational purposes, including monitoring and metering.

docs/index.rst
```diff
@@ -18,13 +18,13 @@ customer datacenters.
...
+and Kibana_. Furthermore, MetalK8s deploys an "ingress controller" based on 
+Nginx_ by default. All of these are managed as Helm_ packages by MetalK8s. See
```
NicolasT: Don't see the use for this "by MetalK8s". The name is coined too many times already in this paragraph. Furthermore, MetalK8s doesn't really manage those packages at all.

docs/index.rst
wabernat: Generally speaking, the purpose of technical communication to describe the things we make, not the reasons we had for building them. Because of this, I have a strong bias towards moving documentation away from answering "why" questions and toward "what" and "how" ones.

The original sentence starts with an ill-defined pronoun, which introduces an ambiguity: "It is..." followed by the infinitive "to...install" can be easily misread to imply that the act of installing a base Kubernetes cluster has the Kubespray project as a precondition. Though this makes no sense, it's not enough to keep people from tripping over the sentence, which is my other reason for suggesting the change.  

So...if you take the intention out of the sentence, you're left with what MetalK8s does. If my edits aren't quite right, at least you know where I'm misunderstanding the sentence and why I made this change.

docs/index.rst
wabernat: The logic of the original sentence is as follows:
* There exists an installation.
* The installation is augmented.
* The thing augmenting the installation is a set. 
* This set is described as "operational tools for metering and monitoring".
* This set contains: Prometheus, Gafana, Elasticsearch, and Kibana.
 
As I read your response, the second-to last point is where the error crept in. If this assessment is accurate, my suggestion is to unpack this phrase. If I read you correctly, the sentence should be: 

"MetalK8s augments this installation with operational tools, including Prometheus, [etc...], for monitoring, metering, and other purposes."

docs/index.rst
wabernat: 1) Don't worry about the repetition. It's better not to be dull, but I'll pick "dull" over "vague" every time. 
2) The original sentence, iirc, was awash in passive constructions, which is why I injected a subject (apparently incorrectly). The passive voice enables us to elide the subject, and that makes it hard for an uninformed reader to find out, what's going on in the product. There's no hard-and-fast rule about this, but if the agent/subject looks clear, I'll add it. I did check with another local engineer before making that change, but it's clearly a wild pitch. 

If not MetalK8s, which component deploys the ingress controller? Does MetalK8s trigger that component's deployment of the ingress controller?

docs/index.rst
NicolasT: First of all, note the target audience of these docs is not only *users*, but also (potential) *contributors* and people who want to *understand* this project, so rationales make sense. This is not a 'Quickstart guide'.

To make it clear, MetalK8s is not *based* on Kubespray (which would make it a fork). Indeed, this is not correctly phrased in the original text, in retrospect.
It *includes* Kubespray to handle one part of what MetalK8s does (this part being 'installing a base Kubernetes cluster, reliably'). It also includes or deploys various other projects.

I fail to see, by the way, why the 'It' is ill-defined: the links in-between both paragraphs may throw you off, but this is not part of the rendered documents. In the rendered version (currently at https://metal-k8s.readthedocs.io/en/latest/) it seems quite clear to me what the 'It' refers to.

docs/index.rst
NicolasT: Fair enough wrt the last suggestion.

docs/index.rst
wabernat: It is clear to you, I trust, that the first word in this sentence is not a reference to a previously-defined term, but introduces the concept soon to follow. This structure and placement is present in the subject sentence. When you start a paragraph, you imply that you're starting with a fresh concept, not continuing the previous one. A reader's natural tendency is going to be "New paragraph. New idea. I may need to reset the 'its' buffer." That the "it" in question above refers to the aforementioned distribution becomes clear enough after one reads the sentence three times. I want to make it clear without having to slow the reader down. Fundamentally, this sentence lacks euphony and requires struggle to parse. Though you know exactly what you're describing, the reader has to pull in every clue available in realtime, and your sentence leaves a trail of breadcrumbs into the woods. 

Based on what you've told me, I suggest:.

> MetalK8s uses Kubespray, among other tools, to install a base Kubernetes cluster, including all dependencies (like etcd), using the Ansible provisioning tool. 

If you're wedded to "it", closing ranks with the preceding paragraph will help, but the sentence remains buggy. 

> It is based on the Kubespray project, and used for installing a base Kubernetes cluster, including all dependencies (like etcd), using the Ansible provisioning tool.

PR #64, opened 2018-05-23
Doc: copyedit cluster-services.rst

wabernatScality:
Grammar, style edit. 
Nicolas: the " :program:`fluentd`" notation is not correct RST notation (see http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#field-lists) please fix or clarify what you want to accomplish with this notation.

### DIFF
--- a/docs/architecture/cluster-services.rst
+++ b/docs/architecture/cluster-services.rst
@@ -1,7 +1,5 @@
 .. spelling::
 
-   Addons
-   addons
    operability
    preconfigured
    prometheus
@@ -10,10 +8,11 @@
 
 Cluster Services
 ================
-A Kubernetes_ cluster deployed on the `Google Cloud Platform`_ using GKE_, on
-`Microsoft Azure`_ using AKS_ or even using Kops_ or similar tools on `Amazon
-AWS`_ comes with built-in tooling for centralized container log management,
-metrics collection, tracing, node health checking and more.
+When deployed on the `Google Cloud Platform`_ (using GKE_), on
+`Microsoft Azure`_ (using AKS_), or on `Amazon AWS`_ (using Kops_
+or similar tools) a Kubernetes_ cluster comes with built-in tooling for
+centralized container log management, metrics collection, tracing, node
+health checking, and more.
 
 .. _Kubernetes: https://kubernetes.io
 .. _Google Cloud Platform: https://cloud.google.com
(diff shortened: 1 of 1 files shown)


NicolasT:
Why would that be incorrect? `program` is a Sphinx-specific RST 'role', so
```
:program:`cat`
```
is perfectly valid: http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-program. Your link is about field lists, which is something different.

docs/architecture/cluster-services.rst
```diff
@@ -23,41 +22,40 @@ metrics collection, tracing, node health checking and more.
...
+the same level of operability.
 
 .. _MetalK8s: https://github.com/scality/metal-k8s/
 .. _Kubespray: https://github.com/kubernetes-incubator/kubespray/
 
-Basic Cluster Addons
+Basic Cluster Add-ons
 --------------------
-On top of the basic Kubernetes services, the following addons are deployed:
+The following add-ons are deployed atop the basic Kubernetes services:
```
NicolasT: `atop`?

docs/architecture/cluster-services.rst
```diff
@@ -23,41 +22,40 @@ metrics collection, tracing, node health checking and more.
...
 
 Heapster
 ********
-Heapster_ is a service which collects and exposes resource consumption metrics
-of containers running in a cluster. The Kubernetes Dashboard uses the Heapster
-service, when available, to display CPU and memory usage of Pods, Deployments
-and more.
+Heapster_ is a service that collects and exposes resource consumption metrics
+of containers running in a cluster. When the Heapster service is available, the
+Kubernetes Dashboard uses it to display CPU and memory usage of pods,
```
NicolasT: Pod, Deployment and others are very specific object types in K8s, hence capitalization. I'd like to keep it that way, maybe should even be emphasized as well, or become terms in a glossary.

docs/architecture/cluster-services.rst
```diff
@@ -97,27 +95,27 @@ Prometheus instances, scrape targets and alerting rules.
...
 .. _node-exporter: https://github.com/prometheus/node_exporter
 
 Grafana
 *******
-To ease cluster operations, several Grafana dashboards are made available,
-including cluster-wide views and health-checks, node OS metrics,
-per-*Deployment* or per-*Pod* resource usage, monitoring of the Prometheus
+To ease cluster operations, several Grafana dashboards are offered,
+providing cluster-wide views and health checks, node OS metrics,
+per-*deployment* and per-*pod* resource usage, monitoring of the Prometheus
```
NicolasT: Same as above.

docs/architecture/cluster-services.rst
```diff
@@ -126,20 +124,20 @@ Log Collection
...
 
-ElasticSearch Curator
+Elasticsearch Curator
 *********************
-To ensure ingested logs don't flood the ElasticSearch resources, `ElasticSearch
-Curator`_ is deployed with a default configuration which drops `logstash-*`
-indices on a given schedule.
+To ensure ingested logs don't flood the Elasticsearch resources,
+`Elasticsearch Curator`_ is deployed with a default configuration
+that drops `logstash-*` indexes on a given schedule.
```
NicolasT: `indices` is correct ;-)

docs/architecture/cluster-services.rst
```diff
@@ -157,11 +155,11 @@ stack.
...
 Kibana
 ******
-To give operators access to the logs stored in ElasticSearch, a `Kibana`_
+To give operators access to the logs stored in Elasticsearch, a `Kibana`_
 instance is provided.
 
-.. note:: When accessing Kibana for the first time, an *index pattern* for the
-   ``logstash-*`` indices needs to be configured, using ``@timestamp`` as *Time
+.. note:: When Kibana is first accessed, an *index pattern* for the
+   ``logstash-*`` indexes must configured, using ``@timestamp`` as *Time
```
NicolasT: `must be configured`.

wabernatScality:
The Git markdown parser does not know how to parse this construct and was rendering it as plain text. Clearly you wanted it to show up as something other than plain text, but I could not parse what the intention was either. So, I hunted down the RST spec and made the assumption that the field list link documentation, showing an RST snippet with identical syntax under the assertion "The following are **not** recognized..." meant it was faulty markdown syntax and that was where the story ended. 

Just for kicks, I ran the rst through pandoc to see if I could simply translate this into markdown. I got:
```
pandoc: ignoring unknown role :program: in "source" (line 151, column 32)
pandoc: ignoring unknown directive: todo "source" (line 125, column 1)
pandoc: ignoring unknown role :command: in "source" (line 41, column 54)
pandoc: ignoring unknown directive: spelling "source" (line 11, column 1)
```
So... even though the syntax is apparently correct, the :program: role remains undefined and difficult to divine.

Given the present context and the semantic limitations of gfm, when the reader sees :program:`fluentd`, Do you want the text to light up blue with a link, and if so, where do you want that link to point? If you do not want it represented as a link, would you prefer is to render as a string-literal code snippet? Or...?

NicolasT:
This documentation is using Sphinx. We use Sphinx to render it (at https://metal-k8s.readthedocs.io). As such, we can (and should) use Sphinx roles, where applicable. The fact that other RST processors may not know these roles doesn't change things.

If you want to render the docs locally, simply run `tox -e docs` in the root of the repository and see `docs/_build/html/`.

docs/architecture/cluster-services.rst
```diff
@@ -10,10 +8,11 @@
 
 Cluster Services
 ================
-A Kubernetes_ cluster deployed on the `Google Cloud Platform`_ using GKE_, on
-`Microsoft Azure`_ using AKS_ or even using Kops_ or similar tools on `Amazon
-AWS`_ comes with built-in tooling for centralized container log management,
-metrics collection, tracing, node health checking and more.
+When deployed on the `Google Cloud Platform`_ (using GKE_), on
+`Microsoft Azure`_ (using AKS_), or on `Amazon AWS`_ (using Kops_
+or similar tools) a Kubernetes_ cluster comes with built-in tooling for
```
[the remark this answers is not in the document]
wabernatScality: Identifying the Kubernetes engines contributes little to the actual meaning you're trying to convey. How 'bout: 

Platform-native Kubernetes engines ensure that Kubernetes_ clusters 
deployed on `Google Cloud Platform`_ , `Microsoft Azure`_ , or 
`Amazon AWS`_ come with built-in tooling for centralized container 
log management, metrics collection, tracing, node health checking and more.

docs/architecture/cluster-services.rst
wabernatScality: "On top of" is an inherently ambiguous phrase. It can mean A) "In addition to" or B) "physically or metaphorically resting on." 
So in the present instance, it could mean:
A) "In addition to basic K8s services, [whatever it is] also deploys the following add-ons..." or 
B) "When you deploy the following add-ons, they run on the basic K8s services..."

I picked B. Did I pick wrong?

docs/architecture/cluster-services.rst
wabernatScality: I am aware that Google does it this way. Google, however, is wrong. The mere fact that they consider their object types somehow special doesn't give them the right to arrogate proper-noun status to them. 

This said, there is some logic in letting the Wookiee win. I'm afraid you're going to have to bear with me on this -- roll back the corrections, but keep vigilant: from time to time, you will continue to catch me reverting to English over Googlese.

docs/architecture/cluster-services.rst
wabernatScality: ibid.

docs/architecture/cluster-services.rst
wabernatScality: Guilty!

PR #66, opened 2018-05-28
Kubespray credentials now under inventory_dir

Zempashi:


Zempashi:
Which file ? Without the PR, it's overriden (by us) to be in {{ playbook_dir + '/credentials/... }}. You have test with master of kubespray ?

metal-k8s.yml
```diff
@@ -57,14 +57,10 @@
...
-        kube_login: 'kube'
-        kube_password: >-
-          {{ lookup('file', 'kubespray/credentials/kube_user') }}
-        ip_master: >-
-          {{ hostvars[groups['kube-master'][0]].ansible_default_ipv4.address }}
+        add_info:
+        - master_dashboard_url: >-
+            https://{{ hostvars[groups['kube-master'][0]].ansible_default_ipv4.address
+            }}:6443/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy
+          kube_login: 'kube'
```
NicolasT: I guess what we really want is to retrieve this out of `kube_users`, but well ;)

metal-k8s.yml
```diff
@@ -57,14 +57,10 @@
...
-          kube_password: '{{ kube_password }}'
       vars:
-        kube_login: 'kube'
-        kube_password: >-
-          {{ lookup('file', 'kubespray/credentials/kube_user') }}
-        ip_master: >-
-          {{ hostvars[groups['kube-master'][0]].ansible_default_ipv4.address }}
+        add_info:
+        - master_dashboard_url: >-
+            https://{{ hostvars[groups['kube-master'][0]].ansible_default_ipv4.address
```
NicolasT: Use `external_apiserver_endpoint`, as also used in `kubespray/roles/kubernetes/client/templates/admin.conf.j2`?

metal-k8s.yml
Zempashi: (was not aware of kube_users). kube_users seems to be a dict containing all user, so how to extract the wanted user ? 
{{ kube_users.popitem() }} (a user at random ?) or something more complex like searching for the first user to be admin ? (but if multiple admin exist , as dict iteration is not consistent, this might be different each time)

NicolasT: No need to change now as far as I'm concerned.

We should, at some point, get rid of 'Basic auth' anyway :)

Zempashi: Ohhh, agreed on that

PR #68, opened 2018-05-30
Node selector use nodename instead of hostname. Use 1.10 API

Zempashi:
On AWS:
kubernetes.io/hostname: `ip-172-16-57-244.eu-west-2.compute.internal`
ansible_hostname: `ip-172-16-57-244`
ansible_nodename: `ip-172-16-57-244.eu-west-2.compute.internal`

On scality.cloud:
kubernetes.io/hostname: `kube-01`
ansible_hostname: `kube-01`
ansible_nodename: `kube-01`

### DIFF
--- a/roles/kube_lvm_storage/templates/local-pv.yml.j2
+++ b/roles/kube_lvm_storage/templates/local-pv.yml.j2
@@ -3,18 +3,6 @@ kind: PersistentVolume
 metadata:
   name: >-
     {{ item.value.vg|replace("_","-") }}-{{ item.value.volume.uuid }}
-  annotations:
-    "volume.alpha.kubernetes.io/node-affinity": ' {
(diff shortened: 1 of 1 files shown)


roles/kube_lvm_storage/templates/local-pv.yml.j2
```diff
@@ -24,3 +12,11 @@ spec:
...
   local:
     path: '{{ item.value.host_path }}/{{ item.value.volume.uuid }}'
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - {{ ansible_nodename }}
```
NicolasT: Whilst in this case it likely doesn't make a difference (`123` would not be a valid `ansible_nodename` anyway), consider quoting templated strings in YAML.

PR #69, opened 2018-05-30
Add storage for prometheus (10Gi as default)

Zempashi:


NicolasT:
In future, consider to point to (and link to) issue IDs (in this case https://github.com/scality/metal-k8s/issues/42) in commit messages, and maybe uses `Fixes:` prefixes etc.

PR #70, opened 2018-06-04
Update reclaim-storage to kubernetes 1.10 pv format

Zempashi:


Zempashi:
Yes indeed, it will not work on 1.9 cluster without an upgrade. I think that kube 1.10 is a good start for backward compat as first POC are being rolled-out.

NicolasT:
I wouldn't put backward-compat guarantees on K8s versions, rather on MetalK8s ones :wink:

PR #71, opened 2018-06-05
Fix info_displayer (var undefined)

Zempashi:


metal-k8s.yml
```diff
@@ -61,7 +61,7 @@
       vars:
         add_info:
         - master_dashboard_url: >-
-            {{ external_apiserver_endpoint
+            {{ hostvars[groups['k8s-cluster']|first].external_apiserver_endpoint
```
NicolasT: `kube-master` instead of `k8s-cluster`, no?

Also, a bit surprised in `kubespray/roles/kubernetes/client/templates/admin.conf.j2` `{{ external_apiserver_endpoint }}` can be used as-is. Why is an attribute lookup required here?

Zempashi: Yes but we are outside a role, so you're not attach to a specific host (yet). Yes for kube-master

Issue #67, opened 2018-05-29
Secure access to `kube-ops` services


NicolasT:
Currently we deploy an `Ingress` object for various browser-based services in `kube-ops`. Anyone who has access to one of the `kube-node` servers (on port 80) can access these, which is obviously problematic from a security PoV. As an example, one can access container logs through Kibana.

I suggest we:
- Remove these `Ingress` objects
- Deploy two `Role` definitions in `kube-ops`: one which has access to metrics and alerts (Prometheus, Grafana and Alertmanager), another which has access to logs (Kibana) throught the API proxy. I'm not sure how to call these, bikeshedding allowed.
- Add `cluster-service` labels on the service object
- Adjust e.g. Kibana's `SERVER_BASE_PATH` to match the API proxy URL
- Document how to access the services: running `kubectl proxy`, then use the correct service URLs

All of this should just work when using `kubectl` with `admin.conf`, yet would also permit more fine-grained access control once we deploy some kind of SSO/authn solution.

PR #72, opened 2018-06-05
Access operational services through K8s API proxy

NicolasT:
Prior to this, services like Grafana and Kibana were exposed (without
authentication) through the ingress controller running on all cluster
nodes, which is a security risk.

Instead, we now no longer expose these services through the ingress
proxy, but encourage administrators to access them through the K8s API
proxy (running `kubectl proxy`).

We aim to list the service URLs in the output of `kubectl -n kube-ops
cluster-info`, though these URLs will be pointing directly to a
`kube-master` IP address, and require to pass authentication in a
browser somehow. We document access through `kubectl proxy` instead.

This requires a role which has access to said API proxy endpoints,
currently e.g. `admin` (so things 'just work' when using when using the
`admin.conf` copied to the deployment host). We aim to create custom
roles which can then be bound to accounts later.

Note: this PR sets some labels on Grafana and Kibana, though these won't
show up in `kubectl -n kube-ops cluster-info` as long as the following
PRs aren't merged and we haven't updated the version of the related
Charts we deploy:

- https://github.com/kubernetes/charts/pull/5901
- https://github.com/coreos/prometheus-operator/pull/1426

Fixes: #67
See: https://github.com/scality/metal-k8s/issues/67


NicolasT:
https://github.com/coreos/prometheus-operator/pull/1426 has been merged, now waiting for progress on https://github.com/kubernetes/charts/pull/5901. In the meantime, will implement @Zempashi's 'fake service in `kube-system`' trick.

PR #74, opened 2018-06-06
Update third party

NicolasT:


Zempashi:
Would like to finish CI to test this one

NicolasT:
Fair enough :) No overly important changes anyway.

Now running in the network size calculation failure though:
```
fatal: [metalk8s-node-01]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was:

'kubelet_max_pods' is undefined

The error appears to have been in 'metal-k8s/kubespray/roles/kubernetes/preinstall/tasks/verify-settings.yml': line 69, column 3,
but may be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

# NOTICE: the check blatantly ignores the inet6-case
- name: Guarantee that enough network address space is available for all pods
  ^ here
"}
```
This gets ignored in the end, but may need to be investigated...

Zempashi:
What ???

NicolasT:
Because you renamed the `master` branch...
[edit] Ugh, no... no clue what's going on here.

NicolasT:
Looks like this branch was overwritten by something completely different :-/

PR #75, opened 2018-06-07
Simple CI with 1 node for metal-k8s

Zempashi:
CI on metal-k8s. Basic test on storage.

https://eve.devsca.com/github/scality/metal-k8s/#/builders/1/builds/51

TODO: 
- since last rebase elasticsearch is constently failing
- make flakyness assessment

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -1,0 +1,89 @@
+version: "0.2"
+
+branches:
+  default:
+    stage: "pre-merge"
+
+stages:
+  pre-merge:
+    worker: &pod
+      type: kube_pod
+      path: eve/workers/zenko.yaml
+    steps:
+    - TriggerStages:
+        name: trigger all the tests
+        stage_names:
+        - single-node
+
+  single-node:
+    worker: &openstack
+      type: openstack
+      image: CentOS 7 (PVHVM)
+      flavor: general1-4
+      #path: eve/workers/centos7
+    steps:
+    - Git: &git_pull
+        name: git pull
+        repourl: "%(prop:git_reference)s"
+        mode: full
+        method: clobber
+        retryFetch: true
+        haltOnFailure: true
+    - ShellCommand:
+        name: 'create loopback blockdevice and configure VM'
+        haltOnFailure: true
+        command: |-
+          sudo ./eve/single_node/setup/single_node.sh
+    - ShellCommand:
+        name: 'Create virtual env'
+        haltOnFailure: true
+        command: |-
+          make shell
+    - ShellCommand:
+        name: 'Run the install with ansible'
+        haltOnFailure: true
+        command: >-
+          ./.shell-env/metal-k8s/bin/ansible-playbook
+          -i eve/single_node/
+          metal-k8s.yml --skip elasticsearch
+        env:
+          ANSIBLE_FORCE_COLOR: 'true'
+    - ShellCommand:
+        name: 'Run sample kubectl command'
+        haltOnFailure: true
+        command: |-
+            export PATH=${PATH}:/usr/local/bin
+            kubectl get nodes
+        env: &env_kubeconfig
+          KUBECONFIG: inventories/single-node-ci/artifacts/admin.conf
+    - ShellCommand:
+        name: 'Test storage'
+        haltOnFailure: true
+        command: |-
+            export PATH=${PATH}:/usr/local/bin
+            make test-simple-storage
+        workdir: build/tests
+        env: *env_kubeconfig
+    - ShellCommand:
+        name: 'Verify that some pv are in released state'
+        command: |-
+          export PATH=${PATH}:/usr/local/bin
+          kubectl get pv
+          kubectl get pv -o jsonpath={.items[*].status.phase} | grep 'Released' > /dev/null
+        env: *env_kubeconfig
+    - ShellCommand:
+        name: 'Reclaim storage'
+        haltOnFailure: true
+        command: >-
+          ./.shell-env/metal-k8s/bin/ansible-playbook
+          -i eve/single_node/
+          reclaim-storage.yml
+        env:
+          ANSIBLE_FORCE_COLOR: 'true'
+    - ShellCommand:
+        name: 'Verify that no pv are in released state'
+        command: |-
+          export PATH=${PATH}:/usr/local/bin
+          kubectl get pv
+          ! kubectl get pv -o jsonpath={.items[*].status.phase}| grep 'Released' > /dev/null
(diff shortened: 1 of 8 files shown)


eve/main.yml
```diff
@@ -0,0 +1,91 @@
...
+          KUBECONFIG: inventories/single-node-ci/artifacts/admin.conf
+    - ShellCommand:
+        name: 'Test storage'
+        haltOnFailure: true
+        command: |-
+            /usr/local/bin/kubectl get pv
+            /usr/local/bin/kubectl apply -f tests/storage/test_pv.yml
+            #retry kubectl get pods test-volume -o jsonpath={.status.phase} | grep -v -E '(Failed|Completed)' ]
+            sleep 10
+            /usr/local/bin/kubectl delete pod test-pv
```
NicolasT: We're not really checking somehow whether the test ran inside the `Pod` succeeded?

eve/main.yml
```diff
@@ -0,0 +1,91 @@
...
+            /usr/local/bin/kubectl delete pvc testclaim
+        env:
+          KUBECONFIG: inventories/single-node-ci/artifacts/admin.conf
```
NicolasT: Consider putting `/usr/local/bin` in `PATH` through `env` such that you can simply run `kubectl`, which will reduce changes required when we do the `make shellrc` thing.

eve/main.yml
```diff
@@ -0,0 +1,91 @@
...
+    - ShellCommand:
+        name: 'Verify that some pv are in released state'
+        command: |-
+          /usr/local/bin/kubectl get pv
+          [ "$(/usr/local/bin/kubectl get pv | grep 'Released')" ]
+        env:
+          KUBECONFIG: inventories/single-node-ci/artifacts/admin.conf
+    - ShellCommand:
+        name: 'Reclaime storage'
```
NicolasT: `Reclaim`

eve/single_node/group_vars/k8s-cluster/single_node.yml
```diff
@@ -0,0 +1,24 @@
...
+  vgs:
+    kubevg:
+      drives: ['/dev/loop0']
+
+metal_k8s_storage_class:
+  storage_classes:
+    local-lvm:
+      is_default: true
+  lvm_conf:
+    default_fstype: 'ext4'
```
NicolasT: Must all of these be set when we could rely on defaults?

eve/single_node/single_node.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
+
+dd if=/dev/zero of=/var/lib/kube_lvm bs=1M count=20480
```
NicolasT: `truncate -s 2G /var/lib/kube_lvm` is likely faster and introduces less needless IO (could also use `dd` to create a sparse file, but I always forget what the right arguments for that are :smile:)

eve/workers/zenko.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "zenko-test-pod"
+spec:
+  containers:
+  - name: aggressor
```
NicolasT: Que?

eve/workers/zenko.yaml
```diff
@@ -0,0 +1,16 @@
...
+    image: zenko/zenko-releng:0.0.7
+    resources:
+      requests:
+        cpu: 500m
+        memory: 1Gi
+      limits:
+        cpu: "1"
+        memory: 2Gi
+    command: ["/bin/sh", "-c", "buildbot-worker create-worker . ${BUILDMASTER}:${BUILDMASTER_PORT} ${WORKERNAME} ${WORKERPASS} && buildbot-worker start --nodaemon"]
```
NicolasT: A bit odd to sometimes use `/bin/sh` as `command`, then pass `args`, and sometimes do everything in `command`.

roles/info_displayer/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
         {{ hostvars.localhost.displayable_info|default([])
             + add_info }}
     ansible_connection: '{{ hostvars.localhost.ansible_connection }}'
-    ansible_python_interpreter: '{{ hostvars.localhost.ansible_python_interpreter }}'
+    ansible_python_interpreter: '{{ hostvars.localhost.ansible_python_interpreter|default(omit) }}'
```
NicolasT: Let's make sure to land the fix in its own PR before merging this, to retain references etc.

eve/main.yml
```diff
@@ -0,0 +1,91 @@
+version: "0.2"
+
+branches:
+  default:
+    stage: "pre-merge"
+
+stages:
+  pre-merge:
```
tcarmet: @Zempashi  Is there any reason why you're not directly calling the `single-node` stage?

eve/main.yml
Zempashi: Not yet indeed

eve/main.yml
Zempashi: I was thinking about that but I'm not sure I can reference $PATH in an env variable of buildbot (the goal is to add the /usr/local/bin/ not to set PATH=/usr/local/bin)

eve/single_node/group_vars/k8s-cluster/single_node.yml
Zempashi: but defaults are replaced not merged. We should convert this in a less tree-ish way, more flat. This allow to override just one key.

eve/workers/zenko.yaml
Zempashi: copy-paste, left-over of sylvain PR, but don't what it mean either

eve/main.yml
Zempashi: To let other people adding tests that run in parallel..

tcarmet: yeah, make sense!

eve/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            export PATH=${PATH}:/usr/local/bin
+            make test-simple-storage
+        workdir: build/tests
+        env:
+          KUBECONFIG: inventories/single-node-ci/artifacts/admin.conf
+    - ShellCommand:
+        name: 'Verify that some pv are in released state'
+        command: |-
+          export PATH=${PATH}:/usr/local/bin
+          kubectl get pv
```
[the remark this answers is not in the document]
Zempashi: After discussion avoid awk to parse the result (because grep 'Released' is too wide), result bakcin two commands

PR #76, opened 2018-06-12
Build doc ci

Zempashi:
Build doc in CI (will help @wabernatScality to see if the syntax is ok or not)

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -14,6 +14,7 @@ stages:
         name: trigger all the tests
         stage_names:
         - single-node
+        - build-doc-inter-stage
 
   single-node:
     worker: &openstack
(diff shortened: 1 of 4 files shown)


NicolasT (changes requested):
This includes the `cert-manager` stuff?

eve/main.yml
```diff
@@ -87,3 +88,25 @@ stages:
           kubectl get pv
           ! kubectl get pv -o jsonpath={.items[*].status.phase}| grep 'Released' > /dev/null
         env: *env_kubeconfig
+
+  build-doc-inter-stage:
+    worker:
+      type: kube_pod
+      path: eve/workers/zenko.yaml
```
NicolasT, line 95: `zenko`?

eve/main.yml
```diff
@@ -87,3 +88,25 @@ stages:
...
+    worker:
+      type: kube_pod
+      path: eve/workers/doc-builder.yaml
+      images:
+        doc-builder: docs/
+    steps:
+    - Git: *git_pull
+    - ShellCommand:
+        name: 'Build doc'
+        command: make -C docs html
```
NicolasT, line 112: This is not the same as what the `tox` target runs: as an example, this will not (though should!) bail out when Sphinx generates warnings.
We'll also want to run the `spelling` target (or however it's called).

Zempashi: Might be better to run tox directly. But I don't know if it's possible to download dependencies in dockerfile and then tell tox to use what's in the docker image

NicolasT: I believe the following in the `Dockerfile` should make `tox` generate the environment (i.e. including installation of the dependencies etc) without running the *test command* (which is in this case, building the docs):

```shell
RUN mkdir docs
COPY tox.ini .
COPY docs/requirements.txt docs/
RUN tox -e docs --notest

CMD tox -e docs
```

Zempashi:
supersed by #248

PR #78, opened 2018-06-18
Update third party

NicolasT:
Cfr. #74

Will update with latest remotes first.


Zempashi:
@NicolasT, could you rebase to get an eve/main.yml ?

Zempashi (dismissed):

NicolasT:
@Zempashi Please re-review (couple of minor changes in new commits)

roles/kube_elasticsearch/tasks/main.yml
```diff
@@ -21,6 +22,26 @@
   run_once: true
   delegate_to: "{{ groups['kube-master'][0] }}"
 
+- name: 'Delete old es-client deployment'
```
Zempashi, line 25: At the end (define: end ?). We want to manage migration from point to point version or from any revision to any revision ? Those code for example has little chance to be triggered in production (as we have no production yet). We should discuss this also for a CI point a view.

Zempashi (approved):

PR #80, opened 2018-06-20
Deploy etcd exporter

Zempashi:


roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,94 @@
...
+  run_once: True
+  delegate_to: '{{ groups.etcd|first }}'
+  when: gen_etcd_exporter_csr is changed
+  register: csr_on_etd_node
+
+- debug:
+    var: csr_on_etd_node
+  when: debug|bool
+
+- name: 'create tempfile to get certificates'
```
NicolasT: A bit odd to use a tempfile here, but use a different method in the step above?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,94 @@
...
+  tempfile:
+    state: file
+    suffix: yml
+  run_once: True
+  delegate_to: "{{ groups.etcd|first }}"
+  when: gen_etcd_exporter_csr is changed
+  register: etcd_exporter_cert_temp
```
NicolasT: This is not used? Is the step necessary?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,94 @@
...
+  register: gen_etcd_exporter_cert
+
+- debug:
+    var: gen_etcd_exporter_cert
+  when: debug|bool
+
+- name: 'copy cert on master node'
+  copy:
+    dest: '{{ etcd_exporter_cert }}'
+    content: '{{ gen_etcd_exporter_csr.stdout }}'
```
NicolasT: `gen_etcd_exporter_cert`, no?

Zempashi: oh right

roles/etcd_prometheus_exporter_cert/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+debug: False
+
+etcd_exporter_key: '{{ etcd_cert_dir }}/etcd-exporter-key.pem'
```
NicolasT: `etcd_cert_dir` is set in another role. Is that best practice in Ansible?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,85 @@
+- name: 'make assertions'
+  assert:
+    that:
+     - groups.etcd is defined
+
+- name: 'create private key on master'
+  command: openssl genrsa -out {{ etcd_exporter_key }} 4096
+  args:
+    creates: '{{ etcd_exporter_key }}'
```
NicolasT, line 9: Does this result in this being done only once? We don't want to re-generate the cert on every invocation of the playbook, right?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,85 @@
...
+  when: debug|bool
+
+- name: 'copy csr over a server capable of signing etcd cert'
+  copy:
+    dest: '/tmp/ansible.{{ lookup("password", "/dev/null chars=ascii_letters length=8") }}'
+    content: '{{ gen_etcd_exporter_csr.stdout }}'
+  run_once: True
+  delegate_to: '{{ groups.etcd|first }}'
+  when: gen_etcd_exporter_csr is changed
+  register: csr_on_etd_node
```
NicolasT: `etcd` instead of `etd`?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,85 @@
...
+
+- name: 'generate certificate'
+  command: >-
+    openssl x509 -req -days 3650
+    -in {{ csr_on_etd_node.dest }}
+    -CAkey {{ etcd_cert_dir }}/ca-key.pem
+    -CA {{ etcd_cert_dir }}/ca.pem
+    -CAcreateserial
+  run_once: True
+  delegate_to: '{{ groups.etcd|first }}'
```
NicolasT, line 53: ... and hope it's still the same... Sometimes I'm wondering whether certs shouldn't be all generated on e.g. the host from which Ansible runs, then copy them over wherever they're required. But that goes back to the whole cert handling discussion :)

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,85 @@
...
+  when: gen_etcd_exporter_csr is changed
+  register: gen_etcd_exporter_cert
+
+- debug:
+    var: gen_etcd_exporter_cert
+  when: debug|bool
+
+- name: 'copy cert on master node'
+  copy:
+    dest: '{{ etcd_exporter_cert }}'
```
NicolasT, line 68: I'm confused. This generates `{{ etcd_exporter_cert }}`, and so does `copy cert on master node` below as well?

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,85 @@
...
+- name: 'download cert from signing server '
+  slurp:
+    path: '{{ etcd_exporter_cert }}'
+  run_once: True
+  register: etcd_exporter_cert_content
+
+- name: 'copy cert on master node'
+  copy:
+    dest: '{{ etcd_exporter_cert }}'
+    content: '{{ etcd_exporter_cert_content.content|b64decode }}'
```
NicolasT: Why `b64encode`? I know this is required when putting the cert etc. in a YAML manifest when creating the `Secret`, but why encode in a PEM-file which one doesn't expect to be encoded in any way? This makes it more difficult to use the cert with `curl` or similar to troubleshoot things.

roles/etcd_prometheus_exporter_cert/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+- import_tasks: gen_cert.yml
+
+- name: 'create tempfile for etcd-exporteer auth secret'
```
NicolasT: `exporter`

roles/etcd_prometheus_exporter_cert/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  tempfile:
+    state: file
+    suffix: yml
+  register: kube_etcd_exporter_auth
+  run_once: True
+
+- name: 'template secret'
+  template:
+    src: etcd_exporter_secret.yml.j2
+    dest: '{{ kube_etcd_exporter_auth.path|default("<check_mode_tempfile>") }}'
```
NicolasT, line 13: What does this line do, exactly?

roles/etcd_prometheus_exporter_cert/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  run_once: True
+
+- name: 'Deploy etcd exporter secret'
+  kube:
+    kubectl: '{{ bin_dir }}/kubectl'
+    filename: '{{ kube_etcd_exporter_auth.path|default }}'
```
NicolasT, line 19: This is not exactly the same as how the path is constructed above, which seems fishy?

roles/etcd_prometheus_exporter_cert/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    namespace: '{{ kube_prometheus_namespace }}'
+    state: 'latest'
+  run_once: True
+
+- name: 'remove uneeded tempfile'
```
NicolasT: `unneeded`

roles/etcd_prometheus_exporter_cert/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  file:
+    path: '{{ kube_etcd_exporter_auth.path|default }}'
+    state: absent
+  run_once: True
+
+- name: 'export sosme vars for later roles'
```
NicolasT: `some`

roles/etcd_prometheus_exporter_cert/templates/etcd_exporter_secret.yml.j2
```diff
@@ -0,0 +1,14 @@
...
+  name: '{{ kube_prometheus_namespace }}'
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kube-prometheus-exporter-kube-etcd
+  namespace: '{{ kube_prometheus_namespace }}'
+type: kubernetes.io/tls
+data:
+  tls.crt: '{{ etcd_exporter_cert_content.content }}'
```
NicolasT: These should be base64 encoded, right?

roles/etcd_prometheus_exporter_cert/templates/etcd_exporter_secret.yml.j2
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: '{{ kube_prometheus_namespace }}'
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kube-prometheus-exporter-kube-etcd
+  namespace: '{{ kube_prometheus_namespace }}'
```
NicolasT: No need to put a `namespace`, we pass it to `kubectl` above.

NicolasT (changes requested):
Some spelling cleanups, some questions.

roles/etcd_prometheus_exporter_cert/defaults/main.yml
Zempashi: I forget to move it from kube_prometheus (which is run is the same play). The real info is in kubespray role but this one install etcd (which I don't want do to here). So I duplicate the info

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
Zempashi: Yes, the `creates` tell that the command should create the file, and should not be run if the file is already here.

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
Zempashi: I will remove this debug and the register (thinking twice, it's not interesting anymore)

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
Zempashi: Yes groups are ordered server (so it's safe). For generating everything locally, why not (but don't see many playbooks that do it this way)

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
Zempashi: Indeed ...(what's wrong with me ?)

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
Zempashi: https://docs.ansible.com/ansible/2.4/slurp_module.html store result in base64 in ansible var

roles/etcd_prometheus_exporter_cert/tasks/main.yml
Zempashi: in checkmode ansible will not execute tempfile, so the path argument will not exist and then result into an exception for the template task. I add a default so that template can execute normaly.

roles/etcd_prometheus_exporter_cert/tasks/main.yml
Zempashi: but filename is not a mandatory argument for kube module so None is ok.

roles/etcd_prometheus_exporter_cert/templates/etcd_exporter_secret.yml.j2
Zempashi: It is already already (see comment above about `slurp`)

Zempashi:
rah I'm not used to "commiting the comment transaction" of github. Sorry for  the delay

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,78 @@
...
+  when: >-
+    gen_etcd_exporter_key is changed or not etcd_exporter_cert_file.stat.exists
+
+- debug:
+    var: gen_etcd_exporter_csr
+  when: debug|bool
+
+- name: 'copy csr over a server capable of signing etcd cert'
+  copy:
+    dest: '/tmp/ansible.{{ lookup("password", "/dev/null chars=ascii_letters length=8") }}'
```
NicolasT, line 34: This file seems not to be deleted.

Zempashi: done

NicolasT:
You'll need to merge `development/0.1` in your branch because of the CI changes.

roles/etcd_prometheus_exporter_cert/tasks/gen_cert.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: 'generate csr'
+  command: >-
+    openssl req -new
+    -subj '/CN=etcd-exporter'
+    -key {{ etcd_exporter_key }}
+  run_once: True
+  register: gen_etcd_exporter_csr
+  when: >-
+    gen_etcd_exporter_key is changed or not etcd_exporter_cert_file.stat.exists
```
NicolasT, line 26: This could, at some point in time, become an issue when the cert expires, I guess :)

NicolasT (dismissed):

NicolasT (approved):

PR #82, opened 2018-06-21
Etcd grafana dashboard

Zempashi:
Add grafana dashboard with new vendoring system embryo


NicolasT:
Will review once #80 is in :)

NicolasT (approved):

PR #90, opened 2018-06-26
Rework tests layout

NicolasT:
These changes
- Import [bash_unit](https://github.com/pgrange/bash_unit/) as a dependency
- Rewrite the existing `single-node` test and its `reclaim-storage` scenario using the `bash_unit` 'framework'

The rationale is to move test implementation out of Eve configuration as much as possible: let Eve simply run (a suite of) tests, ideally with one command invocation, in the various jobs.

Also fixing an oversight in the `make shell C=...` implementation, which I didn't catch earlier because I was testing whilst already inside a `make shell` environment :blush:


Zempashi (approved):

tests/single-node/test.sh
```diff
@@ -0,0 +1,105 @@
...
+        make_shell kubectl get nodes
+}
+
+test_reclaim_storage() {
+        echo "Listing all PVs before test (some should be available)"
+        make_shell kubectl get pv
+
+        echo "Running simple Pod with PersistentVolumeClaim"
+        make_shell kubectl apply -f - << EOF
+kind: PersistentVolumeClaim
```
Zempashi, line 44: Those files has meaning outside of the test, could be nice to have them as separated resources. This break a little the "self-contain" test, but allow us to launch the pod on any cluster without reclaiming the entire storage of the cluster

NicolasT: I'm not sure why you'd want to run this specific Pod & PVC on 'any cluster'? In that case, copy-pasting into some file as a one-off seems reasonable :)

PR #99, opened 2018-07-03
Fix redeployment

Zempashi:


tests/single-node/test.sh
```diff
@@ -35,6 +35,10 @@ setup_suite() {
         make_shell kubectl get nodes
 }
 
+test_deploy_again() {
+        make_shell ansible-playbook -i "$(pwd)/inventory" metal-k8s.yml --skip elasticsearch || die
```
NicolasT: Should this be `die` or use an `assert`?

NicolasT (changes requested):

tests/single-node/test.sh
Zempashi: Is not the same effect ?

NicolasT: No. `die` will stop execution of tests altogether (it's `exit 1`). One of the *bash_unit* `assert`s will make the one test-case fail, but continue running others.

NicolasT (approved):

PR #100, opened 2018-07-03
Fix #98: allow elasticsearch to be disabled via configuration flag

### LINKED ISSUE #98: Make `kube_elasticsearch` deployment optional
Add a var, `metal_k8s_enable_elasticsearch`, which makes `kube_elasticsearch` optional, enabled by default. When disabled but enabled in a later run of the playbook, the services should be deployed. Also, delete the services when disabled again later (`helm delete`, `kubectl delete`,...)

Zempashi:
Installation with flag to metalk8s_elasticsearch_enabled=False will
skip elasticsearch installation.
Setting to flag to true and rerun the playbook will install es.
Disabling the flag and rerun the playbook will not remove elasticsearch.
Should be done manually.

### DIFF
--- a/metal-k8s.yml
+++ b/metal-k8s.yml
@@ -52,6 +52,7 @@
       tags: ['heapster']
     - role: kube_elasticsearch
       tags: ['elasticsearch']
+      when: metalk8s_elasticsearch_enabled|bool
     - role: kube_metrics_server
       tags: ['metrics-server']
 
--- a/roles/kube_elasticsearch/defaults/main.yml
+++ b/roles/kube_elasticsearch/defaults/main.yml
@@ -1,3 +1,6 @@
+metalk8s_elasticsearch_enabled: True
+
+
 debug: false
 
 remove_metal_k8s_temporary_file: true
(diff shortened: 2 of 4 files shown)


metal-k8s.yml
```diff
@@ -52,6 +52,7 @@
       tags: ['heapster']
     - role: kube_elasticsearch
       tags: ['elasticsearch']
+      when: metalk8s_elasticsearch_enabled|default(True)|bool
```
NicolasT: This introduces a 'magic' var. Shouldn't this be set in some role's `defaults` with the appropriate default value?

Zempashi: Yes indeed, variables behave differently in playbook, but now we have the default that stacks before the play to start. Two options:

* just moving default to kubernetes_elasticsearch would work
* having everything in the role (having main.yml that only include (conditionally) the real stuff

EDIT: we could also anyway add a commented variable in group_vars (as a documentation)

NicolasT (dismissed):
We need to work out a story around documentation of vars, role independenc,... in general anyway.

NicolasT (approved):

PR #102, opened 2018-07-03
etcd_prometheus_exporter_cert: Temporary fix uninitialized variable

NicolasT:
See: 9c958c40a73980461cdc7c0b6215de947ca18478
See: https://github.com/scality/metal-k8s/commit/9c958c40a73980461cdc7c0b6215de947ca18478


NicolasT:
Note: this is a stop-gap 'fix', needs a better permanent solution.

NicolasT:
Superseded by #99

PR #105, opened 2018-07-03
Move the `kubespray` subtree under `vendor`

NicolasT:


NicolasT:
Even this looks big, it's merely moving a directory and updating references.

The intent is to simplify the root of the repository and unify vendored packages.

Zempashi (approved):

PR #106, opened 2018-07-03
Move all Ansible playbooks under `playbooks/`

NicolasT:
Backwards incompatible change.


NicolasT:
I'm not sure the way the playbooks have now been split and imported is the right way forward. @Zempashi Please weigh in.

playbooks/roles
```diff
@@ -0,0 +1 @@
+../roles
```
Zempashi, line 1: Why not moving roles as well ?

NicolasT: Because, at least in theory, those roles could be useful by themself as well, and used by other projects, or tested in isolation.

I based this partially on the structure used in https://github.com/openshift/openshift-ansible

BTW, I'd reallly like to get rid of everything we do with `group_vars` (having a `group_vars` directory in the playbook, the symlink in `kubespray`,...)

Zempashi (dismissed):

NicolasT:
@Zempashi I had to merge in `development/0.1` due to other changes. Please re-approve after CI validates?

Also, we'll need to announce the new playbook location(s) widely. Documentation has been updated in this PR.

Zempashi (approved):

Issue #73, opened 2018-06-06
Readme file lacks of some needed dependency


Zempashi:
namely: python-dev and gcc

PR #109, opened 2018-07-03
README: Mention `python-dev` and `gcc` requirements

NicolasT:
Fixes: #73
See: https://github.com/scality/metal-k8s/issues/73


Zempashi (approved):

wabernat (commented):
Insert a space between numerals and abbreviated units of measure: 20 GB (Chicago Manual 16, 10.52)

Issue #56, opened 2018-05-18
Monitor `etcd`-only nodes


NicolasT:
Due to `node_exporter` being deployed as a `DaemonSet`, any node which is not part of the Kubernetes cluster isn't monitored. This can be the case for `etcd` nodes when they're not colocated with `kube-node` or `kube-master` roles.

We should
- Deploy `node_exporter` on them as well (cfr. #31)
- Ensure the Prometheus deployed with `kube-prometheus` is set up to also scrape metrics from these servers

Issue #31, opened 2018-03-26
Monitor all system volumes with node_exporter


NicolasT:
Currently `node_exporter` doesn't monitor all volumes available on the system: as https://github.com/prometheus/node_exporter/blob/1f11a86d594173ca1146ac1d1715cd6263e9959d/README.md#using-docker mentions, we'd need to bind all volumes in a `node_exporter` VM, which is impractical, most certainly because we're not setting up the `node_exporter` `DaemonSet` ourself...

It may be possible to work around this using `MountPropagation` and some other settings (but then see https://github.com/prometheus/node_exporter/pull/672 and https://github.com/prometheus/node_exporter/pull/660), which is far from ideal.

Maybe the best way forward would be to simply deploy `node_exporter` on all nodes, as it's intended to be deployed, and not let `kube-prometheus` (or `prometheus-operator`) manage it, just make sure the metrics are collected as intended.

ballot-scality:
We might rely on https://copr.fedorainfracloud.org/coprs/ibotty/prometheus-exporters/
As stated in the node exporter README it is a 3rdparty maintained closely with upstream.

NicolasT:
Cool. Would you want to work on this? We may need to figure out how to let `kube-prometheus` then use the host-installed exporter and such...

NicolasT:
https://github.com/coreos/prometheus-operator/issues/569#issuecomment-380727999

PR #113, opened 2018-07-04
Improvement/use node exporter pkg

Zempashi:
Fixes: #31 
Fixes: #56

### DIFF
--- a/playbooks/services.yml
+++ b/playbooks/services.yml
@@ -7,6 +7,8 @@
       tags: ['kube-prometheus']
     - role: kube_prometheus
       tags: ['kube-prometheus']
+    - role: kube_prometheus_node_exporter
+      tags: ['kube-prometheus', 'node-exporter']
     - role: kube_heapster
       tags: ['heapster']
     - role: kube_elasticsearch
@@ -14,3 +16,8 @@
       when: metalk8s_elasticsearch_enabled|bool
     - role: kube_metrics_server
       tags: ['metrics-server']
+
+- hosts: k8s-cluster:etcd
+  roles:
+    - role: node_exporter
+      tags: ['kube-prometheus', 'node-exporter', 'node-exporter-pkg']
--- a/roles/kube_prometheus_node_exporter/tasks/main.yml
+++ b/roles/kube_prometheus_node_exporter/tasks/main.yml
@@ -1,0 +1,33 @@
+- name: 'make assertions'
+  assert:
+    that:
+     - prometheus_addon_dir is defined
+     - kube_prometheus_namespace is defined
+
+- name: 'create addons dir'
+  file:
+    dest: '{{ prometheus_addon_dir }}'
+    state: 'directory'
+
+- name: 'Add node_exporter to prometheus monitoring'
+  copy:
+    src: 'node_exporter.yaml'
+    dest: '{{ prometheus_addon_dir }}/node_exporter.yaml'
+  run_once: True
+
+- name: 'Add node_exporter endpoints'
+  template:
+    src: endpoints.yaml.j2
+    dest: '{{ prometheus_addon_dir }}/endpoints.yaml'
+  run_once: True
+
+- name: 'Deploy node-exporter monitoring'
+  kube:
+    kubectl: '{{ bin_dir }}/kubectl'
+    filename: '{{ prometheus_addon_dir }}/{{ item }}'
+    namespace: '{{ kube_prometheus_namespace }}'
+    state: 'latest'
+  run_once: true
+  with_items:
+    - node_exporter.yaml
+    - endpoints.yaml
--- a/roles/node_exporter/tasks/main.yml
+++ b/roles/node_exporter/tasks/main.yml
@@ -1,0 +1,32 @@
+- name: 'set node exporter repository'
+  yum_repository:
+    name: prometheus-rpm_release
+    baseurl: https://packagecloud.io/prometheus-rpm/release/el/$releasever/$basearch
+    description: prometheus-rpm_release
+    enabled: True
+    gpgcheck: False
+    gpgkey: https://packagecloud.io/prometheus-rpm/release/gpgkey
+    sslverify: True
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    repo_gpgcheck: True
+    state: present
+
+# https://github.com/ansible/ansible/issues/20711
+# For ansible < 2.6
+- name: 'rebuild cache'
+  command: yum -y makecache
+  args:
+    creates: /var/lib/yum/repos/{{ ansible_machine }}/{{  ansible_distribution_major_version }}/prometheus-rpm_release/gpgdir/gpg.conf
+    warn: False
+
+- name: 'install Prometheus Node exporter'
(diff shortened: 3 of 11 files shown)


roles/node_exporter/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    description: prometheus-rpm_release
+    enabled: True
+    gpgcheck: False
+    gpgkey: https://packagecloud.io/prometheus-rpm/release/gpgkey
+    sslverify: True
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    repo_gpgcheck: True
+    state: present
+
+# https://github.com/ansible/ansible/issues/20711
```
Zempashi, line 14: @ballot-scality I'm not sure how to detect that the key has been accepted or not.

tests/single-node/test.sh
```diff
@@ -103,3 +105,17 @@ EOF
         assert_fails 'make_shell kubectl get pv -o jsonpath={.items[*].status.phase} | grep Released > /dev/null' \
                 "PVs in Released state found"
 }
+
+test_prometheus_node_exporter_metrics() {
```
Zempashi, line 113: This test rely on being on the server itself (yum and accessing prometheus cluster IP directly) but test are better than no test. As already discussed, we might want to move to a more feature-full framework (list of features to be defined)

metal-k8s.yml
```diff
@@ -48,13 +49,20 @@
...
+    - role: kube_prometheus_node_exporter
+      tags: ['kube-prometheus', 'node-exporter']
     - role: kube_heapster
       tags: ['heapster']
     - role: kube_elasticsearch
       tags: ['elasticsearch']
     - role: kube_metrics_server
       tags: ['metrics-server']
 
+- hosts: kube-node
```
NicolasT: Shouldn't this go on `kube-node`, `kube-master` (so actually `k8s-cluster`) *and* `etcd`?

roles/kube_prometheus_node_exporter/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+- name: 'create addons dir'
+  file:
+    dest: '{{ prometheus_addon_dir }}'
+    state: 'directory'
+
+- name: 'make assertions'
+  assert:
+    that:
+     - prometheus_addon_dir is defined
```
NicolasT: Assertion comes after first usage  (above), which is odd?

playbooks/services.yml
```diff
@@ -7,10 +7,17 @@
...
+      tags: ['kube-prometheus', 'node-exporter']
     - role: kube_heapster
       tags: ['heapster']
     - role: kube_elasticsearch
       tags: ['elasticsearch']
       when: metalk8s_elasticsearch_enabled|bool
     - role: kube_metrics_server
       tags: ['metrics-server']
+
+- hosts: kube-node
```
NicolasT: Shouldn't this be `k8s-cluster:etcd`?

roles/node_exporter/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+- name: 'set node exporter repository'
+  yum_repository:
+    name: prometheus-rpm_release
+    baseurl: https://packagecloud.io/prometheus-rpm/release/el/$releasever/$basearch
+    description: prometheus-rpm_release
+    enabled: True
+    gpgcheck: False
```
NicolasT, line 7: Oh?

roles/node_exporter/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+- name: 'set node exporter repository'
+  yum_repository:
+    name: prometheus-rpm_release
+    baseurl: https://packagecloud.io/prometheus-rpm/release/el/$releasever/$basearch
+    description: prometheus-rpm_release
+    enabled: True
```
NicolasT, line 6: Maybe keep disabled by default and explicitly enable when installing `node_exporter`?

roles/node_exporter/tasks/main.yml
Zempashi: but repo_gpgcheck: True (which is an oddity specific to packagecloud.io)

NicolasT (dismissed):
Didn't test, but LGTM on-sight...

Looking into making the packagecloud.io repo not enabled  by default could be nice, but could be a follow-up PR as far as I'm concerned.

Also, a bit surprised you need the `makecache`, I thought `yum` does that automatically when it finds new repositories.

roles/node_exporter/tasks/main.yml
Zempashi: For update sake ? Talking to @ballot-scality, there is some plugin to install yum-version-lock. Might be better

NicolasT: Update would all of a sudden install 0.16 :) I'm not too fond of version-locking myself. I believe Kubespray even uses a separate yum.conf for Docker installs to keep things separated. You can enable repos on `yum` invocation (so likely also through the `yum` module)

NicolasT:
CI seems unhappy :robot:

roles/node_exporter/tasks/main.yml
Zempashi: repo docker are enabled (and yes install docker-ce will install 18.03, which is bad).
We must avoid doing the same mistake of kubespray. Version locking could fix docker problem (because disabling docker repo after kubespray will not prevent kubespray itself to reenable it).
It will also allow some upgrade and prevent what we have on the ring multiple times: "oh the package is now on epel so the system decide to take the version of epel instead of our repo."

NicolasT: Fair enough :smiley:

tests/single-node/test.sh
```diff
@@ -107,3 +109,17 @@ EOF
...
+        sudo yum install -y jq
+        PROMETHEUS_IP=$(make_shell kubectl \
+                        --namespace ${PROMETHEUS_NAMESPACE} \
+                        get service kube-prometheus -o jsonpath={.spec.clusterIP})
+        echo "Prometheus IP: ${PROMETHEUS_IP}"
+        NB_TARGET=$(curl -s http://${PROMETHEUS_IP}:9090/api/v1/targets| \
+          jq '.data.activeTargets |[.[]|select(.labels.job == $job)]|length' \
+          --arg job "node-exporter")
+        echo "Found ${NB_TARGET} targets"
+        assert [ "${NB_TARGET}" -gt "0" ]
```
NicolasT: For future:
```bash
assert_not_equals 0 ${NB_TARGET}
```

NicolasT:
Fixed CI by applying https://github.com/scality/metal-k8s/pull/113#discussion_r200175576

NicolasT (dismissed):

NicolasT (dismissed):

NicolasT (approved):
I assume this has been tested, and existing Grafana dashboards still work :smiley:

Zempashi:
I've checked that when reverting from 0.16 to 0.15 on one node, some dashboards where back. I'll double check

PR #122, opened 2018-07-05
Fix etcd dashboard

### LINKED ISSUE #107: etcd dashboard is broken
The etcd Grafana dashboard imported in #82 is broken in the JSON file (cfr https://github.com/scality/metal-k8s/blob/a212257129b7494cb2e8e776dbe4afec7772216e/roles/kube_prometheus/files/additionnal_dashboard.yml#L16)

This likely requires fixes in the `Makefile`?

Zempashi:
Fix #107


roles/kube_prometheus/hack/aggregate_dashboard.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def confdir(self):
+        return os.path.dirname(self.configfile)
+
+    def iter_sources(self):
+        for section in self.config.sections():
+            if section.startswith('source:'):
+                yield section
+
+    def iter_filenames_in_source(self, source_name):
```
Zempashi: Some nasty bug here if you have a dir call `src/\*.json` and source that target `src/*.json`. Also look at current directory not relative to the configuration file

roles/kube_prometheus/aggregate_dashboard.ini
```diff
@@ -0,0 +1,8 @@
+[default]
```
NicolasT, line 1: Instead of requiring a config file, wouldn't it be easier to expect files on the command line and print the result to stdout, so one can run
```
aggregate-dashboard.py ../../vendor/etcd/Documentation/op-guide/grafana.json src/*.json > files/additional_dashboard.yml
```

Zempashi: so the configuration lays in the command line. We need a makefile target, or something like this or I'm afraid developers will add their own new dashboard but forget to keep the oldest ones. My belief is that command line is more error prone.

NicolasT: In a Makefile you could use `src/*.json` etc without needing any special support in the tool ;-) Anyway, I'll approve, feel free to merge as-is if you feel like it.

NicolasT (approved):

PR #130, opened 2018-07-09
Implement an example role unit-test

NicolasT:
This PR imports an example unit-test for a role using [Molecule](https://github.com/metacloud/molecule). It also changes some GPG key handling for the Prometheus RPM repository we're using, and relying on the imported tests to validate the changes.

Milestone 0.1.1 because of the role changes.


NicolasT:
TODO: Run in Eve.

roles/node_exporter/molecule/default/Dockerfile.j2
```diff
@@ -0,0 +1,14 @@
+# Molecule managed
+
+{% if item.registry is defined %}
```
Zempashi, line 3: Is it needed to add support for other distribution. (Yes we could, should we ?)

NicolasT: This file is auto-generated by Molecule :smiley:

Indeed, this would allow to test against other platforms, by adding them in `molecule.yml`. However, this role currently doesn't support any platform but RedHat, and MetalK8s doesn't either, so not sure we should spend time on this...

Zempashi: Could we restrict the platform supported in molecule.yml ? Hopefully, the generated code smaller, and not misleading.

Zempashi: hmmm seems not

Zempashi (approved):

roles/node_exporter/molecule/default/Dockerfile.j2
NicolasT: There is already only one platform defined: https://github.com/scality/metal-k8s/pull/130/files#diff-20f42e7f8e26d9ec2ddfd316087ab50bR9

This `Dockerfile.j2` is managed by Molecule. Nobody is supposed to 'understand' it in a way :wink:

Issue #120, opened 2018-07-05
The `services.yml` playbook fails due to some `Undefined` variable which can't be JSON-encoded


NicolasT:
```
TASK [kube_prometheus : copy kube-prometheus values into temporary file] *********************************************************************************************************************************************************************
fatal: [metalk8s-master-01 -> 10.200.4.36]: FAILED! => {
    "changed": false,
    "msg": "AnsibleError: Unexpected templating type error occurred on (deployExporterNode: False\ngrafana:\n  extraVars:\n    - name: 'GF_SERVER_ROOT_URL'\n      value: '%(protocol)s://%(domain)s/api/v1/namespaces/kube-ops/services/kube-prometheus-grafana:http/proxy/'\n  service:\n    labels:\n      kubernetes.io/cluster-service: \"true\"\n      kubernetes.io/name: \"Grafana\"\n\nprometheus:\n  externalUrl: '/api/v1/namespaces/kube-ops/services/kube-prometheus:http/proxy/'\n  service:\n    labels:\n      kubernetes.io/cluster-service: \"true\"\n      kubernetes.io/name: \"Prometheus\"\n  replicaCount: 2\n{% if kube_prometheus_secret|default %}\n  secrets:\n{% for secret in kube_prometheus_secret %}\n  - {{ secret }}\n{% endfor %}\n{% endif %}\n  storageSpec:\n    volumeClaimTemplate:\n      spec:\n        accessModes: [\"ReadWriteOnce\"]\n        resources:\n          requests:\n            storage: {{ prometheus_storage_size }}\n\nalertmanager:\n  externalUrl: '/api/v1/namespaces/kube-ops/services/kube-prometheus-alertmanager:http/proxy/'\n  service:\n    labels:\n      kubernetes.io/cluster-service: \"true\"\n      kubernetes.io/name: \"Alertmanager\"\n  replicaCount: 2\n\nexporter-kube-etcd:\n  etcdPort: 2379\n  endpoints: {{ groups.etcd | map('extract', hostvars, ['ansible_default_ipv4', 'address'])|list|to_json }}\n  scheme: https\n  # Linked to the secret of prometheus\n{% if exporter_kube_etcd_certFile|default %}\n  certFile: {{ exporter_kube_etcd_certFile }}\n{% endif %}\n{% if exporter_kube_etcd_keyFile|default %}\n  keyFile: {{ exporter_kube_etcd_keyFile }}\n{% endif %}\n): Undefined is not JSON serializable"
}

[...]

PR #132, opened 2018-07-09
playbooks: Fix `services` playbook

NicolasT:
Because not all node facts were gathered when only running the
`services` playbook, the `kube_prometheus` role failed since it couldn't
find the IP addresses of all `etcd` and `k8s-cluster` nodes.

This is somewhat of a band-aid which ensures facts *are* gathered. Maybe
adding this to the code which 'requires' these values (i.e. somewhere in
the `kube_prometheus` role) would be better...

Fixes: #120
See: https://github.com/scality/metal-k8s/issues/120

### DIFF
--- a/playbooks/services.yml
+++ b/playbooks/services.yml
@@ -1,3 +1,9 @@
+# https://github.com/scality/metal-k8s/issues/120
+- hosts: k8s-cluster:etcd
+  tasks:
+    - setup:
+  gather_facts: false
+
 - hosts: kube-master
   gather_facts: False
   roles:
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -8,6 +8,9 @@ Bugs fixed
 :ghissue:`103` - set up host anti-affinity for Elasticsearch service scheduling
 (:ghpull:`113`)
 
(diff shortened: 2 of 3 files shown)


playbooks/services.yml
```diff
@@ -1,3 +1,8 @@
+# https://github.com/scality/metal-k8s/issues/120
+- hosts: k8s-cluster:etcd
+  tasks:
+    - debug: msg="Ensure facts are gathered"
```
Zempashi: the tasks: "setup"  (with gather_fatcs: False)

Zempashi (dismissed):

playbooks/services.yml
NicolasT: What do you mean exactly :question:

Zempashi: oups forget the word "or". For gathering facts ansible just execute the special tasks "setup" that you can use instead of debug task (but don't forget to set gather_facts: False, unless you want ansible to gather facts twice). Just to let you know. It's perfectly understandable in your version, which is all I care.

Zempashi (approved):

Issue #150, opened 2018-07-16
Various Grafana dashboards are gone


NicolasT:
The Elasticsearch, Prometheus and *node-exporter-full* dashboards seem to be no longer deployed. This may (though not confirmed) be related to #122?

NicolasT:
Any fix for this should introduce a test-case.

NicolasT:
Fixed in #158

PR #158, opened 2018-07-18
Fix: #150 yaml formating of grafana dashboards

Zempashi:
Fix: #150

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -14,6 +14,8 @@ Bugs fixed
 ----------
 :ghpull:`151` - fix `debug` clause `var` scoping
 
+:ghissue:`150` - fix deployment of Elasticsearch, node and Prometheus Grafana dashboards (:ghpull:`158`)
+
 Release 0.1.1
 =============
 .. note:: Compatibility with future releases of MetalK8s is not guaranteed until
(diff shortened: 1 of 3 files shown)


ChangeLog.rst
```diff
@@ -14,6 +14,8 @@ Bugs fixed
 ----------
 :ghpull:`151` - fix `debug` clause `var` scoping
 
+:ghissue:`150` - Fix yaml formating of grafana dashboards
```
NicolasT: That's the fix, but not the bug fixed :smiley:

```
:ghissue:`150` - fix deployment of Elasticsearch, node and Prometheus Grafana dashboards (:ghpull:`158`)
```

NicolasT (changes requested):
Minor adjustment to ChangeLog entry please, otherwise LGTM.

ChangeLog.rst
Zempashi: what ? that's precisely the bug id ?

NicolasT: No, I meant, the text says what the fix is, not what the bug was. I acknowledge the description of #151 above is similar, but that's a bit of an odd duck.

A user wants to know which bug (i.e. something he could observe) is fixed. `YAML formatting of Grafana dashboards` is not a bug.

NicolasT (approved):
:100:

PR #159, opened 2018-07-18
Elasticsearch: Use upstream (incubator) chart

NicolasT:
BREAKING CHANGE: Historical log data is lost

This commit moves our Elasticsearch deployment from using the vendored manifests from `@pires/kubernetes-elasticsearch-cluster` to deploying the cluster through Helm, using the `elasticsearch` chart from the `@kubernetes/charts` 'incubator' repository. We now also deploy
Elasticsearch Curator using the `elasticsearch-curator` chart.

The configuration of this deployment is such that resource requests and limits are similar to the settings we used before. The Curator configuration is the same as before, i.e. deleting old log indices
after 3 days.

However, unlike the previous manifests which run *master* nodes as a *Deployment*, now we use a *StatefulSet*, so this impacts availability of *PersistentVolume*s. Each *master* node (of which 3 replicas are deployed by default) requests a 4GB volume, which fit with the 5GB volumes that are provisioned by default.

The playbook will remove objects created using the old manifests. However, this functionality is likely to be removed in MetalK8s 1.0.

Since we also deploy new *data* services, the *PersistentVolume* storage volumes used by the old *StatefulSet* are not reused. This results in _loss of historical data_. Furthermore, the *PersistentVolumeClaim*s created by the old *StatefulSet* are kept. You'll want to delete these
using the following command:

```shell
$ kubectl --namespace=kube-ops delete pvc \
    storage-es-data-0 storage-es-data-1 storage-es-data-2
```

Note: this may need to be adjusted if you ran more data replicas.

At this point, you will likely want to run the `reclaim-storage` playbook to make these storage volumes available again for other services.

Fixes: #147
See: https://github.com/scality/metal-k8s/issues/147
See: https://github.com/kubernetes/charts/tree/master/incubator/elasticsearch
See: https://github.com/kubernetes/charts/tree/master/incubator/elasticsearch-curator
See: https://github.com/pires/kubernetes-elasticsearch-cluster/


NicolasT:
If https://github.com/kubernetes/charts/pull/6718 could go in soon, going for that version would be useful.

Zempashi (approved):

PR #165, opened 2018-07-18
Forward-merge development/0.1

NicolasT:


NicolasT:
Hmpf, doesn't work this way.

PR #183, opened 2018-07-23
Metalk8s storage split

Zempashi:


NicolasT:
CI fails :)

roles/metalk8s_lvm_vg/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+- name: "LVM Setup: Detect old configuration style < 0.2.0"
```
ballot-scality: This assertion doesn't belong to this place anymore

ballot-scality (dismissed):
I'm OK with the changes and I need this PR to rework the tests and the check of the devices presence to avoid conflict

ballot-scality (approved):
Rebased on development/1.1

Issue #170, opened 2018-07-19
Enhance Grafana dashboard titles


NicolasT:
`ElasticSearch Example` :arrow_forward: `Elasticsearch`
`Node Exporter Full` :arrow_forward: `Node (Detailed)`

NicolasT:
Before this, I'd like to have #115 in.

PR #188, opened 2018-07-24
Grafana: Rename dashboards

NicolasT:
Fixes: #170
See: https://github.com/scality/metal-k8s/issues/170

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -2,6 +2,10 @@
 
 Release 0.2.1 (in development)
 ==============================
+Bugs fixed
+----------
+:ghissue:`170` - rename `ElasticSearch Example` and `Node Exporter Full` Grafana
+dashboards (:ghpull:`188`)
 
 Release 0.2.0
 =============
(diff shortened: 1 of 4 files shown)


vendor/justwatchcom/elasticsearch_exporter/examples/grafana/dashboard.json
```diff
@@ -1167,6 +1167,6 @@
     ]
   },
   "timezone": "utc",
-  "title": "ElasticSearch Example",
+  "title": "Elasticsearch",
```
Zempashi: Could we use configuration of the aggregation to rewrite the name on the fly, instead of editing vendored stuff directly ?

NicolasT: Good point, I didn't consider that... I'll make those changes.

NicolasT: I now implemented this, thanks for the suggestion!

Zempashi (approved):

Issue #143, opened 2018-07-13
Capture Nginx (ingress) metrics


NicolasT:
`nginx-ingress` can expose Nginx metrics in the Prometheus format, but we're currently not configuring `kube-prometheus` to capture these, nor including a suitable dashboard.

PR #189, opened 2018-07-24
Add Prometheus ServiceMonitor to monitor Nginx

Zempashi:
(related to #143, but only part of it)


playbooks/services.yml
```diff
@@ -9,12 +9,12 @@
   roles:
     - role: helm_install
       tags: ['helm']
-    - role: kube_nginx_ingress
-      tags: ['ingress']
     - role: etcd_prometheus_exporter_cert
       tags: ['kube-prometheus']
     - role: kube_prometheus
       tags: ['kube-prometheus']
+    - role: kube_nginx_ingress
```
NicolasT, line 16: If I'm not mistaken, `prometheus-operator` uses a `Job` to register the CRD. Since we can't really know when this `Job` finished, couldn't this result in a timing problem?

Zempashi: Oh wasn't aware of that. So now even the doc of `prometheus-operator` is wrong stating:
```
helm install coreos/prometheus-operator --name prometheus-operator --namespace monitoring
helm install coreos/kube-prometheus --name kube-prometheus --namespace monitoring
```
There is no guarantee that the second line could execute correctly...

Zempashi: And might explain some transient errors when installing kube-prometheus

Zempashi: Suggestion: add a role "wait_prometheus_crd", which does (with retry on failure):

```
kubectl get customresourcedefinitions alertmanagers.monitoring.coreos.com prometheuses.monitoring.coreos.com servicemonitors.monitoring.coreos.com
```
(Taken from the helm test get-crd-job.yml of kube-prometheus)

playbooks/services.yml
```diff
@@ -7,14 +7,16 @@
 - hosts: kube-master
   gather_facts: False
   roles:
-    - role: helm_install
-      tags: ['helm']
     - role: kube_nginx_ingress
       tags: ['ingress']
+    - role: helm_install
```
NicolasT: How does `kube_nginx_ingress` work when `helm_install` isn't done?

NicolasT: CI fails because of this.

roles/kube_nginx_ingress/files/nginx_ingress_values.yml
```diff
@@ -1,8 +1,14 @@
 controller:
+  image:
+    tag: "0.17.1"
```
NicolasT, line 3: We'll need to monitor closely when this changes upstream (and make sure we don't keep overwriting when upstream is at, say, 0.18.6).

roles/kube_nginx_ingress/tasks/prometheus_nginx.yml
```diff
@@ -0,0 +1,22 @@
...
+- name: 'template nginx monitoring manifests'
+  template:
+    src: '{{ item }}.j2'
+    dest: '{{ nginx_ingress_addons_dir }}/{{ item }}'
+  register: nginx_ingress_manifests
+  with_items:
+    - nginx_ingress_servicemonitor.yaml
+
+- name: 'Nginx ingress monitoring: apply manifests'
+  kube:
```
NicolasT: `namespace`?

roles/kube_nginx_ingress_monitoring/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+- name: 'create addons dir if needed'
```
NicolasT: I'm confused now. How does this relate to `roles/kube_nginx_ingress/tasks/prometheus_nginx.yml`?

roles/kube_nginx_ingress_monitoring/templates/nginx_ingress_servicemonitor.yaml.j2
```diff
@@ -0,0 +1,20 @@
...
+  endpoints:
+  - interval: 15s
+    port: 'metrics'
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+    - {{ nginx_ingress_namespace }}
+  selector:
+    matchLabels:
+      app: nginx-ingress
```
NicolasT: Should we also filter on `release`, in case someone decides to deploy a second `nginx-ingress` in the same namespace?

NicolasT (changes requested):
A bit confused by the duplication?

roles/kube_nginx_ingress_monitoring/tasks/main.yml
Zempashi: It's a move but it did `mv` instead of `git mv`

playbooks/services.yml
```diff
@@ -15,6 +15,8 @@
       tags: ['kube-prometheus']
     - role: kube_prometheus
       tags: ['kube-prometheus']
+    - role: kube_nginx_ingress_monitoring
```
NicolasT: Would it make sense to just merge this into the `kube_nginx_ingress` role (maybe behind a flag), and move that one after `kube_prometheus`?

roles/wait_prometheus_crd/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+- name: 'get prometheus crd'
```
NicolasT: Would it make sense to add these at the end of `kube_prometheus` instead of creating a new role?

playbooks/services.yml
Zempashi: But there is a dependency on `wait_prometheus_etcd` that will fail if kube_prometheus is not installed. And we can install the monitoring without nginx-ingress (just, that endpoints will be empty)

EDIT: removing the role

roles/kube_nginx_ingress/files/nginx_ingress_values.yml
Zempashi: we need to revert this as soon as the PR that introduce the 0.16+ image in nginx-ingress is in.

roles/wait_prometheus_crd/tasks/main.yml
Zempashi: I planed to add a `include_role` in kube_prometheus, but your idea is better

NicolasT (approved):

NicolasT:
LGTM but CI failing :-/

NicolasT (changes requested):
ChangeLog

Zempashi:
For the ChangeLog, I would like to add "added feature" with the issue number and two gitpull

NicolasT (approved):
OK if mentioned in Changelog through #203

PR #199, opened 2018-07-25
CI: Setting up VM on scality.cloud

tcarmet:
# Support scality.cloud

Applying changes to support both rackspace and scality.cloud providers

### DIFF
--- a/tests/single-node/test.sh
+++ b/tests/single-node/test.sh
@@ -19,8 +19,10 @@ setup_suite() {
         sudo losetup /dev/loop0 /kube-lvm || die
 
         echo "Disabling iptables"
-        sudo systemctl disable --now iptables || die
-        sudo systemctl disable --now ip6tables || die
+        if systemctl is-enabled --quiet iptables; then
+          sudo systemctl disable --now iptables || die
+          sudo systemctl disable --now ip6tables || die
+        fi
         sudo iptables -F || die
         sudo iptables -X || die
 
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -20,7 +20,6 @@ stages:
       type: openstack
       image: CentOS 7 (PVHVM)
       flavor: general1-4
-      #path: eve/workers/centos7
     steps:
     - Git: &git_pull
         name: git pull


tests/single-node/requirements.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+yum install iptables-services -y
```
Zempashi: Is this really a requirements or a bug in kubespray (or MetalK8s) ?

Zempashi (dismissed):

Zempashi:
`iptables-services` should not prevent the merge I believe, but It's a concern to be addressed in the product itself.

tests/single-node/requirements.sh
tcarmet: It is required on the `tests/single-node/test.sh` script otherwise the following line fail:
https://github.com/scality/metal-k8s/blob/15e6cf826a684c2fb3da933000fe1bf74acb4d71/tests/single-node/test.sh#L22

If you think that the proper fix would be to deactivate the service if it exists I can change it as well. I don't think it's required on kubespray or metalk8s AFAIK

NicolasT: If whatever image this is now running on doesn't come with the `iptables` service, there's no need to install it if we want to disable it for the test environment anyway.

The `single-node` test script should be updated to
- Check whether the `iptables` service exists
- If it exists, stop it

NicolasT:
> `iptables-services` should not prevent the merge I believe, but It's a concern to be addressed in the product itself.

I somewhat disagree. This PR apparently changes the environment in which tests are executed, so it should fix the tests to *also* work in this new environment. Admittedly it does so right now, but using a work-around rather than a proper fix.

tests/single-node/requirements.sh
tcarmet: Agree
Will fix it in the `single-node` script

tcarmet:
Made the requested changes.

I'm waiting for the test results on a VM running on scality.cloud, do not approve/merge until this build pass: https://eve-dev.devsca.com/github/scality/metal-k8s-thomas/#/builders/5/builds/16

Zempashi:
Ahh not understanding well what Thomas said, I was on the test suite not MetalK8s that we disabled iptables. Yes agree to remove iptables at all, it was a work around for rackspace.

tests/single-node/test.sh
```diff
@@ -19,8 +19,10 @@ setup_suite() {
         sudo losetup /dev/loop0 /kube-lvm || die
 
         echo "Disabling iptables"
-        sudo systemctl disable --now iptables || die
-        sudo systemctl disable --now ip6tables || die
+        if systemctl is-enabled --quiet iptables; then
+          sudo systemctl disable --now iptables || die
```
tcarmet, line 23: I have a question for you @Zempashi, I checked out a VM on rackspace and the services were not enabled, should we still keep those two lines?

NicolasT: It doesn't harm, so I'd keep it as-is.

NicolasT (approved):

tests/single-node/test.sh
tcarmet: ok to me

tcarmet:
The build passed on scality.cloud as well so if there's no comment/request for changes we can merge it.

NicolasT:
@tcarmet In general we let the author of the PR merge after approvals etc :)

tcarmet:
Oh. In that case...

PR #202, opened 2018-07-25
docs: Add logo to PDF version

NicolasT:


wabernat:
Do you have a sample of the pdf output I can take a look at?

NicolasT:
screenshot from 2018-07-25 23-40-00

It may not be the best position etc ever, but the PDF version needs more work anyway wrt LaTeX stylesheet, maybe fonts,... This is mostly a first step.

NicolasT:
To get the docs in PDF format, you can download them from ReadTheDocs: go to https://metal-k8s.readthedocs.io, then on the bottom-left corner, click on 'v:latest', optionally choose another version, and find the download PDF.

To render locally: `tox -e docs -- latexpdf`. However, this requires LaTeX tools and a whole load of packages to be installed on the system. We may want to build a Docker build env for it at some point in time...

alxf (approved):

PR #203, opened 2018-07-25
Nginx ingress dashboards

Zempashi:
Fixes: #143

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -20,6 +20,8 @@ Kubespray
 
 :ghpull:`210` - deploy `metrics-server` using Helm (:ghissue:`146`)
 
+:ghpull:`189`, :ghpull:`203` - collect `nginx-ingress` metrics and deploy a dashboard (:ghissue:`143`)
+
 Release 0.2.1 (in development)
 ==============================
 Bugs fixed
--- a/third-party.yaml
+++ b/third-party.yaml
@@ -52,6 +52,11 @@ modules:
   # the Calico license
   # See https://github.com/projectcalico/calico/pull/1167
   license: Apache-2.0
+- path: vendor/kubernetes/ingress-nginx
+  source:
+    type: url
+    base: https://raw.githubusercontent.com/kubernetes/ingress-nginx/master
+  licence: Aparche-2.0
 
 charts:
 - role: kube_heapster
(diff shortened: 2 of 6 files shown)


NicolasT (approved):
LGTM but let's hold off until #189 lands before merging

NicolasT (changes requested):
Actually, no: ChangeLog requires update.

ChangeLog.rst
```diff
@@ -15,6 +15,8 @@ Features added
 :ghpull:`193` - update versions of `kube-prometheus`, Elasticsearch and
 Kubespray
 
+:ghissue:`143` - Capture Nginx (ingress) metrics (:ghpull:`189`, :ghpull:`203`)
```
NicolasT: Nit-picky, and may be done in a later PR, but for consistency, shouldn't this be

```
:ghpull:`189`, :ghpull:`203` - collect `nginx-ingress` metrics and deploy a dashboard (:ghissue:`143`)
```

NicolasT (dismissed):
Can fixup ChangeLog before release.

NicolasT (dismissed):

NicolasT:
So, our JSON file containing all Grafana dashboards is now so big it doesn't fit in a `ConfigMap` anymore... which means we need to split things up. Luckily this is supported (https://github.com/coreos/prometheus-operator/tree/master/helm/grafana#adding-grafana-dashboards).

I guess we're better of putting every dashboard in its own `ConfigMap`, then set up `values.yml` of `kube-prometheus` properly.

I'll file a ticket for this, it's out of the realm of this one. However, if anyone starts working on it, please assign it to yourself. I may look into it later tonight.

third-party.yaml
```diff
@@ -52,6 +52,11 @@ modules:
   # the Calico license
   # See https://github.com/projectcalico/calico/pull/1167
   license: Apache-2.0
+- path: vendor/kubernetes/ingress-nginx
+  source:
+    type: url
+    base: https://raw.githubusercontent.com/kubernetes/ingress-nginx/master
+  licence: Aparche-2.0
```
NicolasT, line 59: `Apache-2.0`

NicolasT:
I created #214 which may be an approach to split things up (except that we need a fix for `node-exporter-full`). However, this is incompatible with the changes to the 'generate' script in this PR. I believe, however, the changes are fairly easy to port.

PR #215, opened 2018-07-27
Grafana dashboard for Nginx metrics

Zempashi:
Fixes: #143

### DIFF
--- a/roles/kube_prometheus/hack/fix-dashboard.py
+++ b/roles/kube_prometheus/hack/fix-dashboard.py
@@ -12,8 +12,13 @@ DEFAULT_SOURCE = 'DS_DUMMY'
 def is_dashboard(dashboard):
     return all([
         'annotations' in dashboard,
-        'rows' in dashboard,
+        'rows' in dashboard or 'panels' in dashboard,
     ])
+
+def _iter_panels(dashboard):
+    for row in dashboard.get('rows') or [dashboard]:
+        for panel in row['panels']:
+            yield panel
 
 
 def main(dashboard_data, output, filename, title=None):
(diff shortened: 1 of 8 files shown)


roles/kube_prometheus/vars/main.yml
```diff
@@ -0,0 +1,7 @@
+kube_prometheus_dashboards:
+  - kube-prometheus-calico2-dashboard
+  - kube-prometheus-calico-dashboard
+  - kube-prometheus-elasticsearch-dashboard
+  - kube-prometheus-etcd-dashboard
+  #- kube-prometheus-node-exporter-dashboard
+  - kube-prometheus-prometheus-dashboard
```
NicolasT: If the `nginx-ingress` dashboard is not listed here, it will not be deployed :)

NicolasT (changes requested):
I believe the dashboard needs to be added to the list of deployed dashboards in the role

roles/kube_prometheus/vars/main.yml
Zempashi: Yes. We also might want to put this in `defaults/main.yml` to be overriden if needed (a dashboard that get problem to be deployed for example...)

NicolasT: Agree, at some later point :)

third-party.yaml
```diff
@@ -52,6 +52,11 @@ modules:
   # the Calico license
   # See https://github.com/projectcalico/calico/pull/1167
   license: Apache-2.0
+- path: vendor/kubernetes/ingress-nginx
+  source:
+    type: url
+    base: https://raw.githubusercontent.com/kubernetes/ingress-nginx/master
+  licence: Aparche-2.0
```
NicolasT: `Apache-2.0`

roles/kube_prometheus/hack/fix-dashboard.py
```diff
@@ -12,9 +12,14 @@
 def is_dashboard(dashboard):
     return all([
         'annotations' in dashboard,
-        'rows' in dashboard,
+        'rows' in dashboard or 'panels' in dashboard,
     ])
 
+def _iter_panels(dashboard):
+    for row in dashboard.get('rows') or [dashboard]:
```
NicolasT, line 19: I'd write `dashboard.get('rows', [dashboard])`, but fine :)

NicolasT (changes requested):
Please sync with target branch and fix the license typo.

NicolasT:
I'm on it.

NicolasT (approved):

Issue #174, opened 2018-07-20
Kibana: Pre-provision indexing configuration


NicolasT:

alxf:
Documentation can be found here: https://github.com/elastic/kibana/issues/3709

PR #223, opened 2018-07-30
Pre-provision Kibana index configuration

alxf:
Fixes: #174


roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+
+#
+# Script to setup the default index pattern in kibana
+#
+
+INDEX_ID="logstash-*"
+
+KIBANA="http://kibana:443"
```
Zempashi: We should variable this. Suggestion to use an environment variable, that will be injected by another key of the configmap

NicolasT: These are exported as variables already:
```shell
$ kubectl run -ti shell --image=busybox --namespace=kube-ops --restart=Never
If you don't see a command prompt, try pressing enter.
/ # env | grep KIBANA
KIBANA_PORT_443_TCP=tcp://10.233.41.254:443
KIBANA_SERVICE_HOST=10.233.41.254
KIBANA_PORT=tcp://10.233.41.254:443
KIBANA_SERVICE_PORT=443
KIBANA_PORT_443_TCP_ADDR=10.233.41.254
KIBANA_PORT_443_TCP_PORT=443
KIBANA_PORT_443_TCP_PROTO=tcp
```

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Wait for kibana service to be ready
+while [ "$API_READY" -eq 0 ]; do
+    http_code=$(curl -s -o /dev/null -w "%{http_code}" "$KIBANA")
+    if [ "$http_code" = "200" ]; then
+        API_READY=1
+    else
+        sleep 30
+    fi
+done
```
NicolasT: Let's make this a non-infinite loop on failure.

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
...
+
+http_code=$(set_index_pattern)
+
+if [ "$http_code" = "409" ]; then
+    # Index is already registered
+    exit 0
+fi
+
+if [ "$http_code" != "200" ]; then
+    # An error occured
```
NicolasT: Log some messages.

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+
+#
+# Script to setup the default index pattern in kibana
+#
+
+INDEX_ID="logstash-*"
```
NicolasT: Consider moving this outside of the script, into the job definition, and have an assertion it's set here.

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+
```
NicolasT: `set -ue`?

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
...
+    else
+        sleep 30
+    fi
+done
+
+http_code=$(set_index_pattern)
+
+if [ "$http_code" = "409" ]; then
+    # Index is already registered
+    exit 0
```
NicolasT: Index pattern is registered, but maybe default index is not set, as in during a previous run this succeeded but the second step failed.

roles/cluster_logging/templates/job-setup-configmap.yml.j2
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: job-setup-configmap
```
NicolasT: Needs some labels as well, and a more descriptive name.

roles/cluster_logging/templates/job-setup.yml.j2
```diff
@@ -0,0 +1,24 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: provision-kibana-index
+  labels:
```
NicolasT: Make sure it gets eagerly GCed upon success.

roles/cluster_logging/templates/job-setup.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+      name: job-setup-configmap
+    spec:
+      containers:
+        - name: kibana-job-setup
+          image: appropriate/curl
+          command: ['sh', '-c', '/scripts/provision-index.sh']
+          volumeMounts:
+            - name: job-setup-configmap
+              mountPath: /scripts
+      restartPolicy: Never
```
NicolasT: `OnFailure`?

roles/cluster_logging/templates/job-setup.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+      volumes:
+        - name: job-setup-configmap
+          configMap:
+            defaultMode: 0755
```
NicolasT: No need when invoking `sh` on the file.

roles/cluster_logging/templates/job-setup.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+    app: kibana-job-setup
+spec:
+  template:
+    metadata:
+      name: job-setup-configmap
+    spec:
+      containers:
+        - name: kibana-job-setup
+          image: appropriate/curl
+          command: ['sh', '-c', '/scripts/provision-index.sh']
```
NicolasT: Why the `-c`?

NicolasT (changes requested):
Some remarks, and needs an entry in the `ChangeLog`.

roles/cluster_logging/files/kibana/provision-index.sh
Zempashi: Yes but those are environment variable, not config (what if mainstream or us change 'kibana' to 'kibana-ui' or 'kibana-api')

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
```
Zempashi, line 1: Now the integration with the API is only return code. This works with few issues (like kibana claiming uuid problem at start). Could be small enough to decide to change this to python

NicolasT: We discussed this during standup. @alxf asked for a Docker image containing Python & `requests`, which I thought is a bit overkill when the same can, for now, be achieved using shell + `curl`. If need be, this could be changed at some point.

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,69 @@
...
+    http_code=$(curl -s -o /dev/null -w "%{http_code}" "$API_HOSTNAME")
+    if [ "$http_code" = "200" ]; then
+        break
+    else
+        COUNTDOWN=$((COUNTDOWN - 1))
+        if [ "$COUNTDOWN" -le 0 ]; then
+            echo "Error: Can't connect to the kibana api server"
+            exit 1
+        else
+            sleep 30
```
NicolasT: Consider bumping the `COUNTDOWN` initial value, and lowering the sleep time for more responsiveness

roles/cluster_logging/templates/job-setup-configmap.yml.j2
NicolasT: Point being: these objects are 'global' in the `kube-ops` namespace, so a name like `kibana-index-pattern-setup` could be better.

roles/cluster_logging/templates/job-setup.yml.j2
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: provision-kibana-index
+  labels:
+    app: kibana-job-setup
+spec:
+  template:
+    metadata:
+      name: job-setup-configmap
```
NicolasT: This ain't no `ConfigMap` :)

ChangeLog.rst
```diff
@@ -26,6 +26,8 @@ Kubespray
 
 :ghpull:`218` - update versions of Kibana and `fluent-bit`
 
+:ghpull:`223` - Pre-provision kibana index configuration (:ghissue:`174`)
```
NicolasT: For consistency: `pre-provision` (no capital) and `Kibana` (capital).

NicolasT (approved):
Tested manually. If CI passes, this is good to go as far as I'm concerned.

roles/cluster_logging/files/kibana/provision-index.sh
```diff
@@ -0,0 +1,69 @@
...
+
+#
+# Script to setup the default index pattern in kibana
+#
+
+set -ue
+
+INDEX_ID="logstash-*"
+
+API_HOSTNAME="$KIBANA_SERVICE_HOST:$KIBANA_SERVICE_PORT"
```
Zempashi, line 11: Which is the same service hardcoded differently (now the port is variable). I'm good with that for v1, but still something that need a ticket

PR #232, opened 2018-08-01
Validate the device presence on the host

ballot-scality:
Checks implemented now.

Ready for review. Will do some tests with the new framework in another PR

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -10,6 +10,9 @@ Features added
 
 :ghpull:`305` - ensure that journald logs are persisted across reboot (:ghissue:`303`)
 
+:ghpull:`232` - Add more storage checks regarding the device presence and the
+partition existence on specificied drives (:ghissue:`231`)
+
 Bugs fixed
 ----------
 :ghissue:`50` - raise default `etcd` memory limits (:ghpull:`331`)
--- a/roles/metalk8s_lvm_vg/examples/multipath_fact.json
+++ b/roles/metalk8s_lvm_vg/examples/multipath_fact.json
@@ -1,0 +1,11051 @@
+192.168.78.28 | SUCCESS => {
+    "ansible_facts": {
+        "ansible_all_ipv4_addresses": [
+            "192.168.78.28"
+        ],
+        "ansible_all_ipv6_addresses": [
+            "fe80::46a8:42ff:fe49:dc79",
+            "fe80::46a8:42ff:fe49:dc79",
+            "fe80::46a8:42ff:fe49:dc79"
+        ],
+        "ansible_apparmor": {
+            "status": "disabled"
+        },
+        "ansible_architecture": "x86_64",
+        "ansible_bios_date": "06/03/2015",
+        "ansible_bios_version": "1.3.6",
+        "ansible_bond0": {
+            "active": true,
+            "device": "bond0",
+            "features": {
+                "fcoe_mtu": "off [fixed]",
+                "generic_receive_offload": "on",
+                "generic_segmentation_offload": "on",
+                "highdma": "on",
+                "large_receive_offload": "on",
+                "loopback": "off [fixed]",
+                "netns_local": "off [fixed]",
+                "ntuple_filters": "off [fixed]",
+                "receive_hashing": "off [fixed]",
+                "rx_checksumming": "off [fixed]",
+                "rx_vlan_filter": "on",
+                "rx_vlan_offload": "on",
+                "scatter_gather": "on",
+                "tcp_segmentation_offload": "on",
+                "tx_checksum_fcoe_crc": "off [fixed]",
+                "tx_checksum_ip_generic": "on",
+                "tx_checksum_ipv4": "off [fixed]",
+                "tx_checksum_ipv6": "off [fixed]",
+                "tx_checksum_sctp": "off [fixed]",
+                "tx_checksum_unneeded": "off [fixed]",
+                "tx_checksumming": "on",
+                "tx_fcoe_segmentation": "off [fixed]",
+                "tx_gre_segmentation": "off [fixed]",
+                "tx_gso_robust": "off [fixed]",
+                "tx_lockless": "on [fixed]",
+                "tx_scatter_gather": "on",
+                "tx_scatter_gather_fraglist": "off [requested on]",
+                "tx_tcp6_segmentation": "on",
+                "tx_tcp_ecn_segmentation": "on",
+                "tx_tcp_segmentation": "on",
+                "tx_udp_tnl_segmentation": "off [fixed]",
+                "tx_vlan_offload": "on",
+                "udp_fragmentation_offload": "off [fixed]",
+                "vlan_challenged": "off [fixed]"
+            },
+            "hw_timestamp_filters": [],
+            "ipv4": {
+                "address": "192.168.78.28",
+                "broadcast": "192.168.78.255",
+                "netmask": "255.255.255.0",
+                "network": "192.168.78.0"
+            },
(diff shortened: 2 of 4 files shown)


roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,160 @@
...
+    Raises:
+        AssertionError: if the device is not present on the host
+    '''
+
+    for lvm_vg in \
+        hostvars.get(
+            'metalk8s_lvm_all_vgs', {}).keys():
+        mk8s_vg_drives_var = 'metalk8s_lvm_drives_' + lvm_vg
+        for device in hostvars.get(mk8s_vg_drives_var):
+            device_name = device.split('/')[-1]
```
Zempashi: I would try to strip the first '/dev/', and check that the result contains no '/'.

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,160 @@
...
+        def collect_checks():
+            for (name, obj) in inspect.getmembers(sys.modules[__name__]):
+                if name.startswith('check_') and inspect.isfunction(obj):
+                    yield (name, obj)
+
+        errors = []
+        failed = False
+
+        for (name, check) in collect_checks():
+            for host in task_vars['hostvars'].keys():
```
Zempashi: We could also restrict to 'kube-nodes' instead of all nodes

roles/setup_lvm_vg/action_plugins/validate_storage.py
ballot-scality: Good point !

roles/setup_lvm_vg/action_plugins/validate_storage.py
ballot-scality: Indeed. Although it works because the variable metalk8s_lvm_all_vgs will be only defined on hosts with storage, it seems more logical to restrain the scope of the hosts tested.

alxf (dismissed):

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,167 @@
...
+        def collect_checks():
+            for (name, obj) in inspect.getmembers(sys.modules[__name__]):
+                if name.startswith('check_') and inspect.isfunction(obj):
+                    yield (name, obj)
+
+        errors = []
+        failed = False
+
+        for (name, check) in collect_checks():
+            for host in task_vars['groups'].get('kube-node', []):
```
Zempashi: I'm sorry, that's me pointing you to that. But with more thinking of this. It feels a bit wrong to use a group name here in a tasks. What if the tasks is run on another sets of host ? Might be better to use 'ansible_play_hosts' magic variable.

NicolasT:
Target 1.1 iso 1.0?

ballot-scality:
@NicolasT not sure anymore of what you mean with your comment. Do you want me to target 1.0 or 1.1 ?

ChangeLog.rst
```diff
@@ -7,6 +7,8 @@ Features added
 :ghpull:`222` - update Elasticsearch and `node_exporter` dashboards
 
 :ghpull:`225` - Figure out the "fail fast" option of ansible (:ghissue:`129`)
```
Zempashi: we add empty line between entries, don' t know what is a good practice or not

roles/metalk8s_lvm_vg/examples/multipath_fact.json
```diff
@@ -0,0 +1,11051 @@
...
+                "tx_tcp_ecn_segmentation": "on",
+                "tx_tcp_segmentation": "on",
+                "tx_udp_tnl_segmentation": "off [fixed]",
+                "tx_vlan_offload": "on",
+                "udp_fragmentation_offload": "off [fixed]",
+                "vlan_challenged": "off [fixed]"
+            },
+            "hw_timestamp_filters": [],
+            "ipv4": {
+                "address": "192.168.78.28",
```
Zempashi, line 58: You may want to change of the value it give a lot of information publicly (it's already public now...but yet...)

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,460 @@
...
+        'setup' ansible module
+    :returns: True if the device is a raw device. False otherwise
+    :rtype: bool
+    :returns: A tuple with True, and the partition details from ansible_devices
+        if the device is a partition. False otherwise
+    :rtype: tuple
+    '''
+
+    for ansible_device, device_attr in ansible_devices.items():
+        if device in device_attr['partitions']:
```
Zempashi: In python we often try and apologize, rather to ask first:
```
for ansible_device, device_attr in ansible_devices.items():
   try:
        return True, device_attr['partitions'][device]
   except KeyError:
        continue  # May also pass if device_attr has no 'partitions' key
return False, {}
```

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,460 @@
...
+    assert not uuid, (
+        "The device {dev} is used by the fileystem with UUID {uuid}. "
+        "Please remove it.".format(
+            dev=device,
+            uuid=uuid,
+        )
+    )
+
+
+def is_drive_used(device, lvm_vg, hostvars):
```
Zempashi: Create class and access variable via `self` might get rid of the burden of passing variable like hostvars, ansible_devices.
(but I'm definitely a class-person in python development, so sometime I'm wrong)

ballot-scality: Agreed, but let's do that in another PR then :)

roles/metalk8s_lvm_vg/examples/multipath_fact.json
ballot-scality: I'm personally not worried by a class C ip address but I might be wrong...
Any other thoughts on this ?

ballot-scality:
Rebasing to target development/1.0

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,462 @@
...
+
+    .. code::
+
+      ansible-playbook -i {inventory} -t storage playbooks/deploy.yml
+
+A sample of a device dictionary as seen with ansible setup module:
+
+  .. code::
+
+    "vdd": {
```
Zempashi: I would add 'vars' up to the top level tree (how to get to this leaf)

roles/setup_lvm_vg/action_plugins/validate_storage.py
Zempashi: Sure

roles/setup_lvm_vg/action_plugins/validate_storage.py
```diff
@@ -0,0 +1,470 @@
...
+
+    # if the device is a raw device, check its presence
+    if is_drive_partition(device, ansible_devices) \
+       or is_drive_raw_device(device, ansible_devices):
+        return True
+
+    return False, {}
+
+
+def is_used_by_filesystem(device, uuid):
```
Zempashi, line 183: Sorry to be annoying about that: some function is_* return boolean, other raise AssertionError. Could we use two different prefix to distinguish the difference between the API ?

ballot-scality:
Rename branch

Issue #190, opened 2018-07-24
Implement a (new) 'helm' Ansible module


NicolasT:
- Allow to set `values` in task
- Allow to keep `values` in `files` and injected (can use Ansible for this)
- Allow to set `values` in inventory, and apply using proper `-f ...` sequence
- Set repository, name, version,...

I think it'd make sense to first implement this inside MetalK8s, then later we may want to extract in a separate project/repository.

When the above functionality is implemented, #180 becomes possible as a side-effect.

NicolasT:
@Zempashi Since you put this in 1.1.0 (which I was about to do now ;)), are you working on this? If so, please self-assign.

Zempashi:
Yes, I do. We can move this for later release, I wasn't aware we had enough feature to release 1.1.0, it was just show that's something we want to work on "soonish". (And this will help #230)

NicolasT:
Oh no, no 1.1.0 release planned soon, so this can definitely go in, and is indeed a good step to handle #230, #180 and others!

PR #238, opened 2018-08-03
Add an Ansible module  to handle Helm chart installation

Zempashi:
Fixes: #190

### DIFF
--- a/roles/helm_common/library/helm_cli.py
+++ b/roles/helm_common/library/helm_cli.py
@@ -1,0 +1,236 @@
+import os
+import json
+
+from contextlib import contextmanager
+from tempfile import NamedTemporaryFile
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class Helm(AnsibleModule):
+
+    def __init__(self, *args, **kwargs):
+        AnsibleModule.__init__(
+            self,
+            argument_spec=dict(
+                chart=dict(type='dict'),
+                release=dict(type='str', aliases=['name']),
+                namespace=dict(type='str'),
+                binary=dict(type='str'),
+                state=dict(default='present',
+                           choice=['latest', 'present', 'absent', 'purged']),
+                values=dict(type='list'),
+            )
+        )
+        self._helm_bin = None
+        self.delete_temp_file = True
+
+    def get_chart(self):
+        chart = self.params.get('chart')
+        if chart is None:
+            raise HelmError("'chart' is a mandatory argument for "
+                            "state=present|latest")
+        if 'name' not in chart:
+            raise HelmError("'chart' argument required a 'name' key for "
+                            "state=present|latest")
+        return chart
+
+    @property
+    def helm_bin(self):
+        if self._helm_bin is None:
+            binary = self.params.get('binary')
+            if binary:
+                if os.path.isfile(binary) and os.access(binary, os.X_OK):
+                    self._helm_bin = binary
+                else:
+                    raise HelmError(
+                        "No executable at the specified 'binary' location")
+            else:
+                self._helm_bin = self.get_bin_path('helm', required=True)
+        return self._helm_bin
+
+    def get_release_info(self, release, command='get'):
+        assert command in ['get', 'status']
+
+        def failure_detect(rc, out, err):
+            return not(rc == 0 or
+                       'release: "{}" not found'.format(release) in err)
+
+        rc, out, err = self._run_helm([command, release],
+                                      failed_when=failure_detect)
+        if rc != 0:
+            return
+        release_info = self._parse_helm_output(out)
+        return release_info
+
+    def install(self):
+        chart = self.get_chart()
+        release = self.params.get('release')
+        if release is None:
+            cmd = ['install', chart['name']]
+        else:
+            cmd = ['--install', release, 'upgrade', chart['name']]
+        if 'version' in chart:
+            cmd.extend(['--version', chart['version']])
+        if 'repo' in chart:
+            cmd.extend(['--repo', chart['repo']])
+        namespace = self.params.get('namespace')
+        if namespace:
+            cmd.extend(['--namespace', namespace])
+        with self.write_values_files() as values_filenames:
+            for value in values_filenames:
+                cmd.extend(['--values', value])
+            rc, out, err = self._run_helm(cmd)
+        return self._parse_helm_output(out)
+
+    def _run_helm(self, cmd, failed_when=lambda rc, out, err: rc != 0):
+        try:
+            rc, out, err = self.run_command([self.helm_bin] + cmd)
+        except Exception as exc:
+            raise HelmError('error running helm {} command: {}'.format(
+                ' '.join(cmd), exc))
+        else:
+            if failed_when(rc, out, err):
+                raise HelmError(
+                    'Error running helm {} command (rc={}) '
+                    "out='{}' err='{}'".format(' '.join(cmd), rc, out, err))
+        return rc, out, err
+
+    @staticmethod
+    def _helm_std_failure(rc, out, err):
+        return rc != 0
+
+    def ensure_present(self):
+        release = self.params.get('release')
+        if release:
+            release_status = self.get_release_info(release, command='status')
+        else:
+            release_status = None
+
+        if release_status is None or release_status['STATUS'] != 'DEPLOYED':
+            install = self.install()
+            release_info = self.get_release_info(release or install['NAME'])
+            return {'changed': True,
+                    'release': release_info,
+                    'install': install}
+        elif release_status['STATUS'] == 'DEPLOYED':
+            release_info = self.get_release_info(release)
+            return {'changed': False,
+                    'release': release_info}
+        else:
+            raise HelmError("Release found but it's status is not 'DEPLOYED'")
+
+    def install_or_upgrade(self):
+        install = self.install()
+        release_info = self.get_release_info(
+            self.params.get('release', install.get('NAME'))
+        )
+        return {'changed': True,
+                'release': release_info}
+
+    def remove(self, purge=False):
+        release = self.params.get('release')
+        if not release:
+            raise HelmError(
+                "'release' is a mandatory argument for state=absent")
+        release_status = self.get_release_info(release, command='status')
+        release_info = self.get_release_info(release)
+        if release_status is None or \
+                (release_status['STATUS'] == 'DELETED' and not purge):
+            return {'changed': False,
+                    'release': release_info}
+        else:
+            cmd = ['delete', release]
+            if purge:
+                cmd.append('--purge')
+            self._run_helm(cmd)
+            return {'changed': True,
+                    'release': release_info}
+
+    def execute(self):
+        state = self.params.get('state')
+        if state == 'present':
+            return self.ensure_present()
+        elif state == 'latest':
+            return self.install_or_upgrade()
+        elif state == 'absent':
+            return self.remove()
+        elif state == 'purged':
+            return self.remove(purge=True)
+        else:
+            raise HelmError('state not supported')
+
+    @contextmanager
+    def write_values_files(self):
+        values_filenames = []
+        values_content = self.params.get('values', [])
+        for values in values_content:
+            temp = NamedTemporaryFile(
+                suffix='.yaml',
+                prefix='ansible.helm.values.',
+                delete=False,
+            )
+            if isinstance(values, dict):
+                values_str = json.dumps(values)
+            else:
+                values_str = values
+            temp.write(values_str)
+            temp.close()
+            values_filenames.append(temp.name)
+        try:
+            yield values_filenames
+        finally:
+            if self.delete_temp_file:
+                for filename in values_filenames:
+                    os.remove(filename)
+
+    def _parse_helm_output(self, helm_output):
+        output_parsed = {}
+        multiline_key = None
+        for n, line in enumerate(helm_output.splitlines()):
+            line_split = line.split(':', 1)
+            try:
+                (key, value) = line_split
+            except ValueError:
+                pass
+            else:
+                if key.strip().upper() == key:
+                    if value.strip() == '':
+                        multiline_key = key
+                        output_parsed[key] = []
+                        continue
+                    else:
+                        output_parsed[key] = value.strip()
+                        multiline_key = None
+                        continue
+            if multiline_key is not None:
+                output_parsed[multiline_key].append(line)
+                continue
+            if n == 0:
+                output_parsed['message'] = line
+                continue
+            if line == '':
+                continue
+            raise ValueError(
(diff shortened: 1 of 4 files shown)


roles/kube_nginx_ingress/tasks/main.yml
```diff
@@ -1,34 +1,22 @@
...
-    --namespace {{ nginx_ingress_namespace }}
-    -f {{ nginx_ingress_value_file.path }}
+  helm_cli:
+    release: '{{ nginx_ingress_release_name }}'
+    chart:
+      name: nginx-ingress
+      version: '{{ nginx_ingress_version }}'
+      repo: '{{ nginx_ingress_repo }}'
+    namespace: '{{ nginx_ingress_namespace }}'
+    binary: '/usr/local/bin/'
```
NicolasT: Currently the module allows to provide a directory and then looks up `helm` in it. This is overkill, a `binary` configuration entry should point to the binary, not 'a directory in which the binary may be found'.

Also, this should be `'{{ bin_dir }}/helm'`

roles/kube_nginx_ingress/meta/main.yml
```diff
@@ -1,2 +1,3 @@
 dependencies:
+    - role: helm_module
```
NicolasT: Do we really need to introduce a new role for this? Can't this go in `helm_common`?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
+import os
+import json
+
+from contextlib import contextmanager
+from tempfile import NamedTemporaryFile
+
+from ansible.module_utils.basic import *  # noqa
```
NicolasT: :disappointed:

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def __init__(self, *args, **kwargs):
+        AnsibleModule.__init__(
+            self,
+            argument_spec=dict(
+                chart=dict(type='dict'),
+                release=dict(type='str', aliases=['name']),
+                namespace=dict(type='str'),
+                binary=dict(type='str'),
+                state=dict(default='present', choice=['present', 'absent']),
```
NicolasT: `present` or `latest`? Doesn't this module implement the latter behaviour?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+                values_list=dict(type='list'),
```
NicolasT: Change `values_list` -> `values`, which indeed is a list?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+        self._helm_bin = None
+        self.delete_temp_file = True
+
+    def get_chart(self):
+        chart = self.params.get('chart')
+        if chart is None:
+            raise HelmError("'chart' is a mandatory argument for state=present")
+        if 'name' not in chart:
+            raise HelmError("'chart' argument required a 'name' key for state=present")
```
NicolasT: `required` -> `requires`

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        return chart
+
+    @property
+    def helm_bin(self):
+        if self._helm_bin is None:
+            binary = self.params.get('binary')
+            if binary and os.path.isfile(binary):
+                self._helm_bin = binary
+            else:
+                self._helm_bin = self.get_bin_path(
```
NicolasT: See below. `binary` should point to the binary, nothing more nothing less.

NicolasT: If `binary` is not set, then it can be looked up in `$PATH`, but without adding the `binary` value to it first.

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def get_release_info(self, release):
+        def failure_detect(rc, out, err):
+            return not(rc == 0 or
+                       'release: "{}" not found'.format(release) in err)
+
+        rc, out, err = self._run_helm(['get', release],
+                                      failed_when=failure_detect)
+        if rc != 0:
+            return
```
NicolasT: I know it's implicit, but `return None`?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+                continue
+            raise ValueError(
+                'Cannot have an unformatted line in a '
+                'non-multine block: "{}" (n {})'.format(line, n)
+            )
+        return output_parsed
+
+
+class HelmError(Exception):
+    """Error from Kubectl Module"""
```
NicolasT: Not `Kubectl` :wink:

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+            raise HelmError('error running helm {} command: {}'.format(
+                ' '.join(cmd), exc))
+        else:
+            if failed_when(rc, out, err):
+                raise HelmError('Error running helm {} command (rc={}) '
+                                "out='{}' err='{}'".format(' '.join(cmd), rc, out, err))
+        return rc, out, err
+
+    @staticmethod
+    def _helm_std_failure(rc, out, err):
```
NicolasT: This is used in `_run_helm` only, could be inlined there or turned into a lambda.

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+                                      failed_when=failure_detect)
+        if rc != 0:
+            return
+        release_info = self._parse_helm_output(out)
+        return release_info
+
+    def install(self):
+        chart = self.get_chart()
+        release = self.params.get('release')
+        if release is None:
```
NicolasT: There's no way to implement actual `present` without a given release name. Hence, I think (as said above) the actual implementation here is `latest`, not `present`, and if `present` were implemented it'd require `release` to be set.

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+            rc, out, err = self._run_helm(cmd)
+        return self._parse_helm_output(out)
+
+    def _run_helm(self, cmd, failed_when=None):
+        if failed_when is None:
+            failed_when = self._helm_std_failure
+        try:
+            rc, out, err = self.run_command([self.helm_bin] + cmd)
+        except Exception as exc:
+            raise HelmError('error running helm {} command: {}'.format(
```
NicolasT: I thought Ansible modules are never supposed to `raise`, but instead return JSON with `error` set or something along those lines.

NicolasT: Use quotes: `Error running 'helm {}': {}`

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+                ' '.join(cmd), exc))
+        else:
+            if failed_when(rc, out, err):
+                raise HelmError('Error running helm {} command (rc={}) '
```
NicolasT: Quotes

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        return rc != 0
+
+    def check_present(self):
+        release = self.params.get('release')
+        if release:
+            release_info_pre = self.get_release_info(release)
+        else:
+            release_info_pre = None
+
+        install = self.install()
```
NicolasT: So `check_present` actually runs `install`? :confused:

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        release_info_post = self.get_release_info(release or install['NAME'])
+        return {'changed': release_info_pre != release_info_post,
+                'release_info_pre': release_info_pre,
+                'install': install,
+                'release_info_post': release_info_post}
+
+    def execute(self):
+        state = self.params.get('state')
+        if state == 'present':
```
NicolasT: `absent` not handled, despite the code in `__init__`.

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        values_content = self.params.get('values_list', [])
+        for values in values_content:
+            temp = NamedTemporaryFile(
+                suffix='.yaml',
+                prefix='ansible.helm.values.',
+                delete=False,
+            )
+            if isinstance(values, dict):
+                values_str = json.dumps(values)
+            else:
```
NicolasT: What is it if it's not a dict? A list?
What I'm getting at: shouldn't `values` be unconditionally `json.dumps`ed?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+        values_filenames = []
+        values_content = self.params.get('values_list', [])
+        for values in values_content:
+            temp = NamedTemporaryFile(
+                suffix='.yaml',
+                prefix='ansible.helm.values.',
+                delete=False,
+            )
+            if isinstance(values, dict):
+                values_str = json.dumps(values)
```
NicolasT: You can use `json.dump` to `temp`, no?

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+            else:
+                values_str = values
+            temp.write(values_str)
+            temp.close()
```
NicolasT: The code above leaks an FD if `json.dumps` or `temp.write` fails.

roles/helm_module/library/helm_cli.py
```diff
@@ -0,0 +1,190 @@
...
+            values_filenames.append(temp.name)
+        try:
+            yield values_filenames
+        finally:
+            if self.delete_temp_file:
+                for filename in values_filenames:
+                    os.remove(filename)
+
+    def _parse_helm_output(self, helm_output):
```
NicolasT: Please provide a doc-block with an example of input that's being parsed here. Also, I think this needs tests, so we're sure this still works when we bump Helm version.

roles/kube_nginx_ingress/tasks/main.yml
```diff
@@ -1,34 +1,22 @@
...
+    values_list:
+    - "{{ lookup('file', role_path ~ '/files/nginx_ingress_values.yml') }}"
```
NicolasT: Is there a way to include the YAML content of a file here, instead of the file as a string? This would cause Ansible to pre-parse the YAML, ruling out formatting issues at playbook init time rather than runtime.

NicolasT (changes requested):

roles/kube_nginx_ingress/tasks/main.yml
Zempashi: Both works, 'get_bin_path' takes an opt_dirs for optionnal directory to look into

roles/kube_nginx_ingress/meta/main.yml
Zempashi: indeed

roles/helm_module/library/helm_cli.py
Zempashi: I realize this in my holidays, latest could help us to workaround the indempotency of the module. Proposal:
- present: just check that the release with the same chart name is present. If not: install (this one is idempotent
- latest: Create a new revision no matter what. This module should always trigger change. (And might be what we use in our different roles)

roles/kube_nginx_ingress/tasks/main.yml
Zempashi: But agree for '{{ bin_dir }}/helm'

roles/helm_module/library/helm_cli.py
Zempashi: EDIT:
- absent: removal of release without purge
- purge: removal of release with --purge

NicolasT: `present` should be `a release with the same release name is present`, not the same chart name. Otherwise, sounds good to me!

roles/helm_common/library/helm_cli.py
```diff
@@ -0,0 +1,235 @@
...
+        return rc != 0
+
+    def check_present(self):
+        release = self.params.get('release')
+        if release:
+            release_status = self.get_release_info(release, command='status')
+        else:
+            release_status = None
+
+        if release_status is None or release_status['STATUS'] != 'DEPLOYED':
```
Zempashi, line 110: Not sure If we should try to install when "state=present" and a release exist but in "status=DELETED". Gathering advice.

roles/helm_common/library/helm_cli.py
```diff
@@ -0,0 +1,235 @@
...
+        return chart
+
+    @property
+    def helm_bin(self):
+        if self._helm_bin is None:
+            binary = self.params.get('binary')
+            if binary and os.path.isfile(binary):
+                self._helm_bin = binary
+            else:
+                self._helm_bin = self.get_bin_path(
```
NicolasT: I still think supporting `binary` to be a path to a directory rather than the expected binary doesn't make sense :) No need to be too 'clever' and work-around bugs written by a consumer of this module.

roles/helm_common/library/helm_cli.py
```diff
@@ -0,0 +1,235 @@
...
+                ' '.join(cmd), exc))
+        else:
+            if failed_when(rc, out, err):
+                raise HelmError(
+                    'Error running helm {} command (rc={}) '
+                    "out='{}' err='{}'".format(' '.join(cmd), rc, out, err))
+        return rc, out, err
+
+    @staticmethod
+    def _helm_std_failure(rc, out, err):
```
NicolasT, line 100: Unused outside `_run_helm`, so inline this there (as a lambda), or also use it at other places where `rc != 0` is checked, but that feels overengineered.

roles/helm_common/library/helm_cli.py
NicolasT: Bail out. A `DELETED` deployment is not `present`, but there's nothing one can do about it. Similarly, a `FAILED` deployment should be handled somehow I guess.

roles/helm_common/library/helm_cli.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def check_present(self):
+        release = self.params.get('release')
+        if release:
+            release_status = self.get_release_info(release, command='status')
+        else:
+            release_status = None
+
+        if release_status is None or release_status['STATUS'] != 'DEPLOYED':
+            install = self.install()
```
NicolasT, line 111: Still confused `check_present` runs `install`.

Zempashi: 'ensure_present' is a better name ?

Issue #237, opened 2018-08-03
Deployment of `prometheus-operator` sometimes fails


NicolasT:
We've seen the deployment of `prometheus-operator` fail a couple of times, though definitely not repeatedly. Happened just today on @philipyoo's cluster.

When I was able to reproduce this a couple of weeks ago, the root cause was bad network connectivity, which caused the container image for the `prometheus-operator-create-sm` job to be pulled very slowly, which causes Helm to give up after a while. There may, however, be other (unknown) reasons as well.

After this fails, one needs to manually clean up things before the playbook can be restarted:

- `helm delete --purge prometheus-operator`
- `kubectl --namespace=kube-ops delete job prometheus-operator-create-sm`
- `kubectl --namespace=kube-ops delete job prometheus-operator-get-crd`

This is not very user-friendly. We should at least document what to do when the step fails, and ideally increase the Helm deployment timeout for this particular Chart.

NicolasT:
Docs are in, and hopefully the new timeout reduces the odds of running into this issue.

mastachand:
I didn't see an issue with 600 second timeout when i added it manually, however, because it had failed the first time, i got an error and I had to manually delete and purge prometheus-operator on one of the machine or the re-run of the playbook with the new timeout would fail.

this seems to be related to this: https://github.com/helm/helm/issues/3353

NicolasT:
@mastachand Hence the PRs resulting from this ticket :wink: See https://metal-k8s.readthedocs.io/en/latest/faq.html#deployment-failed-at-install-prometheus-operator-what-should-i-do !

PR #243, opened 2018-08-04
docs: Document 'install prometheus-operator' failure

NicolasT:
See: #237
See: https://github.com/scality/metalk8s/issues/237

PR #244, opened 2018-08-04
kube_prometheus: Increase `prometheus-operator` install timeout

NicolasT:
Fixes: #237
See: https://github.com/scality/metalk8s/issues/237
See: #243 
See: https://github.com/scality/metalk8s/pull/243


NicolasT:
Putting this in 1.0.1, unless someone objects.

alxf (approved):

PR #249, opened 2018-08-07
Tag Grafana dashboards

alxf:
Fix #208, but only the tag part of the ticket.

PR #251, opened 2018-08-08
Tag Grafana dashboards

### LINKED ISSUE #208: Label and organize Grafana dashboards
We currently deploy all dashboards at the 'root'. Adding labels, and putting them in folders, would be good.

alxf:
Fix #208

See: #249 
See: https://github.com/scality/metalk8s/pull/249

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -9,6 +9,8 @@ Features added
 
 :ghpull:`222` - update Elasticsearch and `node_exporter` dashboards
 
+:ghpull:`208` - tag Grafana dashboards (:ghissue:`251`)
+
 Release 1.0.1 (in development)
 ==============================
 Features added
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -9,7 +9,7 @@ Features added
 
 :ghpull:`222` - update Elasticsearch and `node_exporter` dashboards
 
-:ghpull:`208` - tag Grafana dashboards (:ghissue:`251`)
+:ghpull:`251` - tag Grafana dashboards (:ghissue:`208`)
 
 Release 1.0.1 (in development)
 ==============================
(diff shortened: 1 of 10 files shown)


roles/kube_prometheus/hack/fix-dashboard.py
```diff
@@ -66,6 +66,14 @@ def main(dashboard_data, output, filename, title=None):
                 annotation['datasource'] = source_template
                 change = True
 
+    sys.stderr.write("{}\n".format(tags))
```
NicolasT: Huh? Debugging leftover?

roles/kube_prometheus/hack/fix-dashboard.py
```diff
@@ -66,6 +66,14 @@ def main(dashboard_data, output, filename, title=None):
...
+    if tags is not None:
+        # reset initial tags list
+        dashboard['tags'] = []
```
NicolasT: I'd still write the one-line comprehension from #249 but sure, ok :wink:

NicolasT (changes requested):
Think there's a debugging line left?

Also, needs a ChangeLog entry...

roles/kube_prometheus/hack/fix-dashboard.py
alxf: good catch!

roles/kube_prometheus/hack/fix-dashboard.py
alxf: mm i think i miss a commit, i made the one line as you suggested.

NicolasT (approved):
LGTM

NicolasT (changes requested):
Ugh hold on

ChangeLog.rst
```diff
@@ -9,6 +9,8 @@ Features added
 
 :ghpull:`222` - update Elasticsearch and `node_exporter` dashboards
 
+:ghpull:`208` - tag Grafana dashboards (:ghissue:`251`)
```
NicolasT: It's the other way around: issue is #208, PR is #251.

NicolasT (approved):

Issue #246, opened 2018-08-06
is there any particular reason why nginx-ingress installed separately from kubespray-integrated one?


virtuman:
Kubespray already comes with a handful of configuration options as well as ability to customize configmaps as defined here:
https://github.com/kubernetes-incubator/kubespray/blob/f086b6824efc6fbc26dfffa248e7e780e88bf11e/inventory/sample/group_vars/k8s-cluster.yml#L208

Is there any good reason not to use nginx-ingress that comes with kubespray, what are the advantages of having your own not-modifiable installation of nginx-ingress?

NicolasT:
In MetalK8s, we use Kubespray to lay out a base Kubernetes cluster, without any extras not required to have a cluster running (i.e. etcd and CNI are managed by Kubespray, for now). Any other components, including an Ingress Controller, which is not strictly required in a K8s cluster, are managed/deployed by MetalK8s.

We're aware e.g. `nginx-ingress` can currently not be deployed with a specific configuration, however this is work-in-progress (as explained in https://github.com/scality/metalk8s/issues/230#issuecomment-409548748), see #190, #238 and especially for `nginx-ingress`, #180.

virtuman:
So is it safe to assume that cert-manager is also going to be moved into metal-k8s from kubespray?
I thought kubespray was positioning itself as production-grade battle-proof cluster deployment tool that people already trust. 
Is it beneficial for the metal-k8s model to not use that kubespray's trust that's already there and to not only build on top? Kubespray is already extensively customizable that comes with sane defaults.

And if things that are currently being managed by kubespray are being moved out - and I'm sure you have a list of your own "must-haves" - is there a list of components in the work pipeline that the community can see and maybe contribute or suggest other additions / priorities?

If cert-manager will be moved out - it would probably be nice to have it pre-configured with let's encrypt similarly to this guide: https://github.com/ahmetb/gke-letsencrypt . (specifically step number 3 )

Thank you.

NicolasT:
For `cert-manager`, there's #26 (which was started at a time when there was no `cert-manager` deployed by Kubespray, IIRC), but indeed, also falls in the 'not required to run a base cluster' category.

Note, we're not "moving" anything 'into MetalK8s' 'from Kubespray'. As mentioned before, we trust Kubespray to be great and battle-tested to deploy a base cluster, which we then augment.

[...]

Issue #230, opened 2018-08-01
Example of resource customization for elasticsearch deployment?


virtuman:
Is there a way to customize elasticsearch resource assignment such as disk space, ram, cpu, heapsize ? or is it expected to be manually updated after the metal-k8s install script is done running?

I guess I just can't figure out if there's a way to override /files/*.yml files

Thank you.

Zempashi:
We plan to provide customization of helm charts (like providing a value file that override defaults of the chart it-self and default of MetalK8s). But sadly, right now, you need to edit `roles/cluster_logging/files/elasticsearch/value.yml`. Yet, we plan for starting soon this feature.

NicolasT:
'Custom' Helm values support was added for `nginx-ingress` and documented at https://metal-k8s.readthedocs.io/en/latest/usage/advanced_configuration.html (see #291). The related change for Elasticsearch and some other charts we deploy are pending (see e.g. #340, #346).

NicolasT:
The ES changes landed. However, please be aware we may slightly change the 'var' naming before 1.1 is released.

Does this suit your need, @virtuman?

thomasdanan:
won't do, since we are not shipping Elasticsearch anymore in metal2.X

PR #253, opened 2018-08-08
Allow to customize services with custom values.yml files

virtuman:
Suggestion for allowing to customize deployed services related to #230 and #246 

Supported customizable roles (can add all of them, but i added only ones that we already customize as an example, to get this going):
- kube_nginx_ingress
- cluster_logging/elasticsearch
- cluster_logging/elasticsearch-curator
- cluster_logging/kibana

place override `values.yml` file in `{{ inventory_dir}}/customization/*role-path*/values.yml` like this:
`{{ inventory_dir }}/customization/kube_nginx_ingress/values.yml`
`{{ inventory_dir }}/customization/cluster_logging/elasticsearch/values.yml`
`{{ inventory_dir }}/customization/cluster_logging/elasticsearch-curator/values.yml`
`{{ inventory_dir }}/customization/cluster_logging/kibana/values.yml`

Intentionally using `with_fileglob` instead of silent file existence check, so that playbook generates additional output, indicating that this playbook supports overriding values.yml. For example, it will say ` [WARNING]: Unable to find '/Users/.../metal-k8s/inventory/devcluster/customizations/cluster_logging/kibana' in expected paths (use -vvvvv to see paths)z


NicolasT:
@virtuman I'm traveling right now so please allow me some time to get back to this.

In order to get this merged eventually, you'll need to put a signature in your commits and assert your work is in line with the DCO. This is currently not enforced by CI but we're working on that. See https://github.com/scality/metalk8s/pull/254/commits/90a90b650fa321097c24c3fdd8b5bc62c1a560b0

Please amend the commit message accordingly, and state your contribution is in line with the DCO in a comment of this PR (for now) if possible :)

virtuman:
Sure thing. Just amended commit message. Mind you - I never used ansible before, and I totally understand if the suggested approach that I used here is completely invalid. This is something that currently works for me for deploying my clusters.

NicolasT:
@virtuman Now #238 landed, do you think it'd make sense to use that approach instead of using files in an inventory directly?

Using the new `helm_cli` module, we can update all roles to use this module, and get 'custom' Chart settings from the inventory. This could be coded up in some `vars` file(s) in-between other Ansible settings, or using a separate file (as in your approach), and use the Ansible `lookup` feature to then include said file in the inventory as value for the appropriate `var`.

@Zempashi Do you think the above makes sense?

Zempashi:
@NicolasT yes I do, I started a new PR (#253) to draft some code for what we envisioned

Issue #255, opened 2018-08-08
Document how to emit an Ansible playbook execution logfile


NicolasT:
- [ ] Document how to let Ansible keep a logfile of playbook execution (also linking to relevant Ansible docs)
- [x] Since this is based on settings in `ansible.cfg`, ship an example `ansible.cfg` which has the required settings commented-out for easy setup

See: RFE-152

PR #271, opened 2018-08-14
docs: Document Ansible logging configuration

NicolasT:
Fixes: #255
See: https://github.com/scality/metalk8s/issues/255
See: RFE-152

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+# We have `ansible.cfg` in VCS, but because one may make local changes, we don't
+# necessarily want 'your' version checked in. When making changes to
+# `ansible.cfg` which are meant to be pushed in the repository, make sure you
+# make those changes on a 'clean' `ansible.cfg` and use `git add -f`.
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,10 @@
+# We have `ansible.cfg` in VCS, but because one may make local changes, we don't
```
Zempashi, line 1: Or part of a make shell, copy over an ansible.cfg.sample. Only the sample is checked in git and allow users to modify the ansible.cfg at will (or loose it if making another git clone)

Zempashi (dismissed):

.gitignore
NicolasT: Yeah, that's a sensible approach (though I wouldn't let `make shell` create a copy, that's a bit IMHO).

wabernat (dismissed):
Approved for language, with edits.

NicolasT:
Needs new branch name.

PR #280, opened 2018-08-21
 Add an Ansible module to handle Helm chart installation 

Zempashi:
Fixes: #190

### DIFF
--- a/roles/helm_common/library/helm_cli.py
+++ b/roles/helm_common/library/helm_cli.py
@@ -1,0 +1,232 @@
+import os
+import json
+
+from contextlib import contextmanager
+from tempfile import NamedTemporaryFile
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class Helm(AnsibleModule):
+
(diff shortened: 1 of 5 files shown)


roles/helm_common/library/helm_cli.py
```diff
@@ -0,0 +1,232 @@
...
+            raise HelmError('error running helm {} command: {}'.format(
+                ' '.join(cmd), exc))
+        else:
+            if failed_when(rc, out, err):
+                raise HelmError(
+                    'Error running helm {} command (rc={}) '
+                    "out='{}' err='{}'".format(' '.join(cmd), rc, out, err))
+        return rc, out, err
+
+    def check_present(self):
```
Zempashi: @NicolasT : Still confused `check_present` runs `install`.
@Zempashi: `ensure_present` is a better name ?

NicolasT: Yes!

NicolasT (approved):

PR #285, opened 2018-08-22
Imported MetalK8s install steps taken from Zenko

wabernat:
Also moved upgrade instructions into a separate file. I recommend changing the directory title from "usage" to "installation_and_upgrade" (or similar).
I also recommend another document to define explicitly the options available for configuration in the kube-node.yml file.

Pardon the volume of changes: no avoiding it.

### DIFF
--- a/docs/usage/upgrade_from_pre-0.2.0.rst
+++ b/docs/usage/upgrade_from_pre-0.2.0.rst
@@ -1,0 +1,71 @@
+Upgrading from MetalK8s < 0.2.0
+===============================
+
+MetalK8s 0.2.0 introduced changes to persistent storage provisioning that are
+not backwards-compatible with MetalK8s 0.1. These changes include:
+
+- The default LVM VG was renamed from `kubevg` to `vg_metalk8s`.
+- Only the PersistentVolumes required by MetalK8s services are created by
+  default.
+- Instead of using dictionaries to configure the storage, these are now
+  flattened.
+
+When a MetalK8s 0.1 configuration is detected, the playbook reports an error.
+
+If you have an early configuration like:
+
+  .. code-block:: yaml
+
+    metal_k8s_lvm:
+      vgs:
+        kubevg:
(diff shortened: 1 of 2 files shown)


NicolasT:
I don't see a second file for upgrade?

We'll also want to get this into 1.0, that'll need a bit of Git magic.

wabernat:
Before we go for Git magic, I'll settle for elementary competence. Apparently, I missed a git add for the new file (docs/usage/upgrade_from_pre-0.2.0.rst). Coulda swore...

I've re-pushed. I hope this comes through better this time. Sorry for the confusion.

NicolasT:
What I meant is we want these changes to go in `development/1.0` (and then percolate into `development/1.1` through @bert-e), which will require some thoughtful rebasing.

docs/usage/upgrade_from_pre-0.2.0.rst
```diff
@@ -0,0 +1,71 @@
+Upgrading from MetalK8s < 0.2.0
```
NicolasT, line 1: I'd like to see this removed from the documentation altogether. This 'upgrade' was never supported in the first place, new deployments shouldn't be using 0.2.0, and any existing 0.2.0 deployments have *or* been upgraded, *or* the owner can figure out how to upgrade by reading old versions of the documentation.

wabernat: I'm not sentimental about this, but before dismissing it, we should ask:

How many pre-0.2.0 downloads are out there? 

Is the removal of these instructions going to result in service calls from customers? 

Will we be perceived as having a lower quality product if this issue is undocumented? 

Resolution is above my pay grade.

--W

wabernat: Also, to clear the air, I knew what you meant. My "basic competence" comment was pointed straight back at me.

wabernat:
Abandon this PR. Superseded by PR 308.

PR #292, opened 2018-08-27
Remove hard dependency to python3.6

Zempashi:
It works fine one python3.7, I would like  to loosen the python dependency without setting up envlist. (If someone has another idea)

### DIFF
--- a/eve/tests/single-node/init.sh
+++ b/eve/tests/single-node/init.sh
@@ -15,5 +15,5 @@ iptables -F || die
 iptables -X || die
 
 yum install -y epel-release
-yum install -y python2-pip python2-devel python36 python36-devel
-pip2 install tox
+yum install -y python36 python36-devel
+pip3 install tox
--- a/tox.ini
+++ b/tox.ini
@@ -60,7 +60,6 @@ whitelist_externals =
     sh
 
 [testenv:tests]
-basepython = python3.6
 skip_install = true
 deps =
     -r{toxinidir}/tests/requirements.txt


NicolasT (dismissed):
Fair enough. When someone breaks something for some version of Python, we'll learn soon enough.

NicolasT:
Interesting that build failed... So we had `python3.6` but there's no `python3` on the host? Odd!

Zempashi:
Seems that effectively centos7 miss a python3 interpreter (repeatable on docker). I worked around on my PC with "ln -s /usr/bin/python3.7 /usr/local/bin/python3.6"... The other option is to create the symlink at cloudinit time.

ballot-scality:
https://www.softwarecollections.org/en/scls/?search=python&policy=&repo=&order_by=-create_date&per_page=10

Seems python 3.7 is not yet part of software collections. As for the fact that "python3" is not present, I'm not really surprised, as rpm doesn't handle very well the "alternatives" that the debian does.

NicolasT:
> As for the fact that "python3" is not present, I'm not really surprised, as rpm doesn't handle very well the "alternatives" that the debian does.

On my F28, `/usr/bin/python3` is simply provided by `python3-3.6.6-1.fc28.x86_64`, as expected...

Zempashi:
Suggestion to remove basepython at all. So some distribution will run python3 other run python2 (likely 2.7 nowadays, so "bearable"). Does it make sense to anyone ?

ballot-scality (dismissed):

eve/workers/centos7_cloudinit/install_python3.sh
```diff
@@ -0,0 +1 @@
+yum install -y python34
```
NicolasT: Guess it's not strictly required with the image we seem to be using, but this would require installing `epel-release` first on a bare image.

Zempashi:
https://github.com/ansible/ansible/blob/devel/setup.py#L243
Sadly ansible is not compatible with python3.4 anymore...
So removing basepython ?

NicolasT:
OK on removing `basepython`. Doesn't harm to try to maintain compatibility with Python2.

eve/tests/single-node/init.sh
```diff
@@ -15,5 +15,5 @@ iptables -F || die
 iptables -X || die
 
 yum install -y epel-release
-yum install -y python2-pip python2-devel python36 python36-devel
-pip2 install tox
+yum install -y python36 python36-devel
```
NicolasT, line 18: Since `basepython` is removed, `python` will be used which is Python 2. Any reason to then install `python36`, `python36-devel` and `pip3 install tox` instead of keeping the original `python2`-based mechanism?

Zempashi: because then it's run on python2, which I don't like...(I make a difference between trying to support python2 and bailing out if we don't support python2 because the CI use it). But anyway python36 doesn't come with pip3.6 neither the pip module (may be a bug on centos package). There is `ensurepip` but I'm not huge fan. Wondering if it's better to keep as now ;)

EDIT: or just simlinking python3.6 to /usr/local/bin/python3 into init.sh

PR #297, opened 2018-08-28
tests: Use Kubernetes Python API to interact with cluster

Zempashi:


tests/utils/helper.py
```diff
@@ -38,3 +50,40 @@ def __init__(self, filename):
 
     def __getattr__(self, key):
         return getattr(self._inventory, key)
+
+
+def _find_resource_name(resource):
+    for resource_name, api in NAMESPACED_RESOURCES.items():
+        if resource == resource_name or resource == resource_name.strip('_'):
```
ballot-scality: For my own curiosity, why the strip('_') ?

Zempashi: To allow 'daemonset' and 'daemon_set' (and eventually 'DaemonSet' if someone ask for)

NicolasT: Is there a particular reason we have to support both? :wink:

Zempashi: Pure cosmetic for feature file. The "daemon_set" version is not what we are used for (because of kubectl that strip underscore)

NicolasT: I'm confused. `str.strip` only strips at the front and end of a string, so how would this work?

Zempashi: hmm indeed a bug. It's not yet used in this PR (this on use `persistent_volume` which looks like better than `persistentvolume` (will be in the following).
```
resource.lower() = resource_name.replace('_', '')
```
Remove the ability for `daemon_set` (but introduce  `DaemonSet`). Does it work for both of you

NicolasT: :+1:

tests/conftest.py
```diff
@@ -53,3 +58,20 @@ def ansible_no_error(request):
...
+        kubectl_command,
+        env={'KUBECONFIG': kubeconfig},
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    if kubectl_run.returncode == 0:
+        request.kubectl_result = kubectl_run
+        return kubectl_run
+    else:
+        logging.error("kubectl stdout: {}\n stderr: {}".format(
```
NicolasT: Shouldn't this raise instead of implicitly returning `None`?

tests/utils/helper.py
ballot-scality: yes

ballot-scality (dismissed):

tests/conftest.py
Zempashi: But some test might require to test error of kubectl (delete a locked pv, or whatever). But indeed this in not ideal (mainly because I don't return the kubectl_run in case of error.)

tests/utils/helper.py
```diff
@@ -38,3 +50,40 @@ def __init__(self, filename):
...
+            return resource_name, api, True  # namespaced_resource
+    for resource_name, api in CLUSTER_RESOURCES.items():
+        if resource == resource_name or resource == resource_name.strip('_'):
+            return resource_name, api, False  # cluster_resource
+    else:
+        raise ValueError('Unknown resource {} should be part of {}'.format(
+            resource, NAMESPACED_RESOURCES + CLUSTER_RESOURCES))
```
Zempashi:
```
NAMESPACED_RESOURCES.keys() | CLUSTER_RESOURCES.keys()
```

Zempashi:
@bert-e reset

NicolasT (dismissed):
Will need to resolve merge, but LGTM

NicolasT (dismissed):

Zempashi:
@bert-e reset

NicolasT (dismissed):

Zempashi:
@bert-e reset

NicolasT (approved):

Issue #260, opened 2018-08-13
Design 'Offline install'


NicolasT:
See: #259

NicolasT:
Won't do.

Issue #259, opened 2018-08-13
Allow MetalK8s to be installed from scratch or upgraded in a fully-offline environment


NicolasT:
A user should be able to download a release package (some binary blob, format to be defined), store this on e.g. a USB key, then take it to a set of servers fully disconnected ('air-gapped') from the public internet (no proxies available), mount the USB key and launch a procedure which eventually ends with a fully running or upgraded MetalK8s cluster.

This projects includes
- Design the technical implementation of this feature and test unknowns
- Automate the build of such release package
- Extend and update deployment code to be able to use this package
- Automatically build the release package in CI
- Document the various procedures
- Test the various procedures
- Run all installation test cases using offline installation, in an environment which is actually offline (to rule out mistakes)

NicolasT:
This was about 1.x, we do this in 2.x and won't support this in 1.x moving forward.

PR #302, opened 2018-08-29
In-cluster package and image repository

NicolasT:
This is a first PR to include a `yum` package repository, Docker image registry, Helm chart repository,... inside the Kubernetes cluster to bootstrap installation of all packages.

These repositories are populated from the internet, so this is definitely no full 'offline installation' solution, rather a first step.

See: #259 
See: #260


NicolasT:
@bert-e wait

NicolasT:
Current open question:

- [x] The repository host(s) need to be added to `--insecure-registry` options passed to Docker on all nodes. Not sure how to achieve this properly...
- [ ] Currently I only tested this on an existing cluster, using one of my `kube-master` nodes as `metalk8s-repository`. This works, but the `kubernetes/node` role basically doesn't do anything. This also pulls in `kubernetes/secrets`. This may not work on 'clean' servers...
- [ ] Currently have to add `vendor/kubespray/roles` under `defaults.roles_path` in `ansible.cfg` for the roles to be properly discovered.
- [x] The Docker repository is 'publicly' running on the host network of the repository nodes, without authentication or whatnot :-1:

NicolasT:
We won't do this in 1.x after all.

PR #310, opened 2018-09-04
Cleanup for test suite. New retry helper

Zempashi:


Zempashi:
@bert-e reset

tests/post/test_storage.py
```diff
@@ -91,8 +91,8 @@ def lauch_test_storage_pod(kubeconfig):
 
 @then("The result of test storage pod should be 'success'")
 def storage_pod_test_result(kubeconfig):
-    count = 5
-    while count:
+    try_ = retry(5)
+    while next(try_):
```
NicolasT, line 95: Because `RetryCountExceededError` is a `StopIteration`, will this actually raise on timeout, or will the loop just exit and the program continue to run?

Zempashi: No it will raise. The non raise version of the retry is
```
for _ in retry(5):
     break
else:
     # After the last retry, custom code without error

NicolasT: Interesting. I had no clue how `while` interacts with a generator, never considered the scenario. Thanks!

NicolasT (approved):

tests/post/test_storage.py
Zempashi: here while don't interact with the generator that's the `next` (and the generator return true for `while` to be happy)

Zempashi: hmm and the `for` version (not used) doesn't work anymore on python3.7 (but works on python2)

EDIT:   https://www.python.org/dev/peps/pep-0479/

PR #335, opened 2018-09-07
Documentation PDF improvements

NicolasT:
- Some styling tweaks
- Rework the `Dockerfile` to include what's needed to build a PDF
- Add PDF build step to CI

The `Dockerfile` changes make the image grow a lot (roughly 1.6GB...), but this makes building a PDF version of the docs significantly easier compared to somehow gathering all required LaTeX packages on a host system.

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -3,26 +3,35 @@
 # Note: the MetalK8s repository root is supposed to be mounted at
 # `/usr/src/metalk8s` when running the container
 
-FROM docker.io/alpine:3.8
+FROM docker.io/ubuntu:18.04
+
+RUN export DEBIAN_FRONTEND=noninteractive && \
+    apt-get update && \
+    apt-get install --no-install-recommends -y \
+        enchant \
+        git \
+        latexmk \
+        make \
+        python2.7 \
+        python3-buildbot-worker \
+        texlive-fonts-extra \
+        texlive-fonts-recommended \
+        texlive-latex-extra \
+        texlive-latex-recommended \
+        tox \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 WORKDIR /usr/src/metalk8s
 
 RUN mkdir docs
 
-RUN apk --no-cache add \
-        enchant \
-        git \
-        make \
-        py-twisted \
-        py2-pip \
-        && \
-    pip install \
-        tox \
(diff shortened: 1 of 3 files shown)


docs/Dockerfile
```diff
@@ -3,26 +3,35 @@
...
         enchant \
         git \
+        latexmk \
         make \
-        py-twisted \
-        py2-pip \
-        && \
-    pip install \
+        python2.7 \
+        python3-buildbot-worker \
```
Zempashi, line 16: Should check with our releng team: but with our version of buildbot(-master), python2 version of buildbot-worker is recommended

NicolasT: @tcarmet Any take on this? It *seems* to work...

NicolasT: As in, there's no Python 2 version packaged in Ubuntu, and I'd rather use a package than random `pip install` calls.

tcarmet: Indeed, it is recommended to use a python2 version of the buildbot worker as we're using python2 with eve.
But I don't think we're gonna face any issue. I've used a few times a python3 worker with eve and it worked perfectly. What I would recommend is to take a look at the version of this worker and ensure it's not higher that `0.9.12`. If it is, then, be careful.

eve/main.yml
```diff
@@ -68,4 +68,4 @@ stages:
     - Git: *git_pull
     - ShellCommand:
         name: 'Build doc'
-        command: tox -e docs
+        command: tox --workdir /tmp/tox -e docs -- html latexpdf
```
tcarmet, line 71: Wouldn't it be also interesting to add an `Upload` step so that you can retrieve the pdf/html documentation in case you want to see the rendering?

docs/Dockerfile
NicolasT: It's 1.1 something.

eve/main.yml
NicolasT: Over time, maybe, however, right now docs are rendered and served on ReadTheDocs already. Figuring out the whole docs workflow is under discussion, also related to our other projects.

docs/Dockerfile
tcarmet: we need to be careful if we get an upgrade that doesn't work with eve anymore. It will be quite explicit if we get an error because of this, so LGTM

Zempashi (approved):

Issue #92, opened 2018-06-26
Implement an upgrade test in CI


NicolasT:
There are at least two scenarios to test:

- Upgrade from some 'old' baseline version (e.g. 0.1.0 and stick to it) to proposed HEAD
- Upgrade from the PR target version/branch to proposed HEAD

NicolasT:
Done in #344.

PR #91, opened 2018-06-26
Test upgrade

NicolasT:

PR #344, opened 2018-09-11
tests: Upgrade test scenario

Zempashi:
Fixes: #92
Closes: #91

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ ansible.cfg
 /.tox/
 /.kube/
 /.shell-env/
+/.pytest_cache/
 
 ansible-playbook.log
--- a/tests/install/test_upgrade.py
+++ b/tests/install/test_upgrade.py
@@ -1,0 +1,36 @@
+import os
+
+import pytest
+
+from pytest_bdd import parsers
+from pytest_bdd import scenario
+from pytest_bdd import when
+
+from utils.helper import create_version_archive
+from utils.helper import run_ansible_playbook
+
+
+@pytest.fixture
+def pytestbdd_strict_gherkin():
+    return False
+
+
+@scenario('features/upgrade.feature', 'Upgrade from 1.0.0')
+def test_upgrade(inventory_tagging):
+    pass
+
+
+@when(parsers.parse(
+    "I launch ansible playbook '{playbook}' from version {version}"
+))
+def ansible_playbook_from_version(
+        request, inventory, playbook, archive_dir, version):
+    archive_version = create_version_archive(version, archive_dir)
+    inventory_path = os.path.abspath(inventory)
+    ansible_process = run_ansible_playbook(
+        playbook,
+        basedir=archive_version,
+        env={'ANSIBLE_INVENTORY': inventory_path}
+    )
+    request.ansible_process = ansible_process
+    return ansible_process
--- a/tests/post/test_kubectl.py
+++ b/tests/post/test_kubectl.py
@@ -12,6 +12,16 @@ def get_node_list(request, inventory_obj):
     if request.kubectl_result.returncode != 0:
         pytest.fail('Error running kubectl command')
     lines = request.kubectl_result.stdout.read().splitlines()
+    # Parsing stdout of a make shell C='kubectl ...' is a bit tricky:
+    # First there is chance that virtualenv is not ready so that
+    # stdout get poluted by "creating virtualenv...", "dowloading kubectl"
+    # => Start parsing after the header line starting with 'NAME' in this case
+    for n, line in enumerate(lines, start=1):
+        if line.startswith('NAME'.encode('utf-8')):
+            break
+    else:
+        raise AssertionError("Kubectl header line cannot be found")
+
     nodes_nb = len(set(inventory_obj.get_groups_dict()['kube-master'] +
                        inventory_obj.get_groups_dict()['kube-node']))
-    assert nodes_nb == len(lines) - 1  # Start with column name line
+    assert nodes_nb == len(lines) - n  # Start after Header line
--- a/tests/utils/helper.py
+++ b/tests/utils/helper.py
@@ -7,8 +7,10 @@ from ansible.inventory.manager import InventoryManager
 from ansible.parsing.dataloader import DataLoader
 
 
-def run_make_shell(args=None, tmpdir=None, **kwargs):
+def run_make_shell(args=None, basedir=None, tmpdir=None, **kwargs):
     make_args = []
+    if basedir:
+        make_args.extend(('-C', basedir))
     if tmpdir:
         make_args.append('SHELL_ENV={}'.format(tmpdir))
     if args:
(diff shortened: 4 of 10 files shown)


Zempashi:
@bert-e wait

Zempashi:
@bert-e reset

hack/shell-bashrc
```diff
@@ -1,4 +1,7 @@
 # Only load `~/.bashrc` in interactive shells
+if [[ ! -z "$TARGET_DIR" ]]; then
```
Zempashi: @NicolasT: I did this at first try to solve the CI failing. I didn't remove this because I found convenient to not change ANSIBLE_INVENTORY (without, we need to patch the ANSIBLE_INVENTORY which might be a relative path evaluated from pwd (that change running under a make shell for an extracted archive)).

This requires to patch the 1.0.0 version (shutil.copy in create_version_archive), which is a bit ugly...
What do you think about that

.gitignore
```diff
@@ -10,5 +10,6 @@ ansible.cfg
 /.tox/
 /.kube/
 /.shell-env/
+/.pytest_cache/
```
NicolasT, line 13: Should also go in `.dockerignore` then.

hack/shell-bashrc
NicolasT: First of all, needs to be moved down (the comment above applies to the section below).
I don't really like this to be part of a shell rcfile... Must be a better way to achieve this... Not sure how though right now.

tests/install/test_upgrade.py
```diff
@@ -0,0 +1,35 @@
...
+    pass
+
+
+@when(parsers.parse(
+    "I launch ansible playbook '{playbook}' from version {version}"
+))
+def ansible_playbook_from_version(
+        request, inventory, playbook, archive_dir, version):
+    archive_version = create_version_archive(version, archive_dir)
+    inventory_path = os.path.abspath(inventory)
```
NicolasT, line 29: This is not used? However, shouldn't it be properly exported in the env?

tests/post/test_kubectl.py
```diff
@@ -12,6 +12,12 @@ def get_node_list(request, inventory_obj):
     if request.kubectl_result.returncode != 0:
         pytest.fail('Error running kubectl command')
     lines = request.kubectl_result.stdout.read().splitlines()
+    for n, line in enumerate(lines, start=1):
```
NicolasT: I don't understand why these changes are required. Is the `kubectl` output ever 'different'?

Also, instead of this, how about changing the test to run `kubectl get nodes -o json`, and `json.loads` the output, then work on the resulting object?

tests/utils/helper.py
```diff
@@ -37,13 +40,20 @@ def run_make_shell(args=None, tmpdir=None, **kwargs):
...
-    return run_make_shell(args=command, env=env_vars, tmpdir=tmpdir)
+    if basedir:
+        env_vars['TARGET_DIR'] = os.getcwd()
+        playbook_path = os.path.join(basedir, 'playbooks', playbook)
+    else:
+        playbook_path = os.path.join('playbooks', playbook)
+    command = "ansible-playbook {}".format(playbook_path)
+    kwargs = dict(basedir=basedir, tmpdir=tmpdir)
+    kwargs.update(args=command, env=env_vars)
+    return run_make_shell(**kwargs)
```
NicolasT:
```python
return run_make_shell(args=command, env=env_vars, basedir=basedir, tmpdir=tmpdir)
```

tests/utils/helper.py
```diff
@@ -65,3 +75,15 @@ def retry(count, wait=2, msg=None):
...
     raise RetryCountExceededError(msg)
+
+
+def create_version_archive(version, tmpdir):
+    git_top = subprocess.check_output(('git', 'rev-parse', '--show-toplevel'))
+    git_top = git_top.strip()
+    git_command = ('git', 'archive', '--prefix={}/'.format(version), version)
+    git = subprocess.Popen(git_command, stdout=subprocess.PIPE, cwd=git_top)
+    subprocess.check_call(('tar', '-C', str(tmpdir), '-x'), stdin=git.stdout)
+    git.wait()
```
NicolasT, line 69: Wait, and ensure it exits successfully.

tests/utils/helper.py
```diff
@@ -65,3 +75,15 @@ def retry(count, wait=2, msg=None):
...
+    shutil.copy('hack/shell-bashrc', os.path.join(str(tmpdir.join(version)), 'hack'))
+    assert git.returncode == 0
```
NicolasT: Ah, OK. Move this higher up.

NicolasT (changes requested):

tests/install/test_upgrade.py
Zempashi: This was the other way around for fixing the CI: converting the ANSIBLE_INVENTORY variable into an absolute path when needed.

(but indeed need to choose one solution, not both)

tests/post/test_kubectl.py
Zempashi: Yes because we can have the virtualenv creation (and downloading of kubectl and helm) that does print to stdout (And it's the case with the upgrade test because installation and upgrade where done in git archive temporary dir)

tests/install/test_upgrade.py
Zempashi: I kept the version of ANSIBLE_INVENTORY exported as an absolute path

Zempashi:
@bert-e reset

.gitignore
Zempashi: done

tests/install/test_upgrade.py
NicolasT: I prefer the absolute path based approach. Remember how there was a 'demo effect' hickup during the SF demo of MetalK8s many months ago, because I used a relative path for `$KUBECONFIG` and moved into another directory? :wink:

tests/post/test_kubectl.py
```diff
@@ -12,6 +12,16 @@ def get_node_list(request, inventory_obj):
...
     lines = request.kubectl_result.stdout.read().splitlines()
+    # Parsing stdout of a make shell C='kubectl ...' is a bit tricky:
+    # First there is chance that virtualenv is not ready so that
+    # stdout get poluted by "creating virtualenv...", "dowloading kubectl"
+    # => Start parsing after the header line starting with 'NAME' in this case
+    for n, line in enumerate(lines, start=1):
+        if line.startswith('NAME'.encode('utf-8')):
+            break
+    else:
+        raise AssertionError("Kubectl header line cannot ba found")
```
NicolasT: Pedantic: `be`.

tests/utils/helper.py
```diff
@@ -23,13 +25,15 @@ def run_make_shell(args=None, tmpdir=None, **kwargs):
...
 
-def run_ansible_playbook(playbook, env=None, tmpdir=None):
+def run_ansible_playbook(playbook, env=None, basedir=None, tmpdir=None):
     env_vars = dict(os.environ)
     env_vars.setdefault('ANSIBLE_FORCE_COLOR', "true")
     if env:
         env_vars.update(env)
     command = "ansible-playbook playbooks/{}".format(playbook)
-    return run_make_shell(args=command, env=env_vars, tmpdir=tmpdir)
+    kwargs = dict(basedir=basedir, tmpdir=tmpdir)
```
NicolasT: I'm not too fond of creating a `dict` by passing two `kwargs` to the `dict` constructor, then updating two (different) keys using `update` in the second line (also using the not-so-common `kwargs`-based approach), then passing that dict as `**kwargs` to some function.

I really think something like

```python
return run_make_shell(
    args=command,
    env=env_vars,
    basedir=basedir,
    tmpdir=tmpdir,
)
```

is much more readable. Though granted, the current code is not 'wrong' per se.

Zempashi:
@bert-e reset

NicolasT (approved):

Issue #340, opened 2018-09-10
Add Elasticsearch external values capacity


alxf:
Give the opportunity to the user to override the values.yaml through the external_values variables.

PR #346, opened 2018-09-11
Add Elasticsearch external values capability

alxf:
fix: #340


Zempashi:
The CI is failing due to a flakiness that is normally solved. Rebase should fix the issue

alxf:
@NicolasT @Zempashi, one question remain, do i need to adapt the variable names for extern_values as i did for elasticsearch, or should i keep this particular case ? (i mean in the yaml values which will be injected)

Zempashi (approved):

PR #347, opened 2018-09-11
Check pep8 syntax before consuming CPU time

Zempashi:


tests/utils/kube.py
```diff
@@ -29,24 +29,24 @@ def _find_resource_name(resource):
 
 def get_kube_resources(kubeconfig, resource, namespace='default', name=None):
     k8s_config.load_kube_config(config_file=kubeconfig)
-    resource_name, api_name, namespaced_resource = _find_resource_name(resource)
+    resrc_name, api_name, namespaced_resource = _find_resource_name(resource)
```
NicolasT: A somewhat more 'standard' shorthand for `resource` is `res` rather than `resrc`.

eve/main.yml
```diff
@@ -34,6 +35,11 @@ stages:
         method: clobber
         retryFetch: true
         haltOnFailure: true
+    - ShellCommand:
+        name: 'Check syntax'
+        haltOnFailure: true
+        command: |-
+          tox -e pep8
```
NicolasT: Could also be a Docker container with all deps pre-installed, even using Alpine.

Zempashi: Yes was doing that ;)

eve/tests/single-node/init.sh
```diff
@@ -13,7 +13,3 @@ if systemctl is-enabled --quiet iptables; then
 fi
 iptables -F || die
 iptables -X || die
-
```
Zempashi, line 16: With kubernetes worker this is not strictly needed anymore. But I find it better like that, so I let that part

Zempashi:
@bert-e reset

Zempashi:
Here, would like a tacit approval for this PR. The build on 1.1 is failling because of other pep8 issue (specific in 1.1), but If need to recreate the integration branch, I will have to re-apply the fix and change the branch each time. Would like to fix that only at the end.

eve/main.yml
```diff
@@ -69,3 +70,18 @@ stages:
...
+      path: eve/workers/pep8.yaml
+      images:
+        pep8-checker:
+          context: '.'
+          dockerfile: eve/images/pep8-checker/Dockerfile
+    steps:
+    - Git: *git_pull
+    - ShellCommand:
+        name: 'Check syntax'
+        haltOnFailure: true
```
NicolasT, line 86: Would this be required on the doc stage as well?

eve/workers/pep8.yaml
```diff
@@ -1,17 +1,23 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: "zenko-test-pod"
+  name: "metal-k8s-pep8"
```
NicolasT: `metalk8s`

NicolasT (dismissed):

eve/main.yml
Zempashi: May be. I'll open a PR for this change

Zempashi:
@bert-e reset

Zempashi:
@bert-e reset

Zempashi:
@bert-e force_reset

NicolasT (dismissed):

Zempashi:
@bert-e wait

Zempashi:
@bert-e force_reset

NicolasT:
Does this now include #360?

NicolasT (approved):
:crossed_fingers:

PR #348, opened 2018-09-12
Improvement/storage check action plugin

ballot-scality:
This is the same as #232 but with the right branch name.


ballot-scality:
@bert-e reset

Zempashi (approved):

PR #357, opened 2018-09-12
docs: Windows compatibility fixes

NicolasT:
- Add `colorama` to docs requirements: this is required by Sphinx on Windows, but `pip-compile` doesn't take platform markers into account, so building `requirements.txt` on Linux generates a file which doesn't include all dependencies required on a Windows host.
- Add a kludge to support building on Windows and Linux
- Support `livehtml` on Windows

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,9 +34,10 @@ deps =
     -r{toxinidir}/docs/requirements.txt
 commands =
     doc8 docs/
-    make -C docs {posargs:html} SPHINXOPTS="-j 4 -n -W"
-whitelist_externals =
-    make
+    {toxinidir}/docs/build.cmd {posargs:html}
+setenv =
+    O=-j4 -n -W
+    SPHINXOPTS=-j4 -n -W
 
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -34,9 +34,10 @@ deps =
     -r{toxinidir}/docs/requirements.txt
 commands =
     doc8 docs/
-    make -C docs {posargs:html} SPHINXOPTS="-j 4 -n -W"
-whitelist_externals =
-    make
+    {toxinidir}/docs/build.cmd {posargs:html}
+setenv =
+    O=-j4 -n -W
```
Zempashi, line 39: For my personnal knowledge: What O stands for ?

Zempashi (approved):

tox.ini
NicolasT: 'options', I guess. It's the same as SPHINXOPTS, just that Makefile and make.bat have different 'preferences' *sigh*

Issue #367, opened 2018-09-17
Do not deactivate SSH root login when the remote user used by ansible is root


ballot-scality:
Ansible-hardening is deactivating the root login by setting
```
PermitRootLogin no
```
However, if the root user is the only user available for SSH, you remove your access to the server.

Adding a check in preflight check with the used user and if root, modify the variable as to no deny complete access would be a nice safety net.

NicolasT:
So, there's #337 which asserts that the remote user is not root. This is, however, currently not included in any released version.

If the solution introduced in #337 is acceptable, I'd prefer to keep it as-is rather than introduce a condition in the playbook and act one way or another depending on it.

Finally, if people feel very strongly against including `ansible-hardening`, we could remove it altogether...

thomasdanan:
not relevant in metalk8s 2.X and won't fix it in metalk8s 1.X, closing the issue

Issue #329, opened 2018-09-05
If setup fails after ansible-hardening (STIG) has been run, it cannot be run as root again


mastachand:
Because the ansible-hardening requirements change some settings in sshd, notably PermitRootLogin to No, it can cause some issues if the install is run with ansible user root.

We should recommend and document how to run the installer with non-root user if we are going to say that running ansible-hardening is the default. Alternatively, we should also document how to run the installer without enabling STIG.

NicolasT:
Agree. We should document this, and add a validation to `preflight-checks` to ensure the user is not using `root` as `ansible_user`.

Issue #128, opened 2018-07-06
Validate inter-node connectivity as a prerequisite


NicolasT:
Early on in the `deploy` (and related) playbooks, where we check all nodes are reachable from the deployment host (`ping`), we should figure out a way to validate all nodes can talk to each other as well, e.g. to make sure no overly restrictive OpenStack *security group* is in place.

NicolasT:
This is partially handled  by #368, but doing TCP-level checks would be useful as well.

thomasdanan:
@gdemonet @NicolasT shouldn't we link this one to the pre check epic?

gdemonet:
Connectivity checks could include verification of configured MTUs (see #1095)

PR #337, opened 2018-09-07
preflight-checks: Disallow `root` `ansible_user`

NicolasT:
Fixes: #329
See: https://github.com/scality/metalk8s/issues/329

PR #368, opened 2018-09-17
node-access-checks: Add role

NicolasT:
This adds a new role which performs various checks before starting
deployment:

- Assert the deployment host can access all `k8s-cluster` and `etcd`
  nodes
- Assert the SSH user is not `root`, or at least to a certain extent
- Assert all nodes can ICMP `ping` all other nodes

See: #367
See: https://github.com/scality/metalk8s/issues/367
See: #329
See: https://github.com/scality/metalk8s/issues/329
See: #128
See: https://github.com/scality/metalk8s/issues/128

### DIFF
--- a/roles/node-access-checks/tasks/main.yml
+++ b/roles/node-access-checks/tasks/main.yml
@@ -1,0 +1,64 @@
+---
+- name: Ping nodes from deployment host
+  ping:
+
+- name: Ping nodes from deployment host, and 'become'
+  ping:
+  become: true
+
+- name: Assert remote SSH user is not 'root'
+  block:
+    - name: Assert 'ansible_become_method' is 'sudo'
+      block:
+        - name: Lookup 'become' '$SUDO_COMMAND'
+          shell: 'echo -n "$SUDO_COMMAND"'
+          become: true
+          register: env_become_sudo_command
+          changed_when: false
+        - name: Assert 'ansible_become_method' is 'sudo'
+          assert:
+            that:
+              - 'env_become_sudo_command.stdout != ""'
+            msg: Using non-'sudo' 'ansible_become_method'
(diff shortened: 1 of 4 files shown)


NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e status

NicolasT:
@bert-e reset

NicolasT:
@bert-e status

roles/node-access-checks/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: Ping nodes from deployment host
+  ping:
+
+- name: Ping nodes from deployment host, and 'become'
+  ping:
+  become: true
+
+- name: Assert remote SSH user is not 'root'
```
Zempashi, line 9: I'm not getting what's the purpose of all of this. For me #337 did it right

NicolasT: Not necessarily: I didn't test, but

- When running `ansible-playbook` as `root` and `ansible_user` isn't set as a var on a host in the inventory, `ansible_user` wouldn't be set and you'd nonetheless be SSH'ing as `root`
- When having some fancy configuration in `~/.ssh/config` to specify a specific user for a specific host

so I thought we're better off adding some actual dynamic checks as well to be (even more) safe.

Zempashi: I think having another solution:
```
- setup:
  become: False
- assert:
       that:
            - "ansible_user_id != 'root'"
- setup: # Restablish default setup with info that only root might be able to get

NicolasT:
Won't do.

PR #372, opened 2018-09-17
Add 'external values' support to kube_heapster and kube_metrics_server

NicolasT:


roles/kube_metrics_server/defaults/main.yml
```diff
@@ -6,6 +7,6 @@ metrics_server_namespace: kube-system
 metrics_server_release_name: metrics-server
 metrics_server_chart: metrics-server
 
-kube_metrics_server_uninstall: false
+metrics_server_external_values: []
```
Zempashi, line 10: Should we name it 'kube_metrics_server__external_values' ?

NicolasT: Ideally, yes. However, not sure to which extent we want to use the 'new' naming scheme next to the 'old' scheme. Maybe better off renaming everything at once?

Zempashi: Definelty not for a huge PR that rename all the variable. But yes we could change do a PR for renaming all extrernal_values, and other PR for other use-cases

Zempashi (dismissed):

roles/kube_metrics_server/defaults/main.yml
NicolasT: Then the one for `external_values` should certainly go in before 1.1 is released (since then it's not backwards-incompatible), and we can make backwards-compatible changes for anything defined pre-1.0 (and deprecate in 2.x or whatever).

Zempashi: Indeed

Zempashi (approved):

PR #373, opened 2018-09-18
Update third-party chart versions

NicolasT:
- Basic test for `metrics-server`
- Updates of various chart versions
- Update Elasticsearch to use `RollingUpdate` deployments

### DIFF
--- a/roles/kube_metrics_server/vars/main.yml
+++ b/roles/kube_metrics_server/vars/main.yml
@@ -2,6 +2,9 @@
 kube_metrics_server__default_values:
   args:
   # Default:
-  #- --source=kubernetes.summary_api:''
-  # For https://github.com/scality/metalk8s/issues/36:
-  - "--source=kubernetes.summary_api:''?useServiceAccount=true&kubeletHttps=true&kubeletPort=10250&insecure=true"
+  - --logtostderr
+  # MetalK8s-specific:
+  - --kubelet-insecure-tls
+  # Required to make sure node hostnames aren't resolved through DNS
+  # See https://github.com/kubernetes-incubator/metrics-server/issues/131
+  - --kubelet-preferred-address-types=InternalIP,Hostname,InternalDNS,ExternalDNS,ExternalIP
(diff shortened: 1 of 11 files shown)


NicolasT:
To be investigated why the `metrics-server` test is failing.

NicolasT:
Tests were failing due to https://github.com/scality/metalk8s/pull/373/files#diff-36d18a34df637e8f16174465ed8e3350R8 not being there (DNS resolution of `ci-metalk8s` fails, of course).

NicolasT:
Ugh, new failure. Potentially timing issue: metrics-server doesn't immediately provide metrics, only after some time...

NicolasT:
Should be fixed... https://github.com/scality/metalk8s/pull/373/commits/69c8125afcc93fd82954d9b41188be21bc91ce55#diff-e7316cf09e38171b2f945d5fa5f92032R5

anurag4DSB (approved):

roles/kube_metrics_server/vars/main.yml
```diff
@@ -2,6 +2,9 @@
 kube_metrics_server__default_values:
   args:
   # Default:
-  #- --source=kubernetes.summary_api:''
-  # For https://github.com/scality/metalk8s/issues/36:
-  - "--source=kubernetes.summary_api:''?useServiceAccount=true&kubeletHttps=true&kubeletPort=10250&insecure=true"
+  - --logtostderr
+  # MetalK8s-specific:
+  - --kubelet-insecure-tls
+  # Required to make sure node hostnames aren't resolved through DNS
```
Zempashi, line 8: Should we adapt the inventory to not change the hostname of the server and get DNS working ?

NicolasT: Then you rely on an external DNS server to resolve the hostname, which is not necessarily the case. Defaulting to InternalIP should be fine in most deployments, no? In the end, that's the IP `kubelet`s are reporting themselves.

Zempashi: Yes, indeed. Now wondering why it's not the default

Zempashi: maybe for tls...

NicolasT: Kubespray not generating certificates correctly is why we need `--kubelet-insecure-tls` in the first place :-(

PR #380, opened 2018-09-19
docs: add Kube Heapster and Kube metrics server to advanced configuration

alxf:
see #372


NicolasT (approved):
Fine, though we really need to improve this document :smiley:

PR #386, opened 2018-09-19
docs: Share introduction between HTML and LaTeX

NicolasT:
In 639c7d13e6e, the introduction text was moved into a separate
document, and replaced by a short description in the generated HTML
index page. This is somewhat sub-optimal.

These changes rearrange the documents to share the text. However, due to
how Sphinx TOC trees work, there's no way to conditionally include or
exclude documents, so a new entry-point document for the LaTeX rendering
is required, and the TOC trees need to be kept in sync.

See: 639c7d13e6ee750e80e2252e335cf1c1bf601ba4
See: #378
See: https://github.com/scality/metalk8s/pull/378

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -157,7 +157,7 @@ latex_elements = {
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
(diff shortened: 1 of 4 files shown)


NicolasT:
@bert-e status

NicolasT:
@bert-e retry

Zempashi (approved):

Issue #364, opened 2018-09-14
Upgrade kubespray with support of 1.11.*


virtuman:
Are you planning to upgrade included kubespray to the most recent version, including support of 1.11.* kubernetes?

Thanks

NicolasT:
Yes we do, however we'll only do so after the test-suite to validate upgrades landed (cfr. #344), and updating to K8s 1.11 will not be in the MetalK8s 1.0 release cycle.

See also https://github.com/scality/metalk8s/projects/4

NicolasT:
See #403.

PR #403, opened 2018-09-25
Update Kubespray to HEAD

NicolasT:
Fixes: #364
Fixes: https://github.com/scality/metalk8s/issues/364


NicolasT:
Mainly looking for CI feedback. Need to update ChangeLog, at least.

Running an upgrade on my test cluster as we speak.

NicolasT:
@bert-e wait

NicolasT:
This is a massive change (which means we really should try to stick closer to upstream...). We do have some testing in CI now, though I really think this needs 'actual review'.

I've gone over the changes already, will do so again. Been running into some troubles on my install, though that one is kinda broken due to #302 testing, so not sure it's an actual issue.

Would be useful if others could try to run `deploy.yml` on a pre-existing cluster...

vendor/kubespray/.gitlab-ci.yml
```diff
@@ -93,7 +93,7 @@ before_script:
     # Check out latest tag if testing upgrade
     # Uncomment when gitlab kubespray repo has tags
     #- test "${UPGRADE_TEST}" != "false" && git fetch --all && git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
-    - test "${UPGRADE_TEST}" != "false" && git checkout f7d52564aad2ff8e337634951beb4a881c0e8aa6
+    - test "${UPGRADE_TEST}" != "false" && git checkout 8b3ce6e418ccf48171eb5b3888ee1af84f8d71ba
```
Zempashi, line 96: The upgrade seems to work accross tags. Shouldn't we use tag version instead of master ?

NicolasT: We could, but haven't before. Kubespray releases tend to be... somewhat ad-hoc.

Zempashi (approved):

NicolasT:
Thx for the review @Zempashi

PR #409, opened 2018-09-26
Check all hostnames are lowercase

### LINKED ISSUE #399: precheck for hostname casing
Seems that calico version of kubespray do not like hostname with capital letter in it.
Add a precheck to fail before install

ballot-scality:
Due to a bug in Calico, the hostname must be composed of only lowercase
characters : https://github.com/projectcalico/calico/issues/1572

Fixes #399

### DIFF
--- a/roles/preflight_checks/action_plugins/validate_inventory.py
+++ b/roles/preflight_checks/action_plugins/validate_inventory.py
@@ -1,6 +1,7 @@
 '''Various inventory validation checks, and an Ansible `Action` to run them'''
 
 import inspect
+import re
 import sys
 
 from ansible.plugins.action import ActionBase
@@ -16,7 +17,7 @@ from ansible.plugins.action import ActionBase
 
 
 def check_etcd_ensemble_size(task_vars):
-    ensemble_size = len(task_vars['groups']['etcd'])
+    ensemble_size = len(task_vars['groups'].get('etcd', []))
 
     assert ensemble_size % 2 == 1, \
         'etcd ensemble size should be odd, currently {}'.format(ensemble_size)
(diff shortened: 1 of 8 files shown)


NicolasT:
@ballot-scality This should go in 1.0

roles/preflight_checks/action_plugins/validate_inventory.py
```diff
@@ -109,6 +109,28 @@ def check_ansible_user_is_not_root(task_vars):
...
+    '''Ensure there is only lower case hostnames in the inventory
+
+    Kubespray set the hostname with lower case based on the inventory content
+    but this breaks calico, therefore we should enforce the usage of lower
+    letter only in the hostnames defined in the inventory
+    Kubnespray uses the `inventory_hostname` variable
+    '''
+
+    for host, hostvars in sorted(task_vars['hostvars'].items()):
+        if any([letter.isupper() for letter in host]):
```
NicolasT, line 135: I think we should extend this and make sure there are no other nonsense characters in the name either. Basically `^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$` or something along those lines ('Starts with a-z0-9, can have a-z0-9 or - in the middle, ends with a-z0-9')

NicolasT:
@bert-e wait

ballot-scality:
Miss click on my part I intended to merge in 1.0 (hence the diverge). Will fix tomorrow as well as the regexp.

roles/preflight_checks/action_plugins/validate_inventory.py
ballot-scality: Hum, what about unicode characters then ? Technically it's valid, but does calico supports that ? If we go that far, shouldn't we do a full validation of the hostname provided as being valid FQDN instead ?

ballot-scality:
@bert-e reset

roles/preflight_checks/action_plugins/validate_inventory.py
NicolasT: I would rule out too 'particular' things for now.

NicolasT:
@ballot-scality You want to implement the tests as part of this PR?

ballot-scality:
Would like too but can do in another PR if you prefer

NicolasT:
Nah here is fine. Guess it shouldn't take too long to get done.

NicolasT (approved):
That's *a lot* more code than I expected :wink: but LGTM.

NicolasT:
@bert-e status

PR #418, opened 2018-10-02
Fix `make shell` on OSX

### LINKED ISSUE #192: `make shell` failing to start on OS X
When executing make-shell on macos we are getting the following error:

```
bash-4.4$ make shell
Creating virtualenv...
Installing Python dependencies...
osc-lib 1.9.0 requires Babel!=2.4.0,>=2.3.4, which is not installed.
osc-lib 1.9.0 requires cliff!=2.9.0,>=2.8.0, which is not installed.
osc-lib 1.9.0 requires keystoneauth1>=3.3.0, which is not installed.
osc-lib 1.9.0 requires openstacksdk>=0.9.19, which is not installed.
osc-lib 1.9.0 requires os-client-config>=1.28.0, which is not installed.
osc-lib 1.9.0 requires oslo.i18n>=3.15.3, which is not installed.
osc-lib 1.9.0 requires oslo.utils>=3.33.0, which is not installed.
osc-lib 1.9.0 requires simplejson>=3.5.1, which is not installed.
osc-lib 1.9.0 requires stevedore>=1.20.0, which is not installed.
python-heatclient 1.14.0 requires Babel!=2.4.0,>=2.3.4, which is not installed.
python-heatclient 1.14.0 requires cliff!=2.9.0,>=2.8.0, which is not installed.
python-heatclient 1.14.0 requires iso8601>=0.1.11, which is not installed.
python-heatclient 1.14.0 requires keystoneauth1>=3.3.0, which is not installed.
python-heatclient 1.14.0 requires oslo.i18n>=3.15.3, which is not installed.
python-heatclient 1.14.0 requires oslo.serialization!=2.19.1,>=2.18.0, which is not installed.
python-heatclient 1.14.0 requires oslo.utils>=3.33.0, which is not installed.
python-heatclient 1.14.0 requires PrettyTable<0.8,>=0.7.1, which is not installed.
python-swiftclient 3.5.0 requires futures>=3.0; python_version == "2.7" or python_version == "2.6", which is not installed.
elasticsearch 6.0.0 has requirement urllib3<1.23,>=1.21.1, but you'll have urllib3 1.23 which is incompatible.
Downloading kubectl...
Downloading Helm...
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [shell] Error 2
```

Looking at the commit that introduced this bug we found the following one: 

```
MacBook-Pro-de-Thomas:metal-k8s thomas$ git bisect good
06c5bbc5dc41afec40f1ce1dd40a089cc7f3984f is the first bad commit
commit 06c5bbc5dc41afec40f1ce1dd40a089cc7f3984f
Author: Nicolas Trangez <ikke@nicolast.be>
Date:   Tue Jun 19 16:59:27 2018 +0200

    Makefile: Add `C` option to `shell` target

:100644 100644 3d072e3aa6a56ad7fe052ef809d88925abc9c6a0 c862ccdfd54980652bed9e1061b5e337b215c5fd M	Makefile
```

Zempashi:
Fix #192

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ $(KUBECTL_BIN): $(SHELL_ENV_EXISTS)
 	$(V)mv $@.tmp $@
 
 KUBECTL = $(VENV_BIN)/kubectl
-KUBECTL_REALPATH = realpath $(KUBECTL_BIN)
+KUBECTL_REALPATH = $(PYTHON) hack/realpath.py $(KUBECTL_BIN)
 $(KUBECTL): $(KUBECTL_BIN) $(VENV_EXISTS)
 	$(V)ln -sf $(shell $(KUBECTL_REALPATH))$(KUBECTL_REALPATH:sh) $@
 
@@ -119,7 +119,7 @@ $(HELM_BIN): $(HELM_SRC_TAR) $(SHELL_ENV_EXISTS)
 	$(V)touch $@
 
 HELM = $(VENV_BIN)/helm
-HELM_REALPATH = realpath $(HELM_BIN)
+HELM_REALPATH = $(PYTHON) hack/realpath.py $(HELM_BIN)
 $(HELM): $(HELM_BIN) $(VENV_EXISTS)
 	$(V)ln -sf $(shell $(HELM_REALPATH))$(HELM_REALPATH:sh) $@
 
--- a/Makefile
+++ b/Makefile
@@ -130,17 +130,17 @@ $(BASHRC): $(SHELL_ENV_EXISTS) hack/shell-bashrc
 
 shell: $(VENV_EXISTS) $(REQUIREMENTS_INSTALLED) $(KUBECTL) $(HELM) $(BASHRC) ## Run a shell with `ansible-playbook`, `kubectl` and `helm` pre-installed
 	$(V)BASH_ENV="$(BASHRC)"; export BASH_ENV; if test -z "$(C)"; then \
-		# Interactive shell \
+		`# Interactive shell` \
 		echo "Launching MetalK8s shell environment. Run 'exit' to quit."; \
 		bash --rcfile $(BASHRC) ||:; \
 	elif test "x$(C)" = "x-"; then  \
-		# Semi-interactive shell, keeping exit-code \
+		`# Semi-interactive shell, keeping exit-code` \
 		bash --rcfile $(BASHRC); \
 	elif test -e "$(C)"; then \
-		# A script \
+		`# A script` \
 		bash --rcfile $(BASHRC) "$(C)"; \
 	else \
-		# A command \
+		`# A command` \
 		bash --rcfile $(BASHRC) -c "$(C)"; \
 	fi
 .PHONY: shell
--- a/hack/realpath.py
+++ b/hack/realpath.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python
+
+import optparse
+import os.path
+
+
+def realpath(files):
+    for file_ in files:
+        print(os.path.realpath(file_))
+
+
+if __name__ == '__main__':
+    parser = optparse.OptionParser()
+    (_, pos_args) = parser.parse_args()
+    realpath(pos_args)
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -27,6 +27,8 @@ Bugs fixed
 :ghissue:`381` - warn when Bash completion is not available in `make shell`
 (:ghpull:`382`)
 
+:ghissue:`192` - `make shell` failing to start on OS X (:ghpull:`418`)
+
 Release 1.0.0
 =============
 This marks the first production-ready release of `MetalK8s`_. Deployments using


Makefile
```diff
@@ -97,9 +97,9 @@ $(KUBECTL_BIN): $(SHELL_ENV_EXISTS)
 	$(V)mv $@.tmp $@
 
 KUBECTL = $(VENV_BIN)/kubectl
-KUBECTL_REALPATH = realpath $(KUBECTL_BIN)
+KUBECTL_REALPATH = $(PWD)/$(KUBECTL_BIN)
 $(KUBECTL): $(KUBECTL_BIN) $(VENV_EXISTS)
-	$(V)ln -sf $(shell $(KUBECTL_REALPATH))$(KUBECTL_REALPATH:sh) $@
+	$(V)ln -sf $(KUBECTL_REALPATH)$(KUBECTL_REALPATH:sh) $@
```
NicolasT: Now, this is no longer POSIX `make` compliant now...

NicolasT:
@bert-e wait

Makefile
Zempashi: Oups sorry, don't know every bit of POSIX. Which part is not posix ?

thomasdanan (dismissed):
works for me on mac os sierra 10.13.6 (17G65)

Makefile
NicolasT: Consider how `KUBECTL_REALPATH` is a *variable* which contains a *command*, instead of being the result of said command.

In GNU `make` one can do `$(shell STRING)`, which is not POSIX. In POSIX `make`, one needs `$(STRING:sh)` to achieve the same (and GNU `make` ignores this, go figure).

So, *or* the value of `KUBECTL_REALPATH` remains a string which represents a command, *or* it's a value and then the `sh` invocation is no longer required.

This applies to all instances.

NicolasT: Also, I don't like it when `Makefile`s depend on `$(PWD)`, but that's just me. Notice how your changes will break things when someone runs `make SHELL_VENV=/tmp/foo`.

NicolasT:
Also, needs a ChangeLog entry.

NicolasT:
Basically, make sure this works also with `make` being `bmake` and `$SHELL` being `dash`.

Makefile
Zempashi: Ah indeed, this will break with setting SHELL_ENV (Can we just make a distinction between shell_env starting with / or not ?). We can also remove this part, and require realpath (which seem to be a C POSIX function and not a command). The proper fix of the ticket is the backquotes bellow

NicolasT: Alternative: add a `realpath.py` using `os.path.realpath` in `hack/` (there are some replacements for standard GNU tools in there already because of OSX), and shell-out to that one.

NicolasT (changes requested):
I'd implement the Python script slightly differently, but LGTM once history gets squashed a bit.

Needs a ChangeLog entry though!

Zempashi:
Still need to test `bmake` (make of any bsd except mac os ?) and `dash` (ubuntu ?)

Zempashi:
Luckily bmake package exists on ubuntu. So, bmake + dash works (should we add a test in the CI for that ?)

NicolasT:
> Luckily bmake package exists on ubuntu. So, bmake + dash works (should we add a test in the CI for that ?)

Given the low number of changes we make to `Makefile`, I think getting that into CI would be a bit wasteful right now.

Maybe consider squashing the ChangeLog update in the actual commit that introduces the fix?

hack/realpath.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def realpath(files):
+    for file_ in files:
+        print(os.path.realpath(file_))
+
+
+if __name__ == '__main__':
+    parser = optparse.OptionParser()
+    (_, pos_args) = parser.parse_args()
```
NicolasT, line 14: Not sure why you need `optparse`, could as well use `sys.argv[1:]` or something alike. Anyway, nothing major.

Zempashi: If you call `python ./hack/realpath.py -s /path/to/file` you have a message that option -s doesn't exist instead of some "realpath" of "-s"

Zempashi:
> Maybe consider squashing the ChangeLog update in the actual commit that introduces the fix?

On the 'backquote' one ?

hack/realpath.py
NicolasT: Sure, yet at the same time:
- We don't use `-s` in `Makefile`
- Now this tool simply disregards options and as such behaves different than `realpath` anyway, actually providing an incorrect result when running with e.g. `-s` :)

NicolasT:
> > Maybe consider squashing the ChangeLog update in the actual commit that introduces the fix?
> 
> On the 'backquote' one ?

Yup

hack/realpath.py
Zempashi:
```
# realpath -a foo
realpath: invalid option -- 'a'            
Try 'realpath --help' for more information.

```
(my GNU realpath has an option '-s', so bad example)
but any invalid option make realpath raise (and as we don't support any option right now, it seems a proper behaviour to raise as well)

NicolasT: Ah, didn't consider `parse_args` raises when unknown args are found.

All good then :+1:

NicolasT (approved):

Zempashi:
@bert-e reset

Zempashi:
wake up bert-e ! The integration branch has survived the CI ;)

Zempashi:
bert-e alive ?

Issue #430, opened 2018-10-10
RTD builds broken


NicolasT:
The recent update of Sphinx version used to build the docs breaks the `sphinxcontrib-googleanalytics` extension, which causes docs building on RTD to fail (cfr. https://readthedocs.org/projects/metal-k8s/builds/7909046/)

Since this extension is not really maintained, having a site-local one is likely the right move here.

PR #431, opened 2018-10-10
docs: Fix Google Analytics Sphinx extension

NicolasT:
The extension we used to use is broken when using Sphinx 1.8, but seems no longer maintained. As such, replacing it with a site-local version.

Also enabling it in CI.

Fixes: #430
Fixes: https://github.com/scality/metalk8s/issues/430
See: https://github.com/sphinx-doc/sphinx/commit/3d8cb12497dfa9a044e4b4eaa98b39c4000c1915#commitcomment-30845327

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,12 +11,17 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
 
 import os
+import sys
+
+sys.path.insert(
+    0,
+    os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        '_lib',
+    )
+)
 
 ON_RTD = os.environ.get('READTHEDOCS') == 'True'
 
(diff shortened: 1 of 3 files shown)


NicolasT:
Even though `development/1.0` is not actually impacted by the original issue (since it's using Sphinx 1.7), I think it's better to use the same extension in all versions if possible.

eve/main.yml
```diff
@@ -98,6 +98,9 @@ stages:
         name: 'Build doc'
         haltOnFailure: true
         command: tox --workdir /tmp/tox -e docs -- html latexpdf
+        env:
+          # Fake we're building in a ReadTheDocs environment
+          READTHEDOCS: True
```
tcarmet: I think that's why we're getting a weird error in eve @NicolasT

Would you mind trying the following:
```yaml
READTHEDOCS: "True"
```

NicolasT: Good catch...

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

eve/main.yml
tcarmet: :)

docs/conf.py
```diff
@@ -11,12 +11,19 @@
...
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
 
 import os
+import sys
+
+sys.path.insert(
+    0,
+    os.path.abspath(
```
Zempashi: joining an abspath with something result in an abspath ? Is this strictly needed ?

NicolasT: Strictly speaking not, I guess.

NicolasT:
@bert-e reset

Zempashi (approved):

PR #450, opened 2018-10-23
Some draft documentation on disk replacement

Zempashi:


docs/operations-guide/replacing_a_disk.rst
```diff
@@ -0,0 +1,62 @@
+Replacing a disk
+================
+
+To change the disk of a server, first put the node into a maintenance
+state. (TODO: link to node maintenance)
```
ballot-scality, line 5: In my experience, having a documentation for the prerequisites of a node maintenance (and recover) as separate docs that are included in each procedure that needs them is a major plus for the operator.

Especially during procedure like that, if there is one thing that I wanted to avoid at all cost was having the operator to follow a set of documentation via links and have the risk to mess up.

Spoiler alert, it still happened but a lot less with a single procedure to follow

PR #454, opened 2018-10-29
Skip elasticsearch external value test when external vars already defined

Zempashi:


tests/install/test_external_values.py
```diff
@@ -22,8 +22,10 @@ def test_nginx_ingress_external_values():
 
 @scenario('features/external_values.feature',
           'Change parameters of Elasticsearch')
-def test_elasticsearch_external_values():
-    pass
+def test_elasticsearch_external_values(inventory_obj):
+    node = inventory_obj.get_groups_dict()['kube-node'][0]
```
Zempashi: 'kube-master' is more relevant

ballot-scality (approved):

Issue #485, opened 2018-11-14
environment setup for documentation writers should be easier


tcarmet:
The documentation is getting (or will get) better everyday with new fonts, pictures, and obviously new tools to build it. 

It's pretty easy to have a working environment when you want to build the doc into html. But things can get a little bit more complicated when building the pdf.

Could we make it easy for our futures contributors and colleagues? Per example make a good use of the docker image that is already available and provide a makefile/script (whatever) to build it and run a proper shell. Any other solution will do as long as it's easy and works either on windows, mac or linux.

Thanks

NicolasT:
Something like https://github.com/scality/metalk8s/pull/486 @tcarmet ?

thomasdanan:
Contributing to documentation in Metal 2 is simple enough

PR #486, opened 2018-11-19
docs: Add `docker`-based targets

NicolasT:
The drawback of this approach is that all generated files are owned by
`root:root`. Running the container using the current user doesn't work,
because the Tox environment is owned by `root`.

A more elaborate building method, where the environment is set up using
the current user from the start, could be coded, though may cause other
complications.

Fixes: https://github.com/scality/metalk8s/issues/485

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -16,6 +16,10 @@ SPHINXAUTOBUILDOPTS = \
 	--re-ignore '4913$$' \
 	--re-ignore '\.git/'
 
+DOCKER        = docker
+DOCKERIMAGE   = metalk8s-docs:latest
+SED           = sed
+
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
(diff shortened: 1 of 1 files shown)


docs/Makefile
```diff
@@ -25,6 +29,10 @@ help:
 livehtml:
 	@$(SPHINXAUTOBUILD) $(SPHINXAUTOBUILDOPTS) $(SPHINXOPTS) $(O) "$(SOURCEDIR)" "$(BUILDDIR)"
 
+docker-%:
```
tcarmet, line 32: Will the `Makefile` work on windows? Or do we have to make the same rule on the `make.bat` file?

tcarmet (commented):
LGTM and answer perfectly to the issue I created! Thanks

docs/Makefile
NicolasT: I guess, though I'm not particularly versed in batch-file coding... nor Docker on Windows.

tcarmet: me neither on batch-file coding, but regarding Docker on windows, the good news is that it works exactly the same.

I don't know how to translate the same rule on batch-file but if we had a simple `shell` rule like the following:

```Makefile
shell:
      cd $(SOURCEDIR)/.. && $(DOCKER) build -t $(DOCKERIMAGE) -f docs/Dockerfile .
      $(DOCKER) run -v $(shell get-proper-path):/usr/src/metalk8s --rm $(DOCKERIMAGE) --entrypoint=bash


```
The equivalent in batch-file would be something like:
```batch
:shell
        %DOCKER% build -t %DOCKERIMAGE% -f Dockerfile ..
        %DOCKER% run -it  -v %~dp0:/usr/src/metalk8s --rm %DOCKERIMAGE% --entrypoint=bash
```
I know that it might not be as powerful that the `docker-%` rule but at least our windows user will get a proper environment to build their docs.

NicolasT: I'd rather keep `tox` to be the entry-point to everything. When adding such `shell` rule, all of a sudden we have to explain how to build the docs using a second mechanism.

tcarmet: Did not thought about that. That's a very good point. agree.

tcarmet (approved):
Tested on my computer and it works perfectly for `latexpdf` `html` and for the `livehtml` we just miss a proper `-P` in the docker run command.

I will provide in the future the same rule for windows Makefiles.

NicolasT:
Right about `livehtml` and `-P`. Would need a change to the `Dockerfile` then as well.

PR #532, opened 2018-12-11
Raise error at the end of the warmup countdown period of metrics server

Zempashi:


Zempashi:
@bert-e approve

tests/post/test_metrics_server.py
```diff
@@ -17,7 +17,7 @@ def wait_until_initialized(kubeconfig):
     # It can take up to a minute before metrics-server scraped some stats, see
     # https://github.com/kubernetes-incubator/metrics-server/issues/134 and
     # https://github.com/kubernetes-incubator/metrics-server/issues/136
-    for _ in utils.helper.retry(90, wait=1):
+    while next(utils.helper.retry(90, wait=1)):
```
NicolasT, line 20: Not sure I like this construct. I'm actually surprised it works. `while EXPR` should re-evaluate `EXPR` on every iteration, right? Otherwise something like `while foo_is_not_ready(): sleep(10)` would not work.

However, because `EXPR` (being `next(utils.helper.retry(90, wait=1))`) is re-executed on every iteration, this means you get a *new* `utils.helper.retry` every single time, no? So this is basically an endless loop...

Zempashi: Ah indeed, I'm stupid

Zempashi: But the for doesn't work either because by definition "for" catch "StopIteration"

PR #553, opened 2019-02-06
Add Salt states to install kubelet and its prerequisites

### LINKED ISSUE #545: Salt State to bring the system up to a running kubelet
**Goal**
Have a salt state (that will be invoked using salt-call) in order to install all kubelet prereqs: containerd, dependencies, enable services ….

**Delivery**
The salt state

TeddyAndrieux:
To test using a `salt-minion` in local mode, assuming MetalK8s is cloned in `$HOME`:

- Execute `salt-call --local --pillar-root $HOME/metalk8s/pillar/ --file-root $HOME/metalk8s/salt/ state.sls containerd` to install `runc`, `containerd`, and enable/start `containerd`.
- Execute `salt-call --local --pillar-root $HOME/metalk8s/pillar/ --file-root $HOME/metalk8s/salt/ state.sls kubelet` to install `kubelet`.

Only 'online' installation is currently supported.

See: #545

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -1,0 +1,3 @@
+metalk8s-2.0:
+  '*':
+    - repositories
--- a/pillar/repositories.sls
+++ b/pillar/repositories.sls
@@ -1,0 +1,4 @@
+# Repositories informations
+repo:
+  # Configure online repositories
+  online_mode: True
--- a/salt/metalk8s/containerd/configured.sls
+++ b/salt/metalk8s/containerd/configured.sls
@@ -1,0 +1,9 @@
+include:
+  - .installed
+
+Start and enable containerd:
+  service.running:
+    - name: containerd
+    - enable: True
+    - require:
+      - pkg: Install containerd
--- a/salt/metalk8s/containerd/init.sls
+++ b/salt/metalk8s/containerd/init.sls
@@ -1,0 +1,2 @@
+include:
+  - .configured
--- a/salt/metalk8s/containerd/installed.sls
+++ b/salt/metalk8s/containerd/installed.sls
@@ -1,0 +1,20 @@
+{%- from "metalk8s/map.jinja" import repo with context %}
+
+include:
+  - metalk8s.repo
+  - metalk8s.runc
+
+Install container-selinux:
+  pkg.installed:
+    - name: container-selinux
+
+Install containerd:
+  pkg.installed:
+    - name: containerd
+    - version: {{ repo.containerd.version }}
+    - fromrepo: {{ repo.containerd.name }}
+    - require:
+      - pkgrepo: Configure Kubernetes repository
+      - pkg: Install runc
+      - pkg: Install container-selinux
+
--- a/salt/metalk8s/defaults.yaml
+++ b/salt/metalk8s/defaults.yaml
@@ -1,0 +1,8 @@
+repo:
+  online_mode: True
+  runc:
+    version: latest
+  containerd:
+    version: latest
+  kubernetes:
+    version: latest
--- a/salt/metalk8s/kubelet/init.sls
+++ b/salt/metalk8s/kubelet/init.sls
@@ -1,0 +1,2 @@
+include:
+  - .installed
--- a/salt/metalk8s/kubelet/installed.sls
+++ b/salt/metalk8s/kubelet/installed.sls
@@ -1,0 +1,20 @@
+{%- from "metalk8s/map.jinja" import repo with context %}
+
+include:
+  - metalk8s.repo
+
+# TODO: Maybe not needed in offline because embedded in the kubernetes repository
+Install kubelet dependencies:
+  pkg.installed:
+    - pkgs:
+      - ebtables
+      - socat
+      - conntrack-tools
+
+Install kubelet:
+  pkg.installed:
+    - name: kubelet
+    - version: {{ repo.kubernetes.version }}
+    - fromrepo: {{ repo.kubernetes.name }}
+    - require:
+      - pkgrepo: Configure Kubernetes repository
--- a/salt/metalk8s/map.jinja
+++ b/salt/metalk8s/map.jinja
@@ -1,0 +1,28 @@
+{% import_yaml 'metalk8s/defaults.yaml' as defaults with context %}
+
+{% set defaults = salt['grains.filter_by']({
+  'default': defaults
+}, merge=salt['pillar.items']()) %}
+
+{% set repo = salt['grains.filter_by']({
+  'RedHat': {
+    'containerd': {
+      'name': 'epel'
+    },
+    'kubernetes': {
+      'name': 'kubernetes'
+    }
+  }
+}, merge=defaults.get('repo')) %}
+
+{% set containerd = salt['grains.filter_by']({
+  'default': {}
+}, merge=defaults.get('containerd')) %}
+
+{% set kubelet = salt['grains.filter_by']({
+  'default': {}
+}, merge=defaults.get('kubelet')) %}
+
+{% set runc = salt['grains.filter_by']({
+  'default': {}
+}, merge=defaults.get('runc')) %}
--- a/salt/metalk8s/repo/configured.sls
+++ b/salt/metalk8s/repo/configured.sls
@@ -1,0 +1,8 @@
+{%- from "metalk8s/map.jinja" import repo with context %}
+
+include:
+{%- if repo.online_mode %}
+  - .online
+{%- else %}
+  - .offline
+{%- endif %}
--- a/salt/metalk8s/repo/init.sls
+++ b/salt/metalk8s/repo/init.sls
@@ -1,0 +1,2 @@
+include:
+  - .configured
--- a/salt/metalk8s/repo/offline.sls
+++ b/salt/metalk8s/repo/offline.sls
@@ -1,0 +1,3 @@
+offline mode not yet implemented:
+  test.fail_without_changes:
+    - failhard: True
--- a/salt/metalk8s/repo/online.sls
+++ b/salt/metalk8s/repo/online.sls
@@ -1,0 +1,26 @@
+{%- from "metalk8s/map.jinja" import repo with context %}
+
+Configure EPEL repository:
+  pkg.installed:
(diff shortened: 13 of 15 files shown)


pillar/top.sls
```diff
@@ -0,0 +1,3 @@
+base:
```
NicolasT: We won't use a `base` environment, instead have that one always-fail. Environments are MetalK8s-version specific.

salt/containerd/installed.sls
```diff
@@ -0,0 +1,24 @@
...
+
+include:
+  - repo
+  - runc
+
+Install containerd:
+  pkg.installed:
+    - name: containerd
+    - version: {{ containerd.version }}
+    - fromrepo: {{ repo.containerd.name }}
```
NicolasT: So, this means we keep the *version* of `containerd` we want to install, and the *repository* from where to retrieve it, are managed in separate locations. Is that a wise approach? Would it make sense to combine these in one place?

salt/containerd/installed.sls
```diff
@@ -0,0 +1,24 @@
...
+    - require:
+      - pkgrepo: Configure Kubernetes repository
+      - pkg: Install runc
+
+Update container-selinux:
```
NicolasT: Would it make sense to install this before `containerd`, so the upgrade isn't required and the `containerd` dependency on `container-selinux` is already met?

salt/repo/map.jinja
```diff
@@ -0,0 +1,18 @@
...
+{%- set repo = salt['grains.filter_by']({
+  'RedHat': {
+    'containerd': {
+      'name': 'epel'
+    },
+    'kubernetes': {
+      'name': 'kubernetes'
+    }
+  },
+  'Debian': {
```
NicolasT: Not supported, so let's not mention it, if possible.

salt/repo/online.sls
```diff
@@ -0,0 +1,36 @@
+{%- from "repo/map.jinja" import repo with context %}
+
+{%- if grains["os_family"] == "RedHat" %}
```
NicolasT: We only support this environment for now, so let's not add `if`/`then` constructions prematurely.

salt/repo/online.sls
```diff
@@ -0,0 +1,36 @@
...
+  pkg.installed:
+  {%- if grains["os"] == "CentOs" %}
+    - name: epel-release
+  {%- else %}
+    - sources:
+      - epel-release: https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ grains['osmajorrelease'] }}.noarch.rpm
+  {%- endif %}
+
+  {%- if grains["os"] == "RedHat" %}
+check that the system is registered:
```
NicolasT: Why?

NicolasT (changes requested):

salt/containerd/installed.sls
TeddyAndrieux: Oh yes it's true I will fix it

- [x] install container-selinx before containerd

pillar/top.sls
TeddyAndrieux: I'm not sure I understand, you mean we should use `metalk8s-2.0` instead of `base` ?

salt/containerd/installed.sls
TeddyAndrieux: Just wanted to separate repo stuff from real program stuff but maybe version should be part of the repo, make sense.

- [x] move version into repo stuff

salt/repo/map.jinja
TeddyAndrieux: True

- [x] Remove Debian from map.jinja

salt/repo/online.sls
TeddyAndrieux: Ok

- [x] do not check if we run on redhat based as it's requirement (for the moment)

salt/repo/online.sls
TeddyAndrieux: It's just a check as we need the system to be registered on RedHat to be able to install online stuff but maybe it's a requirement and we should remove this from the state it's ok for me

pillar/top.sls
NicolasT: Yes. We briefly discussed this in one of the meetings, but maybe it wasn't entirely clear. Should document this, I guess...

NicolasT: Ah, thinking about it, maybe that doesn't apply to pillars, because `master_tops` only applies to states?

TeddyAndrieux: No you right we want to use it with saltenv and pillarenv need to do some research about it

- [x] do not use `base` in pillar top.sls

salt/repo/online.sls
NicolasT: Nah, I've seen this on other prereq lists we have, so if there's a use for this, can stay as far as I'm concerned. I trust your judgement.

salt/containerd/installed.sls
```diff
@@ -0,0 +1,24 @@
...
+  pkg.installed:
+    - name: container-selinux
+    - sources:
+      # TODO: Maybe something to specify link/version
+      - container-selinux: https://buildlogs.centos.org/centos/7/virt/x86_64/container/container-selinux-2.77-1.el7.noarch.rpm
```
NicolasT: Looks like the upstream spec was updated to this required version, so we may be able to retrieve a package from some standard repository soon'ish.

https://git.centos.org/commit/rpms!container-selinux.git/c6b9b7aa65faa452b086fca6fa564639a451e673

salt/containerd/map.jinja
```diff
@@ -0,0 +1,5 @@
+{%- set containerd = salt['grains.filter_by']({
+  'Linux': {
+    'version': 'latest'
```
slaperche-scality: Do we always want the latest? Don't we want to freeze the version?

NicolasT: Eventually, yes.

TeddyAndrieux: Yes agree the point is that by default we use latest but we can overwrite it in the pillar

slaperche-scality: Yeah, I see the logic.

But the default value should be a valid one, and by using latest we take the (limited?) risk to pick "by default" an incompatible/untested version (different major version number).

Anyway, since we're at the start of the project nothing is freezed yet I guess, so we can go with latest for now.

salt/containerd/installed.sls
TeddyAndrieux: True 2.77 now available. I will update

salt/repo/map.jinja
```diff
@@ -0,0 +1,25 @@
+{%- set defaults = salt['grains.filter_by']({
+  'Linux': {
+    'online_mode': True,
+    'runc': {
+      'version': 'latest'
+    },
+    'containerd': {
+      'version': 'latest'
+    },
+    'kubernetes': {
```
NicolasT: I get why this is called `kubernetes`, in a way. However, shouldn't this be called `kubelet`, which is defaulted from some higher-level `kubernetes` version value? The reason is that, when upgrading, the version of `kubernetes` (i.e. contol plane) and `kubelet` on various nodes may not be the same.

salt/runc/installed.sls
```diff
@@ -0,0 +1,10 @@
+{% from "repo/map.jinja" import repo with context %}
```
NicolasT: Is it 'good practice' to let one top-level state tree reference another?

salt/repo/map.jinja
TeddyAndrieux: Yes i could do something to specify version of kubelet which default to version in kubernetes section. 
But we want to be able to have a different version between kubelet and other "kubernetes stuff" ?
During an upgrade we want to upgrade all of them right ?

salt/runc/installed.sls
TeddyAndrieux: Honnestly, I don't really know for that. 

For me we have 2 choice:
1. we import like here
2. we copy paste the map.jinja content (full or only part of it) from repo to runc/map.jinja (and same for others containerd, kubelet, ...)

salt/repo/map.jinja
NicolasT: During an upgrade, you don't upgrade everything all at once, there's a specific order in which to do things.

However, I guess we can figure that out when the case arises, then split version variables or whatever. So not necessarily any changes to be made right now.

salt/runc/installed.sls
NicolasT: I think we may want to have every top-level 'directory' to be fully self-standing, i.e. not depend on anything else. However, that's a project-wide guideline/rule to be defined by the team, if applicable. Since there's no rule right now, can keep this as-is.

NicolasT (dismissed):
Whatever changes we want to do, we can do them later. Basic functionality seems fine.

NicolasT:
@bert-e help

NicolasT:
@bert-e bypass_build_status

salt/runc/installed.sls
ballot-scality: Following our discussion on slack, I would say that a map.jinja should be referenced only by state in the same directory or in sub directories of the said map.jinja.

salt/runc/map.jinja
```diff
@@ -0,0 +1,3 @@
+{%- set runc = salt['grains.filter_by']({
+  'Linux': {}
+}, grain='kernel', merge=salt['pillar.get']('runc')) %}
```
ballot-scality: Please don't use this hack :).

The correct way to set this is using the "default" argument (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.grains.html#salt.modules.grains.filter_by)

```
{%- set runc = salt['grains.filter_by']({
  'default': {}
} merge=salt['pillar.get']('runc')) %}
```

ballot-scality (changes requested):
See my comment about grains.filter_by module in map.jinja

salt/runc/map.jinja
TeddyAndrieux: Oh true it's better :)

ballot-scality (approved):

NicolasT (approved):

TeddyAndrieux:
@bert-e approve

PR #565, opened 2019-02-11
chore: import stub `Vagrantfile`

NicolasT:


Vagrantfile
```diff
@@ -0,0 +1,65 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
```
slaperche-scality: minor: could add `# frozen_string_literal: true` to be compatible with future Ruby versions.

Vagrantfile
```diff
@@ -0,0 +1,65 @@
...
+
+$IMPORT_RELEASE = <<-SCRIPT
```
slaperche-scality: Do we need a global variable here? A constant would be enough, no?

Vagrantfile
```diff
@@ -0,0 +1,65 @@
...
+# This script is supposed to 'install' the build results on the system,
+# i.e. symlink something like `/vagrant/_build/root` to
+# `/srv/scality/metalk8s-$VERSION` or however we build this.
+
+echo "Installed build artifacts on the host"
+
+exit $RC
+SCRIPT
+
+$BOOTSTRAP =<<-SCRIPT
```
slaperche-scality: missing space after `=`

Vagrantfile
NicolasT: This is not really a full-fledged Ruby script, but sure, will do. Also, I'm not a Ruby dev, this file is mostly knit together from stuff found in the Vagrant docs :innocent:

NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #567, opened 2019-02-11
Build calico-cni-plugin RPM package

sayf-eddine-scality:
As there is no `calico-cni-plugin.rpm` in the wild, so we should build it. This PR provides a `Makefile` that will produce the RPM package.

The `Makefile` uses Docker as a build environment.

**Usage**
make CALICO_VERSION=x.x.x
**UPDATE**
Second iteration of the PR
- Now the docker image is very generic and all the logic is done inside the Makefile
- We query the spec file for a couple of things ( version, arch, release ) before building the package
- The docker image acts on a mounted volume to create srpm from sources then to build the rpm
- I couldnt find a querytag to retrieve sources so I fell back 'shamelessly' on `spectool -g -R`


build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# TODO: centrally define a base image 
+ARG BUILD_IMAGE_SHA256=5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121
+ARG BUILD_IMAGE=docker.io/centos
+
+FROM ${BUILD_IMAGE}@sha256:${BUILD_IMAGE_SHA256} as build
+
+# Declare user props
+ARG ARG_UID=1000
+ARG ARG_GID=1000
+ARG ARG_KEY_PUB
```
NicolasT: Unused.

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ARG ARG_USER=scality
+ARG ARG_GRP=scality
+ARG ARG_WORKDIR=/home/buildlab/$ARG_USER
+
+# Make sure we have sudo set up
```
NicolasT: Why? Shouldn't need `sudo` for building when using `mock` or similar.

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN yum install -y sudo \
+    && mkdir -p /home/buildlab \
+    && if ! getent group $ARG_GID; then groupadd -g $ARG_GID $ARG_GRP; fi \
```
NicolasT: What if `ARG_GID` would already be allocated?

In general, these tests etc. aren't really required: we're in a well-known environment. Also, I believe the intent is to keep system UID/GID and container UID/GID the same. There's no real need, all you want to sync is the UID/GID of the 'result' files of a build. Others you actually *want* to be read-only.

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    && if ! getent passwd $ARG_UID; then useradd -m -d "${ARG_WORKDIR}" -u $ARG_UID -g $ARG_GID -s /bin/bash $ARG_USER; fi \
+    && echo "$(getent passwd ${ARG_UID} | cut -d ':' -f 1) ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/$ARG_USER \
+    && echo "Defaults:$(getent passwd ${ARG_UID} | cut -d ':' -f 1) !requiretty" >> /etc/sudoers.d/$ARG_USER
+USER ${ARG_USER}
+WORKDIR ${ARG_WORKDIR}
+
+# Install the build dependencies
+RUN sudo yum install -y yum-utils rpm-build rpmlint rpmdevtools
```
NicolasT: Could move this much earlier, then there's no need for `sudo`.

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Run the build
+CMD rpmdev-setuptree $ARG_WORKDIR \
+    && spectool -g -R ${ARG_WORKDIR}/*.spec \
+    && sudo yum-builddep ${ARG_WORKDIR}/*.spec \
+    && rpmbuild -bb ${ARG_WORKDIR}/*.spec \
+    && cp ${ARG_WORKDIR}/rpmbuild/RPMS/*/*.rpm ${ARG_WORKDIR} \
+    # && rpmlint ${ARG_WORKDIR}/*.rpm \ This will report error the way we 'build' this rpm
```
NicolasT: Why can't we have `rpmlint`?

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    && rpmdev-wipetree ${ARG_WORKDIR}/rpmbuild \
+    && rm -r rpmbuild
```
NicolasT: Since everything gets removed, why not work outside of the bind-mounted user dir?

build/calico-cni-plugin/Makefile
```diff
@@ -0,0 +1,32 @@
...
+ARG_GRP:=$$(getent group $(ARG_UID) | cut -d ':' -f 1)
+
+ifndef CALICO_VERSION
+    $(error CALICO_VERSION not given as arg)
+endif
+
+default: all
+.PHONY: default
+
+all: calico-cni-plugin
```
NicolasT: Instead of a phony target, make a real target (the RPM) with real inputs (the spec, `Dockerfile`,...)

build/calico-cni-plugin/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+calico-builder-image:
+	docker build -t calico-builder \
+       --build-arg ARG_UID=$(ARG_UID) \
+       --build-arg ARG_GUID=$(ARG_GUID) \
+       --build-arg ARG_USER=$(ARG_USER) \
+       --build-arg ARG_GRP=$(ARG_GRP) \
+       .
+
+calico-cni-plugin.spec:
```
NicolasT: See below.

build/calico-cni-plugin/calico-cni-plugin.spec.tmpl
```diff
@@ -0,0 +1,53 @@
+%global provider        github
+%global provider_tld    com
+%global project         projectcalico
+%global repo            cni-plugin
+%global provider_prefix %{provider}.%{provider_tld}/%{project}/%{repo}
+
+%ifarch x86_64
+%global built_arch              amd64
+%global calico_sha256           5736d46dc8f3d17eafd1d7874c49457244b698d5fccf726066c60c136ac5a318
```
NicolasT: So, `CALICO_VERSION` gets templated (at only one location), but these aren't. I.e. we can't change `CALICO_VERSION`, things would break anyway.

build/calico-cni-plugin/calico-cni-plugin.spec.tmpl
```diff
@@ -0,0 +1,53 @@
...
+%global calico_ipam_sha256      ae32622d5c904e216cd914031e560e481c9ea9c213153172d0eacabadbb5e984
+%endif
+
+Name:           calico-cni-plugin
+Version:        %CALICO_VERSION%
```
NicolasT: I prefer to have this hard-coded. One needs to update the spec to get a different RPM.

build/calico-cni-plugin/calico-cni-plugin.spec.tmpl
```diff
@@ -0,0 +1,53 @@
...
+Release:        1%{?dist}
```
NicolasT: Iff we keep `CALICO_VERSION` templated, then this should be too.

NicolasT: Though, again, I don't think we should: otherwise, `Changelog` won't be kept up-to-date etc.

build/calico-cni-plugin/calico-cni-plugin.spec.tmpl
```diff
@@ -0,0 +1,53 @@
...
+install -p -m 755 %{SOURCE2} %{buildroot}/opt/cni/bin/calico-ipam
+
+%files
+/opt/cni/bin/calico
+/opt/cni/bin/calico-ipam
+
+%license LICENSE
+%doc README.md
+
+%changelog
```
NicolasT: We'll likely want to do a 3.5.0 as well :)

NicolasT (changes requested):
Some concerns about the approach: I believe using `mock` (maybe still inside Docker), we could come up with a solution which

- doesn't need `sudo`
- doesn't need too much UID/GID juggling

Related: how can this approach work? At which stage are the binaries downloaded from Github?

Also, over time let's try to do a 2-stage build: one which turns a spec and sources into a source-RPM (and keep it), then one to turn said source-RPM into a binary RPM.

NicolasT:
One more note: in #566 I added a `Makefile` which will put whatever is required in `_build/root/...`, i.e. some target folder. May make sense to do the same or something similar. Also, if we have a top-level `Makefile`, this one should build the resulting RPM (if required) as part of the overall build, either by invoking another `Makefile`, or having the required functionality built-in.

build/calico-cni-plugin/Makefile
```diff
@@ -0,0 +1,32 @@
+ARG_UID:=$$(id -u)
+ARG_GUID:=$$(id -g)
```
slaperche-scality: `ARG_GID` (that's what you use in the Dockerfile, and it's more consistent with the UID above).

build/calico-cni-plugin/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+default: all
+.PHONY: default
+
+all: calico-cni-plugin
+.PHONY: all
+
+calico-cni-plugin: calico-builder-image calico-cni-plugin.spec
+	docker run --rm \
+       -v `pwd`:/home/buildlab/$(ARG_USER)/ \
```
slaperche-scality: use `$()` instead of ` `` ` + you probably need quote around in case the path contains spaces.

slaperche-scality (changes requested):
You wrote a nice description for the PR, but the commit message is almost empty :-(

build/calico-cni-plugin/Dockerfile
sayf-eddine-scality: statically linking binary inside an RPM is not acceptable y rpmlint it seems

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Equal to centos:7.6.1810
+# SHA256 digest of the base image
+# TODO: centrally define a base image
+ARG BUILD_IMAGE_SHA256=5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121
+ARG BUILD_IMAGE=docker.io/centos
+FROM ${BUILD_IMAGE}@sha256:${BUILD_IMAGE_SHA256} as build
+
+RUN yum install -y yum-utils rpm-build rpmlint rpmdevtools
+
+CMD ["/bin/bash"]
```
NicolasT: No need for this, it's the default of the base image.

build/calico-cni-plugin/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Equal to centos:7.6.1810
+# SHA256 digest of the base image
+# TODO: centrally define a base image
+ARG BUILD_IMAGE_SHA256=5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121
+ARG BUILD_IMAGE=docker.io/centos
+FROM ${BUILD_IMAGE}@sha256:${BUILD_IMAGE_SHA256} as build
+
+RUN yum install -y yum-utils rpm-build rpmlint rpmdevtools
```
NicolasT: Two remarks:

- Consider using something like
```
RUN yum install -y \
    rpm-build \
    rpmdevtools \
    rpmlint \
    yum-utils
```
which reduces the size of a diff when a tool needs to be added (adding a single line instead of replacing one is easier to review)

- Add `&& yum clean all` at the end of the `RUN` stanza.

NicolasT (changes requested):
Some initial feedback but needs changes in the `Makefile` as well, will go over them.

NicolasT:
We should spend some time on this tomorrow. I have a branch which implements what I tried to explain, this may help in the conversation.

sayf-eddine-scality:
Superseded by #590

PR #569, opened 2019-02-12
Fix build infrastructure for MacOS X

thomasdanan:
just to have Makefile works on macos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ $(ISO_ROOT)/bootstrap.sh: scripts/bootstrap.sh.in
 $(ISO_ROOT)/salt/%: salt/%
 	mkdir -p $(shell dirname $@)
 	rm -f $@
-	cp --archive $< $@
+	cp -a $< $@
 
 $(ISO_ROOT)/pillar/top.sls: pillar/top.sls.in
 	mkdir -p $(shell dirname $@)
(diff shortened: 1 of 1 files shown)


slaperche-scality:
No problem with the long options of `mkisofs`? Only `cp` was an issue?

NicolasT (dismissed):

NicolasT:
@thomasdanan You'll need to self-approve as well.

thomasdanan:
@slaperche-scality @NicolasT indeed I did not test ISO generation (as it is not called in default target) ... so I have to install crdtools do a kind of manual sym link after brew install and now I get some errors I would need to fix ...
Also that makes me think that we should list pre reqs somewhere

NicolasT:
Default target builds the ISO (as it should, because it's the final and only deliverable) when #572 goes in.

thomasdanan:
so for now the issue is -checksum_algorithm_iso is not recognise option of mkisofs on my env ...

slaperche-scality:
> so for now the issue is -checksum_algorithm_iso is not recognise option of mkisofs on my env ...

Yup, can't find this option on my system as well (see https://github.com/scality/metalk8s/pull/566#discussion_r255935387)
I think it's either something Debian-specific (but seems not) or `genisoimage` embeds a patched version of `mkisofs`.

This option seems related to "Jigdo production" (from https://lists.debian.org/debian-boot/2012/03/msg00445.html).

NicolasT:
Then remove said option.

thomasdanan:
@bert-e approve

NicolasT (approved):

Issue #548, opened 2019-02-05
kubeadm: Salt'ify `kubelet-start` phase


thomasdanan:
**Goal**
Implement a Salt *state* which mimicks `kubeadm`s `kubelet-start` phase.

**Delivery**
The Salt *state*.

PR #570, opened 2019-02-13
Add Salt states for `kubeadm init phase kubelet-start`

TeddyAndrieux:
Add `kubeadm init phase kubelet-start` Salt state.

Included in Salt state `metalk8s.kubeadm.init` or directly `metalk8s.kubeadm.init.kubelet-start`

Refs: #548


salt/metalk8s/defaults.yaml
```diff
@@ -6,3 +6,13 @@ repo:
     version: latest
   kubernetes:
     version: latest
+kubelet:
+  config_file: "/var/lib/kubelet/config.yaml"
+  config:
+    apiVersion: kubelet.config.k8s.io/v1beta1
```
TeddyAndrieux: Don't know how we can infer this one

NicolasT: What do you mean with 'infer'? From what?

salt/metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
```diff
@@ -0,0 +1 @@
+KUBELET_KUBEADM_ARGS={% for key, value in options.items() %}--{{ key }}={{ value }} {% endfor %}
```
slaperche-scality, line 2: simple question: does `value` need to be escaped for a shell (as I see `--` like it's forwarded as command line at one point)?

salt/metalk8s/defaults.yaml
```diff
@@ -6,3 +6,13 @@ repo:
     version: latest
   kubernetes:
     version: latest
```
NicolasT, line 8: Consider adding an empty line here, to lay out various 'sections' of this configuration.

salt/metalk8s/kubeadm/init/kubelet-start/files/service-kubelet-systemd.conf
```diff
@@ -0,0 +1,10 @@
+# Note: This dropin only works with kubelet v1.11+
+[Service]
+Environment="KUBELET_CONFIG_ARGS=--config={{ conf_file }}"
+# This is a file that "kubeadm init" and "kubeadm join" generates at runtime, populating the KUBELET_KUBEADM_ARGS variable dynamically
+EnvironmentFile=-{{ env_file }}
+# This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
+# the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
+EnvironmentFile=-/etc/sysconfig/kubelet
+ExecStart=
```
slaperche-scality, line 10: why this line? Isn't the next line enough?

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,62 @@
...
+    - source: salt://metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - context:
+        options: {{ kubelet.service.options }}
+
+{#- Default dataset for kubelet config #}
+{%- set dataset = salt['grains.filter_by']({
```
NicolasT: Not sure why you're splitting the configuration in two parts, here and above.

If this has a reason, some changes will be required anyway: you specify the `kind` here, and the `apiVersion` above, whilst it's the combination of those two which defines the schema of the configuration/object, so splitting them doesn't make much sense.

salt/metalk8s/defaults.yaml
TeddyAndrieux: I don't know, it's depend of the version of kubelet but maybe we stuck to a specific config apiVersion for a given version of `metalk8s`

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,62 @@
...
+  'default': {
+    'kind': 'KubeletConfiguration',
+    'authentication': {
```
NicolasT: I'm a bit surprised these (and `authorization` below) are required to be specified. I seem to remember they were not before.

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,62 @@
...
+      'anonymous': {'enabled': true},
+      'webhook': {'enabled': false}
+    },
+    'authorization': {'mode': 'AlwaysAllow'}
+  }
+}, merge=kubelet.config) %}
+Create kubelet config file:
```
NicolasT, line 16: This applies to other files created as well (like the environment file above), but: https://github.com/scality/metalk8s/blob/development/2.0/docs/requirements.rst#tagged-generated-files

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,62 @@
...
+    - name: {{ kubelet.service.environment_file }}
+    - source: salt://metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - context:
+        options: {{ kubelet.service.options }}
+
+{#- Default dataset for kubelet config #}
```
slaperche-scality: Is it a comment? Can't be simply `# Default dataset for kubelet config`?

salt/metalk8s/kubeadm/init/kubelet-start/files/service-kubelet-systemd.conf
```diff
@@ -0,0 +1,10 @@
+# Note: This dropin only works with kubelet v1.11+
```
NicolasT, line 2: Same as above: https://github.com/scality/metalk8s/blob/development/2.0/docs/requirements.rst#tagged-generated-files
I know the generated file will then not be 100% equal to the one dropped by the `kubeadm` package, but that's fine.

salt/metalk8s/kubeadm/init/kubelet-start/init.sls
```diff
@@ -0,0 +1,13 @@
+#
+# State for kubeadm init kubelet-start phase.
+#
+# The goal is to have a static kubelet running
+#
+# Available states
+# ================
+#
+# * configured   -> configure kubelet service
```
NicolasT: Configure and start/enable.

NicolasT (changes requested):
Some remarks, LGTM overall.

salt/metalk8s/kubeadm/init/kubelet-start/files/service-kubelet-systemd.conf
NicolasT: It's required to 'clear' the value, otherwise things get appended to what's in the system-global unit file.

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: With my test with kubelet1.13 it's required and also if I only enable anonymous but not disable webhook  authentication that doesn't work.
And for authorization it's seems that it's a known bug https://github.com/kubernetes/kubernetes/issues/68939 should be AlwaysAllow by default but it's not

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: Oh yes True i will do it

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: No because it's a comment about jinja stuff so it's a jinja comment :)

salt/metalk8s/defaults.yaml
TeddyAndrieux: Will do

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: My idea was to be able to change "changing stuff" from the pillar.
The kind for kubelet configuration is always `KubeletConfiguration` no matter the `apiVersion` or i'm wrong ?

But yes I can add it in the default.yaml also make sense. Will do it as the result is the same

salt/metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
TeddyAndrieux: Yes it's args for kubelet service start, should maybe at least quote `value`. I make it create a file like kubeadm does

salt/metalk8s/kubeadm/init/kubelet-start/init.sls
TeddyAndrieux: Yes

salt/metalk8s/defaults.yaml
NicolasT: Partially discussed offline :) It goes with a version of `kubelet`, so it also goes with a specific version of MetalK8s.

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
NicolasT: Discussed offline :)

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
NicolasT: Ah. So, the reason things were 'different' in my testing: I wasn't using a config file at all, but passing required options (actually, only the manifest path) on the command-line.

Given this, the state is now not 'idempotent' with what's going to happen later on, when we create the 'real' `kubelet` configuration file. If we'd re-run the state, we'll end up replacing said 'real' config with this minimal one again, and replace with the 'real' later, etc.

Maybe this file should only be generated with this config if it does not already exist, or rather, it doesn't specify the manifest path properly, and we don't really care about other settings?

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: Btw for a `file.serialize` i don't find any "real" way to add a comment

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: Yes I agree with file.serialize we can merge also and it's maybe make sense for `staticPodPath` but the problem is the authentication and authorization stuff because for kind apiversion and staticPodPath we could always merge I think.

But basically if we want a running kubelet at this step we will have the "not idempotent" somewhere.

TeddyAndrieux: We can maybe have 2 service files for kubelet
one who run kubelet with no webhook
and then we create another real service config file that reset the args of kubelet

salt/metalk8s/kubeadm/init/kubelet-start/files/service-kubelet-systemd.conf
```diff
@@ -0,0 +1,11 @@
+# Kubelet service configuration file generated by metalk8s
```
NicolasT: `MetalK8s`

salt/metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
```diff
@@ -0,0 +1,2 @@
+# Environment file generated by metalk8s
```
NicolasT: `MetalK8s`

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,51 @@
...
+        conf_file: {{ kubelet.config_file }}
+    - require:
+      - file: Create kubelet service environment file
+      - file: Create kubelet config file
+
+Start and enable kubelet:
+  service.running:
+    - name: kubelet
+    - enable: True
+    - watch:
```
ballot-scality, line 51: I think "onchanges" is better here.
"watch" would trigger the restart whatever happens.

TeddyAndrieux: Only if these file are modified so btw it's what we want I think no ?

Don't know if it's also do like a `systemctl daemon-reload`

ballot-scality: After rereading the doc, you're right

NicolasT: You need `daemon-reload` when the unit and/or `EnvironmentFile` gets updated, not when the `kubelet` configuration file gets updated. So yes, you need it.

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
ballot-scality: if we want to achieve idempotency, the only solution I can think of right now is to have a "temporary" kubelet service file generated only if the "final" kubelet service is not here.
This way, the first run is indeed taking 2 steps to reach the final kubelet state, but in case of upgrade or rerun, we have only one, idempotent state.

Something like

```
Create temporary kubelet service:
  file.managed:
    - name: /etc/systemd/system/kubelet-metalk8s-bootstrap.service
    .
    .
    - onlyif: ! test -f /etc/systemd/system/kubelet-metalk8s.service

Create kubelet service:
  file.managed:
    - name: /etc/systemd/system/kubelet-metalk8s.service
    .
  service.running:
     .
     .
```

TeddyAndrieux: Yes was my previous comment just with an only if.
Ok so I think we will do it in a latter PR at the `kubeconfig kubelet` kubeadm phase I think, so we know what exactly needed

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: Indeed I will add it

TeddyAndrieux: Ok no, it's not needed already done by salt since 29 Sep 2016 https://github.com/saltstack/salt/pull/36538

ballot-scality (changes requested):
I think you forgot to update the Makefile with the list of new states :)

```
diff --git a/Makefile b/Makefile
index 7b762e4..107fa31 100644
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,12 @@ ALL = \
        \
        $(ISO_ROOT)/salt/metalk8s/defaults.yaml \
        \
+       $(ISO_ROOT)/salt/metalk8s/kubeadm/init/init.sls \
+       $(ISO_ROOT)/salt/metalk8s/kubeadm/init/kubelet-start/configured.sls \
+       $(ISO_ROOT)/salt/metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env \
+       $(ISO_ROOT)/salt/metalk8s/kubeadm/init/kubelet-start/files/service-kubelet-systemd.conf \
+       $(ISO_ROOT)/salt/metalk8s/kubeadm/init/kubelet-start/init.sls \
+       \
        $(ISO_ROOT)/salt/metalk8s/kubelet/init.sls \
        $(ISO_ROOT)/salt/metalk8s/kubelet/installed.sls \
        \

```

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,51 @@
+{%- from "metalk8s/map.jinja" import kubelet with context %}
+
+Create kubelet service environment file:
```
ballot-scality, line 3: Missing makedirs or creating the parent directories

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
```diff
@@ -0,0 +1,51 @@
...
+    - source: salt://metalk8s/kubeadm/init/kubelet-start/files/kubeadm.env
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - context:
+        options: {{ kubelet.service.options }}
+
+Create kubelet config file:
+  file.serialize:
```
ballot-scality, line 17: Missing makedirs or creating the parent directories

ballot-scality (changes requested):

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
NicolasT: Ah, too bad. Maybe we should do a custom/derived module one day. OK then, for now, though maybe we should create a tech-debt ticket to track cases like this.

TeddyAndrieux: Or if it's only about dump a file then we can use a file.managed and just dump a yaml file. But if one day we just want to merge them then .... I don't know

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: This directory is created by the kubelet package I think but I can add a makedirs just "au cas où"

salt/metalk8s/kubeadm/init/kubelet-start/configured.sls
TeddyAndrieux: Same

TeddyAndrieux:
@bert-e approve

ballot-scality (dismissed):
LGTM

ramanduh (dismissed):

NicolasT (approved):

PR #572, opened 2019-02-13
build: rework `make` targets, and add `vagrantup`

NicolasT:
- By default, build the ISO, which is the final deliverable anyway.
- Add a `vagrantup` target which performs a local build, then runs `vagrant up` to launch the VMs (if not done before) and (re)run the provisioning scripts
- Add a `help` target to display targets and what they do

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,11 @@ ALL = \
 default: all
 .PHONY: default
 
-all: $(ALL)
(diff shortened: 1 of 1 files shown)


slaperche-scality (approved):

NicolasT:
@bert-e approve

Issue #573, opened 2019-02-13
New version of `container-selinux` required


NicolasT:
As discussed in https://github.com/scality/metalk8s/pull/553#pullrequestreview-200788883. The current upstream version of `container-selinux`, 2.74-1, isn't compatible with `containerd`. We need 2.77-1 or newer.

The spec-file for this version is in the CentOS Git repository, but no builds seem to be available. However, there's a ScientificLinux build at ftp://ftp.pbone.net/mirror/ftp.scientificlinux.org/linux/scientific/7x/external_products/extras/x86_64/container-selinux-2.77-1.el7_6.noarch.rpm

Would be good to have a script to test `containerd` functionality, until we have better testing infrastructure in place.

NicolasT:
Closed by #574

PR #574, opened 2019-02-13
fix: install recent version of `container-selinux`

NicolasT:
The default `container-selinux` policies as provided by CentOS (2.74) are not compatible with `containerd`/`runc` (AVC denials when `runc` attempts to `setattr` on `fifo_file` resources, see links below).

This PR forces the install of a third-party package (so this introduces some technical debt...), and includes a simple test-script (to be invoked manually) to check `containerd` is working correctly.

Fixes: https://github.com/scality/metalk8s/issues/573
See: https://github.com/containers/libpod/issues/1980
See: https://github.com/containers/container-selinux/commit/ae6e25bc789d305c4fdebe063b40fafe49614323

### DIFF
--- a/salt/metalk8s/containerd/installed.sls
+++ b/salt/metalk8s/containerd/installed.sls
@@ -6,7 +6,8 @@ include:
 
 Install container-selinux:
   pkg.installed:
-    - name: container-selinux
+    - sources:
+      - container-selinux: ftp://ftp.pbone.net/mirror/ftp.scientificlinux.org/linux/scientific/7x/external_products/extras/x86_64/container-selinux-2.77-1.el7_6.noarch.rpm
 
 Install containerd:
   pkg.installed:
--- a/tests/test-containerd.sh
+++ b/tests/test-containerd.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+set -xue -o pipefail
+
+cat > /etc/cni/net.d/98-containerd-test-bridge.conf << EOF
+{
+    "cniVersion": "0.3.1",
+    "name": "bridge",
+    "type": "bridge",
+    "bridge": "cnio0",
+    "isGateway": true,
+    "ipMasq": true,
+    "ipam": {
+        "type": "host-local",
+        "ranges": [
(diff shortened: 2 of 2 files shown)


NicolasT:
For reviewers: I'm aware the 'test' is very basic and such. However, better than nothing, and we don't really have any kind of test 'framework' right now.

salt/metalk8s/containerd/installed.sls
```diff
@@ -6,7 +6,7 @@ include:
 
 Install container-selinux:
   pkg.installed:
-    - name: container-selinux
+    - source: ftp://ftp.pbone.net/mirror/ftp.scientificlinux.org/linux/scientific/7x/external_products/extras/x86_64/container-selinux-2.77-1.el7_6.noarch.rpm
```
slaperche-scality: It's temporary?

Because relying on "random" FTP seems unreliable/unsafe.

NicolasT: Yes, see #575. If we have the infrastructure in place to generate an offline repository, then this could be part of it (or, hopefully, not because by that time the upstream CentOS repositories have a more recent version of the package).

ballot-scality: This doesn't work for me.
I had to modify like this : 
```
Install container-selinux:
  pkg.installed:
    - sources:
      - container-selinux: ftp://ftp.pbone.net/mirror/ftp.scientificlinux.org/linux/scientific/7x/external_products/extras/x86_64/container-selinux-2.77-1.el7_6.noarch.rpm

```

ballot-scality (changes requested):

ballot-scality (dismissed):
I took the liberty to push the saltstack fix

salt/metalk8s/containerd/installed.sls
NicolasT: That's odd, 'worked on my machine'.
Anyway, given the fix, could you squash commits instead?

NicolasT:
@bert-e approve

TeddyAndrieux (approved):

Issue #576, opened 2019-02-13
Create a Salt module and state to pull or inject images into containerd/CRI


NicolasT:
- Can use `crictl`
- Can use `ctr`
- Can take either a Docker archive file pathto be injected as argument, or an image name to be pulled

NicolasT:
Done in #585

NicolasT:
Done in #585

PR #585, opened 2019-02-13
salt: create modules and states for CRI and `containerd`

NicolasT:
Fixes: https://github.com/scality/metalk8s/issues/576

### DIFF
--- a/salt/_modules/containerd.py
+++ b/salt/_modules/containerd.py
@@ -1,0 +1,60 @@
+'''
+Various functions to interact with :program:`containerd` using :command:`ctr`.
+'''
+
+import re
+import logging
+
+import salt.utils.path
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'containerd'
+
+
+def __virtual__():
+    if not salt.utils.path.which('ctr'):
+        return (False, "'ctr' command not found")
+
+    return __virtualname__
+
+
+_LOAD_RE = re.compile('Loaded image: (?P<name>.*)')
+
+
+def load_cri_image(path):
+    '''
+    Load a Docker image archive into the :program:`containerd` CRI image cache.
+
+    .. note::
+
+       This uses the :command:`ctr` command.
+
+    path
+        Path of the Docker image archive to load
+    '''
+    log.info('Importing image from "%s" into CRI cache', path)
+
+    out = __salt__['cmd.run_all']('ctr cri load "{0}"'.format(path))
+
+    if out['retcode'] != 0:
+        log.error('Failed to load image archive')
+        return None
+
+    stdout = out['stdout']
+
+    ret = {}
+
+    re_match = _LOAD_RE.match(stdout)
+    if re_match:
+        ret['name'] = re_match.group('name')
+
+    if 'name' in ret:
+        log.info(
+            'Imported image from "%s" into CRI cache, name is "%s"',
+            path, ret['name'])
+    else:
+        log.info('Imported image from "%s" into CRI cache, name unknown', path)
+
+    return ret
--- a/salt/_modules/cri.py
+++ b/salt/_modules/cri.py
@@ -1,0 +1,87 @@
+'''
+Various functions to interact with a CRI daemon (through :program:`crictl`).
+'''
+
+import re
+import logging
+
+import salt.utils.json
+
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'cri'
+
+
+def __virtual__():
+    log.debug("Looking for 'crictl'")
+    if not salt.utils.path.which('crictl'):
+        return (False, "'crictl' command not found")
+
+    log.debug("Validating CRI connection / 'crictl' configuration")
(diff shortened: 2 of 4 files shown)


salt/_modules/containerd.py
```diff
@@ -0,0 +1,60 @@
...
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'containerd'
+
+
+def __virtual__():
+    if not salt.utils.path.which('ctr'):
+        return (False, '\'ctr\' command not found')
```
slaperche-scality: Use double quote when the string contains single quote: it's more legible than escaping (or could also use backtick instead of single quote for that purpose).

Also parenthesis are usually omitted here: `return False, "'ctr' command not found"`

salt/_modules/containerd.py
```diff
@@ -0,0 +1,60 @@
...
+
+    return __virtualname__
+
+
+_LOAD_RE = re.compile('Loaded image: (?P<name>.*)')
```
slaperche-scality, line 23: I don't think you need to compile this regexp.

> Note: The compiled versions of the most recent patterns passed to re.compile() and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions.

From https://docs.python.org/3/library/re.html#re.compile

And about the pattern: `.*` accept an empty string, is it what we want?

salt/_modules/containerd.py
```diff
@@ -0,0 +1,60 @@
+'''
```
slaperche-scality, line 1: For this kind of docstring, it's more common to use double quote instead of single quote (on the ring repo it's 1406 vs 29223 occurrences)

salt/_modules/cri.py
```diff
@@ -0,0 +1,87 @@
...
+    if out['retcode'] != 0:
+        log.error('Failed to list images')
+        return None
+
+    return salt.utils.json.loads(out['stdout'])['images']
+
+
+_PULL_RES = {
+    'sha256': re.compile(
+        r'Image is up to date for sha256:(?P<digest>[a-f0-9]{64})'),
```
slaperche-scality: only lowercase hexa?

salt/_states/containerd.py
```diff
@@ -0,0 +1,118 @@
...
+        'pchanges': {},
+        'comment': '',
+    }
+
+    images = __salt__['cri.list_images']()
+    if images is None:
+        ret['comment'] = 'Failed to list CRI images'
+        return ret
+
+    available = False
```
slaperche-scality, line 46: Minor: could be rewritten

```python
    for image in images:
        if name in image.get('repoTags', []):
            available = True
            break
        if name in image.get('repoDigests', []):
            available = True
            break
    else:
        available = False
```

slaperche-scality (changes requested):

salt/_states/containerd.py
NicolasT: I don't like that approach, more of a "Define with failure value, then maybe turn into success" person.

Also, and maybe more importantly, I vehemently dislike the `for/else` construction in Python, which is non-obvious for anyone except those who are familiar with this obscure feature.

slaperche-scality: Fair enough, yeah it's really Python-specific

salt/_modules/containerd.py
ballot-scality: https://docs.saltstack.com/en/latest/topics/development/conventions/style.html

slaperche-scality: Ho, ok then: when in Rome, do as the Romans do.

I didn't know it was Salt style. Then it's correct.

salt/_modules/containerd.py
NicolasT: I wouldn't be surprised Salt (and its modules) use regular expressions all over the place, so there's no way for me to 'predict' this one could be cached automagically.

For empty string: sure, could make it slightly more strict, though the tool will likely never output that, and if it does, then the name of the image would be 'empty string' anyway.

salt/_modules/containerd.py
NicolasT: Yeah, based on formatting used in other Salt modules. However, if we decide to e.g. adopt PEP8 for this project (with proper linting), then I'll adjust accordingly.

salt/_modules/containerd.py
NicolasT: Same as the quotes remark, it's how it's done in other Salt modules.

salt/_modules/cri.py
NicolasT: That's what the tool outputs. Could capture both, I guess. Over time, should rather use the API than invoking a CLI tool and parse its output, though.

- [x] TODO

salt/_modules/containerd.py
NicolasT: Oh, forgot: single-quotes, will do.

NicolasT:
@slaperche-scality Done

NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #586, opened 2019-02-13
build: add product info file to ISO image

NicolasT:
The content of this file can be used to present information in an
'Import ISO' UI, even without mounting the ISO image::

    $ isoinfo -x /PRODUCT.TXT\;1 -i _build/metalk8s.iso


Makefile
```diff
@@ -38,52 +37,66 @@ ALL = \
...
 
-$(ISO): all-local
+$(ISO): $(ALL)
 	mkisofs -output $@ \
 		-rock \
 		-joliet \
 		-joliet-long \
 		-full-iso9660-filenames \
-		-volid 'MetalK8s $(FULL_VERSION)' \
+		-volid '$(shell source $(ISO_ROOT)/product.txt && echo $$NAME) $(shell source $(ISO_ROOT)/product.txt && echo $$VERSION)' \
```
slaperche-scality: Can't the `source` be done only once in the rule (before the call to `mkisofs`, instead of each time for each command.

Or maybe even have a dedicated rule to set the env (since it's already used 4 times).

NicolasT: > Or maybe even have a dedicated rule to set the env (since it's already used 4 times).

How would that work?

Makefile
```diff
@@ -38,52 +37,67 @@ ALL = \
...
 clean: ## Clean the build tree
 	rm -rf $(BUILD_ROOT)
 .PHONY: clean
 
 iso: $(ISO) ## Build the MetalK8s ISO image
 .PHONY: iso
 
-$(ISO): all-local
+$(ISO): $(ALL)
+	source $(ISO_ROOT)/product.txt && \
```
NicolasT, line 94: @slaperche-scality Did it this way, not sure what you meant with a 'target' to achieve this in a prior comment.

Makefile
slaperche-scality: Don't know if it would work, but a maybe Make rule that would do a bunch of `export` (or source a script with a bunch of `export`) of the necessary environment variable.

But if it doesn't work, then we can just `source` the file in each rule that needs it.

Makefile
slaperche-scality: Yup, that looks good to me.

Makefile
NicolasT: Every command in a recipe is executed in a different subshell, so that wouldn't work as-is. Also, the file is generated by this `Makefile`, so the only way to somehow use the variables in a `Makefile` is to re-spawn `make`, which is... not great.

Changing the generated file to do a bunch of `export`s I don't really like because it makes parsing harder, and allows for people to start doing tricks like variable expansion and whatnot, expect it to work, but again makes parsing harder (if not impossible).

I'd like to keep things as-is for now, when it becomes really impractical we can reconsider.

slaperche-scality: Ok for me, the change you already made is good enough for me.

slaperche-scality (approved):

NicolasT:
@bert-e approve

Issue #588, opened 2019-02-14
Install and configure `crictl` when `kubelet` gets installed


NicolasT:

NicolasT:
Done in #589

PR #589, opened 2019-02-14
salt: install and configure `crictl` with `kubelet`

NicolasT:
Fixes: #588

### DIFF
--- a/salt/metalk8s/kubelet/installed.sls
+++ b/salt/metalk8s/kubelet/installed.sls
@@ -11,6 +11,23 @@ Install kubelet dependencies:
       - socat
       - conntrack-tools
 
+Install and configure cri-tools:
+  pkg.installed:
+    - name: cri-tools
+    - fromrepo: {{ repo.kubernetes.name }}
+    - require:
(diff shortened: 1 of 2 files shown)


salt/metalk8s/kubelet/installed.sls
```diff
@@ -11,6 +11,22 @@ Install kubelet dependencies:
...
       - conntrack-tools
 
+Install and configure cri-tools:
+  pkg.installed:
+    - name: cri-tools
+    - fromrepo: {{ repo.kubernetes.name }}
+    - require:
+      - pkgrepo: Configure Kubernetes repository
+  file.serialize:
+    - name: /etc/crictl.yaml
```
ramanduh, line 21: Maybe we can add the merge_if_exists arg so that any custom option in crictl.yaml would not be overwritten

ballot-scality: Good point

NicolasT: Done

NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #590, opened 2019-02-14
Build calico-cni-plugin RPM package

sayf-eddine-scality:
As there is no `calico-cni-plugin.rpm` in the wild, so we should build it. This PR provides a `Makefile` that will produce the RPM package.

The `Makefile` uses Docker as a build environment.


NicolasT (approved):
LGTM :+1:

sayf-eddine-scality:
@bert-e approve

Issue #555, opened 2019-02-06
CI: bootstrap Eve steps - start a worker, run the build step


jbertran:
Using the work of the build team if available, mocking it if not, bootstrap an `eve.yaml` file with the necessary steps to run the build on a worker and retain the output.

jbertran:
Closed by #596

PR #594, opened 2019-02-15
Make eve build and upload the iso file

sayf-eddine-scality:
Eve should spawn an openstack worker to build and upload the iso file

Fixes: #555


NicolasT:
You'll need to fix the branch name.

PR #596, opened 2019-02-15
Make eve build and upload the iso file

sayf-eddine-scality:
Eve should spawn a kube_pod worker to build and upload the iso file

Fixes: #555

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ $(ISO): $(ALL)
 		-input-charset utf-8 \
 		-output-charset utf-8 \
 		$(ISO_ROOT)
+	cd $$(dirname $@) && sha256sum $(notdir $@) > SHA256SUM
 
 
 VAGRANT ?= vagrant
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -8,7 +8,7 @@ stages:
   pre-merge:
     worker:
       type: kube_pod
-      path: eve/workers/pod-builder/pod.yml
+      path: eve/workers/pod-builder/pod.yaml
       images:
           docker-builder: eve/workers/pod-builder
     steps:
@@ -24,9 +24,10 @@ stages:
         command: make
     - ShellCommand:
         name: Put the iso file in a separate folder
-        command: mkdir iso && cp _build/metalk8s.iso iso
+        command: mkdir iso && cp _build/metalk8s.iso _build/SHA256SUM iso
     - Upload:
         name: upload artifacts
         source: iso/
         urls:
           - "*.iso"
+          - "SHA256SUM"
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -15,7 +15,6 @@ stages:
     - Git: &git_pull
         name: git pull
         repourl: "%(prop:git_reference)s"
-        mode: full
         method: clobber
         retryFetch: true
         haltOnFailure: true
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -7,5 +7,26 @@ branches:
 stages:
   pre-merge:
     worker:
-      type: local
-    steps: []
+      type: kube_pod
+      path: eve/workers/pod-builder/pod.yml
+      images:
+          docker-builder: eve/workers/pod-builder
+    steps:
+    - Git: &git_pull
+        name: git pull
+        repourl: "%(prop:git_reference)s"
+        mode: full
+        method: clobber
+        retryFetch: true
+        haltOnFailure: true
+    - ShellCommand:
+        name: build everything
+        command: make
+    - ShellCommand:
+        name: Put the iso file in a separate folder
+        command: mkdir iso && cp _build/metalk8s.iso iso
+    - Upload:
+        name: upload artifacts
+        source: iso/
+        urls:
+          - "*.iso"
--- a/eve/workers/pod-builder/Dockerfile
+++ b/eve/workers/pod-builder/Dockerfile
@@ -4,9 +4,8 @@ ARG BUILDBOT_VERSION=0.9.12
 
 WORKDIR /home/eve/workspace
 
-RUN yum install -y epel-release
-
-RUN yum-config-manager \
+RUN yum install -y epel-release \
+    && yum-config-manager \
     --add-repo \
     https://download.docker.com/linux/centos/docker-ce.repo \
     && yum install -y \
@@ -19,14 +18,7 @@ RUN yum-config-manager \
     yum-utils \
     docker-ce-cli-18.09.2 \
     && adduser -u 1042 --home /home/eve eve --groups docker \
-    && chown eve:eve /home/eve/workspace
-
-COPY requirements-eve.txt /tmp/
-
-RUN pip install -r /tmp/requirements-eve.txt \
+    && chown eve:eve /home/eve/workspace \
     && pip install buildbot-worker==${BUILDBOT_VERSION}
 
 USER eve
-
-CMD buildbot-worker create-worker . "$BUILDMASTER:$BUILDMASTER_PORT" "$WORKERNAME" "$WORKERPASS" \
-    && buildbot-worker start --nodaemon
--- a/eve/workers/pod-builder/Dockerfile
+++ b/eve/workers/pod-builder/Dockerfile
@@ -1,0 +1,32 @@
+FROM centos:7
+
+ARG BUILDBOT_VERSION=0.9.12
+
+WORKDIR /home/eve/workspace
+
+RUN yum install -y epel-release
+
+RUN yum-config-manager \
+    --add-repo \
+    https://download.docker.com/linux/centos/docker-ce.repo \
+    && yum install -y \
+    gcc \
+    make \
+    python-devel \
+    python-pip \
+    genisoimage \
+    git \
+    yum-utils \
+    docker-ce-cli-18.09.2 \
+    && adduser -u 1042 --home /home/eve eve --groups docker \
+    && chown eve:eve /home/eve/workspace
+
+COPY requirements-eve.txt /tmp/
+
+RUN pip install -r /tmp/requirements-eve.txt \
+    && pip install buildbot-worker==${BUILDBOT_VERSION}
+
+USER eve
+
+CMD buildbot-worker create-worker . "$BUILDMASTER:$BUILDMASTER_PORT" "$WORKERNAME" "$WORKERPASS" \
+    && buildbot-worker start --nodaemon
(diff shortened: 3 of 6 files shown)


eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: worker
+      image: docker-builder
+      resources:
+        requests:
+          cpu: 500m
+          memory: 500Mi
+        limits:
+          cpu: 500m
+          memory: 500Mi
+      command: ['sleep', '4000'] # workaround to keep the container "alive"
```
tcarmet: remove this line

eve/workers/pod-builder/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    git \
+    && curl -fsSL https://get.docker.com/ | sh
+    && adduser --home /home/eve eve \
+    && usermod -aG docker eve
+    && pip2 install -r /tmp/requirements-eve.txt \
+    && pip2 install buildbot-worker==0.9.0.post1
+
+USER eve
+WORKDIR /home/eve/workspace
+ENV DOCKER_HOST=localhost
```
tcarmet: IMHO you can remove this line as well. Per example if you're working with this image on your laptop you might bind the docker socket instead of using localhost

eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: worker
+spec:
+  containers:
+    - name: worker
+      image: docker-builder
```
tcarmet: here you need to use a jinja template to refer to your image. eve will then do some magic to point it to the right image on the registry, so it should be `image: {{ images['docker-builder'] }}`

tcarmet: careful, I made a mistake and edited my comment

eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: worker
+spec:
+  containers:
+    - name: worker
+      image: {{ image: images['docker-builder'] }}
```
tcarmet: you're so fast. my bad it's `image: {{ images['docker-builder'] }}`

eve/workers/pod-builder/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM centos/7
+
+COPY requirements-eve.txt /tmp/
+
+RUN yum install -y \
+    python2-pip \
+    genisoimage \
+    git \
+    && curl -fsSL https://get.docker.com/ | sh
```
tcarmet: careful when installing the docker client like that. I really don't recommend it, you don't control the docker version, so everyone and even the CI will have a slightly different image due to when they built it.

eve/workers/pod-builder/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    && adduser --home /home/eve eve \
+    && usermod -aG docker eve
+    && pip2 install -r /tmp/requirements-eve.txt \
+    && pip2 install buildbot-worker==0.9.0.post1
```
tcarmet: please bump the buildbot worker version here to at least `0.9.12`, 

Also eve build the image using a buildarg with buildbot's version so you can use this to always be "up to date" with eve.
`ARG BUILDBOT_VERSION=0.9.12`

eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: worker
+spec:
+  containers:
+    - name: worker
```
tcarmet: in this container your "workspace" is on `/home/eve/workspace` but I don't see any volume attached to it, so you're writing new files on the overlay. I would suggest that you create a new volume of type `emptyDir` just like the `docker-storage` and attach to it the path `/home/eve/workspace`

tcarmet: this will increase the performance :)

eve/main.yml
```diff
@@ -7,5 +7,21 @@ branches:
 stages:
   pre-merge:
```
jbertran, line 8: As we talked about, I'll move the build into a separate step and call it in `TriggerStages` for the pre-merge step in a PR to add linting targets.

eve/main.yml
```diff
@@ -7,5 +7,21 @@ branches:
...
+        mode: full
+        method: clobber
+        retryFetch: true
+        haltOnFailure: true
+    - ShellCommand:
+        name: build everything
+        command: make
+    - Upload:
+        name: upload artifacts
+        source: _build/
```
tcarmet: I would add a `urls` option here to have a link to the iso in the UI.
```yaml
- Upload:
    name: upload
    source: _build/
    urls:
      - *.iso
```

eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,43 @@
+---
```
jbertran: We should probably make a decision on `.yaml` vs. `.yml`.

As it stands, we have `salt/metalk8s/defaults.yaml` and `eve/workers/pod-builder/pod.yml`. Since your PR is the one in review, I'm tempted to say we go for `.yaml` because you're able to make the changes, otherwise it's potato vs. potato.

tcarmet: I think `yaml` is the one used with the kubernetes community, so maybe stick with that ;)

eve/workers/pod-builder/pod.yml
```diff
@@ -0,0 +1,43 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: worker
+spec:
+  containers:
+    - name: worker
```
jbertran: Let's call all our platform CI containers `platform-*` or something.

Makefile
```diff
@@ -130,6 +130,7 @@ $(ISO): $(ALL)
 		-input-charset utf-8 \
 		-output-charset utf-8 \
 		$(ISO_ROOT)
+	sha256sum $@ | cut -d" " -f 1 > $@.sha256sum
```
NicolasT: This way, you're not producing a valid checksum file (i.e. one that can be used as input of `sha256sum --check`). Instead:

```
cd $(dirname $@) && sha256sum $(notdir $@) > SHA256SUM
```
and make sure that file (`SHA256SUM`) gets uploaded.

eve/main.yml
```diff
@@ -7,5 +7,27 @@ branches:
...
-    steps: []
+      type: kube_pod
+      path: eve/workers/pod-builder/pod.yaml
+      images:
+          docker-builder: eve/workers/pod-builder
+    steps:
+    - Git: &git_pull
+        name: git pull
+        repourl: "%(prop:git_reference)s"
+        mode: full
```
NicolasT: Not necessarily.

eve/workers/pod-builder/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM centos:7
+
+COPY requirements-eve.txt /tmp/
```
NicolasT: In general it's better to move 'things that change' after 'things that are less likely to change'. As such, I'd suggest:
```
RUN yum-config-manager --add-repo ... && \
    yum install -y gcc make ... && \
    adduser ... &&
    (no real need for `usermod`, just set things up as part of `adduser`)

ADD requirements-eve.txt /tmp
RUN pip install -r /tmp/... && pip install ...
```

Also, I'm quite surprised why on one end there's a `requirements.txt` for a set of 'fairly random' Python packages, but then there's a separate install for `buildbot-worker`. Why not keep that in `requirements.txt` as well?

sayf-eddine-scality: frankly I dont know, I can do that I just find most of eve dockerfile workers on this format

eve/main.yml
sayf-eddine-scality: which part is not ?

eve/workers/pod-builder/Dockerfile
NicolasT: Ping @tcarmet

eve/main.yml
NicolasT: A full clone instead of a shallow one. One more ping @tcarmet

eve/workers/pod-builder/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    && chown eve:eve /home/eve/workspace
+
+COPY requirements-eve.txt /tmp/
+
+RUN pip install -r /tmp/requirements-eve.txt \
+    && pip install buildbot-worker==${BUILDBOT_VERSION}
+
+USER eve
+
+CMD buildbot-worker create-worker . "$BUILDMASTER:$BUILDMASTER_PORT" "$WORKERNAME" "$WORKERPASS" \
```
jbertran: Consider switching that CMD with the following snippet in `pod.yaml`, so that the concern of starting the pod services remains with the pod file, and setting up the environment is the only concern of the Dockerfile.

```dockerfile
    args: ["-c", "buildbot-worker create-worker . ${BUILDMASTER}:${BUILDMASTER_PORT} ${WORKERNAME} ${WORKERPASS} && buildbot-worker start --nodaemon"]
```

eve/workers/pod-builder/pod.yaml
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: worker
```
jbertran: Let's have a meaningful name here - `build-worker` for example.

eve/workers/pod-builder/pod.yaml
```diff
@@ -0,0 +1,43 @@
...
+spec:
+  containers:
+    - name: worker
```
jbertran: Ditto. Doesn't have to match - `build-container`?

jbertran (changes requested):

eve/workers/pod-builder/pod.yaml
sayf-eddine-scality: as discussed we will keep this name until we get more info from releng

jbertran (dismissed):

eve/main.yml
tcarmet: Yep, not necessary indeed.

jbertran (dismissed):
Good to go :)

eve/workers/pod-builder/Dockerfile
tcarmet: I also appreciate leaving everything in `requirements.txt` as long as you set a version for the buildbot-worker. Could be any version as long as it's newer than the one you have as a build arg.

sayf-eddine-scality:
@bert-e approve

jbertran (approved):

Issue #605, opened 2019-02-18
salt: create a Docker registry state


NicolasT:
Develop a state which allows us to either assert some image exists in some repository, some image tag refers to a known digest, and upload a Docker archive into said repository.

All of this can be achieved using `skopeo`, which may be a prerequisite.

PR #608, opened 2019-02-18
salt: add module and state to manage images in a Docker registry

NicolasT:
Fixes: #605

### DIFF
--- a/salt/_modules/docker_registry.py
+++ b/salt/_modules/docker_registry.py
@@ -1,0 +1,87 @@
+'''
+Functions to interact with a Docker registry.
+'''
+
+import logging
+
+import salt.utils.json
+import salt.utils.path
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'docker_registry'
+
+
+def __virtual__():
+    if not salt.utils.path.which('skopeo'):
+        return (False, "'skopeo' command not found")
+
+    return __virtualname__
+
+
+def get_image_manifest(image, tls_verify=True):
+    '''
+    Retrieve an image manifest from a repository.
+
+    .. note::
+
+       This uses the :command:`skopeo` command.
+
+    image
+        Name of the image (e.g. `localhost:5000/metalk8s/nginx:1.14.2`)
+    tls_verify : True
+        Verify TLS certificate of the repository. Set to `False` if the
+        repository uses HTTP instead of HTTPS.
+    '''
+    log.info('Retrieving image manifest for "%s"', image)
+
+    cmd = ['skopeo', 'inspect']
+
+    cmd.append('--tls-verify={0}'.format('true' if tls_verify else 'false'))
+
+    cmd.append('docker://{0}'.format(image))
+
+    out = __salt__['cmd.run_all'](' '.join(cmd))
+
+    if out['retcode'] != 0:
+        log.error('Failed to retrieve image manifest')
+
+        if 'manifest unknown' in out['stderr']:
+            return None
+        else:
+            raise salt.exceptions.CommandExecutionError(
+                'Failure while retrieving image manifest: {0}'.format(
+                    out['stderr']))
+
+    return salt.utils.json.loads(out['stdout'])
+
+
+def push_docker_archive(archive_path, destination, dest_tls_verify=True):
+    '''
+    Push a Docker image archive to a repository.
+
+    .. note::
+
+       This uses the :command:`skopeo` command.
+
+    archive_path
+        Path of the Docker image archive to push
+    destination
+        Image destination, e.g. `localhost:5000/metalk8s/nginx:1.14.2`
+    dest_tls_verify : True
+        Verify TLS certificate of the repository. Set to `False` if the
+        repository uses HTTP instead of HTTPS.
+    '''
+    log.info('Pushing archive "%s" to "%s"', archive_path, destination)
+
+    cmd = ['skopeo', 'copy']
+    cmd.append(
+        '--dest-tls-verify={0}'.format('true' if dest_tls_verify else 'false'))
+
+    cmd.append('docker-archive:{0}'.format(archive_path))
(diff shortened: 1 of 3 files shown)


salt/_modules/docker_registry.py
```diff
@@ -0,0 +1,87 @@
...
+    destination
+        Image destination, e.g. `localhost:5000/metalk8s/nginx:1.14.2`
+    dest_tls_verify : True
+        Verify TLS certificate of the repository. Set to `False` if the
+        repository uses HTTP instead of HTTPS.
+    '''
+    log.info('Pushing archive "%s" to "%s"', archive_path, destination)
+
+    cmd = ['skopeo', 'copy']
+    cmd.append(
```
slaperche-scality, line 79: very minor: since the opening parenthesis is on its own line, it's better to do the same for the closing one

```
    cmd.append(
        '--dest-tls-verify={0}'.format('true' if dest_tls_verify else 'false')
    )
```

salt/_states/docker_registry.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def __virtual__():
+    if 'docker_registry.get_image_manifest' not in __salt__:
+        return (False, "Missing 'docker_registry' module")
+
+    return __virtualname__
+
+
+def image_managed(name, archive_path=None, digest=None, force=False,
```
slaperche-scality, line 22: minor (layout): isn't

```
def image_managed(name, archive_path=None, digest=None, force=False,
                  tls_verify=True):
```

or

```
def image_managed(
    name, archive_path=None, digest=None, force=False,tls_verify=True
):
```

more legible? Or maybe it's a salt coding style again (if so, feel free to ignore this comment).

salt/_states/docker_registry.py
```diff
@@ -0,0 +1,202 @@
...
+                ret['comment'] = 'Image up-to-date'
+                ret['result'] = True
+            else:
+                ret['comment'] = 'Image digest conflict'
+                ret['result'] = False
+
+        return ret
+
+
+    if image_manifest is not None and digest is not None:
```
slaperche-scality, line 71: minor: Could be computed before l.53, then `digest_match` could be use l.61

salt/_states/docker_registry.py
```diff
@@ -0,0 +1,202 @@
...
+                },
+            })
+            ret['result'] = True
+        else:
+            ret['comment'] = 'Failed to push image'
+            ret['result'] = False
+
+        return ret
+    else:
+        assert False, 'Logic error'
```
slaperche-scality, line 202: This function is too long (~150 lines) with several piece of code duplicated (l.74-81 vs l.61-66)or very similar (`ret['changes'].update`).

This could probably be factored with 1-2 helper functions.

slaperche-scality (changes requested):

salt/_modules/docker_registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+    .. note::
+
+       This uses the :command:`skopeo` command.
+
+    image
+        Name of the image (e.g. `localhost:5000/metalk8s/nginx:1.14.2`)
+    tls_verify : True
+        Verify TLS certificate of the repository. Set to `False` if the
+        repository uses HTTP instead of HTTPS.
```
ballot-scality, line 35: An overall comment. In salt, since you can extract the docstring with 
```
salt-call  -d docker_registry
```
it's a good practice to put one or more examples in the docstring of how you use your module
```
.. code:: bash

  salt '*' docker_registry.get_image_manifest localhost:5000/metalk8s/nginx:1.14.2
```

salt/_states/docker_registry.py
```diff
@@ -0,0 +1,202 @@
...
+        Path of the Docker archive to upload, if image not available.
+    digest : None
+        Expected digest of the image. If no match and `force` is `False`, the
+        state will fail.
+    force : False
+        Force-push the image on digest mismatch, or if no digest is provided.
+    tls_verify : True
+        Verify TLS certificates when retrieving image metadata, passed to
+        `docker_registry.get_image_manifest` and
+        `docker_registry.push_docker_archive` (as `dest_tls_verify`).
```
ballot-scality, line 40: Same for the state module
you can extract the documentation with
```
salt-call sys.state_doc docker_registry.image_managed
```

salt/_modules/docker_registry.py
NicolasT: I know, hence the existing docstrings ;)

Note however these modules are not meant (right now) for any 'public consumption' outside of this project, so I don't want to go overboard with examples and code documentation either.

salt/_modules/docker_registry.py
NicolasT: Done

salt/_states/docker_registry.py
NicolasT: Applying what I remember from PEP8 in the OpenStack days. As long as we don't have automated linting, anything that doesn't look too crazy is good to go. Otherwise, de gustibus and bikeshedding.

Made a modification to the code (will be in another PR, this one is abandoned, hang on).

salt/_states/docker_registry.py
NicolasT: See prior comment.

salt/_states/docker_registry.py
NicolasT: Nope, `image_manifest` may be `None` so can't compute `digest_match` immediately after. If there's no manifest, then the digests don't necessarily not-match (which is a failure).

salt/_states/docker_registry.py
NicolasT: Agree, will do so in other PR.

NicolasT:
https://github.com/scality/metalk8s/pull/604#issuecomment-465233344

PR #614, opened 2019-02-19
Kubeadm init: kubeconfig phase

ramanduh:
Generate kubeconfig files for control plane components.

The goal of this PR is to generate 4 kubeconfig files as describe here: https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#generate-kubeconfig-files-for-control-plane-components

ie salt state implementation of `kubeadm init phase kubeconfig`

The CA cert need to be generated and advertised first with the procedure described here: https://github.com/scality/metalk8s/pull/607#issue-254014586

After that the salt state could be called with:
```
salt <client id> state.sls metalk8s.kubeadm.init.kubeconfig
```

### DIFF
--- a/salt/_states/kubeconfig.py
+++ b/salt/_states/kubeconfig.py
@@ -1,0 +1,195 @@
+# -*- coding: utf-8 -*-
+
+from base64 import b64encode, b64decode
+from datetime import datetime, timedelta
+import os
+import stat
+import yaml
+
+__virtualname__ = 'metalk8s_kubeconfig'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def _validateKubeConfig(filename,
+                        expected_ca_data,
+                        expected_api_server,
+                        expected_cn):
+    """Validate a kubeconfig filename.
+
+    Validate that the kubeconfig provided by filename
+    is conform with config.
+
+    This function is used for managed idempotency.
+
+    :return: True if the kubeconfig file matches expectation
+             False otherwise (ie need to be regenerated)
+    """
+    # Verify if the file exists
+    if not os.path.isfile(filename):
+        return False
+
+    # Verify that the mode is 600
+    if stat.S_IMODE(os.stat(filename).st_mode) != 0o600:
+        return False
+
+    try:
+        with open(filename, 'r') as fd:
+            kubeconfig = yaml.safe_load(fd)
+    except Exception:
+        return False
+
+    # Verify that the current CA cert on disk matches the expected CA cert
+    # and the API Server on the existing file match with the expected
+    try:
+        cluster_info = kubeconfig['clusters'][0]['cluster']
+        current_ca_data = cluster_info['certificate-authority-data']
+        current_api_server = cluster_info['server']
+    except (KeyError, IndexError):
+        return False
+
+    if current_ca_data != expected_ca_data:
+        return False
+
+    if current_api_server != expected_api_server:
+        return False
+
+    # Client Key and certificate verification
+    try:
+        b64_client_key = kubeconfig['users'][0]['user']['client-key-data']
+        b64_client_cert = kubeconfig['users'][0][
+            'user']['client-certificate-data']
+    except (KeyError, IndexError):
+        return False
+
+    try:
+        client_key = b64decode(b64_client_key)
+        client_cert = b64decode(b64_client_cert)
+    except TypeError:
+        return False
+
+    ca_pem_cert = b64decode(current_ca_data)
+
+    client_cert_detail = __salt__['x509.read_certificate'](client_cert)
+
+    # Verify client cn
+    try:
+        current_cn = client_cert_detail['Subject']['CN']
+    except KeyError:
+        return False
+    else:
+        if current_cn != expected_cn:
+            return False
+
+    # Verify client client cert expiration date is > 30days
+    try:
+        expiration_date = client_cert_detail['Not After']
+    except KeyError:
+        return False
+    else:
+        if datetime.strptime(expiration_date, "%Y-%m-%d %H:%M:%S") \
+                - timedelta(days=30) < datetime.now():
+            return False
+
+    if __salt__['x509.verify_signature'](
+            client_cert, ca_pem_cert) is not True:
+        return False
+
+    if __salt__['x509.verify_private_key'](
+            client_key, client_cert) is not True:
+        return False
+
+    return True
+
+
+def managed(name,
+            ca_server,
+            signing_policy,
+            client_cert_info,
+            apiserver,
+            cluster,
+            **kwargs):
+    """Generate kubeconfig file with identities for control plane components"""
+    ret = {
+        'name': name,
+        'changes': {},
+        'comment': "",
+        'result': True,
+    }
+
+    # Get the CA cert from mine
+    try:
+        b64_ca_cert = __salt__['mine.get'](
+            ca_server,
+            'kubernetes_ca_server'
+        )[ca_server]
+    except KeyError:
+        ret.update({
+            'comment':
+                '{0} CA server is not advertized in mine'.format(ca_server),
+            'result': False
+        })
+        return ret
+    else:
+        b64_ca_cert = b64_ca_cert.replace('\n', '')
+
+    user = client_cert_info.get('CN')
+
+    # Validate if a kubeconfig already exists (idempotency)
+    if _validateKubeConfig(name, b64_ca_cert, apiserver, user):
+        ret.update({'comment': 'kubeconfig file exists and is up-to-date'})
+        return ret
+
+    client_priv_key = __salt__['x509.create_private_key'](text=True)
+
+    client_cert = __salt__['x509.create_certificate'](
+        text=True,
+        public_key=client_priv_key,  # pub key is sourced from priv key
+        ca_server=ca_server,
+        signing_policy=signing_policy,
+        **client_cert_info
+    )
+
+    dataset = {
+        'apiVersion': 'v1',
+        'clusters': [
+            {
+                'cluster': {
+                    'certificate-authority-data': b64_ca_cert,
+                    'server': apiserver,
+                },
+                'name': cluster
+            }
+        ],
+        'contexts': [
+            {
+                'context': {
+                    'cluster': cluster,
+                    'user': user,
+                },
+                'name': '{0}@{1}'.format(user, cluster),
+            }
+        ],
+        'current-context': '{0}@{1}'.format(user, cluster),
+        'kind': 'Config',
+        'preferences': {},
+        'users': [
+            {
+                'name': user,
+                'user': {
+                    'client-certificate-data': b64encode(client_cert),
+                    'client-key-data': b64encode(client_priv_key)
+                }
+            }
+        ]
+    }
+
+    return __states__['file.serialize'](
+        name=name,
+        dataset=dataset,
+        formatter="yaml",
+        mode="600",
+        makedirs=True
+    )
--- a/salt/metalk8s/kubeadm/init/certs/ca.sls
+++ b/salt/metalk8s/kubeadm/init/certs/ca.sls
@@ -42,7 +42,8 @@ Advertise CA in the mine:
   module.wait:
     - mine.send:
       - func: 'kubernetes_ca_server'
-      - mine_function: test.ping
+      - mine_function: hashutil.base64_encodefile
+      - /etc/kubernetes/pki/ca.crt
     - watch:
       - x509: Generate CA certificate
 
--- a/salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
+++ b/salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
@@ -1,0 +1,5 @@
+{%- from 'metalk8s/kubeadm/init/kubeconfig/lib.sls' import kubeconfig with context %}
+
+{%- set cert_info = {'CN': 'kubernetes-admin', 'O': 'system:masters'} %}
+
+{{ kubeconfig('admin', cert_info) }}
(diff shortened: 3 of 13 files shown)


salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
```diff
@@ -0,0 +1,12 @@
+Create kubeconf file for admin:
+  metalk8s_kubeconfig.managed:
+    - name: /etc/kubernetes/admin.conf
+    - client_cert_info:
+        signing_private_key: /etc/kubernetes/pki/ca.key
+        signing_cert: /etc/kubernetes/pki/ca.crt
+        CN: kubernetes-admin
+        O: system:masters
+    - config_context:
+        master_address: https://10.200.5.53:6443
```
TeddyAndrieux: Seems really hardcoded here.

And it's not master_address it the kube-apiserver address I think.

salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
```diff
@@ -0,0 +1,12 @@
+Create kubeconf file for admin:
+  metalk8s_kubeconfig.managed:
+    - name: /etc/kubernetes/admin.conf
+    - client_cert_info:
+        signing_private_key: /etc/kubernetes/pki/ca.key
+        signing_cert: /etc/kubernetes/pki/ca.crt
+        CN: kubernetes-admin
+        O: system:masters
```
TeddyAndrieux: For me these value should be part of defaults.yaml

TeddyAndrieux (changes requested):

salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
ramanduh: oups, will change it

salt/_states/kubeconfig.py
```diff
@@ -0,0 +1,44 @@
...
+__virtualname__ = 'metalk8s_kubeconfig'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def managed(name, client_cert_info, config_context, **kwargs):
+    """Generate kubeconfig file with identities for control plane components"""
+    # TODO: manage idempotency
```
NicolasT: This is required for this PR to land, as far as I'm concerned.

salt/_states/kubeconfig.py
```diff
@@ -0,0 +1,44 @@
...
+
+    client_cert = __salt__['x509.create_certificate'](
+        text=True,
+        public_key=client_priv_key,
+        **client_cert_info
+    )  # TODO; use ca server
+
+    ca_cert = client_cert_info.get('signing_cert')
+
+    with open(ca_cert, 'r') as fd:
```
NicolasT: Can we use a `__salt__` function to achieve this?

salt/metalk8s/kubeadm/init/kubeconfig/files/kubeconfig.tpl
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
NicolasT: Could use `file.serialilze` in your state, instead of a template.

NicolasT (dismissed):
See comments.

salt/_states/kubeconfig.py
ramanduh: Yes, It is required for this PR

salt/_states/kubeconfig.py
ramanduh: Ok, I will do. For the ca_cert, I'm thinking about storing a pem version of the ca cert in the pillar (base64), in order to spread it to all minion, so maybe I will drop this part, but still waiting for Teddy's PR for now to test

salt/metalk8s/kubeadm/init/kubeconfig/files/kubeconfig.tpl
ramanduh: OK, I will do

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
```diff
@@ -0,0 +1,15 @@
...
+  metalk8s_kubeconfig.managed:
+    - name: {{ conf.file }}
+    - client_cert_info:
+        signing_private_key: /etc/kubernetes/pki/ca.key
+        signing_cert: /etc/kubernetes/pki/ca.crt
+        {%- for key, value in conf.cert.items() %}
+        {{ key }}: {{ value }}
+        {%- endfor %}
+    - config_context:
+        kube_apiserver: https://10.200.5.53:6443
```
ramanduh: Will update the api server address

salt/_states/kubeconfig.py
TeddyAndrieux: Not in the pillar maybe in the mine it's better

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
TeddyAndrieux: Suggest something like `https://{{ salt['network.ip_addrs'](cidr=network.control_plane)[0] }}:6443` (network.control_plane is part of the pillar in my PR)

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
```diff
@@ -0,0 +1,28 @@
+{%- from "metalk8s/map.jinja" import defaults with context %}
+
+{% macro kubeconfig(name, cert_info) %}
+
+{%- set ca_server = salt['mine.get']('*', 'kubernetes_ca_server').keys() %}
+{%- set apiserver = 'https://' ~ salt['grains.get']('master') ~ ':6443' %}
```
TeddyAndrieux, line 7: Should be configurable. But maybe in another PR/ticket

ramanduh: Yep, agree, I thought about something in pillar, but as we don't have the generate pillar part yet I let it as it was and if it's ok for you, we'll create another ticket

TeddyAndrieux: Yes, you should create a ticket for this :)

salt/metalk8s/kubeadm/init/certs/ca.sls
```diff
@@ -63,6 +64,14 @@ Create CA salt signing_policies:
             - keyUsage: "critical digitalSignature, keyEncipherment"
             - extendedKeyUsage: "serverAuth"
             - days_valid: {{ ca.signing_policy.days_valid }}
+          kube_user_account_policy:
```
TeddyAndrieux: I also created this in my PR but lets merge yours first I will update mine

ramanduh: OK

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
```diff
@@ -0,0 +1,7 @@
+{%- from 'metalk8s/kubeadm/init/kubeconfig/lib.sls' import kubeconfig with context %}
+
+{%- set hostname = salt.network.get_hostname() %}
```
TeddyAndrieux, line 3: I don't know if we want the hostname or the fqdn for me it's more the fqdn

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
ramanduh: Done: https://github.com/scality/metalk8s/issues/644

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
ramanduh: In fact the default value set by kubeadm is the hostname provided by the `hostname` cmd: https://kubernetes.io/docs/reference/access-authn-authz/node/
But in fact the hostname should be a fqdn so it should be ok

salt/_states/kubeconfig.py
```diff
@@ -0,0 +1,146 @@
...
+        return False
+
+    # Verify that the current CA cert on disk matches the expected CA cert
+    # and the API Server on the existing file match with the expected
+    try:
+        cluster_info = kubeconfig['clusters'][0]['cluster']
+        current_ca_data = cluster_info['certificate-authority-data']
+        current_api_server = cluster_info['server']
+    except (KeyError, IndexError):
+        return False
```
TeddyAndrieux, line 51: You also need to verify `client-certificate-data` because if it's a wrong certificate or certificate expired you need to generate one new but keep the same private key I think.

So for me we need something like :
```
if no private_key: 
  # generate_one
else:
  # get the private key

if certificate:
  # Verify certs field like CN, ...
  # Verify that certs signed by the CA public key (x509.verify_signature)
  # Verify if certs expired or almost expired (x509.will_expire)
  # Verifiy that private key in the certs if the one used in the certificate (x509.verify_private_key, you can use directly the certs as public key)
  if something wrong:
    certificate = None

if not certificate:
  # generate one
```

salt/_states/kubeconfig.py
```diff
@@ -0,0 +1,146 @@
...
+    ret = {
+        'name': name,
+        'changes': {},
+        'comment': "",
+        'result': True,
+    }
+
+    # Get the CA cert from mine
+    try:
+        b64_ca_cert = __salt__['mine.get'](
```
TeddyAndrieux, line 124: After checking salt command I suggest to use `x509.get_signing_policy` to get the CA certificate I think it's better

TeddyAndrieux (changes requested):

salt/_states/kubeconfig.py
ramanduh: Yep, you're right it's better, I will change it and I don't even need to store the certif in the mine

salt/_states/kubeconfig.py
ramanduh: Yep, we can do that but it's really depends on our policy. I did it like kubeadm do and it does not verify the `client-certificate-data`, when you run `kubeadm init  phase kubeconfig kubelet`, and you put a random base64 string in `client-certificate-data`, if you re-run the `kubeadm init  phase kubeconfig kubelet`, it will say that the config if up to date. I will see if the change is not too long, I will do it, otherwise, I will create another ticket

TeddyAndrieux: Yes because kubeadm is definitly not idempotent. But us we should be.
But btw yes maybe in another PR/tcket

salt/_states/kubeconfig.py
ramanduh: Seen together but does not work actually as `x509.get_signing_policy` return only the result on the salt master but not the minion, so we can't query the ca cert from minion with `x509.get_signing_policy`

salt/_states/kubeconfig.py
ramanduh: In fact, if the file exists, kubeadm check if it is valid, and for his check, it checks only the ca cert, and the API srv.
```
[root@k8s centos]# kubeadm init  phase kubeconfig kubelet
[kubeconfig] Using existing up-to-date kubeconfig file: "/etc/kubernetes/kubelet.conf"
```
I just updated the PR. I took into account all of your remarks except the keep the private key. AFAIK, we don't have a real reason to keep the same priv key as we don't store it. For the cert expiration check, I did not use x509 will expire module because it does not work on pem cert. The docstring says `string containing the PEM formatted text of the certificate.` but it's not implemented in the code

TeddyAndrieux (approved):
LGTM

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
```diff
@@ -0,0 +1,28 @@
...
+
+{%- if ca_server %}
+
+Create kubeconf file for {{ name }}:
+  metalk8s_kubeconfig.managed:
+    - name: /etc/kubernetes/{{ name }}.conf
+    - ca_server: {{ ca_server[0] }}
+    - signing_policy: kube_user_account_policy
```
TeddyAndrieux: Seems like my PR will get merged first you will need to change this to use `kube_api.cert.client_signing_policy` value from map.jinja

TeddyAndrieux (changes requested):

salt/metalk8s/kubeadm/init/kubeconfig/lib.sls
```diff
@@ -0,0 +1,28 @@
...
+{%- set ca_server = salt['mine.get']('*', 'kubernetes_ca_server').keys() %}
+{%- set apiserver = 'https://' ~ salt['grains.get']('master') ~ ':6443' %}
+
+{%- if ca_server %}
+
+Create kubeconf file for {{ name }}:
+  metalk8s_kubeconfig.managed:
+    - name: /etc/kubernetes/{{ name }}.conf
+    - ca_server: {{ ca_server[0] }}
+    - signing_policy: kube_apiserver_client_policy
```
TeddyAndrieux: Use the defaults.yaml value `kube_api.cert.client_signing_policy`

salt/metalk8s/kubeadm/init/kubeconfig/admin.sls
charlesprost: @ramanduh can we resolve this one?

salt/_states/kubeconfig.py
charlesprost: @ramanduh is this solved?

salt/metalk8s/kubeadm/init/kubeconfig/files/kubeconfig.tpl
charlesprost: @ramanduh is this solved?

TeddyAndrieux (approved):

ramanduh:
@bert-e approve

PR #621, opened 2019-02-20
Add a ticket creation template

sayf-eddine-scality:
This PR will create a template to fulfill when adding a new issue


jbertran (dismissed):
I think we can merge it as is, and submit PRs on it down the road if we need more.

TeddyAndrieux (dismissed):

NicolasT (changes requested):
I would like to take a somewhat different approach.

Github allows to have multiple templates, and will present a choice in the UI when creating an issue. So, let's do this from the start.

This template seems related to 'improvements', 'feature requests',... Let's make that clear, from the title of the template (i.e. `name` field). Also, let's set up some default `labels`, e.g. `enhancement` and `moonshot`.

Also, I'd like all 'example' text (like `Short description`, the 'dependencies' list,...) to be commented-out by default. This eases filling in the template, not requiring to delete the boilerplate.

Finally, 'dependencies' should go at the very bottom, definitely after 'requirements' and 'delivery'.

Issue #557, opened 2019-02-06
CI: Bootstrap automated platform tests entrypoint


jbertran:
~Using eve, launch platform tests on the deployed node. A few basic tests (no more complex than "`kubectl get nodes` returns a node") will suffice.~

Low-hanging fruit: tackle launching lint tests for now, until we are ready to launch a build and start working with other teams to figure out what they wish to test.

jbertran:
Closed by #627

PR #627, opened 2019-02-20
Improvement/557 bootstrap lint test step in eve

jbertran:
~Based on @sayf-eddine-scality 's PR for #596 - will be rebased to only include my commits as soon as that PR is merged.~

__Fixes:__ #557

#### WHAT

Bootstrap ~what will become the unit test stage in Eve.~ From discussing with @gdemonet : let's keep lint and unit tests separate.

* For now, it only contains Python linting so files, steps and targets are named accordingly. 
* Currently no `rpmlint` due to the lack of coooperation from that tool's CentOS7 version with GitHub release artifacts

#### WHY

Get things started, provide a unified way of dealing with tests and step additions.

#### TO DO

* Tests currently in `flunkOnFailure: False`. PR for switching to True by fixing the existing issues coming soon.
* ~Make unit tests and rename the steps/targets/files accordingly~
* Strategy for rpmlint to be formulated
* E2E tests and static analysis to be added in a separate Eve step

#### Commits

* ci: add top level tox.ini, bootstrap tox lint target for Salt
* ci: bootstrap sub-makefile directory, add lint target
* ci: have a separate eve build step triggered by local worker
* ci: add eve lint step to pre-merge step


NicolasT:
FWIW, we already run `rpmlint` as part of the RPM building process.

jbertran:
> FWIW, we already run `rpmlint` as part of the RPM building process.

It runs but in the CentOS7 version we run in the build and test containers (1.5, 6 years old :C), it gives out `invalid-url` warnings that I'd like to mark as failing.

I spent an hour or so trying to work out if I could at least ignore the specific 403 warning caused, but that thing is __ancient__ so I ended up dropping it.

make/lint.mk
```diff
@@ -0,0 +1,9 @@
+# Lint targets for MetalK8S environment
+# Requirements for linting on host:
+#    - python3
+#    - tox
+
+lint-local:
```
gdemonet: Just wondering, do we expect another kind of `lint-<something>` target? Maybe `lint` could be enough

eve/main.yml
```diff
@@ -30,3 +41,17 @@ stages:
         urls:
           - "*.iso"
           - "SHA256SUM"
+
+  lint:
+    worker:
+      type: kube_pod
```
gdemonet, line 47: Why not go with a simpler `docker` type worker? Ditto for build step by the way

eve/main.yml
```diff
@@ -30,3 +41,17 @@ stages:
...
+      path: eve/workers/pod-linter/pod.yaml
+      images:
+        docker-linter: eve/workers/pod-linter
+    steps:
+    - Git: *git_pull
+    - ShellCommand:
+        name: linting steps
```
gdemonet: I think I read somewhere we prefer complete sentences, maybe `Run all linting steps`

eve/main.yml
```diff
@@ -5,7 +5,18 @@ branches:
     stage: pre-merge
 
 stages:
+
   pre-merge:
+    worker:
+      type: local
+    steps:
+      - TriggerStages:
+          name: trigger build and lint stages simultaneously
```
gdemonet: _minor_: Capitalize step name `Trigger build and...`

gdemonet (commented):
Few questions/remarks, looks good to me overall, would need to test I guess...

eve/workers/pod-linter/pod.yaml
```diff
@@ -0,0 +1,22 @@
...
+        cpu: 500m
+        memory: 1Gi
+      limits:
+        cpu: "1"
+        memory: 2Gi
+    securityContext:
+      capabilities:
+        drop:
+        - ALL
+      runAsUSer: 1000
```
sayf-eddine-scality: we have set up eve as the internal user no ? and the capabilities isnt it  by default dropped ?

eve/main.yml
jbertran: For the build step: I believe we can't do docker-in-docker directly in a `docker` worker, but we want to maintain 1:1 between local usage and CI usage, so we build the same way in both (!= RING in that regard).

For the linting step: no argument against. I don't foresee the need for a pod, so I can switch it to docker. We need to discuss more about the lint/test separation.

make/lint.mk
jbertran: Originally I wanted to have an `rpmlint` run included, so my idea was to provide a target assuming dependencies first, then a target running this in a docker so a user wouldn't have to install `rpmlint`. I'll rename it for now, if we revisit that `rpmlint` stuff we can update this at that time.

eve/workers/pod-linter/pod.yaml
jbertran: No idea, that's straight from the metal 1.3 pod file. I'll take a closer look at it.

ballot-scality:
FYI, @jbertran @gdemonet , regarding all Saltstack related stuff, 2 source of information : 
* https://docs.saltstack.com/en/latest/topics/development/conventions/style.html 
* https://github.com/saltstack/salt-pylint

jbertran:
> FYI, @jbertran @gdemonet , regarding all Saltstack related stuff, 2 source of information :
> 
>     * https://docs.saltstack.com/en/latest/topics/development/conventions/style.html
> 
>     * https://github.com/saltstack/salt-pylint

I guess we're switching to Pylint! The plugin interestingly doesn't add `__salt__` or `__opts__` to builtins, but it does do a number of useful checks, thanks.

ballot-scality:
> > FYI, @jbertran @gdemonet , regarding all Saltstack related stuff, 2 source of information :
> > ```
> > * https://docs.saltstack.com/en/latest/topics/development/conventions/style.html
> > 
> > * https://github.com/saltstack/salt-pylint
> > ```
> 
> I guess we're switching to Pylint! The plugin interestingly doesn't add `__salt__` or `__opts__` to builtins, but it does do a number of useful checks, thanks.

I think at least for everything under "salt" directory, it makes sense. IIRC, __salt__ is added as ignored in the .pylintrc provided through the various  "(__.*__)"

eve/main.yml
```diff
@@ -30,3 +41,18 @@ stages:
         urls:
           - "*.iso"
           - "SHA256SUM"
+
+  lint:
+    worker:
+      type: docker
```
tcarmet: tip: try not to use worker of type `docker` anymore! and only use `kube_pod`

eve/workers/pod-linter/pod.yaml
Zempashi: This is because running things as root is ugly. Yes the image setup  the eve-worker manually to use 1000 user (https://github.com/scality/metalk8s/blob/development/1.3/eve/images/pep8-checker/Dockerfile#L13). You can't pass username (just uid) in kubernetes yaml, as far as I know.

eve/main.yml
Zempashi: Yes please, you have sample here that were already do some linting: https://github.com/scality/metalk8s/blob/development/1.3/eve/main.yml#L180-L193

tcarmet: Well actually there's something I'm not understanding. On one side you're setting your worker of type `docker` on the other side you have a `pod` description file [here](https://github.com/scality/metalk8s/pull/627/files#diff-e86b442112f809737963abcac69fe777) which is meant for worker of type `kube_pod`.

I'm confused

jbertran: It was originally a pod, then I took into account a comment while unaware that `kube-pod` is the way to do it, and I haven't committed the new directory structure + file deletion. Reverting to a pod will be quick :)

tcarmet: okkk good good!

gdemonet: > tip: try not to use worker of type `docker` anymore! and only use `kube_pod`

Yup my bad, I was the one that requested the change (it felt unnecessary to have access to Kube for a simple linting step). Maybe you could add a **[deprecated]** mention in Eve docs for that worker type @tcarmet @Zempashi (in case other developers miss that info about preferring `kube_pod`)?

tcarmet: very good point @gdemonet ! [Here you go](https://scality.atlassian.net/browse/EVE-1036)

charlesprost (approved):

jbertran:
@bert-e approve

Issue #561, opened 2019-02-07
Define a Centos7.6 base image that will work for everyone


charlesprost:
We need to identify which Docker image should be the one on which we base everything from now on.

Let's identify it with its digest to be sure.

Let's be sure this image is identified in one place so any evolution should be propagated everywhere.

NicolasT:
This is not just a MetalK8s concern, of course. Don't spend too much time on this topic right now, once we figure out the right technical approach company-wide, any images built as part of MetalK8s shall abide by these rules.

jbertran:
Topic taken internally by release engineering.

PR #628, opened 2019-02-21
Improvement/centralize docker base image

sayf-eddine-scality:
1- Centralize base image for the dockerfiles
Fixes: #561 
2- Bonus: Fix the make help menu


images/salt-master/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:7.6.1810
+FROM base-centos-image
 
 MAINTAINER moonshot-platform <moonshot-platform@scality.com>
```
jbertran, line 3: ~Gonna need that email address now !~ I don't often look at emails.

sayf-eddine-scality: we dont have it already ?

images/salt-master/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:7.6.1810
+FROM base-centos-image
```
NicolasT, line 1: Pass this in as an `ARG`. That way, we can have the 'base image' name based on a name/tag that changes when the base image changes. For tracking purposes.

NicolasT: Better even, retrieve the hash of the base image and use that in FROM.

packages/Dockerfile
```diff
@@ -1,8 +1,4 @@
-# Equal to centos:7.6.1810
-# SHA256 digest of the base image
-ARG BUILD_IMAGE_SHA256=5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121
-ARG BUILD_IMAGE=docker.io/centos
-FROM ${BUILD_IMAGE}@sha256:${BUILD_IMAGE_SHA256} as build
+FROM base-centos-image as build
```
NicolasT, line 1: Same.

images/base/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Equal to centos:7.6.1810
+# SHA256 digest of the base image
+ARG BUILD_IMAGE_SHA256=5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121
```
NicolasT: Can we not have a default, and have this injected by the `Makefile`?

Makefile
```diff
@@ -165,9 +166,14 @@ vagrantup: $(ALL) ## Run 'vagrant up' to (re-)provision a development environmen
 help: ## Show this help message
 	@echo "The following targets are available:"
 	@echo
-	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+	@grep -Eh '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 .PHONY: help
 
+packages/Dockerfile: $(BASE_CENTOS_IMAGE)
+images/salt-master/Dockerfile: $(BASE_CENTOS_IMAGE)
+$(BASE_CENTOS_IMAGE):
```
NicolasT, line 181: This depends on the image's `Dockerfile`, and the value of the base digest etc.

images/salt-master/Dockerfile
ballot-scality, line 3: MAINTAINER is deprecated, please don't use it :) https://docs.docker.com/engine/reference/builder/#maintainer-deprecated
Use a LABEL instead

Makefile
```diff
@@ -165,11 +169,23 @@ vagrantup: $(ALL) ## Run 'vagrant up' to (re-)provision a development environmen
...
 
+.PHONY: basecentos-vars
+basecentos-vars:
+	@if [[ `cat $(BASE_CENTOS_IMAGE) 2>&1` != $(BASE_CENTOS_IMAGE_NAME)$(BASE_CENTOS_IMAGE_SHA256) ]]; then \
+        echo -n $(BASE_CENTOS_IMAGE_NAME)$(BASE_CENTOS_IMAGE_SHA256) > $(BASE_CENTOS_IMAGE) ; \
+    fi
+
+$(BASE_CENTOS_IMAGE):
+	docker build \
+		-t base-centos-image \
```
ballot-scality, line 183: Weird indent :)

Makefile
```diff
@@ -165,11 +169,23 @@ vagrantup: $(ALL) ## Run 'vagrant up' to (re-)provision a development environmen
...
+        --build-arg BUILD_IMAGE=$(BASE_CENTOS_IMAGE_NAME) \
+		--build-arg BUILD_IMAGE_SHA256=$(BASE_CENTOS_IMAGE_SHA256) \
+		images/base
 
-$(BUILD_ROOT)/package-build-container: packages/Dockerfile packages/entrypoint.sh
+$(BUILD_ROOT)/package-build-container: packages/Dockerfile packages/entrypoint.sh $(BASE_CENTOS_IMAGE) | basecentos-vars
```
ballot-scality, line 188: The usage of the "order only prerequisite" seems odd. You usually don't want to use a PHONY target for that (https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html)

Makefile
```diff
@@ -284,7 +300,9 @@ $(ISO_ROOT)/images/$(NGINX_IMAGE_NAME)-$(NGINX_IMAGE_VERSION).tar.gz: \
 $(ISO_ROOT)/images/$(NGINX_IMAGE_NAME)-$(NGINX_IMAGE_VERSION).tar.gz: \
 	IMAGE_TAG=$(NGINX_IMAGE_NAME):$(NGINX_IMAGE_VERSION)
 $(ISO_ROOT)/images/$(SALT_MASTER_IMAGE_NAME)-$(SALT_MASTER_IMAGE_VERSION).tar.gz: \
-	images/salt-master/Dockerfile
+	images/salt-master/Dockerfile \
+	$(BASE_CENTOS_IMAGE)
+$(ISO_ROOT)/images/$(SALT_MASTER_IMAGE_NAME)-$(SALT_MASTER_IMAGE_VERSION).tar.gz: | basecentos-vars
```
ballot-scality, line 305: Same

ballot-scality (commented):
As a rule of thumb, when the result of a command cannot really be "idem potent", I tend to do something like this

```
.PHONY: myphonytarget
myphonytarget: $(build_dir)/.stab_myphonytarget

$(build_dir)/.stab_myphonytarget: $(my_target_I_depend_on) | $(build_dir)
     @mycommands
     touch -m $@
```

This way my phony is always executed, but the real deal is only if the right dependency is OK.
Maybe that's a bad habit though, so I think this is open for debate.

On a side note, I feel like Makefile seems to be less and less appropriate here. Maybe we should consider switching to something else while it's still lfresh.

slaperche-scality:
> On a side note, I feel like Makefile seems to be less and less appropriate here. Maybe we should consider switching to something else while it's stilll fresh.

+1

Any suggestions?

I could propose Rake, but it's Ruby-based and we mainly have Pythonistas here.

Maybe Scons can be a good alternative (you can define custom [builders](https://scons.org/doc/1.3.1/HTML/scons-user.html#chap-builders-writing) and/or [scanners](https://scons.org/doc/1.3.1/HTML/scons-user.html#chap-scanners) (to auto-detect dependencies)), but I don't have that much experience with it so I can't guarantee that it fits the bill.

NicolasT:
@bert-e wait

NicolasT:
I'd like to hold off on this one a little, until we clarified with RELENG what the whole chain of image building, base images,... will be across our projects. Ping @tcarmet @flavien-scality

sayf-eddine-scality:
Once the DoIt PR lands this should be reimplemented in python

Issue #635, opened 2019-02-22
Generic state to restart salt-minion


TeddyAndrieux:
This kind of state should be something generic available upper in the formula (like metalk8s.salt.minion) and be included.

Especially with an ID like this, we risk collision and this is a "trick" (using bg and stuff) that can be missed by someone so I would rather a central place to control the restart of the minion and do a watch_in in this state

_Originally posted by @ballot-scality in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MjU0MDE0NTg2/pull_requests/unread_timeline_

Issue #633, opened 2019-02-22
Use dashes in the filename instead of underscores


TeddyAndrieux:
Consider using dashes in the filename instead of underscores for uniformity.

_Originally posted by @NicolasT in https://github.com/scality/metalk8s/pull/607_

PR #638, opened 2019-02-22
Improvement/saltify kubeadm certs phase root ca

TeddyAndrieux:
Add a generic state to restart salt-minion
Certs generation for apiserver-kubelet-client and SA private and public key.
```
salt-call state.sls metalk8s.kubeadm.init.certs.apiserver-kubelet-client
salt-call state.sls metalk8s.kubeadm.init.certs.sa
```

Fixes: #557 #633 #635

### DIFF
--- a/salt/metalk8s/kubeadm/init/certs/sa.sls
+++ b/salt/metalk8s/kubeadm/init/certs/sa.sls
@@ -1,0 +1,26 @@
+include:
+  - .installed
+
+Create SA private key:
+  x509.private_key_managed:
+    - name: /etc/kubernetes/pki/sa.key
+    - bits: 2048
+    - user: root
+    - group: root
+    - mode: 600
+    - makedirs: True
+    - dir_mode: 755
+    - require:
+      - pkg: Install m2crypto
+
+Store SA public key:
+  file.managed:
+    - name: /etc/kubernetes/pki/sa.pub
+    - contents: __slot__:salt:x509.get_public_key("/etc/kubernetes/pki/sa.key")
+    - user: root
+    - group: root
+    - mode: 600
+    - makedirs: True
+    - dir_mode: 755
+    - require:
+      - x509: Create SA private key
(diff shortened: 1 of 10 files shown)


salt/metalk8s/kubeadm/init/certs/sa.sls
```diff
@@ -0,0 +1,26 @@
+include:
+  - .installed
+
+Create SA private key:
+  x509.private_key_managed:
+    - name: /etc/kubernetes/pki/sa.key
+    - bits: 2048
```
charlesprost, line 7: We may want to customize this as we did for the other one?

charlesprost (dismissed):

salt/metalk8s/kubeadm/init/certs/sa.sls
NicolasT: Meh, do we? Should it be configurable at other occasions?

Here's a rule of thumb:
- CA and intermediate CAs: 4096 bit RSA
- Server or client cert: 2048 bit RSA, or also 4096 (though with additional overhead)

From Wikipedia:

> The NIST recommends 2048-bit keys for RSA. [15] An RSA key length of 3072 bits should be used if security is required beyond 2030.

Over time, using EC could be worthwhile but that's not supported by K8s components right now.

Anyway, in general I don't believe these values should be configurable. We should have defaults which are high enough to satisfy any customer (or which get bumped if some customer has good reasons why they should be), and then be used across the board. There should be *no* deployments where the key size is *weakened*.

TeddyAndrieux: Btw kubeadm generate only 2048 key even for CA.
But maybe we could change to 4096

NicolasT: Tracking ticket pls :)

TeddyAndrieux: Oh ! too late i pushed it :p 
4096 for CA 2048 for others (not configurable)

charlesprost: Yeah, that's good then.

charlesprost (approved):

TeddyAndrieux:
@bert-e approve

PR #643, opened 2019-02-25
587 - fetch packages

alexandre-allard:


sayf-eddine-scality:
what about gpg keys ?

packages.mk
```diff
@@ -0,0 +1,7 @@
+YUM_PACKAGES = \
```
charlesprost, line 1: Hum we probably want to freeze the versions, otherwise running the `make` command a few days later may download different packages, and that's probably not what we want.

NicolasT:
> what about gpg keys ?

Agree, having those offline (maybe even contained in this repository?) would be necessary.

packages.mk
charlesprost: @alexandre-allard-scality @NicolasT what about freezing the package versions? is this a good/bad idea?

NicolasT: @ballot-scality ^^

packages/yum_repositories/docker-ce.repo
```diff
@@ -0,0 +1,6 @@
+[docker-ce]
```
NicolasT: Why would we need Docker repositories?

packages.mk
ballot-scality: Given the very short list, I would say yes, freezing is a good idea.
However, we require a yum plugin for that on our system then : https://access.redhat.com/solutions/98873
We also need to freeze them on the system

packages/yum_repositories/docker-ce.repo
ballot-scality: Indeed. No need for this one AFAIK

alexandre-allard: Oops, leftover, in the beginning I thought I'll use it for containerd but after i saw that it's in epel. I'll change it.

NicolasT:
Let's get this one in and focus on having this repo used on the node, and create a tracking ticket for package version pinning and move that to `-alpha2`.

packages.mk
NicolasT: #666

NicolasT (dismissed):
Didn't test, but LGTM. May use some cleanups/simplifications later.

alexandre-allard:
@bert-e approve

sayf-eddine-scality (dismissed):

tcarmet:
@bert-e after_pull_request=669

tox.ini
```diff
@@ -9,11 +9,10 @@ setenv =
     VIRTUALENV_NO_DOWNLOAD=1
 
 [testenv:lint-shell]
-deps =
-    bashate==0.6.0
+whitelist_externals += shellcheck
```
jbertran, line 12: Does tox complain without that? The inherited bash should be enough since it's the only one we call directly.

jbertran (approved):

tox.ini
alexandre-allard: Tox is ok with that, but you're right it's not necessary, I think.

Issue #556, opened 2019-02-06
CI: bootstrap automated deployment entrypoint


jbertran:
In eve. use the output of the build steps for deploying on VMs. The pros and cons of using the worker as the deployment site of the first node (see 1.x) should be considered.

gdemonet:
Blocked by #634 - need the `product_version` build property to know where to mount the ISO.

jbertran:
Closed by #652

PR #647, opened 2019-02-25
GH-556: Bootstrap deployment in Eve

gdemonet:
Using results from the build step, we mount the ISO and execute the bootstrap script on a fresh OpenStack VM worker (using CentOS 7).

Fixes GH-556.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -45,6 +45,10 @@ stages:
         urls:
           - "*.iso"
           - "SHA256SUM"
+    - SetPropertyFromCommand:
+        name: Set short version property
+        property: metalk8s-short-version
+        command: source _build/root/product.txt && echo $SHORT_VERSION
 
   lint:
     worker:
@@ -87,10 +91,10 @@ stages:
         name: Mount ISO image
         command: >
           sudo mount -o loop metalk8s.iso
-          /srv/scality/metalk8s-%(prop:product_version)s
+          /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
         haltOnFailure: true
     - ShellCommand:
         name: Start the bootstrap process
         command: >
-          bash /srv/scality/metalk8s-%(prop:product_version)s/bootstrap.sh
+          bash /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
         haltOnFailure: true
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -93,6 +93,10 @@ stages:
         command: sha256sum -c SHA256SUM
         haltOnFailure: true
     - ShellCommand:
+        name: Create mountpoint
+        command: >
+          sudo mkdir -p /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+    - ShellCommand:
         name: Mount ISO image
         command: >
           sudo mount -o loop metalk8s.iso
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -105,5 +105,6 @@ stages:
     - ShellCommand:
         name: Start the bootstrap process
         command: >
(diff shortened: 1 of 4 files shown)


eve/workers/openstack-single-node/requirements.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+yum install -y curl \
```
ballot-scality: Trailing backslash

gdemonet: Yup I did that on purpose so we can add dependencies to this list without creating unnecessary diffs when adding the backslash at that time. I can remove it though, I have absolutely no preference in terms of style guides in Shell.

eve/main.yml
```diff
@@ -55,3 +68,38 @@ stages:
...
+    - ShellCommand:
+        name: Mount ISO image
+        command: >
+          mount -o loop metalk8s.iso
+          /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+        haltOnFailure: true
+    - ShellCommand:
+        name: Start the bootstrap process
+        command: >
+            bash /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
```
tcarmet: too many spaces?

eve/main.yml
```diff
@@ -55,3 +68,38 @@ stages:
         command: make lint
         flunkOnFailure: false
         haltOnFailure: false
+
+  single-node:
+    worker:
+      type: openstack
+      image: CentOS 7 (PVHVM)
+      flavor: io1-30
```
tcarmet, line 76: `io1-30`? any particular reason why?

gdemonet: Nope, copy-pasted from metalk8s/1.3. Any advice?

eve/workers/openstack-single-node/requirements.sh
alexandre-allard: If you want to achieve this, you should rather create an array containing package names, this way the diff will be ok.
e.g.:
```
declare -a PACKAGES=(
    curl
)
```
Then, you do:
```
yum install -y "${PACKAGES[@]}"
```

gdemonet: Oh that looks good! Thanks for the tip, let me use it :)

eve/main.yml
tcarmet: no advice, but more like a question, how many ressources do you think you need?

This flavor gives you a VM with 30Gb of memory. If you think that this is what you need, go for it. Otherwise I suggest you select another flavor.

sayf-eddine-scality: platform9 flavors are supported here too no ?

eve/main.yml
```diff
@@ -55,3 +68,43 @@ stages:
...
+        command: >
+          sudo mount -o loop metalk8s.iso
+          /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+        haltOnFailure: true
+    - ShellCommand:
+        name: Start the bootstrap process
+        command: >
+          sudo bash
+          /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
+        haltOnFailure: true
```
tcarmet, line 110: if you want colors and all on the build output you can set this option `usePTY: true` to your `ShellCommand`

gdemonet:
Branch name was inappropriate for Bert-E, new PR at #652

eve/main.yml
gdemonet: Done (c7347ac5b822a1f5b4117c2602f5d342f293cdcd)

eve/main.yml
gdemonet: Chose `m1.medium` (9bc5a0c02c36f276a0c87d60afcd7a5bce352121)

PR #648, opened 2019-02-25
CI: blocking lint and more lint targets

jbertran:
#### WHAT

Make lint blocking and fix python linting issues.

Add linting targets:
* YAML
* Shell

#### WHY

Blocking lint tests is good. So is more linters.

### DIFF
--- a/make/lint.mk
+++ b/make/lint.mk
@@ -4,6 +4,5 @@
 #    - tox
 
 lint:
-	tox -e lint-salt
 
-.PHONY: lint-local
+.PHONY: lint
--- a/make/lint.mk
+++ b/make/lint.mk
@@ -5,5 +5,6 @@
 
 lint:
 	tox -e lint-yaml
+	tox -e lint-shell
 
 .PHONY: lint
--- a/make/lint.mk
+++ b/make/lint.mk
@@ -4,5 +4,6 @@
 #    - tox
 
 lint:
+	tox -e lint-yaml
 
 .PHONY: lint
--- a/salt/_states/containerd.py
+++ b/salt/_states/containerd.py
@@ -115,4 +115,4 @@ def image_managed(name, archive_path=None):
         else:
             ret['comment'] = 'Failed to pull image'
 
-        return ret
+    return ret
--- a/tox.ini
+++ b/tox.ini
@@ -2,10 +2,15 @@
 skipsdist = true
 
 [testenv]
+whitelist_externals = bash
 basepython = python3.6
 skip_install = true
 setenv =
     VIRTUALENV_NO_DOWNLOAD=1
 
+[testenv:lint-shell]
 deps =
+    bashate==0.6.0
 commands =
+    bash -c "bashate **/*.sh"
+
--- a/tox.ini
+++ b/tox.ini
@@ -13,6 +13,7 @@ deps =
     bashate==0.6.0
 commands =
     bash -c "bashate **/*.sh"
+    bash -c "bashate -i E006 **/*.sh.in"
 
 [testenv:lint-yaml]
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -7,14 +7,5 @@ skip_install = true
 setenv =
     VIRTUALENV_NO_DOWNLOAD=1
 
-[testenv:lint-salt]
-whitelist_externals = bash
-basepython = python3.6
-skip_install = true
 deps =
-    -r{toxinidir}/lint-requirements.txt
 commands =
-    bash -c "pylint {posargs} \
-    --persistent=n \
-    --additional-builtins='__salt__,__opts__,__grains__,__pillar__,__context__' \
-    salt/**/*.py"
--- a/tox.ini
+++ b/tox.ini
@@ -13,8 +13,6 @@ basepython = python3.6
 skip_install = true
 deps =
     -r{toxinidir}/lint-requirements.txt
-    SaltPyLint==2019.1.11
-    salt==2018.3.3
 commands =
     bash -c "pylint {posargs} \
     --persistent=n \
--- a/tox.ini
+++ b/tox.ini
@@ -14,3 +14,8 @@ deps =
 commands =
     bash -c "bashate **/*.sh"
 
+[testenv:lint-yaml]
+deps =
+    yamllint==1.15.0
+commands =
+    bash -c "yamllint eve/main.yml salt/**/*.yaml"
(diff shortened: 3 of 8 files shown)


tox.ini
```diff
@@ -16,6 +16,15 @@ deps =
 commands =
     bash -c "bashate **/*.sh"
 
+[testenv:lint-yaml]
+whitelist_externals = bash
+basepython = python3.6
```
slaperche-scality: Do you need to specify `basepython`? It's not inherited from `testenv`?

slaperche-scality (dismissed):

salt/_states/containerd.py
```diff
@@ -114,5 +90,45 @@ def image_managed(name, archive_path=None):
             ret['result'] = True
         else:
             ret['comment'] = 'Failed to pull image'
+        return ret
+
+    return ret
```
charlesprost, line 118: Maybe `return ret` only once?

salt/_states/containerd.py
```diff
@@ -114,5 +90,45 @@ def image_managed(name, archive_path=None):
...
+
 
+def manage_archive_path(name, archive_path, ret):
+    '''Pull image from archive at the specified path
```
charlesprost: Missing endpoint?

salt/_states/containerd.py
```diff
@@ -114,5 +90,45 @@ def image_managed(name, archive_path=None):
...
+
+    name
+        Tag or digest of the image
+    archive_path
+        path to the local Docker archive of the image to load if
```
charlesprost: capitalization of `path`?

salt/_states/containerd.py
```diff
@@ -114,5 +90,45 @@ def image_managed(name, archive_path=None):
...
+                'old': {},
+                'new': result,
+            },
+        })
+        ret['comment'] = 'Imported archive'
+        ret['result'] = True
+    else:
+        ret['comment'] = 'Failed to import archive'
+
+    return ret
```
charlesprost: Same for `return ret`?

charlesprost (dismissed):
minor issues to be fixed

make/lint.mk
```diff
@@ -5,5 +5,7 @@
 
 lint:
 	tox -e lint-salt
+	tox -e lint-yaml
+	tox -e lint-shell
 
 .PHONY: lint-local
```
NicolasT: There's no `lint-local` target?

salt/_modules/containerd.py
```diff
@@ -7,7 +7,7 @@
 
 import salt.utils.path
 
-log = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
```
NicolasT: Most other (upstream) Salt modules use `log` i.s.o. `LOG`. Maybe this should be an exception in the linter.

salt/_states/containerd.py
```diff
@@ -114,5 +90,45 @@ def image_managed(name, archive_path=None):
             ret['result'] = True
         else:
             ret['comment'] = 'Failed to pull image'
+        return ret
+
+    return ret
+
 
+def manage_archive_path(name, archive_path, ret):
```
NicolasT: This naming seems not in line with other states. `manage` is active (i.e. a module).

If the intent is for this to be not 'public API of the state module' make it `_manage_archive_path` or something, though even then, even utility procedures in a state should be... a state-like thing.

NicolasT: What's this `ret` thing?

tox.ini
```diff
@@ -7,14 +7,30 @@ skip_install = true
 setenv =
     VIRTUALENV_NO_DOWNLOAD=1
 
+[testenv:lint-shell]
+whitelist_externals = bash
+basepython = python3.6
+skip_install = true
+deps =
+    -r{toxinidir}/lint-requirements.txt
```
NicolasT: Now you're installing all lint deps in 3 different venvs. Either

- List linter-specific deps here (I think that'd be my preference)
- Set up Tox to re-use the same env between the 3 jobs

salt/_states/containerd.py
jbertran: I really dislike the control flow of these dict-filling steps. Yes, you're absolutely right.

salt/_modules/containerd.py
slaperche-scality: On the other hand, it's pretty common to have UPPERCASE for global variable.

So depends if we want clean code or Salt-like code here.

But if we add an exception to the linter, it should be only for Salt code, as we probably don't want these kind of weird decision to be applied to Python script in general

salt/_states/containerd.py
jbertran: Done

salt/_states/containerd.py
jbertran: Done

salt/_modules/containerd.py
jbertran: Leaving it as `log` for now since I'm dropping Python linting

make/lint.mk
jbertran: Indeed :o

salt/_states/containerd.py
jbertran: Reverting this since I'm dropping Python linting.

jbertran: Reverted. Getting lost in salt return dict side effects.

tox.ini
jbertran: lint deps are minimal, so linter-specific it is.

tox.ini
jbertran: Fixed

salt/_modules/containerd.py
slaperche-scality: Why reverting that change? You'll have to reintroduce it when we re-enable linting anyway, so you could keep it.

jbertran: It's trivial enough to reproduce, and if I want to see if I can work in the Salt-specific exception.

slaperche-scality (dismissed):

tox.ini
```diff
@@ -2,21 +2,20 @@
...
 
-[testenv:lint-salt]
-whitelist_externals = bash
-basepython = python3.6
-skip_install = true
+[testenv:lint-shell]
+deps =
+    bashate==0.6.0
+commands =
+    bash -c "bashate **/*.sh"
```
gdemonet, line 15: Would be nice to add `**/*.sh.in` as well, so we can lint e.g. `scripts/bootstrap.sh.in`

gdemonet (dismissed):

tox.ini
jbertran: Done. We ignore long line warnings for `.sh.in`, since they're needed by some of the heredocs included.

jbertran:
@bert-e approve

gdemonet (approved):

PR #652, opened 2019-02-26
GH-556: Bootstrap deployment in Eve

gdemonet:
Using results from the build step, we mount the ISO and execute the bootstrap script on a fresh OpenStack VM worker (using CentOS 7).

Fixes GH-556.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -106,6 +106,10 @@ stages:
           command: sha256sum -c SHA256SUM
           haltOnFailure: true
       - ShellCommand:
+          name: Create mountpoint
+          command: >
+            sudo mkdir -p /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+      - ShellCommand:
           name: Mount ISO image
           command: >
             sudo mount -o loop metalk8s.iso
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -11,15 +11,15 @@ stages:
     worker:
       type: local
     steps:
-    - TriggerStages:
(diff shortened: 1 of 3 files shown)


eve/main.yml
```diff
@@ -10,18 +10,28 @@ stages:
     worker:
       type: local
     steps:
-      - TriggerStages:
-          name: Trigger build and lint stages simultaneously
-          stage_names:
-            - build
-            - lint
+    - TriggerStages:
```
charlesprost: Is this new indentation yaml-valid?

charlesprost (dismissed):

eve/main.yml
jbertran: It's actually more valid than before - I didn't include Eve yaml in the yaml linting step since it uses some form of templating that's not directly YAML-valid, but the yaml linter _does_ yell on indentation quite a lot on this content taken from other `main.yml`s. Ditto with pod manifests.

gdemonet: I ran `yamllint` to fix the indentation of the whole file (see 749652c)

gdemonet:
@bert-e approve

gdemonet:
@bert-e wait

gdemonet:
> @bert-e wait

This was set so we can warn other PR authors that deployment is now part of the CI and needs to pass (bootstrap script return code 0) to get a green build.

jbertran (approved):
LGTM. Let's add 20 minutes to everyone's build :upside_down_face:

Issue #678, opened 2019-02-28
OCI registry readiness and liveness checks incorrectly configured


NicolasT:
The HTTP checks will query the `podIP`, which is the `hostIP` because the `Pod`s run in the host network. However, we bind the registry to `localhost` only, so the checks fail.

NicolasT:
Fixed by #679

PR #679, opened 2019-02-28
registry: fix readiness and liveness checks

NicolasT:
The HTTP checks will query the `podIP`, which is the `hostIP` because
the `Pods` run in the host network. However, we bind the registry to
`localhost` only, so the checks fail.

Fixes: #678


gdemonet (commented):
LGTM, did not test

NicolasT:
@bert-e approve

NicolasT:
@gdemonet Not a real approve? :wink:

sayf-eddine-scality (approved):

Issue #584, opened 2019-02-13
kubeadm: Salt'ify `addon` phase


TeddyAndrieux:
**Goal**
Implement a Salt _state_ which mimicks `kubeadm`s `addon` phase.

```
addon                      Installs required addons for passing Conformance tests
  /coredns                   Installs the CoreDNS addon to a Kubernetes cluster
  /kube-proxy                Installs the kube-proxy addon to a Kubernetes cluster
```

**Delivery**
The Salt _state_.

NicolasT:
I'll work on deploying `kube-proxy`.

NicolasT:
Done.

PR #684, opened 2019-02-28
Create a state to install coredns addon

ramanduh:
The goal of this PR is to install the coredns addon.

**UPDATE**: all components added, coredns is now part of the addons states
to run it apply the following state
`salt-call state.apply metalk8s.kubeadm.init.addons pillarenv=metalk8s-2.0 saltenv=metalk8s-2.0`
Fixes: #584 
**Note**: will squash once all approve


salt/metalk8s/kubeadm/init/addon/installed.sls
```diff
@@ -0,0 +1,11 @@
+Install pip package:
+  pkg.installed:
+    - name: python2-pip
+
+Install kubernetes python client:
+  pip.installed:
```
NicolasT: No, don't install things using `pip`.

NicolasT (changes requested):

salt/_modules/kubeconfig.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+
+import yaml
+
+__virtualname__ = 'metalk8s_kubeconfigs'
```
NicolasT: Looks like this is also done in the related state, but: let's try to keep Python module names and `__virtualname__` the same.

salt/metalk8s/kubeadm/init/addon/installed.sls
NicolasT: `python2-kubernetes` exists in `extras`. If the Salt state can't work with it, then that needs to be fixed, not using some ancient version of upstream.

Makefile
```diff
@@ -111,7 +113,6 @@ ALL = \
 	$(ISO_ROOT)/salt/_modules/kubernetes.py \
 	\
 	$(ISO_ROOT)/salt/_states/containerd.py \
-	$(ISO_ROOT)/salt/_states/kubeconfig.py \
```
NicolasT: Huh? Isn't this used by other states?

sayf-eddine-scality: Arf, will check it

salt/metalk8s/kubeadm/init/addons/coredns.sls
```diff
@@ -0,0 +1,123 @@
+{%- from "metalk8s/map.jinja" import coredns with context %}
+
+{% set kubeconfig = "/etc/kubernetes/admin.conf" %}
+{% set context = "kubernetes-admin@kubernetes" %}
+
+include:
+  - metalk8s.salt.minion.running
```
NicolasT: We don't do this in the `kube-proxy` state. Not sure it's required: in the end, if the minion is not running, this state likely won't run either...

salt/metalk8s/kubeadm/init/addons/coredns.sls
```diff
@@ -0,0 +1,123 @@
...
+
+Create coredns deployment:
+  kubernetes.deployment_present:
+    - name: coredns
+    - namespace: kube-system
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+    - source: salt://metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
+  require:
+    - kubernetes: Create coredns ConfigMap
```
NicolasT, line 40: If the step before depends on `Install Python Kubernetes client`, then this one (and others?) as well, no?

salt/metalk8s/kubeadm/init/addons/coredns.sls
sayf-eddine-scality: this state will make sure salt-minion restarts, because there was a state that adds some config file to the minion, so true, no need for this anymore

salt/metalk8s/kubeadm/init/addons/coredns.sls
sayf-eddine-scality: hmm yes make send for all of them to depends

salt/metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+    spec:
+      priorityClassName: system-cluster-critical
+      tolerations:
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      containers:
+      - name: coredns
+        image: coredns/coredns:1.3.1
```
TeddyAndrieux: This image should be part of our local registry

TeddyAndrieux (changes requested):

salt/metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
sayf-eddine-scality: yes I noticed that we have one but not the right version

NicolasT: It doesn't seem to be the 'right' version either? In my `kubeadm` `dry-run` output, for K8s 1.11, it uses `coredns:1.2.2`.

sayf-eddine-scality: hmmm, https://github.com/kubernetes/kubernetes/issues/69983
I guess 1.2.2 is too old, found the latest tested version to be 1.3.1 on https://console.cloud.google.com/gcr/images/google_containers/GLOBAL/coredns?gcrImageListsize=50&pli=1
Tell me what do you think about the version

salt/metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: coredns
-        image: coredns/coredns:1.3.1
+        image: k8s.gcr.io/coredns:1.3.1
```
NicolasT: Should be possible to fetch this from `localhost:5000`?

salt/metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+    spec:
+      priorityClassName: system-cluster-critical
+      tolerations:
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      containers:
+      - name: coredns
+        image: localhost:5000/metalk8s-2.0/coredns:1.3.1
```
TeddyAndrieux, line 29: Not really like hardcorded think like that in the yaml file but btw we can't do better using source with `kubernetes` states

salt/metalk8s/kubeadm/init/addons/coredns.sls
```diff
@@ -0,0 +1,126 @@
...
+    - role_ref:
+        apiGroup: rbac.authorization.k8s.io
+        kind: ClusterRole
+        name: system:coredns
+    - subjects:
+      - kind: ServiceAccount
+        name: coredns
+        namespace: kube-system
+  require:
+    - pkg: Install Python Kubernetes client
```
TeddyAndrieux, line 123: If you require it please include the state because here that mean we **can't** call directly the state `coredns`

TeddyAndrieux (changes requested):

salt/metalk8s/kubeadm/init/addons/coredns.sls
```diff
@@ -0,0 +1,126 @@
...
+
+Create coredns service:
+  kubernetes.service_present:
+    - name: coredns
+    - namespace: kube-system
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+    - metadata:
+        name: kube-dns
+        namespace: kube-system
```
TeddyAndrieux: You should remove those 2 metadata taken from the upper section and seems like you use `kube-dns` as name

salt/metalk8s/kubeadm/init/addons/files/coredns_deployment.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: coredns
+  namespace: kube-system
```
TeddyAndrieux: Same as above those should be removed from here

TeddyAndrieux (changes requested):

TeddyAndrieux (approved):

ramanduh:
@bert-e approve

NicolasT (approved):

Issue #663, opened 2019-02-27
Make openstack workers debuggable


jbertran:
We have everyone's SSH keys as secrets in Eve, so we can now set up the spawned openstack VMs to:

* report their own IPs as a build property
* add all of us to authorized_keys

Then, if debug is needed, all that is necessary is to add a sleep step to the Eve `main.yaml` file.

jbertran:
Blocked by GH-652 - need the OpenStack deployment before we can debug it.

jbertran:
Closed by #687

PR #687, opened 2019-03-01
GH-663- add conditional VM debug step

jbertran:
For ease of debug, the single-node deployment exposes a
post-bootstrap step that conditions IP report, addition
of platform team public keys and a 1-hour debugging time
window on the spawned VM behind a `doStepIf` statement.

Issue: GH-663

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -121,3 +121,12 @@ stages:
             sudo bash
             /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
           haltOnFailure: true
+      - ShellCommand:
+          name: Debug step - report IPs, add SSH keys, wait 1 hour
+          timeout: 3600
+          command: >
+            ip a &&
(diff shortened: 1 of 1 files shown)


gdemonet (approved):

gdemonet:
If one would like to test this step in a build, one would need to edit the `eve/main.yaml` line as follows:

```
       - ShellCommand:
           name: Debug step - report IPs, add SSH keys, wait 1 hour
           timeout: 3600
           command: >
             ip a &&
             mkdir -p ~/.ssh &&
             echo "%(secret:ssh_pub_keys)s" >> ~/.ssh/authorized_keys &&
             sleep 3600
-          doStepIf: false
+          doStepIf: true
```

gdemonet:
Until Eve supports casting property values into booleans, editing `eve/main.yaml` is the only way to have this step run in a build. In the future, we expect to link this to a build property, that one could set by hand when forcing a build.

jbertran:
@bert-e approve

Issue #582, opened 2019-02-13
kubeadm: Salt'ify `mark-control-plane` phase


TeddyAndrieux:
**Goal**
Implement a Salt _state_ which mimicks `kubeadm`s `mark-control-planet` phase.

**Delivery**
The Salt _state_.

gdemonet:
Fixed by #692

PR #689, opened 2019-03-01
Deploy kube-proxy

NicolasT:
This set of patches implements deployment of `kube-proxy` as a `DaemonSet`. It adds the `kubernetes` state and module from SaltStack v2019.2.0, and extends these with some code to manage `ClusterRoleBinding`s, `ServiceAccount`s, `Role`s and `RoleBinding`s, as needed to deploy `kube-proxy`.

Fixes: #685

Issue #685, opened 2019-02-28
kubeadm: deploy kube-proxy


NicolasT:
Cfr. #584

NicolasT:
#689

PR #692, opened 2019-03-01
Salt'ify mark-control-plane phase

gdemonet:
Note that we also added an annotation for specifying the CRI socket used, as can be seen in the [kubeadm dry-run output](https://github.com/NicolasT/kubeadm-dry-run/blob/master/kubeadm-dry-run.txt#L518-L524).

Fixes GH-582.

### DIFF
--- a/salt/_modules/kubernetes.py
+++ b/salt/_modules/kubernetes.py
@@ -2093,6 +2093,55 @@ def show_customresourcedefinition(name, **kwargs):
         _cleanup(**cfg)
 
 
+def node_annotations(name, **kwargs):
+    '''
+    Return the annotations of the node identified by the specified name
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_annotations name="minikube"
+    '''
+    match = node(name, **kwargs)
+
+    if match is not None:
+        return match['metadata']['annotations']
+
+    return {}
+
+
+def node_add_annotation(node_name, annotation_name, annotation_value, **kwargs):
+    '''
+    Set the value of the annotation identified by `annotation_name` to
+    `annotation_value` on the node identified by the name `node_name`.
+    Creates the annotation if not present.
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_add_annotation node_name="minikube" \
+            annotation_name="foo" annotation_value="bar"
+    '''
+    cfg = _setup_conn(**kwargs)
+    try:
+        api_instance = kubernetes.client.CoreV1Api()
+        body = {
+            'metadata': {
+                'annotations': {
+                    annotation_name: annotation_value
+                }
+            }
+        }
+        api_response = api_instance.patch_node(node_name, body)
+        return api_response
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception('Exception when calling CoreV1Api->patch_node')
+            raise CommandExecutionError(exc)
+    finally:
+        _cleanup(**cfg)
+
+
 def create_customresourcedefinition(
         name,
         spec,
@@ -2148,6 +2197,38 @@ def replace_customresourcedefinition(
                 'Exception when calling '
                 'ExtensionsV1beta1Api->replace_custom_resource_definition'
             )
+            raise CommandExecutionError(exc)
+    finally:
+        _cleanup(**cfg)
+
+
+def node_remove_annotation(node_name, annotation_name, **kwargs):
+    '''
+    Removes the annotation identified by `annotation_name` from
+    the node identified by the name `node_name`.
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_remove_annotation node_name="minikube" \
+            annotation_name="foo"
+    '''
+    cfg = _setup_conn(**kwargs)
+    try:
+        api_instance = kubernetes.client.CoreV1Api()
+        body = {
+            'metadata': {
+                'annotations': {
+                    annotation_name: None}
+                }
+        }
+        api_response = api_instance.patch_node(node_name, body)
+
+        return api_response.to_dict()
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception('Exception when calling CoreV1Api->patch_node')
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
--- a/salt/_modules/kubernetes.py
+++ b/salt/_modules/kubernetes.py
@@ -2232,3 +2232,46 @@ def node_remove_annotation(node_name, annotation_name, **kwargs):
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
+
+
+def node_taints(node_name, **kwargs):
+    match = node(node_name, **kwargs)
+
+    if match is not None:
+        return match['spec']['taints']
+
+    return []
+
+
+def node_set_taints(node_name, taints, **kwargs):
+    '''
+    Update the array of taints to the provided `taints` value for the node
+    identified by the name `node_name`.
+    Will overwrite any existing taints for this node.
+
+    CLI Examples::
+
+        # To remove all taints
+        salt '*' kubernetes.node_set_taints node_name="minikube" taints=[]
+
+        # To add some taints
+        salt '*' kubernetes.node_set_taints node_name="master1" \
+            taints='[ \
+                {"key": "node-role.kubernetes.io/master", \
+                 "effect" "NoSchedule" } \
+            ]'
+    '''
+    cfg = _setup_conn(**kwargs)
+    try:
+        api_instance = kubernetes.client.CoreV1Api()
+        body = {'spec': {'taints': taints}}
+        api_response = api_instance.patch_node(node_name, body)
+        return api_response
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception('Exception when calling CoreV1Api->patch_node')
+            raise CommandExecutionError(exc)
+    finally:
+        _cleanup(**cfg)
--- a/salt/_states/kubernetes.py
+++ b/salt/_states/kubernetes.py
@@ -1124,6 +1124,64 @@ def node_annotation_present(
     return ret
 
 
+def node_taints_present(name, taints, **kwargs):
+    '''
+    # TODO: describe and provide expls
+    '''
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    existing_taints = {
+        t['key']: t
+        for t in __salt__['kubernetes.node_taints'](name, **kwargs)
+    }
+
+    added = set()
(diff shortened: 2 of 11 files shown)


NicolasT:
@bert-e wait

NicolasT:
Let's wait for CI to merge #689, then rebase this one on `development/2.0` and retarget this PR to `development/2.0` (no need to close this PR, can edit it in-place).

salt/_modules/kubernetes.py
```diff
@@ -1797,3 +1797,132 @@ def replace_role(
...
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception('Exception when calling CoreV1Api->patch_node')
+            raise CommandExecutionError(exc)
+    finally:
+        _cleanup(**cfg)
+
+    return None
```
NicolasT: Is this done in the other procedures? There should be no need, either `return api_response` is reached, or the `return None` or `raise` in the `except` clause, or some other exception bubbles up. This is unreachable.

Also applies to other occurrences.

salt/_modules/kubernetes.py
```diff
@@ -1797,3 +1797,132 @@ def replace_role(
...
+
+
+def node_taints(node_name, **kwargs):
+    match = node(node_name, **kwargs)
+
+    if match is not None:
+        return match['spec']['taints']
+
+    return None
```
NicolasT: This means the node doesn't exist, rather than 'no taints'. If this is how other procedures in the module work, then fine. We should rework it anyway :stuck_out_tongue:

salt/_states/kubernetes.py
```diff
@@ -1009,6 +1009,179 @@ def node_label_present(
...
+            ret['result'] = None
+            ret['comment'] = 'The annotation is going to be set'
+            return ret
+
+        __salt__['kubernetes.node_add_annotation'](
+            annotation_name=name,
+            annotation_value=value,
+            node_name=node,
+            **kwargs
+        )
```
NicolasT, line 1092: In this code-path, `ret['comment']` is not set.

salt/_states/kubernetes.py
```diff
@@ -1009,6 +1009,179 @@ def node_label_present(
...
+        'old': old_annotations,
+        'new': annotations}
+    ret['result'] = True
+
+    return ret
+
+
+def node_taints_present(name, taints, **kwargs):
+    '''
+    # TODO: describe and provide expls
```
NicolasT: I'd prefer no docstring (which can be caught by static analysis) than a docstring which is de-facto empty, no?

salt/_states/kubernetes.py
```diff
@@ -1009,6 +1009,179 @@ def node_label_present(
...
+    '''
+    ret = {'name': name,
+           'changes': {},
+           'result': False,
+           'comment': ''}
+
+    existing_taints = {
+        t['key']: t
+        for t in __salt__['kubernetes.node_taints'](name, **kwargs)
```
NicolasT: Prefer not to call things with side-effects in a comprehension.

salt/_states/kubernetes.py
```diff
@@ -1009,6 +1009,179 @@ def node_label_present(
...
+        )
+        return ret
+
+    __salt__['kubernetes.node_set_taints'](
+        node_name=name,
+        taints=new_taints,
+        **kwargs
+    )
+
+    ret['result'] = True
```
NicolasT, line 1173: `ret['comment']` not set

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
```diff
@@ -0,0 +1,31 @@
+{% set hostname = salt['network.get_hostname']() %}
+
+{% set kubeconfig = "/etc/kubernetes/admin.conf" %}
+{% set context = "kubernetes-admin@kubernetes" %}
+
+Apply control-plane role label:
+  kubernetes.node_label_present:
+    - node: {{ hostname }}
+    - name: "node-role.kubernetes.io/master"
```
NicolasT, line 8: @TeddyAndrieux prefers `name` to go first, then `kubeconfig` and `context`, then `node` and `value`. I kinda agree (had to be convinced though ;))

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+
+Apply control-plane taints:
+  kubernetes.node_taints_present:
+    - name: {{ hostname }}
+    - taints:
+        - key: "node-role.kubernetes.io/master"
+          effect: "NoSchedule"
+    - kubeconfig: {{ kubeconfig }}
```
NicolasT: Same

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+    - context: {{ context }}
+
+Annotate with CRI socket information:
+  kubernetes.node_annotation_present:
+    - node: {{ hostname }}
+    - name: "kubeadm.alpha.kubernetes.io/cri-socket"
+    # NOTE: the socket location may become configurable, we should make sure
```
NicolasT: `TODO` i.s.o. `NOTE`.

NicolasT (changes requested):
Definitely requires rebase after #685 lands.

salt/_modules/kubernetes.py
gdemonet: Indeed, not sure exactly why I added them, will remove when rebasing.

salt/_modules/kubernetes.py
gdemonet: Actually I should have returned an empty list instead (copy-pasted the None as for the other methods I guess), but yes this module needs a rework :)

salt/_states/kubernetes.py
gdemonet: Agreed, I was hoping to find the courage to fill it before pushing...

salt/_states/kubernetes.py
gdemonet: Done.

salt/_states/kubernetes.py
gdemonet: Split the comprehension in two.

salt/_states/kubernetes.py
gdemonet: Done.

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
gdemonet: Done.

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
gdemonet: Done.

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
gdemonet: Done.

gdemonet:
Rebased and integrated review comments from @NicolasT

gdemonet:
Build failed since `mark-control-plane` runs before `addons`, and as such `python2-kubernetes` is not yet installed when we execute this step.

A quick fix could be to include `..addons.dependencies` in `metalk8s.kubeadm.init.mark-control-plane.init`, but this doesn't look "clean".
I'm wondering if doing something like a `metalk8s.python-kubernetes` formula could be useful / acceptable.

salt/_modules/kubernetes.py
```diff
@@ -1975,3 +1975,126 @@ def replace_daemonset(
...
+        salt '*' kubernetes.node_remove_annotation node_name="minikube" \
+            annotation_name="foo"
+    '''
+    cfg = _setup_conn(**kwargs)
+    try:
+        api_instance = kubernetes.client.CoreV1Api()
+        body = {
+            'metadata': {
+                'annotations': {
+                    annotation_name: None}
```
cmonfort, line 2221: Is the annotation removed when its value is set to `None`?
If not, it's not really a removal, and if I get it afterwards, I will not have a failure because it does not exist, but retrieve a `None` value...)

salt/_states/kubernetes.py
```diff
@@ -1009,6 +1009,171 @@ def node_label_present(
...
+        ret['comment'] = 'The annotation is already set, changing the value'
+
+        __salt__['kubernetes.node_add_annotation'](
+            node_name=node,
+            annotation_name=name,
+            annotation_value=value,
+            **kwargs
+        )
+
+    old_annotations = copy.copy(annotations)
```
cmonfort, line 1116: Get the full annotations block? not only the value of one we are changing?

salt/_modules/kubernetes.py
```diff
@@ -1975,3 +1975,126 @@ def replace_daemonset(
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
+
+
+def node_annotations(name, **kwargs):
```
cmonfort, line 2096: For consistency with other annotations commands, `name` parameter should be `node_name`.

gdemonet: I copy-pasted the signature of [`node_labels`](https://github.com/scality/metalk8s/blob/020703de54117108286831e813d3076ecdf39d3a/salt/_modules/kubernetes.py#L321), I'll keep it as is until we refactor this custom module (which is essentially a long list of copy-pasted functions)

salt/_modules/kubernetes.py
gdemonet: Yup, setting an annotation or label to `None` in a PATCH call will effectively remove it - setting a label/annotation with _no value_ would be done by using an empty string instead.

salt/_states/kubernetes.py
gdemonet: Same comment as for the signature of `node_annotations`, I copy-pasted what was done for labels, and they indeed show the whole labels block instead of simply the one changed... To discuss in future improvements.

cmonfort (dismissed):

NicolasT (dismissed):
LGTM (didn't run)

gdemonet:
@bert-e approve

charlesprost (approved):

salt/metalk8s/kubeadm/init/mark-control-plane/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+        - key: "node-role.kubernetes.io/master"
+          effect: "NoSchedule"
+
+Annotate with CRI socket information:
+  kubernetes.node_annotation_present:
+    - name: "kubeadm.alpha.kubernetes.io/cri-socket"
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+    - node: {{ hostname }}
+    # TODO: the socket location may become configurable, we should make sure
```
ramanduh, line 29: I think, it's already configurable in default.yaml: `kubelet.service.options.container-runtime-endpoint`

Issue #656, opened 2019-02-26
Make sure container is running instead of waiting


sayf-eddine-scality:
in the bootstrap we added a sleep command waiting for the container image to be pulled and then executed, if we will keep using salt commands we may need to do this properly

PR #710, opened 2019-03-07
Make sure salt master container is running before declaring success

sayf-eddine-scality:
Fixes: #656
Salt master pod may take some time to start due to image pulling etc, the state deploying it should wait until it is running before declaring success

Fixes: #656

### DIFF
--- a/salt/metalk8s/salt/master/deployed.sls
+++ b/salt/metalk8s/salt/master/deployed.sls
@@ -31,3 +31,14 @@ Install and start salt master manifest:
       salt_master_version: {{ salt_master_version }}
     - require:
       - file: Create salt master directories
+
+Make sure salt master container is up:
+  cmd.run:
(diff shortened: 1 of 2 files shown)


NicolasT:
Doesn't this mean the `sleep` can be removed as well?

sayf-eddine-scality:
> Doesn't this mean the `sleep` can be removed as well?

Done

salt/metalk8s/salt/master/deployed.sls
```diff
@@ -31,3 +31,12 @@ Install and start salt master manifest:
       salt_master_version: {{ salt_master_version }}
     - require:
       - file: Create salt master directories
+
+Make sure salt master container is up:
```
alexandre-allard, line 35: Shouldn't this command require "Install and start salt master manifest" ?

sayf-eddine-scality: right

sayf-eddine-scality:
@bert-e approve

TeddyAndrieux (approved):

Issue #708, opened 2019-03-06
Replacing GNU Make


slaperche-scality:
## Current state

We are using GNU Make to build the ISO and to automate few menial tasks (running the lint tools, spawning a VM with Vagrant, …)

The problems with Make, given that nobody is a Make guru in our team, are:
- it's hard to maintain
- it's hard to evolve
- it's hard to debug

Not convinced?

Let's look at some examples (no finger pointing intended here, everyone reviewed and approved the code):

### Code duplication

- image download: [here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L438-L441) vs [there](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L452-L455) => minor different (gzip vs no gzip).
- templating: [here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L201-L205) and [there](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L212-L215).
- repository building: [here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L345-L357) vs [there](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L382-L393)

### Inconsistency

- `TAG` vs `VERSION`, name with digest included vs dedicated `DIGEST` variable: [here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/container_images.mk#L25-L27) vs [there](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/container_images.mk#L33-L35)
- Use of `DOCKER` [here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L439-L441) vs `docker` command used directly elsewhere ([here](https://github.com/scality/metalk8s/blob/090c5f9d1fb98d37bafa44fd91d55bb6b6b205f7/Makefile#L382) for instance)

### Legibility

[...]

Issue #618, opened 2019-02-20
build: split RPM package creation from main Makefile


jbertran:
* Using an `include` in the main Makefile, split the RPM building targets from the main Makefile.
* Document the different steps as much as possible, so that a newcomer may easily understand the workflow of the Makefile.

Part of #603

gdemonet:
Since RPM packages and repositories are quite independent from the other build targets, one can define an independent `Makefile` and run it as a "sub-`Make`" from the main one.

Here's a draft PR (#748) to show how it might look, so that we have grounds for comparing with `doit` (#708).

slaperche-scality:
Done/made obsolete as a side-effect of https://github.com/scality/metalk8s/issues/708.

PR #748, opened 2019-03-12
Improvement: Split packages Makefile

gdemonet:
Quick draft of how one can split and factorize a Makefile (this works well for generating a slew of packages, even though it's not demonstrated in this PR) - issue #618.

This PR might not necessarily land at any given time, however it should be useful for comparing GNU Make with alternative solutions in the context of this project (see #708).


packages/Makefile
```diff
@@ -0,0 +1,233 @@
...
+PACKAGES_ROOT := $(BUILD_ROOT)/packages
+$(PACKAGES_ROOT):
+	mkdir -p $@
+
+# Sub-directories to create for each package
+PACKAGE_SUBDIRS := SOURCES SRPMS RPMS SPECS
+
+# Templated rules for building a RPM from sources (see packages/packages.mk for
+# an example usage)
+define create-pkg-from-source =
```
NicolasT, line 45: My .02: once starting to use template rules and functions and macros and whatnot in `Makefile`s, we should stop using `make`.

gdemonet: Agreed, which is why I left the PR in draft state. The main goal is to show how bad it can become, in comparison to more powerful solutions (e.g. `doit`, incoming PR from @slaperche-scality)

NicolasT:
Guess this scan be closed given #795.

Issue #724, opened 2019-03-07
Provide a framework to launch additional VMs


charlesprost:
- [x] Use OpenStack credentials + Terraform to instantiate new VMs on top of the one running the tests in the CI
- [x] Use Vagrant multi-machine environement to instantiate new VMs in local space

Ask @tcarmet for a way to access the tenant for metalk8s.

Success is defined by being able to trigger new machine creation from command-line.

PR #759, opened 2019-03-12
Improvement/724 add multiple vms local env vagrant

charlesprost:
Hint: you may want to read this PR commit by commit to understand better the changes.

**Warning**: I didn't test yet this PR to see if the bootstrap node is still working as expected. Changes can happen since there is a new network in place.

Be careful since any new node will take 2GB in your laptop RAM.

Fixes: #724 for the local environment (not the CI one).

On this branch, just perform a `vagrant status`, you should see 9 new VMs available:

```bash

$ vagrant status
Current machine states:

bootstrap                 running (virtualbox)
node1                     not created (virtualbox)
node2                     not created (virtualbox)
node3                     not created (virtualbox)
node4                     not created (virtualbox)
node5                     not created (virtualbox)
node6                     not created (virtualbox)
node7                     not created (virtualbox)
node8                     not created (virtualbox)
node9                     not created (virtualbox)
```

By default they are not started when performing a `vagrant up`, neither there is a conflict when performing a `vagrant ssh` since the bootstrap node stays the main one.

Any action related to these extra nodes needs to be explicit:

- `vagrant up node1`
- `vagrant ssh node1`
- `vagrant destroy -f node1`

All nodes (including the bootstrap one) have now a new interface to communicate on the same network `192.168.42.0/24`.
The bootstrap node is `192.168.42.10`.
Others are incremental: `192.168.42.11` for `node1`, `.12` for `node2`, etc.

With this PR, it will be easy for an E2E test to call for a new VM instantiation.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,18 +52,19 @@ SCRIPT
 Vagrant.require_version(">= 1.8")
 
 Vagrant.configure("2") do |config|
+  config.vm.box = "centos/7"
+  config.vm.box_version = "1811.02"
+
+  config.vm.provider "virtualbox" do |v|
+    v.linked_clone = true
+    v.memory = 2048
+    v.cpus = 2
+  end
+
   config.vm.define :bootstrap do |bootstrap|
-    bootstrap.vm.box = "centos/7"
-    bootstrap.vm.box_version = "1811.02"
     bootstrap.vm.hostname = "bootstrap"
 
-    bootstrap.vm.provider "virtualbox" do |v|
-      v.linked_clone = true
-      v.memory = 2048
-      v.cpus = 2
-
-      bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
-    end
+    bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
 
     bootstrap.vm.provision "import-release",
       type: "shell",
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -48,6 +48,9 @@ echo "Launching bootstrap"
 exec "/srv/scality/metalk8s-$SHORT_VERSION/bootstrap.sh"
 SCRIPT
 
+NETWORK_RANGE = "192.168.42."
+BOOTSTRAP_IP_OFFSET = 10
+
 # To support VirtualBox linked clones
 Vagrant.require_version(">= 1.8")
 
@@ -64,7 +67,13 @@ Vagrant.configure("2") do |config|
   config.vm.define :bootstrap, primary: true do |bootstrap|
     bootstrap.vm.hostname = "bootstrap"
 
-    bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+    bootstrap_ip = "#{NETWORK_RANGE}#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_ip
+
+    bootstrap.vm.provider "virtualbox" do |v|
+      v.memory = 4096
+      bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+    end
 
     bootstrap.vm.provision "import-release",
       type: "shell",
@@ -77,9 +86,12 @@ Vagrant.configure("2") do |config|
 
   (1..9).each do |i|
     node_name = "node#{i}"
+    node_ip = "#{NETWORK_RANGE}#{BOOTSTRAP_IP_OFFSET+i}"
 
     config.vm.define node_name, autostart: false do |node|
       node.vm.hostname = node_name
+
+      node.vm.network "private_network", ip: node_ip
 
       node.vm.synced_folder ".", "/vagrant", disabled: true
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -61,7 +61,7 @@ Vagrant.configure("2") do |config|
     v.cpus = 2
   end
 
-  config.vm.define :bootstrap do |bootstrap|
+  config.vm.define :bootstrap, primary: true do |bootstrap|
     bootstrap.vm.hostname = "bootstrap"
 
     bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
@@ -73,5 +73,18 @@ Vagrant.configure("2") do |config|
     bootstrap.vm.provision "bootstrap",
       type: "shell",
       inline: BOOTSTRAP
+  end
+
+  (1..9).each do |i|
+    node_name = "node#{i}"
+
+    config.vm.define node_name, autostart: false do |node|
+      node.vm.hostname = node_name
+
+      node.vm.synced_folder ".", "/vagrant", disabled: true
+
+      # No need for Guest Additions since there is no synced folder
+      node.vbguest.auto_update = false
+    end
   end
 end


Vagrantfile
```diff
@@ -48,22 +48,29 @@ echo "Launching bootstrap"
...
 Vagrant.configure("2") do |config|
-  config.vm.define :bootstrap do |bootstrap|
-    bootstrap.vm.box = "centos/7"
-    bootstrap.vm.box_version = "1811.02"
+  config.vm.box = "centos/7"
+  config.vm.box_version = "1811.02"
+
+  config.vm.provider "virtualbox" do |v|
+    v.linked_clone = true
+    v.memory = 2048
```
NicolasT, line 63: This (and CPU) is not necessarily the same across all nodes. As an example, I bump the `bootstrap` node memory to 4GB. However, other control-plane nodes could stay at 2GB (for now, may need some more memory once/if we deploy ES), and workload-plane nodes at 1GB, or even lower.

Vagrantfile
```diff
@@ -48,22 +48,29 @@ echo "Launching bootstrap"
...
-    bootstrap.vm.provider "virtualbox" do |v|
-      v.linked_clone = true
-      v.memory = 2048
-      v.cpus = 2
+    bootstrap_ip = NETWORK_RANGE + "#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_ip
 
-      bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
-    end
+    bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
```
NicolasT: This is VirtualBox-specific, hence needs to be in a `provider` block.

Vagrantfile
```diff
@@ -73,4 +80,20 @@ Vagrant.configure("2") do |config|
       type: "shell",
       inline: BOOTSTRAP
   end
+
+  (1..9).each do |i|
```
NicolasT, line 87: Not sure this is the right approach. Once we support multi-node clusters, I'd consider to let Vagrant set up some standard infra, and indeed maybe allow to add some more, though have multi-node by default, with node-names that reference their role.

Also, the Vagrant provisioning should then set up the whole cluster.

Vagrantfile
```diff
@@ -64,6 +67,9 @@ Vagrant.configure("2") do |config|
   config.vm.define :bootstrap, primary: true do |bootstrap|
     bootstrap.vm.hostname = "bootstrap"
 
+    bootstrap_ip = NETWORK_RANGE + "#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_ip
```
NicolasT, line 71: Are we forced to provide an IP address in a private network? Is there no way to let Vagrant allocate IPs, and maybe fill a `hosts` file or similar? It'd be nice not to need to worry about IPAM...

Vagrantfile
```diff
@@ -48,6 +48,9 @@ echo "Launching bootstrap"
 exec "/srv/scality/metalk8s-$SHORT_VERSION/bootstrap.sh"
 SCRIPT
 
+NETWORK_RANGE = "192.168.42."
+BOOTSTRAP_IP_OFFSET = 10
```
NicolasT, line 52: What's the need for this?

Vagrantfile
charlesprost: Yeah, I'll wait for your PR with the 4GB RAM to be merged to integrate this new value.
I'm not sure we want to have all kind of nodes in the Vagrant environment, at least right now, and since it would be difficult to have a lot of those nodes running at the same time, we may not be able to reproduce any kind of environment though.

Vagrantfile
charlesprost: Indeed.

Vagrantfile
charlesprost: About the provisioning, I think we should do everything from the bootstrap node (except starting the new VMs of course). So other nodes won't probably ever have a provisioning step.

Yeah I'm not sure about the approach we want to have. One thing that I think is already settled is that we can not have a too-big cluster otherwise we won't be able to launch it on any laptop. Hence having not autostarted VMs.
But yeah we could have dedicated nodes to some roles. Then we could easily define all this in some yaml file and import it and use it into the Vagrantfile if we need to. We would just have to change the loop.

Since we don't really know what we want to test, I'd say it could be good enough to have access to those extra VMs in the meantime.

Vagrantfile
charlesprost: No we could use a DHCP of course, but then we'll have to SSH into the VMs to know which IP has been assigned, see https://www.vagrantup.com/docs/networking/private_network.html#dhcp.

One thing that we can do, related to my other comment, is having a yaml file looking as:
```yaml
---
- name: bootstrap
  ip: 192.168.42.10
  role: bootstrap
- name: node1
  ip: 192.168.42.11
  role: workload-plane
...
```
and use this architecture file in the Vagrantfile to setup the VMs.

Vagrantfile
charlesprost: Just to define once and for all the IP of the bootstrap node in the network range and use it as an offset for the other IP addresses. We could do without it though.

Vagrantfile
jbertran: From a user perspective:
- having a dedicated bootstrap node running for the E2E bootstrap test is a necessity
- having other nodes started is not - we don't need multiple nodes to confirm high-level bootstrap functions. 

To me the multi-VM thing would only be important for multi-node E2E tests, which we don't necessarily want to launch every time. Maybe we want a dedicated VagrantFile for these?

Although we probably don't need that many nodes to be ready, I'm happy with this. We can revise it once the need for multiple purposes and corresponding VM definitions comes up.

jbertran (dismissed):

charlesprost:
Since the issue is not about having the perfect multi-environments setup that we could dream of, but just having the ability of simple adding more machines for some tests no yet written, I'd say it's good enough for now.
Based on this new Vagrantfile, it's gonna be easy to add new environments in the future.

Vagrantfile
```diff
@@ -74,4 +74,17 @@ Vagrant.configure("2") do |config|
       type: "shell",
       inline: BOOTSTRAP
   end
+
+  (1..9).each do |i|
```
slaperche-scality, line 87: minor: If you don't mind having node [0, 8] instead of [1, 9], you could write `9.times.each` instead of `(1..9).each`.

Vagrantfile
```diff
@@ -48,20 +48,30 @@ echo "Launching bootstrap"
...
+  end
+
+  config.vm.define :bootstrap, primary: true do |bootstrap|
     bootstrap.vm.hostname = "bootstrap"
 
-    bootstrap.vm.provider "virtualbox" do |v|
-      v.linked_clone = true
-      v.memory = 2048
-      v.cpus = 2
+    bootstrap_ip = NETWORK_RANGE + "#{BOOTSTRAP_IP_OFFSET}"
```
slaperche-scality: String interpolation + string concatenation is a bad pattern.

Either use:

`bootstrap_ip = NETWORK_RANGE + BOOTSTRAP_IP_OFFSET.to_s`

or

`bootstrap_ip = "#{NETWORK_RANGE}#{BOOTSTRAP_IP_OFFSET}"`

Vagrantfile
charlesprost: Since bootstrap node may be considered as node0, I prefer it stays this way :)

Vagrantfile
charlesprost: oops, done.

Vagrantfile
sayf-eddine-scality: that information is in the provider not in vagrant, `vboxmanage list runningvms | cut -d" " -f 1 | xargs -L 1 -I {} VBoxManage guestproperty get {} '/VirtualBox/GuestInfo/Net/1/V4/IP'` :stuck_out_tongue:

Vagrantfile
charlesprost: Let's wait to have a clear vision on which kind of environments we want before setting different values for nodes.

Vagrantfile
charlesprost: OK, then let's wait until we have a better understanding of which environments we want to build.

Vagrantfile
charlesprost: Yep we could find it in Virtualbox if need to, but I don't think there is a real difference between finding it out this way and having static IPs. Anyway, that's not for now.

jbertran (approved):
It's as good as we need it for now. Plenty of ideas for additions in the comments though.

charlesprost:
@bert-e approve

Issue #751, opened 2019-03-12
Set up `salt-api` to authenticate against the Kubernetes API


NicolasT:

NicolasT:
#978

Issue #750, opened 2019-03-12
Deploy `salt-api`


NicolasT:
Deploy the `salt-api` service.

NicolasT:
Merged in #760

PR #760, opened 2019-03-12
Integrate SaltAPI

NicolasT:
This PR integrates the Salt HTTP API in the stack.

Fixes: #750 
Fixes: #751

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -244,7 +244,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/_states/containerd.py'),
     Path('salt/_states/kubeconfig.py'),
     Path('salt/_states/docker_registry.py'),
-    Path('salt/_states/kubernetes.py'),
+    Path('salt/_states/kubernetesmod.py'),
 
     targets.RemoteImage(
         registry=constants.GOOGLE_REGISTRY,
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -213,7 +213,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/registry/populated.sls'),
     Path('salt/metalk8s/registry/files/registry-pod.yaml.j2'),
 
-    Path('salt/metalk8s/salt/master/files/master_99-metalk8s.conf'),
+    Path('salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2'),
     Path('salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2'),
     Path('salt/metalk8s/salt/master/configured.sls'),
     Path('salt/metalk8s/salt/master/deployed.sls'),
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -230,6 +230,8 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/ui/deployed.sls'),
     Path('salt/metalk8s/ui/init.sls'),
 
+    Path('salt/_auth/kubernetes_rbac.py'),
+
     Path('salt/_modules/containerd.py'),
     Path('salt/_modules/cri.py'),
     Path('salt/_modules/docker_registry.py'),
@@ -240,6 +242,8 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/_pillar/metalk8s_nodes.py'),
 
     Path('salt/_roster/kubernetes_nodes.py'),
+
+    Path('salt/_runners/metalk8s_saltutil.py'),
 
     Path('salt/_states/containerd.py'),
     Path('salt/_states/kubeconfig.py'),
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -1,0 +1,57 @@
+import base64
+import logging
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'kubernetes_rbac'
+
+def __virtual__():
+    return __virtualname__
+
+
+ADMIN_USERNAME = b'admin'
+ADMIN_PASSWORD = b'admin'
+ADMIN_TOKEN = base64.encodestring(
+    b'{}:{}'.format(ADMIN_USERNAME, ADMIN_PASSWORD)).rstrip()
+AUTH_HANDLERS = {
+    'basic': {
+        'auth': lambda username, token: all([
+            username == ADMIN_USERNAME,
+            token == ADMIN_TOKEN,
+        ]),
+        'groups': lambda _u, _t: [
+            'node_admin',
+        ],
+    },
+}
+
+
+def auth(username, token, token_type, **kwargs):
+    log.info('Authentication request for "%s"', username)
+
+    handler = AUTH_HANDLERS.get(token_type.lower(), None)
+    if not handler:
+        log.warning('Unknown auth token_type: %s', token_type)
+        return False
+
+    result = handler.get('auth', lambda _u, _t: False)(username, token)
+    if result:
+        log.info('Authentication request for "%s" succeeded', username)
+    else:
+        log.warning('Authentication request for "%s" failed', username)
+
+    return result
+
+
+def groups(username, token, token_type, **kwargs):
+    log.info('Groups request for "%s"', username)
+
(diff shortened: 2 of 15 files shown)


charlesprost (commented):
@NicolasT sorry I don't understand how this PR is fixing #751. For #750 it's OK.

NicolasT:
> @NicolasT sorry I don't understand how this PR is fixing #751. For #750 it's OK.

It doesn't (yet), hence the 'Draft' state.

NicolasT:
This is, for now, stuck on design, or rather, how to properly integrate with Salt's authn/authz.

packages/0002-Make-Cache-Control-headers-configurable.patch
```diff
@@ -0,0 +1,81 @@
...
++++ b/saltgui/static/scripts/api.js
+@@ -8,8 +8,12 @@ class HTTPError extends Error {
+ 
+ 
+ class API {
+-  constructor(apiurl="") {
++  constructor(apiurl="", options={}) {
+     this.APIURL = apiurl;
++    this.enableCacheControl =
++        typeof options.enableCacheControl === "undefined"
```
gdemonet: I'm not sure the `typeof` is needed here. Also, this ternary could be rewritten:

```
this.enableCacheControl = options.enableCacheControl === undefined || options.enableCacheControl
```

NicolasT:
@bert-e reset

buildchain/buildchain/salt_tree.py
```diff
@@ -216,10 +218,12 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
 
     Path('salt/_roster/kubernetes_nodes.py'),
 
+    Path('salt/_runners/metalk8s_saltutil.py'),
+
     Path('salt/_states/containerd.py'),
     Path('salt/_states/kubeconfig.py'),
     Path('salt/_states/docker_registry.py'),
-    Path('salt/_states/kubernetes.py'),
+    Path('salt/_states/kubernetesmod.py'),
```
TeddyAndrieux, line 251: Why ? Salt only renamed the kubernetes modules not the state.

Or maybe you got conflict in `_auth/kubernetes_rbac.py`

salt/_auth/kubernetes_rbac.py
```diff
@@ -0,0 +1,69 @@
+import base64
+import logging
+
+log = logging.getLogger(__name__)
+
+try:
+    import kubernetes.client
+    import kubernetes.config
```
TeddyAndrieux: Those import it's only for the futur right ?

salt/metalk8s/salt/master/files/master_99-metalk8s.conf
```diff
@@ -1,9 +1,33 @@
+{% from "metalk8s/map.jinja" import networks with context %}
+
+{% set ip_candidates = salt.network.ip_addrs(cidr=networks.control_plane) %}
```
TeddyAndrieux: I personnaly don't like the fact that we do jinja stuff in a configuration file, should be part of the state and added in `context` or `defaults` of `file.managed`

Also that works ? even if `template` is not set to `jinja` in the state ?

salt/_auth/kubernetes_rbac.py
```diff
@@ -0,0 +1,69 @@
...
+    HAS_DEPS = True
+except ImportError:
+    HAS_DEPS = False
+
+__virtualname__ = 'kubernetes_rbac'
+
+def __virtual__():
+    if not HAS_DEPS:
+        return False, 'Missing dependencies'
```
slaperche-scality: We should mention which deps are missing in the error message, otherwise it's not really useful.

salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2
```diff
@@ -68,6 +72,38 @@ spec:
...
+      args: ['salt-api', '--', '--log-level=info']
+      ports:
+        - name: api
+          containerPort: 4507
+          protocol: TCP
+      livenessProbe:
+        httpGet:
+          path: /
+          port: api
+{% if ip_candidates %}
```
slaperche-scality: Instead of repeating the if/else block many times, you could define a variable and use it instead, no?

tests/post/steps/test_logs.py
```diff
@@ -15,12 +15,14 @@ def check_logs(host):
             '--no-headers -o custom-columns=":metadata.name"'
         )
         for pod_id in pods_list.split('\n'):
+            if 'salt-master' in pod_id:
```
slaperche-scality, line 29: Why do we skip `salt-master`?

Maybe add a comment.

salt/_runners/metalk8s_saltutil.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import, print_function, unicode_literals
```
slaperche-scality, line 1: Is `print_function` used?

salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2
TeddyAndrieux: That exactly the goal of `defaults` and `context` in `file.managed` state

tests/post/steps/test_logs.py
sayf-eddine-scality: we skip salt-master because its logs is in `info` mode so you dont have logs usually

buildchain/buildchain/salt_tree.py
NicolasT: Then there may still an issue in upstream Salt ;-)

salt/_auth/kubernetes_rbac.py
NicolasT: Ugh, right, I had some code in here before which *did* interact with K8s. I'll remove this.

salt/_auth/kubernetes_rbac.py
NicolasT: Given https://github.com/scality/metalk8s/pull/760/files#r273824915 this will be removed.

salt/_runners/metalk8s_saltutil.py
NicolasT: Maybe not, but it's common practice in Salt modules to always include this very line.

salt/metalk8s/salt/master/files/master_99-metalk8s.conf
NicolasT: I believe it works. Also, we're doing the same thing in some other files I believe. Could change though.

Now, given #970, this should be 'fixed' once and for all.

salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2
NicolasT: Fair enough, will do.

tests/post/steps/test_logs.py
NicolasT: So, originally that was indeed the reason @sayf-eddine-scality 

There's no another reason: we run 2 containers in the `salt-master` Pod, which makes `kubectl logs ...` bail out, unless we pass in a container name.

Now, given there's been no interactions with SaltAPI at this point, the logs may again be 'empty' (similar to the original issue), so keeping this as-is. However, I'll add a comment explaining the two reasons (old one and new one).

NicolasT:
@bert-e reset

salt/metalk8s/salt/master/files/master_99-metalk8s.conf
TeddyAndrieux: It's works indeed but I think in the `file.managed` you need to specify `template: jinja` and that not the case for salt-master

salt/metalk8s/salt/master/configured.sls
```diff
@@ -9,6 +17,8 @@ Configure salt master:
     - mode: '0644'
     - makedirs: true
     - backup: false
+    - defaults:
```
TeddyAndrieux, line 21: I think you need to add `template: jinja` here

slaperche-scality: Indeed, and then rename `master_99-metalk8s.conf` to `master_99-metalk8s.conf.j2`

NicolasT: Thanks guys :+1:

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e wait

NicolasT:
@bert-e reset

NicolasT:
@bert-e approve

NicolasT:
@bert-e status

TeddyAndrieux (approved):

Issue #651, opened 2019-02-26
Remove hardcoded paths in salt minion/master confs


sayf-eddine-scality:
the path that points to the fileroots, pillar and states should not be hardcoded and should depend on the metalk8s version

jbertran:
Fileroots tackled in #646

jbertran:
Bumping this to bug, as this makes the bootstrap fail on some stuff now that the expected mountpoint is version-dependent.

Offending statements:

```
salt/metalk8s/registry/populated.sls:35:{% set images_path = '/srv/scality/metalk8s-dev/images' %}
salt/metalk8s/registry/populated.sls:45:    - archive_path: /srv/scality/metalk8s-dev/images/{{ image.name }}-{{ image.tag }}.tar.gz
salt/metalk8s/registry/init.sls:9:    - archive_path: /srv/scality/metalk8s-dev/images/registry-2.7.1.tar
salt/metalk8s/containerd/configured.sls:3:    '/srv/scality/metalk8s-dev/salt/metalk8s/containerd/files/pause-3.1.tar'
```

jbertran:
Current state: in salt files, the version is hard-coded to the correct version (`2.0`). This should depend on short_version instead (as generated in `product.txt`), probably as a custom module whose return values are rendered on build.

We should wait for the build system refactor for this one.

gdemonet:
The `{{ saltenv }}` variable available in SLS files will return what is needed, so we should be able to avoid the need for a custom module. PoC incoming.

gdemonet:
Closed by #761

PR #761, opened 2019-03-13
Improvement: Remove hardcoded version in Salt formulas

gdemonet:
Here is a simplistic solution to the multiple versions problem: we wanted to use multiple Salt environments for having multiple product versions living next to one another. To know which version is being used then, we can simply rely on the `{{ saltenv }}` Jinja variable.

Along the way, we realized another hardcoded info, i.e. *where we mount the ISO*, could not be inferred completely from the active `saltenv`, but instead was a fully customizable value without any implications for the product well-being. We use the newly introduced configuration file to add such info.

Fixes: GH-651

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,7 @@ ALL = \
 	$(ISO_ROOT)/salt/_states/docker_registry.py \
 	$(ISO_ROOT)/salt/_states/kubernetes.py \
 	\
+	$(ISO_ROOT)/pillar/metalk8s.sls \
 	$(ISO_ROOT)/pillar/repositories.sls \
 	$(ISO_ROOT)/pillar/top.sls \
 	\
@@ -221,6 +222,11 @@ $(ISO_ROOT)/salt/%: salt/%
 	mkdir -p $(shell dirname $@)
 	rm -f $@
 	cp -a $< $@
+
+$(ISO_ROOT)/pillar/metalk8s.sls: pillar/metalk8s.sls.in $(ISO_ROOT)/product.txt
+	mkdir -p $(@D)
+	rm -f $@
+	sed s/@VERSION@/$(shell source $(ISO_ROOT)/product.txt && echo $$SHORT_VERSION)/g < $< > $@ || (rm -f $@; false)
 
 $(ISO_ROOT)/pillar/top.sls: pillar/top.sls.in $(ISO_ROOT)/product.txt
 	mkdir -p $(shell dirname $@)
--- a/Makefile
+++ b/Makefile
@@ -213,13 +213,13 @@ all-local: $(ALL) ## Build all artifacts in the build tree
 .PHONY: all-local
 
 $(ISO_ROOT)/bootstrap.sh: scripts/bootstrap.sh.in $(ISO_ROOT)/product.txt
-	mkdir -p $(shell dirname $@)
+	mkdir -p $(@D)
 	rm -f $@
 	sed s/@VERSION@/$(shell source $(ISO_ROOT)/product.txt && echo $$SHORT_VERSION)/g < $< > $@ || (rm -f $@; false)
 	chmod a+x $@
 
 $(ISO_ROOT)/salt/%: salt/%
-	mkdir -p $(shell dirname $@)
+	mkdir -p $(@D)
 	rm -f $@
 	cp -a $< $@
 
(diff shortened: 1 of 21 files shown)


Vagrantfile
```diff
@@ -108,6 +108,7 @@ mkdir -p /etc/metalk8s
 cat > /etc/metalk8s/bootstrap.yaml << EOF
 apiVersion: metalk8s.scality.com/v1alpha1
 kind: BootstrapConfiguration
+isoMountpoint: /srv/scality/metalk8s-$SHORT_VERSION
```
NicolasT: No, we can't have this here. This config is system-wide, across multiple 'versions' of MetalK8s available in the environment.

eve/main.yml
```diff
@@ -221,9 +222,10 @@ stages:
           command: >
             ssh -F bootstrap_ssh_config bootstrap
             sudo mkdir -p /etc/metalk8s &&
-            scp -F bootstrap_ssh_config bootstrap.yaml bootstrap: &&
+            scp -F bootstrap_ssh_config bootstrap.yaml.in bootstrap: &&
```
NicolasT: How about replacing this with something along the lines of

```
sed "s/@VERSION@/%(prop:metalk8s_short_version)s/g" bootstrap.yaml.in | ssh -F bootstrap_ssh_config bootstrap sudo tee /etc/metalk8s/bootstrap.yaml
```

salt/_pillar/metalk8s.py
```diff
@@ -9,25 +9,46 @@
...
+        'pod': networks_data.get('pods', DEFAULT_POD_NETWORK),
+        'service': networks_data.get('services', DEFAULT_SERVICE_NETWORK),
+    }
 
-    assert 'networks' in obj
-    assert 'controlPlane' in obj['networks']
-    assert 'workloadPlane' in obj['networks']
+
+def _load_extra(config_data):
+    assert 'isoMountpoint' in config_data
```
NicolasT: See before.

salt/metalk8s/kubeadm/init/control-plane/lib.sls
```diff
@@ -1,7 +1,7 @@
 # Share macro for building image names for each control-plane component
-{% set repository = "localhost:5000/metalk8s-2.0" %}
-{% set kubernetes_version = "1.11.7" %}
+{%- set registry_url = "localhost:5000" -%}
```
NicolasT: This is not really a URL :-P

salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2
```diff
@@ -1,9 +1,9 @@
 file_roots:
-  metalk8s-{{ metal_version }}:
-    - /srv/scality/metalk8s-2.0/salt
+  {{ saltenv }}:
```
NicolasT, line 2: Once we implement multiple versions and upgrades, this will no longer work correctly (but that's not  regression introduced by this PR, just a heads-up)

NicolasT (changes requested):
The ISO path definition in the bootstrap config won't work...

Vagrantfile
gdemonet: Right, didn't think this through... So this could go in a pillar provided by us? How would a user edit such pillars if they are located in the read-only mounted ISO?

salt/metalk8s/kubeadm/init/control-plane/lib.sls
gdemonet: What would you call it? `registry_host` / `registry_address`?

salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2
gdemonet: Should I create a ticket tracking this as debt? I could split the master conf into multiple files, so that declaring file_roots is done in some `/etc/salt/master.d/99-metalk8s-{{ version }}.conf` which won't get overridden by other Salt environments.

eve/main.yml
gdemonet: It's not needed however, since we're not configuring the mountpoint in this file anymore

Vagrantfile
gdemonet: Moved to a pillar - will need to think about how to customize this...

salt/_pillar/metalk8s.py
gdemonet: Removed

salt/metalk8s/kubeadm/init/control-plane/lib.sls
gdemonet: Used `registry_host`

salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2
gdemonet: Created #800 to track this.

Vagrantfile
NicolasT: Shouldn't this be removed now?

eve/main.yml
```diff
@@ -125,6 +125,7 @@ stages:
             cat > /etc/metalk8s/bootstrap.yaml << END
             apiVersion: metalk8s.scality.com/v1alpha1
             kind: BootstrapConfiguration
+            isoMountpoint: /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
```
NicolasT: Shouldn't this be removed now?

NicolasT (changes requested):
Some leftovers.

Vagrantfile
gdemonet: It should, oops

eve/main.yml
gdemonet: It should, oops

Makefile
```diff
@@ -212,29 +212,34 @@ all-local: $(ALL) ## Build all artifacts in the build tree
...
 	sed s/@VERSION@/$(shell source $(ISO_ROOT)/product.txt && echo $$SHORT_VERSION)/g < $< > $@ || (rm -f $@; false)
 	chmod a+x $@
 
 $(ISO_ROOT)/salt/%: salt/%
-	mkdir -p $(shell dirname $@)
+	mkdir -p $(@D)
 	rm -f $@
 	cp -a $< $@
 
+$(ISO_ROOT)/pillar/metalk8s.sls: pillar/metalk8s.sls.in $(ISO_ROOT)/product.txt
```
NicolasT, line 226: Minor: this could be merged with the rule below for `$(ISO_ROOT)/pillar/%.sls`, but given #708 it's not worth it :)

NicolasT (dismissed):

Makefile
gdemonet: That's what I thought, and we have another `$(ISO_ROOT)/pillar/repositories.sls` which does not follow the same rule, so I was not sure how to write the pattern :/

gdemonet:
@bert-e approve

NicolasT (approved):

Issue #728, opened 2019-03-08
Design & Implement the listing of the nodes


MonPote:
This issue is block by #727 and need more discussion.

- [x]   Validation with Nicolas
- [x]   Design the nodes (wireframe + code)
- [x]   Implement the List component in the core-ui by using react-virtualized
- [x]   Set up K8s Api Server with https://github.com/scality/kubernetes-client-javascript 
- [x]   Implement the nodes list in the Metalk8s bu using  core-ui component
<img width="1440" alt="Screenshot 2019-03-14 at 15 49 39" src="https://user-images.githubusercontent.com/47394132/54429776-c1a7c000-4721-11e9-8cd8-5ac77c9b5560.png">

Issue #727, opened 2019-03-08
Design & Implement login form


MonPote:
Since OIDC will not be ready, the implementation will have to use basic-auth (need discussion)
- [x] Validation with Nicolas
- [x] Design the login (wireframe + code)
- [x] Implement form with redux-saga/formik
- [x] Add Basic Auth

<img width="1440" alt="Screenshot 2019-03-15 at 12 27 57" src="https://user-images.githubusercontent.com/47394132/54429801-d1bf9f80-4721-11e9-9220-5da04d6bd652.png">

Issue #725, opened 2019-03-08
Integrate Platform UI to MetalK8s repository


MonPote:
---
 name: Integrate Platform UI to MetalK8s repository
 about: Integrate Platform UI to MetalK8s repository
 title: 'Integrate Platform UI to MetalK8s repository'
 labels: 'moonshot, UI'
 assignees: ''

---

**GOAL: (What is the problem this ticket solves or the functionality it will add ?)**

[//]: We want to provide an UI for MetalK8s

**Requirements: (What are the nominal use cases)**

[//]: A folder is create and there is at least enough code to run a basic UI (no logic related to MetalK8s) **locally**

Issue #740, opened 2019-03-08
Integrate UI images in the build process


MonPote:
We want to generate the images (build and run) from the Dockerfile.

PR #762, opened 2019-03-14
Integrate Platform UI

carlito-scality:
-https://github.com/scality/metalk8s/issues/725
-https://github.com/scality/metalk8s/issues/727
-https://github.com/scality/metalk8s/issues/728

To test, you have to have a running cluster with basic auth.

Howto

Add  `bootstrap.vm.network "forwarded_port", guest: 6443, host: 8080` into the VagrantFile 

Set up the basic auth:
-----
Create a `htpasswd` file
~~~~~~~~~~~~~~~~~~~~~~~~
Create `/etc/kubernetes/htpasswd`, and add the following content:

```csv
password,user,uid,"system:masters"
```

The format is 'password', 'username', 'unique user ID', 'comma-separated list of groups'.

Note: in the example above, we add the user to the `system:masters` group, which basically gives *root* access to the cluster.

Configure `kube-apiserver` to use HTTP Basic auth
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Create a copy of `/etc/kubernetes/manifests/kube-apiserver.yaml` at some other location, *not in `/etc/kubernetes/manifests`* (so e.g. in `/root/`). We'll need to edit it and want to ensure atomic replacement of the file.

In the copy, perform the following modifications (always ensure proper YAML indentation):

- In the existing `volumes` section, add a new list item:

```yaml
- hostPath:
    path: /etc/kubernetes/htpasswd
    type: File
  name: htpasswd
```

- In the existing `volumeMounts` section, add a new list item:

```yaml
- mountPath: /etc/kubernetes/htpasswd
  name: htpasswd
  readOnly: true
```

- In the existing `command` section, add a new list item:
cors-allowed-origins to avoid CORS problem when calling API server from UI

```yaml
- --basic-auth-file=/etc/kubernetes/htpasswd
- --cors-allowed-origins=http://localhost:3000
```

Save these modifications to the copied file, then copy the modified file to `/etc/kubernetes/manifest/kube-apiserver.yaml`, overwriting the original. At this point, `kubelet` will restart the `Pod`.


NicolasT:
Since you say this is WiP, what's holding back a merge?

As in: I'd rather have the basic UI project in in this PR, without too much functionality, so this can be reviewed, and more functionality can be reviewed in separation as well. If we wait for this to land until a lot of functionality is included, the PR will be too huge to review.

ui/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "scality-ui-template",
```
NicolasT: This can't be right :)

ui/conf/nginx.conf
```diff
@@ -0,0 +1,51 @@
+server {
+    listen       80;
+    server_name  localhost;
+
+    #charset koi8-r;
+    #access_log  /var/log/nginx/host.access.log  main;
+
+    location / {
+        root   /usr/share/nginx/html/sup;
```
NicolasT: What's `sup`?

ui/conf/nginx.conf
```diff
@@ -0,0 +1,51 @@
...
+    #error_page  404              /404.html;
+
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+
+    # proxy the PHP scripts to Apache listening on 127.0.0.1:80
```
NicolasT: No need to keep these comments around, I'd say.

ui/docker/Dockerfile.build
```diff
@@ -0,0 +1,8 @@
+FROM node
```
NicolasT: We'll need to rework the build infrastructure. This can be done in later PRs.

ui/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@fortawesome/fontawesome-free": "^5.7.2",
+    "@kubernetes/client-node": "git://github.com/scality/kubernetes-client-javascript.git#browser",
```
NicolasT, line 7: I guess at some point we'll want to create a tag in that repo, then point to the tag.

ui/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "axios": "^0.18.0",
```
NicolasT, line 8: `kubernetes-client-javascript` uses `fetch`. Do we really need `axios`?

ui/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
```
NicolasT: Nope

ui/public/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Create React App Sample",
```
NicolasT: Also nope :)

ui/src/assets/branding.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+
+const Logo = () => (
```
NicolasT: I have no clue what this thing now looks like. Can't we instead store SVGs as SVG files, and use Webpack's `svg-inline-loader`?

ui/src/containers/App.js
```diff
@@ -0,0 +1,133 @@
...
+
+const messages = {
+  en: translations_en,
+  fr: translations_fr
+};
+
+addLocaleData([...locale_en, ...locale_fr]);
+
+class App extends Component {
+  // componentDidMount = () => {
```
NicolasT: Remove dead/commented code?

ui/src/containers/App.js
```diff
@@ -0,0 +1,133 @@
...
+  //   fetch("/brand/theme.json")
+  //     .then(r => r.json())
+  //     .then(data => {
+  //       console.log("json", data);
+  //     });
+  // };
+
+  render() {
+    const intl = messages[this.props.language];
+    const applications = [{ label: 'Hyperdrive UI', onClick: () => {} }];
```
NicolasT: We'll need to figure out how to handle this. Can't have `Hyperdrive UI` in code in this repo.

ui/src/containers/App.js
```diff
@@ -0,0 +1,133 @@
...
+      onToggleClick: this.props.toggleSidebar,
+      toggleVisible: true,
+      productName: intl.product_name,
+      applications,
+      help,
+      user: this.props.user && user
+    };
+
+    const theme = {
+      brand: {
```
NicolasT, line 82: I'm a bit confused why this is in here, whilst there are other configurations already specifying branding.

ui/src/containers/App.test.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import App from './App';
+
+it('renders without crashing', () => {
```
NicolasT, line 5: Let's make sure to add more tests soon'ish.

ui/src/ducks/login.js
```diff
@@ -0,0 +1,67 @@
...
+        ...state,
+        errors: { authentication: action.payload.message }
+      };
+    default:
+      return state;
+  }
+}
+
+// Action Creators
+export const autheticateAction = payload => {
```
NicolasT: Typo in name.

ui/src/services/api.js
```diff
@@ -0,0 +1,53 @@
+import axios from 'axios';
+import { Config, Core_v1Api } from '@kubernetes/client-node';
+
+const api = axios.create({
+  baseURL: 'https://localhost:8080/api/v1',
+  timeout: 1000
+});
+
+//Basic Auth
+export const authenticate = ({ username, password }) => {
```
NicolasT: If this is the only usage of `axios`, can we remove the dep and simply use `fetch`?

ui/src/services/api.js
```diff
@@ -0,0 +1,53 @@
...
+  localStorage.clear();
+};
+
+export const getNodes = ({ token }) => {
+  const config = new Config('https://localhost:8080', token, 'Basic');
+  const coreV1 = config.makeApiClient(Core_v1Api);
+
+  return coreV1
+    .listNode()
+    .then(nodesResponse => {
```
NicolasT: Consider using `async`/`await` syntax instead of chaining `Promise`s.

ui/src/services/api.js
```diff
@@ -0,0 +1,53 @@
...
+      return nodesResponse.body.items.map(node => ({
```
NicolasT: Hmh... Wouldn't it be better if the code here simply returns plain `Node` objects, and let the renderer code pick whatever fields it wants?

Say we want to display `node.spec.podCIDR` in some widget. Then the code change would require a change in the widget (of course), but also here, which seems odd.

ui/src/services/api.js
```diff
@@ -0,0 +1,53 @@
...
+        name: node.metadata.name,
+        cpu: node.status.capacity.cpu,
+        memory: node.status.capacity.memory,
+        pods: node.status.capacity.pods
+      }));
+    })
+    .catch(error => {
+      console.error('Error retrieving nodes', error.body ? error.body : error);
```
NicolasT: This can be improved upon :) It may make sense to figure out how to do API call error management from the start. Could you spend a bit of time on this?

ui/src/translations/fr.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "Bienvenu",
```
NicolasT: This is not the same as the English version.

NicolasT (changes requested):
First round of feedback.

ui/src/services/api.js
carlito-scality: Yes. We will need a Notifications component to display errors. We can do it later in another ticket

ui/src/containers/App.test.js
carlito-scality: yes! Unit tests is planned 👍

ui/package.json
carlito-scality: https://medium.com/@thejasonfile/fetch-vs-axios-js-for-making-http-requests-2b261cdd3af5

That's is why axios is part of our boilerplate.

ui/src/services/api.js
carlito-scality: I see but I am not fan to have lots of "useless" data in the input of the components.

ui/src/assets/branding.js
carlito-scality: Yes. It is a copy-paste from core-ui where I have some problem when transpiling "svg-inline-loader"...
But here: `import { ReactComponent as Logo } from './logo.svg';`  is enough

ui/src/translations/fr.json
slaperche-scality: Also, it's "Bienvenue"

ui/src/translations/fr.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "Nom",
+  "cpu_capacity": "CPU Capacité",
```
slaperche-scality: Hum, that doesn't make sense in French.
I don't know what we want to express here so I can't propose a better translation (do you mean "Nombre de CPU"?).

ui/package.json
NicolasT: I disagree with that article's second part, which claims that a HTTP 400 response is an 'error': no, it's not. The request was handled just fine (so successful), it's just that the request payload denotes an application-level error which you can easily handle using the `fetch` API.

To decode JSON, it's just a function-call away, and when using `async`/`await` it's literally one line of code. Also, a library behind the scenes decoding JSON is icky IMHO.

Anyway, if you want to stick with Axios, fine. I just doubt it's worth the extra dep.

ui/src/assets/branding.js
NicolasT: Let's try that then? I prefer SVGs to be 'normal' files, otherwise it's very hard to edit them and such.

ui/src/services/api.js
NicolasT: Then do the mangling somewhere else. This is an API accessing module, so it should return API objects as-is, IMHO.

Also, you're not really giving 'useless' input: your component gets (a list of) `Node` objects, that's it. Which parts they consume from such object is up to the component.

ui/src/services/api.js
NicolasT: I think this one is one where it would make sense to handle it in this PR already. What do you tink @MonPote @scality/metalk8s

ui/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+
+# production
+/build
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
gdemonet: _(minor)_ Could do with a glob here:

```diff
- .env.local
- .env.development.local
- .env.test.local
- .env.production.local
+ .env*.local
```

ui/docker/README_BUILD.md
```diff
@@ -0,0 +1,17 @@
+# Building the app
+
+The Dockerfile is here to help you to generate the app's build folder.
+
+## Usage
+
+Run the following command in the docker folder :
```
gdemonet: Would be nice to have appropriate `make` targets for this, let's create a ticket for this.

ui/src/containers/App.js
carlito-scality: let's remove it

ui/src/services/api.js
carlito-scality: https://github.com/scality/metalk8s/issues/774

ui/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "metalk8s-platform-ui",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@fortawesome/fontawesome-free": "^5.7.2",
+    "@kubernetes/client-node": "git://github.com/scality/kubernetes-client-javascript.git#browser",
```
gdemonet, line 7:
```suggestion
    "@kubernetes/client-node": "scality/kubernetes-client-javascript.git#browser",
```

See this shorthand notation for GitHub [here](https://docs.npmjs.com/files/package.json#github-urls)

ui/docker/Dockerfile.build
MonPote: I'm working on it. It will arrive in my next PR very soon.

ui/conf/nginx.conf
MonPote: I just removed the sup part.

ui/conf/nginx.conf
MonPote: I just removed those comments.

ui/src/services/api.js
carlito-scality: OK to move it else where.
The thing is we store the Node objects in the Redux store, I think it is better to keep it as small as possible

ui/src/translations/fr.json
```diff
@@ -0,0 +1,13 @@
...
+  "memory": "Mémoire",
+  "pods_number": "Nombre de pods",
+  "username": "Nom d'utilisateur",
+  "password": "Mot de passe",
+  "submit": "Se connecter",
+  "about": "À propos",
+  "nodes": "Nodes",
+  "log_out": "Se Déconnecter"
```
slaperche-scality: "Se déconnecter" (no uppercase D, like "Se connecter" (no uppercase C)), or simply" Déconnexion" (but then, "Se connecter" should be "Connexion").

ui/src/ducks/nodes.js
```diff
@@ -0,0 +1,41 @@
...
+// Sagas
+function* fetchNodes(token) {
+  try {
+    const nodes = yield call(Api.getNodes, token);
+    yield put(setNodesAction(nodes));
+  } catch (e) {}
+}
+
+export function* nodesSaga() {
+  yield takeEvery(FETCH_NODES, fetchNodes);
```
gdemonet: Why not use a `takeLatest` or something similar instead for (potentially expensive) API calls? This would emulate debouncing of the `fetchNodes` action without too much hassle.

ui/src/containers/App.js
carlito-scality: We will tackle the branding part later.

ui/public/manifest.json
MonPote: Fixed :)

ui/public/manifest.json
MonPote: Fixed :)

ui/src/services/api.js
```diff
@@ -0,0 +1,50 @@
...
+    localStorage.setItem('token', token);
+    return {
+      response: {
+        username,
+        password,
+        token
+      }
+    };
+  } catch (errors) {
+    return { errors: errors.response.data };
```
gdemonet: Let's try to be consistent in those API async methods: either perform the `try/catch` control flow in here, and return an object with `response` and `errors` keys, or perform the `try/catch` in the Saga generator, and only try to `return await <potentiallyBreakingAPICall>()` in here.

gdemonet (changes requested):
Most of my comments are minor, I'd suggest we solve the uniformity in API async methods in this PR though.

ui/docker/README_BUILD.md
carlito-scality: @MonPote could you add it in your PR?

ui/docker/Dockerfile.run
```diff
@@ -0,0 +1,15 @@
+FROM nginx
+
+ENV LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8
+
+RUN apt update
+RUN apt install -y vim
+
+RUN mkdir -p /usr/share/nginx/html/sup
```
NicolasT: So, really, what's this `sup` thin? :wink:

ui/docker/Dockerfile.run
```diff
@@ -0,0 +1,15 @@
+FROM nginx
+
+ENV LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8
+
+RUN apt update
+RUN apt install -y vim
```
NicolasT: Not quite sure you need this in a webserver environment :confused:

ui/src/services/api.js
```diff
@@ -0,0 +1,35 @@
...
+    return { error };
+  }
+}
+
+export const logout = () => {
+  localStorage.removeItem('token');
+};
+
+export async function getNodes({ token }) {
+  const config = new Config('https://localhost:8080', token, 'Basic');
```
NicolasT: It's not really required to create a new `Config` instance for every API call. Rather, create a `Config`, keep it as some variable, and pass it in for every API call.

ui/src/services/api.js
```diff
@@ -0,0 +1,35 @@
...
+  const coreV1 = config.makeApiClient(Core_v1Api);
```
NicolasT: Same for this, actually (cfr. https://github.com/scality/metalk8s/pull/762/files#r266377673).

gdemonet (dismissed):

ui/docker/README_BUILD.md
MonPote: Docker folder is outdated. There is already a issue for this ^^ (cf.  #740)

ui/docker/Dockerfile.run
MonPote: docker folder is outdated, I deleted this folder. For the build part here is the PR : #782

ui/docker/Dockerfile.run
MonPote: docker folder is outdated, I deleted this folder. For the build part here is the PR : #782

ui/src/services/api.js
```diff
@@ -0,0 +1,39 @@
+import axios from 'axios';
+import { Config, Core_v1Api } from '@kubernetes/client-node';
+
+let config, coreV1;
```
NicolasT, line 4: Please have a ticket to get rid of global state, preferably...

ui/src/translations/en.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "Metalk8s Platform",
```
NicolasT: It's actually `MetalK8s`.

ui/src/translations/fr.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "Platforme Metalk8s",
```
NicolasT: It's actually `MetalK8s`.

NicolasT (changes requested):
Please take care of replacing `Metalk8s` with `MetalK8s`, otherwise OK for now (though some comments left, for which most have tracking tickets already).

ui/src/services/api.js
carlito-scality: I do not know how to get rid of them? :)

NicolasT: Construct them at some kind of entry-point of your application, thread them through, and pass when invoking e.g. `getNodes`.

NicolasT (approved):
Let's get this one in and schedule some follow-up holistic reviews.

carlito-scality:
@bert-e approve

PR #764, opened 2019-03-14
Test run bootstrap twice

### LINKED ISSUE #741: Check that running the bootstrap twice is working
Since the bootstrap should be idempotent, running it twice should work as usual.

Blocked by #717 since this double running should be run as a test and not as part of the default deployment process.

ramanduh:
Run a successive run of bootstrap, result we should not have error.

To test: Run the tests as described in `README.md`

Note: test can take time as the bootstrap is run twice

Closes #741

### DIFF
--- a/tests/post/steps/conftest.py
+++ b/tests/post/steps/conftest.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+from configparser import ConfigParser
+from itertools import chain
+
+import pytest
+from pytest_bdd import(
+    given,
+    then,
+    when,
+)
+
+
+# Pytest fixtures
+@pytest.fixture(scope="session")
+def version():
+    """Fixture that read metalk8s version.
+
+    Version is extracted from _build/root/product.txt
+    """
+    version_file = "_build/root/product.txt"
+    parser = ConfigParser()
+
+    with open(version_file) as stream:
+        # add a section header for configparser
+        stream = chain(("[top]",), stream)
+        parser.read_file(stream)
+
+    return dict(parser.items('top'))
+
+
+def _verify_kubeapi_service(host):
+    """Verify that the kubeapi service answer"""
+    with host.sudo():
+        cmd = "kubectl --kubeconfig=/etc/kubernetes/admin.conf cluster-info"
+        retcode = host.run(cmd).rc
+        assert retcode == 0
+
+
+def _run_bootstrap(host, version):
+    cmd = "/srv/scality/metalk8s-{0}/bootstrap.sh".format(
+        version['short_version']
+    )
+    with host.sudo():
+        res = host.run(cmd)
+        assert res.rc == 0, res.stdout
+
+
+# Pytest-bdd steps
+
+# Given
+@given('bootstrap was run once')
+def run_bootstrap(host, version):
+    _run_bootstrap(host, version)
+
+
+@given("the Kubernetes API is available")
+def check_service(host):
+    _verify_kubeapi_service(host)
+
+
+# When
+@when('we run bootstrap a second time')
+def rerun_bootstrap(host, version):
+    _run_bootstrap(host, version)
+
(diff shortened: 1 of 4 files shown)


tests/post/steps/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+    with open(version_file) as stream:
+        # add a section header for configparser
+        stream = chain(("[top]",), stream)
+        parser.readfp(stream)
+
+    return {k: v for k, v in parser.items('top')}
+
+
+@pytest.fixture
+def verify_kubeapi_service(host):
```
sayf-eddine-scality: we can put common steps in the conftest.py (that is why I created it in the first place)

tests/post/steps/test_bootstrap.py
```diff
@@ -0,0 +1,39 @@
+from pytest_bdd import (
+    given,
+    scenario,
+    then,
+    when,
+)
+
+
+def _run_bootstrap(host, version):
```
sayf-eddine-scality: again, I think this should go to conftest.py

sayf-eddine-scality (changes requested):

tests/post/steps/conftest.py
ramanduh: In fact I is not really common steps as it is used by a `Given` steps in the first test and a `Then` step in the 2nd test. I did not find a way to alias the 2 steps, that's why I create the common fixture in the `conftest.py`

ramanduh: I moved it to the conftest

tests/post/steps/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Version is extracted from _build/root/product.txt
+    """
+    version_file = "_build/root/product.txt"
+    parser = ConfigParser()
+
+    with open(version_file) as stream:
+        # add a section header for configparser
+        stream = chain(("[top]",), stream)
+        parser.readfp(stream)
```
sayf-eddine-scality:
```
metalk8s/tests/post/steps/conftest.py:26: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
    parser.readfp(stream)
```

tests/post/steps/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+
+    return {k: v for k, v in parser.items('top')}
```
sayf-eddine-scality: parser.items() is a dict ? why not returning it directly ?

sayf-eddine-scality (changes requested):
ytest_bdd.exceptions.StepDefinitionNotFoundError: Step definition is not found: Given "the Kubernetes API is available". Line 8 in scenario "Exec in Pods" in the feature "metalk8s/tests/post/features/pods_alive.feature

tests/post/steps/conftest.py
ramanduh: fixed

tests/post/steps/conftest.py
ramanduh: It is a list. Something like that:
```
[('name', 'MetalK8s'), ('version', '2.0.0-dev'), ('short_version', '2.0'), ('git', '2.0.0-alpha1-7-gbb0bf9c'), ('development_release', '1'), ('build_timestamp', '2019-03-13T17:09:47Z'), ('build_host', 'duhamel-Latitude-E7450')]
```

tests/post/steps/test_bootstrap.py
```diff
@@ -0,0 +1,12 @@
+from pytest_bdd import (
+    given,
+    scenario,
+    then,
```
sayf-eddine-scality: not important but unused imports

sayf-eddine-scality (approved):

ramanduh:
@bert-e approve

Issue #770, opened 2019-03-15
Use local GPG key for tini image validation


jbertran:
Contacting the external GPG server for validation of the tini binary that we use as PID1 process for metalk8s 2.0 is _very_ flaky. 

To remove that flakiness, the GPG key should be present locally and used locally to verify that binary.

Affects 2.0 only, see #599 comments for more details.

jbertran:
Closed #771

PR #771, opened 2019-03-15
Use local GPG key for tini binary validation

jbertran:
Fixes: GH-770

This frees us from the flakiness of the distant keyserver.

Test with `docker build images/salt-master` in project root.


slaperche-scality (commented):
You should explain *why* we do that in the commit message.

You did it in the PR description (but PR description and GitHub issues are useless when working offline), so you could simply copy your "This frees us from the flakiness of the distant keyserver." in the commit msg.

images/salt-master/Dockerfile
```diff
@@ -10,8 +10,8 @@ ARG TINI_VERSION=v0.18.0
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /tini.asc
 
-RUN gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
- && gpg --batch --verify /tini.asc /tini
+COPY source.key source.key
```
sayf-eddine-scality, line 13: shouldnt the image make target depends on this file too ? ;)

sayf-eddine-scality (changes requested):

images/salt-master/Dockerfile
jbertran: Well... it doesn't _have_ to.

I guess it makes sense to have a more sensible error message than a gpg error message when building if the repo ever gets to a state where the key is removed, but on the other hand we have no way to "build" it anyway so there's little point to having it function as a dependency.

For comprehensiveness though, I'll add it.

sayf-eddine-scality: I meant in case you change the key file the container img wont get rebuilt

sayf-eddine-scality (dismissed):

slaperche-scality (approved):

jbertran:
@bert-e approve

gdemonet (approved):
LGTM, I would add a minor suggestion: `source.key` is not really explicit for a filename, maybe something like `tini_gpg.key`?

gdemonet:
Once again, I'm late to the party

Issue #769, opened 2019-03-15
Configure Vagrant VMs with control-plane and workload-plane networks


NicolasT:
Bugs like #768 would be found earlier if we'd 'properly' emulate a real networking setup.

charlesprost:
@NicolasT 

something like this? I've tested it, it's working. We are limited to 8 network interfaces per VM. That should be enough.

```
diff --git a/Vagrantfile b/Vagrantfile
index f8b11cf..717b5ef 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -48,7 +48,9 @@ echo "Launching bootstrap"
 exec "/srv/scality/metalk8s-$SHORT_VERSION/bootstrap.sh"
 SCRIPT
 
-NETWORK_RANGE = "192.168.42."
+MAIN_RANGE = "192.168.42."
+CONTROL_PLANE_RANGE = "192.168.43."
+WORKLOAD_PLANE_RANGE = "192.168.44."
 BOOTSTRAP_IP_OFFSET = 10
 
 # To support VirtualBox linked clones
@@ -67,8 +69,12 @@ Vagrant.configure("2") do |config|
   config.vm.define :bootstrap, primary: true do |bootstrap|
     bootstrap.vm.hostname = "bootstrap"
 
-    bootstrap_ip = "#{NETWORK_RANGE}#{BOOTSTRAP_IP_OFFSET}"
-    bootstrap.vm.network "private_network", ip: bootstrap_ip
+    bootstrap_main_ip = "#{MAIN_RANGE}#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_main_ip
+    bootstrap_control_plane_ip = "#{CONTROL_PLANE_RANGE}#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_control_plane_ip
+    bootstrap_workload_plane_ip = "#{WORKLOAD_PLANE_RANGE}#{BOOTSTRAP_IP_OFFSET}"
+    bootstrap.vm.network "private_network", ip: bootstrap_workload_plane_ip
 
     bootstrap.vm.provider "virtualbox" do |v|
       v.memory = 4096
@@ -86,12 +92,16 @@ Vagrant.configure("2") do |config|
 
   (1..9).each do |i|
     node_name = "node#{i}"
[...]

PR #778, opened 2019-03-15
Improvement/724 add multiple vms in ci

### LINKED ISSUE #724: Provide a framework to launch additional VMs

charlesprost:
This PR provides a new stage in the CI with multi-VMs spawning.

Fixes #724 

In the single VM environment, the CI worker is also the bootstrap node. Now we spawn thanks to terraform a whole new infra in the PTF9 tenant on which we run the bootstrap script (we run it on one node only of course).

Limitations:

- All nodes can communicate freely on an internal network (192.168.42.0/24); that may be not enough after Nicolas's PR related to the control and workload plane networks; it seems without running a `dhclient` command the IPs are not attributed on this network so the command is run but that may not be a best practice
- Nodes are all identical in sizing, but we can change this later
- The bootstrap node does not contain the SSH private key in order to SSH into the other nodes, but that's easy to do if we want to do it

How to test it?

Just check the steps in the build, or you can edit the debug step in the `eve/main.yaml` file to enable it and then connect into the VM worker. But you won't be able to use terraform commands since they need to be launched by the worker to use the env variables.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -92,13 +92,13 @@ stages:
       - ShellCommand:
           name: Retrieve ISO image
           command: >
-            curl -XGET -o metalk8s.iso
+            curl -s -XGET -o metalk8s.iso
             "%(prop:artifacts_private_url)s/metalk8s.iso"
           haltOnFailure: true
       - ShellCommand:
           name: Retrieve ISO image checksum
           command: >
-            curl -XGET -o SHA256SUM
+            curl -s -XGET -o SHA256SUM
             "%(prop:artifacts_private_url)s/SHA256SUM"
           haltOnFailure: true
       - ShellCommand:
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -177,6 +177,7 @@ stages:
             OS_PASSWORD: "%(secret:scality_cloud_password)s"
             OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
             TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+            TF_VAR_nodes_count: "2"
           haltOnFailure: true
       - ShellCommand:
           name: Configure SSH config for bootstrap node
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -24,9 +24,10 @@ stages:
             . <(curl -s "%(prop:artifacts_private_url)s/product.txt") &&
             echo $SHORT_VERSION'
       - TriggerStages:
-          name: Trigger single-node deployment with built ISO
+          name: Trigger single-node and multiple-nodes steps with built ISO
           stage_names:
             - single-node
+            - multiple-nodes
 
   build:
     worker:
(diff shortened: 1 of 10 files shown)


eve/main.yml
```diff
@@ -130,3 +131,103 @@ stages:
             echo "%(secret:ssh_pub_keys)s" >> ~/.ssh/authorized_keys &&
             sleep 3600
           doStepIf: false
+
+  multiple-nodes:
+    worker:
+      type: openstack
```
NicolasT, line 150: Do I get it correct this node is an OpenStack VM, but it's not used as the bootstrap node itself? If that's the case, can it either be the bootstrap node, or can we run this in a container?

eve/workers/openstack-multiple-nodes/terraform/network.tf
```diff
@@ -0,0 +1,85 @@
...
+  direction         = "ingress"
+  ethertype         = "IPv4"
+  protocol          = "icmp"
+  remote_ip_prefix  = "0.0.0.0/0"
+  security_group_id = "${openstack_networking_secgroup_v2.nodes.id}"
+
+# Main network for nodes
+variable "internal_network_range" {
+  type = "string"
+  default = "192.168.42.0/24"
```
NicolasT, line 35: Once #769 lands, you'll need to create an appropriate initial configuration file on the bootstrap node.

eve/main.yml
charlesprost: Yeah that's right.

In the `single-node` stage, the CI worker (an OpenStack VM) is used as the bootstrap node itself. There is no VM created on top of this worker.

In the new `multiple-nodes` stage, the CI worker (still an OpenStack VM) is used as some kind of bastion to instantiate a fresh environment of VMs with one bootstrap node and X (2 by default) additional nodes.

In this new stage, we are in the same situation as the developer on its laptop running `vagrant`. There is no mix between the source code, and the ISO downloaded on the bootstrap node.

I'd prefer that we stay this way in the `multiple-nodes` stage since I think we want to have a clearly distinct env on which we can really test a fresh install.

eve/workers/openstack-multiple-nodes/terraform/network.tf
charlesprost: Yep.

eve/main.yml
```diff
@@ -24,9 +24,10 @@ stages:
             . <(curl -s "%(prop:artifacts_private_url)s/product.txt") &&
```
jbertran, line 24: Minor (can't do general comments on a commit, that sucks): commit title too long, which messes with github title rendering in commit view. <50 chars is best.

eve/main.yml
```diff
@@ -90,19 +90,19 @@ stages:
       flavor: m1.medium
       path: eve/workers/openstack-single-node
     steps:
-      - ShellCommand:
+      - ShellCommand: &retrieve_iso
```
jbertran, line 93: Minor: since we're starting to reuse these, we could have a YAML section dedicated to reused snippets - see RING `main.yaml` for an example of that.

eve/main.yml
charlesprost: Will check next time :+1:

eve/main.yml
charlesprost: Will keep that in mind for next steps, indeed it could be cool to merge some common steps!

charlesprost:
@bert-e approve

NicolasT (approved):

eve/main.yml
```diff
@@ -143,3 +144,70 @@ stages:
...
+      flavor: m1.medium
+      path: eve/workers/openstack-multiple-nodes
+    steps:
+      - Git: *git_pull
+      - ShellCommand:
+          name: Init terraform
+          command: "terraform init"
+          workdir: "build/eve/workers/openstack-multiple-nodes/terraform/"
+          haltOnFailure: true
+      - ShellCommand:
```
jbertran, line 164: Do we really need that step in a CI context? IMO any change to the terraform plans will be validated beforehand, so we can trust that it will always succeed.

eve/main.yml
```diff
@@ -143,3 +144,70 @@ stages:
...
+          name: Destroy openstack virtual infra
+          command: |-
+            for _ in $(seq 1 3); do
+               terraform destroy -auto-approve && break
+            done;
+          workdir: "build/eve/workers/openstack-multiple-nodes/terraform/"
+          env: *terraform_spawn_multiple_nodes
+          alwaysRun: true
+          sigtermTime: 600
+      - ShellCommand:
```
jbertran, line 247: The debug step should probably take place _before_ the terraform destruction, since we're not really interested in the bastion machine but rather its spawned nodes.

jbertran (changes requested):

eve/main.yml
charlesprost: Yeah it's not really mandatory but it can help separate syntax errors from deployment errors. Since the step is executed in less than 1sec, it's not really an issue.

eve/main.yml
charlesprost: Depends on what you want to debug. And since any debugging will need a manual edition of the file, then I think moving the debug step wherever you want is acceptable.
I'm perfectly OK if in a next PR we move the debug step before the destruction though.

PR #783, opened 2019-03-18
CI: upgrade RAM for build worker

charlesprost:
Building the ISO is very memory-consuming.
500MB limit was hit a few times.


sayf-eddine-scality (approved):

charlesprost:
@bert-e approve

Issue #777, opened 2019-03-15
UI code improvement


carlito-scality:
- [x] add unit tests for saga part
- [x] use core-ui theme in styled-component CSS
- [x] reducer hierarchy
- [x] Nodes memory value should be more human readable

PR #784, opened 2019-03-18
UI Code Optimizing

carlito-scality:
https://github.com/scality/metalk8s/issues/777
- add unit tests for saga part
- use core-ui theme in styled-component CSS
- reducer hierarchy managment
- Nodes memory value should be more human readable

Screenshot 2019-03-19 at 16 02 43


ui/src/containers/App.js
```diff
@@ -17,7 +17,7 @@ import Welcome from '../components/Welcome';
 import Login from './Login';
 import PrivateRoute from './PrivateRoute';
 import { logoutAction } from '../ducks/login';
-import { toggleSidebarAction } from '../ducks/app';
+import { toggleSidebarAction } from '../ducks/layout';
```
gdemonet: If the `layout` sub-state is living under `app`, then maybe the file structure should reflect it as well: `../ducks/app/layout`

ui/src/ducks/reducer.js
```diff
@@ -3,13 +3,15 @@ import { combineReducers } from 'redux';
...
 import login from './login';
-import app from './app';
+import layout from './layout';
 
 const rootReducer = combineReducers({
   language,
   login,
-  nodes,
-  app
+  app: combineReducers({
```
gdemonet, line 11: Same comment as for the `layout`, the file structure of ducks should represent the Redux state structure, for easier navigation and debugging.

ui/src/ducks/login.js
```diff
@@ -61,6 +70,13 @@ function* authenticate({ payload }) {
         token
       }
     });
+    const config = new Config('https://localhost:8080', token, 'Basic');
+    const coreV1 = config.makeApiClient(Core_v1Api);
+    yield put({
+      type: UPDATE_API_COREV1,
+      payload: coreV1
```
gdemonet: Is this a serializable datatype? If not, it would be nice (but not necessary) to have something serializable in Redux (for investigation/debugging/testing) and build the JS object from it to perform an API call.

ui/src/ducks/login.js
```diff
@@ -7,11 +9,13 @@ const AUTHENTICATE = 'AUTHENTICATE';
...
 const AUTHENTICATION_FAILED = 'AUTHENTICATION_FAILED';
 const LOGOUT = 'LOGOUT';
+const UPDATE_API_COREV1 = 'UPDATE_API_COREV1';
 
 // Reducer
 const defaultState = {
   user: null,
-  error: null
+  error: null,
+  api: null
```
gdemonet: So we have a Redux state under `state.login` which, I would guess, holds information about the active user being logged in (or not).

And in this state, I'm storing, for some reason, the API object I'm going to use throughout my application for retrieving data. Are the configuration options only user-related? I'm a bit confused why every API call now will have to do some `select(state => state.login.api)` to retrieve its client... But I guess it's acceptable, just not used to it.

gdemonet (changes requested):
I think we need to be focused on managing properly the Redux state from the earliest steps. Let's make sure the `src/ducks` file structure matches what we have in the Redux store, so we are always aware of where things are happening and if things are getting out of control.

ui/src/ducks/login.js
gdemonet: _Note: it's a minor remark, and may not need to land in this PR if you find some alternative_

ui/src/containers/App.js
carlito-scality: good point

ui/src/ducks/login.js
carlito-scality: not only. api configuration depends on the url endpoint and the user token.
-const config = new Config('https://localhost:8080', token, 'Basic');
-const coreV1 = config.makeApiClient(Core_v1Api);
In the common way, I would store only Config in the redux store and pass it as parameters for API calls. It means we have to duplicate "config.makeApiClient(Core_v1Api);" for every API

ui/src/ducks/login.js
carlito-scality: https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

gdemonet: Maybe not that minor then ^^

ui/src/ducks/login.js
gdemonet: Why not just store the token in Redux? I know it means re-creating a "client" object for each call, but I don't think the computational burden would be too high compared to the HTTP request RTT. And having an active `token` in the `login` state makes a whole lot more sense to me, don't you agree?

carlito-scality: The token is already in Redux :) 
These changes is to avoid re-creating a "client" object for each call. But we have to find a better way the storing "client" object in Redux.
cf https://github.com/scality/metalk8s/pull/762#discussion_r266535836

ui/src/ducks/login.js
carlito-scality: Yep it is not the good idea to stored it "as is" in Redux

ui/src/ducks/login.js
carlito-scality: @NicolasT what do you think?

NicolasT: Not a big fan of re-creating such 'client' objects and API wrappers on every invocation, TBH. Maybe this stuff doesn't fit in the Redux model/store/whatnot, but then that's a limitation of the design IMHO. Yes, compared to performing a query, the overhead (at least time-wise) is minimal, however it's bad form to re-create a (basically) constant object time and time again. If it can't go in the Redux store, then maybe it should go somewhere else with a (serializable) reference to the actual object in the store.

carlito-scality: OK I'll revert these changes and create a debt ticket for it.

ui/src/containers/App.js
```diff
@@ -28,98 +21,22 @@ addLocaleData([...locale_en, ...locale_fr]);
...
         <Switch>
           <Route path="/login" component={Login} />
-          <ThemeProvider theme={theme}>
-            <Layout sidebar={sidebar} navbar={navbar}>
-              <PrivateRoute exact path="/nodes" component={NodeList} />
-              <PrivateRoute exact path="/about" component={Welcome} />
-              <PrivateRoute exact path="/" component={NodeList} />
-            </Layout>
-          </ThemeProvider>
+          <Layout />
```
gdemonet: Wouldn't it be cleaner/safer to do:

```suggestion
          <Route component={Layout} />
```

And include a `Switch` in `Layout`? I think it would best contribute to the separation of concerns and decoupling between `App` and `Layout`.

ui/src/translations/fr.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Platforme MetalK8s",
+  "product_name": "Platforme MetalK8s",
```
gdemonet:
```suggestion
  "product_name": "Plateforme MetalK8s",
```

gdemonet (changes requested):

gdemonet (approved):

MonPote (approved):

carlito-scality:
approve

carlito-scality:
@bert-e approve

carlito-scality:
@bert-e reset

carlito-scality:
@bert-e approve

PR #790, opened 2019-03-19
test pods count and liveness

### LINKED ISSUE #722: Check after deployment that pod number is available through the API
- [ ] Following #716 : have this test in Vagrant environment
- [ ] Following #717 : have this test in CI environment

After a deployment, verify that we have the following pods:
* calico node: `kubectl --kubeconfig=/etc/kubernetes/admin.conf get pod -l 'k8s-app=calico-node' -n kube-system`
* coredns: `kubectl --kubeconfig=/etc/kubernetes/admin.conf get pod -l 'k8s-app=kube-dns' -n kube-system`
* etcd
* kube-scheduler
* kube-controller-manager
* kube-apiserver
* kube-proxy
* salt-master
* package-repositories
* registry

Success is defined having the expected amount of pods reported by the API call.

### LINKED ISSUE #723: Check after deployment that pod list is available
- [ ] Following #716 : have this test in Vagrant environment
- [ ] Following #717 : have this test in CI environment

Success is defined as having a pod list matching expected pods when making the API call.

ramanduh:
Add a test to verify that we have all expected pods running

Closes #722, closes #723

### DIFF
--- a/tests/post/features/pods_alive.feature
+++ b/tests/post/features/pods_alive.feature
@@ -7,3 +7,16 @@ Feature: Pods should be alive
     Scenario: Exec in Pods
         Given the Kubernetes API is available
         Then we can exec 'true' in the 'salt-master-bootstrap' pod in the 'kube-system' namespace
+
+    Scenario: Expected Pods
+        Given the Kubernetes API is available
+        Then we have at least 1 running pod labeled 'component=kube-controller-manager'
+        And we have at least 1 running pod labeled 'component=kube-scheduler'
+        And we have at least 1 running pod labeled 'component=kube-apiserver'
+        And we have at least 1 running pod labeled 'component=etcd'
+        And we have at least 1 running pod labeled 'k8s-app=kube-proxy'
+        And we have at least 1 running pod labeled 'k8s-app=calico-node'
+        And we have at least 2 running pod labeled 'k8s-app=kube-dns'
+        And we have at least 1 running pod labeled 'app=registry'
+        And we have at least 1 running pod labeled 'app=salt-master'
+        And we have at least 1 running pod labeled 'app=package-repositories'
--- a/tests/post/steps/test_liveness.py
+++ b/tests/post/steps/test_liveness.py
@@ -1,12 +1,22 @@
+import json
+import time
+
 from pytest_bdd import scenario, then, parsers
+
 
 # Scenarios
 @scenario('../features/pods_alive.feature', 'List Pods')
 def test_list_pods(host):
     pass
 
+
 @scenario('../features/pods_alive.feature', 'Exec in Pods')
 def test_exec_in_pods(host):
+    pass
+
+
+@scenario('../features/pods_alive.feature', 'Expected Pods')
+def test_expected_pods(host):
     pass
 
 
@@ -19,6 +29,7 @@ def check_resource_list(host, resource, namespace):
         cmd_res = host.check_output(cmd.format(resource, namespace))
     assert len(cmd_res.strip()) > 0, 'No {0} found in namespace {1}'.format(
             resource, namespace)
+
 
 @then(parsers.parse(
     "we can exec '{command}' in the '{pod}' pod in the '{namespace}' namespace"))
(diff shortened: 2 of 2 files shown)


tests/post/steps/test_liveness.py
```diff
@@ -34,3 +45,43 @@ def check_exec(host, command, pod, namespace):
...
+            assert ret.rc == 0, ret.stdout
+
+            pods = json.loads(ret.stdout)
+
+            if len(pods['items']) < expected_nb_pods:
+                time.sleep(interval)
+            else:
+                break
+
+        else:
```
NicolasT, line 81: Why is this an `else` block?

tests/post/features/pods_alive.feature
```diff
@@ -7,3 +7,16 @@ Feature: Pods should be alive
     Scenario: Exec in Pods
         Given the Kubernetes API is available
         Then we can exec 'true' in the 'salt-master-bootstrap' pod in the 'kube-system' namespace
+
+    Scenario: Expected Pods
+        Given the Kubernetes API is available
+        Then we have 1 running pod labeled 'component=kube-controller-manager'
```
NicolasT: Not very fond of hard-coding an expected 'number of nodes' (basically) in a test. This test won't be portable across single-node and multi-node test environments.

tests/post/steps/test_liveness.py
```diff
@@ -34,3 +45,43 @@ def check_exec(host, command, pod, namespace):
...
+        '--namespace kube-system -o json'
+    ).format(label=label)
+
+    # Just after the deployment, the pods may not in running state yet
+    # so implement a timeout
+    interval = 3
+    attempts = 10
+
+    with host.sudo():
+        for _ in range(attempts):
```
sayf-eddine-scality, line 68: shouldnt this be easier with a `while True` ?

tests/post/features/pods_alive.feature
sayf-eddine-scality: Hmm what about creating a test case per host, I guess we can filter on the host name and determine how many pods there should be, or at least a control plane should have > 1 pod for each of the control plane components

sayf-eddine-scality (commented):
I believe we should start using python-kubernetes more often to get rid of parsing kubectl commands, we cannot be sure kubectl will keep the same output format for how long @NicolasT any thoughts ?

tests/post/steps/test_liveness.py
ramanduh: I don't know. I think it just a matter of taste. I prefer the `for` loop because you don't have to manipulate the `attempts` variable. With a `while True`, we'll have to decrease manually the value of `attempts` in the loop and then compare it to 0 to know if we reach the end of attempts or not

tests/post/steps/test_liveness.py
ramanduh: This is a `else` of the `for` block. Without the `else` we don't know if the loop ends because of the break (pods are up) or because we reach the max attempts (pods still down)

tests/post/features/pods_alive.feature
ramanduh: @NicolasT agree with you but for now, I don't know exactly what we should expect on multi-node environments and I can't find a more suitable test input. Unless you have suggestion, I'll let it as it was, and I'll update the test accordingly when we have a multi node environments

NicolasT: For several of the Pods you look for, we want *at least one* instance. Don't really care about the actual count.

ramanduh: Ok, thanks, I updated it to verify that we have at least one pods except for the dns 2

tests/post/features/pods_alive.feature
```diff
@@ -7,3 +7,16 @@ Feature: Pods should be alive
...
+
+    Scenario: Expected Pods
+        Given the Kubernetes API is available
+        Then we have at least 1 running pod labeled 'component=kube-controller-manager'
+        And we have at least 1 running pod labeled 'component=kube-scheduler'
+        And we have at least 1 running pod labeled 'component=kube-apiserver'
+        And we have at least 1 running pod labeled 'component=etcd'
+        And we have at least 1 running pod labeled 'k8s-app=kube-proxy'
+        And we have at least 1 running pod labeled 'k8s-app=calico-node'
+        And we have at least 2 running pod labeled 'k8s-app=kube-dns'
```
NicolasT, line 19: Cosmetic: `pods` :wink:

NicolasT (approved):

ramanduh:
@bert-e approve

sayf-eddine-scality (approved):

Issue #781, opened 2019-03-18
Launch E2E tests in CI


charlesprost:
Needs #717 

E2E tests are currently available for Vagrant environment through a `tox` manual command.

We want to have them automatically executed on the `single-node` and `multiple-nodes` steps of the CI as part of the pre-merge tests.

Success is defined by having those tests executed in every build in the CI.

gdemonet:
Initial working PR (#844) will be merged as-is, improvements will come in a separate PR (#855).

Changes requested by @nootal:
- [x] Change single-node hostname in `requirements.sh` instead of `eve/main.yaml`
- [x] Rename task for tests execution in single-node stage
- [x] Remove unnecessary double-quotes in `eve/main.yaml`

Changes requested by @slaperche-scality:
- [x] `s/Shell files and templates/Shell scripts/` in `tox.ini`
- [x] Use explicit encoding when writing/reading files (kubeconfig)
- [x] Use more explicit check for single element lists
- [x] Factorize "retry `n` times" pattern

gdemonet:
Closed by #844 and #855

PR #792, opened 2019-03-19
Improvement/717 have test framework available in ci

### LINKED ISSUE #717: Have test framework available in CI
Follows #716 

For Vagrant environment, we already have the source code of the test suite available in the local environment.

Currently, in the CI, when having the OpenStack worker, we only have the ISO as an artifact but we don't have access to the test suite.

Success is defined by being able to launch a dummy test checking we are on the CI.

Suggestion from @tcarmet : reclone in the VM, since it's cached in the CI infrastructure.

charlesprost:
Fixes #717

We want to make sure that the `tox` command and the tests suite are available on OpenStack workers otherwise we can not launch the E2E tests.

Source code was already available in the multiple vms environment.

We may want to freeze the tox version but we don't do it in the linter nor in the README on how to run the tests. So it may be part of a debt ticket if need to?

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -90,6 +90,7 @@ stages:
       flavor: m1.medium
       path: eve/workers/openstack-single-node
     steps:
+      - Git: *git_pull
       - ShellCommand: &retrieve_iso
           name: Retrieve ISO image
           command: >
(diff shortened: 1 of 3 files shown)


charlesprost:
Green build, let's move forward!

NicolasT (approved):

NicolasT:
Not overly fond of code changes which don't add any 'feature' (in this case, running the actual tests) though.

charlesprost:
> Not overly fond of code changes which don't add any 'feature' (in this case, running the actual tests) though.

Yeah indeed, will be followed by #781 but everything could have been reunited in one single issue

charlesprost:
@bert-e approve

PR #709, opened 2019-03-07
Upgrade Helm Version

ssalaues:
Implements the Helm version upgrade requested in #707 

Note: There are also references to the helm version in the `vendor/kubespray` folder and was unsure if it is preferred to updates those as well.

Issue #707, opened 2019-03-05
Upgrade to Helm v2.11 or newer


ssalaues:
Helm v2.11 updates their template functions library `Sprig` to v2.16.0 which added several new useful functions that are helpful in advanced templating such as `slice`. It would be nice if we could upgrade to Helm 2.11 on future release versions of MetalK8s.

I don't mind making the changes and PR but wanted to open the issue first.

NicolasT:
PR welcome :)

ssalaues:
Created the PR and tested locally. I don't have permissions to add reviewers so just wanted to update the status here 😃

NicolasT:
This is now in `development/1.1`, about to land in branches up to `development/1.3`.

NicolasT:
All done.

PR #794, opened 2019-03-19
Upgrade Helm Version

NicolasT:
Fixes: GH-707
See: https://github.com/scality/metalk8s/issues/707
Fixes: GH-709
See: https://github.com/scality/metalk8s/pull/709


NicolasT:
@bert-e approve

eve/images/pep8-checker/Dockerfile
```diff
@@ -5,7 +5,15 @@
...
+RUN apk add --no-cache \
+        gcc \
+        git \
+        musl-dev \
+        python \
+        python2-dev \
+        python3 \
+        py2-pip \
+        && \
     pip install tox && \
```
Zempashi, line 17: Could we fix the version here ? == 1.8.0 seems a good version here (sorry my bad, to not have the correct version pinned here), then checking if py-twisted work or if gcc and musl is needed (and in this case remove py-twisted)

NicolasT: Done

NicolasT:
@Zempashi I tried using the package-provided Twisted, but failed (still insisted on pulling a newer version), so had to keep the `musl-dev` etc. in.

NicolasT:
@bert-e approve

Zempashi (approved):

tcarmet:
@NicolasT Bert-e won't merge this change as we recreated the git tree. You can merge this manually (as you have admin rights)

NicolasT:
> @NicolasT Bert-e won't merge this change as we recreated the git tree. You can merge this manually (as you have admin rights)

What about 1.2/1.3?

tcarmet:
Right... forward port it manually?

tcarmet:
cc @DavidPineauScality

NicolasT:
I'll do it manually to unblock @ssalaues.

ghost:
we have an open issue in our tooling to allow configuring the waterfall in a more fine-grained fashion, so that the forward-porting waterfall would be barred at the major version switch for 2.0
Not scheduled at the moment, though.

PR #842, opened 2019-03-25
build: fix `stat` command on MacOS

slaperche-scality:
This should fix the issue on MacOS.

How to test:
- If you run `./doit.sh list` once you'll see that it installs a virtualenv first, then you see the output of `doit`.
- If you run `./doit.sh list` you should only see the output of `doit` (the venv is reused)
- If you `touch buildchain/requirements-Darwin.txt` (to fake an update of the dependencies) and then  `./doit.sh list` you should see like the first time (venv is reinstalled and then `doit` output).


MonPote:
It works like a charm :)

MonPote (approved):

slaperche-scality:
@bert-e approve

slaperche-scality:
@bert-e reset

buildchain/requirements-Darwin.txt
```diff
@@ -0,0 +1,9 @@
+#
```
NicolasT, line 1: Ugh :-( How is one with a Linux machine supposed to update this file now?

slaperche-scality: Yup, Mac people are responsible to update it if they want it to work for them (I asked Carlito to generate this file) :S

I'm not that happy with the current solution but I didn't find a better way with `pip-compile` :-(
What I explored was:
- let people run `pip-compile` themselves: not good, they have to install `pip-tool` (an extra-dep, and extra-step) and we lose the reproducible build feature.
- using environment markers (as supported by pip): can't make it work with `pip-compile` though (it evaluate them at "compile" time instead of copying them to the generated `requirements.txt` so "compiling" on Linux force the use of `inotify` => issue is not solved).

PR #844, opened 2019-03-25
Run tests in the CI

gdemonet:
# What

This PR introduces a run of the post-install test suite on both the single-node and multiple-nodes deployment scenarios in the CI. It is now a blocking condition in the pre-merge build.
Most of the modifications introduced were only made to cope with the CI context: long hostnames created issues with kubelet registration (see cdb5343d075d93124d9af2d76e710bb262edeea3), and the "bastion" VM used to orchestrate the multi-nodes deployment, **and tests**, could not be a part of the control-plane network (which it created). Review through the commit messages to find more details.

# How to test

While a green build is the main goal of these changes (and took some time to achieve), you should still be able to run the test suite locally as before.

If you want to reproduce the single-node tests conditions:

```
# On your host
./doit.sh vagrantup
vagrant ssh
# On your bootstrap node
sudo yum install -y epel-release
sudo yum install -y python34-pip
sudo pip3 install tox
sudo yum install -y --enablerepo=metalk8s-kubernetes kubectl
cd /vagrant
tox -e tests-local
```

The multi-nodes tests are the same as a simple `tox -e tests` on your host.

---

Fixes: GH-781.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -201,6 +201,11 @@ stages:
             # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
             TF_VAR_nodes_count: "2"
           haltOnFailure: true
+      - SetPropertyFromCommand:
+          name: Save spawned bootstrap IP
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          property: bootstrap_ip
+          command: terraform output bootstrap_ip
       - ShellCommand:
           name: Configure SSH config for bootstrap node
           command: >
@@ -255,6 +260,25 @@ stages:
             sudo bash
             /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
           workdir: "build/eve/workers/openstack-multiple-nodes/terraform/"
+          haltOnFailure: true
+      - ShellCommand:
+          name: Install kubectl on the boostrap node
+          command: >
+            ssh -F bootstrap_ssh_config bootstrap
+            sudo yum install -y kubectl --disablerepo=*
+            --enablerepo=metalk8s-kubernetes
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+      - ShellCommand:
+          name: Run tests on the bootstrap node
+          env:
+            SSH_CONFIG_FILE: >-
+              eve/workers/openstack-multiple-nodes/terraform/bootstrap_ssh_config
+            ISO_MOUNTPOINT: >-
+              /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+            TEST_HOSTS_LIST: bootstrap
+          command: >
+            tox -e tests --
+            --bootstrap-ip=%(prop:bootstrap_ip)s --skip-tls-verify
           haltOnFailure: true
       - ShellCommand:
           name: Destroy openstack virtual infra
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -145,6 +145,18 @@ stages:
             /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
           haltOnFailure: true
       - ShellCommand:
+          name: Install kubectl for running tests
+          command: >
+            sudo yum install -y kubectl --disablerepo=*
+            --enablerepo=metalk8s-kubernetes
+      - ShellCommand:
+          name: Run tests on the bootstrap node
+          env:
+            ISO_MOUNTPOINT: >
+              /srv/scality/metalk8s-%(prop:metalk8s_short_version)s
+          command: tox -e tests-local
+          haltOnFailure: true
+      - ShellCommand:
           name: Debug step - report IPs, add SSH keys, wait 1 hour
           timeout: 3600
           command: >
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -134,6 +134,11 @@ stages:
             END
             EOF
       - ShellCommand:
+          # This step is needed so Kubelet registration doesn't fail for DNS
+          # entries longer than 63 characters
+          name: Change hostname to a shorter value
+          command: sudo hostnamectl set-hostname bootstrap
+      - ShellCommand:
           name: Start the bootstrap process
           command: >
             sudo bash
@@ -180,7 +185,8 @@ stages:
             OS_USERNAME: "%(secret:scality_cloud_username)s"
             OS_PASSWORD: "%(secret:scality_cloud_password)s"
             OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
-            TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+            # FIXME: this makes hostnames too long
+            # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
             TF_VAR_nodes_count: "2"
           haltOnFailure: true
       - ShellCommand:
--- a/tests/post/steps/conftest.py
+++ b/tests/post/steps/conftest.py
@@ -1,7 +1,4 @@
 # -*- coding: utf-8 -*-
-from configparser import ConfigParser
-from itertools import chain
-
 import pytest
 from pytest_bdd import (
     given,
@@ -9,24 +6,6 @@ from pytest_bdd import (
     when,
     parsers,
 )
-
-
-# Pytest fixtures
-@pytest.fixture(scope="session")
-def version():
-    """Fixture that read metalk8s version.
-
-    Version is extracted from _build/root/product.txt
-    """
-    version_file = "_build/root/product.txt"
-    parser = ConfigParser()
-
-    with open(version_file) as stream:
-        # add a section header for configparser
-        stream = chain(("[top]",), stream)
-        parser.read_file(stream)
-
-    return dict(parser.items('top'))
 
 
 def _verify_kubeapi_service(host):
@@ -37,10 +16,11 @@ def _verify_kubeapi_service(host):
         assert retcode == 0
 
 
-def _run_bootstrap(host, version):
-    cmd = "/srv/scality/metalk8s-{0}/bootstrap.sh".format(
-        version['short_version']
-    )
+def _run_bootstrap(request, host):
+    # FIXME: this can only run on the bootstrap node, we'd need to skip such
+    #        test if the host fixture is not adapted
+    iso_root = request.config.getoption("--iso-root")
+    cmd = str(iso_root / "bootstrap.sh")
     with host.sudo():
         res = host.run(cmd)
         assert res.rc == 0, res.stdout
@@ -50,8 +30,8 @@ def _run_bootstrap(host, version):
 
 # Given
 @given('bootstrap was run once')
-def run_bootstrap(host, version):
-    _run_bootstrap(host, version)
+def run_bootstrap(request, host):
+    _run_bootstrap(request, host)
 
 
 @given("the Kubernetes API is available")
@@ -61,8 +41,8 @@ def check_service(host):
 
 # When
 @when('we run bootstrap a second time')
-def rerun_bootstrap(host, version):
-    _run_bootstrap(host, version)
+def rerun_bootstrap(request, host):
+    _run_bootstrap(request, host)
 
 
 # Then
--- a/tests/post/steps/conftest.py
+++ b/tests/post/steps/conftest.py
@@ -6,6 +6,44 @@ from pytest_bdd import (
     when,
     parsers,
 )
+import yaml
+
+
+# Pytest fixtures
+@pytest.fixture(scope="module")
+def kubeconfig_data(request, host):
+    """Fixture to generate a kubeconfig file for remote usage."""
+    with host.sudo():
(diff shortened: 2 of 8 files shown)


tests/post/steps/conftest.py
```diff
@@ -37,10 +54,11 @@ def _verify_kubeapi_service(host):
         assert retcode == 0
 
 
-def _run_bootstrap(host, version):
-    cmd = "/srv/scality/metalk8s-{0}/bootstrap.sh".format(
-        version['short_version']
-    )
+def _run_bootstrap(request, host):
+    # FIXME: this can only run on the bootstrap node, we'd need to skip such
+    #        test if the host fixture is not adapted
```
sayf-eddine-scality, line 59: or [limit the tests](https://testinfra.readthedocs.io/en/latest/invocation.html) on the bootstrap by adding 
`testinfra_hosts = ['bootstrap']`

tests/post/steps/test_liveness.py
```diff
@@ -34,58 +35,76 @@ def check_resource_list(host, resource, namespace):
...
+            f=len(pods),
+            l=label,
+            t=str(interval * attempts)
+        )
+
+
+# Utilities
+def _get_pods(host, label, namespace='default', status_phase='Running'):
+    cmd = (
+        'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
```
sayf-eddine-scality, line 101: can we use the kubeconfig fixture here too ?

eve/main.yml
```diff
@@ -133,12 +133,29 @@ stages:
               workloadPlane: 10.0.0.0/8
             END
             EOF
+      - ShellCommand:
+          # This step is needed so Kubelet registration doesn't fail for DNS
+          # entries longer than 63 characters
+          name: Change hostname to a shorter value
```
charlesprost, line 139: minor: that's something you may want to put into the `eve/workers/openstack-single-node/requirements.sh` file instead

eve/main.yml
```diff
@@ -133,12 +133,29 @@ stages:
...
             sudo bash
             /srv/scality/metalk8s-%(prop:metalk8s_short_version)s/bootstrap.sh
           haltOnFailure: true
+      - ShellCommand:
+          name: Install kubectl for running tests
+          command: >
+            sudo yum install -y kubectl --disablerepo=*
+            --enablerepo=metalk8s-kubernetes
+      - ShellCommand:
+          name: Run tests on the bootstrap node
```
charlesprost, line 153: minor: remove `on the bootstrap node` since here we are on a worker that works a bootstrap node. That may be misleading.

eve/main.yml
```diff
@@ -180,9 +197,15 @@ stages:
...
             OS_PASSWORD: "%(secret:scality_cloud_password)s"
             OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
-            TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+            # FIXME: this makes hostnames too long
+            # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
             TF_VAR_nodes_count: "2"
           haltOnFailure: true
+      - SetPropertyFromCommand:
+          name: Save spawned bootstrap IP
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
```
charlesprost, line 206: minor: in this file we use `"` around this path, so you may want to remove them since we obviously don't need them

charlesprost (commented):
Really good work.

tox.ini
```diff
@@ -17,19 +20,27 @@ commands =
     mypy --strict buildchain/dodo.py buildchain/buildchain
 
 [testenv:lint-shell]
-whitelist_externals += shellcheck
+description =
+    Lint Shell files and templates using shellcheck.
```
slaperche-scality, line 24: You can simply mention `Shell script`, because template can be confusing.

tests/post/steps/conftest.py
```diff
@@ -6,6 +6,44 @@
...
+    if request.config.getoption("--skip-tls-verify"):
+        kube_cluster_section["insecure-skip-tls-verify"] = True
+
+    return data
+
+
+@pytest.fixture
+def kubeconfig(kubeconfig_data, tmp_path):
+    kubeconfig_path = tmp_path / "admin.conf"
+    kubeconfig_path.write_text(yaml.dump(kubeconfig_data))
```
slaperche-scality, line 45: `kubeconfig_path.write_text(yaml.dump(kubeconfig_data), encoding='utf-8')` (always specify an encoding, as the default one can change)

tests/post/steps/conftest.py
```diff
@@ -6,6 +6,44 @@
...
+    return str(kubeconfig_path)  # Need Python 3.6 to open() a Path object
```
slaperche-scality, line 46: Note that you could also `kubeconfig_path.open`, but converting to `str` probably require less code changes.

tests/post/steps/conftest.py
```diff
@@ -6,6 +6,44 @@
...
 )
+import yaml
+
+
+# Pytest fixtures
+@pytest.fixture(scope="module")
+def kubeconfig_data(request, host):
+    """Fixture to generate a kubeconfig file for remote usage."""
+    with host.sudo():
+        kubeconfig_file = host.file('/etc/kubernetes/admin.conf')
```
slaperche-scality, line 17: dunno where you can specify an encoding, but if possible specify one (since relying on the default one may bite us in the future).

tests/post/steps/test_liveness.py
```diff
@@ -34,58 +35,76 @@ def check_resource_list(host, resource, namespace):
...
 @then(parsers.parse(
-    "we can exec '{command}' in the "
-    "'{pod}' pod in the '{namespace}' namespace"))
-def check_exec(host, command, pod, namespace):
+    "we can exec '{command}' in the pod labeled '{label}' "
+    "in the '{namespace}' namespace"))
+def check_exec(host, command, label, namespace):
+    candidates = _get_pods(host, label, namespace)
+    try:
+        pod, = candidates
```
slaperche-scality, line 43: I'm still not fond of the trailing comma trick since it's hard to spot.

Could be rewritten:

```
    if len(pos) != 1:
        pytest.fail(
            "Expected one (and only one) pod with label {l}, found {f}".format(
                l=label, f=len(candidates)
            )
        )
    pod = candidates[0]
```
The code is even shorter by one line while being more explicit.

tests/post/steps/test_liveness.py
```diff
@@ -34,58 +35,76 @@ def check_resource_list(host, resource, namespace):
...
     # so implement a timeout
     interval = 3
     attempts = 10
 
-    with host.sudo():
-        for _ in range(attempts):
-
-            ret = host.run(cmd)
-            assert ret.rc == 0, ret.stdout
+    for _ in range(attempts):
```
slaperche-scality, line 72: Is this kind of pattern (retry an action `N` times, waiting `t` second between each retry until a condition `pred` is true) is repeated?

If yes, it could be worth having a class/function to abstract that (can be done in another PR/debt ticket).

Even without repetition, the abstraction could be worth it, to avoid "polluting" the test function with unrelated logic.

tests/post/steps/test_liveness.py
```diff
@@ -34,58 +35,76 @@ def check_resource_list(host, resource, namespace):
...
+        '--namespace {namespace} -o json'
+    ).format(label=label, namespace=namespace, status_phase=status_phase)
 
-            if len(pods['items']) < min_pods_count:
-                time.sleep(interval)
-            else:
-                # if nb pods is >= then it's ok
-                break
+    with host.sudo():
+        result = host.run(cmd)
```
slaperche-scality, line 107: How does that `run` method works?

If it's like `Popen` then it's better to pass a list of string instead of a string (since you would need a shell, and thus escape the parameters).

slaperche-scality (dismissed):

tests/post/steps/conftest.py
gdemonet: Good idea, thanks. I created #848 to track this.

tests/post/steps/test_liveness.py
gdemonet: I'd rather remove the use of `kubectl` on the hosts and use the Python client with the `kubeconfig` fixture, created #850.

eve/main.yml
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

eve/main.yml
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

eve/main.yml
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tox.ini
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tests/post/steps/conftest.py
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tests/post/steps/conftest.py
gdemonet: Yes I found it to be the easiest change, will consider the `Path.open()` method in the improvements PR (it might be good to rely on `pathlib` for everything instead of having some of both approaches).

tests/post/steps/conftest.py
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tests/post/steps/test_liveness.py
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tests/post/steps/test_liveness.py
gdemonet: Tracked [here](https://github.com/scality/metalk8s/issues/781#issuecomment-476565028).

tests/post/steps/conftest.py
slaperche-scality: Yeah, it's clearly better to pick one ans stick to it as much as possible instead of mixing both (but that's clearly out of scope for this PR).

tests/post/steps/test_liveness.py
gdemonet: It's not accepting a list of strings, but it handles quoting the parameters to guard against shell injection, if that's your concern.

The method is documented [here](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.host.Host.run).
It seems to first [`quote` the string](https://github.com/philpep/testinfra/blob/bbab44504eec8d5c8995e3473166f6e57f9d88f9/testinfra/backend/base.py#L151-L155), which is used to format the `command` with `pipes.quote`d positional arguments.

Let's keep it like this for now and see if we encounter issues with it in the future.

gdemonet:
@bert-e approve

tests/post/steps/test_liveness.py
slaperche-scality: IIUC, in our case it's not quoted as we pass directly a whole string (but the method does escape params if called properly).

tests/post/steps/conftest.py
gdemonet: I just remembered I need a string anyway since `kubernetes.load_kube_config(config_file=path)` will do `open(path)` internally (we'll need Python 3.6 or higher to open a `Path` object).

tests/post/steps/test_liveness.py
gdemonet: Right, I'll improve it in the following PR

gdemonet:
@bert-e reset

charlesprost (dismissed):

charlesprost:
> # Build failed
> 
> The build for [commit](https://github.com/scality/metalk8s/commit/13b0060ba7d5cf6e54b9eb5296530727f955e42d) did not succeed in branch w/2.2/feature/781-run-tests-in-ci.
> 
> _The following options are set:_ **approve**

Rebuilding

gdemonet:
Fails in both 2.1 and 2.2 versions, need to investigate and fix.

gdemonet:
@bert-e reset

gdemonet:
@bert-e force_reset

gdemonet:
@bert-e reset

eve/main.yml
```diff
@@ -133,12 +133,29 @@ stages:
               workloadPlane: 10.0.0.0/8
             END
             EOF
+      - ShellCommand:
```
charlesprost, line 136: Thought this one was in the requirements.sh no?

gdemonet: In the follow-up PR yes (#855)

charlesprost (approved):

PR #847, opened 2019-03-25
Improvement/automatically deploy preshared ssh key

### LINKED ISSUE #846: Deploy Vagrant SSH preshared key into the Vagrant environment
Following #739 

We automatically generate a new SSH keypair that can be used to quickly extend the cluster but it's not deployed automatically into the different hosts from the Vagrant env. Let's fix this.

charlesprost:
Fixes #846

We already have a step in `doit` that pre-provisions a SSH key pair in the `.vagrant` folder.

Now we want to deploy automatically the private part in the bootstrap node and the public part in all other nodes so we can use this key pair to extend the cluster through salt-ssh.

How to test it?

Just run `./doit vagrantup` then check the key is present in `/etc/metalk8s/pki`, then `vagrant up node1` and check the key is present in the `.ssh/authorized_keys` file.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -130,6 +130,16 @@ echo "Launching bootstrap"
 exec "/srv/scality/metalk8s-$SHORT_VERSION/bootstrap.sh"
 SCRIPT
 
+DEPLOY_SSH_PUBLIC_KEY = <<-SCRIPT
+#!/bin/bash
+
+set -eu -o pipefail
+
+if ! grep -Fxq "$(cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub)" .ssh/authorized_keys ; then
+   cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub >> .ssh/authorized_keys
+fi
+SCRIPT
+
 # To support VirtualBox linked clones
 Vagrant.require_version(">= 1.8")
 
@@ -181,6 +191,15 @@ Vagrant.configure("2") do |config|
 
       # No need for Guest Additions since there is no synced folder
       node.vbguest.auto_update = false
+
+      node.vm.provision "copy-ssh-public-key",
+        type: "file",
+        source: ".vagrant/#{PRESHARED_SSH_KEY_NAME}.pub",
+        destination: ".ssh/#{PRESHARED_SSH_KEY_NAME}.pub"
+
+      node.vm.provision "add-ssh-public-key-to-authorized-keys",
+        type: "shell",
+        inline: DEPLOY_SSH_PUBLIC_KEY
     end
   end
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -90,6 +90,20 @@ echo "Swap disabled"
 exit $RC
 SCRIPT
 
+PRESHARED_SSH_KEY_NAME = 'preshared_key_for_k8s_nodes'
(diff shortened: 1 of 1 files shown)


slaperche-scality (dismissed):

Vagrantfile
```diff
@@ -90,6 +90,20 @@ echo "Swap disabled"
...
+PRESHARED_SSH_KEY_NAME = 'preshared_key_for_k8s_nodes'
+IMPORT_SSH_PRIVATE_KEY = <<-SCRIPT
+#!/bin/bash
+
+set -eu -o pipefail
+
+echo "Deploying preshared SSH private key on the host"
+
+mkdir -p /etc/metalk8s/pki/
+cp /vagrant/.vagrant/#{PRESHARED_SSH_KEY_NAME} /etc/metalk8s/pki/
```
NicolasT, line 102: `chmod 0600`?

Vagrantfile
```diff
@@ -117,6 +130,16 @@ echo "Launching bootstrap"
 exec "/srv/scality/metalk8s-$SHORT_VERSION/bootstrap.sh"
 SCRIPT
 
+DEPLOY_SSH_PUBLIC_KEY = <<-SCRIPT
+#!/bin/bash
+
+set -eu -o pipefail
+
+if ! grep -Fxq "$(cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub)" .ssh/authorized_keys ; then
+   cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub >> .ssh/authorized_keys
```
NicolasT, line 139: Instead of copying the file into the VM, could as well read it in Ruby, then

```sh
cat >> .ssh/authorized_keys << EOF
#{SSH_PUBLIC_KEY}
EOF
```

(and do the same for the private key).

NicolasT (dismissed):

Vagrantfile
slaperche-scality: +1

Can be read with `File.read(path, :encoding => 'utf-8'))` or, if we are using Ruby >= 1.9: `File.read(path, encoding: 'utf-8')`

NicolasT: @nootal I think the benefit of that approach is that the `Vagrantfile` script could error out, before starting to deploy something, which could be useful if someone runs `vagrant up` (not `doit vagrantup`), and the key hasn't been created before.

charlesprost:
@NicolasT @slaperche-scality could you please check the two fixup commits I've pushed on top of the branch? That should do the trick. Then I'll squash/rebase.

slaperche-scality:
I don't know if it's what Nicolas had in mind or if it can be improved, but as it is I find the previous version more readable ^^"

charlesprost:
@slaperche-scality OK then I'll roll back, add a debt ticket with the diff so we can start again from it, and merge.

slaperche-scality (dismissed):

charlesprost:
@bert-e reset

charlesprost:
@slaperche-scality can you approve again now that I have removed the fixup commits?

Debt ticket is #849

charlesprost:
@bert-e approve

sayf-eddine-scality (dismissed):

alexandre-allard (dismissed):

charlesprost:
@bert-e reset

sayf-eddine-scality (approved):

PR #855, opened 2019-03-26
Follow-up PR: requested changes for CI tests

gdemonet:
This PR only includes small modifications suggested/requested by reviewers in #844 
See each commit independently for a quick review!

---

Relates to GH-781

### DIFF
--- a/eve/workers/openstack-single-node/requirements.sh
+++ b/eve/workers/openstack-single-node/requirements.sh
@@ -13,3 +13,7 @@ yum install -y "${PACKAGES[@]}"
 yum clean all
 
 pip3 install tox
+
+# This step is needed so Kubelet registration doesn't fail for DNS entries
+# longer than 63 characters.
+hostnamectl set-hostname bootstrap
--- a/tests/kube_utils.py
+++ b/tests/kube_utils.py
@@ -1,0 +1,16 @@
+import json
+
+
+def get_pods(host, label, namespace='default', status_phase='Running'):
+    with host.sudo():
+        result = host.run(
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'get pods -l "%s" --field-selector=status.phase=%s '
+            '--namespace %s -o json',
+            label,
+            status_phase,
+            namespace,
+        )
+        assert result.rc == 0, result.stdout
+
+        return json.loads(result.stdout)['items']
--- a/tests/post/steps/test_liveness.py
+++ b/tests/post/steps/test_liveness.py
@@ -4,6 +4,7 @@ import time
 import pytest
 from pytest_bdd import scenario, then, parsers
 
+from tests import kube_utils
 from tests import utils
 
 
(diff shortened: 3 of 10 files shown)


eve/workers/openstack-single-node/requirements.sh
```diff
@@ -13,3 +13,7 @@ yum install -y "${PACKAGES[@]}"
 yum clean all
 
 pip3 install tox
+
+# This step is needed so Kubelet registration doesn't fail for DNS entries
+# longer than 63 characters
```
charlesprost: Missing trailing endpoint.

tests/kube_utils.py
```diff
@@ -0,0 +1,16 @@
...
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'get pods -l "%s" --field-selector=status.phase=%s '
+            '--namespace %s -o json',
+            label,
+            status_phase,
+            namespace,
+        )
+        assert result.rc == 0, result.stdout
+
+        return json.loads(result.stdout)['items']
```
charlesprost: Missing trailing newline.

tests/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    wait=1,
+    error_msg=None,
+    logger='tests.utils.retry',
+):
+    log = logging.getLogger(logger)
+
+    for idx in range(times):
+        try:
+            operation()
+            log.info("Attempt %d/%d succeeded", idx, times)
```
charlesprost: Use an `else` statement to differentiate what needs to be part of the `try` statement and what does not.

charlesprost (commented):
A few minor changes and you're good to merge.

tests/post/steps/test_liveness.py
```diff
@@ -27,84 +30,54 @@ def test_expected_pods(host):
...
     "we can exec '{command}' in the pod labeled '{label}' "
     "in the '{namespace}' namespace"))
 def check_exec(host, command, label, namespace):
-    candidates = _get_pods(host, label, namespace)
-    try:
-        pod, = candidates
-    except ValueError:
+    candidates = kube_utils.get_pods(host, label, namespace)
+
+    if len(candidates) != 1:
```
NicolasT: Instead of `if`, `pytest.fail`, `return`, why not

```python
assert len(candidates) == 1, "Expected one (..."
```

tests/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def retry(
+    operation,
+    times=1,
+    wait=1,
+    error_msg=None,
+    logger='tests.utils.retry',
+):
+    log = logging.getLogger(logger)
```
NicolasT: Create this as a singleton at module scope.

tests/post/steps/test_liveness.py
gdemonet: Indeed, I just converted the `try/except` to this but an `assert` is more idiomatic for `pytest`

tests/utils.py
gdemonet: The point was that callers of this `retry` function could provide their own logger name (could be useful if we start having retries within other retries).

gdemonet: But I'll use a singleton as you requested and pass some string to put in logs instead.

gdemonet:
@bert-e reset

gdemonet:
@bert-e approve

slaperche-scality (dismissed):

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

alexandre-allard (approved):

Issue #757, opened 2019-03-12
Remove address lookup duplication


NicolasT:
Currently, we spread things like

```
{% from "metalk8s/map.jinja" import networks with context %}
{% set ip_candidates = salt.network.ip_addrs(cidr=networks.control_plane) %}
```

around in the SLS files, which ain't nice.

We should calculate this value somehow once, and then be able to reference it from a single location.

Similarly, we should be able to look up such addresses of *other* nodes, e.g. to construct `etcd` peer lists.

How this should be done is TBD.

TeddyAndrieux:
The method to get the Ip is already done by #833 (custom modules `metalk8s.get_control_plane_ips` and `metalk8s.get_workload_plane_ips`) 
Just thinking that could be a custom grains which is better 

Edit: Oh indeed we have an issue for the moment as we need pillar information to get the Ip so we can't put it in a custom grains

NicolasT:
Can minions request grains from other nodes?

TeddyAndrieux:
Not directly but you can have grains value in the mine
```
mine_functions:
  toto:
    - mine_function: grains.get
    - toto
```

NicolasT:
#970

PR #869, opened 2019-03-27
etcd cluster expansion

slaperche-scality:
How to test:

See procedure [here](https://docs.google.com/document/d/1GAMbSTIFhCypBUITg_SunCOeoMdbEfBK_XZZJIw3Wdw/edit).

### DIFF
--- a/salt/_modules/cri.py
+++ b/salt/_modules/cri.py
@@ -148,3 +148,21 @@ def wait_container(name, state, timeout=60, delay=5):
     else:
         log.error('Failed to find container "%s" in state "%s"', name, state)
         return False
+
+
+def component_is_running(name):
+    '''Return true if the specified component is running.
+
+    .. note::
+
+       This uses the :command:`crictl` command, which should be configured
+       correctly on the system, e.g. in :file:`/etc/crictl.yaml`.
+    '''
+    log.info('Checking if compopent %s is running', name)
+    out = __salt__['cmd.run_all'](
+        'crictl pods --label component={} --state=ready -o json'.format(name)
+    )
+    if out['retcode'] != 0:
+        log.error('Failed to list pods')
+        return False
+    return len(salt.utils.json.loads(out['stdout'])['items']) != 0
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -39,3 +39,17 @@ def get_workload_plane_ips():
         return
 
     return __salt__['network.ip_addrs'](cidr=cidr)
+
+
+def get_etcd_endpoint():
+    '''Get endpoint only if it's an etcd node.'''
+    DEFAULT_ETCD_PORT = 2380
+    cidr = __pillar__.get('networks', {}).get('control_plane')
+    if not cidr:
+        return
+    if __salt__['cri.component_is_running'](name='etcd'):
+        ips = __salt__['network.ip_addrs'](cidr=cidr)
+        hostname = __salt__['network.get_hostname']()
+        return '{host}=https://{ip}:{port}'.format(
+            host=hostname, ip=ips[0], port=DEFAULT_ETCD_PORT
+        )
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -4,6 +4,17 @@ Module for handling MetalK8s specific calls.
 '''
 import logging
 
+from salt.exceptions import CommandExecutionError
+
+try:
+    import kubernetes.client
+    import kubernetes.config
+    from kubernetes.client.rest import ApiException
+    from kubernetes.stream import stream as k8s_stream
+    HAS_DEPS = True
+except ImportError:
+    HAS_DEPS = False
+
 
 log = logging.getLogger(__name__)
 
@@ -14,7 +25,11 @@ def __virtual__():
     '''
     Check dependencies
     '''
-    return __virtualname__
+    if HAS_DEPS:
+        return __virtualname__
+    else:
+        return False, 'Unable to load the MetalK8s module: '\
+            'Kubernetes Python client library is missing.'
 
 
 def get_control_plane_ips():
@@ -53,3 +68,44 @@ def get_etcd_endpoint():
         return '{host}=https://{ip}:{port}'.format(
             host=hostname, ip=ips[0], port=DEFAULT_ETCD_PORT
         )
+
+
+def add_etcd_node(host, endpoint):
+    '''Add a new `etcd` node into the `etcd` cluster.
+
+    This module is only runnable from the salt-master on the bootstrap node.
+
+    Arguments:
+        host (str): hostname of the new etcd node
+        endpoint (str): endpoint of the new etcd node
+    '''
+    pod_name = 'etcd-{}'.format(__salt__['network.get_hostname']())
+    try:
+        client = kubernetes.config.new_client_from_config(
+            config_file='/etc/kubernetes/admin.conf'
+        )
+    except:
+        log.exception('failed to load kubeconfig')
+        raise
+    api = kubernetes.client.CoreV1Api(api_client=client)
+    etcd_command = [
+        'etcdctl',
+        '--endpoints', 'https://localhost:2379',
+        '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
+        '--key-file', '/etc/kubernetes/pki/etcd/server.key',
+        '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
+        'member', 'add', host, endpoint
+    ]
+    try:
+        err = k8s_stream(
+            api.connect_get_namespaced_pod_exec,
+            name=pod_name, namespace='kube-system',
+            command=etcd_command,
+            stderr=True, stdin=False, stdout=False, tty=False
+        )
+        if err:
+            log.error('etcdctl error: %s', err)
+            raise CommandExecutionError('etcdctl: {}'.format(err))
+    except ApiException as exn:
+        log.exception('failed to run etcdctl')
+        raise exn
--- a/salt/metalk8s/kubeadm/init/etcd/local.sls
+++ b/salt/metalk8s/kubeadm/init/etcd/local.sls
@@ -8,6 +8,9 @@
 {% if ip_candidates %}
 {% set host = ip_candidates[0] %}
 
+{%- set ca_cert = salt['mine.get']('*', 'kubernetes_etcd_ca_b64').values()[0] %}
+{%- set ca_cert_b64 = salt['hashutil.base64_b64decode'](ca_cert) %}
+
 Create etcd database directory:
   file.directory:
     - name: /var/lib/etcd
@@ -15,6 +18,16 @@ Create etcd database directory:
     - user: root
     - group: root
     - makedirs: True
+
+Ensure etcd CA certificate is present:
+  file.managed:
+    - name: /etc/kubernetes/pki/etcd/ca.crt
+    - user: root
+    - group : root
+    - mode: 644
+    - makedirs: True
+    - dir_mode: 755
+    - contents: {{ ca_cert_b64.split('\n') }}
 
 Create local etcd Pod manifest:
   file.managed:
@@ -55,6 +68,7 @@ Create local etcd Pod manifest:
             readOnly: true
     - require:
       - file: Create etcd database directory
+      - file: Ensure etcd CA certificate is present
 
 
 
--- a/salt/metalk8s/kubeadm/init/etcd/local.sls
+++ b/salt/metalk8s/kubeadm/init/etcd/local.sls
@@ -82,7 +82,13 @@ Create local etcd Pod manifest:
       - file: Create etcd database directory
       - file: Ensure etcd CA certificate is present
 
-
+Advertise etcd node in the mine:
+  module.run:
+    - mine.send:
+      - func: 'etcd_endpoints'
+      - mine_function: metalk8s.get_etcd_endpoint
+    - watch:
+      - file: Create local etcd Pod manifest
 
 {% else %}
 No available advertise IP for etcd:
--- a/salt/metalk8s/kubeadm/init/etcd/local.sls
+++ b/salt/metalk8s/kubeadm/init/etcd/local.sls
@@ -8,8 +8,19 @@
 {% if ip_candidates %}
 {% set host = ip_candidates[0] %}
 
+{% set endpoint  = host_name ~ '=https://' ~ host ~ ':2380' %}
+
 {%- set ca_cert = salt['mine.get']('*', 'kubernetes_etcd_ca_b64').values()[0] %}
 {%- set ca_cert_b64 = salt['hashutil.base64_b64decode'](ca_cert) %}
+
+{#- Get the list of existing etcd node. #}
+{%- set etcd_endpoints = salt['mine.get']('*', 'etcd_endpoints').values() %}
+
+{#- Compute the initial state according to the existing list of node. #}
+{%- set state = "existing" if etcd_endpoints else "new" -%}
+
+{#- Add ourselves to the list #}
+{%- do etcd_endpoints.append(endpoint) %}
 
 Create etcd database directory:
   file.directory:
@@ -49,7 +60,8 @@ Create local etcd Pod manifest:
           - --client-cert-auth=true
           - --data-dir=/var/lib/etcd
           - --initial-advertise-peer-urls=https://{{ host }}:2380
-          - --initial-cluster={{ host_name }}=https://{{ host }}:2380
+          - --initial-cluster={{ etcd_endpoints|unique|join(',') }}
+          - --initial-cluster-state={{ state }}
           - --key-file=/etc/kubernetes/pki/etcd/server.key
           - --listen-client-urls=https://127.0.0.1:2379,https://{{ host }}:2379
           - --listen-peer-urls=https://{{ host }}:2380
--- a/salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
@@ -25,3 +25,19 @@ Refresh the list of etcd nodes:
   salt.function:
     - name: mine.update
     - tgt: '*'
+
+Deploy the new etcd node:
+  salt.state:
+    - tgt: {{ pillar['node_name'] }}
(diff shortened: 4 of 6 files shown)


salt/_modules/cri.py
```diff
@@ -148,3 +148,22 @@ def wait_container(name, state, timeout=60, delay=5):
...
+def component_is_running(name):
+    '''Return true the specified component is running.
+
+    .. note::
+
+       This uses the :command:`crictl` command, which should be configured
+       correctly on the system, e.g. in :file:`/etc/crictl.yaml`.
+    '''
+    log.info('Checking if compopent %s is running', name)
+    out = __salt__['cmd.run_all']('crictl pods -o json')
```
sayf-eddine-scality: wouldnt `crictl ps` be more exact as the pod may exist but it is in a crash loop

salt/_modules/cri.py
```diff
@@ -148,3 +148,22 @@ def wait_container(name, state, timeout=60, delay=5):
     else:
         log.error('Failed to find container "%s" in state "%s"', name, state)
         return False
+
+
+def component_is_running(name):
+    '''Return true the specified component is running.
```
jbertran: +if

salt/_modules/cri.py
```diff
@@ -148,3 +148,22 @@ def wait_container(name, state, timeout=60, delay=5):
...
+       This uses the :command:`crictl` command, which should be configured
+       correctly on the system, e.g. in :file:`/etc/crictl.yaml`.
+    '''
+    log.info('Checking if compopent %s is running', name)
+    out = __salt__['cmd.run_all']('crictl pods -o json')
+    if out['retcode'] != 0:
+        log.error('Failed to list pods')
+        return False
+    for pod in salt.utils.json.loads(out['stdout'])['items']:
+        if pod['labels'].get('component') == name:
```
NicolasT: Just do the filtering on the CLI? `crictl pods --label component=etcd --state=ready -o json`

salt/_modules/metalk8s.py
```diff
@@ -0,0 +1,32 @@
...
+
+def get_etcd_endpoints():
+    '''Get endpoints only if it's an etcd node.'''
+    DEFAULT_ETCD_PORT = 2380
+    cidr = __pillar__.get('networks', {}).get('control_plane')
+    if not cidr:
+        return
+    if __salt__['cri.component_is_running'](name='etcd'):
+        ips = __salt__['network.ip_addrs'](cidr=cidr)
+        if isinstance(ips, list):
```
NicolasT: What else can it be?

salt/_modules/metalk8s.py
```diff
@@ -0,0 +1,32 @@
...
+            return [
+                'https://{ip}:{port}'.format(ip, DEFAULT_ETCD_PORT)
+                for ip in ips
```
NicolasT: Nope, we bind `etcd` to only one of them.

Overall, given #757, I'd like us to have all servers to pick *one* control-plane IP and *one* workload-plane IP, then publish the chosen IPs in the mine, and bind services to only this IP, after which any client of such service(s) can retrieve its address easily, and have it uniquely identified.

Furthermore, over time, it'd allow us to recreate client configuration when relevant values in the mine change by using a reactor (e.g. when a server IP is changed).

All of the above is, by the way, not restricted to `etcd` ;-)

NicolasT (dismissed):

salt/_modules/cri.py
slaperche-scality: Ho, neat!

salt/_modules/metalk8s.py
slaperche-scality: Indeed, that seems useless (I copied that straight from a draft of Teddy's branch).

salt/_modules/metalk8s.py
slaperche-scality: Ok, then I'll always pick the first one for now.

salt/_modules/cri.py
```diff
@@ -148,3 +148,21 @@ def wait_container(name, state, timeout=60, delay=5):
...
+    '''Return true if the specified component is running.
+
+    .. note::
+
+       This uses the :command:`crictl` command, which should be configured
+       correctly on the system, e.g. in :file:`/etc/crictl.yaml`.
+    '''
+    log.info('Checking if compopent %s is running', name)
+    out = __salt__['cmd.run_all'](
+        'crictl pods --label component=etcd --state=ready -o json'
```
NicolasT: Shouldn't this use `name`?

salt/metalk8s/kubeadm/init/etcd/local.sls
```diff
@@ -7,6 +7,24 @@
 {% if ip_candidates %}
 {% set host = ip_candidates[0] %}
 
+{% set endpoint  = host_name ~ '=https://' ~ host ~ ':2380' %}
+
+{%- set ca_cert = salt['mine.get']('*', 'kubernetes_etcd_ca_server').values()[0] %}
```
NicolasT: `kubernetes_etcd_ca_server` is a bit of an odd name for what is actually the CA certificate itself (if I got this correctly). Not in scope of this PR to fix it, I guess, but nonetheless, something to reconsider.

NicolasT (commented):
Mostly LGTM for now.

salt/_modules/cri.py
slaperche-scality: Used the `crictl pods` but with extra params, as suggested by Nicolas.

salt/_modules/cri.py
slaperche-scality: >_<, of course

salt/metalk8s/kubeadm/init/etcd/local.sls
slaperche-scality: Tracked by #890

salt/metalk8s/kubeadm/init/etcd/local.sls
```diff
@@ -8,9 +8,24 @@
...
+{% set endpoint  = host_name ~ '=https://' ~ host ~ ':2380' %}
+
 {%- set ca_cert = salt['mine.get']('*', 'kubernetes_etcd_ca_server').values()[0] %}
 {%- set ca_cert_b64 = salt['hashutil.base64_b64decode'](ca_cert) %}
 
+{#- Get the list of existing etcd node. #}
+{%- set etcd_endpoints = salt.saltutil.runner('mine.get', tgt='*', fun='etcd_endpoints').values() %}
+
+{#- Compute the initial state according to the existing list of node. #}
+{%- if etcd_endpoints %}
```
gdemonet: Not sure if clearer, but you could use `{%- set state = "existing" if etcd_endpoints else "new" -%}`

salt/_modules/metalk8s.py
```diff
@@ -29,3 +44,42 @@ def get_etcd_endpoint():
...
+    '''Add a new `etcd` node into the `etcd` cluster.
+
+    Arguments:
+        input_files (str): hostname of the new etcd node
+        output_file (str): endpoint of the new etcd node
+    '''
+    pod_name = 'etcd-{}'.format(__salt__['network.get_hostname']())
+    try:
+        client = kubernetes.config.new_client_from_config(
+            config_file='/etc/kubernetes/admin.conf'
```
gdemonet, line 85: This file is only available on the bootstrap node (and Salt master). It would be interesting to mention this in the docstring - e.g. "this method should only run on the Salt master".

gdemonet (commented):
So if I understood correctly, we'd like to invoke the orchestrate like so:

```
salt-run state.orchestrate metalk8s.orchestrate.deploy_etcd_on_new_node saltenv=metalk8s-2.0 --pillar='{"node_name": "node1", "node_ip": "10.0.0.2"}'
```

Could we use the Node object from the API server to retrieve the node IP instead? Maybe not for this PR though.

slaperche-scality:
> So if I understood correctly, we'd like to invoke the orchestrate like so:
> 
> ```
> salt-run state.orchestrate metalk8s.orchestrate.deploy_etcd_on_new_node saltenv=metalk8s-2.0 --pillar='{"node_name": "node1", "node_ip": "10.0.0.2"}'
> ```

Yup that's it!


> Could we use the Node object from the API server to retrieve the node IP instead? Maybe not for this PR though.

Not sure to understand, what would be the required changes?

salt/metalk8s/kubeadm/init/etcd/local.sls
slaperche-scality: Looks clearer to me, thanks!

salt/_modules/metalk8s.py
slaperche-scality: Yup, indeed it's worth to mention it.

slaperche-scality:
PR updated with the last fixes from the live debug done with @nootal, @ramanduh and @sayf-eddine-scality.

To be tested, this PR should be rebased on top of @nootal  and @ramanduh PRs.

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
```diff
@@ -0,0 +1,81 @@
...
+        repo:
+          online_mode: false
+          local_mode: false
+          host: {{ bootstrap_ip }}
+        registry_ip: {{ bootstrap_ip }}
+    - require:
+      - salt: Generate etcd certificates on a new node
+      - module: Refresh the list of etcd nodes
+
+Deploy updated manifests for existing etcd nodes:
```
slaperche-scality: Hum, really not sure about this one.

The more i think about it, the more it seems wrong.
I'll redo tests without it, looks like a wild guess made during our troubleshooting session.

NicolasT: I don't think there's a need for this indeed?

sayf-eddine-scality: agree there is no need to touch existing members

salt/_modules/metalk8s.py
TeddyAndrieux: If it's only to run on the `salt master` you should use a runner not a module. 
Here it's works because we use an orchestrate so we can run "modules" on the master but normally it's only runner who run on the master.

https://docs.saltstack.com/en/latest/ref/runners/

slaperche-scality: So where should I put this code?

TeddyAndrieux: Since it's a runner in `salt/_runners/xxx.py`

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
```diff
@@ -0,0 +1,66 @@
...
+        repo:
+          online_mode: false
+          local_mode: false
+          host: {{ bootstrap_ip }}
+        registry_ip: {{ bootstrap_ip }}
+
+Refresh the list of etcd nodes:
+  module.run:
+    - mine.update:
+      - mine_functions: metalk8s.get_etcd_endpoint
```
TeddyAndrieux: This will run the `mine.update` on the master which is basically not what we want.

`mine.update` it's a minion that push information to the master (NOT master who get information from all minions) so you should call mine.update on all salt-minions.

Something like that
```
Refresh the list of etcd nodes:
  salt.function:
    - name: mine.update
    - tgt: '*'
```

`mine_functions` is to update only specific informations maybe we don't care here.
Me personnaly I don't care update all mine or only the one you need, as you want :)

salt/_modules/metalk8s.py
NicolasT: I disagree, @TeddyAndrieux. This is an action that *could* be performed on any node (and, indeed, at some point we may decide to actually run it on a non-master minion, orchestrated by a runner). As such, it's a module. We simply happen to invoke it on the master for now.

Here's how I understand the Salt taxonomy:

- Module procedure: and 'action', something to be executed somewhere (likely a minion), very imperative.
- State procedure: code which exposes functionality to idempotent'ly reach a target state on a system by invoking 'module' actions. Imperative, in a way. In Ansible terms: task implementation I believe.
- State SLS: combination/invocation of state procedures making up a final state. Very declarative. In Ansible terms: role I believe.

- Runner procedure: invoked in context of the master, can invoke module procedures on master or minions, and other runners (again invoked in master context). Used to orchestrate things, mainly. In Ansible terms: playbook I believe.
- `state.orchestrate` runner: special case of a `runner` procedure which uses an SLS as 'input' and interprets it as a DSL instead of requiring one to write the Python code of the same effect.

TeddyAndrieux: I agree but for me the goal was that the master is the only "salt" who talk to the apiserver, No ?

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
NicolasT: I think we're trying to fix an issue we shouldn't have in the first place because the mine is not the right place to figure out the layout of the etcd cluster (cfr. #877).

salt/_modules/metalk8s.py
NicolasT: - Currently this code goes through the API-server because it uses an `exec` in an `etcd` Pod to perform a certain action. That's not the right approach, cfr. #878.
- Even when #878 gets implemented, the call would still happen from the master because that's the only one who's supposed to have the right credentials to invoke the command.
- This is a step which should be executed 'once', so it's part of 'orchestration'.

However.

Whilst we may invoke this procedure on the master, and as part of a larger playbook/orchestration, that doesn't mean the functionality itself (i.e. the procedure) couldn't be useful in other contexts. So, it's a module procedure, which may be invoked from an orchestration/runner, or in a state module if ever, or ...

In general:

- An 'action', imperative procedure to affect 'real world': module procedure.
- An 'orchestration', steps to be followed with centralized command (~ running on the master): runner procedure (or `state.orchestrate` SLS, which is kinda the same).

The question of 'where do *we* currently execute this code' is not part of the equation.

TeddyAndrieux: Ok make sense

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
slaperche-scality: Indeed, I confirmed that it was useless.

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
slaperche-scality: Ha that may explain why it wasn't working, thanks for the explanation @TeddyAndrieux.

gdemonet:
> > Could we use the Node object from the API server to retrieve the node IP instead? Maybe not for this PR though.
>
> Not sure to understand, what would be the required changes?

My point was that we did not need to pass `"node_ip": "..."` as pillar overrides when running this orchestrate, the `node_name`, which is also the minion ID, should be self-sufficient.

I don't think it's used in your code anyway ([in salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls](https://github.com/scality/metalk8s/pull/869/files#diff-1392062438ad991f95f2192e0e1af4f9)):
```
{%- if pillar['node_name'] in control_plane_ips.keys() and control_plane_ips[pillar['node_name']] %}
{%- set node_ip = control_plane_ips[pillar['node_name']][0] %}
{%- else %}
{%- set node_ip = 'localhost' %}
{%- endif %}
```

slaperche-scality:
> I don't think it's used in your code anyway

Yup, this part was reworked and I no longer need the IP to be passed in the pillar.

salt/_modules/metalk8s.py
```diff
@@ -0,0 +1,87 @@
...
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s'
+
+
+def __virtual__():
+    '''Check dependencies.'''
+    if HAS_DEPS:
+        return HAS_DEPS
```
TeddyAndrieux: More common
```suggestion
        return __virtualname__
```

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
```diff
@@ -0,0 +1,66 @@
...
+    - pillar:
+        repo:
+          online_mode: false
+          local_mode: false
+          host: {{ bootstrap_ip }}
+        registry_ip: {{ bootstrap_ip }}
+
+Refresh the list of etcd nodes:
+  salt.function:
+    - name: mine.update
```
slaperche-scality, line 35: Didn't find the right syntax to only update the `etcd` endpoints, so I refresh the whole mine…

salt/_modules/metalk8s.py
slaperche-scality: Done

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
TeddyAndrieux: Should be something like
```
Refresh the list of etcd nodes:
  salt.function:
    - name: mine.update
    - tgt: '*'
    - kwarg:
        mine_function: etcd_endpoints
```
But honnestly for me it's ok to refresh all the mine as we will likely not use mine in the futur for this

slaperche-scality: I think I tried that (maybe I used `kwargs` instead of `kwarg` though) and I got Salt complaining about me doing dirty things that won't work with Salt Fluorine or sth like that.

Agreed, since we won't rely on the mine for that in the future, let's keep it as-is.

charlesprost (approved):

slaperche-scality:
@bert-e approve

slaperche-scality:
@bert-e reset

salt/_modules/metalk8s.py
```diff
@@ -39,3 +54,58 @@ def get_workload_plane_ips():
...
+            name=pod_name, namespace='kube-system',
+            command=etcd_command,
+            stderr=True, stdin=False, stdout=False, tty=False
+        )
+        if err:
+            log.error('etcdctl error: %s', err)
+            raise CommandExecutionError('etcdctl: {}'.format(err))
+    except ApiException as exn:
+        log.exception('failed to run etcdctl')
+        raise exn
```
NicolasT, line 111: Just `raise`, not `raise exn`, right, to keep correct backtrace?

Issue #866, opened 2019-03-27
Add /health post install test for UI


MonPote:
- [ ] Add python test
- [ ] Add HTML file

NicolasT:
What would this HTML file do?

Issue #718, opened 2019-03-07
Check after deployment that the UI is reachable from outside the node


charlesprost:
- [x] Following #716 : have this test in Vagrant environment
- [x] Following #717 : have this test in CI environment

Success is defined by being able to make a successful HTTP request to the Platform UI.

PR #871, opened 2019-03-27
Add UI first test

MonPote:
WHAT ?

- [x] Add post install test to check if the UI is alive or not after the deployment  #866 #718

HOW ?

`tox -e tests -- -k "ui"`

It should be green !

### DIFF
--- a/tests/post/features/ui_alive.feature
+++ b/tests/post/features/ui_alive.feature
@@ -2,4 +2,4 @@
 Feature: The UI should be reachable
     Scenario: Reach the UI
         Given the Kubernetes API is available
-        Then we try to reach the UI
+        Then we can reach the UI
--- a/tests/post/features/ui_alive.feature
+++ b/tests/post/features/ui_alive.feature
@@ -1,0 +1,5 @@
+@post @ui
+Feature: The UI should be reachable
+    Scenario: Reach the UI
+        Given the Kubernetes API is available
+        Then we try to reach the UI
--- a/tests/post/steps/test_ui.py
+++ b/tests/post/steps/test_ui.py
@@ -1,5 +1,6 @@
 import json
 
+import pytest
 from pytest_bdd import scenario, then
 import requests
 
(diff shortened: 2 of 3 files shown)


tests/post/features/ui_alive.feature
```diff
@@ -0,0 +1,5 @@
+@post @ui
+Feature: The UI should be reachable
+    Scenario: Reach the UI
+        Given the Kubernetes API is available
+        Then we try to reach the UI
```
NicolasT:
```suggestion
        Then we can reach the UI
```

tests/post/steps/test_ui.py
```diff
@@ -0,0 +1,29 @@
...
+import requests
+
+
+# Scenarios
+@scenario('../features/ui_alive.feature', 'Reach the UI')
+def test_ui(host):
+    pass
+
+
+@then("we try to reach the UI")
```
NicolasT:
```suggestion
@then("we can reach the UI")
```

NicolasT (dismissed):
Consider making the BDD lines slightly more BDD'y, otherwise LGTM :+1:

MonPote:
@bert-e approve

thomasdanan (dismissed):

MonPote:
@bert-e reset

gdemonet (dismissed):
LGTM

tests/post/steps/test_ui.py
```diff
@@ -0,0 +1,43 @@
...
+@scenario('../features/ui_alive.feature', 'Reach the UI')
+def test_ui(host):
+    pass
+
+
+@then("we can reach the UI")
+def reach_UI(host, version, request):
+    with host.sudo():
+
+        bootstrap_ip = request.config.getoption("--bootstrap-ip")
```
gdemonet, line 17: This will not work, sadly. The MetalK8s UI container is only binding to the workload-plane IP, while this option is another "public" IP of the bootstrap node (neither control- nor workload-plane) that the test runner can see. We really need the test runner to have access to these networks.

It will be hard to find a valid workaround, so I'd suggest you skip the test if in multi-nodes context:

```python
if bootstrap_ip is not None:
    pytest.skip("Cannot yet run this test on multi-nodes deployment.")
else:
    ...
```

gdemonet (changes requested):
The tests will not work on multi-nodes deployment in the CI, and I don't see any "easy" workaround... Let's skip this test until we have access to the control- and workload-plane networks from the bastion VM.

tests/post/steps/test_ui.py
gdemonet: See #889

charlesprost: Agree, let's skip it for now.

MonPote: 👌

MonPote:
@bert-e reset

gdemonet (approved):

MonPote:
@bert-e reset

PR #873, opened 2019-03-27
Fix branding asset "rewrite or internal redirection cycle" problem

carlito-scality:
- Solve the "cycle redirection" of nginx


charlesprost (approved):

carlito-scality:
@bert-e approve

Issue #838, opened 2019-03-23
Keep build infra under `_build`


NicolasT:
Would it be possible to store `.doit.db` in `_build`, as well as the `.venv` we set up to run `doit`? It's good practice to keep 'non-source' stuff outside source folders whenever possible.

slaperche-scality:
Yes, it's doable:
- I create the virtualenv, so obviously I can create it elsewhere
- the location of the `doit` DB can be configured with the `dep_file` entry in `doit.cfg`

Now, I'm not sure that moving those into the `_build` directory is a good move (though I understand the desire/advantage to put build-generated stuffs out of the source tree).

The first problem we have is that currently the creation of the buildroot (`_build`) is handled by `doit`, which needs its virtualenv to run: we would have a " chicken or the egg" problem here.

Could be solved by moving the "ownership" (and responsibility to create it, but also of destroying it) of `_build` from `doit` to `doit.sh`.

Still, I see another problem: now, every `doit` command will create this `_build` directory even if it's not necessary (`doit lint` for instance, but also `doit list`, `doit info`, …) that's not really clean (IMHO).

I know that linter could be run out of `doit`, using `tox`-only directly, but as mentioned in our previous discussion:
- we lack an easy way to run all the linter in a single `tox` command
- if possible I would like to keep using tox for what it's good (creating and managing virtual environment), same for `doit` (running commands).
- that won't solve the spawn of a rogue `_build` directory for other legit but not build-related commands (`list`, `help`, `info`, …).

NicolasT:
> Could be solved by moving the "ownership" (and responsibility to create it, but also of destroying it) of _build from doit to doit.sh.

I disagree. Yes, `doit.sh` should create `_build/doit` (the venv), but the `doit` tasks should still depend on `_build` to exist/be created (which, granted, should always be the case when using `doit.sh`, but not when one happens to have `doit` installed system-wide and invoke it accordingly for whatever reason).

Basically, I think the current approach of "fail when `_build` wasn't created by `doit`, recorded in its database" is wrong. The aim for `doit` (and relevant tasks) is not *to create `_build`*, but the target is *for `_build` to exist* for future use by other targets of which `_build` is a dependency. Whether `doit` created the directory or not is irrelevant.

> Still, I see another problem: now, every doit command will create this _build directory even if it's not necessary (doit lint for instance, but also doit list, doit info, …) that's not really clean (IMHO).

[...]

PR #875, opened 2019-03-28
build: always forget execution of cleaned tasks

slaperche-scality:
This should solve issues where we have stale information in doit.db
after a `clean` (after a call to `clean` we expect to start anew).

If this doesn't solve our problems, we will take more drastic measures.

Refs: #838


charlesprost (approved):

slaperche-scality:
@bert-e approve

Issue #837, opened 2019-03-23
The new build framework doit only support python3.6+ but does not enforce it


charlesprost:
From `doit.sh` we have:
`PYTHON_SYS="${PYTHON_SYS:-python3}"`

But if `python3` redirects to `python3.4` then you hit this error:

```
  File "/home/nootal/projects/metalk8s_bis/buildchain/buildchain/__init__.py", line 14
    ROOT: Path = (Path(__file__)/'../../../').resolve()
        ^
SyntaxError: invalid syntax
 ```

due to https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep526

We need to enforce python3.6 at least.

**Comment by slaperche-scality** [comment]
Yeah, I know.
I wanted to do it, but after looking up how to check a version number in shell I thought "well let's do that later" ^^

I should have created a ticket though…

That being said, we can avoid dirty parsing and complicated comparison in shell by using more Python (thanks `hexversion`)

```python
--- a/doit.sh
+++ b/doit.sh
@@ -2,6 +2,9 @@
 
 # System-wide Python 3 command.
 PYTHON_SYS="${PYTHON_SYS:-python3}"
+# Hexversion for Python 3.6.0
+PYTHON_MIN_VERSION=$(printf "%d\n" 0x3060000)
+PYTHON_SYS_VERSION=$($PYTHON_SYS -c 'import sys; print(sys.hexversion)')
 # Buildchain location.
 BUILDCHAIN=buildchain
 # Location of the virtual environment for the buildchain.
@@ -13,6 +16,13 @@ WITNESS_FILE="${BUILDENV}/installed.tstamp"
 # File containing environment variables.
 DOTENV=./.env
 
+# Check Python version.
+if [ "$PYTHON_SYS_VERSION" -lt "$PYTHON_MIN_VERSION" ]
+then
+    echo "$($PYTHON_SYS -V) too old, 3.6 minimum is required" >&2
+    exit 1
+fi
+
 # Can't use `[ file1 -nt file2 ]` directly because it's not POSIX :'(
 REQ_TSTAMP=$(stat -c %Y "${REQUIREMENTS}")
 WIT_TSTAMP=$(stat -c %Y "${WITNESS_FILE}" 2> /dev/null || echo '0')
```

I can do the patch on Monday if you want ;-)

NicolasT:
Something like this?

```sh
MINIMUM_PYTHON_VERSION='(3, 6)'

[...]

PR #876, opened 2019-03-28
build: add a check to enforce Python version

slaperche-scality:
It is better to check for the version early (and fail with an explicit
error message) that to create the virtual environment and then fail with
a weird syntax error.

Closes: #837


charlesprost (approved):

slaperche-scality:
@bert-e approve

PR #815, opened 2019-03-21
Have salt-minion deployed and configured on a new node

charlesprost:
Fixes #739 

The goal of this PR is to provide a way to have a new node connected to the salt-master through a freshly installed salt-minion.

Installing stuff on this new node to make it part of the control plane is not part of this PR.

Things you need to do to use this PR:

1. `vagrant destroy -f (or vagrant snapshot restore...) && ./doit.sh vagrantup`
1. Start a new node: `vagrant up node1`
1. (_could be done by default later_) Install `kubectl` on the bootstrap node and test it: `vagrant ssh bootstrap` then `sudo su` then `yum install -y kubectl --enablerepo="metalk8s-*" && export KUBECONFIG=/etc/kubernetes/admin.conf && kubectl get pods -n "kube-system"`
1. On the boostrap node, have a file describing node1: `vagrant ssh bootstrap` then `vi node1.yaml` then 
```
apiVersion: v1
kind: Node
metadata:
  name: node1
  labels:
    beta.kubernetes.io/os: linux
  annotations:
    metalk8s.scality.com/ssh-user: vagrant
    metalk8s.scality.com/ssh-port: "22"
    metalk8s.scality.com/ssh-host: <NODE1_IP_ADDRESS>
    metalk8s.scality.com/ssh-key-path: /etc/metalk8s/pki/preshared_key_for_k8s_nodes
    metalk8s.scality.com/ssh-sudo: "true"
```
while inserting the IP address of node1 and create it: `vagrant ssh bootstrap` then `sudo su` then `kubectl create -n "kube-system" -f node1.yaml`
1. On the boostrap node, connect to the salt-master container: `vagrant ssh bootstrap` then `sudo su` then `kubectl -n "kube-system" exec -it salt-master-bootstrap bash`
1. On the salt-master container, ping node1 to ensure SSH connection is OK: `salt-ssh -i node1 test.ping --roster=kubernetes`
1. On the salt-master container, deploy the salt-minion and accept its key: `salt-run state.orch metalk8s.orchestrate.deploy_salt_minion_on_new_node saltenv=metalk8s-2.0 pillar='{"bootstrap_id": "bootstrap", "node_name": "node1"}' ; salt-key -y -a node1`
1. Then wait a few sec and check it's working: `salt '*' test.ping`

Yet to do:

- [x] test if the roster needs to be synced before launch the orchestrate runner in the salt-master container
- [x] put the bootstrap ip (the one linked to the control plane network) into the mine during the bootstrap stage
[...]

Issue #890, opened 2019-03-29
Review name used in the Salt mine


slaperche-scality:
Some names, such as `kubernetes_etcd_ca_server` are a bit odd (e.g. `sth_sth_ca_server` to represent a certificate).

We may want to review the naming of what we put in the mine, to check if the name used are meaningful.

_Originally posted by @NicolasT in https://github.com/scality/metalk8s/pull/869_

sayf-eddine-scality:
target branch should be 2.0

PR #892, opened 2019-03-29
Improvement/orchestrate runner to bootstrap worker node

ramanduh:
WIP but with this PR we should be able to bootstrap without hardcoding IPs everywhere:

Step to tests:
Follow the procedure described here to create bootstrap and node1: https://github.com/scality/metalk8s/pull/815

On the bootstrap node, add a label to the previously created node 1:
```
apiVersion: v1
kind: Node
metadata:
  name: node1
  annotations:
    metalk8s.scality.com/ssh-user: vagrant
    metalk8s.scality.com/ssh-port: "22"
    metalk8s.scality.com/ssh-host: 172.21.254.7
    metalk8s.scality.com/ssh-key-path: /etc/metalk8s/pki/preshared_key_for_k8s_nodes
    metalk8s.scality.com/ssh-sudo: "true"
  labels:
    beta.kubernetes.io/os: linux
```

In the salt master container
```
salt '*' saltutil.sync_all saltenv=metalk8s-2.0
salt-run state.orch metalk8s.orchestrate.worker_plane saltenv=metalk8s-2.0 pillar='{"bootstrap_id": "bootstrap", "node_name": "node1"}'
```

On the bootstrap node:
Restart kube-proxy:
```
kubectl -n kube-system delete pods $(kubectl -n kube-system get pods|grep proxy|awk '{print$1}')
systemctl restart kubelet
```

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -75,6 +75,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/orchestrate/bootstrap_without_master.sls'),
     Path('salt/metalk8s/orchestrate/bootstrap_with_master.sls'),
     Path('salt/metalk8s/orchestrate/deploy_salt_minion_on_new_node.sls'),
+    Path('salt/metalk8s/orchestrate/worker_plane.sls'),
     Path('salt/metalk8s/bootstrap/addons.sls'),
     Path('salt/metalk8s/bootstrap/calico.sls'),
     Path('salt/metalk8s/bootstrap/certs/ca.sls'),
--- a/salt/metalk8s/containerd/installed.sls
+++ b/salt/metalk8s/containerd/installed.sls
@@ -1,5 +1,6 @@
 {%- from "metalk8s/macro.sls" import pkg_installed with context %}
 {%- from "metalk8s/map.jinja" import repo with context %}
+{%- from "metalk8s/map.jinja" import defaults with context %}
 
 include:
   - metalk8s.repo
@@ -22,3 +23,17 @@ Install containerd:
       - test: Repositories configured
       - pkg: Install runc
       - pkg: Install container-selinux
+
+Configure registry IP in containerd conf:
+  file.managed:
+    - name: /etc/containerd/config.toml
+    - makedirs: true
+    - contents: |
+        [plugins]
+          [plugins.cri]
+            [plugins.cri.registry]
+              [plugins.cri.registry.mirrors]
+                [plugins.cri.registry.mirrors."{{ defaults.registry_ip }}"]
+                  endpoint = ["http://{{ defaults.registry_ip }}:5000"]
+    - require:
+      - pkg: Install containerd
--- a/salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
+++ b/salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
@@ -2,9 +2,22 @@
 
 {%- set hostname = salt.network.get_hostname() %}
 
+{%- set ca_cert_b64 = salt['mine.get']('*', 'kubernetes_ca_server').values()[0] %}
+{%- set ca_cert = salt['hashutil.base64_b64decode'](ca_cert_b64) %}
+
 {%- set cert_info = {'CN': 'system:node:' ~ hostname, 'O': 'system:nodes'} %}
 
 {{ kubeconfig('kubelet', cert_info) }}
+
+Ensure CA cert is present:
+  file.managed:
+    - name: /etc/kubernetes/pki/ca.crt
+    - user: root
+    - group : root
+    - mode: 644
+    - makedirs: True
+    - dir_mode: 755
+    - contents: {{ ca_cert.splitlines() }}
 
 Configure kubelet service:
   file.managed:
@@ -28,3 +41,4 @@ Restart kubelet:
     - enable: True
     - watch:
       - file: Configure kubelet service
+      - file: Ensure CA cert is present
--- a/salt/metalk8s/orchestrate/bootstrap_with_master.sls
+++ b/salt/metalk8s/orchestrate/bootstrap_with_master.sls
@@ -31,6 +31,8 @@ Bootstrap control plane:
       - metalk8s.bootstrap.etcd
     - require:
       - salt: Bootstrap client certs
+    - pillar:
+        registry_ip: {{ pillar.get('registry_ip') }}
 
 Bootstrap node:
   salt.state:
@@ -42,3 +44,5 @@ Bootstrap node:
       - metalk8s.bootstrap.calico
     - require:
       - salt: Bootstrap control plane
+    - pillar:
+        registry_ip: {{ pillar.get('registry_ip') }}
--- a/salt/metalk8s/orchestrate/worker_plane.sls
+++ b/salt/metalk8s/orchestrate/worker_plane.sls
@@ -1,0 +1,73 @@
+{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ips') %}
+{%- if pillar['bootstrap_id'] in control_plane_ips.keys() and control_plane_ips[pillar['bootstrap_id']] %}
+{%- set control_plane_ip = control_plane_ips[pillar['bootstrap_id']][0] %}
+{%- else %}
+{%- set control_plane_ip = 'localhost' %}
+{%- endif %}
+
+{% set node = pillar['node_name'] %}
+
+Run bootstrap prechecks:
+  salt.state:
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls: metalk8s.bootstrap.precheck
+
+Run bootstrap kubelet:
+  salt.state:
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls:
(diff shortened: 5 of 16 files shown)


NicolasT:
>     beta.kubernetes.io/os: linux

Please don't use that label ;-) Use one or more 'node-role' ones :)

salt/metalk8s/containerd/installed.sls
```diff
@@ -22,3 +23,18 @@ Install containerd:
       - test: Repositories configured
       - pkg: Install runc
       - pkg: Install container-selinux
+
+# TODO: Use file.serialize (Require salt toml serializer)
```
NicolasT: Please file debt tickets instead of keeping `TODO` lines around.

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
```diff
@@ -2,10 +2,23 @@
 
 {%- set hostname = salt.network.get_hostname() %}
 
+{%- set ca_cert = salt['mine.get']('*', 'kubernetes_ca_server').values()[0] %}
```
NicolasT: #890

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,93 @@
+{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ips') %}
+{%- if pillar['bootstrap_id'] in control_plane_ips.keys() and control_plane_ips[pillar['bootstrap_id']] %}
+{%- set control_plane_ip = control_plane_ips[pillar['bootstrap_id']][0] %}
+{%- else %}
+{%- set control_plane_ip = 'localhost' %}
+{%- endif %}
+
+{% set node = pillar.get('node_name', 'node1') %}
```
NicolasT: This is not a reasonable default. If the value isn't provided, this should error out.

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,93 @@
...
+  salt.state:
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls: metalk8s.bootstrap.precheck
+
+Run bootstrap kube client:
+  salt.state:
+    - saltenv: {{ saltenv }}
+    - tgt: {{ node }}
+    - sls: metalk8s.bootstrap.kubernetes_client
```
NicolasT: Why would you need `python-kubernetes` to be installed on nodes? This should only ever be required in the `salt-master` Pod anyway.

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,93 @@
...
+
+Run bootstrap kubelet:
+  salt.state:
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls:
+      - metalk8s.bootstrap.kubelet_install
+    - pillar:
+        repo:
+          online_mode: false
```
NicolasT: Why do these values need to be overridden?

NicolasT (dismissed):
A couple of comments, but good work overall :+1:

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,93 @@
+{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ips') %}
```
NicolasT, line 1: Since everything here is on the target node and a 'state', over time this `playbook` should basically only invoke `state.highstate` on the target node, given we have an appropriate 'top' generated by a `master_tops` module.

salt/metalk8s/orchestrate/worker_plane.sls
ramanduh: Right, will remove it

salt/metalk8s/orchestrate/worker_plane.sls
ramanduh: No specific reason, just a bad copy paste

salt/metalk8s/orchestrate/worker_plane.sls
ramanduh: Agree

salt/metalk8s/orchestrate/worker_plane.sls
ramanduh: fixed

salt/metalk8s/orchestrate/worker_plane.sls
ramanduh: unused indeed. Removed

salt/metalk8s/containerd/installed.sls
```diff
@@ -22,3 +23,17 @@ Install containerd:
...
+Configure registry IP in containerd conf:
+  file.managed:
+    - name: /etc/containerd/config.toml
+    - makedirs: true
+    - contents: |
+        [plugins]
+          [plugins.cri]
+            [plugins.cri.registry]
+              [plugins.cri.registry.mirrors]
+                [plugins.cri.registry.mirrors."{{ defaults.registry_ip }}"]
```
gdemonet, line 36: I might be wrong, but this TOML can also be written:
```
[plugins.cri.registry.mirrors."{{ defaults.registry_ip }}"]
  endpoint = ...
```
, without the need for all sections before.

ramanduh: Yes, it can be written like that, I just put it like it because it is the output of `containerd config`

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
ramanduh: We may want to rename it in another PR?

charlesprost: It's the opposite in terms of variable names: you have at first `ca_cert_b64` then `ca_cert` when you decode it ;-)

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
```diff
@@ -2,10 +2,23 @@
...
+{%- set ca_cert_b64 = salt['hashutil.base64_b64decode'](ca_cert) %}
+
 {%- set cert_info = {'CN': 'system:node:' ~ hostname, 'O': 'system:nodes'} %}
 
 {{ kubeconfig('kubelet', cert_info) }}
 
+Ensure CA cert is present:
```
charlesprost, line 12: To be done later probably: put this into a dedicated state and call it when necessary because it could be used elsewhere.

salt/metalk8s/orchestrate/bootstrap_with_master.sls
```diff
@@ -31,6 +31,8 @@ Bootstrap control plane:
       - metalk8s.bootstrap.etcd
     - require:
       - salt: Bootstrap client certs
+    - pillar:
+        registry_ip: {{ pillar.get('registry_ip') }}
```
charlesprost, line 28: Can't we have this merged into map.jinja or something alike so we don't have to redefine it from the orchestrate if we don't change the default value?

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,73 @@
...
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls: metalk8s.bootstrap.precheck
+
+Run bootstrap kubelet:
+  salt.state:
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }}
+    - sls:
+      - metalk8s.bootstrap.kubelet_install
```
charlesprost, line 21: minor: always use the same indentation style: either `sls: toto` or `sls: \newline - toto`

salt/metalk8s/orchestrate/worker_plane.sls
```diff
@@ -0,0 +1,73 @@
...
+    - tgt: {{ node }}
+    - sls:
+      -  metalk8s.calico.installed
+      -  metalk8s.calico.configured
+    - pillar:
+        repo:
+          host: {{ control_plane_ip }}
+        registry_ip: {{ control_plane_ip }}
+
+# TODO: Investigate why we have to reload systemd and restart kubelet
```
charlesprost, line 62: debt ticket please :)

buildchain/buildchain/salt_tree.py
```diff
@@ -75,6 +75,7 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
     Path('salt/metalk8s/orchestrate/bootstrap_without_master.sls'),
```
charlesprost, line 76: Generic comment on the commit messages: please add the issue reference in each commit message

charlesprost (changes requested):

salt/metalk8s/kubeadm/init/kubeconfig/kubelet.sls
ramanduh: hum, right, will change

salt/metalk8s/orchestrate/bootstrap_with_master.sls
ramanduh: we already have a ticket for that

charlesprost (dismissed):

ramanduh:
@bert-e approve

ramanduh:
@bert-e reset

charlesprost (approved):

PR #893, opened 2019-03-29
UI - Nodes information page - Pods list 

carlito-scality:
-https://github.com/scality/metalk8s/issues/904
Screenshot 2019-04-01 at 13 39 46


ui/src/containers/NodeInformation.js
```diff
@@ -0,0 +1,190 @@
...
+    return nodes.find(node => node.name === props.match.params.id) || {};
+  } else {
+    return {};
+  }
+};
+
+const getPodsFromUrl = (state, props) => {
+  const pods = state.app.pods.list || [];
+  if (props && props.match && props.match.params && props.match.params.id) {
+    return pods.filter(pod => pod.nodeName === props.match.params.id) || {};
```
MonPote: As we discuss, I think you want to return an array instead of an object.
```js
return pods.filter(pod => pod.nodeName === props.match.params.id) || [];
```

ui/src/containers/NodeInformation.js
```diff
@@ -0,0 +1,190 @@
...
+  } else {
+    return {};
```
MonPote: Same here.

MonPote (approved):
I did the change.

carlito-scality:
@bert-e  approve

PR #894, opened 2019-03-29
Improve default Calico config

NicolasT:
- Ensure Calico uses the node workload-plane IPs instead of something random
- Disable IPIP for non-cross-subnet communication


sayf-eddine-scality (approved):

NicolasT:
@bert-e approve

NicolasT:
@bert-e status

Issue #887, opened 2019-03-28
_vagrankey task may hang while waiting for a confirmation to replace the existing key


charlesprost:
Sometimes, while running `./doit.sh vagrantup` or `./doit.sh _vagrantkey`, the `_vagrantkey` task will hang because it's waiting for a confirmation to replace or not the existing key pair.
Press CTRL+C when stuck on this `. _vagrantkey` task, then you'll see the confirmation prompt.

That may be related to some discrepancy between the doit DB and the presence of the key pair.

slaperche-scality:
Cannot reproduce on my side with "normal" use cases.

The two ways I found to reproduce it are:
- create the key yourself (before `doit` has a chance to create it) and then run `doit vagrantup`
- explicitly ask `doit` to forget about it (not by running `doit clean vagrantup`, because then it would delete the key, but by running `doit forget vagrantup` which would remove the key from the DB but keep it on disk).

Now, the root cause of this is our use of [doit.tools.run_once](http://pydoit.org/uptodate.html#run-once) (and my miscomprehension of its subtleties) which I recommended (mea culpa) because I find it more explicit than a bare `True`. `run_once` does rely on a value inside the DB in addition to the presence of the target(s) on disk.

IAt first I though that `True` and `run_once` were exactly the same (and hence favored `doit.tools.run_once` because it clearer in what it express) but there is a subtle difference that I noticed while re-reading the doc carefully:
- `True`: the computation of `uptodate` only depends on the presence of the target(s).
- ditto for `run_once` BUT it forces at least one execution managed by doit.

To be safe here, even in the "weird" cases (key created externally/`forget` explicit), we could/should use `True` instead of `run_once`.

Also, now that we now about the subtle difference we may want to review the other places where we used `run_once`, to ponder if `True` is more adapted or not.

PR #901, opened 2019-04-01
build: use `True` instad of `run_once`

slaperche-scality:
Unlike what I thought at first, `True` and `doit.tools.run_once` are not
equivalent for `uptodate`:

- `True`: the computation of `uptodate` only depends on the presence of
  the target(s).
- `run_once`: ditto BUT it forces at least one execution managed by
  doit.

In most cases, we only care about the existence of the file, the fact
that if was created by `doit` or not doesn't matter, hence this commit
to reflect that.

Closes: #887

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -29,8 +29,6 @@ Overview;
 from pathlib import Path
 from typing import Dict, Iterator, List, Tuple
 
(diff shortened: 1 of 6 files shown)


alexandre-allard (approved):

slaperche-scality:
@bert-e approve

Issue #833, opened 2019-03-22
Use a mine_function directly in the pillar instead of a state that pushes IPs in the mine


charlesprost:
Following #739

Instead of calling a state that pushes IPs in the mine, let's use a `mine_function` in the pillar instead.

From @TeddyAndrieux 

> Quick explanation have a custom module that return the current ip (or hostname) ONLY IF the host is a control plane so maybe check pods or directly ask api server. 
> And use this custom module as a mine_function so when we do `mine.update` we advertise all REAL control plane nodes automatically

Code from @sayf-eddine-scality that can help to start with:
```
# pillar/mine.py
mine_functions:
  repo_ip:
      - mine_function: metalk8s_mine.get_container_ip

# pillar/top.py
metalk8s-2.0:
  '*':
    - metalk8s
    - repositories
    - mine

# _modules/metalk8s_mine.py
'''
Functions to retrieve infos, IPs and peers.
'''

import logging
import socket

log = logging.getLogger(__name__)


__virtualname__ = 'metalk8s_mine'


def __virtual__():
    return __virtualname__


def get_container_ip():
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    res = sock.connect_ex(('127.0.0.1', 8080))
    if res == 0:
        return __salt__["network.ip_addrs"]("eth1")
    else:
        return "port closed"
```

```
[root@bootstrap vagrant]# ${SALT} bootstrap mine.get '*' repo_ip
bootstrap:
    ----------
    bootstrap:
        172.21.254.4
```

PR #909, opened 2019-04-02
Use a mine_function in the pillar to advertise control_plane and work…

TeddyAndrieux:
Use a mine_function in the pillar to advertise control_plane and workload_plane Ips
Retrieve k8s label in custom pillar

Fixes: #833

### DIFF
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -1,6 +1,8 @@
 import logging
+import os
 import salt.utils.files
 import salt.utils.yaml
+
 
 log = logging.getLogger(__name__)
 
@@ -10,7 +12,7 @@ DEFAULT_SERVICE_NETWORK = '10.96.0.0/12'
 
 
 def _load_config(path):
-    log.debug('Loading MetalK8s configuration from {}'.format(path))
+    log.debug('Loading MetalK8s configuration from %s', path)
 
     config = None
     with salt.utils.files.fopen(path, 'rb') as fd:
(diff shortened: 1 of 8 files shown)


salt/_pillar/metalk8s.py
```diff
@@ -37,9 +39,30 @@ def _load_networks(config_data):
     }
 
 
+def _get_labels(config="/etc/kubernetes/admin.conf", context="kubernetes-admin@kubernetes"):
+    # TODO: Use the "bootstrap_config" instead of /etc/kubernetes/admin.conf
```
slaperche-scality: Do we have an debt issue about that? If yes, you can refer to it in the TODO (and if not, we should create one, otherwise this TODO will probably stay here forever).

salt/_pillar/metalk8s.py
```diff
@@ -37,9 +39,30 @@ def _load_networks(config_data):
     }
 
 
+def _get_labels(config="/etc/kubernetes/admin.conf", context="kubernetes-admin@kubernetes"):
```
slaperche-scality: Line too long:

```
def _get_labels(config="/etc/kubernetes/admin.conf",
                context="kubernetes-admin@kubernetes"): 
```
or
```
def _get_labels(i
    config="/etc/kubernetes/admin.conf", context="kubernetes-admin@kubernetes"
):
```

salt/_pillar/metalk8s.py
```diff
@@ -37,9 +39,30 @@ def _load_networks(config_data):
...
+    if not hostname or not os.path.isfile(config):
+        return labels
+
+    try:
+        labels = __salt__['kubernetes.node_labels'](
+            name=hostname,
+            kubeconfig=config,
+            context=context
+        )
+    except Exception as exc:  # pylint: disable=broad-except
```
slaperche-scality, line 58: Why `Exception`, seems like the code only raises `CommandExecutionError`, no?

slaperche-scality (changes requested):

salt/_pillar/metalk8s.py
TeddyAndrieux: Don't want to break the whole external pillar if we got a ValueError, IndexError, NoneType or what ever

salt/_pillar/metalk8s.py
TeddyAndrieux: Don't know if it's a real debt. It's just that one day we will likely generate a certs for the salt-master to access apiserver then we will need to use this certs (don't know if the ticket exists)
And btw I think it's not `bootstrap_config`.

I will update the TODO and add ticket number

salt/_pillar/metalk8s.py
TeddyAndrieux: True

TeddyAndrieux:
@bert-e reset

salt/_pillar/metalk8s.py
slaperche-scality: Fair enough.

slaperche-scality (approved):

TeddyAndrieux:
@bert-e approve

Issue #888, opened 2019-03-28
Running the deployment orchestrate is working but return rc=1


charlesprost:
While running the deployment of a new node with the orchestrate on the salt master, some errors about `crictl` waiting for the pods to start will appear, and the return code of the salt-run will be 1. Maybe those two events are unrelated though.

Would be great if the RC was 0.

PR #913, opened 2019-04-02
Redirect restarting salt-minion command output to /dev/null

TeddyAndrieux:
Execute this:
`salt-run state.orch metalk8s.orchestrate.deploy_salt_minion_on_new_node saltenv=metalk8s-2.0 pillar='{"bootstrap_id": "bootstrap", "node_name": "node1"}'`

Before this PR:
```
[ERROR   ] No JSON object could be decoded
[...]
Summary for bootstrap_master
------------
Succeeded: 0 (changed=1)
Failed:    1
------------
```
because of the output of the salt-call at the beginning of the state result
```
local:
  True
```

Now:

```
Summary for bootstrap_master
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
```

Fixes: #888

### DIFF
--- a/salt/metalk8s/salt/minion/running.sls
+++ b/salt/metalk8s/salt/minion/running.sls
@@ -1,6 +1,6 @@
 Restart salt-minion:
   cmd.wait:
-    - name: 'salt-call --local service.restart salt-minion'
+    - name: 'salt-call --local service.restart salt-minion > /dev/null'
     - bg: true
 
 Wait until salt-minion restarted:


TeddyAndrieux:
@bert-e approve

slaperche-scality (commented):
Could explain in the commit message why this change (like you did in the PR description: "We redirect the output of the comand to `/dev/null`, otherwise Salt tries to decode it as JSON, which is not valid" or sth like that)

TeddyAndrieux:
> Could explain in the commit message why this change (like you did in the PR description: "We redirect the output of the comand to `/dev/null`, otherwise Salt tries to decode it as JSON, which is not valid" or sth like that)

Yes but the problem is only for orchestrate, classic states works, but indeed I will add it in the commit msg

TeddyAndrieux:
@bert-e reset

slaperche-scality (approved):

PR #921, opened 2019-04-03
CI: restrict control and workload plane networks on the single node env to avoid confusion while picking IPs

### LINKED ISSUE #920: Conflict between control plane network IPs when running tests in CI single node env
In the CI we run the tests, and especially we retry the bootstrap script to see if it's idempotent.

In the CI, in the single node env, we have two IPs on the worker:

1. The one from Openstack: 10.100.A.B
1. The one from the pod network we set: 10.233.C.D

In the CI, we define the control plane network as 10.0.0.0/8 (hence including both previous IPs).

Then when retrying the bootstrap script, we pick one IP on this network, and that could be either the one from Openstack, or the 10.233 one. And if it's the second one, then it means we bind all our services on it, and nothing works anymore because we still are targeting the 10.100 one to address the API server.

Let's be stricter and consider 10.100.0.0/16 as our control plane network (and also workload plane network) in the CI single node env.

charlesprost:
Fixes #920

Our ranges were too wide and could include other IPs that were not part of those networks.
By narrowing them down, we are sure we pick the right IP.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -130,8 +130,8 @@ stages:
             apiVersion: metalk8s.scality.com/v1alpha1
             kind: BootstrapConfiguration
             networks:
-              controlPlane: 10.0.0.0/8
-              workloadPlane: 10.0.0.0/8
+              controlPlane: 10.100.0.0/16
+              workloadPlane: 10.100.0.0/16
             END
             EOF
       - ShellCommand:


gdemonet (approved):

slaperche-scality (approved):

charlesprost:
@bert-e approve

PR #926, opened 2019-04-03
buildchain: enforce strict Dockerfile COPY/ADD filedeps

jbertran:
Aggressively abort a `LocalImage` container build when the task's Dockerfile contains file dependencies (through `COPY` or `ADD` directives) that are not reflected by the task's advertised `file_dep` field in its specification.

This encourages including only necessary files, and prevents desyncs between support files and images.

Play around with it by removing dependencies from any existing `LocalImage` task in the buildchain, and observe the mayhem unfold.

### DIFF
--- a/buildchain/buildchain/targets/local_image.py
+++ b/buildchain/buildchain/targets/local_image.py
@@ -14,8 +14,11 @@ All of these actions are done by a single task.
 
 
 import operator
+import re
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Match, Optional, Union
+
+from doit.exceptions import TaskError # type: ignore
 
 from buildchain import config
 from buildchain import constants
@@ -23,6 +26,55 @@ from buildchain import coreutils
 from buildchain import types
 
 from . import image
+
+
+class DockerFileDep:
+    """A Dockerfile directive line creating external file dependencies."""
+
+    def __init__(
+        self,
+        line_no: int,
+        line_contents: str,
+        line_match: Match[str],
+        dockerfile_path: Path
+    ):
+        self.line_no = line_no
+        self.line_contents = line_contents
+        self.match = line_match
+        self.dockerfile_path = dockerfile_path
+
+    def expand_dep(self) -> List[Path]:
+        """Expand the Dockerfile path dependency to regular files."""
+        source = self.match.group('src')
+        path = self.dockerfile_path.parent/Path(source)
+        # Simple file
+        if path.is_file():
+            return [path]
+        # Directory or `.`
+        if path.is_dir():
+            return list(coreutils.ls_files_rec(path))
+        # Globs - `*` or specific e.g. `*.py`, `*.repo`
+        return [
+            sub_path
+            for sub_path in path.parent.glob(path.name)
+            if sub_path.is_file()
+        ]
+
+    def format_errors(self, error_paths: List[Path]) -> List[str]:
+        """Format paths in error with relevant information."""
+        error_line = "l.{}: '{}' - missing {}"
+        return [
+            error_line.format(self.line_no, self.line_contents, error_path)
+            for error_path in error_paths
+        ]
+
+    def verify(self, file_deps: List[Path]) -> List[str]:
+        """Verify Dockerfile dependencies against external dependency list."""
+        missing_deps = []
+        for dependency in self.expand_dep():
+            if not dependency in file_deps:
+                missing_deps.append(dependency)
+        return self.format_errors(missing_deps)
 
 
 class LocalImage(image.ContainerImage):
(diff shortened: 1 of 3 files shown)


buildchain/buildchain/targets/local_image.py
```diff
@@ -25,6 +28,20 @@
 from . import image
 
 
+def format_deps(
+    missing_deps: List[Tuple[int, str, str]],
+    filename: Path
+) -> str:
+    """Format error message for computed dependencies."""
```
slaperche-scality: Would be nice to have description of the tuple fields in the docstring, or maybe use a `namedtuple` (`DockerFileDep`?) instead.

If you use a namedtuple, you could also use it in the return value of `load_deps_from_dockerfile`.

In fact, you could even write a class `DockerFileDep` which could be created from regexp match (and implement `__str__` for format, …).

That would probably make the code clearer.

buildchain/buildchain/targets/local_image.py
```diff
@@ -84,6 +152,10 @@ def show(_task: types.Task) -> str:
 
     def _build_actions(self) -> List[types.Action]:
         """Build a container image locally."""
+        actions : List[types.Action] = []
```
slaperche-scality: You don't need to annotate local variable.

Also, you could write:

```python
actions = [self.check_dockerfile_dependencies]
```

buildchain/buildchain/targets/local_image.py
```diff
@@ -25,6 +28,20 @@
 from . import image
 
 
+def format_deps(
+    missing_deps: List[Tuple[int, str, str]],
+    filename: Path
```
slaperche-scality: `dockerfile` would be less generic than `filename`.

buildchain/buildchain/targets/local_image.py
```diff
@@ -62,10 +79,61 @@ def __init__(
...
     build_args   = property(operator.attrgetter('_build_args'))
+    dep_re = re.compile(
+        r'^\s*(COPY|ADD)( --[^ ]+)* (?P<src>[^ ]+) (?P<dst>[^ ]+)\s*$'
+    )
+
+    def load_deps_from_dockerfile(self) -> List[Tuple[int, str, str]]:
+        """Compute file dependencies from Dockerfile."""
+        dep_keywords = ('COPY', 'ADD')
+
+        with open(self.dockerfile, 'r') as dockerfile:
```
slaperche-scality: Add `encoding='utf-8'`.

And IIRC `self.dockerfile` is a `Path` object so you can use `self.dockerfile.open('r', encoding='utf8')` or even `lines = self.dockerfile.read_text(encoding='utf-8').split('\n')`

slaperche-scality (changes requested):
Overall LGTM, you may want to do the class approach for code a bit cleaner.

buildchain/buildchain/targets/local_image.py
jbertran: Stealing the first alternative, I find it more legible :)

buildchain/buildchain/targets/local_image.py
```diff
@@ -25,6 +28,58 @@
...
+    def expand_dep(self) -> List[Path]:
+        """Expand the Dockerfile path dependency to regular files."""
+        source = self.match.group('src')
+        path = self.dockerfile_path.parent/Path(source)
+        # Simple file
+        if path.is_file():
+            return [path]
+        # Directory or `.`
+        if path.is_dir():
+            return [
```
slaperche-scality: Could be written `return list(coreutils.ls_files_rec(path))` no?

buildchain/buildchain/targets/local_image.py
```diff
@@ -62,10 +117,56 @@ def __init__(
...
+        """Compute file dependencies from Dockerfile."""
+        dep_keywords = ('COPY', 'ADD')
+
+        with self.dockerfile.open('r', encoding='utf8') as dockerfile:
+            docker_lines = dockerfile.readlines()
+
+        dep_lines = [
+            (pos, line.strip(), self.dep_re.match(line.strip()))
+            for (pos, line) in enumerate(docker_lines, 1)
+            if line.strip().startswith(dep_keywords)
```
slaperche-scality: `lstrip` would be enough for `startswith`

buildchain/buildchain/targets/local_image.py
```diff
@@ -62,10 +117,56 @@ def __init__(
...
+        errors: List[str] = []
+        for dep in deps:
+            errors.extend(dep.verify(self.file_dep))
+        if errors:
+            error_message = error_tplt.format(
+                dockerfile=self.dockerfile,
+                errors='\n\t'.join(errors)
+            )
+            return TaskError(msg=error_message)
+        return True
```
slaperche-scality: `return None` would be more consistent (since you don't return `False` in the other case).

slaperche-scality (dismissed):
Still some remarks, but nothing too serious.

buildchain/buildchain/targets/local_image.py
jbertran: Yup, same return convention as the Docker API PR. Done.

slaperche-scality (dismissed):

jbertran:
@bert-e approve

slaperche-scality (approved):

jbertran:
@bert-e approve

jbertran:
@bert-e reset

PR #928, opened 2019-04-03
Fix pytest filtering

### LINKED ISSUE #879: Do not use `bash` to run `pytest`
In the `tests` environment for `tox`, we invoke `pytest` through `bash -c "..."`.

This allows us to pass an indirected file for `--ssh-config`, but it prevents the usage of quotes in positional arguments passed to tox: `tox -e tests -- -k logs` will result in `pytest -k logs`, but `tox -e tests -- -k "not logs"` will result in `pytest -k not logs`, which in turn breaks with unexpected behaviour.

We should find some way to invoke pytest directly to preserve these functionalities (which are already missing, in particular it would be nice to be able to `tox -e tests -- -k "not bootstrap"`).

ramanduh:
Test call was made with bash command instead of pytest.
The reason is the need of pre-compute ssh-config from vagrant.
Now we pre-compute the ssh-config with tox commands_pre option
and store it in a file before running the tests

Closes #879

Test: run the following command:
```
tox -r -e tests -- --collect-only -k "not logs"
```
without the fix, all tests are deselected, with the fix, only one test is deselected

PS: this pr introduce also a minimum version for tox

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,6 @@
 [tox]
 skipsdist = true
+minversion = 3.4.0
 
 [testenv]
 whitelist_externals = bash
@@ -47,13 +48,16 @@ passenv =
     SSH_CONFIG_FILE
     TEST_HOSTS_LIST
     ISO_MOUNTPOINT
+setenv =
+    VAGRANT_CWD={toxinidir}
+commands_pre =
+    - bash -c "vagrant ssh-config bootstrap > {envdir}/vagrant_ssh_config"
 commands =
-    bash -c "pytest \
-             --ssh-config={env:SSH_CONFIG_FILE:<( \
-                 VAGRANT_CWD={toxinidir} vagrant ssh-config bootstrap)} \
-             --hosts={env:TEST_HOSTS_LIST:bootstrap} \
-             --iso-root={env:ISO_MOUNTPOINT:/vagrant/_build/root} \
-             {posargs} tests"
+    pytest \
+         --ssh-config={env:SSH_CONFIG_FILE:{envdir}/vagrant_ssh_config} \
+         --hosts={env:TEST_HOSTS_LIST:bootstrap} \
+         --iso-root={env:ISO_MOUNTPOINT:/vagrant/_build/root} \
+         {posargs} tests
 
 [testenv:tests-local]
 description =


tox.ini
```diff
@@ -47,13 +48,16 @@ passenv =
     SSH_CONFIG_FILE
     TEST_HOSTS_LIST
     ISO_MOUNTPOINT
+setenv =
+    VAGRANT_CWD={toxinidir}
+commands_pre =
```
gdemonet, line 53: Are commands specified in `commands_pre` run for each invocation of this environment? If not (e.g. only on env creation), this might become problematic.

gdemonet (dismissed):
Just a question about whether this `commands_pre` is executed every time or just on env creation. You may resolve if there is no problem (didn't test yet, and the docs are not explicit enough).

tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
 skipsdist = true
+minversion = 3.4.0
```
slaperche-scality, line 3: I would say `3.6` (for the buildchain)

slaperche-scality (changes requested):

tox.ini
gdemonet: It's the tox version ;P

slaperche-scality: Ho, then forget my comment.

slaperche-scality (dismissed):

tox.ini
gdemonet: So it seems they are. However, we run it even if passing an `SSH_CONFIG_FILE` env variable - which we do in the CI, where there is no Vagrant. As discussed, prefixing the command with a dash `-` should ignore the return code.

_Or_, you could test the environment variable in `commands_pre`, e.g.:
```
whitelist_externals =
    {[testenv]whitelist_externals}
    vagrant
...
commands_pre =
    [[ -z "$SSH_CONFIG_FILE" ]] && vagrant ssh-config bootstrap > {envdir}/vagrant_ssh_config
```

gdemonet (changes requested):
We need to either ignore failure in the CI with the `-` prefix, or only run if no `SSH_CONFIG_FILE` was provide.

tox.ini
ramanduh: yes it is run each invocation of the env

gdemonet (dismissed):

ramanduh:
@bert-e reset

ramanduh:
@bert-e approve

ramanduh:
@bert-e reset

slaperche-scality (approved):

Issue #735, opened 2019-03-08
Orchestrate runner to boostrap worker node


sayf-eddine-scality:
This runner, executed from the salt master, should, on a fresh new node with only salt-minion installed and linked to the master:
* setup containerd
* prepare certs
* setup kubelet (disabling self-registration is important)

In a workload plane node, there is:
* no master label
* no role taint
* no manifests in /etc/kubernetes/manifests
* no files used by those manifests only

Success: the orchestrate can transform a salt-minion-ready node into a node able to launch containers and part of the cluster (that means having kubelet talking to the apiserver)

cmonfort:
Must be done after #626 has landed.

PR #930, opened 2019-04-03
Kubeproxy: Set hostname override dynamically

### LINKED ISSUE #912: kube-proxy is not set with the correct hostname overide when expanding worker plane
In kube-proxy configmap, the `hostnameOverride` is always set to bootstrap no matter of the node name. When performing the expansion, this value should be `node1` on node `node1`.

To do so, we can set the value to `hostnameOverride: ""` and override it from the Command:
```      
/usr/local/bin/kube-proxy
 --config=/var/lib/kube-proxy/config.conf
 --hostname-override=$(NODE_NAME)
```
cf kubeadm

ramanduh:
The value of hostnameOverride was set in the kube-proxy config
to the hostname of the bootstrap node (same for all nodes).
Now, we pass it to kube-proxy command line and we use the node name
as value

Closes #912, Relate #735

Tests performed: run `pgrep -a kube-proxy` after bootstrap and expansion
Result on bootstrap:
```
12587 /usr/local/bin/kube-proxy --config=/var/lib/kube-proxy/config.conf --hostname-override=bootstrap
```

Result on node1:
```
7308 /usr/local/bin/kube-proxy --config=/var/lib/kube-proxy/config.conf --hostname-override=node1
```
=> the hostname override is passed in the arg and different on the node1 and bootstrap

Before, it was set in `/var/lib/kube-proxy/config.conf` with always the value `bootstrap`

### DIFF
--- a/salt/metalk8s/kubeadm/init/addons/kube-proxy.sls
+++ b/salt/metalk8s/kubeadm/init/addons/kube-proxy.sls
@@ -8,7 +8,6 @@
 
 {#- TODO: Not always use local machine as apiserver #}
 {%- set apiserver = 'https://' ~ salt['network.ip_addrs'](cidr=networks.control_plane)[0] ~ ':6443' %}
-{% set host_name = salt.network.get_hostname() %}
 
 Deploy kube-proxy (ServiceAccount):
   kubernetes.serviceaccount_present:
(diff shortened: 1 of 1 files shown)


slaperche-scality (dismissed):

ramanduh:
@bert-e approve

ramanduh:
@bert-e status

ramanduh:
@bert-e reset

slaperche-scality (approved):

PR #936, opened 2019-04-04
salt: ensure that Salt IDs match k8s IDs

slaperche-scality:
How to test:
- follow the [procedure](https://docs.google.com/document/d/1GAMbSTIFhCypBUITg_SunCOeoMdbEfBK_XZZJIw3Wdw/edit) up to the step 6 (deploying a salt minion)
- connect to the new node (`vagrant ssh node1`)
- check the content of `/etc/salt/minion.d/99-metalk8s.conf`: it should contains an `id` that match the node name

(You can also check the value on the bootstrap node).

### DIFF
--- a/salt/metalk8s/salt/minion/configured.sls
+++ b/salt/metalk8s/salt/minion/configured.sls
@@ -16,5 +16,6 @@ Configure salt minion:
     - backup: false
     - defaults:
       master_hostname: {{ defaults.salt.master.host }}
+      minion_id: {{ grains.id }}
     - watch_in:
       - cmd: Restart salt-minion
--- a/salt/metalk8s/salt/minion/files/minion_99-metalk8s.conf.j2
+++ b/salt/metalk8s/salt/minion/files/minion_99-metalk8s.conf.j2
@@ -1,4 +1,5 @@
 master: {{ master_hostname }}
+id: {{ minion_id }}
 
 retry_dns_count: 5
 


NicolasT:
So, `/etc/salt/minion_id` is not used, though it exists? Or is it not generated when `id` is set in the Minion config?

TeddyAndrieux:
> So, `/etc/salt/minion_id` is not used, though it exists? Or is it not generated when `id` is set in the Minion config?

Btw `id` in salt_minion config win over `/etc/salt/minion_id` so I think it's better to put it in the minion config

slaperche-scality:
> So, `/etc/salt/minion_id` is not used, though it exists? Or is it not generated when `id` is set in the Minion config?

I think it's generated from `id` (and used as cache), if I understand this sentence correctly:

> When the minion is started, it will generate an id value, unless it has been generated on a previous run and cached (in /etc/salt/minion_id by default)

See https://docs.saltstack.com/en/latest/topics/tutorials/walkthrough.html

And on the same page:

> Overriding the id
>
> The minion id can be manually specified using the id parameter in the minion config file. If this configuration value is specified, it will override all other sources for the id.

So, LGTM

sayf-eddine-scality (dismissed):

TeddyAndrieux (dismissed):

slaperche-scality:
@bert-e approve

NicolasT (dismissed):

TeddyAndrieux (dismissed):

TeddyAndrieux:
@bert-e reset

slaperche-scality:
@bert-e reset

TeddyAndrieux (approved):

Issue #917, opened 2019-04-02
UI - Node expansion - remove


carlito-scality:
- [x] UI design
- [x] Modify the "Table" Component to handle remove row in the core-ui
- [ ] Implement "Node Remove" (Call api....)

carlito-scality:
On hold until we can delete the node from the cluster

Issue #914, opened 2019-04-02
UI - Node expansion - Creation Form 


carlito-scality:
https://docs.google.com/document/d/1n6R9NhE5u5JdYLcFzIt2U675Sdk0Avk-SdeWK6RDXqI/edit

- Add 'Create New Node' button
- Implement creation Form using formix

PR #959, opened 2019-04-08
Nodes: add remove node button

carlito-scality:
https://github.com/scality/metalk8s/issues/917

Screenshot 2019-04-08 at 17 46 41

### DIFF
--- a/ui/src/containers/NodeList.js
+++ b/ui/src/containers/NodeList.js
@@ -86,7 +86,7 @@ class NodeList extends React.Component {
       return {
         ...node,
         actions:
-          node.status === 'Unknown'
+          !node.status || node.status === 'Unknown'
             ? [
                 {
                   label: this.props.intl.messages.remove,
--- a/ui/src/containers/NodeList.js
+++ b/ui/src/containers/NodeList.js
@@ -5,7 +5,7 @@ import memoizeOne from 'memoize-one';
 import { sortBy as sortByArray } from 'lodash';
 import { injectIntl, FormattedDate, FormattedTime } from 'react-intl';
 
-import { fetchNodesAction } from '../ducks/app/nodes';
+import { fetchNodesAction, deleteNodeAction } from '../ducks/app/nodes';
 
 class NodeList extends React.Component {
   constructor(props) {
@@ -82,9 +82,24 @@ class NodeList extends React.Component {
       this.state.sortDirection
     );
 
+    const nodesWithActions = nodesSortedList.map(node => {
+      return {
+        ...node,
+        actions:
+          node.status === 'Unknown'
+            ? [
+                {
+                  label: this.props.intl.messages.remove,
+                  onClick: this.props.deleteNode
+                }
+              ]
+            : null
+      };
+    });
+
     return (
       <Table
-        list={nodesSortedList}
+        list={nodesWithActions}
         columns={this.state.columns}
         disableHeader={false}
         headerHeight={40}
@@ -105,7 +120,8 @@ function mapStateToProps(state) {
 
 const mapDispatchToProps = dispatch => {
   return {
-    fetchNodes: () => dispatch(fetchNodesAction())
+    fetchNodes: () => dispatch(fetchNodesAction()),
+    deleteNode: node => dispatch(deleteNodeAction(node))
   };
 };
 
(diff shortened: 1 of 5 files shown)


carlito-scality:
On hold. Waiting for https://github.com/scality/metalk8s/issues/914 for integration

NicolasT:
Simply deleting a node from the cluster using the K8s API is not sufficient to 'remove a node from a cluster'.

Not sure why node removal is being added to the UI already, whilst we don't have a design (in the automation) for node removal yet...

charlesprost:
> Not sure why node removal is being added to the UI already, whilst we don't have a design (in the automation) for node removal yet...

Because if we can add a node, then edit/remove it should be feasible too. Or user experience will just be painful, especially when testing.

Of course it should not be available if the node is already part of the cluster because then it's not just about removing it from the apiserver.

NicolasT:
> Of course it should not be available if the node is already part of the cluster because then it's not just about removing it from the apiserver.

Now that's a strange user experience. You'd be able to 'delete' a node from the UI, at first it won't be in the node table indeed, and then only a minute later or so it'll be in the list of nodes again, without you doing anything.

charlesprost:
> Now that's a strange user experience. You'd be able to 'delete' a node from the UI, at first it won't be in the node table indeed, and then only a minute later or so it'll be in the list of nodes again, without you doing anything.

It seems you are describing exactly the scenario I want to avoid by blocking node deletion once deployed.

My idea is: if I'm testing creation/edition/deletion while not having the node deployed, it should be fine, because it's only declaration on the apiserver.

Maybe something is wrong with this idea though. What do you think?

NicolasT:
> > Now that's a strange user experience. You'd be able to 'delete' a node from the UI, at first it won't be in the node table indeed, and then only a minute later or so it'll be in the list of nodes again, without you doing anything.
> 
> It seems you are describing exactly the scenario I want to avoid by blocking node deletion once deployed.

That's adding 'functionality' which we don't want on the long term ;-)

> My idea is: if I'm testing creation/edition/deletion while not having the node deployed, it should be fine, because it's only declaration on the apiserver.

`kubectl delete node ...`

> Maybe something is wrong with this idea though. What do you think?

I think we should

- not expose functionality in the UI which doesn't do what one may intuitively think it does, or which only supports a part of what one may intuitively expect from such feature
- not expose functionality in the UI 'for testing purposes' (which in this case then requires non-trivial state management)

Just my .02.

charlesprost:
I'm not a big fan of having the addition in the UI, and the deletion by the command line. Seems not natural.

I totally agree that the state management may be difficult. But since we can delete a deployed node using the CLI and that will break probably some things, then I think having the same button in the UI saying `for test purposes only` should be fine.

It's just my opinion, nothing too strong in here.

carlito-scality:
I agree with @nootal the creation and deletion should be performed in the same way either by UI or by CLI. It is for the sake of user experience.

By the way, we could rename "Create a new node" to "Register a new node to Api Server"  and "Remove" to "Unregister" to avoid to "expose functionality in the UI which doesn't do what one may intuitively think it does"
What do you think?

NicolasT:
> I'm not a big fan of having the addition in the UI, and the deletion by the command line. Seems not natural.

Agree. My main point is: there *is* no deletion right now, not in UI, nor in CLI, for a somewhat reasonable definition of 'deletion'. If 'deletion' means: from the API, then not exposing this (very crippled) 'deletion' in the UI seems counter-intuitive to me.

> I totally agree that the state management may be difficult. But since we can delete a deployed node using the CLI and that will break probably some things, then I think having the same button in the UI saying `for test purposes only` should be fine.

The 'for test purposes only' functionality in the UI can easily be replicated using CLI, and doesn't require any 'flags' in the UI.

> It's just my opinion, nothing too strong in here.

The more I think about it, the stronger my I feel about it :-P

NicolasT:
> I agree with @nootal the creation and deletion should be performed in the same way either by UI or by CLI. It is for the sake of user experience.

Of course. At some point in time, when *actual deletion* is implemented. Right now it is not. What 'a click on the button' would do given this code is *not* properly deleting a node from the cluster.

> By the way, we could rename "Create a new node" to "Register a new node to Api Server" and "Remove" to "Unregister" to avoid to "expose functionality in the UI which doesn't do what one may intuitively think it does"
> What do you think?

Most certainly not. An end-user doesn't know what "the API server" is, that there's some distinction between "the API server" and "real world", and whatnot. 'Registering' and 'Unregistering' nodes are internal implementation details as to how our product works, not something to expose to a high-level end-user in the UI.

carlito-scality:
I agree with you but I think this PR is a part of the feature "remove node". It is stable and it works. We can merge it while waiting for the real "deletion". Once it is implemented, on the front-side we have just call the right endpoint to remove the node from the cluster and that's it.

carlito-scality:
draft

Issue #964, opened 2019-04-09
Remove unneeded mount for Docker-in-Docker sidecar


gdemonet:
In the CI, we use a Pod to build the ISO. This Pod has a Buildbot worker container, and a Docker-in-Docker sidecar so that the worker can use `docker`.

For this sidecar, we mount `/var/lib/docker` as an `emptyDir` volume, which is not needed (we pass `DOCKER_HOST=localhost` as an environment variable to the worker).

gdemonet:
Actually, this was needed (at least, made on purpose). Using the container overlay for intensive IO is not recommended by Docker: https://docs.docker.com/storage/volumes/

gdemonet:
Closed (see #965 and #976)

PR #965, opened 2019-04-09
Remove unneeded mount for Docker-in-Docker sidecar

gdemonet:
There is no need to mount /var/lib/docker as a volume, the Docker socket
is exposed through the network, using the DOCKER_HOST environment
variable.

Fixes GH-964.

### DIFF
--- a/eve/workers/pod-builder/pod.yaml
+++ b/eve/workers/pod-builder/pod.yaml
@@ -34,12 +34,8 @@ spec:
       securityContext:
         privileged: true
       volumeMounts:
-        - name: docker-storage
-          mountPath: /var/lib/docker
         - name: worker-workspace
           mountPath: /home/eve/workspace
   volumes:
-    - name: docker-storage
-      emptyDir: {}
     - name: worker-workspace
       emptyDir: {}


sayf-eddine-scality (approved):

gdemonet:
@bert-e approve

eve/workers/pod-builder/pod.yaml
```diff
@@ -34,12 +34,8 @@ spec:
       securityContext:
         privileged: true
       volumeMounts:
-        - name: docker-storage
-          mountPath: /var/lib/docker
         - name: worker-workspace
           mountPath: /home/eve/workspace
   volumes:
-    - name: docker-storage
```
tcarmet, line 42: This mount is indeed needed, otherwise you write all the images data in the docker overlay and not in a proper volume and loose some performance. NOT RECOMENDED

tcarmet (changes requested):

eve/workers/pod-builder/pod.yaml
tcarmet: ah, I'm too late :(

Zempashi: https://docs.docker.com/storage/volumes/

gdemonet: Oops, will revert in another PR (and add a comment about it) and explain why in the issue. Good thing you caught it, thanks!

Issue #800, opened 2019-03-20
Do not overwrite existing Salt environments when configuring Salt master


gdemonet:
Currently, the `metalk8s.salt.master.configured` state manages a single master configuration file, located under `/etc/salt/master.d/99-metalk8s.conf`, and containing the following declarations:

```
file_roots:
  {{ saltenv }}:
    - {{ iso_root_path }}/salt
pillar_roots:
  {{ saltenv }}:
    - {{ iso_root_path }}/pillar

[...]
```

However, in some scenario where we need to upgrade the Salt master, applying `metalk8s.salt.master.configured` with a new environment will overwrite the previous one, which is not what we want.

A quick solution could be to define a common configuration file which the latest Salt environment could overwrite, and a specific configuration file per environment (could include the MetalK8s version in its name) to define the `file_` and `pillar_roots`.

slaperche-scality:
So, IIUC the idea is to split `/etc/salt/master.d/99-metalk8s.conf` into two files:
- one, `/etc/salt/master.d/99-metalk8s.conf`, that would contains the version-agnostic settings (such as `ext_pillar`, `roster_defaults`, …)
- one, `/etc/salt/master.d/99-metalk8s-$SHORT_VERSION.conf`, that would contains version-dependant settings (`file_roots` and `pillar_roots`).

To do so, we need to:
- remove the version-specific settings from `salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2`
- add a new template for the version-specific settings
- generate a file from this template in `salt/metalk8s/salt/master/configured.sls`

slaperche-scality:
Ok, after some tests it seems like config files are not merged as we expected, thus this approach doesn't work.

Let's follow @TeddyAndrieux's suggestion and:
- keep generating a single file: `/etc/salt/master.d/99-metalk8s.conf`
- generate it by using `file.serialize` (with `merge_if_exists` set to `true`) instead of `file.managed`).
- get rid of the now unused `salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2`

PR #966, opened 2019-04-09
salt: change salt-master config generation

slaperche-scality:
In order to support several `saltenv` at the same time (e.g. during an
upgrade), we now generate the configuration file with `file.serialize`
and `merge_if_exists` set to `True`.
With that change, when you re-run the state with a new `saltenv`,
instead of overwriting the existing one it will add a new entry in the
dict.

## How to test

- Spawn a bootstrap node: `./doit.sh vagrantup`
- Run `salt-call state.apply metalk8s.salt.master.configured saltenv=metalk8s-2.0`: No changes
- Edit `/etc/salt/master.d/99-metalk8s.conf`, replacing `2.0` by `2.1` (to simulate a new `saltenv`).
- Re-run `salt-call state.apply metalk8s.salt.master.configured saltenv=metalk8s-2.0`: the file is modified, we now have two entries:
  - the manual one (2.1) is still here, not overwritten
  - the current one (2.0) is added

Closes: #800


NicolasT:
IMO we should have two files in `master.d`:

- One which contains 'static' configuration that may get overwritten, but not merged, because then it's hard (to say at least) to somewhat deterministically know what's in the file post deployment
- One which contains `top_roots` and `file_roots`, and which *is* updated by 'merging' values upon updates

slaperche-scality:
> IMO we should have two files in `master.d`:
> 
>     * One which contains 'static' configuration that may get overwritten, but not merged, because then it's hard (to say at least) to somewhat deterministically know what's in the file post deployment
> 
>     * One which contains `top_roots` and `file_roots`, and which _is_ updated by 'merging' values upon updates

Done

slaperche-scality:
@bert-e reset

gdemonet (approved):
LGTM

TeddyAndrieux (approved):

slaperche-scality:
@bert-e approve

Issue #880, opened 2019-03-28
Use Vagrant snapshot features in DoIt


gdemonet:
Currently, developers trying to run a bootstrap deployment in a fresh Vagrant environment can either do so from scratch every time (i.e. `vagrant destroy -f` before running `./doit.sh vagrantup`), or using snapshots (i.e. `vagrant snapshot restore bootstrap my_snapshot`).

In any case, neither of these two approaches are used in the current buildchain, and developers still require manual action.

We advocate that:
- [ ] `doit vagrantup` depends on a snapshot being created for the un-provisioned VMs
- [ ] `doit clean vagrantup` restores said snapshots

slaperche-scality:
For the `doit clean vagrantup` part it's clearly doable with a custom clean actions, see [here](http://pydoit.org/cmd_other.html#clean).

PR #949, opened 2019-04-05
doc: introduce a good practice guide for Python

slaperche-scality:
We recently "standardized" our review process and we mentioned:

> code is done and follow coding standards (enforced by the CI or documented as best practices)

This PR tries to document such best practices for Python. It's based on a draft I started to write for the supervisor team (but I never got to finish it) and can be used as a starting point, so feel free to debate and enrich it :)

Note that this document is not an exhaustive coding style per se, it only contains rules that cannot be enforced by automated tools.

It would be really cool to also have this kind of "guide" for Salt, JavaScript, …

PR #968, opened 2019-04-09
Use vagrant snapshots to shorten dev env launching.

cmonfort:
Use vagrant snapshots to shorten dev env launching.
    
Automatically makes a snapshot after the first `doit vagrant_up` is made
(VM created and Virtualbox Guest Additions installed). Then you can use:
* `doit vagrant_restore` to restore snapshot.
* `doit vagrant_restore vagrant_up` to restore snapshot and re-provision VM.
* `doit clean vagrant_up` to destroy all vagrant environment (aka: `vagrant destroy --force`) and clean up project.
    
Note: `vagrantup` task renamed `vagrant_up` for consistency.
    
Fixes: #880

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -43,6 +43,7 @@ Available options are:
   repository)
 - `VAGRANT_PROVIDER`: type of machine to spawn with Vagrant
 - `VAGRANT_UP_ARGS`: command line arguments to pass to `vagrant up`
+- `VAGRANT_SNAPSHOT_NAME`: name of auto generated Vagrant snapshot
 - `DOCKER_BIN`: Docker binary (name or path to the binary)
 - `GIT_BIN`: Git binary (name or path to the binary)
 - `MKISOFS_BIN`: mkisofs binary (name or path to the binary)
--- a/buildchain/buildchain/vagrant.py
+++ b/buildchain/buildchain/vagrant.py
@@ -19,7 +19,6 @@ from buildchain import constants
 from buildchain import types
 from buildchain import utils
 
-
 def task__vagrantkey() -> types.TaskDict:
     """Generate a SSH key pair in the .vagrant folder."""
     def mkdir_dot_vagrant() -> None:
@@ -40,16 +39,57 @@ def task__vagrantkey() -> types.TaskDict:
         'uptodate': [True],
     }
 
-def task_vagrantup() -> types.TaskDict:
-    """Run `vagrant up` to (re-)provision a development environment."""
-    vagrant = [config.VAGRANT, 'up']
-    vagrant.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+def task__vagrant_up_noprov() -> types.TaskDict:
+    """Run `vagrant up` without provisioning a development environment."""
+    cmd = [config.VAGRANT, 'up']
+    cmd.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+    cmd.append('--no-provision')
+
     return {
-        'actions': [doit.tools.LongRunning(' '.join(vagrant))],
-        'file_dep': [constants.VAGRANT_SSH_KEY_PAIR],
-        'task_dep': ['populate_iso'],
+        'actions': [doit.tools.LongRunning(' '.join(cmd))],
+        'file_dep': [],
+        'task_dep': [],
         'uptodate': [False],
     }
 
+def task__vagrant_snapshot() -> types.TaskDict:
+    """Snapshot development environment."""
+    vagranthalt = [config.VAGRANT, 'halt', 'bootstrap']
+    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap',
+        config.VAGRANT_SNAPSHOT_NAME]
+
+    return {
+        'actions': [vagranthalt, vagrantsnap],
+        'task_dep': ['_vagrant_up_noprov'],
+        'uptodate': [doit.tools.run_once],
+        'verbosity': 2
+    }
+
+def task_vagrant_restore() -> types.TaskDict:
+    """Restore development environment snapshot."""
+    cmd = [config.VAGRANT, 'snapshot', 'restore', 'bootstrap',
+        config.VAGRANT_SNAPSHOT_NAME]
+
+    return {
+        'actions': [cmd],
+        'uptodate': [False],
+        'verbosity': 2
+    }
+
+def task_vagrant_up() -> types.TaskDict:
+    """Run `vagrant up` to (re-)provision a development environment."""
+    vagrantup = [config.VAGRANT, 'up']
+    vagrantup.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+
+    vagrantdestroy = [config.VAGRANT, 'destroy', '--force']
+
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantup))],
+        'file_dep': [constants.VAGRANT_SSH_KEY_PAIR],
+        'task_dep': ['populate_iso', '_vagrant_snapshot'],
+        'uptodate': [False],
+        'clean': [vagrantdestroy],
+        'verbosity': 2
+    }
 
 __all__ = utils.export_only_tasks(__name__)
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -37,6 +37,8 @@ _DEFAULT_VAGRANT_UP_ARGS : str = ' '.join((
     '--provider', VAGRANT_PROVIDER
 ))
 
+VAGRANT_SNAPSHOT_NAME: str = 'bootstrap.autosnapshot'
+
 VAGRANT_UP_ARGS : Tuple[str, ...] = tuple(shlex.split(
     os.getenv('VAGRANT_UP_ARGS', _DEFAULT_VAGRANT_UP_ARGS)
 ))


buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
     }
 
+def task_vagrantupnoprov() -> types.TaskDict:
+    """Run `vagrant up` without provisioning a development environment."""
+    vagrantupnoprov = [config.VAGRANT, 'up']
+    vagrantupnoprov.extend(map(shlex.quote,
+        config.VAGRANT_UP_ARGS + tuple(('--no-provision',))))
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantupnoprov))],
+        'file_dep': [constants.VAGRANT_SSH_KEY_PAIR],
```
charlesprost: Since we don't provision, I don't think we need a dep on this.

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantrestore))],
+        'file_dep': [],
+        'task_dep': [],
+        'uptodate': [False]
+    }
+
 def task_vagrantup() -> types.TaskDict:
     """Run `vagrant up` to (re-)provision a development environment."""
     vagrant = [config.VAGRANT, 'up']
```
charlesprost: rename this variable to `vagrantup` maybe

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantupnoprov))],
+        'file_dep': [constants.VAGRANT_SSH_KEY_PAIR],
+        'task_dep': [],
+        'uptodate': [False],
+    }
+
+def task_vagrantsnapshot() -> types.TaskDict:
+    """Snapshot development environment."""
+    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
```
charlesprost: We may want to save the snapshot while the VM is down. Then we'll have no conflict about port assignment.

BUILDING.md
```diff
@@ -43,6 +43,7 @@ Available options are:
   repository)
 - `VAGRANT_PROVIDER`: type of machine to spawn with Vagrant
 - `VAGRANT_UP_ARGS`: command line arguments to pass to `vagrant up`
```
slaperche-scality, line 45: Commit title is too long (you can see that Github as wrapped it, [see also](https://github.com/scality/metalk8s/blob/development/2.0/CONTRIBUTING.md#committing-guidelines)), could be "use vagrant snapshots to speed up dev env launching" to fit.

buildchain/buildchain/vagrant.py
```diff
@@ -19,6 +19,8 @@
 from buildchain import types
 from buildchain import utils
 
+from doit.tools import run_once
```
slaperche-scality: tl;dr: remove this import and simply use `doit.tools.run_once`

`doit` is an external deps so it should be included after stdlib and before local files so this import should be under the already existing `import doit`.

Moreover `from X import Y` shouldn't be used¹, prefer `import doit.tools as tools` and then use `tools.run_once`.
And in fact, I would just remove this import: to be consistent with `doit.tools.LongRunning`, simply use `doit.tools.run_once`.

----

¹: See [guidelines](https://github.com/scality/metalk8s/blob/development/2.0/docs/best_practices/python.md#avoid-from-module_foo-import-symbol_bar) &(yes I know, wasn't applied elsewhere as [pointed by Guillaume](https://github.com/scality/metalk8s/pull/949#discussion_r273018483), but will be done when we add unit tests, meanwhile let's not adding more).

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+        'task_dep': [],
+        'uptodate': [False]
+    }
+
 def task_vagrantup() -> types.TaskDict:
     """Run `vagrant up` to (re-)provision a development environment."""
     vagrant = [config.VAGRANT, 'up']
     vagrant.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+
+    vagrantdestroy = [config.VAGRANT, 'destroy', '--force']
```
slaperche-scality, line 84: Since this will be used as a string, you should `map(shlex.quote` as well, to be safe)

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantsnap))],
+        'file_dep': [],
+        'task_dep': ['vagrantupnoprov'],
+        'uptodate': [run_once]
+    }
+
+def task_vagrantreset() -> types.TaskDict:
+    """Restore development environment snapshot."""
+    vagrantrestore = [config.VAGRANT, 'snapshot', 'restore', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
```
slaperche-scality: Line too long:

```python
     vagrantrestore = [config.VAGRANT, 'snapshot', 'restore', 'bootstrap',
                       config.VAGRANT_SNAPSHOT_NAME]
```

Also the variable could be simply called `cmd` since you only have one command in this function.

Same remarks apply to `vagrantsnap` and `vagrantupnoprov` above.

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantrestore))],
```
slaperche-scality: Is `LongRunning` needed here? I think it will work without it (if yes, remove it).

Same comment for the other places where you used `LongRunning`.

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
+
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantsnap))],
+        'file_dep': [],
+        'task_dep': ['vagrantupnoprov'],
+        'uptodate': [run_once]
+    }
+
+def task_vagrantreset() -> types.TaskDict:
```
slaperche-scality: When we call `vagrant up` the task is called `vagrantup`: I would keep this logic and call this task `vagrantrestore` (or `vagrant_restore` for the sake of legibility).

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
         'uptodate': [True],
     }
 
+def task_vagrantupnoprov() -> types.TaskDict:
+    """Run `vagrant up` without provisioning a development environment."""
+    vagrantupnoprov = [config.VAGRANT, 'up']
+    vagrantupnoprov.extend(map(shlex.quote,
+        config.VAGRANT_UP_ARGS + tuple(('--no-provision',))))
```
slaperche-scality: Why not `+ ['--no-provision']`?

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
         'uptodate': [True],
     }
 
+def task_vagrantupnoprov() -> types.TaskDict:
```
slaperche-scality: Does the user wants to use this?

If not and the only purpose is to use it as `task_dep`, this should be a private task (`task__vagrantupnoprov`) to avoid polluting the users with useless tasks when they run `doit list`.

Same remark for `task_vagrantsnapshot`.

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +42,53 @@ def mkdir_dot_vagrant() -> None:
...
+
+def task_vagrantsnapshot() -> types.TaskDict:
+    """Snapshot development environment."""
+    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
+
+    return {
+        'actions': [doit.tools.LongRunning(' '.join(vagrantsnap))],
+        'file_dep': [],
+        'task_dep': ['vagrantupnoprov'],
+        'uptodate': [run_once]
```
slaperche-scality: Maybe good for now, but could be better to check if the snapshot exist in vagrant (by checking the output of `vagrant snapshot list`).

You can create a debt ticket for this one, no need to tackle it here.

slaperche-scality (changes requested):

buildchain/buildchain/vagrant.py
cmonfort: It works, but we lose vagrant output:
With `LongRunning`:
```
$ ./doit.sh vagrantreset
.  vagrantreset
==> bootstrap: Forcing shutdown of VM...
==> bootstrap: Restoring the snapshot 'bootstrap.autosnapshot'...
==> bootstrap: Checking if box 'centos/7' version '1811.02' is up to date...
==> bootstrap: Resuming suspended VM...
==> bootstrap: Booting VM...
==> bootstrap: Waiting for machine to boot. This may take a few minutes...
    bootstrap: SSH address: 127.0.0.1:2222
    bootstrap: SSH username: vagrant
    bootstrap: SSH auth method: private key
==> bootstrap: Machine booted and ready!
==> bootstrap: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> bootstrap: flag to force provisioning. Provisioners marked to run always will still run.
$
```
Without:
```
$ ./doit.sh vagrantreset
.  vagrantreset
$
```

slaperche-scality: If you want to see `stdout`, simply add `'verbosity': 2` in your dict (see [here](http://pydoit.org/tasks.html#verbosity)).

`LongRunning` does that indeed, but other things as well. So if you only want `stdout` then setting verbosity is better than using `LongRunning` :slightly_smiling_face:

buildchain/buildchain/vagrant.py
cmonfort: If we call this task `vagrant_restore`, then `vagrantup` should be `vagrant_up` (for consistency)...

buildchain/buildchain/vagrant.py
cmonfort: Because `config.VAGRANT_UP_ARGS` is a tuple, and we cannot concatenate list to tuples... (tried it in the 1st place ;)

buildchain/buildchain/vagrant.py
slaperche-scality: Yup, indeed.

You can either do it here or create a debt issue to do the rename in a dedicated PR.

buildchain/buildchain/vagrant.py
slaperche-scality: Oh, indeed!

Maybe it can be more readable (the single-element tuple with its trailing comma is always tricky…) by doing it in two steps:

```Python
    vagrantupnoprov.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
    vagrantupnoprov.append('--no-provision')
```

buildchain/buildchain/vagrant.py
cmonfort: Done

buildchain/buildchain/vagrant.py
cmonfort: #971

cmonfort:
@bert-e reset

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +39,60 @@ def mkdir_dot_vagrant() -> None:
         'uptodate': [True],
     }
 
-def task_vagrantup() -> types.TaskDict:
+def task__vagrant_up_noprov() -> types.TaskDict:
+    """Run `vagrant up` without provisioning a development environment."""
+    cmd = [config.VAGRANT, 'up']
+    cmd.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+    cmd.extend(['--no-provision'])
```
slaperche-scality: minor: by using `append` instead of `extend` you don't need to wrap the string into a list.

slaperche-scality (dismissed):
Only a minor comment that you can fix before approving yourself.

Other than that, it's good to go :slightly_smiling_face:

cmonfort:
@bert-e approve

cmonfort:
@bert-e reset

slaperche-scality (dismissed):

buildchain/buildchain/vagrant.py
```diff
@@ -40,16 +39,60 @@ def mkdir_dot_vagrant() -> None:
...
+def task__vagrant_snapshot() -> types.TaskDict:
+    """Snapshot development environment."""
+    vagranthalt = map(shlex.quote,
+        [config.VAGRANT, 'halt', 'bootstrap'])
+    vagrantsnap = map(shlex.quote,
+        [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap',
+         config.VAGRANT_SNAPSHOT_NAME])
+
+    return {
+        'actions': [' '.join(vagranthalt), ' '.join(vagrantsnap)],
```
slaperche-scality: Ha, since you're not using `LongRunning` anymore, you can directly use a list of string instead of a string (and thus no need for `shlex.quote` and `join`.

e.g:

```python
    vagranthalt = [config.VAGRANT, 'halt', 'bootstrap'])
    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap',
                   
    return {
        'actions': [vagranthalt, vagrantsnap],
```

Same remark for `task_vagrant_restore`.

cmonfort:
@bert-e reset

slaperche-scality (approved):

PR #970, opened 2019-04-09
salt: use grains to track node IPs

NicolasT:


NicolasT:
@bert-e wait

NicolasT:
This is a draft.

TODO:
- [x] Wherever we need a remote node IP, use the correct mine key
- [x] Wherever we calculate a local IP in SLSs right now, use the grain
- [x] Ensure the grains are set on all nodes we deploy

TeddyAndrieux:
Not really like the grains like that because that mean if you restart a minion you need to call the state again to set the grains.

We should have a custom grains in python in `_grains` but since we use information from the pillar we can't.

Instead of get the ip in the sls using network module from salt we should use the salt module `metalk8s.get_control_plane_ips` .

But what we can do it's get the ip in map.jinja it's not the best thing to do but it's better than what we have now IMO

NicolasT:
> Not really like the grains like that because that mean if you restart a minion you need to call the state again to set the grains.

That is incorrect, because `grains.present` calls into a `grains` procedure which sets the grain on the minion *in its configuration* (more exactly in `/etc/salt/grains`):

Ensure the grains are not present:
```
[root@bootstrap ~]# systemctl stop salt-minion
[root@bootstrap ~]# systemctl start salt-minion
[root@bootstrap ~]# cat /etc/salt/grains
[root@bootstrap ~]# salt-call grains.get metalk8s
local:
```

Call the state setting the grains:
```
[root@bootstrap ~]# salt-call state.sls metalk8s.node.grains saltenv=metalk8s-2.0
local:
----------
          ID: Set control_plane_ip grain
    Function: grains.present
        Name: metalk8s:control_plane_ip
      Result: True
     Comment: Set grain metalk8s:control_plane_ip to 172.21.254.7
     Started: 10:36:57.805679
    Duration: 9.067 ms
     Changes:
              ----------
              metalk8s:
                  ----------
                  control_plane_ip:
                      172.21.254.7
                  workload_plane_ip:
                      172.21.254.39
----------
          ID: Set workload_plane_ip grain
    Function: grains.present
        Name: metalk8s:workload_plane_ip
      Result: True
     Comment: Set grain metalk8s:workload_plane_ip to 172.21.254.39
     Started: 10:36:57.814935
    Duration: 4.462 ms
     Changes:
              ----------
              metalk8s:
                  ----------
                  control_plane_ip:
                      172.21.254.7
                  workload_plane_ip:
                      172.21.254.39

Summary for local
------------
Succeeded: 2 (changed=2)
Failed:    0
------------
Total states run:     2
Total run time:  13.529 ms
```

Ensure the grains are present:
```
[root@bootstrap ~]# salt-call grains.get metalk8s
local:
    ----------
    control_plane_ip:
        172.21.254.7
    workload_plane_ip:
        172.21.254.39
```

Ensure the grains are still present after minion restart:
```
[root@bootstrap ~]# systemctl stop salt-minion
[root@bootstrap ~]# systemctl start salt-minion
[root@bootstrap ~]# salt-call grains.get metalk8s
local:
    ----------
    control_plane_ip:
        172.21.254.7
    workload_plane_ip:
        172.21.254.39
```

Reason:
```
[root@bootstrap ~]# cat /etc/salt/grains
metalk8s:
  control_plane_ip: 172.21.254.7
  workload_plane_ip: 172.21.254.39
```

> We should have a custom grains in python in `_grains` but since we use information from the pillar we can't.

No need :)

> Instead of get the ip in the sls using network module from salt we should use the salt module `metalk8s.get_control_plane_ips` .

`metalk8s.get_control_plane_ips` should be removed since no need. Furthermore, we shouldn't have 'multiple' control-plane IPs. If a node is connected to the contol-plane network using different interfaces/IPs, we should 'pick one', because most services can bind to only one IP anyway (or `0.0.0.0`, of course).

> But what we can do it's get the ip in map.jinja it's not the best thing to do but it's better than what we have now IMO

Not sure what you mean, but also pretty sure it's not required. Storing addresses in a grain makes sense (it's what grains are for...). Then exposing these in the mine makes sense (it's what the mine is for). And no need for any custom modules basically doing the same thing.

We want these IPs in grains since we need them in various SLSs, i.e. every single time we calculate a control-plane (or workload-plane) IP in all SLSs by copying the very same 'Jinja code'.

TeddyAndrieux:
> That is incorrect, because `grains.present` calls into a `grains` procedure which sets the grain on the minion _in its configuration_ (more exactly in `/etc/salt/grains`):

Oh yes it's true, but the thing is that the IP is only "computed" one time. If the ip change one day then you need to advertise again but yes in our case it's not really important so ok.

> Not sure what you mean, but also pretty sure it's not required. Storing addresses in a grain makes sense (it's what grains are for...). Then exposing these in the mine makes sense (it's what the mine is for). And no need for any custom modules basically doing the same thing.

Yes totaly agree this Ip should be in a grains the only thing I not "really" like it's the fact that it's a salt state who push the grains which is not "automatic".

But in metalk8s case I think it's good enough to have it like that

NicolasT:
On Wed, 10 Apr 2019, 13:35 Teddy Andrieux, <notifications@github.com> wrote:

> That is incorrect, because grains.present calls into a grains procedure
> which sets the grain on the minion *in its configuration* (more exactly
> in /etc/salt/grains):
>
> Oh yes it's true, but the thing is that the IP is only "computed" one
> time. If the ip change one day then you need to advertise again but yes in
> our case it's not really important so ok.
>
If an IP of a node changes, you'll need to bring it into highstate again
anyway, because e.g. certificate SAN fields may no longer be 'correct'.


Not sure what you mean, but also pretty sure it's not required. Storing
> addresses in a grain makes sense (it's what grains are for...). Then
> exposing these in the mine makes sense (it's what the mine is for). And no
> need for any custom modules basically doing the same thing.
>
> Yes totaly agree this Ip should be in a grains the only thing I not
> "really" like it's the fact that it's a salt state who push the grains
> which is not "automatic".
>
> But in metalk8s case I think it's good enough to have it like that
>

We could over time have a grains module, sure. However, what I want to
achieve asap is:

- use grains for node-local semi-constant values
- fill mine values from grains instead of custom modules

Nicolas

NicolasT:
@bert-e reset

NicolasT:
Ready for review.

salt/metalk8s/orchestrate/deploy_etcd_on_new_node.sls
```diff
@@ -1,13 +1,13 @@
-{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ips') %}
+{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ip') %}
```
TeddyAndrieux: Not really important but you should remove the `s`
```suggestion
{%- set control_plane_ip = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ip') %}
```

NicolasT:
@bert-e reset

pillar/mine_functions.sls
```diff
@@ -1,7 +1,9 @@
 mine_functions:
-  control_plane_ips:
-    - mine_function: metalk8s.get_control_plane_ips
-  workload_plane_ips:
-    - mine_function: metalk8s.get_workload_plane_ips
+  control_plane_ip:
+    - mine_function: grains.get
+    - metalk8s:control_plane_ip
```
gdemonet: Let's try to keep named arguments as much as possible (YAML syntax could make one think there's a `metalk8s` key with a `control_plane_ip` value):

```suggestion
    - key: metalk8s:control_plane_ip
```

FYI, one can also change the delimiter for traversing the grains dict:

```
    - key: metalk8s/control_plane_ip
    - delimiter: "/"
```

salt/metalk8s/node/grains.sls
```diff
@@ -0,0 +1,9 @@
+Set control_plane_ip grain:
+  grains.present:
+    - name: metalk8s:control_plane_ip
+    - value: {{ salt['network.ip_addrs'](cidr=salt['pillar.get']('networks:control_plane'))[0] }}
+
+Set workload_plane_ip grain:
+  grains.present:
+    - name: metalk8s:workload_plane_ip
+    - value: {{ salt['network.ip_addrs'](cidr=salt['pillar.get']('networks:workload_plane'))[0] }}
```
gdemonet: What if we fail to find IPs in the configured CIDRs? Maybe the user misconfigured these networks... Maybe we could reuse the `if/else` structure here, so that when deploying a new node one quickly identifies which IP was not found.

(This "not finding IPs" issue occured quite frequently in the CI, since we have two private networks, the DHCP request we do during provisioning in Terraform may fail, and we may only have one of the two IPs available)

salt/metalk8s/orchestrate/control_plane.sls
```diff
@@ -1,8 +1,13 @@
-{%- set control_plane_ips = salt.saltutil.runner('mine.get', tgt='*', fun='control_plane_ips') %}
+{%- if pillar['bootstrap_id'] %}
+{%-   set control_plane_ips = salt.saltutil.runner('mine.get', tgt=pillar['bootstrap_id'], fun='control_plane_ip') %}
+{%- else %}
+{%-   set control_plane_ips = {} %}
+{%- endif %}
+
 {%- if pillar['bootstrap_id'] in control_plane_ips.keys() and control_plane_ips[pillar['bootstrap_id']] %}
```
gdemonet: Maybe a single ternary could be more readable than these conditions (not your code, but since you're changing things around it...):

```jinja
{%- set control_plane_ip = control_plane_ips.get(pillar['bootstrap_id']) | default('localhost', true) %}
```

gdemonet (changes requested):
Mostly code style comments, but there's some scenario (fail to find IPs to save into grains) that we need to cover so the other formulas don't explode with `KeyError`s without meaningful error messages.

pillar/mine_functions.sls
NicolasT: Done

salt/metalk8s/node/grains.sls
NicolasT: Done

pillar/mine_functions.sls
NicolasT: Sooooo.... no. I wanted to do this, since it's a valid remark. However, when I applied the change, lookups from the mine no longer work. No clue why.

salt/metalk8s/orchestrate/control_plane.sls
NicolasT: Done

NicolasT:
@bert-e reset

pillar/mine_functions.sls
NicolasT: Sooooooooooooo... yes. Thanks to a hint from @TeddyAndrieux to use the correct YAML encoding, this now works. Thanks! :+1:

NicolasT:
@bert-e reset

gdemonet (approved):

NicolasT:
@bert-e approve

PR #978, opened 2019-04-10
salt: implement true Kubernetes RBAC-based authn/authz

NicolasT:
Fixes: #751

### DIFF
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -81,9 +81,39 @@ def _auth_basic(kubeconfig, username, token):
 
 @_log_exceptions
 def _groups_basic(kubeconfig, username, token):
-    return [
-        'node_admin',
-    ]
+    kubeconfig.api_key = {
+        'authorization': token,
+    }
+    kubeconfig.api_key_prefix = {
+        'authorization': 'Basic',
+    }
+    kubeconfig.username = username
+    kubeconfig.password = None
+    kubeconfig.cert_file = None
+    kubeconfig.key_file = None
+
+    client = kubernetes.client.ApiClient(configuration=kubeconfig)
+
+    authz_api = kubernetes.client.AuthorizationV1Api(api_client=client)
+
+    groups = set()
+
+    result = authz_api.create_self_subject_access_review(
+        body=kubernetes.client.V1SelfSubjectAccessReview(
+            spec=kubernetes.client.V1SelfSubjectAccessReviewSpec(
+                resource_attributes=kubernetes.client.V1ResourceAttributes(
+                    resource='nodes',
+                    verb='*',
+                ),
+            ),
+        ),
+    )
+
+    if result.status.allowed:
+        groups.add('node-admins')
+
+    return list(groups)
+
 
 AUTH_HANDLERS['basic'] = {
     'auth': _auth_basic,
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -1,29 +1,120 @@
 import base64
 import logging
 
+MISSING_DEPS = []
+
+try:
+    import kubernetes.client
+    import kubernetes.config
+except ImportError:
+    MISSING_DEPS.append('kubernetes')
+
+try:
+    import requests
+except ImportError:
+    MISSING_DEPS.append('requests')
+
+
 log = logging.getLogger(__name__)
 
 __virtualname__ = 'kubernetes_rbac'
 
 def __virtual__():
-    return __virtualname__
+    if MISSING_DEPS:
+        return False, 'Missing Python dependencies: {}'.format(
+            ', '.join(MISSING_DEPS))
+    else:
+        return __virtualname__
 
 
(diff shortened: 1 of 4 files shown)


NicolasT:
~Currently based on the branch of #760 so not ready for merge as-is.~ Merged

NicolasT:
@bert-e wait

NicolasT:
~For some reason this doesn't sync with the changes that landed into the target branch in-between, so this looks much 'bigger' than it is. Hmpf...~ Fixed by force-pushing a bogus change (in the commit messages)

NicolasT:
@bert-e reset

salt/_auth/kubernetes_rbac.py
```diff
@@ -1,31 +1,152 @@
...
-    'basic': {
-        'auth': lambda username, token: all([
-            username == ADMIN_USERNAME,
-            token == ADMIN_TOKEN,
-        ]),
-        'groups': lambda _u, _t: [
-            'node_admin',
-        ],
-    },
+    if not HAS_KUBERNETES_DEPS:
```
gdemonet: Small suggestion to avoid masking one of the missing deps if the other one is missing as well:

```python
DEPS_CHECK = {}
try:
    import kubernetes.config
    DEPS_CHECK["python-kubernetes"] = True
except ImportError:
    DEPS_CHECK["python-kubernetes"] = False

try:
    import requests
    DEPS_CHECK["requests"] = True
except ImportError:
    DEPS_CHECK["requests"] = False

...

def __virtual__():
    if not all(DEPS_CHECK.values()):
        return False, 'Missing dependencies: {}'.format(
            ', '.join(dep_name for dep_name, check in DEPS_CHECK.items() if not check)
        )

    return __virtualname__
```

NicolasT: I'll apply a variation of this.

NicolasT: Somewhat simpler IMHO: https://github.com/scality/metalk8s/pull/978/files#diff-4746d4a429f108ddab90fc4a5c9fcb03R4

NicolasT:
@bert-e reset

NicolasT:
@bert-e approve

salt/_auth/kubernetes_rbac.py
```diff
@@ -1,31 +1,152 @@
...
+
+
+@_log_exceptions
+def _auth_basic(kubeconfig, username, token):
+    decoded = base64.decodestring(token)
+    if ':' not in decoded:
+        log.warning('Invalid Basic token format: missing ":"')
+        return False
+
+    (token_username, _) = decoded.split(':', 1)
```
gdemonet, line 51: minor: No need for the parentheses on the left-hand side, and even more explicit to use indexing:

```suggestion
    token_username = decoded.split(':', 1)[0]
```

salt/_auth/kubernetes_rbac.py
```diff
@@ -1,31 +1,152 @@
...
 
+@_log_exceptions
+def _load_kubeconfig(opts):
+    config = {
+        'kubeconfig': None,
+        'context': None,
+    }
+
+    for opt in opts['external_auth'][__virtualname__]:
+        if opt.startswith('^'):
```
gdemonet, line 132: Would you mind adding a comment for this "trick"? I was confused on first read

gdemonet (approved):

salt/_auth/kubernetes_rbac.py
NicolasT: Agree the parentheses are superfluous. I prefer `a, _ = foo.split(b, 1)` over `a = foo.split(b, 1)[0]`, because
- it makes it clearer there's only a single split about to happen
- if there's no split happening, the second version would simply return the whole string, whilst the first would raise an exception, which is exactly the intent here.

salt/_auth/kubernetes_rbac.py
NicolasT: It's (apparently, though admittedly not very fond of it) 'standard' in Salt `eauth` modules and how to configure them, e.g. the built-in `file` and `ldap` ones (which is where I got this idea).

salt/_auth/kubernetes_rbac.py
gdemonet: Agreed, but you have a test `if ':' not in decoded` just above, so the split will always happen. If you prefer the tuple-destructuring syntax, maybe one could use the EAFP pattern (mentioned in [our brand-new Best Practices document](https://github.com/scality/metalk8s/blob/development/2.0/docs/best_practices/python.md#dont-write-code-vulnerable-to-time-of-check-to-time-of-use)), something like:

```python
try:
    token_username, _ = decoded.split(':', 1)
except ValueError:
    log.warning('Invalid Basic token format: missing ":"')
    return False
```

salt/_auth/kubernetes_rbac.py
gdemonet: Oh, wasn't aware of it... Documentation of our internal modules should be improved in the future, so re-using/maintaining them is easier even for non-experts.

PR #984, opened 2019-04-11
Improvement/896 add pillar for bootstrap services ip

ramanduh:
The goal of this PR is to create a pillar named `endpoints` that will contain the IP of pods, running on the bootstrap.
To perform that, we create a service that map to these pods, and we get the pods IP from the endpoints object created with the service.

To test: run the bootstrap, an then query the pillar `metalk8s:endpoints`
```
# crictl exec -ti $(crictl ps | awk '/salt-master/{print $1}') salt "bootstrap" pillar.get metalk8s:endpoints saltenv=metalk8s-2.0
bootstrap:
    ----------
    package-repositories:
        ----------
        hostname:
            None
        ip:
            172.21.254.3
        node_name:
            bootstrap
        ports:
            ----------
            http:
                8080
    registry:
        ----------
        hostname:
            None
        ip:
            172.21.254.3
        node_name:
            bootstrap
        ports:
            ----------
            registry:
                5000
    salt-master:
        ----------
        hostname:
            None
        ip:
            172.21.254.3
        node_name:
            bootstrap
        ports:
            ----------
            api:
                4507
            publisher:
                4505
            requestserver:
                4506

```

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -209,6 +209,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/repo/init.sls'),
     Path('salt/metalk8s/repo/offline.sls'),
     Path('salt/metalk8s/repo/online.sls'),
+    Path('salt/metalk8s/repo/service_deployed.sls'),
 
     Path('salt/metalk8s/runc/init.sls'),
     Path('salt/metalk8s/runc/installed.sls'),
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -115,6 +115,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/bootstrap/preflight.sls'),
     Path('salt/metalk8s/bootstrap/registry.sls'),
     Path('salt/metalk8s/bootstrap/salt_master.sls'),
+    Path('salt/metalk8s/bootstrap/services.sls'),
 
     Path('salt/metalk8s/calico/configured.sls'),
     Path('salt/metalk8s/calico/deployed.sls'),
@@ -221,6 +222,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2'),
     Path('salt/metalk8s/salt/master/configured.sls'),
     Path('salt/metalk8s/salt/master/deployed.sls'),
+    Path('salt/metalk8s/salt/master/service_deployed.sls'),
     Path('salt/metalk8s/salt/master/init.sls'),
     Path('salt/metalk8s/salt/master/accept_keys.sls'),
 
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -216,6 +216,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/registry/init.sls'),
     Path('salt/metalk8s/registry/macro.sls'),
     Path('salt/metalk8s/registry/populated.sls'),
+    Path('salt/metalk8s/registry/service_deployed.sls'),
     Path('salt/metalk8s/registry/files/registry-pod.yaml.j2'),
 
     Path('salt/metalk8s/salt/master/files/master_99-metalk8s.conf.j2'),
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -243,6 +243,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/_modules/metalk8s.py'),
 
     Path('salt/_pillar/metalk8s.py'),
+    Path('salt/_pillar/metalk8s_endpoints.py'),
     Path('salt/_pillar/metalk8s_nodes.py'),
 
     Path('salt/_roster/kubernetes_nodes.py'),
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -1,4 +1,5 @@
 import logging
+
 import salt.utils.files
 import salt.utils.yaml
 
--- a/salt/_pillar/metalk8s_endpoints.py
+++ b/salt/_pillar/metalk8s_endpoints.py
@@ -1,0 +1,66 @@
+"""Store data about bootstrap services ip/port in pillar"""
+
+import logging
+import os.path
+
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_endpoints'
+
+
+def __virtual__():
+    if 'metalk8s_kubernetes.show_endpoint' not in __salt__:
+        return False, 'Missing metalk8s_kubernetes module'
+    else:
+        return __virtualname__
+
+
+def service_endpoints(service, kubeconfig):
+    try:
+        endpoint = __salt__['metalk8s_kubernetes.show_endpoint'](
+            name=service,
+            namespace="kube-system",
+            kubeconfig=kubeconfig,
+        )
+
+        # Extract hostname, ip and node_name
+        res = {
+            k: v
+            for k, v in endpoint['subsets'][0]['addresses'][0].items()
+            if k in ['hostname', 'ip', 'node_name']
+        }
+
+        # Add ports info to res dict
+        ports = {
+            port['name']: port['port']
+            for port in endpoint['subsets'][0]['ports']
+        }
+        res['ports'] = ports
+    except Exception as exc:  # pylint: disable=broad-except
+        log.exception(
+            'Unable to get kubernetes endpoints for %s:\n%s', service, exc
+        )
+        return {}
+    else:
+        return res
+
+
+def ext_pillar(minion_id, pillar, kubeconfig):
+    endpoints = {}
+
+    if not os.path.isfile(kubeconfig):
+        log.warning(
+            '%s: kubeconfig not found at %s', __virtualname__, kubeconfig)
+        return endpoints
+
+    for service in ['salt-master', 'registry', 'package-repositories']:
+        endpoints.update({
+            service: service_endpoints(service, kubeconfig)
+        })
+
+    return {
+        'metalk8s': {
+            'endpoints': endpoints,
+        },
+    }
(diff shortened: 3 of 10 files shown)


salt/_pillar/metalk8s.py
```diff
@@ -49,12 +51,42 @@ def _load_ca(config_data):
     }
 
 
+def _get_endpoints(config="/etc/kubernetes/admin.conf",
```
NicolasT: I'd like to keep pillars-that-dont-need-Kubernetes-available and pillars-that-need-Kubernetes-available separate (in separate modules), so this shouldn't go in here.

NicolasT: Don't use defaults like this, they're arbitrary. Values should be passed in from configuration.

salt/_pillar/metalk8s.py
```diff
@@ -49,12 +51,42 @@ def _load_ca(config_data):
...
+                   context="kubernetes-admin@kubernetes",
+                   namespace="kube-system"):
```
NicolasT: Making `namespace` an input argument, but `service` not, isn't very useful.

No need to make any of this configurable, IMHO. We know what this `ext_pillar` is supposed to provide. If more, then code changes are required.

salt/_pillar/metalk8s.py
```diff
@@ -49,12 +51,42 @@ def _load_ca(config_data):
...
+    """Find out services endpoints."""
+    endpoints = {}
+
+    if 'kubernetes.show_endpoint' not in __salt__ \
+            or not os.path.exists(config):
+        return endpoints
+
+    for service in ('salt-master', 'registry', 'package-repositories'):
```
NicolasT: These strings form a list or a set, not a tuple.

salt/_pillar/metalk8s.py
```diff
@@ -49,12 +51,42 @@ def _load_ca(config_data):
...
+        try:
+            endpoint = __salt__['kubernetes.show_endpoint'](
+                name=service,
+                namespace=namespace,
+                kubeconfig=config,
+                context=context
+            )
+            endpoint_ip = endpoint["subsets"][0]["addresses"][0]['ip']
```
NicolasT: Why throw away port information?

salt/metalk8s/orchestrate/bootstrap_with_master.sls
```diff
@@ -27,6 +27,17 @@ Bootstrap control plane:
...
+Bootstrap services:
+  salt.state:
+    - tgt: {{ pillar['bootstrap_id'] }}
+    - saltenv: {{ saltenv }}
+    - sls:
+      - metalk8s.bootstrap.services
+    - require:
+      - salt: Bootstrap control plane
+    - pillar:
+        registry_ip: {{ pillar.get('registry_ip') }}
```
NicolasT: Is this used?

NicolasT (changes requested):
Couple of requests.

salt/metalk8s/orchestrate/bootstrap_with_master.sls
ramanduh: Not used, will remove it

salt/_pillar/metalk8s.py
ramanduh: No specific reason, I just thought that we'll only use ip in our use case, but it makes sense to add port info as well. BTW, we'll have something like this in the `endpoint["subsets"]` dict:
```
>>> pprint(coreV1.read_namespaced_endpoints('salt-master', 'kube-system').to_dict()['subsets'])
[{'addresses': [{'hostname': None,
                 'ip': '172.21.254.3',
                 'node_name': 'bootstrap',
                 'target_ref': {'api_version': None,
                                'field_path': None,
                                'kind': 'Pod',
                                'name': 'salt-master-bootstrap',
                                'namespace': 'kube-system',
                                'resource_version': '577',
                                'uid': 'f857e39e-5d26-11e9-8308-5254008481d5'}}],
  'not_ready_addresses': None,
  'ports': [{'name': 'requestserver', 'port': 4506, 'protocol': 'TCP'},
            {'name': 'publisher', 'port': 4505, 'protocol': 'TCP'},
            {'name': 'api', 'port': 4507, 'protocol': 'TCP'}]}]
```
Which form do you recommend to return the pillar? Return the dict as it is or make some processing to simplify it (in the pillar I only take the first adress in the subset, but I wonder if it is a good idea)

salt/_pillar/metalk8s.py
ramanduh: Do you mean move it in `metalk8s_nodes.py` or create another module? BTW I just see that in `metalk8s_nodes.py` we use directly the `python-kubernetes` python client. In this PR context, is it preferable to use  `python-kubernetes` client directly or do we want to use salt client and then maintain `salt/_modules/kubernetesmod.py`?

salt/_pillar/metalk8s.py
ramanduh: will fix

salt/_pillar/metalk8s.py
ramanduh: will fix

salt/_pillar/metalk8s.py
ramanduh: will fix

NicolasT: I'd say, add another pillar module.

You can keep using the `__salt__` procedures, that's fine. The other one uses `python-kubernetes` directly since it's based on some work I did a long time ago. So is the roster module, actually.

I guess over time we should harmonize those.

salt/_pillar/metalk8s.py
NicolasT: Taking 'the first one' is fine, for now there will be only one anyway, since the bootstrap node is not HA right now anyway.

I think what you may want to return is something like the following:

```json
[
  {
    "node_name": "bootstrap",
    "hostname": null,
    "ip": "172.21.254.3",
    "ports": {
      "requestserver": 4506,
      "publisher": 4505,
      "api": 4507
    }
  }
]
```

Then, in the *ext_pillar*, we set the pillar value to the 'first' element of what this procedure returns, and then consumer of that pillar value can use something like

```jinja
foo: {{ salt_master.ip }}:{{ salt_master.ports.publisher }}
```

or something along those lines.

buildchain/buildchain/salt_tree.py
```diff
@@ -113,6 +113,7 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
     Path('salt/metalk8s/bootstrap/preflight.sls'),
     Path('salt/metalk8s/bootstrap/registry.sls'),
     Path('salt/metalk8s/bootstrap/salt_master.sls'),
+    Path('salt/metalk8s/bootstrap/services.sls'),
```
gdemonet, line 118: So, I was wondering, do we need such a formula? If we follow the structure defined in https://github.com/scality/metalk8s/pull/967, I guess we could have a `deployed.sls` in all "services", i.e. `metalk8s.registry`, `metalk8s.repo` and `metalk8s.salt.master`. All these will be called as soon as we have an API server ready.

ramanduh: @TeddyAndrieux any thought on this? Is there anything I can change in my PR to facilitate https://github.com/scality/metalk8s/pull/967 ? FYI: for now I can move the service object deployment in `salt/metalk8s/salt/master/deployed.sls` because when the state deployed.sls is applied, the API server is not up yet. Does it make sense for you if this change is managed in https://github.com/scality/metalk8s/pull/967 ?

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -0,0 +1,56 @@
...
+def __virtual__():
+    return __virtualenv__
+
+
+def service_endpoints(service):
+    try:
+        endpoint = __salt__['kubernetes.show_endpoint'](
+            name=service,
+            namespace='kube-system',
+            kubeconfig='/etc/kubernetes/admin.conf',
```
NicolasT: These should ideally come from the `ext_pillar` configuration, not hard-coded.

buildchain/buildchain/salt_tree.py
TeddyAndrieux: As for the moment we already have `deployed.sls` for salt master and repo (which is not real "deployed" as we want to use in our salt states)
I think you should keep the naming `service_deployed.sls` for the moment. I will rename them in my PR

salt/_pillar/metalk8s_endpoints.py
ramanduh: I'll move namespace and kubeconfig to ext_pillar configuration

ramanduh:
@bert-e reset

ramanduh:
@bert-e reset

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -0,0 +1,57 @@
+"""Store data about bootstrap services ip/port in pillar"""
+
+import logging
+
+
+log = logging.getLogger(__name__)
+
+__virtualenv__ = 'metalk8s_endpoints'
```
NicolasT: It's `__virtualname__`, not `__virtualenv__`

NicolasT (changes requested):

salt/_pillar/metalk8s_endpoints.py
ramanduh: oups, will fix

ramanduh:
@bert-e reset

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -0,0 +1,57 @@
...
+import logging
+
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_endpoints'
+
+
+def __virtual__():
+    return __virtualname__
```
NicolasT: Strictly speaking this should indicate failure if `kubernetes.show_endpoint` can't be found in `__salt__`, but fine.

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        # Add ports info to res dict
+        ports = {
+            port['name']: port['port']
+            for port in endpoint['subsets'][0]['ports']
+        }
+        res['ports'] = ports
+    except Exception as exc:  # pylint: disable=broad-except
+        log.error(
```
NicolasT: Could use `log.exception` instead.

NicolasT (dismissed):

ramanduh:
@bert-e reset

ramanduh:
@bert-e help

ramanduh:
@bert-e after_pull_request 1009

ramanduh:
@bert-e after_pull_request=1009

ramanduh:
@bert-e reset

TeddyAndrieux:
@bert-e status

TeddyAndrieux:
@bert-e reset

NicolasT (approved):

NicolasT:
@bert-e bypass_author_approval

PR #985, opened 2019-04-11
salt: set `salt-master` log-level to `info`

NicolasT:


TeddyAndrieux (approved):

NicolasT:
@bert-e approve

salt/metalk8s/salt/master/files/salt-master-pod.yaml.j2
```diff
@@ -27,6 +27,7 @@ spec:
   containers:
     - name: salt-master
       image: {{ build_image_name(salt_master_image, salt_master_version) }}
+      args: ['--log-level=info']
```
sayf-eddine-scality, line 30: we can remove skipping salt master pod log check in tests in this case :)

NicolasT: No, we'd still run into the 'multi-container Pod' issue.

The test to validate logs are tracked does its job, even when skipping `salt-master`. I don't think we should put any extra energy in it.

sayf-eddine-scality: well imo it does not check if salt-master is logging to stdout or to some logfile somewhere, I dont understand how the salt pod is affected by the 'multi-container' pod issue

NicolasT: The test invokes `kubectl logs $POD` as-is. If a Pod has multiple containers in it, one must specify the container name as well to get its logs. Since the test is implemented by looping over Pods it finds, this fails when it tries to retrieve the logs from the `salt-master` Pod. Fixing this would require rewriting the test, to get a list of containers in a Pod from the API, and retrieve logs per-container.

sayf-eddine-scality: given that salt-master and salt-api are in the same component it maybe worth it to use `--all-containers=true` to get the logs or re-implementing the test, tbd later so

PR #988, opened 2019-04-11
Run all tests in multi-nodes deployment in CI

gdemonet:
## What

This PR introduces a new VM in the Terraform multi-nodes spawn process, called _**bastion**_. It allows one to have access to the private networks spawned, unlike the Eve worker.

The post-install tests (defined under `tests/post`) are now all run without custom options in the multi-nodes deployment context (previously, we invoked `pytest` with custom CLI options `--bootstrap-ip` and `--skip-tls-verify` for performing KubeAPI calls from outside the control-plane network).

## How to test

Look at Eve build reports, see that the tests all run in the multi-nodes stage (no test is skipped).

## Left to do

Some tasks, related to this PR, will need to be tackled in (yet-to-be-created) other issues:

- One would want the bastion to run the install scenario instead of the worker
- One would want to have a more friendly Terraform tooling, usable locally
- One would like to remove `kubectl` invocations in tests, and only use the Python Kubernetes client instead


eve/workers/openstack-multiple-nodes/terraform/network.tf
```diff
@@ -6,9 +8,62 @@ variable "openstack_network" {
   }
 }
 
+# Default internal networks
+variable "control_plane_cidr" {
+  type    = "string"
+  default = "172.21.254.0/28"
```
NicolasT: Can I suggest to use *another* subnet for these two, such that we never accidentally make assumptions on the Vagrant environment being 'the default'?

gdemonet: Yes we can, what would fit in your opinion? `172.22.254.X` instead of `172.21.254.X` would be OK?

gdemonet: Done: 070c4a8

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

NicolasT:
Does this actually enable all tests we're currently skipping in multi-node CI?

gdemonet:
> Does this actually enable all tests we're currently skipping in multi-node CI?

Yes, because we were using the `--bootstrap-ip` pytest CLI option to "know" whether the tests run in a multi-node context. However, I forgot to remove that code from the tests themselves, I guess I should do it in this PR.

NicolasT:
> > Does this actually enable all tests we're currently skipping in multi-node CI?
> 
> Yes, because we were using the `--bootstrap-ip` pytest CLI option to "know" whether the tests run in a multi-node context. However, I forgot to remove that code from the tests themselves, I guess I should do it in this PR.

Yes please :)

gdemonet:
> Yes please :)

Done: e528597

gdemonet:
@bert-e approve

NicolasT (approved):

Issue #979, opened 2019-04-11
Create a salt module that `drain` node


TeddyAndrieux:
For maintenance/upgrade we need to be able to `drain` a node.

Need to implement a `kubectl drain <node>` like in salt.

https://kubernetes.io/docs/tasks/administer-cluster/cluster-management/#maintenance-on-a-node

jbertran:
Closed by #992

PR #992, opened 2019-04-12
Add node_drain function

cmonfort:
**Component**: salt

**Context**: To upgrade, we need to be able to drain nodes. Tackles #979 

**Summary**:

Add a `drain` capability to our salt custom modules and states. Since we are beginning to have a *very* significant `metalk8s_kubernetes` custom module/state, we split the `drain` and `cordon` logic outside of it.

We take a few shortcuts to get this to land:

* No use of the `Discovery` API to find out whether or not k-a supports eviction, since all versions we support *do* support eviction
* No management of deletion vs eviction, for the same reason
* No parallel execution of Eviction object creation and waiting, differing from the `kubectl` implementation
* No handling of a potential `TooManyRequests` error, requires more investigation to get it from the raw k-a request.

**Acceptance criteria**:

Test by applying the following patch (edit `node_name` to target a different node):

```diff
diff --git a/buildchain/buildchain/salt_tree.py b/buildchain/buildchain/salt_tree.py
index cb29f6ec..c658fe11 100644
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -199,6 +199,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/orchestrate/bootstrap/init.sls'),
     Path('salt/metalk8s/orchestrate/bootstrap/accept-minion.sls'),
     Path('salt/metalk8s/orchestrate/deploy_new_node.sls'),
+    Path('salt/metalk8s/orchestrate/test_drain.sls'),
     Path('salt/metalk8s/orchestrate/upgrade/etcd.sls'),
 
     Path('salt/metalk8s/registry/deployed.sls'),
diff --git a/salt/metalk8s/orchestrate/test_drain.sls b/salt/metalk8s/orchestrate/test_drain.sls
new file mode 100644
index 00000000..64683ecb
--- /dev/null
+++ b/salt/metalk8s/orchestrate/test_drain.sls
@@ -0,0 +1,8 @@
+Test drain function:
+  module.run:
+    - metalk8s_kubernetes.node_drain:
+        - node_name: "bootstrap"
+        - force: False
+        - ignore_daemonset: True
+        - delete_local_data: True
+        - kubeconfig: {{ pillar['kubeconfig'] }}
```

In your salt-master container, run:

```shell
salt-run saltutil.sync_all saltenv=metalk8s-2.0
salt-run state.orch metalk8s.orchestrate.test_drain saltenv=metalk8s-2.0 pillar='{"kubeconfig": "/etc/kubernetes/admin.conf"}'
```

**Expectations**

When calling the custom state `node_drained` on a node:

* a dummy application pod on the node gets deleted
* the state returns a `CommandExecutionError`:
  *  if there is a daemonset pod on the node unless `force` is set to `True`
  * if there is a pod with local storage unless `_delete_local_data` is set to `True`
  * if there is a pod with the annotation `kubernetes.io/config.mirror`

Additionally, after the state returns, the node should be unschedulable (i.e. still cordoned).


NicolasT:
Let's try to keep the implementation as close to https://github.com/kubernetes/kubernetes/blob/ca89308227abdf4f2c895118528c354c32aa3046/pkg/kubectl/cmd/drain.go#L239 as possible :)

salt/_modules/kubernetesmod.py
```diff
@@ -2276,3 +2276,32 @@ def node_set_taints(node_name, taints, **kwargs):
...
+    '''
+
+    '''
+    Data:
+        * List of namespaces
+        * List of pods to evict (for a given namespace)
+        * List of evicted pods to delete (for a given namespace)
+
+    Get list of all namespaces (need to handle continue parameter?)
+    For given node:
```
sayf-eddine-scality: shouldn't we mark the node as unschedul-able (cordon) so no new pods enter while we are draining ?

sayf-eddine-scality (dismissed):

salt/_modules/kubernetesmod.py
cmonfort: cordon should have been made before calling drain...

NicolasT: Yes (as does `kubectl drain`), otherwise the draining process will never converge.

cmonfort: yep...

salt/_modules/metalk8s_cordon.py
```diff
@@ -0,0 +1,117 @@
...
+        self._conditions = conditions
+
+    setattr(
+        V1beta1CustomResourceDefinitionStatus, 'conditions',
+        property(
+            fget=V1beta1CustomResourceDefinitionStatus.conditions.fget,
+            fset=set_conditions
+        )
+    )
+    # End of workaround
```
TeddyAndrieux: I think this workaround is not needed here

salt/_modules/metalk8s_cordon.py
```diff
@@ -0,0 +1,117 @@
+import logging
+
+from salt.exceptions import CommandExecutionError
+
+import metalk8s_kubernetes
```
TeddyAndrieux: Weird, you "by pass" the "virtual check" + "virtual name" by importing directly the python module.

Would suggest to use `__salt__['metalk8s_kubernetes.<function>']` instead of importing the python module

Or if we have some "private" python function we want to use in several modules then we should have some `utils`
https://docs.saltstack.com/en/latest/topics/utils/index.html

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,516 @@
+import logging
+import operator
+import time
+
+from salt.exceptions import CommandExecutionError
+
+import metalk8s_cordon
+import metalk8s_kubernetes
```
TeddyAndrieux: Same as above

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,516 @@
...
+        self._conditions = conditions
+
+    setattr(
+        V1beta1CustomResourceDefinitionStatus, 'conditions',
+        property(
+            fget=V1beta1CustomResourceDefinitionStatus.conditions.fget,
+            fset=set_conditions
+        )
+    )
+    # End of workaround
```
TeddyAndrieux: Same, this workaround is maybe not needed here

TeddyAndrieux (changes requested):

salt/_modules/metalk8s_cordon.py
jbertran: Seen f2f - the workaround is for most listing operations but operates by modifying an underlying operation.

salt/_modules/metalk8s_cordon.py
jbertran: As we saw, `_utils` is not adapted to grabbing config items. Did the custom module for support functions - in this case, we only need `setup_conn` and `cleanup` from the `metalk8s_kubernetes` module.

salt/_modules/metalk8s_cordon.py
```diff
@@ -0,0 +1,113 @@
...
+    Return unschedulable value of the node identified by the specified name
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_unschedulable node_name="minikube"
+    '''
+    match = __salt__['metalk8s_kubernetes.node'](node_name, **kwargs)
+
+    if match is not None:
+        return match['spec']['unschedulable'] or False
```
gdemonet, line 47: Why do we need this `or False`? `.spec.unschedulable` could be `None`?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,529 @@
+import logging
+import operator
+import time
+
+from salt.exceptions import CommandExecutionError
+
+try:
+    import kubernetes  # pylint: disable=import-self
```
gdemonet: This `pylint` disable isn't required IIUC

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,529 @@
...
+
+class DrainException(Exception):
+    '''General purpose drain exception.'''
+
+    def __init__(self, message):
+        super(DrainException, self).__init__()
+        self.message = message
+
+    def __str__(self):
+        print(self.message)
```
gdemonet: `__str__` shouldn't print anything, but instead return a string... Stray debugging code?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+class DrainTimeoutException(DrainException):
+    '''Timeout-specific drain exception.'''
+    pass
+
+
+def _mirrorpod_filter(pod):
```
gdemonet, line 63: I'm curious, does it mean we can't drain static Pods? (cause that's what the `config.mirror` annotation is for)

salt/_states/metalk8s_drain.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import absolute_import
```
jbertran, line 1: I'm not sure we need this file, if draining will only be done within an orchestrate we can just use a `module.run`

salt/_modules/metalk8s_cordon.py
jbertran: Can be removed since that correction is only for CRDs. It may arise for other listing endpoints though.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+    def localstorage_filter(self, pod):
+        '''Check if given pod can be elected for deletion,
+        regarding local storage
+
+        return: (deletable: bool, warning: str)
+        raises: DrainException if deleting local data is not possible
+        '''
+        if not _has_local_storage(pod):
+            return True, ""
+        if not self._delete_local_data:
```
gdemonet:
```suggestion
        if not self.delete_local_data:
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        return: (deletable flag, warning msg, fatal msg)
+        '''
+        # finished pods can be removed
+        if pod.status.phase in (POD_STATUS_SUCCEEDED, POD_STATUS_FAILED):
+            return True, None, None
+
+        try:
+            controller_ref = self.get_pod_controller(pod)
+        except DrainException as exc:
+            raise
```
gdemonet: Not sure if necessary, this `try/except` is basically a no-op

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            return True, ""
+        if not self._delete_local_data:
+            raise DrainException(self.FATAL_MSG["localStorage"])
+        return True, self.WARNING_MSG["localStorage"]
+
+    def unreplicated_filter(self, pod):
+        '''Check if given pod can be elected for deletion,
+        regarding replication
+
+        return: (deletable flag, warning msg, fatal msg)
```
gdemonet: Signature doesn't seem consistent, you either return:

- `True, None, None` for finished pods
- `True, ""` if you have a controller
- `True, self.WARNING_MSG["unmanaged"]` if you don't and `force == True`

And it seems fatal messages are actually not returned, but instead raised through `DrainException`

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        '''Get the controller object from a reference to it
+
+        return:
+          - the controller object if found
+          - None if not found
+        raises: CommandExecutionError if API fails
+        '''
+        API_CONTROLLERS = {
+            "ReplicationController": {
+                "call": "CoreV1Api->list_namespaced_replication_controller",
```
gdemonet, line 255: (minor) You could build this one on the fly:

```python
api_source = kubernetes.client.CoreV1Api
api_operation = "list_namespaced_replication_controller"
call = "{}->{}".format(api_source.__qualname__, api_operation)
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            )(
+                namespace=namespace,
+                field_selector='metadata.name={0}'.format(controller_ref.name)
+            )
+        except (ApiException, HTTPError) as exc:
+            if isinstance(exc, ApiException) and exc.status == 404:
+                return None
+
+            log.exception('Exception when calling %s', api_call)
+            raise CommandExecutionError(exc)
```
gdemonet: I suppose you need to cast `exc` into a string

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        '''Get a pod's controller object reference
+
+        return: controller
+        raises: DrainException if controller not found
+        '''
+        controller_ref = _get_controller_of(pod)
+        if controller_ref is None:
+            return None
+        # Fire error if - and only if - controller is gone/missing...
+        # TODO: Is it done
```
gdemonet: Can we remove this `TODO`? It's quite obvious that the response of a `list_...` request will be an empty list if no match was found.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        node identified by the name `node_name`.
+        '''
+        warnings = {}
+        failures = {}
+        pods = []
+
+        try:
+            api_instance = kubernetes.client.CoreV1Api()
+            api_response = api_instance.list_pod_for_all_namespaces(
+                field_selector='spec.nodeName={0}'.format(self._node_name)
```
gdemonet:
```suggestion
                field_selector='spec.nodeName={0}'.format(self.node_name)
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            if isinstance(exc, ApiException) and exc.status == 404:
+                return []
+            log.exception(
+                'Exception when calling '
+                'CoreV1Api->list_pod_for_all_namespaces')
+            raise CommandExecutionError(exc)
+
+        for pod in api_response.items:
+            pod_deletable = True
+            for filt in (
```
gdemonet: I don't see the benefit of shortening the variable, `filt` doesn't mean anything, and it's not like we care about the built-in `filter` function here:

```suggestion
            for filter in (
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                    _mirrorpod_filter,
+                    self.localstorage_filter,
+                    self.unreplicated_filter,
+                    self.daemonset_filter
+            ):
+                try:
+                    filter_deletable, warning = filt(pod)
```
gdemonet: As per our Python Best Practices guide, boolean variables should use a prefix like `is_`, `has_`, or similar.

```suggestion
                    is_deletable, warning = filter(pod)
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        except (ApiException, HTTPError) as exc:
+            if isinstance(exc, ApiException) and exc.status == 404:
+                return []
+            log.exception(
+                'Exception when calling '
+                'CoreV1Api->list_pod_for_all_namespaces')
+            raise CommandExecutionError(exc)
+
+        for pod in api_response.items:
+            pod_deletable = True
```
gdemonet:
```suggestion
            is_pod_deletable = True
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                try:
+                    filter_deletable, warning = filt(pod)
+                except DrainException as exc:
+                    failures.setdefault(
+                        exc.message, []).append(pod.metadata.name)
+
+                if warning:
+                    warnings.setdefault(
+                        warning, []).append(pod.metadata.name)
+                pod_deletable = pod_deletable and filter_deletable
```
gdemonet: Assignment operators are common enough in Python to use them for this kind of boolean update:

```suggestion
                is_pod_deletable &= is_deletable
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            if pod_deletable:
+                pods.append(pod)
+
+        def message(errors):
```
gdemonet: This kind of helper method has nothing specific about the context of the `get_pods_for_eviction` method, let's move it out of the `Drain` class instead of instantiating a new one everytime.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            msgs = []
+            for key, msg in errors.items():
+                msgs.append("{0}: {1}".format(key, ", ".join(msg)))
+            return "; ".join(msgs)
```
gdemonet: (minor/nitpick): could use list comprehensions

```suggestion
            return "; ".join(
                "{}: {}".format(key, ", ".join(messages))
                for key, messages in errors.items()
            )
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+
+        if failures:
+            raise DrainException(message(failures))
+        if warnings:
+            log.warning("WARNING: %s", message(warnings))
+        return pods
+
+    def run_drain(self, dry_run=False, **kwargs):
```
gdemonet: Not fond of having `__salt__` appear in the `Drain` class. For instance, it could be used/tested outside of the Salt context if we didn't have this.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        '''Cordon the node if necessary, then drain it.'''
+        node_unschedulable = __salt__['metalk8s_kubernetes.node_unschedulable'](
+            self.node_name, **kwargs
+        )
+        if not node_unschedulable:
```
gdemonet: TOCTOU risk here: why not try to cordon and ignore any error if it's already done?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        '''Delete or evict pods based on eviction support.
+
+        In practice, all the versions we support have eviction, so we just
+        evict the pods.
+        '''
+        try:
+            pods = self.get_pods_for_eviction()
+        except DrainException as exc:
+            if dry_run:
+                pods = None
```
gdemonet: Why? We'll raise the exception anyway, so there is no real need to populate the `pods` variable.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            raise CommandExecutionError(
+                (
+                    "The following are not deletable: {0}. "
```
gdemonet, line 424: Are we sure that `exc.message` will be a list of Pods?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                    "You can ignore DaemonSet pods with the "
+                    "ignore_daemonset flag."
+                ).format(
+                    exc.message
+                )
+            )
+
+        if dry_run:
+            return "Prepared for deletion of pods: {0}".format(
+                ", ".join(pods) if pods else "no pods to delete."
```
gdemonet: Would you rather see `Prepared for deletion of pods: no pods to delete.`, or just `No pod to delete.`? If the latter (which is my stance), then you could rewrite this return as:

```python
return "Prepared for deletion of pods: {}.".format(", ".join(pods)) if pods else "No pod to delete."
```

Note that the period was missing in the format string.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                    field_selector='metadata.name={0}'.format(name)
+                )
+                return api_response
+            except (ApiException, HTTPError) as exc:
+                if isinstance(exc, ApiException) and exc.status == 404:
+                    return None
+
+                log.exception(
+                    'Exception when calling '
+                    'CoreV1Api->list_namespaced_daemon_set')
```
gdemonet:
```suggestion
                    'CoreV1Api->list_namespaced_pod')
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                    exc.message
+                )
+            )
+
+        if dry_run:
+            return "Prepared for deletion of pods: {0}".format(
+                ", ".join(pods) if pods else "no pods to delete."
+            )
+
+        def get_pod(namespace, name):
```
gdemonet: Same as the `message` helper above, there is no reason to instantiate it at every call, let's move it outside of the `Drain` class.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        except DrainTimeoutException as exc:
+            raise CommandExecutionError(
+                "Drain did not complete within {0}".format(exc.message)
+            )
+        except DrainException as exc:
+            remaining_pods = self.get_pods_for_eviction()
+            raise CommandExecutionError(
+                "Pending pods remaining when the error occurred: {0}".format(
+                    exc),
+                [pod.metadata.name for pod in remaining_pods]
```
gdemonet, line 451: Now that's looking really weird, we're instantiating `CommandExecutionError` with a message (rendered from the caught exception) **and** a list of Pod names...

I guess we wanted something closer to:

```python
raise CommandExecutionError(
    "An error occured while draining: {}. The following pods remain: {}.".format(
        exc.message, ", ".join(pod.metadata.name for pod in remaining_pods)
    )
)
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            )
+        return "Eviction complete."
+
+    def evict_pods(self, pods, get_pod_fn):
+        '''Evict all pods from list'''
+        # TODO: there are a lot of shortcuts here, to be checked
```
gdemonet: What does this mean?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                log.exception(
+                    'Exception when calling '
+                    'CoreV1Api->list_namespaced_daemon_set')
+                raise CommandExecutionError(exc)
+
+        try:
+            self.evict_pods(pods, get_pod)
+        except DrainTimeoutException as exc:
+            raise CommandExecutionError(
+                "Drain did not complete within {0}".format(exc.message)
```
gdemonet: This way of passing `self.timeout` is quite convoluted...

```suggestion
                "Drain did not complete within {} seconds".format(self.timeout)
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+                    exc),
+                [pod.metadata.name for pod in remaining_pods]
+            )
+        return "Eviction complete."
+
+    def evict_pods(self, pods, get_pod_fn):
+        '''Evict all pods from list'''
+        # TODO: there are a lot of shortcuts here, to be checked
+
+        global_timeout = self._timeout
```
gdemonet:
```suggestion
        global_timeout = self.timeout
```

gdemonet: But actually, you could just pass `self.timeout` to `self.wait_for_delete`, and even better, just let `wait_for_delete` read it from `self`.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+
+        if pending:
+            raise DrainTimeoutException(global_timeout)
+
+    @staticmethod
+    def not_found(response):
+        '''Check whether a K8s API response indicates a resource not found.'''
+        return response is None or len(response.items) == 0
+
+    def wait_for_delete(self, pods, interval, timeout, get_pod_fn):
```
gdemonet: This function signature could be simply `wait_for_delete(self, pods)`, and:

- replace `interval` with `self.KUBECTL_INTERVAL`
- replace `timeout` with `self.timeout`
- replace `get_pod_fn` with the helper function defined at module-level (see comment above)

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        '''Wait for pods deletion.'''
+        total_t = 0
+        while total_t < timeout and pods:
+            pending = []
+            loop_s = time.time()    # For time processing pods
```
gdemonet:
```suggestion
            loop_s = time.time()  # For time processing pods
```

Oh and I don't really understand this variable name, but fine.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            for pod in pods:
+                response = get_pod_fn(pod.metadata.namespace, pod.metadata.name)
+                if any([
+                        self.not_found(response),
+                        response.items[0].metadata.uid != pod.metadata.uid
```
gdemonet: This will break if `response.items` is empty. We need to use `or` instead of `any`, due to Python way of evaluating expressions.

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        if pending:
+            raise DrainTimeoutException(global_timeout)
+
+    @staticmethod
+    def not_found(response):
+        '''Check whether a K8s API response indicates a resource not found.'''
+        return response is None or len(response.items) == 0
+
+    def wait_for_delete(self, pods, interval, timeout, get_pod_fn):
+        '''Wait for pods deletion.'''
```
gdemonet: What's the difference between `"eviction"` and `"deletion"` in our context?

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            pods = pending
+            if pods and loop_d < interval:
+                time.sleep(interval - loop_d)
+            total_t += time.time() - loop_s
+        return pods
+
+    def evict_pod(self, pod):
+        '''Evict a pod.'''
+        delete_options = V1DeleteOptions()
+        if self._grace_period >= 0:
```
gdemonet:
```suggestion
        if self.grace_period >= 0:
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+            delete_options.grace_period = self._grace_period
```
gdemonet:
```suggestion
            delete_options.grace_period = self.grace_period
```

salt/_modules/metalk8s_drain.py
```diff
@@ -0,0 +1,512 @@
...
+        force=force,
+        grace_period=grace_period,
+        ignore_daemonset=ignore_daemonset,
+        timeout=timeout,
+        delete_local_data=delete_local_data
+    )
+    try:
+        result = drainer.run_drain(dry_run=dry_run, **kwargs)
+    except CommandExecutionError as exc:
+        return False, exc.message
```
gdemonet: Is it common practice to return a tuple from a Salt execution function?

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2306,70 +2310,49 @@ def node_set_taints(node_name, taints, **kwargs):
         _cleanup(**cfg)
 
 
-def node_unschedulable(node_name, **kwargs):
-    '''
-    Return unschedulable value of the node identified by the specified name
+# TODO: Attempt for a central point to call kubernets API,
+#       in order to have a centralized and consistent way
+#       of dealing with errors.
+def call_kube_api(api_instance, method_name, *args, **kwargs):
```
gdemonet: That's not used in this PR, maybe you could keep it a branch for later use?

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+
+    HAS_LIBS = True
+except ImportError:
+    HAS_LIBS = False
+
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'metalk8s_kubernetes'
```
gdemonet, line 39: The fact that we use the same virtual name to merge methods is not obvious. I'd like a module docstring that mentions it, please :)

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    '''
+    Check dependencies
+    '''
+    if HAS_LIBS:
+        return __virtualname__
+
+    return False, 'python kubernetes library not found'
+
+
+def setup_conn_old(**kwargs):
```
gdemonet, line 52: Do we need to keep this `old` method? We don't plan on using the `old`-fashion config, IMHO.

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    kubeconfig_data = kwargs.get('kubeconfig_data') or __salt__['config.option']('kubernetes.kubeconfig-data')
+    context = kwargs.get('context') or __salt__['config.option']('kubernetes.context') or None
+
+    if (kubeconfig_data and not kubeconfig) or (kubeconfig_data and kwargs.get('kubeconfig_data')):
+        with tempfile.NamedTemporaryFile(prefix='salt-kubeconfig-', delete=False) as kcfg:
+            kcfg.write(base64.b64decode(kubeconfig_data))
+            kubeconfig = kcfg.name
+
+    if not kubeconfig:
+        if kwargs.get('api_url') or __salt__['config.option']('kubernetes.api_url'):
```
gdemonet, line 140: Let's remove this part as well as `setup_conn_old`

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+                    'Use \'kubeconfig\' and \'context\' instead.')
+            try:
+                return setup_conn_old(**kwargs)
+            except Exception:
+                raise CommandExecutionError('Old style kubernetes configuration is only supported up to python-kubernetes 2.0.0')
+        else:
+            raise CommandExecutionError('Invalid kubernetes configuration. Parameter \'kubeconfig\' is required.')
+    kubernetes.config.load_kube_config(config_file=kubeconfig, context=context)
+
+    # The return makes unit testing easier
```
gdemonet, line 152: Would even be better to create a client from a new `Configuration` object, instead of using the module-level global `kubernetes.client.configuration`, which could pose issues when testing/debugging. But that's for a future refactor.

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+            salt.utils.files.safe_rm(key)
+        if ca and os.path.exists(ca) and os.path.basename(ca).startswith('salt-kube-'):
+            salt.utils.files.safe_rm(ca)
+    except Exception:
+        pass
+
+
+def cleanup(**kwargs):
+    if not kwargs:
+        return cleanup_old(**kwargs)
```
gdemonet, line 173: Well, that's kind of schizophrenic: _I don't have `kwargs`, let's pass `kwargs`_

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -0,0 +1,182 @@
...
+
+    if 'kubeconfig' in kwargs:
+        kubeconfig = kwargs.get('kubeconfig')
```
gdemonet, line 176: Same story here, you checked that `"kubeconfig"` is in `kwargs`, and then use `.get`, just in case it disappeared :stuck_out_tongue_closed_eyes:

salt/_states/metalk8s_cordon.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import absolute_import
+
+import logging
+
+
+log = logging.getLogger(__name__)
+
+
+def _node_unschedulable(name, value, **kwargs):
```
gdemonet: I know it's not your code, but let's be a tad more explicit with function names (to avoid confusion with the module function `node_unschedulable` that only retrieves the flag value):

```suggestion
def _set_node_unschedulable(name, value, **kwargs):
```

salt/_states/metalk8s_drain.py
gdemonet: Let's keep it for now, might be useful if we want pre-checks for idempotency/early failure.

gdemonet (changes requested):
A large part of the comments are related to the porting of `kubectl drain`, so we may want to move this to a future PR. But I feel like we shouldn't merge this too early, otherwise we'll accumulate way too much debt that will easily be forgotten.

salt/_modules/metalk8s_cordon.py
TeddyAndrieux: Yes it's `None` when `unschedulable` is disabled

salt/_modules/metalk8s_drain.py
jbertran: From copy-paste, will fix. Dunno why we even have pylint annotations on this or other modules, we don't lint or test the salt code.

salt/_modules/metalk8s_drain.py
jbertran: As far as I understand you want mirrors to be deleted manually since they won't be restarted by kube when the node drain completes and is uncordoned.

salt/_modules/metalk8s_drain.py
jbertran: Note to self: need to re-review the control flow in `kubectl`

salt/_modules/metalk8s_drain.py
jbertran: Eeek. Indeed, docstrings in general to be reviewed.

salt/_modules/metalk8s_drain.py
jbertran: Most syntax highlighters will note that as a keyword though, which we don't want. There's little point in making it that terse though, will expand.

salt/_modules/metalk8s_drain.py
jbertran: Indeed. Cordon is idempotent anyway.

salt/_modules/metalk8s_drain.py
jbertran: None. In our context, we only do eviction since a) we support it and b) deletion is a fallback if eviction is not available, according to `kubectl`. A renaming pass is in order.

salt/_modules/metalk8s_kubernetes_utils.py
jbertran: There's a fallback on the "old" method in the "new" one. At this point I'd rather not touch these vendored snippets until we get to refactoring/rewriting the module ourselves.

salt/_modules/metalk8s_cordon.py
gdemonet: Then we can do `bool(match['spec']['unschedulable'])`

salt/_modules/metalk8s_drain.py
gdemonet: > note that as a keyword

which it isn't...

salt/_modules/metalk8s_drain.py
jbertran: no `__qualname__` in Python 2, which we need to support for now.

salt/_modules/metalk8s_drain.py
jbertran: Indeed, the underlying logic was flawed even without that weird no-op. Good catch.

salt/_modules/metalk8s_drain.py
jbertran: The way the caught function is coded, yes.

salt/_modules/metalk8s_drain.py
jbertran: I want to convey the fact that the `drain` function went up to "preparing to delete" the pods (i.e. filtering) explicitly - it didn't just list pods and find none, it found none that fell within its responsibility to delete.

salt/_modules/metalk8s_drain.py
jbertran: There was no correct documentation on this, so I went to look at the source for my own education.

```python
    def __init__(self, message='', info=None):
        [...]
        self.info = info
        if self.info:
            if exc_str_prefix:
                if exc_str_prefix[-1] not in '.?!':
                    exc_str_prefix += '.'
                exc_str_prefix += ' '

            [...]

            # For states, if self.info is a dict also provide an attribute
            # containing a nested output of the info dict without the changes
            # (since they will be in the 'changes' key of the state return and
            # this information would be redundant).
            if isinstance(self.info, dict):
                info_without_changes = copy.deepcopy(self.info)
                info_without_changes.pop('changes', None)
                if info_without_changes:
                    self.strerror_without_changes = \
                        exc_str_prefix + _nested_output(info_without_changes)
                [...
```

So it's OK to pass random stuff to it with the expectation that it'll be represented as usual. But that's not what your comment is about! And indeed this is redundant, especially given that currently there is no `DrainException` to catch so all this is useless.

Again, I need to resync with the `kubectl` code control flow. :angry:

salt/_modules/metalk8s_drain.py
jbertran: Nope! In fact, I should either return `None` or let the `CommandExecutionError` fall through.

salt/_modules/metalk8s_drain.py
gdemonet: Fine by me

salt/_modules/metalk8s_kubernetes.py
jbertran: Pushed in branch `improvement/generic-kube-api-calls`

salt/_modules/metalk8s_drain.py
jbertran: Stray comment, these shortcuts are listed in the PR description.

TeddyAndrieux:
Can you please replace the `TODO` in `salt/metalk8s/orchestrate/deploy_node.sls` with a call to the drain function on `{{ node_name }}`

salt/_modules/metalk8s_drain.py
jbertran: Well not in the strictest sense, no, but some syntax highlighters consider it as one.

gdemonet (approved):
Let's merge for now, refactoring will happen later (#1048).

jbertran:
@bert-e reset

jbertran:
@bert-e approve

jbertran:
> # Not Author
> 
> I'm afraid I cannot do that, @jbertran:
> 
> Only the author of the pull request @cmonfort can use this command.
> 
> Please **delete** the corresponding comment so I can move on.

cjaskljfadjfoidwqjdwqujiwe

cmonfort:
@bert-e approve

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -52,7 +52,14 @@ Cordon the node:
...
 
-# TODO: Call drain
+Drain the node:
+  metalk8s_drain.node_drained:
+    - name: {{ node_name }}
+    - ignore_daemonset: True
+    - delete_local_data: True
+    - force: True
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
```
TeddyAndrieux, line 62: This one should require `Cordon the node` and `Run the highstate` should require the Drain

TeddyAndrieux (dismissed):

PR #995, opened 2019-04-15
Use keepalived for API server failover/VIP

NicolasT:
Please see https://github.com/scality/metalk8s/commit/79251031e8ccb8b590d89e0d3303d34851122fee for some manual steps to be performed in a Vagrant environment before running this.

### DIFF
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -6,7 +6,8 @@
 
 import os
 from pathlib import Path
-from typing import Tuple
+import subprocess
+from typing import Optional, Tuple
 
 from buildchain import ROOT  # Re-export ROOT through this module.
 from buildchain import config
@@ -63,7 +64,12 @@ BUILDER_BASIC_CMD : Tuple[str, ...] = (
 
 K8S_VERSION  : str = '1.11.7'
 SALT_VERSION : str = '2018.3.4'
+KEEPALIVED_VERSION : str = '1.3.5-8.el7_6'
+KEEPALIVED_BUILD_ID : int = 1
 
+CENTOS_BASE_IMAGE : str = 'docker.io/centos'
+CENTOS_BASE_IMAGE_SHA256 : str = \
+    '6ae4cddb2b37f889afd576a17a5286b311dcbf10a904409670827f6f9b50065e'
 
 def load_version_information() -> None:
     """Load version information from `VERSION`."""
@@ -92,5 +98,22 @@ load_version_information()
 
 SHORT_VERSION : str = '{}.{}'.format(VERSION_MAJOR, VERSION_MINOR)
 VERSION : str = '{}.{}{}'.format(SHORT_VERSION, VERSION_PATCH, VERSION_SUFFIX)
+
+
+def git_ref() -> Optional[str]:
+    """Load version information from Git."""
+
+    try:
+        ref : bytes = subprocess.check_output([
+            config.GIT, 'describe', '--always', '--long', '--tags',
+                '--dirty', '--broken',
+        ])
+
+        return ref.decode('utf-8').rstrip()
+    except subprocess.CalledProcessError:
+        return None
+
+
+GIT_REF = git_ref()
 
 # }}}
--- a/buildchain/buildchain/image.py
+++ b/buildchain/buildchain/image.py
@@ -30,6 +30,7 @@ Overview:
 """
 
 
+import datetime
 from pathlib import Path
 from typing import Iterator, Tuple
 
@@ -166,6 +167,11 @@ TO_PULL : Tuple[targets.RemoteImage, ...] = (
 # This should be reset to 1 when SALT_VERSION is changed.
 SALT_MASTER_BUILD_ID : int = 1
 
+KEEPALIVED_IMAGE_TAG = '{}-{}'.format(
+    constants.KEEPALIVED_VERSION,
+    constants.KEEPALIVED_BUILD_ID,
+)
+
 # List of container images to build.
 TO_BUILD : Tuple[targets.LocalImage, ...] = (
     targets.LocalImage(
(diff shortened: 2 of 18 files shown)


buildchain/buildchain/constants.py
```diff
@@ -63,7 +64,15 @@
 
 K8S_VERSION  : str = '1.11.7'
 SALT_VERSION : str = '2018.3.4'
+KEEPALIVED_VERSION : str = '1.3.5-8.el7_6'
+KEEPALIVED_BUILD_ID : int = 1
 
+CENTOS_BASE_IMAGE : str = 'docker.io/centos'
+# pylint: disable=wrong-spelling-in-comment
+# Pylint doesn't like the spelling of `docker.io` ¯\_(ツ)_/¯
```
slaperche-scality: You can add an entry in `.pylint-dict` instead of disabling it ;-)

buildchain/buildchain/constants.py
```diff
@@ -63,7 +64,15 @@
 
 K8S_VERSION  : str = '1.11.7'
 SALT_VERSION : str = '2018.3.4'
+KEEPALIVED_VERSION : str = '1.3.5-8.el7_6'
```
slaperche-scality, line 67: If those constants are only used in `image.py` then they should be defined there, closer to their uses.

buildchain/buildchain/constants.py
```diff
@@ -93,4 +102,21 @@ def load_version_information() -> None:
 SHORT_VERSION : str = '{}.{}'.format(VERSION_MAJOR, VERSION_MINOR)
 VERSION : str = '{}.{}{}'.format(SHORT_VERSION, VERSION_PATCH, VERSION_SUFFIX)
 
+
+def git_ref() -> Optional[str]:
```
slaperche-scality, line 103: We may want to factor our this with `git_revision` from `buildchain/buildchain/iso.py` (or only keep one, are they seems similar enough to be confusing and looks like two different way to get the same info)

Can be done here in a dedicated commit or in another PR (then a debt ticket should be created).

slaperche-scality (changes requested):

slaperche-scality:
I only reviewed the build part.

buildchain/buildchain/constants.py
NicolasT: Done

buildchain/buildchain/constants.py
NicolasT: We're not using `SALT_VERSION` anywhere else either. It's a constant, over time this (and `SALT_VERSION` and whatever else) should go into some central location where Salt SLSs retrieve the values as well (I believe there's a ticket for that), so IMHO this makes sense?

buildchain/buildchain/constants.py
NicolasT: Done

buildchain/buildchain/constants.py
slaperche-scality: Ha yes, for the version indeed (so we can keep it here)

But the build ID and the SHA256 and the image name can be moved (like `SALT_MASTER_BUILD_ID`) or used inline in `image.py` (like the SHA256 and the image name) I think.

buildchain/buildchain/constants.py
```diff
@@ -93,4 +99,22 @@ def load_version_information() -> None:
...
+    """Load version information from Git."""
+    ref = None
```
slaperche-scality: `ref = None` is useless here.
Also, you could return the result of `strip` directly.

Could be written:

```suggestion
def git_ref() -> Optional[str]:
    """Return the current git revision.

    Return None if we can't get the information.
    """
    cmd = [config.GIT, 'describe']
    cmd.extend(['--always', '--long', '--tags','--dirty', '--broken'])
    try:
        stdout : bytes = subprocess.check_output(cmd)
        return stdout.decode('utf-8').rstrip()
    except subprocess.CalledProcessError:
        return None
```

buildchain/buildchain/image.py
```diff
@@ -178,6 +179,28 @@ def task__image_build() -> Iterator[types.TaskDict]:
         task_dep=['_image_mkdir_root'],
         file_dep=[constants.ROOT/'images'/'salt-master'/'source.key'],
     ),
+    targets.LocalImage(
+        name='keepalived',
+        version='{}-{}'.format(
+            constants.KEEPALIVED_VERSION, constants.KEEPALIVED_BUILD_ID),
```
slaperche-scality: minor: when the opening brace is on its own line, we usually put the closing one on its own as well.

```suggestion
        version='{}-{}'.format(
            constants.KEEPALIVED_VERSION, constants.KEEPALIVED_BUILD_ID
        ),
```

OR

```suggestion
        version='{}-{}'.format(constants.KEEPALIVED_VERSION,
                               constants.KEEPALIVED_BUILD_ID),
```

Ditto for `VERSION` in `build_args` (btw, we may want to factor that in a "local" global in this file.

buildchain/buildchain/constants.py
NicolasT: Disagree.

`CENTOS_BASE_IMAGE` and `CENTOS_BASE_IMAGE_SHA256` are constants very much like `SALT_VERSION` which we may want to update from time to time, *without updating the implementation* of how things are built.

Moving `*_BUILD_ID` somewhere else would be counter-productive, because *build-id is to be reset to 1 when the version of something changes*.

buildchain/buildchain/image.py
NicolasT: Then `pylint` or whatever should check for it.

buildchain/buildchain/constants.py
slaperche-scality: Then `SALT_MASTER_BUILD_ID` should be moved, to be consistent (and if we start having a lof of those version-related constant, we could consider having a dedicated file (`versions.py`) for them)

buildchain/buildchain/image.py
slaperche-scality: We could use `black` for formatting (that way code layout would be consistent, instead of having everyone use what it prefers just because CI doens't enforce it…), and enforce it in the CI (code layout aside, my remark about duplication still stands).

buildchain/buildchain/constants.py
NicolasT: No! The point is that `*_BUILD_ID` and `*_VERSION` *go hand in hand*, they're not two distinct unrelated entities. The *version* specifies the version of the software in the container. The *build* specifies the iteration of container building (basically, tracking changes to `Dockerfile`, whatever entrypoint script(s),...) which combined form the *tag* of a container image.

Furthermore, as I said before, *build* gets reset when *version* is bumped (RPM-style), so there again having them in separate places would be counter-productive/counter-intuitive/...

Yes, as I wrote elsewhere, version management must be more centralized (and cross the build-system / Salt boundary) to avoid duplication and divergence, but that's most definitely out of scope for this PR.

NicolasT:
@bert-e reset

buildchain/buildchain/constants.py
slaperche-scality: > No! The point is that *_BUILD_ID and *_VERSION go hand in hand, they're not two distinct unrelated entities. 

Yeah, nothing in my message was against that (if it relates to my suggestion of moving version-related constants in their own file, this included the build ones, as they are tied to the version they are version-related).

But OK to do it in another PR, one day™

slaperche-scality (approved):

NicolasT:
@bert-e approve

PR #1011, opened 2019-04-17
ci: concurrent build

NicolasT:
Ball-park figures: over 4 random successful recent build, average wall-clock time of the `build` step was 15:33. With this patch, on a single build, wall-clock time of the same step is reduced to 6:02.


NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #1012, opened 2019-04-17
Versionlocks docker & node_exporter packages

alexandre-allard:


roles/docker/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+docker_yum_versionlock_file: /etc/yum/pluginconf.d/versionlock.list
+
+docker_package_names:
```
NicolasT, line 5: Should we also lock `container-selinux`? Or is that pinned as reqs of these packages?

alexandre-allard: That's what I did at the beginning, but I've removed it because there are already dependencies in docker-ce & docker-engine RPMs, so no need for that, yum versionlock will prevent from upgrading if dependencies would not be satisfied.

NicolasT:
Could you replace the `eve` update commit with 4d097b7c3a5247f821451f0c45ded162e0fee72d? Will ease upmerges.

NicolasT:
One more: please add an entry to `ChangeLog.rst` for this change. Other than that, LGTM.

alexandre-allard:
@NicolasT I've cherry-picked 4d097b7c3a5247f821451f0c45ded162e0fee72d and added an entry in `ChangeLog.rst`
I've also changed the way node_exporter is locked, because I found a bug when trying to install the version we want and another version is already installed (there is a comment in the code to explain why it's done like that). Basically, we need to unlock the package before doing the installation, otherwise it would fail if the package is already installed and we want a different version than the one which is installed.

NicolasT (approved):
@xaf-scality will be happy ;)

PR #1016, opened 2019-04-17
NodeCreateForm: use Input core-ui

carlito-scality:
-Check if the NodeCreateForm and LoginForm works like before and the style of inputs looks like the on in core-ui.


MonPote (approved):
LGTM

carlito-scality:
@bert-e approve

PR #1020, opened 2019-04-19
Set-up initial documentation tree and first "Integration" docs

### LINKED ISSUE #951: Define a base "integration" document for external solutions
We would like a reference documentation for integrating packaged applications with MetalK8s.
In order to bootstrap this documentation, let's start with writing down a few definitions and principles defining what we mean by integrating a solution with the platform.

To include:

- [ ] Definition of what a "solution" is
- [ ] Definition of MetalK8s responsibility scope in terms of managing solutions
- [ ] User interactions diagram for deploying a solution on MetalK8s

### LINKED ISSUE #952: Define Solutions archiving guidelines
Solutions (as defined in #951) will be expected by MetalK8s to be archived in the ISO image format (to be specified).

This archive should have a pre-defined structure, roughly:
- `product.txt`
- `images/`
- `operator/`
- `ui/`

Let's create a document describing these rules, including:
- [ ] ISO image format specification - with example `mkisofs` invocation
- [ ] File tree expected structure, with short descriptions (to be linked to other documentation pages)
- [ ] Link to the integration document created in #951

### LINKED ISSUE #953: Define Solutions Operator guidelines
Solutions (as defined in #951) will be expected by MetalK8s to be managed by a **single** [Operator](https://coreos.com/blog/introducing-operators.html).

Let's create a document with guidelines and recommendations for building Operators to be deployed on MetalK8s.

Include:
- [ ] Expected manifests for deploying the Operator
- [ ] Monitoring guidelines (how the Operator should use the Prometheus deployed by MetalK8s)
- [ ] Reference links for building/packaging an Operator
- [ ] Link to this doc in the main integration document (#951)

Note: location of the Operator manifests within the archive should be defined in #952

gdemonet:
We define initial documentation for "integrating with MetalK8s", a.k.a. Solutions.

To build the docs, you can use:

```
# Will build HTML files under docs/_build
tox -e docs

# Will watch docs/ sources and serve built HTML files locally on port 8000
tox -e docs -- livehtml
```

Please make sure to read through and discuss anything feeling unclear. Details regarding the integration with Operators and their deployment will be added in the future, during the investigation work on #1060 and #1061.

**Summary**:

- [x] Define a base "integration" document for external solutions
- [x] Define Solutions archiving guidelines
- [x] Define Solutions Operator guidelines

_Note: UI guidelines are not done in this PR_

---

Fixes: #951
Fixes: #952
Fixes: #953

### DIFF
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -67,6 +67,7 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
    │           ├── <layer_digest>
    │           ├── manifest.json
    │           └── version
+   ├── registry-config.inc
    ├── operator
    │   ├── crd
    │   │   └── some_crd_name.yaml
@@ -147,20 +148,30 @@ hardlinks, using the tool hardlink_::
    $ hardlink -c images
 
 A detailed procedure for generating the expected layout is available at
-`NicolasT/static-container-registry`_. You can use the script provided there
-to generate your own NGINX configuration and test this static registry for
-yourself. MetalK8s will generate its own configuration when importing the
-Solution archive.
+`NicolasT/static-container-registry`_. You can use the script provided there,
+or use the one vendored in this repository (located at
+``buildchain/buildchain/static-container-registry``) to generate the NGINX
+configuration to serve these image layers with the Docker Registry API.
+MetalK8s, when deploying the Solution, will include the ``registry-config.inc``
+file provided at the root of the archive. In order to let MetalK8s control
+the mountpoint of the ISO, the configuration **must** be generated using the
+following options::
+
+   $ ./static-container-registry.py \
+       --name-prefix '{{ repository }}' \
+       --server-root '{{ registry_root }}' \
+       /path/to/archive/images > /path/to/archive/registry-config.inc.j2
 
 Each archive will be exposed as a single repository, where the name will be
-computed as ``<NAME>-<VERSION>`` from :ref:`solution-archive-product-info`.
+computed as ``<NAME>-<VERSION>`` from :ref:`solution-archive-product-info`, and
+will be mounted at ``/srv/scality/<NAME>-<VERSION>``.
 
 .. warning::
 
-   Operator Deployments should not rely on this naming pattern for finding
-   the images for their resources. Instead, the full repository prefix will be
-   stored in a ``ConfigMap``, that Deployments will be able to expose as
-   environment variables. See :doc:`./operator` for more details.
+   Operators should not rely on this naming pattern for finding the images for
+   their resources. Instead, the full repository prefix will be exposed to
+   the Operator container as an environment variable when deployed with
+   MetalK8s. See :doc:`./operator` for more details.
 
 The images names and tags will be inferred from the directory names chosen when
 using ``skopeo copy``. Using `hardlink` is highly recommended if one wants to
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -69,9 +69,13 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
    │           └── version
    ├── registry-config.inc
    ├── operator
-   │   ├── crd
-   │   │   └── some_crd_name.yaml
-   │   └── deployment.yaml
+   |   └── deploy
+   │       ├── crds
+   │       |   └── some_crd_name.yaml
+   │       ├── operator.yaml
+   │       ├── role.yaml
+   │       ├── role_binding.yaml
+   │       └── service_account.yaml
    ├── product.txt
    └── ui
        └── deployment.yaml
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -1,0 +1,157 @@
+Solution archive guidelines
+===========================
+
+To provide a predictable interface with packaged Solutions, MetalK8s expects a
+few criteria to be respected, described below.
+
+
+Archive format
+--------------
+
+Solution archives must use the `ISO-9660:1988`_ format, including `Rock Ridge`_
+and Joliet_ directory records. The character encoding must be UTF-8_. The
+conformance level is expected to be at most 3, meaning:
+
+- Directory identifiers may not exceed 31 characters (bytes) in length
+- File name + ``'.'`` + file name extension may not exceed 30 characters
+  (bytes) in length
+- Files are allowed to consist of multiple sections
+
+The generated archive should specify a volume ID, set to
+``{project_name} {version}``.
+
+.. todo::
+
+   Clarify whether Joliet/Rock Ridge records supersede the conformance level
+   w.r.t. filename lengths
+
+
+.. _`ISO-9660:1988`: https://www.iso.org/obp/ui/#iso:std:iso:9660:ed-1:v1:en
+.. _`Rock Ridge`: https://en.wikipedia.org/wiki/Rock_Ridge
+.. _Joliet: https://en.wikipedia.org/wiki/Joliet_(file_system)
+.. _UTF-8: https://tools.ietf.org/html/rfc3629
+
+Here is an example invocation of the common Unix mkisofs_ tool to generate such
+archive::
+
+   mkisofs
+       -output my_solution.iso
+       -R  # (or "-rock" if available)
+       -J  # (or "-joliet" if available)
+       -joliet-long
+       -l  # (or "-full-iso9660-filenames" if available)
+       -V 'MySolution 1.0.0'  # (or "-volid" if available)
+       -gid 0
+       -uid 0
+       -iso-level 3
+       -input-charset utf-8
+       -output-charset utf-8
+       my_solution_root/
+
+.. _mkisofs: https://linux.die.net/man/8/mkisofs
+
+.. todo::
+
+   Consider if overriding the source files UID/GID to 0 is necessary
+
+
+File hierarchy
+--------------
+
+Here is the file tree expected by MetalK8s to exist in each Solution archive::
+
+   .
+   ├── images
+   │   ├── listing.yaml
+   │   └── some_image_name.tar.gz
+   ├── operator
+   │   ├── crd
+   │   │   └── some_crd_name.yaml
+   │   └── deployment.yaml
+   ├── product.txt
+   └── ui
+       └── deployment.yaml
+
+
+Product information
+-------------------
+
+General product information about the packaged Solution must be stored in the
+``product.txt`` file, stored at the archive root.
+
+It must respect the following format (currently version 1, as specified by the
+``ARCHIVE_LAYOUT_VERSION`` value)::
+
+   NAME=Example
+   VERSION=1.0.0-dev
+   REQUIRE_METALK8S=">=2.0"
+   ARCHIVE_LAYOUT_VERSION=1
+
+It is recommended for inspection purposes to include information related to
+the build-time conditions, such as the following (where command invocations
+should be statically replaced in the generated ``product.txt``)::
+
+   GIT=$(git describe --always --long --tags --dirty)
+   BUILD_TIMESTAMP=$(date +%Y-%m-%dT%H:%M:%SZ)
+
+
+.. _solution-archive-images:
+
+OCI images
+----------
+
+.. todo::
+
+   With `#1049`_, we will have another standard way of serving OCI images with
+   a read-only registry directly exposing the ISO contents.
+   A specific process will be required to package these images, which we need
+   to document here.
+
+.. _`#1049`: https://github.com/scality/metalk8s/issues/1049
+
+MetalK8s exposes container images in the OCI_ format through a registry.
+Solutions should thus provide their container images in a compatible format,
+for example using ``docker save``.
+
+For size concerns, we expect such images to be compressed using ``gzip``.
+
+Here is an example of how to build, save and compress an image with
+``docker``::
+
+   docker build --tag my-image:1.0.0 --file my_image.dockerfile
+   docker save my-image:1.0.0 -o my_image_1_0.tar
+   gzip -9 my_image_1_0.tar
+
+.. _OCI: https://github.com/opencontainers/image-spec/blob/master/spec.md
+
+MetalK8s also defines recommended standards for container images, described in
+:ref:`req-container-images`.
+
+In order for MetalK8s to populate its registry with accurate image tags,
+Solutions must provide a ``listing.yaml`` file under ``/images``, with the
+given format::
+
+   apiVersion: metalk8s.scality.com/v1alpha1
+   kind: ImagesList
+   solution: MySolution
+   images:
+     - archive: my_image_1_0.tar.gz
+       tag: my_image:1.0.0
+
+.. note::
+
+   Each Solution ISO will be made available to Pods as a single repository.
+   Operator Deployments will then be required to accept the repository prefix,
+   stored in a ``ConfigMap``, in order to properly configure the resources they
+   manage.
+
+Operator
+--------
+
+See :doc:`./operator` for how the ``/operator`` directory should be
+populated.
+
+Web UI
+------
+
+.. todo:: Create UI guidelines and reference here
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -62,8 +62,11 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
 
    .
    ├── images
-   │   ├── listing.yaml
-   │   └── some_image_name.tar.gz
+   │   └── some_image_name
+   │       └── 1.0.1
+   │           ├── <layer_digest>
+   │           ├── manifest.json
+   │           └── version
    ├── operator
    │   ├── crd
    │   │   └── some_crd_name.yaml
@@ -72,6 +75,7 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
    └── ui
        └── deployment.yaml
 
+.. _solution-archive-product-info:
 
 Product information
 -------------------
@@ -100,50 +104,63 @@ should be statically replaced in the generated ``product.txt``)::
 OCI images
 ----------
 
-.. todo::
+MetalK8s exposes container images in the OCI_ format through a static
+read-only registry. This registry is built with nginx_, and relies on having
+a specific layout of image layers to then replicate the necessary parts of the
+Registry API that CRI clients (such as ``containerd`` or ``cri-o``) rely on.
 
-   With `#1049`_, we will have another standard way of serving OCI images with
-   a read-only registry directly exposing the ISO contents.
-   A specific process will be required to package these images, which we need
-   to document here.
+Using skopeo_, you can save images as a directory of layers::
 
-.. _`#1049`: https://github.com/scality/metalk8s/issues/1049
+   $ mkdir images/my_image
+   $ # from your local Docker daemon
+   $ skopeo copy --format v2s2 --dest-compress docker-daemon:my_image:1.0.0 dir:images/my_image/1.0.0
+   $ # from Docker Hub
+   $ skopeo copy --format v2s2 --dest-compress docker://docker.io/example/my_image:1.0.0 dir:images/my_image/1.0.0
 
-MetalK8s exposes container images in the OCI_ format through a registry.
-Solutions should thus provide their container images in a compatible format,
-for example using ``docker save``.
+Your ``images`` directory should now resemble this::
 
-For size concerns, we expect such images to be compressed using ``gzip``.
+   images
+   └── my_image
+       └── 1.0.0
+           ├── 53071b97a88426d4db86d0e8436ac5c869124d2c414caf4c9e4a4e48769c7f37
+           ├── 64f5d945efcc0f39ab11b3cd4ba403cc9fefe1fa3613123ca016cf3708e8cafb
+           ├── manifest.json
+           └── version
 
-Here is an example of how to build, save and compress an image with
-``docker``::
+Once all your images were stored this way, you can de-duplicate layers using
+hardlinks, using the tool hardlink_::
 
-   docker build --tag my-image:1.0.0 --file my_image.dockerfile
-   docker save my-image:1.0.0 -o my_image_1_0.tar
-   gzip -9 my_image_1_0.tar
+   $ hardlink -c images
 
-.. _OCI: https://github.com/opencontainers/image-spec/blob/master/spec.md
+A detailed procedure for generating the expected layout is available at
+`NicolasT/static-container-registry`_. You can use the script provided there
+to generate your own NGINX configuration and test this static registry for
+yourself. MetalK8s will generate its own configuration when importing the
+Solution archive.
+
+Each archive will be exposed as a single repository, where the name will be
+computed as ``<NAME>-<VERSION>`` from :ref:`solution-archive-product-info`.
+
+.. warning::
+
+   Operator Deployments should not rely on this naming pattern for finding
+   the images for their resources. Instead, the full repository prefix will be
+   stored in a ``ConfigMap``, that Deployments will be able to expose as
+   environment variables. See :doc:`./operator` for more details.
+
+The images names and tags will be inferred from the directory names chosen when
+using ``skopeo copy``. Using `hardlink` is highly recommended if one wants to
+define alias tags for a single image.
 
 MetalK8s also defines recommended standards for container images, described in
 :ref:`req-container-images`.
 
-In order for MetalK8s to populate its registry with accurate image tags,
-Solutions must provide a ``listing.yaml`` file under ``/images``, with the
-given format::
-
-   apiVersion: metalk8s.scality.com/v1alpha1
-   kind: ImagesList
-   solution: MySolution
-   images:
-     - archive: my_image_1_0.tar.gz
-       tag: my_image:1.0.0
-
-.. note::
-
-   Each Solution ISO will be made available to Pods as a single repository.
-   Operator Deployments will then be required to accept the repository prefix,
-   stored in a ``ConfigMap``, in order to properly configure the resources they
-   manage.
+.. _OCI: https://github.com/opencontainers/image-spec/blob/master/spec.md
+.. _nginx: https://www.nginx.com
+.. _skopeo: https://github.com/containers/skopeo
+.. _hardlink: http://man7.org/linux/man-pages//man1/hardlink.1.html
+.. _`NicolasT/static-container-registry`:
+   https://github.com/nicolast/static-container-registry
 
 Operator
 --------
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -91,6 +91,13 @@ It must respect the following format (currently version 1, as specified by the
    REQUIRE_METALK8S=">=2.0"
    ARCHIVE_LAYOUT_VERSION=1
 
+.. note::
+
+   If a Solution can require specific versions of MetalK8s on which to be
+   deployed, requiring specific services (and their respective versions) to be
+   shipped with MetalK8s (e.g. Prometheus/Grafana) is not yet feasible.
+   It will probably be handled in the Operator declaration, maybe using a CR.
+
 It is recommended for inspection purposes to include information related to
 the build-time conditions, such as the following (where command invocations
 should be statically replaced in the generated ``product.txt``)::
--- a/docs/developer/solutions/archive.rst
+++ b/docs/developer/solutions/archive.rst
@@ -91,6 +91,13 @@ It must respect the following format (currently version 1, as specified by the
    REQUIRE_METALK8S=">=2.0"
    ARCHIVE_LAYOUT_VERSION=1
 
+It is recommended for inspection purposes to include information related to
+the build-time conditions, such as the following (where command invocations
+should be statically replaced in the generated ``product.txt``)::
+
+   GIT=$(git describe --always --long --tags --dirty)
+   BUILD_TIMESTAMP=$(date +%Y-%m-%dT%H:%M:%SZ)
+
 .. note::
 
    If a Solution can require specific versions of MetalK8s on which to be
--- a/docs/developer/solutions/deployment.uml
+++ b/docs/developer/solutions/deployment.uml
@@ -1,0 +1,95 @@
+@startuml
+
+actor user as "User"
+control ui as "MetalK8s UI"
+control saltmaster as "Salt Master"
+entity bootstrap as "Bootstrap node"
+control apiserver as "Kubernetes API"
+control registry as "MetalK8s registry"
+
+== Import a new Solution (version) ==
+
+user -> bootstrap : Upload Solution ISO
+user -> bootstrap : Add ISO path to the "SolutionsConfiguration" file
+user -> saltmaster ++ : Request Solutions import
+
+saltmaster <-> apiserver : Retrieve "metalk8s-solutions" ConfigMap
+
+loop For each ISO defined in "SolutionsConfiguration"
+
+    saltmaster -> bootstrap ++ : Check ISO file (against our standards/constraints)
+    |||
+    bootstrap --> saltmaster -- : Return status (valid or not) and metadata if any
+
+    alt ISO is invalid
+        |||
+        saltmaster -> saltmaster : Fail early
+        |||
+    else ISO is valid
+        |||
+        saltmaster -> bootstrap ++ : Run "import_solution_archive" formula
+        |||
+        bootstrap -> bootstrap : Mount ISO
+        bootstrap -> registry : Configure new ISO source
+        |||
+        bootstrap -> saltmaster -- : Solution imported successfully
+    end
+
+end
+
+loop For each latest version newly imported
+
+    saltmaster <-> apiserver : Replace CRDs
+
+    saltmaster <-> apiserver : Create/Update Deployment for the Solution UI
+
+end
+
+loop For each Solution version in "metalk8s-solutions" ConfigMap not in "SolutionsConfiguration"
+    |||
+    saltmaster -> bootstrap ++ : Run "remove_solution_archive" formula
+    |||
+    bootstrap -> registry : Remove configuration for this Solution version
+    bootstrap -> bootstrap : Unmount ISO
+    |||
+    bootstrap -> saltmaster -- : Solution removed successfully
+end
+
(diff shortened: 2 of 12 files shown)


docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+   apiVersion: metalk8s.scality.com/v1alpha1
+   kind: ImagesList
+   solution: MySolution
+   images:
+     - archive: my_image_1_0.tar.gz
+       tag: my_image:1.0.0
+
+.. note::
+
+   Each Solution ISO will be made available to Pods as a single repository.
```
rachedbenmustapha: I don't really understand this paragraph. What I came to understand is that all images listed in all instances of `ImagesList` CR in a Solution will be sort of namespaced in a common "virtual" image registry repository (either by having a cluster-local registry, or by using unique prefixes when injecting the images in node's `containerd`s).

If that is the case then it answers a question that I had: what is the expected behavior if several solutions embed conflicting tags for the same image, for example `latest` or `8-latest`. Even if it is a practice that should be discouraged, it could still happen. Documenting the expected behavior (whether it's a problem or a non-problem) in case of conflict would be useful.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+
+In order for MetalK8s to populate its registry with accurate image tags,
+Solutions must provide a ``listing.yaml`` file under ``/images``, with the
+given format::
+
+   apiVersion: metalk8s.scality.com/v1alpha1
+   kind: ImagesList
+   solution: MySolution
+   images:
+     - archive: my_image_1_0.tar.gz
```
rachedbenmustapha: This format seems to actively discourage using `docker save` of multiple images at once. While this is perfectly fine, are we 100% set on disallowing it? Maybe it has changed since, but from my experience saving multiple images at once does efficient layer deduplication and gives better size after compression than compressing each image layer graph individually. The size and time benefits may not be worth the schema simplicity compromise, just wondering.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+       tag: my_image:1.0.0
+
+.. note::
+
+   Each Solution ISO will be made available to Pods as a single repository.
+   Operator Deployments will then be required to accept the repository prefix,
+   stored in a ``ConfigMap``, in order to properly configure the resources they
```
rachedbenmustapha: Does it mean that only select tailor-made operators with this `ConfigMap` support will work, and that one can't (at least yet) package an existing operator as a solution? May be worth a note if that's the case.

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+**MUST**:
+
+- Comply with the standard archive structure defined by MetalK8s
+- Expose management of its Operator instances in its UI
+- Handle monitoring of its own services (both Operator and application, except
+  the UI)
+
+**MAY**:
+
+- Use MetalK8s monitoring services (Prometheus and Grafana)
```
rachedbenmustapha, line 109: It may be worth taking a stronger stance on this. As a potential Solution developer I understand that I can use MetalK8s monitoring services for convenience, but I am not sure if it's a good (ie sustainable) idea, or if the more sustainable path is to embed my own prometheus stack.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+-------------------
+
+General product information about the packaged Solution must be stored in the
+``product.txt`` file, stored at the archive root.
+
+It must respect the following format::
+
+   NAME=${YOUR_SOLUTION_NAME}
+   VERSION=${YOUR_PROJECT_FULL_VERSION}
+
```
rachedbenmustapha, line 112: This could be a subject that we explicitly don't want to touch, but what about dependencies? While the average Solution could ideally be self-contained, some Solutions could have a need to depend on an already-deployed Solution and be non-functional without it. Examples of dependencies could be:
- A low-level storage backend that would be complex-enough to manage to warrant having its own Solution
- Some high-overhead services like a Kafka+ZooKeeper stack, or ELK stack, that would be best deployed once instead of once per solution that needs them

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+   Solutions can leverage the ``prometheus-operator`` CRs for setting up the
+   monitoring of their components.
+
+.. todo:: Define how Solutions can deploy Grafana dashboards.
+
+Interaction diagrams
+--------------------
+
+We include a detailed interaction sequence diagram for describing how MetalK8s
+will handle user input when deploying / upgrading Solutions.
```
rachedbenmustapha, line 125: Would be worth mentioning rollback (even if it's exactly the same as upgrading) for operations-minded readers.

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+.. todo:: Add specification in a future Reference guide
+
+Here is how it could look::
+
+    apiVersion: metalk8s.scality.com/v1alpha1
+    kind: SolutionsConfiguration
+    solutions:
+      - /solutions/storage_1.0.0.iso
+      - /solutions/storage_latest.iso
+      - /other_solutions/computing.iso
```
rachedbenmustapha, line 57: This example has the advantage of showing 2 versions of the same Solution, I was going to be ask about this case. It leaves me wondering how this case would work: is it designed for parallel run (either because 2 different versions will be needed indefinitely because of independent release tracks, or because we want to support gradual migration from version `v` to version `v+1`) or only one version active at a time that need to be selected/switched by the user through API/UI?

rachedbenmustapha (commented):
First thoughts

docs/developer/solutions/archive.rst
NicolasT: This part is outdated, we're currently integrating another approach to serve container images into MetalK8s (statically, using nothing but `nginx`), and 'solutions' shall use the same approach. This is *not* based on `docker save` but different tooling, and does allow for efficient deduplication (which, indeed, the approach currently described does not) by using `hardlink`(s).

gdemonet: Thanks for the clear answer @NicolasT . Indeed, this is mentioned in the [`todo` above](https://github.com/scality/metalk8s/pull/1020/files#diff-0c567a23053ae27cf85c625171c5d19aR97). Let's track this topic in #1049

docs/developer/solutions/archive.rst
gdemonet: > either by having a cluster-local registry, or by using unique prefixes when injecting the images in node's `containerd`s

There will be a cluster-local _registry_, which role will be to expose all images defined in `ImagesList` under a single _repository_. Let's clarify:

1. Assume I have a "Example v1.0" solution packaged, in which I have a `example:1-latest` image
2. After importing this Solution, MetalK8s will expose a new repository at a URL like `<registry_url>/example-1.0/`
3. The "Example v1.0 Operator" will then rely on this repository prefix to reference the image, so when creating an "Example" instance, use `image: <registry_url>/example-1.0/example:1-latest`

Now, let's assume I want to upgrade to "Example v1.1", which still include the `example:1-latest` image.

1. After importing this new version, I have a new repository at `<registry_url>/example-1.1/`
2. The existing Operator can thus reference either of those images, based on his own logic, still using the `example:1-latest` tag

So there would be no conflict w.r.t. tags. However, how does the Operator know which repository to use? See comment below for more details.

docs/developer/solutions/archive.rst
gdemonet: Operators need to know _where_ the images exist, so they can reference them in the various objects they create. What we suggest here is that we will define a `ConfigMap` to mount as an environment variable, for Operators to use as their `REPOSITORY_PREFIX`.

Let's take our "Example" solution described in the comment above:

- I have imported "Example v1.0", images are available in some repository
- I retrieve the full URL of the Operator image, including the repository (_Note: this is cumbersome, we may want to find a better way_)
- I deploy the Operator for this version, passing into the container spec:

```
image: {{ full_operator_image_url }}
[...]
env:
- name: REPOSITORY_PREFIX
    valueFrom:
    configMapKeyRef:
        name: metalk8s-solutions-repositories
        key: example-1.0
```

- The Operator container can now use this to generate its own resources, using `image: {{ REPOSITORY_PREFIX }}/example:1-latest`

Now, suppose (as in the comment above) that I have deployed version 1.1 of my Example solution. If I want to upgrade the 1.0 Operator, I can edit its Deployment, to use the new Operator image, and pass change the ConfigMap key to use, which will effectively upgrade all instances of `example:1-latest` deployed.

I know this all feels complicated, and thanks a lot for taking the time to review that part. We'll clarify these processes as soon as we know more (working on a _Hello-world Solution_ in the coming weeks should help us on that).

docs/developer/solutions/archive.rst
gdemonet: I agree that requesting external services / Solutions from a Solution has valid use-cases, and we may want to invest time looking at possible ways to manage this. I believe it could become a part of some "Operator of Operators" that we'd like to add in the future, for instance something like OLM from CoreOS.

I'll add a note in these docs, but I don't have a clear answer right now.

However, regarding common services, such as Kafka+Zookeeper or ELK, I believe those should be part of the original offering from MetalK8s. Thoughts @NicolasT?

docs/developer/solutions/introduction.rst
gdemonet: It's designed to run multiple versions of a Solution at the same time, using Namespaces for isolation. This allows users to try a new version in some `staging` namespace first, before upgrading their `prod` namespace.

docs/developer/solutions/introduction.rst
gdemonet: I thought there was no strong requirement, so moved it to _MAY_, but indeed it makes sense to force the usage of the cluster-provided Prometheus stack.

docs/developer/solutions/introduction.rst
gdemonet: We will add more in-depth guides for operational matters w.r.t Solutions, this one is for developers to understand how MetalK8s manages Solutions (which needs improvements as well, granted).

Most importantly, this diagram shows how deployment/upgrade is actually left to the user (manually or through the Solution UI). Rollback would fall into this category as well, IMHO.

Agreed though, we will need to show how a Solution gets **removed** from MetalK8s (which we won't do in the early stages, would be implemented in the generic `import_all_solutions` runner).

docs/developer/solutions/archive.rst
rachedbenmustapha: It's a lot clearer when explained this way :) I figured when looking at referenced issues/PRs but what got me was "what do ISOs have to do with Pods?"

docs/developer/solutions/archive.rst
rachedbenmustapha: Dependency handling is a big subject so it's ok as just a note for now.

If there are services to be provided by the platform, that the Solution expects (which ELK and Prometheus already are), maybe a Solution could mention it in its own manifest, as a capability name (for optional services, should that happen) or a minimum MetalK8s version (for services that will always be provided once introduced to the platform).

docs/developer/solutions/introduction.rst
rachedbenmustapha: I don't have a strong opinion either way, it's mostly about nudging developers in a "preferred" direction so that resources/storage capacity planning for the platform's Prometheus instance are adequate. As a Solution developer I would enjoy not having to manage my own Prom stack if I can trust that I won't have any surprises with retention duration etc :)

docs/developer/solutions/archive.rst
rachedbenmustapha: Once each operator knows where to get images it's pretty straightforward and it's good that it's "managed" by the deployer that way.

> I retrieve the full URL of the Operator image, including the repository (Note: this is cumbersome, we may want to find a better way)

For bootstrapping each Operator and making the initial manifests configurable, any opinions on `kustomize` / `kubectl apply -k`? It can be used [as an API](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/kustomization.go#L42) and allows replacing images (in addition to general patching) pretty easily as input data, leaving operator deployment manifests "clean". I can see how it only shifts one part of the problem elsewhere and does not entirely solve it though :)

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+detailed information.
+
+What is a *Solution*?
+---------------------
+
+We use the term *Solution* to describe a packaged Kubernetes application,
+archived as an ISO disk image, containing:
+
+- A set of OCI images to inject in MetalK8s image registry
+- An `Operator`_, to deploy on the cluster
```
charlesprost:
```suggestion
- An `Operator`_ to deploy on the cluster
```

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+Here is how it could look::
+
+    apiVersion: metalk8s.scality.com/v1alpha1
+    kind: SolutionsConfiguration
+    solutions:
+      - /solutions/storage_1.0.0.iso
+      - /solutions/storage_latest.iso
+      - /other_solutions/computing.iso
+
+There would be no explicit information about what the archives contain.
```
charlesprost:
```suggestion
There would be no explicit information about what an archive contains.
```

Since we use one archive in the line just after.

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,115 @@
...
+- Provide tooling to verify signatures in a Solution ISO
+- Expose management of Solutions in its own UI
+
+Solution
+^^^^^^^^
+
+**MUST**:
+
+- Comply with the standard archive structure defined by MetalK8s
+- Expose management of its Operator instances in its UI
```
charlesprost: A few sections above we state that the UI is optional for a solution, and in here we say that it MUST expose stuff in its UI.

docs/developer/solutions/archive.rst
gdemonet: I'll look into `kustomize`, thanks for the suggestion :)

docs/developer/solutions/operator.rst
```diff
@@ -0,0 +1,84 @@
...
+       ├── role.yaml
+       ├── role_binding.yaml
+       └── service_account.yaml
+
+Most of these files are generated when using the Operator SDK.
+
+.. todo::
+
+   Specify each of them, include example (after `#1060`_ is done).
+   Remember to note specifities about ``OCI_REPOSITORY_PREFIX`` / namespaces.
```
charlesprost:
```suggestion
   Remember to note specificities about ``OCI_REPOSITORY_PREFIX`` / namespaces.
```

docs/developer/solutions/introduction.rst
gdemonet: Would that be better?

```suggestion
- If providing a UI, expose management of its Operator instances
```

docs/developer/solutions/introduction.rst
gdemonet: I improved a few sections to be more explicit about this intent, later improvements will include UML diagrams for these use-cases.

docs/developer/solutions/introduction.rst
gdemonet: I added a warning for removal, and mentioned more details about how an Operator is deployed manually by the user.

docs/developer/solutions/archive.rst
gdemonet: I added a `REQUIRE_METALK8S` variable in the expected `product.txt`, which will allow for a minimal dependency management. Also added a note for handling more specific dependencies, such as monitoring components, TBD later.

docs/developer/solutions/introduction.rst
gdemonet: Changed `MAY` into `SHOULD`, although I don't want to say `MUST`, since deploying your own monitoring stack should not prevent the Solution from working (though I'm not sure if having multiple Prometheus Operators watching for `ServiceMonitor`s could pose issues...). I'll open a tracking ticket.

docs/developer/solutions/archive.rst
gdemonet: So, I looked into `kustomize`, it's really interesting. It doesn't seem to be available as part of `apiserver`, only inside `kubectl`, but it could be a good addition to the manifests provided in the ISO. Let's wait until we know more with #1060, I'll try it out.

I added a note to mention this in the docs.

docs/developer/solutions/archive.rst
gdemonet: I reworked these guidelines to match with what was introduced in #1049

docs/developer/solutions/archive.rst
gdemonet: I explained slightly more in the docs, though this will be detailed in the future with examples and more diagrams.

docs/developer/solutions/introduction.rst
gdemonet: Ok, latest conversations led to a simpler stance: Solutions that want to be deployed on MetalK8s `MUST` use the provided Prometheus Operator to set-up their monitoring. However, it is reasonable to assume Solution developers may want their Operator to work in any K8s cluster, hence could manage their own monitoring stack there. I'm still unsure about how they would do so, let's focus on _using_ MetalK8s stack for now (will consider the alternative use-case in #1158).

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

docs/developer/solutions/operator.rst
```diff
@@ -63,6 +63,12 @@ MetalK8s does not handle the monitoring of a Solution application, which means:
 - Operators should create ``Service`` and ``ServiceMonitor`` objects for each
   deployed component they own
 
+The `Prometheus Operator`_ deployed by MetalK8s will have cluster-scoped
+permissions, and will be able to read the aforementioned ``ServiceMonitor``
+objects to set-up monitoring of your application services.
```
slaperche-scality: s/set-up/setup/

>  Setup vs. Set-Up vs. Set Up
>
> The noun setup is usually styled as a solid compound (that is, as a single word) in American English and as a hyphenated compound (set-up) in British English. The noun may be used attributively (that is, as an adjective) in such phrases as setup man (“someone who makes machine tool adjustments”) and setup tool (“software that facilitates the installation of a computer program”). The verb set up, on the other hand, is usually found as an open compound (two words, no hyphen) in both American and British English.

docs/developer/solutions/archive.rst
```diff
@@ -85,8 +85,16 @@ General product information about the packaged Solution must be stored in the
 
 It must respect the following format::
 
-   NAME=${YOUR_SOLUTION_NAME}
-   VERSION=${YOUR_PROJECT_FULL_VERSION}
+   NAME=Example
```
slaperche-scality, line 94: in commit message "Solutions  to": two many spaces beteen the two words.

docs/developer/solutions/archive.rst
```diff
@@ -94,51 +98,64 @@ It must respect the following format::
 OCI images
 ----------
 
-.. todo::
+MetalK8s exposes container images in the OCI_ format through a static
+read-only registry. This registry is built with NGINX_, and relies on having
```
slaperche-scality: minor: s/NGINX/nginx/g

I have the impression that the uppercase is only used for the name of the company and the logo, when talking about the sofware it's almost always written as `nginx`.

docs/developer/solutions/archive.rst
```diff
@@ -94,51 +98,64 @@ It must respect the following format::
...
-     - archive: my_image_1_0.tar.gz
-       tag: my_image:1.0.0
+   Operator Deployments should not rely on this naming pattern for finding
+   the images for their resources. Instead, the full repository prefix will be
+   stored in a ``ConfigMap``, that Deployments will be able to expose as
+   environment variables. See :doc:`./operator` for more details.
 
-.. note::
+The images names and tags will be inferred from the directory names chosen when
+using ``skopeo copy``. Using `hardlink` is highly recommended if one wants to
```
slaperche-scality, line 181: >  Using `hardlink` is highly recommended if one wants to define alias tags for a single image.

I'm not sure to understand this part. IIUC, hardlinks are used to deduplicate the layers, whereas aliases are provided by some `nginx` config.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+
+
+Archive format
+--------------
+
+Solution archives must use the `ISO-9660:1988`_ format, including `Rock Ridge`_
+and Joliet_ directory records. The character encoding must be UTF-8_. The
+conformance level is expected to be at most 3, meaning:
+
+- Directory identifiers may not exceed 31 characters in length
```
slaperche-scality: What's a characters here: A byte (I suppose)? A Unicode codepoint? A grapheme cluster (highly unlikely)?

Since we're allowing/requiring UTF-8 here, a character is something ambiguous and may not match what people expect.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,152 @@
...
+   Clarify whether Joliet/Rock Ridge records supersede the conformance level
+   w.r.t. filename lengths
+
+
+.. _`ISO-9660:1988`: https://www.iso.org/obp/ui/#iso:std:iso:9660:ed-1:v1:en
+.. _`Rock Ridge`: https://en.wikipedia.org/wiki/Rock_Ridge
+.. _Joliet: https://en.wikipedia.org/wiki/Joliet_(file_system)
+.. _UTF-8: https://tools.ietf.org/html/rfc3629
+
+Here is an example invocation of the common Unix mkisofs_ tool to generate such
```
slaperche-scality, line 34: such an archive => such archive?

docs/developer/solutions/operator.rst
gdemonet: Mmh from what you quoted, it would be `set up`, no? Thanks for the info btw :)

docs/developer/solutions/archive.rst
gdemonet: OK, wasn't sure which one to use

docs/developer/solutions/operator.rst
slaperche-scality: Yup indeed

docs/developer/solutions/archive.rst
gdemonet: Done

docs/developer/solutions/archive.rst
gdemonet: Will the procedure above handle multiple tags for a single image? I thought tags were just derived from directories. Hence, if you want two tags for a single image, you would `skope copy` it twice, then `hardlink` to remove all duplicate layers.

docs/developer/solutions/archive.rst
gdemonet: Nothing clear from what I read so far, feel free to go through the [specs](http://www.ecma-international.org/publications/standards/Ecma-119.htm) ;). In the meantime, I'll mention `bytes`.

gdemonet: TBH, I just read through the docs and tried to translate the command we use in the buildchain into some text, but I have no clue if what I wrote is correct (we probably don't care much if we can mount and read from the ISO).

slaperche-scality: IIUC, ISO 9660 doesn't support Unicode characters so there is no issue (one character == one byte, hence the restrictions expressed in term of character).
Joliet use UCS-2 encoding (which doesn't support all the Unicode characters).

I guess that the UTF-8 is supported through Rock Ridge extensions.

Seems complicated, so let's use mention `bytes` as you suggest :D

docs/developer/solutions/archive.rst
slaperche-scality: Indeed, `nginx` aliases are something else and not related on what you were saying here. I got confused.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+-------------------
+
+General product information about the packaged Solution must be stored in the
+``product.txt`` file, stored at the archive root.
+
+It must respect the following format::
+
+   NAME=Example
+   VERSION=1.0.0-dev
+   REQUIRE_METALK8S=">=2.0"
```
NicolasT, line 96: Do we want to require the inclusion of `GIT`, `BUILD_TIMESTAMP` as done in MetalK8s proper?

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Once all your images were stored this way, you can de-duplicate layers using
+hardlinks, using the tool hardlink_::
+
+   $ hardlink -c images
+
+A detailed procedure for generating the expected layout is available at
+`NicolasT/static-container-registry`_. You can use the script provided there
+to generate your own NGINX configuration and test this static registry for
+yourself. MetalK8s will generate its own configuration when importing the
```
NicolasT: > MetalK8s will generate its own configuration when importing the Solution archive.

Is that so? In this case, we'll need to ship `static-container-registry` with MetalK8s... For MetalK8s proper, we *don't* generate the configuration at runtime, but at build-time...

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+Solution archive.
+
+Each archive will be exposed as a single repository, where the name will be
+computed as ``<NAME>-<VERSION>`` from :ref:`solution-archive-product-info`.
+
+.. warning::
+
+   Operator Deployments should not rely on this naming pattern for finding
+   the images for their resources. Instead, the full repository prefix will be
+   stored in a ``ConfigMap``, that Deployments will be able to expose as
```
NicolasT: Why should it be in a `ConfigMap`? That seems cumbersome?

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+   environment variables. See :doc:`./operator` for more details.
+
+The images names and tags will be inferred from the directory names chosen when
+using ``skopeo copy``. Using `hardlink` is highly recommended if one wants to
+define alias tags for a single image.
+
+MetalK8s also requires that container images respect specific standards, such
```
NicolasT: Does MetalK8s require this? :thinking:

docs/developer/solutions/deployment.uml
```diff
@@ -0,0 +1,89 @@
...
+control apiserver as "Kubernetes API"
+control registry as "MetalK8s registry"
+
+== Import a new Solution (version) ==
+
+user -> bootstrap : Upload Solution ISO
+user -> bootstrap : Add ISO path to the "SolutionsConfiguration" file
+user -> saltmaster ++ : Request Solutions import
+
+saltmaster <-> apiserver : Retrieve "MetalK8sSolutions" ConfigMap
```
NicolasT: In general, object names are not CamelCased in K8s-land.

docs/developer/solutions/deployment.uml
```diff
@@ -0,0 +1,89 @@
...
+    |||
+    bootstrap --> saltmaster -- : Return status (valid or not) and metadata if any
+
+    alt ISO is invalid
+        |||
+        saltmaster -> saltmaster : Fail early
+        |||
+    else ISO is valid and version is present in "MetalK8sSolutions" ConfigMap
+        |||
+        saltmaster -> saltmaster : Skip
```
NicolasT: Why skip? Should be a no-op, so why not just run it?

docs/developer/solutions/introduction.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Note that Solutions will be **imported** based on this file contents, i.e.
+the images they contain will be made available in the registry and the UI
+will be deployed, however **deploying** the Operator and subsequent
+application(s) is left to the user, through manual operations or the Solution
+UI.
+
+.. note::
+
+   Removing an archive path from the ``solutions`` list will effectively
```
NicolasT, line 71: This is not reflected in the UML diagram I believe?

docs/developer/solutions/operator.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Files
+^^^^^
+
+All Operator-related files except for the container images (see
+:ref:`solution-archive-images`) should be stored under ``/operator`` in the ISO
+archive. Those files should be organized as follows::
+
+   operator
+   ├── crd
```
NicolasT: Why not have `crds` (plural) inside `deploy`, as is done by `operator-sdk`?

docs/developer/solutions/operator.rst
```diff
@@ -0,0 +1,96 @@
...
+^^^^^^^^^^
+
+MetalK8s does not handle the monitoring of a Solution application, which means:
+
+- the user, manually or through the Solution UI, should create ``Service`` and
+  ``ServiceMonitor`` objects for each Operator instance
+- Operators should create ``Service`` and ``ServiceMonitor`` objects for each
+  deployed component they own
+
+The `Prometheus Operator`_ deployed by MetalK8s will have cluster-scoped
```
NicolasT: `has`, not `will have`.

docs/developer/solutions/operator.rst
```diff
@@ -0,0 +1,96 @@
...
+permissions, and will be able to read the aforementioned ``ServiceMonitor``
```
NicolasT: `is`

NicolasT (changes requested):

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+
+.. _`ISO-9660:1988`: https://www.iso.org/obp/ui/#iso:std:iso:9660:ed-1:v1:en
+.. _`Rock Ridge`: https://en.wikipedia.org/wiki/Rock_Ridge
+.. _Joliet: https://en.wikipedia.org/wiki/Joliet_(file_system)
+.. _UTF-8: https://tools.ietf.org/html/rfc3629
+
+Here is an example invocation of the common Unix mkisofs_ tool to generate such
+archive::
+
+   mkisofs
```
slaperche-scality, line 37: I was thinking, for this (and how to generate images with `skopeo`/`hardlink`), would it be worth to extract those "functions" of our build system into some sort of tools/library/whatever that then could be used by both our buildchain AND the solutions?

That would avoid the "reverse-engineer what our buildchain does, put it in the doc and take care of keeping both of them in sync" approach.

docs/developer/solutions/archive.rst
gdemonet: We can mention them as useful/recommended, but I don't think our tooling will use them, will it?

docs/developer/solutions/archive.rst
gdemonet: Yes, I don't see how Solutions could include their own without knowing where the ISO will get mounted in our `repositories` container. But maybe you have an idea?

docs/developer/solutions/archive.rst
gdemonet: Because it needs to be readable somewhere that is not Salt, if users want to know where their images are stored. I thought `ConfigMap`s could help provide a consistent storage destination such that mounting its value as an environment variable in Operator Deployments is merely about knowing the `ConfigMap` name and the key for this Solution.

Or do we know the naming of repositories will be stable/predictable and only need to pass the registry address (in which case we probably just need to read the right `Service` object)?

docs/developer/solutions/archive.rst
gdemonet: Dunno, we have a section in `requirements` that mentions OCI standards, I thought Solutions would fall in the same category. I can remove this section if you want.

docs/developer/solutions/deployment.uml
gdemonet: Oops, will change

docs/developer/solutions/introduction.rst
gdemonet: Indeed, I forgot to add it.

docs/developer/solutions/deployment.uml
gdemonet: I'll remove this distinction.

docs/developer/solutions/operator.rst
gdemonet: Because CRDs are deployed differently than the rest, but agreed, it would make sense to keep the whole structure as generated by Operator SDK.

docs/developer/solutions/archive.rst
gdemonet: Yes, providing shared tooling would be great. I think we'll know more what are the requirements once we start working on a minimal Solution implementation (#1060).

docs/developer/solutions/archive.rst
NicolasT: Not necessarily indeed.

docs/developer/solutions/archive.rst
```diff
@@ -0,0 +1,177 @@
...
+
+General product information about the packaged Solution must be stored in the
+``product.txt`` file, stored at the archive root.
+
+It must respect the following format::
+
+   NAME=Example
+   VERSION=1.0.0-dev
+   REQUIRE_METALK8S=">=2.0"
+
```
NicolasT, line 105: One thing must be added here: some kind of 'format version', such that we can change the ISO layout over time, and dispatch accordingly during ISO import (in the MetalK8s code).

docs/developer/solutions/archive.rst
NicolasT: That's why the storage path can be any random string, including... a variable, as explained in the `static-container-registry` `README`, and as used by MetalK8s as well.

docs/developer/solutions/archive.rst
NicolasT: Fair enough, though I don't think solutions should be aware of this `ConfigMap`, or refer to it in their deployment manifest. That's cumbersome. We can provide or inject that value when rendering the manifest directly.

docs/developer/solutions/archive.rst
NicolasT: Merely referring to 'using the same base image' and having specific **LABEL**s. That's not required.

docs/developer/solutions/operator.rst
NicolasT: Their deployment is indeed distinct, but I wouldn't translate into the ISO layout, and keep things the way Operator SDK generates them indeed.

NicolasT (dismissed):

sayf-eddine-scality (dismissed):

gdemonet:
@bert-e approve

sayf-eddine-scality (approved):

Issue #1024, opened 2019-04-19
Form: fix some error management bugs


carlito-scality:
-remove "isValid" flag since it is so buggy
-remove error message onChange
-improve the condition to disable "Submit" button

PR #1025, opened 2019-04-19
NodeCreateForm: error management bug fixs

carlito-scality:
Fix: https://github.com/scality/metalk8s/issues/1024

How to test:
-Fill any form input then empty it, an error should appear and the input is empty ;)
-Submit button is disabled until all field are set correctly

### DIFF
--- a/ui/src/containers/Login.js
+++ b/ui/src/containers/Login.js
@@ -7,6 +7,7 @@ import { brand, padding } from 'core-ui/dist/style/theme';
 import styled from 'styled-components';
 import { authenticateAction } from '../ducks/login';
 import { injectIntl } from 'react-intl';
+import { isEmpty } from 'lodash';
 
 const LoginFormContainer = styled.div`
   height: 100vh;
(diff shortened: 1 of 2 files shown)


ui/src/containers/Login.js
```diff
@@ -48,7 +49,24 @@ const LogoContainer = styled.div`
...
+    touched,
+    errors,
+    dirty,
+    intl,
+    setFieldValue,
+    setFieldTouched,
+    asyncErrors
+  } = props;
+  //handleChange of the Formik props does not update 'values' when field value is empty
+  const handleChange = (e, field) => {
```
gdemonet: You could make it nicer to use with currying:

```suggestion
  const handleChange = field => e => {
```

and below:

```javascript
<Input
  name="username"
  ...
  onChange={handleChange('username')}
/>
```

MonPote (approved):
LGTM, not directly related to the code but I created an issue to discuss if we really want to shake the input when there is an error. #1037

carlito-scality:
@bert-e approve

PR #1026, opened 2019-04-20
salt: don't calculate Pillar top based on cluster

NicolasT:
In 13485d2, changes were made to the way Pillar data is calculated, in an attempt to base this on the 'role' of a minion.

However, this implementation is flawed: instead of matching on the minion ID in the `metalk8s.nodes` structure, it matches on the existence of *any* minion in a specific role in the whole cluster, so de-facto all minions get the same computed Pillar data.

The fix is trivial: instead of matching on `*` in the Pillar glob match, only match on `grains.id` in `metalk8s.nodes`.

See: #1010
See: 13485d2

### DIFF
--- a/pillar/mine_functions.sls
+++ b/pillar/mine_functions.sls
@@ -1,9 +1,0 @@
-mine_functions:
-  control_plane_ip:
-    mine_function: grains.get
-    key: metalk8s:control_plane_ip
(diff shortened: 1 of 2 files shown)


NicolasT:
@bert-e approve

TeddyAndrieux (approved):

Issue #980, opened 2019-04-11
Create an etcd upgrade runner


TeddyAndrieux:
Create a runner that orchestrate the upgrade of all the etcd cluster by upgrading node one by one and checking/waiting for a good status between each nodes

```
Check_etcd_status
For_each etcd_node in etcd_cluster_nodes:
	Upgrade_static_manifest etcd_node
	Check_etcd_status
```

TeddyAndrieux:
Try to follow this kind of orchestrate https://gist.github.com/whiteinge/1bf3b1fa525c2e883b805f271ec6f7d7

Issue #934, opened 2019-04-04
Create a node Upgrade runner


TeddyAndrieux:
Create a salt runner that do the  workload of control plane node upgrade.

The salt runner should look at value from apiserver (though ext_pillar) to know which version of metalk8s **should** be present on a node.

If this is a newer version that the one already deployed run the highstate on this node to upgrade to the required version.

https://kubernetes.io/docs/tasks/administer-cluster/cluster-management/#maintenance-on-a-node

```
# Using salt external pillar reading the apiserver
Get new_version in node “labels” “metalk8s.scality.com/version”
 
Call cordon node				# Mark node as unschedulable
Call drain node				# gracefully terminate all pods on node
   
Call highstate node			        # Salt highstate

Call uncordon node				# Mark node as schedulable
```
To `drain` #979 is needed

gdemonet:
As mentioned in #1030 description (and discussed [here](https://github.com/scality/metalk8s/pull/1030#discussion_r278898634)), this runner should check that we are targeting a version lower than or equal to the version of the runner (i.e. to upgrade for X to Y, you need to use the runner from Y, not X).

TeddyAndrieux:
Sorry my bad but it's more a comment for #981 IMO have a kind of prechecks before starting upgrade (will add a comment)

Issue #896, opened 2019-03-31
Use Kubernetes 'service' objects for bootstrap service discovery


NicolasT:
We currently use a couple of hacks to find the addresses of services running on the bootstrap node (`salt-master`, registry,...).

Instead, we should

- [x] as part of the bootstrap process, create headless `Service` objects in the Kubernetes cluster which map to these services
- [ ] extend an `ext_pillar` to retrieve address(es) for these services and expose the information to the Salt states accordingly

The reason to use headless services: there's no real reason to assign a ClusterIP to these services because we won't have `kube-proxy` running on nodes at the time we need to be able to access these services.

NicolasT:
@ramanduh Here's an example for `salt-master`:

The object:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: salt-master
  labels:
    app: salt-master
    app.kubernetes.io/name: salt-master
    app.kubernetes.io/component: salt
    heritage: metalk8s
    app.kubernetes.io/part-of: metalk8s
    app.kubernetes.io/managed-by: salt
spec:
  clusterIP: None
  ports:
  - name: publisher
    port: 4505
    protocol: TCP
    targetPort: publisher
  - name: requestserver
    port: 4506
    protocol: TCP
    targetPort: requestserver
  selector:
    app.kubernetes.io/component: salt
    app.kubernetes.io/name: salt-master
  type: ClusterIP
```
Apply it:
```
[root@bootstrap ~]# kubectl -n kube-system apply -f salt-master-service.yaml
service/salt-master created
```

Check result:
```
[root@bootstrap ~]# kubectl -n kube-system describe svc salt-master
Name:              salt-master
Namespace:         kube-system
Labels:            app=salt-master
                   app.kubernetes.io/component=salt
                   app.kubernetes.io/managed-by=salt
                   app.kubernetes.io/name=salt-master
                   app.kubernetes.io/part-of=metalk8s
                   heritage=metalk8s
Annotations:       kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"app":"salt-master","app.kubernetes.io/component":"salt","app.kubernetes.io/...
Selector:          app.kubernetes.io/component=salt,app.kubernetes.io/name=salt-master
Type:              ClusterIP
IP:                None
Port:              publisher  4505/TCP
TargetPort:        publisher/TCP
Endpoints:         172.21.254.7:4505
[...]

Issue #878, opened 2019-03-28
Use python-etcd3 instead of call to etcdctl


slaperche-scality:
Instead of issuing call to the `etcdctl` binary, we should use the `etcd` API through  `python-etcd3`.

This would requires:
- provision an etcd client cert for salt-master, similar to how we'll create a kubeconfig file for it
- add the python-etcd3 package in the salt-master image
- create a Salt module to manage etcd members in a cluster, given the above client cert
- use this module to add the new node as a member, on the salt-master host
 
Source: https://github.com/scality/metalk8s/issues/738#issuecomment-477402323

TeddyAndrieux:
Also need to be replaced in etcd upgrade orchestrate 
https://github.com/scality/metalk8s/pull/1030#discussion_r280584127

PR #1030, opened 2019-04-24
Salt orchestrate for etcd upgrade

TeddyAndrieux:
Etcd upgrade and downgrade works:

Upgrade
```
salt-run state.orch metalk8s.orchestrate.upgrade.etcd saltenv=metalk8s-2.2 pillar="{'orchestrate': {'dest_version': '2.2'}, 'bootstrap_id': 'bootstrap'}"
```
Downgrade:
```
salt-run state.orch metalk8s.orchestrate.upgrade.etcd saltenv=metalk8s-2.2 pillar="{'orchestrate': {'dest_version': '2.1'}, 'bootstrap_id': 'bootstrap'}"
```

Note:
The saltenv `metalk8s-2.2` because we could maybe in the futur have action to remove thinks of this new version so we should always use the orchestrate of the newest version between dest_version and current version.

Fixes: #980

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -68,6 +68,40 @@ def get_etcd_endpoint():
         )
 
 
+def _execute_etcd_command(pod_name, cmd):
+    if isinstance(cmd, str):
+        cmd = [cmd]
+    try:
+        client = kubernetes.config.new_client_from_config(
+            config_file='/etc/kubernetes/admin.conf'
+        )
+    except:
+        log.exception('failed to load kubeconfig')
+        raise
+
+    api = kubernetes.client.CoreV1Api(api_client=client)
+    etcd_command = [
+        'etcdctl',
+        '--endpoints', 'https://localhost:2379',
+        '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
+        '--key-file', '/etc/kubernetes/pki/etcd/server.key',
+        '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
+    ] + cmd
+    try:
+        err = k8s_stream(
+            api.connect_get_namespaced_pod_exec,
+            name=pod_name, namespace='kube-system',
+            command=etcd_command,
+            stderr=True, stdin=False, stdout=False, tty=False
+        )
+        if err:
+            log.error('etcdctl error: %s', err)
+            raise CommandExecutionError('etcdctl: {}'.format(err))
+    except ApiException as exn:
+        log.exception('failed to run etcdctl')
+        raise exn
+
+
 @depends('KUBERNETES_PRESENT', fallback_function=deps_missing)
 def add_etcd_node(host, endpoint=None):
     '''Add a new `etcd` node into the `etcd` cluster.
@@ -86,36 +120,19 @@ def add_etcd_node(host, endpoint=None):
         endpoint = 'https://{0}:2380'.format(node_ip)
 
     pod_name = 'etcd-{}'.format(__salt__['network.get_hostname']())
-    try:
-        client = kubernetes.config.new_client_from_config(
-            config_file='/etc/kubernetes/admin.conf'
-        )
-    except:
-        log.exception('failed to load kubeconfig')
-        raise
+    _execute_etcd_command(pod_name, ['member', 'add', host, endpoint])
 
-    api = kubernetes.client.CoreV1Api(api_client=client)
-    etcd_command = [
-        'etcdctl',
-        '--endpoints', 'https://localhost:2379',
-        '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
-        '--key-file', '/etc/kubernetes/pki/etcd/server.key',
-        '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
-        'member', 'add', host, endpoint
-    ]
-    try:
-        err = k8s_stream(
-            api.connect_get_namespaced_pod_exec,
-            name=pod_name, namespace='kube-system',
-            command=etcd_command,
-            stderr=True, stdin=False, stdout=False, tty=False
-        )
-        if err:
-            log.error('etcdctl error: %s', err)
-            raise CommandExecutionError('etcdctl: {}'.format(err))
-    except ApiException as exn:
-        log.exception('failed to run etcdctl')
-        raise exn
+
+def check_etcd_health(hostname):
+    '''Check cluster-health of the `etcd` cluster.
+
+    This module is only runnable from the salt-master on the bootstrap node.
+
+    Arguments:
+        hostname (str): hostname of an etcd node
+    '''
+    pod_name = 'etcd-{}'.format(hostname)
+    _execute_etcd_command(pod_name, 'cluster-health')
 
 
 def format_san(names):
--- a/salt/metalk8s/kubernetes/etcd/init.sls
+++ b/salt/metalk8s/kubernetes/etcd/init.sls
@@ -5,6 +5,7 @@
 # ================
 #
 # * installed     -> deploy etcd manifest
+# * healthy       -> check health of etcd node
 #
 include:
   - .installed
--- a/salt/metalk8s/orchestrate/upgrade/etcd.sls
+++ b/salt/metalk8s/orchestrate/upgrade/etcd.sls
@@ -1,0 +1,44 @@
+{%- set dest_version = pillar.orchestrate.dest_version %}
+{%- set etcd_nodes = salt.metalk8s.minions_by_role('etcd') %}
+
+Check etcd cluster health:
+  module.run:
+    - metalk8s.check_etcd_health:
+      - hostname: {{ pillar.bootstrap_id }}
+
+{%- for node in etcd_nodes | sort %}
+
+Sync {{ node }} minion:
+  salt.function:
+    - name: saltutil.sync_all
+    - tgt: {{ node }}
+    - saltenv: metalk8s-{{ dest_version }}
+
+Upgrade etcd {{ node }} to {{ dest_version }}:
+  salt.state:
+    - tgt: {{ node }}
+    - sls:
+      - metalk8s.kubernetes.etcd.healthy
+    - saltenv: metalk8s-{{ dest_version }}
+    - require:
+      - salt: Sync {{ node }} minion
+      - module: Check etcd cluster health
+  {%- if previous_node is defined %}
+      - module: Check etcd cluster health for {{ previous_node }}
+  {%- endif %}
+
+Check etcd cluster health for {{ node }}:
+  module.run:
+    - metalk8s.check_etcd_health:
+      - hostname: {{ node }}
+    # FIXME: Can't retry because of https://github.com/saltstack/salt/issues/44639
+    # Should be ok for the moment as we check health in etcd.health state.
+    #- retry:
+    #    attempts: 5
+    - require:
+      - salt: Upgrade etcd {{ node }} to {{ dest_version }}
+
+  {#- Ugly but needed since we use have jinja2.7 (`loop.previtem` added in 2.10) #}
+  {%- set previous_node = node %}
+
+{%- endfor %}
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -145,6 +145,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/kubernetes/etcd/certs/peer.sls'),
     Path('salt/metalk8s/kubernetes/etcd/certs/server.sls'),
     Path('salt/metalk8s/kubernetes/etcd/files/manifest.yaml'),
+    Path('salt/metalk8s/kubernetes/etcd/healthy.sls'),
     Path('salt/metalk8s/kubernetes/etcd/init.sls'),
     Path('salt/metalk8s/kubernetes/etcd/installed.sls'),
 
@@ -190,6 +191,7 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/orchestrate/bootstrap/init.sls'),
     Path('salt/metalk8s/orchestrate/bootstrap/accept-minion.sls'),
     Path('salt/metalk8s/orchestrate/deploy_new_node.sls'),
+    Path('salt/metalk8s/orchestrate/upgrade/etcd.sls'),
 
     Path('salt/metalk8s/registry/deployed.sls'),
     Path('salt/metalk8s/registry/files/registry-manifest.yaml.j2'),
--- a/salt/metalk8s/kubernetes/etcd/healthy.sls
+++ b/salt/metalk8s/kubernetes/etcd/healthy.sls
@@ -1,0 +1,15 @@
+include:
+  - .installed
+
+Waiting for etcd running:
+  http.wait_for_successful_query:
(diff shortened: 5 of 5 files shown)


salt/metalk8s/kubernetes/etcd/init.sls
```diff
@@ -5,6 +5,7 @@
 # ================
 #
 # * installed     -> deploy etcd manifest
+# * health        -> check health of etcd node
```
TeddyAndrieux: Not part of `installed` or `init` because for expansions cluster is NOT healthly just after the manifest creation (as we need to execute the runner to add an etcd node first)

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,38 @@
+{%- set dest_version = pillar.orch.dest_version %}
+# TODO: Get the list of nodes waiting for #1028
```
NicolasT: It's merged.

salt/_modules/metalk8s.py
```diff
@@ -118,6 +102,39 @@ def add_etcd_node(host, endpoint=None):
...
+        endpoint (str): endpoint of the new etcd node
+    '''
+    if not endpoint:
+        # If we have no endpoint get it from mine
+        node_ip = __salt__['saltutil.runner'](
+            'mine.get', tgt=host, fun='control_plane_ip'
+        )[host]
+        endpoint = 'https://{0}:2380'.format(node_ip)
+
+    pod_name = 'etcd-{}'.format(__salt__['network.get_hostname']())
```
gdemonet, line 122: It's not your code, but I think this will always return the master's hostname (since we run this module from the master). Probably using `saltutil.runner` will work (as was done for `mine.get` just above).

TeddyAndrieux: Was intended here because we want to connect to the first etcd but yes btw we should not take the hostname of the master but the hostname of the first etcd node

NicolasT: Related: #878

TeddyAndrieux:
@bert-e reset

TeddyAndrieux:
@bert-e reset

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,52 @@
+{%- set dest_version = pillar.orchestrate.dest_version %}
```
gdemonet, line 1: From the PR description:

> we should always use the orchestrate of the newest version between dest_version and current version

Maybe we could implement that check in the `metalk8s.orchestrate.upgrade.<something>` runner(s)? Probably a good fit for a follow-up ticket.

salt/_modules/metalk8s.py
```diff
@@ -118,6 +102,39 @@ def add_etcd_node(host, endpoint=None):
...
+
+    This module is only runnable from the salt-master on the bootstrap node.
+
+    Arguments:
+        host (str): hostname of the new etcd node
+        endpoint (str): endpoint of the new etcd node
+    '''
+    if not endpoint:
+        # If we have no endpoint get it from mine
+        node_ip = __salt__['saltutil.runner'](
```
NicolasT, line 117: This is a `module`, so we shouldn't rely on runners IMHO.

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,52 @@
...
+        'host': control_plane_ip
+    },
+    'registry_ip': control_plane_ip
+} %}
+
+Check etcd cluster health:
+  module.run:
+    - metalk8s.check_etcd_health:
+      - hostname: {{ pillar.bootstrap_id }}
+    - failhard: True
```
NicolasT: I honestly don't like this. Would rather prefer to have a clear `require`-chain.

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,52 @@
...
+
+{%- for node in etcd_nodes %}
```
NicolasT: We should sort the nodes somehow.

Rationale: if one runs this playbook, and it fails, and for whatever reason the playbook is executed again without the cause for failure being fixed, then we want to fail at the same point, rather than potentially first upgrading other nodes that weren't upgraded before, to then at some point bump into the failing node.

salt/metalk8s/kubernetes/etcd/health.sls
```diff
@@ -0,0 +1,15 @@
+include:
+  - .installed
```
NicolasT: I think it's very strange for a `health` state to `include` `installed`.

For sure this may be renamed to `healthy`. Even then, I'm not sure it should `include` `installed`. We may want to be able to run this state as kind-of an 'assertion' without actually installing/upgrading the node. In the playbook, then apply both SLSs.

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,52 @@
...
+    - saltenv: metalk8s-{{ dest_version }}
+    - pillar: {{ pillar_data | tojson }}
+    - require:
+      - salt: Sync {{ node }} minion
+
+Check etcd cluster health for {{ node }}:
+  module.run:
+    - metalk8s.check_etcd_health:
+      - hostname: {{ node }}
+    - failhard: True
```
NicolasT: Same comment.

NicolasT (changes requested):

salt/_modules/metalk8s.py
TeddyAndrieux: Yes but if we call the classic mine.update then if we use this module on the master directly pretty sure that doesn't work. 
And also even if we run on a minion `salt-call saltuil.runner mine.get` works as exepected

salt/metalk8s/kubernetes/etcd/health.sls
TeddyAndrieux: For me it's more
- we want an etcd pod ok call `etcd.installed`
- we want a healthy etcd pod call `etcd.health(y)` 

For me having a state that only do a check it's strange should be a module at least. 
For me should have this step `Waiting for etcd running` directly in the state that install etcd but etcd may not work directly

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Hmm yes ok make sense  i can sort using node name

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Hmm it's harder to do a `require`-chain and also as a real point of view upgrade of one node not really require an upgrade of the previous node.

Also failhard made the output really clear here as if we have tones of nodes if we fail on the first one we will have a huge output with "One or more prequiriste states failed blabla` and only one with an error.

I can try to do a `require` chain but ... 

Edit: Or yes I agree for this one which is out of the `for` I should use a require I agree but for the last one at the end of the for I think it's better to keep the failhard

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: See above :)

salt/_modules/metalk8s.py
TeddyAndrieux: And btw this is only a workaround waiting for #877

TeddyAndrieux:
@bert-e reset

salt/_modules/metalk8s.py
```diff
@@ -118,6 +102,39 @@ def add_etcd_node(host, endpoint=None):
         raise exn
 
 
+@depends('KUBERNETES_PRESENT', fallback_function=deps_missing)
+def add_etcd_node(host, endpoint=None):
+    '''Add a new `etcd` node into the `etcd` cluster.
+
+    This module is only runnable from the salt-master on the bootstrap node.
```
gdemonet, line 109: Is it really? If yes, then maybe it's a good fit for a `runner` module?

gdemonet (commented):
A few notes, shouldn't block the merge. I'll let @NicolasT give the approval though

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: I think we should run like this but that can't be a strong requirement IMO

gdemonet: Disagree - it's a strong requirement, we cannot guarantee future versions won't need specific behaviour. And it's not that much of a disturbance for the end-user, he just needs to use the "good" Salt env, which he already has available

TeddyAndrieux: Yes agree with that but ok lets do this in another state because we need several checks like the fact that we only upgrade from one version, we need to check the version of every nodes but it's maybe more part of the global upgrade runner #934

salt/_modules/metalk8s.py
TeddyAndrieux: Already discussed here https://github.com/scality/metalk8s/pull/869/files#r270963853 :)

salt/metalk8s/orchestrate/upgrade/etcd.sls
gdemonet: Yes, let's do it in another ticket :)

salt/_modules/metalk8s.py
gdemonet: Thanks I forgot about this one :)

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,52 @@
...
+    - metalk8s.check_etcd_health:
+      - hostname: {{ pillar.bootstrap_id }}
+
+{%- for node in etcd_nodes | sort %}
+
+Sync {{ node }} minion:
+  salt.function:
+    - name: saltutil.sync_all
+    - tgt: {{ node }}
+    - saltenv: metalk8s-{{ dest_version }}
```
NicolasT, line 15: I think this should also require the node to be in 'ok' state (`require: ['module: Check etcd cluster health']`). There's no need to sync modules if we're not going to perform the upgrade in the next step anyway.

salt/metalk8s/orchestrate/upgrade/etcd.sls
NicolasT: Still not overly happy with it :P

NicolasT (dismissed):

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: For me it's not a requirements you don't need an healthy etcd cluster to sync modules and if it's just sync at a failing run at the next run he will basically do nothing

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Like I said in the other comment,

Upgrade of one node not really require an upgrade of the previous node.

Most of the case for me `require`-chain is better but for rolling upgrade a `failhard` it's something good no ?
"I'm not able to upgrade one node lets stop here" that make sense no ?

gdemonet: I think I agree with @TeddyAndrieux here: there is no "real" dependency between etcd node upgrades, they are (supposed to be) interchangeable, IIUC. Since it's just semantically about "run this batch in sequence, stop at first failure", I think `failhard: True` makes more sense.

TeddyAndrieux: As discussed directly with Nicolas and Guillaume let's do `require` chain to have a clean ordering and maybe do something smarter in the futur

TeddyAndrieux:
@bert-e reset

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Done

salt/metalk8s/orchestrate/upgrade/etcd.sls
gdemonet: I think that would still need to be done, since we use `require`-chain to define the order of execution, not just to define "real" requirements.

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,57 @@
...
+    'repo': {
+        'host': control_plane_ip
+    },
+    'registry_ip': control_plane_ip
+} %}
+
+Check etcd cluster health:
+  module.run:
+    - metalk8s.check_etcd_health:
+      - hostname: {{ pillar.bootstrap_id }}
```
NicolasT, line 7: Since this is a `runner`, which by definition runs on the Salt master, which is the 'bootstrap' node, can we replace this with the local hostname instead?

salt/metalk8s/orchestrate/upgrade/etcd.sls
```diff
@@ -0,0 +1,57 @@
+{%- set dest_version = pillar.orchestrate.dest_version %}
+{%- set etcd_nodes = salt.metalk8s.minions_by_role('etcd') %}
+{%- set control_plane_ip = salt.saltutil.runner(
+    'mine.get',
+    tgt=pillar.bootstrap_id,
+    fun='control_plane_ip'
+)[pillar.bootstrap_id] %}
+
+{%- set pillar_data = {
+    'repo': {
```
NicolasT: These values are now available in the existing pillar, could you copy them from there? Given that change and the potential improvement mentioned below, it'd remove the need to pass `bootstrap_id` to invoke this.

NicolasT (dismissed):

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Hmm I'm not really like because strictly it's not the hostname of the salt-master that we want. We want an hostname of one node with etcd.

NicolasT: Ugh. Really need to use the etcd Python API, then run this stuff *from the master*. Could you refer to this one in the relevant ticket as well?

salt/metalk8s/orchestrate/upgrade/etcd.sls
NicolasT: Ping @TeddyAndrieux?

TeddyAndrieux: I will remove then since #896 done

salt/metalk8s/orchestrate/upgrade/etcd.sls
TeddyAndrieux: Done

TeddyAndrieux:
@bert-e reset

NicolasT:
@TeddyAndrieux Is this good to go?

TeddyAndrieux:
> @TeddyAndrieux Is this good to go?

For my side yes depends on you now :stuck_out_tongue:

TeddyAndrieux:
@bert-e approve

NicolasT (approved):
Some concerns about updating version labels and such, we may want to reconsider this later. But let's move forward now.

PR #1035, opened 2019-04-26
salt: disable SSH host key checking

NicolasT:
For now, globally disable SSH host key validation. Later on, we can
implement per-node host key validation (e.g. by putting the key
fingerprint as a Node annotation).


slaperche-scality (approved):

NicolasT:
@bert-e approve

Issue #1041, opened 2019-04-30
Create a salt module that `cordon`/`uncordon` a node


cmonfort:
Relates to #979

PR #1043, opened 2019-04-30
Salt module to `cordon` `uncordon` a kubernetes node

TeddyAndrieux:
Adding a salt module + a salt state to `cordon` and `uncordon` a kubernetes node

Fixes: #1041

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -2304,3 +2304,72 @@ def node_set_taints(node_name, taints, **kwargs):
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
+
+
+def node_unschedulable(node_name, **kwargs):
+    '''
+    Return unschedulable value of the node identified by the specified name
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_unschedulable node_name="minikube"
+    '''
+    match = node(node_name, **kwargs)
+
+    if match is not None:
+        return match['spec']['unschedulable'] or False
+
+    return None
+
+
+def node_set_unschedulable(node_name, unschedulable, **kwargs):
+    '''
+    Update the unschedulable flag to the provided `unschedulable` value for the
+    node identified by the name `node_name`.
+
+    CLI Examples::
+
+        # To cordon a node
+        salt '*' kubernetes.node_set_unschedulable node_name="minikube" \
+            unschedulable=True
+
+        # To uncordon a node
+        salt '*' kubernetes.node_set_unschedulable node_name="minikube" \
+            unschedulable=False
+    '''
+    cfg = _setup_conn(**kwargs)
+    try:
+        api_instance = kubernetes.client.CoreV1Api()
+        body = {'spec': {'unschedulable': unschedulable}}
+        api_response = api_instance.patch_node(node_name, body)
+        return api_response.to_dict()
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception('Exception when calling CoreV1Api->patch_node')
+            raise CommandExecutionError(exc)
+    finally:
+        _cleanup(**cfg)
+
+
+def node_cordon(node_name, **kwargs):
+    '''
+    Cordon a node (set unschedulable to True)
+
+    CLI Examples::
+
+        salt '*' kubernetes.node_cordon  node_name="minikube"
+    '''
+    return node_set_unschedulable(node_name, True, **kwargs)
+
+
+def node_uncordon(node_name, **kwargs):
+    '''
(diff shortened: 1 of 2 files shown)


salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2304,3 +2304,65 @@ def node_set_taints(node_name, taints, **kwargs):
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
+
+
+def node_unschedulable(node_name, **kwargs):
```
NicolasT, line 2309: Either make this 'private' (prefix `_`), or add a docstring :)

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1505,3 +1505,79 @@ def customresourcedefinition_present(
             'new': res}
 
     return ret
+
+
+def node_cordon(
```
NicolasT: The *state* would be `node_cordoned` I guess?

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1505,3 +1505,79 @@ def customresourcedefinition_present(
...
+        name,
+        **kwargs):
```
NicolasT: Docstring.

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1505,3 +1505,79 @@ def customresourcedefinition_present(
...
+    ret['changes'] = {
+        'old': {'unschedulable': unschedulable},
+        'new': {'unschedulable': res['spec']['unschedulable']}
+    }
+    ret['comment'] = 'Node {0} cordoned'.format(name)
+
+    return ret
+
+
+def node_uncordon(
```
NicolasT: Same comments as with `node_cordon`.

Also, these functions are basically symmetrical, so would it make sense to factor this out, have one 'helper' (private) state which can then be invoked with a boolean that signals the 'intended' `unschedulable` value?

NicolasT (changes requested):

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1505,3 +1505,79 @@ def customresourcedefinition_present(
...
+    res = __salt__['metalk8s_kubernetes.node_uncordon'](
+        node_name=name,
+        **kwargs)
+
+    ret['result'] = True
+    ret['changes'] = {
+        'old': {'unschedulable': unschedulable},
+        'new': {'unschedulable': res['spec']['unschedulable']}
+    }
+    ret['comment'] = 'Node {0} cordoned'.format(name)
```
NicolasT: A good example of why factoring out (see comment above) makes sense: this is most likely copy-pasted, because the comment is incorrect for this state :wink:

TeddyAndrieux:
@bert-e reset

TeddyAndrieux:
@bert-e reset

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1505,3 +1505,70 @@ def customresourcedefinition_present(
...
+    action = 'cordon' if value else 'uncordon'
+
+    unschedulable = __salt__[
+        'metalk8s_kubernetes.node_unschedulable'](name, **kwargs)
+
+    if unschedulable is None:
+        ret['comment'] = 'Unable to get node {0}'.format(name)
+        return ret
+
+    if unschedulable is value:
```
NicolasT: Use `==`, not `is`. Yes, `True` and `False` are singletons, but that shouldn't be relied on.

NicolasT (changes requested):

salt/_states/metalk8s_kubernetes.py
slaperche-scality: +1

In Python 3, it's kinda safe because `True` and `False` are keywords and can't be changed:

```python
>>> True = 0
  File "<stdin>", line 1
SyntaxError: can't assign to keyword
```
But it's still better to use `==` in this case.

But in Python 2 that's not the case and funny things are possible and you can't rely on it:

```python
>>> True, False = False, True # now, let's the fun begin
>>> False
True
>>> True
False
>>> id(True)
140037444246128
>>> True = 42
>>> id(True)
94042690003080
```

NicolasT: In this case it's an actual comparison between two values (that happen to be Booleans) :)

slaperche-scality: Indeed, I edited my comment after seeing the context ;)

TeddyAndrieux: Hmm ok just use `is` to match only boolean but I don't care let's use `==`

TeddyAndrieux:
@bert-e reset

NicolasT (approved):

TeddyAndrieux:
@bert-e approve

PR #1055, opened 2019-05-02
Salt orchestrate for (single) node upgrade

### LINKED ISSUE #934: Create a node Upgrade runner

TeddyAndrieux:
```
salt-run state.orchestrate metalk8s.orchestrate.deploy_node saltenv=metalk8s-2.0 pillarenv=metalk8s-2.0 pillar="{'orchestrate': {'node_name': 'bootstrap'}}"
```

Fixes: #934

### DIFF
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -1,11 +1,15 @@
-{%- set version = pillar.metalk8s.nodes[pillar.node_name].version %}
+{%- set node_name = pillar.orchestrate.node_name %}
+{%- set version = pillar.metalk8s.nodes[node_name].version %}
 
-{%- if pillar['node_name'] not in salt['saltutil.runner']('manage.up') %}
+{%- set kubeconfig = "/etc/kubernetes/admin.conf" %}
+{%- set context = "kubernetes-admin@kubernetes" %}
+
+{%- if node_name not in salt.saltutil.runner('manage.up') %}
 Deploy salt-minion on a new node:
   salt.state:
     - ssh: true
     - roster: kubernetes
-    - tgt: {{ pillar['node_name'] }}
+    - tgt: {{ node_name }}
     - saltenv: metalk8s-{{ version }}
     - sls:
       - metalk8s.roles.minion
@@ -14,23 +18,25 @@ Accept key:
   module.run:
     - saltutil.wheel:
       - key.accept
-      - {{ pillar['node_name'] }}
+      - {{ node_name }}
     - require:
       - salt: Deploy salt-minion on a new node
 
 Wait minion available:
   salt.runner:
     - name: metalk8s_saltutil.wait_minions
-    - tgt: {{ pillar['node_name'] }}
+    - tgt: {{ node_name }}
     - require:
       - module: Accept key
     - require_in:
       - salt: Set grains
+      - salt: Refresh the mine
+      - salt: Cordon the node
 {%- endif %}
 
 Set grains:
   salt.state:
-    - tgt: {{ pillar['node_name'] }}
+    - tgt: {{ node_name }}
     - saltenv: metalk8s-{{ version }}
     - sls:
       - metalk8s.node.grains
@@ -40,31 +46,48 @@ Refresh the mine:
     - name: mine.update
     - tgt: '*'
 
+Cordon the node:
+  metalk8s_kubernetes.node_cordoned:
+    - name: {{ node_name }}
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+
+# TODO: Call drain
+
 Run the highstate:
   salt.state:
-    - tgt: {{ pillar['node_name'] }}
+    - tgt: {{ node_name }}
     - highstate: True
     - require:
       - salt: Set grains
       - salt: Refresh the mine
+      - metalk8s_kubernetes: Cordon the node
+
+Uncordon the node:
+  metalk8s_kubernetes.node_uncordoned:
+    - name: {{ node_name }}
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+    - require:
+      - salt: Run the highstate
 
 # Work-around for https://github.com/scality/metalk8s/pull/1028
 Kill kube-controller-manager on all master nodes:
   salt.function:
     - name: ps.pkill
-    - tgt: "{{ salt['metalk8s.minions_by_role']('master') | join(',') }}"
+    - tgt: "{{ salt.metalk8s.minions_by_role('master') | join(',') }}"
     - tgt_type: list
     - kwarg:
         pattern: kube-controller-manager
     - require:
       - salt: Run the highstate
 
-{%- if 'etcd' in pillar.get('metalk8s', {}).get('nodes', {}).get(pillar['node_name'], {}).get('roles', []) %}
+{%- if 'etcd' in pillar.get('metalk8s', {}).get('nodes', {}).get(node_name, {}).get('roles', []) %}
 
 Register the node into etcd cluster:
   module.run:
     - metalk8s.add_etcd_node:
-      - host: {{ pillar['node_name'] }}
+      - host: {{ node_name }}
     - require:
       - salt: Run the highstate
 
--- a/ui/src/services/api.js
+++ b/ui/src/services/api.js
@@ -129,11 +129,10 @@ export async function deployNode(url, token, node) {
       {
         client: 'runner',
         fun: 'state.orchestrate',
-        arg: ['metalk8s.orchestrate.deploy_new_node'],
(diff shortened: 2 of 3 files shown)


salt/metalk8s/orchestrate/upgrade/node.sls
```diff
@@ -0,0 +1,61 @@
+{%- set dest_version = pillar.orchestrate.dest_version %}
+{%- set node_name = pillar.orchestrate.node_name %}
+
+{%- set kubeconfig = "/etc/kubernetes/admin.conf" %}
+{%- set context = "kubernetes-admin@kubernetes" %}
+
+{%- if pillar.metalk8s.nodes[node_name]['version'] == dest_version %}
+
+Node {{ node_name }} already in version {{ dest_version }}:
```
NicolasT: Shouldn't we simply call this `deploy_node` or something, and always run everything?

salt/metalk8s/orchestrate/upgrade/node.sls
```diff
@@ -0,0 +1,61 @@
...
+
+Cordon node {{ node_name }}:
+  metalk8s_kubernetes.node_cordoned:
+    - name: {{ node_name }}
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+
+# TODO: Call drain
+
+Run the highstate on {{ node_name }}:
```
NicolasT: I think we should *first* set the version label, *then* cordon/drain/apply highstate/uncordon, and stop.

NicolasT: In which case the pillar override wouldn't be required.

salt/metalk8s/orchestrate/upgrade/node.sls
```diff
@@ -0,0 +1,61 @@
+{%- set dest_version = pillar.orchestrate.dest_version %}
```
NicolasT: Can't this be basically the same as, or almost the same, as `deploy_new_node.sls`?

NicolasT: I.e. don't have two playbooks.

NicolasT (changes requested):

NicolasT:
What I'm trying to say in the above: do we really need a 'single node upgrade' playbook?

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: Hmm for me we should set the version label only when the node is really running this version so at the end. 
And then we not execute anything on node which are already in the correct version and if we got an error during drain cordon highstate or uncordon we don't change the label version.

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: You mean even for deploy a node we will drain cordon and uncordon ?

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: I create one for single as we will use this for control_plane upgrade and for workload_plane_upgrade 
And then have something like that for upgrade global runner
```
└> cat salt/metalk8s/orchestrate/upgrade/init.sls 
Execute the upgrade prechecks:
  salt.runner:
    - name: state.orchestrate
    - mods:
      - metalk8s.orchestrate.upgrade.precheck

Upgrade etcd cluster:
  salt.runner:
    - name: state.orchestrate
    - mods:
      - metalk8s.orchestrate.upgrade.etcd
    - saltenv: {{ saltenv }}

Upgrade control_plane nodes:
  salt.runner:
    - name: state.orchestrate
    - mods:
      - metalk8s.orchestrate.upgrade.control_plane
    - require:
      - salt: Upgrade etcd cluster

Upgrade workload_plane nodes:
  salt.runner:
    - name: state.orchestrate
    - mods:
      - metalk8s.orchestrate.upgrade.workload_plane
    - require:
      - salt: Upgrade control_plane nodes
```
Not finished not tested of course but that my idea

NicolasT: Hmh, not sure that's the best approach. If we were to do a bit more in this playbook, and basically not use any 'sub'-runners, we can more clearly decide which nodes get upgraded in which stage. As an example, there's no need to bring a control-plane node in highstate as part of the *Upgrade workload-plane nodes*, when they've been put in highstate already.

salt/metalk8s/orchestrate/upgrade/node.sls
NicolasT: Yeah, why not? Would basically be a no-op.

salt/metalk8s/orchestrate/upgrade/node.sls
NicolasT: No. We're intent-driven, Kubernetes-style. We set the 'desired' version, then we apply actions (as generic as possible) to make that 'desire' reality.

NicolasT:
Maybe we should discuss more, but I'm still not sure we *need* an orchestration like this. Could have a single playbook that does a whole cluster upgrade. Not sure we even want to be able to upgrade a cluster 'partially', which could result in a broken system with incompatible versions.

Except for etcd, maybe.

salt/metalk8s/orchestrate/upgrade/node.sls
NicolasT: Could we have a short session today to discuss this and the path forward, @TeddyAndrieux @gdemonet ?

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: Edit: Oh for deploy, hmm yes maybe we can

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: Ok following the discussion we had, lets
- have one orchestrate for all the cluster upgrade
- set the version before upgrading a node
- always do cordon, drain, uncordon

TeddyAndrieux:
@bert-e reset

salt/metalk8s/orchestrate/upgrade/node.sls
TeddyAndrieux: This PR is only about one node, I will do the global orchestrate in another PR

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -40,31 +44,46 @@ Refresh the mine:
...
     - require:
       - salt: Set grains
       - salt: Refresh the mine
+      - metalk8s_kubernetes: Cordon the node
+
+Uncordon the node:
+  metalk8s_kubernetes.node_uncordoned:
+    - name: {{ node_name }}
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
```
NicolasT, line 70: Wouldn't this `require` `Run the highstate`?

ui/src/services/api.js
```diff
@@ -129,11 +129,10 @@ export async function deployNode(url, token, node) {
       {
         client: 'runner',
         fun: 'state.orchestrate',
-        arg: ['metalk8s.orchestrate.deploy_new_node'],
+        arg: ['metalk8s.orchestrate.deploy_node'],
         kwarg: {
           saltenv: 'metalk8s-2.0',
```
NicolasT, line 134: @carlito-scality @MonPote Is there a tracking ticket to get rid of this hard-coded value?

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: True, I just forgot this

TeddyAndrieux:
@bert-e reset

NicolasT (approved):

TeddyAndrieux:
@bert-e approve

ui/src/services/api.js
carlito-scality: @NicolasT See https://github.com/scality/metalk8s/issues/1105

PR #1058, opened 2019-05-02
salt: match at least one minion in `wait_minions`

NicolasT:
Right now, the `metalk8s_saltutil.wait_minions` *runner* can 'succeed'
with

```
No minions matched the target. No command was sent, no jid was assigned
```

which is of course not a valid condition.

Instead, wait until at least one minion result is returned. This may
indeed not match *all* expected minions, but we only use the *runner*
with a single target.

### DIFF
--- a/salt/_runners/metalk8s_saltutil.py
+++ b/salt/_runners/metalk8s_saltutil.py
@@ -23,7 +23,17 @@ def wait_minions(tgt='*', retry=10):
     minions = client.cmd(tgt, 'test.ping', timeout=2)
     attempts = 1
 
-    while not all(status for status in minions.values()) and attempts < retry:
+    def condition_reached(minions, attempts):
+        if attempts > retry:
+            return True
+
+        if minions and all(status for status in minions.values()):
+            return True
+
+        # Either `minions` is empty, or not all succeeded
+        return False
+
+    while not condition_reached(minions, attempts):
         log.info(
             "[Attempt %d/%d] Waiting for minions to respond: %s",
             attempts,
@@ -35,7 +45,7 @@ def wait_minions(tgt='*', retry=10):
         minions = client.cmd(tgt, 'test.ping', timeout=2)
         attempts += 1
 
-    if not all(status for status in minions.values()):
+    if not minions or not all(status for status in minions.values()):
         error_message = (
             'Minion{plural} failed to respond after {retry} retries: {minions}'
         ).format(


salt/_runners/metalk8s_saltutil.py
```diff
@@ -23,7 +23,19 @@ def wait_minions(tgt='*', retry=10):
...
+        if not minions:
+            return True
+
+        if not all(status for status in minions.values()):
+            return True
+
+        if attempts > retry:
+            return False
+
+        return False
```
slaperche-scality, line 34: Should be `return True` no?

Unless I'm mistaken, the current code always `return False` no matter the value of `attempts` and `retry`.

Also, should probably be `attempts >= retry` to mirror the old version that was using `attempts < retry` otherwise the behavior change when `retry == attempts`.

slaperche-scality (changes requested):

salt/_runners/metalk8s_saltutil.py
NicolasT: I've been confusing myself here :P Let me take a fresh look.

salt/_runners/metalk8s_saltutil.py
```diff
@@ -23,7 +23,19 @@ def wait_minions(tgt='*', retry=10):
...
 
-    while not all(status for status in minions.values()) and attempts < retry:
+    def keep_trying(minions, attempts):
+        if not minions:
+            return True
+
+        if not all(status for status in minions.values()):
+            return True
+
+        if attempts > retry:
```
gdemonet:
```suggestion
        if attempts < retry:
```

salt/_runners/metalk8s_saltutil.py
```diff
@@ -23,7 +23,19 @@ def wait_minions(tgt='*', retry=10):
...
+            return False
```
gdemonet:
```suggestion
            return True
```

salt/_runners/metalk8s_saltutil.py
NicolasT: I rewrote it, please check again :)

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

salt/_runners/metalk8s_saltutil.py
slaperche-scality: It's better indeed :)

But we still have a subtle change, no? When `retry == attempts` old code would break out of the loop, whereas here we'll return `False` and thus iterate again.

NicolasT: Agree, but it depends on how you define 'retry'. I'd argue the original code was 'wrong', as in: it tried once, and then `retry - 1` times. So there were no `retry` **re**tries.

slaperche-scality: Works for me then :)

slaperche-scality (approved):

NicolasT:
@bert-e approve

gdemonet (approved):
LGTM

PR #1065, opened 2019-05-02
Add a CLI utility

NicolasT:


NicolasT:
@bert-e wait

NicolasT:
This is on-hold.

I think it'd make sense to rewrite this as a `kubectl` plugin.

PR #1040, opened 2019-04-29
bugfix: fix storage validation

ssalaues:
This commit fixes a bug that causes the Ansible installer to fail at the
`setup_lvm_vg` step when trying to validate any specified partition.

See https://github.com/scality/metalk8s/issues/1039

Issue #1039, opened 2019-04-29
Install using a partition fails on 1.x 


ssalaues:
The `setup_lvm_vg` role fails when specifying a partition instead of raw block device

Simple fix is to return `False, None` here:
https://github.com/scality/metalk8s/blob/development/1.3/roles/setup_lvm_vg/action_plugins/validate_storage.py#L135

Steps to reproduce:
1. Define a storage definition pointing to a partition in kube-node.yml:
```
$ cat inventory/group_vars/kube-node.yml
metalk8s_lvm_drives_vg_metalk8s: ['/dev/vdb1']

$ ansible-playbook -i inventory/hosts -b playbooks/deploy.yml
...
TASK [setup_lvm_vg : LVM Setup: Check the storage configuration] *****
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'bool' object is not iterable
fatal: [node-01]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}

NO MORE HOSTS LEFT *****
        to retry, use: --limit @/home/centos/metalk8s/playbooks/deploy.retry

PLAY RECAP *****
localhost                  : ok=1    changed=0    unreachable=0    failed=0
node-01                    : ok=113  changed=27   unreachable=0    failed=1
```

NicolasT:
I guess this can now be closed, @ssalaues ?

Issue #1053, opened 2019-05-01
Kubernetes 1.11 CVE fixes


ssalaues:
These should be applied to the 1.1 branch before release. 
v1.11.7
- Fixes Update to use go1.10.7 with fix for CVE-2018-16875 (#72074)

v1.11.5 (Already applied to the 1.1 branch)
- CVE-2018-1002105: Fix critical security issue in kube-apiserver upgrade request proxy handler (#71411)

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#changelog-since-v1115

I'll open a PR to update to v1.11.7 however since v1.11.9 is the latest and I can just as easily create my PR to update to the latest 1.11 version. Thoughts @NicolasT

gdemonet:
Relates to #902.

NicolasT:
Go for 1.11 latest :)

PR #1066, opened 2019-05-02
1.1 Forward Port - bugfix: fix storage validation

ssalaues:
Forward port from dev/1.0 https://github.com/scality/metalk8s/pull/1040
> This commit fixes a bug that causes the Ansible installer to fail at the
> `setup_lvm_vg` step when trying to validate any specified partition
> 
> see #1039


ssalaues:
@bert-e approve

NicolasT:
In general we don't forward-port individual commits, but `development/*` branches. I'll set up some PRs to do so later today, also taking #1053 into account :)

ssalaues:
@NicolasT Ahh ok I will close these forward port PRs in that case

PR #1068, opened 2019-05-02
1.3 Forward Port - bugfix: fix storage validation

ssalaues:
Forward port from dev/1.0 https://github.com/scality/metalk8s/pull/1040
> This commit fixes a bug that causes the Ansible installer to fail at the
> `setup_lvm_vg` step when trying to validate any specified partition
> 
> see #1039


ssalaues:
Closing PR see https://github.com/scality/metalk8s/pull/1066#issuecomment-489559473

PR #1080, opened 2019-05-06
Store digests of config files in static Pods annotations

### LINKED ISSUE #881: Improve "bootstrap re-run" test
Currently, the test scenario is:

```
Given bootstrap was run once
When we run bootstrap a second time
Then the Kubernetes API is available
```

This means the test execution will run the bootstrap twice, even though the test is run after an initial bootstrap was already run.
An acceptable solution, since we already mark this test as `@post`, would be to only run it in the `When`:

```diff
-Given bootstrap was run once
+Given the Kubernetes API is available
 When we run bootstrap a second time
 Then the Kubernetes API is available
```

### LINKED ISSUE #937: Restart static Pods when configuration changes
Right now, when during (re)deployment e.g. `/etc/salt/master.d/99-metalk8s.conf` is changed, the Salt Master service is not restarted.

To fix this, we should, for every service running as a static Pod which doesn't pick up file changes dynamically:

- Add annotations in the Pod manifest containing checksums for all relevant input files
- Update the manifest when input files change

At this point, `kubelet` will detect the change in the Pod manifest and recreate the Pod, effectively restarting the container/service.

gdemonet:
**Component**: salt, kubernetes
**Type**: bugfix

**Context**:

- In order to have `kubelet` restart static Pod containers on configuration changes, we need to actually change the static Pod manifest

**Summary**:

- Introduce a basic `metalk8s.static_pod_managed` helper state to compute config files digests at runtime (so we can still rely on Salt `require`/`watch` ordering to lazily evaluate the hashes)
    - Hashing algorithm is basically: `md5 ( '-'.join( sha256(file for file in config_files) )`

- Use it for Salt master
- Use it for K8s control plane components (`k-a`, `k-s`, `k-c-m`) and `etcd`
- Use it for the package repositories and OCI registry

- Each static pod includes **every** single file that it wants to read (certificates, private keys, kubeconfig, etc.)

**Acceptance criteria**:

- On changes to the Salt master configuration files (**made using Salt**), `salt-master` Pod should be "restarted" by `kubelet` (actually replaced)
- On changes to other config files (e.g. certificates, kubeconfigs), related Pods should be "restarted"

**Notes**:

When trying to make this PR get a green build, I ran into many CI issues linked to "not waiting long enough" in tests. In the process of fixing this, a few modifications around tests were made:

- Pytest markers are now documented and used by default: see a4f1f09ae0ff04e78c6c59fe0becfe23f4ddf679
- The "bootstrap idempotent" test only runs the bootstrap once, since it's a `@post` test, and is now marked as `@slow`, which runs after the other tests in the CI (and can be filtered out easily with `tox -e tests -- -m "not slow"`)
- A number of tests now wait (through multiple retries) much longer than they used to before attempting operations on `Running` Pods

---

Fixes: GH-937
Fixes: GH-881

### DIFF
--- a/salt/_states/metalk8s.py
+++ b/salt/_states/metalk8s.py
@@ -1,0 +1,35 @@
+"""Custom states for MetalK8s."""
+
+__virtualname__ = "metalk8s"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def static_pod_managed(name, source, config_files=None, context=None):
+    """Simple helper to edit a static Pod manifest if configuration changes.
+
+    Expects the template to use the `config_digest` variable and store it in
+    the `metadata.annotations` section, with the key
+    `metalk8s.scality.com/config-digest`.
+    """
+    config_file_digests = [
+        __salt__["hashutil.digest_file"](config_file, checksum="sha256")
+        for config_file in config_files or []
+    ]
+    config_digest = __salt__["hashutil.md5_digest"](
+        "-".join(config_file_digests)
+    )
+
+    return __states__["file.managed"](
+        name,
+        source,
+        template="jinja",
+        user="root",
+        group="root",
+        mode="0600",
+        makedirs=False,
+        backup=False,
+        context=dict(context or {}, config_digest=config_digest),
+    )
--- a/salt/metalk8s/kubernetes/apiserver/installed.sls
+++ b/salt/metalk8s/kubernetes/apiserver/installed.sls
@@ -80,15 +80,27 @@ Set up default basic auth htpasswd:
 {%- set host = grains['metalk8s']['control_plane_ip'] %}
 
 Create kube-apiserver Pod manifest:
-  file.managed:
+  metalk8s.static_pod_managed:
     - name: /etc/kubernetes/manifests/kube-apiserver.yaml
     - source: salt://metalk8s/kubernetes/files/control-plane-manifest.yaml
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+        - /etc/kubernetes/pki/apiserver.crt
+        - /etc/kubernetes/pki/apiserver.key
+        - /etc/kubernetes/pki/apiserver-etcd-client.crt
+        - /etc/kubernetes/pki/apiserver-etcd-client.key
+        - /etc/kubernetes/pki/apiserver-kubelet-client.crt
+        - /etc/kubernetes/pki/apiserver-kubelet-client.key
+        - /etc/kubernetes/pki/ca.crt
+        - /etc/kubernetes/pki/etcd/ca.crt
+        - /etc/kubernetes/pki/front-proxy-ca.crt
+        - /etc/kubernetes/pki/front-proxy-client.crt
+        - /etc/kubernetes/pki/front-proxy-client.key
+        - /etc/kubernetes/pki/sa.pub
+        - {{ htpasswd_path }}
+{%- if pillar.metalk8s.api_server.keepalived.enabled %}
+        - /etc/keepalived/check-apiserver.sh
+        - /etc/keepalived/keepalived.conf.sh
+{%- endif %}
     - context:
         name: kube-apiserver
         host: {{ host }}
@@ -192,12 +204,16 @@ Create kube-apiserver Pod manifest:
             emptyDir:
               medium: Memory
 {%- endif %}
-    - require:
+    - onchanges:
       - file: Ensure kubernetes CA cert is present
       - file: Ensure etcd CA cert is present
       - file: Ensure front-proxy CA cert is present
       - file: Ensure SA pub key is present
       - file: Set up default basic auth htpasswd
+{%- if pillar.metalk8s.api_server.keepalived.enabled %}
+      - file: Create keepalived check script
+      - file: Create keepalived configuration file generator
+{%- endif %}
 
 Make sure kube-apiserver container is up:
   module.wait:
@@ -205,4 +221,4 @@ Make sure kube-apiserver container is up:
       - name: kube-apiserver
       - state: running
     - watch:
-      - file: Create kube-apiserver Pod manifest
+      - metalk8s: Create kube-apiserver Pod manifest
--- a/salt/metalk8s/kubernetes/controller-manager/installed.sls
+++ b/salt/metalk8s/kubernetes/controller-manager/installed.sls
@@ -5,15 +5,13 @@ include:
   - .kubeconfig
 
 Create kube-controller-manager Pod manifest:
-  file.managed:
+  metalk8s.static_pod_managed:
     - name: /etc/kubernetes/manifests/kube-controller-manager.yaml
     - source: salt://metalk8s/kubernetes/files/control-plane-manifest.yaml
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+        - /etc/kubernetes/controller-manager.conf
+        - /etc/kubernetes/pki/ca.crt
+        - /etc/kubernetes/pki/sa.key
     - context:
         name: kube-controller-manager
         image_name: {{ kubernetes_image("kube-controller-manager") }}
@@ -45,5 +43,5 @@ Create kube-controller-manager Pod manifest:
           - path: /etc/kubernetes/controller-manager.conf
             name: kubeconfig
             type: File
-    - require:
+    - onchanges:
       - metalk8s_kubeconfig: Create kubeconfig file for controller-manager
--- a/salt/metalk8s/kubernetes/files/control-plane-manifest.yaml
+++ b/salt/metalk8s/kubernetes/files/control-plane-manifest.yaml
@@ -3,6 +3,7 @@ kind: Pod
 metadata:
   annotations:
     scheduler.alpha.kubernetes.io/critical-pod: ""
+    metalk8s.scality.com/config-digest: "{{ config_digest }}"
   creationTimestamp: null
   name: {{ name }}
   namespace: "kube-system"
--- a/salt/metalk8s/kubernetes/scheduler/installed.sls
+++ b/salt/metalk8s/kubernetes/scheduler/installed.sls
@@ -4,15 +4,11 @@ include:
   - .kubeconfig
 
 Create kube-scheduler Pod manifest:
-  file.managed:
+  metalk8s.static_pod_managed:
     - name: /etc/kubernetes/manifests/kube-scheduler.yaml
     - source: salt://metalk8s/kubernetes/files/control-plane-manifest.yaml
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+      - /etc/kubernetes/scheduler.conf
     - context:
         name: kube-scheduler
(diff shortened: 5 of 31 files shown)


salt/_states/metalk8s.py
```diff
@@ -0,0 +1,32 @@
...
+        for config_file in config_files
+    ]
+
+    return __states__["file.managed"](
+        name,
+        source,
+        template="jinja",
+        user="root",
+        group="root",
+        mode="0644",
```
NicolasT: Maybe this should be `0600`? No need to be world-readable.

salt/_states/metalk8s.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def static_pod_managed(name, source, config_files, context):
+    """Simple helper to edit a static Pod manifest if configuration changes.
+
+    Expects the template to use the `config_digests` variable and store it in
```
NicolasT: Why plural?

salt/_states/metalk8s.py
```diff
@@ -0,0 +1,32 @@
...
+    return __states__["file.managed"](
+        name,
+        source,
+        template="jinja",
+        user="root",
+        group="root",
+        mode="0644",
+        makedirs=False,
+        backup=False,
+        context=dict(context, config_digests=",".join(config_digests)),
```
NicolasT: Uh, nah. That'd mean we can get some very long but otherwise useless annotation on the Pod. If indeed we want to use `sha256` (which isn't even required here), then the result should be, I don't know, a checksum of the concatenation of all sorted checksums or something? Basically, one single value, ideally short, which is (quasi-)guaranteed to change when the input(s) change.

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -80,15 +80,13 @@ Set up default basic auth htpasswd:
...
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+        - {{ htpasswd_path }}
+        - /etc/keepalived/check-apiserver.sh
+        - /etc/keepalived/keepalived.conf.sh
```
NicolasT, line 102: Also API-server certificates (CA, server cert, `kubelet` client cert,...)

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -80,15 +80,13 @@ Set up default basic auth htpasswd:
...
     - source: salt://metalk8s/kubernetes/files/control-plane-manifest.yaml
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+        - {{ htpasswd_path }}
+        - /etc/keepalived/check-apiserver.sh
```
NicolasT, line 101: These only when `keepalived` is enabled, right?

salt/metalk8s/kubernetes/controller-manager/installed.sls
```diff
@@ -45,5 +41,5 @@ Create kube-controller-manager Pod manifest:
           - path: /etc/kubernetes/controller-manager.conf
             name: kubeconfig
             type: File
-    - require:
+    - onchanges:
```
NicolasT, line 46: Also a requisite, no?

salt/metalk8s/kubernetes/files/control-plane-manifest.yaml
```diff
@@ -3,6 +3,7 @@ kind: Pod
 metadata:
   annotations:
     scheduler.alpha.kubernetes.io/critical-pod: ""
+    metalk8s.scality.com/config-digests: "{{ config_digests }}"
```
NicolasT: Could be singular, if aforementioned changes are made.

salt/metalk8s/kubernetes/scheduler/installed.sls
```diff
@@ -29,5 +25,5 @@ Create kube-scheduler Pod manifest:
           - path: /etc/kubernetes/scheduler.conf
             name: kubeconfig
             type: File
-    - require:
+    - onchanges:
```
NicolasT, line 28: And require?

NicolasT (changes requested):

salt/_states/metalk8s.py
gdemonet: Ok will do

salt/_states/metalk8s.py
gdemonet: Would something along the lines of `md5( '-'.join( sha256(file) for file in config_files )` suit our needs ?

salt/metalk8s/kubernetes/apiserver/installed.sls
gdemonet: OK, I was wondering about certs

salt/metalk8s/kubernetes/controller-manager/installed.sls
gdemonet: So, from what I understand, `onchanges` behaves exactly like a `require` if the referenced state does not return information about changes (so only executes if the requisite returned `True`), otherwise looks at the changes to see if needs to execute. Ordering should be taken care of nicely, if that was your concern.

salt/_states/metalk8s.py
gdemonet: Done

salt/metalk8s/kubernetes/apiserver/installed.sls
gdemonet: What about public/private keys? I added `sa.pub` for `k-a` but was not sure whether to store the private key digest (security?)...

salt/metalk8s/kubernetes/apiserver/installed.sls
gdemonet: Done

gdemonet:
@NicolasT I changed the digest computation (only store in the annotation the MD5 sum of the concatenation of SHA256 digests of listed `config_files`), and added certificates where relevant.

Now, I'm wondering if this would warrant some automatic tests, or, since it's mostly about upgrade/development scenarii, if we can simply check it manually. If the former, how would we do such tests? I'm thinking about simply testing the `metalk8s.static_pod_managed` state with some arbitrary static Pod (not a useful one, just an artificial one for the test) to restart based on some arbitrary file... Would that work? Or do we want to check that, for each service, configuration changes effectively get picked up by said service?

salt/metalk8s/kubernetes/scheduler/installed.sls
```diff
@@ -4,15 +4,11 @@ include:
...
     - name: /etc/kubernetes/manifests/kube-scheduler.yaml
     - source: salt://metalk8s/kubernetes/files/control-plane-manifest.yaml
-    - template: jinja
-    - user: root
-    - group: root
-    - mode: 644
-    - makedirs: True
-    - dir_mode: 750
+    - config_files:
+      - /etc/kubernetes/scheduler.conf
```
NicolasT, line 11: Shouldn't this one have `/etc/kubernetes/pki/ca.crt` just like `k-c-m`?

gdemonet: It does not depend on it, neither does it mount the `/etc/kubernetes/pki` directory, so I guess not? In any case, the certificate data is embedded in the generated kubeconfig, and the `metalk8s_kubeconfig: Create kubeconfig file for scheduler` state retrieves the "latest" certificate data from the mine.

charlesprost:
> Now, I'm wondering if this would warrant some automatic tests, or, since it's mostly about upgrade/development scenarii, if we can simply check it manually. If the former, how would we do such tests? I'm thinking about simply testing the `metalk8s.static_pod_managed` state with some arbitrary static Pod (not a useful one, just an artificial one for the test) to restart based on some arbitrary file... Would that work? Or do we want to check that, for each service, configuration changes effectively get picked up by said service?

Yeah I'd say it would be good to test this feature with a dummy pod. Let's have a pod doing nothing, based on a configuration file with some random string, let's change this string, and see if the pod restarts as expected.

Regarding the image, we could reuse a pod such as nginx without mounting anything.

salt/metalk8s/kubernetes/scheduler/installed.sls
NicolasT: Hmh, then why would `k-c-m` require `ca.crt`?

gdemonet: I put it there because it mounts `/etc/kubernetes/pki` and has a flag for this cert `--cluster-signing-cert-file=/etc/kubernetes/pki/ca.crt`, but maybe it's not required indeed.

salt/metalk8s/kubernetes/apiserver/installed.sls
NicolasT: Let's assume the digest function is strong enough not to be reversible... It's the MD5 of a concatenation of SHA256s of the inputs, should be fine ;)

tests/conftest.py
```diff
@@ -10,3 +14,43 @@ def pytest_addoption(parser):
         type=pathlib.Path,
         help="Root of the ISO file tree."
     )
+
+# Fixtures {{{
+
+
+@pytest.fixture(scope="module")
+def version(request, host):
```
NicolasT: `short_version`?

tests/conftest.py
```diff
@@ -10,3 +14,43 @@ def pytest_addoption(parser):
...
+@pytest.fixture
+def kubeconfig(kubeconfig_data, tmp_path):
+    kubeconfig_path = tmp_path / "admin.conf"
+    kubeconfig_path.write_text(yaml.dump(kubeconfig_data), encoding='utf-8')
+    return str(kubeconfig_path)  # Need Python 3.6 to open() a Path object
+
+
+@pytest.fixture
+def k8s_client(kubeconfig):
+    kubernetes.config.load_kube_config(config_file=kubeconfig)
```
NicolasT: This relies on the 'global' config object in the K8s Python API, which can cause difficult-to-debug issues if we ever want to test RBAC authz and such.

Consider using `new_client_from_config`, then construct an API access object passing in the returned client explicitly.

Also, this actually returns a corev1 API client, not an `ApiClient`, which I'd expect given the name of the procedure.

NicolasT (changes requested):

tests/conftest.py
gdemonet: I just moved it, but yes let's rename it :)

tests/conftest.py
gdemonet: Sure, will do!

gdemonet: So I still have a `k8s_client` for clients exposing a single API, and `k8s_apiclient` that returns an `ApiClient` to use in `k8s_client`. By default, we use `CoreV1Api`, but it can be parametrized with `pytest.mark.parametrize`.

NicolasT (dismissed):
:ship:

gdemonet:
> :ship:

Sadly, I'm having weird results in CI tests execution, so your approval was dismissed @NicolasT - will ping once it's reliably fixed.

gdemonet:
@bert-e reset

gdemonet:
Finally, three green builds! :shipit: :)

Reviewers should note the PR description changed, and the changeset now includes improvements to the test suite (especially #881, but also around `pytest` markers, that one can discover using `tox -e tests -- --markers` for instance).

gdemonet:
@bert-e approve

NicolasT (approved):

PR #1081, opened 2019-05-06
build: don't pass `--broken` to `git`

NicolasT:
In 4979e75e3c03f the arguments passed to `git describe` to find the Git
revision of a build were somewhat changed, adding `--broken`. However,
the version of Git as shipped with CentOS 7 doesn't support this option.

Since it's not strictly required anyway, simply removing it.

See: 4979e75e3c03f2a6786b08b9d1fcc66f92f7548e


NicolasT:
@bert-e approve

gdemonet (approved):

Issue #1087, opened 2019-05-06
Set up fallback loopback CNI


NicolasT:
When setting up a `master` and/or `etcd` node, we don't deploy Calico (and its configuration) on the server, because such node is not required to be part of the overlay network (it's not running any non-`hostNetwork` Pods, and shouldn't connect to any other non-`hostNetwork` Pods).

However, as long as no CNI configuration is present, `kubelet` complains and the Node remains in `NotReady` state.

We can work-around this by either

- explaining this in documentation
- unnecessarily deploying Calico on the node
- providing another valid CNI configuration, e.g. for `localhost`

The latter seem the right approach here: it'll still require documentation ('Why is my Pod not working on a non-`node`/`infra` node after I removed the taints?'), but doesn't unnecessarily impact the networking on a control-plane node, and doesn't keep Nodes in `NotReady` state even though they kind-of are.

NicolasT:
#1089

PR #1089, opened 2019-05-06
cni: deploy a fallback `loopback` CNI configuration

NicolasT:
When setting up a *master* and/or *etcd* node, we don't deploy Calico
(and its configuration) on the server, because such node is not
required to be part of the overlay network (it's not running any
non-`hostNetwork` Pods, and shouldn't connect to any other
non-`hostNetwork` Pods).

However, as long as no CNI configuration is present, `kubelet`
complains and the Node remains in `NotReady` state.

We can work-around this by either

- explaining this in documentation
- unnecessarily deploying Calico on the node
- providing another valid CNI configuration, e.g. for `localhost`

The latter seem the right approach here: it'll still require
documentation ('Why is my Pod not working on a non-node/infra node after
I removed the taints?'), but doesn't unnecessarily impact the networking
on a control-plane node, and doesn't keep Nodes in `NotReady` state even
though they kind-of are.

Fixes: #1087
Fixes: https://github.com/scality/metalk8s/issues/1087

### DIFF
--- a/salt/metalk8s/kubernetes/cni/loopback/init.sls
+++ b/salt/metalk8s/kubernetes/cni/loopback/init.sls
@@ -1,0 +1,27 @@
+# When setting up a *master* and/or *etcd* node, we don't deploy Calico
+# (and its configuration) on the server, because such node is not
+# required to be part of the overlay network (it's not running any
+# non-`hostNetwork` Pods, and shouldn't connect to any other
+# non-`hostNetwork` Pods).
+#
+# However, as long as no CNI configuration is present, `kubelet`
+# complains and the Node remains in `NotReady` state.
+#
+# We can work-around this by either
+#
+# - explaining this in documentation
+# - unnecessarily deploying Calico on the node
+# - providing another valid CNI configuration, e.g. for `localhost`
+#
+# The latter seem the right approach here: it'll still require
+# documentation ('Why is my Pod not working on a non-node/infra node after
+# I removed the taints?'), but doesn't unnecessarily impact the networking
+# on a control-plane node, and doesn't keep Nodes in `NotReady` state even
+# though they kind-of are.
+#
+# Fixes: #1087
+# Fixes: https://github.com/scality/metalk8s/issues/1087
+
+include:
+  - .installed
+  - .configured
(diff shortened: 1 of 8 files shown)


NicolasT:
@bert-e reset

NicolasT:
I guess @xaf-scality could approve this as well ;-)

salt/metalk8s/kubernetes/cni/loopback/installed.sls
```diff
@@ -0,0 +1,10 @@
+{%- from "metalk8s/macro.sls" import pkg_installed with context %}
+{%- from "metalk8s/map.jinja" import repo with context %}
```
gdemonet: This `repo` variable is unused, right?

salt/metalk8s/kubernetes/cni/loopback/init.sls
```diff
@@ -0,0 +1,3 @@
+include:
```
gdemonet, line 25: Could we have some really basic documentation in this file? Just to explain the rationale and expected usage of the formula (mostly an alternative to Calico for `hostNetwork`-only nodes IIUC). A later reference to this comment in a "Troubleshooting guide" would be easy to add.

gdemonet (changes requested):
A few minor comments on the code itself.

I tried deploying a control-plane node starting from this branch, but the `deploy_new_node` orchestrate probably isn't up-to-date, which got my etcd cluster out of shape (new node expects a different cluster ID than the first one is using). Will investigate further, though it's probably not related to this changeset.

salt/metalk8s/kubernetes/cni/loopback/installed.sls
NicolasT: Removed

salt/metalk8s/kubernetes/cni/loopback/init.sls
NicolasT: I added the commit-message as a 'docstring' in the SLS.

NicolasT:
@bert-e reset

gdemonet (approved):

NicolasT:
@bert-e approve

Issue #1034, opened 2019-04-25
Upgrade Calico to 3.7


NicolasT:
No reason to ship an old release from the get-go.

NicolasT:
#1098

PR #1092, opened 2019-05-07
salt: add `kubernetes` renderer

NicolasT:
Add a (very incomplete) `renderer` for Kubernetes YAML manifests.

The intent is to take a standard Kubernetes YAML manifest (or a pipeline generating such manifest, e.g. after Jinja templating), and output a Salt SLS using the `metalk8s_kubernetes` states to realize the various objects contained in said manifest.

It's 'incomplete' because currently, only `v1/ServiceAccount` objects are handled properly. Handlers for other object kinds are to be added gradually.

As an example (and to test), first deploy the renderer:

```
$ salt-call saltutil.sync_renderers saltenv=metalk8s-2.0
```

Then, assuming the following file exists in your Salt tree:

```
$ cat /srv/scality/metalk8s-2.0/salt/metalk8s/k8s-demo.sls
#!jinja | kubernetes

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: calico-node
  namespace: kube-system

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: calico-node2
  namespace: kube-system
```
(note: in the above, the `jinja` part of the rendering pipeline is not necessary since we're not doing any templating anyway)

Then, let's try to render (without validating) the result:

```
$ salt-call slsutil.renderer /srv/scality/metalk8s-2.0/salt/metalk8s/k8s-demo.sls --out=yaml

local:
  Apply v1/ServiceAccount 'kube-system/calico-node':
    metalk8s_kubernetes.serviceaccount_present:
    - name: calico-node
    - namespace: kube-system
  Apply v1/ServiceAccount 'kube-system/calico-node2':
    metalk8s_kubernetes.serviceaccount_present:
    - name: calico-node2
    - namespace: kube-system
```

Finally, rendering and validating the whole SLS:

```
$ salt-call state.show_sls metalk8s.k8s-demo saltenv=metalk8s-2.0
local:
    ----------
    Apply v1/ServiceAccount 'kube-system/calico-node':
        ----------
        __env__:
            metalk8s-2.0
        __sls__:
            metalk8s.k8s-demo
        metalk8s_kubernetes:
            |_
              ----------
              name:
                  calico-node
            |_
              ----------
              namespace:
                  kube-system
            - serviceaccount_present
            |_
              ----------
              order:
                  10000
    Apply v1/ServiceAccount 'kube-system/calico-node2':
        ----------
        __env__:
            metalk8s-2.0
        __sls__:
            metalk8s.k8s-demo
        metalk8s_kubernetes:
            |_
              ----------
              name:
                  calico-node2
            |_
              ----------
              namespace:
                  kube-system
            - serviceaccount_present
            |_
              ----------
              order:
                  10001
```

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -246,6 +246,8 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/_pillar/metalk8s_endpoints.py'),
     Path('salt/_pillar/metalk8s_nodes.py'),
 
+    Path('salt/_renderers/metalk8s_kubernetes.py'),
+
     Path('salt/_roster/kubernetes_nodes.py'),
 
     Path('salt/_runners/metalk8s_saltutil.py'),
--- a/salt/_renderers/metalk8s_kubernetes.py
+++ b/salt/_renderers/metalk8s_kubernetes.py
@@ -1,0 +1,99 @@
+'''
+A renderer for Kubernetes YAML manifests
+
+Given a Kubernetes YAML file (which may be a stream of objects, i.e. YAML
+snippets separated by `---` lines), this will render an SLS which calls Salt
+`*_present` states for every such object.
+
+To use it, add a shebang like `#!kubernetes` as the first line of your manifests
+SLS file. Optionally, you can use rendering pipelines (if templating is
+required), e.g. `#!jinja | kubernetes`.
+'''
+
+import yaml
+
+import salt.utils.yaml
+from salt.utils.odict import OrderedDict
+
+from salt.ext import six
+
+__virtualname__ = 'kubernetes'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def _step_name(obj):
+    namespace = obj['metadata'].get('namespace')
+
+    if namespace:
+        name = '{}/{}'.format(
+            namespace,
+            obj['metadata']['name'],
+        )
+    else:
+        name = obj['metadata']['name']
+
+    return "Apply {}/{} '{}'".format(
+        obj['apiVersion'],
+        obj['kind'],
+        name,
+    )
+
+
+_HANDLERS = {}
+
+def handle(api_version, kind):
+    '''
+    Register a 'handler' (object -> state mapping) for `apiVersion` and `kind`
+    '''
+    tag = (api_version, kind)
+
+    def register(f):
+        assert tag not in _HANDLERS
+
+        _HANDLERS[tag] = f
+
+        return f
+
+    return register
+
+
+@handle('v1', 'ServiceAccount')
+def _handle_v1_serviceaccount(obj):
+    return {
+        'metalk8s_kubernetes.serviceaccount_present': [
+            {'name': obj['metadata']['name']},
+            {'namespace': obj['metadata']['namespace']},
+        ],
+    }
+
+
+del handle
+
+
+def _step(obj):
+    '''
+    Handle a single Kubernetes object, rendering it into a state 'step'
+    '''
+    name = _step_name(obj)
+    api_version = obj['apiVersion']
+    kind = obj['kind']
+
+    handler = _HANDLERS.get((api_version, kind))
+    if not handler:
+        raise ValueError('No handler for {}/{}'.format(api_version, kind))
+
+    state = handler(obj)
+
+    return (name, state)
+
+
+def render(yaml_data, saltenv='', sls='', **kwargs):
+    if not isinstance(yaml_data, six.string_types):
+        yaml_data = yaml_data.read()
+
+    data = yaml.load_all(yaml_data, Loader=salt.utils.yaml.SaltYamlSafeLoader)
+
+    return OrderedDict(_step(obj) for obj in data if obj)


gdemonet:
I'm not sure about the approach... Why would we want to map source YAML files to SLS formulas?

IIUC, the goals are:

- [ ] to not embed Kubernetes objects (retrieved from upstream, e.g. for `kube-prometheus`) into our formulas, so we can easily compare them with newer versions from upstream
- [ ] to reduce boilerplate (since we have the `apiVersion`/`kind` directly in the objects) when deploying them

Now, the renderer approach you used fixes the first goal, but the second is still problematic IMHO (if an object kind/version is not handled yet in `_renderers/kubernetes.py`).
What is `kubectl apply/create -f manifests.yml` doing under the hood? Would it be "bad" to have a simple `metalk8s_kubernetes.object_managed` state which does roughly the same (i.e. take a YAML file and sends it to API server, optionally using a `template` engine)?

Except for the fact we use different APIs from the Kubernetes Python client, I don't see the benefit of using different state functions for different kinds of objects...

NicolasT:
That's not what `kubectl` does, and it's not how the K8s API works. You can not send it 'a bunch of objects', only one object at a time (or multiple objects of the same kind, not a mixture).

So, `kubectl apply` will basically perform a `POST`/`PUT` for every object found in the input, one by one. Which is in the end what this approach does as well.

I guess we could in theory implement some `object_managed` state, and I considered that. At the same time, that won't really work out of the box, though granted it'd be feasible. However, that would preclude any kind of client-side input/model validation, which we do have with the current approach.

gdemonet:
> That's not what `kubectl` does, and it's not how the K8s API works. You can not send it 'a bunch of objects', only one object at a time (or multiple objects of the same kind, not a mixture).

Agreed, but is "sending a bunch of objects of different kinds in a single call" the approach we want to have? I feel like for the rest of the project, we prefer being declarative as much as possible, and manage ordering through Salt requisite system. Having a rendered SLS from a set of "plain K8s" YAML objects could hardly use this system without having either modifications to the original YAMLs to _declare_ ordering, or a set of arbitrary rules to _infer_ ordering (e.g. source file ordering). However, maybe we don't, and won't ever, need specific ordering for this specific use-case of large deployments from other projects such as `kube-prometheus`.

> So, `kubectl apply` will basically perform a `POST`/`PUT` for every object found in the input, one by one. Which is in the end what this approach does as well.

OK, my question was more about "who manages validation of input?": does `kubectl` `POST`/`PUT`s the object unvalidated to `apiserver` on a generic endpoint and `apiserver` does all the heavy lifting? Or does it parse the object's `kind` and `apiVersion` to know which resource should receive the object? In the latter scenario, does it also have the schemas of various resources available locally to perform client-side validation, or is _that_ left to `apiserver`?

> I guess we could in theory implement some `object_managed` state, and I considered that. At the same time, that won't really work out of the box, though granted it'd be feasible. However, that would preclude any kind of client-side input/model validation, which we do have with the current approach.

Indeed, it may be a little more work than the renderer if we want to handle `_managed` as well as `file_managed` does, for instance. In the meantime, we could work on having an `object_present` state that reuses your parsing logic to map `kind`/`apiVersion` pairs to registered handlers, but directly in the `metalk8s_kubernetes` state module. Shouldn't be too different from the existing code IMHO.

Quick example of how I imagine it could look in a formula:

```yaml
Deploy some ServiceAccount:
  metalk8s_kubernetes.object_present:
    - source: salt://{{ slspath }}/files/serviceaccount.yaml

Deploy some Deployment:
  metalk8s_kubernetes.object_present:
    - source: salt://{{ slspath }}/files/deployment.yaml.j2
    - template: jinja
    - require:
      - metalk8s_kubernetes: Deploy some ServiceAccount
```

And having something along those lines for the state function:

```python
@handle(kind='ServiceAccount', version='*')
def serviceaccount_present(*args, **kwargs):
    # do stuff, with validation etc.

def object_present(source, template, *args, **kwargs):
    kind, version, manifest_data = parse(source)
    if (kind, version) in Handlers:
        return Handlers.get(kind, version)(manifest_data, *args, **kwargs)
```

NicolasT:
> > That's not what `kubectl` does, and it's not how the K8s API works. You can not send it 'a bunch of objects', only one object at a time (or multiple objects of the same kind, not a mixture).
> 
> Agreed, but is "sending a bunch of objects of different kinds in a single call" the approach we want to have? I feel like for the rest of the project, we prefer being declarative as much as possible, and manage ordering through Salt requisite system.

Fair point, I could add ordering. However, if I'm not mistaken, `kubectl apply` does not do *requires*: it does apply the manifests in-order, but if one of them fails for whatever reason, it'll happily continue with the next one.

> Having a rendered SLS from a set of "plain K8s" YAML objects could hardly use this system without having either modifications to the original YAMLs to _declare_ ordering, or a set of arbitrary rules to _infer_ ordering (e.g. source file ordering).

So, ordering would come from source file ordering, as is done in `kubectl`, without dependencies.

> However, maybe we don't, and won't ever, need specific ordering for this specific use-case of large deployments from other projects such as `kube-prometheus`.

We do, actually, because you can't create an object of type `version/kind` before `version/kind` is known to the API-server, i.e. when its CRD is deployed, in case of a non-core object kind.

> 
> > So, `kubectl apply` will basically perform a `POST`/`PUT` for every object found in the input, one by one. Which is in the end what this approach does as well.
> 
> OK, my question was more about "who manages validation of input?": does `kubectl` `POST`/`PUT`s the object unvalidated to `apiserver` on a generic endpoint and `apiserver` does all the heavy lifting?

There is no such 'generic endpoint' in the whole API.

> Or does it parse the object's `kind` and `apiVersion` to know which resource should receive the object?

Something like that, yes.

> In the latter scenario, does it also have the schemas of various resources available locally to perform client-side validation, or is _that_ left to `apiserver`?

It does a local validation based (this *may* be based on schemas it receives from the API server) which you can override, but of course the API server itself performs validation as well.

> > I guess we could in theory implement some `object_managed` state, and I considered that. At the same time, that won't really work out of the box, though granted it'd be feasible. However, that would preclude any kind of client-side input/model validation, which we do have with the current approach.
> 
> Indeed, it may be a little more work than the renderer if we want to handle `_managed` as well as `file_managed` does, for instance.

Maybe `_managed` was not the right choice of words here. `object_present` indeed.

> In the meantime, we could work on having an `object_present` state that reuses your parsing logic to map `kind`/`apiVersion` pairs to registered handlers, but directly in the `metalk8s_kubernetes` state module. Shouldn't be too different from the existing code IMHO.

I'm not overly fond of that. I'd rather be able to see using `state.show_sls` what would happen when my SLS is rendered explicitly (i.e. calls to `metalk8s_kubernetes.foobar_present`), which would let me know *at compile time* some object kind I have in the source is not supported/handler, rather than get an output which lists calls to some generic `metalk8s_kubernetes.object_present`, which would then *at runtime* fail.

> 
> Quick example of how I imagine it could look in a formula:
> 
> ```yaml
> Deploy some ServiceAccount:
>   metalk8s_kubernetes.object_present:
>     - source: salt://{{ slspath }}/files/serviceaccount.yaml
> 
> Deploy some Deployment:
>   metalk8s_kubernetes.object_present:
>     - source: salt://{{ slspath }}/files/deployment.yaml.j2
>     - template: jinja
>     - require:
>       - metalk8s_kubernetes: Deploy some ServiceAccount
> ```

One of the primary reasons to have a renderer is *not* to need to write something like the above, and simply take an upstream YAML file, add a shebang and be able to deploy it.

> 
> And having something along those lines for the state function:
> 
> ```python
> @handle(kind='ServiceAccount', version='*')
> def serviceaccount_present(*args, **kwargs):
>     # do stuff, with validation etc.
> 
> def object_present(source, template, *args, **kwargs):
>     kind, version, manifest_data = parse(source)
>     if (kind, version) in Handlers:
>         return Handlers.get(kind, version)(manifest_data, *args, **kwargs)
> ```

So, now you moved something that could be a compile-time error into a run-time error.

NicolasT:
BTW, your `object_present` with a `source` argument would only work for source files which only contain a single object. The intent of the renderer is as well to support YAML object stream documents, as is often used upstream, or would be by us if we use `helm template` to generate some deployment manifests.

gdemonet:
> One of the primary reasons to have a renderer is not to need to write something like the above, and simply take an upstream YAML file, add a shebang and be able to deploy it

Oh, so I think we got misaligned with the objective: I thought we exclusively wanted the ability to deploy objects in the API from files directly. I do get the benefit of using _almost_ the exact same file as exists upstream (we can diff our version with a new release), and it probably outweighs the need for control.

> > However, maybe we don't, and won't ever, need specific ordering for this specific use-case of large deployments from other projects such as kube-prometheus.
>
> We do, actually, because you can't create an object of type version/kind before version/kind is known to the API-server, i.e. when its CRD is deployed, in case of a non-core object kind.

Then we could, I assume, add an `include` section to those YAML-generated formulas, but how would we specify explicit dependencies? For instance:

_crd.sls_
```yaml
#! kubernetes
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata: {...}
spec:
  # ...
  names:
    kind: Example
---
```

_cr.sls_
```yaml
#! kubernetes
# First section, with the shebang, is not parsed as an object to manage

include:
  - .crd
---
apiVersion: example.domain.com/v1alpha1
kind: Example
# ... but how to explicitly `require: metalk8s_kubernetes.crd_present: {{ constructed_name }}`?
---
```

NicolasT:
> > One of the primary reasons to have a renderer is not to need to write something like the above, and simply take an upstream YAML file, add a shebang and be able to deploy it
> 
> Oh, so I think we got misaligned with the objective: I thought we exclusively wanted the ability to deploy objects in the API from files directly. I do get the benefit of using _almost_ the exact same file as exists upstream (we can diff our version with a new release), and it probably outweighs the need for control.

Yeah, not just 'from files'. I ran into this while I tried to implement #1034 with the updated https://docs.projectcalico.org/v3.7/manifests/calico.yaml. Right now, we took the 3.5 manifest and manually turned it into a Salt state calling into `metalk8s_kubernetes.*`. However, that's quite a bit of work, and updating it with a new upstream manifest is... tedious. Furthermore, we can't be sure we correctly represent the upstream manifest.

> 
> > > However, maybe we don't, and won't ever, need specific ordering for this specific use-case of large deployments from other projects such as kube-prometheus.
> > 
> > 
> > We do, actually, because you can't create an object of type version/kind before version/kind is known to the API-server, i.e. when its CRD is deployed, in case of a non-core object kind.
> 
> Then we could, I assume, add an `include` section to those YAML-generated formulas, but how would we specify explicit dependencies? For instance:
> 
> _crd.sls_
> 
> ```yaml
> #! kubernetes
> ---
> apiVersion: apiextensions.k8s.io/v1beta1
> kind: CustomResourceDefinition
> metadata: {...}
> spec:
>   # ...
>   names:
>     kind: Example
> ---
> ```
> 
> _cr.sls_
> 
> ```yaml
> #! kubernetes
> # First section, with the shebang, is not parsed as an object to manage
> 
> include:
>   - .crd
> ---
> apiVersion: example.domain.com/v1alpha1
> kind: Example
> # ... but how to explicitly `require: metalk8s_kubernetes.crd_present: {{ constructed_name }}`?
> ---
> ```

I don't think that'd work: when we use upstream manifests, they're either

- a collection of files, in which case we can create a higher-level SLS which indeed `include`s all those K8s YAML files (as SLSs, so with the renderer shebang added to them). This should take care of ordering?
- a single file, i.e. a single SLS/YAML with the shebang. In that case, we could enforce ordering by https://docs.saltstack.com/en/latest/ref/states/ordering.html#the-order-option though I'm not entirely sure how that then interacts with `include`s

salt/_renderers/kubernetes.py
```diff
@@ -0,0 +1,93 @@
+'''
```
NicolasT: Ugh, I must rename this file or we run into the module name duplication again, potentially.

NicolasT: Done

NicolasT:
@bert-e reset

gdemonet:
> * a collection of files, in which case we can create a higher-level SLS which indeed `include`s all those K8s YAML files (as SLSs, so with the renderer shebang added to them). This should take care of ordering?

Yes I guess we can rely on `include`'s ordering: https://docs.saltstack.com/en/latest/ref/states/compiler_ordering.html#the-include-statement

> * a single file, i.e. a single SLS/YAML with the shebang. In that case, we could enforce ordering by https://docs.saltstack.com/en/latest/ref/states/ordering.html#the-order-option though I'm not entirely sure how that then interacts with `include`s

I'm not sure if it's worth using the `order` option unless we want to re-order the upstream manifest, we should be able to rely on file ordering and the behaviour of `include` described in the link above

NicolasT:
So, keeping things as-is :)

gdemonet (approved):

NicolasT:
@bert-e approve

PR #1094, opened 2019-05-07
Install kubectl on bootstrap node

### LINKED ISSUE #622: Install kubectl during the bootstrap
Install kubectl during the bootstrap

ramanduh:
Install kubectl on the bootstrap node:
* install kubectl package
* configure the default KUBECONFIG by creating a symbolic link `/root/.kube/config -> /etc/kubernetes/admin.conf`

Fixes #622

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -104,6 +104,9 @@ SALT_FILES : Tuple[Union[Path, targets.FileTarget], ...] = (
     Path('salt/metalk8s/internal/preflight/mandatory.sls'),
     Path('salt/metalk8s/internal/preflight/recommended.sls'),
 
+    Path('salt/metalk8s/kubectl/init.sls'),
+    Path('salt/metalk8s/kubectl/installed.sls'),
+
     Path('salt/metalk8s/kubernetes/apiserver/certs/etcd-client.sls'),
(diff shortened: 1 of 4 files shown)


NicolasT:
I'm not sure the symlink is a good idea. Thoughts @scality/metalk8s ?

charlesprost:
> I'm not sure the symlink is a good idea. Thoughts @scality/metalk8s ?

We could probably define the env var in some .bashrc script?

NicolasT:
> > I'm not sure the symlink is a good idea. Thoughts @scality/metalk8s ?
> 
> We could probably define the env var in some .bashrc script?

Oh, no, that'd be even 'worse' IMHO. Just thinking we should not do any of that, simply install the tool and document how to 'make it connect'.

WDYT @xaf-scality ?

xaf-scality:
I don't like the symlink option as this hide something important for this tool.
For me the best option is to show at the end of the installation a quick message explaining the export command to permit kubectl to connect.
Perhaps we can have something asking : do you want to have the export blablabla in your .bashrc env file ?

ramanduh:
I updated the PR so that, we let the operator/user, set his own configuration of kubectl

ramanduh:
@bert-e reset

NicolasT (approved):

ramanduh:
@bert-e approve

PR #1098, opened 2019-05-10
Update Calico to 3.7.2

NicolasT:
Fixes: #1034


salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -85,15 +166,24 @@ def _step(obj):
...
         raise ValueError('No handler for {}/{}'.format(api_version, kind))
 
-    state = handler(obj)
+    state = handler(obj, kubeconfig=kubeconfig, context=context)
 
     return (name, state)
 
 
 def render(yaml_data, saltenv='', sls='', **kwargs):
+    args = six.moves.urllib.parse.parse_qs(kwargs.get('argline', ''))
```
gdemonet, line 175: As discussed, this seems unnecessarily complex - would be easier to set `kubeconfig`/`context` in the master config, since we don't plan on changing them depending on which YAML we are sending to K8s API. See #1099.

salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -61,19 +61,100 @@ def register(f):
...
             {'namespace': obj['metadata']['namespace']},
         ],
     }
 
 
+@handle('v1', 'ConfigMap')
+def _handle_v1_configmap(obj, kubeconfig, context):
+    return {
+        'metalk8s_kubernetes.configmap_present': [
+            {'name': obj['metadata']['name']},
```
gdemonet, line 79: All these states require a `name`, which is always retrieved with `.metadata.name`. Same for `kubeconfig` and `context`.
Maybe it would be interesting to have some helpers to generate them, and mutualize the common parts... Probably not in this PR.

gdemonet (approved):
LGTM! You may want to open a tracking ticket for the _DRY comment_ ;)

NicolasT:
@bert-e approve

salt/_renderers/metalk8s_kubernetes.py
NicolasT: Filed a debt ticket.

PR #1102, opened 2019-05-13
Improvement/878 use python etcd3 instead of call to etcdctl

ramanduh:
**Component**:

salt, deployment

**Context**: 

When adding a new member to the cluster etcd, we used etcdctl, now we use python-etcd3.

python etcd3 is installed only on the salt master container (like python-kubernetes)

The PR includes the 4 following points:

* provision an etcd client cert for salt-master, similar to how we'll create a kubeconfig file for it
* add the python-etcd3 package in the salt-master image
* create a Salt module to manage etcd members in a cluster, given the above client cert
* use this module to add the new node as a member, on the salt-master host

Note: There is no RPM package of python etcd3 so I use pip package

**Summary**:

**Acceptance criteria**: 

Run a successful expansion

### DIFF
--- a/images/salt-master/Dockerfile
+++ b/images/salt-master/Dockerfile
@@ -24,7 +24,9 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
  && yum clean expire-cache \
  && yum install -y epel-release \
  && yum install -y python2-kubernetes salt-master salt-api salt-ssh openssh-clients \
+ && yum install -y python-pip \
  && yum clean all \
+ && pip install etcd3 \
  && chmod +x /tini
 
 # salt-master, salt-api
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -33,7 +33,7 @@ def deps_missing(*args, **kwargs):
     raise CommandExecutionError("Kubernetes python client is not installed")
 
 
-@depends('KUBERNETES_PRESENT', fallback_function=deps_missing)
+@depends(KUBERNETES_PRESENT, fallback_function=deps_missing)
 def wait_apiserver(retry=10, interval=1, **kwargs):
     """Wait for kube-apiserver to respond.
 
@@ -102,7 +102,7 @@ def _execute_etcd_command(pod_name, cmd):
         raise exn
 
 
-@depends('KUBERNETES_PRESENT', fallback_function=deps_missing)
+@depends(KUBERNETES_PRESENT, fallback_function=deps_missing)
 def add_etcd_node(host, endpoint=None):
     '''Add a new `etcd` node into the `etcd` cluster.
 
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -9,17 +9,6 @@ import time
 from salt.exceptions import CommandExecutionError
 from salt.utils.decorators import depends
 
-KUBERNETES_PRESENT = False
-try:
-    import kubernetes.client
-    import kubernetes.config
-    from kubernetes.client.rest import ApiException
-    from kubernetes.stream import stream as k8s_stream
-    KUBERNETES_PRESENT = True
-except ImportError:
-    pass
-
-
 log = logging.getLogger(__name__)
 
 __virtualname__ = 'metalk8s'
@@ -29,11 +18,6 @@ def __virtual__():
     return __virtualname__
 
 
-def deps_missing(*args, **kwargs):
-    raise CommandExecutionError("Kubernetes python client is not installed")
-
-
-@depends(KUBERNETES_PRESENT, fallback_function=deps_missing)
 def wait_apiserver(retry=10, interval=1, **kwargs):
     """Wait for kube-apiserver to respond.
 
@@ -66,52 +50,6 @@ def get_etcd_endpoint():
         return '{host}=https://{ip}:{port}'.format(
             host=hostname, ip=ips[0], port=DEFAULT_ETCD_PORT
         )
-
-
-def _execute_etcd_command(pod_name, cmd):
-    if isinstance(cmd, str):
-        cmd = [cmd]
-    try:
-        client = kubernetes.config.new_client_from_config(
-            config_file='/etc/kubernetes/admin.conf'
-        )
-    except:
-        log.exception('failed to load kubeconfig')
-        raise
-
-    api = kubernetes.client.CoreV1Api(api_client=client)
-    etcd_command = [
-        'etcdctl',
-        '--endpoints', 'https://127.0.0.1:2379',
-        '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
-        '--key-file', '/etc/kubernetes/pki/etcd/server.key',
-        '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
-    ] + cmd
-    try:
-        err = k8s_stream(
-            api.connect_get_namespaced_pod_exec,
-            name=pod_name, namespace='kube-system',
-            command=etcd_command,
-            stderr=True, stdin=False, stdout=False, tty=False
-        )
-        if err:
-            log.error('etcdctl error: %s', err)
-            raise CommandExecutionError('etcdctl: {}'.format(err))
-    except ApiException as exn:
-        log.exception('failed to run etcdctl')
-        raise exn
-
-
-def check_etcd_health(hostname):
-    '''Check cluster-health of the `etcd` cluster.
-
-    This module is only runnable from the salt-master on the bootstrap node.
-
-    Arguments:
-        hostname (str): hostname of an etcd node
-    '''
-    pod_name = 'etcd-{}'.format(hostname)
-    _execute_etcd_command(pod_name, 'cluster-health')
 
 
 def format_san(names):
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -102,27 +102,6 @@ def _execute_etcd_command(pod_name, cmd):
         raise exn
 
 
-@depends(KUBERNETES_PRESENT, fallback_function=deps_missing)
-def add_etcd_node(host, endpoint=None):
-    '''Add a new `etcd` node into the `etcd` cluster.
-
-    This module is only runnable from the salt-master on the bootstrap node.
-
-    Arguments:
-        host (str): hostname of the new etcd node
-        endpoint (str): endpoint of the new etcd node
-    '''
-    if not endpoint:
-        # If we have no endpoint get it from mine
-        node_ip = __salt__['saltutil.runner'](
-            'mine.get', tgt=host, fun='control_plane_ip'
-        )[host]
-        endpoint = 'https://{0}:2380'.format(node_ip)
-
-    pod_name = 'etcd-{}'.format(__salt__['network.get_hostname']())
-    _execute_etcd_command(pod_name, ['member', 'add', host, endpoint])
-
-
 def check_etcd_health(hostname):
     '''Check cluster-health of the `etcd` cluster.
 
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -82,7 +82,7 @@ def _execute_etcd_command(pod_name, cmd):
     api = kubernetes.client.CoreV1Api(api_client=client)
     etcd_command = [
         'etcdctl',
-        '--endpoints', 'https://localhost:2379',
+        '--endpoints', 'https://127.0.0.1:2379',
         '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
         '--key-file', '/etc/kubernetes/pki/etcd/server.key',
         '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
--- a/salt/_modules/metalk8s_etcd.py
+++ b/salt/_modules/metalk8s_etcd.py
@@ -93,6 +93,33 @@ def add_etcd_node(
     return node
 
 
+def urls_exist_in_cluster(
+        peer_urls,
+        endpoint=None,
+        ca_cert='/etc/kubernetes/pki/etcd/ca.crt',
+        cert_key='/etc/kubernetes/pki/etcd/salt-master-etcd-client.key',
+        cert_cert='/etc/kubernetes/pki/etcd/salt-master-etcd-client.crt'):
+    '''Verify if peer_urls exists in cluster.'''
+    if not endpoint:
+        # If we have no endpoint get it from mine
+        endpoint = _get_endpoint_up(
+            ca_cert=ca_cert,
+            cert_key=cert_key,
+            cert_cert=cert_cert
+        )
+
+    with etcd3.client(host=endpoint,
+                      ca_cert=ca_cert,
+                      cert_key=cert_key,
+                      cert_cert=cert_cert,
+                      timeout=TIMEOUT) as etcd:
+        all_urls = []
+        for member in etcd.members:
+            all_urls.extend(member.peer_urls)
+
+    return set(peer_urls).issubset(all_urls)
+
+
 def check_etcd_health(
         minion_id,
         ca_cert='/etc/kubernetes/pki/etcd/ca.crt',
--- a/salt/_modules/metalk8s_etcd.py
+++ b/salt/_modules/metalk8s_etcd.py
@@ -1,0 +1,95 @@
+# -*- coding: utf-8 -*-
+'''
+Module for handling etcd client specific calls.
+'''
+import logging
+
+from salt.exceptions import CommandExecutionError
+from salt.utils.decorators import depends
+
+PYTHON_ETCD_PRESENT = False
+try:
+    import etcd3
+    PYTHON_ETCD_PRESENT = True
+except ImportError:
+    pass
+
+# Timeout when connection to etcd server
+TIMEOUT = 30
+
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_etcd'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def deps_missing(*args, **kwargs):
+    raise CommandExecutionError("python etcd client is not installed")
+
+
+def _get_endpoint_up(ca_cert, cert_key, cert_cert):
+    """Pick an answering etcd endpoint among all etcd servers."""
+    etcd_hosts = __salt__['metalk8s.minions_by_role']('etcd')
+
+    # Get host ip from etcd_hosts
+    endpoints = [
+        __salt__['saltutil.runner'](
+            'mine.get',
+            tgt=host,
+            fun='control_plane_ip'
+        )[host]
+        for host in etcd_hosts
+    ]
+
+    for endpoint in endpoints:
+        try:
+            with etcd3.client(host=endpoint,
+                              ca_cert=ca_cert,
+                              cert_key=cert_key,
+                              cert_cert=cert_cert,
+                              timeout=TIMEOUT) as etcd:
+                etcd.status()
+        except etcd3.exceptions.ConnectionFailedError:
+            pass
+        else:
+            return endpoint
+
+    raise Exception('Unable to find an available etcd member in the cluster')
+
+
+@depends(PYTHON_ETCD_PRESENT, fallback_function=deps_missing)
+def add_etcd_node(
+        peer_urls,
+        endpoint=None,
+        ca_cert='/etc/kubernetes/pki/etcd/ca.crt',
+        cert_key='/etc/kubernetes/pki/etcd/salt-master-etcd-client.key',
+        cert_cert='/etc/kubernetes/pki/etcd/salt-master-etcd-client.crt'):
+    '''Add a new `etcd` node into the `etcd` cluster.
+
+    This module is only runnable from the salt-master on the bootstrap node.
+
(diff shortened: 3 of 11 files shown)


images/salt-master/Dockerfile
```diff
@@ -24,7 +24,9 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
  && yum clean expire-cache \
  && yum install -y epel-release \
  && yum install -y python2-kubernetes salt-master salt-api salt-ssh openssh-clients \
+ && yum install -y python-pip \
  && yum clean all \
+ && pip install etcd3 \
```
NicolasT, line 29: We'll need to package this up. Maybe (*maybe*) not in scope of this PR, but then at least requires a tracking ticket, even part of alpha4. No `pip install` in Docker images.

salt/_modules/metalk8s.py
```diff
@@ -33,7 +33,7 @@ def deps_missing(*args, **kwargs):
     raise CommandExecutionError("Kubernetes python client is not installed")
 
 
-@depends('KUBERNETES_PRESENT', fallback_function=deps_missing)
+@depends(KUBERNETES_PRESENT, fallback_function=deps_missing)
```
NicolasT: @slaperche-scality Weigh in please?

salt/_modules/metalk8s.py
```diff
@@ -82,7 +82,7 @@ def _execute_etcd_command(pod_name, cmd):
     api = kubernetes.client.CoreV1Api(api_client=client)
```
NicolasT: This whole procedure should no longer be required, no?

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,98 @@
...
+    return __virtualname__
+
+
+def deps_missing(*args, **kwargs):
+    raise CommandExecutionError("python etcd client is not installed")
+
+
+def _get_endpoint_up(ca_cert, cert_key, cert_cert):
+    """Find an etcd endpoint among all servers declared in mine"""
+    endpoints = __salt__['saltutil.runner'](
```
NicolasT: No. We know which servers are part (or are supposed to be part of) the `etcd` ensemble. No need to go probe every single minion or whatnot.

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,98 @@
...
+    This module is only runnable from the salt-master on the bootstrap node.
+
+    Arguments:
+        host (str): hostname of the new etcd node
+        endpoint (str): host server in the etcd cluster
+                        IP is expected, not URL
+    '''
+    if not endpoint:
+        # If we have no endpoint get it from mine
+        endpoint = _get_endpoint_up(
```
NicolasT, line 80: One can provide a set of endpoints to an `etcd` client. We know on which minions `etcd` is supposed to run. We can fetch their `control_plane_ip`s from the mine. Hence we can construct such connection string.

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,98 @@
...
+
+    etcd = etcd3.client(
+        host=endpoint,
+        ca_cert=ca_cert,
+        cert_key=cert_key,
+        cert_cert=cert_cert,
+    )
+
+    node = etcd.add_member(peer_urls)
+    etcd.close()
```
NicolasT: Needs a `try`/`finally` then.

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -90,6 +90,15 @@ Bring bootstrap minion to highstate:
     - salt: Sync bootstrap minion
     - salt: Deploy CA role on bootstrap minion
 
+Generate etcd client certs for salt master:
+  salt.state:
+  - sls:
+    - metalk8s.salt.master.certs
+    - metalk8s.kubernetes.ca.etcd.advertised
```
NicolasT: Why? This should be the case already by virtue of `Deploy CA role on bootstrap minion`. You should `require` that stage to be done before this one can run.

salt/metalk8s/salt/master/certs/etcd-client.sls
```diff
@@ -0,0 +1,33 @@
...
+    - require:
+      - pkg: Install m2crypto
+
+Generate salt master etcd client certificate:
+  x509.certificate_managed:
+    - name: /etc/kubernetes/pki/etcd/salt-master-etcd-client.crt
+    - public_key: /etc/kubernetes/pki/etcd/salt-master-etcd-client.key
+    - ca_server: {{ pillar['metalk8s']['ca']['minion'] }}
+    - signing_policy: {{ etcd.cert.apiserver_client_signing_policy }}
+    - CN: kube-apiserver-salt-master-client
```
NicolasT: Why `kube-apiserver`?

salt/metalk8s/salt/master/certs/etcd-client.sls
```diff
@@ -0,0 +1,33 @@
...
+    - O: "system:masters"
```
NicolasT: No.

salt/_modules/metalk8s.py
slaperche-scality: @NicolasT ?

IIRC, this pattern was introduce by @sayf-eddine-scality in acb5f1665480cae91209ec4a9d7f5b6264bccc08, for `add_etcd_node`, and then reuse by @gdemonet for `wait_apiserver`.

But yeah, the fix look correct to me: according to the doc, if you pass a string it should be a module name, but here we're using a boolean so we should pass the variable.

salt/_modules/metalk8s.py
```diff
@@ -82,7 +82,7 @@ def _execute_etcd_command(pod_name, cmd):
...
     etcd_command = [
         'etcdctl',
-        '--endpoints', 'https://localhost:2379',
+        '--endpoints', 'https://127.0.0.1:2379',
```
slaperche-scality: About commit title: s/127.1/127.0.0.1/

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,98 @@
...
+            etcd.status()
+            etcd.close()
+        except etcd3.exceptions.ConnectionFailedError:
+            pass
+        else:
+            return endpoint
+    raise Exception('Fail to find an etcd srv up in the cluster')
+
+
+@depends(PYTHON_ETCD_PRESENT, fallback_function=deps_missing)
```
slaperche-scality: can do `@depends('etcd3', fallback_function=deps_missing)`, no?

salt/_modules/metalk8s_etcd.py
slaperche-scality: +1, or use a with statement (`contextlib.closing` may help if needed)

salt/_modules/metalk8s.py
ramanduh: It is still used for now. This is used by the upgrade orchestrator in order to check the etcd heath. The issue is that `etcd3` does not provide a method to check the cluster health. We have the status method (https://python-etcd3.readthedocs.io/en/latest/_modules/etcd3/client.html#Etcd3Client.status) but it seems, that it does not return the cluster health

salt/_modules/metalk8s.py
ramanduh: It is just a shorthand but ok, I will update

salt/_modules/metalk8s.py
NicolasT: That's because you can't "ask" the cluster about the health of the cluster (you can't do that in a quorum-based distributed system...). All you can do is have a client request/figure out the status of cluster nodes, and derive a cluster health from it.

NicolasT: https://github.com/etcd-io/etcd/blob/master/etcdctl/ctlv2/command/cluster_health.go

salt/_modules/metalk8s_etcd.py
ramanduh: I'm not sure to understand this one, how do we know which servers are part of the etcd cluster? I did not find anything special in grains. Or do you mean query the kubeapi with `python-kubernetes` and find out which node have the etcd roles?

salt/_modules/metalk8s_etcd.py
ramanduh: Should work according to the docs, but I used this syntax to be consistent with the other custom modules

salt/_modules/metalk8s_etcd.py
ramanduh: Right, will fix

salt/metalk8s/orchestrate/bootstrap/init.sls
ramanduh: Ok, I will fix

salt/metalk8s/salt/master/certs/etcd-client.sls
ramanduh: Will fix, copy/paste error

salt/metalk8s/salt/master/certs/etcd-client.sls
ramanduh: Will remove

salt/_modules/metalk8s_etcd.py
NicolasT: There's an existing macro to derive this from the pillar.

salt/metalk8s/orchestrate/deploy_new_node.sls
```diff
@@ -63,7 +63,7 @@ Kill kube-controller-manager on all master nodes:
 
 Register the node into etcd cluster:
   module.run:
-    - metalk8s.add_etcd_node:
+    - metalk8s_etcd.add_etcd_node:
```
TeddyAndrieux: In my opinion we should have an idempotent state, to not trying to add when node already added
(Not tested with this new version with `python etcd3` but) Using `etcdctl` if we try to add a node already in the cluster we got an error
```
                                  An exception occurred in this state: Traceback (most recent call last):
                                    File "/usr/lib/python2.7/site-packages/salt/state.py", line 1919, in call
                                      **cdata['kwargs'])
                                    File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1918, in wrapper
                                      return f(*args, **kwargs)
                                    File "/usr/lib/python2.7/site-packages/salt/utils/decorators/__init__.py", line 558, in _decorate
                                      return self._call_function(kwargs)
                                    File "/usr/lib/python2.7/site-packages/salt/utils/decorators/__init__.py", line 263, in _call_function
                                      raise error
                                  CommandExecutionError: etcdctl: membership: peerURL exists
```

If the python module well handle with NO error for me it's ok to keep a `module.run` if it's not we need a custom state to check list of member and do nothing if member already there

slaperche-scality: >  a custom state to check list of member and do nothing if member already there

This can be racy no?
If possible it's better to do it and handle the error gracefully.

salt/_modules/metalk8s.py
ramanduh: Ok, will fix

salt/_modules/metalk8s_etcd.py
ramanduh: will fix

salt/_modules/metalk8s_etcd.py
ramanduh: I will remove the depends and set up the virtual method instead

salt/_modules/metalk8s_etcd.py
ramanduh: The python-etcd3 class constructor support only one endpoint, so we have to pass it one IP address. I updated the _get_endpoint_up to feed the endpoint only with minions with role `etcd`

salt/metalk8s/orchestrate/deploy_new_node.sls
ramanduh: Oups, sorry @slaperche-scality I did not see your comment, so I went for the custom state (could be reverted), but to be honest I don't understand how it could be racy as the IP address is uniq per node, I may miss something

slaperche-scality: The node could be added/removed between the time you check and the time you act based on this information.

E.g.:
- your code lists the the cluster members, the node isn't in it
- something/someone add the node to the cluster (someone else also run the state).
- your code adds the node: error!

But maybe such case is not possible with Salt (can't run two states concurrently), then it's fine :slightly_smiling_face:

salt/_modules/metalk8s.py
```diff
@@ -11,10 +11,6 @@
 
 KUBERNETES_PRESENT = False
 try:
-    import kubernetes.client
```
NicolasT, line 14: So no more K8s imports? Then the whole `KUBERNETES_PRESENT` and `@depends` on it etc can be removed, no?

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+            'mine.get',
+            tgt=host,
+            fun='control_plane_ip'
+        )[host]
+        for host in etcd_hosts
+    ]
+
+    for endpoint in endpoints:
+        try:
+            with closing(etcd3.client(host=endpoint,
```
NicolasT: `etcd3.Etcd3Client` is a context-manager by itself, so you can use

```python
with etcd3.client(...) as etcd:
    etcd.foo()
```

https://github.com/kragniz/python-etcd3/blob/master/etcd3/client.py#L175

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+                                      ca_cert=ca_cert,
+                                      cert_key=cert_key,
+                                      cert_cert=cert_cert)) as etcd:
```
NicolasT: You'll want to use some reasonable timeout here. The default is `None`, which is never a good idea (even if in reality that means a 300s timeout or whatever, way too high).

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+                etcd.status()
+        except etcd3.exceptions.ConnectionFailedError:
+            pass
+        else:
+            return endpoint
+
+    raise Exception('Fail to find an etcd srv up in the cluster')
```
NicolasT: `'Unable to find an available etcd member in the cluster'` or something along those lines (errors can bubble up to the user)

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+    '''
+    if not endpoint:
+        # If we have no endpoint get it from mine
+        endpoint = _get_endpoint_up(
+            ca_cert=ca_cert,
+            cert_key=cert_key,
+            cert_cert=cert_cert
+        )
+
+    with closing(etcd3.client(host=endpoint,
```
NicolasT: Same as above (context-manager)

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+                              ca_cert=ca_cert,
+                              cert_key=cert_key,
+                              cert_cert=cert_cert)) as etcd:
+        node = etcd.add_member(peer_urls)
```
NicolasT, line 91: Same as above (timeout).

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+
+    with closing(etcd3.client(host=endpoint,
+                              ca_cert=ca_cert,
+                              cert_key=cert_key,
+                              cert_cert=cert_cert)) as etcd:
+        all_urls = []
+        for member in etcd.members:
+            all_urls.extend(member.peer_urls)
+
+    return not bool(set(peer_urls) - set(all_urls))
```
NicolasT: `return set(peer_urls).issubset(all_urls)` ?

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+        hostname,
+        ca_cert='/etc/kubernetes/pki/etcd/ca.crt',
+        cert_key='/etc/kubernetes/pki/etcd/salt-master-etcd-client.key',
+        cert_cert='/etc/kubernetes/pki/etcd/salt-master-etcd-client.crt'):
+    '''Check cluster-health of the `etcd` cluster.
+
+    This module is only runnable from the salt-master on the bootstrap node.
+
+    Arguments:
+        hostname (str): hostname of an etcd node
```
NicolasT: It's not really the hostname, but rather the minion id.

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+    '''
+    # Get host ip from host name
+    endpoint = __salt__['saltutil.runner'](
+        'mine.get', tgt=hostname, fun='control_plane_ip'
+    )[hostname]
+
+    # Get all members
+    try:
+        with closing(etcd3.client(host=endpoint,
```
NicolasT: See above (context-manager and timeout)

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+            etcd_members = list(etcd.members)
+    except Exception:  # pylint: disable=broad-except
+        log.exception("cluster may be unhealthy: failed to list members")
+        raise
+
+    unhealthy_member = 0
+    for member in etcd_members:
+        etcd_url = urlparse(member.client_urls[0])
+        try:
+            with closing(etcd3.client(host=etcd_url.hostname,
```
NicolasT: You know the drill

salt/_modules/metalk8s_etcd.py
```diff
@@ -0,0 +1,168 @@
...
+                                      cert_cert=cert_cert)) as etcd:
+                etcd.status()
+        except Exception:  # pylint: disable=broad-except
+            log.exception(
+                "failed to check the health of member %s", member.name
+            )
+            unhealthy_member += 1
+
+    # Raise on error as this function will be called by module.run in sls file
+    if unhealthy_member == len(list(etcd_members)):
```
NicolasT: It's a list already.

salt/_states/metalk8s_etcd.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def __virtual__():
+    if 'metalk8s_kubernetes.ping' not in __salt__:
+        return False, '`metalk8s_kubernetes.ping` not available'
+    else:
+        return __virtualname__
+
+
+def member_present(name, peer_urls):
```
NicolasT, line 17: This should support `test` mode.

salt/_states/metalk8s_etcd.py
```diff
@@ -0,0 +1,43 @@
...
+    # Check that peer urls does not exist
+    if __salt__['metalk8s_etcd.urls_exist_in_cluster'](peer_urls):
+        ret['result'] = True
+        ret['comment'] = "Peer URLs: {} already exists".format(
+            ' ,'.join(peer_urls)
+        )
+        return ret
+
+    # Add node
+    member = __salt__['metalk8s_etcd.add_etcd_node'](peer_urls)
```
NicolasT: Given below the `result` is always set to `True`, I guess this call can raise.

Rather than raising and relying on that, consider catching the exception and handle `result` and such accordingly.

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -90,6 +90,16 @@ Bring bootstrap minion to highstate:
     - salt: Sync bootstrap minion
     - salt: Deploy CA role on bootstrap minion
 
+Generate etcd client certs for salt master:
```
NicolasT, line 93: Is there a reason to have this here, and not have `metalk8s.salt.master.certs` included in `metalk8s.salt.master`, which is deployed on the `bootstrap` role minion?

salt/metalk8s/roles/etcd/registered.sls
```diff
@@ -0,0 +1,17 @@
+{# Should be run by the orchestrator runner #}
```
NicolasT: No, this doesn't belong here. `metalk8s.roles.*` are roles defined for specific minions. This SLS, which is to be executed on the master, belongs... somewhere else. It's not part of the `etcd` node-role.

NicolasT (changes requested):

salt/_modules/metalk8s.py
ramanduh: we can still use the `@depends` as the `wait_apiserver` function still need the `metalk8s_kubernetes` salt module not the kubernetes import, but true, my `try...except` bloc here does not make sense, will fix it

salt/metalk8s/orchestrate/bootstrap/init.sls
ramanduh: In fact, when the state `metalk8.salt.master` is called, the CA is not deployed yet, so we can't generate the certificate at this moment

salt/metalk8s/roles/etcd/registered.sls
ramanduh: moved

ramanduh:
@bert-e reset

ramanduh:
fixed

NicolasT (approved):
:ship:

ramanduh:
@bert-e approve

Issue #609, opened 2019-02-18
containerd: rework image injection


NicolasT:
Instead of `ctr cri load`, can use `ctr -n k8s.io image import -` (or using a filename).

Given that, we can import image archives created by `skopeo` instead of requiring the `docker pull` / `docker save` kludge.

PR #1109, opened 2019-05-14
Rework offline image serving

### LINKED ISSUE #1049: Rework 'offline image' serving
To support fully-offline installations, we currently provide a Docker *registry* in a deployment, and use `docker pull`/`docker save` to generate image tarballs, which are then pushed into the registry at deployment time.

This has several drawbacks:
- `docker pull/save` to (basically) fetch files from the internet is not necessarily efficient
- Due to `docker save`, layers that are shared across multiple images still end up duplicated multiple times in the various tarballs (which e.g. in case of the CentOS base image means duplicating roughly 200MB several times), inflating the size of the distribution ISO
- The registry image needs to be maintained
- Due to the images being copied from the ISO into the registry, this adds extra space requirements to the bootstrap node
- Currently, the deployed registry is writable by anybody :fearful: 

Ideally, we should be able to deploy a (fake) registry using the `nginx` instance also used to distribute RPM packages/Yum repositories that's part of the bootstrap infrastructure, serving files directly from the mounted ISO.

Given https://github.com/NicolasT/static-container-registry/ this is now possible.

Delivery of this task would include:

- [x] Replace `docker pull` / `docker save` in the build process by `skopeo copy ...` according to the requirements as laid out in the README file of https://github.com/NicolasT/static-container-registry/
- [x] Run `hardlinks -c` on the images directory to remove duplicate files
- [x] Integrate the `static-container-registry.py` script in the build infrastructure to, after `skopeo copy`ing all images into the ISO root, generate a registry locations configuration. This should

  * use `metalk8s-2.0` (or similar) as the `--name-prefix`
  * use `$metalk8s_2_0_images` (or similar) as the `--server-root`
- [x] During deployment, copy the generated registry locations file into an `nginx` `include` directory, e.g. `99-metalk8s-2.0-registry.conf`
- [x] Generate another `include` file (`90-metalk8s-2.0-registry-config.conf` ?) to specify the value of `$metalk8s_2_0_images`, i.e. `set $metalk8s_2_0_images "/srv/scality/metalk8s-2.0/...."`
- [x] Ensure `nginx` gets restarted to take these changes into account (by setting an `annotation` in the Pod manifest to some new value, see #937)
- [x] No longer deploy and populate a `registry`
- [x] Ensure all containers are now pulled from this repository

slaperche-scality:
**Component**: buildchain

**Context**:

To support fully-offline installations, we currently provide a Docker *registry* in a deployment, and use `docker pull`/`docker save` to generate image tarballs, which are then pushed into the registry at deployment time.

This has several drawbacks:
- `docker pull/save` to (basically) fetch files from the internet is not necessarily efficient
- Due to `docker save`, layers that are shared across multiple images still end up duplicated multiple times in the various tarballs (which e.g. in case of the CentOS base image means duplicating roughly 200MB several times), inflating the size of the distribution ISO
- The registry image needs to be maintained
- Due to the images being copied from the ISO into the registry, this adds extra space requirements to the bootstrap node
- Currently, the deployed registry is writable by anybody :fearful: 

Ideally, we should be able to deploy a (fake) registry using the `nginx` instance also used to distribute RPM packages/Yum repositories that's part of the bootstrap infrastructure, serving files directly from the mounted ISO.

**Summary**:

In the buildchain:
- use `skopeo` instead of `docker` to save the images
- use hard links to deduplicate the images' layers
- with the  `static-container-registry.py` script, generate an `nginx` config to serve the images

During deployment:
- mount the new configuration files and the images directory into the `nginx` container
- inject the `nginx` image using `ctr cri load` (to avoid chicken or egg problem).
- add the new configuration files as annotation to the pod manifest (to ensure restart on changes)
- get rid of the existing registry

**Acceptance criteria**:

Bootstrap should be working (so you can run a `doit.sh vagrant_up` and check if everything is green).

---

Fixes: #1049


buildchain/buildchain/image.py
```diff
@@ -244,7 +260,6 @@ def task__image_build() -> Iterator[types.TaskDict]:
         version='2.7.1',
         digest='sha256:870474507964d8e7d8c3b53bcfa738e3356d2747a42adad26d0d81ef4479eb1b',
         destination=ISO_IMAGE_ROOT,
-        for_containerd=True,
```
NicolasT, line 247: This image should no longer be required altogether, right?

NicolasT: Never mind, 'not yet' :)

buildchain/buildchain/targets/remote_image.py
```diff
@@ -77,54 +73,94 @@ def fullname(self) -> str:
...
+                'dir:{}'.format(str(self.dirname))
+            ]
         ]
-        # containerd doesn't support compressed images.
-        if not self.for_containerd:
-            actions.append((coreutils.gzip, [filepath], {}))
-        return actions
+
+
+class RemoteTarImage(base.FileTarget):
```
NicolasT: So, not handling #609 in this PR?

salt/metalk8s/repo/installed.sls
```diff
@@ -19,27 +27,64 @@ Generate package repositories nginx configuration:
...
   module.wait:
     - cri.wait_container:
-      - name: {{ package_repositories_name }}
+      - name: {{ repositories_name }}
       - state: running
     - require:
-      - metalk8s: Install package repositories manifest
+      - metalk8s: Install repositories manifest
+
+Deploy container registry nginx configuration:
```
NicolasT: Shouldn't this happen before we deploy the service manifest? I'm aware the `onchanges` acts as a `require`, so things will be executed by Salt 'in the right order', but I think it also makes it easier to read/follow an SLS when tasks are 'in the expected order of execution' within the source.

Same applies to the step below.

NicolasT (dismissed):
One minor request, just moving two blocks up a couple of lines. Other than that LGTM. CI appears to be failing though? Didn't look at logs.

buildchain/buildchain/targets/remote_image.py
slaperche-scality: I tried to, using `skopeo copy … docker-archive:path/to/image` + using `ctr -n k8s.io image import path/to/image` but bootstrap wasn't working. I'll look at the logs tomorrow, to see if it's easy to fix or not.

salt/metalk8s/repo/installed.sls
slaperche-scality: Done, I moved `Deploy container registry nginx configuration` and `Generate container registry configuration` before `Install repositories manifest`

salt/metalk8s/repo/installed.sls
```diff
@@ -61,12 +66,13 @@ Install package repositories manifest:
       - {{ nginx_registry_config_path }}
     - context:
         container_port: {{ repo.port }}
-        image: {{ package_repositories_image }}
+        image: docker.io/library/nginx:1.15.8
```
jbertran, line 72: Could take the `alpine` image for this, since it's uncompressed - just append `-alpine`. Isofunctional as far as a `doit vagrant_up` is concerned.

salt/metalk8s/repo/files/nginx.conf.j2
```diff
@@ -6,4 +6,7 @@ server {
         root   /var/www/repositories;
         autoindex on;
     }
+
+    include conf.d/{{ nginx_registry_config }};
+    include conf.d/{{ nginx_registry }};
```
jbertran: minor: `s/nginx_registry/nginx_registry_images` for clarity?

salt/metalk8s/repo/installed.sls
```diff
@@ -6,17 +6,18 @@
...
-{%- set nginx_conf_root = '/var/lib/metalk8s/package-repositories/' %}
-{%- set nginx_main_conf = nginx_conf_root ~ 'nginx.conf' %}
-{%- set nginx_confd     = nginx_conf_root ~ 'conf.d/' %}
-{%- set nginx_registry        = '99-' ~ saltenv ~ '-registry.conf' %}
-{%- set nginx_registry_config = '90-' ~ saltenv ~ '-registry-config.conf' %}
+{%- set images_path = metalk8s.iso_root_path ~ '/images/' %}
+
+{%- set nginx_confd  = '/var/lib/metalk8s/package-repositories/conf.d/' %}
+{%- set nginx_default_conf = nginx_confd ~ 'default.conf' %}
+{%- set nginx_registry        = '99-' ~ saltenv ~ '-registry.inc' %}
```
jbertran, line 13: Ditto for the actual image conf filename, for clarity.

salt/metalk8s/repo/installed.sls
```diff
@@ -2,20 +2,20 @@
 {%- from "metalk8s/map.jinja" import metalk8s with context %}
 {%- from "metalk8s/map.jinja" import repo with context %}
 
-{%- set package_repositories_name = 'package-repositories' %}
-{%- set package_repositories_version = '1.0.0' %}
-{%- set package_repositories_image = build_image_name('nginx', '1.15.8') %}
+{%- set repositories_name = 'repositories' %}
+{%- set repositories_version = '1.0.0' %}
+{%- set repositories_image = build_image_name('nginx', '1.15.8') %}
```
jbertran, line 7: If we change the image to alpine, should change here too?

jbertran (commented):
LGTM. I'll do a pass on all pulled images to see if there are any lighter alternatives since we're at it.

buildchain/buildchain/targets/remote_image.py
slaperche-scality: Done now :)

salt/metalk8s/repo/installed.sls
slaperche-scality: As discussed on the channel, we should try to be compliant and use CentOS, so no Alpine.

salt/metalk8s/repo/files/nginx.conf.j2
slaperche-scality: I'm now including `*.inc*.

slaperche-scality:
@bert-e reset

slaperche-scality:
@bert-e reset

slaperche-scality:
@bert-e reset

slaperche-scality:
@bert-e force_reset

buildchain/buildchain/targets/image.py
```diff
@@ -31,41 +29,43 @@ def __init__(
...
     @property
-    def uncompressed_filename(self) -> Path:
-        """Name of the uncompressed image on disk."""
-        filename = self.filename
-        if filename.suffix == COMPRESSED_FILEEXT:
-            return filename.with_suffix('')
-        return filename
+    def dirname(self) -> Path:
+        """Directory where to store the image on disk."""
+        return self.dest_dir/self.tag.replace(':', '/')
```
gdemonet: Is the `:` forbidden in an image name? (would make sense)

In any case, unless we want to magically create subdirectories if an image name contains colons, using `self.dest_dir/self.name/self.version` could look more explicit IMHO.

buildchain/buildchain/targets/local_image.py
```diff
@@ -167,41 +163,41 @@ def check_dockerfile_dependencies(self) -> Union[TaskError, bool]:
...
-            actions.append((coreutils.gzip, [filepath], {}))
+            task['actions'].append(self.mkdirs)
+            task['actions'].append([
+                config.SKOPEO,  'copy',
+                '--format', 'v2s2',
+                '--dest-compress',
+                'docker-daemon:{}'.format(self.tag),
+                'dir:{}'.format(str(self.dirname))
+            ])
+            task['targets'] = [self.dirname/'manifest.json']
```
gdemonet: I'm curious: why do you overwrite `task['targets']` here, while you just `append` below?

buildchain/buildchain/targets/local_image.py
```diff
@@ -167,41 +163,41 @@ def check_dockerfile_dependencies(self) -> Union[TaskError, bool]:
...
-            'title': show,
+            'title': lambda _: self.show('IMG BUILD'),
             'doc': 'Build {} container image.'.format(self.name),
-            'actions': self._build_actions(),
         })
+        self._build_actions(task)
         return task
 
-    def _build_actions(self) -> List[types.Action]:
+    def _build_actions(self, task: types.TaskDict) -> None:
```
gdemonet: Maybe we could rename this method (it's not just creating the `actions` actually), or just merge it with the `task` property?

gdemonet (dismissed):
Really minor comments that can be ignored, let's merge this :)

buildchain/buildchain/targets/image.py
slaperche-scality: You're right, better be explicit here.

buildchain/buildchain/targets/local_image.py
slaperche-scality: I'll rework this part (either rename or merge, depends on what's clearer)

buildchain/buildchain/targets/local_image.py
slaperche-scality: No reason, I'll fix it.

slaperche-scality:
@bert-e approve

gdemonet (approved):
LGTM

Issue #1042, opened 2019-04-30
Node Management: We can not deploy a new node if a node is still not deployed correctly


carlito-scality:
Actually, we can not deploy a new node if a node is still not deployed correctly or only registered.

Typical error:
```
nodes_|-ps.pkill_|-function": {"comment": "Running function ps.pkill failed on minions: node1 Function ps.pkill ran on node1, node2, bootstrap.", "name": "ps.pkill", "start_time": "20:40:43.204051", "result": false, "duration": 351.545, "__run_num__": 6, "__sls__": "metalk8s.orchestrate.deploy_new_node", "changes": {"ret": {"node1": false, "node2": null, "bootstrap": {"killed": [15417]}}, "out": "highstate"}
```

NicolasT:
@carlito-scality Could you try the `bugfix/1042-allow-kcm-kill-to-fail` branch in your environment?

MonPote:
@NicolasT I can try

NicolasT:
@MonPote Ping?

carlito-scality:
@NicolasT I just tested your branch ```bugfix/1042-allow-kcm-kill-to-fail``` but there is still the problem.
scenario: I create the node1 but do not deploy it. then I create the node2 and deploy it => 
![image](https://user-images.githubusercontent.com/47394132/57613452-3018cb00-7577-11e9-9d07-b001f0990c17.png)

PR #1112, opened 2019-05-15
Allow `k-c-m` kill to fail on master minions

NicolasT:
**Component**: salt

**Context**: #1042 

**Summary**:
As a work-around for `k-c-m` not picking up new nodes properly (and schedule `DaemonSet` `Pod`s), we kill `k-c-m` on all *master* nodes. However, we were doing so on all (supposedly) *master* nodes, including ones that are down or not (yet) deployed, hence failing the job of deploying a new node.

It should be OK to simply ignore all failures to perform this kill. Worst case an unkilled `k-c-m` keeps running and retains its leader lease, but chances of this happening are rather slim, and it doesn't 'harm' the cluster: re-running the playbook could fix things.

**Acceptance criteria**: 
- Deploy a cluster
- Add two *master* nodes in the API
- Deploy only one of them
- Deployment should succeed. Bonus points for seeing Salt report how it was unable to perform the action on the inactive node/minion in the job results.


carlito-scality (approved):
test ok from UI.
image

NicolasT:
@bert-e approve

PR #1114, opened 2019-05-15
UI - Configure the metalk8s version

carlito-scality:
**Component**: UI

**Context**: 
Actually, the metalk8s version is hardcoded in the UI when creating and deploying a node
**Summary**:
The idea is to add a "version" field into the creation node form the use its value for deployement
**Acceptance criteria**: 
-No regression when creating and deploying a node
-Check if the metalk8s version is correct.

### DIFF
--- a/ui/src/services/api.js
+++ b/ui/src/services/api.js
@@ -78,7 +78,7 @@ export async function createNode(payload) {
     metadata: {
       name: payload.name,
       labels: {
-        'metalk8s.scality.com/version': '2.0'
+        'metalk8s.scality.com/version': payload.version
       },
       annotations: {
         'metalk8s.scality.com/ssh-user': payload.ssh_user,
@@ -122,7 +122,7 @@ export async function createNode(payload) {
   }
 }
 
-export async function deployNode(url, token, node) {
+export async function deployNode(url, token, node, version) {
   try {
     return await axios.post(
       url,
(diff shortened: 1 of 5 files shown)


ui/src/services/api.js
```diff
@@ -131,8 +131,8 @@ export async function deployNode(url, token, node) {
         fun: 'state.orchestrate',
         arg: ['metalk8s.orchestrate.deploy_node'],
         kwarg: {
-          saltenv: 'metalk8s-2.0',
-          pillar: { orchestrate: { node_name: node }}
+          saltenv: `metalk8s-${version}`,
```
NicolasT, line 134: Hmh, not really, this should be the 'version of the cluster'. Also, I think we should default the version of a node to the 'version of the cluster' at that point in time.

However.

We don't have such things as 'version of the cluster' stored anywhere at this point in time :thinking:

ui/src/translations/en.json
```diff
@@ -32,5 +32,6 @@
   "unknown": "Unknown",
   "ready": "Ready",
   "not_ready": "Not Ready",
-  "deploy": "Deploy"
+  "deploy": "Deploy",
+  "version": "Metalk8s Version"
```
NicolasT: `MetalK8s`

ui/src/translations/fr.json
```diff
@@ -32,5 +32,6 @@
   "unknown": "Inconnu",
   "ready": "Prêt",
   "not_ready": "Pas prêt",
-  "deploy": "Déployer"
+  "deploy": "Déployer",
+  "version": "Metalk8s Version"
```
NicolasT: `MetalK8s`

MonPote (approved):
LGTM. (It works but you might want to do nicolast changed.)

carlito-scality:
@bert-e approve

carlito-scality:
duplicate https://github.com/scality/metalk8s/pull/1117

PR #1117, opened 2019-05-16
Node-Management: configure metal k8s version

carlito-scality:
**Component**: UI (duplicate https://github.com/scality/metalk8s/pull/1114 but change the branch name)

**Context**: 
Actually, the metalk8s version is hardcoded in the UI when creating and deploying a node
**Summary**:
The idea is to add a "version" field into the creation node form the use its value for deployement
**Acceptance criteria**: 
-No regression when creating and deploying a node
-Check if the metalk8s version is correct.


MonPote (approved):

carlito-scality:
@bert-e approve

PR #1118, opened 2019-05-16
Salt daemon not bound to control-plane IP

### LINKED ISSUE #766: Salt daemon not bound to control-plane IP
The Salt master daemon binds to `0.0.0.0`, this shoud be the control-plane IP only.

ramanduh:
The Salt master daemon binds to 0.0.0.0, set it to the control-plane
IP only.

Fixes #766

**Component**:

salt master

**Context**: 

Update salt master listening addr

**Summary**:

**Acceptance criteria**: 

Salt master should listen on control plane IP instead of `0.0.0.0`

### DIFF
--- a/salt/metalk8s/salt/master/configured.sls
+++ b/salt/metalk8s/salt/master/configured.sls
@@ -1,12 +1,6 @@
 {%- from "metalk8s/map.jinja" import metalk8s with context %}
-{%- from "metalk8s/map.jinja" import networks with context %}
 
-{%- set ip_candidates = salt.network.ip_addrs(cidr=networks.control_plane) %}
-{%- if ip_candidates %}
-    {%- set salt_api_ip = ip_candidates[0] %}
-{%- else %}
-    {%- set salt_api_ip = '127.0.0.1' %}
-{%- endif %}
+{%- set salt_ip = grains['metalk8s']['control_plane_ip'] -%}
 
 Configure salt master:
   file.managed:
@@ -19,7 +13,7 @@ Configure salt master:
     - backup: false
     - template: jinja
     - defaults:
-        salt_api_ip: "{{ salt_api_ip }}"
+        salt_ip: "{{ salt_ip }}"
 
 Configure salt master roots paths:
   file.serialize:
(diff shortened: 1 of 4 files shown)


salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
```diff
@@ -1,3 +1,5 @@
+interface: {{ salt_api_ip }}
```
NicolasT: Whilst this is technically correct, maybe we shouldn't call this `salt_api_ip`, but `salt_ip` and rename the other occurrence as well?

NicolasT (changes requested):

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
ramanduh: True

salt/metalk8s/salt/master/configured.sls
```diff
@@ -3,9 +3,9 @@
 
 {%- set ip_candidates = salt.network.ip_addrs(cidr=networks.control_plane) %}
```
NicolasT: Oh, these are leftovers... this `ip_candidates`, `if ip_candidates` (and fallback to `127.0.0.1`) etc. should be no longer required. `salt_ip` should simply be the `metalk8s:control_plane_ip` *grain* of the minion (also in the other instance below).

I think it makes sense to also clean this up as part of this PR.

NicolasT (changes requested):

ramanduh:
@bert-e reset

ramanduh:
@bert-e approve

salt/metalk8s/salt/master/installed.sls
```diff
@@ -7,9 +7,9 @@
 
 {%- set ip_candidates = salt.network.ip_addrs(cidr=networks.control_plane) %}
```
NicolasT: Same as above, should be using the grain.

NicolasT (changes requested):

salt/metalk8s/salt/master/installed.sls
ramanduh: oups

ramanduh:
@bert-e reset

NicolasT (dismissed):

ramanduh:
@bert-e reset

NicolasT (approved):

PR #1119, opened 2019-05-16
Monitoring: Deploy prometheus-operator

alexandre-allard:
**Component**: Prometheus-operator

**Context**: Deployment of the prometheus-operator monitoring components

**Summary**: First step of the monitoring deployment

**Acceptance criteria**: Ability to deploy prometheus-operator through salt using `metalk8s.monitoring.prometheus-operator.deployed` state, prometheus-operator deployment must be up and running.


salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2373,3 +2405,139 @@ def node_uncordon(node_name, **kwargs):
...
+
+def get_crd_plural_from_kind(kind, **kwargs):
+    plural = None
+
+    crds = customresourcedefinition(**kwargs)
+
+    for crd in crds.get('items', []):
+        crd_names = crd.get('spec', {}).get('names', {})
+        if kind == crd_names.get('kind'):
+            plural = crd_names.get('plural')
```
gdemonet: You could have simply done something like:

```suggestion
            return crd_names.get('plural')
```

And get rid of the `break` statement and the `plural` variable

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1572,3 +1576,205 @@ def node_uncordoned(name, **kwargs):
         the name of the node
     '''
     return _node_unschedulable(name, False, **kwargs)
+
+
+def _extract_custom_resource_data(name, namespace, body, source, **kwargs):
+    if body and source:
+        raise salt.exceptions.InvalidConfigError(
```
gdemonet: Is this an `InvalidConfig`? You catch them yourself later, so I guess it doesn't matter much, but I think some native `ValueError` is more adapted to the "bad arguments" situation you have here (and below). To me, the _Salt invalid config_ really is about minion/master configuration, but I might be wrong.

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1572,3 +1576,205 @@ def node_uncordoned(name, **kwargs):
...
+    except ValueError:
+        raise salt.exceptions.InvalidConfigError(
+            "malformed or undefined 'apiVersion' in object definition"
+        )
+
+    plural = __salt__['metalk8s_kubernetes.get_crd_plural_from_kind'](
+        kind, **kwargs)
+    if plural is None:
+        raise salt.exceptions.InvalidConfigError(
+            "'{0}' object kind not found in API".format(kind)
```
gdemonet: So, this error message isn't really explaining the failed check above... In any case, is it really mandatory for CRDs to define a `plural`? I'm not sure it's something that the Operator SDK generates, for example.

buildchain/buildchain/salt_tree.py
```diff
@@ -194,6 +194,8 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
     Path('salt/metalk8s/macro.sls'),
     Path('salt/metalk8s/map.jinja'),
 
+    Path('salt/metalk8s/monitoring/prometheus-operator/deployed.sls'),
```
gdemonet: I think we want monitoring to be a part of `addons`, but not sure actually. @NicolasT @TeddyAndrieux ?

gdemonet (commented):
Few comments/questions

buildchain/buildchain/salt_tree.py
alexandre-allard: Yeah, I was not sure where to put this stuff.

salt/_modules/metalk8s_kubernetes.py
alexandre-allard: Yeah right, at the beginning it was part of another function, I forgot to change it when I moved this.
I will change that.

salt/_states/metalk8s_kubernetes.py
alexandre-allard: I'm not sure if it's mandatory, maybe if you create a CRD without plural it'll fallback on `kind` value.
I'll check that.

alexandre-allard: Looks like it's not possible to create a CRD without `spec.names.plural`:
```
# kubectl apply -f 0prometheus-operator-0servicemonitorCustomResourceDefinition.yaml
error: error validating "0prometheus-operator-0servicemonitorCustomResourceDefinition.yaml": error validating data: ValidationError(CustomResourceDefinition.spec.names): missing required field "plural" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionNames; if you choose to ignore these errors, turn validation off with --validate=false
```
Maybe the error message is not clear enough, but it reflects the test above which is: "Is there a plural for this kind ? if not, this kind does not exist".

salt/_states/metalk8s_kubernetes.py
alexandre-allard: I'll replace with a ValueError exception.

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2373,3 +2405,139 @@ def node_uncordon(node_name, **kwargs):
...
     return node_set_unschedulable(node_name, False, **kwargs)
+
+
+def get_crd_plural_from_kind(kind, **kwargs):
+    plural = None
+
+    crds = customresourcedefinition(**kwargs)
+
+    for crd in crds.get('items', []):
+        crd_names = crd.get('spec', {}).get('names', {})
```
slaperche-scality, line 2414: For now it's use only once, but if this pattern (multiple `get`) becomes more widespread, we will want to use a recursive `get` to make the code easier to read, e.g.:

```python
def get_rec(d, keys, sep='.', default=None):
    try:
        return functools.reduce(operator.getitem, keys.split(sep), d)
    except KeyError:
        return default

def get_crd_plural_from_kind(kind, **kwargs):
    for crd in customresourcedefinition(**kwargs).get('items', []):
        if get_rec(crd, 'spec.names.kind') == kind:
            return get_rec(crd, 'spec.names.plural')
```

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2408,17 +2408,14 @@ def node_uncordon(node_name, **kwargs):
...
 
     for crd in crds.get('items', []):
         crd_names = crd.get('spec', {}).get('names', {})
         if kind == crd_names.get('kind'):
-            plural = crd_names.get('plural')
-            break
+            return crd_names.get('plural')
 
-    return plural
+    return None
```
gdemonet, line 2418: Ok for being explicit, but it's not something required (by default, a Python function with no `return` statement will return `None`)

salt/_modules/metalk8s_kubernetes.py
gdemonet: There might be one in Salt already, since the `__salt__` "dict" uses the same kind of magic for its `__getitem__` (e.g. `__salt__["network.ip_addr"]`)

slaperche-scality: Would be nice if they expose it, but maybe it's part of their internals.

gdemonet (approved):
LGTM - we'll want to change the namespace, to group addons from MetalK8s together, but we can do it later I guess

salt/_modules/metalk8s_kubernetes.py
jbertran: Most linters will yell at you for mixing explicit and implicit returns in a function though.

alexandre-allard:
@bert-e approve

PR #1031, opened 2019-04-25
Add automated tests for cluster expansion

slaperche-scality:

PR #1132, opened 2019-05-20
Fix SSH config file generated by Terraform

### LINKED ISSUE #1038: Fix multi-nodes env in CI
Currently the multi-nodes CI environment *does not* spawn extra nodes (just the bootstrap).
Consequently, we can't run cluster expansion test in the CI.

This is explained by a comment in `eve/workers/openstack-multiple-nodes/terraform/nodes.tf`:

```
  # FIXME: this syntax does not work (but will in v0.12)
  # see https://github.com/hashicorp/terraform/issues/17048
  # nodes = ["${openstack_compute_instance_v2.nodes.*.network.0.fixed_ip_v4}"]
```

So we either have to wait for the release of the v0.12, or find a workaround.

gdemonet:
**Component**: ci, terraform

**Context**:

Tests for cluster expansion (#1031) could not run using the CI Terraform-generated SSH config file, since it did not include information about _extra_ nodes due to limitations in the Terraform syntax.

**Summary**:

- Upgrade to Terraform v0.12 (install a RC version, still no official release)
- Fix the generation of SSH config files from Terraform with the new syntax

**Acceptance criteria**: 

The SSH config files generated by Terraform should be valid and allow to connect to host `node1` from:

- the host that executed `terraform apply`
- the spawned bastion

---

Fixes: #1038

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -214,6 +214,11 @@ stages:
       - ShellCommand: *retrieve_iso_checksum
       - ShellCommand: *check_iso_checksum
       - ShellCommand:
+          name: Set-up Terraform provider plugins for v0.12
+          command: bash scripts/setup-providers.sh
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
+      - ShellCommand:
           name: Init terraform
           command: terraform init
           workdir: build/eve/workers/openstack-multiple-nodes/terraform/
--- a/eve/workers/openstack-multiple-nodes/requirements.sh
+++ b/eve/workers/openstack-multiple-nodes/requirements.sh
@@ -5,6 +5,7 @@ yum install -y epel-release
 PACKAGES=(
     curl
     git
+    make
     python36-pip
     unzip
 )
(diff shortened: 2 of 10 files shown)


eve/main.yml
```diff
@@ -213,6 +213,11 @@ stages:
       - ShellCommand: *retrieve_iso
       - ShellCommand: *retrieve_iso_checksum
       - ShellCommand: *check_iso_checksum
+      - ShellCommand:
+          name: Set-up Terraform provider plugins for v0.12
+          command: bash scripts/setup-providers.sh
```
charlesprost, line 218: Don't we want to run this during the `requirements.sh` step?

charlesprost, line 218: Don't we want to run this during the `requirements.sh` step?

charlesprost (approved):

eve/main.yml
gdemonet: I left it there so that we see it more easily than in the `requirements.sh` script. The rationale is that we will want to remove that step as soon as 0.12 is officially released (then, providers will be fetched by `terraform init`).

gdemonet:
@bert-e approve

PR #1142, opened 2019-05-22
Deploy Prometheus, AlertManager, Kube-State-Metrics & Grafana

alexandre-allard:
**Component**: prometheus, alertmanager, kube-state-metrics & grafana

**Context**: Deployment of the monitoring stack

**Summary**: Deployment of prometheus, alertmanager, kube-state-metrics & grafana with its dashboards

**Acceptance criteria**: Ability to deploy the components aforementioned through salt using metalk8s.addons.{prometheus,alertmanager,kube-state-metrics,grafana}.deployed states, all components must be up and running.

There is still an ongoing issue with alertmanager, but we found the root cause and I'm trying to fix it. The PR can still be reviewed.


salt/metalk8s/addons/prometheus/deployed.sls
```diff
@@ -0,0 +1,1464 @@
...
+    prometheus: k8s
+  name: k8s
+  namespace: monitoring
+spec:
+  alerting:
+    alertmanagers:
+    - name: alertmanager-main
+      namespace: monitoring
+      port: web
+  baseImage: quay.io/prometheus/prometheus
```
jbertran: Ditto.

salt/metalk8s/addons/alertmanager/deployed.sls
```diff
@@ -0,0 +1,60 @@
...
+---
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  labels:
+    alertmanager: main
+  name: main
+  namespace: monitoring
+spec:
+  baseImage: quay.io/prometheus/alertmanager
```
jbertran: Any chance we can freeze that version? I realize we're importing these almost directly from the prometheus repo so we might not want to do this.

salt/metalk8s/addons/alertmanager/deployed.sls
```diff
@@ -15,7 +27,7 @@ metadata:
   name: main
   namespace: monitoring
 spec:
-  baseImage: quay.io/prometheus/alertmanager
+  baseImage: {{ build_image_name('alertmanager') }}
```
jbertran: Should be frozen here as well if we can freeze it in the manifest.

salt/metalk8s/addons/prometheus/deployed.sls
```diff
@@ -137,7 +149,7 @@ spec:
     - name: alertmanager-main
       namespace: monitoring
       port: web
-  baseImage: quay.io/prometheus/prometheus
+  baseImage: {{ build_image_name('prometheus') }}
```
jbertran: Ditto.

jbertran (commented):
Minor stuff on version freezing for some images, otherwise LGTM

salt/metalk8s/addons/alertmanager/deployed.sls
alexandre-allard: Already done if you check `version`  key few lines below.

salt/metalk8s/addons/prometheus/deployed.sls
```diff
@@ -0,0 +1,1478 @@
...
+        message: Errors while reconciling Prometheus in {{ $labels.namespace }} Namespace.
+      expr: |
+        rate(prometheus_operator_node_address_lookup_errors_total{job="prometheus-operator"}[5m]) > 0.1
+      for: 10m
+      labels:
+        severity: warning
+{%- endraw %}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+items:
```
gdemonet: Do we really need to do this kind of thing? Because if we do, Prometheus will never be able to watch resources in new namespaces, such as the ones used for deploying Solutions. I think we want to move to using `ClusterRole`/`ClusterRoleBinding`.

salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -227,8 +267,16 @@ def render(yaml_data, saltenv='', sls='', **kwargs):
 
     data = yaml.load_all(yaml_data, Loader=salt.utils.yaml.SaltYamlSafeLoader)
 
+    objects = []
+    for obj in data:
+        if not obj:
+            continue
+        if isinstance(obj.get('items'), list):
```
gdemonet, line 288: Did you really have to do this? IIUC, the idea is to push a `RoleBindingList`, which is supported by the K8s API: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#rolebindinglist-v1-rbac-authorization-k8s-io

In any case, I believe we'll want to replace this list with a `ClusterRoleBinding` instead.

salt/metalk8s/addons/grafana/deployed.sls
```diff
@@ -0,0 +1,7513 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: grafana
+  namespace: monitoring
+---
+{%- raw %}
+apiVersion: v1
+items:
+- apiVersion: v1
```
gdemonet: :poop:

Agreed that it's done automatically with the bundler tool, but we are just pushing a bunch of JSON dashboards into ConfigMaps, let's just store the JSON files somewhere and include them in this YAML, instead of having a 7k lines file.

Also, dashboard addition will need to be "standardized" so that Solutions can add theirs as well (maybe we'll need to rework something in the Grafana Pod spec).

salt/metalk8s/addons/kube-state-metrics/deployed.sls
```diff
@@ -0,0 +1,289 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: kube-state-metrics
+subjects:
+- kind: ServiceAccount
+  name: kube-state-metrics
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
gdemonet: Why do we have a `RoleBinding` above for the same SA?

gdemonet (changes requested):
Few comments around Grafana dashboards (really think we need to get them out of the main YAML, too verbose), and RoleBindings. Those may exist in the upstream YAMLs, but I think we can do better on our side.

Also, just noticed we stashed everything under `metalk8s.addons`, but I think it would make sense to group them under `metalk8s.addons.monitoring`, to separate from the UI, and later, other cluster services such as log aggregation.

salt/_renderers/metalk8s_kubernetes.py
gdemonet: Oh and same for `ConfigMapList`: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#configmaplist-v1-core

alexandre-allard: I did this because in the renderer we're generating the name/id of the state with the name & kind of the objet we want to create. When we've a *list kind, we don't have a name so there is nothing to create an unique name/id.

gdemonet: My point was we could add the `*_list_[pre|ab]sent` states since they exist in the API, but I guess it's simpler this way :)

alexandre-allard:
@bert-e reset

alexandre-allard:
@bert-e reset

jbertran (approved):

alexandre-allard:
@bert-e reset

alexandre-allard:
@bert-e force_reset

jbertran (dismissed):

salt/metalk8s/addons/kube-state-metrics/deployed.sls
gdemonet: The two roles don't have the same permissions, let's look at this later.

salt/metalk8s/addons/grafana/deployed.sls
gdemonet: So, let's leave it as is for now. However, it would be interesting to consider using something like [`kustomize`](https://kustomize.io/) in the future (they have a feature to generate ConfigMaps from files).

salt/metalk8s/addons/prometheus/deployed.sls
gdemonet: Let's look at permissions later.

gdemonet (dismissed):
I trust you guys that it works... Didn't take time to test.

jbertran (approved):
Third time's the charm?

alexandre-allard:
@bert-e approve

Issue #924, opened 2019-04-03
Mount all iso listed in BootstrapConfig


TeddyAndrieux:
BootstrapConfig should include a list of iso that a salt modules/states automatically mount all of them in the right place.

**When ?**
It's basically the first action the user should do, call the right salt state that "prepare" the iso

**Where ?**
Proposition:
Should be at `/srv/scality/metalk8s-<version>` but we need to be able to know the version of the iso before extracting.
We could use something like this:
```
isoinfo -x /PRODUCT.TXT\;1 -i _build/metalk8s.iso
NAME=MetalK8s
VERSION=2.2.0-dev
SHORT_VERSION=2.2
GIT=2.0.0-alpha1-377-gc692ee47
DEVELOPMENT_RELEASE=1
BUILD_TIMESTAMP=2019-04-03T08:22:16Z
BUILD_HOST=teddy-Latitude-E7450
```

The salt states should:
- mount all the iso listed in the BootstrapConfig file
- (re)configure salt master
- "deploy" all registry image in the local regisitry
- (re)configure nginx to serve the new packagges

gdemonet:
So would we consider the new workflows to look like:

1. Initial install
    + First mount your `metalk8s-X` ISO wherever
    + Specify where your `metalk8s-X` ISO is located in `BootstrapConfiguration`
    + Running the `bootstrap.sh` script first runs this state to (re-)mount said ISO before anything

2. Upgrade
    + Add the new `metalk8s-Y` ISO location in `BootstrapConfiguration`
    + Run some `salt bootstrap metalk8s.iso.all_mounted saltenv=metalk8s-X` to mount `metalk8s-Y` as well
    + Then run the upgrade scenario from `metalk8s-Y` saltenv

I'm not sure what you meant by "the salt states should...", but I don't think we want the ISO-related states to populate the Docker registry/Yum repositories ; it should remain in the `metalk8s.registry` and `metalk8s.repo` states.

TeddyAndrieux:
For 1. I agree
 
2. For me should be more an "include state" like `metalk8s.prepare` that btw do the `iso.all_mounted` and also include the registriy and repo configuration salt states.

> I'm not sure what you meant by "the salt states should...", but I don't think we want the ISO-related states to populate the Docker registry/Yum repositories ; it should remain in the `metalk8s.registry` and `metalk8s.repo` states.

It's not really the iso-related state yes it's more a "preparation state"

NicolasT:
[...]

PR #1149, opened 2019-05-24
Mount all iso listed in bootstrap config file

ramanduh:
WIP from Nicolas's stash: https://github.com/scality/metalk8s/commit/c2aa4c538a0f2b8f3912887873e05899375a85a2
cf context for description

Mount all iso listed in /etc/metalk8s/bootstrap.yaml and configure salt-master

**Component**:

deployment

**Context**: 

Add a products entry in /etc/metalk8s/bootstrap.yaml which will contain
the product path or the product iso. If product path is provided,
expose saltstack file_roots and pillar_roots
into salt master configuration. If a iso image is provided, mount the iso
first before performing the same operation.

As set of assertion is also made to verify the product version. In this
case, if the products:metalk8s field contains an iso, isoinfo binary is
required to extract the product version

Change in the bootstrap config:

```
apiVersion: metalk8s.scality.com/v1alpha2
kind: BootstrapConfiguration
networks:
  controlPlane: 172.21.254.0/28
  workloadPlane: 172.21.254.32/27
ca:
  minion: bootstrap
apiServer:
  host: 172.21.254.14
  keepalived:
    enabled: true
products:
  metalk8s:
  - /srv/scality/metalk8s-2.0.0-dev/
```

**Note**
The PR is not fully functional right now. In fact, the `iso_root_path` pillar is moved to ext_pillar and generated by the salt master. The issue is that is the iso path provided by the user is not accessible from the salt master container, the ext_pillar will fail

**Summary**:

**Acceptance criteria**: 

---

See: #924

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -132,6 +132,9 @@ apiServer:
   host: #{IPAddr.new(CONTROL_PLANE_IP).mask(CONTROL_PLANE_NETMASK).to_range.last(2).first.to_s}
   keepalived:
     enabled: true
+products:
+  metalk8s:
+  - /srv/scality/metalk8s-$SHORT_VERSION
 EOF
 
 echo "Launching bootstrap"
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -3,11 +3,13 @@
 Module for handling MetalK8s specific calls.
 '''
 import logging
+import os.path
+import re
 import socket
 import time
 
 from salt.exceptions import CommandExecutionError
-from salt.utils.decorators import depends
+import salt.utils.files
 
 log = logging.getLogger(__name__)
 
@@ -104,3 +106,47 @@ def minions_by_role(role, nodes=None):
         for (node, node_info) in nodes.items()
         if role in node_info.get('roles', [])
     ]
+
+
+def _get_product_version(info):
+    """Extract production version from info
+
+    Args:
+        info (str): content of metalk8s product.txt file
+    """
+    match = re.search(r'^SHORT_VERSION=(?P<version>.+)$', info, re.MULTILINE)
+    return match.group('version') if match else None
+
+
+def product_version_from_tree(path):
+    log.debug('Reading product version from %r', path)
+
+    product_txt = os.path.join(path, 'product.txt')
+
+    if not os.path.isfile(product_txt):
+        raise CommandExecutionError(
+            'Path {} has no "product.txt"'.format(path))
+
+    with salt.utils.files.fopen(os.path.join(path, 'product.txt')) as fd:
+        return _get_product_version(fd.read())
+
+
+def product_version_from_iso(path):
+    log.debug('Reading product version from %r', path)
+
+    cmd = ' '.join([
+        'isoinfo',
+        '-x', '/PRODUCT.TXT\;1',
+        '-i', '"{}"'.format(path),
+    ])
+    result = __salt__['cmd.run_all'](cmd=cmd)
+    log.debug('Result: %r', result)
+
+    if result['retcode'] != 0:
+        raise CommandExecutionError(
+            'Failed to run isoinfo: {}'.format(
+                result.get('stderr', result['stdout'])
+            )
+        )
+
+    return _get_product_version(result['stdout'])
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -109,16 +109,43 @@ def minions_by_role(role, nodes=None):
 
 
 def _get_product_version(info):
-    """Extract production version from info
+    """Extract product version from info
 
-    Args:
+    Arguments:
         info (str): content of metalk8s product.txt file
     """
     match = re.search(r'^SHORT_VERSION=(?P<version>.+)$', info, re.MULTILINE)
     return match.group('version') if match else None
 
 
-def product_version_from_tree(path):
+def _get_product_name(info):
+    """Extract product name from info
+
+    Arguments:
+        info (str): content of metalk8s product.txt file
+    """
+    match = re.search(r'^NAME=(?P<name>.+)$', info, re.MULTILINE)
+    return match.group('name') if match else None
+
+
+def _get_product_info(info):
+    """Extract product information from info
+
+    Arguments:
+        info (str): content of metalk8s product.txt file
+    """
+    return {
+        'version': _get_product_version(info),
+        'name': _get_product_name(info)
+    }
+
+
+def product_info_from_tree(path):
+    """Extract product information from a directory
+
+    Arguments:
+        path (str): path to a directory
+    """
     log.debug('Reading product version from %r', path)
 
     product_txt = os.path.join(path, 'product.txt')
@@ -128,10 +155,15 @@ def product_version_from_tree(path):
             'Path {} has no "product.txt"'.format(path))
 
     with salt.utils.files.fopen(os.path.join(path, 'product.txt')) as fd:
-        return _get_product_version(fd.read())
+        return _get_product_info(fd.read())
 
 
-def product_version_from_iso(path):
+def product_info_from_iso(path):
+    """Extract product information from an iso
+
+    Arguments:
+        path (str): path to an iso
+    """
     log.debug('Reading product version from %r', path)
 
     cmd = ' '.join([
@@ -149,4 +181,52 @@ def product_version_from_iso(path):
             )
         )
 
-    return _get_product_version(result['stdout'])
+    return _get_product_info(result['stdout'])
+
+
+def get_products(products=None):
+    """Get a matching between version and path from products or
+    `metalk8s.products` from pillar if products is None
+
+    Arguments:
+        products (list): list of path to directory or iso
+    """
+    if not products:
+        products = __pillar__.get('metalk8s', {}).get('products', [])
+
+    if isinstance(products, str):
+        products = [products]
+    if not isinstance(products, list):
+        raise CommandExecutionError(
+            'Invalid products list or string expected and got: {0}'.format(
+                products
+            )
+        )
+
+    res = {}
+
+    for prod in products:
+        if os.path.isdir(prod):
+            iso = None
+            info = product_info_from_tree(prod)
+            path = prod
+            version = info['version']
+        elif os.path.isfile(prod):
+            iso = prod
+            info = product_info_from_iso(prod)
+            version = info['version']
+            path = '/srv/scality/metalk8s-{0}'.format(version)
+        else:
+            log.warning(
+                'Skip, invalid products path %s, should be an iso or a '
+                'directory.',
+                prod
+            )
+            continue
+
+        info.update({'path': path, 'iso': iso})
+        res.update(
+            {'metalk8s-{0}'.format(version): info}
+        )
+
+    return res
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -20,6 +20,7 @@ def _load_config(path):
 
     assert config.get('apiVersion') == 'metalk8s.scality.com/v1alpha2'
     assert config.get('kind') == 'BootstrapConfiguration'
+    assert 'products' in config
 
     return config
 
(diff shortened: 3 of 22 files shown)


salt/_modules/metalk8s.py
```diff
@@ -104,3 +110,43 @@ def minions_by_role(role, nodes=None):
         for (node, node_info) in nodes.items()
         if role in node_info.get('roles', [])
     ]
+
+
+def _product_version_from_fd(fd):
```
gdemonet: I'd rather have this helper work with strings, much easier to test in isolation. Also, you can use `re.MULTILINE` to avoid testing all the lines by hand:

```python
def _get_product_version(info):
    match = re.search(r'^VERSION=(?P<version>.+)$', info, re.MULTILINE)
    return match.group('version') if match else None
```

salt/_modules/metalk8s.py
```diff
@@ -104,3 +110,43 @@ def minions_by_role(role, nodes=None):
...
+    return None
+
+
+def product_version_from_tree(path):
+    log.debug('Reading product version from %r', path)
+
+    product_txt = os.path.join(path, 'product.txt')
+
+    if not os.path.isfile(product_txt):
+        log.warning('Path %r has no "product.txt"', path)
```
gdemonet: Isn't this an error rather than a warning? I guess raising `CommandExecutionError` from here would make sense.

salt/_modules/metalk8s.py
```diff
@@ -104,3 +110,43 @@ def minions_by_role(role, nodes=None):
...
+
+    cmd = ' '.join([
+        'isoinfo',
+        '-x', '/PRODUCT.TXT\;1',
+        '-i', '"{}"'.format(path),
+    ])
+    result = __salt__['cmd.run_all'](cmd=cmd)
+    log.debug('Result: %r', result)
+
+    if result['retcode'] != 0:
```
gdemonet, line 177: That would be an error as well IIUC

gdemonet (commented):
Few light comments on the loading of `product.txt`

gdemonet:
Some stronger concerns:

- `products` seems rather vague, and since the current `BootstrapConfiguration` file is only for MetalK8s, I find it too generic (we wouldn't store Solution ISOs in this file, IIUC): we need to discuss this, as it's part of the user experience when installing MetalK8s. In any case, the `metalk8s.products` formula only works for MetalK8s ISOs (there are some `Assert product '{{ product }}' is MetalK8s` states).
- We allow either mounted or unmounted ISOs in this file, but it goes against the idea described in #924: we want to control where ISOs are mounted, so we can be sure we'll mount what is needed in the `salt-master`  and `repositories` containers.
- We now use the _full_ version: IMHO, it's out of this PR's scope, but if we want to do so, it will change a lot of documentation/UI w.r.t. the `saltenv` variable (we would need something like `saltenv=metalk8s-2.0.0-alpha1` instead of `saltenv=metalk8s-2.0`)
- We missed the `repositories` configuration (for the images), we only did `[file|pillar]_roots` for Salt master. This has to be part of this PR IMO, since the `metalk8s.products` formula is now part of the bootstrap script and upgrade procedure, as explained in #924
- [This comment](https://github.com/scality/metalk8s/issues/924#issuecomment-492202747) from @TeddyAndrieux also requests changes to the Salt master manifest to mount the accurate `[file|pillar]_roots`

salt/metalk8s/products/init.sls
```diff
@@ -0,0 +1,81 @@
...
+
+Assert product '{{ product }}' is MetalK8s:
+  cmd.run:
+  - name: grep '^NAME=MetalK8s$' {{ product_path }}/product.txt > /dev/null
+  - stateful: true
+  - require:
+    - file: Assert '{{ product_path }}/product.txt' exists
+
+{%   for (root, subdir) in [('file_roots', 'salt'), ('pillar_roots', 'pillar')] %}
+Expose Salt '{{ root }}' for {{ product_path }}:
```
TeddyAndrieux: We should directly call the salt-master state instead of doing it here

Vagrantfile
```diff
@@ -132,10 +132,13 @@ apiServer:
   host: #{IPAddr.new(CONTROL_PLANE_IP).mask(CONTROL_PLANE_NETMASK).to_range.last(2).first.to_s}
   keepalived:
     enabled: true
+products:
+  metalk8s:
+  - /srv/scality/metalk8s-$VERSION
```
TeddyAndrieux: No that's not what we want it's the oposite.

The path where to extract the iso should be hardcoded
We should have something like
```
products:
  metalk8s:
    - /root/metalk8s.iso
    - /var/tmp/metalk8s.iso
    - /toto/tata/titi.iso
```
And then the goal is to mount these iso in function of the version so for each `/srv/scality/metalk8s-<version>`

TeddyAndrieux (changes requested):

salt/_modules/metalk8s.py
ramanduh: will fix

salt/_modules/metalk8s.py
ramanduh: will fix

salt/_modules/metalk8s.py
ramanduh: will fix

Vagrantfile
ramanduh: Yes, it's already the case, the path where to mount the iso is hardcoded in `/srv/scality`, it is just that both are supported: product path and iso (if the iso is given, mount it otherwise, just config salt and perform a set of assert)

salt/metalk8s/products/init.sls
ramanduh: Ok, will do

ramanduh:
@gdemonet thanks for reviews:
* Point 1: Ok for discussion about it. Is it only for the naming? (in that case, I will ping you directly) or do you want to discuss about the fact that we may want to manage solution iso as well (in that case, I will create an other question issue)
* Point 2: the mounted iso is allowed as the state will be bundled in the iso itself, so my understanding is that this state will be called necessary when at least one iso is mounted
* Point 3: I will see with @NicolasT using version instead short version is necessary or not before planing the doc/ui update
* Point 4 and 5: Ok, working on it

Vagrantfile
```diff
@@ -80,8 +80,8 @@ die() {
 echo "Installing build artifacts on the host"
 
 mkdir -p /srv/scality || die "Failed to create /srv/scality"
-rm -f "/srv/scality/metalk8s-$SHORT_VERSION" || die "Failed to unlink symlink destination"
```
slaperche-scality: The move from `SHORT_VERSION` to `VERSION` clearly deserve a commit (and commit message) on its own.
(It's not even mentioned in the current one…)

gdemonet:
> @gdemonet thanks for reviews:
> 
> * Point 1: Ok for discussion about it. Is it only for the naming? (in that case, I will ping you directly) or do you want to discuss about the fact that we may want to manage solution iso as well (in that case, I will create an other question issue)

No, the point is I don't think we want to manage **all** ISOs in this file, thus renaming/moving `products:metalk8s` to `archives`. Let's discuss it today.

> * Point 2: the mounted iso is allowed as the state will be bundled in the iso itself, so my understanding is that this state will be called necessary when at least one iso is mounted

See the original ticket (and comments): we don't wan't to let the user choose where ISOs will be mounted, even _the first one_. This could allow a user to mount its newest ISO at a random path, e.g. `/tmp/metalk8s-latest`, without any trouble w.r.t Pillar values (in particular, `metalk8s:iso_root_path`, which is hardcoded to `/srv/scality/metalk8s-@@VERSION`).

> * Point 3: I will see with @NicolasT using version instead short version is necessary or not before planing the doc/ui update

I don't see why it would be necessary. It's useful (because we'll be able to test upgrades between patch versions), but not mandatory in this specific PR.

Vagrantfile
ramanduh: @gdemonet @slaperche-scality I'm gonna remove this change from this PR, you're right, it deserves a new issue/ticket

salt/metalk8s/products/init.sls
```diff
@@ -0,0 +1,81 @@
...
+{%         set product_path = base ~ '/metalk8s-' ~ product_version %}
+{%         set product_path_available = "- mount: Product " ~ product ~ " available at " ~ product_path %}
+Product path {{ product_path }} exists:
+  file.directory:
+  - name: {{ product_path }}
+  - mode: 0755
+  - makedirs: true
+
+Product {{ product }} available at {{ product_path }}:
+  mount.mounted:
```
ycointe: It would be nice to add `persist: True` to have the iso mounted automatically and to be resilient when a restart occur.

gdemonet: Note that it's the default, but we want to be explicit

ramanduh: fixed

salt/_pillar/metalk8s.py
```diff
@@ -78,12 +79,34 @@ def _load_apiserver(config_data):
...
 
+def _load_iso_path(config):
+    """Load iso path from BootstrapConfiguration
+
+    """
+    res = {}
+
+    for product in config['products']['metalk8s']:
+        if __salt__['file.directory_exists'](product):
+            version = __salt__['metalk8s.product_version_from_tree'](product)
```
TeddyAndrieux: Do we really want to support this ? For me we don't.
Any news about this @gdemonet or can I remove it ?

salt/metalk8s/products/mounted.sls
```diff
@@ -0,0 +1,65 @@
+{% import_yaml "/etc/metalk8s/bootstrap.yaml" as bootstrap_config %}
```
TeddyAndrieux: Not really like this approach, strongly prefer using ext_pillar informations, anyone against that ?

salt/metalk8s/products/mounted.sls
```diff
@@ -0,0 +1,65 @@
...
+{% set base = '/srv/scality' %}
+{% set roots_file = '/etc/salt/master.d/99-metalk8s-roots.conf' %}
+
+{% for product in products %}
+{%   if salt['file.file_exists'](product) %}
+{%     set product_version = salt['metalk8s.product_version_from_iso'](product) %}
+{%     if not product_version %}
+{%         set product_path = product %}
+{%         set product_version = 'unknown' %}
+{%         set product_path_available = "- test: Product " ~ product ~ " available" %}
```
TeddyAndrieux: All these informations should be retrieved from ext_pillar

salt/_pillar/metalk8s.py
```diff
@@ -78,12 +79,34 @@ def _load_apiserver(config_data):
...
+        else:
+            version = __salt__['metalk8s.product_version_from_iso'](product)
```
TeddyAndrieux: Ok we can't do that in ext_pillar as we need the iso to be accessible from the master if we want to do this

salt/metalk8s/products/mounted.sls
TeddyAndrieux: Fixed

salt/_pillar/metalk8s.py
TeddyAndrieux: Yes we need it, done

salt/_pillar/metalk8s.py
TeddyAndrieux: Only retrieve list of iso/directory in ext_pillar, use them in a custom module

TeddyAndrieux:
@bert-e reset

TeddyAndrieux:
@bert-e reset

TeddyAndrieux:
@bert-e reset

salt/_modules/metalk8s.py
```diff
@@ -104,3 +106,127 @@ def minions_by_role(role, nodes=None):
         for (node, node_info) in nodes.items()
         if role in node_info.get('roles', [])
     ]
+
+
+def _get_product_version(info):
+    """Extract production version from info
```
sayf-eddine-scality: product~ion~

TeddyAndrieux: Dudu's fault :p

TeddyAndrieux:
@bert-e reset

sayf-eddine-scality (approved):
LGTM
minor note: If we have two isos with the same mount folder the first is ignored without warning maybe we want to add one @TeddyAndrieux  ?

TeddyAndrieux:
> LGTM
> minor note: If we have two isos with the same mount folder the first is ignored without warning maybe we want to add one @TeddyAndrieux ?

Lets merge this one as it's here since long time and do it in another PR

TeddyAndrieux (approved):

tcarmet:
@bert-e bypass_author_approval

PR #1153, opened 2019-05-24
build: fix skopeo usage on MacOS

### LINKED ISSUE #1152: buildchain: fix skopeo usage on MacOS
**Component**:

buildchain

**What happened**:

Cannot get remote images on MacOS with `skopeo`.

**What was expected**:

The build works and an ISO is produced.

**Steps to reproduce**:

Run `./doit.sh` on MacOS and you'll have the following error:

```
time="2019-05-24T10:24:10+02:00" level=fatal msg="Error loading trust policy: open /etc/containers/policy.json: no such file or directory" 
TaskFailed - taskid:_deploy_salt_tree:salt:pause
Command failed: '['skopeo', 'copy', '--format', 'v2s2', 'docker://k8s.gcr.io/pause@sha256:f78411e19d84a252e53bff71a4407a5686c46983a2c2eeed83929b888179acea', 'docker-archive:/Users/charlesnguyen/metalk8s/_build/root/salt/metalk8s/container-engine/containerd/files/pause-3.1.tar']' returned 1
```

**Resolution proposal**:

This can be fixed by passing the `--insecure-policy` flags to the `skopeo` command (which should be fine, as the default policy seems to allow insecure anyway).
But that's not enough, because it then fails with the following error:

```
time="2019-05-24T11:13:03+02:00" level=fatal msg="Error choosing an image from manifest list docker://calico/kube-controllers@sha256:d5533cb8df6150123cad26e369383a3e665c5e376f8c9dd7d80a8c86fa907e7c: no image found in manifest list for architecture amd64, OS darwin"
TaskFailed - taskid:_image_pull:calico-kube-controllers
Command failed: '['skopeo', '--insecure-policy', 'copy', '--format', 'v2s2', '--dest-compress', 'docker://calico/kube-controllers@sha256:d5533cb8df6150123cad26e369383a3e665c5e376f8c9dd7d80a8c86fa907e7c', 'dir:/Users/charlesnguyen/metalk8s/_build/root/images/calico-kube-controllers/3.7.2']' returned 1
```

Since it tries to fetch an image for MacOS, we also have to pass `--override-os linux`.

slaperche-scality:
**Component**:

buildchain

**Context**: 

Since the implementation of #1049, the buildchain uses `skopeo` and this breaks the build on MacOS.

**Summary**:

This fix two issues related to skopeo on MacOS:
- first, the policy file is, for some reason (missing? need root?), not readable => this is fixed by passing the `--insecure-policy` flags (which should be OK: IIUC the default conf allows insecure)
- then, being on MacOS `skopeo` tried to fetch images for MacOS (which may not exists), since we target Linux we specify `--override-os linux` to enforce the download of images for Linux.

**Acceptance criteria**: 

The build should work on MacOS.

Closes: #1152

### DIFF
--- a/buildchain/buildchain/targets/remote_image.py
+++ b/buildchain/buildchain/targets/remote_image.py
@@ -103,7 +103,10 @@ class RemoteImage(image.ContainerImage):
 
     def _skopeo_copy(self) -> List[str]:
         """Return the command line to execute skopeo copy."""
-        cmd = [config.SKOPEO, 'copy', '--format', 'v2s2']
+        cmd = [
+            config.SKOPEO, '--override-os', 'linux', '--insecure-policy',
+            'copy', '--format', 'v2s2'
+        ]
         if not self._use_tar:
             cmd.append('--dest-compress')
         cmd.append('docker://{}'.format(self.fullname))


buildchain/buildchain/targets/remote_image.py
```diff
@@ -103,7 +103,10 @@ def task(self) -> types.TaskDict:
 
     def _skopeo_copy(self) -> List[str]:
         """Return the command line to execute skopeo copy."""
-        cmd = [config.SKOPEO, 'copy', '--format', 'v2s2']
+        cmd = [
+            config.SKOPEO, '--override-os', 'linux', '--insecure-policy',
```
jbertran, line 107: Could make that little guy configurable, since in the future we'll want to verify image signatures in CI at least.

jbertran (approved):

buildchain/buildchain/targets/remote_image.py
slaperche-scality: Then, it will be made configurable at that time :slightly_smiling_face: 

There is no point in making something configurable for "one day"™
I mean, we don't even have a ticket to track it this "verify image in CI", so I'm not expecting it to land any time soon (if ever).

Otherwise, I can also make `override-os` configurable ("one day" we may support MacOS or Windows) :stuck_out_tongue:

slaperche-scality:
@bert-e approve

Issue #1018, opened 2019-04-18
External pillar should "log" error in `_errors` key of the resulting dict


TeddyAndrieux:
Currently in external pillars we log in error when we got some error but for the first install it's expected error as for example apiserver currently not configured.

We should put all error message in `_errors` key of the resulting external pillar for example for endpoint if we got some error we are able to get the error message just by getting the `'_errors` key from the pillar
`salt-call pillar.get metalk8s.endpoints._errors`

PR #1155, opened 2019-05-24
Add prometheus url to ui config

### LINKED ISSUE #1131: Cluster Health - Add url_prometheus into metalk8s-ui ConfigMap
**Component**: Salt, Prometheus

**Why this is needed**:

Metalk8s GUI will need the URL of prometheus in order to get information like alerts or status from it.
As a front-end developer I want to be able to access to the prometheus via a URL

**What should be done**:

We should add the prometheus link to metalk8s-ui config map.

In `metalk8s/salt/metalk8s/addons/ui/deployed.sls`, we should add in the config.json a item `"url_prometheus": "{{ prometheus_api }}"`.
`prometheus_api` information should be provide by the pillar.

**Test plan**:
Fetch the <API_SERVER>/config.json
It should return something like this :
```json
{
    "url": "https://172.21.254.13:6443", 
    "url_salt": "http://172.21.254.13:4507",
    "url_prometheus": "http://172.21.254.13:9090"
}
```

jbertran:
**Component**:

salt, ui, monitoring

**Context**: We want to supply the UI ConfigMap with the Prometheus URL, in order for it to use Prometheus for monitoring purposes.

**Summary**:

* Make endpoint lookups namespace-sensitive in the ext_pillar
* Add the prometheus_url key to the config map
* Add a NodePort service to expose Prometheus on the node

**Acceptance criteria**: After a successful deployment, we have the `prometheus_url` key in `<UI_URL>/config.json`, and the value of this key is an accessible URL to Prometheus on the node where it runs.

---

Closes: #1131


salt/_pillar/metalk8s_endpoints.py
```diff
@@ -16,11 +16,11 @@ def __virtual__():
         return __virtualname__
 
 
-def service_endpoints(service, kubeconfig):
+def namespaced_service_endpoints(service, namespace, kubeconfig):
```
slaperche-scality: minor: You could have kept the old name

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -16,6 +16,48 @@ def __virtual__():
         return __virtualname__
 
 
+def namespaced_service_endpoints_nodeport(service, namespace, kubeconfig):
```
slaperche-scality: ditto, I'm not sure that it's useful to add the `namespaced` prefix: eveything seems to be namespaced in k8s anyway.

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -16,6 +16,48 @@ def __virtual__():
...
+    endpoint = namespaced_service_endpoints(service, namespace, kubeconfig)
+    node_name = endpoint.get('node_name')
+
+    if not node_name:
+        log.exception('Cannot get `node_name` from %s.', endpoint)
```
slaperche-scality: why `log.exception`, there is no exception raised here?

salt/metalk8s/addons/ui/deployed.sls
```diff
@@ -3,7 +3,7 @@
 
 {%- set apiserver = 'https://' ~ pillar.metalk8s.api_server.host ~ ':6443' %}
 {%- set saltapi = 'http://' ~ pillar.metalk8s.endpoints['salt-master'].ip ~ ':' ~ pillar.metalk8s.endpoints['salt-master'].ports.api %}
-{%- set prometheus = 'http://' ~ pillar.metalk8s.endpoints['prometheus'].ip ~ ':' ~ pillar.metalk8s.endpoints['prometheus'].ports.api %}
+{%- set prometheus = 'http://' ~ pillar.metalk8s.endpoints['prometheus'].ip ~ ':' ~ '30222' %}
```
slaperche-scality: Why this change?

salt/metalk8s/addons/monitoring/prometheus/exposed.sls
```diff
@@ -0,0 +1,28 @@
...
+          protocol: TCP
+          node_port: 30222
+          targetPort: api
+        selector:
+          app: prometheus
+          prometheus: k8s
+        type: NodePort
+  require:
+    - pkg: Install Python Kubernetes client
+    - salt: metalk8s.addons.monitoring.prometheus.deployed
```
TeddyAndrieux: These requires are useless here as it's a yaml key in `Expose Prometheus` and btw you should not require something not in the current sls (or in an included one)

buildchain/buildchain/salt_tree.py
```diff
@@ -150,6 +150,7 @@ def _run(self) -> None:
     Path('salt/metalk8s/addons/monitoring/kube-state-metrics/deployed.sls'),
     Path('salt/metalk8s/addons/monitoring/node-exporter/deployed.sls'),
     Path('salt/metalk8s/addons/monitoring/prometheus/deployed.sls'),
+    Path('salt/metalk8s/addons/monitoring/prometheus/exposed.sls'),
```
TeddyAndrieux: Why separate `deployed` and `exposed` ? Currently the `exposed` state "deploy" a config map :stuck_out_tongue: 
We said few week ago that all state that should only be launched on the master should be called `deployed.sls`

salt/metalk8s/deployed.sls
```diff
@@ -4,10 +4,10 @@ include:
   - metalk8s.kubernetes.coredns.deployed
   - metalk8s.repo.deployed
   - metalk8s.salt.master.deployed
-  - metalk8s.addons.ui.deployed
```
TeddyAndrieux: For me you should not remove this from here just put it in the right order here and if needed add a logic in the `prometheus` state to check the status of the ConfigMap before finishing the state.

Maybe just add a state in `prometheus.deployed` that retrieve the configmap from k8s api with a retry just after the configmap deployement

TeddyAndrieux (dismissed):

buildchain/buildchain/salt_tree.py
gdemonet: Agreed, though `deployed` is actually the upstream YAML that gets rendered through Salt into proper states... Maybe the upstream one should get renamed and `include`d in `deployed.sls`

salt/_pillar/metalk8s_endpoints.py
gdemonet: Yes, `log.error` would be better. Even then, we should start implementing the `_errors` key as described in #1018

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -16,11 +16,53 @@ def __virtual__():
         return __virtualname__
 
 
-def service_endpoints(service, kubeconfig):
+def namespaced_service_endpoints_nodeport(service, namespace, kubeconfig):
```
gdemonet: Hum, no let's not do this _workaround_ thing in the ext_pillar. The service IP is still the cluster IP, so we shouldn't get out of our way just to _make things work_. Instead, you should use `grains['workload_plane_ip']` in `addons.ui.deployed`, until we can reach the real service address through some LB/Ingress.

buildchain/buildchain/salt_tree.py
jbertran: The idea here comes from that this state was to a large extent autogenerated from [coreos upstream](https://github.com/coreos/kube-prometheus), so I want to separate them from what's really ours.

More precisely:

1. We're not the authors per se of the contents of `deployed`, whereas we are in `exposed` or however we choose to call the ingress deployment.
2. Being "pulled" from upstream, the `deployed` files should not be modified after their initial version adjustment, whereas the `exposed` files could be, so they're not similar lifecycle-wise.

If we want to keep the naming convention on master-only states though, we can definitely have a `metalk8s.addons.monitoring.prometheus-ingress` or similar state instead - I wasn't aware of it!

salt/metalk8s/deployed.sls
TeddyAndrieux: @jbertran ok you just have to add endpoint like `salt-master` and `repositories` in `pillar_data` in `salt/metalk8s/orchestrate/bootstrap/init.sls`

buildchain/buildchain/salt_tree.py
jbertran: From discussion with @gdemonet, another sane option would be to store the "upstream" states in `upstream.sls` and just include these states in `deployed.sls`, where we could have our custom states for exposing services to the outside.

salt/_pillar/metalk8s_endpoints.py
jbertran: From f2f discussions. This is indeed ugly but we're keeping it for the short duration between here and LB/Ingress addition.

MonPote:
@bert-e reset

jbertran:
Done little bot

MonPote (dismissed):
LGTM

jbertran:
Fixed, bert-e boy

jbertran:
@bert-e approve

jbertran:
@bert-e reset

sayf-eddine-scality (approved):

jbertran:
2.1 done bert-e, gimme 2.2 now

jbertran:
Gottem berte-e boy. Let's merge!

PR #1161, opened 2019-05-28
Job Events Parser

### LINKED ISSUE #1145: Parse Salt job results in UI in case of failure
**Component**: ui

**Why this is needed**:

When a runner (`deploy_node` in particular) fails, the orchestrate sends a large JSON payload with all states return information (whether they failed or not). A non-technical user would find it difficult to understand.

**What should be done**:

The UI should provide a simple parsing logic to extract the "first" failing state in case of orchestrate failure.

Given a JSON of state results in the `return.data.<master_name>` key, order its elements by `__run_num__`, and pick the first one with `result: False`.

Display to the user: "Orchestration failed at step `${step_id}`: `${step.comment}`".

Each element of the aforementioned JSON has a key in the format:

`${module}_|-${step_id}_|-${step_name}_|-${function}`

**Test plan**:

TODO

carlito-scality:
**Component**: UI

**Context**: When a runner (deploy_node in particular) fails, the orchestrate sends a large JSON payload with all states return information (whether they failed or not). A non-technical user would find it difficult to understand.

**Summary**:
The UI should provide a simple parsing logic to extract the "first" failing state in case of orchestrate failure to display in the node list and the node deploy page

**Acceptance criteria**: 
Have a clear deployment status in the node list and the node deploy page when the deployment  failed.

image

Screenshot 2019-05-29 at 11 14 10

Workaround: As the nodes list is refreshed every 10s, so the error message is only available during max 10s since there is no way to persist it... => We display notifications instead

image

Closes: https://github.com/scality/metalk8s/issues/1145

### DIFF
--- a/ui/src/containers/NodeDeployment.js
+++ b/ui/src/containers/NodeDeployment.js
@@ -4,15 +4,17 @@ import { injectIntl } from 'react-intl';
 import { createSelector } from 'reselect';
 import styled from 'styled-components';
 import ReactJson from 'react-json-view';
-import { Button } from 'core-ui';
+import { Button, Loader } from 'core-ui';
 import { withRouter } from 'react-router-dom';
 
 import { subscribeDeployEventsAction } from '../ducks/app/nodes';
+import { getJobStatusFromEventRet } from '../../src/services/salt/utils';
 import {
   fontWeight,
   grayLightest,
   padding,
-  fontSize
+  fontSize,
+  brand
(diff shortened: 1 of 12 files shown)


carlito-scality:
@bert-e reset

gdemonet (approved):
Code looks good, GUI as well, and everything works!

carlito-scality:
@bert-e approve

PR #1177, opened 2019-06-03
Remove VERSION_SUFFIX  from Eve "product_version"

### LINKED ISSUE #1166: eve/get_product_version.sh should only output numbers separated by a dot
**Component**:

'build'

**What happened**:

artifacts are not processed by expiration and promotion, due to a version considered as invalid in the artifacts_name property

**What was expected**:

correct processing of artifacts expiration (ie good format of get_product_version.sh output)

**Steps to reproduce**

see builds listing, entries with a version suffix have not been expired.

fix all dev branches having the issue by removing the VERSION_SUFFIX of the output. If you want to use the current format of the suffixed version, please do not use get_product_version for that, as it is designed and used by the CI with its own requirements.

gdemonet:
**Component**: build

---

Closes: #1166

### DIFF
--- a/eve/get_product_version.sh
+++ b/eve/get_product_version.sh
@@ -4,5 +4,4 @@ script_full_path=$(readlink -f "$0")
 file_dir=$(dirname "$script_full_path")/..
 # shellcheck disable=SC1090
 source "$file_dir/VERSION" && \
-    echo "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_SUFFIX}"
-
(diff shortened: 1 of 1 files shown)


slaperche-scality (changes requested):
Should explain WHY this change in the commit message.

gdemonet:
> Should explain WHY this change in the commit message.

Done

gdemonet:
@bert-e reset

slaperche-scality (approved):

gdemonet:
@bert-e approve

PR #1179, opened 2019-06-03
UI: improve Dockerfile for faster build

### LINKED ISSUE #1139: Optimize UI container build
**Component**: ui

**Why this is needed**: The UI container is extremely slow to build (~15 minutes). This should be remedied now rather than later.

**What should be done**: Make the UI container build fast again (MTUICBFA?).

**Implementation proposal** (strongly recommended): Most likely some optimization to be done around docker file.

**Test plan**:

carlito-scality:
**Component**: UI

**Context**: 
The UI is extremely slow to build (~15 minutes). This should be remedied now rather than later.

**Summary**:
Some docker layers could be cached.

**Acceptance criteria**: 
The UI build is faster after the first build since some docker layers are cached.
Limitations: when rebuilding UI, we have to restart the pods manually to pull the new UI image (correctly generated) since the UI version does not change every times. Otherwise, the new image will be not correctly deployed.
``` kubectl delete pods metalk8s-ui-xxx-xxx -n kube-system ```

Closes: https://github.com/scality/metalk8s/issues/1139

### DIFF
--- a/buildchain/buildchain/image.py
+++ b/buildchain/buildchain/image.py
@@ -87,7 +87,7 @@ def task__image_dedup_layers() -> types.TaskDict:
 
 
 NGINX_IMAGE_VERSION : str = '1.15.8'
-
+NODE_IMAGE_VERSION : str = '10.16.0'
 
 # List of container images to pull.
 #
@@ -301,9 +301,20 @@ TO_BUILD : Tuple[targets.LocalImage, ...] = (
         dockerfile=constants.ROOT/'ui'/'Dockerfile',
         destination=constants.ISO_IMAGE_ROOT,
         save_on_disk=True,
-        build_args={'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION},
+        build_args={
+            'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION,
+            'NODE_IMAGE_VERSION': NODE_IMAGE_VERSION
+        },
         task_dep=['_image_mkdir_root'],
-        file_dep=list(coreutils.ls_files_rec(constants.ROOT/'ui')),
+        file_dep=(
+            list(coreutils.ls_files_rec(constants.ROOT/'ui'/'public')) +
+            list(coreutils.ls_files_rec(constants.ROOT/'ui'/'src')) +
+            [
+                constants.ROOT/'ui'/'package.json',
+                constants.ROOT/'ui'/'package-lock.json',
+                constants.ROOT/'ui'/'conf'/'nginx.conf'
+            ]
+        )
     ),
 )
 
--- a/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
+++ b/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
@@ -27,6 +27,7 @@ spec:
       containers:
         - name: metalk8s-ui
           image: {{ build_image_name('metalk8s-ui', '0.2') }}
+          imagePullPolicy: Always
           resources:
             limits:
               memory: 170Mi
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -3,8 +3,11 @@ ARG NGINX_IMAGE_VERSION=1.15.8
 FROM node AS build-step
 
 WORKDIR /home/node
-COPY . /home/node/
+COPY package*.json /home/node/
 RUN npm install
+
+COPY public /home/node/public
+COPY src /home/node/src
 RUN npm run build
 
 FROM nginx:${NGINX_IMAGE_VERSION}
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -1,6 +1,7 @@
 ARG NGINX_IMAGE_VERSION=1.15.8
+ARG NODE_IMAGE_VERSION=10.16.0
 
-FROM node AS build-step
+FROM node:${NODE_IMAGE_VERSION} AS build-step
 
 WORKDIR /home/node
 COPY package*.json /home/node/
--- a/ui/src/containers/Layout.js
+++ b/ui/src/containers/Layout.js
@@ -99,7 +99,6 @@ class Layout extends Component {
             <PrivateRoute exact path="/about" component={Welcome} />
             <PrivateRoute exact path="/" component={NodeList} />
           </Switch>
-          )}
         </CoreUILayout>
       </ThemeProvider>
     );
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Review by gdemonet — DISMISSED** [review]
LGTM, great speed improvements!

File: ui/Dockerfile [hunk 1]
```diff
@@ -3,8 +3,11 @@ ARG NGINX_IMAGE_VERSION=1.15.8
 FROM node AS build-step
 
 WORKDIR /home/node
-COPY . /home/node/
+COPY package*.json /home/node/
```
slaperche-scality, line 7: You should also update the `file_dep` for the `targets.LocalImage` for the UI image in the buildchain, no?

ui/Dockerfile
```diff
@@ -1,6 +1,7 @@
 ARG NGINX_IMAGE_VERSION=1.15.8
+ARG NODE_IMAGE_VERSION=10.16.0
 
-FROM node AS build-step
+FROM node:${NODE_IMAGE_VERSION} AS build-step
```
slaperche-scality, line 4: The value for this variable comes from where?

carlito-scality: ARG NODE_IMAGE_VERSION=10.16.0

carlito-scality: I will put it in targets.LocalImage of image.py as well

ui/Dockerfile
carlito-scality: I think we do not have to update since package*.json is part of '/ui'.

slaperche-scality: Yeah, but currently the buildchain will re-trigger the build if any file under `/ui` is modified, whereas with you change we only care about the `package*.json`.

If you don't update then you'll trigger useless builds (that will be fast thanks to better caching with this change), so it's better to update to avoid it :)

BTW, shouldn't your check complains about this mismatch @jbertran ?

buildchain/buildchain/image.py
```diff
@@ -304,11 +304,11 @@ def show() -> str:
         build_args={'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION, 'NODE_IMAGE_VERSION': NODE_IMAGE_VERSION},
         task_dep=['_image_mkdir_root'],
         file_dep=(
```
slaperche-scality, line 309: This commit should be squashed with the previous one.

buildchain/buildchain/image.py
```diff
@@ -301,7 +301,7 @@ def show() -> str:
         dockerfile=constants.ROOT/'ui'/'Dockerfile',
         destination=constants.ISO_IMAGE_ROOT,
         save_on_disk=True,
-        build_args={'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION},
+        build_args={'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION, 'NODE_IMAGE_VERSION': NODE_IMAGE_VERSION},
```
slaperche-scality: line too long, please wrap it as:

```
        build_args={
            'NGINX_IMAGE_VERSION': NGINX_IMAGE_VERSION,
            'NODE_IMAGE_VERSION': NODE_IMAGE_VERSION
        },
```

salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
```diff
@@ -27,6 +27,7 @@ spec:
       containers:
         - name: metalk8s-ui
           image: {{ build_image_name('metalk8s-ui', '0.2') }}
+          imagePullPolicy: Always
```
slaperche-scality, line 30: unrelated change, shouldn't be in this commit

ui/src/containers/Layout.js
```diff
@@ -99,7 +99,6 @@ class Layout extends Component {
             <PrivateRoute exact path="/about" component={Welcome} />
             <PrivateRoute exact path="/" component={NodeList} />
           </Switch>
-          )}
```
slaperche-scality, line 102: unrelated change, shouldn't be in this commit

slaperche-scality (changes requested):

salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
carlito-scality: I rename the commit in the squash

ui/src/containers/Layout.js
carlito-scality: I rename the commit in the squash

buildchain/buildchain/image.py
carlito-scality: done

carlito-scality:
@bert-e reset

carlito-scality:
@bert-e reset

buildchain/buildchain/image.py
```diff
@@ -87,7 +87,7 @@ def show() -> str:
 
 
 NGINX_IMAGE_VERSION : str = '1.15.8'
-
+NODE_IMAGE_VERSION : str = '10.16.0'
```
slaperche-scality, line 90:
```
-update file_dep
-change the pull policy deployment.yaml
-remove useless html element
```

Those 3 changes are totally unrelated (unless I miss something) and should have dedicated commits.

A commit is "atomic" and has one goal: what if I want to revert/cherry-pick the removal of the useless HTML element without losing/picking the `file_dep` update?

slaperche-scality (changes requested):

buildchain/buildchain/image.py
carlito-scality: Two first changes are part of the build configuration. Ok the last one is a missing error but I will not create another for 2 characters removal.

slaperche-scality: the deployment.yaml is not related to the buildchain, but to the deployment so they don't belong to the same commit.

Having a dedicated commit for this one would allow you to explains why you added ` imagePullPolicy: Always` in the commit message (useful when someone will use `git log` or `git blame`).

And you should create a dedicated commit, even for your two characters removal since it has absolutely no relation with the rest (a little commit "fix typo" cost nothing and keep a usable and useful git history, which save a lot more time in the future than splitting the commit).

slaperche-scality (changes requested):

buildchain/buildchain/image.py
carlito-scality: Ok for the last one. but for the two first, I do not agree. It depends on how we define the "atomicity" of a commit.

carlito-scality:
@bert-e reset

slaperche-scality (approved):

carlito-scality:
@bert-e approve

Issue #1170, opened 2019-05-31
Grafana pod crashes on startup


jbertran:
<!-- Please use this template while reporting a bug and provide as much info as possible. Not doing so may result in your bug not being addressed in a timely manner. Thanks!

If the matter is security related, please disclose it privately to security@scality.com
-->

**Component**: monitoring

<!-- E.g. 'salt', 'containers', 'kubernetes', 'build', 'tests'... -->

**What happened**: Grafana manifest is present in API but the pod crashes on startup. From the logs, the culprit is the b64'd Prometheus configuration mounted as a `Secret`.

**What was expected**: Grafana pod starts normally.

**Steps to reproduce**: Deploy the product.

**Resolution proposal** (optional):

The following diff fixes this issue by mounting this as a `ConfigMap` instead.

```diff
From e85680b640f5e46e2a8bb99587ac1580d642f158 Mon Sep 17 00:00:00 2001                                                                                                                                                                                                  [65/150]
From: Jordi Bertran de Balanda <jordi.bertran@scality.com>                                                                                                                                                                                                                      
Date: Fri, 31 May 2019 11:28:38 +0200                                                                                                                                                                                                                                           
Subject: [PATCH] monitoring: fix Grafana manifest                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
The way the Prometheus configuration was exposed made it crash                                                                                                                                                                                                                  
on startup.

Issue: #<>
---
 .../addons/monitoring/grafana/deployed.sls    | 25 ++++++++++++++-----
 1 file changed, 19 insertions(+), 6 deletions(-)

diff --git a/salt/metalk8s/addons/monitoring/grafana/deployed.sls b/salt/metalk8s/addons/monitoring/grafana/deployed.sls
index c46c0108..1e1257b8 100644
--- a/salt/metalk8s/addons/monitoring/grafana/deployed.sls
+++ b/salt/metalk8s/addons/monitoring/grafana/deployed.sls
@@ -12,12 +12,25 @@
 ---
 apiVersion: v1
 data:
[...]

PR #1186, opened 2019-06-05
Fix alertmanager manifest

### LINKED ISSUE #1175: Alertmanager container fails to start
**Component**: monitoring

**What happened**: the alertmanager container fails to start

**What was expected**: alertmanager container starts correctly

**Steps to reproduce**: `doit vagrant_up`

This seems to be an issue similar to #1170, i.e. a configuration issue in our imported manifests that causes the container to fail to start because of bad YAML.

Logs from the crashing container:

```
level=info ts=2019-06-02T15:10:48.575872663Z caller=main.go:174 msg="Starting Alertmanager" version="(version=0.15.2, branch=HEAD, revision=d19fae3bae451940b8470abb680cfdd59bfa7cfa)"
level=info ts=2019-06-02T15:10:48.591503808Z caller=main.go:175 build_context="(go=go1.10.3, user=root@3101e5b68a55, date=20180814-10:53:39)"
level=warn ts=2019-06-02T15:11:18.716938801Z caller=cluster.go:219 component=cluster msg="failed to join cluster" err="3 errors occurred:\n\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout"
level=info ts=2019-06-02T15:11:18.717078759Z caller=cluster.go:221 component=cluster msg="will retry joining cluster every 10s"
level=warn ts=2019-06-02T15:11:18.717123411Z caller=main.go:265 msg="unable to join gossip mesh" err="3 errors occurred:\n\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout\n* Failed to join 82.216.111.26: dial tcp 82.216.111.26:6783: i/o timeout"
level=info ts=2019-06-02T15:11:18.717340867Z caller=main.go:322 msg="Loading configuration file" file=/etc/alertmanager/config/alertmanager.yaml
level=error ts=2019-06-02T15:11:18.717747141Z caller=main.go:325 msg="Loading configuration file failed" file=/etc/alertmanager/config/alertmanager.yaml err="yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `Z2xvYmF...` into config.plain"
```

jbertran:

**Context**:

The `alertmanager` manifest, like the `grafana` manifest, used a `Secret` with some base64 data, and crashed on an attempt to marshal that data into YAML like `grafana` (see #1170).

The root cause (thanks @gdemonet) is the kubernetes salt module, which expects the data we pass it to be plain text, and does the encoding itself, so we had a double-encoded value passed to K8s, which only decoded it once.

**Summary**:

We fix this in the renderer for now, by decoding the data before passing it to the salt state, since: 

* Reusing the workaround wouldn't work in this case since the `Secret` for `alertmanager` seems to be not directly mounted in the upstream manifest but used by the CR. 
* Rewriting some *more* of the vendored version of the kubernetes salt module means more "magic" and divergence with the upstream of that module.

Commits:

* Handle secret data correctly in the renderer for the saltified upstream manifests
* Revert the `grafana` workaround, since it's no longer needed

**Acceptance criteria**: 

All pods in the `monitoring` namespace should be in Running state upon completion of a bootstrap. All pods in the `monitoring` namespace should eventually be in Running state upon reload of their `deployed` state.

---

Closes: #1175

### DIFF
--- a/salt/_renderers/metalk8s_kubernetes.py
+++ b/salt/_renderers/metalk8s_kubernetes.py
@@ -10,6 +10,7 @@ SLS file. Optionally, you can use rendering pipelines (if templating is
 required), e.g. `#!jinja | kubernetes`.
 '''
 
+import base64
 import yaml
 
 import salt.utils.yaml
@@ -210,11 +211,21 @@ def _handle_v1_namespace(obj, kubeconfig, context):
 
 @handle('v1', 'Secret')
 def _handle_v1_secret(obj, kubeconfig, context):
+    # FIXME #1183: this is a compatibility band-aid for the vendored salt module
+    if 'data' in obj:
+        data = {
+            key: base64.b64decode(value)
+            for key, value in obj['data'].items()
+        }
+    elif 'stringData' in obj:
+        data = obj['stringData']
+    else:
+        data = None
     return {
         'metalk8s_kubernetes.secret_present': [
             {'name': obj['metadata']['name']},
             {'namespace': obj['metadata']['namespace']},
-            {'data': obj['data']},
+            {'data': data},
             {'kubeconfig': kubeconfig},
             {'context': context},
         ],
--- a/salt/metalk8s/addons/monitoring/grafana/upstream.sls
+++ b/salt/metalk8s/addons/monitoring/grafana/upstream.sls
@@ -12,25 +12,12 @@
 ---
 apiVersion: v1
 data:
-  prometheus.yaml: |-
-    {
-        "apiVersion": 1,
-        "datasources": [
-            {
-                "access": "proxy",
-                "editable": false,
-                "name": "prometheus",
-                "orgId": 1,
-                "type": "prometheus",
-                "url": "http://prometheus-k8s.monitoring.svc:9090",
-                "version": 1
-            }
-        ]
-    }
-kind: ConfigMap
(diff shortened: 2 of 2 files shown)


salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -210,11 +211,21 @@ def _handle_v1_namespace(obj, kubeconfig, context):
 
 @handle('v1', 'Secret')
 def _handle_v1_secret(obj, kubeconfig, context):
+    # FIXME #1183: this is a compatibility band-aid for the vendored salt module
+    if 'data' in obj:
+        data = {
+            key: base64.b64decode(value)
+            for key, value in obj['data'].items()
+        }
+    elif 'stringData' in obj:
```
jbertran, line 220: Not sure about these next lines: should we just die in a fire if there is no `data` key?

gdemonet: No it looks good, at least it should work for both cases (don't know if all upstream manifests encode their secrets).

gdemonet (commented):
LGTM, testing it on my machine

gdemonet (approved):
Works fine!

sayf-eddine-scality (approved):
LGTM, and tested

jbertran:
@bert-e approve

PR #1188, opened 2019-06-05
UI- I18n: translation in redux sagas management

### LINKED ISSUE #1154: UI - I18n - translation in redux-sagas
**Component**: UI

Actually, there is no way to translate "text" in saga (for ex: notification title and message => hardcoded). So far, all translations are executed by injectIntl HOC in react containers.

**Why this is needed**:
To be i18n compatible
**What should be done**:
-figure out a solution to translate those labels in saga
...
**Test plan**:

carlito-scality:
**Component**: UI

**Context**: 
Actually, there is no way to translate "text" in saga (for ex: notification title and message => hardcoded). So far, all translations are executed by injectIntl HOC in react containers.

**Summary**:
-Add IntlGlobalProvider which wraps a Singleton to expose ```translate``` function
**Acceptance criteria**: 
All labels in sagas (notifications for example) have to be translated

Closes: https://github.com/scality/metalk8s/issues/1154

### DIFF
--- a/ui/src/translations/fr.json
+++ b/ui/src/translations/fr.json
@@ -37,12 +37,14 @@
   "deployment": "Déploiement",
   "version": "MetalK8s Version",
   "node_deployment": "Déploiement du node",
+  "node_deployment_success": "Déploiement du node {name} a terminé avec succèss",
+  "node_deployment_failed": "Déploiement du node {name} a échoué. {step} - {reason}",
   "back_to_node_list": "Retour à la liste des nodes",
   "success": "Succès",
   "error": "Erreur",
   "completed": "Terminé",
   "node_registered": "Node enregistré",
-  "deployment_started": "Deploiement commencé",
+  "deployment_started": "Déploiement commencé",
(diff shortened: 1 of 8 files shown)


ui/src/translations/fr.json
```diff
@@ -37,18 +37,23 @@
   "deployment": "Déploiement",
   "version": "MetalK8s Version",
   "node_deployment": "Déploiement du node",
```
slaperche-scality, line 39: Not really related to this PR (but since you're updating the translation…), but:
- do we really want to use `grappe` instead of `cluster`?
- on the other hand, we should clearly use `nœud` instead of `node`
- in the English strings, we have `At least on role has to be selected!` => `s/on role/one role/`.

carlito-scality: I'll create another ticket to review all translations https://github.com/scality/metalk8s/issues/1192

MonPote (approved):
LGTM

carlito-scality:
@bert-e approve

Issue #1185, opened 2019-06-05
Improve the UI resiliency to Prometheus API failure. 


MonPote:
**Component**: ui

**Why this is needed**:
As a user, when the API is not reachable the UI should inform the user and work "correctly" (don't crash).

**What should be done**:

- [ ] Design of a way to handle API failure

**Implementation proposal** (strongly recommended):
*in progress*

**Test plan**:
*in progress*

PR #1190, opened 2019-06-05
Fix-ui-crash-if-prometheus-is-not-available

MonPote:
**Component**: ui

**Context**: 
When the prometheus api is not available the ui might break (it cannot create node properly for example)

**Summary**:
It's a programmation error, the promise is directly. We should return the result of a promise not the promise itself.

---

See: #1185

### DIFF
--- a/ui/src/services/prometheus/api.js
+++ b/ui/src/services/prometheus/api.js
@@ -2,7 +2,7 @@ import axios from 'axios';
 
 export async function getAlerts(api) {
   try {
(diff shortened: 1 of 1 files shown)


carlito-scality (approved):
LGTM.
You should add a notification to warn the user  and the cluster status should not be "Down" in this case.

MonPote:
@bert-e approval

MonPote:
@bert-e approve

Issue #1189, opened 2019-06-05
Provide `clean` definition for local images


jbertran:
**Component**: buildchain

**Why this is needed**:

Currently, we do not have a custom `clean` definition for local images, which would do `docker rmi`.

**What should be done**:

Use the Python API to clean images that were built.

**Implementation proposal** (strongly recommended):

Just do it!

**Test plan**:

We should now be able to:

1. Build a local image
2. Ask `doit` to clean that local image
3. Observe via `docker images` that the image was deleted from the local registry

slaperche-scality:
@jbertran is this still relevant?

After you've implemented it, I think we said let's not do it (because it would negate the optimisation that was done for the UI image build)

jbertran:
Ah I closed the PR that tackled this among other things. No longer relevant indeed.

Issue #839, opened 2019-03-23
Improve tasks that save container images


slaperche-scality:
Tasks that save containers images relies on the presence of a file on the disk (either the saved image or a `touch`ed file) to know if the image was pulled/built.

It's not enough though, because if you want to use the image later (e.g. with `docker run`) the image must still exists in Docker (having a local save on the disk is not enough because it could have been `docker rmi`ed), thus to be considered up-to-date those tasks must require that:
- a file exists on the disk (already done)
- the image exists in Docker

To meet the last constraint, we can implement a custom method that would check the existence of the image (by checking the return code of `docker inspect --type=image …`) and use this method for the `uptodate` attribute of the task (see doc [here](http://pydoit.org/dependencies.html#attr-uptodate)).

Since this check is common to both pulled and built images, a good place to implement this method is the base class `ContainerImage`.

EDIT: So far, this only concerns the `metalk8s-build` image.

---

To be considered: is the `touch`ed file still useful now that we also check in Docker? I think not, but to be confirmed.

NicolasT:
This would be fixed if we simply use `skopeo copy` to pull-and-archive the image into a target archive, or not even into an archive (#1049) instead of the Docker kludge, so I'd link the tickets and fix this one by merits of #1049.

slaperche-scality:
Hum I'm not sure about it.

What we will tackle in #1049 is the images we serve from the ISO, and the problem here (seeing that I mention `docker run`) is with images such as `metalk8s-build` which are built and then used to build other artifacts (they are not shipped with the ISO).

I'm not sure `skopeo` will help here.

Edit: In fact I think only `metalk8s-build` is concerned here, so I could as well reformulate the issue to clarify it.

NicolasT:
Ah, good point, my bad.

PR #1191, opened 2019-06-06
Better buildchain docker image management

jbertran:
**Component**:

builchain

**Context**: 

Issues: #839 and #1189 

Some more usage of the docker Python API to further improve how we manage images.

**Summary**:

* Add `docker rmi` callable via the docker API
* Add an image existence check in the local registry via the docker API
* Remove `touch`ed files as image presence flags
* Add `clean` target to `LocalImage`, equivalent to `docker rmi --no-prune=true`

**Acceptance criteria**: 

Isofunctional, no rebuild of images existing in local registry.

---

Closes: #839 
Closes: #1189

### DIFF
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -88,7 +88,18 @@ def docker_image_exists(target_image: image.ContainerImage) -> bool:
         return False
 
 
-
+def docker_remove_image(
+    target_image: image.ContainerImage
+) -> Optional[TaskError]:
+    """
+    Run the `docker rmi` command on the given image through the API client.
+    """
+    try:
+        return DOCKER_CLIENT.images.remove(
+            target_image.tag, force=False, noprune=False
+        )
+    except docker.errors.ImageNotFound:
+        return None
 
 
 @task_errors(docker.errors.BuildError, docker.errors.APIError)
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -79,6 +79,22 @@ def task_errors(*expected_exn: Type[Exception]) -> Callable[[Any], Any]:
 
 # The call method is not counted as a public method
 # pylint: disable=too-few-public-methods
+class DockerExists:
+    """A class providing an image existence check through the API client."""
+    def __init__(self, name: str, version: str):
+        self.name = name
+        self.version = version
+
+    def __call__(self) -> bool:
+        try:
+            docker_image = DOCKER_CLIENT.images.get(
+                '{name}:{version}'.format(name=self.name, version=self.version)
+            )
+            return docker_image is not None
+        except docker.errors.ImageNotFound:
+            return False
+
+
 class DockerBuild:
     """A class to expose the `docker build` command through the API client."""
 
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -77,56 +77,33 @@ def task_errors(*expected_exn: Type[Exception]) -> Callable[[Any], Any]:
     return wrapped_task
 
 
-# The call method is not counted as a public method
-# pylint: disable=too-few-public-methods
-class DockerExists:
-    """A class providing an image existence check through the API client."""
-    def __init__(self, name: str, version: str):
-        self.name = name
-        self.version = version
-
-    def __call__(self) -> bool:
-        try:
-            docker_image = DOCKER_CLIENT.images.get(
-                '{name}:{version}'.format(name=self.name, version=self.version)
-            )
-            return docker_image is not None
-        except docker.errors.ImageNotFound:
-            return False
+def docker_image_exists(target_image: image.ContainerImage) -> bool:
+    """
+    Verify image existence in the local registry through the API client.
+    """
+    try:
+        docker_image = DOCKER_CLIENT.images.get(target_image.tag)
+        return docker_image is not None
+    except docker.errors.ImageNotFound:
+        return False
 
 
-class DockerBuild:
-    """A class to expose the `docker build` command through the API client."""
 
-    def __init__(
-        self,
-        tag: str,
-        path: Path,
-        dockerfile: Path,
-        buildargs: Dict[str, Any]
-    ):
-        """Initialize a `docker tag` callable object.
 
-        Arguments:
-            tag:        the tag to add to the resulting image
-            path:       the build context path
-            dockerfile: the Dockerfile path
-            buildargs:  dict of CLI-equivalent `--build-arg` parameters
-        """
-        self.tag = tag
-        self.path = str(path)
-        self.dockerfile = str(dockerfile)
-        self.buildargs = buildargs
 
-    @task_errors(docker.errors.BuildError, docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
-        return DOCKER_CLIENT.images.build(
-            tag=self.tag,
-            path=self.path,
-            dockerfile=self.dockerfile,
-            buildargs=self.buildargs,
-            forcerm=True,
-        )
+@task_errors(docker.errors.BuildError, docker.errors.APIError)
+def docker_build(target_image: Any) -> Optional[TaskError]:
+    """
+    Run the `docker build` command on the given image through the API client.
+    """
+    docker_context = target_image.dockerfile.parent
+    return DOCKER_CLIENT.images.build(
+        tag=target_image.tag,
+        path=str(docker_context),
+        dockerfile=str(target_image.dockerfile),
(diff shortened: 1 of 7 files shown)


buildchain/buildchain/docker_command.py
```diff
@@ -77,6 +77,25 @@ def decorated_task(*args: Any, **kwargs: Any) -> Optional[TaskError]:
     return wrapped_task
 
 
+class DockerRMI:
```
slaperche-scality: `DockerRemoveImage` is more explicit (and we have autocomplete).

buildchain/buildchain/docker_command.py
```diff
@@ -77,6 +77,25 @@ def decorated_task(*args: Any, **kwargs: Any) -> Optional[TaskError]:
...
+    """
+    A class to expose the `docker rmi` command through the API client.
+
+    This command has `--no-prune` set to True in order to avoid cleaning up
+    intermediary layers that speed up container builds (most notably the GUI).
+    """
+    def __init__(self, name: str, version: str):
```
slaperche-scality: Could take a `ContainerImage` object instead of two separate string.

I mean, it's called `DockerRemoveImage` so I wouldn't be surprised to give an `Image` instance as parameter.

buildchain/buildchain/targets/local_image.py
```diff
@@ -162,9 +162,20 @@ def check_dockerfile_dependencies(self) -> Union[TaskError, bool]:
             return TaskError(msg=error_message)
         return None
 
+    def clean_registry(self) -> Optional[TaskError]:
```
slaperche-scality: minor: you're not really cleaning the registry, `rm_from_registry` would be a slightly better name IMHO

buildchain/buildchain/targets/local_image.py
```diff
@@ -162,9 +162,20 @@ def check_dockerfile_dependencies(self) -> Union[TaskError, bool]:
...
+        """Docker API call for removing the image from local registry."""
+        docker_remove = docker_command.DockerRMI(self.name, self.version)
```
slaperche-scality: Why not simply `return docker_command.DockerRMI(self)()`?

buildchain/buildchain/targets/local_image.py
```diff
@@ -162,9 +162,20 @@ def check_dockerfile_dependencies(self) -> Union[TaskError, bool]:
...
+        return docker_remove()
+
     @property
     def task(self) -> types.TaskDict:
         task = self.basic_task
+
+        if self.save_on_disk:
```
slaperche-scality, line 169:
```
clean =  [self.clean_registry]
if self.save_on_disk:
    clean.append(super().clean)
```

buildchain/buildchain/docker_command.py
```diff
@@ -77,6 +77,24 @@ def decorated_task(*args: Any, **kwargs: Any) -> Optional[TaskError]:
     return wrapped_task
 
 
+# The call method is not counted as a public method
+# pylint: disable=too-few-public-methods
```
slaperche-scality: Do you still need this one? Seems not required for `DockerRmi` which has the same number of method.

buildchain/buildchain/docker_command.py
```diff
@@ -77,6 +77,24 @@ def decorated_task(*args: Any, **kwargs: Any) -> Optional[TaskError]:
...
+class DockerExists:
+    """A class providing an image existence check through the API client."""
+    def __init__(self, name: str, version: str):
```
slaperche-scality: I would take either a single parameter `tag` (or whatever is the proper naming) or directly an `ContainerImage` instance.

buildchain/buildchain/packaging.py
```diff
@@ -139,7 +139,7 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
     dockerfile=constants.ROOT/'packages'/'Dockerfile',
     destination=config.BUILD_ROOT,
     save_on_disk=False,
-    task_dep=['_build_root'],
+    task_dep=[],
```
slaperche-scality: remove the `task_dep` parameter totally, it will default to an empty list.

slaperche-scality (changes requested):

buildchain/buildchain/docker_command.py
jbertran: This is complicated in the current order of things, as we don't want to depend on `image` for things in `docker_command` or we get hit with circular imports when we type them.

buildchain/buildchain/targets/local_image.py
jbertran: I'd rather have an explicit setup of the `self.save_on_disk` divergences at the cost of very minor code duplication.

slaperche-scality: But it's less clear.
In both case we do need to clear the image on the registry and iif we save it on disk then we also need to clean it on disk.

buildchain/buildchain/docker_command.py
jbertran: Stray rebase removal, it's needed on both (at least it is locally).

slaperche-scality: You sure? `doit.sh lint` doesn't complain, even locally for me.
Are you sure it's not your weird editor linter that complains :p ?

Since we froze the version of `pylint` we should have the same results.

Edit: Ok, now it's weird.

If I remove all the `disable=too-few-public-methods` in the file (the on `DockerExists` and the one on `DockerBuild` I get three errors:

```
R: 81, 0: Too few public methods (1/2) (too-few-public-methods)
R: 97, 0: Too few public methods (1/2) (too-few-public-methods)
R:117, 0: Too few public methods (1/2) (too-few-public-methods)
```
It complains on `DockerExists`, `DockerBuild` and `DockerRmi` oO
If I add back the disable on `Exists` and `Build`, no more warning on `Rmi`.

slaperche-scality: Ok, seems like our `disable` are not scope to the class, but global to the file.

Either we got the syntax wrong, either it's a bug in pylint. Will need to check the doc…

buildchain/buildchain/docker_command.py
slaperche-scality: The thing is, we already do: you already have an `from buildchain.targets import image` in this file so I don't see where the problem is.

buildchain/buildchain/docker_command.py
slaperche-scality: Ok, the correct syntax seems to be:

```
class DockerExists:
    # The call method is not counted as a public method
    # pylint:disable=too-few-public-methods
```

To disable it only for the class and not the whole file (I also noticed that we usually don't put spaces before `disable`, but that's minor).

See this [FAQ entry](https://docs.pylint.org/en/1.6.0/faq.html#is-it-possible-to-locally-disable-a-particular-message):

> This may be done by adding “#pylint: disable=some-message,another-one” **at the desired block level** or at the end of the desired line of code

jbertran: Removed the need for these by switching to functions.

jbertran:
Blocked by the lack of understanding of the origin of the following error:

```
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/buildchain/.venv/lib64/python3.6/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/buildchain/.venv/lib64/python3.6/site-packages/urllib3/connectionpool.py", line 384, in _make_request
    six.raise_from(e, None)
  File "<string>", line 2, in raise_from
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/buildchain/.venv/lib64/python3.6/site-packages/urllib3/connectionpool.py", line 380, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib64/python3.6/http/client.py", line 1331, in getresponse
    response.begin()
  File "/usr/lib64/python3.6/http/client.py", line 297, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python3.6/http/client.py", line 279, in _read_status
    raise BadStatusLine(line)
http.client.BadStatusLine: 0
```

It seems to be pointing to a protocol mismatch in the docker client and daemon configurations, but so far I haven't been able to fix it.

The PR may stay `blocked` as it is non-essential, and could advantageously be replaced by another container management tool.

jbertran:
Most of this is tackled by #1846
By this point the buildchain has changed so much that the PR no longer has any remaining purpose.

slaperche-scality:
The part about `rm`ing the image could be kept though (but yeah, maybe in another PR, this one would be too hard to rebase :grin: )

Issue #791, opened 2019-03-19
Deploy `ingress-nginx`


NicolasT:
- [ ] Deploy `ingress-nginx` as the cluster-provided `Ingress` controller
- [ ] Expose `ingress-nginx` to the outside
  * [ ] Initially: using `hostNetwork`
  * [ ] Once #779 lands: as a `Service` of type `LoadBalancer`

Once this is available, #729 can be updated to use an `Ingress` object.

PR #1194, opened 2019-06-06
Add documentation about how install the root CA on web browser

ycointe:
**Component**: docs

**Context**:

Currently, the procedure for reaching MetalK8s UI requires one to navigate to `kube-apiserver` URL in one's browser to accept its server certificate.
While this procedure works fine for now, it will become cumbersome as soon as other services become secured with TLS. Instead, we can trust the root CA to automatically accept all certificates signed by this CA.

**Summary**: How to  install the root CA on web browser

**Acceptance criteria**: Be able to install the root CA on web browser and don't use the link "Accept SSL certificate" on the metalk8s UI login page.

---

Closes:

### DIFF
--- a/docs/add-root-ca.rst
+++ b/docs/add-root-ca.rst
@@ -1,0 +1,46 @@
+Add a root CA to the web browser
+================================
+
+.. todo::
+
+  The following test has been realized on Vagrant only
+  Web browsers version used for tests:
+
+   * Googe Chrome: Version 75.0.3770.80
+   * Mozilla Firefox: Version 67.0
+
+Run the following command as ``root`` on the bootstrap node to copy the root CA
+from the bootstrap on your host:
+
+.. code-block:: shell
+
+  $ scp /etc/kubernetes/pki/ca.crt <user>@<IP>:<directory>
+
+Then go back on you host and open you web browser ton install the root CA.
+
+Install on Google Chrome
+------------------------
+
+  * Open your web browser
+  * Navigate on ``chrome://settings``
+  * On the search setting bar, search ``manage certificates``.
+  * Then go in the tab ``Authorities``, click on :kbd:`IMPORT`.
+  * A new window appear, then find the directory where you upload the root CA
+    then select.
+  * A pop up apear to ask you to set the Trust settings:
+  * Select the line ``Trust this certificate for identifying websites``,
+    then click on :kbd:`OK` to confirm.
+
+Install on Mozilla Firefox
+--------------------------
+
+  * Open your web browser
+  * Navigate on ``about:preferences``
+  * On the search bar, search ``certificates``. In the search result
+  * click on the button :kbd:`View certificates...`
+  * Go on the tab ``Authorities`` then click on :kbd:`IMPORT`
+  * A new window appear, then find the directory where you upload the root CA
+    then select.
+  * A pop up apear to ask you to set the Trust settings:
+  * Select the line ``Trust this CA for identifying websites``,
+    then click on :kbd:`OK` to confirm.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,7 +21,7 @@ Welcome to MetalK8s's documentation!
 
    installation-guide/index
    developer/index
-
+   add-root-ca
 
 
 Indices and tables


docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
+Add a root CA to the web browser
+================================
+
+.. todo::
+
+  The following test has been realized on vagrant only - One should list
```
slaperche-scality: s/vagrant/Vagrant/

docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
...
+  all requirements for reaching the UI using an operationnal cluster.
+  Web browsers version used for tests:
+  Googe Chrome: Version 75.0.3770.80 | Mozilla Firefox: Version 67.0
```
slaperche-scality: Why not a list formatting (it would be more readable).

docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Run the following command as ``root`` on the bootstrap node:
```
slaperche-scality: Missing command? (there is nothing after your `:` but another sentence that ends with `:`)

docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
...
+.. code-block:: shell
+
+  $ scp /etc/kubernetes/pki/ca.crt <user>@<IP>:<directory>
+
+Then go back on you host and open you web browser ton install the root CA.
+
+Install on Google Chrome
+------------------------
+
+Navigate on ``chrome://settings``
```
slaperche-scality: Since this is a list of instruction to follow sequentially, wouldn't it make sense to format it as a bullet list?

docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
...
+then select.
+
+A pop up apear to ask you to set the Trust settings:
+Select the line ``Trust this certificate for identifying websites``
+Then click on :kbd:`OK` to confirm.
+
+Install on Mozilla Firefox
+--------------------------
+
+Navigate on ``about:preferences``
```
slaperche-scality: ditto

slaperche-scality (changes requested):
You could use

> Currently, the procedure for reaching MetalK8s UI requires one to navigate to kube-apiserver URL in one's browser to accept its server certificate.
> While this procedure works fine for now, it will become cumbersome as soon as other services become secured with TLS. Instead, we can trust the root CA to automatically accept all certificates signed by this CA.

as commit message

docs/add-root-ca.rst
ycointe: Correction done.

docs/add-root-ca.rst
ycointe: I tried to do a list, please tell me if it's better like this.

docs/add-root-ca.rst
ycointe: Correction done.

docs/add-root-ca.rst
ycointe: I tried to do a list, please tell me if it's better like this.

docs/add-root-ca.rst
ycointe: Done the same.

docs/add-root-ca.rst
```diff
@@ -0,0 +1,45 @@
+Add a root CA to the web browser
+================================
+
+.. todo::
+
+  | The following test has been realized on Vagrant only
```
slaperche-scality: I don't think you need the `|`.

```
.. todo::

  The following test has been realized on Vagrant only
  Web browsers version used for tests:
 
  * Googe Chrome: Version 75.0.3770.80
  * Mozilla Firefox: Version 67.0

Run the following command as ``root`` on the bootstrap node to copy the root CA
from the bootstrap on your host:
```

slaperche-scality (changes requested):
Commit message still not updated, otherwise LGTM

docs/add-root-ca.rst
slaperche-scality: it works!

docs/add-root-ca.rst
slaperche-scality: It works!

docs/add-root-ca.rst
ycointe: I removed them.

gdemonet:
As @NicolasT explained, we shouldn't do this, as it is really bad practice.

> The K8s CA(s) are cluster-internal only and shall never be used as a CA outside it, except when provided in a 'kubeconfig' client configuration file which doesn't outlive a cluster (authn would fail anyway).

Instead, we should focus on building a real solution for exposing such services to outside the cluster, namely #791.

I'm closing this PR.

Issue #1116, opened 2019-05-16
buildchain: add tasks to check external dependencies


slaperche-scality:
**Component**:

Buildchain

**Why this is needed**:

Currently, when you're missing an external command used by the build chain (such
as `docker`, `skopeo`, …) you'll get a traceback like this one:

```
 Traceback (most recent call last):
   File "[…]/.venv/lib/python3.7/site-packages/doit/doit_cmd.py", line 177, in run
     return command.parse_execute(args)
*  File "[…]/.venv/lib/python3.7/site-packages/doit/cmd_base.py", line 127, in parse_execute*What should be done**:
     return self.execute(params, args)
*  File "[…]/.venv/lib/python3.7/site-packages/doit/cmd_base.py", line 420, in execute      *Implementation proposal** (strongly recommended):
     return self._execute(**exec_params)
*  File "[…]/.venv/lib/python3.7/site-packages/doit/cmd_run.py", line 260, in _execute      *Test plan**:
     return runner.run_all(self.control.task_dispatcher())
   File "[…]/.venv/lib/python3.7/site-packages/doit/runner.py", line 256, in run_all
     self.run_tasks(task_dispatcher)
   File "[…]/.venv/lib/python3.7/site-packages/doit/runner.py", line 222, in run_tasks
     catched_excp = self.execute_task(node.task)
   File "[…]/.venv/lib/python3.7/site-packages/doit/runner.py", line 177, in execute_task
     return task.execute(self.stream)
   File "[…]/.venv/lib/python3.7/site-packages/doit/task.py", line 449, in execute
     action_return = action.execute(task_stdout, task_stderr)
   File "[…]/.venv/lib/python3.7/site-packages/doit/action.py", line 219, in execute
     **subprocess_pkwargs)
   File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
     restore_signals, start_new_session)
   File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
     raise child_exception_type(errno_num, err_msg, err_filename)
 FileNotFoundError: [Errno 2] No such file or directory: 'skopeo': 'skopeo'
```

While it does mention the missing binary, it's not really user-friendly.

**What should be done**:

Following a suggestion from @NicolasT, we could define tasks that would check
the existence of the binary (and if not found, it would bail out with a nice
error message).

Then, tasks that use an external binary would have a dependence on the
corresponding task that check its presence (using `task_dep`).

**Implementation proposal**:

[...]

PR #1233, opened 2019-06-14
Feature/1116 add new check external dependency tasks

Ebaneck:
**Component**:

Buildchain

**Context**: 

Issue: https://github.com/scality/metalk8s/issues/1116

**Summary**:

Improve the management of external dependencies properly and become gentle with traceback printed to the user

**Acceptance criteria**: 

Generate a clean trackback if a required external dependency is missing.

E.g
`TaskError - taskid:check_for:skopeo
command skopeo not found in $PATH

########################################
check_for:skopeo <stdout>:`

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -47,7 +47,6 @@ VAGRANT_UP_ARGS : Tuple[str, ...] = tuple(shlex.split(
 
 # Name of the command (if in the PATH) or path to the binary.
 
-DOCKER   : str = os.getenv('DOCKER_BIN',   'docker')
 GIT      : str = os.getenv('GIT_BIN',      'git')
 HARDLINK : str = os.getenv('HARDLINK_BIN', 'hardlink')
 MKISOFS  : str = os.getenv('MKISOFS_BIN',  'mkisofs')
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -9,6 +9,7 @@ Those can be edited by the user to customize/configure the build system.
 
 import os
 import shlex
+import enum
 
 from typing import Tuple
 from pathlib import Path
@@ -45,12 +46,16 @@ VAGRANT_UP_ARGS : Tuple[str, ...] = tuple(shlex.split(
 
 # External commands {{{
 
-# Name of the command (if in the PATH) or path to the binary.
+# Name of the external commands (if in the PATH) or path to the binary.
 
-GIT      : str = os.getenv('GIT_BIN',      'git')
-HARDLINK : str = os.getenv('HARDLINK_BIN', 'hardlink')
-MKISOFS  : str = os.getenv('MKISOFS_BIN',  'mkisofs')
-SKOPEO   : str = os.getenv('SKOPEO_BIN',   'skopeo')
-VAGRANT  : str = os.getenv('VAGRANT_BIN',  'vagrant')
+
+class ExtCommand(enum.Enum):
+    """External commands used by the build chain."""
+
+    GIT      = os.getenv('GIT_BIN',      'git')
+    HARDLINK = os.getenv('HARDLINK_BIN', 'hardlink')
+    MKISOFS  = os.getenv('MKISOFS_BIN',  'mkisofs')
+    SKOPEO   = os.getenv('SKOPEO_BIN',   'skopeo')
+    VAGRANT  = os.getenv('VAGRANT_BIN',  'vagrant')
 
 # }}}
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -91,8 +91,8 @@ def git_ref() -> Optional[str]:
 
     try:
         ref : bytes = subprocess.check_output([
-            config.GIT, 'describe', '--always', '--long', '--tags',
-                '--dirty',
+            config.ExtCommand.GIT.value, 'describe', '--always', '--long',
+            '--tags', '--dirty',
         ])
 
         return ref.decode('utf-8').rstrip()
--- a/buildchain/buildchain/deps.py
+++ b/buildchain/buildchain/deps.py
@@ -1,4 +1,3 @@
-
 """Dependency checker for skopeo, vagrant, git, mkisofs and hard-link."""
 
 
--- a/buildchain/buildchain/deps.py
+++ b/buildchain/buildchain/deps.py
@@ -1,0 +1,61 @@
+
+"""Dependency checker for skopeo, vagrant, git, mkisofs and hard-link."""
+
+
+from pathlib import Path
+from typing import Optional, Iterator
+import shutil
+
+from doit.exceptions import TaskError  # type: ignore
+
+from buildchain import config
+from buildchain import constants
+from buildchain import types
+from buildchain import utils
+
+
+def task_check_for() -> Iterator[types.TaskDict]:
+    """Check the existence of the external commands."""
+    def get_command_location(
+        command_name: str, binary_path: Optional[str]
+    ) -> Optional[TaskError]:
+        """Check the existence of the given command.
+
+        If the user already specified a path for it, check for the existence
+        of the binary at the given location. Otherwise, look it up from PATH.
+
+        If the binary is not found, the task fails.
+        """
+        if binary_path:
+            path = Path(binary_path)
+            cmd_path = shutil.which(path.name, path=path.parent)
+        else:
+            cmd_path = shutil.which(command_name)
+        if cmd_path is None:
+            return TaskError(msg='command {} not found in {}'.format(
+                command_name, Path(binary_path).parent
+                if binary_path else '$PATH'
+            ))
+        return None
+
+    for ext_cmd in config.ExtCommand:
+        cmd_name = ext_cmd.name.lower()
+        cmd_path = ext_cmd.value if ext_cmd.value != cmd_name else None
+
+        def show(name: str) -> str:
+            return '{cmd: <{width}} {name}'.format(
+                cmd='CHECK CMD', width=constants.CMD_WIDTH, name=name
+            )
+
+        yield {
+            'name': cmd_name,
+            'title': lambda _, name=cmd_name: show(name),
+            'doc': 'Check the presence of the {} command.'.format(cmd_name),
+            'actions': [(get_command_location, [cmd_name, cmd_path], {})],
+            'file_dep': [],
+            'task_dep': [],
+            'uptodate': [False],
+        }
+
+
+__all__ = utils.export_only_tasks(__name__)
--- a/buildchain/buildchain/image.py
+++ b/buildchain/buildchain/image.py
@@ -80,9 +80,13 @@ def task__image_dedup_layers() -> types.TaskDict:
             path=utils.build_relpath(constants.ISO_IMAGE_ROOT)
         )
 
-    task = targets.Target(task_dep=['_image_pull', '_image_build']).basic_task
+    task = targets.Target(
+        task_dep=['check_for:hardlink', '_image_pull', '_image_build']
+    ).basic_task
     task['title']   = lambda _: show()
-    task['actions'] = [[config.HARDLINK, '-c', constants.ISO_IMAGE_ROOT]]
+    task['actions'] = [
+        [config.ExtCommand.HARDLINK.value, '-c', constants.ISO_IMAGE_ROOT]
+    ]
     return task
 
 
--- a/buildchain/buildchain/iso.py
+++ b/buildchain/buildchain/iso.py
@@ -160,7 +160,7 @@ def task__iso_generate_product_txt() -> types.TaskDict:
 def task__iso_build() -> types.TaskDict:
     """Create the ISO from the files in ISO_ROOT."""
     mkisofs = [
-        config.MKISOFS, '-output',  ISO_FILE,
+        config.ExtCommand.MKISOFS.value, '-output',  ISO_FILE,
         '-quiet',
         '-rock',
         '-joliet',
@@ -186,7 +186,7 @@ def task__iso_build() -> types.TaskDict:
         'actions': [mkisofs],
         'targets': [ISO_FILE],
         'file_dep': depends,
-        'task_dep': ['_build_root', '_iso_mkdir_root'],
+        'task_dep': ['check_for:mkisofs', '_build_root', '_iso_mkdir_root'],
         'clean': True,
     }
 
--- a/buildchain/buildchain/targets/local_image.py
+++ b/buildchain/buildchain/targets/local_image.py
@@ -107,6 +107,7 @@ class LocalImage(image.ContainerImage):
         self._save = save_on_disk
         self._build_args = build_args or {}
         kwargs.setdefault('file_dep', []).append(self.dockerfile)
+        kwargs.setdefault('task_dep', []).append('check_for:skopeo')
         super().__init__(
             name=name, version=version, destination=destination, **kwargs
         )
@@ -192,8 +193,9 @@ class LocalImage(image.ContainerImage):
         # If a destination is defined, let's save the image there.
         if self.save_on_disk:
             cmd = [
-                config.SKOPEO, '--override-os', 'linux', '--insecure-policy',
-                'copy', '--format', 'v2s2', '--dest-compress',
+                config.ExtCommand.SKOPEO.value, '--override-os', 'linux',
+                '--insecure-policy', 'copy', '--format', 'v2s2',
+                '--dest-compress',
             ]
             docker_host = os.getenv('DOCKER_HOST')
             if docker_host is not None:
--- a/buildchain/buildchain/targets/remote_image.py
+++ b/buildchain/buildchain/targets/remote_image.py
@@ -52,6 +52,7 @@ class RemoteImage(image.ContainerImage):
         self._digest = digest
         self._remote_name = remote_name or name
         self._use_tar = save_as_tar
+        kwargs.setdefault('task_dep', []).append('check_for:skopeo')
         super().__init__(
             name=name, version=version,
             destination=destination,
@@ -104,8 +105,8 @@ class RemoteImage(image.ContainerImage):
     def _skopeo_copy(self) -> List[str]:
         """Return the command line to execute skopeo copy."""
         cmd = [
-            config.SKOPEO, '--override-os', 'linux', '--insecure-policy',
-            'copy', '--format', 'v2s2'
+            config.ExtCommand.SKOPEO.value, '--override-os', 'linux',
+            '--insecure-policy', 'copy', '--format', 'v2s2'
         ]
         if not self._use_tar:
             cmd.append('--dest-compress')
--- a/buildchain/buildchain/vagrant.py
+++ b/buildchain/buildchain/vagrant.py
@@ -41,52 +41,53 @@ def task__vagrantkey() -> types.TaskDict:
 
 def task__vagrant_up_noprov() -> types.TaskDict:
     """Run `vagrant up` without provisioning a development environment."""
-    cmd = [config.VAGRANT, 'up']
+    cmd = [config.ExtCommand.VAGRANT.value, 'up']
     cmd.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
     cmd.append('--no-provision')
 
     return {
         'actions': [doit.tools.LongRunning(' '.join(cmd))],
         'file_dep': [],
-        'task_dep': [],
+        'task_dep': ['check_for:vagrant'],
         'uptodate': [False],
     }
 
 def task__vagrant_snapshot() -> types.TaskDict:
     """Snapshot development environment."""
-    vagranthalt = [config.VAGRANT, 'halt', 'bootstrap']
-    vagrantsnap = [config.VAGRANT, 'snapshot', 'save', '--force', 'bootstrap',
-        config.VAGRANT_SNAPSHOT_NAME]
+    vagranthalt = [config.ExtCommand.VAGRANT.value, 'halt', 'bootstrap']
+    vagrantsnap = [config.ExtCommand.VAGRANT.value, 'snapshot', 'save',
+                   '--force', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
 
     return {
         'actions': [vagranthalt, vagrantsnap],
-        'task_dep': ['_vagrant_up_noprov'],
+        'task_dep': ['_vagrant_up_noprov', 'check_for:vagrant'],
         'uptodate': [doit.tools.run_once],
         'verbosity': 2
     }
 
 def task_vagrant_restore() -> types.TaskDict:
     """Restore development environment snapshot."""
-    cmd = [config.VAGRANT, 'snapshot', 'restore', 'bootstrap',
+    cmd = [config.ExtCommand.VAGRANT.value, 'snapshot', 'restore', 'bootstrap',
         config.VAGRANT_SNAPSHOT_NAME]
 
     return {
         'actions': [cmd],
         'uptodate': [False],
-        'verbosity': 2
+        'verbosity': 2,
+        'task_dep': ['check_for:vagrant']
     }
 
 def task_vagrant_up() -> types.TaskDict:
     """Run `vagrant up` to (re-)provision a development environment."""
-    vagrantup = [config.VAGRANT, 'up']
+    vagrantup = [config.ExtCommand.VAGRANT.value, 'up']
     vagrantup.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
 
-    vagrantdestroy = [config.VAGRANT, 'destroy', '--force']
+    vagrantdestroy = [config.ExtCommand.VAGRANT.value, 'destroy', '--force']
 
     return {
         'actions': [doit.tools.LongRunning(' '.join(vagrantup))],
         'file_dep': [constants.VAGRANT_SSH_KEY_PAIR],
-        'task_dep': ['populate_iso', '_vagrant_snapshot'],
+        'task_dep': ['check_for:vagrant', 'populate_iso', '_vagrant_snapshot'],
         'uptodate': [False],
         'clean': [vagrantdestroy],
         'verbosity': 2
--- a/buildchain/dodo.py
+++ b/buildchain/dodo.py
@@ -9,6 +9,7 @@
 import doit  # type: ignore
 
 from buildchain.build import *
+from buildchain.deps import *
 from buildchain.image import *
 from buildchain.iso import *
 from buildchain.lint import *


buildchain/buildchain/image.py
```diff
@@ -82,7 +83,7 @@ def show() -> str:
 
     task = targets.Target(task_dep=['_image_pull', '_image_build']).basic_task
```
gdemonet: Would need a dependency on `hardlink` then, no?

```suggestion
     task = targets.Target(
         task_dep=['check_for:hardlink', '_image_pull', '_image_build']
     ).basic_task
```

charlesprost:
@Ebaneck just got this, any clue?

```
$ ./doit.sh 
-- MKDIR        _build
-- MKDIR        _build/root
-- MKDIR        _build/root/examples
.  check_for:docker
-- MKDIR        _build/root/images
-- MKTREE       _build/root/pillar/metalk8s/roles
.  CPTREE       pillar
.  RENDER       _build/root/pillar/top.sls
.  MKTREE       _build/root/salt/metalk8s/addons/monitoring/grafana
.  CPTREE       salt
-- RENDER       _build/root/salt/top.sls
-- RENDER       _build/root/salt/metalk8s/kubernetes/mark-control-plane/deployed.sls
-- MKDIR        _build/packages
-- MKDIR        _build/root/packages
-- MKDIR        _build/root/packages/metalk8s-scality-el7
-- MKDIR        _build/root/packages/metalk8s-scality-el7/x86_64
.  GENERATE     _build/root/product.txt
.  RENDER       _build/root/bootstrap.sh
-- GENERATE     _build/root/examples/new-node.yaml
.  check_for:skopeo
-- PULL IMG     coredns:1.3.1
-- PULL IMG     etcd:3.2.18
-- PULL IMG     grafana:5.2.4
-- PULL IMG     kube-apiserver:1.11.7
-- PULL IMG     kube-controller-manager:1.11.7
-- PULL IMG     kube-proxy:1.11.7
-- PULL IMG     kube-rbac-proxy:v0.3.1
-- PULL IMG     kube-scheduler:1.11.7
-- PULL IMG     kube-state-metrics:v1.3.1
-- PULL IMG     nginx:1.15.8
-- PULL IMG     node-exporter:v0.17.0
-- PULL IMG     prometheus:v2.4.3
-- PULL IMG     prometheus-config-reloader:v0.23.2
.  PULL IMG     addon-resizer:1.8.3
TaskError - taskid:_image_pull:addon-resizer
CmdAction Error creating command string
  File "/home/nootal/projects/metalk8s/buildchain/.venv/lib/python3.6/site-packages/doit/action.py", line 196, in execute
    action = self.expand_action()
  File "/home/nootal/projects/metalk8s/buildchain/.venv/lib/python3.6/site-packages/doit/action.py", line 277, in expand_action
    msg % (self.task.name, element, type(element)))
doit.exceptions.InvalidTask: _image_pull:addon-resizer. CmdAction element must be a str or Path from pathlib. Got 'None' (<class 'NoneType'>)

########################################
_image_pull:addon-resizer <stdout>:
```

buildchain/buildchain/deps.py
```diff
@@ -0,0 +1,36 @@
...
+def task_check_for():
+
+    def get_command_location(command_name: str, binary_path: Optional[str]
+                             ) -> Optional[TaskError]:
+        """Search the location of the command `command_name`.
+
+        If the user already specified a path for it, check for the existence
+        of the binary at the given location. Otherwise, look it up from PATH.
+
+        If the binary is found, the `CMD` dict is updated, otherwise it fails.
```
gdemonet: I'm quite lost. Let me detail the few code paths I understand of your code:

_Case 1:_ `SOME_CMD_BIN` is not defined in the environment, so this task will search in `$PATH` and override `CMD['some-cmd']`

_Case 2_: `SOME_CMD_BIN` is set to `/some/path/some-cmd`, so this task will search in `/some/path` and set `CMD['some-cmd']` to the same value as it was already

_Case 3_: `SOME_CMD_BIN` is set to `/some/path/alias-cmd` (note the different binary name), so this task will search for `some-cmd` in `/some/path`, and **will fail**.

I believe we need to be more declarative and less magic about finding binaries we rely on.

First, using a mutable dict like this prevents static checkers from knowing whether the relevant command is defined in `config.CMD`. We could have used an `enum.Enum` to fix that aspect.

Second, you are mutating the dict at every execution of `check_for:some-cmd`, redundant though not critical.

Finally, I'd rather let the user responsible of managing his commands: the task could just _check_ if the provided path exists, complain if not, and get on with it. No need to mutate anything, nor to run this check for each task that rely on it.

gdemonet (changes requested):
Good usability improvement, although I think we're missing a situation where the user wants to define an alias for some binaries (rename `vagrant` to `vagrant-X.Y` for instance).

Another point, though I'm not sure it's a use-case we want to support, we can't use the `XXX_BIN` env vars to add options, such as verbosity.

Ebaneck:
> @Ebaneck just got this, any clue?
> 
> ```
> $ ./doit.sh 
> -- MKDIR        _build
> -- MKDIR        _build/root
> -- MKDIR        _build/root/examples
> .  check_for:docker
> -- MKDIR        _build/root/images
> -- MKTREE       _build/root/pillar/metalk8s/roles
> .  CPTREE       pillar
> .  RENDER       _build/root/pillar/top.sls
> .  MKTREE       _build/root/salt/metalk8s/addons/monitoring/grafana
> .  CPTREE       salt
> -- RENDER       _build/root/salt/top.sls
> -- RENDER       _build/root/salt/metalk8s/kubernetes/mark-control-plane/deployed.sls
> -- MKDIR        _build/packages
> -- MKDIR        _build/root/packages
> -- MKDIR        _build/root/packages/metalk8s-scality-el7
> -- MKDIR        _build/root/packages/metalk8s-scality-el7/x86_64
> .  GENERATE     _build/root/product.txt
> .  RENDER       _build/root/bootstrap.sh
> -- GENERATE     _build/root/examples/new-node.yaml
> .  check_for:skopeo
> -- PULL IMG     coredns:1.3.1
> -- PULL IMG     etcd:3.2.18
> -- PULL IMG     grafana:5.2.4
> -- PULL IMG     kube-apiserver:1.11.7
> -- PULL IMG     kube-controller-manager:1.11.7
> -- PULL IMG     kube-proxy:1.11.7
> -- PULL IMG     kube-rbac-proxy:v0.3.1
> -- PULL IMG     kube-scheduler:1.11.7
> -- PULL IMG     kube-state-metrics:v1.3.1
> -- PULL IMG     nginx:1.15.8
> -- PULL IMG     node-exporter:v0.17.0
> -- PULL IMG     prometheus:v2.4.3
> -- PULL IMG     prometheus-config-reloader:v0.23.2
> .  PULL IMG     addon-resizer:1.8.3
> TaskError - taskid:_image_pull:addon-resizer
> CmdAction Error creating command string
> Traceback (most recent call last):
>   File "/home/nootal/projects/metalk8s/buildchain/.venv/lib/python3.6/site-packages/doit/action.py", line 196, in execute
>     action = self.expand_action()
>   File "/home/nootal/projects/metalk8s/buildchain/.venv/lib/python3.6/site-packages/doit/action.py", line 277, in expand_action
>     msg % (self.task.name, element, type(element)))
> doit.exceptions.InvalidTask: _image_pull:addon-resizer. CmdAction element must be a str or Path from pathlib. Got 'None' (<class 'NoneType'>)
> 
> ########################################
> _image_pull:addon-resizer <stdout>:
> ```

@nootal This is a work in progress... We are currently refactoring some of the logic + this is marked a draft PR.. Will ping you when it is fully ready for review

buildchain/buildchain/deps.py
Ebaneck: @gdemonet I would certainly look into the potential use of enumerators.

In regards to the second potential issue of mutating Dicts each time; I think this is pinned to the underline functioning of **Python Doit**.

**step i.** All tasks are created + graphs are built.
**step ii.** Tasks are executed.

What does this mean for us??
Once tasks are created, the dict is initialized with none until the tasks are ran  For example, the dict will only be updated if  `check_for:docker` had ran before hence the mutation:

`TaskError - taskid:check_for:skopeo
command skopeo {'docker': '/usr/bin/docker', 'skopeo': None, 'vagrant': None, 'mkisofs': None, 'hardlink': None} not found in $PATH
`

@slaperche-scality has proposed a better way of handling this and we are looking into this.

And yes with this new approach, we potentially will take off mutations each time using `        'actions': [CmdAction(create_cmd_string)],`

slaperche-scality:
> Good usability improvement, although I think we're missing a situation where the user wants to define an alias for some binaries (rename `vagrant` to `vagrant-X.Y` for instance).

If someone has an alias, it can uses the `XXX_BIN`, no?
That's the point of those actually, to handle case when we are not in the default setup (binary with default name in the `PATH`).

> Another point, though I'm not sure it's a use-case we want to support, we can't use the `XXX_BIN` env vars to add options, such as verbosity.

It more or less supported for `vagrant` (with `VAGRANT_UP_ARGS`), but not sure we want to go that way.
We could also leverage the `verbose` options of `doit` and patch the commands ourself when it's set.
Or in some other way.

Anyway, this is not related to this PR (but could be tracked in a dedicated issue :slightly_smiling_face: )

buildchain/buildchain/deps.py
slaperche-scality: > Case 2: SOME_CMD_BIN is set to /some/path/some-cmd, so this task will search in /some/path and set CMD['some-cmd'] to the same value as it was already

Yeah, we don't change the value here.
The point is to check that we indeed have a `some-cmd` and that it's executable, … in the given path.

> No need to mutate anything, nor to run this check for each task that rely on it.
You need to run it for every task that rely on it because you don't know WHICH command will be run, and as said before it should be run only once per execution.

> Second, you are mutating the dict at every execution of check_for:some-cmd, redundant though not critical.

For that point, we could probably have a custom `uptodate` that run the task only if the dict entry is `None`: that way it'll only be run once, even if multiple tasks depends on it.
But I think even as it is, the task is only run once (to be confirmed) so it's unnecessary.

gdemonet:
> > Good usability improvement, although I think we're missing a situation where the user wants to define an alias for some binaries (rename `vagrant` to `vagrant-X.Y` for instance).
> 
> If someone has an alias, it can uses the `XXX_BIN`, no?
> That's the point of those actually, to handle case when we are not in the default setup (binary with default name in the `PATH`).

No, that's my point actually (see https://github.com/scality/metalk8s/pull/1233#discussion_r294186349):

[...] 3. `SOME_CMD_BIN` is set to `/some/path/alias-cmd` (note the different binary name), so this task will search for `some-cmd` in `/some/path`, and **will fail**.


> > Another point, though I'm not sure it's a use-case we want to support, we can't use the `XXX_BIN` env vars to add options, such as verbosity.
> 
> It more or less supported for `vagrant` (with `VAGRANT_UP_ARGS`), but not sure we want to go that way.
> We could also leverage the `verbose` options of `doit` and patch the commands ourself when it's set.
> Or in some other way.
> 
> Anyway, this is not related to this PR (but could be tracked in a dedicated issue  )

Agreed, let's not focus on this right now.

buildchain/buildchain/deps.py
gdemonet: > In regards to the second potential issue of mutating Dicts each time; I think this is pinned to the underline functioning of Python Doit.
>
> **step i.** All tasks are created + graphs are built.
> **step ii.** Tasks are executed.

I don't see how this means you **need** mutability. Most importantly, my point is that we don't need to mutate _anything_.

> For that point, we could probably have a custom `uptodate` that run the task only if the dict entry is `None`: that way it'll only be run once, even if multiple tasks depends on it.
> But I think even as it is, the task is only run once (to be confirmed) so it's unnecessary.

If the task is run once only, then let it fail if the path is not found, and that's enough. No need to store it in a dict (either it's in `$PATH`, or it's provided by `os.getenv("XXX_BIN")`).

gdemonet: Also, how would this behaviour of mutating dicts work in a concurrent execution? Is `doit` responsible for passing the shared state to spawned tasks?

slaperche-scality: > If the task is run once only, then let it fail if the path is not found, and that's enough. No need to store it in a dict (either it's in $PATH, or it's provided by os.getenv("XXX_BIN")).

Hum, fair enough.
This will simplify the code!

> Also, how would this behaviour of mutating dicts work in a concurrent execution? Is doit responsible for passing the shared state to spawned tasks?

Good point.

slaperche-scality:
> /some/path/alias-cmd



> No, that's my point actually (see [#1233 (comment)](https://github.com/scality/metalk8s/pull/1233#discussion_r294186349)):
> 
> [...] 3. `SOME_CMD_BIN` is set to `/some/path/alias-cmd` (note the different binary name), so this task will search for `some-cmd` in `/some/path`, and **will fail**.

If someone tells use `look for /some/path/alias-cmd`, of course we aren't going to find `some-cmd` (because they explicitly told us to look for `alias-cmd`).

buildchain/buildchain/config.py
```diff
@@ -47,11 +47,18 @@
 
 # Name of the command (if in the PATH) or path to the binary.
 
-DOCKER   : str = os.getenv('DOCKER_BIN',   'docker')
+# DOCKER   : str = os.getenv('DOCKER_BIN',   'docker')
```
slaperche-scality: Following @gdemonet recommendation, we will give up the `dict`-based approach and move onto `enum`.

This will simplify the code (we no longer need to use `CmdAction`).

This should become:

```python
# External commands {{{

# Name of the external commands (if in the PATH) or path to the binary.

class ExtCommand(enum.Enum):
    """External commands used by the build chain."""

    GIT      = os.getenv('GIT_BIN',      'git')
    HARDLINK = os.getenv('HARDLINK_BIN', 'hardlink')
    MKISOFS  = os.getenv('MKISOFS_BIN',  'mkisofs')
    SKOPEO   = os.getenv('SKOPEO_BIN',   'skopeo')
    VAGRANT  = os.getenv('VAGRANT_BIN',  'vagrant')

# }}}
```

buildchain/buildchain/deps.py
```diff
@@ -0,0 +1,36 @@
+from doit.exceptions import TaskError
+from pathlib import Path
+from typing import Optional
+import shutil
+from buildchain import config
+
+
+def task_check_for():
```
slaperche-scality: Following @gdemonet  recommendations, this would become

```python
def task_check_for():
    """Check the existence of the external commands."""
    def get_command_location(
        command_name: str, binary_path: Optional[str]
    ) -> Optional[TaskError]:
        """Check the existence of the given command.

        If the user already specified a path for it, check for the existence
        of the binary at the given location. Otherwise, look it up from PATH.

        If the binary is not found, the task fails.
        """
        if binary_path:
            path = Path(binary_path)
            cmd_path = shutil.which(path.name, path=path.parent)
        else:
            cmd_path = shutil.which(command_name)
        if cmd_path is None:
            return TaskError(msg='command {} not found in {}'.format(
                command_name, Path(binary_path).parent
                if binary_path else '$PATH'
            ))
        return None

    for ext_cmd in config.ExtCommand:
        cmd_name = ext_cmd.name.lower()
        cmd_path = ext_cmd.value if ext_cmd.value != cmd_name else None

        def show(name) -> str:
            return '{cmd: <{width}} {name}'.format(
                cmd='CHECK CMD', width=constants.CMD_WIDTH, name=name
            )

        yield {
            'name': cmd_name,
            'title': lambda _, name=cmd_name: show(name),
            'doc': 'Check the presence of the {} command.'.format(cmd_name),
            'actions': [(get_command_location, [cmd_name, cmd_path], {})],
            'file_dep': [],
            'task_dep': [],
            'uptodate': [False],
        }
```

I added a `title` field for a nice output + we  no longer update a `dict`.

buildchain/buildchain/image.py
```diff
@@ -79,10 +79,10 @@ def show() -> str:
             cmd='HARDLINK', width=constants.CMD_WIDTH,
             path=utils.build_relpath(constants.ISO_IMAGE_ROOT)
         )
-
-    task = targets.Target(task_dep=['_image_pull', '_image_build']).basic_task
+    task = targets.Target(task_dep=['check_for:hardlink',
```
slaperche-scality: This can be wrapped as:

```
    task = targets.Target(
        task_dep=['_image_pull', '_image_build', 'check_for:hardlink']
    ).basic_task
```

which is a bit more readable.

buildchain/buildchain/image.py
```diff
@@ -79,10 +79,10 @@ def show() -> str:
...
+                                    '_image_pull', '_image_build']).basic_task
     task['title']   = lambda _: show()
-    task['actions'] = [[config.HARDLINK, '-c', constants.ISO_IMAGE_ROOT]]
+    task['actions'] = [[config.CMD['hardlink'], '-c', constants.ISO_IMAGE_ROOT]]
```
slaperche-scality: Using the enum approach, this would become `config.ExtCommand.HARDLINK.value`

buildchain/buildchain/deps.py
```diff
@@ -0,0 +1,36 @@
+from doit.exceptions import TaskError
```
slaperche-scality: Import should be split in 3 groups:
- standard library
- third-party library
- local module

e.g.:

```
from pathlib import Path
from typing import Optional
import shutil

from doit.exceptions import TaskError

from buildchain import config
from buildchain import constants
from buildchain import utils
```

buildchain/buildchain/iso.py
```diff
@@ -160,7 +160,7 @@ def action(targets: Sequence[str]) -> None:
 def task__iso_build() -> types.TaskDict:
     """Create the ISO from the files in ISO_ROOT."""
     mkisofs = [
-        config.MKISOFS, '-output',  ISO_FILE,
+        config.CMD['mkisofs'], '-output',  ISO_FILE,
```
slaperche-scality: `config.ExtCommand.MKISOFS.value`

buildchain/buildchain/targets/local_image.py
```diff
@@ -169,6 +170,7 @@ def task(self) -> types.TaskDict:
             'title': lambda _: self.show('IMG BUILD'),
             'doc': 'Build {} container image.'.format(self.name),
             'actions': self._build_actions(),
+            'task_dep': ['check_for:skopeo', 'check_for:docker'],
```
slaperche-scality: No need for `check_for:docker` + you shouldn't assign `task_dep` directly: if the user specified its own `task_dep` you'll overwrite them.

I recommend to append to `task_dep`, in the `__init__`:

```diff
@@ -107,6 +107,7 @@ class LocalImage(image.ContainerImage):
         self._save = save_on_disk
         self._build_args = build_args or {}
         kwargs.setdefault('file_dep', []).append(self.dockerfile)
+        kwargs.setdefault('task_dep', []).append('check_for:skopeo')
         super().__init__(
             name=name, version=version, destination=destination, **kwargs
         )
```

buildchain/buildchain/deps.py
slaperche-scality: You need a docstring to describe what your module is doing, e.g.: `"""Tasks to check the dependencies on external commands."""` as the first line of the file

gdemonet:
> > /some/path/alias-cmd
> 
> > No, that's my point actually (see [#1233 (comment)](https://github.com/scality/metalk8s/pull/1233#discussion_r294186349)):
> > [...] 3. `SOME_CMD_BIN` is set to `/some/path/alias-cmd` (note the different binary name), so this task will search for `some-cmd` in `/some/path`, and **will fail**.
> 
> If someone tells use `look for /some/path/alias-cmd`, of course we aren't going to find `some-cmd` (because they explicitly told us to look for `alias-cmd`).

Yes, the code needs to be fixed.

buildchain/buildchain/targets/local_image.py
```diff
@@ -177,6 +179,21 @@ def task(self) -> types.TaskDict:
         })
         return task
 
+    def _build_skopeo_cmd(self) -> List[str]:
+        """Return the command line to execute skopeo cmd."""
+        cmd = [
```
slaperche-scality:
```python
            cmd = [
                config.ExtCommand.SKOPEO.value,
                '--override-os', 'linux', '--insecure-policy',
                'copy', '--format', 'v2s2', '--dest-compress',
            ]
```

buildchain/buildchain/targets/remote_image.py
```diff
@@ -58,6 +60,7 @@ def __init__(
             **kwargs
         )
         self._targets = [self.filepath]
+        self.task_dep.append('check_for:skopeo')
```
slaperche-scality: I would use the same pattern as for `local_image`, e.g:

```diff
+++ b/buildchain/buildchain/targets/remote_image.py
@@ -52,6 +52,7 @@ class RemoteImage(image.ContainerImage):
         self._digest = digest
         self._remote_name = remote_name or name
         self._use_tar = save_as_tar
+        kwargs.setdefault('task_dep', []).append('check_for:skopeo')
         super().__init__(
             name=name, version=version,
             destination=destination,
```

buildchain/buildchain/targets/remote_image.py
```diff
@@ -92,19 +95,23 @@ def task(self) -> types.TaskDict:
...
+                            ],
                 'clean':   [self.clean],
             })
         return task
 
     def _skopeo_copy(self) -> List[str]:
         """Return the command line to execute skopeo copy."""
         cmd = [
-            config.SKOPEO, '--override-os', 'linux', '--insecure-policy',
+            config.CMD['skopeo'], '--override-os', 'linux', '--insecure-policy',
```
slaperche-scality:
```
        cmd = [
            config.ExtCommand.SKOPEO.value,
            '--override-os', 'linux', '--insecure-policy',
            'copy', '--format', 'v2s2'
        ]
```

buildchain/buildchain/deps.py
```diff
@@ -0,0 +1,36 @@
...
+        # return None
+    for cmd, binary_path in config.CMD.items():
+        yield {
+            'name': cmd,
+            'doc': 'Check the presence of the {} command.'.format(cmd),
+            'actions': [(get_command_location, [cmd, binary_path], {})],
+            'file_dep': [],
+            'task_dep': [],
+            'uptodate': [False],
+        }
```
slaperche-scality, line 57: You need to add `__all__ = utils.export_only_tasks(__name__)` at the end of the file, to only export the tasks (it's also done for the others, see `iso.py` for instance).

buildchain/buildchain/vagrant.py
```diff
@@ -38,55 +39,62 @@ def mkdir_dot_vagrant() -> None:
...
+    
+def task__vagrant_up_noprov() -> types.TaskDict:
+    """Run `vagrant up` without provisioning a development environment."""
+    # cmd = [config.CMD['vagrant'], 'up']
+    # cmd.extend(map(shlex.quote, config.VAGRANT_UP_ARGS))
+    # cmd.append('--no-provision')
 
     return {
-        'actions': [doit.tools.LongRunning(' '.join(cmd))],
+        'actions': [doit.action.CmdAction(_build_vagrant_cmd, shell=True)],
```
slaperche-scality: Since we no longer use a dict, you can keep the `LongRunning` as it was before.

buildchain/buildchain/vagrant.py
```diff
@@ -38,55 +39,62 @@ def mkdir_dot_vagrant() -> None:
         'targets': [constants.VAGRANT_SSH_KEY_PAIR],
         'uptodate': [True],
     }
-
-def task__vagrant_up_noprov() -> types.TaskDict:
-    """Run `vagrant up` without provisioning a development environment."""
-    cmd = [config.VAGRANT, 'up']
+    
+def _build_vagrant_cmd(self) -> List[str]:
+    cmd = [config.CMD['vagrant'], 'up']
```
slaperche-scality: `s/config.CMD['vagrant']/config.ExtCommand.VAGRANT.value/`, same for the other occurrence in the file.

slaperche-scality (changes requested):
We will follow @gdemonet good recommendation on getting rid of the `dict`: this will simplify our code (no longer need to use `CmdAction`).

Also, I think it would be nice if you PR was split in 3 commits:
- the first to delete `DOCKER   : str = os.getenv('DOCKER_BIN',   'docker')` because it's no longer used since Jordi's PR (should be mentionned in the commit message)
- the second one to introduce the enum and `deps.py`
- the third one to replace the `config.XXX` by `config.ExtCommand.XXX.value` and adding the `task_dep`

buildchain/buildchain/config.py
Ebaneck: I think that the GIT_BIN should not be included here. no??

Ebaneck:
@bert-e reset

buildchain/buildchain/config.py
```diff
@@ -47,7 +47,6 @@
 
 # Name of the command (if in the PATH) or path to the binary.
 
-DOCKER   : str = os.getenv('DOCKER_BIN',   'docker')
```
slaperche-scality, line 50: Comments about the commit message:

The PR that introduced the docker API is #939, not #1191 (which is a work in progress).

`donot` => either `don't` or `do not`.

e.g.:

```
build: remove DOCKER_BIN

Since the introduction #939, we use the Docker API instead of the binary
directly, thus it's no longer used and can be safely removed.
```

buildchain/buildchain/deps.py
```diff
@@ -1,3 +1,7 @@
+
```
slaperche-scality: No need to skip a line here, the docstring can be the first line of the file (IIRC, that's what we do in the other files).

buildchain/buildchain/config.py
```diff
@@ -9,6 +9,7 @@
 
 import os
 import shlex
+import enum
```
slaperche-scality, line 12: Commit message could be more explicit, like:

```
build: add a test to check for external binaries

This commit introduces a new set of tasks to check for the presence of some required binaries.
If a required binary is missing, the task will fail and stop the build with a clear error message (instead of a big stacktrace).
```

Otherwise, the code LGTM.

buildchain/buildchain/image.py
```diff
@@ -80,9 +80,13 @@ def show() -> str:
             path=utils.build_relpath(constants.ISO_IMAGE_ROOT)
         )
 
-    task = targets.Target(task_dep=['_image_pull', '_image_build']).basic_task
+    task = targets.Target(
```
slaperche-scality, line 83: minor: code layout could be improved a little bit

Align closing parenthesis:
```
    task = targets.Target(
        task_dep=['check_for:hardlink', '_image_pull', '_image_build']
    ).basic_task
```

Try to keep the command on a single line:

```
task['actions'] = [
    [config.ExtCommand.HARDLINK.value, '-c', constants.ISO_IMAGE_ROOT]
]
```

buildchain/buildchain/constants.py
```diff
@@ -91,8 +91,8 @@ def git_ref() -> Optional[str]:
 
     try:
         ref : bytes = subprocess.check_output([
-            config.GIT, 'describe', '--always', '--long', '--tags',
-                '--dirty',
+            config.ExtCommand.GIT.value, 'describe', '--always', '--long',
```
slaperche-scality, line 94: about commit message: the reference to the issue is usually put at the end of the commit, not in the title.
Also, we usually prefix the title with the component/module we are working on (in that case `build` or `buildchain`).

e.g.:

```
build: add task dep on binaries checking

For every task that rely on an external command, we add a dependence to the task that check the existence of the required binary.
```

slaperche-scality (changes requested):
Code is pretty good (only have a few minor remarks), most comments are about commit messages.

buildchain/buildchain/image.py
Ebaneck: Great, I learnt we have a vertical ruler of 79 enforced. This is the only reason I reformated this. I am going to keep the single line

Ebaneck:
@bert-e reset

Ebaneck:
@bert-e force_reset

Ebaneck:
@bert-e reset

slaperche-scality (dismissed):

buildchain/buildchain/image.py
```diff
@@ -80,9 +80,13 @@ def show() -> str:
...
+        task_dep=['check_for:hardlink', '_image_pull', '_image_build']
+        ).basic_task
```
gdemonet: This one should be unindented:

```suggestion
    ).basic_task
```

gdemonet (dismissed):
Great work, thanks :) Will be much friendlier for newcomers with this addition!

(just a minor linting comment)

Ebaneck:
@bert-e reset

gdemonet (approved):

slaperche-scality (approved):

Ebaneck:
@bert-e approve

PR #1246, opened 2019-06-18
UI: remove imagePullPolicy: Always

### LINKED ISSUE #1238: UI - remove imagePullPolicy: Always
**Component**: UI

**What happened**:
- ``` imagePullPolicy: Always ```because we want to pull a new image for the same ui version after each modification. It is only for development purpose. Consequently, we do not need ```imagePullPolicy: Always``` anymore since we can launch UI in local.
**What was expected**:

**Steps to reproduce**

carlito-scality:
**Component**:UI

**Context**: 
- Actually we do not need  ``` imagePullPolicy: Always ``` for development anymore since we can launch the UI on locals. Let's put ``` imagePullPolicy: IfNotPresent ``` to be more explicit.
**Summary**:

**Acceptance criteria**: 

---

Closes: https://github.com/scality/metalk8s/issues/1238

### DIFF
--- a/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
+++ b/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
@@ -27,7 +27,6 @@ spec:
       containers:
         - name: metalk8s-ui
           image: {{ build_image_name('metalk8s-ui', '0.2') }}
-          imagePullPolicy: Always
           resources:
             limits:
               memory: 170Mi
--- a/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
+++ b/salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
@@ -27,6 +27,7 @@ spec:
       containers:
         - name: metalk8s-ui
           image: {{ build_image_name('metalk8s-ui', '0.2') }}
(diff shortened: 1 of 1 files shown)


salt/metalk8s/addons/ui/files/metalk8s-ui-deployment.yaml
```diff
@@ -27,7 +27,6 @@ spec:
       containers:
         - name: metalk8s-ui
           image: {{ build_image_name('metalk8s-ui', '0.2') }}
-          imagePullPolicy: Always
```
charlesprost, line 30: Why not putting `IfNotPresent` instead as we do for some other images? Better be explicit than implicit I'd say.

carlito-scality: @TeddyAndrieux @nootal : it is not by default?

TeddyAndrieux: It's the default value but better to be explicit like for other manifest

carlito-scality: ok

charlesprost (dismissed):

carlito-scality:
@bert-e approve

charlesprost (approved):

PR #1257, opened 2019-06-19
Configure salt minion for local mode when mounting iso

### LINKED ISSUE #1256: `products/mounted.sls` should also configure a salt minion local file_roots
**Component**:

'salt'

**Why this is needed**:

For upgrade process we will need to use salt-minion in local mode to configure the new salt-master and repositories (+registry).

**What should be done**:

When we mount or configure the ISOs we should also create on the same salt minion a configuration file for the file_roots and pillar_roots to be able to use this salt-minion with `--local`

Add a state in salt minion directory to configure file_roots and pillar_roots that state should only be included in the `metalk8s.products.mounted`.

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

For upgrade process we will need to use salt-minion in local mode to configure the new salt-master and repositories (+registry).

**Summary**:

Adding a state to configure salt-minion to be able to use local mode after deployment and use this states in the bootstrap script and when mounting/configuring new ISOs

**Acceptance criteria**: 

Be able to use some states with `salt-call --local`
(not all available as the external pillar that need k8s not working in local mode)

---

Fixes: #1256

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -324,6 +324,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2'),
     Path('salt/metalk8s/salt/minion/init.sls'),
     Path('salt/metalk8s/salt/minion/installed.sls'),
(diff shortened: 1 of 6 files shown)


sayf-eddine-scality (approved):

TeddyAndrieux:
@bert-e approve

PR #1261, opened 2019-06-20
Separate `common location` in registry nginx configuration from version specific

### LINKED ISSUE #1150: Duplicate v2 in nginx conf
**Component**:

'salt', 'upgrade', 'deployment'

**What happened**:

We generate one nginx config file per version to serve container images.
But the config file conflict 
```
2019-05-24T08:46:38.17258157Z stderr F 2019/05/24 08:46:38 [emerg] 1#1: duplicate location "/v2" in /etc/nginx/conf.d/99-metalk8s-2.1-registry.inc:2
2019-05-24T08:46:38.17263509Z stderr F nginx: [emerg] duplicate location "/v2" in /etc/nginx/conf.d/99-metalk8s-2.1-registry.inc:2
``` 

**What was expected**:

No conflict

**Steps to reproduce**

Classic install, mount a new iso and just run the state `metalk8s.repo.installed` with the new version.

Have common location in a single file not depending on the version like `default.conf`

TeddyAndrieux:
**Component**:

'salt', 'registry', 'upgrade'

**Context**: 

We generate one nginx config file per version to serve container images.
But the config file conflict.

**Summary**:

Generate 2 separate file, one with constants location registry nginx configuration and one with version specific location registry nginx configuration

**Acceptance criteria**: 

Have 2 version of metalk8s registry served at the same time

---

Fixes: #1150

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -59,6 +59,47 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
         yield from file_tree.execution_plan
 
 
+class CommonStaticContainerRegistry(targets.AtomicTarget):
+    """Generate a nginx configuration to serve common static container
+    registry."""
+    def __init__(
+        self,
+        destination: Path,
+        **kwargs: Any
+    ):
+        """Set target destination.
+
+        Arguments:
+            destination: path to the nginx configuration file to write
+
+        Keyword Arguments:
+            They are passed to `Target` init method.
+        """
+        kwargs['targets'] = [destination]
+        super().__init__(task_name=destination.name, **kwargs)
+
+    @property
+    def task(self) -> types.TaskDict:
+        task = self.basic_task
+        task.update({
+            'title': self._show,
+            'doc': 'Generate the nginx config to serve a common static '
+                   'container registry.',
+            'actions': [self._run],
+        })
+        return task
+
+    @staticmethod
+    def _show(task: types.Task) -> str:
+        """Return a description of the task."""
+        return utils.title_with_target1('NGINX_CFG', task)
+
+    def _run(self) -> None:
+        """Generate the nginx configuration."""
+        with Path(self.targets[0]).open('w', encoding='utf-8') as fp:
+            fp.write(container_registry.CONSTANTS)
+
+
 class StaticContainerRegistry(targets.AtomicTarget):
     """Generate a nginx configuration to serve a static container registry."""
     def __init__(
@@ -105,7 +146,7 @@ class StaticContainerRegistry(targets.AtomicTarget):
         """Generate the nginx configuration."""
         with Path(self.targets[0]).open('w', encoding='utf-8') as fp:
             parts = container_registry.create_config(
-                self._img_root, self._srv_root, self._name_pfx
+                self._img_root, self._srv_root, self._name_pfx, False
             )
             for part in parts:
                 fp.write(part)
@@ -367,6 +408,13 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
         save_as_tar=True,
     ),
 
+    CommonStaticContainerRegistry(
+        destination=Path(
+            constants.ISO_ROOT,
+            'salt/metalk8s/repo/files',
+            '{}-registry-common.inc'.format(config.PROJECT_NAME.lower())
+        )
+    ),
     StaticContainerRegistry(
         root=constants.ISO_IMAGE_ROOT,
         server_root='${}_{}_images'.format(
--- a/buildchain/static-container-registry/static-container-registry.py
+++ b/buildchain/static-container-registry/static-container-registry.py
@@ -88,8 +88,9 @@ def find_images(root):
             yield (name, tag)
 
 
-def create_config(root, server_root, name_prefix):
-    yield CONSTANTS
+def create_config(root, server_root, name_prefix, with_constants=True):
+    if with_constants:
+        yield CONSTANTS
 
     images = {}
     for (name, tag) in find_images(root):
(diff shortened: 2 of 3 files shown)


buildchain/static-container-registry/static-container-registry.py
```diff
@@ -88,8 +88,9 @@ def find_images(root):
             yield (name, tag)
 
 
-def create_config(root, server_root, name_prefix):
-    yield CONSTANTS
+def create_config(root, server_root, name_prefix, with_constants=True):
```
slaperche-scality, line 91: We  probably want to patch upstream as well, no?

buildchain/buildchain/salt_tree.py
```diff
@@ -59,6 +59,46 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
         yield from file_tree.execution_plan
 
 
+class ConstantContainerRegistry(targets.AtomicTarget):
+    """Generate a nginx configuration to serve constants container registry."""
```
slaperche-scality: What is a `constants container registry`?

buildchain/buildchain/salt_tree.py
```diff
@@ -59,6 +59,46 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
...
+    def __init__(
+        self,
+        destination: Path,
+        **kwargs: Any
+    ):
+        """Configure the static-container-registry script.
```
slaperche-scality: Not really, you just specify the output path (which is not used by the script).

slaperche-scality (commented):
(I don't "request changes" because I won't be here until Thursday and I don't want to block the PR)

buildchain/buildchain/salt_tree.py
TeddyAndrieux: It's like `static container registry` ...

buildchain/buildchain/salt_tree.py
TeddyAndrieux: True just copy paste, I will change this one

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Btw if you have better naming I'm ok to change this

buildchain/static-container-registry/static-container-registry.py
TeddyAndrieux: Upstream ? I did not know it was a downloaded script.
But btw I patched this only for our use case, if we want to keep it the same as upstream I can also do some "filtering" directly in the doit task

slaperche-scality: [Upstream](https://github.com/NicolasT/static-container-registry/) is from @NicolasT so you can see with him if he want the change or not.

buildchain/buildchain/salt_tree.py
slaperche-scality: `static container registry` means, I guess, that's it's a read-only registry whose list of images is generated at build time (hence the `static`, opposite to `dynamic` where you could push images on the fly).

NicolasT: Yeah, please don't use the term 'constant' here, that's not the right vocabulary.

gdemonet (commented):
Maybe instead of adding a new `Target` type for handling these ~constant~ common routes/paths for serving a _static_ container registry, you could have changed the existing `StaticContainerRegistry` to include both files?

salt/metalk8s/repo/installed.sls
```diff
@@ -8,6 +8,7 @@
 
 {%- set nginx_confd  = '/var/lib/metalk8s/repositories/conf.d/' %}
 {%- set nginx_default_conf = nginx_confd ~ 'default.conf' %}
+{%- set nginx_const_conf = nginx_confd ~ '99-registry-consts.inc' %}
```
NicolasT: `constants` or something along those lines, not fond of `consts`

buildchain/static-container-registry/static-container-registry.py
NicolasT: I'll take care of 'upstreaming' and then re-syncing our subtree at a later point in time, never mind for now.

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Done

TeddyAndrieux:
> Maybe instead of adding a new `Target` type for handling these ~constant~ common routes/paths for serving a _static_ container registry, you could have changed the existing `StaticContainerRegistry` to include both files?

I tried but I can't specify 2 target for one class, so I can generate it "quietly" but I think we don't want to do that

salt/metalk8s/repo/installed.sls
TeddyAndrieux: I think I will use "common" like @gdemonet suggested, look better no ?

TeddyAndrieux:
@bert-e reset

buildchain/buildchain/salt_tree.py
```diff
@@ -105,7 +146,7 @@ def _run(self) -> None:
         """Generate the nginx configuration."""
         with Path(self.targets[0]).open('w', encoding='utf-8') as fp:
             parts = container_registry.create_config(
-                self._img_root, self._srv_root, self._name_pfx
+                self._img_root, self._srv_root, self._name_pfx, False
```
charlesprost, line 149:
```suggestion
                self._img_root, self._srv_root, self._name_pfx, with_constants=False
```

charlesprost: Easier to understand I guess

charlesprost (approved):

TeddyAndrieux:
@bert-e approve

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Oh yes indeed

PR #1268, opened 2019-06-20
ClusterHealth: Prometheus connection failed management

### LINKED ISSUE #1185: Improve the UI resiliency to Prometheus API failure.

carlito-scality:
**Component**: UI

**Context**: 
Actually, Prometheus failed connection case is not managed. We only display "down" in the cluster health. It is confusing. 
**Summary**:
- Catch the error and display it when the prometheus API is not connected
- Add ApiClient to manage the 3 apis as the same way
- Refactor https://github.com/scality/metalk8s/pull/1273 

**Acceptance criteria**: 

---

Closes: https://github.com/scality/metalk8s/issues/1185


carlito-scality:
@bert-e reset

carlito-scality:
@bert-e reset

carlito-scality:
@bert-e reset

ui/src/translations/en.json
```diff
@@ -51,6 +51,7 @@
   "message": "Message",
   "active_at": "Active Since",
   "cluster_up_and_running": "Everything is up and running",
+  "prometheus_unavailable": "Prometheus unavailable",
```
gdemonet: To be consistent with the French translation:

```suggestion
  "prometheus_unavailable": "Prometheus is unavailable",
```

carlito-scality:
@bert-e reset

ui/src/ducks/app/monitoring.js
```diff
@@ -1,32 +1,35 @@
-import { put, takeEvery, select, call } from 'redux-saga/effects';
+import { put, takeEvery, call } from 'redux-saga/effects';
 import { getAlerts, getClusterStatus } from '../../services/prometheus/api';
+import { intl } from '../../translations/IntlGlobalProvider';
```
gdemonet: You shouldn't translate messages in the Redux state, as it hinders reproducibility. It's a presentational component's responsibility actually, IMO.

carlito-scality:
@bert-e force_reset

ui/src/services/ApiClient.js
```diff
@@ -0,0 +1,84 @@
+import axios from 'axios';
+
+class ApiClient {
```
gdemonet, line 3: :+1: I like that one :)

Is there a way you could reduce the size for definitions of `get`/`post`/...? Maybe something like:

```javascript
  async send(request) {
    const {
        method = 'get',
        endpoint,
        payload = {},
        headers = {},
        params
    } = request;

    try {
        const response = await axios({
        method,
        headers: { ...this.headers, ...headers },
        params,
        url: endpoint,
        data: payload,
        ...this.settings
        });
        return response.data;
    } catch (error) {
        return { error };
    }
  }

  get = (...request) => send({ ...request, method: 'get' })

  // ...
```

ui/src/ducks/app/monitoring.js
```diff
@@ -39,23 +42,40 @@ export const setClusterStatusAction = payload => {
...
+        clusterHealth.statusLabel = intl.translate('cluster_up_and_running');
+      } else {
+        clusterHealth.status = CLUSTER_STATUS_DOWN;
+        clusterHealth.statusLabel = intl.translate('down');
+      }
+    }
+  } else {
+    clusterHealth.status = CLUSTER_STATUS_ERROR;
+    if (resultAlerts.error.response) {
+      clusterHealth.statusLabel = `Prometheus - ${
```
gdemonet: Since that one is not translated (and probably cannot be), I think it should be handled differently (not use the same state key).

ui/src/ducks/app/monitoring.js
```diff
@@ -39,23 +42,40 @@ export const setClusterStatusAction = payload => {
...
 export function* fetchClusterStatus() {
-  const api = yield select(state => state.config.api);
-  const result = yield call(getClusterStatus, api.url_prometheus);
-  if (!result.error) {
-    yield put(setClusterStatusAction(result.data.data.result));
+  const clusterHealth = {
+    status: '',
+    label: ''
+  };
+  const resultAlerts = yield call(getAlerts); // Check if Prometheus API is available
```
gdemonet: Why should we rely on the alert list to check if Prom API is available? I think there's a simpler `/health` endpoint for this purpose, which you could poll every `N` seconds and store the result in Redux state.

Otherwise, a good pattern is to try to get a result, and react accordingly if it fails:

> **It's easier to ask forgiveness than it is to get permission.**
> __ Grace Hopper

In any case, I'm not sure merging the two sagas is a good idea.

carlito-scality: We do not have any endpoint like /health :(

ui/src/ducks/app/monitoring.js
carlito-scality: Good point. By the way core-ui components does not support i18n yet such as Notifications so we have to do it in  sagas

ui/src/services/ApiClient.js
carlito-scality: Let's try it

carlito-scality:
@bert-e approve

ui/src/ducks/app/monitoring.js
gdemonet: Wrap them in _container_ components then ;)

IMHO, the Redux state is not about user-facing data, it's about the application state. Your components are responsible for displaying meaningful information, derived from this application state. So, instead of storing status labels in there, just compute it from the status of components (UP vs. DOWN), and translate it on-the-fly.

ui/src/ducks/app/monitoring.js
gdemonet: There is one, it's used for liveness/readiness probes. However, as I said in my comment, the EAFP pattern should be the way to go:

1. Keep two distinct sagas, one for health computation, the other for fetching the list of alerts: this means that they can run on different polling schedules, and are tested independently
2. For each of them, try to perform the API call:
  a. If it passes, you got the result, and are happy (just as before)
  b. If it fails, you have error handling in both sagas (so maybe you get the query but somehow the alerts call fails due to wrong authorization..)
  c. If the API is unavailable / unresponsive, you can catch it differently, and write a specific state key (e.g. `app.monitoring.api_available`, set to `False`) that your components can rely on to show something different (between having a query-related error and a broken API)

ui/src/services/ApiClient.js
MonPote: Issue created #1338

carlito-scality:
@bert-e reset

MonPote (dismissed):

gdemonet (dismissed):

MonPote (approved):

PR #1254, opened 2019-06-18
1076: Test monitoring

alexandre-allard:
**Component**: Monitoring, Tests

**Acceptance criteria**: Tests are all green :) if tests in the CI are OK, it should be OK.

---
Closes: #1076

PR #1273, opened 2019-06-21
Rework compute of cluster health

### LINKED ISSUE #1248: Rework compute of cluster status
**Component**: ui

**What happened**:
The computation of cluster status is not correct on the UI.
Right now if at least 1 apiserver, kube-scheduler or kube-controller-manager is up we display the `Everything is up and running`.

**What was expected**:
It should be when at least 1 job of apiserver, kube-scheduler and kube-controller-manager is up that we display `Everything is up and running`.

**Steps to reproduce**

2 way : 
- Create a single query that do that
- Create 3 queries `sum(up{job=~"apiserver"} == 0)` & `sum(up{job=~"kube-scheduler"} == 0)` & `sum(up{job=~"kube-controller-manager"} == 0)` and check that they are good.

MonPote:
**Component**: ui

**Context**: 
#1254 fix issue that need some changes in the compute of cluster status

**Summary**:
Cluster status will now display `Everything is up and running` when there is a least one job up for apiserver, kube-scheduler and kube-controller-manager.

**Acceptance criteria**: 

To Test this, you will need to rebase the fix from #1254 
Check that cluster status display `Everythign is up and running`.

---

Closes: #1248

### DIFF
--- a/ui/src/ducks/app/monitoring.js
+++ b/ui/src/ducks/app/monitoring.js
@@ -1,15 +1,26 @@
 import { put, takeEvery, select, call } from 'redux-saga/effects';
-import { getAlerts, getClusterStatus } from '../../services/prometheus/api';
+import { getAlerts, queryPrometheus } from '../../services/prometheus/api';
 
 const FETCH_ALERTS = 'FETCH_ALERTS';
 const SET_ALERTS = 'SET_ALERTS';
 
 const FETCH_CLUSTER_STATUS = 'FETCH_CLUSTER_STATUS';
 const SET_CLUSTER_STATUS = 'SET_CLUSTER_STATUS';
+const SET_APISERVER_STATUS = 'SET_APISERVER_STATUS';
+const SET_KUBESCHEDULER_STATUS = 'SET_KUBESCHEDULER_STATUS';
+const SET_KUBECONTROLLER_MANAGER_STATUS = 'SET_KUBECONTROLLER_MANAGER_STATUS';
 
 const defaultState = {
   alertList: [],
-  clusterStatus: []
+  clusterStatus: [],
+
+  cluster: {
+    // Will be change with Carlito PR
+    status: false,
+    apiServerStatus: [],
+    kubeSchedulerStatus: [],
+    kubeControllerManagerStatus: []
+  }
 };
 
 export default function(state = defaultState, action = {}) {
@@ -17,7 +28,10 @@ export default function(state = defaultState, action = {}) {
     case SET_ALERTS:
       return { ...state, alertList: action.payload };
     case SET_CLUSTER_STATUS:
-      return { ...state, clusterStatus: action.payload };
+      return {
+        ...state,
+        cluster: { ...state.cluster, status: action.payload }
+      };
     default:
       return state;
   }
@@ -39,6 +53,18 @@ export const setClusterStatusAction = payload => {
   return { type: SET_CLUSTER_STATUS, payload };
 };
 
+export const setApiServerStatusAction = payload => {
+  return { type: SET_APISERVER_STATUS, payload };
+};
+
+export const setKubeSchedulerStatusAction = payload => {
+  return { type: SET_KUBESCHEDULER_STATUS, payload };
+};
+
+export const setKubeControllerManagerStatus = payload => {
+  return { type: SET_KUBECONTROLLER_MANAGER_STATUS, payload };
+};
+
 export function* fetchAlerts() {
   const api = yield select(state => state.config.api);
   const result = yield call(getAlerts, api.url_prometheus);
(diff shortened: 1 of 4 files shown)


MonPote:
@bert-e after_pull_request #1254

ui/src/ducks/app/monitoring.js
```diff
@@ -49,9 +75,64 @@ export function* fetchAlerts() {
 
 export function* fetchClusterStatus() {
```
carlito-scality, line 93: you should add unit tests please.

carlito-scality (changes requested):

ui/src/ducks/app/monitoring.js
```diff
@@ -49,9 +75,64 @@ export function* fetchAlerts() {
...
   const api = yield select(state => state.config.api);
-  const result = yield call(getClusterStatus, api.url_prometheus);
-  if (!result.error) {
-    yield put(setClusterStatusAction(result.data.data.result));
+
+  const apiserverQuery = 'sum(up{job="apiserver"})';
+  const kubeSchedulerQuery = 'sum(up{job="kube-scheduler"})';
+  const kubeControllerManagerQuery = 'sum(up{job="kube-controller-manager"})';
+
+  const apiserver = yield call(
```
carlito-scality: All the 3 query can be executed at the same time with ``` yield all ```

carlito-scality (dismissed):
LGTM

ui/src/ducks/app/monitoring.js
```diff
@@ -49,9 +92,61 @@ export function* fetchAlerts() {
...
+
+  const kubeScheduler = results[1];
+  const kubeSchedulerValue = kubeScheduler.data.data.result[0].value;
+
+  const kubeControllerManager = results[2];
+  const kubeControllerManagerValue =
+    kubeControllerManager.data.data.result[0].value;
+
+  if (apiserver && apiserver.data && apiserver.data.status === 'success') {
+    const apiServerValue = apiserver.data.data.result[0].value;
```
carlito-scality: to be removed since already exist

MonPote:
@bert-e reset

ui/src/ducks/app/monitoring.js
```diff
@@ -49,9 +92,60 @@ export function* fetchAlerts() {
...
+  const kubeControllerManagerQuery = 'sum(up{job="kube-controller-manager"})';
+
+  const results = yield all([
+    call(queryPrometheus, api.url_prometheus, apiserverQuery),
+    call(queryPrometheus, api.url_prometheus, kubeSchedulerQuery),
+    call(queryPrometheus, api.url_prometheus, kubeControllerManagerQuery)
+  ]);
+
+  const apiserver = results[0];
+  const apiServerValue = apiserver.data.data.result[0].value;
```
gdemonet, line 107: Maybe this pattern for checking status before putting the value could be shared (you have it three times already):

```javascript
export async function queryPrometheus(api, query) {
  try {
    const result = await axios.get(api + '/api/v1/query?query=' + query);
    return { status: result.data.status, value: result.data.data.result[0].value }
  } catch (error) {
    return { error };
  }
}
```

If you need special handling of missing `.data` / or empty `.value`, you could do it in this method as well, maybe returning an error.

ui/src/ducks/app/monitoring.js
```diff
@@ -39,6 +70,18 @@ export const setClusterStatusAction = payload => {
...
 
+export const setApiServerStatusAction = payload => {
+  return { type: SET_APISERVER_STATUS, payload };
+};
+
+export const setKubeSchedulerStatusAction = payload => {
+  return { type: SET_KUBESCHEDULER_STATUS, payload };
+};
+
+export const setKubeControllerManagerStatus = payload => {
```
gdemonet, line 81:
```suggestion
export const setKubeControllerManagerStatusAction = payload => {
```

... I guess.

ui/src/ducks/app/monitoring.js
```diff
@@ -1,23 +1,54 @@
...
+        cluster: {
+          ...state.cluster,
+          kubeControllerManagerStatus: action.payload
+        }
+      };
     case SET_CLUSTER_STATUS:
-      return { ...state, clusterStatus: action.payload };
+      return {
+        ...state,
+        cluster: { ...state.cluster, status: action.payload }
```
gdemonet, line 50: Do you really need this one in the state? If it's directly computable from the other `<component>Status` from this state, maybe we don't need it? (not to mention it's not a good pattern, may induce desynchronized values)

Thinking of it, that's actually a good candidate for a memoized selector.

MonPote: `cluster.status` will be change in the next PR #1268
The idea will be to set the status translated value directly from saga.

ui/src/ducks/app/monitoring.js
carlito-scality: good point.
=> See PR #1268

ui/src/ducks/app/monitoring.js
gdemonet: Kindly disagree: see https://github.com/scality/metalk8s/pull/1268#discussion_r298224343

carlito-scality:
merged in https://github.com/scality/metalk8s/issues/1185

PR #1295, opened 2019-06-25
Does not drain when upgrading a single node cluster

### LINKED ISSUE #1291: Skip draining when upgrading single-node
**Component**:

'salt', 'upgrade'

**What happened**:

When upgrading we drain node one by one, When we have only one node we will stop pods like `alertmanager` and `prometheus` and then prometheus IP needed by UI no longer available so ui deployed state will fail.

**What was expected**:

Working single-node upgrade

**Steps to reproduce**

Try to upgrade a single node cluster

Skip draining when we have only one node as the pod will not be scheduled somewhere else.

TeddyAndrieux:
**Component**:

'salt', 'upgrade'

**Context**: 

When upgrading we drain node one by one, When we have only one node we will stop pods like alertmanager and prometheus and then prometheus IP needed by UI no longer available so ui deployed state will fail.

**Summary**:

Skip draining when we have only one node as the pods will not be scheduled somewhere else.

---

Fixes: #1291

### DIFF
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -52,6 +52,8 @@ Cordon the node:
     - kubeconfig: {{ kubeconfig }}
     - context: {{ context }}
 
+{%- if not pillar.orchestrate.get('skip_draining', True) %}
+
 Drain the node:
   metalk8s_drain.node_drained:
     - name: {{ node_name }}
@@ -62,6 +64,10 @@ Drain the node:
     - context: {{ context }}
     - require:
       - metalk8s_cordon: Cordon the node
+    - require_in:
+      - salt: Run the highstate
+
+{%- endif %}
 
 Run the highstate:
   salt.state:
@@ -70,7 +76,7 @@ Run the highstate:
     - require:
       - salt: Set grains
       - salt: Refresh the mine
-      - metalk8s_drain: Drain the node
+      - metalk8s_cordon: Cordon the node
 
 Uncordon the node:
   metalk8s_cordon.node_uncordoned:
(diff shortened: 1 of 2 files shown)


gdemonet:
Draining a one-node cluster can be problematic, agreed, though I'm not sure it's a valid design to remove it entirely. Maybe the formula for deploying the UI could be more resilient to such scenario?

TeddyAndrieux:
> Draining a one-node cluster can be problematic, agreed, though I'm not sure it's a valid design to remove it entirely. Maybe the formula for deploying the UI could be more resilient to such scenario?

Hmm but if the UI need it we have no real solution for that.

Maybe a better error message but .... that not fix the issue

OR 

We can separate deployment of the UI from other stuff... Maybe a better solution btw (I will do some testing)

TeddyAndrieux:
> > Draining a one-node cluster can be problematic, agreed, though I'm not sure it's a valid design to remove it entirely. Maybe the formula for deploying the UI could be more resilient to such scenario?
> 
> Hmm but if the UI need it we have no real solution for that.
> 
> Maybe a better error message but .... that not fix the issue
> 
> OR
> 
> We can separate deployment of the UI from other stuff... Maybe a better solution btw (I will do some testing)

Ok for me this will be fixed with ingress so for the moment just skip draining when one node it's good enough IMO

TeddyAndrieux:
> Draining a one-node cluster can be problematic, agreed, though I'm not sure it's a valid design to remove it entirely. Maybe the formula for deploying the UI could be more resilient to such scenario?

And also btw for me the UI should not have the prometheus endpoint directly hardcoded in the deployment as the EP can change, the UI should retrieve itself the IP/port using the apiserver

gdemonet (approved):

TeddyAndrieux:
@bert-e approve

Issue #1281, opened 2019-06-24
Upgrading salt-master to dev2.1 require kubelet 1.12


TeddyAndrieux:
**Component**:

'salt', 'upgrade'

**What happened**:

In dev2.1 we use `spec.shareProcessNamespace: true` which is not enabled by default in kubelet 1.11 but only in 1.12 so when we upgrade salt-master without upgrading kubelet first we got an error
```
bootstrap.novalocal kubelet[16488]: E0621 17:06:04.057277   16488 file.go:182] Can't process manifest file "/etc/kubernetes/manifests/salt-master.yaml": invalid pod: [spec.securityContext.shareProcessNamespace: Forbidden: Process Namespace Sharing is disabled by PodShareProcessNamespace feature-gate]
```

**What was expected**:

Have a working salt-master.

**Steps to reproduce**

Install classic 2.0 and try to upgrade to 2.1

**Resolution proposal** (optional):

**TBD**
Some proposal:
- Upgrade kubelet before upgrading salt-master pod.
Easiest solution but not recommended `kubelet instances on all nodes are at version 1.n or 1.(n-1) (this ensures they are not newer than the existing API server version, and are within 2 minor versions of the new API server version)` https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-component-upgrade-order
- Enable ` --feature-gates=PodShareProcessNamespace=true` in kubelet.
Simple but maybe not a good idea as share-process-namespace is in alpha version in 1.11  (https://v1-11.docs.kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/)
- Make the pod template dependant on the version of kubelet installed

TeddyAndrieux:
@NicolasT any suggestion on this one ?

PR #1325, opened 2019-06-27
Enable podshareprocessnamespace for kubelet 1.11

TeddyAndrieux:
**Component**:

'salt', 'upgrade'

**Context**: 

In dev2.1 we use `spec.shareProcessNamespace: true` which is not enabled by default in kubelet 1.11 but only in 1.12 so when we upgrade salt-master without upgrading kubelet first we got an error

**Summary**:

In salt-master installed state enable feature gate `PodShareProcessNamespace` if we are in kubelet 1.11.

**Acceptance criteria**: 

Salt-master manifest from development/2.1 works with kubelet 1.11

---

Fixes: #1281

### DIFF
--- a/salt/metalk8s/salt/master/installed.sls
+++ b/salt/metalk8s/salt/master/installed.sls
@@ -19,6 +19,27 @@ Create salt master directories:
       - /var/cache/salt
       - /var/run/salt
 
+{%- if salt.pkg.version('kubelet').startswith('1.11.') %}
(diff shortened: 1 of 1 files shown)


TeddyAndrieux:
@bert-e reset

sayf-eddine-scality (approved):

TeddyAndrieux:
@bert-e approve

PR #1336, opened 2019-07-01
build/deps: fix the type of `path` for `which`

### LINKED ISSUE #1335: AttributeError: 'PosixPath' object has no attribute 'split' in buildchain/buildchain/deps.py
**Component**:

buildchain

**What happened**:

When trying to lookup a binary in a custom location, we die with the following error:

```
TaskError - taskid:check_for:skopeo
PythonAction Error
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.7/site-packages/doit/action.py", line 424, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/home/sylvain/dev/metalk8s/buildchain/buildchain/deps.py", line 30, in get_command_location
    cmd_path = shutil.which(path.name, path=path.parent)
  File "/usr/lib/python3.7/shutil.py", line 1144, in which
    path = path.split(os.pathsep)
AttributeError: 'PosixPath' object has no attribute 'split'
```

The problem is that we're passing a `Path` object where `shutil.which` expects a `str`.

**What was expected**:

The binary is found and the build can proceed.

**Steps to reproduce**

Create a `.env` file at the root of the repository with the following content:

```
export SKOPEO_BIN=/some/path/to/skopeo
```

and then run `./doit.sh`.

**Resolution proposal**:

Simply convert the `Path` object to `str` before passing it to `shutil.which`.

slaperche-scality:
**Component**:

buildchain

**Context**: 

The buildchain can be configured to use binaries in location out of `PATH`, but currently this cause the code to die with an exception.
This bug stems from the fact that `shutil.which` doesn't work with `Path` object.

**Summary**:

We convert the `path` argument to `str` before passing it to `shutil.which`.

**Acceptance criteria**: 

You can build MetalK8s with binaries in custom location (i.e. defined in `.env`).

---

Closes: #1335

### DIFF
--- a/buildchain/buildchain/deps.py
+++ b/buildchain/buildchain/deps.py
@@ -27,7 +27,7 @@ def task_check_for() -> Iterator[types.TaskDict]:
         """
         if binary_path:
             path = Path(binary_path)
-            cmd_path = shutil.which(path.name, path=path.parent)
+            cmd_path = shutil.which(path.name, path=str(path.parent))
         else:
             cmd_path = shutil.which(command_name)
         if cmd_path is None:


ChengYanJin (approved):
It works well! Thanks for fixing!

gdemonet (approved):

slaperche-scality:
@bert-e approve

PR #1342, opened 2019-07-02
Do a `yum clean all` when configuring the repositories

### LINKED ISSUE #1298: During upgrade yum cache is not uptodate
**Component**:

'salt', 'upgrade'

**What happened**:

When we try to upgrade kubelet the new version of kubelet is not available in the yum cache. Just removing `/var/cache/yum` fix the problem
```
Error occurred installing package(s). Additional info follows:
                                                  
errors:
     - Running scope as unit run-30943.scope.
       Loaded plugins: fastestmirror, versionlock
       Loading mirror speeds from cached hostfile
       No package kubelet-1.12.6-0 available.
       Error: Nothing to do
```

**What was expected**:

Kubelet get upgraded

**Steps to reproduce**

Make a new iso available, reconfigure repo on a host and try to upgrade kubelet

Maybe one of those is enough:
- refreshing db before upgrading using refresh_db https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.yumpkg.html#salt.modules.yumpkg.refresh_db
- set `refresh: True` in the pkg.installed macro https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html#salt.states.pkg.installed
Too be investigate

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

We use the same repositories name for all MetalK8s versions so we need to clean all the yum cache to be able to have the right list of packages

**Summary**:

When one of the repositories changes do a `yum clean all` + refresh_db

---

Fixes: #1298

### DIFF
--- a/salt/metalk8s/orchestrate/upgrade/init.sls
+++ b/salt/metalk8s/orchestrate/upgrade/init.sls
@@ -30,11 +30,6 @@ Upgrade etcd cluster:
   {%- set kubeconfig = "/etc/kubernetes/admin.conf" %}
   {%- set context = "kubernetes-admin@kubernetes" %}
 
-Refresh the package repository for {{ node }}:
-  salt.function:
-    - tgt: {{ node }}
-    - name: pkg.refresh_db
-
 Set node {{ node }} version to {{ dest_version }}:
   metalk8s_kubernetes.node_label_present:
     - name: metalk8s.scality.com/version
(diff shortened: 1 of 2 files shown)


slaperche-scality (approved):

TeddyAndrieux:
@bert-e approve

PR #1346, opened 2019-07-02
build/vagrant: fix `uptodate` for `vagrant_snapshot`

### LINKED ISSUE #1345: `_vagrant_snapshot` is not always executed
**Component**:

buildchain

**What happened**:

When running `vagrant_up`, even with no existing snapshot, the task `_vagrant_snapshot` is not always executed.

**What was expected**:

If the snapshot doesn't exist, it should be created.

**Steps to reproduce**

Starting from a clean state (i.e run `vagrant destroy` and `./doit.sh forget _vagrant_snapshot`):
- `vagrant snapshot list`: no snapshot exist
- `./doit.sh _vagrant_snapshot`: the task is executed (you should see `.  _vagrant_snapshot`) and the snapshot is created (you can check with `vagrant snapshot list`)
- `vagrant destroy bootstrap`: destroy the VM (and its associated snapshot, you can check with `vagrant snapshot list`)
- `./doit.sh _vagrant_snapshot`: the task is **NOT** executed (you should see `--  _vagrant_snapshot`) and the snapshot is not recreated (you can check with `vagrant snapshot list`)

**Resolution proposal**:

The problem comes from the use of `doit.tools.run_once` for `uptodate`: this helper ensures that we execute the task `_vagrant_snapshot` once and only once (an entry is stored in the `doit`'s DB to keep track of that).

Consequently, you'll never run `_vagrant_snapshot` twice, even if the snapshot is destroyed in the future (by destroying the VM, deleting the snapshot manually, …).

A good way to fix this issue would be to provide a custom `uptodate` definition that would check for the existence of the snapshot to decide if we need to execute `_vagrant_snapshot` or not.

slaperche-scality:
**Component**:

buildchain

**Context**: 

When running vagrant_up, even with no existing snapshot, the task _vagrant_snapshot is not always executed.

**Summary**:

We were using `doit.tools.run_once` which is not what we want here (the snapshot may have been deleted since its first creation).

Now we rely on the output of `vagrant snapshot list` to decide if we need to create a snapshot or not.

**Acceptance criteria**: 

Executing `./doit.sh _vagrant_snapshot` should create a snapshot if it doesn't already exists (and do nothing if it already exists).

---

Closes:  #1345

### DIFF
--- a/buildchain/buildchain/vagrant.py
+++ b/buildchain/buildchain/vagrant.py
@@ -11,6 +11,7 @@ environment.
 
 
 import shlex
+import subprocess
 
 import doit  # type: ignore
 
@@ -54,6 +55,12 @@ def task__vagrant_up_noprov() -> types.TaskDict:
 
 def task__vagrant_snapshot() -> types.TaskDict:
     """Snapshot development environment."""
+    def check_snapshot_existence() -> bool:
+        snapshots : bytes = subprocess.check_output([
+            config.ExtCommand.VAGRANT.value, 'snapshot', 'list'
+        ])
+        return config.VAGRANT_SNAPSHOT_NAME in snapshots.decode('utf-8')
+
     vagranthalt = [config.ExtCommand.VAGRANT.value, 'halt', 'bootstrap']
     vagrantsnap = [config.ExtCommand.VAGRANT.value, 'snapshot', 'save',
                    '--force', 'bootstrap', config.VAGRANT_SNAPSHOT_NAME]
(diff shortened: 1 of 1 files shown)


slaperche-scality:
@bert-e reset

jbertran (approved):
Run 1:

```
==> bootstrap: Snapshotting the machine as 'bootstrap.autosnapshot'...
==> bootstrap: Snapshot saved! You can restore the snapshot at any time by
==> bootstrap: using `vagrant snapshot restore`. You can delete it using
==> bootstrap: `vagrant snapshot delete`.
```

Run 2:

```
-- MKDIR        _build/packages
-- _vagrant_snapshot
-- MKDIR        _build/root
```

Run 3 (after destroy):

```
.  _vagrant_snapshot
==> bootstrap: Attempting graceful shutdown of VM...
==> bootstrap: Snapshotting the machine as 'bootstrap.autosnapshot'...
==> bootstrap: Snapshot saved! You can restore the snapshot at any time by
==> bootstrap: using `vagrant snapshot restore`. You can delete it using
==> bootstrap: `vagrant snapshot delete`.
```

LGTM!

jbertran:
`requirements.sh` failed on single and multi node, rebuild triggered.

slaperche-scality:
@bert-e approve

PR #1347, opened 2019-07-03
Separate config generation from manifest for repositories

### LINKED ISSUE #1344: Repositories separate config generation from manifest deployment
**Component**:

'salt', 'kubernetes'

**Why this is needed**:

To make new iso version available we need to reconfigure repositories without upgrading the static pod, so we need to separate config generation from manifest deployment

**What should be done**:

Separate config generation from manifest generation in repositories

Create a states to generate all the configuration and include it in `installed.sls`

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Context**: 

To make new iso version available we need to reconfigure repositories without upgrading the static pod, so we need to separate config generation from manifest deployment

**Summary**:

Separate config generation from manifest generation in repo states

---

Fixes: #1344

### DIFF
--- a/salt/_states/metalk8s.py
+++ b/salt/_states/metalk8s.py
@@ -10,16 +10,48 @@ def __virtual__():
     return __virtualname__
 
 
-def static_pod_managed(name, source, config_files=None, context=None):
+def static_pod_managed(name,
+                       source,
+                       config_files=None,
+                       config_files_opt=None,
+                       context=None,
+                       **kwargs):
     """Simple helper to edit a static Pod manifest if configuration changes.
 
     Expects the template to use the `config_digest` variable and store it in
     the `metadata.annotations` section, with the key
     `metalk8s.scality.com/config-digest`.
+
+    name:
+        Path to the static pod manifest.
+
+    source:
+        Source file used to render the manifest.
+
+    config_files:
+        List of file to use to generate a digest store in `config_digest` in
+        the source template.
+
+    config_files_opt:
+        Same as config_files but these files are optional (ignored if the file
+        does not exists).
+
+    context:
+        Context to use to render the source template.
+
+    kwargs:
+        Any arguments supported by `file.managed` are supported.
     """
+    if not config_files:
+        config_files = []
+
+    for config_file in config_files_opt or []:
+        if __salt__["file.file_exists"](config_file):
+            config_files.append(config_file)
+
     config_file_digests = [
         __salt__["hashutil.digest_file"](config_file, checksum="sha256")
-        for config_file in config_files or []
+        for config_file in config_files
     ]
     config_digest = __salt__["hashutil.md5_digest"](
         "-".join(config_file_digests)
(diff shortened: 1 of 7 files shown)


salt/_states/metalk8s.py
```diff
@@ -10,16 +10,25 @@ def __virtual__():
     return __virtualname__
 
 
-def static_pod_managed(name, source, config_files=None, context=None):
+def static_pod_managed(name, source,
```
jbertran: minor: indentation?

salt/_states/metalk8s.py
```diff
@@ -10,16 +10,25 @@ def __virtual__():
...
+                       config_files=None, config_files_opt=None, context=None,
+                       **kwargs):
     """Simple helper to edit a static Pod manifest if configuration changes.
 
     Expects the template to use the `config_digest` variable and store it in
     the `metadata.annotations` section, with the key
     `metalk8s.scality.com/config-digest`.
```
jbertran, line 23: A more verbose docstring with arguments would be helpful here.

salt/_states/metalk8s.py
```diff
@@ -10,16 +10,25 @@ def __virtual__():
     return __virtualname__
 
 
-def static_pod_managed(name, source, config_files=None, context=None):
+def static_pod_managed(name, source,
+                       config_files=None, config_files_opt=None, context=None,
```
jbertran: You could use 2 read-only arguments and create a third list out of the contents of the 2 first - that would allow you to specify empty list as default.

salt/metalk8s/repo/installed.sls
```diff
@@ -70,9 +19,16 @@ Install repositories manifest:
     - name: /etc/kubernetes/manifests/repositories.yaml
     - source: salt://{{ slspath }}/files/repositories-manifest.yaml.j2
     - config_files:
-      - {{ nginx_default_conf }}
-      - {{ nginx_registry_path }}
-      - {{ nginx_registry_config_path }}
+      - {{ salt.file.join(repo.config.directory, repo.config.default) }}
```
jbertran, line 22: minor: could be a jinja for loop over `repo` keys? That might be ugly and too long though.

jbertran (commented):
LGTM, minor comments.

salt/metalk8s/repo/installed.sls
TeddyAndrieux: No I think we don't want to do that because it's something merged with the pillar so we don't want to take all of them

salt/_states/metalk8s.py
TeddyAndrieux: Hmm, you already can provide empty list, I don't really get the point

salt/_states/metalk8s.py
TeddyAndrieux: Indeed, I will add one

salt/_states/metalk8s.py
jbertran: The intent is to have empty lists in the function definition, for clarity. That way you don't need the `None` checks that are below :)

TeddyAndrieux: I think I will keep the `None` because here we handle if the user give `None` as input which is the case when for example we do some jinja with a for loop in yaml file.
```
toto:
{%- for i in [] %}
  - {{ i }}
{%- endfor %}
```
"toto" will be `None`

TeddyAndrieux:
@bert-e reset

salt/_states/metalk8s.py
jbertran: Indeed. Works for me then!

jbertran (approved):

TeddyAndrieux:
@bert-e approve

salt/_states/metalk8s.py
gdemonet: Let's not do that kind of thing anyway, passing mutable defaults to functions is not recommended in Python (even if your function is supposed to be pure).

PR #1365, opened 2019-07-04
Improvement/950 add sosreport collecting

sayf-eddine-scality:
**Component**: salt, tests

**Context**:  We need to collect logs and config files to ease debugging and reproducing all kind of issues

**Summary**:
Sosreport can now be installed with two plugins that help recover all trace of logs from the platform
**Acceptance criteria**: 
running `salt-call state.sls metalk8s.internal.sreport saltenv=metalk8s-2.0.0-dev` shoud install both sosreport and our custom plugins then you need to run
`sosreport --all-logs -o metalk8s -o containerd -kmetalk8s.podlogs=True -kcontainerd.all=True -kcontainerd.logs=True --batch --tmp-dir <chose a dir>` to generate the report

---

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -151,6 +151,19 @@ CALICO_CNI_PLUGIN_VERSION : str = '3.8.0'
 # Packages per repository.
 PACKAGES : Dict[str, Tuple[targets.Package, ...]] = {
     'scality': (
+        # SOS report custom plugins.
+        targets.Package(
+            basename='_build_packages',
+            name='metalk8s-sosreport',
+            version=constants.SHORT_VERSION,
+            build_id=1,
+            sources=[
+                Path('metalk8s.py'),
+                Path('containerd.py'),
+            ],
+            builder=BUILDER,
+            task_dep=['_package_mkdir_root', '_build_container'],
+        ),
         # Calico Container Network Interface Plugin.
         targets.Package(
             basename='_build_packages',
--- a/buildchain/buildchain/targets/package.py
+++ b/buildchain/buildchain/targets/package.py
@@ -27,6 +27,7 @@ Overview;
 import os
 import operator
 import re
+import shutil
 import urllib.parse
 import urllib.request
 from pathlib import Path
@@ -188,7 +189,7 @@ class Package(base.CompositeTarget):
         targets = [self.srcdir]
         targets.extend(self.sources)
         actions = directory.Mkdir(directory=self.srcdir).task['actions']
-        actions.append(self._download_sources)
+        actions.append(self._get_sources)
         task = self.basic_task
         task.update({
             'name': 'pkg_get_source',
@@ -235,12 +236,16 @@ class Package(base.CompositeTarget):
                                                self.MKDIR_TASK_NAME))
         return task
 
-    def _download_sources(self) -> None:
-        """Return a list of actions to download the source files."""
+    def _get_sources(self) -> None:
+        """Gather the package resources."""
         for srcfile, url in self._get_source_files_urls().items():
-            with urllib.request.urlopen(url) as conn:
-                with open(srcfile, 'wb') as fp:
-                    fp.write(conn.read())
+            if urllib.parse.urlparse(url).scheme:
+                with urllib.request.urlopen(url) as conn:
+                    with open(srcfile, 'wb') as fp:
+                        fp.write(conn.read())
+            else:
+                url = os.path.join(constants.ROOT/'packages', url)
+                shutil.copyfile(url, srcfile)
 
     def _get_source_files_urls(self) -> Dict[Path, str]:
         """Extract source file URLs from .meta file."""
--- a/buildchain/buildchain/targets/package.py
+++ b/buildchain/buildchain/targets/package.py
@@ -209,6 +209,7 @@ class Package(base.CompositeTarget):
             'SPEC': self.spec.name,
             'SRPM': self.srpm.name,
             'SOURCES': ' '.join(source.name for source in self.sources),
+            'VERSION': self.version,
         }
 
         buildsrpm_callable = docker_command.DockerRun(
--- a/packages/metalk8s-sosreport.spec
+++ b/packages/metalk8s-sosreport.spec
@@ -1,0 +1,38 @@
+Name:		metalk8s-sosreport
+Version:	_VERSION_
+Release:	1%{?dist}
+Summary: 	Metalk8s SOS report custom plugins
+
+
+Requires: python >= 2.6, python < 2.8
+Requires: sos >= 3.1
+# NameError on FileNotFoundError in sos 3.5 python2.7
+Conflicts: sos = 3.5
+
+ExclusiveArch:  x86_64
+ExclusiveOS:    Linux
+
+License:        GPL+
+Source0:        metalk8s-sosreport/src/metalk8s.py
+Source1:        metalk8s-sosreport/src/containerd.py
+
+%description
+%{Summary}
+
+%install
+install -m 755 -d %{buildroot}/%{python_sitelib}/sos/plugins
+install -p -m 755 %{_topdir}/SOURCES/metalk8s.py %{buildroot}/%{python_sitelib}/sos/plugins/metalk8s.py
+install -p -m 755 %{_topdir}/SOURCES/containerd.py %{buildroot}/%{python_sitelib}/sos/plugins/containerd.py
+
+%files
+%defattr(-,root,root)
+%{python_sitelib}/sos/plugins/metalk8s.py
+%{python_sitelib}/sos/plugins/containerd.py
+%{python_sitelib}/sos/plugins/metalk8s.pyc
+%{python_sitelib}/sos/plugins/containerd.pyc
+%{python_sitelib}/sos/plugins/metalk8s.pyo
+%{python_sitelib}/sos/plugins/containerd.pyo
+
+%changelog
+* Fri Jul 05 2019 SayfEddine Hammemi <sayf-eddine.hammemi@scality.com> - 1.0.0-1
+- Initial build
--- a/packages/metalk8s-sosreport/src/metalk8s.py
+++ b/packages/metalk8s-sosreport/src/metalk8s.py
@@ -1,0 +1,119 @@
+#! /bin/env python
+
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+from os import path
+
+
+class metalk8s(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    """Metalk8s plugin"""
+
+    plugin_name = 'metalk8s'
+    packages = ('kubernetes', 'kubernetes-master')
+    profiles = ('container',)
+    files = ('/etc/kubernetes/admin.conf',)
+
+    option_list = [
+        ('all', 'also collect all namespaces output separately',
+            'slow', False),
+        ('describe', 'capture descriptions of all kube resources',
+            'fast', False),
+        ('podlogs', 'capture logs for pods', 'slow', False),
+    ]
+
+    def check_is_master(self):
+        return any([path.exists("/etc/kubernetes/admin.conf")])
+
+    def setup(self):
+        self.add_copy_spec('/etc/kubernetes/manifests')
+        self.add_copy_spec('/var/log/pods')
+        self.add_copy_spec('/var/log/metalk8s-bootstrap.log')
+
+        services = [
+            'kubelet',
+        ]
+
+        for service in services:
+            self.add_journal(units=service)
+
+        # We can only grab kubectl output from the master
+        if self.check_is_master():
+            kube_cmd = 'kubectl '
+            if path.exists('/etc/kubernetes/admin.conf'):
+                kube_cmd += '--kubeconfig=/etc/kubernetes/admin.conf'
+
+            kube_get_cmd = 'get -o json '
+            for subcmd in ['version', 'config view']:
+                self.add_cmd_output('{0} {1}'.format(kube_cmd, subcmd))
+
+            # get all namespaces in use
+            namespaces_result = self.get_command_output('{0} get namespaces'.format(kube_cmd))
+            kube_namespaces = [n.split()[0] for n in namespaces_result['output'].splitlines()[1:] if n]
+
+            resources = [
+                'pods',
+                'deploy',
+                'rc',
+                'services',
+                'ds',
+                'cm',
+            ]
+
+            # nodes and pvs are not namespaced, must pull separately.
+            # Also collect master metrics
+            self.add_cmd_output([
+                '{} get -o json nodes'.format(kube_cmd),
+                '{} get -o json pv'.format(kube_cmd),
+                '{} get --raw /metrics'.format(kube_cmd)
+            ])
+
+            for n in kube_namespaces:
+                kube_namespace = '--namespace={}'.format(n)
+                if self.get_option('all'):
+                    kube_namespaced_cmd = '{0} {1} {2}'.format(kube_cmd, kube_get_cmd, kube_namespace)
+
+                    self.add_cmd_output('{} events'.format(kube_namespaced_cmd))
+
+                    for res in resources:
+                        self.add_cmd_output('{0} {1}'.format(kube_namespaced_cmd, res))
+
+                    if self.get_option('describe'):
+                        # need to drop json formatting for this
+                        kube_namespaced_cmd = '{0} get {1}'.format(kube_cmd, kube_namespace)
+                        for res in resources:
+                            r = self.get_command_output(
+                                '{0} {1}'.format(kube_namespaced_cmd, res))
+                            if r['status'] == 0:
+                                kube_cmd_result = [k.split()[0] for k in
+                                          r['output'].splitlines()[1:]]
+                                for k in kube_cmd_result:
+                                    kube_namespaced_cmd = '{0} {1}'.format(kube_cmd, kube_namespace)
+                                    self.add_cmd_output(
+                                        '{0} describe {1} {2}'.format(kube_namespaced_cmd, res, k))
+
+                if self.get_option('podlogs'):
+                    kube_namespaced_cmd = '{0} {1}'.format(kube_cmd, kube_namespace)
+                    r = self.get_command_output('{} get pods'.format(kube_namespaced_cmd))
+                    if r['status'] == 0:
+                        pods = [p.split()[0] for p in
+                                r['output'].splitlines()[1:]]
+                        for pod in pods:
+                            self.add_cmd_output('{0} logs {1}'.format(kube_namespaced_cmd, pod))
+
+            if not self.get_option('all'):
+                kube_namespaced_cmd = '{} get --all-namespaces=true'.format(kube_cmd)
+                for res in resources:
+                    self.add_cmd_output('{0} {1}'.format(kube_namespaced_cmd, res))
+
+    def postproc(self):
+        # First, clear sensitive data from the json output collected.
+        # This will mask values when the 'name' looks susceptible of
+        # values worth obfuscating, i.e. if the name contains strings
+        # like 'pass', 'pwd', 'key' or 'token'
+        env_regexp = r'(?P<var>{\s*"name":\s*[^,]*' \
+            r'(pass|pwd|key|token|cred|PASS|PWD|KEY)[^,]*,\s*"value":)[^}]*'
+        self.do_cmd_output_sub('kubectl', env_regexp,
+                               r'\g<var> "********"')
+        # Next, we need to handle the private keys and certs in some
+        # output that is not hit by the previous iteration.
+        self.do_cmd_private_sub('kubectl')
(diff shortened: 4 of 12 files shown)


salt/metalk8s/internal/sreport/installed.sls
```diff
@@ -0,0 +1,11 @@
...
+  - metalk8s.repo
+
+Install sos report:
+  pkg.installed:
+    - name: sos
+
+Inject custom plugins:
+  file.recurse:
+    - source: salt://metalk8s/internal/sreport/files
+    - name: /usr/lib/python2.7/site-packages/sos/plugins
```
gdemonet: Oh, it's sad that they don't define a cleaner way to register new plugins...

salt/metalk8s/internal/sreport/installed.sls
```diff
@@ -0,0 +1,11 @@
+include:
+  - metalk8s.repo
+
+Install sos report:
+  pkg.installed:
+    - name: sos
```
gdemonet: Different name on Ubuntu, will that be a problem?

gdemonet (commented):
Early thoughts, will test and review in depth tomorrow! Looks like a good tool though :+1:

salt/metalk8s/internal/sreport/installed.sls
```diff
@@ -0,0 +1,11 @@
+include:
+  - metalk8s.repo
+
+Install sos report:
+  pkg.installed:
```
TeddyAndrieux: You should use the macro here

salt/metalk8s/internal/sreport/installed.sls
```diff
@@ -0,0 +1,11 @@
...
+    - name: sos
+
+Inject custom plugins:
+  file.recurse:
```
TeddyAndrieux: `- include_pat: '*.py'` :)

salt/metalk8s/internal/sreport/collected.sls
```diff
@@ -0,0 +1,3 @@
+Collect reports:
+  cmd.run:
+    - name: sosreport -o metalk8s -o containerd --batch -kmetalk8s.podlogs -kmetalk8s.describe -kcontainerd.logs -kcontainerd.all
```
TeddyAndrieux: Suggest this one to include and require the `installed` state and btw remove the `installed` from `init.sls`

salt/metalk8s/internal/sreport/installed.sls
TeddyAndrieux: This state should require the `- test: Repositories configured`

salt/metalk8s/internal/sreport/installed.sls
TeddyAndrieux: This one should require the installation just above

NicolasT: Yes. This implies we're dropping 'random files' in `/usr`, which is not according to the project goals. A blocker as far as I'm concerned.

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
+from sos.plugins import Plugin, RedHatPlugin
+from os import path
+
+
+class metalk8s(Plugin, RedHatPlugin):
```
NicolasT: What makes this a `RedHatPlugin` and not an `UbuntuPlugin`?

salt/metalk8s/internal/sreport/collected.sls
```diff
@@ -0,0 +1,3 @@
+Collect reports:
```
NicolasT: I fail to see how this is a 'state'.

salt/metalk8s/internal/sreport/installed.sls
sayf-eddine-scality: the clearer way would be to create an rpm package that drop the plugin here, I ll look of we can use safer location

packages/packages.list
```diff
@@ -10,3 +10,4 @@ runc
 salt-minion-2018.3.4-1.el7
 skopeo
 yum-plugin-versionlock
+sos
```
slaperche-scality: minor: missing newline at the end of the file (as indicated by GitHub :stuck_out_tongue: )

(also the case in several other files you added in this PR)

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class containerd(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    """containerd engine
```
slaperche-scality: PEP-257

> The closing quotes are on the same line as the opening quotes. This looks better for one-liners.

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    plugin_name = 'containerd'
+    profiles = ('container',)
+    packages = ("cri-tools")
```
slaperche-scality: Why the double quotes here (when the two lines above are using single quotes)?

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+
+    option_list = [
+        ("all", "enable capture for all containers, even containers "
+            "that have terminated", 'fast', False),
+        ("logs", "capture logs for running containers",
```
slaperche-scality: This fit on a single line, so why are you wrapping it?

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+
+    """containerd engine
+    """
+
+    plugin_name = 'containerd'
+    profiles = ('container',)
+    packages = ("cri-tools")
+
+    option_list = [
+        ("all", "enable capture for all containers, even containers "
```
slaperche-scality: Same here (and below): why are you mixing single and double quote?

Pick one and be consistent.

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+            'pods',
+            'ps',
+            'ps -a',
+            'ps -v',
+            'stats',
+            'version',
+        ]
+
+        self.add_journal(units="containerd")
+        self.add_cmd_output(["crictl %s" % s for s in subcmds])
```
slaperche-scality: prefer `format` over `%`, (the later is kinda deprecated, except for logger).

ditto elsewhere in the file.

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+            self.add_cmd_output("crictl inspecti %s" % image)
+
+        for pod in pods:
+            self.add_cmd_output("crictl inspectp %s" % pod)
+
+    def _get_crio_list(self, cmd):
+        ret = []
+        result = self.get_command_output(cmd)
+        if result['status'] == 0:
+            for ent in result['output'].splitlines():
```
slaperche-scality: `ent`? do you mean `entry` (or maybe `line`?)

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+                ret.append(ent)
+            # Prevent the socket deprecation warning from being iterated over
+            if 'deprecated' in ret[0]:
```
slaperche-scality: instead of pop after the append, you could write the loop as:

```python
for ent in result['output'].splitlines():
    # Don't append the socket deprecation warning.
    if 'deprecated'  not in ret[0]:
        ret.append(ent)
```

salt/metalk8s/internal/sreport/files/containerd.py
```diff
@@ -0,0 +1,73 @@
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
```
slaperche-scality: Most of the comments here also applies to `metalk8s.py`

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+
+    option_list = [
+        ("all", "also collect all namespaces output separately",
+            'slow', False),
+        ("describe", "capture descriptions of all kube resources",
+            'fast', False),
+        ("podlogs", "capture logs for pods", 'slow', False),
+    ]
+
+    def check_is_master(self):
```
slaperche-scality: Simply write:

```python
    def check_is_master(self):
        return any([path.exists("/etc/kubernetes/admin.conf")])
```

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+        ]):
+            return True
+        return False
+
+    def setup(self):
+        self.add_copy_spec("/etc/kubernetes")
+        self.add_copy_spec("/var/log/pods")
+        self.add_copy_spec("/var/log/metalk8s-bootstrap.log")
+
+        svcs = [
```
slaperche-scality: Could be wrapped on a single line (or do you expect to add a lot of entries in the future?)

slaperche-scality: svcs? do you mean `services` (if yes, then use `services`, it's way more understandable)

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+            kube_cmd = "kubectl "
+            if path.exists('/etc/kubernetes/admin.conf'):
+                kube_cmd += "--kubeconfig=/etc/kubernetes/admin.conf"
+
+            kube_get_cmd = "get -o json "
+            for subcmd in ['version', 'config view']:
+                self.add_cmd_output('%s %s' % (kube_cmd, subcmd))
+
+            # get all namespaces in use
+            kn = self.get_command_output('%s get namespaces' % kube_cmd)
```
slaperche-scality: `kn`, `knsps`?
We should clearer names.

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+                    self.add_cmd_output('%s events' % k_cmd)
+
+                    for res in resources:
+                        self.add_cmd_output('%s %s' % (k_cmd, res))
+
+                    if self.get_option('describe'):
+                        # need to drop json formatting for this
+                        k_cmd = '%s get %s' % (kube_cmd, knsp)
+                        for res in resources:
+                            r = self.get_command_output(
```
slaperche-scality: `r` => `res` or `result`, avoid one letter variable name as much as possible…

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+            self.add_cmd_output([
+                "{} get -o json nodes".format(kube_cmd),
+                "{} get -o json pv".format(kube_cmd),
+                "{} get --raw /metrics".format(kube_cmd)
+            ])
+
+            for n in knsps:
+                knsp = '--namespace=%s' % n
+                if self.get_option('all'):
+                    k_cmd = '%s %s %s' % (kube_cmd, kube_get_cmd, knsp)
```
slaperche-scality: `k_cmd`, why the `k_` prefix? What does it means?

If you can't use `cmd` because you already have variable named like that, maybe it's the sign that we should split this function (it becames too big).

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+                        self.add_cmd_output('%s %s' % (k_cmd, res))
+
+                    if self.get_option('describe'):
+                        # need to drop json formatting for this
+                        k_cmd = '%s get %s' % (kube_cmd, knsp)
+                        for res in resources:
+                            r = self.get_command_output(
+                                '%s %s' % (k_cmd, res))
+                            if r['status'] == 0:
+                                k_list = [k.split()[0] for k in
```
slaperche-scality: what is a `k_list`?

salt/metalk8s/internal/sreport/files/metalk8s.py
```diff
@@ -0,0 +1,123 @@
...
+                k_cmd = '%s get --all-namespaces=true' % kube_cmd
+                for res in resources:
+                    self.add_cmd_output('%s %s' % (k_cmd, res))
+
+    def postproc(self):
+        # First, clear sensitive data from the json output collected.
+        # This will mask values when the "name" looks susceptible of
+        # values worth obfuscating, i.e. if the name contains strings
+        # like "pass", "pwd", "key" or "token"
+        env_regexp = r'(?P<var>{\s*"name":\s*[^,]*' \
```
slaperche-scality: Not very readable/understandable.

Either give an example string of what it's supposed to match, or use the verbose regexp mode that allows to insert comments in the regexp.

slaperche-scality (dismissed):

salt/metalk8s/internal/sreport/files/metalk8s.py
sayf-eddine-scality: I agree this is not the best code you can read but this is half vendored/modified code and I fear it become hard to adapt to the upstream if we derive so much

slaperche-scality: Hum, not very fond of the half vendored/modified…

- Either it's vendored code, and we should keep it as-is, with no modification.
- Either we take the ownership and we can rewrite it as we want.

Falling between two stools gives us the worse of both world:
- you don't want to fix the code, because it's vendored
- but you've still modified it enough to be pain when we will want to follow upstream (conflicts).

sayf-eddine-scality:
@bert-e reset

salt/metalk8s/internal/sreport/files/metalk8s.py
sayf-eddine-scality: make sens, going to modify it, still it is ready for a functional review

sayf-eddine-scality:
@bert-e reset

sayf-eddine-scality:
@bert-e reset

buildchain/buildchain/packaging.py
```diff
@@ -148,9 +148,23 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
 
 
 CALICO_CNI_PLUGIN_VERSION : str = '3.8.0'
+METALK8S_SOSREPORT_VERSION : str = '1.0.0'
```
slaperche-scality: Since it's used only at a single place, I don't think you need the constant: you can directly write `version='1.0.0'` l.159.

salt/metalk8s/internal/sreport/installed.sls
```diff
@@ -1,7 +1,9 @@
+{%- from "metalk8s/macro.sls" import pkg_installed with context %}
```
slaperche-scality: This diff should be squashed into the commit `Add salt states to install sos and inject custom plugins`, no?

buildchain/buildchain/targets/package.py
```diff
@@ -238,9 +239,13 @@ def build_srpm(self) -> types.TaskDict:
     def _download_sources(self) -> None:
```
slaperche-scality: Function should be renamed (`_import_sources`? `_get_sources`?) and docstring updated.

buildchain/buildchain/targets/package.py
```diff
@@ -238,9 +239,13 @@ def build_srpm(self) -> types.TaskDict:
...
         """Return a list of actions to download the source files."""
         for srcfile, url in self._get_source_files_urls().items():
-            with urllib.request.urlopen(url) as conn:
-                with open(srcfile, 'wb') as fp:
-                    fp.write(conn.read())
+            if url.startswith("http"):
```
slaperche-scality: This is too restrictive. We could download from HTTPS or FTP.

I would write the code like that instead:

```python
            if urllib.parse.urlparse(url).scheme:
                with urllib.request.urlopen(url) as conn:
                    with open(srcfile, 'wb') as fp:
                        fp.write(conn.read())
            else:
                url = os.path.join(constants.ROOT/'packages', url)
                shutil.copyfile(url, srcfile)
```

If we have a scheme => we download, if not it's a local file and we copy it.

packages/metalk8s-sosreport.spec
```diff
@@ -0,0 +1,38 @@
+Name:		metalk8s-sosreport
+Version:	1.0.0
+Release:	1%{?dist}
+Summary: 	Metalk8s SOS report custom plugins
+
+
+Requires: python >= 2.6, python < 2.8
```
slaperche-scality, line 7: Is the `python < 2.8` really needed? I don't think Python 2.8 will ever exists. Or maybe simple `python = 2.7`?

slaperche-scality (dismissed):

packages/metalk8s-sosreport.spec
sayf-eddine-scality: we want all versions of `2.7.X` so we go for ` < 2.8`

sayf-eddine-scality:
@bert-e reset

packages/metalk8s-sosreport/src/metalk8s.py
```diff
@@ -0,0 +1,119 @@
...
+            self.add_cmd_output([
+                '{} get -o json nodes'.format(kube_cmd),
+                '{} get -o json pv'.format(kube_cmd),
+                '{} get --raw /metrics'.format(kube_cmd)
+            ])
+
+            for n in kube_namespaces:
+                kube_namespace = '--namespace={}'.format(n)
+                if self.get_option('all'):
+                    kube_namedspaced_cmd = '{0} {1} {2}'.format(kube_cmd, kube_get_cmd, kube_namespace)
```
gdemonet: minor (also on other lines):

```suggestion
                    kube_namespaced_cmd = '{0} {1} {2}'.format(kube_cmd, kube_get_cmd, kube_namespace)
```

buildchain/buildchain/packaging.py
```diff
@@ -151,6 +151,19 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
 # Packages per repository.
 PACKAGES : Dict[str, Tuple[targets.Package, ...]] = {
     'scality': (
+        # SOS report custom plugins.
+        targets.Package(
+            basename='_build_packages',
+            name='metalk8s-sosreport',
+            version='1.0.0',
```
TeddyAndrieux: I think this package version should be the version of metalk8s

salt/metalk8s/internal/sreport/init.sls
```diff
@@ -0,0 +1,2 @@
+include:
+  - .installed
```
TeddyAndrieux: I think you need to include it somewhere in the deployment of the nodes to install it anywhere automatically. Maybe part of the `minion` roles

packages/metalk8s-sosreport/src/metalk8s.py
```diff
@@ -0,0 +1,119 @@
...
+        ('describe', 'capture descriptions of all kube resources',
+            'fast', False),
+        ('podlogs', 'capture logs for pods', 'slow', False),
+    ]
+
+    def check_is_master(self):
+        return any([path.exists("/etc/kubernetes/admin.conf")])
+
+    def setup(self):
+        self.add_copy_spec('/etc/kubernetes')
```
gdemonet: That includes sensitive data (private keys, certs, kubeconfigs...), not sure if we want them :/

sayf-eddine-scality: right, will make this precise

sayf-eddine-scality: okay everything here is sensitive, removing this

salt/metalk8s/internal/sreport/init.sls
sayf-eddine-scality: as it is in internal, make sens to include it in its init.sls ?

packages/metalk8s-sosreport/src/metalk8s.py
gdemonet: maybe `/etc/kubernetes/manifests` could be useful, not sure though

sayf-eddine-scality: the manifests are not changed in runtime so they re already in the codebase

TeddyAndrieux: I agree with @gdemonet we should include the manifest if for some reason a user edited the manifest we want to have them

sayf-eddine-scality: hmmm yes makes sens

salt/metalk8s/internal/sreport/init.sls
TeddyAndrieux: We never run the entire `metalk8s.internal`. 
And btw this sosreport package it's not "internal" you should move it directly to `salt/metalk8s`. 
`internal` directory is more for salt internal stuff. That mean states that should not be call directly.

sayf-eddine-scality:
@bert-e reset

slaperche-scality (approved):

sayf-eddine-scality:
@bert-e approve

PR #1368, opened 2019-07-04
calico: upgrade to Calico 3.8.0

NicolasT:
- The manifest was retrieved from
  https://docs.projectcalico.org/v3.8/manifests/calico.yaml and edited
  for our purposes, disabling the FlexVolume driver deployment for
  Dikastes support, which we don't provide right now.

- CNI plugin version got bumped to 0.3.1.

- The new manifest uses stable APIs/versions, so the Kubernetes rendered
  required a couple of tweaks to support these.

**Component**: CNI


NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #1369, opened 2019-07-04
packages: use CentOS upstream `container-selinux`

### LINKED ISSUE #575: debt: Fix for #573 (#574) should be reverted once CentOS packages get updated
The fix for #573 in #574 introduces some technical debt by installing a non-official package from a third-party repository. Once a sufficiently-recent version of `container-selinux` lands in CentOS, we can undo this, or decide to roll our own build.

NicolasT:
The upstream CentOS 'extras' repository now ships
`container-selinux-2.99-1.el7_6.noarch` which is sufficiently new enough
(i.e., >= 2.77) for `containerd` to work properly. As a result, we no
longer need the version from ScientificLinux.

Since this was the only package in the `metalk8s-external` repository,
this whole repository and the build infrastructure coming with it is now
gone.

Fixes: #575
See: https://github.com/scality/metalk8s/issues/575
See: https://github.com/scality/metalk8s/issues/573
See: https://github.com/scality/metalk8s/pull/574

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -185,12 +185,6 @@ REPOSITORIES : Tuple[targets.Repository, ...] = (
     ),
     targets.Repository(
         basename='_build_repositories',
-        name='external',
-        builder=BUILDER,
-        task_dep=['_download_packages'],
-    ),
-    targets.Repository(
-        basename='_build_repositories',
         name='extras',
         builder=BUILDER,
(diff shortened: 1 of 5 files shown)


NicolasT:
@bert-e reset

NicolasT:
@bert-e approve

slaperche-scality (approved):

gdemonet (approved):
LGTM

PR #1374, opened 2019-07-05
fix cleanup condition for local images

### LINKED ISSUE #1373: fix LocalImage cleanup
**Component**:

buildchain

**What happened**:

`doit.sh clean` doesn't remove the locally built images from the build tree.

**What was expected**:

The locally built images are properly deleted during the cleanup.

**Steps to reproduce**

From a clean build tree, run:
- build the images: `./doit.sh -n 4 _image_build`
- clean them up: `./doit.sh clean`
- check that they aren't deleted with `ls _build/root/images/*/*`

**Resolution proposal**:

In `buildchain/buildchain/targets/local_image.py`, there is this code to select the cleanup action.

```python
'clean': True if self.save_on_disk else [self.clean]
```

Simply swap the branches of the condition…

slaperche-scality:
**Component**:

buildchain

**Context**: 

`doit.sh clean` doesn't remove the locally built images from the build tree.

**Summary**:

When we save on disk, we use `skopeo` and it creates a file tree that should be cleaned up with a recursive `rm`
Otherwise, we simply touch a file and using `True` is enough to let `doit` do the cleanup.

We simply have to  swap the branches to fix the issue.

**Acceptance criteria**: 

From a clean build tree, run:
- build the images: `./doit.sh -n 4 _image_build`
- clean them up: `./doit.sh clean`
- eveything should be removed (i.e. `_build` no longer exists)

---

Closes: #1373

### DIFF
--- a/buildchain/buildchain/targets/local_image.py
+++ b/buildchain/buildchain/targets/local_image.py
@@ -174,7 +174,7 @@ class LocalImage(image.ContainerImage):
                 self.dirname/'manifest.json' if self.save_on_disk
                 else self.dest_dir/self.tag
             ],
-            'clean': True if self.save_on_disk else [self.clean]
+            'clean': [self.clean] if self.save_on_disk else True
         })
         return task
 


gdemonet (approved):
Works as advertised!

slaperche-scality:
@bert-e approve

slaperche-scality:
@bert-e reset

PR #1377, opened 2019-07-08
Run multi-nodes CI stage offline

### LINKED ISSUE #713: Make sure nothing is pulled from the outside when installing
The ISO should be self-sufficient and contains everything necessary for an offline installation.
Any connection to the outside (internet, proxy cache, etc.) should be forbidden in the CI environments.

However, Vagrant being essentially a development tool, we will not enforce offline mode there (we may add an option to enable it in the future).

### LINKED ISSUE #1093: Ship a debug/utilities container image
For operational use, it would be useful to have a container (image) available offline which contains tooling like `ping`, `telnet`, `curl`,...

gdemonet:
**Component**: ci, tests

**Context**: We need to catch undesired behaviours of the product w.r.t. online connectivity.

**Summary**:

- run the multi-nodes deployment stage offline (using Openstack security groups)
- ship a simple `metalk8s-utils` container image to replace previous use of `busybox` in tests
    - NB: the ISO only increases in size for around 15MB, thanks to the same base image as other containers and the use of `hardlink`

**Acceptance criteria**:

- tests in CI should pass
- one should be able to install an offline platform, and run a container using the utils container image `{bootstrap_control_plane_ip}:8080/metalk8s-{version}/metalk8s-utils:{version}`

---

Closes: #713 
Closes: #1093

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,13 +1,14 @@
+import json
 import pathlib
 
 import kubernetes
 import pytest
 from pytest_bdd import given, parsers, then
-import testinfra
 import yaml
 
 from tests import kube_utils
 from tests import utils
+
 
 # Pytest command-line options
 def pytest_addoption(parser):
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -106,6 +107,48 @@ def bootstrap_config(host):
...
+    with host.sudo():
+        ip_json = host.check_output(
+            "salt-call grains.get metalk8s:control_plane_ip --out json"
+        )
+
+    return json.loads(ip_json)["local"]
+
+
+@pytest.fixture
+def build_image_name(k8s_client, control_plane_ip, version):
```
slaperche-scality: Does this need to be a fixture or can it be simply an helper?

If we want a fixture, maybe rename it `image_name_builder`.
Usually, fixture name behave like "variable" instead of "function" (i.e.: they're not named like action => `control_plane_ip` vs `get_control_plane_ip`).

gdemonet: I wanted a fixture for easy access, since I don't see other use cases for a `registry_address` fixture. The thing is, fixtures are named as the variables that would hold their return value. In that particular case, the return value of the fixture **is** a function, so I find it more explicit than the abstract `_builder` suffix.

gdemonet:
@bert-e reset

gdemonet:
@bert-e reset

slaperche-scality (dismissed):

gdemonet:
@bert-e approve

gdemonet:
@bert-e reset

slaperche-scality (approved):

PR #1409, opened 2019-07-11
salt: Remove retry output from etcd status check

### LINKED ISSUE #1363: Remove "retry output" when checking etcd status
**Component**:

'salt', 'kubernetes', 'upgrade'

**Why this is needed**:

When we upgrade etcd cluster we check etcd status and we get "error" output because we do some retry.

```
[ERROR   ] cluster may be unhealthy: failed to list members
  File "/var/cache/salt/master/extmods/modules/metalk8s_etcd.py", line 147, in check_etcd_health
    etcd_members = list(etcd.members)
  File "/usr/lib/python2.7/site-packages/etcd3/client.py", line 893, in members
    metadata=self.metadata
  File "/usr/lib64/python2.7/site-packages/grpc/_channel.py", line 565, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/usr/lib64/python2.7/site-packages/grpc/_channel.py", line 467, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
_Rendezvous: <_Rendezvous of RPC that terminated with:
    status = StatusCode.UNAVAILABLE
    details = "failed to connect to all addresses"
    debug_error_string = "{"created":"@1562239014.171578976","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3528,"referenced_errors":[{"created":"@1562239014.171573960","description":"failed to connect to all addresses","file":"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":399,"grpc_status":14}]}"
>
[ERROR   ] Unhandled exception occurred in function "run: <_Rendezvous of RPC that terminated with:
    status = StatusCode.UNAVAILABLE
    details = "failed to connect to all addresses"
    debug_error_string = "{"created":"@1562239014.171578976","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3528,"referenced_errors":[{"created":"@1562239014.171573960","description":"failed to connect to all addresses","file":"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":399,"grpc_status":14}]}"
```

**What should be done**:

The error during retry should be hidden

TBD
Error happen in `salt/metalk8s/orchestrate/upgrade/etcd.sls`
```
Check etcd cluster health for {{ node }}:
  metalk8s.module_run:
    - metalk8s_etcd.check_etcd_health:
      - minion_id: {{ node }}
    - attemps: 5
    - require:
      - salt: Upgrade etcd {{ node }} to {{ dest_version }}
```

**Test plan**:

Ebaneck:
**Component**:

'salt', 'kubernetes', 'upgrade'

**Context**: 

Issue: #1363 

**Summary**:

During an upgrade, we perform an etcd status health check which prints a messy stack trace.

We do not want to see this mess, so we have decided to hide it.

**Acceptance criteria**: 

- No stacktrace when etcd performs status check during upgrade

---

Closes: #1363

### DIFF
--- a/salt/_modules/metalk8s_etcd.py
+++ b/salt/_modules/metalk8s_etcd.py
@@ -138,16 +138,12 @@ def check_etcd_health(
     )[minion_id]
 
     # Get all members
-    try:
-        with etcd3.client(host=endpoint,
-                          ca_cert=ca_cert,
-                          cert_key=cert_key,
-                          cert_cert=cert_cert,
-                          timeout=TIMEOUT) as etcd:
-            etcd_members = list(etcd.members)
-    except Exception:  # pylint: disable=broad-except
-        log.exception("cluster may be unhealthy: failed to list members")
-        raise
+    with etcd3.client(host=endpoint,
+                      ca_cert=ca_cert,
+                      cert_key=cert_key,
+                      cert_cert=cert_cert,
+                      timeout=TIMEOUT) as etcd:
+        etcd_members = list(etcd.members)
 
     unhealthy_member = 0
     for member in etcd_members:
@@ -161,7 +157,7 @@ def check_etcd_health(
                               timeout=TIMEOUT) as etcd:
                 etcd.status()
         except Exception:  # pylint: disable=broad-except
-            log.exception(
+            log.debug(
                 "failed to check the health of member %s", member.name
             )
             unhealthy_member += 1


salt/_states/metalk8s.py
```diff
@@ -87,8 +87,10 @@ def module_run(name, attemps=1, sleep_time=10, **kwargs):
                 **kwargs
             )
         except Exception as exc:  # pylint: disable=broad-except
-            ret['comment'] = str(exc)
-
+            if name == str('metalk8s_etcd.check_etcd_health'):
+                ret['comment'] = None
+            else:
+                ret['comment'] = str(exc)
```
TeddyAndrieux: Nop it's not here.
You need to change in the module itself the log.warning.
But we may want to have this log somehow maybe we should log it in debug and raise it so we will have it in the resulting output and when we retry it's only printed in debug mode

gdemonet: @TeddyAndrieux Since we own the `metalk8s_etcd` module, why not give a `retry` argument directly to the `check_etcd_health` method?

gdemonet: Also, that's not the only retry that will print error logs... Either each `retry` we need will print logs unless we catch it here (looks like a hack to me), or we find some way to catch logs while retrying (which we may not want to do, since logs are useful).

TeddyAndrieux: We can do it directly in `check_etcd_health` but it's kind of strange to me that a module automatically retry. It's more something we want to do in a state not when calling a module...
But I think in the module in case of critical error we should only raise this error and not log anything and then it's the state that handle this, so in our case we controle the retrying state so when retrying we just ignore (or maybe log in debug to say "attemps x: got this error: blabla")

TeddyAndrieux: And yes it's not the only retry we will have but I think we should just not log in error in our modules just raise in case of error and let the caller handle this

Ebaneck:
@bert-e reset

salt/_modules/metalk8s_etcd.py
```diff
@@ -146,7 +146,6 @@ def check_etcd_health(
                           timeout=TIMEOUT) as etcd:
             etcd_members = list(etcd.members)
     except Exception:  # pylint: disable=broad-except
-        log.exception("cluster may be unhealthy: failed to list members")
         raise
```
TeddyAndrieux: You can remove the try except

Ebaneck:
@bert-e reset

TeddyAndrieux (approved):

Ebaneck:
@bert-e approve

Ebaneck:
@bert-e reset

Ebaneck:
@bert-e reset

Ebaneck:
@bert-e reset

Ebaneck:
@bert-e reset

PR #1414, opened 2019-07-11
Correction of the vagrantfile

### LINKED ISSUE #1413: Python2.7 installed on  vagrant Ubuntu node
**Component**: Vagrant, ubuntu

**What happened**: python is not installed at the VM boot

**What was expected**: Python already installed when Ubuntu is deployed

**Steps to reproduce**: 
-  $ vagrant up ubuntu1 
- $ vagrant ssh ubuntu1

ycointe:
Permit to install Python2.7 on Ubuntu node
during the VM deployment.

Issue: GH-1413

**Component**:

**Context**: Need python2.7 on Ubuntu nodes for salt minion installation

**Summary**: Permits to have some Ubuntu nodes on Vagrant with python 2.7 installed

**Acceptance criteria**: Have python 2.7 already installed on Ubuntu nodes.

---

Closes: #1413

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -235,7 +235,7 @@ Vagrant.configure("2") do |config|
           type: "shell",
           inline: DEPLOY_SSH_PUBLIC_KEY
 
-        if os == "ubuntu"
+        if os == :ubuntu
           node.vm.provision "install-python",
             type: "shell",
             inline: INSTALL_PYTHON


jbertran (dismissed):

ycointe:
@bert-e approve

ycointe:
@bert-e approve

slaperche-scality (changes requested):
Code LGTM, but commit message could be improved to mention what was the problem (and how we solve it).
Especially since people are not very familiar with Ruby, that can be helpful.

I would suggest something like:

```
vagrant: fix OS comparison

During the review of the previous PR, we replaced the hash keys that
were strings by symbols, which is more idiomatic in Ruby (it's also more
efficient, but we don't care in this specific use case) We updated the
code accordingly (hash lookup) but forgot to fix the comparison.

As a consequence, the condition was always false (because we were
comparing a symbol against a string) and Python was never installed,
even on Ubuntu.

Using a symbol literal instead of a string literal is enough to fix the
issue.
```

ycointe:
@bert-e reset

ycointe:
@bert-e reset

slaperche-scality (dismissed):

jbertran (approved):
LGTM!

Issue #1353, opened 2019-07-03
Ubuntu workload plane node deployment online


ycointe:
**Component**: packages, salt

**Why this is needed**: The salt formula permit to install some packages on the Ubuntu node.
Actually, these packages are configured to be installed on Centos only.

**What should be done**: Configure salt to install required packages on Ubuntu minions

**Implementation proposal** (strongly recommended):
- List Ubuntu required packages
- Adapt salt formula (configure repo and install packages)

**Test plan**:
Deploy Ubuntu node in the workload plane.

PR #1416, opened 2019-07-12
Improvement/1353 ubuntu formula

ycointe:
**Component**: formula, packages, salt

**Context**: Metalk8s over Ubuntu

**Summary**: Modify salt formula to permits workload plane Ubuntu nodes deployment.

**Acceptance criteria**: 

- Deploy a Ubuntu Workload plane node
- Verify every pods are scheduled and working well

---

Closes: #1353


salt/metalk8s/repo/debian.sls
```diff
@@ -0,0 +1,27 @@
+Install python-apt:
+  pkg.installed:
+    - name: python-apt
```
jbertran, line 3: Let's freeze the version of that python package.

salt/metalk8s/repo/init.sls
```diff
@@ -1,2 +1,2 @@
 include:
-  - .offline
+  - .redhat
```
NicolasT, line 2: This is a bit odd. Shouldn't we dispatch on host OS here and include the 'correct' file accordingly?

NicolasT:
There's no Ubuntu support in MetalK8s.

PR #1420, opened 2019-07-12
salt: remove node taint(node.kubernetes.io/unschedulable) before uncordon

### LINKED ISSUE #1408: Cordon when downgrading from 1.12 to 1.11 make the node unschedulable even after uncordon
**Component**:

'salt', 'kubernetes', 'upgrade'

**What happened**:

With version 1.12 of k-c-m when cordoning a node k-c-m add a taints to the node
```
  - effect: NoSchedule
    key: node.kubernetes.io/unschedulable
    timeAdded: 2019-07-11T12:46:25Z
```
```
2019-07-11T12:46:25.703186432Z stderr F I0711 12:46:25.702868       1 controller_utils.go:248] Added [&Taint{Key:node.kubernetes.io/unschedulable,Value:,Effect:NoSchedule,TimeAdded:2019-07-11 12:46:25.669028408 +0000 UTC m=+92.589019139,}] Taint to Node single-node.novalocal
```
But this taint is not handle by k-c-m 1.11 so when we downgrade from 1.12 to 1.11 we cordon with k-c-m 1.12 and uncordon with k-c-m 1.11 so the taints remaining and the nodes stay unschedulable

**What was expected**:

Uncordon make the node schedulable again

**Steps to reproduce**

Install metalk8s-2.0.0 make an iso 2.1.0 available (https://github.com/scality/metalk8s/blob/development/2.0/docs/operation/preparation.rst),
Upgrade k-c-m 
```
salt-call state.sls metalk8s.kubernetes.controller-manager saltenv=metalk8s-2.1.0-dev
```

Cordon the node:
```
crictl exec -it $(crictl ps -q --label io.kubernetes.container.name=salt-master) bash
salt-run salt.cmd state.single metalk8s_cordon.node_cordoned <node_name> kubeconfig=/etc/kubernetes/admin.conf
```

Downgrade k-c-m
```
salt-call state.sls metalk8s.kubernetes.controller-manager saltenv=metalk8s-2.0.0-dev
```

Uncordon the node:
```
crictl exec -it $(crictl ps -q --label io.kubernetes.container.name=salt-master) bash
salt-run salt.cmd state.single metalk8s_cordon.node_uncordoned <node_name> kubeconfig=/etc/kubernetes/admin.conf
```

Node still unschedulable due to taints (even if he have status `Ready`)

The `uncordon` module should remove this taint

Ebaneck:
**Component**:

'salt', 'kubernetes', 'downgrade'

**Context**: 

During uncordon in an upgrade or downgrade procedure, we do not taint nodes which seem to pose problems.
The issue becomes even more evident when nodes are marked unschedulable during downgrade as reported here #1408.

Issue: #1408

**Summary**:

- Added  a taint removal method for the above existing taint.
- Method should only be applied during uncordon

**Acceptance criteria**: 

- Uncordon should remove taint(node.kubernetes.io/unschedulable)
- Nodes should become schedulable after a downgrade

---

Closes: #1408

### DIFF
--- a/salt/_modules/metalk8s_cordon.py
+++ b/salt/_modules/metalk8s_cordon.py
@@ -99,4 +99,24 @@ def node_uncordon(node_name, **kwargs):
 
         salt '*' kubernetes.node_uncordon  node_name="minikube"
     '''
+    # k-c-m 1.12 adds taint `node.kubernetes.io/unschedulable` when cordonning
+    # a node but k-c-m 1.11 does not remove this taint when uncordonning so,
+    # we need to manually remove it when necessary by patching taints.
+
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](
+        node_name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    if key_to_remove in (taint['key'] for taint in all_node_taints):
+        new_taints = [
+            taint for taint in all_node_taints
+            if taint['key'] != key_to_remove
+        ]
+        __salt__['metalk8s_kubernetes.node_set_taints'](
+            node_name=node_name,
+            taints=new_taints,
+            **kwargs
+        )
+
     return node_set_unschedulable(node_name, False, **kwargs)
--- a/salt/_states/metalk8s_cordon.py
+++ b/salt/_states/metalk8s_cordon.py
@@ -52,7 +52,6 @@ def _node_set_unschedulable(name, value, **kwargs):
     return ret
 
 
-
 def node_cordoned(name, **kwargs):
     '''
     Ensures that the named node is cordoned.


gdemonet (dismissed):
Huh, this PR is quite confusing.

The title is 'Add node taint before uncordon', while the changes (and the original issue) is about removing the `unschedulable` taint when uncordoning, were it present.

Also, in `salt/_modules/metalk8s_kubernetes.py`, there are already methods for managing taints: `node_taints` and `node_set_taints`. Make sure to use those for this specific use-case, the logic shouldn't be complicated.

Finally, all operations requiring interaction with the K8s API from Salt **must not** use `kubectl`. There's a Python client to interface with the API, which we already use throughout the various custom modules that we ship.

Ebaneck:
> Huh, this PR is quite confusing.
> 
> The title is 'Add node taint before uncordon', while the changes (and the original issue) is about removing the `unschedulable` taint when uncordoning, were it present.
> 
> Also, in `salt/_modules/metalk8s_kubernetes.py`, there are already methods for managing taints: `node_taints` and `node_set_taints`. Make sure to use those for this specific use-case, the logic shouldn't be complicated.
> 
> Finally, all operations requiring interaction with the K8s API from Salt **must not** use `kubectl`. There's a Python client to interface with the API, which we already use throughout the various custom modules that we ship.

Completely my bad about the PR title... I have fixed this. @gdemonet 

I certainly will look into using existing modules for managing taints as you suggested + the existing python K8s wrapper.

salt/_modules/metalk8s_cordon.py
```diff
@@ -99,4 +99,13 @@ def node_uncordon(node_name, **kwargs):
 
         salt '*' kubernetes.node_uncordon  node_name="minikube"
     '''
-    return node_set_unschedulable(node_name, False, **kwargs)
+    # need to remove all taints before node uncordon
+    set_no_taints = []
+    remove_all_taint = __salt__['metalk8s_kubernetes.node_set_taints'](
```
NicolasT: Wait no, this would remove *all* taints on a node when uncordon'ing it? That ain't right at all!

NicolasT (dismissed):

salt/_modules/metalk8s_cordon.py
Ebaneck: @NicolasT from the issue, it seems the node taint key to be removed happens to be `node.kubernetes.io/unschedulable`

Maybe I should specify the node taint to be removed?

NicolasT: Nodes (can) have taints, for any key. We even rely on this for control-plane/infra/workload-plane separation, and other things.

Given this change, when uncordoning a node, all its taints would be removed, which is of course not what you intend to do: you're only looking to handle one taint. Can't remove any of the others.

Ebaneck: Thanks for the heads-up @NicolasT

salt/_modules/metalk8s_cordon.py
```diff
@@ -99,4 +99,13 @@ def node_uncordon(node_name, **kwargs):
...
+        node_name, set_no_taints, **kwargs
+    )
+
+    if remove_all_taint is not None:
+        return node_set_unschedulable(node_name, False, **kwargs)
```
TeddyAndrieux: Not in an if you need to remove unschedulable always.
You ~may~ have some version (1.11) where you will have basically no taint to remove normally.

The module should remove the taint `node.kubernetes.io/unschedulable` (and only this one) AND set `unschedulable` flag to `False` (even if no taint get removed)

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1182,6 +1182,40 @@ def node_taints_present(name, taints, **kwargs):
...
+           'result': False,
+           'comment': ''
+    }
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](name, **kwargs)
+    existing_taints = {t['key']: t for t in all_node_taints}
+
+    for taint in taints:
+        key_to_remove = taint['key']
+        if key_to_remove in existing_taints:
+            existing_taints.pop(key_to_remove)
```
TeddyAndrieux: Do not edit `existing_taints` create a new dict `new_taints`

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1182,6 +1182,40 @@ def node_taints_present(name, taints, **kwargs):
...
+
+    ret['changes'] = {
+        'old': list(existing_taints.values()),
```
TeddyAndrieux: As you `pop` from `existing_taints` you can't consider this as old

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1182,6 +1182,40 @@ def node_taints_present(name, taints, **kwargs):
...
+        'new': new_taints,
```
TeddyAndrieux: `new_taints` does not exists

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1182,6 +1182,40 @@ def node_taints_present(name, taints, **kwargs):
...
+    }
+    ret['comment'] = '{} will be removed'
```
TeddyAndrieux: You need to `format` this string

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1182,6 +1182,40 @@ def node_taints_present(name, taints, **kwargs):
...
+           'comment': ''
+    }
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](name, **kwargs)
+    existing_taints = {t['key']: t for t in all_node_taints}
+
+    for taint in taints:
+        key_to_remove = taint['key']
+        if key_to_remove in existing_taints:
+            existing_taints.pop(key_to_remove)
+
```
TeddyAndrieux: You need to return here if taint already absent (I know it's not in `node_taints_present` but it's something missing)

Ebaneck:
@bert-e reset

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,19 @@ def node_uncordoned(name, **kwargs):
...
+        name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    new_taints = [
+        t for t in all_node_taints
+        if t['key'] not in key_to_remove
+    ]
+
+    __salt__['metalk8s_kubernetes.node_set_taints'](
```
TeddyAndrieux: Add an `if` here to not call this one if not needed `if new_taints != all_node_taints`

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,19 @@ def node_uncordoned(name, **kwargs):
...
         the name of the node
     '''
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](
+        name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    new_taints = [
+        t for t in all_node_taints
+        if t['key'] not in key_to_remove
```
TeddyAndrieux: You need to use a list for `key_to_remove`
Here you will remove all taint that have a key who contains a part of  "node.kubernetes.io/unschedulable" so for example you will remove taint "kubernetes.io/unsche" (that does not exists btw but we should not do that).

OR much better you should change this line
```suggestion
        if t['key'] == key_to_remove
```

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,19 @@ def node_uncordoned(name, **kwargs):
     name
         the name of the node
     '''
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](
+        name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    new_taints = [
+        t for t in all_node_taints
```
TeddyAndrieux: Consider using `taint` instead of `t`

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,19 @@ def node_uncordoned(name, **kwargs):
     name
         the name of the node
     '''
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](
+        name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
```
TeddyAndrieux: Add a comment to say why we remove this taint
Something like
```
k-c-m 1.12 add taint `node.kubernetes.io/unschedulable` when cordonning a node but k-c-m 1.11 does not remove this one when uncordonning
```

TeddyAndrieux (changes requested):

salt/_states/metalk8s_cordon.py
Ebaneck: yeah I can add a comment within the codebase... But the commit message already has this

salt/_states/metalk8s_cordon.py
Ebaneck: Yes.. makes sense here to be specific on the taint we want to remove

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,25 @@ def node_uncordoned(name, **kwargs):
...
+    ]
+    if new_taints == all_node_taints:
+        return _node_set_unschedulable(name, False, **kwargs)
+    else:
+        __salt__['metalk8s_kubernetes.node_set_taints'](
+            node_name=name,
+            taints=new_taints,
+            **kwargs
+        )
+        return _node_set_unschedulable(name, False, **kwargs)
```
TeddyAndrieux: You don't need to do if else just do
```
    if new_taints != all_node_taints:
        __salt__['metalk8s_kubernetes.node_set_taints'](
            node_name=name,
            taints=new_taints,
            **kwargs
        )

    return _node_set_unschedulable(name, False, **kwargs)
```

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,25 @@ def node_uncordoned(name, **kwargs):
...
+    # we need to manually remove it when necessary by patching taints.
+
+    all_node_taints = __salt__['metalk8s_kubernetes.node_taints'](
+        name, **kwargs
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    new_taints = [
+        taint for taint in all_node_taints
+        if not (taint['key'] == key_to_remove)
```
TeddyAndrieux:
```suggestion
        if taint['key'] != key_to_remove
```

salt/_states/metalk8s_cordon.py
Ebaneck: ok

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,25 @@ def node_uncordoned(name, **kwargs):
...
+    ]
+    if new_taints == all_node_taints:
```
gdemonet: That's going to be flaky, I don't think we should trust ordering in the API response.

You have multiple solutions:

- You could make the taint `dict`s hashable (e.g. `serialized = "key={};value={};effect={}".format(t['key'], t['value'], t['effect'])`), then compare `set`s of these serialized values
- A **much** simpler/efficient approach is to test `if key_to_remove in (t['key'] for t in all_node_taints):` before attempting to remove the taint

TeddyAndrieux: I think the order should be the same as we create the `new_taints` list using `all_node_taints`.

But I agree with the second point check this before "computing" the `new_taints` list it's much better

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,26 @@ def node_uncordoned(name, **kwargs):
...
+    )
+    key_to_remove = "node.kubernetes.io/unschedulable"
+
+    if key_to_remove in (taint['key'] for taint in all_node_taints):
+        new_taints = [
+            taint for taint in all_node_taints
+            if taint['key'] != key_to_remove
+        ]
+
+    if new_taints != all_node_taints:
```
TeddyAndrieux: You don't need this `if` as if you are in the previous `if` you need to "remove" the taint

salt/_states/metalk8s_cordon.py
```diff
@@ -70,4 +69,26 @@ def node_uncordoned(name, **kwargs):
     name
         the name of the node
     '''
+    # k-c-m 1.12 adds taint `node.kubernetes.io/unschedulable` when cordonning
```
TeddyAndrieux: Oh wait you should do it in the module not in the state
`salt/_modules/metalk8s_cordon.py`

salt/_states/metalk8s_cordon.py
Ebaneck: ok

TeddyAndrieux (approved):
LGTM

Ebaneck:
@bert-e  reset

Ebaneck:
@bert-e approve

Ebaneck:
@bert-e approve

Issue #1410, opened 2019-07-11
Update buildchain/CI to build the Storage Operator


slaperche-scality:
**Component**:

buildchain, CI

**Why this is needed**:

We need to be able to build and deploy a Kubernetes operator written in Go.

**What should be done**:

- Update the documentation about the new building dependencies (go, operator-sdk 0.9.0, …)
- Add new dependencies in the CI build container
- Build the container image for the operator
- Lint the Go code of the operator (at least enforce go-fmt, maybe linting as well)
- Ensure `operator-sdk generate k8s` and `operator-sdk generate openapi` don’t result in a diff, which would otherwise mean codegen wasn’t run as part of the commit/changes

**Test plan**:

Being able to build an ISO that contains the operator container, both locally and on the CI.

PR #1432, opened 2019-07-15
Update buildchain and CI for operator

slaperche-scality:
**Component**:

ci, buildchain

**Context**: 

In order to deploy the operator, we need to be able to build it and add it in the ISO.

**Summary**:

Add new dependencies in the CI to build the operator (`go`, `operator-sdk` and `dep`).
Update the buildchain with new tasks:
- to build the operator image and add it to the ISO
- to lint the Go code (and check that the generated files are up-to-date)
- to help formatting Go code (using `gofmt`) and generating boilerplate code (using `operator-sdk generate`).

**Acceptance criteria**: 

When building the ISO, you should have the operator image in the build tree (check with `ls _build/root/images/storage-operator/*`)
If you're familiar with Go, you can also edit `storage-operator/pkg/apis/storage/v1alpha1/volume_types.go` to test the new tasks:
- `./doit.sh format` should reformat the code according to `gofmt`
- `./doit.sh codegen` should re-generate the boilerplate code
- `./doit.sh lint:go` should complains if either the formatting is not good or the generated files are outdated.

---

Closes: #1410

### DIFF
--- a/buildchain/buildchain/codegen.py
+++ b/buildchain/buildchain/codegen.py
@@ -1,0 +1,49 @@
+# coding: utf-8
+
+
+"""Tasks for code generation."""
+
+
+import shlex
+from typing import Callable, Iterator, Tuple
+
+import doit  # type: ignore
+
+from buildchain import config
+from buildchain import constants
+from buildchain import types
+from buildchain import utils
+
+def task_codegen() -> Iterator[types.TaskDict]:
+    """Run the code generation tools."""
+    for create_codegen_task in CODEGEN:
+        yield create_codegen_task()
+
+
+def codegen_go() -> types.TaskDict:
+    """Generate Go code using operator-sdk."""
+    cwd  = constants.STORAGE_OPERATOR_ROOT
+    actions = []
+    for target in ('k8s', 'openapi'):
+        cmd = ' '.join(map(shlex.quote, [
+            config.ExtCommand.OPERATOR_SDK.value, 'generate', target
+        ]))
+        actions.append(doit.action.CmdAction(cmd, cwd=cwd))
+
+    return {
+        'name': 'go',
+        'title': lambda task: utils.title_with_subtask_name('CODEGEN', task),
+        'doc': codegen_go.__doc__,
+        'actions': actions,
+        'task_dep': ['check_for:operator-sdk'],
+        'file_dep': list(constants.STORAGE_OPERATOR_SOURCES),
+    }
+
+
+# List of available code generation tasks.
+CODEGEN : Tuple[Callable[[], types.TaskDict], ...] = (
+    codegen_go,
+)
+
+
+__all__ = utils.export_only_tasks(__name__)
--- a/buildchain/buildchain/codegen.py
+++ b/buildchain/buildchain/codegen.py
@@ -32,7 +32,7 @@ def codegen_go() -> types.TaskDict:
 
     return {
         'name': 'go',
-        'title': lambda task: utils.title_with_subtask_name('CODEGEN', task),
+        'title': utils.title_with_subtask_name('CODEGEN'),
         'doc': codegen_go.__doc__,
         'actions': actions,
         'task_dep': ['check_for:operator-sdk'],
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -52,11 +52,12 @@ VAGRANT_UP_ARGS : Tuple[str, ...] = tuple(shlex.split(
 class ExtCommand(enum.Enum):
     """External commands used by the build chain."""
 
-    GIT      = os.getenv('GIT_BIN',      'git')
-    HARDLINK = os.getenv('HARDLINK_BIN', 'hardlink')
-    MKISOFS  = os.getenv('MKISOFS_BIN',  'mkisofs')
-    SKOPEO   = os.getenv('SKOPEO_BIN',   'skopeo')
-    VAGRANT  = os.getenv('VAGRANT_BIN',  'vagrant')
+    GIT          = os.getenv('GIT_BIN',          'git')
+    HARDLINK     = os.getenv('HARDLINK_BIN',     'hardlink')
+    MKISOFS      = os.getenv('MKISOFS_BIN',      'mkisofs')
+    SKOPEO       = os.getenv('SKOPEO_BIN',       'skopeo')
+    VAGRANT      = os.getenv('VAGRANT_BIN',      'vagrant')
+    OPERATOR_SDK = os.getenv('OPERATOR_SDK_BIN', 'operator-sdk')
 
     @property
     def command_name(self) -> str:
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -58,6 +58,7 @@ class ExtCommand(enum.Enum):
     SKOPEO       = os.getenv('SKOPEO_BIN',       'skopeo')
     VAGRANT      = os.getenv('VAGRANT_BIN',      'vagrant')
     OPERATOR_SDK = os.getenv('OPERATOR_SDK_BIN', 'operator-sdk')
+    GOFMT        = os.getenv('GOFMT_BIN',        'gofmt')
 
     @property
     def command_name(self) -> str:
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -63,6 +63,8 @@ class ExtCommand(enum.Enum):
         """Return the name of the command."""
         if self is self.OPERATOR_SDK:
             return 'operator-sdk'
+        # See https://github.com/PyCQA/pylint/issues/2062
+        # pylint: disable=no-member
         return self.name.lower()
 
 # }}}
--- a/buildchain/buildchain/config.py
+++ b/buildchain/buildchain/config.py
@@ -58,4 +58,11 @@ class ExtCommand(enum.Enum):
     SKOPEO   = os.getenv('SKOPEO_BIN',   'skopeo')
     VAGRANT  = os.getenv('VAGRANT_BIN',  'vagrant')
 
+    @property
+    def command_name(self) -> str:
+        """Return the name of the command."""
+        if self is self.OPERATOR_SDK:
+            return 'operator-sdk'
+        return self.name.lower()
+
 # }}}
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -106,12 +106,11 @@ GIT_REF = git_ref()
 
 # }}}
 
-# Only keep directories (except `vendor`) and top-level Go source files.
+# Only keep directories and top-level Go source files.
 STORAGE_OPERATOR_FMT_ARGS : FrozenSet[str] = frozenset([
     path.name for path in STORAGE_OPERATOR_ROOT.glob('*')
-    if (path.is_dir() and path.name != 'vendor') or path.suffix == '.go'
+    if path.is_dir() or path.suffix == '.go'
 ])
 STORAGE_OPERATOR_SOURCES : FrozenSet[Path] = frozenset([
     filepath for filepath in STORAGE_OPERATOR_ROOT.rglob('*.go')
-    if 'vendor' not in str(filepath.parent)
 ])
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -106,6 +106,11 @@ GIT_REF = git_ref()
 
 # }}}
 
+# Only keep directories (except `vendor`) and top-level Go source files.
+STORAGE_OPERATOR_FMT_ARGS : FrozenSet[str] = frozenset([
+    path.name for path in STORAGE_OPERATOR_ROOT.glob('*')
+    if (path.is_dir() and path.name != 'vendor') or path.suffix == '.go'
+])
 STORAGE_OPERATOR_SOURCES : FrozenSet[Path] = frozenset([
     filepath for filepath in STORAGE_OPERATOR_ROOT.rglob('*.go')
     if 'vendor' not in str(filepath.parent)
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -6,7 +6,7 @@
 
 from pathlib import Path
 import subprocess
-from typing import Optional
+from typing import Optional, FrozenSet
 
 from buildchain import ROOT  # Re-export ROOT through this module.
 from buildchain import config
@@ -39,6 +39,8 @@ VAGRANT_ROOT : Path = ROOT/'.vagrant'
 STATIC_CONTAINER_REGISTRY : Path = Path(
     ROOT, 'buildchain/static-container-registry'
 )
+# Path to the storage-operator source directory.
+STORAGE_OPERATOR_ROOT : Path = ROOT/'storage-operator'
 
 # }}}
 # Vagrant parameters {{{
@@ -103,3 +105,8 @@ def git_ref() -> Optional[str]:
 GIT_REF = git_ref()
 
 # }}}
+
+STORAGE_OPERATOR_SOURCES : FrozenSet[Path] = frozenset([
+    filepath for filepath in STORAGE_OPERATOR_ROOT.rglob('*.go')
+    if 'vendor' not in str(filepath.parent)
+])
--- a/buildchain/buildchain/format.py
+++ b/buildchain/buildchain/format.py
@@ -31,7 +31,7 @@ def format_go() -> types.TaskDict:
 
     return {
         'name': 'go',
-        'title': lambda task: utils.title_with_subtask_name('FORMAT', task),
+        'title': utils.title_with_subtask_name('FORMAT'),
         'doc': format_go.__doc__,
         'actions': [doit.action.CmdAction(cmd, cwd=cwd)],
         'task_dep': ['check_for:gofmt'],
--- a/buildchain/buildchain/format.py
+++ b/buildchain/buildchain/format.py
@@ -23,14 +23,10 @@ def task_format() -> Iterator[types.TaskDict]:
 
 def format_go() -> types.TaskDict:
     """Format Go code using gofmt."""
-    # Only keep directories (except `vendor`) and top-level Go source files.
-    targets = [
-        path.name for path in constants.STORAGE_OPERATOR_ROOT.glob('*')
-        if (path.is_dir() and path.name != 'vendor') or path.suffix == '.go'
-    ]
     cwd  = constants.STORAGE_OPERATOR_ROOT
     cmd = ' '.join(map(shlex.quote, [
-        config.ExtCommand.GOFMT.value, '-s', '-w', *targets
+        config.ExtCommand.GOFMT.value, '-s', '-w',
+        *tuple(constants.STORAGE_OPERATOR_FMT_ARGS)
     ]))
 
     return {
--- a/buildchain/buildchain/format.py
+++ b/buildchain/buildchain/format.py
@@ -1,0 +1,60 @@
+# coding: utf-8
+
+
+"""Tasks for code auto-formatting."""
+
+
+import shlex
+from typing import Callable, Iterator, Tuple
+
+import doit  # type: ignore
+
+from buildchain import config
+from buildchain import constants
+from buildchain import types
+from buildchain import utils
+
+
+def task_format() -> Iterator[types.TaskDict]:
+    """Run the code auto-formatting tools."""
+    for create_format_task in FORMATTERS:
+        yield create_format_task()
+
+
+def format_go() -> types.TaskDict:
+    """Format Go code using gofmt."""
+    # Only keep directories (except `vendor`) and top-level Go source files.
+    targets = [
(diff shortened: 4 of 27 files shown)


slaperche-scality:
Need to be sure it works on the CI (i.e.: green build) before removing the Draft label.
Also want to check if `dep` is really required, it may not be.

Other than that, you can start reviewing the code (lot of commit, but several of them are pretty small).

NicolasT:
We don't (or shouldn't, at least) rely on `dep`, we're using Go modules and no vendoring.

buildchain/buildchain/config.py
```diff
@@ -52,10 +52,21 @@
...
-    MKISOFS  = os.getenv('MKISOFS_BIN',  'mkisofs')
-    SKOPEO   = os.getenv('SKOPEO_BIN',   'skopeo')
-    VAGRANT  = os.getenv('VAGRANT_BIN',  'vagrant')
+    GIT          = os.getenv('GIT_BIN',          'git')
+    HARDLINK     = os.getenv('HARDLINK_BIN',     'hardlink')
+    MKISOFS      = os.getenv('MKISOFS_BIN',      'mkisofs')
+    SKOPEO       = os.getenv('SKOPEO_BIN',       'skopeo')
+    VAGRANT      = os.getenv('VAGRANT_BIN',      'vagrant')
+    OPERATOR_SDK = os.getenv('OPERATOR_SDK_BIN', 'operator-sdk')
+    GOFMT        = os.getenv('GOFMT_BIN',        'gofmt')
```
NicolasT, line 61: `go fmt` ?

buildchain/buildchain/constants.py
```diff
@@ -103,3 +105,13 @@ def git_ref() -> Optional[str]:
 GIT_REF = git_ref()
 
 # }}}
+
+# Only keep directories (except `vendor`) and top-level Go source files.
+STORAGE_OPERATOR_FMT_ARGS : FrozenSet[str] = frozenset([
+    path.name for path in STORAGE_OPERATOR_ROOT.glob('*')
+    if (path.is_dir() and path.name != 'vendor') or path.suffix == '.go'
```
NicolasT: There should be no `vendor` directory. If there is one, that's a 'bug'.

eve/workers/pod-builder/Dockerfile
```diff
@@ -26,4 +26,20 @@ RUN yum install -y epel-release \
     && chown eve:eve /home/eve/workspace \
     && pip install buildbot-worker==${BUILDBOT_VERSION}
 
+# Install Go tooling to build the Kubernetes Operators
+ENV GO_VERSION 1.12.7
+ENV OP_SDK_VERSION v0.9.0
```
NicolasT: I'd avoid using shorthands.

eve/workers/pod-builder/Dockerfile
```diff
@@ -26,4 +26,20 @@ RUN yum install -y epel-release \
...
+ENV GOCACHE /home/eve/workspace/.cache
+ENV PATH $GOPATH/bin:$GOROOT/bin:$PATH
+
+RUN curl -ORL https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz      \
+    && tar xzvf go${GO_VERSION}.linux-amd64.tar.gz                             \
+    && rm go${GO_VERSION}.linux-amd64.tar.gz                                   \
+    && mv go /usr/local                                                        \
+    && curl -RLo /usr/bin/operator-sdk                                         \
+        https://github.com/operator-framework/operator-sdk/releases/download/${OP_SDK_VERSION}/operator-sdk-${OP_SDK_VERSION}-x86_64-linux-gnu \
+    && chmod +x /usr/bin/operator-sdk
```
NicolasT, line 44: Consider checking the signature of the downloaded file?

buildchain/buildchain/config.py
slaperche-scality: I tried using `go fmt` but it only supports a subset of what `goftmt` does (no `-s` and more importantly, no `-d`)

```
% go help fmt
usage: go fmt [-n] [-x] [packages]

Fmt runs the command 'gofmt -l -w' on the packages named
by the import paths. It prints the names of the files that are modified.

For more about gofmt, see 'go doc cmd/gofmt'.
For more about specifying packages, see 'go help packages'.

The -n flag prints commands that would be executed.
The -x flag prints commands as they are executed.

To run gofmt with specific options, run gofmt itself.

See also: go fix, go vet.
% gofmt -h
usage: gofmt [flags] [path ...]
  -cpuprofile string
        write cpu profile to this file
  -d    display diffs instead of rewriting files
  -e    report all errors (not just the first 10 on different lines)
  -l    list files whose formatting differs from gofmt's
  -r string
        rewrite rule (e.g., 'a[b:len(a)] -> a[b:]')
  -s    simplify code
  -w    write result to (source) file instead of stdout
```
See also: https://stackoverflow.com/questions/55319495/what-is-the-difference-between-gofmt-go-fmt

Since both are available with a standard Go installation (`gofmt` isn't an extra-dependency), I went with `gofmt`.

buildchain/buildchain/constants.py
slaperche-scality: Ho, you're right!!
The one I had is a vestigial directory of when I was using an old `operator-sdk`.

I guess we could also remove the entry from the `.gitignore` then?

eve/workers/pod-builder/Dockerfile
slaperche-scality: :+1: 
I'll add it (anyway I have to modify this Dockerfile, things aren't working yet)

eve/workers/pod-builder/Dockerfile
slaperche-scality: Done

eve/workers/pod-builder/Dockerfile
slaperche-scality: That may be harder than expected, because I don't have the maintainers public key, so I have to download it but to get the ID I need to parse the error message from `gpg`

> To download the key, use the following command, replacing $KEY_ID with the RSA key string provided in the output of the previous command:

:/

buildchain/buildchain/constants.py
slaperche-scality: Done

slaperche-scality:
> We don't (or shouldn't, at least) rely on `dep`, we're using Go modules and no vendoring.

Yes, you're right: I got rid of it.

slaperche-scality:
A green build \o/, I'm passing this as ready for review then :slightly_smiling_face: 

That being said, since it's adding 20 min to the CI time (because of the retries) we probably want to fix that.
But "it works" so it can be reviewed…

NicolasT:
Adding 20min to CI is a no-go as far as I'm concerned. I believe there's a fairly easy fix though: we're running the build in a container, right? (If not, shouldn't we? ;-))

We can then apply the same 'trick' as we do for the doc building which downloads the Tox/Sphinx env, something along the lines of
```
WORKDIR /usr/src/storage-operator
COPY go.mod go.sum /usr/src/storage-operator/
RUN go mod download
COPY . /usr/src/storage-operator
RUN operator-sdk build ...
```

storage-operator/.gitignore
```diff
@@ -1,5 +1,3 @@
-vendor/
```
NicolasT: I wouldn't remove it, TBH. *Not* having it wouldn't prevent anyone from checking things into `vendor/`. Having it requires one to add `-f` to `git add`, otherwise Git won't be willing to touch files in that folder.

slaperche-scality: But how do they check things into `vendor`? The directory no longer exists when using 0.9.0.

slaperche-scality:
@bert-e reset

eve/workers/pod-builder/Dockerfile
slaperche-scality: Or should I hardcode the key when doing the `recv-keys`?

NicolasT: Ugh, maybe file a TODO ticket but leave it for now.

eve/workers/pod-linter/Dockerfile
```diff
@@ -20,8 +21,28 @@ RUN dnf install -y git \
   ShellCheck \
   && dnf clean all \
   && adduser -u 1042 --home /home/eve eve \
-  && chown eve:eve /home/eve/workspace \
+  && chown -R eve:eve /home/eve \
   && pip install tox \
   && pip install buildbot-worker==${BUILDBOT_VERSION}
 
+# Install Go tooling to lint the Kubernetes Operators.
```
NicolasT, line 30: Given we seem to be creating (almost) the same container image twice, does it really make sense to make a distinction between a `linter` and `builder` container?

storage-operator/.gitignore
NicolasT: `mkdir vendor`? `go mod vendor`?...

Boils down to: having `vendor/` in here doesn't do any harm, it might even provide a benefit, so I don't see any reason to remove it.

slaperche-scality: Fair enough, I'll add it back then.

But if we could have `vendor` here from `go mod vendor` I should keep the code to filter it then, no?

NicolasT: No, because we *shouldn't* have a `vendor` directory. This is a way to ensure we don't ever get one (unless one is very stubborn).

eve/workers/pod-builder/Dockerfile
slaperche-scality: Tracked by #1444.

storage-operator/.gitignore
slaperche-scality: Ok seen with you directly: will keep the `.gitignore` entry but remove the code

eve/workers/pod-linter/Dockerfile
slaperche-scality: I think the reason behind the split is that we want to use `centos:7` as base image (and that's what we do for the pod-builder).

Now, the issue with pod-linter is that we need a not too old `shellcheck`, which is not available on CentOS 7 so Jordi switched the base image to `fedora:29` for this pod (IIRC it was too cumbersome to get a recent `shellcheck` on CentOS)

That's the reason behind the split I think (keep using CentOS 7 everywhere, except for linting).

slaperche-scality:
@bert-e reset

slaperche-scality:
@bert-e reset

buildchain/buildchain/config.py
```diff
@@ -52,10 +52,21 @@
...
+
+    @property
+    def command_name(self) -> str:
+        """Return the name of the command."""
+        if self is self.OPERATOR_SDK:
```
gdemonet, line 66: Mmh, works but could have been simpler (any command name containing dashes will need this special case). On the other hand, the heuristic is quite simple:

```python
def command_name(self):
    return self.name.lower().replace('_', '-')
```

buildchain/buildchain/image.py
```diff
@@ -344,6 +344,13 @@ def show() -> str:
         },
         task_dep=['_image_mkdir_root'],
     ),
+    targets.OperatorImage(
+        name='storage-operator',
+        version='latest',
```
gdemonet, line 349: Don't we want the MetalK8s version here?

eve/workers/pod-builder/Dockerfile
```diff
@@ -19,11 +19,32 @@ RUN yum install -y epel-release \
...
     yum-utils \
     docker-ce-cli-18.09.6 \
     && adduser -u 1042 --home /home/eve eve --groups docker \
-    && chown eve:eve /home/eve/workspace \
+    && chown -R eve:eve /home/eve \
     && pip install buildbot-worker==${BUILDBOT_VERSION}
 
+# Install Go tooling to build the Kubernetes Operators.
+ENV GO_VERSION 1.12.7
+ENV OPERATOR_SDK_VERSION v0.9.0
```
gdemonet: Shouldn't these versions be set in `ARG` commands instead? Same as for `BUILDBOT_VERSION`

buildchain/buildchain/format.py
```diff
@@ -0,0 +1,48 @@
...
+        'name': 'go',
+        'title': lambda task: utils.title_with_subtask_name('FORMAT', task),
+        'doc': format_go.__doc__,
+        'actions': [doit.action.CmdAction(cmd, cwd=cwd)],
+        'task_dep': ['check_for:gofmt'],
+        'file_dep': list(constants.STORAGE_OPERATOR_SOURCES),
+    }
+
+
+# List of available formatting task.
```
gdemonet: minor:

```suggestion
# List of available formatting tasks.
```

buildchain/buildchain/utils.py
```diff
@@ -61,3 +61,19 @@ def title_with_target1(command: str, task: types.Task) -> str:
         cmd=command, width=constants.CMD_WIDTH,
         path=build_relpath(Path(task.targets[0])),
     )
+
+
+def title_with_subtask_name(command: str, task: types.Task) -> str:
```
gdemonet: minor: Could be interesting to have these `title_with_<something>` wrapped in another function, taking only the `command` argument, since we'll always need a `function(task: types.Task) -> str` to pass in the task dict.

buildchain/buildchain/lint.py
```diff
@@ -64,27 +75,86 @@ def lint_shell() -> types.TaskDict:
...
+    base = subprocess.check_output(git_diff)
+    for target in ('k8s', 'openapi'):
+        cmd = [config.ExtCommand.OPERATOR_SDK.value, 'generate', target]
+        subprocess.check_call(cmd, cwd=cwd)
+    current = subprocess.check_output(git_diff)
+    # If the diff changed after running the code generation that means that
+    # the generated
+    if current != base:
+        return doit.exceptions.TaskError(
+            msg='outdated generated Go files, did you run `doit.sh codegen:go`?'
```
gdemonet, line 128: minor: for consistency with `check_go_fmt`

```suggestion
            msg='outdated generated Go files, please run `doit.sh codegen:go`'
```

buildchain/buildchain/lint.py
```diff
@@ -64,27 +75,86 @@ def lint_shell() -> types.TaskDict:
...
+    """Check if the generated files are up to date."""
+    cwd  = constants.STORAGE_OPERATOR_ROOT
+    git_diff = [config.ExtCommand.GIT.value, 'diff']
+    base = subprocess.check_output(git_diff)
+    for target in ('k8s', 'openapi'):
+        cmd = [config.ExtCommand.OPERATOR_SDK.value, 'generate', target]
+        subprocess.check_call(cmd, cwd=cwd)
+    current = subprocess.check_output(git_diff)
+    # If the diff changed after running the code generation that means that
+    # the generated
```
gdemonet: This comment is missing an end I believe ;)

buildchain/buildchain/lint.py
```diff
@@ -64,27 +75,86 @@ def lint_shell() -> types.TaskDict:
...
+    if current != base:
```
gdemonet, line 126: Should we reset the changes in that situation?

eve/workers/pod-builder/Dockerfile
```diff
@@ -19,11 +19,32 @@ RUN yum install -y epel-release \
...
+    && rm go${GO_VERSION}.linux-amd64.tar.gz                                   \
+    && mv go /usr/local                                                        \
+    && curl -RLo /usr/bin/operator-sdk                                         \
+        https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu \
+    && chmod +x /usr/bin/operator-sdk
+
 USER eve
+
+# Pre-download the Go dependencies.
+COPY storage-operator/go.mod storage-operator/go.sum ./
```
gdemonet: Since you're only `COPY`ing from `storage-operator/`, maybe the build context could be just `'storage-operator/'` instead of `'.'`?

gdemonet (commented):
Few questions and minor comments, most of it looks good to me!

buildchain/buildchain/config.py
slaperche-scality: The problem is that some command does contains `_` in their name and that would break them.

True, we aren't using such command yet, but if one day we add one, I don't want the poor guy to go looking all over the buildchain trying to understand why the command he added doesn't work.

buildchain/buildchain/image.py
slaperche-scality: In the long term, we do want a real version number here (not latest).
For now: no, we will stay with latest.

Extract from a Slack discussion I had with Nicolas:

> Can use `latest` for now
> Until we figure out how to properly version `storage-operator`, also in its `version.go` file and such
> Can of worms

eve/workers/pod-builder/Dockerfile
slaperche-scality: Done

eve/workers/pod-builder/Dockerfile
slaperche-scality: Indeed. Done.

buildchain/buildchain/lint.py
slaperche-scality: If we reset, we could lose the existing diff of the user.

The buildchain should avoid to mess with the git state as much as possible (yes, here we are modifying files, but anyway the guy is gonna run `./doit.sh codegen` so it's not a big deal).

buildchain/buildchain/lint.py
slaperche-scality: Since the lint does run the codegen, the user don't really have to run it: we have done it for him already (would be better if `operator-sdk` had a dry-run mode, but that's not the case, so…)

buildchain/buildchain/format.py
slaperche-scality: Done (also fixed the same in `codegenpy`)

buildchain/buildchain/lint.py
slaperche-scality: Indeed >_<

Done

buildchain/buildchain/utils.py
slaperche-scality: Good idea!

Done in a dedicated commit.

slaperche-scality:
@bert-e reset

NicolasT:
It would make sense to hold off on this PR until #1447 is merged, then rebase it on `development/2.4` and change the target branch, so other version branches aren't impacted by the time spent on this in CI.

slaperche-scality:
@bert-e reset

buildchain/buildchain/codegen.py
```diff
@@ -0,0 +1,49 @@
...
+    """Run the code generation tools."""
+    for create_codegen_task in CODEGEN:
+        yield create_codegen_task()
+
+
+def codegen_go() -> types.TaskDict:
+    """Generate Go code using operator-sdk."""
+    cwd  = constants.STORAGE_OPERATOR_ROOT
+    actions = []
+    for target in ('k8s', 'openapi'):
```
NicolasT, line 27: To be really picky, this should be `['k8s', 'openapi']` (since it's a `list<str>` rather than a `(str, str)`), but let's not let that keep this PR unmerged :-P

storage-operator/.dockerignore
```diff
@@ -0,0 +1,5 @@
+cmd/
```
NicolasT, line 1: Not very fond we need this, but sure, for now. At some point I'd like to roll building of the whole thing in a container-based environment (indeed, maybe not using `operator-sdk build ...`), at which point this would need to be removed. Someday.

NicolasT (approved):

slaperche-scality:
@bert-e approve

buildchain/buildchain/codegen.py
gdemonet: Why? Tuples are lighter than lists

slaperche-scality:
@bert-e approve

storage-operator/.dockerignore
slaperche-scality: Yes I added it to only upload the strict minimum to the docker daemon.

But that looks like premature optimization (and I even broke `operator-sdk build` when I added `build/` to the ignore => proves that it wasn't a great idea ^^").

I won't be against its removal when it becomes an hindrance :slightly_smiling_face:

buildchain/buildchain/codegen.py
slaperche-scality: I used tuple for its immutability (since the list is kind of a constant here), but in this context tuple and list are really alike: won't make any (observable) difference.

In a statically typed language, tuples and lists are a different beasts (lists are homogeneous whereas tuples aren't), but in Python where you can mix apples and oranges in both lists and tuples the key difference is the mutability.

NicolasT: The remark came indeed from a 'static typing' background:

- A *list* is an *ordered* collection of an *unknown* number of values of a *single* type
- A *tuple* is an *ordered* collection of a *known* number of values of (potentially) *different* (but known) types

As such, the type constructor of a list has a single argument, whilst the type constructor of an N-tuple has N arguments.

The immutability of a tuple doesn't really play a role in the argument.

Anyway, pedantic, though in general I'm not overly fond of conflating the two (yes, despite duck-typing).

Issue #1204, opened 2019-06-11
Setting ssh key path when adding a node could be more intuitive


thomasdanan:
**Component**:

metalk8s-ui

**Why this is needed**:

It is easy to do mistakes when setting the ssh key path (when adding a node)

**What should be done**:

From what I understood, the path must always contain /etc/metalk8s/pki/ so may be this should be enforced so that user only choose the ssh key name (or even choose among available ssh key ?)

**Implementation proposal** (strongly recommended):

default value with the absolute path: /etc/metalk8s/pki/salt-bootstrap

**Test plan**:

ycointe:
Would it be possible to add a question mark icon to the right of the field which will explain that is the path where is the id_rsa file on the bootstrap node ?

thomasdanan:
@ycointe I think what you are suggesting is super important: we must have online help available in the UI ... I suggest you open an improvement issue to tackle this.
On this specific issue, I think the user should not even enter this path if this is the only path that can work ...

PR #1435, opened 2019-07-16
NodeCreationForm: usability improvement

### LINKED ISSUE #1226: Some field labels can be misleading while creating a node
**Component**:

documentation, ui, salt

**What happened**:

I was trying to add a new node and to extend the cluster to it.
I used "node1" as the name where my hostname was something different, and for the SSH key I was not sure if it was the private key path or the public key path.
Hence the expansion failed.

Even if it's detailed in the description of each field in the doc, labels are not precised enough in the UI nor in the documentation when you add manually a node using files and commands.

**What was expected**:

Detailed labels that can not mislead.

**Steps to reproduce**:

Use a random name (not matching the hostname) when adding a node, or provide the public key path instead of the private key path.

Change the labels:
* in the UI
  * `Name` could become `Hostname` or `Complete Hostname`
  * `Hostname or IP` could become `SSH Hostname or IP`
  * `SSH Key Path` could become `SSH Private Key Path`
  * Not related but `Sudo required` could become `Sudo Required`
* in the doc (and in the UI/salt code I guess because the metadata keys would be different)
  * `name` could become `hostname`
  * `metalk8s.scality.com/ssh-key-path` could become `metalk8s.scality.com/ssh-private-key-path`

Other option is to have distinct boxs for each set of info:
1 box: Node Name + roles
2 box: Ssh Connection infos (including sudo)
IMG_20190709_153626

carlito-scality:
**Component**: UI

**Context**: 
- Actually, the node creation form is quite confusing. We decided to split into 2 sections to make it clearer.
- SSH Port defaults to ```22```
- SSH Key Path defaults to ```/etc/metalk8s/pki/salt-bootstrap```
- Roles selection is improved
 
**Summary**:

**Acceptance criteria**: 
Screenshot 2019-07-16 at 15 06 35

Screenshot 2019-07-16 at 15 06 53

image

- no regression on node expansion
---

Closes: https://github.com/scality/metalk8s/issues/1226 https://github.com/scality/metalk8s/issues/1204

### DIFF
--- a/ui/src/containers/NodeCreateForm.js
+++ b/ui/src/containers/NodeCreateForm.js
@@ -98,7 +98,10 @@ const validationSchema = Yup.object().shape({
   version: Yup.string().required(),
   ssh_user: Yup.string().required(),
   hostName_ip: Yup.string().required(),
-  ssh_port: Yup.string().required(),
+  ssh_port: Yup.number()
+    .min(0)
+    .max(65535)
+    .required(),
   ssh_key_path: Yup.string().required(),
   sudo_required: Yup.boolean().required(),
   workload_plane: Yup.boolean().required(),
--- a/ui/src/containers/NodeCreateForm.js
+++ b/ui/src/containers/NodeCreateForm.js
@@ -5,8 +5,8 @@ import { Formik, Form } from 'formik';
 import * as Yup from 'yup';
 import { withRouter } from 'react-router-dom';
 import { injectIntl } from 'react-intl';
-import { Button, Input } from '@scality/core-ui';
-import { padding } from '@scality/core-ui/dist/style/theme';
+import { Button, Input, Checkbox } from '@scality/core-ui';
+import { padding, fontSize, gray } from '@scality/core-ui/dist/style/theme';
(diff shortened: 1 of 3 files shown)


gdemonet:
Stupid question: can the `SSH Port` input element enforce `type="number"` and `min=0`/`max=65535`? This would be more intuitive to experienced HTML users ;)

carlito-scality:
> Stupid question: can the `SSH Port` input element enforce `type="number"` and `min=0`/`max=65535`? This would be more intuitive to experienced HTML users ;)

yes. It's a good idea

thomasdanan (dismissed):
perfect ...
only the ssh key path is truncated, but nothing you can really do about that ...

carlito-scality:
> perfect ...
> only the ssh key path is truncated, but nothing you can really do about that ...

We will put a prefix ```/etc/metalk8s/pki``` for SSH key path in the next PR and it wanna solve this problem.

carlito-scality:
@bert-e reset

MonPote (approved):
LGTM

carlito-scality:
@bert-e approve

Issue #1394, opened 2019-07-09
Need to have MetalK8S with k8S 1.14 and 1.15


thomasdanan:
**Component**:

build

**Why this is needed**:

current k8s version is 1.15 and next deliveries of Metal should be on this version.
Also, we want ou new dev to be on this version and not on Metal2.0 which embed 1.11

**What should be done**:

- [x] Add 2 new branches 2.3 (for k8s 1.14) and 2.4 (for k8s 1.15)
- [x] Make sure MetalK8s.iso is generated for those new versions
- [x] Make `development/2.4` the default branch for the repository
- [x] Communicate to the team once branches are ready

**Implementation proposal** (strongly recommended):

**Test plan**:
No other tests than the CI autotests

Issue #640, opened 2019-02-22
De-duplicate image listing, versions,...


NicolasT:
It'd be really good to have this list somewhere else, where the build infrastructure also retrieves it from, etc.

_Originally posted by @NicolasT in https://github.com/scality/metalk8s/pull/604/review_comment/create_

jbertran:
This should wait for a build system refactor - then we can specify on build, image by image whether or not it should be added to the registry.

gdemonet:
Could be useful to be able to read this source of information from docs as well, since some expected outputs include versions (notably, K8s version).

gdemonet:
Done (see #1477).

PR #1443, opened 2019-07-17
Bump K8s version to 1.14

gdemonet:
**Component**: kubernetes

**Context**: Creation of the `development/2.3` branch

---

See: #1394


slaperche-scality:
You only need to bump k8s, not the others images (prometheus, etcd, …)?

Also commit message is missing (you can simply put "Creation of the development/2.3 branch" as body)

TeddyAndrieux:
Hmm here you bump version for `development/2.2`

Ebaneck:
> You only need to bump k8s, not the others images (prometheus, etcd, …)?
> 
> Also commit message is missing (you can simply put "Creation of the development/2.3 branch" as body)

@gdemonet just like @slaperche-scality mentioned... I think you will need to update the images in ` buildchain/buildchain/image.py`

NicolasT (changes requested):
- Need fix of target branch as @TeddyAndrieux mentioned
- Image hashes need to be updated as @Ebaneck mentioned
- `etcd` version to be checked with what `kubeadm` does in 1.14

gdemonet:
@slaperche-scality @Ebaneck  I'll have a look at `etcd`, but no need to bump Prometheus/Calico etc. for now
@NicolasT @TeddyAndrieux yes I misclicked apparently, fixed it

gdemonet:
@NicolasT `etcd` version was bumped in 1.13, not 1.14: https://github.com/kubernetes/kubernetes/commit/862b945e5ab33d7e15f9a3a624c53eb0371627b0
Should I create another PR?

There is also: https://github.com/kubernetes/kubernetes/blob/v1.14.4/cmd/kubeadm/app/constants/constants.go#L398-L403

```go
	// SupportedEtcdVersion lists officially supported etcd versions with corresponding Kubernetes releases
	SupportedEtcdVersion = map[uint8]string{
		12: "3.2.24",
		13: "3.2.24",
		14: "3.3.10",
	}
```

NicolasT:
> @NicolasT `etcd` version was bumped in 1.13, not 1.14: [kubernetes/kubernetes@862b945](https://github.com/kubernetes/kubernetes/commit/862b945e5ab33d7e15f9a3a624c53eb0371627b0)
> Should I create another PR?
> 
> There is also: https://github.com/kubernetes/kubernetes/blob/v1.14.4/cmd/kubeadm/app/constants/constants.go#L398-L403
> 
> ```go
> 	// SupportedEtcdVersion lists officially supported etcd versions with corresponding Kubernetes releases
> 	SupportedEtcdVersion = map[uint8]string{
> 		12: "3.2.24",
> 		13: "3.2.24",
> 		14: "3.3.10",
> 	}
> ```

From the second part, it appears 3.3.10 is supported by K8s 1.14, and K8s 1.12/1.13 use etcd 3.2.24. Your first link may seem to contradict this, saying the etcd version was updated to 3.3.10 for kubeadm 1.13, but this seems like a mistake in the commit message to me: according to Github, the relevant commit is only part of `v1.14.0-alpha.1` up to `v1.16.0-alpha.0`, i.e. K8s 1.14.

So, we'll want to properly deploy (and upgrade to) etcd 3.3.10 in MetalK8s 2.3, no?

thomasdanan (commented):
@gdemonet  did you test node expansion with this new version of k8s ? not working for me but maybe related to my env ... forget about it, I was using the wrong version (2.0.0 instead of 2.2) for metalk8s when adding the node ...

buildchain/buildchain/image.py
```diff
@@ -151,8 +151,8 @@ def show() -> str:
     targets.RemoteImage(
         registry=constants.GOOGLE_REGISTRY,
         name='etcd',
-        version='3.2.24',
-        digest='sha256:905d7ca17fd02bc24c0eba9a062753aba15db3e31422390bc3238eb762339b20',
+        version='3.3.10',
```
TeddyAndrieux, line 154: You need to change it in the manifest also :)

thomasdanan (dismissed):
actually, you probably need to update metalk8s version since in what I deployed it is still 2.2

gdemonet:
> /me throws in #640

Will do in a future PR ;)

NicolasT:
@thomasdanan is right, oversight. The `VERSION` file needs to be bumped as well, ideally before any other commits.

gdemonet:
@thomasdanan  @NicolasT `VERSION` file updated
@TeddyAndrieux `metalk8s.kubernetes.etcd.installed` updated

NicolasT (approved):
LGTM if CI agrees.

Actually, it may be worth to add a test as well, which validates (`kubectl version` or similar) the version of K8s that's running is the one we expect... May be worth another ticket, or done in scope of this PR.

gdemonet:
> LGTM if CI agrees.
> 
> Actually, it may be worth to add a test as well, which validates (`kubectl version` or similar) the version of K8s that's running is the one we expect... May be worth another ticket, or done in scope of this PR.

Good idea, I'll do it in this PR

TeddyAndrieux:
> > LGTM if CI agrees.
> > Actually, it may be worth to add a test as well, which validates (`kubectl version` or similar) the version of K8s that's running is the one we expect... May be worth another ticket, or done in scope of this PR.
> 
> Good idea, I'll do it in this PR

For me this test should be done on every branches so you can't do it in this PR.

gdemonet:
@bert-e approve

PR #1159, opened 2019-05-27
Deploy MetalLB and nginx-ingress

NicolasT:
**Component**: salt, kubernetes, metallb, ingress, nginx-ingress

**Summary**: deploy MetalLB and `nginx-ingress` as part of the cluster.

**Acceptance criteria**: 
- Deploy the cluster
- Find the `nginx-ingress-controller` `LoadBalancer` service (now managed by MetalLB):

```
$ kubectl -n nginx-ingress-system get svc
...
nginx-ingress-system   nginx-ingress-controller        LoadBalancer   10.101.134.174   172.21.254.47   80:30107/TCP,443:31177/TCP   6m
...
```

- Ensure the `metalk8s-ui` `Ingress` exists:

```
$  kubectl -n kube-system get ingress metalk8s-ui
NAME          HOSTS      ADDRESS   PORTS     AGE
metalk8s-ui   metalk8s             80        3h
```

- Either set up `/etc/hosts` for `metalk8s` to point to the LB address, or use `curl` and name resolution override to validate access to the admin UI service through the LB address using

```
$  curl http://metalk8s --resolve 'metalk8s:80:172.21.254.47'
...
```

Closes: #779
Closes: #791

PR #1445, opened 2019-07-18
Deploy `nginx-ingress`

NicolasT:
**Context**: we need an ingress controller running in the cluster

**Acceptance criteria**: requests to port 80 (HTTP) and port 443 (HTTPS) on a node workload-plane IP should return HTTP `502 Bad Gateway`. Access on the control-plane IP (if different) should fail.

Currently, the upstream Helm chart is rendered 'manually' and committed as-is, this should be merged in the buildchain.

For reviewers: better do this one commit-by-commit, because the 'import chart' and 'add deployment file' commits are large, but contain 100% downloaded or generated files.

Fixes: #791
See: #1159

### DIFF
--- a/buildchain/buildchain/image.py
+++ b/buildchain/buildchain/image.py
@@ -253,6 +253,23 @@ TO_PULL : Tuple[targets.RemoteImage, ...] = (
         destination=constants.ISO_IMAGE_ROOT,
         task_dep=['_image_mkdir_root'],
     ),
+    targets.RemoteImage(
+        registry='quay.io/kubernetes-ingress-controller',
+        name='nginx-ingress-controller',
+        version='0.25.0',
+        digest='sha256:464db4880861bd9d1e74e67a4a9c975a6e74c1e9968776d8d4cc73492a56dfa5',
+        destination=constants.ISO_IMAGE_ROOT,
+        task_dep=['_image_mkdir_root'],
+    ),
+    targets.RemoteImage(
+        registry=constants.GOOGLE_REGISTRY,
+        remote_name='defaultbackend-amd64',
+        name='nginx-ingress-defaultbackend-amd64',
+        version='1.5',
+        digest='sha256:4dc5e07c8ca4e23bddb3153737d7b8c556e5fb2f29c4558b7cd6e6df99c512c7',
+        destination=constants.ISO_IMAGE_ROOT,
+        task_dep=['_image_mkdir_root'],
+    ),
 )
 
 
--- a/charts/render.py
+++ b/charts/render.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+
+'''
+This script takes a Helm release name, a namespace name, a Helm chart
+`values` file and a chart, and turns it into a YAML document that can be
+deployed in a Kubernetes cluster as part of MetalK8s.
+
+It performs the following tasks:
+
+- Run `helm template` to render the chart, passing in the values provided on the
+  command-line
+- Fix up the resulting objects to include the desired namespace (which is not
+  always part of chart templates) in the metadata section
+- Fix up the resulting objects labels and annotations replace their `Tiller`
+  heritage by `metalk8s`, set the `app.kubernetes.io/part-of` and
+  `app.kubernetes.io/managed-by` to `metalk8s`, and copy any `app` and
+  `component` fields to the canonical `app.kubernetes.io/name` and
+  `app.kubernetes.io/component` fields
+'''
+
+import sys
+import subprocess
+
+import yaml
+
+
+BOILERPLATE = '''
+#!jinja | kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+{%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+'''
+
+
+def fixup_metadata(namespace, doc):
+    if 'metadata' in doc and 'namespace' not in doc['metadata']:
+        doc['metadata']['namespace'] = namespace
+
+    return doc
+
+
+def maybe_copy(doc, src, dest):
+    try:
+        doc[dest] = doc[src]
+    except KeyError:
+        pass
+
+
+def fixup_dict(doc):
+    if doc.get('heritage') == 'Tiller':
+        maybe_copy(doc, 'app', 'app.kubernetes.io/name')
+        maybe_copy(doc, 'component', 'app.kubernetes.io/component')
+
+        doc['heritage'] = 'metalk8s'
+        doc['app.kubernetes.io/part-of'] = 'metalk8s'
+        doc['app.kubernetes.io/managed-by'] = 'metalk8s'
+
+    return dict((key, fixup_doc(value)) for (key, value) in doc.items())
+
+
+def fixup_doc(doc):
+    if isinstance(doc, dict):
+        return fixup_dict(doc)
+    elif isinstance(doc, list):
+        return [fixup_doc(d) for d in doc]
+    else:
+        return doc
+
+
+def main():
+    (name, namespace, values, path) = sys.argv[1:]
+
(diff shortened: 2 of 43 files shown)


NicolasT:
I'll add a minimal test for this as well.

buildchain/buildchain/image.py
```diff
@@ -253,6 +253,23 @@ def show() -> str:
         destination=constants.ISO_IMAGE_ROOT,
         task_dep=['_image_mkdir_root'],
     ),
+    targets.RemoteImage(
+        registry='quay.io/kubernetes-ingress-controller',
+        name='nginx-ingress-controller',
+        version='0.25.0',
+		digest='sha256:464db4880861bd9d1e74e67a4a9c975a6e74c1e9968776d8d4cc73492a56dfa5',
```
slaperche-scality: Bad indent (tab instead of spaces I guess)

charts/render.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
slaperche-scality, line 1: > This is a short-term hack which should be integrated in the proper buildchain.

Do we have an issue to track that?

NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

NicolasT:
@bert-e approve

charts/render.py
NicolasT: #1449

charts/render.py
```diff
@@ -0,0 +1,79 @@
...
+def fixup_doc(doc):
+    if isinstance(doc, dict):
+        return fixup_dict(doc)
+    elif isinstance(doc, list):
+        return [fixup_doc(d) for d in doc]
+    else:
+        return doc
+
+
+def main():
```
slaperche-scality, line 68: I know it's hack and as such is not supposed to stay that way for a long time, that being said I think it would be nice to have some docstring/comments (or even in the commit message) explaining what is done here.

I mean, I get the big picture (we "fix" the output of `helm`), but not the details: What are we fixing? Why are we fixing it? What's so special about `Tiller` heritage? …

Understanding the script will help the one who will be tasked to rewrite or integrate the things in the build chain later :)

tests/post/steps/test_ingress.py
```diff
@@ -0,0 +1,101 @@
...
+
+import requests
+import requests.exceptions
+
+import pytest
+from pytest_bdd import given, parsers, scenario, then, when
+
+
+@pytest.fixture(scope='function')
+def context():
```
slaperche-scality: Looks like we're trying to reimplement `behave` (and its `context` object) here ^^

tests/post/steps/test_ingress.py
```diff
@@ -0,0 +1,101 @@
...
+    pass
+
+
+@scenario(
+    '../features/ingress.feature', 'Access HTTP services on control-plane IP')
+def test_access_http_services_on_control_plane_ip(host):
+    pass
+
+
+@given("the node control-plane IP is not equal to its workload-plane IP")
```
slaperche-scality: Why the double quote here? (you used simple quote for most of the other strings)

tests/post/steps/test_ingress.py
```diff
@@ -0,0 +1,101 @@
...
+        'control-plane': 'metalk8s:control_plane_ip',
+    }
+
+    if plane not in grains:
+        raise NotImplementedError
+
+    with host.sudo():
+        ip_output = host.check_output(' '.join([
+            'salt-call --local',
+            'grains.get {grain}'.format(grain=grains[plane]),
```
slaperche-scality, line 71: minor: could be written `'grains.get {grains[plane]}'.format(grains=grains)` if you prefer.

Code is fine as-is, just mentioning the possibility.

tests/post/steps/test_ingress.py
```diff
@@ -0,0 +1,101 @@
...
+            '--out json',
+        ]))
+    ip = json.loads(ip_output)['local']
+
+    protocols = {
```
slaperche-scality: If `protocols` is declared at the beginning of the function, you could rewrite the test l.52 as `if protocol not in protocols:`.

That would avoid the need to duplicate the list of supported protocols.

Edit: in fact the test l.52 seems useless as you are already doing what I suggest below.

slaperche-scality (changes requested):
Most of the comments are minor, the only one that need change is the duplicated test in `perform_request` (except if I missed something), and maybe documenting the small Python script also.

tests/post/steps/test_ingress.py
NicolasT: This pattern is used in other places as well. The only way I know to pass data from one step to another in pytest-bdd... It's ugly but :man_shrugging:

gdemonet: Givens/whens/thens can return a value to be stored in an arbitrary fixture for later consumption. It's done in the static Pod test, with `@given("I have set up a static Pod", target_fixture="static_pod_id")`.

Reference: https://github.com/pytest-dev/pytest-bdd/blob/master/README.rst#override-fixtures-via-given-steps

charts/render.py
```diff
@@ -0,0 +1,79 @@
...
+
+def fixup_metadata(namespace, doc):
+    if 'metadata' in doc and 'namespace' not in doc['metadata']:
+        doc['metadata']['namespace'] = namespace
+
+    return doc
+
+
+def maybe_copy(doc, src, dest):
+    obj = object()
```
gdemonet: I was confused by this pattern when first reading it :s IIUC, you use a unique object to check identity of the fallback for getting the `src` key in `doc`.

I think a more _Pythonic_ implementation would look like:

```python
def maybe_copy(doc, src, dest):
    try:
        doc[dest] = doc[src]
    except KeyError:
        pass
```

It's a temporary hack anyway, so don't bother to change the code, it was just FYI :)

salt/metalk8s/addons/nginx-ingress/deployed/namespace.sls
```diff
@@ -0,0 +1,10 @@
+#! kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: metalk8s-ingress
```
gdemonet, line 6: So, we decided for multiple namespaces? Worth opening a tracking ticket to apply the same strategy for other services we deploy IMO

tests/post/steps/test_ingress.py
gdemonet: Or just use `protocol.lower()` instead of this mapping and keep the test l.52?

gdemonet (dismissed):
LGTM!

salt/metalk8s/addons/nginx-ingress/deployed/namespace.sls
NicolasT: Ingress doesn't belong in `kube-system`, it's not a cluster-critical service. Neither do `monitoring` and such. However, I'd prefer the `monitoring` namespace to be (at least) prefixed by `metalk8s-` as well for consistency, and namespacing of namespaces.

tests/post/steps/test_ingress.py
NicolasT: Didn't know, but not sure I like it either TBH. Rather pass minimal context into the template string rather than more than strictly required.

NicolasT:
@bert-e reset

NicolasT:
I believe all comments are handled appropriately @gdemonet @slaperche-scality. Will re-run tests locally but need to redeploy Vagrant env first.

NicolasT:
@bert-e reset

NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #1467, opened 2019-07-23
Node Management: use cluster version to deploy

### LINKED ISSUE #1252: Node Management: saltenv metalk8s version
**Component**: kubernetes ?!

**Why this is needed**:
- Actually,  the metalk8s version set by user when creating a node is used in the ```saltenv```for the node deployment through the UI. This should be the 'version of the cluster'. But we don't have such things as 'version of the cluster' stored anywhere at this point in time.

**What should be done**:
- Store the 'version of the cluster' somewhere (#1440)
- Use this version in the UI for the Salt deployment

We can rely on the annotation `metalk8s.scality.com/cluster-version` of the `kube-system` Namespace for storing the cluster version.

**Test plan**:

carlito-scality:
**Component**: UI

**Context**: 
- Actually, the metalk8s version set by user when creating a node is used in the saltenvfor the node deployment through the UI. This should be the 'version of the cluster'. But we don't have such things as 'version of the cluster' stored anywhere at this point in time.

**Summary**:
- Retrieve the cluster version using the annotation of the namespace 'kube-system'
- Display cluster version the the About page and NodeCreationForm

**Acceptance criteria**: 
- No regression when deploying a node
- cluster version is used when creating a node

---

Closes: https://github.com/scality/metalk8s/issues/1252


carlito-scality:
close since the name branch is incorrect. see https://github.com/scality/metalk8s/pull/1468

Issue #947, opened 2019-04-05
Review and correct references to docker terms in buildchain


jbertran:
In the buildchain image, local_image and remote_image, we define and use tags, versions, repositories and registries in a manner that is not entirely accurate or consistent.

This needs to be fixed, for the sanity of future buildchain maintainers.

_Originally posted by @slaperche-scality in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjIyNzM4NDQ4/pull_request_reviews/more_threads_

Issue #1455, opened 2019-07-19
Package manager list wrong deps


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**What happened**:

If several version of a package is available in our repos then the module `metalk8s_package_manager.list_pkg_deps` will take the wrong one

```
[root@bootstrap ~]# repoquery --whatrequires --recursive --qf '%{NAME} %{VERSION}-%{RELEASE}' kubelet-1.12.9 --disablerepo '*' --enablerepo 'metalk8s-*'
calico-cni-plugin 3.8.0-1.el7
kubelet 1.12.9-0
kubelet 1.13.7-0
kubernetes-cni 0.7.5-0
[root@bootstrap ~]# salt-call metalk8s_package_manager.list_pkg_deps kubelet 1.12.9 fromrepo="metalk8s-*"
local:
    ----------
    calico-cni-plugin:
        3.8.0-1.el7
    kubelet:
        1.13.7-0
    kubernetes-cni:
        0.7.5-0
```

**What was expected**:

Get the right kubelet version.

**Steps to reproduce**

Configure repo to have several version of kubelet (or another package) available. 

**Resolution proposal** (optional):

Does not use `--recursive` in the module and do the recursive search manually filtering out package already found.

PR #1477, opened 2019-07-25
Define a single source of version information

gdemonet:
**Component**: build, salt, kubernetes, tests

**Context**:

Version information is currently hardcoded in multiple places (buildchain, Salt formulas, docs), making it difficult when some of them need to be changed.

**Summary**:

- New single source in `buildchain/buildchain/versions.py`
- Generated `salt/metalk8s/versions.json` containing versions for packages and container images
- Adapted macros for `pkg_installed` and `build_image_name`
- New test scenario, to verify the running version of the cluster matches configuration
- Small clean-up of the Docker client usage in the buildchain (see #947)
- [Not yet included in this draft PR] Fix of `metalk8s_package_manager` (see #1455)

**Acceptance criteria**: 

- Everything should keep working as expected
- Integration branches should only need to change the `buildchain/buildchain/versions.py` file

---

Closes: #640
Closes: #947


salt/metalk8s/macro.sls
```diff
@@ -1,14 +1,19 @@
...
     - fromrepo: {{ repo.repositories.keys() | join(',') }}
     {%- if package.version | default(None) %}
     - version: {{ package.version }}
     - hold: True
     - update_holds: True
+    - ignore_epoch: True
     {%- endif %}
     - reload_modules: True
+  {%- else %}
+  {{ raise('Missing information for package "' ~ name ~ '"') }}
```
TeddyAndrieux, line 18: Maybe better to use do for that
```suggestion
  {%- do raise('Missing information for package "' ~ name ~ '"') %}
```

gdemonet: No no, this usage is documented: https://docs.saltstack.com/en/latest/topics/jinja/index.html#errors

TeddyAndrieux: Oh ! Ok

.pylintrc
```diff
@@ -42,11 +42,12 @@ confidence=
...
-        bad-whitespace,     # Alignment is a good thing.
-        bad-continuation,   # I kindly disagree :)
-        locally-disabled,   # Sometime we need to disable a warning locally.
-        suppressed-message  # Don't pollute the output with useless info.
+disable=abstract-method,     # Too many false positive :/
+        bad-whitespace,      # Alignment is a good thing.
+        bad-continuation,    # I kindly disagree :)
+        locally-disabled,    # Sometime we need to disable a warning locally.
+        suppressed-message,  # Don't pollute the output with useless info.
+        useless-return       # Disagreement with mypy, see https://github.com/python/mypy/issues/3974
```
slaperche-scality, line 50: Hum, I disagree with that change.

I think for the `Optional` case it doesn't hurt to be explicit about returning `None`/
If you took the time to annotate the code with "I will return a value OR `None`", then take the time to return the `None`.

And as GvR commented in the issue: it's about the style recommended in PEP8:

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable).

buildchain/buildchain/constants.py
```diff
@@ -17,11 +17,13 @@
 CMD_WIDTH : int = 12
 
 # URLs of the main container repositories.
-GOOGLE_REGISTRY : str = 'k8s.gcr.io'
-DOCKER_REGISTRY : str = 'docker.io/library'
-COREOS_REGISTRY : str = 'quay.io/coreos'
-PROMETHEUS_REGISTRY : str = 'quay.io/prometheus'
-GRAFANA_REGISTRY : str = 'docker.io/grafana'
+CALICO_REPOSITORY     : str = 'quay.io/calico'
```
slaperche-scality, line 20: minor: since you are modifying each line, you could use the occasion to order them by alphabetical order.

buildchain/buildchain/docker_command.py
```diff
@@ -141,13 +141,14 @@ def __init__(
     @task_error(docker.errors.BuildError, handler=build_error_handler)
     @task_error(docker.errors.APIError)
     def __call__(self) -> Optional[TaskError]:
-        return DOCKER_CLIENT.images.build(
+        DOCKER_CLIENT.images.build(
```
slaperche-scality, line 144: I don't understand the relation between "this is a patch about renaming" and "let's change the return value of the code"?

buildchain/buildchain/targets/remote_image.py
```diff
@@ -62,39 +61,18 @@ def __init__(
...
-        """Complete image name.
-
-        Usable by `docker` commands.
-        """
-        return '{obj.registry}/{obj._remote_name}@{obj.digest}'.format(
-            obj=self
-        )
+    def remote_fullname(self) -> str:
+        """Complete image name retrieved from the remote repository."""
+        return "{i.repository}/{i._remote_name}:{i.version}".format(i=self)
```
slaperche-scality: `i`? Could use `img` at least.

Same remarks for the occurrences below.

buildchain/buildchain/targets/remote_image.py
```diff
@@ -150,12 +131,14 @@ def _skopeo_copy(self) -> List[str]:
...
             cmd.append('--dest-compress')
-        cmd.append('docker://{}'.format(self.fullname))
+        cmd.append('docker://{}'.format(self.remote_fullname))
         cmd.append(self._skopeo_dest())
         return cmd
 
     def _skopeo_dest(self) -> str:
         """Return the destination, formatted for skopeo copy."""
         if self._use_tar:
+            # NOTE: unused, since `docker-archive` format doesn't work with
```
slaperche-scality: Ha good catch, we forgot to remove this code when we reverted back to using Docker.

I think the code can be removed in a dedicated commit (with your comment as commit message, and maybe a link to the related issues).

buildchain/buildchain/image.py
```diff
@@ -89,244 +90,127 @@ def show() -> str:
     ]
     return task
 
+# Helpers
+def _get_image_info(name: str) -> versions.Image:
+    try:
+        return versions.CONTAINER_IMAGES_MAP[name]
+    except KeyError:
+        raise Exception(
```
slaperche-scality: Could be a `ValueError` (the `name` was wrong) I guess.

buildchain/buildchain/image.py
```diff
@@ -89,244 +90,127 @@ def show() -> str:
...
+            'Missing version for container image "{}"'.format(name)
+        )
 
-NGINX_IMAGE_VERSION : str = '1.15.8'
-NODE_IMAGE_VERSION : str = '10.16.0'
+def _remote_image(name: str, repository: str, **overrides: Any) \
```
slaperche-scality: I prefer this wrapping:

```python
def _remote_image(
    name: str, repository: str, **overrides: Any
) -> targets.RemoteImage:
```

It avoids the trailing `\` and is consistent with the layout used elsewhere in the build chain.

buildchain/buildchain/image.py
```diff
@@ -89,244 +90,127 @@ def show() -> str:
     ]
     return task
 
+# Helpers
```
slaperche-scality: Some basic docstrings would be good for those helpers.

minor: also I guess the `Helpers`, `Container images to pull` (`List of` is superfluous) and `Container images to build` could be folds.

buildchain/buildchain/packaging.py
```diff
@@ -142,52 +139,75 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
...
+        constants.ROOT/'packages/yum_repositories/saltstack.repo'  # TODO: template this with versions.SALT_VERSION
     ],
 )
 
+# Packages to build, per repository.
+def _package(name: str, sources: List[Path]) -> targets.Package:
+    try:
+        pkg_info = versions.PACKAGES_MAP[name]
+    except KeyError:
+        raise Exception(
```
slaperche-scality: Same remark as in previous commit: I think a `ValueError` could make sense here and be more specific than `Exception`.

buildchain/buildchain/packaging.py
```diff
@@ -142,52 +139,75 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
...
+            'Missing version for package "{}"'.format(name)
+        )
+
+    try:
```
slaperche-scality: Could this `try`/`except` be replaced by: `build_id_str, _, _ = pkg_info.release.partition('.')`?

e.g.:

```
>>> no_dot = 'foo'
>>> one_dot = 'foo.bar'
>>> more_dot = 'foo.bar.baz'
>>> res, _ = no_dot.split('.', 1); print(res)
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 1)
>>> res, _ = one_dot.split('.', 1); print(res)
foo
>>> res, _ = more_dot.split('.', 1); print(res)
foo
>>> res, _, _ = no_dot.partition('.'); print(res)
foo
>>> res, _, _ = one_dot.partition('.'); print(res)
foo
>>> res, _, _ = more_dot.partition('.'); print(res)
foo
```

buildchain/buildchain/packaging.py
```diff
@@ -142,52 +139,75 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
...
         ),
     ),
 }
 
+_TO_BUILD_PKG_NAMES : FrozenSet[str] = frozenset(
+    pkg.name for pkgs in TO_BUILD.values() for pkg in pkgs
+)
+
+# All packages not referenced in `TO_BUILD` but listed in `versions.PACKAGES`
+# is supposed to be downloaded.
```
slaperche-scality: `are supposed`

buildchain/buildchain/packaging.py
```diff
@@ -142,52 +139,75 @@ def task__build_repositories() -> Iterator[types.TaskDict]:
...
+                Path('v{}.tar.gz'.format(versions.CALICO_VERSION)),
             ],
-            builder=BUILDER,
-            task_dep=['_package_mkdir_root', '_build_container'],
         ),
     ),
 }
 
+_TO_BUILD_PKG_NAMES : FrozenSet[str] = frozenset(
+    pkg.name for pkgs in TO_BUILD.values() for pkg in pkgs
```
slaperche-scality: nested loop in comprehension are not very readable :S

buildchain/buildchain/salt_tree.py
```diff
@@ -29,8 +29,8 @@
 
 
 import importlib
-import sys
 from pathlib import Path
+import sys
```
slaperche-scality, line 33: About commit message: why JSON?

I guess it's because you don't need an extra-deps (unlike YAML), but it can be good to mention if someone does archeology and ever wonder why JSON was chosen instead of YAML.

buildchain/buildchain/salt_tree.py
```diff
@@ -180,6 +180,24 @@ def _run(self) -> None:
         file_dep=[versions.VERSION_FILE],
     ),
 
+    targets.DataFile(
```
slaperche-scality: `SerializedObject` (or `SerializedData`) instead of `DataFile`.

buildchain/buildchain/targets/data_file.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Any, Callable
+
+from buildchain import types
+from buildchain import utils
+
+from . import base
+
+
+def render_json(obj: types.JSON, filepath: Path) -> None:
+    """Serialize a dict as JSON to a given file path."""
```
slaperche-scality: `a dict`: not only according to the type signature ;)

buildchain/buildchain/targets/data_file.py
```diff
@@ -0,0 +1,79 @@
...
+    with filepath.open('w', encoding='utf-8') as file_obj:
+        json.dump(obj, file_obj, sort_keys=True, indent=2)
+
+
+class DataFile(base.AtomicTarget):
```
slaperche-scality: `SerializedObject` (or `SerializedData`) instead of `DataFile`.

(the docstring seems to agree with me :p)

buildchain/buildchain/types.py
```diff
@@ -31,3 +31,10 @@
 
 # A docker mount
 Mount = docker.types.Mount
+
+# A JSON-serializable data structure (almost)
+# NOTE: recursive type definitions are not supported, otherwise a better
+#       definition would have been:
+# pylint: disable=wrong-spelling-in-comment
+#       Union[str, int, float, bool, None, Dict[str, 'JSON'], List['JSON']]
+JSON = Union[str, int, float, bool, None, Dict[str, Any], List[Any]]
```
slaperche-scality: Hum, why bother?

I would go for `Any`, ok it's not accurate but it's simpler.

I mean, we already can't accurately encode the type "JSON-encodable", so let's keep things simple :)

And even now we can have `NaN` in `float` and this is not supported by JSON (Python allows it by default though)

buildchain/buildchain/targets/data_file.py
```diff
@@ -0,0 +1,79 @@
...
+        Keyword Arguments:
+            They are passed to `Target` init method
+        """
+        kwargs['targets'] = [destination]
+        super().__init__(**kwargs)
+
+        self._data = data
+        self._dest = destination
+
+        assert renderer in self.RENDERERS, (
```
slaperche-scality: An enum would be better than a string here, the set of supported rendered is known, finite and small.

slaperche-scality (changes requested):
Posting that before I forget…

I mainly reviewed the buildchain, also took a glance at Salt stuff but `Fix package manager listing of requires` would certainly benefit from a deeper review.

.pylintrc
gdemonet: Well, I agree with you and Guido, but `pylint` doesn't (complaining with `useless-return`), hence the disable here.

buildchain/buildchain/constants.py
gdemonet: Oops I misplaced `INGRESS_...`

buildchain/buildchain/docker_command.py
gdemonet: I roughly stashed all my changes to `docker_command` in this commit :/ I'll try to split it

buildchain/buildchain/targets/remote_image.py
gdemonet: Yeaaahhh, line length made `i` more appealing, but w/e

buildchain/buildchain/packaging.py
gdemonet: Oh, really? I like it actually, but if you prefer I can do something like:

```python
_TO_BUILD_PKG_NAMES : List[str] = []

for pkgs in TO_BUILD.values():
    for pkg in pkgs:
        _TO_BUILD_PKG_NAMES.append(pkg.name)
```

buildchain/buildchain/types.py
gdemonet: I'll go with `Any`

.pylintrc
slaperche-scality: Ho right, I went to read the issue and thought we wanted to get rid of a warning about "missing" return, not about "useless" one.

Then it's good of course >_<

That being said we should probably link this issue instead: https://github.com/PyCQA/pylint/issues/2300
But it seems to be fixed…

On which code did you had the warning?
Either Pylint was right (maybe you had an if/else with a return in both branches already, so the last return was really useless) or we may have a minimal example to reopen the bug.

buildchain/buildchain/packaging.py
slaperche-scality: Or maybe split the line as

```python
_TO_BUILD_PKG_NAMES : FrozenSet[str] = frozenset(
    pkg.name
    for pkgs in TO_BUILD.values()
    for pkg in pkgs
)
```
It's a bit more readable like that.

salt/_modules/metalk8s_package_manager.py
```diff
@@ -15,64 +14,133 @@ def __virtual__():
...
+
+def list_pkg_dependents(
+    name, version=None, fromrepo=None, pkgs_info=None
+):
+    '''
+    Check dependents of the package `name`-`version` to install, to add in a
+    later `pkg.installed` state along with the original package.
+
+
+
```
TeddyAndrieux: Maybe too many empty line here :stuck_out_tongue:

TeddyAndrieux (dismissed):
LGTM: Only one minor comment

.pylintrc
slaperche-scality, line 50: Going back to that, I still think it's not needed.

I think we may have messed up our type annotations.

Seeing the examples on this issue https://github.com/python/mypy/issues/3157, one can see that the decorated function must be annotated with what it really returns (so if our function returns nothing, the return type must be `None`), because the decorator can have its own signature.

So in our case:
- `__call__` always return `None`¹
- the decorator returns a function that return `Optional[TaskError]`

With the following diff, both `pylint` and `mypy` are happy and we don't need to disable anything.

```diff
diff --git a/.pylintrc b/.pylintrc
index ce7a5299..c4aae53d 100644
--- a/.pylintrc
+++ b/.pylintrc
@@ -47,7 +47,6 @@ disable=abstract-method,     # Too many false positive :/
         bad-continuation,    # I kindly disagree :)
         locally-disabled,    # Sometime we need to disable a warning locally.
         suppressed-message,  # Don't pollute the output with useless info.
-        useless-return       # Disagreement with mypy, see https://github.com/python/mypy/issues/3974
 
 # }}}
 # Basic {{{
diff --git a/buildchain/buildchain/docker_command.py b/buildchain/buildchain/docker_command.py
index 56e53c71..4cf26444 100644
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -140,7 +140,7 @@ class DockerBuild:
 
     @task_error(docker.errors.BuildError, handler=build_error_handler)
     @task_error(docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
         DOCKER_CLIENT.images.build(
             tag=self.tag,
             path=self.path,
@@ -148,7 +148,6 @@ class DockerBuild:
             buildargs=self.buildargs,
             forcerm=True,
         )
-        return None
 
 
 class DockerRun:
@@ -266,14 +265,13 @@ class DockerRun:
     @task_error(docker.errors.ContainerError, handler=container_error_handler)
     @task_error(docker.errors.ImageNotFound)
     @task_error(docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
         run_config = self.expand_config()
         DOCKER_CLIENT.containers.run(
             image=self.builder.tag,
             command=self.command,
             **run_config
         )
-        return None
 
 
 class DockerTag:
@@ -294,10 +292,9 @@ class DockerTag:
 
     @task_error(docker.errors.BuildError, handler=build_error_handler)
     @task_error(docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
         to_tag = DOCKER_CLIENT.images.get(self.full_name)
         to_tag.tag(self.repository, tag=self.version)
-        return None
 
 
 class DockerPull:
@@ -338,7 +335,6 @@ class DockerPull:
                     s=self, observed_digest=pulled.id
                 )
             )
-
         return None
 
 
@@ -358,12 +354,10 @@ class DockerSave:
 
     @task_error(docker.errors.APIError)
     @task_error(OSError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
         to_save = DOCKER_CLIENT.images.get(self.tag)
         image_stream = to_save.save(named=True)
 
         with self.save_path.open('wb') as image_file:
             for chunk in image_stream:
                 image_file.write(chunk)
-
-        return None
```

¹: except the one you modified to check the digest that explicitly returns a task error (we may want to raise an exception instead and use `task_error` for that, to be consistent).

slaperche-scality (changes requested):
LGTM except for one comment.

slaperche-scality (approved):

gdemonet:
@bert-e approve

PR #1478, opened 2019-07-25
Improvement/add babel plugin without ejecting cra

MonPote:
**Component**: ui

**Context**: 
Some APIs like `api-server` or `salt-api` response are usually very big and nested object.
It add a lot of repetitive checking that do not add any value to the code.

For example if you want to get the id of the url instead of 
```javascript
if (props && props.match && props.match.params && props.match.params.id) {
  console.log(props.match.params.id);
}
```
You have something like this
```javascript
if (props?.match?.params?.id) {
  console.log(props.match.params.id);
}
```
**Summary**:
This PR :
- Update create react app 2 to 3.
- Add optional-chaining plugin

**Acceptance criteria**:
You will need to do an `npm install` before launching the app.
Just make sure you can launch the app.

### DIFF
--- a/ui/.babelrc
+++ b/ui/.babelrc
@@ -1,0 +1,5 @@
+{
+  "plugins":[
+    ["@babel/plugin-proposal-optional-chaining", { "loose": false }],
+  ]
+}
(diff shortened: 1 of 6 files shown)


MonPote:
@bert-e reset

MonPote:
@bert-e ping

Issue #1439, opened 2019-07-17
Reduce code duplication iso.py


slaperche-scality:
**Component**:

buildchain

**Why this is needed**:

There are several places in `iso.py` where we need to copy files into the build tree.
This as led to the creation of three task that share a lot of code:

- `task__iso_add_iso_manager`
- `task__add_upgrade`
- `task__iso_add_node_manifest`

**What should be done**:

Factor our the duplicated code.

**Implementation proposal**:

For the examples, we can use a `targets.FileTree`.
For the scripts, I would go with a simple task generator that iterates over the list of scripts and generate a task for each of them. 

**Test plan**:

No visible impact on the buildchain: files are still copied as they used to be.

PR #1487, opened 2019-07-30
WIP: Improvement/956 saltify solution lifecycle

### LINKED ISSUE #956: Provide a runner for importing Solutions

**Why this is needed**: Ease of Solution deployment is the responsibility of MetalK8s

**What should be done**:

Define a Salt runner to _import_ all Solution archives listed in `/etc/metalk8s.solutions.yaml`, handling addition and removal.

1. For each ISO defined:
    a. Check format and retrieve Solution name and version
    b. Mount ISO at `/srv/scality/{{name}}-{{version}}` (replace/fail if already exists?)
    c. Configure NGINX registry (template of config file included in ISO)
    d. Add this version to the `metalk8s-solutions` ConfigMap

2. Given the contents of `metalk8s-solutions` ConfigMap, remove any version not seen in step 1:
    a. Remove NGINX configuration
    b. Unmount the ISO
    c. Remove the version from the ConfigMap

3. For each Solution name, pick the latest version imported:
    a. Deploy/replace the Operator CRDs
    b. Deploy/replace the UI Deployment
    c. Mark the version as being "active" in the ConfigMap (and remove that flag from any other version from that name)

**Test plan**: Manual, until #1061 
Tasks:

- [x] Setup config-maps/config-files to hold solutions state
- [x] Add salt module to handle read/write on configs
- [x] Add ext pillar for solutions
- [x] Add salt states that mounts, config and deploy solutions
- [x] Manage solution versioning
- [x] Add salt states that unmonut, unconfigure and undeploy solutions
- [x] Config repositories volumes/nginx config
- [x] Add an utility script to add/remove solutions
- [x] Add an orchestrate runner that manage solutions lifecycle
- [x] Adjust/Fix metalk8s_kubernetes for removing objects
- [x] Add deletion renderer
- [x] Update the buildchain

sayf-eddine-scality:
**Component**:
solutions, salt

**Context**: 
Solutions ISOs need a salt runners to check and import them into the cluster
**Summary**:
Salt runners all batteries included has been added with a bash script to automate these actions
**Acceptance criteria**: 

---

Closes: #956


salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    import kubernetes.client
+    from kubernetes.client.rest import ApiException
+    from pprint import pprint
+except ImportError:
+    HAS_LIBS = False
+
+log = logging.getLogger(__name__)
+
+
+__virtualname__ = 'solutions'
```
TeddyAndrieux: Not really like `solutions` if one day salt create a module named solutions that will override it. IMO you should prefix your module with `metalk8s`

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    file (required).
+    """
+    configmap_solutions = get_solutions_from_configmap()
+    config_file_solutions = get_solutions_form_configfile()
+    result = {k: config_file_solutions[k]
+              for k in config_file_solutions if k not in configmap_solutions}
+    print(result)
+    return result
+
+def set_configured(
```
TeddyAndrieux: Not really like the naming here. It's a module for me it should be `add_solution` or something like that

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+__virtualname__ = 'solutions'
+
+
+def __virtual__():
+    if HAS_LIBS:
+        return __virtualname__
+    else:
+        return False, 'python kubernetes library not found'
+
+def get_solutions_from_configmap(
```
TeddyAndrieux: `get_configured_solutions` to be consistent

salt/metalk8s/solutions/configured.sls
```diff
@@ -0,0 +1,18 @@
+{%- set solutions = pillar.metalk8s.solutions.unconfigured %}
```
TeddyAndrieux: For me this state should include the mounted one (or the init.sls should include both) and you don't need an orchestrate for mounted IMO

TeddyAndrieux (changes requested):
IMHO the approach here is not good as it's really strange to maintain a configmap when we configure solution.
For me the configmap should contain a list of solutions and salt should only read it (from an ext pillar) and configure/unconfigure the ISOs to match what is in the configmap. And if needed maybe the script can edit the configmap or something in the UI but for me the salt states should only apply what is in configmap and not maintain it. 

But it's only my opinion :)

salt/metalk8s/solutions/configured.sls
sayf-eddine-scality: yes i guess one init.sls will be enough

sayf-eddine-scality:
> IMHO the approach here is not good as it's really strange to maintain a configmap when we configure solution.
> For me the configmap should contain a list of solutions and salt should only read it (from an ext pillar) and configure/unconfigure the ISOs to match what is in the configmap. And if needed maybe the script can edit the configmap or something in the UI but for me the salt states should only apply what is in configmap and not maintain it.
> 
> But it's only my opinion
@gdemonet what do you think  ?

gdemonet:
> IMHO the approach here is not good as it's really strange to maintain a configmap when we configure solution.
> For me the configmap should contain a list of solutions and salt should only read it (from an ext pillar) and configure/unconfigure the ISOs to match what is in the configmap. And if needed maybe the script can edit the configmap or something in the UI but for me the salt states should only apply what is in configmap and not maintain it.
> 
> But it's only my opinion :)

Mmh we designed it to be some kind of _state persistence_ to be read/written by Salt when managing Solutions. The declarative listing comes from a file, `/etc/metalk8s/solutions.yaml`, and the `ConfigMap` serves as an observation of what is really deployed (even if it can be tampered with, granted). We need it for two reasons:

- comparing the desired Solution versions (from the file) with what is deployed (from the CM)
- list deployed Solution versions from the UI

buildchain/buildchain/salt_tree.py
```diff
@@ -324,6 +324,8 @@ def _run(self) -> None:
     Path('salt/metalk8s/orchestrate/deploy_node.sls'),
     Path('salt/metalk8s/orchestrate/downgrade/init.sls'),
     Path('salt/metalk8s/orchestrate/downgrade/precheck.sls'),
+    Path('salt/metalk8s/orchestrate/solutions/mounted.sls'),
+    Path('salt/metalk8s/orchestrate/solutions/configured.sls'),
```
gdemonet: As Teddy said, I'm not sure we need two different orchestrates for this task, since it can all happen in a single run. Maybe `metalk8s.orchestrate.import_solutions` is sufficient?

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+import logging
+import time
+import os
+import yaml
+
+import salt.utils.json
+from salt.exceptions import CommandExecutionError
+
+HAS_LIBS = True
+SOLUTIONS_CONFIG_MAP = 'solutions'
```
gdemonet: Let's name it `metalk8s-solutions` for clarity

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+SOLUTIONS_CONFIG_MAP_NAMESPACE = 'default'
```
gdemonet: Let's have a namespace for this, `metalk8s-solutions` seems like a good candidate for me, we could deploy Solution UIs there as well.

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+        return __virtualname__
+    else:
+        return False, 'python kubernetes library not found'
+
+def get_solutions_from_configmap(
+        context="kubernetes-admin@kubernetes",
+        kubeconfig="/etc/kubernetes/admin.conf"
+        ):
+    """Get list of imported solution from solutions config map.."""
+    cfg = __salt__['metalk8s_kubernetes.setup_conn'](
```
gdemonet: There is a `metalk8s_kubernetes.show_configmap` for your convenience, so you don't bother with error management here ;)

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+                sol_dict = solutions.setdefault(sol_id, {})
+                sol_dict[att] = v
+        return solutions
+    except ApiException as exc:
+        log.exception(
+            'Exception when calling '
+            'CoreV1Api->read_namespaced_config_map')
+        raise CommandExecutionError(exc)
+
+def get_solutions_form_configfile():
```
gdemonet: minor: 
```suggestion
def get_solutions_from_configfile():
```

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    """Get list of solutions defined in solution.yaml."""
+    solutions = {}
+    if os.path.exists(SOLUTIONS_CONFIG_FILE):
+        with open(SOLUTIONS_CONFIG_FILE, 'r') as fd:
+            solutions = yaml.safe_load(fd)
```
gdemonet: This would need some processing IMO, as we do for `bootstrap.yaml`:

- [x] First attempt to load as YAML (if it fails, complain about the format)
- [ ] Then, check the `kind` and `apiVersion` match what was expected (we may want to change the format in the future), and complain if it doesn't

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    else:
+        log.exception(
+            'Cannot find {}'.format(SOLUTIONS_CONFIG_FILE))
```
gdemonet: So, maybe the log is fine here, but we need the message to land in the `_errors` pillar key. Otherwise the Jinja exceptions will become a pain to manage.

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
+'''
+Various utilities to handle solution.
```
gdemonet: So, this module can be improved IMHO to provide a simpler interface, related to our needs.

For instance, it could expose:

- `metalk8s_solutions.list_configured` to list all Solutions defined in `/etc/metalk8s/solutions.yaml`, including parsing of the `product.txt` file without mounting the ISO
    - the result can have the format `{name: [{"version": version, "archive": archive_path}, ...], ...}`
- `metalk8s_solutions.list_deployed` to list all Solutions defined in the `metalk8s-solutions` `ConfigMap`
    - keys are Solution names
    - values are JSON, with following format: `{"version": version, "mountpoint": mountpoint, "deployed": true|false}`
- `metalk8s_solutions.add_solution` to add a Solution to the `ConfigMap` (`name`, `version`, and `mountpoint`)
- `metalk8s_solutions.remove_solution` to remove a Solution from the `ConfigMap`

Mounting and registry configuration (with their counterparts) can be left to the formulas/orchestrate, that will rely on those few methods to decide what to do. Difference between ConfigMap and file configuration should be done at that time as well.

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    if os.path.exists(SOLUTIONS_CONFIG_FILE):
+        with open(SOLUTIONS_CONFIG_FILE, 'r') as fd:
+            solutions = yaml.safe_load(fd)
+    else:
+        log.exception(
+            'Cannot find {}'.format(SOLUTIONS_CONFIG_FILE))
+        raise CommandExecutionError()
+    return solutions
+
+def get_unconfigured_solutions():
```
gdemonet: This can be removed from the module IMO, and left to the formula/orchestrate that needs such a list

salt/_pillar/metalk8s.py
```diff
@@ -120,6 +120,13 @@ def _load_iso_path(config_data):
 
     return res
 
+def _load_solutions():
+    """Load configured solutions from configmap."""
```
gdemonet: I'm not sure if we want the `ext_pillar` to load the `ConfigMap`, but if we do, I'd suggest something like:

```
def _load_solutions():
    """Load Solutions information."""
    return {
        'configured': __salt__['metalk8s_solutions.list_configured'](),
        'deployed': __salt__['metalk8s_solutions.list_deployed'](),
    }
```

Manipulating those two values (which are deployed but shouldn't, which are the "latest" versions, ...) can be done from formulas.

salt/metalk8s/orchestrate/solutions/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+
+Configure unconfigured solutions:
+  salt.state:
+  - tgt: {{ pillar.bootstrap_id }}
+  - sls:
+    - metalk8s.solutions.configured
+  - saltenv: metalk8s-{{ version }}
+
+
+{%- set solutions = pillar.metalk8s.solutions.unconfigured.items() %}
```
gdemonet: We can consider all _configured_ (from the `solutions.yaml` file) values here, since the states below should be idempotent

salt/metalk8s/orchestrate/solutions/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+{%- for _, solution in  solutions %}
+Update configmap solutions for {{ solution.name }}:
+  module.run:
+    - solutions.set_configured:
+      - solution: {{ solution }}
+      - kubeconfig: {{ kubeconfig }}
+      - context: {{ context }}
+{%- endfor %}
+
+Update registry for unconfigured solutions:
```
gdemonet: Did you add this state after updating the CM to ensure you'll have the freshest value from your ext_pillar? I think we want to do the opposite, and only use the `metalk8s:solutions:configured` info in the `metalk8s.repo.installed` formula.

salt/metalk8s/repo/files/repositories-manifest.yaml.j2
```diff
@@ -64,3 +68,9 @@ spec:
         path: {{ product.path ~ image_path }}
         type: Directory
     {%- endfor %}
+    {%- for name, solution in solutions.items() | sort(attribute='0') %}
+    - name: registry-{{ name | replace('.', '-') }}
+      hostPath:
+        path: /srv/scality/{{ name }}-{{ solution.version }}/images/
```
gdemonet: What happens if this is not available? Does the container fail to start?

salt/metalk8s/solutions/configured.sls
```diff
@@ -0,0 +1,18 @@
...
+  file.managed:
+    - source: {{ path }}/registry-config.inc.j2
+    - name: /var/lib/metalk8s/repositories/conf.d/{{solution.name}}-{{ solution.version }}-registry-config.inc
+    - template: jinja
+    - defaults:
+      repository: {{ solution.name }}
+      registry_prefix: {{ path }}/images
+{%- endfor %}
+{%- else %}
+No unconfigured solutions detected:
```
gdemonet: Probably not needed, `file.managed` is clearly idempotent. However, we need to think about Solutions to remove as well (the ones in `metalk8s:solutions:deployed` but not in `...:configured`).

salt/metalk8s/solutions/deployed.sls
```diff
@@ -0,0 +1,5 @@
+{%- for _, solution in salt.solutions.get_unconfigured_solutions().items() %}
+Deploy the operator for solution {{ solution.name }}:
```
gdemonet: That formula will be responsible for deploying the **latest version** of a Solution `name`, and in this case, deploying means creating/updating:

- the UI `Deployment`
- the CRDs needed by the Operator

Deploying the Operator itself is not part of MetalK8s per-se (even though we may want to provide a script to help users).

salt/metalk8s/solutions/mounted.sls
```diff
@@ -0,0 +1,41 @@
+
+{%- set solutions = pillar.metalk8s.solutions.unconfigured %}
+{%- if solutions %}
+{%- for _, solution in solutions.items() %}
+  {% set path = "/srv/scality/" ~ solution.name ~ "-" ~ solution.version %}
+  {%- if solution.iso %}
```
gdemonet: We really need to have one, that check can be made in the module function `metalk8s_solutions.list_configured`

salt/metalk8s/solutions/mounted.sls
```diff
@@ -0,0 +1,41 @@
...
+  - match_on:
+    - name
+  - require:
+    - file: Solution path {{ path }} exists
+  - require_in:
+    - file: Assert '{{ path }}/product.txt' exists
+
+  {%- endif %}
+
+Assert '{{ path }}/product.txt' exists:
```
gdemonet, line 30: Not sure why we need this, since we'll have read the file (without mounting) in `metalk8s_solution.list_configured` anyway

scripts/solution-manager.sh
```diff
@@ -0,0 +1,212 @@
...
+DRY_RUN="False"
+SALTENV=""
+PILLAR=""
+SOLUTION_PATH=""
+SOLUTION_NAME=""
+SOLUTION_VERSION=""
+
+
+_usage() {
+    echo "iso-manager.sh [options]"
```
gdemonet:
```suggestion
    echo "solution-manager.sh [options]"
```

scripts/solution-manager.sh
```diff
@@ -0,0 +1,212 @@
...
+    echo "Options:"
+    echo "-s/--solution <solution path>:   Path to solution ISO"
+    echo "-n/--name <solution name>:       Solution name"
+    echo "--version <solution version>:    Version fo the solution"
```
gdemonet: Neither `name` nor `solution` are needed, since we can read them from the ISO directly

scripts/solution-manager.sh
```diff
@@ -0,0 +1,212 @@
...
+}
+
+_add_solution() {
+    # Check solution args before continuing
+    if [[ -z "$SOLUTION_PATH" || -z "$SOLUTION_NAME" || -z "$SOLUTION_VERSION" ]]
+    then
+        echo "solution path name and version are required"
+        exit 1
+    fi
+    echo "Updating metalk8s.solutions.yaml"
```
gdemonet:
```suggestion
    echo "Updating $SOLUTION_CONFIG"
```

scripts/solution-manager.sh
```diff
@@ -0,0 +1,212 @@
...
+        merge_if_exists=True \
+        formatter=yaml \
+        show_changes=True \
+        --retcode-passthrough
+    return $?
+}
+
+_configure_solutions() {
+    # Mount solutions
+    echo "Mounting solutions..."
```
gdemonet: Maybe the vocabulary should be more vague, such as `Importing Solutions`, so that it reflects that we not only `mount`, nor only `configure` ;)

gdemonet (changes requested):
Some design issues: I think we don't want the ext_pillar to compute anything, only forward data sources (a config file, or the ConfigMap) to the minions.

Also, the format of `solutions.yaml` needs to be reviewed - it should look like:

```yaml
kind: SolutionsConfiguration
apiVersion: v1alpha1
archives:
  - /archives/hyperdrive-1.0.0.iso
  - /archives/hyperdrive-1.1.0.iso
  - /zenko/archives/zenko.iso
```

Our tooling should behave roughly like we do for MetalK8s ISOs: first read config file to list all ISOs we are concerned with, then read product info from the ISO without mounting it, then mount the ISO, finally configure the repositories to include the mounted ISO.

The ConfigMap doesn't need to exist for this functionality - however, it serves two other purposes:

1. it is read from the Platform UI to know about deployed Solutions
2. it is read by the `metalk8s.orchestrate.solutions` runner to know which Solution versions need to be removed

Otherwise, that's great progress, thanks!

salt/_modules/solutions.py
NicolasT: Definitely not `default` :)

salt/_modules/solutions.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    Return the difference between the configmap (configured) and the config 
+    file (required).
+    """
+    configmap_solutions = get_solutions_from_configmap()
+    config_file_solutions = get_solutions_form_configfile()
+    result = {k: config_file_solutions[k]
+              for k in config_file_solutions if k not in configmap_solutions}
+    print(result)
```
NicolasT: Oops

salt/metalk8s/orchestrate/solutions/configured.sls
```diff
@@ -0,0 +1,31 @@
...
+  - sls:
+    - metalk8s.solutions.configured
+  - saltenv: metalk8s-{{ version }}
+
+
+{%- set solutions = pillar.metalk8s.solutions.unconfigured.items() %}
+{%- for _, solution in  solutions %}
+Update configmap solutions for {{ solution.name }}:
+  module.run:
+    - solutions.set_configured:
```
NicolasT: Can the module be called `metalk8s_solutions` instead?

salt/metalk8s/solutions/mounted.sls
sayf-eddine-scality: to validate the iso internal structure, I added more asserts

salt/_modules/solutions.py
sayf-eddine-scality: oups

buildchain/buildchain/iso.py
```diff
@@ -130,16 +130,31 @@ def task__iso_add_utilities_scripts() -> types.TaskDict:
     upgrade_src = constants.ROOT/'scripts'/'upgrade.sh'
     upgrade_dest = constants.ISO_ROOT/'upgrade.sh'
     upgrade = [upgrade_src, upgrade_dest]
+
```
slaperche-scality: Could be simplified as:

```python
def task__iso_add_utilities_scripts() -> types.TaskDict:
    """Copy the ISO manager script to scripts."""
    files = (
        (
            constants.ROOT/'scripts'/'iso-manager.sh',
            constants.ISO_ROOT/'iso-manager.sh'
        ), (
            constants.ROOT/'scripts'/'downgrade.sh'
            constants.ISO_ROOT/'downgrade.sh'
        ), (
            constants.ROOT/'scripts'/'upgrade.sh'
            constants.ISO_ROOT/'upgrade.sh'
        ), (
            constants.ROOT/'scripts'/'solution-manager.sh'
            constants.ISO_ROOT/'solution-manager.sh'
        )
    )
    return {
        'title': utils.title_with_target1('COPY'),
        'actions': [(coreutils.cp_file, filepair) for filepair in files],
        'targets': [filepair[1] for filepair in files],
        'task_dep': ['_iso_mkdir_root'],
        'file_dep': [filepair[0] for filepair in files],
        'clean': True,
    }
```

Or maybe consider using a `targets.FileTree` targets (if it applies here, but I think it could).

salt/metalk8s/repo/files/repositories-manifest.yaml.j2
sayf-eddine-scality: this will only fail if the iso is not mounted for any reason

buildchain/buildchain/iso.py
sayf-eddine-scality: This code is changing a lot until we finish the PoC, I prefer to fix this here #1439

slaperche-scality: yeah, you're still in draft it's OK, just fix it before it get merged.

As for `#1439` it's something else: it to factorize between function.
Here I want to avoid duplication inside a single function.

salt/_pillar/metalk8s_solutions.py
```diff
@@ -0,0 +1,56 @@
+import os.path
```
gdemonet: Unused ;)

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,158 @@
...
+# Init
+Make sure solutions namespace exist:
+  metalk8s_kubernetes.namespace_present:
+    - name: metalk8s-solutions
+    - kubeconfig: {{ kubeconfig }}
+    - context: {{ context }}
+
+Make sure UIs namespace exist:
+  metalk8s_kubernetes.namespace_present:
+    - name: admin-uis
```
gdemonet: Now that I think of it, since they are all in the same Namespace, we can use `metalk8s-solutions` for the UIs and ConfigMaps IMO.

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,158 @@
...
+
+Configure unconfigured solutions:
+  salt.state:
+  - tgt: {{ pillar.bootstrap_id }}
+  - sls:
+    - metalk8s.solutions.configured
+  - saltenv: metalk8s-{{ version }}
+
+{%- set solutions_list = pillar.metalk8s.solutions.configured %}
+{%- if solutions_list %}
```
gdemonet: Do we need this test? I guess if the list is empty, we'll just never loop.

salt/metalk8s/repo/installed.sls
```diff
@@ -44,6 +44,13 @@ Install repositories manifest:
         name: {{ repositories_name }}
         version: {{ repositories_version }}
         products: {{ products }}
+{%- if 'solutions' in pillar.metalk8s %}
+  {%- set solutions = pillar.metalk8s.solutions.deployed %}
```
gdemonet: I think you want the `solutions.configured` here.

Also, could have done `{%- set solutions = pillar.metalk8s.solutions.configured | default({}) %}` I think.

salt/metalk8s/solutions/mounted.sls
gdemonet: This is probably something to check in some module function related to `get_product_info_from_iso`, as it could be done before mounting.

salt/metalk8s/repo/installed.sls
sayf-eddine-scality: this to prevent crash when calling locally where we override all the pillar and solutions are not in it ( refer to the bootstrap script L262)

salt/metalk8s/orchestrate/solutions/init.sls
sayf-eddine-scality: isnt it hardcoded in the deployment.yaml of the ui ?

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1582,6 +1582,72 @@ def customresourcedefinition_present(
...
+        'changes': {},
+        'result': False,
+        'comment': ''
+    }
+
+    customresourcedefinition = __salt__[
+        'metalk8s_kubernetes.show_customresourcedefinition'](name, **kwargs)
+
+    if customresourcedefinition is None:
+        ret['result'] = True if not __opts__['test'] else None
```
TeddyAndrieux: No, not salt good practice 

In test mode:
- Error when checking current status: `result=False`
- Nothing to do: `result=True`
- Some change needed: `result=None`

So here we should have `True` even if test mode enable

salt/_pillar/metalk8s.py
```diff
@@ -120,7 +120,6 @@ def _load_iso_path(config_data):
 
     return res
 
-
```
TeddyAndrieux: `salt/_pillar/metalk8s.py:123:1: E302 expected 2 blank lines, found 1`

salt/_pillar/metalk8s.py
```diff
@@ -132,7 +131,7 @@ def ext_pillar(minion_id, pillar, bootstrap_config):
         metal_data = {
             'products': _load_iso_path(config),
             'ca': _load_ca(config),
-            'api_server': _load_apiserver(config)
+            'api_server': _load_apiserver(config),
```
TeddyAndrieux: btw modification of this file does not seems legit :)

salt/_pillar/metalk8s_solutions.py
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if errors:
+        result.update(__utils__['pillar_utils.errors_to_dict'](errors))
+
+    return result
+
+
+def ext_pillar(minion_id, pillar):
+    solutions = _load_solutions
```
TeddyAndrieux: Useless variable

TeddyAndrieux: Also `solutions` is a function here

salt/metalk8s/solutions/prereq.sls
```diff
@@ -0,0 +1,10 @@
+Make sure solutions namespace exist:
```
TeddyAndrieux: This file seems no longer needed

scripts/solution-manager.sh
```diff
@@ -0,0 +1,229 @@
...
+SALTENV=""
+PILLAR=""
+
+
+_usage() {
+    echo "solution-manager.sh [options]"
+    echo "Options:"
+    echo "-s/--solution <solution path>:   Path to solution ISO"
+    echo "-n/--name <solution name>:       Solution name"
+    echo "--version <solution version>:    Version fo the solution"
```
TeddyAndrieux:
```suggestion
    echo "--version <solution version>:    Version of the solution"
```

scripts/solution-manager.sh
```diff
@@ -0,0 +1,229 @@
...
+      ;;
+    -v|--verbose)
+      VERBOSE=1
+      shift
+      ;;
+    -l|--log-file)
+      LOGFILE="$2"
+      [ -z "$DEBUG" ] && DEBUG="on"
+      shift 2
+      ;;
```
TeddyAndrieux, line 39: `--name` and `--version` seems not supported :)

scripts/solution-manager.sh
```diff
@@ -0,0 +1,229 @@
+#!/bin/bash
+set -e
+set -u
+set -o pipefail
+
+SOLUTION_CONFIG="/etc/metalk8s/solutions.yaml"
+VERBOSE=${VERBOSE:-0}
+LOGFILE="/var/log/metalk8s-solution-manager.log"
```
TeddyAndrieux, line 8: No related to this PR but IMO we should create a `/var/log/metalk8s/` directory as we have many log file now

scripts/solution-manager.sh
```diff
@@ -0,0 +1,229 @@
...
+    [ ${#SOLUTIONS[@]} -lt 1 ] && return 0
+    # The use salt file.serialize merge require having full list
+    # salt-call output example:
+    # local: ["/tmp/solution1.iso", "/tmp/solution2.iso"]
+    # parsed products:
+    # ("/tmp/solution1.iso" "/tmp/solution2.iso")
+    local existent_solutions=()
+    IFS=" " read -r -a \
+        existent_solutions <<< "$(salt-call pillar.get metalk8s:solutions:configured \
+        --out txt | cut -d' ' -f2- | tr -d '[],{}')"
```
TeddyAndrieux: Maybe simpler to do something in jinja here with `slsutil.renderer`

buildchain/buildchain/iso.py
```diff
@@ -119,29 +119,29 @@ def task__iso_add_node_manifest() -> types.TaskDict:
...
+            constants.ROOT/'scripts'/'iso-manager.sh',
+            constants.ISO_ROOT/'iso-manager.sh'
+        ), (
+            constants.ROOT/'scripts'/'downgrade.sh',
+            constants.ISO_ROOT/'downgrade.sh'
+        ), (
+            constants.ROOT/'scripts'/'upgrade.sh',
+            constants.ISO_ROOT/'upgrade.sh'
+        ), (
+            constants.ROOT/'scripts'/'solution-manager.sh',
```
gdemonet, line 133: Not sure if we really want such a script honestly, but fine for now I guess.

salt/_pillar/metalk8s.py
gdemonet: @TeddyAndrieux You know we don't have linting on Salt files? If that's something we want to add, let's open a ticket to run it in CI, instead of spending time for it in manual reviews.

salt/metalk8s/addons/monitoring/process-exporter/deployed.sls
```diff
@@ -0,0 +1,2 @@
+include:
+ - .upstream
```
gdemonet: Why is this part of this PR?

salt/metalk8s/orchestrate/solutions/available.sls
```diff
@@ -0,0 +1,48 @@
...
+{%- set apiserver = 'https://' ~ pillar.metalk8s.api_server.host ~ ':6443' %}
+{%- set version = pillar.metalk8s.nodes[pillar.bootstrap_id].version %}
+{%- set kubeconfig = "/etc/kubernetes/admin.conf" %}
+{%- set context = "kubernetes-admin@kubernetes" %}
+{%- set custom_renderer =
+  "jinja | kubernetes kubeconfig=" ~ kubeconfig ~ "&context=" ~ context
+%}
+
+# Init
+Make sure "metalk8s-solutions" Namespace exists:
```
gdemonet, line 13: We want to change this approach IIRC:

- only use a single Namespace to deploy all UIs (including Platform UI), let's call it `metalk8s-uis`
- define the `metalk8s-solutions` ConfigMap in this Namespace
- make sure the `ui-config` or `ui-branding` (name TBD) CM is the same for both Platform UI and the Solution ones

As a first iteration (this PR), let's not bother about Platform UI, and keep a dedicated namespace for Solutions: `metalk8s-solutions`. So, just remove the `admin-uis` NS and move the `ui-branding` CM to the `metalk8s-solutions` NS.

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,249 @@
...
+      - archive_path: {{ solution_info.iso }}
+      - deployed: {{ was_deployed }}
+      - kubeconfig: {{ kubeconfig }}
+      - context: {{ context }}
+    - require:
+      - salt: Prepare registry configuration for declared Solutions
+{%- endfor %} {# fullname, solution_info in solutions_info.items() #}
+
+
+# undeploy solutions
```
gdemonet, line 133: This loop is trickier than the current implementation. Let's consider a Solution named `s`, and versions `v1` and `v2`:

- case 0: nothing deployed => nothing to do
- case 1: `s-v1` deployed, `s-v1` and `s-v2` configured => the previous loop already deployed `s-v2` in place of `s-v1`, so nothing to do here
- case 2: `s-v1` deployed, only `s-v2` configured => same as above, the deployment for `s-v2` is done, so nothing to remove
- case 3: `s-v1` deployed, nothing configured => now we need to remove the deployment of `s-v1`

(considering `s-v2` being deployed beforehand is the same w.r.t this loop)

gdemonet: You could use the `highest_versions` dict (defined above) to check if a given Solution name is among the configured ones.

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,249 @@
...
+{%- set custom_absent_renderer =
+  "jinja | kubernetes kubeconfig=" ~ kubeconfig ~ "&context=" ~ context ~ "&absent=True"
+%}
+
+{%- set configured = pillar.metalk8s.solutions.configured or [] %}
```
gdemonet: No need to retrieve those values again, once per formula should be enough.

salt/metalk8s/solutions/mounted.sls
gdemonet: Let's add a comment + issue to track this effort (moving asserts to **before** mounting the ISO) in the future, should be enough for this iteration.

scripts/solution-manager.sh
gdemonet: This script was left aside when working on the demo, since it doesn't bring much value. Adding a line to `/etc/metalk8s/solutions.yaml` is quite straightforward, and this script can't do much more. Maybe we could remove it altogether, not sure though.

gdemonet (dismissed):
Few comments to address, I think it would be a smart choice to split the `solution-manager.sh` script out of this PR to discuss the improvements / utility later (as the feature is definitely not urgent IMO).

salt/_pillar/metalk8s.py
TeddyAndrieux: We should have one but it's not only about enable pylint/flake8/whatever because salt have some specific variables.
Just lets try to follow python good practice even if we have not automatic linting

scripts/solution-manager.sh
TeddyAndrieux: But this script should also configure and mount the solutions ISOs ? No ?

TeddyAndrieux: Do almost same thing as the `iso-manager.sh` script ?

gdemonet: Mmh, maybe. To me, a good workflow would be:

1- Edit the config file to match your needs
2- Run the _good_ command: `kubectl metalk8s import-solutions`

But maybe we prefer a flow where adding/removing a Solution version is the unit of work (and we run the `import-solutions` on each addition/removal).

salt/_modules/solutions.py
sayf-eddine-scality: will open a separate issue to think why we need this and which design of the file we need #1582 1582

scripts/solution-manager.sh
sayf-eddine-scality: #1583

scripts/solution-manager.sh
sayf-eddine-scality: This should be dont as a proper kubectl plugin, it is more of a helper script for the moment

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,256 @@
...
+  - saltenv: metalk8s-{{ version }}
+
+# Unmount
+Unmount removed Solutions archives:
+  salt.state:
+  - tgt: {{ pillar.bootstrap_id }}
+  - sls:
+    - metalk8s.solutions.unmounted
+  - saltenv: metalk8s-{{ version }}
+  require:
```
TeddyAndrieux:
```suggestion
  - require:
```

TeddyAndrieux (changes requested):

salt/metalk8s/orchestrate/solutions/init.sls
```diff
@@ -0,0 +1,254 @@
...
+  salt.state:
+  - tgt: {{ pillar.bootstrap_id }}
+  - sls:
+    - metalk8s.solutions.configured
+  - saltenv: metalk8s-{{ version }}
+
+# Compute information about Solutions from their ISO files
+{%- set solutions_info = {} %} {# indexed by <name>-<version> #}
+{%- set highest_versions = {} %} {# indexed by <name> #}
+{%- for solution_iso in pillar.metalk8s.solutions.configured %}
```
TeddyAndrieux: Why you not use the `configured` variable instead of retrieve pillar again

salt/metalk8s/solutions/mounted.sls
```diff
@@ -0,0 +1,55 @@
+
```
TeddyAndrieux: Remove this line please

TeddyAndrieux (dismissed):
LGTM

TeddyAndrieux (approved):

salt/_modules/metalk8s_solutions.py
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+
+def list_configured():
+    """Get list of Solution archives paths defined in a config file."""
+    try:
+        with open(SOLUTIONS_CONFIG_FILE, 'r') as fd:
+            content = yaml.safe_load(fd)
+    except Exception as exc:
+        log.exception('Failed to load "%s": %s',
```
TeddyAndrieux: Do not log.exception in module that run in ext_pillar please.

For me in module we should never `log.exception` we should only raise the exception and the caller handle the exception like he want

TeddyAndrieux (changes requested):

TeddyAndrieux (approved):

Issue #1459, opened 2019-07-22
Add unit tests and polish volume creation feature.


MonPote:
**Component**: ui, storage, tests

**Why this is needed**:
This issue is the follow-up of #1454

**What should be done**:
- Error management
  - ~~Display notification when there is an error during the creation of volume~~
- Validation management
  - Add form validation : 
    - ~~`size` -> `^[1-9]+[0-9]*$`must be a positive integer~~ 
    - ~~`path` -> Start with `/`~~
    - ~~`name` -> `'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'` and it should be unique~~
- Refresh Management
  - ~~The volume list should refresh itself~~
- Unit tests
  - ~~Add unit tests to redux sagas that has been created.~~
- The sort on the Storage Capacity column
Given the following sizes: 1Gi, 100Mi, 10Gi, 1Mi
```Ascending order: we expect 1Mi, 100Mi, 1Gi, 10Gi, but we get 100Mi, 10Gi, 1Gi, 1Mi```
```Descending order: we expect 10Gi, 1Gi, 100Mi, 1Mi, but we get 1Mi, 1Gi, 10Gi, 100Mi```
- ~~Catch error when fetching volumes to avoid this error:~~
![image](https://user-images.githubusercontent.com/47394132/61776014-620d9f80-adfa-11e9-88a2-44e07574e93f.png)

**Test plan**:
There will be unit tests and manual tests for the rest because we are still not sure how to do E2E tests for feature that require to create something (ie: Creation of Node and Volume).

PR #1488, opened 2019-07-30
Improvement/1459 add volume validation

ChengYanJin:
**Component**: ui, storage

**Context**: 
When we create the volume, we need to validate the form before create.

**Summary**:
Add form validation :
name -> ```/^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/ ```
size -> ``` /^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$/```
path -> Start with /

**Acceptance criteria**: 
UI will provide the error message to help the user give the correct information.
Screenshot 2019-07-30 at 17 38 40

**Note**
We will split the size field in two. One is input, the other is a select for the unit.

See: #1459

### DIFF
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -155,8 +155,6 @@ const CreateVolume = props => {
               dirty
             } = formikProps;
 
-            console.log('errors', errors);
-
             //touched is not "always" correctly set
             const handleOnBlur = e => setFieldTouched(e.target.name, true);
             return (
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -29,16 +29,15 @@ const FormSection = styled.div`
   display: flex;
   flex-direction: column;
 
-  .input_item {
-    width: 147px;
+  .sc-input-wrapper {
+    width: 200px;
   }
 `;
 
 const ActionContainer = styled.div`
   display: flex;
   margin: ${padding.large} 0;
-  margin-left:300px
-  justify-content: flex-start;
+  justify-content: flex-end;
   button {
     margin-right: ${padding.large};
   }
@@ -51,6 +50,7 @@ const CreateVolumeLayout = styled.div`
     .sc-input {
       display: inline-flex;
       margin: ${padding.smaller} 0;
+
       .sc-input-label {
         width: 200px;
       }
@@ -70,7 +70,7 @@ const SelectField = styled.div`
 `;
 
 const SelectFieldItem = styled.select`
-  width: 169px;
+  width: 200px;
 `;
 
 const CreateVolume = props => {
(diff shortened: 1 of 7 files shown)


ui/src/containers/CreateVolume.js
```diff
@@ -109,27 +140,45 @@ const CreateVolume = props => {
...
+              handleChange,
+              errors,
+              touched,
+              setFieldTouched,
+              dirty
+            } = formikProps;
+
+            console.log('errors', errors);
+
+            //touched is not "always" correctly set
```
slaperche-scality, line 158: Is this a known bug/documented behavior?

ui/src/translations/en.json
```diff
@@ -84,5 +84,8 @@
   "no": "No",
   "volume_creation": "Volume creation",
   "volume_creation_success": "Volume {name} has been created successfully.",
-  "volume_creation_failed": "Volume {name} creation has failed."
+  "volume_creation_failed": "Volume {name} creation has failed.",
+  "volume_name_error": "Name must be lowercase letters, numbers and hyphens.",
```
slaperche-scality, line 88: Not totally accurate (`you can use `.`, you can't use `-` at the start or end, …) but good enough I think.
An accurate message would be too long for an error message (probably OK for tooltip though)

ui/src/translations/en.json
```diff
@@ -84,5 +84,8 @@
...
+  "volume_path_error": "Device Path must start with a '/'",
```
slaperche-scality: `path`, not `Path`.

ui/src/translations/fr.json
```diff
@@ -84,5 +84,8 @@
   "no": "Non",
   "volume_creation": "Création du volume",
   "volume_creation_success": "Le volume {name} a été créé avec succèss",
-  "volume_creation_failed": "La création du volume {name} a échoué."
+  "volume_creation_failed": "La création du volume {name} a échoué.",
+  "volume_name_error": "Le nom doit être composé de caractères minuscules, nombres ou de tiret.",
```
slaperche-scality: - `caractères minuscules` => `minuscules` (otherwise we may think you talk about very small characters, it's a bit ambiguous :p)
- `nombres` => `chiffres` (there is a distinction between those two terms in French).

ui/src/translations/fr.json
```diff
@@ -84,5 +84,8 @@
...
+  "volume_path_error": "Le chemin du disque doit commencer par un '/'",
```
slaperche-scality, line 89: `périphérique` would probably be more accurate technically, that being said I think `disque` is more understandable so we can keep `disque`

slaperche-scality (dismissed):
>  We will split the size field in two. One is input, the other is a select for the unit.

:+1: 

Code review done, now I'm gonna test it

slaperche-scality:
Works as expected.

Note that the regexp used to validate the size is not perfect (you can still input invalid size) but that's the one used by Kubernetes itself that we reuse so I guess it's good enough, meh…

ui/src/translations/en.json
```diff
@@ -84,5 +84,8 @@
...
+  "volume_size_error": "Size must be composed of a number and a unit like 250Gi or 1Ti."
```
slaperche-scality: s/`number`/`positive integer`/ (and for the French s/`nombre`/`entier positif`/)

MonPote (dismissed):

slaperche-scality (approved):

ChengYanJin:
@bert-e approve

Issue #1489, opened 2019-07-31
Improve task callable type hints


gdemonet:
We disabled a `pylint` rule in #1477 to cope with type hints indicating a `Optional[TaskError]` return value. It seems we can avoid this altogether by adjusting the type hints, as explained below:

> Going back to that, I still think it's not needed.

> I think we may have messed up our type annotations.

> Seeing the examples on this issue https://github.com/python/mypy/issues/3157, one can see that the decorated function must be annotated with what it really returns (so if our function returns nothing, the return type must be `None`), because the decorator can have its own signature.

> So in our case:
> - `__call__` always return `None`¹
> - the decorator returns a function that return `Optional[TaskError]`

> With the following diff, both `pylint` and `mypy` are happy and we don't need to disable anything.

```diff
diff --git a/.pylintrc b/.pylintrc
index ce7a5299..c4aae53d 100644
--- a/.pylintrc
+++ b/.pylintrc
@@ -47,7 +47,6 @@ disable=abstract-method,     # Too many false positive :/
         bad-continuation,    # I kindly disagree :)
         locally-disabled,    # Sometime we need to disable a warning locally.
         suppressed-message,  # Don't pollute the output with useless info.
-        useless-return       # Disagreement with mypy, see https://github.com/python/mypy/issues/3974
 
 # }}}
 # Basic {{{
diff --git a/buildchain/buildchain/docker_command.py b/buildchain/buildchain/docker_command.py
index 56e53c71..4cf26444 100644
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -140,7 +140,7 @@ class DockerBuild:
 
     @task_error(docker.errors.BuildError, handler=build_error_handler)
     @task_error(docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
         DOCKER_CLIENT.images.build(
             tag=self.tag,
             path=self.path,
@@ -148,7 +148,6 @@ class DockerBuild:
             buildargs=self.buildargs,
             forcerm=True,
         )
-        return None
 
 
 class DockerRun:
@@ -266,14 +265,13 @@ class DockerRun:
     @task_error(docker.errors.ContainerError, handler=container_error_handler)
     @task_error(docker.errors.ImageNotFound)
     @task_error(docker.errors.APIError)
-    def __call__(self) -> Optional[TaskError]:
+    def __call__(self) -> None:
[...]

PR #1490, opened 2019-07-31
Fix task callables signatures and type hints

gdemonet:
In an attempt to clean the signature of the `docker_command` task
callables, we used the `Optional[TaskError]` return type hint.
However, this expressely required us to add a `return None` statement
for `mypy`, and disable a `pylint` rule to do so.

Instead, we use the return type hint `None` and don't `return` from such
callables, leaving the `Optional[TaskError]` to the `task_error`
decorator.

Fixes: #1489
See: #1477


slaperche-scality (approved):
Good job!

gdemonet:
@bert-e approve

PR #1452, opened 2019-07-19
Setup operator authentication

slaperche-scality:
**Component**:

operator, salt

**Context**: 

The operator will need to interact with salt-api, and thus it should be able to authenticate.

**Summary**:

In the PR we:
- update the `kubernetes_rbac` Salt authentication module to support `Bearer` authorization
- add a Salt API client in the storage operator such as we can authenticate against the Salt API server and run a dummy command (`test.ping`) for now

**Acceptance criteria**: 

You can test either with the operator running locally (will use HTTP Basic Authentication with default hardcoded creds) or inside the cluster (will use the ServiceAccount token, which is the expected behavior in production).

##### Setup

1. Connect to the bootstrap node: `vagrant ssh bootstrap`
2. Become root: `sudo su`
3. Register the new CRD: `kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f /vagrant/storage-operator/deploy/crds/storage_v1alpha1_volume_crd.yaml`
4. Get the "public" IP of the Salt API server (the one that use the port `4507`): `kubectl --kubeconfig /etc/kubernetes/admin.conf  -n kube-system describe svc salt-master | grep 'Endpoints'`
5. Copy the `/etc/kubernetes/admin.conf` from the bootstrap node to your local machine.

##### Testing

1. Start the operator locally:

```
cd storage-operator
export KUBECONFIG=<path-to-the-admin.cong-you-copied-locally>
export METALK8S_SALT_MASTER_ADDRESS=http://<IP-OF-SALT-API-WITHOUT-PORT>
operator-sdk up local 
```

2. Create a new volume: `kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f /vagrant/storage-operator/deploy/crds/storage_v1alpha1_volume_cr.yaml`

3. Check that in the Operator log you see a lines like:

```json
{"level":"info","ts":1563867500.5190809,"logger":"controller_volume","msg":"Reconciling Volume","Request.Name":"example-volume"}
{"level":"info","ts":1563867500.5191145,"logger":"salt_api","msg":"test.ping"}
{"level":"info","ts":1563867500.5191262,"logger":"salt_api","msg":"Auth","username":"admin","type":"Basic"}
{"level":"info","ts":1563867500.7507794,"logger":"salt_api","msg":"POST","url":"http://172.21.254.13:4507/login","StatusCode":200,"duration":231}
[...]

PR #1491, opened 2019-08-01
salt: Add salt pillar to get and list volumes per node

### LINKED ISSUE #1473: Salt external pillar to get volumes
**Component**: salt, storage

**Why this is needed**: To be able to list all volumes on a per node basis

**What should be done**: Create an external pillar to get the list of volumes, their configurations and status

**Test plan**: Be able to list the volumes using Salt (pillar.get)

Ebaneck:
**Component**:

'salt', 'kubernetes', 'storage'

**Context**: 

see: #1473

**Summary**:

- Be able to list all volumes on a node basis.
- Volume specific information is what we really want.

**Acceptance criteria**: 

1. You can test this locally by following : https://github.com/scality/metalk8s/pull/1452 (local deployment) to setup the storage operator and create test volumes.

2. Connect to the salt-master container and refresh the pillars  using `salt '*' saltutil.pillar_refresh`

3. On the bootstrap node, list the newly computed volume pillars using `salt-call pillar.get metalk8s:volumes`

4 The output should look like ;
```[root@bootstrap vagrant]# salt-call pillar.get metalk8s:volumes
    volumes:
            ----------
            example-volume:
                None
            example-volume3:
                ----------
                apiVersion:
                    storage.metalk8s.scality.com/v1alpha1
                kind:
                    Volume
                metadata:
                    ----------
                    annotations:
                        ----------
                        kubectl.kubernetes.io/last-applied-configuration:
                            {"apiVersion":"storage.metalk8s.scality.com/v1alpha1","kind":"Volume","metadata":{"annotations":{},"name":"example-volume3"},"spec":{"LoopDevice":{"size":"2Gi"},"nodeName":"bootstrap","storageClassName":"standard"}}
                    creationTimestamp:
                        2019-08-01T09:47:12Z
                    generation:
                        4
                    name:
                        example-volume3
                    resourceVersion:
                        104581
                    selfLink:
                        /apis/storage.metalk8s.scality.com/v1alpha1/volumes/example-volume3
                    uid:
                        a5fbfe40-c8e3-4623-ac59-c797739fa5d7
                spec:
                    ----------
                    LoopDevice:
                        ----------
                        size:
                            2Gi
                    nodeName:
                        bootstrap
                    storageClassName:
                        ----------
                        allow_volume_expansion:
                            None
                        allowed_topologies:
                            None
                        api_version:
                            None
                        kind:
                            None
                        metadata:
                            ----------
                            annotations:
                                ----------
                                kubectl.kubernetes.io/last-applied-configuration:
                                    {"apiVersion":"storage.k8s.io/v1","kind":"StorageClass","metadata":{"annotations":{},"name":"standard"},"mountOptions":["debug"],"parameters":{"type":"gp2"},"provisioner":"kubernetes.io/aws-ebs","reclaimPolicy":"Retain","volumeBindingMode":"Immediate"}
                            cluster_name:
                                None
                            creation_timestamp:
                                2019-07-31T10:02:11+00:00
                            deletion_grace_period_seconds:
                                None
                            deletion_timestamp:
                                None
                            finalizers:
                                None
                            generate_name:
                                None
                            generation:
                                None
                            initializers:
                                None
                            labels:
                                None
                            name:
                                standard
                            namespace:
                                None
                            owner_references:
                                None
                            resource_version:
                                1696
                            self_link:
                                /apis/storage.k8s.io/v1/storageclasses/standard
                            uid:
                                b1849754-e4bd-433a-b5de-8a1246febc4b
                        mount_options:
                            - debug
                        parameters:
                            ----------
                            type:
                                gp2
                        provisioner:
                            kubernetes.io/aws-ebs
                        reclaim_policy:
                            Retain
                        volume_binding_mode:
                            Immediate


```

---

Closes: #1473

### DIFF
--- a/salt/_pillar/metalk8s_nodes.py
+++ b/salt/_pillar/metalk8s_nodes.py
@@ -1,9 +1,12 @@
 import os.path
 import logging
+import datetime
 
 try:
     import kubernetes.client
+    from kubernetes.client.rest import ApiException
     import kubernetes.config
+    from urllib3.exceptions import HTTPError
     HAS_DEPS = True
 except ImportError:
     HAS_DEPS = False
@@ -50,6 +53,101 @@ def node_info(node, ca_minion):
     return result
 
 
+def get_cluster_version(api_client):
+    coreV1 = kubernetes.client.CoreV1Api(api_client=api_client)
+    try:
+        namespace = coreV1.read_namespace(name="kube-system")
+    except (ApiException, HTTPError) as exc:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to read namespace information {}'.format(exc)
+        )
+    annotations = namespace.metadata.annotations
+    annotation_key = 'metalk8s.scality.com/cluster-version'
+    if not annotations or annotation_key not in annotations:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to retrieve cluster version'
+        )
+
+    return namespace.metadata.annotations[annotation_key]
+
+
+def iso_timestamp_converter(timestamp):
+    if timestamp is None:
+        return None
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+    storageclass_list = storageV1.list_storage_class()
+    for storageclass in storageclass_list.items:
+        # Need to convert the datetime object in storageclass to ISO format in
+        # order to make them serializable.
+        storageclass.metadata.creation_timestamp = iso_timestamp_converter(
+            storageclass.metadata.creation_timestamp
+        )
+        storageclass.metadata.deletion_timestamp = iso_timestamp_converter(
+            storageclass.metadata.deletion_timestamp
+        )
+        storage_classes[storageclass.metadata.name] = storageclass.to_dict()
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+    customObjectsApi = kubernetes.client.CustomObjectsApi(
+        api_client=api_client
+    )
+    try:
+        storage_classes = get_storage_classes(api_client)
+    except (ApiException, HTTPError) as exc:
+        error_tplt = (
+            'Exception while calling StorageV1->list_storage_class {}'
+        )
+        return __utils__['pillar_utils.errors_to_dict'](
+            [error_tplt.format(exc)]
+        )
+
+    if not storage_classes:
+        return {}
+
+    try:
+        volumes = customObjectsApi.list_cluster_custom_object(
+            group="storage.metalk8s.scality.com",
+            version="v1alpha1",
+            plural="volumes"
+        )
+    except (ApiException, HTTPError) as exc:
+        error_tplt = (
+            'Exception while calling CustomObjectsAPi->list_custom_object {}'
+        )
+        return __utils__['pillar_utils.errors_to_dict'](
+            [error_tplt.format(exc)]
+        )
+
+    results = {}
+    local_volumes = (
+        volume for volume in volumes['items']
+        if volume['spec']['nodeName'] == minion_id
+    )
+    for volume in local_volumes:
+        name = volume['metadata']['name']
+        storageclass = storage_classes.get(
+            volume['spec']['storageClassName']
+        )
+        if storageclass:
+            volume['spec']['storageClassName'] = storageclass
+            name = volume['metadata']['name']
+            results[name] = volume
+        else:
+            log.error('Unknown StorageClass "{}" for Volume "{}"'.format(
+                volume['spec']['storageClassName'], name)
+            )
+
+    return results
+
+
 def ext_pillar(minion_id, pillar, kubeconfig):
     if not os.path.isfile(kubeconfig):
         error_tplt = '{}: kubeconfig not found at {}'
@@ -57,6 +155,7 @@ def ext_pillar(minion_id, pillar, kubeconfig):
             error_tplt.format(__virtualname__, kubeconfig)
         ])
         cluster_version = pillar_nodes
+        volume_information = pillar_nodes
 
     else:
         ca_minion = None
@@ -72,29 +171,22 @@ def ext_pillar(minion_id, pillar, kubeconfig):
 
         node_list = coreV1.list_node()
 
-        cluster_version = None
-        namespace_information = coreV1.read_namespace(name="kube-system")
-
-        if namespace_information:
-            ns_info_dict = namespace_information.to_dict()
-            annotation_info = ns_info_dict.get(
-                'metadata', {}).get('annotations')
-            if annotation_info:
-                cluster_version = annotation_info.get(
-                    'metalk8s.scality.com/cluster-version'
-                )
-        if not cluster_version:
-            cluster_version = __utils__['pillar_utils.errors_to_dict'](
-                'Unable to retrieve cluster version'
-            )
+        cluster_version = get_cluster_version(api_client=client)
         pillar_nodes = dict(
             (node.metadata.name, node_info(node, ca_minion))
             for node in node_list.items
         )
 
-    return {
+        volume_information = list_volumes(client, minion_id)
+
+    result = {
         'metalk8s': {
             'nodes': pillar_nodes,
             'cluster_version': cluster_version,
+            'volumes': volume_information,
         },
     }
+    for key in ['nodes', 'volumes']:
+        __utils__['pillar_utils.promote_errors'](result['metalk8s'], key)
+
+    return result


salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,40 @@ def ext_pillar(minion_id, pillar, kubeconfig):
             for node in node_list.items
         )
 
+        custom_object = []
```
TeddyAndrieux: `volume_information` make more sense IMO as you not store all "custom objects" in this dict

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,40 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+        try:
+            object_response = customObjectsApi.list_cluster_custom_object(
+                group="storage.metalk8s.scality.com", version="v1alpha1",
+                plural="volumes"
+            )
+            if not object_response:
+                custom_object = __utils__['pillar_utils.errors_to_dict']([
+                    'Unable to retrieve node volume list'
+                ])
+            for node in node_list.items:
```
slaperche-scality: Maybe I'm missing something, but how is the volume list dispatched between nodes?

Normally each node should only have its volumes in its pillar, is it done here?

TeddyAndrieux: Indeed no it's not done. You should not do a `for` loop and use the `minion_id` to only retrieve the node you need

Ebaneck: @slaperche-scality, I see your point. I can add something like

`if items['spec']['nodeName'] == node:`

We previously had this check. Maybe i took this off at the last minute.

Completely my bad (:

slaperche-scality: @Ebaneck actually I think you don't need to filter the list yourself, I think the `list_cluster_custom_object` allows you to pass a `label_selector` which you can use to filter the returned result (filter on `spec.nodeName=<minion_id>`)

gdemonet (changes requested):
If I may, I think the pillar keys naming could be improved: maybe go for `metalk8s:volumes` instead of `metalk8s:volume_information`, and within a single Volume details, don't prefix every key with `volume_`, and maybe rename `class` to `storage_class`.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,40 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                for k, v in object_response.items():
+                    if k == "items":
+                        for items in v:
+                            volume_type = list(items['spec'].keys())[0]
+                            result_vol_object = {
+                             'volume_name': items['metadata']['name'],
+                             'volume_uid': items['metadata']['uid'],
+                             'volume_type': volume_type,
+                             'volume_capacity': items['spec'][
```
NicolasT: This won't fly. You can't return the same (structure) of data for every type of volume.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,40 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                                 'sparseLoopDevice']['size'],
+                             'volume_class': items['spec']['storageClassName']}
```
NicolasT: This won't be sufficient. Every volume object in the pillar will need to contain the values we care about retrieved from the storage class (hint: add a cache when looking up StorageClasses). Just the name is not useful.

NicolasT (changes requested):

salt/_pillar/metalk8s_nodes.py
Ebaneck: Looks like the current kubernetes python client does not support filters based on `field_selector` for the custom object type. so we are marking this comment as resolved.

Resolution is to filter based on `items['spec']['nodeName'] == <minion_id>:`

salt/_pillar/metalk8s_nodes.py
Ebaneck: @NicolasT Thanks for picking this out. I am going to fix this.

salt/_pillar/metalk8s_nodes.py
Ebaneck: @NicolasT it would be helpful if you could provide a pre-computed list of  `storageclass` attributes we will like to add to each individual volume pillar. `mount_options`, `reclaim_policy`  does not seem sufficient to me.

NicolasT: I can't :) @slaperche-scality ? This is mainly about the file-system to be used, the `mkfs` options,... stored in annotations.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                group="storage.metalk8s.scality.com", version="v1alpha1",
+                plural="volumes"
+            )
+            if not object_response:
+                volume_information = __utils__['pillar_utils.errors_to_dict']([
+                    'Unable to retrieve node volume list'
+                ])
+            for items in object_response.get('items', []):
+                if items.get('spec', {}).get('nodeName') == minion_id:
+                    volume_type = list(items.get('spec', {}).keys())[0]
```
NicolasT: What's this supposed to do?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                    result_vol_object = {
+                        'name': items.get('metadata', {}).get(
+                            'name', {}),
+                        'uid': items.get('metadata', {}).get(
+                            'uid', {}),
+                        'type': volume_type,
+                        'capacity': items.get('spec').get(
+                            'sparseLoopDevice', {}).get('size', {}),
```
NicolasT: Why would the default be a dict?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                        'storage_class': items.get('spec', {}).get(
+                            'storageClassName', {})
```
NicolasT: Why would the default be a dict?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+            if not object_response:
+                volume_information = __utils__['pillar_utils.errors_to_dict']([
+                    'Unable to retrieve node volume list'
+                ])
+            for items in object_response.get('items', []):
+                if items.get('spec', {}).get('nodeName') == minion_id:
+                    volume_type = list(items.get('spec', {}).keys())[0]
+                    result_vol_object = {
+                        'name': items.get('metadata', {}).get(
+                            'name', {}),
```
NicolasT: `metadata.name` must always be set. If it's not, the input is invalid.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                        'uid': items.get('metadata', {}).get(
```
NicolasT: `metadata.uid` must always be set. If it's not , the input is invalid.

NicolasT (changes requested):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
             for node in node_list.items
         )
 
+        volume_information = []
```
NicolasT: This should really be split out into a separate function (or more). Even now, `ext_pillar` is way too big. The namespace annotation lookup added above shouldn't be in the function body either.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+        try:
+            object_response = customObjectsApi.list_cluster_custom_object(
+                group="storage.metalk8s.scality.com", version="v1alpha1",
+                plural="volumes"
+            )
+            if not object_response:
+                volume_information = __utils__['pillar_utils.errors_to_dict']([
+                    'Unable to retrieve node volume list'
+                ])
+            for items in object_response.get('items', []):
```
NicolasT: You're listing all volumes. Including the ones that are to-be-deleted but are not yet deleted. Either those need to be filtered out (which would be bad, because then implementing volume reclaim will be difficult), or the deleted-or-not status should be a field in the volume information, such that a state which is supposed to apply all volumes on a system can filter those out.

@slaperche-scality can you elaborate on this if needed?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+                if items.get('spec', {}).get('nodeName') == minion_id:
```
NicolasT: Here as well, no `nodeName` in the spec would be an invalid object, so just error out.

NicolasT (changes requested):

salt/_pillar/metalk8s_nodes.py
NicolasT: Given object names are unique, it could make sense to turn this into a dict keyed by the name instead of a list, which will then ease lookup in SLSs.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -92,9 +96,42 @@ def ext_pillar(minion_id, pillar, kubeconfig):
             for node in node_list.items
         )
 
+        volume_information = []
+
+        try:
```
NicolasT: If this were put in a separate function, there'd be no need for a massive `try`/`except` block (and the subsequent indent).

salt/_pillar/metalk8s_nodes.py
slaperche-scality: Yup, we may want to have a boolean field to discriminate "normal" volume from "terminating" one.

To know if a volume is terminating, you can check the field `Status.Phase` => if present AND the value is "Terminating" then you know that this volume is on its way out.

Ebaneck: `status` has been added to volume listing so i guess i can mark this as resolved

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,94 @@ def node_info(node, ca_minion):
     return result
 
 
+def cluster_version_info(namespace_information):
+    cluster_version = None
+
+    if namespace_information:
+        ns_info_dict = namespace_information.to_dict()
```
gdemonet: Why not use the object itself? I guess the whole method could benefit from a simpler syntax. Also, the `cluster-version` annotation is specific to the `kube-system` namespace, there's no real value in passing any namespace object.

```python
def get_cluster_version():
    namespace = coreV1.read_namespace(name="kube-system")
    annotation_key = 'metalk8s.scality.com/cluster-version'

    if annotation_key not in namespace.metadata.annotations:
        # return pillar `_errors`

    return namespace.metadata.annotations[annotation_key]
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,94 @@ def node_info(node, ca_minion):
...
+            )
+    if not cluster_version:
+        cluster_version = __utils__['pillar_utils.errors_to_dict'](
+            'Unable to retrieve cluster version'
+        )
+
+    return cluster_version
+
+
+def get_storageclass(storageClassV1):
```
gdemonet: The API to use is an implementation detail of this method IMO, shouldn't be the responsibility of the caller to know which one to pass. I'd suggest a simpler signature:

```python
def list_storage_classes(api_client):
    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
    storage_classes = {}
    ...
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,94 @@ def node_info(node, ca_minion):
...
+    storageclass_list = storageClassV1.list_storage_class()
+    if not storageclass_list:
+        return {}
+    for storageclass in storageclass_list.items:
+        storage_classes[storageclass.metadata.name] = storageclass
+
+    return storage_classes
+
+
+def volume_info(customObjectApi, minion_id, storageclassObject):
```
gdemonet: Ditto, let's go for something like:

```python
def list_volumes(api_client, node_name):
    customObjectApi = kubernetes.client.CustomObjectsApi(api_client=api_client)

    # NOTE: I'd be wary of API errors when listing storage classes as well
    storage_classes = list_storage_classes(api_client)

    try:
        volume_list = customObjectApi.list_cluster_custom_object(
            group="storage.metalk8s.scality.com",
            version="v1alpha1",
            plural="volumes",
        )
    except ApiException as exc:
        return __utils__['pillar_utils.errors_to_dict'](...)

    return dict(
        (volume.metadata.name, volume_info(volume, storage_classes))
        for volume in volume_list
    )
```

And add a `volume_info` helper to generate the dict you want from a `Volume` object and a dict of `StorageClass`es.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,94 @@ def node_info(node, ca_minion):
...
+                    'No none storage class found for {}'.format(
+                        items['metadata']['name']
+                    )
+                ])
+            else:
+                result_vol_object = {
+                    'name': items['metadata']['name'],
+                    'uid': items['metadata']['uid'],
+                    'capacity': items.get('spec').get(
+                        'sparseLoopDevice', {}).get('size', ""),
```
gdemonet: We need to do something for all those `.get(...)` syntaxes, I fear it's getting completely unreadable. Can't you use the usual `<obj_name>.<attr_name>` syntax instead of the `<dict_name>["<attr_name>"]` one?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,94 @@ def node_info(node, ca_minion):
...
+                        'storageClassName', ""),
+                    'status': items.get('status', {}).get('phase', ""),
+                    'storageclass_annotations':
+                        storageclass.metadata.annotations
+                }
+                volume_information[items['metadata']['name']] = \
+                    result_vol_object
+
+    except Exception as exc:
+        volume_information = __utils__['pillar_utils.errors_to_dict']([
```
gdemonet: When faced with such errors, you could use an early `return` statement instead of setting the final return value in a variable, makes it much more readable (you know the function stops there).

gdemonet (changes requested):
Few suggested improvements to the code, most of it is about having clear separation of concerns (SRP FTW) and using the simple `<obj>.<attr>` syntax for improved readability.

salt/_pillar/metalk8s_nodes.py
Ebaneck: I am not sure we can get rid of the `.get(..)` syntaxes

salt/_pillar/metalk8s_nodes.py
slaperche-scality: We won't use annotation but `parameters` to put our fields (`fsType` and `mkfsOptions`):

Here is an example:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: standard
provisioner: kubernetes.io/no-provisioner
parameters:
  fsType: ext4
  mkfsOptions: '["-m", "0", "-b", "1024", "-E", "stride=16,stripe-width=64"]'
reclaimPolicy: Retain
mountOptions:
  - debug
volumeBindingMode: Immediate
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+                volume_information = __utils__['pillar_utils.errors_to_dict']([
+                    'No known storage class found for {}'.format(
+                        items['metadata']['name']
+                    )
+                ])
+            else:
+                result_vol_object = {
+                    'name': items['metadata']['name'],
+                    'uid': items['metadata']['uid'],
+                    'capacity': items.get('spec').get(
```
slaperche-scality: If `spec` is missing, the `get` won't save you here: it will return `None` and the next `get` will raise an exception.

When I have to do nested lookup like that, I usually define a small helper like this one:

```python
def rec_get(dictionary, key, **kwargs):
    try:
        return functools.reduce(operator.getitem, key.split('.'), dictionary)
    except KeyError:
        if 'default' in kwargs:
            return kwargs['default']
        raise
```

And then you can write:
```python
result_vol_object = {
    'name': rec_get(items, 'metadata.name'),
    'uid': rec_get(items, 'metadata.uid'),
    'capacity': rec_get(items, 'spec.sparseLoopDevice.size', default=''),
    'storage_class': rec_get(items, 'spec.storageClassName', default=''),
    'status': rec_get(items, 'status.phase', default='Unknown'),
    'storageclass_annotations': storageclass.metadata.annotations
}
```
Which is a bit more readable.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+                        'sparseLoopDevice', {}).get('size', ""),
+                    'storage_class': items.get('spec', {}).get(
+                        'storageClassName', ""),
+                    'status': items.get('status', {}).get('phase', ""),
+                    'storageclass_annotations':
+                        storageclass.metadata.annotations
```
slaperche-scality: See my other comment: we won't use annotations but `parameters` ;)

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+    return namespace.metadata.annotations[annotation_key]
+
+
+def list_storage_classes(api_client):
+
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+
+    storageclass_list = storageV1.list_storage_class()
+    if not storageclass_list:
```
NicolasT: Does `list_storage_class` ever return `None` or similar?! I'd expect it to *always* return a `StorageClassList`, whose `items` may be empty, but that shouldn't matter in your code.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+            version="v1alpha1",
+            plural="volumes"
+        )
+
+        if not volume_list:
+            volume_information = __utils__['pillar_utils.errors_to_dict']([
+                'Unable to retrieve node volumes'
+            ])
+
+        storage_classes = list_storage_classes(api_client)
```
NicolasT: Again?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+
+    customObjectApi = kubernetes.client.CustomObjectsApi(api_client=api_client)
+    volume_information = {}
+
+    storage_classes = list_storage_classes(api_client)
```
NicolasT: What if this fails?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+            ])
+
+        storage_classes = list_storage_classes(api_client)
+
+        if not storage_classes:
+            return __utils__['pillar_utils.errors_to_dict'](
+                ['Unable to retrieve node volume list']
+            )
+
+        for items in volume_list.get('items'):
```
NicolasT: This variable should be logically called `item`.

NicolasT: Why use `volume_list.get`? A `VolumeList` should *always* have an `items` field, otherwise something is fishy. Even then, if it wouldn't exist, then because you use `dict.get(key)`, the result would be `None`, and `for items in None` would bail out.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +50,103 @@ def node_info(node, ca_minion):
...
+            if items['spec']['nodeName'] != minion_id:
```
NicolasT: I may be too much of a functional programmer, but I'd split this out:
- get all volumes
- filter the volumes we're interested in (i.e., the ones for this minion)
- do something with the resulting set

`for item in collection: if not predicate(item): continue` is a smell to me.

salt/_pillar/metalk8s_nodes.py
NicolasT: And really, don't pass it in like this. We're making things more complicated, less future-proof etc than it could be.

salt/_pillar/metalk8s_nodes.py
NicolasT: This approach simply can't work? Given this code, how is any Salt state using this pillar supposed to know which type of volume to create/prepare?

I kinda fail to understand why we are rewriting and peeling out values from the `Storage` object in the first place, instead of simply passing it into the pillar as-is, after replacing `storageClassName` with the actual `StorageClass` object. Then a pillar consumer only needs to know the schema of `Volume` and `StorageClass` to achieve whatever is required.

Ebaneck:
@bert-e wait

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +52,100 @@ def node_info(node, ca_minion):
...
+
+def list_storage_classes(api_client):
+
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+
+    storageclass_list = storageV1.list_storage_class()
+    if not storageclass_list:
+        return {}
+    for storageclass in storageclass_list.items:
```
TeddyAndrieux: That cannot work
```suggestion
    for storageclass in storageclass_list.items():
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +52,100 @@ def node_info(node, ca_minion):
...
+            )
+            if storageclass is None:
+                volume_information[item['metadata']['name']] = \
+                    'No known storage class found for {}'.format(
+                        item['metadata']['name']
+                    )
+
+            else:
+                item['spec']['storageClassName'] = storageclass
+                # need to serialize the object
```
gdemonet: Why?

salt/_pillar/metalk8s_nodes.py
Ebaneck: We are resolved to generally not peeling any values for the Storage object unless in extreme cases where we could not do without.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +52,92 @@ def node_info(node, ca_minion):
     return result
 
 
+def get_cluster_version(api_client):
+
```
gdemonet: Empty line is unnecessary here, would have preferred a docstring, but removing it is enough I guess.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +52,92 @@ def node_info(node, ca_minion):
...
+def iso_timestamp_converter(timestamp):
+    if not timestamp:
+        return None
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+    for storageclass in storageV1.list_storage_class().items:
```
gdemonet: You need to wrap the `storageV1.list_storage_class` call in a `try/except` block, otherwise things can go wrong and break things in other places (actually a failing `ext_pillar` will prevent you from doing many things with a minion).

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +52,92 @@ def node_info(node, ca_minion):
...
+                )
+            continue
+        volume['spec']['storageClass'] = storageclass
+
+        results[name] = volume
+    return results
+
+
 def ext_pillar(minion_id, pillar, kubeconfig):
     if not os.path.isfile(kubeconfig):
```
gdemonet, line 152: I'd rather have:

```python
if not os.path.isfile(kubeconfig):
    error_tplt = '...'
    error_dict = __utils__['pillar_utils.errors_to_dict'](...)
    return {'metalk8s': error_dict}
```

and continue from there, instead of assigning this error dict to some variables we re-use later.

gdemonet (changes requested):
We're close!

salt/_pillar/metalk8s_nodes.py
Ebaneck: Well spotted, I think the `get_cluster_version` method also needs a try/except block

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+        )
+    except (ApiException, HTTPError) as exc:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to get volumes {}'.format(exc)
+        )
+
+    for volume in volumes['items']:
+        name = volume['metadata']['name']
+
+        if volume['spec']['nodeName'] != minion_id:
```
NicolasT: As said before, I really think `for e in c: if not pred(e): continue` is a really bad pattern.

In any case, if you want to stick with it, move it to the very beginning of the loop, for readability sake.

Though I'd prefer something like

```python
local_volumes = (volume for volume in volumes['items'] if volume['spec']['nodeName'] == minion_id)

for volume in local_volumes:
    ...
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+            continue
+        storageclass = storage_classes.get(
+            volume['spec'].pop('storageClassName')
```
NicolasT: No reason to `pop`. `storageClassName` *is* part of the schema.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+        )
+        if not storageclass:
+            results[volume['metadata']['name']] = \
```
NicolasT: You already looked up `name` before.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+
+    for volume in volumes['items']:
+        name = volume['metadata']['name']
+
+        if volume['spec']['nodeName'] != minion_id:
+            continue
+        storageclass = storage_classes.get(
+            volume['spec'].pop('storageClassName')
+        )
+        if not storageclass:
```
NicolasT: I'd rewrite this again, to get rid of the `continue` below:

```python
if storageclass:
    volume['spec']['storageCass'] = storageClass
    result = volume
else:
    result = 'Invalid StorageClass'

results[name] = result
```

However, now there's a problem: we end up with values in the `results` dict (and as such in the pillar) that are (error) strings, and do not look like `Volume` objects. This makes writing SLSs using this pillar value *much* more cumbersome.

As such, instead of putting this error string in the pillar (especially repeating the list of StorageClasses in it, which seems quite superfluous?), I'd instead put `None` as value for the corresponding name (key in the dict), and log a proper error message.

NicolasT (dismissed):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,103 @@ def node_info(node, ca_minion):
...
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+    try:
+        storageclass_list = storageV1.list_storage_class()
+    except (ApiException, HTTPError) as exc:
+        return __utils__['pillar_utils.errors_to_dict'](
```
slaperche-scality: should be `return {}`, because at the call site you have this code

```python
    storage_classes = get_storage_classes(api_client)

    if not storage_classes:
        return __utils__['pillar_utils.errors_to_dict'](
            'Cannot list storage classes'
        )
```

TeddyAndrieux: I think it's a good idea to get the content of the exception

slaperche-scality: Fine by me, but then you need to update the calling code otherwise it's broken…

Or does `__utils__` stuff returns None already?

salt/_pillar/metalk8s_nodes.py
Ebaneck: I am not sure of the consequences this will have, but looking through the repo on how we generally handle errors with external pillar, I see similar patterns.

I am willing to discuss this more with salt experts herein.

This issue is being tracked by #1498

salt/_pillar/metalk8s_nodes.py
Ebaneck: Well, we can actually return None and log the exception accordingly.

I also think once we call __utils__ then the error will be logged to `_error`

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+    local_volumes = (
+        volume for volume in volumes['items']
+        if volume['spec']['nodeName'] == minion_id
+    )
+    for volume in local_volumes:
+        name = volume['metadata']['name']
+        storageclass = storage_classes.get(
+            volume['spec']['storageClassName']
+        )
+        if storageclass:
```
slaperche-scality, line 139:
```python
        if storageclass:
            volume['spec']['storageClassName'] = storageclass
            results[name] = result
        else:
            log.error('Invalid storageclass for {} volume'.format(name))
```

I would get rid of the `result` var that just make things confusing w.r.t. `results` and simply skip the invalid volume instead of adding `None` (so that the SLS doesn't have to filter anything).

NicolasT:
In that case good error logging is pertinent, otherwise you'd end up with a
pillar that for whatever reason doesn't include some volumes.

salt/_pillar/metalk8s_nodes.py
Ebaneck: @NicolasT as a baseline requirement, should we stick to listing all volumes in the pillar even if some volumes do not have storage classes?

Does it become problematic filtering out None in SLS files versus filtering out Objects for some reason? Because initially, we were returning strings for volumes with no storage class and I think you did mention that `None` should be used instead.

salt/_pillar/metalk8s_nodes.py
TeddyAndrieux: `__utils__|'pillarutils.errors_to_dict']` return a dict with `_errors` key with a list of messages

salt/_pillar/metalk8s_nodes.py
gdemonet: Change is made, so let's keep it. But for the record, filtering in a `for` loop is an idiomatic pattern in Python.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+    results = {}
+    customObjectApi = kubernetes.client.CustomObjectsApi(api_client=api_client)
+    storage_classes = get_storage_classes(api_client)
+
+    if not storage_classes:
```
gdemonet, line 112: No, let's not do it this way. There are two **very** different cases covered by this test:

- `storage_classes is None`, which actually means that there was an API failure when we tried listing: this case should make `list_volumes` return a `pillar_utils.errors_to_dict`
- `storage_classes == {}`, which means there is no StorageClass in the API: this case can still be covered as it is right now (`return {}`)

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+
+    try:
+        volumes = customObjectApi.list_cluster_custom_object(
+            group="storage.metalk8s.scality.com",
+            version="v1alpha1",
+            plural="volumes"
+        )
+    except (ApiException, HTTPError) as exc:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to get volumes {}'.format(exc)
```
gdemonet: Minor:

```suggestion
            'Unable to get volumes: {}'.format(exc)
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
+    try:
+        storageclass_list = storageV1.list_storage_class()
+    except (ApiException, HTTPError) as exc:
+        log.exception('Exception when calling storageV1->list_storage_class')
```
gdemonet: Minor:

```suggestion
        log.exception('Exception when calling StorageV1Api->list_storage_class')
```

salt/_pillar/metalk8s_nodes.py
NicolasT: Either not list them (but then log proper errors, as @slaperche-scality menioned), or `None`, but not strings. Checking for `None` in an SLS is much easier than checking or 'an object that looks like the kind of object I expect'.

However.

I guess @slaperche-scality is right and we should skip-but-log 'invalid' volumes and not represent them in the pillar.

The 'error' object should only be used when we can't properly render the pillar, i.e. there's a configuration issue or something along those lines.

gdemonet (changes requested):
There's one more case to cover w.r.t. to the behaviour in case of API failure when listing StorageClasses.

Also, and I think @NicolasT said otherwise, we don't want to log from external pillars: they are re-computed on every call to `state.sls`, hence polluting the output for no good reason.
The `_errors` key pattern, introduced by #1018, is a specific solution to this problem. Yes, consumers need to check `{% if '_errors' in volume %}` instead of `{% if not volume %}`. Is it really that cumbersome?
In addition, it allows a formula to display the appropriate error message in a `test.fail_without_changes` for example.
It also allows, if we use the `pillar_utils.promote_errors` utility method, to bubble up such `_errors` to the `metalk8s:volumes` level, allowing a consumer to fail early by checking `{% if '_errors' in pillar.metalk8s.volumes %}`.

NicolasT:
> There's one more case to cover w.r.t. to the behaviour in case of API failure when listing StorageClasses.
> 
> Also, and I think @NicolasT said otherwise, we don't want to log from external pillars: they are re-computed on every call to `state.sls`, hence polluting the output for no good reason.
> The `_errors` key pattern, introduced by #1018, is a specific solution to this problem. Yes, consumers need to check `{% if '_errors' in volume %}` instead of `{% if not volume %}`. Is it really that cumbersome?
> In addition, it allows a formula to display the appropriate error message in a `test.fail_without_changes` for example.
> It also allows, if we use the `pillar_utils.promote_errors` utility method, to bubble up such `_errors` to the `metalk8s:volumes` level, allowing a consumer to fail early by checking `{% if '_errors' in pillar.metalk8s.volumes %}`.

Works for me.

As long as we don't have either an object or a string in the pillar as the value of what's supposed to be an object in the happy path. Way too easy to end up with `AttributeError` kind of issues during SLS rendering, and type-checking in an SLS is meh.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+
+def iso_timestamp_converter(timestamp):
+    if not timestamp:
+        return None
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    storage_classes = {}
```
charlesprost: Create this variable after the try/except statement because we don't need it before and it troubles reading.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+            storageclass.metadata.deletion_timestamp
+        )
+        storage_classes[storageclass.metadata.name] = storageclass.to_dict()
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+    results = {}
+    customObjectApi = kubernetes.client.CustomObjectsApi(api_client=api_client)
```
charlesprost: If you use method names as inspiration to name your variables, then it's better to take the exact name so there is no confusion, especially if you switch from plural (`CustomObjectsApi`) to singular (`customObjectApi`) because that can be misleading.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+        storageclass.metadata.deletion_timestamp = iso_timestamp_converter(
+            storageclass.metadata.deletion_timestamp
+        )
+        storage_classes[storageclass.metadata.name] = storageclass.to_dict()
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+    results = {}
```
charlesprost: Do no create variables if you are not likely to use them in the next lines.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
     return result
 
 
+def get_cluster_version(api_client):
+    coreV1 = kubernetes.client.CoreV1Api(api_client=api_client)
```
charlesprost, line 57: Is it possible that this variable could be a `None` value for example if there is no connection? Or are we sure we'll always have a proper object?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+            version="v1alpha1",
+            plural="volumes"
+        )
+    except (ApiException, HTTPError) as exc:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to get volumes {}'.format(exc)
+        )
+
+    local_volumes = (
+        volume for volume in volumes['items']
```
charlesprost, line 131: Here we use the syntax `dict[key]`, but above we were using `storageclass_list.items`; if objects can be accessed through two different methods, let's have the same way of doing it.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+        if volume['spec']['nodeName'] == minion_id
+    )
+    for volume in local_volumes:
```
charlesprost, line 134: We could mix the two loops in one bigger. That could be easier to understand.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+        name = volume['metadata']['name']
```
charlesprost, line 135: Variable defined but not used just after.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,100 @@ def node_info(node, ca_minion):
...
+        storageclass = storage_classes.get(
+            volume['spec']['storageClassName']
+        )
+        if storageclass:
+            volume['spec']['storageClassName'] = storageclass
+            result = volume
+        else:
+            result = None
+            log.error('Invalid storageclass for {} volume'.format(name))
```
charlesprost: Could be interesting to have the content of the `storageclass` variable if it's evaluated as `False`. Is it a None or something else?

charlesprost:
```suggestion
            log.error("Invalid storageclass for volume '{}'".format(name))
```
I'm not sure that's how we do this in this project, but to me it's easier to read this way.

salt/_pillar/metalk8s_nodes.py
slaperche-scality: Yup, it's inconsistent but that's how the API works:

- StorageClass is a "standard" Kubernetes object, as such it is represent by a real Python class/object.
- Volume is a custom object (defined by us) and thus the API doesn't know about it and returns them as `dict`

Hence the difference in how we access them.

salt/_pillar/metalk8s_nodes.py
gdemonet: We'll always have an object, no worries to be had here.

salt/_pillar/metalk8s_nodes.py
gdemonet: It's a limitation of the client when using `CustomObjectsApi`: there's no actual Python class to cast the dict into in the client code, so you only receive a dict. And keeping the other syntax when possible eases readability, so let's keep it.

salt/_pillar/metalk8s_nodes.py
gdemonet: So, see previous comments from @NicolasT, but we are actually just using a generator expression here, which means it's a single loop in practice.
The rationale is readability: first define the items to loop over, then loop and perform actions on each iteration. Instead of: loop over everything, sometimes skip, sometimes perform actions.

salt/_pillar/metalk8s_nodes.py
gdemonet: Actually, the error message should be:

```suggestion
            log.error('Missing StorageClass "{}" for Volume "{}"'.format(storageClassName, name))
```

salt/_pillar/metalk8s_nodes.py
Ebaneck: @gdemonet I think here, it was previously agreed upon that we will not log errors to pillar but rather use `log.exception` and return None

Ebaneck: It is fine

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+    return timestamp.isoformat()
+
+
+def get_storage_classes(api_client):
+    storageV1 = kubernetes.client.StorageV1Api(api_client=api_client)
+    try:
+        storageclass_list = storageV1.list_storage_class()
+    except (ApiException, HTTPError) as exc:
+        error_tplt = 'Exception while calling StorageV1->list_storage_class {}'
+        return __utils__['pillar_utils.errors_to_dict'](
```
gdemonet: That's problematic, see below in `list_volumes`.

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,102 @@ def node_info(node, ca_minion):
...
+        storage_classes[storageclass.metadata.name] = storageclass.to_dict()
+
+    return storage_classes
+
+
+def list_volumes(api_client, minion_id):
+    customObjectsApi = kubernetes.client.CustomObjectsApi(
+        api_client=api_client
+    )
+    storage_classes = get_storage_classes(api_client)
```
gdemonet: With your current implementation, this may be a dict with `_errors` inside. So, not going to work well in case of failure.

Here's what needs to be done:

1. Remove the `try/except` logic from `get_storage_classes`
2. Add a `try/except` in this function:

```
def list_volumes(api_client, minion_id):
    # ...
    try:
        storage_classes = get_storage_classes(api_client)
    except (APIException, HTTPError) as exc:
        return __utils__['...'](error_tplt.format(exc))

    # keep on with the current body
```

gdemonet (changes requested):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,107 @@ def node_info(node, ca_minion):
...
+    if not annotations or annotation_key not in annotations:
+        return __utils__['pillar_utils.errors_to_dict'](
+            'Unable to retrieve cluster version'
+        )
+
+    return namespace.metadata.annotations[annotation_key]
+
+
+def iso_timestamp_converter(timestamp):
+    if not timestamp:
```
slaperche-scality: `if timestamp is None:` would be safer

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,107 @@ def node_info(node, ca_minion):
...
+        name = volume['metadata']['name']
+        storageclass = storage_classes.get(
+            volume['spec']['storageClassName']
+        )
+        if storageclass:
+            volume['spec']['storageClassName'] = storageclass
+            name = volume['metadata']['name']
+            results[name] = volume
+        else:
+            log.error('Missing StorageClass for Volume "{}"'.format(name))
```
slaperche-scality: `Missing` => `Unknown`

slaperche-scality (commented):
two minor remarks: otherwise it looks good to me and can be merged

slaperche-scality (dismissed):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -72,29 +169,21 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+        volume_information = list_volumes(client, minion_id)
+
+    result = {
         'metalk8s': {
             'nodes': pillar_nodes,
             'cluster_version': cluster_version,
+            'volumes': volume_information,
         },
     }
+    __utils__['pillar_utils.promote_errors'](result['metalk8s'], 'volumes')
```
gdemonet:
```suggestion
    for key in ['nodes', 'cluster_version', 'volumes']:
        __utils__['pillar_utils.promote_errors'](result['metalk8s'], key)
```

salt/_pillar/metalk8s_nodes.py
```diff
@@ -50,13 +53,107 @@ def node_info(node, ca_minion):
...
+        name = volume['metadata']['name']
+        storageclass = storage_classes.get(
+            volume['spec']['storageClassName']
+        )
+        if storageclass:
+            volume['spec']['storageClassName'] = storageclass
+            name = volume['metadata']['name']
+            results[name] = volume
+        else:
+            log.error('Unknown StorageClass for Volume "{}"'.format(name))
```
gdemonet: The unknown StorageClass name would have been useful as well

gdemonet (dismissed):
Approving, but will need a rebase once #1504 is merged (until then, no need to rebuild, it will always fail).

slaperche-scality (dismissed):

Ebaneck:
@bert-e approve

slaperche-scality (approved):

PR #1492, opened 2019-08-01
split the volume size field into two

ChengYanJin:
**Component**: ui, storage

**Context**:  We would like to split the volume size into two. One is input for the positive integer, the other is select for the unit.

**Summary**: For the input field, we check if it's the positive integer. And users can select the unit from the dropdown list.

**Acceptance criteria**: 
image

See: #1459

### DIFF
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -264,6 +264,12 @@ const CreateVolume = props => {
                           onBlur={handleOnBlur}
                         >
                           {sizeUnits
+                            /**
+                             * `sizeUnits` have a base 2 and base 10 units
+                             * (ie. KiB and KB).
+                             * We chose to only display base 2 units
+                             * to improve the UX.
+                             */
                             .filter((size, idx) => idx < 6)
                             .map(({ label, value }, idx) => {
                               return (
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -111,7 +111,7 @@ const CreateVolume = props => {
     storageClass: storageClassesName[0],
     type: types[0].value,
     path: '',
-    selectedUnit: sizeUnits[0].value,
+    selectedUnit: sizeUnits[3].value,
     sizeInput: ''
   };
   const validationSchema = yup.object().shape({
@@ -121,7 +121,12 @@ const CreateVolume = props => {
         /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/,
         intl.messages.volume_name_error
       )
-      .required(),
+      .required(
+        intl.formatMessage(
+          { id: 'generic_missing_field' },
+          { field: intl.messages.name.toLowerCase() }
+        )
+      ),
     storageClass: yup.string().required(),
     type: yup.string().required(),
     path: yup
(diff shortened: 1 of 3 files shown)


ui/src/services/utils.js
```diff
@@ -98,3 +98,11 @@ export const useRefreshEffect = (refreshAction, stopRefreshAction) => {
     };
   }, [dispatch, refreshAction, stopRefreshAction]);
 };
+
+export const sizeUnits = [
+  { label: 'ki', value: 'Ki' },
```
slaperche-scality: `ki` => `KiB`

I know that I told you that it was a small `k`, but actually it's more complicated:
- a small `k` for [kilobyte](https://en.wikipedia.org/wiki/Kilobyte) (`kB`) 
- a big `K` for [kibibyte](https://en.wikipedia.org/wiki/Kibibyte) (`KiB`)

ui/src/services/utils.js
```diff
@@ -98,3 +98,11 @@ export const useRefreshEffect = (refreshAction, stopRefreshAction) => {
     };
   }, [dispatch, refreshAction, stopRefreshAction]);
 };
+
+export const sizeUnits = [
```
slaperche-scality: I would add the `B` of byte for the labels, so: `KiB`, `MiB`, …

Also, I would add `{ label: 'B', value: '' }` as one may want to specify the size in bytes directly.

ui/src/translations/en.json
```diff
@@ -87,5 +87,5 @@
   "volume_creation_failed": "Volume {name} creation has failed.",
   "volume_name_error": "Name must be lowercase letters, numbers and hyphens.",
   "volume_path_error": "Device path must start with a '/'",
-  "volume_size_error": "Size must be composed of a positive integer and a unit like 250Gi or 1Ti."
+  "volume_size_error": "The size must be a postive integer."
```
slaperche-scality: `postive` => `positive`

ui/src/containers/CreateVolume.js
```diff
@@ -88,37 +101,44 @@ const CreateVolume = props => {
...
-      )
-      .when('type', {
-        is: SPARSE_LOOP_DEVICE,
-        then: Yup.string().required()
-      })
+    sizeInput: yup.number().when('type', {
+      is: SPARSE_LOOP_DEVICE,
+      then: yup
+        .number()
+        .integer(intl.messages.volume_size_error)
```
slaperche-scality: This is way more readable than a regex :+1: 

That being said, it's buggy :disappointed: 

If you input a "big" number, say 2 GiB as bytes (which is pretty small actually), e.g. `2147483648`, you have the error `The size must be a postive integer.` which is pretty weird from a user point of view (it is a positive integer).

I guess we wouldn't have the problem with the good ol’ regex :stuck_out_tongue:

slaperche-scality (changes requested):

ui/src/containers/CreateVolume.js
```diff
@@ -127,17 +127,16 @@ const CreateVolume = props => {
...
       }),
-    sizeInput: yup.number().when('type', {
+    sizeInput: yup.string().when('type', {
       is: SPARSE_LOOP_DEVICE,
       then: yup
-        .number()
-        .integer(intl.messages.volume_size_error)
-        .positive(intl.messages.volume_size_error)
+        .string()
+        .matches(/^[1-9][0-9]*$/, intl.messages.volume_size_error)
```
slaperche-scality: It would be nice to explain in the commit message why we went back to the regex

slaperche-scality (commented):
Works fine :smiley: 

Two remarks though:

1. Not in the scope of this PR but I realized that we're not checking if the volume name is unique.
2. I have the error message `sizeInput is a required field` in English, even in French (and `sizeInput` may not be very clear to the user). I guess it's an error message generated by `Yup`, I don't know if we can override it (if yes, we should do it).

carlito-scality:
Select Input style should be improved to be consistent with other inputs but can be managed later.

slaperche-scality (dismissed):

ChengYanJin:
@bert-e approve

MonPote (approved):

PR #1493, opened 2019-08-01
fix the volume size sorting

ChengYanJin:
**Component**: ui, storage

**Context**: The sorting is not sorted by capacity.

**Summary**: We sort the volume according to the size.

**Acceptance criteria**: 
image

See: #1459

### DIFF
--- a/ui/src/services/utils.js
+++ b/ui/src/services/utils.js
@@ -51,6 +51,81 @@ export const sortSelector = createSelector(
   list => list
 );
 
+/**
+ * This function will sort the `array` by capacity.
+ * The capacity should follow k8s regex rules.
+ * Each elements is an object that have a storageCapacity field.
+ *
+ * @param {array} list - The array that will be sort
+ * @param {string} sortBy - The field that will be sort
+ * @param {string} sortDirection - The direction of the sort.
+ *
+ * @example
+ * const capacities = [
+ *  { capacity: '1Ki' },
+ *  { capacity: '1Gi' },
+ *  { capacity: '100Mi' },
+ * ];
+ *
+ * const sortedCapacity = sortCapacity(capacities, 'capacity', 'DESC')
+ */
+export const sortCapacity = createSelector(
+  (list = [], sortBy = 'storageCapacity', sortDirection = 'ASC') => {
+    if (
+      Array.isArray(list) &&
+      typeof sortBy === 'string' &&
+      typeof sortDirection === 'string'
+    ) {
+      const k8sCapacityRegex = /^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$/;
+      const notSortableList = list.filter(
+        item => !k8sCapacityRegex.test(item?.[sortBy])
+      );
+
+      const sortedList = list
+        .filter(item => k8sCapacityRegex.test(item?.[sortBy]))
+        .map(item => {
+          const unit = item[sortBy].slice(-2);
+          const size = item[sortBy].substring(0, item[sortBy].length - 2);
+          const unitBase =
+            sizeUnits.find(sizeUnit => sizeUnit.value === unit)?.base ??
+            sizeUnits[0].value;
+
+          return {
+            ...item,
+            unit,
+            size,
+            unitBase
+          };
+        })
+        .sort((item1, item2) => {
+          const rawValue1 = item1.unitBase * item1.size;
+          const rawValue2 = item2.unitBase * item2.size;
+
+          if (sortDirection === 'ASC') {
+            return rawValue1 - rawValue2;
+          } else if (sortDirection === 'DESC') {
+            return rawValue2 - rawValue1;
+          } else {
+            return 0;
+          }
+        })
+        // Cleanup the object
+        .map(item => {
+          const cleanItem = { ...item };
+          delete cleanItem.unit;
+          delete cleanItem.size;
+          delete cleanItem.unitBase;
+          return cleanItem;
+        });
+
+      return [...sortedList, ...notSortableList];
+    } else {
+      return [];
+    }
+  },
+  list => list
+);
+
 export const getNodeNameFromUrl = (state, props) => {
   if (props && props.match && props.match.params && props.match.params.id) {
     return props.match.params.id;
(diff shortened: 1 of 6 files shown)


ui/src/services/utils.js
```diff
@@ -98,3 +173,11 @@ export const useRefreshEffect = (refreshAction, stopRefreshAction) => {
     };
   }, [dispatch, refreshAction, stopRefreshAction]);
 };
+
+export const sizeUnits = [
+  { label: 'ki', value: 'Ki', base: 10 ** 3 },
```
slaperche-scality: The `i` units use power of two, not power of ten.

- `Ki` => 2**10 bytes
- `Mi` => 2**20 bytes
- `Gi` => 2**30 bytes
- `Ti` => 2**40 bytes
- `Pi` => 2**50 bytes

ui/src/services/utils.js
```diff
@@ -51,6 +51,81 @@ export const sortSelector = createSelector(
...
+    ) {
+      const k8sCapacityRegex = /^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$/;
+      const notSortableList = list.filter(
+        item => !k8sCapacityRegex.test(item?.[sortBy])
+      );
+
+      const sortedList = list
+        .filter(item => k8sCapacityRegex.test(item?.[sortBy]))
+        .map(item => {
+          const unit = item[sortBy].slice(-2);
```
slaperche-scality: I think the `size` and `unit` could be extracted with the `k8sCapacityRegex` regex which would be simpler (with a little twist of the regex maybe).
I'll do some tests on my side and come back to you if I've get something good :slightly_smiling_face:

slaperche-scality (changes requested):
I think I've found an issue.

I created a volume with a size of `42949670000` (which is ~40Gi), and it seems like when creating a PV, Kubernetes converted it to `42949670k` (which is OK because it's equivalent, no loss of precision).

But I think the `k` is breaking the sorting (because we only added support for `*i` units), because this volume always stay at the top, no matter the sorting order :/

Also have some issues with `11111111111` (which is a bit more then 10Gi), not sorted properly.
I guess it's because it doesn't have a unit, but that can happen because here you can't have a unit without losing precision (since we only uses integers).

If you want to tests, here are the sizes of my volumes YAMLs:

```
    Size:              42949670000
    Size:              100Mi
    Size:              250Gi
    Size:              1Gi
    Size:              1Mi
    Size:              11111111111
    Size:              10Gi
```

ui/src/services/utils.test.js
```diff
@@ -8,6 +8,16 @@ const testcases = [
   { storageCapacity: '1Mi' }
 ];
 
+const testcases2 = [
+  { storageCapacity: '42949670000' },
```
slaperche-scality: Actually, the test case should be `42949670k`, because I do create the volume with `42949670000` but Kubernetes stores `42949670k`, and that's what you get in the table (so that's what we should be able to sort).

ui/src/services/utils.js
```diff
@@ -84,8 +84,19 @@ export const sortCapacity = createSelector(
...
+           *
+           * We might want to improve the regex to get rid of the useless empty
+           * string at the 1st and 4th place of the array later but right now
+           * just keep the k8s regex
+           *
+           * @example
+           * "1Gi" => ["", "1", "Gi", ""]
+           * "123" => ["", "123", "", ""]
+           */
+          const [empty, size, unit] = item[sortBy].split(k8sCapacityRegex);
```
slaperche-scality: As discussed on our common Slack channel, using `match` instead of `split` would make for a clearer code and remove the need for the big comment above.
Especially if you use named capture group.

For the record, I propose that we simplify the regex used, by using `^(?<size>[1-9][0-9]*)(?<unit>[kKMGTP]i?)?`
And then (to be checked if the code below is correct, I'm not a JS guru ^^):

```javascript
const capacityRegex = /^(?<size>[1-9][0-9]*)(?<unit>[kKMGTP]i?)?/
[…]
const match= item[sortBy].match(capacityRegex).groups;
const size = match.size
const unit = match.unit
```
Or something like that.

ui/src/services/utils.test.js
```diff
@@ -17,6 +27,17 @@ it('should sort correctly the array', () => {
...
   ]);
+
+  const result2 = sortCapacity(testcases2, 'storageCapacity');
+  expect(result2).toEqual([
+    { storageCapacity: '1Mi' },
+    { storageCapacity: '100Mi' },
+    { storageCapacity: '1Gi' },
+    { storageCapacity: '10Gi' },
+    { storageCapacity: '11111111111' },
+    { storageCapacity: '42949670000' },
```
slaperche-scality: ditto

slaperche-scality (changes requested):
Good progress on this one!

Almost everything is good, except `42949670k`, but I think it can be simply fixed by adding required entries in `sizeUnits` :slightly_smiling_face:

ui/src/services/utils.js
```diff
@@ -82,35 +86,34 @@ export const sortCapacity = createSelector(
...
-           * "1Gi" => ["", "1", "Gi", ""]
-           * "123" => ["", "123", "", ""]
+           * "1Gi" => { 'groups': { size: '1', unit: 'Gi'} }
+           * "123" => { 'groups': { size: '1', unit: undefined } }
            */
-          const [empty, size, unit] = item[sortBy].split(k8sCapacityRegex);
-          const unitBase =
-            sizeUnits.find(sizeUnit => sizeUnit.value === unit)?.base ??
-            sizeUnits[0].value;
+          const sizeRegex = /^(?<size>[1-9][0-9]*)(?<unit>[ikKMGTP]{1,2})?$/;
```
slaperche-scality: minor: could be improved again a little bit by using `^(?<size>[1-9][0-9]*)(?<unit>[kKMGTP]i?)?`

The current one accept `20ii` or `20GG` which is incorrect, the new one is more strict (while not becoming more complex).

ui/src/services/utils.js
```diff
@@ -82,35 +86,34 @@ export const sortCapacity = createSelector(
...
+          const { groups } = item[sortBy].match(sizeRegex);
+
+          const tmpInternalUnit = groups?.unit ?? '';
+          const tmpInternalSize = groups?.size ?? '0';
```
slaperche-scality: You'll always have a size, because the first capture group is not optional: if you don't have a `size`, the regex won't match.
But you're correct for the `unit`, it's optional (the second matching group is using `?`).

ui/src/services/utils.js
```diff
@@ -82,35 +86,34 @@ export const sortCapacity = createSelector(
       );
 
       const sortedList = list
+        // Filter wrong value (ie: null or incorrect unit)
         .filter(item => k8sCapacityRegex.test(item?.[sortBy]))
```
slaperche-scality: We should also use the new regex here, instead of the old `k8sCapacityRegex`

MonPote (approved):

ChengYanJin:
@bert-e approve

slaperche-scality (approved):
Good to go o/

PR #1497, opened 2019-08-05
Improvement/1457 notice the user when no storage class

### LINKED ISSUE #1457: We should notice the user when the ptf doesn't have any storage classes.
**Component**: ui

**Why this is needed**:
When a user does not have any Storage Class he is stuck at the creation of volume without any hint to solve the issues.

**What should be done**:
We should told him, he does not have any storage class and he should create one.
However this message should be display only when we are sure he does not have any.
We don't want this message to be display when the app is fetching storage class for example.

**Test plan**:

ChengYanJin:
**Component**: ui, storage

**Context**: As a user, I want to see the error message and a hint to help me solve the problem instead of a blank page when there is no storage class.

**Summary**: We display a banner to show the error message and give a hint to users.

**Acceptance criteria**: 
image

Closes: #1457

### DIFF
--- a/ui/src/components/Banner.js
+++ b/ui/src/components/Banner.js
@@ -3,17 +3,25 @@ import styled from 'styled-components';
 import {
   fontSize,
   padding,
-  yellowOrange
+  jade,
+  yellowOrange,
+  warmRed
 } from '@scality/core-ui/dist/style/theme';
+import { STATUS_BANNER_WARNING, STATUS_BANNER_ERROR } from '../constants';
 
 const BannerContainer = styled.div`
   display: flex;
   margin: ${padding.base} 0 0 ${padding.larger};
   padding: ${padding.small} ${padding.small} ${padding.small} 0;
   font-size: ${fontSize.small};
-  border: 1px solid ${yellowOrange};
-  border-left: 5px solid ${yellowOrange};
+
+  border: 1px solid;
+  border-left: 5px solid;
   border-radius: 3px;
+  border-color: ${props => {
+    return props.themeBanner;
+  }};
+
   a {
     text-decoration: none;
     margin-left: ${padding.smaller};
@@ -22,7 +30,9 @@ const BannerContainer = styled.div`
     display: flex;
     align-items: center;
     margin-left: ${padding.small};
-    color: ${yellowOrange};
+    color: ${props => {
+      return props.themeBanner;
+    }};
   }
 `;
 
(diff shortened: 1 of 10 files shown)


ui/src/containers/App.js
```diff
@@ -34,7 +34,6 @@ class App extends Component {
 
   render() {
     const { language, api, theme } = this.props.config;
-
```
slaperche-scality, line 37: minor: this seems totally unrelated to the current PR (the file is not touched except for this change).

It's nice to do cosmetics changes like that, but they deserve to be in a dedicated commit.

Don't bother splitting it this time, but please keep it in mind for a future PR.

ui/src/components/Banner.js
```diff
@@ -0,0 +1,61 @@
...
+`;
+
+const Banner = props => {
+  const { intlMsg } = props;
+
+  return (
+    <BannerContainer>
+      <i className="fas fa-exclamation-triangle" />
+      <TextContainer>
+        <TextLine>{intlMsg.no_storage_class_found}</TextLine>
```
slaperche-scality: This is a kind of title, do we want it in bold or something?

ui/src/containers/CreateVolume.js
```diff
@@ -100,6 +104,7 @@ const CreateVolume = props => {
   const storageClassesName = props.storageClass.map(
     storageClass => storageClass.metadata.name
   );
+  const isSCLoading = useSelector(state => state.app.volumes.isSCLoading);
```
slaperche-scality: The meaning of `SC` may not be obvious to the reader that wasn't involved in the development (myself I was wondering about the meaning for a while).

I suggest doing like l.168 and use `StorageClass` instead of `SC`.

ui/src/translations/fr.json
```diff
@@ -82,11 +82,15 @@
...
   "volume_creation": "Création du volume",
   "volume_creation_success": "Le volume {name} a été créé avec succèss.",
   "volume_creation_failed": "La création du volume {name} a échoué.",
   "volume_name_error": "Le nom doit être composé de minuscules, chiffres ou de tiret.",
   "volume_path_error": "Le chemin du disque doit commencer par un '/'.",
   "volume_size_error": "La capacité du volume doit être un entier positif.",
-  "generic_missing_field": "Veuillez entrer un {field}."
+  "generic_missing_field": "Veuillez entrer un {field}.",
+  "no_storage_class_found": "Aucune classe de stockage trouvée",
+  "storage_class_is_required": "Une classe de stockage est nécessaire pour créer un volume. Veuillez en créer une à partir de Kubernetes.",
```
slaperche-scality: `necessaire` => `requise` (to match the English version)

Also `à partir` is mainly used with a temporal meaning, so I would suggest `depuis`.
Or, to be consistent with the English sentence: `dans`.

i.e.: `Une classe de stockage est requise pour créer un volume. Veuillez en créer une dans Kubernetes.`

slaperche-scality (commented):
Works as expected, good job :slightly_smiling_face: 

Only have a few comments about the "implementation", but nothing too serious.

ui/src/containers/App.js
ChengYanJin: > minor: this seems totally unrelated to the current PR (the file is not touched except for this change).
> 
> It's nice to do cosmetics changes like that, but they deserve to be in a dedicated commit.
> 
> Don't bother splitting it this time, but please keep it in mind for a future PR.

I'll keep in mind.  Thanks for the kind advice!

slaperche-scality (dismissed):
Good to go :smiley:

ChengYanJin:
@bert-e approve

MonPote (approved):
LGTM.
A few comment of possible improvement:
- `Banner.js` can be greatly improve IMO. Like adding the possibility of passing anything in the children.
- I'm not a fan of a or i in `BannerContainer` because it change all the icons and link in the Banner and it's maybe not what we really want to.

That said, it does the job. So let merge it now. We will have plenty of time to improve it when we have more use cases.

PR #1501, opened 2019-08-06
Update `static-container-registry`

NicolasT:
Update the subtree'd `static-container-registry` to the latest version, including @gdemonet fixes from https://github.com/NicolasT/static-container-registry/pull/12

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -28,6 +28,7 @@ Overview:
 """
 
 
+import abc
 import importlib
 from pathlib import Path
 import sys
@@ -61,9 +62,9 @@ def task__deploy_salt_tree() -> Iterator[types.TaskDict]:
         yield from file_tree.execution_plan
 
 
-class CommonStaticContainerRegistry(targets.AtomicTarget):
-    """Generate a nginx configuration to serve common static container
-    registry."""
+class _StaticContainerRegistryBase(targets.AtomicTarget, abc.ABC):
+    """Common code to interact with `static-container-registry`."""
+
     def __init__(
         self,
         destination: Path,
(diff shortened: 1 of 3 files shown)


gdemonet (dismissed):

slaperche-scality:
Maybe we can patch the `_run` method of `CommonStaticContainerRegistry` to take advantage of the new signature of `create_config` instead of reaching for `CONSTANTS` directly as we currently do.

Would be more "future-proof".

NicolasT:
Fair enough, I'll apply that change.

NicolasT:
@bert-e reset

NicolasT:
@slaperche-scality Done. Also factored out some duplication between the two task types related to `static-container-registry` interactions.

buildchain/buildchain/salt_tree.py
```diff
@@ -85,8 +87,7 @@ def task(self) -> types.TaskDict:
         task = self.basic_task
         task.update({
             'title': self._show,
-            'doc': 'Generate the nginx config to serve a common static '
-                   'container registry.',
+            'doc': self._TASK_DOC,
```
slaperche-scality: Since the `_TASK_DOC` is really similar to the docstring of the class, we may want to use `self.__doc__` instead of a dedicated variable.

buildchain/buildchain/salt_tree.py
```diff
@@ -98,12 +99,36 @@ def _show(task: types.Task) -> str:
...
+        parts = self._get_parts()
         with Path(self.targets[0]).open('w', encoding='utf-8') as fp:
-            fp.write(container_registry.CONSTANTS)
+            for part in parts:
+                fp.write(part)
+
+    def _get_parts(self) -> Iterator[str]:
+        """Yield all parts that should go in the generated configuration file.
+        """
+        raise NotImplementedError
```
slaperche-scality, line 111: minor: could inherit from `abc.ABC` and use the `@abc.abstractmethod` decorator.

NicolasT:
@bert-e reset

NicolasT:
@bert-e force_reset

NicolasT:
@bert-e approve

slaperche-scality (approved):

PR #1502, opened 2019-08-06
Add metalk8s-sosreport to the debian packages

### LINKED ISSUE #1471: Create metalk8s-sos-report debian package
**Component**: package

**Why this is needed**: We need metalk8s-sos-report for ubuntu

**What should be done**: create a package sos-report and its plugins for debian

**Test plan**: verify that sos-report and plugins are well installed

ycointe:
**Component**: packages

**Context**: Metalk8s over Ubuntu

**Summary**: Need to add sources to build metalk8s-sosreport

**Acceptance criteria**: 
Be able to build metalk8s-sosreport:

- Move into metalk8s-sosreport package forlder: ` $ cd packages/debian/metalk8s-sosreport-deb-2.4.0`
- Build the package: ` $ debuild -b -uc -us`
 -> You should have `metalk8s-sosreport-deb_2.4.0_amd64.deb` in the parent folder.

You can verify if the package is working well:
- Download and install sosreport: ` $ sudo apt install sosreport`
- Install metalk8s-sosreport: ` $ dpkg -i metalk8s-sosreport-deb_2.4.0_amd64.deb `
-> then verify the installation: ` $ sosreport -l | grep -E "metalk8s|containerd" `

---

Closes: #1471


packages/debian/metalk8s-sosreport-deb-2.4.0/debian/changelog
```diff
@@ -0,0 +1,12 @@
+metalk8s-sosreport (2.4.0) bionic; urgency=low
+
+  * Initial release for test.
+
+ -- Yohann Cointe <yohann.cointe@scality.com>  Wed, 24 Jul 2019 17:11:02 +0200
+
+metalk8s-sosreport (2.4.0) UNRELEASED; urgency=low
```
sayf-eddine-scality: useless input, only keep latest changelog

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/control
```diff
@@ -0,0 +1,13 @@
+Source: metalk8s-sosreport
+Section: base
+Priority: extra
+Maintainer: Yohann Cointe <yohann.cointe@scality.com>
```
sayf-eddine-scality: maintainer should be Moonshot team, for future support
(changelog) too

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/control
```diff
@@ -0,0 +1,13 @@
...
+Build-Depends: debhelper (>=9)
+Standards-Version: 2.4.0 
+Homepage: https://github.com/scality/metalk8s
+
+Package: metalk8s-sosreport-deb
```
sayf-eddine-scality: no need to `-deb` as the file literally ends with `.deb`

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/metalk8s-sosreport-deb/DEBIAN/control
```diff
@@ -0,0 +1,12 @@
+Package: metalk8s-sosreport-deb
```
sayf-eddine-scality: dublicate file ? one of them should be unnecessary

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/metalk8s-sosreport-deb/usr/share/sosreport/sos/plugins/containerd.py
```diff
@@ -0,0 +1,73 @@
+#! /bin/env python
```
sayf-eddine-scality: I guess you did a `git add -A` some files are generated on build time and do not need to be in the cvs

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/metalk8s-sosreport-deb/usr/share/sosreport/sos/plugins/metalk8s.py
```diff
@@ -0,0 +1,119 @@
+#! /bin/env python
```
sayf-eddine-scality: same

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
```diff
@@ -0,0 +1,73 @@
+#! /bin/env python
```
sayf-eddine-scality: so this file and the next are common to debian and redhat, they should reside in one place to ease maintenance and should be copied (linked) in build time.

sayf-eddine-scality (dismissed):

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
ycointe: Is there no chance that these files could evolve with the time ?
If one file change for an OS family and not for another, it could be a problem if different OS families use the same.

sayf-eddine-scality: good point but I am not sure we want to support two codebases or have one file that supports both OSs, tbd

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/copyright
```diff
@@ -0,0 +1,9 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: metalk8s-sosreport
+Source: <url://example.com>
+#
+# Please double check copyright with the licensecheck(1) command.
```
slaperche-scality: I guess we will want to use the Apache License (which is what we use for the project).

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
slaperche-scality: Please, don't use `python`: use either `python2` or `python3`
Only use `python` when your code support both version.

I've had to patch too many scripts that were broken on ArchLinux because they now use Python 3 as default but people still think `python` == `python2` which is wrong (see [PEP-394](https://www.python.org/dev/peps/pep-0394/))

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
```diff
@@ -0,0 +1,73 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class containerd(Plugin, RedHatPlugin, UbuntuPlugin):
```
slaperche-scality: If we do keep two versions of the code, then I think the Ubuntu version doesn't need to inherit from `RedHatPlugin` and the CentOS version doesn't need to inherit from `UbuntuPlugin`.

But seeing the quantity of duplicated code, I think we will want to keep a single implementation that support both OSes.

Worst case (if they do diverge a lot) we could have something like:

```python
class UbuntuContainerd(Plugin, UbuntuPlugin):
    …

class RedHatContainerd(Plugin, RedHatPlugin):
    …

if os_is_ubuntu:
     containerd = UbuntuContainerd
elif os_is_redhat:
     containerd = RedHatContainerd
else:
    raise some_error
```

but that won't be needed for now.

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/metalk8s-sosreport-deb.substvars
```diff
@@ -0,0 +1,2 @@
+misc:Depends=
+misc:Pre-Depends=
```
slaperche-scality: Don't wee need to depends on some version of python and `sos`?

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
ycointe: As python3 will be the next standard and will be installed by default on Debian systems, I suggest that we only use python3.

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/.debhelper/generated/metalk8s-sosreport-deb/installed-by-dh_install
```diff
@@ -0,0 +1,2 @@
+./debian/src/metalk8s.py
```
jbertran: I don't think these `.debhelper` generated files should be committed, should they?

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/README.Debian
```diff
@@ -0,0 +1,8 @@
+metalk8s-sosreport for Debian
+
+Please edit this to provide information specific to
```
jbertran: Let's have a significant description

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/README.Debian
```diff
@@ -0,0 +1,8 @@
+metalk8s-sosreport for Debian
+
+Please edit this to provide information specific to 
+this metalk8s-sosreport Debian package.
+
+    (Automatically generated by debmake Version 4.2.9)
+
+ -- Yohann <ycointe@localhost>  Wed, 24 Jul 2019 15:49:02 +0200
```
jbertran: Let's generate these files with the correct email address for you.

See [here](https://people.debian.org/~osamu/maint-guide.html#email-setup) for that.

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/control
```diff
@@ -0,0 +1,13 @@
+Source: metalk8s-sosreport
+Section: base
+Priority: extra
+Maintainer: Moonshot Team <moonshot-platform@scality.com>
```
jbertran: Platform team

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/copyright
```diff
@@ -0,0 +1,9 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: metalk8s-sosreport
+Source: <url://example.com>
```
jbertran: Stray example URL

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/metalk8s-sosreport-deb.substvars
jbertran: This is `substvars`, we don't actually need this file to have contents (or at all???) if we fill up what we need in `control`.

@ycointe please check whether or not we need this file to be present at all

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/watch
```diff
@@ -0,0 +1,2 @@
+# You must remove unused comment lines for the released package.
```
jbertran: Let's follow the instructions :)

jbertran (commented):
I'm by no means an expert on debian packaging, ignore at will if you have better information.

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/src/containerd.py
jbertran: In the meantime, this is a possibility for inheriting from the correct plugin:

```
In [1]: class Both: 
   ...:     pass 
   ...:                                                             

In [2]: class One: 
   ...:     def a(self): 
   ...:         print('toto') 
   ...:                                                             

In [3]: class Other: 
   ...:     def a(self): 
   ...:         print('tata') 
   ...:                                                             

In [4]: selector = True                                             

In [5]: class Inheritor(Both, One if selector else Other): 
   ...:     pass 
   ...:                                                             

In [6]: i = Inheritor()                                             

In [7]: i.a()                                                       
toto
```

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/README.Debian
ycointe: Thx !

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/README.Debian
```diff
@@ -0,0 +1,5 @@
+metalk8s-sosreport for Debian
+
+Metalk8s sos-report custom plugins
```
jbertran: Let's be consistent between OS for the naming.

It's `Metalk8s SOS report custom plugins` in the `.spec`.

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/rules
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/make -f
+# You must remove unused comment lines for the released package.
```
jbertran: Let's follow the instructions :)

packages/debian/metalk8s-sosreport-deb-2.4.0/debian/source/local-options
```diff
@@ -0,0 +1,2 @@
+#abort-on-upstream-changes
```
jbertran: I'm guessing we can remove this file if we keep these options commented. From [here](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#sourcel) it looks like these are for a specific workflow which is not our own - even if we will need that file later, we can add it back when we do the buildchain integration.

jbertran (commented):
We can still trim it down a little :)

Let's also think about not having the custom plugin code twice in git.

packages/common/containerd.py
```diff
@@ -0,0 +1,73 @@
+#! /bin/env python
```
slaperche-scality: Please specify either `python2` or `python3` ditto for the other file.

Also, about the location, I would put the files under `packages/common/metalk8s-sosreport` (or even `packages/common/metalk8s-sosreport/src`).

In case we have more packages in the future we shouldn't put stuff directly under `common`.

slaperche-scality (dismissed):

packages/common/metalk8s-sosreport/containerd.py
```diff
@@ -1,4 +1,4 @@
-#! /bin/env python
```
jbertran, line 1: Can we have a separate commit for these changes ?

packages/debian/README.md
```diff
@@ -0,0 +1,31 @@
+<h1>
+    <img src="artwork/generated/metalk8s-logo-wide-black-400.png" width="400" height="100%" alt="MetalK8s logo" title="Debian packages" />
+</h1>
+
+## Requirement
+
+To build mannually Debian packages, install `devscripts`, `debhelper`
+and `buid-essential`:
```
jbertran: Typo, and let's have the commands with a helpful package name since we don't all run Ubuntu.

`to build manually, the commands dh and debuild are required on top of the MetalK8s build requirements.` or something like that

packages/debian/README.md
```diff
@@ -0,0 +1,31 @@
+<h1>
+    <img src="artwork/generated/metalk8s-logo-wide-black-400.png" width="400" height="100%" alt="MetalK8s logo" title="Debian packages" />
+</h1>
+
+## Requirement
+
+To build mannually Debian packages, install `devscripts`, `debhelper`
```
jbertran: `s/manually/manually`

jbertran (commented):
Lookin' good! Minor stuff to amend :)

packages/debian/README.md
```diff
@@ -0,0 +1,31 @@
+<h1>
```
slaperche-scality, line 1: Is it needed?

ycointe: It's not mandatory but it can be useful for now.

gdemonet: Let's not start littering the repo with `.md` files again. We have real documentation, let's use it please :)

packages/debian/metalk8s-sosreport/debian/control
```diff
@@ -0,0 +1,13 @@
+Source: metalk8s-sosreport
+Section: base
+Priority: extra
+Maintainer: Plateform Team <moonshot-platform@scality.com>
```
gdemonet:
```suggestion
Maintainer: Scality Platform Team <moonshot-platform@scality.com>
```

packages/debian/metalk8s-sosreport/debian/copyright
```diff
@@ -0,0 +1,12 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: metalk8s-sosreport
+Source: <https://github.com/scality/metalk8s>
+#----------------------------------------------------------------------------
+#
+# 		    GNU GENERAL PUBLIC LICENSE
```
gdemonet: No I don't think we want to use `GPL` licenses. Let's stick with the same licensing policy as for the parent repo, Apache v2 (cf. [here](https://github.com/scality/metalk8s/blob/development/2.4/LICENSE)).

packages/debian/metalk8s-sosreport/debian/src/containerd.py
```diff
@@ -0,0 +1,73 @@
+#! /bin/env python3
```
gdemonet: Don't think we still need these, since they also are under a `common` directory, right?

gdemonet (dismissed):

packages/debian/README.md
jbertran: This is dev only and only for a very specific purpose though (building packages by hand).

gdemonet: > This is dev only 

`docs/developer`

> and only for a very specific purpose though (building packages by hand).

We have other examples of temporary docs in there as well (accessing services without an Ingress for instance).

Also, building packages by hand may be a valid use-case that we'll want to keep in the future for dev purposes.

packages/debian/metalk8s-sosreport/debian/copyright
slaperche-scality: Depends if SOS is GPL and requires its plugin to be GPL as well.

To be checked, but yeah if possible let's stick to a single license.

packages/debian/metalk8s-sosreport/debian/src/containerd.py
slaperche-scality: Why should we remove the shebang?

packages/debian/README.md
ycointe: The file will be kept until it is automated.
When it will be automated, the .md content file will be integrated in the documentation and will be deleted.
Does everyone is agree ?

packages/debian/metalk8s-sosreport/debian/copyright
ycointe: I checked, SOS is under GPL v2 licence.
Do we keep this licence or do we keep the Apache licence ?

packages/debian/metalk8s-sosreport/debian/src/containerd.py
ycointe: These files (containerd.py and metalk8s.py) should have been deleted after tests for the common folder.
I will take care of it.

packages/debian/README.md
slaperche-scality: Agree. What we want to document is how to build package using a container +  `entrypoint.sh`

This is only temporary note until we're here.

packages/debian/metalk8s-sosreport/debian/src/containerd.py
gdemonet: (not related to the shebang sorry)

packages/debian/metalk8s-sosreport/debian/copyright
NicolasT: This can be Apache2'd just like other parts of MetalK8s.

ycointe: I used the Apache 2 licence.
Thank you for your help about it.

ycointe:
@bert-e approve

jbertran (approved):

jbertran:
Everything square after deployment, all comments taken into account.

Nice work! Now for the real fun - calico, and the buildchain next :smile:

PR #1470, opened 2019-07-24
Implement storage operator reconciliation loop

slaperche-scality:
**Component**:

operator

**Context**:

PV should be "hidden" from the user, and thus all their handling (creation and deletion) should be automated by the reconciliation loop and Volume CR.

**Summary**:

Implement all the necessary logic to:
- create a PV when a volume is created
- delete a PV when a volume is deleted
- handle PV and Volume lifetimes (using finalizer and ownership)
- prevent the deletion of Volume/PV is the PV is bound

**Acceptance criteria**:

Start the operator locally following the instructions given in https://github.com/scality/metalk8s/pull/1452

#### Create a volume

Create new volume (using `kubectl apply` or the UI):
-  The volume should be picked up by the Operator and will go into `Pending` state.
-  After some random amount of time, a PV with the proper size/path will be created
- The volume will become `Available`.

You can monitor the progression by watching the UI or with `watch  kubectl --kubeconfig=/etc/kubernetes/admin.conf describe volume <VOLUME-NAME` (that way you'll also see the events, those are not yet shown in the UI)

##### Bonus test

Stop the `salt-api` container while the volume creation is in progress:
- the error should be handled gracefully (no crash)
- we have periodic retry
- the creation will resume automatically once `salt-api` is brought back up by Kubernetes.

#### Delete a volume

Use `kubectl delete` to delete the volume:
- the volume will go into `Terminating` state
- After a while it'll be deleted (and the backing PV with it).

##### Bonus test 1

Stop the `salt-api` container while the volume deletion is in progress:
- the error should be handled gracefully (no crash)
- we have periodic retry
- the deletion will resume automatically once `salt-api` is brought back up by Kubernetes.

##### Bonus test 2

Create a PVC to bound the PV, then delete the volume: the volume isn't be deleted.
Now, delete the PVC: the PV is no longer bound, the deletion is resumed automatically

#### Delete the PV

Use `kubctl delete` the PV directly: the PV becomes `Terminating` but isn't deleted until the volume exists

**How to review**:

To review, you'll need:
- to squat (remember tovarish: "heel on the ground: a comrade found, but heel in the sky: it's a Western spy")
- a shot of vodka
- a stack of блин

More seriously, it will be easier to do the review commit by commit (several of them are small, and the big ones have useful commit messages).

**Afterwords**

That was a big and long PR. It wouldn't have been possible without:
- the `artist-mode` of Emacs for the schemas
- блин and плов for brain's calories intake
[...]

PR #1511, opened 2019-08-09
Update bootstrap to deploy the Volume CRD and the operator with lucky parrot

slaperche-scality:
See  #1470


slaperche-scality:
me dummy, bad branch name

PR #1512, opened 2019-08-09
Implement storage operator reconciliation loop lucky parrot

### LINKED ISSUE #1462: Implement the reconciliation loop
**Component**:

operator

**Why this is needed**:

PV should be "hidden" from the user, and thus all their handling should be automated by the reconciliation loop and Volume CR.

**What should be done**:

We should handle creation (when a new Volume CR is created), deletion (lifecycle management) and failure (error while preparing the disk).

**Test plan**:

Test manually that you can create a volume (automated tests will be tacked in another issue) and that the PV has the expected properties (`path`, `size`, …).
Also test the deletion.

slaperche-scality:
Same as #1470 but with 
Hopefully the bot won't get stuck this time…

---

Closes: #1462

### DIFF
--- a/storage-operator/README.md
+++ b/storage-operator/README.md
@@ -15,7 +15,7 @@
 
 ```shell
 export KUBECONFIG=<path-to-the-admin.cong-you-copied-locally>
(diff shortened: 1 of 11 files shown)


slaperche-scality:
not needed

Issue #1475, opened 2019-07-25
Unprepare volume with Salt


alexandre-allard:
**Component**: salt, storage

**Dependencies**:
- [x] #1473 
- [x] #1474

**Why this is needed**: To be able to remove device previously created with Salt

**What should be done**: Create a salt formula which will do the reverse of #1474

**Implementation proposal** (strongly recommended):

**Test plan**: We need to be able to completely "unprepare" a volume.

NicolasT:
For reference: while #1474 is about a *state* that can be invoked for something to exist (idempotently), this is about an *action* that can be invoked for something to not-exist (kinda idempotently, but only to be invoked in known, well-defined cases).

NicolasT:
(That action can be a state, of course, my bad)

slaperche-scality:
What does it means to "unprepare" in this case?
- `losetup --detach` and `rm the/sparse/file` for `sparseLoopDevice`?
- `wipefs` (through the Salt module `disk.wipe`) on the device for `rawBlockDevice`?

NicolasT:
Something like that, yes.

Although we may want to check whether we actually want to 'unprepare' behind the users' back... Any opinion @xaf-scality ?

NicolasT:
Checked with @xaf-scality in-person.

For `rawBlockDevice`, `unprepare` should basically do nothing. Let's keep the FS around (such that if someone deletes a `Volume` by mistake, data can still be recovered).
For `sparseLoopDevice`, because this accumulates on `/var/lib/metalk8s/...`, let's remove the backing sparse file upon deletion. This is for test/dev/demo purposes only anyway.

NicolasT:
Over time, we could add a `reclaimPolicy` similar to how it exists on a `PersistentVolume` which could then be `Retain` (don't delete anything), `Delete` (`wipefs`/remove sparse file/...), or whatnot. Given some sensible default (`Retain`, most likely), we could allow a user to choose what must happen.

PR #1513, opened 2019-08-09
Volume preparation using Salt formulas

### LINKED ISSUE #1474: Prepare volume with Salt
**Component**: salt, storage

**Dependencies**:
- [ ] #1473 

**Why this is needed**: To be able to write persistent data on disks (for Prometheus, Solutions, ...)

**What should be done**: Add Salt formulas to handle the volume creation, for the moment we only want to handle raw & loop devices.
* Check StorageClass validity (ensure that the requested storage class exists and the requested parameters associated to it are set and valid)
* Loop devices (Create the loop device and the sparse file associated to it)
* Create FS on devices (mkfs)
* Mechanism to create automatically loop devices at system start-up

**Implementation proposal**: 
The state should be able to take 1 or 0 device as parameter, if nothing is provided, we must prepare all the volumes not prepared yet.
For loop devices, we can use a systemd unit to ensure their creation at the system start-up, but another solution would be to use some kind of event provided by Salt when minion starts (this solution would be cleaner, we need to be check if it's doable)

**Test plan**: We should be able to prepare a device (a loop and a raw device).

Ebaneck:
**Component**:

'salt', 'storage', 'kubernetes'

**Context**: 

see:  #1474

**Summary**:

A combination of Salt state and custom salt modules should be able to take 0 or 1 device as arguments and prepare the specified volume if no argument is passed, we should prepare all volumes not yet prepared within the pillars.

**Acceptance criteria**: 

- Be able to successfully prepare loop devices and block devices as metalk8s volumes.
- Ensure that prepared volumes are always available even after a restart.
- Ensure that prepared volumes have a valid file system
- Be able to write persistent data on all prepared volumes 

---

Closes: #1474


Ebaneck:
@bert-e wait

salt/metalk8s/volume/init.sls
```diff
@@ -0,0 +1,6 @@
+#
```
jbertran: Here a call on `metalk8s.volume` would prepare then unprepare a volume.

Let's leave this as a prepared volume if you don't specify the underlying state.

salt/metalk8s/volume/files/loopdevice.service.j2
```diff
@@ -0,0 +1,17 @@
+{# source salt linux formulars #}
```
jbertran: Typo: `formulas`

salt/_states/metalk8s_volume.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_storageclass_validity(name, **kwargs):
+    '''
+    Ensures that a storageclass exist and is valid.
+
+    name
+        the name of the storageclass
+    '''
+    # add valid state definition for this
```
jbertran: Is that a TODO?

salt/metalk8s/volume/prepare/init.sls
```diff
@@ -0,0 +1,84 @@
...
+{%- set storageclass_name =  v_info['spec']['storageClassName']['metadata']['name'] %}
+
+#check the validity of specified volume storageclass
+#could refresh pillar here and there will be no need to check storageclass validity
+
+Verify volume storageclass:
+  metalk8s_volume.get_storageclass_validity:
+  - name: {{ storageclass_name }}
+
+    {%- if 'sparseLoopDevice' in v_info['spec'].keys() %}
```
jbertran: More indentation please! It's hard to follow what belongs to the ``sparseLoopDevice` check and what doesn't below.

salt/metalk8s/volume/prepare/init.sls
```diff
@@ -0,0 +1,84 @@
...
+
+# need to convert kubernetes size object(2Gi) to human readable format
+{%- set readable_v_size = metalk8s_volume.to_bytes_converter(v_size) %}
+
+#could check here if system hard disk has sufficient disk space
+# maybe not recommended for a sparsefile
+
+Create sparseLoopDevice {{ v_name }} directory:
+  file.directory:
+  - name: /dev/{{ v_name }}
```
jbertran: Are we OK with crashing in a `file.directory` state if the volume name is already taken? Even if volume name collision checks are handled at a higher level, we should probably check here as well (the usual TOCTOU issues)

salt/metalk8s/volume/prepare/init.sls
```diff
@@ -0,0 +1,84 @@
...
+  - user: root
+  - group: root
+  - dir_mode: 750
+  - makedirs: True
+  - require_in:
+    - file: Create sparsefile {{ v_name }}
+
+Create sparsefile {{ v_name }}:
+  cmd.run:
+  - name: "truncate --size {{ readable_v_size }} {{ v_name }}"
```
jbertran: Shouldn't these be `/dev/{{v_name}}/{{v_name}}`, according to your init scripts?

salt/_modules/metalk8s_volume.py
```diff
@@ -0,0 +1,54 @@
...
+        storageclassname, **kwargs
+    )
+    if storageclass_info is None:
+        return None
+    return storageclass_info
+
+
+# simple helper to convert kubernetes size object to human readable form
+# e.g 43Gi from Kubernetes yields 46170898432
+def to_bytes_converter(quantity):
```
jbertran: I feel like this is a common enough problem to have an existing importable utility, but my google-fu fails me...

@slaperche-scality ?

jbertran (commented):
Some formatting issues, some stray stuff. Please mark comments about stuff remaining to implement as `TODO` so we can clearly differentiate which comments are indicative and which are for ongoing items.

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2728,3 +2728,29 @@ def show_api_service(name, **kwargs):
             raise CommandExecutionError(exc)
     finally:
         _cleanup(**cfg)
+
+
+def read_storage_class(name, **kwargs):
```
NicolasT: Why would we need this? All required info should be in the pillar.

Ebaneck: @NicolasT I am assuming here that at the start of execution we do not refresh pillar and for some reason, a volume in the pillar has a storage class which no longer exist(volume storageclass was deleted).

`But again, I think pillars get refreshed every now and then`

So,
- We either refresh the pillars and make sure that all volumes in the pillars actually have a valid storage class
- Or Check that each volume to be provisioned has a storage class indeed(reason for the above method).

NicolasT: Volumes in the pillar always have a valid storageclass, right? That's checked in the `ext_pillar`.

The main reason I was asking is as follows: I assume you're adding this code to eventually be used in some state and a state SLS. These SLSs are to be applied on minions, which is in the end where the storage sits.

However.

This code connects to the K8s API. And minions can't connect to the K8s API. Hence all info a minion needs that's derived from info that's stored in the K8s API needs to be passed in as pillar values.

salt/_modules/metalk8s_volume.py
NicolasT: Can't find anything in `python-kubernetes`.

salt/metalk8s/volume/files/loopdevice.service.j2
NicolasT: I'm not sure this is the right approach (using a `systemd` unit to set up a loop device).

- We'll have multiple such devices
- How to properly handle cleanup?

Instead, I'd argue we could apply a state at minion startup to get the same thing done.

salt/metalk8s/volume/files/loopdevice.service.j2
```diff
@@ -0,0 +1,17 @@
...
+Description=Setup loop devices for {{ device_name }}
+DefaultDependencies=no
+After=systemd-udev-settle.service
+Before=local-fs.target
+Wants=systemd-udev-settle.service
+
+[Service]
+Type=oneshot
+{# The command is prefixed with '-' to consider it a success if the loopback device is already setup #}
+ExecStart=-/sbin/losetup {{ device_name }} {{ file }}
```
NicolasT: How would you figure out a reasonable `device_name`?

salt/metalk8s/volume/prepare/init.sls
NicolasT: Huh, what's this supposed to do?

salt/metalk8s/volume/prepare/init.sls
NicolasT: We can't have those sparse files somewhere under `/dev` to begin with... `/dev` is not exactly persistent.

salt/metalk8s/volume/prepare/init.sls
```diff
@@ -0,0 +1,84 @@
...
+  - creates: {{ v_name }}
+
+
+add sparsefile_{{ v_name }}_init_script:
+  file.managed:
+  - name: /usr/lib/systemd/system/loopdevice-{{ v_name }}.service
```
NicolasT: Drop-ins don't go in `/usr`.

salt/metalk8s/volume/files/loopdevice.service.j2
Ebaneck: I think we are almost certain that having something like 
`/sbin/losetup/ <volume_uid> <sparsefile_name>` could work no?? Since it is impossible for `device_name` with similar volume_uid or volume_name to exist.

For `device_name` I probably will use the unique `volume_uid`

salt/metalk8s/volume/prepare/init.sls
Ebaneck: Before creating a sparsefile, create a corresponding directory for it in `/dev/volume_name/...`.
`/dev/volume_name/sparsefile_name.img`

salt/_modules/metalk8s_volume.py
slaperche-scality: Not in the stdlib I'm affraid.

As to bring an external dependency for a 10 lines (counting code only) function, I'm not convinced by the JS way of life…

That being said, would be nice to have some unit tests for small helpers like that.
Probably out of scope of this PR, but we could have some doctest at least in the docstring at least (instead of having a simple comment).

salt/metalk8s/volume/prepare/init.sls
Ebaneck: Well noted

I see it should go to `/etc/systemd/system/....`

salt/metalk8s/volume/files/loopdevice.service.j2
Ebaneck: Hmmm, seems we will be having multiple loop devices for one particular volume ??

salt/metalk8s/volume/prepare/init.sls
slaperche-scality: Also, sparse files are supposed to live under `/var/lib/metalk8s/storage/sparse/`

salt/metalk8s/volume/prepare/init.sls
Ebaneck: well noted

salt/metalk8s/volume/prepare/init.sls
NicolasT: You don't ever need to create anything in `/dev` yourself, that's why we have `udev` nowadays. The sparse files are definitely not supposed to be in `/dev`, see other comments.

salt/metalk8s/volume/prepare/init.sls
NicolasT: Also, don't use `cmd.run` with `truncate --size`. https://github.com/scality/metalk8s/issues/1474#issuecomment-518308958

salt/_modules/metalk8s_kubernetes.py
Ebaneck: Thanks for the clarification. It will be useful to simply refresh the pillars hence ensure that all volumes in the pillar are valid(have storageclasses).

salt/metalk8s/volume/prepare/init.sls
Ebaneck: ok

salt/metalk8s/volume/files/loopdevice.service.j2
slaperche-scality: Yes, this may help: https://docs.saltstack.com/en/latest/ref/states/startup.html

Ebaneck: Actually, I see two options here.
- use a startup state as you have suggested or
- use a minion start event 

I am going to experiment with either and see what works well.

slaperche-scality:
Current state:
- [x] Prepare a `sparseLoopDevice`
- [x] Prepare a `rawBlockDevice`
- [x] Calling the state with no arguments prepare all the volumes of the node
- [x] Handle reboot for `sparseLoopDevice`

storage-operator/pkg/salt/client.go
```diff
@@ -53,27 +53,36 @@ func NewClient(creds *Credential) *Client {
...
 ) (string, error) {
 	// Use rand_sleep to emulate slow operation for now
 	payload := map[string]interface{}{
 		"client": "local_async",
 		"tgt":    nodeName,
-		"fun":    "test.rand_sleep",
+		"fun":    "state.sls",
+		"kwarg": map[string]interface{}{
+			"mods":    "metalk8s.volumes",
+			"saltenv": "metalk8s-2.4.0-dev",
```
slaperche-scality: Don't want to hardcode that I guess.

Should I retrieve it from the cluster (with the namespace annotations)?

gdemonet: Indeed, but maybe use the target Node version and not the cluster one? Not sure though :/

salt/metalk8s/volumes/init.sls
```diff
@@ -0,0 +1,2 @@
+include:
+  - .prepare
```
gdemonet: Let's try to fix this early: formulas are usually named as describing a **state**, not an **action**. So that would be `prepared.sls`, and could be accompanied by an `unprepared.sls` counterpart.

slaperche-scality: Ok, will rename it to `prepared`.

As for `unprepared` it's is planned (#1475)

storage-operator/pkg/salt/client.go
TeddyAndrieux: IMO the CRD (and CR, ...) deployed in apiserver should be the one of the cluster_version so I think we should deploy volume using this same version

slaperche-scality: Seen with Teddy: cluster version seems the most pertinent here.

I'll add some code to deal with that.

TeddyAndrieux: Oh sorry misunderstanding what is done in the called state.
Maybe like @gdemonet said we should use the version of the node as it's mainly change on the node.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        return __salt__['disk.fstype'](self.block_device) == fs_type
+
+    def format(self):
+        """Format the volume.
+
+        The volume is formatted according to its StorageClass.
+        """
+        params = self['spec.storageClassName.parameters']
+        # Options are stored as JSON-encoded list.
+        mkfs_options = json.loads(params['mkfsOptions'])
```
NicolasT: What if no `mkfsOptions` are set (which I think should be valid)?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        command = ['mkfs']
```
NicolasT: From `man mkfs`:

>        This mkfs frontend is deprecated in favour of filesystem specific mkfs.<type> utils.

We shouldn't use `mkfs`, but only support some 'known' filesystems, and use their respective `mkfs.TYPE` utilities, and the correct options.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        command.extend(['-t', params['fsType']])
+        command.extend(['-U', self['metadata.uid']])
```
NicolasT: This may be valid for `mkfs.ext4`, but is not valid for e.g. `mkfs.xfs`, hence comment above.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        command.extend(mkfs_options)
+        command.append(self.block_device)
```
NicolasT: In case of `SparseLoopDevice`, we'll want to format the backing file before setting up the loop-device. Or does `udev` get properly triggered after `mkfs` to then create the `/dev/disk/by-uuid/...` symlink?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        _run_cmd(' '.join(command))
+
+    def __getitem__(self, path):
```
NicolasT: That's a very strange/non-standard behaviour/definition of `__getitem__`, not sure I like it.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+
+    @property
+    def size(self):
+        return _quantity_to_bytes(self['spec.sparseLoopDevice.size'])
+
+    @property
+    def exists(self):
+        path = self.sparse_file
+        return os.path.isfile(path) and os.path.getsize(path) == self.size
```
NicolasT: Why the size check for an existence check? What happens if the size doesn't match (for whatever reason)?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def create(self):
+        # Try to create a sparse file, don't clobber existing one!
+        try:
```
NicolasT, line 232: This construct could leak the FD if someone adds code before the next `try` and not taking this into account.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        command = ' '.join(['losetup', '--associated', self.sparse_file])
+        pattern = r'\({}\)'.format(re.escape(self.sparse_file))
+        result  = _run_cmd(command)
+        return re.search(pattern, result['stdout']) is not None
+
+    def initialize(self):
+        # Recent losetup support `--nooverlap` but not the one shipped with
+        # CentOS 7.
+        command = ' '.join(
+            ['losetup', '--find', '--partscan', '--show', self.sparse_file]
```
NicolasT: Why would we need `--partscan`?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+        result  = _run_cmd(command)
+        return result['stdout'].strip()
```
NicolasT: What for? I don't think we should ever use the `/dev/loopN` path.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+
+    @property
+    def block_device(self):
```
NicolasT: I guess this is used to format. I'd rather not do it that way, but format *what we know* (i.e., what's passed in by a user with a `rawBlockDevice`, or in this case, the backing sparse file, which can be formatted in-place, rather than formatting after creating the block device, which is more failure-prone).

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,360 @@
...
+        """Check if the volume is already formatted."""
+        fs_type = self['spec.storageClassName.parameters.fsType']
+        return __salt__['disk.fstype'](self.block_device) == fs_type
+
+    def format(self):
+        """Format the volume.
+
+        The volume is formatted according to its StorageClass.
+        """
+        params = self['spec.storageClassName.parameters']
```
NicolasT: I'd argue this needs a failsafe which checks whether the device is formatted already, and if so, errors out.

Otherwise, if someone creates a `rawBlockDevice` `Volume` for `/dev/sda1` by mistake, we could wipe a whole server (root FS at least). Not cool.

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
```diff
@@ -8,3 +8,10 @@ use_superseded:
   - module.run
 
 log_level_logfile: info
+
+saltenv: {{ saltenv }}
+
+# Prepare volume at startup (required for loop devices persistency).
+startup_states: sls
+sls_list:
+  - /metalk8s/volumes/prepared
```
NicolasT: Is there a way to specify the `saltenv` here instead of setting it globally for the minion?

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -0,0 +1,51 @@
+include:
+  - .installed
+
+Refresh pillars for all minions:
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
```
NicolasT: Why?

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -0,0 +1,51 @@
...
+    - name: /var/lib/metalk8s/storage/sparse/
+    - makedirs: True
+
+{%- for volume in volumes_to_create %}
+
+Create backing storage for {{ volume }}:
+  metalk8s_volumes.present:
+    - volume: {{ volume }}
+    - require:
+      - file: Create the sparse file spool
```
NicolasT: Strictly speaking this is only required for `sparseLoopDevice` volumes, but fine.

salt/metalk8s/volumes/prepared/installed.sls
```diff
@@ -0,0 +1,9 @@
+{%- from "metalk8s/macro.sls" import pkg_installed with context %}
+
+include:
+  - metalk8s.repo
+
+Install e2fsprogs:
```
NicolasT, line 6: Please also support XFS from the start.

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -0,0 +1,51 @@
+include:
+  - .installed
+
+Refresh pillars for all minions:
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
+
+{%- set volumes_to_create = [] %}
```
NicolasT, line 4: Instead of this, could we have a `single` SLS which requires a single volume name to be present and process it (which can then be called from the operator through SaltAPI), and have this one loop over everything, and include `single` for every loop iteration with the loop value in the context as `volume`?

I.e. instead of supporting both handling one single volume and all of them in the same SLS/Jinja template, split it up to make things easier to comprehend.

NicolasT (changes requested):
Good stuff, some comments.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah, I should probably use a "normal" function for that, instead of overloading the behavior of a well known operator.

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
slaperche-scality: @TeddyAndrieux any idea about that?

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: We wanted to be sure that we had an up-to-date pillar at that point, but I guess it's useless/won't do what we want?

storage-operator/pkg/salt/client.go
slaperche-scality: As seen with Nicolas, we want to use the Node version, because that's the one used by the Salt minion so let's stay in sync with it.

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
TeddyAndrieux: From my last search I not found any way to do this. 

I would prefer also to not define it in the minion config, but after some reflection as it's the version of the minion and when we will upgrade/downgrade this node we will also update the minion configuration.... for me it's ok to have it here. (And btw it's maybe also mandatory for startup_states :stuck_out_tongue: )

TeddyAndrieux: Also with this change if salt handle the mount of the volumes that mean if we do a downgrade to something before 2.4 we will no longer mount our volumes (maybe we don't care, maybe not :nerd_face:)

NicolasT: `kubelet` manages mounts, not Salt.

salt/metalk8s/volumes/prepared/init.sls
NicolasT: There should be no need. The pillar should be recalculated when `state.sls` is invoked. Also, the `tgt: '*'` is quite unexpected. Not even sure how this can work in a `state.sls`-interpreted SLS.

salt/metalk8s/volumes/prepared/installed.sls
slaperche-scality: Done

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
NicolasT: Meh, it's fine I guess... in the end, the minion should be in that saltenv in the first place anyway.

salt/_modules/metalk8s_volumes.py
slaperche-scality: If the size doesn't match that means someone other than us created the file, so I wanted to play it safe and don't touch the file.

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: Good catch, I'll will use `params.get('mkfsOptions', '[]')` to be safe.

salt/_modules/metalk8s_volumes.py
slaperche-scality: As seen with you, I misunderstood the role of this option.
We don't need it and I'll remove it (same goes for `--show`).

salt/_modules/metalk8s_volumes.py
slaperche-scality: Actually the return value is not even used by the caller, so yes let's remove it.

slaperche-scality:
We now wandle the reboot for `sparseLoopDevice` so the PR is no longer a draft.

This new patchset started to address some of the comments, but a new patchset is coming soon to address the remaining ones: don't need to re-review now (you can wait the next patchset).

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: Ok, will remove it then.

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
slaperche-scality: Ok, then let's keep it as is (thanks for the feedback @TeddyAndrieux)

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: Yeah, it avoids the need for an `if` in the Jinja and it simplifies the looping when you create all the volumes.

We could also update the module to create the parent directories when creating the sparse file, but that's less explicit.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed & Done.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Solved by using using a contextmanager.

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: Done.

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed, need to format before `losetup` otherwise the symlink won't be created by `udev`.
Done

salt/_states/metalk8s_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+def __virtual__():
+    return __virtualname__
+
+
+def present(name, volume):
+    """Ensure that the backing storage exists for the specified volume.
+
+    Args:
+        name   (str): SLS caller name
+        volume (str): Volume name
```
TeddyAndrieux: It's not common practice in salt normaly name is something usefull and mandatory for state, it's not only a "name"

For me you should remove the volume argument and use name as volume name

TeddyAndrieux: (And it's the same for other function of this file)

salt/_states/metalk8s_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Returns:
+        dict: state return value
+    """
+    ret = {'name': name, 'changes': {}, 'result': False, 'comment': ''}
+    # Idempotence.
+    if __salt__['metalk8s_volumes.exists'](volume):
+        ret['result'] = True
+        ret['comment'] = 'storage for volume {} already exists'.format(volume)
```
TeddyAndrieux: Please capitalize at the beginning of comments

salt/metalk8s/volumes/prepared/init.sls
TeddyAndrieux: But you can't pass information between state.
You only can with orchestrate with pillar but I think we don't want to have an orchestrate to prepare volume as this definitly not need any orchestration

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+
+# }}}
+# SparseLoopDevice {{{
+
+
+class SparseLoopDevice(Volume):
+    @property
+    def path(self):
+        return '/var/lib/metalk8s/storage/sparse/{}'.format(
+            self.get('metadata.name')
```
NicolasT: I'm not entirely sure using `name` is the right approach, I'd argue UUID would be better.

Rationale: one can create a volume with some name, then things can go wrong, one could then (possibly much later) create another volume with the same name, and get a conflict.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+            os.path.isfile(self.path) and
+            os.path.getsize(self.path) == self.size
+        )
+
+    def create(self):
+        # Try to create a sparse file, don't clobber existing one!
+        open_flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
+        try:
+            with _open_fd(self.path, open_flags) as fd:
+                os.ftruncate(fd, self.size)
```
NicolasT: The calculation of `self.size` should be moved *before* the file is opened/created.

Rationale: if `self.size` calculation fails (e.g., invalid format of the value), the `open` will still have happened, so the file will (zero-length) exist on disk. Now when a user would fix the size field, creation will bail out again (size mismatch).

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        except OSError as exn:
+            raise Exception('cannot create sparse file at {}: {}'.format(
+                self.path, exn
+            ))
+
+    @property
+    def is_finalized(self):
```
NicolasT: `finalized` may not be the right term, actually. To me it relates to being removed (e.g., an object finalizer in a GCd language).

NicolasT: `is_provisioned` and `provision`?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        # A sparse loop device is finalized when a sparse file is associated
+        # to a loop device.
+        command = ' '.join(['losetup', '--associated', self.path])
+        pattern = r'\({}\)'.format(re.escape(self.path))
+        result  = _run_cmd(command)
+        return re.search(pattern, result['stdout']) is not None
+
+    def finalize(self):
```
NicolasT: See above.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        return  # Nothing to do
+
+    @property
+    def path(self):
+        return self.get('spec.rawBlockDevice.devicePath')
+
+    def format(self, force=False):
+        # We format an entire device, not just a partition: we need force=True.
```
NicolasT, line 292: That surprises me, to be honest...

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+    """Get a Volume object from the pillar."""
+    volume = __pillar__['metalk8s']['volumes'].get(name)
+    if volume is None:
+        raise ValueError('volume {} not found in pillar'.format(name))
+    if 'rawBlockDevice' in volume['spec']:
+        return RawBlockDevice(volume)
+    elif 'sparseLoopDevice' in volume['spec']:
+        return SparseLoopDevice(volume)
+    else:
+        raise ValueError('unsupported Volume type')
```
NicolasT: Add the volume name in this message.

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+    try:
+        yield fd
+    finally:
+        os.close(fd)
+
+
+# Cannot use `disk.fstype` from Salt because you can't trust its output…
+#
+# Example:
+# - On a unformated sparse file:
```
NicolasT, line 382: Ugh that's terrible :-D

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+#     returns "/dev/sda1      ext4  41152736 7765776  31273476  20% /"
+#
+# - On a raw (non-formatted) block device:
+#     __salt__['disk.fstype']('/dev/sdb')
+#     returns "devtmpfs       devtmpfs   1932084     0   1932084   0% /dev"
+#
+# So yeah, let's not rely on this…
+def _get_fstype(path):
+    try:
+        res = _run_cmd('blkid {}'.format(path))
```
NicolasT: Do we assume this command is pre-installed or should we force-install it, just like `xfsprogs` and such?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        pat = r'TYPE="(?P<fstype>.+)"'
```
NicolasT: Instead of using `blkid` and a regex to parse its output, consider using `lsblk`:
```
$ lsblk -o FSTYPE -r -n /dev/sda2
ext4
```
for actual block devices. Sadly enough, this seems not to work for files:
```
$ truncate -s 1M volume.dat
$ lsblk -o FSTYPE -r -n ./volume.dat
lsblk: ./volume.dat: not a block device
```

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        match = re.search(pat, res['stdout'])
+        return match.groupdict().get('fstype') if match else None
+    except CommandExecutionError:  # Happens on non-formatted sparse file.
```
NicolasT: Could we then assert the given path is a file?

salt/_states/metalk8s_volumes.py
```diff
@@ -0,0 +1,125 @@
...
+        ret['comment'] = 'cannot create storage for volume {}: {}'\
+            .format(volume, exn)
+    else:
+        ret['changes'][volume] = 'Present'
+        ret['result'] = True
+        ret['comment'] = 'storage for volume {} created'.format(volume)
+    return ret
+
+
+def finalized(name, volume):
```
NicolasT: See above.

salt/metalk8s/volumes/prepared/init.sls
NicolasT: Is there no way to 'include' another SLS with a given context dict to render that sub-SLS before pasting in the result?

salt/_modules/metalk8s_volumes.py
```diff
@@ -0,0 +1,430 @@
...
+        """Finalize the setup of the backing storage device."""
+        return
+
+    @abc.abstractproperty
+    def path(self):
+        """Path to the backing device."""
+        return
+
+    @property
+    def is_formatted(self):
```
NicolasT, line 173: A case I believe we're not handling right now, and should do (likely by erroring out): what if the volume exists (and has the expected size and such), it is formatted, but its FS UUID is *not* the UUID we expect it to have?

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -0,0 +1,47 @@
...
+    {%- do volumes_to_create.append(target_volume) %}
+  {%- else %}
+Volume "{{ target_volume }}" not found in pillar:
+  test.fail_without_changes: []
+  {%- endif %}
+{%- else %}
+    {%- do volumes_to_create.extend(all_volumes.keys()|list) %}
+{%- endif %}
+
+Create the sparse file spool:
```
NicolasT: `spool`? `pool`?

storage-operator/pkg/salt/client.go
NicolasT: This is still outstanding?

NicolasT (changes requested):

salt/metalk8s/volumes/prepared/init.sls
TeddyAndrieux: Not as far as I know, I think it's good question for saltstack slack :)

TeddyAndrieux: Seems not supported but asked since long time https://github.com/saltstack/salt/issues/8878

storage-operator/pkg/salt/client.go
slaperche-scality: Yup, but that'll be addressed by my next commit (wasn't ready yet when I made the last push)

salt/_states/metalk8s_volumes.py
slaperche-scality: Ho thanks for the explanation, yeah I'll get rid of the `volume` parameter and use name instead!

salt/_states/metalk8s_volumes.py
slaperche-scality: My rationale here was that you never now how the log message get mix and matched, and thus having proper sentence (starts with a capital, ends with a period) sometime gives weird message in the end.

But yeah, for Salt that's not really an issue so I'll fix that.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Good catch!! I'll move it before the `open`.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah UUID is better, the PR started with `name` and I kept it but as you said: UUID is safer;
I'll make the change.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah, I also wasn't happy with `finalize`…

I'll use `is_provisioned` and `provision`, way better. Thanks!

salt/_modules/metalk8s_volumes.py
slaperche-scality: Me too ! But indeed:

```
# mkfs.ext4 /dev/sdb
mke2fs 1.42.9 (28-Dec-2013)
/dev/sdb is entire device, not just one partition!
Proceed anyway? (y,n)
```

salt/_modules/metalk8s_volumes.py
slaperche-scality: Provided by `util-linux`:

```
# rpm -q --whatprovides /usr/sbin/blkid
util-linux-2.23.2-59.el7_6.1.x86_64
```

I guess we can assume it's here because we require it for the pre-flight.

```yaml
kubeadm_preflight:
  mandatory:
    packages:
      - util-linux            # provides nsenter, mount
      - iproute               # provides ip
      - iptables
```

salt/_modules/metalk8s_volumes.py
slaperche-scality: The output of `lsblk` seems to be variable as well:
```
# lsblk -o FSTYPE -r -n /dev/sda1
ext4
# lsblk -o FSTYPE -r -n /dev/sda  # Why do I have an empty line in that case? oO

ext4
# lsblk -o FSTYPE -r -n /dev/sdb

```
And yeah, it doesn't work on files (even after the formatting it), so I'll stay on `blkid` I think (since it's already installed anyway), otherwise we can't check for the sparse file.

salt/_modules/metalk8s_volumes.py
slaperche-scality: What do you mean?

I'll rewrite this part to be more fine-grained (return code==2 I'll treat as "No FS", return code non-zero and !=2 I'll keep raising).

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: I meant `spool` but after checking it seems I had a wrong definition of this word so I'll use `directory` instead (let's keep things simple).

storage-operator/pkg/salt/client.go
slaperche-scality: Done

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: I guess we will have to keep it like this then, right?

TeddyAndrieux: Yes I think so, except if someone else have an idea

salt/_states/metalk8s_volumes.py
slaperche-scality: Done

salt/_states/metalk8s_volumes.py
slaperche-scality: Done

salt/_modules/metalk8s_volumes.py
slaperche-scality: Hum, we will still have the issue if the `truncate` fails no?

Instead of moving the calcultation, I'll add a try/except around the `truncate` to `os.unlink` in case of error.

NicolasT: But only if the error is not `EEXIST` or whatever, then... Otherwise it kinda defeats the purpose :D

salt/_modules/metalk8s_volumes.py
NicolasT: OK, works for me.

I just wanted to make sure that `CommandExecutionError` (i.e., `rc != 0`) doesn't implicitly mean "no FS present", while there are plenty of other reasons why `blkid` can fail...

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
NicolasT: Looking at the man page, checking for rc 2 catches quite a lot of things as well, including "impossible to gather any information"...

salt/_modules/metalk8s_volumes.py

salt/metalk8s/volumes/prepared/init.sls
NicolasT: No, except writing code. That's really too bad.

salt/_states/metalk8s_volumes.py
slaperche-scality: Done

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: So I'll resolve the conversation.

But yes, it's a pity that we can't "reuse" easily a templated SLS into another one :(

NicolasT: (Or using a Jinja macro, but bleh)

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: Done

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah you're right.

"impossible to gather any information" is a bit like `lsblk` failing on non-block device: in the end you don't know if it's formatted or not :/

I guess we could use `file` but the output format is a mess…

```
# truncate -s 1G bar
# file -sbL bar
data
# truncate -s 2G foo
# mkfs.xfs -f foo > /dev/null
# file -sbL foo
SGI XFS filesystem data (blksz 4096, inosz 512, v2 dirs)
# wipefs -af /dev/sdb > /dev/null
# file -sbL /dev/sdb
data
# mkfs.ext4 -F /dev/sdb > /dev/null
mke2fs 1.42.9 (28-Dec-2013)
# file -sbL /dev/sdb
Linux rev 1.0 ext4 filesystem data, UUID=e3276629-5051-431b-83b5-da808679f77a (extents) (64bit) (large files) (huge files)
```

Could `(?P<fstype>.+) filesystem data` and apply a `tolower` on the result, but that looks more hackish than reliable…

salt/_modules/metalk8s_volumes.py
slaperche-scality: yup, very surprised by the behavior there…

salt/_modules/metalk8s_volumes.py
slaperche-scality: True, but what I had in mind was:

```python
        open_flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
        try:
            with _open_fd(self.path, open_flags) as fd:
                try:
                    os.ftruncate(fd, self.size)
                except Exception:
                    os.unlink(self.path)
                    raise
        except OSError as exn:
            raise Exception('cannot create sparse file at {}: {}'.format(
                self.path, exn
            ))
```
So it should be safe (the unlink only happen if `truncate` (or `self.size`), not for `open`).

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: Maybe `is_formatted` should only check the UUID in fact: that's the proof that we have formatted it ourself.

Checking the FS type can be dangerous no?
Imagine we have a StorageClass `Foo` with `fsType` set to `xfs`, we format the device as described.
Later someone change the `fsType` of the `StorageClass` (I think you can't edit `parameters` of a `StorageClass` but then simply delete/recreate a new with the same name…)  to `ext4`.
Now at the next call for this volume (like when you reboot) we will refresh the pillar, store the new version of the StorageClass and see `xfs != ext4` => not formatted.
Then we go to format => "ho it's already formatted so we bail out".
No data loss (thanks to the safeguard we added after the previous reviews) but an ugly failure.

Or I'm missing something and this scenario can't happen?

slaperche-scality: So I think I'll rewrite `is_formatted` to check the UUID only:
- they match => nothing to do
- they mismatch => either we don't have one (so we should go format) or we have another one (we go format but then we detect the existing FS and bail out => we're safe).

slaperche-scality: Done

salt/_modules/metalk8s_volumes.py

slaperche-scality:
Ok, new version ready for review.

This time everything was addressed.

There are two new commits that may require your attention:
1. https://github.com/scality/metalk8s/pull/1513/commits/ba91f008c93112293d2df567a5a62eb7bacf53ae: this one handle the issue of "We can't run two state in parallel"
2. https://github.com/scality/metalk8s/pull/1513/commits/be6d91d88d9ff65653cb33b166eb9ff796661970: as discussed on Slack this one tries to call into `libblkid` directly instead of calling the shell command

About ①, I'm wondering if we could use [parallel states](https://docs.saltstack.com/en/latest/ref/states/parallel.html), but to do so we would need to move the installation of the packages out of this state (where could we put them?).

salt/_modules/metalk8s_blkid.py
```diff
@@ -0,0 +1,140 @@
...
+So let's directly attack the library, making sure that any API call will result
+in an exception and thus if we end up with "no data" then that truly means
+"everything went well, but there was nothing there".
+"""
+
+import collections
+import ctypes
+import functools
+
+import metalk8s_blkid_c as blkid_c
```
NicolasT: You can't do this in Salt. It may work, but it's not guaranteed to work.

If you have code that's 'utility code', i.e. not 'real' Salt-module functionality, have it as a 'utils' module and consume it from `__utils__`.

Although in this case I'd merge both files in one.

salt/_modules/metalk8s_blkid.py
```diff
@@ -0,0 +1,140 @@
...
+        return cls(probe)
+
+    def __init__(self, probe=None):
+        """Initialize a new probe."""
+        self._probe = blkid_c.new_probe() if probe is None else probe
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
```
NicolasT: I'm going to be a bit pedantic, but because this interacts with system-level C stuff, better be safe than sorry.

So, this `probe` thing clearly has a `free_probe` function for cleanup (it's C, after all). However, the current Python interface leaves ways for this release function never to be called:

- I can call `Probe.from_filename` and never use the result in a context manager (i.e., `__exit__` will never be called).
- I can construct a new `Probe` as-is, and never use the result in a context manager.

Since `free_probe` *must* be (promptly) called after `new_probe` is executed, we should change the API to enforce this.

This is, actually, quite simple: instead of making `Probe` a public type, make it 'private' (rename to `_Probe`, only have the `configure` and `probe` members)), then create two top-level functions that enforce use as a context-manager (can use this utility function in `contextlib` with `yield` to implement them, I forgot the name), one `def probe_frome_filename(filepath)` and one `def probe(probe=None)`, and done.

salt/_modules/metalk8s_blkid.py
```diff
@@ -0,0 +1,140 @@
...
+        if self._probe:
+            blkid_c.free_probe(self._probe)
+            self._probe = None
```
NicolasT: This introduces a second 'issue': when an API allows one to create some value, and then use it as a context manager, nothing prohibits reuse of said value after the context-managed exited. However, if a consumer of this API would do so, things will act really weird.

salt/_modules/metalk8s_blkid.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @_need_valid_probe
```
NicolasT: I guess `need_valid_probe` is related to the face `_probe` *can* be `None`, but never should be.

So, instead of allowing it to be `None`, how about requiring it to be passed to the constructor (but see above, this would be 'private', only used by the context manager functions), in which case it never *can* be `None`, and there's no need to validate this (outside of the constructor), because we tie the lifetime of the probe to the lifetime of this `_Probe` instance (actually, the lifetime of the probe spans the lifetime of the `_Probe` instance).

Granted, this wouldn't prevent nonsense like

```python
p = None
with probe_from_filename("/foo/bar") as q:
    p = q

p.probe()
```

but constructs like the above can't be expected to be OK in general anyway.

salt/_modules/metalk8s_blkid.py
```diff
@@ -0,0 +1,140 @@
...
+    def configure(self, config):
+        """Configure the probe.
+
+        Valid keys for `config` are:
```
NicolasT: That's a very Javascript-esque calling convention. Why not use plain arguments? Much more explicit.

salt/_modules/metalk8s_blkid_c.py
```diff
@@ -0,0 +1,138 @@
+# coding: utf-8
+"""Thin wrapper exposing libblkid functions."""
+
+import ctypes
+import ctypes.util
+
+soname = ctypes.util.find_library('blkid')
+if soname is None:
+    raise Exception('cannot find blkid library')
```
NicolasT: Would an `ImportError` make sense?

salt/_modules/metalk8s_blkid_c.py
```diff
@@ -0,0 +1,138 @@
...
+PARTS_ENTRY_DETAILS = 1 << 2
+PARTS_MAGIC         = 1 << 3
+
+
+# }}}
+# Superblocks prober {{{
+
+
+# blkid_probe_enable_superblocks
+probe_enable_superblocks= blkid.blkid_probe_enable_superblocks
```
NicolasT: Missing space before `=`

salt/_modules/metalk8s_volumes.py
```diff
@@ -13,6 +13,8 @@
 
 from salt.exceptions import CommandExecutionError
 
+import metalk8s_blkid as blkid
```
NicolasT: Again, can't do this. Need to look things up from `__salt__` or something across modules IIRC.

salt/_modules/metalk8s_volumes.py
```diff
@@ -180,7 +182,8 @@ def format(self, force=False):
         """
         # Check that the backing device is not already formatted.
         # Bail out if it is: we don't want data loss because of a typo…
-        if _get_from_blkid(self.path, 'TYPE'):
+        device_info = _get_from_blkid(self.path)
+        if device_info.fstype or device_info.has_partition:
             raise Exception('backing device `{}` already formatted'.format(
```
NicolasT: Exception no longer matches the causes for errors. Split them up in 2 cases with an appropriate error message.

salt/_modules/metalk8s_volumes.py
```diff
@@ -382,20 +385,17 @@ def _open_fd(*args, **kwargs):
...
+def _get_from_blkid(path):
+    probe_config = {
+        'superblocks': True,
+        'superblocks_flags': [
+            blkid.SuperblockFlags.UUID, blkid.SuperblockFlags.TYPE
+        ],
+        'partitions': True,
+    }
+    with blkid.Probe.from_filename(path) as probe:
+        probe.configure(probe_config)
```
NicolasT: Would it make sense to allow configuration from the 'creation' function directly?

NicolasT (changes requested):

salt/_modules/metalk8s_blkid.py
slaperche-scality: I'll move those into `_utils` then.

I split the code between two files to separate the "low-level" binding that only defines the "interfacing" from the "high-level" binding that provides a Python-like interface on top of it.

But yeah, I can merge those, since I'll already using folds I can use them to keep the separate.

Done

salt/_modules/metalk8s_blkid_c.py
slaperche-scality: Yup, I think it's better to use `ImportError` here.
Done.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yup, now using `__utils__`.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Having a FS or a partition table are both a proof that the device is already formatted, no?
Do we really want to make a distinction here?

salt/_modules/metalk8s_blkid.py
slaperche-scality: Yeah, that's why I added this `need_valid_probe` band-aid thing: to avoid segfault if one reuse the object after the `with`.

But it won't be needed after reworking the API as suggested in another comment.

salt/_modules/metalk8s_blkid.py
slaperche-scality: Very good point!

I'll rework the interface to enforce correct use.

salt/_modules/metalk8s_volumes.py
NicolasT: Yes.

Either it's a device that's already formatted but not 'by us', so there's a 'conflict'.

Or, the device path we got (e.g. in case of a `rawBlockDevice`) is a block device that contains partitions. As such this is more likely a configuration error (someone provided `/dev/sda` whilst it should've been `/dev/sda1`, for example).

Granted, in both cases there's nothing *we* can do except for raising an error. However, the root cause why such exception is raised is quite different, and pointing the user in the right direction ('Already formatted' vs. 'Contains partitions, not something we can format') is useful.

salt/_modules/metalk8s_blkid.py
NicolasT: Making 'invalid state' unrepresentable (through 'expected' use of an API, of course... in Python you can't prohibit someone doing crazy things) always trumps runtime validation :)

salt/_modules/metalk8s_blkid_c.py
slaperche-scality: Done

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yup, I'll move it there.

Done.

salt/_modules/metalk8s_blkid.py
slaperche-scality: I'll replace the dict by a namedtuple, more explicit (while avoiding long signature/long function call).

NicolasT: That would still require one to look at the documentation instead of the function prototype to know what defaults are.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed, will do.

salt/_modules/metalk8s_blkid.py
slaperche-scality: Fair enough. Done

salt/_modules/metalk8s_blkid.py
slaperche-scality: Removed by reworking the API

salt/_utils/volume_utils.py
```diff
@@ -0,0 +1,297 @@
...
+    ):
+        """Initialize a new probe for the device at `filepath`."""
+        self._probe = new_probe_from_filename(filepath)
+        self._configure(
+            use_superblocks, superblocks_flags, use_partitions, partitions_flags
+        )
+
+    def destroy(self):
+        """Free the resources allocated by the probe."""
+        free_probe(self._probe)
```
NicolasT: And now you should set `self._probe` to None (and check for it before), otherwise one can call `destroy` (which is public API!), and cause a double-free.

NicolasT: I'd argue the constructor should take a `probe` which is created externally to this class code, and then create/destroy the C-level 'probe' inside the context manager code.

salt/_utils/volume_utils.py
```diff
@@ -0,0 +1,297 @@
...
+        probe.destroy()
+
+
+def get_superblock_flags(*args):
+    return _get_flags(
+        SuperblockFlags, 'superblocks', SuperblockFlags.DEFAULT, *args
+    )
+
+
+def get_parition_flags(*args):
```
NicolasT: `parition`?

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
```diff
@@ -37,4 +37,7 @@ external_auth:
       - '@jobs'
     storage-operator:
       - '*':
-        - 'test.ping'
+        - 'test.rand_sleep'
+        - 'disk.dump'
+        - 'state.sls'
```
NicolasT, line 42: We can further limit this, please file a ticket.

storage-operator/deploy/role.yaml
```diff
@@ -59,3 +60,9 @@ rules:
   - '*'
   verbs:
   - '*'
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - '*'
```
NicolasT: Only `GET`, no?

NicolasT (changes requested):
Minor, mostly good.

salt/_utils/volume_utils.py
slaperche-scality: meh, typo >_<
Done.

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
slaperche-scality: Like that:

```yaml
        - 'test.rand_sleep'
        - 'disk.dump'
        - 'state.sls':
            args:
              - 'metalk8s.volumes'
```
?

storage-operator/deploy/role.yaml
slaperche-scality: I'll try, but yeah that should be enough.

salt/_utils/volume_utils.py
slaperche-scality: Done

storage-operator/deploy/role.yaml
slaperche-scality: Not enough, need `get`, `list` and `watch`.

Done

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
NicolasT: Something along those lines. Not required in this PR as far as I'm concerned, but would be nice to limit access.

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -213,6 +213,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 		scheme:   mgr.GetScheme(),
 		recorder: mgr.GetRecorder("volume-controller"),
 		salt:     salt.NewClient(getAuthCredential(mgr.GetConfig())),
+		saltenv:  "",
```
NicolasT: This can't be right, similar how `fetchSaltEnv` can't be right accordingly.

This structure is global to the operator. However, we retrieve a `saltenv` from a `Node`, as intended: the `saltenv` to be used when invoking Salt can be different from one `Volume` to another, when those sit on different `Node`s.

So we'd need something like `getNodeSaltenv` or something (not even cached, as far as I'm concerned), and use the resulting value when invoking Salt, within the scope of handling a single `Volume`.

NicolasT (changes requested):
I believe there's one last bug, otherwise LGTM, but would appreciate a second +1.

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: It was working fine because `saltenv` was recomputed at each reconciliation, but it was confusing.
I'll remove the field and use a local variable in the reconciliation loop instead.

Done.

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
slaperche-scality: Tracked by #1528

salt/metalk8s/orchestrate/etcd.sls
```diff
@@ -32,6 +32,7 @@ Check etcd cluster health for {{ node }}:
     - metalk8s_etcd.check_etcd_health:
       - minion_id: {{ node }}
     - attemps: 5
+    - until: "cluster is healthy"
```
NicolasT: Where does this come from?

slaperche-scality: Was from a TEST commit where I was trying stuff to make the expansion working.

The idea was to execute `state.running` to check if something was running after restarting the minion (if not == then startup is over, we can proceed with the other states).

To do so, I had to add the support for `until` to `metalk8s.module_run` and thus update its use here.

But in the end it didn't work, so it was removed ;)

Now trying to use a dedicated parallel state for startup (seems to be working so far).

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
```diff
@@ -14,4 +14,4 @@ saltenv: {{ saltenv }}
 # Prepare volume at startup (required for loop devices persistency).
 startup_states: sls
 sls_list:
-  - /metalk8s/volumes/prepared
+  - /metalk8s/volumes/provisioned
```
NicolasT, line 17: Applying this state may easily fail (potentially only partially). Is that handled gracefully by Salt?

NicolasT: What I'm aiming at: it's possible `provision` fails because `exists` is `False` (e.g., we didn't prepare the sparse file yet).

TeddyAndrieux: I think the output of the state is just ignored

NicolasT: If that's the case (and logs aren't being spammed), then that's OK for me.

slaperche-scality: That doesn't prevent the minion to start.

Just have an error logged:

```
[DEBUG   ] minion return: {u'fun_args': [[u'/metalk8s/volumes/provisioned']], u'jid': u'req', u'return': {u'test_|-Provision backing storage for kotlet_|-Provision backing storage for kotlet_|-fail_without_changes': {u'comment': u'Failure!', u'name': u'Provision backing st
orage for kotlet', u'start_time': '15:07:37.884802', u'result': False, u'duration': 0.948, u'__run_num__': 0, u'__sls__': u'/metalk8s/volumes/provisioned', u'changes': {}, u'__id__': u'Provision backing storage for kotlet'}}, u'retcode': 2, u'success': True, u'fun': u'stat
e.sls'}
```

NicolasT:
I think this can be +1'd?

NicolasT (approved):

jbertran (approved):

PR #1522, opened 2019-08-13
Improvement/1506 delete volume in volumelist

### LINKED ISSUE #1506: Delete volume from volumelist
**Component**: UI, storage

**Why this is needed**:  
As a user, I want to delete volume from UI.

**What should be done**:  
In the volume list, I can delete the volume by click on the DELETE TARSH BUTTON at the right side of each volume when the volume is deletable. Otherwise, the DELETE TARSH BUTTON  will be disabled and when the user hovers on the button it will give more information.

When we disable the DELETE TARSH BUTTON?

The status of volume is `PENDING, TERMINATING, UNKNOWN`. In order to give some information to the user, we will show a dialog when the user hovers on the button.
`PENDING => Create volume is on progress`
`TERMINATING => Delete volume in on progress`
`UNKNOWN => The volume is on transient status`

When we enable the DELETE TARSH BUTTON?

The status of volume is `FAILED, AVAILABLE` we need a further check on PV.
  FAILED => no PV => YES
  FAILED => have PV => if the status of PV is `AVAILABLE, RELEASED, FAILED`=> YES
  FAILED => have PV => if the status of PV is `PENDING, BOUND`=> NO => show the message `The volume is busy`

The step of delete volume(with mockup)
1.  We can find the DELETE TARSH BUTTON on the right side of each volume( If the volume is deletable, otherwise we may want to explain to the user why it cannot be deleted)
image

2. When we click on the DELETE button, there will be a modal pop up to help users confirm if they really want to delete the certain volume.
image

3. After the confirmation, there will be a notification pop up. And the status of the volume will change to `TERMINATING`
image

In order to delete the volume, you need to use ```Custom_objectsApiapi```
```extension = config.makeApiClient(Custom_objectsApi);```

```javascript
const ressourceDeleted = extension.deleteClusterCustomObject(
    'storage.metalk8s.scality.com',
    'v1alpha1',
    'volumes',
    'yanjin-test',
     {}
  );
```

**Test plan**: 
Create unit test for the deleteVolume saga.

ChengYanJin:
**Component**: ui, storage

**Context**:  
As a user, I would like to delete the volume from the volumelist.

**Summary**: 

**Acceptance criteria**: 

- If the volume is deletable when you click on the delete button, it should popup a modal to help you confirm. 

image

- If the volume is undeletable when you hover on the delete button, it should display the hint that explains why you cannot delete this volume.

image
 
- There will be a notification pop up.

image

Closes: #1506

### DIFF
--- a/ui/src/constants.js
+++ b/ui/src/constants.js
@@ -12,9 +12,9 @@ export const STATUS_BANNER_WARNING = 'banner_warning';
 export const STATUS_BANNER_ERROR = 'banner_error';
 export const STATUS_BANNER_SUCCESS = 'banner_success';
 
-export const STATUS_READY = 'ready';
-export const STATUS_NOT_READY = 'not_ready';
-export const STATUS_API_UNKNOWN = 'unknown';
+export const API_STATUS_READY = 'ready';
+export const API_STATUS_NOT_READY = 'not_ready';
+export const API_STATUS_UNKNOWN = 'unknown';
 
 export const STATUS_BOUND = 'Bound';
 export const STATUS_PENDING = 'Pending';
--- a/ui/src/constants.js
+++ b/ui/src/constants.js
@@ -14,14 +14,15 @@ export const STATUS_BANNER_SUCCESS = 'banner_success';
 
 export const STATUS_READY = 'ready';
 export const STATUS_NOT_READY = 'not_ready';
-export const STATUS_UNKNOWN = 'unknown';
+export const STATUS_API_UNKNOWN = 'unknown';
+
 export const STATUS_BOUND = 'Bound';
 export const STATUS_PENDING = 'Pending';
 export const STATUS_TERMINATING = 'Terminating';
 export const STATUS_FAILED = 'Failed';
 export const STATUS_AVAILABLE = 'Available';
 export const STATUS_RELEASED = 'Released';
-export const STATUS_VOLUME_UNKNOWN = 'Unknown';
+export const STATUS_UNKNOWN = 'Unknown';
 
 export const SPARSE_LOOP_DEVICE = 'sparseLoopDevice';
 export const RAW_BLOCK_DEVICE = 'rawBlockDevice';
--- a/ui/src/constants.js
+++ b/ui/src/constants.js
@@ -16,6 +16,12 @@ export const STATUS_READY = 'ready';
 export const STATUS_NOT_READY = 'not_ready';
 export const STATUS_UNKNOWN = 'unknown';
 export const STATUS_BOUND = 'Bound';
+export const STATUS_PENDING = 'Pending';
+export const STATUS_TERMINATING = 'Terminating';
+export const STATUS_FAILED = 'Failed';
+export const STATUS_AVAILABLE = 'Available';
+export const STATUS_RELEASED = 'Released';
+export const STATUS_VOLUME_UNKNOWN = 'Unknown';
 
 export const SPARSE_LOOP_DEVICE = 'sparseLoopDevice';
 export const RAW_BLOCK_DEVICE = 'rawBlockDevice';
--- a/ui/src/containers/NodeVolumes.js
+++ b/ui/src/containers/NodeVolumes.js
@@ -142,7 +142,7 @@ const NodeVolumes = props => {
               return false;
             default:
               console.error(
-                `Unexpected state for PersistentVolume ${persistentVolume}:${persistentVolumeStatus}`
+                `Unexpected state for PersistentVolume ${volumeName}:${persistentVolumeStatus}`
               );
               return false;
           }
--- a/ui/src/containers/NodeVolumes.js
+++ b/ui/src/containers/NodeVolumes.js
@@ -4,6 +4,8 @@ import { injectIntl } from 'react-intl';
 import styled from 'styled-components';
 import { FormattedDate, FormattedTime } from 'react-intl';
 import { withRouter } from 'react-router-dom';
+import Modal from 'react-modal';
+import Tooltip from '@material-ui/core/Tooltip';
 import { Button, Table, Loader } from '@scality/core-ui';
 import { padding } from '@scality/core-ui/dist/style/theme';
 import NoRowsRenderer from '../components/NoRowsRenderer';
@@ -18,6 +20,14 @@ import {
   refreshPersistentVolumesAction,
   stopRefreshPersistentVolumesAction
 } from '../ducks/app/volumes';
+import {
+  STATUS_VOLUME_UNKNOWN,
+  STATUS_TERMINATING,
+  STATUS_PENDING,
+  STATUS_FAILED,
+  STATUS_AVAILABLE,
+  STATUS_BOUND
+} from '../constants';
 
 const ButtonContainer = styled.div`
   margin-top: ${padding.small};
@@ -38,18 +48,58 @@ const NodeVolumes = props => {
     stopRefreshPersistentVolumesAction
   );
   const volumes = useSelector(state => state.app.volumes);
+  const persistentVolumes = useSelector(state => state.app.volumes.pVList);
   const [sortBy, setSortBy] = useState('name');
   const [sortDirection, setSortDirection] = useState('ASC');
+  const [
+    isDeleteConfirmationModalOpen,
+    setisDeleteConfirmationModalOpen
+  ] = useState(false);
+
   const onSort = ({ sortBy, sortDirection }) => {
     setSortBy(sortBy);
     setSortDirection(sortDirection);
   };
 
+  const isVolumeDeletable = rowData => {
+    const volumeStatus = rowData.status;
+    const volumeName = rowData.name;
+
+    if (
+      volumeStatus === STATUS_VOLUME_UNKNOWN ||
+      volumeStatus === STATUS_PENDING ||
+      volumeStatus === STATUS_TERMINATING
+    ) {
+      return false;
+    } else if (
+      volumeStatus === STATUS_FAILED ||
+      volumeStatus === STATUS_AVAILABLE
+    ) {
+      if (persistentVolumes === []) {
+        return true;
+      } else {
+        const persistentVolume = persistentVolumes.find(
+          pv => pv?.metadata?.name === volumeName
+        );
+        const persistentVolumeStatus = persistentVolume?.status?.phase;
+        if (
+          persistentVolumeStatus === STATUS_BOUND ||
+          persistentVolumeStatus === STATUS_AVAILABLE
+        ) {
+          return true;
+        } else if (persistentVolumeStatus === STATUS_PENDING) {
+          return false;
+        } else {
+          return true;
+        }
+      }
+    }
+  };
+
   const columns = [
     {
       label: props.intl.messages.name,
-      dataKey: 'name',
-      flexGrow: 1
+      dataKey: 'name'
     },
     {
       label: props.intl.messages.status,
@@ -66,9 +116,10 @@ const NodeVolumes = props => {
     {
       label: props.intl.messages.creationTime,
       dataKey: 'creationTime',
+      flexGrow: 1,
       renderer: data => (
         <span>
-          <FormattedDate value={data} />
+          <FormattedDate value={data} />{' '}
           <FormattedTime
             hour="2-digit"
             minute="2-digit"
@@ -77,6 +128,38 @@ const NodeVolumes = props => {
           />
         </span>
       )
+    },
+    {
+      label: 'Action',
+      dataKey: 'action-test',
+      disableSort: true,
+      width: 80,
+      renderer: (data, rowData) => {
+        const isEnableClick = isVolumeDeletable(rowData);
+        const CustI = styled.i`
+          cursor: ${props => {
+            if (props.isEnableClick) {
+              return `pointer`;
+            } else {
+              return `not-allowed`;
+            }
+          }};
+        `;
+        return (
+          <div>
+            <Tooltip
+              title="Why it cannot be deleted?"
+              onClick={e => {
+                e.stopPropagation();
+                setisDeleteConfirmationModalOpen(true);
+              }}
+              disableHoverListener={isEnableClick}
+            >
+              <CustI className="fas fa-trash" isEnableClick={isEnableClick} />
+            </Tooltip>
+          </div>
+        );
+      }
     }
   ];
   const onRowClick = row => {
@@ -97,6 +180,7 @@ const NodeVolumes = props => {
 
   return (
     <>
+      <Modal isOpen={isDeleteConfirmationModalOpen} ariaHideApp={false} />
       <ButtonContainer>
         <Button
           text={props.intl.messages.create_new_volume}
--- a/ui/src/containers/NodeVolumes.js
+++ b/ui/src/containers/NodeVolumes.js
@@ -41,6 +41,9 @@ const ButtonContainer = styled.div`
 const VolumeTable = styled.div`
   flex-grow: 1;
   margin-top: ${padding.small};
+  .sc-table-column-cell-container-action {
+    justify-content: center;
+  }
 `;
 
 const DeleteConfirmationModal = {
@@ -136,13 +139,18 @@ const NodeVolumes = props => {
               return true;
             case STATUS_PENDING:
             case STATUS_BOUND:
+              return false;
             default:
-              console.error('Error! There is error from PV status.');
+              console.error(
+                `Unexpected state for PersistentVolume ${persistentVolume}:${persistentVolumeStatus}`
+              );
               return false;
           }
         }
       default:
-        console.error('Error! There is error from volume status.');
+        console.error(
+          `Unexpected state for Volume ${volumeName}:${volumeStatus}`
+        );
         return false;
     }
   };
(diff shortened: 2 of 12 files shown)


ui/src/containers/NodeVolumes.js
```diff
@@ -1,9 +1,11 @@
 import React, { useState } from 'react';
-import { useSelector } from 'react-redux';
+import { connect, useSelector } from 'react-redux';
 import { injectIntl } from 'react-intl';
 import styled from 'styled-components';
 import { FormattedDate, FormattedTime } from 'react-intl';
 import { withRouter } from 'react-router-dom';
+import Modal from 'react-modal';
+import Tooltip from '@material-ui/core/Tooltip';
```
carlito-scality: Why do we need a new library for Tooltip? we have already ```rc-tooltip``` and it is not reusable as it is. I think we should implement a Tooltip based on ```rc-tooltip``` or from scratch using styled-component in core-ui which covers all the use cases we need.

ui/src/containers/NodeVolumes.js
```diff
@@ -95,11 +229,47 @@ const NodeVolumes = props => {
...
     <>
+      <Modal
+        isOpen={isDeleteConfirmationModalOpen}
+        ariaHideApp={false}
+        style={DeleteConfirmationModal}
+      >
+        <h2>{intl.messages.delete_a_volume}</h2>
+        <div>{intl.messages.delete_a_volume_warning}</div>
+        <div>
+          {intl.messages.delete_a_volume_confirm}{' '}
```
carlito-scality: Should you pass ```deleteVolumeName``` as ```delete_a_volume_confirm``` 's argument

ui/src/containers/NodeVolumes.js
```diff
@@ -95,11 +229,47 @@ const NodeVolumes = props => {
...
+    setisDeleteConfirmationModalOpen(false);
+  };
+
+  const onClickCancelButton = () => {
+    setisDeleteConfirmationModalOpen(false);
+  };
+
   return (
     <>
+      <Modal
```
carlito-scality, line 296: - It would be nice to create a reusable component and It should be moved to ```core-ui```.
something like:
```
<Modal>
    <ModalTitle>{props.title}</ModalTitle>
    <ModalBody>{props.body}</ModalBody>
    <ModalFooter>{props.footer}</ModalFooter>
</Modal>
```
- we need to have only one instance of "Modal" per application like Notifications, but it can be managed later.

carlito-scality (dismissed):

ui/src/constants.js
```diff
@@ -16,6 +16,12 @@ export const STATUS_READY = 'ready';
 export const STATUS_NOT_READY = 'not_ready';
 export const STATUS_UNKNOWN = 'unknown';
 export const STATUS_BOUND = 'Bound';
+export const STATUS_PENDING = 'Pending';
+export const STATUS_TERMINATING = 'Terminating';
+export const STATUS_FAILED = 'Failed';
+export const STATUS_AVAILABLE = 'Available';
+export const STATUS_RELEASED = 'Released';
+export const STATUS_VOLUME_UNKNOWN = 'Unknown';
```
slaperche-scality: It's a bit weird to have `STATUS_UNKNOWN` and `STATUS_VOLUME_UNKNOWN`, where the real difference is not really about volume but "ha this one starts with an uppercase".

If in the future, if we have an object like `ServiceAccount` and it also uses "Unknown" then we will have a weird code that check against `STATUS_VOLUME_UNKNOWN` for something that's not a volume.

I don't have a really good solution, we could:
- have a set of constants (I don't know if JS supports `enum`? Seems not, but can be [emulated](https://stijndewitt.com/2014/01/26/enums-in-javascript/)) per object
- or use caseless comparison for status
- or an even better idea?

ui/src/translations/fr.json
```diff
@@ -92,5 +92,17 @@
   "generic_missing_field": "Veuillez entrer un {field}.",
   "no_storage_class_found": "Aucune classe de stockage trouvée.",
   "storage_class_is_required": "Une classe de stockage est requise pour créer un volume. Veuillez en créer une dans Kubernetes.",
-  "learn_more": "En savoir plus"
+  "learn_more": "Learn more",
```
slaperche-scality: Oups, I think you want to keep `En savoir plus` here ^^

ui/src/translations/fr.json
```diff
@@ -92,5 +92,17 @@
...
+  "volume_deletion": "Suppression du volume",
+  "volume_delete_success": "Volume {name} est en cours de suppression.",
+  "volume_delete_failed": "La suppression du volume {name} a échoué.",
+  "volume_status_pending_hint": "Création du volume en cours.",
+  "volume_status_terminating_hint": "Suppression du volume en cours.",
+  "volume_status_unknown_hint": "Le volume est dans un état transitoire.",
+  "volume_status_failed_hint": "Le volume est occupé.",
+  "delete_a_volume": "Supprimer un volume",
+  "delete_a_volume_warning": "La suppression de ce volume entraînera la perte des données qu'il contient.",
+  "delete_a_volume_confirm": "Voulez-vous supprimez"
```
slaperche-scality: We probably need a question mark here: `Voulez-vous supprimez ?`

ui/src/containers/NodeVolumes.js
ChengYanJin: > Why do we need a new library for Tooltip? we have already `rc-tooltip` and it is not reusable as it is. I think we should implement a Tooltip based on `rc-tooltip` or from scratch using styled-component in core-ui which covers all the use cases we need.

Thanks Carlito. I didn't know we already have tooltip before. Will use rc-tooltip then.

ui/src/translations/en.json
```diff
@@ -92,5 +92,17 @@
...
+  "volume_deletion": "Volume deletion",
+  "volume_delete_success": "Volume {name} is being deleted.",
+  "volume_delete_failed": "Volume {name} deleted failed.",
+  "volume_status_pending_hint": "Volume creation is in progress.",
+  "volume_status_terminating_hint": "Volume deletion is in progress.",
+  "volume_status_unknown_hint": "The volume is in a transient state.",
+  "volume_status_failed_hint": "The volume is busy.",
+  "delete_a_volume": "Delete a volume",
+  "delete_a_volume_warning": "Deleting this volume will permanently delete the data it contains.",
+  "delete_a_volume_confirm": "Do you want to delete"
```
slaperche-scality: We probably need a question mark here: `Do you want to delete?`

ui/src/translations/en.json
```diff
@@ -92,5 +92,17 @@
   "generic_missing_field": "Please enter a {field}.",
   "no_storage_class_found": "No storage class found.",
   "storage_class_is_required": "A storage class is required in order to create a volume. Please create one in Kubernetes.",
-  "learn_more": "Learn more"
+  "learn_more": "Learn more",
+  "action": "Action",
+  "delete": "Delete",
+  "volume_deletion": "Volume deletion",
+  "volume_delete_success": "Volume {name} is being deleted.",
+  "volume_delete_failed": "Volume {name} deleted failed.",
```
slaperche-scality: `Volume {name} deletion failed.`

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +42,109 @@ const VolumeTable = styled.div`
...
+      if (persistentVolumes === []) {
+        return true;
+      } else {
+        const persistentVolume = persistentVolumes.find(
+          pv => pv?.metadata?.name === volumeName
+        );
+        const persistentVolumeStatus = persistentVolume?.status?.phase;
+        if (
+          persistentVolumeStatus === STATUS_BOUND ||
+          persistentVolumeStatus === STATUS_AVAILABLE
```
slaperche-scality: PV can also be `Released` and it's fine to delete them.

See [here](https://godoc.org/k8s.io/api/core/v1#PersistentVolumePhase) for a list of PV status:

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +42,109 @@ const VolumeTable = styled.div`
...
   const onSort = ({ sortBy, sortDirection }) => {
     setSortBy(sortBy);
     setSortDirection(sortDirection);
   };
 
+  const isVolumeDeletable = rowData => {
+    const volumeStatus = rowData.status;
+    const volumeName = rowData.name;
+
+    if (
```
slaperche-scality: Why not using switch here?
Would be more readable than `if/else`.

This kind of control flow (branching on a finite set of value) is one of the best use case for switch.

e.g.:

```javascript
switch (volumeStatus) {
  case STATUS_VOLUME_UNKNOWN:
  case STATUS_PENDING:
  case STATUS_TERMINATING:
    return false;
  case STATUS_FAILED:
  case STATUS_AVAILABLE: {
    const persistentVolume = persistentVolumes.find(
      pv => pv?.metadata?.name === volumeName
    );
    const persistentVolumeStatus = persistentVolume?.status?.phase;
    switch (persistentVolumeStatus) {
      case STATUS_FAILED:
      case STATUS_AVAILABLE:
      case STATUS_RELEASED:
        return true;
      case STATUS_PENDING:
      case STATUS_BOUND:
        return false;
      default:
        // Log a warning/error here, not supposed to happen.
        return false;
    }
  }
  default:
    // Log a warning/error here, not supposed to happen.
    return false;
}
```

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +42,109 @@ const VolumeTable = styled.div`
...
+    ) {
+      if (persistentVolumes === []) {
+        return true;
+      } else {
+        const persistentVolume = persistentVolumes.find(
+          pv => pv?.metadata?.name === volumeName
+        );
+        const persistentVolumeStatus = persistentVolume?.status?.phase;
+        if (
+          persistentVolumeStatus === STATUS_BOUND ||
```
slaperche-scality: For `BOUND` you want to return `false` (a PV Bound is a PV in use => cannot delete).

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +42,109 @@ const VolumeTable = styled.div`
...
+          persistentVolumeStatus === STATUS_AVAILABLE
+        ) {
+          return true;
+        } else if (persistentVolumeStatus === STATUS_PENDING) {
+          return false;
+        } else {
```
slaperche-scality, line 129: for the default case I would play it safe and return `false`.

ui/src/containers/NodeVolumes.js
```diff
@@ -77,6 +153,64 @@ const NodeVolumes = props => {
...
+            case STATUS_PENDING:
+              hintMessage = intl.messages.volume_status_pending_hint;
+              break;
+            case STATUS_TERMINATING:
+              hintMessage = intl.messages.volume_status_terminating_hint;
+              break;
+            case STATUS_VOLUME_UNKNOWN:
+              hintMessage = intl.messages.volume_status_unknown_hint;
+              break;
+            case STATUS_FAILED:
```
slaperche-scality, line 219: I'm not sure the string is good here: a `Failed` volume is not busy (and if the `PV` is not bound you can delete it)

If the volume is `Available` or `Failed` you need to check the `PV` to know what happen:
- PV is `Bound` => we should display a message saying `Volume is bound, cannot delete`
- PV is `Pending` => here is the `busy` case I guess (shouldn't happen so don't have a better message for this case).

slaperche-scality (changes requested):
Haven't tested yet, only code review (in my limited capacity).

Feel free to come to me if some Status/Phase are unclear.

ui/src/constants.js
ChengYanJin: We changed the `STATUS_UNKNOWN` to `STATUS_API_UNKNOWN` because the lowercase `unknow` is come from the API.  So later we can reuse STATUS_UNKNOWN in any other object.

slaperche-scality: Works for me :slightly_smiling_face:

ui/src/containers/NodeVolumes.js
ChengYanJin: The reason I didn't pass `deleteVolumeName` inside the translation is that I would like to give bold to the deleteVolumeName in the Modal. So I will keep in this way ^^

ui/src/containers/NodeVolumes.js
ChengYanJin: Will improve it later. Thanks

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +43,134 @@ const VolumeTable = styled.div`
...
+            pv => pv?.metadata?.name === volumeName
+          );
+          const persistentVolumeStatus = persistentVolume?.status?.phase;
+
+          switch (persistentVolumeStatus) {
+            case STATUS_FAILED:
+            case STATUS_AVAILABLE:
+            case STATUS_RELEASED:
+              return true;
+            case STATUS_PENDING:
```
slaperche-scality, line 140: Hum, you'll want:

```javascript
            case STATUS_PENDING:
            case STATUS_BOUND:
              return false;
            default:
              console.error('Error! There is error from PV status.');
              return false;
```

Otherwise you'll log an error also for `STATUS_PENDING` and `STATUS_BOUND`: we don't want that.

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +43,134 @@ const VolumeTable = styled.div`
...
+            case STATUS_BOUND:
+            default:
+              console.error('Error! There is error from PV status.');
```
slaperche-scality: Don't need `Error` in the message: you already use `console.error`.

The message could be: `Unexpected state for PersistentVolume {here put pv.name}: {here put pv.status}.`

ui/src/containers/NodeVolumes.js
```diff
@@ -31,44 +43,134 @@ const VolumeTable = styled.div`
...
+              return false;
+          }
+        }
+      default:
+        console.error('Error! There is error from volume status.');
```
slaperche-scality: Ditto.


The message could be: `Unexpected state for Volume {here put volume.name}: {here put volume.status}.`

ui/src/containers/NodeVolumes.js
```diff
@@ -77,8 +179,85 @@ const NodeVolumes = props => {
...
+            case STATUS_PENDING:
+              hintMessage = intl.messages.volume_status_pending_hint;
+              break;
+            case STATUS_TERMINATING:
+              hintMessage = intl.messages.volume_status_terminating_hint;
+              break;
+            case STATUS_UNKNOWN:
+              hintMessage = intl.messages.volume_status_unknown_hint;
+              break;
+            case STATUS_FAILED:
```
slaperche-scality, line 219: This should be:

```
case STATUS_FAILED:
case STATUS_AVAILABLE:
```

Otherwise you'll fall into the `default` case l.229 and will have the behavior I noticed (see my message on our Slack channel).

ui/src/translations/en.json
```diff
@@ -84,13 +84,26 @@
...
+  "action": "Action",
+  "delete": "Delete",
+  "volume_deletion": "Volume deletion",
+  "volume_delete_success": "Volume {name} is being deleted.",
+  "volume_delete_failed": "Volume {name} deletion failed.",
+  "volume_status_pending_hint": "Volume creation is in progress.",
+  "volume_status_terminating_hint": "Volume deletion is in progress.",
+  "volume_status_unknown_hint": "The volume is in a transient state.",
+  "volume_status_failed_pv_pending_hint": "The volume is busy.",
+  "volume_statue_failed_pv_bound_hint": "The volume is bound to a storage request",
```
slaperche-scality: missing `.` at the end of the sentence.

ui/src/translations/en.json
```diff
@@ -84,13 +84,26 @@
   "no": "No",
   "ok": "OK",
   "volume_creation": "Volume creation",
-  "volume_creation_success": "Volume {name} has been created successfully.",
+  "volume_creation_success": "Volume {name} is being created.",
```
slaperche-scality, line 87: You also need to fix the French translation: `Volume {name} est en cours de création.`

slaperche-scality (changes requested):
A little bug related to the hint message + few minors remarks.

Next round will be good I think :slightly_smiling_face:

ui/src/containers/NodeVolumes.js
```diff
@@ -29,46 +41,144 @@ const ButtonContainer = styled.div`
...
+            case STATUS_AVAILABLE:
+            case STATUS_RELEASED:
+              return true;
+            case STATUS_PENDING:
+            case STATUS_BOUND:
+              return false;
+            default:
+              console.error(
+                `Unexpected state for PersistentVolume ${persistentVolume}:${persistentVolumeStatus}`
```
slaperche-scality: `persistentVolume` is the whole PV object, not only the name, right?

Anyway, since we name our PV according to the Volume, I guess you can use `volumeName` here.

slaperche-scality (changes requested):
One last small remarks and we're good to go!!

slaperche-scality (dismissed):

slaperche-scality (approved):

ChengYanJin:
@bert-e approve

PR #1523, opened 2019-08-14
WIP: Update prometheus deployment to use a pvc

### LINKED ISSUE #1515: Update Prometheus deployment to use a PVC
**Component**:

salt, monitoring, tests, documentation

**Why this is needed**:

In order to have persistent storage for the monitoring stack, we need Prometheus to us a PV for its storage.

**What should be done**:

- update the Prometheus deployment to come with a PVC (which size do we request?)
- update the E2E tests: between the bootstrap and the execution of the E2E test we need to create a PV that can satisfy the PVC (otherwise Prometheus won't start and lot of things won't work…)
- update the documentation to inform the users that they should create a PV for Prometheus after the bootstraping.

**Test plan**:

Check that Prometheus does use persistent storage (data should be preserved accross pod restart).

Ebaneck:
**Component**:

'salt', 'monitoring', 'tests', documentation

**Context**: 

See: #1515 

**Summary**:

We should be able to use the Prometheus volume to store metrics even after pod restarts or upgrade.

For this, we need to add a persistent volume claim within the Prometheus deployment file, create a  persistent volume,  in the `metalk8s-Prometheus` storageclass and bind it to the PVC.

**Acceptance criteria**: 

- Boostrapping metalk8s should still work
- CI e2e test should continue to work.
- The Prometheus deployment should include a Persistent Volume Claim automatically.
- For e2e test,  create a PV that can satisfy the PVC
- Documentation on how to create a PV that binds to the PVC should be added.
---

Closes: #1515

### DIFF
--- a/salt/metalk8s/addons/monitoring/prometheus/upstream.sls
+++ b/salt/metalk8s/addons/monitoring/prometheus/upstream.sls
@@ -1356,6 +1356,19 @@ spec:
   - key: "node-role.kubernetes.io/infra"
     operator: "Exists"
     effect: "NoSchedule"
+  storage:
+    volumeClaimTemplate:
+      apiVersion: v1
+      metadata:
+        name: prometheus-monitoring
+      kind: PersistentVolumeClaim
+      spec:
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 5Gi
+        storageClassName: metalk8s-prometheus
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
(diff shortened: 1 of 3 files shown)


Ebaneck:
To any of the reviewers, take note that this breaks the METALK8S bootstrap for the following reason.

-  Kubernetes enables automatically "Dynamic Volume Provisioning" which breaks the setup since no volume is immediately available for provisioning during the bootstrap phase.

We end up with messages like the following after a failed metalks8s bootstrap:

```Events:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  24s (x28 over 31m)  default-scheduler  pod has unbound immediate PersistentVolumeClaims
```

Let me know your views on how to go about this.

PR #1535, opened 2019-08-20
Add automated tests for the entire volume workflow

### LINKED ISSUE #1417: Add automated tests for the volume workflow
**Component**:

tests

**Why this is needed**:

To ensure that the volume creation workflow is working and keep working.

**What should be done**:

Add automated E2E tests.

Write a new feature (volume.feature) and the associated steps to be able to test the following case:
- test volume creation for each supported Volume type: kubectl apply and then check the existence of the corresponding PV.
- test volume deletion (check that finalizer is removed from the PV/check that the PV is removed (need a reasonable number of retries/timeout if relying on the GC)/check the physical volume is removed)
- test PV deletion (the volume should be deleted in the next reconciliation loop)
- test some semantically invalid cases: bad fsType, non-existing `storageClassName`, non-existing `nodeName`, missing both `rawBlockDevice` and `sparseLoopDevice`, …

**Test plan**:

- [x] Test **Volume** creation (nominal case)

```cucumber
Scenario: Test volume creation (sparseLoopDevice)
    When I create the following object:
        apiVersion: storage.metalk8s.scality.com/v1alpha1
        kind: Volume
        metadata:
          name: volume1
        spec:
          nodeName: bootstrap
          storageClassName: metalk8s-prometheus
          sparseLoopDevice:
            size: 10Gi
    Then the Volume 'volume1' is 'Available'
    And the PersistentVolume 'volume1' exists with size '1Gi'
```

- [x] Test **Volume** deletion (nominal case)
```cucumber
Scenario: Test volume deletion (sparseLoopDevice)
    Given a Volume 'volume2' exists
    When I delete the Volume 'volume2'
    Then the Volume 'volume2' does not exists
    And the PersistentVolume 'volume2' does not exists
```

- [x] Test PV-protection: a **PV** cannot be deleted while referenced by a **Volume**
```cucumber
Scenario: Test PV-protection
    Given a Volume 'volume3' exists
    When I delete the PersistentVolume 'volume3'
    Then the PersistentVolume 'volume3' is marked for deletion
    And  the Volume 'volume3' is 'Available'
    When I delete the Volume 'volume3'
    Then the Volume 'volume3' does not exists
    And the PersistentVolume 'volume3' does not exists
```

- [x] Create a **Volume** without volume type: you must specify a volume type for the creation
```cucumber
Scenario: Create a volume without volume type
    When I create the following object:
        apiVersion: storage.metalk8s.scality.com/v1alpha1
        kind: Volume
        metadata:
          name: volume4
        spec:
          nodeName: bootstrap
          storageClassName: metalk8s-prometheus
    Then the Volume 'volume4' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
```

- [x] Create a **Volume** with invalid volume type: the volume type must be supported
```cucumber
Scenario: Create a volume with invalid volume type
    When I create the following object:
        apiVersion: storage.metalk8s.scality.com/v1alpha1
        kind: Volume
        metadata:
          name: volume5
        spec:
          nodeName: bootstrap
          storageClassName: metalk8s-prometheus
          someRandomDevice:
            capacity: 10Gi
    Then the Volume 'volume5' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
```

- [x] Create a **Volume** with an unsupported FS type (we only support ext4 and xfs for now)
```cucumber
Scenario: Create a volume with unsupported FS type
    Given the existence of the StorageClass:
      apiVersion: storage.k8s.io/v1
      kind: StorageClass
      metadata:
        name: btrfs-volume
      provisioner: kubernetes.io/no-provisioner
      parameters:
        fsType: btrfs
        mkfsOptions: '["-d", "raid0"]'
      reclaimPolicy: Retain
      mountOptions: []
      volumeBindingMode: WaitForFirstConsumer
    When I create the following object:
        apiVersion: storage.metalk8s.scality.com/v1alpha1
        kind: Volume
        metadata:
          name: volume6
        spec:
          nodeName: bootstrap
          storageClassName: btrfs-volume
          sparseLoopDevice:
            size: 10Gi
    Then the Volume 'volume6' is 'Failed' with code 'CreationError' and message matches '??TBD??'
```

- [x] Create a **Volume** referencing a non-existing **StorageClass**: the **StorageClass** must exists at creation time
```cucumber
Scenario: Create a volume using a non-existing StorageClass
    Given the StorageClass 'not-found' does not exists
    When I create the following object:
        apiVersion: storage.metalk8s.scality.com/v1alpha1
        kind: Volume
        metadata:
          name: volume7
        spec:
          nodeName: bootstrap
          storageClassName: not-found
          sparseLoopDevice:
            size: 10Gi
    Then the Volume 'volume7' is 'Failed' with code 'CreationError' and message matches '??TBD??'
```

- [x] Test in-use Volume protection: you cannot delete a bound volume
```cucumber
Scenario: Test in-use Volume protection
    Given a Volume 'volume8' exists
    And I create the following object:
        kind: PersistentVolumeClaim
        apiVersion: v1
        metadata:
          name: my-pvc
        spec:
          storageClassName: metalk8s-prometheus
          accessModes:
              - ReadWriteOnce
          resources:
              requests:
                storage: 10Gi
    When I delete the Volume 'volume8'
    Then the Volume 'volume3' is marked for deletion
    And  the PersistentVolume 'volume3' is marked for deletion
    And  the Volume 'volume3' is 'Available'
    When I delete the PersistentVolumeClaim 'my-pvc'
    Then the Volume 'volume3' does not exists
    And the PersistentVolume 'volume3' does not exists
```

Ebaneck:
**Component**:

'test', 'storage', 'kubernetes'

**Context**: 

See issue #1417 
- The ticket contains an extensive description of some of the test scenarios we would like to have.

**Summary**:

- We would love to automate CI tests for volume creation, volume deletion, persistent volume creation and deletion including test cases where semantic errors might occur like a bad storage class naming, invalid filesystem types which we do not support and much more.

How to test?
`tox -e tests -- -m 'volume'
`

**Acceptance criteria**: 

- CI tests should continue to work normally
- All volume workflow scenarios should be accounted for in the test cases described. 

---

Closes: #1417

### DIFF
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -64,3 +64,16 @@ Feature: Volume management
               someRandomDevice:
                 capacity: 10Gi
         Then the Volume 'volume5' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
+
+    Scenario: Test in-use protection
+        Given a Volume 'volume6' exist
+        And a PersistentVolumeClaim exists for 'volume6'
+        And a Pod using volume 'volume6' and running '["sleep", "60"]' exist
+        When I delete the Volume 'volume6'
+        Then the Volume 'volume6' is 'Available'
+        And the Volume 'volume6' is marked for deletion
+        And the PersistentVolume 'volume6' is marked for deletion
+        When I delete the Pod using 'volume6'
+        And I delete the PersistentVolumeClaim on 'volume6'
+        Then the Volume 'volume6' does not exist
+        And the PersistentVolume 'volume6' does not exist
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -77,3 +77,11 @@ Feature: Volume management
         And I delete the PersistentVolumeClaim on 'volume6'
         Then the Volume 'volume6' does not exist
         And the PersistentVolume 'volume6' does not exist
+
+    Scenario: Volume usage (data persistency)
+        Given a Volume 'volume7' exist
+        And a PersistentVolumeClaim exists for 'volume7'
+        And a Pod using volume 'volume7' and running '["sh", "-c", "echo 'foo' > /mnt/bar.txt"]' exist
+        When I delete the Pod using 'volume7'
+        And I create a Pod using volume 'volume7' and running '["sleep", "60"]'
+        Then the Pod using volume 'volume7' has a file '/mnt/bar.txt' containing 'foo'
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -50,3 +50,17 @@ Feature: Volume management
               nodeName: bootstrap
               storageClassName: metalk8s-prometheus
         Then the Volume 'volume4' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
+
+    Scenario: Create a volume with an invalid volume type
+        Given the Kubernetes API is available
+        When I create the following Volume:
+            apiVersion: storage.metalk8s.scality.com/v1alpha1
+            kind: Volume
+            metadata:
+              name: volume5
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s-prometheus
+              someRandomDevice:
+                capacity: 10Gi
+        Then the Volume 'volume5' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -38,3 +38,15 @@ Feature: Volume management
         When I delete the Volume 'volume3'
         Then the Volume 'volume3' does not exist
         And the PersistentVolume 'volume3' does not exist
+
+    Scenario: Create a volume with no volume type
+        Given the Kubernetes API is available
+        When I create the following Volume:
+            apiVersion: storage.metalk8s.scality.com/v1alpha1
+            kind: Volume
+            metadata:
+              name: volume4
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s-prometheus
+        Then the Volume 'volume4' is 'Failed' with code 'InternalError' and message matches 'volume type not found'
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -8,3 +8,18 @@ Feature: Volume management
     Scenario: The storage operator is up
         Given the Kubernetes API is available
         Then we have 1 running pod labeled 'name=storage-operator' in namespace 'kube-system'
+
+    Scenario: Test volume creation (sparseLoopDevice)
+        Given the Kubernetes API is available
+        When I create the following Volume:
+            apiVersion: storage.metalk8s.scality.com/v1alpha1
+            kind: Volume
+            metadata:
+              name: volume1
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s-prometheus
+              sparseLoopDevice:
+                size: 10Gi
+        Then the Volume 'volume1' is 'Available'
+        And the PersistentVolume 'volume1' has size '10Gi'
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -29,3 +29,12 @@ Feature: Volume management
         When I delete the Volume 'volume2'
         Then the Volume 'volume2' does not exist
         And the PersistentVolume 'volume2' does not exist
+
+    Scenario: Test PersistentVolume protection
+        Given a Volume 'volume3' exist
+        When I delete the PersistentVolume 'volume3'
+        Then the PersistentVolume 'volume3' is marked for deletion
+        And the Volume 'volume3' is 'Available'
+        When I delete the Volume 'volume3'
+        Then the Volume 'volume3' does not exist
+        And the PersistentVolume 'volume3' does not exist
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -23,3 +23,9 @@ Feature: Volume management
                 size: 10Gi
         Then the Volume 'volume1' is 'Available'
         And the PersistentVolume 'volume1' has size '10Gi'
+
+    Scenario: Test volume deletion (sparseLoopDevice)
+        Given a Volume 'volume2' exist
+        When I delete the Volume 'volume2'
+        Then the Volume 'volume2' does not exist
+        And the PersistentVolume 'volume2' does not exist
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -53,6 +53,12 @@ def test_volume_creation(host):
 def test_volume_deletion(host):
     pass
 
+@scenario('../features/volume.feature',
+          'Test PersistentVolume protection',
+          strict_gherkin=False)
+def test_pv_protection(host):
+    pass
+
 # }}}
 # Given {{{
 
@@ -82,6 +88,16 @@ def delete_volume(host, name, k8s_custom_client):
         body=kubernetes.client.V1DeleteOptions(),
         grace_period_seconds=0
     )
+
+
+@when(parsers.parse("I delete the PersistentVolume '{name}'"))
+def delete_pv(host, name, k8s_client):
+    k8s_client.delete_persistent_volume(
+        name=name,
+        body=kubernetes.client.V1DeleteOptions(),
+        grace_period_seconds=0
+    )
+
 
 # }}}
 # Then {{{
@@ -152,21 +168,33 @@ def check_volume_absent(host, name, k8s_custom_client):
 @then(parsers.parse("the PersistentVolume '{name}' does not exist"))
 def check_pv_absent(name, k8s_client):
     def _check_pv():
-        try:
-            k8s_client.read_persistent_volume(name)
-        except (ApiException, HTTPError) as exc:
-            if isinstance(exc, ApiException) and exc.status == 404:
-                return
-            raise
-        assert False, 'PersistentVolume {} exist'.format(name)
+        assert _get_pv(k8s_client, name) is None,\
+            'PersistentVolume {} exist'.format(name)
 
     utils.retry(
         _check_pv, times=10, wait=2,
         name='checking the absence of PersistentVolume {}'.format(name)
     )
 
+
+@then(parsers.parse("the PersistentVolume '{name}' is marked for deletion"))
+def check_pv_deletion_marker(name, k8s_client):
+    def _check_pv_deletion_marker():
+        pv = _get_pv(k8s_client, name)
+        assert pv is not None, 'PersistentVolume {} not found'.format(name)
+        assert pv.metadata.deletion_timestamp is not None,\
+            'PersistentVolume {} is not marked for deletion'.format(name)
+
+    utils.retry(
+        _check_pv_deletion_marker, times=10, wait=2,
+        name='checking that PersistentVolume {} is marked for deletion'.format(
+            name
+        )
+    )
+
 # }}}
 # Helpers {{{
+# Volume {{{
 
 def _create_volume(k8s_client, body):
     k8s_client.create_cluster_custom_object(
@@ -190,4 +218,16 @@ def _get_volume(k8s_client, name):
             return None
         raise
 
+# }}}
+# PersistentVolume {{{
+
+def _get_pv(k8s_client, name):
+    try:
+        return k8s_client.read_persistent_volume(name)
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        raise
+
+# }}}
 # }}}
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -78,8 +78,8 @@ spec:
 # Fixture {{{
 
 @pytest.fixture
-def volume_client(k8s_apiclient):
-    return VolumeClient(CustomObjectsApi(api_client=k8s_apiclient))
+def volume_client(k8s_apiclient, ssh_config):
+    return VolumeClient(CustomObjectsApi(api_client=k8s_apiclient), ssh_config)
 
 @pytest.fixture
 def pv_client(k8s_client):
@@ -427,10 +427,11 @@ class Client(abc.ABC):
 # VolumeClient {{{
 
 class VolumeClient(Client):
-    def __init__(self, k8s_client):
+    def __init__(self, k8s_client, ssh_config):
         super().__init__(
             k8s_client, kind='Volume', retry_count=30, retry_delay=2
         )
+        self._ssh_config = ssh_config
         self._group="storage.metalk8s.scality.com"
         self._version="v1alpha1"
         self._plural="volumes"
@@ -441,6 +442,10 @@ class VolumeClient(Client):
         )['items']
 
     def _create(self, body):
+        # Fixup the hostname.
+        body['spec']['nodeName'] = utils.resolve_hostname(
+            body['spec']['nodeName'], self._ssh_config
+        )
         self._client.create_cluster_custom_object(
             group=self._group, version=self._version, plural=self._plural,
             body=body
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -473,7 +473,7 @@ def _create_pod(k8s_client, volume_name, image_name, full_command):
     )
 
     utils.retry(
-        kube_utils.wait_for_pod(k8s_client, pod_name),
+        kube_utils.check_pod_status(k8s_client, pod_name),
         times=30, wait=2,
         name="wait for pod {}".format(pod_name)
     )
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -1,3 +1,4 @@
+import re
 from urllib3.exceptions import HTTPError
 
 import kubernetes.client
@@ -55,6 +56,10 @@ def test_volume_deletion(host):
 
 @scenario('../features/volume.feature', 'Test PersistentVolume protection')
 def test_pv_protection(host):
+    pass
+
+@scenario('../features/volume.feature', 'Create a volume with no volume type')
+def test_no_volume_type(host):
     pass
 
 # }}}
@@ -188,6 +193,33 @@ def check_pv_deletion_marker(name, k8s_client):
         name='checking that PersistentVolume {} is marked for deletion'.format(
             name
         )
+    )
+
+
+@then(parsers.parse("the Volume '{name}' is 'Failed' "
+                    "with code '{code}' and message matches '{pattern}'"))
+def check_volume_error(host, name, code, pattern, k8s_custom_client):
+    def _check_error():
+        volume = _get_volume(k8s_custom_client, name)
+        assert volume is not None, 'Volume {} not found'.format(name)
+        status = volume.get('status')
+        assert status is not None, 'no status for volume {}'.format(name)
+        assert status['phase'] == 'Failed',\
+            'Unexpected status: expected Failed, got {}'.format(
+                status, status['phase']
+            )
+        assert status['errorCode'] == code,\
+            'Unexpected error code: expected {}, got {}'.format(
+                code, status['errorCode']
+            )
+        assert re.search(pattern, status['errorMessage']) is not None,\
+            "error message `{}` doesn't match `{}`".format(
+                status['errorMessage'], pattern
+            )
+
+    utils.retry(
+        _check_error, times=30, wait=2,
+        name='checking error for Volume {}'.format(name)
     )
 
 # }}}
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -62,6 +62,11 @@ def test_pv_protection(host):
 def test_no_volume_type(host):
     pass
 
+@scenario('../features/volume.feature',
+          'Create a volume with an invalid volume type')
+def test_invalid_volume_type(host):
+    pass
+
 # }}}
 # Given {{{
 
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -1,6 +1,9 @@
+import ast
+import json
 import re
 from urllib3.exceptions import HTTPError
 
+import kubernetes as k8s
 import kubernetes.client
 from kubernetes.client import CustomObjectsApi
 from kubernetes.client import StorageV1Api
@@ -10,6 +13,7 @@ from pytest_bdd import given, when, scenario, then, parsers
 import yaml
 
 from tests import utils
+from tests import kube_utils
 
 
 # Constants {{{
@@ -53,7 +57,8 @@ spec:
   containers:
     - name: {volume_name}-pod-container
       image: {image_name}
-      command: {command}
+      command: [{command}]
+      args: {args}
       volumeMounts:
         - mountPath: "/mnt/"
           name: {volume_name}-pod-storage
@@ -127,6 +132,10 @@ def test_invalid_volume_type(host, cleanup_volumes):
 def test_in_use_protection(host, cleanup_volumes):
     pass
 
+@scenario('../features/volume.feature', 'Volume usage (data persistency)')
+def test_volume_data_persistency(host, cleanup_volumes):
+    pass
+
 # }}}
 # Given {{{
 
@@ -159,21 +168,8 @@ def create_pvc_for_volume(host, volume_name, k8s_client):
 @given(parsers.parse(
     "a Pod using volume '{volume_name}' and running '{command}' exist"
 ))
-def create_pod_for_volume(host, volume_name, command, k8s_client, utils_image):
-    pod_name = '{}-pod'.format(volume_name)
-    if _get_pod(k8s_client, pod_name) is not None:
-        return
-    body = POD_TEMPLATE.format(
-        volume_name=volume_name, image_name=utils_image, command=command,
-    )
-    k8s_client.create_namespaced_pod(
-        namespace='default', body=yaml.safe_load(body)
-    )
-    utils.retry(
-        kube_utils.wait_for_pod(k8s_client, pod_name),
-        times=30, wait=2,
-        name="wait for pod {}".format(pod_name)
-    )
+def pod_exists_for_volume(host, volume_name, command, k8s_client, utils_image):
+    _create_pod(k8s_client, volume_name, utils_image, command)
 
 # }}}
 # When {{{
@@ -229,6 +225,13 @@ def delete_pv_claim(host, volume_name, k8s_client):
         _check_pv_claim_absent, times=10, wait=2,
         name='checking for the absence of PersistentVolumeClaim {}'.format(name)
     )
+
+
+@when(parsers.parse(
+    "I create a Pod using volume '{volume_name}' and running '{command}'"
+))
+def create_pod_for_volume(host, volume_name, command, k8s_client, utils_image):
+    _create_pod(k8s_client, volume_name, utils_image, command)
 
 # }}}
 # Then {{{
@@ -357,6 +360,32 @@ def check_volume_deletion_marker(name, k8s_custom_client):
         name='checking that Volume {} is marked for deletion'.format(name)
     )
 
+
+@then(parsers.parse("the Pod using volume '{volume_name}' "
+                    "has a file '{path}' containing '{content}'"))
+def check_file_content_inside_pod(volume_name, path, content, k8s_client):
+    name = '{}-pod'.format(volume_name)
+
+    def _check_file_content():
+        try:
+            result = k8s.stream.stream(
+                k8s_client.connect_get_namespaced_pod_exec,
+                name=name, namespace='default',
+                command=['cat', path],
+                stderr=True, stdin=False, stdout=True, tty=False
+               )
+        except ApiException:
+            assert False
+        assert result.rstrip('\n') == content,\
+            'unexpected data in {}: expected "{}", got "{}"'.format(
+                path, content, result
+            )
+
+    utils.retry(
+        _check_file_content, times=10, wait=2,
+        name='checking content of {} on Pod {}'.format(path, name)
+    )
+
 # }}}
 # Helpers {{{
 # Volume {{{
@@ -431,6 +460,25 @@ def _get_pv_claim(k8s_client, name, namespace='default'):
 
 # }}}
 # Pod {{{
+
+def _create_pod(k8s_client, volume_name, image_name, full_command):
+    pod_name = '{}-pod'.format(volume_name)
+    if _get_pod(k8s_client, pod_name) is not None:
+        return
+    command, *args = ast.literal_eval(full_command)
+    body = POD_TEMPLATE.format(
+        volume_name=volume_name, image_name=image_name,
+        command=json.dumps(command), args=json.dumps(args)
+    )
+    k8s_client.create_namespaced_pod(
+        namespace='default', body=yaml.safe_load(body)
+    )
+
+    utils.retry(
+        kube_utils.wait_for_pod(k8s_client, pod_name),
+        times=30, wait=2,
+        name="wait for pod {}".format(pod_name)
+    )
 
 def _get_pod(k8s_client, name, namespace='default'):
     try:
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -1,3 +1,4 @@
+import abc
 import ast
 import json
 import re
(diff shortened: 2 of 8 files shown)


Ebaneck:
@scality/metalk8s and any other person reviewing this,  feel free to leave a comment stating if this is the right path to follow and/or some test scenarios we have not accounted for as specified in the original ticket.

tests/post/features/volume.feature
```diff
@@ -11,15 +11,6 @@ Feature: Volume management
...
-            apiVersion: storage.metalk8s.scality.com/v1alpha1
-            kind: Volume
-            metadata:
-              name: volume1
-            spec:
-              nodeName: bootstrap
-              storageClassName: metalk8s-prometheus
-              sparseLoopDevice:
-                size: 10Gi
+        When I create a Volume 'volume1'
```
slaperche-scality: No, when you create a volume you want to specify the body.

This should be left untouched

tests/post/steps/test_volume.py
```diff
@@ -105,4 +100,38 @@ def _check_size():
...
+  sparseLoopDevice:
+    size: 10Gi
+"""
+
+# }}}
+
+# Helpers {{{
+
+
+def _create_volume(host, name, k8s_custom_client):
```
slaperche-scality: not exactly, here we want an helper to create a volume from a body.

```python
def _create_volume(k8s_client, body):
    try:
        k8s_client.create_cluster_custom_object(
            group="storage.metalk8s.scality.com",
            version="v1alpha1",
            plural="volumes",
            body=body)
        )
        if isinstance(exc, ApiException) and exc.status == 404:
            assert False, 'Exception while creating volume {}'.format(name)
        raise
```

tests/post/steps/test_volume.py
```diff
@@ -105,4 +100,38 @@ def _check_size():
...
     )
 
-# }}}
\ No newline at end of file
+# }}}
+
+# Constants {{{
+
+
+DEFAULT_VOLUME = """
```
slaperche-scality: This should be introduced in the next commit, because you don't need it here; this commit only introduce the helper.

Also, constants are usually defined at the top of the fileso I would move the `Constants` fold before the `Fixtures` one.

tests/post/steps/test_volume.py
```diff
@@ -34,14 +34,9 @@ def test_volume_creation(host):
...
-def create_volume(host, body, k8s_custom_client):
-    k8s_custom_client.create_cluster_custom_object(
-        group="storage.metalk8s.scality.com",
-        version="v1alpha1",
-        plural="volumes",
-        body=yaml.safe_load(body)
-    )
+@when(parsers.parse("I create a Volume '{name}'"))
+def create_volume(host, name, k8s_custom_client):
+    _create_volume(host, name, k8s_custom_client)
```
slaperche-scality: This should become `_create_volume(k8s_custom_client, body)`

tests/post/features/volume.feature
slaperche-scality: You should explain why we suddenly introduce an helper.

Something like:

```
test/volume: refactor volume creation test to use helpers 
In future test we will need to create Volume by giving only a name
(we will use a default template for the other fields) as well as
creating Volume from a YAML manifest (what we are already doing).

Since the creation code will be the same, let's factor in into an helper.
```

tests/post/steps/test_volume.py
```diff
@@ -100,6 +110,31 @@ def _check_size():
         name='checking size of PersistentVolume {}'.format(name)
     )
 
+
+@then(parsers.parse("the Volume '{name}' does not exists"))
+def volume_not_exist(host, name, k8s_custom_client, k8s_client):
```
slaperche-scality: `check_volume_absent`, `then` statement usually have function name starting with `check`.

tests/post/steps/test_volume.py
```diff
@@ -134,4 +169,38 @@ def _create_volume(host, name, k8s_custom_client):
         if isinstance(exc, ApiException) and exc.status == 404:
             assert False, 'Exception while creating volume {}'.format(name)
         raise
+
+
+def _volume_exist(host, name, k8s_custom_client):
```
slaperche-scality: This return an object, not a Boolean: call it `_get_volume`.

`host` is unused: can be removed.
Here we're not using the fixture (it's not a `then`/`when`/`given`, so fixture are not magically applied): we can name the parameter `k8s_client`).


Also we know that we're fetching a Volume here, so the variable can be named `volume` instead of `custom_object`

```python
 def _get_volume(k8s_client, name):
    try:
        return k8s_custom_client.get_cluster_custom_object(
            group="storage.metalk8s.scality.com",
            version="v1alpha1",
            plural="volumes",
            name=name
        )
        if isinstance(exc, ApiException) and exc.status == 404:
            return None
        raise
```

tests/post/steps/test_volume.py
```diff
@@ -31,13 +32,32 @@ def test_deploy_operator(host):
...
 # When {{{
 
 @when(parsers.parse("I create a Volume '{name}'"))
 def create_volume(host, name, k8s_custom_client):
     _create_volume(host, name, k8s_custom_client)
 
+
+@when(parsers.parse("I delete the Volume '{name}'"))
+def delete_volume(host, name, k8s_custom_client):
+    _delete_volume(host, name, k8s_custom_client)
```
slaperche-scality: `_delete_volume` is only used here, so I don't think you need an helper.

Simply write:

```python
@when(parsers.parse("I delete the Volume '{name}'"))
def delete_volume(host, name, k8s_custom_client):
        k8s_custom_client.delete_cluster_custom_object(
            group="storage.metalk8s.scality.com",
            version="v1alpha1",
            plural="volumes",
            name=name,
            body=kubernetes.client.V1DeleteOptions(),
            grace_period_seconds=0
        )
```

We don't need to handle exception here, if something fail the test should fail.

tests/post/steps/test_volume.py
```diff
@@ -31,13 +32,32 @@ def test_deploy_operator(host):
...
+
+# }}}
+
+# Given {{{
+
+@given(parsers.parse("a Volume '{name}' exists"))
+def volume_exist(host, name, k8s_custom_client):
+    volume = _volume_exist(host, name, k8s_custom_client)
+    if volume is None:
+        _create_volume(host, name, k8s_custom_client)
```
slaperche-scality: Here you should have:

```python
if volume is None:
    body = DEFAULT_VOLUME.format(name=name)
    _create_volume(k8s_custom_client, body)
```

tests/post/steps/test_volume.py
```diff
@@ -100,6 +110,31 @@ def _check_size():
...
+
+@then(parsers.parse("the Volume '{name}' does not exists"))
+def volume_not_exist(host, name, k8s_custom_client, k8s_client):
+    def _check_volume():
+        volume = _volume_exist(host, name, k8s_custom_client)
+        if volume is not None:
+            assert False, 'Volume {} still exist'.format(name)
+    utils.retry(
+        _check_volume, times=30, wait=2,
+        name='checking if volume {} exist'.format(name)
```
slaperche-scality: `checking the absence of volume {}`

tests/post/steps/test_volume.py
```diff
@@ -100,6 +110,31 @@ def _check_size():
...
+    )
+
+
+@then(parsers.parse("the PersistentVolume '{name}' does not exists"))
+def check_pv(host, name, k8s_client):
```
slaperche-scality: Hum, this should be similar to `check_volume_absent`, but for the PV instead (so rename it, use a `retry`, …)

slaperche-scality (changes requested):

tests/post/steps/test_volume.py
Ebaneck: yes we do not need the helper yet , but the next test case makes use of a delete function definition hence the helper

tests/post/steps/test_volume.py
Ebaneck: > This return an object, not a Boolean: call it `_get_volume`.
> 
> `host` is unused: can be removed.
> Here we're not using the fixture (it's not a `then`/`when`/`given`, so fixture are not magically applied): we can name the parameter `k8s_client`).

Not sure this works as stated, tested and it only works with the right fixtures (`k8s_custom_client`). Safe to say that for some reason, fixtures are magically applied here.

> 
> Also we know that we're fetching a Volume here, so the variable can be named `volume` instead of `custom_object`
> 
> ```python
>  def _get_volume(k8s_client, name):
>     try:
>         return k8s_custom_client.get_cluster_custom_object(
>             group="storage.metalk8s.scality.com",
>             version="v1alpha1",
>             plural="volumes",
>             name=name
>         )
>     except (ApiException, HTTPError) as exc:
>         if isinstance(exc, ApiException) and exc.status == 404:
>             return None
>         raise
> ```

tests/post/steps/test_volume.py
slaperche-scality: We also probably need to have a wait here, until the Volume becomes `Available`.

tests/post/steps/test_volume.py
slaperche-scality: > the next test case makes use of a delete function definition hence the helper

Then introduce it in the next test case.
I'm not even sure the next test case use it actually.

The next test case use ` When I delete the Volume 'volume3'` which is exactly what we use here. No code to add for that.

tests/post/features/volume.feature
slaperche-scality: This is not resolved: you didn't restore the step as it was…

tests/post/steps/test_volume.py
slaperche-scality: I applied this diff (which is what I was recommending) on your current PR:

```diff
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -160,9 +160,9 @@ def check_pv_absent(name, k8s_client):
 # Helpers {{{
 
 
-def _create_volume(body, k8s_custom_client):
+def _create_volume(body, k8s_client):
     try:
-        k8s_custom_client.create_cluster_custom_object(
+        k8s_client.create_cluster_custom_object(
             group="storage.metalk8s.scality.com",
             version="v1alpha1",
             plural="volumes",
@@ -174,9 +174,9 @@ def _create_volume(body, k8s_custom_client):
         raise
 
 
-def _get_volume(name, k8s_custom_client):
+def _get_volume(name, k8s_client):
     try:
-        return k8s_custom_client.get_cluster_custom_object(
+        return k8s_client.get_cluster_custom_object(
             group="storage.metalk8s.scality.com",
             version="v1alpha1",
             plural="volumes",
@@ -187,9 +187,8 @@ def _get_volume(name, k8s_custom_client):
             return None
 
 
-def _delete_volume(name, k8s_custom_client):
-
-    k8s_custom_client.delete_cluster_custom_object(
+def _delete_volume(name, k8s_client):
+    k8s_client.delete_cluster_custom_object(
         group="storage.metalk8s.scality.com",
         version="v1alpha1",
         plural="volumes",
```

And it works…

So no, fixtures are **not** magically applied :slightly_smiling_face:

tests/post/steps/test_volume.py
Ebaneck: I fail to see why we need to wait here. We are making a request to the API and once the request is submitted, we should terminate the transaction.

Checking the Availability of the created volume should be handled in the next stage which is what we do actually.

slaperche-scality: Nope, the role of the `Given` is to ensure that the volume exists and thus should wait for it to becomes available.

The test scenario is:

```cucumber
Scenario: Test volume deletion (sparseLoopDevice)
    Given a Volume 'volume2' exists
    When I delete the Volume 'volume2'
    Then the Volume 'volume2' does not exists
    And the PersistentVolume 'volume2' does not exists
```

We try to delete right after we are sure that the Volume exists: there is no explicit `wait` to add in the scenario because it must be handled by the given.

tests/post/features/volume.feature
```diff
@@ -8,3 +8,59 @@ Feature: Volume management
...
+              storageClassName: metalk8s-prometheus
+              sparseLoopDevice:
+                size: 10Gi
+        Then the Volume 'volume1' is 'Available'
+        And the PersistentVolume 'volume1' has size '10Gi'
+
+    Scenario: Test volume deletion (sparseLoopDevice)
+        Given a Volume 'volume2' exists
+        When I delete the Volume 'volume2'
+        Then the Volume 'volume2' does not exists
```
charlesprost:
```suggestion
        Then the Volume 'volume2' does not exist
```

tests/post/features/volume.feature
```diff
@@ -8,3 +8,59 @@ Feature: Volume management
...
+        And the PersistentVolume 'volume2' does not exists
```
charlesprost:
```suggestion
        And the PersistentVolume 'volume2' does not exist
```

charlesprost: Same below

tests/post/steps/test_volume.py
```diff
@@ -1,10 +1,39 @@
...
-from pytest_bdd import scenario, then, parsers
+from pytest_bdd import given, when, scenario, then, parsers
+import yaml
 
+from tests import utils
+
+
+# Constants {{{
+
+DEFAULT_VOLUME = """
```
charlesprost, line 22: If there is a default volume we might want to avoid passing it in the scenarios themselves.

tests/post/steps/test_volume.py
```diff
@@ -15,7 +44,86 @@ def test_deploy_storage_class(host):
...
+def test_invalid_volume_type(host):
+    pass
+
+# }}}
+# Given {{{
+
+@given(parsers.parse("a Volume '{name}' exists"))
+def volume_exist(host, name, k8s_custom_client):
+    volume = _get_volume(k8s_custom_client, name)
+    if volume is None:
```
charlesprost: Python tip: when 90% of the function is indented after a `if`, it could be better to do something like:

```
if volume is not None:
    return
#blablabla
```

tests/post/steps/test_volume.py
```diff
@@ -26,3 +134,152 @@ def check_storage_class(host, name, k8s_apiclient):
         if isinstance(exc, ApiException) and exc.status == 404:
             assert False, 'StorageClass {} not found'.format(name)
         raise
+
+
+@then(parsers.parse("the Volume '{name}' is '{status}'"))
+def check_volume_status(host, name, status, k8s_custom_client):
+    def _check_status():
```
charlesprost: Define only once `_check_status`

tests/post/steps/test_volume.py
```diff
@@ -26,3 +134,152 @@ def check_storage_class(host, name, k8s_apiclient):
...
+                size, pv.spec.capacity['storage']
+            )
+
+    utils.retry(
+        _check_size, times=30, wait=2,
+        name='checking size of PersistentVolume {}'.format(name)
+    )
+
+
+@then(parsers.parse("the Volume '{name}' does not exists"))
```
charlesprost:
```suggestion
@then(parsers.parse("the Volume '{name}' does not exist"))
```

tests/post/steps/test_volume.py
slaperche-scality: Nope, two different use cases here.

You want to create a volume, and that's the point of the test?
=> then pass a YAML

Using a templated stuff for that will just give unreadable long line like:
`And I create a volume titled 'test-volume' of type 'sparseLoopDevice' with storageclass 'random-storageclass' of size '1Gi' for the 'bootstrap' node with a random device type 'rawSparseDevice'`
So, please no…

On the other hand, when you just need a volume to exists (because you want delete it or whatever) then we can use an hardcoded default one.

tests/post/steps/test_volume.py
Ebaneck: The scope of the function is local so, I think this should not be a problem.
 By the way, the check status function is specific in each case it has been used. Maybe we should really give this nested functions proper names

slaperche-scality: >  Maybe we should really give this nested functions proper names

We should (and we started to do so with `_check_status` and `_check_size`), it's easier for grepping & jumping in the code.

tests/post/steps/test_volume.py
```diff
@@ -26,3 +56,54 @@ def check_storage_class(host, name, k8s_apiclient):
...
+            if isinstance(exc, ApiException) and exc.status == 404:
+                assert False, 'PersistentVolume {} not found'.format(name)
+            raise
+        assert pv.spec.capacity['storage'] == size, \
+            'Unexpected PersistentVolume size: expected {}, got {}'.format(
+                size, pv.spec.capacity['storage']
+            )
+
+    utils.retry(
+        _check_size, times=30, wait=2,
```
slaperche-scality: Seeing that PV doesn't trigger long action at creation time, I think we may reduce the retry parameters.

Maybe `times=10, wait=2` is enough.

tests/post/steps/test_volume.py
```diff
@@ -12,7 +13,6 @@
 
 # Constants {{{
 
-
```
slaperche-scality: Why this change?
If you really want to to it, should be done in the previous commit.

tests/post/steps/test_volume.py
```diff
@@ -48,13 +48,57 @@ def test_deploy_operator(host):
...
+
+@given(parsers.parse("a Volume '{name}' exists"))
+def volume_exist(host, name, k8s_custom_client):
+    volume = _get_volume(k8s_custom_client, name)
+    if volume is not None:
+        return
+    body = DEFAULT_VOLUME.format(name=name)
+    _create_volume(k8s_custom_client, body)
+
+    def _check_status(status='Available'):
```
slaperche-scality: I think it could be simplified to:

```python
    def _check_availability():
        volume = _get_volume(k8s_custom_client, name)
        assert volume['status']['phase'] == 'Available',\
            'Volume not available, status={}'.format(volume['status']['phase'])
```

tests/post/steps/test_volume.py
```diff
@@ -48,13 +48,57 @@ def test_deploy_operator(host):
...
+                'Unexpected status: expected {}, got {}'.format(
+                    status, volume['status']['phase']
+                )
+        except KeyError:
+            assert status == 'Unknown', \
+                'Unexpected status: expected {}, got none'.format(status)
+
+    utils.retry(
+        _check_status, times=30, wait=2,
+        name='checking status of Volume {}'.format(name)
```
slaperche-scality: `checking availability of Volume {}`

tests/post/steps/test_volume.py
```diff
@@ -117,6 +151,34 @@ def _check_size():
...
+    )
+
+
+@then(parsers.parse("the PersistentVolume '{name}' does not exists"))
+def check_pv_absent(name, k8s_client):
+    def _check_pv():
+        try:
+            pv = k8s_client.read_persistent_volume(name)
+        except (ApiException, HTTPError) as exc:
+            if isinstance(exc, ApiException) and exc.status == 404:
```
slaperche-scality:
```python
if isinstance(exc, ApiException) and exc.status == 404:
    return None
raise
```

tests/post/steps/test_volume.py
```diff
@@ -117,6 +151,34 @@ def _check_size():
...
+                return None
+        if pv is not None:
+            assert False, 'PersistentVolume {} exist'.format(name)
+    utils.retry(
+        _check_pv, times=30, wait=2,
```
slaperche-scality: As before, I think we can decrease the retry parameters: `times=10, wait=1` or `times=10, wait=2`.

tests/post/steps/test_volume.py
```diff
@@ -117,6 +151,34 @@ def _check_size():
         name='checking size of PersistentVolume {}'.format(name)
     )
 
+
+@then(parsers.parse("the Volume '{name}' does not exists"))
+def check_volume_absent(host, name, k8s_custom_client, k8s_client):
+    def _check_volume():
+        volume = _get_volume(k8s_custom_client, name)
```
slaperche-scality: Can be simplified to:

```python
    def _check_volume_absent():
        assert _get_volume(k8s_custom_client, name) is None,\
            'Volume {} still exist'.format(name)
```

tests/post/steps/test_volume.py
```diff
@@ -130,7 +192,19 @@ def _create_volume(k8s_client, body):
...
 
+def _get_volume(k8s_client, name):
+    try:
+        return k8s_client.get_cluster_custom_object(
+            group="storage.metalk8s.scality.com",
+            version="v1alpha1",
+            plural="volumes",
+            name=name
+        )
+    except (ApiException, HTTPError) as exc:
```
slaperche-scality: You should re-raise if it's not a 404:

```python
        if isinstance(exc, ApiException) and exc.status == 404:
            return None
        raise
```

tests/post/steps/test_volume.py
```diff
@@ -107,3 +118,19 @@ def _check_size():
...
+    try:
+        k8s_client.create_cluster_custom_object(
+            group="storage.metalk8s.scality.com",
+            version="v1alpha1",
+            plural="volumes",
+            body=yaml.safe_load(body)
+        )
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            assert False, 'Exception while creating volume {}'.format(name)
```
slaperche-scality: While only error when 404? Other errors are also errors.

Should be defined as:

```python
def _create_volume(k8s_client, body):
    k8s_client.create_cluster_custom_object(
        group="storage.metalk8s.scality.com",
        version="v1alpha1",
        plural="volumes",
        body=yaml.safe_load(body)
    )
```

tests/post/steps/test_volume.py
```diff
@@ -48,13 +48,57 @@ def test_deploy_operator(host):
...
+@scenario('../features/volume.feature',
+          'Test volume deletion (sparseLoopDevice)')
+def test_volume_deletion(host):
+    pass
+
+# }}}
+# Given {{{
+
+@given(parsers.parse("a Volume '{name}' exists"))
+def volume_exist(host, name, k8s_custom_client):
```
slaperche-scality: nested function are usually defined at the top of the outer function, not in the middle:

```python
def volume_exist(host, name, k8s_custom_client):
    def _check_availability():
        volume = _get_volume(k8s_custom_client, name)
        assert volume['status']['phase'] == 'Available',\
            'Volume not available, status={}'.format(volume['status']['phase'])

    volume = _get_volume(k8s_custom_client, name)
    if volume is not None:
        return
    body = DEFAULT_VOLUME.format(name=name)
    _create_volume(k8s_custom_client, body)
    utils.retry(
        _check_availability, times=30, wait=2,
        name='checking availability of Volume {}'.format(name)
    )
```

tests/post/steps/test_volume.py
```diff
@@ -179,6 +195,31 @@ def _check_pv():
         name='checking the absence of PersistentVolume {}'.format(name)
     )
 
+
+@then(parsers.parse("the PersistentVolume '{name}' is marked for deletion"))
+def check_pv_deletion_timestamp(name, k8s_client):
+    def _check_deletion_timestamp():
+        try:
+            pv = k8s_client.read_persistent_volume(name)
```
slaperche-scality: Like you have a `_get_volume` helper, you should have a `_get_pv` helper.

```python
def _get_volume(k8s_client, name):
    try:
        return k8s_client.read_persistent_volume(name)
        if isinstance(exc, ApiException) and exc.status == 404:
            return None
        raise
```

then you could write:
```python
    def _check_deletion_timestamp():
        pv = _get_pv(name)
        assert ps is not None, 'PersistentVolume {} not found'.format(name)
        assert pv.metadata.deletion_timestamp is not None,\
            'PersistentVolume {} is not marked for deletion'.format(name)
```

tests/post/steps/test_volume.py
```diff
@@ -218,6 +223,33 @@ def _check_deletion_timestamp():
         name='checking the termination of PersistentVolume {}'.format(name)
     )
 
+
+@then(parsers.parse("the Volume '{name}' is 'Failed' "
+                    "with code '{code}' and message matches '{pattern}'"))
+def check_volume_error(host, name, code, pattern, k8s_custom_client):
+    def _check_status():
```
slaperche-scality: `def _check_error():`

tests/post/features/volume.feature
```diff
@@ -115,3 +115,88 @@ Feature: Volume management
         When I delete the PersistentVolumeClaim 'my-pvc'
         Then the Volume 'volume8' does not exists
         And the PersistentVolume 'volume8' does not exists
+
+    Scenario: Complete e2e volume workflow
```
slaperche-scality: `E2E`, it's an acronym.
So: `Complete E2E volume workflow`.

tests/post/steps/test_volume.py
```diff
@@ -94,6 +99,22 @@ def _check_status(status='Available'):
         name='checking status of Volume {}'.format(name)
     )
 
+
+
+@given(parsers.parse("I create the following object:\n{body}"))
```
slaperche-scality: no, don't use some generic stuff such as "object'
It would be good if the step could create arbitrary object, but that's not the case (yet).

Also a `given` is about a state, not about action thus you can't write `I create`.

Either:
- `@given(parsers.parse('The following PersistentVolumeClaim exist:\n{body}'))`
- `@then(parsers.parse('I create the following PersistentVolumeClaim:\n{body}'))`

tests/post/steps/test_volume.py
```diff
@@ -94,6 +99,22 @@ def _check_status(status='Available'):
...
+def create_pvc(host, body, k8s_client):
+    k8s_client.create_namespaced_persistent_volume_claim(
+        namespace='monitoring',
+        body=yaml.safe_load(body)
+    )
+
+
+@given(parsers.parse("I create the following Pod:\n{body}"))
```
slaperche-scality: Same remarks, either:
- `@given(parsers.parse("the following Pod exist:\n{body}"))`
- `@then(parsers.parse("I create the following Pod:\n{body}"))`

tests/post/steps/test_volume.py
```diff
@@ -255,6 +285,24 @@ def _check_status():
...
+        if volume is None:
+            assert False, 'Volume {} not found'.format(name)
+
+        assert volume['metadata']['deletionTimestamp'] is not None,\
+            'Unexpected deletion timestamp got {}'.format(
+                volume['metadata']['deletionTimestamp']
+            )
+
+    utils.retry(
+        _check_deletion_timestamp, times=30, wait=2,
```
slaperche-scality: This should appear quite fast (we don't need to call Salt or whatever) so we can reduce the settings here: `times=10, wait=1` or `times=10, wait=2`

tests/post/steps/test_volume.py
```diff
@@ -255,6 +285,24 @@ def _check_status():
...
+
+@then(parsers.parse("the Volume '{name}' is marked for deletion"))
+def check_volume_deletion_timestamp(name, k8s_custom_client):
+    def _check_deletion_timestamp():
+        volume = _get_volume(k8s_custom_client, name)
+        if volume is None:
+            assert False, 'Volume {} not found'.format(name)
+
+        assert volume['metadata']['deletionTimestamp'] is not None,\
+            'Unexpected deletion timestamp got {}'.format(
```
slaperche-scality: How is it `Unexpected deletion timestamp` when the point of the test is that you expect one?

Can be rewritten as:

```python
    def _check_volume_deletion_timestamp():
        volume = _get_volume(k8s_custom_client, name)
        assert volume is not None, 'Volume {} not found'.format(name)
        assert volume['metadata']['deletionTimestamp'] is not None,\
            'Volume {} is not marked for deletion'.format(name)
```

tests/post/features/volume.feature
```diff
@@ -115,3 +115,88 @@ Feature: Volume management
...
+                operator: "Exists"
+                effect: "NoSchedule"
+              - key: "node-role.kubernetes.io/infra"
+                operator: "Exists"
+                effect: "NoSchedule"
+              - key: "node-role.kubernetes.io/master"
+                operator: "Exists"
+                effect: "NoSchedule"
+        Then a pod 'pod-test-volume-e2e-2' exist with file '/mnt/toto.log'
+        # Enter the pod and check that the file exist
```
slaperche-scality: Is this implemented or not?
If so, remove the comment.

tests/post/steps/test_volume.py
```diff
@@ -153,6 +157,49 @@ def delete_pvc(host, name, k8s_client):
         grace_period_seconds=0
     )
 
+
+@when(parsers.parse("I delete the pod '{name}'"))
+def delete_namespaced_pod(host, name, k8s_client):
+    # ensure pod is available before deleting
```
slaperche-scality: No.

Pods are created with a `given`, the `given` should wait for availability. Not the delete.

tests/post/steps/test_volume.py
```diff
@@ -94,6 +99,22 @@ def _check_status(status='Available'):
...
     )
 
+
+
+@given(parsers.parse("I create the following object:\n{body}"))
+def create_pvc(host, body, k8s_client):
+    k8s_client.create_namespaced_persistent_volume_claim(
+        namespace='monitoring',
+        body=yaml.safe_load(body)
+    )
```
slaperche-scality: It's a `given` you should follow the same pattern as form Volume:

```
if exist(…):
    return
create(…)
wait for exist/available (depending on the object)
```

tests/post/steps/test_volume.py
```diff
@@ -94,6 +99,22 @@ def _check_status(status='Available'):
...
+
+
+@given(parsers.parse("I create the following Pod:\n{body}"))
+def create_namespace_pod(host, body, k8s_client):
+    k8s_client.create_namespaced_pod(
+        namespace='monitoring',
+        body=yaml.safe_load(body)
+    )
```
slaperche-scality: ditto

```
if exist(…):
    return
create(…)d
wait for exist/available (depending on the object)
```

Also, keep an empty line before closing `# }}}`.

tests/post/steps/test_volume.py
```diff
@@ -303,6 +350,45 @@ def _check_deletion_timestamp():
...
+        except KeyError as exc:
+            assert False, 'Unexpected key error {} for pod {}'.format(
+                exc,
+                name
+            )
+    utils.retry(
+        _check_pod_status, times=30, wait=2,
+        name='checking status of Pod {}'.format(name)
+    )
+    with host.sudo():
```
slaperche-scality: I don't think you need to have a retry for the pod availability in a retry.

Just have a retry around the code that check the for the file.
Also, we may want to check file using the API instead of `kubectl` + we also want to check the content of the file.

slaperche-scality (changes requested):

tests/post/steps/test_volume.py
slaperche-scality: And for the wait on a pod, you should reuse `kube_utils.wait_for_pod`.

tests/post/steps/test_volume.py
Ebaneck: Not sure, i can think of a better way to check the file using the API and it's content.

Can you enlighten me here??

slaperche-scality: Take a look in `tests/install/steps/test_expansion.py`.

We're running an `etcdctl member list` command from inside the `etcd` container and get the output.

```python
etcd_member_list = etcdctl(k8s_client, ['member', 'list'], ssh_config)
[…]
def etcdctl(k8s_client, command, ssh_config):
    """Run an etcdctl command inside the etcd container."""
    name = 'etcd-{}'.format(
        utils.resolve_hostname('bootstrap', ssh_config)
    )
    etcd_command = [
        'etcdctl',
        '--endpoints', 'https://localhost:2379',
        '--ca-file', '/etc/kubernetes/pki/etcd/ca.crt',
        '--key-file', '/etc/kubernetes/pki/etcd/server.key',
        '--cert-file', '/etc/kubernetes/pki/etcd/server.crt',
    ] + command
    output = k8s.stream.stream(
        k8s_client.connect_get_namespaced_pod_exec,
        name=name, namespace='kube-system',
        command=etcd_command,
        stderr=True, stdin=False, stdout=True, tty=False
    )
    return output
```

tests/post/features/volume.feature
```diff
@@ -8,3 +8,194 @@ Feature: Volume management
...
+              volumes:
+                - name: pod-storage
+                  persistentVolumeClaim:
+                    claimName: e2e-pvc
+              containers:
+                - name: container-test-volume-e2e
+                  image:  busybox
+                  command:
+                  - sh
+                  args: ["-c", 'echo "toto" >> /mnt/toto.log && sleep 60']
```
Ebaneck: Might be useful here to add a longer sleep interval since the container might exit after the sleep duration.
Also we have to account for the following
- pod creation time
- time for pod to become available(PV and PVC bound time)

I have noticed situations in which the check for pod availability fails.
And by the way, extending the sleep duration to something like 120s is not harmful since we delete this container thereafter.

tests/post/steps/test_volume.py
Ebaneck: @slaperche-scality, i think i implemented the wait here because in the complete E2E test workflow, when we create the pod, it cannot be in available state because it has a pvc which is not yet fulfilled since the volume is only created in the `when`

Ebaneck: Thinking we should create the volume in the `given` to furfill this check

BTW, there is nothing special about the volume definition.

Something like: 
```
    Scenario: Complete E2E volume workflow
        Given the following PersistentVolumeClaim 'e2e-pvc' exist:
         .....
        And a Volume 'volume9' exist```

slaperche-scality (approved):

PR #1539, opened 2019-08-21
Prepare environment for calico integration for Debian

### LINKED ISSUE #1538: Preparation for calico integration
**Component**: buildchain.

**Why this is needed**: Need to prepare the environment before to modify buildchain scrits to integrate calico for Debian.

**What should be done**: Prepare the environment.

- Add some constants for Debain
- Create function for debian packages
- Prepare the containerd to convert rpm files to deb files

**Test plan**:
Deployment still should work.

ycointe:
**Component**: Buildchain

**Summary**: Prepare the environment before to modify buildchain scrits to integrate calico for Debian.

**Acceptance criteria**: 
- Deploy cluster as usual.

---

Closes: #1538

### DIFF
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -33,12 +33,16 @@ ISO_ROOT : Path = config.BUILD_ROOT/'root'
 REPO_ROOT : Path = ISO_ROOT/'packages'
 # Root of the RedHat repositories on the ISO.
 REPO_RPM_ROOT : Path = REPO_ROOT/'redhat'
+# Root of the Debian repositories on the ISO.
+REPO_DEB_ROOT : Path = REPO_ROOT/'debian'
 # Root for the images on the ISO.
 ISO_IMAGE_ROOT : Path = ISO_ROOT/'images'
 # Root for the packages that we build ourselves.
 PKG_ROOT : Path = config.BUILD_ROOT/'packages'
 # Root for the RPM packages that we build ourselves.
 PKG_RPM_ROOT : Path = PKG_ROOT/'redhat'
+# Root for the Debian packages that we build ourselves.
+PKG_DEB_ROOT : Path = PKG_ROOT/'debian'
 # Root of the Vagrant environment folder.
 VAGRANT_ROOT : Path = ROOT/'.vagrant'
 # Path to the static-container-registry module.
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -73,6 +73,12 @@ def task__package_mkdir_rpm_root() -> types.TaskDict:
         directory=constants.PKG_RPM_ROOT, task_dep=['_package_mkdir_root']
     ).task
 
+def task__package_mkdir_deb_root() -> types.TaskDict:
+    """Create the Debian packages root directory."""
+    return targets.Mkdir(
+        directory=constants.PKG_DEB_ROOT, task_dep=['_package_mkdir_root']
+    ).task
+
 def task__package_mkdir_iso_root() -> types.TaskDict:
     """Create the packages root directory on the ISO."""
     return targets.Mkdir(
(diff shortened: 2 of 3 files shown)


buildchain/buildchain/packaging.py
```diff
@@ -73,6 +73,12 @@ def task__package_mkdir_rpm_root() -> types.TaskDict:
         directory=constants.PKG_RPM_ROOT, task_dep=['_package_mkdir_root']
     ).task
 
+def task__package_mkdir_deb_root() -> types.TaskDict:
+    """Create the Debian packages root directory."""
+    return targets.Mkdir(
+        directory=constants.PKG_DEB_ROOT, task_dep=['_package_mkdir_root']
+    ).task
+
 def task__package_mkdir_iso_root() -> types.TaskDict:
```
slaperche-scality, line 82: Something that we missed in the previous PR: this function should be renamed `task__package_mkdir_rpm_iso_root`

packages/redhat/Dockerfile
```diff
@@ -16,6 +16,9 @@ RUN yum install -y \
         rpmlint \
         yum-utils \
         && \
+    yum install -y \
```
slaperche-scality: Why running a new `yum install -y` when you could simply add `alien` to the list of installed package by the existing `yum install -y`?

slaperche-scality (changes requested):

buildchain/buildchain/packaging.py
ycointe: I don't understand, the task `task__package_mkdir_rpm_iso_root` exist 6 lines after. 
The tasks tagged with an OS familly like `rpm` or `deb` call the tasks untagged like, in this case, `task__package_mkdir_iso_root`
Please tell me if I missed something

packages/redhat/Dockerfile
ycointe: Because during my tests, when I add `alien` with all other packages, it doesn't work.
I think it's a problem of dependencies uninstalled until all the previous packages are installed.
Add a new line with a new `yum install -y` is the only way i found to install `alien`

NicolasT: `alien` sits in EPEL, not in the core repositories. Since we only install `epel-release` (which configures the EPEL repository on the system) in the first `yum install` invocation. Before that, EPEL is not available, so trying to install something from it fails.

So, not an issue and the second `yum install` invocation is fine (heck, there's no other way).

buildchain/buildchain/constants.py
```diff
@@ -33,12 +33,16 @@
 REPO_ROOT : Path = ISO_ROOT/'packages'
 # Root of the RedHat repositories on the ISO.
 REPO_RPM_ROOT : Path = REPO_ROOT/'redhat'
+# Root of the Debian repositories on the ISO.
+REPO_DEB_ROOT : Path = REPO_ROOT/'debian'
```
NicolasT, line 37: So, is it Debian or is it Ubuntu? Those are not the same...

buildchain/buildchain/packaging.py
slaperche-scality: Meh, go cheated by the Github display:

meh

Didn't see that there was a line gap…
Everything is fine, forget it ^^

packages/redhat/Dockerfile
slaperche-scality: Weird, what kind of error do you get?

buildchain/buildchain/constants.py
ycointe: Debian, we sort by os-family.

packages/redhat/Dockerfile
slaperche-scality: Ho ok, makes sense.

slaperche-scality (dismissed):

ycointe:
@bert-e approve

slaperche-scality (approved):

PR #1543, opened 2019-08-21
Don't use matchFields in generated PersistentVolume

### LINKED ISSUE #1542: Workaround k8s bug regarding matchFields on PersistentVolume
**Component**:

operator

**What happened**:

PV created by a Volume can't be used by Pods.
See https://github.com/kubernetes/kubernetes/issues/81725 for the underlying issue.

**What was expected**:

Pod scheduling succeeds.

**Steps to reproduce**:

- Create a Volume using `metalk8s-prometheus` as StorageClass
- Create a PVC that would bind to this PV (i.e., use the SC, compatible size,...)
- Create a Pod that uses said PVC as a volume

**Resolution proposal**:

When creating the PV from the Volume, don't use `matchFields` (only use `matchExpressions`).

slaperche-scality:
**Component**:

operator

**Context**: 

Workaround https://github.com/kubernetes/kubernetes/issues/81725

**Summary**:

Don't use `matchFields` in the generated PersistentVolume.

**Acceptance criteria**: 

Created PV can be used by Pod.

---

Closes: #1542

### DIFF
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -824,13 +824,6 @@ func nodeAffinity(node types.NodeName) *corev1.VolumeNodeAffinity {
 						Values:   []string{string(node)},
 					},
 				},
-				MatchFields: []corev1.NodeSelectorRequirement{
-					{
-						Key:      "metadata.name",
-						Operator: corev1.NodeSelectorOpIn,
-						Values:   []string{string(node)},
-					},
-				},
 			},
 		},
 	}


NicolasT (approved):
We may want to file a ticket to track the fact this should be reverted (and the label match removed) once https://github.com/kubernetes/kubernetes/pull/81784 (or another fix for https://github.com/kubernetes/kubernetes/issues/81725) has landed, and we ship a K8s version which contains said fixes.

slaperche-scality:
> We may want to file a ticket to track the fact this should be reverted (and the label match removed) […]

Tracked by #1548

PR #1544, opened 2019-08-22
Split classes to handle different kind of packages

### LINKED ISSUE #1472: integrate sos-report and calico-cni debian packages
**Component**: package

**Why this is needed**: Because Alien doesn't work for sos-report to convert it to a rpm to deb package

**What should be done**: 
- create sos-report debian source 
- integrate it in the buildchain
- modify the buildchain to build the sources
- modify the buildchain to convert calico

- [x] Split packages for debian packages integration

- [x] Split the repository class to handle Debian & RedHat repo separatly

- [x]  modify docker calls to use a specify entrypoint script  in depending of the os family 

- [x]  Correct the path of sos-resources  sources used in the mounted volum to build it in Debian  package

- [x]  integrate Debian package builder classes in the top level to permit to use them

- [x]  Modify entrypoint to build & convert packages

**Test plan**: verify that calico is well installed

ycointe:
**Component**: Buildchain, package

**Summary**: We need to to split some classes in the buildchain to integrate Debian and build its packages.

**Acceptance criteria**: 
- classes separated
- Build still available 
- Centos cluster deployment available

---

Closes: #1472

### DIFF
--- a/buildchain/buildchain/targets/package.py
+++ b/buildchain/buildchain/targets/package.py
@@ -34,6 +34,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Sequence
 
 from buildchain import constants
+from buildchain import coreutils
 from buildchain import types
 from buildchain import utils
 from buildchain import docker_command
@@ -319,3 +320,114 @@ def _file_from_url(url: str) -> str:
     """Get filename from a URL."""
     path = urllib.parse.urlparse(url).path
     return urllib.parse.unquote(os.path.basename(path))
+
+
+class DEBPackage(Package):
+    """A DEB software package for Ubuntu 18.04."""
+
+    ARCH = 'amd64'
+
+    def __init__(
+        self,
+        basename: str,
+        name: str,
+        version: str,
+        sources: Path,
+        build_id: int,
+        builder: image.ContainerImage,
+        **kwargs: Any
+    ):
+        super().__init__(
+            basename, name, version, build_id, builder,
+            constants.PKG_DEB_ROOT, **kwargs
+         )
+        self._sources = sources
+
+    sources = property(operator.attrgetter('_sources'))
+
+    @property
+    def deb(self) -> Path:
+        """DEB path."""
+        fmt = '{pkg.name}_{pkg.version}-{pkg.build_id}_{pkg.ARCH}.deb'
+        return self.rootdir/fmt.format(pkg=self)
+
+    @property
+    def debuild_sources(self) -> Path:
+        """Path to the directory that contains input files for debuild."""
+        return constants.ROOT.joinpath('packages','debian',self.name)
+
+    @property
+    def execution_plan(self) -> List[types.TaskDict]:
+        tasks = [self.make_package_directory()]
+        if self.sources.suffix == '.rpm':
+            tasks.append(self.convert_package())
+        else:
+            tasks.append(self.build_package())
+        return tasks
+
+    def build_package(self) -> types.TaskDict:
+        """Build DEB packages from source files."""
+        mounts = [
+            utils.bind_ro_mount(
+                source=self.sources, target=Path('/debbuild/pkg-src')
+            ),
+            utils.bind_ro_mount(
+                source=self.debuild_sources, target=Path('/debbuild/pkg-meta')
+            ),
+            utils.bind_mount(
+                source=self.rootdir, target=Path('/debbuild/result')
+            ),
+        ]
+        builddeb_callable = docker_command.DockerRun(
+            command=['/entrypoint.sh', 'builddeb'],
+            builder=self.builder,
+            run_config=docker_command.DEB_BASE_CONFIG,
+            mounts=mounts,
+            environment={
+                'VERSION': '{}-{}'.format(self.version, self.build_id)
+            },
+        )
+        task = self.basic_task
+        task.update({
+            'name': 'build_deb_pkg',
+            'actions': [builddeb_callable],
+            'doc': 'Build DEB package from sources for {}'.format(self.name),
+            'title': utils.title_with_target1('BUILD DEB'),
+            'targets': [self.deb],
+        })
+        task['file_dep'].extend(coreutils.ls_files_rec(self.sources))
+        task['file_dep'].extend(coreutils.ls_files_rec(self.debuild_sources))
+        task['task_dep'].append('_package_mkdir_deb_iso_root')
+        task['task_dep'].append('_build_deb_container')
+        return task
+
+    def convert_package(self) -> types.TaskDict:
+        """Build a DEB package from a RPM one."""
+        mounts = [
+            utils.bind_ro_mount(
+                source=self.sources,
+                target=Path('/rpmbuild/source.rpm')
+            ),
+            utils.bind_mount(
+                source=self.rootdir,
+                target=Path('/debbuild/result')
+            ),
+        ]
+        builddeb_callable = docker_command.DockerRun(
+            command=['/entrypoint.sh', 'rpm2deb'],
+            builder=self.builder,
+            run_config=docker_command.DEB_BASE_CONFIG,
+            mounts=mounts
+        )
+        task = self.basic_task
+        task.update({
+            'name': 'convert_rpm_pkg_to_deb',
+            'actions': [builddeb_callable],
+            'doc': 'Build DEB package from RPM for {}'.format(self.name),
+            'title': utils.title_with_target1('RPM2DEB'),
+            'targets': [self.deb],
+        })
+        task['file_dep'].append(self.sources)
+        task['task_dep'].append('_package_mkdir_deb_iso_root')
+        task['task_dep'].append('_build_deb_container')
+        return task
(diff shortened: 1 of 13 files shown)


packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+set -u
+set -o pipefail
+
+rpm2deb() {
+	set -x
+	chown build:build /home/build
```
alexandre-allard: Why chowning the home directory, the ownership should already be right, no ?

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+	cp -p /rpmbuild/SRC/*.deb /debbuild/PKG
+	chown "${TARGET_UID}:${TARGET_GID}" \
+		/debbuild/DEBS/*
+}
+
+builddeb() {
+	#need to be done with user "build"
+	set -x
+	chown build:build /home/build
+        pushd /debbuild/PKG 1>/dev/null
```
alexandre-allard: Why moving in this directory since the line just after you use a full path and right after you move again ?
For me, this line is useless, must be a leftover of a copy/paste from your function above.

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -e
+set -u
+set -o pipefail
+
+rpm2deb() {
+	set -x
+	chown build:build /home/build
+	pushd /debbuild/PKG 1>/dev/null
+	for RPM in /rpmbuild/SRC/calico*.rpm; do
```
alexandre-allard: Should be based on a list of packages to alienize, not hardcoded.

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+builddeb() {
+	#need to be done with user "build"
+	set -x
+	chown build:build /home/build
+        pushd /debbuild/PKG 1>/dev/null
+	for DEBSRC in /debbuild/PKG; do
+		pushd /debbuild/PKG/metalk8s-sosreport/debian
+		debuild -b -uc -us
+	done
+	cp -p /debbuild/SRC/* /debbuild/PKG
```
alexandre-allard: Not really an error, but is there any good reason for using `-p` ? (just for my personal understanding)

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+	chown "${TARGET_UID}:${TARGET_GID}" \
+		/debbuild/DEBS/*
+}
+
+builddeb() {
+	#need to be done with user "build"
+	set -x
+	chown build:build /home/build
+        pushd /debbuild/PKG 1>/dev/null
+	for DEBSRC in /debbuild/PKG; do
```
alexandre-allard: Will not work, you need to use a wildcard:
```
for debsrc in /debbuild/PKG/*; do
```
And conventionally, you use lowercased variable when the purpose of the variable is local.
Uppercase is reserved for global variables.

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+		pushd /debbuild/PKG/metalk8s-sosreport/debian
```
alexandre-allard: Hardcoded package name, you're not using the variable set before `$DEBSRC`

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+	done
+	cp -p /rpmbuild/SRC/*.deb /debbuild/PKG
+	chown "${TARGET_UID}:${TARGET_GID}" \
+		/debbuild/DEBS/*
+}
+
+builddeb() {
+	#need to be done with user "build"
+	set -x
+	chown build:build /home/build
```
alexandre-allard: Same as above

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+        pushd /debbuild/PKG 1>/dev/null
+	for DEBSRC in /debbuild/PKG; do
+		pushd /debbuild/PKG/metalk8s-sosreport/debian
+		debuild -b -uc -us
```
alexandre-allard: Need to be run as build user:
```
su build -c "debuild -b -uc -us"
```

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+set -o pipefail
+
+rpm2deb() {
+	set -x
+	chown build:build /home/build
+	pushd /debbuild/PKG 1>/dev/null
+	for RPM in /rpmbuild/SRC/calico*.rpm; do
+		alien --bump=0 "$RPM"
+	done
+	cp -p /rpmbuild/SRC/*.deb /debbuild/PKG
```
alexandre-allard: Useless as you already moved to `/debbuild/PKG` before your loop and the result files (*.deb) from alien are generated in the current directory.

packages/debian/entrypoint.sh
ycointe: Indeed, It was more a kind of "security" to be sur. I delete it.

packages/debian/entrypoint.sh
ycointe: It was to "preserve" rights on its but no, it was not very needed because of the `chown` applied after.

packages/debian/entrypoint.sh
ycointe: Not necessary, i prefer to use a `chown` after which will apply another owner that `root`.

buildchain/buildchain/targets/package.py
```diff
@@ -319,3 +320,114 @@ def _file_from_url(url: str) -> str:
...
+        builder: image.ContainerImage,
+        **kwargs: Any
+    ):
+        super().__init__(
+            basename, name, version, build_id, builder,
+            constants.PKG_DEB_ROOT, **kwargs
+         )
+        self._sources = sources
+
+    sources = property(operator.attrgetter('_sources'))
```
jbertran, line 346: I think that with a little work the `sources` management can me mutualized into the `Package` class. This can be an improvement ticket I guess.

E: ~~most likely we want to tend towards the current RPM behaviour, where it would be easy to specify 2 different source paths whereas this would be a problem in the current DEB behaviour.~~ after discussion with @slaperche-scality , just create a debt ticket for reworking/rewording the `sources` thing - it turns out I didn't understand it fully but we still have an issue with it.

buildchain/buildchain/targets/package.py
```diff
@@ -319,3 +320,114 @@ def _file_from_url(url: str) -> str:
...
+    @property
+    def execution_plan(self) -> List[types.TaskDict]:
+        tasks = [self.make_package_directory()]
+        if self.sources.suffix == '.rpm':
+            tasks.append(self.convert_package())
+        else:
+            tasks.append(self.build_deb_package())
+        return tasks
+
+    def build_deb_package(self) -> types.TaskDict:
```
jbertran: Can be renamed `build_package` like the `RPMPackage` equivalent.

jbertran (commented):
Dayum, dayum, dayum.

LGTM. Great work!

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+    cp -r /debbuild/pkg-meta/* /debbuild/build-pkg
+    cp -r /debbuild/pkg-src/* /debbuild/build-pkg/src
+
+    # Overwrite version if supported
+    sed -i 's/_VERSION_/'"${VERSION}"'/g' "/debbuild/build-pkg/debian/changelog"
+
+    pushd /debbuild/build-pkg 1>/dev/null
+    debuild -b -uc -us
+
+    cp -r /debbuild/*.deb /debbuild/result
```
alexandre-allard: Not sure you need `-r`, except if you're trying to copy directories named `*.deb`. :)

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+
+# Build a DEB from an RPM file, using alien.
+rpm2deb() {
+    set -x
+
+    alien --bump=0 /rpmbuild/source.rpm
+
+    cp /*.deb /debbuild/result
```
alexandre-allard: And what about:
```
pushd /debbuild/result 1> /dev/null
alien --bump=0 /rpmbuild/source.rpm
chown "$TARGET_UID:$TARGET_GID" *
```

packages/debian/entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+    chown -R "${TARGET_UID}:${TARGET_GID}" /debbuild/result
+}
+
+
+case ${1:-''} in
```
alexandre-allard: Details: Useless quotes, but it does not hurt, I think. :)

packages/debian/entrypoint.sh
ycointe: nop, indeed! :) 
Thanks

buildchain/buildchain/targets/package.py
ycointe: I create a ticket, you can follow the issue REF #1623

packages/debian/entrypoint.sh
ycointe: Change for Debian and RedHat entrypoint :)

ycointe:
@bert-e approve

jbertran (approved):

PR #1547, opened 2019-08-22
ui: update `package-lock.json`

NicolasT:
Due to a change in dependencies of our fork of
`@kubernetes/client-javascript`, we somehow need to recreate this
project's `package-lock.json` to include these new deps ¯\_(ツ)_/¯

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
-      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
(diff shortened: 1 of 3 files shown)


NicolasT:
Let's hope this unblocks CI...

carlito-scality (dismissed):

carlito-scality (commented):
It could unblock the build but the UI should not work because some APIs are deprecated with the new version of the branch ‘browser-dist’...we need to cherry-pick the fix from development/2.4

NicolasT:
Hmh, this is not sufficient to make the Cypress tests pass... so still blocked :(

NicolasT:
Success on 2.[0123], flaky on 2.4, rebuild launched. If successful, this should go in.

Note: the 2.4 integration branch is fixed by hand, no diff with current development/2.4 as intended, except for some updates in `package-lock.json`.

carlito-scality (approved):

carlito-scality:
@NicolasT ```package-lock.json``` of 2.[0123] and that of 2.4 are not the same because dependencies are different. we should update them separately

NicolasT:
> @NicolasT `package-lock.json` of 2.[0123] and that of 2.4 are not the same because dependencies are different. we should update them separately

They are updated separately:

```
$ for x in 1 2 3 4; do echo -n "2.0 -> 2.$x: "; git diff scality/bugfix/npm-deps..scality/w/2.$x/bugfix/npm-deps ui/package-lock.json | wc -l; done
2.0 -> 2.1: 0
2.0 -> 2.2: 0
2.0 -> 2.3: 0
2.0 -> 2.4: 14688
```

PR #1551, opened 2019-08-23
Unprepare volume with Salt

### LINKED ISSUE #1475: Unprepare volume with Salt

slaperche-scality:
**Component**:

salt

**Context**:

Cleanup resources when a `Volume` is deleted.

**Summary**:

Add a new Salt state to cleanup volumes.

**Acceptance criteria**: 

When you delete a `SparseLoopDevice`:
- the loop device disappeared from the output of `losetup -l`
- the backing file was removed from `/var/lib/metalk8s/storage/sparse/`

When you delete a `rawBlockDevice`:
- the disk is still formatted as before, data are left untouched

---

Closes: #1475

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -37,7 +37,7 @@ def exists(name):
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.exists example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.exists example-volume
     """
     return _get_volume(name).exists
 
@@ -48,23 +48,20 @@ def create(name):
     Args:
         name (str): volume name
 
-    Returns:
-        None
-
     CLI Example:
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.create example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.create example-volume
     """
-    return _get_volume(name).create()
+    _get_volume(name).create()
 
 
 def is_provisioned(name):
     """Check if the backing storage device is provisioned for the given volume.
 
     Args:
-        path (str): path of the sparse file
+        name (str): volume name
 
     Returns:
         bool: True if the backing storage device is provisioned, otherwise False
@@ -73,7 +70,7 @@ def is_provisioned(name):
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.is_provisioned example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.is_provisioned example-volume
     """
     return _get_volume(name).is_provisioned
 
@@ -84,16 +81,13 @@ def provision(name):
     Args:
         name (str): volume name
 
-    Returns:
-        str: path to the associated loop device
-
     CLI Example:
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.provision example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.provision example-volume
     """
-    return _get_volume(name).provision()
+    _get_volume(name).provision()
 
 
 def is_formatted(name):
@@ -109,7 +103,7 @@ def is_formatted(name):
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.volume_is_formatted example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.volume_is_formatted example-volume
     """
     return _get_volume(name).is_formatted
 
@@ -120,14 +114,11 @@ def format(name):
     Args:
         name (str): volume name
 
-    Returns:
-        None
-
     CLI Example:
 
     .. code-block:: bash
 
-        salt '*' metalk8s_volumes.format example-volume
+        salt '<NODE_NAME>' metalk8s_volumes.format example-volume
     """
     _get_volume(name).format()
 
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -4,6 +4,7 @@
 import abc
 import contextlib
 import errno
+import fcntl
 import functools
 import json
 import re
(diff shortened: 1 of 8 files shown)


salt/_modules/metalk8s_volumes.py
```diff
@@ -260,6 +307,29 @@ def format(self, force=False):
...
+        try:
+            device_path = '/dev/disk/by-uuid/{}'.format(
+                self.get('metadata.uid')
+            )
+            command = ' '.join(['losetup', '--detach', device_path])
+            _run_cmd(command)
+        except CommandExecutionError as exn:
+            # Gracefully handle if the loop device is already detached.
+            # Return code is too generic, have to check the error message…
+            if 'No such file or directory' not in exn.message:
```
slaperche-scality: Not a big fan of this…

Maybe I should use `os.open` + `ioctl(fd, LOOP_CLR_FD, 0)`, error handling would be more fine-grained.

salt/_modules/metalk8s_volumes.py
```diff
@@ -85,7 +87,7 @@ def provision(name):
         name (str): volume name
 
     Returns:
-        str: path to the associated loop device
+        None
```
NicolasT: Then I guess this `Returns: None` can be removed altogether?

salt/_modules/metalk8s_volumes.py
```diff
@@ -132,6 +134,42 @@ def format(name):
...
+        name (str): volume name
+
+    Returns:
+        bool: True if the backing storage device is cleaned up, otherwise False
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' metalk8s_volumes.is_cleaned_up example-volume
```
NicolasT: You won't want to run this on `*`, no? A volume sits on a particular minion/host.

salt/_modules/metalk8s_volumes.py
```diff
@@ -132,6 +134,42 @@ def format(name):
...
+        name (str): volume name
+
+    Returns:
+        None
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' metalk8s_volumes.clean_up example-volume
```
NicolasT: Same.

salt/_modules/metalk8s_volumes.py
```diff
@@ -260,6 +308,25 @@ def format(self, force=False):
         # We format a "normal" file, not a block device: we need force=True.
         super(SparseLoopDevice, self).format(force=True)
 
+    @property
+    def is_cleaned_up(self):
+        return not (self.is_provisioned or self.exists)
+
+    def clean_up(self):
+        LOOP_CLR_FD = 0x4C01  # From /usr/include/linux/loop.h
```
NicolasT, line 304: Is this portable across archs?

salt/_modules/metalk8s_volumes.py
```diff
@@ -260,6 +308,25 @@ def format(self, force=False):
...
+        device_path = '/dev/disk/by-uuid/{}'.format(self.get('metadata.uid'))
+        try:
+            with _open_fd(device_path, os.O_RDONLY) as fd:
+                try:
+                    fcntl.ioctl(fd, LOOP_CLR_FD, 0)
+                except Exception:
+                    raise
```
NicolasT: So, no need to `try`/`except`

storage-operator/pkg/salt/client.go
```diff
@@ -95,25 +95,35 @@ func (self *Client) PrepareVolume(
...
 //     ctx:      the request context (used for cancellation)
 //     nodeName: name of the node where the volume will be
+//     volumeName: name of the volume to prepare
+//     saltenv:    saltenv to use
 //
 // Returns
 //     The Salt job ID.
 func (self *Client) UnprepareVolume(
-	ctx context.Context, nodeName string,
+	ctx context.Context, nodeName string, volumeName string, saltenv string,
```
NicolasT, line 105: I'm surprised `go fmt` or `go vet` doesn't turn this into `ctx context.Context, nodeName, volumeName, saltenv string`

storage-operator/pkg/salt/client.go
```diff
@@ -95,25 +95,35 @@ func (self *Client) PrepareVolume(
...
+	self.logger.Info(
+		"UnprepareVolume",
+		"Volume.NodeName", nodeName, "Volume.Name", volumeName,
+	)
 
 	ans, err := self.authenticatedRequest(ctx, "POST", "/", payload)
 	if err != nil {
 		return "", errors.Wrapf(
-			err, "UnprepareVolume failed (target=%s)", nodeName,
+			err, "UnrepareVolume failed (target=%s, volume=%s)",
```
NicolasT: Maybe also log the `saltenv`.

NicolasT (changes requested):

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed. Done

salt/_modules/metalk8s_volumes.py

salt/_modules/metalk8s_volumes.py
slaperche-scality: arch as in `endianness`?

I think so, there is no `#define` around the constant in the header.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed, it's a vestigial structure.

storage-operator/pkg/salt/client.go
slaperche-scality: Indeed, maybe it's not a good practice or not good enough to be enforced.

storage-operator/pkg/salt/client.go
slaperche-scality: Yeah, could be useful.

Done

salt/_modules/metalk8s_volumes.py
slaperche-scality: I've just tried on big endian (Debian on MIPS) and it works:

```
root@debian:~# truncate -s 1G foo
root@debian:~# losetup -f foo
[   79.088643] loop: module loaded
root@debian:~# losetup -l
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE DIO LOG-SEC
/dev/loop0         0      0         0  0 /root/foo   0     512
root@debian:~# python2
Python 2.7.16 (default, Apr  6 2019, 01:42:57)
[GCC 8.3.0] on linux2
>>> import platform
>>> platform.machine()
'mips'
>>> import sys
>>> sys.byteorder
'big'
>>> import os
>>> import fcntl
>>> fd = os.open('/dev/loop0', os.O_RDONLY)
>>> fcntl.ioctl(fd, 0x4C01, 0)
0
>>> os.close(fd)
>>>
root@debian:~# losetup -l
root@debian:~#
```

NicolasT: :100:

jbertran (approved):
LGTM

NicolasT (approved):

NicolasT:
I scheduled a rebuild.

PR #1553, opened 2019-08-23
ci: use `npm ci` instead of `npm install`

NicolasT:
**Component**: ci, ui
**Summary**: Instead of using `npm install`, use `npm ci`
**Acceptance criteria**: CI succeeds

Rationale:
- `npm ci` ought to be faster than `npm install`
- `npm install` doesn't necessarily honor `package-lock.json`


gdemonet (approved):
Cool feature, didn't know it existed!

PR #1564, opened 2019-08-26
ui: add `.dockerignore`

NicolasT:
Forgot to add this as part of 46f0a1699b8da2352308a3b64ab40de9d0188cbd.

See: 46f0a1699b8da2352308a3b64ab40de9d0188cbd

### DIFF
--- a/ui/.dockerignore
+++ b/ui/.dockerignore
@@ -1,0 +1,4 @@
+build/
+junit/
+node_modules/
+tmp/


ui/.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
+build/
+tmp/
```
carlito-scality, line 4: I think ```tmp```is not necessary here.

carlito-scality: We can add ```junit/``` to this list too

NicolasT: Cypress created `tmp/cypress/screenshots/...` on my system.

carlito-scality: oh yes. it is for debug

carlito-scality (approved):

PR #1579, opened 2019-08-27
github: add PR labeling action

NicolasT:
See: https://github.com/actions/labeler

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,0 +1,12 @@
+name: Labeler
+on: [pull_request]
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/labeler@v2
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/label.yml
```diff
@@ -0,0 +1,12 @@
+name: Labeler
+on: [pull_request]
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
```
slaperche-scality, line 7: Why the empty lines around this one?

NicolasT: https://github.com/actions/starter-workflows/blob/master/automation/label.yml

slaperche-scality: Ho ok, fair enough.

I simply looked at the the README from your link (https://github.com/actions/labeler) where there were no spaces around this line.

slaperche-scality (approved):

PR #1581, opened 2019-08-28
Use leader-with-lease election mode for the storage-operator

### LINKED ISSUE #1580: Deadlock when the operator's pod is evicted
**Component**:

operator

**What happened**:

I had a cluster where I wanted to create a lot of volumes.
At one point, while I was continuously creating new volumes, the storage resource become low and Kubernetes evicted the pod of the storage-operator:

```
# kubectl -n kube-system describe pods storage-operator-56f98777b9-mr9lp
Events:
  Type     Reason   Age    From                                         Message
  ----     ------   ----   ----                                         -------
  Warning  Evicted  8m44s  kubelet, metalk8s-2de5k-bootstrap.novalocal  The node was low on resource: ephemeral-storage. Container storage-operator was using 11538531, which exceeds its request of 0.
  Normal   Killing  8m44s  kubelet, metalk8s-2de5k-bootstrap.novalocal  Stopping container storage-operator
```

No problem though, because Kubernetes scheduled a new Pod to replace the evicted one:

```
# kubectl -n kube-system get pods
NAME                                                         READY   STATUS    RESTARTS   AGE
[…]
storage-operator-56f98777b9-mr9lp                            0/1     Evicted   0          4h13m
storage-operator-56f98777b9-qz798                            1/1     Running   0          5m31s
```

Unfortunately, even though the Pod is running, the volumes creation was still stalled: the operator wasn't doing its job:

```
# kubectl -n kube-system logs storage-operator-56f98777b9-qz798
{"level":"info","ts":1566929966.220363,"logger":"cmd","msg":"Go Version: go1.12.8"}
{"level":"info","ts":1566929966.2204573,"logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":1566929966.220469,"logger":"cmd","msg":"Version of operator-sdk: v0.9.0"}
{"level":"info","ts":1566929966.2207386,"logger":"leader","msg":"Trying to become the leader."}
{"level":"info","ts":1566929966.2983456,"logger":"leader","msg":"Found existing lock","LockOwner":"storage-operator-56f98777b9-mr9lp"}
{"level":"info","ts":1566929966.3117368,"logger":"leader","msg":"Not the leader. Waiting."}
{"level":"info","ts":1566929967.4399626,"logger":"leader","msg":"Not the leader. Waiting."}
[…]
{"level":"info","ts":1566930468.4896085,"logger":"leader","msg":"Not the leader. Waiting."}
{"level":"info","ts":1566930485.007126,"logger":"leader","msg":"Not the leader. Waiting."}
```

Looks like the evicted operator was still holding onto the leader lock, preventing the new one to step in and resume the work.

**What was expected**:

The new Pod should be able to become leader and take over the work left by the evicted one.

**Steps to reproduce**

- deploy a MetalK8s cluster 
- put yourself in a "low-resource" situation
- start creating a bunch of volume
- at one point the storage-operator Pod will be evicted
- find that the new Pod isn't doing anything, even though it's running

**Analysis**:

Seems like we're affected by this bug: https://github.com/operator-framework/operator-sdk/issues/1305

By reading [this documentation page](https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-leader-election.html) we understand that by default our operator is running with the "leader-for-life" election mode which only release the leadership when the Pod is deleted (because then the `ConfigMap` containing the lock can be garbage collected).

The problem is that evicted Pod are not deleted, thus our evicted Pod keeps a reference onto the `ConfigMap` which then can't be GC'ed. When the new Pod arrive it find a lock already held and kindly wait (forever since the other pod is not running…) => we have a deadlock.

It can be manually solved by deleting the evicted Pod, but this is a manual solution.

To workaround this bug in a gracious way, we will now use the "leader-with-lease" election mode.
That way, the lock will be released after some period of time: since an evicted pod isn't running, it can't renew its lease and will eventually release the lock, allowing the new Pod to step in as leader.

Note that, as documented, this election mode doesn't fully prevent the risk of split brain.
But that should be fine in our case because Salt acts as a serialization point: you can't have two prepare/finalize states running in parallel, and running them one after the other is totally safe (thanks to idempotency).

**Resolution proposal**:

Use "leader-with-lease" election mode.

slaperche-scality:
**Component**:

operator

**Context**: 

Avoid deadlock when the operator's Pod is evicted.

**Summary**:

Use leader-with-lease election mode. 

**Acceptance criteria**: 

- deploy a MetalK8s cluster
- put yourself in a "low-resource" situation
- start creating a bunch of volume
- at one point the storage-operator Pod will be evicted
- find that the new Pod has taken over (you can create a new volume to check).

---

Closes: #1580

### DIFF
--- a/storage-operator/cmd/manager/main.go
+++ b/storage-operator/cmd/manager/main.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
-	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	"github.com/operator-framework/operator-sdk/pkg/restmapper"
@@ -80,19 +79,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctx := context.TODO()
-	// Become the leader before proceeding
-	err = leader.Become(ctx, "storage-operator-lock")
-	if err != nil {
-		log.Error(err, "")
-		os.Exit(1)
-	}
-
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace:          namespace,
-		MapperProvider:     restmapper.NewDynamicRESTMapper,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		Namespace:               namespace,
+		MapperProvider:          restmapper.NewDynamicRESTMapper,
+		MetricsBindAddress:      fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		LeaderElection:          true,
+		LeaderElectionNamespace: "kube-system",
+		LeaderElectionID:        "storage-operator-lock",
 	})
 	if err != nil {
 		log.Error(err, "")
@@ -123,6 +117,7 @@ func main() {
 		{Port: operatorMetricsPort, Name: metrics.CRPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: operatorMetricsPort}},
 	}
 	// Create Service object to expose the metrics port(s).
+	ctx := context.TODO()
 	_, err = metrics.CreateMetricsService(ctx, cfg, servicePorts)
 
 	if err != nil {


storage-operator/cmd/manager/main.go
```diff
@@ -80,19 +79,14 @@ func main() {
...
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace:          namespace,
-		MapperProvider:     restmapper.NewDynamicRESTMapper,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		Namespace:               namespace,
+		MapperProvider:          restmapper.NewDynamicRESTMapper,
+		MetricsBindAddress:      fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		LeaderElection:          true,
+		LeaderElectionNamespace: "kube-system",
+		LeaderElectionID:        "storage-operator-lock",
```
slaperche-scality, line 89: We may want to configure `LeaseDuration` (default 15s) and `RenewDeadline` (default 10s) values if we have an idea of what a "right" value would be.
Otherwise let's keep the default for now…

See https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/manager#Options

charlesprost: I guess the default values are fine as they are.

charlesprost (approved):
Seems good according to documentation. Logic is definitely better than before.

storage-operator/cmd/manager/main.go
slaperche-scality: I think so, we will see when it more widely use if we need to tailor it

NicolasT: We run a single-replica *Deployment* anyway... (which can in some circumstances lead to multiple replicas running concurrently, but anyway).

10s renewal, 15s lease (i.e., 'only' 5s diff) *can* be an issue when a system is under high churn, the process has unreasonably long GC pauses (could be caused by other processes on the same image), when running in VMs on a heavily over-allocated host,...

I guess indeed, we'll see how it plays out in practice. We *must* make sure the whole system can work without data loss/corruption in case multiple operators would run concurrently anyway, i.e. fully idempotent Salt states/actions (slightly concerned about the 'forced' formatting of `sparseLoopDevice`s here), handling conflicting updates to the API-server gracefully (I think we got that covered),...

slaperche-scality: Yup, the conflicting updates are covered, we have some log message about it and we have automatic retry for those.

PR #1584, opened 2019-08-28
UI: Update the core-ui version to be up-to-date with the new build process

carlito-scality:
**Component**:UI

**Context**: 
- We currently have to commit the build 'dist' to the core-ui's source code. This was not a neat solution. The build should be generated when installing core-ui like for ```@kubernetes/client-node```.
See https://github.com/scality/core-ui/pull/69

**Summary**:

**Acceptance criteria**: 
- No regression

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.7.2",
     "@kubernetes/client-node": "github:scality/kubernetes-client-javascript.git#browser-0.10.2-62-g61d8543",
-    "@scality/core-ui": "github:scality/core-ui.git#094b88f",
+    "@scality/core-ui": "github:scality/core-ui.git#e492300",
     "axios": "^0.18.0",
     "formik": "1.5.1",
     "lodash": "4.17.13",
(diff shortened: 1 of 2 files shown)


ui/package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.7.2",
     "@kubernetes/client-node": "github:scality/kubernetes-client-javascript.git#browser-0.10.2-62-g61d8543",
-    "@scality/core-ui": "github:scality/core-ui.git#094b88f",
+    "@scality/core-ui": "github:scality/core-ui.git#feature/improve-build-process",
```
NicolasT: I don't really like it when we rely on branches in remote repositories.

Reason is simple: if the branch is removed from the remote repository, the build here will break. Similarly, when doing a `git bisect` for whatever reason, a rebuild of an old revision may not be possible.

Unless `package-lock.json` is the source of truth and the branch information is not actually used after the lock-file is generated?

carlito-scality: This branch is for test to make sur the the build in metalk8s is green. when it is merged, we will update it with the development/1.0 latest commit hash.

MonPote (approved):

carlito-scality:
@bert-e approve

PR #1594, opened 2019-08-30
Fine grained error handling for Salt state

### LINKED ISSUE #1593: storage-operator: fine-grained handling of state failure
**Component**:

operator

**Why this is needed**:

Because there is some failure that are transient and that we want to retry (cannot run the state because another one is already running) and some that we don't want to retry (because retry it will likely result in the same failure)

**What should be done**:

Handle differently Salt state failure according to the underlying cause.

**Implementation proposal**:

Seems like we get a `retcode` of 1 when a concurrent state is running vs a `retcode` of 2 when our state fails.
Let's use that as discriminant instead of trying to parse the returned error payload.

**Test plan**:

Existing tests should still pass.
And we should be able to add those new test cases:
-  Create a volume with unsupported FS type
- Create a volume using a non-existing StorageClass

slaperche-scality:
**Component**:

operator, test, salt

**Context**:

Make the distinction between Salt state error resulting from the execution of the state itself and the others.

**Summary**:

Check `retcode`.
Add some new tests (and fix bugs uncovered by those).

**Acceptance criteria**: 

The test suite is passing.

---

Closes: #1593

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -226,7 +226,11 @@ class Volume(object):
         if device_info.has_partition:
             raise Exception('backing device `{}` contains a partition table'\
                             .format(self.path))
-        params = self.get('spec.storageClassName.parameters')
+        storage_class = self.get('spec.storageClassName')
+        # If we got a string that means the name wasn't replaced by the object.
+        if isinstance(storage_class, basestring):
+            raise Exception('StorageClass {} not found'.format(storage_class))
+        params = storage_class['parameters']
         # mkfs options, if any, are stored as JSON-encoded list.
         options = json.loads(params.get('mkfsOptions', '[]'))
         fs_type = params['fsType']
--- a/salt/metalk8s/volumes/unprepared/init.sls
+++ b/salt/metalk8s/volumes/unprepared/init.sls
@@ -8,5 +8,5 @@ Clean up backing storage for {{ volume }}:
     - name: {{ volume }}
 {%- else %}
 Volume "{{ volume }}" not found in pillar:
-  test.fail_without_changes: []
+  test.succeed_without_changes: []
 {%- endif %}
--- a/salt/metalk8s/volumes/unprepared/init.sls
+++ b/salt/metalk8s/volumes/unprepared/init.sls
@@ -7,6 +7,13 @@ Clean up backing storage for {{ volume }}:
   metalk8s_volumes.removed:
     - name: {{ volume }}
 {%- else %}
-Volume "{{ volume }}" not found in pillar:
-  test.succeed_without_changes: []
+
+{%- do salt.log.warning('Volume ' ~ volume ~ ' not found in pillar') -%}
+
+Volume {{ volume }} not found in pillar:
+  test.configurable_test_state:
+    - name: {{ volume }}
+    - changes: False
+    - result: True
+    - comment: Volume {{ volume }} not found in pillar
 {%- endif %}
--- a/storage-operator/pkg/salt/client.go
+++ b/storage-operator/pkg/salt/client.go
@@ -198,11 +198,17 @@ func (self *Client) PollJob(
 	// running for Foo…
 	//
 	// So let's check `retcode` to be 100% sure it succeedeed.
-	if int(retcode) != 0 {
+	switch int(retcode) {
+	case 0:
+		jobLogger.Info("Salt job succeeded")
+		return nodeResult, nil
+	case 1: // Concurrent state execution.
 		return nil, fmt.Errorf("Salt job %s failed to run", jobId)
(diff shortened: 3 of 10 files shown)


salt/metalk8s/volumes/unprepared/init.sls
```diff
@@ -8,5 +8,5 @@ Clean up backing storage for {{ volume }}:
     - name: {{ volume }}
 {%- else %}
 Volume "{{ volume }}" not found in pillar:
-  test.fail_without_changes: []
+  test.succeed_without_changes: []
```
NicolasT: Is there some way to still render some kind of 'warning' message in the state's output, e.g. using `test.configurable_test_state` and setting the `comment` to something descriptive?

While indeed this should be 'success', it's also something that should not occur all too frequently, and something we may want to report to a user (beyond the `name` of the state).

storage-operator/pkg/salt/client.go
```diff
@@ -198,11 +199,38 @@ func (self *Client) PollJob(
 	// running for Foo…
 	//
 	// So let's check `retcode` to be 100% sure it succeedeed.
-	if int(retcode) != 0 {
+	switch int(retcode) {
+	case 0:
+		jobLogger.Info("Salt job succeedeed")
```
NicolasT:
```suggestion
		jobLogger.Info("Salt job succeeded")
```

NicolasT (dismissed):

storage-operator/pkg/salt/client.go
slaperche-scality: Done

salt/metalk8s/volumes/unprepared/init.sls
TeddyAndrieux: For me you should log warning in jinja so that you will have real warning output in salt log and in output https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.logmod.html
Because put something directly in the comment of the state result is not really usefull if it's success nobody will read it IMO.

NicolasT: Then do both. Keep in mind that, when the operator invokes Salt, no human will see any Salt logs (except when explicitly looking at the logs on the minion system at that point in time).

slaperche-scality: Done

storage-operator/pkg/salt/client.go
```diff
@@ -206,11 +207,32 @@ func (self *Client) PollJob(
...
 }
 
+func getStateFailureRootCause(output interface{}) string {
+	const non_root_error_prefix string = "One or more requisite failed"
+
+	switch error := output.(type) {
+	case string:
+		return error
+	case map[string]interface{}:
+		for state := range error {
```
jbertran: Oh boy that's confusing... Can we have `state, _` or does Go not like that?

salt/_modules/metalk8s_volumes.py
```diff
@@ -226,7 +226,10 @@ def format(self, force=False):
         if device_info.has_partition:
             raise Exception('backing device `{}` contains a partition table'\
                             .format(self.path))
-        params = self.get('spec.storageClassName.parameters')
+        storage_class = self.get('spec.storageClassName')
+        if isinstance(storage_class, basestring):
```
jbertran, line 231: Much confused here, could we have a comment explaining why the negative case is a `basestring`? It's not immediately obvious since we're departing from the usual `get` behaviour.

tests/post/steps/test_volume.py
```diff
@@ -373,6 +379,34 @@ def _check_file_content():
         name='checking content of {} on Pod {}'.format(path, name)
     )
 
+
+@then(parsers.parse("the backing storage for Volume '{name}' is created"))
+def check_storage_is_created(context, host, name):
+    volume = context.get(name)
+    assert volume is not None, 'volume {} not found in context'.format(name)
+    assert 'sparseLoopDevice' in volume['spec'], 'unsupported volume type'
```
jbertran: Worth mentioning in the error message that it's unsupported _for this test_.

jbertran (commented):
Noice. Just a couple of minor comments.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah, deserve a comment indeed.

tests/post/steps/test_volume.py
slaperche-scality: Done

storage-operator/pkg/salt/client.go
slaperche-scality: renamed the variable to be more explicit

jbertran (approved):

PR #1598, opened 2019-08-31
Support solution removal through solution-manager

### LINKED ISSUE #1596: Add the possibility to remove a solution through solution manager
**Component**:
solutions

**Why this is needed**:
until now the solution-manager script only adds new solutions, it will be useful to have solution deletion supported
**What should be done**:

**Test plan**:

sayf-eddine-scality:
Fixes: #1596

**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #1596

### DIFF
--- a/scripts/solution-manager.sh
+++ b/scripts/solution-manager.sh
@@ -16,16 +16,23 @@ PILLAR=""
 _usage() {
     echo "solution-manager.sh [options]"
     echo "Options:"
-    echo "-s/--solution <solution path>:   Path to solution ISO"
+    echo "-a/--add <solution path>:        Add solution from ISO path"
+    echo "-d/--del <solution path>:        Uninstall solution from ISO path"
     echo "-l/--log-file <logfile_path>:    Path to log file"
     echo "-v/--verbose:                    Run in verbose mode"
 }
 
-SOLUTIONS=()
+SOLUTIONS_ADD=()
+SOLUTIONS_REMOVE=()
+EXISTENT_SOLUTIONS=()
 while (( "$#" )); do
   case "$1" in
-    -s|--solution)
-      SOLUTIONS+=("$2")
+    -a|--add)
+      SOLUTIONS_ADD+=("$2")
+      shift 2
+      ;;
+    -d|--del)
+      SOLUTIONS_REMOVE+=("$2")
       shift 2
       ;;
     -v|--verbose)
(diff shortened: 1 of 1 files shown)


scripts/solution-manager.sh
```diff
@@ -16,16 +16,23 @@ PILLAR=""
 _usage() {
     echo "solution-manager.sh [options]"
     echo "Options:"
-    echo "-s/--solution <solution path>:   Path to solution ISO"
+    echo "-a/--add <solution path>:        Add solution from ISO path"
+    echo "-d/--del <solution path>:        uninstall solution from ISO path"
```
slaperche-scality: Uninstall (use uppercase, to be consistent with the other strings)

scripts/solution-manager.sh
```diff
@@ -16,16 +16,23 @@ PILLAR=""
...
     echo "-l/--log-file <logfile_path>:    Path to log file"
     echo "-v/--verbose:                    Run in verbose mode"
 }
 
-SOLUTIONS=()
+SOLUTIONS_ADD=()
+SOLUTIONS_REMOVE=()
+existent_solutions=()
```
slaperche-scality: Why this one is not in uppercase?

scripts/solution-manager.sh
```diff
@@ -175,29 +182,52 @@ containsElement () {
...
-        fi
-    done
+    # Add new solutions
+    if [ ${#SOLUTIONS_ADD[@]} -ge 1 ]; then
+        for solution in "${SOLUTIONS_ADD[@]}"; do
+            if ! containsElement "'$solution'" \
+                 "${EXISTENT_SOLUTIONS[@]+"${EXISTENT_SOLUTIONS[@]}"}"; then
+                EXISTENT_SOLUTIONS+=("'$solution'")
+            fi
+        done
```
TeddyAndrieux: To be consistent can you create the function `add_solutions`

sayf-eddine-scality: does it make sense to have a function with just `EXISTENT_SOLUTIONS+=("'$solution'")` ?

TeddyAndrieux: The function will also do the `for` loop + the `if ! containsElement`.
So you will have something like
```
if "SOLUTIONS_ADD not empty" then
  add_soltuions SOLUTIONS_ADD
if "SOLTUIONS_REMOVE not empty" then
  remove_soltuions SOLUTIONS_REMOVE
```
Which is cleaner IMO

TeddyAndrieux (approved):
LGTM

PR #1608, opened 2019-09-03
Add gzip to nginx conf

MonPote:
**Component**: ui

**Context**: 
Currently the browser need to download 3Mb of JS. We can drastically decrease this number by compression thanks to nginx.

**Summary**:

- Enable gzip

This will decrease the size of the file the client need to download
In the current version we have ~7x improvement.
From 3.17Mb to 0.470Mb send to the client.

**Acceptance criteria**: 

You can reach the UI and play with it. Everything should be the same.
When you load the UI for the first time, open the console before and on the network tab you should see for the JS files something like [this](https://puu.sh/EcU7g/4a22ffde50.png) instead of something like [this](https://puu.sh/EcU6m/9ff1243eed.png)


ui/conf/nginx.conf
```diff
@@ -12,13 +28,14 @@ server {
         try_files $uri $uri/ /index.html;
     }
 
+    # Provide the ConfigMaps generated by MetalK8s
```
slaperche-scality, line 30: not related to gzip

ui/conf/nginx.conf
```diff
@@ -12,13 +28,14 @@ server {
...
     location /config.json {
         alias /etc/metalk8s/ui/config.json;
     }
 
+    # Use the theme provided by the customer
```
slaperche-scality, line 35: ditto (should be in a dedicated commit "add comments" or something like that.

ui/conf/nginx.conf
```diff
@@ -2,8 +2,24 @@ server {
     listen       80;
     server_name  localhost;
 
-    #charset koi8-r;
-    #access_log  /var/log/nginx/host.access.log  main;
+    # https://www.digitalocean.com/community/tutorials/how-to-increase-pagespeed-score-by-changing-your-nginx-configuration-on-ubuntu-16-04
```
slaperche-scality: I would rather link the official nginx documentation here: https://nginx.org/en/docs/http/ngx_http_gzip_module.html instead of a "random" blog.

ui/conf/nginx.conf
```diff
@@ -2,8 +2,24 @@ server {
...
+    gzip on;
+    gzip_comp_level    5;
+    gzip_min_length    256;
+    gzip_proxied       any;
```
slaperche-scality: Do we need this one?

ui/conf/nginx.conf
```diff
@@ -2,8 +2,24 @@ server {
...
+    gzip_vary          on;
```
slaperche-scality: Ditto, do we need this.

Seems useful to handle a very rare case so not sure it's worth it.

slaperche-scality (dismissed):

ui/config-overrides.js
```diff
@@ -1,3 +1,13 @@
-const { useBabelRc, override, useEslintRc } = require('customize-cra');
+const {
```
slaperche-scality, line 1: Hum, wrong commit?

Looks like this change belongs to "Add compression-webpack-plugin to create gziped file durint the build" instead of "Add gzip to nginx conf".

ui/conf/nginx.conf
```diff
@@ -2,8 +2,23 @@ server {
...
-    #charset koi8-r;
-    #access_log  /var/log/nginx/host.access.log  main;
+    # https://nginx.org/en/docs/http/ngx_http_gzip_module.html
+    gzip on;
+    gzip_static on;
+    gzip_comp_level    5;
+    gzip_min_length    256;
+
+    gzip_types
+    application/javascript
```
NicolasT: Consider indenting these a bit more, for clarity.

ui/config-overrides.js
```diff
@@ -1,3 +1,13 @@
...
+  useEslintRc,
+  addWebpackPlugin
+} = require('customize-cra');
+const CompressionPlugin = require('compression-webpack-plugin');
 
-module.exports = override(useBabelRc(), useEslintRc());
+module.exports = override(
+  useBabelRc(),
+  useEslintRc(),
+  addWebpackPlugin(new CompressionPlugin())
```
NicolasT: Add a `,` at the end so later commits won't change this line for no good reason.

NicolasT (changes requested):

ui/conf/nginx.conf
MonPote: Yep, I just add an indent.

ui/config-overrides.js
MonPote: I just add the rule on .prettierc that will also enforce that rule in the entire project :)

NicolasT (approved):
OK. However:

- Could you file a ticket for code/bundle-splitting.
- Could you file a ticket for configuring Webpack's bundle-size-limits-becomes-a-build-error feature, so CI catches bundle growth accordingly?

Both of these could be handled in the same PR of course.

MonPote:
@bert-e approve

NicolasT:
For reference, https://webpack.js.org/configuration/performance/#performancehints

MonPote:
> For reference, https://webpack.js.org/configuration/performance/#performancehints



> OK. However:
> 
> * Could you file a ticket for code/bundle-splitting.
> * Could you file a ticket for configuring Webpack's bundle-size-limits-becomes-a-build-error feature, so CI catches bundle growth accordingly?
> 
> Both of these could be handled in the same PR of course.

Ok let me do that.

Issue #1624, opened 2019-09-05
v1beta1.metrics.k8s.io not working


NicolasT:
The `v1beta1.metrics.k8s.io` *APIService* is not working:

```
[root@bootstrap ~]# kubectl describe apiservice v1beta1.metrics.k8s.io
Name:         v1beta1.metrics.k8s.io
Namespace:
Labels:       <none>
Annotations:  <none>
API Version:  apiregistration.k8s.io/v1
Kind:         APIService
Metadata:
  Creation Timestamp:  2019-09-05T09:32:38Z
  Resource Version:    923
  Self Link:           /apis/apiregistration.k8s.io/v1/apiservices/v1beta1.metrics.k8s.io
  UID:                 19da6c42-cfc0-11e9-a93b-08002791ec69
Spec:
  Group:                   metrics.k8s.io
  Group Priority Minimum:  100
  Service:
    Name:            prometheus-adapter
    Namespace:       metalk8s-monitoring
  Version:           v1beta1
  Version Priority:  100
Status:
  Conditions:
    Last Transition Time:  2019-09-05T09:32:38Z
    Message:               no response from https://10.96.10.86:443: Get https://10.96.10.86:443: dial tcp 10.96.10.86:443: connect: connection refused
    Reason:                FailedDiscoveryCheck
    Status:                False
    Type:                  Available
Events:                    <none>
```

This causes namespace deletion to fail (the `kubernetes` finalizer isn't removed because the controller is unable to validate all known object types are deleted from the *Namespace*, since listing all known object types fails: more background in https://github.com/kubernetes/kubernetes/issues/66735).

NicolasT:
The issue is 'fixed' (manually) when
- editing the `prometheus-adapter` *Service* to set a proper `targetPort` (named or numbered)
- editing the `v1beta1.metrics.k8s.io` *APIService* to set `insecureSkipTLSVerify: true` properly in its `spec`

Given these seem to be set correctly in the manifest we apply, this is likely a bug in the `metalk8s_kubernetes` Salt *renderer*.

Lincoln-dac:
i have no prometheus-adapter but how fix

NicolasT:
> i have no prometheus-adapter but how fix

In MetalK8s? This is fixed in https://github.com/scality/metalk8s/pull/1630 so if you're running a sufficiently recent build, this should not be an issue.

PR #1630, opened 2019-09-05
salt: fix deployment of `metrics.k8s.io` *APIService*

NicolasT:
This PR eventually fixes the `metrics.k8s.io` `APIService` deployment, but takes a couple of detours to also fix latent bugs discovered by introducing a fix for the primary aim/issue of this patch series...

Fixes: https://github.com/scality/metalk8s/issues/1624

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -1671,6 +1671,8 @@ def __dict_to_service_spec(spec):
                     port_kwargs = {"port": port}
                 kube_port = kubernetes.client.V1ServicePort(**port_kwargs)
                 spec_obj.ports.append(kube_port)
+        elif key == 'clusterIP':
+            spec_obj.cluster_ip = value
         elif hasattr(spec_obj, key):
             setattr(spec_obj, key, value)
 
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -1663,6 +1663,10 @@ def __dict_to_service_spec(spec):
                     for port_key, port_value in iteritems(port):
                         if port_key in ("name", "port", "protocol", "node_port", "target_port"):
                             port_kwargs[port_key] = port_value
+                        elif port_key == 'nodePort':
+                            port_kwargs['node_port'] = port_value
+                        elif port_key == 'targetPort':
+                            port_kwargs['target_port'] = port_value
                 else:
                     port_kwargs = {"port": port}
                 kube_port = kubernetes.client.V1ServicePort(**port_kwargs)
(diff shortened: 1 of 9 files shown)


NicolasT:
@bert-e reset

NicolasT:
@bert-e reset

salt/metalk8s/addons/monitoring/prometheus/deployed.sls
```diff
@@ -27,7 +27,7 @@ Expose Prometheus:
           port: 9090
           protocol: TCP
           node_port: 30222
-          targetPort: api
+          targetPort: web
```
slaperche-scality, line 30: So upstream uses a mix of snake\_case (`node_port`) and camelCase (`targetPort`)? meh…

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -1663,6 +1663,10 @@ def __dict_to_service_spec(spec):
                     for port_key, port_value in iteritems(port):
                         if port_key in ("name", "port", "protocol", "node_port", "target_port"):
                             port_kwargs[port_key] = port_value
+                        elif port_key == 'nodePort':
```
slaperche-scality, line 1666: So we support several letter casing for the same parameter?

slaperche-scality (approved):

salt/_modules/metalk8s_kubernetes.py
NicolasT: `node_port` is 'standard Salt' (and Python K8s API client). `nodePort` is 'standard K8s YAML'.

NicolasT: So, in a sense this is a bit of an ugly work-around to support both use-cases... The whole interaction with K8s API from Salt ought to be revised anyway (there are some old tickets about this)

salt/metalk8s/addons/monitoring/prometheus/deployed.sls
NicolasT: No. Upstream is definitely 'standard K8s YAML', i.e.`nodePort`. This `deployed.sls` is *not* from upstream, but added by us to create the `NodePort` *Service* for Prometheus. As such:
- `node_port` was used 'to make it work' (I guess)
- Due to the bug in the name handling, `targetPort` was simply ignored
- Things still 'worked' because the (numeric) `targetPort` is 9090, and `targetPort` defaults to `port` if not set (and even though it's set here, it wasn't passed to K8s API, so that one rightfully did what it's supposed to do)
- Due to my fix, `targetPort` is no longer not passed (heh), so it was set to the named port `api`, but our Prometheus *Pod*s don't *have* a port named `api`, so this was incorrect from the beginning.

PR #1632, opened 2019-09-05
Improvement/1618 import only useful function from lodash and kubernetes/client-node

### LINKED ISSUE #1618: Add code splitting with webpack
**Component**: ui, build

**Why this is needed**:
https://github.com/scality/metalk8s/pull/1608#pullrequestreview-283667885

Right now we have 2 big chunks of JS (1 of 3Mb and 1 of 400kb). Webpack can help us to split it in smaller chunk so that the user does not have to download everything at first because using the app.

**What should be done**:

We might want to use this plugin to help us achieve that.
https://webpack.js.org/plugins/split-chunks-plugin/

MonPote:
**Component**: ui, build

**Context**: 
Our build is pretty big, 3MB before gziped. This is because we import a lot of unnecessary dependency.
  
**Summary**:
- Improve import for `lodash`
- Improve import for `@kubernetes/client-node`

Improve by 30% the size of the build.

**Acceptance criteria**: 
When you do a `npm run build`, you have something like: 

```
File sizes after gzip:

  327.41 KB  build/static/js/2.9128ae09.chunk.js
  24.09 KB   build/static/js/main.df19a729.chunk.js
  14.65 KB   build/static/css/2.98cea2f2.chunk.css
  787 B      build/static/js/runtime~main.e26c4261.js
  746 B      build/static/css/main.f9c7b7f2.chunk.css
```

The biggest size should be around 330KB instead of 450KB
And the original file size in `/build/static/js` should be around 2MB instead of 3MB.

---

Closes: #1618


ui/src/services/k8s/api.js
```diff
@@ -1,11 +1,9 @@
 import ApiClient from '../ApiClient';
 
-import {
-  Config,
-  CoreV1Api,
-  CustomObjectsApi,
-  StorageV1Api
-} from '@kubernetes/client-node';
+import { Config } from '@kubernetes/client-node/dist/config';
```
NicolasT: Given this, you'll want to update the title of this PR.

NicolasT: Shouldn't this be `@kubernetes/client-node/dist/browser/config`? Or will the replacements defined in said package's `package.json` also handle this case?

NicolasT (changes requested):
Could we, given this, only depend on `lodash.isempty` and `lodash.sortby` instead of all of `lodash`? Or is that an indirect dependency anyway?

If not, given the two security-related PRs we had recently due to bugs in `lodash`, limiting how much of `lodash` we depend on may be worth that effort.

ui/src/services/k8s/api.js
MonPote: Updated :)

MonPote: > Could we, given this, only depend on `lodash.isempty` and `lodash.sortby` instead of all of `lodash`? Or is that an indirect dependency anyway?
> 
> If not, given the two security-related PRs we had recently due to bugs in `lodash`, limiting how much of `lodash` we depend on may be worth that effort.

Done :)

MonPote: > Shouldn't this be `@kubernetes/client-node/dist/browser/config`? Or will the replacements defined in said package's `package.json` also handle this case?

~~I just tested both, we can use this one. Actually, if we are not using this specific import, I don't understand why we forked `@kubernetes/client-node`. The project seems to work on both and we are not using `browser` stuff from the beginning I think (or we thought we are using it ?)~~
Any ideas @carlito-scality ?
Forget it Carlito explain me that the `package.json` do the trick to redirect the import.

```
  "browser": {
    "request": "./dist/browser/request-fetch.js",
    "./dist/config.js": "./dist/browser/config.js",
    "./dist/watch.js": "./dist/browser/watch.js"
  },
```

NicolasT: Exactly, so that's why the *non-browser-version* also worked from the start, most likely.

It *is* a kludge. I'd love to, at some point, work with upstream to make their library 'cross-platform' out of the box, both NodeJS as well as browser.

NicolasT (approved):

MonPote:
@bert-e approve

Issue #1558, opened 2019-08-24
On-disk encryption of secrets etc.


NicolasT:
Kubernetes has built-in support for encryption of data being stored in `etcd`, which would otherwise end up in plaintext on-disk.

Let's make sure to set this up.

jbertran:
First iteration plans:

* On-disk key only - a KMS provider can be a second step
* `aescbc` since it's the most conservative choice
* Enable encryption via salt (i.e. create encryption key, encrypt and replace all secrets)
* Disable encryption via salt (decrypt and replace all secrets, remove encryption key)

We can either wait for a KMS provider and set up rotation when we get that, set up key rotation with an on-disk key now, or set up key rotation with an on-disk key later. I'm not sure what the scope of this ticket should be in that regard.

@NicolasT does that match with your expectations?

NicolasT:
- Chances are slim we'd ever need/want KMS
- I'd start with implementing `identity` such that all configuration is in place
- Then indeed set up `aescbc`
- The key is to be generated during bootstrap, but should obviously not be recreated when re-applying some state...

So,

> Enable encryption via salt (i.e. create encryption key, encrypt and replace all secrets)

We only support encrypted data. One single key generated during bootstrap. No need to 'encrypt and replace'. We may implement key rotation later, if need be.

> Disable encryption via salt (decrypt and replace all secrets, remove encryption key)

Nope. Why would we want this?

No need for any kind of rotation, for now. This 'feature' is merely there such that a disk which contains the `etcd` database doens't contain any plain-text secret data. As such, if `/etc` and `/var/lib/etcd` sit on the same disk, no level of security is provided anyway. *Rotation* is not much of an issue either (at least for now): if one is able to retrieve the encryption key, one can likely at the same point in time read the `etcd` data.

This 'feature' is merely about: if someone deploys MetalK8s (with `/var/lib/etcd` on a separate device), then throws away that disk (only) without erasing it, an 'attacker' shouldn't be able to read secrets from said database.

So, KIS:
- Deploy config file, `identity` 'encryption'
- Deploy config file but with `aescbc` enabled, using a deployment-time-generated key
- Ensure this config (including the key) ends up on all API server nodes (i.e., may need to go in the `metalk8s_private` pillar?)
- Ensure that re-applying the deployment state (the one which initially would generate a new key) will not regenerate the config with a *different* key (hence, this may require a custom state/module to manage this configuration file properly)

NicolasT:
Keep in mind
[...]

PR #1633, opened 2019-09-06
Improvement/1558 encrypt k8s secrets on disk

jbertran:
**Component**: salt/apiserver

**Context**:

Currently the K8s `Secret` objects are saved as plain text in etcd, which is perfectible™.

**Summary**:

* Set up an encryption configuration state pulled by the `kube-apiserver` installation state, which is itself pulled by the role setup.
* Use the private pillar to ensure that the encryption key is generated once and only once, and then replicated across the other nodes running `kube-apiserver`.
* Make `kube-apiserver` use that encryption configuration by mounting it into the container and adding the relevant flag to its launch options.

**Acceptance criteria**: 

1. Get a cluster up with this branch, go into any machine of the cluster and switch to root
2. Create a secret: `kubectl create secret generic secret1 -n default --from-literal=mykey=mydata`
3. Verify the secret is legible: `kubectl get secret secret1 -o yaml` - the data is encoded in base64.
4. Verify the secret is encrypted in etcd:
     a. exec into the etcd container: `kubectl -ti etcd-bootstrap sh -n kube-system --kubeconfig /etc/kubernetes/admin.conf`
     b. dump the secret: `ETCDCTL_API=3 etcdctl get /registry/secrets/default/secret1 --cert=/etc/kubernetes/pki/etcd/server.crt --key=/etc/kubernetes/pki/etcd/server.key --cacert=/etc/kubernetes/pki/etcd/ca.crt | hexdump -C`
     c. The data should be illegible, and should be prefixed by `:enc:aescbc:v1`

---

Closes: #1558

### DIFF
--- a/salt/_pillar/metalk8s_private.py
+++ b/salt/_pillar/metalk8s_private.py
@@ -3,6 +3,7 @@ import logging
 
 
 SA_PRIVATE_KEY_PATH = "/etc/kubernetes/pki/sa.key"
+APISERVER_KEY_PATH = "/etc/metalk8s/crypt/apiserver.key"
 
 log = logging.getLogger(__name__)
 
@@ -28,6 +29,23 @@ def _read_sa_private_key():
     return {"sa_private_key": key_data}
 
 
+def _read_apiserver_key():
+    # FIXME: we only have access to this file because:
+    #   1- Salt master Pod mounts all of /etc/metalk8s
+    #   2- Salt master Pod runs on the 'bootstrap' minion, which is, by chance,
+    #      also the initial master node
+
+    if not os.path.isfile(APISERVER_KEY_PATH):
+        return {"_errors": [
+            "Missing secrets encryption key to share with master node."
+        ]}
+
+    with open(APISERVER_KEY_PATH, "r") as config_file:
+        encryption_key = config_file.read()
+
+    return {"apiserver_key": encryption_key}
+
+
 def ext_pillar(minion_id, pillar):
     nodes_info = pillar.get("metalk8s", {}).get("nodes", {})
 
@@ -42,7 +60,10 @@ def ext_pillar(minion_id, pillar):
     private_data = {'private': None}
 
     if "master" in node_info["roles"]:
-        private_data['private'] = _read_sa_private_key()
+        data = {}
+        data.update(_read_sa_private_key())
+        data.update(_read_apiserver_key())
+        private_data['private'] = data
         __utils__['pillar_utils.promote_errors'](private_data, 'private')
 
     result = {"metalk8s": private_data}
--- a/salt/_pillar/metalk8s_private.py
+++ b/salt/_pillar/metalk8s_private.py
@@ -1,4 +1,4 @@
-import os.path
+import errno
 import logging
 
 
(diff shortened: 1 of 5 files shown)


salt/_pillar/metalk8s_private.py
```diff
@@ -1,8 +1,10 @@
 import os.path
 import logging
+import yaml
 
 
 SA_PRIVATE_KEY_PATH = "/etc/kubernetes/pki/sa.key"
+APISERVER_KEY_PATH = "/etc/metalk8s/apiserver.key"
```
charlesprost:
```suggestion
APISERVER_KEY_PATH = "/etc/metalk8s/crypt/apiserver.key"
```
So we have a clear distinction between PKI stuff and encryption stuff

salt/_pillar/metalk8s_private.py
```diff
@@ -28,6 +30,23 @@ def _read_sa_private_key():
     return {"sa_private_key": key_data}
 
 
+def _read_apiserver_key():
+    # FIXME: we only have access to this file because:
+    #   1- Salt master Pod mounts all of /etc/kubernetes
```
charlesprost: You are talking about `/etc/kubernetes` but the key file is stored under `/etc/metalk8s`.

salt/_pillar/metalk8s_private.py
```diff
@@ -28,6 +30,23 @@ def _read_sa_private_key():
...
+    #   2- Salt master Pod runs on the 'bootstrap' minion, which is, by chance,
+    #      also the initial master node
+
+    if not os.path.isfile(APISERVER_KEY_PATH):
```
slaperche-scality: TOCTOU:, use a try/except (you can have two except, one for not found and one for the others) below instead.

slaperche-scality (changes requested):

salt/_pillar/metalk8s_private.py
jbertran: Oof, indeed. The remark still holds, edited the path :)

salt/_pillar/metalk8s_private.py
jbertran: I was unsure about the path, good suggestion, done.

salt/_pillar/metalk8s_private.py
jbertran: cargo culting is bad, mmmkay?

salt/_pillar/metalk8s_private.py
```diff
@@ -35,14 +36,14 @@ def _read_apiserver_key():
     #   2- Salt master Pod runs on the 'bootstrap' minion, which is, by chance,
     #      also the initial master node
 
-    if not os.path.isfile(APISERVER_KEY_PATH):
+    try:
```
slaperche-scality: Actually, the real equivalent would be:

```python
    try:
        with open(APISERVER_KEY_PATH, "r") as config_file:
            encryption_key = config_file.read()
    except FileNotFoundError:
        return {"_errors": [
            "Missing secrets encryption key to share with master node."
        ]}
```

Just catch file missing with a proper errors, and let other exceptions bubble up.

slaperche-scality: Meh, no `FileNotFoundError` in Python 2, so it would be more like (need `import errno`):

```python
    try:
        with open(APISERVER_KEY_PATH, "r") as config_file:
            encryption_key = config_file.read()
    except IOError as exn:
        if exn.errno == errno.ENOENT:
            return {"_errors": [
                "Missing secrets encryption key to share with master node."
            ]}
```

salt/_pillar/metalk8s_private.py
```diff
@@ -14,36 +14,39 @@ def __virtual__():
     return __virtualname__
 
 
+def _read_private_key(key_name, key_path):
+    try:
+        with open(key_path, "r") as key_file:
+            key_data = key_file.read()
+    except IOError as exn:
+        # File not found
```
slaperche-scality: IMHO, `ENOENT` is clear enough (that's the point of importing `errno` to use the named constant instead of `2`)

salt/_pillar/metalk8s_private.py
```diff
@@ -14,36 +14,39 @@ def __virtual__():
...
+        if exn.errno == errno.ENOENT:
+            err_msg = "Missing {0} key file to share with master node at {1}."
+            return {"_errors": [
+                err_msg.format(key_name, key_path)
+            ]}
+        # Other failures to read the file
+        err_msg = "Unable to read {0} key file contents at {1}."
+        return {"_errors": [err_msg.format(key_name, key_path)]}
```
slaperche-scality: could be good to have the exception message somewhere here.

salt/_pillar/metalk8s_private.py
```diff
@@ -14,36 +14,39 @@ def __virtual__():
     return __virtualname__
 
 
+def _read_private_key(key_name, key_path):
```
slaperche-scality, line 17: Could explain in the commit message why this reword (TOCTOU, obvious in the context of this review but not when finding the commit alone with a git blame)

slaperche-scality (changes requested):

salt/_pillar/metalk8s_private.py
jbertran: Done

slaperche-scality (approved):

charlesprost (approved):

PR #1635, opened 2019-09-06
Create a function to download Debian packages

### LINKED ISSUE #1616: Create a script to download Debian packages from official repositories
**Component**:packages, build

**Why this is needed**: We need to download some Debian packages which will be used to deploy Ubuntu nodes during the cluster deployment.

**What should be done**: create a script in the Debian entrypoint Implement a solution to download Debian packages from official repositories with all dependencies required to install these packages offline.

- [x] List all the packages required by a cluster node on Ubuntu
- [x] Find a tool to download the Debian packages and their dependencies from repositories
- [x] Add all required repositories to download packages

**Test plan**: build the MetalK8s iso and verify that you have all the required packages

ycointe:
**Component**: build, packages

**Summary**: We need a function in the Debian entrypoint to download Debian packages from official repositories

**Acceptance criteria**: 
We need to have the required Debian packages in the iso.

---

Closes: #1616

### DIFF
--- a/packages/debian/Dockerfile
+++ b/packages/debian/Dockerfile
@@ -11,6 +11,7 @@ RUN apt update -y \
         devscripts \
         build-essential \
         lintian \
+        apt-rdepends \
         alien
 
 RUN useradd -m build
--- a/packages/debian/entrypoint.sh
+++ b/packages/debian/entrypoint.sh
@@ -32,20 +32,54 @@ rpm2deb() {
     pushd /debbuild/result 1> /dev/null
     alien --bump=0 /rpmbuild/source.rpm
 
-    chown "${TARGET_UID}:${TARGET_GID}" * 
+    chown "${TARGET_UID}:${TARGET_GID}" *
 }
 
-case ${1:- } in
-    builddeb)
-        builddeb
-        ;;
-    rpm2deb)
-        rpm2deb
-        ;;
-    '')
-        exec /bin/bash
-        ;;
-    *)
-        exec "$@"
-        ;;
-esac
+
+download_packages() {
+    set -x
+    local -a list_pkg=("$@")
+
+    # Salt-minion GPG Key
+    curl -s https://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3/SALTSTACK-GPG-KEY.pub | apt-key add -
+    echo 'deb http://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3 bionic main' >> /etc/apt/sources.list.d/saltstack.list
+
+    # Kubectl & Kubelet GPG key
+    echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" >>/etc/apt/sources.list.d/kubernetes.list
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+    apt update
+
+    for pkg in $(list_pkg)
+    do
+        echo "$pkg" >> packages_list.txt
+        for dep in $(apt-rdepends "$pkg" | grep -E 'Depends' | cut -d " " -f4)
+            do echo "$dep" >> packages_list.txt
+        done
+    done
+
+    mkdir /packages/pkg_downloaded
+    pushd /packages/pkg_downloaded
+    while IFS=$'\n' read -r pkg2dl
+    do
+    	apt-get download "$pkg2dl"
+    done
+    chown "${TARGET_UID}:${TARGET_GID}" *
+}
+
+    case ${1:- } in
+        builddeb)
+            builddeb
+            ;;
+        rpm2deb)
+            rpm2deb
+            ;;
+        download_packages)
+            download_packages
+            ;;
+        '')
+            exec /bin/bash
+            ;;
+        *)
+            exec "$@"
+            ;;
+    esac


packages/debian/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apt update -y \
         devscripts \
         build-essential \
         lintian \
+	apt-rdepends \
```
slaperche-scality: wrong indentation (I bet your editor added a tab, whereas the other lines are using spaces)

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,22 @@ rpm2deb() {
     chown "${TARGET_UID}:${TARGET_GID}" * 
 }
 
+
+download_packages() {
+#<Package_name>=<Package_version>
+for pkg in $(cat list.txt)
+do
+        echo "$pkg" >> packages_list.txt
+        for dep in $(apt-rdepends "$pkg" | grep -E 'Depends|Recommends|Suggests' | cut -d " " -f4)
```
slaperche-scality: I don't think we want the `Suggests`, and maybe not the `Recommends`.

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,22 @@ rpm2deb() {
...
+                do echo "$dep" >> packages_list.txt
+        done
+done
+for todl in $(sort -u packages_list.txt)
```
slaperche-scality: I'm not sure that's the way to read a file line by line, my dusted memories about Bash reminds me of something like

```sh
while IFS= read -r line
do
  echo "$line"
done < "packages_list.txt"
```

TBC.

slaperche-scality (commented):
first quick review.

packages/debian/entrypoint.sh
ycointe: This is just temporary names waiting for a convention naming for these parameters

packages/debian/Dockerfile
ycointe: @jbertran made me the same remark but there is no tindentation issue in the file in my computeur and neither on the files on github (see clicking on `view file`).

packages/debian/entrypoint.sh
slaperche-scality: I'm not talking about naming of variable here.
I'm talking about the deps we want to download.

packages/debian/Dockerfile
slaperche-scality: Likely because we indent with 4 spaces and by default tabs are displayed with a 4-spaces width so visually it may look OK.

But it's still bad to mix tab and spaces, and this will be quite ugly if you editor use a custom width for the tabs and/or display tab differently from spaces.

ycointe: Indeed, I did a tabulation and not spaces.
It's corrected !

packages/debian/entrypoint.sh
ycointe: My bad, I misunderstood.
I corrected it !

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,53 @@ rpm2deb() {
...
+
+    # Salt-minion GPG Key
+    curl -s https://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3/SALTSTACK-GPG-KEY.pub | apt-key add -
+    echo 'deb http://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3 bionic main' >> /etc/apt/sources.list.d/saltstack.list
+
+    # Kubectl & Kubelet GPG key
+    echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" >>/etc/apt/sources.list.d/kubernetes.list
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+
+    mkdir /packages/pkg_downloaded
```
charlesprost: You can create it later when you will be using it

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,53 @@ rpm2deb() {
...
+    do
+        echo "$pkg" >> packages_list.txt
+        for dep in $(apt-rdepends "$pkg" | grep -E 'Depends' | cut -d " " -f4)
+            do echo "$dep" >> packages_list.txt
+        done
+    done
+
+    pushd /packages/pkg_downloaded
+    for pkg2dl in $(sort -u /packages_list.txt)
+    #Do not copy this way to itterate packages, it's work because there is no space
```
charlesprost:
```suggestion
    #Do not copy this way to iterate on packages, it's working because there is no space in the names
```

charlesprost: Are we sure this is a very reliable way of doing it? Because as you are saying, if we had a name with a space, it could break

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,53 @@ rpm2deb() {
...
-        exec /bin/bash
-        ;;
-    *)
-        exec "$@"
-        ;;
-esac
+
+download_packages() {
+
+    local -a list_pkg("$@")
```
jbertran: Missing `=` here

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,53 @@ rpm2deb() {
...
+    for pkg in $(list_pkg)
+    do
+        echo "$pkg" >> packages_list.txt
+        for dep in $(apt-rdepends "$pkg" | grep -E 'Depends' | cut -d " " -f4)
+            do echo "$dep" >> packages_list.txt
+        done
+    done
+
+    pushd /packages/pkg_downloaded
+    for pkg2dl in $(sort -u /packages_list.txt)
```
jbertran: `pkg_to_dl` please - let's not make this more cryptic than we have to :)

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,53 @@ rpm2deb() {
...
+    #Do not copy this way to itterate packages, it's work because there is no space
+    do
+    	apt-get download "$pkg2dl"
+    done
+}
```
jbertran, line 67: In order to have the right permissions for everything on the host FS, let's `chown` these artifacts to the right UID/GID.

ycointe: Done, thank you for the reminder :)

ycointe:
@bert-e approve

packages/debian/entrypoint.sh
```diff
@@ -32,20 +32,54 @@ rpm2deb() {
...
+download_packages() {
+    set -x
+    local -a list_pkg=("$@")
+
+    # Salt-minion GPG Key
+    curl -s https://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3/SALTSTACK-GPG-KEY.pub | apt-key add -
+    echo 'deb http://repo.saltstack.com/apt/ubuntu/18.04/amd64/2018.3 bionic main' >> /etc/apt/sources.list.d/saltstack.list
+
+    # Kubectl & Kubelet GPG key
+    echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" >>/etc/apt/sources.list.d/kubernetes.list
```
charlesprost, line 48:
```suggestion
    echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" >> /etc/apt/sources.list.d/kubernetes.list
```

charlesprost: Just for consistency sake

charlesprost (approved):

PR #1638, opened 2019-09-09
Allow to use a node name different from the hostname

### LINKED ISSUE #1391: bootstrap node deployment or node expansion fails if node name does not equals to hostname
**Component**:

'salt', 'ui'

**What happened**:

bootstrap node deployment or node expansion fails if node name does not equal to hostname
This is true when you fill bootstrap.yaml (when deploying bootstrap node)
It is also trie when filling UI form when adding a node

**What was expected**:

Node name is what will appear in k8s node resource and it could be different from hostname
Hostname (or ip) is just a way to instruct salt and our tooling on how to reach a specific node

**Steps to reproduce**

Deploy the bootstrap node and use random name (!= from hostname) ...
Deployment will fail
**UPDATED**
**Why is this happening**
both `kubelet` and `salt-minion` have their internal algorithm to determine which hostname to use, if these hostnames differ from the one we set in `bootstrap.yaml` file then everything will break down
We need to force both `kubelet` and `salt-minion` to use the same hostname, to do that:
- **kubelet** has an option `--override-hostname` to set it
- **salt-minion** reads a file `minion_id` that we can put the hostname in it

slaperche-scality:
**Component**:
salt

**Context**: 

We want to be able to choose a node name different from the node hostname.

**Summary**:

The salt-minion configuration already use `grains.id` so nothing to do on this side.
As for kubelet, we specify the `--hostname-override` option.

**Acceptance criteria**: 

Try to create a new node with a name that doesn't match the hostname of the node: it should works.

---

Closes: #1391


charlesprost (dismissed):

TeddyAndrieux (approved):

salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -14,8 +14,13 @@ Create kubelet service environment file:
...
     - dir_mode: 750
     - context:
-        options: {{ kubelet.service.options }}
-        node_ip: {{ grains['metalk8s']['control_plane_ip'] }}
+    - context:
+        options:
+      {%- for opt, value in kubelet.service.options.items() %}
+          {{ opt }}: {{ value }}
+      {%- endfor %}
+          node_ip: {{ grains['metalk8s']['control_plane_ip'] }}
```
NicolasT: Is it `node_ip`? Before, it was `node-ip`...

TeddyAndrieux: Oh yes agree

TeddyAndrieux (changes requested):

salt/metalk8s/kubernetes/kubelet/standalone.sls
slaperche-scality: Good catch!

Seems like it didn't broke the expansion (in a very visible way at least);

TeddyAndrieux (approved):

NicolasT:
Argh, forgot! This should've gone in `development/2.0` as well of course @slaperche-scality...

slaperche-scality:
> Argh, forgot! This should've gone in `development/2.0` as well of course @slaperche-scality...

Ha this one also?
Isn't it about ease of use/UX only?

Or does the fact that we touch `/var/lib/kubelet/kubeadm-flags.env` will make upgrade manual and thus we must have it from the start?

NicolasT:
Yes, rather have it all 'correct' from the start. This is not 'a new feature', this is really needed to have the nodes function 'properly'.

It may not be *strictly* required, but backporting will likely make things somewhat less error-prone.

In general, new features: 2.4. Things that touch 'core' K8s cluster/nodes: 2.0 and up.

slaperche-scality:
Will backport then, this one will be trivial (and will be more careful next time ^^)

Issue #1115, opened 2019-05-15
TLS for SaltAPI


NicolasT:
We currently deploy SaltAPI as a plain HTTP service. This should, however, be an HTTPS service to ensure secrecy of authentication, as well as data transmitted between Salt and the client.

It likely makes sense to use a certificate signed by the 'API server' root CA for this purpose.

slaperche-scality:
So, to clarify a bit what we need here is:
- to have a new SLS in `salt/metalk8s/kubernetes/apiserver/certs/` to create a certs for salt-api
- update the `rest_cherrypy` section in the Salt master config file to use those certs
- update the code calling the Salt API to use HTTPS (storage-operator, UI also maybe?, …)

Did I miss something?

TeddyAndrieux:
For me the SLS for certificate should be in `salt/metalk8s/salt/master/certs/` as it's a cert for salt-api

NicolasT:
> to have a new SLS in salt/metalk8s/kubernetes/apiserver/certs/ to create a certs for salt-api

Not sure that's the right SLS, because `salt-api` is deployed elsewhere. It may also require us to create the K8s API server CA earlier (i.e., before we actually deploy K8s API server, because `salt-master` comes really early)

> update the code calling the Salt API to use HTTPS (storage-operator, UI also maybe?, …)

UI should use HTTPS but you could run into cert acceptance issues (still self-signed...). For the operator, yes indeed, and you can even let it validate the provided cert because we have the K8s API server root CA in `/var/run/kubernetes/....`.

sayf-eddine-scality:
LGTM, As per saltstack docs check CherryPy version https://github.com/cherrypy/cherrypy/issues/1298 and that PyOpenSSL is insatlled

slaperche-scality:
@NicolasT 

> For the operator, yes indeed, and you can even let it validate the provided cert because we have the K8s API server root CA in /var/run/kubernetes/....

Do I need to update the Operator deployment for that? Or it's already mounted/somehow available?

Also, what about running the operator outside of the cluster?
Do I need to document a new step (copy the ca.crt locally and give the path in a env variable)?

NicolasT:
> @NicolasT
> 
> > For the operator, yes indeed, and you can even let it validate the provided cert because we have the K8s API server root CA in /var/run/kubernetes/....
> 
> Do I need to update the Operator deployment for that? Or it's already mounted/somehow available?

[...]

PR #1641, opened 2019-09-09
Backport PR 1631 to 2.0

slaperche-scality:
**Component**:

salt

**Context**: 

Since HTTPS is a security feature, we want it in 2.0 (and also to ease the migration).

**Summary**:

Thanks to atomic commits, it's only a matter of cherry-picking:
- 5e29246e, 09a7fad1, fbbab308 were left out because UI doesn't exist (was removed in fact) before 2.4
- 428ea9b2 was left out because the storage operator was introduced in 2.4

All the other commit were cherry-picked as-is (just a small conflict resolution for 96416a50).

**Acceptance criteria**: 

See https://github.com/scality/metalk8s/pull/1631

---

See: #1115


NicolasT:
LGTM on first sight.

slaperche-scality:
No diff in https://github.com/scality/metalk8s/pull/1645, as expected/as it should be.

charlesprost (approved):

slaperche-scality:
Now we have a diff for #1645 (and it's expected).

The test with the ping was buggy in 2.4 to begin with, but went unnoticed because my new test "auth with token" shadowed/replaced the "auth with Basic auth"…

I'll do a fix in 2.4 once this PR is merged (could have done it here, but I'm not fond of fixing bug during merge commit, bad for tracking)

Issue #780, opened 2019-03-15
Remove hard-coded service IPs


NicolasT:
In several places service IPs like `10.96.0.10` for cluster DNS are hard-coded. This should be calculated from `BootstrapConfiguration` (or the default service network), and then either stored as a static value in the pillar as well, or stored in the mine, TBD

NicolasT:
This is a release-blocker for https://github.com/scality/metalk8s/milestone/15

NicolasT:
Some clarification: we currently *assume* the `serviceNetwork` is `10.96.0.0/12`:
```
$ git grep '10\.96'
docs/developer/architecture/deployment.rst:  Default: ``10.96.0.0/12``
docs/quickstart/introduction.rst:- A network for :term:`Services <Service>`, defaulting to ``10.96.0.0/12``
salt/_pillar/metalk8s.py:DEFAULT_SERVICE_NETWORK = '10.96.0.0/12'
salt/metalk8s/defaults.yaml:  service_ip: "10.96.0.1"
salt/metalk8s/defaults.yaml:  cluster_dns_ip: 10.96.0.10
salt/metalk8s/kubernetes/kubelet/standalone.sls:          - 10.96.0.10
```

However, the `serviceNetwork` can be configured in `BootstrapConfiguration`, so those addresses would be incorrect if the default `serviceNetwork` is not used.

Hence, we should properly calculate these addresses from the (configured or default, but always available in Pillar) `service_network`, and use them accordingly. We could, in line with what's there now, decide that

- the `kubernetes` *ClusterIP* address is the second address in the `service_network`. I.e., if the `service_network` is `10.96.0.0/12`, then the `kubernetes` *ClusterIP* is `10.96.0.1`
- the `coredns` *ClusterIP* address is the 11th address in the `service_network`. I.e., if the `service_network` is `10.96.0.0/12`, then the `coredns` *CusterIP* is `10.96.0.10`.

Not sure where that calculation should happen. In any case, the 'default' in `salt/metalk8s/default.yaml` has to be gone because we *can't* have a non-parametrized default.

When calculating the address, we must also assert the result *is within the `service_network`*. As an example, if `service_network` is `192.168.0.0/29` we would calculate the `coredns` *ClusterIP* as `192.168.0.10`, but **that's not within the `service_network`**. In that case, we should error out during deployment/SLS rendering/whatever (a `/29` `service_network` is way too small anyway).

Ebaneck:
Todo:

[...]

PR #1651, opened 2019-09-10
Remove hard coded service ips for coredns and kubernetes service 

Ebaneck:
**Component**:

'salt', 'kubernetes', 'networking'

**Context**: 

See #780 

**Summary**:

- We need to remove the hard-coded values related to the Coredns cluster IP and Kubernetes cluster IP.
For this, we have written a salt execution module to pick the following:
- The `kubernetes ClusterIP` address which is agreed to be the 2nd address in the service_network.
- The `coredns ClusterIP` address which is the 10th address in the service_network
 For this to work properly, we need to define a proper service network within the pillars

How to test?

```
[root@bootstrap vagrant]# kubectl get pods,svc --namespace=kube-system | grep dns

pod/coredns-69bb6dc74d-48glm                 0/1       ContainerCreating   0          1m
pod/coredns-69bb6dc74d-4tqtj                 0/1       ContainerCreating   0          1m
service/coredns                                        ClusterIP   10.96.0.10       <none>        53/UDP,53/TCP,9153/TCP       1m
```
**Acceptance criteria**: 

- Working metalk8s cluster after bootstrap
- CI build should still pass.

Needed:

Once this ticket is closed, I plan to open another to update the documentation.
---

Closes: #780

### DIFF
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -1,0 +1,70 @@
+'''Metalk8s network related utilities.'''
+
+import itertools
+import logging
+
+from salt._compat import ipaddress
+from salt.exceptions import CommandExecutionError
+
+K8S_CLUSTER_ADDRESS_NUMBER = 0
+COREDNS_ADDRESS_NUMBER = 9
+
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_network'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def _pick_nth_service_ip(n):
+    '''
+    Pick the nth IP address from a network range, based on pillar
+    configurations provided in CIDR notation.
+    '''
+    cidr = __pillar__.get('networks', {}).get('service')
+    if cidr is None:
+        raise CommandExecutionError(
+            'Pillar key "networks:service" must be set.'
+        )
+
+    network = ipaddress.IPv4Network(unicode(cidr))
+    # NOTE: hosts() method below returns usable hosts in a network.
+    # The usable hosts are all the IP addresses that belong to the network,
+    # except the network address itself and the network broadcast address.
+    ip = next(
+        itertools.islice(network.hosts(), n, None),
+        None
+    )
+    if ip is None:
+        raise CommandExecutionError(
+            'Could not obtain an IP in the network range {}'.format(
+                cidr
+            )
+        )
+    return str(ip)
+
+
+def get_kubernetes_service_ip():
+    '''
+    Return the Kubernetes service cluster IP.
+
+    This IP is arbitrarily selected as the first IP from the usable hosts
+    range.
+    Note:
+    In kube-apiserver Pod manifest, we define a service-cluster-ip-range which
+    sets the kube-api address to the first usable host.
+    '''
+    return _pick_nth_service_ip(K8S_CLUSTER_ADDRESS_NUMBER)
+
+
+def get_cluster_dns_ip():
+    '''
+    Return the CoreDNS cluster IP.
+
+    This IP is arbitrarily selected as the tenth IP from the usable hosts
+    range.
+    '''
+    return _pick_nth_service_ip(COREDNS_ADDRESS_NUMBER)
--- a/salt/metalk8s/defaults.yaml
+++ b/salt/metalk8s/defaults.yaml
@@ -113,7 +113,6 @@ kube_api:
   cert:
     server_signing_policy: kube_apiserver_server_policy
     client_signing_policy: kube_apiserver_client_policy
-  service_ip: "10.96.0.1"
 
 etcd:
   ca:
@@ -137,7 +136,6 @@ front_proxy:
     client_signing_policy: front_proxy_client_policy
 
 coredns:
-  cluster_dns_ip: 10.96.0.10
   cluster_domain: cluster.local
   reverse_cidrs: in-addr.arpa ip6.arpa
 
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -402,6 +402,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/_modules/metalk8s_etcd.py'),
     Path('salt/_modules/metalk8s_kubernetes_utils.py'),
     Path('salt/_modules/metalk8s.py'),
+    Path('salt/_modules/metalk8s_network.py'),
     Path('salt/_modules/metalk8s_package_manager.py'),
 
 
--- a/salt/metalk8s/kubernetes/apiserver/certs/server.sls
+++ b/salt/metalk8s/kubernetes/apiserver/certs/server.sls
@@ -1,5 +1,7 @@
 {%- from "metalk8s/map.jinja" import kube_api with context %}
 
+{%- set kubernetes_service_ip = salt.metalk8s_network.get_kubernetes_service_ip() %}
+
 include:
   - metalk8s.internal.m2crypto
 
@@ -22,7 +24,7 @@ Create kube-apiserver private key:
     'kubernetes.default',
     'kubernetes.default.svc',
     'kubernetes.default.svc.cluster.local',
-    kube_api.service_ip,
+    kubernetes_service_ip,
     grains['metalk8s']['control_plane_ip'],
     pillar['metalk8s']['api_server']['host'],
 ]
--- a/salt/metalk8s/kubernetes/cni/calico/configured.sls
+++ b/salt/metalk8s/kubernetes/cni/calico/configured.sls
@@ -1,6 +1,8 @@
 {%- from "metalk8s/map.jinja" import kube_api with context %}
 {%- from "metalk8s/map.jinja" import kubernetes with context %}
 
+{%- set kubernetes_service_ip = salt.metalk8s_network.get_kubernetes_service_ip() %}
+
 include:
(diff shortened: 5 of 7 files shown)


Ebaneck:
This is a WIP, I am going to ask for reviewers when a fully functional draft is ready.

slaperche-scality:
You should target your PR on 2.0, not 2.4 :wink:

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,45 @@
+'''Metalk8s network related utilities.'''
+
+import ipaddress
```
slaperche-scality: `New in version 3.3.`

I thought we were stuck to Python 2 for Salt (see this [reply](https://github.com/scality/metalk8s/issues/780#issuecomment-528441446)) ?
Or are we somehow magically pulling the backport to 2.X from pip or something?

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,45 @@
...
+
+__virtualname__ = 'metalk8s_network'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def pick_nth_ip(service_network, index):
+
```
slaperche-scality: this empty line should be removed.

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,45 @@
...
+    :param service_network (str): CIDR presentation format read from pillars.
+     For example, '192.168.1.0/24'.
+    :param index (int): Kubernetes cluster IP address is set to 2nd address in
+     service_network.
+     coredns cluster IP address is set to the 10th address in the
+     service network
+
+    :returns (str):
+        An index param of 10 returns => coredns cluster ip
```
slaperche-scality: What is `an index param`?
Also, what the heck does 2 and 10 means in this context?

We should have named constant or something, rather than relying on those some magical values?

And/or have two public functions `get_service_ip` and `get_dns_ip` and that will call that one.

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def pick_nth_ip(service_network, index):
+
+    '''
+    Pick the coredns and kubernetes cluster ip address from service network.
```
slaperche-scality: use proper capitalization: `CoreDNS`, `Kubernetes`, `IP address`.

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,45 @@
...
+     service network
+
+    :returns (str):
+        An index param of 10 returns => coredns cluster ip
+        An index param of 2 returns => kubernetes cluster ip
+    '''
+    log.info('Picking the nth address from service network')
+
+    network = ipaddress.IPv4Network(unicode(service_network))
+    for i, ip in enumerate(network):
```
slaperche-scality: could be simplified as:

```python
ip = next(itertools.islice(network.hosts(), index, None), None)
if ip is None:
    raise CommandExecutionError(
        'Could not obtain an IP in the network range {}'.format(
            service_network
        )
    )
return str(ip)
```

salt/metalk8s/defaults.yaml
```diff
@@ -113,7 +113,6 @@ kube_api:
   cert:
     server_signing_policy: kube_apiserver_server_policy
     client_signing_policy: kube_apiserver_client_policy
```
slaperche-scality, line 115: Could be squashed with the previous commit, no?

slaperche-scality (changes requested):

salt/_modules/metalk8s_network.py
Ebaneck: Looks like the module is available by default in python2.

Here: 
```
[root@bootstrap vagrant]# python
Python 2.7.5 (default, Oct 30 2018, 23:45:53) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
>>> import ipaddress
>>> 
```

salt/_modules/metalk8s_network.py
Ebaneck: I think we had clearly defined what an index `param` signifies. no??

Also FYI:
```
 :param index (int): Kubernetes cluster IP address is set to 2nd address in
     service_network.
     coredns cluster IP address is set to the 10th address in the
     service network
```
As such index=10 => we are computing the ip address for cluster DNS

I do not think it would be any better creating two functions that perform the same actions and so this is the reason we have one generic function doing the same action.

Would mark this as resolved unless another thinks otherwise

salt/_modules/metalk8s_network.py
Ebaneck: ok

salt/metalk8s/defaults.yaml
Ebaneck: It could be.

But this still looks okay to me.
one commit => add module + add module path to salt_tree.
one commit => use module in sls files while replacing variables were need be.
one commit => remove the hard coded service IP addresses within the `defaults.yaml`.

Looks ok to me

salt/_modules/metalk8s_network.py
Ebaneck: Maybe it should read `An index value of 10 returns a CoreDNS cluster IP`

salt/_modules/metalk8s_network.py
Ebaneck: - Adds one more import for `itertools` python module
- `network.hosts()` looks to return strictly the usable set of IP addresses meaning passing an index of 2 returns =>  a CoreDNS IP of 192.168.1.3.
This is not exactly what we want even though we can make a few changes else where to match this.

I prefer to stick to what is currently written. :)

salt/_modules/metalk8s_network.py
gdemonet: No it's not _available by default_, but actually comes from a backport as @slaperche-scality suggested: https://pypi.org/project/ipaddress/

However, this module is `Required-by` the `docker` client, which you probably have installed on your machine. We could add the dependency to the project (as we do with `m2crypto` for instance), or move back to relying on a more manual approach (using `__utils__` for instance).

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,44 @@
...
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_network'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def pick_nth_ip(service_network, index):
```
gdemonet: As @slaperche-scality suggested, it makes sense to define this as a _private_ helper function, and define specific methods for `get_kubernetes_service_ip` and `get_cluster_dns_ip`.

The rationale is simple: even if `pick_nth_ip` is an explicit name w.r.t what it does, it doesn't tell us **why** we want to pick the 10th for DNS, nor the first for the `kubernetes` service.

So, this module could instead look like:

```
# imports and other init stuff...

def _pick_nth_ip(network_range, n):
    """Pick the nth IP address from a network range, provided in CIDR notation."""
   # Same code...
   return ip

def get_cluster_dns_ip():
    """Return the CoreDNS cluster IP based on pillar configuration.

    This IP is arbitrarily selected as the tenth IP from the Service network range.
    """
    return _pick_nth_ip(__pillar__['networks.service'], 10)


def get_kubernetes_service_ip():
    """Return the Kubernetes service cluster IP based on pillar configuration.

    This IP is arbitrarily selected as the second IP from the Service network range.
    """
    return _pick_nth_ip(__pillar__['networks.service'], 2)
```

It yields that, in SLS files, no one will wonder "_why do we pick these values?!_", and instead will refer to these methods for understanding. You could have achieved the same result by adding a comment in those SLS files, but since these IPs are used several times, you gain the benefit of de-duplication.

salt/_modules/metalk8s_network.py
gdemonet: I'm not convinced it's a "simplification", I feel it's quite harder to read :/

However, since using `ipaddress` may not be the chosen strategy, it may not make sense to bother further with this...

salt/metalk8s/defaults.yaml
gdemonet: @Ebaneck I agree with Sylvain, one commit to introduce functionality (the module), one to use the functionality (replace values where need be AND remove now-unused values). If you keep the last commit as you suggest, its value is roughly the same as a `fix typo` commit, which we don't like much ;)

gdemonet (dismissed):

salt/_modules/metalk8s_network.py
```diff
@@ -0,0 +1,61 @@
...
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'metalk8s_network'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def get_service_ip(service_network):
```
gdemonet: You can factorize these with a helper method, called `_pick_nth_ip` ;)

salt/_modules/metalk8s_network.py
gdemonet: Now, Salt uses this as well, relying on their `salt._compat` layer: [see this import in the `salt.utils.network` module](https://github.com/saltstack/salt/blob/develop/salt/utils/network.py#L30). Let's do the same, should be sufficient:

```suggestion
from salt._compat import ipaddress
```

slaperche-scality (dismissed):

slaperche-scality (approved):

PR #1666, opened 2019-09-11
Replace etcd endpoints mine by ext pillar

### LINKED ISSUE #877: Use K8s API instead of Salt's mine to know etcd cluster members
This would be more reliable, for more info see Nicolas' comments:
- https://github.com/scality/metalk8s/issues/738#issuecomment-477364132
- https://github.com/scality/metalk8s/issues/738#issuecomment-477401597

sayf-eddine-scality:
Ext pillar is calculated each time it is accessed, so it makes
sense to use it to hold etcd endpoints rather than the occasionally
updated mine

Fixes: #877

**Component**:
salt, etcd

**Context**: 
etcd enpoints were set in the salt mine, but the latter is occasionally updated and will not hold the correct information always, this is an issue when in expansion the new etcd member does not have the correct information about the state of the existing cluster
**Summary**:
The etcd endpoints have been moved into the ext pillar
**Acceptance criteria**: 
node expansion should work perfectly, the new etcd instance should be in **existing** mode rather than **new** and the initial cluster should be correct

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -466,6 +466,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/_pillar/metalk8s_nodes.py'),
     Path('salt/_pillar/metalk8s_private.py'),
     Path('salt/_pillar/metalk8s_solutions.py'),
+    Path('salt/_pillar/metalk8s_etcd.py'),
 
     Path('salt/_renderers/metalk8s_kubernetes.py'),
 
--- a/pillar/metalk8s/roles/etcd.sls
+++ b/pillar/metalk8s/roles/etcd.sls
@@ -1,3 +1,0 @@
-mine_functions:
-  etcd_endpoints:
-    mine_function: metalk8s.get_etcd_endpoint
--- a/salt/_pillar/metalk8s_etcd.py
+++ b/salt/_pillar/metalk8s_etcd.py
@@ -1,0 +1,33 @@
+import logging
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = "metalk8s_etcd"
+
+
+def __virtual__():
+    return __virtualname__
+
+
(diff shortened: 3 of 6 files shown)


buildchain/buildchain/salt_tree.py
```diff
@@ -157,7 +157,6 @@ def _get_parts(self) -> Iterator[str]:
 PILLAR_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('pillar/metalk8s/roles/bootstrap.sls'),
     Path('pillar/metalk8s/roles/ca.sls'),
-    Path('pillar/metalk8s/roles/etcd.sls'),
```
TeddyAndrieux: No, keep it even if it's empty

pillar/metalk8s/roles/etcd.sls
```diff
@@ -1,3 +0,0 @@
-mine_functions:
```
TeddyAndrieux, line 1: Keep the empty file please

salt/_pillar/metalk8s_etcd.py
```diff
@@ -0,0 +1,34 @@
...
+		errors.append(
+			'Error while retrieving etcd cluster members: {}'.format(
+				exc
+			)
+		)
+	result = {'members': members}
+
+	if errors:
+		result.update(__utils__['pillar_utils.errors_to_dict'](errors))
+	return result
```
TeddyAndrieux: You used tabs every where here ?

salt/_pillar/metalk8s_etcd.py
```diff
@@ -0,0 +1,34 @@
+import logging
+from salt.exceptions import CommandExecutionError
```
TeddyAndrieux: Seems not used

TeddyAndrieux (commented):
Just one minor comment

TeddyAndrieux (approved):
LGTM

PR #1670, opened 2019-09-11
Use python-apt to download DEB packages

### LINKED ISSUE #1668: transpose the download_packages script in python
**Component**: build

**Why this is needed**: The download_package script in bash doesn't permit to handle virtual packages. 

**What should be done**: Transpose this script in python to use `python-apt`

- transpose the script in python using python-apt

**Test plan**:
launch the script, verify that we have all the required Debian packages and their dependencies.

slaperche-scality:
**Component**:

buildchain

**Context**: 

Got more logic than expected to download the package, move away from Bash and use Python.

**Summary**:

We use `python-apt` to compute the list of packages to download from a list of packages to explicitly install (possibly with a pinned version).
Then we download the required DEB package, dispatching them between repositories according to their origins.

**Acceptance criteria**: 

Spawn the build container:

```
% docker run --rm -it                                                         \
    -v $(pwd)/packages/debian/download_packages.py:/download_packages.py:ro \
    -v $(pwd)/_build/root/packages/debian:/repositories                     \
    --env SALT_VERSION=2018.3.4                                             \
    --env TARGET_UID=1000                                                   \
    --env TARGET_GID=100                                                    \
    metalk8s-deb-build:latest bash
```

Run the script:

```
# /download_packages.py containerd cri-tools coreutils ebtables ethtool e2fsprogs genisoimage iproute2 iptables salt-minion=2018.3.4 kubectl=1.15.3 kubelet=1.15.3 kubernetes-cni python-m2crypto runc socat sosreport util-linux xfsprogs 
```

Check that the packages are present and have the right permissions:

```
% ls -l $(pwd)/_build/root/packages/debian/*
```

---

Closes: #1668

### DIFF
--- a/packages/debian/download_packages.py
+++ b/packages/debian/download_packages.py
@@ -71,7 +71,9 @@ def select_package_origin(package: apt.package.Version) -> str:
 # In the same vein, we are doing checksum validation manually because
 # `apt_pkg.AcquireFile` is failing on some packages (the ones from the external
 # repositories…).
-def fetch_binary(package: apt.package.Version, destdir: pathlib.Path) -> None:
+def fetch_binary(
+    package: apt.package.Version, destdir: pathlib.Path
+) -> pathlib.Path:
     """Download the DEB file corresponding to the given package."""
     destdir.mkdir(exist_ok=True)  # TODO: to be done by the build chain.
 
@@ -90,6 +92,7 @@ def fetch_binary(package: apt.package.Version, destdir: pathlib.Path) -> None:
         if apt_pkg.sha256sum(fp) != package.sha256:
             raise Exception('Failed to download package {}: checksum failed'\
                             .format(filename))
+    return destfile
 
 
 # }}}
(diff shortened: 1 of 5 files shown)


packages/debian/download_packages.py
```diff
@@ -0,0 +1,70 @@
...
+    """Add a new packages source called `name` using the give URL as source."""
+    filepath = '/etc/apt/sources.list.d/{}.list'.format(name)
+    with open(filepath, 'w', encoding='utf-8') as fp:
+        print(url, file=fp)
+
+
+# }}}
+
+
+def add_external_repositories(salt_version: str) -> None:
```
jbertran, line 101: We may want to declare these externally at some point

packages/debian/download_packages.py
```diff
@@ -53,6 +54,44 @@ def select_package_version(
...
+# In the same vein, we are doing checksum validation manually because
+# `apt_pkg.AcquireFile` is failing on some packages (the ones from the external
+# repositories…).
+def fetch_binary(package: apt.package.Version, destdir: pathlib.Path) -> None:
+    """Download the DEB file corresponding to the given package."""
+    destdir.mkdir(exist_ok=True)  # TODO: to be done by the build chain.
+
+    filename = pathlib.Path(package.filename).name
+    destfile = destdir/filename
+    print('Downloading package {}'.format(filename))
```
jbertran, line 82: stray print?

ycointe (dismissed):
LGTM.

jbertran (commented):
LGTM, will approve if you want me to

packages/debian/download_packages.py
slaperche-scality: Yeah, `version.py` may be a good place.
I'll add a comment.

packages/debian/download_packages.py
slaperche-scality: Nope, on purpose: to have some input during the download (which can take a while on "slow" connection).

jbertran (approved):

PR #1672, opened 2019-09-11
Use the `prometheus-operator` chart to deploy monitoring tools

NicolasT:
Rework the monitoring deployment to use the `prometheus-operator` chart.

### TODO
- [x] Use offline containers, currently pulled from online registries
- [ ] Ensure all custom dashboards we added are still deployed (right now they're not)
- [x] Ensure the 'right' dashboards are deployed per Kubernetes version (hence `kubeTargetVersionOverride` is to be set correctly across various branches)


NicolasT:
@bert-e reset

NicolasT:
@bert-e help

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
Open for review. I may postpone the 'adding custom dashboards' to a later PR.

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -2730,3 +2730,97 @@ def show_api_service(name, **kwargs):
...
+    finally:
+        _cleanup(**cfg)
+
+
+def create_podsecuritypolicy(
+        name,
+        metadata,
+        spec,
+        **kwargs):
+    if not metadata:
```
gdemonet, line 2769: Really minor: an idiomatic one-liner for this kind of thing in Python would be `metadata = metadata or {}`.

buildchain/buildchain/salt_tree.py
```diff
@@ -204,22 +204,11 @@ def _get_parts(self) -> Iterator[str]:
...
-    Path('salt/metalk8s/addons/monitoring/node-exporter/upstream.sls'),
-    Path('salt/metalk8s/addons/monitoring/prometheus/deployed.sls'),
-    Path('salt/metalk8s/addons/monitoring/prometheus/upstream.sls'),
-    Path('salt/metalk8s/addons/monitoring/prometheus-operator/deployed.sls'),
-    Path('salt/metalk8s/addons/monitoring/prometheus-operator/upstream.sls'),
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/chart.sls'),
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/init.sls'),
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/namespace.sls'),
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/'
+            'prometheus-nodeport.sls'),
```
gdemonet, line 211: Indentation is weird, should be aligned on the line above IMHO:

```python
    Path('salt/metalk8s/addons/prometheus-operator/deployed/'
         'prometheus-nodeport.sls'),
```

buildchain/buildchain/versions.py
```diff
@@ -175,18 +180,18 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
-        version='v0.23.2',
-        digest='sha256:df1453c7c69e4f2ab8a86fc18fe3b890ce2f80fed6d6519dc9d33927451b214d',
+        version='v0.32.0',
+        digest='sha256:f1e57817dcfdb2c76e8a154b39180c6c8f3f16b990fe9cc41bee34cca0784a64',
     ),
     Image(
         name='prometheus-operator',
-        version='v0.23.2',
-        digest='sha256:8211b3eb30cb8591ddf536f1cf62100f5c97659c14d18dd45001acf94dafd713',
+        version='v0.32.0',
```
gdemonet, line 193: Maybe these version changes could warrant a small explanation in the commit message: the chart _knows_ the `kubeTargetVersion` and _decides_ which component versions can/should be used accordingly. The previous versions used were actually largely outdated, and as such, replaced in this commit.

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+  tolerations:
+    - key: 'node-role.kubernetes.io/bootstrap'
+      operator: 'Exists'
+      effect: 'NoSchedule'
+    - key: 'node-role.kubernetes.io/infra'
+      operator: 'Exists'
+      effect: 'NoSchedule'
+
+  image:
+    repository: '{%- endraw -%}{{ build_image_name(\"prometheus-operator\", False) }}{%- raw -%}'
```
gdemonet, line 46: Why the backslashes here and nowhere else?

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+    nodeSelector:
+      node-role.kubernetes.io/infra: ''
+
+    podAntiAffinity: 'soft'
+
+
+grafana:
+  image:
+    repository: '{%- endraw -%}{{ build_image_name(\"grafana\", False) }}{%- raw -%}'
```
gdemonet, line 79: Same question, why do we need backslashes here?

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+  sidecar:
+    image: '{%- endraw -%}{{ build_image_name(\"k8s-sidecar\", False) }}{%- raw -%}:0.1.20'
```
gdemonet, line 82: Why not simply use the full `build_image_name` here? If the chart doesn't know/care about the sidecar version, maybe our single source of truth (`buildchain/buildchain/versions.py`) is enough?

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+      operator: 'Exists'
+      effect: 'NoSchedule'
+    - key: 'node-role.kubernetes.io/infra'
+      operator: 'Exists'
+      effect: 'NoSchedule'
+
+
+kube-state-metrics:
+  image:
+    repository: '{%- endraw -%}{{ build_image_name(\"kube-state-metrics\", False) }}{%- raw -%}'
```
gdemonet, line 98: ditto backslashes

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+      operator: 'Exists'
+      effect: 'NoSchedule'
+    - key: 'node-role.kubernetes.io/infra'
+      operator: 'Exists'
+      effect: 'NoSchedule'
+
+
+prometheus-node-exporter:
+  image:
+    repository: '{%- endraw -%}{{ build_image_name(\"node-exporter\", False) }}{%- raw -%}'
```
gdemonet, line 114: ditto

charts/prometheus-operator.yaml
```diff
@@ -0,0 +1,114 @@
...
+    repository: '{%- endraw -%}{{ build_image_name(\"prometheus-operator\", False) }}{%- raw -%}'
+
+  configmapReloadImage:
+    repository: '{%- endraw -%}{{ build_image_name("configmap-reload", False) }}{%- raw -%}'
+
+  prometheusConfigReloaderImage:
+    repository: '{%- endraw -%}{{ build_image_name("prometheus-config-reloader", False) }}{%- raw -%}'
+
+  hyperkubeImage:
+    repository: '''{%- endraw -%}{{ build_image_name("hyperkube", False) }}{%- raw -%}'''
```
gdemonet, line 55: How will YAML interpret those triple quotes? I'm surprised with all these different notations around image names...

salt/metalk8s/addons/prometheus-operator/deployed/prometheus-nodeport.sls
```diff
@@ -18,17 +9,20 @@ Expose Prometheus:
     - context: {{ context }}
     - metadata:
         labels:
-          run: prometheus
-          prometheus: k8s
+          app: prometheus
+          app.kubernetes.io/managed-by: metalk8s
```
gdemonet, line 13: Don't we use `managed-by: salt` instead?

gdemonet (dismissed):
Only minor comments and a few questions w.r.t how we inject some Jinja in charts with values (specifically the `{{ build_image_name(...) }}` stuff and quoting around it).

PS: I did not test, but CI seems unhappy

sayf-eddine-scality:
branch 2.4 should be updated according to the addition of the `absent` argument to the renderer

buildchain/buildchain/salt_tree.py
NicolasT: Then the linter should catch that.

Now we're bikeshedding anyway: I disagree with the indentation you suggest. If I read code like this:

```
foo("abc"
    "def")
```

I mentally add a `,` after the first argument:

```
foo("abc",
    "def")
```

whilst in

```
foo("abc"
        "def")
```

I don't.

buildchain/buildchain/versions.py
NicolasT: These are not dependent on the version of the cluster.

The only diff I found is in the dashboards, not in any of the containers that are deployed.

charts/prometheus-operator.yaml
NicolasT: As explained IRL: this is because of how the render script works, how Helm splices in things in templates, whether or not the source chart templates uses quotes, which type of quotes,...

Basically, trial-and-error until rendering succeeds, and such.

charts/prometheus-operator.yaml
NicolasT: YAML escapes single-quotes as `''`, not `\'` or something along those lines. There's little to no rationale in all of this, trying options until things work.

charts/prometheus-operator.yaml
NicolasT: Because of how the chart is written. In general, they have something like

```
image:
  repository: foo/bar
  tag: v1.0.0
```

For this particular one, that split is not made, so we need to provide both. I keep the `0.1.20` here instead of only having it in `buildchain.versions` because we need to review/update this one when the chart gets updated.

salt/_modules/metalk8s_kubernetes.py
NicolasT: Copied from other similar procedures in the module.

salt/metalk8s/addons/prometheus-operator/deployed/prometheus-nodeport.sls
NicolasT: Good point.

buildchain/buildchain/salt_tree.py
slaperche-scality: BTW, `Path` can take a list of string instead of a single string, so

```
    Path('salt/metalk8s/addons/prometheus-operator/deployed/',
         'prometheus-nodeport.sls'),
```

is an option :wink:

NicolasT:
@bert-e reset

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
> branch 2.4 should be updated according to the addition of the absent argument to the renderer

Done (I believe)

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

charts/prometheus-operator.yaml
gdemonet: OK, I get the rationale. Maybe worth adding a comment next to it then, but feel free to ignore if CI is happy now :)

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e reset

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
Fixed a (hopefully) last issue in the 2.4 integration.

I'd like to add the extra dashboards in another PR.

gdemonet:
@bert-e status

buildchain/buildchain/versions.py
gdemonet: Oh indeed, the `kubeTargetVersion` checks in the charts are only used for dashboards and Prometheus rules, so I guess it's fine :)

gdemonet (approved):
LGTM!

Issue #1332, opened 2019-07-01
Documentation about `products` in bootstrap.yaml


TeddyAndrieux:
**Component**:

'docs'

**Why this is needed**:

Currently we don't give strict instruction about what is needed in `products:metalk8s` in `bootstrap.yaml`

**What should be done**:

In `products:metalk8s` we can either put path to a directory where the iso is already mounted/extracted or a path to an iso. 
We should only describe one approach in the documentation.

2 options (IMO):

1. Path to a directory: describe how mount the iso and configure the fstab manually and then launch the bootstrap.sh
2. Path to the iso: mount the iso anywhere and just launch the bootstrap.sh (iso will be automatically mounted in the right directory and fstab configured automatically

**Implementation proposal** (strongly recommended):

Suggest solution 2. just above

sayf-eddine-scality:
rename `products` into more meaningful name

charlesprost:
As discussed, we could use `archives` instead of `products` and use a plain list instead of having the `:metalk8s` sub tree

charlesprost:
Should be available starting from 2.0 also

PR #1674, opened 2019-09-12
Improve products field from bootstrap config

slaperche-scality:
**Component**:

salt, documentation

**Context**: 

See #1332 

**Summary**:

- Rename `products` to `archives`
- Use a list instead of a dict of list
- Document what to put in this field
- Update code accordingly

**Acceptance criteria**: 

Everything is working as before.

---

Closes: #1332


slaperche-scality:
Was wondering, do we still need to support the `give path to an already mounted ISO`?

If not, we can probably remove the code to handle that and only keep the code to deal with path to ISO file.

TeddyAndrieux:
> Was wondering, do we still need to support the `give path to an already mounted ISO`?
> 
> If not, we can probably remove the code to handle that and only keep the code to deal with path to ISO file.

It's still needed at least for vagrant as we not build iso when we use vagrant. Also it's good to have this in case we want live debugging we need a NOT read only mounted iso

salt/_modules/metalk8s.py
```diff
@@ -181,61 +181,57 @@ def product_info_from_iso(path):
...
+    if not archives:
+        archives = __pillar__.get('metalk8s', {}).get('archives', [])
+
+    if isinstance(archives, basestring):
+        archives = [str(archives)]
+    elif not isinstance(archives, list):
         raise CommandExecutionError(
-            'Invalid products list or string expected and got: {0}'.format(
-                products
+            'Invalid archives list or string expected and got: {0}'.format(
```
jbertran: Let's add a colon after `archives` so this sentence makes sense

jbertran (dismissed):
Looking good!

salt/_modules/metalk8s.py
slaperche-scality: Indeed.

Done

docs/quickstart/bootstrap.rst
```diff
@@ -40,15 +40,17 @@ Configuration
         minion: <hostname-of-the-bootstrap-node>
       apiServer:
         host: <IP-of-the-bootstrap-node>
-      products:
-        metalk8s:
-          - <path-to-extracted-directory-or-iso>
+      archives:
+        - <path-to-metalk8s-iso>
+
+The ``archives`` field is a list of path to MetalK8s ISO files. When the
```
gdemonet: Minor:

```suggestion
The ``archives`` field is a list of paths to MetalK8s ISO files. When the
```

gdemonet: Also, if you make the change, could be good to specify that they are **absolute** paths, to avoid questions :)

docs/quickstart/bootstrap.rst
```diff
@@ -40,15 +40,17 @@ Configuration
...
+bootstrap script is executed, those ISOs are automatically mounted and the
+``fstab`` is updated accordingly.
```
gdemonet: Let's keep it for now, but we may want to find a more generic formulation that even Windows users could understand (e.g. `and the mounting will be resilient to reboots.`).

salt/_modules/metalk8s.py
```diff
@@ -128,25 +128,25 @@ def _get_product_name(info):
     return match.group('name') if match else None
 
 
-def _get_product_info(info):
-    """Extract product information from info
+def _get_archive_info(info):
```
gdemonet, line 131: I think we could have left this one (and other similar references) as _**product** info_, since we basically extract it from `product.txt`.

- `archive` describes the ISO or directory, which we configure in `BootstrapConfiguration`
- `product` describes the software that is shipped within an `archive`

gdemonet (dismissed):
Nitpicking on a nuance between `archive` and `product` in the code, feel free to ignore ;)

docs/quickstart/bootstrap.rst
slaperche-scality: Well, we probably have several formulation to improve to be really noob-friendly, can be done more globally.

salt/_modules/metalk8s.py
slaperche-scality: Was discussed on Slack yesterday: do we only rename the visible part or the whole pillar/salt stuff.
Was agreed to rename everything to be consistent so I'll keep it like that.

docs/quickstart/bootstrap.rst
slaperche-scality: Done

gdemonet (approved):
LGTM

Vagrantfile
```diff
@@ -132,8 +132,7 @@ apiServer:
   host: #{IPAddr.new(CONTROL_PLANE_IP).mask(CONTROL_PLANE_NETMASK).to_range.last(2).first.to_s}
   keepalived:
     enabled: true
-products:
-  metalk8s:
+archives:
   - /srv/scality/metalk8s-$VERSION
```
charlesprost, line 136: Don't we need to reindent something? it seems weird removing one level and still having the inner list at the same level

docs/operation/preparation.rst
```diff
@@ -8,5 +8,5 @@ with the current installation:
 
    .. code::
 
-     /srv/scality/metalk8s-X.X.X/iso-manager.sh -p <path_to_iso>
+     /srv/scality/metalk8s-X.X.X/iso-manager.sh -a <path_to_iso>
```
charlesprost, line 11: Could we use the long option name instead? I always find it more obvious when used in scripts (let's keep the short name for manual interactions)

docs/quickstart/bootstrap.rst
```diff
@@ -40,15 +40,17 @@ Configuration
         minion: <hostname-of-the-bootstrap-node>
       apiServer:
         host: <IP-of-the-bootstrap-node>
-      products:
-        metalk8s:
-          - <path-to-extracted-directory-or-iso>
+      archives:
+        - <path-to-metalk8s-iso>
```
charlesprost, line 44: Here we reindent, that's cool.

salt/_modules/metalk8s.py
```diff
@@ -181,61 +181,57 @@ def product_info_from_iso(path):
...
+            path = archive
             version = info['version']
-        elif os.path.isfile(prod):
-            iso = prod
-            info = product_info_from_iso(prod)
+        elif os.path.isfile(archive):
+            iso = archive
+            info = archive_info_from_iso(archive)
             version = info['version']
             path = '/srv/scality/metalk8s-{0}'.format(version)
```
charlesprost, line 217: I find the `path` variable name here confusing because from what I understand by reading the code a few times, it's a dir path, but by looking at this 217th line quickly, it could look like a file path. Could we have it renamed `dirpath` for example?

Vagrantfile
gdemonet: YAML... Personally, I prefer to indent lists, but it's not strictly required.

PR #1675, opened 2019-09-12
ui: integrate Flow for type checking

### LINKED ISSUE #1663: integrate Flow in Metalk8s
**Component**: ui

**Why this is needed**:
1. Since we have Flow in core-ui, if we implement it in metalk8s, it will give us more info for the parameters.
2. Help all the dev to follow the structure of parameters.

**What should be done**:
Integrate Flow in Metalk8s 

TBD

**Test plan**:
TBD

ChengYanJin:
**Component**: ui

**Context**: 
We would like to integrate Flow in MetalK8S.

**Summary**:

**Acceptance criteria**: 
If we execute `npm run flow` in the command line, Flow will help us check the type.
We currently just use Flow in the specific component, `Banner`, `InformationList`, `Loader`, will progressively use it into all the components and containers.

follows: #1677 
Closes: #1663

### DIFF
--- a/ui/.flowconfig
+++ b/ui/.flowconfig
@@ -1,0 +1,12 @@
+[ignore]
+./node_modules/.*
+.*/node_modules/module-deps/.*
+[include]
+src/.*
+[libs]
+
+[lints]
+
+[options]
+esproposal.optional_chaining=enable
+[strict]
--- a/ui/config-overrides.js
+++ b/ui/config-overrides.js
@@ -15,8 +15,9 @@ const CompressionPlugin = require('compression-webpack-plugin');
 const setWebpackPerformance = () => config => {
   config.performance = {
     hints: 'error',
-    // ~390 KiB
-    maxAssetSize: 400000,
+    // ~390 KiB for production
+    // ~1953 KiB for development because flow increase the size of assets.
+    maxAssetSize: process.env.NODE_ENV === 'production' ? 400000 : 2000000,
     assetFilter: assetFilename => {
       return (
         !assetFilename.endsWith('.map.gz') && assetFilename.endsWith('.gz')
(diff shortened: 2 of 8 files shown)


ui/config-overrides.js
```diff
@@ -16,7 +16,7 @@ const setWebpackPerformance = () => config => {
   config.performance = {
     hints: 'error',
     // ~390 KiB
-    maxAssetSize: 400000,
+    maxAssetSize: process.env.NODE_ENV === 'production' ? 400000 : 2000000,
```
slaperche-scality, line 20: I don't understand the relation with this change and Flow (and nothing mention it in the commit message).

ui/.flowconfig
```diff
@@ -0,0 +1,12 @@
+[ignore]
```
slaperche-scality, line 1: noob question: why `ignore` instead of `untyped`?

ui/config-overrides.js
MonPote: We will add a comment for this specific line. But the reason si that flow generate a lot of stuff for development. It result to a big build (over 1MiB) asset, flow is not include in the production build so we don't have this issue.

slaperche-scality: Yeah, we should have an explanation like your reply here in the commit message.

ui/.flowconfig
MonPote: It's  because we don't want flow to parse `node_modules` at all (it contains the entire internet).
```
[ignore] - Have flow not parse
[untyped] - Have flow parse but throw away types
```

MonPote (dismissed):

ui/.flowconfig
slaperche-scality: Fair enough, that would probably make everything slower.

I wonder what benefit does Flow have to parse stuff but typing everything as `any` (which is what `untyped` does).
Maybe it can at least check that the types does exists and catch typo? Anyway, probably not worth the slowdown.

Thanks for the answer!

slaperche-scality (dismissed):

MonPote (approved):

ChengYanJin:
@bert-e approve

PR #1676, opened 2019-09-12
Improvement/1634 implement download packages

### LINKED ISSUE #1634: Implement the download debian package script in the buildchain
**Component**: build

**Why this is needed**: We have a script in the Debian entrypoint to download Debian packages that we have to integrate it in the buildchain tasks

**What should be done**:
- [ ] Implement this solution in the buildchain

**Test plan**:

ycointe:
**Component**: Buildchain, package

**Summary**: We need to implement the script download_packages.py in the buildchain to download some Debian packages from official repositories.

**Acceptance criteria**: 
Build the MetalK8s ISO with the Debian packages downloaded.

---

Closes: #1634

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -54,6 +54,7 @@ def task_packaging() -> types.TaskDict:
             '_build_rpm_packages:*',
             '_build_rpm_repositories:*',
             '_build_deb_packages:*',
+            '_download_deb_packages',
         ],
     }
 
@@ -134,7 +135,7 @@ def task__download_rpm_packages() -> types.TaskDict:
         run_config=docker_command.RPM_BASE_CONFIG
     )
     return {
-        'title': utils.title_with_target1('GET PKGS'),
+        'title': utils.title_with_target1('GET RPM PKGS'),
         'actions': [dl_packages_callable],
         'targets': [constants.PKG_RPM_ROOT/'var'],
         'task_dep': [
@@ -147,6 +148,40 @@ def task__download_rpm_packages() -> types.TaskDict:
         # Prevent Docker from polluting our output.
         'verbosity': 0,
     }
+
+
+def task__download_deb_packages() -> types.TaskDict:
+    """Download Debian packages locally."""
+    # TODO: Clean the repository
+    mounts = [
+        utils.bind_ro_mount(
+            source=constants.ROOT/'packages'/'debian'/'download_packages.py',
+            target=Path('/download_packages.py'),
+        ),
+        utils.bind_mount(
+            source=constants.REPO_DEB_ROOT,
+            target=Path('/repositories')
+        ),
+    ]
+    dl_packages_callable = docker_command.DockerRun(
+        command=['/download_packages.py', *DEB_TO_DOWNLOAD],
+        builder=DEB_BUILDER,
+        mounts=mounts,
+        environment={'SALT_VERSION': versions.SALT_VERSION},
+        run_config=docker_command.DEB_BASE_CONFIG
+    )
+    return {
+        'title': utils.title_with_target1('GET DEB PKGS'),
+        'actions': [dl_packages_callable],
+        'targets': [constants.REPO_DEB_ROOT/'.witness'],
+        'task_dep': [
+            '_package_mkdir_deb_root',
+            '_package_mkdir_deb_iso_root',
+            '_build_deb_container'
+        ],
+        'uptodate': [config_changed(_TO_DOWNLOAD_DEB_CONFIG)],
+    }
+
 
 def task__build_rpm_packages() -> Iterator[types.TaskDict]:
     """Build a RPM package."""
@@ -274,6 +309,7 @@ RPM_TO_BUILD : Dict[str, Tuple[targets.RPMPackage, ...]] = {
 }
 
 _RPM_TO_BUILD_PKG_NAMES : List[str] = []
+_DEB_TO_BUILD_PKG_NAMES : List[str] = []
 
 for pkgs in RPM_TO_BUILD.values():
     for pkg in pkgs:
@@ -287,12 +323,20 @@ RPM_TO_DOWNLOAD : FrozenSet[str] = frozenset(
     if package.name not in _RPM_TO_BUILD_PKG_NAMES
 )
 
+
 # Store these versions in a dict to use with doit.tools.config_changed
 _TO_DOWNLOAD_CONFIG : Dict[str, str] = {
     pkg.name: "{p.version}-{p.release}".format(p=pkg)
     for pkg in versions.RPM_PACKAGES
     if pkg.name not in _RPM_TO_BUILD_PKG_NAMES
 }
+
+_TO_DOWNLOAD_DEB_CONFIG : Dict[str, str] = {
+    pkg.name: "{p.version}-{p.release}".format(p=pkg)
+    for pkg in versions.DEB_PACKAGES
+    if pkg.name not in _DEB_TO_BUILD_PKG_NAMES
+}
+
 
 SCALITY_RPM_REPOSITORY = targets.RPMRepository(
     basename='_build_rpm_repositories',
@@ -356,5 +400,10 @@ DEB_TO_BUILD : Dict[str, Tuple[targets.DEBPackage, ...]] = {
     )
 }
 
+DEB_TO_DOWNLOAD : FrozenSet[str] = frozenset(
+        "{p.name}".format(p=package)
+        for package in versions.DEB_PACKAGES
+        if package.name not in DEB_TO_BUILD
+)
 
 __all__ = utils.export_only_tasks(__name__)
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -280,17 +280,17 @@ for pkgs in RPM_TO_BUILD.values():
         _RPM_TO_BUILD_PKG_NAMES.append(pkg.name)
 
 # All packages not referenced in `RPM_TO_BUILD` but listed in
-# `versions.PACKAGES` are supposed to be downloaded.
+# `versions.RPM_PACKAGES` are supposed to be downloaded.
 RPM_TO_DOWNLOAD : FrozenSet[str] = frozenset(
     "{p.name}-{p.version}-{p.release}".format(p=package)
-    for package in versions.PACKAGES
+    for package in versions.RPM_PACKAGES
     if package.name not in _RPM_TO_BUILD_PKG_NAMES
 )
 
 # Store these versions in a dict to use with doit.tools.config_changed
 _TO_DOWNLOAD_CONFIG : Dict[str, str] = {
     pkg.name: "{p.version}-{p.release}".format(p=pkg)
-    for pkg in versions.PACKAGES
+    for pkg in versions.RPM_PACKAGES
     if pkg.name not in _RPM_TO_BUILD_PKG_NAMES
 }
 
--- a/buildchain/buildchain/versions.py
+++ b/buildchain/buildchain/versions.py
@@ -230,7 +230,7 @@ CONTAINER_IMAGES_MAP = {image.name: image for image in CONTAINER_IMAGES}
 # }}}
 # Packages {{{
 
-PACKAGES = (
+RPM_PACKAGES = (
     # Remote packages
     Package(
         name='containerd',
@@ -360,6 +360,106 @@ PACKAGES = (
     ),
 )
 
-PACKAGES_MAP = {pkg.name: pkg for pkg in PACKAGES}
+PACKAGES_MAP = {pkg.name: pkg for pkg in RPM_PACKAGES}
+
+
+DEB_PACKAGES = (
+    Package(
+        name='containerd',
+        version='',
+        release='',
+    ),
+    Package(
+        name='cri-tools',
+        version='',
+        release='',
+    ),
+    Package(
+        name='coreutils',
+        version='',
+        release='',
+    ),
+    Package(
+        name='ebtables',
+        version='',
+        release='',
+    ),
+    Package(
+        name='ethtool',
+        version='',
+        release='',
+    ),
+    Package(
+        name='e2fsprogs',
+        version='',
+        release='',
+    ),
+    Package(
+        name='genisoimage',
+        version='',
+        release='',
+    ),
+    Package(
+        name='iproute2',
+        version='',
+        release='',
+    ),
+    Package(
+        name='iptables',
+        version='',
+        release='',
+    ),
+    Package(
+        name='salt-minion={}'.format(SALT_VERSION),
+        version='',
+        release='',
+    ),
+    Package(
+        name='kubectl={}'.format(K8S_VERSION),
+        version='',
+        release='',
+    ),
+    Package(
+        name='kubelet={}'.format(K8S_VERSION),
+        version='',
+        release='',
+    ),
+    Package(
+        name='kubernetes-cni',
+        version='',
+        release='',
+    ),
+    Package(
+        name='python-m2crypto',
+        version='',
+        release='',
+    ),
+    Package(
+        name='runc',
+        version='',
+        release='',
+    ),
+    Package(
+        name='socat',
+        version='',
+        release='',
+    ),
+    Package(
+        name='sosreport',
+        version='',
+        release='',
+    ),
+    Package(
+        name='util-linux',
+        version='',
+        release='',
+    ),
+    Package(
+        name='xfsprogs',
+        version='',
+        release='',
+    ),
+)
+
 
 # }}}
--- a/packages/debian/download_packages.py
+++ b/packages/debian/download_packages.py
@@ -176,6 +176,11 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
 
 def main(packages: Sequence[str], env: Mapping[str, str]) -> None:
     """Download the packages specified on the command-line."""
+    witness_file = pathlib.Path('/repositories/.witness')
+    try:
+        witness_file.unlink()
+    except FileNotFoundError:
+        pass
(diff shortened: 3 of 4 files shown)


buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
         'verbosity': 0,
     }
 
+
+def task__download_deb_packages() -> types.TaskDict:
+    """Download Debian packages locally."""
+    # TO DO: Clean the repository
```
jbertran: That's `TODO` if you want it to show up all pretty in Python syntax highlighters :)

buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
...
+    def clean() -> None:
+        """Delete cache and repositories on the ISO."""
+        coreutils.rm_rf(constants.REPO_RPM_ROOT/'repositories')
```
jbertran: Did you mean `REPO_DEB_ROOT`?

buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
...
+        for repository in DEB_REPOSITORIES:
+            # Repository with an explicit list of packages are created by a
```
jbertran:
```suggestion
            # Repositories with an explicit list of packages are created by a
```

buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
...
     }
 
+
+def task__download_deb_packages() -> types.TaskDict:
+    """Download Debian packages locally."""
+    # TO DO: Clean the repository
+    def clean() -> None:
+        """Delete cache and repositories on the ISO."""
+        coreutils.rm_rf(constants.REPO_RPM_ROOT/'repositories')
+        for repository in DEB_REPOSITORIES:
```
jbertran: This isn't defined in this file, nor is it imported.

buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
...
+    ]
+    dl_packages_callable = docker_command.DockerRun(
+        command=['/download_packages.py', *DEB_TO_DOWNLOAD],
+        builder=DEB_BUILDER,
+        mounts=mounts,
+        environment={'SALT_VERSION': versions.SALT_VERSION},
+        run_config=docker_command.DEB_BASE_CONFIG
+    )
+    return {
+        'title': utils.title_with_target1('GET DEB PKGS'),
```
jbertran, line 174: Reflect this on the RPM task - should be `GET RPM PKGS`

buildchain/buildchain/packaging.py
```diff
@@ -287,13 +333,21 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
...
 _TO_DOWNLOAD_CONFIG : Dict[str, str] = {
     pkg.name: "{p.version}-{p.release}".format(p=pkg)
     for pkg in versions.PACKAGES
     if pkg.name not in _RPM_TO_BUILD_PKG_NAMES
 }
 
+_TO_DOWNLOAD_DEB_CONFIG : Dict[str, str] = {
+    pkg.name: "{p.version}-{p.release}".format(p=pkg)
+    for pkg in versions.PACKAGE_DEB
+        if pkg.name not in _DEB_TO_BUILD_PKG_NAMES
```
jbertran: this should be at the same indent level as the rest

buildchain/buildchain/packaging.py
```diff
@@ -356,5 +410,10 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
     )
 }
 
+DEB_TO_DOWNLOAD : FrozenSet[str] = frozenset(
+        "{p.name}".format(p=package)
+        for package in versions.PACKAGE_DEB
+            if package.name not in DEB_TO_BUILD
```
jbertran: ditto

packages/debian/download_packages.py
```diff
@@ -176,16 +176,22 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
 
 def main(packages: Sequence[str], env: Mapping[str, str]) -> None:
     """Download the packages specified on the command-line."""
+    #pathlib.Path('/repositories/.witness').unlink()
```
jbertran: Stray comment

packages/debian/download_packages.py
```diff
@@ -176,16 +176,22 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
...
     cache = apt.cache.Cache()
     to_download = {}
+    dest = pathlib.Path('/repositories')
     for package in packages:
         deps = get_package_deps(package, cache)
         to_download.update(deps)
     for pkg in to_download.values():
         filepath = download_package(pkg)
         os.chown(filepath, int(env['TARGET_UID']), int(env['TARGET_GID']))
+    #pathlib.Path('/repositories/.witness').touch()
```
jbertran: Stray comment again, but we're going to need this one (or a variation of it) in order to be coherent with the declared `target` in your task.

Also, this guy should be at the very end of your script.

packages/debian/download_packages.py
```diff
@@ -176,16 +176,22 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
...
+    for rep in dest.iterdir():
```
jbertran: let's keep english names - `for directory in dest.iterdir()` :)

jbertran (commented):
Nice work! I'll let @gdemonet comment on the version management

buildchain/buildchain/versions.py
```diff
@@ -362,4 +362,104 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
+        version='',
+        release='',
+    ),
+    Package(
+        name='iptables',
+        version='',
+        release='',
+    ),
+    Package(
+        name='salt-minion={}'.format(SALT_VERSION),
```
gdemonet, line 413: Version in the name defeats the purpose of this listing. We should instead find some way to only mark `version`/`release` when needed (and, doing so, remove them where not needed for RPM packages as well).

I'll fork something from your branch to show you and let you clean it up afterwards.

slaperche-scality: Can probably be handled when we generate the strings for `DEB_TO_DOWNLOAD `, or maybe you have an even better idea :slightly_smiling_face:

gdemonet: Yup, that would be a quick win.

Best solution would be to handle "OSes"/"Package managers" in a more generic way (the only difference is the separator being a `-` for `yum` and `=` for `apt` apparently). This way, we could also avoid duplication for "common" packages, though, granted, it would cost slightly more upfront.

slaperche-scality: > the only difference is the separator being a - for yum and = for apt apparently

Actually, in my PoC for `use Yum in Python` I went back to `=` instead of `-`, rational being that since we're giving a string to describe a package and that package name can contains `-` (`cri-tools` for instance) it can be hard to know where to split to separate `name` and `version`, less issue with `=` (or even `;` I guess).

But at that point, if we're gonna use Python anyway we could generate a JSON file and mount it in the container, let the script parse it and use the content instead of relying on plain "strings" with custom "encoding".

The `-` vs `=` things mainly come from the use of CLI tools, we're not limited by that if we use Python instead (where API functions usually take name and version as distinct parameters instead of a whole string).

buildchain/buildchain/versions.py
```diff
@@ -362,4 +362,104 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
 
 PACKAGES_MAP = {pkg.name: pkg for pkg in PACKAGES}
 
+
```
slaperche-scality, line 365: About commit message

> RedHat and Debian don't use the same packages so we need a new list of these
packages.

You could even cite `iproute2` vs `iproute` as example :wink:

buildchain/buildchain/versions.py
```diff
@@ -362,4 +362,104 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
+PACKAGE_DEB = (
+    Package(
+        name='containerd',
+        version='',
```
slaperche-scality, line 369: When no version and release are specified I would rather use None than an empty string, to catch misuse more easily.

And to avoid boilerplate, I would set None as default value so that you don't need to specify them.
See [here](https://stackoverflow.com/a/18348004) how to do that.

Since we're not requiring Python 3.7, we can do something like that:

```diff
 Package = namedtuple('Package', ('name', 'version', 'release'))
+Package.__new__.__defaults__ = (None, None)
```
Since fields with a default value must come after any fields without a default, the defaults are applied to the rightmost parameters :wink: 

Which then will make the code shorter:
```python

PACKAGE_DEB = (
    Package(name='containerd'),
    Package(name='cri-tools'),
    Package(name='coreutils'),
    Package(name='ebtables'),
    Package(name='ethtool'),
    Package(name='e2fsprogs'),
    Package(name='genisoimage'),
    Package(name='iproute2'),
    Package(name='iptables'),
    Package(name='salt-minion', version=SALT_VERSION),
    Package(name='kubectl', version=K8S_VERSION),
    Package(name='kubelet', version=K8S_VERSION),
    Package(name='kubernetes-cni'),
    Package(name='python-m2crypto'),
    Package(name='runc'),
    Package(name='socat'),
    Package(name='sosreport'),
    Package(name='util-linux'),
    Package(name='xfsprogs'),
)
```

buildchain/buildchain/versions.py
```diff
@@ -362,4 +362,104 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
 
 PACKAGES_MAP = {pkg.name: pkg for pkg in PACKAGES}
 
+
+PACKAGE_DEB = (
```
slaperche-scality: IMHO, it should be `DEB_PACKAGES`.

Also, you should rename `PACKAGES` to `RPM_PACKAGES` to be consistent
And we may want to have fold for each of those big list (that being said, as we can make them a lot shorter, see below, maybe it's not required).

buildchain/buildchain/packaging.py
slaperche-scality: And to be removed, since the TODO seems implemented

buildchain/buildchain/packaging.py
slaperche-scality: Yeah that, and also as we discussed yesterday `repositories` only exist in the container, not on your local FS so this `rm` is not gonna do something…

buildchain/buildchain/packaging.py
```diff
@@ -274,6 +319,7 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
 }
 
 _RPM_TO_BUILD_PKG_NAMES : List[str] = []
+_DEB_TO_BUILD_PKG_NAMES : List[str] = []
```
slaperche-scality, line 312: Where is this variable populated?

buildchain/buildchain/packaging.py
slaperche-scality: Also, since the code is a straight copy-paste from above, this should probably be factored…

packages/debian/download_packages.py
slaperche-scality: You shouldn't need that, as explained in the TODO of this script the directory must be created by the buildchain (and thus you won't need the `chown` here).

packages/debian/download_packages.py
slaperche-scality: How is this gonna work if you keep the code commented?

BTW the use of witness here AND in the buildchain should be introduced in the same commit (with a nice commit message explaining what they are/why we need them).

buildchain/buildchain/packaging.py
```diff
@@ -148,6 +149,50 @@ def clean() -> None:
...
+        command=['/download_packages.py', *DEB_TO_DOWNLOAD],
+        builder=DEB_BUILDER,
+        mounts=mounts,
+        environment={'SALT_VERSION': versions.SALT_VERSION},
+        run_config=docker_command.DEB_BASE_CONFIG
+    )
+    return {
+        'title': utils.title_with_target1('GET DEB PKGS'),
+        'actions': [dl_packages_callable],
+        'targets': [constants.REPO_DEB_ROOT/'.witness'],
```
slaperche-scality, line 176: `.witness` should be introduced later, when you also implement it in the script, otherwise it doesn't make sense.

slaperche-scality (changes requested):

buildchain/buildchain/packaging.py
ycointe: I should not have commited this, it won't be used until I create the repository for Debian.
I will corrected it.

buildchain/buildchain/packaging.py
ycointe: I used ` _RPM_TO_BUILD_PKG_NAMES : List[str] = [ ` as exemple to created it.
It's used by `DEB/RPM_TO_DOWNLOAD`

buildchain/buildchain/packaging.py
ycointe: But if I add nothing after `'targets': [constants.REPO_DEB_ROOT/'`, doit.sh crash because there is already something on this target

buildchain/buildchain/versions.py
ycointe: I will use the exemple of m2crypto / python-m2crypto, iproute is about a  package depreciation :)

packages/debian/download_packages.py
ycointe: Agree with that, but the repository folder will be created by the class DEBRepository in targets/repository.py like it's done for RedHat... so it will be handled at the next step about the repository... what do you think about it ?

buildchain/buildchain/packaging.py
slaperche-scality: Yeah, you use it.
But you never put any value into it, which seems buggy to me.

You do have 2 packages to build for Debian, so it's weird that `_DEB_TO_BUILD_PKG_NAMES` is empty.
Either put something into it, or remove it.
But as it is, it's really weird.

buildchain/buildchain/packaging.py
slaperche-scality: even with  `'targets': []`?

packages/debian/download_packages.py
slaperche-scality: Why not, but then add a TODO like I did (`# TODO: to be done by the build chain.`)

buildchain/buildchain/packaging.py
ycointe: Factorization will be handled by the issue #1693

buildchain/buildchain/versions.py
ycointe: This will be handled by the issue #1693

buildchain/buildchain/versions.py
ycointe: This will be handled by the issue #1693

buildchain/buildchain/packaging.py
ycointe: Will be handled by the issue #1693

jbertran (dismissed):
I, holder of most changesets for the tracking ticket, approve this message.

buildchain/buildchain/versions.py
```diff
@@ -360,6 +360,106 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
     ),
 )
 
-PACKAGES_MAP = {pkg.name: pkg for pkg in PACKAGES}
+PACKAGES_MAP = {pkg.name: pkg for pkg in RPM_PACKAGES}
```
slaperche-scality, line 363: Should probably be renamed `RPM_PACKAGES_MAP`, but anyway this part will be reworked by Jordi

buildchain/buildchain/packaging.py
slaperche-scality: Still here, should have been introduced in `build/package: add a witness to verify the well task running`

packages/debian/download_packages.py
```diff
@@ -176,16 +176,28 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
...
     to_download = {}
+    dest = pathlib.Path('/repositories')
     for package in packages:
         deps = get_package_deps(package, cache)
         to_download.update(deps)
     for pkg in to_download.values():
         filepath = download_package(pkg)
         os.chown(filepath, int(env['TARGET_UID']), int(env['TARGET_GID']))
+    # TODO: need to be done by the build chain
+    for directory in dest.iterdir():
```
slaperche-scality, line 196: Why do we have the loop twice? oO

packages/debian/download_packages.py
```diff
@@ -176,16 +176,28 @@ def download_package(package: apt.package.Version) -> pathlib.Path:
...
+        os.chown(directory, int(env['TARGET_UID']), int(env['TARGET_GID']))
+    pathlib.Path('/repositories/.witness').touch()
```
slaperche-scality: Probably need to chown this file as well.

packages/debian/download_packages.py
```diff
@@ -196,7 +196,7 @@ def main(packages: Sequence[str], env: Mapping[str, str]) -> None:
         os.chown(directory, int(env['TARGET_UID']), int(env['TARGET_GID']))
     pathlib.Path('/repositories/.witness').touch()
     for directory in dest.iterdir():
-        os.chown(dirctory, int(env['TARGET_UID']), int(env['TARGET_GID']))
+        os.chown(directory, int(env['TARGET_UID']), int(env['TARGET_GID']))
```
slaperche-scality: I think the code for this commit is in the previous commit :/

slaperche-scality (changes requested):

slaperche-scality (approved):

PR #1681, opened 2019-09-13
monitoring: use persistent storage for Prometheus

### LINKED ISSUE #1515: Update Prometheus deployment to use a PVC

NicolasT:
These changes configure the Prometheus stateful services (i.e.,
Prometheus and AlertManager) to use a persistent volume for storage
instead of ephemeral `emptyDir` volumes.

As a result, these services don't 'work' out of the box after initial
deployment. We add some example `Volume` templates using the
`sparseLoopDevice` storage type to create proper volumes on a node after
deployment, which should allow the services to start.

Furthermore, the Vagrant deployment and CI are extended to create the
required `Volume`s such that the Prometheus services *are* available
after bootstrapping the system.

Note: during an earlier implementation of this feature (#1523), we ran
into a dependency issue between Prometheus deployment and the MetalK8s
UI deployment, which requires the first to be available. Instead, given
this PR, we use a hard-coded `NodePort` port number to access
Prometheus, not requiring the service to be available during bootstrap,
and point the UI to the workload-plane IP of the bootstrap node. This
ought to be improved once we use an `Ingress` to access the UI.

Fixes: #1515
See: https://github.com/scality/metalk8s/issues/1515
See: https://github.com/scality/metalk8s/pull/1523

### DIFF
--- a/eve/create-volumes.sh
+++ b/eve/create-volumes.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+# Note: this script's *defaults* must adhere to the Vagrant environment in
+# which it is used.
+
+set -eu -o pipefail
+
+BOOTSTRAP_NODE_NAME=${BOOTSTRAP_NODE_NAME:-$(hostname)}
+PRODUCT_TXT=${PRODUCT_TXT:-/vagrant/_build/root/product.txt}
+
+# shellcheck disable=SC1090
+source "${PRODUCT_TXT}"
+
+PRODUCT_MOUNT=${PRODUCT_MOUNT:-/srv/scality/metalk8s-${VERSION}}
+
+if ! test -f "${PRODUCT_MOUNT}/examples/prometheus-sparse.yaml"; then
+    echo 1>&2 "Storage provisioning template not found."
+    exit 1
+fi
+
+KUBECONFIG=${KUBECONFIG:-/etc/kubernetes/admin.conf}
+export KUBECONFIG
+
+echo "Creating storage volumes"
+sed "s/BOOTSTRAP_NODE_NAME/${BOOTSTRAP_NODE_NAME}/" "${PRODUCT_MOUNT}/examples/prometheus-sparse.yaml" | \
+    kubectl apply -f -
+
+OK=0
+echo "Waiting for PV 'bootstrap-alertmanager' to be provisioned"
+for _ in $(seq 1 60); do
+    if kubectl get pv bootstrap-alertmanager > /dev/null 2>&1; then
+        OK=1
+        break
+    fi
+    sleep 1
+done
+[ $OK -eq 1 ] || (echo "PV not created"; false)
+
+OK=0
+echo "Waiting for PV 'bootstrap-prometheus' to be provisioned"
+for _ in $(seq 1 60); do
+    if kubectl get pv bootstrap-prometheus > /dev/null 2>&1; then
+        OK=1
+        break
+    fi
+    sleep 1
+done
+[ $OK -eq 1 ] || (echo "PV not created"; false)
+
+OK=0
+echo 'Waiting for AlertManager to be running'
+for _ in $(seq 1 60); do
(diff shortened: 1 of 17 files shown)


NicolasT:
@bert-e wait

NicolasT:
Once #1711 is in, this could be picked up again.

NicolasT:
@bert-e wait

NicolasT:
TODOs:

- [x] Docs

eve/create-volumes.sh
```diff
@@ -0,0 +1,72 @@
...
+    fi
+    sleep 1
+done
+[ $OK -eq 1 ] || (echo "AlertManager not Running"; false)
+
+OK=0
+echo 'Waiting for Prometheus to be running'
+for _ in $(seq 1 60); do
+    PHASE=$(kubectl -n metalk8s-monitoring get pod prometheus-prometheus-operator-prometheus-0 -o jsonpath="{.status.phase}")
+    if [ "x${PHASE}" = "xRunning" ]; then
```
slaperche-scality, line 66: why the `x`?

eve/create-volumes.sh
```diff
@@ -0,0 +1,72 @@
...
+        OK=1
+        break
+    fi
+    sleep 1
+done
+[ $OK -eq 1 ] || (echo "Prometheus not Running"; false)
```
slaperche-scality, line 72: I guess the `wait for PV` and `wait for Pod` could be in function, and then we would call each of them twice, but meh it's shell scripts anyway…

salt/metalk8s/addons/prometheus-operator/deployed/storageclass.sls
```diff
@@ -1,14 +1,18 @@
-#!jinja | kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+#! kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
```
slaperche-scality, line 1: the move (and update) could have been in a dedicated commit, to explain the rationale behind the change/move.

Could also edit the current commit message to explain it.

eve/create-volumes.sh
NicolasT: Old habits don't go away easily... https://unix.stackexchange.com/a/136632

eve/create-volumes.sh
NicolasT: Not worth the trouble IMHO.

salt/metalk8s/addons/prometheus-operator/deployed/storageclass.sls
NicolasT: I'll amend the commit message. Not fond of doing moves like this (which don't bring any value per-se outside of the context of the actual commit doing the PV stuff and such) in separate commits.

eve/create-volumes.sh
slaperche-scality: Ha good to know, thanks for the pointer :slightly_smiling_face:

eve/create-volumes.sh
slaperche-scality: Yeah, especially if we are going to rewrite this in Python at some point :stuck_out_tongue:

slaperche-scality (approved):

Issue #1684, opened 2019-09-13
Use fake endpoint for container image registry


TeddyAndrieux:
**Component**:

'salt', 'containers'

**Why this is needed**:

For futur  HA bootstrap and being able to change registry without re-configuring/creating all pods

**What should be done**:

Use `metalk8s-registry-from-config.invalid` as registry and redirect this to the right IP in containerd config

gdemonet:
See the [relevant `containerd` documentation](https://github.com/containerd/cri/blob/master/docs/registry.md#configure-registry-endpoint).

Would it be useful to also add a catch-all mirror: `[plugins.cri.registry.mirrors."*"]`?

TeddyAndrieux:
> See the [relevant `containerd` documentation](https://github.com/containerd/cri/blob/master/docs/registry.md#configure-registry-endpoint).
> 
> Would it be useful to also add a catch-all mirror: `[plugins.cri.registry.mirrors."*"]`?

For me we shouldn't

NicolasT:
No we shouldn't. We *know* that what we deploy is in the local repo. Any MetalK8s user can also deploy things with containers from other repos (heck, Zenko today needs online repos). Having such catchall would basically make it impossible to install anything not part of what's in the offline repos... Not cool.

PR #1685, opened 2019-09-13
Use fake endpoint for registry name

TeddyAndrieux:
**Component**:

'salt', 'containers'

**Summary**:

Using registry name instead of ip:port for endpoint allow for futur
HA registry and able to change the registry without re-configuring and
restarting all pods

---

Fixes: #1684

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -112,14 +112,15 @@ def bootstrap_config(host):
 
 
 @pytest.fixture
-def registry_address(host):
+def registry_address(host, version):
     with host.sudo():
         registry_json = host.check_output(
-            "salt-call pillar.get metalk8s:endpoints:repositories --out json"
+            "salt-call --out json slsutil.renderer string='"
+            "{% from \"metalk8s/map.jinja\" import repo with context %}"
+            "{{ repo.registry_endpoint }}' "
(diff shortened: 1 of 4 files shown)


slaperche-scality (approved):
LGTM, but I haven't followed the whole story

tests/conftest.py
```diff
@@ -112,14 +112,15 @@ def bootstrap_config(host):
...
 
 @pytest.fixture
-def registry_address(host):
+def registry_address(host, version):
     with host.sudo():
         registry_json = host.check_output(
-            "salt-call pillar.get metalk8s:endpoints:repositories --out json"
+            "salt-call --out json slsutil.renderer string='"
+            "{% from \"metalk8s/map.jinja\" import repo with context %}"
+            "{{ repo.registry_endpoint }}' "
```
gdemonet, line 120: ~Can't we just read from `defaults.yaml`?~

EDIT: meh, in case we override this value in `map.jinja`... OK, fine with me.

gdemonet (approved):

tests/conftest.py
TeddyAndrieux: Yes btw I don't think we will override it in map.jinja one day but for me it's better to really get the final value :)

PR #1689, opened 2019-09-13
ui/bug: fix the bug when there is no PV && add unit test for isVolumeDeletable

### LINKED ISSUE #1673: Impossible to delete a volume in status failed
**Component**:
Volume Web GUI management

**What happened**:
I created a volume and the creation has failed. I would like to remove it using the GUI but, first the tips is empty and second when I click on the trash bin nothing happen.

If I list my current pv with kubectl it doesn't appeared in the list. The information seems to be only in the GUI db.

**What was expected**:
Being able to remove it to recreate it with the same name.

**Steps to reproduce**
Create a raw block device with a wrong device name it will fail.

ChengYanJin:
**Component**: ui

**Context**: 
When the volume is failed, if there is no PV found, the volume is not deletable.

**Summary**:
Fix the bug.
Add a unit test for `isVolumeDeletable` function in NodeVolumes. 
There are in total of 17 cases with the different combinations of volume and PV.

**Acceptance criteria**: 
Pass all the 17 cases

Closes: #1673

### DIFF
--- a/ui/src/containers/NodeVolumes.js
+++ b/ui/src/containers/NodeVolumes.js
@@ -13,6 +13,7 @@ import {
   sortCapacity,
   useRefreshEffect,
 } from '../services/utils';
+import { isVolumeDeletable } from '../services/NodeVolumesUtils';
 import {
   refreshVolumesAction,
   stopRefreshVolumesAction,
@@ -27,7 +28,6 @@ import {
   STATUS_FAILED,
   STATUS_AVAILABLE,
   STATUS_BOUND,
-  STATUS_RELEASED,
 } from '../constants';
 
 const VolumeTable = styled.div`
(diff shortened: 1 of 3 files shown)


ui/src/containers/NodeVolumes.js
```diff
@@ -119,7 +119,7 @@ const NodeVolumes = props => {
     setSortDirection(sortDirection);
   };
 
-  const isVolumeDeletable = rowData => {
+  const isVolumeDeletable = (rowData, persistentVolumes) => {
```
slaperche-scality: Why do we need this change?

Before we were already using `persistentVolumes` l.136, it wasn't working?

ui/src/containers/NodeVolumesUtils.js
```diff
@@ -0,0 +1,55 @@
...
+  STATUS_UNKNOWN,
+  STATUS_TERMINATING,
+  STATUS_PENDING,
+  STATUS_FAILED,
+  STATUS_AVAILABLE,
+  STATUS_BOUND,
+  STATUS_RELEASED,
+} from '../constants';
+
+export const isVolumeDeletable = (rowData, persistentVolumes) => {
```
slaperche-scality: This seems copied from `ui/src/containers/NodeVolumes.js`, is it normal to keep two versions of the same code?

If the goal is to make the function usable in the test, I guess `ui/src/containers/NodeVolumes.js` should use this version as well (so that the code tested is the one used for real).

carlito-scality: Where do you use it?

carlito-scality (changes requested):

ui/src/containers/NodeVolumesUtils.js
carlito-scality: As discussed before, please put it in 'services' folder instead of 'containers'.

ui/src/containers/NodeVolumesUtils.test.js
```diff
@@ -0,0 +1,256 @@
+import { isVolumeDeletable } from './NodeVolumesUtils';
+
+const testcase1 = {
```
slaperche-scality: We probably could have more meaningful name that `testcaseX`, `testcaseY`, …

- `testcase1` => `testcaseVolumeWithoutStatus`
- `testcase2` => `testcaseVolumePending`
- `testcase3` => `testcaseVolumeTerminating`
- `testcase4` => `testcaseVolumeFailedWithoutPv`
- `testcase5` => `testcaseVolumeFailedWithPvFailed`
- […]
- `testcase11` => `testcaseVolumeAvailableWithPvFailed`
- […]
- `testcase17` => testcaseVolumeAvailableWithoutPv

slaperche-scality (commented):
I think we could have 2 commits here:
- one for the bugfix
- one for the tests

Also, I don't know much about JS tests frameworks, but maybe there is some way to have less boilerplate here.

MonPote (approved):

carlito-scality (approved):

ChengYanJin:
@bert-e approve

PR #1694, opened 2019-09-16
ui/bug: add the title for breadcrumb

ChengYanJin:
**Component**: ui

**Context**:  
As a user, I want to know the whole name when the path is too long that automatically folded.

**Summary**:
When you hover the long path which reaches the max length, it will pop up the text to show the whole name

**Acceptance criteria**: 
image

### DIFF
--- a/ui/src/containers/VolumeInformation.js
+++ b/ui/src/containers/VolumeInformation.js
@@ -9,24 +9,24 @@ import { makeGetNodeFromUrl, makeGetVolumesFromUrl } from '../services/utils';
 import {
   SPARSE_LOOP_DEVICE,
   RAW_BLOCK_DEVICE,
-  STATUS_BOUND
+  STATUS_BOUND,
 } from '../constants';
 import {
   fetchVolumesAction,
   fetchPersistentVolumeAction,
-  fetchStorageClassAction
+  fetchStorageClassAction,
 } from '../ducks/app/volumes';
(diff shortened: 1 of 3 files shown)


ui/src/containers/VolumeInformation.js
```diff
@@ -73,10 +73,10 @@ const VolumeInformation = props => {
           activeColor={theme.brand.secondary}
           paths={[
             <StyledLink to="/nodes">{intl.messages.nodes}</StyledLink>,
-            <StyledLink to={`/nodes/${node.name}/volumes`}>
+            <StyledLink to={`/nodes/${node.name}/volumes`} title={node.name}>
               {node.name}
             </StyledLink>,
-            <BreadcrumbLabel>{match.params.volumeName}</BreadcrumbLabel>
+            <BreadcrumbLabel>{match.params.volumeName}</BreadcrumbLabel>,
```
carlito-scality: need ```title``` here too? I think you should add it to every items of the breadcrumbs

carlito-scality (changes requested):

ui/src/containers/VolumeInformation.js
ChengYanJin: Thanks. We should do that yes.

MonPote (approved):

carlito-scality (approved):

ChengYanJin:
@bert-e approve

PR #1702, opened 2019-09-17
#1693: Refactor versions listing for packages

### LINKED ISSUE #1693: Refactor versions
**Component**: build 

**Why this is needed**: We need it to have a code more readable and to have a better Debian integration

**What should be done**: refactor version.py in the buildchain

**Test plan**:

jbertran:
**Component**:

**Context**: We're adding a new OS, so we need OS-aware versions definition.

**Summary**:

The goal is to provide an authoritative listing of required packages that it as easy as possible to manage.

```
# Common packages are packages for which we need not care about OS-specific
# divergences.
#
# In this case, either:
#   * the _latest_ version is good enough, and will be the one
#     selected by the package managers (so far: apt and yum).
#   * we have strict version requirements that span OS families, and the
#     versioning schemes _and_ package names do not diverge
#
# Strict version requirements are notably:
#   * kubelet and kubectl which _make_ the K8s version of the cluster
#   * salt-minion which _makes_ the Salt version of the cluster
#
# These common packages may be overridden by OS-specific packages if package
# names or versioning conventions diverge.
#
# Packages that we build ourselves require a version and release as part of
# their build process.
```

**Acceptance criteria**: tests pass, debian files are generated with the correct versions.

---

Closes: #1693

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -41,6 +41,30 @@ from buildchain import utils
 from buildchain import versions
 
 
+# Utilities {{{
+
+def _list_packages_to_build(
+    pkg_cats: Dict[str, Tuple[targets.Package, ...]]
+) -> List[str]:
+    return [
+        pkg.name for pkg_list in pkg_cats.values() for pkg in pkg_list
+    ]
+
+
+def _list_packages_to_download(
+    package_versions: Tuple[versions.PackageVersion, ...],
+    packages_to_build: List[str]
+) -> Dict[str,str]:
+    return {
+        pkg.name: "{p.version}-{p.release}".format(p=pkg)
+        for pkg in package_versions
+        if pkg.name not in packages_to_build
+    }
+
+
+# }}}
+# Tasks {{{
+
 def task_packaging() -> types.TaskDict:
     """Build the packages and repositories."""
     return {
@@ -202,6 +226,10 @@ def task__build_rpm_repositories() -> Iterator[types.TaskDict]:
     for repository in RPM_REPOSITORIES:
         yield from repository.execution_plan
 
+
+# }}}
+# Builders {{{
+
 # Image used to build the packages
 RPM_BUILDER : targets.LocalImage = targets.LocalImage(
     name='metalk8s-rpm-build',
@@ -239,6 +267,8 @@ DEB_BUILDER : targets.LocalImage = targets.LocalImage(
     task_dep=['_build_root'],
 )
 
+# }}}
+# RPM packages and repository {{{
 
 # Packages to build, per repository.
 def _rpm_package(name: str, sources: List[Path]) -> targets.RPMPackage:
@@ -262,27 +292,6 @@ def _rpm_package(name: str, sources: List[Path]) -> targets.RPMPackage:
         task_dep=['_package_mkdir_rpm_root', '_build_rpm_container'],
     )
 
-
-def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
-    try:
-        pkg_info = versions.PACKAGES_MAP[name]
-    except KeyError:
-        raise ValueError(
-            'Missing version for package "{}"'.format(name)
-        )
-
-    # In case the `release` is of form "{build_id}.{os}", which is standard
-    build_id_str, _, _ = pkg_info.release.partition('.')
-
-    return targets.DEBPackage(
-        basename='_build_deb_packages',
-        name=name,
-        version=pkg_info.version,
-        build_id=int(build_id_str),
-        sources=sources,
-        builder=DEB_BUILDER,
-        task_dep=['_package_mkdir_deb_root', '_build_deb_container'],
-    )
 
 # Calico Container Network Interface Plugin.
 CALICO_RPM = _rpm_package(
@@ -308,6 +317,58 @@ RPM_TO_BUILD : Dict[str, Tuple[targets.RPMPackage, ...]] = {
     ),
 }
 
+
+_RPM_TO_BUILD_PKG_NAMES : List[str] = _list_packages_to_build(RPM_TO_BUILD)
+
+# All packages not referenced in `RPM_TO_BUILD` but listed in
+# `versions.RPM_PACKAGES` are supposed to be downloaded.
+RPM_TO_DOWNLOAD : FrozenSet[str] = frozenset(
+    "{p.name}-{p.version}-{p.release}".format(p=package)
+    for package in versions.RPM_PACKAGES
+    if package.name not in _RPM_TO_BUILD_PKG_NAMES
+)
+
+
+# Store these versions in a dict to use with doit.tools.config_changed
+_TO_DOWNLOAD_RPM_CONFIG : Dict[str, str] = _list_packages_to_download(
+    versions.PACKAGES,
+    _RPM_TO_BUILD_PKG_NAMES
+)
+
+
+SCALITY_RPM_REPOSITORY = targets.RPMRepository(
+    basename='_build_rpm_repositories',
+    name='scality',
+    builder=RPM_BUILDER,
+    packages=RPM_TO_BUILD['scality'],
+    task_dep=['_package_mkdir_rpm_iso_root'],
+)
+
+# }}}
+# Debian packages and repositories {{{
+
+def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
+    try:
+        pkg_info = versions.DEB_PACKAGES_MAP[name]
+    except KeyError:
+        raise ValueError(
+            'Missing version for package "{}"'.format(name)
+        )
+
+    # In case the `release` is of form "{build_id}.{os}", which is standard
+    build_id_str, _, _ = pkg_info.release.partition('.')
+
+    return targets.DEBPackage(
+        basename='_build_deb_packages',
+        name=name,
+        version=pkg_info.version,
+        build_id=int(build_id_str),
+        sources=sources,
+        builder=DEB_BUILDER,
+        task_dep=['_package_mkdir_deb_root', '_build_deb_container'],
+    )
+
+
 DEB_TO_BUILD : Dict[str, Tuple[targets.DEBPackage, ...]] = {
     'scality': (
         # SOS report custom plugins.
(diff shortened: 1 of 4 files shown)


buildchain/buildchain/versions.py
```diff
@@ -363,6 +363,106 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
 PACKAGES_MAP = {pkg.name: pkg for pkg in RPM_PACKAGES}
 
 
+DEB_PACKAGES = (
```
slaperche-scality: This commit should probably be removed (`DEB_PACKAGES` is already defined l.466)

buildchain/buildchain/packaging.py
```diff
@@ -324,18 +324,27 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
...
+    }
+
 # Store these versions in a dict to use with doit.tools.config_changed
-_TO_DOWNLOAD_CONFIG : Dict[str, str] = {
-    pkg.name: "{p.version}-{p.release}".format(p=pkg)
-    for pkg in versions.RPM_PACKAGES
-    if pkg.name not in _RPM_TO_BUILD_PKG_NAMES
-}
+_TO_DOWNLOAD_RPM_CONFIG : Dict[str, str] = _list_packages_to_download(
+    versions.PACKAGES,
```
slaperche-scality: minor: can be wrapped:

```python
_TO_DOWNLOAD_RPM_CONFIG : Dict[str, str] = _list_packages_to_download(
    versions.PACKAGES, _RPM_TO_BUILD_PKG_NAMES
)
```

ditto for `_TO_DOWNLOAD_DEB_CONFIG`

buildchain/buildchain/packaging.py
```diff
@@ -324,18 +324,27 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
...
+    _RPM_TO_BUILD_PKG_NAMES
+)
 
-_TO_DOWNLOAD_DEB_CONFIG : Dict[str, str] = {
-    pkg.name: "{p.version}-{p.release}".format(p=pkg)
-    for pkg in versions.DEB_PACKAGES
-    if pkg.name not in _DEB_TO_BUILD_PKG_NAMES
-}
+# Store these versions in a dict to use with doit.tools.config_changed
```
slaperche-scality, line 331: Not sure we want to duplicate this comment

buildchain/buildchain/packaging.py
```diff
@@ -308,12 +308,32 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
...
+            sources=SCALITY_RPM_REPOSITORY.get_rpm_path(CALICO_RPM)
+        ),
+    )
+}
+
+def _list_packages_to_build(
+    pkg_cats: Dict[str, Tuple[targets.Package, ...]]
+) -> List[str]:
+    return [
+        pkg.name for pkg_list in pkg_cats.values()
```
slaperche-scality: Can be squashed:

```python
    return [
        pkg.name for pkg_list in pkg_cats.values() for pkg in pkg_list
    ]
```

buildchain/buildchain/packaging.py
```diff
@@ -202,6 +227,12 @@ def task__build_rpm_repositories() -> Iterator[types.TaskDict]:
     for repository in RPM_REPOSITORIES:
         yield from repository.execution_plan
 
+
+# }}}
+
```
slaperche-scality: Remove empty lines between folds

buildchain/buildchain/packaging.py
```diff
@@ -239,6 +270,10 @@ def task__build_rpm_repositories() -> Iterator[types.TaskDict]:
     task_dep=['_build_root'],
 )
 
+# }}}
+
```
slaperche-scality: ditto

buildchain/buildchain/packaging.py
```diff
@@ -381,6 +349,65 @@ def _list_packages_to_download(
     task_dep=['_package_mkdir_rpm_iso_root'],
 )
 
+# }}}
+
```
slaperche-scality: ditto

buildchain/buildchain/packaging.py
```diff
@@ -328,8 +327,8 @@ def _rpm_package(name: str, sources: List[Path]) -> targets.RPMPackage:
 # All packages not referenced in `RPM_TO_BUILD` but listed in
 # `versions.RPM_PACKAGES` are supposed to be downloaded.
 RPM_TO_DOWNLOAD : FrozenSet[str] = frozenset(
-    "{p.name}-{p.version}-{p.release}".format(p=package)
-    for package in versions.RPM_PACKAGES
+    package.rpm_full_name
+    for package in versions.PACKAGES
```
slaperche-scality: Why is `RPM_PACKAGES` renamed `PACKAGES`?

buildchain/buildchain/versions.py
```diff
@@ -364,200 +425,65 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
 
 
 DEB_PACKAGES = (
-    Package(
+    PackageVersion(
```
slaperche-scality: Save some screen real estate and squash into a single line: `PackageVersion(name='containerd'),` (ditto for the others)

slaperche-scality (changes requested):

buildchain/buildchain/versions.py
jbertran: Done in a following commit

buildchain/buildchain/versions.py
jbertran: Oops, indeed. Rebase gone wrong

buildchain/buildchain/packaging.py
jbertran: Gah, stray rebase conflict resolution, thank you

buildchain/buildchain/packaging.py
```diff
@@ -1,6 +1,4 @@
 # coding: utf-8
-
```
slaperche-scality: Why?

Also, it become inconsistent with the other files.

buildchain/buildchain/versions.py
```diff
@@ -228,132 +228,193 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
+        Arguments:
+            name: the name of the package
+            version: the version of the package
+            release: the release of the package
+        """
+        self._name = name
+        self._version = version
+        self._release = release
+
+    name = property(operator.attrgetter('_name'))
```
slaperche-scality, line 261: minor: would align

```
    name    = property(operator.attrgetter('_name'))
    version = property(operator.attrgetter('_version'))
    release = property(operator.attrgetter('_release'))
```

buildchain/buildchain/versions.py
```diff
@@ -228,132 +228,193 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
+        if self.version:
+            full_version = self.version
+            if self.release:
+                full_version = '{}-{}'.format(self.version, self.release)
+        return full_version
+
+    @property
+    def rpm_full_name(self) -> str:
+        """The package's full name in RPM conventions."""
+        name_fmt = '{p.name}'
```
slaperche-scality: could it be rewritten:

```python
if self.version:
    return '{}-{}'.format(self.name, self.full_version)
return self.name
```
?

buildchain/buildchain/versions.py
```diff
@@ -228,132 +228,193 @@ def _version_prefix(version: str, prefix: str = 'v') -> str:
...
+        if self.version:
+            name_fmt += '-{p.version}'
+            if self.release:
+                name_fmt += '-{p.release}'
+        return name_fmt.format(p=self)
+
+    @property
+    def deb_full_name(self) -> str:
```
slaperche-scality, line 284: ditto, could we do something like:

```
if self.version:
    return '{}={}'.format(self.name, self.full_version)
return self.name
```
?

buildchain/buildchain/versions.py
```diff
@@ -291,201 +294,113 @@ def deb_full_name(self) -> str:
...
-    ),
-)
+    Arguments:
+        os_family: OS_family for which to list packages
+    """
+    common_pkgs = PACKAGES['common']
+    os_family_pkgs = PACKAGES.get(os_family)
+
+    if os_family_pkgs is None:
+        raise Exception('Unlisted OS family: {}'.format(os_family))
```
slaperche-scality: s/Unlisted/Unsupported/?

buildchain/buildchain/packaging.py
```diff
@@ -52,7 +52,7 @@ def _list_packages_to_build(
 def _list_packages_to_download(
     package_versions: Tuple[versions.PackageVersion, ...],
     packages_to_build: List[str]
-) -> Dict[str,str]:
+) -> Mapping[str, Optional[str]]:
```
slaperche-scality: Why the change here? (and BTW, why the change in general: a commit message would be welcome).

IIRC, the good practice is to use "precise" type for return type and "abstract" one for input.

See https://docs.python.org/3/library/typing.html#typing.Dict:

> Useful for annotating return types. To annotate arguments it is preferred to use an abstract collection type such as Mapping.

buildchain/buildchain/packaging.py
```diff
@@ -346,6 +339,61 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
     task_dep=['_package_mkdir_rpm_iso_root'],
 )
 
+# }}}
+# Debian packages and repositories {{{
+
+def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
+    try:
+        pkg_info = versions.RPM_PACKAGES_MAP[name]
```
slaperche-scality: `_deb_package` using `RPM_PACKAGES_MAP`?

slaperche-scality (changes requested):

buildchain/buildchain/packaging.py
jbertran: Oh boy that's bad.

buildchain/buildchain/versions.py
jbertran: Oh indeed, that's basic DRY. Thanks.

buildchain/buildchain/versions.py
jbertran: Eeeeh technically we don't support Ubuntu yet but we download its packages. As `versions` is supposed to be absolutely standalone I'd rather observe things that pertain to it and only to it, hence unlisted in the context of the package dict keys.

buildchain/buildchain/versions.py
jbertran: Damn that triggers my linter like crazy. Done.

buildchain/buildchain/versions.py
slaperche-scality: Make sense, but I think we can have a better word than `unlisted`.
Maybe `no packages for OS family {}`?

jbertran: Suits me, thanks

buildchain/buildchain/packaging.py
```diff
@@ -27,7 +27,7 @@
 
 
 from pathlib import Path
-from typing import Dict, FrozenSet, Iterator, List, Tuple
+from typing import Dict, FrozenSet, Iterator, List, Mapping, Optional, Tuple
```
slaperche-scality, line 30: Still no explanation on why this switch in the commit message

slaperche-scality (approved):

Issue #1696, opened 2019-09-16
Build the needed ISO to perform upgrade/downgrade tests


sayf-eddine-scality:
**Component**:
upgrade, downgrade, tests

**Why this is needed**:
We need the N and the N-1 ISOs images in order to run upgrade or downgrade tests.

**What should be done**:
Build both ISOs in the build step and provide them as artifacts

**Implementation proposal** (strongly recommended):
Add an eve step to checkout the respective branches we would want to test then build a Metalk8s ISO out of each branch.
We should end up with two separate directories/filenames having two different Metalk8s versions based on the branches we have checked out.
Two approaches recommended:

- Build the ISO images by checking out the respective branches in parallel (strongly recommended)
- Retrieve the ISO image from artefact(basing on promoted builds)
*UPDATE*: retreiving the ISO from the artifacts is not a viable option if a PR is backported into older branches, the ISO wont be available until all the integration branches finish the build which is incrementally long

**Test plan**:
Manually check the ISOs versions, SHA256sum checksums and the product.txt

PR #1704, opened 2019-09-17
Build the MetalK8s version N and N-1 ISO

sayf-eddine-scality:
Fixes: #1696

**Component**:
upgrade, downgrade, tests

**Context**: 
to be able to launch upgrade e2e tests we need the ISO of the version N-1
**Summary**:
The version N-1 is built and stored in the artifacts
**Acceptance criteria**: 
Check the artifacts for `pre/metalk8s-xxxx-dev.iso`

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -14,6 +14,7 @@ stages:
           name: Trigger build, docs and lint stages simultaneously
           stage_names:
             - build
+            - buildprev
             - docs
             - lint
           haltOnFailure: true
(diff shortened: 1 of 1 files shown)


NicolasT:
I hope, once we have 'released' versions, we can use promoted images instead...

sayf-eddine-scality:
> I hope, once we have 'released' versions, we can use promoted images instead...

Yes that would be so useful instead of doing the guess work on the branch name

Ebaneck:
LGTM! ..Waiting for green build

gdemonet:
> > I hope, once we have 'released' versions, we can use promoted images instead...
> 
> Yes that would be so useful instead of doing the guess work on the branch name

Mmh IMO you still need to rely on tags (from `git`) to find out where such artifacts would be. The issue here is that we're guessing the branch since we lack some tags

Ebaneck (approved):

eve/main.yml
```diff
@@ -77,6 +78,68 @@ stages:
...
+              sleep 2
+            done
+            echo "Could not reach Docker daemon from buildbot worker" >&2
+            exit 1'
+          haltOnFailure: true
+      - SetPropertyFromCommand:
+          name: set product_version_prev
+          property: product_version_prev
+          command: >
+              major=$(echo "%(prop:product_version)s" | cut -d'.' -f1) &&
```
alexandre-allard, line 104: Details: prefer the herestring syntax rather than using `echo |` which is ugly.
```
major=$(cut -d'.' -f1 <<< "%(prop:product_version)s")
```
And, btw `&&` are useless since there is no chance that your cut command will fail, you can use simple `;`.

sayf-eddine-scality: ups will fix this in the next PRs

Issue #1706, opened 2019-09-17
Bootstrap still need to use hostname as node_name


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**What happened**:

We use `hostname` to mark the control plane node so the node_name need to be the hostname

**What was expected**:

Do not need to use hostname as node_name and minion_id

**Steps to reproduce**

Run a bootstrap with minion_id different not equal to hostname

**Resolution proposal** (optional):

Use node name instead of hostname in `salt/metalk8s/kubernetes/mark-control-plane/deployed.sls.in`

NicolasT:
Ref #1700

PR #1709, opened 2019-09-17
Fix iso-manager script

gdemonet:
We forgot to remove the extra `metalk8s` level during renaming from
`products:metalk8s` to `archives`.

See #1332
See 1975ec5
See 6eccd63


sayf-eddine-scality (dismissed):

slaperche-scality (approved):

gdemonet:
Closing, will tackle another issue (#1706) at the same time to reduce build delays.

PR #1711, opened 2019-09-17
Add PV templating to Volume

### LINKED ISSUE #1500: Add template field in Volume to customize PV creation
Likely to be done later, but we should either have some kind of `template` in the `Volume` spec such that one can define `label`s of the `PersistentVolume` that get created (and other fields like `annotations`), as done in the `podSpec` of a `Deployment`, for example, or we should copy over said values from the `Volume` as-is.

_Originally posted by @NicolasT in https://github.com/scality/metalk8s/pull/1470_

slaperche-scality:
**Component**:

operator

**Context**: 

We want to be able to "customize" the underlying PV.
See #1515

**Summary**:

Add a new field in the Volume CRD, `template`, which allows to specify meta and spec for the underlying PV.
Those fields are used as base and overwritten according to our needs (`name`, `finalizers`, …).

Since it's an optional field, the backward compat is preserved and we don't need to bump the version (I did a manual test for that).

**Acceptance criteria**: 

Tests are passing (I updated an existing tests to check for labels)

---

Closes: #1500


storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -809,32 +809,31 @@ func newPersistentVolume(
 		)
```
jbertran, line 809: minor: `s/templace/template` in commit message

slaperche-scality: Done

jbertran (dismissed):

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -809,32 +809,31 @@ func newPersistentVolume(
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:       volume.Name,
-			Labels:     map[string]string{},
-			Finalizers: []string{VOLUME_PROTECTION},
+	pv := corev1.PersistentVolume{
+		ObjectMeta: volume.Spec.Template.Metadata,
+		Spec:       volume.Spec.Template.Spec,
+	}
+	pv.ObjectMeta.Name = volume.Name
+	pv.ObjectMeta.Finalizers = []string{VOLUME_PROTECTION}
```
NicolasT: I think we may want to add a finalizer rather than overwrite them.

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -809,32 +809,31 @@ func newPersistentVolume(
...
+	pv.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{
+		corev1.ReadWriteOnce,
+	}
+	pv.Spec.Capacity = map[corev1.ResourceName]resource.Quantity{
+		corev1.ResourceStorage: volumeSize,
+	}
+	pv.Spec.MountOptions = storageClass.MountOptions
```
NicolasT, line 826: Overwrite or merge? Not sure.

NicolasT (changes requested):
Minor questions

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: yeah, good point: I'll do that.

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: If you can specify the `MountOptions` directly in the PV, it kinds of defeat the purpose of `StorageClass` no?

I mean, one seeing a `Volume` with the `StorageClass` `Foo` will expect it to have the same mount options as the others `Volume` using that `StorageClass`.

I see it like `class` and `object` in languages.
I expect all the object instantiated from the class `Foo` to have the same interface. Now, sure, I can monkey-patch one specific instance but that's usually confusing and frowned upon (save for a few use case I guess).

Knowing it's always easier to ease a restrictions later than to reinforce it, I would go with `overwrite` now and if one day we have a good use case for the "monkey-patching" approach we can reconsider it, no?

NicolasT: Agree. Over time we could maybe consider not enforcing a `StorageClass` in the `Volume`, only 'use' it if it's specified in the `PVTemplate`, and if none is specified, create a classless PV.

So, for now, all good, no need to change, and if we ever want to do the above change, will require a version bump.

NicolasT (dismissed):

slaperche-scality:
Thanks for the approval.

I'll have to rebase (and lose it) though, to get the thrift fix (I'm stuck in the CI because of that one ^^)

NicolasT (approved):
I'll approve, for now. However, there's another thing to consider:

We currently use the provided template when the PV is constructed. However, e.g. in the Pod template in the spec of a Deployment, when changing this, eventually the Pods will be changed as well.

So, in theory, one could argue that, when e.g. the labels in the PV template in a Volume are changed, then the labels on the resulting PV ought to be updated as well.

Not sure that's super useful and a must-have, but something to consider/track.

slaperche-scality:
> I'll approve, for now. However, there's another thing to consider:
> 
> We currently use the provided template when the PV is constructed. However, e.g. in the Pod template in the spec of a Deployment, when changing this, eventually the Pods will be changed as well.
> 
> So, in theory, one could argue that, when e.g. the labels in the PV template in a Volume are changed, then the labels on the resulting PV ought to be updated as well.
> 
> Not sure that's super useful and a must-have, but something to consider/track.

Tracked with the more general issue #1718

PR #1712, opened 2019-09-17
Pre `2.4.0-beta1` release bugfixes

### LINKED ISSUE #1706: Bootstrap still need to use hostname as node_name

gdemonet:
Fix some small bugs required for the `2.4.0-beta1`:

- Remove the extra `metalk8s` level in `iso-manager.sh` script when adding archives
- Fix #1706 (`mark-control-plane` relying on hostname instead of `pillar.bootstrap_id` to accurately set labels and annotations on the good `Node` object)

### DIFF
--- a/salt/metalk8s/kubernetes/mark-control-plane/deployed.sls.in
+++ b/salt/metalk8s/kubernetes/mark-control-plane/deployed.sls.in
@@ -1,4 +1,4 @@
-{% set hostname = pillar.get('mark_control_plane_hostname', salt['network.get_hostname']()) %}
+{% set node_name = pillar.bootstrap_id %}
 
 {% set kubeconfig = "/etc/kubernetes/admin.conf" %}
 {% set context = "kubernetes-admin@kubernetes" %}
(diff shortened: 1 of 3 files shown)


slaperche-scality (dismissed):

packages/Dockerfile
```diff
@@ -8,6 +8,9 @@ ARG SALT_VERSION
 ADD yum_repositories/*.repo /etc/yum.repos.d/
 RUN sed -i s/@SALT_VERSION@/$SALT_VERSION/ /etc/yum.repos.d/saltstack.repo
 
+# FIXME: CentOS Vault is broken...
```
slaperche-scality: We probably want an issue (which explains the context & so on) to track that (build succeed so the fix works).

slaperche-scality: Hum, but it breaks the bootstrap :sweat_smile:

slaperche-scality: Ok, seems like the issues resolved itself so no need for a workaround on our side.

slaperche-scality:
Rebased

packages/Dockerfile
gdemonet: ~I'll remove the commit~ Oh you did it already, thanks!

sayf-eddine-scality (approved):

Issue #1100, opened 2019-05-10
Simplify `salt/_renderers/metalk8s_kubernetes.py` module


NicolasT:
All these states require a `name`, which is always retrieved with `.metadata.name`. Same for `kubeconfig` and `context`.
Maybe it would be interesting to have some helpers to generate them, and mutualize the common parts... Probably not in this PR.

_Originally posted by @gdemonet in https://github.com/scality/metalk8s/pull/1098_

Issue #1099, opened 2019-05-10
Set kubeconfig/context in Salt master config


gdemonet:
**Component**: salt

**Why this is needed**:

Currently, for each state calling `metalk8s_kubernetes`, we explicitly pass values for `kubeconfig` and `context`. The renderer defined in #1092 and modified in #1098 also accepts these arguments to later set them in the states definitions.

This leads to:

- repetition
- hardcoded values (what if we change the kubeconfig file to use for Salt master?)
- weird shebang for the renderer

**What should be done**: Set the `kubeconfig`/`context` globally to avoid specifying them in each state definition.

**Implementation proposal**:

The `metalk8s_kubernetes` (taken from the upstream `kubernetes`) execution module allows one to set `kubeconfig`/`context` in a pillar or in the master/minion configuration.
We'd like the `metalk8s.kubernetes.apiserver.kubeconfig` formula to create/update a master configuration file, setting the `kubernetes.kubeconfig` and `kubernetes.context` options.

**Test plan**: Irrelevant, it's an implementation detail.

PR #1713, opened 2019-09-17
[WIP] Improve Salt management of Kubernetes objects

### LINKED ISSUE #1046: Kubernetes python API error handling
In `metalk8s/salt/_modules/metalk8s_kubernetes.py`, errors when calling Kubernetes API are - mostly - handled this way:
```
    try:
        api_instance = kubernetes.client.CoreV1Api()
        api_response = api_instance.list_node()

        # ... do something with response ...

        if isinstance(exc, ApiException) and exc.status == 404:
            return None
        else:
            log.exception('Exception when calling CoreV1Api->list_node')
``` 
So, basically the only error handled is a HTTP 404, with a big assumption it is returned when Kubernetes does not find the object requested... IMHO this is wrong, and we should get a 404 when we call an API that does not exist (so it should be considered a fatal error), and by experiencing with different calls to Kubernetes API it looks like when object is not found a valid response is sent (aka. no HTTP error), but the `response.items` array is just empty.

### LINKED ISSUE #1047: Kubernetes python API calls and reponse processing
In many places in `metalk8s/salt/_modules/metalk8s_kubernetes.py`, calls and response processing are sub-optimal, for example:
```
def nodes(**kwargs):
    cfg = _setup_conn(**kwargs)
    try:
        api_instance = kubernetes.client.CoreV1Api()
        api_response = api_instance.list_node()

        return [k8s_node['metadata']['name'] for k8s_node in api_response.to_dict().get('items')]
```
there is no need to convert to dict(), and the returned value should be:
`return [k8s_node.metadata.name for k8s_node in api_response.items]`
because even when nothing is found a valid reponse - with an empty `response.items` array is returned, and no error is fired.

In a similar way:
```
def node(name, **kwargs):
    cfg = _setup_conn(**kwargs)
    try:
        api_instance = kubernetes.client.CoreV1Api()
        api_response = api_instance.list_node()
        if isinstance(exc, ApiException) and exc.status == 404:
            return None
        else:
            log.exception('Exception when calling CoreV1Api->list_node')
    finally:
        _cleanup(**cfg)

    for k8s_node in api_response.items:
        if k8s_node.metadata.name == name:
            return k8s_node.to_dict()

    return None
```
We loop on all the returned nodes to find out the one we want, whereas we could have use `field_selector` and let Kubernetes do the job:
`api_response = api_instance.list_node(field_selector='metadata.name={0}'.format(name))`
then the `api_response.items` contains only the targeted node, or nothing... (extra question: do we need/is it better to convert response to dict(), or directly use the node object returned ?)

Note: those are just examples, and the same patterns are used in many different places in the file.

### LINKED ISSUE #1048: Kubernetes Salt execution/state modules refactoring
The current `metalk8s_kubernetes` execution module is basically a fork of the upstream [`kubernetes` module from Saltstack](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.kubernetes.html).

It is however filled with bad practices, and needs refactoring:

- [x] to ease maintenance and improve readability
- [x] to reduce code duplication
- [x] to reduce ineffective API calls
- [x] to have better error management

Here are all the tickets related to this task, with additional details:

- #1047 
- #1046

### LINKED ISSUE #1183: Remove compatibility band-aid in Secret handler
**Component**: salt

**Why this is needed**:

Currently we vendor the monitoring components as upstreamed, autogenerated salt states using a custom renderer. Using this renderer to glue the manifests with the k8s salt module means that all secrets **should** be plain text to be used by the salt module (which doesn't necessarily make a lot of sense).

This means that we have a compatibility patch on the renderer for Secrets for the sole purpose of passing data for Secrets to the salt module.

**What should be done**:

Remove the need for this layer.

Ideally, this takes place during a refactor of the whole salt module.

**Test plan**:

No more secrets in plain text, isofunctional monitoring stack otherwise.

gdemonet:
**Component**: salt

**Context**: to do

**Summary**: to do

**Acceptance criteria**:

- Everything keeps working
- A new suite of tests should validate the execution/state modules introduced by this PR

TODO:

- [ ] Finish to write the execution module (connection setup and extra methods for drain/cordon)
- [x] Rename everything to overwrite the previous `metalk8s_kubernetes` modules
- [ ] Use these modules instead of the previous ones everywhere (not only through the renderer)
- [ ] Check that bootstrap/expansion/upgrade still work
- [ ] Write some tests to check both standard and custom objects can be manipulated
- [ ] Handle #1100 and #1099
- [ ] Clean-up commits

Fixes: #1048 
Fixes: #1047
Fixes: #1046
Fixes: #1183


salt/_states/better_kubernetes.py
```diff
@@ -0,0 +1,87 @@
...
+    return ret
+
+
+def object_present(name, kubeconfig, context, manifest):
+    """Ensure that the object is present."""
+    ret = {'name': name, 'changes': {}, 'result': False, 'comment': ''}
+
+    obj = __salt__['metalk8s_kubernetes.get_object'](
+        name, kubeconfig, context, manifest
+    )
```
TeddyAndrieux: IMO we really want to handle idempotancy

TeddyAndrieux (commented):
Good job it's really simpler than before

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _extract_kwargs(obj, details):
+    kwargs = {'name': obj.metadata.name}
+
+    if details.scope == 'namespaced':
+        kwargs['namespace'] = obj.metadata.namespace
+
+    if hasattr(details, 'group'):
+        # This object represents a CustomResource
```
NicolasT: Why? All K8s objects belong to a 'group'?

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+    kwargs = _extract_kwargs(obj, details)
+
+    try:
+        result = details.retrieve(**kwargs)
+    except Exception as exc:  # TODO: which Exception could be raised, and why?
+        raise CommandExecutionError(
+            'Failed to retrieve object: {!s}'.format(exc)
+        )
+
+    if hasattr(result, 'to_dict'):
```
NicolasT: Maybe put a comment in which case this attribute *doesn't* exist, and why we're sure `result` is in that case a valid `dict` already.

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+        # TODO: check if useful, probably for CRs, could be nice to clean
+        result = result.to_dict()
+
+    return result
+
+
+def delete_object(manifest, kubeconfig, context):
+    obj, details = _extract_obj_and_details(manifest)
+
+    kwargs = {'name': obj.metadata.name, 'body': k8s_client.V1DeleteOptions()}
```
NicolasT: Overlap with `extract_kwargs` in these lines...

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+        })
+
+    try:
+        details.delete(**kwargs)
+    except Exception as exc:
+        raise CommandExecutionError(
+            'Failed to delete object: {!s}'.format(exc)
+        )
+
+    return None
```
NicolasT: Remove

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+    if hasattr(result, 'to_dict'):
+        # TODO: check if useful, probably for CRs, could be nice to clean
+        result = result.to_dict()
+
+    return result
+
+
+def create_object(name, kubeconfig, context, manifest):
```
NicolasT: There's a bit of duplication in these functions, could abstract them out (e.g., `create_object = _object_manipulation_function('create')`, etc).

salt/_renderers/better_kubernetes.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _step(manifest, kubeconfig=None, context=None, absent=False):
+    """Render a single Kubernetes object into a state 'step'."""
+    try:
+        assert __utils__['kubernetes.validate_manifest'](manifest)
+    except (AssertionError, ValueError) as exc:
+        raise SaltRenderError('Invalid manifest: {!s}'.format(exc))
+
+    obj = __utils__['kubernetes.convert_manifest_into_object'](manifest)
```
NicolasT: This is already done by `validate_manifest`, maybe this could return this as well? Not sure what such API would look like though.

salt/_states/better_kubernetes.py
```diff
@@ -0,0 +1,87 @@
...
+Those will then simply delegate all the logic to the `metalk8s_kubernetes`
+execution module, only managing simple dicts in this state module.
+"""
+from salt.exceptions import CommandExecutionError
+
+__virtualname__ = 'metalk8s_kubernetes'
+
+
+def __virtual__():
+    if 'metalk8s_kubernetes.ping' not in __salt__:
```
NicolasT: I'd expect this module to (also?) require the `better_kubernetes` module, though we'll want to rename those ;-)

salt/_states/better_kubernetes.py
```diff
@@ -0,0 +1,87 @@
...
+    if obj is None:
+        try:
+            __salt__['metalk8s_kubernetes.create_object'](
+                name, kubeconfig, context, manifest
+            )
+        except CommandExecutionError as exc:
+            ret['comment'] = 'Something went wrong: {}'.format(exc)
+        else:
+            ret['result'] = True
+            ret['changes'] = {'old': 'absent', 'new': 'absent'}
```
NicolasT: This ain't right.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+
+try:
+    import kubernetes.client as k8s_client
+    from kubernetes.client.apis import CustomObjectsApi
+except ImportError:
+    HAS_LIBS = False
+else:
+    HAS_LIBS = True
+
+__virtualname__ = 'kubernetes'
```
NicolasT: I think we should namespace this as well, in case upstream decides to create a `kubernetes` utils thing at some point.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+    CRDetails('monitoring.coreos.com/v1', 'PrometheusRule',
+              scope='namespaced', plural='prometheusrules'),
+    CRDetails('monitoring.coreos.com/v1', 'ServiceMonitor',
+              scope='namespaced', plural='servicemonitors'),
+]
+
+KNOWN_CUSTOM_OBJECTS = {details.key: details for details in _CUSTOM_OBJECTS}
+
+
+class _DictWrapper(object):
```
NicolasT, line 397: This needs some more documentation (also `CustomObject`) IMHO as to what the intent is and how it's achieved. Not immediately obvious.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+        )
+
+    object_details = KNOWN_CUSTOM_OBJECTS.get(
+        (api_version, kind),
+        KNOWN_STD_OBJECTS.get((api_version, kind))
+    )
+
+    if object_details is None:
+        raise ValueError(
+            'Unknown object type provided: {}/{}. Make sure that it is'
```
NicolasT:
```suggestion
            'Unknown object type provided: {}/{}. Make sure it is'
```

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            return super(CustomObject, self).__getattribute__(name)
+        except AttributeError:
+            return getattr(self._attr_dict, name)
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def get_object_details(manifest):
```
NicolasT: Nit: the naming is a bit confusing/conflates two concepts: it *takes* a `manifest`, which is an object (think 'instance'), but it doesn't return any kind of 'details' about this object, instead, it returns *meta-information* about the *type* of the object.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+DICT_PATTERN = re.compile(r'^dict\(str,\s?(?P<value_type>\S+)\)$')
+LIST_PATTERN = re.compile(r'^list\[(?P<value_type>\S+)\]$')
+
+
+def _cast_value(value, type_string):
+    """Attempt to cast a value given a type declaration as a string."""
+    if type_string == 'str':
+        if not isinstance(value, str):
```
NicolasT: Should we be forward-compatible and use the types from `salt.whatever.six`?

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            raise _type_error(value, expected='a string')
+        return value
+
+    if type_string == 'bool':
+        if not isinstance(value, bool):
+            raise _type_error(value, expected='a boolean')
+        return value
+
+    if type_string == 'int':
+        if not isinstance(value, int):
```
NicolasT: What about `long`s?

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            raise _type_error(value, expected='an integer')
+        return value
+
+    if type_string == 'float':
+        if not isinstance(value, float) or isinstance(value, int):
```
NicolasT: What about `long`s?

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            raise _type_error(value, expected='a date-time string')
+
+    dict_match = DICT_PATTERN.match(type_string)
+    if dict_match is not None:
+        if not isinstance(value, dict):
+            raise _type_error(value, expected='a dictionary')
+
+        value_type_str = dict_match.group('value_type')
+        return {
+            key: _cast_value(val, value_type_str)
```
NicolasT, line 631: Should assert the key is a string.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            for key, value in data.items()
+        }
+
+    if isinstance(data, list):
+        return [_cast_dict_keys(value, key_cast) for value in data]
+
+    return data
+
+
+def _camel_to_snake(key):
```
NicolasT: Not entirely sure it's the exact same, but:

```
'_'.join(p.lower() for p in re.split(r'([A-Z][a-z0-9]*)', 'camelCaseValue') if p)
```

slaperche-scality: If we don't have to support things like `CamelCase` (`camel_case`) or `HTTPServer` (`http_server`), then a simple regex can do the trick.

```python
return re.sub('([a-z0-9])([A-Z])', r'\1_\2', key).lower()
```

Also, would be nice to have doctest that would serve as example of what is supported/what's the output:
- `camelCalse`?
- `CamelCase`?
- `camel_case`?
- `camelHTTPCase`?

salt/_utils/kubernetes_utils.py
slaperche-scality: Also, note that `isinstance` works with tuple. Thus you condition can be simplified as:

```python
if not isinstance(value, (float, int)):
```

salt/_utils/kubernetes_utils.py
slaperche-scality: Or ` six.integer_types` if we want to support py3.

salt/_utils/kubernetes_utils.py
slaperche-scality: If py2 only, I would go for `basestring` instead of `str` only.
If py3 compat, then yeah go for `six.{string_types,text_type,binary_type}`

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return __virtualname__
+
+
+def _extract_obj_and_details(manifest):
+    try:
+        details = __utils__['kubernetes.get_object_details'](manifest)
+        obj = __utils__['kubernetes.convert_manifest_into_object'](manifest)
+    except ValueError as exc:
+        raise CommandExecutionError('Invalid manifest: {!s}'.format(exc))
```
slaperche-scality: You're not really executing a command, are you?

Or is this exception the "used-for-everything" Salt exception?

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+            'version': details.version,
+            'plural': details.plural,
+        })
+
+    return kwargs
+
+
+def get_object(manifest, kubeconfig, context):
+    obj, details = _extract_obj_and_details(manifest)
+
```
slaperche-scality: why empty line here?

Empty lines are usually used to enhance readability, by separating logical block.
But here I don't really see the gain.

salt/_modules/better_kubernetes.py
slaperche-scality: https://metal-k8s.readthedocs.io/en/latest/developer/best_practices/python.html#don-t-use-hasattr-in-python-2

salt/_renderers/better_kubernetes.py
```diff
@@ -0,0 +1,104 @@
...
+__virtualname__ = 'kubernetes'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def _step_name(obj, absent=False):
+    try:
+        namespace = obj.metadata.namespace
```
slaperche-scality: `namespace = getattr(obj.metadata, 'namespace', None)`?
Or could `metadata` be missing as well?

salt/_renderers/better_kubernetes.py
```diff
@@ -0,0 +1,104 @@
...
+    except AttributeError:
+        namespace = None
+
+    try:
+        name = obj.metadata.name
+    except AttributeError:
+        raise SaltRenderError('Object `metadata.name` must be set.')
+
+    if namespace is not None:
+        full_name = '{}/{}'.format(
```
slaperche-scality: `        full_name = '{}/{}'.format(namespace,name)` fit on a single line, so let's save some screen real estate.

salt/_renderers/better_kubernetes.py
```diff
@@ -0,0 +1,104 @@
...
+    """Render a single Kubernetes object into a state 'step'."""
+    try:
+        assert __utils__['kubernetes.validate_manifest'](manifest)
+    except (AssertionError, ValueError) as exc:
+        raise SaltRenderError('Invalid manifest: {!s}'.format(exc))
+
+    obj = __utils__['kubernetes.convert_manifest_into_object'](manifest)
+    step_name = _step_name(obj, absent)
+
+    state_func = 'better_kubernetes.object_{}'.format(
```
slaperche-scality: `better_kubernetes` :grin:

salt/_renderers/better_kubernetes.py
```diff
@@ -0,0 +1,104 @@
...
+
+    kubeconfig = args.get('kubeconfig', [None])[0]
+    context = args.get('context', [None])[0]
+    absent = args.get('absent', [False])[0]
+
+    if not isinstance(source, six.string_types):
+        # Assume it is a file handle
+        source = source.read()
+
+    data = yaml.load_all(source, Loader=SaltYamlSafeLoader)
```
slaperche-scality: `safe_load_all`?

salt/_states/better_kubernetes.py
```diff
@@ -0,0 +1,87 @@
...
+            ret['comment'] = 'The object was created'
+
+        return ret
+
+    try:
+        new = __salt__['metalk8s_kubernetes.replace_object'](
+            name, kubeconfig, context, manifest
+        )
+    except CommandExecutionError as exc:
+        ret['comment'] = 'Something went wrong: {}'.format(exc)
```
slaperche-scality: s/Something went wrong/Cannot replace object/

salt/_states/better_kubernetes.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+        return ret
+
+    if obj is None:
+        try:
+            __salt__['metalk8s_kubernetes.create_object'](
+                name, kubeconfig, context, manifest
+            )
+        except CommandExecutionError as exc:
+            ret['comment'] = 'Something went wrong: {}'.format(exc)
```
slaperche-scality: s/Something went wrong/Cannot create object/

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
+"""Utility methods for manipulation of Kubernetes objects in Python."""
+import datetime
+import enum
```
slaperche-scality: `enum` was introduced in 3.4 and here we are in Python 2.

Maybe Salt provides a compat for it (like for `ipaddress`) but then we should import it as a compat, no?

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+    plural = property(operator.attrgetter('_plural'))
+    create = property(operator.attrgetter('_create'))
+    retrieve = property(operator.attrgetter('_retrieve'))
+    update = property(operator.attrgetter('_update'))
+    delete = property(operator.attrgetter('_delete'))
+    replace = property(operator.attrgetter('_replace'))
+
+    @property
+    def api_version(self):
+        return '/'.join([self.group, self.version])
```
slaperche-scality: Since you have a fixed number of field, and this number is small, `return '{s.group}/{s.version}'.format(s=self)` may be more readable.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+            return cls(value)
+        if isinstance(value, list):
+            return [cls.from_value(val) for val in value]
+        return value
+
+    def __repr__(self):
+        return repr(self._fields)
+
+    def __getattribute__(self, name):
+        try:
```
slaperche-scality, line 436: Why not:

```python
        try:
            return super(_DictWrapper, self).__getattribute__(name)
        except AttributeError:
            try:
                return self._fields[name]
            except KeyError:
                raise AttributeError(
                    "Custom object has no attribute '{}'".format(name)
                )
```
Then you avoid the `pass` and the need for a comment.

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+                'Unsupported attribute {} for "{}" object.'.format(
+                    src_key, model.__name__
+                )
+            )
+
+        try:
+            value = _cast_value(src_value, type_str)
+        except TypeError as exc:
+            raise ValueError(
+                'Invalid type for attribute {} of a "{}" object: {}.'.format(
```
slaperche-scality: Invalid value rather than type, no?

salt/_utils/kubernetes_utils.py
```diff
@@ -0,0 +1,535 @@
...
+    # manifest, as is done by the `kubernetes` client.
+    desired = _cast_dict_keys(manifest, key_cast=_camel_to_snake)
+
+    # We only check fields provided in the source, since the conversion
+    # may add empty or default values for optional fields
+    dictdiff = recursive_diff(desired, result)
+    return not dictdiff.removed() and not dictdiff.changed()
+
+
+def _cast_dict_keys(data, key_cast):
```
slaperche-scality, line 693: So `_cast_dict_keys` handles `list`? That's confusing.

salt/_modules/better_kubernetes.py
gdemonet: So, what I understand of this `CommandExecutionError` is that it's the standard one to raise from **execution modules**, so that invoking `salt 'target' my_module.my_command` will output a proper message. This is just a pattern I observed, so not entirely sure about it. Neither is it documented in [Writing Execution Modules](https://docs.saltstack.com/en/latest/ref/modules/), nor anything related to raising exceptions from them.

Any more info on this @TeddyAndrieux @sayf-eddine-scality @NicolasT ?

NicolasT: All I can say is: I've seen `CommandExecutionError` being used all over the place in Salt modules as well, so this seems legit.

salt/_modules/better_kubernetes.py
gdemonet: Yeah, that part doesn't look nice to me... However, that's one specificity from the Python K8s client when it comes to managing custom objects and the signature of the client methods.

For instance, a _standard_ object would be created with [`CoreV1Api.create_namespaced_pod(namespace, body)`](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#create_namespaced_pod).

On the other hand, custom objects need [`CustomObjectsApi.create_namespaced_custom_object(group, version, namespace, plural, body)`](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#create_namespaced_custom_object).

Now that I look again at those examples, it seems I made another unrelated mistake: `create` methods don't accept a `name` argument...

gdemonet: However, I _could_ inject these arguments when constructing the `CRDetails` object. :+1: this comment if that would suit you :)

salt/_modules/better_kubernetes.py
gdemonet: Indeed, did not remove the space when refactored the `_extract...` helpers. Will do!

salt/_modules/better_kubernetes.py
gdemonet: @NicolasT Yes, similar to the special case for `kwargs` with custom objects, I'd like to remove this check altogether. We know that all models for standard objects have a `to_dict` method, so I'll just have something similar for custom objects.

@slaperche-scality Oopsie, forgot that one! Anyway, as explained above, I'll remove the check altogether.

salt/_modules/better_kubernetes.py
gdemonet: Yup, I did not finish going through this refactoring... Will make it all consistent in the end.

salt/_modules/better_kubernetes.py
gdemonet: We could, but I'd rather have clear helpers that explain what's happening and let these methods define the specificities, instead of a bunch of `if method == 'create':` tests in the generic `_object_manipulation_function(method)`.

If you disagree, please feel free to detail why :)

salt/_renderers/better_kubernetes.py
gdemonet: It could: `CustomObject`s don't have validation of the manifest except for `apiVersion` and `kind`. I could add it to the constructor though, to make it clearer. Or, for simplicity, just check `obj.metadata.name` first since it's mandatory, and then use `getattr(obj.metadata, 'namespace', None)` without fear.

salt/_renderers/better_kubernetes.py
gdemonet: Yeah I wondered... Since `__utils__` are only callable from other Python modules, I'd say we can safely change the API for validation to something along those lines:

```python
try:
    obj = __utils__['kubernetes.convert_manifest_into_object'](manifest)
except ValueError as exc:
    raise SaltRenderError('Invalid manifest: {!s}'...)

conversion_valid, diff = __utils__['kubernetes.validate_conversion'](obj, manifest)
if not conversion_valid:
    raise SaltRenderError(
        'Conversion of manifest is invalid. Removed fields: {}. Changed fields: {}.'.format(
            ', '.join(diff.removed()), ', '.join(diff.changed())
        )
    )
```

salt/_renderers/better_kubernetes.py
gdemonet: Yup, that's to help me progressively test the module while not changing all the Salt states :)

salt/_renderers/better_kubernetes.py
gdemonet: Makes sense, I took the pattern from the previous renderer, which, I guess, took it from existing Salt renderers (e.g. [the `yaml` one](https://github.com/saltstack/salt/blob/develop/salt/renderers/yaml.py)).

Apparently, the `salt.utils.yamlloader.SaltYamlSafeLoader` allows for some Salt-friendly loading, e.g. casting mappings into `OrderedDict`s. See [source code](https://github.com/saltstack/salt/blob/develop/salt/utils/yamlloader.py#L35-L166).

I'd say it's fine to assume simple, standard YAML in our K8s manifests, so I'll remove this specificity and use the original `yaml.safe_load_all`.

Or maybe you think we should preserve order in there as well @NicolasT ?

salt/_states/better_kubernetes.py
gdemonet: Indeed it should be `better_kubernetes` until we rename them :)

salt/_states/better_kubernetes.py
gdemonet: You mean avoid replacing the object if it matches the desired manifest? I'd love to, and will give it a try once the rest works fine.

However, from discussions with @NicolasT, we anticipated complications:
- how do we handle default values for fields unset in the source manifest (and what would be the difference between a default value and an unset field, from the client perspective)?
- how do we know which fields are not supposed to be managed by the client (e.g. `status`, `metadata.creationTimestamp`, ...)?

Those two issues have simple solutions that contradict each other:
- if we only check fields set in the source manifest, we can't ensure other fields were not set manually by a user, and that's undesired behaviour
- if we check all fields except a known set of exceptions, we may replace greedily and lose idempotency for all default values that were unset in the source manifest

I'd settle for the second approach, as it's still better than not even trying to be idempotent. But that wouldn't be idempotent most of the time, since **a lot** of fields in standard objects are optional. Maybe, though that would require some extensive investigations, we could ignore fields set to `None`, but I'm not convinced that it would map to unset optional fields directly.

salt/_utils/kubernetes_utils.py
gdemonet: Oopsie, forgot that, so used to `enum`... Didn't find much when [searching their repo](https://github.com/saltstack/salt/search?q=enum&unscoped_q=enum) though.

I'll do some python2-compatible variant of `enum` then (don't want to go down the road of installing/importing the [`enum34` backport](https://pypi.org/project/enum34/)).

salt/_utils/kubernetes_utils.py
gdemonet: Right, we could namespace the `renderer` as well, no? In case they get the same idea :)

salt/_utils/kubernetes_utils.py
gdemonet: Sure, will do.

salt/_utils/kubernetes_utils.py
gdemonet: OK, will try to give more details.

salt/_utils/kubernetes_utils.py
gdemonet: Agreed, I'm thinking these `ObjectDetails` / `CRDetails` could be renamed to something like `Kind` / `CustomKind`, though I'm not super fond of it either... Any suggestions?

salt/_utils/kubernetes_utils.py
gdemonet: Yes, the mention of wrong type would be in `str(exc)` anyway.

salt/_utils/kubernetes_utils.py
gdemonet: Even handles neither `list` nor `dict` as a passthrough. It warrants a docstring, agreed that it can be confusing.

salt/_utils/kubernetes_utils.py
gdemonet: Will try the regex approach, and will include doctests!

salt/_modules/better_kubernetes.py
slaperche-scality: Ok, let's keep it then :slightly_smiling_face:

salt/_renderers/better_kubernetes.py
slaperche-scality: Ha, haven't notice the `SaltYamlSafeLoader`, so I guess it's safe to keep it as is then.

salt/_utils/kubernetes_utils.py
slaperche-scality: IIRC, the py2-way of doing enum is simply to use class attribute like that:

```
class ObjectScope(object):
    NAMESPACE = 'namespaced'
    CLUSTER = 'cluster'
```

You don't get something as fancy as py3-enum, but for the use we have here it's probably good enough.

salt/_modules/better_kubernetes.py
```diff
@@ -0,0 +1,148 @@
...
+except ImportError:
+    HAS_LIBS = False
+else:
+    HAS_LIBS = True
+
+__virtualname__ = 'metalk8s_kubernetes'
+
+
+def __virtual__():
+    if not HAS_LIBS:
```
gdemonet: Will also check that the `__utils__` module is properly loaded

salt/_utils/kubernetes_utils.py
gdemonet: I did something slightly more fancy, for comfort:

```python
# Roughly equivalent to an Enum, for Python 2
class ObjectScope:
    NAMESPACE = 'namespaced'
    CLUSTER = 'cluster'

    def __init__(self, value):
        allowed_values = (ObjectScope.NAMESPACE, ObjectScope.CLUSTER)
        if value not in allowed_values:
            raise ValueError('Value must be one of {}.'.format(allowed_values))
        self.value = value

    def __eq__(self, other):
        if isinstance(other, ObjectScope):
            return self.value == other.value
        if isinstance(other, str):
            return self.value == other
        return NotImplemented

    def __repr__(self):
        return 'ObjectScope({})'.format(self.value)
```

salt/_utils/kubernetes_utils.py
gdemonet: OK, so after going through all attributes in the models from `kubernetes.client.models`, I found that the rules are quite strange.

Here's what I ended up with:

```python
def _convert_attribute_name(key):
    if keyword.iskeyword(key):
        return '_{}'.format(key)

    if key.startswith('$'):
        # Only two supported values, '$ref' and '$schema'
        return key[1:]

    for pattern in ['([a-z])([A-Z0-9])', '([A-Z0-9])([A-Z0-9][a-z])']:
        key = re.sub(pattern, r'\1_\2', key)

    return key.lower()
```

To check that it matches all examples across the original models, here's what I did:
```python
attr_map = {}
for model in kubernetes.client.models.__dict__:
    attr_map.update(getattr(model, 'attribute_map', {}))

assert all(
    python_key == _convert_attribute_name(k8s_key)
    for python_key, k8s_key in attr_map.items()
)
```

salt/_utils/kubernetes_utils.py
slaperche-scality: I guess `allowed_values` don't need to be recreated at each `__init__` call and could use `basestring` (or whatever `six` provides` in the `isinstance` instead of `str`).

Otherwise LGTM :slightly_smiling_face:

salt/_modules/better_kubernetes.py
gdemonet: Changed my mind when the methods got more complex :sweat_smile:

salt/_modules/better_kubernetes.py
gdemonet: I went for the injection of special params when manipulating custom objects, for simplicity on the caller side.

NicolasT:
Look out for https://github.com/kubernetes-client/python/issues/974 :-/

gdemonet:
@TeddyAndrieux now that you're taking over this task (thanks a lot, by the way), you may want to create a new PR

TeddyAndrieux:
> @TeddyAndrieux now that you're taking over this task (thanks a lot, by the way), you may want to create a new PR

I will do it when I have something that seems to work (so today or tomorrow I think :smiley: )

NicolasT:
Let's try to avoid doing PRs of non-BertE-compatible branches in general in the future. Doing 'another' PR to continue this work loses all original PR context, unless we explicitly link back and whatnot.

salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -1,327 +1,108 @@
...
-import salt.utils.yaml
-from salt.utils.odict import OrderedDict
-
+snippets separated by `---` lines), this will render a sequence of states
+(represented as an OrderedDict), mapping every such object to an invocation
+of our custom `object_[present|absent]` state function.
+
+To use it, add a shebang like `#!kubernetes` as the first line of your
+manifests SLS file. Optionally, you can use rendering pipelines (if templating
+is required), e.g. `#!jinja | metalk8s_kubernetes`.
```
NicolasT, line 10: Is it `kubernetes` (2 lines above) or `metalk8s_kubernetes`?

salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -1,327 +1,108 @@
...
-    '''
-    Handle a single Kubernetes object, rendering it into a state 'step'
-    '''
-    name = _step_name(obj)
-    api_version = obj['apiVersion']
-    kind = obj['kind']
+def _step(manifest, kubeconfig=None, context=None, absent=False):
+    """Render a single Kubernetes object into a state 'step'."""
+    try:
+        obj = __utils__['metalk8s_kubernetes.convert_manifest_into_object'](
```
NicolasT, line 61: This should be called `convert_manifest_to_object` (or even `manifest_to_object`).

TeddyAndrieux:
> Let's try to avoid doing PRs of non-BertE-compatible branches in general in the future. Doing 'another' PR to continue this work loses all original PR context, unless we explicitly link back and whatnot.

Indeed, I will create a new PR soon and link this one (just in case)

salt/_renderers/metalk8s_kubernetes.py
TeddyAndrieux: `metalk8s_kubernetes` indeed I will update this

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1,1938 +1,164 @@
...
+    """Ensure that the object is absent.
 
-    namespace
-        The name of the namespace
-    '''
+    Arguments:
+        name (str): Path to a manifest yaml file or just a name
+        manifest (dict): Manifest content
+    """
+    ret = {'name': name, 'changes': {}, 'result': True, 'comment': ''}
```
NicolasT, line 33: I think, in general, it's safer to set `result` to `False` when constructing the 'base' result, and explicitly set it to `True` *when something succeeded*.

This is in line with other Salt code.

salt/_renderers/metalk8s_kubernetes.py
TeddyAndrieux: Hmm btw the old one is `kubernetes` and now renamed to `metalk8s_kubernetes`

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1,1938 +1,164 @@
...
-    ret = {'name': name,
-           'changes': {},
-           'result': False,
-           'comment': ''}
+    Arguments:
+        name (str): Path to a manifest yaml file
+                    or just a name if manifest provided
+        manifest (dict): Manifest content
+    """
+    ret = {'name': name, 'changes': {}, 'result': True, 'comment': ''}
```
NicolasT, line 73: See above.

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1,1938 +1,164 @@
...
-        __salt__['metalk8s_kubernetes.namespace_add_annotation'](
-            name=name,
-            annotation_key=annotation_key,
-            annotation_value=annotation_value,
+    if obj is None:
+        __salt__['metalk8s_kubernetes.create_object'](
+            name=name_arg, manifest=manifest,
             **kwargs
         )
+        ret['changes'] = {'old': 'absent', 'new': 'present'}
```
NicolasT, line 94: So we bubble up any exceptions `create_object` could raise?

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1,1938 +1,164 @@
...
-    }
-    ret['comment'] = '{} will change'.format(
-        'Some taints' if (added | updated) else 'No taint'
-    )
+    Arguments:
+        name (str): Path to a manifest yaml file
+                    or just a name if manifest provided
+        manifest (dict): Manifest content
+    """
+    ret = {'name': name, 'changes': {}, 'result': True, 'comment': ''}
```
NicolasT, line 123: See above.

salt/_states/metalk8s_kubernetes.py
```diff
@@ -1,1938 +1,164 @@
...
-            ret['result'] = None
-            ret['comment'] = 'The custom resource is going to be created'
-            return ret
-
-        res = __salt__['metalk8s_kubernetes.create_namespaced_custom_object'](
-            cr_data['group'], cr_data['version'], namespace,
-            cr_data['plural'], cr_data['body'], **kwargs
-        )
+    if cmp_manifest:
+        new_obj = __utils__['metalk8s_kubernetes.caml_to_snake'](cmp_manifest)
```
NicolasT, line 145: `camel`?

salt/_states/metalk8s_kubernetes.py
TeddyAndrieux: Salt will automatically raise this exception directly so for me it's good enough, except if we really want to hide exception but I think it's not a good idea 
state return will look like this (don't take care of the error message, was only a typo :stuck_out_tongue: )
```
metalk8s_kubernetes_|-toto_|-toto_|-object_updated:
    ----------
    __id__:
        toto
    __run_num__:
        0
    __sls__:
        None
    changes:
        ----------
    comment:
        An exception occurred in this state: Traceback (most recent call last):
          File "/usr/lib/python2.7/site-packages/salt/state.py", line 1919, in call
            **cdata['kwargs'])
          File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1918, in wrapper
            return f(*args, **kwargs)
          File "/var/cache/salt/master/extmods/states/metalk8s_kubernetes.py", line 136, in object_updated
            diff = __utils__['dictdiffer.recursive_diff'](obj, new_obj.to_dict())
        AttributeError: 'dict' object has no attribute 'to_dict'
    duration:
        35.622
    name:
        toto
    result:
        False
    start_time:
        14:08:48.872329
```

salt/_states/metalk8s_kubernetes.py
TeddyAndrieux: We can but as we never return False I don't know if it's really usefull (the only way to get a false is because of an exception raised and salt automatically set the result to false)

NicolasT:
I believe this was included in #1972 so can be closed now @gdemonet?

gdemonet:
> I believe this was included in #1972 so can be closed now @gdemonet?

Indeed.

PR #1714, opened 2019-09-18
Create deb repository root

### LINKED ISSUE #1701: Create the debian repository root on the iso
**Component**: build

**Why this is needed**: we need to create directories which will be mounted to create the internal repository.

**What should be done**: modify the buildchain to create the directories where packages will be downloaded the mounted in a container to create a repository.

**Test plan**:
build the image then verify that we have packages in the good path on the iso.

ycointe:
**Component**: Buildchain

**Context**: MetalK8son Ubuntu

**Summary**: We need to create Debian repository root directly in the buildchain and not in the script which download packages.

**Acceptance criteria**: Execute task download_deb_packages with the command: `./doit.sh _download_deb_packages` the verify in the folder ` _build/root/packages/` that you have all repository folder with .deb corresponding inside

---

Closes: #1701

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -175,7 +175,22 @@ def task__download_rpm_packages() -> types.TaskDict:
 
 def task__download_deb_packages() -> types.TaskDict:
     """Download Debian packages locally."""
-    # TODO: Clean the repository
+    def clean() -> None:
+        """Delete cache and repositories on the ISO."""
+        for repository in DEB_REPOSITORIES:
+            # Repository with an explicit list of packages are created by a
+            # dedicated task that will also handle their cleaning, so we skip
+            # them here.
+            if repository.packages:
+                continue
+            coreutils.rm_rf(repository.rootdir)
+        constants.REPO_DEB_ROOT.rmdir()
+
+    def mkdirs() -> None:
+        """Create directories for the repositories."""
+        for repository in DEB_REPOSITORIES:
+            repository.rootdir.mkdir(exist_ok=True)
+
     mounts = [
         utils.bind_ro_mount(
             source=constants.ROOT/'packages'/'debian'/'download_packages.py',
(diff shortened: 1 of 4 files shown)


packages/debian/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt update -y \
         build-essential \
         lintian \
         apt-rdepends \
-        python-apt \
+        reprepro \
```
slaperche-scality: Not related to this PR

buildchain/buildchain/packaging.py
```diff
@@ -404,6 +404,48 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
         "{p.name}".format(p=package)
         for package in versions.DEB_PACKAGES
         if package.name not in DEB_TO_BUILD
+    )
+
+SCALITY_DEB_REPOSITORY = targets.DEBRepository(
```
slaperche-scality: Why not declare it directly in `DEB_REPOSITORIES` like the others?

buildchain/buildchain/packaging.py
```diff
@@ -153,6 +153,10 @@ def clean() -> None:
 def task__download_deb_packages() -> types.TaskDict:
     """Download Debian packages locally."""
     # TODO: Clean the repository
```
slaperche-scality: Now that we `mkdirs` we could have a commit implementing the cleanup.

slaperche-scality (changes requested):

packages/debian/Dockerfile
ycointe: Indeed, I cleaned up.

buildchain/buildchain/packaging.py
```diff
@@ -175,7 +175,21 @@ def clean() -> None:
 
 def task__download_deb_packages() -> types.TaskDict:
     """Download Debian packages locally."""
-    # TODO: Clean the repository
+    def clean() -> None:
+        """Delete cache and repositories on the ISO."""
+        coreutils.rm_rf(constants.REPO_DEB_ROOT)
```
slaperche-scality: If you `rm -rf` that directory, you also remove all the repo at once (so the loop is useless).
I think we want something like that instead:

```python
        for repository in DEB_REPOSITORIES:
            # Repository with an explicit list of packages are created by a
            # dedicated task that will also handle their cleaning, so we skip
            # them here.
            if repository.packages:
                continue
            coreutils.rm_rf(repository.rootdir)
        constants.REPO_DEB_ROOT.rmdir()
```

slaperche-scality (changes requested):

slaperche-scality (dismissed):

slaperche-scality (dismissed):

slaperche-scality (approved):

PR #1715, opened 2019-09-18
Remove unused images

### LINKED ISSUE #1678: Prune shipped images
Given #1672 some images (at least `addon-resizer-amd64`) are no longer used. However, we still include it in the built ISO.

Should perform a bootstrap install, check which images are pulles (`crictl images`), then remove any unused (and unwanted) ones from the buildchain (but not, e.g. `metalk8s-utils`, of course).

slaperche-scality:
**Component**:

buildchain

**Context**: 

Remove unused images (faster build, smaller ISO).

**Summary**:

Simply removes the images from the buildchain.

**Acceptance criteria**: 

Bootstrap is still working.

---

Closes: #1678

### DIFF
--- a/buildchain/buildchain/image.py
+++ b/buildchain/buildchain/image.py
@@ -166,7 +166,6 @@ IMGS_PER_REPOSITORY : Dict[str, List[str]] = {
     ],
     constants.COREOS_REPOSITORY: [
         'configmap-reload',
-        'kube-rbac-proxy',
         'kube-state-metrics',
         'prometheus-config-reloader',
         'prometheus-operator',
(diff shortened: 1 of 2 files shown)


slaperche-scality:
Open as draft because not tested yet (Vagrant and CI are "broken" right now).

NicolasT (approved):
In theory this could go back to 2.0, but meh :man_shrugging: 

We *may* however want to turn this into a test-case, which after bootstrap asks `containerd` (through `crictl`) which images it pulled, gets a list of available images from the version ISO, and validates every image on the ISO has been pulled by `containerd`, except some whitelisted ones (think `metalk8s-utils`), so we don't accumulate cruft over time.

slaperche-scality:
> In theory this could go back to 2.0, but meh man_shrugging

Ho right, #1672 was merged from 2.0 >_<
 
> We _may_ however want to turn this into a test-case, which after bootstrap asks `containerd` (through `crictl`) which images it pulled, gets a list of available images from the version ISO, and validates every image on the ISO has been pulled by `containerd`, except some whitelisted ones (think `metalk8s-utils`), so we don't accumulate cruft over time.

Tracked by #1719 (I've added some questions there).

PR #1723, opened 2019-09-19
Fix release in debpackage definition

### LINKED ISSUE #1721: Fix release in DEBPackage definition
**Component**: build

**What happened**: The Calicol release number is wrong once the conversion from rpm to deb is done.
This is due to a wrong release definition in the buildchain

**What was expected**:
We expect that the task is finding the target.

**Steps to reproduce**
`./doit.sh info _build_deb_packages:calico-cni-plugin:convert_rpm_pkg_to_deb` 
in `Status` you can read that the target doesn't exist

Fix the release number.
Use this release number for the naming.

ycointe:
**Component**: Build

**Summary**: We need to pin the release version of Calico and use it for the package naming.

**Acceptance criteria**: 
- launch `./doit iso` 
- launch `./doit.sh info _build_deb_packages:calico-cni-plugin:convert_rpm_pkg_to_deb`
- verify that the status is up-to-date 

---

Closes: #1721

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -352,14 +352,11 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
             'Missing version for package "{}"'.format(name)
         )
 
-    # In case the `release` is of form "{build_id}.{os}", which is standard
-    build_id_str, _, _ = pkg_info.release.partition('.')
-
     return targets.DEBPackage(
         basename='_build_deb_packages',
         name=name,
         version=pkg_info.version,
-        build_id=int(build_id_str),
+        build_id=int(pkg_info.release),
         sources=sources,
         builder=DEB_BUILDER,
         task_dep=['_package_mkdir_deb_root', '_build_deb_container'],
--- a/buildchain/buildchain/versions.py
+++ b/buildchain/buildchain/versions.py
@@ -350,13 +350,13 @@ PACKAGES: Dict[str, Tuple[PackageVersion, ...]] = {
         PackageVersion(
             name='calico-cni-plugin',
             version=CALICO_VERSION,
-            release='0'
+            release='1'
         ),
         PackageVersion(name='iproute2', override='iproute'),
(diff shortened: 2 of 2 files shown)


buildchain/buildchain/versions.py
```diff
@@ -350,7 +350,7 @@ def deb_full_name(self) -> str:
         PackageVersion(
             name='calico-cni-plugin',
             version=CALICO_VERSION,
-            release='0'
+            release='1'
```
slaperche-scality, line 353: You need to fix `sos-report` as well (it's build ID should be 1 not 0)

The only reason why it was working for `sos` is because WE build the package (and thus were passing 0 to the docker) so passing the buggy ID at two places made things work by accident (but it's still wrong).

(dont' forget to update the commit message accordingly when you amend the commit)

buildchain/buildchain/packaging.py
```diff
@@ -352,14 +352,11 @@ def _deb_package(name: str, sources: Path) -> targets.DEBPackage:
...
 
-    # In case the `release` is of form "{build_id}.{os}", which is standard
-    build_id_str, _, _ = pkg_info.release.partition('.')
-
     return targets.DEBPackage(
         basename='_build_deb_packages',
         name=name,
         version=pkg_info.version,
-        build_id=int(build_id_str),
+        build_id=int(pkg_info.release),
```
slaperche-scality, line 359: commit message:
- in the title, s/nuber/number
- in the body, s/gave/given/

Also it's not "to avoid parsing" but "because the release number IS the build id for DEB package that we build ourselve".

slaperche-scality (changes requested):

buildchain/buildchain/versions.py
ycointe: Done

buildchain/buildchain/packaging.py
ycointe: It's changed

slaperche-scality (approved):

Issue #1697, opened 2019-09-16
Convert single node step into a downgrade scenario


sayf-eddine-scality:
**Component**:
upgrade, downgrade, tests

**Why this is needed**:
Test downgrade scenario to N-1 version

**What should be done**:
Install the current version of MetalK8s, run the tests (fast and slow), upgrade to N-1, rerun tests from N-1 branch

**Implementation proposal** (strongly recommended):
- Grab the ISO version N
- Mount the ISO
- Run the bootstrap script
- Run the test suite
- Grab the ISO version N-1
- Mount the ISO
- Run the downgrade script

**Test plan**: None

PR #1730, opened 2019-09-19
test/downgrade: Convert the single node stage to a downgrade test

sayf-eddine-scality:
Fixes: #1697

**Component**:
upgrade, downgrade, tests

**Context**: 
Upgrade/downgrade are not tested by eve

**Summary**:
We run downgrade tests to N-1 version in the CI

**Acceptance criteria**: 
The upgrade stage should be successful 

---

Closes: #1697

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -308,7 +308,7 @@ stages:
           name: Run fast tests locally
           env:
             ISO_MOUNTPOINT: >
-              /srv/scality/metalk8s-%(prop:metalk8s_version)s
+              /srv/scality/metalk8s-%(prop:metalk8s_prev_version)s
           command: >
             git checkout "development/%(prop:product_version_prev)s" &&
             tox -e tests-local --
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -197,14 +197,15 @@ stages:
     steps:
       - Git: *git_pull
       - ShellCommand: *setup_cache
-      - SetPropertyFromCommand:
+      - SetPropertyFromCommand: *product_version_prev
+      - SetPropertyFromCommand: &set_prev_version
           name: Set previous version
           property: metalk8s_prev_version
           command: >
             bash -c '
-              source VERSION &&
-              VERSION_MINOR=$((VERSION_MINOR-1))
-            echo $VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH$VERSION_SUFFIX'
+              git checkout development/%(prop:product_version_prev)s --quiet &&
+              source VERSION ;
+              echo $VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH$VERSION_SUFFIX'
       - ShellCommand: &ssh_ip_setup
           name: Report IP and install SSH keys
           command: >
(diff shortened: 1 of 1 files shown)


eve/main.yml
```diff
@@ -216,7 +224,6 @@ stages:
           command: >
             sudo mount -o loop metalk8s.iso
             /srv/scality/metalk8s-%(prop:metalk8s_version)s
-          haltOnFailure: true
```
Ebaneck: I think we should always keep `haltOnFailure: true`.. no??

eve/main.yml
```diff
@@ -277,7 +284,34 @@ stages:
...
+            shasumurl="/pre/SHA256SUM-%(prop:metalk8s_prev_version)s";
+            curl -s -XGET -o /tmp/SHA256SUM
+            "%(prop:artifacts_private_url)s$shasumurl"
+          haltOnFailure: true
+      - ShellCommand:
+          name: Check N-1 image with checksum
+          command: cd /tmp && sha256sum -c SHA256SUM
+          haltOnFailure: true
           alwaysRun: true
+      - ShellCommand:
```
Ebaneck: Ditto, we should add `haltOnFailure: true`

eve/main.yml
```diff
@@ -277,7 +284,34 @@ stages:
...
+          command: >
+            shasumurl="/pre/SHA256SUM-%(prop:metalk8s_prev_version)s";
+            curl -s -XGET -o /tmp/SHA256SUM
+            "%(prop:artifacts_private_url)s$shasumurl"
+          haltOnFailure: true
+      - ShellCommand:
+          name: Check N-1 image with checksum
+          command: cd /tmp && sha256sum -c SHA256SUM
+          haltOnFailure: true
           alwaysRun: true
```
TeddyAndrieux: I think you don't want to alwaysRun this one

eve/main.yml
TeddyAndrieux: Agree

eve/main.yml
```diff
@@ -277,7 +284,34 @@ stages:
...
+      - ShellCommand:
+          name: Run downgrade
+          command: >
+            sudo bash
+            /srv/scality/metalk8s-%(prop:metalk8s_version)s/iso-manager.sh
+            --archive /tmp/metalk8s.iso &&
+            sudo bash
+            /srv/scality/metalk8s-%(prop:metalk8s_version)s/downgrade.sh
+            --destination-version %(prop:metalk8s_prev_version)s
```
TeddyAndrieux: `haltOnFailure: true` for this one make sense IMO

TeddyAndrieux (dismissed):

eve/main.yml
```diff
@@ -277,7 +284,34 @@ stages:
...
+          name: Retrieve ISO N-1 image checksum
+          command: >
+            shasumurl="/pre/SHA256SUM-%(prop:metalk8s_prev_version)s";
+            curl -s -XGET -o /tmp/SHA256SUM
+            "%(prop:artifacts_private_url)s$shasumurl"
+          haltOnFailure: true
+      - ShellCommand:
+          name: Check N-1 image with checksum
+          command: cd /tmp && sha256sum -c SHA256SUM
+          haltOnFailure: true
```
TeddyAndrieux: IMO you should retrieve and check both ISO at the very beginning (and also you should use the same yaml tag for upgrade and downgrade)

eve/main.yml
```diff
@@ -277,7 +284,34 @@ stages:
       - Upload:
           name: upload artifacts
           source: sosreport
```
TeddyAndrieux: Like in Upgrade please move `upload artifacts` and sosreport to the end of the step (`alwaysRun:true` so we will always have the report

sayf-eddine-scality: I wanted to have 2 sos report collections if someday we want to see the difference between the two states

TeddyAndrieux: Then please rename the file like `sosrepot-blabla-post-install.tar.xz` `sosreport-blabla-post-upgrade.tar.xz` `sosreport-blabla-post-downgrade`

eve/main.yml
```diff
@@ -277,7 +313,6 @@ stages:
...
-          alwaysRun: true
```
TeddyAndrieux: Keep always run I think

eve/main.yml
```diff
@@ -177,22 +177,30 @@ stages:
...
-  single-node:
+  single-node-downgrade-test:
     worker:
       type: openstack
       image: CentOS-7-x86_64-GenericCloud-1809.qcow2
       flavor: m1.large
       path: eve/workers/openstack-single-node
     steps:
       - Git: *git_pull
+      - SetPropertyFromCommand:
```
TeddyAndrieux: Please use yaml anchor (`&set_prev_vers` `*set_prev_vers`) for all common step for downgrade/upgrade to remove copy paste. At least :
- `Set previous version`
- `Retrieve ISO version N-1 image`
- `Retrieve ISO N-1 image checksum`
- `Check N-1 image with checksum`

Ebaneck: @TeddyAndrieux lets merge one first and then we can add and re-use `anchors` in the second PR. This is the reason why the second PR on `Upgrade` is in draft mode till now

TeddyAndrieux: Ok yes you can add/use them in a single PR, as you want

TeddyAndrieux: @sayf-eddine-scality please do this, as the upgrade merged now :)

Ebaneck (approved):

PR #1732, opened 2019-09-19
Build internal Debian repository

ycointe:
**Component**: Repository, container

**Summary**: We need to build a Debian repository to dispense Debian packages required for Ubuntu nodes deployment.

**Acceptance criteria**: 
- We need to create a script to execute the `reprepro` command to build a repository inside a container.

**Tests**: (Need to bue done after ISO buit with `./doit.sh`)
- go on the branch `improvement/1617-implement-Debian-repository`
- build Docker image from the Debian Dockerfile (`docker build -t myubuntu:18.04 packages/debian/` )
- Lauch the container with the good volumes mounted:
  `docker run --rm -it -v $(pwd)/packages/debian/entrypoint.sh:/entrypoint.sh -v $(pwd)/packages/debian/distributions:/distributions -v $(pwd)/_build/root/packages/debian/:/metalk8s-deb myubuntu:18.04 bash`
- copie the file `distributions` in the good path (it will be automatically mounted after):
`cp distributions metalk8s-deb/`
- Then lauch the script entrypoint.sh calling `buildrepo` : `./entrypoint.sh buildrepo`

- Once the execution finished, verify in the folder `/repository/debian/dists/bionic/metalk8s-*` that the file `Release` is well here ! (the packages are in the folder `/repository/debian/pool`)
---

Closes:

### DIFF
--- a/packages/debian/Dockerfile
+++ b/packages/debian/Dockerfile
@@ -11,7 +11,6 @@ RUN apt update -y \
         devscripts \
         build-essential \
         lintian \
-        apt-rdepends \
         python-apt \
         reprepro \
         alien
--- a/packages/debian/Dockerfile
+++ b/packages/debian/Dockerfile
@@ -13,6 +13,7 @@ RUN apt update -y \
         lintian \
         apt-rdepends \
         python-apt \
+        reprepro \
         alien
 
 RUN useradd -m build
--- a/packages/debian/distributions
+++ b/packages/debian/distributions
@@ -1,0 +1,7 @@
+Origin: Scality
+Label: metalk8s
+Suite: stable
+Codename: bionic
+Architectures: amd64
+Components: metalk8s-bionic metalk8s-bionic-backports metalk8s-bionic-updates metalk8s-kubernetes-xenial metalk8s-salt_ubuntu1804 metalk8s-scality
+Description: MetalK8s repository for bionic
--- a/packages/debian/entrypoint.sh
+++ b/packages/debian/entrypoint.sh
@@ -36,9 +36,26 @@ rpm2deb() {
 }
 
 
+buildrepo() {
+    set -x
+    mkdir -p /repository/debian/{conf,incoming}
+    # All downloaded packages must be mounted in /metalk8s-deb
+    cp /metalk8s-deb/distributions /repository/debian/conf
+    #TODO: Mount `distributions` file in  /repository/debian/conf
+    reprepro -b /repository/debian export
+    for dir in /metalk8s-deb/metalk8s-*; do
+        reprepro -C "${dir##*/}" -b /repository/debian \
+            includedeb bionic "$dir"/*.deb
+    done
+}
+
+
 case ${1:- } in
     builddeb)
         builddeb
+        ;;
+    buildrepo)
+        buildrepo
         ;;
     rpm2deb)
         rpm2deb


packages/debian/Dockerfile
```diff
@@ -11,8 +11,8 @@ RUN apt update -y \
         devscripts \
         build-essential \
         lintian \
-        apt-rdepends \
```
slaperche-scality, line 14: Ha good catch

Forgot to remove this one when I rewrite the download in Python.

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,19 @@ rpm2deb() {
...
 }
 
+buildrepo(){
+    set -x
+    mkdir -p /var/www/debian/conf
+    mkdir -p /var/www/debian/incoming
+    # Mount `distributions` file in  /var/www/debian/conf
+    pushd /var/www/debian
+    reprepro -b . export
+    # We download packages that we want
```
slaperche-scality: We download?

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,19 @@ rpm2deb() {
...
+    reprepro -vb . includedeb <codename> <Path to the .deb>
+    # A folder `pool` is created whith all .deb insinde
+
+
+## TODO : Handle the GPG key or find a way to do without.
```
slaperche-scality: Are we OK "with a way without", I'm not sure: I guess we want signed stuff for security purpose.
To be discussed with the team.

packages/debian/Dockerfile
ycointe: I commit this change in a separate commit ;)

packages/debian/entrypoint.sh
ycointe: Comment deleted

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,19 @@ rpm2deb() {
...
+    mkdir -p /var/www/debian/conf
+    mkdir -p /var/www/debian/incoming
+    # We assume that all downloaded packages are mounted in /metalk8s-deb/
+    cp /metalk8s-deb/distributions /var/www/debian/conf
+    # Mount `distributions` file in  /var/www/debian/conf
+    reprepro -b /var/www/debian export
+    for dir in /metalk8s-deb/metalk8s-*; do
+        reprepro -C ${dir##*/} -b /var/www/debian includedeb bionic "$dir"/*.deb
+    done
+
```
alexandre-allard: It's missing the closing curly brace.

alexandre-allard (dismissed):

packages/debian/entrypoint.sh
```diff
@@ -35,6 +35,19 @@ rpm2deb() {
     chown "${TARGET_UID}:${TARGET_GID}" *
 }
 
+buildrepo(){
```
alexandre-allard: details: missing space after the closing parenthesis

packages/debian/entrypoint.sh
```diff
@@ -35,11 +35,28 @@ rpm2deb() {
...
+    mkdir -p /var/www/debian/conf
+    mkdir -p /var/www/debian/incoming
+    # We assume that all downloaded packages are mounted in /metalk8s-deb/
+    cp /metalk8s-deb/distributions /var/www/debian/conf
+    # Mount `distributions` file in  /var/www/debian/conf
+    reprepro -b /var/www/debian export
+    for dir in /metalk8s-deb/metalk8s-*; do
+        reprepro -C ${dir##*/} -b /var/www/debian includedeb bionic "$dir"/*.deb
+    done
+
```
alexandre-allard: extra space

packages/debian/entrypoint.sh
```diff
@@ -36,10 +36,27 @@ rpm2deb() {
 }
 
 
+buildrepo() {
+    set -x
+    mkdir -p /var/www/debian/{conf,incoming}
+    # We assume that all downloaded packages are mounted in /metalk8s-deb/
+    cp /metalk8s-deb/distributions /var/www/debian/conf
+    # Mount `distributions` file in  /var/www/debian/conf
```
gdemonet: It's not a mount, you're copying it. IMO, the comment can be removed.

packages/debian/entrypoint.sh
```diff
@@ -36,10 +36,27 @@ rpm2deb() {
 }
 
 
+buildrepo() {
+    set -x
+    mkdir -p /var/www/debian/{conf,incoming}
```
gdemonet: Why use `/var/www/debian`? I'm confused, is it supposed to be served by some webserver?

gdemonet: Also, if you want to set a basedir for `reprepro`, instead of duplicating the `-b` argument for each invocation, you could just set the `REPREPRO_BASE_DIR` environment variable.

packages/debian/distributions
```diff
@@ -0,0 +1,7 @@
+Origin: Scality
+Label: metalk8s
+Suite: stable
+Codename: bionic
+Architectures: amd64
+Components: main metalk8s-bionic metalk8s-bionic-backports metalk8s-bionic-updates metalk8s-kubernetes-xenial metalk8s-salt_ubuntu1804 metalk8s-scality
+Description: Build repository for Debian
```
gdemonet: Not sure this description is accurate. It should describe the repository, so let's go for something like:

```
Description: MetalK8s repository for bionic
```

packages/debian/entrypoint.sh
```diff
@@ -36,10 +36,27 @@ rpm2deb() {
...
+    # We assume that all downloaded packages are mounted in /metalk8s-deb/
+    cp /metalk8s-deb/distributions /var/www/debian/conf
+    # Mount `distributions` file in  /var/www/debian/conf
+    reprepro -b /var/www/debian export
+    for dir in /metalk8s-deb/metalk8s-*; do
```
gdemonet, line 46: So, IIUC (this loop would benefit from a comment explaining what it does, for readers unaware of Debian repositories building practices), we're adding all packages for `metalk8s-*` components.

Don't we also need to do the same for the `main` component? At least that's what I would expect from the `distributions` file.

packages/debian/distributions
```diff
@@ -0,0 +1,7 @@
+Origin: Scality
+Label: metalk8s
+Suite: stable
+Codename: bionic
+Architectures: amd64
+Components: main metalk8s-bionic metalk8s-bionic-backports metalk8s-bionic-updates metalk8s-kubernetes-xenial metalk8s-salt_ubuntu1804 metalk8s-scality
```
gdemonet: Why not `metalk8s-main`?

gdemonet (dismissed):
Bunch of questions, would like to have some comments in the code to get more info on this process.

Also, how can we make sure this script works, if it's not yet used in the buildchain? Would help to catch errors, and help reviewers to test what it's doing (also, we can't rely on CI since it won't hit this code).

packages/debian/entrypoint.sh
ycointe: The file `distributions` is created before the repository build, it should be mounted in the container then, paste in the folder `/var/www/debian/conf`.
This comment is just a reminder for the next PR, integrate this script in the buildchain (it will be deleted after)

packages/debian/entrypoint.sh
ycointe: In fact there is no main component, I will remove it from the `distributions` file.

packages/debian/distributions
ycointe: Indeed, it's from my tests, I will change it.

packages/debian/distributions
ycointe: In fact I have to delete main... there is no need of this component

packages/debian/entrypoint.sh
ycointe: we donc need the `/var/www` for now, I will remove it

slaperche-scality (approved):

Issue #1735, opened 2019-09-20
Delete ISO on error


slaperche-scality:
**Component**:

build

**What happened**:

If the `mkisofs` fails (e.g. because the disk is full), it may leave behind a partial or corrupted ISO file.
This is a problem, because those ISOs may be used if people don't know that the build failed.

**What was expected**:

If `mkisofs` fails, then no ISO should be present.

**Steps to reproduce**

- Run the build once `./doit.sh -n4`
- Forbid processes to create big files (so that ISO creation will fail): `ulimit -f 262144`
- Run the build again `./doit.sh` to recreate the ISO, this time it should fails

Now, you should have a truncated/corrupted ISO at `_build/metalk8s.iso`.

**Resolution proposal**:

Wrap the `mkisofs` call in a `try/except` block and delete the ISO fail on failure.

PR #1736, opened 2019-09-20
Delete truncated/corrupted ISO on failure

slaperche-scality:
**Component**:

build

**Context**: 

We want to cleanup truncated/corrupted ISO to avoid them being used by mistake.

**Summary**:

Use a `try/except` block to handle failure of `mkisofs` and cleanup afterward.

**Acceptance criteria**: 

- Run the build once `./doit.sh -n4`
- Forbid processes to create big files (so that ISO creation will fail): `ulimit -f 262144`
- Run the build again `./doit.sh` to recreate the ISO, this time it should fails
- the file `_build/metalk8s.iso` doesn't exist

---

Closes: #1735

### DIFF
--- a/buildchain/buildchain/iso.py
+++ b/buildchain/buildchain/iso.py
@@ -29,6 +29,7 @@ Overview:
 
 import datetime as dt
 import socket
+import subprocess
 from pathlib import Path
 from typing import Sequence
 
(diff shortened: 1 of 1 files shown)


buildchain/buildchain/iso.py
```diff
@@ -188,21 +189,36 @@ def action(targets: Sequence[str]) -> None:
...
-        '-full-iso9660-filenames',
-        '-volid', '{} {}'.format(config.PROJECT_NAME, versions.VERSION),
-        '--iso-level', '3',
-        '-gid', '0',
-        '-uid', '0',
-        '-input-charset', 'utf-8',
-        '-output-charset', 'utf-8',
-        constants.ISO_ROOT
-    ]
+    def unlink_if_exist(filepath: Path) -> None:
```
jbertran, line 192: IIRC we have a couple of these now (I recall @ycointe inserted something similar). Could be a `utils` candidate.

buildchain/buildchain/iso.py
```diff
@@ -188,21 +189,36 @@ def action(targets: Sequence[str]) -> None:
...
+            '--iso-level', '3',
+            '-gid', '0',
+            '-uid', '0',
+            '-input-charset', 'utf-8',
+            '-output-charset', 'utf-8',
+            constants.ISO_ROOT
+        ]
+        try:
+            subprocess.run(cmd, check=True)
+        except:
```
jbertran, line 218: Can't we just have `except CalledProcessError`? We only want to catch errors that occurred during our `run`, since they're the only ones that _may_ generate partial artifacts we want to clean up.

jbertran (commented):
Shame we can't specify cleanup on error with `doit`.

buildchain/buildchain/iso.py
slaperche-scality: Yup, but the one you mention is out of the buildchain so not that easy to share.

buildchain/buildchain/iso.py
slaperche-scality: Since the cleanup handle the case of "missing file" I guess we can always cleanup, to be on the safe side.

jbertran (approved):

slaperche-scality:
> Shame we can't specify cleanup on error with `doit`.

Would be nice indeed, I may open an issue on their side about that.

Issue #1738, opened 2019-09-20
Uncordon does not remove the unschedulable taint in 2.0


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**What happened**:

#1408  not fixed anymore

#992 split node cordon in another module `metalk8s_cordon.py` with same virtualname `metalk8s_kubernetes` but does not removed the old `node_uncordon` from `metalk8s_kubernetes.py` so the new `node_uncordon` from `metalk8s_cordon.py` is never used.

**What was expected**:

Uncordon in 2.0 when cordon in 2.1 should remove the unschedulable taint

**Resolution proposal** (optional):

Remove function in `metalk8s_cordon.py` from `metalk8s_kubernetes.py`

PR #1741, opened 2019-09-20
Remove (un)cordon functions from classic module

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Summary**:

Older cordon and uncordon functions still present in
`metalk8s_kubernetes.py` when we want to use the ones from
`metalk8s_cordon.py`. Both module use `metalk8s_kubernetes` as
virtualname so the function from `metalk8s_cordon.py` were ignored.

---

Fixes: #1738


NicolasT:
Or, maybe better: have every module we have use `__name__` as its `__virtualname__`, and call *what we want* *exactly* instead of relying on this Salt module shadowing nonsense?

TeddyAndrieux:
> Or, maybe better: have every module we have use `__name__` as its `__virtualname__`, and call _what we want_ _exactly_ instead of relying on this Salt module shadowing nonsense?

It's allow us to split functions in different python file and keep one entry point. Because if we separate `kubernetes` and `cordon` why we not also separate for example `node` and `pod`.
Maybe make sense to separate everything but I think have a separate "virtualname" only for cordon is not really good as both `metalk8s_kubernetes` and `metalk8s_cordon` are only about Apiserver interactions.

NicolasT:
Fair enough :)

NicolasT (approved):

Issue #1724, opened 2019-09-19
Doc abour how to access grafana service is wrong


thomasdanan:
**Component**:

doc

**What happened**:

To get the port used by grafana, we say in the doc to run the following command:

`kubectl --kubeconfig=/etc/kubernetes/admin.conf get service grafana -n metalk8s-monitoring`

but actually service named grafana does not exist. It is now named prometheus-operator-grafana

**What was expected**:

to get working commands (i.e. right grafana service name) in the doc to get grafana port.

**Steps to reproduce**

follow the doc "Accessing cluster services"

**Resolution proposal** (optional):

put the right grafana service name in the doc

PR #1742, opened 2019-09-20
doc: change procedure in order to access grafana service

thomasdanan:
**Component**:

doc: accessing cluster services

**Context**: 

procedure to access grafana service has to change since it is now deployed by prometheus operator 

**Summary**:

added a procedure to get grafana pod name and just port-forward it on client

**Acceptance criteria**: 

following the doc should be enough to access grafana and consult dashboards

---
Closes: #1724

### DIFF
--- a/docs/quickstart/services.rst
+++ b/docs/quickstart/services.rst
@@ -8,17 +8,6 @@ Grafana
 
 You will first need the latest ``kubectl`` version installed on your host.
 
-From the bootstrap node, get the port used by Grafana:
-
-.. code-block:: shell
-
-   root@bootstrap $ kubectl --kubeconfig=/etc/kubernetes/admin.conf get service grafana -n metalk8s-monitoring
-
-   NAME      TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
-   grafana   ClusterIP   10.109.125.193   <none>        3000/TCP   1h
-
-Please note the port used by Grafana (here ``3000``)
-
 To authenticate with the cluster, retrieve the admin kubeconfig on your host:
 
 .. code-block:: shell
@@ -29,7 +18,7 @@ Forward the port used by Grafana:
 
 .. code-block:: shell
 
-   user@your-host $ kubectl --namespace metalk8s-monitoring port-forward svc/grafana 3000
+   user@your-host $ kubectl -n metalk8s-monitoring port-forward svc/prometheus-operator-grafana 3000:80
 
 Then open your web browser and navigate to ``http://localhost:3000``
 


NicolasT:
I believe nowadays you can also `port-forward` to a *Service*, which could simplify the documentation a lot: instead of instructing the user to look up the *Pod* name, then alter the second command invocation, we know the *Service* name and can provide the exact `kubectl port-forward svc/...` to run.

gdemonet:
There's no `Service` deployed with Grafana anymore?

EDIT: As @NicolasT suggested, especially since that's what was done before.

thomasdanan:
Indeed @NicolasT @gdemonet so something like that would do the job: `kubectl --namespace metalk8s-monitoring port-forward svc/prometheus-operator-grafana 3000:80 --kubeconfig=/tmp/admin.conf`

docs/quickstart/services.rst
```diff
@@ -8,16 +8,14 @@ Grafana
 
 You will first need the latest ``kubectl`` version installed on your host.
 
-From the bootstrap node, get the port used by Grafana:
```
gdemonet, line 11: Indeed, we can remove this section regarding Grafana's port, as it'll always be 3000.

docs/quickstart/services.rst
```diff
@@ -8,16 +8,14 @@ Grafana
...
+From the bootstrap node, get the Grafana pod name:
```
gdemonet: No need, see below.

docs/quickstart/services.rst
```diff
@@ -29,7 +27,7 @@ Forward the port used by Grafana:
 
 .. code-block:: shell
 
-   user@your-host $ kubectl --namespace metalk8s-monitoring port-forward svc/grafana 3000
+   user@your-host $ kubectl port-forward -n metalk8s-monitoring prometheus-operator-grafana-54b8dc5844-sf5x4 3000
```
gdemonet: Using a Service is sufficient for port-forwarding. Also, we need to specify the kubeconfig to use (which was retrieved locally just before):

```suggestion
   user@your-host $ kubectl --kubeconfig=./admin.conf port-forward -n metalk8s-monitoring svc/prometheus-operator-grafana 3000:<local_port>
```

If you want, you could leave the default `<local_port>` to 3000.

docs/quickstart/services.rst
```diff
@@ -29,7 +27,7 @@ Forward the port used by Grafana:
...
 
 Then open your web browser and navigate to ``http://localhost:3000``
```
gdemonet, line 23: If you mention a `<local_port>` as suggested above, this would become:

```suggestion
Then open your web browser and navigate to ``http://localhost:<local_port>``
```

gdemonet (changes requested):

docs/quickstart/services.rst
thomasdanan: @gdemonet the thing is grafana sevice port is now 80 (not anymore 3000) and you cannot use 80 for local port ... so what works on my env is the following: `kubectl --namespace metalk8s-monitoring port-forward svc/prometheus-operator-grafana 3000:80 --kubeconfig=/tmp/admin.conf`. works for you ?

gdemonet (approved):
LGTM

PR #1743, opened 2019-09-20
Bugfix/1705/improve the volume detail page when failed to create volume

### LINKED ISSUE #1705: Volume: Device path is not fullfil properly
**Component**:
Web Gui volume information on a volume that fails to create

**What happened**:
The Path field is empty: may be because we have a device path, because of the RawBlockDevice type.

**What was expected**:
Having the device path in the path field. This could help to investigate. If I ran kubectl describe <MY-VOLUME> I get the device path. So the information exists
```
Spec:
  Node Name:  metal-k8s-servers01.novalocal
  Raw Block Device:
    Device Path:       /dev/vdb1
  Storage Class Name:  zenko
```

**Steps to reproduce**
Create a volume that has RawBlockDeviceType I guess the creation must fail.

Get the Device Path when the Type is RawBlockDevice

ChengYanJin:
**Component**: ui, storage

**Context**: 
As a user, when I failed to create a volume of `rawBlockDevice` I would like to get the path instead of an empty field

**Summary**:
Add a banner to show the error message when failed to create volume.
rawBlockDevice: get the path from the volume.
sparseLoopDevice: the path from PV is not useful so we just show no applicable. 

**Acceptance criteria**: 
- on a running MetalK8s platform, try to create a Volume with of type rawBlockDevice with a path that doesn't exists
- wait for the Volume to go in Failed state
- click on the volume, you should see a banner with the error message and the path you've entered in the form is displayed in the details
Screenshot 2019-09-20 at 16 32 54

Closes: #1705

### DIFF
--- a/ui/src/containers/VolumeInformation.js
+++ b/ui/src/containers/VolumeInformation.js
@@ -71,9 +71,7 @@ const VolumeInformation = props => {
   );
   const pV = pVList.find(pv => pv.metadata.name === currentVolumeName);
   const storageClass = storageClasses.find(
-    SC =>
-      SC.metadata.name ===
-      (volume && volume.spec && volume.spec.storageClassName),
+    SC => SC.metadata.name === volume?.spec?.storageClassName,
   );
 
   return (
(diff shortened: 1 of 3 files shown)


ui/src/containers/VolumeInformation.js
```diff
@@ -22,6 +24,7 @@ import {
   BreadcrumbLabel,
   StyledLink,
 } from '../components/BreadcrumbStyle';
+import Banner from '../components/Banner';
```
carlito-scality, line 27: It should be moved to core-ui

carlito-scality (commented):
The code looks good to me

ui/src/containers/VolumeInformation.js
```diff
@@ -82,6 +92,23 @@ const VolumeInformation = props => {
...
+      <VolumeInformationTitle>
+        {intl.messages.detailed_information}
+      </VolumeInformationTitle>
+
+      {volume?.status?.phase === STATUS_FAILED ? (
+        <Banner
+          type={STATUS_BANNER_ERROR}
+          icon={<i className="fas fa-exclamation-triangle" />}
+          title={
+            volume?.status?.errorCode === 'CreationError'
```
slaperche-scality, line 102: I could give you an exhaustive list of error codes so that you can map each to a string (instead of only `CreationError`) but let's keep it like that => this will be reworked when we move to `conditions` (and I plan to work on it next week).

ui/src/containers/VolumeInformation.js
```diff
@@ -127,13 +155,14 @@ const VolumeInformation = props => {
         <InformationSpan>
           <InformationLabel>{intl.messages.path}</InformationLabel>
           <InformationValue>
-            {pV && pV.spec && pV.spec.local && pV.spec.local.path}
+            {volume?.spec?.rawBlockDevice?.devicePath ||
```
slaperche-scality: minor: IMHO, this change should be in a dedicated commit: this THE change that fix the bug.

The banner is a great UX improvement, but not (directly) related to the bug reported so it would make sense to have one commit for the banner and one commit for the bugfix.

ui/src/containers/VolumeInformation.js
```diff
@@ -127,13 +155,14 @@ const VolumeInformation = props => {
...
+              intl.messages.not_applicable}
```
slaperche-scality, line 151: The logic here works but is not very future proof.
But since I don't see how it could fail for now, let's keep it (will probably need to be reworked when we have more than two types though).

slaperche-scality (commented):
Feature works as expected

Only have a few comments, mainly for information (nothing really big justifying a change, except if you do want to split the first commit)

So, LGTM but I'll let the other front guys to review the code before giving my approval :wink:

ui/src/containers/VolumeInformation.js
slaperche-scality: Can be done in another PR I think.

ChengYanJin: Yeah. I will do it later.

slaperche-scality (approved):

ChengYanJin:
@bert-e approve

PR #1745, opened 2019-09-20
Migrate Select component to MetalK8S

ChengYanJin:
**Component**: ui

**Context**: 
In core-ui, we replace `react-virtualized-select` by `react-select` because `react-virtualized-select` works only with the version 1.x of `react-select`
We would like to update the Select component in MetalK8S.

**Summary**:
Update the dependency:
We replaced `react-virtualized-select` by `react-select` in the dependencies.
So we suggest that you do `npm install`

Remove the import css:
import "react-select/dist/react-select.css";
import "react-virtualized-select/styles.css";

**Acceptance criteria**: 
Screenshot 2019-09-20 at 17 08 58

### DIFF
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -63,6 +63,10 @@ const CreateVolumeLayout = styled.div`
       }
     }
   }
+  .sc-select-option-label,
+  .sc-select__placeholder {
+    font-size: 14px;
+  }
 `;
 
 const SizeFieldContainer = styled.div`
@@ -70,14 +74,15 @@ const SizeFieldContainer = styled.div`
   align-items: flex-start;
   .sc-input-wrapper,
   .sc-input-type {
-    width: 120px;
+    width: 100px;
     box-sizing: border-box;
+    height: 38px;
   }
 `;
 
 const SizeUnitFieldSelectContainer = styled.div`
   .sc-select {
-    width: 75px;
+    width: 100px;
     padding-left: 5px;
   }
 `;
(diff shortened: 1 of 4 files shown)


ui/src/containers/CreateVolume.js
```diff
@@ -279,11 +284,10 @@ const CreateVolume = props => {
                     clearable={false}
                     type="select"
                     options={optionsStorageClasses}
-                    placeholder={intl.messages.select_a_storageClass}
-                    noResultsText={intl.messages.no_results}
+                    noOptionsMessage={() => intl.messages.no_results}
                     name="storageClass"
                     onChange={handleSelectChange('storageClass')}
-                    value={values.storageClass}
+                    value={optionsStorageClasses[0]}
```
carlito-scality, line 290: It should be based on values.storageClass even if now we have only one value

ui/src/containers/CreateVolume.js
```diff
@@ -279,11 +284,10 @@ const CreateVolume = props => {
                     clearable={false}
                     type="select"
                     options={optionsStorageClasses}
-                    placeholder={intl.messages.select_a_storageClass}
```
carlito-scality, line 282: Leave placeholder here

ui/src/containers/CreateVolume.js
```diff
@@ -293,11 +297,12 @@ const CreateVolume = props => {
                     clearable={false}
                     type="select"
                     options={optionsTypes}
-                    placeholder={intl.messages.select_a_type}
```
carlito-scality, line 296: Leave placeholder here

ui/src/containers/CreateVolume.js
```diff
@@ -63,21 +63,26 @@ const CreateVolumeLayout = styled.div`
       }
     }
   }
+  .sc-select-option-label,
+  .sc-select__placeholder {
+    font-size: 14px;
```
carlito-scality, line 68: use core-ui theme font-size

carlito-scality (changes requested):

slaperche-scality:
minor nitpick on the PR description:

> Context:  We would like to migrate the Select component to MetalK8S.

Would be nice to add "why" we want to migrate?
Also isn't it migrating Select component to `migrate the Select component to core-ui` or `migrate the Select component of MetalK8s`?

ChengYanJin:
wrong branch name. Going to recreate a PR

PR #1757, opened 2019-09-23
Add PDF documentation in the ISO

### LINKED ISSUE #1695: Include documentation PDFs on ISO
Somewhat related to #1595: having the docs HTML included in the UI would be great at some point. Next to this (and somewhat as an intermediate step as well, maybe?), having a rendered PDF of the docs included on the ISO would be pretty useful as well.

slaperche-scality:
**Component**:

build

**Context**: 

Add the PDF documentation in the ISO we ship

**Summary**:

- add tasks to build the documentation (`docs:html` and `docs:pdf`)
- add task to copy the generated PDF on the ISO

**Acceptance criteria**: 

When running `doit.sh -n4`, you will have `user-guide.pdf` under `_build/root/documentation`.

---

Closes: #1695


buildchain/buildchain/docs.py
```diff
@@ -0,0 +1,98 @@
...
+        doc_format = target.name.upper()
+        yield {
+            'name': target.name,
+            'title': utils.title_with_target1('DOC {}'.format(doc_format)),
+            'doc': 'Generate {} {} documentation'.format(
+                config.PROJECT_NAME, doc_format
+            ),
+            'actions': [['tox', '-e', 'docs', '--', target.command]],
+            'targets': [target.target],
+            'file_dep': list(utils.git_ls('docs')),
```
gdemonet, line 107: Neat trick, I'll use something like this for the Makefile in `examples/example-solution`

buildchain/buildchain/docs.py
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+
+def task_doc() -> Iterator[types.TaskDict]:
+    """Generate the documentation."""
+    def clean(target: DocTarget) -> Callable[[], None]:
+        """Delete the build sub-directory for the given target."""
+        return lambda: coreutils.rm_rf(target.target.parent)
+
+    doc_targets = (
```
gdemonet, line 88: Fine for now, but could be useful to be able to choose, no? If developing docs, I guess `tox -e docs -- livehtml` is more comfortable, but `./doit.sh doc_html` could be cool as well IMHO

slaperche-scality: I thought about supporting `livehtml`, but with `doit.sh auto doc:html` you'll have auto-recompilation as well no?

NicolasT: That's not the only thing `livehtml` does: it recompiles, but also embeds some JS in the result such that when recompiling, the server can send an event to the browser to trigger a reload.

slaperche-scality: Yeah, and it spawn a small webserver to serve the doc as well I think.

But meh, close enough :stuck_out_tongue:  (there is still `tox` if you really want to), or I could add a target `doc:live` (or `doc:auto`, or `doc:watch`), won't be hard.

I'll see once I have something working on the CI

slaperche-scality: Done

eve/main.yml
```diff
@@ -123,6 +135,8 @@ stages:
               bash -c '
               source metalk8s-%(prop:product_version_prev)s/VERSION &&
               echo %(prop:product_version_prev)s.$VERSION_PATCH$VERSION_SUFFIX'
+
+      - ShellCommand: *download_pdf_doc
```
NicolasT, line 139: So now we'd end up with a 'previous version' build/ISO containing the docs of the 'current version'?

slaperche-scality: Yeah, I guess for upgrade/downgrade we don't care about the embedded doc anyway.
It's just a matter of having a build that works.

Now if we have a simple way to retrieve the matching doc, why not.
If it's more involved then I'm not sure it's worth the effort.

NicolasT: Can't we simply not embed any doc in the resulting ISO, then?

slaperche-scality: But we always want the doc in the ISO, no?
Why should me make it optional? Only to workaround a quirk in our CI workflow?

This ISO is only built for testing upgrade/downgrade, it won't be uploaded anywhere to be reused.
I would say that embdedding a wrong PDF in a transient ISO is a lesser evil than making doc embedding optional in the buildchain (which could bite us later it the codepath is triggered by error when we build the real ISO we ship)

NicolasT (approved):
Didn't test, but LGTM.

PR #1763, opened 2019-09-23
Reduce code duplication in iso.py

### LINKED ISSUE #1439: Reduce code duplication iso.py

slaperche-scality:
**Component**:

build

**Context**: 

Several tasks in `iso.py` contains code that is very similar.

**Summary**:

Leverage existing targets (such as `FileTree` and `SerialiazedData`) to reduce code duplication.

**Acceptance criteria**: 

Build is working as before.

---

Closes: #1439

### DIFF
--- a/buildchain/buildchain/iso.py
+++ b/buildchain/buildchain/iso.py
@@ -22,7 +22,7 @@ Overview:
                 └────────┘ ╲───>│  salt_tree  │────╱   ^
                         │       └─────────────┘        │
                         │       ┌─────────────┐        │
-                        └──────>│ product.txt │────────┘
+                        └──────>│   iso_tree  │────────┘
                                 └─────────────┘
 """
 
@@ -31,7 +31,7 @@ import datetime as dt
 import socket
 import subprocess
 from pathlib import Path
-from typing import Iterator, Sequence, Tuple
+from typing import Iterator, Tuple
 
 import doit  # type: ignore
 
@@ -42,6 +42,7 @@ from buildchain import targets as helper
 from buildchain import types
 from buildchain import utils
 from buildchain import versions
+from buildchain.targets.serialize import Renderer
 
 
 ISO_FILE : Path = config.BUILD_ROOT/'{}.iso'.format(config.PROJECT_NAME.lower())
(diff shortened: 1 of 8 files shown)


buildchain/buildchain/targets/file_tree.py
```diff
@@ -70,7 +70,10 @@ def __init__(
             for directory
             in self._compute_dir_tree(self.files)
         ]
-        self._root = self.directories[-1].relative_to(self.destination)
+        if not self.directories:
```
jbertran, line 76: can we avoid using properties in `init`? This way there's no need to look at code _after_ the function to figure out what `directories` or `destination` mean - they're simple accessors anyway, so NBD.

buildchain/buildchain/iso.py
```diff
@@ -42,7 +42,6 @@
 from buildchain import types
```
jbertran: Minor: could do with inverting the last 2 commits, so you have the `ENV` renderer available directly via the same namespace.

jbertran (commented):
LGTM! Much simpler indeed.

buildchain/buildchain/iso.py
slaperche-scality: I think there is some logic in this order:
- we use it once more
- we realize it perfectible™
- we add a commit to export it

Instead of:
- I know that it will be painful, let's export it
- Use it once more: see I was right

But I guess both order can be justified, I just prefer this one ^^

buildchain/buildchain/targets/file_tree.py
slaperche-scality: I understand you point.

But I think once you have a property, it's better to always use it (if possible).

That way, if one day the property does some `assert`, logging, type casting or whatever you won't have to screen the code/update the `__init__` code: just change the property.

jbertran (approved):

PR #1774, opened 2019-09-24
Implement the build repository script in the buildchain

### LINKED ISSUE #1746: Implement the build repository script in the buildchain
**Component**: build 

**Why this is needed**: We have a scipt which build an internal Debian repository. We want to call this script from the buildchain

**What should be done**: 
Implement the script in the buildchain

**Test plan**:

ycointe:
**Component**: Build

**Summary**: We want to implement the script which will create Debian repositories

**Acceptance criteria**:
Launch the ISO build with `./doit.sh` then verify that we have all the Debian repositories in `_build/root/packages/debian/`

---

Closes: #1746

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -176,20 +176,20 @@ def task__download_rpm_packages() -> types.TaskDict:
 def task__download_deb_packages() -> types.TaskDict:
     """Download Debian packages locally."""
     def clean() -> None:
-        """Delete cache and repositories on the ISO."""
+        """Delete downloaded Debian packages."""
         for repository in DEB_REPOSITORIES:
             # Repository with an explicit list of packages are created by a
             # dedicated task that will also handle their cleaning, so we skip
             # them here.
             if repository.packages:
                 continue
-            coreutils.rm_rf(repository.rootdir)
+            coreutils.rm_rf(repository.pkgdir)
         constants.REPO_DEB_ROOT.rmdir()
 
     def mkdirs() -> None:
         """Create directories for the repositories."""
         for repository in DEB_REPOSITORIES:
-            repository.rootdir.mkdir(exist_ok=True)
+            repository.pkgdir.mkdir(exist_ok=True)
 
     mounts = [
         utils.bind_ro_mount(
@@ -197,7 +197,7 @@ def task__download_deb_packages() -> types.TaskDict:
             target=Path('/download_packages.py'),
         ),
         utils.bind_mount(
-            source=constants.REPO_DEB_ROOT,
+            source=constants.PKG_DEB_ROOT,
             target=Path('/repositories')
         ),
     ]
@@ -211,7 +211,7 @@ def task__download_deb_packages() -> types.TaskDict:
     return {
         'title': utils.title_with_target1('GET DEB PKGS'),
         'actions': [mkdirs, dl_packages_callable],
-        'targets': [constants.REPO_DEB_ROOT/'.witness'],
+        'targets': [constants.PKG_DEB_ROOT/'.witness'],
         'task_dep': [
             '_package_mkdir_deb_root',
             '_package_mkdir_deb_iso_root',
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -78,6 +78,7 @@ def task_packaging() -> types.TaskDict:
             '_build_rpm_repositories:*',
             '_build_deb_packages:*',
             '_download_deb_packages',
+            '_build_deb_repositories:*',
         ],
     }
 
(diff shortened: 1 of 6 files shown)


buildchain/buildchain/packaging.py
```diff
@@ -176,28 +176,28 @@ def clean() -> None:
 def task__download_deb_packages() -> types.TaskDict:
     """Download Debian packages locally."""
     def clean() -> None:
-        """Delete cache and repositories on the ISO."""
+        """Delete local Debian packages."""
```
slaperche-scality: `Delete downloaded Debian packages`

buildchain/buildchain/packaging.py
```diff
@@ -176,28 +176,28 @@ def clean() -> None:
...
         for repository in DEB_REPOSITORIES:
             # Repository with an explicit list of packages are created by a
             # dedicated task that will also handle their cleaning, so we skip
             # them here.
             if repository.packages:
                 continue
-            coreutils.rm_rf(repository.rootdir)
+            coreutils.rm_rf(constants.PKG_DEB_ROOT/repository.fullname)
```
slaperche-scality: I would add:

```python
def pkgdir(self) -> Path:
    return constants.PKG_DEB_ROOT/self.fullname
```

in the class `DEBRepository` and the use `repository.pkgdir` here and l.192

buildchain/buildchain/packaging.py
```diff
@@ -211,7 +211,7 @@ def mkdirs() -> None:
     return {
         'title': utils.title_with_target1('GET DEB PKGS'),
         'actions': [mkdirs, dl_packages_callable],
-        'targets': [constants.REPO_DEB_ROOT/'.witness'],
+        'targets': [constants.PKG_DEB_ROOT/'.witness'],
```
slaperche-scality, line 218: Commit message should explain why better, like:

```
When building repositories, the packages are copied into the repo hierarchy.
Until now, we were downloading package on the ISO directly (under `_build/root`) but since the repo copy them, we shouldn't download them here (they will be duplicated).
This commit updates the code to download the packages in the "staging" area (under `_build/packages`).
```

Something like that.

buildchain/buildchain/targets/package.py
```diff
@@ -349,7 +349,7 @@ def __init__(
     def deb(self) -> Path:
         """DEB path."""
         fmt = '{pkg.name}_{pkg.version}-{pkg.build_id}_{pkg.ARCH}.deb'
-        return self.rootdir/fmt.format(pkg=self)
+        return constants.PKG_DEB_ROOT/fmt.format(pkg=self)
```
jbertran, line 352: In the final staging area, this looks odd since we have every other `.deb` that was downloaded set inside its directory (so `_build/packages/metalk8s-bionic/*.deb` etc), but the built packages are set in `_build/packages`.

Could we make that `constants.PKG_DEB_ROOT / 'metalk8s-scality' / fmt.format(pkg=self)` ?

buildchain/buildchain/targets/package.py
```diff
@@ -375,7 +372,7 @@ def build_package(self) -> types.TaskDict:
                 source=self.debuild_sources, target=Path('/debbuild/pkg-meta')
             ),
             utils.bind_mount(
-                source=self.rootdir, target=Path('/debbuild/result')
+                source=constants.PKG_DEB_ROOT, target=Path('/debbuild/result')
```
jbertran, line 375: From my comment above, this one would change as well.

buildchain/buildchain/targets/package.py
```diff
@@ -404,14 +401,10 @@ def build_package(self) -> types.TaskDict:
...
-                source=self.sources,
-                target=Path('/rpmbuild/source.rpm')
-            ),
-            utils.bind_mount(
-                source=self.rootdir,
-                target=Path('/debbuild/result')
-            ),
+            utils.bind_ro_mount(source=self.sources,
+                                target=Path('/rpmbuild/source.rpm')),
+            utils.bind_mount(source=constants.PKG_DEB_ROOT,
```
jbertran, line 406: And this one.

buildchain/buildchain/targets/repository.py
```diff
@@ -331,10 +331,14 @@ def fullname(self) -> str:
     @property
     def pkgdir(self) -> Path:
         """Repository where to download the packages."""
+        if self.packages:
+            # Built packages are not under a sub-directory.
```
jbertran, line 335: But they should be, I guess.

jbertran (commented):
A comment on the built packages staging area.

From what I gather you handled the `.witness` deletion issue while I was reviewing.

buildchain/buildchain/targets/package.py
ycointe: It will be handled by the issue #1781

buildchain/buildchain/targets/repository.py
ycointe: It will be handled by the issue #1781

buildchain/buildchain/targets/package.py
ycointe: It will be handled by the issue #1781

buildchain/buildchain/targets/package.py
ycointe: It will be handled by the issue #1781

slaperche-scality (approved):

PR #1778, opened 2019-09-25
Reinstate 'node-exporter-full' dashboard

### LINKED ISSUE #1688: Re-introduce node-exporter-full and CoreDNS dashboards
In #1672, some additional dashboards we added before (in 11f0c8b218872977d57705063d14c88189df73f5 and d0ed38a26d006d90e649a676ba79ff635da8926d) were temporarily removed. These should be added again.

NicolasT:
Fixes: #1688

### DIFF
--- a/salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls
@@ -1,6 +1,11 @@
 #!jinja | kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
 
 {%- set dashboards = [
+        {
+            'name': 'node-exporter-full',
+            'title': 'Nodes (Detailed)',
+            'tags': ['nodes'],
+        },
     ]
 %}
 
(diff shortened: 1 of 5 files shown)


salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls
```diff
@@ -0,0 +1,34 @@
+#!jinja | kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+
+{%- set dashboards = [
```
sayf-eddine-scality, line 3: We re saving data into an sls file, should we create a debt ticket for this to move it somewhere else more maintainable ?

sayf-eddine-scality (approved):
one comment maybe better to be tracked in a debt ticket, lets push beta2 forward

salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls
NicolasT: It's not really 'data' IMHO, more of an `import ... with context` or rather `include`... Don't think 'magically' deploying all `files/*.json` dashboards would be wise, because the transform function may not suit all possible dashboard inputs (in the MetalK8s 1.x `hack/fix-dashboard.py` or something script, there's a whole bunch of quirks).

So, not sure... What would you think would be 'more maintainable'?

PR #1782, opened 2019-09-25
ui/bug: enable to click on trash button when there is no pv

ChengYanJin:
**Component**: ui

**Context**: 
When the status of volume is failed and no PV attached, we should be able to click on the trash button.

**Summary**:

**Acceptance criteria**: 
When there is no pv and status of volume is failed, the trash button should be enable and without hint message.
image

Fix this bug
Refs: #1673


slaperche-scality (dismissed):

slaperche-scality (approved):

ChengYanJin:
@bert-e approve

PR #1783, opened 2019-09-25
SolutionUI:  improvements

carlito-scality:
**Component**:SolutionUI

**Context**: 
- rework Solution UI to make it more user friendly

**Summary**:

**Acceptance criteria**: 
image
image
image
image

---

Closes: #ISSUE_NUMBER


examples/metalk8s-solution-example/ui/src/containers/Component.js
```diff
@@ -0,0 +1,217 @@
...
+};
+
+export default injectIntl(
+  withRouter(
+    connect(
+      mapStateToProps,
+      mapDispatchToProps
+    )(Component)
+  )
+);
```
gdemonet, line 217: Maybe use a more explicit name for this, such as `ComponentLists` or `ComponentTables`, even `ComponentsGroup` if you'd like. But since we'd like this code to be easily copy-pasted by other Solution developers, good naming is useful :)

MonPote:
When an environment does not have a `version`, we should not be able to create a `cr`.

examples/metalk8s-solution-example/ui/src/containers/Environment.js
```diff
@@ -0,0 +1,186 @@
...
+
+const ModalBody = styled.div`
+  padding-bottom: ${padding.base};
+`;
+
+const ModalBodyTitle = styled.div`
+  padding-bottom: ${padding.base};
+`;
+
+const Environment = props => {
```
gdemonet, line 44: Same for this one, I guess `EnvironmentsList` would be more explicit

examples/metalk8s-solution-example/ui/src/containers/ClockServerCreation.js
```diff
@@ -0,0 +1,237 @@
...
+
+import { isVersionSupported } from '../services/utils';
+import { createClockServerAction } from '../ducks/app/clockServer';
+import {
+  BreadcrumbContainer,
+  BreadcrumbLabel,
+  StyledLink
+} from '../components/BreadcrumbStyle';
+
+const CreateClockServerContainter = styled.div`
```
MonPote: I guess it's `CreateClockServerContainer` instead of `CreateClockServerContainter `

examples/metalk8s-solution-example/ui/src/containers/ClockServerEdit.js
```diff
@@ -0,0 +1,253 @@
...
+        width: 200px;
+        box-sizing: border-box;
+      }
+    }
+  }
+`;
+
+const InputLabel = styled.label`
+  width: 150px;
+  padding: 10px;
```
MonPote: I guess we want `${padding.small}`

examples/metalk8s-solution-example/ui/src/containers/ClockServerEdit.js
```diff
@@ -0,0 +1,253 @@
...
+import semver from 'semver';
+import { editClockServerAction } from '../ducks/app/clockServer';
+import {
+  BreadcrumbContainer,
+  BreadcrumbLabel,
+  StyledLink
+} from '../components/BreadcrumbStyle';
+import { isVersionSupported } from '../services/utils';
+
+const CreateClockServerContainter = styled.div`
```
MonPote: `CreateClockServerContainer` I think

examples/metalk8s-solution-example/ui/src/containers/Layout.js
```diff
@@ -101,20 +101,35 @@ const Layout = props => {
           <PrivateRoute exact path="/about" component={Welcome} />
           <PrivateRoute
             exact
-            path="/customResource"
-            component={CustomResource}
+            path="/environments/:name/clockServer/create"
```
gdemonet, line 104: I think we should keep case-insensitive URLs:

```suggestion
            path="/environments/:name/clock-servers/create"
```

examples/metalk8s-solution-example/ui/src/containers/Environment.js
```diff
@@ -0,0 +1,186 @@
...
+      dataKey: 'version',
+      renderer: (data, rowData) => {
+        if (!data) {
+          return (
+            <Button
+              text={intl.messages.add_solution}
+              size="small"
+              type="button"
+              onClick={e => {
+                e.stopPropagation();
```
MonPote, line 80: Do we really need `e.stopPropagation()` ? You already specify the type.

examples/metalk8s-solution-example/ui/src/containers/Layout.js
```diff
@@ -101,20 +101,35 @@ const Layout = props => {
...
-            path="/customResource/create"
-            component={CustomresourceCreation}
+            path="/environments/:name/versionServer/:id/edit"
+            component={VersionServerEdit}
+          />
+          <PrivateRoute
+            exact
+            path="/environments/:name/version/:version/prepare"
+            component={EnvironmentPreparation}
+          />
```
gdemonet, line 126: Not super fond of this route... Is it feasible to do it in two steps?

1. Save desired version in the `Environment` object itself
2. Go to `/environments/:name/prepare` to run the preparation step

Would make it safer to go back on this page, and would also allow unifying the preparation and upgrade IMO

examples/metalk8s-solution-example/ui/src/containers/VersionServerCreation.js
```diff
@@ -42,32 +51,33 @@ const ActionContainer = styled.div`
...
-  const { intl, namespaces } = props;
+const VersionServerCreationForm = props => {
+  const { intl, match, config, environments } = props;
+  const environment = match.params.name;
+  const currentEnvironment = environments.find(
+    item => item.name === environment
+  );
+  const currentEnvironmentVersion = currentEnvironment
+    ? currentEnvironment.version
+    : '';
```
gdemonet, line 72: Can't use the optional chaining syntax here instead?

```javascript
const currentEnvironmentVersion = currentEnvironment?.version ?? '';
```

examples/metalk8s-solution-example/ui/src/containers/VersionServerEdit.js
```diff
@@ -74,24 +78,25 @@ const InputValue = styled.label`
...
-      ? customResource.namespace
-      : namespaces.length
-      ? namespaces[0].metadata.name
-      : '',
-    version: customResource ? customResource.version : '',
-    replicas: customResource ? customResource.replicas : '',
-    name: customResource ? customResource.name : ''
+    version: versionServer ? versionServer.version : '',
+    replicas: versionServer ? versionServer.replicas : '',
+    name: versionServer ? versionServer.name : '',
```
gdemonet, line 95: Could use the optional chaining syntax for those

examples/metalk8s-solution-example/ui/src/containers/VersionServerEdit.js
```diff
@@ -74,24 +78,25 @@ const InputValue = styled.label`
...
-    namespaces: customResource
-      ? customResource.namespace
-      : namespaces.length
-      ? namespaces[0].metadata.name
-      : '',
-    version: customResource ? customResource.version : '',
-    replicas: customResource ? customResource.replicas : '',
-    name: customResource ? customResource.name : ''
+    version: versionServer ? versionServer.version : '',
+    replicas: versionServer ? versionServer.replicas : '',
```
gdemonet, line 94: Default value should be an integer I believe, is it some way for Formik to handle an unset field?

examples/metalk8s-solution-example/ui/src/containers/VersionServerEdit.js
```diff
@@ -5,29 +5,38 @@ import { Formik, Form } from 'formik';
...
-const CreateCustomresourceContainter = styled.div`
+import { editVersionServerAction } from '../ducks/app/versionServer';
+import {
+  BreadcrumbContainer,
+  BreadcrumbLabel,
+  StyledLink
+} from '../components/BreadcrumbStyle';
+import { isVersionSupported } from '../services/utils';
+
+const CreateVersionServerContainter = styled.div`
```
gdemonet, line 20: Oops:

```suggestion
const CreateVersionServerContainer = styled.div`
```

examples/metalk8s-solution-example/ui/src/ducks/app/clockServer.js
```diff
@@ -0,0 +1,154 @@
...
+
+import { REFRESH_TIMEOUT } from '../../constants';
+
+// Actions
+const REFRESH_CLOCK_SERVER = 'REFRESH_CLOCK_SERVER';
+const STOP_REFRESH_CLOCK_SERVER = 'STOP_REFRESH_CLOCK_SERVER';
+const CREATE_CLOCK_SERVER = 'CREATE_CLOCK_SERVER';
+
+const UPDATE_CLOCK_SERVER = 'UPDATE_CLOCK_SERVER';
+const EDIT_CLOCK_SERVER = 'EDIT_CLOCK_SERVER';
```
gdemonet, line 13: IIRC, a perk of using well isolated _ducks_ is to name those constants with shorter names:

```javascript
const REFRESH = 'REFRESH_CLOCK_SERVER';
const STOP_REFRESH = 'STOP_REFRESH_CLOCK_SERVER';
const CREATE = 'CREATE_CLOCK_SERVER';
const EDIT = 'EDIT_CLOCK_SERVER';
const UPDATE = 'UPDATE_CLOCK_SERVER';
```

Which makes the usual `actionCreators` and `reducer` quite simple to generate (or copy-paste).
If you ever need to import them from another duck (should not happen often, but it can), export it in the source:

```javascript
const REFRESH = 'REFRESH_CLOCK_SERVER';
export { REFRESH as REFRESH_CLOCK_SERVER };
```

MonPote (dismissed):
A few remark in general : 
- We might want to use `optional chaining` to shorten a bit the code and prevent a few bug
- Need to add test to sagas and `isVersionSupported` function
- We might want to start using hooks for `react-router` and `redux` soon

All those remarks are non-blocker IMO so I approve.
Good Job :)

examples/metalk8s-solution-example/ui/src/ducks/app/clockServer.js
```diff
@@ -0,0 +1,154 @@
...
+      })
+    );
+  }
+  return results;
+}
+
+export function* createClockServer({ payload }) {
+  const { name, environment, timezone, version } = payload;
+  const body = {
+    apiVersion: 'example-solution.metalk8s.scality.com/v1alpha1',
```
gdemonet, line 74: Ideally (worth tracking in another issue), we should store this value (and others, such as the `SOLUTION_NAME`, which is `'example-solution'` in this case) as constants and share them where necessary, to avoid boring renaming steps if ever needed.

examples/metalk8s-solution-example/ui/src/ducks/app/deployment.js
```diff
@@ -1,111 +1,132 @@
...
-import history from '../../history';
 
 // Constants
 const LABEL_COMPONENT = 'app.kubernetes.io/component';
 const LABEL_NAME = 'app.kubernetes.io/name';
 export const LABEL_PART_OF = 'app.kubernetes.io/part-of';
-const LABEL_VERSION = 'app.kubernetes.io/version';
-const SOLUTION_NAME = 'example-solution';
+export const LABEL_VERSION = 'app.kubernetes.io/version';
+export const SOLUTION_NAME = 'example-solution';
```
gdemonet: Could we have some re-usable constants in a dedicated file (see comment above), e.g. `src/ducks/constants.js`?

examples/metalk8s-solution-example/ui/src/ducks/app/deployment.js
```diff
@@ -1,111 +1,132 @@
...
-      name: item.metadata.name,
-      namespace: item.metadata.namespace,
-      image: item.spec.template.spec.containers['0'].image,
-      version:
-        (item.metadata.labels && item.metadata.labels[LABEL_VERSION]) || ''
-    }));
-
-    yield put(updateDeploymentAction({ list: flattenedItems }));
-  }
+export function* fetchOpertorDeployments(namespaces) {
```
gdemonet, line 26: Oops:

```suggestion
export function* fetchOperatorDeployments(namespaces) {
```

examples/metalk8s-solution-example/ui/src/ducks/app/environment.js
```diff
@@ -0,0 +1,232 @@
...
+  }
+  return results;
+}
+export function* updateEnvironment(environment) {
+  const results = yield call(
+    fetchOpertorDeployments,
+    `${environment.metadata.name}-example-solution`
+  );
+  // One operator per environment
+  const operator = results.body.items.length ? results.body.items[0] : null;
```
gdemonet, line 97:
```suggestion
  const operator = results?.body?.items?.[0];
```

examples/metalk8s-solution-example/ui/src/ducks/app/environment.js
```diff
@@ -0,0 +1,232 @@
...
+  yield put(
+    addEnvironmentAction({
+      name: environment.metadata.name,
+      status: '', //TO be implemented
+      description: environment.spec.description,
+      version:
+        operator && operator.metadata.labels
+          ? operator.metadata.labels[LABEL_VERSION]
+          : '',
```
gdemonet, line 106: Could be:

```javascript
version: operator?.metadata.labels?.[LABEL_VERSION] ?? '',
```

examples/metalk8s-solution-example/ui/src/ducks/app/environment.js
```diff
@@ -0,0 +1,232 @@
...
+      solutions: environment.spec.solutions || [],
+      operator: operator
+        ? {
+            name: operator.metadata.name,
+            namespace: operator.metadata.namespace,
+            image: operator.spec.template.spec.containers['0'].image,
```
gdemonet, line 112: Why `containers['0']` instead of `containers[0]`?

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -37,40 +38,98 @@ export const updateApiServerConfig = (url, token) => {
   rbacAuthorizationV1Api = config.makeApiClient(RbacAuthorizationV1Api);
 };
 
-export async function getCustomResource() {
+export async function getEnvironment() {
```
gdemonet, line 40: This method would be better named `listEnvironments`

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -37,40 +38,98 @@ export const updateApiServerConfig = (url, token) => {
...
+          'Content-Type': 'application/merge-patch+json'
+        }
+      }
+    );
+  } catch (error) {
+    return { error };
+  }
+}
+
+export async function getClockServer(namespaces) {
```
gdemonet, line 71: This method would be better named `listClockServers`, and its argument should be named `namespace`, not `namespaces`, to avoid confusion as to what kind of value should be passed.

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -37,40 +38,98 @@ export const updateApiServerConfig = (url, token) => {
...
+      namespaces,
+      'clockservers'
     );
   } catch (error) {
     return { error };
   }
 }
 
-export async function createCustomResource(body, namespaces) {
+export async function getVersionServer(namespaces) {
```
gdemonet, line 84: Ditto

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -37,40 +38,98 @@ export const updateApiServerConfig = (url, token) => {
...
+      'v1alpha1',
+      namespaces,
+      'versionservers'
+    );
+  } catch (error) {
+    return { error };
+  }
+}
+
+export async function createClockServer(body, namespaces) {
```
gdemonet, line 97: Single `namespace`, everywhere please :)

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -84,26 +143,34 @@ export async function updateCustomResource(body, namespaces, name) {
...
   } catch (error) {
     return { error };
   }
 }
 
-export async function getSolutionDeployment(version) {
+export async function getOperatorDeployments(namespaces, name) {
   try {
-    return await appsV1Api.listDeploymentForAllNamespaces(
+    return await appsV1Api.listNamespacedDeployment(
```
gdemonet, line 167: Why not use the `readNamespacedDeployment` method? `namespace` and `name` are enough to find the object I believe: https://github.com/kubernetes-client/javascript/blob/master/src/gen/api/appsV1Api.ts#L3912-L3920

You shouldn't need the `part-of` label filter anyway

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -119,6 +186,20 @@ export async function createNamespacedDeployment(namespaces, body) {
   }
 }
 
+export async function listNamespacedServiceAccount(namespaces, name) {
```
gdemonet, line 188: As above, if you have the `name`, don't do unnecessary `list` + `labelSelector`, just use the `read` method from the client.

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -127,6 +208,20 @@ export async function createNamespacedServiceAccount(namespaces, body) {
   }
 }
 
+export async function listNamespacedRole(namespaces, name) {
```
gdemonet, line 210: Ditto

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -135,6 +230,20 @@ export async function createNamespacedRole(namespaces, body) {
   }
 }
 
+export async function listNamespacedRoleBinding(namespaces, name) {
```
gdemonet, line 232: Ditto

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
```diff
@@ -164,3 +273,34 @@ export async function updateDeployment(body, namespaces, name) {
...
+      null,
+      `metadata.name=${name}`
+    );
+  } catch (error) {
+    return { error };
+  }
+}
+export async function getSolutionsConfigMap() {
+  try {
+    return await coreV1.listConfigMapForAllNamespaces(
```
gdemonet, line 298: Would be OK to use `read` and rely on knowledge of this CM's namespace (`metalk8s-solutions`), since we also know its name and existence.

examples/metalk8s-solution-example/ui/src/services/utils.js
```diff
@@ -30,3 +31,40 @@ export const useRefreshEffect = (refreshAction, stopRefreshAction) => {
     };
   }, [dispatch, refreshAction, stopRefreshAction]);
 };
+
+/* Check if a version is supported by comparaing it to the 'environmentVersion'
+   Here we decide that all the  "preminor of 1" versions are allowed
+   Example :
+   const versions = [
+              { version: '1.1.4-prod' },
```
gdemonet, line 39: Such pre-release _suffix_ wouldn't be used I guess, something like `-beta2` or `-rc1` would be more explicit)

examples/metalk8s-solution-example/ui/src/containers/Environment.js
carlito-scality: Yes. Otherwise Table's onRowClick will be called

examples/metalk8s-solution-example/ui/src/services/utils.js
```diff
@@ -30,3 +31,40 @@ export const useRefreshEffect = (refreshAction, stopRefreshAction) => {
...
+      environmentVersionMinor > 1 ? environmentVersionMinor - 1 : 0
+    }.0`;
+
+    return (
+      semver.gte(item.version, minimunSupportedVersion) &&
+      semver.lte(item.version, parsedEnvironmentVersion)
+    );
+  } else {
+    return true;
+  }
```
gdemonet: I'd say we can come up with a less verbose implementation, using `semver.parse`. Here's what I would imagine:

```javascript
export const isVersionSupported = envVersionStr => item => {
  let envVersion = semver.parse(envVersionStr);
  envVersion.prerelease = [];  // Reset pre-release info
  if (envVersionObj) {
    const minVersion = `${envVersion.major}.${Math.max(envVersion.minor - 1, 0)}`
    const versionRange = `${minVersion} - ${envVersion.format()}`
    return semver.statisfies(item.version, versionRange);
  } else {
    return true;
  }
}
```

examples/metalk8s-solution-example/ui/src/translations/en.json
```diff
@@ -5,19 +5,32 @@
...
+  "create_version_server": "Create a Version Server",
+  "create_clock_server": "Create a Clock Server",
+  "edit_version_server": "Edit a Version Server",
+  "edit_clock_server": "Edit a Clock Server",
+  "no_component_available": "No Component Available",
+  "components": "Components",
+  "information": "Information",
+  "kind": "Kind",
+  "edit_environment_version": "Edit Environment Version",
+  "timezone": "Time Zone",
```
gdemonet, line 30: This can be a single word

examples/metalk8s-solution-example/ui/src/translations/fr.json
```diff
@@ -5,19 +5,14 @@
...
   "edit": "Éditer",
-  "namespace": "Namespace",
   "version": "Version",
   "replicas": "Duplication",
   "no_data_available": "Pas de données disponibles",
-  "custom_resource": "Custom Resource",
-  "select_a_namespace": "Sélectionner un namespace…",
   "not_found": "Aucun résultat",
   "not_valid_version": "Ce n'est pas une version valide"
 }
```
gdemonet, line 18: Seems like we are missing French translations for the newly added texts?

examples/metalk8s-solution-example/ui/src/translations/en.json
```diff
@@ -5,19 +5,32 @@
...
+  "prepare_environment": "Preparing environment...",
+  "select_a_version": "Select a Version",
```
gdemonet: No need for a capital `V` in front of `version`.

gdemonet (changes requested):
Bunch of changes to clean up the code, otherwise most of it looks good! Some improvements can be relegated to a future PR if that's easier for you.

I'd really like if we can tackle the following points before merging:
- [x] Reword the "Add Solution" button/modal on the main `/environments` page (use "Prepare" instead)
- [x] Clean the SemVer code to make it more obvious
- [x] Move constants into a single file, for Solution developers to just change them in a few places without much effort

MonPote (approved):

carlito-scality:
I just found a bug relative to the refresh of clockServer and versionServer. When we switch from ```environment 1``` to ```environment 2```, there are 2 refresh of clockServer (and 2 refresh of versionServer) running instead of one in the ```environment 2``` since the refresh of the  ```environment 1``` is not cancelled correctly. Please see the ticket below https://github.com/scality/metalk8s/issues/1820

examples/metalk8s-solution-example/ui/src/translations/fr.json
carlito-scality: Yes. because I18n is not implemented in the UI yet.

carlito-scality:
@bert-e approve

gdemonet (approved):

examples/metalk8s-solution-example/ui/src/services/k8s/api.js
carlito-scality: readNamespacedDeployment works as well. However, when the deployment does not exist, it returns a 404 error whereas listNamespacedDeployment return a 200 with an empty list. We use it to populate the environment state, so we do not want to deal with the error.

gdemonet: Mmh, you're not using this Deployment to fill an array in the state, this is, as you said, an optional attribute for each Environment. So it could(should) try to get the object and return nothing if a caught error has `status_code == 404`, failing otherwise.

examples/metalk8s-solution-example/ui/src/containers/Layout.js
carlito-scality: See https://github.com/scality/metalk8s/issues/1917

Issue #1726, opened 2019-09-19
Configure Debian repository on bootstrap node


ycointe:
**Component**: Bootstrap

**Why this is needed**: To deploy a bootstrap on an Ubuntu node, we need to configure our Debian repository to provide the required packages.

**What should be done**: 
Modify the bootstrap.sh.in script in the buildchain to configure the Debian repository in local for the Ubuntu bootstrap node according to the os.

**Implementation proposal** (strongly recommended):
implement a solution to be sur that the script do the os selection correctly .
Create a function for Debian repository configuration on an Ubuntu bootstrap node according to the os.

**Test plan**: Launch the bootstrap deployment on an Ubuntu node then verify our Debian repository is well configured.

PR #1786, opened 2019-09-25
Add Ubuntu bootstrap VM in Vagrant

jbertran:
**Component**: local build

**Context**: #1726 is the tipping point at which we'd like to start testing the bootstrap process on an Ubuntu node. We provide an Ubuntu bootstrap VM in Vagrant that will serve hereafter for local tests on Ubuntu.

CI testing should follow when an Ubuntu bootstrap is fully functional.

**Summary**:

* Refactor bootstrap node creation into a separate function
* Provide an Ubuntu bootstrap VM
* Allow the definition of custom scripts to be called for provisioning depending on OS.

**Acceptance criteria**: 

The Ubuntu bootstrap node can be provisioned (and predictably fails early on the bootstrap for now).

```
vagrant up bootstrap_ubuntu
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -152,6 +152,8 @@ if ! grep -Fxq "$(cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub)" .ssh/authorized_keys 
 fi
 SCRIPT
 
+INSTALL_PYTHON = 'DEBIAN_FRONTEND=noninteractive apt install python -yq'
+
 # To support VirtualBox linked clones
 Vagrant.require_version(">= 1.8")
 
@@ -163,6 +165,13 @@ def declare_bootstrap(machine, os_data)
   machine.vm.provider "virtualbox" do |v|
     v.memory = 4086
     machine.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+  end
+
+  if os_data.fetch(:scripts, []).each do |script|
+       machine.vm.provision script[:name],
+         type: script[:type],
+         inline: script[:data]
+     end
   end
 
   machine.vm.provision "import-release",
@@ -193,14 +202,17 @@ Vagrant.configure("2") do |config|
     },
     ubuntu: {
       name: 'ubuntu/bionic64',
-      version: '20190514.0.0'
+      version: '20190514.0.0',
+      scripts: [
+        {
+          name: 'install-python',
+          type: 'shell',
+          data: INSTALL_PYTHON
+        }
+      ]
     }
   }
 
-  INSTALL_PYTHON = <<-SCRIPT
-  #!/bin/bash
-  apt install python -y
-  SCRIPT
   config.vm.box = os_data[:centos][:name]
   config.vm.box_version = os_data[:centos][:version]
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -155,9 +155,54 @@ SCRIPT
 # To support VirtualBox linked clones
 Vagrant.require_version(">= 1.8")
 
+def declare_bootstrap(machine, os_data)
+  machine.vm.box = os_data[:name]
+  machine.vm.box_version = os_data[:version]
+
+  machine.vm.hostname = "bootstrap"
+  machine.vm.provider "virtualbox" do |v|
+    v.memory = 4086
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -155,50 +155,37 @@ SCRIPT
...
+    inline: CREATE_VOLUMES
 
-    bootstrap.vm.provision "import-ssh-private-key",
-      type: "shell",
-      inline: IMPORT_SSH_PRIVATE_KEY
 
-    bootstrap.vm.provision "bootstrap",
-      type: "shell",
-      inline: BOOTSTRAP
+  return machine
```
slaperche-scality: `machine` is enough, explicit return are not idiomatic in Ruby: https://github.com/rubocop-hq/ruby-style-guide#no-explicit-return

Vagrantfile
```diff
@@ -194,14 +206,17 @@ Vagrant.configure("2") do |config|
     },
     ubuntu: {
       name: 'ubuntu/bionic64',
-      version: '20190514.0.0'
+      version: '20190514.0.0',
+      scripts: [
+        {
+          name: "install-python",
+          type: "shell",
+          data: INSTALL_PYTHON
```
slaperche-scality, line 210: Since the script is a one-liner, you could write `data: 'apt install python -y'` directly.

Vagrantfile
```diff
@@ -194,14 +206,17 @@ Vagrant.configure("2") do |config|
     },
     ubuntu: {
       name: 'ubuntu/bionic64',
-      version: '20190514.0.0'
+      version: '20190514.0.0',
+      scripts: [
+        {
+          name: "install-python",
```
slaperche-scality: Use single-quote (since it's what we are using above): https://github.com/rubocop-hq/ruby-style-guide#consistent-string-literals

(would really love to run `rubocop` on this file in the CI, but would need a tailored config since it's Ruby used as a DSL).

Vagrantfile
```diff
@@ -165,6 +170,14 @@ def declare_bootstrap(machine, os_data)
     machine.vm.synced_folder ".", "/vagrant", type: "virtualbox"
   end
 
+  if os_data.has_key?(:scripts)
```
slaperche-scality: Can be written as:

```ruby
os_data.fetch(:script, []).each do |script|
  machine.vm.provision script[:name],
    type: script[:type],
    inline: script[:data]
end
```

slaperche-scality (changes requested):

Vagrantfile
```diff
@@ -152,53 +152,52 @@ if ! grep -Fxq "$(cat .ssh/#{PRESHARED_SSH_KEY_NAME}.pub)" .ssh/authorized_keys
 fi
 SCRIPT
 
+INSTALL_PYTHON = <<-SCRIPT
+  #!/bin/bash
```
NicolasT: Why this indentation?

Vagrantfile
```diff
@@ -207,14 +206,43 @@ Vagrant.configure("2") do |config|
...
+  config.vm.network "private_network",
+    type: :dhcp,
+    nic_type: 'virtio',
+    **WORKLOAD_PLANE_NETWORK
+
+  config.vm.define :bootstrap, primary: true do |machine|
+    declare_bootstrap machine, os_data[:centos]
+  end
+
+  config.vm.define :bootstrap_ubuntu do |machine|
```
NicolasT: Won't we need `autostart: false` on this one as well?

Vagrantfile
jbertran: We use the same for regular Ubuntu nodes too

Vagrantfile
jbertran: Gah it's inconsistent in this file, I was matching what's in the other `vm.provision` declarations

Vagrantfile
jbertran: I figured there'd be a ruby way of `.get(key, default)`. Thanks

Vagrantfile
jbertran: Fixed by addressing @slaperche-scality 's comments.

Vagrantfile
jbertran: Oops indeed.

slaperche-scality (approved):

PR #1787, opened 2019-09-25
containerd: set `LimitNOFILE` on the service

### LINKED ISSUE #1785: containers' file limitation is too low
**containers**:

**What happened**:
When we deploy HD thru our k8s controller, we have a very small limitation of max open files allowed to the process run in the container.
The current limitation allowed is only 1014, which is way too low for processed implied in HDs 

It seems that containerd does not allow to change ulimit values
```
ulimit -n 819200
bash: ulimit: open files: cannot modify limit: Operation not permitted
```

Also, trying to use the operator-sdk API to add additionnal parameters to pod deployment doesn't work because they are _unsafe_

```
diff --git a/pkg/controller/hdservice/hdcontroller/hdcontroller.go b/pkg/controller/hdservice/hdcontroller/hdcontroller.go
index 77adc07..3475843 100644
--- a/pkg/controller/hdservice/hdcontroller/hdcontroller.go
+++ b/pkg/controller/hdservice/hdcontroller/hdcontroller.go
@@ -240,6 +240,9 @@ func NewDeployment(cr *v1alpha1.HdService, ID, cMapName, cType, relName string)
                                                },
                                        },
                                        Volumes: volumes,
+                                       SecurityContext: &corev1.PodSecurityContext{
+                                               Sysctls: []corev1.Sysctl{corev1.Sysctl{Name: "fs.file-max", Value: "100000"}},
+                                       },
                                },
                        },
                },

```
```
kubectl get pod
NAME                                      READY   STATUS             RESTARTS   AGE
default-http-backend-5d5ff5d4f5-8j7v7     1/1     Running            28         8d
....
my-hdservice-monitor-1-7b8968497b-l8dsf   0/1     **SysctlForbidden**    0          1s

```

**What was expected**:

To have a way to change this value, either globally, or directly in the HD operator

NicolasT:
The default 'open files' limit, 1024, is way too low for some of our
applications. Since there's no way to bump this limit from within a
container, and the artificial limit of 1024 file descriptors stems from
an era where FDs were costly, bump the limit to a more reasonable (in
2019) default.

Manual test:

```
[root@bootstrap ~]# kubectl taint node bootstrap node-role.kubernetes.io/bootstrap-
node/bootstrap untainted
[root@bootstrap ~]# kubectl taint node bootstrap node-role.kubernetes.io/infra-
node/bootstrap untainted
[root@bootstrap ~]# kubectl run --restart=Never --image=busybox shell -- sleep 9999
pod/shell created
[root@bootstrap ~]# kubectl exec -ti shell -- sh -c 'ulimit -n'
1048576
```

Fixes: #1785
See: https://github.com/scality/metalk8s/issues/1785
See: https://github.com/containerd/containerd/issues/3201
See: https://github.com/containerd/containerd/pull/3202

### DIFF
--- a/salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf
+++ b/salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf
@@ -1,0 +1,4 @@
+[Service]
+# See https://github.com/containerd/containerd/issues/3201
+# See https://github.com/containerd/containerd/pull/3202
+LimitNOFILE=1048576
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -255,6 +255,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/nginx-ingress/deployed/namespace.sls'),
 
     Path('salt/metalk8s/container-engine/containerd/configured.sls'),
+    Path('salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf'),
     Path('salt/metalk8s/container-engine/containerd/init.sls'),
     Path('salt/metalk8s/container-engine/containerd/installed.sls'),
     Path('salt/metalk8s/container-engine/init.sls'),
--- a/salt/metalk8s/container-engine/containerd/configured.sls
+++ b/salt/metalk8s/container-engine/containerd/configured.sls
@@ -11,6 +11,7 @@ Start and enable containerd:
       - metalk8s_package_manager: Install containerd
     - watch:
       - file: Configure registry IP in containerd conf
+      - file: Create containerd service drop-in
 
 Inject pause image:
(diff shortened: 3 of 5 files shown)


NicolasT:
~Need to test this first.~

slaperche-scality (dismissed):

NicolasT:
I'll approve once @fferrandis signed off in the issue :)

salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf
```diff
@@ -0,0 +1,4 @@
+[Service]
+# See https://github.com/containerd/containerd/issues/3201
+# See https://github.com/containerd/containerd/pull/3202
+LimitNOFILE=1048576
```
fferrandis: <3

fferrandis (approved):

fferrandis:
Thank you team !

slaperche-scality (approved):

PR #1791, opened 2019-09-26
Use conditions for volume (instead of Phase, ErrorCode and ErrorMessage)

### LINKED ISSUE #1507: Use conditions instead of error code/message for Volume
**Component**:

storage

**Why this is needed**:

In order to be more future-proof (it's easier to extend) and allows the UI to gives better feedback, we want to move away from the traditional error code/message pair to use a list of "conditions" instead.

**What should be done**:

Replace the error code/error message pair by a list of conditions (like what's done for `Node` object).
This may also supersedes the `Phase` field.

**Implementation proposal**:

## Changes to the Volume schema

New definition for `VolumeStatus`: we replace `Phase`, `ErrorCode` and `ErrorMessage` by a list of `VolumeCondition`.

``` go
type VolumeStatus struct {
    Conditions []VolumeCondition
    Job string
}
```

A `VolumeCondition` is defined as follow (straight copy from `NodeCondition`, see [here](https://godoc.org/k8s.io/api/core/v1#NodeCondition)):

```go
type VolumeCondition struct {
    // Type of volume condition.
    Type VolumeConditionType
    // Status of the condition, one of True, False, Unknown.
    Status ConditionStatus
    // Last time the condition was updated (optional).
    LastUpdateTime metav1.Time
    // Last time the condition transit from one status to another (optional)
    LastTransitionTime metav1.Time
    // Unique, one-word, CamelCase reason for the condition's last transition.
    Reason string
    // Human readable message indicating details about last transition.
    Message string
}
```

Given the short delay to make this new schema land, we will go with a single `VolumeConditionType` for now:

```go
type VolumeConditionType string

const (
    // VolumeReady means Volume is ready to be used.
    VolumeReady VolumeConditionType = "Ready"
)
```

Later on (i.e. in other issues), we will add more fine-grained conditions (and improve the UI to leverage those extra information).

No version bump for the CRD.

## Impact on the operator

Update code to update `Conditions` instead of `Phase`, `ErrorCode` and `ErrorMessage`.

## Impact on the Salt module

None.

## Impact on the UI

Code that uses the fields `status.phase`, `status.errorCode` or `status.errorMessage` will need to be adapted to work with `status.conditions` instead.

Since we won't exploit yet the whole power of `Conditions` we can keep the current paradigm in order to minimize the required code changes.
To do so, we will add a set of simple helper functions that extract/compute a `phase`, `errorCode` and `errorMessage` from a Volume object using the `Conditions` field.
Then we can use those functions where we are accessing the removed fields.

| `status.phase` | `Conditions[Ready]`                       |
| ---            | ---                                       |
| `Unknown`      | missing                                   |
| `Failed`       | `False`                                   |
| `Pending`      | `Unknown` + `Reason` ==  `Pending` |
| `Available`    | `True`                                    |
| `Terminating`  | `Unknown` + `Reason` == `Terminating`      |

For `status.errorCode` we can read the `Reason` field of the `VolumeCondition`.
For `status.errorMessage` we can read the `Message` field of the `VolumeCondition`.

**Test plan**:

Existing tests should work as before.

slaperche-scality:
**Component**:

storage, ui, tests

**Context**: 

Update the Volume CRD to use a list of `Conditions` instead of the `Phase/ErrorCode/ErrorMessage` triptych

**Summary**:

- update the CRD
- update the operator code with a compat layer
- update the E2E tests with a compat layer
- update the UI code with a compat layer
- add some unit tests for the compat layer

**Acceptance criteria**: 

Everything should work as before, no visible change (except the Volume schema of course)

---

Closes: #1507


storage-operator/deploy/crds/storage_v1alpha1_volume_crd.yaml
```diff
@@ -78,28 +78,42 @@ spec:
...
-            errorMessage:
-              description: Volume failure error message
-              type: string
+            conditions:
+              description: List of conditions through which the Volume has or has
+                not passed.
+              items:
+                properties:
+                  lastTransitionTime:
+                    description: Last time the condition transit from one status to
```
charlesprost:
```suggestion
                    description: Last time the condition transited from one status to
```

storage-operator/deploy/crds/storage_v1alpha1_volume_crd.yaml
```diff
@@ -78,28 +78,42 @@ spec:
...
+                      another (optional)
```
charlesprost:
```suggestion
                      another (optional).
```

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
 
 type VolumePhase string
 
-// "Enum" representing the phase/status of a volume.
+// TODO: kept for temporary compatibility, to be removed.
```
charlesprost, line 66: Temporary? What would condition the end of this temporary time?

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
...
+// "Enum" representing the phase of a volume.
```
charlesprost, line 67: So we still have a phase for a volume but it's not directly in the CRD but it comes from the computation of all conditions?

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
...
 	VolumeFailed      VolumePhase = "Failed"
 	VolumePending     VolumePhase = "Pending"
 	VolumeAvailable   VolumePhase = "Available"
 	VolumeTerminating VolumePhase = "Terminating"
 )
 
-type VolumeErrorCode string
+type ConditionReason string
 
+// TODO: replace those by more fine-grained one.
```
charlesprost:
```suggestion
// TODO: replace those by more fine-grained ones.
```

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
...
+)
+
+type VolumeCondition struct {
+	// Type of volume condition.
+	Type VolumeConditionType `json:"type"`
+	// Status of the condition, one of True, False, Unknown.
+	Status ConditionStatus `json:"status"`
+	// Last time the condition was updated (optional).
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
+	// Last time the condition transit from one status to another (optional)
```
charlesprost:
```suggestion
	// Last time the condition transited from one status to another (optional).
```

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
charlesprost: Oh I've just discovered the `computePhase` later, so I guess that's it.

charlesprost (commented):
Very minor stuff to fix (or not).
Not sure I'm able to review the Go syntax but the logic seems clear (or I'm just tired maybe ^^).
Didn't test it but if our tests are green, that should be fine.
Good job.

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: Yeah, we still have the concept of a global "Phase" because lot of logic relies on that.

Not sure we will keep it when we make the real move to conditions, we will see.
But for now we have a computed one to ease the transition.

storage-operator/deploy/crds/storage_v1alpha1_volume_crd.yaml
slaperche-scality: This file is auto generated, but I'll fix the comment from which this is generated.
Done

storage-operator/deploy/crds/storage_v1alpha1_volume_crd.yaml
slaperche-scality: ditto and done.

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: Done

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: Done

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: When we will have fine-grained conditions representing each steps, the concept of global phase may become irrelevant/useless and thus it could be removed (some k8s objects only have conditions).
I'll try to open a new issue about this, containing the more completed design I first proposed for this one (but which I then simplified due to the time constraint)

charlesprost: OK thx for the explanation.

charlesprost (dismissed):

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
...
+)
+
+type VolumeConditionType string
+
+const (
+	// VolumeReady means Volume is ready to be used.
+	VolumeReady VolumeConditionType = "Ready"
 )
 
+type ConditionStatus string
```
NicolasT: Can't we reuse the type from `k8s.io/pkg/apis/core` (or whatever the right import path is)?

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -63,44 +63,72 @@ type PersistentVolumeTemplateSpec struct {
...
+
+const (
+	ConditionTrue    ConditionStatus = "True"
```
NicolasT: (and then these values as well)

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: Ho yes, you're right, let's use `corev1.ConditionStatus`

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
slaperche-scality: Done

NicolasT (approved):
Somewhat of a cursory review, but LGTM!

PR #1801, opened 2019-09-27
Improvement/1366 implement the tooltip from coreui

### LINKED ISSUE #1366: Implement Tooltips from core-ui
**Component**: ui, component

**Why this is needed**:
Currently, we are using a library of `rc-tooltip`. But it causes problem when we are trying to change the background color of the tooltip. 
We decide to develop the tooltip by ourselves in core-ui instead of using the library. And want to implement our Tooltip metal.

**What should be done**:
Use tooltip from core-ui

ChengYanJin:
**Component**: ui, component

**Context**:  
In order to customize the style of Tooltip, we developed our own tooltip in core-ui,  and we want to replace the rc-tooltip with the one from core-ui. 

**Summary**: 
In order to get the new version of core-ui, we suggest you to do `npm install` before running the code locally.

**Acceptance criteria**: 
The Tooltip from monitoring should be displayed like this.
image

The tooltip in volume list should be displayed like this.
image

Closes: #1366


ui/src/containers/NodeVolumes.js
```diff
@@ -40,8 +39,15 @@ import {
 const VolumeTable = styled.div`
   flex-grow: 1;
   margin-top: ${padding.small};
-  .sc-table-column-cell-container-action {
+  /* .sc-table-column-cell-container-action {
```
slaperche-scality: Zombie code, kill it! :stuck_out_tongue: 

We should avoid keeping commented out code: if it's useless remove it (it's not lost, we have the Git history).

See also https://www.bitnative.com/2012/10/22/kill-the-zombies-in-your-code/

ui/src/containers/NodeVolumes.js
```diff
@@ -4,10 +4,17 @@ import { injectIntl } from 'react-intl';
...
+  Table,
+  Loader,
+  Modal,
+  SearchInput,
+  Tooltip,
+} from '@scality/core-ui';
 import { padding, grayLight } from '@scality/core-ui/dist/style/theme';
 import NoRowsRenderer from '../components/NoRowsRenderer';
-import Tooltip from '../components/Tooltip';
+//import Tooltip from '../components/Tooltip';
```
slaperche-scality: This line should be deleted, not commented.

ui/src/containers/NodeVolumes.js
```diff
@@ -40,8 +39,12 @@ import {
 const VolumeTable = styled.div`
   flex-grow: 1;
   margin-top: ${padding.small};
-  .sc-table-column-cell-container-action {
-    justify-content: center;
+  /*solve the tooltip display issue*/
```
slaperche-scality, line 42: `//` instead of `/* … */` for single line comment, no?

Also, you should describe here or in the commit message what is the issue your worked around.

slaperche-scality (changes requested):

ui/src/containers/NodeVolumes.js
ChengYanJin: /*...*/  is the way to give commend in CSS ^^
But yes, for sure it's good to describe what's the issue. Will add it.
Thanks for the review!

MonPote (approved):

ui/src/containers/NodeVolumes.js
```diff
@@ -33,8 +39,12 @@ import {
...
+  .ReactVirtualized__Grid {
+    overflow: visible !important;
```
carlito-scality, line 44: The use of '!important' should be avoided but I can not find a better solution. We can leave it here

carlito-scality (approved):

slaperche-scality (approved):

ChengYanJin:
@bert-e approve

Issue #1800, opened 2019-09-27
Decide the TLS termination for the Ingress objects


charlesprost:
Relates to #1797 #1798 #1799 

**Component**:

'kubernetes', 'network', 'ingress'

**Why this is needed**:

Because for some components (such as K8S API), we probably want to keep the TLS termination on the component itself (so we use HTTPS while using kubectl).
And for other components (such as the UI), we may want to have HTTP available internally and using HTTPS from the outside.

**What should be done**:

Decide component by component what to do.

**Implementation proposal** (strongly recommended):

**Test plan**:

NicolasT:
Not sure what the question is here...

- For K8s API and SaltAPI, we *need* them to be bound to the control-plane IP of nodes (there's no CNI available when they're deployed, all we have is `hostNetwork`), and as such most definitely require TLS, and the cert chain their clients expect.
- K8s API is accessible to outside consumers (not the UI) on the control-plane network, using TLS, and so is SaltAPI, for any non-UI API interactions with the system (`kubectl`, our CLI,...)

The only 'corner case' is the UI.

Since the UI requires access to K8s API and SaltAPI, and we don't want the user to be forced to accept the certificates they use (which are, and must remain, self-signed), we need them to be accessible through the same channel as the UI. Since these APIs are sensitive, and request include credentials, we need them to be exposed over TLS, but this can be the *same* TLS cert as the one used for the UI (managed by the ingress controller proxy), so there's no issue.

Anyway.

I think the split-up between various tickets that was made isn't right. We don't want to have an *Ingress* for every service separately, because most *are* already accessible from the outside. It's the UI, Grafana and maybe AlertManager we want to expose.

So we need *at most* 3 Ingress objects (in the according namespaces), unless strictly required otherwise.

Every single one of them, I believe, should be using the same host (or can be the only service if we have an ingress controller on the control-plane IP), then do path-based routing between them.

For Grafana and AlertManager, this should be trivial (though we'll likely want to put them behind something like `oidc-proxy` or similar! So we'll really need to tackle the authn topic before exposing any services...). For the UI, it's slightly more complicated because **the UI requires services not in its own namespace to be available as part of its routing structure**.

However, luckily, this is not too much of a concern, thanks to `ExternalName` *Service*s.

So, here's the trick for the UI (though needs cleanup and enhancements):

```yaml
---
kind: Service
apiVersion: v1
metadata:
  name: kubernetes-api
  namespace: metalk8s-ui
spec:
[...]

PR #1802, opened 2019-09-27
Use an ingress controller on the control-plane network for UI and API (through UI) access

NicolasT:
Cfr. https://github.com/scality/metalk8s/issues/1800#issuecomment-536082408

Still a bit of work to be done w.r.t. inter-service TLS traffic, and eventually turning this into a single *Ingress* object, but this should be a good start.

TODO:
- [x] Docs
- [ ] Consider using the host-network for the ingress controller, iff we can bind the resulting `nginx` to only the control-plane IP of a node, to recover source/client IPs (which doesn't work with `externalIPs` in a *Service*)


NicolasT:
This is a bit stuck on the Cypress tests, because we now expose the UI on HTTPS using an IP-address and a self-signed certificate.

I believe this could be fixed by https://github.com/cypress-io/cypress/pull/4947 (cfr. https://github.com/cypress-io/cypress/issues/771), but there's no Cypress release including this change, and I seem to be unable to install an unreleased version for testing :roll_eyes:

NicolasT:
Jay, found a work-around :tada:

docs/quickstart/services.rst
```diff
@@ -64,23 +47,8 @@ This page displays two monitoring indicators:
...
-Forward the port used by Grafana:
-
-.. code-block:: shell
-
-   user@your-host $ kubectl -n metalk8s-monitoring port-forward svc/prometheus-operator-grafana 3000:80
-
-Then open your web browser and navigate to ``http://localhost:3000``
-
+Grafana is available on the same host as the MetalK8s UI, under ``/grafana``.
+Log in with the default credentials: ``grafana`` / ``prom-operator``.
```
sayf-eddine-scality: Still having issue to login the default creds are not working

charts/nginx-ingress-control-plane.yaml
```diff
@@ -0,0 +1,42 @@
+controller:
+  image:
+    repository: '{%- endraw -%}{{ build_image_name("nginx-ingress-controller", False) }}{%- raw -%}'
+
+  defaultBackendService: 'metalk8s-ingress/nginx-ingress-default-backend'
```
sayf-eddine-scality, line 5: we may need a good 404 page with the same theme as the UI

sayf-eddine-scality (changes requested):
two minor comments the rest is LGTM

charts/nginx-ingress-control-plane.yaml
gdemonet: Indeed, could even redirect to MetalK8s UI 404 page.

NicolasT: Good idea, but I think that's out of scope for this PR.

docs/quickstart/services.rst
NicolasT: That's because I made a mistake in the docs ;-P Will fix.

NicolasT: I changed the default one to `admin`/`admin`. Should get OIDC up soon!

sayf-eddine-scality (approved):

Issue #1836, opened 2019-10-04
Upgrade may fail if apiserver is not ready


charlesprost:
**Component**:

'containers', 'tests', 'ci'

**What happened**:

While running the 2.4.0 upgrade step on a single-node-upgrade-test stage, after having deployed a 2.3.0 single node, we face this error:
```
          ID: Set node bootstrap version to 2.4.0-dev
    Function: metalk8s_kubernetes.node_label_present
        Name: metalk8s.scality.com/version
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1919, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1918, in wrapper
                  return f(*args, **kwargs)
                File "/var/cache/salt/master/extmods/states/metalk8s_kubernetes.py", line 1053, in node_label_present
                  labels = __salt__['metalk8s_kubernetes.node_labels'](node, **kwargs)
                File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 350, in node_labels
                  match = node(name, **kwargs)
                File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 331, in node
                  raise CommandExecutionError(exc)
              CommandExecutionError: HTTPSConnectionPool(host='10.100.2.172', port=6443): Max retries exceeded with url: /api/v1/nodes (Caused by ProtocolError('Connection aborted.', error(111, 'Connection refused')))
     Started: 09:49:12.805768
    Duration: 33.734 ms
     Changes:
```
So in 33ms, we try to talk to the apiserver, and if it's not available, we fail abruptly.

See build here: https://eve.devsca.com/github/scality/metalk8s/#/builders/25/builds/297/steps/17/logs/stdio
See sosreport here: [sosreport-bootstrap-2019-10-01-dbibmzl.zip](https://github.com/scality/metalk8s/files/3689767/sosreport-bootstrap-2019-10-01-dbibmzl.zip)

**What was expected**:

The upgrade should have been running OK.

**Steps to reproduce**

Not sure because it's flaky.

**Resolution proposal** (optional):

Let's wait for apiserver to be up and running before continuing with the upgrade process.

PR #1837, opened 2019-10-04
upgrade: wait for API server before continuing

charlesprost:
**Component**:

'containers', 'tests', 'ci'

**Context**: 

If we don't wait for apiserver to be up and ready during the upgrade process, then apiserver may be still booting, then any request to it will fail hence the whole upgrade process will fail.

Thx @TeddyAndrieux for helping on this one.

**Acceptance criteria**: 

Upgrade process should not fail because apiserver is not available.

---

Closes: #1836

### DIFF
--- a/salt/metalk8s/orchestrate/upgrade/init.sls
+++ b/salt/metalk8s/orchestrate/upgrade/init.sls
@@ -41,6 +41,18 @@ Skip node {{ node }}, already in {{ node_version }} newer than {{ dest_version }
 
   {%- else %}
 
+Wait for API server to be available on {{ node }}:
+  http.wait_for_successful_query:
+  - name: https://{{ pillar.metalk8s.api_server.host }}:6443/healthz
+  - match: 'ok'
+  - status: 200
+  - verify_ssl: false
+  - require:
+    - salt: Upgrade etcd cluster
+  {%- if previous_node is defined %}
+    - salt: Deploy node {{ previous_node }}
+  {%- endif %}
+
 Set node {{ node }} version to {{ dest_version }}:
   metalk8s_kubernetes.node_label_present:
     - name: metalk8s.scality.com/version
@@ -49,10 +61,7 @@ Set node {{ node }} version to {{ dest_version }}:
     - kubeconfig: {{ kubeconfig }}
     - context: {{ context }}
     - require:
-      - salt: Upgrade etcd cluster
-    {%- if previous_node is defined %}
-      - salt: Deploy node {{ previous_node }}
-    {%- endif %}
+      - http: Wait for API server to be available on {{ node }}
 
 Deploy node {{ node }}:
   salt.runner:


salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -41,6 +41,13 @@ Skip node {{ node }}, already in {{ node_version }} newer than {{ dest_version }
 
   {%- else %}
 
+Wait for API server to be available:
+  http.wait_for_successful_query:
+  - name: https://{{ pillar.metalk8s.api_server.host }}:6443/healthz
+  - match: 'ok'
+  - status: 200
+  - verify_ssl: false
```
TeddyAndrieux, line 49: I think you should move the requires here
```
      - salt: Upgrade etcd cluster
    {%- if previous_node is defined %}
      - salt: Deploy node {{ previous_node }}
    {%- endif %}
```

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -41,6 +41,13 @@ Skip node {{ node }}, already in {{ node_version }} newer than {{ dest_version }
 
   {%- else %}
 
+Wait for API server to be available:
```
TeddyAndrieux: That will not work as you are in a `for` two state will have the same name you need to add the node name here.
Maybe `Wait for API server to be available for {{ node }}`

TeddyAndrieux (commented):
IMO same thing should be done in the downgrade orchestrate.

And that should target development/2.0 no ?

salt/metalk8s/orchestrate/upgrade/init.sls
NicolasT: `on`, not `for`.

TeddyAndrieux: Hmm it's not really `on` because we not check API server on this node but we check that API server available to be able to upgrade the node

TeddyAndrieux: Oh but yes we wait **on** this node so `on` maybe :stuck_out_tongue:

salt/metalk8s/orchestrate/upgrade/init.sls
charlesprost: Done.

charlesprost:
> IMO same thing should be done in the downgrade orchestrate.

I'll create a ticket for this.

> 
> And that should target development/2.0 no ?

No because in 2.0 we don't have the upgrade test (since we don't test from a version below 2.0).

TeddyAndrieux:
> No because in 2.0 we don't have the upgrade test (since we don't test from a version below 2.0).

Yes but your change are in the upgrade orchestrate so part of 2.0. And if one day we go for a 2.0.1 we will include upgrade test also in 2.0. (I don't know if we will do this one day but for me we should also do this in the 2.0 branch)

charlesprost:
> > No because in 2.0 we don't have the upgrade test (since we don't test from a version below 2.0).
> 
> Yes but your change are in the upgrade orchestrate so part of 2.0. And if one day we go for a 2.0.1 we will include upgrade test also in 2.0. (I don't know if we will do this one day but for me we should also do this in the 2.0 branch)

You're right, I've rebased and changed the target branch.

TeddyAndrieux (dismissed):
LGTM

thomasdanan (approved):

PR #1846, opened 2019-10-04
Build documentation and storage-operator inside containers

### LINKED ISSUE #839: Improve tasks that save container images

### LINKED ISSUE #1830: Build the documentation inside a container
**Component**:

buildchain

**Why this is needed**:

We are now shipping the documentation as a PDF in the ISO, consequently we now build the PDF during the build process and this requires some new dependencies (such as a LaTeX distribution, which can be quite big…).

Since people may simply want generate an ISO and have nothing to do with the doc itself, it's annoying to require them to install those new dependencies.

To circumvent this issue, we can instead use a container that contains the dependencies to build the documentation.

**What should be done**:

Use a container to build the documentation (we already use one in the CI, try to reuse it)

- add a new builder image in the buildchain
- use the builder image for the `doc:*` targets

**Test plan**:

Try to build an ISO without LaTeX installed on your machine.

### LINKED ISSUE #1831: Build the storage-operator inside a container
**Component**:

buildchain

**Why this is needed**:

Similar reasons as #1830.

To build an ISO you need to install Go, the Operator SDK and download all the required deps (which can be quite large).

Since you may simply want to build and ISO and not hack on Go code, it would be nice to be able to build an ISO without having to install all those deps.

Also, using a builder image we can fix the Go version used to build the operator which will avoid discrepencies in how Go react to fix depending on its version (such as the fix we made to solve the `Thrift Apache repository` problem, works with Go 1.12, but not with Go 1.13…).

**What should be done**:

Use a container to build the `storage-operator` image.

- add a new builder for Go code
- use it to build the `storage-operator` image

**Test plan**:

Build an ISO without Go installed on your machine

slaperche-scality:
**Component**:

buildchain, ci

**Context**: 

Some people just want to build a MetalK8s ISO and be done with it, thus requiring them to install developer dependencies such as Go toolchain and LaTeX distribution is perfectible™

**Summary**:

- rework the handling of "build container" (a.k.a. builder) in the buildchain
- rework doc-building tasks to use a builder
- rework go-buidling tasks to use a builder
- revert some changes that were made in the CI to build Go code and PDF doc

Note that we only move build-related task inside a container, things such as live doc rendering and Go codegen/linting still requires you to have dev dependencies (which is acceptable).

**Acceptance criteria**: 

You can build an ISO without Go, Operator SDK or LaTeX installed on your machine.

---

Closes: #839
Closes: #1830
Closes: #1831


jbertran (approved):
Damn, nice work. You tackled everything I wanted to tackle in my ancient docker PR, I'll remove that one.

slaperche-scality:
> Damn, nice work. You tackled everything I wanted to tackle in my ancient docker PR, I'll remove that one.

Thanks :slightly_smiling_face: 

Everything but the `rm` from the local registry!
It wasn't really related to this PR (whereas removing the `touch`ed file was a nice addition since I'll reworking this part anyway) so I left it aside.

You may want to reopen a PR for that part (seems like the weird CI you had disappeared, maybe related to the `dind` update…)

PR #1853, opened 2019-10-07
Generate a build log

### LINKED ISSUE #1849: Generate a build log
**Component**:

build, ci

**Why this is needed**:

If we want to track flakiness and or build time increase, it would really help to have build log that contains, for each task, status, timing information, generated errors, etc.

**What should be done**:

Update the buildchain to generate a build log, in JSON (for easier post-processing/consumption), at the end of the build process.

**Implementation proposal**:

Use a custom reporter (see [here](https://pydoit.org/cmd_run.html#custom-reporter)), likely derived from the existing `JsonReporter`.

**Test plan**:

None

slaperche-scality:
**Component**:

buildchain, ci

**Context**: 

Generate a build log.

**Summary**:

- update our custom reporter to generate a JSON build log at the end of the build
- update the CI to upload the build log to artifacts.

**Acceptance criteria**: 

a `build.log` file is uploaded to artifacts and it contains build information as JSON.

---

Closes: #1849

### DIFF
--- a/buildchain/dodo.py
+++ b/buildchain/dodo.py
@@ -5,10 +5,12 @@
 
 """Build entry point."""
 
+from typing import Any, Dict, List, TextIO
 import sys
 
 import doit  # type: ignore
 
+from buildchain import constants
 from buildchain.build import *
 from buildchain.builder import *
 from buildchain.codegen import *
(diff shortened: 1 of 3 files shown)


eve/main.yml
```diff
@@ -70,18 +70,22 @@ stages:
           usePTY: true
           haltOnFailure: true
       - ShellCommand:
-          name: Put the iso file in a separate folder
+          name: Put the artifacts to upload in a separate folder
           command: >
-            mkdir iso &&
-            cp _build/metalk8s.iso _build/SHA256SUM _build/root/product.txt iso
+            mkdir artifacts &&
+            cp build.log artifacts &&
```
jbertran, line 76: A loop wouldn't be out of place here

buildchain/dodo.py
```diff
@@ -23,29 +25,92 @@
...
+
+    def complete_run(self) -> None:
+        """Called when finished running all tasks."""
+        # If test fails print output from failed task.
+        for result in self.failures:
+            task = result['task']
+            # Makes no sense to print output if task was not executed.
+            if not task.executed:
+                continue
+            self._write('#'*40 + '\n')
```
jbertran: Could make this a variable - `failure_sep`?

buildchain/dodo.py
```diff
@@ -23,29 +25,92 @@
...
+            out = ''.join([action.out for action in task.actions if action.out])
+            self._write('{} <stdout>:\n{}\n'.format(task.name, out))
+
+        if self.runtime_errors:
+            self._write('#'*40 + '\n')
+            self._write('Execution aborted.\n')
+            self._write('\n'.join(self.runtime_errors))
+            self._write('\n')
+        # Generate the build log.
+        build_log = constants.ROOT/'build.log'
```
jbertran, line 96: This variable might be better suited higher in the hierarchy - or even in `constants` - as it's a fairly important one (also, a constant would be cool for @NicolasT 's eve-in-python PoC)

buildchain/dodo.py
```diff
@@ -23,29 +25,92 @@
...
+        super().__init__(outstream, options)
+        self.failures : List[Dict[str, Any]] = []
+        self.runtime_errors : List[str] = []
 
     def execute_task(self, task: doit.task.Task) -> None:
         """Called when a task is executed."""
-        # Ignore tasks that do not define actions.
-        if task.actions:
-            self.write('.  {}\n'.format(task.title()))
+        super().execute_task(task)
```
jbertran, line 40: I'm sure I'm missing something here, but why do we call `super().everything` when we previously just printed?

jbertran (commented):
Nice work, this is a great addition for build visibility :)

buildchain/dodo.py
slaperche-scality: It's only used locally, so not sure it make sense to expose it in a wider scope.

Can be moved when the need arises.

buildchain/dodo.py
slaperche-scality: The `super` calls the `JsonReporter` methods, which populate an list or dict to keep track of the status of each tasks, the elapsed time, …
This is then used to compute the final JSON at the end (see https://github.com/pydoit/doit/blob/master/doit/reporter.py#L191).

eve/main.yml
slaperche-scality: Yeah but it's YAML stuff, I prefer keep things simple here :p

Wasted too many time with the weird YAML syntax rules, the line continuations, …

Let's hope we will get rid of this unreadable mess this has become since the addition of upgrade/downgrade, and replace it by some code that generate it :D

buildchain/dodo.py
jbertran: Ah right whereas we just hooked into the calls for console since we just needed printing. Thanks

buildchain/dodo.py
jbertran: Moving it to class variable would be better for visibility imo, in case anyone should be looking for it.

Not a big issue though.

buildchain/dodo.py
slaperche-scality: Done!

jbertran (approved):

PR #1871, opened 2019-10-08
UI: Add grafana link in the navbar

carlito-scality:
**Component**:UI

**Context**: 
- Once logged in metalk8s UI (or any other solution admin UI actually), I would like to quickly access grafana dashboard in case I need to do deeper investigation.

**Summary**:

**Acceptance criteria**: 
- Build and launch the UI 
- Click on the "Advanced Monitoring" button in the Cluster Health Page to access Grafana dashboard in a new tab.

image

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/public/config.json
+++ b/ui/public/config.json
@@ -1,5 +1,6 @@
 {
-    "url": "https://172.21.254.13:6443",
-    "url_salt": "http://172.21.254.13:4507",
-    "url_prometheus": "http://172.21.254.46:31266"
+    "url": "https://{ip}:8443/api/kubernetes",
+    "url_salt": "https://{ip}:8443/api/salt",
+    "url_prometheus": "https://{ip}:8443/api/prometheus",
+    "url_grafana": "https://{ip}:8443/grafana"
 }
--- a/salt/metalk8s/addons/ui/deployed/ui.sls
+++ b/salt/metalk8s/addons/ui/deployed/ui.sls
@@ -43,7 +43,8 @@ Create metalk8s-ui ConfigMap:
           {
             "url": "/api/kubernetes",
             "url_salt": "/api/salt",
-            "url_prometheus": "/api/prometheus"
+            "url_prometheus": "/api/prometheus",
+            "url_grafana": "/grafana"
           }
 
 Create ui-branding ConfigMap:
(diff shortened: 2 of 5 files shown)


ui/src/containers/Layout.js
```diff
@@ -38,6 +38,12 @@ const Layout = props => {
   }, [fetchClusterVersion]);
 
   const help = [
+    {
+      label: props.intl.messages.advanced_monitoring,
+      onClick: () => {
+        props.history.push('/grafana');
```
NicolasT: May want to have this in the 'config' document as well, where we have `/api/kubernetes` etc. Otherwise, why have the config at all, and not hard-code the other paths as well?

carlito-scality: Yes good point!
And we can mock the `config.js` for the development mode.

thomasdanan:
I would prefer the link to be always visible (vs within a menu) and have an associated icon with it (ex: https://www.plesk.com/extensions/monitoring/)

ui/public/config.json
```diff
@@ -1,5 +1,6 @@
 {
-    "url": "https://172.21.254.13:6443",
-    "url_salt": "http://172.21.254.13:4507",
-    "url_prometheus": "http://172.21.254.46:31266"
+    "url": "https://{ip}:8443/api/kubernetes",
+    "url_salt": "https://{ip}:8443/api/salt",
+    "url_prometheus": "https://{ip}:8443/api/prometheus",
+    "url_grafana": "https://{ip}:8443/grafana"
```
gdemonet, line 5: I think those paths are intended to be exposed (via Ingress) on the same host as the UI itself, so maybe using:

```
{
  "url": "/api/kubernetes",
  "url_salt": "/api/salt",
  "url_prometheus": "/api/prometheus",
  "url_grafana": "/grafana",
}
```

is acceptable, even locally. No?

gdemonet: If the goal is to pass some `INGRESS_HOST` when running the UI from your laptop, maybe that's something to add in some `npm` script when running a dev server?

gdemonet (approved):
Not tested but screenshot looks good :) small remark on local `config.json`

ui/public/config.json
carlito-scality: The goal is to be able to hook the local development UI (accessible at localhost:3000) to any backend (cluster) which serves as a mock. This local development UI is not part of the cluster (so Ingress). We can discuss it IRL ;)

Issue #1861, opened 2019-10-07
Expose component statuses with example Solution Operator


gdemonet:
**Component**: solutions

**Why this is needed**: Currently, the example Solution does not expose any status, defeating the purpose of operating this Solution through CustomResources.

**What should be done**: Bubble-up status of Deployments to CRs with the Solution Operator, expose them from the Solution UI.

**Test plan**: Check CR status from CLI and UI tests

Issue #1860, opened 2019-10-07
Deploy Solution Grafana dashboards with Salt


gdemonet:
**Component**: salt, solutions

**Why this is needed**: Grafana dashboards are passed to the Grafana instance running in the cluster through `ConfigMap`s. We cannot ask from Solutions to perform this configuration themselves, since implementation may evolve.

**What should be done**: Deploy Grafana dashboards at the same time as the Admin UI and CRDs.

**Test plan**: Check that dashboards defined in #1859 are well deployed.

Issue #1859, opened 2019-10-07
Create Grafana dashboards for example Solution


gdemonet:
**Component**: monitoring, solutions

**Why this is needed**: To be able to test our tooling responsible for the deployment of said dashboards.

**What should be done**: Implement a dashboard displaying some charts for all components in a given Namespace. Requires metrics from #1857

Issue #1857, opened 2019-10-07
Expose Prometheus metrics from example Solution


gdemonet:
**Component**: solution

**Why this is needed**: To validate integration with MetalK8s Prometheus stack.

**What should be done**:

- [ ] Implement management of `ServiceMonitor`s and/or appropriate labels on `Service` objects for Solution components and Operator
- [ ] Ensure documentation from #1158 is in line with the implementation

**Test plan**: Check that metrics are available for the Solution Operator after tests from #1850

NicolasT:
To be clear, setting up *ServiceMonitor*s is part of the deployment of a solution, i.e., something up to the operator to do. Unless this is primarily about monitoring the operator?

gdemonet:
It's about both, the goal is to implement this for the example Solution we embed in the repo.

PR #1876, opened 2019-10-08
Adapt Solutions integration

gdemonet:

### Context

Solution Admin UIs, as well as MetalK8s UI, recently (see #1810 and #1783) adopted a concept of `Environment` (actual CR for now, though implementation may change in the future) to manage isolated deployments of Solutions, grouped into functional units.
To provide a more meaningful example Solution with respect to the actual integration tooling provided by MetalK8s, we also extend the example applications provided (add a "clock" server using the same base HTTP server, create `Service` objects).
Future improvements will include exposing `status` information (see #1861) and adding monitoring of the deployed components (see #1857, #1859, #1860).

### Summary

- [salt] Deploy `Environment` CRD with `metalk8s.addons.solutions.deployed`
- [salt] Use `metalk8s-registry-from-config.invalid` as registry prefix for deployed Solutions
- [example] Replace `Example` CR with two new: `VersionServer` and `ClockServer`
- [example] Expose `Service` objects for each component

#### Ability to deploy the example Solution

- Build the example Solution ISO:

```
cd examples/metalk8s-solution-example
make iso
```

- On the Bootstrap (for Vagrant), register the built ISO:

```
sudo /srv/scality/metalk8s-2.4.0-dev/solution-manager.sh \
  --add /vagrant/examples/metalk8s-solution-example/_build/example-solution-0.1.0-dev.iso
```

- From MetalK8s UI (`<control-plane-ip>/solutions`), create an environment, and add `Example Solution 0.1.0-dev` to it.

- This should redirect you to `localhost:<port>/environments/<env_name>/version/0.1.0-dev/prepare` (make sure URLs are the same).
    - To actually go to the Solution Admin UI, replace `localhost` with the control-plane IP
    - This will lead you to a login page, then to the main page listing Environments
    - _Note: the redirect does not know about the host, but this issue will be void once #1855 is done)_

- Once you have logged in to the Solution UI, you want to "prepare" the Environment for deployment of the Solution Application (which actually creates a Namespace and a Deployment for the Solution Operator in this namespace)
    - Either re-use the URL from previous step `.../version/0.1.0-dev/prepare`
    - Or simply click the "Prepare" button on the line for your Environment

```
# If your Environment was named "testenv", Namespace should be "<env_name>-<sol_name>"
$ kubectl get deploy -n testenv-example-solution
NAME               DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
example-operator   1         0         0            0           1s
```

- This will lead you to the view of the Environment from the Solution UI. You can now deploy components:
    - Adding a `VersionServer` will create a Deployment (with the same `.spec.replicas` as the original CR) and corresponding Service
    - Adding a `ClockServer` will do roughly the same, although the Deployment will only request a single replica

#### Ability to upgrade the example Solution

- To upgrade, you'll need a new version of the Solution ISO. One can do this by bumping the values in `examples/metalk8s-solution-example/VERSION`, and retrigger a `make iso`.

- To make a new version available to the cluster, run the same `solution-manager.sh --add <iso_path>` command, using the new ISO.

- Once that's done, it should be visible in MetalK8s UI `/solutions` page (and this page should also display the `example-solution-0.1.0-dev` version as being part of the previously created Environment).

- From the Solution UI (refer to steps above for how to access it, clicking the button in the Environment will yield the same `localhost` situation), go into the Environment where `0.1.0-dev` is active.

- Edit the "version" field in the "Information" pane: this will trigger an upgrade of this Solution Operator for this Environment (namespace being `<env_name>-<sol_name>`). You can check this version with the following command:

```
$ kubectl get deploy -n testenv-example-solution example-operator -o jsonpath='{.spec.containers[*].image}'
metalk8s-registry-from-config.invalid/example-solution-<new_version>/example-operator:<new_version>
```

- Now that the images are available to the cluster and the Operator is up-to-date, you can edit each component to set its `.spec.version` to the new one. Upgrade is done (check the Deployments or Pods to make sure of it).


alexandre-allard:
LGTM, just few things, not necessarily related to this PR:
  - To acces to the solution UI, you must use the `workload_plane_ip` instead of the `control_plane_ip`
  - In order to be able to login on the solution UI, we must first export the kube api server certificate
  - The `Solutions > Create a New Environment > Name` field allows spaces (and probably some other forbidden characters) and then fails silently when submitting the form (error 422 in the console)
  - There is an issue with the script managing /etc/metalk8s/solutions.yaml, at some point I ended up with the same solutions defined multiple times, which led to Salt conflicting ids when rendering formulas

gdemonet:
> LGTM, just few things, not necessarily related to this PR:
> 
> * To acces to the solution UI, you must use the `workload_plane_ip` instead of the `control_plane_ip`

Indeed, that's expected, the provided steps are not up-to-date w.r.t #1802 

> * In order to be able to login on the solution UI, we must first export the kube api server certificate

Expected, will change once we implement #1855 

> * The `Solutions > Create a New Environment > Name` field allows spaces (and probably some other forbidden characters) and then fails silently when submitting the form (error 422 in the console)

Not expected, need to open a tracking issue

> * There is an issue with the script managing /etc/metalk8s/solutions.yaml, at some point I ended up with the same solutions defined multiple times, which led to Salt conflicting ids when rendering formulas

Expected, this script is being rewritten in #1854

alexandre-allard (dismissed):

alexandre-allard (approved):

Issue #1872, opened 2019-10-08
Building the documentation may fail


charlesprost:
Follows: #1830 

**Component**:

'build', 'ci', 'documentation'

**What happened**:

When building the ISO, and specifically the documentation part, we can hit this error:

```
.  BUILD RPM REPO _build/root/packages/redhat/metalk8s-saltstack-el7/repodata/repomd.xml
.  BUILD RPM REPO _build/root/packages/redhat/metalk8s-extras-el7/repodata/repomd.xml
.  BUILD RPM REPO _build/root/packages/redhat/metalk8s-updates-el7/repodata/repomd.xml
.  BUILD RPM REPO _build/root/packages/redhat/metalk8s-epel-el7/repodata/repomd.xml
.  BUILD SRPM     _build/packages/redhat/metalk8s-sosreport-2.4-1.el7.src.rpm
TaskError - taskid:doc:pdf
Command '['/entrypoint.sh', 'latexpdf']' in image 'metalk8s-doc-builder:latest' returned non-zero exit status 1: b'':
b''
program finished with exit code 2
elapsedTime=793.921439
```

See this build: https://eve.devsca.com/github/scality/metalk8s/#/builders/1/builds/12658

**What was expected**:

The documentation should have been built smoothly.

**Steps to reproduce**

Not sure since it's a flaky. But it seems it happens a lot on 2.4 integration branches.

**Resolution proposal** (optional):

Not sure.

slaperche-scality:
Now that we have proper output, here is the error I got (see https://eve.devsca.com/github/scality/metalk8s/#/builders/1/builds/13016):

```
docs run-test: commands[1] | /usr/src/metalk8s/docs/build.cmd latexpdf
make: Entering directory '/usr/src/metalk8s/docs'
Running Sphinx v2.0.1
Initializing Spelling Checker
Initializing GitHub plugin         
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [latex]: all documents
updating environment: 45 added, 0 changed, 0 removed
reading sources... [ 20%] developer/architecture/deployment .. developer/building/index
reading sources... [ 40%] developer/building/requirements .. developer/solutions/deploy
reading sources... [ 60%] developer/solutions/index .. operation/bootstrap_backup_restore
reading sources... [ 80%] operation/changing_node_hostname .. quickstart/bootstrap
reading sources... [100%] quickstart/expansion .. quickstart/setup
                                                                    
waiting for workers...
[...]

PR #1881, opened 2019-10-09
Improve output in case of error during the doc building

slaperche-scality:
**Component**:

build

**Context**: 

The doc-builder sometimes fail, but we won't know why (no error message are reported).

**Summary**:

Work around the weird behavior of `tox` (or what it calls) of printing error on `stdout`…

**Acceptance criteria**: 

- edit the `docs/entrypoint.sh` and replace `"$@"` by `xylospongium`
- run `./doit.sh doc:pdf`

You should see the `tox` output with the error message.

---

See: #1872

### DIFF
--- a/docs/entrypoint.sh
+++ b/docs/entrypoint.sh
@@ -4,5 +4,5 @@ set -e
 set -u
 set -o pipefail
 
-tox --workdir /tmp/tox -e docs -- "$@"
+tox --workdir /tmp/tox -e docs -- "$@" 1>&2
 chown -R "${TARGET_UID}:${TARGET_GID}" /usr/src/metalk8s/docs/_build


charlesprost (approved):

charlesprost:
@slaperche-scality you don't want to target 2.0 for this one? I guess it only fails in 2.4 so that's fine, but i wanted to ask anyway.

slaperche-scality:
> @slaperche-scality you don't want to target 2.0 for this one? I guess it only fails in 2.4 so that's fine, but i wanted to ask anyway.

Haha this time no! (I was careful this time :stuck_out_tongue: )
The "build-doc-in-docker" was introduced in 2.4 (because before no PDF in the ISO, so no need to build the doc during the build process).

Issue #1835, opened 2019-10-04
Running the bootstrap twice may fail because of some timeout on salt-master


charlesprost:
**Component**:

'containers', 'ci', 'tests'

**What happened**:

This is a flaky test that occurs very often in those days.

For example this build: https://eve.devsca.com/github/scality/metalk8s/#/builders/14/builds/5497

sosreport: [sosreport-bootstrap-2019-10-03-yivvutl.zip](https://github.com/scality/metalk8s/files/3689632/sosreport-bootstrap-2019-10-03-yivvutl.zip)

Error looks like when running the bootstrap a second time in the tests, either on single-node or multiple-nodes stage:
```
E               Name: saltutil.sync_all - Function: salt.runner - Result: Changed Started: - 18:02:56.408077 Duration: 3198.467 ms
E               Name: saltutil.sync_roster - Function: salt.runner - Result: Changed Started: - 18:02:59.606782 Duration: 3478.215 ms
E               Name: metalk8s_saltutil.sync_auth - Function: salt.runner - Result: Changed Started: - 18:03:03.085247 Duration: 3310.672 ms
E               Name: saltutil.sync_all - Function: salt.function - Result: Changed Started: - 18:03:06.396165 Duration: 827.052 ms
E               Name: Deploy CA role on bootstrap minion - Function: salt.state - Result: Changed Started: - 18:03:07.224550 Duration: 14151.445 ms
E               Name: Bring bootstrap minion to highstate - Function: salt.state - Result: Changed Started: - 18:03:21.377617 Duration: 29886.327 ms
E             ----------
E                       ID: Generate etcd client certs for salt master
E                 Function: salt.state
E                   Result: False
E                  Comment: An exception occurred in this state: Traceback (most recent call last):
E                             File "/usr/lib/python2.7/site-packages/salt/state.py", line 1919, in call
E                               **cdata['kwargs'])
E                             File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1918, in wrapper
E                               return f(*args, **kwargs)
E                             File "/usr/lib/python2.7/site-packages/salt/states/saltmod.py", line 335, in state
E                               cmd_ret = __salt__['saltutil.cmd'](tgt, fun, **cmd_kw)
E                             File "/usr/lib/python2.7/site-packages/salt/modules/saltutil.py", line 1400, in cmd
E                               client, tgt, fun, arg, timeout, tgt_type, ret, kwarg, **kwargs)
[...]

PR #1883, opened 2019-10-09
salt: Wait API server after bootstrap highstate

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Timeout in ext_pillar when retrieving information from API server

**Summary**:

Wait for API Server to be available right after the bootstrap highstate

---

Fixes: #1835


slaperche-scality (approved):

gdemonet (approved):
LGTM for now, would be nice to have such a method defined in our Salt modules interacting with K8s API (maybe I'll add it to #1713)

Issue #1841, opened 2019-10-04
Building some RPM packages may fail without error displaying


charlesprost:
**Component**:

'build'

**What happened**:

In this build https://eve.devsca.com/github/scality/metalk8s/#/builders/23/builds/1051/steps/7/logs/stdio we have this error:
```
.  GET_SRC      _build/packages/redhat/calico-cni-plugin/SOURCES
.  GET_SRC      _build/packages/redhat/metalk8s-sosreport/SOURCES
.  BUILD SRPM   _build/packages/redhat/metalk8s-sosreport-2.2-1.el7.src.rpm
.  HARDLINK     _build/root/images
.  BUILD SRPM   _build/packages/redhat/calico-cni-plugin-3.8.2-1.el7.src.rpm
.  NGINX_CFG    _build/root/salt/metalk8s/repo/files/metalk8s-registry.inc
.  BUILD RPM    _build/root/packages/redhat/metalk8s-scality-el7/x86_64/metalk8s-sosreport-2.2-1.el7.x86_64.rpm
.  BUILD RPM    _build/root/packages/redhat/metalk8s-scality-el7/x86_64/calico-cni-plugin-3.8.2-1.el7.x86_64.rpm
TaskError - taskid:_build_repositories:scality:build_rpm:calico-cni-plugin
Command '['/entrypoint.sh', 'buildrpm']' in image 'metalk8s-build:latest' returned non-zero exit status 1: b'+ yum-builddep -y /rpmbuild/SRPMS/calico-cni-plugin-3.8.2-1.el7.src.rpm\n':
b'+ yum-builddep -y /rpmbuild/SRPMS/calico-cni-plugin-3.8.2-1.el7.src.rpm\n'
program finished with exit code 2
elapsedTime=210.425599
```

There is no explicit error.

See another one here on https://eve.devsca.com/github/scality/metalk8s/#/builders/1/builds/12376 :
```
.  BUILD RPM    _build/root/packages/redhat/metalk8s-scality-el7/x86_64/metalk8s-sosreport-2.2-1.el7.x86_64.rpm
.  NGINX_CFG    _build/root/salt/metalk8s/repo/files/metalk8s-registry.inc
TaskError - taskid:_build_repositories:scality:build_rpm:metalk8s-sosreport
[...]

PR #1884, opened 2019-10-09
build: Redirect `yum-builddep` output to stderr

TeddyAndrieux:
**Component**:

**Context**: 

No information during build when error in `yum-builddep` step

**Summary**:

By default `yum-builddep` only output on stdout, in doit we only print
stderr, lets always redirect `yum-builddep` output to stderr

**Acceptance criteria**: 

Have more information when doit fail in builddep step

---

See: #1841

### DIFF
--- a/packages/redhat/entrypoint.sh
+++ b/packages/redhat/entrypoint.sh
@@ -36,7 +36,7 @@ buildsrpm() {
 
 buildrpm() {
     set -x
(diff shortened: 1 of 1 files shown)


slaperche-scality (dismissed):
Ha, so that's not only `tox`…

sayf-eddine-scality (approved):

Issue #1840, opened 2019-10-04
Multiple nodes stage tests can fail if some containers are not up and ready yet


charlesprost:
**Component**:

'tests', 'ci'

**What happened**:

When running the tests on the multiple nodes stage, after the bootstrap node has been provisioned, we can see this error in this build: https://eve.devsca.com/github/scality/metalk8s/#/builders/15/builds/5368/steps/21/logs/stdio
```
=================================== FAILURES ===================================
_______________________ test_logs[paramiko://bootstrap] ________________________
host = <testinfra.host.Host object at 0x7f6386faf710>
request = <FixtureRequest for <Function test_logs[paramiko://bootstrap]>>
              'check logs from all containers')
>   def test_logs(host):
tests/post/steps/test_logs.py:8: 

[...]

        if not 200 <= r.status <= 299:
>           raise ApiException(http_resp=r)
E           kubernetes.client.rest.ApiException: (400)
E           Reason: Bad Request
E           HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Fri, 27 Sep 2019 15:13:48 GMT', 'Content-Length': '234'})
E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"container \"nginx-ingress-controller\" in pod \"nginx-ingress-controller-f2x9n\" is waiting to start: ContainerCreating","reason":"BadRequest","code":400}
```

It seems we retrieve the logs from the containers but we don't check that they are ready for this.

Having the same kind of error in the build https://eve.devsca.com/github/scality/metalk8s/#/builders/15/builds/5481:
```
>               assert logs.strip(), (
                    "Couldn't find logs for container '{}' in Pod '{}' (status {})"
                ).format(container.name, pod_name, pod.status.phase)
E               AssertionError: Couldn't find logs for container 'kube-proxy' in Pod 'kube-proxy-95n2f' (status Pending)
E               assert ''
E                +  where '' = <built-in method strip of str object at 0x7f81a2980ab0>()
E                +    where <built-in method strip of str object at 0x7f81a2980ab0> = ''.strip
```

That may also happen for other containers as well.

**What was expected**:

The container should have been running so retrieving the logs should have worked.

**Steps to reproduce**

Not sure since it's flaky.

**Resolution proposal** (optional):

Check that container is in the proper state before retrieving its logs.

PR #1889, opened 2019-10-10
tests/containers: Add retry when retrieving pod logs

Ebaneck:
**Component**:

'ci', 'flakiness'

**Context**: 

See #1840 

**Summary**:

- Ensure pods are in running state before reading pod logs by adding a wait/retry when pods are not ready.

**Acceptance criteria**: 

- Local test should be passing.

---

Closes: #1840

### DIFF
--- a/tests/post/steps/test_logs.py
+++ b/tests/post/steps/test_logs.py
@@ -2,6 +2,8 @@ import re
 
 from pytest_bdd import scenario, then
 
+from tests import kube_utils, utils
+
 # Scenarios
 @scenario('../features/log_accessible.feature',
           'check logs from all containers')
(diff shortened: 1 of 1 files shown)


tests/post/steps/test_logs.py
```diff
@@ -2,6 +2,9 @@
 
 from pytest_bdd import scenario, then
 
+from tests import kube_utils
+from tests import utils
```
TeddyAndrieux:
```suggestion
from tests import kube_utils, utils
```
Please use only one line here

tests/post/steps/test_logs.py
```diff
@@ -19,10 +22,14 @@ def test_logs(host):
     'nginx-ingress-default-backend',
 ))
 
+# Constants
+POD_STATUS = 'Running'
```
TeddyAndrieux: I think you don't need a global variable for this. Just add a variable in the function directly if you want

TeddyAndrieux (changes requested):
2 minor comment, but LGTM

tests/post/steps/test_logs.py
```diff
@@ -19,10 +21,12 @@ def test_logs(host):
     'nginx-ingress-default-backend',
 ))
 
+
 @then("all Pod logs should be non-empty")
 def check_logs(k8s_client):
+    POD_STATUS = 'Running'
```
TeddyAndrieux: No capital for local variable :)

TeddyAndrieux (approved):
LGTM

PR #1892, opened 2019-10-10
Update sphinxcontrib-plantuml

slaperche-scality:
**Component**:

docs

**Context**: 

The build of the doc is flaky, and the real error is hidden by a bug in `sphinxcontrib-plantuml`.

**Summary**:

Update doc depedencies
Bump required minimal version for `sphinxcontrib-plantuml`

**Acceptance criteria**: 

Doc build is still working, no visible change right now (but if we do get a build failure, we should have the real exception logged)

---

See: #1872

### DIFF
--- a/docs/requirements.in
+++ b/docs/requirements.in
@@ -4,7 +4,7 @@ colorama >= 0.3.9
 Sphinx >= 1.7
 sphinx_rtd_theme >= 0.3
 sphinxcontrib_github_alt >= 1.0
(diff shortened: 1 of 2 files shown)


TeddyAndrieux (changes requested):
I think this PR should target 2.0

slaperche-scality:
The flaky is only present in 2.4 but yeah, it doesn't hurt to update from 2.0.

Will do.

TeddyAndrieux (approved):

PR #1894, opened 2019-10-11
Backport Salt K8s module fixes

NicolasT:
These only landed in `development/2.1`, which causes trouble when upgrading/downgrading from/to `development/2.0`, while the code in `development/2.0` is as wrong as the code in `development/2.1` was before the fixes landed.


sayf-eddine-scality (approved):

NicolasT:
@bert-e create_pull_requests

TeddyAndrieux (approved):

Issue #1899, opened 2019-10-12
Show pod list when test fails if applicable


sayf-eddine-scality:
**Component**:

tests

**Why this is needed**:
Most of the flaky failed tests are hard to debug because we dont know the status of the cluster in the test time

**What should be done**:
Log/show the pod list of the cluster when a test fail (if applicable meaning the pod list can show information about the given test failure status)

**Implementation proposal** (strongly recommended):
tbd

**Test plan**:
tbd

PR #1900, opened 2019-10-12
Improvement/1899 show more info on test failure

sayf-eddine-scality:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #1899 

Test output
```
            if result.rc != 0:
                pytest.fail("Cannot resolve {}: {}".format(hostname, result.stderr))
>       pytest.fail("Dummy fail")
E       Failed: Dummy fail

tests/post/steps/test_dns.py:68: Failed
----------------------------- Captured stdout teardown ------------------------------
test failed! tests/post/steps/test_dns.py::test_dns[paramiko://bootstrap]
Running pods for namespace kube-system
calico-kube-controllers-6898dc8bd5-87xzm
calico-node-vjbgl
coredns-676c4dfdd9-6vzn9
coredns-676c4dfdd9-hq64c
etcd-bootstrap
kube-apiserver-bootstrap
kube-controller-manager-bootstrap
kube-proxy-zjrsk
kube-scheduler-bootstrap
repositories-bootstrap
salt-master-bootstrap
storage-operator-cccbfbf9-gtcm8
Running pods for namespace metalk8s-monitoring
alertmanager-prometheus-operator-alertmanager-0
prometheus-operator-grafana-6db87dcf6b-8qq65
prometheus-operator-kube-state-metrics-5ccc6d96ff-rqkp7
prometheus-operator-operator-5b597ff7b-ztl8x
prometheus-operator-prometheus-node-exporter-znlpx
prometheus-prometheus-operator-prometheus-0
Running pods for namespace metalk8s-ui
metalk8s-ui-854bf777c5-z4bwp
====================== 1 failed, 2 deselected in 12.63 seconds ======================
ERROR: InvocationError for command '/home/sayf/work/metalk8s/.tox/tests/bin/pytest --ssh-config=/home/sayf/work/metalk8s/.tox/tests/vagrant_ssh_config --hosts=bootstrap --iso-root=/vagrant/_build/root -m post and not slow and not multinode -x tests' (exited with code 1)
______________________________________ summary ______________________________________
ERROR:   tests: commands failed
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -203,3 +203,29 @@ def _verify_kubeapi_service(host):
             pytest.fail(res.stderr)
 
 # }}}
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    rep = outcome.get_result()
+    # attach the failing result to the report object
+    setattr(item, "rep", rep)
+
+
+@pytest.fixture
+def showpods(request, k8s_client):
+    yield
+    # request.node is an "item" because we use the default
+    # "function" scope
+    if request.node.rep.failed:
+        ssh_config = request.config.getoption('--ssh-config')
+        print("test failed!", request.node.nodeid)
+        for namespace in ("kube-system", "metalk8s-monitoring", "metalk8s-ui"):
+            pods = kube_utils.get_pods(
+                k8s_client, ssh_config, label=None, node="bootstrap",
+                namespace=namespace, state="Running",
+            )
+            print("Running pods for namespace {}".format(namespace))
+            print('\n'.join([p.metadata.name for p in pods]))
--- a/tests/kube_utils.py
+++ b/tests/kube_utils.py
@@ -14,7 +14,11 @@ def get_pods(
     if node:
         nodename = utils.resolve_hostname(node, ssh_config)
         field_selector.append('spec.nodeName={}'.format(nodename))
-
+    if not label:
+        return k8s_client.list_namespaced_pod(
+            namespace,
+            field_selector=','.join(field_selector),
+        ).items
     return k8s_client.list_namespaced_pod(
         namespace,
         field_selector=','.join(field_selector),
--- a/tests/post/steps/test_dns.py
+++ b/tests/post/steps/test_dns.py
@@ -54,7 +54,7 @@ def test_dns(host):
 
 
 @then(parsers.parse("the hostname '{hostname}' should be resolved"))
-def resolve_hostname(utils_pod, host, hostname):
+def resolve_hostname(utils_pod, host, hostname, showpods):
     with host.sudo():
         # test dns resolve
         result = host.run(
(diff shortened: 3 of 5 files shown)


gdemonet:
Good initiative, showing a state of the cluster after a test failure could definitely help in some situations. However, I'm not entirely sure if a list of `Running` Pods scheduled on the `bootstrap` Node would help that much. Here are a few concerns:
- when a test fails to find a Pod due to a timing issue, we would not see it appear in this list (or maybe we would, but then that would be a timing issue between test execution and this teardown fixture)
- when a test fails to find a Pod in the expected state, we would not see the actual state, nor would we have more information (e.g. if I have a Pod in `CrashLoopBackOff`, I'd like to see some logs/events for instance)
- when a test fails due to API errors (`k-a-s` is not responding/in a bad state, DNS resolution fails...), then I wouldn't get much from this list (even if the Pod is running, we don't know what's happening inside)
- when a Pod is not on the expected Node, then I won't be able to see where they are

I agree we can't tackle all issues at once, so let me suggest something slightly different:
- show a list of Pods in all namespaces (don't bother with selecting some of them)
- show them without consideration for their Node, state, or anything else
- order them by namespace
- show meaningful information for each (Node, state, and maybe recent events if the state != `Running`)

(also, small remark, but the `pytest` hook _ninja technique_ you used here could benefit from a comment in the code, took me some time to get that it's purpose is to mark the `request.node` for fixtures to know if a test run failed)

sayf-eddine-scality:
> Good initiative, showing a state of the cluster after a test failure could definitely help in some situations. However, I'm not entirely sure if a list of `Running` Pods scheduled on the `bootstrap` Node would help that much. Here are a few concerns:
> 
> * when a test fails to find a Pod due to a timing issue, we would not see it appear in this list (or maybe we would, but then that would be a timing issue between test execution and this teardown fixture)

the magic around `tryfirst=True` and `hookwrapper=True` will make sure the hook is executed as early as possible in order to catch the bad environment status while keeping a clean implementation of the test itself (better than running the code after the assert)
> * when a test fails to find a Pod in the expected state, we would not see the actual state, nor would we have more information (e.g. if I have a Pod in `CrashLoopBackOff`, I'd like to see some logs/events for instance)
> * when a test fails due to API errors (`k-a-s` is not responding/in a bad state, DNS resolution fails...), then I wouldn't get much from this list (even if the Pod is running, we don't know what's happening inside)
> * when a Pod is not on the expected Node, then I won't be able to see where they are
> 
Yes I agree, should have put the PR in draft mode as this is a POC for me, showing pods is just a the first try, we can access whatever information we judge useful. as we have ssh access to the node containerd can be accessed too
> I agree we can't tackle all issues at once, so let me suggest something slightly different:
> 
> * show a list of Pods in all namespaces (don't bother with selecting some of them)
> * show them without consideration for their Node, state, or anything else
> * order them by namespace
> * show meaningful information for each (Node, state, and maybe recent events if the state != `Running`)
> 
> (also, small remark, but the `pytest` hook _ninja technique_ you used here could benefit from a comment in the code, took me some time to get that it's purpose is to mark the `request.node` for fixtures to know if a test run failed)

Oups. I wrote a bit in the commit message but I agree the method should be explained as code comments

Lets discuss this next week maybe we change the design

sayf-eddine-scality:
Not ready for now, nor is in the near future plan, closing this PR

Issue #1839, opened 2019-10-04
Upgrade may fail when running the higstate if apiserver is not ready


charlesprost:
Probably relates to #1836 

**Component**:

'containers', 'upgrade', 'ci', 'tests'

**What happened**:

When running the highstate during the upgrade test, we face this error:
```
                          salt_|-Run the highstate_|-Run the highstate_|-state:
                              ----------
                              __id__:
                                  Run the highstate
                              __jid__:
                                  20190927150938505970
                              __run_num__:
                                  3
                              __sls__:
                                  metalk8s.orchestrate.deploy_node
                              changes:
                                  ----------
                                  out:
                                      highstate
                                  ret:
                                      ----------
                                      bootstrap:
                                          - Rendering SLS 'metalk8s-2.2.0-dev:metalk8s.salt.minion.configured' failed: Jinja variable 'dict object' has no attribute 'ip'
                              comment:
                                  Run failed on minions: bootstrap
                              duration:
                                  8212.407
                              name:
                                  Run the highstate
                              result:
                                  False
                              start_time:
                                  15:09:38.391750
```

See https://eve.devsca.com/github/scality/metalk8s/#/builders/25/builds/207/steps/17/logs/stdio

See sosreport here: [sosreport-bootstrap-2019-09-27-rznleio.zip](https://github.com/scality/metalk8s/files/3689926/sosreport-bootstrap-2019-09-27-rznleio.zip)

When the Salt states are rendered, the Salt master uses the external pillar to retrieve the IP address, and for this it calls the apiserver to find the endpoint. But if the apiserver is not available, then it fails and the pillar contains no IP.

See the start time of the salt step at `15:09:38.391750`
And we only wait for APIserver to be available at `15:09:46.605898`

May be fixed by #1836 

We have a similar one here https://eve.devsca.com/github/scality/metalk8s/#/builders/25/builds/375/steps/17/logs/stdio:
```
                                          - Rendering SLS 'metalk8s-2.3.0-dev:metalk8s.salt.minion.configured' failed: Jinja variable 'dict object' has no attribute 'ip'
                                          - Rendering SLS 'metalk8s-2.3.0-dev:metalk8s.repo.offline' failed: Jinja variable 'dict object' has no attribute 'ports'
                                          - Rendering SLS 'metalk8s-2.3.0-dev:metalk8s.container-engine.containerd.installed' failed: Jinja variable 'dict object' has no attribute 'ports'
```

**What was expected**:

[...]

PR #1903, opened 2019-10-14
salt: Delay before checking etcd status

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Flaky error missing `etcd` information right after the etcd upgrade when calling highstate on node.

**Summary**:

Wait before checking etcd node status as the pod may not be restarted
immediatly after the manifest edition

---

Fixes: #1839

### DIFF
--- a/salt/metalk8s/kubernetes/etcd/healthy.sls
+++ b/salt/metalk8s/kubernetes/etcd/healthy.sls
@@ -1,6 +1,13 @@
 include:
   - .installed
 
+Delay after etcd pod deployment:
+  module.wait:
+    - test.sleep:
+      - length: 10
+    - watch:
+      - metalk8s: Create local etcd Pod manifest
+
 Waiting for etcd running:
   http.wait_for_successful_query:
     - name: https://127.0.0.1:2379/health
@@ -12,4 +19,4 @@ Waiting for etcd running:
     - status: 200
     - match: '{"health": "true"}'
     - require:
-      - metalk8s: Create local etcd Pod manifest
+      - module: Delay after etcd pod deployment


Ebaneck:
@TeddyAndrieux just a few questions and I will approve this :)

After reading https://github.com/saltstack/salt/pull/36578, it looks like `retry` is a global state feature now.

This lets me think that we could have something like:
```
Waiting for etcd running:
  http.wait_for_successful_query:
    - name: https://127.0.0.1:2379/health
    - verify_ssl: True
    - ca_bundle: /etc/kubernetes/pki/etcd/ca.crt
    - cert:
      - /etc/kubernetes/pki/etcd/server.crt
      - /etc/kubernetes/pki/etcd/server.key
    - status: 200
    - match: '{"health": "true"}'
    - retry:
        until: True
```

or
```
Waiting for etcd running:
  http.wait_for_successful_query:
    - name: https://127.0.0.1:2379/health
    - verify_ssl: True
    - ca_bundle: /etc/kubernetes/pki/etcd/ca.crt
    - cert:
      - /etc/kubernetes/pki/etcd/server.crt
      - /etc/kubernetes/pki/etcd/server.key
    - status: 200
    - match: '{"health": "true"}'
    - retry:
        attempts: 12
        interval: 5
```

What is your take on this??

salt/metalk8s/kubernetes/etcd/healthy.sls
```diff
@@ -1,6 +1,13 @@
 include:
   - .installed
 
+Delay after etcd pod deployment:
```
gdemonet, line 4: I'm wondering, could we maybe add something in `metalk8s.static_pod_managed` that waits for `kubelet` to pick up the change instead of sporadically adding random delays?

TeddyAndrieux: It's more containerd but yes we could do that good idea. Don't really know how we can do that but I think it's doable. I will create a ticket for this but let merge this for the moment

TeddyAndrieux:
> @TeddyAndrieux just a few questions and I will approve this :)
> 
> After reading [saltstack/salt#36578](https://github.com/saltstack/salt/pull/36578), it looks like `retry` is a global state feature now.
> 
> This lets me think that we could have something like:
> 
> ```
> Waiting for etcd running:
>   http.wait_for_successful_query:
>     - name: https://127.0.0.1:2379/health
>     - verify_ssl: True
>     - ca_bundle: /etc/kubernetes/pki/etcd/ca.crt
>     - cert:
>       - /etc/kubernetes/pki/etcd/server.crt
>       - /etc/kubernetes/pki/etcd/server.key
>     - status: 200
>     - match: '{"health": "true"}'
>     - retry:
>         until: True
> ```
> 
> or
> 
> ```
> Waiting for etcd running:
>   http.wait_for_successful_query:
>     - name: https://127.0.0.1:2379/health
>     - verify_ssl: True
>     - ca_bundle: /etc/kubernetes/pki/etcd/ca.crt
>     - cert:
>       - /etc/kubernetes/pki/etcd/server.crt
>       - /etc/kubernetes/pki/etcd/server.key
>     - status: 200
>     - match: '{"health": "true"}'
>     - retry:
>         attempts: 12
>         interval: 5
> ```
> 
> What is your take on this??

Yes btw `http.wait_for_successful_query` already retry for 300sec by default so it's not an issue of retrying. The problem is when you run the state you update the local manifest and directly after you check the status so if containerd not aware of this change he not stopped the old pod so the http check may succeed using the old pod which is wrong.

salt/metalk8s/kubernetes/etcd/healthy.sls
gdemonet: > It's more containerd

Hum, no I think `kubelet` is the one watching `/etc/kubernetes/manifests` and asking to `containerd` for a restart.

> I will create a ticket for this but let merge this for the moment

:+1:

gdemonet (approved):

salt/metalk8s/kubernetes/etcd/healthy.sls
TeddyAndrieux: > Hum, no I think `kubelet` is the one watching `/etc/kubernetes/manifests` and asking to `containerd` for a restart.

Oh yes indeed it's kubelet watching the manifest and containerd doing the start/stop true

gdemonet:
> Yes btw `http.wait_for_successful_query` already retry for 300sec by default so it's not an issue of retrying. The problem is when you run the state you update the local manifest and directly after you check the status so if containerd not aware of this change he not stopped the old pod so the http check may succeed using the old pod which is wrong.

Or the state could break (yeah, too hard for Salt's `http` module, I guess :man_shrugging:) because of a `ECONNREFUSED` if the request was sent during Pod replacement.

PR #1910, opened 2019-10-14
build/debian: Do not build repo if no package

TeddyAndrieux:
**Component**:

'build'

**Summary**:

When building debian repo `metalk8s-bionic-security` we get an error
because repo has no `.deb` package.
```
Error 2 opening /packages/*.deb: No such file or directory
There have been errors!
```

---

Lets fix the build (and the queue) and investigate what need to be done here after

### DIFF
--- a/packages/debian/entrypoint.sh
+++ b/packages/debian/entrypoint.sh
@@ -45,7 +45,9 @@ buildrepo() {
     cp /distributions /repository/conf
     sed -i 's/_REPONAME_/'"${reponame}"'/g' /repository/conf/distributions
     reprepro -b /repository/ export
-    reprepro -C "$reponame" -b /repository/ includedeb bionic /packages/*.deb
+    if ls /packages/*.deb ; then
+        reprepro -C "$reponame" -b /repository/ includedeb bionic /packages/*.deb
+    fi
     chown -R "${TARGET_UID}:${TARGET_GID}" /repository
 }
 


storage-operator/go.mod
```diff
@@ -1,7 +1,6 @@
 module github.com/scality/metalk8s/storage-operator
 
 require (
-	github.com/NYTimes/gziphandler v1.0.1 // indirect
```
slaperche-scality: shouldn't commit that

packages/debian/entrypoint.sh
```diff
@@ -45,7 +45,9 @@ buildrepo() {
     cp /distributions /repository/conf
     sed -i 's/_REPONAME_/'"${reponame}"'/g' /repository/conf/distributions
     reprepro -b /repository/ export
-    reprepro -C "$reponame" -b /repository/ includedeb bionic /packages/*.deb
+    if ls /packages/*.deb ; then
```
slaperche-scality, line 48: Hope the buildchain won't complain of a missing target file, we will see :crossed_fingers:

slaperche-scality (changes requested):

storage-operator/go.mod
TeddyAndrieux: Oh indeed `git commit -a` MB :stuck_out_tongue:

packages/debian/entrypoint.sh
TeddyAndrieux: tested locally seems good

slaperche-scality (approved):
Don't forget to create an issue so that we can solve it in a nicer way

TeddyAndrieux:
> Don't forget to create an issue so that we can solve it in a nicer way

Done #1911

Issue #1588, opened 2019-08-29
Rework bootstrap script to Python


NicolasT:
Instead of using a bunch of shell-scripts for bootstrap, use a Python script instead, using the Salt Python APIs where feasible.

---

This command should be part of `metalk8sctl` python CLI

```
metalk8sctl bootstrap deploy
```

This command will have the same behavior as the current bootstrap bash script.

Except that it will be written in Python and should **not** use salt binaries at all, only interact directly with Salt using Python library (check design documentation about Centralized CLI for more detail)

TBD:
This `metalk8sctl` is part of a package and need at least python to work so may need a bootstrap

Proposition:
Having a bootstrap bash script only in the ISO (not in the `metalk8sctl` package) that will configure local repositories and install `metalk8sctl` package (+ eventual dependencies)

---
Note by @TeddyAndrieux description according to Centralized CLI design

PR #1912, opened 2019-10-14
Rewrite Solutions management

### LINKED ISSUE #1582: Design better format for solutions.yaml
**Component**:
solutions, salt

**Why this is needed**:
we need to version-ize solutions.yaml and design a format for it
**What should be done**:

**Test plan**:

### LINKED ISSUE #1614: Reword the solution manager script
**Component**:

**Why this is needed**:
Add/delete solution output is confusing
**What should be done**:
output add when adding and delete when deleting

**Test plan**:

### LINKED ISSUE #1852: Split `solution-manager` script
**Component**: salt, scripts

**Why this is needed**:

Currently, adding a Solution to the cluster using the `solution-manager.sh` script will result in two operations:

- First, mounting the ISO and exposing the container images in the cluster registry
- Then, deploying the highest version of Admin UI / CRDs / ... for a given Solution name

These two actions have different repercussions on the life of the cluster and the applications running on top of it, so having a clear separation is meaningful for end-users.

The second action could be naturally moved into Platform UI as well, in a future iteration (see #1610).

**What should be done**:

Split the Salt formulas / orchestrate runners and the `solution-manager` script to provide:

- `-a/--add ARCHIVE`: mount the archive and expose its images
- `-r/--remove ARCHIVE`: stop exposing the images and unmount the archive
- `--deploy NAME:VERSION`: deploy the singleton components for Solution `NAME` in version `VERSION`

An optional `undeploy` functionality could be interesting, though will wait for a rewrite of the script ~as a `kubectl` plugin~ in Python (see #1854).

**Test plan**:

Include this new approach in tests from #1850

### LINKED ISSUE #1854: Rewrite Solutions manager script (in Python and Go)
**Component**: scripts

**Why this is needed**: Logic is getting too complex for Bash

**What should be done**:

- Rewrite the Bootstrap-local parts in Python (import/unimport, maybe activate/deactivate)
- Rewrite the other parts in a `kubectl` plugin

**Test plan**: Update tests introduced by #1850 and #1852 

---
Epic: #2163

gdemonet:
**Components**: salt, solutions, scripts

**Context**: Improvement of Solutions management UX

---

**To do**:

- [ ] Rework packaging of the CLI (no need to make it redistributable)
- [ ] Rework logging with verbosity toggle (need to think about more than two verbosity levels, if needed in the future)
- [x] Remove `Environment` CRD:
    - [x] Replace with `Namespace`(s) metadata:
        - Label `solutions.metalk8s.scality.com/environment=<environment_name>`
        - Annotation `solutions.metalk8s.scality.com/environment-description=<short_desc>`
    - [x] Adapt MetalK8s UI:
        - List environments based on the label (group namespaces having the same env name)
        - Create such a namespace instead of an `Environment` CR when asked to
        - Add/update ConfigMap describing deployed versions of Solution Operators in that namespace
- [x] Expose available versions to Admin UIs through ConfigMap (even if Salt doesn't use it for state management, it updates it for consumption)
- [x] Rework Solution images repository handling
    - [x] Rework configuration of Operators (need to have some config file, not just env var)
    - [ ] Rework configuration of Admin UI (similar to operator, to know which repositories exist for a Solution)
- [ ] Document how Operators/UIs should be defined so they can still work outside of MetalK8s (maybe requiring some manual Operations for now)

---

**Summary**:

### Part I - Rework Salt implementation

_The previous approach relied on some custom state management, where Salt uses a `ConfigMap` to store a list of available Solution versions, to compare with the desired state configured in `/etc/metalk8s/solutions.yaml`._

- Rework configuration file format to let users declare which Solution versions are *active* (meaning, their Admin UIs / CRDs should be deployed)
- Rework ext_pillar `metalk8s:solutions` to list *available* Solutions in the cluster, based on simple observations (active mounts, existing UI Service objects in `metalk8s-solutions` Namespace...) of the actual state, instead of the `ConfigMap` used before
- Use a single formula (`metalk8s.solutions.available`) to (un)mount archives and (un)configure the cluster OCI registry
- Use a single orchestrate (`metalk8s.orchestrate.solutions.deploy-components`) to (un)deploy Solution _components_ (Admin UI, CRDs)

### Part II - Rewrite `solution-manager.sh` in Python

_To use the new states and config file format, and provide a comprehensive user experience, we rewrite the script in Python with `argparse`._

- Define the skeleton for a `metalk8s` CLI entrypoint, using git-style subcommands (implemented with `argparse`)
- Add logging to match existing behaviour in Shell scripts
- Add Salt helpers for running commands on either the Salt Minion or the Salt Master
- Add subcommands to manage Solutions:
    - `/path/to/metalk8s solutions add /root/some-solution.iso`
    - `/path/to/metalk8s solutions remove /root/some-solution.iso`
    - `/path/to/metalk8s solutions deploy some-solution --latest`
    - `/path/to/metalk8s solutions deploy some-solution --use-version 1.2.3-alpha1`
    - `/path/to/metalk8s solutions deploy some-solution --delete`

**Acceptance criteria**:

- Bootstrap still works, and after it's done, all Solution dependencies are deployed:
    - `metalk8s-solutions` Namespace exists
    - `ui-branding` ConfigMap exists in `metalk8s-solutions` namespace
    - _Environment_ CRD exists (`environments.solutions.metalk8s.scality.com`)

- One can add a Solution archive to the cluster (see #1876 for how to build the example ISO):
    - `/srv/scality/metalk8s-2.4.0-dev/metalk8s solutions add /vagrant/.../example-solution-0.1.0-dev.iso`
    - After this is done, one can retrieve container images from this ISO: `crictl pull -D metalk8s-registry-from-config.invalid/example-solution-0.1.0-dev/base-server:0.1.0-dev` should work fine
    - `salt-call pillar.get metalk8s:solutions:config:archives` should contain `/vagrant/.../example-solution-0.1.0-dev.iso`
    - `salt-call pillar.get metalk8s:solutions:available:example-solution` should show the relevant information about this archive and its mountpoint, and `active` will be `False`

- One can deploy components for a Solution version:
    - `/srv/scality/metalk8s-2.4.0-dev/metalk8s solutions deploy example-solution --latest`
    - After this is done, one can access the Solution UI (`kubectl get svc -n metalk8s-solutions example-solution-ui` to retrieve the UI port, its IP should be the workload-plane IP)

---

Fixes: #1582
Fixes: #1614
Fixes: #1852 
Fixes: #1854

See: #1588


gdemonet:
Forgot that I need to adjust the UIs to work without the ConfigMap... For reviewers, please don't focus on UIs right now.

salt/_modules/metalk8s_solutions_k8s.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if labels.get("app.kubernetes.io/component") != "ui":
+            # Not an Admin UI, ignoring for this list
+            continue
+
+        try:
+            solution_name = labels["app.kubernetes.io/part-of"]
+            solution_version = labels["app.kubernetes.io/version"]
+        except KeyError:
+            # FIXME: ignoring invalid Service objects for now, though it may
```
TeddyAndrieux: At least warning IMO

salt/_modules/metalk8s_solutions.py
```diff
@@ -1,186 +1,122 @@
...
+    return __virtualname__
+
+
+def read_config():
+    """Read the SolutionsConfiguration file and return its contents.
+
+    Empty containers will be used for `archives` and `active` in the return
+    value.
+
+    The format is expected to resemble the following example:
```
TeddyAndrieux:
```suggestion
    The format should look like the following example:
```

salt/metalk8s/solutions/available.sls
```diff
@@ -0,0 +1,120 @@
...
+  file.managed:
+    - source: {{ mount_path }}/registry-config.inc.j2
+    - name: {{ repo.config.directory }}/{{ machine_name }}-registry-config.inc
+    - template: jinja
+    - defaults:
+        repository: {{ machine_name }}
+        registry_root: {{ mount_path }}/images
+    - require:
+      - file: Container images for Solution {{ display_name }} exist
+    - onchanges_in:
```
TeddyAndrieux: Would suggest `require_in` because `onchanges_in` that mean if for one reason only one `Expose container images for Solution blabla` fail you will not run the `metalk8s.repo.installed` state and if you run again and no changes made you will not run the state. (This sentence does not look really clear)

Example:
You want to make toto and tata solutions available:
- toto not configured at all
- tata already has a good `registry-config.inc` but for what ever reason product.txt not here

1- Launch this state => toto get configured with good `registry-config.inc` tata fail => `metalk8s.repo.installed` is not executed because one "onchanges" state failed
2- Fix tata product.txt
3- Launch this state again => All green nothing change (repo still not configured)

With a `require_in` the 3- will run the `metalk8s.repo.installed` state.

(For me `onchanges` it's something only usefull in really specific case for state that is not idempotent and make "dangerous" changes)

salt/metalk8s/solutions/available.sls
```diff
@@ -0,0 +1,120 @@
...
+    - persist: true
+    - match_on:
+      - name
+    - require:
+      - file: Mountpoint for Solution {{ display_name }} exists
+
+{# Validate the archive contents
+   TODO: This should be moved before mounting the solution's ISO, using some
+   custom module #}
+Product information for Solution {{ display_name }} exists:
```
TeddyAndrieux, line 56: If you try to make 2 different version of the same solution at the same time you will have conflict here as you will have 2 states with the same name.

You should change the display_name to also contains the version or add something unique for each state so "display_name" + "version" or "mount_path" but not "display_name" only

salt/metalk8s/solutions/available.sls
```diff
@@ -0,0 +1,120 @@
...
+
+  {#- Unmount all Solution ISOs mounted in /srv/scality not referenced in
+      the configuration file #}
+  {%- set available = pillar.metalk8s.solutions.available %}
+  {%- for name, versions in available.items() %}
+    {%- for info in versions %}
+      {%- if info.archive not in configured %}
+        {%- set display_name = info.display_name ~ ' ' ~ info.version %}
+        {%- if info.active %}
+Cannot remove archive for active Solution {{ display_name }}:
```
TeddyAndrieux, line 91: Ditto for remove you need to have a unique state name or if you remove 2 different versions of the same solution the state will fail

salt/metalk8s/solutions/available.sls
```diff
@@ -0,0 +1,120 @@
...
+  test.fail_without_changes:
+    - name: Solution {{ display_name }} is still active, cannot remove it
+
+        {%- else %}
+Remove container images for Solution {{ display_name }}:
+  file.absent:
+    - name: {{ repo.config.directory }}/{{ name }}-registry-config.inc
+    - onchanges_in:
```
TeddyAndrieux: ditto :)

salt/_modules/metalk8s_solutions.py
```diff
@@ -1,186 +1,122 @@
...
+    for mountpoint, mount_info in solution_mounts:
+        solution_info = __salt__['metalk8s.archive_info_from_iso'](
+            mount_info['alt_device']
+        )
+        machine_name = solution_info['name'].replace(' ', '-').lower()
+        version = solution_info['version']
+
+        result[machine_name].append({
+            'display_name': solution_info['name'],
+            'machine_id': '{}-{}'.format(machine_name, version),
```
TeddyAndrieux: `solution_id` may be better no ? Because nothing really related to "machine" here right ?

TeddyAndrieux (changes requested):
First quick review (only for salt part, will do another one for scripts :) )

But looks pretty good

scripts/metalk8s
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# set -e
+set -u
+set -o pipefail
+
+pushd "$(dirname ${BASH_SOURCE[0]})" > /dev/null
+SCRIPTS_DIR="$(pwd)"
+popd > /dev/null
```
TeddyAndrieux: Hmm personnaly prefer somethink like that 
```
SCRIPTS_DIR="$(dirname "$(readlink -f "$0")")"
```
but maybe `pushd` `popd` better I'm not a bash expert @alexandre-allard-scality ? :)

scripts/metalk8s
```diff
@@ -0,0 +1,45 @@
...
+
+if [ ! -d "${SCRIPTS_DIR}" ]; then
```
TeddyAndrieux: Wut ? It's possible ?

scripts/metalk8s_cli/parsers.py
```diff
@@ -0,0 +1,30 @@
...
+    """Allow toggling verbose mode."""
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('-v', '--verbose', action='store_true')
+    return parser
+
+
+def build_logfile_parser():
+    """Allow defining a logfile for this command execution."""
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('--logfile',
```
TeddyAndrieux: `-l` also

salt/metalk8s/solutions/available.sls
gdemonet: See L.29: `{% set display_name = solution.name ~ ' ' ~ solution.version %}`

salt/_modules/metalk8s_solutions.py
gdemonet: The point is this name/ID is for machine use, while `display_name` is intended for humans. Maybe I can add a comment to explicit this distinction.

salt/_modules/metalk8s_solutions_k8s.py
gdemonet: Good idea, will do

scripts/metalk8s_cli/mixins/log.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def prepare_logfile(self):
+        if os.path.exists(self.logfile):
+            return
+
+        dirname = os.path.dirname(self.logfile)
+        if not os.path.exists(dirname):
+            os.makedirs(dirname)
+
+        open(self.logfile, 'a').close()
```
TeddyAndrieux: Does this really needed ?

salt/metalk8s/solutions/available.sls
gdemonet: OK I get the reasoning, thanks for the explanation! Will do :)

scripts/metalk8s_cli/mixins/log.py
```diff
@@ -0,0 +1,251 @@
...
+    """Provide methods to log messages from a command.
+
+    Always log to a file, and log partial info to the console.
+    A logging command should rely on its `log_active_run` and `log_step`
+    context managers within its `run` method.
+    """
+    def __init__(self, args):
+        super(LoggingCommandMixin, self).__init__(args)
+        self._logger = None
+        self.logfile = os.path.abspath(args.logfile)
```
TeddyAndrieux: For me we should do like in the ring installer with one global log file and one log file specific for each steps :)

So if you run bootstrap each log are in /var/log/metalk8s/bootstrap.log and /var/log/metalk8s/metalk8s.log, if you run upgrade each log are in /var/log/metalk8s/upgrade.log and /var/log/metalk8s/metalk8s.log ... (but maybe not part of this issue/PR IMO)

scripts/metalk8s_cli/mixins/log.py
```diff
@@ -0,0 +1,251 @@
...
+            # Errors should be caught and logged by the wrapped code, and
+            # re-raised as `CommandError`s
+            yield
+        except CommandError:
+            self.print_and_log('The script will now exit')
+            sys.exit(1)
+        finally:
+            self._stop_run_log()
+
+    def print_and_log(self, message, level=logging.INFO, keep_line=False):
```
TeddyAndrieux: Hmm it's weird to call this function each time you want to log something.

For me this class should inherit from classic Logging class and dynamicaly change the level of the console handler when verbose mode enable

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+
+class SaltCommandMixin(object):
+    def __init__(self, args):
+        super(SaltCommandMixin, self).__init__(args)
+
+    def run_salt_minion(
+        self, command, local=False, outputter=None, color=True,
+        saltenv=None, pillar=None,
+    ):
+        full_cmd = ['salt-call']
```
TeddyAndrieux: `--retcode-passthrough` I think even if we don't want to fail I think it's good to have the return code

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if saltenv is not None:
+            full_cmd.append('saltenv={}'.format(saltenv))
+        if pillar is not None:
+            full_cmd.append("pillar='{}'".format(json.dumps(pillar)))
+
+        try:
+            return subprocess.check_output(full_cmd, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as exc:
+            raise CommandError(
```
TeddyAndrieux: Do we really want to raise here ? For me we may want some time to do something with the output in case of error.

For me we should either:
1- always return a dict with `stdout` `stderr` `retcode`
2- when `retcode != 0` raise something that contain `stderr` `stdout` and `retcode` (I mean a dict containing these informations)

For me 2- look like a good solution most of the time we will just not catch the exception and will get the msg just here but for rare special case we can `try except` and get the stderr stdout, retcode and do what ever we want with this

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+                'Failed to execute "{}" [retcode {}]:\n{}'.format(
+                    ' '.join(full_cmd), exc.returncode, exc.output
+                )
+            )
+
+    def run_salt_master(self, command):
+        # TODO: add cache for the salt-master container ID
```
TeddyAndrieux: I don't know if we can/want really to have cache for this.
Because if we cache this we need either
1- a variable to "force caching again" 
2- something to check that the container ID still good (which is basically the command executed in `get_salt_master_container`)

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+    def _build_args(self, cmd):
+        if isinstance(cmd, six.string_types):
+            return shlex.split(cmd)
+        return cmd
+
+    def _get_json_from_minion(self, command, **kwargs):
+        out_string = self.run_salt_minion(
+            command, outputter='json', color=False, **kwargs
+        )
+        return json.loads(out_string)["local"]
```
TeddyAndrieux: Care here since you do `stderr=subprocess.STDOUT` in subprocess call if for what ever reason salt log something to stderr you will likely not be able to parse this as json.

scripts/metalk8s_cli/solutions/config.py
```diff
@@ -0,0 +1,123 @@
...
+        kind = data.get('kind')
+        assert api_version and kind, (
+            'SolutionsConfiguration must have a `kind` and `apiVersion`.'
+        )
+        assert kind == self.KIND, (
+            'Wrong kind ({}), must be "{}".'.format(kind, self.KIND)
+        )
+        assert api_version in self.API_VERSIONS, (
+            'API version not supported: {}'.format(api_version)
+        )
```
TeddyAndrieux, line 21: Hmm you do the same check as in the salt module. That mean if for what ever reason we need to change something we need to change it in salt and in the script. Maybe you should use a salt module to check this config file.

TeddyAndrieux (changes requested):
Several comments (not tested for the moment) but looks really better than our current bash scripts gj

salt/metalk8s/solutions/available.sls
TeddyAndrieux: Oh yes MB through it was the `display_name` from `list_available` salt modules

salt/_modules/metalk8s_solutions.py
TeddyAndrieux: Oh okay machine in that sense

slaperche-scality: Maybe simply call them `name` and `id`, as `name` are usually for human whereas `id` are usually for computer.

scripts/metalk8s
gdemonet: I stole this whole Bash script from @NicolasT ;)
But I agree your one-liner looks more explicit to me, being a noob at Shell scripting.

scripts/metalk8s
gdemonet: :man_shrugging: I'll remove the check I guess

scripts/metalk8s_cli/parsers.py
gdemonet: OK

scripts/metalk8s_cli/mixins/log.py
gdemonet: Indeed it's not required, file will get created at first log (so should be the `self._init_run_log` call actually). Will remove it :)

scripts/metalk8s_cli/mixins/log.py
gdemonet: Good suggestion! I was not sure about splitting command logs, but having a _catch-all_ logfile in addition is smart :)

I'll do it in this PR, it's not much of a change anyway (iff we agree that `/var/log/metalk8s/metalk8s.log` is a non-configurable value for the catch-all logfile).

scripts/metalk8s_cli/mixins/log.py
gdemonet: > this class should inherit from classic Logging class

WDYM?

> dynamicaly change the level of the console handler when verbose mode enable

Huh, I did this first, then decided otherwise. The rationale is the following: it's a CLI, so there's a big difference between what goes in `stdout` and `stderr`. IMO (probably wrong), _verbosity_ of the script affects what goes in `stdout`, while log messages go in `stderr` (and we may give knobs to configure this logging for users). This way, if commands are supposed to return parsable output, a consumer can easily redirect `stderr` and avoid any risk when parsing (as you mentioned for the `SaltCommandMixin._get_json_from_minion` :wink:).

Now, I wanted to have the `print`-ed messages in the logfile as well, which is why there's this `print_and_log` method.

> it's weird to call this function each time you want to log something.

I think some occasions may justify using "log-only" messages for `INFO` level and below, but for now all of those were user-oriented. Note that errors are not `print`-ed.

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Good point, will do (those helpers are really a draft, so thanks for the help!)

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Indeed, I was thinking of implementing the second approach. Using `check_output(cmd, stderr=PIPE)` should do the trick.

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Indeed, I was actually thinking about giving control to the caller (same as for `saltenv` in `run_salt_minion`, callers store it once and can pass it to each call). This way, if a command expects the container ID to have changed, it can refresh its value, while otherwise saving the cost of running the subprocess again.

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Indeed, will change. I think I'll remove `check_output` and use `run` instead, that will allow more control for callers.

scripts/metalk8s_cli/solutions/config.py
gdemonet: Good point. Given it's a simple `salt-call --local`, I guess the performance penalty of delegating this to Salt is acceptable (this CLI is already slow as hell due to the numerous calls to Salt master - might be good to profile a bit more what takes time, I fear having expensive ext_pillars is not that good for responsiveness of the system).

scripts/metalk8s_cli/mixins/salt.py
slaperche-scality: `subprocess.run` was introduced in Python 3.5, dunno if it's "old enough" to be used in this context.

gdemonet: Again... OK so I'll do another monkey-patch of this `subprocess.run` as we do for `argparse`, the ugliness of these should encourage us in only using Python 3 :)

slaperche-scality: I mean, it's our custom script so maybe we can use Python 3, but even then I'm not sure CentOS 7 comes with Python 3.5+ (it's Python 3.4, no?)

gdemonet: There's no Python 3 pre-installed on CentOS 7. There's 3.6 in `base`, though most of the usual dependencies (e.g. `six` or `PyYAML`) are not, so we'd need to include them in our archive, and somehow install them before execution of the bootstrap script (since it's intended to run through this Python CLI).

I think it could be feasible, but would cost more efforts for us. So let's go with something simpler for now :) and once everything is moved to Python 3 (Salt, notably), then we'll be able to enforce it for the script as well.

gdemonet: Mmh, monkey-patch is not that easy, there's quite a lot of changes in the `run` API actually, when compared to the capabilities of `subprocess` in 2.7... So I'll ignore `run` and only expose `stderr` in case of error.

TeddyAndrieux: Why do not use directly `subprocess.Popen` ?

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+        full_cmd = [
+            'crictl', 'exec', self.get_salt_master_container()
+        ]
+        master_cmd = ['salt-run', '--force-color'] + self._build_args(command)
+        full_cmd.extend(master_cmd)
+
+        try:
+            return subprocess.check_output(
+                # FIXME: pillar args... need to escape?
+                ' '.join(full_cmd), shell=True,
```
slaperche-scality: hum yeah, this looks dangerous, passing a string coming from non-controlled input (pillar) and let the shell interpret it, with no escaping.

Why not use a list of string here?
And if you absolutely want to use a string, then use `shlex.quote`.

Now, being Python 2 it's probably not available (otherwise it's not fun), but IIRC the implementation is not that complex (but still, would be better to use a list of string and keep `shell=False`).

scripts/metalk8s_cli/mixins/salt.py
```diff
@@ -0,0 +1,119 @@
...
+
+        full_cmd.extend(self._build_args(command))
+
+        if saltenv is not None:
+            full_cmd.append('saltenv={}'.format(saltenv))
+        if pillar is not None:
+            full_cmd.append("pillar='{}'".format(json.dumps(pillar)))
+
+        try:
+            return subprocess.check_output(full_cmd, stderr=subprocess.STDOUT)
```
slaperche-scality: Why do we, at several places, redirect `stderr` into `stdout`?

Why not using `subprocess.Popen` and then use `communicate` to get the output (`retcode` will be in the object returned by `Popen`)?

scripts/metalk8s_cli/mixins/salt.py
gdemonet: My bad, the `FIXME` comment was not explicit. The issue I was faced with when using a list of strings, the value `"pillar='{bootstrap_id: bootstrap}'"` is not well interpreted by Salt, and fails with some `invalid pillar` exception. Maybe it's not well escaped when sent to the execution context, so I'll give a try to `shlex.quote` on this specific argument.

scripts/metalk8s_cli/mixins/salt.py
gdemonet: The goal was to print `stdout`/`stderr` if the user requests `--verbose` mode only, but as @TeddyAndrieux suggested, we should not mix them if we want to parse the `stdout` result.
I moved to `Popen.communicate` since then, push is coming soon

scripts/metalk8s_cli/mixins/salt.py
gdemonet: `run` provides a bit more flexibility, so I made a _poor man's_ variant of it using `Popen.communicate`. It lacks `KeyboardInterrupt` or timeout management though, but should be fine for a first version.

scripts/metalk8s_cli/mixins/salt.py
slaperche-scality: Weird, the whole point of using a list of string is to avoid the need for escaping (because the values are directly passed to the process).

This works¹ for me:

```python
pillar = {"update_archive_name": "new_archive_name.zip"}
cmd = ['salt', '*', 'state.highstate', 'pillar', json.dumps(pillar)]
subprocess.run(cmd)
```

¹: works as in `doesn't complain about the pillar`, but it fails with `No Top file or master_tops data matches found` because my local salt is not configured :stuck_out_tongue:

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Now there's a `_run_process` helper, which gives you a `CompletedProcess` namedtuple, with keys `args`, `returncode`, `stdout` and `stderr`. In case `returncode` was not 0, it instead raises a `CommandSubprocessError`, with the same attributes, for callers to catch and log as they want.

scripts/metalk8s_cli/mixins/salt.py
slaperche-scality: What's the deal with `KeyboardInterrupt`?

For timeout indeed, there is nothing in Python 2…

gdemonet: So, `run` in Python 3 relies on using `Popen` as a contextmanager, which performs clean-up of open streams and handles `KeyboardInterrupt` errors when doing so. I just ignored the clean-up of stdin (which may raise other errors), as it's not in our use-cases for now. See [`subprocess.Popen.__exit__`](https://github.com/python/cpython/blob/master/Lib/subprocess.py#L998-L1024)

scripts/metalk8s_cli/mixins/salt.py
gdemonet: I did not have issues with direct usage of `salt-call` either, the issue arose when sending to Salt Master through `crictl exec`... Did not test yet if the `shlex.quote` approach helps, so we'll see :)

slaperche-scality: Ha, you may need to wrap the pillar between single quote then.

I think we're doing something like that in the expansion test `tests/install/steps/test_expansion.py`, maybe you can get an idea there.

scripts/metalk8s_cli/solutions/config.py
gdemonet: Also, since Salt is responsible for defining format/usage of this config file (same as it would be for `/etc/metalk8s/bootstrap.yaml` I guess), I think it should also be responsible for managing its path (for now, it's hardcoded in `__salt__['metalk8s_solutions.SOLUTIONS_CONFIG_FILE']`). Which means I should remove the `--solutions-config` argument.

scripts/metalk8s_cli/mixins/salt.py
gdemonet: Huh, it seems the tests are doing it with `shell=True` as well (using `host.run(some_format_string)`) :/ and through `kubectl`. Tried switching to `kubectl` as well, used different escaping strategies, nothing works. I'll go back to `shell=True` as well here, actually `pillar` is not provided by the user, so I'd say it's not really _unsafe_ to use it within the CLI code.

slaperche-scality: Yeah indeed, the pillar value is hardcoded here so it should be safe: let's go for that.
(but it's weird that we can't make it works without this)

buildchain/buildchain/iso.py
```diff
@@ -107,7 +107,53 @@
         destination_directory=constants.ISO_ROOT,
         source_prefix=Path('scripts'),
         task_dep=['_iso_mkdir_root']
-    )
+    ),
+    helper.FileTree(
```
NicolasT, line 126: I wonder whether it wouldn't be better to turn `metalk8s_cli` into a proper Python package, `sdist` it during build into `_build/...` (not `_build/root`), then 'install' it in `_build/root/whatever` (also during build).

buildchain/buildchain/iso.py
```diff
@@ -107,7 +107,53 @@
...
+        basename='_iso_add_tree',
+        files=(
+            Path('metalk8s'),
+
+            Path('metalk8s_cli/__init__.py'),
```
NicolasT, line 131: Not sure this should be in the project root. Is this *the* CLI to work with MetalK8s? I.e., will it include the stuff we wanted to put in the `kubectl` plugin?

buildchain/buildchain/iso.py
```diff
@@ -107,7 +107,53 @@
...
+            Path('metalk8s_cli/__main__.py'),
+            helper.TemplateFile(
```
NicolasT, line 133: Not sure we need this, can read it (at runtime) from `product.txt`, since this is not a redistributable/system-wide-installable package

buildchain/buildchain/salt_tree.py
```diff
@@ -221,8 +221,10 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/addons/ui/deployed/namespace.sls'),
     Path('salt/metalk8s/addons/ui/deployed/ui.sls'),
 
-    Path('salt/metalk8s/addons/solutions/deployed.sls'),
-    Path('salt/metalk8s/addons/solutions/environment-crd.sls'),
+    Path('salt/metalk8s/addons/solutions/deployed/init.sls'),
+    Path('salt/metalk8s/addons/solutions/deployed/environment-crd.sls'),
```
NicolasT: Oh, an *environment* is an actual CRD? I though it maps to a *Namespace*?

buildchain/buildchain/salt_tree.py
```diff
@@ -453,16 +451,16 @@ def _get_parts(self) -> Iterator[str]:
...
-    Path('salt/_modules/metalk8s.py'),
+    Path('salt/_modules/metalk8s_kubernetes.py'),
     Path('salt/_modules/metalk8s_network.py'),
     Path('salt/_modules/metalk8s_package_manager.py'),
-    Path('salt/_modules/metalk8s_volumes.py'),
+    Path('salt/_modules/metalk8s_solutions_k8s.py'),
     Path('salt/_modules/metalk8s_solutions.py'),
-
+    Path('salt/_modules/metalk8s_volumes.py'),
+    Path('salt/_modules/metalk8s.py'),
```
NicolasT: That's a surprising reordering.

NicolasT (changes requested):
Intermediate POST

buildchain/buildchain/salt_tree.py
gdemonet: No, an `Environment` is actually a "group" of `Namespace`s with information about which version of a Solution (Operator) is running in such `Namespace`s (one namespace per Solution, to keep things organized by default). So it's a CRD (for simplicity, although we may want to change this in the future), but it's not tied to any automatic action through a controller.

buildchain/buildchain/iso.py
gdemonet: ~OK, will do~ Well, depends on the discussion regarding `kubectl` plugins being included in this Python package (which may contain multiple entrypoints). It may make sense to know the sources version when they are copied outside of the ISO.

buildchain/buildchain/iso.py
gdemonet: I think we can make it _the_ CLI for MetalK8s, and as a Python package, export `entrypoints` for some commands we want to install/use as `kubectl` plugin(s). Would this be a problem?
I also don't see why it should not be an installable/shareable package, as `kubectl` plugins would be.

buildchain/buildchain/iso.py
gdemonet: I'll give it a try :)

buildchain/buildchain/salt_tree.py
gdemonet: Indeed, my editor shows them in this order, but I prefer the "usual" lexicographical ordering (such as the one for Python strings). I'll change this.

buildchain/buildchain/salt_tree.py
gdemonet: This concept was introduced to help users unfamiliar with namespaces to correctly manage Solutions without using a single namespace for a bunch of Solutions (the UI enforces a single Operator per Solution in each Environment, so actually each Operator has its own Namespace)

buildchain/buildchain/iso.py
gdemonet: OK, I think it's worth a dedicated issue, so we can think about different technical solutions (and what we want to do in the future for this CLI). I'll open one

scripts/metalk8s_cli/solutions/config.py
gdemonet: Done, see 65d8738

salt/_modules/metalk8s_solutions.py
```diff
@@ -1,186 +1,203 @@
...
-SOLUTIONS_CONFIG_MAP = 'metalk8s-solutions'
-SOLUTIONS_CONFIG_MAP_NAMESPACE = 'metalk8s-solutions'
-SOLUTIONS_CONFIG_FILE = '/etc/metalk8s/solutions.yaml'
-try:
-    import kubernetes.client
-    from kubernetes.client.rest import ApiException
-    from urllib3.exceptions import HTTPError
-except ImportError:
-    HAS_LIBS = False
+import yaml
```
TeddyAndrieux, line 11: Why did you move `import yaml` ?
```
C: 11, 0: third party import "import yaml" should be placed before "from salt.exceptions import CommandExecutionError" (wrong-import-order)
```

scripts/metalk8s
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# set -e
+set -u
+set -o pipefail
+
+SCRIPT_PATH="$(readlink -f "${BASH_SOURCE[0]}")"
+SCRIPTS_DIR="$(cd "$(dirname "${SCRIPT_PATH}")" && pwd)"
```
TeddyAndrieux, line 7: Hmm why `cd` + `pwd` ? `dirname "$SCRIPT_PATH"` is not enough ?

scripts/metalk8s_cli/mixins/log.py
TeddyAndrieux: > WDYM?

Why not create a `class Metalk8sLogger(logging.getLoggerClass())` ?
Then we could use directly the classic `debug`, `info` ...

> I think some occasions may justify using "log-only" messages for `INFO` level and below, but for now all of those were user-oriented. Note that errors are not `print`-ed.

For me the only things we want to print is the name of the step + result "Deploying bootstrap node (this may take a while)... done[200s]" others stuff may only go in log and if you enable verbose mode then you change the handler for console so that log also printed in the console.

So IMO printing of the beginning of the step + result should be done in `log_step` function directly. Instead of `print_and_log` we should `print` and `log` (maybe something with more detail if needed)

(Btw it's only my opinion feel free to dismiss this one if you want :) )

scripts/metalk8s_cli/solutions/config.py
```diff
@@ -0,0 +1,51 @@
+"""Helper module for managing a SolutionsConfiguration file."""
+import argparse
+import os.path
+
+import yaml
+
+from metalk8s_cli.exceptions import CommandInitError, CommandError
```
TeddyAndrieux, line 7: Look like all these imports are no longer used here. Only salt is needed

TeddyAndrieux (changes requested):
Few comments about the code (not tested yet, but I will)

salt/_modules/metalk8s_solutions.py
gdemonet: Wut?! As your linter (`pylint` I guess) says, `yaml` is third party, not standard lib. So I put them together, and since `Y > S`, I put it after `salt`. Now, maybe you have a `pylint` plugin that wants `salt` imports after those, but I'm quite unsure as to why that would be a good idea.

scripts/metalk8s
gdemonet: I think the rationale was "if `dirname ...` is not an actual directory, then it will fail early"

gdemonet: But I can remove it if you'd like

scripts/metalk8s_cli/mixins/log.py
gdemonet: > Why not create a class `Metalk8sLogger(logging.getLoggerClass())` ?
Then we could use directly the classic `debug`, `info` ...

Why would we need a custom logger? Isn't the native one sufficient? If you want the `LoggingCommandMixin` to **be** a `Logger` subclass, then I disagree, we should favor composition in that kind of situation.

> if you enable verbose mode then you change the handler for console so that log also printed in the console

See above as to why I don't want to do this ;)

> printing of the beginning of the step + result should be done in `log_step` function directly. Instead of `print_and_log` we should `print` and `log` (maybe something with more detail if needed)

OK, I think that's a good middle-ground. I'll need to filter out `INFO` messages from the `console` handler to avoid duplication, but OK.

salt/_modules/metalk8s_solutions.py
TeddyAndrieux: Oh yes indeed I use saltpylint. 
https://docs.saltstack.com/en/latest/topics/development/conventions/style.html#linting

So yes it's 
```
# import builtins

# import thirdparty

# import salt_stuff
```

scripts/metalk8s
TeddyAndrieux: Hmm maybe simpler/cleaner to do `test -d` for this.
But I don't see a case where a `dirname` of something who come from a `readlink -f` can give a non-directory file :confused:

scripts/metalk8s_cli/mixins/log.py
TeddyAndrieux: > Why would we need a custom logger? Isn't the native one sufficient? If you want the `LoggingCommandMixin` to **be** a `Logger` subclass, then I disagree, we should favor composition in that kind of situation.

Not really needed but if we want to have some special feature this can help but for the moment maybe not needed.

> > if you enable verbose mode then you change the handler for console so that log also printed in the console
> 
> See above as to why I don't want to do this ;)

Which one ?
You can change the log level of an handler without having a subclass of `Logger`

> OK, I think that's a good middle-ground. I'll need to filter out `INFO` messages from the `console` handler to avoid duplication, but OK.

Why ? We should have no duplication here if you never print in a logging function but only log

buildchain/buildchain/iso.py
gdemonet: Discussion concluded that we will not include `kubectl` plugin(s) in this package, as they would probably be written in Go to make use of the available libraries.

gdemonet:
Closing as this PR is not actively maintained. The code here should be useful for #2163 efforts (especially on bootstrapping a Python CLI).

Issue #1361, opened 2019-07-04
Write a "Commit Best Practice" guide


slaperche-scality:
**Component**:

documentation

**Why this is needed**:

We still have a lot of commits with no proper commit messages.
This may be due to people still not used to write them, but can also be due to people not being aware/convinced by the benefits of writing good commit message.

**What should be done**:

Write a best practice document about commit messages, with rules, explanation and examples.
During the review, everyone will be able to give their feedback until we reach a consensus.

Exceptionally, this pull request will require everyone's approval to be merged

**Implementation proposal**:

Write a document similar to `docs/developer/best_practices/python.rst`

**Test plan**: N/A

PR #1927, opened 2019-10-16
Add commit best practices

slaperche-scality:
**Component**:

docs

**Context**: 

We want to improve the long term maintainability, and having goog Git history is a important part of it

**Summary**:

Define commit best practices.

**Acceptance criteria**:

Everyone's approval

---

Closes: #1361

### DIFF
--- a/docs/developer/best_practices/commit.rst
+++ b/docs/developer/best_practices/commit.rst
@@ -1,0 +1,370 @@
+Commit Best Practices
+---------------------
+
+How to split a change into commits
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Why do we need to split changes into commits
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This has several advantages amongst which are:
+
+- small commits are easier to review (a large pull request correctly divided
+  into commits is easier/faster to review than a medium-sized one with less
+  thought-out division)
+
+- simple commits are easier to revert (`e866b01f0553 <https://github.com/scality/metalk8s/commit/e866b01f05535925e80da20aca00417904422433>`_/`8208a170ac66 <https://github.com/scality/metalk8s/commit/8208a170ac66912ace018bcd00c058ad214d169b>`_)/cherry-pick
+  (`Pull request #1641 <https://github.com/scality/metalk8s/pull/1641>`_)
+
+- when looking for a regression (e.g. using ``git bisect``) it is easier to
+  find the root cause
+
+- make ``git log`` and ``git blame`` way more useful
+
+Examples
+^^^^^^^^
+
+The golden rule to create good commits is to ensure that there is only one
+"logical" change per commit.
+
+Cosmetic changes
+::::::::::::::::
+
+Use a dedicated commit when you want to make cosmetic changes to the code
+(linting, whitespaces, alignment, renaming, etc.).
+
+Mixing cosmetics and functional changes is bad because the cosmetics (which
+tend to generate a lot of diff/noise) will obscure the important functional
+changes, making it harder to correctly determine whether the change is correct
+during the review.
+
+Example (Pull request `#1620 <https://github.com/scality/metalk8s/issues/1620>`_):
+
+- one commit for the cosmetic changes: `766f572e462c6933c8168a629ed4f479bb68a803 <https://github.com/scality/metalk8s/commit/766f572e462c6933c8168a629ed4f479bb68a803>`_
+
+- one commit for the functional changes: `3367fabdefc0b35d34bf7cf2fb0d33ff81f9fd5a <https://github.com/scality/metalk8s/commit/3367fabdefc0b35d34bf7cf2fb0d33ff81f9fd5a>`_
+
+Ideally, purely cosmetic changes which inflate the number of changes in a PR
+significantly, should go in a separate PR
+
+Refactoring
+:::::::::::
+
+When introducing new features, you often have to add new helpers or refactor
+existing code. In such case, instead of having single commit with everything
+inside, you can either:
+
+1. first add a new helper: `29f49cbe9dfa <https://github.com/scality/metalk8s/commit/29f49cbe9dfa0b824c818d25d4a2f6965351e65d>`_
+
+2. then use it in new code: `7e47310a8f20 <https://github.com/scality/metalk8s/commit/7e47310a8f20fd49f0ad36707b20e6c2a53df638>`_
+
+Or:
+
+1. first add the new code: `5b2a6d5fa498 <https://github.com/scality/metalk8s/commit/5b2a6d5fa49815180a2effdd37cb58542e83b5a5>`_
+
+2. then refactor the now duplicated code: `ac08d0f53a83 <https://github.com/scality/metalk8s/commit/ac08d0f53a835a0b2bc61c1fe5b7317bf4d6550c>`_
+
+Mixing unrelated changes
+::::::::::::::::::::::::
+
+It is sometimes tempting to do small unrelated changes as you are working on
+something else in the same code area.
+Please refrain to do so, or at least do it in a dedicated commit.
+
+Mixing non-related changes into the same commit makes revert and cherry-pick
+harder (and understanding as well).
+
+The pull request `#1846 <https://github.com/scality/metalk8s/pull/1846>`_ is a good example. It tackles three issues at once: `#1830 <https://github.com/scality/metalk8s/issues/1830>`_
+and `#1831 <https://github.com/scality/metalk8s/issues/1831>`_ (because they are similar) and `#839 <https://github.com/scality/metalk8s/issues/839>`_ (because it was making the other
+changes easier), but it uses distincts commits for each issue.
+
+How to write a commit message
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Why do we need commit messages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+After comments in the code, commit messages are the easiest way to find context
+for every single line of code: running ``git blame`` on a file will give you,
+for each line, the identifier of the last commit that changed the line.
+
+Unlike a comment in the code (which applies to a single line or file), a commit
+message applies to a logical change and thus can provide information on the
+design of the code and why the change was done. This makes commit messages a
+part of the code documentation and makes them helpful for other developers to
+understand your code.
+
+Last but not least: commit messages can also be used for automating tasks such
+as issue management.
+
+Note that it is important to have all the necessary information in the commit
+message, instead of having them (only) in the related issue, because:
+
+- the issue can contain troubleshooting/design discussion/investigation with a
+  lot of back and forth, which makes hard to get the gist of it.
+
+- you need access to an external service to get the whole context, which goes
+  against one of biggest advantage of the distributed SCM (having all the
+  information you need offline, from your local copy of the repository).
+
+- migration from one tracking system to another will invalidate the
+  references/links to the issues.
+
+Anatomy of a good commit message
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A commit is composed of a subject, a body and a footer. A blank line separates
+the subject from body and the body from the footer.
+
+The body can be omitted for trivial commit. That being said, be very careful:
+a change might seem trivial when you write it but will seem totally awkward
+the day you will have to understand why you made it. If you think your patch is
+trivial and somebody tells you he does not understand your patch, then your
+patch is not trivial and it requires a detailed description.
+
+The footer contains references for issue management (``Refs``, ``Closes``,
+etc.) or other relevant annotations (cherry-pick source, etc.).
+Optional if your commit is not related to any issue (should be pretty rare).
+
+Subject
+:::::::
+
+A good commit message should start with a short summary of the change: the
+subject line.
+
+This summary should be written using the imperative mood and carry as much
+information as possible while staying short, ideally under 50 characters (this
+is a goal, the hard limit is 72).
+
+Subject topic and description shouldn't start with a capital.
+
+It is composed of:
+
+- a topic, usually the name of the affected component (``ui``, ``build``,
+  ``docs``, etc.)
+
+- a slash and then the name of the sub-component (optional)
+
+- a colon
+
+- the description of the change
+
+Examples:
+
+- ``ci: use proxy-cache to reduce flakiness``
+
+- ``build/package: factorize task_dep in DEBPackage``
+
+- ``ui/volume: add banner when failed to create volume``
+
+If several components are affected:
+
+- split your commit (preferred)
+
+- pick only the most affected one
+
+- entirely omit the component (happen for truly global change, like renaming
+  ``licence`` to ``license`` over the whole codebase)
+
+As for "what is the topic?", the following heuristic works quite well for
+MetalK8s: take the name of the top-level directory (``ui``, ``salt``, ``docs``,
+etc.) except for ``eve`` (use ``ci`` instead). ``buildchain`` could also be
+shortened to ``build``.
+
+Having the topic in the summary line allows for faster peering over ``git log``
+output (you can know what the commit is about just by reading a few characters,
+not need to check the entire commit message or the associated diff).
+It also helps the review process: if you have a big pull request affecting
+front-end and back-end, front-end people can only review commits starting with
+``ui`` (not need to read over the whole diff, or to open each commit one by one
+in Github to see which ones are interesting).
+
+Body
+::::
+
+The body should answer the following questions:
+
+- Why did you make this change? (is this for a new feature, a bugfix - then,
+  why was it buggy? -, some cleanup, some optimization, etc.).
+  It is really important to describe the intent/motivation behind the changes.
+
+- What change did you make? Document what the original problem was and how it
+  is being fixed (can be omitted for short obvious patches).
+
+- Why did you make the change in that way and not in another (mention alternate
+  solutions considered but discarded, if any)?
+
+When writing your message you must consider that your reader does not know
+anything about the code you have patched.
+
+You should also describe any limitations of the current code. This will avoid
+reviewer pointing them out, and also inform future people looking at the code
+which tradeoffs were made at the time.
+
+Lines must be wrapped at 72 characters.
+
+Footer
+::::::
+
+Use `references <https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords>`_
+such as ``Refs``, ``See``, ``Fixes`` or ``Closes`` followed by
+an issue number to automate issue management.
+
+In addition to the references, you can also provide the URLs (it will be
+quicker to access them from the terminal).
+
+Example:
+
+.. code:: text
+
+    topic: description
+
+    [ commit message body ]
+
+    Refs: #XXXXX
+    Refs: #YYYYY
+    Closes: #ZZZZZ
+    See: https://github.com/scality/metalk8s/issues/XXXXX
+    See: https://github.com/scality/metalk8s/issues/YYYYY
+    See: https://github.com/scality/metalk8s/issues/ZZZZZ
+
+Footer can also contain a signature (``git commit -s``) or cherry-pick source
+(``git cherry-pick -x``).
+
+Examples
+^^^^^^^^
+
+Bad commit message
+::::::::::::::::::
+
+- ``Quick fix for service port issue``: what was the issue? It is a quick fix,
+  why not a proper fix? What are the limitations?
+
+- ``fix glitchs``: as expressive and useful as ~fix stuff~
+
+- ``Bump Create React App to v3 and add optional-chaining``: Why? What are the
+  benefits?
+
+- ``Add skopeo & m2crypto to packages list``: Why do we need them?
+
+- ``Split certificates bootstrap between CA and clients``: Why do we need this
+  split? What is the issue we are trying to solve here?
+
+Note that none of these commits contain a reference to an issue (which
+could have been used as an (invalid) excuse for the lack of information): you
+really have no more context/explanation than what is shown here.
+
+Good commit message
+:::::::::::::::::::
+
+Commit `b531290c04c4 <https://github.com/scality/metalk8s/commit/b531290c04c45a01cd103a85431e2428b98d340e>`_
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+.. code:: text
+
+    Add gzip to nginx conf
+
+    This will decrease the size of the file the client need to download
+    In the current version we have ~7x improvement.
+    From 3.17Mb to 0.470Mb send to the client
+
+Some things to note about this commit message:
+
+- Reason behind the changes are explained: we want to decrease the size of
+  the downloaded resources.
+
+- Results/effects are demonstrated: measurements are given.
+
+Commit `82d92836d4ff <https://github.com/scality/metalk8s/commit/82d92836d4ff78c623a0e06302c94cfa5ff79908>`_
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+.. code:: text
+
+    Use safer invocation of shell commands
+
+    Running commands with the "host" fixture provided by testinfra was done
+    without concern for quoting of arguments, and might be vulnerable to
+    injections / escaping issues.
+
+    Using a log-like formatting, i.e. `host.run('my-cmd %s %d', arg1, arg2)`
+    fixes the issue (note we cannot use a list of strings as with
+    `subprocess`).
+
+    Issue: GH-781
+
+Some things to note about this commit message:
+
+- Reasons behind the changes are explained: potential security issue.
+
+- Solution is described: we use log-like formatting.
+
+- Non-obvious parts are clarified: cannot use a list of string (as expected)
+  because it is not supported.
+
+Commit `f66ac0be1c19 <https://github.com/scality/metalk8s/commit/f66ac0be1c191be8fa31a925c28d34c113eb172c>`_
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+.. code:: text
+
+    build: fix concurrent build on MacOS
+
+    When trying to use the parallel execution feature of `doit` on Mac, we
+    observe that the worker processes are killed by the OS and only the
+    main one survives.
+
+    The issues seems related to the fact that:
+    - by default `doit` uses `fork` (through `multiprocessing`) to spawn its
+      workers
+    - since macOS 10.13 (High Sierra), Apple added a new security measure[1]
+      that kill processes that are using a dangerous mix of threads and
+      forks[2])
+
+    As a consequence, now instead of working most of the time (and failing
+    in a hard way to debug), the processes are directly killed.
+
+    There are three ways to solve this problems:
+    1. set the environment variable `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES.`
+    2. don't use `fork`
+    3. fix the code that uses a dangerous mix of thread and forks
+
+    (1) is not good as it doesn't fix the underlying issue: it only disable
+    the security and we're back to "works most of the time, sometimes does
+    weird things"
+    (2) is easy to do because we can tell to `doit` to uses only threads
+    instead of forks.
+    (3) is probably the best, but requires more troubleshooting/time/
+
+    In conclusion, this commit implements (2) until (3) is done (if ever) by
+    detecting macOS and forcing the use of threads in that case.
+
+    [1]: http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html
+    [2]: https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
+
+    Closes: #1354
+
+Some things to note about this commit message:
+
+- Observed problem is described: parallel builds crash on macOS.
+
+- Root cause is analyzed: OS security measure + thread/fork mix.
+
+- Several solution are proposed: disable the security, workaround the problem
+  or fix the root cause.
+
+- Selection of a solution is explained: we go for the workaround because it is
+  easy and faster.
+
+- Extra-references are given: links in the footer gives more in-depth
+  explanations/context.
+
+Conclusion
+~~~~~~~~~~
+
+When reviewing a change, do not simply look at the correctness of the code:
+review the commit message itself and request improvements to its content.
+Look out for commits that can be divided, ensure that cosmetic changes are not
+mixed with functional changes, etc.
+
+The goal here is to improve the long term maintainability, by a wide variety of
+developers who may only have the Git history to get some context so it is
+important to have a useful Git history.
--- a/docs/developer/best_practices/index.rst
+++ b/docs/developer/best_practices/index.rst
@@ -3,4 +3,5 @@ Development Best Practices
 
 .. toctree::
 
+   commit
    python


ycointe (dismissed):
LGTM !

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
+Commit best practices
+---------------------
+
+How to split a changes into commits
```
charlesprost:
```suggestion
How to split a change into commits
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+^^^^^^^^
+
+Cosmetic changes
+::::::::::::::::
+
+Use a dedicated commit when you want to make cosmetic changes to the code
+(linting, whitespaces, alignment, renaming, …).
+
+Mixing cosmetics and functional changes is bad because the cosmetics (which
+tends to generate a lot of diff/noise) will obscure the important functional
```
charlesprost:
```suggestion
tend to generate a lot of diff/noise) will obscure the important functional
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+as well.
+
+Mixing unrelated changes
+::::::::::::::::::::::::
+
+It is sometimes tempting to do small unrelated changes as you are working on
+something else in the same code area.
+Please refrain to do so, or at least do it in a dedicated commit.
+
+Mixing non-related changes into the same commit makes revert and cherry-pick
```
charlesprost, line 74: And understanding as well!

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+
+- you need access to an external service to get the whole context, which goes
+  against one of biggest advantage of the distributed SCM (having all the
+  information you need offline, from your local copy of the repository).
+
+Anatomy of a good commit message
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A commit is composed of a subject, a body and a footer. A blank line separates
+the subject from body.
```
charlesprost: And what separates the footer from the body?

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+    detecting macOS and forcing the use of threads in that case.
+
+    [1]: http://sealiesoftware.com/blog/archive/2017/6/5/Objective-C_and_fork_in_macOS_1013.html
+    [2]: https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
+
+    Closes: #1354
+
+Some things to note about this commit message:
+
+- Observed problem is described: parallel builds crash on MacOS
```
charlesprost:
```suggestion
- Observed problem is described: parallel builds crash on MacOS.
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+
+- Root cause is analyzed: OS security measure + thread/fork mix
```
charlesprost:
```suggestion
- Root cause is analyzed: OS security measure + thread/fork mix.
```

charlesprost (dismissed):

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: A blank line as well, I'll add it.

docs/developer/best_practices/commit.rst
slaperche-scality: I'll add it, good point :slightly_smiling_face:

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+part of the code documentation and makes them helpful for other developers to
+understand your code.
+
+Last but not least: commit messages can also be used for automating tasks such
+as issue management and changelog generation.
+
+Note that it is important to have all the necessary information in the commit
+message, instead of having them (only) in the related issue, because:
+
+- issues can contains troubleshooting/design discussion/investigation with a
```
rdebay-scality: "issue can contain" ?

slaperche-scality: Done.

rdebay-scality (dismissed):

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,357 @@
...
+
+When reviewing a change, do not simply look at the correctness of the code:
+review the commit message itself and request improvements to its content.
+Look out for commits that can be divided, ensure that cosmetic changes are not
+mixed with functional changes, …
+
+The goal here is to improve the long term maintainability, by a wide variety of
+developers. Considering that we are an Open Source project, external
+contributors may only have the Git history to get some context so it is
+important to have a useful Git history.
```
TeddyAndrieux: This doc should be used by open source contributors as well. IMO we should not talk about the fact that we are an "Open Source project" in the doc. 
I would suggest something 
```suggestion
The goal here is to improve the long term maintainability, by a wide variety of
developers who may only have the Git history to get some context so it is
important to have a useful Git history.
```

(Suggestion does not work very well with multi line comment :stuck_out_tongue: )

slaperche-scality: Fair enough. Done!

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
+Commit best practices
```
NicolasT: Shouldn't we use *titlecase*?

```suggestion
Commit Best Practices
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+---------------------
+
+How to split a change into commits
```
NicolasT, line 4: The *why* should come before the *how*.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The golden rule to create good commits is to ensure that there is only one
+logical change per commit.
```
NicolasT: I'd rephrase this, because *logical* is quite broad and non-precise.

IMO, the content of a commit should be `a reversible unit`, i.e., something which, when `git revert`ed, doesn't break the product. Related, but less important (and likely more related to what you describe as `logical unit`) is something which can be `git cherry-pick`ed.

NicolasT: Looks like you refer to this before. However, I still think *logical* is too imprecise.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+
+- make ``git log`` and ``git blame`` way more useful
+
+Examples
+^^^^^^^^
+
+Cosmetic changes
+::::::::::::::::
+
+Use a dedicated commit when you want to make cosmetic changes to the code
```
NicolasT, line 33: *unless these cosmetic changes change code that's introduced in the same PR*

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+(linting, whitespaces, alignment, renaming, …).
+
+Mixing cosmetics and functional changes is bad because the cosmetics (which
+tend to generate a lot of diff/noise) will obscure the important functional
+changes, making it harder to correctly determine whether the change is correct
+during the review.
```
NicolasT, line 39: *Ideally, purely cosmetic changes which inflate the number of changes in a PR significantly, should go in a separate PR*

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+- one commit for the functional changes: `3367fabdefc0b35d34bf7cf2fb0d33ff81f9fd5a <https://github.com/scality/metalk8s/commit/3367fabdefc0b35d34bf7cf2fb0d33ff81f9fd5a>`_
+
+Refactoring
+:::::::::::
+
+When introducing new features, you often have to add new helpers or refactor
+existing code. In such case, instead of having single commit with everything
+inside, you can either:
+
+1. first add a new helper: `29f49cbe9dfa <https://github.com/scality/metalk8s/commit/29f49cbe9dfa0b824c818d25d4a2f6965351e65d>`_
```
NicolasT, line 57: I disagree with this one and don't consider it a 'good' commit. It adds utility code which is not used anywhere.

This breaks the `git revert` rule.

Instead, the commit adding the utility function could've also updated code to *use* this utility, if any. If no existing code can use it, then introduce it *with* the 'feature' code that *will*, otherwise the `git revert`/`git cherry-pick` rule doesn't apply again.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+
+2. then use it in new code: `7e47310a8f20 <https://github.com/scality/metalk8s/commit/7e47310a8f20fd49f0ad36707b20e6c2a53df638>`_
+
+Or:
+
+1. first add the new code: `5b2a6d5fa498 <https://github.com/scality/metalk8s/commit/5b2a6d5fa49815180a2effdd37cb58542e83b5a5>`_
```
NicolasT, line 63: Again, I'm not sure doing this in 2 commits brings any value. It'd require *both* commits to be reverted and/or cherry-picked to 'work'.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+
+2. then refactor the now duplicated code: `ac08d0f53a83 <https://github.com/scality/metalk8s/commit/ac08d0f53a835a0b2bc61c1fe5b7317bf4d6550c>`_
+
+If you add tests (and you should), they can be introduced in a dedicated commit
```
NicolasT: No. This relates to `git bisect`. Say, the code is wrong. Then, the test fail. Now, if you `git bisect`, you'd *find* the commit introducing the *test*, not the one which contains the 'bad code'.

So, no. Code and tests for said code go in one commit.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+for each line, the identifier of the last commit that changed the line.
+
+Unlike a comment in the code (which applies to a single line or file), a commit
+message applies to a logical change and thus can provide information on the
+design of the code and why the change was done. This makes commit messages a
+part of the code documentation and makes them helpful for other developers to
+understand your code.
+
+Last but not least: commit messages can also be used for automating tasks such
+as issue management and changelog generation.
```
NicolasT: We don't want to use commit messages for changelog generation because they're immutable (among other reasons).

See https://docs.google.com/document/d/1NUUShY9VBdLa3s0ePR5swptBtszXwM9qvFsF52L22L4/edit?usp=sharing

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+harder (and understanding as well).
+
+The pull request `#1846 <https://github.com/scality/metalk8s/pull/1846>`_ is a good example. It tackles three issues at once: `#1830 <https://github.com/scality/metalk8s/issues/1830>`_
+and `#1831 <https://github.com/scality/metalk8s/issues/1831>`_ (because they are similar) and `#839 <https://github.com/scality/metalk8s/issues/839>`_ (because it was making the other
+changes easier), but it uses distincts commits for each issue.
+
+How to write a commit message
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Why do we need commit message
```
NicolasT:
```suggestion
Why do we need commit messages
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+Note that it is important to have all the necessary information in the commit
+message, instead of having them (only) in the related issue, because:
+
+- the issue can contain troubleshooting/design discussion/investigation with a
+  lot of back and forth, which makes hard to get the gist of it.
+
+- you need access to an external service to get the whole context, which goes
+  against one of biggest advantage of the distributed SCM (having all the
+  information you need offline, from your local copy of the repository).
+
```
NicolasT, line 109: Related to the above, switching tracking system may lose history. Switching VCS should not.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+It is composed of:
+
+- a topic, usually the name of the affected component (``ui``, ``build``,
+  ``docs``, …)
+
+- a slash and then the name of the sub-component (optional)
+
+- a colon
+
+- the description of the change
```
NicolasT, line 150: - no capital

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+
+- ``build/package: factorize task_dep in DEBPackage``
+
+- ``ui/volume: add banner when failed to create volume``
+
+If several components are affected:
+
+- split your commit (preferred)
+
+- pick only the most affected one
```
NicolasT, line 164: or comma-separate?

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+
+Example:
+
+.. code:: text
+
+    topic: description
+
+    [… commit message body …]
+
+    Refs: #XXXXX, #YYYYY
```
NicolasT: Prefer one per line.
Also, I tend to do provide not only `See: #XXXX`, but also `See: https://github.com/scality/metalk8s/issues/XXXX` or whatever, such that one can easily click-to-open in a terminal.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,356 @@
...
+Commit `b531290c04c4 <https://github.com/scality/metalk8s/commit/b531290c04c45a01cd103a85431e2428b98d340e>`_
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+.. code:: text
+
+    Add gzip to nginx conf
+
+    This will decrease the size of the file the client need to download
+    In the current version we have ~7x improvement.
+    From 3.17Mb to 0.470Mb send to the client
```
NicolasT, line 269: And this is exactly why we can't generate changelogs from the commits ;-)

TeddyAndrieux (dismissed):
LGTM

NicolasT (dismissed):

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: True for the revert/cherry-pick part.

But it does make the review easier: you can review the feature and the refactor separately, while conflating both may be harder to review.

docs/developer/best_practices/commit.rst
slaperche-scality: If it's code introduced in this PR, then it's new code and you will write it "correctly" when you introduce it.

Or do you talk about a commit that changes some existing code, and thus reformat it as it modifies it?
Yeah, in this case you won't do two commits (one which update the code but keep the bad formatting and then one to fix the formatting), you'll edit the code and fix the formatting directly).

docs/developer/best_practices/commit.rst
slaperche-scality: Indeed, I'll add it.

docs/developer/best_practices/commit.rst
slaperche-scality: The function to check if an image exists is not tied to the `uptodate` one.

I mean, yes `uptodate` does need this helper (and thus if you want to cherry-pick this "feature" you'll have to cherry-pick the helper as well), but the helper could also be used in another independent context (pre-check that some images exists, …) and thus could be cherry-picked independently.
Likewise, you may want to revert the `uptodate` stuff while keeping the helper around for another use.

docs/developer/best_practices/commit.rst
slaperche-scality: Hum, good point indeed!

I was thinking  how we sometimes wrote test today (especially E2E one):  write a huge chunk of code conflating required steps and the test scenario.
Adding this to the feature would make for a huge commit.

But I guess if we wrote the steps and everything needed for the tests in prior commits, and then have a commit with the feature AND only the test scenario, that would be doable.

I'll remove this sentence then :slightly_smiling_face: 

(now, to be honest, I doubt that `bisect` works well on a codebase where the CI doesn't enforce that the code at least "build" for each commit :stuck_out_tongue: )

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: Done (I think)

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah, @gdemonet shared this document with me afterward.
That being said, in general it's a possible use case of commit messages.

But yeah, since it's doesn't fit our needs + those guideline are specific to this project, mentioning it here is irrelevant, so I'll remove the changelog example.

docs/developer/best_practices/commit.rst
slaperche-scality: It's an option, but it will nibble at the already tight character budget of the subject line.

So I'm not sure it's a good one :/

docs/developer/best_practices/commit.rst
slaperche-scality: Ok for one per line, I'll update the example.

Good idea for also providing a real hyperlink!

docs/developer/best_practices/commit.rst
slaperche-scality: Indeed ^^

docs/developer/best_practices/commit.rst
slaperche-scality: :+1: 

And since subject line are more like a title, they shouldn't end with a period (I'll remove that part).

docs/developer/best_practices/commit.rst
slaperche-scality: Yup, I forgot to mention that one. I'll add it.

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah, `logical` is kinda unclear I agree.

I'll rewrite this part using a definition more similar to your yours.

ChengYanJin (dismissed):

sayf-eddine-scality (dismissed):

docs/developer/best_practices/commit.rst
carlito-scality: Personally, it is easier for me to review if they are in one commit since I would like know why is it refactored and how is it used

slaperche-scality: >  I would like know why is it refactored and how is it used

But you'll know why it is refactored (because the previous commit implementing the feature introduced code duplication) and you'll see how it's used.

Having two commit also have the following advantage:
- first you implement the new feature, testing it manually/adding unit test
- then you refactor and you know that it works because you have the unit test/manual test from the previous commit.

Now, and maybe that's not clear here, the goal of those examples is to help people who have no idea how they could split things.
These aren't hard and fast rules to be applied in all cases, I trust people to follow their judgement here.

jbertran (dismissed):

docs/developer/best_practices/commit.rst
slaperche-scality: Now, if this example is really controversial/confusing I can remove it.

It's no big deal, given that:
- people usually split their changes into commit already (most of the time at least), our current biggest problem are the commit messages…
- those examples are only a guidance at best, not something to strictly enforce.

Opinion @NicolasT , @carlito-scality ?

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
+Commit Best Practices
+---------------------
+
+How to split a change into commits
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Why do we need to split changes into commits
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This has several advantages amongst whose:
```
Ebaneck: This has several advantages amongst which are: ??
or
This has several advantages including: ??

alexandre-allard (dismissed):

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+Unlike a comment in the code (which applies to a single line or file), a commit
+message applies to a logical change and thus can provide information on the
+design of the code and why the change was done. This makes commit messages a
+part of the code documentation and makes them helpful for other developers to
+understand your code.
+
+Last but not least: commit messages can also be used for automating tasks such
+as issue management.
+
+Note that it is important to have all the necessary information in the commit
```
Ebaneck, line 100: We could wrap the `note` in an `rst` note format since it is information you want the readers to retain always

For example ```..Note: ```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+    [… commit message body …]
+
+    Refs: #XXXXX
+    Refs: #YYYYY
+    Closes: #ZZZZZ
+    See: https://github.com/scality/metalk8s/issues/XXXXX
+    See: https://github.com/scality/metalk8s/issues/YYYYY
+    See: https://github.com/scality/metalk8s/issues/ZZZZZ
+
+Footer can also contains signature (``git commit -s``) or cherry-pick source
```
Ebaneck: Footer can also contain a signature??

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+
+- ``Bump Create React App to v3 and add optional-chaining``: Why? What are the
+  benefits?
+
+- ``Add skopeo & m2crypto to packages list``: Why do we need them?
+
+- ``Split certificates bootstrap between CA and clients``: Why do we need this
+  split? What is the issue we are trying to solve here?
+
+Note that none of these commits contain a reference to an issue (which
```
Ebaneck, line 253: Ditto, maybe an rst note format here??

Ebaneck (changes requested):
LGTM... just a couple of minors while reading through

docs/developer/best_practices/commit.rst
slaperche-scality: `Note that` here is more an expression than really introducing a note (I could totally remove it and just start the sentence with `It is important…`), si I'll keep it like that.

docs/developer/best_practices/commit.rst
slaperche-scality: Ditto

docs/developer/best_practices/commit.rst
slaperche-scality: Ditto

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah they could.
Most of my commits end with `Signed-off-by: Sylvain Laperche <…> `

And actually, it's even required for contributors outside of Scality: https://github.com/scality/metalk8s/blame/development/2.4/CONTRIBUTING.md#L110-L121

docs/developer/best_practices/commit.rst
slaperche-scality: I'll use `This has several advantages amongst which are:`, thanks!

carlito-scality (dismissed):

Ebaneck (dismissed):

docs/developer/best_practices/commit.rst
Ebaneck: Yes yes...
Here, it was more of an issue with
`s/contains/contain a/`

slaperche-scality: Oh right. Done.

MonPote (dismissed):

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+
+- make ``git log`` and ``git blame`` way more useful
+
+Examples
+^^^^^^^^
+
+The golden rule to create good commits is to ensure that the commit is atomic.
+A commit is atomic when:
+
+- it can be reverted without breaking the product
```
gdemonet: Maybe that's a property we *want*, but it's definitely harder to achieve than having the ability to `checkout` any commit and have working code.

To me, `revert` ing a commit is a dangerous operation, and will always be: you never know what code was introduced after this commit, and if it relies on the contents you're removing. We are not reverting code often, and when we do, it's often one of two cases:
- disabling a commit in forward-merges (e.g. some implementation only needed for K8s versions before 1.15)
- removing a commit that works around a limitation of a dependency of our product, that we want to update

IMHO, `revert` is rarely used to remove an error unvoluntarily introduced. We have tests and reviews to prevent bugs from landing in the code, and shouldn't optimize our workflows to "patch" our mistakes after the act.

So, atomicity may be defined in terms of this ability to `revert`, but I'd rather focus on having the ability to **use** the code in any given point of history rather than being able to remove parts of it (not that usual of an operation). Furthermore, when `bisect`ing, ability to `revert` is irrelevant: the most important is that the code should not be broken at all, and we're trying to find one commit that is actually breaking something (maybe due to some history manipulation, or considering another, newer, view of _what is a correct state_).

slaperche-scality: > use the code in any given point of history 

This seems pretty hard to achieve, because most of the code we add depends on already existing code, no?
So if some required code is renamed or moved, then past that point you won't be able to "use it" as-is.

---

Side note: I don't think we will ever come up with a precise definition for an atomic (reversible, logical, you name it) commit that'll stand a reality check (and that's why I didn't try to define it in my first version :stuck_out_tongue: ).

I think it's intrinsically ambiguous and everyone will see it a bit different.
Is it a big deal?
I'm not sure. Most of our current problem are regarding commit message, split is mostly OK.

/me currently wondering if he should go back to an intendedly open "definition"

slaperche-scality: For the record, the concept of "logical commit" (without more definition) is coming from [here](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes), which was one of the three main sources that have influenced me.

gdemonet: > > use the code in any given point of history

> This seems pretty hard to achieve, because most of the code we add depends on already existing code, no?
> So if some required code is renamed or moved, then past that point you won't be able to "use it" as-is.

That's not what I meant. My point is, any **merged** commit should work as is, i.e. for any commit in `development/x.y`, I can ` git checkout` said commit and have working code.

This implies:
- reordering is a dangerous operation
- `cherry-pick`ing and `revert`ing are dangerous as well

---

I like the "open" definition from your link, that would suit me.

docs/developer/best_practices/commit.rst
gdemonet: I'd even argue those *topics* in a commit subject are already expensive to use: e.g. `examples/solution: ` costs 19 out of 50 characters, which makes meaningful subjects harder to write.

~I'm wondering what the use-case of *topics* really is. Are we trying to have easy to understand lists of commits, from `git log --oneline` and the likes? Or to have easy to filter lists of commits? If the latter, then having such information in the subject is unnecessary.~
OK, just read through and found the explanation.

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+  why was it buggy? —, some cleanup, some optimization, …).
+  It is really important to describe the intent/motivation behind the changes.
+
+- What change did you make? Document what the original problem was and how it
+  is being fixed (can be omitted for short obvious patches).
+
+- Why did you make the change in that way and not in another (mention alternate
+  solutions considered but discarded, if any)?
+
+When writing your message you must consider that your reader does not know
```
gdemonet, line 197: I'd somewhat reduce this statement: it's almost impossible to write _"noob-friendly"_ commits while carrying out meaningful information for e.g. knowledgeable reviewers, without writing full pages for each change.

Maybe we could go for something closer to "Try to write your message in an explicit manner, adding references when concepts would need explanation. If the reader comes from outside of the core developers, she needs to be able to navigate through to understand the change".

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+anything about the code you have patched.
+
+You should also describe any limitations of the current code. This will avoid
+reviewer pointing them out, and also inform future people looking at the code
+which tradeoffs were made at the time.
+
+Lines should be wrapped at 72 characters.
```
gdemonet:
```suggestion
Lines must be wrapped at 72 characters.
```

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+  lot of back and forth, which makes hard to get the gist of it.
+
+- you need access to an external service to get the whole context, which goes
+  against one of biggest advantage of the distributed SCM (having all the
+  information you need offline, from your local copy of the repository).
+
+- migration from one tracking system to another will invalidate the
+  references/links to the issues.
+
+Anatomy of a good commit message
```
gdemonet, line 113: Can we provide a template? Users could "install" it locally, with `git config commit.template` :)

docs/developer/best_practices/commit.rst
```diff
@@ -0,0 +1,372 @@
...
+reviewer pointing them out, and also inform future people looking at the code
+which tradeoffs were made at the time.
+
+Lines should be wrapped at 72 characters.
+
+Footer
+::::::
+
+Use references such as ``Refs``, ``See``, ``Fixes`` or ``Closes`` followed by
+an issue number to automate issue management.
```
gdemonet, line 211: Such references can have a special meaning, let's mention it here: https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

cmonfort (dismissed):

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah, that's the how they "automate issue management", but I'll add the link, you're right.

docs/developer/best_practices/commit.rst
slaperche-scality: I don't have one ready right now.

Also, I'm worried it would be more of a hindrance than helpful since you'll always have text to delete (because you can hardly generate a useful default).

docs/developer/best_practices/commit.rst
slaperche-scality: Done

docs/developer/best_practices/commit.rst
slaperche-scality: It's not about noob-friendliness, as in "explain acronyms such as IPIP" or "what is a Pod?".
It's more about anyone should be able to review the code, even if it's not in their comfort zone.

That kinda goes hand in hand with the bullet points above: give context and explain why such as an outsider understand the commit.

Now, if we have a more clear way than "doesn't know anything about the code patched", I can reformulate :slightly_smiling_face:

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah, topic can be quite a mouthful as well, but in you example I would replace `examples/solution` by `solution` since it stays unambiguous.

The "heuristic" I give (take the name of the top-level directory) is not perfect, it's more like an hint that works is most cases, but is not always optimal.

docs/developer/best_practices/commit.rst
slaperche-scality: > That's not what I meant. My point is, any merged commit should work as is, i.e. for any commit in development/x.y, I can git checkout said commit and have working code.

Ok, I agree with that.

> I like the "open" definition from your link, that would suit me.

Then I'll revert back to my "open" denomination. And if the need for a more precise definition occurs, we can then edit this document (it's not set in stone).

docs/developer/best_practices/commit.rst
gdemonet: There's a slight difference between `examples/solution` and `salt/solutions` though. But yes, let's do our best :)

docs/developer/best_practices/commit.rst
gdemonet: We can use comments for non-required content, I guess.

Something like:

```
topic: subject (stay below 50 chars)

# Commit message body
# Explains:
#   - why the change,
#   - what was changed,
#   - discarded alternatives

# Commit message footer
# References:
#   - issues fixed by this change, using e.g. `Fixes: #XXXXX`
#   - issues linked to this change, using e.g. `Refs: #XXXXX`
#   - external links, using e.g. `See: kubernetes/kubernetes/issues/XXXXX`
```

gdemonet (dismissed):

docs/developer/best_practices/commit.rst
slaperche-scality: Yeah, we could interpret `solutions` as the `solutions` itself, and `salt/solutions` as the Salt code to deploy the solution.
But yeah, no hard rules (that's why I haven't listed a closed-set of accepted topic: the goal is mainly informative).

docs/developer/best_practices/commit.rst
slaperche-scality: Where would that template lives?
- Inline in this doc?
- Or somewhere in the repo?

(still not really convinced of the help of this template though ^^)

gdemonet (approved):

PR #1929, opened 2019-10-17
Do not build empty repository

### LINKED ISSUE #1911: Error when building empty debian repo
**Component**:

'build'

**What happened**:

When building empty repo `reprepro` fail
```
+ reprepro -C metalk8s-bionic-security -b /repository/ includedeb bionic '/packages/*.deb'
Error 2 opening /packages/*.deb: No such file or directory
There have been errors!
```

**What was expected**:

A working build, even with empty repositories (those should be skipped).

Use [delayed task creation](https://pydoit.org/task_creation.html?#delayed-task-creation) to postpone the creation of the `build_repositories` tasks **after** the download of the packages.
That way, we can check if there is packages for the repository, and if not we simply skip it.

slaperche-scality:
**Component**:

build

**Context**: 

We sometimes end up trying to build empty repository on Ubuntu, which is bad (`reprepro` doesn't work, and anyway it doesn't make sense).

**Summary**:

Use delayed task creation to postpone the creation of the "build repository" tasks after the "download packages" task: that way we can check if a repository is empty or not (and if yes, we can skip it).

Also had to fix the naming scheme of the nested tasks (was subtly broken).

**Acceptance criteria**: 

Run `./doit.sh -n4`, at the end `_build/root/packages/debian/metalk8s-bionic-security` should not exist.

---

Closes: #1911


buildchain/buildchain/targets/repository.py
```diff
@@ -112,12 +109,23 @@ def build_repo(self) -> types.TaskDict:
     def build_packages(self) -> List[types.TaskDict]:
         """Build the packages for the repository."""
 
+    def _get_task_name(self, taskname: str, with_basename: bool=False) -> str:
+        """Return a fully qualified task name.
+
+        The task name is prefixed by the repository name.
+        Use the given basename if any.
+        """
+        suffix = '{}:'.format(self.basename) if with_basename else ''
```
gdemonet: Wouldn't that be a `prefix`?

gdemonet (dismissed):
LGTM :) Just a small nitpick on variable naming

buildchain/buildchain/targets/repository.py
slaperche-scality: Raaah, indeed!
I'll fix it :grin:

gdemonet (approved):

PR #1930, opened 2019-10-17
Solution UI: fix components list refresh management

### LINKED ISSUE #1820: SolutionUI: Component StopRefresh is not cancelled correctly
**Component**:Solution UI

**What happened**:

- When we switch from ```environment 1``` to ```environment 2```, there are 2 refresh of clockServer running instead of one in the ```environment 2``` since the refresh of the ```environment 1``` is not cancelled correctly. The root cause is the redux state "isRefreshing" in ```examples/metalk8s-solution-example/ui/src/ducks/app/clockServer.js``` is shared for all environments. It should be part of one single environment
- same bug for ```versionServer```.

**What was expected**:

**Steps to reproduce**

carlito-scality:
**Component**:Solution UI

**Context**: 
- When we switch from environment 1 to environment 2, there are 2 refresh of clockServer running instead of one in the environment 2 since the refresh of the environment 1 is not cancelled correctly. The root cause is the redux state "isRefreshing" in examples/metalk8s-solution-example/ui/src/ducks/app/clockServer.js is shared for all environments. It should be part of one single environment

**Summary**:

**Acceptance criteria**: 
- No regression
- When switching from one environment to another one, the component refreshs are correctly managed

---

Closes: https://github.com/scality/metalk8s/issues/1820


gdemonet (approved):

carlito-scality:
@bert-e approve

PR #1933, opened 2019-10-17
tests/logs: Fix name of helper method

gdemonet:
The helper method `wait_for_pod`, from `tests.kube_utils`, was renamed
in 5b82a73, but not taken into account by ce6ff39.


TeddyAndrieux (approved):
Good catch !

Ebaneck (approved):

PR #1944, opened 2019-10-22
doc: Document how to administrate metalk8s services

### LINKED ISSUE #1940: Explain how to change passwd to access K8S, Salt APIs and Grafana
**Component**:

documentation

**Why this is needed**:

Currently, we pre-provision user admin with password admin with all privileges !!! As a consequence, anyone could access the UI (or APIs) and trigger dangerous MetalK8s operations. This behaviour is not secure enough and it is very likely that the customer will ask for defining his own admin password for both APIs and Grafana

**What should be done**:

We should document in official MatelK8s documentation how to change the password for APIs and Grafana

**Test plan**:

Ebaneck:
**Component**:

'Doc'

**Context**: 

see #1940 

**Summary**:

- Add procedures on how to manage default user account for Grafana and Metalk8s UI

**Acceptance criteria**: 

---

Closes: #1940

### DIFF
--- a/docs/operation/account_administration.rst
+++ b/docs/operation/account_administration.rst
@@ -1,0 +1,88 @@
+
+Account Administration
+======================
+
+This section highlights **MetalK8s Account Administration** which covers
+changing the default username and password for some MetalK8s services.
+
+
+Administering Grafana
+*********************
+
+A fresh install of MetalK8s has a Grafana service instance with default
+credentials: ``admin`` / ``admin``. For more information on how to access
+Grafana, please refer to :ref:`this procedure <quickstart-services-grafana>`
+
+Changing Grafana username and password
+--------------------------------------
+
+To change the default username and password for Grafana on a MetalK8s cluster,
+perform the following procedures:
+
+#. Create a file named ``patch-secret.yaml`` that has the following content:
+
+   .. code-block:: yaml
+
+      stringData:
+        admin-password: <password-in-clear>
+        admin-user: <username-in-clear>
+
+#. Apply the patch file by running:
+
+   .. code-block:: shell
+
+      $ kubectl --kubeconfig /etc/kubernetes/admin.conf patch secrets prometheus-operator-grafana --patch "$(cat patch-secret.yaml)" -n metalk8s-monitoring
+
+#. Now, roll out the new updates for Grafana.
+
+   .. code-block:: shell
+
+      $ kubectl --kubeconfig /etc/kubernetes/admin.conf rollout restart deploy prometheus-operator-grafana -n metalk8s-monitoring
+
+#. Access the Grafana instance and authenticate yourself using the new Account
+   credentials.
+
+
+Administering MetalK8s GUI, Kubernetes(K8S) and Salt API
+********************************************************
+
+A fresh install of MetalK8s has a GUI instance with default credentials:
+``admin`` / ``admin``. For more information on how to access the GUI, please
+refer to :ref:`this procedure <quickstart-services-admin-ui>`
+
+Managing MetalK8s GUI, K8S and Salt API username and password
+-------------------------------------------------------------
+
+  .. warning::
+
+     #. In a typical **Metalk8s** cluster, ``/etc/kubernetes/htpasswd``
+        provides access control credentials to the following 3 components:
+        **MetalK8s GUI**, **K8S** and **Salt API**. Any changes made to this
+        file will directly impact authentication for the 3 mentioned components.
+
+     #. The procedures mentioned below must be carried out on every cluster
+        node having an instance of the Kubernetes API server.
+
+#. From a node running an instance of the Kubernetes API server,
+   edit the file ``/etc/kubernetes/htpasswd`` replacing the
+   username and/or password fields as below.
+
+   .. code-block:: shell
+
+      <username-in-clear>,<password-in-clear>,123,"system:masters"
+
+#. From a node running an instance of the Kubernetes API server,
+   force a restart of the Kubernetes API server.
+
+   .. code-block:: shell
+
+      $ crictl stop $(crictl ps -q --label io.kubernetes.pod.namespace=kube-system --label io.kubernetes.container.name=kube-apiserver --state Running)
+
+#. Access a service (for example, MetalK8s GUI) and authenticate yourself
+   using the new Account credentials.
+
+   .. note::
+
+      Upon changing the username and/or password, a fresh logout then login is
+      required for accessing the MetalK8s GUI.
+
--- a/docs/operation/index.rst
+++ b/docs/operation/index.rst
@@ -16,3 +16,4 @@ do not have a working MetalK8s_ setup.
    downgrade
    changing_node_hostname
    volume_management/index
+   account_administration


docs/operation/account_administration.rst
```diff
@@ -0,0 +1,86 @@
...
+      $ echo -n 'new-password' | base64
+
+#. Create a file named ``patch-secret.yaml`` that has following content:
+
+   .. code-block:: yaml
+
+      data:
+        admin-password: bmV3cGFzc3dvcmQ=   <-- Replace with base64 password
+        admin-user: bmV3cGFzc3dvcmQ=       <-- Replace with base64 username
+        ldap-toml: ""
```
gdemonet: Let's remove the `base64` encoding for the user, since API Server supports doing it server-side. Use this for the patch file instead:

```
stringData:
  admin-user: <username-in-clear>
  admin-password: <password-in-clear>
```

(Note that we don't need to patch `ldap-toml`)

Also, could be nice mentioning that either one of the two keys can be omitted from this patch, depending on what the user wants to achieve.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,86 @@
...
+
+   .. code-block:: shell
+
+      $ kubectl patch secrets prometheus-operator-grafana --patch "$(cat patch-secret.yml)" -n metalk8s-monitoring
+
+#. Verify that the secret was updated successfully.
+
+   .. code-block:: shell
+
+      $ kubectl get secrets prometheus-operator-grafana --output yaml -n metalk8s-monitoring
```
gdemonet: Could be omitted, since the output will be base64-encoded

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,86 @@
...
+
+#. Now, roll out the new updates for the grafana deployment.
```
gdemonet:
```suggestion
#. Now, roll out the new updates for Grafana.
```

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,86 @@
...
+   .. code-block:: shell
+
+      $ kubectl rollout restart deploy prometheus-operator-grafana -n metalk8s-monitoring
+
+#. Access the Grafana instance and authenticate yourself using the new Account
+   credentials.
+
+
+
+Administering MetalK8s GUI
```
gdemonet: This section needs rewording: we are actually changing the Kubernetes API user/password here. Which implies not only changes to the GUI, but also to SaltAPI credentials (we could explain these implications in this document, sure).

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,86 @@
...
+-------------------------------------------
+
+#. From the Bootstrap node, edit the file ``/etc/kubernetes/htpasswd``
+   replacing the username and password fields as below.
+
+   .. code-block:: shell
+
+      new_username,new_password,123,"system:masters"
+
+#. From the Bootstrap node, re-configure the Kubernetes API server.
```
gdemonet:
```suggestion
#. From the Bootstrap node, force a restart of the Kubernetes API server.
```

gdemonet: Now that I think of it, running it on the Bootstrap node only won't be sufficient if there are multiple `master` nodes.

gdemonet (changes requested):
Early feedback, we need to do some extra checks to:

- validate authentication changes for SaltAPI
- adapt the procedure in case of multiple `master` Nodes

We'll open a tracking issue for having this file-based declaration of user/passwords managed with Salt.

Ebaneck:
> Early feedback, we need to do some extra checks to:
> 
>     * validate authentication changes for SaltAPI
> 
I think salt-API authentication was moved from basic_auth to certificate-based authentication on 2.4.
no??
>     * adapt the procedure in case of multiple `master` Nodes
> 
> 
> We'll open a tracking issue for having this file-based declaration of user/passwords managed with Salt.

docs/operation/account_administration.rst
Ebaneck: Looks like SaltAPI authentication was moved to certificate based auth in the `2.4` branch so, i doubt this has an impact.

Ebaneck: https://github.com/scality/metalk8s/pull/1944#issuecomment-544963797 ??

docs/operation/account_administration.rst
Ebaneck: Yes, I agree!
As of now, there isn't any better approach. Maybe we wait for the Dex integration since it is something planned in the roadmap for the next iteration.

gdemonet:
> > * validate authentication changes for SaltAPI
> 
> I think salt-API authentication was moved from basic_auth to certificate-based authentication on 2.4.
> no??

So, from the current code in `salt/_auth/`, it seems SaltAPI is always using token-based authentication with `external_auth` modules. These tokens are mapped to different handlers, which allows to support both `Basic` (basic auth) and `Bearer` token types. `Bearer` tokens are for `ServiceAccount`s, such as the Storage Operator. `Basic` tokens are checked against K8s API.

docs/operation/account_administration.rst
gdemonet: Well, we could do something using Salt I guess. For now, let's just add a note mentioning this issue (procedure to repeat on each node hosting an API server instance).

docs/operation/account_administration.rst
carlito-scality: I don't think the UI still work when you update the login/password while playing with it. There will be probably some work to do on the UI to force user to log in again by going to the '/login'

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+
+      stringData:
+        admin-password: <username-in-clear>
+        admin-user: <password-in-clear>
+
+#. Apply the patch file by running:
+
+   .. code-block:: shell
+
+      $ kubectl patch secrets prometheus-operator-grafana --patch "$(cat patch-secret.yml)" -n metalk8s-monitoring
```
TeddyAndrieux: Should we add the kubeconfig directly in each and every command line that use kubectl ?
 `--kubeconfig=/etc/kubernetes/admin.conf`

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+      <username-in-clear>,<password-in-clear>,123,"system:masters"
+
+#. From a node running an instance of the Kubernetes API server,
+   force a restart of the Kubernetes API server.
+
+   .. code-block:: shell
+
+      $ crictl stop $(crictl ps -q --label io.kubernetes.pod.namespace=kube-system --label io.kubernetes.container.name=kube-apiserver --state Running)
+
+#. Access the service for example; MetalK8s GUI and authenticate yourself
```
gdemonet:
```suggestion
#. Access a service (for example, MetalK8s GUI) and authenticate yourself
```

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+   using the new Account credentials.
+
+   .. note::
+
+      Upon changing the username and/or password, a fresh logout then login is
+      required for continuous access to the MetalK8s GUI.
```
gdemonet:
```suggestion
      required for accessing the MetalK8s GUI.
```

gdemonet (dismissed):
Minor wording suggestions, can be ignored IMO

gdemonet (approved):
LGTM

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+     #. The procedures mentioned below must be carried out on every cluster
+        node having an instance of the Kubernetes API server.
+
+#. From a node running an instance of the Kubernetes API server,
+   edit the file ``/etc/kubernetes/htpasswd`` replacing the
+   username and/or password fields as below.
+
+   .. code-block:: shell
+
+      <username-in-clear>,<password-in-clear>,123,"system:masters"
```
thomasdanan, line 72: @Ebaneck @TeddyAndrieux  Maybe I did something wrong but to make it work, first param has to be password and second is the user.
@carlito-scality actually I am guessing it is a bug on the UI side no?

TeddyAndrieux: Hmm yes I think it's an issue on the UI side then

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+
+To change the default username and password for Grafana on a MetalK8s cluster,
+perform the following procedures:
+
+#. Create a file named ``patch-secret.yaml`` that has the following content:
+
+   .. code-block:: yaml
+
+      stringData:
+        admin-password: <password-in-clear>
```
NicolasT, line 27: I'd switch these 2 lines around. People think 'username', 'password'. Too easy to now use the wrong value in the wrong place now.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   .. code-block:: shell
+
+      $ kubectl --kubeconfig /etc/kubernetes/admin.conf rollout restart deploy prometheus-operator-grafana -n metalk8s-monitoring
+
+#. Access the Grafana instance and authenticate yourself using the new Account
+   credentials.
+
+
+Administering MetalK8s GUI, Kubernetes(K8S) and Salt API
```
NicolasT, line 46:
```suggestion
Administering MetalK8s GUI, Kubernetes and Salt API
```

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+********************************************************
+
+A fresh install of MetalK8s has a GUI instance with default credentials:
+``admin`` / ``admin``. For more information on how to access the GUI, please
+refer to :ref:`this procedure <quickstart-services-admin-ui>`
+
+Managing MetalK8s GUI, K8S and Salt API username and password
```
NicolasT, line 53: The way it's written could make one assume the UI *has* its own username/password, which is not true. Similarly, Salt doesn't have a username/password, but uses K8s as an auth backend.

So, where the current writing makes one believe you're talking about three distinct username/password combinations, there's really only one.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+-------------------------------------------------------------
+
+  .. warning::
+
+     #. In a typical **Metalk8s** cluster, ``/etc/kubernetes/htpasswd``
+        provides access control credentials to the following 3 components:
+        **MetalK8s GUI**, **K8S** and **Salt API**. Any changes made to this
```
NicolasT, line 60: See comment above. This can be confusing, because

- what's a non-typical cluster and how would thing work there?
- That file only contains access credentials to the K8s API. UI doesn't have any creds/access control itself, and SaltAPI federates access to the K8s API.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+        file will directly impact authentication for the 3 mentioned components.
+
+     #. The procedures mentioned below must be carried out on every cluster
```
NicolasT, line 63:
```suggestion
     #. The procedures mentioned below must be carried out on every control-plane node in cluster
```

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+        node having an instance of the Kubernetes API server.
+
+#. From a node running an instance of the Kubernetes API server,
```
NicolasT, line 66:
```suggestion
#. On every control-plane node,
```

docs/operation/account_administration.rst
NicolasT: Why would that be 'an issue on the UI side'? The UI never accesses this file.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+   edit the file ``/etc/kubernetes/htpasswd`` replacing the
+   username and/or password fields as below.
+
+   .. code-block:: shell
+
+      <username-in-clear>,<password-in-clear>,123,"system:masters"
+
+#. From a node running an instance of the Kubernetes API server,
```
NicolasT, line 74: Not *a* node, the node on which the file was changed, and so on for every node.

docs/operation/account_administration.rst
```diff
@@ -0,0 +1,88 @@
...
+   force a restart of the Kubernetes API server.
+
+   .. code-block:: shell
+
+      $ crictl stop $(crictl ps -q --label io.kubernetes.pod.namespace=kube-system --label io.kubernetes.container.name=kube-apiserver --state Running)
+
+#. Access a service (for example, MetalK8s GUI) and authenticate yourself
```
NicolasT, line 81: Mixing use of `MetalK8s UI` and `MetalK8s GUI` throughout the document.

NicolasT:
Hmh, this is in the queue already :-/

docs/operation/account_administration.rst
thomasdanan: I am guessing when you enter something in username field in the UI, it is used as a password and what you put in password field is used as a username

TeddyAndrieux: Yes right it's the password for kubernetes basic auth, so why this one is `<password>:<user>` when auth in the UI :confused:

ChengYanJin: I checked the UI code of login. I don't think we inversed the username and password there :/

NicolasT: Well, the doc is wrong: the format must be

```
password,user,uid,"group1,group2,group3"
```

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-password-file

So, if you'd put `username,password,123,"system:masters"` in the file, the declaration is wrong, and the UI would appear to be 'wrong' whilst it's right.

gdemonet: Oh OK

PR #1946, opened 2019-10-23
Refactor Eve build description

gdemonet:
**Component**: ci

**Context**: The existing `eve/main.yml` was littered with repetitions, making adjustments to certain steps cumbersome.

**Summary**:

- Refactor of the whole file using YAML anchors, references, and extends (see [this blog post](https://blog.daemonl.com/2016/02/yaml.html) for more info)
- Fine-tune timeout durations for artifacts retrieval
- Add some retry logic for retrieving large artifacts (sometimes, a `curl` request will fail with an empty response)

**Acceptance criteria**: 

- Builds still pass
- Less flakiness on artifacts retrieval

---

**Note**: this is a temporary refactoring - we plan on writing a custom "DSL" for generating this file at build-time, instead of these complex YAML tricks.

**To do**:

- [x] Rebase on `development/2.0` and adjust upgrade/downgrade specificities in according branches

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -111,6 +111,8 @@ models:
       command: >
         curl -s -XGET -o "${DEST_DIR}/metalk8s.iso" "${BASE_URL}/metalk8s.iso"
       haltOnFailure: true
+      # Increase default timeout for ISOs, as artifacts may be too slow
+      timeout: 2400
   - ShellCommand: &retrieve_iso_checksum
       <<: *retrieve_iso
       name: Retrieve ISO image checksum
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -103,23 +103,36 @@ models:
         DEST_DIR: sosreport/sosreport
       command: mkdir -p "${DEST_DIR}" && cp /var/tmp/sosreport* "${DEST_DIR}"
       alwaysRun: true
-  - ShellCommand: &retrieve_iso
-      name: Retrieve ISO image
-      env:
+  - ShellCommand: &retrieve_iso_checksum
+      name: Retrieve ISO image checksum
+      env: &_env_retrieve_artifact
         BASE_URL: "%(prop:artifacts_private_url)s"
         DEST_DIR: "."
-      command: >
-        curl -s -XGET -o "${DEST_DIR}/metalk8s.iso" "${BASE_URL}/metalk8s.iso"
-      haltOnFailure: true
-      # Increase default timeout for ISOs, as artifacts may be too slow
-      timeout: 2400
-  - ShellCommand: &retrieve_iso_checksum
-      <<: *retrieve_iso
-      name: Retrieve ISO image checksum
       command: >
         curl -s -XGET -o "${DEST_DIR}/SHA256SUM" "${BASE_URL}/SHA256SUM"
       # Three minutes should be enough for this small file
       timeout: 180
+  - ShellCommand: &retrieve_iso
+      name: Retrieve ISO image
+      env: &_env_retrieve_artifact_retry
+        <<: *_env_retrieve_artifact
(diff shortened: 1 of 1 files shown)


eve/main.yml
```diff
@@ -5,6 +5,205 @@ branches:
...
+      # There are 3 CPUs available for Docker, and 1 for `doit` in the Pod.
+      # Given the network IO-bound nature of some of the build steps, and
+      # most build steps running in Docker, set concurrency to 4.
+      command: ./doit.sh -n 4
+      usePTY: true
+      haltOnFailure: true
+  - ShellCommand: &ssh_ip_setup
+      name: Report IP and install SSH keys
+      command: >
+        ip a &&
```
alexandre-allard: Not related to the PR, but this command can be improved to only return the IP we want (eth0):
```
ip -f inet addr show eth0 | sed -En 's/^.*inet ([0-9.]+).*$/\1/p'
```
Or if we want all interfaces:
```
ip -f inet addr show | sed -En 's/^.*inet ([0-9.]+).* (.+)$/\2: \1/p'
```

eve/main.yml
```diff
@@ -5,6 +5,205 @@ branches:
...
+        sudo sosreport --all-logs
+        -o metalk8s -kmetalk8s.podlogs=True
+        -o containerd -kcontainerd.all=True -kcontainerd.logs=True
+        --batch --tmp-dir /var/tmp &&
+        sudo chown eve:eve /var/tmp/sosreport*
+  - ShellCommand: &wait_debug
+      name: Debug step - wait befor allowing resource destruction
+      timeout: 3600
+      command: sleep 3600
+      doStepIf: false
```
alexandre-allard: I think we should use some property defaulted to False for this doStepIf, so we would be able to trigger debug on build without changing this file and doing a debug commit.

alexandre-allard (dismissed):

eve/main.yml
```diff
@@ -5,6 +5,256 @@ branches:
...
+            REASON="step not in \"$DEBUG_STEPS\""
+          fi
+        else
+          REASON="invalid \"debug\" property value"
+          cat >&2 << EOF
+        Invalid "debug" build property value "$DEBUG_STEPS".
+        Must use either:
+          - "all", to select all debug steps
+          - a single step name
+          - a list of step names, separated by tilde signs "~",
```
alexandre-allard, line 112: Weird choice `~` as separation character between step names.
I would have used commas or semicolons, but ok, it's a detail. :)
Otherwise, I like this debugging system, way more usable than the previous one.

eve/main.yml
```diff
@@ -5,6 +5,256 @@ branches:
...
+        --batch --tmp-dir /var/tmp &&
+        sudo chown eve:eve /var/tmp/sosreport*
+  - ShellCommand: &wait_debug
+      name: Debug step - wait before allowing resource destruction
+      env:
+        # Use this step with STEP_NAME and DURATION environment values to
+        # customize the command behaviour. DURATION default value is 3600
+        # (wait duration, in seconds).
+        STEP_NAME: default
+      timeout: 3600
```
alexandre-allard, line 81: Timeout after 3600, so basically you can't put a `$DURATION` longer than 1 hour. :)
If I'm right, the timeout is triggered if there is no output during the period defined, but the step can be longer if there is something on stdout/err.
So maybe, you can have a loop sleeping for N minutes max (where N is less than global timeout) and printing something between each sleep.

gdemonet: See https://github.com/scality/metalk8s/pull/1946/files/916275ac08f58c8dc937240e6c8975fa4a2c3882#diff-73e74177595a435f88a0701e605cf08eR592-R596

alexandre-allard (approved):

Issue #1234, opened 2019-06-14
Add option in BootstrapConfiguration for always using IP-IP


gdemonet:
**Component**: salt, network

**Why this is needed**:

In case the networks backing a cluster drop any packet from an unknown origin, Pod-to-Pod communication through DNS resolution will be broken (since it will rely on ClusterIPs).
Always enabling IP-IP encapsulation would prevent this kind of issue.

**What should be done**:

Add a configuration option in the `BootstrapConfiguration` file (`/etc/metalk8s/bootstrap.yaml`) to control usage of IP-IP encapsulation. Values should be either `Always` or `CrossSubnet` (we use the latter in the current version of MetalK8s).

This configuration value should be reflected in the `calico-node` DaemonSet, replacing the `CALICO_IPV4POOL_IPIP` environment variable (see `salt/metalk8s/kubernetes/cni/calico/deployed.sls#L620`).

See https://docs.projectcalico.org/v3.5/usage/configuration/ip-in-ip for more details.

**Test plan**:

Pod-to-Pod communication should work with both variants of this option in a network that doesn't enforce the _known origin_ policy.
Pod-to-Pod communication should work with IP-IP set to `Always` in a network that do enforce this policy.

PR #1951, opened 2019-10-23
docs: Add enable IP-in-IP encapsulation

### LINKED ISSUE #1949: Document how to enable IP-in-IP encapsulation for virtual networks
**Component**: networking, calico

**Why this is needed**:

For installations of MetalK8s on virtual hardware/networks (e.g. public clouds such as AWS EC2), we cannot rely on Calico's `CrossSubnet` encapsulation strategy, because packets will be dropped due to unknown destination addresses (`ClusterIP`s).

**What should be done**:

Provide in documentation (in a subsection of the Quickstart guide for now) a procedure to enable IP-in-IP encapsulation even within a subnet (Calico's `Always` strategy).

Use the `kubectl patch` from `eve/workers/openstack-multiple-nodes/terraform/scripts/enable-ipip.sh`

**Test plan**: not applicable

TeddyAndrieux:
**Component**:

'docs'

**Context**: 

Currently Calico encapsulation method is not configurable before the
installation it's always "CrossSubnet" by default.

**Summary**:

Add documentation about how to enable IP-in-IP "Always".

---

Sees: #1234
Fixes: #1949

### DIFF
--- a/docs/operation/enable_ip_ip.rst
+++ b/docs/operation/enable_ip_ip.rst
@@ -1,0 +1,22 @@
+.. _enable IP-in-IP:
+
+Enable IP-in-IP encapsulation
+=============================
+
+.. _IP-in-IP: https://en.wikipedia.org/wiki/IP_in_IP
+.. _Calico: https://docs.projectcalico.org/
+.. _IP-in-IP Calico configuration: https://docs.projectcalico.org/v3.7/networking/vxlan-ipip
+
+By default Calico_ in MetalK8s is configured to use IP-in-IP_ encapsulation
+only for cross-subnet communication.
+
+To always use IP-in-IP_ encapsulation run the following command:
+
+.. code-block:: shell
+
+    $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
+      patch ippool default-ipv4-ippool --type=merge \
+      --patch '{"spec": {"ipipMode": "Always"}}'
+
+
+For more details refer to `IP-in-IP Calico configuration`_.
--- a/docs/quickstart/bootstrap.rst
+++ b/docs/quickstart/bootstrap.rst
@@ -104,6 +104,12 @@ Bootstrap node.
 
    root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh
 
+.. warning::
+
+    In case of virtual network (or any network which enforces source and
+    destination fields of IP packets to correspond to the MAC address(es))
+    :ref:`IP-in-IP needs to be enabled<enable IP-in-IP>`.
+
 Provision storage for Prometheus services
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 After bootstrapping the cluster, the Prometheus and AlertManager services used
--- a/docs/operation/index.rst
+++ b/docs/operation/index.rst
@@ -11,6 +11,7 @@ do not have a working MetalK8s_ setup.
 .. toctree::
 
    bootstrap_backup_restore
+   enable_ip_ip
    preparation
    upgrade
    downgrade


docs/quickstart/bootstrap.rst
```diff
@@ -104,6 +104,11 @@ Bootstrap node.
 
    root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh
 
+.. warning::
+
+    In case of virtual network (or any network not able to route workload IP
```
gdemonet: _"Workload IP addresses"_ is unclear to me, maybe you meant _"cluster-internal IP"_

docs/quickstart/bootstrap.rst
```diff
@@ -104,6 +104,11 @@ Bootstrap node.
...
+    addresses). :ref:`IP-in-IP need to be enabled<enable IP-in-IP>`.
```
gdemonet:
```suggestion
    addresses), :ref:`IP-in-IP need to be enabled<enable IP-in-IP>`.
```

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,17 @@
+.. _enable IP-in-IP:
+
+Enable IP-in-IP encapsulation
+=============================
+
+.. _IP-in-IP: https://en.wikipedia.org/wiki/IP_in_IP
+
+By default Calico in MetalK8s is configured to use IP-in-IP_ encapsulation only
+for Cross Subnet communication.
```
gdemonet:
```suggestion
for cross-subnet communication.
```

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,17 @@
...
+
+To Always use IP-in-IP_ encapsulation edit the IPPool used by Calico:
```
gdemonet:
```suggestion
To always use IP-in-IP_ encapsulation edit the ``IPPool`` used by Calico:
```

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,17 @@
+.. _enable IP-in-IP:
+
+Enable IP-in-IP encapsulation
+=============================
+
+.. _IP-in-IP: https://en.wikipedia.org/wiki/IP_in_IP
+
+By default Calico in MetalK8s is configured to use IP-in-IP_ encapsulation only
```
gdemonet: Calico would benefit from an explanation in the glossary + external link. Can become a tracking issue if you prefer.

docs/operation/enable_ip_ip.rst
gdemonet: Since we don't want to dive too deep in the internals of Calico for MetalK8s users, the "edit the IPPool used by Calico" could become "apply/run the following command"

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,18 @@
+.. _enable IP-in-IP:
+
+Enable IP-in-IP encapsulation
+=============================
+
+.. _IP-in-IP: https://en.wikipedia.org/wiki/IP_in_IP
+.. _Calico: https://docs.projectcalico.org/
+
+By default Calico_ in MetalK8s is configured to use IP-in-IP_ encapsulation
+only for Cross-Subnet communication.
```
gdemonet:
```suggestion
only for cross-subnet communication.
```

docs/quickstart/bootstrap.rst
```diff
@@ -104,6 +104,12 @@ Bootstrap node.
 
    root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh
 
+.. warning::
+
+    In case of virtual network (or any network not able to route
+    cluster-internal IP addresses),
+    :ref:`IP-in-IP need to be enabled<enable IP-in-IP>`.
```
gdemonet:
```suggestion
    :ref:`IP-in-IP needs to be enabled<enable IP-in-IP>`.
```

docs/quickstart/bootstrap.rst
```diff
@@ -104,6 +104,12 @@ Bootstrap node.
 
    root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh
 
+.. warning::
+
+    In case of virtual network (or any network not able to route
```
gdemonet: Not sure if _routing_ is the issue, since this is managed by Calico. The issue is that Calico is not encapsulating packets within a single network, leaving the origin and destination to internal IP addresses (each node being configured to answer for ARP requests on virtual network interfaces for those internal IPs).

Most virtual routers (such as OpenStack Neutron, see [here](https://docs.openstack.org/dragonflow/latest/specs/mac_spoofing.html)) enforce that all traffic leaving a VM can only contain packets with a `src` matching the actual MAC-IP pair on the virtual port (interface):

> Without this protection a VM in a network can spoof and answer to ARP requests that don’t actually belong to it, or fake response for other VM.

Not sure how we want to translate this into documentation, but it might be worth the effort (since we'll make this configurable in the future, see #1234).

gdemonet (changes requested):

docs/quickstart/bootstrap.rst
TeddyAndrieux: It's not only for virtual networks it's for all networks not able to route packet targetting internal IP.

If for what ever reason (firewall, some internal router in the network or .... what ever), a server or router or switch is not able to determine where he need to route this internal IP we need to have IP-in-IP encapsulation (or adding some default routes somewhere but this solution is not on our side)

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,18 @@
...
+By default Calico_ in MetalK8s is configured to use IP-in-IP_ encapsulation
+only for cross-subnet communication.
+
+To always use IP-in-IP_ encapsulation run the following command:
+
+.. code-block:: shell
+
+    $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
+      patch ippool default-ipv4-ippool --type=merge \
+      --patch '{"spec": {"ipipMode": "Always"}}'
```
NicolasT, line 19: I think it would make sense to also provide pointers to more in-depth information in our documentation. As an example, we could link to https://docs.projectcalico.org/v3.7/networking/vxlan-ipip from this document.

docs/quickstart/bootstrap.rst
NicolasT: I agree with @gdemonet, *routing* per se is not the issue.

> it's for all networks not able to route packet targetting internal IP.

Well, no, that's why we enable IPIP for *CrossSubnet* traffic by default.

I'm not sure what you mean with 'internal IP'. Do you mean PodIP? That's *not* how the routing works, as in, packets to PodIPs are not sent to any routers, they're sent immediately to the *machine* that's hosting said PodIP (because we're in-subnet).

I.e., if I have PodIP 1.0.0.1, and I'm in network 2.0.0.0/8, and the Pod is running on a host with IP 2.0.0.2, then if a Pod on another host wants to connect to this 1.0.0.1 Pod, there'll be a routing rule (not literally, Calico uses ranges, but bear with me) which says "1.0.0.1/32 via 2.0.0.2 dev ..." (and there's another entry saying "2.0.0.0/8 dev ... proto kernel scope link" or similar). So, if a packet is sent to 1.0.0.1, it will be sent to the *MAC* of the machine which ARP-responds to 2.0.0.2.
Basically, in this whole transmission, there's no real 'routing' involved in the sense that there's no host through which the packets flow which is not the destination *machine* of the packets, this is direct one-to-one communication.

docs/operation/enable_ip_ip.rst
TeddyAndrieux: Yes good one I will add this one. Already have the global link for calico project but this one is really about IP-IP so I will add it

docs/quickstart/bootstrap.rst
TeddyAndrieux: Hmm ok so it's only for virutal network ?

NicolasT: No, it's for any network which enforces source and destination fields of IP packets to correspond to the address(es) this security system expects/allows.

As such, in a 'standard' network where no link-level security is in place, IPIP is not required. However, in some virtual networks (AWS EC2, OpenStack with port security enabled,...) this is the case, and there may be physical networks where such packet spoofing is not permitted either.

TeddyAndrieux: Ok got it

docs/operation/enable_ip_ip.rst
```diff
@@ -0,0 +1,22 @@
...
+
+
+For more detail refer to `IP-in-IP Calico configuration`_.
```
gdemonet:
```suggestion
For more details refer to `IP-in-IP Calico configuration`_.
```

gdemonet (approved):
LGTM! Thanks :)

NicolasT:
There's one more issue with the current PR: right now, it explains *why* you'd want to enable IPIP in the quickstart, but it *doesn't* say *why* you'd want it on the page that explains how to enable it...

PR #1957, opened 2019-10-24
Authentication requirements in MetalK8s

thomasdanan:
**Component**:

Requirements doc

**Context**: 

Requirements about Authentication and roles

**Summary**:

listing context, requirements and user story

### DIFF
--- a/docs/developer/architecture/authentication.rst
+++ b/docs/developer/architecture/authentication.rst
@@ -28,7 +28,7 @@ for both local identity provider or proxy to external identity provider.
 Integration with LDAP or Microsoft AD are the most important ones to support.
 
 We may want to keep the concept of pre provisioned superadmin user
-whatever Identity provider method we use ... to be discussed. 
+whatever Identity provider method we use ... to be discussed.
 
 User Stories:
 -------------
@@ -63,9 +63,26 @@ already provisioned users & groups.
 The way we do that integration is ideally through a page in MetalK8s UI, but
 probably acceptable to make the feature available through CLI only.
 
-**configuration persistence**
+**Authentication check**
+
+UI should make sure user is well authenticated and if not, redirect to
+the login page (internal or external)
+
+**Dex as Identity Provider and login page**
+
+Dex login page theme should be as closed as possible as the rest of UIs themes.
+
+**Configuration persistence**
 
 Upgrading MetalK8s should not affect configuration that was done earlier. If
 embedded Identity Provider was used and a list of users created, they still
 exist after an upgrade. If configuration to integrate with external identity
 provider was set up, it should remain after an Upgrade.
+
+Open questions:
+---------------
+
+Should we define default role binding for this first iteration?
+If we don't, new created users or the once retrieved from external
+Identity Provider won't be able to do anything with UIs or APIs, except
+maybe accessing Grafana, Prometheus and Alert Manager
--- a/docs/developer/architecture/authentication.rst
+++ b/docs/developer/architecture/authentication.rst
@@ -1,0 +1,99 @@
+Authentication Architecture Document
+====================================
+
+Context:
+--------
+Currently, when we deploy MetalK8s we pre provision super admin user with admin
+password and with all privileges.
+This is not good enough in term of security and flexibility.
+
+The one deploying MetalK8s would like to decide who can do what (with UI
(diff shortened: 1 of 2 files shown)


docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,71 @@
...
+-------------
+
+Basically we are talking about
+
+- Being able to provision users with an embedded identity provider
+- Being able to integrate with an external Identity provider
+
+Dex https://github.com/dexidp/dex seems to be the preferred solution for now,
+for both local identity provider or proxy to external identity provider.
+Integration with LDAP or Microsoft AD are the most important ones to support.
```
Ebaneck, line 28: These are Dex **connector** features which come out of the box and should be straight forward to configure.

Q1. Should we consider handling/provide the guidelines for installation/configuration, for example of an Active Directory server??

We can enable the Dex connector for LDAP and AD in the `oidc configmap` but setup help from client-side might be required in terms of documentation.. TBD

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,71 @@
...
+
+We may want to keep the concept of pre provisioned superadmin user
+whatever Identity provider method we use ... to be discussed.
```
Ebaneck: Agree that we might need to have a statically pre-defined Dex superuser. If we intend to support some sort of user management(add/delete) in the Metalk8s UI someday, then this is mandatory since User addition over the Dex GRPC API would only work when the user performing the action has admin privileges.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,71 @@
...
+As an IT Generalist, I want to update super admin user password in order to
+secure access to all UIs / APIs
+
+Right after bootstrap node installation, the first time I log into the MetalK8s
+admin UI,
+I will be asked to change the super admin password. I would probably need to
+re-login just after that.
+
+In any cases, I should be able to reset this password, from CLI, in case it has
+been lost.
```
Ebaneck, line 46: Possible from the CLI, but a `configmap` edit is required to have this work correctly.

Static user passwords are stored as  bcrypt hashes only so this needs careful documentation

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Ideally, we have one dedicated page in MetalK8s admin UI allowing the IT
+Generalist to manage the list of users. Maybe for a first iteration, this is
+done from CLI with well documented procedure. In any cases, password should
+not be stored in clear text and easily accessible.
+
+**External Identity Provider Integration**
+
+As an IT Generalist, I want to leverage my Identity provider in order to reuse
+already provisioned users & groups.
```
Ebaneck, line 61: This would raise the need for having a Kubernetes role binding for your Identity provider(Active Directory) user. Role binding creation from Metalk8s UI is not an available feature at this time hence this needs CLI documentation

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,71 @@
...
+
+The way we do that integration is ideally through a page in MetalK8s UI, but
+probably acceptable to make the feature available through CLI only.
+
+**configuration persistence**
+
+Upgrading MetalK8s should not affect configuration that was done earlier. If
+embedded Identity Provider was used and a list of users created, they still
+exist after an upgrade. If configuration to integrate with external identity
+provider was set up, it should remain after an Upgrade.
```
Ebaneck, line 80: Ideally, this does not seem to be a problem since Dex provides storage.

Two of the storage methods we can greatly leverage are: **ETCD** or **Kubernetes Custom Resources** which comes default with Dex config.

We currently support backup and restore of ETCD server on dev 2.4.

We could streamline and discuss the merits and demerits of the storage medium to use in the next meeting.

I dont think we will ever consider Postgresql or an sqlite db

Ebaneck (commented):
Quick first round!

docs/developer/architecture/authentication.rst
thomasdanan: I think by default we use Dex as the Identity provider and then we have a documented procedure to configure Dex connector for LDAP or AD ... not sure to understand what kind of help we need? you mean to better know mandatory inputs we need to properly configure integration to LDAP or AD ?

PR #1959, opened 2019-10-25
salt: Waiting for salt-minion startup state to finish

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Salt can not have several state running at the same time. In salt-minion
configuration we have startup state to provision volumues it is done in
`parallel=True` but the state may take time to retrieve pillar
information or stuff like that, so we need to wait before running another
state.
We can't do it directly in the salt minion state as it is a state running
on the minion we will always have state running on this minion...

**Summary**:

So we check it in the `metalk8s_saltutil.wait_minions` runner called in
the `deploy_node` orchestrate (where we got issue because trying to
launch state right after restarting the salt-minion)

**Acceptance criteria**: 

To test it run a state that sleep and try running the runner

```
[root@bootstrap-rhel ~]# crictl exec -it $(crictl ps -q --label io.kubernetes.container.name=salt-master) sh -c "salt 'bootstrap' state.single cmd.run 'sleep 20' & salt-run metalk8s_saltutil.wait_minions bootstrap -l info"
[INFO    ] [Attempt 1/10] Waiting for running jobs to complete: State on minion "bootstrap": PID=29135 JID=20191025101037404502
[INFO    ] [Attempt 2/10] Waiting for running jobs to complete: State on minion "bootstrap": PID=29135 JID=20191025101037404502
[INFO    ] [Attempt 3/10] Waiting for running jobs to complete: State on minion "bootstrap": PID=29135 JID=20191025101037404502
[INFO    ] [Attempt 4/10] Waiting for running jobs to complete: State on minion "bootstrap": PID=29135 JID=20191025101037404502
[INFO    ] [Attempt 5/10] Waiting for running jobs to complete: State on minion "bootstrap": PID=29135 JID=20191025101037404502
bootstrap:
----------
          ID: sleep 20
    Function: cmd.run
      Result: True
     Comment: Command "sleep 20" run
     Started: 06:10:40.623805
    Duration: 20019.585 ms
     Changes:   
              ----------
              pid:
                  29361
              retcode:
                  0
              stderr:
              stdout:

Summary for bootstrap
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:  20.020 s
comment:
    All minions matching "bootstrap" responded and finished startup state: bootstrap
result:
    True
[INFO    ] Runner completed: 20191025101037306164
```

### DIFF
--- a/salt/_runners/metalk8s_saltutil.py
+++ b/salt/_runners/metalk8s_saltutil.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, print_function, unicode_literals
 import logging
+import time
 
 import salt.client
 import salt.utils.extmods
@@ -61,11 +62,57 @@ def wait_minions(tgt='*', retry=10):
             'error': error_message
         }
 
+    # Waiting for running states to complete
+    state_running = client.cmd(tgt, 'saltutil.is_running', arg=['state.*'])
+    attempts = 1
+
+    # If we got only empty result then no state running
+    while attempts <= retry and any(state_running.values()):
+        log.info(
+            "[Attempt %d/%d] Waiting for running jobs to complete: %s",
+            attempts,
+            retry,
+            ' - '.join(
+                'State on minion "{minion}": {states}'.format(
+                    minion=minion,
+                    states=', '.join(
+                        "PID={state[pid]} JID={state[jid]}".format(state=state)
+                        for state in running_states
+                    )
+                )
+                for minion, running_states in state_running.items()
+                if running_states
+            )
+        )
+        time.sleep(5)
+        state_running = client.cmd(tgt, 'saltutil.is_running', arg=['state.*'])
+        attempts += 1
+
+    if any(state_running.values()):
+        error_message = (
+            'Minion{plural} still have running state after {retry} retries: '
+            '{minions}'
+        ).format(
+            plural='s' if len(state_running) > 1 else '',
+            retry=retry,
+            minions=', '.join(
+                minion for minion, running_state in state_running.items()
+                if running_state
+            )
+        )
+        log.error(error_message)
+        return {
+            'result': False,
+            'error': error_message
+        }
+
     return {
         'result': True,
-        'comment': 'All minions matching "{}" responded: {}'.format(
-            tgt, ', '.join(minions)
-        )
+        'comment':
+            'All minions matching "{}" responded and finished startup '
+            'state: {}'.format(
+                tgt, ', '.join(minions)
+            )
     }
 
 


alexandre-allard (dismissed):

alexandre-allard:
Some typos in the commit message:
several stateS
volumes instead of volumues
And
```
We can't do it directly in the salt minion state as it is a state running
on the minion we will always have state running on this minion...
```
Should be reworded with something like:
```
We can't do it directly in the salt minion state as it would always see itself as a running state.
```

salt/_runners/metalk8s_saltutil.py
```diff
@@ -61,11 +62,50 @@ def condition_reached(minions, attempts):
             'error': error_message
         }
 
+    # Waiting for startup state to finish
```
gdemonet: `saltutil.running` shows more than just "startup states", IIUC, it's actually showing all the running jobs on the minion (which are module executions, not states, so we would get a bunch of `{"fun": "state.sls", "jid": ..., "pid": ..., "tgt": ...}` in `state_running`).

I think we could adjust comments and log messages to mark them as "running jobs" instead of "startup state".
It would also make sense to show info about the running jobs themselves:

```python
log.debug(
    "[Attempt %d/%d] Waiting for running jobs to complete: %s",
    attempts,
    retry,
    ' - '.join(
        'Jobs on minion "{minion}": {jobs}'.format(
            minion=minion,
            jobs=', '.join(
                '"{job[fun]}" ({job[pid]})'.format(job=job)
                for job in minion_jobs
            )
        )
        for minion, minion_jobs in running_jobs
    ),
)
```

_Note that I used `log.debug`, not sure this kind of "waiting" log warrants to be printed by default..._

gdemonet (changes requested):
Super cool, this is going to reduce CI flakiness a lot!

I would just request that this check actually prints information about the running jobs, and, if possible, to add an option to the `wait_minions` method for enabling this check.
If we really want to always check running jobs, then it would make sense to *not* use `test.ping` at all: `client.cmd(tgt, 'saltutil.running', ...)` would answer that the minion did not answer or something, I guess, and we could mutualize the retries.
But changing this can be done later, I'd rather focus on having more useful info on the running jobs, in case we need to kill a hanging process.

salt/_runners/metalk8s_saltutil.py
TeddyAndrieux: `info` is not printed by default (I mean in the output but it in the log file)

gdemonet (dismissed):

gdemonet (approved):

PR #1961, opened 2019-10-28
test/ui: Use released Cypress 3.5.0

gdemonet:
This reverts the use of a development version and temporary build
artifacts from Cypress own CI, which always jeopardized our CI builds in
case the temporary URL disappears.
Since we required 3.5.0, to cope with self-signed certificates, and this
version is now released officially, we can remove this workaround.

See: d9c6024
See: 0140af9


NicolasT (approved):
LGTM if CI agrees.

NicolasT:
Though I guess CI won't agree since this fixes CI :thinking:

PR #1964, opened 2019-10-28
Authentication requirements in MetalK8s

thomasdanan:
**Component**:

Requirements doc

**Context**: 

Requirements about Authentication in MetalK8s

**Summary**:

Listing context, requirements and user story

**Acceptance criteria**: 

Agree on requirements with contributors

### DIFF
--- a/docs/developer/architecture/authentication.rst
+++ b/docs/developer/architecture/authentication.rst
@@ -5,84 +5,92 @@ Context:
 --------
 Currently, when we deploy MetalK8s we pre provision super admin user with admin
 password. Meaning anyone wanting to use K8S/Salt APIs should authenticate using
-this single super admin user.
+this single super admin user. Another way to access the APIs is by using K8S
+admin certificate that is stored in /etc/kubernetes/admin.conf. We could also
+manually provisioned other users and credentials as well as roles binding
+but the way the functionality is accessible currently is too complex and not
+enough secured since user / password are stored in clear text.
 This is not good enough in term of security and flexibility.
 
 We would like at least to be able to add different users with different
 credentials and ideally to integrate K8S authentication system with external
 identity provider.
 
-Managing role binding between user / groups and roles is not part of this
-specification.
+Managing K8S role binding between user / groups High level roles and K8S roles
+is not part of this specification.
 
 Requirements:
 -------------
 
 Basically we are talking about
 
-- Being able to provision users with an embedded identity provider
-- Being able to integrate with an external Identity provider
+- Being able to provision users with an embedded Identity provider (IDP)
+- Being able to integrate with an external IDP
 
-Dex https://github.com/dexidp/dex seems to be the preferred solution for now,
-for both local identity provider or proxy to external identity provider.
+`Dex <https://github.com/dexidp/dex>`_ seems to be the preferred solution for
+now, for both local identity provider or proxy to external IDP.
 Integration with LDAP or Microsoft AD are the most important ones to support.
-
-We may want to keep the concept of pre provisioned superadmin user
-whatever Identity provider method we use ... to be discussed.
 
 User Stories:
 -------------
-**Super admin user password**
+**User Management with Dex as an IDP**
 
-As an IT Generalist, I want to update super admin user password in order to
-secure access to all UIs / APIs
+As an IT Generalist, I want to provision/edit usersa and high level roles.
+group provisioning not supported by Dex from what I understand.
+The Product high level roles are:
+- Cluster Admin role
+- Solution Admin role
+- Read Only
 
-Right after bootstrap node installation, the first time I log into the MetalK8s
-admin UI,
-I will be asked to change the super admin password. I would probably need to
-re-login just after that.
+This is done from CLI with well documented procedure.
+Entered passwords are never visible and encrypted when stored in DEX DB.
+The CLI tool enables to add / delete and edit passwords and roles.
 
-In any cases, I should be able to reset this password, from CLI, in case it has
-been lost.
+The CLI tool is used right after bootstrap node deployment to provision the
+very first user (which won't be provisioned anymore by default)
 
-**User Management with Dex as an Identity provider**
 
-As an IT Generalist, I want to provision/edit users
-group provisioning not supported by Dex from what I understand
+**External IDP Integration**
 
-Ideally, we have one dedicated page in MetalK8s admin UI allowing the IT
-Generalist to manage the list of users. Maybe for a first iteration, this is
-done from CLI with well documented procedure. In any cases, password should
-not be stored in clear text and easily accessible.
+As an IT Generalist, I want to leverage my IDP in order to reuse already
+provisioned users & groups.
+The way we do that integration is through a CLI tool. This CLI tool will take
+LDAP or AD integration configuration file as an input.
+When External IDP Integration is set up, we can always use embedded IDP to
+authenticate.
 
-**External Identity Provider Integration**
-
-As an IT Generalist, I want to leverage my Identity provider in order to reuse
-already provisioned users & groups.
-
-The way we do that integration is ideally through a page in MetalK8s UI, but
-probably acceptable to make the feature available through CLI only.
 
 **Authentication check**
 
 UI should make sure user is well authenticated and if not, redirect to
-the login page (internal or external)
-
-**Dex as Identity Provider and login page**
-
-Dex login page theme should be as closed as possible as the rest of UIs themes.
+the DEX login page. In that DEX login page, the user should choose between
+authenticating with DEX local users or with external IDP.
+If no external IDP is configured, no choice is presented to the user.
+This DEX login page should be styled so that it looks like any other Scality
+web pages.
 
 **Configuration persistence**
 
-Upgrading MetalK8s should not affect configuration that was done earlier. If
-embedded Identity Provider was used and a list of users created, they still
-exist after an upgrade. If configuration to integrate with external identity
-provider was set up, it should remain after an Upgrade.
+Upgrading or redeploying MetalK8s should not affect configuration that was done
+earlier (i.e. local users and credentials as well as external IDP integration
+and configuration)
+
+**Securing access to stats and alerts**
+
+Our customers, may want to collect some statistics out of our Prometheus and
+Alert Manager instances. Some other may want to leverage K8S / Salt APIs to
+drive administration of the system. Those APIs should be secured enough either
+through OIDC, Service account or because only accessible through control plane
+Network.
+
+**SSO between Admin UIs**
+
+Once DEX is in place and user are provisioned through external IDP one
+authenticated user can easily navigate to the other admin UIs without having
+to re-authenticate.
 
 Open questions:
 ---------------
 
-Should we define default role binding for this first iteration?
-If we don't, new created users or the once retrieved from external
-Identity Provider won't be able to do anything with UIs or APIs, except
-maybe accessing Grafana, Prometheus and Alert Manager
+- Authentication across multiple sites
+- SSO across Admin UIs (control plane) and Zenko Management UIs (on data plane)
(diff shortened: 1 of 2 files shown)


docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,88 @@
+Authentication Architecture Document
+====================================
+
+Context:
+--------
+Currently, when we deploy MetalK8s we pre provision super admin user with admin
+password. Meaning anyone wanting to use K8S/Salt APIs should authenticate using
```
NicolasT: One can trivially add more users as well, with whatever group mappings, and set up custom `(Cluster)RoleBinding`s, so this isn't entirely true.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,88 @@
...
+this single super admin user.
+This is not good enough in term of security and flexibility.
+
+We would like at least to be able to add different users with different
```
NicolasT: As said above, you *can* do that today (though there are security aspects in the current solution that can't be ignored).

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Requirements:
+-------------
+
+Basically we are talking about
+
+- Being able to provision users with an embedded identity provider
+- Being able to integrate with an external Identity provider
+
+Dex https://github.com/dexidp/dex seems to be the preferred solution for now,
```
NicolasT: Let's write proper ReST documents with real links and such ;-)

NicolasT (dismissed):
WiP

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,88 @@
...
+probably acceptable to make the feature available through CLI only.
+
+**Authentication check**
+
+UI should make sure user is well authenticated and if not, redirect to
+the login page (internal or external)
+
+**Dex as Identity Provider and login page**
+
+Dex login page theme should be as closed as possible as the rest of UIs themes.
```
NicolasT:
```suggestion
Dex login page theme should be as close as possible as the rest of UIs themes.
```

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+
+`Dex <https://github.com/dexidp/dex>`_ seems to be the preferred solution for
+now, for both local identity provider or proxy to external IDP.
+Integration with LDAP or Microsoft AD are the most important ones to support.
+
+User Stories:
+-------------
+**User Management with Dex as an IDP**
+
+As an IT Generalist, I want to provision/edit usersa and high level roles.
```
Ebaneck: _Minor typos_
As an IT Generalist, I want to provision/edit **users** and high-level roles.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+earlier (i.e. local users and credentials as well as external IDP integration
+and configuration)
+
+**Securing access to stats and alerts**
+
+Our customers, may want to collect some statistics out of our Prometheus and
+Alert Manager instances. Some other may want to leverage K8S / Salt APIs to
+drive administration of the system. Those APIs should be secured enough either
+through OIDC, Service account or because only accessible through control plane
+Network.
```
gdemonet: So, what's the actual requirement? Is being accessible only in the CP network sufficient? If so, then it's done already.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+
+
+**Authentication check**
+
+UI should make sure user is well authenticated and if not, redirect to
+the DEX login page. In that DEX login page, the user should choose between
+authenticating with DEX local users or with external IDP.
+If no external IDP is configured, no choice is presented to the user.
+This DEX login page should be styled so that it looks like any other Scality
+web pages.
```
gdemonet: Would make sense to mention that all Admin UIs should rely on this same authentication mechanism

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+The CLI tool is used right after bootstrap node deployment to provision the
+very first user (which won't be provisioned anymore by default)
+
+
+**External IDP Integration**
+
+As an IT Generalist, I want to leverage my IDP in order to reuse already
+provisioned users & groups.
+The way we do that integration is through a CLI tool. This CLI tool will take
+LDAP or AD integration configuration file as an input.
```
gdemonet: I'd argue this is not in the realm of "requirements", but an implementation proposal.

Requirements could be summed up like:
- Users want to be able to configure external identity providers,
- they want to be able to do so without specific K8s/Dex knowledge

To which an implementation design would suggest developing a CLI tool to abstract K8s/Dex details.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+Requirements:
+-------------
+
+Basically we are talking about:
+
+- Being able to provision users with an embedded Identity provider (IDP)
+- Being able to integrate with an external IDP
+
+`Dex <https://github.com/dexidp/dex>`_ seems to be the preferred solution for
+now, for both local identity provider or proxy to external IDP.
```
gdemonet: Mentions of `Dex` in the requirements seem inappropriate. To me, `Dex` is _one_ solution to the set of requirements, and this choice should be documented in the "design" section of this architecture document, along with the discarded alternatives.

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+- Cluster Admin role
+- Solution Admin role
+- Read Only
+
+This is done from CLI with well documented procedure.
+Entered passwords are never visible and encrypted when stored in DEX DB.
+The CLI tool enables to add / delete and edit passwords and roles.
+
+The CLI tool is used right after bootstrap node deployment to provision the
+very first user (which won't be provisioned anymore by default)
```
gdemonet: That's a really good point, but is "lockdown by default" really a user requirement?

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Once DEX is in place and user are provisioned through external IDP one
+authenticated user can easily navigate to the other admin UIs without having
+to re-authenticate.
+
+Open questions:
+---------------
+
+- Authentication across multiple sites
+- SSO across Admin UIs (control plane) and Zenko Management UIs (on data plane)
```
gdemonet: Let's not mention other products in MetalK8s documents, or only as examples. In here, the question would be "Unified authentication and SSO between Admin UIs and other workload UIs".

Since those lie in different networks and namespaces, it's probably not achievable - but discussing this point would go in the "design" section of this document, IMHO.

docs/developer/architecture/authentication.rst
Ebaneck: I would argue that this will not be the first user to be provisioned.

I still think that the `Scality-Dex` superuser needs to be statically defined else we will have no way to connect Dex to the API-server.
Adding to this, the user experience becomes even more awful when we have to ask users to configure the API-server/Dex integration and force reboot the API-server.
https://github.com/dexidp/dex/blob/master/Documentation/kubernetes.md#configuring-the-openid-connect-plugin

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,97 @@
...
+Alert Manager instances. Some other may want to leverage K8S / Salt APIs to
+drive administration of the system. Those APIs should be secured enough either
+through OIDC, Service account or because only accessible through control plane
+Network.
+
+**SSO between Admin UIs**
+
+Once DEX is in place and user are provisioned through external IDP one
+authenticated user can easily navigate to the other admin UIs without having
+to re-authenticate.
```
Ebaneck: Once DEX is in place and **users** are provisioned through external IDP, authenticated users can easily navigate......
??

Ebaneck (changes requested):
A couple of Minors to fix

docs/developer/architecture/authentication.rst
thomasdanan: updated in last commit

docs/developer/architecture/authentication.rst
thomasdanan: updated

docs/developer/architecture/authentication.rst
gdemonet: No, you don't need a user to have K8s API accept OIDC tokens issued from Dex. As explained in your docs, the setup requires certificates to be generated (that's going to be done with Salt and the `ca:minion`), said certs must then be exposed to API Server, and that should be enough.

docs/developer/architecture/authentication.rst
thomasdanan: ok so will replace DEX with local IDP

docs/developer/architecture/authentication.rst
thomasdanan: done

docs/developer/architecture/authentication.rst
thomasdanan: done

docs/developer/architecture/authentication.rst
thomasdanan: done

docs/developer/architecture/authentication.rst
thomasdanan: done

docs/developer/architecture/authentication.rst
thomasdanan: agree, this was confusing ... wanted to say whatever IDP is in place ... will fix

docs/developer/architecture/authentication.rst
thomasdanan: Actually I changed my mind after discussing with other folks ... we want one pre provisioned admin and we want to ask him to change the default password at very first connection ... I put back the original requirement.

NicolasT: There's another req that should be added, IMHO, about the Vagrant deployment.

Similar to how we create (sparse) volumes for Prometheus and such by default when running `vagrant up`, I think it could make sense to also pre-provision some well-known user in said environment to retain a cluster that's fully working out of the box.

NicolasT: Dex connecting to K8s API is using a `ServiceAccount`, so no need to provision any OIDC user. And as @gdemonet said, you can set up K8s API to accept OIDC tokens *even* when the OIDC service is not (yet) running. No need for any restart.

Which is a good thing, otherwise your whole cluster would stop operating when there's an issue with some external component, i.e. the OIDC IdP in this case?

To define new users (in the case of Dex), we could either use its gRPC API (though I have some concerns about authn towards said API...), or work at the `ConfigMap` level, which is not an issue since we *have* `/etc/kubernetes/admin.conf` available when the user management script is executed. No need for any OIDC 'user' either.

thomasdanan: Actually I changed this req and put back the need of having a pre-provisioned user so I guess it would not be different with vagrant N

docs/developer/architecture/authentication.rst
thomasdanan: moved it to open question

docs/developer/architecture/authentication.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Open questions
+--------------
+
+- Authentication across multiple sites
+- SSO across MetalK8s and solutions Admin UIs and other workload Management UIs
+- Our customers may want to collect some statistics out of our Prometheus
+  instances. This API should be secured enough and could be made available
+  either through OIDC, Service account tokens or be made accessible only
+  through the control plane network.
```
gdemonet: This sentence could be reformulated to include the known considerations, e.g.:

```
This API could be authenticated using OIDC, using an OIDC proxy, or stay unauthenticated. One should consider:
- the low sensitivity of the exposed data
- the fact that it is only exposed on the control-plane network
- the fact that most consumers of Prometheus stats are not human (e.g. Grafana, a federating Prometheus, scripts...), hence not well-suited to performing the OIDC dance
```

Ebaneck (approved):
Let's merge it

Issue #1970, opened 2019-10-30
Bootstrap fails with locked Yum database


gdemonet:
**Component**: salt

**What happened**:

Locally, using Vagrant, we hit an issue during bootstrap (most often during the `Deploy early-stage bootstrap node` step), showing errors when trying to install packages, e.g.:

```
    bootstrap: Failure while running step 'Deploying early-stage bootstrap node in local mode (this may take a while)'
[...]
    bootstrap: Output:
    bootstrap: 
    bootstrap: << BEGIN >>
    bootstrap: warning: /vagrant/_build/root/packages/redhat/metalk8s-kubernetes-el7/14bfe6e75a9efc8eca3f638eb22c7e2ce759c67f95b43b16fae4ebabde1549f3-cri-tools-1.13.0-0.x86_64.rpm: Header V4 RSA/SHA512 Signature, key ID 3e1ba8d5: NOKEY
    bootstrap: [ERROR   ] Command '[u'systemd-run', u'--scope', u'yum', u'-y', u'install', u'yum-plugin-versionlock']' failed with return code: 1
    bootstrap: [ERROR   ] stdout: Running scope as unit run-6189.scope.
    bootstrap: Loaded plugins: fastestmirror
    bootstrap: Loading mirror speeds from cached hostfile
    bootstrap:  * base: distrib-coffee.ipsl.jussieu.fr
    bootstrap:  * extras: centos.crazyfrogs.org
    bootstrap:  * updates: ftp.pasteur.fr
    bootstrap: Resolving Dependencies
    bootstrap: --> Running transaction check
    bootstrap: ---> Package yum-plugin-versionlock.noarch 0:1.1.31-52.el7 will be installed
    bootstrap: --> Finished Dependency Resolution
    bootstrap: 
    bootstrap: Dependencies Resolved
    bootstrap: 
    bootstrap: ================================================================================
    bootstrap:  Package                      Arch         Version             Repository  Size
    bootstrap: ================================================================================
    bootstrap: Installing:
    bootstrap:  yum-plugin-versionlock       noarch       1.1.31-52.el7       base        36 k
    bootstrap: 
    bootstrap: Transaction Summary
    bootstrap: ================================================================================
    bootstrap: Install  1 Package
    bootstrap: 
    bootstrap: Total size: 36 k
    bootstrap: Installed size: 53 k
    bootstrap: Downloading packages:
    bootstrap: Running transaction check
    bootstrap: Running transaction test
[...]

PR #1971, opened 2019-10-30
salt: close rpm DB & yum repo/cache

alexandre-allard:
We forgot to close databases after checking the presence
of packages in repositories, it can lead to DB related errors
when trying to use yum just after calling the salt custom module

Refs: #1970

**Component**: salt

**Context**: RPM & yum DB not closed

**Acceptance criteria**: successful `./doit.sh vagrant_up`

---
See: #1970

### DIFF
--- a/salt/_modules/metalk8s_package_manager_yum.py
+++ b/salt/_modules/metalk8s_package_manager_yum.py
@@ -4,6 +4,7 @@ so that we can support downgrade in metalk8s
 '''
 import logging
 
+from contextlib import contextmanager
 from salt.exceptions import CommandExecutionError
 
 try:
@@ -168,6 +169,16 @@ def _disable_yum_loggers():
         yum_logger.setLevel(yum.logginglevels.__NO_LOGGING)
 
 
+@contextmanager
+def _get_yum_base():
+    ybase = yum.YumBase()
+    try:
+        yield ybase
+    finally:
+        ybase.closeRpmDB()
+        ybase.close()
+
+
 def check_pkg_availability(pkgs_info):
     '''
     Check that provided packages and their dependencies are available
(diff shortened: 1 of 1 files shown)


gdemonet (approved):

salt/_modules/metalk8s_package_manager_yum.py
```diff
@@ -198,6 +198,8 @@ def check_pkg_availability(pkgs_info):
         try:
             ybase.install(name=name, version=version, release=release)
         except yum.Errors.InstallError as exc:
+            ybase.closeRpmDB()
```
NicolasT: Not too fond of doing it this way, because there's now one `close` path in an `except` and a later `finally`. There are better ways to do error handling with resource releasing.

sayf-eddine-scality: implementing a python context and using `with` will close the object every time I guess

TeddyAndrieux: As this except raise right after this do not really matter but indeed I agree it's cleaner to do one big `try - finally`.

Exceot if something else is raised ... true

gdemonet: Makes sense, you can wrap the whole `for` loop in `try/finally`

gdemonet (changes requested):

gdemonet:
Just tested on my machine, it fixed the issue described in #1970 for me

salt/_modules/metalk8s_package_manager_yum.py
slaperche-scality: context manager are the way to handle this kind of things, they are the Python way of implementing RAII-like behavior.

gdemonet (approved):
Thanks! This PR would be a good candidate for a changelog entry IMO, so let's not close the issue yet

NicolasT:
Consider adding a GNU-style `CHANGELOG` file in the root of the project for now in which we can track things, then later on migrate those into whichever change tracking system we put in place.

alexandre-allard:
Creating another PR to add `CHANGELOG`, so we can have the fix for now.

PR #1976, opened 2019-10-31
1963: RHEL bootstrap installation in pre-merge

### LINKED ISSUE #1963: Test RHEL bootstrap in CI pre-merge
**Component**: ci

**Why this is needed**: Ensure that deployment works on RHEL

**What should be done**: Implement bootstrap deployment on RHEL

**Implementation proposal** (strongly recommended): Same thing that's already did for CentOS, we just need to handle the subscription-manager register before & unregister at the end.

alexandre-allard:
**Component**: ci

**Context**: Since we now support metalk8s over RHEL, we need to test it in the CI.

**Summary**: Add bootstrap installation test in the CI for RHEL.

**Acceptance criteria**: Green build.

---

Closes: #1963

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -266,6 +266,64 @@ models:
       workdir: build/eve/workers/openstack-multiple-nodes/terraform/
       haltOnFailure: true
 
+  # --- Re-usable steps related to terraform actions ---
+  - ShellCommand: &terraform_install
+      name: Download and install terraform
+      env:
+        TF_VERSION: "0.12.3"
+      command: >
+        TF_URL="https://releases.hashicorp.com/terraform/${TF_VERSION}/";
+        TF_FILE="terraform_${TF_VERSION}_linux_amd64.zip";
+        curl --retry 5 -O "${TF_URL}${TF_FILE}" &&
+        sudo unzip "${TF_FILE}" -d /usr/local/sbin/ &&
+        rm -f "${TF_FILE}"
+      haltOnFailure: true
+  - ShellCommand: &terraform_install_check
+      name: Check that terraform was installed
+      command: |-
+        if ! terraform --version 2&> /dev/null; then
+          echo "aborting - terraform not installed and required" >&2
+          exit 1
+        fi
+      haltOnFailure: true
+  - ShellCommand: &terraform_init
+      name: Init terraform
+      command: |-
+        for _ in $(seq 1 12); do
+          if terraform init; then
+            break
+          else
+            rm -rf .terraform/
+            sleep 5
+          fi
+        done;
+      haltOnFailure: true
+  - ShellCommand: &terraform_validate
+      name: Validate terraform definition
+      command: terraform validate
+      haltOnFailure: true
+  - ShellCommand: &terraform_apply
+      name: Spawn openstack virtual infra
+      command: terraform apply -auto-approve
+      env: &_env_terraform
+        OS_AUTH_URL: "%(secret:scality_cloud_auth_url)s"
+        OS_REGION_NAME: "%(secret:scality_cloud_region)s"
+        OS_USERNAME: "%(secret:scality_cloud_username)s"
+        OS_PASSWORD: "%(secret:scality_cloud_password)s"
+        OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
+        # FIXME: this makes hostnames too long
+        # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+      haltOnFailure: true
+  - ShellCommand: &terraform_destroy
+      name: Destroy openstack virtual infra
+      command: |-
+        for _ in $(seq 1 3); do
+           terraform destroy -auto-approve && break
+        done;
+      env: *_env_terraform
+      alwaysRun: true
+      sigtermTime: 600
+
   # --- Previous version (for Upgrade/Downgrade) ---
   - Git: &git_pull_prev
       name: clone previous version branch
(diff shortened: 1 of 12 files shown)


eve/workers/openstack-single-node-rhel/terraform/image.tf
```diff
@@ -0,0 +1,9 @@
+variable "openstack_image_name" {
+  type    = string
+  default = "rhel-server-updated-7.6-x86_64-kvm.qcow2"
+}
+
+variable "openstack_flavour_name" {
+  type    = string
+  default = "m1.medium"
```
gdemonet: This will probably be too small - I think we use `m1.large` on `single-node-[up|down]grade` currently.

Now, note that we don't do it in `multiple-nodes`, which may explain why it tends to be more flaky.

eve/workers/openstack-single-node-rhel/terraform/common.tf
```diff
@@ -0,0 +1,16 @@
+resource "random_string" "current" {
+  length  = 5
+  special = false
+}
+
+locals {
+  prefix = "metalk8s-${random_string.current.result}"
```
gdemonet, line 7: I'm thinking this could also be an `output`, so we can easily find the VMs from a build an clean the ones that are not from an active build (in case some were not removed).

eve/workers/openstack-single-node-rhel/terraform/bootstrap.tf
```diff
@@ -0,0 +1,54 @@
...
+  provisioner "remote-exec" {
+    inline = [
+      "sudo bash scripts/rhsm-register.sh '${var.rhsm_username}' '${var.rhsm_password}'",
+      "sudo bash scripts/bootstrap-config.sh",
+    ]
+  }
+
+  provisioner "remote-exec" {
+     when = "destroy"
+     on_failure = "continue"
```
gdemonet, line 45: What will this mean if we fail to unregister? How can we manually unregister it after resource deletion?

eve/main.yml
```diff
@@ -266,6 +266,64 @@ models:
...
+      name: Spawn openstack virtual infra
+      command: terraform apply -auto-approve
+      env: &_env_terraform
+        OS_AUTH_URL: "%(secret:scality_cloud_auth_url)s"
+        OS_REGION_NAME: "%(secret:scality_cloud_region)s"
+        OS_USERNAME: "%(secret:scality_cloud_username)s"
+        OS_PASSWORD: "%(secret:scality_cloud_password)s"
+        OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
+        # FIXME: this makes hostnames too long
+        # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
```
gdemonet, line 315: We may want to reintroduce this at some point, since it will help finding lost VMs, and we don't rely on hostnames anymore.

eve/main.yml
```diff
@@ -308,6 +366,7 @@ stages:
           stage_names:
             - single-node-upgrade
             - single-node-downgrade
+            - single-node-install-rhel
```
gdemonet, line 400: Would it make sense to rename the others with `-centos`?

eve/main.yml
```diff
@@ -605,6 +664,116 @@ stages:
...
+            sudo bash /var/tmp/metalk8s/bootstrap.sh --verbose
+          workdir: build/eve/workers/openstack-single-node-rhel/terraform/
+          haltOnFailure: true
+      - ShellCommand:
+          name: Provision Prometheus and AlertManager storage
+          env:
+            SSH_CONFIG: >-
+              eve/workers/openstack-single-node-rhel/terraform/ssh_config
+          command: >
+            scp -F $SSH_CONFIG eve/create-volumes.sh bootstrap:/tmp/create-volumes.sh &&
```
gdemonet, line 758: This `create-volumes.sh` script could actually be copied to the machine with Terraform, as we copy other scripts. Not necessarily in your PR though.

eve/main.yml
```diff
@@ -605,6 +664,116 @@ stages:
...
+      - ShellCommand:
+          name: Collect sosreport from the bootstrap node
+          env:
+            SSH_CONFIG: >-
+              eve/workers/openstack-single-node-rhel/terraform/ssh_config
+          command: >
+            mkdir -p sosreport/sosreport/single-node-install-rhel &&
+            scp -F $SSH_CONFIG bootstrap:/var/tmp/sosreport-* \
+              sosreport/sosreport/single-node-install-rhel/bootstrap-sosreport.tar.xz
+          alwaysRun: true
```
gdemonet: Could be worth factorizing with `multiple-nodes` equivalent steps IMO

gdemonet (changes requested):

eve/main.yml
gdemonet: ~I'll open an issue for this.~ See #1982

eve/workers/openstack-single-node-rhel/terraform/bootstrap.tf
alexandre-allard: From the RedHat web site, anyway it should not fail, but at least we destroy the VM and release openstack resources.
If it fails, the worker will be destroyed at some point and ssh keys to access VM will be lost, so not sure it's better to have both the VM to destroy and the subscription to remove.

gdemonet: I'm not saying it should fail, just wondering what will happen if we fail to remove the subscription and the VM gets destroyed.

eve/workers/openstack-single-node-rhel/terraform/common.tf
alexandre-allard: done

eve/main.yml
alexandre-allard: done

eve/workers/openstack-single-node-rhel/terraform/image.tf
alexandre-allard: done

eve/main.yml
alexandre-allard: done

eve/workers/openstack-single-node-rhel/terraform/bootstrap.tf
alexandre-allard: The `on_failure: "continue"` is more here to avoid the destruction to fail if there is an error when trying to spawn the VM or if the machine is not registered yet (or anymore).

gdemonet: OK now with your explanations I'm lost.

What I understand of a Terraform provisioner `on_failure` argument is that it controls how the provisioner's result affects the creation (or deletion, in that case) of the resource, here, the VM.
So, having `on_failure = "continue"` here means, if the "unregister_rhsm" script fails, the VM will still be destroyed.

My question is only: if the script fails, for some unknown reason, and the VM then gets destroyed as expected, do we need some manual action on RedHat side to unregister this now non-existent VM?

alexandre-allard: Yes, if the VM has been registered.

gdemonet: OK, will see if we can remember those to cleanup after the fact.

gdemonet (dismissed):

TeddyAndrieux (approved):

PR #1978, opened 2019-10-31
Add architectural and implementation documentation for Authentication

Ebaneck:
**Component**:

'documentation'

**Context**: 

See requirements as listed here : https://github.com/scality/metalk8s/pull/1964

**Summary**:

Add design choices and implementation guide for authentication in MetalK8s.

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/docs/developer/architecture/authentication.rst
+++ b/docs/developer/architecture/authentication.rst
@@ -21,6 +21,7 @@ identity provider.
 Managing K8S role binding between user/groups High level roles and K8S roles
 is not part of this specification.
 
+.. _authentication-requirements:
 
 Requirements
 ------------
@@ -108,3 +109,183 @@ Open questions
   - the fact that most consumers of Prometheus stats are not human
     (e.g. Grafana, a federating Prometheus, scripts and others), hence not
     well-suited for performing the OIDC flow
+
+Design Choices
+--------------
+
+`Dex <https://github.com/dexidp/dex/>`_ is chosen as an Identity Provider(IdP)
+in MetalK8s based on the above :ref:`authentication-requirements` for the
+following reasons:
+
+* Dex's support for multiple plugins enable integrating the OIDC flow
+  with existing user management systems such as Active Directory,
+  LDAP, SAML and others.
+* Dex can be seamlessly deployed in a Kubernetes cluster.
+* Dex provides access to a highly customizable UI which is a step closer to
+  good user experience which we advocate for.
+* Dex can act as a fallback Identity Provider in cases where the external
+  providers become unavailable or are not configured.
+
+Rejected design choices
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Static password file Vs OpenID Connect
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using static password files involves adding new users by updating a static file
+located on every control-plane Node. This method requires restarting the
+Kubernetes API server for every new change introduced.
+
+This was rejected since it is inflexible to operate, requires storing user
+credentials and there is no support for a pluggable external identity provider
+such as LDAP.
+
+X.509 certificates Vs OpenID Connect
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Here, each user owns a signed certificate that is validated by the Kubernetes
+API server.
+
+This approach is not user-friendly that is each certificate has to be manually
+signed. Providing certificates for accessing the MetalK8s UI needs
+much more efforts since these certificates are browser incompatible.
+Using certificates is tedious since the certificate revocation process is also
+cumbersome.
+
+Keycloak Vs Dex
+~~~~~~~~~~~~~~~
+
+Both systems use OpenID Connect(OIDC) to authenticate a user using a
+standard OAuth2 flow.
+
+They both offer the ability to have short lived sessions so that user access
+can be rotated with minimum efforts.
+
+Finally, they both provide a means for identity management to be handled by an
+external service such as LDAP, Active Directory, SAML and others.
+
+Why not `Keycloak <https://www.keycoak.org/>`_?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Keycloak while offering similar features as Dex and even much more was rejected
+for the following reasons:
+
+- Keycloak is complex to operate(requires its own standalone database) and
+  manage(frequent database backups are required).
+
+- Currently, no use case exist for implementing a sophisticated Identity
+  Provider like Keycloak when the minimal Identity Provider from Dex is
+  sufficient.
+
+Note that, Keycloak is considered a future fallback Identity Provider if
+the need ever arises from a customer standpoint.
+
+Unexploited choices
+~~~~~~~~~~~~~~~~~~~
+
+* `Guard <https://github.com/appscode/guard/>`_
+
+A Kubernetes webhook authentication server by AppsCode, allowing you to
+log into your Kubernetes cluster by using various identity providers such as
+LDAP.
+
+* `ORY Hydra <https://github.com/ory/hydra/>`_
+
+It's an OpenID Connect provider optimized for low resource consumption.
+ORY Hydra is not an identity provider but it is able to connect to existing
+identity providers.
+
+Implementation Details
+----------------------
+
+Iteration 1
+~~~~~~~~~~~
+
+- Using Salt, generate self-signed certificates needed for Dex deployment
+- Deploy Dex in MetalK8s from the official **Dex Charts** while making use of
+  the generated certificates above
+- Provision an admin superuser
+- Configure Kubernetes API server flags to use Dex
+- Expose Dex on the control-plane using Ingress
+- Print the admin super user credentials to the CLI after MetalK8s bootstrap is
+  complete
+- Implement MetalK8s UI integration with Dex
+- Theme the Dex GUI to match MetalK8s UI specs(optional for iteration 1)
+
+Iteration 2
+~~~~~~~~~~~
+
+- Provide documentation on how to integrate with these external Identity
+  Providers especially LDAP and Microsoft Active Directory.
+
+Iteration 3
+~~~~~~~~~~~
+
+- Provide Single sign-on(SSO) for Grafana
+- Provide SSO between admin UIs
+
+Iteration 4
+~~~~~~~~~~~
+
+- Provide a CLI command to change the default superuser password as a prompt
+  after bootstrap
+- Provide a CLI for user management and provisioning
+
+The following operations will be supported using the CLI tool:
+
+- Create users password
+- List existing passwords
+- Delete users password
+- Edit existing password
+
+The CLI tool will also be used to create MetalK8s dedicated roles as already
+specified in the requirements section of this document.
+(see high-level roles from the requirements document)
+
+Since it is not advisable to perform the above mentioned operations at the Dex
+ConfigMap level, using the Dex gRPC API could be the way to go.
+
+Iteration 5
+~~~~~~~~~~~
+
+- Demand for a superuser's default password change upon first UI access
+- Provide UI integration that performs similar CLI operations for user
+  management and provisioning
+
+This means from the MetalK8s UI, a Cluster administrator should be able to do
+the following:
+
+- Create passwords for users
+- List existing passwords
+- Delete users password
+- Edit existing password
+
+.. note::
+
+   This iteration is completely optional for reasons being that the Identity
+   Provider from Dex acts as a fallback for Kubernetes Administrators who do
+   not want to use an external Identity Provider(mostly because they
+   have a very small user store).
+
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document the predefined dex roles(Cluster Admin role, Solution Admin role,
+  Read Only role), their access levels and how to create them.
+* Document how to create users and the secrets associated to them.
+* Document how to integrate Dex with external Identity Providers such as LDAP
+  and Microsoft Active Directory.
+
+In the Installation/Quickstart Guide
+
+* Document how to setup/change the superuser password
+
+
+Test Plan
+---------
+
+We could add some automated end-to-end tests for Dex user creation,
+and deletion using the CLI and then setup a mini-lab on scality cloud to try
+out the UI integration.


docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,137 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Iteration 1
+~~~~~~~~~~~
+
+- Deploy Dex in MetalK8s using the official **Dex Charts**
+- Setup Dex to use self-signed certificates
+- Expose Dex on the control plane using Ingress
+- Configure Kubernetes API server to use Dex
+- Provide documentation(Dex + API server integration)
```
thomasdanan: Why should contain this documentation if DEX + API server integration is done automatically when deploying bootstrap node and expanding control plane ? is it for explaining how to integrate external IDP ?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+- Provide documentation on how to create Dex users and secrets.
+
+Iteration 3
+~~~~~~~~~~~
+
+- Provide CLI Integration to create/manage users and corresponding
+  documentation.
+- Provide documentation on how to integrate with an external Identity Provider
+  focusing on Active Directory and LDAP.
+
```
thomasdanan: Shouldn't we split those iterations so that each one is bringing some value and is potentially demo-able and shippable to production?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
+Authentication
+==============
+
+Todo
+----
+
+- Provide uml diagrams for Dex + MetalK8s authentication flow
```
gdemonet: Not sure it's really required, we can link to the relevant parts of Dex / K8s docs regarding OIDC flow.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Open questions
+--------------
+
+- Is support for Dex monitoring and alerting on Grafana
+  something we should think about now?
```
gdemonet: Not sure what you mean by that.

- Should we monitor Dex? Yes
- Should we authenticate Prometheus and Grafana through Dex? Not sure, still an open question in the requirements

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Abstract
+--------
+
+To be able to identify who a user is and to manage(Create/Edit/Delete)
+these user identities, MetalK8s needs a middleman that acts in this light
+as an Identity Provider.
+
+To accomplish this, we introduce the concept of an OpenID Connect identity
+provider which has as role identity federation in a typical MetalK8s cluster.
```
gdemonet: Not exactly why we do this. To me, the rationale can be summed up in three points:
- we already have authentication with K8s API, but its static declaration is cumbersome, which explains why we want a dynamic IdP to replace it
- we also want to integrate with existing IdPs a customer would like to plug in
- finally, and most importantly, we don't want to store user credentials, which is where OIDC (or OAuth, for that matter), come into play

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+User Requirements for Authentication in MetalK8s has been discussed and is
+well documented. For more information, refer to -> [link to Requirements DOC]
+
+
+Design
+------
+
+For authentication in **MetalK8s**, we are going to use Dex, Dex
+connectors/plugins and it's integrated support for other
```
gdemonet: `it's` is an abbreviation for `it is`. I think you meant `its`.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+External IDP Integration as an Identity Provider.
```
gdemonet: Not super clear nor sufficient. We will mostly use Dex as an **OIDC Provider**, i.e. some service that can issue OIDC tokens for a user to authenticate against a third-party application (here, K8s API). As a bonus, Dex also can be used as an actual **Identity Provider**, in which case it acts as both parties in the OIDC flow.

To me, the design should first state that we **want** to use OIDC as the authentication mechanism against K8s API for human users, and then explain that we chose Dex for its ability to connect non-OIDC compatible IdPs, such as Active Directory, in that OIDC flow. And in the details of why we chose Dex, add a mention that we can still use it as a fallback IdP, in case the external ones are not reachable/configured.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Design choices
+^^^^^^^^^^^^^^
+
+Dex is chosen as an Identity Provider in MetalK8s based on the above
+requirements enforced by the following reasons:
+
+* Dex's support for multiple plugins greatly increases the potential for
+  integrating with existing user management system such as Active Directory,
+  LDAP, SAML and others.
```
gdemonet: "increases the potential" - be more affirmative, like "Dex's multiple plugins enable integrating the OIDC flow with existing ..."

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+* The ability for Dex to issue tokens, revoke tokens, refresh tokens and even
+  control the lifetime of these tokens makes **Dex** suitable since we can
+  force users to re-authenticate when need arises.
```
gdemonet: Not sure how that's a requirement (don't remember it being defined in #1964)

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+* Dex can be seamlessly deployed in a Kubernetes cluster.
+* Dex provides access to a highly customizable UI which is a step closer to
+  good user experience which we advocate for.
+* Dex provides support for multiple backends for storage. This guarantees user
+  account information persistence after upgrades.
```
gdemonet: Not sure how that's useful. We use `etcd`, Dex can use `etcd` and is well integrated in Kubernetes (uses CRDs and such), as you mentioned 2 lines above, so I think the point is made.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Rejected design choices
```
gdemonet: So, in those rejected choices, you mention static passwords and certificates. I agree they should be there, but you can't oppose those to Dex directly.

There are two levels of choice we made here:

- First choice was the **authentication method for end-users**: OIDC (the winner), vs. static passwords vs. X.509 certificates
- Second choice was the **OIDC provider**: Dex, vs. Keycloak (or other in-cluster providers) vs. public providers (such as Google, Facebook or GitHub)

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,154 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Static passwords
+~~~~~~~~~~~~~~~~
+
+This involves adding new users by updating a static file located on every
+control-plane Node and then restarting the API server.
+
+This was rejected since there is no support for a pluggable user management
+system and it is cumbersome to operate.
```
gdemonet: As mentioned above, it also means storing actual user credentials (contrary to what OIDC enables), which we'd prefer to avoid if possible.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,160 @@
...
+
+X.509 certificates
+~~~~~~~~~~~~~~~~~~
+
+Here, each user owns a signed certificate that is validated by the API
+server.
+
+This approach is not user-friendly meaning each certificate has to be manually
+signed. Also providing certificates for accessing the MetalK8s UI needs
+much more effort.
```
gdemonet: It's actually one of the common approaches in Kubernetes clusters. As a user, you create a `CertificateSigningRequest` object in the API, and a cluster administrator will validate it for you, effectively issuing a valid certificate for you to use. More info [here](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#requesting-a-certificate).

Now, that's indeed quite cumbersome to manage, and is not compatible with browser-based authentication as you mentioned.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Considered as one of the heavyweight identity management systems offering a
+full blown out Identity Management.
+
+Not suitable for our use case since it entails much efforts in terms of
+deployment and management. Also since we drive for a simplistic approach to
+Identity Management, Keycloak falls short in this aspect.
+
+Keycloak is considered a future fall-back Identity Provider if the need ever
+arises from our customers.
```
gdemonet: Keycloak offers everything Dex provides, and much more. We didn't choose it because of its complexity (it requires its own database and such) and also because it's overkill since we won't probably have a use-case for hosting complex user organizations (which is why we think the minimal IdP from Dex is sufficient).

It is indeed considered as fallback in case some customers require a more advanced user management **inside** MetalK8s itself (which we don't anticipate, because most of them already have such a system in place they can plug in, with either Keycloak or Dex).

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Unexploited choices
+~~~~~~~~~~~~~~~~~~~
+* ORY Hydra
```
gdemonet: That's actually an interesting example to detail. because ORY Hydra is actually providing _half_ of the requirements: it's an OIDC provider, but **not** an Identity provider, which means we can't have some cluster-local users in case there is no external system, or it is unreachable.

gdemonet (changes requested):
Feedback on the Design section first, Implementation review to come soon ;)

docs/developer/design/authentication.rst
gdemonet: I don't see how that's resolved, the question is still there. If the question is "should we provide a Grafana dashboard to consume monitoring data of Dex?", the answer is still **yes**. That's how we monitor things :)

A question worth asking is **when** will we do this, in your implementation plan. But such questions don't have a place in a design document IMHO, it's more a debate to have either in this peer review or during a design session / team conversation.

So, instead of resolving the conversation and asking the question in the document, just leave it open and start a discussion by mentioning people here, suggesting the various options you thought of and maybe some of the difficulties you anticipate. Once you get a final answer, then you'll know what to write in the document and where, and you'll be able to resolve this conversation. But let's not use merged code for conversation ;)

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
...
+- Configure Kubernetes API server flags to use Dex
+- Expose Dex on the control plane using Ingress
+- Implement MetalK8s UI integration with Dex
+- Theme the Dex GUI to match MetalK8s UI specs
+
+Iteration 2
+~~~~~~~~~~~
+
+- Provide integration with an external Identity Provider focusing on
+  LDAP and Active Directory
```
gdemonet: This part is shipped by Dex, not us, so not relevant in our implementation plan. Documentation is enough IMO.

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
...
+Iteration 1
+~~~~~~~~~~~
+
+- Using Salt, generate self-signed certificates needed for Dex deployment
+- Deploy Dex in MetalK8s from the official **Dex Charts** while making use of
+  the generated certificates above
+- Configure Kubernetes API server flags to use Dex
+- Expose Dex on the control plane using Ingress
+- Implement MetalK8s UI integration with Dex
+- Theme the Dex GUI to match MetalK8s UI specs
```
gdemonet: This may not be required for a first iteration... @thomasdanan?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
...
+
+- Provide integration with an external Identity Provider focusing on
+  LDAP and Active Directory
+- Provide documentation on how to integrate with these external Identity
+  Providers
+
+Iteration 3
+~~~~~~~~~~~
+
+- Provide a CLI for user management and provisioning
```
gdemonet: Maybe add some details, such as:

- what kind of operations should be supported
- which roles (see high-level roles from the requirements document)
- how it should work (use the gRPC API of Dex, IIRC)
- and more, if you think it can help evaluate the complexity of the task

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Iteration 4
+~~~~~~~~~~~
+
+- Provide UI integration that performs similar CLI operations for user
+  management and provisioning.
```
gdemonet: Same as above, some details to help understand the complexity of the task. The first iteration above is good (generate certs, use Helm charts, configure API server...)

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
...
+  LDAP and Active Directory
+- Provide documentation on how to integrate with these external Identity
+  Providers
+
+Iteration 3
+~~~~~~~~~~~
+
+- Provide a CLI for user management and provisioning
+
+Iteration 4
```
gdemonet: Mention that this iteration may be entirely optional, as the integrated IdP from Dex is more of a fallback for platforms that don't want to use an existing IdP (most likely, for platforms with few users).

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,187 @@
+Authentication
+==============
+
+Open questions
+--------------
+
+- Is support for Dex monitoring and alerting coupled with Grafana charts
+  something we should think about now?
```
gdemonet: See https://github.com/scality/metalk8s/pull/1978#discussion_r341353115

gdemonet (commented):
Review of implementation plan is done, good job! We should be ready to start working soon :)

docs/developer/design/authentication.rst
thomasdanan: @gdemonet I agree, this is the only one that is potentially optional for a first iteration

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,210 @@
...
+
+- Provide a CLI for user management and provisioning
+
+The following operations will be supported using the CLI tool:
+
+- Create users password
+- List existing passwords
+- Delete users password
+- Edit existing password
+- Revoke a users token
```
thomasdanan: Why do we need to be able to revoke users token?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,210 @@
...
+
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document the predefined dex roles(Cluster Admin role, Solution Admin role,
+  Read Only role), their access levels and how to create them.
+* Document how to create users and the secrets associated to them.
+* Document how to integrate Dex with external Identity Provider(LDAP).
```
thomasdanan: and Microsoft AD

docs/developer/design/authentication.rst
thomasdanan: @gdemonet @Ebaneck so to me, having SSO between admin UIs and Grafana is important ...I have added it to the spec. It could be tackled in a dedicated iteration if needs be.

docs/developer/design/authentication.rst
NicolasT: Good question, since you can't ;P

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,210 @@
...
+ORY Hydra is not an identity provider but it is able to connect to existing
+identity providers.
+
+Implementation Details
+----------------------
+
+Implementation steps(High level)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Iteration 1
```
thomasdanan: I don't see anything about pre-provisioned admin user and the ability to enforce password change at first login ... Is it something that should appear in another iteration? or maybe we don't have a solution for now?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,210 @@
...
+- List existing passwords
+- Delete users password
+- Edit existing password
+- Revoke a users token
+
+The CLI tool will also be used to create MetalK8s dedicated roles as already
+specified in the requirements section of this document.
+(see high-level roles from the requirements document)
+
+Since the above mentioned operations can not be performed at the Dex ConfigMap
```
NicolasT: Why not?

docs/developer/design/authentication.rst
```diff
@@ -0,0 +1,210 @@
...
+
+
+Test Plan
+---------
+
+We could add some automated end-to-end tests for Dex user creation,
+and deletion using the CLI and then setup a mini-lab on scality cloud to try
+out the UI integration.
+
+Provide a virtualbox disk image of Microsoft server configured with
```
NicolasT: We don't do VirtualBox in CI and such. Need to sync with the team that's done SAML integration for S3C on how to test this in CI.

docs/developer/design/authentication.rst
Ebaneck: @thomasdanan I think having a pre-provisioned admin user can be accomplished at the ConfigMap level during Dex deployment using the Helm charts.

Having a password change policy on first login would require mostly frontend work and interactions with the Dex gRPC API if we go down this road. 

or

Put in place the required documentation on how to change the static admin credentials which is usually encrypted and can be found in the Dex ConfigMap.

@gdemonet what's your take on this?

docs/developer/design/authentication.rst
Ebaneck: My bad here, you can only revoke tokens for registered Dex client applications

docs/developer/design/authentication.rst
Ebaneck: Well, it is true that these operations can be done at the ConfigMap level.

Maybe I should rephrase to `not advisable to be done at the ConfigMap level`

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
   - the fact that most consumers of Prometheus stats are not human
     (e.g. Grafana, a federating Prometheus, scripts and others), hence not
     well-suited for performing the OIDC flow
+
+Design Choices
+--------------
+
+`Dex <https://github.com/dexidp/dex/>`_ is chosen as an Identity Provider(IdP)
+in MetalK8s based on the above :ref:`authentication-requirements` enforced
+by the following reasons:
```
gdemonet: _enforced by the following reasons_ doesn't make much sense to me. Maybe just use _for the following reasons_

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+
+* Dex's support for multiple plugins enable integrating the OIDC flow
+  with existing user management systems such as Active Directory,
+  LDAP, SAML and others.
+* Dex can be seamlessly deployed in a Kubernetes cluster.
+* Dex provides access to a highly customizable UI which is a step closer to
+  good user experience which we advocate for.
+* Dex can act as a fall back Identity Provider incases were the external
```
gdemonet:
```suggestion
* Dex can act as a fallback Identity Provider in cases where the external
```

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+  providers become unable or are not configured.
```
gdemonet:
```suggestion
  providers become unavailable or are not configured.
```

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+- Provide documentation on how to integrate with these external Identity
+  Providers especially LDAP and Microsoft Active Directory.
+
+Iteration 3
+~~~~~~~~~~~
+
+- Provide Single sign-on(SSO) for Grafana
+- Provide SSO between admin UIs
+
+Iteration 4
```
gdemonet, line 227: In this iteration, add an item for: `Ask for changing the admin superuser password after bootstrap, providing an example CLI command`

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+- Edit existing password
+
+The CLI tool will also be used to create MetalK8s dedicated roles as already
+specified in the requirements section of this document.
+(see high-level roles from the requirements document)
+
+Since it is not advisable to perform the above mentioned operations at the Dex
+ConfigMap level, using the Dex gRPC API could be the way to go.
+
+Iteration 5
```
gdemonet, line 248: If this is done, one could remove the message from the previous iteration to change the default admin password, and instead request that change from the UI itself (that's how most Web applications do it, such as Grafana)

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+Test Plan
+---------
+
+We could add some automated end-to-end tests for Dex user creation,
+and deletion using the CLI and then setup a mini-lab on scality cloud to try
+out the UI integration.
+
+.. note::
+   Sync with the team that has done SAML integration for S3C to discuss on how
+   to test MetalK8s & Dex integration in the CI.
```
gdemonet: Mmh that's really a comment internal to Scality, I'd remove it from this doc.

docs/developer/architecture/authentication.rst
```diff
@@ -108,3 +109,180 @@ Open questions
...
+
+   This iteration is completely optional for reasons being that the Identity
+   Provider from Dex acts as a fallback for Kubernetes Administrators who do
+   not want to use an external Identity Provider(mostly because they
+   have a very small user store).
+
+Documentation
+-------------
+
+In the Operational Guide:
```
gdemonet, line 273: Also mention how to setup/change the admin password in the Installation/Quickstart guide(s).

gdemonet (changes requested):

gdemonet (approved):
LGTM, good job :)

PR #1981, opened 2019-10-31
Update the 'prometheus-operator' addon to the version 7.1.1 of the chart

NicolasT:
This could make porting to K8s 1.16 easier.

To reviewers: everything in `charts/prometheus-operator/` is a vendored package. You could review the changes, but in the end, only what's in `salt/metalk8s/addons/prometheus-operator/` is actually deployed, so changes there (which are more minimal than the changes in `charts/prometheus-operator`) are what's likely to be reviewed in detail.


salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -92,15 +92,23 @@ def _handle_v1_serviceaccount(obj, kubeconfig, context, absent):
...
+        'name': obj['metadata']['name'],
+        'kubeconfig': kubeconfig,
+        'context': context,
+        'namespace': obj['metadata']['namespace'],
+        'metadata': obj['metadata'],
+    }
+    if not absent:
+        args.update({
+            'data': obj['data'],
+        })
```
gdemonet, line 107: :+1:

gdemonet (approved):
LGTM (not sure about the `cleanup` state, because it would only work in the upgrade "direction", but can't think of anything better atm)

gdemonet:
We'll need to do the same for `nginx-ingress` by the way

NicolasT:
> LGTM (not sure about the `cleanup` state, because it would only work in the upgrade "direction", but can't think of anything better atm)

Not sure what you mean? In the 'downgrade' direction, the objects that used to exist should be created when applying the previous version of the manifest(s), no?

NicolasT:
> We'll need to do the same for `nginx-ingress` by the way

Yes, but this PR is big enough as-is :wink:

gdemonet:
> > LGTM (not sure about the `cleanup` state, because it would only work in the upgrade "direction", but can't think of anything better atm)
> 
> Not sure what you mean? In the 'downgrade' direction, the objects that used to exist should be created when applying the previous version of the manifest(s), no?

I meant that we won't be able to remove objects that only exist in the upper version, since the `deployed` state in the lower version cannot know about them.

PR #1979, opened 2019-10-31
UI: provision volumes with labels

carlito-scality:
**Component**:UI

**Context**: 
- After bootstrap, a user must create Volumes to provide storage for Prometheus and AlertManager. We deploy a StorageClass for this, but the PersistentVolumeClaims for these services also use a `labelSelector` to pick the 'right' PersistentVolume.

**Summary**:

**Acceptance criteria**: 
image

Screenshot 2019-10-31 at 14 17 59

---

Closes: https://github.com/scality/metalk8s/issues/1891 https://github.com/scality/metalk8s/issues/1977

PR #1984, opened 2019-11-04
UI: Add cluster status warning before provisioning the volume in Prometheus

### LINKED ISSUE #1890: Reduce Prometheus unavailability warning in UI when storage is not yet provisioned
As reported by @xaf-scality:

After bootstrap, but before (manually) provisioning the storage volumes required by Prometheus and AlertManager to operate, Prometheus can't start (by design). At this point, when accessing the UI, one gets strong warnings about Prometheus not being accessible.

While in general this is OK, we should tone this down a bit when this situation (i.e., the Prometheus Pod is Pending because its PVC can't be bound) by detecting it (we have access to the Kubernetes API). In this case, we should have some kind of 'orange' state, referring to the documentation (ideally linking to it, #1595) explaining how to resolve the situation.

ChengYanJin:
**Component**: ui/monitoring

**Context**:  
As a TS user, I would like to know how to deal with the error status of the cluster(no volume provisioned for Prometheus pod) instead of the red state of Prometheus.

**Summary**:

- Add the `Unknown` status for the cluster when there isn't provision volume for Prometheus. 

- Add a warning banner to explain the status `Prometheus is not available.` and how to deal with the issue which should refer the position of the doc.

**Acceptance criteria**: 
In order to reproduce the issue, you can simply comment on the following line in `Vagrantfile` and then do `vagrant up` so that we won't provision any volume during deploy the bootstrap node.
```
machine.vm.provision "create-volumes",
type: "shell",
inline: CREATE_VOLUMES
```
When you go to the monitoring page, you will wait a moment(until the api timeout) and get the following page.
image

Closes: #1890

### DIFF
--- a/ui/src/containers/ClusterMonitoring.js
+++ b/ui/src/containers/ClusterMonitoring.js
@@ -28,6 +28,9 @@ import { sortSelector } from '../services/utils';
 import NoRowsRenderer from '../components/NoRowsRenderer';
 import Banner from '../components/Banner';
 
+const VOLUME_PROVISION_DOC_REFERENCE =
+  'MetalK8s Quickstart Guide > Deployment of the Bootstrap node > Installation > Provision storage for Prometheus services';
+
 const PageContainer = styled.div`
   box-sizing: border-box;
   display: flex;
@@ -258,12 +261,10 @@ const ClusterMonitoring = props => {
         <Banner
           type={STATUS_BANNER_WARNING}
           icon={<i className="fas fa-exclamation-triangle" />}
-          title={'Prometheus is not available.'}
+          title={intl.messages.prometheus_not_available}
           messages={[
             <>
-              {
-                'Please refer to MetalK8s Quickstart Guide > Deployment of the Bootstrap node > Installation > Provision storage for Prometheus services'
-              }
+              {`${intl.messages.please_refer_to} ${VOLUME_PROVISION_DOC_REFERENCE}`}
             </>,
           ]}
         />
--- a/ui/src/containers/ClusterMonitoring.js
+++ b/ui/src/containers/ClusterMonitoring.js
@@ -255,9 +255,7 @@ const ClusterMonitoring = props => {
           </Tooltip>
         </RightClusterStatusContainer>
       </ClusterStatusTitleContainer>
-      {clusterStatus.value === CLUSTER_STATUS_UNKNOWN &&
-      cluster.prometheusPodStatusMessage ===
-        `0/1 nodes are available: 1 node(s) didn't find available persistent volumes to bind.` ? (
+      {cluster.isPrometheusVolumeProvisioned ? null : (
         <Banner
           type={STATUS_BANNER_WARNING}
           icon={<i className="fas fa-exclamation-triangle" />}
@@ -268,7 +266,7 @@ const ClusterMonitoring = props => {
             </>,
           ]}
         />
-      ) : null}
+      )}
       <PageSubtitle>
         {intl.messages.alerts}
         {alerts.isLoading ? <LoaderCoreUI size="small" /> : null}
@@ -308,7 +306,7 @@ const makeClusterStatus = (state, props) => {
     value = CLUSTER_STATUS_DOWN;
     label = intl.messages[cluster.error] || cluster.error;
   }
-  if (cluster.prometheusPodStatusMessage) {
+  if (!state.app.monitoring.isPrometheusApiUp) {
     value = CLUSTER_STATUS_UNKNOWN;
     label = intl.messages[cluster.error] || cluster.error;
   }
--- a/ui/src/containers/ClusterMonitoring.js
+++ b/ui/src/containers/ClusterMonitoring.js
@@ -17,10 +17,16 @@ import {
   stopRefreshClusterStatusAction,
   CLUSTER_STATUS_UP,
   CLUSTER_STATUS_DOWN,
+  CLUSTER_STATUS_UNKNOWN,
 } from '../ducks/app/monitoring';
-import { STATUS_CRITICAL, STATUS_SUCCESS } from '../constants';
+import {
+  STATUS_CRITICAL,
+  STATUS_SUCCESS,
+  STATUS_BANNER_WARNING,
+} from '../constants';
 import { sortSelector } from '../services/utils';
 import NoRowsRenderer from '../components/NoRowsRenderer';
+import Banner from '../components/Banner';
 
 const PageContainer = styled.div`
   box-sizing: border-box;
@@ -78,8 +84,16 @@ const RightClusterStatusContainer = styled.div`
 const ClusterStatusValue = styled.span`
   margin: 0 ${padding.small};
   font-weight: bold;
-  color: ${props =>
-    props.isUp ? props.theme.brand.success : props.theme.brand.danger};
+  color: ${props => {
+    switch (props.value) {
+      case CLUSTER_STATUS_UNKNOWN:
+        return props.theme.brand.warning;
+      case CLUSTER_STATUS_UP:
+        return props.theme.brand.success;
+      default:
+        return props.theme.brand.danger;
+    }
+  }};
 `;
 
 const QuestionMarkIcon = styled.i`
@@ -187,7 +201,7 @@ const ClusterMonitoring = props => {
       <ClusterStatusTitleContainer>
         <LeftClusterStatusContainer>
           <PageSubtitle>{intl.messages.cluster_status + ' :'}</PageSubtitle>
-          <ClusterStatusValue isUp={clusterStatus.value === CLUSTER_STATUS_UP}>
+          <ClusterStatusValue value={clusterStatus.value}>
             {clusterStatus.label}
           </ClusterStatusValue>
           <Tooltip
@@ -238,7 +252,22 @@ const ClusterMonitoring = props => {
           </Tooltip>
         </RightClusterStatusContainer>
       </ClusterStatusTitleContainer>
-
+      {clusterStatus.value === CLUSTER_STATUS_UNKNOWN &&
+      cluster.prometheusPodStatusMessage ===
+        `0/1 nodes are available: 1 node(s) didn't find available persistent volumes to bind.` ? (
+        <Banner
+          type={STATUS_BANNER_WARNING}
+          icon={<i className="fas fa-exclamation-triangle" />}
+          title={'Prometheus is not available.'}
+          messages={[
+            <>
+              {
+                'Please refer to MetalK8s Quickstart Guide > Deployment of the Bootstrap node > Installation > Provision storage for Prometheus services'
+              }
+            </>,
+          ]}
+        />
+      ) : null}
       <PageSubtitle>
         {intl.messages.alerts}
         {alerts.isLoading ? <LoaderCoreUI size="small" /> : null}
@@ -266,7 +295,6 @@ const makeClusterStatus = (state, props) => {
   const cluster = state.app.monitoring.cluster;
   let label = intl.messages.down;
   let value = CLUSTER_STATUS_DOWN;
-
   if (
     cluster.apiServerStatus > 0 &&
     cluster.kubeSchedulerStatus > 0 &&
@@ -275,12 +303,14 @@ const makeClusterStatus = (state, props) => {
     value = CLUSTER_STATUS_UP;
     label = intl.messages.cluster_up_and_running;
   }
-
   if (cluster.error) {
     value = CLUSTER_STATUS_DOWN;
     label = intl.messages[cluster.error] || cluster.error;
   }
-
+  if (cluster.prometheusPodStatusMessage) {
+    value = CLUSTER_STATUS_UNKNOWN;
+    label = intl.messages[cluster.error] || cluster.error;
+  }
   return { value, label, isLoading: cluster.isLoading };
 };
 
--- a/ui/src/ducks/app/monitoring.js
+++ b/ui/src/ducks/app/monitoring.js
@@ -105,7 +105,7 @@ export function* handlePrometheusError(clusterHealth, result) {
         c => c.type === 'PodScheduled',
       );
       if (
-        scheduledCondition.message.includes(
+        scheduledCondition?.message?.includes(
           `didn't find available persistent volumes to bind`,
         )
       ) {
--- a/ui/src/ducks/app/monitoring.js
+++ b/ui/src/ducks/app/monitoring.js
@@ -29,9 +29,9 @@ const defaultState = {
     kubeSchedulerStatus: 0,
     kubeControllerManagerStatus: 0,
     error: null,
-    prometheusPodStatusMessage: null,
     isLoading: false,
     isRefreshing: false,
+    isPrometheusVolumeProvisioned: true,
   },
   isPrometheusApiUp: false,
 };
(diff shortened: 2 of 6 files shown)


ui/src/ducks/app/monitoring.js
```diff
@@ -121,14 +128,32 @@ export function* fetchClusterStatus() {
...
+    const prometheusPod = yield call(
+      ApiK8s.getPodInNamespace,
+      'metalk8s-monitoring',
+      'prometheus',
+    );
+
+    if (!prometheusPod.error) {
+      const prometheusPodCondition =
+        prometheusPod?.body?.items[0]?.status?.conditions[0]?.type;
+      if (prometheusPodCondition === 'PodScheduled') {
```
carlito-scality: Why "PodScheduled" instead of "Ready"? You should also check the status of the condition

ui/src/containers/ClusterMonitoring.js
```diff
@@ -238,7 +255,22 @@ const ClusterMonitoring = props => {
           </Tooltip>
         </RightClusterStatusContainer>
       </ClusterStatusTitleContainer>
-
+      {clusterStatus.value === CLUSTER_STATUS_UNKNOWN &&
+      cluster.podStatusMessage ===
```
carlito-scality: This check is not robust IMO. @gdemonet any ideas?

ui/src/ducks/app/monitoring.js
```diff
@@ -20,17 +22,18 @@ const defaultState = {
...
     isLoading: false,
-    isRefreshing: false
+    isRefreshing: false,
   },
   cluster: {
     apiServerStatus: 0,
     kubeSchedulerStatus: 0,
     kubeControllerManagerStatus: 0,
     error: null,
+    podStatusMessage: null,
```
carlito-scality: you should rename it to "prometheusPodStatus" instead to be clearer. and It should contain the status of the pod instead of only the message

carlito-scality (changes requested):

ui/src/ducks/app/monitoring.js
ChengYanJin: The status of the condition is `False` so I think it's better to check the type of condition?

ui/src/ducks/app/monitoring.js
ChengYanJin: I will change the name from `podStatusMessage` to `prometheusPodStatusMessage`
Not sure if it's useful to store the status of the pod as in this case it's `pending`

gdemonet: IMO, this kind of information would make more sense in the `monitoring` Redux state tree, not in `monitoring.cluster`. But we can move this to a debt ticket, as it's not critical.

ui/src/containers/ClusterMonitoring.js
```diff
@@ -238,7 +255,16 @@ const ClusterMonitoring = props => {
           </Tooltip>
         </RightClusterStatusContainer>
       </ClusterStatusTitleContainer>
-
+      {clusterStatus.value === CLUSTER_STATUS_UNKNOWN &&
```
gdemonet: Instead of checking this `UNKNOWN` status, why not use the `monitoring.isPrometheusApiUp` boolean that already exists in the state?

`UNKNOWN` could exist while a query is in progress, for instance.

ui/src/containers/ClusterMonitoring.js
```diff
@@ -238,7 +255,16 @@ const ClusterMonitoring = props => {
...
+      cluster.prometheusPodStatusMessage ===
```
gdemonet: Indeed, I'm not sure if this value will always match this string. Let's be slightly more flexible:

```javascript
cluster.prometheusPodStatusMessage.includes(`didn't find available persistent volumes to bind`)
```

ui/src/ducks/app/monitoring.js
```diff
@@ -20,17 +22,18 @@ const defaultState = {
...
   },
   cluster: {
     apiServerStatus: 0,
     kubeSchedulerStatus: 0,
     kubeControllerManagerStatus: 0,
     error: null,
+    prometheusPodStatusMessage: null,
     isLoading: false,
-    isRefreshing: false
+    isRefreshing: false,
```
gdemonet, line 33: What's the difference between `loading` and `refreshing`? Confuses me, though not related to this PR.

ui/src/ducks/app/monitoring.js
```diff
@@ -20,17 +22,18 @@ const defaultState = {
...
   },
   cluster: {
     apiServerStatus: 0,
     kubeSchedulerStatus: 0,
     kubeControllerManagerStatus: 0,
     error: null,
+    prometheusPodStatusMessage: null,
```
gdemonet: IMHO, this should not be in `monitoring.cluster`, simply `monitoring`, or maybe even go as far as:

```
const defaultState = {
  // ...
  cluster: { ... },
  prometheus: {
    isAvailable: false, // the old `isPrometheusApiUp`
    podStatusMessage: null,
  },
}
```

That could help keep things organized. But feel free to put this in a debt ticket.

ui/src/ducks/app/monitoring.js
```diff
@@ -121,14 +128,31 @@ export function* fetchClusterStatus() {
...
+        prometheusPod?.body?.items[0]?.status?.conditions[0]?.type;
+      if (prometheusPodCondition === 'PodScheduled') {
+        clusterHealth.prometheusPodStatusMessage =
+          prometheusPod?.body?.items[0]?.status?.conditions[0]?.message;
+      } else {
+        console.error('This is an error in the cluster.');
+      }
+    } else {
+      console.error('Cannot find the Prometheus pod.');
+    }
```
gdemonet: I would move this logic in `handleClusterError` (which could be renamed `handlePrometheusError`, for clarity).

ui/src/ducks/app/monitoring.js
```diff
@@ -121,14 +128,31 @@ export function* fetchClusterStatus() {
...
+      'prometheus',
+    );
+    if (!prometheusPod.error) {
+      const prometheusPodCondition =
+        prometheusPod?.body?.items[0]?.status?.conditions[0]?.type;
+      if (prometheusPodCondition === 'PodScheduled') {
+        clusterHealth.prometheusPodStatusMessage =
+          prometheusPod?.body?.items[0]?.status?.conditions[0]?.message;
+      } else {
+        console.error('This is an error in the cluster.');
```
gdemonet: Not sure it's good practice to print error messages in the console. Also, this message doesn't help understand the actual state or why we hit it.

ui/src/ducks/app/monitoring.js
```diff
@@ -121,14 +128,31 @@ export function* fetchClusterStatus() {
...
+      }
+    } else {
+      console.error('Cannot find the Prometheus pod.');
```
gdemonet: This could be due to another error, you should check the HTTP status code is `404` before saying that. Anyway (as said above), I don't think we should print error messages in the console.

ui/src/ducks/app/monitoring.js
```diff
@@ -121,14 +128,31 @@ export function* fetchClusterStatus() {
...
     clusterHealth.kubeControllerManagerStatus = getClusterQueryStatus(
-      results[2]
+      results[2],
     );
     yield put(setPrometheusApiAvailable(true));
   } else {
+    const prometheusPod = yield call(
+      ApiK8s.getPodInNamespace,
+      'metalk8s-monitoring',
+      'prometheus',
```
gdemonet: Is this supposed to be the _Pod_ `metadata.name`? If it is, this won't work, as the actual _Pod_ is created through a _StatefulSet_. If it isn't, then the `getPodInNamespace` method is confusing (but that's debt, not your PR).

ui/src/ducks/app/monitoring.js
gdemonet: Mmh, actually this is not super robust. If one day another `condition` is added to _Pod_ `status`es, this will become somehow flaky. Instead, try something like:

```javascript
const conditions = prometheusPod?.body?.items[0]?.status?.conditions;
const scheduledCondition = conditions?.find(c => c.type === 'PodScheduled');
if (!scheduledCondition.status) {
  // update state with `scheduledCondition.message`
} else {
  // that's an unknown situation, handling TBD
}
```

ui/src/ducks/app/monitoring.js
```diff
@@ -90,6 +93,9 @@ export function* handleClusterError(clusterHealth, result) {
   if (result.error.response) {
     yield put(setPrometheusApiAvailable(true));
     clusterHealth.error = `Prometheus - ${result.error.response.statusText}`;
+  } else if (clusterHealth.prometheusPodStatusMessage) {
+    yield put(setPrometheusApiAvailable(false));
+    clusterHealth.error = 'Unknown';
```
gdemonet: No, actually if you want to have a different state in `clusterHealth` in case there's a missing Volume for Prometheus, then it's not "unknown", it's "missing a Volume". But I'd argue this can remain `prometheus_unavailable`, and the banner is enough to explain to the user **why** it is unavailable.

ui/src/ducks/app/monitoring.js
carlito-scality: Indeed, you have a status for each type of the condition. you should filter `prometheusPod?.body?.items[0]?.status?.conditions` to get the "Ready" (or "PodScheduled" ?)condition instead of getting the first condition. Then check the status of this condition if it is `false` then store the message

ui/src/services/k8s/api.js
```diff
@@ -183,3 +183,18 @@ export async function createEnvironment(body) {
...
+      null,
+      null,
+      null,
+      null,
+      `app=${podLabel}`,
+    );
+  } catch (error) {
+    return { error };
+  }
+}
```
gdemonet: Oh, that's not a good name/signature for this method IMHO (for instance, it's a `list` action, not a `get`, since you retrieve multiple objects that match a given `labelSelector`). But let's move it to a debt ticket, I think there are other methods like this one that could be improved.

gdemonet (changes requested):
Numerous comments, but let's focus on the actual handling of the message to merge this PR (keep the same `clusterHealth` that says "Prometheus is not available", and rely on the Pod status to display the banenr).

ui/src/ducks/app/monitoring.js
carlito-scality: it should be 'unknown' instead of "Unknown" to be translated

ui/src/containers/ClusterMonitoring.js
```diff
@@ -275,12 +300,14 @@ const makeClusterStatus = (state, props) => {
     value = CLUSTER_STATUS_UP;
     label = intl.messages.cluster_up_and_running;
   }
-
   if (cluster.error) {
     value = CLUSTER_STATUS_DOWN;
     label = intl.messages[cluster.error] || cluster.error;
   }
-
+  if (cluster.prometheusPodStatusMessage) {
```
carlito-scality: Here you use `prometheusPodStatusMessage ` message to compute the status of the cluster. It makes more sense to use the status of the prometheusPod instead. So after calling ApiK8s.getPodInNamespace, store the status in clusterHealth and use it here and when setting ` clusterHealth.error = 'Unknown';` in the saga

carlito-scality (dismissed):

ui/src/ducks/app/monitoring.js
ChengYanJin: > IMHO, this should not be in `monitoring.cluster`, simply `monitoring`, or maybe even go as far as:
> 
> ```
> const defaultState = {
>   // ...
>   cluster: { ... },
>   prometheus: {
>     isAvailable: false, // the old `isPrometheusApiUp`
>     podStatusMessage: null,
>   },
> }
> ```
> 
> That could help keep things organized. But feel free to put this in a debt ticket.

Will use`isPrometheusVolumeProvisioned` prop in cluster to control the banner visibility.

ui/src/ducks/app/monitoring.js
ChengYanJin: > I would move this logic in `handleClusterError` (which could be renamed `handlePrometheusError`, for clarity).

Refactoring the handlePrometheusError method in a separate commit

ui/src/services/k8s/api.js
ChengYanJin: > Oh, that's not a good name/signature for this method IMHO (for instance, it's a `list` action, not a `get`, since you retrieve multiple objects that match a given `labelSelector`). But let's move it to a debt ticket, I think there are other methods like this one that could be improved.

Make sense!
Will rename the method to `queryPodInNameSpace` maybe?

ui/src/ducks/app/monitoring.js
ChengYanJin: > This could be due to another error, you should check the HTTP status code is `404` before saying that. Anyway (as said above), I don't think we should print error messages in the console.

Removed the console.error!
Set clusterHealth.error='Unkown' instead of.

ui/src/containers/ClusterMonitoring.js
ChengYanJin: I use `isPrometheusVolumeProvisioned` to control the banner visibility and store it in clusterHealth. Is that make sense to you？

ui/src/ducks/app/monitoring.js
```diff
@@ -86,13 +89,37 @@ function getClusterQueryStatus(result) {
...
+        clusterHealth.isPrometheusVolumeProvisioned = false;
+        clusterHealth.error = 'unknown';
+      } else {
+        yield put(setPrometheusApiAvailable(false));
+        clusterHealth.error = 'unknown';
+      }
+    } else {
+      yield put(setPrometheusApiAvailable(false));
+      clusterHealth.error = 'prometheus_unavailable';
+    }
```
gdemonet: I don't get why we still have an `unknown` status. `prometheus_unavailable` should be enough:

```javascript
// Check if the error is due to an unschedulable Pod, because of missing PV
const prometheusPod = yield call(
  ApiK8s.queryPodInNamespace,
  'metalk8s-monitoring',
  'prometheus',
);

if (!prometheusPod.error) {
  const conditions = prometheusPod?.body?.items[0]?.status?.conditions;
  const scheduledCondition = conditions?.find(
    c => c.type === 'PodScheduled',
  );
  if (
    scheduledCondition.message.includes(
      `didn't find available persistent volumes to bind`,
    )
  ) {
    clusterHealth.isPrometheusVolumeProvisioned = false;
  }
}

// In any case, set these values
yield put(setPrometheusApiAvailable(false));
clusterHealth.error = 'prometheus_unavailable';
```

ui/src/translations/en.json
```diff
@@ -124,5 +124,7 @@
   "add_solution_to_environment": "Add solution to the environment {environment}",
   "version_env": "Versions",
   "add_solution": "Add the solution",
-  "advanced_monitoring": "Advanced Monitoring"
+  "advanced_monitoring": "Advanced Monitoring",
+  "prometheus_not_available":"Prometheus is not available.",
+  "provision_prometheus_volume":"Please refer to MetalK8s Quickstart Guide > Deployment of the Bootstrap node > Installation > Provision storage for Prometheus services"
```
gdemonet: We could just translate `Please refer to`, and leave the rest untranslated (since the doc is in English anyway).

gdemonet (changes requested):
Last changes, I promise ;)

gdemonet:
By the way, this PR will probably conflict with #1979 (since it changes some translations as well), so it may make sense to rebase as soon as it gets merged.

ui/src/containers/ClusterMonitoring.js
```diff
@@ -275,12 +305,14 @@ const makeClusterStatus = (state, props) => {
     value = CLUSTER_STATUS_UP;
     label = intl.messages.cluster_up_and_running;
   }
-
   if (cluster.error) {
     value = CLUSTER_STATUS_DOWN;
     label = intl.messages[cluster.error] || cluster.error;
   }
-
+  if (!cluster.isPrometheusVolumeProvisioned) {
```
gdemonet: Almost, that should be:

```suggestion
  if (!state.app.monitoring.isPrometheusApiUp) {
```

gdemonet (changes requested):

ChengYanJin:
> By the way, this PR will probably conflict with #1979 (since it changes some translations as well), so it may make sense to rebase as soon as it gets merged.

Thanks for the review!!!

gdemonet (dismissed):
LGTM! Thanks for the reactivity ;)

ui/src/containers/ClusterMonitoring.js
```diff
@@ -182,12 +200,14 @@ const ClusterMonitoring = props => {
 
   const sortedAlerts = sortSelector(alertsList, sortBy, sortDirection);
 
-  return (
+  return alerts.isLoading ? (
```
carlito-scality: You should not use `alerts.isLoading` here otherwise the whole page will refresh every 20s. Let's remove it for now and will improve it later.

carlito-scality (changes requested):
One last request ;)

ui/src/containers/ClusterMonitoring.js
ChengYanJin: ok. So I removed the `<Loader/>` for the moment Will find a better solution.

carlito-scality (dismissed):

gdemonet (dismissed):

gdemonet (approved):

PR #1987, opened 2019-11-05
Set version to 2.4.1 dev

alexandre-allard:
**Component**: metalk8s, version

**Context**: Bump metalk8s version to 2.4.1-dev (also push tag 2.4.0)

**Acceptance criteria**:


gdemonet (approved):

Ebaneck (commented):
@alexandre-allard-scality @gdemonet 

Since this looks like it is the first time we are introducing a version patch, What should the upgrade/downgrade path look like?

What we are testing in the CI now is deploy a 2.3.0 stable branch then upgrade to 2.4.1.

It should be deploy a 2.4.0 then test upgrade to 2.4.1... no??

alexandre-allard:
@Ebaneck not sure we want to test minor version upgrades, as we mostly want to check Kubernetes version upgrades.

gdemonet:
@alexandre-allard-scality @Ebaneck We will want to test "patch" upgrades, yes, but that's not in the scope of this PR IMO. We should still test "minor" upgrades, so nothing to be done on that side.

TeddyAndrieux:
@gdemonet @alexandre-allard-scality  @Ebaneck for me since the 2.4.0 is released we should have upgrade(downgrade) test from(to) dev2.4 to(from) 2.4.0 before releasing 2.4.1 (but it looks like we will not have them) 
But indeed it's not part of this PR

alexandre-allard:
@TeddyAndrieux I think you can create an issue to track this. :)

NicolasT:
@gdemonet This should've been done when `2.4.0` is merged into `development/2.4`, as part of that merge, right?

gdemonet:
> @gdemonet This should've been done when `2.4.0` is merged into `development/2.4`, as part of that merge, right?

This is the PR to merge the tag actually (quite late indeed, my bad), could have made this more obvious in the PR title/description.

PR #2024, opened 2019-11-07
Update `nginx-ingress` addon to the 1.24.7 version of the chart

gdemonet:
**Component**: salt, kubernetes, helm

**Context**:

K8s 1.16 effectively stops serving deprecated APIs, that some objects relied on in the existing Helm chart for `nginx-ingress`. The latest version takes that deprecation into account when rendering the chart, so a simple upgrade should be enough to support upgrading to K8s 1.16.

**Summary**:

Similar to #1981 - source chart and rendered SLS are updated. Reviewers should focus on the SLS changes, since the source chart is just vendored (with a minor edit to avoid Helm vs. Jinja conflicts).

**Acceptance criteria**: 

Ingress objects keep working, for both control-plane (check the UI for instance) and workload-plane (the default, one can try to create an `Ingress` object without the `ingress-class` annotation and test it from there).

Automatic tests should be added, but not in the scope of this PR.


NicolasT:
With the use of `raw` and `endraw` (injected by the render script, but not at the time the `nginx-ingress` add-on landed), the patch we have on top of the upstream chart may no longer be required...

gdemonet:
> With the use of `raw` and `endraw` (injected by the render script, but not at the time the `nginx-ingress` add-on landed), the patch we have on top of the upstream chart may no longer be required...

I needed to add some escaping to quotes in the values files (`nginx-ingress.yaml` and the control-plane equivalent), and managed to remove the patch on upstream. Good catch!

alexandre-allard (approved):

Issue #2007, opened 2019-11-05
Deploy Dex in MetalK8s from the Helm charts


Ebaneck:
**Component**:

'salt', 'kubernetes', 'containers'

**Why this is needed**:

To integrate a flexible identity and access management(IAM) in Metalk8s based on the authentication user requirements, we need to deploy an OIDC provider.

Dex is our choice of OIDC and needs to be deployed in a MetalK8s cluster.

**What should be done**:

- Using the official Dex charts, deploy 2 replicas of the latest release of Dex(v2.19) in a Metalk8s cluster.

**Implementation proposal** (strongly recommended):

**Test plan**:

- Have a Dex pod running in Metalk8s under the `metalk8s-auth` namespace

---

Epic iteration: #1988

NicolasT:
Please make sure to follow the exact mechanism used to install the other chart-based addons we embed.

NicolasT:
Why would we want/need to run this in `kube-system`?

Ebaneck:
> Why would we want/need to run this in `kube-system`?

@NicolasT since the Dex storage backend we will most likely be using is the Kubernetes backend, i can't think of any reason to not leave it in the `kube-system` namespace.

Generally what is considered best practice for such??

gdemonet:
> > Why would we want/need to run this in `kube-system`?
> 
> @NicolasT since the Dex storage backend we will most likely be using is the Kubernetes backend, i can't think of any reason to not leave it in the `kube-system` namespace.

Dex is using K8s API for interfacing with its storage, so it can run as any other workload (e.g. all Operators use K8s API as well). Having a dedicated namespace, as we do for the `nginx-ingress-controller`s (which use K8s API too), is recommended: go for `metalk8s-dex`. 

> Generally what is considered best practice for such??

`kube-system` is meant to group applications critical for the cluster functioning, such as `etcd` or `kube-apiserver`. One could argue `salt-master` shouldn't be part of it (it is today), but that's not the scope of this discussion. Dex isn't strictly necessary for the cluster to function (it's only necessary for end-users to authenticate against the API through OIDC, which they could circumvent using certificate-based auth if the need arises).

NicolasT:
> go for metalk8s-dex

`metalk8s-auth` or something along those lines. If ever we change IdP, we can keep things in the same namespace and such.

Issue #2011, opened 2019-11-05
Expose Dex on the control-plane ingress controller


Ebaneck:
**Component**: 

'salt', 'kubernetes',  'authentication'

**Why this is needed**:
To access Dex outside the Metalk8s cluster, we need to expose it over the control plane ingress

**What should be done**:
- Add an ingress object to expose the Dex web GUI

**Implementation proposal** (strongly recommended):

**Test plan**:
- Be able to navigate to the Dex login landing page 

---

Epic iteration: #1988

Issue #2010, opened 2019-11-05
Setup the Dex-to-Kubernetes API server integration 


Ebaneck:
**Component**: 

'salt', 'kubernetes', 'containers', 'authentication'

**Why this is needed**:

The Kubernetes API server needs to be configured to communicate with Dex to support user authentication.

**What should be done**:

- Setup the Kubernetes API server flags to integrate with Dex.
- Ensure the CA certificate generated for Dex is available to the Kubernetes API server container.

**Implementation proposal** (strongly recommended):

- Since we run Kubernetes API server as a container probably the easiest method is to use a hostPath volume to mount the CA file directly.

**Test plan**:
- The Kubernetes API server should continue to work after these changes.

---

Epic iteration: #1988

Issue #1996, opened 2019-11-05
Cluster and Services Configuration & Persistence


thomasdanan:
**Abstract**:

We require a unified design (as much as possible) to provide means for cluster administrators to configure the deployment, and have these settings persisted under common operations (upgrade/downgrade, reboots, restoration procedures...).

**Example settings**:

- Number of replicas for addons (Prometheus, Alertmanager, Grafana, Dex...)
- Dex configuration
- Alert/Rules configuration
- etc ...

PR #2025, opened 2019-11-07
Deploy dex in metalk8s from helm charts

Ebaneck:
**Component**:

'authentication', 'salt', 'kubernetes'

**Context**: 

See #2007 

**Summary**:

**Acceptance criteria**: 

- Have Dex pods running in MetalK8s cluster deployed from helm charts
- Access the dex URL exposed by ingress and verify that it works : https://{{ control_plane_address }}:8443/oidc/.well-known/openid-configuration

```
{
  "issuer": "https://172.21.254.3:8443/oidc",
  "authorization_endpoint": "https://172.21.254.3:8443/oidc/auth",
  "token_endpoint": "https://172.21.254.3:8443/oidc/token",
  "jwks_uri": "https://172.21.254.3:8443/oidc/keys",
  "userinfo_endpoint": "https://172.21.254.3:8443/oidc/userinfo",
  "response_types_supported": [
    "code",
    "id_token",
    "token"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "scopes_supported": [
    "openid",
    "email",
    "groups",
    "profile",
    "offline_access"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_basic"
  ],
  "claims_supported": [
    "aud",
    "email",
    "email_verified",
    "exp",
    "iat",
    "iss",
    "locale",
    "name",
    "sub"
  ]
}
```

---

Closes: #2007

### DIFF
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -19,6 +19,7 @@ CMD_WIDTH : int = 14
 # URLs of the main container repositories.
 CALICO_REPOSITORY     : str = 'quay.io/calico'
 COREOS_REPOSITORY     : str = 'quay.io/coreos'
+DEX_REPOSITORY        : str = 'quay.io/dexidp'
 DOCKER_REPOSITORY     : str = 'docker.io/library'
 GOOGLE_REPOSITORY     : str = 'k8s.gcr.io'
 GRAFANA_REPOSITORY    : str = 'docker.io/grafana'
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -216,6 +216,16 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
         renderer=targets.Renderer.JSON,
     ),
 
+    Path('salt/metalk8s/addons/dex/ca/init.sls'),
+    Path('salt/metalk8s/addons/dex/ca/installed.sls'),
+    Path('salt/metalk8s/addons/dex/ca/advertised.sls'),
+    Path('salt/metalk8s/addons/dex/certs/init.sls'),
+    Path('salt/metalk8s/addons/dex/certs/server.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/chart.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/init.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/namespace.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/tls-secret.sls'),
+
     Path('salt/metalk8s/addons/prometheus-operator/deployed/chart.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/cleanup.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls'),
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -225,6 +225,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/dex/deployed/init.sls'),
     Path('salt/metalk8s/addons/dex/deployed/namespace.sls'),
     Path('salt/metalk8s/addons/dex/deployed/tls-secret.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/clusterrolebinding.sls'),
 
     Path('salt/metalk8s/addons/prometheus-operator/deployed/chart.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/cleanup.sls'),
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -292,6 +292,8 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/nginx-ingress-control-plane/deployed/chart.sls'),
     Path('salt/metalk8s/addons/nginx-ingress-control-plane/deployed/',
          'tls-secret.sls'),
+    Path('salt/metalk8s/addons/nginx-ingress-control-plane/',
+         'control-plane-ip.sls'),
 
     Path('salt/metalk8s/container-engine/containerd/configured.sls'),
     Path('salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf'),
--- a/charts/dex.yaml
+++ b/charts/dex.yaml
@@ -1,0 +1,94 @@
+image: '{% endraw %}{{ build_image_name(\"dex\", False) }}{% raw %}'
+
+nodeSelector:
+  node-role.kubernetes.io/infra: ''
+
+tolerations:
+  - key: "node-role.kubernetes.io/bootstrap"
+    operator: "Exists"
+    effect: "NoSchedule"
+  - key: "node-role.kubernetes.io/infra"
+    operator: "Exists"
+    effect: "NoSchedule"
+
+replicas: 2
+
+# grpc support
+grpc: false
+
+# https termination by dex itself
+https: true
+
+service:
+  clusterIP: '{% endraw %}{{ salt.metalk8s_network.get_oidc_service_ip() }}{% raw %}'
+
+ingress:
+  enabled: true
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+    kubernetes.io/ingress.class: "nginx-control-plane"
+  path: /oidc
+  hosts:
+    - null
+
+# extraVolumes:
+#   - name: theme
+#     configMap:
+#        name: dex-branding
+
+# extraVolumeMounts:
+#   - name: theme
+#     mountPath: /web/themes/custom/
+
+certs:
+  web:
+    create: false
+  grpc:
+    create: false
+
+config:
+  issuer: '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc{% raw %}'
+  web:
+    tlsCert: /etc/dex/tls/https/server/tls.crt
+    tlsKey: /etc/dex/tls/https/server/tls.key
+  frontend:
+    theme: "coreos" #metalk8s-ui
+    # dir: /web/themes/custom/
+
+  connectors: {}
+
+  oauth2:
+    alwaysShowLoginScreen: true
+    skipApprovalScreen: true
+    responseTypes: ["code", "token", "id_token"]
+
+  expiry:
+    signingKeys: "6h"
+    idTokens: "24h"
+
+  staticClients:
+  - id: oidc-auth-client
+    redirectURIs:
+    - 'urn:ietf:wg:oauth:2.0:oob'
+    name: 'oidc-auth-client'
+    secret: "lkfa9jaf3kfakqyeoikfjakf93k2l"
+    trustedPeers:
+    - metalk8s-ui
+    - grafana-ui
+  - id: metalk8s-ui
+    redirectURIs:
+    - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/oauth2/callback{% raw %}'
+    name: 'MetalK8s UI'
+    secret: "ybrMJpVMQxsiZw26MhJzCjA2ut"
+  - id: grafana-ui
+    name: 'Grafana UI'
+    redirectURIs:
+    - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{% raw %}'
+    secret: "4lqK98NcsWG5qBRHJUqYM1"
+
+  staticPasswords:
+    - email: "admin@metalk8s.invalid"
+      # bcrypt hash of the string "password"
+      hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+      username: "admin"
+      userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
--- a/charts/dex/templates/clusterrole.yaml
+++ b/charts/dex/templates/clusterrole.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+{{ include "dex.labels" . | indent 4 }}
+  name: {{ template "dex.fullname" . }}
+rules:
+- apiGroups: ["dex.coreos.com"] # API group created by dex
+  resources: ["*"]
+  verbs: ["*"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["create"] # To manage its own resources, dex must be able to create customresourcedefinitions
+{{- end -}}
--- a/charts/dex/templates/deployment.yaml
+++ b/charts/dex/templates/deployment.yaml
@@ -1,0 +1,114 @@
+{{ $fullname := include "dex.fullname" . }}
+{{ $httpsTlsBuiltName := printf "%s-tls" $fullname }}
+{{ $httpsTlsSecretName := default $httpsTlsBuiltName .Values.certs.web.secret.tlsName }}
+{{ $grpcTlsServerBuiltName := printf "%s-server-tls" $fullname }}
+{{ $grpcTlsServerSecretName := default $grpcTlsServerBuiltName .Values.certs.grpc.secret.serverTlsName }}
+{{ $grpcCaBuiltName := printf "%s-ca" $fullname }}
+{{ $grpcCaSecretName := default $grpcCaBuiltName .Values.certs.grpc.secret.caName }}
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "dex.fullname" . }}
+  labels:
+{{ include "dex.labels" . | indent 4 }}
+    app.kubernetes.io/component: dex
+spec:
+  replicas: {{ .Values.replicas }}
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "dex.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      app.kubernetes.io/component: dex
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "dex.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/component: dex
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- if .Values.podAnnotations }}
+{{ toYaml .Values.podAnnotations | indent 8 }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ template "dex.serviceAccountName" . }}
+      nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 10 }}
+{{- with .Values.tolerations }}
+      tolerations:
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- with .Values.affinity }}
+      affinity:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+      containers:
+      - name: main
+        image: "{{ .Values.image }}:{{ .Values.imageTag }}"
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        command:
+        - /usr/local/bin/dex
+        - serve
+        - /etc/dex/cfg/config.yaml
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        ports:
+        - name: {{ if .Values.https }}https{{ else }}http{{ end }}
+          containerPort: {{ .Values.ports.web.containerPort }}
+          protocol: TCP
+        {{- if .Values.grpc }}
+        - name: grpc
+          containerPort: {{ .Values.ports.grpc.containerPort }}
+          protocol: TCP
+        {{- end }}
+        env:
+{{ toYaml .Values.env | indent 10 }}
+        volumeMounts:
+        - mountPath: /etc/dex/cfg
+          name: config
+{{- if .Values.https }}
+        - mountPath: /etc/dex/tls/https/server
+          name: https-tls
+{{- end }}
+{{- if .Values.grpc }}
+        - mountPath: /etc/dex/tls/grpc/server
+          name: grpc-tls-server
+        - mountPath: /etc/dex/tls/grpc/ca
+          name: grpc-tls-ca
+{{- end }}
+{{- if ne (len .Values.extraVolumeMounts) 0 }}
+{{ toYaml .Values.extraVolumeMounts | indent 8 }}
+{{- end }}
+      volumes:
+      - secret:
+          defaultMode: 420
+          items:
+          - key: config.yaml
+            path: config.yaml
+          secretName: {{ template "dex.fullname" . }}
+        name: config
+{{- if .Values.https }}
+      - name: https-tls
+        secret:
+          defaultMode: 420
+          secretName: {{ $httpsTlsSecretName | quote }}
+{{- end }}
+{{- if .Values.grpc }}
+      - name: grpc-tls-server
+        secret:
+          defaultMode: 420
+          secretName: {{ $grpcTlsServerSecretName | quote }}
+      - name: grpc-tls-ca
+        secret:
+          defaultMode: 420
+          secretName: {{ $grpcCaSecretName| quote }}
+{{- end }}
+{{- if ne (len .Values.extraVolumes) 0 }}
+{{ toYaml .Values.extraVolumes | indent 6 }}
+{{- end }}
--- a/charts/dex/templates/secret.yaml
+++ b/charts/dex/templates/secret.yaml
@@ -1,0 +1,51 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+{{ include "dex.labels" . | indent 4 }}
+  name: {{ template "dex.fullname" . }}
+stringData:
+  config.yaml: |-
+    {{- with .Values.config }}
+    issuer: {{ .issuer }}
+    storage:
+{{ toYaml .storage | indent 6 }}
+    logger:
+{{ toYaml .logger | indent 6 }}
+    web:
+      {{- if $.Values.https }}
+      https: {{ $.Values.config.web.address }}:{{ $.Values.ports.web.containerPort }}
+      tlsCert: {{ .web.tlsCert }}
+      tlsKey: {{ .web.tlsKey }}
+      {{- else }}
+      http: {{ $.Values.config.web.address }}:{{ $.Values.ports.web.containerPort }}
+      {{- end }}
+    {{- if $.Values.grpc }}
+    grpc:
+      addr: {{ $.Values.config.grpc.address }}:{{ $.Values.ports.grpc.containerPort }}
+      tlsCert: {{ .grpc.tlsCert }}
+      tlsKey: {{ .grpc.tlsKey }}
+      tlsClientCA: {{ .grpc.tlsClientCA }}
+    {{- end }}
+    {{- if .connectors }}
+    connectors:
+{{ toYaml .connectors | indent 4 }}
+    {{- end }}
+    oauth2: {{ toYaml .oauth2 | nindent 6 }}
+    {{- if .staticClients }}
+    staticClients:
+{{ toYaml .staticClients | indent 4 }}
+    {{- end }}
+    enablePasswordDB: {{ .enablePasswordDB }}
+    {{- if .staticPasswords }}
+    staticPasswords:
+{{ toYaml .staticPasswords | indent 4 }}
+    {{- end }}
+    {{- if .expiry }}
+    expiry:
+{{ toYaml .expiry | indent 6 }}
+    {{- end }}
+    {{- if .frontend }}
+    frontend: {{ toYaml .frontend | nindent 6 }}
+    {{- end }}
+    {{- end }}
--- a/charts/dex/templates/secret.yaml
+++ b/charts/dex/templates/secret.yaml
@@ -31,7 +31,8 @@ stringData:
     connectors:
 {{ toYaml .connectors | indent 4 }}
     {{- end }}
-    oauth2: {{ toYaml .oauth2 | nindent 6 }}
+    oauth2:
+{{ toYaml .oauth2 | indent 6 }}
     {{- if .staticClients }}
     staticClients:
 {{ toYaml .staticClients | indent 4 }}
@@ -46,6 +47,7 @@ stringData:
 {{ toYaml .expiry | indent 6 }}
     {{- end }}
     {{- if .frontend }}
-    frontend: {{ toYaml .frontend | nindent 6 }}
+    frontend:
+{{ toYaml .frontend | indent 6 }}
     {{- end }}
     {{- end }}
--- a/charts/dex/templates/service.yaml
+++ b/charts/dex/templates/service.yaml
@@ -1,0 +1,38 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "dex.fullname" . }}
+  labels:
+{{ include "dex.labels" . | indent 4 }}
+{{- if .Values.service.annotations }}
+  annotations:
+{{ toYaml .Values.service.annotations | indent 4 }}
+{{- end }}
+spec:
+  type: {{ .Values.service.type}}
+  sessionAffinity: None
+  ports:
+  - name: {{ if .Values.https }}https{{ else }}http{{ end }}
+    targetPort: {{ if .Values.https }}https{{ else }}http{{ end }}
+{{- if eq "NodePort" .Values.service.type }}
+    nodePort: {{ .Values.ports.web.nodePort }}
+{{- end }}
+    port: {{ .Values.ports.web.servicePort }}
+{{- if .Values.grpc }}
+  - name: grpc
+    targetPort: grpc
+  {{- if eq "NodePort" .Values.service.type }}
+    nodePort: {{ .Values.ports.grpc.nodePort }}
+  {{- end }}
+    port: {{ .Values.ports.grpc.servicePort }}
+{{- end }}
+{{- if hasKey .Values.service "externalIPs" }}
+  externalIPs:
+{{ toYaml .Values.service.externalIPs | indent 4 }}
+{{- end }}
+{{- if hasKey .Values.service "loadBalancerIP" }}
+  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
+{{- end }}
+  selector:
+    app.kubernetes.io/name: {{ include "dex.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
--- a/charts/dex/templates/service.yaml
+++ b/charts/dex/templates/service.yaml
@@ -33,6 +33,9 @@ spec:
 {{- if hasKey .Values.service "loadBalancerIP" }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
 {{- end }}
+{{- if hasKey .Values.service "clusterIP" }}
+  clusterIP: {{ .Values.service.clusterIP | quote }}
+{{- end }}
   selector:
     app.kubernetes.io/name: {{ include "dex.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
--- a/charts/render.py
+++ b/charts/render.py
@@ -18,10 +18,13 @@ It performs the following tasks:
   `app.kubernetes.io/component` fields
 '''
 
+import re
 import sys
 import subprocess
 
 import yaml
+from yaml.dumper import SafeDumper
+from yaml.representer import SafeRepresenter
 
 
 BOILERPLATE = '''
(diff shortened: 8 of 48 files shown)


charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
+image: '{%- endraw -%}{{ build_image_name(\"dex\", False) }}{%- raw -%}'
+
+nodeSelector:
+  node-role.kubernetes.io/master: ''
```
NicolasT: No, Dex is infra, not control-plane.

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+tolerations:
+  - key: "node-role.kubernetes.io/bootstrap"
+    operator: "Exists"
+    effect: "NoSchedule"
+  - key: "node-role.kubernetes.io/master"
```
NicolasT: I don't think we want a control-plane toleration, only `bootstrap` and `infra`.

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+    operator: "Exists"
+    effect: "NoSchedule"
+
+replicas: 2
+
+# grpc support
+grpc: false
+
+# https termination by dex itself
+https: false
```
NicolasT: Why not?

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+ports:
+  web:
+    containerPort: 5556
```
NicolasT: Do we need to override any defaults?

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+    servicePort: 32000
+  # grpc:
+    # containerPort: 5000
+    # servicePort: 35000
+
+service:
+  type: ClusterIP
```
NicolasT: I'd assume this is the default as well, though didn't check.

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+ingress:
+  enabled: true
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: '/$2'
+    nginx.ingress.kubernetes.io/use-regex: "true"
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
+    kubernetes.io/ingress.class: "nginx-control-plane"
+  path: /dex(/|$)(.*)
```
NicolasT: I'd refrain from using `dex` in the URL (cfr. namespace thing). Consider `/oidc` or such?

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+  hosts:
+    - null
+  tls: []
```
NicolasT: I don't think not setting one will work. The ingress will get a new cert at every start (which is another issue, but anyway), so the cert registered with clients will never be correct, isn't it?

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+  #      - dex.example.com
+
+certs:
+  web:
+    create: false
+  grpc:
+    create: false
+
+rbac:
+  create: true
```
NicolasT: Isn't this the default as well? (Also applies to various other settings in this file)

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+  name:
+
+# ensure dex pods are running only on master nodes
+affinity:
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: "node-role.kubernetes.io/master"
```
NicolasT: Infra. And because we already have a `nodeSelector`, there should be no more affinity defined?

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+    alwaysShowLoginScreen: false
+    skipApprovalScreen: true
+
+  expiry:
+    signingKeys: "6h"
+    idTokens: "24h"
+
+  enablePasswordDB: true
+  staticPasswords:
+    - email: "oidc_admin@metalk8s.com"
```
NicolasT: There is no such domain. Need to find a better value for this, similar to our fake registry address.

charts/dex.yaml
```diff
@@ -0,0 +1,102 @@
...
+      # bcrypt hash of the string "password"
+      hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+      username: "oidc_admin"
```
NicolasT: Why this username?

salt/metalk8s/addons/dex/deployed/namespace.sls
```diff
@@ -0,0 +1,10 @@
+#! kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: metalk8s-auth
+  labels:
+    app.kubernetes.io/managed-by: metalk8s
```
NicolasT: `managed-by` `salt`

NicolasT (dismissed):

charts/dex/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+{{- with .Values.tolerations }}
+      tolerations:
+{{ toYaml . | indent 8 }}
+{{- end }}
+{{- with .Values.affinity }}
+      affinity:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+      containers:
+      - name: main
```
Ebaneck, line 52: @gdemonet this seems to be an issue from the Dex stable helm charts.
Dex Container names are hard-coded to `main`
I think we are not allowed to make changes to this file in the way we deploy using charts.. no ??

NicolasT: Why would this name be an issue in the first place?

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  web:
+    # port is taken from ports section above
+    address: 0.0.0.0
+    # tlsCert: /etc/dex/tls/https/server/tls.crt
+    # tlsKey: /etc/dex/tls/https/server/tls.key
+  connectors: {}
+
+  staticPasswords:
+    - email: "admin@example.com"
```
NicolasT: I commented on this before but the comment seems to be gone?

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+   - secretName: dex-tls-secret
+
+certs:
+  web:
+    create: false
+  grpc:
+    create: false
+
+config:
+  issuer: https://{{ pillar['metalk8s']['api_server']['host'] }}:32000
```
NicolasT: I'm surprised this works without an `endraw`/`raw`.

charts/dex.yaml
Ebaneck: It was not marked resolved yet

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -0,0 +1,139 @@
...
+
+apiVersion: v1
+kind: Secret
+metadata:
+  labels: {app.kubernetes.io/instance: dex, app.kubernetes.io/managed-by: salt, app.kubernetes.io/name: dex,
+    app.kubernetes.io/part-of: metalk8s, app.kubernetes.io/version: 2.19.0, helm.sh/chart: dex-2.4.0,
+    heritage: metalk8s}
+  name: dex
+  namespace: metalk8s-auth
+stringData: {config.yaml: "issuer: https://{{ pillar['metalk8s']['api_server']['host']\
```
NicolasT: As hinted above (in the values file), this ain't right. Even adding `endraw`/`raw` there may make this fail... Not sure how to fix though.

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -0,0 +1,139 @@
...
+  namespace: metalk8s-auth
+spec:
+  rules:
+  - host: null
+    http:
+      paths:
+      - backend: {serviceName: dex, servicePort: 32000}
+        path: /oidc(/|$)(.*)
+  tls:
+  - {hosts: null, secretName: dex-tls-secret}
```
NicolasT: Where is this created?

charts/dex.yaml
NicolasT: Sure, but you did make a change to the line, and as such GitHub removed the comment from the diff view. Which can be considered an issue in the GitHub UI.

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: Working on this, I am going to add a `secret` that would hold `cert` and `key`

charts/dex.yaml
gdemonet: What's this e-mail used for anyway? Could it be `admin@metalk8s.invalid` or something along those lines?

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  web:
+    # port is taken from ports section above
+    address: 0.0.0.0
```
gdemonet: Shouldn't it listen on the control-plane IP only, as suggested in #2011?

charts/dex.yaml
gdemonet: Indeed, even if we expose Dex through an Ingress object, it would make sense to still have it secured and exposed using a `NodePort` (though not entirely sure about that part, cc @NicolasT), as we do for K8s and Salt APIs.

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+tolerations:
+  - key: "node-role.kubernetes.io/bootstrap"
+    operator: "Exists"
+    effect: "NoSchedule"
+  - key: "node-role.kubernetes.io/infra"
+    operator: "Exists"
+    effect: "NoSchedule"
+
+replicas: 2
```
gdemonet, line 14: Why? The default is 1, and the initial deployment is on Bootstrap only. Would make sense to handle this as we do for Prom/AM, i.e. use 1 replica for now, and make it configurable somehow (that'd be part of the #1996 epic I guess).

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+# grpc support
+grpc: false
```
gdemonet, line 17: Didn't we say we'll need gRPC for #1990?

charts/dex.yaml
```diff
@@ -0,0 +1,56 @@
...
+ingress:
+  enabled: true
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: '/$2'
+    nginx.ingress.kubernetes.io/use-regex: "true"
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+    kubernetes.io/ingress.class: "nginx-control-plane"
+  path: /oidc(/|$)(.*)
+  hosts:
+    - null
```
gdemonet, line 32: Control-plane IP here?

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -0,0 +1,139 @@
+#!jinja | kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
+{%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+
+{% raw %}
+
+apiVersion: v1
+kind: Secret
+metadata:
+  labels: {app.kubernetes.io/instance: dex, app.kubernetes.io/managed-by: salt, app.kubernetes.io/name: dex,
```
gdemonet: This display of YAML is unreadable for me, I'd rather use block style consistently. See: https://github.com/scality/metalk8s/pull/2024/commits/3cdd70704250346fc938d778311114414171543b

salt/metalk8s/addons/dex/deployed/chart.sls
gdemonet: `endraw`/`raw` should work, don't see why it wouldn't... Jinja doesn't care about what the supposed format underneath is, IIRC.

charts/dex.yaml
Ebaneck: Not sure we can have an IP address in ingress's host

charts/dex.yaml
NicolasT: Dex is stateless, so having 2 shouldn't harm. Even better, given the `updateStrategy`, there will always be one running even during upgrades and such.

Initially both may run on the `bootstrap` node, fine, but once there are more `infra` ones things can be spread.

So I agree with 2 :)

charts/dex.yaml
NicolasT: Do we really need GRPC? I'm afraid of getting into an authn/authz  mess, whilst access to the configuration/configmap is much easier to manage.

charts/dex.yaml
NicolasT: No no no, we don't want/need a `NodePort`.

However, when we *can* secure in-cluster over-the-wire traffic, we should.

charts/dex.yaml
NicolasT: No, not control plane IP... I'm not even sure `null` is right, maybe the lists needs to be empty?

In any case, we want to be the 'fallback' entry (like we do for UI, Grafana,...) for the ingress controller.

charts/dex.yaml
NicolasT: Huh, no.

This stuff is running in a container, not in the host network, so binding to a control-plane IP (which IP would that be, by the way? Of which node?) can't work.

`0.0.0.0` is fine and will basically bind to `lo` and the only other interface in the container.

However, again, isn't this the default in the chart's `values`, so do we really need to set it here?!

charts/dex.yaml
NicolasT: The e-mail address is used in the OIDC profile of this 'user', and as such can show up e.g. in Grafana.

It should indeed be set to some 'invalid' address.

salt/metalk8s/addons/dex/deployed/chart.sls
NicolasT: It could not work because of the `\`s that get injected when serializing the YAML.

charts/dex.yaml
gdemonet: OK OK OK 😄 No `NodePort` then!

salt/metalk8s/addons/dex/deployed/chart.sls
gdemonet: Maybe with block-style display this wouldn't need to inject `\`s... We need to test, ping @Ebaneck ;)

charts/dex.yaml
gdemonet: OK, I get why having more than one can help for `updateStrategy`. Would still be useful to make it configurable in the long run, IMHO.

charts/dex.yaml
gdemonet: If it's acceptable to restart Dex for each config change, then yes avoid the whole gRPC stuff can be much simpler.

charts/dex.yaml
gdemonet: Unfortunately, this may be hard to do using the Helm chart, given this [template](https://github.com/helm/charts/blob/master/stable/dex/templates/ingress.yaml#L27-L35):

```yaml
  rules:
  {{- range .Values.ingress.hosts }}
    - host: {{ . | quote }}
      http:
        paths:
          - path: {{ $ingressPath }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: {{ $servicePort }}
  {{- end }}
```

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: I can confirm that endraw / raw does not work

charts/dex.yaml
Ebaneck: it's default in the charts.. taking it off

charts/dex/templates/deployment.yaml
Ebaneck: Not a big deal, maybe while listing the running containers, it looked a little inconsistent

charts/dex.yaml
Ebaneck: Yeah,  I am going to enable TLS here

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: Even with block-style display, it injects `\ s..` Any other hints i could try out?? 

Worst case, I add an sls that creates the same secret holding the dex config params

charts/dex.yaml
Ebaneck: Yeah it doen't

salt/metalk8s/addons/dex/deployed/dex-conf.sls
```diff
@@ -0,0 +1,75 @@
...
+    app.kubernetes.io/name: dex
+    app.kubernetes.io/part-of: metalk8s
+    app.kubernetes.io/version: "2.19.0"
+    helm.sh/chart: dex-2.4.0
+    heritage: metalk8s
+  name: dex
+  namespace: metalk8s-auth
+stringData:
+  config.yaml: |-
+      issuer: 'https://{{ oidc_service_ip }}:32000'
```
gdemonet: What would it imply to mention use the Ingress in this config file? Does it only affect what is exposed through `.well-know/openid-configuration`?

```suggestion
      issuer: 'https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc'
```

pillar/metalk8s/roles/ca.sls
```diff
@@ -51,3 +59,10 @@ x509_signing_policies:
     - keyUsage: critical digitalSignature, keyEncipherment
     - extendedKeyUsage: clientAuth
     - days_valid: 365
+  dex_server_policy:
+    - minions: '*'
+    - signing_private_key: /etc/kubernetes/pki/dex-ca.key
+    - signing_cert: /etc/kubernetes/pki/dex-ca.crt
+    - keyUsage: critical digitalSignature, keyEncipherment
+    - extendedKeyUsage: serverAuth, clientAuth
```
TeddyAndrieux: Do we really need `clientAuth` Dex cert is only used as server no ?

charts/dex/templates/clusterrole.yaml
```diff
@@ -9,7 +9,44 @@ rules:
...
+    - clusterroles
+    - roles
+    - events
+  verbs:
+    - get
+    - watch
+    - list
+    - create
+    - patch
+    - update
```
TeddyAndrieux: Wait what ??? Do we really want to have (almost) everything ??
If I well understand it's only stuff doable by the UI no ? So only listing/getting some stuff + maybe some edition no ?

gdemonet: Agree with Teddy, this doesn't make sense. This ClusterRole is for Dex, so that it can use K8s API as its storage, it should have write permissions on such generic resources. Where does this come from? The `clusterrole` from upstream looks better: https://github.com/helm/charts/blob/master/stable/dex/templates/clusterrole.yaml

Ebaneck: yeah, i am going to keep the upstream one

charts/dex.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+service:
+  clusterIP: '{% endraw %}{{ salt.metalk8s_network.get_oidc_service_ip() }}{% raw %}'
+
+ingress:
+  enabled: true
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+    nginx.ingress.kubernetes.io/proxy-ssl-verify: "on"
+    nginx.ingress.kubernetes.io/proxy-ssl-secret: "dex-ca-cert"
```
TeddyAndrieux: I think since we have a default cert for ingress we no longer need this dex CA and server cert

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -4,9 +4,40 @@
...
+{%- endif %}
+
+{%- if bootstrap_id is none %}
+  {{ raise('Missing bootstrap Node in pillar, cannot proceed.') }}
+{%- elif bootstrap_id == grains.id %}
+  {%- set bootstrap_control_plane_ip = grains.metalk8s.control_plane_ip %}
+{%- else %}
+  {%- set bootstrap_control_plane_ip = salt['mine.get'](bootstrap_id,
+                                                        'control_plane_ip') %}
+{%- endif %}
```
TeddyAndrieux: Why we want to use bootstrap ip, that mean if we lose the bootstrap node oidc will no longer work,

Also meaning that we need to update each and every apiserver in restore script

gdemonet: We need a single IP, so a LoadBalancer IP would be ideal, but we don't have a way to create one yet (see #1788).
In the meantime, we use the only available externalIP exposed by the nginx-ingress-control-plane Service, which the Bootstrap's control-plane IP.

Indeed, it does mean that a lot of things will stop working if Bootstrap goes down. The solution is just not available at the moment.

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -4,9 +4,40 @@
 {%- set htpasswd_path = "/etc/kubernetes/htpasswd" %}
 {%- set encryption_k8s_path = "/etc/kubernetes/encryption.conf" %}
 
+{# This whole block is used to "know" the Ingress external IP used by Dex.
```
gdemonet: I split that into a macro, why put it back here? It's a huge block of Jinja code that's not specific to API Server, hindering readability... But not critical, just questioning.

Ebaneck: Yeah, I can add this split back.

The split was done in your PR that adds `Nginx certs`
Reviewer requested it to be removed since it was not used.

buildchain/buildchain/salt_tree.py
```diff
@@ -216,6 +216,10 @@ def _get_parts(self) -> Iterator[str]:
         renderer=targets.Renderer.JSON,
     ),
 
+    Path('salt/metalk8s/addons/dex/deployed/init.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/chart.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/namespace.sls'),
+
```
gdemonet: This is not related to the commit message, which states two things:

- "salt: deploy dex" - I would expect this commit to actually **deploy** Dex using Salt, not just prepare the buildchain
- "This commit adds the dex image to the buildchain which is required for dex deployment" - Then I don't expect anything related to Salt states in there

I think we could settle for a single commit, that handles:

- adding required images in the buildchain
- rendering `chart.sls`
- adding the few dependencies (like `tls-secret.sls`)
- adding those files in the buildchain `salt_tree.py`

The whole "Deploy Dex" feature is thus self-contained in this commit. Fixing the render script or other unrelated stuff should go in other commits (if required by this one, then go before it, if not, after).

charts/dex.yaml
```diff
@@ -0,0 +1,98 @@
+image: '{%- endraw -%}{{ build_image_name(\"dex\", False) }}{%- raw -%}'
```
gdemonet: (unrelated to this line, cannot comment on the commit message)

This commit message states: "add `dex` deployment configuration", but it actually also includes the rendered chart. To me, this commit and the previous one should be squashed.

gdemonet: Also, remove the note about `Secret`  not being rendered properly, as this is fixed in the same PR (just re-order commit to fix it before rendering, and squash all commits including the rendered chart).

salt/metalk8s/addons/dex/deployed/namespace.sls
```diff
@@ -0,0 +1,10 @@
+#!jinja | metalk8s_kubernetes kubeconfig=/etc/kubernetes/admin.conf&context=kubernetes-admin@kubernetes
```
gdemonet, line 1: As it's directly tied to Dex and its deployment, could be squashed with the chart commit.

salt/_modules/metalk8s_network.py
```diff
@@ -68,3 +69,13 @@ def get_cluster_dns_ip():
     range.
     '''
     return _pick_nth_service_ip(COREDNS_ADDRESS_NUMBER)
+
+
+def get_oidc_service_ip():
```
gdemonet, line 74: This is used in the rendered chart and chart values, so either in the same commit, or before it.

salt/_modules/metalk8s_network.py
```diff
@@ -68,3 +69,13 @@ def get_cluster_dns_ip():
...
+    '''
+    Return the OIDC service cluster IP.
+
+    This IP is arbitrarily selected as the seventh IP from the usable hosts
+    range.
+    '''
+    return _pick_nth_service_ip(OIDC_ADDRESS_NUMBER)
```
gdemonet: Missing newline here

salt/_modules/metalk8s_network.py
gdemonet: The commit message should remove the mention about Ingress not being a valid approach, as that's what we use in the end.

charts/dex.yaml
```diff
@@ -1,4 +1,4 @@
-image: '{%- endraw -%}{{ build_image_name(\"dex\", False) }}{%- raw -%}'
```
gdemonet: Can be squashed with the chart commit (no real value in making another commit here, the link was just for you to look at :wink:)

charts/render.py
```diff
@@ -18,10 +18,13 @@
   `app.kubernetes.io/component` fields
 '''
 
+import re
```
gdemonet, line 21: (unrelated to this line, for the commit)

This commit is required before rendering the chart, so let's move it before.
Also, the commit message could be updated to:
- include a reference to the Dex deployment using charts issue (so we know that it's linked somehow)
- detail why multi-line strings are an issue in rendered charts (it's not a problem in YAML per-se, but doesn't work well when Jinja "commands" are split)

charts/dex/templates/secret.yaml
```diff
@@ -31,7 +31,8 @@ stringData:
     connectors:
 {{ toYaml .connectors | indent 4 }}
     {{- end }}
-    oauth2: {{ toYaml .oauth2 | nindent 6 }}
+    oauth2:
```
gdemonet, line 34: This commit needs to come just after importing the chart, before rendering.

charts/dex.yaml
```diff
@@ -32,8 +32,6 @@ ingress:
   path: /oidc
   hosts:
     - null
-  tls:
-   - secretName: dex-tls-ingress
```
gdemonet: No need to make this a dedicated commit, just squash with the chart rendering.

salt/metalk8s/roles/infra/init.sls
```diff
@@ -1,2 +1,2 @@
 include:
-- metalk8s.roles.node
+  - metalk8s.roles.node
```
gdemonet: Unrelated to this PR, let's remove it

salt/metalk8s/salt/master/certs/init.sls
```diff
@@ -8,8 +8,8 @@
 # * salt-api              -> generate Salt API key and certificate
 #
 include:
-  # Some server certs are required to be published in K8s API by Salt master
```
gdemonet, line 11: Why remove the comment? Since these aren't actual certs required by Salt master itself, it explains why they are here.

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -1,12 +1,14 @@
 {%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
 {%- from "metalk8s/map.jinja" import networks with context %}
+{%- from "metalk8s/addons/nginx-ingress-control-plane/control-plane-ip.jinja" import ingress_control_plane with context %}
```
gdemonet: Line is too long to read in GH Web UI, please split at 80 chars or so

salt/metalk8s/addons/dex/deployed/tls-secret.sls
```diff
@@ -0,0 +1,31 @@
...
+kind: Secret
+metadata:
+  name: dex-ca-cert
+  namespace: metalk8s-auth
+type: Opaque
+data:
+  ca.crt: "{{
+    salt['hashutil.base64_encodefile']('/etc/metalk8s/pki/dex/ca.crt')
+    | replace('\n', '')
+  }}"
```
gdemonet: Please remove this altogether, as it's not needed. (not removing it with a commit, it shouldn't appear in history at all)

charts/dex.yaml
```diff
@@ -26,8 +26,6 @@ ingress:
   enabled: true
   annotations:
     nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
-    nginx.ingress.kubernetes.io/proxy-ssl-verify: "on"
-    nginx.ingress.kubernetes.io/proxy-ssl-secret: "dex-ca-cert"
```
gdemonet: Why remove those?

buildchain/buildchain/salt_tree.py
```diff
@@ -216,9 +216,15 @@ def _get_parts(self) -> Iterator[str]:
         renderer=targets.Renderer.JSON,
     ),
 
+    Path('salt/metalk8s/addons/dex/ca/init.sls'),
+    Path('salt/metalk8s/addons/dex/ca/installed.sls'),
+    Path('salt/metalk8s/addons/dex/ca/advertised.sls'),
+    Path('salt/metalk8s/addons/dex/certs/init.sls'),
+    Path('salt/metalk8s/addons/dex/certs/server.sls'),
```
gdemonet, line 223: Dex CA and server certificates are actually part of the Dex deployment, should be squashed with the chart render commit.

However, configuring k-a's OIDC plugin to point to Ingress, that **is** a separate commit.

charts/dex/templates/service.yaml
```diff
@@ -32,6 +32,9 @@ spec:
 {{- end }}
 {{- if hasKey .Values.service "loadBalancerIP" }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
+{{- end }}
+{{- if hasKey .Values.service "clusterIP" }}
+  clusterIP: {{ .Values.service.clusterIP | quote }}
```
gdemonet, line 37: That's a manual fix on a vendored chart. It deserves a dedicated commit, explaining what is done (add support for cluster IP)

salt/metalk8s/addons/dex/ca/installed.sls
```diff
@@ -0,0 +1,42 @@
...
+      - x509: Create dex CA private key
+
+Advertise dex CA certificate in the mine:
+  module.wait:
+    - mine.send:
+      - func: dex_ca_b64
+      - mine_function: hashutil.base64_encodefile
+      - /etc/metalk8s/pki/dex/ca.crt
+    - watch:
+      - x509: Generate dex CA certificate
```
gdemonet: Missing newline

salt/metalk8s/addons/dex/certs/init.sls
```diff
@@ -0,0 +1,2 @@
+include:
+  - .server
```
gdemonet: Missing newline

buildchain/buildchain/salt_tree.py
```diff
@@ -291,6 +291,8 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/addons/nginx-ingress-control-plane/deployed/chart.sls'),
     Path('salt/metalk8s/addons/nginx-ingress-control-plane/deployed/',
          'tls-secret.sls'),
+    Path('salt/metalk8s/addons/nginx-ingress-control-plane/',
+         'control-plane-ip.jinja'),
```
gdemonet: Two things:

- this commit could introduce the helper with the code that uses it, but not critical
- I don't see why it should use the `.jinja` extension, instead of `.sls` - it relies on `pillar` and `grains` (that one could setup manually, sure), so users can't expect it to work with "simple" Jinja templating

salt/metalk8s/addons/nginx-ingress-control-plane/control-plane-ip.jinja
```diff
@@ -0,0 +1,29 @@
...
+  {# Only one node (or even, zero) can/should only happen during bootstrap #}
+  {%- set bootstrap_id = grains.id %}
+{%- else %}
+  {%- set bootstrap_id = none %}
+  {%- for minion_id, node in pillar.metalk8s.nodes.items() %}
+    {%- if 'bootstrap' in node.roles %}
+      {%- set bootstrap_id = minion_id %}
+      {%- break %}
+    {%- endif %}
+  {%- endfor %}
```
gdemonet: This loop could be rewritten with something that we already have (forgot about it):

```jinja
{%- set bootstrap_nodes = salt.metalk8s.minions_by_role('bootstrap') %}
{%- if bootstrap_nodes %}
  {%- set bootstrap_id = bootstrap_nodes | first %}
{%- else %}
  {{ raise('Missing bootstrap node') }}
{%- endif %}
```

gdemonet (changes requested):
Mostly re-ordering of commits, and grouping things logically. The most important one is about splitting out the chart modification for support of `clusterIP`, but gathering "everything that deploys Dex" in a single commit makes sense. API Server configuration is another commit (and it's even another issue, #2010).

charts/dex.yaml
Ebaneck: From the above review, we remove the secret `dex-ca-cert` so it should not be needed here again I think!

buildchain/buildchain/salt_tree.py
```diff
@@ -216,6 +216,16 @@ def _get_parts(self) -> Iterator[str]:
         renderer=targets.Renderer.JSON,
     ),
 
+    Path('salt/metalk8s/addons/dex/deployed/chart.sls'),
+    Path('salt/metalk8s/addons/dex/deployed/namespace.sls'),
```
TeddyAndrieux: Minor: Alphabetic order you shoud move these two with the other `deployed`

salt/metalk8s/addons/nginx-ingress-control-plane/control-plane-ip.sls
```diff
@@ -0,0 +1,28 @@
...
+  {%- endif %}
+{%- endif %}
+
+{%- if bootstrap_id is none %}
+  {{ raise('Missing bootstrap Node in pillar, cannot proceed.') }}
+{%- elif bootstrap_id == grains.id %}
+  {%- set bootstrap_control_plane_ip = grains.metalk8s.control_plane_ip %}
+{%- else %}
+  {%- set bootstrap_control_plane_ip = salt['mine.get'](bootstrap_id,
+                                                        'control_plane_ip') %}
```
TeddyAndrieux: Here you have an issue you will get a dict contains control_plane_ip of each and every node so if you want to do it you need to take only the bootstrap one so

```
  {%- set bootstrap_control_plane_ip = salt['mine.get'](bootstrap_id,
                                                        'control_plane_ip')[bootstrap_id] %}
```

charts/dex.yaml
gdemonet: Oh right that's why I left the CA cert... My bad. Let's leave it as-is for now, open a debt ticket to set this up (and test it).

charts/dex/templates/service.yaml
```diff
@@ -35,4 +35,4 @@ spec:
 {{- end }}
   selector:
     app.kubernetes.io/name: {{ include "dex.name" . }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
```
gdemonet: There's no reason to do this

buildchain/buildchain/constants.py
```diff
@@ -19,6 +19,7 @@
 # URLs of the main container repositories.
 CALICO_REPOSITORY     : str = 'quay.io/calico'
 COREOS_REPOSITORY     : str = 'quay.io/coreos'
+DEX_REPOSITORY        : str = 'quay.io/dexidp'
```
gdemonet, line 22: Squash this with the "Deploy Dex" commit, there's no need to have a separate commit (if you reverted "Deploy Dex', then this commit still present would mean we ship unneeded container images).

charts/dex/templates/service.yaml
```diff
@@ -32,6 +32,9 @@ spec:
 {{- end }}
 {{- if hasKey .Values.service "loadBalancerIP" }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
+{{- end }}
```
gdemonet, line 35: Commit message talks about the actual Dex cluster IP being the 7th Service IP. This is true, but not related to this commit: explanation should be added to the "Deploy Dex" commit.

charts/dex.yaml
```diff
@@ -0,0 +1,94 @@
...
+    redirectURIs:
+    - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{% raw %}'
+    secret: "4lqK98NcsWG5qBRHJUqYM1"
+
+  staticPasswords:
+    - email: "admin@metalk8s.invalid"
+      # bcrypt hash of the string "password"
+      hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+      username: "admin"
+      userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
```
gdemonet: Missing newline

salt/metalk8s/addons/dex/deployed/clusterrolebinding.sls
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: dex-administrator
+subjects:
+- kind: User
+  name: "admin@metalk8s.invalid"
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
```
gdemonet: This file could have been added in another commit, or even another PR. But let's leave it, not critical.

Also, missing newline.

gdemonet (changes requested):
FYI, Dex is not an acronym, so you shouldn't write it in full-caps.

gdemonet (dismissed):
LGTM - added a wait to not merge this in 2.4.

In the meantime (waiting for 2.5 branch), could you please:
- start adding some tests (e.g. try to reach the openid-configuration and check the issuer_url)
- investigate SSL verification from the Ingress (the thing that we removed) - does it work? or is it just ignored?

TeddyAndrieux (approved):

PR #2032, opened 2019-11-12
1952: flaky downgrade/upgrade

alexandre-allard:
**Component**: salt

**Context**: flaky upgrade / downgrade tests

**Summary**: Salt is checking every X seconds the liveness of a running job, sometimes it sends the request during the minion restart, so the query fails & the master assumes that the job has also failed.

**Acceptance criteria**: green build

### DIFF
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -74,6 +74,23 @@ Refresh pillar before highstate:
     - name: saltutil.refresh_pillar
     - tgt: {{ node_name }}
 
+{%- if node_name in salt.saltutil.runner('manage.up') %}
+
+Reconfigure salt-minion:
+  salt.state:
+    - tgt: {{ node_name }}
+    - saltenv: metalk8s-{{ version }}
+    - sls:
+      - metalk8s.salt.minion.configured
+    - require:
+      - salt: Set grains
+      - salt: Refresh the mine
+      - salt: Refresh pillar before highstate
+    - require_in:
+      - salt: Run the highstate
+
+{%- endif %}
+
 Run the highstate:
   salt.state:
     - tgt: {{ node_name }}


TeddyAndrieux (dismissed):
Great LGTM

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -79,6 +79,9 @@ Run the highstate:
     - tgt: {{ node_name }}
     - highstate: True
     - saltenv: metalk8s-{{ version }}
+    {#- we set a timeout of 45s before checking for job liveness, otherwise
+        we can end up checking the job while minion is restarting #}
+    - timeout: 45
```
gdemonet: What happens if the highstate completes in less than 45 seconds?

gdemonet (commented):
Small question, but could be troublesome if that slows down _real_ deployments

salt/metalk8s/orchestrate/deploy_node.sls
alexandre-allard: The result will not be read by the master before the 45 seconds.
Maybe we could lower down the timeout, but with the risk of bringing back some flakies.

alexandre-allard: Otherwise, as talked offline, we could move the minion configuration & the restart oustide of the highstate.

TeddyAndrieux: > Otherwise, as talked offline, we could move the minion configuration & the restart oustide of the highstate.

We can "pre" reconfigure the minion before the highstate but the minion configuration should be part of the highstate anyway (but indeed if we pre-reconfigure the minion normally the minion will not restart in the highstate).
For me increasing the timeout here is good enough, pretty sure we will not win time by doing a call to pre-reconfigure salt minion instead of increase timeout here

gdemonet: > We can "pre" reconfigure the minion before the highstate but the minion configuration should be part of the highstate anyway (but indeed if we pre-reconfigure the minion normally the minion will not restart in the highstate).

Yes that seems like a good idea, just increase the timeout on the expected minion restart (while with highstate, we don't know _when_ the minion will restart, as it depends on ordering). That won't fix the highstate if run manually, but should be equivalent to the current approach.

> For me increasing the timeout here is good enough, pretty sure we will not win time by doing a call to pre-reconfigure salt minion instead of increase timeout here

I don't think the goal is to win time, but mostly to make sure it's not flaky. What if we need to add a bunch of states **before** the minion restart in the highstate? That timeout may become too small...

gdemonet:
@alexandre-allard-scality Don't we need to do the same in `upgrade`, `downgrade` and `bootstrap` orchestrations as well?

TeddyAndrieux:
> @alexandre-allard-scality Don't we need to do the same in `upgrade`, `downgrade` and `bootstrap` orchestrations as well?

`upgrade` and `downgrade` use the `deploy_node` sls, bootstrap may not have the issue as we configure the salt-minion before running the highstate

gdemonet:
> `upgrade` and `downgrade` use the `deploy_node` sls,

:+1:

> bootstrap may not have the issue as we configure the salt-minion before running the highstate

OK, wouldn't it make sense to do the same in the orchestration then? Related to your suggestion for "pre-configuring" the minion, we'll have better control over the minion restart (and could even improve the `timeout` method with some retry in the orchestration itself).

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: From my understanding `timeout` is just the max time to wait before getting a minion status for a job so it should not change even if we add a bunch of states in the highstate. It may change if for what ever reason a minion take more time to restart

TeddyAndrieux:
> OK, wouldn't it make sense to do the same in the orchestration then? Related to your suggestion for "pre-configuring" the minion, we'll have better control over the minion restart (and could even improve the `timeout` method with some retry in the orchestration itself).

Yes may make sense

alexandre-allard:
I'll do some tests to reconfigure salt-minion before running the highstate and change this PR if it solves the issue.

salt/metalk8s/orchestrate/deploy_node.sls
gdemonet: Oh, that was not my understanding (based on @alexandre-allard-scality explanations, so I'm not 100% certain). To me, `timeout` is actually a "delay" to observe before polling _the_ job (here, the  `state.highstate` job), not all "sub-jobs". If for some reason the minion is still restarting 45 seconds after the request for running `highstate` was sent, the master will poll the job and hit the same issue. If we can guarantee that restarting the minion is the first thing we do in the `highstate`, then it's not necessary to split them.

alexandre-allard: Actually, it's used for both things:
The master waits `timeout` seconds before firing `find_job`, then waits another `timeout` seconds for the answer.

TeddyAndrieux (approved):

PR #2050, opened 2019-11-15
Improvement/1099 set kubeconfig in master config

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Context**: 

We need to have a `kubeconfig` to interact with kubernetes API, before this PR we give the `kubeconfig` to each and every call we do.

**Summary**:

Store the `kubeconfig` and `context` directly in the pillar and salt-master config

**Acceptance criteria**: 

Working bootstrap, upgrade, downgrade

---

Closes: #1099 
Closes: #1100

### DIFF
--- a/salt/_modules/metalk8s_drain.py
+++ b/salt/_modules/metalk8s_drain.py
@@ -228,8 +228,7 @@ class Drain(object):
             kind=controller_ref['kind'],
             apiVersion=controller_ref['api_version'],
             namespace=namespace,
-            kubeconfig=self._kwargs.get('kubeconfig'),
-            context=self._kwargs.get('context')
+            **self._kwargs
         )
 
     def get_pod_controller(self, pod):
@@ -303,8 +302,7 @@ class Drain(object):
             apiVersion='v1',
             all_namespaces=True,
             field_selector='spec.nodeName={0}'.format(self.node_name),
-            kubeconfig=self._kwargs.get('kubeconfig'),
-            context=self._kwargs.get('context')
+            **self._kwargs
         )
 
         for pod in all_pods:
@@ -399,8 +397,7 @@ class Drain(object):
                 name=pod['metadata']['name'],
                 namespace=pod['metadata']['namespace'],
                 grace_period=self.grace_period,
-                kubeconfig=self._kwargs.get('kubeconfig'),
-                context=self._kwargs.get('context')
+                **self._kwargs
             )
 
         pending = self.wait_for_eviction(pods)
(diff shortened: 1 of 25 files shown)


NicolasT:
Do we really need to thread a `context` through? Can't we create a `KubeConfig` with the `current-context` set to the 'right' value (likely only one), and use that?

NicolasT (dismissed):
See question

TeddyAndrieux:
> Do we really need to thread a `context` through? Can't we create a `KubeConfig` with the `current-context` set to the 'right' value (likely only one), and use that?

Yes, we can basically just remove the context. `admin.conf` already have the right `current-context` set, even before this PR `context` was not necessary.

salt/_modules/metalk8s_drain.py
```diff
@@ -228,8 +228,7 @@ def get_controller(self, namespace, controller_ref):
             kind=controller_ref['kind'],
             apiVersion=controller_ref['api_version'],
             namespace=namespace,
-            kubeconfig=self._kwargs.get('kubeconfig'),
-            context=self._kwargs.get('context')
+            **self._kwargs
```
gdemonet, line 231: I thought you didn't want to do this because `self._kwargs` are the `Drain` object's kwargs? Minor though, feel free to leave them.

salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -40,7 +40,42 @@ def __virtual__():
...
+
+    .. code-block:: python
+
+        kubeconfig, context = __salt__['metalk8s_kubernetes.get_kubeconfig'](**kwargs)
+    """
+    pillar_dict = __pillar__.get('metalk8s', {}).get('api_server', {})
+
+    kubeconfig = kwargs.get('kubeconfig') or \
+        pillar_dict.get('kubeconfig') or \
+        __salt__['config.option']('kubernetes.kubeconfig')
```
gdemonet, line 70: Don't you have access to `__opts__`? Or is that unrelated?

gdemonet (dismissed):
Minor comments, LGTM

salt/_modules/metalk8s_drain.py
TeddyAndrieux: But since we accept kwargs everywhere I think it's ok, at the beginning in kubernetes module we didn't have any kwargs I added it at the end (after your first comment so btw I should have done this before)

salt/_modules/metalk8s_kubernetes_utils.py
TeddyAndrieux: We have access to `__opts__` just the initial `kubernetes` module from salt use this options so just kept it

gdemonet (approved):

PR #2057, opened 2019-11-19
Deploy prometheus-adapter for metrics.k8s.io to be available

NicolasT:
Deploy and configure `prometheus-adapter` such that `kubectl top nodes` and `kubectl top pods` works.


NicolasT:
~~Want to add some test(s) before moving forward with this.~~ Tests implemented.

NicolasT:
Turns out due to changes in `cAdvisor` in Kubernetes 1.16, this didn't work as-is on `development/2.5` (cfr. https://stackoverflow.com/a/58572044/1285443)

Changing the Prometheus queries fixes this, and is luckily backwards-compatible with 1.15/2.4.

alexandre-allard (dismissed):
I tried an install with this on 2.4.2, it worked like a charm.
I was able to do some kubectl top nodes & pods
LGTM.

alexandre-allard (dismissed):

gdemonet (approved):

PR #2061, opened 2019-11-21
Test bootstrap restore in CI

### LINKED ISSUE #1687: Add backup/restore test in the CI
**Component**:

'tests'

**Why this is needed**:

Currently backup is launch at the end of each bootstrap but restore is never tested

**What should be done**:

Add a test for bootstrap recovery

Spawn a metalk8s cluster retrieve the backup file from`/var/lib/metalk8s/` destroy the bootstrap node, spawn a new machine and use it to spawn a new bootstrap node

alexandre-allard:
**Component**: ci, tests

**Context**: 

**Summary**:

---

Closes: #1687

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -951,7 +951,8 @@ stages:
           name: Run slow tests locally
           env:
             <<: *_env_local_tests
-            PYTEST_FILTERS: "post and ci and not multinode and slow"
+            PYTEST_FILTERS: >-
+              post and ci and not multinode and slow and not restore
       - ShellCommand:
           name: Cypress test
           command: bash cypress.sh
@@ -1626,7 +1627,8 @@ stages:
           env:
             <<: *_env_bastion_tests
             PYTEST_ARGS: "--suppress-no-test-exit-code"
-            PYTEST_FILTERS: "post and ci and slow and not bootstrap"
+            PYTEST_FILTERS: >-
+              post and ci and slow and not bootstrap and not restore
       - ShellCommand:
           <<: *generate_report_over_ssh
           env:
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -537,6 +537,7 @@ stages:
             - single-node-upgrade-centos
             - single-node-downgrade-centos
             - single-node-patch-version
+            - bootstrap-restore
           waitForFinish: True
       - ShellCommand: *add_final_status_artifact_success
       - Upload: *upload_final_status_artifact
@@ -1459,7 +1460,7 @@ stages:
         CLASS_NAME: multi node.centos7
         TEST_NAME: 1 bootstrap 1 master,etcd
     simultaneous_builds: 20
-    worker:
+    worker: &multi_node_worker
       <<: *single_node_worker
       flavor: m1.medium
       path: eve/workers/openstack-multiple-nodes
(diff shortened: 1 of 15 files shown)


gdemonet:
Great effort, seems like you encountered much more problems than anticipated. However, I think we don't want to keep adding things to Eve specifics, but instead extend the work that was done in `tests/install` for cluster expansion, and/or move more logic into Terraform and make this accessible to users. Anyway, going from this to something more generic shouldn't be too hard now that you solved most issues.

alexandre-allard-scality:
@gdemonet Are you talking only about the node expansion or is there something else you want me to change ?

eve/main.yml
```diff
@@ -1315,6 +1317,103 @@ stages:
...
+          haltOnFailure: true
+      - ShellCommand:
+          name: Recreate the bootstrap node
+          env: *_env_terraform_multiple_nodes
+          command:
+            terraform apply -auto-approve -refresh
+            -target openstack_compute_instance_v2.bootstrap
+            -target null_resource.ssh_config
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
```
TeddyAndrieux: Maybe could be simpler/better to install `etcdctl` on the bastion and directly do the member list and remove from here using the right endpoint

TeddyAndrieux (dismissed):
Small comment but LGTM

eve/main.yml
```diff
@@ -1597,7 +1597,7 @@ stages:
           name: Run installation scenarii on the bastion
           env:
             <<: *_env_bastion_tests
-            PYTEST_FILTERS: "install and ci and multinodes"
+            PYTEST_FILTERS: "install and ci and multinodes and not node2"
```
gdemonet, line 1598: Beurk. Let's keep it, but really not what BDD tags should be used for, IMHO

eve/main.yml
```diff
@@ -448,6 +448,17 @@ models:
...
+      name: Send bastion public key to nodes
+      command: >
+        for host in $HOSTS_LIST; do
+          ssh -F ssh_config bastion "cat .ssh/bastion.pub" |
+          ssh -F ssh_config $host "cat >> .ssh/authorized_keys"
+        done
+      env:
+        HOSTS_LIST: bootstrap
+      workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+
```
gdemonet, line 461: Why not do this in Terraform?

eve/main.yml
```diff
@@ -1658,6 +1659,193 @@ stages:
...
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
+      - ShellCommand:
+          name: Retrieve the backup archive from the bootstrap node
+          command: >
+            scp -F ssh_config
+            bootstrap:"/var/lib/metalk8s/%(prop:bootstrap_backup_archive)s"
+            /tmp
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
```
gdemonet, line 1735: Would have been simpler in practice to just scp from Bootstrap to Node 2 directly ;)

eve/main.yml
```diff
@@ -1658,6 +1659,193 @@ stages:
...
+      - ShellCommand:
+          name: Remove bootstrap node object
+          command: >
+            ssh -F ssh_config node1
+            sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf delete
+            node --selector="node-role.kubernetes.io/bootstrap"
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
```
gdemonet, line 1785: Why isn't all this orchestrated in the test? Looks like it's not in "the right place"...

gdemonet (approved):
Nothing preventing the merge, but Terraform, our tests, and Eve's YAML need to be cleaned-up and refactored. For the future :)

eve/main.yml
TeddyAndrieux: Why ? We create a new node for bootstrap we do not use the node 2

eve/main.yml
TeddyAndrieux: Yes but we do not have better solution here for expansion right now I think

eve/main.yml
TeddyAndrieux: It was not done before but yes maybe this could go in terraform

eve/main.yml
TeddyAndrieux: Yes could be part of the pytest test directly but then we need to do it after spawning the new bootstrap node, anyway yes it's doable

TeddyAndrieux (approved):

eve/main.yml
gdemonet: Ah right my bad.

PR #2066, opened 2019-11-22
UI: Bugfix - provision volumes with labels

### LINKED ISSUE #1891: UI: Provision volumes with labels (in the PV template)
After bootstrap, a user must create *Volume*s to provide storage for Prometheus and AlertManager. We deploy a *StorageClass* for this, but the *PersistentVolumeClaim*s for these services also use a `labelSelector` to pick the 'right' *PersistentVolume*.

It's currently not possible to specify these when creating a *Volume* through the UI. As such, one can't complete the installation using the UI only, and needs to resort to using `kubectl` (e.g. by editing and applying the example manifest shipped on the ISO), which is sub-optimal.

carlito-scality:
**Component**: UI 

**Context**: 
- We need to add labels in the templates instead of the Volume object itself when creating volumes so labels can be propagated to PersistentVolume.

**Summary**:

**Acceptance criteria**: 

Screenshot 2019-11-22 at 10 08 17

---

Closes: https://github.com/scality/metalk8s/issues/1891

### DIFF
--- a/ui/src/ducks/app/volumes.js
+++ b/ui/src/ducks/app/volumes.js
@@ -217,6 +217,11 @@ export function* createVolumes({ payload }) {
     spec: {
       nodeName,
       storageClassName: newVolume.storageClass,
+      template: {
+        metadata: {
+          labels: newVolume.labels,
+        },
+      },
     },
   };
 
--- a/ui/cypress.json
+++ b/ui/cypress.json
@@ -4,8 +4,10 @@
     "username": "admin",
     "password": "admin",
     "device_path": "/dev/disk1",
-    "volume_capacity": "1",
-    "storage_class":"metalk8s-prometheus"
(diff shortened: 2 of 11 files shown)


ChengYanJin (dismissed):
LGTM

ui/src/ducks/app/volumes.js
```diff
@@ -212,11 +212,15 @@ export function* createVolumes({ payload }) {
     kind: 'Volume',
     metadata: {
       name: newVolume.name,
-      labels: newVolume.labels,
```
gdemonet: No, keep it

gdemonet (changes requested):

ChengYanJin (approved):

gdemonet (approved):

PR #2042, opened 2019-11-13
UI: Dex integration

carlito-scality:
**Component**: UI 

**Context**: 
- Integrate Dex login using redux-oidc library

**Summary**:

**Acceptance criteria**: 
- The MetalK8s UI should directly use dex to authenticate a user using OAuth2 flow.
- Once logged in, the reste should works like before without regressions

---

Closes: https://github.com/scality/metalk8s/issues/2015

Issue #1528, opened 2019-08-19
Limit which Salt states can be executed by the storage-operator


slaperche-scality:
**Component**:

salt

**Why this is needed**:

Currently the storage-operator is allowed to execute `state.sls` which is too wide (it can run any kind of states).
Processes should only have the minimum permissions to allow them to do their job, not more.

**What should be done**:

Update the ACL for the `storage-operator` account in `salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2`.

**Implementation proposal** (strongly recommended):

We should limit which states can be given to `state.sls`, using [this feature](https://docs.saltstack.com/en/latest/topics/eauth/index.html#limiting-by-function-arguments).

**Test plan**:

Try to run arbitrary Salt state while using storage-operator credentials: it should fails.

PR #2068, opened 2019-11-22
salt/auth: Add groups support for bearer auth

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

Currently we have no permissions when using bearer token auth

**Summary**:

add a `groups` function to authenticate bearer auth using the same approach as
the basic auth for the moment
Also update test accordingly

### DIFF
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -58,6 +58,40 @@ def _check_k8s_creds(kubeconfig, token):
         raise
 
 
+def _check_node_admin(kubeconfig):
+    client = kubernetes.client.ApiClient(configuration=kubeconfig)
+
+    authz_api = kubernetes.client.AuthorizationV1Api(api_client=client)
+
+    result = authz_api.create_self_subject_access_review(
+        body=kubernetes.client.V1SelfSubjectAccessReview(
+            spec=kubernetes.client.V1SelfSubjectAccessReviewSpec(
+                resource_attributes=kubernetes.client.V1ResourceAttributes(
+                    resource='nodes',
+                    verb='*',
+                ),
+            ),
+        ),
+    )
+
+    return result.status.allowed
+
+
+AVAILABLES_GROUPS = {
+    'node-admins': _check_node_admin
+}
+
+
+def _get_groups(kubeconfig):
+    groups = set()
+
+    for group, func in AVAILABLES_GROUPS.items():
+        if func(kubeconfig):
+            groups.add(group)
+
+    return list(groups)
+
+
 @_log_exceptions
 def _auth_basic(kubeconfig, username, token):
     decoded = base64.decodestring(token)
@@ -86,27 +120,7 @@ def _groups_basic(kubeconfig, username, token):
     kubeconfig.cert_file = None
     kubeconfig.key_file = None
 
-    client = kubernetes.client.ApiClient(configuration=kubeconfig)
-
-    authz_api = kubernetes.client.AuthorizationV1Api(api_client=client)
-
-    groups = set()
-
-    result = authz_api.create_self_subject_access_review(
-        body=kubernetes.client.V1SelfSubjectAccessReview(
-            spec=kubernetes.client.V1SelfSubjectAccessReviewSpec(
-                resource_attributes=kubernetes.client.V1ResourceAttributes(
-                    resource='nodes',
-                    verb='*',
-                ),
-            ),
-        ),
-    )
-
-    if result.status.allowed:
-        groups.add('node-admins')
-
-    return list(groups)
+    return _get_groups(kubeconfig)
 
 
 AUTH_HANDLERS['basic'] = {
@@ -114,13 +128,33 @@ AUTH_HANDLERS['basic'] = {
     'groups': _groups_basic,
 }
 
+
 @_log_exceptions
 def _auth_bearer(kubeconfig, username, token):
     return _check_k8s_creds(kubeconfig, 'Bearer {}'.format(token))
 
+
+@_log_exceptions
+def _groups_bearer(kubeconfig, _username, token):
+    kubeconfig.api_key = {
+        'authorization': token,
+    }
+    kubeconfig.api_key_prefix = {
+        'authorization': 'Bearer',
+    }
+    kubeconfig.username = None
+    kubeconfig.password = None
+    kubeconfig.cert_file = None
+    kubeconfig.key_file = None
+
+    return _get_groups(kubeconfig)
+
+
 AUTH_HANDLERS['bearer'] = {
     'auth': _auth_bearer,
+    'groups': _groups_bearer
 }
+
 
 @_log_exceptions
 def _load_kubeconfig(opts):
--- a/tests/post/features/salt_api.feature
+++ b/tests/post/features/salt_api.feature
@@ -9,6 +9,15 @@ Feature: SaltAPI
         And we have '@runner' perms
         And we have '@jobs' perms
 
+    Scenario: Login to SaltAPI using an admin ServiceAccount
+        Given the Kubernetes API is available
+        When we login to SaltAPI with an admin ServiceAccount
+        Then we can ping all minions
+        And we can invoke '[".*"]' on '*'
+        And we have '@wheel' perms
+        And we have '@runner' perms
+        And we have '@jobs' perms
+
     Scenario: Login to SaltAPI using a ServiceAccount
         Given the Kubernetes API is available
         When we login to SaltAPI with the ServiceAccount 'storage-operator'
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -102,6 +102,111 @@ def k8s_client(request, k8s_apiclient):
     return api_cls(api_client=k8s_apiclient)
 
 
+@pytest.fixture
+def admin_sa(k8s_client, k8s_apiclient):
+    """Fixture to create a ServiceAccount which is bind to `cluster-admin`
+    ClusterRole and return the ServiceAccount name
+    """
+    rbac_k8s_client = kubernetes.client.RbacAuthorizationV1Api(
+        api_client=k8s_apiclient
+    )
+    sa_name = 'test-admin'
+    sa_namespace = 'default'
+    sa_manifest = {
+        'apiVersion': 'v1',
+        'kind': 'ServiceAccount',
+        'metadata': {
+            'name': sa_name,
+            'namespace': sa_namespace
+        }
+    }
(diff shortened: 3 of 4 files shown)


tests/post/features/salt_api.feature
```diff
@@ -12,7 +12,10 @@ Feature: SaltAPI
     Scenario: Login to SaltAPI using a ServiceAccount
         Given the Kubernetes API is available
         When we login to SaltAPI with the ServiceAccount 'storage-operator'
-        Then we can invoke '["disk.dump", "state.sls"]' on '*'
+        Then we can ping all minions
+        And we can invoke '[".*"]' on '*'
+        And we have '@wheel' perms
+        And we have '@runner' perms
```
gdemonet: Uh, really? Should it?

salt/_auth/kubernetes_rbac.py
```diff
@@ -114,14 +114,54 @@ def _groups_basic(kubeconfig, username, token):
...
+    result = authz_api.create_self_subject_access_review(
+        body=kubernetes.client.V1SelfSubjectAccessReview(
+            spec=kubernetes.client.V1SelfSubjectAccessReviewSpec(
+                resource_attributes=kubernetes.client.V1ResourceAttributes(
+                    resource='nodes',
+                    verb='*',
+                ),
+            ),
+        ),
+    )
```
gdemonet: Could be worth factorizing with the code in `_groups_basic`

gdemonet (changes requested):

tests/post/features/salt_api.feature
gdemonet: Because I don't think the Storage Operator needs to have `*` perms on Nodes

TeddyAndrieux: > Should it?

I think no, but since we use Bearer token for storage-operator ... it is.
Since we have no "roles" we can only have every perms or None for the moment

> Because I don't think the Storage Operator needs to have `*` perms on Nodes

I agree

salt/_auth/kubernetes_rbac.py
TeddyAndrieux: Since this one should change in the futur (and I hope before 2.5.0 release), to handle different roles I think it's not really needed

tests/post/features/salt_api.feature
slaperche-scality: Maybe related to https://github.com/scality/metalk8s/issues/1528

TeddyAndrieux: Hmm yes true we have some rules in master config maybe something can be done.
Honnestly I don't really searched how our auth module work, just did a quick fix to have `groups` for Bearer auth.

salt/_auth/kubernetes_rbac.py
gdemonet: Disagree. The "node-admins" group for Salt will still exist I guess, and even if the way to know if a user is in that group changes in the future, there's common logic between Basic and Bearer tokens.

tests/post/features/salt_api.feature
gdemonet: In master config, we have two "groups". One for node-admins, another for storage-operator. The SA for the Operator shouldn't have permissions that make it end up in the node-admins group.

> Maybe related to #1528

No, that one would be nice as well, but that's about reducing the permissions of the storage-operator group. Here, this test effectively confirms that we have the node-admins group, and that shouldn't be the case.

salt/_auth/kubernetes_rbac.py
NicolasT: Agree with @gdemonet. The mapping of K8s rights to Salt groups is intricate, and should be *the same* whether you're using Basic auth, Bearer tokens or whatever. The only difference is the mangling of the `kubeconfig` object (which is an ugly hack anyway).

salt/_auth/kubernetes_rbac.py
```diff
@@ -114,14 +114,54 @@ def _groups_basic(kubeconfig, username, token):
...
+
+@_log_exceptions
+def _groups_bearer(kubeconfig, username, token):
+    kubeconfig.api_key = {
+        'authorization': token,
+    }
+    kubeconfig.api_key_prefix = {
+        'authorization': 'Bearer',
+    }
+    kubeconfig.username = username
```
NicolasT: This shouldn't be set I believe, this information is contained in the token.

tests/post/features/salt_api.feature
NicolasT: Note: `storage-operator` is not a *group*, it's a SaltAPI *user*. Only `node-admins@` is a *group*.

tests/post/features/salt_api.feature
```diff
@@ -12,7 +12,10 @@ Feature: SaltAPI
     Scenario: Login to SaltAPI using a ServiceAccount
         Given the Kubernetes API is available
         When we login to SaltAPI with the ServiceAccount 'storage-operator'
-        Then we can invoke '["disk.dump", "state.sls"]' on '*'
+        Then we can ping all minions
```
NicolasT, line 15: Basically, the fact that this test requires changes shows that there's an issue somewhere in the implementation. This test should remain as-is.

TeddyAndrieux: Btw this test do not require change but since `storage-opeator` is able to do a `*` on `nodes` he also have this perms.

Basically storage-operator perms in this PR are:
```
        "perms": [
            {
                "*": [
                    "disk.dump", 
                    "state.sls"
                ]
            }, 
            "@jobs", 
            {
                "*": [
                    ".*"
                ]
            }, 
            "@wheel", 
            "@runner", 
            "@jobs"
        ], 
```

tests/post/features/salt_api.feature
TeddyAndrieux: Ok just investigate, it's not really easy just retrieve the id_token from dex. (if you have idea do not hesitate)

So several choice for this PR:
- We test with `storage-operator` token as this specific user has the right perms (see https://github.com/scality/metalk8s/pull/2068/files#r351360140)
- We create a new service account just to do this test
- We not test in this PR as this full auth process will be tested by cypress test in #2042

gdemonet: I'm getting confused. Why are we talking about Dex? Anything that runs in the cluster can rely on ServiceAccounts, and that's what `storage-operator` uses (or should use) as well.

See:
https://github.com/scality/metalk8s/blob/development/2.4/storage-operator/deploy/role.yaml
https://github.com/scality/metalk8s/blob/development/2.4/storage-operator/deploy/service_account.yaml

TeddyAndrieux: Since we use dex we can authenticate without service account using directly dex that will give you a token for a user which is bind in `dex-administrator` cluster role binding. But I didn't found any simple solution to retrieve this token (and also it's not really part of salt-api stuff)

I know it's a service account that my point, 
So we need to create a new service account for this test or use the storage-operator one.

salt/_auth/kubernetes_rbac.py
```diff
@@ -58,6 +58,30 @@ def _check_k8s_creds(kubeconfig, token):
         raise
 
 
+def _get_groups(kubeconfig):
```
gdemonet: Works for now, since we only support one group, but I would have imagined something that checks whether to add the 'node-admins' group and only return this information, and be able to call a bunch of such single-focus checks to compute the list/set of groups for a token.

gdemonet (dismissed):

tests/post/features/salt_api.feature
gdemonet: But it shouldn't have those perms?!? https://github.com/scality/metalk8s/blob/development/2.4/storage-operator/deploy/role.yaml

tests/post/features/salt_api.feature
gdemonet: > we need to create a new service account for this test or use the storage-operator one.

I don't understand the problem. In the Gherkin, we have:

```
When we login to SaltAPI with the ServiceAccount 'storage-operator'
```

I would then expect that's already what the test does. We shouldn't try to authenticate without a SA from those tests (yet, that's something to do, for sure, but will take some time to implement, with the whole OAuth dance).

tests/post/features/salt_api.feature
TeddyAndrieux: As the cluster role say that he can do `*` on nodes the check we do for groups is approved so we add node-admin groups thats why but if we dont want that the storage operator can do this then we need to either change the check from groups function or changes the rule of his cluster role

salt/_auth/kubernetes_rbac.py
TeddyAndrieux: Ok make sense i will update it

tests/post/features/salt_api.feature
TeddyAndrieux: Yes but storage operator is a special case as this user is define directly in the salt-master it may make sense to have a new SA to test the groups function for bearer auth

salt/_auth/kubernetes_rbac.py
TeddyAndrieux: Done

tests/post/features/salt_api.feature
NicolasT: Ah, right, I think I get it now.

So indeed, there is currently no test to check whether *with a given token* for a user that's *not* `storage-operator` but *does* have the rights to do `*` on `Nodes` would get assigned to the 'right' group (i.e., `node-admins`).

So, yes, maybe creating a SA which then gets a `ClusterRoleBinding` to `system:masters`, then check whether we get the expected access to SaltAPI when using this SA's token would work.

No need for Dex or whatnot, so could (should?) be part of this PR.

tests/post/features/salt_api.feature
gdemonet: Oh right, I read the Role too fast. Then that would be a bug I guess, we shouldn't give that many permissions to `storage-operator`.

gdemonet: From [this issue](https://github.com/operator-framework/operator-sdk/issues/1226), it seems we should not have kept the default ClusterRole that gets created when the project is initialized with `operator-sdk new`. Since `operator-sdk generate k8s` doesn't generate this file, it would make sense to update it to match our needs and commit it.

gdemonet: See #2084

TeddyAndrieux: Then it's not part of this PR, because we may want to fix it also in 2.4

carlito-scality (approved):

PR #2069, opened 2019-11-23
Rework management of Salt jobs in the UI

gdemonet:
**Component**: ui

**Context**:

As the UI will need to handle more and more long-running Salt operations (such as importing a Solution or preparing an Environment), the pattern introduced for deployment of Nodes needs to be extended.

**Summary**:

- extract all logic related to management of Salt jobs from `app.nodes` to `app.salt`
- refine the pattern to make it more generic: a deployment job is now named `deploy-node/<node-name>` to avoid future collisions
- store the completion time of a job in localStorage, to provide efficient reloading of the application
- keep jobs in localStorage after their completion and implement a garbage collection saga to clean them up after some time
- trigger notifications from the related duck (here, `app.nodes`), while the action gets sent from the `app.salt` duck
- rework the NodeDeployment page slightly (simplification and handling of some edge-cases)

(read commit messages for more info)

**Acceptance criteria**: 

Everything still works, and one can now find info about previously failed jobs even after a refresh. Unit tests were also slightly reworked to make use of features of `jest` to avoid code duplication - make sure you (UI devs) understand how this works and feel comfortable with it.

### DIFF
--- a/ui/src/containers/NodeDeployment.js
+++ b/ui/src/containers/NodeDeployment.js
@@ -5,6 +5,7 @@ import styled from 'styled-components';
 import ReactJson from 'react-json-view';
 import { Button, Loader, Steppers } from '@scality/core-ui';
 import { useRouteMatch, useHistory } from 'react-router';
+import isEmpty from 'lodash.isempty';
 
 import {
   fontWeight,
@@ -19,6 +20,12 @@ const NodeDeploymentContainer = styled.div`
   flex-direction: column;
   padding: ${padding.larger};
   overflow: auto;
+`;
+
+const InfoMessage = styled.div`
+  color: ${props => props.theme.brand.text};
+  font-size: ${fontSize.base};
+  padding: ${padding.base};
 `;
 
 const NodeDeploymentTitle = styled.div`
@@ -122,6 +129,13 @@ const NodeDeployment = ({ intl }) => {
         <NodeDeploymentTitle>
           {intl.messages.node_deployment}
         </NodeDeploymentTitle>
+        {activeJob === undefined ? (
+          <InfoMessage>
+            {intl.translate('no_deployment_found', { name: nodeName })}
+          </InfoMessage>
+        ) : activeJob.completed && isEmpty(activeJob.status) ? (
+          <InfoMessage>{intl.messages.refreshing_job}</InfoMessage>
+        ) : (
           <NodeDeploymentContent>
             <NodeDeploymentStatus>
               <Steppers steps={steps} activeStep={activeStep} />
@@ -134,6 +148,7 @@ const NodeDeployment = ({ intl }) => {
               />
             </NodeDeploymentEvent>
           </NodeDeploymentContent>
+        )}
       </NodeDeploymentWrapper>
     </NodeDeploymentContainer>
   );
--- a/ui/src/containers/NodeDeployment.js
+++ b/ui/src/containers/NodeDeployment.js
@@ -1,13 +1,11 @@
-import React, { useEffect, useState } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import React from 'react';
+import { useSelector } from 'react-redux';
 import { injectIntl } from 'react-intl';
 import styled from 'styled-components';
 import ReactJson from 'react-json-view';
 import { Button, Loader, Steppers } from '@scality/core-ui';
 import { useRouteMatch, useHistory } from 'react-router';
 
-import { subscribeDeployEventsAction } from '../ducks/app/nodes';
-import { getJobStatusFromEventRet } from '../../src/services/salt/utils';
 import {
   fontWeight,
   grayLighter,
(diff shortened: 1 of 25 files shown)


ui/src/ducks/app/salt.js
```diff
@@ -1,31 +1,255 @@
...
+  const savedJobs = listJobsFromLocalStorage();
+  yield all(savedJobs.map(job => put(addJobAction(job))));
+  yield all(savedJobs.map(job => call(refreshJobStatus, job)));
+}
+
+function* manageLocalStorage() {
+  // NOTE: instead of creating a saga per action type, since the operations on
+  // localStorage rely on its previous state, we need to execute them serially.
+  // We don't set up a buffer size as operations should be infrequent and fast.
+  const channel = actionChannel([ADD_JOB, REMOVE_JOB, JOB_COMPLETED]);
```
MonPote: Very cool pattern :)

ui/src/containers/NodeDeployment.js
```diff
@@ -140,14 +129,26 @@ const NodeDeployment = ({ intl }) => {
...
-          <NodeDeploymentStatus>
-            <Steppers steps={steps} activeStep={activeStep} />
-          </NodeDeploymentStatus>
-          <NodeDeploymentEvent>
-            <ReactJson src={events} name={nodeId} collapsed={true} />
-          </NodeDeploymentEvent>
-        </NodeDeploymentContent>
+        {activeJob === undefined ? (
+          <InfoMessage>
+            {`No deployment job found for Node ${nodeName}.`}
```
carlito-scality: translation here

ui/src/containers/NodeDeployment.js
```diff
@@ -140,14 +129,26 @@ const NodeDeployment = ({ intl }) => {
...
+          </InfoMessage>
+        ) : activeJob.completed && isEmpty(activeJob.status) ? (
+          <InfoMessage>{`Refreshing job info...`}</InfoMessage>
```
carlito-scality: translation here

ui/src/ducks/app/nodes.js
```diff
@@ -434,104 +361,60 @@ export function* deployNode({ payload }) {
...
-      jobs.map(jid => {
-        if (data.tag.includes(jid)) {
-          return call(updateDeployEvents, jid, data);
-        }
-        return data;
-      })
+    yield put(
+      addJobAction({
+        name: `deploy-node/${payload.name}`,
+        jid: result.return[0].jid,
```
carlito-scality, line 371: why don't we add `node: payload.name` to avoid `job.name.replace(/^(deploy-node\/)/, '')`?

ui/src/services/salt/utils.js
```diff
@@ -1,7 +1,50 @@
...
+  }
+}
+
+export function markJobCompleteInLocalStorage(jid, completedAt) {
+  const jobs = listJobsFromLocalStorage();
+  const job = jobs.find(item => item.jid === jid);
+  if (job) {
+    job.completedAt = completedAt;
+  }
+  localStorage.setItem(JOBS, JSON.stringify(jobs));
```
carlito-scality: to be replaced by `saveJobsToLocalStorage(jobs)`

ui/src/ducks/app/salt.js
```diff
@@ -1,31 +1,258 @@
...
+        break;
+    }
+  }
+}
+
+export function* refreshJobStatus(job) {
+  const result = yield call(ApiSalt.printJob, job.jid);
+  // TODO: error handling?
+  const status = getJobStatusFromPrintJob(result, job.jid);
+  return yield* updateJobStatus(job, status);
```
carlito-scality: why `yield*` instead id `yield call`?

ui/src/ducks/app/salt.js
```diff
@@ -1,31 +1,258 @@
...
+  }
+}
+// }}}
+
+export function* saltSaga() {
+  yield all([
+    fork(manageLocalStorage),
+    fork(garbageCollectJobs),
+    takeEvery(CONNECT_SALT_API, initialize),
+    takeEvery(CONNECT_SALT_API, watchSaltEvents),
```
carlito-scality: I would merge those two in one saga and call `initialize` before `watchSaltEvents ` so we are sure we do not miss active jobs

ui/src/ducks/app/salt.js
```diff
@@ -1,31 +1,258 @@
...
+  status: {},
+  events: [],
+};
+
 export function addJobAction(job) {
-  return { type: ADD_JOB, payload: job };
+  const payload = { ...defaultJob, ...job };
+  if (payload.completedAt !== null) {
+    payload.completed = true;
+  }
```
carlito-scality, line 102: I wonder if it is necessary here since we will overwrite it in `updateJobStatus`

carlito-scality (changes requested):

ui/src/ducks/app/salt.js
gdemonet: It's for the `refreshNodes` saga actually, so that they don't show the node as deploying while we wait for an update of the status. We know the job is completed, and the consumer uses `completed`, not `completedAt` :/

ui/src/ducks/app/salt.js
gdemonet: :man_shrugging: I read that `yield*` is the idiomatic way of doing this with ES6 generators if you want sequential execution. But true, `call` is useful if we want to test that this saga delegates to another.

ui/src/ducks/app/salt.js
gdemonet: Good point! I'll do that

ui/src/ducks/app/nodes.js
gdemonet: I actually didn't want to add too much in `localStorage`. But since we don't care too much about "dimensions" when it comes to JSON, I think it actually makes a lot of sense to switch to `type` + some arbitrary number of attributes for consumers to read from, in place of some known format to encode these attributes in a single `name` string.

Good point :)

ui/src/ducks/app/salt.js
```diff
@@ -1,31 +1,267 @@
...
+
+export function setJobStatusAction(jid, status) {
+  return { type: SET_JOB_STATUS, payload: { jid, status } };
+}
+
+// Selectors
+export const allJobsSelector = state => state.app.salt.jobs;
+
+// Sagas
+export function* initialize() {
```
carlito-scality: you need to get a `payload` here and pass it to `yield fork(watchSaltEvents, payload);`

carlito-scality (changes requested):

carlito-scality (approved):
LGTM

PR #2071, opened 2019-11-25
set NO_PROXY env var for containerd

### LINKED ISSUE #2052: Proxy configuration for Image Pull only
**Component**:

'containerd'

**Why this is needed**:

In some customer environments, cluster nodes are not connected to the internet. When it comes to download images from the internet, we usually use a proxy, which is external to the cluster. We suggested in our [documentation](https://metal-k8s.readthedocs.io/en/latest/quickstart/setup.html#proxies) to add proxy configuration in `/etc/environment` on all hosts so that 'containerd' pulls image through the proxy. The issue with that approach is that all pods running on the host will also try to access other services through this proxy and it will fail unless you configure every endpoint, which is not what we want.

**What should be done**:

We would like to configure the system so that a proxy can be used, only for pulling images but without affecting the traffic between pods.

**Test plan**:

alexandre-allard:
**Component**: salt

**Context**: Containerd is failing to pull images when a http(s) proxy is set system wide through environment variables in `/etc/environment`

**Summary**: Set NO_PROXY environment variable with control, workload plane and K8s internal
networks in containerd systemd unit file, to avoid using system wide defined HTTP(S)
proxy, if any, when trying to pull resources from metalk8s registry.

**Acceptance criteria**: Be able to install metalk8s with a proxy configured following the doc https://metal-k8s.readthedocs.io/en/latest/quickstart/setup.html#proxies

---

Closes: #2052

### DIFF
--- a/salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2
+++ b/salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2
@@ -2,3 +2,4 @@
 # See https://github.com/containerd/containerd/issues/3201
 # See https://github.com/containerd/containerd/pull/3202
 LimitNOFILE=1048576
+Environment="{{ environment }}"
--- a/salt/metalk8s/container-engine/containerd/installed.sls
+++ b/salt/metalk8s/container-engine/containerd/installed.sls
@@ -2,6 +2,8 @@
 {%- from "metalk8s/map.jinja" import metalk8s with context %}
 {%- from "metalk8s/map.jinja" import kubelet with context %}
 {%- from "metalk8s/map.jinja" import repo with context %}
+{%- from "metalk8s/map.jinja" import networks with context %}
+{%- from "metalk8s/map.jinja" import proxies with context %}
 
 {%- set registry_ip = metalk8s.endpoints['repositories'].ip %}
 {%- set registry_port = metalk8s.endpoints['repositories'].ports.http %}
@@ -36,12 +38,28 @@ Install containerd:
 Create containerd service drop-in:
   file.managed:
     - name: /etc/systemd/system/containerd.service.d/50-metalk8s.conf
-    - source: salt://{{ slspath }}/files/50-metalk8s.conf
+    - source: salt://{{ slspath }}/files/50-metalk8s.conf.j2
+    - template: jinja
     - user: root
     - group: root
     - mode: 0644
     - makedirs: true
     - dir_mode: 0755
+    - context:
+      environment: >-
+      {%- if proxies %}
+        {%- set no_proxy = ["localhost", "127.0.0.1"] + networks.values %}
+        {%- if proxies.no_proxy | default %}
+          {%- do no_proxy.extend(proxies.no_proxy) %}
+        {%- endif %}
+        NO_PROXY={{ no_proxy | unique | join(",") }}
+        {%- if proxies.http | default %}
+        HTTP_PROXY={{ proxies.http }}
+        {%- endif %}
+        {%- if proxies.https | default %}
+        HTTPS_PROXY={{ proxies.https }}
+        {%- endif %}
+      {%- endif %}
     - require:
       - metalk8s_package_manager: Install containerd
 
--- a/salt/metalk8s/container-engine/containerd/installed.sls
+++ b/salt/metalk8s/container-engine/containerd/installed.sls
@@ -31,6 +31,8 @@ Install containerd:
     - require:
       - test: Repositories configured
       - metalk8s_package_manager: Install runc
+      - file: Create containerd service drop-in
+      - file: Configure registry IP in containerd conf
       {%- if grains['os_family'].lower() == 'redhat' %}
       - metalk8s_package_manager: Install container-selinux
       {%- endif %}
@@ -60,8 +62,6 @@ Create containerd service drop-in:
         HTTPS_PROXY={{ proxies.https }}
         {%- endif %}
       {%- endif %}
-    - require:
-      - metalk8s_package_manager: Install containerd
 
 Install and configure cri-tools:
   {{ pkg_installed('cri-tools') }}
@@ -85,5 +85,3 @@ Configure registry IP in containerd conf:
     - contents: |
         [plugins.cri.registry.mirrors."{{ repo.registry_endpoint }}"]
           endpoint = ["http://{{ registry_ip }}:{{ registry_port }}"]
-    - require:
-      - metalk8s_package_manager: Install containerd
--- a/salt/metalk8s/repo/redhat.sls
+++ b/salt/metalk8s/repo/redhat.sls
@@ -61,6 +61,10 @@ Configure {{ repo_name }} repository:
   {%- endif %}
     - repo_gpg_check: {{ repo_config.repo_gpg_check }}
     - enabled: {{ repo_config.enabled }}
+    # URL to the proxy server for this repository.
+    # Set to '_none_' to disable the global proxy setting
+    # for this repository.
+    - proxy: _none_
     - refresh: false
     - onchanges_in:
       - cmd: Refresh yum cache
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -290,7 +290,7 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/nginx-ingress-control-plane/pre-upgrade.sls'),
 
     Path('salt/metalk8s/container-engine/containerd/configured.sls'),
-    Path('salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf'),
+    Path('salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2'),
     Path('salt/metalk8s/container-engine/containerd/init.sls'),
     Path('salt/metalk8s/container-engine/containerd/installed.sls'),
     Path('salt/metalk8s/container-engine/init.sls'),
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -36,6 +36,12 @@ Configuration
       networks:
         controlPlane: <CIDR-notation>
         workloadPlane: <CIDR-notation>
+      proxies:
+        http: <http://proxy-ip:proxy-port>
+        https: <https://proxy-ip:proxy-port>
+        no_proxy:
+          - <host>
+          - <ip/cidr>
       ca:
         minion: <hostname-of-the-bootstrap-node>
       archives:
@@ -44,6 +50,13 @@ Configuration
 The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
 the bootstrap script is executed, those ISOs are automatically mounted and the
 system is configured to re-mount them automatically after a reboot.
+
+The ``proxies`` field can be omitted if there is no proxy to configure.
+The 2 entries ``http`` and ``https`` are used to configure the containerd
+daemon proxy to fetch extra container images from outstide the MetalK8s
+cluster.
+The ``no_proxy`` entry specifies IPs that should be excluded from proxying,
+it must a list of hosts, IP addresses or IP ranges in CIDR format.
 
 .. todo::
 
--- a/docs/installation/setup.rst
+++ b/docs/installation/setup.rst
@@ -23,14 +23,7 @@ partition larger than 40 GB.
 
 Proxies
 -------
-For nodes operating behind a proxy, add the following lines to each cluster
-member's :file:`/etc/environment` file:
-
-.. code-block:: shell
-
-   http_proxy=http://user;pass@<HTTP proxy IP address>:<port>
-   https_proxy=http://user;pass@<HTTPS proxy IP address>:<port>
-   no_proxy=localhost,127.0.0.1,<local IP of each node>
+For nodes operating behind a proxy, see :ref:`Bootstrap Configuration`
 
 Provisioning
 ------------
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -104,7 +104,8 @@ def ext_pillar(minion_id, pillar, bootstrap_config):
 
     result = {
         'networks': _load_networks(config),
-        'metalk8s': metal_data
+        'metalk8s': metal_data,
+        'proxies': config.get('proxies', {})
     }
 
     if not isinstance(metal_data['archives'], list):
--- a/salt/metalk8s/defaults.yaml
+++ b/salt/metalk8s/defaults.yaml
@@ -94,3 +94,5 @@ coredns:
   reverse_cidrs: in-addr.arpa ip6.arpa
 
 upgrade: false        # define if we're on an upgrade case
+
+proxies: {}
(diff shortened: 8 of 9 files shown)


salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -36,12 +37,16 @@ Install containerd:
...
+    - source: salt://{{ slspath }}/files/50-metalk8s.conf.j2
+    - template: jinja
     - user: root
     - group: root
     - mode: 0644
     - makedirs: true
     - dir_mode: 0755
+    - context:
+      environment: >-
+        NO_PROXY=127.0.0.1,localhost,{{ networks.values() | join(",") }}
```
sayf-eddine-scality: shouldnt we set `.svc,.default,.local,.cluster.local` too ?

alexandre-allard: Sure
`.local` & `.cluster.local` ? Isn't `.local` sufficient ?

sayf-eddine-scality: hmm yes i guess it is

salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2
```diff
@@ -2,3 +2,4 @@
 # See https://github.com/containerd/containerd/issues/3201
 # See https://github.com/containerd/containerd/pull/3202
 LimitNOFILE=1048576
+Environment="{{ environment }}"
```
gdemonet, line 5: Is it picking up whatever is in `/etc/environment` on its own? Or should we manually copy its contents here?

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -36,12 +37,17 @@ Install containerd:
...
+    - source: salt://{{ slspath }}/files/50-metalk8s.conf.j2
+    - template: jinja
     - user: root
     - group: root
     - mode: 0644
     - makedirs: true
     - dir_mode: 0755
+    - context:
+      environment: >-
+        NO_PROXY=127.0.0.1,localhost,.svc,.default,.local,{{
```
gdemonet: I don't know the allowed syntax of `NO_PROXY`, but this will probably won't be enough.

First question: what's the exact symptom? When we have a `HTTP_PROXY` set up, is `containerd` failing on retrieval of new container images? Are containers also failing to reach other services/cluster IPs?

Second question: if `containerd` uses the proxy for container networking as well, then we'll need a lot more than this (I don't even think we could be exhaustive) - all Service DNS names can use `.svc.<namespace>`, or `.svc.<namespace>.cluster`, or `.svc.<namespace>.cluster.local`, or even remove this whole suffix and only use the Service actual name. We naturally cannot specify all of them here, so maybe we could instead try to resolve names with CoreDNS before trying to go through the proxy?

gdemonet (commented):
Some questions...

salt/metalk8s/container-engine/containerd/installed.sls
alexandre-allard: During my tests, only containerd was failing to retrieve container images.
I don't think that these environment variables are used inside running containers.

salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2
alexandre-allard: In my understanding, it's a merge between what's already defined in environment and what we define here, that's why I only defined NO_PROXY, so that we use HTTP(S)_PROXY defined system wide.

salt/metalk8s/container-engine/containerd/installed.sls
gdemonet: OK, then I guess we don't need all those rules, just the registry address, no? Something like `metalk8s-registry-from-config.invalid` (or its actual IP:port, not sure)?

salt/metalk8s/container-engine/containerd/files/50-metalk8s.conf.j2
gdemonet: OK perfect.

salt/metalk8s/container-engine/containerd/installed.sls
NicolasT: I seem to remember from the Slack chats with TS, the processes *inside* containers were also using the proxy variables, then failing to connect to e.g. `10.96.0.1` or whatnot. Please validate this more in-depth...

In any case:
- Where does `.default` come from?
- Where does `.local` come from?
- Indeed, what about the registry 'DNS name'?

alexandre-allard-scality: I think that only the registry address could be sufficient, when I try to deploy setting only this address in the no_proxy variable, it seems to work fine.
Maybe adding the whole network would be a better idea, just in case we want to handle multiple replicas for registry.
As I said above, processes inside containers don't seem to be affected by these environment variables.

docs/quickstart/bootstrap.rst
```diff
@@ -36,6 +36,9 @@ Configuration
       networks:
         controlPlane: <CIDR-notation>
         workloadPlane: <CIDR-notation>
+      proxies:
+        http: <http://ip-of-the-proxy>
+        https: <https://ip-of-the-proxy>
```
Ebaneck: Should be <http://ip-of-the-proxy:proxy-port>  ??

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -36,12 +38,24 @@ Install containerd:
...
+    - template: jinja
     - user: root
     - group: root
     - mode: 0644
     - makedirs: true
     - dir_mode: 0755
+    - context:
+      environment: >-
+      {%- if proxies %}
+        NO_PROXY={{ networks.values() | join(",") }}
```
Ebaneck: Don't we need loopback addresses here?? `127.0.0.1` and `localhost` ??

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -36,12 +38,24 @@ Install containerd:
...
+        {%- if 'http' in proxies %}
+        HTTP_PROXY={{ proxies.http }}
+        {%- endif %}
+        {%- if 'https' in proxies %}
+        HTTPS_PROXY={{ proxies.https }}
+        {%- endif %}
+      {%- endif %}
     - require:
       - metalk8s_package_manager: Install containerd
```
Ebaneck: If we amended the contanerd service configs then we need a restart for our changes to take effect. no??
The way i see the sequence
- Install containerd
- patch the configuration file
Could be inverted ??

Ebaneck (commented):
LGTM, just a few questions that need clarifications

docs/quickstart/bootstrap.rst
alexandre-allard: Yes, could be, will change it.

salt/metalk8s/container-engine/containerd/installed.sls
TeddyAndrieux: It's not needed since in `configured` sls we have a `service.running` that watch this state
BUT
for me this service configuration should be in `configured` sls and it's not really part of this PR (but feel free to do it :) )

salt/metalk8s/repo/redhat.sls
```diff
@@ -61,6 +61,7 @@ Configure {{ repo_name }} repository:
   {%- endif %}
     - repo_gpg_check: {{ repo_config.repo_gpg_check }}
     - enabled: {{ repo_config.enabled }}
+    - proxy: _none_
```
gdemonet, line 67: What does this do? I can't find it in the [docs](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html) nor the [source code](https://github.com/saltstack/salt/blob/develop/salt/states/pkgrepo.py) for `pkgrepo` :/

salt/metalk8s/container-engine/containerd/installed.sls
alexandre-allard: Yep, not really the scope of the changes, but yeah, could be improved.

salt/metalk8s/container-engine/containerd/installed.sls
alexandre-allard: If I refer to my tests, it doesn't seem to be mandatory, I can still add them, just in case.

gdemonet (commented):
Small question, looks good to me otherwise, can we open a debt ticket for adding some test behind a proxy?

salt/metalk8s/repo/redhat.sls
Ebaneck: Me too, i couldn't figure it upon initial review then I stumbled on this  https://github.com/saltstack/salt/issues/35760

alexandre-allard: It sets the entry `proxy=_none_` in each repository configuration file as it is explained in `man yumconf`:

**proxy** URL to the proxy server for this repository. Set to '\_none\_' to disable the global proxy setting
for this repository. If this is unset it inherits it from the global setting

The goal is to handle the case when the user has a yum or system-wide defined proxy.

It's not in the doc, but you can see in the code (if you go a little deeper in [salt/modules/yumpkg.py](https://github.com/saltstack/salt/blob/develop/salt/modules/yumpkg.py)) that almost any kwargs is a valid parameter (except salt vars & those starting with `__`).

slaperche-scality: If two people (and even one person in fact) didn't understand the meaning of this, that's a clear indication that this deserves a comment :wink: 
Especially since it's not documented.

alexandre-allard:
> Small question, looks good to me otherwise, can we open a debt ticket for adding some test behind a proxy?

@gdemonet Yup, easily doable.
Since we don't need any image from the internet, we can set a fake proxy pointing to nothing (e.g. http://10.10.10.10:1234).
I'll open an issue to track this.

salt/metalk8s/repo/redhat.sls
alexandre-allard: Thought it was pretty clear, but I'm used to yum syntax, maybe it's not.
I've added a comment.

Ebaneck (dismissed):

gdemonet (dismissed):

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -29,21 +31,33 @@ Install containerd:
     - require:
       - test: Repositories configured
       - metalk8s_package_manager: Install runc
+      - file: Create containerd service drop-in
+      - file: Configure registry IP in containerd conf
```
TeddyAndrieux, line 35: For me it's not intuitive to do like that because that mean on the first install you will do in this order
```
- Configure repo
- Install runc
- Create containerd service file
- Create containerd config file toml
- Install containerd
[...]
- Restart containerd <--- Because the `service.running` in "configured.sls" watch containerd service file and containerd config toml
```
And btw you really want to have this 2 "watch" because if containerd already installed and you change something in one of this two file you want the state to restart containerd
 
For me this two state should be in "configured.sls"

alexandre-allard: IIRC, containerd is not automatically started after its installation, that's the goal of this.
You ensure that you only start it when everything is ready.

docs/quickstart/bootstrap.rst
```diff
@@ -47,6 +50,11 @@ The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
 the bootstrap script is executed, those ISOs are automatically mounted and the
 system is configured to re-mount them automatically after a reboot.
 
+The ``proxies`` field can be omitted if there is no proxy to configure.
+The 2 entries ``http`` and ``https`` are used to configure containerd
```
NicolasT:
```suggestion
The 2 entries ``http`` and ``https`` are used to configure the containerd
```

docs/quickstart/bootstrap.rst
```diff
@@ -47,6 +50,11 @@ The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
...
+daemon proxy to fetch extra container images from outstide the metalk8s
```
NicolasT:
```suggestion
daemon proxy to fetch extra container images from outstide the MetalK8s
```

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -29,21 +31,33 @@ Install containerd:
...
     - group: root
     - mode: 0644
     - makedirs: true
     - dir_mode: 0755
-    - require:
-      - metalk8s_package_manager: Install containerd
+    - context:
+      environment: >-
+      {%- if proxies %}
+        NO_PROXY=localhost,127.0.0.1,{{ networks.values() | join(",") }}
```
NicolasT: Shouldn't we allow to set *other* `NO_PROXY` values in configuration, similar to how HTTP/HTTPS proxies can be defined?

docs/quickstart/bootstrap.rst
```diff
@@ -36,6 +36,9 @@ Configuration
       networks:
         controlPlane: <CIDR-notation>
         workloadPlane: <CIDR-notation>
+      proxies:
```
NicolasT: Since we only use these values to *specifically* configure `containerd`, not any other services that may be running, or containers, or whatnot, maybe using a more specific name would be wise? WDYT @scality/metalk8s

TeddyAndrieux: Yes maybe make sense to know just by reading the file that this proxy will only be used by the container engine to fetch images. 
Maybe something like
- `registry_proxies`
- `extra_container_images_proxies`
- `images_proxies`

(I'm not sure we want to talk about containerd here)

NicolasT: Well, no, I don't think we need to do *multiple* proxy declarations (if there's some proxy in place, there is one, used for all traffic).
However, as-is, it may not be clear we *only* use this configuration *for one very specific service*.

NicolasT (dismissed):
See comments.

docs/quickstart/bootstrap.rst
TeddyAndrieux: Oh no not saying we need multiple just suggest few key name

alexandre-allard: I've put `proxies` to be more general, because I don't know if we'll use this entry to set the proxy on other stuff in the future.
But ok, I will change it.

NicolasT: That makes sense, and maybe `proxies` *is* the right key. But in that case, it should be explicitly documented where exactly this configuration is used/applied, such that there can be no false assumptions about other subsystems being configured using it.

TeddyAndrieux: It's already the case just bellow in the doc no ?

salt/metalk8s/container-engine/containerd/installed.sls
alexandre-allard: Make sense, will do.

docs/quickstart/bootstrap.rst
alexandre-allard: done

docs/quickstart/bootstrap.rst
alexandre-allard: done

salt/metalk8s/container-engine/containerd/installed.sls
alexandre-allard: done

Ebaneck (approved):
Looks good to go!

Issue #2077, opened 2019-11-26
Installer assumes versionlock is installed


bkettler:
When performing the bootstrap step if the yum-versionlock plugin is not installed and/or disabled the installer fails.

**Component**:
installer

**What happened**:
The installer fails if the yum versionlock plugin is not installed and enabled.

**What was expected**:
The installer should check for dependencies and install them if missing or instruct the user how to rectify the issue.

**Steps to reproduce**
Uninstall the yum-versionlock plugin and/or disable plugins in /etc/yum.conf then run the bootstrap.sh script.

**Resolution proposal** (optional):
The bootstrap script should either install and configure the yum-versionlock plugin or fail immediately and instruct the user to do so.

PR #2082, opened 2019-11-27
salt: ensure yum versionlock plugin is installed & enabled

alexandre-allard:
**Component**: salt

**Context**:
bootstrap installation fails if yum versionlock plugin is disabled

**Summary**:
If the plugin is disabled, we can't install packages through our salt states because we lock each and every packages we install.
Plugin is enabled by default when installed, but it could have been disabled for some reasons, so we
need to ensure it's enabled before trying to install any package.

**Acceptance criteria**: bootstrap installation passes even if yum versionlock plugin is disabled prior to installation

---

Closes: #2077

### DIFF
--- a/salt/metalk8s/repo/offline.sls
+++ b/salt/metalk8s/repo/offline.sls
@@ -9,6 +9,15 @@ Install yum-plugin-versionlock:
     - name: yum-plugin-versionlock
     - fromrepo: {{ repo.repositories.keys() | join(',') }}
 
+Ensure yum plugins are enabled:
+  ini.options_present:
+    - name: /etc/yum.conf
+    - sections:
+        main:
+          plugins: 1
+    - require_in:
+      - test: Repositories configured
+
 Ensure yum versionlock plugin is enabled:
   ini.options_present:
     - name: /etc/yum/pluginconf.d/versionlock.conf
--- a/salt/metalk8s/repo/offline.sls
+++ b/salt/metalk8s/repo/offline.sls
@@ -8,8 +8,17 @@ Install yum-plugin-versionlock:
   pkg.installed:
     - name: yum-plugin-versionlock
     - fromrepo: {{ repo.repositories.keys() | join(',') }}
+
+Ensure yum versionlock plugin is enabled:
+  ini.options_present:
+    - name: /etc/yum/pluginconf.d/versionlock.conf
+    - sections:
+        main:
+          enabled: 1
     - require_in:
       - test: Repositories configured
+    - require:
+      - pkg: Install yum-plugin-versionlock
 
 {%- for repo_name, repo_config in repo.repositories.items() %}
   {%- if repo.local_mode %}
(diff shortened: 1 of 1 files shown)


slaperche-scality (commented):
Good job on the commit message :+1:

salt/metalk8s/repo/offline.sls
```diff
@@ -41,6 +41,8 @@ Configure {{ repo_name }} repository:
     - refresh: false
     - onchanges_in:
       - cmd: Refresh yum cache
+    - require_in:
+      - pkg: Install yum-plugin-versionlock
```
gdemonet, line 63: Not sure I understand why there is this requirement. From your commit message, you said we need `yum-plugin-versionlock` to be installed before other states depending on `Repositories configured`. That makes sense. But why would `yum-plugin-versionlock` installation require all **custom** repos to be configured? It actually lives in `base`, which we don't need to configure.

To me, it would make more sense to have here something like:

```
- require_in:
```

gdemonet (dismissed):
Minor change request for some requisites. Otherwise, LGTM :)

salt/metalk8s/repo/offline.sls
TeddyAndrieux: > But why would `yum-plugin-versionlock` installation require all **custom** repos to be configured? It actually lives in `base`, which we don't need to configure.

False, in metalk8s < 2.4.

LGTM to keep it like this for the moment and when we will backport the fact that we not embed `base` repos in our ISO in metalk8s-2.0 we just need to remove this `require_in`

salt/metalk8s/repo/offline.sls
```diff
@@ -57,8 +77,4 @@ Refresh yum cache:
       - cmd: Refresh yum cache
 
 Repositories configured:
-  test.succeed_without_changes:
-    - require:
-{%- for repository_name in repo.repositories.keys() %}
-      - pkgrepo: Configure {{ repository_name }} repository
-{%- endfor %}
```
TeddyAndrieux, line 64: If you remove this you need to add a `require_in` in `Configure {{ repostiroy_name }} repository` state

alexandre-allard: It's actually not really needed since there is a require_in on it in the yum plugin versionlock package configuration, which requires yum plugin versionlock installation, which requires repositories to be setup... so
repo setup > yum install yum-plugin-version > yum versionlock plugin configuration > repositories configured

But I could add it to be more explicit.

TeddyAndrieux: Even if it's implicit for the moment because of the `require_in` on yum-versionlock like I said in my comment just above (https://github.com/scality/metalk8s/pull/2082/files#r351709055) if we remove the `require_in` on yum-versionlock we want this `Repositories configured` state to require all `Configure ... ` states

alexandre-allard: Yeah, sure, will do it.

TeddyAndrieux (approved):
LGTM, enjoy flaky rebuilds :smiley:

Issue #2085, opened 2019-11-28
UI: update core-ui


carlito-scality:
**Component**:UI

**Why this is needed**:
- As we added new features above all the improvements of navbar to be aligned with Use Design design and they introduced breaking changes, we need to address them as soon as possible.

**What should be done**:
`ToggleVisible, languages, applications, help, user` should be removed

**Implementation proposal** (strongly recommended):

**Test plan**:
- No regressions

ChengYanJin:
already merged by PR #2086

PR #2086, opened 2019-11-28
UI: Refactor navbar to align with the update of core ui

ChengYanJin:
**Component**: ui

**Context**: 
As there are breaking changes in core-ui(release version 0.3.0), we will need to update the version of core-ui and refactor the Navbar at the same time. (covered metalk8s-UI and example-solution UI)

**Summary**:

- Removed toggleVisible, applications, help, user and languages from Navbar.
- Add `rightActions` with the list of actions in Navbar.
- Implement the language the logic.

**Acceptance criteria**: 
We won't see the dropdown in Navbar if there is no item. 

---

Closes: #2085

### DIFF
--- a/examples/metalk8s-solution-example/ui/src/containers/Layout.js
+++ b/examples/metalk8s-solution-example/ui/src/containers/Layout.js
@@ -47,22 +47,6 @@ const Layout = props => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const applications = [];
-
-  const help = [
-    {
-      label: intl.messages.about,
-      onClick: () => {
-        history.push('/about');
-      }
-    }
-  ];
-
-  const userConfig = {
-    name: user && user.username,
-    actions: [{ label: intl.messages.log_out, onClick: logout }]
-  };
-
   const sidebarConfig = {
     expanded: sidebar.expanded,
     actions: [
@@ -87,13 +71,31 @@ const Layout = props => {
     ]
   };
 
+  const rightActions = [
+    {
+      type: 'dropdown',
+      icon: <i className="fas fa-question-circle" />,
+      items: [
+        {
+          label: intl.messages.about,
+          onClick: () => {
+            history.push('/about');
+          }
+        }
+      ]
+    },
+    {
+      type: 'dropdown',
+      text: user && user.username,
+      icon: <i className="fas fa-user" />,
+      items: [{ label: intl.messages.log_out, onClick: logout }]
+    }
+  ];
+
   const navbar = {
     onToggleClick: toggleSidebar,
-    toggleVisible: true,
     productName: intl.messages.product_name,
-    applications,
-    help,
-    user: user && userConfig,
+    rightActions,
     logo: (
       <img
         alt="logo"
--- a/ui/src/containers/Layout.js
+++ b/ui/src/containers/Layout.js
@@ -44,26 +44,10 @@ const Layout = props => {
   const toggleSidebar = () => dispatch(toggleSideBarAction());
   const { intl } = props;
   const history = useHistory();
-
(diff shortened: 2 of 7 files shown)


ui/src/containers/Layout.js
```diff
@@ -143,14 +128,44 @@ const Layout = props => {
     },
   ];
 
+  const filterLanguage = languages.find(lang => lang.name !== language);
```
carlito-scality: you should use `filter` instead of `find` here. In our case, we have 2 languages, that's why it works.

ui/src/containers/Layout.js
```diff
@@ -143,14 +128,44 @@ const Layout = props => {
...
+
+  const rightActions = [
+    {
+      type: 'dropdown',
+      text: language,
+      icon: <i className="fas fa-globe" />,
+      items: filterLanguage === null ? null : [filterLanguage],
```
carlito-scality: `items: filterLanguage` is enough if you use `filter` above.

ui/src/containers/Layout.js
```diff
@@ -143,14 +128,44 @@ const Layout = props => {
...
+    },
+    {
+      type: 'dropdown',
+      icon: <i className="fas fa-th" />,
+      items: applications === null ? null : [applications],
```
carlito-scality: `items: applications` is enough because applications is an array. if `applications` is null, you should not add it to `rightActions`

carlito-scality (changes requested):

ui/src/containers/Layout.js
```diff
@@ -143,14 +128,44 @@ const Layout = props => {
...
+    {
+      type: 'dropdown',
+      text: language,
+      icon: <i className="fas fa-globe" />,
+      items: filterLanguage,
+    },
+    {
+      type: 'dropdown',
+      icon: <i className="fas fa-th" />,
+      items: applications,
```
carlito-scality: you should check if there is applications not before adding this `rightActions`

examples/metalk8s-solution-example/ui/src/containers/Layout.js
```diff
@@ -87,13 +73,36 @@ const Layout = props => {
     ]
   };
 
+  const rightActions = [
+    {
+      type: 'dropdown',
+      icon: <i className="fas fa-th" />,
+      items: applications
```
carlito-scality: same here, you should check if there is applications not before adding this rightActions

examples/metalk8s-solution-example/ui/src/containers/Layout.js
```diff
@@ -49,20 +49,6 @@ const Layout = props => {
 
   const applications = [];
```
carlito-scality: you should update the core-ui in package.json of metalk8s-solution-example as well

carlito-scality (changes requested):

examples/metalk8s-solution-example/ui/src/containers/Layout.js
ChengYanJin: I am wondering in solution-example UI do we need applications item in `Navbar`?

ui/src/containers/Layout.js
```diff
@@ -143,14 +127,46 @@ const Layout = props => {
...
+      icon: <i className="fas fa-globe" />,
+      items: filterLanguage,
+    },
+    applications == null
+      ? {}
+      : {
+          type: 'dropdown',
+          icon: <i className="fas fa-th" />,
+          items: applications,
+        },
```
carlito-scality: please move this check outside this array and insert 
```
{
  type: 'dropdown',
  icon: <i className="fas fa-th" />,
  items: applications,
 }
```
only if `applications && applications.length`

carlito-scality (changes requested):

carlito-scality (approved):

PR #2087, opened 2019-11-29
Improvement/no longer use hostname in ci test

TeddyAndrieux:
**Component**:

'tests', 'build'

**Context**: 

We no longer need to use hostname to deploy and extend metalk8s cluster.

**Summary**:

Do not use hostname to deploy, extend, test in CI tests

**Acceptance criteria**: 

Green build

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -211,7 +211,7 @@ models:
           controlPlane: 10.100.0.0/16
           workloadPlane: 10.100.0.0/16
         ca:
-          minion: $(hostname)
+          minion: "bootstrap"
         apiServer:
           host: $(ip route get 10.100.0.0 | awk '/10.100.0.0/{ print $6 }')
         archives:
@@ -531,8 +531,7 @@ stages:
             OS_USERNAME: "%(secret:scality_cloud_username)s"
             OS_PASSWORD: "%(secret:scality_cloud_password)s"
             OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
-            # FIXME: this makes hostnames too long
-            # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+            TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
             TF_VAR_nodes_count: "2"
           haltOnFailure: true
       - ShellCommand:
--- a/eve/workers/openstack-multiple-nodes/terraform/scripts/bootstrap-config.sh
+++ b/eve/workers/openstack-multiple-nodes/terraform/scripts/bootstrap-config.sh
@@ -7,7 +7,7 @@ OUTPUT_FILE="/etc/metalk8s/bootstrap.yaml"
 mkdir -p "$(dirname $OUTPUT_FILE)"
 
 mkdir -p /etc/salt
-hostname > /etc/salt/minion_id
+echo "bootstrap-node" > /etc/salt/minion_id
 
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -42,6 +42,19 @@ def hostname(host):
...
+@pytest.fixture(scope="module")
+def nodename(host):
+    """Return the kubernetes node name of the `host`
+
+    Node name need to be equal to the salt minion id so just retrieve the
+    salt minion id
+    """
+    with host.sudo():
+        return host.check_output(
+            'salt-call --out txt grains.get id | cut -c 8-'
```
NicolasT: Could this use `--local`?

tests/utils.py
```diff
@@ -52,11 +52,12 @@ def get_ip_from_cidr(host, cidr):
...
+def get_node_name(nodename, ssh_config=None):
+    """Get a node name (from SSH config)."""
     if ssh_config is not None:
         node = testinfra.get_host(nodename, ssh_config=ssh_config)
-        nodename = node.check_output('hostname')
-    else:
-        assert nodename == 'bootstrap'
+        with node.sudo():
+            return node.check_output(
+                'salt-call grains.get --out txt id | cut -c 8-'
```
NicolasT: `--local`?

tests/conftest.py
TeddyAndrieux: Yes agree will be better

eve/main.yml
```diff
@@ -531,8 +531,7 @@ stages:
             OS_USERNAME: "%(secret:scality_cloud_username)s"
             OS_PASSWORD: "%(secret:scality_cloud_password)s"
             OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
-            # FIXME: this makes hostnames too long
-            # TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
+            TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
```
gdemonet, line 534: Cool!

eve/main.yml
```diff
@@ -211,7 +211,7 @@ models:
           controlPlane: 10.100.0.0/16
           workloadPlane: 10.100.0.0/16
         ca:
-          minion: $(hostname)
+          minion: bootstrap
```
gdemonet: I think it could make sense to leave one of the tested scenarios to the default, which is hostname for Salt. That's what some, if not most users, will do, probably makes sense to test it regularly.

gdemonet (commented):
Changes overall look good to me, just one remark on CI coverage - what's your opinion?

eve/main.yml
TeddyAndrieux: Yes make sense

gdemonet (approved):
LGTM.

I'm not entirely sure, which CI stage is still using hostname? Or are they all using `bootstrap` now instead?

TeddyAndrieux:
> I'm not entirely sure, which CI stage is still using hostname? Or are they all using `bootstrap` now instead?

I tried to use hostname for one but was not able to have something working well without setting minion_id to hostname before launching bootstrap .... but it's almost the same as setting minion_id directly need to dig into salt code 

Look like salt first try to get fqdn using hostname if empty he will use hostname directly ...
```
    hosts = DistinctList().append(
        salt.utils.stringutils.to_unicode(socket.getfqdn(salt.utils.stringutils.to_bytes(hostname)))
    ).append(platform.node()).append(hostname)
```

Issue #2078, opened 2019-11-26
Store the access_token in the localStorage


ChengYanJin:
**Component**:  Dex, ui

**Why this is needed**:
After login, if we open another tab in the same browser, we are redirected to the login page again.

**What should be done**:

The hint is to change the user data into localStorage instead of sessionStorage, which is the current case for redux-oidc library. (But we still need to investigate if this is the real issue)
```
const userManagerSettings = {
...,
userStore: new WebStorageStateStore({ store: localStorage})
}
```
According to the suggestion of the redux-oidc git, we will need to look at the `oidc-client documentation` when configuring the userManager.

**oidc-client documentation**
https://github.com/IdentityModel/oidc-client-js/wiki#configuration

> userStore: (default: session storage): Storage object used to persist User for currently authenticated user. E.g. userStore: new WebStorageStateStore({ store: window.localStorage })

**Implementation proposal** (strongly recommended):
TBD

**Test plan**:
TBD

PR #2088, opened 2019-11-29
Improvement/2078 store user credentials in localstorage

ChengYanJin:
**Component**: UI, Dex

**Context**: 
We will want to share the user credentials across different tabs so that users won't need to be authenticated again.

**Summary**:
We set `userStore: new WebStorageStateStore({ store: localStorage }),` by default in userManagerConfig.

**Acceptance criteria**: 
Once login, we should find the `oidc.user:https://172.21.254.12:8443/oidc:metalk8s-ui` oidc credential stored in localStorage.

---

Closes: #2078

### DIFF
--- a/ui/config-overrides.js
+++ b/ui/config-overrides.js
@@ -15,7 +15,7 @@ const CompressionPlugin = require('compression-webpack-plugin');
 const setWebpackPerformance = () => config => {
   config.performance = {
     hints: 'error',
-    // ~586 KiB for production
+    // ~684 KiB for production
     // ~1953 KiB for development because flow increase the size of assets.
     maxAssetSize: process.env.NODE_ENV === 'production' ? 700000 : 2000000,
     assetFilter: assetFilename => {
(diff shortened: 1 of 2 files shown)


ui/package.json
```diff
@@ -10,7 +10,7 @@
     "formik": "1.5.1",
     "lodash.isempty": "^4.4.0",
     "lodash.sortby": "^4.7.0",
-    "oidc-client": "^1.8.0",
+    "oidc-client": "^1.9.1",
```
carlito-scality: why do you need to update?

carlito-scality (changes requested):
Works great but the logout button does not work anymore.

ui/package.json
ChengYanJin: oh. ok. You used v 1.8.0. I will keep the same as well.

carlito-scality (approved):
I rebuilt and tested again. Everything is ok. I cannot reproduce the logout button problem.

PR #2091, opened 2019-12-02
CI: Add upgrade test for PATCH VERSIONS

### LINKED ISSUE #2009: Add steps to upgrade between patch versions
**Component**: upgrade, downgrade, tests, ci

**Why this is needed**:
Test that upgrade/downgrade between patch versions is working fine

**What should be done**:
We should do what's already done for minor version upgrade/downgrade.
Add two steps in the CI (1 for upgrade and 1 for downgrade)

Upgrade:
- Get version X.Y.(Z - 1) ISO
- Install version X.Y.(Z - 1)
- Launch tests
- Get version X.Y.Z ISO
- Upgrade to version X.Y.Z
- Launch tests

Downgrade:
- Get version X.Y.Z ISO
- Install version X.Y.Z
- Launch tests
- Get version X.Y.(Z - 1) ISO
- Downgrade to version X.Y.(Z - 1)
- Launch tests

**Test plan**: N/A

Ebaneck:
**Component**:

'ci', 'tests'

**Context**: 

See #2009 

**Summary**:

We should be able to automatically test patch versions using existing promoted artefacts.
Targeting 2.4 branches only because patch version inclusion was first introduced here.

**Acceptance criteria**: 

Please don't come with any more flakies

---

Closes: #2009

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -428,6 +428,7 @@ stages:
             - single-node-downgrade-centos
             - single-node-install-rhel
             - multiple-nodes-centos
+            - single-node-patch-version
       - ShellCommand: *add_final_status_artifact_success
       - Upload: *upload_final_status_artifact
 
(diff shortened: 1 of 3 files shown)


eve/main.yml
```diff
@@ -405,6 +405,14 @@ stages:
...
               minor=$(echo "%(prop:product_version)s" | cut -d'.' -f2) &&
               echo "$major.$(( $minor-1 ))"
+      - SetPropertyFromCommand:
+          name: Set previous patch version to upgrade from and downgrade to
+          property: product_patch_version_prev
+          command: >
+              major=$(echo "%(prop:product_version)s" | cut -d'.' -f1) &&
+              minor=$(echo "%(prop:product_version)s" | cut -d'.' -f2) &&
+              patch=$(echo "%(prop:product_version)s" | cut -d'.' -f3) &&
+              echo "$major.$minor.$(( $patch-1 ))"
```
alexandre-allard: You must do some checks before doing that, otherwise you could end up with a version `2.5.-1`, not sure that what we want. :)
I would do something like:
```
              major=$(echo "%(prop:product_version)s" | cut -d'.' -f1) &&
              minor=$(echo "%(prop:product_version)s" | cut -d'.' -f2) &&
              patch=$(echo "%(prop:product_version)s" | cut -d'.' -f3) &&
              echo "$major.$minor.$(( patch > 0 ? patch - 1 : 0 ))"
```
And... I'm not sure how to handle it (except by commenting the whole `single-node-upgrade-patch-centos` step in 2.5.0), but we must not do patch version upgrade/downgrade tests if `product_patch_version_prev == product_version`.
Would be very nice to have the `doStepIf` for such cases.
Maybe we can leave it as is right now and comment the whole thing in 2.5.0, until we've a released patch version or a better way to activate/deactivate checks depending on variables.

Ebaneck: As agreed in person, we will deactivate upgrade testing for patch versions on the 2.5 branch since `doStepIf` is buggy at this time.

gdemonet: > Would be very nice to have the `doStepIf` for such cases.

Working on it :)

eve/main.yml
```diff
@@ -927,6 +928,129 @@ stages:
...
+            ISO_PATH: /tmp/metalk8s.iso
+      - ShellCommand:
+          <<: *bootstrap_config
+          env: *_env_prev_patch_version
+      # --- Install patch version N-1 ---
+      - ShellCommand:
+          <<: *run_bootstrap
+          env: *_env_prev_patch_version
+      - ShellCommand:
+          name: Provision Prometheus and AlertManager storage
```
alexandre-allard, line 1012: Not sure this step is really mandatory since we don't test anything after patch version installation, and you already create the volumes after the upgrade, before doing the tests.

eve/main.yml
```diff
@@ -927,6 +928,129 @@ stages:
...
+      - ShellCommand: &retrieve_prev_patch_version_iso_checksum
+          <<: *retrieve_iso_checksum
+          name: Retrieve previous PATCH ISO image checksum
+          env: &_env_previous_patch
+            DEST_DIR: "/tmp"
+            BASE_URL: "http://artifacts/builds/github:scality:metalk8s:\
+                      promoted-%(prop:product_version_prev_patch)s"
+      - ShellCommand: &retrieve_prev_patch_version_iso
+          <<: *retrieve_iso
+          name: Retrieve previous PATCH ISO image
```
alexandre-allard: detail: be consistent, use either `PATCH`, `Patch` or `patch` everywhere.

alexandre-allard (dismissed):
details, otherwise LGTM.

eve/main.yml
Ebaneck: Yes, it is not mandatory.
But if we finish the installation and try to upgrade immediately, we endup with cluster resources which are not ready.
BTW, during downgrade for patch versions, we will re-use this step

eve/main.yml
```diff
@@ -927,6 +928,131 @@ stages:
...
+          name: Set the patch version flag
+          property: patch_present
+          env:
+            PRODUCT_VERSION: "%(prop:product_version)s"
+            PRODUCT_VERSION_PREV_patch: "%(prop:product_version_prev_patch)s"
+          command: |-
+              if [ "$PRODUCT_VERSION" != "$PRODUCT_VERSION_PREV_patch" ]; then
+                 echo "true"
+              else
+                echo "false"
```
alexandre-allard: detail: indentation is not consistent here

Ebaneck: merde!

alexandre-allard (dismissed):

alexandre-allard (approved):

PR #2096, opened 2019-12-04
Generate Eve's YAML build plan from Python

gdemonet:
## Work In Progress

Plan is to provide tooling for generating the `eve/main.yml` build plan that Eve requires.
Goal is to be faster when adjusting this build plan, and add more complex logic in there that pure YAML would have trouble with.

(dump of the current state of cleaned-up commits, will mark PR as ready once the bulk of `eve/main.yml` is properly rendered)


deploy/README.md
```diff
@@ -0,0 +1,67 @@
...
+_OpenStack_:
+
+- `openstack_image_name` (default `"CentOS-7-x86_64-GenericCloud-1809.qcow2"`):
+  Image name to use for all spawned VMs.
+- `openstack_flavour_name` (default `"m1.medium"`): VM flavour to use.
+
+_Networks_:
+
+- `default_network_name` (default `"tenantnetwork1"`): The default network to
+  use for remote SSH access (from your host and from the Bastion).
```
slaperche-scality, line 25: Is `Bastion` a standard/well-know term? If not, could be worth have a definition somewhere.

deploy/README.md
```diff
@@ -0,0 +1,67 @@
...
+- `use_private_networks` (default `true`): Create and use two private networks
+  (instead of the default network) for both workload- and control-plane. Values
```
slaperche-scality, line 27: Which `values`? It's a bit ambiguous, is the value of this option or the value for the networks?

deploy/scripts/create-volumes.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Note: this script's *defaults* must adhere to the Vagrant environment in
```
slaperche-scality, line 3: Not sure refering to Vagrant is relevant here since it's about deploying a cluster "in the Cloud".

eve/buildplan/core.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(self):
+        self._stages = []
+
+    stages = property(operator.attrgetter("_stages"))
+
+    def add(self, stage):
+        existing_stage = next(
+            (s for s in self.stages if s.name == stage.name), None
+        )
```
slaperche-scality: I wonder if you could use an `OrderedSet` instead of a list here?
(yeah, there is no `OrderedSet`, but an `OrderedMap` with `None` value could do I guess)

eve/buildplan/core.py
```diff
@@ -0,0 +1,162 @@
...
+    """A collection of Steps to run on a Worker."""
+
+    def __init__(self, name, worker, steps, branches=None):
+        self._name = name
+        self._worker = worker
+        self._steps = steps
+        self._branches = branches or []
+
+    def __hash__(self):
+        return hash(yamlprint.dump(self.dump()))
```
slaperche-scality: a bit heavy to do a YAML serialization to compute the hash :/

I understand that you can't hash directly the objects (because `OrderedDict` is not hashable), but maybe using `json.dumps(self.dump(), sort_keys=True)` is a bit less expensive.

eve/buildplan/core.py
```diff
@@ -0,0 +1,162 @@
...
+            ]
+        )
+
+
+# Workers {{{
+class Worker(metaclass=abc.ABCMeta):
+    """A worker defines how/where to run a stage."""
+
+    @abc.abstractmethod
+    def dump(self):
```
slaperche-scality: Depends on your version of Python, but:

```
    @abc.abstractmethod
    def dump(self):
        """"here document the interface."""
```
works (and thus, no need for the nocover).

eve/buildplan/core.py
```diff
@@ -0,0 +1,162 @@
...
+        return collections.OrderedDict([("type", "local")])
+
+
+# }}}
+# Steps {{{
+class Step(metaclass=abc.ABCMeta):
+    """A step describes what needs to be executed within a stage."""
+
+    # TODO: consider changing halt_on_failure default value to True
+    def __init__(self, name, halt_on_failure=None, always_run=None):
```
slaperche-scality: Why `None` instead of `False` if `halt_on_failure` is a Boolean?
Ditto for `always_run`.

I guess `None` means `default to whatever is the default in Eve` which may not be very reproducible if Eve ever changes its default values.

eve/buildplan/test_core.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class TestProject:
+    def test_add(self):
+        project = core.Project()
+
+        stage = core.Stage(name="simple", worker=core.LocalWorker(), steps=[])
+
+        project.add(stage)
+        assert sum(stage.name == s.name for s in project.stages) == 1
```
slaperche-scality: maybe an `assert len(filter(…, project.stages)) == 1` (or using list comprehension) would be more explicit that relying on the fact that `True` is automagically casted to `1`

eve/buildplan/.coveragerc
```diff
@@ -1,2 +1,6 @@
 [run]
-omit=*/test_*.py
+omit=
+    # don't cover test files
```
slaperche-scality: Why? Would allow to catch dead code in tests, no?

eve/buildplan/main.py
```diff
@@ -0,0 +1,36 @@
+"""Generate the Eve YAML description of build plan for MetalK8s."""
```
slaperche-scality, line 1: Shouldn't you have a shebang to at least clarify if you're using Python 2 or 3?

eve/buildplan/main.py
```diff
@@ -0,0 +1,36 @@
...
+            "documentation/*",
+            "release/*",
+        ],
+        steps=[],
+    )
+
+
+if __name__ == "__main__":
+    build_plan = build_project().dump()
+    yamlprint.dump(build_plan, stream=sys.stdout)
```
slaperche-scality, line 429: instead of hardcoding `stdout`, take a filename as input from the CLI (default to `stdout` if not provided if you want)
That way it's easier to test, and you don't need the `bash -c` in tox (because you no longer need shell redirection).

eve/buildplan/main.py
```diff
@@ -58,7 +60,16 @@ def build():
...
-        steps=[],
+        steps=[
+            build_all(),
+            *dsl.copy_artifacts(
+                [
+                    "_build/metalk8s.iso",
+                    "_build/SHA256SUM",
+                    "_build/root/product.txt",
+                ]
+            ),
```
slaperche-scality, line 76: Couldn't this be handled by the `WithArtifacts` decorator?

eve/main.yml
```diff
@@ -82,7 +82,7 @@ stages:
           _build/root/product.txt; do cp -r "$artifact" artifacts; done
     - Upload:
         name: Upload artifacts
-        source: artifacts
+        source: &id002 artifacts
```
slaperche-scality, line 113: minor: `id002`? Cant' we generate better name (using `name` above`)?

eve/buildplan/buildplan/shell.py
```diff
@@ -0,0 +1,84 @@
+"""Helpers to manipulate `core.ShellCommand`."""
+
+from buildplan import core
+
+
+# Helpers {{{
+def _fmt_args(*args, join_with=" "):
+    """Join an iterable of `Optional[str]`, omitting `None`s."""
```
slaperche-scality, line 8: Any reason why `None` instead of falsy value (i.e. None and empty string)?

eve/buildplan/buildplan/dsl/artifacts.py
```diff
@@ -0,0 +1,79 @@
+from collections import Mapping
```
slaperche-scality, line 1: I would have done the split into a dedicated commit, because here it's hard to tell the difference between the new code that we should review and the old code that is simply moved around.

tox.ini
```diff
@@ -88,11 +88,11 @@ commands =
 
 [testenv:lint-yaml]
 description =
-    Lint Salt and Eve YAML files using yamllint.
+    Lint Salt YAML files using yamllint.
 deps =
     yamllint==1.15.0
 commands =
-    bash -c "yamllint eve/main.yml salt/metalk8s/defaults.yaml"
+    bash -c "yamllint salt"
```
slaperche-scality, line 95: Why aren't we linting the `eve/main.yaml`?

eve/buildplan/buildplan/core.py
```diff
@@ -260,6 +260,7 @@ class ShellCommand(Step):
         workdir="workdir",
         sigterm_time="sigtermTime",
         use_pty="usePTY",
+        timeout="timeout",
```
slaperche-scality, line 263: `ci/buildplan; ` => should be `:` (in the commit message)

eve/buildplan/buildplan/core.py
```diff
@@ -273,10 +274,8 @@ def dump_arguments(self):
 
 
 class Upload(Step):
-    OPTIONAL_ARGS = dict(
-        Step.OPTIONAL_ARGS,
-        urls="urls",
-    )
+    OPTIONAL_ARGS = dict(Step.OPTIONAL_ARGS, urls="urls")
```
slaperche-scality, line 277: unrelated change :stuck_out_tongue:

Issue #2074, opened 2019-11-25
Add tests for OIDC authentication flow in the CI


Ebaneck:
**Component**:

'salt', 'kubernetes', 'tests'

**Why this is needed**:

We need to add some tests inthe CI to ensure that authentication via OIDC flow works as expected.

**What should be done**:

We could add the following tests:

- Request the openid-configuration located at <<https://{{ control_plane_address }}:8443/oidc/.well-known/openid-configuration>> and verify that the issuer_url is valid
- Deploy an `openLDAP` container in the CI and verify authentication with MetalK8s works.

**Implementation proposal** (strongly recommended):

**Test plan**:

Tests :)

PR #2098, opened 2019-12-05
Add tests for OIDC authentication flow

Ebaneck:
**Component**:

'tests'

**Context**: 

See #2074 

**Summary**:

Ensure OIDC service works as expected

**Acceptance criteria**: 

All tests should pass
---

Closes: #2074

### DIFF
--- a/tests/post/features/authentication.feature
+++ b/tests/post/features/authentication.feature
@@ -1,0 +1,37 @@
+@post @local @ci @authentication
+Feature: Authentication is up and running
+    Scenario: List Pods
+        Given the Kubernetes API is available
+        Then the 'pods' list should not be empty in the 'metalk8s-auth' namespace
+
+    Scenario: Expected Pods
+        Given the Kubernetes API is available
+        And pods with label 'release=nginx-ingress-control-plane' are 'Ready'
+        Then we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth' on node 'bootstrap'
+
+    Scenario: Reach the OpenID Config
+        Given the Kubernetes API is available
+        And the control-plane Ingress path '/oidc' is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        Then we can reach the OIDC openID configuration
+
+    Scenario: Access HTTPS service
+        Given the Kubernetes API is available
+        And the control-plane Ingress path '/oidc' is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        When we perform a request on '/oidc/' with port '8443' on control-plane IP
+        Then the server returns '404' with message '404 page not found'
+
+    Scenario: Login to Dex using incorrect email
+        Given the Kubernetes API is available
+        And the control-plane Ingress path '/oidc' is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        When we login to Dex as 'admin@metalk8s.com' using password 'password'
+        Then authentication fails with login error
+
+    Scenario: Login to Dex using correct email and password
+        Given the Kubernetes API is available
+        And the control-plane Ingress path '/oidc' is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        When we login to Dex as 'admin@metalk8s.invalid' using password 'password'
+        Then the server returns '303' with an ID token
--- a/tests/post/steps/test_authentication.py
+++ b/tests/post/steps/test_authentication.py
@@ -1,0 +1,257 @@
+import json
+import re
+
+import requests
+import requests.exceptions
+
+import pytest
+from pytest_bdd import scenario, given, then, when, parsers
+
+import kubernetes.client
+from kubernetes.client.rest import ApiException
+
+from tests import kube_utils
+from tests import utils
+
+
+# Constants {{{
+
+INGRESS_PORT = 8443
+
+# }}}
+
+
+# Scenarios {{{
+
+@scenario('../features/authentication.feature', 'List Pods')
+def test_list_pods(host):
+    pass
+
+
+@scenario('../features/authentication.feature', 'Expected Pods')
+def test_expected_pods(host):
+    pass
+
+
+@scenario('../features/authentication.feature', 'Reach the OpenID Config')
+def test_reach_openid_config(host):
+    pass
+
+
+@scenario('../features/authentication.feature', 'Access HTTPS service')
+def test_access_https_service(host):
+    pass
+
+
+@scenario('../features/authentication.feature',
+          'Login to Dex using incorrect email')
+def test_failed_login(host):
+    pass
+
+
+@scenario('../features/authentication.feature',
+          'Login to Dex using correct email and password')
+def test_login(host):
+    pass
+
+
+# }}}
+
+
+# Fixtures {{{
+
+@pytest.fixture(scope='function')
+def context():
+    return {}
+
+
+@pytest.fixture(scope='function')
+def control_plane_ip(host):
+    with host.sudo():
+        output = host.check_output(' '.join([
+            'salt-call', '--local', '--out=json',
+            'grains.get', 'metalk8s:control_plane_ip',
+        ]))
+        ip = json.loads(output)['local']
+    return ip
+
+
+# }}}
+
+
+# Given {{{
+
+@given(parsers.parse(
+    "the control-plane Ingress path '{path}' is available"))
+def check_cp_ingress_pod(request, host, k8s_client, control_plane_ip):
+    ssh_config = request.config.getoption('--ssh-config')
+    label = "release=nginx-ingress-control-plane"
+
+    # Ensure nginx ingress control-plane controller is running
+    def _wait_for_ingress():
+        namespace = "metalk8s-ingress"
+        pods = kube_utils.get_pods(
+            k8s_client, ssh_config, label, namespace=namespace
+        )
+        assert len(pods) > 0
+    utils.retry(
+        _wait_for_ingress,
+        times=10,
+        wait=3,
+        name="wait for pod labeled '{}'".format(label)
+    )
+    # Todo: use the ingress object to check it's addresses
+
+
+# }}}
+
+
+# When {{{
+
+
+@when(parsers.parse(
+    "we perform a request on '{path}' with port '{port}' on control-plane IP"))
+def perform_request(host, context, control_plane_ip, path, port):
+    try:
+        context['response'] = requests.get(
+            'https://{ip}:{port}{path}'.format(
+                ip=control_plane_ip, port=port, path=path
+            ),
+            verify=False,
+        )
+    except requests.exceptions.ConnectionError as exc:
+        pytest.fail(
+            "Failed to access oidc url path with error: {}".format(exc)
+        )
+
+
+@when(parsers.parse(
+    "we login to Dex as '{username}' using password '{password}'"))
+def dex_login(host, control_plane_ip, username, password, context):
+    context['login_response'] = _dex_auth_request(
+        control_plane_ip, username, password
+    )
+
+
+# }}}
+
+
+# Then {{{
+
+
+@then("we can reach the OIDC openID configuration")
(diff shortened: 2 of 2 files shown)


tests/post/features/authentication.feature
```diff
@@ -0,0 +1,41 @@
...
+
+#Cannot test this since we use an implicit oidc flow and need an in-browser redirect
+# if we consider out-of-browser redirects using (urn:ietf:wg:oauth:2.0:oob), `requests` library raises schema exception
+# wouldn't spend much time here except otherwise a better approach is suggested by a Good Samaritan
+    Scenario: Access Node resources using Dex static user
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        When we login to Dex as 'admin@metalk8s.invalid' using password 'password'
+        Then the server returns '303' with an ID token
+        And we can use the ID token to list cluster resources
```
gdemonet: I didn't think we were going to test the whole authentication flow in Python :/

IMO, getting Cypress to login was sufficient for a first iteration. Implementing OIDC from the CLI only is not possible, we would need some way of working around the lack of actual client with a callback URL. Maybe we could set something up for the sake of a test, but that would require much more efforts than using Metal UI...

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def _get_control_plane_ip(host):
+    with host.sudo():
+        output = host.check_output(' '.join([
+            'salt-call', '--local', '--out=json',
+            'grains.get', 'metalk8s:control_plane_ip',
+        ]))
+        ip = json.loads(output)['local']
+    return ip
```
gdemonet: This takes a long time to evaluate, let's make it a `@pytest.fixture(scope='session')`

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def _dex_auth_request(control_plane_ip, username, password):
+    try:
+        response = requests.post(
+            'https://{}:{}/oidc/auth?'.format(control_plane_ip, INGRESS_PORT),
+            data={
+                'response_type': 'id_token',
+                'client_id': 'metalk8s-cli',
```
gdemonet: Couldn't use `metalk8s-ui` client instead of adding a new one?

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+            ),
+            verify=False,
+        )
+    except requests.exceptions.ConnectionError as exc:
+        raise ConnectionError(exc)
+
+    assert response.status_code == 200
+    response_body = response.json()
+    # check for the existence of  keys[issuer, authorization_endpoint]
+    assert 'issuer' and 'authorization_endpoint' in response_body
```
gdemonet: Would make sense to actually expect the full config, as we know statically what it **should** be.

tests/post/features/authentication.feature
```diff
@@ -0,0 +1,41 @@
...
+
+    Scenario: Login to Dex using correct email and password
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        When we login to Dex as 'admin@metalk8s.invalid' using password 'password'
+        Then the server returns '303' with an ID token
+
+#Cannot test this since we use an implicit oidc flow and need an in-browser redirect
+# if we consider out-of-browser redirects using (urn:ietf:wg:oauth:2.0:oob), `requests` library raises schema exception
+# wouldn't spend much time here except otherwise a better approach is suggested by a Good Samaritan
```
gdemonet: So yeah, let's not spend time on this, you checked that we can access Dex, that its config is valid, even went far enough to check the default user credentials are understood by Dex, now that should be enough. Let's leave the whole flow to Cypress tests.

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -57,6 +58,12 @@ stringData:
       - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%
         raw %}'
       secret: 4lqK98NcsWG5qBRHJUqYM1
+    - id: metalk8s-cli
+      name: MetalK8s CLI
+      redirectURIs:
+      - http://localhost
+      - http://127.0.0.1
+      secret: jfakfka9i32242k4j2k42j42249kah
```
gdemonet: Let's not add "virtual" clients if possible - these tests should work with `metalk8s-ui` I believe.

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+    ip = _get_control_plane_ip(host)
+    try:
+        response = requests.get(
+            'https://{ip}:8443/oidc/.well-known/openid-configuration'.format(
+                ip=ip
+            ),
+            verify=False,
+        )
+    except requests.exceptions.ConnectionError as exc:
+        raise ConnectionError(exc)
```
gdemonet: Is `ConnectionError` in stdlib? What does it mean? I'd rather have a `pytest.fail` here.

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+    assert auth_response.headers.get('location') is None
+
+
+@then(parsers.parse("the server returns '{status_code}' with an ID token"))
+def successful_login(host, context, status_code):
+    auth_response = context.get('login_response')
+    if auth_response.text is None:
+        assert False
+    assert auth_response.status_code == int(status_code)
+    assert auth_response.headers.get('location') is not None
```
gdemonet, line 199: Can we check the value? It should be the callback URL, right?

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+            },
+            verify=False,
+        )
+    except requests.exceptions.ConnectionError as exc:
+        raise ConnectionError(exc)
+
+    auth_request = response.text  # response is an html form
+    # Obtain the request id using regex search from the return html form
+    # form action looks like:
+    # <a href="/oidc/auth/local?req=ovc5qdll5zznlubewjok266rl" target="_self">
```
gdemonet, line 231: Maybe take the whole `href` in your test then...

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,207 @@
...
+
+# Then {{{
+
+
+@then("we can reach the OIDC openID configuration")
+def reach_openid_config(host):
+    ip = _get_control_plane_ip(host)
+    try:
+        response = requests.get(
+            'https://{ip}:8443/oidc/.well-known/openid-configuration'.format(
```
gdemonet: Use the `INGRESS_PORT` constant here

gdemonet (changes requested):

tests/post/steps/test_authentication.py
Ebaneck: It is actually an `Approval URL`. It only occurs when login was successful and Redirect(callback) has happened. But since we do not have a proper in-browser or out-of-browser redirect, we can simply check that this field exist when login passes.

tests/post/steps/test_authentication.py
Ebaneck: We cannot access the 'module' scoped fixture 'host' with a 'session' scoped request object so i am going to leave it as a 'function' fixture

tests/post/steps/test_authentication.py
Ebaneck: Suggesting we `assert` for all the `11` JSON keys present in the OpenID configuration?? 😞
Maybe this is sufficient??
```   
    assert 'issuer' and 'authorization_endpoint' in response_body
    assert response_body.get('issuer') == 'https://{}:{}/oidc'.format(
        control_plane_ip, INGRESS_PORT
    )
    assert response_body.get(
        'authorization_endpoint') == 'https://{}:{}/oidc/auth'.format(
        control_plane_ip, INGRESS_PORT
    )
```

gdemonet: Was thinking about something more compact:

```python
basepath = "https://{}:{}/oidc".format(control_plane_ip, INGRESS_PORT)
assert response_body == {
    "issuer": f"{basepath}",
    "authorization_endpoint": f"{basepath}/auth",
    # other keys are easy to add
}
```

But fine, let's not go too far anyway

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,224 @@
...
+        pytest.fail("Dex authentication request failed with error: {}".format(
+            exc
+            )
+        )
+
+    auth_request = response.text  # response is an html form
+    # form action looks like:
+    # <a href="/oidc/auth/local?req=ovc5qdll5zznlubewjok266rl" target="_self">
+    try:
+        req = re.search('href=[\'"].oidc.([^\'" >]+)', auth_request).group(0)
```
gdemonet: Regex could be slightly improved:

```python
reqpath = re.search(r'href=[\'"](?P<reqpath>/oidc/\S+)[\'"] ', auth_request).group('reqpath')
```

gdemonet (commented):
Just a last comment on the parsing of the 303 HTML response. You may need to wait for Ingress to be ready as well before performing requests.

Ebaneck:
> Just a last comment on the parsing of the 303 HTML response. You may need to wait for Ingress to be ready as well before performing requests.

Looks like we cannot have multiple `Givens` that checks the readiness of pods in each scenario. Pytest will fail indiscriminately with 
```
pytest_bdd.exceptions.GivenAlreadyUsed: Fixture "check_pod_status" that implements this "pods with label 'app.kubernetes.io/name=dex' are 'Ready'" given step has been already used.
```
I tried something like 

```
And pods with label 'release=nginx-ingress-control-plane' are 'Ready'
And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
```
So we check for readiness of Ingress in the first request only as seen in the scenario file of this PR.

Also i realised that the only specific label that differentiates nginx ingress and nginx-ingress-control-plane pod is the `release`.

tests/post/steps/test_authentication.py
```diff
@@ -0,0 +1,232 @@
...
+                'https://{ip}:{port}{path}'.format(
+                    ip=control_plane_ip, port=port, path=path
+                ),
+                verify=False,
+            )
+        except requests.exceptions.ConnectionError as exc:
+            pytest.fail(
+                "Failed to access oidc url path with error: {}".format(exc)
+            )
+    utils.retry(_get_urlpath_resource, times=10, wait=3)
```
gdemonet: I think there shouldn't be a `ConnectionError` here, can't we wait for `ingress` to be ready?

tests/post/features/authentication.feature
```diff
@@ -0,0 +1,34 @@
...
+        Then the 'pods' list should not be empty in the 'metalk8s-auth' namespace
+
+    Scenario: Expected Pods
+        Given the Kubernetes API is available
+        And pods with label 'release=nginx-ingress-control-plane' are 'Ready'
+        Then we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth' on node 'bootstrap'
+
+    Scenario: Reach the OpenID Config
+        Given the Kubernetes API is available
+        And the control-plane Ingress path '/oidc' is available
```
gdemonet, line 14: Would make sense to also use it in other scenarii that rely on it (Access HTTPS service, Login to Dex using incorrect email, Login to Dex using correct email and password)

gdemonet (changes requested):
Should be good once we share this check (each feature scenario should be executable independently).

What could be interesting, in the future, is to use the `Background` feature of Gherkin (see [here](https://pytest-bdd.readthedocs.io/en/stable/#backgrounds)).

Ebaneck:
> Should be good once we share this check (each feature scenario should be executable independently).
> 
> What could be interesting, in the future, is to use the `Background` feature of Gherkin (see [here](https://pytest-bdd.readthedocs.io/en/stable/#backgrounds)).

I would want to think the deadlock situation of a `Given` already being used will come up.
It remains to be tested!

gdemonet (approved):

Issue #2083, opened 2019-11-27
etcd initial-cluster Contains Duplicate Entries


bkettler:
After expanding from a 2 node MetalK8s cluster to a 3 node MetalK8s the initial-cluster value in etcd.yaml contains duplicate URLs on the new host.

`        - --initial-cluster==https://10.25.144.162:2380,app-1=https://10.25.144.151:2380,storage-1=https://10.25.144.161:2380,storage-2=https://10.25.144.162:2380
`

When duplicate entries are defined the the etcd container will repeatedly fail to start with the following error.

`2019-11-27 17:29:40.350157 C | etcdmain: initial cluster =https://10.25.144.162:2380,app-1=https://10.25.144.151:2380,storage-1=https://10.25.144.161:2380,storage-2=https://10.25.144.162:2380 has duplicate url
`

Version : MetalK8s 2.4.1

gdemonet:
Thanks for the bug report!

It is a surprising issue, as the code we have on our side is supposed to handle deduplication. It seems something went wrong in a strange way, as the duplicated peer is definitely missing half of the declaration (no name before the `=` sign).

We will add an extra safeguard, but I don't understand why it happened. Anything peculiar in your setup that might give us a hint @bkettler?

NicolasT:
The duplicate IP address in there is quite surprising as well...

NicolasT:
@bkettler What could help us to debug: could you execute the following and provide the output of:

```sh
salt-call pillar.items metalk8s:etcd:members
```
(as `root`) on your 'bootstrap' node, as well as something along the following lines:

```
[root@bootstrap ~]# kubectl -n kube-system get pods -l component=etcd
NAME             READY   STATUS    RESTARTS   AGE
etcd-bootstrap   1/1     Running   4          2d

[root@bootstrap ~]# kubectl -n kube-system exec -ti etcd-bootstrap sh -- -c "ETCDCTL_API=3 etcdctl --endpoints=https://[127.0.0.1]:2379 --cacert=/etc/kubernetes/pki/etcd/ca.crt --cert=/etc/kubernetes/pki/etcd/healthcheck-client.crt --key=/etc/kubernetes/pki/etcd/healthcheck-client.key member list"
37bacaf3fa2d1df9, started, bootstrap, https://172.21.254.13:2380, https://172.21.254.13:2379
```
where in the second command, you replace `etcd-bootstrap` with the Pod name as found in the output of the first command.

Thanks!

[...]

PR #2099, opened 2019-12-05
salt: Store more etcd members informations in pillar (and not use not started etcd)

TeddyAndrieux:
**Component**:

'salt'

**Summary**: 

Add all etcd member informations in the pillar retrived from the etcd3
API and use them to compute the `initial-cluster` value for etcd
manifest (filtering out the member with empty name)

---

Fixes: #2083

### DIFF
--- a/salt/_modules/metalk8s_etcd.py
+++ b/salt/_modules/metalk8s_etcd.py
@@ -176,14 +176,14 @@ def check_etcd_health(
     else:
         return "cluster is healthy"
 
+
 def get_etcd_member_list(
         endpoint=None,
         nodes=None,
         ca_cert='/etc/kubernetes/pki/etcd/ca.crt',
         cert_key='/etc/kubernetes/pki/etcd/salt-master-etcd-client.key',
         cert_cert='/etc/kubernetes/pki/etcd/salt-master-etcd-client.crt'):
-    '''Get the list of etcd members peer urls using the python etcd3 client.'''
-    member_peer_urls = []
+    '''Get the list of etcd members using the python etcd3 client.'''
     if not endpoint:
         # If we have no endpoint get it from mine
         try:
@@ -194,14 +194,18 @@ def get_etcd_member_list(
                 cert_cert=cert_cert
             )
         except:
-            return member_peer_urls
+            return []
+
     with etcd3.client(host=endpoint,
                       ca_cert=ca_cert,
                       cert_key=cert_key,
                       cert_cert=cert_cert,
                       timeout=TIMEOUT) as etcd:
-        for member in etcd.members:
-            member_peer_urls.append('{0}={1}'.format(
-                member.name,
-                member.peer_urls[0]))
-    return list(set(member_peer_urls))
+        return [
+            {
+                'id': member.id,
+                'name': member.name,
+                'peer_urls': list(member.peer_urls),
+                'client_urls': list(member.client_urls)
+            } for member in etcd.members
+        ]
--- a/salt/metalk8s/kubernetes/etcd/installed.sls
+++ b/salt/metalk8s/kubernetes/etcd/installed.sls
@@ -4,19 +4,31 @@ include:
   - metalk8s.kubernetes.ca.etcd.advertised
   - .certs
 
-{%- set host_name = grains['id'] %}
-{%- set host = grains['metalk8s']['control_plane_ip'] %}
+{%- set node_name = grains['id'] %}
+{%- set node_ip = grains['metalk8s']['control_plane_ip'] %}
 
-{%- set endpoint  = host_name ~ '=https://' ~ host ~ ':2380' %}
+{%- set endpoint = 'https://' ~ node_ip ~ ':2380' %}
 
-{#- Get the list of existing etcd node. #}
-{%- set etcd_endpoints = pillar.metalk8s.etcd.members %}
+{#- Get the list of existing etcd member. #}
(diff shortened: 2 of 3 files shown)


salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -4,19 +4,33 @@ include:
...
+{%- set endpoint = 'https://' ~ node_ip ~ ':2380' %}
 
-{#- Get the list of existing etcd node. #}
-{%- set etcd_endpoints = pillar.metalk8s.etcd.members %}
+{#- Get the list of existing etcd member. #}
+{%- set etcd_members = pillar.metalk8s.etcd.members %}
 
 {#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_endpoints else "new" %}
+{%- set state = "existing" if etcd_members else "new" %}
```
NicolasT, line 16: Not sure this is the right approach (though indeed, it's not related to this PR/issue).

In essence, what we're saying here is "If all nodes are unavailable (which causes the member list to be empty), we consider this to be a new cluster", which is definitely not a given. Actually, only during very first bootstrap this can be the case, otherwise we're disaster-recovering or so.

Worth an issue IMO.

TeddyAndrieux: I agree it's not a safe approach
#2102

salt/_modules/metalk8s_etcd.py
```diff
@@ -194,14 +194,20 @@ def get_etcd_member_list(
...
-                member.name,
-                member.peer_urls[0]))
-    return list(set(member_peer_urls))
\ No newline at end of file
+            members_list.append({
+                'id': member.id,
+                'name': member.name,
+                'peer_urls': list(member.peer_urls),
+                'client_urls': list(member.client_urls)
+            })
```
gdemonet: nit: use a list comprehension

```python
with etcd3.client(**kwargs) as etcd:
    return [_dict_from_member(member) for member in etcd.members]
```

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -4,19 +4,33 @@ include:
...
 
-{#- Add ourselves to the list. #}
-{%- do etcd_endpoints.append(endpoint) %}
+{%- set etcd_endpoints = {} %}
+{%- for member in etcd_members %}
+  {#- NOTE: Filter out members with empty name as they are not started yet. #}
+  {%- if member['name'] %}
```
gdemonet: Jinja provides filters for this:

```jinja
{#- NOTE: Filter out members with empty name as they are not started yet. #}
{%- for member in etcd_members | selectattr("name") %}
```

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -4,19 +4,33 @@ include:
...
+    {#- NOTE: Only take first peer_urls for endpoint. #}
+    {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
+  {%- endif %}
+{%- endfor %}
+
+{#- Add ourselves to the endpoints. #}
+{%- do etcd_endpoints.update({node_name: endpoint}) %}
+
+{%- set etcd_initial_cluster = [] %}
+{%- for name, ep in etcd_endpoints.items() %}
```
gdemonet, line 29: Did we really need a second loop over a dict? Do we fear duplicate names?

gdemonet (dismissed):
Some nitpicks, but the functionality looks good to me :)

salt/metalk8s/kubernetes/etcd/installed.sls
TeddyAndrieux: Yes used this one when WIP, just thinking that an `if` is maybe simpler to understand but indeed I could change it

salt/metalk8s/kubernetes/etcd/installed.sls
TeddyAndrieux: I think it's better to not break the etcd cluster if we have one node that changed peer urls (since we only support one peer urls here)

Unfortunatly I didn't found any good method to do it without this second loop, if you have an idea do not hesitate
At the beginning just wanted to do something like that in python directly for `initial-cluster`
```
','.join('{}={}'.format(name, ip) for name, ip in etcd_endpoints.items())
```
But not doable in jinja without a custom filter

salt/_modules/metalk8s_etcd.py
TeddyAndrieux: Yes I can, I just kept the initial approach but I will change it

salt/metalk8s/kubernetes/etcd/installed.sls
gdemonet: My point was that we're supposed to have a unique set of members (with unique peer_urls, I'd imagine), so we could have just done:

```jinja
{%- set initial_cluster = [] %}
{%- for member in etcd_members | selectattr('name') %}
  {%- do initial_cluster.append(member['name'] ~ '=' ~ member['peer_urls'][0]) %}
{%- endfor %}

{%- do initial_cluster.append(node_name ~ '=' ~ endpoint %}
```

But it's really minor, and your approach is safer against duplicate values in `etcd_members`, so let's leave it as is.

gdemonet (approved):

PR #2100, opened 2019-12-06
deploy nodes with terraform

alexandre-allard:
**Component**: terraform

**Context**: Metalk8s deployment with terraform

**Summary**: Add possiblity to deploy nodes using terraform (for now only bootstrap, but will add the expansion really soon)

**Acceptance criteria**: Nothing is broken ?
Ability to deploy a cluster with the bootstrap node installed:
`terraform apply -auto-approve -var=metalk8s_iso=/path/to/metalk8s.iso`

### DIFF
--- a/eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
+++ b/eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
@@ -1,0 +1,60 @@
+variable "metalk8s_iso" {
+  type        = string
+  description = "Metalk8s ISO file path to upload to bootstrap node"
+  default     = ""
+}
+
+variable "metalk8s_iso_dest" {
+  type        = string
+  description = "Location where the Metalk8s ISO will be copied on the bootstrap node"
+  default     = "/home/centos/metalk8s.iso"
+}
+
+variable "metalk8s_iso_mountpoint" {
+  type        = string
+  description = "Mountpoint for the Metalk8s ISO"
+  default     = "/var/tmp/metalk8s"
+}
+
+resource "null_resource" "metalk8s_upload_iso" {
+  count = "${var.metalk8s_iso != "" ? 1 : 0}"
+
+  connection {
+    host        = openstack_compute_instance_v2.bootstrap.access_ip_v4
+    type        = "ssh"
+    user        = "centos"
+    private_key = file("~/.ssh/terraform")
+  }
+
+  depends_on = [
+    openstack_compute_instance_v2.bootstrap,
+  ]
+
+  provisioner "file" {
+    source      = "${var.metalk8s_iso}"
+    destination = "${var.metalk8s_iso_dest}"
+  }
+}
+
+resource "null_resource" "metalk8s_run_bootstrap" {
+  count = "${var.metalk8s_iso != "" ? 1 : 0}"
+
+  connection {
+    host        = openstack_compute_instance_v2.bootstrap.access_ip_v4
+    type        = "ssh"
+    user        = "centos"
+    private_key = file("~/.ssh/terraform")
+  }
+
+  depends_on = [
+    null_resource.metalk8s_upload_iso,
+  ]
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p ${var.metalk8s_iso_mountpoint}",
+      "sudo mount -o loop ${var.metalk8s_iso_dest} ${var.metalk8s_iso_mountpoint}",
+      "sudo ${var.metalk8s_iso_mountpoint}/bootstrap.sh --verbose",
+    ]
+  }
+}
--- a/eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
+++ b/eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
@@ -58,3 +58,34 @@ resource "null_resource" "metalk8s_run_bootstrap" {
     ]
   }
 }
+
+variable "metalk8s_provision_volumes" {
+  type    = bool
+  default = false
+}
+
+resource "null_resource" "metalk8s_provision_volumes" {
+  count = "${var.metalk8s_provision_volumes ? 1 : 0}"
+
+  connection {
+    host        = openstack_compute_instance_v2.bootstrap.access_ip_v4
+    type        = "ssh"
+    user        = "centos"
+    private_key = file("~/.ssh/terraform")
+  }
+
+  depends_on = [
+    null_resource.metalk8s_run_bootstrap
+  ]
+
+  provisioner "file" {
+    source      = "${path.module}/../../../create-volumes.sh"
+    destination = "/home/centos/create-volumes.sh"
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo PRODUCT_TXT=${var.metalk8s_iso_mountpoint}/product.txt PRODUCT_MOUNT=${var.metalk8s_iso_mountpoint} bash create-volumes.sh",
+    ]
+  }
+}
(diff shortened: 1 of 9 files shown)


eve/workers/openstack-multiple-nodes/terraform/scripts/network-iface-config.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
NicolasT, line 1: Can't we do all of this through `cloud-init`?

eve/workers/openstack-multiple-nodes/terraform/network.tf
```diff
@@ -94,3 +104,87 @@ resource "openstack_networking_secgroup_v2" "bastion" {
...
+
+  security_group_ids = [
+    openstack_networking_secgroup_v2.control_plane.id,
+  ]
+
+  fixed_ip {
+    subnet_id  = openstack_networking_subnet_v2.control_plane.id
+  }
+
+  allowed_address_pairs {
```
NicolasT, line 171: Instead of all this mockery, if we have a private control-plane network, can't we simply get rid of port-security on it, create a port for the VIP so it's not being handed out by DHCP, and get rid of IPIP?

eve/workers/openstack-multiple-nodes/terraform/network.tf
```diff
@@ -6,6 +6,17 @@ variable "openstack_network" {
   }
 }
 
+locals {
+  control_plane_network = {
+    name    = "${local.prefix}-control-plane"
+    cidr    = "192.168.1.0/28"
+    gateway = "192.168.1.14"
```
NicolasT: It's counter-intuitive to have the gateway address as the last address (before broadcast) in a network, in general it'd be the first, but sure this works :man_shrugging:

eve/workers/openstack-multiple-nodes/terraform/scripts/network-iface-config.sh
alexandre-allard: Yes, that's exactly what the comment below says, I don't know yet how to do it through terraform, but I'll search.
The script is more to make the rest of the stuff work until I found the right solution.

eve/workers/openstack-multiple-nodes/terraform/network.tf
alexandre-allard: It's only a detail, but I'll change it if it bother you.

eve/workers/openstack-multiple-nodes/terraform/network.tf
alexandre-allard-scality: We can already get rid of IPIP with the current changes, but I'll try so simplify this as you suggested.

alexandre-allard-scality: I tried to disable port security, but I never managed to make it work.
Every time, I end up with some (not really useful) errors when spawning instances, like this one:
```
Error: Error waiting for instance (37abba31-95c7-47df-8be5-bbf1ae8e6033) to become ready: unexpected state 'ERROR', wanted target 'ACTIVE'. last error: %!s(<nil>)
```
I'm still investigating... but I don't know if it worth spending too much time.
Still, IPIP can be removed with these changes.

eve/workers/openstack-multiple-nodes/terraform/network.tf
```diff
@@ -94,3 +104,87 @@ resource "openstack_networking_secgroup_v2" "bastion" {
...
+
+resource "openstack_networking_network_v2" "control_plane" {
+  name = local.control_plane_network.name
+}
+
+resource "openstack_networking_subnet_v2" "control_plane" {
+  name        = "${local.prefix}-control-plane"
+  network_id  = openstack_networking_network_v2.control_plane.id
+  enable_dhcp = true
+  cidr        = "${local.control_plane_network.cidr}"
```
gdemonet: Minor: now that we use Terraform 0.12, you can just write:

```
  cidr        = local.control_plane_network.cidr
```

Same for other occurences below

eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
```diff
@@ -0,0 +1,60 @@
...
+  type        = string
+  description = "Location where the Metalk8s ISO will be copied on the bootstrap node"
+  default     = "/home/centos/metalk8s.iso"
+}
+
+variable "metalk8s_iso_mountpoint" {
+  type        = string
+  description = "Mountpoint for the Metalk8s ISO"
+  default     = "/var/tmp/metalk8s"
+}
```
gdemonet, line 17: Neat! Even if we may not really need it in the CI, it will be helpful for sharing with users :+1:

eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
```diff
@@ -58,3 +58,34 @@ resource "null_resource" "metalk8s_run_bootstrap" {
     ]
   }
 }
+
+variable "metalk8s_provision_volumes" {
+  type    = bool
+  default = false
+}
```
gdemonet, line 65: For future improvement: we could (should?) mount an OpenStack volume to store `SparseLoopDevice`s backing files, and maybe let users choose the total size of this volume.

eve/workers/openstack-multiple-nodes/terraform/ssh.tf
```diff
@@ -46,3 +50,10 @@ resource "null_resource" "ssh_config" {
     )
   }
 }
+
+# Generate Bastion SSH keypair
+resource "null_resource" "bastion_ssh_keys" {
+  provisioner "local-exec" {
+    command  = "ssh-keygen -t rsa -b 4096 -N '' -f ${local.bastion_ssh_key}"
+  }
+}
```
gdemonet, line 59: Didn't want to generate it on the host, but yeah, makes sense to do it that way.

gdemonet (dismissed):
Just a few minor nitpicks, most of it looks good to me! I'll make sure to use these recent changes in my PR as well, let's merge this one :)

eve/workers/openstack-multiple-nodes/terraform/network.tf
alexandre-allard: Nice, will do.

eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
alexandre-allard: Yep, that's the goal

eve/workers/openstack-multiple-nodes/terraform/metalk8s.tf
alexandre-allard: OpenStack volume ? You want to bring flakies ?
Yes, could be a good idea.

gdemonet (dismissed):

tests/post/steps/test_salt_api.py
```diff
@@ -48,7 +48,8 @@ def login_salt_api(host, username, password, version, context, request):
     ])
     with host.sudo():
         cmd_output = host.check_output(cmd_ip)
-    ip = json.loads(cmd_output)['local'][0]
+    # We take the last IP, otherwise we can have the control plane VIP instead
+    ip = json.loads(cmd_output)['local'][-1]
```
gdemonet: Isn't it the opposite in Vagrant?

alexandre-allard: Just checked, looks like it is...

gdemonet: We should probably replace this whole way of retrieving the control-plane IP and use the grain `metalk8s.control_plane_ip` instead (`metalk8s.node.grains` already handles ignoring any apiserver VIP configured)

alexandre-allard: smart

PR #2105, opened 2019-12-06
Bugfix/scroll volume list and refresh volume information page

ChengYanJin:
**Component**: ui

**Context**: 
As a user, I want to see the update of the volume. ( For the moment, if I go to volumeInformation page right away creating the volume, I couldn't see the status changing)

**Summary**:

- Fix the two bugs:

1.  When there is no data in `clusterMonitoring`, the text `no data available` moving from top to down.
2. The volume list table is not scrollable.

- Add refresh in `volumeInfomation` page.

**Acceptance criteria**: 

- Add the update in volumeInformation page

image

- The volume list table is scrollable.
image

### DIFF
--- a/ui/src/containers/ClusterMonitoring.js
+++ b/ui/src/containers/ClusterMonitoring.js
@@ -40,6 +40,7 @@ const PageContainer = styled.div`
 `;
 
 const TableContainer = styled.div`
+  height: 100%;
   flex-grow: 1;
 
   .sc-table-column-cell-container-severity {
--- a/ui/src/containers/VolumeInformation.js
+++ b/ui/src/containers/VolumeInformation.js
@@ -4,8 +4,12 @@ import { useRouteMatch } from 'react-router';
 import { injectIntl, FormattedDate, FormattedTime } from 'react-intl';
 import styled from 'styled-components';
 import { padding } from '@scality/core-ui/dist/style/theme';
-import { Breadcrumb, Table } from '@scality/core-ui';
-import { makeGetNodeFromUrl, makeGetVolumesFromUrl } from '../services/utils';
+import { Breadcrumb, Table, Loader } from '@scality/core-ui';
+import {
+  makeGetNodeFromUrl,
+  makeGetVolumesFromUrl,
+  useRefreshEffect,
+} from '../services/utils';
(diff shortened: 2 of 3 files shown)


ui/src/containers/VolumeInformation.js
```diff
@@ -60,14 +66,23 @@ const InformationValueSection = styled.div`
   min-width: 500px;
 `;
 
+const LoaderContainer = styled(Loader)`
+  position: absolute;
```
carlito-scality: Avoid `position: absolute;`  if you can please

ui/src/containers/ClusterMonitoring.js
```diff
@@ -40,6 +40,7 @@ const PageContainer = styled.div`
 `;
 
 const TableContainer = styled.div`
+  height: 100%;
```
carlito-scality, line 43: it is relative to volume?

carlito-scality (changes requested):

ui/src/containers/ClusterMonitoring.js
ChengYanJin: Nope. it's not related to the volume.  
I found this bug, it's just one line change, so want to fix it here as well.

ui/src/containers/VolumeInformation.js
ChengYanJin: Ok. I removed it. 
Using `visibility` to control the visibility of the `<Loader>`

ui/src/containers/VolumeInformation.js
```diff
@@ -143,6 +167,8 @@ const VolumeInformation = props => {
         />
       ) : null}
 
+      {<LoaderContainer isLoading={isLoading} size="small" />}
```
carlito-scality, line 170: Need to find a better place for the Loader but fine for now

carlito-scality (approved):

ui/src/containers/VolumeInformation.js
ChengYanJin: ok! I did a lot of thinkings about where is the best place to put loader.  
Yeah, we will discuss later to find a good place.

PR #2106, opened 2019-12-06
HA access to `kube-apiserver` using a host-local proxy

### LINKED ISSUE #2103: Control-plane without LB for API-server is a SPOF
When deploying without creating an external load-balancer for the API server, or enabling the `keepalived` VIP management, we end up configuring all services who require 'external' access to the API server, including `kube-controller-manager`, `kube-scheduler`, `kubelet`, `kube-proxy` to connect to `apiServer.host` from the `BootstrapConfiguration`, which would then be the bootstrap node. As such, this becomes a SPOF for the control-plane, even when multiple control-plane nodes are up and running.

This can be resolved by either using an external LB or use the built-in `keepalived` functionality. However, this may not be feasible in all environments.

As such, we could consider adopting the Kubespray approach of deploying an `nginx` instance on every host in the cluster (using the `hostNetwork`), listening on `127.0.0.1:6443`, which then balances/proxies to all known API server instances in the cluster. At that point, we can configure all `k-c-m`s, `k-s`s, `kubelet`s, `kube-proxy`s and other consumers of the API server to connect to `127.0.0.1:6443`.

This does *not* fix the issue of API server accessibility from *outside* the cluster, i.e. an admin's workstation or similar.

NicolasT:
Implement the 'local `nginx`' approach as described in #2103.

Fixes: #2103


NicolasT:
There's a leftover reference to the `metalk8s:api_server:host` Pillar value in `salt/metalk8s/orchestrate/solutions/available.sls`, however, I'm not sure that's *supposed* to be there.

Need input @gdemonet.

NicolasT:
CI failures because in the upgrade/downgrade scenarios, `apiServer` config is still expected in `BootstrapConfiguration` (of course). Will fix.

salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.yaml.j2
```diff
@@ -0,0 +1,59 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: apiserver-proxy
+  namespace: kube-system
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
```
gdemonet, line 7: Oh, I was surprised by this "addon manager", looks like it is a "Bash Operator" :) https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/kube-addons.sh

buildchain/buildchain/image.py
```diff
@@ -204,7 +207,9 @@ def _operator_image(name: str, **kwargs: Any) -> targets.OperatorImage:
     'nginx-ingress-defaultbackend-amd64': 'defaultbackend-amd64',
 }
 
-SAVE_AS_TAR : FrozenSet[str] = frozenset(('nginx', 'pause'))
+SAVE_AS_TAR : FrozenSet[str] = frozenset(['pause'])
+
+SAVE_AS_TAR_AND_IMAGE : FrozenSet[str] = frozenset(['nginx'])
```
slaperche-scality: I'm not very fond of this approach.

It's really ad-hoc/hackish, you can see it from the name `SAVE_AS_TAR_AND_IMAGE`.
Like if we ever add a third format, we will have `SAVE_AS_TAR_AND_IMAGE_AND_FOO` (unlikely I know, but it's still a code smell :wink: )

This aside, the current implementation breaks the cleanup: a `doit clean` on `_image_pull:nginx` won't delete the tarball :confused: 

I propose getting rid of the Boolean to use a list of formats instead. As a nice side effect it allows to have a simpler code in `RemoteImage` by moving the variant parts into dedicated class.
You can cherry-pick https://github.com/scality/metalk8s/commit/ac15e324be8b9deb186bfa805b0cdb841e6c5dbd if you want (should apply on top of your PR without conflict :slightly_smiling_face: ).

slaperche-scality (dismissed):

NicolasT:
This seems to somehow break downgrade, because `etcd` doesn't get restarted. Further investigation shows this is due to the image not getting pulled?!?
Indeed, there's no `/var/lib/metalk8s/repositories/conf.d/99-metalk8s-2.3.0-dev-registry.inc` or similar on my test system. No clue what's causing that, though...

buildchain/buildchain/image.py
NicolasT: Done

salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
```diff
@@ -0,0 +1,65 @@
...
+
+stream {
+  upstream apiserver {
+    least_conn;
+{%- for apiserver in apiservers %}
+{%-     if apiserver == grains['metalk8s']['control_plane_ip'] %}
+{%-         set weight = 100 %}
+{%-      else %}
+{%-         set weight = 1 %}
+{%-      endif %}
```
gdemonet, line 34: FWIW, you can use [inline conditional expressions](https://jinja.palletsprojects.com/en/2.10.x/templates/#if-expression) with Jinja:

```
{%-     set weight = 100 if apiserver == grains['metalk8s']['control_plane_ip'] else 1 %}
```

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -153,72 +93,14 @@ Create kube-apiserver Pod manifest:
           - path: {{ htpasswd_path }}
             type: File
             name: htpasswd
-{%- if pillar.metalk8s.api_server.keepalived.enabled %}
-          - path: /etc/keepalived
-            name: keepalived-config
-{%- endif %}
         sidecars:
```
TeddyAndrieux: Maybe you should remove this key since it's empty

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -153,72 +93,14 @@ Create kube-apiserver Pod manifest:
...
-            volumeMounts:
-              - name: keepalived-config
-                mountPath: /etc/keepalived
-                readOnly: true
-              - name: keepalived-varrun
-                mountPath: /var/run
-              - name: keepalived-tmp
-                mountPath: /tmp
-{%- endif %}
         extra_volumes:
```
TeddyAndrieux: ditto

salt/metalk8s/kubernetes/apiserver/kubeconfig.sls
```diff
@@ -4,7 +4,7 @@
 include:
   - metalk8s.internal.m2crypto
 
-{%- set apiserver = 'https://' ~ pillar.metalk8s.api_server.host ~ ':6443' %}
+{%- set apiserver = 'https://' ~ grains['metalk8s']['control_plane_ip'] ~ ':6443' %}
```
TeddyAndrieux, line 7: Why do not use the "proxy" here ?
That mean if for what ever reason local apiserver do not work the `admin.conf` will not work and since we use this kubeconfig everywhere it's not really good IMO

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -39,6 +39,15 @@ Wait for pillar refresh to complete:
...
 {%- set other_nodes = pillar.metalk8s.nodes.keys() | difference(cp_nodes) | sort %}
 
+Ensure all apiservers serve a certificate for 127.0.0.1:
+  salt.state:
+    - tgt: {{ cp_nodes | join(",") }}
+    - tgt_type: list
+    - sls:
+      - metalk8s.internal.upgrade.apiserver-cert-localhost
+    - saltenv: {{ saltenv }}
+    - batch: 1
```
TeddyAndrieux, line 49: I think this state should require `module: Wait for pillar refresh to complete`

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -53,9 +62,16 @@ Skip node {{ node }}, already in {{ node_version }} newer than {{ dest_version }
 
   {%- else %}
 
+Install apiserver-proxy on {{ node }}:
+  salt.state:
+    - tgt: {{ node }}
+    - sls:
+      - metalk8s.kubernetes.apiserver-proxy
+    - saltenv: {{ saltenv }}
```
TeddyAndrieux, line 72: This one should require `salt: Ensure all apiservers serve a certificate for 127.0.0.1`

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -53,9 +62,16 @@ Skip node {{ node }}, already in {{ node_version }} newer than {{ dest_version }
...
+
 Wait for API server to be available on {{ node }}:
```
TeddyAndrieux, line 74: This one should require `salt: Install apiserver-proxy on {{ node }}` we don't want to run this one if proxy do not work otherwise we may wait for a timeout for ... nothing

salt/metalk8s/salt/master/configured.sls
```diff
@@ -13,10 +13,7 @@ Configure salt master:
     - template: jinja
     - defaults:
         salt_ip: "{{ salt_ip }}"
-        kubeconfig: "{{ pillar['metalk8s']['api_server']['kubeconfig'] }}"
-        {%- if pillar['metalk8s']['api_server'].get('context') %}
-        kubecontext: "{{ pillar['metalk8s']['api_server']['context'] }}"
-        {%- endif %}
+        kubeconfig: "/etc/kubernetes/admin.conf"
```
TeddyAndrieux, line 16: Like already said on slack for me this change should land in `dev/2.0` to not have to edit `scripts/iso-manager.sh` in this PR

TeddyAndrieux (commented):
Looks good just few comments

salt/metalk8s/kubernetes/apiserver/installed.sls
NicolasT: Done

salt/metalk8s/kubernetes/apiserver/installed.sls
NicolasT: Done

salt/metalk8s/kubernetes/apiserver/kubeconfig.sls
NicolasT: Because this conf is perfectly valid to be used from *outside* the node, e.g. an admin copying it to his machine (aren't we even doing that for tests?)

Note how *most* `KubeConfig`s are rewritten to use the local proxy, so the services using those shouldn't be impacted. AFAIK only `salt-master` is using `admin.conf` (which is a bug by itself...).

In practice, not much changes, because in our current deployments the `apiServer.host` value is set to the hostname or IP of the bootstrap node anyway...

NicolasT: Not that any *script* which uses `admin.conf` and is *certain* it's running on a host on which the proxy is deployed, could pass `-s`/`--server` to the `kubectl` invocation to use the proxy anyway.

salt/metalk8s/salt/master/configured.sls
NicolasT: I'll do a separate PR to handle that. Don't want this one to go in 2.0, and don't want to wait for something to land in 2.0 which is primarily 'cosmetic'.

salt/metalk8s/orchestrate/upgrade/init.sls
NicolasT: I'll do it, but could you explain why?

salt/metalk8s/orchestrate/upgrade/init.sls
NicolasT: No, there's no dependency on that. `kubelet` reconfiguration requires the proxy to be up, and the cert to be resigned. But the proxy can happily be deployed and run, it doesn't really care about the services it's proxying to (at the TCP level).

salt/metalk8s/orchestrate/upgrade/init.sls
NicolasT: Agree,  done.

salt/metalk8s/orchestrate/upgrade/init.sls
NicolasT: I added a `require` on `salt: Ensure all apiservers serve a certificate for 127.0.0.1` to `Deploy node {{ node }}` instead.

salt/metalk8s/kubernetes/apiserver/kubeconfig.sls
TeddyAndrieux: Ok but then I think we need to change something in our current restore script because if I well remember we need to contact apiserver before configuring the local one and we basically call `kubeconfig` salt state from apiserver and then use this new admin,conf

salt/metalk8s/orchestrate/upgrade/init.sls
TeddyAndrieux: Because we want the pillar to be refreshed before calling this, as you use information from the pillar and as just before this highstate we just upgrade the bootstrap our pillar may-be empty

salt/metalk8s/kubernetes/apiserver/kubeconfig.sls
NicolasT: That wouldn't work today in the first place then, unless the APIserver address was HA...

TeddyAndrieux: Indeed 
```
.. warning::

    You cannot use the restore script if you do not have High Availability
    apiserver because some information required to reconfigure the others
    nodes are stored in the apiserver.
```
PS: Oh btw you should remove this note in this PR also

NicolasT: No I mean, that's not a reasonable requirement in the first place.

Instead, the recover script should take, as an argument, the address of *an* API-server, and use that address when connecting to do whatever it needs to do.

TeddyAndrieux: With this PR yes indeed but when restore script get writed in our bootstrap config the goal was to have an IP able to hit apiserver, and not only bound to bootstrap IP, either with keepalived or with any other system because configuring all server with the bootstrap IP as apiserver endpoint is really wrong.
But yes now we need to change this

NicolasT: My point is that a VIP or LB for API-server has *always* been *optional*, so the restore script *relying* on that is wrong in the first place.

The bug fixed in this PR is the fact that sometimes we relied on this given address to be HA, given the context of #2108 

The restore script should also be updated to support this properly, and I'm happy to open a bug for this, but I'd rather not move this into this PR, since the context is completely different IMHO.

Finally, the lack of CI for the restore operation (#1687) basically *allows* me to merge this without remorse ;-P

NicolasT: Ugh, I keep putting comments, but the topic keeps resonating.

What I was thinking, primarily: there *should* be no `admin.conf` in the backup, because it's not at all required to restore a cluster. An `admin.conf` can be generated from the CA data that *does* need to be backed up.

The whole restore process should be constructive, starting from the minimal set of required data as kept in a backup.

NicolasT: So, I went to see, and apparently there's not even an `admin.conf` in the backup archive (which is good!)

However, what *is* in there has a really strange structure... We really need to spend some time on that it seems.

TeddyAndrieux (approved):
LGTM (except restore script which is broken now but will be tackle in another PR)

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Release 2.4.2 (in development)
 
+### Breaking changes
```
gdemonet, line 5: Would probably make sense to bump the version of `BootstrapConfiguration`

gdemonet (changes requested):
Works on my machine, and code looks good to me, except for the `apiVersion` of the `BootstrapConfiguration` which should change (as you documented it, it's a breaking change).

CHANGELOG.md
NicolasT: No. What you can put in the 'object' doesn't really change: there are no backwards-incompatible changes or anything. We just now ignore information we did use before (and as such, no longer require them to be set). It's a required value that now becomes kinda 'optional' (and unused, at all).

Basically, any existing `v1alpha2` `BootstrapConfiguration` *still* works.

Issue #2101, opened 2019-12-06
Downgrade process is wrong


TeddyAndrieux:
**Component**:

'salt', 'downgrade'

**What happened**:

Current downgrade process:
```
- Downgrade workload plane nodes and then control plane nodes (except bootstrap one) <-- (including etcd ....)
- Downgrade etcd cluster <-- (automatically done when downgrading other nodes)
- Downgrade k8s objects
- Downgrade bootstrap stuff (salt-master, repositories, ...)
```
Several issues:
- if we need `ext_pillar` informations when downgrading nodes or etcd that use some stuff only available with old `salt-master` we can't as we downgrade the salt-master at the end
- downgrading of `etcd` it's done when downgrading node and downgrade of etcd basically do nothing

**What was expected**:

Being able to use old `ext_pillar` only available with old `salt-master` when downgrading

**Resolution proposal** (optional):

TBD:
Changes the downgrade process order:
```
- Apply bunch of pre-downgrade k8s states (cleanup no longer used objects, ...)
- Downgrade bootstrap stuff (salt-master, repositories, ...)
- Downgrade workload plane nodes and then control plane nodes (removing `etcd` role from pillar when calling highstate)
- Downgrade etcd cluster
- Downgrade kubernetes objects (NOTE: this one may be moved right after "downgrade of bootstrap stuff", TBD)
```

PR #2110, opened 2019-12-09
Change downgrade process (downgrade bootstrap first)

TeddyAndrieux:
**Component**:

'salt', 'scripts', 'downgrade'

**Context**: 

Downgrade process is wrong, some ext_pillar need a salt-master in the same version to be computed

**Summary**:

- Downgrade bootstrap stuff before calling the downgrade orchestrate
- No longer downgrade etcd at the same time of the node

---

Fixes: #2101

### DIFF
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -1,6 +1,8 @@
 {%- set node_name = pillar.orchestrate.node_name %}
 {%- set version = pillar.metalk8s.nodes[node_name].version %}
 
+{%- set skip_roles = pillar.metalk8s.nodes[node_name].get('skip_roles', []) %}
+
 {%- if node_name not in salt.saltutil.runner('manage.up') %}
 Deploy salt-minion on a new node:
   salt.state:
@@ -116,6 +118,15 @@ Run the highstate:
     - tgt: {{ node_name }}
     - highstate: True
     - saltenv: metalk8s-{{ version }}
+    {#- Add ability to skip node roles to not apply all the highstate
+         e.g.: Skipping etcd when downgrading #}
+    {%- if skip_roles %}
+    - pillar:
+        metalk8s:
+          nodes:
+            {{ node_name }}:
+              skip_roles: {{ skip_roles }}
+    {%- endif %}
     - require:
       - salt: Set grains
       - salt: Refresh the mine
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -39,11 +39,6 @@ Wait for pillar refresh to complete:
 Skip node {{ node }}, already in {{ node_version }} older than {{ dest_version }}:
   test.succeed_without_changes
 
-  {%- elif 'bootstrap' in pillar.metalk8s.nodes[node].roles %}
-
-Skip node {{ node }}, bootstrap node downgrade should be done later:
-  test.succeed_without_changes
-
   {%- else %}
 
 Wait for API server to be available on {{ node }}:
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -10,6 +10,20 @@ Execute the downgrade prechecks:
         orchestrate:
           dest_version: {{ dest_version }}
 
+# In salt 2018.3 we can not do synchronous pillar refresh, so add a sleep
+# see https://github.com/saltstack/salt/issues/20590
+Wait for pillar refresh to complete:
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
+    - require:
+      - salt: Execute the downgrade prechecks
+  module.run:
+    - test.sleep:
+      - length: 20
+    - require:
+      - salt: Wait for pillar refresh to complete
+
 {%- set cp_nodes = salt.metalk8s.minions_by_role('master') | sort %}
 {%- set other_nodes = pillar.metalk8s.nodes.keys() | difference(cp_nodes) | sort %}
 
@@ -39,7 +53,7 @@ Wait for API server to be available on {{ node }}:
   - status: 200
   - verify_ssl: false
   - require:
-    - salt: Execute the downgrade prechecks
+    - module: Wait for pillar refresh to complete
   {%- if previous_node is defined %}
     - salt: Deploy node {{ previous_node }}
   {%- endif %}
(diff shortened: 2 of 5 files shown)


salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -10,6 +10,20 @@ Execute the downgrade prechecks:
         orchestrate:
           dest_version: {{ dest_version }}
 
+# In salt 2018.3 we can not do synchronous pillar refresh, so add a sleep
+# see https://github.com/saltstack/salt/issues/20590
+Refresh pillar everywhere and wait a bit:
```
Ebaneck: Minor: ```Wait for pillar refresh to complete``` ??

Ebaneck (changes requested):
Slight grammar check

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -10,6 +10,20 @@ Execute the downgrade prechecks:
...
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
+    - require:
+      - salt: Execute the downgrade prechecks
+  module.run:
+    - test.sleep:
+      - length: 20
+    - require:
+      - salt: Refresh pillar everywhere and wait a bit
```
Ebaneck: ditto

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -39,7 +48,7 @@ Wait for API server to be available on {{ node }}:
   - status: 200
   - verify_ssl: false
   - require:
-    - salt: Execute the downgrade prechecks
+    - module: Refresh pillar everywhere and wait a bit
```
Ebaneck: ditto

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -22,6 +22,20 @@ Upgrade etcd cluster:
     - require:
       - salt: Execute the upgrade prechecks
 
+# In salt 2018.3 we can not do synchronous pillar refresh, so add a sleep
+# see https://github.com/saltstack/salt/issues/20590
+Refresh pillar everywhere and wait a bit:
```
Ebaneck: ditto

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -22,6 +22,20 @@ Upgrade etcd cluster:
...
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
+    - require:
+      - salt: Upgrade etcd cluster
+  module.run:
+    - test.sleep:
+      - length: 20
+    - require:
+      - salt: Refresh pillar everywhere and wait a bit
```
Ebaneck: ditto

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -46,7 +60,7 @@ Wait for API server to be available on {{ node }}:
   - status: 200
   - verify_ssl: false
   - require:
-    - salt: Upgrade etcd cluster
+    - module: Refresh pillar everywhere and wait a bit
```
Ebaneck: ditto

Ebaneck (dismissed):

salt/metalk8s/orchestrate/downgrade/init.sls
TeddyAndrieux: True it's better, I will change

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -116,6 +116,13 @@ Run the highstate:
     - tgt: {{ node_name }}
     - highstate: True
     - saltenv: metalk8s-{{ version }}
+    - pillar:
+        # Add ability to override node roles to not apply all the highstate
+        # e.g.: Skipping etcd when downgrading
```
gdemonet: It works, sure, but doesn't look super good IMHO... Could we use something dedicated to "skipping" roles? E.g. `pillar.highstate.skip_roles", and let the `top.sls` filter this out?

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -10,6 +10,20 @@ Execute the downgrade prechecks:
...
+# see https://github.com/saltstack/salt/issues/20590
+Wait for pillar refresh to complete:
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: '*'
+    - require:
+      - salt: Execute the downgrade prechecks
+  module.run:
+    - test.sleep:
+      - length: 20
```
gdemonet, line 23: 20 seconds for a pillar refresh?! Isn't that a bit too much?

gdemonet (dismissed):
Minor comments, otherwise looks good to me

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Sure could be better i will

salt/metalk8s/orchestrate/downgrade/init.sls
TeddyAndrieux: Since we do several call to apiserver that may be on another machine I think it's quite ok and it's only one time during all the upgrade/downgrade process so it's ok IMO

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Ok just thinking about it and that mean that we will not support it for current released version as we can't change the top.sls of already released version but I think it's quite ok

salt/metalk8s/orchestrate/downgrade/init.sls
gdemonet: OK. We should do some profiling on our various Salt orchestrations, some time (not urgent at all, but would be interesting for us).

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -1,6 +1,8 @@
 {%- set node_name = pillar.orchestrate.node_name %}
 {%- set version = pillar.metalk8s.nodes[node_name].version %}
 
+{%- set skip_roles = pillar.metalk8s.nodes[node_name].get('skip_roles', []) %}
```
gdemonet, line 4: Maybe source it from `pillar.orchestrate`?

gdemonet (approved):

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Hmm no, it's not really intuitif that mean if someone want to skip etcd from CLI directly when running highstate he will need to change the one in `metalk8s.nodes` and if he want to skip using `deploy_node` he can only use `orchestrate` one.
Even if it's not a feature we should use really often for me it's better to have only one place to put it in the pillar

Issue #2070, opened 2019-11-24
Use a YAML Renderer to create ConfigMap SLS


ChengYanJin:
Both of `salt_embed_text` and  `salt_embed_bytes` could rely on `yaml.dump` (see https://github.com/scality/metalk8s/pull/2042/commits/1ada06c5769e987afc7473cb58250419e413b5bd for an example of how to customize representers). 

_Originally posted by @gdemonet in https://github.com/scality/metalk8s/pull/2062_

slaperche-scality:
Solved with https://github.com/scality/metalk8s/pull/2120

PR #2120, opened 2019-12-10
Use Renderer to generate the ConfigMap

slaperche-scality:
**Component**:

buildchain

**Context**: 

We want to generate SLS from the buildchain, and we have to shove huge string inside it while trying to keep it readable AND have a valid YAML in the end.

**Summary**:

Instead of creating/formatting the SLS content by hand, let's use a YAML renderer.

**Acceptance criteria**: 

You can build a working ISO

---

Closes: #2070


NicolasT (dismissed):
LGTM. One minor concern: the SLS renderer takes a `content` object, whilst we often put multiple objects in a single SLS file as a YAML document stream (objects separated by `---`). This renderer could be more generalized by taking a list of objects as `objects` and render them accordingly.

slaperche-scality:
> LGTM. One minor concern: the SLS renderer takes a `content` object, whilst we often put multiple objects in a single SLS file as a YAML document stream (objects separated by `---`). This renderer could be more generalized by taking a list of objects as `objects` and render them accordingly.

Good point, and shouldn't be too hard to support: I can add a new commit for that.

slaperche-scality:
Done, I've added support for list of objects.

NicolasT (approved):
Not 100% sure I'm 100% comfortable with this approach, since now we have an SLS in the Salt tree that's not represented by any file in the 'source' `salt/` tree, but OK.

buildchain/buildchain/targets/serialize.py
```diff
@@ -95,3 +121,64 @@ def _render(self) -> Callable[[Any, Path], None]:
...
+    dumper.add_representer( # type: ignore
+        YAMLDocument.Literal, _literal_representer
+    )
+    dumper.add_representer( # type: ignore
+        YAMLDocument.ByteString, _bytestring_representer
+    )
+    try:
+        dumper.open() # type: ignore
+        for document in data:
+            dumper.represent(document) # type: ignore
```
NicolasT, line 178: I assume this then adds the `---`?

slaperche-scality: Yes, I've tested locally.

And the code is highly inspired from [official PyYAML code](https://github.com/yaml/pyyaml/blob/master/lib/yaml/__init__.py#L278-L284).

Issue #2064, opened 2019-11-22
Builchain not working with Python 3.8


slaperche-scality:
**Component**:

buildchain

**What happened**:

When I try to build the ISO, some dependencies (`cloudpickle-1.0.0`) fail to install in the virtualenv:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.8/runpy.py", line 142, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/usr/lib/python3.8/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/doit/__init__.py", line 33, in <module>
    from doit.doit_cmd import get_var
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/doit/doit_cmd.py", line 14, in <module>
    from .cmd_run import Run
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/doit/cmd_run.py", line 8, in <module>
    from .runner import Runner, MRunner, MThreadRunner
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/doit/runner.py", line 8, in <module>
    import cloudpickle
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/cloudpickle/__init__.py", line 3, in <module>
    from cloudpickle.cloudpickle import *
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/cloudpickle/cloudpickle.py", line 152, in <module>
    _cell_set_template_code = _make_cell_set_template_code()
  File "/home/sylvain/dev/metalk8s/buildchain/.venv/lib/python3.8/site-packages/cloudpickle/cloudpickle.py", line 133, in _make_cell_set_template_code
    return types.CodeType(
TypeError: an integer is required (got type bytes)
```

Same kind of things happen when I try to lint the code (this time it's related to `typed-ast-1.3.5`:

```
    ast3/Python/ast.c:22:1: error: conflicting types for '_PyObject_FastCall'
     _PyObject_FastCall(PyObject *func, PyObject *const *args, int nargs)
[...]

PR #2126, opened 2019-12-11
Fix requirements on MacOS

slaperche-scality:
**Component**:

buildchain

**Context**: 

When I've updated the requirements.txt for MacOS in 8bc1bc792e13 I've somehow missed websocket-client, which then broke the use of `doit.sh` on MacOS

**Summary**:

Add the missing dependency.

**Acceptance criteria**: 

`./doit.sh` works on MacOS

---

See: #2064


ChengYanJin (approved):
It works for me!!
Thanks for the fix!!!

PR #2133, opened 2019-12-13
Improvement/2121 update the brand to adapt new design

### LINKED ISSUE #2121: Update the color palette to adapt to the new design
**Component**: ui

**Why this is needed**:
We need to update the color palette in order to meet the new design.

**What should be done**:
- Update the color palette (Change the primary color)
- Add more layer
- Handle the hover

TBD

**Test plan**:
TBD

ChengYanJin:
**Component**: ui

**Context**: We want to update our brand theme(light/dark mode) to meet the new design. And will support the theme switch in dev/2.5

**Summary**: 
- update the light/dark theme.json, add the `custom` in order to customise the theme later.
- update the secondary button style and tooltip overlay style
- cherry-pick some bug fix from dev/2.5 ( 3 commits related to volume)
- update the ui-branding configMap

**Acceptance criteria**: 
light mode:
lightmode
lightmode2

dark mode:
darkmode
darkmode2

---

Closes: #2121


ui/public/brand/assets/branding-light.svg
```diff
@@ -0,0 +1,50 @@
+<svg
```
carlito-scality, line 1: - I think we do not need to add a new `brand[light/dark].svg` because the difference is small. Only the text color is changed. We can just add `class="theme"` to where should be themed in the current `brand.svg` and add CSS to handle it. Thus, we don't need the `logoPath`
- for the "custom" theme, we definitely need the FULL `logoPath`, since the `logo` could be mounted somewhere else than `process.env.PUBLIC_URL`. For now, we don't need to support this yet. @gdemonet any thoughts?

ui/src/ducks/config.js
```diff
@@ -68,11 +75,23 @@ export function updateLanguageAction(language) {
...
+
 // Sagas
 export function* fetchTheme() {
   const result = yield call(Api.fetchTheme);
   if (!result.error) {
-    result.brand = mergeTheme(result, defaultTheme);
+    // get the default theme from configMap
+    const defaultThemeMode = result.default;
+    yield put(setThemesAction(result.theme));
+    yield put(setLogoPathAction(result.theme[defaultThemeMode].logo_path));
```
carlito-scality: no need to set logoPath since it is already in  `theme `

carlito-scality (changes requested):

ui/src/containers/CreateVolume.js
```diff
@@ -384,6 +384,7 @@ const CreateVolume = props => {
                           type="button"
                           onClick={addLabel}
                           data-cy="add-volume-labels-button"
+                          outlined
```
carlito-scality, line 387: why `outlined`? because for me `outlined` button means `secondary` action and in this case `add volume` is a `primary` action.

ChengYanJin: I just want to meet Use Design's proposal ;) The button from them is like this
image

ui/public/brand/assets/branding-light.svg
gdemonet: > I think we do not need to add a new `brand[light/dark].svg` because the difference is small. [...] Thus, we don't need the `logoPath`

True, we can add styling directly to the SVG. We could also handle it in JS, FWIW. In the meantime, having a single path to set a logo, be it a default Scality-provided theme or a custom one, seems a reasonable first iteration. I think we can stick to 2 SVGs for now and open an improvement ticket to handle it some time later.

> for the "custom" theme, we definitely need the FULL `logoPath`, since the `logo` could be mounted somewhere else than `process.env.PUBLIC_URL`. For now, we don't need to support this yet. @gdemonet any thoughts?

The only good reason for the logo not being exposed under the same URL as the UI itself is to use an absolute URL (say, from a CDN). IMHO, it's not urgent.
For now, changing a logo would mean:
- adding it as binary data to the `metalk8s-ui/ui-branding` _ConfigMap_
- referencing it from the root of the application (which is what `PUBLIC_URL` represents) in the `theme.json` `"logoPath"`, e.g. `"/brand/custom_logo.svg"`

I'm however now doubting about the _ConfigMap_ approach. It's sufficient for simple theming through `theme.json` values, but as soon as we start adding multiple files, it becomes cumbersome to say the least. As an example, adding sub-directories under `brand` from this single _ConfigMap_ wouldn't be feasible.
Also, the whole idea of sharing this theming configuration between UIs in the cluster means we need a more flexible approach than this single file.

Some solutions come to mind:
- have _another_ static container for branding assets, and let all UIs expose it with their own _Ingress_
    - (+) platform operators can simply edit files on the Bootstrap (including `theme.json`) and the changes are exposed through all the UIs
    - (-) need to backup, propagate to other Bootstrap nodes (maybe even Infra? along with the UI containers...)
- generate the _ConfigMap_**(s)** from static configuration and files using Salt:
    - (+) users still have a single entrypoint
    - (+) this could be used by Solutions, if they want to share theming assets, e.g. for their NavBar icon
    - (-) all UI containers need to be updated when we add a new _ConfigMap_

... :thinking: What if we try both approaches at once? Have a static `nginx`, with its Ingress for e.g. `/brand`, that exposes files from two _ConfigMap_ sources: system defaults and user configuration. As in the current UI containers, system defaults are used as fallback when user-provided values are not available. However, instead of having those stored at build-time, we will need to mount them (this is necessary to add new system defaults dynamically, e.g. for Solution UIs). User-provided configuration can be generated from Salt / Metal UI / whatever.

If you need to upgrade/downgrade Metal / a Solution, change the system defaults _ConfigMap_(s). The user-provided (or generated) _ConfigMap_(s) will be persisted.

ui/src/containers/CreateVolume.js
carlito-scality: It should look like this IMOimage

ui/public/brand/theme.json
```diff
@@ -1,17 +1,43 @@
...
-    "info": "#00B2A9",
-    "warning": "#F1B434",
-    "danger": "#EF3340",
-    "background": "#26232A",
-    "backgroundContrast1": "#2E2B32",
-    "backgroundContrast2": "#353239",
-    "text": "#ffffff",
-    "border": "#ffffff"
-  }
+  "theme":{
```
carlito-scality, line 2: You should to update the config.json configMap as well

carlito-scality (changes requested):
you need to update the

ui/src/containers/CreateVolume.js
ChengYanJin: This is `add label` though, but I changed other buttons which should be the primary.

carlito-scality (approved):
LGTM

Issue #2132, opened 2019-12-13
Add label in every k8s metalk8s object deployed 


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Why this is needed**:

Currently we cannot know, for each and every object we deploy in metalk8s, the metalk8s version deployed. This version could really help to do proper upgrade and downgrade.

**What should be done**:

When deploying an object in metalk8s, add a label about the metalk8s version used to deploy this object `metalk8s.scality.com/version`

**Implementation proposal** (strongly recommended):

TBD:
- Adding autmoatically in our custom module/state/utils
- Added manually in each and every object

gdemonet:
:+1: for the automatic annotation added by Salt. Maybe use something like `metalk8s.scality.com/deployed-with-version`, or `.../saltenv`, to be super explicit.

PR #2135, opened 2019-12-13
salt: Add a version label on each Kubernetes objects

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Context**: 

For upgrade and downgrade we want to know from which version we come so
we need to have for each and every kubernetes objects we deploy the
current version deployed.

**Summary**:

 This commit add a label `metalk8s.scality.com/version` for deployed object
NOTE: this label only affect the direct object we deploy and not all the
child (e.g.: For a deployment only the `Deployment` will have this label
and not all the child `Pod`)

---

Fixes: #2132

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -10,6 +10,7 @@ module, while other methods can be found in `metalk8s_kubernetes_utils.py`,
 """
 
 import logging
+import re
 
 from salt.exceptions import CommandExecutionError
 from salt.utils import yaml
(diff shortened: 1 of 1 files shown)


salt/_modules/metalk8s_kubernetes.py
```diff
@@ -132,6 +133,13 @@ def method(manifest=None, name=None, kind=None, apiVersion=None,
                 )
             )
 
+        # Adding label containing metalk8s version (retrieved from saltenv)
+        match = re.search(r'^metalk8s-(?P<version>.+)$', saltenv)
+        if action in ['create', 'replace'] and match:
```
NicolasT: Would it make sense to set the label value to `unknown` or something along those lines if there's *no* 'match'?

TeddyAndrieux: Yes True, I will

Ebaneck (approved):

PR #2140, opened 2019-12-16
ci: don't re-run bootstrap in multi-node tests

NicolasT:
Re-running bootstrap should not affect anything but the bootstrap node.
Since we run it in single-node CI tests, there's not much use re-running
it in a multi-node environment, especially since we don't actually
*validate* there's no impact on the cluster.

If we were to extend the test-suite to validate this, or when putting
less focus on the single-node environments, we could enable this again.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -251,7 +251,7 @@ models:
         export SSH_CONFIG_FILE=\"${SSH_CONFIG_FILE}\" &&
         export ISO_MOUNTPOINT=\"${ISO_MOUNTPOINT}\" &&
         export TEST_HOSTS_LIST=\"${TEST_HOSTS_LIST}\" &&
-        tox -e tests -- -m \"${PYTEST_FILTERS}\""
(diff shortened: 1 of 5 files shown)


alexandre-allard (dismissed):

eve/main.yml
```diff
@@ -251,7 +251,7 @@ models:
         export SSH_CONFIG_FILE=\"${SSH_CONFIG_FILE}\" &&
         export ISO_MOUNTPOINT=\"${ISO_MOUNTPOINT}\" &&
         export TEST_HOSTS_LIST=\"${TEST_HOSTS_LIST}\" &&
-        tox -e tests -- -m \"${PYTEST_FILTERS}\""
+        tox -e tests -- ${PYTEST_ARGS:-""} -m \"${PYTEST_FILTERS}\""
```
alexandre-allard, line 254: Double quotes are useless here.

NicolasT: I like to be explicit...

alexandre-allard (approved):

gdemonet (approved):

Issue #2122, opened 2019-12-11
UI: intl.translate does not work anymore in React component


carlito-scality:
**Component**: UI

**What happened**:
- Once IntlGlobalProvider is introduced to allow us to translate text in sagas, `intl.translate` does not work anymore in React components, only `intl.messages` works. It triggers an JS error 'translate is not a function'.

**What was expected**:

**Steps to reproduce**

**Resolution proposal** (optional):

PR #2144, opened 2019-12-16
UI - Fix intl.translate in react components

carlito-scality:
**Component**:UI

**Context**: 

- Once IntlGlobalProvider is introduced to allow us to translate text in sagas, intl.translate does not work anymore in React components, only intl.messages works. It triggers an JS error 'translate is not a function'. We need to use ONLY IntlGlobalProvider's intl.translate everywhere (saga and react components) for translation to be consistent. Thus we can get rid of HOC injectIntl

**Summary**:

**Acceptance criteria**: 

---

Closes: https://github.com/scality/metalk8s/issues/2122

### DIFF
--- a/ui/src/translations/IntlGlobalProvider.js
+++ b/ui/src/translations/IntlGlobalProvider.js
@@ -36,6 +36,9 @@ class IntlTranslator {
   // Formatting Functions
   // ------------------------------------
   translate(key, values) {
+    if (!key) {
+      return '';
+    }
     const description = defineMessages({
       message: {
         id: key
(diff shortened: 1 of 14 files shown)


ui/src/translations/IntlGlobalProvider.js
```diff
@@ -36,10 +36,13 @@ class IntlTranslator {
   // Formatting Functions
   // ------------------------------------
   translate(key, values) {
+    if (!key) {
+      return '';
+    }
```
slaperche-scality, line 41: this is the only relevant change, the rest of the diff is linting/cosmetics and belong to a dedicated commit

ChengYanJin (approved):
LGTM!

PR #2145, opened 2019-12-17
salt,build: Add version label on static manifest deployed

TeddyAndrieux:
**Component**:

'salt', 'build'

**Context**: 

Add version in static manifest
Use `dest_version` on `deploy_node` in downgrade orchestrate

**Summary**:

For upgrade and downgrade we want to know from which version we come.
See dde1aa163c943fc68e0b6118e9a7156712ddee8f for more detail

### DIFF
--- a/salt/_states/metalk8s.py
+++ b/salt/_states/metalk8s.py
@@ -1,6 +1,7 @@
 """Custom states for MetalK8s."""
 
 import time
+import re
 
 
 __virtualname__ = "metalk8s"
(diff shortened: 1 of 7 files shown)


NicolasT (commented):
Instead of using build-time templating: the manifest YAMLs are (or could be) Jinja templates, and rendered through `file.managed`, splicing in whatever version string we want at that time...

In general, I think we should use Salt based rendering instead of build-time 'templating' where possible. The rationale: mainly consistency.

Makes sense?

TeddyAndrieux:
> Instead of using build-time templating: the manifest YAMLs are (or could be) Jinja templates, and rendered through `file.managed`, splicing in whatever version string we want at that time...
> 
> In general, I think we should use Salt based rendering instead of build-time 'templating' where possible. The rationale: mainly consistency.
> 
> Makes sense?

Hmm a user can put what ever he want in saltenv but since we use saltenv for k8s object it make sense to use saltenv here also. I will change this

gdemonet (approved):
LGTM

PR #2147, opened 2019-12-18
WIP: Split deploy_node and prepare etcd members for cluster addition

### LINKED ISSUE #2028: Control plane Expansion may leads to broken etcd cluster
**Component**:

'etcd', 'salt'

**What happened**:

While doing a control plane expansion (on vagrant environment), the system is trying to register the new node in etcd cluster even if etcd is not running in the new node. As a result, etcd cannot start and the cluster becomes unusable.

**What was expected**:

If for any reason, we are not able to start etcd on the new added control plane node, we should immediately abort and more importantly not try to register it to the existing etcd cluster.

**Steps to reproduce**

on Vagrant Environment
Deploy bootstrap
Add a node with infra / control plane / worker plane roles
Click on Deploy node and follow the progress
It ends up with the following error:
```
Error: Register the node into etcd cluster - Runner function 'state.orchestrate' failed.`
```
and etcd on bootstrap is not able to start anymore:
```
[root@bootstrap containers]# crictl ps -a | grep etcd
cb9f42e21e7a4       2c4adeb21b4ff       5 minutes ago       Exited              etcd                            13                  5eaa0b27d9c3e
```
etcd on newly added node was never scheduled (no logs available)

Restarting **kubelet** on both nodes is fixing the issue.

Ebaneck:
**Component**:

'salt'

**Context**: 

See #2028 

**Summary**:

- Split deploy_node into generic stages
- Add prepared state to prepare etcd deps and register etcd node into etcd cluster
- Add running state to create manifest and wait for member to be healthy

**Acceptance criteria**: 

---

Closes: #2028


salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -80,3 +80,24 @@ Create local etcd Pod manifest:
...
+    - name: https://127.0.0.1:2379/health
+    - verify_ssl: True
+    - ca_bundle: /etc/kubernetes/pki/etcd/ca.crt
+    - cert:
+      - /etc/kubernetes/pki/etcd/server.crt
+      - /etc/kubernetes/pki/etcd/server.key
+    - status: 200
+    - match: '{"health":"true"}'
+    - require:
+      - module: Delay after etcd pod is deployed
```
TeddyAndrieux: If you do this here you can basically remove the `metalk8s.kubernetes.etcd.healthy` sls (and change it in `metalk8s.orchestrate.etcd`)

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +113,21 @@ Wait minion available:
...
+
+Prepare etcd dependencies:
+  salt.state:
+    - tgt: {{ node_name }}
+    - sls: metalk8s.roles.etcd.prepared
+    - pillar: {{ pillar | json }}
+    - require:
+      - salt: Set grains
+      - salt: Refresh the mine
+      - salt: Refresh and check pillar before highstate
```
TeddyAndrieux: Add a `require_in` here on `Run the highstate`

salt/metalk8s/roles/etcd/prepared.sls
```diff
@@ -0,0 +1,7 @@
+# add member to cluster before etcd starts
+Register the node into etcd cluster:
+  salt.runner:
+    - name: state.orchestrate
+    - pillar: {{ pillar | json  }}
+    - mods:
+      - metalk8s.orchestrate.register_etcd
```
TeddyAndrieux: Maybe an `include` of this state is enough here.

Also it's weird to me to have a sls in `roles` directory that need to run on the salt-master (cc @NicolasT )

salt/metalk8s/roles/etcd/running.sls
```diff
@@ -0,0 +1,3 @@
+# creates the manifest YAML and wait for the member to be up and running
+include:
+  - metalk8s.kubernetes.etcd
```
TeddyAndrieux: This state really needed ?

For me it's not needed and we shouldn't create a state like that except if we want somewhere to run only this state and not all the `roles.etcd` (which is not the case for the moment)

TeddyAndrieux (changes requested):

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +113,21 @@ Wait minion available:
 
 {%- endif %}
 
+# Prepare etcd and dependencies
+{%- if 'etcd' in pillar.get('metalk8s', {}).get('nodes', {}).get(node_name, {}).get('roles', []) %}
+
+Prepare etcd dependencies:
+  salt.state:
+    - tgt: {{ node_name }}
+    - sls: metalk8s.roles.etcd.prepared
```
TeddyAndrieux: You can't run this state on the minion as this is something that need `etcd3` python lib which is only available (by default) on the salt-master

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +113,23 @@ Wait minion available:
...
 {%- endif %}
 
+# Prepare etcd and dependencies
+{%- if 'etcd' in pillar.get('metalk8s', {}).get('nodes', {}).get(node_name, {}).get('roles', []) %}
+
+Prepare etcd dependencies:
+  salt.runner:
+    - tgt: {{ node_name }}
+    - sls: metalk8s.kubernetes.etcd.prepared
+    - pillar: {{ pillar | json }}
```
TeddyAndrieux: That cannot work ! 
The `metalk8s.orchestrate.register_etcd` need to run on the salt-master not on the minion so you cannot use `salt.runner` targetting a node.

IMO you should call the register here and remove all `prepared` stuff as the prepare is only registering and cannot run on the salt-minion

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +113,35 @@ Wait minion available:
...
+    - pillar: {{ pillar | json  }}
+    - mods:
+      - metalk8s.orchestrate.register_etcd
+    - require:
+      - salt: Set grains
+      - salt: Refresh the mine
+      - salt: Refresh and check pillar before highstate
+      - salt: Prepare etcd dependencies
+    - require_in:
+      - salt: Run the highstate
```
gdemonet: Maybe add `metalk8s.kubernetes.etcd` between registering and highstate, to make sure `etcd` is unavailable the least amount of time possible. WDYT @NicolasT @TeddyAndrieux ?

gdemonet (commented):
Other than a suggestion to deploy `etcd` manifest before running highstate, looks good to me!

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: I think/hope that etcd cluster work when a node get registered but not joined. So IMO no.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +113,35 @@ Wait minion available:
 
 {%- endif %}
 
+{%- if 'etcd' in pillar.get('metalk8s', {}).get('nodes', {}).get(node_name, {}).get('roles', []) %}
```
TeddyAndrieux: Please skip this part if `etcd` is in `skip_roles`

salt/metalk8s/roles/etcd/prepared.sls
```diff
@@ -0,0 +1,4 @@
+include:
+  - metalk8s.roles.internal.node-without-calico
+  - metalk8s.container-engine.containerd
```
TeddyAndrieux: Hmmm it's already part of `metalk8s.roles.internal.node-without-calico` so not needed

salt/metalk8s/kubernetes/etcd/prepared.sls
```diff
@@ -0,0 +1,9 @@
+{%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+
+{%- set image = build_image_name("etcd") -%}
+
+Pre-pull the etcd image:
+  containerd.image_managed:
+    - name: {{ image }}
+    - require:
+      - metalk8s_package_manager: Install and configure cri-tools
```
TeddyAndrieux: Please no ! Remove this require

gdemonet:
Another thing, the commit message explains well *what* has been done (splitting `deploy_node` and handling `etcd` first), but it doesn't explain *why*, which is crucial in this case.

First, explain the issue at hand: expanding the `etcd` cluster from 1 to 2 necessarily puts the cluster in a blocked state during the time the new member is declared and not yet started. If this time is too long, liveness probes on the initial `etcd` Pod will fail, it will be restarted, entering a back-off loop. Even after the new `etcd` Pod has been created, this one won't be able to start, since the initial is in back-off, and will enter a back-off as well.

Then, explain that this re-ordering, inspired by what kubespray does for pre-pulling the `etcd` image on the target Node, ensures the shortest time possible between the new member being registered and its corresponding Pod being created.

salt/metalk8s/kubernetes/etcd/prepared.sls
```diff
@@ -0,0 +1,9 @@
+{%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+
+{%- set image = build_image_name("etcd") -%}
```
TeddyAndrieux: Do we really need to set this jinja var ? Putting directly `build_image_name("etcd")` in `name` of `containerd.image_managed` is simpler IMO

salt/metalk8s/orchestrate/deploy_node.sls
gdemonet: > etcd cluster work when a node get registered but not joined

No it does not

salt/metalk8s/roles/etcd/prepared.sls
```diff
@@ -0,0 +1,3 @@
+include:
```
NicolasT: This is not a 'role', not sure it belongs here.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +114,35 @@ Wait minion available:
...
 {%- endif %}
 
+{%- if 'etcd' in roles and 'etcd' not in skip_roles %}
+
+# Prepare etcd and dependencies
+Prepare etcd dependencies:
+  salt.state:
+    - tgt: {{ node_name }}
+    - saltenv: metalk8s-{{ version }}
+    - sls: metalk8s.roles.etcd.prepared
```
NicolasT: See other comment.

salt/metalk8s/orchestrate/deploy_node.sls
NicolasT: When going from 1 to 2 nodes indeed it does not, so we want to make the "time" during which 2 nodes are registered but only 0 or 1 of them are running as short as possible, i.e., apply an as-specific-state as possible.

NicolasT (dismissed):

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Then I suggest we use the `skip_roles` so, we first run highstate without `etcd` and then we do etcd steps in the right order

salt/metalk8s/orchestrate/deploy_node.sls
Ebaneck: Done

salt/metalk8s/roles/etcd/prepared.sls
Ebaneck: Done

salt/metalk8s/roles/etcd/init.sls
```diff
@@ -1,3 +1,3 @@
 include:
-  - metalk8s.roles.internal.node-without-calico
-  - metalk8s.kubernetes.etcd
+  - metalk8s.kubernetes.etcd.configured
```
TeddyAndrieux: This state does not exist and IMO you should not change the roles state but only include the `prepared` in the `metalk8s.kubernetes.etcd.init`or directly in `metalk8s.kubernetes.etcd.installed`

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -1,7 +1,12 @@
 {%- set node_name = pillar.orchestrate.node_name %}
 {%- set version = pillar.metalk8s.nodes[node_name].version %}
 
-{%- set skip_roles = pillar.metalk8s.nodes[node_name].get('skip_roles', []) %}
+{%- set roles_to_skip = pillar.metalk8s.nodes[node_name].get('skip_roles', []) %}
+
+{#- Add etcd to skip_roles in order not to apply all the highstate #}
+{%- set skip_roles = roles_to_skip.append('etcd') %}
```
TeddyAndrieux: `append` do not return the list it's just alter the list so you can't do this as `skip_roles` will always be empty

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +115,42 @@ Wait minion available:
...
+
+Prepare etcd dependencies:
+  salt.state:
+    - tgt: {{ node_name }}
+    - saltenv: metalk8s-{{ version }}
+    - sls:
+      - metalk8s.roles.internal.node-without-calico
+      - metalk8s.kubernetes.etcd.prepared
+    - require_in:
+      - salt: Run the highstate
```
TeddyAndrieux: Remove this one only keep the `require_in` of the last one

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +115,42 @@ Wait minion available:
...
+  salt.runner:
+    - name: state.orchestrate
+    - pillar: {{ pillar | json  }}
+    - mods:
+      - metalk8s.orchestrate.register_etcd
+    - require:
+      - salt: Prepare etcd dependencies
+    - require_in:
+      - salt: Run the highstate
+      - salt: Install and wait for healthy etcd member
```
TeddyAndrieux: ditto, and you don't need to `require_in` on `Install and wait for healthy etcd member` since you already `require` it in next state

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -113,6 +115,47 @@ Wait minion available:
...
+      - salt: Prepare etcd dependencies
+
+Install and wait for healthy etcd member:
+  salt.state:
+    - tgt: {{ node_name }}
+    - saltenv: metalk8s-{{ version }}
+    - timeout: 120
+    - sls: metalk8s.kubernetes.etcd.installed
+    - require:
+      - salt: Register the node into etcd cluster
```
TeddyAndrieux: Since register break the etcd cluster from 1 to 2 IMO no point to register before installing etcd.
I suggest we register just after `etcd.installed` then we shouldn't get the error we have before as `metalk8s.kubernetes.etcd.installed` will basically only render one file

TeddyAndrieux (approved):

Issue #650, opened 2019-02-26
Support salt 3000.2


TeddyAndrieux:
Currently salt 2019.2.0 is not supported as we use yaml information directly. 
So unicode instead of string

https://docs.saltstack.com/en/latest/topics/releases/2019.2.0.html#non-backward-compatible-change-to-yaml-renderer

```
bootstrap: [ERROR   ] Command '[u'systemd-run', u'--scope', u'yum', u'-y', u'install', u"u'util-linux'", u"u'iptables'", u"u'iproute'"]' failed with return code: 1
    bootstrap: [ERROR   ] stdout: Running scope as unit run-396.scope.
    bootstrap: Loaded plugins: fastestmirror
    bootstrap: Loading mirror speeds from cached hostfile
    bootstrap:  * base: distrib-coffee.ipsl.jussieu.fr
    bootstrap:  * epel: mirror.ibcp.fr
    bootstrap:  * extras: centos.quelquesmots.fr
    bootstrap:  * updates: centos.mirrors.proxad.net
    bootstrap: No package u'util-linux' available.
    bootstrap: No package u'iptables' available.
    bootstrap: No package u'iproute' available.
    bootstrap: Error: Nothing to do
    bootstrap: [ERROR   ] retcode: 1
    bootstrap: [ERROR   ] Error occurred installing package(s). Additional info follows:
    bootstrap: 
    bootstrap: errors:
    bootstrap:     - Running scope as unit run-396.scope.
    bootstrap:       Loaded plugins: fastestmirror
    bootstrap:       Loading mirror speeds from cached hostfile
    bootstrap:        * base: distrib-coffee.ipsl.jussieu.fr
    bootstrap:        * epel: mirror.ibcp.fr
    bootstrap:        * extras: centos.quelquesmots.fr
    bootstrap:        * updates: centos.mirrors.proxad.net
    bootstrap:       No package u'util-linux' available.
    bootstrap:       No package u'iptables' available.
    bootstrap:       No package u'iproute' available.
    bootstrap:       Error: Nothing to do
    bootstrap: [ERROR   ] Command '[u'systemd-run', u'--scope', u'yum', u'-y', u'install', u"u'ethtool'", u"u'socat'", u"u'ebtables'", u"u'coreutils'", u"u'iproute'"]' failed with return code: 1
    bootstrap: [ERROR   ] stdout: Running scope as unit run-470.scope.
    bootstrap: Loaded plugins: fastestmirror
[...]

PR #2151, opened 2019-12-18
salt: use `tojson` when splicing in complex structures

NicolasT:
Sometimes we use a complex structure as an interpolated value in an SLS
file. Whilst this works, this is merely by accident, and Salt 2019.2
changes the behaviour: complex values would be spliced in using their
`unicode` representation, including `u""` markers.

As such, use a proper `tojson` filter in the Jinja templates instead.

See: #650
See: https://github.com/scality/metalk8s/issues/650
See: https://docs.saltstack.com/en/latest/topics/releases/2019.2.0.html#non-backward-compatible-change-to-yaml-renderer


TeddyAndrieux (approved):
LGTM

Issue #2148, opened 2019-12-18
Dex logout doesn't work when the language is set to French


ChengYanJin:
**Component**: ui, authentication

**What happened**: When the French language is set, the logout function doesn't work.

**What was expected**: Redirect to the dex login page

**Steps to reproduce** Login the Metalk8s => change the language to french  => LOGOUT

**Resolution proposal** (optional):

The way we do logout maybe is not enough
```
export function* logout() {
  const userManager = yield select(state => state.config.userManager);
  if (userManager) {
    userManager.removeUser(); // removes the user data from sessionStorage
  }
}
```
Maybe we can get some inspiration from the `USER_SIGNED_OUT` action
https://github.com/maxmantz/redux-oidc/blob/master/docs/Actions.md

gdemonet:
Closed by #2153

PR #2153, opened 2019-12-18
UI: Fix logout problem when changing languages

carlito-scality:
**Component**:UI

**Context**: 
- When the French language is set, the logout function doesn't work.

**Summary**:
- When the French language is set in localStorage and the users log in, they have to log in twice because the `fetchConfigAction` is executed twice due to the language state changes. Thus, 2 "user" sessions in the localStorage. When log out, only one user is clear from localStorage. That's why the logout does not work.

**Acceptance criteria**: 

---

Closes: https://github.com/scality/metalk8s/issues/2148


ChengYanJin (approved):
LGTM!!!

PR #2154, opened 2019-12-19
1595 include docs in ui

### LINKED ISSUE #1595: Include documentation in the UI
It'd be valuable to have direct access to the MetalK8s documentation from within the UI (a link from one of the menus on the top, I guess, later on maybe to particular sections of the docs from within particular views/forms/... in the UI).

There are several ways to achieve this:

## Include a HTML rendering of the docs in the UI container and have this served by its `nginx`
The difficulty here lies in the fact that the docs sources are outside of the Docker build context of the UI container.

Some possible approaches:
- Change the build context of the UI container to `/`, and render the HTML as part of the UI `Dockerfile` execution (in a separate intermediate container), then copy things over
- Change the way we handle UI builds: instead of building it in Docker, built an SRPM/RPM, do the same for the (rendered) docs, and build the final UI container by installing those RPMs (next to `nginx` and whatnot)
- Build the docs as a separate container which only includes `busybox` and the rendered HTML, then, in the `metalk8s-ui` *Pod*, introduce an `emptyDir` volume, an `initContainer` which copies the HTML into this volume, and let the UI container's `nginx` serve the docs from this volume.

## Consider the docs/HTML to be a separate application
In this case, we can build the docs in some `Dockerfile`, `FROM nginx`, and serve the docs as-is. We can then either run `metalk8s-docs` as a separate *Pod* and let `/docs` (or whatever) be served by it in the `metalk8s-ui` `Ingress`, or run this *docs* container as a second container in the `metalk8s-ui` *Pod*, on a different port (bound to `127.0.0.1` only), and configure the main `metalk8s-ui` `nginx` to proxy `/docs` (or whatever) requests to this one.

There may be other approaches... Having the docs as static files in the the `metalk8s-ui` container somehow seems like the 'best' approach to me (no need to run one more `nginx` or other nonsense, and those HTMLs belong with the UI container if the UI actually links to them in the first place...), however not sure how to achieve this technically (at least with Docker. `buildah` would make this trivial).

Currently putting in milestone [2.4.0-beta1](https://github.com/scality/metalk8s/milestone/15) because it'd be really cool to have this from the start, given the UX evaluation feedback.

alexandre-allard:
**Component**: buildchain, doc, ui

**Context**: we want to embed the documentation inside the metalk8s-ui container image

**Summary**: split build of the UI NodeJS code and container image, then embed doc inside the container image.

**Acceptance criteria**: Nothing is broken and you can access the documentation through the UI ip:port/docs

---

Closes: #1595


buildchain/buildchain/builder.py
```diff
@@ -67,6 +67,9 @@ def _builder_image(
     build_context=constants.ROOT,
     file_dep=[
         constants.ROOT/'tox.ini', constants.ROOT/'docs/requirements.txt'
+    ],
```
slaperche-scality: Good catch!

But this applies to all the builder images (we always touch a file under `_build`) so I would do the fix in the `_builder_image` helper (that way, it will applies to every image).

```diff
@@ -39,6 +39,7 @@ def _builder_image(
         dockerfile=dockerfile,
         destination=config.BUILD_ROOT,
         save_on_disk=False,
+        task_dep=['_build_root']
         **kwargs
     )
```

buildchain/buildchain/builder.py
```diff
@@ -82,9 +82,24 @@ def _builder_image(
...
 
+UI_BUILDER : LocalImage = _builder_image(
+    name='ui',
+    dockerfile=constants.ROOT/'ui'/'Dockerfile',
+    build_args={
+        'NODE_IMAGE_VERSION': versions.NODEJS_IMAGE_VERSION,
+    },
+    file_dep=[
+        constants.ROOT/'ui'/'entrypoint.sh',
+    ]
```
slaperche-scality: minor: can be squashed

```python
    build_args={'NODE_IMAGE_VERSION': versions.NODEJS_IMAGE_VERSION},
    file_dep=[constants.ROOT/'ui'/'entrypoint.sh']
```

buildchain/buildchain/ui.py
```diff
@@ -0,0 +1,73 @@
+# coding: utf-8
+
+"""Tasks to build the MetalK8s UI.
+"""
```
slaperche-scality: can be squashed: `"""Tasks to build the MetalK8s UI."""`

or you could a a little schema like:

```
┌────┐    ┌───────┐    ┌───────┐
│ ui │───>│ mkdir │───>│ build │
└────┘    └───────┘    └───────┘
```
as we do for other module.

buildchain/buildchain/ui.py
```diff
@@ -0,0 +1,73 @@
...
+
+def task__ui_mkdir_build_root() -> types.TaskDict:
+    """Create the MetalK8s UI build root directory."""
+    return targets.Mkdir(
+        directory=constants.UI_BUILD_ROOT, task_dep=['_build_root']
+    ).task
+
+
+def task__ui_build() -> types.TaskDict:
+    """Build the MetalK8s UI NodeJS"""
```
slaperche-scality: minor: Missing period at the end.

ui/Dockerfile
```diff
@@ -1,27 +1,13 @@
...
-COPY public /home/node/public
+RUN mkdir -p ~/ui
 
-COPY package*.json /home/node/
-RUN npm ci
+WORKDIR /home/node/ui
 
-COPY src /home/node/src
-RUN npm run build
+COPY entrypoint.sh /entrypoint.sh
```
slaperche-scality, line 13: Why do you need to copy the entrypoint? You're mounting it when you use it to build the UI.

buildchain/buildchain/image.py
```diff
@@ -234,20 +234,18 @@ def _operator_image(name: str, **kwargs: Any) -> targets.OperatorImage:
     ),
     _local_image(
         name='metalk8s-ui',
-        dockerfile=constants.ROOT/'ui'/'Dockerfile',
+        build_context=config.BUILD_ROOT,
```
slaperche-scality, line 226: commit message: `s/We know build/We now build/`

images/metalk8s-ui/conf/nginx.conf
```diff
@@ -0,0 +1,44 @@
+server {
```
slaperche-scality: This is a duplicate of `ui/conf/nginx.conf`, no?
Do we need to keep these two copies,

buildchain/buildchain/ui.py
```diff
@@ -70,4 +73,22 @@ def clean() -> None:
...
+    def clean(target: Path) -> Callable[[], None]:
+        return lambda: coreutils.rm_rf(target)
+
+    return {
+        'title': utils.title_with_target1('COPY'),
+        'actions': [(coreutils.cp_file, (source, target))],
+        'targets': [target],
+        'task_dep': ['_build_root'],
+        'file_dep': [source],
+        'clean': [clean(target)],
```
slaperche-scality: want to delete `target`, then you can simply use `'clean': True`.

> A common scenario is a task that needs to “revert” its actions. A task may include a clean attribute. This attribute can be True to remove all of its target files.

See https://pydoit.org/cmd_other.html#clean

You only need a custom callable when you want to delete stuff that aren't in `targets`.

buildchain/buildchain/ui.py
```diff
@@ -36,7 +39,7 @@ def task__ui_mkdir_build_root() -> types.TaskDict:
 
 
 def task__ui_build() -> types.TaskDict:
-    """Build the MetalK8s UI NodeJS"""
+    """Build the MetalK8s UI NodeJS code"""
```
slaperche-scality: minor: missing period.

slaperche-scality (changes requested):

buildchain/buildchain/builder.py
alexandre-allard: Yeah, I saw it afterward, forgot to change it.

buildchain/buildchain/ui.py
alexandre-allard: Oh yeah, wanted to add the schema like the other modules, just forgot to do it before pushing. Will do

images/metalk8s-ui/conf/nginx.conf
alexandre-allard: Hum no, forgot to remove it.

buildchain/buildchain/ui.py
alexandre-allard: Nice. Will do.

images/metalk8s-ui/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG NGINX_IMAGE_VERSION=1.15.8
+
+FROM nginx:${NGINX_IMAGE_VERSION}
+
+COPY metalk8s-ui-nginx.conf /etc/nginx/conf.d/default.conf
+
+RUN rm -rf /usr/share/nginx/html/*
+
+COPY ui /usr/share/nginx/html/
+COPY docs/html /usr/share/nginx/docs
```
NicolasT: No, not fond of this. `docs` are part of *DocumentRoot* as is `ui`. So, copy `ui` into `/usr/share/nginx/html`, copy `docs` into `/usr/share/nginx/html`, set up `/usr/share/nginx/html` as the server root, and that should basically be it.

NicolasT (dismissed):

images/metalk8s-ui/Dockerfile
alexandre-allard: Ok, wasn't sure of how we wanted to do this, will do.

buildchain/buildchain/docs.py
```diff
@@ -45,28 +45,35 @@
 
 
 DocTarget = namedtuple('DocTarget', ('name', 'command', 'target'))
-DOC_BUILD_DIR : Path = constants.ROOT/'docs/_build'
-DOC_PDF_FILE  : Path = DOC_BUILD_DIR/'latex/{}.pdf'.format(config.PROJECT_NAME)
+DOC_PDF_FILE : Path = constants.DOCS_BUILD_ROOT/'latex/{}.pdf'.format(
```
slaperche-scality, line 48: minor: when the opening parenthesis is on its own line, the closing as well:

```python
DOC_PDF_FILE : Path = constants.DOCS_BUILD_ROOT/'latex/{}.pdf'.format(
    config.PROJECT_NAME
)
```

buildchain/buildchain/ui.py
```diff
@@ -0,0 +1,73 @@
...
+            utils.bind_mount(
+                target=Path('/home/node/build'),
+                source=constants.UI_BUILD_ROOT,
+            ),
+        ],
+    )
+
+    return {
+        'actions': [build_ui],
+        'title': utils.title_with_target1('UI BUILD'),
```
slaperche-scality: s/UI BUILD/NPM BUILD/ maybe

buildchain/buildchain/ui.py
```diff
@@ -4,15 +4,16 @@
 
 Overview:
 
-┌───────┐    ┌──────────┐
-│ mkdir │───>│ build ui │
-└───────┘    └──────────┘
+┌────┐    ┌───────┐    ┌──────────┐    ┌───────────┐
+│ ui │───>│ mkdir │───>│ ui build │───>│ ui config │
```
slaperche-scality: UI configuration (that way you don't need to add `config` to the dict`)

buildchain/buildchain/ui.py
```diff
@@ -0,0 +1,73 @@
+# coding: utf-8
+
+"""Tasks to build the MetalK8s UI.
+
+Overview:
+
+┌───────┐    ┌──────────┐
+│ mkdir │───>│ build ui │
```
slaperche-scality: build UI

slaperche-scality (commented):
only a few minor remarks :wink:

ui/Dockerfile
alexandre-allard: So that we can also use that Dockerfile to build the npm stuff regardless of the buildchain.

slaperche-scality (approved):

PR #2155, opened 2019-12-19
salt: Rename storageClassName to storageClass

Ebaneck:
**Component**:

'salt'

**Context**: 

Keep consistency

**Summary**:

The storageClassName key in the pillars actually contains the entire description
of a storageclass and not just the name. The key name should be reflected as such

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -226,7 +226,7 @@ class Volume(object):
         if device_info.has_partition:
             raise Exception('backing device `{}` contains a partition table'\
                             .format(self.path))
-        storage_class = self.get('spec.storageClassName')
+        storage_class = self.get('spec.storageClass')
         # If we got a string that means the name wasn't replaced by the object.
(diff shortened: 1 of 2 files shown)


slaperche-scality:
> The storageClassName key in the pillars actually contains the entire description
of a storageclass and not just the name.

Actually it can contains just the name (if we failed to get the StorageClass information) :stuck_out_tongue: 

Nothing to handle on the compatibility side? (I guess it's OK since we refresh the pillar in our states, but worth a check).

Ebaneck:
> > The storageClassName key in the pillars actually contains the entire description
> > of a storageclass and not just the name.
> 
> Actually it can contains just the name (if we failed to get the StorageClass information) stuck_out_tongue
> 
> Nothing to handle on the compatibility side? (I guess it's OK since we refresh the pillar in our states, but worth a check).

As hinted by you in person, there is nothing to handle on the compatibility side...

slaperche-scality (approved):

PR #2174, opened 2020-01-06
charts: Fix the missing `record: node_namespace_pod_container`

### LINKED ISSUE #2172: CPU usage graph is broken due to "outdated" query
**Component**: Grafana dashboard / Prometheus operator

**What happened**: The graph of CPU usage in Grafana is not showing anything.

Screenshot

Note that this is not the only dashboard having this symptom, everything that uses the same query is broken, therefore, every dashboard under `Kubernetes / Compute Resources /`

**What was expected**: I was expecting to see a graph showing the CPU usage of each resources. But I have empty graph even though there's resources running in my cluster.

**Steps to reproduce**: install metalk8s 2.4.1

**Resolution proposal** : I might be wrong here, so don't follow this blindly, but I think that the following query is outdated
```
sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{cluster="$cluster"}) by (namespace)
```
And should be replaced by:
```
sum(namespace:container_cpu_usage_seconds_total:sum_rate{cluster="$cluster"}) by (namespace)
```

Ebaneck:
**Component**:

'charts', 'grafana',

**Context**: 

See #2172 

**Summary**:

Update the prometheus-operator charts in order to fix missing CPU usage graph stats.

**Acceptance criteria**: 

- Working CPU usage graphs

cpu

---

Closes: #2172


NicolasT:
Is there an upstream bug about this, and can we upstream the fix?

NicolasT:
Related, if fixed upstream already (to be checked of course), can we update the chart instead?

Ebaneck:
> Is there an upstream bug about this, and can we upstream the fix?

Yes, i think there is an upstream bug ... I would have to double-check now

Here: https://github.com/helm/charts/issues/19597

Ebaneck:
> Related, if fixed upstream already (to be checked of course), can we update the chart instead?

Yes, updating the charts looks like the best thing to do since it has been fixed as from this line

https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules-1.14/k8s.rules.yaml#L30

tcarmet (approved):
I'm glad this was fixed upstream, LGTM for the upstream chart upgrade

tcarmet:
and thanks for fixing the issue :heart:

thomasdanan (approved):
I have checked all dashboards and everything seems to work fine

PR #2178, opened 2020-01-09
docs: Enhance Bootstrap node deployment docs

### LINKED ISSUE #2114: Enhance Bootstrap Node Deployment documentation
**Component**:

'documentation'

**Why this is needed**:

https://metal-k8s.readthedocs.io/en/latest/quickstart/bootstrap.html
Some information are missing in this section

**What should be done**:

- Better explain bootstrap configuration file
- Describe all possible options (including proxy, ipip, etc …)
- remove prometheus volume provisioning (as it is better to do it at the end of cluster installation)
- build troubleshooting content in ops guide and provide a short link from there
- Validate the install (pods running and access to the UI). Explain that monitoring from UI will not be available until prometheus volumes are provisioned

**Test plan**:

Ebaneck:
**Component**:

'docs'

**Context**: 

See #2114 

**Summary**:

This PR adds the following: 
- Attempt to better explain  the bootstrap configuration file
- Adds minor troubleshooting guide
- Move the Prometheus volume provisioning step to post-install

**Acceptance criteria**: 

- Enhanced bootstrap documentation

---

Closes: #2114

### DIFF
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -27,8 +27,11 @@ Configuration
 
       root@bootstrap $ mkdir /etc/metalk8s
 
-#. Create the :file:`/etc/metalk8s/bootstrap.yaml` file. Change the networks,
-   IP address, and hostname to conform to your infrastructure.
+#. Create the :file:`/etc/metalk8s/bootstrap.yaml` file.
+   This file contains initial configuration settings which are mandatory for
+   setting up a MetalK8s :term:`Bootstrap node`.
+   Change the networks, IP address, and hostname fields to conform to your
+   infrastructure.
 
    .. code-block:: yaml
 
@@ -37,6 +40,8 @@ Configuration
       networks:
         controlPlane: <CIDR-notation>
         workloadPlane: <CIDR-notation>
+        pods: <CIDR-notation>
+        services: <CIDR-notation>
       proxies:
         http: <http://proxy-ip:proxy-port>
         https: <https://proxy-ip:proxy-port>
@@ -48,20 +53,58 @@ Configuration
       archives:
         - <path-to-metalk8s-iso>
 
-The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
-the bootstrap script is executed, those ISOs are automatically mounted and the
-system is configured to re-mount them automatically after a reboot.
+The ``networks`` field specifies a range of IP addresses written in CIDR
+notation for it's various subfields.
+
+      The ``controlPlane`` and ``workloadPlane`` entries are **mandatory**.
+      These values specify the range of IP addresses that will be used at the
+      host level for each member of the cluster.
+
+      .. code-block:: yaml
+
+            networks:
+              controlPlane: 10.200.1.0/28
+              workloadPlane: 10.200.1.0/28
+
+      All nodes within the cluster **must** connect to both the control plane
+      and workload plane networks. If the same network range is chosen for both
+      the control plane and workload plane networks then the same interface
+      may be used.
+
+      The ``pods`` and ``services`` fields are not mandatory, though can be
+      changed to match the constraints of existing networking infrastructure
+      (for example, if all or part of these default subnets is already routed).
+      During installation, by default ``pods`` and ``services`` are set to the
+      following values below if omitted.
+
+      For **production clusters**, we advise users to anticipate future
+      expansions and use sufficiently large networks for pods and services.
+
+      .. code-block:: yaml
+
+            networks:
+              pods: 10.233.0.0/16
+              services: 10.96.0.0/12
 
 The ``proxies`` field can be omitted if there is no proxy to configure.
 The 2 entries ``http`` and ``https`` are used to configure the containerd
 daemon proxy to fetch extra container images from outstide the MetalK8s
 cluster.
 The ``no_proxy`` entry specifies IPs that should be excluded from proxying,
-it must a list of hosts, IP addresses or IP ranges in CIDR format.
+it must be a list of hosts, IP addresses or IP ranges in CIDR format.
+For example;
 
-.. todo::
+   .. code-block:: shell
 
-   - Explain the role of this config file and its values
+      no_proxy:
+        - localhost
+        - 127.0.0.1
+        - 10.10.0.0/16
+        - 192.168.0.0/16
+
+The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
+the bootstrap script is executed, those ISOs are automatically mounted and the
+system is configured to re-mount them automatically after a reboot.
 
 
 .. _Bootstrap SSH Provisioning:
@@ -105,6 +148,8 @@ SSH Provisioning
          user@host $ ssh-copy-id -i /tmp/salt-bootstrap.pub root@<node_hostname>
 
 
+.. _Bootstrap installation:
+
 Installation
 ------------
 
@@ -123,29 +168,18 @@ Bootstrap node.
     destination fields of IP packets to correspond to the MAC address(es)),
     :ref:`IP-in-IP needs to be enabled<enable IP-in-IP>`.
 
-Provision Storage for Prometheus Services
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-After bootstrapping the cluster, the Prometheus and AlertManager services used
-to monitor the system will not be running (their respective :term:`Pods <Pod>`
-will remain in a **Pending** state), because they require persistent storage to
-be available. You can either provision these storage volumes on the bootstrap
-node, or later on other nodes joining the cluster. Templates for the required
-volumes are available in :download:`examples/prometheus-sparse.yaml
-<../../examples/prometheus-sparse.yaml>`. Note, however, these templates use
-the `sparseLoopDevice` *Volume* type, which is not suitable for production
-installations. Refer to :ref:`volume-management` for more information on how to
-provision persistent storage.
+Validate the install
+^^^^^^^^^^^^^^^^^^^^
+- Check that all :term:`Pods <Pod>` on the Bootstrap node are in the
+  **Running** state. Note that Prometheus and Alertmanager pods will remain in
+  a **Pending** state until their respective persistent storage volumes are
+  provisioned.
 
 .. note::
 
-   When deploying using Vagrant, persistent volumes for Prometheus and
-   AlertManager are already provisioned.
-
-Validate the Installation
-^^^^^^^^^^^^^^^^^^^^^^^^^
-Check if all :term:`Pods <Pod>` on the Bootstrap node are running.
-
-.. note::
+   The administrator :term:`kubeconfig` file is used to configure access to
+   Kubernetes when used with :term:`kubectl` as shown below. This file contains
+   sensitive information and should be kept securely.
 
    On all subsequent :term:`kubectl` commands, you may omit the
    ``--kubeconfig`` argument if you have exported the ``KUBECONFIG``
@@ -162,42 +196,51 @@ Check if all :term:`Pods <Pod>` on the Bootstrap node are running.
 
    root@bootstrap $ kubectl get nodes --kubeconfig /etc/kubernetes/admin.conf
    NAME                   STATUS    ROLES                         AGE       VERSION
-   bootstrap              Ready     bootstrap,etcd,infra,master   17m       v1.11.7
+   bootstrap              Ready     bootstrap,etcd,infra,master   17m       v1.15.5
 
    root@bootstrap $ kubectl get pods --all-namespaces -o wide --kubeconfig /etc/kubernetes/admin.conf
-   NAMESPACE             NAME                                             READY     STATUS    RESTARTS   AGE       IP              NODE        NOMINATED NODE
-   kube-system           calico-kube-controllers-b7bc4449f-6rh2q          1/1       Running   0          4m        10.233.132.65   bootstrap   <none>
-   kube-system           calico-node-r2qxs                                1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           coredns-7475f8d796-8h4lt                         1/1       Running   0          4m        10.233.132.67   bootstrap   <none>
-   kube-system           coredns-7475f8d796-m5zz9                         1/1       Running   0          4m        10.233.132.66   bootstrap   <none>
-   kube-system           etcd-bootstrap                                   1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           kube-apiserver-bootstrap                         2/2       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           kube-controller-manager-bootstrap                1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           kube-proxy-vb74b                                 1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           kube-scheduler-bootstrap                         1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           repositories-bootstrap                           1/1       Running   0          4m        172.21.254.12   bootstrap   <none>
-   kube-system           salt-master-bootstrap                            2/2       Running   0          4m        172.21.254.12   bootstrap   <none>
-   metalk8s-ingress      nginx-ingress-controller-46lxd                   1/1       Running   0          4m        10.233.132.73   bootstrap   <none>
-   metalk8s-ingress      nginx-ingress-default-backend-5449d5b699-8bkbr   1/1       Running   0          4m        10.233.132.74   bootstrap   <none>
-   metalk8s-monitoring   alertmanager-main-0                              2/2       Running   0          4m        10.233.132.70   bootstrap   <none>
-   metalk8s-monitoring   alertmanager-main-1                              2/2       Running   0          3m        10.233.132.76   bootstrap   <none>
-   metalk8s-monitoring   alertmanager-main-2                              2/2       Running   0          3m        10.233.132.77   bootstrap   <none>
-   metalk8s-monitoring   grafana-5cb4945b7b-ltdrz                         1/1       Running   0          4m        10.233.132.71   bootstrap   <none>
-   metalk8s-monitoring   kube-state-metrics-588d699b56-d6crn              4/4       Running   0          3m        10.233.132.75   bootstrap   <none>
-   metalk8s-monitoring   node-exporter-4jdgv                              2/2       Running   0          4m        172.21.254.12   bootstrap   <none>
-   metalk8s-monitoring   prometheus-k8s-0                                 3/3       Running   1          4m        10.233.132.72   bootstrap   <none>
-   metalk8s-monitoring   prometheus-k8s-1                                 3/3       Running   1          3m        10.233.132.78   bootstrap   <none>
-   metalk8s-monitoring   prometheus-operator-64477d4bff-xxjw2             1/1       Running   0          4m        10.233.132.68   bootstrap   <none>
+   NAMESPACE             NAME                                                      READY   STATUS    RESTARTS   AGE     IP               NODE            NOMINATED NODE   READINESS GATES
+   kube-system           calico-kube-controllers-7c9944c5f4-h9bsc                  1/1     Running   0          6m29s   10.233.220.129   bootstrap   <none>           <none>
+   kube-system           calico-node-v4qhb                                         1/1     Running   0          6m29s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           coredns-ff46db798-k54z9                                   1/1     Running   0          6m29s   10.233.220.134   bootstrap   <none>           <none>
+   kube-system           coredns-ff46db798-nvmjl                                   1/1     Running   0          6m29s   10.233.220.132   bootstrap   <none>           <none>
+   kube-system           etcd-bootstrap                                            1/1     Running   0          5m45s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           kube-apiserver-bootstrap                                  1/1     Running   0          5m57s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           kube-controller-manager-bootstrap                         1/1     Running   0          7m4s    10.200.3.152     bootstrap   <none>           <none>
+   kube-system           kube-proxy-n6zgk                                          1/1     Running   0          6m32s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           kube-scheduler-bootstrap                                  1/1     Running   0          7m4s    10.200.3.152     bootstrap   <none>           <none>
+   kube-system           repositories-bootstrap                                    1/1     Running   0          6m20s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           salt-master-bootstrap                                     2/2     Running   0          6m10s   10.200.3.152     bootstrap   <none>           <none>
+   kube-system           storage-operator-7567748b6d-hp7gq                         1/1     Running   0          6m6s    10.233.220.138   bootstrap   <none>           <none>
+   metalk8s-ingress      nginx-ingress-control-plane-controller-5nkkx              1/1     Running   0          6m6s    10.233.220.137   bootstrap   <none>           <none>
+   metalk8s-ingress      nginx-ingress-controller-shg7x                            1/1     Running   0          6m7s    10.233.220.135   bootstrap   <none>           <none>
+   metalk8s-ingress      nginx-ingress-default-backend-7d8898655c-jj7l6            1/1     Running   0          6m7s    10.233.220.136   bootstrap   <none>           <none>
+   metalk8s-monitoring   alertmanager-prometheus-operator-alertmanager-0           0/2     Pending   0          6m1s    <none>           <none>      <none>           <none>
+   metalk8s-monitoring   prometheus-operator-grafana-775fbb5b-sgngh                2/2     Running   0          6m17s   10.233.220.130   bootstrap   <none>           <none>
+   metalk8s-monitoring   prometheus-operator-kube-state-metrics-7587b4897c-tt79q   1/1     Running   0          6m17s   10.233.220.131   bootstrap   <none>           <none>
+   metalk8s-monitoring   prometheus-operator-operator-7446d89644-zqdlj             1/1     Running   0          6m17s   10.233.220.133   bootstrap   <none>           <none>
+   metalk8s-monitoring   prometheus-operator-prometheus-node-exporter-rb969        1/1     Running   0          6m17s   10.200.3.152     bootstrap   <none>           <none>
+   metalk8s-monitoring   prometheus-prometheus-operator-prometheus-0               0/3     Pending   0          5m50s   <none>           <none>      <none>           <none>
+   metalk8s-ui           metalk8s-ui-6f74ff4bc-fgk86                               1/1     Running   0          6m4s    10.233.220.139   bootstrap   <none>           <none>
 
-Check that you can access the MetalK8s GUI, following
-:ref:`this procedure <installation-services-admin-ui>`.
+- From the console output above, :term:`Prometheus` and :term:`Alertmanager`
+  pods are in a ``Pending`` state because their respective persistent
+  storage volumes need to be provisioned. To provision these persistent storage
+  volumes, follow :ref:`this procedure <Provision Prometheus storage>`.
 
-.. todo::
+- Check that you can access the MetalK8s GUI after the
+  :ref:`installation <Bootstrap installation>` is completed by following
+  :ref:`this procedure <installation-services-admin-ui>`.
 
-   Troubleshooting section
+- At this stage, the MetalK8s GUI should be up and ready for you to
+  explore.
 
-   - Mention ``/var/log/metalk8s-bootstrap.log`` and the command-line options
-     for verbosity.
-   - Add Salt master/minion logs, and explain how to run a specific state from
-     the Salt master.
-   - Then refer to a troubleshooting section in the installation guide.
+  .. note::
+
+     Monitoring through the MetalK8s GUI will not be available until persistent
+     storage volumes for both Prometheus and Alertmanager have been successfully
+     provisioned.
+
+- If you encouter an error during installation or have difficulties
+  validating a fresh MetalK8s installation, visit our
+  :ref:`Troubleshooting guide <Troubleshooting Guide>`.
--- a/docs/installation/post-install.rst
+++ b/docs/installation/post-install.rst
@@ -1,6 +1,29 @@
 Post-Installation Procedure
 ===========================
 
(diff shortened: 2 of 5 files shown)


docs/installation/bootstrap.rst
```diff
@@ -47,21 +50,37 @@ Configuration
...
 The ``no_proxy`` entry specifies IPs that should be excluded from proxying,
-it must a list of hosts, IP addresses or IP ranges in CIDR format.
+it must be a list of hosts, IP addresses or IP ranges in CIDR format.
+For example;
 
-.. todo::
+   .. code-block:: shell
 
-   - Explain the role of this config file and its values
+      NO_PROXY="localhost,127.0.0.1,10.10.0.0/16,192.168.0.0/16"
```
NicolasT: This seems confusing: not in YAML-format, and not using the same key name (caps vs. lower). Also, the YAML value must be a list (type), not a comma-separate string.

docs/installation/bootstrap.rst
```diff
@@ -120,31 +141,17 @@ Bootstrap node.
...
-.. note::
-
-   When deploying using Vagrant, persistent volumes for Prometheus and
-   AlertManager are already provisioned.
-
 Validate the install
 ^^^^^^^^^^^^^^^^^^^^
-Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
-``Running`` state.
+- Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
```
NicolasT: If we don't mention provisioning storage for Prometheus and such before this, then they *can't* be all in `Running` state.

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,54 @@
...
+output does not provide sufficient information in order to pin-point the cause
+of failure, then re-run the installation with verbose arguments(--verbose).
+
+.. parsed-literal::
+
+   root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh --verbose
+
+Errors after restarting the Bootstrap node
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you restart the Bootstrap node and for some reason, some containers
```
NicolasT: Missing an `and` or something here.

NicolasT:
Please review @lucieleonard (or someone in your team).

thomasdanan:
Shouldn't we explain from where get the iso?

docs/installation/bootstrap.rst
```diff
@@ -47,21 +50,41 @@ Configuration
...
-   - Explain the role of this config file and its values
+      no_proxy:
+        - localhost
+        - 127.0.0.1
+        - 10.10.0.0/16
+        - 192.168.0.0/16
 
+The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
+the bootstrap script is executed, these ISOs are automatically mounted and the
+system is configured to re-mount them automatically after a reboot.
```
thomasdanan, line 107: Is the bootstrap script really mounting those iso when we run it? isn't it a manual step to perform? and when we execute the bootstrap script, we update system configuration to mount the iso automatically after a reboot?

docs/installation/bootstrap.rst
```diff
@@ -187,16 +205,13 @@ Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
    metalk8s-monitoring   prometheus-k8s-1                                 3/3       Running   1          3m        10.233.132.78   bootstrap   <none>
```
thomasdanan: Ideally use the right version of K8S instead of 1.11.7

docs/installation/bootstrap.rst
```diff
@@ -187,16 +205,13 @@ Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
...
-:ref:`this procedure <quickstart-services-admin-ui>`.
-
-Troubleshooting
-^^^^^^^^^^^^^^^
+- Check and verify that you can access the MetalK8s GUI after the
+  :ref:`installation <Bootstrap installation>` is completed by following
+  :ref:`this procedure <quickstart-services-admin-ui>`.
 
-.. todo::
+- At this stage, the MetalK8s GUI should be up and ready for you to
```
thomasdanan: even if prom and am PV have not been provisioned?

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Forgot the MetalK8s GUI password
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you forget the MetalK8s GUI username and/or password combination,
+follow
+:ref:`this procedure <Administering-MetalK8s-GUI-Kubernetes-API-and-Salt-API>`
+to reset or change it.
+
+
+.. todo::
```
thomasdanan, line 198: are you going to handle this in this PR?

thomasdanan (dismissed):
Nice job !!!
Just a few little comments + 

- Shouldn't we explain from where get the iso? and remind it should not be stored on /tmp?
- I thought we wanted to manage IP IP configuration through the bootstrap.yaml config file but it may not be available yet?

docs/installation/bootstrap.rst
Ebaneck: fixed

docs/installation/bootstrap.rst
Ebaneck: Rephrased and fixed!

Ebaneck:
> Nice job !!!
> Just a few little comments +
> 
>     * Shouldn't we explain from where get the iso? and remind it should not be stored on /tmp?
> 

We could but it seems the built and released ISO's are strictly internal.
Also the ISO store path now reads it should be kept in the users home directory.

>     * I thought we wanted to manage IP IP configuration through the bootstrap.yaml config file but it may not be available yet?

I think this feature is not ready yet for 2.4.2 and btw, we have a small guide added in the `Installation` section which hints users on how to enable `IP-in-IP`.

NicolasT:
> Shouldn't we explain from where get the iso? and remind it should not be stored on /tmp?

I believe that's handled in @lucieleonard's PR.

Ebaneck:
> > Shouldn't we explain from where get the iso? and remind it should not be stored on /tmp?
> 
> I believe that's handled in @lucieleonard's PR.

True.

docs/operation/troubleshooting.rst
Ebaneck: I think the two last ones are not mission critical

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
+The ``networks`` field specifies a range of IP addresses written in CIDR
+notation for the various fields.
+
+In production clusters, we advise users to choose a
+subnet range of atleast `/16` in order to have sufficient IP addresses keeping
+in mind that users infrastructure will grow in the near future.
+
+      The ``controlPlane`` and ``workloadPlane`` entries are **mandatory**.
+      These values specify the range of IP addresses that will be attributed to
+      controlPlane and workloadPlane nodes.
```
```suggestion
      control plane and workload plane nodes.
```

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
 
-The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
-the bootstrap script is executed, those ISOs are automatically mounted and the
-system is configured to re-mount them automatically after a reboot.
+
+The ``networks`` field specifies a range of IP addresses written in CIDR
+notation for the various fields.
+
+In production clusters, we advise users to choose a
+subnet range of atleast `/16` in order to have sufficient IP addresses keeping
```
gdemonet: `/16` is definitely **not** a requirement for control/workload plane networks: those are real IPs, so as long as you have all your nodes in them, it's enough (and 5 nodes fit in a `/28` if needed).
`/16` is however recommended for internal, virtual networks that are used for Pod and Service ClusterIPs.

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
+in mind that users infrastructure will grow in the near future.
```
gdemonet: `users infrastructure` doesn't mean much to me. The recommendation makes sense however, so maybe rephrasing it to something like:

> For production clusters, we advise users to anticipate future expansions and use sufficiently large networks for control plane and workload plane. A `/24` should be enough. Note that aggregation of subnet segments is not yet achievable with this configuration file.

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
+
+      The ``controlPlane`` and ``workloadPlane`` entries are **mandatory**.
+      These values specify the range of IP addresses that will be attributed to
```
gdemonet: Not `will be attributed`: those networks are the ranges of IP addresses used at the host level for each member of the cluster. It could also make sense to mention that **all** nodes in the cluster **must** have an interface connected to each of those networks (CP and WP) - this may be the same interface if we only use a single network range for both.

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
+      controlPlane and workloadPlane nodes.
+
+      .. code-block:: shell
+
+            networks:
+              controlPlane: 10.200.0.0/16
+              workloadPlane: 10.200.0.0/16
+
+      The ``pods`` and ``services`` fields at not mandatory but can be set
+      based on users requirements.
```
gdemonet: Maybe suggest why one would want to change these (`users requirements` is too vague IMHO). Something like:

> The ``pods`` and ``services`` fields are not mandatory, though can be changed to match the constraints of existing networking infrastructure (e.g. if all or part of these default subnets is already routed).

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,57 @@ Configuration
...
+      During installation, by default ``pods`` and ``services`` are set to the
+      following values below if omitted. This is done in order that pods and
+      services within the cluster obtain unique IP addresses and avoid any
+      conflicts.
```
gdemonet: The explanation lacks a useful disclaimer: those are internal, virtual networks. Nothing has to be done on the infrastructure side, as our cluster will handle configuration of the routing tables and allocation pools on its own (at the `iptables` level). We just need to make sure that they don't conflict with existing, routed networks.

docs/installation/bootstrap.rst
```diff
@@ -160,43 +191,51 @@ Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
...
+
+- Check that you can access the MetalK8s GUI after the
+  :ref:`installation <Bootstrap installation>` is completed by following
+  :ref:`this procedure <quickstart-services-admin-ui>`.
+
+  .. note::
+
+     The MetalK8s GUI will only become accessible when persistent storage
+     volumes for both Prometheus and Alertmanager have been successfully
+     provisioned.
```
gdemonet: That's not true anymore, UI is accessible, but Monitoring features will be missing.

docs/installation/post-install.rst
```diff
@@ -1,2 +1,25 @@
...
+<../../examples/prometheus-sparse.yaml>`.
+
+Note, however, these templates use the `sparseLoopDevice` *Volume* type, which
+is not suitable for production installations. Refer to :ref:`volume-management`
+for more information on how to provision persistent storage.
+
+.. note::
+
+   When deploying using Vagrant, persistent volumes for Prometheus and
+   AlertManager are already provisioned.
```
gdemonet, line 25: Should this still be part of the Installation guide? cc @thomasdanan

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
+
+.. _Troubleshooting Guide:
+
+Troubleshooting Guide
+^^^^^^^^^^^^^^^^^^^^^
+
+This section highlights some of the common problems users face during and
+after a MetalK8s installation. If you do not find a solution to a problem you
+are facing, please reach out to us or create a
+`Github issue <https://github.com/scality/metalk8s/issues>`_.
```
gdemonet: Mention Scality support?

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Bootstrap Installation Errors
++++++++++++++++++++++++++++++
+
+Bootstrap Installation fails with no straightforward reason
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If during a MetalK8s installation you encounter a failure and the console
+output does not provide sufficient information in order to pin-point the cause
+of failure, then re-run the installation with verbose arguments(--verbose).
```
gdemonet:
```suggestion
of failure, then re-run the installation with the verbose flag (``--verbose``).
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. parsed-literal::
+
+   root@bootstrap $ /srv/scality/metalk8s-|release|/bootstrap.sh --verbose
+
+Errors after restarting the Bootstrap node
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you restart the Bootstrap node and for some reason, some containers
```
gdemonet: nitpick: `reboot` is more obvious than `restart`, maybe I'm wrong.

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+.. code-block:: shell
+
+   kubectl -n <namespace-name> describe pod <pod name>
+
+   kubectl -n <namespace-name> logs -p  <pod name>
+
+Persistent Volume Claim(PVC) stuck in "Pending" state
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If after provisioning a volume for a Pod say Prometheus and the PVC still
```
gdemonet: For clarity:

```suggestion
If after provisioning a volume for a Pod (e.g. Prometheus) and the PVC still
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+hangs in a ``Pending`` state, then try checking the following.
```
gdemonet: Use an unordered list for the various checks

```suggestion
hangs in a ``Pending`` state, then try checking the following:

- Volumes have been provisioned and their status is ``Ready``: [...]
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Check that the volumes have been provisioned and it's status is ``Ready``:
```
gdemonet: Volume being an actual CRD, it could make sense to capitalize occurences of `Volumes`

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. code-block:: shell
+
+   kubectl describe volume <volume name>
+
+Check that a corresponding Persistent volume exist:
```
gdemonet:
```suggestion
Check that a corresponding PersistentVolume exist:
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+.. code-block:: shell
+
+   kubectl get pv
+
+If no Persistent volume exist, then check that the storage operator is up
+and running.
+
+.. code-block:: shell
+
+   kubectl -n kube-system get pods
```
gdemonet: Better to look at the Deployment object I believe:

```suggestion
   kubectl -n kube-system get deployments storage-operator
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+To get possible clues about this error, run the following commands and inspect
+the output.
+
+.. code-block:: shell
+
+   kubectl -n <namespace-name> describe pod <pod name>
+
+   kubectl -n <namespace-name> logs -p  <pod name>
+
+Persistent Volume Claim(PVC) stuck in "Pending" state
```
gdemonet, line 86: Great troubleshooting suggestions, thanks!! I think we could improve those a bit by providing expected output, this would help users being more autonomous (if they have a hard time understanding Kubernetes)

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+Access to MetalK8s GUI fails with "undefined backend"
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If in the cause of using the MetalK8s GUI, you encounter an "undefined
+backend" error then perform the following checks.
+
+Check that the Ingress pods are running:
+
+.. code-block:: shell
+
+   kubectl -n metalk8s-ingress get pods
```
gdemonet: Ditto, Deployments give a simpler information, including the number of expected Pods compared to the actual number.

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Check the Ingress controller logs:
+
+.. code-block:: shell
+
+   kubectl -n metalk8s-ingress logs <nginx-ingress-pod-name>
```
gdemonet: Since Metal UI runs behind the control-plane Ingress, it would make sense to:

- provide a query to list all relevant Pods (using a label selector, e.g. `-l release=nginx-ingress-control-plane`)
- then show how to read the logs using the target Pod's name

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Pod and Service CIDR conflicts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If after installation of a MetalK8s cluster you notice that Pod to pod
```
gdemonet:
```suggestion
If after installation of a MetalK8s cluster you notice that Pod-to-Pod
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+communication has routing problems, perform the following.
```
gdemonet: Ditto, use a colon and list all checks in a unordered list

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Check that Pod network subnets do not conflict with host networks by verifying
+allocated pod IP addresses:
```
gdemonet: Maybe not the most straightforward approach. Let's add a first check on the actual network values:

```rst
- Check the configured values for the internal Pod and Service networks:

   .. code-block:: shell

      root@bootstrap # salt-call pillar.get metalk8s:networks

   Make sure the configured IP ranges (CIDR notation) do not conflict with your infrastructure.
```

After a check like this one, there shouldn't be a need to go through the `kubectl get pods` command. Going further may look into more advanced commands, such as `traceroute` or `tcpdump`. Not sure it's worth going too far in this iteration though.

docs/operation/troubleshooting.rst
gdemonet: Troubleshooting for networking is indeed a large part and may be relegated to a future enhancement (worth opening a ticket to track this however, cc @thomasdanan).

For Salt logs, I think it would make sense to at least document it somewhere. Let's keep it for 2.4.2, but not this PR, which is already quite big.

gdemonet (changes requested):
Some comments, but good job overall, this doc is starting to look professional :)

docs/operation/troubleshooting.rst
Ebaneck: True... only that the deployments returns the following which is not sufficient i think :<

```
[root@bootstrap vagrant]# kubectl -n metalk8s-ingress get deployments
NAME                            READY   UP-TO-DATE   AVAILABLE   AGE
nginx-ingress-default-backend   1/1     1            1           4d18h

```

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,60 @@ Configuration
...
+              workloadPlane: 10.200.1.0/28
+
+      The ``pods`` and ``services`` fields are not mandatory, though can be
+      changed to match the constraints of existing networking infrastructure
+      (for example, if all or part of these default subnets is already routed).
+      During installation, by default ``pods`` and ``services`` are set to the
+      following values below if omitted.
+
+      All nodes within the cluster **must** connect to both the control plane
+      and workload networks. If the same network range is chosen for both the
```
```suggestion
      and workload plane networks. If the same network range is chosen for both the
```

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,60 @@ Configuration
...
+      control plane and workload plane networks then the same interface maybe
```
```suggestion
      control plane and workload plane networks then the same interface may be
```

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,60 @@ Configuration
...
+notation for the various fields.
+
+For production clusters, we advise users to anticipate future expansions and
+use sufficiently large networks for control plane and workload plane.
+
+      The ``controlPlane`` and ``workloadPlane`` entries are **mandatory**.
+      These values specify the range of IP addresses that will be attributed to
+      control plane and workload plane nodes.
+
+      .. code-block:: shell
```
gdemonet: For this one, and others below, why not use `.. code-block:: yaml`?

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,60 @@ Configuration
...
+
+            networks:
+              controlPlane: 10.200.1.0/28
+              workloadPlane: 10.200.1.0/28
+
+      The ``pods`` and ``services`` fields are not mandatory, though can be
+      changed to match the constraints of existing networking infrastructure
+      (for example, if all or part of these default subnets is already routed).
```
gdemonet, line 76: :+1:

docs/installation/bootstrap.rst
```diff
@@ -47,21 +52,60 @@ Configuration
...
+      During installation, by default ``pods`` and ``services`` are set to the
+      following values below if omitted.
+
+      All nodes within the cluster **must** connect to both the control plane
+      and workload networks. If the same network range is chosen for both the
+      control plane and workload plane networks then the same interface maybe
+      used.
```
gdemonet: I think this paragraph should be moved just after the `controlPlane`/`workloadPlane` example, for clarity. Here, we're talking about `pods`/`services`, so it's a bit confusing.

docs/installation/bootstrap.rst
```diff
@@ -160,43 +194,51 @@ Check if all :term:`Pods <Pod>` on the Bootstrap node are in the
...
+- At this stage, the MetalK8s GUI should be up and ready for you to
+  explore.
+
+  .. note::
+
+     Monitoring through the MetalK8s GUI will not be available until persistent
+     storage volumes for both Prometheus and Alertmanager have been successfully
+     provisioned.
+
+- If you encouter an error during **installation** or have difficulties
```
gdemonet: Not sure the bold font is useful here.

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Errors after restarting the Bootstrap node
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you reboot the Bootstrap node and for some reason, some containers
+(especially the salt-master container) refuses to start then perform the
+following checks:
+
+- Check and ensure that the **MetalK8s ISO** is mounted properly.
+
+  .. code-block:: shell
```
gdemonet: You need to use `parsed-literal` for substitutions like `|release|` to work

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+    Priority:             2000000000
+    Priority Class Name:  system-cluster-critical
+
+Persistent Volume Claim(PVC) stuck in "Pending" state
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If after provisioning a Volume for a Pod (e.g. Prometheus) and the PVC still
+hangs in a ``Pending`` state, then try checking the following:
+
+- Check that the volumes have been provisioned and it's status is ``Ready``:
```
gdemonet:
```suggestion
- Check that the volumes have been provisioned and are in a **Ready** state:
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+    Name:                 <pod name>
+    Namespace:            kube-system
+    Priority:             2000000000
+    Priority Class Name:  system-cluster-critical
+
+Persistent Volume Claim(PVC) stuck in "Pending" state
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If after provisioning a Volume for a Pod (e.g. Prometheus) and the PVC still
+hangs in a ``Pending`` state, then try checking the following:
```
gdemonet:
```suggestion
hangs in a **Pending** state, then try checking the following:
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+  .. code-block:: shell
+
+     [root@bootstrap vagrant]# kubectl -n kube-system get deployments storage-operator
+     NAME               READY   UP-TO-DATE   AVAILABLE   AGE
+     storage-operator   1/1     1            1           4d22h
+
+Access to MetalK8s GUI fails with "undefined backend"
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If in the cause of using the MetalK8s GUI, you encounter an "undefined
+backend" error then perform the following checks.
```
gdemonet:
```suggestion
backend" error then perform the following checks:
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+        Build:         git-2de5a893a
+        Repository:    https://github.com/kubernetes/ingress-nginx
+        nginx version: openresty/1.15.8.2
+
+Pod and Service CIDR conflicts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If after installation of a MetalK8s cluster you notice that Pod-to-Pod
+communication has routing problems, perform the following:
+
+- Check the configured values for the internal POD and Service networks:
```
gdemonet:
```suggestion
- Check the configured values for the internal Pod and Service networks:
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+         control_plane:
+             172.21.254.0/28
+         pod:
+             10.233.0.0/16
+         service:
+             10.96.0.0/12
+         workload_plane:
+             172.21.254.32/27
+
+  Make sure the configured IP ranges (CIDR notation) do not conflicts with your
```
gdemonet:
```suggestion
  Make sure the configured IP ranges (CIDR notation) do not conflict with your
```

gdemonet (changes requested):
Last changes and it'll be all good!

docs/installation/bootstrap.rst
Ebaneck: true

docs/installation/bootstrap.rst
```diff
@@ -27,8 +27,11 @@ Configuration
 
       root@bootstrap $ mkdir /etc/metalk8s
 
-#. Create the :file:`/etc/metalk8s/bootstrap.yaml` file. Change the networks,
-   IP address, and hostname to conform to your infrastructure.
+#. Create the :file:`/etc/metalk8s/bootstrap.yaml` file.
+   This file contains initial configuration settings which are mandatory for
+   settings up a MetalK8s :term:`Bootstrap node`.
```
gdemonet: Whoops, missed that one:

```suggestion
   setting up a MetalK8s :term:`Bootstrap node`.
```

docs/installation/bootstrap.rst
```diff
@@ -123,29 +168,18 @@ Bootstrap node.
...
-node, or later on other nodes joining the cluster. Templates for the required
-volumes are available in :download:`examples/prometheus-sparse.yaml
-<../../examples/prometheus-sparse.yaml>`. Note, however, these templates use
-the `sparseLoopDevice` *Volume* type, which is not suitable for production
-installations. Refer to :ref:`volume-management` for more information on how to
-provision persistent storage.
+Validate the install
+^^^^^^^^^^^^^^^^^^^^
+- Check that all :term:`Pods <Pod>` on the Bootstrap node are in the
+  ``Running`` state. Note that Prometheus and Alertmanager pods will remain in
```
gdemonet: From what I've seen in @lucieleonard 's PR, user-facing values such as an object's `state` are shown as bold, not monospace:

```suggestion
  **Running** state. Note that Prometheus and Alertmanager pods will remain in
```

docs/installation/bootstrap.rst
```diff
@@ -123,29 +168,18 @@ Bootstrap node.
...
+  a ``Pending`` state until their respective persistent storage volumes are
```
```suggestion
  a **Pending** state until their respective persistent storage volumes are
```

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+  .. parsed-literal::
+
+     [root@bootstrap vagrant]# mount | grep /srv/scality/metalk8s-|release|
+     /home/centos/metalk8s.iso on /srv/scality/metalk8s-|release| type iso9660 (ro,relatime)
+
+
+- If the ISO is unmounted, run the following command which will check the
+  the status of the ISO file and remount it automatically.
+
+  .. code-block:: shell
```
gdemonet:
```suggestion
  .. parsed-literal::
```

gdemonet (changes requested):
Almost :innocent:

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,178 @@
...
+     [root@bootstrap vagrant]# kubectl describe volume test-volume
+      Name:         <volume-name>
+      Status:
+        Conditions:
+          Last Transition Time:  2020-01-14T12:57:56Z
+          Last Update Time:      2020-01-14T12:57:56Z
+          Status:                True
+          Type:                  Ready
+
+- Check that a corresponding PersistentVolume exist:
```
slaperche-scality, line 106: Another thing worth checking is that the PV matches the PVC constraints (size, labels, storage class).

docs/operation/troubleshooting.rst
```diff
@@ -0,0 +1,202 @@
...
+
+  .. code-block:: shell
+
+     [root@bootstrap vagrant]# kubectl get pv
+     NAME                     CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS  STORAGECLASS             AGE       CLAIM
+     <volume-name>              10Gi       RWO            Retain          Bound  <storage-class-name>     4d22h     <persistentvolume-claim-name>
+
+- Check that the PersistentVolume matches the PersistentVolume Claim
+  constraints(size, labels, storage class) by doing the following:
```
slaperche-scality: missing space before `(`

gdemonet (approved):

PR #2179, opened 2020-01-09
Add link to build requirements in README

slaperche-scality:
**Component**:

documentation

**Context**: 

With the current README people will try to build an ISO by running doit right away (as we suggest) and it will likely fail because of some missing requirement and people will have no clue on what to do next because there is no mention nor link of those build requirements here.

**Summary**:

In the README, add a link to the part of the documentation that lists the build requirements.


gdemonet (approved):
Thanks for this :)

PR #2181, opened 2020-01-10
salt: use all etcd servers as apiserver backends

### LINKED ISSUE #2080: `kube-apiserver` instances only rely on a local `etcd` member
**Component**: kubernetes

**What happened**:

In the manifest for `kube-apiserver` we deploy, we set the `etcd` servers only to the local server: `--etcd-servers=https://{{ grains.metalk8s.control_plane_ip }}:2379`

**What was expected**:

We should not rely on the fact that we usually deploy `etcd` next to `k-a`. Instead, all members should be listed in this `--etcd-servers` flag, and the manifests should be updated everytime a new member is added/removed to the `etcd` cluster.

alexandre-allard:
**Component**: salt, apiserver

**Context**:
Until now, we only use the local etcd as the apiserver backend

**Summary**:
Define all clients URLs as etcd servers for API server backend, this way, if the local etcd is
down, the API server is still able to answer.

---

Closes: #2080

### DIFF
--- a/salt/metalk8s/kubernetes/apiserver/installed.sls
+++ b/salt/metalk8s/kubernetes/apiserver/installed.sls
@@ -21,6 +21,16 @@ Set up default basic auth htpasswd:
     - replace: False
 
 {%- set host = grains['metalk8s']['control_plane_ip'] %}
+{%- set etcd_servers = [] %}
+{#- NOTE: Filter out members with empty name as they are not started yet. #}
+{%- for member in pillar.metalk8s.etcd.members | selectattr('name') %}
+{%-   do etcd_servers.extend(member['client_urls']) %}
+{%- endfor %}
+{%- set etcd_servers = etcd_servers | sort %}
+{%- if 'etcd' in pillar.metalk8s.nodes[grains.id].roles %}
+{%-   do etcd_servers.insert(0, "https://" ~ host ~ ":2379") %}
+{%- endif %}
+{%- set etcd_servers = etcd_servers | unique %}
 
 Create kube-apiserver Pod manifest:
   metalk8s.static_pod_managed:
@@ -61,7 +71,7 @@ Create kube-apiserver Pod manifest:
           - --etcd-cafile=/etc/kubernetes/pki/etcd/ca.crt
           - --etcd-certfile=/etc/kubernetes/pki/apiserver-etcd-client.crt
           - --etcd-keyfile=/etc/kubernetes/pki/apiserver-etcd-client.key
-          - --etcd-servers=https://{{ grains.metalk8s.control_plane_ip }}:2379
+          - --etcd-servers={{ etcd_servers | join(",") }}
           - --insecure-port=0
           - --kubelet-client-certificate=/etc/kubernetes/pki/apiserver-kubelet-client.crt
           - --kubelet-client-key=/etc/kubernetes/pki/apiserver-kubelet-client.key


salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -61,7 +65,7 @@ Create kube-apiserver Pod manifest:
           - --etcd-cafile=/etc/kubernetes/pki/etcd/ca.crt
           - --etcd-certfile=/etc/kubernetes/pki/apiserver-etcd-client.crt
           - --etcd-keyfile=/etc/kubernetes/pki/apiserver-etcd-client.key
-          - --etcd-servers=https://{{ grains.metalk8s.control_plane_ip }}:2379
+          - --etcd-servers={{ etcd_servers | unique | sort | join(",") }}
```
alexandre-allard-scality: Maybe we want to keep the local etcd as the first server ?

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -21,6 +21,12 @@ Set up default basic auth htpasswd:
     - replace: False
 
 {%- set host = grains['metalk8s']['control_plane_ip'] %}
+{%- set etcd_servers = [] %}
+{%- for member in pillar.metalk8s.etcd.members | default([]) %}
+{%-   do etcd_servers.extend(member['client_urls']) %}
+{%- endfor %}
+{%- set etcd_servers =
+      (["https://" ~ host ~ ":2379"] + etcd_servers | sort) | unique %}
```
piratos: with the sort doesn't that mean that sometimes the first etcd server kubeapiserver will hit maybe not the local one ? also  did you check when killing the first etcd in the list that kubeapiserver fall over to the next ? I can see some bugs 
here https://github.com/kubernetes/kubernetes/issues/83028 and here https://github.com/kubernetes/kubernetes/issues/72102#issuecomment-460303835

sayf-eddine-scality: sorry used my personal account again :p

gdemonet: > with the sort doesn't that mean that sometimes the first etcd server kubeapiserver will hit maybe not the local one ?

The `etcd_servers` variable will always have the local member in first position, so it shouldn't change this behaviour. However, now that I think of it, we should **only** add this _local member_ **if** it has the `etcd` role. Something like:

```sls
{%- if 'etcd' in pillar.metalk8s.nodes[grains.id].roles %}
{%-   do etcd_servers.insert(0, "https://" ~ host ~ ":2379") %}
{%- endif %}
```

> also did you check when killing the first etcd in the list that kubeapiserver fall over to the next? I can see some bugs here kubernetes/kubernetes#83028 and here kubernetes/kubernetes#72102 (comment)

Indeed, @alexandre-allard-scality is aware of this issue. It is supposed to be fixed in K8s 1.16, from the issues you linked.
If we were to have a failing etcd member, the apiserver deployed on that node will not work either, and this is current behaviour (as we only list the assumed local member). In 2.5 and above, we _should_ have a working fallback to other `etcd` members.
I think it would make sense to add a BDD test for this, and mark it as expected fail if the version is lower than when the fix was introduced. Do you think you can squeeze something like this in this PR @alexandre-allard-scality?

sayf-eddine-scality: oh yes sort is only applied to the remote list mb !

NicolasT: I'd add some parenthesis to make the desired outcome more clear. Precedence between `+` and `|` is not necessarily... 'obvious'.

What we *want* (I guess?) is
```
{%- set etcd_servers =
      (["https://" ~ host ~ ":2379"] + (etcd_servers | sort)) | unique %}
```
This may be the case without the extra parenthesis (and operator precedence could be inferred from the use of parenthesis to bind `| unique` to the combined list), but I prefer to be explicit.

Though not a blocker for this PR as far as I'm concerned :man_shrugging:

alexandre-allard: Applied comment of @gdemonet, we don't need extra parenthesis with this change.

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -21,6 +21,15 @@ Set up default basic auth htpasswd:
     - replace: False
 
 {%- set host = grains['metalk8s']['control_plane_ip'] %}
+{%- set etcd_servers = [] %}
+{%- for member in pillar.metalk8s.etcd.members | default([]) %}
```
TeddyAndrieux: Please filterout etcd member without "name" (no name mean not started yet) (like in installed state of etcd)
`selectattr('name')`

alexandre-allard: done

TeddyAndrieux (dismissed):
LGTM

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -21,6 +21,16 @@ Set up default basic auth htpasswd:
     - replace: False
 
 {%- set host = grains['metalk8s']['control_plane_ip'] %}
+{%- set etcd_servers = [] %}
+{#- NOTE: Filter out members with empty name as they are not started yet. #}
+{%- for member in pillar.metalk8s.etcd.members | selectattr('name') %}
+{%-   do etcd_servers.extend(member['client_urls']) %}
+{%- endfor %}
+{%- set etcd_servers = etcd_servers | sort %}
```
gdemonet, line 29: If you like Jinja tricks:

```sls
{%- set etcd_servers = pillar.metalk8s.etcd.members
                       | selectattr('name')
                       | sum(attribute='client_urls', start=[])
                       | sort %}
```

gdemonet (approved):

PR #2189, opened 2020-01-15
build/debian: Fix Saltstack repository URL

gdemonet:
We pin an old version of Saltstack to use in our product. This version
is no longer available on the "main" repo domain, but instead under the
"archive" domain. URL thus need to change, though only for Debian (as we
already used the "archive" domain for RedHat).

---

**NB:** this currently prevents any CI run for 2.4 or 2.5 from passing the `build` stage.

### DIFF
--- a/packages/debian/download_packages.py
+++ b/packages/debian/download_packages.py
@@ -98,8 +98,6 @@ def fetch_binary(
 
 def add_external_repositories(salt_version: str) -> None:
     """Register Salt and Kubernetes repository."""
-    # Only keep the major and minor version number.
-    salt_version = '.'.join(salt_version.split('.')[:2])
     # TODO: move these static info in `versions.py`.
     repositories = [
         {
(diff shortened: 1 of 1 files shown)


packages/debian/download_packages.py
```diff
@@ -111,12 +111,12 @@ def add_external_repositories(salt_version: str) -> None:
             'name': 'saltstack',
             'key': urllib.parse.urljoin(
                 'https://repo.saltstack.com/',
-                'apt/ubuntu/18.04/amd64/{}/SALTSTACK-GPG-KEY.pub'.format(
+                'apt/ubuntu/18.04/amd64/archive/{}/SALTSTACK-GPG-KEY.pub'.format(
```
Ebaneck: I think i am unable to access the GPG key using an incomplete version prefix such as `2018.3`

We probably need to pass the complete version info `2018.3.4`

Ebaneck (changes requested):

packages/debian/download_packages.py
Ebaneck: Something like this works for me: `https://repo.saltstack.com/apt/ubuntu/18.04/amd64/archive/2018.3.4/SALTSTACK-GPG-KEY.pub`

packages/debian/download_packages.py
```diff
@@ -111,12 +111,12 @@ def add_external_repositories(salt_version: str) -> None:
...
                     salt_version
                 )
             ),
             'source': 'deb http://repo.saltstack.com/apt/ubuntu/'\
-                      '18.04/amd64/{} bionic main'.format(salt_version)
+                      '18.04/amd64/archive/{} bionic main'.format(salt_version)
```
Ebaneck, line 117: Ditto .. i think

gdemonet: Indeed, I used the full version locally and wanted to check with @slaperche-scality to see if I could commit it. Just pushed a fix!

slaperche-scality (dismissed):

slaperche-scality (approved):

Ebaneck (approved):

PR #2197, opened 2020-01-21
tests: Use nodename instead of hostname to compute pod name

TeddyAndrieux:
**Component**:

'tests'

**Context**: 

We still use hostname in our test which is wrong

**Summary**:

Since we can deploy nodes using a node name different from the hostname,
pod deployed can have a name that does no contain the hostname

**Acceptance criteria**: 

Green build

---


gdemonet (approved):
LGTM!

PR #2198, opened 2020-01-21
salt: fix unsafe approach for etcd `initial-cluster-state

### LINKED ISSUE #2102: Not safe approach for etcd `initial-cluster-state`
**Component**:

'salt'

**What happened**:

In our salt states to deploy etcd we set `initial-cluster-state` to `new` if we get an empty member list from pillar, that approach may create a `initial-cluster-state` `new` for a new node deployment if for what ever reason we are not able to retrieve the list of etcd member.

initial comment https://github.com/scality/metalk8s/pull/2099#discussion_r354703435

**What was expected**:

`initial-cluster-state` should be set to `new` only for the first etcd node (likely the bootstrap one)

TBD:
Suggestion: this value should be `existing` by default and we force to `new` only in the bootstrap script (also make the state raise an error if state=`existing` and we have empty etcd member list)

Ebaneck:
**Component**:

'salt', 'bootstrap'

**Context**: 

See #2102 

**Summary**:

This PR ensures that we set the etcd initial cluster state to `existing` by default and then override this value only during a Bootstrap orchestration.

We also ensure to raise an error when the etcd member list is empty because If all nodes are unavailable we should never set the cluster state to `new`

**Acceptance criteria**: 

Etcd deployment should be resilient.
---

Closes: #2102

### DIFF
--- a/salt/metalk8s/kubernetes/etcd/installed.sls
+++ b/salt/metalk8s/kubernetes/etcd/installed.sls
@@ -12,18 +12,24 @@ include:
 {#- Get the list of existing etcd member. #}
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
-{#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if pillar.get('is_bootstrap') and not etcd_members %}
+  {%- set state = 'new' %}
+  {%- set etcd_endpoints = {node_name: endpoint} %}
+{%- else %}
+  {%- set state = 'existing' %}
 
-{%- set etcd_endpoints = {} %}
-{#- NOTE: Filter out members with empty name as they are not started yet. #}
-{%- for member in etcd_members | selectattr('name') %}
-  {#- NOTE: Only take first peer_urls for endpoint. #}
-  {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
-{%- endfor %}
+  {%- set etcd_endpoints = {} %}
+  {#- NOTE: Filter out members with empty name as they are not started yet. #}
+  {%- for member in etcd_members | selectattr('name') %}
+    {#- NOTE: Only take first peer_urls for endpoint. #}
+    {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
+  {%- else %}
+    {{ raise('List of active etcd members is empty, cannot reference the existing cluster state.') }}
+  {%- endfor %}
 
-{#- Add ourselves to the endpoints. #}
-{%- do etcd_endpoints.update({node_name: endpoint}) %}
+  {#- Add ourselves to the endpoints. #}
+  {%- do etcd_endpoints.update({node_name: endpoint}) %}
+{%- endif %}
 
 {%- set etcd_initial_cluster = [] %}
 {%- for name, ep in etcd_endpoints.items() %}
--- a/salt/metalk8s/orchestrate/bootstrap/init.sls
+++ b/salt/metalk8s/orchestrate/bootstrap/init.sls
@@ -30,6 +30,7 @@
 
 {%- set pillar_data = {
         'bootstrap_id': pillar.bootstrap_id,
+        'is_bootstrap': True,
         'metalk8s': {
             'nodes': {
                 pillar.bootstrap_id: {


salt/metalk8s/defaults.yaml
```diff
@@ -70,6 +70,7 @@ etcd:
     healthcheck_client_signing_policy: etcd_client_policy
     peer_signing_policy: etcd_server_client_policy
     server_signing_policy: etcd_server_client_policy
+  initial_cluster_state: existing
```
gdemonet: Not sure about this... IMO, there's no good reason to let users set this value by hand, hence pillar doesn't seem like a good fit.

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -13,7 +14,15 @@ include:
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
 {#- Compute the initial state according to the existing list of node. #}
```
gdemonet: This comment is somehow outdated, no?

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -13,7 +14,15 @@ include:
...
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if etcd.initial_cluster_state | string() == 'new' %}
+     {%- set state = etcd.initial_cluster_state %}
+{%- else %}
+    {%- if etcd_members is not none %}
+        {%- set state = etcd.initial_cluster_state %}
+    {%- else %}
+        {{ raise ('etcd member list cannot be empty for existing state') }}
+    {%- endif %}
+{%- endif %}
```
gdemonet: Not fond of this approach. If `etcd.initial_cluster_state` is set to a wrong value (e.g. `'toto'`), this will break.

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -13,7 +14,15 @@ include:
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
 {#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if etcd.initial_cluster_state | string() == 'new' %}
+     {%- set state = etcd.initial_cluster_state %}
+{%- else %}
+    {%- if etcd_members is not none %}
```
gdemonet: Can it be `None`? Wouldn't it be an empty list, most of the time?

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -13,7 +14,15 @@ include:
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
 {#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if etcd.initial_cluster_state | string() == 'new' %}
```
gdemonet: I'd rather have something like:

```sls
{%- set initial_cluster_state = 'new' if pillar.during_bootstrap else 'existing' %}    
```

It's much easier to read, and it doesn't require some weird way of marking the initial cluster state in the pillar.
It also allows other components to react to the `during_bootstrap` special case (e.g. apiserver-proxy, which checks if the mine is empty to "know" that we are bootstrapping).

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -13,7 +14,15 @@ include:
...
+     {%- set state = etcd.initial_cluster_state %}
+{%- else %}
+    {%- if etcd_members is not none %}
+        {%- set state = etcd.initial_cluster_state %}
+    {%- else %}
+        {{ raise ('etcd member list cannot be empty for existing state') }}
```
gdemonet: Raising here is a good idea I think, before trying to deploy anything.
You could actually move this to the for loop below:

```sls
{%- for member in etcd_members | selectattr('name') %}
  {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
{%- else %}
  {{ raise('List of active etcd members is empty, cannot reference the existing cluster state.') }}
{%- endfor %}
```

It's a better check, because you'll get rid of not started members in the list.

gdemonet (changes requested):
Logic in `etcd.installed` is good, but I'm not sure about this pillar value... WDYT @TeddyAndrieux @NicolasT ?

salt/metalk8s/kubernetes/etcd/installed.sls
Ebaneck: will fix

salt/metalk8s/kubernetes/etcd/installed.sls
TeddyAndrieux: Yes ok LGTM, having a pillar value for cluster state does not hurt for me but ok with this approach also

salt/metalk8s/kubernetes/etcd/installed.sls
Ebaneck: True, only that during Bootstrap, this value is actually empty and hence the raise will be called always

gdemonet: Add a shortcut ;)

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -12,8 +12,12 @@ include:
 {#- Get the list of existing etcd member. #}
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
-{#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if pillar.is_bootstrap %}
+  {%- set state = 'new' %}
+{%- else %}
+  {%- set state = 'existing' if etcd_members else
+        raise ('etcd member list cannot be empty for existing state')  %}
```
TeddyAndrieux: No really like the fact to raise in the `set`, keep a `if` `else` block please

gdemonet: Just do both at once:

```sls
{%- if pillar.is_bootstrap %}
  {%- set state = 'new' %}
  {%- set etcd_endpoints = {node_name: endpoint} %}
{%- else %}
  {%- set state = 'existing' %}

  {%- set etcd_endpoints = {} %}
  {#- NOTE: Filter out members with empty name as they are not started yet. #}
  {%- for member in etcd_members | selectattr('name') %}
    {#- NOTE: Only take first peer_urls for endpoint. #}
    {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
  {%- else %}
    {{ raise('Br0ken') }}
  {%- endfor %}

  {#- Add ourselves to the endpoints. #}
  {%- do etcd_endpoints.update({node_name: endpoint}) %}
{%- endif %}
```

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -12,18 +12,24 @@ include:
...
+    {#- NOTE: Only take first peer_urls for endpoint. #}
+    {%- do etcd_endpoints.update({member['name']: member['peer_urls'][0]}) %}
+  {%- else %}
+    {{ raise('List of active etcd members is empty, cannot reference the existing cluster state.') }}
+  {%- endfor %}
 
-{#- Add ourselves to the endpoints. #}
-{%- do etcd_endpoints.update({node_name: endpoint}) %}
+  {#- Add ourselves to the endpoints. #}
+  {%- do etcd_endpoints.update({node_name: endpoint}) %}
```
TeddyAndrieux: Not really like this approach, 
For me `state` should be `new` only if we are bootstraping AND we didn't have any etcd members.

Ebaneck: fixed

TeddyAndrieux (approved):
LGTM

salt/metalk8s/kubernetes/etcd/installed.sls
gdemonet: Indeed, that would help in re-running the bootstrap script/orchestration after the cluster was already deployed and expanded. Not sure if there's a use-case, but no good reason for it to break everything either.

gdemonet (dismissed):
All good!

salt/metalk8s/kubernetes/etcd/installed.sls
```diff
@@ -12,18 +12,24 @@ include:
 {#- Get the list of existing etcd member. #}
 {%- set etcd_members = pillar.metalk8s.etcd.members %}
 
-{#- Compute the initial state according to the existing list of node. #}
-{%- set state = "existing" if etcd_members else "new" %}
+{%- if pillar.is_bootstrap and not etcd_members %}
```
TeddyAndrieux:
```suggestion
{%- if pillar..get('is_bootstrap') and not etcd_members %}
```

TeddyAndrieux (dismissed):
Build fail

salt/metalk8s/kubernetes/etcd/installed.sls
gdemonet: Oops! And indeed, there's no sense in always having this key in the pillar.

gdemonet (approved):

PR #2213, opened 2020-01-24
Add extra check for control plane ingress pods and containers

Ebaneck:
**Component**:

'test', 'authentication'

**Context**: 

Authentication tests are flaky.
It looks like when we invoke these tests, pods are ready but the containers are not.

Here is a sample output of pods status vs running container status.

```
tests/post/steps/test_authentication.py:133: Failed
----------------------------- Captured stdout call -----------------------------
NAMESPACE          NAME                                             READY   STATUS    RESTARTS   AGE
metalk8s-ingress   nginx-ingress-control-plane-controller-ptr56     0/1     Running   0          14s
metalk8s-ingress   nginx-ingress-controller-t9gr7                   1/1     Running   0          27s
metalk8s-ingress   nginx-ingress-default-backend-69558976b7-75ftd   1/1     Running   0          51s
```

**Summary**:

Ensure that we check the container status for Nginx-ingress-control plane pods before running authentication tests else we will result in flaky tests.

Increase the delay between a failed tests and a fired retry so that requests lib does not complain of handling an exception while another exception was triggered.

**Acceptance criteria**: 

All authentication tests should pass
Less falky

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/tests/post/steps/test_authentication.py
+++ b/tests/post/steps/test_authentication.py
@@ -81,26 +81,43 @@ def control_plane_ip(host):
 
 # Given {{{
 
+
 @given(parsers.parse(
     "the control-plane Ingress path '{path}' is available"))
-def check_cp_ingress_pod(request, host, k8s_client, control_plane_ip):
+def check_cp_ingress_pod_and_container(
+    request,
+    host,
+    k8s_client,
+    control_plane_ip
+):
     ssh_config = request.config.getoption('--ssh-config')
     label = "release=nginx-ingress-control-plane"
+    namespace = "metalk8s-ingress"
 
-    # Ensure nginx ingress control-plane controller is running
-    def _wait_for_ingress():
-        namespace = "metalk8s-ingress"
-        pods = kube_utils.get_pods(
-            k8s_client, ssh_config, label, namespace=namespace
-        )
-        assert len(pods) > 0
+    def _wait_for_ingress_pod_and_container():
+        try:
+            pods = kube_utils.get_pods(
+                k8s_client, ssh_config, label, namespace=namespace
+            )
+        except Exception as exc:
+            pytest.fail("unable to get pods with error: {}".format(exc))
+
+        assert pods, ("No pod with label {} found".format(label))
+
+        for pod in pods:
+            assert all(
+                container.ready == True
+                for container in pod.status.container_statuses
+            )
+
     utils.retry(
-        _wait_for_ingress,
+        _wait_for_ingress_pod_and_container,
         times=10,
-        wait=3,
+        wait=5,
         name="wait for pod labeled '{}'".format(label)
     )
-    # Todo: use the ingress object to check it's addresses
+    # Todo: check the provided path and ensure it does not redirect to the
+    # default-backend
 
 
 # }}}
@@ -168,7 +185,7 @@ def reach_openid_config(host, control_plane_ip):
             control_plane_ip, INGRESS_PORT
         )
 
-    utils.retry(_get_openID_config, times=10, wait=3)
+    utils.retry(_get_openID_config, times=10, wait=5)
 
 
 @then(parsers.parse(
--- a/tox.ini
+++ b/tox.ini
@@ -130,6 +130,7 @@ markers =
     ingress: tag a BDD feature as related to ingress
     volume: tag a BDD feature as related to Volume management
     bootstrap: tag a BDD feature as related to bootstrap
+    authentication: tag a BDD feature as related to authentication
 filterwarnings =
     ignore:encode_point has been deprecated on EllipticCurvePublicNumbers and will be removed in a future version. Please use EllipticCurvePublicKey.public_bytes to obtain both compressed and uncompressed point encoding.:UserWarning
     ignore:Support for unsafe construction of public numbers from encoded data will be removed in a future version. Please use EllipticCurvePublicKey.from_encoded_point:UserWarning


tests/post/steps/test_authentication.py
```diff
@@ -81,6 +81,50 @@ def control_plane_ip(host):
 
 # Given {{{
 
+
+@given(parsers.parse(
```
alexandre-allard, line 85: `parsers.parse` is not needed if there is nothing to parse :)

tests/post/steps/test_authentication.py
```diff
@@ -81,6 +81,50 @@ def control_plane_ip(host):
...
+    "the control-plane Ingress container is Ready"))
+def check_cp_ingress_containers(request, host, k8s_client):
```
alexandre-allard: Can't you use the same `given` you use for `And pods with label 'app.kubernetes.io/name=dex' are 'Ready'` ? To me, it looks like the same.
```
And pods with label 'release=nginx-ingress-control-plane' are 'Ready'
```
edit: nvm, didn't see that you check the container status here, not pod one, I though it was the same (so a pod can be ready and not the containers inside ?).
Still, this function could be variabilized (is that even a word ?) to do a `check_pod_containers_status`, rather than this very specific use case.

tests/post/features/authentication.feature
```diff
@@ -11,6 +11,7 @@ Feature: Authentication is up and running
 
     Scenario: Reach the OpenID Config
         Given the Kubernetes API is available
+        And the control-plane Ingress container is Ready
```
alexandre-allard: Shouldn't we add this line in each and every scenarios below ?
Maybe, we can even add a `Background` at the beginning of the feature, since we always need the same things:
```
    Background:
        And the control-plane Ingress container is Ready
        And the control-plane Ingress path '/oidc' is available
        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'

    Scenario: Reach the OpenID Config
        Then we can reach the OIDC openID configuration

    Scenario: Access HTTPS service
        When we perform a request on '/oidc/' with port '8443' on control-plane IP
        Then the server returns '404' with message '404 page not found'

    Scenario: Login to Dex using incorrect email
        When we login to Dex as 'admin@metalk8s.com' using password 'password'
        Then authentication fails with login error

    Scenario: Login to Dex using correct email and password
        When we login to Dex as 'admin@metalk8s.invalid' using password 'password'
        Then the server returns '303' with an ID token
```
see: https://pytest-bdd.readthedocs.io/en/latest/#backgrounds

Ebaneck: Agreed that we could improve the entire tests suite and not limited to authentication by adding `Background`  scenarios.

Worth creating an issue to track this refactoring.
I have added the line to each scenario for now

tests/post/steps/test_authentication.py
Ebaneck: The extra check for container status has been added to the same function that checks ingress pods.
so now the function has the following submethods
- check ingress pod status
- obtain the full name of ingress pods
- then finally check the status of containers in that pod

tests/post/features/authentication.feature
Ebaneck: So instead of adding the `Given` to each line, I merged the check for container status in the same function that checks for pod status using submethods

tests/post/steps/test_authentication.py
```diff
@@ -81,25 +81,64 @@ def control_plane_ip(host):
 
 # Given {{{
 
+
 @given(parsers.parse(
     "the control-plane Ingress path '{path}' is available"))
```
alexandre-allard, line 86: What this `path` is used for ? Does not seem to be used anywhere.

Ebaneck: True... here is why: https://github.com/scality/metalk8s/pull/2213/files#diff-6065c212c4eed6c66088271ca7d2eddfR142

tests/post/features/authentication.feature
gdemonet: Let's create this tracking issue then @Ebaneck

tests/post/features/authentication.feature
```diff
@@ -12,26 +12,30 @@ Feature: Authentication is up and running
     Scenario: Reach the OpenID Config
         Given the Kubernetes API is available
         And the control-plane Ingress path '/oidc' is available
```
gdemonet: If I may suggest something, we have a "Ingress-specific `Given`", which is supposed to make sure the specified path is available. This check could perform all the required checks to move forward, including the container check:

- First, check that the `nginx-ingress-control-plane` DaemonSet is OK (there's a Pod running on all `master` Nodes, or maybe at least one, not sure)
- Then, check that for each Pod belonging in this DaemonSet, all of its containers are Ready
- Then, through one of the available control plane IPs (in our case, there's only the Bootstrap's that should work), try to connect to `https://<cp-ip>:8443/<path>`, with `<path>` being parsed in the `Given`

If all of these succeed (the last check should just validate that we do not receive a 502 or a 404 from the default backend), then any following query should work. And there's no need to _know_ in advance which Pod make up the service that the Ingress object is pointing to.

tests/post/steps/test_authentication.py
```diff
@@ -254,4 +287,16 @@ def _dex_auth_request(control_plane_ip, username, password):
...
+def _get_pod_name(k8s_client, ssh_config, label, namespace):
+    try:
+        pods = kube_utils.get_pods(
+            k8s_client, ssh_config, label, namespace=namespace
+        )
+    except Exception as exc:
+        pytest.fail("Unable to get pod name with error: {}".format(exc))
+
+    for pod in pods:
+        return pod.metadata.name
```
gdemonet: Not fond of this. It's misleading, as you're getting "the name of the first Pod that matched this query", not "the Pod name". Why not check all Pods?

tox.ini
```diff
@@ -130,6 +130,7 @@ markers =
     ingress: tag a BDD feature as related to ingress
     volume: tag a BDD feature as related to Volume management
     bootstrap: tag a BDD feature as related to bootstrap
+    authentication: tag a BDD feature as related to authentication
```
gdemonet, line 133: Good point, however this should have its dedicated commit (not related to your changes above).

gdemonet (changes requested):
The added checks are OK but could benefit from some clean-up, at least to make the flow obvious for readers. You could open a follow-up PR, since this one is supposed to reduce flakiness.

tests/post/features/authentication.feature
Ebaneck: I think point 1 & 2 have been addressed in this PR maybe not word for word.

I am going to open a second PR to address point 3 separately

tests/post/steps/test_authentication.py
```diff
@@ -81,26 +81,45 @@ def control_plane_ip(host):
...
+    def _wait_for_ingress_pod_and_container():
+        try:
+            pods = kube_utils.get_pods(
+                k8s_client, ssh_config, label, namespace=namespace
+            )
+        except Exception as exc:
+            pytest.fail("unable to get pods with error: {}".format(exc))
+
+        assert len(pods) > 0, (
+            "Expected only one Pod with label {}, but found {}"
```
alexandre-allard: For me, the assertion and the message are not consistent, either we have:
```
assert pods, "No Pod with label {} found".format(label)
```
Or:
```
assert len(pods) == 1, (
    "Expected only one Pod with label {}, but found {}"
).format(label, len(pods))
```

alexandre-allard (dismissed):

tests/post/steps/test_authentication.py
gdemonet: Unfortunately, you should have picked the first option. We can't know how many Pods we'll find, as it depends on the number of `master` Nodes in the cluster.

gdemonet (dismissed):
LGTM!

alexandre-allard (dismissed):

alexandre-allard (approved):

Issue #2220, opened 2020-02-05
Build of salt-master image is broken


alexandre-allard:
**Component**: build

**What happened**:
At build time, the salt-master image build is failing

**What was expected**:
Build went successfully and the salt-master image is present. 

**Steps to reproduce**:
Remove the salt-master image from your local cache, then try to build the images
```
./doit.sh images
```
You can also trigger a build in the CI.

**Resolution proposal** (optional):

alexandre-allard:
Looks like the issue is when we try to install etcd3 with pip:
```
#  pip install "etcd3 != 0.11.0"
Collecting etcd3!=0.11.0
  Downloading https://files.pythonhosted.org/packages/30/1b/6402bb034afd85a33f3fe0ea064a496f3f96917e21ddefbe70c89735ccbf/etcd3-0.11.1.tar.gz (63kB)
    100% |################################| 71kB 1.9MB/s
Collecting grpcio>=1.24.0 (from etcd3!=0.11.0)
  Downloading https://files.pythonhosted.org/packages/c3/a1/89b7905c9540ad2eac024630abece06d2d563b5fb06e10f6597dcd33972f/grpcio-1.27.0.tar.gz (16.9MB)
    100% |################################| 16.9MB 72kB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-glvgiO/grpcio/setup.py", line 191, in <module>
        if check_linker_need_libatomic():
      File "/tmp/pip-build-glvgiO/grpcio/setup.py", line 152, in check_linker_need_libatomic
        stderr=PIPE)
      File "/usr/lib64/python2.7/subprocess.py", line 711, in __init__
        errread, errwrite)
      File "/usr/lib64/python2.7/subprocess.py", line 1327, in _execute_child
        raise child_exception
    OSError: [Errno 2] No such file or directory

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-glvgiO/grpcio/
```

NicolasT:
This seems fixed?

slaperche-scality:
It is.
The PR was merged and was referencing the issue correctly (but it haven't been auto-closed though…)

PR #2221, opened 2020-02-05
build: upgrade pip version prior to install package

alexandre-allard:
**Component**: build

**Context**: 
The pip version shipped in this dist is very old 8.1.2 (2016-05-10), which prevented us from using it to install etcd client and its dependencies.

**Summary**:
We now use the latest pip version available.

**Acceptance criteria**: we can build salt-master image 

---

Closes: #2220

### DIFF
--- a/images/salt-master/Dockerfile
+++ b/images/salt-master/Dockerfile
@@ -26,6 +26,7 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
  && yum install -y python2-kubernetes salt-master salt-api salt-ssh openssh-clients \
  && yum install -y python-pip \
  && yum clean all \
+ && pip install pip==20.0.2 \
  && pip install etcd3 \
  && chmod +x /tini
 


images/salt-master/Dockerfile
```diff
@@ -26,6 +26,7 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
  && yum install -y python2-kubernetes salt-master salt-api salt-ssh openssh-clients \
  && yum install -y python-pip \
  && yum clean all \
+ && pip install -U pip \
```
alexandre-allard: Maybe I should freeze the pip version to the latest version available now (20.0.2), rather than doing that.

slaperche-scality: Yup, better freeze the version (more reproducible build, and if it fails again one day we can bump it).

gdemonet: Maybe, though probably much less critical than freezing `etcd3` just below, since we rely on it at runtime.

gdemonet (dismissed):

alexandre-allard:
I will need a by pass of the build status as we can't build previous branches without the fix.

slaperche-scality:
Not the first time this happen, that may point a flaw in our design here.

gdemonet:
> we can't build previous branches without the fix.

This will happen again and again. We need to stop doing this ASAP - can you open a ticket @alexandre-allard-scality?

gdemonet (approved):

slaperche-scality (approved):

Ebaneck:
> I will need a by pass of the build status as we can't build previous branches without the fix.

You might also require queue bypass from Releng

alexandre-allard:
@gdemonet created #2222

PR #2224, opened 2020-02-06
Fix doit clean

### LINKED ISSUE #2223: ./doit.sh clean doesn't work
**Component**:

buildchain

**What happened**:

When I run `./doit.sh clean`, `_build` is not removed (some files & directories are still lingering there).

**What was expected**:

The `_build` is deleted.

**Steps to reproduce**

Run `./doit.sh clean` after a `./doit.sh`.

**Resolution proposal**:

There are several things to fix here:
- local images that we don't save in the build tree create a witness file which isn't cleaned up => this witness file is no longer required, thus we can remove the code that creates it
- We are now building the doc inside the build tree and Sphinx creates a `doctrees` directory, being created by Sphinx it's not tracked (and thus cleaned up) by the buildchain => use a custom clean action to handle it
- race conditions when cleaning images: sometimes the `metalk8s-ui` image is deleted **AFTER** its parent directory: this is caused by a missing deps => simply fix the deps (same things for `_ui_build` and the `UI_BUILD_ROOT`).
- double delete of `REPO_DEB_ROO` (both by `_package_mkdir_deb_iso_root` and `_download_deb_packages`, but the latter doesn't handle the fact that it's already deleted) => only `_package_mkdir_deb_iso_root` should delete it (it's the owner, the one which create the directory, and thus the one responsible for its deletion)
- cannot cleanup `UI_BUILD_ROOT` because of UID and permissions: this happens to work fine locally by chance, but not in the CI

Also, to avoid future silent breakage of `./doit.sh clean` I suggest that we test it in the CI (that's super easy).

slaperche-scality:
**Component**:

buildchain

**Context**: 

`./doit.sh clean` doesn't work (i.e. the `_build` directory still exists and isn't empty).

**Summary**:

- fix task to build local images (don't `touch` a file on disk)
- fix race conditions due to missing dependencies
- fix doc task cleanup (handle garbage from Sphinx)
- fix wrong usage of `setdefault` for `task_dep` (this one could be backported to 2.0 I believe)
- fix "double-delete" of a directory
- fix `_ui_build` cleanup (was tricky because of the `node` user…)
- bonus: simplify code of `Mkdir` task
- add check for `./doit.sh clean` in the CI

**Acceptance criteria**: 

Green build in CI (we now ensure that `./doit.sh clean` works in the CI)

---

Closes: #2223

### DIFF
--- a/buildchain/buildchain/builder.py
+++ b/buildchain/buildchain/builder.py
@@ -34,7 +34,7 @@ def _builder_image(
         They are passed to `LocalImage` init method.
     """
     img_name = '{}-{}-builder'.format(config.PROJECT_NAME.lower(), name)
-    kwargs.setdefault('task_dep', ['_build_root'])
+    kwargs.setdefault('task_dep', []).append('_build_root')
 
     return LocalImage(
         name=img_name,
--- a/buildchain/buildchain/docs.py
+++ b/buildchain/buildchain/docs.py
@@ -70,9 +70,14 @@ def task__doc_mkdir_iso_root() -> types.TaskDict:
 
 def task__doc_mkdir_build_root() -> types.TaskDict:
     """Create the documentation build root directory."""
-    return targets.Mkdir(
+    def clean_doctrees() -> None:
+        coreutils.rm_rf(constants.DOCS_BUILD_ROOT/'doctrees')
+
+    task = targets.Mkdir(
         directory=constants.DOCS_BUILD_ROOT, task_dep=['_build_root']
     ).task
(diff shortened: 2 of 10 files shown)


gdemonet (approved):
LGTM! May make sense to backport in 2.0 one day, not urgent though.

slaperche-scality:
> LGTM! May make sense to backport in 2.0 one day, not urgent though.

Most of the stuff was introduced in 2.4, but yeah some `setdefault` misused existed in 2.0 so may be worth to backport if we have the occasion.

alexandre-allard (approved):

PR #2225, opened 2020-02-07
ci: stop using OpenStack server groups in terraform

alexandre-allard:
**Component**: ci, tf

**Context**: 

**Summary**: removal of server groups in terraform

**Acceptance criteria**:  builds using terraform still work :)

### DIFF
--- a/eve/workers/openstack-multiple-nodes/terraform/nodes.tf
+++ b/eve/workers/openstack-multiple-nodes/terraform/nodes.tf
@@ -1,17 +1,8 @@
-resource "openstack_compute_servergroup_v2" "all" {
-  name     = "${local.prefix}-servergroup"
-  policies = []
-}
-
 resource "openstack_compute_instance_v2" "bastion" {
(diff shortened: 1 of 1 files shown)


eve/workers/openstack-multiple-nodes/terraform/nodes.tf
```diff
@@ -1,18 +1,9 @@
-resource "openstack_compute_servergroup_v2" "all" {
```
slaperche-scality, line 1: Commit message could mention **WHY** we remove them.

And as they seems useless (as we just remove them and be done with it), maybe explain what was their role/why it's not an issue to simply remove them.

alexandre-allard: Yep, good point, I wanted to see first if it does not break anything in the CI, but it looks good.
Is the commit message sufficient or do you think we must put more information in it ?

slaperche-scality (dismissed):

gdemonet (approved):

PR #2226, opened 2020-02-07
docs: add `release` tag

NicolasT:
This commit adds a `release` tag to the Sphinx project. Currently, when
set, `todo` elements in the documentation will *not* be rendered. This
is, however, not the default.

A related change is made to pass the `O` environment variable from the
`tox` invocation all the way down to arguments to `sphinx-build`. As
such, one can now run

```
O="-t release" tox -e docs
```

to create 'release' builds of the documentation.

### DIFF
--- a/buildchain/buildchain/docs.py
+++ b/buildchain/buildchain/docs.py
@@ -116,6 +116,9 @@ def task_doc() -> Iterator[types.TaskDict]:
             command=['/entrypoint.sh', target.command],
             builder=builder.DOC_BUILDER,
             run_config=run_config,
+            environment={
+                'O': '-t release',
+            },
             mounts=[
                 utils.bind_mount(
                     target=Path('/usr/src/metalk8s/docs/_build/'),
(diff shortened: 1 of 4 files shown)


lucieleonard (dismissed):

lucieleonard (commented):
I would add the new command in README.md

slaperche-scality:
Why not have a release config by default and add a `dev` flag to enable TODOs and stuff?

Less risky if we generate the doc in a hurry and forget the flag, at least it'll be "clean" by default.

NicolasT:
> Why not have a release config by default and add a `dev` flag to enable TODOs and stuff?
> 
> Less risky if we generate the doc in a hurry and forget the flag, at least it'll be "clean" by default.

I'd like all non-customer builds (i.e., ReadTheDocs, open-source builds,...) to include these TODOs since we welcome contributions to fix these.

slaperche-scality:
Ha, we have no control on how RTD build the doc (i.e. can't switch the flag for RTD)?

NicolasT:
> Ha, we have no control on how RTD build the doc (i.e. can't switch the flag for RTD)?

Maybe we can (didn't check), but even if we can, I'd like any default build to still include the tags.

However. This does raise the question about the docs we include in a build (HTML in the UI, or PDF)...

NicolasT:
I added a patch which ensures the docs are built using the `release` tag while building the ISO. Manual testing seems to validate this is working.

slaperche-scality (dismissed):

slaperche-scality (approved):

PR #2239, opened 2020-02-13
ci: Upload final_status at the artifacts root

gdemonet:
We used to upload all status files in a subdirectory. The "root" status
(for the whole build execution) should however be stored at the root of
the artifacts directory, to ensure artifacts promotion will work
properly.


bertranddemiddelaer (approved):
The .final_status is now also present at the build root. Thanks !

alexandre-allard (approved):

Issue #2232, opened 2020-02-12
Add a software requirements spec document for user customization and persistence


Ebaneck:
**Component**:

'Docs'

**Why this is needed**:

Our new approach warrants that for each new feature being developed, we provide a good, well written and validated requirements document.

**What should be done**:

**Implementation proposal** (strongly recommended):

**Test plan**:

Epic: #1996

PR #2240, opened 2020-02-13
Cluster and Services Configuration: add Requirements document

thomasdanan:
First list of requirements for the topic of 'Cluster and Services Configurations & Persistence'.

This focuses on high-level requirements, and provides a first list of settings a cluster administrator should be able to configure.

---
Closes: #2232
Epic: #1996

### DIFF
--- a/docs/developer/architecture/configurations.rst
+++ b/docs/developer/architecture/configurations.rst
@@ -1,0 +1,75 @@
+Cluster and Services Configurations and Persistence
+===================================================
+
+Context
+-------
+
+MetalK8s comes with a set of tools and services that may need to be configured
+on site. At the same time, we don't want the administrator of the cluster to
+master each and every service of the cluster. We also don't want to allow all
+kind of configurations since it would make the system even more complex to test
+and maintain over time.
+
+In addition to those services, MetalK8s deployment may have to be adapted
+depending on the architecture of the platform or depending on the different
+use cases and applications running on top of it.
+
+It can be:
+
+- The BootstrapConfig,
+- The various roles and taints we set on the node objects of the cluster
+- The configurations associated to solutions, such as the list of available
+  solutions, the environments and namespaces created for a solution
+
+Be it services or MetalK8s configurations, we need to ensure it is persisted
+and resilient to various type of events such as node reboot, upgrade,
+downgrade, backup, restore.
+
+.. _configurations-requirements:
+
+Requirements
+------------
+
+User Stories
+^^^^^^^^^^^^
+
+Available Settings
+~~~~~~~~~~~~~~~~~~
+As a cluster administrator, I have access to a finite list of settings I can
+customize on-site in order to match with my environment specificities:
+
+- List of static users and credentials configured in Dex
+- Integration with an external IDP configuration in Dex
+- Existing Prometheus rules edition and new rules addition
+- Alert notifications configuration in Alert Manager
+- New Grafana dashboards or new Grafana datasources
+- Number of replicas for the Prometheus, Alert Manager, Grafana or Dex
+  deployments
+
+.. note::
+
+   Other items will appear as we add new configurable features in MetalK8s
+
+Settings Documentation
+~~~~~~~~~~~~~~~~~~~~~~
+As a cluster administrator, I have access to a documented list of settings I
+can configure in the Operational Guide.
+
+Persistence of Configurations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+As a cluster administrator, I can upgrade or downgrade my cluster without
+losing any of the customised settings described above.
+
+Backup and Restoration
+~~~~~~~~~~~~~~~~~~~~~~
+As a cluster administrator, when I am doing a backup of my cluster, I backup
+all the customised settings described above and I can restore it when restoring
+the MetalK8s cluster or I can re apply part or all of it on a fresh new
+cluster.
+
+Expert-mode Access
+~~~~~~~~~~~~~~~~~~
+As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
+settings that are exposed. The intent is to have a method / API that an expert
+could use, if the right CLI tool or GUI is not available or not functioning as
+expected.
--- a/docs/developer/architecture/index.rst
+++ b/docs/developer/architecture/index.rst
@@ -6,4 +6,5 @@ Architecture Documents
    authentication
    deployment
    monitoring
+   configurations
    requirements


docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
+User Customisations and Persistence
+===================================
+
+Context
+-------
+
+MetalK8s comes with a set of tools and services that may need to be configured
+on site. At the same time, we don't want the administrator of the cluster to
+master each and every service of the cluster. We also don't want to allow any
```
gdemonet: We don't want to allow **all** kinds of customisations (_not any_ means nothing)

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+kind of customisations since it would make the system even more complex to test
+and maintain over the time.
```
gdemonet: over ~~the~~ time

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+
+In addition to those services, MetalK8s deployment may have to be adapted
```
gdemonet: I'd like more details about what you call "MetalK8s configurations"... Something else than what we have in BootstrapConfig + Node objects?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+Requirements
+------------
+
+As a MetalK8s administrator, I have access to a finite list of settings I can
+customise on-site in order to match with my environment specificities:
+
+- list of static users and credentials configured in Dex
+- Integration with an external IDP configuration in Dex
+- Alert notification configuration in Alert Manager
+- Number of replicas for the Prometheus and Alert Manager cluster
```
gdemonet: Grafana could also get more replicas, no?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+- Other items will appear as we add new customisable features in MetalK8s
+
+Depending on the the setting, impact of the customisation can be at cluster
```
gdemonet: Depending on the ~~the~~ setting

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+level or namespace level (not sure it is needed)
```
gdemonet: None of the use cases above require cluster-level config, but if you have ideas where this would be required, we'll need to take it into account for the design (which uses namespace-scoped ConfigMaps at the moment).

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+
+As a MetalK8s administrator, I have a uniq CLI/API way of doing MetalK8s
```
gdemonet: Can we specify what "way of doing" means? Is it a single entrypoint (`kubectl edit cm ...`, or `kubectl metalk8s userconfig ...` if we have our CLI) ? A single customisation format? Should it be even a single "source of config", like a single file?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+customisation, whatever the type of customisation is.
+From a UI, I could have different type of forms in order to make the
```
gdemonet: This could/should also be achieved with CLI in my opinion

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+
+As a MetalK8s administrator, I have access to a documented list of settings I
+can configure in the MetalK8s Operation section.
+
+As a MetalK8s administrator, I can upgrade or downgrade my cluster without
+losing any of the customised settings described above.
+
+As a MetalK8s administrator, when I am doing a backup of my cluster, I backup
+all the customised settings described above and I can restore it when restoring
+the MetalK8s cluster?
```
gdemonet: Why the question mark?

gdemonet (commented):
Some comments, looks good to me overall

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,51 @@
...
+configuration easier depending on the type of setting.
+
+As a MetalK8s administrator, I have access to a documented list of settings I
+can configure in the MetalK8s Operation section.
+
+As a MetalK8s administrator, I can upgrade or downgrade my cluster without
+losing any of the customised settings described above.
+
+As a MetalK8s administrator, when I am doing a backup of my cluster, I backup
+all the customised settings described above and I can restore it when restoring
```
TeddyAndrieux: > I backup all the customised settings

No, we don't backup anything about user custos, but they are in etcd so even if we lose node they are still available in the etcd cluster. 
For me this section should just be removed

docs/developer/architecture/user_customisations.rst
thomasdanan: done

docs/developer/architecture/user_customisations.rst
thomasdanan: done

docs/developer/architecture/user_customisations.rst
thomasdanan: So I was mainly referring to the number of replicas for the MetalK8s services with deploy such as Prometheus or AlertManager. It is true that content of BootstrapConfig or NodeObjects config (roles and taints) as well as solution configs (what solution is available on the bootstrap or envs or namespaces created when deploying solutions) are also part of MetalK8s settings we need to expose to the MetalK8s administrator and that need to persist. Will update this part to be more precise.

docs/developer/architecture/user_customisations.rst
thomasdanan: Yes I can add it but should I also talk about grafana configurations? (dashboards, datasources ...)?

docs/developer/architecture/user_customisations.rst
thomasdanan: done

docs/developer/architecture/user_customisations.rst
thomasdanan: Dex configuration is not a cluster-level config ? or actually maybe Dex deployment is part of a specific namespace such as prometheus or alert manager so having all custos at namespace level is probably enough ?

docs/developer/architecture/user_customisations.rst
thomasdanan: So I was mainly thinking of common way of updating one setting using `kubectl edit cm` indeed. will update. Does not mean it should be the only way but at least we always have this backing door

docs/developer/architecture/user_customisations.rst
thomasdanan: Yes indeed, I think bottom line we have the `kubectl edit cm` backing door for an IT Specialist or Scality expert to update all exposed settings and we have a set of distinct CLI commands and UI, for an IT Generalist to perform different type of configurations.

docs/developer/architecture/user_customisations.rst
thomasdanan: to make sure the reviewer catches everything !!! done ...

docs/developer/architecture/user_customisations.rst
thomasdanan: My point is to make sure that if we lose the cluster or etcd we have a backup of user custos that we can restore. So for everything in etcd we have the etcd backup but what about BootstrapConfig, Node Objects config, Solution / Environment configs etc ... If everything is already implemented this is great but the requirement remains valid no ?

docs/developer/architecture/user_customisations.rst
thomasdanan: done

docs/developer/architecture/user_customisations.rst
TeddyAndrieux: In the backup we have all `/etc/metalk8s` directory and etcd snapshot but currently we only do backup at the end of bootstrap and restore.

The think is that currently we do not have automatic backup. We only backup at the end of the bootstrap and at the end of restoration but that it. That mean if a user do a custos in etcd or bootstrap config, or whatever, after bootstrap (which will always be the case) that will not be in the backup.

Anyway we should often backup.

If you restore with an etcd cluster still alive then you will not lose any tunning in the etcd stuff.
But for all tuning in bootstrap file AND/OR if you lose all etcd cluster we can't guarantee that your backup archive is uptodate with all the tuning you did.

Anyway I'm ok with this requirement just for me the work that need to be done on backuo/restore is not related to this EPIC

thomasdanan: @TeddyAndrieux But we can always launch a backup manually at any time right?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,59 @@
...
+
+As a MetalK8s administrator, I have access to a finite list of settings I can
+customise on-site in order to match with my environment specificities:
+
+- list of static users and credentials configured in Dex
+- Integration with an external IDP configuration in Dex
+- Alert notification configuration in Alert Manager
+- Number of replicas for the Prometheus, Alert Manager, Grafana or Dex
+- Other items will appear as we add new customisable features in MetalK8s
+
```
ChengYanJin: It's a question from UI part.
- Currently, we store the language preference in local storage.
- We are going to support light/dark theme switch very sson.

Wondering do we need to take these settings(language/theme) into consideration?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,59 @@
...
+In addition to those services, MetalK8s deployment may have to be adapted
+depending on the architecture of the platform or depending on the different
+use cases and applications running on the top of it.
+
+It can be:
+
+- The BootstrapConfig,
+- The various roles and taints we set on the node objects of the cluster
+- The labels and annotations we would set on all K8s objects
```
gdemonet: This is too much free form, and contradicts the sentence `We also don't want to allow all kind of customisations`. We cannot allow users to add whichever labels/annotations on any K8s object **we** deploy - it may break the deployment itself. I also don't see why they would like to do so on our K8s objects. Whatever isn't deployed by MetalK8s is not in scope for this feature.

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,59 @@
...
+- The configurations associated to solutions, such as the list of available
+  solutions, the environments and namespaces created for a solution
+
+Be it services or MetalK8s customisations, we need to ensure it is persisted
+and resistant to various type of events such as node reboot, upgrade, downgrade
```
gdemonet:
```suggestion
and resilient to various type of events such as node reboot, upgrade, downgrade
```

gdemonet (dismissed):

docs/developer/architecture/user_customisations.rst
gdemonet: Requirement is valid and we should keep it. It is a good reason to choose `etcd` as a storage media, since we'll need to back it up if we want to restore after complete disaster (everything lost).
Although we speak a lot about this kind of disaster recovery, I'm curious if this is a real need, and if it is, what's the urgency. Losing a cluster is a lot more problems than just a loss of small configurations, and nothing we can manage on MetalK8s' side.

docs/developer/architecture/user_customisations.rst
gdemonet: Good question. I think we're not talking about the same topics however, so no, shouldn't be in here.
Calling this feature "User customizations" is actually misleading: there's nothing related to a user here, we're talking about "Cluster services configuration".
In the UI case however, this is indeed "User preferences". It could warrant a dedicated architecture document (cc @thomasdanan), but do you think `LocalStorage` isn't a good design for it @ChengYanJin?

docs/developer/architecture/user_customisations.rst
gdemonet: > No, we don't backup anything about user custos, but they are in etcd so even if we lose node they are still available in the etcd cluster.
> For me this section should just be removed

Don't use the design to adjust the requirements.

docs/developer/architecture/user_customisations.rst
gdemonet: Everything is cluster-level config, in a sense that we won't have different configurations of the same service in different places. But configuration will always be tied to the service itself, which sits in a namespace managed by MetalK8s.
Also, this makes the name of this feature really misleading: there's nothing related to "users" here, except that it's a user who will do the configuration.

docs/developer/architecture/user_customisations.rst
gdemonet: Yes you could, if you think it's something we should support.

docs/developer/architecture/user_customisations.rst
ChengYanJin: Yeah. it's good to store the settings in Localstorage!
Maybe we will want to store the "user preference" during the upgrade as well?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,59 @@
...
+customise on-site in order to match with my environment specificities:
+
+- list of static users and credentials configured in Dex
+- Integration with an external IDP configuration in Dex
+- Alert notification configuration in Alert Manager
+- Number of replicas for the Prometheus, Alert Manager, Grafana or Dex
+- Other items will appear as we add new customisable features in MetalK8s
+
+Depending on the setting, impact of the customisation can be at cluster
+level or namespace level (not sure it is needed).
```
xaf-scality: For me there is no customisation for a namespace, it's more on the solution level if needed

nloewensen (commented):
From a high-level point of view we should be able to keep any customization across reboots/upgrades etc. and potentially even be able to restore them to a newly installed system, so that they do not need to be applied again manually (site-disaster).
Hence some kind of backup procedure is needed and potentially we should allow to actually copy this backup to a location “outside” of the MetalK8s cluster in order to allow a restorability in case the entire clusters is “lost”. In addition, we need a restore procedure that will allow to restore settings. Ideally with a selection option for all or "specific" items.
The procedures need to be fully documented and the respective tooling provided.

docs/developer/architecture/user_customisations.rst
thomasdanan: Okay so I remove the part about labels and annotations but if we don't want to allow the administrator setting such labels or annotation, maybe we should make sure he cannot do so on our K8s objects?

docs/developer/architecture/user_customisations.rst
thomasdanan: done

docs/developer/architecture/user_customisations.rst
thomasdanan: ok adding it

docs/developer/architecture/user_customisations.rst
thomasdanan: would you agree with MetalK8s customizations or cluster customizations?

docs/developer/architecture/user_customisations.rst
```diff
@@ -0,0 +1,60 @@
...
+Requirements
+------------
+
+As a MetalK8s administrator, I have access to a finite list of settings I can
+customise on-site in order to match with my environment specificities:
+
+- list of static users and credentials configured in Dex
+- Integration with an external IDP configuration in Dex
+- Alert notification configuration in Alert Manager
+- New Grafana dashboards or new Grafana datasources
```
Ebaneck: @NicolasT @gdemonet Should the persistence of Grafana dashboard be handled like this? From the Prometheus operator charts, it looks like the production recommended way was to use volumes and PVC's just like we did for Alertmanager & Prometheus.

gdemonet: The requirement exists however. It may imply that the design also includes "exceptions" (e.g. provide a PV for Grafana, although I'm not sure how that fits with dashboards configured in ConfigMaps...), or simply acknowledges that this requirement is not handled (which could obviously be dismissed as not being acceptable).

docs/developer/architecture/user_customisations.rst
gdemonet: Agreed, let's remove this sentence.

docs/developer/architecture/user_customisations.rst
bkettler: I'd also like users to be able to add their own alerts to Prometheus. Differing customer workloads often surface new/unknown bottlenecks, issues, etc that we may need to monitor. Users or the CSE should be able to easily and persistently edit the alerts.

gdemonet: Good point @bkettler, will add this to the list of topics.

docs/developer/architecture/user_customisations.rst
gdemonet: Making sure they can't do stuff we don't document would mostly be managed through RBAC. Though I don't see how we can prevent a cluster admin from doing whatever he needs/wants.

docs/developer/architecture/user_customisations.rst
gdemonet: Went with `Cluster and Services Configurations and Persistence`.

gdemonet (approved):

Issue #2233, opened 2020-02-12
Add a design document for User customization and Persistence


Ebaneck:
**Component**:

'docs'

**Why this is needed**:

Our new approach warrants that for each new feature being developed, we provide a good, well written and validated design document.

This will be a follow up to the requirements document from #2232 

**What should be done**:

**Implementation proposal** (strongly recommended):

**Test plan**:

Epic: #1996

Issue #2218, opened 2020-02-04
On Site Grafana Dashboard and configuration Persistence


thomasdanan:
**Component**:

'grafana'

**Why this is needed**:

Even if MetalK8s is coming with a bunch of Grafana dashboards, IT Specialist might need to create its own data sources and dashboards on site. If he is doing so, he wants those dashboards to remains available when servers are restarted and when MetalK8s is upgraded. Not talking about the dashboards we deliver as part of MetalK8s but only the ones IT Specialist would create on site. Same apply to specific Grafana configurations if any

**What should be done**:

Make sure that when new Grafana data source and dashboard is created, this data source or dashboard remains available after reboot or upgrade.

**Implementation proposal** (strongly recommended):

**Test plan**:

Issue #2163, opened 2019-12-26
Centralised CLI for low level administration


thomasdanan:
**Component**:

'cli'

**Why this is needed**:

The number of tools, scripts may increase with the time and in order to make it easy for the IT Specialist, it would make sense to have one single CLI tool, with reasonable help and documentation, that can be used to perform most common low-level administrative tasks such as:

- Upgrade/Downgrade
- Backup/Restore
- Expansion 
- removing a Node (whatever roles are set to it)
- storage management
- more to be defined

**What should be done**:

kubectl plugin
Specific MetalK8s operation tooling

**Implementation proposal** (strongly recommended):

**Test plan**:

NicolasT:
References:
https://github.com/scality/metalk8s/issues/1854
https://github.com/scality/metalk8s/issues/1852
https://github.com/scality/metalk8s/issues/1588
https://github.com/scality/metalk8s/pull/1912
https://github.com/scality/metalk8s/pull/1065 (prior art)

PR #2241, opened 2020-02-13
Design document for user customization and persistence

Ebaneck:
**Component**:

'docs', 'design'

**Context**: 

see #2233 

**Summary**:

This document explains some of the design choices considered while
designing a simplistic tool for MetalK8s that guarantees that user-defined
configurations are persisted throughout.

**Acceptance criteria**: 

- Most be validated by Architect and Team Lead

---

Closes: #2233

### DIFF
--- a/docs/developer/architecture/configurations.rst
+++ b/docs/developer/architecture/configurations.rst
@@ -73,3 +73,204 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
 settings that are exposed. The intent is to have a method / API that an expert
 could use, if the right CLI tool or GUI is not available or not functioning as
 expected.
+
+Design Choices
+--------------
+
+:term:`ConfigMap` is chosen as a unified data access and storage media for
+cluster and service configurations in a MetalK8s cluster based on the above
+requirements for the following reasons:
+
+* Ability to support Update operations on ConfigMaps with CLI and UI easily
+  using our already existing python kubernetes module.
+* Guarantee of adaptability and ease of changing the design and implementation
+  in cases where customer needs evolve rapidly.
+* ConfigMaps are stored in the :term:`etcd` database which is generally being
+  backed up. This ensures that user settings cannot be lost easily.
+
+How it works
+^^^^^^^^^^^^
+
+During the Bootstrap stages, when we are assertive that the K8s cluster is
+fully ready and available we could perform the following actions:
+
+  - Create and deploy ConfigMaps that hold cluster and service configurations
+    and pre-fill them with default values.
+  - Template service pods and deployments to consume configuration data
+    directly from the above deployed ConfigMaps
+
+This approach works because in a MetalK8s cluster, ConfigMaps for cluster and
+service configurations are available before we deploy the configured services.
+
+**Using Salt states**
+
+Once a ConfigMap is updated by the user (say a user changes the number of
+replicas for Prometheus deployments to a new value), then perform the
+following actions:
+
+  - Apply a Salt state that reads the ConfigMap object, validates the schema
+    and checks the new values passed and re-applies this configuration value to
+    the deployment in question.
+  - Restart the Kubernetes deployment to pickup newly applied service
+    configurations.
+
+Storage format
+~~~~~~~~~~~~~~
+
+A YAML (K8s-like) format was chosen to represent the data field instead of a
+flat key-value structure for the following reasons:
+
+ - YAML formatted configurations are easy to write and understand hence it will
+   be simpler for users to edit configurations.
+ - The YAML format benefits from bearing a schema version, which can be checked
+   and validated against a version we deploy.
+ - YAML is a format for describing hierarchical data structures, while using a
+   flat key-value format would require a form of encoding (and then, decoding)
+   of this hierarchical structure.
+
+A sample ConfigMap can be defined with the following fields.
+
+.. code-block:: yaml
+
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      namespace: <namespace>
+      name: <config-name>
+    data:
+      config.yaml: |-
+        apiVersion: <object-version>
+        kind: <kind>
+        spec:
+          <key>: <values>
+
+**Use case 1:**
+
+Configure and store the number of replicas for service specific Deployments
+found in the `metalk8s-monitoring` namespace using the ConfigMap format.
+
+.. code-block:: yaml
+
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      namespace: metalk8s-monitoring
+      name: metalk8s-grafana-config
+    data:
+      config.yaml: |-
+        apiVersion: metalk8s.scality.com/v1alpha1
+        kind: GrafanaConfig
+        spec:
+          deployment:
+            replicas: 2
+
+Non-goals
+~~~~~~~~~
+
+This section contains requirements stated above which the current design choice
+does not cater for and will be addressed later:
+
+- Persisting newly added Grafana dashboards or new Grafana datasources
+  especially for modifications added via the Grafana UI cannot be stored in
+  ConfigMaps and hence will be catered for later.
+
+- As stated in the requirements, adding and editing Prometheus alert rules
+  is also not covered by the chosen design choice and will need to be addressed
+  differently. Even if we could use ConfigMaps for Prometheus rules, we prefer
+  relying on the Prometheus Operator and it's CRD (PrometheusRule).
+
+Rejected design choices
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Consul KV vs ConfigMap
+~~~~~~~~~~~~~~~~~~~~~~
+
+This approach offers a full fledge KV store with a /kv endpoint which allows
+CRUD operations on all KV data stored in it.
+Consul KV also allows access to past versions of objects and has an optimistic
+concurrency when manipulating multiple objects.
+
+Note that, Consul KV store was rejected because managing operations such as
+performing full backups, system restores for a full fledged KV system
+requires time and much more efforts than the ConfigMap approach.
+
+Operator (Custom Controller) vs Salt
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Operators are useful in that, they provide self-healing functionalities on a
+reactive basis. When a user changes a given configuration, it is easy to
+reconcile and apply these changes to the in-cluster objects.
+
+The Operator approach was rejected because it is much more complex, requires
+much more effort to realize and there is no real need for applying changes
+using this method because configuration changes are not frequent
+(for a typical MetalK8s admin, changing the number of replicas for a given
+deployment could happen once in 3 months or less) as such, having an operator
+watch for object changes is not significant and not very useful at this point
+in time.
+
+In the Salt approach, Salt Formulas are designed to be idempotent ensuring that
+service configuration changes can be applied each time a new configuration is
+introduced.
+
+Implementation Details
+----------------------
+
+Iteration 1
+^^^^^^^^^^^
+
+- Define and deploy new ConfigMap stores that will hold cluster and service
+  configurations as listed in the requirements. For each ConfigMap, define its
+  schema, its default values, and how it impacts the configured services
+- Template and render Deployment and Pod manifests that will make use of
+  this persisted cluster and service configurations
+- Document how to change cluster and service configurations using kubectl
+- Document the entire list of configurations which can be changed by the user
+
+Iteration 2
+^^^^^^^^^^^
+
+- Provide a CLI tool for changing any of the cluster and service
+  configurations:
+
+    - Count of replicas for chosen Deployments (Prometheus)
+    - Updating a Dex authentication connector (OpenLDAP, AD and
+      staticUser store)
+    - Updating the Alertmanager notification configuration
+
+- Provide a UI interface for adding, updating and deleting service specific
+  configurations for example Dex-LDAP connector integration.
+- Provide a UI interface for listing MetalK8s available/supported Dex
+  authentication Connectors
+- Provide a UI interface for enabling or disabling Dex authentication
+  connectors (LDAP, Active Directory, StaticUser store)
+- Add a UI interface for listing Alertmanager notification systems MetalK8s
+  will support (Slack, email)
+- Provide a UI interface for adding, modifying and deleting Alertmanager
+  configurations from the listing above
+
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document how to customize or change any given service settings using the CLI
+  tool
+* Document how to customize or change any given service settings using the UI
+  interface
+* Document the list of service settings which can be configured by the user
+
+Test Plan
+---------
+
+- Add test that ensures that update operations on user configurations are
+  propagated down to the various services
+
+- Add test that ensures that after a MetalK8s upgrade, we do not lose previous
+  customizations.
+
+- Other corner cases that require testing to reduce error prone setups include:
+
+   - Checking for invalid values in a user defined configuration (e.g setting
+     the number of replicas to a string ("two"))
+   - Checking for invalid formats in a user configuration
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -31,6 +31,17 @@ Glossary
      and from where the cluster will be deployed to other machines. It also
      serves as the entrypoint for upgrades of the cluster.
 
+   ConfigMap
+     A ConfigMap is a Kubernetes object that allows one to store general
+     configuration information such as environment variables in a key-value
+     pair format.
+     ConfigMaps can only be applied to namespaces and once created, they can
+     be updated automatically without the need of restarting containers that
+     depend on it.
+
+     |see K8s docs|
+     `ConfigMap <https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#understanding-configmaps-and-pods/>`_.
+
    Controller Manager
    ``kube-controller-manager``
      The Kubernetes controller manager embeds the core control loops shipped


docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+        receivers.slack_configs.send_resolved: 'true'
+
+Iteration 1
+~~~~~~~~~~~
+
+- Define and deploy new namespaced bounded ConfigMap KV stores that will hold
+  user configurations as listed in the requirements
+- Template and render Deployment and Pod manifests that will make use of
+  this user customizable configurations using default pillar values
+- Provide a CLI tool(Bash script) for changing any of the user configurations:
```
slaperche-scality: Why Bash instead of Python?
I thought we were moving away from Shell script.

Ebaneck: Very true @slaperche-scality 
I think we have an entire EPIC to handle this migration and it is debatable if we should wait or have something working(shell scripts) while waiting.

Maybe, after all, we do not even have to handle this point within this EPIC

slaperche-scality: The other EPIC is about migrating existing shellscript.

Meanwhile, for new ones that we creates from scratch, it could be a good idea to use Python from the beginning (except if it's really more costly than Bash for some reasons): would be less work for the other EPIC, while adding no extra-work to the current one (anyway you have to write a new script)

gdemonet: In my opinion, wait for the Python CLI epic (#2163) to have the foundations set in place before implementing a CLI for this. In any case, the CLI / UI implementation of those customizations is less urgent than having a framework to persist said customizations over upgrade/downgrade/... IMO, CLI / UI shouldn't be part of the first iteration.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+
+.. todo::
+
+   This section will be handled by the requirements PR.
+
+
+Design Choices
+--------------
+
+:term:`ConfigMap` KV(Key-Value) store is chosen as a unified data access and
```
gdemonet: I'm uncomfortable with this focus on "key-value" nature of a `ConfigMap` - does it imply we will use a single `ConfigMap` with a key for each "service" to be configured? I'm not sure that's what we want actually, for multiple reasons (which may be adjusted after we get the Requirements btw):

- CMs are namespace-scoped, and so are most of the services we deploy with MetalK8s. Wouldn't it make more sense if we keep a configuration CM next to what it configures? E.g. `metalk8s-auth-userconfig` in the `metalk8s-auth` namespace?
- Editing a single CM with many keys is less user-friendly than having smaller CMs per service, each with a single key. Although we plan on providing better interfaces through CLI and UI, it would make sense to make the underlying format as friendly as possible (this will make a first iteration with no CLI/GUI much more useful)

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+storage media for user editable configurations in a MetalK8s cluster based on
+the above requirements for the following reasons:
+
+* The limited number of read/write operations even for multi-node setups means
+  that concurrency issues on CRUD operations will not lag even if we do not
+  implement cross-app locking.
```
gdemonet: "concurrency issues [...] will not lag" - not sure what you meant?

In any case, is it important to mention in the design? We're not looking for a database, we're looking for means of configuration. As a matter of fact, we **do have** a database (`etcd`), and a nice schema already in place that we can use (the K8s API resources, including `ConfigMap`s) for such configurations.

The concurrency matters on configuration of cluster-wide services are, as you suggest here, not a problem. Let's move this point to the "non-goals" in the Requirements (cc @thomasdanan), and leave it aside.

We need our design choices to clearly explain **why** we chose this approach according to our "goals", not to our "non-goals", IMO.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+* Ability to support CRUD operations on ConfigMap's with CLI and UI easily
```
gdemonet: Why this focus on CRUD? I'm actually thinking, we only want the users to perform **U**pdate operations on those objects, since MetalK8s is responsible for **C**reating and **D**eleting those objects - so the fact that all CRUD ops are supported could be a problem (luckily, we can simply have a `Role` which only allows `patch` on the right objects).

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+  using our already existing python kubernetes module.
+* Guarantee of adaptability and ease of changing the design and implementation
+  in cases were customer needs evolve rapidly(modular design nature)
```
gdemonet: What is modular in this design? Especially with the "single ConfigMap KVS" approach.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+* Ease of operation and a guarantee of no downtime in a MetalK8s cluster
+  especially when configuration changes are made on the fly.
```
gdemonet: How does using ConfigMaps handle "no-downtime" guarantees?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+Context
+-------
+
+.. todo::
+
+   This section will be handled by the requirements PR.
+
+
+Design Choices
+--------------
```
gdemonet: This design only discusses the choice of storage (method and media). What about the concerns of translating these user configurations into services configurations? Are the services supposed to know about our user configs and parse them? Are we (Salt) deriving actual service configs from our user configs? Why Salt and not a K8s Operator?

docs/developer/architecture/user_customization_and_persistence.rst
gdemonet: > changes are made on the fly.

WDYM? Do you expect a change to impact the target service right away? Or not?
"made on the fly" would mean something like `kubectl edit cm ...` or something else?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+requires time and much more efforts than the ConfigMap KV store which is
+simplistic and matches the requirements stated.
+
+
+Implementation Details
+----------------------
+
+A sample ConfigMap KV store can be defined with the following fields. These
+objects will be namespace bounded ensuring that sensitive information is
+restricted from a cluster-wide access.
```
gdemonet: CMs are always namespace-scoped.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+
+       apiVersion: v1
+       kind: ConfigMap
+       metadata:
+         namespace: metalk8s-monitoring
+         name: <example-name>
+       data:
+         replicas.grafana: 2
+         replicas.prometheus: 2
+         replicas.alertmanager: 1
```
gdemonet: This kind of format is not straightforward to manage with code, since you'll need to parse the flat keys into hierarchical structures. As discussed during the short design session, we'd rather have something more "K8s-like":

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: metalk8s-grafana-userconfig
data:
  config.yaml: |-
    apiVersion: metalk8s.scality.com/v1alpha1
    kind: GrafanaUserConfig
    spec:
      replicas: 2
```

This also has the benefit of bearing a schema version, which we can check before inferring the underlying service configuration.

docs/developer/architecture/user_customization_and_persistence.rst
gdemonet: CLI tooling doesn't need to be part of the first iteration IMO - well formatted CMs are easy enough to manage both manually and in tests/scripts.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+- Template and render Deployment and Pod manifests that will make use of
+  this user customizable configurations using default pillar values
+- Provide a CLI tool(Bash script) for changing any of the user configurations:
+
+    - Count of replicas for chosen Deployments(Prometheus)
+    - Adding/modifying/removing a Dex authentication connector(OpenLDAP, AD and
+      staticUser store)
+    - Adding/modifying/removing Alertmanager notification configuration
+
+Iteration 2
```
gdemonet: UI and CLI could both be part of iteration 2 I think, and we could add a third iteration to include pre-edit validation hooks, display of both user config and generated service configs... We'll see with the reqs.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+- Provide a UI interface for adding, modifying and deleting Alertmanager
+  configurations from the listing above
+
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document how to customize or change any given setting using the CLI tool
+* Document how to customize or change any given setting using the UI interface
```
gdemonet: OK, maybe a "common doc" for CLI / UI. But we'll also need this info next to the services we want our users to configure (e.g. show how to configure Dex in the section/page dedicated to Dex)

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,214 @@
...
+
+Dex Static User authentication is currently covered in our test-suite and it
+will make sense to cover atleast one other authentication connector with the
+easiest being LDAP since we readily have access to OpenLDAP Docker images and
+automating this process is possible on Scality Cloud.
+
+CRUD Operations on the ConfigMap KV store should be tested both from the UI
+test libraries and the CLI tool provided(An initial idea here is to have a
+simplistic configurator.sh bash script(CLI tool) performing sets of CRUD
+operations on a sample ConfigMap KV store).
```
gdemonet: Standard operations to edit configmaps are not super useful to test - what we want to check is propagation of those changes down to the configured services.
E2E tests with CLI and UI are a good thing, because (I guess) those would also trigger the right Salt calls after the CM(s) update(s).

We also need to test edge-cases:
- invalid format for the user configs
- invalid values in the user configs
- lost a user config (someone deleted it)

gdemonet (changes requested):
Early review, we'll see how it goes with requirements.
I think you should focus more on "what happens after the user has configured something", not only on "how will the user configure something".

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Well, this was supposed to mean; the fact that we are making use of ConfigMaps, I think once a user updates a given configuration, we do not require Pod restarts for the specific services to pick up the changes. no?

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: ok

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: ok

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Done

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Done, removed the focus on `key-value` nature

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: ok

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Thanks for the reminder! even though i think at the end of the day, we would have littered the entire namespace with many small ConfigMap objects.

What are the chances that we could group the number of replicas specific ConfigMap for all services in the Monitoring namespace?

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Moved UI and CLI operations to iteration 2 as suggested. Not sure about third iteration so i'm gonna have to pass!

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: I added a `How it works section` that takes care of the questions raised here. This might need validation from You

docs/developer/architecture/user_customization_and_persistence.rst
TeddyAndrieux: Not really agree with your approach here.
If we use a config map IMO we should have just "key value" like Claude did and when we will use specific CRD then we could have object like you suggest.
Also IMO we should do one configmap per "object/service" (e.g.: One configmap for auth/dex, one for grafana, one for prometheus (or maybe one for monitoring regrouping prom, grafana altermanager), one for ...)

Ebaneck: > Thanks for the reminder! even though i think at the end of the day, we would have littered the entire namespace with many small ConfigMap objects.
> 
> What are the chances that we could group the number of replicas specific ConfigMap for all services in the Monitoring namespace?

Oh, let us keep it service-specific as suggested since the configmap of one particular service might grow drastically when new customizations are included

gdemonet: > If we use a config map IMO we should have just "key value" like Claude did and when we will use specific CRD then we could have object like you suggest.

Then it means we'll need to implement the parsing logic twice? Why not stick with the same YAML style/format?

TeddyAndrieux: The configmap will be read from salt so if we use the "kubernetes style" that mean we need to implement it in our salt modules as we will have a configmap with several "key=string" and we will need to then load and parse the string to get the yaml we want.
If we stick with key=value we will just have to read a classic dict in salt so it's really easy.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,225 @@
...
+**Using Saltstates**
+
+Once a ConfigMap KV store is updated by the user(say a user changes the
+number of replicas for Prometheus deployments to a new value), then perform the
+following actions;
+
+  - Apply a salt state that reads the ConfigMap object, validates the schema
+    based on MetalK8s defined standards and checks the new values passed.
+  - If the ConfigMap object is valid, the new values passed by the user are
+    then re-rendered to the pillars.
```
TeddyAndrieux: Not sure we really want this value to be in the pillar, IMO just reading the CM directly in the state is better so that user can NOT change, for example, the number of replicas just on CLI using `pillar="blabla"` but he always need to edit the CM

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,225 @@
...
+  - Finally, we make sure that the updated values are picked up by their
+    respective consumers(this might require Pod restarts for changes to take
+    effect).
+
+Note that, salt-states are used to sync data and update consumers
+of new configurations changes mainly because of the minimum effort it takes to
+setup this flow(i.e. the act of configuration update by the user and it's
+propagation to the consumers) but the K8s Operator pattern could be use to
+replace configuration synchronization between user defined configurations and
+consumers.
```
TeddyAndrieux: Not sure we should talk about it here, for me we should take a desicion in this document and maybe do not implement the operator in the first iteration and only use salt but at least having the final view here is better IMO

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,225 @@
...
+  - The Operator is connected to the API server to watch for changes in the
+    ConfigMap.
+  - If the Operator detects a modified ConfigMap, it then determines which line
+    of action it should take which are;
+
+      - Extract the ConfigMap name and object fields
+      - Extract the pods associated to this ConfigMap based on it's labels
+      - Read and validate the ConfigMap data and schema
+      - If the ConfigMap is valid, update the pillars and restart the
+        respective pods such that they pick-up new configs from the pillars.
```
TeddyAndrieux: Not from the pillar in this approach

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,225 @@
...
+changing the number of replicas for a given deployment could happen once in 3
+months) as such, having an operator watch for object changes is not significant
+and not very useful at this point in time.
+
+**Using Operator architecture(Custom Controllers)**
+
+When using an Operator(a Custom Controller that works with CRDs), we create a
+Custom Resource Definition (CRD) that references a ConfigMap. Once a ConfigMap
+is updated by the user, then the Operator is designed to perform the following
+actions;
```
TeddyAndrieux: Operator and CRD are not related we could use CRD with salt and use Operator with ConfigMap.

Also if we use Operator and CRD then we no longer need ConfigMap IMO we directly use CRD as entry point

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,225 @@
...
+    of action it should take which are;
+
+      - Extract the ConfigMap name and object fields
+      - Extract the pods associated to this ConfigMap based on it's labels
+      - Read and validate the ConfigMap data and schema
+      - If the ConfigMap is valid, update the pillars and restart the
+        respective pods such that they pick-up new configs from the pillars.
+      - If the ConfigMap is invalid, log the error and perform no further
+        action because a bad ConfigMap being applied could lead to cluster
+        outages
```
TeddyAndrieux: Hmm I not really like this point, we not directly apply the ConfigMap but we use it to modify another object and i don't see case where an invalid value can cause outages for me apiServer will just not set the wrong value.
For example if you use a string as replica then opeator will try to edit an object and apiServer will just reject the change, ... 
Maybe for other custom tuning that could lead to cluster outages but ... I don't see where

Ebaneck: ok... resolved!

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: As discussed in person... very true!

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Fixed

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+:term:`ConfigMap` store is chosen as a unified data access and
+storage media for user editable configurations in a MetalK8s cluster based on
+the above requirements for the following reasons:
+
+* Ability to support Update operations on ConfigMap's with CLI and UI easily
+  using our already existing python kubernetes module.
+* Guarantee of adaptability and ease of changing the design and implementation
+  in cases were customer needs evolve rapidly.
+* ConfigMap data is store in the :term:`etcd` database which is generally being
```
gdemonet:
```suggestion
* ConfigMaps are stored in the :term:`etcd` database which is generally being
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+  backed up. This ensures that user settings cannot be lost easily.
+
+.. note::
+
+   Persisting newly added Grafana dashboards or new Grafana datasources
+   especially for modifications added via the Grafana UI cannot be stored in
+   ConfigMaps.
+   To handle this particular user case, there is a need to provision persistent
+   storage volumes to handle the persistence of these settings across Pod
+   restarts.
```
gdemonet: True. I would create a section for "non-goals" of this design, including this point on Grafana, but also other aspects (providing a general solution for recovery after total site loss, periodic backups, ...).

docs/developer/architecture/user_customization_and_persistence.rst
gdemonet: Good content, though I'd rather state this in the _Design Choices_ section. In addition, there seems to be two implementation alternatives (Salt vs. Operator). Since we apparently decide to use Salt (and not an Operator), why not show Salt as the design + Operator as a rejected option?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+Note that, Consul KV store was rejected because managing operations such as
+performing full backups, system restores for a full fledged KV system
+requires time and much more efforts than the ConfigMap KV store which is
+simplistic and matches the requirements stated.
+
+
+Implementation Details
+----------------------
+
+Storage format
```
gdemonet: The actual format is part of the design. It would be worth adding an explanation in the _Design Choices_ as to why we want to use the traditional versioned `apiVersion/kind` YAML style instead of flat KV style (what benefits it brings, what constraints / extra efforts it requires...).

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+      namespace: metalk8s-monitoring
+      name: metalk8s-grafana-userconfig
+    data:
+      config.yaml: |-
+        apiVersion: metalk8s.scality.com/v1alpha1
+        kind: GrafanaUserConfig
+        spec:
+          replicas: 2
+        podSelector:
+          app: grafana
```
gdemonet: Probably missing an indent?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+~~~~~~~~~~~~
+
+During Bootstrap, pillar values for the various services will be pre-filled
+with default values. Service pods and deployments will be configured to either
+consume configuration data directly from their ConfigMap store or fallback to
+the default pillar values if the service specific ConfigMap store was not found.
+
+**Using Saltstates**
+
+Once a ConfigMap KV store is updated by the user(say a user changes the
```
gdemonet:
```suggestion
Once a ConfigMap KV store is updated by the user (say a user changes the
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+          replicas: 2
+        podSelector:
+          app: grafana
+
+How it works
+~~~~~~~~~~~~
+
+During Bootstrap, pillar values for the various services will be pre-filled
+with default values. Service pods and deployments will be configured to either
+consume configuration data directly from their ConfigMap store or fallback to
```
gdemonet: Using simply _ConfigMap_ i.s.o "ConfigMap store" or "ConfigMap KV store" should be sufficient for readers - could you change this here and in other relevant places?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+the default pillar values if the service specific ConfigMap store was not found.
+
+**Using Saltstates**
```
gdemonet: Let's write this "Salt states" and use the same notation everywhere (no "Saltstate", "salt state", or "salt-state")

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Once a ConfigMap KV store is updated by the user(say a user changes the
+number of replicas for Prometheus deployments to a new value), then perform the
+following actions;
```
gdemonet:
```suggestion
following actions:
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+  - Apply a salt state that reads the ConfigMap object, validates the schema
+    and checks the new values passed.
+
+  - Finally, we make sure that the updated values are picked up by their
+    respective consumers(this might require Pod restarts for changes to take
```
gdemonet:
```suggestion
    respective consumers (this might require Pod restarts for changes to take
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+    effect). The podSelector field has labels which can be used to detect which
```
gdemonet: There should be no need for us to "detect" what needs to restart: we own these objects, and the configuration options we expose, so it's safe to assume we'll always know what to do after a re-configuration state is applied.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+setup this flow but the K8s Operator pattern could be use to replace
+configuration synchronization between user defined configurations and
+consumers.
+
+The Operator approach is much more complex, requires much more effort
+to realize and there is no real need for applying changes using this method
+because configuration changes are not frequent(for a typical MetalK8s admin,
+changing the number of replicas for a given deployment could happen once in 3
+months) as such, having an operator watch for object changes is not significant
+and not very useful at this point in time.
```
gdemonet: Agreed with your points. Would just add that Operators are especially useful to provide self-healing functionalities on a reactive basis. When a user changes configuration, it is easy to apply (in the same flow) those configurations to the in-cluster objects. Once deployed, those configurations will be handled by the cluster controllers (Deployments, ConfigMap/Secret-based volume mounts, ...), and self-healing comes out of the box. No real need for more than what Salt provides.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+**Using Operator architecture(Custom Controllers)**
```
gdemonet: No need to provide a design if it's rejected. Simply explain why it's rejected, this will help readers quickly understand the overall architecture document.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+      - If the ConfigMap is invalid, log the error and perform no further
+        action.
+
+Iteration 1
+~~~~~~~~~~~
+
+- Define and deploy new ConfigMap stores that will hold user configurations
+  as listed in the requirements
+- Template and render Deployment and Pod manifests that will make use of
+  this user customizable configurations using pillar values
```
gdemonet: Are we using Pillar values or directly call into `metalk8s_kubernetes` module in the end? cc. @TeddyAndrieux 

Maybe this could be a good fit for `__slots__` (and especially with [dict support in Salt 3000](https://docs.saltstack.com/en/latest/topics/slots/index.html))?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+- Document how to change user configurations using kubectl
+- Create and deploy persistent storage volumes for Grafana dashboards and
+  datasources
```
gdemonet: We need to configure Grafana to use a PVC (not create the volumes ourselves). Just do the same as Prom.
Also, if it's out of the design (and I think we have an epic for Grafana: #2218), maybe just don't state it here.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+- Document how to create these persistent volume for Grafana dashboards and
+  datasources
+- Document how to add Prometheus alerting rules using a ConfigMap. This
```
gdemonet: Why not use the existing `PrometheusRule` CRD that comes with `prometheus-operator`? If it's a conscious decision, it would make sense to explain in the design choices.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+  documentation should explain to users that the Prometheus object has a
+  ruleSelector field, which selects the ConfigMaps from which rules files are
+  loaded from. This means that to add new Prometheus alert rules, Users need to
+  create new ConfigMaps containing alert rules with labels matching what was
+  defined in the Prometheus ruleSelector field.
```
gdemonet: Ah! Actually, this is true if you replace `ConfigMap` with `PrometheusRule`, IIUC. Or maybe not in our version of `prometheus-operator`...? Didn't check

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Iteration 2
+~~~~~~~~~~~
+
+- Provide a CLI tool for changing any of the user configurations:
+
+    - Count of replicas for chosen Deployments(Prometheus)
```
gdemonet: General comment: add a space before opening a parenthesis.

```suggestion
    - Count of replicas for chosen Deployments (Prometheus)
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+    - Updating a Dex authentication connector(OpenLDAP, AD and
```
gdemonet:
```suggestion
    - Updating a Dex authentication connector (OpenLDAP, AD and
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+      staticUser store)
+    - Updating the Alertmanager notification configuration
+
+- Provide a UI interface that allows Update operations on all user customizable
+  settings based on the above requirements
```
gdemonet: From the requirements, CLI provides this "all settings possible" capability, while GUI provides more user-friendly / less detailed forms to configure certain services (as you described below).
I think (need to be confirmed) this point can be removed.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document how to customize or change any given service settings using the CLI
+  tool
+* Document how to customize or change any given service settings using the UI
+  interface
+* Document how to add and edit Prometheus alerting rules using the CLI
```
gdemonet: Why this specific example and not all services?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,224 @@
...
+
+Test Plan
+---------
+
+- Dex Static User authentication is currently covered in our test-suite and it
+  will make sense to cover atleast one other authentication connector with the
+  easiest being LDAP since we readily have access to OpenLDAP Docker images and
+  automating this process is possible on Scality Cloud
```
gdemonet: Out of scope - this relates to testing Dex, which is part of the Authentication document.

gdemonet (dismissed):

docs/developer/architecture/user_customization_and_persistence.rst
TeddyAndrieux: IMO we should use directly `metalk8s_kubernetes` module.

I think we don't really need `__slots__` here, basic jinja is enough

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Will keep it simple and resolve to users editing the original CRD!

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Not tested, so let's leave it at editing the original CRD.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+  backed up. This ensures that user settings cannot be lost easily.
+
+
+How it works
+~~~~~~~~~~~~
+
+During Bootstrap, pillar values for the various services will be pre-filled
+with default values. Service pods and deployments will be configured to either
+consume configuration data directly from their ConfigMap or fallback to
+the default pillar values if the service specific ConfigMap was not found.
```
NicolasT: What if we did this differently?
None of these 'user configurable' settings (at this level) should influence how the K8s cluster behaves: this should be dictated by the `BootstrapConfiguration`. As such during bootstrap, at some point *before we need any of these settings*, the K8s cluster is fully available.

Which means that at this point, we could create all these `ConfigMap`s, filled with whatever default values we choose.

At this point, we can consume them while deploying various services without any more 'defaults' defined anywhere.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+Once a ConfigMap is updated by the user (say a user changes the number of
+replicas for Prometheus deployments to a new value), then perform the
+following actions:
+
+  - Apply a Salt states that reads the ConfigMap object, validates the schema
+    and checks the new values passed.
+
+  - Finally, we make sure that the updated values are picked up by their
+    respective consumers (this might require Pod restarts for changes to take
+    effect).
```
NicolasT: The fact that Pods may be restarted is always the case in K8s, not sure this should be mentioned. In the end, Salt (MetalK8s) is not restarting any Pods. This should all be managed by the `Deployment`/`StatefulSet`/`DaemonSet` controllers because Salt updates some high-level object(s).

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Note that, Salt states are used to sync data and update consumers
+of new configurations changes mainly because of the minimum effort it takes to
+setup this flow but the K8s Operator pattern could be use to replace
```
NicolasT:
```suggestion
setup this flow but the K8s Operator pattern could be used to replace
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+configuration synchronization between user defined configurations and
```
NicolasT: `... could be used to synchronize configuration between user-defined settings and the applications.`

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+consumers.
+
+Non-goals
+~~~~~~~~~
+
+- Persisting newly added Grafana dashboards or new Grafana datasources
+  especially for modifications added via the Grafana UI cannot be stored in
+  ConfigMaps.
+  To handle this particular user case, we need to configure Grafana deployments
```
NicolasT:
```suggestion
  To handle this particular use-case, we need to configure Grafana deployments
```

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+  to use persistent volume claims and then document how to create volumes that
```
NicolasT: Here you're prescribing *how* a certain feature *not* covered by this design ought to be implemented. That's not OK. Indeed, this document should describe what it *won't* cover, but whether this other feature is implemented using PVs or some other mechanism is to be determined later/elsewhere.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+  will handle the persistence of these settings across Pod restarts.
+
+- ConfigMaps as used in the above design for storing and persisting user
+  settings need periodic backups for recovery purposes. This implies that
```
NicolasT: I get what you mean but I don't like how it's phrased. We don't need to put backups in place to store these user settings and such. Instead, using `ConfigMap`s (or other K8s object types) has as a benefit that K8s cluster backups include these settings. The fact that the K8s `etcd` cluster needs to be backed up is not driven by this feature.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Operators are useful in that, they provide self-healing functionalities on a
+reactive basis. When a user changes a given configuration, it is easy to
+reconcile and apply these changes to the in-cluster objects.
+
+The Operator approach was rejected because it is much more complex, requires
+much more effort to realize and there is no real need for applying changes
+using this method because configuration changes are not frequent
+(for a typical MetalK8s admin, changing the number of replicas for a given
+deployment could happen once in 3 months) as such, having an operator watch for
```
NicolasT: Or less.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      namespace: metalk8s-monitoring
+      name: metalk8s-grafana-userconfig
+    data:
+      config.yaml: |-
+        apiVersion: metalk8s.scality.com/v1alpha1
+        kind: GrafanaUserConfig
```
NicolasT: Not sure I like the `GrafanaUserConfig` name. We're not configuring Grafana users here... `GrafanaCustomizations`? `GrafanaConfig`? ...

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+        spec:
+          replicas: 2
```
NicolasT: This is a low-level setting at top level in the config. Not sure that's great.
```yaml
spec:
  deployment:
    replicas: 2
```
(then other `deployment`-related settings can go there as well)

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Well, we still have an open discussion about the format to use and from what I see, the flat KV style sits well with our already existing python-Kubernetes module....

docs/developer/architecture/user_customization_and_persistence.rst
NicolasT: I'd try to steer away from `__slots__` as well. The values for these settings can be calculated/retrieved before templating starts, so let's do that.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Iteration 1
+~~~~~~~~~~~
+
+- Define and deploy new ConfigMap stores that will hold user configurations
+  as listed in the requirements
+- Template and render Deployment and Pod manifests that will make use of
+  this user customizable configurations using pillar values
+- Document how to change user configurations using kubectl
+- Document how to add Prometheus alerting rules by editing the Prometheus rule
```
NicolasT: I don't see how this is in scope of this design?

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,199 @@
...
+- Provide a UI interface for adding, updating and deleting service specific
+  configurations for example Dex-LDAP  connector integration.
+- Provide a UI interface for listing MetalK8s available/supported Dex
+  authentication Connectors
+- Provide a UI interface for enabling or disabling Dex authentication
+  connectors (LDAP, Active Directory, StaticUser store)
+- Provide a UI interface for changing the number of replicas for a chosen set
+  of MetalK8s deployments (Prometheus, ....)
+- Add a UI interface for listing Alertmanager notification systems MetalK8s
+  will support (Slack, email, hipchat)
```
NicolasT: HipChat is dead, no need to list it anywhere.

NicolasT (dismissed):

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: True, I will move it to non-goals and clearly state that the current design does not cater for this. No need to mention an implementation choice.

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: fixed

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: ok

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: Picked the latter!

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: removed

docs/developer/architecture/user_customization_and_persistence.rst
Ebaneck: True! after a couple of discussions, an alternative we considered and updated the design docs was keeping these 'defaults' in our `default.yaml` file and then doing something like ```{%- from "metalk8s/map.jinja" import addons with context %}```. This meant in our add-ons charts we had to implement some conditional to read from the Configmaps or fall back to these defined defaults(just like pillar pre-filling idea but cleaner).

Thanks to your approach, it appears we do not even need to declare defaults in `defaults.yaml`.  Waiting to hear from others what they think.

TeddyAndrieux: Like I told you Monday 3 choice:
- We have `defaults.yaml` containing this default number of replica
- we create the custos CM and "pre-fill" them **if they do not exists**
- We just put default value directly in the `sls` file

Personally suggest the second one as it will be much simpler for the user to change them if CM already exists and already contain value we want

gdemonet: Option 2 looks good to me as well. A simple `kubectl delete cm ..` followed by `salt-run state.orch metalk8s.deployed` to restore defaults is acceptable I guess.

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,187 @@
...
+This approach offers a full fledge KV store with a /kv endpoint which allows
+CRUD operations on all KV data stored in it.
+Consul KV also allows access to past versions of objects and has an optimistic
+concurrency when manipulating multiple objects.
+
+Note that, Consul KV store was rejected because managing operations such as
+performing full backups, system restores for a full fledged KV system
+requires time and much more efforts than the ConfigMap approach.
+
+Operator (Custom Controller) vs ConfigMap
```
gdemonet: `vs. Salt` - as Teddy said, Operator is not related to a given storage medium, and could work with plain CM

docs/developer/architecture/user_customization_and_persistence.rst
```diff
@@ -0,0 +1,187 @@
...
+replicas for Prometheus deployments to a new value), then perform the
+following actions:
+
+  - Apply a Salt states that reads the ConfigMap object, validates the schema
+    and checks the new values passed.
+
+Storage format
+~~~~~~~~~~~~~~
+
+A sample ConfigMap can be defined with the following fields.
```
gdemonet: Why this format? Explain in the design, compare with flat KV approach and why we prefer this one.

Ebaneck: Done

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
 expected.
+
+Design Choices
+--------------
+
+:term:`ConfigMap` is chosen as a unified data access and storage media for
+cluster and service configurations in a MetalK8s cluster based on the above
+requirements for the following reasons:
+
+* Ability to support Update operations on ConfigMap's with CLI and UI easily
```
gdemonet:
```suggestion
* Ability to support Update operations on ConfigMaps with CLI and UI easily
```

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+  using our already existing python kubernetes module.
+* Guarantee of adaptability and ease of changing the design and implementation
+  in cases were customer needs evolve rapidly.
```
gdemonet:
```suggestion
  in cases where customer needs evolve rapidly.
```

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+    and pre-fill them with default values.
+  - Template service pods and deployments to consume configuration data
+    directly from the above deployed ConfigMaps
+
+This approach works because in a MetalK8s cluster, cluster and service
+configurations do not influence how the cluster behaves but rather the
+BootstrapConfiguration dictates the cluster behaviour. Also the creation of
+cluster and service configurations takes place before the respective consumers
+(MetalK8s addons) are deployed hence configuration data is always ready for
+consumers and not the other way round.
```
gdemonet: I think you're trying to say that service configurations (and maybe others that affect the cluster **after** initial deployment) are a good fit for ConfigMaps, whereas BootstrapConfiguration needs to stay as a file, since we need it before initial deployment. It also evokes the need for us to create said CMs before deploying the configured services.

Rewording a bit, we could simply state that:
- we need special casing for some configs (what goes in BootstrapConfig)
- we can store all the rest in the API, as long as we make sure these CMs are available **before** we deploy the configured services

Also:
> cluster [...] configurations do not influence how the cluster behaves

This looks incorrect, or if it's not, then I guess those configurations are useless? :thinking:

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+replicas for Prometheus deployments to a new value), then perform the
+following actions:
+
+  - Apply a Salt states that reads the ConfigMap object, validates the schema
+    and checks the new values passed.
+
+Storage format
+~~~~~~~~~~~~~~
+
+A YAML(K8s-like) format was chosen to represent the data field instead of a
```
gdemonet:
```suggestion
A YAML (K8s-like) format was chosen to represent the data field instead of a
```

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+flat key-value structure for the following reasons:
+
+ - YAML formatted configurations are easy to write and understand hence it will
+   be simpler for users to edit configurations.
+ - The YAML format benefits from bearing a schema version, which can be checked
+   before inferring the underlying service configuration if need arises.
```
gdemonet: > checked before inferring

What do you mean? We don't want to hold default values or infer anything on our side. We just want to validate that the "version" of the configuration matches what we deploy.

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+ - The YAML format utilizes white spaces to define variable hierarchies and as
+   such it is easier to manage with code mean while the flat key-value format
+   requires parsing of flat keys into hierarchical structures which can be
+   tedious.
```
gdemonet: True, but not super clear. Maybe simply state that YAML **is** a format for describing hierarchical data structures, while using a flat structure would require a form of encoding (and then, decoding) of this hierarchical structure, which we know we'll need.

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+This section contains requirements stated above which the current design choice
+does not cater for and will be addressed later:
+
+- Persisting newly added Grafana dashboards or new Grafana datasources
+  especially for modifications added via the Grafana UI cannot be stored in
+  ConfigMaps and hence will be catered for later.
+
+- As stated in the requirements, adding and editing Prometheus alert rules
+  is also not covered by the chosen design choice and will need to be addressed
+  differently.
```
gdemonet: Not clear enough why we don't cover Prometheus rules (and not only alerting rules are affected, btw). Explain that even if we **could** use ConfigMaps for said rules, we **prefer** relying on the Prometheus Operator (which we deploy) and its CRDs (especially, the `PrometheusRule` one).

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+reactive basis. When a user changes a given configuration, it is easy to
+reconcile and apply these changes to the in-cluster objects.
+
+The Operator approach was rejected because it is much more complex, requires
+much more effort to realize and there is no real need for applying changes
+using this method because configuration changes are not frequent
+(for a typical MetalK8s admin, changing the number of replicas for a given
+deployment could happen once in 3 months or less) as such, having an operator
+watch for object changes is not significant and not very useful at this point
+in time.
```
gdemonet, line 210: Good explanation. Maybe add a small paragraph explaining that we will design Salt formulas to be idempotent (as we do for all formulas), making it easy to apply after a config change and "reconcile the desired state", and just loose the "reactive" aspect of it by making this design choice.

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+
+
+Implementation Details
+----------------------
+
+Iteration 1
+^^^^^^^^^^^
+
+- Define and deploy new ConfigMap stores that will hold cluster and service
+  configurations as listed in the requirements
```
gdemonet: Indeed. To provide a bit more details, maybe add:

> for each CM, define its schema, its default values, and how it impacts the configured services

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+    - Updating the Alertmanager notification configuration
+
+- Provide a UI interface for adding, updating and deleting service specific
+  configurations for example Dex-LDAP connector integration.
+- Provide a UI interface for listing MetalK8s available/supported Dex
+  authentication Connectors
+- Provide a UI interface for enabling or disabling Dex authentication
+  connectors (LDAP, Active Directory, StaticUser store)
+- Provide a UI interface for changing the number of replicas for a chosen set
+  of MetalK8s deployments (Prometheus, ....)
```
gdemonet:
```suggestion
  of MetalK8s deployments (Prometheus, ...)
```

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,205 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+- Add a UI interface for listing Alertmanager notification systems MetalK8s
```
gdemonet, line 247: It's not that "MetalK8s will support", we want to allow anything Alertmanager supports, IMO.

The comment from Nicolas was just that HipChat is EOL, so no need to mention it as an example (no one will use it).

gdemonet (commented):
Almost perfect! Good job ;)

docs/developer/architecture/configurations.rst
Ebaneck: As earlier indicated by Nicolas, these are cases that our design does not cater for and it is not required state the solution we want to use.

docs/developer/architecture/configurations.rst
Ebaneck: Config changes in this design and our current approach are not handled by salt I think...it is a Kubernetes thing right?? (kubectl edit). Also, state reconciliation is done by Kubernetes.. no??

docs/developer/architecture/configurations.rst
Ebaneck: We are talking about 9 different Alertmanager channels for notifications and i am not sure we want to support all those IMHO.

So, i prefer to leave it generic

docs/developer/architecture/configurations.rst
gdemonet: Sorry if I wasn't clear. In the current approach, we have:

- Salt creates ConfigMaps for all services with default values during bootstrap
- Salt deploys services with templated values coming from those default ConfigMaps
- Admin sets the desired configurations by `kubectl edit cm ...` if required (or through CLI/UI)

_At that time, the deployed services are still using default values_

- Admin requests (which a CLI/UI could do after editing the ConfigMap) a re-deployment of the desired service, for Salt to orchestrate

This implies a two-step process (whereas the Operator approach only requires the CM edition to have changes applied), *unless* we implement a reactive behaviour (through Beacon+Reactor pairs) directly in Salt master(s). _(this makes for a nice benefit of ConfigMaps vs. CRDs, as the former can be mounted as files, and reacting to file changes is rather easy with Salt)_

TeddyAndrieux: Personnaly prefer to have operator instead of beacon + reactor for this (even if for first iteration having a two-step process with salt is better/simpler IMO)

gdemonet: Well, we can easily transform Helm charts into Operators using Operator SDK, could be worth looking into.

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,197 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+
+**Using Salt states**
+
+Once a ConfigMap is updated by the user (say a user changes the number of
+replicas for Prometheus deployments to a new value), then perform the
+following actions:
+
+  - Apply a Salt states that reads the ConfigMap object, validates the schema
+    and checks the new values passed.
+
```
thomasdanan, line 116: @Ebaneck so If I understand well when we want to change one CSC, manually, it will be done in 2 steps: changing the conf and then applying the conf right?
I guess for some configurations, that may happen frequently (like setting alert notification), a UI or a CLI will automate those 2 steps?

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,197 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+Non-goals
+~~~~~~~~~
+
+This section contains requirements stated above which the current design choice
+does not cater for and will be addressed later:
+
+- Persisting newly added Grafana dashboards or new Grafana datasources
+  especially for modifications added via the Grafana UI cannot be stored in
+  ConfigMaps and hence will be catered for later.
```
thomasdanan, line 175: So what will be the solution for this?

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,197 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+
+- As stated in the requirements, adding and editing Prometheus alert rules
+  is also not covered by the chosen design choice and will need to be addressed
+  differently. Even if we could use ConfigMaps for Prometheus rules, we prefer
+  relying on the Prometheus Operator and it's CRD (PrometheusRule).
```
thomasdanan, line 180: Does it mean someone wanting to change the threshold of one rule can do it but will lose its configuration after an Upgrade? what solution do we suggest for this specific case? Is alert notification configuration is one of the things that is not supported by this design?

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,197 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+
+Documentation
+-------------
+
+In the Operational Guide:
+
+* Document how to customize or change any given service settings using the CLI
+  tool
+* Document how to customize or change any given service settings using the UI
+  interface
```
thomasdanan, line 260: So If it is supported through a UI, can we say it is supported through the API? (and not only CLI). Again if changing configuration is done in 2 steps, first step would be through K8s API, to change config map and second one will be through Salt API to apply the change?

docs/developer/architecture/configurations.rst
Ebaneck: The quickest approach will be to provision persistent volumes for Grafana just like we did for Prometheus & Alertmanager. The effort of doing this is really minimal.

docs/developer/architecture/configurations.rst
Ebaneck: Alertmanager notification configurations are handled in our current design since we can store whatever notification configuration ( e.g slack, email) we wish to implement in ConfigMaps. PS We have a ticket for storing the Alertmanager notification configurations.

For changing those thresholds, we found out that you can actually do it in the original PrometheusRule CRD deployed by the Prometheus-operator and I want to believe the Crds are persisted even after upgrades.
We just need to provide adequate documentation explaining how this can be done.

Remember that this design doc is about using ConfigMaps to store and persist Service & Cluster configurations. Persisting or changing Prometheus rules is not handled using ConfigMaps and hence we were advised not to add that particular solution in this document.

docs/developer/architecture/configurations.rst
```diff
@@ -73,3 +73,197 @@ As a MetalK8s expert, I can use ``kubectl`` command(s) in order to edit all
...
+* Document the list of service settings which can be configured by the user
+
+Test Plan
```
thomasdanan, line 263: Shouldn't you check that an upgrade of MetalK8s doesn't break previous customizations?

Ebaneck: Indeed, we should

docs/developer/architecture/configurations.rst
gdemonet: > may happen frequently (like setting alert notification)

"Frequently" here is still closer to monthly than daily IMO, and so gathering both steps into one shouldn't be a top priority. That being said, it's entirely achievable to have those grouped in a single "action", be it in CLI (once #2163 is done) or GUI.

docs/developer/architecture/configurations.rst
gdemonet: > The effort of doing this is really minimal.

Adding a PV is indeed no effort on our side. However, be it for Grafana or the rest of our services which use persistent storage, I'm not sure how we'll handle upgrade (and, if required, migration of their data). This question may not be as trivial to solve as "adding a PV".

docs/developer/architecture/configurations.rst
gdemonet: > do it in the original PrometheusRule CRD deployed by the Prometheus-operator and I want to believe the Crds are persisted even after upgrades.

CRDs belong to the Operator - a cluster admin could add rules by creating `PrometheusRule` objects, as he would add volumes by creating `Volume` objects.
Those *new* objects are not known by us, so we will leave them untouched under all circumstances.

> we were advised not to add that particular solution in this document.

At least in this first design PR. It may make sense to group all possible solutions to the general "configuration" problem under a single umbrella, but for now, we focus on what's missing: said ConfigMaps.

docs/developer/architecture/configurations.rst
gdemonet: Indeed, v1 of the docs should document the ConfigMaps, how to edit them (`kubectl edit` / `patch` or directly using the HTTP API), and how to apply changes using Salt.

Ebaneck: True and should be straight forward to handle in the docs
E.g

i. Configmap: metalk8s-dex-config

ii. How to edit it?

`kubectl edit configmaps metalk8s-dex-config -n metalk8s-auth`

ii. Apply the changes to take effect?

`salt-run state.sls metalk8s.addons.dex.deployed.chart saltenv=metalk8s-2.5.0-dev`

`kubectl rollout restart deployment dex -n metalk8s-auth`

TeddyAndrieux: > kubectl rollout restart deployment dex -n metalk8s-auth

Why ?

Ebaneck: Well, it looks like we need to forcefully restart the deployments...

Switching my dex replicas count from (2) to (1) and then applying the state didn't seem to change anything until I restarted the deployment.

thomasdanan (dismissed):

thomasdanan (approved):

PR #2249, opened 2020-02-19
test: add requests library retry to authentication test in order to reduce flakiness

Ebaneck:
**Component**:

'test'

**Context**: 

see #2229 

**Summary**:

This PR adds a retry logic at the requests library level to the authentication test-suite ensuring that we retry requests errors originating from internal server errors(500) & service unavailability(503).

**Acceptance criteria**: 

- Burn authentication test flaky

---

Closes: #2229

PR #2251, opened 2020-02-20
test: refactor Requests retry helper and make it generic

Ebaneck:
**Component**:

'test'

**Context**: 

See: https://github.com/scality/metalk8s/pull/2249#discussion_r381489467

**Summary**:

Made the requests retry helper generic by declaring it in `tests/utils.py` before importing it into the authentication test.

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -5,6 +5,10 @@ import testinfra
 import time
 from typing import Optional, Dict
 
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
 import pytest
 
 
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -61,3 +65,34 @@ def get_node_name(nodename, ssh_config=None):
...
+    status_forcelist=(500, 503),
+    method_whitelist=frozenset(['GET', 'POST']),
+    session=None
+):
+    """Request library retry helper
+
+    This helper performs 3 retries with an exponential sleep interval between
+    each request.
+
+    We only retry internal server errors(500) & service unavailable errors(503)
```
gdemonet: Those are default values, given this function, it seems we can override many arguments. Docstring should explain this, and maybe explain the argument values?

tests/utils.py
```diff
@@ -61,3 +65,34 @@ def get_node_name(nodename, ssh_config=None):
                 'salt-call --local --out txt grains.get id | cut -c 8-'
             )
     return nodename
+
+
+# Source: https://www.peterbe.com/plog/best-practice-with-retries-with-requests
+
```
slaperche-scality: you can remove this empty line

tests/utils.py
```diff
@@ -61,3 +65,34 @@ def get_node_name(nodename, ssh_config=None):
...
+    method_whitelist=frozenset(['GET', 'POST']),
+    session=None
+):
+    """Request library retry helper
+
+    This helper performs 3 retries with an exponential sleep interval between
+    each request.
+
+    We only retry internal server errors(500) & service unavailable errors(503)
+
```
slaperche-scality: you can remove this empty line

tests/utils.py
slaperche-scality: This is the default behavior (so you could start the sentence with "By default this helper…"

Also would be nice to have doc for the arguments, something like that:

```python
    """Configure a `requests.Session` for retry on error.

    By default this helper performs 3 retries with an exponential sleep interval
    between each request and only retry internal server errors(500) & service
    unavailable errors(503).

    Arguments:
        retries:          […]
        backoff_factor:   […]
        status_forcelist: […]
        method_whitelist: […]
        session:          […]

    Returns:
        A `requests.Session` object configured for retry.
    """
```

gdemonet (changes requested):
Small change on the docstring to make it more obvious that this helper is configurable.

If you really want to make it super easy to use, you could make it a Pytest fixture:

```python
@pytest.fixture(scope="function")
def request_retry_session(request):
    # Callers can inject arguments using `pytest.mark.parametrize`
    params = getattr(request, 'param', {})

    return utils.request_retry_session(**params)


def test_flaky_request(request_retry_session):
    resp = request_retry_session.get("/flaky")
    assert resp == 'OK'


@pytest.mark.parametrize('request_retry_session', [
    {'retries': 1, 'backoff_factor': 0.5},
], indirect=True)
def test_stable_request(request_retry_session):
    resp = request_retry_session.get("/stable")
    assert resp == 'OK'
```

Although I think I remember using `pytest.mark.parametrize` for fixtures in `pytest-bdd` scenarii is kind of painful...

tests/utils.py
Ebaneck: Done

tests/utils.py
gdemonet: Still here...

gdemonet (changes requested):

slaperche-scality (approved):

gdemonet (approved):

Issue #2084, opened 2019-11-27
Storage Operator ClusterRole has too much permissions


gdemonet:
**Component**: storage-operator

**Why this is needed**: We don't want to setup permissions for components that don't need them specifically, as a good practice (wouldn't want a bug in there that creates turmoil in the cluster for no good reason).

**What should be done**: Remove all unneeded permissions from the ClusterRole defined in `storage-operator/deploy/role.yaml`.

PR #2256, opened 2020-02-24
storage-operator: upgrade to `operator-sdk` 0.15.2

NicolasT:


eve/workers/pod-linter/Dockerfile
```diff
@@ -1,19 +1,21 @@
 # CI lint container - Fedora for recent ShellCheck version
-FROM fedora:29
+FROM fedora:31
```
slaperche-scality, line 2: this bump (and the install of `git` and `findutils`) is related to the bump of the operator SDK?

If yes, would be nice to have a little explanation on why those new reqs in the commit message.
If no, could be done in a dedicated commit.

storage-operator/cmd/manager/main.go
```diff
@@ -79,14 +82,18 @@ func main() {
...
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace:               namespace,
-		MapperProvider:          restmapper.NewDynamicRESTMapper,
-		MetricsBindAddress:      fmt.Sprintf("%s:%d", metricsHost, metricsPort),
-		LeaderElection:          true,
-		LeaderElectionNamespace: "kube-system",
-		LeaderElectionID:        "storage-operator-lock",
+		Namespace:          namespace,
+		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
```
slaperche-scality, line 96: IIRC this is reverting bff55c256fff5dff9525f86bd0b6b7dfbc4735b7

If this is done on purpose, this should be at least justified in the commit message.
If this an error, let's fix it :wink:

storage-operator/deploy/role.yaml
```diff
@@ -10,14 +10,21 @@ rules:
...
+  - services/finalizers
   - endpoints
   - persistentvolumes
   - persistentvolumeclaims
   - events
   - configmaps
   - secrets
   verbs:
-  - '*'
+  - create
```
slaperche-scality, line 21: Seems like this started to tackle https://github.com/scality/metalk8s/issues/2084 as a side effet :slightly_smiling_face:

storage-operator/go.mod
```diff
@@ -1,35 +1,44 @@
...
-	k8s.io/client-go => k8s.io/client-go v0.0.0-20190228174230-b40b2a5939e4
+	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.4.0
+	k8s.io/api v0.0.0
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v12.0.0+incompatible
+	sigs.k8s.io/controller-runtime v0.4.0
 )
 
+// Pinned to kubernetes-1.16.2
```
slaperche-scality, line 17: Should this be pinned to different version of k8s depending on version of MetalK8s? Or we don't care?

storage-operator/pkg/apis/storage/v1alpha1/volume_types.go
```diff
@@ -103,20 +101,19 @@ type VolumeCondition struct {
...
 
 // VolumeStatus defines the observed state of Volume
-// +k8s:openapi-gen=true
 type VolumeStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
 
 	// List of conditions through which the Volume has or has not passed.
-	// +kubebuilder:validation:Enum=Available,Pending,Failed,Terminating
```
slaperche-scality, line 119: I would almost fix that in a dedicated commit, since it's a bugfix and not exactly related to the upgrade per se.

buildchain/buildchain/codegen.py
```diff
@@ -24,7 +24,7 @@ def codegen_go() -> types.TaskDict:
     """Generate Go code using operator-sdk."""
     cwd  = constants.STORAGE_OPERATOR_ROOT
     actions = []
-    for target in ('k8s', 'openapi'):
+    for target in ('k8s', 'crds'):
```
slaperche-scality, line 27: ho, so we don't have to call `openapi-gen` explicitly like Benoit did?

(I thought it was required to update the `zz_generated.openapi.go`)

slaperche-scality (changes requested):
Possible regression on leader election

buildchain/buildchain/codegen.py
NicolasT: FWIW, IIUC in general you don't need those `zz_generated.openapi.go` files at all.

storage-operator/go.mod
NicolasT: I'm going to close this PR eventually in favor of #2544, but to answer the question: no, it doesn't. This is a *client* version, so if we'd be using APIs/types in *our* code that doesn't work with a deployed version of K8s anyway, we'd be stuck.

NicolasT:
Closing in favor of #2544.

Issue #2255, opened 2020-02-24
Template and store the number of replicas for Grafana, Prometheus and Alertmanager as service configurations


Ebaneck:
**Component**:

'salt', 'charts', 'kubernetes'

**Why this is needed**:

To guarantee the persistence of the number replicas for Grafana, Alertmanager and Prometheus deployments, we need to store these service configurations using a ConfigMap as documented in the design phase(see #2233).

**What should be done**:

- Template and render the service charts(Prometheus-operator) to read cluster and service configurations from ConfigMaps directly.
- Add a salt module that retrieves the number of replicas from a specified Configmap
- Deploy the ConfigMap holding the respective service configurations before the consumers are deployed

**Implementation proposal** (strongly recommended):

**Test plan**:

Epic: #1996

PR #2258, opened 2020-02-25
Template and store replicas count as service configuration in Metalk8s cluster

Ebaneck:
**Component**:

'salt', 'charts', 

**Context**: 

See #2255 

**Summary**:

Template and render the charts to consume service configuration values directly.

**Acceptance criteria**: 

---

Closes: #2255


charts/render.py
```diff
@@ -114,25 +120,45 @@ def keep_doc(doc):
...
+    parser.add_argument('values', help="Our custom chart values")
+    parser.add_argument(
+        '--service-config',
+        action='append',
+        nargs=2,
+        dest="service_config",
+        help="Ex: --service-config service_name service_configmap_name"
+    )
+    parser.add_argument('path')
+    args, extra = parser.parse_known_args()
```
gdemonet: Not sure you want to allow (and ignore) extra arguments. Just using `args = parser.parse_args()` will fail if unknown arguments are provided.

charts/render.py
```diff
@@ -114,25 +120,45 @@ def keep_doc(doc):
 
 
 def main():
-    (name, namespace, values, path) = sys.argv[1:]
+    parser = argparse.ArgumentParser()
+    parser.add_argument('name', help="Denotes the name of the chart")
+    parser.add_argument('namespace', help="Metalk8s namespace to deploy unto")
```
gdemonet: Let's make this optional:

```suggestion
    parser.add_argument(
        '-n',
        '--namespace',
        default="default",
        help="Namespace to deploy this chart in",
    )
```

charts/render.py
```diff
@@ -27,17 +28,22 @@
 from yaml.representer import SafeRepresenter
 
 
-BOILERPLATE = '''
+BOILERPLATE = """
```
slaperche-scality: Even though I personally prefer `"""` over `'''`, let's be consistent here and use `'''` (since  `SUB_BOILERPLATE` and `SUB_BOILERPLATE_END` already use `'''`)

charts/render.py
```diff
@@ -114,25 +120,45 @@ def keep_doc(doc):
...
+    parser.add_argument('namespace', help="Metalk8s namespace to deploy unto")
+    parser.add_argument('values', help="Our custom chart values")
+    parser.add_argument(
+        '--service-config',
+        action='append',
+        nargs=2,
+        dest="service_config",
+        help="Ex: --service-config service_name service_configmap_name"
+    )
+    parser.add_argument('path')
```
gdemonet: Let's add some help as well:

```suggestion
    parser.add_argument('path', help="Path to the chart directory")
```

charts/render.py
```diff
@@ -114,25 +120,45 @@ def keep_doc(doc):
...
+    parser.add_argument('name', help="Denotes the name of the chart")
+    parser.add_argument('namespace', help="Metalk8s namespace to deploy unto")
+    parser.add_argument('values', help="Our custom chart values")
+    parser.add_argument(
+        '--service-config',
+        action='append',
+        nargs=2,
+        dest="service_config",
+        help="Ex: --service-config service_name service_configmap_name"
+    )
```
gdemonet: Cool stuff! Could add a small explanation about what those "service-config"s are in the help message (and adjust the example to make it understandable that `service_name` is actually the Jinja variable which will hold this ConfigMap's contents).

(minor) To make it even friendlier to use/read in code, you can cast parsed arguments into arbitrary types:

```python
from collections import namedtuple

ServiceConfig = namedtuple('ServiceConfig', ('var_name', 'configmap_name')

# When building the parser...
parser.add_argument(
    '--service-config',
    action='append',
    nargs=2,
    type=ServiceConfig,
    dest="service_configs",
    help="...",
)

# When printing the Jinja declarations...
sys.stdout.write(
    BOILERPLATE.format(
        configlines='\n'.join(
            ("{{%- set {s.var_name} = salt.metalk8s_service_configurator"
             ".get_service_conf('{namespace}', '{s.configmap_name}') %}}").format(
                 s=service_config, namespace=args.namespace
            ) for service_config in args.service_configs
        )
    ).lstrip()
)
```

Note that you don't need to inject curly braces with `format` arguments, simply double them (`{{` or `}}`) to escape the characters. (cf. https://docs.python.org/library/string.html#formatstrings)

charts/render.py
```diff
@@ -114,25 +120,45 @@ def keep_doc(doc):
...
-    sys.stdout.write(BOILERPLATE.lstrip())
+    sys.stdout.write(
+        BOILERPLATE.format(
+            configlines='\n'.join(
+                "{} set {} = salt.metalk8s_service_configurator.get_service_conf('{}', '{}') {}".format(
+                    '{%-', p[0], args.namespace, p[1], '%}'
+                ) for p in args.service_config)
+        ).lstrip()
+    )
+    sys.stdout.write(SUB_BOILERPLATE)
```
gdemonet: Not sure the split between `BOILERPLATE` and `SUB_BOILERPLATE` is needed or helpful. Actually, since we now template in some variable declarations, it's not only boilerplate anymore. Maybe just call them `[START|END]_TEMPLATE` or `[START|END]_BLOCK`?

gdemonet (dismissed):

charts/render.py
Ebaneck: Well, we redirect the output to a file using something like `> salt/metalk8s/addons/prometheus-operator/deployed/chart.sls`

We actually want to omit this extra argument

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        A dict of a specific service configuration
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_service_configurator.service_conf "metalk8s-monitoring" "metalk8s-prometheus-config"
+    """
+
+    if configmap_name is None:
```
TeddyAndrieux:
```suggestion
    if not configmap_name:
```
If user give `0` or `''` as name you want to raise the  same message

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        conf_section = manifest.get('data', {}).get('config.yaml', {})
+        config = yaml.safe_load(conf_section) or {}
+    except Exception as exc:
+        raise CommandExecutionError(
+            'Failed to get Config section {} as manifest with error {}'
+            .format(configmap_name, str(exc))
+        )
+    if not config:
+        raise CommandExecutionError(
+            'Expected configmap as yaml but got {}'.format(config)
```
TeddyAndrieux: I got what you mean but I think the message is not really clear maybe something like

```suggestion
            'Expected `config.yaml` as yaml in the ConfigMap {} but got {}'.format(configmap_name, config)
```

buildchain/buildchain/salt_tree.py
```diff
@@ -221,6 +221,10 @@ def _get_parts(self) -> Iterator[str]:
         renderer=targets.Renderer.JSON,
     ),
 
+    Path('salt/metalk8s/addons/addons-service-config/configmap.sls'),
+    Path('salt/metalk8s/addons/addons-service-config/deployed.sls'),
+    Path('salt/metalk8s/addons/addons-service-config/init.sls'),
```
TeddyAndrieux: You should move these files, we said that state that should only run on the salt-master need to be called `deployed`.
Here you can call `addons.addons-service-config` and `addons.addons-service-config.configmap` and both will deploy kubernetes objects.

So here you should have
```
salt/metalk8s/addons/addons-service-config/deployed/init.sls
salt/metalk8s/addons/addons-service-config/deployed/configmap.sls
```
And that it, then you will be able to do `state.sls metalk8s.addons.addons-service-config.deployed`

Also IMO the `addons` in `addons-service-config` is a bit redundant

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
+'''Metalk8s cluster and service configuration utility.'''
+
+import logging
+import re
+
+from salt.exceptions import CommandExecutionError
+
+import yaml
+from yaml.dumper import SafeDumper
+from yaml.representer import SafeRepresenter
```
TeddyAndrieux: Look like these are not really used here

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
+'''Metalk8s cluster and service configuration utility.'''
+
+import logging
+import re
```
TeddyAndrieux: Same, not used here

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+
+from salt.exceptions import CommandExecutionError
```
TeddyAndrieux: Please respect pylint import suggestion
```
C:  8, 0: third party import "import yaml" should be placed before "from salt.exceptions import CommandExecutionError" (wrong-import-order)
```

charts/render.py
TeddyAndrieux: The redirect is catch by your shell directly and will **not** be part of the arguments python will get

salt/_modules/metalk8s_service_configurator.py
slaperche-scality: Maybe we should setup pylint for Salt tree (or `_modules` at least)?
That would automatically flag unused import and import ordering.

That way we can enforce it in CI and focus our brain's bandwidth on functional issues.

buildchain/buildchain/salt_tree.py
NicolasT: Can we have these *under the applicable addons or whatnot directories* rather than somewhere completely else? This is not an 'addon'.

buildchain/buildchain/salt_tree.py
```diff
@@ -552,6 +556,7 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/_modules/metalk8s_package_manager_yum.py'),
     Path('salt/_modules/metalk8s_package_manager_apt.py'),
     Path('salt/_modules/metalk8s_volumes.py'),
+    Path('salt/_modules/metalk8s_service_configurator.py'),
```
NicolasT: Is it a `configurator` or is it about `service_configuration`?

charts/render.py
NicolasT:
```suggestion
    parser.add_argument('namespace', help="MetalK8s namespace to deploy into")
```
Also, this is not strictly a `MetalK8s` namespace.

salt/_modules/metalk8s_service_configurator.py
TeddyAndrieux: Agree we should but we need some extra stuff about "salt" stuffs like all the `__opts__` `__salt__` `__pillar__` ... but anyway agree we should have salt module checking in the CI so linting + unit test

charts/render.py
NicolasT: I'm quite confused by this naming as well.

salt/_modules/metalk8s_service_configurator.py
gdemonet: If this is something we really want to enforce, it would make sense to add it in CI. Especially since not everyone here uses `pylint` + the special Salt plugin to lint their Salt modules (I don't).

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_service_configurator.service_conf "metalk8s-monitoring" "metalk8s-prometheus-config"
+    """
+
+    if configmap_name is None:
+        raise Exception(
+            'Expected a configmap name but got {}'.format(configmap_name)
```
NicolasT:
```suggestion
            'Expected a ConfigMap name but got {}'.format(configmap_name)
```
Similar changes below. Be consistent.

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def get_service_conf(
+    namespace,
+    configmap_name,
+):
+    """Reads  a ConfigMap from the name specified
+
+    Arguments:
+        configmap_name: Configmap name
```
NicolasT:
```suggestion
        configmap_name: ConfigMap name
```

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        namespace: the namespace where the configmap is found
```
NicolasT:
```suggestion
        namespace: the Namespace where the ConfigMap is stored
```

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+            'metalk8s_kubernetes.get_object'
+        ](
+            kind='ConfigMap',
+            apiVersion='v1',
+            namespace=namespace,
+            name=configmap_name
+        )
+    except ValueError as exc:
+        raise CommandExecutionError(
+            'Failed to read Configmap object {} with error {}'
```
NicolasT:
```suggestion
            'Failed to read ConfigMap object {}: {}'
```

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+            .format(configmap_name, str(exc))
```
NicolasT: Use a modifier in the template instead of `str` here.

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    if manifest is None:
+        raise CommandExecutionError(
+            'Expected Configmap object but got {}'.format(manifest)
```
NicolasT:
```suggestion
            'Expected ConfigMap object but got {}'.format(manifest)
```

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    try:
+        # read the dict object
+        # we are only interested in the data section of the configmap for now
+        conf_section = manifest.get('data', {}).get('config.yaml', {})
+        config = yaml.safe_load(conf_section) or {}
+    except Exception as exc:
```
NicolasT: Do we really need to catch `Exception`?

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        raise CommandExecutionError(
+            'Failed to get Config section {} as manifest with error {}'
```
NicolasT:
```suggestion
            'Failed to get config section {}: {}'
```

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+            .format(configmap_name, str(exc))
```
NicolasT: Same as above, use a modifier in the template, no `str`.

salt/_modules/metalk8s_service_configurator.py
slaperche-scality: I thin there is a pylint plugin for Salt, otherwise we can simply ignore Salt stuff by tweaking Pylint conf (you can probably consider some stuff as blackbox).

salt/_modules/metalk8s_service_configurator.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+    if not config:
+        raise CommandExecutionError(
+            'Expected configmap as yaml but got {}'.format(config)
+        )
+    return config  # a classic python dict
```
NicolasT: Too early. Could as well do some super basic validation of this value. Even more, we may want to pass the `apiVersion` and `kind` of the object we expect (at call site) and assert these are 'correct'. And over time, do proper validation of the whole structure.

salt/metalk8s/addons/addons-service-config/configmap.sls
```diff
@@ -0,0 +1,41 @@
+#! metalk8s_kubernetes
```
NicolasT: As mentioned before, these shouldn't go here.

salt/metalk8s/addons/addons-service-config/configmap.sls
```diff
@@ -0,0 +1,41 @@
...
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: metalk8s-monitoring
+  name: metalk8s-grafana-config
+data:
+  config.yaml: |-
+    apiVersion: metalk8s.scality.com/v1alpha1
```
NicolasT: We may want to give these their own group. WDYT @gdemonet

salt/metalk8s/addons/addons-service-config/deployed.sls
```diff
@@ -0,0 +1,5 @@
+# Create the namespace before deploying configmap objects
```
NicolasT: This is exactly why the default configs belong with the addons and not here.

NicolasT (dismissed):

salt/_modules/metalk8s_service_configurator.py
TeddyAndrieux: https://github.com/saltstack/salt-pylint (Personnaly never tested)

salt/_modules/metalk8s_service_configurator.py
gdemonet: (for those who don't know what this means:

```python
'Failed to read ConfigMap object {}: {!s}'.format(configmap_name, exc)
```
)

salt/metalk8s/addons/addons-service-config/configmap.sls
gdemonet: `addons.metalk8s.scality.com`? Or even more specialized, like `<addon-name>.addons.metalk8s.scality.com`?

salt/metalk8s/addons/addons-service-config/deployed.sls
Ebaneck: ok

salt/_modules/metalk8s_service_configurator.py
slaperche-scality: I think @jbertran tried it (but I may remember wrong)

charts/render.py
Ebaneck: ok

buildchain/buildchain/salt_tree.py
Ebaneck: service_configuration

salt/_modules/metalk8s_service_configurator.py
Ebaneck: True! no need to catch it

gdemonet: You do need to catch some exception, in case loading of the ConfigMap as YAML fails. Nicolas' point was: don't use bare `Exception`, look for something more specific (like `yaml.YAMLError`).

Leaving something like:
```python
except Exception:
    pass
```

is worse than not doing anything (otherwise your code will break in very strange ways).

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  namespace: metalk8s-monitoring
+  name: metalk8s-alertmanager-config
+data:
+  config.yaml: |-
+    apiVersion: addons.metalk8s.scality.com
+    kind: AlertmanagerConfig
+    spec:
+      deployment:
+        replicas: 1
```
TeddyAndrieux: IMO we should do some jinja here to check that these ConfigMap not already exists to not override customers changes if this state run (and then when we have this, this state could be included in the `init.sls` of each addons)

Ebaneck: Done!

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -133,6 +133,13 @@ Sync module on salt-master:
     - name: saltutil.sync_all
     - saltenv: metalk8s-{{ dest_version }}
 
+Deploy Kubernetes service config objects:
+  salt.runner:
+  - name: state.orchestrate
+  - mods:
+    - metalk8s.service-configuration.deployed
+  - saltenv: metalk8s-{{ dest_version }}
```
TeddyAndrieux, line 149: You need to add some checks here because you don't want to do this if `dest_version < 2.5.0`

salt/metalk8s/service-configuration/deployed/init.sls
```diff
@@ -0,0 +1,10 @@
+#
+# States to deploy cluster and service configurations
+#
+
+# Create the Namespace before deploying ConfiMap objects
+# If we include service configurations in their respective addons deploy states,
+# then salt will fail while rendering the jinja since the Namespace object does
```
TeddyAndrieux: I think you should say that is that '"ConfigMap' does not exist yet" even if the first error you get is because the namespace do not exist the issue here is that the CM do not exist (even if you create the namespace you will get an error because the ConfigMap do not exist)

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -133,6 +133,23 @@ Sync module on salt-master:
     - name: saltutil.sync_all
     - saltenv: metalk8s-{{ dest_version }}
 
+{#- In previous MetalK8s releases, Cluster and Service configurations where not
+    persisted throughout. Now in `2.5.0` we have added ConfigMaps to store
+    service configurations. When running a downgrade to a version inferior to
+    `2.5.0` we do not need to deploy these ConfigMap objects. #}
+{#- Only run this step if `dest_version < 2.5.0` #}
```
TeddyAndrieux: No, we want to run it only if `dest_version >= 2.5.0` :)

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -133,6 +133,23 @@ Sync module on salt-master:
...
+
+{%- if salt.pkg.version_cmp(dest_version, '2.5.0') == -1  %}
```
TeddyAndrieux:
```suggestion
{%- if salt.pkg.version_cmp(dest_version, '2.5.0') >= 0  %}
```

salt/metalk8s/orchestrate/downgrade/init.sls
Ebaneck: Thinking this still works because we would have already deployed the ConfigMaps in the Bootstrap orchestrate before the downgrade.

salt/metalk8s/orchestrate/downgrade/init.sls
Ebaneck: Done

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -0,0 +1,105 @@
...
+        raise CommandExecutionError(
+            'Expected ConfigMap object but got {}'.format(manifest)
+        )
+
+    try:
+        conf_section = manifest.get('data', {}).get('config.yaml', {})
+        config = yaml.safe_load(conf_section) or {}
+    except yaml.YAMLError as exc:
+        raise CommandExecutionError(
+            'Invalid YAML format in Configmap {}: {!s}'.format(
```
TeddyAndrieux:
```suggestion
            'Invalid YAML format in ConfigMap {}: {!s}'.format(
```

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -0,0 +1,105 @@
...
+                configmap_name, exc
+            )
+        )
+    except Exception as exc:
+        raise CommandExecutionError(
+            'Failed loading `config.yaml` from Configmap {}: {!s}'.format(
```
TeddyAndrieux:
```suggestion
            'Failed loading `config.yaml` from ConfigMap {}: {!s}'.format(
```

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -0,0 +1,105 @@
...
+                config['apiVersion'], apiVersion
+            )
+        )
+
+    if config['kind'] != kind:
+        raise CommandExecutionError(
+            'Expected value {} for key kind, got {}'.format(
+                config['kind'], kind
+            )
+        )
```
TeddyAndrieux: I don't really get the logic here, because you return config if `kind` and `apiVersion` are both None, but if only one is None that will fail.

If I well understand what you want to do is if `kind` or `apiVersion` is `None` just dont check it 
If yes then you need to also check for `None` in the 2 last `if` because one of `apiVersion` or `kind` could be `None`

```suggestion
    if not apiVersion and not kind:
        return config

    if apiVersion and config['apiVersion'] != apiVersion:
        raise CommandExecutionError(
            'Expected value {} for key apiVersion, got {}'.format(
                apiVersion, config['apiVersion']
            )
        )

    if kind and config['kind'] != kind:
        raise CommandExecutionError(
            'Expected value {} for key kind, got {}'.format(
                kind, config['kind']
            )
        )
```

Also, it's only my opinion but for me it do not hurt to check for empty string or 0 ... for `apiVersion` and `kind` not only for `None`

TeddyAndrieux (commented):
Few minor comment but LGTM

salt/_modules/metalk8s_service_configuration.py
Ebaneck: This is getting out of hand.. Going to fix

salt/_modules/metalk8s_service_configuration.py
Ebaneck: Accepted and will fix after discussing it

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -0,0 +1,104 @@
...
+            .format(configmap_name, config)
+        )
+
+    if not apiVersion and not kind:
+        return config
+
+    if apiVersion and config['apiVersion'] != apiVersion:
+        raise CommandExecutionError(
+            'Expected value {} for key apiVersion, got {}'.format(
+                config['apiVersion'], apiVersion
```
TeddyAndrieux:
```suggestion
                apiVersion, config['apiVersion']
```
Like I told you on slack, you expect the `apiVersion` args from this function, not the one from configmap

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -0,0 +1,104 @@
...
+            )
+        )
+    if kind and config['kind'] != kind:
+        raise CommandExecutionError(
+            'Expected value {} for key kind, got {}'.format(
+                config['kind'], kind
```
TeddyAndrieux:
```suggestion
                kind, config['kind']
```

TeddyAndrieux (changes requested):

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -133,6 +133,23 @@ Sync module on salt-master:
...
+{#- Only run this step if `dest_version >= 2.5.0` #}
+
+{%- if salt.pkg.version_cmp(dest_version, '2.5.0') >= 0 %}
+
+Deploy Kubernetes service config objects:
+  salt.runner:
+  - name: state.orchestrate
+  - mods:
+    - metalk8s.service-configuration.deployed
+  - saltenv: metalk8s-{{ dest_version }}
```
TeddyAndrieux, line 149: Just add a require_in on Deploy kubernetes objects here (you don't want to run the kubernetes object deployment if this state failed)

And also a require on Sync module on salt-master

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
```diff
@@ -0,0 +1,83 @@
...
+          name: metalk8s-grafana-config
+          namespace: metalk8s-monitoring
+        data:
+          config.yaml: |-
+            apiVersion: addons.metalk8s.scality.com
+            kind: GrafanaConfig
+            spec:
+              deployment:
+                replicas: 1
+{%- endif %}
```
TeddyAndrieux, line 50: An sls cannot be empty or you will get an error (or maybe just a warning don't remember) and here if all config already exists after jinja rendering you will get empty sls.

So suggest to just add an `else` block with `test.succeed_without_changes` explaining that the configmap already exists nothing need to be done

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -115,6 +115,13 @@ Sync module on salt-master:
     - name: saltutil.sync_all
     - saltenv: metalk8s-{{ dest_version }}
 
+Deploy Kubernetes service config objects:
+  salt.runner:
+  - name: state.orchestrate
+  - mods:
+    - metalk8s.service-configuration.deployed
+  - saltenv: metalk8s-{{ dest_version }}
```
TeddyAndrieux, line 123: Require Sync module on salt-master here

TeddyAndrieux (changes requested):

TeddyAndrieux (approved):
LGTM

Issue #2222, opened 2020-02-05
Upgrade/downgrade tests from development branches


alexandre-allard:
**Component**: build, ci

**What happened**:
Sometimes, we have issues in the buildchain impacting all versions of the product (e.g. #2220).
This lead to a problem when trying to test the fix in the CI, because we're building current (from the branch with the fix) and previous (without the fix, which will obviously fail to build) versions to test downgrade/upgrade.
In this case we need to bypass the build status, to force the merge of the branch even if tests don't pass, which is definitely not what we want.

**What was expected**:
Be able to successfully build and test in CI, without having to bypass build status.

**Steps to reproduce**: n/a

**Resolution proposal** (optional):
Maybe we could build previous branches from the integration branches when there is one and fallback on development branch.
We could also fetch a previous build from artifacts, so we would also avoid rebuilding each time.
The best solution would be to only tests upgrade/downgrade paths with already released versions and maybe have some post-merge tests with the development branches.

alexandre-allard:
No longer needed as we no longer build previous/next versions in pre-merge tests anymore.

Issue #1993, opened 2019-11-05
Solution Lifecycle: tooling and solution example


thomasdanan:
Framework and tools to deploy solutions on top of MetalK8s cluster:

- Import solution
- Deploy Operator and UI
- UI to create Environment (i.e. namespace)
- Fined grain salt tooling
- Upgrade / Downgrade Operator and UI
- Solution Example

gdemonet:
This iteration is now complete. Next steps are all consigned in the [Solutions Lifecycle project](https://github.com/scality/metalk8s/projects/18).

PR #2265, opened 2020-02-26
docs: Add CI architecture document

TeddyAndrieux:
**Component**:

'docs'

**Context**: 

CI test

**Summary**:

Add a new architecture document for CI test, describing what we should
do in the CI and in which step (pre-merge, nightly, weekly, ...)

---

### DIFF
--- a/docs/developer/architecture/ci.rst
+++ b/docs/developer/architecture/ci.rst
@@ -1,0 +1,168 @@
+Continuous Testing
+==================
+
+
+This document will not describe how to write a test but just the list of tests
+that should be done and when.
+
+The goal is to:
+
+- have day-to-day development and PRs merged faster
+- have a great test coverage
+
+Lets define 2 differents stages of continuous testing:
+
+- Pre-merge: Run during development process on changes not yet merged
+- Post-merge: Run on changes already approved and merged in development
+  branches
+
+Pre-merge
+---------
+
+What should be tested in pre-merge on every branch used during development
+(``user/*``, ``feature/*``, ``improvement/*``, ``bugfix/*``, ``w/*``).
+The pre-merge test should not long too much time (less than 40 minutes
+is great) so we can't test everything in pre-merge, we should only test
+building of the product and check that product still usable.
+
+- Building tests
+
+  - Build
+  - Lint
+  - Unit tests
+
+- Installation tests
+
+  - Simple install RHEL
+  - Simple install CentOs + expansion
+
+When merging several pull requests at the same time, given that we are on a
+queue branch (``q/*``), we may require additional tests as a combination of
+several PRs could have a larger impact than all individual PR:
+
+- Simple upgrade/downgrade
+
+.. _post-merge:
+
+Post-merge
+----------
+
+On each and every ``development/2.*`` branches we want to run complex tests,
+that take more time or need more ressources than classic tests that run during
+pre-merge, to ensure that the current product continues to work well.
+
+Nightly
+^^^^^^^
+
+- Upgrade, downgrade tests:
+
+  - For previous development branch
+
+    e.g.: on ``development/2.x`` test upgrade from ``development/2.(x-1)``
+    and downgrade to ``development/2.(x-1)``
+
+    - Build branch ``development/2.(x-1)`` (or retrieve it if available)
+    - Tests:
+
+      - Single node test
+      - Complex deployment test
+
+  - For last released version of current minor
+
+    e.g.: on ``development/2.x`` when developing "2.x.y-dev" test upgrade from
+    ``metalk8s-2.x.(y-1)`` and downgrade to ``metalk8s-2.x.(y-1)``
+
+    - Single node test
+    - Complex deployment test
+
+  - For last released version of previous minor
+
+    e.g.: on ``development/2.x`` when developing "2.x.y-dev" test upgrade from
+    ``metalk8s-2.(x-1).z`` and downgrade to ``metalk8s-2.(x-1).z`` where
+    "2.(x-1).z" is the last patch released version for "2.(x-1)" (``z`` may be
+    different from ``y``)
+
+    - Single node test
+    - Complex deployment test
+
+- Backup, restore tests:
+
+  - Environment with at least 3-node etcd cluster, destroy the bootstrap node
+    and spawning a new fresh node for restoration
+  - Environment with at least 3-node etcd cluster, destroy the bootstrap node
+    and use one existing node for restoration
+
+- Solutions tests
+
+.. note::
+
+  Complex deployment is (to be validated):
+
+  - 1 bootstrap
+  - 1 etcd and control
+  - 1 etcd and control and workload
+  - 1 workload and infra
+  - 1 workload
+  - 1 infra
+
+.. todo::
+
+   - Describe solutions tests (#1993)
+
+Weekly
+^^^^^^
+
+More complex tests:
+
+- Performance/conformance tests
+- Validation of `contrib` tooling (Heat, terraform, ...)
+- Installation of "real" Solution (Zenko, ...)
+- Long lifecycle metalk8s tests (several upgrade, downgrade, backup/restore,
+  expansions, ...)
+
+.. todo::
+
+  Validate the list of Weekly test to do and define exactly what need to be
+  tested
+
+Adaptive test plan
+------------------
+
+CI pre-merge may be more flexible by including some logic about the content
+of the changeset.
+
+The goal here is to test only what needed according to the content of the
+commit.
+
+For example:
+
+- For a commit that changes uniquely documentation, we don't need to run the
+  entire installation test suite but rather run tests related to documentation.
+- For a commit touching upgrade orchestrate we want to test upgrade directly
+  in pre-merge and not wait :ref:`Post merge<post-merge>` build to get the
+  test result.
+
+.. todo::
+
+   Several questions:
+
+   - How to get the change of one commit ?
+
+     - Depending on the files changed
+
+       - How do you know when you change something in salt if this changeset
+         touch upgrade for example ?
+
+         - ...
+
+     - A tag in the commit message
+
+       - maybe ?
+
+   - How to get the bunch of commit to test ?
+
+     - Get commit between HEAD and target branch
+
+       - How to get this target ?
+
+         - ...
--- a/docs/developer/architecture/index.rst
+++ b/docs/developer/architecture/index.rst
@@ -5,3 +5,4 @@ Architecture Documents
 
    deployment
    requirements
+   ci


docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
+CI
+==
+
+Pre merge
+---------
```
Ebaneck: `Pre-merge` ??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+What should be tested in pre-merge on every ``user/*``, ``feature/*``,
+``improvement/*``, ``bugfix/*``, ``w/*`` branches.
```
Ebaneck: I think you could at least state one reason here why we should only perform the following test (Building tests & Installation tests) in Pre-merge. ??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+  - Lint
+  - Unit tests
+
+- Installation tests
+
+  - Simple install RHEL
+  - Simple install CentOs + expansion
+
+When merging several PRs at the same time, so when we are on a ``q/*``
+branches, we may want to have additionnal test like:
```
Ebaneck: Could be rewritten `When merging several pull requests at the same time, given that we are on a queue branch (`q/*`), we may want to have or we may require additional tests like:`

Ebaneck: And maybe add a reason why we require these additional tests??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+- Simple upgrade/downgrade
+
+Pre-merge shouldn't be too long as it run on every push on the feature
+branches.
+
+.. _post-merge:
+
+Post merge
```
Ebaneck: Post-merge ??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+----------
+
+On each and every ``development/2.*`` branches we want to run complex tests to
+ensure that the current developmnent product is well working.
```
Ebaneck: Change to `development` or simply have `On each and every ``development/2.*`` branches we want to run complex tests to ensure that the current product continues to work`??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+More complex tests (TBD)
+
+Adaptive test plan
+------------------
+
+CI pre-merge may be more flexible by including some logic about the content
+of the changeset.
+
+For example :
```
Ebaneck: Why the space before the colon (:)??

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+- for a commit changing only documentation we don't need/want to run
+  all the test about installation but just documentation part.
+- for a commit touching upgrade orchestrate we want to test upgrade directly
+  in pre-merge and not wait :ref:`Post merge<post-merge>` build to get the
```
Ebaneck: - Please begin the sentences in capitals (For)
- We could rewrite the first sentence to read
`For a commit that changes solely/uniquely documentation, we don't need/want to run all the installation test but rather run tests related to documentation`

Ebaneck (changes requested):

docs/developer/architecture/ci.rst
TeddyAndrieux: Hmm yes true I will add it

docs/developer/architecture/ci.rst
TeddyAndrieux: Hmmm `¯\_(ツ)_/¯`

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
+CI
```
gdemonet: General notes on the document:

- the title could be more specific, e.g. "Continuous Testing"
- would be great to have an introductory section explaining what the scope of this document is, what are the goals and non-goals

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+    - Build branch ``development/2.(x-1)``
+    - Tests:
+
+      - Single node test
+      - Complex deployment test
+
+  - For last released version of current minor
+
+    e.g.: on ``development/2.x`` when developing "2.x.y" test upgrade from
```
gdemonet: During development, our version is `2.x.y-dev`. Also, what about pre-releases (e.g. `2.x.y-alpha1`)? Should they be treated just the same as a development version, or as a released version?

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,120 @@
...
+
+    - Single node test
+    - Complex deployment test
+
+- Backup, restore tests
+- Solutions tests
+
+.. todo::
+
+   - Define what is "complex deployment"
```
gdemonet: Oh didn't see :) Yes please!

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
+CI
+==
+
+Pre-merge
```
gdemonet, line 19: Let's remind what the dichotomy "pre-merge"/"post-merge" means and why we want to run different tests in each.

Keep in mind that such document can be read by any open-source contributor, as a means for her/him to understand how MetalK8s project is designed. Context and definitions are really useful if not mandatory.

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+---------
+
+What should be tested in pre-merge on every ``user/*``, ``feature/*``,
+``improvement/*``, ``bugfix/*``, ``w/*`` branches. The pre-merge test should
```
gdemonet: Remind what these branch prefixes mean, what they're used for, and why we have them all run in pre-merge (and not somewhere else).

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+not long too much time (less than 40 minutes is great) so we can't
+test everything in pre-merge, we should only test building of the product
+and check that product still usable.
```
gdemonet: Let's try to split goals from solutions in such architecture documents. Our goal here: "have short CI runs (<40mn) before merge". Ideally, explain why we chose this goal (quick feedback loop during development, faster merges means less conflicts... the usual). And then, our solution to this goal: reduce the scope of pre-merge tests, focus on critical aspects (can build, can install from scratch).

gdemonet: Related: what about other targets, e.g. failure rates? And how do we plan on tracking these goals? Or even enforce them?

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+- Installation tests
+
+  - Simple install RHEL
+  - Simple install CentOs + expansion
+
+When merging several pull requests at the same time, given that we are on a
+queue branch (``q/*``), we may want to have or we may require additional tests
+as combinaison of several PRs could have larger impact than all individual PR.
+
+- Simple upgrade/downgrade
```
gdemonet, line 43: At first pass, this looks like a stray bullet point. Maybe add a colon `:` at the end of the previous sentence.

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. _post-merge:
+
+Post-merge
+----------
+
+On each and every ``development/2.*`` branches we want to run complex tests
```
gdemonet: "Complex" doesn't mean much to me. Specifically, we want to cover as much as possible of what isn't handled in pre-merge tests: multiple-nodes scenarii, upgrade/downgrades, disaster recoveries, conformance/performance tests...

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+^^^^^^^
+
+- Upgrade, downgrade tests:
+
+  - For development branch just before
+
+    e.g.: on ``development/2.x`` test upgrade from ``development/2.(x-1)``
+    and downgrade to ``development/2.(x-1)``
+
+    - Build branch ``development/2.(x-1)``
```
gdemonet: Could we rely on existing build artifacts in post-merge? I don't see the benefit of building fresh ones, except if we add some changes to the build itself (e.g. try to update dependencies).

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+    - Tests:
+
+      - Single node test
+      - Complex deployment test
```
gdemonet, line 68: What does this mean? That for each ugrade/downgrade we run them both as single node and "complex" (whatever that means)?

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+On each and every ``development/2.*`` branches we want to run complex tests
+to ensure that the current product continues to work.
+
+Nightly
+^^^^^^^
+
+- Upgrade, downgrade tests:
+
+  - For development branch just before
```
gdemonet:
```suggestion
  - For previous development branch
```

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+  - For last released version of current minor
+
+    e.g.: on ``development/2.x`` when developing "2.x.y" test upgrade from
+    ``metalk8s-2.x.(y-1)`` and downgrade to ``metalk8s-2.x.(y-1)``
+
+    - Single node test
+    - Complex deployment test
+
+  - For last released version of minor just before
```
gdemonet:
```suggestion
  - For last released version of previous minor
```

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+    e.g.: on ``development/2.x`` when developing "2.x.y" test upgrade from
+    ``metalk8s-2.(x-1).z`` and downgrade to ``metalk8s-2.(x-1).z`` where
+    "2.(x-1).z" is the last patch released version for "2.(x-1)" (``z`` may be
+    different from ``y``)
+
+    - Single node test
+    - Complex deployment test
+
+- Backup, restore tests
```
gdemonet: Can we add more details? Should we only cover a Bootstrap node restoration? From a fresh VM or an existing Node? What should we verify after such a restoration?

For backup, we also said we want to check its contents during pre-merge, as it's really inexpensive.

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+- Solutions tests
```
gdemonet, line 95: Same, would need more details around this. Not part of this PR, so adding a `.. todo::` with reference to the related Epic (#1993) would be nice.

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. todo::
+
+   - Define what is "complex deployment"
+
+Weekly
+^^^^^^
+
+More complex tests (TBD)
```
gdemonet: Some suggestions: performance/conformance tests, validation of `contrib` tooling (Heat/Terraform), installation of a "real" Solution like Zenko...

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Adaptive test plan
+------------------
+
+CI pre-merge may be more flexible by including some logic about the content
+of the changeset.
```
gdemonet, line 132: Let's remind what the problem is, what the goals are, and why we think adapting the pre-merge test plan is acceptable (queue builds cover everything, post-merge will always run anyway...), before jumping into the implementation (which may not be valid once we compare it with goals/requirements).

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,122 @@
...
+
+For example:
+
+- For a commit that changes solely/uniquely documentation, we don't need/want
+  to run all the installation test but rather run tests related to
+  documentation.
+- For a commit touching upgrade orchestrate we want to test upgrade directly
+  in pre-merge and not wait :ref:`Post merge<post-merge>` build to get the
+  test result.
```
gdemonet: Those are actual requirements, formulated through examples. Please abstract away:

- "As a developer, I want the ability to reduce the scope of pre-merge tests running for my branch" (and add your docs example)
- "As a developer, I want the ability to augment the scope of pre-merge tests" (and add your upgrade example)
- "As a reviewer, I want the ability to easily verify which tests are scheduled for a given PR"

Also remind that we want this to be an _opt-in_ feature: by default (without specific author action), all tests run. This could prevent untested code to be merged without reviewers noticing.

gdemonet (changes requested):

docs/developer/architecture/ci.rst
TeddyAndrieux: It's should not be part of this document IMO, we should have this in another developper guide about branch naming.

About the "why in pre-merge" I will add it

docs/developer/architecture/ci.rst
TeddyAndrieux: I will add the `-dev`

Do we really plan to release `alpha` version ? Anyway for me alpha and beta are released version

docs/developer/architecture/ci.rst
TeddyAndrieux: > Related: what about other targets, e.g. failure rates? And how do we plan on tracking these goals? Or even enforce them?

?? What do you talk about ?

docs/developer/architecture/ci.rst
TeddyAndrieux: It's not released version so we can't rely on artifacts as we don't know if we already have one or not

docs/developer/architecture/ci.rst
TeddyAndrieux: Yes just need to define complex deployment

docs/developer/architecture/ci.rst
gdemonet: Maybe add a `todo` item to create said document then

docs/developer/architecture/ci.rst
gdemonet: You said one target is `<40mn runs`. What about others? We're trying to write down a set of "requirements" before drafting a "design".

docs/developer/architecture/ci.rst
gdemonet: It's post-merge, so we have the pre-merge artifacts available ;)

docs/developer/architecture/ci.rst
gdemonet: I was wondering if you meant "simple + complex for **each** [up/down]grade scenario". If yes, I think it's overkill. Either just run it in single node or multiple nodes, but I don't see much benefits in running all those cases nightly. Maybe have full coverage on weekly builds.

docs/developer/architecture/ci.rst
TeddyAndrieux: No not necessarily post-merge should run even if nothing is merged on the branch since long time to check that product is not broken due to dependency ... (that's why also we should re-built it) 
Anyway artifacts get removed after ... some time so for previous version we don't know if we will have one available or not

docs/developer/architecture/ci.rst
TeddyAndrieux: Single node upgrade/downgrade is REALLY different from a full cluster with several etcd upgrade/downgrade.

We can easily merge something that will work for multi-node upgrade/downgrade but not for single node (and "vice-versa")

docs/developer/architecture/ci.rst
TeddyAndrieux: > others ?

Goals are now define at the very beginning of the document, here just say that because of `having day-to-day development and PRs merged faster` goal we need to reduce pre-merge

docs/developer/architecture/ci.rst
gdemonet: Mmmh, still not really convinced by the need to rebuild in post-merge. The product dependencies are checked at runtime, when we install. Having a broken build will be visible whenever we run any pre-merge, and if you're really scared that it breaks on some old, inactive branch, you can run a build stage for the current branch in post merge. But doing it twice for each branch (when it's `N` and `N-1`), really not convinced by the benefit (vs. the fact that we'd loose the rest of the tests if this one fails).
W.r.t maintaining build artifacts for old dev branches, I'd argue this is not hard to remedy (define a persistent location for "latest dev/" builds, re-build only when there are no artifacts available, ...).

docs/developer/architecture/ci.rst
gdemonet: Agreed, both cases need to be tested. But should they be tested for each scenario? We're talking two architectures, for six scenarii (3 upgrades, 3 downgrades), for seven development branches. This means we're running 84 install + [up/down]grade every night.

TeddyAndrieux: IMO yes but it's not 84 as you do not test it for dev/2.0 and also for the moment we did not have released version for 2.0 2.1 2.2 2.3 2.5 2.6 but anyway IMO we should test everything nightly, 
For me weekly we should do long test like for example creating a mixed OS cluster and doing several test for the same env like:
- install bootstrap
- Upgrade
- Downgrade
- expansion
- Upgrade
- Crash bootstrap, restore a new one
- expansion
- Upgrade
- Downgrade
So we test "complex stuff" like that backup/restore still working after upgrade/downgrade, expansion still working after upgrade/downgrade and after restore, ...

Anyway if we have too much install to do nightly we can think about skipping commit already tested (mean if a branch did not change since the day before we do not test it)

docs/developer/architecture/ci.rst
TeddyAndrieux: > W.r.t maintaining build artifacts for old dev branches, I'd argue this is not hard to remedy (define a persistent location for "latest dev/" builds, re-build only when there are no artifacts available, ...).

Yes it's doable, but except if we have something directly in `bert-e` to keep the last build of the dev branches, we may need to build

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,155 @@
...
+  - Unit tests
+
+- Installation tests
+
+  - Simple install RHEL
+  - Simple install CentOs + expansion
+
+When merging several pull requests at the same time, given that we are on a
+queue branch (``q/*``), we may want to have or we may require additional tests
+as combinaison of several PRs could have larger impact than all individual PR:
```
Ebaneck: `When merging several pull requests at the same time, given that we are on a
queue branch (``q/*``), we may require additional tests
as a combination of several PRs could have a larger impact than all individual PR:`

docs/developer/architecture/ci.rst
```diff
@@ -0,0 +1,155 @@
...
+------------------
+
+CI pre-merge may be more flexible by including some logic about the content
+of the changeset.
+
+For example:
+
+- For a commit that changes solely/uniquely documentation, we don't need/want
+  to run all the installation test but rather run tests related to
+  documentation.
```
Ebaneck: ` For a commit that changes uniquely documentation, we don't need
  to run the entire installation test suite but rather run tests related to
  documentation.`

Ebaneck (dismissed):

docs/developer/architecture/ci.rst
gdemonet: OK, let's do everything until we find problems.

docs/developer/architecture/ci.rst
gdemonet: Care to put them in the doc or a follow-up ticket, please?

docs/developer/architecture/ci.rst
gdemonet: ping
If you don't want to adress this atm, create a follow-up issue with this comment linked, please

TeddyAndrieux: Just added a sentence about the goal.

But I don't see what you want to say about "why is acceptable ?" your commit change doc you only want to test doc it's just ... obvious IMO I don't see.
I can add a sentence about the fact that we test the product in post-merge  but btw queue build is not acceptable here because they only run for multi PRs merges (or not uptodate ones)

docs/developer/architecture/ci.rst
TeddyAndrieux: Done

docs/developer/architecture/ci.rst
gdemonet: I'm not sure if we'll do pre-releases. At least, providing RCs could make sense.
But agree with you, we can consider them released versions.

docs/developer/architecture/ci.rst
gdemonet: Keeping this conversation open to track when we start writing tickets about this ;)

TeddyAndrieux: We already have this one #2222

docs/developer/architecture/ci.rst
gdemonet: > but btw queue build is not acceptable here because they only run for multi PRs merges (or not uptodate ones)

Yes, but I feel that's a problem. I'd really want anything merged to have been tested with the full pre-merge suite at least once. IMO, adjusting the test plan should be a tool for developers to focus the feedback CI gives them on a given patchset. While the review process is ongoing, focusing tests on what the changes affect allows for quicker iterations. Once approval is given, assuming no flakiness at all in pre-merge tests, adding a full suite in a merge queue shouldn't cause trouble, and safeguard us against merging broken code (of course, that means broken queues, but that's why we put them in place).

gdemonet (dismissed):
LGTM! Let's make sure we add more docs along the way (when we implement post-merge tests, jUnit reports, etc.)

docs/developer/architecture/ci.rst
TeddyAndrieux: That mean lot of change in current eve/bert-e work, because here that only check that a commit get success build so you need to make a distinction between build from pre-merge and build for queue both have a success final status for the same commit but one allow to merge and the other one not ....

gdemonet (approved):

Issue #2261, opened 2020-02-26
Template and store the external IDP configurations and Dex static user store as a service configuration


Ebaneck:
**Component**:

'salt', 'kubernetes', 'charts'

**Why this is needed**:

To guarantee the persistence of external IDP configurations and the Dex static user we provision by default, we need to store these service configurations using a ConfigMap.

**What should be done**:

- For the given ConfigMap to be created, define its schema, its default values
- Deploy the ConfigMap holding the respective Dex service configurations before the consumers are deployed.
- Template and render the service charts(Dex) to read cluster and service configurations from the ConfigMaps directly.

**Implementation proposal** (strongly recommended):

**Test plan**:

Epic: #1996

PR #2274, opened 2020-02-28
chart: template and store Dex backend config in a ConfigMap

Ebaneck:
**Component**:

'salt', 'kubernetes', 'chart'

**Context**: 

See #2261 

**Summary**:

- Have Dex configurations stored and persisted in a ConfigMap.
- For Dex/external IDP config, please reference this doc: https://github.com/dexidp/dex/blob/master/Documentation/connectors/ldap.md#configuration

**Acceptance criteria**: 

- Working cluster

---

Closes: #2261

### DIFF
--- a/charts/render.py
+++ b/charts/render.py
@@ -129,9 +129,23 @@ def keep_doc(doc):
 def replace_magic_strings(rendered_yaml):
     # Handle __var__
     result = re.sub(
-        r'__var__\((?P<varname>[\w\-_]+(?:\.[\w\-_]+)*)\)',
+        r'__var__\((?P<varname>[\w\-_]+(?:\.[\w\-_()]+)*)\)',
         r'{% endraw -%}{{ \g<varname> }}{%- raw %}',
         rendered_yaml,
+    )
+
+    # Handle __var_tojson__
+    result = re.sub(
+        r'__var_tojson__\((?P<varname>[\w\-_]+(?:\.[\w\-_()|]+)*)\)',
+        r'  {% endraw -%}{{ \g<varname> | tojson }}{%- raw %}',
+        result,
+    )
+
+    # Handle __url__
+    result = re.sub(
+        r'__url__\((?P<varname>.*)\)',
+        r'"{% endraw -%}\g<varname>{%- raw %}"',
+        result,
     )
 
     # Handle __image__
--- a/salt/metalk8s/addons/dex/deployed/chart.sls
+++ b/salt/metalk8s/addons/dex/deployed/chart.sls
@@ -1,5 +1,7 @@
 #!jinja | metalk8s_kubernetes
+
 {%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+{%- set dex = salt.metalk8s_service_configuration.get_service_conf('metalk8s-auth', 'metalk8s-dex-config') %}
 
 {% raw %}
 
@@ -18,7 +20,7 @@ metadata:
   namespace: metalk8s-auth
 stringData:
   config.yaml: |-
-    issuer: {% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc{% raw %}
+    issuer: "{% endraw -%}https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc{%- raw %}"
     storage:
       config:
         inCluster: true
@@ -29,6 +31,8 @@ stringData:
       https: 0.0.0.0:5556
       tlsCert: /etc/dex/tls/https/server/tls.crt
       tlsKey: /etc/dex/tls/https/server/tls.key
+    connectors:
+      {% endraw -%}{{ dex.spec.connectors | tojson }}{%- raw %}
     oauth2:
       alwaysShowLoginScreen: true
       responseTypes:
@@ -48,21 +52,16 @@ stringData:
     - id: metalk8s-ui
       name: MetalK8s UI
       redirectURIs:
-      - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/oauth2/callback{%
-        raw %}'
+      - "{% endraw -%}https://{{ grains.metalk8s.control_plane_ip }}:8443/oauth2/callback{%- raw %}"
       secret: ybrMJpVMQxsiZw26MhJzCjA2ut
     - id: grafana-ui
       name: Grafana UI
       redirectURIs:
-      - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%
-        raw %}'
+      - "{% endraw -%}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%- raw %}"
       secret: 4lqK98NcsWG5qBRHJUqYM1
-    enablePasswordDB: true
+    enablePasswordDB: {% endraw -%}{{ dex.spec.localuserstore.enabled }}{%- raw %}
     staticPasswords:
-    - email: admin@metalk8s.invalid
-      hash: $2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W
-      userID: 08a8684b-db88-4b73-90a9-3cd1661f5466
-      username: admin
+      {% endraw -%}{{ dex.spec.localuserstore.userlist | tojson }}{%- raw %}
     expiry:
       idTokens: 24h
       signingKeys: 6h
@@ -147,8 +146,7 @@ metadata:
   name: dex
   namespace: metalk8s-auth
 spec:
-  clusterIP: '{% endraw %}{{ salt.metalk8s_network.get_oidc_service_ip() }}{% raw
-    %}'
+  clusterIP: {% endraw -%}{{ salt.metalk8s_network.get_oidc_service_ip() }}{%- raw %}
   ports:
   - name: https
     port: 32000
@@ -174,7 +172,7 @@ metadata:
   name: dex
   namespace: metalk8s-auth
 spec:
-  replicas: 2
+  replicas: {% endraw -%}{{ dex.spec.deployment.replicas }}{%- raw %}
   selector:
     matchLabels:
       app.kubernetes.io/component: dex
@@ -188,7 +186,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: 278f2b27e9441887e4070365aa3df2ccf668da0025ebf565605555c2b7b16042
+        checksum/config: ad6c30825bdc913fdb7dd6486fd137e443001efd3f183aa7ae13f013bc6f6c38
       labels:
         app.kubernetes.io/component: dex
         app.kubernetes.io/instance: dex
@@ -200,7 +198,7 @@ spec:
         - serve
         - /etc/dex/cfg/config.yaml
         env: []
-        image: '{% endraw %}{{ build_image_name("dex", False) }}{% raw %}:v2.19.0'
+        image: {% endraw -%}{{ build_image_name("dex", False) }}{%- raw %}:v2.19.0
         imagePullPolicy: IfNotPresent
         name: main
         ports:
--- a/salt/metalk8s/addons/dex/deployed/init.sls
+++ b/salt/metalk8s/addons/dex/deployed/init.sls
@@ -13,5 +13,6 @@ include:
 - .namespace
 - .tls-secret
 - .theme-configmap
+- .service-configuration
 - .chart
 - .clusterrolebinding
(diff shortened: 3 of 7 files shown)


salt/metalk8s/addons/dex/deployed/init.sls
```diff
@@ -15,3 +15,4 @@ include:
 - .theme-configmap
 - .chart
 - .clusterrolebinding
+- .service-configuration
```
TeddyAndrieux: This one should be before `chart` I think

salt/metalk8s/addons/dex/deployed/service-configuration.sls
```diff
@@ -0,0 +1,66 @@
...
+                    idAttr: <uidNumber>
+                    emailAttr: <mail>
+                    nameAttr: <name>
+                  groupSearch:
+                    baseDN: <baseDN>
+                    filter: "(objectClass=group)"
+                    userMatchers:
+                      userAttr: <uid>
+                      groupAttr: <member>
+                    nameAttr: <name>
```
TeddyAndrieux: All the field with `<blabla>` are mandatory ? 
If no I think we may want to pre-fill field with default value (maybe `null`) so that a user can easily just change mandatory field and set `enabled` to `True` and ... let's go

salt/metalk8s/addons/dex/deployed/service-configuration.sls
```diff
@@ -0,0 +1,66 @@
...
+                hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+              connectors:
+                enabled: false
+                type: ldap
+                name: LDAP
+                id: ldap
+                config:
+                  host: ldap.metalk8s.invalid:636
+                  startTLS: true
+                  rootCAData: "{{ salt['hashutil.base64_encodefile']('/etc/metalk8s/pki/dex/ca.crt') | replace('\n', '') }}"
```
NicolasT: Why would the Dex CA be the root CA of the LDAP directory server?

salt/metalk8s/addons/dex/deployed/service-configuration.sls
```diff
@@ -0,0 +1,66 @@
...
+            apiVersion: addons.metalk8s.scality.com
+            kind: DexConfig
+            spec:
+              deployment:
+                replicas: 2
+              localuserstore:
+                enabled: true
+                email: "admin@metalk8s.invalid"
+                hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+              connectors:
```
NicolasT: This field is called connector**s** but it's not some kind of list?

salt/metalk8s/addons/dex/deployed/service-configuration.sls
```diff
@@ -0,0 +1,66 @@
...
+                enabled: false
+                type: ldap
```
NicolasT: I don't like having this in the 'default' config. The 'default' config should represent the config that's running and not include 'stuff' with placeholders that's not used at all. That ought to be found in documentation.

salt/metalk8s/addons/dex/deployed/service-configuration.sls
NicolasT: IMHO (see above) those fields shouldn't be here at all. It's super confusing.

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -29,6 +31,31 @@ stringData:
...
+          userMatchers:
+            groupAttr: {% endraw -%}{{ dex.spec.connectors.config.groupSearch.userMatchers.groupAttr }}{%- raw %}
+            userAttr: {% endraw -%}{{ dex.spec.connectors.config.groupSearch.userMatchers.userAttr }}{%- raw %}
+        host: {% endraw -%}{{ dex.spec.connectors.config.host }}{%- raw %}
+        rootCAData: {% endraw -%}{{ dex.spec.connectors.config.rootCAData }}{%- raw %}
+        startTLS: true
+        userSearch:
+          baseDN: {% endraw -%}{{ dex.spec.connectors.config.userSearch.baseDN }}{%- raw %}
+          emailAttr: {% endraw -%}{{ dex.spec.connectors.config.userSearch.emailAttr }}{%- raw %}
+          filter: (&(objectclass=user))
```
NicolasT: `objectclass` or `objectClass`?

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -29,6 +31,31 @@ stringData:
...
+          idAttr: {% endraw -%}{{ dex.spec.connectors.config.userSearch.idAttr }}{%- raw %}
+          nameAttr: {% endraw -%}{{ dex.spec.connectors.config.userSearch.nameAttr }}{%- raw %}
+          username: {% endraw -%}{{ dex.spec.connectors.config.userSearch.username }}{%- raw %}
+        usernamePrompt: SSO Username
```
NicolasT: Wouldn't it make sense for this to be configurable? What if the login system is not known as `SSO` within the organization?

salt/metalk8s/addons/dex/deployed/service-configuration.sls
Ebaneck: This should not be the case, i only found out that without a value in this field, the Dex containers crashed.

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -48,19 +75,17 @@ stringData:
...
-      - '{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%
-        raw %}'
+      - "{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{% raw %}"
       secret: 4lqK98NcsWG5qBRHJUqYM1
-    enablePasswordDB: true
+    enablePasswordDB: {% endraw -%}{{ dex.spec.localuserstore.enabled }}{%- raw %}
     staticPasswords:
-    - email: admin@metalk8s.invalid
-      hash: $2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W
+    - email: {% endraw -%}{{ dex.spec.localuserstore.email }}{%- raw %}
```
NicolasT: Why would we allow changing the 'email' when we don't allow to change the 'username'?

Also, shouldn't this be a *list* of accounts in the config...

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -188,7 +212,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: 278f2b27e9441887e4070365aa3df2ccf668da0025ebf565605555c2b7b16042
+        checksum/config: 82272216f0373dd6510af9cea925a8642900f1fe7400d2cbcf4ef54e59ac7d9c
```
NicolasT: We're changing this when needed, right?

NicolasT (dismissed):

salt/metalk8s/addons/dex/deployed/service-configuration.sls
NicolasT: Then let's use... 'some' other value, not an actually used cert.

salt/metalk8s/addons/dex/deployed/service-configuration.sls
TeddyAndrieux: Yes agree we discussed this directly with @Ebaneck also look like the `enabled` key do not work so basically with default bootstrap deployment you will have an "ldap" button in UI that basically do not work.
Just asked, for first iteration, to just copy `connectors` contains in the dex configuration and not checking field one by one + adding some documentation for ldap

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: Yes, going to make this configurable

salt/metalk8s/addons/dex/deployed/service-configuration.sls
TeddyAndrieux: IMO just not prefill `connectors` and just have an empty list by default with documentation explaning how to fill it

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: True, it should be a list of accounts!

I think i left out `username` because the APIserver flag we set was `--oidc-username-claim=email`

Now that I remember that the UI might want to display the username, i will make this field configurable

salt/metalk8s/addons/dex/deployed/chart.sls
Ebaneck: I thought this checksum hashes are generated each time we render the charts? Don't we need to use the new values generated?
Also, when are we required to change it ?

salt/metalk8s/addons/dex/deployed/service-configuration.sls
Ebaneck: Agreed!

salt/metalk8s/addons/dex/deployed/secret.sls
```diff
@@ -0,0 +1,110 @@
...
+                      filter: {{ connector['config']['groupSearch']['filter'] }}
+                      userMatchers:
+                        userAttr: {{ connector['config']['groupSearch']['userMatchers']['userAttr'] }}
+                        groupAttr: {{ connector['config']['groupSearch']['userMatchers']['groupAttr'] }}
+                      nameAttr: {{ connector['config']['groupSearch']['nameAttr'] }}
+
+              {%- endif %}
+
+              {%- endfor %}
+            {%- endif %}
```
TeddyAndrieux: I don't get why you do this for me you just have to directly use the connectors content

```suggestion
            connectors: {{ dex.spec.connectors | tojson }}
```
And then you don't need to extract secret from the chart

Ebaneck: True!

charts/render.py
```diff
@@ -129,11 +129,25 @@ def keep_doc(doc):
...
+    result = re.sub(
+        r'__var_tojson__\((?P<varname>[\w\-_]+(?:\.[\w\-_()|]+)*)\)',
+        r'  {% endraw -%}{{ \g<varname> | tojson }}{%- raw %}',
+        result,
+    )
+
+    # Handle __url__
+    result = re.sub(
+        r'__url__\((?P<varname>.*)\)',
+        r'"{% endraw %}\g<varname>{% raw %}"',
```
TeddyAndrieux: To be consistent 
```suggestion
        r'"{% endraw -%}\g<varname>{%- raw %}"',
```

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -29,6 +31,8 @@ stringData:
       https: 0.0.0.0:5556
       tlsCert: /etc/dex/tls/https/server/tls.crt
       tlsKey: /etc/dex/tls/https/server/tls.key
+    connectors:
+      {% endraw -%}{{ dex.spec.connectors | tojson }}{%- raw %}
```
TeddyAndrieux, line 35: This works ?
For me you need to do 
```suggestion
    connectors: {% endraw -%}{{ dex.spec.connectors | tojson }}{%- raw %}
```
Anyway even if it's work I think it's clearer to do like I suggest

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -48,21 +52,16 @@ stringData:
...
+      - "{% endraw %}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{% raw %}"
       secret: 4lqK98NcsWG5qBRHJUqYM1
-    enablePasswordDB: true
+    enablePasswordDB: {% endraw -%}{{ dex.spec.localuserstore.enabled }}{%- raw %}
     staticPasswords:
-    - email: admin@metalk8s.invalid
-      hash: $2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W
-      userID: 08a8684b-db88-4b73-90a9-3cd1661f5466
-      username: admin
+      {% endraw -%}{{ dex.spec.localuserstore.userlist | tojson }}{%- raw %}
```
TeddyAndrieux, line 64: ditto

salt/metalk8s/addons/dex/deployed/chart.sls
TeddyAndrieux: Ok it's because of default manifest from DEX lets keep it

salt/metalk8s/addons/dex/deployed/chart.sls
TeddyAndrieux: ditto x2

TeddyAndrieux (approved):
LGTM

Issue #2286, opened 2020-03-04
Ubuntu package download is broken (again)


alexandre-allard:
**Component**: build

**What happened**:
We're no longer able to download Ubuntu packages
See here for the exact error: https://eve.devsca.com/github/scality/metalk8s/#/builders/11/builds/12858/steps/8/logs/stdio

**What was expected**: build go well :) 

**Steps to reproduce**: `./doit.sh packaging`

**Resolution proposal** (optional):

alexandre-allard:
Looks like the problem comes from downloading the `salt-minion` package.

PR #2287, opened 2020-03-04
build: escape "+" char in package URL

alexandre-allard:
**Component**: build

**Context**: 

**Summary**:

**Acceptance criteria**:
We can download Ubuntu packages and the build is green.
Basically, `./doit.sh packaging` must work

---

Closes: #2286

### DIFF
--- a/packages/debian/download_packages.py
+++ b/packages/debian/download_packages.py
@@ -78,7 +78,9 @@ def fetch_binary(
     filename = pathlib.Path(package.filename).name
     destfile = destdir/filename
     print('Downloading package {}'.format(filename))
-    with urllib.request.urlopen(package.uri) as response:
+    pkg_uri = urllib.parse.urlparse(package.uri)
+    pkg_uri = pkg_uri._replace(path=urllib.parse.quote(pkg_uri.path))
+    with urllib.request.urlopen(urllib.parse.urlunparse(pkg_uri)) as response:
         destfile.write_bytes(response.read())
     # Check package size.
     if destfile.stat().st_size != package.size:


packages/debian/download_packages.py
```diff
@@ -78,7 +78,9 @@ def fetch_binary(
     filename = pathlib.Path(package.filename).name
     destfile = destdir/filename
     print('Downloading package {}'.format(filename))
-    with urllib.request.urlopen(package.uri) as response:
+    pkg_uri = urllib.parse.urlparse(package.uri)
+    pkg_uri = pkg_uri._replace(path=urllib.parse.quote(pkg_uri.path))
+    with urllib.request.urlopen(urllib.parse.urlunparse(pkg_uri)) as response:
```
slaperche-scality, line 83: Why not simply:

```python
uri = urllib.parse.quote(package.uri)
with urllib.request.urlopen(uri) as response:
```
?

Also, if the problem is a bad char in the URL, why does it breaks just now?

URL changed from upstream recently? (if so, could be mentioned in the commit message, even giving an example of problematic URI would be cool)

alexandre-allard: The current URL is:
```
http://repo.saltstack.com/apt/ubuntu/18.04/amd64/archive/2018.3.4/pool/main/s/salt/salt-minion_2018.3.4+ds-1_all.deb
```
Problem is the `+` character is interpreted as a space, so we need to escape it, otherwise you end up trying to fetch:
```
http://repo.saltstack.com/apt/ubuntu/18.04/amd64/archive/2018.3.4/pool/main/s/salt/salt-minion_2018.3.4
```

Can't quote the whole URI, otherwise it also escapes `:` after the scheme:
```
http%3A//repo.saltstack.com/apt/ubuntu/18.04/amd64/archive/2018.3.4/pool/main/s/salt/salt-minion_2018.3.4%2Bds-1_all.deb
```

I don't know if the URI changed recently, it seems not:
```
2019-10-03T22:34:59.000Z        26.9 kB        salt-minion_2018.3.4+ds-1_all.deb
```
Maybe the change comes from elsewhere, urllib, the remote server...  I'll investigate, I just wanted to fix it quickly first, as it blocks the builds (and it was midnight).

slaperche-scality: Ok, works for me.
But would be good to find where the change came from, to be sure this fix is the right one.

slaperche-scality (dismissed):

packages/debian/download_packages.py
alexandre-allard: It seems that's not related to `urllib`, as it succeed to fetch other packages with `+` sign in the URL.
I'm 99.9% sure there is a change on repo.saltstack.com on how it handles URL.

alexandre-allard: Ok, easy to reproduce with a browser go to:
```
http://archive.ubuntu.com/ubuntu/pool/main/libz/libzstd/libzstd1_1.3.3+dfsg-2ubuntu1.1_amd64.deb
```
The package is downloaded.
Then, go to:
```
http://repo.saltstack.com/apt/ubuntu/18.04/amd64/archive/2018.3.4/pool/main/s/salt/salt-minion_2018.3.4+ds-1_all.deb
```
It does not work.

alexandre-allard: I'll change the commit message for something more accurate.

slaperche-scality (approved):
Good job for the commit message, at least now we have a clearer idea of the issue :+1:

Issue #2262, opened 2020-02-26
Store the Alertmanager notification configurations as a service configuration


Ebaneck:
**Component**:

'salt', 'kubernetes', 'alertmanager'

**Why this is needed**:

For persistence of Alertmanager notification configurations, we need to store these service configurations using a ConfigMap.

Here, we are talking about integrations to Slack, email(smtp), victorops etc

**What should be done**:

- Define the schema and default values for this ConfigMap
- Deploy the ConfigMap holding these service values
- Template and render the prometheus-operator charts to read cluster and service configurations from the ConfigMaps directly.

**Implementation proposal** (strongly recommended):

**Test plan**:

Epic #1996

PR #2289, opened 2020-03-05
Add Alertmanager conf as a customizable service configuration

Ebaneck:
**Component**:

'salt', 'charts', 'prometheus-operator'

**Context**: 

See #2262 

**Summary**:

Template and store `alertmanager.yaml` as a service configuration in a MetalK8s cluster making it possible that users can customize and add new alert channels.

**Acceptance criteria**: 

---

Closes: #2262

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -269,6 +269,8 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/prometheus-adapter/deployed/chart.sls'),
     Path('salt/metalk8s/addons/prometheus-adapter/deployed/init.sls'),
 
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/'
+         'alertmanager-configuration-secret.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/chart.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/cleanup.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/dashboards.sls'),
--- a/charts/prometheus-operator.yaml
+++ b/charts/prometheus-operator.yaml
@@ -38,6 +38,18 @@ alertmanager:
             matchLabels:
               app.kubernetes.io/name: prometheus-operator-alertmanager
 
+  ## Passing the below configuration directives through the helm template
+  ## engine leads to no substitution since the variables passed will be base64
+  ## encoded directly hence we end up with `alertmanager.yaml` being invalid
+
+  ## So let us disable the creation of this secret from the default charts
+  ## and then create a new secret that replaces the
+  ## alertmanager-prometheus-operator-alertmanager secret with new config
+  ## read from a ConfigMap (metalk8s-alertmanager-config)
+
+  # config: '__var_tojson__(alertmanager.spec.notification.config)'
+
+    useExistingSecret: true
 
 prometheusOperator:
   tlsProxy:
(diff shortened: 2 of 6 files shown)


salt/metalk8s/addons/prometheus-operator/deployed/alertmanager-configuration-secret.sls
```diff
@@ -0,0 +1,20 @@
...
+    - kind: Secret
+    - namespace: metalk8s-monitoring
+    - patch:
+        stringData:
+          alertmanager.yaml: |-
+            global: {{ alertmanager.spec.notification.config.global | tojson }}
+            templates: {{ alertmanager.spec.notification.config.templates | tojson }}
+            route: {{ alertmanager.spec.notification.config.route | tojson }}
+            receivers: {{ alertmanager.spec.notification.config.receivers | tojson }}
+            inhibit_rules: {{ alertmanager.spec.notification.config.inhibit_rules | tojson }}
```
TeddyAndrieux: Why not put directly the config dict ?

```suggestion
          alertmanager.yaml: {{ alertmanager.spec.notification.config | tojson }}
```

charts/prometheus-operator.yaml
```diff
@@ -38,6 +38,19 @@ alertmanager:
             matchLabels:
               app.kubernetes.io/name: prometheus-operator-alertmanager
 
+  ## Passing the below configuration directives through the helm template
+  ## engine leads to no changes.
+  ## So let us create a new secret that overrides/patches
+  ## alertmanager-prometheus-operator-alertmanager secret with new config
+  ## read from a ConfigMap (metalk8s-alertmanager-config)
+  ## This implies we don't have to worry about patching the local charts
```
TeddyAndrieux: Do we need to patch lot of stuff in the chart ? Because if for what ever reason the secret chart change in another version we will not get the changes

Ebaneck: I think the main issue here was that each variable passed was directly base64 encoded. This is the current behaviour of the chart.

If we rationally go by the way we write, substitute and dump the jinja variables, this meant we will have something like `global: __var_tojson__(alertmanager.spec.notification.config.global)` inside `alertmanager.yaml` once we base64 decode.

buildchain/buildchain/salt_tree.py
```diff
@@ -268,6 +268,8 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/addons/prometheus-adapter/deployed/chart.sls'),
     Path('salt/metalk8s/addons/prometheus-adapter/deployed/init.sls'),
 
+    Path('salt/metalk8s/addons/prometheus-operator/deployed/'
+            'alertmanager-configuration-secret.sls'),
```
slaperche-scality: weird indentation

Ebaneck: Ok! will fix

charts/prometheus-operator.yaml
TeddyAndrieux: Ok make sense, let keep your approach then

salt/metalk8s/addons/prometheus-operator/deployed/alertmanager-configuration-secret.sls
```diff
@@ -0,0 +1,16 @@
...
+Patch Alertmanager Configuration Secret:
+  metalk8s_kubernetes.object_updated:
+    - name: alertmanager-prometheus-operator-alertmanager
+    - apiVersion: v1
+    - kind: Secret
+    - namespace: metalk8s-monitoring
+    - patch:
+        stringData:
+          alertmanager.yaml: |-
+            {{ alertmanager.spec.notification.config | tojson }}
```
TeddyAndrieux, line 23: Does it work ??? because the `|-` will get the bellow block as a string where we expect a dict
For me you should do 
```suggestion
          alertmanager.yaml: {{ alertmanager.spec.notification.config | tojson }}
```
As here we control it, it's not from the template

TeddyAndrieux: Ok no i'm wrong I was thinking it's a subkey of the yaml alertmanager but no the spec contain the full yaml so yes you right sorry for the noise

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):
LGTM

Issue #2292, opened 2020-03-06
Add tests to ensure that update operations on service configs are propagated to the underlying service


Ebaneck:
**Component**:

'tests', 'kubernetes'

**Why this is needed**:

We need to be able to assert that any configuration changes on a Cluster and Service configuration is picked-up by the underline service

**What should be done**:

- Edit a ConfigMap containing service configurations e.g metalk8s-dex-config
- Bump the replica count to say (3) replicas
- Run the salt state to apply the changes to the deployment and maybe restart the deployment
- Check and verify that 3 replicas of the Dex pod exist within your cluster

**Implementation proposal** (strongly recommended):

**Test plan**:

gdemonet:
> Run the salt state to apply the changes to the deployment and maybe restart the deployment

We should not need to restart anything, that's the responsibility of Salt.

> Check and verify that 3 replicas of the Dex pod exist within your cluster

No, don't look at the Pod, look at the actual object that will be updated: the Deployment `dex` in `metalk8s-auth` namespace.

PR #2300, opened 2020-03-11
Add tests to check service config propagation

Ebaneck:
**Component**:

'tests', 'kubernetes', 'salt'

**Context**: 

See #2292 

**Summary**:

- Add tests to ensure that Service Configurations are propagated to the underlying services.

**Acceptance criteria**: 

---

Closes: #2292

### DIFF
--- a/tests/post/features/service_configuration.feature
+++ b/tests/post/features/service_configuration.feature
@@ -1,0 +1,10 @@
+@post @local @ci @csc
+Feature: Cluster and Services Configurations
+    Scenario: Propagation of Service Configurations to underlying Services
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        And we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth'
+        And we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth' with 'spec.deployment.replicas' equal to '2'
+        When we update 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth' 'spec.deployment.replicas' to '3'
+        And we apply the 'metalk8s.addons.dex.deployed' salt state
+        Then we have '3' at 'status.available_replicas' for 'dex' Deployment in namespace 'metalk8s-auth'
--- a/tests/post/steps/test_service_configuration.py
+++ b/tests/post/steps/test_service_configuration.py
@@ -1,0 +1,217 @@
+from ast import literal_eval
+import yaml
+
+import pytest
+from pytest_bdd import scenario, given, then, when, parsers
+
+import kubernetes.client
+from kubernetes.client import AppsV1Api
+from kubernetes.client.rest import ApiException
+
+from tests import utils
+
+
+# Fixtures {{{
+
+
+@pytest.fixture
+def k8s_appsv1_client(k8s_apiclient):
+    return AppsV1Api(api_client=k8s_apiclient)
+
+# }}}
+
+
+# Scenarios {{{
+
+
+@scenario('../features/service_configuration.feature',
+          'Propagation of Service Configurations to underlying Services')
+def test_service_config_propagation(host):
+    pass
+
+
+# }}}
+
+
+# Given {{{
+
+
+@given(parsers.parse(
+    "we have a '{name}' CSC in namespace '{namespace}' with "
+    "'{path}' equal to '{value}'"))
+def check_csc_configuration(
+    k8s_client,
+    name,
+    namespace,
+    path,
+    value
+):
+    csc_response = read_namespace_config_map(k8s_client, name, namespace)
+
+    assert csc_response, (
+        "No ConfigMap with name {} in namespace {} found".format(
+            name, namespace
+        )
+    )
+
+    csc_obj = csc_load(csc_response, name, namespace)
+    response_value = utils.get_dict_element(csc_obj, path)
+
+    assert literal_eval(value) == response_value, (
+        "Expected value {} for key {} in ConfigMap {} found in namespace {}, "
+        "got {}".format(value, path, name, namespace, response_value)
+    )
+
+
+# }}}
+
+
+# When {{{
+
+
+@when(parsers.parse(
+    "we update '{name}' CSC in namespace '{namespace}' "
+    "'{path}' to '{value}'"))
+def update_service_configuration(
+    k8s_client,
+    name,
+    namespace,
+    path,
+    value
+):
+
+    full_csc = read_namespace_config_map(k8s_client, name, namespace)
+    csc_obj = utils.set_dict_element(
+        csc_load(full_csc, name, namespace), path, literal_eval(value)
+    )
+    patch = {
+        'data': {
+            'config.yaml': yaml.safe_dump(
+                csc_obj, default_flow_style=False
+            )
+        }
+    }
+    response = patch_namespace_config_map(
+        k8s_client, name, namespace, patch
+    )
+
+    assert response
+
+    patched_csc = read_namespace_config_map(
+        k8s_client, name, namespace
+    )
+    patched_value = utils.get_dict_element(
+        csc_load(patched_csc, name, namespace), path
+    )
+
+    assert literal_eval(value) == patched_value, (
+        "Expected value {} for key {} in ConfigMap {} found in namespace {}, "
+        "got {}".format(value, path, name, namespace, patched_value)
+    )
+
+
+@when(parsers.parse("we apply the '{state}' salt state"))
+def apply_service_config(host, version, request, k8s_client, state):
+    ssh_config = request.config.getoption('--ssh-config')
+
+    cmd = [
+        'salt-run', 'state.sls', '{}'.format(state),
+        'saltenv=metalk8s-{}'.format(version)
+    ]
+    utils.run_salt_command(host, cmd, ssh_config)
+
+
+# }}}
+
+
+# Then {{{
+
+
+@then(parsers.parse(
+    "we have '{value}' at '{path}' for '{service}' Deployment in namespace "
+    "'{namespace}'"))
+def get_deployments(
+    host,
+    version,
+    k8s_appsv1_client,
+    k8s_client,
+    request,
+    value,
+    path,
+    service,
+    namespace
+
+):
+    def _wait_for_deployment():
+        try:
+            response = k8s_appsv1_client.read_namespaced_deployment(
+                name=service, namespace=namespace
+            ).to_dict()
+        except Exception as exc:
+            pytest.fail(
+                "Unable to read {} Deployment with error: {!s}".format(
+                    service, exc
+                )
+            )
+        assert literal_eval(value) == utils.get_dict_element(response, path), (
+            "Expected value {} for service {}, got {}".format(
+                value, service, utils.get_dict_element(response, path)
+            )
+        )
+    utils.retry(
+        _wait_for_deployment,
+        times=10,
+        wait=5,
+        name="wait for available Deployments"
+    )
+
+
+# }}}
+
+
+# Helpers {{{
+
+
+def read_namespace_config_map(k8s_client, name, namespace):
+    try:
+        response = k8s_client.read_namespaced_config_map(
+            name, namespace
+        )
+    except Exception as exc:
+        pytest.fail(
+            "Unable to read {} ConfigMap  in namespace {} with error: {!s}"
+            .format(name, namespace, exc)
+        )
+    return response
+
+
+def patch_namespace_config_map(k8s_client, name, namespace, patch):
+    try:
+        response = k8s_client.patch_namespaced_config_map(
+            name, namespace, patch
+        )
+    except Exception as exc:
+        pytest.fail(
+            "Unable to patch ConfigMap {} in namespace {} with error {!s}"
+            .format(name, namespace, exc)
+        )
+    return response
+
+
+def csc_load(full_csc, name, namespace):
+    try:
+        csc_obj = yaml.safe_load(full_csc.data['config.yaml'])
+    except yaml.YAMLError as exc:
+        raise Exception(
+            'Invalid YAML format in ConfigMap {} found in namespace {}: {!s}'
+            .format(name, namespace, exc)
+        )
+    except Exception as exc:
+        raise Exception(
+            "Failed loading `config.yaml` from ConfigMap {} in namespace {}: "
+            "{!s}".format(name, namespace, exc)
+        )
+    return csc_obj
+
+
+# }}}
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,6 +1,8 @@
+import functools
 import ipaddress
 import logging
 import re
+import operator
 import testinfra
 import time
 from typing import Optional, Dict
@@ -160,3 +162,21 @@ def run_salt_command(host, command, ssh_config):
             output.stdout,
             output.stderr
         )
+
+
+def get_dict_element(data, path, delimiter='.'):
+    """
+    Traverse a dict using a 'delimiter' on a target string.
+    getitem(a, b) returns the value of a at index b
+    """
+    return functools.reduce(operator.getitem, path.split(delimiter), data)
+
+
+def set_dict_element(data, path, value, delimiter='.'):
+    """
+    Traverse a nested dict using a delimiter on a target string
+    replaces the value of a key within a dictionary and returns the new dict
+    """
+    path, _, key = path.rpartition(delimiter)
+    (get_dict_element(data, path) if path else data)[key] = value
+    return data
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -109,3 +109,54 @@ def requests_retry_session(
     session.mount("http://", adapter)
     session.mount("https://", adapter)
     return session
+
+
+def kubectl_exec(
+    host,
+    command,
+    pod,
+    kubeconfig='/etc/kubernetes/admin.conf',
+    **kwargs
+):
+    """Grab the return code from a `kubectl exec`"""
+    kube_args = ['--kubeconfig', kubeconfig]
+
+    if kwargs.get('container'):
+        kube_args.extend(['-c', kwargs.get('container')])
+    if kwargs.get('namespace'):
+        kube_args.extend(['-n', kwargs.get('namespace')])
+
+    kubectl_cmd_tplt = 'kubectl exec {} {} -- {}'
+
+    with host.sudo():
+        output = host.run(
+            kubectl_cmd_tplt.format(
+                pod,
+                ' '.join(kube_args),
+                ' '.join(command)
+            )
+        )
+        return output
+
+
+def run_salt_command(host, command, ssh_config):
+    """Run a command inside the salt-master container."""
+
+    pod = 'salt-master-{}'.format(
+        get_node_name('bootstrap', ssh_config)
+    )
+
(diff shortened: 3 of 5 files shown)


tests/post/features/service_configuration.feature
```diff
@@ -0,0 +1,11 @@
+@post @local @ci @persistence
+Feature: Cluster and Services Configurations
+    Scenario: Propagation of Service Configurations to underlying Services
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        And we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth' on node 'bootstrap'
```
TeddyAndrieux: Do not specify the `on node 'bootstrap'` please
```suggestion
        And we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth'
```

tests/post/features/service_configuration.feature
```diff
@@ -0,0 +1,11 @@
...
+        And we have a 'metalk8s-dex-config' ConfigMap with replicas count '2'
```
TeddyAndrieux: I think you should use something more generic than that, also here you match directly in the config and `ConfigMap` is not really a keyword here. 

IMO we should have something more generic that you can use for different test something like that
```
we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth' with '.spec.deployment.replicas' equal to '2'
```

So a fixture look like this 
```
we have a '<name>' CSC in namespace '<namespace>' with '<path>' equal to '<value>'
```

tests/post/features/service_configuration.feature
```diff
@@ -0,0 +1,11 @@
...
+        When I update the replicas count to '3'
+        And I apply a salt-state to propagate the changes to Dex Service
```
TeddyAndrieux: Those two should be more generic also

```
I update '<name>' CSC in namespace '<namespace>' '<path>' to '<value>'
```

```
I apply the '<salt states>' salt state
```

tests/post/features/service_configuration.feature
```diff
@@ -0,0 +1,11 @@
...
+        Then we have '3' available Deployments for the Dex app
```
TeddyAndrieux:
```
we have 3 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth'
```

TeddyAndrieux (changes requested):
First quick review

tests/post/features/service_configuration.feature
TeddyAndrieux: Oh and btw to be consistent lets use `we` instead of `I`

tests/post/features/service_configuration.feature
Ebaneck: Please see : https://github.com/scality/metalk8s/issues/2292#issuecomment-596614157

TeddyAndrieux: Ok as you want but still need something more generic IMO like

```
we have '3' at '.spec.status.available_replicas' for 'dex' Deployment in namespace 'metalk8s-auth'
```

tests/post/features/service_configuration.feature
```diff
@@ -0,0 +1,10 @@
+@post @local @ci @persistence
+Feature: Cluster and Services Configurations
+    Scenario: Propagation of Service Configurations to underlying Services
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        And we have 2 running pod labeled 'app.kubernetes.io/name=dex' in namespace 'metalk8s-auth'
+        And we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth' with '['spec']['deployment']['replicas']' equal to '2'
+        When we update 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth' '['spec']['deployment']['replicas']' to '3'
```
TeddyAndrieux: I really prefer you use jsonpath syntax here (pretty easy in python you just have to split on `.` and get)

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+    path,
+    value
+):
+    # This patch here overrides the entire config.yaml
+    try:
+        response = k8s_client.patch_namespaced_config_map(
+            name, namespace, yaml.safe_load(BODY.format(
+                replicas_count=value)
+            )
+        )
```
TeddyAndrieux: No please do not override the full config,
Retrieve it load the yaml and edit the field you want and then push it back

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+# }}}
+
+
+# Given {{{
+
+
+@given(parsers.parse(
+    "we have a '{name}' CSC in namespace '{namespace}' with "
+    "'{path}' equal to '{value}'"))
+def check_configmap_configuration(
```
TeddyAndrieux: Minor: 
```suggestion
def check_csc_configuration(
```

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+                name, exc
+            )
+        )
+
+    response_value = \
+        eval("response_yaml"+"{}".format(path))
+
+    assert int(value) == response_value, (
+        "Expected value {} for key {}, got {}".format(
+            value, path, response_value
```
TeddyAndrieux: Please add the name of the config map and the namespace in the message

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+        response_yaml = yaml.safe_load(response.data['config.yaml'])
+    except yaml.YAMLError as exc:
+        raise Exception(
+            'Invalid YAML format in ConfigMap {}: {!s}'.format(name, exc)
+        )
+    except Exception as exc:
+        raise Exception(
+            'Failed loading `config.yaml` from ConfigMap {}: {!s}'.format(
+                name, exc
+            )
```
TeddyAndrieux: Add the namespace also please

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+    response = read_namespace_config_map(k8s_client, name, namespace)
+
+    assert response, ("No ConfigMap with name {} found".format(name))
+
+    try:
+        response_yaml = yaml.safe_load(response.data['config.yaml'])
+    except yaml.YAMLError as exc:
+        raise Exception(
+            'Invalid YAML format in ConfigMap {}: {!s}'.format(name, exc)
+        )
```
TeddyAndrieux: + namespace :smiley:

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+        k8s_client, name, namespace
+    )
+
+    try:
+        config_yaml = yaml.safe_load(
+            read_patch_configmap.data['config.yaml']
+        )
+    except yaml.YAMLError as exc:
+        raise Exception(
+            'Invalid YAML format in ConfigMap {}: {!s}'.format(name, exc)
```
TeddyAndrieux: ditto

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+        )
+    except Exception as exc:
+        raise Exception(
+            'Failed loading `config.yaml` from ConfigMap {}: {!s}'.format(
```
TeddyAndrieux: dittoto

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,261 @@
...
+    ssh_config = request.config.getoption('--ssh-config')
+
+    cmd = [
+        'salt-run', 'state.sls', '{}'.format(state),
+        'saltenv=metalk8s-{}'.format(version)
+    ]
+    output = utils.run_salt_command(host, cmd, ssh_config)
+
+    assert output.exit_status == 0, \
+        'Failed applying salt-state: \nout: {}\nerr:'.format(
```
TeddyAndrieux: Add the name of the salt state you try to apply please

tests/post/features/service_configuration.feature
gdemonet: You also have those `salt.utils` methods: [`salt.utils.data.traverse_dict`](https://github.com/saltstack/salt/blob/2018.3/salt/utils/data.py#L533) and [`salt.utils.data.traverse_dict_and_list`](https://github.com/saltstack/salt/blob/2018.3/salt/utils/data.py#L551)

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+    except Exception as exc:
+        raise Exception(
+            "Failed loading `config.yaml` from ConfigMap {} in namespace {}: "
+            "{!s}".format(name, namespace, exc)
+        )
+    return csc_obj
+
+
+def traversed_dict(data, key):
+
```
TeddyAndrieux: Remove this empty line please

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+    '''
+    Traverse a dict using a 'delimiter' on a  target string.
```
TeddyAndrieux: Minor: SInce you talk about delimiter either you should add a delimiter arg to this function (that default to `.` or talk about dot here (personnally prefer to have delimiter as argument :) )

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+    csc_obj['spec']['deployment']['replicas'] = 2
+    patch = \
+        {'data': {'config.yaml': yaml.safe_dump(
+            csc_obj, default_flow_style=False)
+        }}
+    response = patch_namespace_config_map(
+        k8s_client, name, namespace, patch
+    )
+
+    return response
```
TeddyAndrieux: Hmm why ? Do we really care about it ? (Also here `metalk8s-dex-config` is hardcoded not really like it)

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+):
+
+    full_csc = read_namespace_config_map(k8s_client, name, namespace)
+    csc_obj = yaml_load(full_csc, name, namespace)
+
+    csc_obj['spec']['deployment']['replicas'] = int(value)
+    patch = \
+        {'data': {'config.yaml': yaml.safe_dump(
+            csc_obj, default_flow_style=False)
+        }}
```
TeddyAndrieux: Minor: Personnaly don't really like this indentation
```suggestion
    patch = {
        'data': {
            'config.yaml': yaml.safe_dump(csc_obj, default_flow_style=False)
        }
    }
```

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+                name=service, namespace=namespace
+            )
+        except Exception as exc:
+            pytest.fail(
+                "Unable to read {} Deployment with error: {!s}".format(
+                    service, exc
+                )
+            )
+
+        assert int(value) == eval("response"+"{}".format(path)), (
```
TeddyAndrieux: ??? I think you forgot to use `traversed_dict` here.

And global python comment: better to avoid using `eval` as much as possible

slaperche-scality: Yeah, `eval` is a no-go.

tests/utils.py
```diff
@@ -104,3 +104,49 @@ def requests_retry_session(
...
+
+    output = kubectl_exec(
+        host,
+        command,
+        pod,
+        container='salt-master',
+        namespace='kube-system'
+    )
+
+    return output
```
slaperche-scality: In `install/steps/test_expansion.py` we have an helper like that:

```python
def run_salt_command(host, command, ssh_config):
    """Run a command inside the salt-master container."""

    pod = 'salt-master-{}'.format(
        utils.get_node_name('bootstrap', ssh_config)
    )

    output = kubectl_exec(
        host,
        command,
        pod,
        container='salt-master',
        namespace='kube-system'
    )

    assert output.exit_status == 0, \
        'deploy failed with: \nout: {}\nerr:'.format(
            output.stdout,
            output.stderr
        )
```

You should either reuse it/move it here (with a bit of rework) OR replace it by yours.
But we shouldn't keep both.

tests/utils.py
```diff
@@ -104,3 +104,49 @@ def requests_retry_session(
...
+
+    with host.sudo():
+        output = host.run(
+            kubectl_cmd_tplt.format(
+                pod,
+                ' '.join(kube_args),
+                ' '.join(command)
+            )
+        )
+        return output
```
slaperche-scality, line 141: In `install/steps/test_expansion.py` we already have an helper like that:

```python
def kubectl_exec(
    host,
    command,
    pod,
    kubeconfig='/etc/kubernetes/admin.conf',
    **kwargs
):
    """Grab the return code from a `kubectl exec`"""
    kube_args = ['--kubeconfig', kubeconfig]

    if kwargs.get('container'):
        kube_args.extend(['-c', kwargs.get('container')])
    if kwargs.get('namespace'):
        kube_args.extend(['-n', kwargs.get('namespace')])

    kubectl_cmd_tplt = 'kubectl exec {} {} -- {}'

    with host.sudo():
        output = host.run(
            kubectl_cmd_tplt.format(
                pod,
                ' '.join(kube_args),
                ' '.join(command)
            )
        )
        return output
```

You should either reuse it/move it here (with a bit of rework) OR replace it by yours.
But we shouldn't keep both.

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,271 @@
...
+    try:
+        for each in key.split('.'):
+            ptr = ptr[each]
+    except Exception as exc:
+        raise Exception(
+            "Failed generating a traversed dictionary with error: {!s}".format(
+                exc
+            )
+        )
+    return ptr
```
slaperche-scality: I think this could be simplified as:

```python
def traversed_dict(data, key, delimiter='.'):
    return functools.reduce(operator.getitem, key.split(delimiter), data)
```

tests/utils.py
Ebaneck: ok

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,229 @@
...
+        'saltenv=metalk8s-{}'.format(version)
+    ]
+    output = utils.run_salt_command(host, cmd, ssh_config)
+
+    assert output.exit_status == 0, \
+        'Failed applying salt-state {}: \nout: {}\nerr:'.format(
+            state,
+            output.stdout,
+            output.stderr
+        )
```
TeddyAndrieux: I think this assert could be directly in `run_salt_command`

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,229 @@
...
+    except Exception as exc:
+        pytest.fail(
+            "Unable to patch ConfigMap {} in namespace {} with error {!s}"
+            .format(name, namespace, exc)
+        )
+    return response
+
+
+def yaml_load(full_csc, name, namespace):
```
TeddyAndrieux: Minor: Maybe just rename this function as it's not used to load a "yaml" but specific to CSC maybe "csc_load"

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,229 @@
...
+    k8s_client,
+    name,
+    namespace,
+    path,
+    value
+):
+
+    full_csc = read_namespace_config_map(k8s_client, name, namespace)
+    csc_obj = yaml_load(full_csc, name, namespace)
+    csc_obj['spec']['deployment']['replicas'] = int(value)
```
TeddyAndrieux: Nop, use `path` also here

TeddyAndrieux: I also don't really like the hardcoded `int` here because we want to use this same function for others service configuration (that may be not int).

Maybe we should have 2 separate `when` one with quoted `{path}` for string and one without for int

Then you should create an helper to update service configuration that take the value and do no cast it

slaperche-scality: > Maybe we should have 2 separate `when` one with quoted `{path}` for string and one without for int

I think you can use a single `when`, and it can works fine if you use `"foo"` for string and `4` for integer.

`eval` is really bad, but [`literal_eval`](https://docs.python.org/3.6/library/ast.html#ast.literal_eval) ("Safely evaluate an expression node or a string containing a Python literal") is pretty cool.

```
>>> ast.literal_eval('3')
3
>>> ast.literal_eval('"foo"')
'foo'
```

TeddyAndrieux: oh ok i didn't know this one, cool :+1:

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,239 @@
...
+    '''
+    Traverse a nested dict using a delimiter on a target string
+    replaces the value of a key within a dictionary and returns the new dict
+    '''
+    required_key = path.split(delimiter)[-1]
+    for k, v in data.items():
+        if isinstance(v, dict):
+            set_dict_element(v, path, new_value)
+        if k == required_key:
+            data[k] = new_value
```
TeddyAndrieux: What ? No !!

Use `.get` or like I suggest you, you can basically use `get_dict_element` with `path.split(delimiter)[:-1]` (with a join or making `get_dict_element` handle list (which is the cleaner solution IMO, except if we want to properly handle KeyError, but I think we don't really care here))

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,239 @@
...
+    return data
```
TeddyAndrieux: Minor: Btw this return is not mandatory, it's a dict so anyway the input "data" will change

tests/post/steps/test_service_configuration.py
```diff
@@ -0,0 +1,239 @@
...
+
+def get_dict_element(data, path, delimiter='.'):
+    '''
+    Traverse a dict using a 'delimiter' on a target string.
+    getitem(a, b) returns the value of a at index b
+    '''
+    return functools.reduce(operator.getitem, path.split(delimiter), data)
+
+
+def set_dict_element(data, path, new_value, delimiter='.'):
```
TeddyAndrieux: I think `value` is enough do not need to specify `new_`

TeddyAndrieux: Minor: Maybe you could move this two function in `utils` as we may want to use them somewhere else

tests/post/steps/test_service_configuration.py
slaperche-scality: Instead of making `get_dict_element` handling both a list and a string (which is not *that* clean) I think it can also be expressed as:

```python
def set_dict_element(data, path, value, delimiter='.'):
    path, _, key = path.rpartition('.')
    (get_dict_element(data, path) if path else data)[key] = value
    return data
```

Ebaneck: ok

TeddyAndrieux (dismissed):
LGTM

slaperche-scality (dismissed):

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):

PR #2304, opened 2020-03-12
Documentation/install setup

wabernat:
**Component**:

Documentation

**Context**: 

**Summary**:

General language fixes for setup.rst

**Acceptance criteria**: 

if it sounds good, it is good.

Closes no issues.


docs/installation/introduction.rst
```diff
@@ -18,62 +17,61 @@ the following steps:
 #. :doc:`Post installation <./post-install>` configuration steps and sanity
    checks
 
+.. warning::
+
+   MetalK8s is not designed to handle world-distributed multi-site
+   architectures. Instead, it provides a highly resilient cluster at the
+   datacenter scale. To manage multiple sites, look into application-level
+   solutions or alternatives from such the communities as `the Multicluster SIG
```
gdemonet: > from such the communities as

That's weird to read. And a SIG isn't a community itself, it's just a group within Kubernetes community. I'd rather have this sentence read: "or alternatives from the community (e.g. in the scope of the Multicluster SIG)".

docs/installation/introduction.rst
```diff
@@ -18,62 +17,61 @@ the following steps:
...
-.. warning::
-
-   MetalK8s is not designed to handle world-distributed multi-sites
-   architectures. Instead, it focuses on providing a highly resilient cluster
-   at the datacenter scale. To manage multiple sites, look into solutions
-   provided at the application level, or alternatives from the community
-   (such as what `the SIG Multicluster
-   <https://github.com/kubernetes/community/tree/master/sig-multicluster>`_
-   provides).
+Before starting the installation, you must choose an architecture.
```
gdemonet: The meaning changed. It was a recommendation, and became a requirement. The fact is, you don't **need** to have an architecture first - expanding, replacing servers, changing roles, is exactly where MetalK8s is much better than the RING.

docs/installation/introduction.rst
```diff
@@ -84,128 +82,133 @@ offering the finest control over the entire platform:
...
 Variations
 """"""""""
-It is possible to customize the chosen architecture using combinations of
-:ref:`roles<node-roles>` and :ref:`taints<node-taints>`, which are described
-below, to adapt to the available infrastructure.
 
-As a general recommendation, it is easier to monitor and operate well-isolated
-groups of machines in the cluster, where hardware issues would only impact one
-group of services.
+You can customize your architecture using combinations of
```
gdemonet, line 113: I thought our style guide was to never directly address the reader, hence omitting any use of "You". Could we have a clear document to answer this kind of question, so we don't do back-n-forth PRs to change this kind of details?
cc @lucieleonard

docs/installation/introduction.rst
```diff
@@ -84,128 +82,133 @@ offering the finest control over the entire platform:
...
 .. note::
 
    Nodes may belong to both planes, so that one can run applications
    alongside the control plane services.
 
 Control plane nodes often are responsible for providing storage for
-:term:`API Server`, by running :term:`etcd`. This responsibility may be
-offloaded to other nodes from the workload plane (without the ``etcd`` taint).
+API Server, by running :term:`etcd`. This responsibility may be
+offloaded to other nodes from the workload plane (without the etcd taint).
```
gdemonet, line 169: Can we keep the monospace formatting on this value? It expresses a value to use, not just a meaning.

gdemonet: Oh, actually you removed all of this formatting. Can we please revert this? I want to make sure it's obvious that those are not just "terms", but describe the actual value (omitting the common prefix) to input.

docs/installation/introduction.rst
```diff
@@ -258,26 +261,25 @@ To achieve this, use the following commands after deployment:
 
 Networks
 """"""""
+
 A MetalK8s cluster requires a physical network for both the control plane and
 the workload plane Nodes. Although these may be the same network, the
 distinction will still be made in further references to these networks, and
-when referring to a Node IP address. Each Node in the cluster **must** belong
+when referring to a Node IP address. Each Node in the cluster *must* belong
```
gdemonet, line 269: Do we have guidelines for this kind of formatting as well? I honestly prefer bold than italic to emphasize verbs.

docs/installation/introduction.rst
```diff
@@ -288,25 +290,25 @@ Additional Notes
...
 +================+=======================+==========+========+=================================+=============================+
 | bootstrap      | Package repositories, | 1 core   | 2 GB   | Sufficient space for the        |                             |
 |                | container registries, |          |        | product ISO archives            |                             |
 |                | Salt master           |          |        |                                 |                             |
 +----------------+-----------------------+----------+--------+---------------------------------+-----------------------------+
-| etcd           | ``etcd`` database     | 0.5 core | 1 GB   | 1 GB for                        |                             |
-|                | for K8s API           |          |        | ``/var/lib/etcd``               |                             |
+| etcd           | etcd database for the | 0.5 core | 1 GB   | 1 GB for                        |                             |
+|                | K8s API               |          |        | /var/lib/etcd                   |                             |
```
gdemonet, line 312: Definitely this `/var/lib/etcd` is a path, so, monospace.

docs/installation/setup.rst
```diff
@@ -11,131 +11,137 @@ Prerequisites
 .. _SaltStack: https://www.saltstack.com
 .. _Puppet: https://puppet.com
 
-MetalK8s_ clusters require machines running CentOS_ / RHEL_ 7.6 or higher as
+MetalK8s_ clusters require machines running CentOS_//RHEL_ 7.6 or higher as
```
gdemonet: Why double the slash? Doesn't make sense to me.

docs/installation/setup.rst
```diff
@@ -11,131 +11,137 @@ Prerequisites
...
 
-        uname -r
+    $ uname -r
 
-If the installed version is lower than the one above, it must
-be upgraded:
+If the installed version is lower than the one above, upgrade it with:
 
-    .. code-block:: shell
+.. code-block:: shell
```
gdemonet: Inconsistent use of indentation compared to the change above.

gdemonet (changes requested):
Thanks for the rewording effort, it indeed reads better. For the formatting questions however, I'd rather have a consistent guide (that TechPubs may provide) and have us refer to it, than have you or Lucie spend time fixing everything when you go over our docs.

For two reasons:
- it distracts us when reading the diff from the meaningful changes
- we may not agree :wink:

docs/installation/introduction.rst
wabernat: uh... yeah. That "the" has no business in there. That's an error, so good catch. Rewording.

docs/installation/introduction.rst
wabernat: fixed.

docs/installation/introduction.rst
wabernat: I appreciate your attention to the rules, but I have never followed that rule, and I am pushing hard to relegate it to history: it does not make for good writing or reading, hides agency, and makes for long, tortured sentences. 

Let me talk through this with Lucie and I'll revise the rules. Until then, please let this stand.

docs/installation/introduction.rst
wabernat: In this instance, the topic of discussion is the taint, and "etcd" is the taint's name. Again, I may be rebelling against one of my predecessors' edicts here, but it's best to reserve monospace for CLI interactions only. If we are not asking the reader to enter a string into the terminal, or describing what the terminal returns, it should not be set monospace.

docs/installation/introduction.rst
wabernat: I'll revisit our style guide on this. The rule I follow is that emphases are italic, not bold (like the `<em>` tag does). We reserve boldface for GUI elements.

docs/installation/setup.rst
wabernat: That was an attempt to escape the slash. Sphinx behaves oddly when links get stacked next to slashes. Nonetheless, this does not render correctly. Fixed

docs/installation/setup.rst
wabernat: fixed

wabernat:
@gdemonet: I apologize for the confusion, but the style guide you are using is obsolete and we have not yet replaced it. We cannot generate the one you're using from Flare source (Flare machine is on the wrong side of the quarantine), and I cannot locate a PDF copy. Please either send me the PDF you're using or a link to it. Once I have it, I will rewrite it in a briefer and more maintainable format.  

The formatting we're using and enforcing is considerably simpler than what you've been trying to implement. In a nutshell:

* Bold is reserved for elements in a GUI (button, tab legends, e.g.).
* Monospace is reserved for CLI interactions (text that the reader must enter or the machine shows in a CLI).
* Italic is for natural language use: emphasis, titles of works, etc. 

We want to keep things simple. The old style guide had unconventional meanings for bold, italic, bold-italic, bold monospace, etc. that meant nothing to readers and slowed down the writers. Furthermore, dual-formatted sections (bold mono, bold italic, italic hypertext e.g,) are not possible in reStructuredText.

docs/installation/introduction.rst
wabernat: The rule you are following is bad. When I get my hands on it, I will change it. The only reason to set a path as monospace is to indicate that it's something the reader sees or is expected to enter in a command line. For all other mentions, set it roman. Because paths are set with slashes in them, they cannot be mistaken for ordinary language. They are self-evidently paths, and do not need type effects to be easily identified as such by the reader.

gdemonet: https://kubernetes.io/docs/contribute/style/style-guide/#use-code-style-for-filenames-directories-and-paths

That's one topic I'd like us to reconsider.

wabernat: This is not the only editorial decision by Google that gives me heartburn. If you can send me the Scality style guide you're using, I will generate a style guide that we can all live with (with only a few screams from the dungeon).

wabernat:
@bert-e approve

wabernat:
@gdemonet

docs/installation/introduction.rst
gdemonet: OK, let's find some time to gather all the rules TechPubs want us to use, write it somewhere, so we don't need to debate it again :)

gdemonet (approved):
Hey @wabernatScality, really sorry I couldn't find time to come back to you. I understand your point, and would really appreciate having a definitive style guide to make these reviews more efficient in the future :)

> Please either send me the PDF you're using or a link to it. Once I have it, I will rewrite it in a briefer and more maintainable format.

We're not using any Scality-provided style guide at the moment, this was a topic we discussed early in the project but decided could be postponed. Seems like we really need one now :)

PR #2307, opened 2020-03-17
tests: Keep last assert error message when retrying

TeddyAndrieux:
**Component**:

'tests'

**Context**: 

When we use `retry` test function and after all retries the test fail we
do no have any assertion error just a "This failed".

**Summary**:

Keep the last assertion and add it to the error message

---

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -12,10 +12,12 @@ LOGGER = logging.getLogger(__name__)
 
 
 def retry(operation, times=1, wait=1, error_msg=None, name="default"):
+    last_assert = None
     for idx in range(times):
         try:
             res = operation()
         except AssertionError as exc:
+            last_assert = str(exc)
             LOGGER.info(
                 "[%s] Attempt %d/%d failed: %s", name, idx, times, str(exc)
             )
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -12,10 +12,12 @@
 
 
 def retry(operation, times=1, wait=1, error_msg=None, name="default"):
```
gdemonet, line 14: Starting to think this implementation (it's mine, let me shame myself ok) isn't super readable. Maybe we could build some wrapper for running the operation, akin to the way we do `ok, result := SomeOp()` in Go, or how we use `Result` types with `Err` and `Ok` variants in some languages:

```python
def retry(operation, times=1, wait=1, error_msg=None, name="default"):
    def run_and_assert():
        try:
            return None, operation()
        except AssertionError as exc:
            return str(exc), None

    for attempt in range(times):
        error, result = run_and_assert()
        if error is not None:
            LOGGER.info("failed with...")
        else:
            LOGGER.info("success...")
            return result
    else:
        error_msg = (
            error_msg if error_msg is not None
            else default_error_msg(name, attempts, total)
        )
        pytest.fail(error_msg + ": " + error)
```

gdemonet (approved):
I get why we need this (although looking at what `LOGGER.info` outputs should help, if we don't have it maybe we need to setup something in `pytest`...), so here's an approval.

Just thinking the `retry` helper could be improved on a bit.

Issue #2277, opened 2020-03-02
Rewrite Solutions Management (Salt + CLI)


alexandre-allard:
**Component**: salt, scripts

**Why this is needed**:
To implement various changes on how we handle Solutions such as:
  - Remove Environment CRD and replace it by Namespace label (`solutions.metalk8s.scality.com/environment=<environment_name>`)
  - Rework configuration of Operators (need to have some config file, not just env var)
  - Rework configuration of Admin UI (similar to operator, to know which repositories exist for a Solution)
  - Expose available versions to Admin UIs through ConfigMap

**What should be done**:
The previous approach relied on some custom state management, where Salt uses a ConfigMap to store a list of available Solution versions, to compare with the desired state configured in /etc/metalk8s/solutions.yaml.

- Rework configuration file format to let users declare which Solution versions are active (meaning, their Admin UIs / CRDs should be deployed)
- Rework ext_pillar metalk8s:solutions to list available Solutions in the cluster, based on simple observations (active mounts, existing UI Service objects in metalk8s-solutions Namespace...) of the actual state, instead of the ConfigMap used before
- Use a single formula (metalk8s.solutions.available) to (un)mount archives and (un)configure the cluster OCI registry
- Use a single orchestrate (metalk8s.orchestrate.solutions.deploy-components) to (un)deploy Solution components (Admin UI, CRDs)

**Implementation proposal** (strongly recommended):
We need to follow what's inside the Solutions architecture document  https://metal-k8s.readthedocs.io/en/latest/developer/architecture/solutions.html (not yet merged, but can be found as a PR at https://github.com/scality/metalk8s/pull/2246)

**Test plan**:
Manual for now, through the CLI (import and deploy the solution example, create env then populate the env), but we need to add post-merge tests very soon.

gdemonet:
Closed by #2279

PR #2310, opened 2020-03-18
Bugfix/2277 fix solutions management

alexandre-allard:
**Component**: scripts, solutions

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

See: #2277

### DIFF
--- a/scripts/solutions.sh
+++ b/scripts/solutions.sh
@@ -138,7 +138,7 @@ while :; do
     case $1 in
         -a|--archive)
             shift
-            ARCHIVES+=("$1")
+            ARCHIVES+=("$(readlink -f "$1")")
             ;;
         -d|--description)
             shift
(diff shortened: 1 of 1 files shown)


TeddyAndrieux (approved):
LGTM

scripts/solutions.sh
```diff
@@ -261,6 +259,11 @@ configure_archives() {
     local removed=${1:-False}
 
     for archive in "${ARCHIVES[@]}"; do
+        if file "$archive" | grep -vq 'ISO 9660'; then
```
Ebaneck, line 262: Minor and can be omitted
We are searching for the absence of a  fixed string here `ISO 9660` so we can make this faster with `fgrep`?? no??

Ebaneck (approved):

scripts/solutions.sh
slaperche-scality: Or `grep -F` (`fgrep` is deprecated, as explained in the man page).

But I don't think we will observe a perf improvement here ^^

Issue #1605, opened 2019-09-03
Build 'example' solution in CI


NicolasT:
The 'example' solution was imported in https://github.com/scality/metalk8s/pull/1601 but isn't (AFAIK) built as part of a CI run, so the code could bit-rot easily. Furthermore, we need the resulting ISO image to perform any kind of automated testing of solutions.

The building of this 'example' solution should *not* be integrated in the MetalK8s build-system, since they're not related. At the same time, this (easily) allows to add a build-step to the Eve configuration which builds (and uploads) the solution ISO concurrently with the build of MetalK8s.

Next to pure building, setting up some linting, running unit-tests,... would be worth some time.

gdemonet:
Closed by #2312

Issue #1856, opened 2019-10-07
Integration of Solution tests in CI


gdemonet:
**Component**: tests

**Why this is needed**: To maintain quality of the features from the Solutions Lifecycle project.

**What should be done**: Integrate tests introduced by #1850 and #1851

For this to happen we require two nodes to be set up using terraform:

- Spawn a new bootstrap node with MetalK8s installed 
- Spawn a second node for Cypress tests(might require installing the cypress deps)

PR #2312, opened 2020-03-19
Build example solution in the CI

alexandre-allard:
**Component**: ci, build

**Context**: We want to be able to tests the whole Solution framework in the CI

**Summary**: Building of the example-solution in the CI 

**Acceptance criteria**: an example-solution ISO is well generated and available for download in artifacts. 

---

Closes: #1605

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -571,6 +571,48 @@ stages:
             STEP_NAME: buildprev
       - Upload: *upload_final_status_artifact
 
+  build-example-solution:
+    worker:
+      type: kube_pod
+      path: eve/workers/pod-example-solution-builder/pod.yaml
+      images:
+        example-solution-builder: eve/workers/pod-example-solution-builder
+    steps:
+      - ShellCommand: *wait_for_docker
+      - Git: *git_pull
+      - ShellCommand:
+          <<: *add_final_status_artifact_failed
+          env:
+            <<: *_env_final_status_artifact_failed
+            STEP_NAME: build-example-solution
+      - ShellCommand: *setup_cache
+      - SetPropertyFromCommand:
+          name: Set Example Solution version property
+          property: example_solution_version
+          command: >
+            . examples/metalk8s-solution-example/VERSION &&
+            echo "$VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH$VERSION_SUFFIX"
+      - ShellCommand:
+          name: Build Example Solution ISO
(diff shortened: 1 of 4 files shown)


eve/main.yml
```diff
@@ -571,6 +572,59 @@ stages:
...
+            OPERATOR_SDK_VERSION: 'v0.16.0'
+          command: >
+            sudo yum -y remove git perl-Git &&
+            sudo yum -y install https://centos7.iuscommunity.org/ius-release.rpm &&
+            sudo yum -y install genisoimage golang skopeo git2u &&
+            curl -LO https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu &&
+            chmod +x operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu &&
+            sudo mkdir -p /usr/local/bin/ &&
+            sudo cp operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu /usr/local/bin/operator-sdk &&
+            rm -f operator-sdk-$OPERATOR_SDK_VERSION-x86_64-linux-gnu
```
gdemonet: If the deps are really that different, would be worth creating another builder container image IMO

gdemonet: We could also argue that building this ISO is really not "important" and only targetted for a single stage in the post-merge (the stage that checks Solutions, which we'll add in #1856). Hence, just build on the worker VM (no need for the Docker-in-Docker trick for the Pod worker).

gdemonet (commented):
Some comments on how/when we build this ISO

eve/main.yml
```diff
@@ -485,6 +485,7 @@ stages:
           name: Trigger build of previous version
           stage_names:
             - buildprev
+            - build-example-solution
```
Ebaneck: We could make the stage trigger name generic since we are not building the previous version of the example-solution
e.g `Trigger build for previous version and example-solution stages simultaneously` or something along those lines...

alexandre-allard: yep, done

eve/main.yml
alexandre-allard: > If the deps are really that different, would be worth creating another builder container image IMO

Yes, I've created another builder.

Ebaneck (approved):
LGTM

PR #2315, opened 2020-03-20
fix example solution & add --config option

alexandre-allard:
**Component**: examples

**Context**: 

**Summary**:

**Acceptance criteria**: 

---


gdemonet:
Did you try upgrading to the latest Operator SDK?

examples/metalk8s-solution-example/operator/go.mod
```diff
@@ -1,10 +1,11 @@
-module example-operator
+module example-solution-operator
 
 require (
 	github.com/NYTimes/gziphandler v1.0.1 // indirect
 	github.com/go-openapi/spec v0.19.0
 	github.com/operator-framework/operator-sdk v0.9.1-0.20190805223000-66e78cc576ef
 	github.com/spf13/pflag v1.0.3
+	gopkg.in/yaml.v2 v2.2.2
```
gdemonet, line 11: Did you look into [`spf13/viper`](https://github.com/spf13/viper)?

gdemonet: It supports a bunch of features out-of-the-box, such as live reloading of config files, that would help us a lot.

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"gopkg.in/yaml.v2"
+)
+
+type TOperatorConfig struct {
+        Kind         string `yaml:"kind"`
+        APIVersion   string `yaml:"apiVersion"`
```
gdemonet: Those values should be checked before the complete `yaml.Unmarshal`, to ensure we understand the provided format and warn the user properly if we don't.

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
```diff
@@ -109,9 +110,12 @@ func BuildLabelSelector(labels map[string]string) metav1.LabelSelector {
 // buildImageName builds a complete image name based on the version provided
 // for the `base-server` component, which is the only one deployed for now
 func buildImageName(version string) string {
-	prefix, found := os.LookupEnv("REGISTRY_PREFIX")
-	if !found {
-		prefix = "docker.io/metalk8s"
+	var prefix string = "docker.io/metalk8s"
```
gdemonet: Let's define this as a default value for the config and only read from config in controllers.

See https://github.com/spf13/viper#establishing-defaults

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
```diff
@@ -109,9 +110,12 @@ func BuildLabelSelector(labels map[string]string) metav1.LabelSelector {
...
+
+	for _, registry := range config.OperatorConfig.Repositories {
+		if registry.Version == version {
+			prefix = registry.Repository
+		}
```
gdemonet: Not entirely sure about this logic. Consider a complex Operator which not only deploys custom images (which will have the same version as the Operator, probably), but also some open-sourced components (e.g. a well-known database or queue system). The version of said images would not be the same as the `repository.Version`, yet would live in one specific repository.

Mapping between those would need to live somewhere, and I don't know what would be the most flexible approach here. Thoughts @benoit-a @RachedBenMustapha?

examples/metalk8s-solution-example/operator/go.mod
alexandre-allard: No, didn't check this, I will.

alexandre-allard:
> Did you try upgrading to the latest Operator SDK?

No, why ?

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
alexandre-allard: Yeah right, didn't though of other Solutions here.
Will try to find a better format for the repository mapping.

examples/metalk8s-solution-example/operator/go.mod
NicolasT: Why would you want live-reloading of configs? :-/

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,39 @@
...
+        Repositories []struct {
+                Version    string `yaml:"version"`
+                Repository string `yaml:"repository"`
+        } `yaml:"repositories"`
+}
+
+var OperatorConfig TOperatorConfig
```
NicolasT: Not sure some global mutable config 'singleton' is a great pattern.

gdemonet:
> > Did you try upgrading to the latest Operator SDK?
> 
> No, why ?

To see if it changes anything :innocent: we can track this in another ticket, no worries.

examples/metalk8s-solution-example/operator/go.mod
gdemonet: Importing a new Solution version would immediately allow Operators to pick it up if we propagate the repositories in their ConfigMaps.

NicolasT: But that operator will not know about said version (or how to deploy it). So a restart would be required anyway.

In general, an app which starts, loads config, logs it and then uses that (static) config until it gets restarted is much easier to troubleshoot than some app which can dynamically at random points in time change its config, IMHO.

Also, operators are (or should be) trivial to restart (that's the whole point of stateless control loops).

gdemonet: OK, then maybe we don't need those extra features.

I was looking at `prometheus-operator`, and they seem to use the standard [`flag`](https://golang.org/pkg/flag/) package: see [here](https://github.com/coreos/prometheus-operator/blob/3a86bb3115d59db459c92fb54ea63248ee4e4d3a/cmd/operator/main.go#L124-L158)
No way to pass a config file here.

`kubectl` uses `spf13/cobra` and some complex patterns to load its kubeconfig (tried to read through the code but it's confusing for a noob like me), registering all config options with command-line flag overrides.

Since Operator SDK sets up the manager with `spf13/pflag`, it's not too far from how `spf13/viper` works (e.g. see [this PR](https://github.com/operator-framework/operator-sdk/pull/950/files) converting from one to the other). Maybe we don't care about live-reload, but having a config-file with command-line flags overrides is nice for testing or simpler setups.
We can build our own in-house parsing of such files, but it seems the ecosystem often relies on `viper` and `cobra` for Go CLIs, so I wouldn't worry too much about it and just follow the trend.

Also, not sure if there's a standard way of validating an `apiVersion`/`kind` tuple in the tooling controllers rely on, but I'd assume there is...

gdemonet: (of course, all those changes can be considered in a future PR! we just need to think about "real" Operators that will be built around this tooling, and the closer we can get to this with the example, the more likely we'll define better tooling)

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
rachedbenmustapha: @gdemonet for Zenko Operator, since we have all possible cases you mentioned (including being used outside of the solution framework), we use a combination of an image prefix (not yet populated, always empty, but used in all generated ~workflows~ workloads) and explicit repos/versions in a [ZenkoVersoin CR](https://github.com/scality/zenko-operator/blob/master/doc/examples/zenkoversion-1.1-dev.yaml) that is going to be generated in the release process and shipped inside a version's solution ISO.

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: I agree, I've changed this a little bit.

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: I've added a validator, tell me if you have a better idea in mind.

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+type Repository struct {
+	Endpoint string `yaml:"endpoint" validate:"required"`
+	Images   []string `yaml:"images" validate:"required,ne=0"`
+}
+
+type OperatorConfig struct {
+	Kind         string `yaml:"kind" validate:"required,eq=OperatorConfig"`
+	APIVersion   string `yaml:"apiVersion" validate:"required"`
```
gdemonet: Would be nice to validate we know the APIVersion as well ;)

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,153 @@
...
+	case "eq":
+		message += fmt.Sprintf("field '%v' must be equal to '%v'.", field, err.Param())
+	default:
+		message += fmt.Sprintf("field '%v' with tag '%v'", field, err.Tag())
+	}
+
+	return message
+}
+
+func validate() error {
```
gdemonet: Why not make this as a method on any `OperatorConfig` struct?

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,153 @@
...
+			err,
+		)
+	}
+
+	err = validate()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
gdemonet: I'd expect this method to return the config struct if it could load. Something like:

```go
func LoadConfiguration(input io.Reader) (*OperatorConfig, error)  {
```

So we can easily call it with the usual:

```go
config, err := LoadConfiguration(file)
if err != nil {
    // do something with err
}
// consume config
```

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
```diff
@@ -108,16 +109,28 @@ func BuildLabelSelector(labels map[string]string) metav1.LabelSelector {
...
+						prefix = repository.Endpoint
+					}
+				}
+			}
+		}
+	}
+
+	if prefix == "" {
+		return imageName
 	}
```
gdemonet: This will happen if an image is not registered in the list of `Repositories` for the given Solution version. I'd rather fail in the Operator, rather than rely on the fact that _it could work, maybe_.

gdemonet (changes requested):

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -0,0 +1,153 @@
...
+	Images   []string `yaml:"images" validate:"required,ne=0"`
+}
+
+type OperatorConfig struct {
+	Kind         string `yaml:"kind" validate:"required,eq=OperatorConfig"`
+	APIVersion   string `yaml:"apiVersion" validate:"required"`
+	Repositories map[string][]Repository `yaml:"repositories" validate:"required,ne=0"`
+}
+
+var operatorConfig OperatorConfig
```
slaperche-scality: Kept the approach with a global (but made it private to limit the risk) because we couldn't find a way to thread something from the main to the reconciliation loop (the loop has access to a `reconcile.Reconciler` and the only thing we get when creating it is the `manager.Manager`)

If you have a better suggestion, we're open for it.

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: Yes, sure.

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
alexandre-allard: Yes, wasn't sure how to handle this.
Will change it and make it fail.

examples/metalk8s-solution-example/operator/pkg/config/config.go
slaperche-scality: After discussing with @benoit-a I've implemented [this](https://github.com/scality/metalk8s/compare/bugfix/fix-example-solution...user/slaperche/fix-example-solution?expand=1) which allows us to get rid of the global config and thread the config variable from the main to the reconciler.

But this require some modification in the code generated by the SDK, thus I don't know if that's OK for you @gdemonet / @NicolasT or if you prefer the "private global" approach?
Or if you have a 3rd way :slightly_smiling_face: ?

alexandre-allard: I think your Solution is better and we can change the code generated, it's not a problem, but I'm not a Go expert, so waiting for feedback from them.

slaperche-scality: Yeah it's not a problem, except that it may make the upgrade of the SDK a bit more tedious (but  it's already a very manual process).

alexandre-allard: I've cherry-picked your commit.

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: done

examples/metalk8s-solution-example/operator/pkg/controller/util/baseserver.go
alexandre-allard: done

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: done

examples/metalk8s-solution-example/operator/pkg/config/config.go
alexandre-allard: done

examples/metalk8s-solution-example/operator/pkg/config/config.go
```diff
@@ -80,15 +78,15 @@ func (opConfig *OperatorConfig) validate() error {
...
 	})
 
-	err := validate.Struct(opConfig)
+	err := validate.Struct(config)
 
 	if err != nil {
 		for _, e := range err.(validator.ValidationErrors) {
 			errorList = append(errorList, getErrorMessage(e, ""))
 		}
 	}
```
slaperche-scality: Minor: could use the short form (like l.135 of `main.go`)

```go
	if err := validate.Struct(config); err != nil {
		for _, e := range err.(validator.ValidationErrors) {
			errorList = append(errorList, getErrorMessage(e, ""))
		}
	}
```

examples/metalk8s-solution-example/config.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: solutions.metalk8s.scality.com/v1alpha1
+kind: SolutionConfig
```
gdemonet, line 2: Would be cool to show everything we support in this file. Next PR!

gdemonet (approved):
LGTM!

examples/metalk8s-solution-example/config.yaml
alexandre-allard: Yes, good idea.

PR #2316, opened 2020-03-20
[security] Upgrade PyYAML from 5.1 to 5.3.1 (and other dependencies)

gdemonet:
Original change was proposed by Snyk. PR format was invalid however, and
edit of the requirements required running the `pip-compile` tox target.

The following vulnerabilities are fixed by pinning transitive dependencies:
- https://snyk.io/vuln/SNYK-PYTHON-PYYAML-559098

All integration branches were updated (by running `tox -e pip-compile` and bumping PyYAML version in `.in` requirement files).

---
Original PR: #2314


slaperche-scality (changes requested):
That's not the only issue with Snyk's PR…

- we should also update the buildchain's requirements (it also uses PyYAML)
- if it's really a security issue it should probably be fixed in the oldest branches (for tests I don't know which one it is, but for buildchain we use PyYAML since 2.5)

Also it's still not clear to me that we're affected by the mentioned vulnerabilities (but anyway, it won't hurt to update).

gdemonet:
> That's not the only issue with Snyk's PR…
> 
> * we should also update the buildchain's requirements (it also uses PyYAML)

Indeed, will go through the other places where PyYAML is used.

> * if it's really a security issue it should probably be fixed in the oldest branches (for tests I don't know which one it is, but for buildchain we use PyYAML since 2.5)

Ah right, the Snyk PR was targeting 2.0 already (so I changed the target of this branch). Maybe that's why buildchain was not changed by Snyk.

> Also it's still not clear to me that we're affected by the mentioned vulnerabilities (but anyway, it won't hurt to update).

Indeed, we probably aren't (it's affecting the `full_load` method, which I don't recall us using - but we depend on libraries which use `PyYAML`, so better upgrade than be sorry later).

TeddyAndrieux (dismissed):

slaperche-scality (dismissed):
What about ` buildchain/requirements-Darwin.txt`?

gdemonet:
Ah, right. I'll borrow a Mac.

TeddyAndrieux (approved):

PR #2318, opened 2020-03-21
Feature/2278 update solutions ui

### LINKED ISSUE #2278: Adapt Solutions UI
**Component**: ui

**Why this is needed**:
Because we change a lot of stuff on how we manage Solutions in #2277 

**What should be done**: 

- Be able to create an environment from UI, we need to warn the users before they delete the environment.

- Be able to add the solution from the available solutions in the specific environment.
The following screenshot is the current UI in order to give a context:
image
1. After click on the `add the solution` button, we should display a loader and notice the user it's preparing the environment. [sketch1](https://user-images.githubusercontent.com/18453133/76859496-ba961f00-6859-11ea-9601-dc037bfb7d71.png)

2. When the preparation is done,  we could use a collapsible panel inside the solution column which could display all the solutions name+version and trash button ( for this PR, we won't implement the remove the solution from the specific environment, but the idea is here)  [sketch2](https://user-images.githubusercontent.com/18453133/76855711-00e78000-6852-11ea-84f0-09e8797a3aff.png)

1. import the solution ISO
2. activate the solution:
`./solutions.sh activate --name example-solution --version 0.1.0-dev`
3. create an environment
4. click on the `add a solution` button

**Test plan**:
Cypress test: Continue what Carlito did for the solution UI.
Will handle Cypress test in another PR

 **Next step**:
We will handle `delete an environment` in another PR.

ChengYanJin:
**Component**: ui, solution

**Context**:  Because we change a lot of stuff on how we manage Solutions in #2277

**Summary**:
 - From UI, I am able to create an environment
- From UI,  I am able to add the solution from the available solutions in the specific environment.
- If the solution deployment is done,  we should be able to see the solution name with version displayed in the specific environment row
- From UI, I am able to delete the Environment by click on the trash button 

**Acceptance criteria**: 

1. build the solution ISO

   Go to `bugfix/fix-example-solution` branch to make iso.

 - To get a new version of the same Solution, I edited the VERSION file
 - To get a "new" Solution, I passed PRODUCT_NAME="Other Solution" PRODUCT_LOWERNAME="other-solution" to make iso

   Copy the iso to `_build/root`

2. import the solution ISO
`./solutions.sh import --archive </path/to/solution.iso>`
3. activate the solution:
`./solutions.sh activate --name <solution-name> --version <solution-version>`
4. create an environment
5. click on the add a solution button

screenshots:
1. create a new environment: [image1](https://user-images.githubusercontent.com/18453133/77231089-f3573080-6b98-11ea-93ac-d3c633e0d652.png)
2. add a solution: [image2](https://user-images.githubusercontent.com/18453133/77231213-da9b4a80-6b99-11ea-8c79-8ede3ea73df4.png)

---

Closes: #2278


gdemonet (commented):
Some general comments first, related to previous commits you had to work on top of:

- there are still some commits left as `WIP`: probably worth to rename them if they are now valid (from your standpoint), or simply remove them if irrelevant
- the `[WIP] ui: Run app behind /metalk8s` shouldn't be done in this PR I believe, not related to the Solutions page IMHO (would make sense in some "Shell integration" PR though :wink:) (the `serve fonts CSS...` commit should move out as well)

ui/src/translations/en.json
```diff
@@ -135,5 +136,10 @@
   "please_refer_to": "Please refer to",
   "import_solution_hint": "You should import a solution to prepare your environment",
   "redirecting": "Redirecting…",
-  "documentation": "Documentation"
+  "documentation": "Documentation",
+  "env_preparation": "Environment preparation",
+  "env_preparation_success": "Environment {envName} is being prepared.",
+  "env_preparation_failed": "Environemnt {envName} preparation failed.",
```
gdemonet:
```suggestion
  "env_preparation_failed": "Environment {envName} preparation failed.",
```

ui/src/translations/en.json
```diff
@@ -135,5 +136,10 @@
...
+  "preparing_environemnt": "preparing the environment",
```
gdemonet:
```suggestion
  "preparing_environment": "preparing the environment",
```

gdemonet: Also, not sure why we need this key (aren't `env_preparation` / `env_preparation_success` sufficient?)

ui/src/translations/en.json
```diff
@@ -135,5 +136,10 @@
...
+  "environment_creation_failed": "Environment {name} creation has failed."
```
gdemonet: For consistency with other keys:

```suggestion
  "env_creation_failed": "Environment {envName} creation has failed."
```

ui/src/translations/fr.json
```diff
@@ -135,5 +136,10 @@
   "please_refer_to": "Veuillez vous référer à",
   "import_solution_hint": "Vous devez importer une solution pour préparer votre environnement",
   "redirecting": "Redirection…",
-  "documentation": "Documentation"
+  "documentation": "Documentation",
+  "env_preparation": "Environment preparation",
```
gdemonet:
```suggestion
  "env_preparation": "Préparation d'environnement",
```

ui/src/translations/fr.json
```diff
@@ -135,5 +136,10 @@
...
+  "env_preparation_success": "L'environnement {envName} est en cours de préparation.",
+  "env_preparation_failed": "La préparation de l'environnement {envName} a échoué.",
+  "preparing_environemnt": "préparer l'environnement",
```
gdemonet: If this is supposed to mean the same thing as the english variant, it should translate to:

```suggestion
  "preparing_environemnt": "environnement en cours de préparation",
```

ui/src/translations/fr.json
```diff
@@ -123,7 +124,7 @@
   "create_new_environment": "Créer un nouvel environnement",
   "description": "Description",
   "add_solution_to_environment": "Ajout d'une solution à l'environnement {environment}",
-  "version_env": "Versions",
+  "version_env": "Version",
```
slaperche-scality, line 127: Why `version_env` instead of `version`?

ui/src/translations/en.json
slaperche-scality: `has failed` (to be consistent with the message for `environment_creation_failed`)

ui/src/translations/fr.json
slaperche-scality: Probably not a correct French translation, what's the context/use for this string?
`Préparation de l’environnement` maybe

ui/src/translations/fr.json
slaperche-scality: This one is a bit weird too, what's the use/context?

If it's for a progress bar or some kind of load, then `Préparation de l’environnement…` would be better (and English could be `Preparing the environment…`)

ui/src/services/k8s/solutions.js
```diff
@@ -96,18 +98,57 @@ export async function getEnvironmentAdminUIs(environment) {
...
     );
+  } catch (error) {
+    return { error };
+  }
+}
 
-    if (!result.error) {
-      // Return the first one found among the Environment namespaces
-      return { ...result?.data };
-    }
```
gdemonet: Not sure this is a good change. Under `services/k8s/solutions`, the goal is to have methods that are specific to Solutions, while keeping generic things under `services/k8s/core` (but could also be split up and extended if required).

This method is retrieving the "environment ConfigMap" for a given environment. It's built on the premise that a caller has a access to an "environment" object, as returned by `listEnvironments` above.
Such "environments" can have multiple `namespaces`, and this is why the `getEnvironmentConfigMap` goes through this list and returns the first CM it finds. Not the best solution (merging them into a higher-level structure would be great, but takes time, so, later :innocent:), but at least we have room for improvement directly in this helper method.

If you really want helpers around `get/patch/create` for ConfigMaps, please add them to `services/k8s/core` instead, and build on them in this file.

ui/src/translations/en.json
```diff
@@ -96,6 +96,7 @@
   "no_storage_class_found": "No storage class found.",
   "storage_class_is_required": "A storage class is required in order to create a volume. Please create one in Kubernetes.",
   "learn_more": "Learn more",
+  "solution": "Solution",
```
slaperche-scality, line 99: Don't we have something to handle the pluralization in the framework we use fo the translation?

Edit: Yeah, not sure that would help in that case.

ui/src/services/k8s/solutions.js
```diff
@@ -96,18 +98,57 @@ export async function getEnvironmentAdminUIs(environment) {
...
+    metadata: {
+      name: ENVIRONMENT_CONFIGMAP_NAME,
+    },
+  };
+  try {
+    return await coreV1.createNamespacedConfigMap(namespace, body);
+  } catch (error) {
+    return { error };
   }
 }
```
gdemonet: Creating an environment ConfigMap could benefit from some pre-filled `data` keys in the future, so let's make this a helper method:

```javascript
export async function createEnvironmentConfigMap(namespace) {
  // Add default values in here
  const data = {};

  // Note to YanJin: this `core.createConfigMap` helper would need
  // to be defined in `services/k8s/core`
  return core.createConfigMap(ENVIRONMENT_CONFIGMAP_NAME, namespace, data);
}
```

ui/src/services/k8s/solutions.js
```diff
@@ -96,18 +98,57 @@ export async function getEnvironmentAdminUIs(environment) {
...
+      {
+        headers: {
+          'Content-Type': 'application/merge-patch+json',
+        },
+      },
+    );
+  } catch (error) {
+    return { error };
+  }
+}
```
gdemonet: Same principles here: define the generic "patch ConfigMap" method in `core`, and build on it in `solutions` for just your needs!

```javascript
export async function addSolutionToEnv(namespace, solName, solVersion) {
  const patch = { op: "add", path: `/data/${solName}`, value: solVersion };

  return core.patchConfigMap(
    ENVIRONMENT_CONFIGMAP_NAME, namespace,
    // use an object instead of many arguments, for easier support of
    // multiple methods (e.g. both `jsonPatch` and `mergePatch` could be
    // allowed, headers being inferred from which is used).
    { jsonPatch = [patch] },
  );
}

export async function updateSolutionInEnv(namespace, solName, solVersion) {
  const patch = { op: "replace", path: `/data/${solName}`, value: solVersion };

  return core.patchConfigMap(
    ENVIRONMENT_CONFIGMAP_NAME, namespace, { jsonPatch = [patch] },
  );
}

export async function removeSolutionFromEnv(namespace, solName) {
  const patch = { op: "remove", path: `/data/${solName}` };

  return core.patchConfigMap(
    ENVIRONMENT_CONFIGMAP_NAME, namespace, { jsonPatch = [patch] },
  );
}
```

ui/src/translations/fr.json
ChengYanJin: Thanks for the French speaker^^
The context of `preparing_environemnt` is that we want to display a text next to the loader when we are waiting for preparing the environment.

ui/src/services/k8s/solutions.js
```diff
@@ -96,18 +98,57 @@ export async function getEnvironmentAdminUIs(environment) {
...
+
+export async function getNamespacedDeployment(name, namespace) {
+  try {
+    return await appsV1.readNamespacedDeployment(name, namespace);
+  } catch (error) {
+    return { error };
+  }
+}
```
gdemonet: Definitely a helper to put in `core` (or `app`, if you want to add one, but I'd say `core` is for everything "core" K8s, while others are for Metal-specific concepts).

ui/src/translations/fr.json
ChengYanJin: because we have another translation  `"version": "MetalK8s Version",`

gdemonet (changes requested):
Early dump of comments, mostly related to the `services/k8s` helpers.

IMO, we should make sure that:

- as much business logic for managing APIs (such as K8s or Prom) as possible should go into the `services` helpers
- ducks are UI logic: "when this action was dispatched, trigger this call to a `services` helper, and let it do the business logic" - "when this call completed, store the result in the state, and trigger a notification"
- we don't put some logic in both places

The special logic we need for Solutions and Volumes are worth extracting entirely into `services/k8s/solutions` and `s/k/volumes` respectively. Ducks can then rely on the helpers, making sure that if for some reason we change the backend implementation of any of them, the helpers can hide this change behind a constant format, and ducks will remain unchanged.

ui/src/translations/fr.json
slaperche-scality: Then `"version": "MetalK8s Version"` should probably be renamed `"metalk8s_version": "MetalK8s Version"` instead

ui/src/services/k8s/solutions.js
ChengYanJin: Good point!!! We may have multiple namespaces in one environment. I will just as you said leave space for future improvement, anyway in UI we don't support multiple namespaces yet.

ui/src/translations/en.json
ChengYanJin: ok. At the beginning, I would like to have this prop used for the text of the loader.
But yeah, it makes sense to reuse`env_preparation_success`

ui/src/services/k8s/solutions.js
ChengYanJin: Good to know the jsonPatch!!!

ui/src/translations/en.json
gdemonet: Oh OK I see! So no, let's keep a loading message as this one.

However, `env_preparation_success` should show a "successful" status, no? If yes, maybe reword to show:
- in English: `Environment {envName} preparation succeeded.`
- in French: `La préparation de l'environnement {envName} a réussi.`

The current message for `_success` is actually meaning something is `_in_progress` ;)

ui/src/ducks/app/solutions.js
```diff
@@ -65,52 +78,184 @@ export function createEnvironmentAction(newEnvironment) {
...
-        ...environments,
-        [env.name]: {
-          adminUIs: call(SolutionsApi.getEnvironmentAdminUIs, env),
-          config: call(SolutionsApi.getEnvironmentConfigMap, env),
-        },
-        };
+  const jobs = yield select(state => state.app.salt.jobs);
+  const preparingEnvs = jobs
+    ?.filter(job => job.name.startsWith('prepare-env/') && !job.completed)
+    .map(job => job.name.replace(/^(prepare-env\/)/, ''));
```
gdemonet: Can we use `job.type === "prepare-env"` instead of prefixing the `job.name`, as we did for Node deployments?

ui/src/translations/en.json
ChengYanJin: Ok. For the loading message, I will keep the previous one.

However, I don't think it's a good idea to change the text of `env_preparation_success`. Probably we should change the key to `env_preparation_in_progress` though.

I am going to give the context. The text will be displayed as the notification after the Salt job is done. So in theory, we couldn't say it's already succeeded because we still need to check the deployment of the solution.

So IMO, `Environment {envName} is being prepared.` is better. What do you think?

ui/package.json
```diff
@@ -7,7 +7,9 @@
     "@kubernetes/client-node": "github:scality/kubernetes-client-javascript.git#browser-0.10.2-62-g61d8543",
     "@scality/core-ui": "github:scality/core-ui.git#v0.5.0",
     "axios": "^0.18.0",
+    "eslint": "6.8.0",
```
gdemonet: Is this really a project dependency? Or just one for `devDepencies`? ;)

ui/src/containers/SolutionList.js
```diff
@@ -60,13 +66,14 @@ const FormStyle = styled.div`
...
-`;
-
 const TableContainer = styled.div`
   height: 40%;
   margin: 0 0 50px 0;
+
+  /* solve the tooltip display issue */
+  .sc-table-column-cell-action {
+    overflow: visible !important;
+  }
```
gdemonet: Something wrong with this? Ticket in `core-ui` maybe? If there's one already, having it in the comment can help :)

ui/src/containers/SolutionList.js
```diff
@@ -135,56 +165,68 @@ const SolutionsList = props => {
...
+      renderer: (solutions, environment) => {
+        const isEnvironmentPreparing = environment.isPreparing;
+        const deployedSolutions = environment.solutions;
+        const solutionsList =
+          deployedSolutions &&
+          deployedSolutions.map((deployedSolution, idx) => {
+            return (
+              <span
+                key={idx}
+              >{`${deployedSolution.name}(v.${deployedSolution.version})`}</span>
```
gdemonet: minor: let's show a space after the solution name
```suggestion
              >{`${deployedSolution.name} (v.${deployedSolution.version})`}</span>
```

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
 export const solutionServicesSelector = state => state.app.solutions.services;
 
 // Sagas
+export function* fetchEnvironments() {
+  const jobs = yield select(state => state.app.salt.jobs);
+  const preparingEnvs = jobs
+    ?.filter(job => job.name.startsWith('prepare-env/') && !job.completed)
+    .map(job => job.name.replace(/^(prepare-env\/)/, ''));
+  const environments = yield call(SolutionsApi.listEnvironments);
+  const updatedEnvironments = yield call(updateEnvironments, environments);
```
gdemonet, line 106: Is it useful to split this "hydration" step from the original `listEnvironments`, in the end?

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
+  const resultCreateEnvironment = yield call(
+    SolutionsApi.createEnvironment,
+    action.payload,
+  );
+
+  if (!resultCreateEnvironment.error) {
+    const resultCreateNamespacedConfigMap = yield call(
+      SolutionsApi.createNamespacedConfigMap,
+      name,
+    );
```
gdemonet: Could be part of `SolutionsApi.createEnvironment` I believe. Especially since we're not sending any specific information to the user about an error, would it occur.

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
 }
 
-export function* fetchUIServices() {
-  const result = yield call(ApiK8s.getUIServiceForAllNamespaces);
-  if (!result.error) {
-    yield put(setServicesAction(result.body.items));
+export function* prepareEnvironment(action) {
+  const { envName, solution } = action.payload;
+  const addedSolutionName = solution.solution.value;
+  const addedSolutionVersion = solution.version.value;
```
gdemonet: We're extracting `.value` because it comes from a form field? I'd argue that's not the sagas' responsibility.
In general, we should strive to only give each function what it needs and nothing else. In this case, extracting the `.value`s into a usable payload should be done in the component which displays the form and triggers the saga.

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
+        }),
+      );
+      yield call(watchPrepareJobs);
+    }
+  } else {
+    yield put(
+      addNotificationErrorAction({
+        title: intl.translate('prepare_environment'),
+        message: intl.translate('env_preparation_failed', { envName }),
+      }),
```
gdemonet: This was an error when adding the Solution to the Env's ConfigMap. Maybe another message could be useful here.

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
+        const solutionOperatorDeployment = yield call(
+          CoreApi.getNamespacedDeployment,
+          `${solution}${OPERATOR_}`,
+          env.name,
+        );
+        const solutionUIDeployment = yield call(
+          CoreApi.getNamespacedDeployment,
+          `${solution}${UI_}`,
+          env.name,
+        );
```
gdemonet, line 205: Mmmh, we should not rely on the name of these resources, as it may change. Instead, let's use sufficient selectors (cc @alexandre-allard-scality - maybe you know which ones we set?).

ui/src/ducks/app/solutions.js
```diff
@@ -67,113 +80,257 @@ export function createEnvironmentAction(newEnvironment) {
...
+      }
     }
   }
 }
 
-export function* stopRefreshSolutions() {
-  yield put(setSolutionsRefeshingAction(false));
+export function* deleteEnvironment(action) {
+  const envName = action.payload;
+  const result = yield call(CoreApi.deleteEnvironment, envName);
```
gdemonet: This `deleteEnvironment` should live in `SolutionsApi`, not `CoreApi`. "Environments" are not even a concept in K8s world, it's just our creation :)

ui/src/services/k8s/core.js
```diff
@@ -0,0 +1,129 @@
...
+      undefined,
+      {
+        headers: {
+          'Content-Type': 'application/merge-patch+json',
+        },
+      },
+    );
+  } catch (error) {
+    return { error };
+  }
```
gdemonet: I think there was a misunderstanding here. JSON Patch **is** supported by K8s API. You just need to send it with the right header:

```javascript
export async function patchNamespacedConfigMap(name, namespace, { jsonPatch, mergePatch }) {
  if (jsonPatch !== undefined) {
    cTypeHeader = 'application/json-patch+json';
    body = jsonPatch;
  } else {
    cTypeHeader = 'application/merge-patch+json';
    body = mergePatch;
  }

  try {
    return await coreV1.patchNamespacedConfigMap(
      name,
      namespace,
      body,
      undefined,
      undefined,
      { headers: { 'Content-Type': cTypeHeader } },
    );
  } catch (error) {
    return { error };
  }
}
```

gdemonet: (which means you could also get rid of the `jsonpatch` dependency)

ui/src/services/k8s/core.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export async function getNamespacedDeployment(name, namespace) {
+  try {
+    return await appsV1.readNamespacedDeployment(name, namespace);
+  } catch (error) {
+    return { error };
+  }
+}
```
gdemonet, line 128: Not convinced about the need for such transparent helpers, but we can leave it as is.

gdemonet (changes requested):
Some general comments on usability of the feature (most of these remarks can be dispatched in future iterations):

- `solutions.sh import -a /path/to/my.iso` doesn't put it in the `metalk8s-solutions` ConfigMap, which means UI needs a call to `solutions.sh activate ...` before seeing a newly added Solution ISO (that's not related to UI, more of a Salt problem I think)
- upgrading a Solution instance in an Environment is tricky: you need to **Add** the new version, which will override the old one - IMO, that's counter-intuitive
- there's a bug in the "add Solution to Env" modal: if I have Solution 1 with version A, switching to Solution 2 will keep version A selected, even if there's no version A for Solution 2 - I can submit the form, and the Salt job then fails because the version is not available
- notifications are bugged for me: when a "prepare" job fails, I get the error notification twice - when it succeeds, I get  **four** success notifications at once!
- not sure why exactly, but I deployed two Solution instances in my environment, and only one shows up in the table (the first one I deployed)

ui/src/ducks/app/solutions.js
alexandre-allard: I don't see any useful label for this, especially for the Operator which contains only a single label refering to the MetalK8s cluster version:
```
metadata:
  [...]
  labels:
    metalk8s.scality.com/version: 2.4.3-dev
```
For the UI part, there are labels, but they're all computed on the same basis as the deployment name... so, not more reliable than the deployment name:
```
metadata:
  [...]
  labels:
    app: example-solution-ui
    app.kubernetes.io/component: ui
    app.kubernetes.io/managed-by: salt
    app.kubernetes.io/name: example-solution-ui
    app.kubernetes.io/part-of: example-solution
    app.kubernetes.io/version: 0.1.0-dev
    heritage: example-solution
    metalk8s.scality.com/version: 2.4.3-dev
```
Computed from:
```
metadata:
  name: {{ solution }}-ui
  namespace: {{ namespace }}
  labels:
    app: {{ solution }}-ui
    heritage: {{ solution }}
    app.kubernetes.io/name: {{ solution }}-ui
    app.kubernetes.io/version: {{ version }}
    app.kubernetes.io/component: ui
    app.kubernetes.io/part-of: {{ solution }}
    # UIs are deployed and managed by Salt, provided with MetalK8s
    app.kubernetes.io/managed-by: salt
```
Should we had some labels to these ?

ui/src/containers/SolutionList.js
ChengYanJin: Yeah. I just create a ticket in core-ui: https://github.com/scality/core-ui/issues/176
Will put it in the comment as well

ui/src/ducks/app/solutions.js
ChengYanJin: Response from Alexandre:
I don't see any useful label for this, especially for the Operator which contains only a single label refering to the MetalK8s cluster version...

For the UI part, there are labels, but they're all computed on the same basis as the deployment name... so, not more reliable than the deployment name:
```
  labels:
    app: example-solution-ui
    app.kubernetes.io/component: ui
    app.kubernetes.io/managed-by: salt
    app.kubernetes.io/name: example-solution-ui
    app.kubernetes.io/part-of: example-solution
    app.kubernetes.io/version: 0.1.0-dev
    heritage: example-solution
    metalk8s.scality.com/version: 2.4.3-dev
```
Computed from:

```
metadata:
  name: {{ solution }}-ui
  namespace: {{ namespace }}
  labels:
    app: {{ solution }}-ui
    heritage: {{ solution }}
    app.kubernetes.io/name: {{ solution }}-ui
    app.kubernetes.io/version: {{ version }}
    app.kubernetes.io/component: ui
    app.kubernetes.io/part-of: {{ solution }}
    # UIs are deployed and managed by Salt, provided with MetalK8s
    app.kubernetes.io/managed-by: salt
```

ui/src/containers/SolutionList.js
ChengYanJin: > Something wrong with this? Ticket in `core-ui` maybe? If there's one already, having it in the comment can help :)

Oh. Sorry, I should have removed those tooltip things. 
At first, I wanted to add a tooltip for the trash button. 
We may need the tooltip later, but I created a ticket in core-ui already!

ui/src/ducks/app/solutions.js
gdemonet: These labels are what we should rely on IMO, not the names (we may want to change them for whatever reason). If I'm looking for _the UI of Solution "toto" in version "1.0.0", deployed in the Env namespace "prod"_, I'll do the query:

```shell
kubectl get deploy -n prod \
    -l app.kubernetes.io/part-of=toto \
    -l app.kubernetes.io/component=ui \
    -l app.kubernetes.io/version=1.0.0
```

ui/src/translations/en.json
gdemonet: > The text will be displayed as the notification after the Salt job is done. So in theory, we couldn't say it's already succeeded because we still need to check the deployment of the solution.

That's where I disagree ;) When Salt returns after the `prepare-environment`, it means (at least, from Salt PoV) that the env has been prepared.
It may not mean that all resources (UI/Operator) are available in the current implem, but I'd rather add this check on Salt side than needing an extra step in UI or CLI.

ChengYanJin:
> Some general comments on usability of the feature (most of these remarks can be dispatched in future iterations):
> 
> * `solutions.sh import -a /path/to/my.iso` doesn't put it in the `metalk8s-solutions` ConfigMap, which means UI needs a call to `solutions.sh activate ...` before seeing a newly added Solution ISO (that's not related to UI, more of a Salt problem I think)
> * upgrading a Solution instance in an Environment is tricky: you need to **Add** the new version, which will override the old one - IMO, that's counter-intuitive
> * there's a bug in the "add Solution to Env" modal: if I have Solution 1 with version A, switching to Solution 2 will keep version A selected, even if there's no version A for Solution 2 - I can submit the form, and the Salt job then fails because the version is not available
> * notifications are bugged for me: when a "prepare" job fails, I get the error notification twice - when it succeeds, I get  **four** success notifications at once!
> * not sure why exactly, but I deployed two Solution instances in my environment, and only one shows up in the table (the first one I deployed)

Thanks for the review!
Please find the following responses which map to each of the points.

-  Alexandre will help me to handle this, will do in a sperate PR
-  I got your point, how do we handle the upgrading the solution instances I will create a ticket and do it in another PR.
- Fix the bug in `ae07f9b2 - ui/solutions: Fix the bug for Formik form`
- Fix the bug in `ff7fc846 - ui/solutions: Fix the bug that it displays multiple notifications at the same time`
- Fix the bug in `045f3feb - ui/solutions: Fix the display bug when there are more than one solutions deployed`

For the other's comments, I addressed them all but the jsonPatch, for some reason, I got 422 when I patch...

ui/src/ducks/app/solutions.js
```diff
@@ -330,12 +297,41 @@ export function* deleteEnvironment(action) {
...
+}
+
 // Sagas
 export function* solutionsSaga() {
   yield all([
     fork(refreshSolutions),
     takeEvery(CREATE_ENVIRONMENT, createEnvironment),
     takeEvery(PREPARE_ENVIRONMENT, prepareEnvironment),
     takeEvery(DELETE_ENVIRONMENT, deleteEnvironment),
+    takeEvery(JOB_COMPLETED, notifyDeployJobCompleted),
```
gdemonet, line 324: Obviously! Don't know why I made my own with `while (true) { take(JOB_COMPLETED); ... }`... :man_facepalming:

ui/src/ducks/app/solutions.js
```diff
@@ -175,8 +177,10 @@ export function* prepareEnvironment(action) {
   } else {
     yield put(
       addNotificationErrorAction({
-        title: intl.translate('prepare_environment'),
-        message: intl.translate('env_preparation_failed', { envName }),
+        title: intl.translate('add_solution_to_configMap', {
```
gdemonet: Let's be consistent and keep translation keys always lowercase please :) so `configmap`, not `configMap` (here and everywhere else, of course)

gdemonet:
> the jsonPatch, for some reason, I got 422 when I patch...

What's the error message? Maybe the patch body is not in the right format, or some keys can't be patched as we'd like to...

ChengYanJin:
> > the jsonPatch, for some reason, I got 422 when I patch...
> 
> What's the error message? Maybe the patch body is not in the right format, or some keys can't be patched as we'd like to...

Done with`12ca05f3 - ui/solutions: Apply json patch provided by k8s api to patch solution for configmap`

ChengYanJin:
> > > the jsonPatch, for some reason, I got 422 when I patch...
> > 
> > 
> > What's the error message? Maybe the patch body is not in the right format, or some keys can't be patched as we'd like to...
> 
> Done with `12ca05f3 - ui/solutions: Apply json patch provided by k8s api to patch solution for configmap`

ui/src/ducks/app/solutions.js
ChengYanJin: In the end, I still prefer to sperate the `updateEnvironments` from `fetchEnvironments`.
For me the purpose of updateEnvironments is to update the Environment with the deployed solutions it's a bit more than fetchEnvironments IMO

ui/src/services/k8s/core.js
```diff
@@ -79,10 +78,12 @@ export async function queryPodInNamespace(namespace, podLabel) {
 
 export async function createNamespacedConfigMap(name, namespace, restProps) {
   const body = {
+    // it actually be removed by the k8s when the data is empty.
```
slaperche-scality: Since it's removed, maybe you don't need to put it (anyway you have to create it with the first patch)

ui/src/services/k8s/solutions.js
```diff
@@ -119,7 +119,17 @@ export async function getEnvironmentConfigMap(environment) {
 }
 
 export async function addSolutionToEnvironment(namespace, solName, solVersion) {
-  const patch = [{ op: 'add', path: `/data/${solName}`, value: solVersion }];
+  const result = await coreV1.readNamespacedConfigMap(
+    ENVIRONMENT_CONFIGMAP_NAME,
+    namespace,
+  );
+  let patch;
+  // the first patch we need to create data key. otherwise we will get 422 patch error.
```
slaperche-scality: Or simply: `// Create the `data` key if it doesn't exists, otherwise we can't insert the solution field inside`

ui/src/services/k8s/core.js
```diff
@@ -79,10 +78,12 @@ export async function queryPodInNamespace(namespace, podLabel) {
 
 export async function createNamespacedConfigMap(name, namespace, restProps) {
```
slaperche-scality, line 79: In the commit message:

- `create the data object,` => `create the field "data"` (object is a bit ambiguous in this context).
- `we could using the path` => `we can use the path`

slaperche-scality (commented):
Some minor comments (mainly about comments ^^)

ui/src/ducks/app/solutions.js
ChengYanJin: I will create a debt ticket in order to check if we have the deployed UI and Solution service

gdemonet (approved):
Didn't test manually again, but let's merge it and test all of this automatically in post-merge :)

Great job @ChengYanJin ! :clap:

PR #2320, opened 2020-03-23
Enforce the usage of --kubeconfig flag in Solutions script

Ebaneck:
**Component**:

'kubernetes'

**Context**: 

While using Solutions.sh script to provision new solutions, users fall into the following situations:
The connection to the server localhost:8080 was refused - did you specify the right host or port?

This is likely because we didn't specify the `--kubeconfig` flag while executing kubectl commands

**Summary**:

Most Devs will have an environment with the exported KUBECONFIG env but this is not the case in the CI. And since we need to test the example-solution in the CI, we require to explicitly set the `--kubeconfig` flag.

**Acceptance criteria**: 

- Working Solutions.sh script

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/scripts/solutions.sh
+++ b/scripts/solutions.sh
@@ -20,6 +20,8 @@ SOLUTION=''
 VERBOSE=${VERBOSE:-0}
 VERSION=''
 
+export KUBECONFIG
+
 declare -A COMMANDS=(
     [import]=import_solution
     [unimport]=unimport_solution


NicolasT (changes requested):
Can't you just `export KUBECONFIG` somewhere in the beginning of this script and be done with it? :thinking:

scripts/solutions.sh
```diff
@@ -20,6 +20,8 @@ SOLUTION=''
 VERBOSE=${VERBOSE:-0}
 VERSION=''
 
+export KUBECONFIG="$KUBECONFIG"
```
NicolasT: `export KUBECONFIG` should be enough (and more portable).

NicolasT (changes requested):

scripts/solutions.sh
Ebaneck: Ok, I see

Done!

alexandre-allard (approved):

NicolasT (approved):

Issue #1850, opened 2019-10-07
Add E2E testing of Solutions CLI integration


gdemonet:
**Component**: tests, solutions

**Why this is needed**: Prepare future integration to CI (in post-merge test suite), validate first iteration

**What should be done**: Implement a simple test of build and deployment for the `examples/metalk8s-solution-example` embedded in the repo

**Implementation proposal**:

- [ ] In some Eve step (not triggered by pre-merge)
  - [x] Build Solution ISO (`cd examples/metalk8s-solution-example && make iso`)
  - [x] Add archive with `solutions-manager.sh`
  - [x] Deploy Solution Operator in some Namespace (no `Environment` involved there)
  - [x] Add CRs and check the exposed HTTP endpoints

alexandre-allard:
Done in #2321

PR #2321, opened 2020-03-23
Add the integration of example-solution tests in ci

Ebaneck:
**Component**:

'ci'

**Context**: 

See #1856 and #1850 

**Summary**:

Be able to:
 - Spawn two nodes (Bootstrap node with MetalK8s installed and a second node1)
 - Prepare node1 with Cypress deps
 - Add example-solution iso file built from this PR (https://github.com/scality/metalk8s/pull/2312)
 - Deploy Solution Operator in some Namespace (example-solution)
 - Apply example-solution CRs

**Acceptance criteria**: 

Debts on : https://github.com/scality/metalk8s/pull/2315 since the Pod containers cannot startup because of missing binaries 
and https://github.com/scality/metalk8s/pull/2320 since we require a Kubeconfig to run any kubectl command.

---

Closes: #1850

### DIFF
--- a/tests/post/features/solutions.feature
+++ b/tests/post/features/solutions.feature
@@ -1,0 +1,27 @@
+@local @solution
+Feature: Solutions
+    Scenario: Deploy Solution
+        Given the Kubernetes API is available
+        And no Solution 'example-solution' is imported
+        And no Solution environment 'example-environment' is available
+        And the Solution Configuration file is absent
+        When we import a Solution archive '/var/tmp/example-solution.iso'
+        Then Solution archive 'example-solution' is imported correctly
+        And Solution 'example-solution' version '0.1.0-dev' is available
+        When we activate Solution 'example-solution' version '0.1.0-dev'
+        Then Solution 'example-solution' version '0.1.0-dev' is activated
+        And CRD 'versionservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
+        And CRD 'clockservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
+        When we create a solution environment 'example-environment'
+        Then solution environment 'example-environment' is available
+        When we remove Taints on node 'bootstrap' before deployment
+        And we deploy Solution 'example-solution' in environment 'example-environment' with version '0.1.0-dev'
+        Then we have 1 running pod labeled 'name=example-solution-operator' in namespace 'example-environment'
+        And we have 1 running pod labeled 'app=example-solution-ui' in namespace 'example-environment'
+        When we deactivate Solution 'example-solution'
+        And we delete Solution 'example-solution' in environment 'example-environment'
+        And we delete Solution environment 'example-environment'
+        And we unimport Solution archive '/var/tmp/example-solution.iso'
+        And we delete the Solution Configuration file
+        Then we have no Solution 'example-solution' archive mounted
+        And we have no Solution Configuration file present
--- a/tests/post/steps/test_solutions.py
+++ b/tests/post/steps/test_solutions.py
@@ -1,0 +1,350 @@
+import json
+from ast import literal_eval
+
+import requests
+import requests.exceptions
+
+import pytest
+from pytest_bdd import scenario, given, then, when, parsers
+
+import kubernetes.client
+from kubernetes.client import AppsV1Api
+from kubernetes.client.rest import ApiException
+
+from tests import kube_utils
+from tests import utils
+
+# Constants {{{
+
+DEFAULT_SOLUTION_MOUNTPOINT = "/srv/scality"
+SOLUTION_PILLAR_KEY = "metalk8s:solutions:available"
+SOLUTION_CONFIGURATION_FILE = "/etc/metalk8s/solutions.yaml"
+INFRA_TAINT = "node-role.kubernetes.io/infra:NoSchedule-"
+BOOTSTRAP_TAINT = "node-role.kubernetes.io/bootstrap:NoSchedule-"
+SOLUTION_CONFIGMAP = "metalk8s-solutions"
+SOLUTION_NAMESPACE = "metalk8s-solutions"
+ENVIRONMENT_LABEL = "solutions.metalk8s.scality.com/environment"
+
+# }}}
+
+# Fixtures {{{
+
+
+@pytest.fixture(scope='function')
+def context():
+    return {}
+
+
+# }}}
+
+
+# Scenarios {{{
+
+@scenario('../features/solutions.feature', 'Deploy Solution')
+def test_deploy_solution(host):
+    pass
+
+# }}}
+
+
+# Given {{{
+
+@given(parsers.parse("no Solution '{name}' is imported"))
+def is_absent_solution(host, name, k8s_client):
+    with host.sudo():
+        assert name not in host.mount_point.get_mountpoints(), \
+            'Solution mountpoint {} exists'.format(name)
+
+
+@given(parsers.parse("no Solution environment '{environment}' is available"))
+def is_absent_solution_environment(host, environment, k8s_client):
+    solution_environment = get_environment(k8s_client, environment)
+    assert environment not in solution_environment, \
+        'Solution environment {} exists'.format(environment)
+
+
+@given(parsers.parse("the Solution Configuration file is absent"))
+def is_absent_solution_config(host):
+    with host.sudo():
+        if host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, \
+                'Solution Config file exists at path {}'.format(
+                    SOLUTION_CONFIGURATION_FILE
+                )
+
+# }}}
+
+
+# When {{{
+
+@when(parsers.parse("we import a Solution archive '{path}'"))
+def import_solution(request, host, path):
+    run_solutions_command(request, host, ["import", "--archive", path])
+
+
+@when(parsers.parse("we activate Solution '{name}' version '{version}'"))
+def activate_example_solution(request, host, name, version):
+    run_solutions_command(
+        request,
+        host,
+        ["activate", "--name", name, "--version", version]
+    )
+
+
+@when(parsers.parse(
+    "we create a solution environment '{environment}'"))
+def create_solution_environment(request, host, environment):
+    run_solutions_command(request, host, ["create-env", "--name", environment])
+
+
+@when(parsers.parse(
+    "we remove Taints on node '{node_name}' before deployment"))
+def remove_node_taints(request, host, node_name):
+    # Remove Taints on the Bootstrap node before deploying
+    # example-solution-operator else, the operator pods will not startup
+
+    with host.sudo():
+        res = host.run((
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'taint nodes {} {}'
+        ).format(node_name, BOOTSTRAP_TAINT))
+        assert res.rc == 0, res.stdout
+
+    with host.sudo():
+        res = host.run((
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'taint nodes {} {}'
+        ).format(node_name, INFRA_TAINT))
+        assert res.rc == 0, res.stdout
+
+
+@when(parsers.parse(
+    "we deploy Solution '{name}' in environment '{environment}' with version "
+    "'{solution_version}'"))
+def deploy_solution(request, host, name, environment, solution_version):
+    run_solutions_command(
+        request,
+        host,
+        [
+            "add-solution",
+            "--name",
+            environment,
+            "--solution",
+            name,
+            "--version",
+            solution_version
+        ]
+    )
+
+
+@when(parsers.parse("we deactivate Solution '{name}'"))
+def deactivate_solution(request, host, name):
+    run_solutions_command(request, host, ["deactivate", "--name", name])
+
+
+@when(parsers.parse(
+    "we delete Solution '{name}' in environment '{environment}'"))
+def delete_solution(host, request, name, environment):
+    run_solutions_command(
+        request,
+        host,
+        ["delete-solution", "--name", environment, "--solution", name]
+    )
+
+
+@when(parsers.parse("we delete Solution environment '{environment}'"))
+def delete_solution_env(host, request, environment):
+    run_solutions_command(request, host, ["delete-env", "--name", environment])
+
+
+@when(parsers.parse("we unimport Solution archive '{path}'"))
+def unimport_solution_archive(host, request, path):
+    run_solutions_command(request, host, ["unimport", "--archive", path])
+
+
+@when("we delete the Solution Configuration file")
+def remove_solution_config(host, request):
+    with host.sudo():
+        res = host.run(('rm -rf {}').format(SOLUTION_CONFIGURATION_FILE))
+        assert res.rc == 0, res.stdout
+
+# }}}
+
+
+# Then {{{
+
+@then(parsers.parse("Solution archive '{name}' is imported correctly"))
+def is_imported_solution_archive(host, k8s_client, name):
+    # Solution is imported correctly if
+    # 1. Solution Configuration is created at /etc/metalk8s/solution.yaml
+    # 2. Solution Configuration file has new archive path
+    # 3. Solutions ConfigMap has the new Solution imported
+    with host.sudo():
+        assert host.file(SOLUTION_CONFIGURATION_FILE).exists, \
+            'No Solution Configuration file exists at {}'.format(
+                SOLUTION_CONFIGURATION_FILE
+            )
+    # Todo: SOLUTION_CONFIGURATION_FILE is yaml, we could parse it
+    # and ensure the archive path is at the right location
+    with host.sudo():
+        assert name in host.file(SOLUTION_CONFIGURATION_FILE).content_string, \
+            'No {} Solution ISO path found in Solution Config {}'.format(
+                name, SOLUTION_CONFIGURATION_FILE
+            )
+
+    def _wait_for_imported_archive():
+
+        configmap = get_configmap(
+            k8s_client, SOLUTION_CONFIGMAP, SOLUTION_NAMESPACE
+        )
+        assert configmap is not None, \
+            'Failed to read Solution ConfigMap {}'.format(SOLUTION_CONFIGMAP)
+
+        data = configmap.get('data', {})
+        assert any(k == name for k, v in data.items()), \
+            '{} Configmap does not contain solution {}'.format(
+                SOLUTION_CONFIGMAP, name
+            )
+    utils.retry(
+        _wait_for_imported_archive,
+        times=10,
+        wait=5,
+        name="wait for imported archive"
+    )
+
+
+@then(parsers.parse(
+    "Solution '{name}' version '{solution_version}' is available"))
+def is_available_solution(host, request, name, solution_version):
+    available_solution = get_available_solution(host)
+    assert name in available_solution
+    for k, v in available_solution.items():
+        if k == name:
+            for item in v:
+                assert item.get('version', None) == solution_version, \
+                    'Expected solution version {}: got {}'.format(
+                        solution_version, item.get('version', None)
+                    )
+
+
+@then(parsers.parse(
+    "Solution '{name}' version '{solution_version}' is activated"))
+def is_activated_solution(host, name, solution_version, k8s_client):
+    # Solution is marked activated in Solutions ConfigMap
+
+    def _wait_for_activated_solution():
+        configmap = get_configmap(
+            k8s_client, SOLUTION_CONFIGMAP, SOLUTION_NAMESPACE
+        )
+        assert name in configmap.get('data', {}), \
+            '{} Configmap does not contain {}'.format(SOLUTION_CONFIGMAP, name)
+
+        # Solution is marked activated in the Pillars
+        available_solutions = get_available_solution(host)
+        for k, v in available_solutions.items():
+            if k == name:
+                assert any([item.get('active') and item.get('version')
+                           == solution_version for item in v]) is True, \
+                    'Solution {} in version {} is not active'.format(
+                        name, solution_version
+                    )
+
+    utils.retry(
+        _wait_for_activated_solution,
+        times=10,
+        wait=5,
+        name="wait for activated solution"
+    )
+
+
+@then(parsers.parse("CRD '{name}' exists in Kubernetes API"))
+def is_available_crds(request, host, name):
+    with host.sudo():
+        res = host.run((
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'get crd {}'
+        ).format(name))
+        assert res.rc == 0, res.stdout
+
+
+@then(parsers.parse("solution environment '{name}' is available"))
+def read_solution_environment(k8s_client, name):
+    assert get_environment(k8s_client, name) is not None, \
+        'Solution environment {} not available'.format(name)
+
+
+@then(parsers.parse("we have no Solution '{name}' archive mounted"))
+def no_solution_mountpoint(host, name, k8s_client):
+    with host.sudo():
+        assert name not in host.mount_point.get_mountpoints(), \
+            'Found Solution mountpoint {}'.format(name)
+
+
+@then(parsers.parse("we have no Solution environment '{name}'"))
+def no_solution_environment(k8s_client, name):
+    assert get_environment(k8s_client, name) is None, \
+        'No environment with name {} found'.format(name)
+
+
+@then(parsers.parse("we have no available Solution '{name}'"))
+def no_available_solution(host, name):
+    assert name not in get_available_solution(host), \
+        'Solution with name {} found'.format(name)
+
+
+@then("we have no Solution Configuration file present")
+def no_solution_config(host):
+    with host.sudo():
+        if host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, \
+                'Solution Configuration file exists at {}'.format(
+                    SOLUTION_CONFIGURATION_FILE
+                )
+
+
+# }}}
+
+
+# Helper {{{
+
+def get_configmap(k8s_client, name, namespace):
+    try:
+        response = k8s_client.read_namespaced_config_map(
+            name, namespace
+        ).to_dict()
+    except Exception as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        raise
+    return response
+
+
+def get_environment(k8s_client, name):
+    try:
+        response = k8s_client.list_namespace(
+            label_selector="{}={}".format(ENVIRONMENT_LABEL, name)
+        )
+    except (ApiException) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        raise
+    return response.items
+
+
+def get_available_solution(host):
+    with host.sudo():
+        output = host.check_output(
+            'salt-call --out=json pillar.get "{}"'.format(SOLUTION_PILLAR_KEY)
+        )
+    data = json.loads(output)['local']
+    return data
+
+
+def run_solutions_command(request, host, args):
+    iso_root = request.config.getoption("--iso-root")
+    cmd = [str(iso_root / "solutions.sh")] + args
+    with host.sudo():
+        res = host.run(' '.join(cmd))
+        assert res.rc == 0, res.stdout
+
+# }}}
--- a/tox.ini
+++ b/tox.ini
@@ -131,6 +131,7 @@ markers =
     ingress: tag a BDD feature as related to ingress
     volume: tag a BDD feature as related to Volume management
     bootstrap: tag a BDD feature as related to bootstrap
+    solution: tag a BDD feature as related to solution
 filterwarnings =
     ignore:encode_point has been deprecated on EllipticCurvePublicNumbers and will be removed in a future version. Please use EllipticCurvePublicKey.public_bytes to obtain both compressed and uncompressed point encoding.:UserWarning
     ignore:Support for unsafe construction of public numbers from encoded data will be removed in a future version. Please use EllipticCurvePublicKey.from_encoded_point:UserWarning


eve/main.yml
```diff
@@ -1466,3 +1469,277 @@ stages:
...
+            BOOTSTRAP_CR_PATH: "/tmp/examplesolution_v1alpha1_versionserver_cr.yaml"
+          command: >
+            scp -F $SSH_CONFIG "${HOST_CR_PATH}" bootstrap:"${BOOTSTRAP_CR_PATH}"
+          haltOnFailure: true
+      - ShellCommand:
+          name: Apply example-solution CRs on Bootstrap node
+          command: >
+            ssh -F ssh_config bootstrap
+            sudo kubectl --kubeconfig /etc/kubernetes/admin.conf apply -f
+            /tmp/examplesolution_v1alpha1_versionserver_cr.yaml
```
alexandre-allard: This CR must be put in the right namespace `example-environment`.

Ebaneck: https://github.com/scality/metalk8s/pull/2321/files#diff-fa1f5352a12d0b35290d51280da84271R5

alexandre-allard: Oh! Yes! :p

eve/main.yml
```diff
@@ -1466,3 +1469,277 @@ stages:
...
+              if [ $ATTEMPTS -eq 10 ]; then
+                echo "Could not obtain Replicas after ${ATTEMPTS} attempts" >&2
+                exit 1
+              fi
+              sleep 2
+            done'
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
+      - SetPropertyFromCommand:
+        name: Set Cluster IP as Property for example-versionserver
```
alexandre-allard: There is an issue with indentation here.

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+              if [ $ATTEMPTS -eq 10 ]; then
+                echo "Could not obtain Replicas after ${ATTEMPTS} attempts" >&2
+                exit 1
+              fi
+              sleep 3
+            done'
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
+      - ShellCommand:
+          name: Verify exampleversions Deployment
```
alexandre-allard: Looks very similar to the `ShellCommand` above, maybe we could factorize this ?

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+            REPLICAS=$(sudo kubectl "$KUBECONF" get deployments example-versionserver -n example-environment -o custom-columns=:status.availableReplicas)
+            until [ $REPLICAS -eq 3 ]; do
+              $REPLICAS
+              ((ATTEMPTS++))
+              if [ $ATTEMPTS -eq 10 ]; then
+                echo "Could not obtain Replicas after ${ATTEMPTS} attempts" >&2
+                exit 1
+              fi
+              sleep 3
+            done'
```
alexandre-allard: You're not executing the above command here, only executing the result of it.
Should be something like:
```suggestion
            bash -c '
            ATTEMPTS=0
            declare -a GET_REPLICAS=(
                sudo kubectl $KUBECONF get deployments example-versionserver
                -n example-environment -o custom-columns=:status.availableReplicas
            )
            REPLICAS=$("${GET_REPLICAS[@]}")
            until [ $REPLICAS -eq 3 ]; do
              ((ATTEMPTS++))
              if [ $ATTEMPTS -eq 10 ]; then
                echo "Could not obtain Replicas after ${ATTEMPTS} attempts" >&2
                exit 1
              fi
              sleep 3
              REPLICAS=$("${GET_REPLICAS[@]}")
            done'
```
Or even (more compact way):
```suggestion
            bash -c '
            declare -a GET_REPLICAS=(
                sudo kubectl $KUBECONF get deployments example-versionserver
                -n example-environment -o custom-columns=:status.availableReplicas
            )
            for ((attempts=0, replicas=$("${GET_REPLICAS[@]}"); replicas != 3; replicas=$("${GET_REPLICAS[@]}"), ++attempts)); do
              if [ $attempts -eq 10 ]; then
                echo "Could not obtain Replicas after $attempts attempts" >&2
                exit 1
              fi
              sleep 3
            done'
```
Also, note that you can't protect `$KUBECONF` variable with double quotes since you did put the flag and its value in the same variable, otherwise you end up with a single argument and it does not work.
I think you should only put the path `/etc/kubernetes/admin.conf` in this variable and rename it to `$KUBECONFIG` to be consistent.

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+          name: Set Cluster IP as Property for example-versionserver
+          property: versionserver_cluster_ip
+          env:
+            KUBECONF: "--kubeconfig /etc/kubernetes/admin.conf"
+            SSH_CONFIG: >-
+              eve/workers/openstack-multiple-nodes/terraform/ssh_config
+          command: |
+            ssh -F $SSH_CONFIG bootstrap --
+            bash -c '
+            IP=$(sudo kubectl "$KUBECONF" get service example-versionserver -n example-environment -o custom-columns=:spec.clusterIP) &&
```
alexandre-allard: Same as above, will not work.

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+            echo $IP'
+          haltOnFailure: true
+      - ShellCommand:
+          name: Verify versionserver http endpoints
+          env:
+            HTTP: "http://"
```
alexandre-allard: Should be name `SCHEME` and I'm not sure that we need an environment variable for this, normally it should never be changed.

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+            IP: "%(prop:versionserver_cluster_ip)s"
+            SUB_URL: ":8080/version"
```
alexandre-allard: Same here, not sure the variable is useful, and it should not be named this way, at least have a `PORT` variable and `URL_PATH`

eve/main.yml
```diff
@@ -1553,3 +1556,277 @@ stages:
...
+          haltOnFailure: true
+      # Taint bootstrap node to allow deployment of example-solution-operator
+      - ShellCommand:
+          name: Taint Bootstrap node before deploying example-solution-operator
+          env:
+            KUBECONF: "--kubeconfig /etc/kubernetes/admin.conf"
+          command: >
+            ssh -F ssh_config bootstrap
+            sudo kubectl ${KUBECONF} taint nodes --all node-role.kubernetes.io/infra:NoSchedule-
+            sudo kubectl ${KUBECONF} taint nodes --all node-role.kubernetes.io/bootstrap:NoSchedule-
```
alexandre-allard: It will not work this way, everything will be put in a single kubectl command.

Ebaneck: yep it does not

eve/main.yml
```diff
@@ -1553,3 +1556,300 @@ stages:
...
+      - ShellCommand:
+          name: Taint Bootstrap node before deploying example-solution-operator
+          env:
+            KUBECONFIG: "/etc/kubernetes/admin.conf"
+            INFRA_TAINT: "node-role.kubernetes.io/infra:NoSchedule-"
+            BOOTSTRAP_TAINT: "node-role.kubernetes.io/bootstrap:NoSchedule-"
+          command: >
+            ssh -F ssh_config bootstrap -- "
+            sudo kubectl --kubeconfig ${KUBECONFIG} taint nodes --all ${BOOTSTRAP_TAINT} &&
+            sudo kubectl --kubeconfig ${KUBECONFiG} taint nodes --all ${INFRA_TAINT}
```
alexandre-allard: Wouw, this one was hard to find, you have a typo:
```suggestion
            sudo kubectl --kubeconfig ${KUBECONFIG} taint nodes --all ${INFRA_TAINT}
```

eve/main.yml
```diff
@@ -1553,3 +1556,300 @@ stages:
...
+          command: |
+            ssh -F $SSH_CONFIG bootstrap --
+            bash -c '
+            declare -a GET_IP=(
+              sudo kubectl --kubeconfig "$KUBECONFIG" get service
+              example-versionserver -n example-environment
+              -o custom-columns=:spec.clusterIP
+            )
+            IP=$("${GET_IP[@]}")
+            echo $IP'
```
alexandre-allard: Why not just execute the command instead of this:
```suggestion
            ssh -F $SSH_CONFIG bootstrap --
            sudo kubectl --kubeconfig "$KUBECONFIG" get service
            example-versionserver -n example-environment
            -o custom-columns=:spec.clusterIP
```

Ebaneck: Done

eve/main.yml
```diff
@@ -1553,3 +1556,295 @@ stages:
...
+              ((ATTEMPTS++))
+              if [ $ATTEMPTS -eq 10 ]; then
+                echo "Failed to obtain Replicas after ${ATTEMPTS} attempts" >&2
+                exit 1
+              fi
+              sleep 3
+              REPLICAS=$("${GET_REPLICAS[@]}")
+            done'
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
```
alexandre-allard: Not tested, but something like that should be better (less code duplication):
```suggestion
      - ShellCommand: &check_k8s_resource_column_ssh
          name: Verify example-solution-operator Deployment
          env: &_env_check_k8s_resource_column_ssh
            KUBECONFIG: "/etc/kubernetes/admin.conf"
            NAMESPACE: "example-environment"
            RESOURCE_TYPE: "deployment"
            RESOURCE_NAME: "example-solution-operator"
            RESOURCE_COLUMN: ":status.availableReplicas"
            RESOURCE_VALUE: "1"
            RETRIES: "10"
            SLEEP: "3"
          command: |
            ssh -F ssh_config bootstrap --
            bash -c '
            ATTEMPTS=0
            declare -a GET_COLUMN_VALUE=(
                sudo kubectl --kubeconfig "$KUBECONFIG" get "$RESOURCE_TYPE"
                "$RESOURCE_NAME" -n "$NAMESPACE"
                -o custom-columns="$RESOURCE_COLUMN"
            )
            CURRENT_VALUE=$("${GET_COLUMN_VALUE[@]}")
            until [ "$CURRENT_VALUE" -eq "$RESOURCE_VALUE" ]; do
              (( ATTEMPTS++ ))
              if [ "$ATTEMPTS" -eq "$RETRIES" ]; then
                echo "$RESOURCE_COLUMN still not equal to $RESOURCE_VALUE after $ATTEMPTS attempts" >&2
                exit 1
              fi
              sleep "$SLEEP"
              CURRENT_VALUE=$("${GET_COLUMN_VALUE[@]}")
            done'
          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
          haltOnFailure: true
      - ShellCommand:
          <<: *check_k8s_resource_column_ssh
          name: Verify example-versionserver Deployment
          env:
            <<: *_env_check_k8s_resource_column_ssh
            RESOURCE_NAME: "example-versionserver"
            RESOURCE_VALUE: "3"
```

alexandre-allard: Anyway, I'm not even sure the first approach or this one can work, using simple quotes will prevent environment variables from being expanded in the string passed to `bash -c`... and if you use double quotes, some variables will be expanded before the script got executed.
Maybe it would be better to put this in a dedicated `.sh`.

Ebaneck: Hmm, i would prefer doing it without a dedicated shell script

alexandre-allard: You can, you just need to use double quotes around the whole bash script and escape things that must not be expanded the first time.

eve/main.yml
```diff
@@ -1553,3 +1556,269 @@ stages:
...
+            ssh -F ssh_config bootstrap --
+            bash -c '
+            ATTEMPTS=0
+            declare -a GET_STATUS_CODE=(
+                curl --write-out %{http_code} --silent --output /dev/null
+                "http://$IP:$PORT$URL_PATH"
+            )
+            status_code=$("${GET_STATUS_CODE[@]}")
+            until [ "$status_code" -eq 200 ]; do
+              $status_code
```
alexandre-allard: The script will fail here for sure:
```
200: command not found
```

eve/main.yml
```diff
@@ -1553,3 +1556,269 @@ stages:
...
+              ((ATTEMPTS++))
+              if [ "$ATTEMPTS" -eq 10 ]; then
+                echo "Failed to obtain endpoint after ${ATTEMPTS} attempts" >&2
```
alexandre-allard:
```suggestion
                echo "Failed to reach endpoint after ${ATTEMPTS} attempts" >&2
```

examples/metalk8s-solution-example/get_resource.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
alexandre-allard: Not sure this is the best place to put this script as it's more related to CI than examples.
Should be put under `eve/`

eve/main.yml
```diff
@@ -1553,3 +1556,269 @@ stages:
...
+          haltOnFailure: true
+      - ShellCommand:
+          name: Copy example-solution CRs & Bash script to Bootstrap node
+          env:
+            SSH_CONFIG: >-
+              eve/workers/openstack-multiple-nodes/terraform/ssh_config
+            HOST_CR_PATH: "examples/metalk8s-solution-example/example-operator/deploy/crds/examplesolution_v1alpha1_versionserver_cr.yaml"
+            GET_RESOURCE: "examples/metalk8s-solution-example/get_resource.sh"
+          command: >
+            scp -F $SSH_CONFIG "${HOST_CR_PATH}" bootstrap:/tmp &&
```
alexandre-allard: You can copy multiple files using only 1 scp command (like cp):
```
scp -F "$SSH_CONFIG" "$HOST_CR_PATH" "$GET_RESOURCE" bootstrap:/tmp
```

eve/main.yml
```diff
@@ -1553,3 +1556,269 @@ stages:
...
+            NAMESPACE: "example-environment"
+            RESOURCE_TYPE: "deployments"
+            RESOURCE_NAME: "example-solution-operator"
+            RESOURCE_COLUMN: ":status.availableReplicas"
+            RESOURCE_VALUE: "1"
+            RETRIES: "10"
+            SLEEP: "3"
+          command: |
+            ssh -F ssh_config bootstrap
+            sudo bash /tmp/get_resource.sh
```
alexandre-allard: Here you're executing 2 commands, it must either be:
```suggestion
          command: |
            ssh -F ssh_config bootstrap \
            sudo bash /tmp/get_resource.sh
```
Or:
```suggestion
          command: >
            ssh -F ssh_config bootstrap
            sudo bash /tmp/get_resource.sh
```

eve/main.yml
```diff
@@ -1553,3 +1556,269 @@ stages:
...
+            "${SOLUTION_VERSION}" --verbose
+          workdir: build/eve/workers/openstack-multiple-nodes/terraform/
+          haltOnFailure: true
+      - ShellCommand:
+          name: Copy example-solution CRs & Bash script to Bootstrap node
+          env:
+            SSH_CONFIG: >-
+              eve/workers/openstack-multiple-nodes/terraform/ssh_config
+            HOST_CR_PATH: "examples/metalk8s-solution-example/example-operator/deploy/crds/examplesolution_v1alpha1_versionserver_cr.yaml"
+            GET_RESOURCE: "examples/metalk8s-solution-example/get_resource.sh"
```
alexandre-allard: I don't see the point of using environment variables for these.

eve/main.yml
```diff
@@ -1553,3 +1556,275 @@ stages:
...
+            sudo env
+            KUBECONFIG=/etc/kubernetes/admin.conf
+            NAMESPACE=example-environment
+            RESOURCE_TYPE=deployments
+            RESOURCE_NAME=example-solution-operator
+            RESOURCE_COLUMN=:status.availableReplicas
+            RESOURCE_VALUE=1
+            RETRIES=10
+            SLEEP=3
+            /tmp/get_resource.sh
```
alexandre-allard: Will not work, this file is not executable.

gdemonet (dismissed):
Why are we doing this as Eve steps using bash scripts, instead of extending the Python test suite?

eve/main.yml
```diff
@@ -458,6 +459,7 @@ stages:
           stage_names:
             - single-node-install-rhel
             - multiple-nodes-centos
+            - multiple-nodes-example-solution-centos #remove me
```
TeddyAndrieux: Do we really need a new environment to test solution ?? ( For me no) Couldn't be part of the multi node we already have ?

alexandre-allard: No, because we will do it in postmerge and we only want to test solution deployment, not necessarily the whole expansion.

eve/get_resource.sh
```diff
@@ -0,0 +1,17 @@
...
+    kubectl --kubeconfig "${KUBECONFIG}" get "${RESOURCE_TYPE}"
+    "${RESOURCE_NAME}" -n "${NAMESPACE}" -o "custom-columns=${RESOURCE_COLUMN}"
+)
+CURRENT_VALUE=$("${GET_COLUMN_VALUE[@]}")
+until [ "${CURRENT_VALUE}" -eq "${RESOURCE_VALUE}" ]; do
+    (( ATTEMPTS++ ))
+    if [ "${ATTEMPTS}" -eq "${RETRIES}" ]; then
+    echo "${RESOURCE_COLUMN} still not equal to ${RESOURCE_VALUE} after ${ATTEMPTS} attempts" >&2
+    exit 1
+    fi
```
TeddyAndrieux:
```suggestion
    if [ "${ATTEMPTS}" -eq "${RETRIES}" ]; then
        echo "${RESOURCE_COLUMN} still not equal to ${RESOURCE_VALUE} after ${ATTEMPTS} attempts" >&2
        exit 1
    fi
```

eve/main.yml
TeddyAndrieux: Hmm yes true, then you also need to add like for all other step the right env variable (and add these env variable on `add_final_stauts` like for all other steps)
```
_metalk8s_internal_info:
      junit_info: &_install_multi-node_solution_junit_info
....
```
To generate the right junit result file.

gdemonet: I don't think it should be a "multi nodes" cluster. IMO, we want:

- one Bootstrap, to install Metal, and then import + run the Solution 
- one extra VM to run Cypress tests

This should also be the case for "single node" tests, so yeah, maybe merge them.

TeddyAndrieux: Then you need to remove taint on the bootstrap node, but it's doable yes (or add toleration for example solution)

Ebaneck: Removing taints on the Bootstrap node was handled here: https://github.com/scality/metalk8s/pull/2321/files#diff-73e74177595a435f88a0701e605cf08eR1686-R1698

TeddyAndrieux: Hmmm then why you need multi node ?
Anyway I agree with @gdemonet comment https://github.com/scality/metalk8s/pull/2321#pullrequestreview-383610719 this test should be directly in pytest

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,22 @@
...
+        And we have 3 running pod labeled 'app.kubernetes.io/name=example-versionserver' in namespace 'example-environment'
+        When we check the available replicas for 'example-versionserver' deployment in namespace 'example-environment'
+        Then we have '3' at 'status.available_replicas' for 'example-versionserver' Deployment in namespace 'example-environment'
+
+    Scenario: Expected HTTP endpoints example versionserver
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=example-versionserver' are 'Ready'
+        And we have 3 running pod labeled 'app.kubernetes.io/name=example-versionserver' in namespace 'example-environment'
+        When we check the http endpoint '/version' on port '8080' with IP 'spec.cluster_ip' for 'example-versionserver' in namespace 'example-environment'
+        Then service returns '200' status code
```
gdemonet: Those are not scenarii, merely static checks that could be in a `Given` or a `Then`.

You want to test the effect of running the CLI `solutions.sh`, build one scenario per command:
- Given no Solution is imported, And the config file is absent, When I run the `solutions.sh import ...` command, Then ...
- Given Solution toto is imported, And it's not marked as active, When I run `solutions.sh activate ...`, Then ...
- etc.

Do the same for creation of environment, adding a solution instance to an environment, and optionally (but really, that's not the point of those tests) check that the Operator works by creating a CR and checking that it deploys a component as expected.

Later iterations will also validate the upgrade of a Solution, but for now, let's focus on those simple actions. Also make sure to clean-up the Solution deployment after the test run (use a `session`-scoped fixture, add the teardown commands to it when you managed to setup something, like importing a solution, and make sure it runs on test completion, be it successful or not).

Ebaneck: ok

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+
+@then(parsers.parse("we have no activated Solution '{name}'"))
+def no_activated_solution(host, request, name):
+    with host.sudo():
+        output = host.check_output(
+            'salt-call --out=json pillar.get "{}"'.format(SOLUTION_PILLAR_KEY)
+        )
+    data = json.loads(output)['local']
+
+    assert name not in data
```
alexandre-allard: This function is a mostly a duplicate of `is_activated_solution()`, you should create an helper and use it in both functions.
And btw, you're not testing if the solution is activated, but if it's available.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    with host.sudo():
+        res = host.run(' '.join([
+            cmd,
+            'create-env --name {}'.format(environment),
+        ]))
+        assert res.rc == 0, res.stdout
+
+
+@when("we remove Taints on Bootstrap node before deployment")
+def remove_bootstrapNode_taints(host):
```
alexandre-allard:
```suggestion
def remove_bootstrap_node_taints(host):
```

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    "I create a solution environment '{environment}' for Solution '{name}'"))
+def create_solution_environment(request, host, environment, name):
+    iso_root = request.config.getoption("--iso-root")
+    cmd = str(iso_root / "solutions.sh")
+    with host.sudo():
+        res = host.run(' '.join([
+            cmd,
+            'create-env --name {}'.format(environment),
+        ]))
+        assert res.rc == 0, res.stdout
```
alexandre-allard: Same here, code duplication, you could have an helper for this, with something like:
```
def run_solutions_command(request, host, args):
    iso_root = request.config.getoption("--iso-root")
    cmd = [str(iso_root / "solutions.sh")] + args
    with host.sudo():
        res = host.run(' '.join(cmd))
        assert res.rc == 0, res.stdout


@when(parsers.parse("I create a solution environment '{environment}'"))
def create_solution_environment(request, host, environment):
    run_solutions_command(request, host, ["create-env", "--name", environment])
```
And the `{name}` parameter is never used.

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,31 @@
...
+        And we delete Solution 'example-solution' in environment 'example-environment'
+        And we delete Solution environment 'example-environment'
+        And we unimport Solution archive '/var/tmp/example-solution.iso'
+        And we delete the Solution Configuration file
+        Then we have no Solution 'example-solution' archive mounted
+        And we have no Solution environment 'example-environment'
+        And we have no activated Solution 'example-solution'
+        And we have no Solution Configuration file present
+        And we have no Solution deployment 'example-solution-operator' in environment 'example-environment'
+        And we have no Solution deployment 'example-solution-ui' in environment 'example-environment'
```
alexandre-allard: These 2 will always be true since you're testing `And we have no Solution environment 'example-environment'` just before.

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,31 @@
...
+    Scenario: Deploy Solution
+        Given the Kubernetes API is available
+        And no Solution 'example-solution' is imported
+        And no Solution environment 'example-environment' is available
+        And the Solution Configuration file is absent
+        When we import a Solution archive '/var/tmp/example-solution.iso'
+        Then Solution archive 'example-solution' is imported correctly
+        When I activate Solution 'example-solution' with version '0.1.0-dev'
+        Then Solution 'example-solution' is activated
+        When I create a solution environment 'example-environment' for Solution 'example-solution'
```
alexandre-allard:
```suggestion
        When we activate Solution 'example-solution' version '0.1.0-dev'
        Then Solution 'example-solution' version '0.1.0-dev' is activated
        When we create a solution environment 'example-environment' for Solution 'example-solution'
```

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+            'taint nodes --all {}'.format(BOOTSTRAP_TAINT),
+        ]))
+        assert res.rc == 0, res.stdout
+
+    with host.sudo():
+        res = host.run(' '.join([
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf ',
+            'taint nodes --all {}'.format(INFRA_TAINT),
+        ]))
+        assert res.rc == 0, res.stdout
```
alexandre-allard: Why using 2 times `host.sudo()` and why are you using lists if you're not really splitting arguments ?
```
        res = host.run(' '.join([
            'kubectl' '--kubeconfig' '/etc/kubernetes/admin.conf',
            'taint' 'nodes' '--all', INFRA_TAINT,
        ]))
```
Or
```
        res = host.run((
            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
            'taint nodes --all {}'
        ).format(INFRA_TAINT))
```

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    data = json.loads(output)['local']
+
+    assert name not in data
+
+
+@then(parsers.parse("we have no Solution Configuration file present"))
+def no_solution_config(host, name):
+    with host.sudo():
+        if host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, 'Solution Configuration file exist at {}'.format(
```
alexandre-allard:
```suggestion
            assert False, 'Solution Configuration file exists at {}'.format(
```

tests/post/steps/test_example_solution.py
Ebaneck: Lets shoot for the split argument version

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    # Solution is imported correctly if
+    # 1. Solution Configuration is created at /etc/metalk8s/solution.yaml
+    # 2. Solution Configuration has new archive path
+    # 3. Solutions ConfigMap has new Solution imported
+
+    with host.sudo():
+        if not host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, 'No Solution Configuration file exist at {}'.format(
+                        SOLUTION_CONFIGURATION_FILE
+                    )
```
alexandre-allard: Why not directly asserting ? (and weird indentation)

```suggestion
    assert host.file(SOLUTION_CONFIGURATION_FILE).exists, \
        'No Solution Configuration file exists at {}'.format(
            SOLUTION_CONFIGURATION_FILE
        )
```

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    with host.sudo():
+        file_contents = host.file(SOLUTION_CONFIGURATION_FILE).content_string
+        if name not in file_contents:
```
alexandre-allard: This file is valid YAML, you could easily parse it to ensure the string is at the right place. :)

Ebaneck: Would add a Todo here for some other time!

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+            assert False, 'Found Solution mountpoint {}'.format(name)
+    return name
+
+
+@given(parsers.parse("no Solution environment '{environment}' is available"))
+def is_absent_solution_environment(host, environment, k8s_client):
+    solution_environment = get_environment(k8s_client, environment)
+    if solution_environment is not None:
+        pytest.fail("Found Solution environment {}".format(environment))
+    return environment
```
alexandre-allard: Could be an assertion, and what's the return used for ?

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+@given(parsers.parse("the Solution Configuration file is absent"))
+def is_absent_solution_config(host):
+    with host.sudo():
+        if host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, 'Solution Config file exist at path {}'.format(
+                        SOLUTION_CONFIGURATION_FILE
+                    )
```
alexandre-allard: Really weird indentation here.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+        if not host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, 'No Solution Configuration file exist at {}'.format(
+                        SOLUTION_CONFIGURATION_FILE
+                    )
+    with host.sudo():
+        file_contents = host.file(SOLUTION_CONFIGURATION_FILE).content_string
+        if name not in file_contents:
+            assert False, 'No Solution ISO found in Solution Config {}'.format(
+                        SOLUTION_CONFIGURATION_FILE, name
+                    )
```
alexandre-allard: indentation

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,398 @@
...
+    assert name not in data
+
+
+@then(parsers.parse("we have no Solution Configuration file present"))
+def no_solution_config(host, name):
+    with host.sudo():
+        if host.file(SOLUTION_CONFIGURATION_FILE).exists:
+            assert False, 'Solution Configuration file exist at {}'.format(
+                        SOLUTION_CONFIGURATION_FILE
+                    )
```
alexandre-allard: identation

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,31 @@
...
+        Then solution environment 'example-environment' is available
+        When we remove Taints on Bootstrap node before deployment
+        And we deploy Solution 'example-solution' in environment 'example-environment' with version '0.1.0-dev'
+        Then the 'pods' list should not be empty in the 'example-environment' namespace
+        When we check the available replicas 'status.available_replicas' for 'example-solution-operator' deployment in environment 'example-environment'
+        Then we have '1' available 'status.available_replicas' for Solution 'example-solution-operator' Deployment in environment 'example-environment'
+        When we check the available replicas 'status.available_replicas' for 'example-solution-ui' deployment in environment 'example-environment'
+        Then we have '1' available 'status.available_replicas' for Solution 'example-solution-ui' Deployment in environment 'example-environment'
```
alexandre-allard:
```suggestion
        Then the 'pods' list should not be empty in the 'example-environment' namespace
        And we have '1' available 'status.available_replicas' for Solution 'example-solution-operator' Deployment in environment 'example-environment'
        And we have '1' available 'status.available_replicas' for Solution 'example-solution-ui' Deployment in environment 'example-environment'
```

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def get_environment(k8s_client, name):
+    try:
+        response = k8s_client.read_namespace(name)
+    except (ApiException) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        raise
+    return response
```
alexandre-allard: An environment should not be matched on the namespace name but on a specific label `metadata:labels:solutions.metalk8s.scality.com/environment` with the environment name as the value.
So you must use `list_namespace()` with a label selector, should be something like that:
```
ENVIRONMENT_LABEL = "solutions.metalk8s.scality.com/environment"
response = k8s_client.list_namespace(label_selector='{}={}'.format(ENVIRONMENT_LABEL, name))
```

Ebaneck: Ok, i will fix this!

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,27 @@
...
+Feature: Example solution
+    Scenario: Deploy Solution
+        Given the Kubernetes API is available
+        And no Solution 'example-solution' is imported
+        And no Solution environment 'example-environment' is available
+        And the Solution Configuration file is absent
+        When we import a Solution archive '/var/tmp/example-solution.iso'
+        Then Solution archive 'example-solution' is imported correctly
+        When we activate Solution 'example-solution' version '0.1.0-dev'
+        Then Solution 'example-solution' is available
```
alexandre-allard:
```suggestion
        Then Solution archive 'example-solution' is imported correctly
        And Solution 'example-solution' version '0.1.0-dev' is available
        When we activate Solution 'example-solution' version '0.1.0-dev'
        Then Solution 'example-solution' version '0.1.0-dev' is activated
        And CRD 'versionservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
        And CRD 'clockservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
```

alexandre-allard: The Solution is made available at import, not at activation.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+# }}}
+
+
+# Given {{{
+
+@given(parsers.parse("no Solution '{name}' is imported"))
+def is_absent_solution(host, name, k8s_client):
+    with host.sudo():
+        assert name not in host.mount_point.get_mountpoints(), \
+            'No Solution Configuration file exists at {}'.format(name)
```
alexandre-allard: The error message is not related to what's asserted.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+            'taint nodes --all {}'
+        ).format(BOOTSTRAP_TAINT))
+        assert res.rc == 0, res.stdout
+
+    with host.sudo():
+        res = host.run((
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'taint nodes --all {}'
+        ).format(INFRA_TAINT))
+        assert res.rc == 0, res.stdout
```
alexandre-allard: Should either be `remove_nodes_taints()` or better: you should only remove taints on bootstrap node here (not using the `--all flag`), otherwise, if someone runs this scenario on a multi nodes cluster, it'll have unintended behavior.

tests/post/features/example_solution.feature
```diff
@@ -0,0 +1,27 @@
+@local @solution
+Feature: Example solution
```
alexandre-allard: Feature is not "Example solution", but "Solutions CLI" or just "Solutions", file should be also renamed accordingly.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+    assert solution_environment_obj is None
+
+
+@then(parsers.parse("we have no available Solution '{name}'"))
+def no_available_solution(host, name):
+    assert name not in get_available_solution(host)
+
+
+@then(parsers.parse("we have no Solution Configuration file present"))
+def no_solution_config(host, name):
```
alexandre-allard: No need for parsers.parse() if there is nothing to parse.
And from where does the `name` parameter comes from ?

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+    return response
+
+
+def list_environments(k8s_client):
+    try:
+        response = k8s_client.list_namespace()
+    except Exception as exc:
+        pytest.fail("Unable to read namespaces: {!s}".format(exc))
+    return response.items
```
alexandre-allard: This function is not used anywhere.

tests/utils.py
```diff
@@ -66,3 +72,111 @@ def get_node_name(nodename, ssh_config=None):
...
+        total=retries,
+        read=retries,
+        connect=retries,
+        backoff_factor=backoff_factor,
+        status_forcelist=status_forcelist,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount("http://", adapter)
+    session.mount("https://", adapter)
+    return session
```
alexandre-allard: Not used

tests/conftest.py
```diff
@@ -150,6 +150,14 @@ def ssh_config(request):
...
 
 
+@pytest.fixture(scope="function")
+def request_retry_session(request):
+    # Callers can inject arguments using `pytest.mark.parametrize`
+    params = getattr(request, 'param', {})
+
+    return utils.requests_retry_session(**params)
+
+
```
alexandre-allard: not used

tests/utils.py
```diff
@@ -66,3 +72,111 @@ def get_node_name(nodename, ssh_config=None):
...
+        container='salt-master',
+        namespace='kube-system'
+    )
+
+    assert output.exit_status == 0, \
+        'command {} failed with: \nout: {}\nerr:'.format(
+            command,
+            output.stdout,
+            output.stderr
+        )
```
alexandre-allard: These two functions are not used either.

tests/utils.py
Ebaneck: Yes, will remove this one!
It was a cherry-pick from an old commit

tests/utils.py
```diff
@@ -66,3 +72,111 @@ def get_node_name(nodename, ssh_config=None):
...
+
+
+def set_dict_element(data, path, value, delimiter='.'):
+    """
+    Traverse a nested dict using a delimiter on a target string
+    replaces the value of a key within a dictionary and returns the new dict
+    """
+    path, _, key = path.rpartition(delimiter)
+    (get_dict_element(data, path) if path else data)[key] = value
+    return data
```
alexandre-allard: Not used

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def get_environment(k8s_client, name):
+    try:
+        response = k8s_client.read_namespace(name)
+    except (ApiException) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
```
alexandre-allard: Not sure the `isinstance(exc, ApiException)` is needed.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+def get_configmap(k8s_client, name, namespace):
+    try:
+        response = k8s_client.read_namespaced_config_map(
+            name, namespace
+        ).to_dict()
+    except Exception as exc:
+        pytest.fail(
+            "Unable to read {} ConfigMap  in namespace {} with error: {!s}"
+            .format(name, namespace, exc)
+        )
```
alexandre-allard: Why not doing the same as for `get_environment()`, returning an empty dict for a `404` status code and letting the caller decide what to do when the `ConfigMap` cannot be found.

tests/post/steps/test_example_solution.py
```diff
@@ -0,0 +1,331 @@
...
+def delete_solution_env(host, request, environment):
+    run_solutions_command(request, host, ["delete-env", "--name", environment])
+
+
+@when(parsers.parse("we unimport Solution archive '{path}'"))
+def unimport_solution_archive(host, request, path):
+    run_solutions_command(request, host, ["unimport", "--archive", path])
+
+
+@when(parsers.parse("we delete the Solution Configuration file"))
```
alexandre-allard: no need for `parsers.parse()`

tests/utils.py
Ebaneck: It is a cherry-pick from a commit that was only merged in 2.5
Atleast we use the `get_dict_element` method, so!

tests/post/steps/test_solutions.py
```diff
@@ -0,0 +1,383 @@
...
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'get crd {}'
+        ).format(name))
+        assert res.rc == 0, res.stdout
+
+
+@then(parsers.parse("solution environment '{name}' is available"))
+def read_solution_environment(k8s_client, name):
+    solution_environment_obj = get_environment(k8s_client, name)
+    assert any(name == env.metadata.name for env in solution_environment_obj)
```
alexandre-allard: Environment name != namespace name, anyway it should be:
```suggestion
    assert get_environment(k8s_client, name) is not None
```

tests/post/steps/test_solutions.py
```diff
@@ -0,0 +1,383 @@
...
+def no_solution_mountpoint(host, name, k8s_client):
+    with host.sudo():
+        assert name not in host.mount_point.get_mountpoints(), \
+            'Found Solution mountpoint {}'.format(name)
+
+
+@then(parsers.parse("we have no Solution environment '{name}'"))
+def no_solution_environment(k8s_client, name):
+    solution_environment_obj = get_environment(k8s_client, name)
+    assert name not in solution_environment_obj
```
alexandre-allard:
```suggestion
    assert get_environment(k8s_client, name) is None
```
And you should write explicit assertion error message.

tests/post/steps/test_solutions.py
```diff
@@ -0,0 +1,383 @@
...
+        assert name in configmap.get('data', {}), \
+            '{} Configmap does not contain {}'.format(SOLUTION_CONFIGMAP, name)
+
+        # Solution is marked activated in the Pillars
+        available_solutions = get_available_solution(host)
+        for k, v in available_solutions.items():
+            if k == name:
+                for items in v:
+                    assert bool(items.get('active', None))
+                    assert items.get('version', None) == solution_version
```
alexandre-allard: You should not do it that way, here it'll be okay because we ensure there is only one Solution deployed at a time, but you should not rely on it:
```suggestion
                assert any([item.get('active') and item.get('version') == solution_version for item in v]) is True, \
                    "Solution {} in version {} is not active".format(name, solution_version)
```

tests/post/steps/test_solutions.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+@then(parsers.parse("CRD '{name}' exists in Kubernetes API"))
+def is_available_crds(request, host, name):
+    with host.sudo():
+        res = host.run((
+            'kubectl --kubeconfig=/etc/kubernetes/admin.conf '
+            'get crd {}'
+        ).format(name))
+        assert res.rc == 0, res.stdout
```
alexandre-allard-scality, line 265: Please write assertion error messages.

Ebaneck: `Stdout` is sufficient no??

tests/post/features/solutions.feature
```diff
@@ -0,0 +1,30 @@
...
+        When we create a solution environment 'example-environment'
+        Then solution environment 'example-environment' is available
+        When we remove Taints on node 'bootstrap' before deployment
+        And we deploy Solution 'example-solution' in environment 'example-environment' with version '0.1.0-dev'
+        Then the 'pods' list should not be empty in the 'example-environment' namespace
+        And we have '1' available 'status.available_replicas' for Solution 'example-solution-operator' Deployment in environment 'example-environment'
+        And we have '1' available 'status.available_replicas' for Solution 'example-solution-ui' Deployment in environment 'example-environment'
+        When we deactivate Solution 'example-solution'
+        And we delete Solution 'example-solution' in environment 'example-environment'
+        And we delete Solution environment 'example-environment'
```
alexandre-allard, line 23: If we delete a Solution in an Environment and then delete the Environment right after, without checking if the Solution was successfully removed, we can't ensure the first command has worked as expected.

tests/post/features/solutions.feature
```diff
@@ -0,0 +1,30 @@
...
+        Then Solution 'example-solution' version '0.1.0-dev' is activated
+        And CRD 'versionservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
+        And CRD 'clockservers.example-solution.metalk8s.scality.com' exists in Kubernetes API
+        When we create a solution environment 'example-environment'
+        Then solution environment 'example-environment' is available
+        When we remove Taints on node 'bootstrap' before deployment
+        And we deploy Solution 'example-solution' in environment 'example-environment' with version '0.1.0-dev'
+        Then the 'pods' list should not be empty in the 'example-environment' namespace
+        And we have '1' available 'status.available_replicas' for Solution 'example-solution-operator' Deployment in environment 'example-environment'
+        And we have '1' available 'status.available_replicas' for Solution 'example-solution-ui' Deployment in environment 'example-environment'
```
alexandre-allard-scality, line 21: You could have used what's already done for the rest of the tests (defined in conftest.py).
```
we have 1 running pod labeled 'name=example-solution-operator' in namespace 'example-environment'
we have 1 running pod labeled 'app=example-solution-ui' in namespace 'example-environment'
```

alexandre-allard: You could have used what's already done for the rest of the tests (defined in conftest.py).

```suggestion
        Then we have 1 running pod labeled 'name=example-solution-operator' in namespace 'example-environment'
        And we have 1 running pod labeled 'app=example-solution-ui' in namespace 'example-environment'
```

Ebaneck: Done

tests/post/steps/test_solutions.py
```diff
@@ -0,0 +1,350 @@
...
+def is_available_solution(host, request, name, solution_version):
+    available_solution = get_available_solution(host)
+    assert name in available_solution
+    for k, v in available_solution.items():
+        if k == name:
+            for item in v:
+                assert item.get('version', None) == solution_version, \
+                    'Expected solution version {}: got {}'.format(
+                        solution_version, item.get('version', None)
+                    )
```
alexandre-allard, line 227: Not the good way of doing this, you should check each item to see if there is a matching version.

alexandre-allard (approved):

PR #2325, opened 2020-03-23
[Centralized CLI] Requirements

### LINKED ISSUE #2322: Add a requirements document for centralised CLI
**Component**:

'documentation', 'requirements'

**Why this is needed**:

To make sure we are aligned on requirements for centralized CLI (#2163)

**What should be done**:

Write the doc and push a PR for review

**Test plan**:

thomasdanan:
Centralized CLI requirements (Draft)

**Component**:

'doc', 'requirements'

**Context**: 

Centralized CLI requirements

Closes: #2322

### DIFF
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -14,9 +14,11 @@ categorized as follow:
 - Cluster Service Administration (Configure Dex, Prometheus, Alert Manager,
   ...)
 
-K8s provides the kubectl CLI, which is distributed and enables
-all kind of interaction with all Kubernetes resources, through apiserver, but
-its usage often requires to build verbose YAML files.
+K8s provides the kubectl CLI, enabling all kind of interactions with all
+Kubernetes resources, through apiserver, but its usage often requires to build
+verbose YAML files.
+It is shipped as an independent package and can be deployed and run from
+anywhere, on any OS.
 
 Currently, MetalK8s provides other set of scripts or manual procedures, but
 those are located in various locations, their usage may vary and they are not
@@ -49,12 +51,12 @@ CLIs:
 
 When it is possible, it would make sense to leverage kubectl plugin
 
-All functionalities are exposed through 2 distincts CLI:
+All functionalities are exposed through 2 distinct CLI:
 - kubectl (enriched using kubectl approach)
 - mk8sctl: a new CLI, exposing specific MetalK8s functionalities
 
 In order to operate the cluster with mk8sctl from outside of the cluster, a
-specific pkg (for each OS) or a procedure explaining how to deploy it is
+specific pkg (for each OS) or a procedure explaining how to deploy it, is
 available.
 The mk8scli is deployed and available by default on bootstrap nodes
 
@@ -131,13 +133,11 @@ Solution Administration
 +------------+------------+-----------------------------------------------+
 | Resource   | action     | parameters                                    |
 +============+============+===============================================+
-| solution   | import     |                                               |
-+------------+------------+-----------------------------------------------+
 | solution   | import     | path_to_iso                                   |
 +------------+------------+-----------------------------------------------+
 |environment | add        | name                                          |
 +------------+------------+-----------------------------------------------+
-|environment | delete     |                                               |
+|environment | delete     | name                                          |
 +------------+------------+-----------------------------------------------+
 
 Cluster Service Administration
@@ -156,7 +156,7 @@ Cluster Service Administration
 +------------+------------+-----------------------------------------------+
 | alert-rule | add        | conf-param=conf-value                         |
 +------------+------------+-----------------------------------------------+
-|prometheuse | edit-conf  | conf-param=conf-value                         |
+| prometheus | edit-conf  | conf-param=conf-value                         |
 +------------+------------+-----------------------------------------------+
 | grafana    | deploy     |                                               |
 +------------+------------+-----------------------------------------------+
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -14,9 +14,9 @@ categorized as follow:
 - Cluster Service Administration (Configure Dex, Prometheus, Alert Manager,
   ...)
 
-K8s provides the kubectl CLI, which is distributed and enables all kind of
-interaction with all Kubernetes resources, through apiserver, but its usage
-often requires to build verbose YAML files.
+K8s provides the kubectl CLI, which is distributed and enables
+all kind of interaction with all Kubernetes resources, through apiserver, but
+its usage often requires to build verbose YAML files.
 
 Currently, MetalK8s provides other set of scripts or manual procedures, but
 those are located in various locations, their usage may vary and they are not
@@ -29,12 +29,13 @@ The goal of the project is to provide MetalK8s administrator with an intuitive
 and easy to use set of tools in order to administrate and operate a finite set
 of functionalities.
 
-Because kubectl CLI is already in place and is well known by Kubernetes
+Because kubectl is already in place and is well known by Kubernetes
 administrators, it will be used as a standard to follow for all other MetalK8s
 CLIs:
 
 - CLI follows kubectl style: kubectl <action> <resource>
 - CLI provides an exhaustive help, per action, with relevant examples
+- CLI provides <action> help when the command is not valid
 - CLI is not interactive (except if password input is needed)
 - CLI should not require password input
 - CLI provides a dryrun mode for intrusive operations
@@ -48,13 +49,27 @@ CLIs:
 
 When it is possible, it would make sense to leverage kubectl plugin
 
+All functionalities are exposed through 2 distincts CLI:
+- kubectl (enriched using kubectl approach)
+- mk8sctl: a new CLI, exposing specific MetalK8s functionalities
+
+In order to operate the cluster with mk8sctl from outside of the cluster, a
+specific pkg (for each OS) or a procedure explaining how to deploy it is
+available.
+The mk8scli is deployed and available by default on bootstrap nodes
+
 Requirements
 ------------
+
+Not listing all commands that are already available through kubectl.
+Only describing commands that are missing or commands that can be simplified
+using new command line arguments and predefined manifest.
+
 
 Cluster Resources Administration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-tools: kubectl
+**tool: kubectl**
 
 +------------+------------------+-----------------------------------------+
 | Resource   | action           | parameters                              |
@@ -69,28 +84,28 @@ tools: kubectl
 +------------+------------------+-----------------------------------------+
 | node       | edit             | name, hostname (to change node hostname)|
 +------------+------------------+-----------------------------------------+
-| node       | deploy(or apply?)| <list of nodes>, dry-run                |
+| node       | deploy           | <list of nodes>, dry-run                |
 +------------+------------------+-----------------------------------------+
 | node       | delete, drain,   | <list of nodes>                         |
 |            | taint, label     |                                         |
 +------------+------------------+-----------------------------------------+
 | volume     | add              | name, nodeName, storageClassName,       |
 |            |                  | devicePath, manifest-template-path,     |
-|            |                  | labels                                  |
+|            |                  | labels, type                            |
 +------------+------------------+-----------------------------------------+
-| volume     | delete, label    | name                                    |
+| volume     | delete, labels   | name                                    |
 +------------+------------------+-----------------------------------------+
 
 
 Cluster Administration
 ^^^^^^^^^^^^^^^^^^^^^^
 
-tools: mk8sctl
+**tool: mk8sctl**
 
 +------------+------------+-----------------------------------------------+
 | Resource   | action     | parameters                                    |
 +============+============+===============================================+
-| bootstrap  | install    |                                               |
+| bootstrap  | deploy     |                                               |
 +------------+------------+-----------------------------------------------+
 | cluster    | import-iso | path_to_iso                                   |
 +------------+------------+-----------------------------------------------+
@@ -111,7 +126,7 @@ tools: mk8sctl
 Solution Administration
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-tools: mk8sctl
+**tool: mk8sctl**
 
 +------------+------------+-----------------------------------------------+
 | Resource   | action     | parameters                                    |
@@ -128,12 +143,26 @@ tools: mk8sctl
 Cluster Service Administration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-tools: mk8sctl
+**tool: mk8sctl**
 
 +------------+------------+-----------------------------------------------+
 | Resource   | action     | parameters                                    |
 +============+============+===============================================+
 | grafana    | edit-conf  | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+|alertmanager| edit-conf  | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+| alert-rule | edit-conf  | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+| alert-rule | add        | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+|prometheuse | edit-conf  | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+| grafana    | deploy     |                                               |
++------------+------------+-----------------------------------------------+
+|alertmanager| deploy     |                                               |
++------------+------------+-----------------------------------------------+
+| prometheus | deploy     |                                               |
 +------------+------------+-----------------------------------------------+
 | user       | add        | name, email, passwd, mk8s-roles               |
 +------------+------------+-----------------------------------------------+
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -7,10 +7,12 @@ Context
 MetalK8s comes with a set of services to operate and monitor the K8s cluster.
 All operations that need to be performed by the Platform Administrator could be
 categorized as follow:
+
 - Cluster Resources Administration (Nodes, Volumes, Deployments, ...)
 - Cluster Administration (Install, Upgrade, Downgrade, Backup, Restore, ...)
 - Solution Administration (CRUD Environment, Import/Remove Solution, ...)
-- Cluster Service Administration (Configure Dex, Prometheus, Alert Manager, ...)
+- Cluster Service Administration (Configure Dex, Prometheus, Alert Manager,
+  ...)
 
 K8s provides the kubectl CLI, which is distributed and enables all kind of
 interaction with all Kubernetes resources, through apiserver, but its usage
@@ -30,6 +32,7 @@ of functionalities.
 Because kubectl CLI is already in place and is well known by Kubernetes
 administrators, it will be used as a standard to follow for all other MetalK8s
 CLIs:
+
 - CLI follows kubectl style: kubectl <action> <object>
 - CLI provides an exhaustive help, per action, with relevant examples
 - CLI is not interactive (except maybe for very first install script)
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -35,7 +35,7 @@ CLIs:
 
 - CLI follows kubectl style: kubectl <action> <resource>
 - CLI provides an exhaustive help, per action, with relevant examples
-- CLI is not interactive (except for static user provisioning)
+- CLI is not interactive (except if password input is needed)
 - CLI should not require password input
 - CLI provides a dryrun mode for intrusive operations
 - CLI provides a verbose (or debug) mode
@@ -90,7 +90,7 @@ tools: mk8sctl
 +------------+------------+-----------------------------------------------+
 | Resource   | action     | parameters                                    |
 +============+============+===============================================+
-| cluster    | install    |                                               |
+| bootstrap  | install    |                                               |
 +------------+------------+-----------------------------------------------+
 | cluster    | import-iso | path_to_iso                                   |
 +------------+------------+-----------------------------------------------+
@@ -101,8 +101,6 @@ tools: mk8sctl
 | etcd       | health     |                                               |
 +------------+------------+-----------------------------------------------+
 | bootstrap  | backup     |                                               |
-+------------+------------+-----------------------------------------------+
-| bootstrap  | restore    | backup-file                                   |
 +------------+------------+-----------------------------------------------+
 | bootstrap  | restore    | backup-file                                   |
 +------------+------------+-----------------------------------------------+
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -33,10 +33,12 @@ Because kubectl CLI is already in place and is well known by Kubernetes
 administrators, it will be used as a standard to follow for all other MetalK8s
 CLIs:
 
-- CLI follows kubectl style: kubectl <action> <object>
+- CLI follows kubectl style: kubectl <action> <resource>
 - CLI provides an exhaustive help, per action, with relevant examples
-- CLI is not interactive (except maybe for very first install script)
+- CLI is not interactive (except for static user provisioning)
 - CLI should not require password input
+- CLI provides a dryrun mode for intrusive operations
+- CLI provides a verbose (or debug) mode
 - CLI implementation relies on secure APIs
 - CLI support <action> completion for easy discovery
 - CLI output is standardized and human readable by default
@@ -52,11 +54,92 @@ Requirements
 Cluster Resources Administration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+tools: kubectl
+
++------------+------------------+-----------------------------------------+
+| Resource   | action           | parameters                              |
++============+==================+=========================================+
+| node       | add              | name, ssh-user, hostname or ip,         |
+|            |                  | ssh port, ssh-key-path, sudo-required,  |
+|            |                  | roles, manifest-template-path           |
++------------+------------------+-----------------------------------------+
+| the manifest-template-path is the path to the manifest describing the   |
+| resource. Everything that is in the command line, override the manifest |
+| content. This apply to Node and Volumes.                                |
++------------+------------------+-----------------------------------------+
+| node       | edit             | name, hostname (to change node hostname)|
++------------+------------------+-----------------------------------------+
+| node       | deploy(or apply?)| <list of nodes>, dry-run                |
++------------+------------------+-----------------------------------------+
+| node       | delete, drain,   | <list of nodes>                         |
+|            | taint, label     |                                         |
++------------+------------------+-----------------------------------------+
+| volume     | add              | name, nodeName, storageClassName,       |
+|            |                  | devicePath, manifest-template-path,     |
+|            |                  | labels                                  |
++------------+------------------+-----------------------------------------+
+| volume     | delete, label    | name                                    |
++------------+------------------+-----------------------------------------+
+
+
 Cluster Administration
 ^^^^^^^^^^^^^^^^^^^^^^
+
+tools: mk8sctl
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| cluster    | install    |                                               |
++------------+------------+-----------------------------------------------+
+| cluster    | import-iso | path_to_iso                                   |
++------------+------------+-----------------------------------------------+
+| cluster    | upgrade    | dest-version, dry-run                         |
++------------+------------+-----------------------------------------------+
+| cluster    | downgrade  | dest-version, dry-run                         |
++------------+------------+-----------------------------------------------+
+| etcd       | health     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | backup     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | restore    | backup-file                                   |
++------------+------------+-----------------------------------------------+
+| bootstrap  | restore    | backup-file                                   |
++------------+------------+-----------------------------------------------+
+| cluster    | edit-conf  | conf-param=conf-value                         |
+|            |            | ex: ip-in-ip=true                             |
++------------+------------+-----------------------------------------------+
 
 Solution Administration
 ^^^^^^^^^^^^^^^^^^^^^^^
 
+tools: mk8sctl
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| solution   | import     |                                               |
++------------+------------+-----------------------------------------------+
+| solution   | import     | path_to_iso                                   |
++------------+------------+-----------------------------------------------+
+|environment | add        | name                                          |
++------------+------------+-----------------------------------------------+
+|environment | delete     |                                               |
++------------+------------+-----------------------------------------------+
+
 Cluster Service Administration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+tools: mk8sctl
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| grafana    | edit-conf  | conf-param=conf-value                         |
++------------+------------+-----------------------------------------------+
+| user       | add        | name, email, passwd, mk8s-roles               |
++------------+------------+-----------------------------------------------+
+| user       | delete     | name                                          |
++------------+------------+-----------------------------------------------+
+| user       | edit       | name, email, passwd, mk8s-roles               |
++------------+------------+-----------------------------------------------+
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -15,8 +15,9 @@ categorized as follow:
   ...)
 
 K8s provides the kubectl CLI, enabling all kind of interactions with all
-Kubernetes resources, through apiserver, but its usage often requires to build
-verbose YAML files.
+Kubernetes resources, through k8s apiserver, but its usage often requires to
+build verbose YAML files. Also it does not leverage everything MetalK8s exposes
+through the salt API.
 It is shipped as an independent package and can be deployed and run from
 anywhere, on any OS.
 
@@ -32,10 +33,9 @@ and easy to use set of tools in order to administrate and operate a finite set
 of functionalities.
 
 Because kubectl is already in place and is well known by Kubernetes
-administrators, it will be used as a standard to follow for all other MetalK8s
-CLIs:
+administrators, it will be used as a standard to follow, as much as possible,
+for all other MetalK8s CLIs:
 
-- CLI follows kubectl style: kubectl <action> <resource>
 - CLI provides an exhaustive help, per action, with relevant examples
 - CLI provides <action> help when the command is not valid
 - CLI is not interactive (except if password input is needed)
@@ -43,22 +43,32 @@ CLIs:
 - CLI provides a dryrun mode for intrusive operations
 - CLI provides a verbose (or debug) mode
 - CLI implementation relies on secure APIs
-- CLI support <action> completion for easy discovery
+- CLI support action completion for easy discovery
 - CLI output is standardized and human readable by default
 - CLI output can be formatted in JSON or YAML
-- CLI can be executed from outside of the cluster (except maybe for very first
-  install script)
 
 When it is possible, it would make sense to leverage kubectl plugin
 
-All functionalities are exposed through 2 distinct CLI:
-- kubectl (enriched using kubectl approach)
-- metalk8scli: a new CLI, exposing specific MetalK8s functionalities
+Most functionalities are exposed through 2 distinct CLI:
 
-In order to operate the cluster with metalk8scli from outside of the cluster, a
-specific pkg (for each OS) or a procedure explaining how to deploy it, is
-available.
-The metalk8scli and kubectl are deployed and available by default on bootstrap
+- kubectl: enriched with metalk8s plugin, to interact with both k8s apiserver
+  and salt API, and that can be executed from outside of the cluster.
+- metalk8sctl: a new CLI, exposing specific MetalK8s functionalities, that are
+  not interacting with k8s apiserver, and that must be executed on cluster node
+  host.
+
+Some cluster configurations will be achievable through documented procedures,
+such as changing one cluster server hostname.
+
+Other specific solution kubectl plugin may also be provided by a solution.
+
+To know which command must be used, administrator will rely on MetalK8s
+documentation. Documentation will be updated accordingly.
+
+In order to operate the cluster with kubectl plugins from outside of the
+cluster, plugin binary will be available for download from the bootstrap node
+or from MetalK8s release repository.
+The metalk8sctl and kubectl are deployed and available by default on bootstrap
 nodes.
 
 Requirements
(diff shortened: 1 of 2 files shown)


docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+MetalK8s comes with a set of services to operate and monitor the K8s cluster.
+All operations that need to be performed by the Platform Administrator could be
+categorized as follow:
+- Cluster Resources Administration (Nodes, Volumes, Deployments, ...)
+- Cluster Administration (Install, Upgrade, Downgrade, Backup, Restore, ...)
+- Solution Administration (CRUD Environment, Import/Remove Solution, ...)
+- Cluster Service Administration (Configure Dex, Prometheus, Alert Manager, ...)
+
+K8s provides the kubctl CLI, which is distributed and enables all kind of
+interaction with all resources available in etcd DB but its usage often requires
```
TeddyAndrieux: Not sure you want to talk about etcd here as kubectl basically just interact with apiserver he don't really care about etcd

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+to build verbose yaml files. Also kubectl tool only expose the standard K8s
+resources but not the custom ones, such as Volumes.
```
TeddyAndrieux: > kubectl tool only expose the standard K8s resources but not the custom ones, such as Volumes.

Hmm, he do. You can do `kubectl get volume`

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Currently, MetalK8s provides other set of CLI or manual procedures, but those
+are located in various locations, their usage may vary and they are not
+developed using the same technologies.
```
TeddyAndrieux: IMO we don't want to talk about technologies but logic here.
```suggestion
developed using the same logic.
```

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+
+The goal of the project is to provide MetalK8s administrator an intuitive and
+easy to use set of CLI in order to administrate and operate a finite set of
+functionalities.
+
+Because kubectl CLI is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+- CLI follows kubctl style
+- CLI is not interactive
```
TeddyAndrieux: except for some password that should be asked at the very beginning of the script

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+- CLI should not require password input
```
TeddyAndrieux, line 42: Except for user administration right ?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+- CLI implementation relies on secured APIs
+- CLI support completion for easy discovery
+- CLI output is standardized (yaml, json, ...)
+- CLI can be executed from outside of the cluster
```
TeddyAndrieux: Not all CLI

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Because kubectl CLI is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+- CLI follows kubctl style
+- CLI is not interactive
+- CLI should not require password input
+- CLI implementation relies on secured APIs
+- CLI support completion for easy discovery
+- CLI output is standardized (yaml, json, ...)
```
TeddyAndrieux: IMO we should follow kubectl style so standardized yes but not yaml or json by default
Something "pretty" and human readable by default with an option to change it to yaml, json, ...

TeddyAndrieux (dismissed):
Do we plan to have new CLI only in metalk8s-2.6 ?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+resources but not the custom ones, such as Volumes.
+
+Currently, MetalK8s provides other set of CLI or manual procedures, but those
+are located in various locations, their usage may vary and they are not
+developed using the same technologies.
+
+This makes the CLI and associated documentation not super intuitive and it
+also makes the maintenance more expensive in the long term.
+
+The goal of the project is to provide MetalK8s administrator an intuitive and
```
sayf-eddine-scality:
```suggestion
The goal of the project is to provide MetalK8s administrator with an intuitive and
```

docs/developer/architecture/centralized-cli.rst
sayf-eddine-scality: apiserver is just a giant wrapper in front of the etcd cluster anyway

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+functionalities.
+
+Because kubectl CLI is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+- CLI follows kubctl style
+- CLI is not interactive
+- CLI should not require password input
+- CLI implementation relies on secured APIs
+- CLI support completion for easy discovery
```
sayf-eddine-scality: on medium to large platforms it becomes super slow if this feature supports autocompleting pods/svc/volumes names. this needs to be counted for

TeddyAndrieux: Hmm for me we definitly not want completion for object name but just for different operation we can do with the CLI

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: For the moment yes but maybe one day they will use another backend :)

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+
+MetalK8s comes with a set of services to operate and monitor the K8s cluster.
+All operations that need to be performed by the Platform Administrator could be
+categorized as follow:
+- Cluster Resources Administration (Nodes, Volumes, Deployments, ...)
+- Cluster Administration (Install, Upgrade, Downgrade, Backup, Restore, ...)
+- Solution Administration (CRUD Environment, Import/Remove Solution, ...)
+- Cluster Service Administration (Configure Dex, Prometheus, Alert Manager, ...)
+
+K8s provides the kubctl CLI, which is distributed and enables all kind of
```
slaperche-scality: `s/kubctl/kubectl`, BTW `kubectl` is described as a `command line tool ` not `CLI`

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+interaction with all resources available in etcd DB but its usage often requires
+to build verbose yaml files. Also kubectl tool only expose the standard K8s
```
slaperche-scality: `s/yaml/YAML:`

docs/developer/architecture/centralized-cli.rst
slaperche-scality: yeah, you can manipulate custom resources with kubectl.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+resources but not the custom ones, such as Volumes.
+
+Currently, MetalK8s provides other set of CLI or manual procedures, but those
```
slaperche-scality: `a set of tools` not CLI.

I don't think CLI means what you think. A CLI is defined as `A user interface in which the user interacts with the computer through lines of text instead of graphical images.`

What we have is a bunch of script that we can run, and they are not really interactive.
You could say that this bunch of scripts defines a CLI for MetalK8s, but a script by itself isn't a CLI.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+also makes the maintenance more expensive in the long term.
+
+The goal of the project is to provide MetalK8s administrator an intuitive and
+easy to use set of CLI in order to administrate and operate a finite set of
+functionalities.
+
+Because kubectl CLI is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+- CLI follows kubctl style
```
slaperche-scality: `s/kubctl/kubectl/`

I may have missed some occurrences, but this typo is present several times (may be worth a search & replace).

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,57 @@
...
+- CLI is not interactive
+- CLI should not require password input
+- CLI implementation relies on secured APIs
```
slaperche-scality: Not sure about that, but I think `secure APIs` instead of `secured APIs`.

docs/developer/architecture/centralized-cli.rst
slaperche-scality: Hum, completion with <TAB> can be shell-specific (for the different operation)
I'm not sure we want to thread into this direction.

I think it's easier and as important to have a useful `-h/--help`, like `foo -h` list the top level commands, then a `foo command -h` gives the help of `command` and so on.
Like what `kubectl` (and others) does.

docs/developer/architecture/centralized-cli.rst
slaperche-scality: `s/yaml/YAML/` and `s/json/JSON/`

slaperche-scality (commented):
I think there is a misuse of the term CLI in the document.

docs/developer/architecture/centralized-cli.rst
Ebaneck: When provisioning new OIDC static users we mandatorily need to provide password hashes so I accept with @TeddyAndrieux that user administration should be an exception here.

docs/developer/architecture/centralized-cli.rst
Ebaneck: For me, autocompletion for both object names and operations is a nice to have. Btw, @sayf-eddine-scality what is the cause of the slowness when autocompleting pods/svc/volumes names??

sayf-eddine-scality: https://github.com/kubernetes/kubernetes/issues/82883
loading a huge number of objects and then parsing their metadata takes some time (counting also network latency) and this happens every time you press Tab.

docs/developer/architecture/centralized-cli.rst
Ebaneck: `s/Also/Also,`
`s/expose/exposes`

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Yes, but like I said in my previous comment this should be at the very beginning of the script

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Auto completion for objects name is a bad idea, because to discover btw you need to access apiserver so you likely need a kubeconfig ... and yes if you have tones of objects ect ...

> I think it's easier and as important to have a useful -h/--help, like foo -h list the top level commands, then a foo command -h gives the help of command and so on.
Like what kubectl (and others) does.

Agree, personnaly think having nice help text is good enough but with `cmd` lib from python we can have also an interpreter with completion

slaperche-scality: > Agree, personnaly think having nice help text is good enough but with cmd lib from python we can have also an interpreter with completion

Ha I was thinking about `argparse`, but if we have a lib that does it for us then let's have completion on subcommands :slightly_smiling_face:

Ebaneck: > Auto completion for objects name is a bad idea, because to discover btw you need to access apiserver so you likely need a kubeconfig ... and yes if you have tones of objects ect ...
> 
If we provide OIDC login creds at the very beginning of the script as you say then we can use the resulting OIDC auth token to generate a kubeconfig and use for subsequent operations

> > I think it's easier and as important to have a useful -h/--help, like foo -h list the top level commands, then a foo command -h gives the help of command and so on.
> > Like what kubectl (and others) does.
> 
> Agree, personnaly think having nice help text is good enough but with `cmd` lib from python we can have also an interpreter with completion

docs/developer/architecture/centralized-cli.rst
thomasdanan: yes agree @TeddyAndrieux will change to apiserver

docs/developer/architecture/centralized-cli.rst
thomasdanan: okay my bad ... will remove this

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,62 @@
...
+and easy to use set of tools in order to administrate and operate a finite set
+of functionalities.
+
+Because kubectl CLI is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+
+- CLI follows kubectl style: kubectl <action> <object>
+- CLI provides an exhaustive help, per action, with relevant examples
+- CLI is not interactive (except maybe for very first install script)
```
TeddyAndrieux: > except maybe for very first install script

No for me it's really only for password and that's it. And this password should be ask at the very beginning of the command execution so that the user launch the command, enter the password if any and then not more interactivity

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,62 @@
...
+- CLI should not require password input
+- CLI implementation relies on secure APIs
+- CLI support <action> completion for easy discovery
+- CLI output is standardized and human readable by default
+- CLI output can be formatted in JSON or YAML
+- CLI can be executed from outside of the cluster (except maybe for very first
+  install script)
```
TeddyAndrieux: > except maybe for very first install script

No, not only. You also have for example the iso-manager, upgrade and downgrade script that need to run on the bootstrap node. For me you should remove this requirement or have a better "except" part but ... I don't know how you can define it for me it's something like "except for script that need to (re)start salt-master container" but it maybe not only this

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,145 @@
...
++------------+------------+-----------------------------------------------+
+| cluster    | downgrade  | dest-version, dry-run                         |
++------------+------------+-----------------------------------------------+
+| etcd       | health     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | backup     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | restore    | backup-file                                   |
++------------+------------+-----------------------------------------------+
+| bootstrap  | restore    | backup-file                                   |
```
TeddyAndrieux: 2 time restore

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Cluster Administration
+^^^^^^^^^^^^^^^^^^^^^^
+
+tools: mk8sctl
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| cluster    | install    |                                               |
```
TeddyAndrieux: What `install cluster` mean ? It's bootstrap right ? Or it's something with a platform description file and this do bootstrap + expansions ?

Also I think we want a command to `install bootstrap`.

Since we use `<action> <resource>` style I think you should have `deploy` or `install` but not a mix because command will look like 
`kubectl install cluster`

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,145 @@
...
+|            |                  | ssh port, ssh-key-path, sudo-required,  |
+|            |                  | roles, manifest-template-path           |
++------------+------------------+-----------------------------------------+
+| the manifest-template-path is the path to the manifest describing the   |
+| resource. Everything that is in the command line, override the manifest |
+| content. This apply to Node and Volumes.                                |
++------------+------------------+-----------------------------------------+
+| node       | edit             | name, hostname (to change node hostname)|
++------------+------------------+-----------------------------------------+
+| node       | deploy(or apply?)| <list of nodes>, dry-run                |
```
TeddyAndrieux: Hmm it's not part of `kubectl`

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,145 @@
...
+Requirements
+------------
+
+Cluster Resources Administration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+tools: kubectl
+
++------------+------------------+-----------------------------------------+
+| Resource   | action           | parameters                              |
```
TeddyAndrieux: If we follow kubectl style I'm not sure it's good to describe it like this. 

I think you should list for each `action` what `resource` you can use. 

BUT from my understanding we want several different `tool` (at least 2) so that we have one that can be run from anywhere with the right kubeconfig and another that need to be on the bootstrap node  cc @NicolasT

thomasdanan: We want all commands to behave the same, i.e. <tool> <action> <resource>. But in terms of requirements, I prefer to express the list of tasks we want to expose through this CLI and organize the tasks per domain.

I am not sure to understand why we should enforce the fact that some tools must run on the bootstrap. It looks like more a design constraint than a requirement no? Maybe at the end, we will say, some actions cannot be run from outside of the cluster and that's okay ...

docs/developer/architecture/centralized-cli.rst
thomasdanan: isn't it the kind of thing we can achieve with the kubectl plugin? i.e. enrich kubectl with new functionalities? but from a usage point of view, the administrator will invoke kubectl?

docs/developer/architecture/centralized-cli.rst
thomasdanan: you are right, it should be `install` `bootstrap`.
Not sure I understand your second comment

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Hmm for this one is exacly one we do NOT want to put in a kubectl plugin because it's an interaction with salt directly so not part of kubectl at all

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: True it's not part of the requirement directly

docs/developer/architecture/centralized-cli.rst
gdemonet: (can't review the whole PR atm, but just my 2 cents on this question)

We **can** call SaltAPI from `kubectl`. We **could** also send commands in the `salt-master` container. Nothing prevents us from interacting with Salt from a `kubectl` plugin.

There are some commands which will require direct interaction on the host however, maybe because the operation will trigger a master restart (e.g. upgrade/downgrade), or because it does not make sense to do it from somewhere else (e.g. managing Metal or Solution ISOs on the Bootstrap). But deploying a Node is a standard operation you may want to do from outside the cluster (we do it from the UI already).

docs/developer/architecture/centralized-cli.rst
thomasdanan: @TeddyAndrieux the requirement is to be able to run everything from outside of the cluster, as we can do with kubectl. It is maybe not possible, you will say it in the design.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: For node you have `node` `deploy` and for bootstrap it's `bootstrap` `install` for me you should have only `deploy` or only `install`

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Of course we can do it from a kubectl plugin but for me it's a bad idea. Kubectl plugin shouldn't interact with salt IMO.

> There are some commands which will require direct interaction on the host however, maybe because the operation will trigger a master restart (e.g. upgrade/downgrade), or because it does not make sense to do it from somewhere else (e.g. managing Metal or Solution ISOs on the Bootstrap). But deploying a Node is a standard operation you may want to do from outside the cluster (we do it from the UI already).

???
I do not see why it's related to kubectl plugin or not ?? You can execute your kubectl plugin from outside the cluster, you can also execute our own CLI from outside the cluster

gdemonet: > you can also execute our own CLI from outside the cluster

No. This "extra CLI" we'll build for Bootstrap-local operations will not be a "redistributable package", just an executable in an ISO.

Otherwise, how do you allow this CLI to authenticate against Salt without a kubeconfig? Would you reimplement the login dance with K8s API to then be able to reach Salt API?
Also, what's the benefit of having two CLIs if they can both work anywhere?

docs/developer/architecture/centralized-cli.rst
thomasdanan: I think it is okay and easier to talk about CLI

docs/developer/architecture/centralized-cli.rst
thomasdanan: okay understood

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: > No. This "extra CLI" we'll build for Bootstrap-local operations will not be a "redistributable package", just an executable in an ISO.

So for you all extra CLI are for stuff that cannot go outside the cluster otherwise it's a kubectl plugin ??
For me no.

To add a node you can perfectly have a CLI that work outside of the cluster or a kubectl plugin but IMO kubectl plugins shouldn't interact with salt

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+===============
+
+Context
+-------
+
+MetalK8s comes with a set of services to operate and monitor the K8s cluster.
+All operations that need to be performed by the Platform Administrator could be
+categorized as follow:
+
+- Cluster Resources Administration (Nodes, Volumes, Deployments, ...)
```
NicolasT, line 11: What are *Deployments* in this context?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+- Cluster Administration (Install, Upgrade, Downgrade, Backup, Restore, ...)
+- Solution Administration (CRUD Environment, Import/Remove Solution, ...)
+- Cluster Service Administration (Configure Dex, Prometheus, Alert Manager,
+  ...)
+
+K8s provides the kubectl CLI, which is distributed and enables
```
NicolasT: *distributed*?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+
+- CLI follows kubectl style: kubectl <action> <resource>
+- CLI provides an exhaustive help, per action, with relevant examples
+- CLI provides <action> help when the command is not valid
+- CLI is not interactive (except if password input is needed)
+- CLI should not require password input
+- CLI provides a dryrun mode for intrusive operations
+- CLI provides a verbose (or debug) mode
```
NicolasT, line 44: Note, `kubectl` verbosity is not a toggle but a value.

docs/developer/architecture/centralized-cli.rst
NicolasT: I'm with @TeddyAndrieux here. Putting something in a requirement means "We really need this for some customer use-case". As such, either it's a must (requirement) and we need to figure out some design to accommodate the need, or it's not a requirement.

Now, in the case where 'local things' are required (indeed, e.g. ISO import, where one needs to refer to a *local* path to an ISO), *requiring* this to be available 'from outside' seems odd.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+- CLI implementation relies on secure APIs
+- CLI support <action> completion for easy discovery
+- CLI output is standardized and human readable by default
+- CLI output can be formatted in JSON or YAML
+- CLI can be executed from outside of the cluster (except maybe for very first
```
NicolasT: This seems to imply there's a single CLI (tool) somewhere. I don't think that's (necessarily) the intent?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+  install script)
+
+When it is possible, it would make sense to leverage kubectl plugin
+
+All functionalities are exposed through 2 distincts CLI:
+- kubectl (enriched using kubectl approach)
+- mk8sctl: a new CLI, exposing specific MetalK8s functionalities
```
NicolasT: `metalk8sctl` or similar, please... We aim for tab-completion of actions, so why not use a shell's tab-completion.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+- CLI implementation relies on secure APIs
+- CLI support <action> completion for easy discovery
+- CLI output is standardized and human readable by default
+- CLI output can be formatted in JSON or YAML
+- CLI can be executed from outside of the cluster (except maybe for very first
+  install script)
+
+When it is possible, it would make sense to leverage kubectl plugin
+
+All functionalities are exposed through 2 distincts CLI:
```
NicolasT: Here you say *all* functionality should be exposed using *both* a `kubectl` extension *as well as* some other tool. That seems odd, unintuitive and confusing.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+- kubectl (enriched using kubectl approach)
+- mk8sctl: a new CLI, exposing specific MetalK8s functionalities
+
+In order to operate the cluster with mk8sctl from outside of the cluster, a
```
NicolasT: - For any `kubectl` extra functionality, some binary/binaries need to be installed anyway as well.
- I'd say: `metalk8scli` for on-node actions, `kubectl` plugin for anything that can be done remotely.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+specific pkg (for each OS) or a procedure explaining how to deploy it is
+available.
+The mk8scli is deployed and available by default on bootstrap nodes
```
NicolasT: As well as the `kubectl` plugin, then.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Cluster Resources Administration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**tool: kubectl**
+
++------------+------------------+-----------------------------------------+
+| Resource   | action           | parameters                              |
++============+==================+=========================================+
+| node       | add              | name, ssh-user, hostname or ip,         |
```
NicolasT: That's not how it works. It'll be something like `kubectl metalk8s create node ...`

NicolasT: The interactive action of `kubectl` is `create`, not `add`.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+|            |                  | ssh port, ssh-key-path, sudo-required,  |
+|            |                  | roles, manifest-template-path           |
++------------+------------------+-----------------------------------------+
+| the manifest-template-path is the path to the manifest describing the   |
```
NicolasT: That's not in line with other `kubectl` commands. Why would one want to provide such manifest?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+| resource. Everything that is in the command line, override the manifest |
+| content. This apply to Node and Volumes.                                |
++------------+------------------+-----------------------------------------+
+| node       | edit             | name, hostname (to change node hostname)|
```
NicolasT: No. `kubectl edit` opens an editor to interactively edit an object. One can already `kubectl edit node ...` and do whatever.

NicolasT: What would 'changing the hostname' do? You can't change the `name` of an object in K8s...

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
++------------+------------------+-----------------------------------------+
+| node       | deploy           | <list of nodes>, dry-run                |
++------------+------------------+-----------------------------------------+
+| node       | delete, drain,   | <list of nodes>                         |
+|            | taint, label     |                                         |
++------------+------------------+-----------------------------------------+
+| volume     | add              | name, nodeName, storageClassName,       |
```
NicolasT: This (`devicePath` and such) is too tied to the current type of volume we may support (and only one of the two we do support). This needs to be revised into a proper CLI UX.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+| cluster    | downgrade  | dest-version, dry-run                         |
++------------+------------+-----------------------------------------------+
+| etcd       | health     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | backup     |                                               |
++------------+------------+-----------------------------------------------+
+| bootstrap  | restore    | backup-file                                   |
++------------+------------+-----------------------------------------------+
+| cluster    | edit-conf  | conf-param=conf-value                         |
+|            |            | ex: ip-in-ip=true                             |
```
NicolasT: I'm pretty sure you'll run into trouble with this.

`BootstrapConfiguration` is a structured object, which here you're flattening into... what? Which subactions would you support?...

What you'll want instead is a config check (is the YAML valid?), and a (re)deploy-with-new-config.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Solution Administration
+^^^^^^^^^^^^^^^^^^^^^^^
+
+**tool: mk8sctl**
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| solution   | import     |                                               |
```
NicolasT: Without a path to some ISO?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Cluster Service Administration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**tool: mk8sctl**
+
++------------+------------+-----------------------------------------------+
+| Resource   | action     | parameters                                    |
++============+============+===============================================+
+| grafana    | edit-conf  | conf-param=conf-value                         |
```
NicolasT: This (and the ones below) won't work. If there's structured editing going on, then present the user with an editor.

docs/developer/architecture/centralized-cli.rst
slaperche-scality: @TeddyAndrieux That's the second time you say `IMO kubectl plugins shouldn't interact with salt`, do you have reasons/arguments for that?

I've no opinion on that part, but giving your argument for this could help convince people (or at least understand your point of view).

NicolasT: Why not, @TeddyAndrieux ? There's SaltAPI they can work with, using the `KubeConfig` token.

Now, to be clear: `kubectl` plugins *only allow to add top-level commands*. So you *can't* write a plugin which all of a sudden would give you a `kubectl node deploy` command or something along those lines. You could have `kubectl metalk8s deploy node`.

NicolasT: OK, in the above: you *could* have `kubectl node deploy` because `kubectl node` doesn't exist., but that's mere coincidence and I think it'd be very bad style (and potential future-incompatibility) to hijack random nouns.

docs/developer/architecture/centralized-cli.rst
xaf-scality: Deployments means Expansions ?

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,172 @@
...
+The goal of the project is to provide MetalK8s administrator with an intuitive
+and easy to use set of tools in order to administrate and operate a finite set
+of functionalities.
+
+Because kubectl is already in place and is well known by Kubernetes
+administrators, it will be used as a standard to follow for all other MetalK8s
+CLIs:
+
+- CLI follows kubectl style: kubectl <action> <resource>
+- CLI provides an exhaustive help, per action, with relevant examples
```
xaf-scality: auto completion for actions (and not values) like ringsh ?

docs/developer/architecture/centralized-cli.rst
xaf-scality: The intent is to have less # of CLI, so the more we can aggregate in one CLI the best it is.

docs/developer/architecture/centralized-cli.rst
xaf-scality: Well you can execute remotely some 'local things' with local path, and yes, it's a bit weird.
The point is why we don't want this ? Because of technical difficulties or conceptual things or security things ?
Moreover could the CLI upgrade itself ?

docs/developer/architecture/centralized-cli.rst
xaf-scality: I agree why having two CLI that will do the same ?

docs/developer/architecture/centralized-cli.rst
xaf-scality: this one for CLI will be a very long command line with all the "optional values" for each type of volume. It's almost like writing the whole yaml file describing the volume in this case

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: For me kubectl plugins should only act on apiserver and not on something we deploy on our side, for me it's something that can work on every kubernetes deployment ...

I agree and I know everything is doable from kubectl plugin but for me it's not a good idea. Anyway as you want

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT @xaf-scality @TeddyAndrieux let's discuss this one during today's meeting.

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: okay so If I understand correctly, with kubectl, we can add a new action but we can't override an existing one (like kubectl create node, but with additional arguments)?

docs/developer/architecture/centralized-cli.rst
thomasdanan: I was referring to Deployment Object on K8s API

docs/developer/architecture/centralized-cli.rst
thomasdanan: Indeed, not the right term, wanted to say that we can execute it anywhere since we have Network access to the cluster ...

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT my suggestion is to have, in addition to kubectl, one other CLI: mk8sctl for all platform operations. Note that with solutions, we may have additional CLI for Zenko, HD and SOFS. I think it makes sense to limit the number of CLI we expose.

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT actually it will do all of this: https://metal-k8s.readthedocs.io/en/latest/operation/changing_node_hostname.html

docs/developer/architecture/centralized-cli.rst
gdemonet: > for me it's something that can work on every kubernetes deployment 

Not really. Core `kubectl` functionalities, yes. But **plugins** are "additional functionality". For instance, we could have a `kubectl-prom` plugin which helps manage whatever comes with the Prometheus Operator.

> with kubectl, we can add a new action but we can't override an existing one (like kubectl add node, but with additional arguments)?

The way plugins work is quite simple. Put a binary called `kubectl-<plugin-name>` in the right place, and whenever you will run `kubectl <plugin-name> ...`, the call will be forwarded to this binary by `kubectl`, nothing more.
So indeed, there's no way to override existing `kubectl` functionality with plugins, only to extend it. In our case, as Nicolas said, we'll likely have all our commands under `kubectl metalk8s ...`.

docs/developer/architecture/centralized-cli.rst
thomasdanan: @xaf-scality @NicolasT What I was suggesting is to have manifest templates and being able to override some values in command line ... to be discussed during today's meeting

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT can we do everything with kubectl and kubectl plugin?

thomasdanan: @xaf-scality the 2 commands are not doing the same things but again if we can do everything with kubectl that's even better ...

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT, in order to simplify the number of inputs a user should provide in order to perform a specific operation.

docs/developer/architecture/centralized-cli.rst
thomasdanan: @NicolasT my problem is when we have to edit such a yaml file, it is super scary, it is error prone, and unless you are familiar with kubernetes and yaml, you will be very likely unable to perform your operation.

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
gdemonet: Solutions could provide their own `kubectl` plugin ;)

docs/developer/architecture/centralized-cli.rst
gdemonet: We could indeed have a "simplified" node object with just the few fields we know are editable. But that's quite a lot of effort, whereas we could expose the `set` verb that `kubectl` has for some resouces.

Something like `kubectl metalk8s set ssh-host node/master-1 10.200.1.2`.

gdemonet: Oh, that's not "changing a hostname". This procedure is basically replacing a Node entirely (it's not even draining it beforehand). Not sure we want to expose this as a command, especially not one that makes you believe it's "just changing a value in some YAML".

docs/developer/architecture/centralized-cli.rst
thomasdanan: Okay, thanks for the explanation ... I think I understand now

docs/developer/architecture/centralized-cli.rst
NicolasT: That's kinda confusing because MetalK8s doesn't really 'care' that much about (K8s) *Deployments*. I'd suggest another term ('cluster services'?)

gdemonet: Maybe 'addons'?

docs/developer/architecture/centralized-cli.rst
thomasdanan: removed, only kubectl metalk8s plugin should be executable from outside of the cluster

docs/developer/architecture/centralized-cli.rst
thomasdanan: Actually, I am talking about any K8s resource we can create, edit, delete through k8s apiserver. I suggest k8s resource

docs/developer/architecture/centralized-cli.rst
thomasdanan: smart !!!

docs/developer/architecture/centralized-cli.rst
thomasdanan: will remove this one

docs/developer/architecture/centralized-cli.rst
thomasdanan: will remove this concept of manifest template

docs/developer/architecture/centralized-cli.rst
thomasdanan: will remove this one

docs/developer/architecture/centralized-cli.rst
thomasdanan: done

docs/developer/architecture/centralized-cli.rst
gdemonet: But this should be left to `kubectl` as we said... Do you still want to mention this in the requirements?

xaf-scality (approved):

docs/developer/architecture/centralized-cli.rst
```diff
@@ -0,0 +1,151 @@
...
+- CLI output can be formatted in JSON or YAML
+
+When it is possible, it would make sense to leverage kubectl plugin
+
+Most functionalities are exposed through 2 distinct CLI:
+
+- kubectl: enriched with metalk8s plugin, to interact with both k8s apiserver
+  and salt API, and that can be executed from outside of the cluster.
+- metalk8sctl: a new CLI, exposing specific MetalK8s functionalities, that are
+  not interacting with k8s apiserver, and that must be executed on cluster node
```
TeddyAndrieux, line 57: > that are not interacting with k8s apiserver

It's not true for me, some of them interact with apiserver just they must be run on the machine because they interact also with local machine

thomasdanan: ok, will do that in one other PR if you don't mind

Issue #2267, opened 2020-02-26
Generate junit (or whatever) for each steps of CI


TeddyAndrieux:
**Component**:

'CI'

**Why this is needed**:

We have bunch of steps in a CI but we do not generate any report

**What should be done**:

We should generate a report for each and every step of the CI so that we can include this report in something (like testrail for example)

**Implementation proposal** (strongly recommended):

Generate a junit report to use testrail to see all CI tests result

PR #2326, opened 2020-03-23
Improvement/gh 2267 generate simple junit for eve steps

TeddyAndrieux:
**Component**:

'build'

**Context**: 

Currently we do not have any report for build status.

**Summary**:

This commit add a very very simple script to generate really simple junit for each steps
of the build. 1 junit with 1 test suite and 1 test case per step.
A better junit reporting should be done directly by eve in the futur,
this junit just allow us to have a report giving use the result of each
build steps

Junit example:

Success:
```
<?xml version="1.0" ?>
<testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
	<testsuite disabled="0" errors="0" failures="0" name="upgrade" skipped="0" tests="1" time="0">
		<testcase name="simple environment" classname="single-node.centos7"/>
	</testsuite>
</testsuites>
```

Fail:
```
<?xml version="1.0" ?>
<testsuites disabled="0" errors="0" failures="1" tests="1" time="0.0">
	<testsuite disabled="0" errors="0" failures="1" name="upgrade" skipped="0" tests="1" time="0">
		<testcase name="simple environment" classname="single-node.centos7">
			<failure type="failure" message="FAILED"/>
		</testcase>
	</testsuite>
</testsuites>
```

---

Fixes: #2267

### DIFF
--- a/eve/generate_junit_result.sh
+++ b/eve/generate_junit_result.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# Simple ugly bash script to generate really really simple junit file
+# with just the result of the build step, waiting for something integrated
+# directly in `eve`
+
+FAILURE=0
+ERROR=0
+TYPE=""
+
+case "$FINAL_STATUS" in
+    "SUCCESSFUL")
+        ;;
+    "FAILED")
+        FAILURE=1
+        TYPE="failure"
+        ;;
+    *)
+        ERROR=1
+        TYPE="error"
+        ;;
+esac
+
+
+cat << EOF
+<?xml version="1.0" ?>
+<testsuites disabled="0" errors="$ERROR" failures="$FAILURE" \
+tests="1" time="0.0">
+    <testsuite disabled="0" errors="$ERROR" failures="$FAILURE" \
+name="$TEST_SUITE" skipped="0" tests="1" time="0">
+EOF
+
+if [ "$TYPE" ]; then
+
+    cat << EOF
+        <testcase name="$TEST_NAME" classname="$CLASS_NAME">
+            <$TYPE type="$TYPE" message="$FINAL_STATUS"/>
+        </testcase>
+EOF
+
+else
+    cat << EOF
+        <testcase name="$TEST_NAME" classname="$CLASS_NAME"/>
+EOF
+fi
+
+cat << EOF
+    </testsuite>
+</testsuites>
+EOF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -444,7 +444,7 @@ stages:
           name: Run all linting targets
           command: ./doit.sh lint
           usePTY: true
-          haltOnFailure: false
+          haltOnFailure: true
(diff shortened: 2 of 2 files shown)


eve/generate_junit_result.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Simple ugly bash script to generate really really simple junit file
+# with just the result of the build step, waiting for something integrated
+# directly in `eve`
```
gdemonet, line 5: Don't when something will land in Eve, but in the meantime, can we use "real" tooling instead of some half-assed XML rendering from Bash? Something like [python-junit-xml](https://github.com/kyrus/python-junit-xml) could be nice :)

eve/main.yml
```diff
@@ -290,6 +290,11 @@ models:
           [[ ${STEP_NAME:-} ]] && BUILD_STATUS_DIR+="/build_status/$STEP_NAME"
           mkdir -p "$BUILD_STATUS_DIR"
           echo -n "$FINAL_STATUS" > "$BUILD_STATUS_DIR/.final_status"
+          if [[ ${STEP_NAME:-} ]]; then
+              git checkout "%(prop:branch)s" && \
+              eve/generate_junit_result.sh \
+                  > "$BUILD_STATUS_DIR/junit_status.xml"
+          fi
```
gdemonet: Oh, I see, we're only generating the report for the "Upload .final_status" step, relying on the `FINAL_STATUS` env var to know the "step" outcome (while in practice, we're using it to show the "stage" outcome).

Definitely a hack, but a good enough first step!

gdemonet (dismissed):
Not sure if we can use it already, but as a first implem, looks sufficient to me :) Let's just not keep this approach eternally and forget about it...

eve/generate_junit_result.sh
```diff
@@ -0,0 +1,39 @@
...
+        FAILURE=1
+        TYPE="failure"
+        ;;
+    *)
+        ERROR=1
+        TYPE="error"
+        ;;
+esac
+
+echo '<?xml version="1.0" ?>'
```
NicolasT: Wouldn't it be much nicer to use a heredoc? :thinking:

eve/generate_junit_result.sh
TeddyAndrieux: Then we need to install a new package (only available from epel btw) for a really simple xml for me it's not worth for this first step

eve/generate_junit_result.sh
TeddyAndrieux: Yes true

eve/generate_junit_result.sh
gdemonet: Honestly, I'd rather have a single step running after a full Eve build in a `kube_pod` worker, with the dependencies it needs to generate whatever report TestRail needs. But yeah, no need for this PR.

gdemonet (approved):

Issue #2331, opened 2020-03-25
Failing post-merge test for Cluster Service Configurations on branch 2.6


Ebaneck:
**Component**:

'test', 'kubernetes'

**What happened**:

Post-merge tests on branch 2.6 continuously fail mainly because after the test setup and execution phase, we do not tear down and return the ConfigMap to its original state.

We end up with the following error output:

```>           pytest.fail(error_msg)
E           Failed: There is not exactly '2' pod(s) labeled 'app.kubernetes.io/name=dex' running in namespace 'metalk8s-auth': assert 3 == 2
E            +  where 3 = len([{'api_version': None,\n 'kind': None,\n 'metadata': {'annotations': {'checksum/config': 'ad6c30825bdc913fdb7dd6486fd137...',\n            'reason': None,\n            'start_time': datetime.datetime(2020, 3, 23, 0, 56, 50, tzinfo=tzlocal())}}])
```

Since during testing we bump the Dex replicas count to 3, we need to find a way to revert this value to 2

**What was expected**:

Post-merge test should pass!

**Steps to reproduce**

**Resolution proposal** (optional):

Add a teardown method that returns the ConfigMap to the original state.

PR #2332, opened 2020-03-26
tests: Refactor Cluster and Service Configuration tests to include teardown phase

Ebaneck:
**Component**:

'test', 'csc'

**Context**: 

See #2331

**Summary**:

Refactor CSC to include a teardown method after setup and execution phase

**Acceptance criteria**: 

---

Closes: #2331

### DIFF
--- a/tests/post/steps/test_service_configuration.py
+++ b/tests/post/steps/test_service_configuration.py
@@ -18,6 +18,11 @@ from tests import utils
 def k8s_appsv1_client(k8s_apiclient):
     return AppsV1Api(api_client=k8s_apiclient)
 
+@pytest.fixture
+def csc(k8s_client):
+    return ClusterServiceConfiguration(k8s_client)
+
+
 # }}}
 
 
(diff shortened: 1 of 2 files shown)


tests/post/steps/test_service_configuration.py
```diff
@@ -158,6 +160,21 @@ def _wait_for_deployment():
...
+        original_csc_obj = check_csc_configuration['csc_obj']
+
+        patch = {
+            'data': {
+                'config.yaml': yaml.safe_dump(
+                    original_csc_obj, default_flow_style=False
+                )
+            }
+        }
+        csc.update(original_name, original_namespace, patch)
```
TeddyAndrieux: I would prefer to do that in a teardown function
Because here if for what ever reason we want to add another `then` check .... that will not work as after this `then` you reset the csc

slaperche-scality: Maybe we need an explicit `then` step, like `Then we restore the original state of the config map foo in namespace bar`.
It may be annoying, but at least it's explicit;

Otherwise, yeah we have to assume that this `then` comes last or that we won't have more `then` checks on the modified object after this one.

TeddyAndrieux: Hmm adding this "restore `then`" look good and clean to me

Ebaneck: ok

TeddyAndrieux (approved):
LGTM

Issue #2260, opened 2020-02-26
Solution Framework fails to render some YAML from a Solution


slaperche-scality:
**Component**:

solution, salt

**What happened**:

When trying to deploy a Solution using `solution-manager.sh`, it dies with the following error:

```
[CRITICAL] Rendering SLS 'metalk8s-2.4.1:metalk8s.orchestrate.solutions' failed: while parsing a block mapping

  in "<unicode string>", line 37, column 9:

          - tem: "{Apply apiextensions.k8s.i ...

            ^

expected <block end>, but found '<scalar>'

  in "<unicode string>", line 85, column 59:

     ...             whose key field is "key", the operator is "In",

                                         ^

app-1_master:

    Data failed to compile:

----------

    Rendering SLS 'metalk8s-2.4.1:metalk8s.orchestrate.solutions' failed: while parsing a block mapping

  in "<unicode string>", line 37, column 9:

          - tem: "{Apply apiextensions.k8s.i ...

            ^

expected <block end>, but found '<scalar>'

  in "<unicode string>", line 85, column 59:

     ...             whose key field is "key", the operator is "In",

                                         ^
```

I can provide the problematic YAML if needed.

**What was expected**:

Given that the YAML of the CRD is valid, it should be handled without error by the Solution Framework.

**Steps to reproduce**

Simply run `/srv/scality/metalk8s-2.4.1/solution-manager.sh -v -a /srv/scality/solution.iso` on a MetalK8s 2.4.1 cluster (I can provide the ISO if needed)

alexandre-allard:
fixed by #2342

PR #2342, opened 2020-03-30
Bugfix/2260 fail to render crds

alexandre-allard:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2260

### DIFF
--- a/salt/metalk8s/orchestrate/solutions/deploy-components.sls
+++ b/salt/metalk8s/orchestrate/solutions/deploy-components.sls
@@ -16,10 +16,10 @@ include:
 {%- macro manipulate_solution_components(solution, present=true) %}
   {%- if present %}
     {%- set action = "Apply" %}
-    {%- set renderer = kubernetes_present_renderer %}
+    {%- set state = "metalk8s_kubernetes.object_present" %}
   {%- else %}
     {%- set action = "Remove" %}
-    {%- set renderer = kubernetes_absent_renderer %}
+    {%- set state = "metalk8s_kubernetes.object_absent" %}
   {%- endif %}
 
   {# CRDs management #}
@@ -35,18 +35,10 @@ include:
         )[pillar.bootstrap_id]['ret'] %}
 
   {%- for crd_file in crd_files %}
-    {%- set sls_content = salt.saltutil.cmd(
-            tgt=pillar.bootstrap_id,
-            fun='slsutil.renderer',
-            kwarg={
-                'path': crd_file,
-                'default_renderer': renderer,
-            },
-          )[pillar.bootstrap_id]['ret'] %}
+
 {{ action }} CRD "{{ crd_file }}" for Solution {{ solution.name }}:
-  module.run:
-    - state.template_str:
-      - tem: "{{ sls_content | yaml }}"
+  {{ state }}:
+    - name: {{ crd_file }}
 
   {%- endfor %} {# crd_file in crd_files #}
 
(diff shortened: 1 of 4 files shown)


salt/_modules/metalk8s.py
```diff
@@ -275,3 +275,14 @@ def check_pillar_keys(keys, refresh=True, pillar=None, raise_error=True):
...
+
+def file_read(path):
+    """Returns the content of a given file
+
+    Arguments:
+        path (str): path to a file
+    """
+
+    with open(path, 'r') as fd:
+        return fd.read()
```
slaperche-scality: > passing the result to the yaml filter is failing, because it sees the content as a single string

That's weird, because here you return a single string as well so why does it work better?

gdemonet (commented):
I'm not sure that's the problem.

It's not a problem of Jinja markup, but YAML markup. We were trying to espace the YAML data as a string, using:

```jinja
module.run:
  - state.template_str:
    - tem: "{{ sls_content | yaml }}"
```

If the YAML file contains quotes `"`, then the escaping will stop mid-string and the resulting YAML will be invalid.

IMO, inlining the file, which comes from an unknown source, isn't a great idea.
From what I understand, we still want to gather the file from the bootstrap minion, because we may not have the files in the Salt Master container. Honestly, I think having access to `/srv/scality` entirely in the Salt Master container would be a much more acceptabe solution.

If this is true, then you could replace the whole `file_read` with a simple:

```jinja
  {%- for crd_file in crd_files %}
{{ action }} CRD "{{ crd_file }}" for Solution {{ solution.name }}:
  metalk8s_kubernetes.object_present:
    - name: {{ crd_file }}

  {%- endfor %}
```

salt/metalk8s/orchestrate/solutions/deploy-components.sls
```diff
@@ -37,16 +37,16 @@ include:
...
             },
           )[pillar.bootstrap_id]['ret'] %}
+
 {{ action }} CRD "{{ crd_file }}" for Solution {{ solution.name }}:
-  module.run:
-    - state.template_str:
-      - tem: "{{ sls_content | yaml }}"
+  {{ state }}:
+    - manifest:
+        {{ sls_content | yaml(False) | indent(8) }}"
```
slaperche-scality: What is the use of `False` here?

alexandre-allard:
Ok, it wasn't really clear in the issue, I though it was due to some Jinja markup as written in my commit message.
My first idea was to use the `name` argument but the file is not accessible on the master, I'll try to mount the whole `/srv/scality` directory in the salt master container.

salt/metalk8s/orchestrate/solutions/deploy-components.sls
alexandre-allard: It corresponds to the `default_flow_style=False` argument of the `yaml.dump()` function in Python.

salt/_modules/metalk8s.py
alexandre-allard: Well, I must admit that the commit message is not really good.
But, Salt is returning something that's not parsable by the yaml filter.
It's like the whole content is escaped with quotes, the YAML filter sees only a single string, instead of a complex object and renders it badly.
Anyway, I've removed this part. :)

gdemonet (dismissed):
LGTM, what do you think @NicolasT @TeddyAndrieux?

alexandre-allard:
We could also mount each and every solution archives at import/activation time, but it means restarting the salt master container every time there is a new solution or when we remove an old one.

gdemonet (dismissed):

gdemonet (dismissed):

gdemonet (approved):

PR #2350, opened 2020-04-02
[Solutions] Upgrade/downgrade the deployed solutions from UI

### LINKED ISSUE #2337: Upgrade the deployed solutions from UI
**Component**: ui, solutions

**Why this is needed**:

Currently, in order to upgrade the deployed solutions the only way we have is to click on the `add` button and fill in the form with the new version, which is counter-intuitive.

**What should be done**:
We should be able to upgrade the solutions by a dedicate button `edit` or `upgrade` button ( in the future we may need a button to redirect to the solution UI)

There are two probable proposals:
**1) Expandable row table**

image

- Pros: Centralized the information for Environment.

- Cons: We need to include the development cost for the Table component. Currently, we don't support the expandable row table, so we need first implement it in core-ui. 
Implement detail of the table: We should be able to dynamically change the rowHeight of the table
（https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md）

**2) Redirect to the environment detail page by clicking on the specific environment in the Environment List**

image

- Pros: Easier to implement and have more space to display more environment detail( e.g. a link redirect to the solution UI...)
- Cons: A new page without enough information, I mean, the content is very few.

Requirements:
1. Be able to click on the row when there is a deployed version.
2. `Upgrade` & `Downgrade` button should appear only if there are some available versions to upgrade / downgrade.

**Test plan**:
TBD

ChengYanJin:
**Component**: ui, solutions

**Context**:  Currently, in order to upgrade the deployed solutions the only way is to click on the add button and fill in the form with the new version, which is counter-intuitive.

**Summary**:
- We can access the solution detail page from the environment list by click on the specific row
- Upgrade & Downgrade button should appear only if there are some available versions to upgrade / downgrade.
- When click on the `upgrade/downgrade` button, it will pop up a modal with available upgrade/downgrade version 
- Rename the sidebar from `solutions` to `environment` as well as the URL from `/solutions` to `/evironments` 

**Acceptance criteria**: 

1. Environment list page

 image
By clicking on the specific environment `dev` for example,  we will go to the dev detail page, given the condition that we already deployed at least one solution in this environment.

2. Solution detail page

image

3. Upgrade/downgrade a solution

image

After clicking on the`Upgrade/Downgrade` button, there will be a modal pop up to edit the deployed solution version

---

Closes: #2337

### DIFF
--- a/ui/src/ducks/app/solutions.js
+++ b/ui/src/ducks/app/solutions.js
@@ -64,7 +64,7 @@ export function setSolutionsRefeshingAction(payload) {
   return { type: SET_SOLUTIONS_REFRESHING, payload };
 }
 
-export const setEnvironmentsAction = environments => {
+export const setEnvironmentsAction = (environments) => {
   return { type: SET_ENVIRONMENTS, payload: environments };
 };
 
@@ -92,15 +92,15 @@ export function deleteEnvironmentAction(envName) {
 }
 
 // Selectors
-export const solutionsRefreshingSelector = state =>
+export const solutionsRefreshingSelector = (state) =>
   state.app.solutions.isSolutionsRefreshing;
-export const solutionServicesSelector = state => state.app.solutions.services;
+export const solutionServicesSelector = (state) => state.app.solutions.services;
 
 // Sagas
 export function* fetchEnvironments() {
-  const jobs = yield select(state => state.app.salt.jobs);
+  const jobs = yield select((state) => state.app.salt.jobs);
   const preparingEnvs = jobs?.filter(
-    job => job.type === 'prepare-env/' && !job.completed,
+    (job) => job.type === 'prepare-env/' && !job.completed,
   );
   const environments = yield call(SolutionsApi.listEnvironments);
   const updatedEnvironments = yield call(updateEnvironments, environments);
@@ -132,9 +132,9 @@ export function* createEnvironment(action) {
 export function* prepareEnvironment(action) {
   const { envName, solName, solVersion } = action.payload;
 
-  const existingEnv = yield select(state => state.app.solutions.environments);
+  const existingEnv = yield select((state) => state.app.solutions.environments);
 
-  const preparingEnv = existingEnv.find(env => env.name === envName);
+  const preparingEnv = existingEnv.find((env) => env.name === envName);
 
   if (preparingEnv === undefined) {
     console.error(`Environment '${envName}' does not exist`);
@@ -151,7 +151,7 @@ export function* prepareEnvironment(action) {
 
   if (!addSolutionToEnvironmentResult.error) {
     const clusterVersion = yield select(
-      state => state.app.nodes.clusterVersion,
+      (state) => state.app.nodes.clusterVersion,
     );
     const result = yield call(
       SaltApi.prepareEnvironment,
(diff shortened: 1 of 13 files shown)


thomasdanan (commented):
how and when do you choose to which version you want to upgrade or downgrade?

ChengYanJin:
> how and when do you choose to which version you want to upgrade or downgrade?

When you click on the upgrade/downgrade button, it will pop up a modal with the available version.
(I'll update a screenshot for it)

thomasdanan (dismissed):

thomasdanan:
also don't hesitate to add @xaf-scality @bkettler to review the PR

gdemonet:
I'm a bit worried with the "Upgrade" / "Downgrade" buttons:

- there's no real difference between the two, so not sure there's much value in having two buttons, a single one with "Edit" label could be sufficient (could also allow for more editions than just the version, in the future)
- it's not going to upgrade or downgrade the components deployed, only the Operator and UI - making this information obvious in our UI would make sense, to avoid unnecessary confusions (not sure how though)

ui/src/ducks/app/solutions.js
```diff
@@ -204,16 +204,41 @@ export function* updateEnvironments(environments) {
           env.name,
         );
         if (!solutionOperatorDeployment.error && !solutionUIDeployment.error) {
+          const currentSolutionVersion = envConfig?.[solution];
+          // when update the environment, we add `availableUpgradeVersion` and `availableDowngradeVersion` field
```
gdemonet: We shouldn't do this kind of thing, it's duplicating existing data we have in the state. We should just make sure whenever it's consumed that we know how to compute these lists from a selected version and all the available ones (that we have stored in the state).

So, the code you wrote here should either go directly in the presentation component, or as a utility method. But the computation isn't expensive, so there's no need to cache the result in Redux.

ui/src/ducks/app/solutions.js
```diff
@@ -203,9 +203,22 @@ export function* updateEnvironments(environments) {
           `${solution}${UI_}`,
           env.name,
         );
+
         if (!solutionOperatorDeployment.error && !solutionUIDeployment.error) {
-          const currentSolutionVersion = envConfig?.[solution];
+          // in order to get the deployed version, we temporarily check the image version deployed Operator & UI.
+          // but we should add a label with the solution version in both Operator & UI.
```
gdemonet: Labels were added in #2367 ;)

ui/src/ducks/app/solutions.js
```diff
@@ -217,26 +230,42 @@ export function* updateEnvironments(environments) {
           const availableDowngradeVersion = [];
 
           availableSolutionVersions.versions.forEach(avaSolVer => {
-            if (avaSolVer.version.localeCompare(currentSolutionVersion) <= -1) {
+            if (
+              avaSolVer.version.localeCompare(
+                isDeploySolutionSucceed
+                  ? currentSolutionVersion
+                  : solutionOperatorDeploymentVersion,
```
gdemonet: No need for this check, both branches always yield the `solutionOperatorDeploymentVersion` anyway.

What would like to show differently if the configured version is different from the deployed one? For me, the only  thing to show would be that I can only see a "apply" button if there's a difference between `current` and `observed` versions.

gdemonet (changes requested):

ui/src/ducks/app/solutions.js
ChengYanJin: Hello @gdemonet,

So what you mean is to store `availableUpgradeVersion` and `availableDowngradeVersion` data in the local component state instead of caching it in redux store?

ui/src/ducks/app/solutions.js
ChengYanJin: Yes, you are right! We should go for Operator Version anyways because it's the actual deployed version.

ui/src/ducks/app/solutions.js
gdemonet: Yeah, that's the idea. A single row can compute whatever it needs from its props and the list of available solutions in the global state :)

ui/src/services/k8s/solutions.js
```diff
@@ -3,7 +3,7 @@ import * as core from './core';
 
 const _K8S = 'app.kubernetes.io';
 const _METAL = 'solutions.metalk8s.scality.com';
-
+const LABEL_VERSION = 'app.kubernetes.io/version';
```
gdemonet: super minor:

```suggestion
const LABEL_K8S_VERSION = `${_K8S}/version`;
```

gdemonet (dismissed):
LGTM!

gdemonet (approved):

Issue #2335, opened 2020-03-27
Create testrail object during CI build


TeddyAndrieux:
**Component**:

'build'

**Why this is needed**:

To upload test result to testrail we first need to create testrail objects

**What should be done**:

Create a YAML file describing all the TestRail objects we want to create and during post-merge, run a script to create all these testrail objects
(Only do this in post-merge as we do not want to upload all pre-merge result in testRail as IMO it's not relevant since it's during development process)

PR #2354, opened 2020-04-02
Add example solution build and patch version upgrade downgrade test to TestRail

TeddyAndrieux:
**Component**:

'build', 'tests'

**Context**: 

All post-merge test should be part of TestRail

**Summary**:

Add TestRail objects creation for example solution build and patch version upgrade and downgrade

---

Sees: #2335

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -648,7 +648,7 @@ stages:
     _metalk8s_internal_info:
       junit_info: &_build_ex-solution_junit_info
         TEST_SUITE: build
-        TEST_NAME: example-solution
+        TEST_NAME: example solution
     worker:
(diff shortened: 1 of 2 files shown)


gdemonet (approved):
LGTM. Wondering, would it be a problem if pre-merge also uploads results to TestRail?

TeddyAndrieux:
> LGTM. Wondering, would it be a problem if pre-merge also uploads results to TestRail?

I think yes, because you upload tones of result and some that will always fail because you are in development process and btw not sure we can configure alerting only for post-merge result ... anyway if we have all pre-merge I thin all global information in TestRail will not really be relevant

PR #2367, opened 2020-04-06
Add default labels to solution operator deployment & pod

### LINKED ISSUE #2366: Solutions: Add the current Solution version in Operator & UI deployment

**Why this is needed**:
To be able to check the current version of Solution Operator & UI

**What should be done**:
Add a label (`app.kubernetes.io/version` ?) on both deployment with the Solution version as the value

**Test plan**:

alexandre-allard:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2366

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -139,6 +139,9 @@ def _object_manipulation_function(action):
             manifest.setdefault('metadata', {}).setdefault('labels', {})[
                 'metalk8s.scality.com/version'
             ] = match.group('version') if match else "unknown"
+            manifest['metadata']['labels']['app.kubernetes.io/managed-by'] = \
+                'salt'
+            manifest['metadata']['labels']['heritage'] = 'salt'
 
         log.debug(
             '%sing object with manifest: %s',
--- a/salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml
+++ b/salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml
@@ -3,15 +3,29 @@ kind: Deployment
 metadata:
   name: {{ solution }}-operator
   namespace: {{ namespace }}
+  labels:
+    app: {{ solution }}-operator
+    app.kubernetes.io/name: {{ solution }}-operator
+    app.kubernetes.io/instance: {{ solution }}-operator
+    app.kubernetes.io/version: {{ version }}
+    app.kubernetes.io/component: operator
+    app.kubernetes.io/part-of: {{ solution }}
 spec:
   replicas: 1
   selector:
     matchLabels:
-      name: {{ solution }}-operator
+      app: {{ solution }}-operator
   template:
     metadata:
       labels:
-        name: {{ solution }}-operator
+        app: {{ solution }}-operator
+        app.kubernetes.io/name: {{ solution }}-operator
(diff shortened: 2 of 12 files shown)


salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml
```diff
@@ -3,15 +3,29 @@ kind: Deployment
...
+    app.kubernetes.io/version: {{ version }}
+    app.kubernetes.io/component: operator
+    app.kubernetes.io/part-of: {{ solution }}
+    app.kubernetes.io/managed-by: salt
 spec:
   replicas: 1
   selector:
     matchLabels:
-      name: {{ solution }}-operator
+      app: {{ solution }}-operator
```
Ebaneck, line 17: I think this change here will break the Solutions test.
It is related to this line and must be fixed
`        Then we have 1 running pod labeled 'name=example-solution-operator' in namespace 'example-environment'
`

salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml
```diff
@@ -38,4 +52,4 @@ spec:
       volumes:
         - name: operator-config
           configMap:
-            name: {{ solution }}-operator-config
\ No newline at end of file
+            name: {{ solution }}-operator-config
```
Ebaneck: What changed here??

Ebaneck (dismissed):

salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml
alexandre-allard: New line at the end of the file

gdemonet (changes requested):
Could you also add `heritage` (equivalent of `managed-by`, often used by Helm)? Basically, make sure that both Operator and UI deployments use the same labelling principles (not just the *Deployment*s, all objects can/should be labelled).

It could be a feature of our `metalk8s_kubernetes` module to enforce those labels are always set, like it does for automatically adding the `metalk8s.scality.com/version` label... WDYT @TeddyAndrieux @NicolasT?

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -139,6 +139,9 @@ def method(manifest=None, name=None, kind=None, apiVersion=None,
             manifest.setdefault('metadata', {}).setdefault('labels', {})[
                 'metalk8s.scality.com/version'
             ] = match.group('version') if match else "unknown"
+            manifest['metadata']['labels']['app.kubernetes.io/managed-by'] = \
+                'salt'
+            manifest['metadata']['labels']['heritage'] = 'salt'
```
gdemonet, line 144: :+1:

gdemonet (approved):
LGTM!

salt/_modules/metalk8s_kubernetes.py
TeddyAndrieux: Right now for things like coredns we put "metalk8s" as heritage

Edit: And for ingress we use "metalk8s" as managed-by also .... anyway we should have the same thing everywhere
Heritage = metalk8s
Managed-by = salt ?

TeddyAndrieux: Also I think we shouldn't override these value if they are already in the manifest (and indeed it's the same for version just above)

Issue #2259, opened 2020-02-26
No seccomp support


samcv:
**Component**:

**What happened**:
When attempting to use Kubernetes dashboard, I get the error: `feb 26 08:08:21 worker-3 kubelet[1500]: E0226 08:08:21.242021    1500 remote_runtime.go:105] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to generate seccomp spec opts: seccomp is not supported` which shows up in my journal

**What was expected**:
For the pod to start.
**Steps to reproduce**
kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/v2.0.0-rc5/aio/deploy/recommended.yaml
**Resolution proposal** (optional):
Support seccomp. I am not sure but I think containerd needs to be compiled with seccomp support (just guessing)

NicolasT:
Indeed, the `containerd` package as distributed in EPEL (which is the one we install) is not compiled with `seccomp` support. Adding this would require us to ship our own build of `containerd`, which would require quite a bit of work.

Meanwhile, you could remove the `seccomp`-related annotation from the dashboard *Deployment* *PodSpec* and get it to run.

samcv:
Would it be appropriate to get them to include it? Or is there a reason 
you think it is not included?

NicolasT:
> Would it be appropriate to get them to include it? Or is there a reason you think it is not included?

I never tried, to be honest. One reason why it's not included could be if the version of `libseccomp` that ships with RHEL7 is too old.

Anyway, the version of `containerd` as shipped in EPEL is somewhat old as well, so building our own could make sense. We'll need to look into it: given the age of RHEL7, packaging Go applications is... _not trivial_.

NicolasT:
I built a `containerd` 1.2.13 RPM with `seccomp` enabled, and after deploying it (manually) to one node in my test-cluster, I can deploy the upstream dashboard manifest (slightly changing the manifest to ensure the `dashboard-metrics-server` gets scheduled on this one node) and everything seems to work.

I'll spend some time to integrate this in the project.

PR #2369, opened 2020-04-06
Upgrade containerd

NicolasT:
Upgrade `containerd` to 1.2.13, among others to include a fix for https://github.com/containerd/containerd/issues/3125. Also, enable `seccomp` support to handle #2259.

This implies we no longer ship the `containerd` package as found in EPEL, but instead build one as part of the MetalK8s build and ship it in the `metalk8s-scality` repository.


NicolasT:
For reviewers, the most important changes are

- Bumping the `containerd` version to 1.2.13, based on the RPM spec of 1.2.4 as we currently ship from EPEL (imported separately) at https://github.com/scality/metalk8s/pull/2369/commits/f5869403a1c5fcd1c5701f5c011494cd726c0329#diff-bddba0a782eb642c0a25329157bdf322
- Enabling `seccomp` support in `containerd` at https://github.com/scality/metalk8s/pull/2369/commits/05ea261e7916dd466e2fbd6110be96ace735fb7e#diff-bddba0a782eb642c0a25329157bdf322

NicolasT:
@bert-e approve

NicolasT:
Moved into 2.6.0. Needs field testing, too big a change for a .1.

tests/post/steps/conftest.py
```diff
@@ -46,11 +46,17 @@ def _wait_for_status():
...
+
+
 # Given {{{
 
 
-@given(parsers.parse("pods with label '{label}' are '{expected_status}'"))
-def check_pod_status(request, host, k8s_client, label, expected_status):
+@given(_PARSE_PODS_WITH_LABEL_STATUS)
+def given_check_pod_status(request, host, k8s_client, label, expected_status):
     ssh_config = request.config.getoption('--ssh-config')
```
gdemonet, line 60: I know it's not related to your changes, but there's a `ssh_config` fixture to simplify this a bit:

```suggestion
def given_check_pod_status(ssh_config, host, k8s_client, label, expected_status):
```

tests/post/steps/test_seccomp.py
```diff
@@ -0,0 +1,59 @@
...
+
+@when("we create a utils Pod with labels {'test': 'seccomp1'} "
+      "and annotations "
+      "{'seccomp.security.alpha.kubernetes.io/pod': 'runtime/default'}")
+def create_utils_pod(utils_pod):
+    pass
+
+
+@pytest.fixture
+def utils_pod(k8s_client, utils_image):
```
gdemonet, line 28: You don't need to create another fixture, `@when` already makes the `create_utils_pod` a fixture. This would allow us to make use of a BDD parser:

```python
_UTILS_POD_PARSER = parsers.parse(
    "we create a utils Pod with labels {labels:JSON} "
    "and annotations {annotations:JSON}",
    extra_types=dict(JSON=json.loads),
)

@when(_UTILS_POD_PARSER)
def utils_pod(k8s_client, utils_image, labels, annotations):
    # create the Pod with right labels and annotations, which are
    # passed in as Python dicts already
    try:
        yield pod_name
    finally:
        # clean-up of the Pod

@then("I can run '{command}' in the utils Pod")
def run_in_utils_pod(utils_pod, command):
     # If we had such a helper, we could do:
    kubectl_exec(pod_name=utils_pod, command=command)
```

gdemonet (commented):
LGTM for the most part! Some suggested changes on tests, maybe we want to move them in a debt ticket, as you wish.

tests/post/steps/test_seccomp.py
NicolasT: Well, that's what I tried at first (how could one perform resource cleanup otherwise?). However, when executing the test using that implementation, nothing happened. My `when` thing, which is then a generator, would not be `next`ed, so no code executed, no Pod created,... (to my surprise). Skimming the docs back then, there's no reference of `when`s being permitted to be a generator.

I'll try again with a super-simple test...

NicolasT: As discussed on Slack, one can't use a generator in a `@when` implementation.

tests/post/steps/conftest.py
NicolasT: #2387

tests/post/steps/test_seccomp.py
gdemonet: Indeed, but it seems we can in a `@given` (and in your context, creating a utils Pod could make sense as a "Given" step).

NicolasT: Well, no :) I am not actually *using* the `utils` Pod, its merely a canary (hence, not executing any commands in it). What this test wants to validate is that *any* Pod with some specific set of annotations will be fully functional at some point (i.e., our CRI supports the `seccomp` profile annotations). The Pod is not an input to the test, its creation and validation *is* the test.

gdemonet (approved):

Issue #2352, opened 2020-04-02
Adding a new user should not require to fetch and deploy new packages


thomasdanan:
**Component**:

'bootstrap', 'deployment', 'authentification'

**Why this is needed**:

If the Platform administrator wants to provision a new admin user, following documentation: https://metal-k8s.readthedocs.io/en/development-2.6/operation/cluster_and_service_configuration.html, It may be needed to deploy `apache-utils` package, which may not be possible if the site is offline.

**What should be done**:

It should not be required to fetch and deploy packages in order to provision a new user. I guess we should either embed and deploy apache-utils during bootstrap deployment, either use some other available tooling

PR #2374, opened 2020-04-06
images: enrich `metalk8s-utils` image

### LINKED ISSUE #2156: Extend `metalk8s-utils` image content
If thanks to #2146 the `metalk8s-utils` image starts to get used more often, we must ensure it contains all tools people use to debug/troubleshoot a system. We currently only include a couple:

- `bind-utils`
- `curl`
- `httpie`
- `iperf`
- `iproute`
- `socat`
- `telnet`

I bet there's a bunch of other useful tools we could/should add. This ticket is meant to collect those, so please comment.

NicolasT:
This patch extends the `metalk8s-utils` `Dockerfile` to add a bunch of
packages (and make some related changes):

- The Kubernetes and SaltStack repositories are enabled in the image
- No longer override the entrypoint
- Fail when a given package (name) doesn't exist
- Add the following packages to the image:

  * `bash-completion`
  * `bash-completion-extras`
  * `bzip2`
  * `conntrack-tools`
  * `cri-tools`, for `crictl`
  * `e2fsprogs`
  * `ebtables`
  * `etcd`, for `etcdctl`
  * `ethtool`
  * `gdb`
  * `git`
  * `htop`
  * `httpd-tools`, for `htpasswd` (see #2352)
  * `iotop`
  * `iperf3` (next to `iperf` which was already installed)
  * `ipset`
  * `iptables`
  * `ipvsadm`
  * `jnettop`
  * `jq`
  * `kubectl` (pinned to the version of Kubernetes this MetalK8s release ships with)
  * `less`
  * `lsof`
  * `ltrace`
  * `lvm2`
  * `net-tools`
  * `nethogs`
  * `nmap`
  * `nmap-ncat`, for `nc`
  * `openssh-clients`
  * `openssh-server`
  * `openssl`, for its CLI interface
  * `parted`
  * `perf`
  * `rsync`
  * `salt-master`, for `salt` and `salt-run`
  * `salt-minion`, for `salt-call`
  * `strace`
  * `sysstat`
  * `tcpdump`
  * `util-linux`, for `nsenter` and others
  * `vim`
  * `wget`
  * `wireshark`, for `tshark`
  * `xfsprogs`

This resolves #2156 to a great extent. However, documentation on how to
use the image is still lacking, as well as some of the tools referred to
in the ticket, for which no upstream packages are available.

See: https://github.com/scality/metalk8s/issues/2156
See: https://github.com/scality/metalk8s/issues/2352

### DIFF
--- a/images/metalk8s-utils/configure-repos.sh
+++ b/images/metalk8s-utils/configure-repos.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -xue -o pipefail
+
+CENTOS_VERSION=$1
+SALT_VERSION=$2
+
+cat > /etc/yum.repos.d/kubernetes.repo << EOF
+[kubernetes]
+name=Kubernetes
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el$CENTOS_VERSION-x86_64
+enabled=1
+gpgcheck=1
+repo_gpgcheck=1
+gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg
+       https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
+EOF
+
+cat > /etc/yum.repos.d/saltstack.repo << EOF
+[saltstack]
+name=SaltStack repo for RHEL/CentOS \$releasever
+baseurl=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/$SALT_VERSION
+enabled=1
+gpgcheck=1
+gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/$SALT_VERSION/SALTSTACK-GPG-KEY.pub
+       https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/$SALT_VERSION/base/RPM-GPG-KEY-CentOS-$CENTOS_VERSION
+EOF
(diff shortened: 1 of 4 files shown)


NicolasT:
@bert-e approve

images/metalk8s-utils/configure-repos.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -xue -o pipefail
+
+CENTOS_VERSION=$1
+SALT_VERSION=$2
+
+cat > /etc/yum.repos.d/kubernetes.repo << EOF
+[kubernetes]
+name=Kubernetes
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el$CENTOS_VERSION-x86_64
```
gdemonet, line 11: Could have used the Yum vars, like `\$releasever` instead of this `CENTOS_VERSION` variable, no? And replaced `x86_64` with `\$basearch` maybe :)

gdemonet (approved):

images/metalk8s-utils/configure-repos.sh
NicolasT: Hmh, yes, I guess... The 'source' (in `packages/redhat/repos/...`) from where I took this snippet doesn't use those either, I did add the `$CENTOS_VERSION`. But indeed, the 'default' variables would've been better. But merged now :man_shrugging: and not sure it's worth a follow-up :man_shrugging:

gdemonet: Yeah not critical at all :ok_hand:

Issue #2371, opened 2020-04-06
Add build url in JUnit file generated by eve


TeddyAndrieux:
**Component**:

'build'

**Why this is needed**:

Currently we didn't have any information in TestRail to investigate post-merge result

**What should be done**:

Add a link to buildbot result to have all information about the build stage

---

~Blocked:~

~This ticket first need one ticket to be done on eve side~

PR #2376, opened 2020-04-07
eve: Add branch and commit revision to `text` in JUnit

TeddyAndrieux:
**Component**:

'build'

**Context**: 

Currently we use JUnit file to upload test result to TestRail but we
only have artifacts URL to investigate in case of test fail.

**Summary**:

Add branch name and commit revision in the JUnit to upload this to
TestRail, this is a second step waiting for #2371

---

Sees: #2371

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -325,6 +325,8 @@ models:
         FINAL_STATUS: ''
         TEXT: |-
           Artifact Url: %(prop:artifacts_public_url)s
+          Branch: %(prop:branch)s
+          Commit: %(prop:revision)s
       haltOnFailure: True
   - ShellCommand: &add_final_status_artifact_success
       <<: *add_final_status_artifact


eve/main.yml
```diff
@@ -325,6 +325,8 @@ models:
         FINAL_STATUS: ''
         TEXT: |-
           Artifact Url: %(prop:artifacts_public_url)s
+          Branch: %(prop:branch)s
+          Commit: %(prop:revision)s
```
gdemonet, line 329: Cool ! Could we also add a link to the build?

```
Build URL: https://eve.devsca.com/github/scality/metalk8s/#/builders/1/builds/%(prop:buildnumber)s
```

TeddyAndrieux: That do not work the `1` builder is not always 1 and this information is currently not in eve property a ticket is in progress about this on there side see #2371, once this is done I will add it

gdemonet (approved):

PR #2378, opened 2020-04-07
Enable Login for Grafana with Dex(OIDC)

Ebaneck:
**Component**:

'charts', 'kubernetes', 'oidc', 'authentication'

**Context**: 

Basic authentication was deprecated in 2.5 branch and Dex(OIDC) was introduced. It is only fair that we make use of the Dex user store as a source of truth for Grafana authentication before a 2.5 release.

Why?
- Users will not lose Grafana username/password customizations when they henceforth upgrade from 2.5 to 2.6
- We can gracefully update the documentation to reflect a common procedure for changing username and passwords for both Grafana UI and MetalK8s UI instead of using two different documented procedure.

Things to keep:

- Continue to warn users of the possible loss of Grafana username/password customizations when they upgrade from 2.4 to 2.5 as it is already the case in the documentation

**Summary**:

This PR adds the following changes:
- Adds and enable Single Sign-On for Dex-Grafana as the default means of authentication into Grafana

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/charts/prometheus-operator.yaml
+++ b/charts/prometheus-operator.yaml
@@ -148,10 +148,21 @@ grafana:
 
   grafana.ini:
     server:
-      root_url: /grafana
+      root_url: '__url__(https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana)'
     analytics:
       reporting_enabled: false
       check_for_updates: false
+    auth:
+      oauth_auto_login: true
+    auth.generic_oauth:
+      enabled: true
+      tls_skip_verify_insecure: true
+      scopes: "openid profile email groups"
+      client_id: "grafana-ui"
+      client_secret: "4lqK98NcsWG5qBRHJUqYM1"
+      auth_url: '__url__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/auth)'
+      token_url:  '__url__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/token)'
+      api_url: '__url__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/userinfo)'
 
   testFramework:
     enabled: false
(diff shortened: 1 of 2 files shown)


charts/prometheus-operator.yaml
```diff
@@ -148,10 +148,21 @@ grafana:
 
   grafana.ini:
     server:
-      root_url: /grafana
+      root_url: '__url__(https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana)'
```
gdemonet, line 151: What's this `__url__` thing doing?

gdemonet: OK I looked at the code. The name is bad (we're expecting something special about URLs), `__escape__` could be more explicit.

Ebaneck: It's just one of our magical ways of substituting unquoted URLs from jinja to YAML.

See: https://github.com/scality/metalk8s/blob/development/2.6/charts/render.py#L144

Ebaneck: > OK I looked at the code. The name is bad (we're expecting something special about URLs), `__escape__` could be more explicit.

Ok, but wouldn't fix it in this PR since the Prometheus-operator charts make use of this entry

I will fix this probably next week when I am about to bump the versions for Dex and Prometheus-operator chart

charts/prometheus-operator.yaml
```diff
@@ -148,10 +148,21 @@ grafana:
...
     analytics:
       reporting_enabled: false
       check_for_updates: false
+    auth:
+      oauth_auto_login: true
+    auth.generic_oauth:
+      enabled: true
+      tls_skip_verify_insecure: true
```
gdemonet, line 159: Mmmh... Really? Can't we give it the Ingress cert?

gdemonet (changes requested):

gdemonet (approved):

Issue #2372, opened 2020-04-06
Design multiple versions support for Solution CRDs


gdemonet:
**Component**: solutions, salt, kubernetes

**Why this is needed**:

Solution Operators most often will rely on CustomResourceDefinitions to extend Kubernetes API and allow users to manage the application(s) this Solution provides.
When the requirements of the Operator (or the application) change over time, the need for new versions of said CRDs will inevitably arise.

As the management of multiple versions isn't an obvious task, it cannot be left to a user.
And since CRDs are cluster-scoped, while Solution Operators are namespaced-scoped (and shouldn't create their own CRDs), these cannot be managed by the Solution itself.

**What should be done**:

MetalK8s should extend its tooling to provide first-class support for multi-version CRDs, including:

- bundling the single-version CRDs provided in Solution ISOs into a single multi-version CRD
- registering (and generating the required certs) for conversion webhooks
- changing the `stored` version in a live cluster, and converting stored objects

For reference, see:

- [`kubebuilder`'s tutorial on multi-versions](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html)
- [Kubernetes documentation on CRD versioning](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/)

**Implementation proposal** (strongly recommended):

TBD

**Test plan**:

The example Solution should provide two API versions, and its buildchain should generate two different ISOs for testing the upgrade / downgrade path.
Special care should be taken for manipulation of existing objects and running Operators.

---

Epic: #1994

PR #2391, opened 2020-04-08
Build a second example-solution ISO for upgrade/downgrade tests

### LINKED ISSUE #2390: Build another version of example-solution for upgrade/downgrade tests
**Component**: ci, examples

**Why this is needed**:
To be able to tests upgrade / downgrade of Solutions

**What should be done**:
Add a mechanism in the CI to build 2 version of the example solution, let say a `0.1.0-dev` and a `0.2.0-dev`

Build a first version, then patch the VERSION file, clean up a bit the old images and then build a second ISO

**Test plan**: n/a

alexandre-allard:
**Component**: ci

**Context**:
In order to be able to tests upgrade/downgrade of Solutions, we need to have at least 2 different version of the example-solution.

**Summary**:
Build a second example-solution ISO with the patch version bumped by 1.

**Acceptance criteria**:
- [x] A second archive is well generated and uploaded to artifacts.
- [x] The import of the Solution archive works.

---

Closes: #2390

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -704,19 +704,36 @@ stages:
           command: >
             . examples/metalk8s-solution-example/VERSION &&
             echo "$VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH$VERSION_SUFFIX"
-      - ShellCommand:
+      - ShellCommand: &build_example_solution_iso
           name: Build Example Solution ISO
           command: >
             cd examples/metalk8s-solution-example &&
-            DOCKER_SOCKET=http://localhost:2375 make iso
+            DOCKER_SOCKET=http://localhost:2375 make iso &&
+            rm -rf examples/metalk8s-solution-example/_build/images
+            examples/metalk8s-solution-example/_build/root/images
           haltOnFailure: True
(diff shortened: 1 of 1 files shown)


gdemonet (approved):
OK for now, let's create an improvement issue to track following suggestions:
- would be better if the example Solution's `Makefile` would provide the two versions instead of those hacky `sed`s
- would be nice if there were some real differences between the versions (like, there could be a CRD version change... see #2372)
- building those two versions shouldn't require a clean of the temporary `_build` directory (instead, isolate versioned components in there, or only archive some specific files and not the whole `images` directory)

alexandre-allard:
> OK for now, let's create an improvement issue to track following suggestions:
> 
> * would be better if the example Solution's `Makefile` would provide the two versions instead of those hacky `sed`s
> * would be nice if there were some real differences between the versions (like, there could be a CRD version change... see #2372)
> * building those two versions shouldn't require a clean of the temporary `_build` directory (instead, isolate versioned components in there, or only archive some specific files and not the whole `images` directory)

@gdemonet I agree, tracked this here #2394

PR #2399, opened 2020-04-10
docs: Add design choices for Centralized CLI

TeddyAndrieux:
**Component**:

'docs'

**Summary**:

Add documentation describing all the design choices and rejected choice
for Centralized CLI, also define the different iteration for development
of this feature

---

### DIFF
--- a/docs/developer/architecture/centralized-cli.rst
+++ b/docs/developer/architecture/centralized-cli.rst
@@ -106,7 +106,11 @@ Cluster Administration
 +============+============+===============================================+
 | bootstrap  | deploy     |                                               |
 +------------+------------+-----------------------------------------------+
-| bootstrap  | import-iso | path_to_iso                                   |
+| archive    | import     | path_to_iso                                   |
++------------+------------+-----------------------------------------------+
+| archive    | get        | <name>                                        |
++------------+------------+-----------------------------------------------+
+| archive    | delete     | path_to_iso or path_to_mountpoint or name     |
 +------------+------------+-----------------------------------------------+
 | cluster    | upgrade    | dest-version, <dry-run>                       |
 +------------+------------+-----------------------------------------------+
@@ -122,13 +126,40 @@ Cluster Administration
 Solution Administration
 ^^^^^^^^^^^^^^^^^^^^^^^
 
+.. note::
+
+  Import and unimport of solution are done the same way as MetalK8s ``archive``
+  using ``metalk8sctl archive import ...``.
+
 **tool: metalk8sctl**
 
-+------------+------------+-----------------------------------------------+
-| Resource   | action     | parameters                                    |
-+============+============+===============================================+
-| solutions  | import-iso | path_to_iso                                   |
-+------------+------------+-----------------------------------------------+
++----------+------------+-------------------+
+| Resource | action     | parameters        |
++==========+============+===================+
+| solution | activate   | name, version     |
++----------+------------+-------------------+
+| solution | deactivate | name              |
++----------+------------+-------------------+
+| solution | get        | <name>, <version> |
++----------+------------+-------------------+
+
+**tool: kubectl metalk8s**
+
++----------+---------------+------------------------------------------+
+| action   | resource type | parameters                               |
++==========+===============+==========================================+
+| create   | environment   | name, <description>, <namespace>         |
++----------+---------------+------------------------------------------+
+| delete   | environment   | name, <namespace>                        |
++----------+---------------+------------------------------------------+
+| get      | environment   | <name>                                   |
++----------+---------------+------------------------------------------+
+| add      | solution      | name, version, environment, <namespace>  |
++----------+---------------+------------------------------------------+
+| delete   | solution      | name, environment, <namespace>           |
++----------+---------------+------------------------------------------+
+| get      | solution      | <name>, environment                      |
++----------+---------------+------------------------------------------+
 
 Cluster Service Administration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -149,3 +180,239 @@ Cluster Service Administration
 +----------+---------------+-------------+----------------------------+
 | edit     | dex-config    | name        | open an editor             |
 +----------+---------------+-------------+----------------------------+
+
+Design Choices
+--------------
+
+Two distinct CLI:
+
+- a ``metalk8s`` kubectl plugin with subcommands to interact with Kubernetes
+  API, and Salt API if needed.
+- a ``metalk8sctl`` CLI with subcommands for action that need to interact with
+  the local machine, but may also interact with Kubernetes API and Salt API if
+  needed.
+
+
+``metalk8s`` kubectl plugin
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Language
+""""""""
+
+``Go`` is chosen as the language for kubectl plugin for the following reasons:
+
+- Great interactions with Kubernetes API.
+- Often used for operators and kubectl plugins
+  (`Sample CLI plugin <https://github.com/kubernetes/sample-cli-plugin>`_,
+  `Helpers for kubectl plugins <https://github.com/kubernetes/cli-runtime>`_).
+- Easy to ship because it's a statically compiled binary, no deps to provide.
+- Simple deployment (no real requirements), just drop a binary in the PATH.
+
+.. todo::
+
+  - As kubectl plugin may run from outside of the cluster, think about
+    compilation of binaries for non-"classic linux x86_64"
+
+Input and Output
+""""""""""""""""
+
+Each command should follow the kubectl style and standard as much as possible:
+
+- Command style:
+
+  .. code-block:: bash
+
+    kubectl metalk8s <action> <resource>
+
+- Interactive:
+
+  No interaction with the user, except when it's an ``edit`` command an editor
+  is opened (if needed) and when a password is required a prompt appears to ask
+  it.
+
+- Output style:
+
+  Default human-readable output (``<object> <action>ed``).
+
+  A ``--output``, ``-o`` option to change output format, at least support
+  for ``json`` and ``yaml`` (``jsonpath`` and ``go-template`` when it's
+  possible).
+
+  Internally each action result should be an "object"
+  (e.g.: single-level dictionary) containing several informations,
+  at least:
+
+  - name
+  - message
+  - result (True or False)
+  - an elapsed time (to know each action time)
+
+  A ``--verbose``, ``-v`` option to change log level verbosity (default output
+  to ``stderr``), using
+  `Kubernetes log library <https://github.com/kubernetes/klog>`_.
+
+  By default each command will wait for a result but, when it's possible, a
+  ``--async`` option should allow to do not wait for a result and just output
+  an ID (e.g.: Job ID for Salt) that can be used to watch for the result.
+
+SaltAPI interaction
+"""""""""""""""""""
+
+If the plugin needs to access Salt API then it should use the service proxy
+``http://<apiserver_host>/api/v1/namespaces/kube-system/services/https:salt-master:api/proxy/``.
+
+For each and every Salt API call plugin will need authentication on
+apiserver to access the Salt API service proxy and also to Salt API.
+
+.. note::
+
+  Right now, Salt API only accepts authentication using Bearer token, but in
+  kubeconfig we could have certificates authentication so this kind of
+  kubeconfig will not work with this kubectl plugin.
+
+  Add support for certificates based authentication in Salt API look quite hard
+  and costly.
+
+.. todo::
+
+  Check what can be done to support certificates based kubeconfig or if we
+  consider that we only support token-based kubeconfig for our plugin commands
+  interacting with Salt API.
+
+Deployment
+""""""""""
+
+Plugin should be developed as one single binary ``kubectl-metalk8s`` available
+from the ISO, easy buildable from GitHub repository and also as a System
+Package for Operating System supported by MetalK8s.
+
+The package should install the plugin in ``/usr/bin`` directory by default.
+
+This package should be installed on the bootstrap node by default.
+
+Rejected design choice
+""""""""""""""""""""""
+
+- ``Bash`` kubectl plugin: Bash is great to do simple actions but not when
+  you need to do interaction with some API like Kubernetes API or Salt API.
+- ``Python`` kubectl plugin: Python allows us to do complicated actions and
+  great interactions with APIs but interactions between ``Go`` and Kubernetes
+  are much easier, given the large number of example available.
+
+``metalk8sctl`` CLI
+^^^^^^^^^^^^^^^^^^^
+
+.. _`Salt Python client API`: https://docs.saltstack.com/en/latest/ref/clients/index.html
+
+Language
+""""""""
+
+``Python`` is chosen as the language for ``metalk8sctl`` for the following
+reasons:
+
+- Ability to interact with `Salt Python client API`_.
+- Python installation needed anyway by Salt-minion.
+
+.. note::
+
+  Python version 3 will be used as version 2 is end of life since beginning of
+  2020.
+
+Input and Output
+""""""""""""""""
+
+- Command style:
+
+  .. code-block:: bash
+
+    metalk8sctl <resource> <action>
+
+- Interactive:
+
+  Never.
+
+- Output style:
+
+  Human readable output, do not necessarily need for "machine output" like
+  JSON and YAML.
+
+  The output should display useful information from Salt returns when needed,
+  and in case of error, only show relevant error message(s) from Salt.
+
+.. todo::
+
+  - Define logging informations
+
+Salt interaction
+""""""""""""""""
+
+All Salt interaction should be done using `Salt Python client API`_
+and not use the ``salt-call``, ``salt``, ``salt-run`` binary at all.
+
+This `Salt Python client API`_ allows us to interact with Salt-master directly
+from the host machine as Python API directly acts on the Salt sockets and does
+not need to execute a command inside the Salt-master container.
+
+Deployment
+""""""""""
+
+``metalk8sctl`` should be available from the ISO and also as a System
+Package for Operating System supported by MetalK8s.
+
+This package should be installed on the bootstrap node automatically after
+a fresh install.
+
+As this CLI is used to do the first bootstrap deployment we will need another
+script (likely ``bash``) to configure local repositories and install
+``metalk8sctl`` package with all dependencies.
+
+.. note::
+
+  This CLI cannot run from outside of the cluster and need to have root
+  access on the machine to run.
+
+  That's why this CLI do not need any specific authentication on the cluster
+  itself, interaction with all machines will be done using Salt.
+
+Rejected design choice
+""""""""""""""""""""""
+
+- ``bash`` MetalK8s CLI: Bash is great to do simple actions but not when you
+  need to do interaction with Salt, Salt API, and Kubernetes API.
+- Do not follow kubectl style for the command (``<action> <resource>``), it
+  does not make sense to regroup command per action as actions are really
+  different and this CLI only manages a few resources.
+
+Implementation Details
+----------------------
+
+Two different projects that can be started in parallel.
+
+First have a simple framework to implement a simple command,
+then each command would update the framework if needed.
+
+Check `Requirements`_ for a full list of commands.
+
+Documentation
+-------------
+
+All command should be documented in the Operational Guide with a reference to
+it when it's needed in the Installation Guide.
+
+All commands and sub-commands should have a ``--help`` option to explain a bit
+of usage of this specific command and available options.
+
+Test plan
+---------
+
+For ``metalk8sctl``:
+
+- Add unit tests for internal functions using Pytest
+- Most of the command are already used during functional test (some may need
+  to be added in PyTest BDD)
+
+For ``metalk8s`` kubectl plugin:
+
+- Add unit tests for internal functions using
+  `Golang testing framework <https://golang.org/pkg/testing/>`_
+- Add functional test for all plugin commands in PyTest BDD


docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+
+If the plugin needs to access Salt API a proxy should be configured by the
+plugin.
+
+And then plugin can access Salt API directly using the proxy
+``http://<server-ip>:<proxy-port>/api/v1/namespaces/kube-system/services/https:salt-master:api/proxy/``.
+
+This ``proxy-port`` should be configurable as it needs to be reachable by the
+plugin that can run from outside the MetalK8s cluster.
+
```
Ebaneck: Any chance we could consider something  like `pepperlib` from https://github.com/saltstack/pepper

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Language
+""""""""
+
+``Go`` is chosen as the language for kubectl plugin for the following reasons:
+
+- Great interactions with Kubernetes API.
+- Often used for operators and kubectl plugins
+  (`Sample CLI plugin <https://github.com/kubernetes/sample-cli-plugin>`_).
```
gdemonet: Maybe even mention the set of tools for building such plugins: https://github.com/kubernetes/cli-runtime

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+- Command style:
+
+  .. code-block:: bash
+
+    kubectl metalk8s <action> <resource>
+
+- Interactive:
+
+  No interaction with the user, except when it's an ``edit`` command an editor
+  is opened (if needed).
```
gdemonet: We may also need prompt/editor for secrets management (thinking of Dex static users, Alertmanager settings...)

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+  `Kubernetes log library <https://github.com/kubernetes/klog>`_.
+
+
+SaltAPI interaction
+"""""""""""""""""""
+
+If the plugin needs to access Salt API a proxy should be configured by the
+plugin.
+
+And then plugin can access Salt API directly using the proxy
```
gdemonet: What do you mean by "a proxy should be configured"? IIUC, this "proxy" is always available: that's what API Server exposes for all Service objects.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+``http://<server-ip>:<proxy-port>/api/v1/namespaces/kube-system/services/https:salt-master:api/proxy/``.
```
gdemonet: We should just be able to talk to the same K8s API as we are for other requests:

```suggestion
``http://<apiserver-host>/api/v1/namespaces/kube-system/services/https:salt-master:api/proxy/``.
```

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+  - message
+  - result (True or False)
+  - an elapsed time (to know each action time)
+
+  A ``--verbose``, ``-v`` option to change log level verbosity (default output
+  to ``stderr``), using
+  `Kubernetes log library <https://github.com/kubernetes/klog>`_.
+
+
+SaltAPI interaction
```
gdemonet, line 258: We will also need to somehow pass a Bearer token (that we probably already have since interacting with K8s API) to Salt API.
Maybe worth mentioning that we delegate the authz flow to Salt, which will use this Bearer token to check permissions against K8s/whatever.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+Two distinct CLI:
+
+- a ``metalk8s`` kubectl plugin with subcommands to interact with Kubernetes
+  API, and Salt API if needed.
+- a ``metalk8sctl`` CLI with subcommands for action that need to interact with
+  the local machine, but may also interact with Kubernetes API and Salt API if
+  needed.
+
+
+``metalk8s`` kubectl plugin
```
gdemonet, line 196: One thing we forgot to mention: `kubectl` authentication is based on kubeconfig files, with certificates, whereas our users will rely on an OIDC provider to authenticate, most of the time.

It would be nice to include in this design an approach for generating the required certs (maybe even the kubeconfig, I think @NicolasT  mentioned tools to do that) based on an OIDC-provided identity.
Getting a kubeconfig will be required from our users to use the CLI from outside the cluster (we don't want them to copy the `admin.conf`, let alone use it...).

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Nop, we do not need it, we will interact with API directly and we use Go
Anyway pepper is good to interact with salt-api directly as a CLI but it's not what we want here and even on the bootstrap we have access to the salt socket so we do not need pepper at all here.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+  possible).
+
+  Internally each action result should be an "object"
+  (e.g.: single-level dictionary) containing several informations,
+  at least:
+
+  - name
+  - message
+  - result (True or False)
+  - an elapsed time (to know each action time)
```
gdemonet, line 248: Something we discussed: some actions may be ran asynchronously (e.g. `deploy node --async`), outputting some form of reference to the active job. And we may have actions to "watch" jobs, like `kubectl metalk8s get job <jid> -w`. 

Not sure how that'll play with the output of these commands, just thought it'd be worth mentioning here.
Maybe even in the requirements... :man_shrugging:

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+Deployment
+""""""""""
+
+Plugin should be developed as one single binary ``kubectl-metalk8s`` available
+from the ISO, easy buildable from GitHub repository and also as a System
+Package for Operating System supported by MetalK8s.
+
+The package should install the plugin in ``/usr/bin`` directory by default.
+
+This plugin should be installed on the bootstrap node by default.
```
gdemonet: Maybe it's better to write more specifically:
```suggestion
This package should be installed on the bootstrap node by default.
```

Also, we'd probably like to install it in the `metalk8s-utils` image.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+
+All Salt interaction should be done using
+`Salt Python API <https://docs.saltstack.com/en/latest/ref/clients/index.html>`_
+and not use the ``salt-call``, ``salt``, ``salt-run`` binary at all.
+
+This Python API allows us to interact with Salt-master directly from the host
+machine as Python API directly actis on the Salt socket and does not need to
+execute a command inside the Salt-master container.
+
+Deployment
```
gdemonet, line 356: I think we also want to mention how it should be made available on **bare** CentOS 7 machines, so without many required dependencies, Python 3 being the most critical.

So, we'll need some form of entrypoint from the ISO that will "bootstrap" this CLI, and we may also need to package some dependencies to install along with it (some of which we install with `pip` in the Salt master container, IIRC...).

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+
+Salt interaction
+""""""""""""""""
+
+All Salt interaction should be done using
+`Salt Python API <https://docs.saltstack.com/en/latest/ref/clients/index.html>`_
+and not use the ``salt-call``, ``salt``, ``salt-run`` binary at all.
+
+This Python API allows us to interact with Salt-master directly from the host
+machine as Python API directly actis on the Salt socket and does not need to
```
gdemonet:
```suggestion
machine as Python API directly acts on the Salt sockets and does not need to
```

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+Iteration 1
+^^^^^^^^^^^
+
+``metalk8sctl``: (everything ?)
+
+- bootstrap deploy
+- bootstrap import-iso ?
+- cluster upgrade ?
+- cluster downgrade ?
+- etcd health
```
gdemonet: Good point, we didn't talk about `etcd` interactions - should we use Salt? IMO, that'd be the best approach, because then we'd have all the tools needed to implement self-healing in case of trouble with `etcd`.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+- bootstrap backup ?
+- bootstrap restore ?
+- solutions import-iso ?
```
gdemonet: Maybe the resource isn't really Solutions here... What would you think about an `archive` resource, and then:

```shell
# Both can (and will, anyway) look into the ISO, so the tool can
# know if it's a Solution or not, to configure it in the right place
metalk8sctl archive import /path/to/metalk8s.iso
metalk8sctl archive import /path/to/hyperdrive.iso

# Could also add some others, like
metalk8sctl archive list
metalk8sctl archive remove --iso /path/to/metalk8s.iso
metalk8sctl archive remove --id metalk8s-2.4.2
```

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +149,203 @@ Cluster Service Administration
...
+
+Iteration 2
+^^^^^^^^^^^
+
+``metalk8s`` kubectl plugin: (everything ?)
```
gdemonet: We don't have `get` or `describe` commands? Could be useful to show Salt-related info (like grains/pillar for a Node's minion), or adjust how some things are displayed.

I'd love to get some wrappers on Environments, which are just groups of namespaces with a common label:

```shell
kubectl metalk8s get env
NAME    NAMESPACES                  SOLUTIONS
prod    prod-storage, prod-zenko    storage(1.0.2), zenko(2.0.1)
test    test                        storage(1.0.2), zenko(2.1.0-rc1)

kubectl metalk8s get env prod -o json
{
  "name": "prod",
  "namespaces": ["prod-storage", "prod-zenko"],
  "solutions": [
    {
      "name": "storage",
      "version": "1.0.2",
      "operator": "prod-storage/storage-operator",
      "ui": "prod-storage/storage-admin-ui",
    },
    ...
  ]
}

# We could even provide wrappers to the usual get/delete/... commands
kubectl metalk8s get deploy --env prod
#   - This would return Deployments in both prod-storage and prod-zenko
```

gdemonet (changes requested):
Great start! We may be missing some details, and we'll surely find out about more problems once we start implementing, but it already looks good to me :)

If some areas are too unclear to decide right now, maybe just create follow-up tickets and we'll update the design accordingly once investigations are made.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: True, done

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: For secrets yes it's like a config map so it's an edit, but for static user hmm yes we want to prompt for password also, I will add it

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: From what I understood we do not want to hit directly apiserver to reach this proxy as you need to be authenticated through apiserver to hit this proxy but you also need to authenticate on salt-api so .... For me we want to proxy apiserver

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: https://github.com/scality/metalk8s/pull/2399#discussion_r411979294

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Yes true I will add it

docs/developer/architecture/centralized-cli.rst
gdemonet: > we want to proxy apiserver

But we're already authenticated to apiserver, and can already talk to it, since we're in a `kubectl` plugin, right?
And we can pass the token we already have from apiserver to Salt API, that should be enough.
I don't understand what we are missing...

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Is it really part of this design ? For me here we don't really care what kind of kubeconfig user may use no ?
But yes agree we may want to document how to get a kubeconfig (and I think for user it will be a kubeconfig with a token not a certificate, no ?)

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Hmm true I can talk about it here

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: > Also, we'd probably like to install it in the metalk8s-utils image.

Is it really relevant to have a kubectl plugin inside a container image ? (I think we do not have kubectl also for the moment)

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: > and we may also need to package some dependencies to install along with it (some of which we install with pip in the Salt master container, IIRC...).

Which one(s) ? For me, for the moment, the only dependents would be likely python3, argparse and salt that could be installed directly from our repos or "basic" ones from RHEL/CentOs. 
If we need python package that are not packaged then we will need to package them ourself as we provide a package for this `metalk8sctl` all depends should be available as package (or part of our package (but IMO it's not a good idea to vendor each depends directly in one package))

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Honestly, for this one I don't know yet, because we may want to have something that works without salt to check etcd status (using directly the etcd python lib) but we can also have some salt module that we run with `--local` to do that ..... 

Not sure about this one. Maybe yes only with salt it's good and for all other etcd interaction we have etcdctl installed with a configuration file

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Hmm yes we could, true. Then we also need to change the requirement for this one, but agree

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Yes true we should have some get, describe, etc. I just took command from the requirements for the moment.

Not sure we want to list all "utility commands" here ?

docs/developer/architecture/centralized-cli.rst
gdemonet: > Which one(s) ? For me, for the moment, the only dependents would be likely python3, argparse and salt that could be installed directly from our repos or "basic" ones from RHEL/CentOs.

We'll probably need a system package for the `etcd3` Python package we're depending on (in the Salt Master only, atm, and maybe we don't want minions to use it... :shrug:).

> If we need python package that are not packaged then we will need to package them ourself as we provide a package for this metalk8sctl all depends should be available as package (or part of our package (but IMO it's not a good idea to vendor each depends directly in one package))

Agreed, that was indeed my point :)

docs/developer/architecture/centralized-cli.rst
gdemonet: Let's list all ideas in some table, and only implement a subset in the first iteration. But yeah, having all the "possible" commands can help us adjust the design for the first commands we'll implement.

docs/developer/architecture/centralized-cli.rst
gdemonet: cc @thomasdanan

docs/developer/architecture/centralized-cli.rst
gdemonet: I'd say our CLI should focus on Salt implementations, because that's how we'll get the UI to do the same features. Not sure we want to add and `etcd` client in the UI (and even, I don't think that would work, because `etcd` shouldn't be listening to the outside world...).

EDIT: uh, but this is part of the host-local CLI (I guess because that's how you would try to fix `etcd`, while any `kubectl` would be broken if `etcd` is down) - in this case, maybe we can focus on the minimal features that a minion would need anyway (like checking the local `etcd` member, restarting its container, restoring from a backup...)

docs/developer/architecture/centralized-cli.rst
gdemonet: Well, if it's not part of this one, then it needs to be part of another - otherwise the `kubectl` plugin can't be an acceptable solution for our users, who mainly don't care about K8s itself and its technologies.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Ok just did some test it's doable giving two token on every call, right I will talk about it here

```
curl -kvvi -X POST https://10.100.2.176:6443/api/v1/namespaces/kube-system/services/https:salt-master:api/proxy/ -H "Authorization: Bearer $id_token" -H "X-Auth-Token: $salt_token" -d 'client=local' -d 'tgt=*' -d 'fun=test.ping'
```

TeddyAndrieux: Done

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Done

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Done in requirement, I have updated the list of commands

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,187 @@ Cluster Service Administration
...
+  great interactions with APIs but interactions between ``Go`` and Kubernetes
+  are much better, with a sample CLI available.
+
+``metalk8sctl`` CLI
+^^^^^^^^^^^^^^^^^^^
+
+Language
+""""""""
+
+``Python`` is chosen as the language for kubectl plugin for the following
```
alexandre-allard:
```suggestion
``Python`` is chosen as the language for ``metalk8sctl`` for the following
```

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,187 @@ Cluster Service Administration
...
+
+  Never.
+
+- Output style:
+
+  Human readable output, do not necessarily need for "machine output" like
+  JSON and YAML.
+
+  Output should extract useful information from Salt output when needed, in
+  case of error only output useful error message(s) from Salt output
```
alexandre-allard: Many `output` here:
```suggestion
  The output should display useful information from Salt returns when needed, and
  in case of error, only show relevant error message(s) from Salt.
```

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Oh shit, Bad copy pasta

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: :+1:

thomasdanan (changes requested):
excellent, I just think we are missing to explain what test strategy we are putting in place

docs/developer/architecture/centralized-cli.rst
```diff
@@ -122,13 +126,40 @@ Cluster Administration
...
++----------+----------+-------------------+
+| solution | get      | <name>, <version> |
++----------+----------+-------------------+
+
+**tool: kubectl metalk8s**
+
++----------+---------------+------------------------------------------+
+| action   | resource type | parameters                               |
++==========+===============+==========================================+
+| create   | environement  | name, <description>, <namespace>         |
```
xaf-scality: typo here: environment

docs/developer/architecture/centralized-cli.rst
```diff
@@ -122,13 +126,40 @@ Cluster Administration
...
++----------+---------------+------------------------------------------+
+| delete   | environment   | name, <namespace>                        |
++----------+---------------+------------------------------------------+
+| get      | environment   | <name>                                   |
++----------+---------------+------------------------------------------+
+| add      | solution      | name, version, environment, <namespace>  |
++----------+---------------+------------------------------------------+
+| delete   | solution      | name, environment, <namespace>           |
++----------+---------------+------------------------------------------+
+| get      | solution      | <name>, environement                     |
```
xaf-scality: typo here: environment

docs/developer/architecture/centralized-cli.rst
xaf-scality: perhaps worth also to explain how authentication is working with the tool metalk8sctl

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,217 @@ Cluster Service Administration
...
+^^^^^^^^^^^^^^^^^^^
+
+Language
+""""""""
+
+``Python`` is chosen as the language for ``metalk8sctl`` for the following
+reasons:
+
+- Ability to interact with Salt directly with Python API.
+- Python installation needed anyway by Salt-minion.
```
xaf-scality, line 314: Perhaps a side note to explain that we will rely on python 3.4/3.6/3.xxx because 2.7 will be deprecated end of year.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,217 @@ Cluster Service Administration
...
+First have a simple framework to implement a simple command,
+then each command would update the framework if needed.
+
+Check `Requirements`_ for a full list of commands.
+
+Documentation
+-------------
+
+All command should be documented in the Operational Guide with a reference to
+it when it's needed in the Installation Guide.
```
xaf-scality, line 400: all commands should have a local help if we do not have the guide near us.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Hmm we do not have reall authentication in this tool as it's only local action but yes I should add it I agree

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: It's deprecated since the beginning of the year but right :+1: I will add it

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Indeed, I will add it

docs/developer/architecture/centralized-cli.rst
```diff
@@ -106,7 +106,11 @@ Cluster Administration
 +============+============+===============================================+
 | bootstrap  | deploy     |                                               |
 +------------+------------+-----------------------------------------------+
-| bootstrap  | import-iso | path_to_iso                                   |
+| archive    | import     | path_to_iso                                   |
++------------+------------+-----------------------------------------------+
+| archive    | get        | <name>                                        |
++------------+------------+-----------------------------------------------+
+| archive    | delete     | path_to_iso or path_to_mountpoint or name     |
```
gdemonet, line 113: Maybe `remove` or `unimport` would be better, but it's really minor so feel free to ignore

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,233 @@ Cluster Service Administration
...
+Language
+""""""""
+
+``Go`` is chosen as the language for kubectl plugin for the following reasons:
+
+- Great interactions with Kubernetes API.
+- Often used for operators and kubectl plugins
+  (`Sample CLI plugin <https://github.com/kubernetes/sample-cli-plugin>`_,
+  `Helpers for kubectl plugins <https://github.com/kubernetes/cli-runtime>`_).
+- Portable as it's just a binary.
```
slaperche-scality: It's "portable" because it's a statically compiled binary, being a binary doesn't make you portable at all per se.
Also, "portable" may not be the right term here (won't run on other OSes or CPU architecture, not our use case anyway).

It's more like "easy to ship because it's a statically compiled binary, no deps to provide".

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,233 @@ Cluster Service Administration
...
+This package should be installed on the bootstrap node by default.
+
+Rejected design choice
+""""""""""""""""""""""
+
+- ``Bash`` kubectl plugin: Bash is great to do simple actions but not when
+  you need to do interaction with some API like Kubernetes API or Salt API.
+- ``Python`` kubectl plugin: Python allows us to do complicated actions and
+  great interactions with APIs but interactions between ``Go`` and Kubernetes
+  are much better, with a sample CLI available.
```
slaperche-scality: I'm not sure I understand what you mean by `interactions between ``Go`` and Kubernetes
  are much better, with a sample CLI available.` :thinking:

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,233 @@ Cluster Service Administration
...
+``metalk8sctl`` CLI
+^^^^^^^^^^^^^^^^^^^
+
+Language
+""""""""
+
+``Python`` is chosen as the language for ``metalk8sctl`` for the following
+reasons:
+
+- Ability to interact with Salt directly with Python API.
```
slaperche-scality: You mean you can reuse Salt code as a library (since you can `import salt`)?

It's not very clear here, it sounds like `can interact with Salt through an API`, which any language can do.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: You right, I will update this.

> Also, "portable" may not be the right term here (won't run on other OSes or CPU architecture, not our use case anyway).

Hmm it's one use case here because this kubectl plugin may run from customer laptops so likely any OSes

thomasdanan (dismissed):

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Yes but it's not really library code it's really dedicated function in Python used to interact with Salt it's an API in Python to interact with Salt itself 
So it's a Python API. I agree it's not really clear here maybe `Python client API` could be better and match salt documentation
https://docs.saltstack.com/en/latest/ref/clients/

docs/developer/architecture/centralized-cli.rst
slaperche-scality: > Hmm it's one use case here because this kubectl plugin may run from customer laptops so likely any OSes

As long as it's x86_64 Linux OS it should be fine (i.e, it's portable across Ubuntu, Fedora, Archlinux, CentOS, …).

But if the customer wants to run it from MacOS, Windows, or some ARM board I'm not sure it will work out of the box.

`Go` can cross-compile and generate binaries for those platform quite easily (as long as our dependencies also works on those systems/architecture), but that will be a distinct binary (example [here](https://github.com/helm/helm/releases), where `helm` provide one binary for each case).

So yeah, it can be truly portable, but that's not "one binary to run them all".

So, given our dependencies itself are portable, our plugin can be portable too but the user may have to compile it ourself (or we need to provide/release cross-compiled binary like `helm` does).

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: `much simpler` maybe as you have several samples and sample CLI it's just another point for `Go` vs `Python` just the fact that you have a "Sample CLI plugin" from Kubernetes written in Go

docs/developer/architecture/centralized-cli.rst
slaperche-scality: Yeah, let's call it `Python client API`, and you can even add the link so it's even clearer :slightly_smiling_face:

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Ok good to know then I will changes this one for the moment and we will think about non-linux x86_64 later (I will add a todo for this) :+1:

docs/developer/architecture/centralized-cli.rst
slaperche-scality: Ok I see.
That doesn't make the interaction easier per se, but indeed make the development easier because you have lot of example around.

Maybe:

> but interactions between ``Go`` and Kubernetes are much easier, given the large number of example available.

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: We already have the link in `Salt interaction` section but right I will add the link for both :+1:

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: #2487

NicolasT: > One thing we forgot to mention: kubectl authentication is based on kubeconfig files, with certificates, whereas our users will rely on an OIDC provider to authenticate, most of the time.

That's not true, a `KubeConfig` object can have certs embedded, or a token, or HTTP basic auth.

> It would be nice to include in this design an approach for generating the required certs (maybe even the kubeconfig, I think @NicolasT mentioned tools to do that) based on an OIDC-provided identity.

Those tools don't generate certs or anything like that. They generate a `KubeConfig` including the JWT token.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,239 @@ Cluster Service Administration
...
+
+  Human readable output, do not necessarily need for "machine output" like
+  JSON and YAML.
+
+  The output should display useful information from Salt returns when needed,
+  and in case of error, only show relevant error message(s) from Salt.
+
+.. todo::
+
+  - Define logging informations
```
NicolasT, line 344: IMO, any kind of *persisted* logging shall be done at the Salt/API/... level (i.e., some kind of audit logs), *not* by some CLI tool/client itself.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -122,13 +126,40 @@ Cluster Administration
...
-| solutions  | import-iso | path_to_iso                                   |
-+------------+------------+-----------------------------------------------+
++----------+------------+-------------------+
+| Resource | action     | parameters        |
++==========+============+===================+
+| solution | activate   | name, version     |
++----------+------------+-------------------+
+| solution | deactivate | name              |
++----------+------------+-------------------+
+| solution | get        | <name>, <version> |
```
gdemonet, line 143: Right now, we need these to be host-local (because the whole system depends on the `/etc/metalk8s/solutions.yaml` file), so let's keep it like this.

But in a later iteration, I think this would make sense to manage in K8s only (only the archives part need to be on the bootstrap, and as such, could just be added to the list in `/etc/metalk8s/bootstrap.yaml`... cc @alexandre-allard-scality).

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,239 @@ Cluster Service Administration
...
+  kubeconfig will not work with this kubectl plugin.
+
+  Add support for certificates based authentication in Salt API look quite hard
+  and costly.
+
+.. todo::
+
+  Check what can be done to support certificates based kubeconfig or if we
+  consider that we only support token-based kubeconfig for our plugin commands
+  interacting with Salt API.
```
gdemonet, line 280: Do we have a ticket for this one? Would be good to dump the conversation we had in there.

docs/developer/architecture/centralized-cli.rst
```diff
@@ -149,3 +180,239 @@ Cluster Service Administration
...
+Check `Requirements`_ for a full list of commands.
+
+Documentation
+-------------
+
+All command should be documented in the Operational Guide with a reference to
+it when it's needed in the Installation Guide.
+
+All commands and sub-commands should have a ``--help`` option to explain a bit
+of usage of this specific command and available options.
```
gdemonet, line 403: If we have this, would be great to have it in a Reference Guide in the docs as well :)

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: For `metalk8sctl` IMO we should log stuff in the CLI as we may interact with the machine itself, for the kubectl plugin I agree it's only API driven so it's ok.

gdemonet (approved):
:+1: Great stuff :) Just do it™

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: I hesitate a bit with `unimport` but it's a kind of deletion but ... yes I agree, lets keep `delete` for this PR and change it later if we want to

docs/developer/architecture/centralized-cli.rst
NicolasT: After bootstrap, is there anything we'd do on the node that's *not* performed through Salt?

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: No, but I think we should discuss a bit more about this one before creating a ticket

slaperche-scality (approved):

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: No you right, but I think we would like to have log to know for example when an upgrade is launched with the result etc, because in Salt logs you will have each Salt call but not a global view.

For me since a command consists of the execution of several API calls (which is not the case for kubectl plugin normally) we want to have a log file for this. (but maybe we can discuss about it)

docs/developer/architecture/centralized-cli.rst
TeddyAndrieux: Reference Guide ?

alexandre-allard (approved):

thomasdanan (approved):

docs/developer/architecture/centralized-cli.rst
NicolasT: `eject`?

xaf-scality (approved):

PR #2405, opened 2020-04-14
[UI] Extend the Eslint configure for react app

### LINKED ISSUE #2400: Extend the ESLint config for React App
**Component**: UI

**Why this is needed**:
Currently, we only have the `react` ESLintConfig extend, which is not enough.

```
  "eslintConfig": {
    "extends": "react-app"
  },
```

For ESLint, No rules are enabled by default. We should at least enable the  "extends": "eslint:recommended" property in a configuration file enables rules that report common problems.
(https://eslint.org/docs/rules/)

Otherwise, for the following code, we won't display any error: `Arrow function should not return assignment  no-return-assign`
```
    const availableDeployingSolution = availableSolutions?.find(
      (avaSol) => avaSol.name === solName,
    );
```

**What should be done**:
Enable the  "extends": "eslint:recommended" property in a configuration file.

TBD
**Test plan**:
TED

ChengYanJin:
**Component**: UI

**Context**:  We should extend the eslint config to include the recommended rules.

**Summary**: 
- Add the extends`eslint:recommended` and `no-return-assign` rule
```
    "extends": [
      "react-app",
      "eslint:recommended"
    ],
    "rules": {
      "no-return-assign": [
        "error",
        "always"
      ]
    }
```
- fix errors after the enable the recommended esline configure

**Acceptance criteria**: 
No lint error.

---

Closes: #2400

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -46,7 +46,16 @@
     "flow": "flow check"
   },
   "eslintConfig": {
-    "extends": "react-app"
+    "extends": [
+      "react-app",
+      "eslint:recommended"
+    ],
+    "rules": {
+      "no-return-assign": [
+        "error",
+        "always"
+      ]
+    }
   },
   "browserslist": [
     ">0.2%",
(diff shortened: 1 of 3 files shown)


MonPote (approved):
LGTM

Issue #2409, opened 2020-04-15
Volume stuck in Pending state forever after a kubectl delete


slaperche-scality:
**Component**:

operator

**What happened**:

When trying to delete a volume whose creation was in progress, the volume stay stuck in Pending state forever.

**What was expected**:

We let the volume creation proceed until the end (be it with an error or a success) and then the deletion happen.

**Steps to reproduce**

`kubectl apply -f volume.yaml && sleep 1 && kubectl delete -f volume.yaml`

**Resolution proposal**:

Do not exit the reconciliation loop too early, in order to let `deployVolume` moving the process forward.

PR #2410, opened 2020-04-15
Fix deletion on pending volume

slaperche-scality:
**Component**:

operator

**Context**: 

When trying to delete a volume whose creation was in progress, the volume stay stuck in Pending state forever.

**Summary**:

Don't exit the reconciliation loop too early, to let `deployVolume` move the creation process forward.

**Acceptance criteria**: 

A new test was added, so the CI should be green.
Otherwise, you can test with `kubectl apply -f volume.yaml && sleep 1 && kubectl delete -f volume.yaml` => it should work.

---

Closes: #2409

### DIFF
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -588,7 +588,14 @@ func (r *ReconcileVolume) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 	// Check if the volume is marked for deletion (i.e., deletion tstamp is set).
 	if !volume.GetDeletionTimestamp().IsZero() {
-		return r.finalizeVolume(ctx, volume, saltenv)
+		// Pending volume: can do nothing but wait for stabilization.
+		if volume.ComputePhase() == storagev1alpha1.VolumePending {
+			reqLogger.Info("pending volume cannot be finalized: requeue")
+			// Do not return here! We need to re-enter deployVolume to keep
+			// polling the Salt job and make progress.
+		} else {
+			return r.finalizeVolume(ctx, volume, saltenv)
+		}
 	}
 	// Skip volume stuck waiting for deletion or a manual fix.
 	if condition := volume.IsInUnrecoverableFailedState(); condition != nil {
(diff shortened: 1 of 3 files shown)


storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -676,13 +683,6 @@ func (self *ReconcileVolume) finalizeVolume(
 	saltenv string,
 ) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Volume.Name", volume.Name)
-
-	// Pending volume: can do nothing but wait for stabilization.
```
NicolasT, line 680: Does this *need* to be removed, or would it make sense to keep this around for safety sake?

slaperche-scality: We could keep it, it wouldn't be dangerous as it would be dead code.
But it would be redundant with the check done at the upper level, no?

tests/post/features/volume.feature
```diff
@@ -148,3 +148,20 @@ Feature: Volume management
...
+            metadata:
+              name: test-volume11
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s-prometheus
+              sparseLoopDevice:
+                size: 10Gi
+        Then the Volume 'test-volume11' is 'Pending'
+        When I delete the Volume 'test-volume11'
+        Then the Volume 'test-volume11' does not exist
```
gdemonet, line 166: Are we sure this is going to test "deletion while creation"?

It's the same TOCTOU as you often remember me about :) Sure, volume was `Pending`, but when you hit the `delete Volume` line, maybe it's not `Pending` anymore...

Maybe a single `When I create and immediately delete the following Volume: ...` would help.

gdemonet (approved):
Nitpick on the BDD test, but the PR could be merged as is IMO, especially if someone's waiting for it.

tests/post/features/volume.feature
slaperche-scality: Yeah, it's racy but I trust Salt to be sloooooow enough to work (if it's flaky the test will succeed anyway so it won't bother us too much).

Having a single step may help but would still be racy (I had the case when testing manually where the delete would happen fast enough that the volume was still in "no state", not yet in Pending)

Issue #2411, opened 2020-04-15
Downgrade from ... whatever to 2.4.2 when bootstrap already in 2.4.2 broken


TeddyAndrieux:
**Component**:

'salt', 'lifecycle'

**What happened**:

When we downgrade and the bootstrap node is already in 2.4.2 with salt module sync to 2.4.2 ext_pillar for solutions fail because trying to launch a module on the bootstrap node not available in 2.4.2

```
[ERROR   ] Execption caught loading ext_pillar 'metalk8s_solutions':
  File "/usr/lib/python2.7/site-packages/salt/pillar/__init__.py", line 956, in ext_pillar
    key)
  File "/usr/lib/python2.7/site-packages/salt/pillar/__init__.py", line 875, in _external_pillar_data
    ext = self.ext_pillars[key](self.minion_id, pillar, **val)
  File "/var/cache/salt/master/extmods/pillar/metalk8s_solutions.py", line 100, in ext_pillar
    return {"metalk8s": {'solutions': _load_solutions(bootstrap_id)}}
  File "/var/cache/salt/master/extmods/pillar/metalk8s_solutions.py", line 54, in _load_solutions
    for solution, versions in result['available'].items():

[CRITICAL] Pillar render error: Failed to load ext_pillar metalk8s_solutions: 'unicode' object has no attribute 'items'
```

**What was expected**:

ext_pillar function should just report the error in `_errors` and not raise

**Steps to reproduce**

```
salt-call saltutil.sync_all saltenv=2.4.2
```
And then try to run the downgrade from a higher version. 

**Resolution proposal** (optional):

In solution ext_pillar check for `saltutil_cmd` retcode to report the `ret` field to `errors` list if `retcode` is not equal to 0

```
salt/_pillar/metalk8s_solutions.py:32:        result['available'] = __salt__['saltutil.cmd'](
salt/_pillar/metalk8s_solutions.py-33-            tgt=bootstrap_id,
salt/_pillar/metalk8s_solutions.py-34-            fun='metalk8s_solutions.list_available',
salt/_pillar/metalk8s_solutions.py-35-        )[bootstrap_id]['ret']
```

PR #2414, opened 2020-04-16
salt: Do not raise when `saltutil_cmd` failed in solutions ext_pillar

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

If for whatever reason `saltutil_cmd` command to list available
solutions in solution ext_pillar fail, the `ret` would be a string
containing the error message so `result['available'].items()` would
raise.

**Summary**:

Check `retcode` of `saltutil_cmd` and add the return in `errors` list if
`retcode` is not equal to `0`

---

Fixes: #2411

### DIFF
--- a/salt/_pillar/metalk8s_solutions.py
+++ b/salt/_pillar/metalk8s_solutions.py
@@ -29,10 +29,17 @@ def _load_solutions(bootstrap_id):
 
     errors = []
     try:
-        result['available'] = __salt__['saltutil.cmd'](
+        available_ret = __salt__['saltutil.cmd'](
             tgt=bootstrap_id,
             fun='metalk8s_solutions.list_available',
-        )[bootstrap_id]['ret']
+        )[bootstrap_id]
+        if available_ret['retcode'] != 0:
+            raise Exception('[{}] {}'.format(
+                available_ret['retcode'],
+                available_ret['ret']
+            ))
+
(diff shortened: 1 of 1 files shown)


salt/_pillar/metalk8s_solutions.py
```diff
@@ -29,10 +29,17 @@ def _load_solutions(bootstrap_id):
...
+        available_ret = __salt__['saltutil.cmd'](
             tgt=bootstrap_id,
             fun='metalk8s_solutions.list_available',
-        )[bootstrap_id]['ret']
+        )[bootstrap_id]
+        if available_ret['retcode'] != 0:
+            raise Exception('[{}] {}'.format(
+                available_ret['retcode'],
+                available_ret['ret']
+            ))
```
Ebaneck, line 40: I am not sure we want to raise like that ... no>>
why not something like :
```
return __utils__['pillar_utils.errors_to_dict']([
    '{}'.format(available_ret['ret'])
])
```

TeddyAndrieux: If we raise that will be handle by the except block which is good IMO, we will have something like 
```
Error when listing available Solutions:[<error_code>] <error_msg>
```

And we do not want to return here btw

Ebaneck (approved):

Issue #2125, opened 2019-12-11
Migration of customised admin credentials when upgrading to 2.5


thomasdanan:
**Component**:

'upgrade'

**Why this is needed**:

Some users may have configured admin credentials for both k8s APIs and Grafana, using Metalk8s 2.4.X. Since we are changing the authentication backend, we should make sure those customizations are not lost during an Upgrade to MetalK8s 2.5.X

**What should be done**:

Ideally, the upgrade script automatically migrates customized admin credentials to the new authentication framework. At worst, we have a documented procedure to do this migration.

**Implementation proposal** (strongly recommended):

**Test plan**:

PR #2416, opened 2020-04-16
Add warning concerning the migration of customized admin credentials

Ebaneck:
**Component**:

'docs'

**Context**: 

See #2125 

**Summary**:

MetalK8s administrators should be aware of the breaking changes in 2.5.0 regarding 
the new authentication mechanism and how to migrate their old creds.

Should land in 2.4.3!!

**Acceptance criteria**: 

---

Closes: #2125

### DIFF
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -29,9 +29,16 @@ command.
 This will simulate the upgrade prechecks and provide an overview of the
 changes to be carried out in your MetalK8s cluster.
 
+.. important::
+
+    The version prefix metalk8s-**X.X.X** as used below during a MetalK8s
+    upgrade must be the new MetalK8s version you would like to upgrade
+    to.
+
    .. code::
 
-     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version <destination_version> --dry-run --verbose
+     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version \
+       <destination_version> --dry-run --verbose
 
 Upgrade Steps
 *************
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -24,6 +24,8 @@ Upgrade Pre-requisites
 Before proceeding with the upgrade procedure, make sure to complete the
 pre-requisites listed in :doc:`/operation/preparation`.
 
+Run pre-check
+-------------
 You can test if your environment will successfully upgrade with the following
 command.
 This will simulate the upgrade prechecks and provide an overview of the
@@ -39,6 +41,26 @@ changes to be carried out in your MetalK8s cluster.
 
      /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version \
        <destination_version> --dry-run --verbose
+
+Backup old credentials
+----------------------
+Starting 2.5.0, MetalK8s will henceforth implement OpenID Connect (OIDC) based
+authentication. Both K8s and Grafana will be configured to make use of the same
+OIDC provider.
+
+.. warning::
+
+    Before running an upgrade from 2.4.x to 2.5.0 or higher, MetalK8s
+    administrators **must** ensure all static users defined in
+    ``/etc/kubernetes/htpasswd`` can be recreated, and if any, all users that
+    were defined in Grafana.
+    The upgrade procedure will result in all admin credentials being reset
+    to their default values, and any additional user being removed. MetalK8s
+    administrators need to remember and reconfigure these username/password
+    pairs.
+
+    After upgrade is complete, a procedure for configuring the OIDC provider
+    (Dex) user store will be provided in the next version.
 
 Upgrade Steps
 *************
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -26,7 +26,7 @@ pre-requisites listed in :doc:`/operation/preparation`.
 
 You can test if your environment will successfully upgrade with the following
 command.
-This will simulate the upgrade procedure and provide an overview of the
+This will simulate the upgrade prechecks and provide an overview of the
 changes to be carried out in your MetalK8s cluster.
 
    .. code::
--- a/docs/operation/downgrade.rst
+++ b/docs/operation/downgrade.rst
@@ -29,9 +29,15 @@ command.
 This will simulate the downgrade prechecks and provide an overview of the
 changes to be carried out in your MetalK8s cluster.
 
+.. important::
+
+    The version prefix metalk8s-**X.X.X** as used below during a MetalK8s
+    downgrade must be the currently-installed MetalKs8 version.
+
    .. code::
 
-     /srv/scality/metalk8s-X.X.X/downgrade.sh --destination-version <destination_version> --dry-run --verbose
+     /srv/scality/metalk8s-X.X.X/downgrade.sh --destination-version \
+       <destination_version> --dry-run --verbose
 
 Downgrade Steps
 ***************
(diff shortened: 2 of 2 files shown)


docs/operation/upgrade.rst
```diff
@@ -24,14 +24,40 @@ Upgrade Pre-requisites
...
 
-     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version <destination_version> --dry-run --verbose
+     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version \
+       <destination_version> --dry-run --verbose
+
+Backup old credentials
+----------------------
+Starting 2.5.0, MetalK8s will henceforth implement OpenID Connect (OIDC) based
+authentication for both the MetalK8s UI and Grafana UI. Both UIs now make use
+of a common user store and ofcourse a common access credential.
```
gdemonet: Not sure we want to talk about UIs here: it's K8s and Grafana APIs which are authenticated.
Now, an interesting note to give to users is that both K8s and Grafana will be configured to use the same OIDC provider, as you stated in the last sentence (but using "OIDC provider" instead of "user store and access credential" would make it more obvious, IMHO).

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,40 @@ Upgrade Pre-requisites
...
+
+.. warning::
+
+    To run an upgrade to version 2.5.0, MetalK8s administrators **must**
```
gdemonet:
```suggestion
    Before running an upgrade from 2.4.x to 2.5.0 or higher, MetalK8s administrators **must**
```

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,40 @@ Upgrade Pre-requisites
...
+    perform a manual backup/save the passwords set for MetalK8s UI and Grafana
+    UI. Failure to save these credentials will mean administrators accept
+    losing old credentials upon upgrade.
```
gdemonet: No, actually we shouldn't formulate it like this. Instead, say that the upgrade procedure will be destructive, and that passwords will be reset to their default values, so the user needs to save them.

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,40 @@ Upgrade Pre-requisites
...
+
+After upgrade is complete, administrators can later follow
+:ref:`this procedure <Change-dex-static-user-password>` to re-instate/restore
+admin credentials.
```
gdemonet: Not sure referencing these docs in 2.4 will work well. Instead, maybe just mention that a procedure for configuring the OIDC provider's user store will be provided in the next version.

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,40 @@ Upgrade Pre-requisites
...
+Backup old credentials
+----------------------
+Starting 2.5.0, MetalK8s will henceforth implement OpenID Connect (OIDC) based
+authentication for both the MetalK8s UI and Grafana UI. Both UIs now make use
+of a common user store and ofcourse a common access credential.
+
+.. warning::
+
+    To run an upgrade to version 2.5.0, MetalK8s administrators **must**
+    perform a manual backup/save the passwords set for MetalK8s UI and Grafana
```
gdemonet: We can't really "backup" anything, they are passwords. Once set, they are encrypted, so no way to put those back into a new system. You need to "remember" them :innocent: 

Maybe rephrase to:
```suggestion
    ensure all static users defined in `/etc/htpasswd` can be recreated,
    and if any, all users that were defined in Grafana.
```

gdemonet (changes requested):
The backup note should actually be reworded more strongly IMHO (cc @thomasdanan):

We are going to **lose** all the users/passwords that were set, both in `/etc/htpasswd` and in Grafana configuration (any user created from Grafana UI would be lost on reboot today, even in 2.6, because Grafana isn't configured to use persistent storage - only the admin password can be saved in configuration).

The note should focus on this, because there will be no form of "restoration from a backup" as one could expect. It's going to be a complete re-creation of the user store (and maybe even remove some static users to replace them with an external provider, now that it's available as well :wink:).

docs/operation/upgrade.rst
Ebaneck: Done

docs/operation/upgrade.rst
Ebaneck: Done

docs/operation/upgrade.rst
Ebaneck: ok

docs/operation/upgrade.rst
Ebaneck: Done

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,43 @@ Upgrade Pre-requisites
...
    .. code::
 
-     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version <destination_version> --dry-run --verbose
+     /srv/scality/metalk8s-X.X.X/upgrade.sh --destination-version \
+       <destination_version> --dry-run --verbose
+
+Backup old credentials
+----------------------
+Starting 2.5.0, MetalK8s will henceforth implement OpenID Connect (OIDC) based
+authentication. Both K8s and Grafana will be configured to make use of thesame
```
gdemonet:
```suggestion
authentication. Both K8s and Grafana will be configured to make use of the same
```

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,43 @@ Upgrade Pre-requisites
...
+OIDC provider.
+
+.. warning::
+
+    Before running an upgrade from 2.4.x to 2.5.0 or higher, MetalK8s
+    administrators **must** ensure all static users defined in
+    ``/etc/kubernetes/htpasswd`` can be recreated, and if any, all users that
+    were defined in Grafana.
+    The upgrade procedure from 2.4.x to 2.5.0 or higher will result to admin
+    set usernames and passwords being resetted to default values. MetalK8s
```
gdemonet:
```suggestion
    The upgrade procedure will result in all admin credentialss being reset
    to their default values, and any additional user being removed. MetalK8s
```

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,43 @@ Upgrade Pre-requisites
...
+    administrators need to remember and reconfigure these username/password
+    pairs.
+
+After upgrade is complete, a procedure for configuring the OIDC provider (Dex)
+user store will be provided in the next version.
```
gdemonet: Maybe leave this in the warning block

gdemonet (changes requested):

docs/operation/upgrade.rst
Ebaneck: done

docs/operation/upgrade.rst
Ebaneck: Done

docs/operation/upgrade.rst
Ebaneck: Done

docs/operation/upgrade.rst
```diff
@@ -24,14 +24,43 @@ Upgrade Pre-requisites
...
+authentication. Both K8s and Grafana will be configured to make use of the same
+OIDC provider.
+
+.. warning::
+
+    Before running an upgrade from 2.4.x to 2.5.0 or higher, MetalK8s
+    administrators **must** ensure all static users defined in
+    ``/etc/kubernetes/htpasswd`` can be recreated, and if any, all users that
+    were defined in Grafana.
+    The upgrade procedure will result in all admin credentialss being reset
```
gdemonet:
```suggestion
    The upgrade procedure will result in all admin credentials being reset
```

gdemonet (approved):

Ebaneck:
//bypass_build_status

PR #2417, opened 2020-04-16
Enable grain caching and only spawn 1 node for multi-node CI stage

TeddyAndrieux:
**Component**:

'build', 'salt'

**Context**: 

1. Grains compute may take time on salt-master and they are computed at each and every salt command
2. We spawn 2 extra node in the multi-node CI stage when we only expand to the first one

**Summary**:

1. Enable grains caching for salt-master
2. Only spawn 1 extra node in the multi-node CI stage

---


gdemonet (approved):
:ok_hand:

PR #2424, opened 2020-04-17
Metalk8s UI adptation requirements

### LINKED ISSUE #2428: MetalK8s UI requirements
**Component**:

'ui'

**Why this is needed**:

Redefined the scope and general concepts of MetalK8s UI. List the functionalities and problems to solve. Describe the pages that will be displayed to the Platform Administrator

**What should be done**:

Requirements Doc

thomasdanan:
**Component**:

'ui'

**Context**: 

goal is to write down what the MetalK8s UI should deliver in terms of functionalities, define the high level screens/pages to develop and the way to navigate through them. 

Closes: #2428


docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,176 @@
...
+- Volume provisioning (adding/removing Volumes, configuring volumes)
+- Solution lifecycle (adding/deleting envs, activate/adding solutions)
+- MetalK8s services administration (configuring alerts, scaling service)
+
+Some of the operations are not exposed through API and thus cannot be exposed
+in the UI. The operation that are accessible in the UI are served with either
+K8S or Salt APIs. Not all operations are critical to have in the UI.
+
+Some operations, such as cluster expansion or volume addition can be time
+consuming. MetalK8s UI should provide precise progress and ETA. It should also
```
ChengYanJin: Not sure how we can do the ETA( Estimated Time of Arrival)?

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,176 @@
...
+
+
+*volumes*
+
+A table with the list of Volumes created on this node. For each volume, status,
+health, bounded / available, type and size is available. When clicking on one
+Volume, it is possible to access various information about this specific
+Volume in the Volume page.
+
+.. image:: img/node-view-volumes.jpg
```
ChengYanJin: We should decide the logic of breadcrumb. Since we may want to navigate back to the volume page from the specific volume.
Could be `platform > nodes > node-4> volumes > volume-1`?

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,283 @@
...
+Volume, it is possible to access various information about this specific
+Volume in the Volume page.
+
+*pods*
+
+A table with the list of pods scheduled on this node. For each pod, status,
+health, age, namespace are displayed.
+
+.. image:: img/nodes-view.jpg
+
```
nloewensen, line 166: In the image it is unclear why the "time-line" goes from -7days to now to sometime in the future (the "now" is not at the end of the line but somewhere 3/4 of the line). Is that really a forecast into the future or should the "now" be the right end of the line?

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,283 @@
...
+their health and status. From the Volumes page it is possible to create a new
+Volume.
+
+.. note:: One of the duty of the Platform Administrator is also to have a view
+          of the available disks or devices out of which we could create
+          Volumes / PVs. The Platform Administrator will want to know what are
+          the available devices and if they are healthy. Also when a disk is
+          not healthy, the Platform Administrator will want this disk to be
+          easily identified in the data center (i.e. blinking the disk)
+
```
nloewensen, line 204: For this "duty" wouldn't it be helpful to also show here (in the Volume page) which disks make up a volume/PV?

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,283 @@
...
+**Cards vs Table for Nodes ad Volumes**
+
+Cards offer a more sexy way of presenting instances. We can even layout or
+group cards according to some business logic criteria. However we can't display
+lot of information or sort the cards. The table is may be less fancy, however
+it helps the user to visualize lot of information at the same time, it usually
+embeds out of the box sorting, filtering and top 10 capabilities. Also each
+cell can be rendered using fancy components (an not only text).
+
```
nloewensen, line 243: can we "toggle" the view between cards (for a "sexier" look and a quick overview) and table (in case ore details are needed?

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,283 @@
...
+as well as Alert Manager to manage the alerts.
+
+As a Platform Administrator I want to know if there are some actives alerts on
+a service and I want to visualize the history of alerts In order to act on it
+to fix the issue.
+
+As a Platform Administrator I want to to know how a service is behaving in
+terms of performances (CPU, Load, Memory, IO), In order to anticipate potential
+failure events.
+
```
nloewensen, line 273: Perhaps this is already included but for many storage/storage-service, we also look at bandwidth, latency etc. so if not included, these should be in the monitoring.

docs/developer/architecture/metalk8s-ui.rst
```diff
@@ -0,0 +1,283 @@
...
+No detailed functionalities for now. This could be from where the Platform
+Administrator would trigger platform upgrade, downgrade, restore or backup.
+
+**Alerts Page**
+
+From this page the Platform Administrator can visualize all past and current
+alerts belonging to any entity of the platform. When clicking on one specific
+alert, the user is redirected the specific entity / health page on which the
+alert was fired.
+
```
nloewensen, line 306: In general, the ability to "acknowledge" alerts once they have been acted upon or once the root cause has been solved would be helpful in the workflows for the Administrator, since they can easily keep track of the alerts they still need to look at/work on.

docs/developer/architecture/metalk8s-ui.rst
thomasdanan: Maybe we can't do it but at least we want to tell the user it is progressing

docs/developer/architecture/metalk8s-ui.rst
thomasdanan: I think it is explained in the Overall Navigation schema no?

docs/developer/architecture/metalk8s-ui.rst
thomasdanan: It is going from -7 to now ... This is what Use Design suggested but we may want to put real day date instead. Also we want to have a time window selector in order to see the health over the last hour or the last 30 days for example.

docs/developer/architecture/metalk8s-ui.rst
thomasdanan: device or disk? With Zenko, we are going to slice a disk into multiple LVs so multiple devices at the end. Nevertheless, we can definitely add this info in the Volume Info box

docs/developer/architecture/metalk8s-ui.rst
thomasdanan: Yes that would be the ideal solution even if in a first iteration we would focus on having the table with all details.

xaf-scality (dismissed):

alexandre-allard (approved):

Issue #2434, opened 2020-04-20
Bootstrap fails if trying to get the Salt Master container while two exist


gdemonet:
**Component**: salt, scripts

**What happened**:

During execution of `bootstrap.sh` (observed at least on 2.6):

```shell
> Syncing Utility modules on Salt master...
time="2020-04-20T07:14:03-04:00" level=fatal msg="execing command in container failed: rpc error: code = Unknown desc = failed to find container \"112e206b7a22bbfd7eb49ccad329db2558622326a7961e28f91d9530131b6175\\n173aa24cfd67ce8ab440becf5f8fe753b027f02c48230a3888699e761bed3227\" in store: does not exist
```

**What was expected**: The bootstrap script shouldn't fail for this kind of issue

**Steps to reproduce**: It's a timing issue, and only rarely happens.

**Resolution proposal** (optional):

In the `get_salt_container` function, defined in `scripts/common.sh`, wait for the `crictl ps` query to only return a single container (if two are Running, we can't know for sure which one we should use, so better wait than try to guess).

PR #2435, opened 2020-04-20
scripts: Wait for a single Salt Master container

gdemonet:
Sometimes, if kubelet restarted the `salt-master` static Pod after an
operation, two containers matching the usual selector will co-exist for
a small time window.
If we use the `scripts/common.sh:get_salt_container` function at that
point in time, we may return a string with two container IDs instead of
just one, and subsequent commands will fail.
Instead, we now wait for a single container to exist (and also add a
sleep time between two attemps, which we didn't before).

Fixes: #2434

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -313,13 +313,21 @@ pre_minion_checks() {
 get_salt_container() {
     local -r max_retries=10
     local salt_container='' attempts=0
+    local -a found_containers=()
 
-    while [ -z "$salt_container" ] && [ $attempts -lt $max_retries ]; do
-        salt_container="$(crictl ps -q \
+    while [[ $attempts -lt $max_retries ]]; do
+        IFS=$'\n' read -r -d '' -a found_containers < <(crictl ps -q \
             --label io.kubernetes.pod.namespace=kube-system \
             --label io.kubernetes.container.name=salt-master \
-            --state Running)"
+            --state Running && printf '\0')
+
+        if [[ "${#found_containers[@]}" -eq 1 ]]; then
+            salt_container=${found_containers[0]}
+            break
+        fi
+        echo "Invalid number of candidates: ${#found_containers[@]}" >&2
         (( attempts++ ))
+        sleep 3
     done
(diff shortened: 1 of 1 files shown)


scripts/common.sh
```diff
@@ -313,13 +313,21 @@ pre_minion_checks() {
...
+    local -a found_containers=()
 
-    while [ -z "$salt_container" ] && [ $attempts -lt $max_retries ]; do
-        salt_container="$(crictl ps -q \
+    while [[ $attempts -lt $max_retries ]]; do
+        IFS=$'\n' read -r -d '' -a found_containers < <(crictl ps -q \
             --label io.kubernetes.pod.namespace=kube-system \
             --label io.kubernetes.container.name=salt-master \
-            --state Running)"
+            --state Running && printf '\0')
```
alexandre-allard, line 322: You should simply do:
```
 readarray -t found_containers < <(crictl ps -q \
            --label io.kubernetes.pod.namespace=kube-system \
            --label io.kubernetes.container.name=salt-master \
            --state Running)
```

alexandre-allard (approved):
minor comment, otherwise LGTM.

scripts/common.sh
gdemonet: I've read somewhere that `readarray` and `mapfile` could have both been used, but both depend on Bash version, while my approach should be POSIX compliant. True? If not, I can change to `readarray`.

alexandre-allard: It's true, `readarray` comes in bash 4, but since we're using bashism everywhere, I don't think it's really important to be POSIX compliant.
And we're deploying MetalK8s only on systems that ship bash version 4+.
But anyway, we can keep your approach, my point was more I don't understand why your chaning `IFS` and use `printf`, it should be sufficient to do something like:
```
        read -r -d '' -a found_containers < <(crictl ps -q \
            --label io.kubernetes.pod.namespace=kube-system \
            --label io.kubernetes.container.name=salt-master \
            --state Running)
```

PR #2439, opened 2020-04-21
ci: does not rely on worker_uuid anymore

alexandre-allard:
**Component**: ci

**Context**: 
In the CI context we were using worker_uuid property as a unique identifier prefix to all terraform resources we created, to avoid collision between these resources.
It happens that this UUID is not really unique and sometimes we have uncleaned resources on the MetalK8s scality.cloud CI tenant, so we end up having collision between old resources and new ones, this causes builds to fail.

**Summary**:
We now use the build number + the stage name as a prefix.
The build number should be unique accross any build and the stage name will be used to easily retrieve the resources of a specific stage of a build.

**Acceptance criteria**: 
Resources should be spawned with the right prefix and everything should still work fine.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -392,7 +392,7 @@ models:
         OS_USERNAME: "%(secret:scality_cloud_username)s"
         OS_PASSWORD: "%(secret:scality_cloud_password)s"
         OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
-        TF_VAR_worker_uuid: "%(prop:worker_uuid)s"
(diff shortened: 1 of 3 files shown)


gdemonet (approved):
OK for this branch, guess you'll need to adjust this for integration branches...

alexandre-allard:
> OK for this branch, guess you'll need to adjust this for integration branches...

@gdemonet I don't think so, I only target 2.5. Maybe you wanted me to target the 2.0 branch ?

gdemonet:
> I only target 2.5.

Ah, my bad thought we were on 2.0.

> Maybe you wanted me to target the 2.0 branch ?

No no, that's fine :)

Issue #2431, opened 2020-04-20
Bump Prometheus-adapter version to 0.6.0


Ebaneck:
**Component**:

'charts', 'monitoring', 'prometheus-adapter'

**Why this is needed**:

We need to update the Prometheus-adapter charts to the latest stable release which is v0.6.0

**What should be done**:

- Bump the currently deployed chart version to v0.6.0

**Implementation proposal** (strongly recommended):

**Test plan**:

no regression

PR #2441, opened 2020-04-21
Update Prometheus-adapter addon to version 0.6.0

Ebaneck:
**Component**:

'charts', 'prometheus-adapter', 

**Context**: 

See #2431 

**Summary**:

Bump Prometheus-adapter version to 0.6.0

**Acceptance criteria**: 

No regression

```
[root@bootstrap vagrant]# kubectl top node bootstrap
NAME        CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
bootstrap   346m         17%    2995Mi          81%     
```

---

Closes: #2431

### DIFF
--- a/charts/render.py
+++ b/charts/render.py
@@ -190,6 +190,7 @@ def main():
         '--service-config',
         action='append',
         nargs=2,
+        required=False,
         dest="service_configs",
         help="Example: --service-config grafana metalk8s-grafana-config"
     )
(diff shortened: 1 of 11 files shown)


salt/metalk8s/addons/prometheus-adapter/deployed/chart.sls
```diff
@@ -1,55 +1,117 @@
 #!jinja | metalk8s_kubernetes
+
 {%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+{%- set prometheus_adapter = salt.metalk8s_service_configuration.get_service_conf('metalk8s-monitoring', 'unavailable') %}
```
TeddyAndrieux, line 4: I think this line should not be in this commit

charts/render.py
```diff
@@ -212,17 +213,20 @@ def main():
...
+                    ("{{%- set {} = salt.metalk8s_service_configuration"
+                     ".get_service_conf('{}', '{}') %}}").format(
+                        service_config[0], args.namespace, service_config[1]
+                    ) for service_config in args.service_configs
+                )
+            ).lstrip()
+        )
+        sys.stdout.write('\n')
+    else:
+        sys.stdout.write(START_BLOCK.format(configlines='').lstrip())
```
TeddyAndrieux: Would prefer an `if` only to set a `configlines` variable and then having the write with format

TeddyAndrieux (approved):
Not tested but changes LGTM

Issue #2430, opened 2020-04-20
Bump Prometheus-operator version to 8.13.0


Ebaneck:
**Component**:

'charts', 'monitoring', 'prometheus-operator'

**Why this is needed**:

We need to update the chart version we currently deploy for Prometheus-operator to the latest stable release.

**What should be done**:

- Bump the version of the chart to 8.13.0
- Ensure grafana-piechart-panel plugin is present (no missing plugins)
- Check and validate that Grafana dashboards have data

**Implementation proposal** (strongly recommended):

**Test plan**:

Ebaneck:
Bumping the Prometheus-operator version to 8.13.0 results in the following
```
> Deploying bootstrap node (this may take a while)...[ERROR   ] An exception occurred in this state: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/state.py", line 1919, in call
    **cdata['kwargs'])
  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1918, in wrapper
    return f(*args, **kwargs)
  File "/var/cache/salt/master/extmods/states/metalk8s_kubernetes.py", line 132, in object_present
    **kwargs
  File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 158, in method
    force_custom_object=action in ['retrieve', 'delete', 'update']
  File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 54, in _extract_obj_and_kind_info
    raise CommandExecutionError('Invalid manifest: {!s}'.format(exc))
CommandExecutionError: Invalid manifest: Unsupported attribute preserveUnknownFields for "V1beta1CustomResourceDefinitionSpec" object.
```
This is likely due to the fact that the python2-kubernetes version we currently use does not have support for `preserve_unknown_fields(bool)` which was recently introduced in v11.0.0 here: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1beta1CustomResourceDefinitionSpec.md.

Issue #2029, opened 2019-11-08
Upgrade python-kubernetes version in salt-master


TeddyAndrieux:
**Component**:

'kubernetes', 'build'

**Why this is needed**:

Currently we use the `python2-kubernetes` version available in epel repository which is not the last one so we need to do some tricks to try to handle object with older APIs for example in this commit for Ingress using `networking.k8s.io/v1beta1` Api a008d388.
https://github.com/kubernetes-client/python#compatibility

**What should be done**:

Upgrade python kubernetes in salt-master container to have a version able to handle each and every object of current kubernetes version we use.

**Implementation proposal** (strongly recommended):

TBD
Some possibility:
- Package python-kubernetes version we want and install it in the salt-master container
- Install python kubernetes directly in the salt-master dockerfile (using pip or something else)

gdemonet:
Sadly, it seems github.com/kubernetes-client/python does not have (yet) released versions for K8s 1.15 or 1.16... Although 1.15 seems to be in progress (see the [11.0-alpha1](https://github.com/kubernetes-client/python/releases/tag/v11.0.0a1))

NicolasT:
Then maybe we can help maintaining that library? ;-)

TeddyAndrieux:
Version 1.10.1 just get pushed in epel repository so a008d38813e8cb88f53d1d2bd38184b66289488d is not longer an issue.

But something still need to be done to have a python kubenretes version able to support the k8s version deployed

TeddyAndrieux:
In recent version the kubernetes lib has been bumped in the salt-master container

PR #2449, opened 2020-04-23
containers: upgrade python kubernetes version deployed on salt master

Ebaneck:
**Component**:

'containers', 'kubernetes'

**Context**: 

See #2029 

**Summary**:

`Use pip install one last time.`
Updating the Prometheus-operator charts to 8.13.0 requires at least v11 of python-kubernetes package see https://github.com/scality/metalk8s/issues/2430#issuecomment-617692984.

Let us add this package to the salt-master image at build time knowing that we might need to build and/or package it as an rpm package shortly after.

We are not going to close issue #2029 for now but this PR will ensure we can close issue #2430 at least for now.

**Acceptance criteria**: 

- Tested the Prometheus-operator 8.13.0 charts and upgrade seems to work
- From salt-master container, verify the package version installed
```
[root@bootstrap vagrant]# crictl exec -it a32b3c04fe529 bash
[root@bootstrap /]# python
Python 2.7.5 (default, Oct 30 2018, 23:45:53) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
>>> import kubernetes
>>> print kubernetes.__version__
11.0.0
>>> 

```

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/images/salt-master/Dockerfile
+++ b/images/salt-master/Dockerfile
@@ -15,9 +15,18 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
  && rpm --import https://repo.saltstack.com/yum/redhat/7/x86_64/archive/${SALT_VERSION}/SALTSTACK-GPG-KEY.pub \
  && yum clean expire-cache \
  && yum install -y epel-release \
- && yum install -y python2-kubernetes salt-master salt-api salt-ssh openssh-clients \
+ && yum install -y salt-master salt-api salt-ssh openssh-clients \
  && yum install -y python-pip \
  && pip install pip==20.0.2 \
+ # TODO: issue #2454
+ # Docker build will die here because we have PyYAML==3.11 installed
+ # ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and
+ # thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall
+ # Since kubernetes-client requires PyYAML==3.12 ipaddress==1.0.17 we can ignore the installed versions using:
+ && pip install setuptools==40.3.0 \
+ && pip install --ignore-installed PyYAML==3.12 ipaddress==1.0.17 \
+ && pip install kubernetes==11.0.0 \
+ # End
  && pip install "etcd3 != 0.11.0" \
  && yum clean all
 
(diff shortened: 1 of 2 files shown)


images/salt-master/Dockerfile
```diff
@@ -15,9 +15,19 @@ gpgkey=https://repo.saltstack.com/yum/redhat/\$releasever/\$basearch/archive/%s/
...
+ && yum install -y salt-master salt-api salt-ssh openssh-clients \
  && yum install -y python-pip \
  && pip install pip==20.0.2 \
+ # Start of deps for python-kubernetes
+ # Note: Will replace pip install kubernetes & its deps below with a built rpm package
+ # Docker build will die here because we have PyYAML==3.11 installed
+ # ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and
+ # thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall
+ # Since kubernetes-client requires PyYAML==3.12 ipaddress==1.0.17 we can ignore the installed versions using:
+ && pip install setuptools==40.3.0 \
```
NicolasT, line 26: Install-from-PyPI/pip is against the container building guidelines we tend to use. This should be packaged, then installed... We basically don't want to put any 'random' files (i.e., files from which the origin can't be retrieved from `rpm -qf ...` or similar) in either the container images we ship, or when installed on a machine, except for configuration files in `/etc` when applicable.

gdemonet: We discussed this offline, and decided to provide a first version using pip, open a debt ticket to create some RPMs, so that we can close the addons upgrade tasks, currently blocked by this too old version of the client.

We are also installing `etcd3` with pip, so this one would require us to do some packaging as well. And there's the Python 3 dependencies... This will take some time, and while I agree we should enforce the guidelines you mentioned ASAP, I'd advocate we don't block other tasks while we are setting everything up.

(However, the comment in the Dockerfile could use a `TODO:` instead of just a `Note:` @Ebaneck, best would be to even have the issue number in there)

Ebaneck: We now have a dedicated issue here https://github.com/scality/metalk8s/issues/2454 to track the packaging of the k8s client

Ebaneck:
Closing this, since the python-kubernetes package has been upgraded in 2.6

PR #2458, opened 2020-04-24
[ui] delete the job when there is error occur

ChengYanJin:
**Component**: node

**Context**: 
There is no error handler when SALT indicates the error inside the JSON

**Summary**:
Since it's not easy to reproduce the bug, I'll just first add an error handler.

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/src/ducks/app/salt.js
+++ b/ui/src/ducks/app/salt.js
@@ -44,12 +44,12 @@ export default function reducer(state = defaultState, action = {}) {
     case REMOVE_JOB:
       return {
         ...state,
-        jobs: state.jobs.filter(job => job.jid !== action.payload.jid),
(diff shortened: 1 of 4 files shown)


gdemonet:
Aren't we missing error handling if SaltAPI doesn't answer / gives a bad status **at the HTTP level**?

ChengYanJin:
> Aren't we missing error handling if SaltAPI doesn't answer / gives a bad status **at the HTTP level**?

You are right! There are two errors we need to handle actually
- HTTP error (4xx, 5xx)
- Salt error (code 2xx, but error in the response)

MonPote (approved):
LGTM.
We really need to design a better way to handle salt error. But it's not the purpose of this PR.

Issue #2488, opened 2020-04-27
CSC defaults must be upgrade-able 


alexandre-allard:
**Component**: salt, csc

**Why this is needed**:
To be able to change the default values for the components using CSC during an upgrade or a downgrade (e.g. Alertmanager configuration file format changes between 2 versions).

**What should be done**:
We need to provide a mechanism to be able to change default values in CSC.

**Implementation proposal** (strongly recommended):
Instead of hardcoding default values in Salt formulas, we could use the `salt/metalk8s/defaults.yaml` to define these values.
So, we could simply use them through Jinja templating using a `| default()` or we could even do better, developing Salt modules taking care of merging the content of the CSC `ConfigMap` with the default values and returning the configuration as a dict.

**Test plan**:

PR #2501, opened 2020-04-28
docs: rework cluster and service configuration to make defaults updat…

Ebaneck:
**Component**:

'documentation', 'csc'

**Context**: 

See reason for this rework in: #2488

**Summary**:

Today, the default service configurations are hard-coded in service-config ConfigMaps and will remain untouched during upgrade/downgrade even if we specify new configurations values in the new version being deployed.

We absolutely need to consume new values if a user decides to add new stuff or fall back to defaults. Also upgrading from one version to the other may bring about new configs which need to be taken into account.
- The end goal is to be able to consume service configurations aggregated from user-customizable Configmaps + those defined in `defaults.yaml` using whatever merge strategy we decide while still keeping user-defined configurations intact.

**Acceptance criteria**: 

Must be validated by architects...

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/docs/developer/architecture/configurations.rst
+++ b/docs/developer/architecture/configurations.rst
@@ -98,37 +98,100 @@ actions:
   - Firstly, create and deploy ConfigMaps that will hold customizable cluster
     and service configurations.
     These ConfigMaps should define an empty `config.yaml` in the data section
-    of the ConfigMap for later use. A standard layout for each customizable
-    field could be added in the documentation to assist MetalK8s administrator
-    in adding and modifying customizations.
+    of the ConfigMap for later use.
 
-  - In a `csc.yaml` file located at the root directory of each MetalK8s Addon,
-    define the keys and values for default service configurations in a
-    YAML structured format.
+    A standard layout for each customizable field could be added in the
+    documentation to assist MetalK8s administrator in adding and modifying
+    customizations.
 
-      - The layout of service configurations within this file could have root
-        key being the impacted ConfigMap name. For example:
+    To simplify the customizing efforts required from MetalK8s administrators,
+    each customizable ConfigMap will include an example section with inline
+    documented directives that highlight how users should add, edit and remove
+    customizations.
 
-    .. code-block:: yaml
+  - In an Addon config file for example;
+    `salt/metalk8s/addons/prometheus-operator/config/alertmanager.yaml`, define
+    the keys and values for default service configurations in a YAML structured
+    format.
 
-        metalk8s-prometheus-config:
+      - The layout of service configurations within this file could follow the
+        format:
+
+        .. code-block:: yaml
+
+            # Configuration of the Alertmanager service
+            apiVersion: addons.metalk8s.scality.com/v1alpha1
+            kind: AlertmanagerConfig
+            spec:
+              # Configure the Alertmanager Deployment
+              deployment:
+                replicas: 1
+
+  - During Addon manifest rendering, call a Salt module that will merge
+    the configurations defined within the customizable ConfigMap to those
+    defined in `alertmanager.yaml` using a Salt merge strategy.
+
+    Amongst other merge technique such as `aggregate`, `overwrite`, `list`, the
+    `recurse` merge technique is chosen to merge the two data structures
+    because it allows deep merging of python dict objects while
+    being able to support the aggregation of list structures within the python
+    object.
+
+    Aggregating list structures is particularly useful when merging the
+    pre-provisioned Dex static users found in the default configurations to
+    those newly defined by Administrators especially during upgrade. Without
+    support for list merge, pre-provisioned Dex static users will be
+    overwritten during merge time.
+
+    `Recurse` merge strategy example:
+
+    Merging the following structures using `salt.utils.dictupdate.merge`:
+
+      - Object (a) (MetalK8s defaults):
+
+        .. code-block:: yaml
+
+          apiVersion: addons.metalk8s.scality.com/v1alpha1
+          kind: AlertmanagerConfig
           spec:
             deployment:
               replicas: 1
 
-  - During Addon manifests rendering, call a Salt module that will merge
-    the configurations defined within the customizable ConfigMap to those
-    defined in `csc.yaml` using the recommended merge strategy.
+      - Object (b) (User-defined configurations from ConfigMap):
 
-    The resulting configuration (a python dict object) will be used to populate
+        .. code-block:: yaml
+
+          apiVersion: addons.metalk8s.scality.com/v1alpha1
+          kind: AlertmanagerConfig
+          spec:
+            deployment:
+              replicas: 2
+            notification:
+              config:
+                global:
+                  resolve_timeout: 5m
+
+      - Result of Salt `recurse` merge:
+
+        .. code-block:: yaml
+
+          apiVersion: addons.metalk8s.scality.com/v1alpha1
+          kind: AlertmanagerConfig
+          spec:
+            deployment:
+              replicas: 2
+            notification:
+              config:
+                global:
+                  resolve_timeout: 5m
+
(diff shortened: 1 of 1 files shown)


gdemonet:
> We absolutely need to consume new values if a user decides to add new stuff or fall back to defaults. Also upgrading from one version to the other may bring about new configs which need to be taken into account.
> - The end goal is to be able to consume service configurations aggregated from user-customizable Configmaps + those defined in defaults.yaml using whatever merge strategy we decide.

Haven't read the doc yet, but I'm not really understanding this goal... In the sentences above I understand: "we need to make sure we can merge default values with provided user customizations in the ConfigMap(s)". This is an implementation detail. The goal we are actually trying to achieve is: "we need to make sure default values are upgraded while keeping user modifications intact".

Ebaneck:
> > We absolutely need to consume new values if a user decides to add new stuff or fall back to defaults. Also upgrading from one version to the other may bring about new configs which need to be taken into account.
> > 
> > * The end goal is to be able to consume service configurations aggregated from user-customizable Configmaps + those defined in defaults.yaml using whatever merge strategy we decide.
> 
> Haven't read the doc yet, but I'm not really understanding this goal... In the sentences above I understand: "we need to make sure we can merge default values with provided user customizations in the ConfigMap(s)". This is an implementation detail. The goal we are actually trying to achieve is: "we need to make sure default values are upgraded while keeping user modifications intact".

Well noted!

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,36 @@ requirements for the following reasons:
...
+    adding and modifying customizations.
+  - In the `defaults.yaml` located in `salt/metalk8s` directory,
+    define the keys and values for default service configurations in a
+    structured format.
+
+      - The layout of service configurations within this file could have root
+        key being the impacted ConfigMap name.
+  - During Addon manifests rendering, call a Salt module that will merge
+    the configurations defined within the customizable ConfigMap to those
+    defined in `defaults.yaml` using the recommended merge strategy.
```
TeddyAndrieux: As I discussed with you @Ebaneck, it's not a good idea to have this default value in `defaults.yaml` if you do not use the pillar to merge it with.

For me you should have a `csc.yaml` (or another name) just next to each CSC salt state (e.g.: for dex `salt/metalk8s/addons/dex/deployed/csc.yaml`) containing all the defaults value for the config and in the SLS you just load this yaml file and get + merge with the associated config map.

TeddyAndrieux (approved):

TeddyAndrieux (dismissed):
ups (misclick the previous approve)

docs/developer/architecture/configurations.rst
Ebaneck: @TeddyAndrieux your comments have been taken into consideration but I prefer to have a single `csc.yaml` found at the root of `salt/metalk8s`

gdemonet: Then what's the difference between a key in `defaults.yaml` and a single `csc.yaml` file? I agree with Teddy's point, maintaining these defaults close to where we will use them makes it simpler to develop (and also, if you need to add/remove addons, you won't need to care about which parts of some common files you need to change, just create/delete a new directory in `salt/metalk8s/addons` and be done with it)

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,45 @@ requirements for the following reasons:
...
-  - Create and deploy ConfigMaps that hold cluster and service configurations
-    and pre-fill them with default values.
-  - Template service pods and deployments to consume configuration data
-    directly from the above deployed ConfigMaps
+  - Firstly, create and deploy ConfigMaps that will hold customizable cluster
+    and service configurations.
+    These ConfigMaps should define an empty `config.yaml` in the data section
+    of the ConfigMap for later use. A standard layout for each customizable
+    field could be added in the documentation to assist MetalK8s administrator
+    in adding and modifying customizations.
```
gdemonet: One thing we could do to make this simpler: on Install/Upgrade/Downgrade, push the current default values in another key of this ConfigMap (say, `defaults.yaml`), and if possible (should be okay if we just needed to paste the contents of a single `csc.yaml` :wink:), with inline comments.

This makes the ConfigMaps self-documented. Now, I'm not saying we shouldn't document all of this, but when in a terminal to `kubectl edit` such a configuration, having everything at hand would be quite comfortable.

Of course, this key in the ConfigMap would only serve as documentation, and should never be used for deploying (in case someone edits it).

docs/developer/architecture/configurations.rst
Ebaneck: ok, I see

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,45 @@ requirements for the following reasons:
...
+
+      - The layout of service configurations within this file could have root
+        key being the impacted ConfigMap name. For example:
+
+    .. code-block:: yaml
+
+        metalk8s-prometheus-config:
+          spec:
+            deployment:
+              replicas: 1
```
gdemonet: So, as mentioned above, I'd rather have a single "config" in those `csc.yaml` files (I don't like the name, will suggest another below), and make it exhaustive (if some value **must not** have a default, still provide an example as a comment). For instance, the `salt/metalk8s/addons/prometheus-operator/config/grafana.yaml`:

```
# Configuration of the Grafana service
apiVersion: addons.metalk8s.scality.com/v1alpha1
kind: GrafanaConfig
spec:
  # Configure the Grafana Deployment
  deployment:
    replicas: 1
```

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,45 @@ requirements for the following reasons:
...
+
+  - During Addon manifests rendering, call a Salt module that will merge
+    the configurations defined within the customizable ConfigMap to those
+    defined in `csc.yaml` using the recommended merge strategy.
```
gdemonet: What does "recommended" mean? Let's choose one as part of this design, it's not trivial.

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,45 @@ requirements for the following reasons:
...
+
+    The resulting configuration (a python dict object) will be used to populate
+    the desired configuration fields within each Addon chart at render time.
 
 This approach works because in a MetalK8s cluster, ConfigMaps for cluster and
 service configurations are available before we deploy the configured services.
```
gdemonet: This is not really a relevant comment anymore, right? Also, what about the behavior if the ConfigMaps are missing? Can we fallback on defaults only, or should we fail?

docs/developer/architecture/configurations.rst
```diff
@@ -91,17 +91,45 @@ requirements for the following reasons:
...
 
+Secondly, we need to ensure that the default values specified in
+`csc.yaml` are upgrade-able while user-defined configurations remain
+untouched.
```
gdemonet: That's part of the requirements. The choice of defining default values in Salt and merging them with the ConfigMap data is a solution to this requirement.

docs/developer/architecture/configurations.rst
gdemonet: Let's add an item for having these files included in the docs. That'll make things easier on our side.

gdemonet (dismissed):

alexandre-allard (approved):
LGTM

docs/developer/architecture/configurations.rst
Ebaneck: Yes, we will fall back to defaults if CSC Configmaps are empty. But if Configmaps are missing because an Admin deleted them, then the CSC module will fail then raise an error

docs/developer/architecture/configurations.rst
Ebaneck: Merge strategy has been added

Issue #2495, opened 2020-04-28
Monitoring Table can't display all the Alerts


ChengYanJin:
**Component**: ui 

**What happened**:
When the number of alerts is more than the capacity of the table, we can't scroll down to check the last few alerts.
![image](https://user-images.githubusercontent.com/18453133/80455522-33de6280-892c-11ea-9a3b-54be00fc2720.png)

**What was expected**:
We can scroll down until the end of the table.

**Steps to reproduce**
I was trying with dev/2.6, but we have the same issue in 2.5 branch
There are many ways to create many alerts. What I did is to create many new Nodes but without deploying them and wait 15mins. 

**Resolution proposal** (optional):
The height problem of the table.

gdemonet:
@ChengYanJin this can be closed I guess?

ChengYanJin:
Merged in PR #2503

PR #2503, opened 2020-04-30
[ui] Bugfix: monitoring page cannot display all the alerts

ChengYanJin:
**Component**: ui

**Context**: 
When the number of alerts is more than the capacity of the table, we can't scroll down to check the last few alerts.

**Summary**:
Change the height of the table from 100% to 80%

**Acceptance criteria**: 
We can scroll down until the end of the table.
image

---

Closes: #2495

### DIFF
--- a/ui/src/containers/ClusterMonitoring.js
+++ b/ui/src/containers/ClusterMonitoring.js
@@ -33,14 +33,13 @@ const PageContainer = styled.div`
   flex-direction: column;
   height: 100%;
   padding: ${padding.larger};
-
   .sc-loader {
     padding: 0 ${padding.smaller};
   }
 `;
 
 const TableContainer = styled.div`
-  height: 100%;
+  height: 80%;
   flex-grow: 1;
 
   .sc-table-column-cell-container-severity {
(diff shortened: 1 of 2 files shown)


ui/src/containers/ClusterMonitoring.js
```diff
@@ -33,14 +33,13 @@ const PageContainer = styled.div`
...
   padding: ${padding.larger};
-
   .sc-loader {
     padding: 0 ${padding.smaller};
   }
 `;
 
 const TableContainer = styled.div`
-  height: 100%;
+  height: 80%;
```
slaperche-scality, line 42: > For some reason, if we set the height of the table to 100% we can't see the last few rows.

If the component is from us, we should investigate why.
If it's a third-party library, we could report the bug to them.

thomasdanan:
@ChengYanJin how do we scroll down? I don't see any scroll bar ...

ChengYanJin:
> @ChengYanJin how do we scroll down? I don't see any scroll bar ...

The scroll bar doesn't occur all the time ;)
It appears when we are scrolling down.

ui/src/containers/ClusterMonitoring.js
ChengYanJin: Yes! We should know the root problem.

We are planning to replace this table library actually.
Because it causes too much issue than the value it brings. 
We will do some research on the table.

thomasdanan (approved):

Issue #2193, opened 2020-01-16
Basic Alerting functionalities


thomasdanan:
**Component**:

'alertmanager'

**Why this is needed**:

MetalK8s is coming with Prometheus and Alertmanager in order to efficiently define, fire and manage alerts for both k8s cluster and solutions. In the current situation, there is no doc or tools explaining how to perform very basic operations such as:

- configuring email notifications
- silencing/deactivating an alert
- customizing a rule (mainly to configure thresholds)
- Adding a new rule / alert (probably just documentation)

**What should be done**:

Tools or documentation should be produced in order to perform those 3 operations. Remember that this may have to be done in a completely offline environment where Alertmanager doc is maybe not accessible. 

**Estimation**: 1W

---

Depends on #1996

thomasdanan:
First iteration delivered

PR #2511, opened 2020-05-04
Alert silences management

### LINKED ISSUE #2464: Alert silences management
**Component**: monitoring, salt

**Why this is needed**:
For the user to be able to silence a specific alert when needed (e.g. maintenance).

**What should be done**:
Create a Salt module which exposes methods to manage silences (add, list & delete).  

To silence an alert, we need to send a query to the Alertmanager API.
This can be done using the following route::

    http://<alertmanager-ip>:9093/api/v1/silences

With a POST query content formatted as below::

    {
      "matchers": [
        {
          "name": "<alert-name>",
          "value": ".*",
          "isRegex": true
        }
      ],
      "startsAt": "2020-04-10T12:12:12",
      "endsAt": "2020-04-10T13:12:12",
      "createdBy": "<author>",
      "comment": "Maintenance is planned",
      "status": {
        "state": "active"
      }
    }

We must also be able to retrieve silenced alerts and to remove a silence.
This will be done using the API, with the same route using GET and DELETE word
respectively::

    # GET - to list all silences
    http://<alertmanager-ip>:9093/api/v1/silences

    # DELETE - to delete a specific silence
    http://<alertmanager-ip>:9093/api/v1/silence/<silence-id>

We will need to provide these functionnalities through a Salt module
``metalk8s_monitoring`` which could then be wrapped in a CLI and UI.

**Test plan**:

alexandre-allard:
**Component**: salt

**Context**: See #2193

**Summary**:
Add a Salt module to manage alert silences.

**Acceptance criteria**:
After having deployed a bootstrap node, being able to put a silence on an alert, list it and then remove it, from the Salt master:
```
salt-run salt.cmd metalk8s_monitoring.get_alerts
salt-run salt.cmd metalk8s_monitoring.add_silence KubeMemOvercommit
salt-run salt.cmd metalk8s_monitoring.get_silences active
salt-run salt.cmd metalk8s_monitoring.delete_silence ${silence_id}
```

---

Closes: #2464

### DIFF
--- a/salt/_modules/metalk8s_monitoring.py
+++ b/salt/_modules/metalk8s_monitoring.py
@@ -1,0 +1,213 @@
+"""Utiliy methods to interact with MetalK8s monitoring.
+"""
+
+from salt.exceptions import CommandExecutionError
+
+from datetime import datetime, timedelta
+
+MISSING_DEPS = []
+
+try:
+    import requests
+except ImportError:
+    MISSING_DEPS.append('requests')
+
+__virtualname__ = 'metalk8s_monitoring'
+
+
+def __virtual__():
+    if MISSING_DEPS:
+        error_msg = 'Missing dependencies: {}'.format(', '.join(MISSING_DEPS))
+        return False, error_msg
+
+    return __virtualname__
+
+
+def add_silence(value, name='alertname', is_regex=False, starts_at=None,
+                duration=3600, ends_at=None, time_format='%Y-%m-%dT%H:%M:%S',
+                author='', comment='', **kwargs):
+    """Add a new silence in Alertmanager.
+
+    Arguments:
+
+        value (str): value to check
+        name (str): label to check
+        is_regex (bool): Whether `value` should be treated as a regular
+            expression or not, defaults to False.
+        starts_at (str): Date when the silence starts, defaults to `now`.
+        duration (int): Duration of the silence in seconds, defaults to `3600`.
+        ends_at (str): Date when the silence ends, defaults to
+            `starts_at` + `duration`.
+        time_format (str): Time format for `starts_at` and `ends_at` arguments.
+            Support the `datetime` Python library flags.
+        author (str): Creator of the silence.
+        comment (str): A description of why this silence has been put.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_monitoring.add_silence KubeMemOvercommit
+        salt-call metalk8s_kubernetes.add_silence none name=severity
+            starts_at="2020-05-05T07:14:52" duration=7200
+    """
+    if starts_at is None:
+        starts_at = datetime.now()
+    else:
+        starts_at = datetime.strptime(starts_at, time_format)
+
+    if ends_at is None:
+        ends_at = starts_at + timedelta(seconds=duration)
+    else:
+        ends_at = datetime.strptime(ends_at, time_format)
+
+    body = {
+        "matchers": [{
+            "name": name,
+            "isRegex": is_regex,
+            "value": value,
+        }],
+        "startsAt": starts_at.strftime('%Y-%m-%dT%H:%M:%SZ'),
+        "endsAt": ends_at.strftime('%Y-%m-%dT%H:%M:%SZ'),
+        "createdBy": author,
+        "comment": comment,
+        "status": {"state": "active"},
+    }
+
+    response = _requests_alertmanager_api(
+        'api/v1/silences',
+        'POST',
+        json=body,
+        **kwargs
+    )
+
+    return response['silenceId']
+
+
+def delete_silence(silence_id, **kwargs):
+    """Delete a silence in Alertmanager
+
+    Arguments:
+
+        silence_id (str): ID of the silence to delete.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_monitoring.delete_silence \
+            64d84a9e-cc6e-41ce-83ff-e84771ff6872
+    """
+    _requests_alertmanager_api(
+        'api/v1/silence/{}'.format(silence_id),
+        'DELETE',
+        **kwargs
+    )
+
+
+def get_silences(state=None, **kwargs):
+    """Get the list of all silences in Alertmanager
+
+    Arguments:
+
+        state (str): Filter silences on their state (e.g. `active`),
+            if None, return all silences, defaults to `None`.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_monitoring.get_silences
+        salt-call metalk8s_monitoring.get_silences state=active
+    """
+    response = _requests_alertmanager_api(
+        'api/v1/silences',
+        'GET',
+        **kwargs
+    )
+
+    if state is not None:
+        silences = [
+            silence for silence in response
+            if silence['status']['state'] == state
+        ]
+    else:
+        silences = response
+
+    return silences
+
+
+def get_alerts(state=None, **kwargs):
+    """Get the list of all alerts in Alertmanager
+
+    Arguments:
+
+        state (str): Filter alerts on their state (e.g. `active`),
+            if None, return all alerts, defaults to `None`.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_monitoring.get_alerts
+        salt-call metalk8s_monitoring.get_alerts state=suppressed
+    """
+    response = _requests_alertmanager_api(
+        'api/v1/alerts',
+        'GET',
+        **kwargs
+    )
+
+    if state is not None:
+        alerts = [
+            alert for alert in response
+            if alert['status']['state'] == state
+        ]
+    else:
+        alerts = response
+
+    return alerts
+
+
+def _requests_alertmanager_api(route, method='GET', **kwargs):
+    endpoint = __salt__['metalk8s_kubernetes.get_service_endpoints'](
+        'prometheus-operator-alertmanager',
+        'metalk8s-monitoring',
+        kwargs.pop('kubeconfig', None),
+    )
+
+    url = 'http://{}:{}/{}'.format(
+        endpoint['ip'],
+        endpoint['ports']['web'],
+        route,
+    )
+
+    try:
+        session = __utils__['metalk8s.requests_retry_session']()
+        response = session.request(method, url, **kwargs)
+    except Exception as exc:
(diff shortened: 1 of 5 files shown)


salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+    url = 'https://{}:{}/alertmanager/{}'.format(
+        host, CONTROL_PLANE_INGRESS_PORT, route
+    )
+
+    try:
+        response = requests.request(method, url, verify=False, **kwargs)
+    except Exception as exc:
+        raise CommandExecutionError(
+            "Unable to query Alertmanager API on {}: {}".format(url, str(exc))
+        )
```
Ebaneck: I hear it is good practice to use `{!s}` instead of converting with str(exc)

salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+    if MISSING_DEPS:
+        error_msg = 'Missing dependencies: {}'.format(', '.join(MISSING_DEPS))
+        return False, error_msg
+
+    requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
+
+    return __virtualname__
+
+
+def _requests_alertmanager_api(route, method='GET', **kwargs):
```
Ebaneck: I think this is not the most important thing, to begin with, per say! the helper section could come last?
Btw, what is the pep8 standard for this?

salt/_modules/metalk8s_monitoring.py
Ebaneck: This is a request query that might fail for many different reasons, maybe we should handle retries? See example here: https://github.com/scality/metalk8s/pull/2249/files

salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+        'GET',
+    )
+
+    if state is not None:
+        silences = [
+            silence for silence in response
+            if silence['status']['state'] == state
+        ]
+    else:
+        silences = response
```
Ebaneck, line 135: Could this be a one-liner ??
`silences = [silence for silence in response if silence['status']['state'] == state] if state else response`

Ebaneck (changes requested):

salt/_modules/metalk8s_monitoring.py
alexandre-allard: I don't know for this, maybe... For me it does the exact same thing, but it's shorter.
Will change it.

salt/_modules/metalk8s_monitoring.py
alexandre-allard: It could be, but I don't see the point in doing this, other than making the code more compact and less readable.

salt/_modules/metalk8s_monitoring.py
alexandre-allard: Good point.

salt/_modules/metalk8s_monitoring.py
Ebaneck: Ok, let's keep it as you have done then...

salt/_modules/metalk8s_monitoring.py
alexandre-allard: I don't know for the PEP8 standard, is there anything about that ? I'm not sure. But we don't follow any standard for sure in our modules, should we ?

salt/_modules/metalk8s_monitoring.py
gdemonet: Modifiers like this are especially useful when you are passing a complex object to `str.format`:

```python
class MagicSpell:
    incantation = "Wingardium Leviosa"
    effects = ["Levitation", "CharismaBonus"]
    level = 3

    def __repr__(self):
        return (
            "<MagicSpell "
            "incantation={self.incantation} "
            "effect={self.effects!r} "
            "level={self.level!s}>"
        ).format(self=self)
```

It's even more common when using Python 3.6 f-strings:

```python
try:
    # break things
except Exception as exc:
    raise MyError(f"Couldn't break things: {exc!s}")
```

salt/_modules/metalk8s_monitoring.py
gdemonet: From the Zen of Python:

> Simple is better than complex.

,

> Sparse is better than dense.

and:

> Readability counts.

Ebaneck: Well noted!

salt/_modules/metalk8s_monitoring.py
Ebaneck: None that I know of.

gdemonet: Not sure if PEP8 says anything about this, as it's not really about formatting the code but more about organizing modules. If you want to make sure, feel free to read it and tell us :wink:

On the other hand, we could define our own standards for how to organize these modules, and indeed, where to put helper functions.
As a common practice, I usually try to define pieces of code before they are used. However, in Salt modules, it would probably make sense to have all the public methods first.

salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+
+CONTROL_PLANE_INGRESS_PORT = 8443
+
+
+def __virtual__():
+    if MISSING_DEPS:
+        error_msg = 'Missing dependencies: {}'.format(', '.join(MISSING_DEPS))
+        return False, error_msg
+
+    requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```
gdemonet: I'd rather make sure we trust the CA than go out of our way to ignore warnings globally (what about other consumers of `requests`?). It's the Ingress CA, so you can get the bundle at `/etc/metalk8s/pki/nginx-ingress/ca.[crt|key]`.

salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+        json = response.json()
+    except ValueError as exc:
+        if response.status_code != requests.codes.ok:
+            error = (
+                "Received HTTP code {} when querying Alertmanager API on {}"
+                .format(response.status_code, url)
+            )
+        else:
+            error = (
+                "Malformatted response returned from Alertmanager API: {}: {}"
```
gdemonet: Mmh, I think it'd be either `Misformatted` or `Malformed`

salt/_modules/metalk8s_monitoring.py
```diff
@@ -0,0 +1,173 @@
...
+            "{}: {}".format(json['errorType'], json['error'])
+        )
+
+    return json.get('data')
+
+
+def add_silence(value, name='alertname', is_regex=False, starts_at=None,
+                duration=3600, ends_at=None, time_format='%Y-%m-%dT%H:%M:%S',
+                author='', comment=''):
+    """Add a new silence in Alertmanager.
```
gdemonet, line 29: It's an execution module, I think Salt standard practice recommends adding a few examples to use the module from CLI.

salt/_modules/metalk8s_monitoring.py
slaperche-scality: FWIW, when the language allows it, I usually organize the code is a top-down fashion (i.e the helpers tends to be further down or at the end of the file).

Rationale: I assume that you start reading a file from the top (to read the module doc for instance). Then you read down from here.
If you're only looking for high level info, you'll get what you want in the first functions.
If you need more details, you keep reading further into the file, to dive into the implementation details.

When looking for high level info, you don't have to read each function to know if it's an internal helper (and you need to skip it) or if it's an "interesting" function.


> On the other hand, we could define our own standards for how to organize these modules, and indeed, where to put helper functions.

This could go in our Python best practice, if we come up with something that please everyone.

salt/metalk8s/addons/prometheus-operator/deployed/ingress.sls
```diff
@@ -0,0 +1,25 @@
+#! metalk8s_kubernetes
+
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: prometheus-operator-alertmanager
+  namespace: metalk8s-monitoring
```
gdemonet: Now that I'm thinking about it, do we **really** need the Ingress to talk to Alertmanager from Salt Master?

It's the first time that we need access to a Service from an execution module, so we didn't really have precedence, but we do look for Service Endpoints in the external pillar `metalk8s_endpoints`, and use those values in some formulas.
IMO, we should do the same here, and maybe refactor slightly to have:
- an execution module for gathering Endpoints for a Service
- this module used in the pillar module
- and use it instead of the Ingress in your `metalk8s_monitoring` module

IIRC, `requests` expose a way to pass multiple hosts to try from (but IIRC, an Endpoint is created if the Pod is Ready, so you shouldn't need it).

salt/_modules/metalk8s_monitoring.py
gdemonet: Well, that's only if we use the Ingress. But we're inside the cluster, so we shouldn't.

gdemonet (changes requested):
Ingress (even control-plane) objects are for exposing services to the outside. From Salt, we shouldn't need it.

salt/_modules/metalk8s_monitoring.py
alexandre-allard: done

salt/_modules/metalk8s_monitoring.py
alexandre-allard: I've moved the helper method to the bottom of the file.

salt/_modules/metalk8s_monitoring.py
alexandre-allard: Done

salt/_modules/metalk8s_monitoring.py
alexandre-allard: Removed

salt/_modules/metalk8s_monitoring.py
alexandre-allard: Changed to `Malformed`

salt/metalk8s/addons/prometheus-operator/deployed/ingress.sls
alexandre-allard: Removed, I now use alertmanager service endpoint.

gdemonet (approved):

Ebaneck (approved):

PR #2513, opened 2020-05-04
csc: Add ability to update CSC defaults during upgrade/downgrade

Ebaneck:
**Component**:

'salt', 'kubernetes', 'csc'

**Context**: 

See #2488 

**Summary**:

- This PR adds the ability to update default CSC values during upgrade and downgrade while keeping the CSC Configmap intact. This works because we are able to merge the user customizations defined within Configmaps into the defaults packaged alongside each chart and then use the resulting dict object to render & deploy Metalk8s addons.

Flow:
1. Create CSC defaults for each Addon  e.g `alertmanager.yaml`
2. Import the CSC defaults within the charts using the `render.py` script
3. Call a salt module that takes arguments `csc ConfigMap name`, `namespace` and `default csc`
4. The module should merge the configs read in `csc ConfigMap` to those of `default CSC` and use  the resulting object to deploy the Addon

**Acceptance criteria**: 

no regression

---

Closes: #2488

### DIFF
--- a/charts/render.py
+++ b/charts/render.py
@@ -41,6 +41,7 @@ START_BLOCK = """
 #!jinja | metalk8s_kubernetes
 
 {{%- from "metalk8s/repo/macro.sls" import build_image_name with context %}}
+{csc_defaults}
 {configlines}
 
 {{% raw %}}
@@ -212,15 +213,27 @@ def main():
                 doc=doc
             )
         )
+    if args.service_configs:
+        import_csc_yaml = '\n'.join(
+            ("{{% import_yaml 'metalk8s/addons/{0}/config/{1}.yaml' as "
+                "{1}_defaults with context %}}").format(
+                args.name, service_config[0]
+            ) for service_config in args.service_configs
+        )
 
-    config = '\n'.join(
-        ("{{%- set {} = salt.metalk8s_service_configuration"
-            ".get_service_conf('{}', '{}') %}}").format(
-            service_config[0], args.namespace, service_config[1]
-        ) for service_config in args.service_configs
-    ) if args.service_configs else ''
+        config = '\n'.join(
+            ("{{%- set {0} = salt.metalk8s_service_configuration"
+                ".get_service_conf('{1}', '{2}', {0}_defaults) %}}").format(
+                service_config[0], args.namespace, service_config[1]
+            ) for service_config in args.service_configs
+        )
+    else:
+        import_csc_yaml = ''
+        config = ''
 
-    sys.stdout.write(START_BLOCK.format(configlines=config).lstrip())
+    sys.stdout.write(START_BLOCK.format(
+        csc_defaults=import_csc_yaml, configlines=config).lstrip()
+    )
     sys.stdout.write('\n')
 
     stream = io.StringIO()
--- a/salt/_modules/metalk8s_service_configuration.py
+++ b/salt/_modules/metalk8s_service_configuration.py
@@ -3,6 +3,7 @@
 import logging
 import yaml
 
+import salt
 from salt.exceptions import CommandExecutionError
 
 
@@ -18,6 +19,7 @@ def __virtual__():
 def get_service_conf(
     namespace,
     configmap_name,
+    default_csc,
     apiVersion=None,
     kind=None,
     **kwargs
@@ -27,6 +29,7 @@ def get_service_conf(
     Arguments:
         configmap_name: the ConfigMap name
         namespace: the Namespace where the ConfigMap is stored
+        default_csc: the CSC imported from YAML file
 
     Returns:
         A dict of a specific service configuration
@@ -35,12 +38,20 @@ def get_service_conf(
 
     .. code-block:: bash
 
-        salt-call metalk8s_service_configuration.get_service_conf "metalk8s-monitoring" "metalk8s-prometheus-config"
+        salt-call metalk8s_service_configuration.get_service_conf
+            "metalk8s-monitoring" "metalk8s-prometheus-config"
+            '{"apiVersion":"addons.metalk8s.scality.com","kind":"PrometheusConfig","spec":{"deployment":{"replicas":1}}}'
     """
 
     if not configmap_name:
-        raise Exception(
+        raise CommandExecutionError(
             'Expected a ConfigMap name but got {}'.format(configmap_name)
+        )
+    if not isinstance(default_csc, dict):
+        raise CommandExecutionError(
+            'Expected default CSC for ConfigMap {} but got {}'.format(
+                configmap_name, default_csc
+            )
         )
 
     try:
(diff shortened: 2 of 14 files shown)


charts/render.py
```diff
@@ -212,15 +213,27 @@ def main():
...
-    ) if args.service_configs else ''
+        config = '\n'.join(
+            ("{{%- set {0} = salt.metalk8s_service_configuration"
+                ".get_service_conf('{1}', '{2}', {0}_defaults) %}}").format(
+                service_config[0], args.namespace, service_config[1]
+            ) for service_config in args.service_configs
+        )
+    else:
+        import_csc_yaml = ''
+        config = ''
```
Ebaneck, line 232: should we tolerate `a = b = <value>`

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -35,13 +38,19 @@ def get_service_conf(
...
     if not configmap_name:
         raise Exception(
             'Expected a ConfigMap name but got {}'.format(configmap_name)
         )
+    if not default_csc:
+        raise Exception(
+            'Expected default CSC for ConfigMap {} but got {}'.format(
+                configmap_name, default_csc
+            )
+        )
```
alexandre-allard: In a module, you should use `CommandExecutionError` rather than an `Exception`.
Also, for the second `if` I would check that the argument is a `dict` rather than checking if it's evaluated to false or not.
```suggestion
    if not configmap_name:
        raise CommandExecutionError(
            'Expected a ConfigMap name but got {}'.format(configmap_name)
        )
    if not isinstance(default_csc, dict):
        raise CommandExecutionError(
            'Expected default CSC for ConfigMap {} but got {}'.format(
                configmap_name, default_csc
            )
        )
```

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -99,6 +104,14 @@ def get_service_conf(
...
+    merged_config = salt.utils.dictupdate.merge(
+        default_csc, config, strategy='recurse',
+        merge_lists=True
+    )
+    if not merged_config:
+        raise CommandExecutionError(
+            'Expected to merge configs {} into {}'.format(
+                config, default_csc
+            )
+        )
```
alexandre-allard: Is this even possible to have an empty dict (or anything evaluated to False) here ?

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -35,13 +38,19 @@ def get_service_conf(
 
     .. code-block:: bash
 
-        salt-call metalk8s_service_configuration.get_service_conf "metalk8s-monitoring" "metalk8s-prometheus-config"
+        salt-call metalk8s_service_configuration.get_service_conf "metalk8s-monitoring" "metalk8s-prometheus-config" default_csc
```
alexandre-allard: Your example should be valid (or at least what's `default_csc`), this is not the case here, we don't really understand what's `default_csc`.

salt/_modules/metalk8s_service_configuration.py
```diff
@@ -27,6 +29,7 @@ def get_service_conf(
     Arguments:
         configmap_name: the ConfigMap name
         namespace: the Namespace where the ConfigMap is stored
+        default_csc: the csc configurations imported from YAML file
```
alexandre-allard: Should be `CSC`, not `csc`. And isn't `csc configurations` redundant as it stands for `Cluster & Services Configurations` ?

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
```diff
@@ -39,9 +39,16 @@ Create grafana-config ConfigMap:
...
-                replicas: 1
+            spec: {}
+        # example configuration
+          # example.yaml: |-
+          #   apiVersion: addons.metalk8s.scality.com
+          #   kind: GrafanaConfig
+          #   spec:
+          #     # Configure the Grafana Deployment
+          #     deployment:
+          #       replicas: 1
```
alexandre-allard: Not sure we want to keep this here as it does not really help, we should rather have some documentation exposing the CSC defaults.

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
```diff
@@ -63,9 +70,16 @@ Create prometheus-config ConfigMap:
...
+            spec: {}
+        # example configuration
+          # example.yaml: |-
+          #   apiVersion: addons.metalk8s.scality.com
+          #   kind: PrometheusConfig
+          #   spec:
+          #     # Configure the Prometheus Deployment
+          #     deployment:
+          #       replicas: 1
+
```
alexandre-allard: Same as above

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
```diff
@@ -87,27 +101,17 @@ Create alertmanager-config ConfigMap:
...
+        # example configuration
+          # example.yaml: |-
+          #   apiVersion: addons.metalk8s.scality.com
+          #   kind: AlertmanagerConfig
+          #   spec:
+          #     # Configure the Alertmanager Deployment
+          #     deployment:
+          #       replicas: 1
+          #     notification:
+          #       config: {}
```
alexandre-allard: Same

salt/metalk8s/addons/dex/deployed/service-configuration.sls
```diff
@@ -23,18 +23,23 @@ Create dex-config ConfigMap:
...
+          #     deployment:
+          #       replicas: 2
+          #     localuserstore:
+          #       enabled: true
+          #       userlist:
+          #       - email: "read_only@metalk8s.invalid"
+          #         hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
+          #         username: "read_only"
+          #         userID: "89b3684b-db28-b7v3-a99c-j99305v2"
+          #     connectors: []
```
alexandre-allard: Not sure we want to keep this here as it does not really help, we should rather have some documentation exposing the CSC defaults.

salt/_modules/metalk8s_service_configuration.py
Ebaneck: Not sure it is possible to have an empty dict so i will remove

salt/metalk8s/addons/prometheus-operator/deployed/service-configuration.sls
Ebaneck: Ok, going to open a documentation ticket to track this
Here is the ticket: https://github.com/scality/metalk8s/issues/2522

salt/_modules/metalk8s_service_configuration.py
Ebaneck: Done

salt/_modules/metalk8s_service_configuration.py
Ebaneck: Done

salt/_modules/metalk8s_service_configuration.py
Ebaneck: Let's go with `CSC`
Done

alexandre-allard (approved):
LGTM,

Issue #2530, opened 2020-05-06
Upgrading a cluster with deployed workloads may fail due to disallowed disruptions


gdemonet:
**Component**: salt, kubernetes, lifecycle

**What happened**:

When running the `upgrade.sh` script on a multi-nodes cluster with some workload (here, Keycloak) deployed on it, the rolling upgrade failed with:

```
[ERROR   ] An exception occurred in this state: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/state.py", line 1981, in call
    **cdata['kwargs'])
  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1977, in wrapper
    return f(*args, **kwargs)
  File "/var/cache/salt/master/extmods/states/metalk8s_drain.py", line 25, in node_drained
    res = __salt__['metalk8s_kubernetes.node_drain'](name, **kwargs)
  File "/var/cache/salt/master/extmods/modules/metalk8s_drain.py", line 532, in node_drain
    return drainer.run_drain(dry_run=dry_run)
  File "/var/cache/salt/master/extmods/modules/metalk8s_drain.py", line 381, in run_drain
    [pod['metadata']['name'] for pod in remaining_pods]
CommandExecutionError: Failed to evict pod "keycloak-0" in namespace "keycloak": (429)
Reason: Too Many Requests
HTTP response headers: HTTPHeaderDict({'date': 'Wed, 06 May 2020 10:14:20 GMT', 'content-length': '322', 'content-type': 'application/json', 'cache-control': 'no-cache, private'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Cannot evict pod as it would violate the pod's disruption budget.","reason":"TooManyRequests","details":{"causes":[{"reason":"DisruptionBudget","message":"The disruption budget keycloak needs 1 healthy pods and has 1 currently"}]},"code":429}
```

The custom drain fails with an uncaught error because some DisruptionBudget doesn't allow the eviction: the previous Node we had to upgrade was just uncordoned when we try to drain the following Node, hence workloads on the previous Node weren't back up yet.

**What was expected**:

For the upgrade script not to fail due to instability it induces by draining Nodes in a rolling fashion.

**Steps to reproduce**:

[...]

PR #2531, opened 2020-05-07
salt/drain: Handle retryable errors when evicting

gdemonet:
The Eviction API sends '429 Too Many Requests' when a requested eviction
can't be applied due to some disruption budget. This way, the client can
wait and retry later.
We didn't handle this error in our implementation, hence rolling
upgrades were failing as soon as we hit this situation.
We add support for this, and also move the "timeout" scope to the whole
eviction process (since we now may be stuck at the eviction creation).

Fixes: #2530

---

Closes: #2530

### DIFF
--- a/salt/_modules/metalk8s_drain.py
+++ b/salt/_modules/metalk8s_drain.py
@@ -359,6 +359,9 @@ class Drain(object):
             )
 
         if dry_run:
+            # Would be nice to create the Eviction in dry-run mode and see if
+            # we hit some 429 Too Many Requests (because a disruption budget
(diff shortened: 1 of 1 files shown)


gdemonet:
Tested with a 3 workers deployment of Zenko, didn't get any issue with the drain during an upgrade!

TeddyAndrieux (approved):
I would have liked more debug log but LGTM

gdemonet:
> I would have liked more debug log but LGTM

You're right, I'll open a follow-up PR.

Issue #2533, opened 2020-05-07
Upgrading the Bootstrap Node may fail if kube-apiserver is too slow to start


gdemonet:
**Component**: salt

**What happened**:

After upgrading the Bootstrap node (during the execution of the `deploy_node` orchestrate triggered by the `upgrade.sh` script), the uncordoning fails with `ConnectionRefused` against the Bootstrap's `kube-apiserver`.

**What was expected**:

This orchestration should not fail because of the changes it creates.

In this case, the particular error shouldn't even be a problem: we are waiting for an API server to answer by asking through the local proxy (:7443). However, since the kubeconfig we use references the local instance of `kube-apiserver` (:6443), this one may not be ready yet (if others were, the proxy would redirect requests to them).

**Steps to reproduce**:

Run the upgrade script on a cluster with highly-available control plane, it may fail when upgrading the Bootstrap. It could also work if things are fast enough to start.

**Resolution proposal** (optional):

Salt Master shouldn't use the generated `admin.conf`, which is intended for direct access to a specific master's instance of `kube-apiserver` (it references `https://<control_plane_ip>:6443`).
Instead, we should generate another kubeconfig for Salt Master, which should point to its local `apiserver-proxy` (so, `https://127.0.0.1:7443`).

To make things clean, we should also generate a custom certificate for this kubeconfig, with:
- `CN: "salt-master-{{ node_name }}"` (username)
- `O: "system:masters"` (group)

In the future, we should use a different group (or set of groups, depending on what (Cluster)Role(s) MetalK8s will provide by default), in the same spirit as #1775.

PR #2534, opened 2020-05-07
salt: Generate custom kubeconfig for Salt Master

gdemonet:
Instead of using the `/etc/kubernetes/admin.conf` file which points to a
specific master's `kube-apiserver` instance, we generate another one
dedicated to Salt Master, and configure it to point to the local
`apiserver-proxy` (which can then route to other masters if the local
one isn't available).

In addition, this kubeconfig generates its own certificate, which could
later map to another group (and thus, other (Cluster)Role(s) than the
current "system:masters").

Note that we reduce the unneeded `/etc/kubernetes` mount to
`/etc/kubernetes/pki` (for SA signing key and etcd encryption key) in
both Salt Master and Salt API containers (SaltAPI didn't need it since
41ba749).

Fixes: #2533


gdemonet:
This doesn't work with upgrade: obviously, haven't run the `metalk8s.salt.master.kubeconfig` formula yet...

Trying a refactor, moving this PR back to draft state.

TeddyAndrieux (dismissed):

TeddyAndrieux (dismissed):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -36,39 +26,43 @@ def node_info(node, ca_minion):
 
     roles = set()
 
-    if VERSION_LABEL in node.metadata.labels:
-        result['version'] = node.metadata.labels[VERSION_LABEL]
+    if VERSION_LABEL in node['metadata']['labels']:
+        result['version'] = node['metadata']['labels'][VERSION_LABEL]
 
-    for (label, value) in node.metadata.labels.items():
+    for (label, _) in node['metadata']['labels'].items():
```
gdemonet: `.keys()`?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -36,39 +26,43 @@ def node_info(node, ca_minion):
...
     try:
-        namespace = coreV1.read_namespace(name="kube-system")
-    except (ApiException, HTTPError) as exc:
+        namespace = __salt__['metalk8s_kubernetes.get_object'](
+            name="kube-system",
+            kind="Namespace",
+            apiVersion="v1",
+            kubeconfig=kubeconfig
+        )
+    except Exception as exc:  # pylint: disable=broad-except
```
gdemonet: It's an execution module, so I guess it should only raise a `CommandExecutionError`?

salt/_pillar/metalk8s_nodes.py
```diff
@@ -156,21 +144,28 @@ def ext_pillar(minion_id, pillar, kubeconfig):
...
+            )
+        except Exception as exc:
+            log.exception(
+                "Failed to retrieve nodes for ext_pillar", exc_info=exc
+            )
+            pillar_nodes = __utils__['pillar_utils.errors_to_dict']([
+                "Failed to retrieve NodeList: {!s}".format(exc)
+            ])
+        else:
+            log.info("Successfully retrieved nodes for ext_pillar")
```
gdemonet: I put this for debug purposes, maybe it should be removed or at least toned down to `debug` level

gdemonet (commented):
Few things to adjust, will take care of them tomorrow. Thanks a lot @TeddyAndrieux for taking care of this!

TeddyAndrieux (approved):

PR #2536, opened 2020-05-08
Add a sleep between failed eviction attempts, and add more logs to drain module

gdemonet:
**Component**: salt, kubernetes

**Context**: See #2530, https://github.com/scality/metalk8s/pull/2531#pullrequestreview-407585510

**Summary**:

- Add a `time.sleep(5)` after 429 failures as `kubectl` does
- Add some debug/info logs to the drain module to make it easier to troubleshoot when things go wrong

---

Closes: #2530

### DIFF
--- a/salt/_modules/metalk8s_drain.py
+++ b/salt/_modules/metalk8s_drain.py
@@ -8,6 +8,7 @@ This module's functions are merged into the `metalk8s_kubernetes`
 module when called by salt by virtue of its `__virtualname__` attribute.
 '''
 
+import json
 import logging
 import operator
 import time
@@ -345,6 +346,7 @@ class Drain(object):
         Returns: string message
         Raises: CommandExecutionError in case of timeout or eviction failure
         '''
+        log.debug("Beginning drain of Node %s", self.node_name)
         try:
             pods = self.get_pods_for_eviction()
         except DrainException as exc:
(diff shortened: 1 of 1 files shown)


salt/_modules/metalk8s_drain.py
```diff
@@ -345,6 +346,7 @@ def run_drain(self, dry_run=False):
         Returns: string message
         Raises: CommandExecutionError in case of timeout or eviction failure
         '''
+        log.debug("Beginning drain of Node {}".format(self.node_name))
```
TeddyAndrieux: Do not use `format` when logging please :)

slaperche-scality: Because it allows lazy evaluation of the formatting.

Cf. https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation

We should probably add it here https://metal-k8s.readthedocs.io/en/latest/developer/development/python.html

gdemonet: Done.

I'm not convinced this kind of change/policy brings much value however, except maybe in the number of characters used. Laziness can be useful, but in our situation here (and most often), it really isn't.
What about consistency? Why not always use the curly braces from `.format` instead of sometimes use printf-style `%s`?

slaperche-scality: It's useful when you have costly formatting: you don't pay for it when the log level is disabled.
May not be the case here, but again if we want to be consistent I think it's better to have:
- `%` for log
- `format` everywhere else

rather than:
- `%` for costly logging (for some value of costly, everyone can use its own definition, and don't forget to change the formatting method as the code evolve if the log becomes more costly because you add context)
- `format` in the other cases

Now, Salt being already slow as hell it probably doesn't matter indeed…

gdemonet: Makes sense, I'm not sure formatting logs would really be the costliest part of what we do, but as a general practice it does ensure "optimal" consistency. Let's indeed put it in our Python Best Practices.

TeddyAndrieux (approved):
:+1:

PR #2537, opened 2020-05-11
salt: catch TypeError in get_service_endpoints()

alexandre-allard:
**Component**: salt

**Context**: 

**Summary**:

**Acceptance criteria**:


gdemonet (approved):
OK for a quick fix, but it'd be better to split this method and explain what is not defined in K8s API to the user (maybe there's something wrong on his side, and it's not just a transient error).

PR #2539, opened 2020-05-11
Allow customization of Prometheus rules (node-exporter)

### LINKED ISSUE #2448: Allow customization of pre-defined Prometheus alert rules

**Why this is needed**:
To allow the user to tune some alert rules threshold to better meet his needs.

**What should be done**:
Customize pre-defined Prometheus alert rules from the user input.  

In order to allow the user to customize the alert rules, we will pick up
some of them (for now, we'll only expose node-exporter alert rules) and expose only few parts of their configurations (e.g. threshold) to be customized.

Since in `Prometheus` rules, there are duplicated `group name` + `alert rule
name`, we also need to take the `severity` into account to understand which
specific alert we're editing.

These customization will be stored in the ``metalk8s-prometheus-config``
`ConfigMap` with something like::

    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: metalk8s-prometheus-config
    data:
      config.yaml: |-
        apiVersion: addons.metalk8s.scality.com
        kind: PrometheusConfig
        spec:
          deployment:
            replicas: 1
          rules:
            <alertGroupName>:
              <alertName>:
                warning:
                  threshold: 30
                critical:
                  threshold: 10
            <anotherAlertGroupName>:
              <anotherAlertName>:
                critical:
                  threshold: 20
                  anotherThreshold: 10

The `PrometheusRules` object manifests
`salt/metalk8s/addons/prometheus-operator/deployed/chart.sls` need
to be templatized to consume these customizations through Jinja.

Default values for customizable alert rules to fallback on, if not defined
in the `ConfigMap`, will be set in ``salt/metalk8s/defaults.yaml``.

**Test plan**:
Using pytest-bdd, we need to add a new feature alerting that will edit a pre-defined Prometheus rule and then check if the modification has been successfully applied.

alexandre-allard:
**Component**: salt, charts

**Context**: 

**Summary**:

**Acceptance criteria**: Tests are OK

---

Closes: #2448


tests/post/steps/test_service_configuration.py
```diff
@@ -1,37 +1,24 @@
 from ast import literal_eval
+import json
 import yaml
+import requests
```
Ebaneck: Minor: might want to **order** them?

tests/post/steps/test_service_configuration.py
```diff
@@ -238,15 +261,35 @@ def load(self, full_csc, name, namespace):
...
 
+def _query_prometheus_api(host, route):
+    cp_ip = _get_local_grain(host, 'metalk8s:control_plane_ip')
+
+    return requests.get(
+        'https://{ip}:8443/api/prometheus/api/v1/{route}'
+        .format(ip=cp_ip, route=route),
+        verify=False,
+    )
+
```
Ebaneck: We should refactor this particular method and the following to reduce duplication.
We have the same method in `test_monitoring.py`

We could keep them in  `utils` then call them from there just like `utils.get_dict_element`

Ebaneck (changes requested):
Nicely done but still has minors to fix!

charts/render.py
```diff
@@ -166,6 +166,23 @@ def replace_magic_strings(rendered_yaml):
...
+            for rule in group.get('rules', [])[:]:
+                if rule.get('alert') in drop_rules[group['name']] \
+                       or rule.get('record') in drop_rules[group['name']]:
+                    group['rules'].remove(rule)
+        if group.get('rules'):
+            groups.append(group)
+
+    template['spec']['groups'] = groups
+
+    return template
```
gdemonet: Do we really need to make this a mutating method? A pure variant (which can directly handle returning `None` if there's no more rules defined) would be simpler/safer I think.

charts/render.py
```diff
@@ -166,6 +166,23 @@ def replace_magic_strings(rendered_yaml):
...
+    groups = []
+
+    for group in template.get('spec', {}).get('groups', []):
+        if group.get('name') in drop_rules:
+            for rule in group.get('rules', [])[:]:
+                if rule.get('alert') in drop_rules[group['name']] \
+                       or rule.get('record') in drop_rules[group['name']]:
+                    group['rules'].remove(rule)
+        if group.get('rules'):
+            groups.append(group)
```
gdemonet: Using many `.get()` in the various `if`/`for` statements makes it not super readable IMHO... Let's try something with more variables to make it more obvious:

```python
existing_groups = template.get('spec', {}).get('groups', [])
for group in existing_groups:
    group_rules = group.get('rules', [])
    new_rules = group_rules[:]
    to_drop = drop_rules.get(group.get('name'), [])
    if to_drop:
        for rule in group_rules:
            if any(rule.get(key) in to_drop for key in ['alert', 'record']):
                new_rules.remove(rule)
    if new_rules:
        groups.append(dict(group, rules=new_rules))
```

Of course, we could also make it work with `filter` (but the benefit isn't obvious, matter of taste):

```python
    new_rules = list(filter(
        lambda rule: not any(
            rule.get(key) in to_drop for key in ['alert', 'record']
        ),
        group_rules
    ))
```

charts/render.py
```diff
@@ -236,11 +269,16 @@ def main():
     )
     sys.stdout.write('\n')
 
+    charts = []
```
gdemonet: It's more a list of `manifests` or `docs` than a list of `charts` (we are reading the templated output of **one** chart).

tests/post/steps/test_service_configuration.py
```diff
@@ -92,30 +78,26 @@ def update_service_configuration(
...
-        DEFAULT_CSC_CONFIG_YAML.format(replicas=value)
-    )
+    csc_obj = csc.get(name, namespace)
+    csc_content = csc.load(csc_obj, name, namespace)
+
+    current = csc_content
+    elements = path.split('.')
+    for element in elements[:-1]:
+        current = current.setdefault(element, {})
+    current[elements[-1]] = value
```
gdemonet: don't we have this defined in a helper (`utils.set_dict_element` or something) already?

tests/post/steps/test_service_configuration.py
```diff
@@ -92,30 +78,26 @@ def update_service_configuration(
     value
 ):
 
-    csc_from_yaml = yaml.safe_load(
-        DEFAULT_CSC_CONFIG_YAML.format(replicas=value)
-    )
+    csc_obj = csc.get(name, namespace)
+    csc_content = csc.load(csc_obj, name, namespace)
```
gdemonet: Sorry it's probably just existing code, but what are those two methods doing? Why not just have one?

tests/conftest.py
```diff
@@ -32,6 +32,7 @@ def version(request, host):
             'source %s && echo $VERSION', str(product_path)
         )
 
+
 @pytest.fixture(scope="module")
 def hostname(host):
```
gdemonet, line 40: We'd need to rename this fixture, since it's the minion ID, not hostname... For later.

tests/utils.py
```diff
@@ -180,3 +178,13 @@ def set_dict_element(data, path, value, delimiter='.'):
...
     (get_dict_element(data, path) if path else data)[key] = value
     return data
+
+
+def get_grain(host, key):
+    with host.sudo():
+        output = host.check_output(
+            'salt-call --local --out=json grains.get "{}"'.format(key)
+        )
+        grain = json.loads(output)['local']
```
gdemonet, line 190: Can be done outside of the `host.sudo()` context

tests/post/steps/test_service_configuration.py
```diff
@@ -180,6 +186,47 @@ def restore_csc(
     )
 
 
+@then(parsers.parse(
+    "we have an alert rule '{rule_name}' in group '{group_name}' with "
+    "'{path}' equal to '{value}'"
+))
```
gdemonet: Would make sense to define this new parser in the same commit that uses it.

tests/post/steps/test_service_configuration.py
```diff
@@ -180,6 +186,47 @@ def restore_csc(
...
+def check_prometheus_alert_rule(
+        control_plane_ip, rule_name, group_name, path, value):
+    """Retrieve an alert rule from the Prometheus API, then
+    checks if the given path matches the value.
+    """
+
+    def _wait_prometheus_config_reload():
+        response = requests.get(
+            'https://{}:8443/api/prometheus/api/v1/rules'.format(
```
gdemonet: If you're considering alert rules only (from the parser text), shouldn't you filter this query? `/api/v1/rules?type=alert`

tests/post/steps/test_service_configuration.py
```diff
@@ -180,6 +186,47 @@ def restore_csc(
...
+                control_plane_ip
+            ),
+            verify=False
+        )
```
gdemonet: Configuration reload is handled without a Pod restart (I think Prom has support for both `SIGHUP` and a `/-/reload` API, probably `prometheus-operator` handles this)? No chance of getting HTTP errors when doing this query?

gdemonet (changes requested):
Design is good (except maybe putting the "drop rules" feature in `charts/render.py`, but let's tackle this later), mostly code polishing left now :)

tests/post/steps/test_service_configuration.py
alexandre-allard: Yes, but it'll not work if the key does not already exist.

charts/render.py
alexandre-allard: true

tests/post/steps/test_service_configuration.py
alexandre-allard: Yes, it's already existing code and I agree that it should be done in the same method.
I wanted to change this, but I forgot. Will do it.

tests/utils.py
alexandre-allard: Does not seem to work, at least on vagrant box.

tests/post/steps/test_service_configuration.py
alexandre-allard: I don't know, maybe, didn't had issue during my tests.

tests/post/steps/test_service_configuration.py
alexandre-allard: Totally.

charts/render.py
alexandre-allard: Haha that's what I was doing at the beginning, but though it would be more efficient to directly edit the dict, that's why there is still this return here which is useless for the moment. Will change this.

charts/render.py
alexandre-allard: Seems readable enough for you to understand. :D
Ok, will change.

tests/post/steps/test_service_configuration.py
alexandre-allard: I tried, but IIRC it does not work.

tests/utils.py
gdemonet: Oh, strange. OK let's keep it like this and investigate later.

tests/conftest.py
TeddyAndrieux: Nop, this one return the real hostname, not the minion ID you have another one just bellow `nodename` to retrieve the minion ID/node name (but maybe this "hostname" one is no longer needed)

gdemonet: :rofl: got tricked by the collapsed lines :facepalm: please ignore!

tests/post/steps/test_service_configuration.py
gdemonet: Maybe it should be fixed?

charts/render.py
gdemonet: Efficiency really isn't a concern here :P Whereas non-mutating methods are easier to test ;)

tests/post/steps/test_service_configuration.py
```diff
@@ -63,15 +54,14 @@ def check_csc_configuration(
...
 
-    assert csc_response, (
+    assert csc_obj, (
         "No ConfigMap with name {} in namespace {} found".format(
             name, namespace
         )
     )
 
-    csc_obj = csc.load(csc_response, name, namespace)
     return dict(csc_obj=csc_obj)
```
gdemonet: Why do we return here? And why wrap it in a higher-level dict? (not your code again :/ but since you're refactoring...)

tests/post/features/service_configuration.feature
```diff
@@ -9,3 +9,14 @@ Feature: Cluster and Services Configurations
...
+
+    Scenario: Customization of pre-defined Prometheus rules
+        Given the Kubernetes API is available
+        And pods with label 'app=prometheus' are 'Ready'
+        And we have a 'metalk8s-prometheus-config' CSC in namespace 'metalk8s-monitoring'
+        When we update 'metalk8s-prometheus-config' CSC in namespace 'metalk8s-monitoring' 'spec.rules.node_exporter.node_filesystem_space_filling_up.warning.hours' to '48'
+        And we apply the 'metalk8s.addons.prometheus-operator.deployed.prometheus-rules' salt state
+        Then we have an alert rule 'NodeFilesystemSpaceFillingUp' in group 'node-exporter' with 'annotations.summary' equal to 'Filesystem is predicted to run out of space within the next 48 hours.'
+        And we restore original state of 'metalk8s-prometheus-config' CSC in namespace 'metalk8s-monitoring'
+        And we apply the 'metalk8s.addons.prometheus-operator.deployed.prometheus-rules' salt state
```
gdemonet: Oh, sorry I didn't catch this before! It's really not a good way of doing cleanup: the `given` used to find a CSC should register teardown code for resetting to the default state. Same goes for the `then` used to trigger the update with Salt, I think. Let's keep it this way for now, as we really wish to get rid of BDD, but it's not a safe approach (if the first `then` assertion fails, you will not run other `then`s, hence will not run your teardown logic).

tests/post/steps/test_service_configuration.py
gdemonet: Actually, could be smart to return the `csc_obj` as a fixture and consume it from the `when`. And change the `ClusterServiceConfiguration` class to manage a single instance of CSC (not just a client), giving it methods to handle teardown at different fixture scopes... For the future ;)

tests/utils.py
```diff
@@ -180,3 +180,54 @@ def set_dict_element(data, path, value, delimiter='.'):
...
+
+    return grain
+
+
+class PrometheusApiError(Exception):
+    pass
+
+
+class PrometheusApi():
+    def __init__(self, host, port=CONTROL_PLANE_INGRESS_PORT):
```
gdemonet: If we really want a default value, maybe use the default Prometheus port (9090) and not the Ingress?

tests/utils.py
```diff
@@ -180,3 +180,54 @@ def set_dict_element(data, path, value, delimiter='.'):
...
+        grain = json.loads(output)['local']
+
+    return grain
+
+
+class PrometheusApiError(Exception):
+    pass
+
+
+class PrometheusApi():
```
gdemonet: (nit) No need for the parentheses

tests/post/steps/test_service_configuration.py
```diff
@@ -199,6 +184,45 @@ def restore_csc(
...
+                        break
+                else:
+                    continue
+                break
+        else:
+            pytest.fail(
+                "Alert rule {} in group {} not found".format(
+                    rule_name, group_name
+                )
+            )
```
gdemonet: Since you've created your own helper to manage Prom API, why not give it this logic to find a specific rule? Could be more re-usable.

gdemonet: And just handle comparing values in this test

tests/post/steps/test_service_configuration.py
```diff
@@ -219,7 +243,17 @@ def get(self, name, namespace):
                 "Unable to read {} ConfigMap  in namespace {} with error: {!s}"
                 .format(name, namespace, exc)
             )
-        return response
+
+        try:
+            csc_key = "config.yaml"
+            csc_object = yaml.safe_load(response.data[csc_key])
+        except yaml.YAMLError as exc:
+            raise Exception(
```
gdemonet: As a general practice, only put the expression that is expected to break within a `try` block. This makes it easier for the reader to understand what part of your code is supposed to raise the exception that you're catching with `except`.

In this case, it'd be nice to add another guard against the absence of the `config.yaml` key:

```suggestion
        csc_key = "config.yaml"
        csc_data = response.data.get(csc_key)
        if csc_data is None:
            raise Exception("Missing {} key in ConfigMap".format(csc_key))

        try:
            csc_object = yaml.safe_load(csc_data)
        except yaml.YAMLError as exc:
            raise Exception(
```

gdemonet (dismissed):
Some more :sweat_smile:

tests/utils.py
TeddyAndrieux: I disagree to me you should inherit from `object`, this one is old style Class pretty sure pylint or flake8 say something about this one

tests/post/steps/test_service_configuration.py
alexandre-allard: Yes, why not.

tests/utils.py
slaperche-scality: Guillaume is right, in Python 3 you don't need to inherit from object explicitly.
old-style class are a thing of the past (Python 2).

Edit: see [here](https://stackoverflow.com/a/45062077) for instance.

alexandre-allard: Who are you calling old?! :older_man:

tests/post/steps/test_service_configuration.py
alexandre-allard: Yeah... I know, just copy/pasted what was below, I should have changed it...

tests/utils.py
TeddyAndrieux: Hmm ok you right, we use python3 in pytest

tests/post/steps/test_service_configuration.py
alexandre-allard: Agree, just trying to change as less as possible as it's not related to the ticket linked to this PR.

tests/post/steps/test_service_configuration.py
alexandre-allard: Maybe, didn't know if it was intended or not... but since this function is not used anywhere, I'll change it;

tests/utils.py
gdemonet: > old-style class are a thing of the past (Python 2).

Python 2 is a thing of the past :trollface:

Ebaneck (dismissed):
Not tested entirely!

PR #2541, opened 2020-05-12
Simple kubectl plugin interacting with APIServer and Salt API

### LINKED ISSUE #2489: Write a simple kubectl plugin in Go able to interract with ApiServer and Salt API
**Component**:

'salt', 'kubectl-plugin'

**Why this is needed**:

Centralized CLI

**What should be done**:

Write a really simple Go kubectl plugin to interact with apiserver and Salt API.

`kubectl-metalk8s`

This Go kubectl plugin should be built as part of classic `doit` build.

https://github.com/kubernetes/sample-cli-plugin
https://github.com/kubernetes/cli-runtime

Only have a simple list nodes on apiserver and `test.ping` on Salt API.

It will be written in Go and should **not** use salt binaries at all, only interact directly with Salt API, and should be able to run from outside of the cluster (check design documentation about Centralized CLI for more detail).

Follow kubectl best practice as much as possible.

TeddyAndrieux:
**Component**:

'salt', 'kubectl-plugin'

**Context**: 

Centralized CLI

**Summary**:

First simple command to list nodes and `test.ping` using salt-api

Right now you need to build it manually using `go build` and then you can run it with a kubeconfig like a classic kubectl command (for salt-api you need a bearer token)

```
[root@bootstrap ~]# kubectl metalk8s list-nodes --kubeconfig=/etc/kubernetes/admin.conf 
List of all nodes:
	bootstrap
[root@bootstrap ~]# kubectl metalk8s test-ping --kubeconfig=/etc/kubernetes/admin.conf --token="$id_token"
bootstrap: true
```

---

Fixes: #2489


NicolasT:
Do we need to *copy* the Salt library? We should reorganize our repository to be able to share it...

slaperche-scality:
Since it's a new "codebase", I suggest to setup some linting right from the start.
On another project we've got good result with [`golangci-lint`](https://github.com/golangci/golangci-lint).

Here is the output I get when I reuse this configuration:

```
pkg/cmd/listNodes.go:79: unnecessary leading newline (whitespace)
func (opt *ListNodesOptions) Run() error {

pkg/salt/client.go:29:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self" (golint)
func (self *AsyncJobFailed) Error() string {
^
pkg/salt/client.go:128:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self" (golint)
func (self *Client) LocalAsync(
^
pkg/salt/client.go:159:1: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self" (golint)
func (self *Client) PollJob(
^
pkg/salt/client.go:160:23: method parameter `jobId` should be `jobID` (golint)
        ctx context.Context, jobId string,
                             ^
pkg/salt/client.go:195:8: don't use underscores in Go names; const non_root_error_prefix should be nonRootErrorPrefix (golint)
        const non_root_error_prefix string = "One or more requisite failed"
              ^
pkg/salt/client.go:321:2: don't use underscores in Go names; method parameter is_auth should be isAuth (golint)
        is_auth bool,
        ^
pkg/salt/client.go:377:64: don't use underscores in Go names; method parameter is_auth should be isAuth (golint)
        verb string, endpoint string, payload map[string]interface{}, is_auth bool,
                                                                      ^
pkg/salt/client.go:423:6: func decodeApiResponse should be decodeAPIResponse (golint)
func decodeApiResponse(response *http.Response) (map[string]interface{}, error) {
     ^
pkg/salt/client_test.go:43:3: don't use underscores in Go names; struct field is_auth should be isAuth (golint)
                is_auth  bool
                ^
pkg/salt/token_test.go:22:4: don't use underscores in Go names; var `is_expired` should be `isExpired` (golint)
                        is_expired := token.isExpired()
                        ^
pkg/salt/client.go:194:6: `getStateFailureRootCause` is unused (deadcode)
func getStateFailureRootCause(output interface{}) string {
     ^
cmd/kubectl-metalk8s.go:4: File is not `goimports`-ed (goimports)
        "flag"
pkg/salt/client.go:258:13: ineffectual assignment to `err` (ineffassign)
                response, err = self.doRequest(ctx, verb, endpoint, payload, true)
                          ^
pkg/salt/client.go:137:5: S1009: should omit nil check; len() for nil maps is defined as zero (gosimple)
        if kwarg != nil && len(kwarg) > 0 {
           ^
pkg/salt/client.go:55:25: ST1005: error strings should not be capitalized (stylecheck)
                return nil, fmt.Errorf("Empty CA certificate")
                                      ^
pkg/salt/client.go:94:25: ST1005: error strings should not be capitalized (stylecheck)
                return nil, fmt.Errorf("Salt API only support BearerToken authentication")
```

(yeah most of those comes from my Salt module :sweat_smile:, yet another reason to share it instead of duplicated it: to have less fix to do ^^).

TeddyAndrieux:
> Do we need to _copy_ the Salt library? We should reorganize our repository to be able to share it...

I totally agree, it's in the commit message, but right now I don't know how we will do it and I think we should do it in another PR so that we can start implementing plugin command in parallel

kubectl-metalk8s/pkg/cmd/listNodes.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// NewListNodesOptions provides an instance of ListNodesOptions with default values
+func NewListNodesOptions(streams genericclioptions.IOStreams) *ListNodesOptions {
+	return &ListNodesOptions{
+		IOStreams: streams,
+	}
+}
+
+// NewCmdListNodes provide a cobra command wrapping ListNodesOptions
```
slaperche-scality: `provides` (+s)

kubectl-metalk8s/pkg/cmd/listNodes.go
```diff
@@ -0,0 +1,106 @@
...
+func NewCmdListNodes(streams genericclioptions.IOStreams) *cobra.Command {
+	opt := NewListNodesOptions(streams)
+
+	cmd := &cobra.Command{
+		Use:          "list-nodes",
+		Short:        "Test list all Kubernetes nodes",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			klog.V(5).Info("listNodes called")
```
slaperche-scality: What is `5`?

Maybe we should define a named constant to make it clearer.

kubectl-metalk8s/pkg/cmd/listNodes.go
```diff
@@ -0,0 +1,106 @@
...
+func (opt *ListNodesOptions) Validate() error {
+	klog.V(5).Info("List nodes options validated")
+
+	return nil
+}
+
+// Run the command
+func (opt *ListNodesOptions) Run() error {
+
+	clientset, err := kubernetes.NewForConfig(opt.config)
```
slaperche-scality: why `clientset` instead of `client`?

kubectl-metalk8s/pkg/salt/client.go
```diff
@@ -62,6 +70,51 @@ func NewClient(creds *Credential, caCertData []byte) (*Client, error) {
 	}, nil
 }
 
+// Create a new Salt API client from a given config (basically a kubeconfig)
+func NewForConfig(config *rest.Config) (*Client, error) {
```
slaperche-scality: `NewFromConfig` I guess.

kubectl-metalk8s/pkg/salt/client.go
```diff
@@ -160,7 +160,7 @@ func (self *Client) PollJob(
 	ctx context.Context, jobId string,
 ) (map[string]interface{}, error) {
 	jobLogger := self.logger.WithValues("Salt.JobId", jobId)
-	jobLogger.Info("polling Salt job")
+	jobLogger.V(3).Info("polling Salt job")
```
slaperche-scality, line 163: Ditto, use a constant here.
If 3 is used only for Salt log, then could be called `SALT_LOG_LEVEL`

kubectl-metalk8s/pkg/cmd/testPing.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !opt.wait {
+		fmt.Fprintf(opt.IOStreams.Out, "Test ping sended: JID %s\n", jid)
+		return nil
+	}
+
+	for {
```
slaperche-scality: Looks like a busy loop, no?

I guess in that case we could simply make a synchronous call, no?

kubectl-metalk8s/pkg/cmd/listNodes.go
TeddyAndrieux: It's the name of the struct but yes I should use client

kubectl-metalk8s/pkg/salt/client.go
TeddyAndrieux: Just took the same name as the kubernetes one it's "new client for this config" .
But here we extract informations from config to create a client so... Yes `from` is maybe better anyway here :thinking:

kubectl-metalk8s/pkg/cmd/listNodes.go
TeddyAndrieux: It's a log level you will see this message only if you do a `-v=5` (or more than 5). 
I agree we should define somewhere what should be logged in which level and then maybe have constant for this

kubectl-metalk8s/pkg/cmd/testPing.go
TeddyAndrieux: I hesitated, different call depending on the "wait" parameter or just have a loop waiting for result if wait True.

Not sure for this one, TBD

kubectl-metalk8s/pkg/cmd/listNodes.go
NicolasT: https://github.com/kubernetes/community/blob/1d747b3720c06e795b0ef6be3cfff71fc13e8b34/contributors/devel/sig-instrumentation/logging.md

kubectl-metalk8s/pkg/cmd/testPing.go
```diff
@@ -0,0 +1,123 @@
...
+// Validate provided informations
+func (opt *TestPingOptions) Validate() error {
+	klog.V(5).Info("Test ping options validated")
+
+	return nil
+}
+
+// Run the command
+func (opt *TestPingOptions) Run() error {
+	ctx := context.TODO()
```
NicolasT: This ain't right, we should have a background context that spans command execution.

kubectl-metalk8s/pkg/cmd/testPing.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return nil
+}
+
+func init() {
+	streams := genericclioptions.IOStreams{
+		In: os.Stdin, Out: os.Stdout, ErrOut: os.Stderr,
+	}
+
+	rootCmd.AddCommand(NewCmdTestPing(streams))
```
NicolasT: There's a better way to achieve this: in the top-level `main` (or elsewhere), construct the `rootCmd` and add a subcommand by calling functions which create those `*cobra.Command`s. 

This then allows to pass in values to these command-generation-functions which then removes the need for these `streams` globals you have, and increase test'ability.

kubectl-metalk8s/cmd/kubectl-metalk8s.go
```diff
@@ -0,0 +1,21 @@
...
+	"k8s.io/klog"
+)
+
+func main() {
+	flags := pflag.NewFlagSet("kubectl-metalk8s", pflag.ExitOnError)
+	pflag.CommandLine = flags
+
+	klog.InitFlags(nil)
+
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
NicolasT: There should be no need to do all of this when using Cobra, or at least, it can be done differently.

kubectl-metalk8s/cmd/kubectl-metalk8s.go
```diff
@@ -0,0 +1,21 @@
...
+
+	cmd.Execute()
```
NicolasT: Exceptions are ignored?

kubectl-metalk8s/pkg/cmd/testPing.go
TeddyAndrieux: Yes was one idea but then you will have all the `AddCommand(NewMyCmd(...))` in the top-level `main` when in this approach to add a new command you just create a new go file in `cmd` package and add it to rootCmd in the `init()` function so .... I don't know 

For global streams value I agree, for testability I do not see why ?

Anyway ok I can do it it's pretty simple change

kubectl-metalk8s/cmd/kubectl-metalk8s.go
TeddyAndrieux: To handle glog (klog) flags + Cobra one's I did not find any other method

kubectl-metalk8s/cmd/kubectl-metalk8s.go
TeddyAndrieux: Exceptions handled by `Execute` function itself (`cmd` here is the Go package cmd not the Cobra command)
```
// Execute adds all child commands to the root command and sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
        if err := rootCmd.Execute(); err != nil {
                fmt.Println(err)
                os.Exit(1)
        }
}

```

kubectl-metalk8s/cmd/kubectl-metalk8s.go
NicolasT: I'll create a fork to explain, will be easier than doing it here.

kubectl-metalk8s/pkg/cmd/testPing.go
NicolasT: The "benefit" of "I can drop a new file and it magically shows up in as a subcommand" isn't worth it to me. Really, 'registering' the subcommand when implementing it isn't a lot of 'work', and allows to parametrize subcommand creation (which is often very useful) :)

kubectl-metalk8s/pkg/cmd/listNodes.go
slaperche-scality: Ho, good to know.

Then those are some well-known values I guess.

kubectl-metalk8s/pkg/cmd/testPing.go
slaperche-scality: And also, explicit is better than implicit (even if we're not in Python).
Especially since the boilerplate is not that big in that case.

kubectl-metalk8s/pkg/cmd/listNodes.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// NewCmdListNodes provide a cobra command wrapping ListNodesOptions
+func NewCmdListNodes(streams genericclioptions.IOStreams) *cobra.Command {
+	opt := NewListNodesOptions(streams)
+
+	cmd := &cobra.Command{
+		Use:          "list-nodes",
+		Short:        "Test list all Kubernetes nodes",
+		SilenceUsage: true,
```
NicolasT: Why?

kubectl-metalk8s/pkg/cmd/testPing.go
TeddyAndrieux: Ok good to me, I agree with you :+1:

kubectl-metalk8s/pkg/cmd/listNodes.go
TeddyAndrieux: Oh ! Cool ! I will refer to it to set level

kubectl-metalk8s/pkg/cmd/listNodes.go
TeddyAndrieux: Because if you do not silence when `RunE` return an error it will show usage even if the error is not related to some arguments.

So either we silence usage either we handle error output ourselves and exit ourselves when running the command. (So basically mean here we always `return nil` in Run function)

TeddyAndrieux: Note that my approach is the same as sample-cli-plugin one (kubectl has a special function to check for err and log it + Exit)

So both doable

Personally I don't really like the "usage" output when you give invalid keyword, argument or whatever but it's only my opinion :smiley:

kubectl-metalk8s/cmd/kubectl-metalk8s.go
NicolasT: https://github.com/scality/metalk8s/compare/improvement/GH-2498-simple-kubectl-plugin...user/nicolast/kubectl-plugin

kubectl-metalk8s/pkg/cmd/listNodes.go
NicolasT: The problem is indeed when `RunE` returns a non-nil `error`, Cobra will show usage.

However, there's a way around to *both* show 'usage' when passing invalid arguments and such, as well as not show 'usage' when input validation has been done: https://github.com/scality/metalk8s/compare/scality:227c0b0...scality:3782a97#diff-8375618fe20c3e0fb1000ac2044b44b6R22

kubectl-metalk8s/pkg/cmd/testPing.go
NicolasT: See https://github.com/scality/metalk8s/compare/scality:227c0b0...scality:3782a97#diff-0f61793fd40f9245e11120c25113edcfR18 and further usage of `cobra.Command.ExecuteContext` as well as `cobra.Command.Context()`

kubectl-metalk8s/pkg/cmd/listNodes.go
NicolasT: https://github.com/scality/metalk8s/compare/improvement/GH-2498-simple-kubectl-plugin...user/nicolast/kubectl-plugin#diff-8375618fe20c3e0fb1000ac2044b44b6R22

Issue #2538, opened 2020-05-11
Extract Prometheus rules tooling


alexandre-allard:
**Component**: tests, documentation

**Why this is needed**:
To be able to retrieve an exhaustive list of the alert rules we deploy (to expose alert rules in the documentation and also to add some tests).

**What should be done**:
A script that extract all Prometheus rules deployed from a running cluster and store them.

**Implementation proposal** (strongly recommended):
Use the Prometheus API, GET on the /api/v1/rules, it returns a json formatted output of all the alert and record rules currently configured in Prometheus.

**Test plan**:

PR #2542, opened 2020-05-12
Tooling to extract deployed Prometheus alert rules + Cluster monitoring doc

Ebaneck:
**Component**:

''tooling', 'basic alerting'

**Context**: 

See #2538 

**Summary**:

This PR brings along the following changes:
- Tooling to read and dump deployed Prometheus alerting and recording rules to file. Dumped files will later be used during testing to assert that rules deployed match exactly what we know or raise so that Dev's can adjust and document new changes.
- Introduce a Cluster monitoring section in the Operation guide and then document default Alerting rules deployed by the Prometheus-operator.
- A PR for BDD testing the rules will follow subsequently

**Acceptance criteria**: 

---

Closes: #2538

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -58,6 +58,12 @@ Glossary
 
      For more information, see `etcd.io <https://etcd.io>`_.
 
+   Grafana
+     Grafana is a service for analysing and visualizing metrics scraped by
+     Prometheus.
+
+     For more information, see `Grafana <https://grafana.com/docs/grafana/latest/>`_.
+
    Kubeconfig
      A configuration file for :term:`kubectl`, which includes authentication
      through embedded certificates.
@@ -70,6 +76,13 @@ Glossary
 
      |see K8s docs|
      `kubelet <https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/>`_.
+
+   Kube-state-metrics
+     The kube-state-metrics service listens to the Kubernetes API server and
+     generates metrics about the state of the objects.
+
+     |see K8s docs|
+     `kube-state-metrics <https://github.com/kubernetes/kube-state-metrics/tree/master/docs/>`_.
 
    Namespace
      A Namespace is a Kubernetes abstraction to support multiple virtual
@@ -105,6 +118,13 @@ Glossary
      provided by the cluster or installed afterwards.
 
      For more details, see `prometheus.io <https://prometheus.io>`_.
+
+   Prometheus Node-exporter
+     The Prometheus node-exporter is an exporter for exposing hardware and
+     OS metrics read from the Linux Kernel. Users can typically obtain the
+     following metrics; cpu, memory, filesystem for each Kubernetes node.
+
+    For more details, see `prometheus node-exporter <https://prometheus.io/docs/guides/node-exporter>`_.
 
    SaltAPI
      SaltAPI is an HTTP service for exposing operations to perform with a
--- a/docs/operation/cluster_monitoring.rst
+++ b/docs/operation/cluster_monitoring.rst
@@ -1,0 +1,60 @@
+Cluster Monitoring
+==================
+
+This section contains information describing the MetalK8s monitoring and
+alerting stack, metric resources that are automatically monitored once MetalK8s
+is deployed, a list of alerting and recording rules which are pre-configured
+and much more.
+
+Monitoring stack
+****************
+
+MetalK8s ships with a monitoring stack that provides a cluster-wide
+view of cluster health, pod status, node status, network traffic status and
+much more. These view points are usually represented as charts,
+counts and graphs. For a closer look, access the
+:ref:`Grafana Service<installation-services-grafana>` to get more insights on
+monitoring stats provided once MetalK8s is deployed.
+
+The MetalK8s monitoring stack consist of the following main components;
+
+  - :term:`Alertmanager`
+  - :term:`Grafana`
+  - :term:`Kube-state-metrics`
+  - :term:`Prometheus`
+  - :term:`Prometheus Node-exporter`
+
+.. todo::
+
+   - For each of the components list above, provide a detail description of
+     its role within the Monitoring stack.
+   - How to customize default alerting & recording rules
+   - Default alerting & recording rules are available as a json file, we should
+     use the json to generate a corresponding rst table as below
+
+Prometheus
+^^^^^^^^^^
+In a MetalK8s cluster, the Prometheus service is responsible for recording
+real-time metrics in a time series database. Prometheus is capable of querying
+a list of datasources called `exporters` at specific polling frequency and then
+aggregating this data across the various sources.
+Prometheus makes use of a special language Prometheus Query Language - PromQL
+for writing alerting and recording rules which we will later see.
+
+Default Alerting rules
+""""""""""""""""""""""
+
+Alerting rules enable a user to specify a condition that must occur before an
+external system like slack is notified. For example, MetalK8s administrators
+could want to raise an alert for any node that is unreachable for a duration
+>1 minutes.
+
+Out-of-the-box, MetalK8s ships with preconfigured alerting rules. These
+alerting rules are typically written as PromQL queries.
+The table below outlines some of the preconfigured alerting rules exposed from
+a newly deployed MetalK8s cluster.
+
+.. csv-table:: Default Prometheus Alerting rules
+   :file: ../../tools/rule_extractor/alerting_rules.csv
+   :header: "Name", "Severity", "Description"
+
--- a/tools/rule_extractor/rule_extractor.py
+++ b/tools/rule_extractor/rule_extractor.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python3
+
+'''
+Utility script to read and write Prometheus Alert rules to file.
+Ensures we can test the Alert rules deployed and for documentation
+purposes.
+'''
+
+import argparse
+import csv
+import ipaddress
+import json
+import pathlib
+import requests
+import subprocess
+import sys
+
+
+def query_prometheus_api(ip, port, route):
+    """Read the Prometheus route contents into memory."""
+    if not route:
+        sys.stderr.write("Expected endpoint route but got {}".format(route))
+        sys.exit(1)
+
+    endpoint = "https://{}:{}/api/prometheus/api/v1/{}".format(ip, port, route)
+    try:
+        response = requests.get(url=endpoint, verify=False)
+    except requests.exceptions.ConnectionError as exc:
+        sys.stderr.write(
+            "Failed to reach endpoint {}: {!s}".format(route, exc)
+        )
+        sys.exit(1)
+
+    if response.status_code != 200:
+        sys.stderr.write(
+            "Expected status code <200> got {}".format(response.status_code)
+        )
+        sys.exit(1)
+
+    return response.json()
+
+
+def write_json_to_file(content, rule_name):
+    filename = '{}.json'.format(rule_name)
+    filepath = pathlib.Path(__file__).parent/filename
+    try:
+        with open('{}'.format(filepath), 'w', encoding='utf-8') as f:
+            json.dump(content, f, indent=4, sort_keys=True)
+    except IOError as exc:
+        sys.stderr.write("Failed to write json file: {!s}".format(exc))
+        sys.exit(1)
+
+
+def write_csv_to_file(content, rule_name):
+    filename = '{}.csv'.format(rule_name)
+    filepath = pathlib.Path(__file__).parent/filename
+    csv_columns = ['name', 'severity', 'message']
+    try:
+        with open('{}'.format(filepath), 'w') as csvfile:
+            writer = csv.DictWriter(
+                csvfile, fieldnames=csv_columns, extrasaction='ignore'
+            )
+            writer.writerows(content)
+    except IOError as exc:
+        sys.stderr.write("Failed to write csv file: {!s}".format(exc))
+        sys.exit(1)
+
+
+def main():
+    """
+        python rule_extractor --ip 127.0.0.1 --port 9090 -t rules
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-i',
+        '--ip',
+        dest='ip',
+        help='Specify the target control-plane ingress IP address'
+    )
+    parser.add_argument(
+        '-p',
+        '--port',
+        dest='port',
+        default='8443',
+        help='Specify the target control-plane ingress port',
+        type=int
+    )
+    parser.add_argument(
+        '-t',
+        required=True,
+        dest='object_type',
+        help='Specify the type of objects to fetch',
+        choices=['rules']
+    )
+
+    args = parser.parse_args()
+    if args.ip:
+        ip = args.ip
+    else:
+        # Obtain VM ip by connecting to it and running a command
+        # Requires that the VM is up and running.
+        ip = subprocess.check_output(
+            [
+                "vagrant",
+                "ssh",
+                "--command",
+                "hostname -I | cut -d' ' -f1"
+            ]
+        ).decode(sys.stdout.encoding).strip()
+    # Validate that we got a real ipv4 address
+    try:
+        ipaddress.IPv4Address(ip)
+    except ValueError:
+        sys.stderr.write("Ip address: {} is invalid".format(ip))
+        sys.exit(1)
+
+    if args.object_type == "rules":
+        result = query_prometheus_api(ip=ip, port=args.port, route="rules")
+        write_json_to_file(result, "rules")
+        rule_group = result.get('data', {}).get('groups', [])
+        recording_rules = []
+        alerting_rules = []
+        for item in rule_group:
+            for rule in item.get('rules', []):
+                # rule type can be alerting or recording
+                # For now, we only need alerting rules
+                if rule['type'] == "alerting":
+                    message = rule['annotations'].get('message') or \
+                        rule['annotations'].get('summary')
+                    fixup_alerting_rule = {
+                        'name': rule['name'],
+                        'severity': rule['labels']['severity'],
+                        'message': message,
+                        'query': rule['query']
+                    }
+                    alerting_rules.append(fixup_alerting_rule)
+
+                elif rule['type'] == "recording":
+                    fixup_record_rules = {
+                        'name': rule['name'],
+                        'query': rule['query']
+                    }
+                    recording_rules.append(fixup_record_rules)
+
+        write_json_to_file(alerting_rules, "alerting_rules")
+        write_csv_to_file(alerting_rules, "alerting_rules")
+    else:
+        sys.stderr.write("Input argument {} is unsupported".format(
+            args.object_type)
+        )
+        sys.exit(1)
+
+
+if __name__ == '__main__':
+    main()
--- a/docs/operation/index.rst
+++ b/docs/operation/index.rst
@@ -20,4 +20,5 @@ do not have a working MetalK8s_ setup.
    volume_management/index
    account_administration
    cluster_and_service_configuration
+   cluster_monitoring
    troubleshooting
--- a/tools/rule_extractor/alerting_rules.csv
+++ b/tools/rule_extractor/alerting_rules.csv
@@ -1,0 +1,102 @@
+AlertmanagerConfigInconsistent,critical,The configuration of the instances of the Alertmanager cluster `{{$labels.service}}` are out of sync.
+AlertmanagerFailedReload,warning,Reloading Alertmanager's configuration has failed for {{ $labels.namespace }}/{{ $labels.pod}}.
+AlertmanagerMembersInconsistent,critical,Alertmanager has not found all other members of the cluster.
+etcdInsufficientMembers,critical,"etcd cluster ""{{ $labels.job }}"": insufficient members ({{ $value }})."
+etcdNoLeader,critical,"etcd cluster ""{{ $labels.job }}"": member {{ $labels.instance }} has no leader."
+etcdHighNumberOfLeaderChanges,warning,"etcd cluster ""{{ $labels.job }}"": instance {{ $labels.instance }} has seen {{ $value }} leader changes within the last hour."
+etcdHighNumberOfFailedGRPCRequests,warning,"etcd cluster ""{{ $labels.job }}"": {{ $value }}% of requests for {{ $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}."
+etcdHighNumberOfFailedGRPCRequests,critical,"etcd cluster ""{{ $labels.job }}"": {{ $value }}% of requests for {{ $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}."
+etcdGRPCRequestsSlow,critical,"etcd cluster ""{{ $labels.job }}"": gRPC requests to {{ $labels.grpc_method }} are taking {{ $value }}s on etcd instance {{ $labels.instance }}."
+etcdMemberCommunicationSlow,warning,"etcd cluster ""{{ $labels.job }}"": member communication with {{ $labels.To }} is taking {{ $value }}s on etcd instance {{ $labels.instance }}."
+etcdHighNumberOfFailedProposals,warning,"etcd cluster ""{{ $labels.job }}"": {{ $value }} proposal failures within the last hour on etcd instance {{ $labels.instance }}."
+etcdHighFsyncDurations,warning,"etcd cluster ""{{ $labels.job }}"": 99th percentile fync durations are {{ $value }}s on etcd instance {{ $labels.instance }}."
+etcdHighCommitDurations,warning,"etcd cluster ""{{ $labels.job }}"": 99th percentile commit durations {{ $value }}s on etcd instance {{ $labels.instance }}."
+etcdHighNumberOfFailedHTTPRequests,warning,{{ $value }}% of requests for {{ $labels.method }} failed on etcd instance {{ $labels.instance }}
+etcdHighNumberOfFailedHTTPRequests,critical,{{ $value }}% of requests for {{ $labels.method }} failed on etcd instance {{ $labels.instance }}.
+etcdHTTPRequestsSlow,warning,etcd instance {{ $labels.instance }} HTTP requests to {{ $labels.method }} are slow.
+TargetDown,warning,"{{ printf ""%.4g"" $value }}% of the {{ $labels.job }} targets in {{ $labels.namespace }} namespace are down."
+Watchdog,none,"This is an alert meant to ensure that the entire alerting pipeline is functional.
+This alert is always firing, therefore it should always be firing in Alertmanager
+and always fire against a receiver. There are integrations with various notification
+mechanisms that send a notification when this alert is not firing. For example the
+""DeadMansSnitch"" integration in PagerDuty."
+AlertmanagerDown,critical,Alertmanager has disappeared from Prometheus target discovery.
+KubeAPIDown,critical,KubeAPI has disappeared from Prometheus target discovery.
+KubeControllerManagerDown,critical,KubeControllerManager has disappeared from Prometheus target discovery.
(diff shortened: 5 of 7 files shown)


tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        try:
+            ipaddress.ip_address(ip)
+        except ValueError:
+            sys.stderr.write("Ip address: {} is invalid".format(ip))
+        except Exception as exc:
+            sys.stderr.write(
+                "Syntax error while checking ip address validity: {!s}".format(
+                    exc
+                )
+            )
```
alexandre-allard: Could be also tested if the IP is passed as argument.
And you should probably exit in error when catching one of these exceptions.
Also, I'm not sure that the bare `Exception` is needed, `ipaddress.ip_address` seems to always return `ValueError` in case of bad argument.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        dest='ip',
+        help='Specify the target control-plane ingress IP address'
+    )
+    parser.add_argument(
+        '-p',
+        '--port',
+        dest='port',
+        default='8443',
+        help='Specify the target control-plane ingress port'
+    )
```
alexandre-allard:
```suggestion
    parser.add_argument(
        '-p',
        '--port',
        dest='port',
        default='8443',
        help='Specify the target control-plane ingress port',
        type=int
    )
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+    parser.add_argument(
+        '-t',
+        '--type',
+        required=True,
+        dest='rule_type',
+        help='Specify the rule type. For example: rules or alerts'
+    )
```
alexandre-allard: Use `choices=[]` to provide a list of possible values for this option, and you should rename this `dest` and `help` as it does not really reflect what it is used for:
```suggestion
    parser.add_argument(
        '-t',
        '--type',
        required=True,
        dest='object_type',
        help='Specify the type of objects to fetch',
        choices=['rules', 'alerts']
    )
```
(not sure for the `objects` name, if you find a more relevant naming...)

docs/operation/cluster_monitoring.rst
```diff
@@ -0,0 +1,240 @@
...
+queries which ensures that MetalK8s administrators no longer have to
+remember complicated expressions which are most of the time computationally
+expensive.
+
+Out-of-the-box, MetalK8s ships with preconfigured recording rules. The
+results of these preconfigured recording rules are saved in the Prometheus
+time series database store.
+
+The contents of :download:`recording rules
+<../../tools/rule_extractor/recording_rule_doc.json>` outlines some of the
```
alexandre-allard: Not sure we want to show these recording rules to the user, not even sure we want to talk about recording rules here.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        out = alertrule.query_prometheus_api(route="rules")
+        alertrule.write_to_file(out, args.rule_type)
+        # For documentation purposes, we don't want to litter the whole thing
+        # with all fields, so lets strip off some "ambiguous" fields
+        rule_group = alertrule.get_dict_element(out, "data.groups")
+        recording_rules = {}
+        alerting_rules = {}
+        for item in rule_group:
+            if 'rules' in item:
+                for rule in item['rules']:
```
alexandre-allard:
```suggestion
            for rule in item.get('rules', []):
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+                    # let's format things nicely
+                    # rule type can be alerting or recording
+                    if rule['type'] == "recording":
+                        fixup_record_rule = {}
+                        fixup_record_rule['name'] = rule['name']
+                        fixup_record_rule['query'] = rule['query']
+                        rule_name = rule['name']
+                        if rule_name not in recording_rules:
+                            recording_rules[rule_name] = {}
```
alexandre-allard-scality, line 131: You can't do that as the `rule_name` is not necessarily unique.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+                    if rule['type'] == "alerting":
+                        fixup_alert_rule = {}
+                        fixup_alert_rule['name'] = rule['name']
+                        fixup_alert_rule['severity'] = \
+                            rule['labels']['severity']
+                        fixup_alert_rule['message'] = \
+                            rule['annotations'].get('message', None)
+                        alert_name = rule['name']
+                        if alert_name not in alerting_rules:
+                            alerting_rules[alert_name] = {}
```
alexandre-allard: You can't do that as the `alert_name` is not necessarily unique. You can overwrite another alert rule.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+                    if rule['type'] == "recording":
+                        fixup_record_rule = {}
+                        fixup_record_rule['name'] = rule['name']
+                        fixup_record_rule['query'] = rule['query']
+                        rule_name = rule['name']
+                        if rule_name not in recording_rules:
+                            recording_rules[rule_name] = {}
+                        recording_rules[rule_name].update(fixup_record_rule)
+
+                    if rule['type'] == "alerting":
```
alexandre-allard: `elif` to avoid having to evaluate the second `if` when the first is evaluated to `True`, or even `else`, because I'm not sure there is another kind of rule type.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        alertrule = AlertRuleFile(ip=args.ip, port=args.port)
+
+    if args.rule_type == "rules":
+        # Returns list of rules that are currently loaded
+        out = alertrule.query_prometheus_api(route="rules")
+        alertrule.write_to_file(out, args.rule_type)
+        # For documentation purposes, we don't want to litter the whole thing
+        # with all fields, so lets strip off some "ambiguous" fields
+        rule_group = alertrule.get_dict_element(out, "data.groups")
```
alexandre-allard: Wouw, that's a bit over engineered, no ?
```suggestion
        rule_group = alertrule.get('data', {}).get('groups', [])
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+
+        return response.json()
+
+    def write_to_file(self, content, rule_name):
+        with open('{}.json'.format(rule_name), 'w', encoding='utf-8') as f:
+            f.write(json.dumps(content, ensure_ascii=True))
+
+    def get_dict_element(self, data, path, delimiter='.'):
+        return functools.reduce(operator.getitem, path.split(delimiter), data)
```
alexandre-allard: IMO, not really relevant to put these 2 methods in this class, at least they should be `@staticmethod`.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+                    # let's format things nicely
+                    # rule type can be alerting or recording
+                    if rule['type'] == "recording":
+                        fixup_record_rule = {}
+                        fixup_record_rule['name'] = rule['name']
+                        fixup_record_rule['query'] = rule['query']
+                        rule_name = rule['name']
+                        if rule_name not in recording_rules:
+                            recording_rules[rule_name] = {}
+                        recording_rules[rule_name].update(fixup_record_rule)
```
alexandre-allard: Could be replaced by:
```suggestion
                        recording_rules.setdefault(rule['name'], {}).update({
                            'name': rule['name'],
                            'query': rule['query'],
                        })
```
But it should even be:
```suggestion
                        recording_rules[rule['name']] = {
                            'name': rule['name'],
                            'query': rule['query'],
                        }
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+
+                    if rule['type'] == "alerting":
+                        fixup_alert_rule = {}
+                        fixup_alert_rule['name'] = rule['name']
+                        fixup_alert_rule['severity'] = \
+                            rule['labels']['severity']
+                        fixup_alert_rule['message'] = \
+                            rule['annotations'].get('message', None)
```
alexandre-allard: This is not true for all rules, sometimes this is stored in `summary`:
```suggestion
                        fixup_alert_rule['message'] = \
                            rule['annotations'].get('message') or \
                            rule['annotations'].get('summary')
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+
+    if args.rule_type == "rules":
+        # Returns list of rules that are currently loaded
+        out = alertrule.query_prometheus_api(route="rules")
+        alertrule.write_to_file(out, args.rule_type)
+        # For documentation purposes, we don't want to litter the whole thing
+        # with all fields, so lets strip off some "ambiguous" fields
+        rule_group = alertrule.get_dict_element(out, "data.groups")
+        recording_rules = {}
+        alerting_rules = {}
```
alexandre-allard: Should be list of rules, you can't index this by name, because alerts can have the same name with different severities, for example.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+import ipaddress
+import json
+import operator
+import requests
+import subprocess
+import sys
+import yaml
+
+
+class AlertRuleFile:
```
alexandre-allard: I don't really see the point of having a class here and it's poorly named, the class is mainly a wrapper around requests on Prometheus API.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+    """A class for manipulating Prometheus Alert & Recording Rules."""
+
+    def __init__(self, ip, port):
+        self.ip = ip
+        self.port = port
+
+    def query_prometheus_api(self, route=False):
+        """Read the Prometheus route contents into memory."""
+        if not route:
+            raise Exception("Expected endpoint route but got None")
```
alexandre-allard: `route` is a string and is defaulted to `False` ? That's a bit weird.
And why trying to handle the case when there is no argument provided, Python can do it itself.
```suggestion
    def query_prometheus_api(self, route):
        """Read the Prometheus route contents into memory."""
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        except requests.exceptions.ConnectionError as exc:
+            raise Exception("Failed to reach endpoint {}: {!s}".format(
+                route, exc
+                )
+            )
+        if response.status_code != 200:
+            raise Exception("Expected status code <200> but got {}".format(
+                response.status_code
+                )
+            )
```
alexandre-allard: Weird indentation:
```suggestion
            raise Exception(
                "Expected status code <200> but got {}".format(
                    response.status_code
                )
            )
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        endpoint = "https://{}:{}/api/prometheus/api/v1/{}".format(
+            self.ip, self.port, route
+        )
+        try:
+            response = requests.get(url=endpoint, verify=False)
+        except requests.exceptions.ConnectionError as exc:
+            raise Exception("Failed to reach endpoint {}: {!s}".format(
+                route, exc
+                )
+            )
```
alexandre-allard:
```suggestion
            raise Exception(
                "Failed to reach endpoint {}: {!s}".format(route, exc)
            )
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+
+import argparse
+import functools
+import ipaddress
+import json
+import operator
+import requests
+import subprocess
+import sys
+import yaml
```
alexandre-allard: never used.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+                    exc
+                )
+            )
+        alertrule = AlertRuleFile(ip=ip, port=args.port)
+    else:
+        alertrule = AlertRuleFile(ip=args.ip, port=args.port)
+
+    if args.rule_type == "rules":
+        # Returns list of rules that are currently loaded
+        out = alertrule.query_prometheus_api(route="rules")
```
alexandre-allard: You should catch exceptions from this function, print a nice error message and exit in error.

tools/rule_extractor/rule_extractor.py
Ebaneck: Yeah, true!

docs/operation/cluster_monitoring.rst
```diff
@@ -0,0 +1,240 @@
...
+    | PrometheusRemoteWriteDesiredShards              | warning      | Prometheus remote write desired shards calculation wants to run more than configured max shards.                    |
+    +-------------------------------------------------+--------------+---------------------------------------------------------------------------------------------------------------------+
+    | PrometheusRuleFailures                          | critical     | Prometheus is failing rule evaluations.                                                                             |
+    +-------------------------------------------------+--------------+---------------------------------------------------------------------------------------------------------------------+
+    | PrometheusMissingRuleEvaluations                | warning      | Prometheus is missing rule evaluations due to slow rule group evaluation.                                           |
+    +-------------------------------------------------+--------------+---------------------------------------------------------------------------------------------------------------------+
```
alexandre-allard: I think it should be automatically generated from the json.
Not sure if there is sphinx extension to do that or if we need to develop one.
Maybe we could also generate an rst from the script below (or another one) and include it here.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        if response.status_code != 200:
+            raise Exception("Expected status code <200> but got {}".format(
+                response.status_code
+                )
+            )
+
+        return response.json()
+
+    def write_to_file(self, content, rule_name):
+        with open('{}.json'.format(rule_name), 'w', encoding='utf-8') as f:
```
alexandre-allard: So, files are generated in the current working directory, not in `tools/rule_extractor/`, I first need to `cd` into this directory to launch this script and update the committed JSON files.
I would like to be able to launch it directly from the root of the repository `./tools/rule_extractor/rule_extractor.py`.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+            f.write(json.dumps(content, ensure_ascii=True))
```
alexandre-allard: I think you should pretty print the resulting JSON file and sort the keys, so the diff will be easier:
```suggestion
            f.write(json.dumps(content, ensure_ascii=True, indent=4, sort_keys=True))
```

slaperche-scality: If you want to ease the diff, you'll want `sort_keys=True` as well.

slaperche-scality: BTW, why do we `ensute_ascii` when we use `UTF-8` as encoding anyway?

docs/glossary.rst
```diff
@@ -58,6 +58,13 @@ Glossary
 
      For more information, see `etcd.io <https://etcd.io>`_.
 
+   Grafana
+     Grafana is a service for analysing and visualizing metrics scraped by
+     Prometheus.
+
+     See also the official Grafana documentation for
+     `Grafana <https://grafana.com/docs/grafana/latest/>`_.
```
alexandre-allard: Grafana documentation for Grafana ?

docs/glossary.rst
```diff
@@ -71,6 +78,13 @@ Glossary
      |see K8s docs|
      `kubelet <https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/>`_.
 
+   Kube-state-metrics
+     The kube-state-metrics service listens to the Kubernetes API server and
+     generates metrics about the state of the objects.
+
+     |see K8s docs|
+     `kube-state-metrics <https://kubernetes.io/docs/tutorials/stateless-application/guestbook-logs-metrics-with-elk/#install-kube-state-metrics/>`_.
```
alexandre-allard: Not sure the link points to kube-state-metrics documentation.

docs/operation/cluster_monitoring.rst
Ebaneck: Ok, going to remove this then

docs/glossary.rst
Ebaneck: Fixed

tools/rule_extractor/rule_extractor.py
Ebaneck: Fixed

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,147 @@
...
+        except ValueError:
+            sys.stderr.write("Ip address: {} is invalid".format(ip))
+            sys.exit(1)
+    else:
+        ip = args.ip
+        try:
+            ipaddress.ip_address(ip)
+        except ValueError:
+            sys.stderr.write("Ip address: {} is invalid".format(ip))
+            sys.exit(1)
```
alexandre-allard-scality, line 102:
```suggestion        
    else:
        ip = args.ip

    # Validate that we got a real ipv4 address
    try:
        ipaddress.ip_address(ip)
    except ValueError:
        sys.stderr.write("Ip address: {} is invalid".format(ip))
        sys.exit(1)
```

alexandre-allard:
```suggestion
    if args.ip:
        ip = args.ip
    else:
        # Obtain VM ip by connecting to it and running a command
        # Requires that the VM is up and running.
        ip = subprocess.check_output(
            [
                "vagrant",
                "ssh",
                "--command",
                "hostname -I | cut -d' ' -f1"
            ]
        ).decode(sys.stdout.encoding).strip()

    # Validate that we got a real ipv4 address
    try:
        ipaddress.ip_address(ip)
    except ValueError:
        sys.stderr.write("Ip address: {} is invalid".format(ip))
        sys.exit(1)
```

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,147 @@
...
+        # We do not need the recording rules at this point in time
+        # write_to_file(recording_rules, "recording_rules")
+        write_to_file(alerting_rules, "alerting_rules")
+    else:
+        # Returns a list of all active alerts if route <alerts> is queried
+        # we do not care about this for now
+        sys.stderr.write("Input argument {} is unsupported".format(
+            args.object_type)
+        )
+        sys.exit(1)
```
alexandre-allard: Why having a mandatory option for something not implemented ?
You should maybe remove this option and this block, no need to complexify this script for nothing.

Ebaneck: Will remove it

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,147 @@
...
+        except ValueError:
+            sys.stderr.write("Ip address: {} is invalid".format(ip))
+            sys.exit(1)
+
+    if args.object_type == "rules":
+        result = query_prometheus_api(ip=ip, port=args.port, route="rules")
+        # For documentation purposes, we don't want to litter the whole thing
+        # with all fields, so lets strip off some "ambiguous" fields
+        # Also we need to test that this generated default alerting rules are
+        # actually same as the deployed alerting rules
```
alexandre-allard: I don't understand the purpose of this comment, it's not related to this script, no tests here.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,147 @@
...
+            [
+                "vagrant",
+                "ssh",
+                "--command",
+                "hostname -I | cut -d' ' -f1"
+            ]
+        ).decode(sys.stdout.encoding).strip()
+        # Validate that we got a real ipv4 address
+        try:
+            ipaddress.ip_address(ip)
```
slaperche-scality: No, this validate you got a valid IP address: it can be either v4 or v6.
- If we want to accept both, we should update the comment.
- If we only v4, the we should update the code (using `ipaddress.IPv4Address(ip)`)

Ebaneck: ok

slaperche-scality: In fact, to be 100% safe you want `ip = str(ipaddress.IPv4Address(ip))` or check `ip == str(ipaddress.IPv4Address(ip))`

Because `IPv4Address` accepts `3232235521` as IP (which is technically valid), but I don't think Prometheus will be happy with that.

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+        sys.exit(1)
+
+    return response.json()
+
+
+def write_to_file(content, rule_name):
+    filename = '{}.json'.format(rule_name)
+    filepath = pathlib.Path(__file__).parent/filename
+    with open('{}.json'.format(filepath), 'w', encoding='utf-8') as f:
+        f.write(json.dumps(content, indent=4, sort_keys=True))
```
TeddyAndrieux: Minor: consider using `json.dump` instead of `write` + `json.dumps`

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+Utility script to read and write Prometheus Alert rules to file.
+Ensures we can test the Alert rules deployed and for documentation
+purposes.
+'''
+
+import argparse
+import ipaddress
+import json
+import pathlib
+from os.path import dirname, abspath
```
alexandre-allard: Not used

tools/rule_extractor/rule_extractor.py
Ebaneck: Done

docs/operation/cluster_monitoring.rst
Ebaneck: Done

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,149 @@
...
+
+def write_csv_to_file(content, rule_name):
+    filename = '{}.csv'.format(rule_name)
+    filepath = pathlib.Path(__file__).parent/filename
+    csv_columns = ['name', 'severity', 'message']
+    try:
+        with open('{}'.format(filepath), 'w') as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=csv_columns)
+            for data in content:
+                writer.writerow(data)
```
alexandre-allard-scality, line 58:
```suggestion
            write.writerows(content)
```

alexandre-allard:
```suggestion
            writer.writerows(data)
```

sayf-eddine-scality: `writer.writerows(content)` you mean ?

tools/rule_extractor/rule_extractor.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def write_json_to_file(content, rule_name):
+    filename = '{}.json'.format(rule_name)
+    filepath = pathlib.Path(__file__).parent/filename
+    with open('{}'.format(filepath), 'w', encoding='utf-8') as f:
+        json.dump(content, f, indent=4, sort_keys=True)
+
+
+def write_csv_to_file(content, rule_name):
```
gdemonet, line 54: CSV? Oh no, the nightmares are coming back...

Ebaneck: We only use this for automatically generating a table in the docs

gdemonet (dismissed):
Sorry, I'm coming after the battle, so maybe we can stick with this approach for now and open debt tickets.

IMO, we're doing it all wrong. The goal of this extraction is actually multiple:
- we want to have generated docs about all Prometheus rules (maybe we only show alerting today, but in a reference guide, I'd rather have them all) deployed by MetalK8s
- we want to have automated tests to help us see when those rules evolve (especially useful if we are upgrading the charts)
- we want to help consumers of those alerts find what they need and how they can fetch them (think about the UI)

So, the current approach only solves (partially) the documentation purpose. We would need more info to understand how the "default alerting rules" behave: what's the `expr`? what are the `labels` attached? maybe some `annotations` could help me filter the alerts I'm interested in?

Also, the CSV approach is way too flat IMO. I understand why it's easier to include in Sphinx, but that shouldn't be a good reason long-term (again, can be a debt ticket). JSON is good, maybe not a flat list if we can use `annotations` on the rules to group them coherently.

Then, with a fully detailed JSON (maybe just save the response from Prom and don't change it at all actually), we can build our Sphinx extension to render it however we want: show only a subset of fields we care about, filter on `type`, ...
We can also build other tooling around it (maybe that's something our UIs can consume and display).
And our tests will be able to pickup subtle changes that a reduced format could miss.

gdemonet:
A few extra comments:

- do we really need a new `tools` directory? IMO, that would make a lot of sense to just implement as a `doit` target (we have Python and dependencies setup), or as a "test" module that we could call directly - a big reason is that I don't expect users (maybe developers would, but a test is fine) to want to use this tool, that's why we have the rules included in the docs
- would be nice to add the test in this same PR, to make it obvious we want to use it this way (it's just a matter of testing two Python dicts equality, shouldn't take too long)
- @ChengYanJin can you have a look at the info in either JSON or rendered docs and see if you're missing details? I think UI will be the first consumer of this info

tools/rule_extractor/rule_extractor.py
gdemonet: Yeah I understand why, I'm just a :trollface: 

However, still thinking this intermediate format just to be able to re-use the existing `csv-table` directive is only OK short-term. I'd rather have our own extension build this table directly from the JSON output of Prometheus (single data source, flexible display).

tools/rule_extractor/rule_extractor.py
alexandre-allard: well, this is obviously obvious. :)

Ebaneck:
> Sorry, I'm coming after the battle, so maybe we can stick with this approach for now and open debt tickets.
> 
> IMO, we're doing it all wrong. The goal of this extraction is actually multiple:
> 
>     * we want to have generated docs about all Prometheus rules (maybe we only show alerting today, but in a reference guide, I'd rather have them all) deployed by MetalK8s
> 
>     * we want to have automated tests to help us see when those rules evolve (especially useful if we are upgrading the charts)
> 
>     * we want to help consumers of those alerts find what they need and how they can fetch them (think about the UI)
> 
> 
> So, the current approach only solves (partially) the documentation purpose. We would need more info to understand how the "default alerting rules" behave: what's the `expr`? what are the `labels` attached? maybe some `annotations` could help me filter the alerts I'm interested in?
> 
> Also, the CSV approach is way too flat IMO. I understand why it's easier to include in Sphinx, but that shouldn't be a good reason long-term (again, can be a debt ticket). JSON is good, maybe not a flat list if we can use `annotations` on the rules to group them coherently.
> 
> Then, with a fully detailed JSON (maybe just save the response from Prom and don't change it at all actually), we can build our Sphinx extension to render it however we want: show only a subset of fields we care about, filter on `type`, ...
> We can also build other tooling around it (maybe that's something our UIs can consume and display).
> And our tests will be able to pickup subtle changes that a reduced format could miss.

Accepted...we need to rework this in a later time.

ChengYanJin (commented):
Hello @gdemonet  
I just took a look at the default alerts. It's not sufficient for us for the moment, e.g. there isn't any alert for latency. But I guess we could preconfigure it.

By saying that, I think we should first define one KPI for Volume. I will create a ticket so @Ebaneck could help work on it.

alexandre-allard (dismissed):
Let's merge this as is and try to improve it later (soon).

alexandre-allard (approved):

Issue #2199, opened 2020-01-21
Customize label selector for `prometheus-operator`-managed ServiceMonitors


gdemonet:
**Component**: monitoring, salt

**Why this is needed**:

When Solutions (and most likely, their Operators) set up metrics scraping by the Prometheus provided by MetalK8s, they can either:

- rely on [Prometheus Service discovery]()
- create `ServiceMonitor` objects

Our current deployment of `prometheus-operator` only looks at `ServiceMonitor`s that match the `release=prometheus-operator` label. This can be too generic if other Prometheus Operators are deployed in the cluster, and doesn't make much sense outside of the Helm chart context.

**What should be done**:

Instead, we should define and use a custom label for these, e.g. `metalk8s.scality.com/monitor=""`.
We should also document this label clearly in Solution developer docs (see #1994).

**Implementation proposal** (strongly recommended):

**Test plan**:

---

Epic: #1994

PR #2544, opened 2020-05-12
Update `operator-sdk`

NicolasT:
This set of patches upgrades the `storage-operator` to `operator-sdk` 0.17.0 (not a fun exercise :sob:)

Likely the biggest change is the code handling leader-election, be advised.

In the end, we end up with a `go.mod` which is somewhat sane, so we may be able to depend on libraries part of the `storage-operator` tree (for now) in other packages.

Closes: #2256.


storage-operator/cmd/manager/main.go
```diff
@@ -116,39 +161,47 @@ func main() {
...
-	log.Info("Starting the Cmd.")
+	// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
+	// necessary to configure Prometheus to scrape metrics from this operator.
+	services := []*v1.Service{service}
 
-	// Start the Cmd
-	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
-		log.Error(err, "Manager exited non-zero")
-		os.Exit(1)
+	// The ServiceMonitor is created in the same namespace where the operator is deployed
```
gdemonet, line 171: Unfortunately, those ServiceMonitors won't be picked up by our prometheus-operator (since it uses the `release=prometheus-operator` label selector, see #2199)... Looked quickly at their code, not sure we can inject custom labels, the `metrics.CreateServiceMonitors` relies on a custom method on `v1.Service` instance called [`GenerateServiceMonitor`](https://github.com/operator-framework/operator-sdk/blob/master/pkg/metrics/service-monitor.go#L73-L104), which basically copies the Service labels for the ServiceMonitor... And the Service created by `metrics.CreateMetricsService` only has one label: `name=<operator-name>`.

So I guess we'd need to patch the ServiceMonitor(s) after creation to add those labels if we wanted a workaround, and probably should open an issue on Operator SDK's tracker.

gdemonet (commented):
Can't speak much for the go.mod changes, only had a small comment about how this SDK manages metrics and how it doesn't fit with our prometheus-operator deployment today. Probably not related to this PR, so feel free to dismiss :)

I'll let @slaperche-scality give his seal of approval ;)

slaperche-scality:
> This set of patches upgrades the `storage-operator` to `operator-sdk` 0.17.0 (not a fun exercise sob)

Seeing how upgrade goes in another project, I'm starting to wonder if this SDK still brings more advantages than pain to us :thinking:

NicolasT:
> > This set of patches upgrades the `storage-operator` to `operator-sdk` 0.17.0 (not a fun exercise sob)
> 
> Seeing how upgrade goes in another project, I'm starting to wonder if this SDK still brings more advantages than pain to us

Agree. Now, these changes are actually somewhat limited, in the end, and I did make the effort of going through every version, which is strictly speaking not required.

Now, for another operator project, I *am* considering to move to Kubebuilder instead... Less 'magic' (but also less features in some regards compared to latest versions of OperatorSDK, like in the testing department).

storage-operator/cmd/manager/main.go
NicolasT: Well, `CreateServiceMonitors` takes optional functions to modify the `ServiceMonitor` object before it being created (https://github.com/operator-framework/operator-sdk/blob/0fa4c824583d60fab2f5c0de3275d6a8bde56982/pkg/metrics/service-monitor.go#L37). So, adding this `release=prometheus-operator` label to them should fix this, right?

NicolasT: Now, in general, I'm not that fond of the *operator* creating its own `ServiceMonitor`, `Service` and whatnot objects *at runtime*.

NicolasT: In any case, this should not 'block' this PR:

- The same issue exists with the 'old' code, so metrics aren't captured today.
- We wouldn't be doing anything *useful* with these metrics anyway for now.
- Solution (whilst feasible to implement, cfr. comment above) would better wait for #2199 instead of now patching in the `release=prometheus-operator` label
- I'd rather 'deploy a `ServiceMonitor` with the operator deployment' and get rid of this 'runtime generation'.

slaperche-scality (dismissed):

NicolasT:
Not fond of https://github.com/scality/metalk8s/pull/2544/commits/3efec35b772388e2c2ceabab611070f46b0124a8 but seems required.

storage-operator/cmd/manager/main.go
```diff
@@ -80,14 +83,28 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Create a new Cmd to provide shared dependencies and start components
-	mgr, err := manager.New(cfg, manager.Options{
+	ctx := context.TODO()
+
+	// Set default manager options
+	options := manager.Options{
 		Namespace:               namespace,
```
slaperche-scality, line 90: > Note: because of bff55c2, the 'standard' leader election mechanism/code
is *not* used.

Now that I think about it, this was fixed so we can probably come back to the standard election mechanism.
See https://github.com/operator-framework/operator-sdk/issues/1305

NicolasT: Maybe in some other PR at some point :) FWIW, a Kubebuilder-scaffolded operator uses this mechanism (`LeaderElection` in the manager's `Options`) by default.

NicolasT:
@bert-e approve

slaperche-scality (approved):

storage-operator/cmd/manager/main.go
gdemonet: Didn't catch the `updaters` argument! So we're fine :)

> I'd rather 'deploy a ServiceMonitor with the operator deployment' and get rid of this 'runtime generation'

Well, I kind of agree in a sense, but also like the idea of considering the Operator as a self-bootstrapping installer... For this specific Operator, it doesn't change much because we deploy it with Salt, so still nothing to do for the user, but for an Operator in the general sense, I understand why this approach is the default.

gdemonet (approved):

PR #2546, opened 2020-05-13
tests: Add tests to validate Prometheus (deployed alerting rules == default alerting rules)

Ebaneck:
**Component**:

'tests', 'basic alerting', 'tooling'

**Context**: 

The following test is added to ensure that we can rightfully validate the deployed Prometheus alerting rules against the defaults at all times. 

See the issue here: #2538 

**Summary**:

Follow suit of PR #2542 as promised.
The guard (Heimdall the Gatekeeper) to catch Prometheus-operator chart upgrades that bring new 
undocumented alerting rules is in place.

**Acceptance criteria**: 

- Rightfully tested!
- Green build

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/tests/post/features/monitoring.feature
+++ b/tests/post/features/monitoring.feature
@@ -38,3 +38,8 @@ Feature: Monitoring is up and running
     Scenario: Node metrics can be retrieved using metrics.k8s.io/v1beta1
         Given the Kubernetes API is available
         Then a node with label 'node-role.kubernetes.io/bootstrap=' has metrics
+
+    Scenario: Ensure deployed Prometheus rules match the default
+        Given the Kubernetes API is available
+        And we have 1 running pod labeled 'prometheus=prometheus-operator-prometheus' in namespace 'metalk8s-monitoring'
+        Then the deployed Prometheus alert rules are the same as the default alert rules
--- a/tests/post/steps/test_monitoring.py
+++ b/tests/post/steps/test_monitoring.py
@@ -1,17 +1,23 @@
-import requests
 import json
-
-import pytest
-from pytest_bdd import scenario, given, then, parsers
+import pathlib
 
 import kubernetes.client
 from kubernetes.client.rest import ApiException
 
-from tests import kube_utils
+import pytest
+from pytest_bdd import scenario, given, then, parsers
+
 from tests import utils
 
+# Constants {{{
 
+ALERT_RULE_FILE_NAME = 'alerting_rules.json'
+ALERT_RULE_FILE_PATH = (pathlib.Path(__file__)/'..'/'..'/'..'/'..'
+            /'tools'/'rule_extractor'/ALERT_RULE_FILE_NAME).resolve()
+
+# }}}
 # Scenarios {{{
+
 
 @scenario('../features/monitoring.feature', 'List Pods')
 def test_list_pods(host):
@@ -39,6 +45,13 @@ def test_pod_metrics(host):
     '../features/monitoring.feature',
     'Node metrics can be retrieved using metrics.k8s.io/v1beta1')
 def test_node_metrics(host):
+    pass
+
+
+@scenario(
+    '../features/monitoring.feature',
+    'Ensure deployed Prometheus rules match the default')
+def test_deployed_prometheus_rules(host):
     pass
 
 
@@ -177,6 +190,51 @@ def node_has_metrics(label, k8s_apiclient):
 
     # Metrics are only available after a while (by design)
     utils.retry(_node_has_metrics, times=60, wait=3)
+
+
+@then(
+    "the deployed Prometheus alert rules are the same as the default "
+    "alert rules")
+def check_deployed_rules(host, prometheus_api):
+    try:
+        deployed_rules = prometheus_api.get_rules()
+    except utils.PrometheusApiError as exc:
+        pytest.fail(str(exc))
+
+    rule_group = deployed_rules.get('data', {}).get('groups', [])
+    deployed_alert_rules = []
+    for item in rule_group:
+        for rule in item.get('rules', []):
+            # rule type can be alerting or recording
+            # For now, we only need alerting rules
+            if rule['type'] == "alerting":
+                message = rule['annotations'].get('message') or \
+                    rule['annotations'].get('summary')
+                fixup_alerting_rule = {
+                    'name': rule['name'],
+                    'severity': rule['labels']['severity'],
+                    'message': message,
+                    'query': rule['query']
+                }
+                deployed_alert_rules.append(fixup_alerting_rule)
+
+    try:
+        with open(ALERT_RULE_FILE_PATH) as f:
+            try:
+                default_alert_rules = json.load(f)
+            except json.JSONDecodeError as exc:
+                pytest.fail("Failed to decode JSON from {}: {!s}".format(
+                    ALERT_RULE_FILE_PATH, exc)
+                )
+    except IOError as exc:
+        pytest.fail(
+            "Failed to open file {}: {!s}"
+            .format(ALERT_RULE_FILE_NAME, exc)
+        )
+
+    assert default_alert_rules == deployed_alert_rules, (
+        "Expected default Prometheus rules to be equal to deployed rules."
+    )
 
 
 # }}}
--- a/tests/post/steps/test_monitoring.py
+++ b/tests/post/steps/test_monitoring.py
@@ -1,6 +1,7 @@
 import requests
 import json
 
+import pytest
 from pytest_bdd import scenario, given, then, parsers
 
 import kubernetes.client
@@ -45,10 +46,11 @@ def test_node_metrics(host):
 # Given {{{
 
 @given("the Prometheus API is available")
-def check_prometheus_api(host):
-    response = _query_prometheus_api(host, 'targets')
-
-    assert response.status_code == 200, response.text
+def check_prometheus_api(prometheus_api):
+    try:
+        prometheus_api.get_targets()
+    except utils.PrometheusApiError as exc:
+        pytest.fail(str(exc))
 
 
 @given(parsers.parse("the '{name}' APIService exists"))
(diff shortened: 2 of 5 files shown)


tests/post/features/monitoring.feature
```diff
@@ -38,3 +38,8 @@ Feature: Monitoring is up and running
     Scenario: Node metrics can be retrieved using metrics.k8s.io/v1beta1
         Given the Kubernetes API is available
         Then a node with label 'node-role.kubernetes.io/bootstrap=' has metrics
+
+    Scenario: Monitor deployed Prometheus rules
+        Given the Kubernetes API is available
+        And we have 1 running pod labeled 'prometheus=prometheus-operator-prometheus' in namespace 'metalk8s-monitoring'
+        Then the deployed Prometheus alerting rules is the same as the default alerting rules
```
gdemonet:
```suggestion
        Then the deployed Prometheus alert rules are the same as the default alert rules
```

tests/post/steps/test_monitoring.py
```diff
@@ -1,6 +1,8 @@
-import requests
 import json
+import pathlib
+import requests
```
gdemonet: Since you're re-ordering imports, `requests` is a third party library, as `pytest` and `kubernetes` are. The usual structure is the following:

```python
# Stdlib imports
import json
import pathlib

# 3rd party imports
import kubernetes.client
import pytests
import requests

# Local imports
from tests import kube_utils
from tests import utils
```

tests/post/features/monitoring.feature
```diff
@@ -38,3 +38,8 @@ Feature: Monitoring is up and running
     Scenario: Node metrics can be retrieved using metrics.k8s.io/v1beta1
         Given the Kubernetes API is available
         Then a node with label 'node-role.kubernetes.io/bootstrap=' has metrics
+
+    Scenario: Monitor deployed Prometheus rules
```
gdemonet: I don't understand what "monitor rules" means...

```suggestion
    Scenario: Ensure deployed Prometheus rules match the default
```

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
     utils.retry(_node_has_metrics, times=60, wait=3)
 
 
+@then(
+    "the deployed Prometheus alerting rules is the same as the default "
+    "alerting rules")
+def monitor_deployed_rules(host):
```
gdemonet: Same comment, this is not about monitoring rules...

```suggestion
def check_deployed_rules(host):
```

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+    """
+        This test is meant to be a guard to catch Prometheus-operator charts
+        upgrade changes that might introduce new alerting rules which warrant
+        some documentation efforts.
+
+        Only a Prometheus alerting rule change should trigger this to fail.
+        When this happens, Developers need to regenerate the default alerting
+        rules using the rule_extractor python script, compare and document the
+        new changes in the Cluster monitoring guide before moving on.
+    """
```
gdemonet: Please comply with [PEP 257](https://www.python.org/dev/peps/pep-0257/), so we ensure consistency across our project. And if you find it hard to remember, maybe add those files to the `lint` stage in CI :wink:?

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+    def _wait_prom_rule_cmp():
+        response = _query_prometheus_api(host, 'rules')
+        assert response.status_code == 200, \
+            "Expect status code <200> but got {} for route rules".format(
+                response.status_code
+            )
+        deployed_rules = response.json()
```
gdemonet: If you wait a bit for #2539, there will be a `PrometheusAPI` helper client to do this for you ;)

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+                # rule type can be alerting or recording
+                # For now, we only need alerting rules
+                if rule['type'] == "alerting":
+                    message = rule['annotations'].get('message') or \
+                        rule['annotations'].get('summary')
+                    fixup_alerting_rule = {
+                        'name': rule['name'],
+                        'severity': rule['labels']['severity'],
+                        'message': message
+                    }
```
gdemonet: Is this the only info we save in the extracted `alerting_rules.json`? If yes, we can't detect if labels changed, if the expr changed... Not super useful like this, but it should at least help us detect when alerts are added/removed.

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+        response = _query_prometheus_api(host, 'rules')
+        assert response.status_code == 200, \
+            "Expect status code <200> but got {} for route rules".format(
+                response.status_code
+            )
+        deployed_rules = response.json()
+
+        # we cannot simply compare the obtained value because it contains
+        # alerts with an "activeAt": "2020-05-15T09:23:15.279979961Z",
+        # so lets filter and use only certain alerting rule fields
```
gdemonet: In the general sense, when you want to ensure two datasets are the same except for a few fields, don't select a subset of the fields you care about. Just ignore the fields you don't care about when comparing.

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+                    fixup_alerting_rule = {
+                        'name': rule['name'],
+                        'severity': rule['labels']['severity'],
+                        'message': message
+                    }
+                    deployed_alerting_rules.append(fixup_alerting_rule)
+
+        filename = 'alerting_rules.json'
+        filepath = (pathlib.Path(__file__)/'..'/'..'/'..'/'..'
+                    /'tools'/'rule_extractor'/filename).resolve()
```
gdemonet: Let's make this a constant outside of the test function.

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+
+        try:
+            with open(filepath) as f:
+                default_alerting_rules = json.load(f)
+        except IOError as exc:
```
gdemonet: Would also need to catch JSON errors ;)

tests/post/steps/test_monitoring.py
```diff
@@ -174,6 +182,69 @@ def _node_has_metrics():
...
+            pytest.fail("Failed to open file {}: {!s}".format(filename, exc))
+
+        assert json.dumps(default_alerting_rules, sort_keys=True) \
+            == json.dumps(deployed_alerting_rules, sort_keys=True), \
```
gdemonet: That works, but I'd suggest you use Python dicts equality, `pytest` automatically displays the diff when the assertion fails.

If you really want to compare a sorted list, you can do:
```python
sorted(
    default_alerting_rules,
    key=lambda rule: hash(json.dumps(rule, sort_keys=True))
)
```

But that's way more complex than required. Instead, just build a hierarchy of group/alerts and compare the final dict:

```python
# this is an example, adapt it to your code
deployed_rules = {
    group['name']: {rule['name']: rule for rule in group.rules}
    for group in prom_data['data']['groups']
}
```

gdemonet (dismissed):

tests/post/steps/test_monitoring.py
Ebaneck: We have decided to also compare the `expr` as you have suggested and as such we have updated the alerting_rules.json to have this field.

For the label key in general, we have just `severity` as the subkey and yes, we do compare the severity. Maybe it would have been better to dump the label structure without altering it.

tests/post/steps/test_monitoring.py
Ebaneck: ok

tests/post/steps/test_monitoring.py
```diff
@@ -1,17 +1,24 @@
...
+import pytest
+from pytest_bdd import scenario, given, then, parsers
+
 from tests import utils
 
+# Constants {{{
 
+FILE_NAME = 'alerting_rules.json'
+FILE_PATH = (pathlib.Path(__file__)/'..'/'..'/'..'/'..'
+            /'tools'/'rule_extractor'/FILE_NAME).resolve()
```
alexandre-allard: Could we use more explicit names here ?

Ebaneck: Done

tests/post/steps/test_monitoring.py
```diff
@@ -179,4 +192,63 @@ def _node_has_metrics():
...
+        assert default_alert_rules == deployed_alert_rules, (
+            "Expected default Prometheus rules to be equal to deployed rules."
+        )
+
+    utils.retry(
+        _wait_prom_rule_cmp,
+        times=5,
+        wait=3,
+        name="checking deployed Prometheus rules are equal to default rules"
+    )
```
alexandre-allard: Not sure we need to implement a retry here since we do not change anything in the step before, just reading the API, the PrometheusAPI wrapper will take care of retrying in case of non 200 exit code.

tests/post/steps/test_monitoring.py
```diff
@@ -179,4 +192,63 @@ def _node_has_metrics():
...
+    def _wait_prom_rule_cmp():
+        try:
+            deployed_rules = prometheus_api.get_rules()
+        except utils.PrometheusApiError as exc:
+            pytest.fail(str(exc))
+
+        if not deployed_rules:
+            pytest.fail(
+                "Expected default deployed Prometheus rules but got {}"
+                .format(deployed_rules)
```
alexandre-allard: Don't see the point of showing the content of `deployed_rules`, with the `if` before, we know its empty.
Moreover, I'm not even sure that this case can happen since it's the return from the API in case of success, so you should at least have the status and the `data` key.
For me, you can remove this whole `if` and let the assertion fails at the bottom of this function.

alexandre-allard (approved):

PR #2549, opened 2020-05-14
Upgrade Salt to `3000.3`

TeddyAndrieux:
**Component**:

'salt', 'build'

**Summary**:

Change salt-master image to salt 3000.3 and download salt-minion version
3000.3

https://docs.saltstack.com/en/latest/topics/releases/3000.3.html


NicolasT:
What would change for us?

TeddyAndrieux:
> What would change for us?

Fixed some bugs in Salt but nothing for us because we are not using the "things" that have been fixed. But anyway always good to have the last patch release

gdemonet (approved):

Issue #2551, opened 2020-05-15
Unable to launch downgrade if label version already set to dest


TeddyAndrieux:
**Component**:

'lifecycle'

**What happened**:

If for whatever reason (downgrade already failed or ....) all the node already have the label version set to the destination version then downgrade fail.

NOTE: The label version is set before downgrading the node so this version does not necessarily represent the running version but the desired version!

```
[root@test-solution ~]# kubectl get node -o go-template='{{range .items}}{{index .metadata.labels "metalk8s.scality.com/version"}}{{end}}{{printf "\n"}}'
2.4.2
[root@test-solution ~]# /srv/scality/metalk8s-2.4.3/downgrade.sh --destination-version 2.4.2
> Performing Pre-Downgrade checks... fail [45s]

Failure while running step 'Performing Pre-Downgrade checks'

Command: precheck_downgrade

Output:

<< BEGIN >>
[ERROR   ] Failure!
bootstrap_master:
----------
          ID: Invalid saltenv "metalk8s-2.4.3" consider using "metalk8s-2.4.2"
    Function: test.fail_without_changes
      Result: False
     Comment: Failure!
     Started: 08:44:54.030307
    Duration: 2.577 ms
     Changes:   
----------
          ID: Node bootstrap already in version 2.4.2
    Function: test.succeed_without_changes
      Result: True
     Comment: Success!
     Started: 08:44:54.033391
    Duration: 0.728 ms
     Changes:   

Summary for bootstrap_master
------------
Succeeded: 1
Failed:    1
------------
Total states run:     2
Total run time:   3.305 ms
time="2020-05-15T08:44:54Z" level=fatal msg="execing command in container failed: command terminated with exit code 1"
<< END >>

This script will now exit
```

**What was expected**:

This pre-check shouldn't fail if the script version is higher than ALL nodes versions

**Steps to reproduce**

Set label `metalk8s.scality.com/version` of all nodes to destination version and try to downgrade

**Resolution proposal** (optional):

Check for saltenv version to be higher than all nodes version and not strictly equal

---

**Workaround**:

Set `metalk8s.scality.com/version` of at least one node to the current running version

E.g.: Downgrading from 2.4.3 to 2.4.2 just set one node version to 2.4.3 and then you can run the downgrade script again

PR #2552, opened 2020-05-15
salt: Downgrade pre-check allow saltenv >= higher node

TeddyAndrieux:
**Component**:

'salt', 'lifecycle'

**Summary**:

Node version represent the version desired on a specific node and not
the actual deployed version so if for any reason a downgrade failed
after all node version being set to the destination one, we want to be
able to run the downgrade again so the saltenv specified may be higher
that all the node versions

---

Fixes: #2551

### DIFF
--- a/salt/metalk8s/orchestrate/downgrade/precheck.sls
+++ b/salt/metalk8s/orchestrate/downgrade/precheck.sls
@@ -4,7 +4,7 @@
 {%- set nodes_versions = pillar.metalk8s.nodes.values() | map(attribute='version') | list %}
 {%- do nodes_versions.sort(cmp=salt.pkg.version_cmp, reverse=True) %}
 {%- set expected = nodes_versions | first %}
-{%- if saltenv != 'metalk8s-' ~  expected %}
+{%- if salt.pkg.version_cmp(saltenv | replace('metalk8s-', ''), expected) >= 0 %}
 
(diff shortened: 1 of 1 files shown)


salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -4,7 +4,7 @@
 {%- set nodes_versions = pillar.metalk8s.nodes.values() | map(attribute='version') | list %}
 {%- do nodes_versions.sort(cmp=salt.pkg.version_cmp, reverse=True) %}
 {%- set expected = nodes_versions | first %}
-{%- if saltenv != 'metalk8s-' ~  expected %}
+{%- if salt.pkg.version_cmp(saltenv, 'metalk8s-' ~ expected) >= 0 %}
```
gdemonet: `version_cmp` on stuff with 'metalk8s-' as a prefix? Wouldn't it make more sense with:

```suggestion
{%- if salt.pkg.version_cmp(saltenv | replace('metalk8s-', ''), expected) >= 0 %}
```

or maybe `saltenv[9:]`, but that's less explicit.

TeddyAndrieux: Does not really change much but ok let's compare just version

gdemonet (dismissed):

gdemonet (approved):
:shipit:

PR #2555, opened 2020-05-17
salt: Fix apiserver-proxy upstreams

gdemonet:
The previous approached relied on the pillar target
`metalk8s:nodes:*:roles:master`, which will always match all minions
since they all have access to the full `metalk8s:nodes` pillar value.
Instead, we use the `metalk8s.minions_by_role` execution method to
define the right target when retrieving control plane IPs from the mine.

We also change the `metalk8s.minions_by_role` method to handle
pillar errors gracefully, and let clients handle an empty list as they wish
(which was already the approach taken in the apiserver-proxy
configuration template, considering an empty result implies we are
bootstrapping).

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -87,6 +87,12 @@ def minions_by_role(role, nodes=None):
             Defaults to `pillar.metalk8s.nodes`.
     '''
     nodes = nodes or __pillar__['metalk8s']['nodes']
+    pillar_errors = nodes.pop('_errors', None)
+    if pillar_errors:
+        raise CommandExecutionError(
+            "Can't retrieve minions by role because of errors in pillar "
+            "'metalk8s:nodes': {}".format(', '.join(pillar_errors))
+        )
 
     return [
         node
--- a/salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
+++ b/salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
@@ -13,10 +13,11 @@ events {
 {%- if pillar.get('apiserver_ip') %}
 {%-     set masters_request = {'ip': pillar.apiserver_ip} %}
 {%- else %}
+{%-     set master_minions = salt['metalk8s.minions_by_role']('master') %}
 {%-     set masters_request = salt['mine.get'](
-            tgt='I@metalk8s:nodes:*:roles:master',
+            tgt=master_minions | join(','),
             fun='control_plane_ip',
-            tgt_type='compound',
+            tgt_type='list',
         )
 %}
 {%- endif %}
--- a/salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
+++ b/salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
@@ -11,21 +11,26 @@ events {
 }
 
 {%- if pillar.get('apiserver_ip') %}
-{%-     set masters_request = {'ip': pillar.apiserver_ip} %}
+{%-     set apiservers = [pillar.apiserver_ip] %}
 {%- else %}
-{%-     set master_minions = salt['metalk8s.minions_by_role']('master') %}
-{%-     set masters_request = salt['mine.get'](
-            tgt=master_minions | join(','),
-            fun='control_plane_ip',
-            tgt_type='list',
-        )
+{%-     set masters = salt['metalk8s.minions_by_role']('master') %}
+{%-     if not masters %}
+            {{ raise("No master nodes in pillar 'metalk8s:nodes', "
+                     ~ "can't deploy apiserver-proxy") }}
+{%-     else %}
+{%-         set master_ips = salt['mine.get'](
+                tgt=masters | join(','),
+                fun='control_plane_ip',
+                tgt_type='list',
+            )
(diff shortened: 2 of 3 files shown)


salt/_modules/metalk8s.py
```diff
@@ -87,6 +87,13 @@ def minions_by_role(role, nodes=None):
             Defaults to `pillar.metalk8s.nodes`.
     '''
     nodes = nodes or __pillar__['metalk8s']['nodes']
+    if '_errors' in nodes:
+        log.debug(
+            "Can't retrieve minions by role because of errors in pillar "
+            "'metalk8s:nodes': %s",
+            ', '.join(nodes['_errors']),
+        )
+        return []
```
TeddyAndrieux: Why you do not raise ? With this we cannot see the difference in output between "no minions with this roles" and "not able to retrieve minions list"

gdemonet: If this method is meant to be used directly at render-time, I'd rather have it not fail... But maybe we should just not call it if there are `_errors` in the pillar, and do those checks at every call site.

TeddyAndrieux: Hmm not sure if something is doable in Salt to catch that kind of raise :thinking: , but what is doable is to return a string if you have an error and a list otherwise so that you just check the result

gdemonet: > return a string if you have an error and a list otherwise

Meh, I'm not fond of this kind of trick to handle errors. What I'll do instead, is let this method raise as you suggested, and make sure to not call it if we expect errors:

```sls
{%- if '_errors' in pillar.metalk8s.nodes %}
    {# Bootstrap #}
    {%- set apiservers = ["127.0.0.1"] %}
{%- else %}
    {%- set masters = salt['metalk8s.minions_by_role']('master') %}
    {%- set master_ips = salt['mine.get'](...) %}
    {%- if not master_ips %}
        {{ raise("No master node in pillar!") }}
    {%- endif %}
    {%- set apiservers = master_ips.values() | sort %}
{%- endif %}
```

TeddyAndrieux: Hmm you sure we have errors during bootstrap ? Because to me we fill the pillar "nodes" properly during bootstrap

gdemonet: There's a moment when this gets rendered and there's no kubeconfig yet, so yeah there are some errors in the pillar (I had a failure with something like `None object has no attribute 'items'` from the `minions_by_role` call).

TeddyAndrieux: Hmm ok weird to me as during bootstrap we override all the pillar contain ...But ok, right

gdemonet: Oh right I know what happened! The error wasn't what I shared, but `list item has not attribute 'get'`. It comes for this piece of code:

```python
return [
    node
    for (node, node_info) in nodes.items()
    if role in node_info.get('roles', [])
]
```

The issue: in `metalk8s:nodes`, even if we pass our own value, it'll get merged, so we still have an `_errors` key in the pillar...

gdemonet: Guess we can trick it by changing slightly to:

```python
nodes = nodes or __pillar__['metalk8s']['nodes']
pillar_errors = nodes.pop('_errors', None)
if pillar_errors is not None:
    raise CommandExecutionError(
        "Can't retrieve minions by role because of errors in pillar "
        "'metalk8s:nodes': {}".format(', '.join(pillar_errors))
    )
```

And override the `_errors` manually when bootstrapping?

TeddyAndrieux: Yes agree with the last one to me we should override errors when bootstrapping, we fill the pillar so we don't care about ext_pillar errors in these sections

gdemonet: Done!

salt/_modules/metalk8s.py
```diff
@@ -87,6 +87,12 @@ def minions_by_role(role, nodes=None):
             Defaults to `pillar.metalk8s.nodes`.
     '''
     nodes = nodes or __pillar__['metalk8s']['nodes']
+    pillar_errors = nodes.pop('_errors', None)
+    if pillar_errors is not None:
```
TeddyAndrieux: Minor but I think if it's an empty list it's ok also :smiley:

salt/metalk8s/kubernetes/apiserver-proxy/files/apiserver-proxy.conf.j2
```diff
@@ -11,20 +11,21 @@ events {
...
+                tgt_type='list',
+            )
 %}
-{%- endif %}
-{%- if not masters_request %}
-{# In this case, we're (likely) bootstrapping #}
-{%-   set apiservers = ['127.0.0.1'] %}
-{%- else %}
-{%-   set apiservers = masters_request.values() | sort %}
+{%-         set apiservers = master_ips.values() | sort %}
```
TeddyAndrieux: You still need to do the bootstrap trick here since mine does not work without master (I think you can check that `grains.id` is in the `masters` list and add 127.0.0.1 in this case (only if mine empty of course) it will be PERFECT :smiling_face_with_three_hearts:)

gdemonet: Ah, right :sweat_smile:

TeddyAndrieux (approved):
LGTM

PR #2556, opened 2020-05-18
Log Centralization Requirements Document

thomasdanan:
**Component**:

'logs', 'kubernetes', 'deployment'

**Context**: 

Log Centralisation Requirements Document

**Summary**:

Main requirements and user stories related to Log service n MetalK8s

**Acceptance criteria**: 

---

Closes: #2574


docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Context
+-------
+
+MetalK8s value is to provide, out of the box, some services in order to ease
+the monitoring and operation of the platform as well as workload running on top
+of it.
+
+MetalK8s provides monitoring service with Prometheus and AlertManager in
+order to expose system performance time series and alerts.
```
gdemonet: "system performance time series" -> "metrics" ? They can be of any kind actually, more than just performance indicators.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The logs generated by the platform and the workloads constitute an essential
+piece of information that can be very helpful when it comes to understand the
+root cause of a failure or a performance degradation. Because of the
+distributed nature of MetalK8s and workloads running of top of it, it is
+important to have some tooling to collect and aggregate all logs in order to
+make them browsable from a central endpoint. It is also important to ease
```
gdemonet: It's not exactly why we need a system like Loki IMHO. We already have a central endpoint to reach all logs of any container in the cluster: K8s API (`kubectl logs` command).
On the other hand, we are missing key features:
- logs are not persisted over time (a container can maintain at most 5MB of logs IIRC), and are host-local (can be lost in case of server loss)
- logs are not unified and cannot be queried / searched (and `kubectl logs` really isn't a great tool for investigating)

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+browsing and correlation with other type of indicators such as KPIs and alerts.
```
gdemonet: This I'm not sure we can do. IMO, the best we can achieve is to have a "source" component (e.g. `cloudserver`) for logs, and check if this "source" has metrics/alerts.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+
+In addition to the application logs, the platform and workload are also
+producing some audit logs. If the same system is used to manage application and
+audit logs, then it is important to provide the end users of the workloads
+some capabilities to secure and query the audit logs belonging to their usage.
```
gdemonet: > "end-users of the workloads [...] secure and query the audit logs belonging to their usage"

I'm a bit lost. A user wants to secure audit logs? But that'd be more for an admin, to make sure that when such "end-users" reach the workload, it appears in the audit logs, and the "end-user" can't remove this line (or even read it).

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Requirements
+------------
+
+- Collect both containers and system logs (STDOUT or STDERR)
```
gdemonet: system logs? What does this mean? Everything managed by `systemd`? Maybe just dump `journalctl`? What about other non-recurrent logs (like the ones we write when performing MetalK8s operations, under `/var/log/metalk8s`)?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+- Persistent storage of logs
+- Horizontally scalable (capacity and performance)
+- Accessible from a central end point
+- Highly Available
+- Easy to use query language
+- Service accessible through a UI
```
gdemonet: "Service" isn't specific enough. You mean "log browsing/search", right? Because that could be provided by Grafana out of the box, but if we're considering other parts of the system, like defining custom parsers for extracting logs of your workloads into Loki, that won't be UI-based.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+In addition to the application logs, the platform and workload are also
+producing some audit logs. If the same system is used to manage application and
+audit logs, then it is important to provide the end users of the workloads
+some capabilities to secure and query the audit logs belonging to their usage.
+
+Requirements
+------------
+
+- Collect both containers and system logs (STDOUT or STDERR)
+- Persistent storage of logs
```
gdemonet: For how long? Also, there's no mention of "downsampling" (doesn't make much sense for logs per-se, but extracting long-term metrics out of logs can be useful data).

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+- Horizontally scalable (capacity and performance)
+- Accessible from a central end point
+- Highly Available
+- Easy to use query language
+- Service accessible through a UI
+- Monitorable service (health, performances and alerts)
+- Ability to correlate logs with KPIs and alerts
+- Multitenancy for audit logs
```
gdemonet: So, multi-tenancy means you want RBAC. Maybe you want different log sources, maybe you want to filter based on roles. But that's not going to be part of Metal's log aggregation system I think, multi-tenancy isn't really our concern, it's only at the application level. If they want multi-tenant S3 audit logs in cloudserver, this will probably need another design and shouldn't be aggregated in the central log collector.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+- Performances (TBD)
+
+
+User Stories
+------------
+
+- As a Platform Administrator, I want to browse all MetalK8s containers logs
```
gdemonet, line 122: Is this log-source RBAC really required? I mean, will a Storage Admin really go down into the container logs if he can't do more than managing the CRs we expose for him?
The thing is, I'm really not sure this feature (RBAC) is available (neither in Loki nor ES).

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+  (from all servers) from a unique endpoint, in order to ease distributed
+  system error investigation
+
+- As a Platform Administrator, I want to browse all system logs (from all
```
gdemonet: Again, what are system logs? How do I group logs? Per service? Per server?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+  the log.
+
+
+Documentation
+-------------
+
+Deployment & Configuration
+--------------------------
+
+indexing log streams instead of indexing log content
```
gdemonet: ?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,90 @@
...
+  (from all servers) from a unique endpoint, in order to ease Hyperdrive error
+  investigation
+
+- As a Storage Administrator, I want to browse all Zenko containers logs (from
+  all servers) from a unique endpoint, in order to ease Zenko error
+  investigation
+
+- As a Storage Administrator, I want to log all actions performed by users
+  (whatever their Role) on the RING X product, in order to be able to sustain
+  an audit
```
gdemonet: We can't put this as a requirement on the log collection system: Loki/ES/anything doesn't care about what the logs are made of, it's the responsibility of the sender to define. So, if you want audit logs to cover all the product, maybe worth another epic.

Also, this will extend the requirement a bit: we said container logs (so, `stdout`/`stderr`) and system logs (to be defined, but probably journald entries). Now, if we consider audit logs (e.g. from kube-apiserver), then we also need to tail arbitrary files written by containers (not just the standard streams).

gdemonet (commented):
Some comments/questions!

docs/developer/architecture/logs.rst
thomasdanan: will change with MetalK8s and workloads metrics and alerts

docs/developer/architecture/logs.rst
thomasdanan: ok changed it a bit ... without going in the details since it is the next section

docs/developer/architecture/logs.rst
thomasdanan: Can't we use labels? and have shared labels between Prometheus KPIs and Logs? For example if we have some label for PUT Object OPS metric, can't we have the same label for logs about Object Put (from cloudserver, or/and other components)? If we ca correlate alert, kpis and logs within a single dashboard it will drastically speed up troubleshooting.

Now Loki is not indexing the content of the logs but only offer the ability to label the different log streams.

docs/developer/architecture/logs.rst
thomasdanan: to query is enough. I wanted to say that a tenant doesn't want his audit logs to be consumable by another tenant. By the way, I have no idea if we want to use Loki for audit logs, maybe I should completely remove this part.

docs/developer/architecture/logs.rst
thomasdanan: I want to talk about things that are not running in containers
At least, kubelet and containerd and as you are saying some MetalK8s logs, salt minion ... but maybe interested to also have /var/log/messages (to see with @xaf-scality and @bkettler ...) will change that

docs/developer/architecture/logs.rst
thomasdanan: yes browsing will specify

docs/developer/architecture/logs.rst
thomasdanan: I don't know yet but it has to be configurable anyway. This should be part of CSC and centralized CLI..
I see some value in extracting some KPIs like nbr of occurrences or frequency of a certain type of log, do you think about something else?

docs/developer/architecture/logs.rst
thomasdanan: okay will remove everything about tenant audit logs

docs/developer/architecture/logs.rst
thomasdanan: Don't want to have any kind of RBAC for logs, just saying that only users having the Platform and Storage Admin role will use the service. Other types of users will never need to access those logs. This is how the US are written in the service map

docs/developer/architecture/logs.rst
thomasdanan: Here I am talking about the System so maybe I should say Platform? I don't know how we should group logs ... you mean how we should label logs? to ease the browsing?

docs/developer/architecture/logs.rst
thomasdanan: to be removed ... it is a draft :) but this is the key difference between Loki and Kibana/ES. I wonder if it is going to be a problem or not for workload logs browsing

docs/developer/architecture/logs.rst
thomasdanan: Ok I removed everything about audit logs. Again, those US are the ones from the RINGX service map, just to give a little bit of context on what we need at a higher level

docs/developer/architecture/logs.rst
gdemonet: Right, we have arbitrary labels, so indeed, we can use them to provide correlation. Good point.

docs/developer/architecture/logs.rst
gdemonet: Yes, the requirements for audit logs on various components would need another design, because they do not serve the same purpose (the ones we consider here are there for monitoring/troubleshooting, whereas audit can have many other uses).

docs/developer/architecture/logs.rst
gdemonet: It needs to be configurable, but we need sensible defaults that come from "real data". That should be part of the requirements I believe (defining the sensible defaults based on various architectures).

> nbr of occurrences or frequency of a certain type of log, do you think about something else?

Not really, except if the log itself is outputting numeric values (e.g. an old component without an exporter could just log the duration of an operation after completion).

docs/developer/architecture/logs.rst
gdemonet: Oh yes right, you define what a role does, but it doesn't imply another can't do it per-se. It's a bit confusing, especially if you need to derive RBAC rules from those stories.

docs/developer/architecture/logs.rst
gdemonet: I think we should stick to a single definition of "system logs", as opposed to "container logs". Since we can manage them easily by name, let's only consider "systemd-managed services" under "system logs" for now. The exceptions (like `/var/log/metalk8s`) will need to be treated as exceptions anyway.

Yes, for browsing / dashboards that we would have to provide (we don't have any today, maybe should be in the requirements? also extending the existing dashboards with some log panels?)

docs/developer/architecture/logs.rst
gdemonet: That's an implementation detail, so not part of the requirements IMO. If we want full-text indexing on the logs, let's state it in the requirements and discard Loki today. If we don't want such full-text search, let's state it as well to make sure it is approved before we start the design.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Context
+-------
+
+MetalK8s value is to provide, out of the box, some services in order to ease
+the monitoring and operation of the platform as well as workload running on top
+of it.
+
+MetalK8s provides monitoring service with Prometheus and AlertManager in
+order to expose MetalK8s and workloads metrics and alerts.
```
Ebaneck: Could read better this way?

`Metalk8s provides a monitoring stack consisting of Prometheus and Alertmanager
with a goal to expose and monitor metrics and/or alerts for both
control-plane and workload-plane components.`

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
+Log Centralisation
+==================
+
+Context
+-------
+
+MetalK8s value is to provide, out of the box, some services in order to ease
+the monitoring and operation of the platform as well as workload running on top
```
Ebaneck: `s/workload/workloads`

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+of it.
+
+MetalK8s provides monitoring service with Prometheus and AlertManager in
+order to expose MetalK8s and workloads metrics and alerts.
+
+The logs generated by the platform and the workloads constitute an essential
+piece of information that can be very helpful when it comes to understand the
+root cause of a failure or a performance degradation. Because of the
```
Ebaneck: Let's be assertive here: If logs are essential then they are useful, so let's strike-off `can`.

- `s/understand/understanding`

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+distributed nature of MetalK8s and workloads running of top of it, it is
+important to have some tooling to ease the analysis of near real time and past
+logs from a central endpoint. As such those logs should be stored on the system
+for a configurable period of time. The logs analysis functionalities should be
```
Ebaneck: `s/those/these` => need to remove the passive voice
`s/period of time/period`

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+accessible through an API and a UI. It is also important to ease correlation
+with other type of indicators such as KPIs and alerts.
```
Ebaneck: `s/type/types`

The sentence is unclear or incomplete? + using KPIs as an example indicator feels incorrect or something...

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+The following requirements are not about audit logs but only focus on
+application logs.
```
Ebaneck: Could read:

`The following requirements focus on application logs with the ([pick one] exemption/exclusion) of audit logs.`

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+namespaces as well as kubelet, containerd, salt-minion, metalk8s or CLI logs in
+Grafana. One can correlate logs and metrics or alerts in one single Grafana
+Dashboard.
+Browsing logs can be achieved through a documented API.
+
+- Persistence, Retention
+
+Logs should be stored on a persistent storage. Some automatic purging mechanism
+should ensure we are never filling the persistent storage. The Platform
+Administrator can configure the retention period.
```
Ebaneck: Hmmm, if I set up a 50Gb PV with a retention period configured to last 10 days and for some reason, my PV is filled in 5 days because one of my applications was started in debug mode.

What then sets in here?
Do we automatically start purging things so as to not fill the volume or we respect the Admins retention period (meaning we might lose logs when Volume is filled)?

Seems like here, we have to prioritize one or this is a design issue and should not be raised here?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Log collection, storage and browsing services should remain available when
+losing one node or one disk.
+
+- log indexing
+
+For MetalK8s use case we want a lightweight tool to store and expose logs.
+Indexing log on their content provides a powerfull way to analyze logs but it
+will require too much resources both in terms of CPU and Storage. Instead, the
+log centralisation service should offer basic features to group logs.
```
Ebaneck: Could be awesome if the requirements layout examples of such groupings from the very start.
For example,  log groupings should be based on Container Name or POD Name or Container ID etc

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,120 @@
...
+
+- As a Platform Administrator, I want to browse all containers logs (from all
+  servers) from a unique endpoint, in order to ease distributed Platform and
+  workloads error investigation.
+
+- As a Platform Administrator, I want to browse non container (kubelet,
+  containerd, salt-minion, metalk8s CLI logs) logs (from all servers) from a
+  unique endpoint, in order to ease Platform and workloads error investigation.
+
+- As a Storage Administrator, I want to browse browse all Solution instance
```
Ebaneck: - Remove one `browse`

Ebaneck (dismissed):
Quick look!

docs/developer/architecture/logs.rst
Ebaneck: When we talk about accessibility through a UI, are we talking about the Grafana UI or MetalK8s UI ( can be achieved in the MetalK8s UI by consuming for example Loki's HTTP API as documented here: https://github.com/grafana/loki/blob/master/docs/api.md#lokis-http-api)

docs/developer/architecture/logs.rst
thomasdanan: updated

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: updated

docs/developer/architecture/logs.rst
thomasdanan: updated

docs/developer/architecture/logs.rst
thomasdanan: clarified

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,168 @@
...
+  the log.
+
+
+Deployment & Configuration
+--------------------------
+
+The log centralisation service is scheduled on infra nodes.
+A platform Administrator can operate the service as follow:
+
+- add persistent storage
```
xaf-scality, line 162: and remove

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,168 @@
...
+certain period of time.
+
+*Monitorable/Observable service (health, performances and alerts)*
+
+The Platform Administrator can monitor capacity usage, IOPS, latency and
+bandwidth of the Log centralisation service. He can also monitor the health of
+the service (i.e. if some active alerts exist). He can visualize active and
+past alerts impacting the service.
+
+*Performances (TBD)*
```
xaf-scality, line 115: If the load of incoming logs is too heavy we should drop some lines (and say it) to not impact the global performance.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,168 @@
...
+Dashboard.
+Browsing logs can be achieved through a documented API in order to expose logs
+in MetalK8s UIs or other workloads UIs.
+
+*Persistence, Retention*
+
+Logs should be stored on a persistent storage. Platform Administrator can
+configure a max retention period. Some automatic purging mechanism is triggered
+when logs are older than the retention period or when the persistent storage is
+about to reach its capacity limit.
```
xaf-scality: and log those operations

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,168 @@
...
+
+*Horizontally scalable (capacity and performance)*
+
+The Platform Administrator can add nodes or disks in order to increase storage
```
xaf-scality: and remove nodes or disks to decrease

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,168 @@
...
+*Monitorable/Observable service (health, performances and alerts)*
+
+The Platform Administrator can monitor capacity usage, IOPS, latency and
+bandwidth of the Log centralisation service. He can also monitor the health of
+the service (i.e. if some active alerts exist). He can visualize active and
+past alerts impacting the service.
+
+*Performances (TBD)*
+
+
```
xaf-scality, line 118: should we add that if any specific certificat (or files ...) are needed we will add them in MK8s backup ?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+Context
+-------
+
+MetalK8s value is to provide, out of the box, some services in order to ease
+the monitoring and operation of the platform as well as workloads running on
+top of it.
+
+It currently provides monitoring service, powered by Prometheus and
+AlertManager, in order to expose metrics and alerts for both control-plane and
+workload-plane components.
```
alexandre-allard: And the machines through node-exporter.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+The logs generated by the platform and the workloads constitute an essential
+piece of information when it comes to understanding the root cause of a failure
+or a performance degradation. Because of the distributed nature of MetalK8s and
+workloads running of top of it, the administrators need some tooling to ease
```
alexandre-allard:
```suggestion
workloads running on top of it, the administrators need some tooling to ease
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+the analysis of near real time and past logs, from a central endpoint. As such
+these logs should be stored on the platform, for a configurable period.
+Browsing the logs is accessible through an API and a UI. The UI should ease
+correlation between logs and health/performance KPIs as well as alerts.
+
+For organisations having their own Log centralisation system (like Splunk or
```
alexandre-allard:
```suggestion
For organisations having their own log centralisation system (like Splunk or
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+A lightweight tool to store and expose logs is required in order to minimize
+the HW footprints (CPU, RAM, Disks):
+
+- limited history: Storing the logs for very large period (3 years) is not
+  something metalK8s needs to provide as a feature. This can be achieved using
+  external log centralisation system.
+
+- stable ingestion: It is important to guaranty stable ingestion of the logs
+  and less important to guaranty stable performances when browsing/searching
```
alexandre-allard: I'm not sure, but I think the correct modern spelling is guarantee
```suggestion
- stable ingestion: It is important to guarantee stable ingestion of the logs
  and less important to guarantee stable performances when browsing/searching
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+  impact the application workloads and deploying log storage and search service
+  on infra nodes might help achieving this isolation.
+
+- stream indexing: It is not required to have automatic indexing of logs
+  content. Instead, the log centralisation service should offer basic features
+  to group/filter logs per tag/metadata defining the log stream.
+
+*Accessible from a central UI/API*
+
+Platform Admin or Storage Admin can visualize logs from all containers from all
```
alexandre-allard:
```suggestion
Platform Admin or Storage Admin can visualize logs from all containers in all
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+about to reach its capacity limit.
+A typical and default retention period is of 2 weeks.
+
+*Horizontally scalable (capacity and performance)*
+
+The Platform Administrator can add nodes or disks in order to increase storage
+capacity for logs (to keep more history or to store more workloads logs) or to
+increase performances for ingesting or querying the logs. It is also possible
+to fine tune the deployment configuration in order to have more replicas and
+ideally the Log centralisation service supports horizontal POD autoscaling.
```
alexandre-allard:
```suggestion
ideally the log centralisation service supports horizontal POD autoscaling.
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+*Log Querying*
+
+Get all logs for a given period, node(s), pod regex, limited list of predefined
+labels and free keywords text.
+Typical Zenko use case: collecting all logs across several components, related
+to a S3 uniq request.
+
+*Log statistics (nice to have)*
+
+The Log centralisation service also offer the ability to consume statistics
```
alexandre-allard:
```suggestion
The log centralisation service also offers the ability to consume statistics
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+about the logs like the number of occurrences of one type of log during a
+certain period of time.
+
+*Monitorable/Observable service (health, performances and alerts)*
+
+The Platform Administrator can monitor capacity usage, IOPS, latency and
+bandwidth of the Log centralisation service. He can also monitor the health of
```
alexandre-allard:
```suggestion
bandwidth of the log centralisation service. He can also monitor the health of
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+- As a Platform Administrator, I want to browse non container (kubelet,
+  containerd, salt-minion, cron) logs (from all servers) from a
+  unique endpoint, in order to ease System error investigation.
+
+- As a Storage Administrator, I want to browse all Solution instance containers
+  logs (from all servers) from a unique endpoint, in order to ease Solution
+  instance Error investigation.
+
+- As a Platform Administrator, I want to push all containers logs to an
+  external log centralisation system, In Order to archive it or aggregate it
```
alexandre-allard: For consistency across US, should be:
```suggestion
  external log centralisation system, in order to archive it or aggregate it
```
Or if you want to keep it this way, the two US above should be changed.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+  which produced those timeseries at the time of the alert.
+
+- After a pod or node disappears, I want to be able to retrieve logs from just
+  before it died, so I can diagnose why it died.
+
+- After discovering an ongoing issue with my service, I want to extract a
+  metric from some logs and combine it with my existing time series data.
+
+- I have a legacy job which does not expose metrics about errors - it only logs
+  them.  I want to build an alert based on the rate of occurrences of errors in
```
alexandre-allard: extra space
```suggestion
  them. I want to build an alert based on the rate of occurrences of errors in
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+*Performances (TBD)*
+
+Typical workloads can generate around 1000 logs per second per node.
+
+User Stories
+------------
+
+- As a Platform Administrator, I want to browse all MetalK8s containers logs
+  (from all servers) from a unique endpoint, in order to ease distributed
+  k8s and metalk8s services error investigation.
```
alexandre-allard:
```suggestion
  K8s and MetalK8s services error investigation.
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Logs can be seen in Grafana.
+Log centralisation monitoring information are displayed in the MetalK8s UI
+overview page.
+A Grafana dashboard gathering health/performance KPIs, as well as alerts for
+log centralisation service is available when deploying/upgrading MetalK8s.
+
+Documentation
+-------------
+The sizing section in Introduction page is updated to include log
```
alexandre-allard: Add a new line for consistency
```suggestion
-------------

The sizing section in Introduction page is updated to include log
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+centralisation service impact. The sizing rule takes in account the retention
+period, workloads expected log rate and workload pre defined indices. This rule
```
alexandre-allard:
```suggestion
period, workloads expected log rate and workload predefined indices. This rule
```

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+is to be known by solution developers to properly size the service based on
+the workload properties.
+
+The Post Installation page is updated to indicate that persistent storage is
+needed for log centralisation service.
+
+A new section should be added under Cluster and Configuration services explains
```
alexandre-allard:
```suggestion
A new section should be added under Cluster and Services Configurations to explain
```

alexandre-allard (commented):
Really cool reqs, only minor details (typo, cap, ...), LGTM.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+capacity for logs (to keep more history or to store more workloads logs) or to
+increase performances for ingesting or querying the logs. It is also possible
+to fine tune the deployment configuration in order to have more replicas and
+ideally the Log centralisation service supports horizontal POD autoscaling.
+
+*Highly Available*
+
+Log collection, storage and browsing services can be replicated in order to
+ensure the same level of High Availability Prometheus has i.e. can lose all
+replicas but one.
```
TeddyAndrieux: In this case, of HA with replicas everywhere, then you can **not** add a new node to extend the storage (as you mentioned in the previous section)

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+- I have a legacy job which does not expose metrics about errors - it only logs
+  them.  I want to build an alert based on the rate of occurrences of errors in
+  the log.
+
+
+Deployment & Configuration
+--------------------------
+
+The log centralisation service is scheduled on infra nodes.
```
TeddyAndrieux: > on infra nodes

Not only you also have, likely, promtail (or fluentd, fluent-bit) to install on every nodes to send logs

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+A platform Administrator can operate the service as follow:
+
+- add persistent storage
+- configure max retention period
+- adjust the number of replicas
+- push logs to an external log centralisation service
+- configure alerting related to log centralisation service
+
+Those operations are accessible through kubectl, kubectl metalk8s plugin or
+metalk8scli.
```
TeddyAndrieux: To me it's fully part of kubectl plugin we do not need anything in the `metalk8scli`

docs/developer/architecture/logs.rst
thomasdanan: @TeddyAndrieux right ... so how it works with Prometheus? extending the storage means extending PV size?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+piece of information when it comes to understanding the root cause of a failure
+or a performance degradation. Because of the distributed nature of MetalK8s and
+workloads running of top of it, the administrators need some tooling to ease
+the analysis of near real time and past logs, from a central endpoint. As such
+these logs should be stored on the platform, for a configurable period.
+Browsing the logs is accessible through an API and a UI. The UI should ease
+correlation between logs and health/performance KPIs as well as alerts.
+
+For organisations having their own Log centralisation system (like Splunk or
+Elasticsearch), MetalK8s should provide a way to push logs into it.
```
gdemonet: Careful with the wording: should MetalK8s (or its services) **push** logs, or expose them so external services can **pull** them? Could have a big impact on design.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
+Log Centralisation
```
gdemonet: Here, and everywhere in the doc: it's spelled `Centralization` with a `z`.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+- limited history: Storing the logs for very large period (3 years) is not
+  something metalK8s needs to provide as a feature. This can be achieved using
+  external log centralisation system.
+
+- stable ingestion: It is important to guaranty stable ingestion of the logs
+  and less important to guaranty stable performances when browsing/searching
+  the logs. However, peak loads related to complex logs queries should not
+  impact the application workloads and deploying log storage and search service
+  on infra nodes might help achieving this isolation.
```
gdemonet, line 47: The suggestion to deploy on infra nodes is nice, but not expressed as a requirement: is it one? Or is the ability to isolate the log service from actual workloads sufficient (maybe we just want to make sure those are Burstable, for instance)?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+- stream indexing: It is not required to have automatic indexing of logs
+  content. Instead, the log centralisation service should offer basic features
+  to group/filter logs per tag/metadata defining the log stream.
+
+*Accessible from a central UI/API*
+
+Platform Admin or Storage Admin can visualize logs from all containers from all
+namespaces as well as journal logs, including (kubelet, containerd, salt-minion
+, kernel, initrd, services, etc ...) in Grafana. One can correlate logs and
+metrics or alerts in one single Grafana Dashboard.
```
gdemonet, line 58: ?? A single Grafana dashboard for correlating "logs from all containers in all namespaces as well as journal logs" with "metrics"? Seems impossible, and highly undesirable.
I'd like us to specify what dashboards (among the ones we deploy today with Metal) would need to evolve to include logs (and maybe alerts? most of them aren't displayed in Grafana, IIRC).
And also to remind that such dashboards need to be maintained by the application producing the logs, and can't be made generic.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+Browsing logs can be achieved through a documented API in order to expose logs
+in MetalK8s UIs or other workloads UIs if needed.
+
+*Persistence, Retention*
+
+Logs should be stored on a persistent storage. Platform Administrator can
```
gdemonet, line 64: Didn't we mention we needed some _formula_ to compute the required storage space from a few factors (namely, ingestion rate, number of indices/cardinality, and retention time)? Being able to warn the admin about the expected space requirements when she's looking at changing the retention time would be a good addition.

docs/developer/architecture/logs.rst
gdemonet: Never full-caps for pods: either lowercase (`pod`) when referencing the concept, or capitalized (`Pod`) when referencing the API resources.
Note that if you want to refer to the HPA mechanism built in Kubernetes, then just refer to it directly (with maybe a link). When it comes to the service being able to "support" such scaling strategy, it's not the **auto**scaling part that matters most, but the **horizontal** scaling part (and this requirement is already expressed as "have more replicas").
In the end, unless you want HPA enabled by default, I don't think it brings much value to specify it here.

docs/developer/architecture/logs.rst
gdemonet: IMHO that's why we have two sections: one about scaling and one about HA. Implementation will differ for both use-cases, and we **should** be able to handle both.
For the Prometheus question, in the current state you can only increase the PV size. But in production, we can exploit more advanced patterns, like sharding, to achieve the same results (each "shard" can be managed by multiple replicas). There's also Federation, which allows to build slaves and aggregate each slaves' metrics... Many solutions (e.g. [this blog post](https://www.robustperception.io/scaling-and-federating-prometheus)), out of the scope of HA concerns however.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+*Highly Available*
+
+Log collection, storage and browsing services can be replicated in order to
+ensure the same level of High Availability Prometheus has i.e. can lose all
+replicas but one.
+
+*Log Querying*
+
+Get all logs for a given period, node(s), pod regex, limited list of predefined
+labels and free keywords text.
```
gdemonet, line 87: For predefined labels, don't you want to give more details? I'm fine with leaving this to the design, but maybe some of them must be implemented whichever design is chosen.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+Typical Zenko use case: collecting all logs across several components, related
+to a S3 uniq request.
```
gdemonet, line 89:
```suggestion
to a unique S3 request.
```

docs/developer/architecture/logs.rst
gdemonet: This covers resource consumption by the log centralization service. What about usage metrics from the service itself? E.g. ingestion rate (number of log lines per second), counter of indexed streams...

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+The Log centralisation service also offer the ability to consume statistics
+about the logs like the number of occurrences of one type of log during a
+certain period of time.
+
+*Monitorable/Observable service (health, performances and alerts)*
+
+The Platform Administrator can monitor capacity usage, IOPS, latency and
+bandwidth of the Log centralisation service. He can also monitor the health of
+the service (i.e. if some active alerts exist). He can visualize active and
```
gdemonet: Can we define the scope of those alerts beforehand? E.g. monitoring the remaining capacity may simply be handled at the Volume level.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+past alerts impacting the service.
+
+*Performances (TBD)*
+
+Typical workloads can generate around 1000 logs per second per node.
```
gdemonet, line 117: OK. But how many "workloads"? And what about the cardinality of those streams (the number of unique combinations of labels)?

docs/developer/architecture/logs.rst
gdemonet: No good reason to capitalize, let's change this one

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+  (from all servers) from a unique endpoint, in order to ease distributed
+  k8s and metalk8s services error investigation.
+
+- As a Platform Administrator, I want to browse non container (kubelet,
+  containerd, salt-minion, cron) logs (from all servers) from a
+  unique endpoint, in order to ease System error investigation.
+
+- As a Storage Administrator, I want to browse all Solution instance containers
+  logs (from all servers) from a unique endpoint, in order to ease Solution
+  instance Error investigation.
```
gdemonet:
```suggestion
  instance error investigation.
```

Would need to express the filtering/search capabilities in the User Story as well, IMHO (make it clear that, as a Storage Admin, I can easily drill down into the affected workloads that I want logs from).

docs/developer/architecture/logs.rst
gdemonet: That's not mandatory, and the requirements may impose that we have the ability to only run ingestors on a subset of the cluster nodes.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+- I have a legacy job which does not expose metrics about errors - it only logs
+  them.  I want to build an alert based on the rate of occurrences of errors in
+  the log.
+
+
+Deployment & Configuration
+--------------------------
+
+The log centralisation service is scheduled on infra nodes.
+A platform Administrator can operate the service as follow:
```
gdemonet:
```suggestion
A Platform Administrator can operate the service as follows:
```

docs/developer/architecture/logs.rst
gdemonet: The requirement should express: "operations are accessible from any host also able to access the control plane network, and should be wrapped in a common user-facing CLI". Which actual tool is chosen for implementation is out of scope for the requirements.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+
+- add persistent storage
+- configure max retention period
+- adjust the number of replicas
```
gdemonet, line 164: So, autoscaling or not?

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+- push logs to an external log centralisation service
```
gdemonet: Manual operation? Pushing a dump, or configuring a Job to push incremental snapshots? Would need much more details around what's to be done with external services (do we need filtering/downsampling/deduplication? should we obfuscate externalized logs?)

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,197 @@
...
+- configure alerting related to log centralisation service
```
gdemonet: What configuration? Thresholds/silences/... (the usual)? Define new alerts? For monitoring what, the log service or the app(s) from which we are collecting logs?

docs/developer/architecture/logs.rst
gdemonet: Number of indices cannot be "pre-defined": it depends on the deployment + usage. Imagine a label extracting the 'bucket name' from a cloudserver log line: if you had a million log lines with a single bucket, you'll get a single index, whereas a thousand log lines with each a different buckets will generate a thousand indices.
It's the whole cardinality topic, which is also really important to manage for Prometheus.

docs/developer/architecture/logs.rst
gdemonet: IMHO, we should remove the CSC section of the Operations guide, it doesn't mean anything for a user. Instead, keep focused pages, each dedicated to a single service.

docs/developer/architecture/logs.rst
gdemonet: (seen that it's requested in Documentation, could be sufficient)

gdemonet (commented):
Some comments!

docs/developer/architecture/logs.rst
TeddyAndrieux: Yes, sure both doable but not at the same time and I think if we keep these 2 requirements we need to say it somewhere in this document.
Anyway as first iteration we will likely do the HA one as it's the simplest IMO

For prometheus (and it will be the same for Log centralization) a lot of possibility but pros and cons for each ....
"on pourra pas sauver tout le monde" :smiley:

docs/developer/architecture/logs.rst
thomasdanan: will specify log storage service to run on infra nodes

docs/developer/architecture/logs.rst
thomasdanan: I was more thinking about our system is able to push logs to different locations: either loki or an external splunk (even if it means deploying a specific forwarder on cluster nodes).

docs/developer/architecture/logs.rst
thomasdanan: actually I used centralisation everywhere but changing with centralization

docs/developer/architecture/logs.rst
thomasdanan: Correlation can be only visual or based on a specific time frame, it does not necessarily mean that the dashboard automatically guess the relation between one specific alert, log line and perf metric but being able to see specific KPIs, alerts and log streams at the same time is already with a big value. Also, it could be interesting to visualize, in the same Volume/PV dashboard, IOPS/latency KPIs and Alerts next to some RAID card logs so that a RAID card log error could be correlated with a drop of disk IO performances (or peak of latency) and this can help to understand the root cause very quickly. I don't want to change the 3rd party dashboard we have today but with the time we could build our own dashboards based on real-life experience and leveraging those 3 types of information (kpi, alerts, logs and later on tracing ...)

docs/developer/architecture/logs.rst
thomasdanan: yes in the sizing part

docs/developer/architecture/logs.rst
thomasdanan: ok removing this vert last part about autoscaling

docs/developer/architecture/logs.rst
thomasdanan: I am thinking about severity and namespace

docs/developer/architecture/logs.rst
thomasdanan: I gave 2 examples of things that could impact the log service but at the end what the platform admin wants is to be notified when the service is degraded. Root cause is part of the alert notification.

docs/developer/architecture/logs.rst
thomasdanan: It is just an indication of what a one Zenko Solution instance could generate in terms of logs. So one namespace and for the combination of labels I don't know yet

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: removed

docs/developer/architecture/logs.rst
thomasdanan: reword in configuring the service so that it can push logs to external system ...

docs/developer/architecture/logs.rst
thomasdanan: add details

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: so now we are talking about journal logs.
About Grafana dashboards, what is important is the ability to display logs in Grafana. With time and experience, we will then build predefined dashboards (but can't specify it now)

docs/developer/architecture/logs.rst
thomasdanan: updated

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
thomasdanan: Okay actually the requirement is about being able to scale the service in order ingest and store more logs

docs/developer/architecture/logs.rst
thomasdanan: it is expressed as a requirement in the Deployment & configuration section.
I think it makes sense to schedule all APM (monitoring, alerting, logging, ...) services on infra nodes in order to isolate the services from workloads and k8s master services.

docs/developer/architecture/logs.rst
thomasdanan: Yes for ingestion rate, for counter indexed streams I am not sure ... not adding it right now. I guess what would be interesting is to detect workloads that generate crazy amount of logs

docs/developer/architecture/logs.rst
nloewensen: I think that audit-logs are important to be collected (it is a frequent customer request) and since we are providing a multi-tenancy solution, each tenant should be able to query the audit-logs related to his/her tenant, i.e. which user in my tenant did what, who/when were changes done to my tenant (e.g. changing quotas, resetting passwords etc.). I cannot comment on which system to use or if that is more application, i.e. Zenko related and hence should be handled somewhere else but I think in general audit-logs/trails is something we need to provide for our "overall" solution.

docs/developer/architecture/logs.rst
nloewensen: RBAC and multi-tenancy are two completely different features to me. Probably RBAC is better addressed on the "viewing/display" instance rather than on the collection.
Same goes for Multi-tenancy, which is probably handled better somewhere else (higher in the solution, i.e. UI, CLI) rather than on the collection. Still (see above), I think we need to be able to collect audit-logs and for all logsg include the appropriate "hooks", so that they can be filtered appropriately to be only displayed to a specific tenant.

docs/developer/architecture/logs.rst
NicolasT: This is why audit logs are not logs as (IMHO) we're discussing here.

'Log aggregation' is about logs an operator uses to debug/monitor/... some software.

Audit logs are consumed by a *user* of said software (or some third-party solution this *user* has in place to analyze said logs).

A *user* doesn't care about, say, `Unable to connect to Zookeeper, retrying in 5s`, this is for operators. Similarly, an *operator* shouldn't be inspecting things like `User XYZ has created bucket ABC`. Heck, `XYZ` and `ABC` may be PII this operator is not even supposed/allowed to see! However, the *user* (of this account, for example) should be able to inspect this.

Please don't merge the two concepts.

docs/developer/architecture/logs.rst
TeddyAndrieux: > That's not mandatory, and the requirements may impose that we have the ability to only run ingestors on a subset of the cluster nodes.

Really ? Then you may have log centralization with only a part of all the logs ? For example you will have salt-minion, kubelet, ... logs of infra nodes and that it...

To me, we really want to have it everywhere, it's weird to have log centralization for some nodes but not for others ... it's not really centralized in this case.
To me, if you add "ability to run ingestors on a subset of the cluster nodes" then you want to do the same for node exporter (and to me both does not make sense)

gdemonet: No you're right, I didn't mean to ingest only logs for a subset of Nodes. I don't know why, but I had the wrong idea ingestors could consume logs from other Nodes, but that's not how it works in `promtail` or `fluent-bit` (even for Pod logs, they read from Node-local logfiles).
What would be nice is to make sure in the design that storage + as much compute resources as possible are consumed on the infra nodes, and leave the bare minimum in the ingestors (ideally just forward the text streams?).

docs/developer/architecture/logs.rst
xaf-scality: From my point of view pull is more "easy" as we can throttle the incoming request. push is more complex because we need to handle retries by example.

docs/developer/architecture/logs.rst
xaf-scality: The goal is to have the ability to isolate the log service if needed

docs/developer/architecture/logs.rst
xaf-scality: one single dashboard is ambitious. Do you mean that correlate logs is more in a way of counting error/warn logs in this single dashboard ?

xaf-scality (dismissed):

docs/developer/architecture/logs.rst
gdemonet: > we could build our own dashboards based on real-life experience and leveraging those 3 types of information (kpi, alerts, logs and later on tracing ...)

OK that's what I thought. Maybe reformulate to make this obvious we are not building "one singe dashboard", but some dashboards with one clear purpose (e.g. disk issues investigations), combining views for logs ([Logs panel](https://grafana.com/docs/grafana/latest/panels/visualizations/logs-panel/)), metrics and alerts (using all the usual panels). Traces are really for deep-dive however, so I think keeping them as they are exposed in Grafana 7 (split-screen between the dashboard and the tracing view) is sufficient, no need to get them always visible in a dashboard.

docs/developer/architecture/logs.rst
thomasdanan: indeed, will add one section about upgrade & backup to make sure that service customization is backed up and persist following an upgrade

docs/developer/architecture/logs.rst
thomasdanan: done

docs/developer/architecture/logs.rst
TeddyAndrieux: My point is that you cannot match the 2 requirements in a deployment, either it's horizontally scalable either it's HA (with lose all but one).
And btw horizontally scalable need more logic on our side.

Anyway we can have the ability to handle both different deployment (likely with some CSC tunning) but a single deployment cannot match both. 

To me, it should be mentioned somewhere in this document and if we really want to support both different deployment then you should add a point about it in the `Deployment & Configuration` section

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,215 @@
...
+  (from all servers) from a unique endpoint, in order to ease distributed
+  K8s and MetalK8s services error investigation.
+
+- As a Platform Administrator, I want to browse non container (kubelet,
+  containerd, salt-minion, cron) logs (from all servers) from a
+  unique endpoint, in order to ease System error investigation.
+
+- As a Storage Administrator, I want to browse all Solution instance containers
+  logs (from all servers) from a unique endpoint, in order to ease Solution
+  instance error investigation.
```
TeddyAndrieux, line 132: NOTE that if you are "horizontally scalable" you will likely not be able to have all servers logs from only one endpoint as we do not have one server storing all servers data it's spread in this case.
So logs will come from different data sources
(At least it's the behavior for prometheus by default and it will likely be the same with Loki&co)

alexandre-allard (approved):

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,215 @@
...
+correlation between logs and health/performance KPIs as well as alerts.
+
+For organisations having their own Log centralization system (like Splunk or
+Elasticsearch), MetalK8s should provide some documentation in order guide the
+customer to deploy and configure its own log collection agent.
+
+The following requirements focus on application logs. Audit logs are not part
+of the requirements.
+
+Requirements
```
benoit-a, line 31: I would add a requirement on getting the log "visible". IE : the difference between the time when the logs is generated and "printed" (visible in kubectl logs) and the time when it appears in the UI.
Also, I would precise if loss of logs is acceptable and in which circumstances it is and how (if ?) the end user is notified of the loss.

docs/developer/architecture/logs.rst
```diff
@@ -0,0 +1,215 @@
...
+order to ensure that we can lose at least one server in the cluster without
+impacting availability and reliability of the service.
+
+*Log Querying*
+
+Get all logs for a given period, node(s), pod regex, limited list of predefined
+labels and free keywords text.
+Typical Zenko use case: collecting all logs across several components, related
+to a S3 uniq request.
+Typical predefined labels are severity and namespace.
```
benoit-a, line 90: I am probably late to the party ; but this could lead to requirements on the server producing the logs themselves (like "structured logs" or using the same labels/names across the stack). Not exactly a requirement on the system itself, more of a pre-requisite, IMO

PR #2557, opened 2020-05-18
Bump prometheus-operator version to 8.13.0

Ebaneck:
**Component**:

'charts', 'prometheus-operator'

**Context**: 

See #2430 

**Summary**:

Now, that the python-kubernetes version has been updated, we can proceed
with the chart upgrade.

The following issue https://github.com/scality/metalk8s/issues/2430#issuecomment-617692984 raised previously while upgrading the charts should have been fixed after the python-kubernetes version was bumped.

**Acceptance criteria**: 

Green build

---

Closes: #2430


gdemonet:
Would need to regenerate the deployed rules with what you defined in #2538. ~We need a test to enforce this (otherwise the tooling has close to no purpose)~ _Edit: didn't see #2546_.

alexandre-allard (approved):

Issue #2422, opened 2020-04-17
Solutions config.yaml & product.txt should be merged in a single manifest.yaml file


alexandre-allard:
**Component**: solutions

**Why this is needed**:
Actually we have 2 files describing the Solution instead of one, moreover the `config.yaml` file is badly name as it suggests this is a configuration file rather than a manifest of the Solution.

**What should be done**:
* Merge `config.yaml` and `product.txt` in a single file and name it with a more relevant name such as `manifest.yaml`.
* Adapt Salt to use the new file
* Adapt the Solutions post-merge tests in the CI

**Implementation proposal** (strongly recommended):

**Test plan**:
There is already Solutions tests with pytest-bdd and cypress, they must not break.

gdemonet:
What about MetalK8s ISO and `product.txt`? Could we unify the tooling for this?

NicolasT:
I'd rather stick with `product.txt`, for one very simple reason: it can be read, parsed,... without any tooling, libraries,... For bootstrapping purposes, this is quite important.

gdemonet:
OK, fair point. But we don't have the same need for Solution ISOs, right?

NicolasT:
> OK, fair point. But we don't have the same need for Solution ISOs, right?

Likely, indeed. My remark mostly concerns MetalK8s ISOs that need to be bootstrap'able.

miniscruff:
I am wondering if this is going to be done in the next month or so as I am building the zenko solution. Or alternatively, if the current implementation will be maintained for a minor release or two for compatibility.

gdemonet:
@miniscruff we'll make sure to have it available in the next iteration of Solutions, which we really want to work on for next month. In any case, you'll need #2389, which will change the format of this file (at the very least, remove the `customApiGroups` section), so if you can wait a bit, we'll update this ticket / our docs to reflect the next format (single `manifest.yaml` file).
For your question on compatibility, we could maintain it for a minor version (2.5.x), and drop it in the following, if that's needed. But I'd rather have the "next" approach for Zenko Solution ISO, since it's starting just now :)

miniscruff:
I think the zenko solution is meant to be code complete this month, not sure if we can wait on that or if we just need to do an update when this is ready. @rahulreddy thoughts?

gdemonet:
[...]

PR #2565, opened 2020-05-20
Solutions manifest

alexandre-allard:

**Context**: See #2422

**Summary**:
Replace `config.yaml` + `product.txt` with a single `manifest.yaml`.
Also the format has changed a bit to look more like a valid CR.

**Acceptance criteria**:
Solution workflow is still working (import, activate, ...) and the tests are green.

---

Closes: #2422


salt/_modules/metalk8s_solutions.py
```diff
@@ -250,6 +251,58 @@ def read_solution_config(mountpoint, name, version):
...
+    log.debug('Result: %r', result)
+
+    if result['retcode'] != 0:
+        raise CommandExecutionError(
+            'Failed to run isoinfo: {}'.format(
+                result.get('stderr', result['stdout'])
+            )
+        )
+
+    manifest = yaml.safe_load(result['stdout'])
```
gdemonet: We should catch YAML errors here and re-raise as `CommandExecutionError`

salt/_modules/metalk8s_solutions.py
```diff
@@ -202,52 +201,110 @@ def list_solution_images(mountpoint):
...
+
+    return manifest, info
+
+
+def _archive_info_from_manifest(manifest):
+    info = {}
+
+    try:
+        info['name'] = manifest.get('metadata', {})['name']
+        info['version'] = manifest.get('spec', {})['version']
```
TeddyAndrieux: Minor: Since you check for key error to me you do not want to use `get` here

TeddyAndrieux (commented):
LGTM (just a minor comment but feel free to ignore it)

TeddyAndrieux (dismissed):
(ups wanted to approve)

salt/_modules/metalk8s_solutions.py
alexandre-allard: Yes, right, will keep it as is if it's the only requested change.

salt/_modules/metalk8s_solutions.py
```diff
@@ -202,52 +201,110 @@ def list_solution_images(mountpoint):
...
+    except KeyError:
+        info['display_name'] = info['name']
+
+    info['id'] = "{}-{}".format(info['name'], info['version'])
+
+    return info
+
+
+def archive_info_from_iso(path):
+    """Extract archive information from a Solution ISO
```
gdemonet: We could call this method `manifest_from_iso` now that we have a good name for it :)

salt/_modules/metalk8s_solutions.py
```diff
@@ -202,52 +201,110 @@ def list_solution_images(mountpoint):
...
+
+    manifest = salt.utils.dictupdate.merge(
+        default_manifest, manifest, strategy='recurse'
+    )
+
+    return manifest, info
+
+
+def _archive_info_from_manifest(manifest):
+    info = {}
```
gdemonet: As a general comment, this way of writing methods that build and return an object, although widespread in Salt modules, makes the code quite hard to read. You need to parse through each branch to understand what the contents of an object will be when it's returned. Sprinkle some early returns in there and it's quickly becoming a mess.

Instead, I'd suggest we only build the structure in the final `return` statement (even in an early `return`?). Something like this:

```python
def _archive_info_from_manifest(manifest):
    name = manifest.get('metadata', {}).get('name')
    version = manifest.get('spec', {}).get('version')

    if any(key is None for key in [name, version]):
        raise CommandExecutionError(
            'Missing mandatory key(s) in Solution "{}": must provide '
            '"metadata.name" and "spec.version"'
            .format(SOLUTION_MANIFEST)
        )

    display_name = manifest.get('annotations', {}).get(
        'solutions.metalk8s.scality.com/display-name', name
    )

    return {
        'name': name,
        'version': version,
        'display_name': display_name,
        'id': '-'.join([name, version]),
    }
```

salt/_modules/metalk8s_solutions.py
```diff
@@ -202,52 +201,110 @@ def list_solution_images(mountpoint):
...
+
+    try:
+        info['name'] = manifest.get('metadata', {})['name']
```
gdemonet: Although this comes from `metadata.name` in the manifest, this isn't an actual CRD (yet), so we need to perform validation manually and ensure this is lowercased and URL-encoded (because we'll use it, through the computed `info['id']`, as a path in our custom registry).

examples/metalk8s-solution-example/manifest.py
```diff
@@ -0,0 +1,114 @@
...
+
+SOLUTION_MANIFEST = {
+    'apiVersion': 'solutions.metalk8s.scality.com/v1alpha1',
+    'kind': 'Solution',
+    'metadata': {
+        'name': '',
+        'annotations': {},
+        'labels': {},
+    },
+    'spec': {
```
gdemonet: The list of extra images could be added as an option to the script, so we can check that overriding the default behaviour works as expected (we could build an ISO with a wrong manifest and check how our tooling behaves). But not needed for now, so feel free to leave this as debt (with a ticket to track and test) :)

gdemonet (changes requested):
Outside of the validation of the Solution `name` field, only a few minor comments.

Two more troubling questions:
- in 2.4? not sure that's a good strategy, it's a breaking change (would put it in 2.5.1 at the earliest)
- what about docs? I'd like to see at least an updated Solutions guide with an explanation and example of this new format, maybe even an extra note in the `CHANGELOG.md` warning about this change

alexandre-allard:
> Outside of the validation of the Solution `name` field, only a few minor comments.
> 
> Two more troubling questions:
> 
> * in 2.4? not sure that's a good strategy, it's a breaking change (would put it in 2.5.1 at the earliest)
> * what about docs? I'd like to see at least an updated Solutions guide with an explanation and example of this new format, maybe even an extra note in the `CHANGELOG.md` warning about this change

@gdemonet I wanted to put it in 2.4 to avoid Solution developers to maintain 2 format to support 2.4 and 2.5+, but maybe that's not what we want and they should just not use 2.4.

For the doc part... I would say that I use the "Code is the new Doc" or "Doc As Code" approach.
Joke aside, I agree, I will add some and also a note in the CHANGELOG.

salt/_modules/metalk8s_solutions.py
alexandre-allard: As you prefer, I did this because as you said, most of the time it's the approach used in Salt modules, so I tried to be consistent.

gdemonet: Consistently bad is still bad :P I'd rather have us change this "standard" progressively by checking for this pattern in reviews than do a full refactor on thousands of lines in a few months...

examples/metalk8s-solution-example/manifest.py
alexandre-allard: Added the list, so we just have to add the test now.
I'll create a ticket.

miniscruff:
I was wondering if I should/need to adjust my PR here https://github.com/scality/Zenko/pull/1018 or if the versions expected / required make it not an issue. ( It is not a big change regardless but not sure of order or release timings ) @rahulreddy FYI.

alexandre-allard:
> I was wondering if I should/need to adjust my PR here [scality/Zenko#1018](https://github.com/scality/Zenko/pull/1018) or if the versions expected / required make it not an issue. ( It is not a big change regardless but not sure of order or release timings ) @rahulreddy FYI.

@miniscruff Yes, it would be better if you could adjust your PR accordingly, otherwise you'll have to manage different versions/formats.
There is also another change coming for the Operator roles (#2576).

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 
+### Breaking changes
+
+- Solutions product information format has changed, there is a new
+  `manifest.yaml` file to describe the whole Solution instead of the
+  `product.txt` and `config.yaml`
+  ([#2422](https://github.com/scality/metalk8s/issues/2422)).
+  See [Solutions documentation](https://metal-k8s.readthedocs.io/en/development-2.5/developer/solutions/archive.html#product-information)
+  for details about the new format.
+
```
gdemonet: Can you add a statement explaining the impact? Like `Solution archives working on previous versions of MetalK8s will no longer be compatible and will need to be regenerated` or something...

docs/developer/architecture/solutions.rst
```diff
@@ -116,35 +116,32 @@ Here is how it will look::
...
+        - solution-name-operator:1.0.0
+        - solution-name-ui:1.0.0
+      operator:
+        image:
+          name: solution-name-operator
+          tag: 1.0.0
+      ui:
+        image:
+          name: solution-name-ui
+          tag: 1.0.0
```
gdemonet, line 141: So we're still deploying UIs with MetalK8s... Let's open a debt ticket to remove this behaviour. Also, we may want to add some docs explaining that if this section is not specified, we won't try to deploy the UI ourselves (recommended approach, let the Operator do it).

docs/developer/solutions/archive.rst
```diff
@@ -67,56 +67,56 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
...
+       - solution-name-operator:1.0.0
+       - solution-name-ui:1.0.0
+     operator:
+       image:
+         name: solution-name-operator
+         tag: 1.0.0
+     ui:
+       image:
+         name: solution-name-ui
+         tag: 1.0.0
```
gdemonet, line 109: Maybe remind that this is optional and conditions the deployment of said UI by MetalK8s.

docs/developer/solutions/archive.rst
```diff
@@ -67,56 +67,56 @@ Here is the file tree expected by MetalK8s to exist in each Solution archive::
...
+     version: 1.0.0
+
+It is recommended for inspection purposes to include some annotations related
+to the build-time conditions, such as the following (where command invocations
+should be statically replaced in the generated ``manifest.yaml``)::
+
+   solutions.metalk8s.scality.com/build-timestamp: \
+     $(date -u +%Y-%m-%dT%H:%M:%SZ)
+   solutions.metalk8s.scality.com/git-revision: \
+     $(git describe --always --long --tags --dirty)
```
gdemonet, line 125: Maybe add a note linking to your simple script for generating the manifest? Just in case readers want to copy-paste it.

docs/developer/solutions/introduction.rst
```diff
@@ -28,49 +27,48 @@ For more details, see the following documentation pages:
 - :doc:`./operator`
 - (TODO) Solution UI guidelines
 
-Once a Solution is deployed on MetalK8s, a user can deploy one or more versions
-of the Solution Operator, using either the Solution UI or the Kubernetes API,
-into separate namespaces. Using the Operator-defined ``CustomResource(s)``, the
-user can then effectively deploy the application packaged in the Solution.
+Once a Solution is imported in MetalK8s, a user can deploy one or more versions
+of the Solution Operator, using either the Solution UI or CLI, into separate
```
gdemonet: `Solution UI or CLI`: what's that? I think you mean the `solutions.sh` script or the "Environments" page in MetalK8s UI?
Interacting with the CRs would then use a `Solution UI` or `CLI` (probably kubectl plugins).

docs/developer/solutions/operator.rst
```diff
@@ -69,28 +56,38 @@ to set up monitoring of your application services.
...
+     <solution-version-x>:
+       - endpoint: metalk8s-registry/<solution-name>-<solution-version-x>
+         images:
+           - <image-x>:<tag-x>
+           - <image-y>:<tag-y>
+     <solution-version-y>:
+       - endpoint: metalk8s-registry/<solution-name>-<solution-version-y>
+         images:
+           - <image-x>:<tag-x>
+           - <image-y>:<tag-y>
```
gdemonet, line 79: Cool, but not really explaining why we use this format or what it enables them to do with it.
To make this a bit more obvious, you could show this format as "what MetalK8s will inject when deploying your Operator", and show another example of what the Operator could use as defaults when running with online access, e.g.:

```yaml
apiVersion: solutions.metalk8s.scality.com/v1alpha1
kind: OperatorConfig
repositories:
  1.0.0:
    - endpoint: registry.scality.com/zenko
      images:
        - cloudserver:1.0.0
        - zenko-quorum:1.0.0
    - endpoint: quay.io/coreos
      images:
        - prometheus-operator:v0.34.0
```

docs/developer/solutions/operator.rst
```diff
@@ -69,28 +56,38 @@ to set up monitoring of your application services.
...
 
 .. note::
 
-   Future improvements to MetalK8s may include the addition of an "Operator for
-   Operators", such as the `Operator Lifecycle Manager`_.
-
-.. _`Operator Lifecycle Manager`:
-   https://github.com/operator-framework/operator-lifecycle-manager
+   ``Role`` and ``RoleBinding`` should be preferred over ``ClusterRole`` and
+   ``ClusterRoleBinding``, if there is no need for cluster-scoped permissions.
```
gdemonet: Would be nice to track this topic in a ticket: most importantly, we don't want to allow cluster-scoped permissions for Environment-scoped Operators. But maybe we'll need other scopes, and the question of multiple namespaces within an environment is also non-trivial, so let's tackle it later.

examples/metalk8s-solution-example/manifest.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    return parser.parse_args()
+
+
+def main():
+    """Script entrypoint."""
+    args = parse_arguments()
+
+    manifest = SOLUTION_MANIFEST
```
gdemonet: FYI: if it's something you're mutating in your method, then it can't be a constant. Hence, can't be using the constants naming conventions, to avoid confusing a user.

If the intention is to generate a new object from a set of args, then just define this object in the method body. This way, there's no risk of side effects:

```python
def _img_dict(name, tag):
    return {'image': {'name': name, 'tag': tag}}

def build_manifest(args):
    return {
        'apiVersion': 'solutions.metalk8s.scality.com/v1alpha1',
        'kind': 'Solution',
        'metadata': {
            'name': args.name,
            'annotations': dict(args.annotations),
            'labels': dict(args.labels),
        },
        'spec': {
            'images': list(set(
                ':'.join(image) for image in (
                    args.extra_images + [args.ui_image, args.operator_image]
                )
            )),
            'ui': _img_dict(*args.ui_image),
            'operator': _img_dict(*args.operator_image),
            'version': args.version,
        },
    }

# Try to make your functions testable
# Here, we can pass in a stream to output the manifest string,
# so in a test we can just pass an in-memory buffer and inspect it
def main(stream=sys.stdout):
    args = parse_arguments()
    manifest = build_manifest(args)

    # Also, no need to manually write,
    # just pass the stream to `yaml.safe_dump`
    yaml.safe_dump(manifest, stream, default_flow_style=False)
```

gdemonet: It's just a helper script, so no need to change anything. But it's not longer to write code "the right way", so I hope you'll think about it in the future ;)

gdemonet (changes requested):
We have some time before releasing 2.5.1 I think, so let's improve a bit more :)

examples/metalk8s-solution-example/manifest.py
alexandre-allard: Yes, I know this, but since it's just a simple script to generate a YAML, I thought it would be sufficient.
Oh cool for the stream stuff, didn't know about that one.

docs/developer/architecture/solutions.rst
alexandre-allard: Ok, opened the ticket #2581

docs/developer/solutions/archive.rst
alexandre-allard: Yep

docs/developer/solutions/introduction.rst
alexandre-allard: Yes, that's what I meant, I'll update this to make it more clear.

examples/metalk8s-solution-example/manifest.py
gdemonet: ping

docs/developer/solutions/operator.rst
gdemonet: ping

salt/metalk8s/orchestrate/solutions/prepare-environment.sls
```diff
@@ -155,8 +154,8 @@ Cannot deploy Solution {{ name }}-{{ version }} for environment {{ env_name }}:
                                  | selectattr('version', 'equalto', version)
                                  | first %}
 
-              {{- deploy_operator(namespace, solution) }}
-              {{- deploy_admin_ui(namespace, solution) }}
+              {{- deploy_operator(namespace, name, solution) }}
+              {{- deploy_admin_ui(namespace, name, solution) }}
```
gdemonet, line 158: Maybe calling this macro should be made optional based on the presence of a `spec.ui.image` key in the manifest? Or we can just ignore until #2581 is done

gdemonet (approved):
LGTM!

examples/metalk8s-solution-example/manifest.py
alexandre-allard: #2618

docs/developer/solutions/operator.rst
alexandre-allard: #2619

PR #2568, opened 2020-05-25
vagrant: add workaround for VirtualBox audio issue

alexandre-allard:
This issue appears with VirtualBox version 5.2.8+ which seems to have an issue when trying to add
the audio to a VM.
We can either shorten the VM name to workaround this issue or disable the audio.
Since we don't need the audio on our boxes, we choose to deactivate the audio.

Without thix fix we end up with such errors:
```
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["startvm", "20b29baa-fc92-45e5-9ebd-e43d5f87228d", "--type", "headless"]

Stderr: VBoxManage: error: The specified string / bytes buffer was to small.
Specify a larger one and retry. (VERR_CFGM_NOT_ENOUGH_SPACE)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole
```

See hashicorp/vagrant#9524 for details.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -161,7 +161,7 @@ Vagrant.configure("2") do |config|
     v.linked_clone = true
     v.memory = 2048
     v.cpus = 2
-    v.customize ["modifyvm", :id, "--chipset", "ich9"]
+    v.customize ["modifyvm", :id, "--chipset", "ich9", "--audio", "none"]
   end
 
   config.vm.network "private_network",
@@ -178,6 +178,7 @@ Vagrant.configure("2") do |config|
 
     bootstrap.vm.provider "virtualbox" do |v|
       v.memory = 4096
+      v.customize ["modifyvm", :id, "--audio", "none"]
       bootstrap.vm.synced_folder ".", "/vagrant", type: "virtualbox"
     end
 


Vagrantfile
```diff
@@ -161,7 +161,7 @@ Vagrant.configure("2") do |config|
     v.linked_clone = true
     v.memory = 2048
     v.cpus = 2
-    v.customize ["modifyvm", :id, "--chipset", "ich9"]
+    v.customize ["modifyvm", :id, "--chipset", "ich9", "--audio", "none"]
```
slaperche-scality, line 164: What's the relation between `audio` and the length of the name?

alexandre-allard: I don't really know, but it works.
I tried to find more information on the Internet on why it's related, but didn't manage to find much.
It's not directly related to the VM name as far as I understand, check this comment: https://github.com/hashicorp/vagrant/issues/9524#issuecomment-389269817 (the only relevant thing I found).
It seems that there is an issue when virtualbox is adding the audio to the VM, because of a too long path compared to the buffer used to store it. We could also shorten the VM name to workaround this issue... but since I assume that we don't care about audio on our vagrant box...

alexandre-allard: I'll update the commit message with more precise information.

gdemonet (approved):

slaperche-scality (approved):

PR #2571, opened 2020-05-25
Use async call for disk.dump

### LINKED ISSUE #2493: Handle error on disk.dump Salt call
**Component**:

operator

**What happened**:

Sometimes, Volume creation fails with:

```
Message:               call to GetVolumeSize failed: disk.dump failed (target=worker-3, path=/dev/disk/by-uuid/567c960e-dcba-4cef-b5b2-4d8bb0a3446e): Salt API failed with code 500: {"status": 500, "return": "An unexpected error occurred"}
```

If we enable debug log on Salt API side, we get:

```
[DEBUG   ] Trying to connect to: tcp://192.168.1.40:4506
[DEBUG   ] Error while processing request for: /
  File "/usr/lib/python2.7/site-packages/salt/netapi/rest_cherrypy/app.py", line 858, in hypermedia_handler
    ret = cherrypy.serving.request._hypermedia_inner_handler(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/cherrypy/_cpdispatch.py", line 60, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/usr/lib/python2.7/site-packages/salt/netapi/rest_cherrypy/app.py", line 1300, in POST
    token=cherrypy.session.get('token')))
  File "/usr/lib/python2.7/site-packages/salt/netapi/rest_cherrypy/app.py", line 1192, in exec_lowstate
    ret = self.api.run(chunk)
  File "/usr/lib/python2.7/site-packages/salt/netapi/__init__.py", line 75, in run
    return l_fun(*f_call.get('args', ()), **f_call.get('kwargs', {}))
  File "/usr/lib/python2.7/site-packages/salt/netapi/__init__.py", line 103, in local
    return local.cmd(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/client/__init__.py", line 754, in cmd
    **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/client/__init__.py", line 358, in run_job
    raise SaltClientError(general_exception)
SaltClientError: Salt request timed out. The master is not responding. You may need to run your command with `--async` in order to bypass the congested event bus. With `--async`, the CLI tool will print the job id (jid) and exit immediately without listening for responses. You can then use `salt-run jobs.lookup_jid` to look up the results of the job in the job cache later.
[INFO    ] 10.233.132.74 - - [27/Apr/2020:15:56:05] "POST /?timeout=1 HTTP/1.1" 500 57 "" "Go-http-client/1.1"
```

The problem may be related to a busy Salt master and/or more than one state trying to run at once.

**What was expected**:

The Operator should retry this operation and eventually the Volume should be created.

**Steps to reproduce**

@gdemonet managed to reproduce it by creating 10 sparseloop volumes at once on the same node

**Resolution proposal**:

Use async call for `disk.dump` (it's currently sync), that way we will benefit from the auto-retry logic that we've already implemented

slaperche-scality:
**Component**:

operator

**Context**: 

Sometimes the call to `disk.dump` fails because we have another state running concurrent and this permanently put the volume in a failed state.

**Summary**:

Given the lack of details in Salt-API response (the server return an error 500 with a generic message, not much you can infer from that…) the simplest solution is to make the call to `disk.dump` an async one (that way we leverage the existing handling of concurrent states).
As a side bonus, we also no longer need a timeout (which wasn't handled properly in our case…).

The biggest change is the enrichment of the job ID (which becomes a Job Handle).

**Acceptance criteria**: 

No regression (existing tests still pass)
Cannot reproduce the described issue.

---

Closes: #2493

### DIFF
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -5,6 +5,7 @@ import (
 	b64 "encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"strconv"
 	"time"
 
 	errorsng "github.com/pkg/errors"
@@ -97,10 +98,13 @@ If we do have a job ID, then something is in progress and we monitor it until
 it's over.
 If it has ended with an error, we move the volume into a failed state.
 
-Otherwise we proceed with the PersistentVolume creation (which require an extra
-Salt call, synchronous this time, to get the volume size), taking care of
-putting a finalizer on the PersistentVolume (so that its lifetime is tied to
-ours) and setting ourself as the owner of the PersistentVolume.
+Otherwise we make another asynchronous Salt call to get the size of the backing
+storage device (the polling is done exactly as described above).
+
+If we successfully retrive the size, we proceed with the PersistentVolume
+creation, taking care of putting a finalizer on the PersistentVolume (so that
+its lifetime is tied to ours) and setting ourself as the owner of the
+PersistentVolume.
 
 Once we have successfuly created the PersistentVolume, we can move into the
 `Available` state and reschedule the request (the next iteration will check the
@@ -412,18 +416,6 @@ func (self *ReconcileVolume) removeVolumeFinalizer(
 	return self.client.Update(ctx, volume)
 }
 
-// Get the disk size for the volume.
-func (self *ReconcileVolume) getDiskSizeForVolume(
-	ctx context.Context, volume *storagev1alpha1.Volume,
-) (resource.Quantity, error) {
-	nodeName := string(volume.Spec.NodeName)
-	size, err := self.salt.GetVolumeSize(ctx, nodeName, volume.GetPath())
-	if err != nil {
-		return *resource.NewQuantity(0, resource.BinarySI), err
-	}
-	return *resource.NewQuantity(size, resource.BinarySI), err
-}
-
 // Get the PersistentVolume associated to the given volume.
 //
 // Return `nil` if no such volume exists.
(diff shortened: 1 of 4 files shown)


NicolasT (commented):
LGTM but would like more review.

gdemonet (commented):
I'm a bit curious: do we still need a special `JOB_DONE_MARKER` JID value? Wouldn't adding an extra field about the job status to the `JobHandle` struct make things simpler?

Also, do we have "migration" tests? Like what happens if the Operator is updated and picks up an old Volume with an active Job in the old format? Is your `salt.JobFromString` method handling old-style JIDs?

slaperche-scality:
> I'm a bit curious: do we still need a special `JOB_DONE_MARKER` JID value? Wouldn't adding an extra field about the job status to the `JobHandle` struct make things simpler?

As we need to serialize it as a string in the end (to store in in the CR's status), we wouldn't gain much.
On one hand, it may make the code a bit simpler since you can `if job.IsDone` (not sure it's a real benefit, since we still need a to test the ID for the empty/non-empty case anyway).
On the other hand, we would need an extra-field in the string form when we serialize/parse it back and forth (e.g.: `name/ID/isDone/result`).

I think the current solution is OK, albeit hackish since I'm piggybacking stuff into the string field.

But if we really have more needs, I would start over and have a structured `Job` field in the CR's status (but I think this should be done in a new 2.X release since it's changing the CRD, not in a patch release).

> Also, do we have "migration" tests? Like what happens if the Operator is updated and picks up an old Volume with an active Job in the old format? Is your `salt.JobFromString` method handling old-style JIDs?

Currently no.
Current behavior is: `Volume` will goes in Failed state, since we can't parse back the JID, and will need to be recreated.
I think it should be unlikely enough to be acceptable.

It we really want to support it, it may be doable will make the code a bit more complex (because you need to find, somehow, if the current job is `Prepare` or `Unprepare`).

gdemonet (approved):
OK thanks for the explanation. Indeed, it's easier to keep the current approach for now. In any case, the approach for "polling Salt jobs" may change in the future depending on how we make it evolve.
And for migration, I guess it's OK as long as we mention it in release notes.

Issue #2569, opened 2020-05-25
Dex configuration change do not apply properly on running Dex Pods


TeddyAndrieux:
**Component**:

'tests', 'authentication'

**What happened**:

The Dex configuration is stored in a secret, mounted in the Dex Pods. If the secret change the file is automatically changed in the Dex Pods but as Dex process does not restart he still uses the old config.

Like explained in the docs for CSC to apply a change for Dex you need to manually restart the Dex Pods

**What was expected**:

Do not need user action to properly deploy the Dex config changes on all running Dex deployment (handled automatically by the salt state)

This Bug should have been seen in CSC test.

**Steps to reproduce**

Change the Dex secret (or the Dex CSC) and apply Dex salt state.

**Resolution proposal** (optional):

- ~Add an annotations  `metalk8s.scality.com/config-digest`~ Use the `checksum/config` annotation on the Dex Pod, containing a digest of the config file. So that when you apply the salt state and the config content changed the Pod will be re-created automatically. (same as the one we have for static pods)

- Remove the rollout command from the CSC doc

- Add a new test to check that after a CSC update for Dex the change get applied on the running Pods as excepted

PR #2573, opened 2020-05-26
Add digest for dex config

TeddyAndrieux:
**Component**:

'salt', 'authentication'

**Context**: 

#2569 

**Summary**:

- Add simple "salt slots" handling in Kubernetes manifest when using `metalk8s_kubernetes` module
- Add digest of secret to `checksum/config` annotation for Dex pods
- Remove `kubectl rollout restart` command from doc
- Add test for Dex static user password change

---

Fixes: #2569

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -10,6 +10,7 @@ import time
 
 from salt.pillar import get_pillar
 from salt.exceptions import CommandExecutionError
+import salt.utils.args
 import salt.utils.files
 
 log = logging.getLogger(__name__)
@@ -287,3 +288,49 @@ def check_pillar_keys(keys, refresh=True, pillar=None, raise_error=True):
             return False
 
     return True
+
+
+def format_slots(data):
+    """Helper to replace slots in nested dictionnary
+
+    "__slots__:salt:module.function(arg1, arg2, kwarg1=abc, kwargs2=cde)
+
+    Arguments:
+        data: Data structure to format
+    """
+    slots_callers = {
+        "salt": __salt__
+    }
+
+    if isinstance(data, list):
+        return [format_slots(elt) for elt in data]
+
+    if isinstance(data, dict):
+        return {key: format_slots(value) for key, value in data.items()}
+
+    if isinstance(data, basestring) and data.startswith('__slot__:'):
+        fmt = data.split(":", 2)
+        if len(fmt) != 3:
+            log.warning(
+                "Malformed slot %s: expecting "
+                "'__slot__:<caller>:<module>.<function>(...)'",
+                data
+            )
+            return data
+        if fmt[1] not in slots_callers:
+            log.warning(
+                "Malformed slot '%s': invalid caller, must use one of '%s'",
+                data, "', '".join(slots_callers.keys())
+            )
+            return data
+
+        fun, args, kwargs = salt.utils.args.parse_function(fmt[2])
+
+        try:
+            return slots_callers[fmt[1]][fun](*args, **kwargs)
+        except Exception as exc:
+            raise CommandExecutionError(
+                "Unable to compute slot '{}': {}".format(data, exc)
+            )
+
+    return data
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -14,6 +14,7 @@ import re
 
 from salt.exceptions import CommandExecutionError
 from salt.utils import yaml
+import salt.utils.data
 
 MISSING_DEPS = []
 
@@ -354,3 +355,34 @@ def list_objects(kind, apiVersion, namespace='default', all_namespaces=False,
         raise CommandExecutionError('{}: {!s}'.format(base_msg, exc))
 
     return [obj.to_dict() for obj in result.items]
+
+
+def get_object_digest(path=None, checksum='sha256', *args, **kwargs):
+    """
+    Helper to get the digest of one kubernetes object or from a specific key
+    of this object using a path
+    (usefull to get the digest of one config from ConfigMap)
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_kubernetes.get_object_digest kind="ConfigMap" apiVersion="v1" name="my-config-map" path="data:config.yaml"
+        salt-call metalk8s_kubernetes.get_object_digest kind="Pod" apiVersion="v1" name="my-pod"
+    """
+    obj = get_object(*args, **kwargs)
+
+    if not obj:
+        raise CommandExecutionError('Unable to find the object')
+
+    if path:
+        obj = salt.utils.data.traverse_dict_and_list(obj, path, delimiter=':')
+
+        if not obj:
+            raise CommandExecutionError(
+                'Unable to find key "{}" in the object'.format(
+                    path
+                )
+            )
+
+    return __salt__.hashutil.digest(str(obj), checksum=checksum)
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -133,6 +133,9 @@ def _object_manipulation_function(action):
                 )
             )
 
+        # Format slots on the manifest
+        manifest = __salt__.metalk8s.format_slots(manifest)
+
         # Adding label containing metalk8s version (retrieved from saltenv)
         if action in ['create', 'replace']:
             match = re.search(r'^metalk8s-(?P<version>.+)$', saltenv)
--- a/tests/post/features/service_configuration.feature
+++ b/tests/post/features/service_configuration.feature
@@ -7,3 +7,13 @@ Feature: Cluster and Services Configurations
         And we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth'
         When we update the CSC 'spec.deployment.replicas' to '3'
         Then we have '3' at 'status.available_replicas' for 'dex' Deployment in namespace 'metalk8s-auth'
+
+    Scenario: Update Admin static user password
+        Given the Kubernetes API is available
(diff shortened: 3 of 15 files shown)


salt/_modules/metalk8s.py
```diff
@@ -287,3 +288,40 @@ def check_pillar_keys(keys, refresh=True, pillar=None, raise_error=True):
...
+
+    if isinstance(data, basestring) and data.startswith('__slot__:'):
+        fmt = data.split(":", 2)
+        if len(fmt) != 3:
+            log.warning("Malformed slot: %s", data)
+            return data
+        if fmt[1] not in slots_callers:
+            log.warning(
+                "Malformed slot, only %s supported: %s",
+                slots_callers.keys(), data
```
gdemonet: The ordering of values makes things weird to read IMHO:

```suggestion
                "Malformed slot '%s': invalid caller, must use one of '%s'",
                data, "', '".join(slots_callers.keys())
```

salt/_modules/metalk8s.py
```diff
@@ -287,3 +288,40 @@ def check_pillar_keys(keys, refresh=True, pillar=None, raise_error=True):
...
+    if isinstance(data, list):
+        return [format_slots(elt) for elt in data]
+
+    if isinstance(data, dict):
+        return {key: format_slots(value) for key, value in data.items()}
+
+    if isinstance(data, basestring) and data.startswith('__slot__:'):
+        fmt = data.split(":", 2)
+        if len(fmt) != 3:
+            log.warning("Malformed slot: %s", data)
```
gdemonet: (minor) Could provide the expected format in this log line

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -351,3 +354,37 @@ def list_objects(kind, apiVersion, namespace='default', all_namespaces=False,
         raise CommandExecutionError('{}: {!s}'.format(base_msg, exc))
 
     return [obj.to_dict() for obj in result.items]
+
+
+def get_object_digest(object_key=None, checksum='sha256', *args, **kwargs):
```
gdemonet: Maybe just `key` as an argument?

salt/_modules/metalk8s.py
```diff
@@ -287,3 +288,40 @@ def check_pillar_keys(keys, refresh=True, pillar=None, raise_error=True):
...
+        if fmt[1] not in slots_callers:
+            log.warning(
+                "Malformed slot, only %s supported: %s",
+                slots_callers.keys(), data
+            )
+            return data
+
+        fun, args, kwargs = salt.utils.args.parse_function(fmt[2])
+
+        return slots_callers[fmt[1]][fun](*args, **kwargs)
```
gdemonet: Shouldn't we catch errors from this expression? Like, what if the function isn't in `__salt__`? What if the function raises ValueError/TypeError because we're calling it wrong? Maybe it's fine to leave `CommandExecutionError`s pass through, but some "classic" errors like the ones mentioned above would benefit from pre-formatted error messages for the user.

salt/_modules/metalk8s_kubernetes.py
gdemonet: Oh you're actually using a full colon-separated path on the object (I see `data:config.yaml`, when I was expecting `config.yaml`). Maybe `path` is better then.

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -351,3 +354,37 @@ def list_objects(kind, apiVersion, namespace='default', all_namespaces=False,
...
+
+        for key in object_key:
+            try:
+                obj = obj[key]
+            except KeyError:
+                raise CommandExecutionError(
+                    'Unable to find key "{}" in the object'.format(
+                        '.'.join(object_key)
+                    )
+                )
```
gdemonet: Can't we use Salt utils for this?

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -351,3 +354,37 @@ def list_objects(kind, apiVersion, namespace='default', all_namespaces=False,
         raise CommandExecutionError('{}: {!s}'.format(base_msg, exc))
 
     return [obj.to_dict() for obj in result.items]
+
+
+def get_object_digest(object_key=None, checksum='sha256', *args, **kwargs):
+    """
+    Helper to get the digest of one kubernetes object or from a specific key
+    of this object (usefull to get the digest of one config from ConfigMap)
```
gdemonet: One single object may not be sufficient in all cases: some Pods will for instance consume configuration data from both a Secret and a ConfigMap. But acceptable to keep as a debt ticket, as we don't need it for Dex.

gdemonet (commented):
Those slots look cool :100:!

I'm a bit concerned of the flexibility however:
- Hashing string(s) in a slot is a good idea, could be reused for other sources (like local files), but not in the current approach which is tied to K8s
- Getting more than a single K8s object will very likely become a need, as workloads often consume data from both ConfigMap(s) and Secret(s)

TeddyAndrieux:
> I'm a bit concerned of the flexibility however:

I don't see why

> * Hashing string(s) in a slot is a good idea, could be reused for other sources (like local files), but not in the current approach which is tied to K8s

Why ? For local file you already have `hashutil.digest_file` from salt you do not need this function to handle it

> * Getting more than a single K8s object will very likely become a need, as workloads often consume data from both ConfigMap(s) and Secret(s)

Then we will just need a new function that do several call on `metalk8s_kubernetes.get_object_digest` but right now it's not needed

salt/_modules/metalk8s.py
TeddyAndrieux: sure

salt/_modules/metalk8s_kubernetes.py
TeddyAndrieux: Hmm yes `path` make sense

salt/_modules/metalk8s_kubernetes.py
TeddyAndrieux: The goal of this function is to retrieve one if you need to retrieve several then you should just do several calls on this one (and maybe create another module function to do it if you need to).

To me it's not a debt and not an issue it's just a feature we do not need right now so ....

salt/_modules/metalk8s.py
TeddyAndrieux: Sure I can, just so you know here the salt implementation of slots https://github.com/saltstack/salt/blob/master/salt/state.py#L2285-L2306

salt/_modules/metalk8s_kubernetes.py
TeddyAndrieux: Which one ?

gdemonet: [This one](https://github.com/saltstack/salt/blob/master/salt/utils/data.py#L785)?

salt/_modules/metalk8s_kubernetes.py
gdemonet: We're kind of forced to have an extensible method if we need more than one config to hash, as the hash should be stored in a single annotation. OK for the debt topic, let's wait for a need to implement this feature, you're right :)

gdemonet (dismissed):
:shipit:

salt/_modules/metalk8s_kubernetes.py
TeddyAndrieux: Right, I will

gdemonet (dismissed):

tests/post/features/service_configuration.feature
```diff
@@ -7,3 +7,14 @@ Feature: Cluster and Services Configurations
         And we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth'
         When we update the CSC 'spec.deployment.replicas' to '3'
         Then we have '3' at 'status.available_replicas' for 'dex' Deployment in namespace 'metalk8s-auth'
+
+  Scenario: Update Admin static user password
+        Given: the Kubernetes API is available
```
alexandre-allard:
```suggestion
        Given the Kubernetes API is available
```

TeddyAndrieux: Ohhh right good catch :+1:

gdemonet (dismissed):

alexandre-allard (dismissed):

tests/post/features/service_configuration.feature
```diff
@@ -7,3 +7,14 @@ Feature: Cluster and Services Configurations
...
+  Scenario: Update Admin static user password
+        Given the Kubernetes API is available
+        And we have a 'metalk8s-dex-config' CSC in namespace 'metalk8s-auth'
+        # NOTE: Consider that admin user is the first of the userlist
+        # Update password to "new-password"
+        When we update the CSC 'spec.localuserstore.userlist.0.hash' to '$2y$14$pDe0vj917rR3XJQ5iEZvhuSGoWkZg2/qBN/mMLqwFSz9S7EYcbIpO'
+        Then the control-plane Ingress path '/oidc' is available
+        And pods with label 'app.kubernetes.io/name=dex' are 'Ready'
+        And we are not able to login to Dex as 'admin@metalk8s.invalid' using password 'password'
+        And we are able to login to Dex as 'admin@metalk8s.invalid' using password 'new-password'
```
alexandre-allard: The corresponding methods only exist in `tests/post/steps/test_authentication.py`, I don't think it'll work.
You should either put them in `tests/conftest.py` to make these `Then` available to all your scenario, or declare them in `tests/post/steps/test_service_configuration.py`.
Tests have passed probably because of the typo you had on the first line `Given: the Kubernetes API is available` (https://github.com/scality/metalk8s/pull/2573#discussion_r430894649)

TeddyAndrieux: Right, cool pytest-bdd feature (typo = success) .... 
Anyway ok should be fixed but I don't know if my test work as he actually never run for the moment

gdemonet (approved):

alexandre-allard (approved):

Issue #2601, opened 2020-06-05
Flaky during bootstrap restore


TeddyAndrieux:
**Component**:

'salt', 'restore'

**What happened**:

During bootstrap restore some time we are not able to mark the new bootstrap node as this one does not exist (as kubelet not registered yet)

```
----------
          ID: Mark control plane node
    Function: metalk8s_kubernetes.object_updated
        Name: bootstrap
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1981, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1977, in wrapper
                  return f(*args, **kwargs)
                File "/var/cache/salt/master/extmods/states/metalk8s_kubernetes.py", line 205, in object_updated
                  **kwargs
                File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 214, in method
                  return _handle_error(exc, action)
                File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 72, in _handle_error
                  raise CommandExecutionError(base_msg + str(exception).decode('utf-8'))
              CommandExecutionError: Failed to update object: (404)
              Reason: Not Found
              HTTP response headers: HTTPHeaderDict({'date': 'Fri, 05 Jun 2020 09:09:21 GMT', 'content-length': '188', 'content-type': 'application/json', 'cache-control': 'no-cache, private'})
              HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"nodes \"bootstrap\" not found","reason":"NotFound","details":{"name":"bootstrap","kind":"nodes"},"code":404}
     Started: 09:09:21.731846
    Duration: 40.144 ms
     Changes:
```

**What was expected**:

No fail

**Steps to reproduce**

Flaky

**Resolution proposal** (optional):

Add a retry on this specific state

PR #2610, opened 2020-06-08
salt: Ensure node exists before marking it as bootstrap

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

#2601 

**Summary**:

In restore script, we configure a new bootstrap node according to a
backup file and just after kubelet installation we try to mark the node
as bootstrap node but kubelet may take some time to register so the
salt state may fail.
This commit adds a state to check that the node exists with a retry logic
in case kubelet not register yet when calling the salt state

---

Fixes: #2601

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -299,6 +299,22 @@ get_object = _object_manipulation_function('retrieve')
 update_object = _object_manipulation_function('update')
 
 
+# Check if a specific object exists
+def object_exists(kind, apiVersion, name, **kwargs):
+    """
+    Simple helper to check if an object exists or not
+
+    CLI Examples:
+
+    .. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


salt/_modules/metalk8s_kubernetes.py
```diff
@@ -299,6 +299,22 @@ def method(manifest=None, name=None, kind=None, apiVersion=None,
...
+    """
+    Simple helper to check if an object exists or not
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt-call metalk8s_kubernetes.object_exists kind="Node" apiVersion="v1" name="MyNode"
+    """
+    return bool(
```
gdemonet: Any falsy result outside of `None` would be a bug, so let's make this method somewhat more obvious:

```python
return get_object(...) is not None
```

gdemonet (changes requested):
Small change on the helper, otherwise LGTM!

gdemonet (dismissed):

gdemonet (approved):

alexandre-allard (approved):

Issue #2621, opened 2020-06-12
Delete CustomObject does not work


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**What happened**:

When we try to delete a CustomObject we get a traceback (for example during upgrade)
```
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1981, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1977, in wrapper
                  return f(*args, **kwargs)
                File "/var/cache/salt/master/extmods/states/metalk8s_kubernetes.py", line 68, in object_absent
                  **kwargs)
                File "/var/cache/salt/master/extmods/modules/metalk8s_kubernetes.py", line 212, in method
                  result = method_func(**call_kwargs)
                File "/var/cache/salt/master/extmods/utils/kubernetes_utils.py", line 442, in method
                  result = base_method(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/kubernetes/client/api/custom_objects_api.py", line 586, in delete_namespaced_custom_object
                  (data) = self.delete_namespaced_custom_object_with_http_info(group, version, namespace, plural, **kwargs)  # noqa: E501
                File "/usr/lib/python2.7/site-packages/kubernetes/client/api/custom_objects_api.py", line 625, in delete_namespaced_custom_object_with_http_info
                  " to method delete_namespaced_custom_object" % key
              TypeError: Got an unexpected keyword argument 'name' to method delete_namespaced_custom_object
```

**What was expected**:

Working delete function

**Steps to reproduce**

Try to delete a CustomObject with Salt

**Resolution proposal** (optional):

The function used to delete a CustomObject is wrong we should not use `delete_cluster_custom_object` or `delete_scoped_custom_object` but instead use `delete_cluster_custom_object_0` and `delete_scoped_custom_object_0`

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#delete_cluster_custom_object
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#delete_cluster_custom_object_0

NOTE it's only the case because we have a snapshot version of `python-kubernetes` no released version have this 2 different function right now.
02047bbc68a1a454833e93f3dcd028106c37708e

PR #2622, opened 2020-06-12
salt: Fix `delete` for CustomObject in `metalk8s_kubernetes`

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Context**: 

`delete` of CustomObject in our custom Salt module `metalk8s_kubernetes`
does not work as we use `delete_namespaced_custom_object` function that
does not take `name` argument instead of
`delete_namespaced_custom_object_0` that take `name`.

**Summary**:

This commit change the function used to delete a CustomObject

---

Fixes: #2621

### DIFF
--- a/salt/_utils/kubernetes_utils.py
+++ b/salt/_utils/kubernetes_utils.py
@@ -452,6 +452,18 @@ class CustomApiClient(ApiClient):
 
         return method
 
+    def _method_name(self, verb):
+        # Override super(_method_name) for `delete` as two different functions
+        # exist to delete a custom object, one that take a `name` argument
+        # and another one that does not
+        # In our case we want to use the one with `name`
+        # See: https://github.com/scality/metalk8s/issues/2621
+        # NOTE: This may need to be removed once we will change the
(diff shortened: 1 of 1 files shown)


alexandre-allard (dismissed):

salt/_utils/kubernetes_utils.py
```diff
@@ -457,6 +457,16 @@ def method(*args, **kwargs):
 
         return method
 
+    def _method_name(self, verb):
+        # Override super(_method_name) for `delete` as two different function
+        # exists to delete a custom object, one that take `name` argument
+        # and another one that didn't
+        # In our case we want to use the one with `name`
+        # Sees: https://github.com/scality/metalk8s/issues/2621
```
alexandre-allard:
```suggestion
        # Override super(_method_name) for `delete` as two different functions
        # exist to delete a custom object, one that take a `name` argument
        # and another one that does not.
        # In our case, we want to use the one with `name`
        # See: https://github.com/scality/metalk8s/issues/2621
```

alexandre-allard (approved):

PR #2623, opened 2020-06-12
eve: replace old uis repository in example-solution builder

alexandre-allard:
**Component**: eve

**Context**:
Solution example cannot be built anymore in the CI

**Summary**:
UIS repository is used to install git2 on CentOS7 because otherwise the latest version available is 1.8.3.1 which is not compatible with the operator-sdk.

The repository has recently changed its URL, so we update it accordingly.
Plus, the package name has changed from git2u to git224.

We also add an option to yum install for it to fail when it cannot find a requested package, avoiding
ourselves to lost time on searching why the build is not working anymore (just because git was not in the good version).

**Acceptance criteria**:
We can build the Solution Example from the CI (in post-merge)

### DIFF
--- a/eve/workers/pod-example-solution-builder/Dockerfile
+++ b/eve/workers/pod-example-solution-builder/Dockerfile
@@ -5,12 +5,12 @@ ARG OPERATOR_SDK_VERSION=v0.16.0
 
 WORKDIR /home/eve/workspace
 
-RUN yum install -y \
+RUN yum install -y --setopt=skip_missing_names_on_install=False \
     epel-release \
     https://repo.ius.io/ius-release-el7.rpm \
     && yum-config-manager --add-repo \
     https://download.docker.com/linux/centos/docker-ce.repo \
-    && yum install -y \
+    && yum install -y --setopt=skip_missing_names_on_install=False \
     sudo \
     gcc \
     hardlinks \
--- a/eve/workers/pod-example-solution-builder/Dockerfile
+++ b/eve/workers/pod-example-solution-builder/Dockerfile
@@ -13,7 +13,7 @@ RUN yum install -y --setopt=skip_missing_names_on_install=False \
     && yum install -y --setopt=skip_missing_names_on_install=False \
     sudo \
     gcc \
-    hardlinks \
+    hardlink \
     make \
     genisoimage \
     golang \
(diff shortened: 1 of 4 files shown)


eve/workers/pod-example-solution-builder/Dockerfile
```diff
@@ -5,25 +5,25 @@ ARG OPERATOR_SDK_VERSION=v0.16.0
 
 WORKDIR /home/eve/workspace
 
-RUN yum install -y \
+RUN yum install -y --setopt=skip_missing_names_on_install=False \
```
Ebaneck, line 8: Not sure we need this extra check here!
should `yum` not automatically fail if it cannot find any of the packages

eve/workers/pod-example-solution-builder/Dockerfile
```diff
@@ -5,25 +5,25 @@ ARG OPERATOR_SDK_VERSION=v0.16.0
...
     genisoimage \
-    git2u \
     golang \
     python-devel \
     python-pip \
     skopeo \
     yum-utils \
+    yum-plugin-replace \
     docker-ce-cli-18.09.6 \
+    && yum replace -y git --replace-with git224 \
```
Ebaneck: Seems the `IUS` repo does not have the `git` package at all...

And if we must pin a git version may we just use `git224-all`?
Source: https://repo.ius.io/7/x86_64/packages/g/

Ebaneck (commented):
Few comments!

eve/workers/pod-example-solution-builder/Dockerfile
Ebaneck: Hmm, seems yum will not fail when it cannot find a package from the list provided so let's keep it.

eve/workers/pod-example-solution-builder/Dockerfile
alexandre-allard: No, git package comes from CentOS repositories.
That's basically what's done here.

eve/workers/pod-example-solution-builder/Dockerfile
alexandre-allard: Yup, that's why I've added this option, the build would have failed here with a clear error message if we had this option set.

Ebaneck: ok

alexandre-allard: Well... not a clear message because it would have failed when provisioning the worker, so it returns nothing... but at least it's easily reproducible, just by building this Dockerfile.

Ebaneck (approved):
Ship it!

PR #2625, opened 2020-06-12
Feature/2615 annotate pv with device name

### LINKED ISSUE #2615: Annotate PV with the name of the underlying block device
**Component**:

operator, salt

**Why this is needed**:

Prometheus exports, through node-exporter, some I/O related stats.
Those stats are labeled with the name of the block device, but we would like to show them on the Volume dashboard.
This proved to be difficult, because the path used to create the PV may be an alias (like `dev/disk/by-uuid/…`) which the UI can't resolve into an actual block device name, and thus can't find the right stats for the Volume.

**What should be done**:

The storage-operator should resolve the name of the underlying block device and store it somewhere the UI can retrieve it.

When a PV is created, the storage-operator should resolve the name of the underlying block device (using Salt) and annotate the PV with it (using `storage.metalk8s.scality.com/device`, and set it to `sda` for instance).
This annotation must be updated during the reconciliation, as the name of the underlying block device may change in some occasion (e.g. reboot, disk replacement, …)

**Test plan**:

Add a step to existing tests that create a volume to ensure that the annotation is added to the PV.

slaperche-scality:
**Component**:

salt, operator

**Context**: 

In order to expose I/O metrics we need a way to link a PV/Volume (which is what we expose in our UI) to the underlying block device (which is the label used to export I/O metrics). 

**Summary**:

Add a new annotation on the PV created by our operator.
This annotation contains the name of the underlying block device and is refreshed at each reconciliation iteration (because it may change in some occasion, like a reboot).

**Acceptance criteria**: 

When you create a volume, the PV should have an annotation `storage.metalk8s.scality.com/device` providing the name of the underlying block device.
A test was added to check that and should pass.

---

Closes: #2615

### DIFF
--- a/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
+++ b/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
@@ -45,6 +45,7 @@ external_auth:
     storage-operator:
       - '*':
         - 'disk.dump'
+        - 'metalk8s_volumes.device_name'
         - 'state.sls'
       - '@jobs'
 
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -719,11 +719,16 @@ func (self *ReconcileVolume) refreshDeviceName(
 	volume *storagev1alpha1.Volume,
 	pv *corev1.PersistentVolume,
 ) (reconcile.Result, error) {
-	path := pv.Spec.PersistentVolumeSource.Local.Path
 	nodeName := string(volume.Spec.NodeName)
 	reqLogger := log.WithValues(
 		"Volume.Name", volume.Name, "Volume.NodeName", nodeName,
 	)
+
+	if pv.Spec.PersistentVolumeSource.Local == nil {
+		reqLogger.Info("skipping volume: not a local storage")
+		return endReconciliation()
+	}
+	path := pv.Spec.PersistentVolumeSource.Local.Path
 
 	name, err := self.salt.GetDeviceName(ctx, nodeName, volume.Name, path)
 	if err != nil {
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -201,7 +201,7 @@ reclaim its storage and remove the finalizers to let the object be deleted.
 
 }}} */
 
-const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device"
+const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device-name"
 const VOLUME_PROTECTION = "storage.metalk8s.scality.com/volume-protection"
 const JOB_DONE_MARKER = "DONE"
 
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -201,6 +201,7 @@ reclaim its storage and remove the finalizers to let the object be deleted.
 
 }}} */
 
+const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device"
 const VOLUME_PROTECTION = "storage.metalk8s.scality.com/volume-protection"
 const JOB_DONE_MARKER = "DONE"
 
@@ -633,7 +634,7 @@ func (r *ReconcileVolume) Reconcile(request reconcile.Request) (reconcile.Result
 		return delayedRequeue(err)
 	}
 	reqLogger.Info("backing PersistentVolume is healthy")
-	return endReconciliation()
+	return r.refreshDeviceName(ctx, volume, pv)
 }
 
 // Deploy a volume (i.e prepare the storage and create a PV).
@@ -711,6 +712,38 @@ func (self *ReconcileVolume) finalizeVolume(
 
 	// PersistentVolume still in use: wait before reclaiming the storage.
 	return delayedRequeue(nil)
+}
+
+func (self *ReconcileVolume) refreshDeviceName(
+	ctx context.Context,
+	volume *storagev1alpha1.Volume,
+	pv *corev1.PersistentVolume,
+) (reconcile.Result, error) {
+	path := pv.Spec.PersistentVolumeSource.Local.Path
+	nodeName := string(volume.Spec.NodeName)
+	reqLogger := log.WithValues(
+		"Volume.Name", volume.Name, "Volume.NodeName", nodeName,
+	)
+
+	name, err := self.salt.GetDeviceName(ctx, nodeName, volume.Name, path)
+	if err != nil {
+		self.recorder.Event(
+			volume, corev1.EventTypeNormal, "SaltCall",
+			"device path resolution failed",
+		)
+		reqLogger.Error(err, "cannot get device name from Salt response")
+		return delayedRequeue(err)
+	}
+	if pv.ObjectMeta.Annotations[DEVICE_ANNOTATION] != name {
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, DEVICE_ANNOTATION, name)
+		if err := self.client.Update(ctx, pv); err != nil {
+			reqLogger.Error(err, "cannot update device name: requeue")
+			return delayedRequeue(err)
+		}
+		reqLogger.Info("update device name", "Volume.DeviceName", name)
+	}
+
+	return endReconciliation()
 }
 
 func (self *ReconcileVolume) prepareStorage(
--- a/storage-operator/pkg/controller/volume/volume_controller.go
+++ b/storage-operator/pkg/controller/volume/volume_controller.go
@@ -201,7 +201,6 @@ reclaim its storage and remove the finalizers to let the object be deleted.
 
 }}} */
 
-const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device-name"
 const VOLUME_PROTECTION = "storage.metalk8s.scality.com/volume-protection"
 const JOB_DONE_MARKER = "DONE"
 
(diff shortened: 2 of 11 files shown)


storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -201,6 +201,7 @@ reclaim its storage and remove the finalizers to let the object be deleted.
 
 }}} */
 
+const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device"
```
NicolasT: Considered `.../device-name`?

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -713,6 +714,38 @@ func (self *ReconcileVolume) finalizeVolume(
 	return delayedRequeue(nil)
 }
 
+func (self *ReconcileVolume) refreshDeviceName(
+	ctx context.Context,
+	volume *storagev1alpha1.Volume,
+	pv *corev1.PersistentVolume,
+) (reconcile.Result, error) {
+	path := pv.Spec.PersistentVolumeSource.Local.Path
```
NicolasT: Maybe have an assertion that the given PV is actually of type `Local`, i.e. `Path` is not `""`?

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -713,6 +714,38 @@ func (self *ReconcileVolume) finalizeVolume(
...
+	if pv.ObjectMeta.Annotations[DEVICE_ANNOTATION] != name {
+		metav1.SetMetaDataAnnotation(&pv.ObjectMeta, DEVICE_ANNOTATION, name)
+		if err := self.client.Update(ctx, pv); err != nil {
+			reqLogger.Error(err, "cannot update device name: requeue")
+			return delayedRequeue(err)
+		}
+		reqLogger.Info("update device name", "DeviceName", name)
+	}
+
+	return endReconciliation()
```
NicolasT, line 747: This implies this call is made at the end of a reconciliation loop, right? Which is not necessarily the case?

In general, shouldn't only `Reconcile` ever return a `reconcile.Result`, and (e.g.) this one an `error`, and let `Reconcile` translate this in a `delayedRequeue` or whatever?

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: Yup, I'll rename.

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: It's supposed to be PV we created no?
But will do, it doesn't hurt to be extra-safe on that point.

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: > This implies this call is made at the end of a reconciliation loop, right? Which is not necessarily the case?

Not necessarily.
It may be ill-named, but the role of this function is to return a `reconcile.Result` that end the reconciliation (i.e. doesn't trigger a requeue), unlike `requeue` and `delayedRequeue`.

> In general, shouldn't only `Reconcile` ever return a `reconcile.Result`, and (e.g.) this one an `error`, and let `Reconcile` translate this in a `delayedRequeue` or whatever?

Hum, that would be like a global try/catch, where we inspect the "exception" and decide how to react to it.
Could be done, but would need to define a bunch of error types to be able to tell the difference between the different sources of error.
Could be done, but deserve a dedicated issue.

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -201,7 +201,7 @@ reclaim its storage and remove the finalizers to let the object be deleted.
 
 }}} */
 
-const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device"
+const DEVICE_ANNOTATION = "storage.metalk8s.scality.com/device-name"
```
NicolasT: Now thinking of it... Should this be an annotation, or rather in `status`? :thinking: WDYT @scality/metalk8s

storage-operator/pkg/controller/volume/volume_controller.go
NicolasT: > > This implies this call is made at the end of a reconciliation loop, right? Which is not necessarily the case?
> 
> Not necessarily.
> It may be ill-named, but the role of this function is to return a `reconcile.Result` that end the reconciliation (i.e. doesn't trigger a requeue), unlike `requeue` and `delayedRequeue`.

I know. However, imagine this were *not* the last step/call in `Reconcile`. Now in order to know whether to continue with the next operation or ask for a re-enqueue without running the 'next step', `Reconcile` would need to look 'inside' this `reconcile.Result`...

> 
> > In general, shouldn't only `Reconcile` ever return a `reconcile.Result`, and (e.g.) this one an `error`, and let `Reconcile` translate this in a `delayedRequeue` or whatever?
> 
> Hum, that would be like a global try/catch, where we inspect the "exception" and decide how to react to it.
> Could be done, but would need to define a bunch of error types to be able to tell the difference between the different sources of error.
> Could be done, but deserve a dedicated issue.

Oh, for sure, out of scope.

Maybe the API could be as follows: instead of `(reconcile.Result, error)`, the functions could *still* return `(reconcile.Result, error)`, but the `Result` *must* be `nil` when `error` is `nil`, and may *not* be `nil` when `error` is not `nil`. As such, the `Result` is returned after the `error` is logged on error, otherwise `Reconcile` continues whatever it needs to do next, and in the end returns its own 'success' `Result`.

slaperche-scality: Would be better yeah, I'll open a debt ticket for that.

storage-operator/pkg/controller/volume/volume_controller.go
gdemonet: IMO, that's what `status` is for, and since we own it... I'd vote for using it. Feels minor however, and we could easily convert between one and the other in the future, if need be.

NicolasT: Except that then the UI would need to dispatch between multiple possible locations :)

gdemonet: Tie this schema to an `apiVersion`?

slaperche-scality: Done, last commit move things into status.

gdemonet: Oh thanks, was not sure if it needed to go in this PR.

storage-operator/pkg/controller/volume/volume_controller.go
```diff
@@ -713,6 +713,40 @@ func (self *ReconcileVolume) finalizeVolume(
...
+		self.recorder.Event(
+			volume, corev1.EventTypeNormal, "SaltCall",
+			"device path resolution failed",
+		)
+		reqLogger.Error(err, "cannot get device name from Salt response")
+		return delayedRequeue(err)
+	}
+	if volume.Status.DeviceName != name {
+		volume.Status.DeviceName = name
+		reqLogger.Info("update device name", "DeviceName", name)
```
gdemonet: I'm not familiar with how structured logging should work with controller-runtime, but given the `WithValues` call above, shouldn't we also namespace `DeviceName`?

```suggestion
		reqLogger.Info("update device name", "Volume.Status.DeviceName", name)
```

storage-operator/pkg/salt/client.go
```diff
@@ -232,25 +205,73 @@ func (self *Client) GetVolumeSize(
...
-		"GetVolumeSize", "Volume.NodeName", nodeName, "Volume.Name", volumeName,
+		jobName, "Volume.NodeName", nodeName, "Volume.Name", volumeName,
+	)
+
+	return self.submitJob(ctx, jobName, volumeName, payload)
+}
+
+// Return the name of the block device designed by `devicePath`.
+//
+// This request is synchronous.
```
gdemonet, line 216: Back to using synchronous requests against SaltAPI? Isn't it a recipe for failure?

gdemonet (dismissed):
Some minor questions, can safely be handled in follow-up tickets :)

storage-operator/pkg/controller/volume/volume_controller.go
slaperche-scality: I could "namespace" it yeah, I'll use `Volume.DeviceName` (I think the fact that this lives inside the Status sub-resource is an implementation detail, the one inspecting the log doesn't need this knowledge).

storage-operator/pkg/salt/client.go
slaperche-scality: As I tried to explain in the commit message, I think here it's acceptable because unlike `disk.dump` (which is in the "critical path" of volume creation):

- it doesn't prevent the volume to be used: a failed `disk.dump` can block/delay the creation of the volume, where a failed `metalk8s_volume.device_name` won't change the status/avaibility of the volume (volume is already available and will stay that way, worst case is we'll miss some state in the UI until the call succeed, but this doesn't prevent the volume to be ready and used).
- it's always retried on failure, no matter the "type" of error: unlike `disk.dump` (which only retry on some error and "abort" on some others), `metalk8s_volumes.device_name` is always retried on error

Given those difference, I think we can use a sync call (and have a big code simplification), without the drawbacks we used to have with `disk.dump`.

gdemonet: Sorry I missed the commit message, OK for me!

gdemonet (dismissed):

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
```diff
@@ -45,6 +45,7 @@ external_auth:
     storage-operator:
       - '*':
         - 'disk.dump'
+        - 'metalk8s_volumes.device_name'
```
slaperche-scality, line 48: Weirdly enough, it was working fine in 2.5 without that, but was failing in 2.6 with a 401 error from SaltAPI (as expected, since the permission was missing).

gdemonet: That's a bug, here's a ticket to investigate: #2634

gdemonet (approved):

PR #2627, opened 2020-06-15
CI: Split post-merge in two stages

alexandre-allard:
**Component**: ci (post-merge)

**Context**: 
We were building previous version and example-solution in a first step, then
triggering tests on solutions and upgrade/downgrade of the platform as a second step.
If, for example, the building of example solution failed, we weren't doing the upgrade/downgrade tests, while they're not related.

**Summary**:
We now split the post-merge in two distinct parts, one for lifecycle (build previous version, upgrade, downgrade, restore) and one for solution (build solution and deploy it).
So, if one of the two builds fails, the tests for the other are still run.

This has been done to avoid losing all the benefits of the post-merge when a single component is failing.

It also reduces time needed for the post-merge tests as the Example Solution build can take ages.

**Acceptance criteria**: post-merge succeed

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -512,6 +512,34 @@ stages:
           haltOnFailure: True
       - SetPropertyFromCommand: *set_version_prop
       - SetPropertyFromCommand: *set_short_version_prop
+      - TriggerStages:
+          name: Trigger post-merge lifecycle and solutions stages simultaneously
+          stage_names:
+            - post-merge-lifecycle
+            - post-merge-solutions
+          haltOnFailure: true
+      - ShellCommand: *add_final_status_artifact_success
+      - Upload: *upload_final_status_artifact
+      - TriggerStages:
+          name: Trigger TestRail objects creation and results upload
+          stage_names:
+            - create-upload-testrail-objects
+          alwaysRun: True
+
+  post-merge-solutions:
+    worker:
+      type: local
+    steps:
+      - TriggerStages:
+          name: Trigger build of Example Solution archive
+          stage_names:
+            - build-example-solution
+          haltOnFailure: true
+
+  post-merge-lifecycle:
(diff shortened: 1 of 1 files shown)


eve/main.yml
```diff
@@ -512,6 +512,34 @@ stages:
...
+
+  post-merge-solutions:
+    worker:
+      type: local
+    steps:
+      - TriggerStages:
+          name: Trigger build of Example Solution archive
+          stage_names:
+            - build-example-solution
+          haltOnFailure: true
```
alexandre-allard, line 537: It's a stage triggering a single stage, it seems pretty useless, but it's done this way for the code to be similar between 2.4 and 2.5+, so we don't have conflicts if we change something in 2.4.
Because, in later versions, we also trigger Solutions frameworks tests right after this stage.

gdemonet: Not yet though, right (at least, can't see any difference in integration branches)?

gdemonet:
FWIW, this approach should work, until we need to do e.g. upgrade Metal while maintaining a Solution active. We need a real dependency graph, not just this manual handling provided by Eve.

alexandre-allard:
> FWIW, this approach should work, until we need to do e.g. upgrade Metal while maintaining a Solution active. We need a real dependency graph, not just this manual handling provided by Eve.

I agree, this is just a quick improvement for now.

eve/main.yml
alexandre-allard: Now it's done: https://github.com/scality/metalk8s/blob/w/2.6/improvement/rework-post-merge-stages/eve/main.yml

gdemonet (approved):

alexandre-allard:
Post merge tests look great (except some flakies.. as always), let's merge this.

PR #2630, opened 2020-06-17
WIP: Log centralization design document

### LINKED ISSUE #2626: Write design document for Centralized Logging
**Component**:

'logging', 'design', 'kubernetes'

**Why this is needed**:

- Need to provide a design document that satisfies the requirements as documented here: https://github.com/scality/metalk8s/pull/2556/

**What should be done**:

- Write a design document that answers to the requirements.
- Should be validated by Architect & Team

**Test plan**:

Ebaneck:
**Component**:

'logs', 'monitoring', 'design'

**Context**: 

see #2626 

**Summary**:

- A first design draft of the log centralization service. Let's use this to schedule a design session with the team.

**Acceptance criteria**: 

---

Closes: #2626


docs/developer/architecture/logging.rst
```diff
@@ -0,0 +1,235 @@
+Logging
+=======
+
+This section summarizes centralized logging design architecture which is closely
+linked to the above requirements (to be merged).
+
+The following terms will be used throughout the context of this document so it
+is fair that we describe their meanings.
```
gdemonet: Not just fair, useful, if not mandatory ;)

docs/developer/architecture/logging.rst
```diff
@@ -0,0 +1,235 @@
...
+
+Collector/Shipper
```
gdemonet, line 10: OK for this one, "collector" is better than "shipper" IMO (it's the tool that "knows where to gather logs from and where to send them", whereas "shipping" may mean it knows how to distribute to multiple consumers).

docs/developer/architecture/logging.rst
```diff
@@ -0,0 +1,235 @@
...
+-----------------
+A log Collector allows you to collect logs from different sources, unify
+and send them to multiple destinations. An example of such a system is the
+Linux utility Rsyslog
+
+Distributor
+-----------
+A log Distributor is a utility capable of adding metadata to logs and
```
gdemonet, line 18: "Distributor" feels strange as a word for this: I would expect such a component is only responsible of handling the dispatching of logs received from the "Collector" to one or more consumers (so, that relates to the "Shipper" concept from above). Its responsibility could also be to filter logs (based on some metadata?), to deduplicate them, to forward only part of them to a specific consumer, etc.

IIRC, the usual term for such a component is "Aggregator", but I may be wrong. In any case, I think enriching a log stream should be made by another component: either at the Collector level, or at the Database level.

docs/developer/architecture/logging.rst
```diff
@@ -0,0 +1,235 @@
+Logging
+=======
+
+This section summarizes centralized logging design architecture which is closely
+linked to the above requirements (to be merged).
+
+The following terms will be used throughout the context of this document so it
```
gdemonet, line 7: Maybe we're missing one key component: the Database (or Storage, whichever name fits).
It would help to "see" the pipeline (ASCII schema can be sufficient):

```
Collector -> Aggregator -> Database -> Visualizer
```

Each component can be scaled and scheduled independently.

gdemonet: Also, in this document, it's important to consider these "components" and the split between them as **part of the design**. Nothing forces us to think of this log centralization system in this manner, it's a choice to match our requirements.

docs/developer/architecture/logging.rst
```diff
@@ -0,0 +1,235 @@
...
+Distributor
+-----------
+A log Distributor is a utility capable of adding metadata to logs and
+forwarding these logs to unique sources.
+The metadata (labels) set on log streams are used by log visualizers for
+querying purposes. An example of such system is Logstash.
+
+Visualizer
+----------
+A log Visualizer allows one to query and visualize logs. This can be possible
```
gdemonet, line 34: Query as a part of the "Visualizer"? That would be part of the Database IMO, or, if we really want to decorrelate them, an extra component (one that would maintain extra indices, for instance).

gdemonet (commented):
Didn't look at the implementation details yet, but some comments on the terminology used and concepts presented. It's a good idea to clarify those concepts first, but keep in mind that the choice of components and how they work together is not **just** something to explain, it's part of the design.

PR #2632, opened 2020-06-17
salt: raise if solution manifest does not exist

alexandre-allard:

**Context**:
When trying to deploy old Solution archive, without the manifest.yaml file, salt crashes with a not really clear error.

**Summary**:
We now have a clear error message when there is no manifest.yaml found at the root of the Solution archive.

**Acceptance criteria**:

### DIFF
--- a/salt/_modules/metalk8s_solutions.py
+++ b/salt/_modules/metalk8s_solutions.py
@@ -315,6 +315,13 @@ def manifest_from_iso(path):
             )
         )
 
+    if not result['stdout']:
+        raise CommandExecutionError(
+            "Solution ISO at '{}' must contain a '{}' file".format(
+                path, SOLUTION_MANIFEST
+            )
+        )
+
     try:
         manifest = yaml.safe_load(result['stdout'])
     except yaml.YAMLError as exc:


salt/_modules/metalk8s_solutions.py
```diff
@@ -324,6 +324,11 @@ def manifest_from_iso(path):
             )
         )
 
+    if manifest is None:
```
gdemonet: Does it mean `result['stdout']` is an empty string? If that's the case, can't we tune the `isoinfo` command to make sure it breaks on a missing file?

salt/_modules/metalk8s_solutions.py
```diff
@@ -324,6 +324,11 @@ def manifest_from_iso(path):
...
+        raise CommandExecutionError(
+            "No such {} file found in {}".format(SOLUTION_MANIFEST, path)
```
gdemonet: Error message could be more "imperative": `"Solution ISO at {} must contain a '{}' file".format(...)`

salt/_modules/metalk8s_solutions.py
alexandre-allard: Doesn't seem to be achievable, at least with the current version (1.1.11) we ship.

alexandre-allard: Or we could do it with an ugly thing like:
```
isoinfo -x 'MANIFEST.YAML;1' -f </path/to/the/solution.iso> | grep ''
```
grep will fail because there is nothing to read.
But, we consume an extra process and it makes the command cryptic for a non initiated.
And, at least, with the current approach, we have clear error messages if it's the run of isoinfo that fails (for an obscure reason) or if the file does not exist.

gdemonet: OK thanks for checking :) sorry I made you, the point was more that it's weird to say the file doesn't exist after having tried to parse its contents as YAML, but it's really minor

gdemonet (dismissed):

salt/_modules/metalk8s_solutions.py
alexandre-allard: Oh ok, I've moved the test before the YAML parsing.

gdemonet (approved):

Issue #2634, opened 2020-06-18
Storage Operator can use SaltAPI without RBAC in 2.5


gdemonet:
**Component**: salt

**What happened**:

When implementing #2625, we found that forgetting the permission for a newly added Salt execution function (namely, `metalk8s_volumes.device_name`) didn't prevent `storage-operator` from using it, which is a serious security hole.
This appeared in 2.5.1-dev, whereas the RBAC was properly enforced in 2.6.0-dev.

**What was expected**:

Without the relevant permissions, one shouldn't be able to use SaltAPI freely.
If there is a hole in our setup in 2.5 around RBAC with Salt(API), we need it fixed ASAP.

**Steps to reproduce**: TBD

**Resolution proposal** (optional): TBD (we need a reproducing example, and tests to ensure this behaviour is fixed)

gdemonet:
First observation: in `salt/_auth/kubernetes_rbac.py`, the `auth` method is not checking that the username provided matches the ServiceAccount for which the token was issued. That in itself is a security hole (cc @NicolasT)

NicolasT:
There's no difference in the eauth code between 2.5 and 2.6, so it's odd as to why it worked in 2.5 and not 2.6.

NicolasT:
Actually not sure there's something wrong here. When looking at `storage-operator/deploy/role.yaml`, the `storage-operator` `ClusterRole` allows create/delete/get/list/patch/update/watch on `node` objects. As such, it's deemed to be part of the Salt `node-admins` group (which requires `*` on `nodes` in K8s, see the eauth code).

This `node-admins`  group can un `.*` on `*` (see `salt/metalk8s/salt/master/files/master-99-metallk8s.conf.j2`).

So everything works as designed (except that maybe the `storage-operator` should *not* have `*` rights on `node` objects, of course!)

NicolasT:
> First observation: in `salt/_auth/kubernetes_rbac.py`, the `auth` method is not checking that the username provided matches the ServiceAccount for which the token was issued. That in itself is a security hole (cc @NicolasT)

This is indeed a bug and needs to be fixed.

NicolasT:
Seems introduced in 56eda6d12c02082a14a864b97266ec9623011d6c @slaperche-scality. Giving the `storage-operator` role access to `nodes` gives it `node-admins` permissions in Salt.

NicolasT:
[...]

PR #2635, opened 2020-06-18
Tighten storage-operator permissions against Salt

gdemonet:
**Component**: salt, storage-operator

**Context**: see conversation in #2634

**Summary**: to do

**Acceptance criteria**: to do

---

Closes: #2634 
Fixes: #1528
Fixes: #2084

### DIFF
--- a/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
+++ b/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
@@ -45,7 +45,9 @@ external_auth:
     storage-operator:
       - '*':
         - 'disk.dump'
-        - 'state.sls'
+        - 'state.sls':
+            kwargs:
+              mods: 'metalk8s\.volumes.*'
       - '@jobs'
 
 # `kubeconfig` file and `context` used by salt to interact with apiserver
--- a/storage-operator/deploy/role.yaml
+++ b/storage-operator/deploy/role.yaml
@@ -4,29 +4,30 @@ metadata:
   creationTimestamp: null
   name: storage-operator
 rules:
+# For recording transition events
 - apiGroups:
   - ""
   resources:
-  - pods
-  - nodes
-  - services
-  - endpoints
-  - persistentvolumes
-  - persistentvolumeclaims
   - events
-  - configmaps
-  - secrets
   verbs:
-  - '*'
+  - create
+  - patch
+# For setting up monitoring for itself
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - get
+  - create
+  - update
 - apiGroups:
   - apps
   resources:
   - deployments
-  - daemonsets
   - replicasets
-  - statefulsets
   verbs:
-  - '*'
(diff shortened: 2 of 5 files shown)


storage-operator/deploy/role.yaml
```diff
@@ -4,36 +4,23 @@ metadata:
   creationTimestamp: null
   name: storage-operator
 rules:
+# For recording transition events
```
slaperche-scality, line 7: Good idea to add comments in this file :+1:

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
```diff
@@ -45,7 +45,9 @@ external_auth:
     storage-operator:
       - '*':
         - 'disk.dump'
-        - 'state.sls'
+        - 'state.sls':
```
slaperche-scality, line 48: Ho, this was simpler than I thought ^^

slaperche-scality (dismissed):
LGTM

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
gdemonet: That's the theory, not working atm :sweat_smile:

gdemonet: OK it looks like it's working actually!

slaperche-scality (dismissed):

storage-operator/deploy/role.yaml
```diff
@@ -4,36 +4,63 @@ metadata:
...
 - apiGroups:
   - apps
   resources:
   - deployments
-  - daemonsets
   - replicasets
-  - statefulsets
   verbs:
-  - '*'
+  - get
```
NicolasT, line 30: Why does the storage-operator need to read deployments and replicasets?

storage-operator/deploy/role.yaml
NicolasT: FWIW, if we update `operator-sdk`/`controller-runtime` (or move to `kubebuilder`), the RBAC rules are to be defined in the Go code using annotations, and this file is to be generated by the tooling.

storage-operator/deploy/role.yaml
gdemonet: It comes from `kubemetrics`: when creating the Service, it reads its Pod's owner controller (a ReplicaSet) and this controller's controller (a Deployment) to determine the right `selector` to use, I think. The information is disseminated in several issues and pull requests in `operator-sdk`, and I found most of them when investigating the `Failed to list *unstructured.Unstructured` error logs. Could do some digging up my history if you want, but I can confirm that there were unauthorized errors if I remove those permissions.

slaperche-scality (approved):

storage-operator/deploy/role.yaml
gdemonet: Oh yeah... Debt ticket #2644

PR #2642, opened 2020-06-22
salt-api: Prevent impersonation of ServiceAccounts using Bearer tokens

### LINKED ISSUE #2634: Storage Operator can use SaltAPI without RBAC in 2.5

gdemonet:
**Note for reviewers**:

This PR adds some changes to Basic auth handling, and does so in different manners:

- in 2.4, Basic auth support in SaltAPI is maintained (through the `password` value for `POST /login`), but now uses plaintext password instead of the `user:password` base64-encoded token
- in 2.5 and upwards, Basic auth support in SaltAPI is entirely removed, and logic for having multiple handlers based on authentication method is removed - as such, we also remove this particular logic from our clients, namely `storage-operator` and `metalk8s-ui`

<details>
  <summary>Commits</summary>

### salt-api: Rework Basic auth handling

Instead of relying on the strange assumption that the /login 'password'
argument is actually the full 'user:password' base64-encoded token, we
adjust the storage-operator and metalk8s-ui to pass in this password in
cleartext.

This commit also updates the `_auth_basic` handler to rely on the same
SelfSubjectAccessReview API as the  `_groups_basic` handler did.

---

### salt-api: Prevent SA impersonation

When using ServiceAccount, we were not checking if the provided Bearer
`token` actually belonged to the claimed `username`. This is now done,
and we add tests to prevent regressions.

NOTE: storage-operator username, from SaltAPI PoV, is now
'system:serviceaccount:kube-system:storage-operator'.

Fixes: #2634

---

### tests: Add positive check for auth success

We only assessed if an authentication attempt failed, while tried to
check the actual permissions for an authn assumed successful.
To make errors more explicit, we add a `then` check that the
"authentication succeeds".

See: #2634

</details>

### DIFF
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -84,6 +84,11 @@ def _review_access(kubeconfig, resource, verb):
     client = kubernetes.client.ApiClient(configuration=kubeconfig)
     authz_api = kubernetes.client.AuthorizationV1Api(api_client=client)
 
+    # NOTE: any authenticated user can use this API.
+    # This comes from the fact that an authenticated user will always belong to
+    # the `system:authenticated` group, and this group is bound to the
+    # `system:basic-user` ClusterRole, which enables creating
+    # SelfSubjectAccessReviews and SelfSubjectRulesReviews.
     return authz_api.create_self_subject_access_review(
         body=kubernetes.client.V1SelfSubjectAccessReview(
             spec=kubernetes.client.V1SelfSubjectAccessReviewSpec(
--- a/salt/_auth/kubernetes_rbac.py
+++ b/salt/_auth/kubernetes_rbac.py
@@ -1,4 +1,5 @@
 import base64
+from functools import wraps
 import logging
 
 MISSING_DEPS = []
@@ -6,6 +7,7 @@ MISSING_DEPS = []
 try:
     import kubernetes.client
     import kubernetes.config
+    from kubernetes.client.rest import ApiException
 except ImportError:
     MISSING_DEPS.append('kubernetes')
 
@@ -58,23 +60,32 @@ def _check_k8s_creds(kubeconfig, token):
         raise
 
 
-@_log_exceptions
-def _auth_basic(kubeconfig, username, token):
-    decoded = base64.decodestring(token)
-    if ':' not in decoded:
-        log.warning('Invalid Basic token format: missing ":"')
-        return False
+def _check_auth_args(f):
+    @wraps(f)
+    def wrapped(username, password=None, token=None, **kwargs):
+        error = None
 
-    (token_username, _) = decoded.split(':', 1)
-    if token_username != username:
-        log.warning('Invalid Basic token: username mismatch')
-        return False
+        if token and password:
+            error = 'cannot provide both "token" and "password"'
+        if not any([token, password and username]):
+            error = (
+                'must provide either a "token" or both a "username" and '
+                'a "password"'
+            )
 
-    return _check_k8s_creds(kubeconfig, 'Basic {}'.format(token))
+        if error is not None:
+            log.error('Invalid authentication request: %s', error)
+            raise CommandExecutionError(
+                'Invalid authentication request: {}'.format(error)
+            )
+
+        return f(username, password=password, token=token, **kwargs)
+
+    return wrapped
 
 
-@_log_exceptions
-def _groups_basic(kubeconfig, username, token):
+def _patch_kubeconfig_for_basic(kubeconfig, username, password):
+    token = base64.encodestring(':'.join([username, password])).rstrip('\n')
     kubeconfig.api_key = {
         'authorization': token,
     }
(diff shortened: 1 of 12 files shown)


gdemonet:
_Need to fix the storage-operator unit tests_

tests/post/steps/test_salt_api.py
```diff
@@ -146,6 +146,12 @@ def run_state_on_targets(host, context, negated, module, targets):
 def authentication_fails(host, context):
     assert context['salt-api']['login-status-code'] == 401
 
+
+@then('authentication succeeds')
+def authentication_fails(host, context):
```
alexandre-allard: bad copy/paste ?
```suggestion
def authentication_succeeds(host, context):
```

gdemonet: whoops

salt/_auth/kubernetes_rbac.py
```diff
@@ -41,68 +43,101 @@ def wrapped(*args, **kwargs):
...
         ),
     )
 
+
+@_log_exceptions
+def _auth_basic(kubeconfig, username, password):
+    _patch_kubeconfig_for_basic(kubeconfig, username, password)
+
+    try:
+        access_review = _review_access(
```
NicolasT, line 109: Can't a user be valid, but not have access to the `SelfSubjectAccessReview` API?

salt/_auth/kubernetes_rbac.py
```diff
@@ -116,7 +151,34 @@ def _groups_basic(kubeconfig, username, token):
...
+            log.error(
+                "Provided token belongs to '%s', does not match '%s'",
+                token_review.status.user.username,
+                username,
+            )
+            return False
+        else:
+            return True
+    else:
+        log.error('Provided token failed to authenticate')
```
NicolasT: Can we log the provided `username` here?

salt/_auth/kubernetes_rbac.py
```diff
@@ -148,23 +210,13 @@ def _load_kubeconfig(opts):
     return kubeconfig
 
 
+@_check_auth_args()
```
NicolasT: When calling with an incorrect combination of arguments, shouldn't this `raise` instead of return `False` as it does now?

salt/_auth/kubernetes_rbac.py
```diff
@@ -184,22 +236,10 @@ def auth(username, password=None, token=None, **kwargs):
     return result
 
 
+@_check_auth_args(on_fail=lambda: [])
```
NicolasT: When calling with an incorrect combination of arguments, shouldn't this `raise` instead of return `[]` as it does now?

tests/post/steps/test_salt_api.py
```diff
@@ -54,30 +65,49 @@ def context():
...
+        name=service_account.secrets[0].name, namespace=namespace
     )
     token = base64.decodebytes(secret.data['token'].encode('utf-8'))
     context['salt-api'] = _salt_api_login(
-        address, username=account_name, token=token
+        address, username=username, token=token
     )
 
 
+@when(parsers.parse(
```
NicolasT, line 84: There's a fair bit of duplication between this procedure and the one above. Could be split out?

ui/src/services/salt/api.js
```diff
@@ -14,7 +14,7 @@ export function authenticate(user) {
   return saltApiClient.post('/login', {
     eauth: 'kubernetes_rbac',
     username: user.username,
-    password: user.token,
```
NicolasT, line 17: So where did `user.token` go? Should it still be set? Does it have a different meaning now?

tests/post/steps/test_salt_api.py
gdemonet: Yes,  I'll cherry-pick the `_login_salt_api_sa` from 9d7e63d7d7f3e7601c0af02d2b1a61ae5ab73c82 (which I already customized in the 2.5 branch)

ui/src/services/salt/api.js
gdemonet: In 2.4, UI relies on basic auth. I didn't check the `user` state management in details, but both `password` and `token` are set. The API changed however, `password` isn't expecting a token anymore, but the password itself.

salt/_auth/kubernetes_rbac.py
gdemonet: Makes sense, and would make the code simpler :) I blindly followed the previous approach

salt/_auth/kubernetes_rbac.py
gdemonet: IIUC, any authenticated user will have the `system:authenticated` group. By default, this group is bound to the `system:basic-user` ClusterRole, which enables these permissions:

```yaml
rules:
- apiGroups:
  - authorization.k8s.io
  resources:
  - selfsubjectaccessreviews
  - selfsubjectrulesreviews
  verbs:
  - create
```

gdemonet: Should I add this as a comment inline?

tests/post/steps/test_salt_api.py
gdemonet: Done (not a real cherry-pick, but backported the same logic/name)

salt/_auth/kubernetes_rbac.py
gdemonet: Done

salt/_auth/kubernetes_rbac.py
gdemonet: Done

salt/_auth/kubernetes_rbac.py
gdemonet: Done

salt/_auth/kubernetes_rbac.py
NicolasT: Yes please.

ui/src/services/salt/api.js
NicolasT: Yes, but the `token` attribute on this `user` object now still exists. For no good reason?

gdemonet: Indeed, I don't think it's used. OK will remove it, and get some more conflicts, yay :trollface:!

salt/_auth/kubernetes_rbac.py
gdemonet: Will do!

ui/src/services/salt/api.js
NicolasT: Got `git rerere` enabled, right?

salt/_auth/kubernetes_rbac.py
gdemonet: Done: 54aca8a1be857f6547f93f610f145af5e645d2d9

ui/src/services/salt/api.js
gdemonet: Done: 94251f5

gdemonet: > Got `git rerere` enabled, right?

Yes of course, but now I'm changing things that I didn't before :)

NicolasT (approved):

Issue #2421, opened 2020-04-17
Support raw block volume


slaperche-scality:
**Component**:

operator, salt

**Why this is needed**:

There are some specialized applications (e.g.: database, SDS system, …) that require direct access to a block device because, for example, the file system layer introduces unneeded overhead.

To address this use case, we should be able to manage Volume that will backed by `PersistentVolume` with `volumeMode` set to `Block`.

**What should be done**:

Currently, the `volumeMode` is hardcoded to `Filesystem` in the operator, this need to change.
In the same vein, the formatting step is always executed by the Salt state, this also need to be changed.

**Implementation proposal**:

We need a way to know if we want to format the volume or not.
This could be either at the `Volume` level or at the `StorageClass` level.

I suggest that we control that at the `StorageClass` level, because:
- that's already the place where we configure the formatting, so it's consistent
- if we put it at the `Volume` level, we will have to sanity check against illogical combination (a `StorageClass` with `ext4` on a non-formatted `Volume`)
- by setting it as the `StorageClass` level it could apply to both `SparseLoopDevice` and `RawBlockDevice`.

The idea is to accept a special value "none" for the parameter `fsType` in the `StorageClass`.
When `fsType` is set to "none" then:
- `PersistentVolume` are created with `volumeMode` set to `Block`
- Salt state skip the formatting.

We could have made the field optional instead of using a special value, but I think that it's better to be explicit here.

**Test plan**:

If we can add support for `SparseLoopDevice` (may be tricky because we can't store the ID in the FS in that case), then we can add tests with `SparseLoopDevice`: create a volume and check that the PV has the right mode and the backing storage is not formatted.

Otherwise manuel testing I guess (since CI doesn't attach "real" volume to the VM).

slaperche-scality:
Comments are welcome! @NicolasT  @gdemonet @benoit-a

Or I can make a design section in the doc, like what I've done for volume v1 (will be volume v1.1 this time :stuck_out_tongue: )

gdemonet:
> Comments are welcome! @NicolasT @gdemonet @benoit-a

I don't have any strong opinion - the implementation proposed looks acceptable to me, especially since it's a feature supported by `PersistentVolume`s in the first place.

> Or I can make a design section in the doc, like what I've done for volume v1 (will be volume v1.1 this time  )

[...]

PR #2651, opened 2020-06-29
Add support for raw (non-formatted) volume

slaperche-scality:
**Component**:

operator, salt

**Context**: 

We want to support the provisioning of non-formatted block device.

**Summary**:

Add a new parameter `noFormat` in the Volume CRD and update Salt code to adapt the provisioning accordingly.

**Acceptance criteria**: 

Tests are passing (`sparseLoopDevice`)
Manual testing (`rawBlockDevice`)

---

Closes: #2421, #2963


slaperche-scality:
Will do some more manual testing, but code is ready for review

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -48,4 +48,11 @@ Provision backing storage for {{ volume }}:
     - require:
       - metalk8s_volumes: Format backing storage for {{ volume }}
 
+Update pillar after volume provisionning:
+  module.run:
+    - saltutil.refresh_pillar:
+      - wait: True
```
gdemonet, line 55: To make this call (and the above) synchronous, you'll need to explicitly pass `async=False`.
For reference, see:
https://github.com/saltstack/salt/issues/20590
https://github.com/saltstack/salt/pull/50168

gdemonet: Which makes me think there are some workarounds we could remove now that we upgraded to Salt 3000 (cc @TeddyAndrieux)

docs/developer/architecture/volume.rst
```diff
@@ -148,6 +148,28 @@ Similarly, our **Volume** object will have the following states:
...
+  setting the label of the partition to the **Volume** UUID. We can then use
+  ``/dev/disk/by-partlabel/<volume-uuid>`` as device path.
+* **rawBlockDevice** without filesystem:
+
+  * the **rawBlockDevice** is a disk (e.g. ``/dev/sdb``): we use the same
+    strategy as above.
+  * the **rawBlockDevice** is a partition (e.g. ``/dev/sdb1``): we re-label the
+    partition using the **Volume** UUID and use
+    ``/dev/disk/by-partlabel/<volume-uuid>`` as device path.
+  * The **rawBlockDevice** is a LVM volume: we use the exitisting LVM UUID and
```
gdemonet:
```suggestion
  * The **rawBlockDevice** is a LVM volume: we use the existing LVM UUID and
```

salt/_modules/metalk8s_volumes.py
```diff
@@ -451,6 +451,15 @@ def __init__(self, volume):
             self._kind = DeviceType.PARTITION
         self._kind = DeviceType.DISK
 
+    @property
+    def persistent_path(self):
+        if self._kind == DeviceType.LVM:
+            for symlink in glob.glob('/dev/disk/by-id/dm-uuid-LVM-*'):
+                realpath = os.path.realpath(symlink)
+                if os.path.basename(realpath) == name:
```
gdemonet: IIUC, `name` should be computed (maybe make this an attribute or a dynamic property?):

```suggestion
                if os.path.basename(realpath) == device_name(self.path):
```

gdemonet (commented):
Some early comments, most of it looks good to me :)

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: Ho good catch, will add `async=False` then.

docs/developer/architecture/volume.rst
slaperche-scality: Done

salt/_modules/metalk8s_volumes.py
slaperche-scality: Indeed, `name` is not defined >_<

Fixed.

salt/metalk8s/volumes/prepared/init.sls
slaperche-scality: Will revert this for now, seems not supported (or I got the syntax wrong):

```
      message: 'Salt job ''PrepareVolume'' failed with: ''saltutil.refresh_pillar''
        failed: refresh_pillar() got an unexpected keyword argument ''async'''
```

gdemonet: No indeed my bad, it seems the `wait: True` is the most recent implementation.

docs/operation/volume_management/volume_creation_deletion_cli.rst
```diff
@@ -26,13 +26,16 @@ This section describes how to create a **Volume** from the **CLI**.
...
            devicePath: <device_path>
+           noFormat: false
 
    Set the following fields:
 
    - **name**: the name of your volume, must be unique
    - **nodeName**: the name of the node where the volume will be located.
    - **storageClassName**: the **StorageClass** to use
    - **devicePath**: path to the block device (for example, `/dev/sda1`).
+   - **noFormat**: skip the volume formatting if **true**, optional (default to
```
NicolasT: Should `noFormat` be inside the `rawBlockDevice` section? Why not top-level?

Also, instead of `noFormat`, could we re-use the `volumeMode` as used `PersistentVolume(Claim)`?

salt/_modules/metalk8s_volumes.py
```diff
@@ -180,6 +182,32 @@ def device_name(path):
     return os.path.basename(realpath)
 
 
+def device_info(name):
+    """Return information about the backing storage device of the given volume.
+
+    Returned information are:
+    - an reliable path (e.g.: constant accross reboot) to the storage device
```
NicolasT:
```suggestion
    - a stable path (e.g.: constant across reboots) to the storage device
```

salt/_modules/metalk8s_volumes.py
```diff
@@ -339,6 +378,27 @@ def clean_up(self):
...
+    @property
+    def persistent_path(self):
+        return '/dev/disk/by-partlabel/{}'.format(self.get('metadata.uid'))
+
+    @property
+    def is_prepared(self):
+        """Check if the volume is already prepared."""
+        # Partition 1 should be labelled with the volume UUID.
+        pattern = '1.+{}'.format(self.get('metadata.uid').lower())
+        result  = _run_cmd(' '.join(['parted', self.path, 'print']))
```
NicolasT: This requires `parted` to be installed. Can we build this on top of plain `sfdisk` instead?

salt/_modules/metalk8s_volumes.py
```diff
@@ -377,6 +437,79 @@ def is_cleaned_up(self):
...
+        pattern = '{}.+{}'.format(partition, self.get('metadata.uid').lower())
+        result  = _run_cmd(' '.join(['parted', self.device_path, 'print']))
+        return re.search(pattern, result['stdout']) is not None
+
+    def prepare(self, force=False):
+        # Nothing to do in LVM case.
+        if self._kind == DeviceType.LVM:
+            return
+        uuid = self.get('metadata.uid')
+        # In case of partition, set the label to volume UUID.
```
NicolasT: Setting the *label* to the volume UUID? Can't we change the partition UUID instead?

NicolasT (dismissed):
Some questions and suggestions.

I think we should target `sfdisk` instead of `parted` to 'get things done', if feasible...

salt/_modules/metalk8s_volumes.py
slaperche-scality: From `man 8 sfdisk`:

> sfdisk doesn't understand the GUID Partition Table (GPT) format and it is not designed for large partitions.  In these cases use the more advanced GNU parted(8).

So I guess `parted` is a safer choice (will need to install it through Salt to be sure it's present though).

salt/_modules/metalk8s_volumes.py
slaperche-scality: Yeah, I wanted to use partition UUID at first, but couldn't find a way to set it :confused:  so I went with partition label instead.

salt/_modules/metalk8s_volumes.py
NicolasT: Modern `fdisk` does know GPT, but maybe not on CentOS 7. Anyway, on those systems there should be `sgdisk` for `fdisk`-like GPT management.

salt/_modules/metalk8s_volumes.py
NicolasT: Not fond of that at all, because super confusing. Instead, I'd set the *name* of the GPT partition to the `metadata.name` of the `Volume` object...

salt/_modules/metalk8s_volumes.py

docs/operation/volume_management/volume_creation_deletion_cli.rst
slaperche-scality: Yeah, good point.

I'll add a commit to change that.

salt/_modules/metalk8s_volumes.py
slaperche-scality:
```
[root@test-bootstrap centos]# sgdisk
bash: sgdisk: command not found
```

Seems not installed by default, so what's the advantage of using it instead of `parted` (since I'll have to make sure it's installed through Salt anyway)?

salt/_modules/metalk8s_volumes.py
slaperche-scality: I used UUID for consistency with the others, but I can change it to `name`.

salt/_modules/metalk8s_volumes.py
NicolasT: Fair enough. I was thinking `parted` has (many) more deps than `gdisk`, but that seems not to be the case (they're comparable), so if `parted` does what you need, go for it.

salt/_modules/metalk8s_volumes.py
NicolasT: I'm not sure we're aligned here :)

In GPT, a partition can have a *name* and a *GUID*. So, what I suggest is to set its *GUID* to the `metadata.uid` of the `Volume` object, and the *name* to `metadata.name` of the `Volume`. Then, always use the `/dev/disk/by-partuuid/...` path to *refer* to it. The *name* is merely cosmetic.

slaperche-scality: Yup, but:

> I wanted to use partition UUID at first, but couldn't find a way to set it :confused: 

Haven't find a way to set the partition UID with parted.
Will give a try to `sgdisk`, maybe it has more options.

slaperche-scality: Easy to do with `sgdisk`, I've added a new commit to use `sgdisk` instead of `parted`.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Done, I'm now using `sgdisk` since it's possible to set the partition GUID with it.

slaperche-scality:
Ready for a new round of review, I've applied all the requested changes and tested manually the different cases for Block mode (whole disk, partition and LVM volume) and they work.

That being said, unlike Filesystem mode, we now some distinct codepaths for Block mode (partition and LVM cases are specific to `rawBlockDevice`): at one point, we may want to have tests in the CI for `rawBlockDevice` as well.

docs/developer/architecture/volume.rst
```diff
@@ -157,6 +148,28 @@ Similarly, our **Volume** object will have the following states:
...
+In order to have a reliable automount through kubelet, we need to create the
+underlying **PersistentVolume** using a persistent name for the backing storage
+device. We use different strategies according to the **Volume** type:
+
+* **sparseLoopDevice** and **rawBlockDevice** with a filesystem: during the
+  formatting, we set the filesystem UUID to the **Volume** UUID and use
+  ``dev/disk/by-uuid/<volume-uuid>`` as device path.
+* **sparseLoopDevice** without filesystem: we create a GUID Partition Table on
+  the sparse file and create a single partition encompassing the whole device,
+  setting the GUID of the partition to the **Volume** UUID. We can then use
```
alexandre-allard-scality, line 162: Should be `the UUID of the partition`, no ? Or maybe I'm lost with all these UUID/GUID. 
```suggestion
  setting the UUID of the partition to the **Volume** UUID. We can then use
```

salt/_modules/metalk8s_volumes.py
```diff
@@ -377,6 +442,76 @@ def is_cleaned_up(self):
...
+        super(RawBlockDeviceBlock, self).__init__(volume)
+        # Detect which kind of device we have: a real disk, only a partition or
+        # an LVM volume.
+        name = device_name(self.path)
+        self._partition = None
+        for symlink in glob.glob('/dev/disk/by-id/dm-uuid-LVM-*'):
+            realpath = os.path.realpath(symlink)
+            if os.path.basename(realpath) == name:
+                self._kind = DeviceType.LVM
+                return
```
alexandre-allard: Minor, it looks like the code to find a LVM symlink here and in `persistent_path()` could be factorized in a function:
```
def get_lvm_path(self):
    name = device_name(self.path)
    for symlink in glob.glob('/dev/disk/by-id/dm-uuid-LVM-*'):
        realpath = os.path.realpath(symlink)
        if os.path.basename(realpath) == name:
            return symlink
    return None
```
```suggestion
        self._partition = None
        if self.get_lvm_path():
            self._kind = DeviceType.LVM
            return
```

alexandre-allard (dismissed):
LGTM.

salt/_modules/metalk8s_volumes.py
slaperche-scality: Good point, I've added a commit to factorize it!

alexandre-allard (approved):

Issue #2266, opened 2020-02-26
Add salt module unit test


TeddyAndrieux:
**Component**:

'tests', 'salt'

**Why this is needed**:

Currently we didn't do any test for salt custom module and states

**What should be done**:

Add unit test for salt module (when it's possible).
This test should run in pre-merge like all other unit tests

**Implementation proposal** (strongly recommended):

https://docs.saltstack.com/en/master/topics/development/tests/unit.html

alexandre-allard:
Closing this one as we now have 100% coverage on Salt execution modules.

PR #2654, opened 2020-07-02
Add unit tests for `containerd` and `cri` Salt module

TeddyAndrieux:
**Component**:

'salt,' 'tests'

**Context**: 

#2266 

**Summary**:

- Add tests utils to simulate  salt `cmd.run_all` output
- Small fix in `cri.execute`
- Add unit tests for `containerd` and `cri` custom Salt module

---

Refs: #2266

### DIFF
--- a/salt/tests/unit/modules/test_cri.py
+++ b/salt/tests/unit/modules/test_cri.py
@@ -1,0 +1,194 @@
+import json
+
+from parameterized import parameterized
+
+from salttesting.mixins import LoaderModuleMockMixin
+from salttesting.unit import TestCase
+from salttesting.mock import MagicMock, patch
+
+from tests.unit import utils
+
+import cri
+
+
+IMAGES_LIST = [{
+    "id": "sha256:da86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e",
+    "repoTags": [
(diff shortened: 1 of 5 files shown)


salt/tests/unit/modules/test_cri.py
```diff
@@ -0,0 +1,194 @@
...
+from salttesting.mixins import LoaderModuleMockMixin
+from salttesting.unit import TestCase
+from salttesting.mock import MagicMock, patch
+
+from tests.unit import utils
+
+import cri
+
+
+IMAGES_LIST = [{
```
Ebaneck, line 14: Since we are starting to have variables of 21+ lines, what is the good practice here...
 - To define a global helper file that contains these variables and simply import them from there or
 - Move the variable declaration to the bottom of this file

TeddyAndrieux: To me I think it's good to have example of output you get from command in test so that it's easier to understand the code, anyway to me here it's still simple ones so it's ok to have them directly in the python test file IMHO.

For complicated ones I move to file check https://github.com/scality/metalk8s/pull/2656

Ebaneck (commented):
Seems ok for me...
cc @alexandre-allard-scality

Ebaneck (approved):

PR #2655, opened 2020-07-06
salt/tests: Add unit tests for `metalk8s_etcd` salt module

TeddyAndrieux:
**Component**:

'salt,' 'tests'

**Context**: 

#2266 

**Summary**:

Add unit tests for `metalk8s_etcd` salt custom module, install etcd3 in
pytest environment and mock all etcd3 call needed

---

Refs: #2266


salt/tests/unit/modules/test_metalk8s_etcd.py
```diff
@@ -0,0 +1,258 @@
...
+        "https://10.11.12.14:2379"
+    ],
+    "id": "17971792102091431978L",
+    "name": "node1",
+    "peer_urls": [
+        "https://10.11.12.14:2380",
+        "https://11.11.12.14:2380"
+    ]
+}]
+MEMBERS_LIST = [MagicMock(**(member)) for member in MEMBERS_LIST_DICT]
```
alexandre-allard: details, superfluous parenthesis:
```suggestion
MEMBERS_LIST = [MagicMock(**member) for member in MEMBERS_LIST_DICT]
```

alexandre-allard (dismissed):
LGTM, but I still don't get the point of testing functions with all the logic inside them mocked (e.g. add_etcd_node). To me, it looks like we're testing nothing in these cases (maybe if these functions change at some point, it will be useful).

salt/tests/unit/modules/test_metalk8s_etcd.py
TeddyAndrieux: Right i will change it

TeddyAndrieux:
> LGTM, but I still don't get the point of testing functions with all the logic inside them mocked (e.g. add_etcd_node). To me, it looks like we're testing nothing in these cases (maybe if these functions change at some point, it will be useful).

Yes I agree, btw for `add_etcd_node` we check at least that we do one call to `add_member` function with the expected url

alexandre-allard (approved):

PR #2661, opened 2020-07-08
salt/tests: Add unit tests for `metalk8s` salt module

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

#2266 

**Summary**:

- Salt: Fix `if` block in `metalk8s` Salt module when we have multiple metalk8s archive with the same version
- Tests: Unit tests for `metalk8s` Salt module

---

Refs: #2266

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -227,12 +227,13 @@ def get_archives(archives=None):
         env_name = 'metalk8s-{0}'.format(version)
 
         # Warn if we have 2 archives with the same version
-        if env_name in archives:
+        if env_name in res:
             archive = res[env_name]
             log.warning(
-                'Skip, archive %s has the same version as %s: %s.',
-                archive, archive['iso'] or archive['path'], version
+                'Archives have the same version: %s is overridden by %s.',
+                archive, info
             )
+
         res.update({env_name: info})
     return res
 
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -52,7 +52,7 @@ def format_san(names):
         for af_name in ['AF_INET', 'AF_INET6']:
             try:
                 af = getattr(socket, af_name)
-            except AttributeError:
+            except AttributeError:  # pragma: no cover
                 log.info('Unkown address family: %s', af_name)
                 continue
 
--- a/salt/tests/unit/modules/test_metalk8s.py
+++ b/salt/tests/unit/modules/test_metalk8s.py
@@ -1,0 +1,276 @@
+import os.path
+import yaml
+
+from parameterized import param, parameterized
+
+from salt.exceptions import CommandExecutionError
+
+from salttesting.mixins import LoaderModuleMockMixin
+from salttesting.unit import TestCase
+from salttesting.mock import MagicMock, mock_open, patch
+
+from tests.unit import utils
+
+import metalk8s
+
+
+YAML_TESTS_FILE = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)),
+    "files", "test_metalk8s.yaml"
+)
+
+PRODUCT_TXT = '''
+NAME=MetalK8s
+VERSION=2.5.0
+SHORT_VERSION=2.5
+GIT=2.5.0-0-g47705aa
+DEVELOPMENT_RELEASE=0
+BUILD_TIMESTAMP=2020-04-18T01:46:10Z
+BUILD_HOST=kw029-prod-metalk8s-backend-0-22391-14681
+'''
+
+
+class Metalk8sTestCase(TestCase, LoaderModuleMockMixin):
+    """
+    TestCase for `metalk8s` module
+    """
+    loader_module = metalk8s
+    loader_module_globals = {
+        "__grains__": {
+            "id": "my_node_1"
+        }
+    }
+
+    def test_virtual(self):
+        """
+        Tests the return of `__virtual__` function
+        """
+        self.assertEqual(metalk8s.__virtual__(), 'metalk8s')
+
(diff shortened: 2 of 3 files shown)


salt/_modules/metalk8s.py
```diff
@@ -227,12 +227,13 @@ def get_archives(archives=None):
         env_name = 'metalk8s-{0}'.format(version)
 
         # Warn if we have 2 archives with the same version
-        if env_name in archives:
+        if env_name in res:
             archive = res[env_name]
             log.warning(
-                'Skip, archive %s has the same version as %s: %s.',
-                archive, archive['iso'] or archive['path'], version
+                'Archives has the same version: %s is overrided by %s.',
```
alexandre-allard:
```suggestion
                'Archives have the same version: %s is overridden by %s.',
```

salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -0,0 +1,272 @@
...
+                    set(metalk8s.minions_by_role(role, nodes)),
+                    set(result)
+                )
+
+    @parameterized.expand([
+        (PRODUCT_TXT, {'version': '2.5.0', 'name': 'MetalK8s'}),
+        (None, 'Path /my/path/ has no "product.txt"', True),
+        ("Not a good product txt", {'version': None, 'name': None})
+    ])
+    def test_archive_info_from_tree(self, product, result, raises=False):
```
alexandre-allard, line 133: Can't we just use real files instead of mocking the `isfile()` and `fopen()` functions ?
i.e. we could have (I'm not saying we should do it this way):
```
./files/good/product.txt
./files/bad/product.txt
/non/existent/path/product.txt
```
It's not really important as we don't really want to tests these functions, it's just that this test function is more complicated than the code it tests. :/
In the end, here we're only testing 2 really simple regexps...
Or we could slightly change the `archive_info_from_tree()` function signature to be able to change the `product.txt` filename.
Maybe what you did is how we should test this, I don't really know what's good practices in tests.

salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -0,0 +1,272 @@
...
+                    result
+                )
+                open_file_mock.assert_called_once_with("/my/path/product.txt")
+                read_mock.assert_called_once()
+
+            is_file_mock.assert_called_once_with("/my/path/product.txt")
+
+    @parameterized.expand([
+        (PRODUCT_TXT, {'version': '2.5.0', 'name': 'MetalK8s'}),
+        (None, 'Failed to run isoinfo: Error occured', True),
```
alexandre-allard: `occurred` takes 2 R:
```suggestion
        (None, 'Failed to run isoinfo: An error has occurred', True),
```

salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -0,0 +1,272 @@
...
+    ])
+    def test_archive_info_from_iso(self, product, result, raises=False):
+        """
+        Tests the return of `archive_info_from_iso` function
+        """
+        iso_info_cmd = utils.cmd_output(
+            retcode=0 if product is not None else 1,
+            stderr='' if product is not None else 'Error occured',
+            stdout=product if product is not None else ''
+        )
```
alexandre-allard:
```suggestion
        iso_info_cmd = utils.cmd_output(
            retcode=1 if product is None else 0,
            stderr='Error occured' if product is None else '',
            stdout='' if product is None else product
```
or even:
```suggestion
        if product is None:
            iso_info_cmd_kwargs = {
                'retcode': 1,
                'stderr': 'An error has occurred',
                'stdout': '',
            }
        else:
            iso_info_cmd_kwargs = {
                'retcode': 0,
                'stderr': '',
                'stdout': product,
            }
        iso_info_cmd = utils.cmd_output(**iso_info_cmd_kwargs)
```
which is longer, but IMHO more readable.

alexandre-allard (commented):
Details, otherwise LGTM.

salt/tests/unit/modules/test_metalk8s.py
TeddyAndrieux: I don't know either what is the best practices in tests, for this one, we could have real txt file but for `archive_info_from_iso` you will need real ISO and I think it's not really good so here I do exactly the same tests for both function but indeed it could be files.

But I just saw `mock_open` in doc that will be much simpler than what we have right now, I will change it
https://cpython-test-docs.readthedocs.io/en/latest/library/unittest.mock.html#mock-open

salt/tests/unit/modules/test_metalk8s.py
TeddyAndrieux: Yes sure that why it's an error :sweat_smile: (right I will fix it)

salt/tests/unit/modules/test_metalk8s.py
alexandre-allard: Ok, my point was only for this function, not the ISO one (still we could build an ISO with the same `product.txt` files aforementioned as the first step in the test function..., but we would also need to ensure that we have the binaries to handle the ISO, so I'm not sure it's a good idea).
What you did seems OK to me, just wanted to ask, because as I said I'm not really aware of how things should be done.

alexandre-allard (approved):

salt/tests/unit/modules/test_metalk8s.py
gdemonet: FWIW, best practice in unit tests is to mock everything outside of the tested function, and to never use real IO (to avoid flakiness / concurrency issues). So it seems like a good approach to me.
It could make sense for us to add such rules in our developer guidelines ;)

PR #2664, opened 2020-07-09
salt/tests: Add unit tests for `metalk8s_package_manager_yum` salt module

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

#2266 

**Summary**:

Add unit tests for `metalk8s_package_manager_yum` salt custom module,
use a dedicated yaml file containing input, arguments and expected
result for some functions

NOTE: This PR does not test `metalk8s_package_manager_apt` as right now it does not work as expected and it's not used

---

Refs: #2266

### DIFF
--- a/salt/tests/unit/modules/files/test_metalk8s_package_manager_yum.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_package_manager_yum.yaml
@@ -1,0 +1,194 @@
+list_dependents:
+  # 3 dependents - all of them in allowed_versions
+  - repoquery_output: |
+      calico-cni-plugin 3.8.2-1.el7
+      kubelet 1.15.11-0
+      kubernetes-cni 0.7.5-0
+    allowed_versions:
+      kubelet: 1.15.11-0
+      kubernetes-cni: 0.7.5-0
+      calico-cni-plugin: 3.8.2-1.el7
+      kubectl: 1.15.11-0
+    fromrepo: "metalk8s-kubernetes"
+    result:
+      kubelet: 1.15.11-0
+      kubernetes-cni: 0.7.5-0
+      calico-cni-plugin: 3.8.2-1.el7
+
+  # 3 dependents - only one in allowed_versions
+  - repoquery_output: |
+      my_unknown_package 1.2.3
+      my_second_unknown_package 4.5.6
+      kubelet 1.15.11-0
+    allowed_versions:
+      kubelet: 1.15.11-0
+      kubernetes-cni: 0.7.5-0
+      calico-cni-plugin: 3.8.2-1.el7
+      kubectl: 1.15.11-0
+    result:
+      kubelet: 1.15.11-0
+
+  # 1 dependents - not in the allowed_versions
+  - repoquery_output: |
+      kubelet 1.15.10-0
+    allowed_versions:
+      kubelet: 1.15.11-0
+    result: {}
+
+  # 3 dependents - empty allowed_versions
+  - repoquery_output: |
+      calico-cni-plugin 3.8.2-1.el7
+      kubelet 1.15.11-0
+      kubernetes-cni 0.7.5-0
+    result: {}
+
+  # 0 dependents - empty allowed_versions
+  - repoquery_output: ""
+    result: {}
+
+  # 0 dependents - few allowed_versions
+  - repoquery_output: ""
+    allowed_versions:
+      kubelet: 1.15.11-0
+      kubernetes-cni: 0.7.5-0
+      calico-cni-plugin: 3.8.2-1.el7
+      kubectl: 1.15.11-0
+    result: {}
+
+  # Error when retrieving dependents
+  - repoquery_output: null
+    result: null
+
+list_pkg_dependents:
+  # No version specified
+  - result:
+      my_package: null
+
+  # version specified - no dependents
+  - version: "3.11.12"
+    result:
+      my_package: "3.11.12"
+
+  # version specified - no dependents - in pkgs_info
+  - version: "3.11.12"
+    pkgs_info:
+      my_package:
+        version: "3.11.12"
+    result:
+      my_package: "3.11.12"
+
+  # version specified - no dependents - version not in pkgs_info
+  - version: "3.11.13"
+    pkgs_info:
+      my_package:
+        version: "3.11.12"
+    result: null
+
+  # version specified - no dependents - package not in pkgs_info
+  - name: "my_unknown_package"
+    version: "3.11.12"
+    pkgs_info:
+      my_package:
+        version: "3.11.12"
+    result: null
+
+  # version specified - 2 dependents not installed
+  - version: "3.11.12"
+    list_dependents:
+      my_second_package: "1.11.12"
+      my_third_package: "2.14.15"
+    result:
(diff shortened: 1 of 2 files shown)


salt/tests/unit/modules/files/test_metalk8s_package_manager_yum.yaml
```diff
@@ -0,0 +1,147 @@
+list_dependents:
```
gdemonet, line 1: Do we really need to store test cases in an external file? Makes it harder to spot mistakes / missing cases when reading a test's code, IMHO.

Also, in general I'd recommend having a label to explicit what's happening (e.g. "all dependents found", "only one dependent found", "no dependent found", ...).

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
```diff
@@ -0,0 +1,196 @@
...
+        with ForceImportErrorOn("yum"):
+            reload(metalk8s_package_manager_yum)
+            self.assertEqual(
+                metalk8s_package_manager_yum.__virtual__(),
+                False
+            )
+
+    @parameterized.expand(
+        param.explicit(kwargs=test_case)
+        for test_case in yaml.safe_load(open(YAML_TESTS_FILE))["list_dependents"]
```
gdemonet: Instead of loading this file in the decorator, can't we load it only once in some `YAML_TEST_CASES` variable?

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
```diff
@@ -0,0 +1,196 @@
...
+                ),
+                result
+            )
+            repoquery_cmd_mock.assert_called_once()
+            self.assertEqual(
+                repoquery_cmd_mock.call_args[0][0][0],
+                "repoquery"
+            )
+            if "fromrepo" in kwargs:
+                self.assertTrue(
```
gdemonet: There's an `assertIn` method for this: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIn

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
```diff
@@ -0,0 +1,196 @@
...
+                    if isinstance(rpm_qa_outputs, dict):
+                        if rpm_qa_outputs.get(command[2]) is None:
+                            out_kwargs['retcode'] = 1
+                            out_kwargs['stderr'] = 'An error has occured'
+                        else:
+                            out_kwargs['stdout'] = rpm_qa_outputs[command[2]]
+                    else:
+                        out_kwargs['stdout'] = rpm_qa_outputs
+                return utils.cmd_output(**out_kwargs)
+            return None
```
gdemonet, line 127: I had a hard time understanding all the cases coming from different inputs to this test function. A test shouldn't have too much logic in it, so it's clear where the error comes from when the test breaks. Can't we split test cases, at least between nominal and error cases?

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            process_trans_mock.side_effect = Exception("An error has occurred")
+
+        with patch.dict("sys.modules", {"yum": MagicMock()}):
+            reload(metalk8s_package_manager_yum)
+            with patch("yum.YumBase", ybase_mock), \
+                    patch("yum.Errors.InstallError", Exception), \
+                    patch("yum.Errors.YumDownloadError", Exception), \
+                    patch("yum.Errors.YumRPMCheckError", Exception), \
+                    patch("logging.getLogger", MagicMock()):
```
gdemonet, line 187: (can be considered as debt) Maybe we should capture and check what/when we log from this module.

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
```diff
@@ -0,0 +1,196 @@
...
+            process_trans_mock.return_value = True
+        else:
+            process_trans_mock.side_effect = Exception("An error has occurred")
+
+        with patch.dict("sys.modules", {"yum": MagicMock()}):
+            reload(metalk8s_package_manager_yum)
+            with patch("yum.YumBase", ybase_mock), \
+                    patch("yum.Errors.InstallError", Exception), \
+                    patch("yum.Errors.YumDownloadError", Exception), \
+                    patch("yum.Errors.YumRPMCheckError", Exception), \
```
gdemonet, line 186: Why are we mocking those? Isn't the `MagicMock` used in `sys.modules` sufficient?

gdemonet (changes requested):
I think all meaningful cases were covered, but it's not easy to say with this big YAML file separated from the tests :confused:

IMHO, tests should be:
- simple (an `if/else` should be seen as a red flag)
- self-explanatory
- split by expected behaviour ("nominal", "error", "side effect", ...)

salt/tests/unit/modules/files/test_metalk8s_package_manager_yum.yaml
TeddyAndrieux: You just have to read the yaml file to see all the test cases. If we put all in the python it will be unreadable to me.
We could add labels for tests  it's true

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
TeddyAndrieux: This function is complicated? 
This function do not test anything about our product is just to simulate `rpm -qa` output from yaml file so that you can have different output depending on the package. It's just a helper to test

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
TeddyAndrieux: No, because we want to reach the `except` block so we need to catch Exceptions

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
TeddyAndrieux: For the moment we didn't log anything we just disable the yum logs, so let's not invest time on it right now

salt/tests/unit/modules/files/test_metalk8s_package_manager_yum.yaml
gdemonet: I get why we don't want to clutter code files with many repetitive datasets. Just not fond of having things requiring my brain to split attention between multiple files when trying to reason about a piece of logic.

Let's keep this approach for now, better have tests than debate for too long, we'll see how that works in the long run.

salt/tests/unit/modules/test_metalk8s_package_manager_yum.py
gdemonet: I'll try to explain my point later when implementing tests for drain. I understand how this helper is just here to build a parametrized mock of `rpm -qa`, but it took me some time (maybe because there's no docstring, maybe because the test case which help me understand what `rpm_qa_outputs` look like are in another file, maybe because I'm dumb...).

gdemonet (approved):

Issue #2508, opened 2020-04-30
Alerts List in Volume page 


ChengYanJin:
**Component**:
ui

**Why this is needed**:
We want to have the alerts list for the specific volume.
In the long term, the user can be redirected to this alert table by click on the alert related to this volume on the alert page.

**What should be done**:
Display the alerts of a specific volume.

**Implementation proposal** (strongly recommended):
Given #2502(Get the alerts for the specific resource) is done, we should be able to fetch the alerts related to the volume through the labels.

We have three default `critical` alerts for volume by default expose by Prometheus Operator. 
```
KubePersistentVolumeUsageCritical,critical,The PersistentVolume claimed by {{ $labels.persistentvolumeclaim }} in Namespace {{ $labels.namespace }} is only {{ $value | humanizePercentage }} free.

KubePersistentVolumeFullInFourDays,critical,"Based on recent sampling, the PersistentVolume claimed by {{ $labels.persistentvolumeclaim }} in Namespace {{ $labels.namespace }} is expected to fill up within four days. Currently {{ $value | humanizePercentage }} is available."

KubePersistentVolumeErrors,critical,The persistent volume {{ $labels.persistentvolume }} has status {{ $labels.phase }}. 
```
The 3rd alert should be triggered if the volume is `Failed or Pending` after 5mins

**Test plan**:
TBD

Issue #2507, opened 2020-04-30
Volume page - Performance charts


ChengYanJin:
**Component**:
ui

**Why this is needed**:
These charts provide us an intuitive way to check the performance of a specific volume.

**What should be done**:
We should get the data through Prometheus API and display it in the sparkline chart.

**Implementation proposal** (strongly recommended):
@thomasdanan will provide a more precise graph from Grafana. 

![image](https://user-images.githubusercontent.com/18453133/80733814-7f506680-8b0e-11ea-95aa-b7062012edd9.png)

Got inspiration from the Grafana dashboard. 
![Untitled](https://user-images.githubusercontent.com/18453133/84632942-f770c100-aeef-11ea-9cf4-40c1336c2368.png)

The difficulties facing in these performance charts:
- [vega-lite] How to add the label with the conversion from bytes to the units
- [prometheus] Need to figure out which metric we could use for the KPI`IOPS`.

We should properly handle the **downtime** of the machine
when there is no data we shouldn't display anything
![image](https://user-images.githubusercontent.com/18453133/86479045-aebc6300-bd4b-11ea-87f8-d4782c5aa7e3.png)

**Test plan**:
TBD

leeoniya:
may be of interest: https://github.com/leeoniya/uPlot

gdemonet:
Thanks for sharing your project @leeoniya :) Unfortunately, we're already building a simplified components library on top of Vega(Lite) - see [scality/core-ui](https://github.com/scality/core-ui) - and focus more on having a clear interoperable language for defining graphs than on performance (for this, we use Grafana :wink:, maybe they should use uPlot?).

leeoniya:
i'm told it's in the works ;)

https://github.com/grafana/grafana/pull/21835
https://news.ycombinator.com/item?id=23228327

Issue #2506, opened 2020-04-30
Fill the Volume left-side Page with volume table 


ChengYanJin:
**Component**: 
ui

**Why this is needed**:
For the 1st iteration, we want to display the volume list as a table instead of cards.

**What should be done**:
- Will work with @thomasdanan to define what is missing in the current volume table.
  - should be able to sort the volume by usage
- When we click on a specific volume, the right-side panel should update with the volume panel.
- When we click on the `Create Volume` button, the form appears.

**Implementation proposal** (strongly recommended):
![image](https://user-images.githubusercontent.com/18453133/80732501-b58ce680-8b0c-11ea-8dcb-1e45c9e82c0f.png)

**Test plan**:
Adapt the current cypress test

Issue #2505, opened 2020-04-30
Set up the style & layout for Volume page


ChengYanJin:
**Component**:
 ui

**Why this is needed**:
We would like to adapt the current volume page to the new design with the volume list on the left side and volume panel on the right side, including `volume info/health/alerts/performance`.

**What should be done**:

**Prerequisites: the design file from UX**:
Will work with @Cuervino to clarify the following questions in the design file:
- How the volume page looks like when landing the page
- Where should we display the `Create Volume Form` (either display it on the right side the panel or go for a new page)

Dev implementation:
- Add the link in the detail node page to access the volume page with the filter.
- Add the entry point in the side-bar to access the global volumes.
- We should still keep the tabs (Details/Volumes/Pods) on the node page. And we could access the `volume page` when I click on a certain volume in the volume list.
- Need to decide how to display the breadcrumb
- Remove the VolumeInformation page and move the volume detail data in the Volume Page. 

There are some additional fields, e.g. `Workload Bound` which is missing right now.
There is no direct way to retrieve the data.
What we could do is when the Volume/PV is bound => find the name of the PVC => list all the pods on this node => find which on has the PVC. Then the Pod should be the `Workload Bound`
Note that a Volume/PV can be bound by more than one node (not in our case I think, but we should still handle it)

**Implementation proposal** (strongly recommended):
![image](https://user-images.githubusercontent.com/18453133/82414464-6258e480-9a77-11ea-8271-580717722ade.png)

**Test plan**:
TBD

Cuervino:
About: 
> How the volume page looks like when landing the page

Here is a proposal:
![volumes view test](https://user-images.githubusercontent.com/28654802/80970701-e0807e80-8e1b-11ea-839d-c7316f1676a5.png)
On the right panel, just a notification "No data selected"
About:
> Where should we display the Create Volume Form (either display it on the right side the panel or go for a new page)

I would go for the right side panel, especially if the whole form is short enough, and if there is only 1 step.

slaperche-scality:
In term of wording, I would use `Available` instead of `Unbound`

gdemonet:
> In term of wording, I would use `Available` instead of `Unbound`

Related to this, I found myself trying to reuse an `Available` Volume after it had been previously consumed by another workload. This is because the PVC has been removed (so the Volume isn't bound), but since the `reclaimPolicy` was set to `Retain` in the StorageClass, we still have the underlying PV available (in case we want to extract the data manually).

[...]

PR #2665, opened 2020-07-13
[UI]The new volume page

ChengYanJin:
**Component** ui, volumes, monitoring

**Context**:  Replace the old volume information page with the new Volume Page

**Summary**:
- Volume list is implemented by `react-table`, 
① the selected row is highlighted so that we can clearly know which volume we are looking for)
② store the search bar content in `localStorage`, so we won't lose them during the refresh and API call.
③ add scroll bar 

- Volume Detail Card ( Note that I remove the Volume Health Status Bar since we can't get the history of the alerts yet)
- Volume Active Alert Card ( Note that I still get the alerts through `Prometheus` not `AlertManager` yet, the alert of AertManager has a weird behavior we need a further investigation on it, please see #2662)
- Metrics Chart Card 
① We can change the time span from the dropdown, by default is `last 7 days`
② rename it from PERFORMANCE to METRICS
③  add auto-refreshing( refresh the metrics graph every 1 min)

- Add a node field in Create Volume Page.
- Update the URL and Breadcrumb

URL for volume page:
`/volumes`: Access the volume page same as from the sidebar
`/volumes/<volume-name>`: Check the detail of volume <volume-name> 
`/volumes/?node=<node-name>`: Add the node filter, if we go to the volume page from node page
`/volumes/createVolume`: Create a volume from the `/volumes` page
`/volumes/createVolume/?node=<node-name>`: Create a volume from with node filter, the `node` field prefilled by node filter

We don't want to re-render the whole VolumePage when we select to another volume
How to design the components: 

- `VolumePage` component (smart)
     - get the list of volumes
- `VolumeContent` component (smart)
    - extract the volume from the URL
    - tell the sub-router component the selected volume
    - get the metric data from monitoring ( for the MetricChart)
- `VolumeListTable` component (dumb)
    - get the volumelist and selected volume through props 
- `RightSidePanel`component
   - `VolumeDetailCard` component (dumb)    
   - `ActiveAlertCard`  component (dumb)
   - `MetricGraphCard` component (dumb)

We design those three components as a dumb component is because we may want to add a tab to switch between those components. And we don't want them to be refreshed. 

**Acceptance criteria**: 
image
image
image

How to fill up the volume? take alertmanger volume as example which the size is `10GiB` the devicepath is `/dev/vdb`
1. ssh centos@10.200.5.215
2. cd dev/
3. mount /dev/vdb my-volume
4. dd if=/dev/urandom of=my-volume/bigfile bs=1M count=9500 seek=0

Note that this PR should be merged in 2.6

Closes: #2505 
Closes: #2506
Closes: #2507 
Closes: #2508


NicolasT:
I think `Performances` should be `Performance` singular. Or, rather, `Metrics`/`Statistics`, because `Usage` for example is not (necessarily) related to 'performance'.

ui/src/translations/en.json
```diff
@@ -136,5 +136,11 @@
   "import_solution_hint": "You should import a solution to prepare your environment",
   "redirecting": "Redirecting…",
   "documentation": "Documentation",
-  "salt_job": "Salt job"
+  "salt_job": "Salt job",
+  "platform": "Platform",
+  "node": "Node",
+  "used_by": "Used By",
+  "backend_disk": "Backend Disk",
+  "active_alerts": "ACTIVE ALERTES",
```
NicolasT:
```suggestion
  "active_alerts": "ACTIVE ALERTS",
```

ui/src/translations/en.json
```diff
@@ -73,7 +73,7 @@
   "create_new_volume": "Create a New Volume",
   "no_volume_found": "No volumes on this node",
   "no_searched_volume_found": "No volumes matching the search",
-  "type": "Type",
+  "volume_type": "Volume Type",
   "volume_size": "Volume Capacity",
   "device_path": "Device path",
```
NicolasT: There appears to be a casing difference in here, e.g. between `Volume Type` (`T`) and `Device path` (`p`). Is that intentional? Any 'rule'?

ui/src/components/BreadcrumbStyle.js
```diff
@@ -3,10 +3,10 @@ import { Link } from 'react-router-dom';
 import { fontSize, padding } from '@scality/core-ui/dist/style/theme';
 
 export const BreadcrumbContainer = styled.div`
-  margin-left: ${padding.small};
   .sc-breadcrumb {
-    padding: ${padding.smaller};
+    height: 24px;
```
MonPote: Is it normal the height is hardcoded ?

ui/src/containers/NodeVolumes.js
```diff
@@ -87,6 +91,12 @@ const LoaderContainer = styled(Loader)`
   padding-right: ${padding.smaller};
 `;
 
+const TrashButtonContainer = styled(Button)`
+  ${props => {
+    if (props.disabled) return { opacity: 0.2 };
```
MonPote, line 96: Nothing to change here but we might want to discuss about the opacity for disabled stuff with Valamir.

ui/src/containers/VolumeDetailCard.js
```diff
@@ -0,0 +1,249 @@
...
+`;
+
+const VolumeInformation = styled.div`
+  width: 40vw;
+`;
+
+const VolumeNameTitle = styled.div`
+  color: ${(props) => props.theme.brand.textPrimary};
+  font-size: ${fontSize.large};
+  padding: 10px 0 10px 22px;
```
MonPote: is 22px really necessary ?

ui/src/containers/VolumeDetailCard.js
```diff
@@ -0,0 +1,249 @@
...
+`;
+
+const InformationSpan = styled.div`
+  padding-bottom: 7px;
+  padding-left: 22px;
```
MonPote: Same here for padding

ui/src/containers/VolumeDetailCard.js
```diff
@@ -0,0 +1,249 @@
...
+  color: ${(props) => props.theme.brand.textPrimary};
+  font-size: ${fontSize.base};
+`;
+
+const DeleteButton = styled(Button)`
+  margin-right: 100px;
+  width: 135px;
+  height: 30px;
+  font-size: ${fontSize.small};
+  background-color: #be2443;
```
MonPote: We maybe should use the theme here.

ui/src/containers/VolumeDetailCard.js
```diff
@@ -0,0 +1,249 @@
...
+  display: flex;
+  justify-content: center;
+  padding-top: ${padding.larger};
+`;
+
+const VolumeGraph = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  width: 37vw;
```
MonPote: That's a bit weird to me, can you explain why it's `37vw` ?

ui/src/containers/VolumePage.js
```diff
@@ -0,0 +1,750 @@
...
+    pod.volumes.find((volume) => volume.persistentVolumeClaim === PVCName),
+  );
+
+  // get the alert base on the current
+  const alertlist = alerts?.list?.filter(
+    (alert) => alert.labels.persistentvolumeclaim === PVCName,
+  );
+
+  // Data for volume table
+  const volumeListData = volumes.map((volume) => {
```
MonPote: This one is a bit tricky, we might want to use reselect for this one to prepare the data before arriving to this component, you can achieve that with something like this :

```js
// VolumePage.js
const volumeListData = useSelector((state) => makeVolumeList(state, match));
```

```js
// src/services/utils.js
const getVolumeFromUrl = createSelector(
  getVolumeNameFromUrl,
  getVolumes,
  (volumeName, volumes) => {
    return volumes.find(
      (volume) =>
        volume && volume.metadata && volume.metadata.name === volumeName,
    );
  },
);

const makeGetNodeFromUrl2 = createSelector(
  (state, match) => {
    if (match && match.params && match.params.id) {
      return match.params.id;
    } else {
      return '';
    }
  },
  getNodes,
  (nodeName, nodes) => nodes.find((node) => node.name === nodeName) || {},
);

export const makeTestVolumes = createSelector(
  getVolumeFromUrl,
  getVolumes,
  (state) => state.app.volumes.pVList,
  (state) => state.app.volumes.pVCList,
  (state) => state.app.monitoring.volumeStats.volumeUsedCurrent,
  (state) => state.app.monitoring.alert,
  makeGetNodeFromUrl2,
  (state) => state.app.monitoring.volumeStats.volumeLatencyCurrent,
  (
    volume,
    volumes,
    pVList,
    pVCList,
    volumeUsedCurrentList,
    alerts,
    node,
    volumeLatencyCurrent,
  ) => {
    // console.log('makeTestVolumes');
    if (!volume) {
      return [];
    }

    return volumes.map((volume) => {
      const volumePV = pVList.find(
        (pV) => pV.metadata.name === volume.metadata.name,
      );
      // find the mapping PVC of this specific volume
      const volumePVC = pVCList.find(
        (pVC) => pVC.spec.volumeName === volume.metadata.name,
      );

      const volumeComputedCondition = computeVolumeCondition(
        computeVolumeGlobalStatus(volume.metadata.name, volume?.status),
        volumePV?.status?.phase === STATUS_BOUND
          ? intl.translate('yes')
          : intl.translate('no'),
      );

      let volumeUsedCurren = null;
      let volumeAlerts = [];
      let volumeHealth = '';

      // if volume is bounded
      if (volumePVC) {
        volumeUsedCurren = volumeUsedCurrentList?.find(
          (volStat) =>
            volStat.metric.persistentvolumeclaim === volumePVC.metadata.name,
        );

        // filter the alerts related to the current volume.
        volumeAlerts = alerts?.list?.filter(
          (alert) =>
            alert.labels.persistentvolumeclaim === volumePVC.metadata.name,
        );
      } else {
        volumeHealth = STATUS_NONE;
      }

      // THE RULES TO COMPUTE THE HEALTH
      // compute the volume health based on the severity of the alerts
      // critical => if there is at least one critical
      if (volumeAlerts.length) {
        volumeHealth = volumeAlerts.find(
          (vol) => vol.labels.severity === STATUS_CRITICAL,
        )
          ? STATUS_CRITICAL
          : STATUS_WARNING;
      } else if (volumeComputedCondition === ('exclamation' || 'unlink')) {
        volumeHealth = STATUS_NONE;
      } else {
        volumeHealth = STATUS_HEALTH;
      }

      return {
        name: volume.metadata.name,
        node: node.name,
        usage: volumeUsedCurren?.value[1]
          ? Math.round(
              (volumeUsedCurren?.value[1] /
                (volumePV?.spec?.capacity?.storage &&
                  allSizeUnitsToBytes(volumePV?.spec?.capacity?.storage))) *
                100,
            )
          : 0,
        status: volumeComputedCondition,
        bound:
          volumePV?.status?.phase === STATUS_BOUND
            ? intl.translate('yes')
            : intl.translate('no'),
        storageCapacity:
          volumePV?.spec?.capacity?.storage || intl.translate('unknown'),
        storageClass: volume.spec.storageClassName,
        creationTime: formatDate(volume.metadata.creationTimestamp),
        usageRawData: volumeUsedCurren?.value[1]
          ? bytesToSize(volumeUsedCurren?.value[1])
          : 0,
        health: volumeHealth,
        // when the data is not ready, display `loading` for the moment
        latency: volumeLatencyCurrent
          ? Math.round(
              volumeLatencyCurrent?.find(
                (vLV) => vLV.metric.device === volume.status.deviceName,
              )?.value[1] * 100,
            ) + 'ms'
          : 'loading',
      };
    });
  },
);
```

Please note that this is a quick draft, there might need some refinement. But that way it's already "easier" to test this critical part of the page (basically most login behind you `Table`.

ui/src/containers/VolumePage.js
```diff
@@ -0,0 +1,750 @@
...
+        : 'loading',
+    };
+  });
+
+  const currentVolume = volumeListData.find(
+    (vol) => vol.name === currentVolumeName,
+  );
+
+  // Define a default UI for filtering
+  function GlobalFilter({
```
MonPote: I believe this should be extract outside of the `VolumePage` component (not necessary in a new file but outside of the component.

ui/src/containers/VolumePage.js
```diff
@@ -0,0 +1,750 @@
...
+          onClick={() => {
+            history.push(`/nodes/${node.name}/createVolume`);
+          }}
+        />
+      </ActionContainer>
+    );
+  }
+
+  // React Table for the volume list
+  function Table({ columns, data, getTrProps }) {
```
MonPote: Same comment than GlobalFilter.

ui/src/containers/VolumePage.js
```diff
@@ -0,0 +1,750 @@
...
+              volumeIOPSRead
+                ? addMissingDataPoint(
+                    jointDataPointBaseonTimeSeries(volumeIOPSRead),
+                    queryStartingTime,
+                    7,
+                    1,
+                  )
+                : null
+            }
+            volumeIOPSWrite={
```
MonPote: I think that `volumeUsed `, `volumeStorageCapacity `, `volumeThroughputWrite `, `volumeThroughputRead `, `volumeIOPSRead ` and `volumeIOPSWrite ` should be defined inside `PerformanceGraphCard` and not in `VolumePage`.

It's a 1 to 1 relationship and add a lots of noise to the component for nothing. `PerformanceGraphCard` will become a `container` then.

If you want to do something reusable then you may want to do it a bit lower.

ui/src/containers/VolumePage.js
```diff
@@ -0,0 +1,750 @@
...
+            volumeUsagePercentage={currentVolume?.usage}
+            volumeUsageBytes={currentVolume?.usageRawData ?? 0}
+            storageCapacity={
+              volumeListData?.find((vol) => vol.name === currentVolumeName)
+                .storageCapacity
+            }
+            health={
+              volumeListData?.find((vol) => vol.name === currentVolumeName)
+                .health
+            }
```
MonPote: Same comment than for `PerformanceGraphCard ` *but I'm less sure*.
It may be a good thing to leave it that way. Can you have a look ?

ui/src/containers/VolumePage.js
ChengYanJin: I extracted the `getVolumeListData` function to `NodeVolumesUtils.js` since it's not a global utility function yet.

ui/src/containers/VolumePage.js
ChengYanJin: Here you definitely right. I should make `PerformanceGraphCard` a container, and it's going to fetch its own data.

ui/src/containers/VolumePage.js
ChengYanJin: Extracted the volume table to a separate component.

ui/src/containers/VolumePage.js
ChengYanJin: IMO, I think we could leave it in this way since we just pass some data to the detail card component. This is no big difference if we fetch the state in detail card component or volume page component.

gdemonet: When not sure whether to implement a component as pure display or as smart / stateful, I try to think about lifecycle. More specifically, about comparing the lifecycle of the component vs. the lifecycle of the data it uses.

Let's take an example: say I'm displaying the Node Details tab component. This component's data lives with the Node object, and for the most part (except the metrics), will not change a lot. The component itself lives much shorter: switching between tabs will recreate new components everytime (which can be even shorter than metrics refresh rate).
In this case, I would build the tab component to be purely about display (dumb component), and make sure the parent would provide it with a fresh Node object (including its stats).

(as you may have guessed, this means that IMO, neither of the components aforementioned should be smart, outside the "data extraction" logic they embed)

Related to [the discussion above](https://github.com/scality/metalk8s/pull/2665#discussion_r462082994), I think the reasoning was biased. The symptom @MonPote observed is well summed up in:

> It's a 1 to 1 relationship and add a lots of noise to the component for nothing

This is entirely true, and we should think of a solution to this problem.

But the proposed one is only solving it by side-effect: if the `PerformanceGraphCard` is a container, it means we change the ownership of data, and as such, will not need to worry about passing down values anymore since the component will fetch it on its own once mounted.
Another solution we could consider to the above problem, without changing ownership of data, is to change the contract between parent and child: pass a full `volumeStats` object as a prop to `PerformanceCard`, and let it handle the extraction of information to display (specific time series for each graph it needs based on known keys), and let the parent (`VolumePage`) do the mapping between a `volume` object and the relevant `volumeStats` required by the performance card.

A "dumb component" is not as dumb as to need flattened information. "Dumb" is to distinguish components managing state (performing creative/destructive actions on application state, from user operations or external data consumption) versus components consuming this state managed for them. It doesn't mean you can't write code outside of HTML/JSX in them ;)

The only component in this page which makes sense to be a container is the `VolumePage`. It manages a list of volumes, and dispatches relevant information to each display components it shows, maintaining a cached (though probably refreshed over time) view of each volume.

One could argue that we don't want to refresh stats for an unselected volume, especially if we have a lot of them (though honestly not sure that would make a big impact if we group everything in a single query to Prometheus). If that's the case, then the next level would be to consider a component taking care of "one selected volume" (which we don't have in the current implementation). This component would be made a container, able to fetch and maintain the metrics for a volume which was passed by its parent, the `VolumePage`.

However, let's consider the lifecycle question again: the `VolumePage` component receives the URL `match` as a prop. If another volume than the current one is selected, the URL will change, and the component will be re-rendered (which, in this case, also means dispatching some actions to the store which in turn will send calls to the K8s API). A big chunk of `VolumePage` managed data (the list of all volumes, the selected filters, the active tab), on the other hand, would (probably) outlive the change between selected Volumes.
So... we need to make this page a "dumb" component as well? Who takes care of retrieving those volumes in that case? IMO, that's a sign that a container is owning a piece of state that it should not (in this case, the selected Volume name).

Imagine if the `VolumePage` was mounted and kept intact for all URLs (`useRouteMatch("/volumes")`). This container can manage a list of volumes, handle modals and their interaction with sagas (and I would argue for it to also maintain stats, but this can be debated, so let's say it won't for now, it will make the example more interesting). This container also renders a sub-router component, which in turn will match more specifically on the rest of the route (`useRouteMatch("/volumes/:id")`) and handle volume-specific information (such as metrics), which don't need to be maintained when the volume isn't selected.
Now you can work with a single source of the list of volumes, managed by the page container. A child container manages the selected volume (through URL) and receives the list of volumes from its parent. This child container can mount a dumb component for the table on the left, highlighting the selected volume, and another child container for the right-side panel, responsible for refreshing live information about the selected volume it received as a prop.

I hope this long message wasn't too long and that it made my point clear. Let's sum it up:
- For the performance card component, making it a container would have created a lifecycle mismatch (managed data lives longer than the component), and we solved it by bubbling up the state management in its parent component.
- For the Volume page component, we had the opposite mismatch (the information of which volume is selected lives, or should live, shorter than the full Volume page), but couldn't solve it by the same method. Instead, we reduced the scope of managed data in the parent and bubbled down the state management into a child container.

thomasdanan:
Great page, few things I noticed though:

- Need to fix/improve URL and breadcrumb (the node filter is not necessarily always required), think about switching from one volume to another one belonging to one other node
- not able to see all volumes (a scrollbar or pagination is missing)
- When creating a Volume the label text is not visible
- to me, we are missing the labels in the Volume info zone
- the time window selector for charts is not working (I only see the last 7 days option)
- not always able to see latency in table and charts correctly (except %usage)
- I think we should remove the Created from the Volumes table (and have it only in the Volume info zone)
- If we can Create a Volume from Volumes page we probably need to have a field or selector for the node
- When filtering the Volumes (using the search bar), the filter is lost as soon as the page is refreshed. It should be reapplied

ChengYanJin:
Recreate PR in order to merge in 2.6 branch
Merged in PR #2744

PR #2668, opened 2020-07-15
Log centralization design

### LINKED ISSUE #2626: Write design document for Centralized Logging

alexandre-allard:
**Component**: documentation

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2626

### DIFF
--- a/docs/developer/architecture/logs-components-diagram.uml
+++ b/docs/developer/architecture/logs-components-diagram.uml
@@ -1,0 +1,49 @@
+@startuml
+
+node "node1" {
+    journal_logs1 <-- [collector1] : reads from
+    container_logs1 <-- [collector1] : reads from
+}
+
+node "node2" {
+    journal_logs2 <-- [collector2] : reads from
+    container_logs2 <-- [collector2] : reads from
+}
+
+package "distributors" {
+    [distributor1]
+    [distributor2] 
+}
+
+package "ingesters" {
+    [ingester1]
+    [ingester2]
+}
+
+database "storage" {
+    [volume1]
+    [volume2]
+}
+
+package "queriers" {
+    [querier1]
+    [querier2]
+}
+
+collector1 --> distributor1
+collector1 --> distributor2
+collector2 --> distributor1
+collector2 --> distributor2
+
+distributor1 --> ingester1
+distributor2 --> ingester2
+
+ingester1 --> volume1 : writes to
+ingester2 --> volume2 : writes to
+
+querier1 --> ingester1 : reads from
+querier1 --> volume1 : reads from
+querier2 --> ingester2 : reads from
+querier2 --> volume2 : reads from
+
+@enduml
--- a/docs/developer/architecture/logs.rst
+++ b/docs/developer/architecture/logs.rst
@@ -22,7 +22,7 @@ Browsing the logs is accessible through an API and a UI. The UI should ease
 correlation between logs and health/performance KPIs as well as alerts.
 
 For organisations having their own Log centralization system (like Splunk or
-Elasticsearch), MetalK8s should provide some documentation in order guide the
+Elasticsearch), MetalK8s should provide some documentation to guide the
 customer to deploy and configure its own log collection agent.
 
 The following requirements focus on application logs. Audit logs are not part
@@ -197,6 +197,531 @@ overview page.
 A Grafana dashboard gathering health/performance KPIs, as well as alerts for
 log centralization service is available when deploying/upgrading MetalK8s.
 
+Components
+----------
+
+Our Log Centralization system can be split into several components as follows:
+
+.. uml:: logs-components-diagram.uml
+
+Collector
+~~~~~~~~~
+
+The collector is responsible for processing logs from all the sources
+(files, journal, containers, ...), enriching the logs with metadata (labels)
+coming from parsing/filtering them (e.g. drop record on regexp match), from the
+context (e.g. file path, host) or by querying external sources such as APIs,
+then finally forwarding these logs to one or multiple distributors.
+
+Distributor (Router)
+~~~~~~~~~~~~~~~~~~~~
+
+The distributor is the component that receives incoming streams from the
+collectors, it validates them (e.g. labels format, timestamp), then forwards
+them to the ingester.
+The distributor can also do parsing/filtering on the streams to enrich
+them with metadata (labels), route to a specific ingester or even drop them.
+It can as well do some buffering to avoid nagging the ingester with queries
+or to wait a bit in case the ingester would be unresponsive for a moment.
+
+Ingester (Storage)
+~~~~~~~~~~~~~~~~~~
+
+The ingester serves as a buffer between distributor and storage, because
+writing large chunks of data is more efficient than writing each event
+individually as it arrives.
+As such, a querier may need to ask an ingester about what is in the buffer.
+
+Querier
+~~~~~~~
+
+The querier interprets the queries it receives from clients and then asks the
+ingesters for the corresponding data, then fallback on storage backend if not
+present in memory and returns it to the clients.
+It also takes care of deduplication of data because of the replication.
+
+Design Choices
+--------------
+
+To choose which solution fits best our needs, we did a benchmark of the
+shortlisted collectors to compare their performances, on a 4 K8s nodes
+architecture (3 infra + 1 workload), with Loki as backend.
+
+Our choice for the final design has been greatly motivated by these numbers,
+it represents the global resources consumption for the whole log centralization
+stack (Loki included).
+
+With 10k events/sec, composed of 10 distinct streams:
+
+   +----------------------+------------------+--------------------+
+   |                      | **CPU avg in m** | **RAM avg in MiB** |
+   +----------------------+------------------+--------------------+
+   | promtail             | 975              | 928                |
+   +----------------------+------------------+--------------------+
+   | fluent-bit           | 790              | 830                |
+   +----------------------+------------------+--------------------+
+   | fluent-bit + fluentd | 1311             | 1967               |
+   +----------------------+------------------+--------------------+
+
+With 10k events/sec, composed of 1000 distinct streams:
+
+   +----------------------+------------------+--------------------+
+   |                      | **CPU avg in m** | **RAM avg in MiB** |
+   +----------------------+------------------+--------------------+
+   | promtail             | 1292             | 1925               |
+   +----------------------+------------------+--------------------+
+   | fluent-bit           | 1040             | 834                |
+   +----------------------+------------------+--------------------+
+   | fluent-bit + fluentd | 1447             | 1902               |
+   +----------------------+------------------+--------------------+
+
+We can see that the Fluent Bit + Loki couple has the smallest impact on
+resources, but also that the Fluent Bit + Fluentd + Loki architecture seems
+to offer a better scaling, with the possiblity of keeping less pressure on
+workload nodes (Fluent Bit), relying more on dedicated infra nodes (Fluentd).
+
+.. todo::
+
+   define sizing rules for CPU & RAM based on log streams and rate
+
+Fluent Bit + Loki
+~~~~~~~~~~~~~~~~~
+
+Fluent Bit
+""""""""""
+
+We choose Fluent Bit as the collector because it allows to scrape all the logs
+we need (journal & containers), enriches them with the Kubernetes API and it
+has a very low resources footprint.
+
+Moreover, it supports multiple backend such as Loki, ES, Splunk, etc. at
+the same time, which is a very important point if a user also wants to forward
+the logs to an external log centralization system (e.g. long term archiving).
+
+Check `here <https://docs.fluentbit.io/manual/pipeline/outputs>`__
+for the official list of supported outputs.
+
+Loki
+""""
+
+Loki has been chosen as the distributor, ingester & querier because, like
+Fluent Bit, it has a really small impact on resources and is very cost
+effective regarding storage needs.
+
+It also uses the LogQL syntax for queries, which is pretty close to what
+we already have with Prometheus and PromQL, so it eases the integration
+in our tools.
+
+Rejected Design Choices
+-----------------------
+
+Promtail + Loki
+~~~~~~~~~~~~~~~
+
+Promtail has been rejected because, even if it consumes very few resources,
+it can only integrate with Loki and we need something more versatile, with
+the ability to interact with different distributors.
+
+Fluentd + Loki
+~~~~~~~~~~~~~~
+
+This architecture has been rejected because it means we need 1 Fluentd instance
+per node, which increases a lot the resources consumption compared to other
+solutions.
+
+Fluent Bit + Fluentd + Loki
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We have considered this architecture as the Fluent Bit + Fluentd couple seems
+to be a standard in the industry, but we didn't find any reason of keeping
+Fluentd, apart for its large panel of plugins which we don't really need.
+Fluent Bit alone seems to be sufficient for what we want to achieve and adding
+an extra Fluentd means more resources consumption and add unnecessary
+complexity in the log centralization stack.
+
+Logging Operator
+~~~~~~~~~~~~~~~~
+
+`Logging Operator`_ seemed to be a good candidate for the implementation
+we choose, offering the ability to deploy and configure Fluent Bit and Fluentd,
+but Fluentd is not optional and seems to have a central place as most of the
+parsing/filtering is done by this one, which means a bigger footprint on the
+hardware resources.
+
+.. _Logging Operator: https://github.com/banzaicloud/logging-operator
+
+Logstash + Elasticsearch
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This architecture is probably the most common one, but it has not been taken
+into consideration because we want to focus on having the minimum resources
+consumption and these components can really hog RAM & CPU. Beats could be used
+as the log collector to reduce the impact of Logstash, but Elasticsearch still
+consumes a lot of resources.
+Even if this solution offers a lot of powerful functionalities
+(e.g. distributed storage, full-text indexing), we don't really need them and
+want to focus on the smallest hardware footprint.
+
+Implementation Details
+----------------------
+
+Deployment
+~~~~~~~~~~
+
+All the components will be deployed using Kubernetes manifests through Salt
+inside a ``metalk8s-logging`` Namespace.
+
+Fluent Bit
+""""""""""
+
+Fluent Bit will be deployed as a DaemonSet, because we need one collector on
+each node to be able to collect logs from both the Kubernetes platform, the
+applications that run on top of it and all the system daemons running alongside
+(e.g. Salt minion, Kubelet).
+
+This DaemonSet and Fluent Bit configuration will be handled by the Fluent Bit
+Operator.
+For this, we need to deploy the manifests found
+`here <https://github.com/kubesphere/fluentbit-operator/tree/master/manifests/setup>`__,
+using our Salt Kubernetes renderer.
+
+We will then need to also automatically deploy CRs with Fluent Bit default
+configuration, examples can be found
+`here <https://github.com/kubesphere/fluentbit-operator/tree/master/manifests/logging-stack>`__.
+
+Loki
+""""
+
+Loki has 2 deployment mode, either as microservices (with distributor, ingester
+and querier in distinct pods), either as a monolith.
+We must use the monolithic mode, because we use filesystem as the storage
+backend and microservice mode does not support it.
+
+Loki will be deployed as a ``StatefulSet`` as it needs ``PersistentVolume``
+to write the logs. It will be running only on infra nodes and we need at least
+2 replicas of it (except on a single node platform), to ensure its high
+availability.
+
+As we already did for other components (e.g. Prometheus Operator), the
+manifests for Loki will be generated statically by rendering
+the `loki helm chart`_::
+
+  helm repo add loki https://grafana.github.io/loki/charts
+  helm repo update
+  helm fetch -d charts --untar loki/loki
+  ./charts/render.py loki --namespace metalk8s-logging \
+    charts/loki.yaml charts/loki/ \
+    > salt/metalk8s/addons/loki/deployed/charts.sls
+
+.. _loki helm chart: https://github.com/grafana/loki/tree/master/production/helm
+
+Loki Storage
+____________
+
+Since we're using filesystem to store Loki's data, we have basically 2 ways
+for having multiple Loki instances running at the same time, with the same
+set of data.
+
+Either we do like Prometheus and we store everything on every instance of
+Loki, but it means we raise the storage, RAM and CPU needs for each additionnal
+instance, either we use a `new experimental feature`_ from Loki 1.5.0 where
+Loki ingesters use a hash ring and talk between them to route the queries to
+the right one. This approach needs an external KV store to work such as
+etcd or Consul.
+
+We choose to use the first approach as the second is not production ready
+and since we plan to only do short term retention on Loki, the impact on
+storage will not be that much important.
+Moreover, it eases the deployment and maintenance since there is not an
+extra component.
+
+.. _new experimental feature: https://github.com/grafana/loki/blob/master/docs/operations/storage/filesystem.md#new-and-very-experimental-in-150-horizontal-scaling-of-the-filesystem-store
+
+.. todo::
+
+  define loki storage sizing depending on retention, ingestion rate,
+  average size of logs, number of streams, ...
+
+Configuration
+~~~~~~~~~~~~~
+
+Fluent Bit
+""""""""""
+
+Fluent Bit needs to be configured to scrape and handle properly journal and
+containers logs by default.
+
+For containers logs, we want to add the following labels:
+
+ - node: the node it comes from
+ - namespace: the namespace the pod is running in
+ - instance: the name of the pod
+ - container: the name of the container
+
+For journal we want these labels:
+
+ - node: the node it comes from
+ - unit: the name of the unit generating these logs
+
+This configuration will also be customizable by the user to be able to
+add new routes (Output) to push the log streams to.
+
+This configuration will be done through various CRs provided by the Fluent Bit
+Operator:
+
+ - FluentBit: Defines Fluent Bit instances and its associated config
+ - FluentBitConfig: Select input/filter/output plugins and generates the final
+   config into a Secret
+ - Input: Defines input config sections
+ - Filter: Defines filter config sections
+ - Output: Defines output config sections
+
+For example, if a user wants to forward all Kubernetes logs to an external log
+centralization system (e.g. Elasticsearch), he will need to define an
+``Output`` CR as follows:
+
+.. code-block:: yaml
+
+  kind: Output
+  metadata:
+    name: my-output-to-external-es
+    namespace: my-namespace
+  spec:
+    match: kube.*
+    es:
+      host: 10.0.0.1
+      port: 9200
+
+More details can be found on `Fluent Bit Operator repository`_ and manifest
+samples are
+`here <https://github.com/kubesphere/fluentbit-operator/tree/master/manifests/logging-stack>`_.
+
+.. _Fluent Bit Operator repository: https://github.com/kubesphere/fluentbit-operator
+
+Loki
+""""
+
+Loki's configuration is stored as a Secret.
+we need to expose few parameters to the user for customization
+(e.g. retention).
+Since we do not have Operator and CRs for Loki, we will use the CSC mechanism
+to provide the interface for customization, with a ConfigMap
+``metalk8s-loki-config`` in the ``metalk8s-logging`` Namespace.
+CSC is not as powerful as an Operator with CRs (no watch and reconciliation on
+resources and need to run Salt state manually), but the Loki configuration will
+not change that much, probably during deployment and to tune few parameters
+afterwards, so it does not worth to invest on an Operator.
+
+The CSC ``ConfigMap`` will look like the followings:
+
+.. code-block:: yaml
+
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: metalk8s-loki-config
+    namespace: metalk8s-logging
+  data:
+    config.yaml: |-
+      apiVersion: addons.metalk8s.scality.com
+      kind: LokiConfig
+      spec:
+        deployment:
+          replicas: 1
+        config:
+          auth_enabled: false
+          chunk_store_config:
+            max_look_back_period: 168h
+          ingester:
+            chunk_block_size: 262144
+            chunk_idle_period: 3m
+            chunk_retain_period: 1m
+            [...]
+
+With default values fetched from a YAML file as it is already done for Dex,
+Alertmanager and Prometheus.
+
+Monitoring
+~~~~~~~~~~
+
+Prometheus
+""""""""""
+
+To monitor every services in our log centralization stack, we will need to
+deploy ``ServiceMonitor`` object and expose the ``/metrics`` route of all
+these components.
+It will allow Prometheus Operator to configure Prometheus and automatically
+start scraping these services.
+For Loki, this can be achieved by adding the following configuration in its
+helm chart ``charts/loki.yaml`` values:
+
+.. code-block:: yaml
+
+  serviceMonitor:
+    enabled: true
+    additionalLabels:
+      release: "prometheus-operator"
+
+For Fluent Bit, we will need to define a ``ServiceMonitor`` object:
+
+.. code-block:: yaml
+
+  apiVersion: monitoring.coreos.com/v1
+  kind: ServiceMonitor
+  metadata:
+    labels:
+      app: fluent-bit
+      release: prometheus-operator
+    name: fluent-bit
+    namespace: metalk8s-logging
+  spec:
+    endpoints:
+    - path: /api/v1/metrics/prometheus
+      port: http-metrics
+    namespaceSelector:
+      matchNames:
+      - metalk8s-logging
+    selector:
+      matchLabels:
+        app: fluent-bit
+
+We also need to define alert rules based on metrics exposed by these services.
+This is done deploying new ``PrometheusRules`` object in ``metalk8s-logging``
+Namespace:
+
+.. code-block:: yaml
+
+  apiVersion: monitoring.coreos.com/v1
+  kind: PrometheusRule
+  metadata:
+    labels:
+      release: prometheus-operator
+    name: loki.rules
+    namespace: metalk8s-logging
+  spec:
+    groups:
+    - name: loki.rules
+      rules: <RULES DEFINITION>
+
+There is some recording and alert rules defined in the `Loki repository`_ that
+could be used as a base, then we could enrich these rules later when we will
+have better operational knowledge.
+
+.. _Loki repository: https://github.com/grafana/loki/tree/master/production/loki-mixin
+
+Grafana
+"""""""
+
+To be able to query logs from Loki, we need to add a Grafana datasource,
+this is done adding a ConfigMap ``loki-grafana-datasource`` in
+``metalk8s-monitoring`` Namespace as follows:
+
+.. code-block:: yaml
+
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: loki-grafana-datasource
+    namespace: metalk8s-monitoring
+    labels:
+      grafana_datasource: "1"
+  data:
+    loki-datasource.yaml: |-
+      apiVersion: 1
+      datasources:
+      - name: Loki
+        type: loki
+        access: proxy
+        url: http://loki.loki.svc.cluster.local:3100
+        version: 1
+
+To display the logs we also need a dashboard, adding a ConfigMap
+``loki-logs-dashboard`` in Namespace ``metalk8s-monitoring``:
+
+.. code-block:: yaml
+
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: loki-logs-dashboard
+    namespace: metalk8s-monitoring
+    labels:
+      grafana_dashboard: "1"
+  data:
+    loki-logs.json: <DASHBOARD DEFINITION>
+
+For the dashboard we will use a view with a simple log panel and variables
+representing labels to filter on.
+Since journal and kubernetes logs will not have the same labels, we could
+either have 2 distinct dashboards or 2 log panels in the same.
+An example of what we want is `Loki dashboard`_.
+
+.. _Loki dashboard: https://github.com/grafana/loki/blob/master/production/loki-mixin/dashboard-loki-logs.json
+
+.. note::
+
+  The ``grafana_datasource: "1"`` and ``grafana_dashboard: "1"`` labels are
+  what is used by the Prometheus Operator to retrieve datasource and dashboard
+  for Grafana, resources must be deployed in `metalk8s-monitoring` namespace.
+
+Loki Volume Purge
+~~~~~~~~~~~~~~~~~
+
+Even with a max retention period, the logs could grow faster than what was
+expected and fill up the volume.
+Since there is no retention based on size in Loki yet, we need to add some
+specific monitoring (with prediction on volume usage) and alerting to ensure
+that an administrator will be warned if such a case would happen.
+The alert message should be clear and provide an URL to a run book to help
+the administrator resolving the issue.
+
+To fix this issue, the administrator should purge oldest log chunk files
+from Loki volume, which can be achieved by connecting to the pod and manually
+removing them.
+If the growth of logs is not something transient, the administrator should
+also be advised to lower the retention period or replace the Loki volume
+by a bigger one.
+
+.. todo::
+
+   Define what should be done for the purge, is it automatic (side-car pod)
+   or is it only alerts with manual operations?
+   If we go with an automatic purge, we should at least expose few parameters
+   for the user to be able to customize this mechanism:
+     - whether this mechanism is activated or not (defaulted to yes or no?)
+     - % of space used for the purge to be triggered (90%?)
+     - minimum days of retention (even if we're above the threshold) to ensure
+       we're not removing everything in case of massive amount of logs
+   We also need to trigger alerts everytime this purge is activated to be sure
+   that the operator is aware that Loki volume is undersized.
+
+Iterations
+~~~~~~~~~~
+
+Iteration 1
+"""""""""""
+
+The goal is to have a working log centralization system, with logs accessible
+from Grafana:
+
+* Deploy Fluent Bit and Loki
+* Customization of Loki with CSC mechanisms
+* Document customization of Loki through CSC
+* Deploy Grafana datasource & dashboard
+* Document the log centralization system (sizing, configuration, ...)
+* Simple pre-merge test to ensure the default log pipeline is working
+
+Iteration 2
+"""""""""""
+
+* Deployment of Fluent Bit with Fluent Bit Operator
+* Document customization of Fluent Bit through CRs
+* Define Prometheus record and alert rules
+* Deploy Loki volumes purge mechanism (TBD)
+* Display log centralization system status on the MetalK8s UI
+* Post-merge tests to ensure customization is working (replicas, custom
+  parser/filter rules, ...)
+
 Documentation
 -------------
 
@@ -209,7 +734,19 @@ the workload properties.
 The Post Installation page is updated to indicate that persistent storage is
 needed for log centralization service.
 
-A new page should be added to explains how to operate the service.
+A new page should be added to explain how to operate the service and how to
+forward logs to an external log centralization system.
 
 The Cluster Monitoring page is updated to describe the log centralization
 service.
+
+Test Plan
+---------
+
+Log centralization system will be deployed by default with MetalK8s, so its
+deployment will automatically be tested during pre-merge integration tests.
+However, we still need to develop specific pytest-bdd test scenario to ensure
+that the default logging pipeline is fully functionnal, and run it during
+these pre-merge tests.
+We will also add more complex tests in post-merge such as configuring
+specific parsers/filters, scaling the system, etc.
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -15,6 +15,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         latexmk \
         make \
         plantuml \
+        graphviz \
         python3.6 \
         python3-buildbot-worker \
         python3-pip \
--- a/docs/developer/architecture/index.rst
+++ b/docs/developer/architecture/index.rst
@@ -14,3 +14,4 @@ Architecture Documents
    solutions
    ci
    volume
+   logs


docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,422 @@ overview page.
...
+effective regarding storage needs.
+
+It also uses the LogQL syntax for queries, which is pretty close to what
+we already have with Prometheus and PromQL, so it eases the integration
+in our tools.
+
+Fluentd (optional)
+""""""""""""""""""
+
+Fluentd could be optionally added between Fluent Bit and Loki to offers
```
sayf-eddine-scality: offer

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,422 @@ overview page.
...
+the logs to an external log centralization system (e.g. long term archiving).
+
+Check here_ for the official list of supported outputs.
+
+.. _here: https://docs.fluentbit.io/manual/pipeline/outputs
+
+Loki
+""""
+
+Loki has been choosen as the distributor, ingester & querier because, like
```
sayf-eddine-scality: chosen

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,422 @@ overview page.
...
+collectors, validates and prepares the streams for storage, then forwards
+them to the right ingester.
+The distributor can also do some processing on the streams to enrich them
+with metadata (labels).
+
+Ingester (Storage)
+~~~~~~~~~~~~~~~~~~
+
+The job of the ingester is to write log streams on persistent storage such
+as a filesystem or a database, but also to return the data requested by
```
sayf-eddine-scality: in Loki ecosystem Ingesters only return data that have not been flashed to storage they re not used to read the data back,no?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,422 @@ overview page.
...
+
+.. _new experimental feature: https://github.com/grafana/loki/blob/master/docs/operations/storage/filesystem.md#new-and-very-experimental-in-150-horizontal-scaling-of-the-filesystem-store
+
+.. note::
+
+  What do we do ? Do we use metalk8s etcd to store loki ring ? :/
+
+  Or do we add a new instance of etcd or consul ?
+
+  This feature is marked as experimental, should we even use it ?
```
sayf-eddine-scality: Looks very prone to changes if you ask me :/

docs/developer/architecture/logs.rst
alexandre-allard: true, it's not clear here, even if this part of the documentation is not directly related to Loki, but more a general concept.

docs/developer/architecture/logs.rst
alexandre-allard: I agree, and probably bugged too.
But otherwise we can only do "replication" like Prometheus, meaning we have to send all streams on all Loki instances with the overhead of resources it implies.
Maybe we can go for the dirty solution and then if it becomes stable at some point, implement this.
But, to me, the goal of using Loki is to reduce storage (and also RAM & CPU) needs, if we go this way, it'll obviously not be the best.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,447 @@ overview page.
...
+
+Fluent Bit will be deployed as a DaemonSet, because we need one collector on
+each control or workload plane node to be able to collect logs from both the
+Kubernetes platform and the applications that run on top of it.
+
+Loki
+""""
+
+Loki will be deployed as a StatefulSet as it needs persistent volumes to write
+the logs. It will be running only on control plane nodes and we need at least
```
TeddyAndrieux: To me loki will be stuck to infra nodes not control plane

```suggestion
the logs. It will be running only on infra nodes and we need at least
```

docs/developer/architecture/logs.rst
TeddyAndrieux: Since the Loki part of MetalK8s is only here for short term retention to me it's ok to store everything on every instance of Loki.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,447 @@ overview page.
...
+
+  Or do we add a new instance of etcd or consul ?
+
+  This feature is marked as experimental, should we even use it ?
+
+Fluentd
+"""""""
+
+Fluentd will be deployed as a Deployment with constraints to schedule it only
+on control plane nodes.
```
TeddyAndrieux: Same as Loki to me it's on infra nodes

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,447 @@ overview page.
...
+ - container: the name of the container
+
+For journal we want these labels:
+
+ - node: the host it comes from
+ - unit: the name of the unit generating these logs
+
+.. note::
+
+  What if host != node ?
```
TeddyAndrieux: To me we may want to have node name everywhere and not hostname (or maybe having both :thinking:) , no ?

docs/developer/architecture/logs.rst
alexandre-allard: true.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, I mean we want to be able to filter with a single name to get both logs from journal and k8s, but the hostname is not necessarily equal to the node name.
Maybe we can find a way to enrich the containers logs with the hostname.

docs/developer/architecture/logs.rst
alexandre-allard: Yes

docs/developer/architecture/logs.rst
gdemonet: As discussed offline, we want to consider our infra services as simple _poor man's solutions_, and document how to get metrics/logs/alerts/... out of the cluster into other systems (e.g. Splunk for logs).

alexandre-allard: Ok, good to me, just wanted to be sure everyone is aware and agree on this.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,447 @@ overview page.
...
+in our tools.
+
+Fluentd (optional)
+""""""""""""""""""
+
+Fluentd could be optionally added between Fluent Bit and Loki to offer
+more plugins and reduce the impact of Fluent Bit on workload nodes.
+
+It's not recommended for small platforms (few nodes) as it adds a big overhead
+in resources consumption.
```
gdemonet: It's not clear what role Fluentd would play out in this case: Distributor, or a part of the Collector pipeline with Fluentbit?

For the "optional" part of this, is the design stating our implementation may include Fluentd, or is it saying we will have both options available in the product and expose this to our users?

alexandre-allard: Oh yes, I agree, that's not crystal clear.
Fluentd will be part of the Collector pipeline, I'll explain it.

It means we will have both options available, depending on the needs, I will also try to explain a bit more.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+For journal we want these labels:
+
+ - node: the host it comes from
+ - unit: the name of the unit generating these logs
+
+.. note::
+
+  What if host != node ?
+
+  Do we want more labels (e.g. "app", "environment") ?
```
thomasdanan: since we have namespace and container, I would say it is sufficient for a first implem

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+  How do we allow Solutions to add filters/parsers (not in the reqs) ?
```
thomasdanan: Adding filters means, adding label to the container logs?
Not sure to understand what means adding parser?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+.. code-block:: yaml
+
+  metrics:
+    serviceMonitor:
+      enabled: true
+      additionalLabels:
+        release: "prometheus-operator"
+
+We also need to define alert rules based on metrics exposed by these services.
```
thomasdanan, line 588: what would be the formula to compute Log centralization Health?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+* Simple pre-merge test to ensure the log pipeline is working
+* Customization of Loki, Fluent Bit and Fluentd (editing the CSC ConfigMap)
+
+Iteration 2
+~~~~~~~~~~~
+
+* Deployment of Fluentd
+* Customization of Loki, Fluent Bit and Fluentd (simple + advanced modes)
+* Document how to configure these components
+* Define monitoring rules
```
thomasdanan: and display in MetalK8s Overview page

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+""""""""""
+
+Fluent Bit will be deployed as a DaemonSet, because we need one collector on
+each control or workload plane node to be able to collect logs from both the
+Kubernetes platform and the applications that run on top of it.
+
+Loki
+""""
+
+Loki will be deployed as a StatefulSet as it needs persistent volumes to write
```
thomasdanan: Don't we need specific labels when provisioning PVs for Loki? 
Do we recommend having dedicated disks or can we use LVs? do we need SSD or HDD is okay? Can we easily increase PV capacity to handle more logs?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
 A Grafana dashboard gathering health/performance KPIs, as well as alerts for
 log centralization service is available when deploying/upgrading MetalK8s.
 
+Components
+----------
+
+Our Log Centralization system can be splitted into several components as
```
gdemonet: "split", not "splitted"

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+follows:
+
+.. parsed-literal::
+
+  +-----------+     +-------------+     +----------+     +---------+
+  | Collector | ==> | Distributor | ==> | Ingester | <=> | Querier |
+  +-----------+     +-------------+     +----------+     +---------+
+
+.. todo:: do a better schema. :)
```
gdemonet: What would you add? The only thing missing from the schema, IMHO, is a way to see how these components can scale (e.g. if I have multiple collectors, do they all send to a single distributor? or do they each have their own? something in between?)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+Collector
+~~~~~~~~~
+
+The collector is responsible for processing logs from all the sources
+(files, journal, containers, ...), enriching the logs with metadata (labels)
+coming from parsing/filtering them or by querying external sources such as
```
gdemonet: Or coming from context (the source itself can provide metadata: e.g. the file path, container name and parent pod...)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+APIs, then finally forwarding these logs to one or multiple distributors.
+
+Distributor (Router)
+~~~~~~~~~~~~~~~~~~~~
+
+The distributor is the component that receives incoming streams from the
+collectors, validates and prepares the streams for storage, then forwards
+them to the right ingester.
+The distributor can also do some processing on the streams to enrich them
```
gdemonet: Can't a distributor have other functions, like filtering or buffering?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+The collector is responsible for processing logs from all the sources
+(files, journal, containers, ...), enriching the logs with metadata (labels)
+coming from parsing/filtering them or by querying external sources such as
+APIs, then finally forwarding these logs to one or multiple distributors.
+
+Distributor (Router)
+~~~~~~~~~~~~~~~~~~~~
+
+The distributor is the component that receives incoming streams from the
+collectors, validates and prepares the streams for storage, then forwards
```
gdemonet: What do we mean by validation and preparation of log streams? Is the distributor responsible for chunking/formatting? I thought that was the ingester's job.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+them to the right ingester.
+The distributor can also do some processing on the streams to enrich them
+with metadata (labels).
+
+Ingester (Storage)
+~~~~~~~~~~~~~~~~~~
+
+The job of the ingester is to write log streams on persistent storage such
+as a filesystem or a database, but also to return the data requested by
+the queriers, if still in memory.
```
gdemonet: To make it more obvious, the ingester serves as a buffer between distributor and storage, because writing large (potentially compressed) chunks of data (potentially over the network) is more efficient than writing each event individually as it arrives. As such, a querier may need to ask an ingester about what is in the buffer.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+Querier
+~~~~~~~
+
+The querier interprets the queries it receives from clients and then asks the
+ingesters for the corresponding data, then fallback on storage backend if not
```
gdemonet, line 239: Is it a mandatory behavior? Can't a querier ignore ingesters entirely, focusing on querying the storage? How does a querier know about ingesters? What if the ingester is busy, will the query wait? Isn't this pattern encouraging tight coupling between querier and ingester?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+present in memory and returns it to the clients.
+It also takes care of deduplication of data because of the replication.
```
gdemonet, line 241: Can it aggregate results from multiple storage backends? Imagine: I'm configuring my ingesters to save all logs on filesystem, with 1 week retention, and to also save error logs in an external Cassandra DB, with 2 months retention. Is this feasible? How do queries behave in this case?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+Design Choices
+--------------
+
+To choose which solution fits the most our needs, we did a benchmark of some
```
gdemonet: "fits best" i.s.o. "fits the most"

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+of the solutions to compare their performances.
+Our choice for the final design has been greatly motivated by these numbers,
```
gdemonet: Ending with a comma, the suspense is too damn high! ;)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+Check `this document`_ for details.
+
+.. _this document: https://docs.google.com/spreadsheets/d/1f2Pz9zBbkDzpHr8KzST3nTcfvWJqgoNcCEd0-4WIeow/edit#gid=0
```
gdemonet: Private docs in a public repo? Could be worth copying over the results, either as some CSV or directly in a RST table.

(Edit: just read the to-do below, let's have it done before merging please :smirk:)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+It also takes care of deduplication of data because of the replication.
+
+Design Choices
+--------------
+
+To choose which solution fits the most our needs, we did a benchmark of some
+of the solutions to compare their performances.
+Our choice for the final design has been greatly motivated by these numbers,
+
+Check `this document`_ for details.
```
gdemonet: FWIW, you can inline links using the following syntax:

```rst
`my link text <link_url>`_
```

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+.. _this document: https://docs.google.com/spreadsheets/d/1f2Pz9zBbkDzpHr8KzST3nTcfvWJqgoNcCEd0-4WIeow/edit#gid=0
+
+.. todo::
+
+  export the spreadsheet and include it or do directly a quick recap tab here.
+
+Fluent Bit + Fluentd (optional) + Loki
```
gdemonet: Let's have a small schema showing the role of each entity (collector/distributor/...) and how we imagine them being scheduled in a cluster.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Fluent Bit
+""""""""""
+
+We choose Fluent Bit as the collector because it allows to scrape all the logs
+we need (journal & containers), enriches them with the Kubernetes API and it
```
gdemonet, line 294: Is it using the API? I thought it was only reading local files.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+has a very low resources footprint.
+
+Moreover, it supports multiple distributors such as Loki, ES, Splunk, etc. at
```
gdemonet: Distributors?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+It also uses the LogQL syntax for queries, which is pretty close to what
+we already have with Prometheus and PromQL, so it eases the integration
+in our tools.
+
+Fluentd (optional)
+""""""""""""""""""
+
+Fluentd could be optionally added between Fluent Bit and Loki to offer
+more plugins and reduce the impact of Fluent Bit on workload nodes.
```
gdemonet: Let's try to be slightly more explicit:
- more plugins are great, but what for? (For customers trying to push logs in some exotic backend? Some other parsing feature Fluent Bit is missing?)
- reduce the impact how? (IIUC, by removing the need from Fluent Bit to parse log events. Is it still collecting metadata from context?)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+solutions.
+
+Logstash + Elasticsearch
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This architecture is probably the most common one, but it has not been taken
+into consideration because we want to focus on having the minimum resources
+consumption and these components can really hog RAM & CPU. Beats could be used
+as the log collector to reduce the impact of Logstash, but Elasticsearch still
+consumes a lot of resources.
```
gdemonet, line 360: Well, it does a bunch more than Loki as well: distributed storage, full-text indexing...
Given our requirements, we especially want the ingestion load to be small, and that's where Loki shines, while we don't care as much about querying (and that's where ES is strong).

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+Implementation Details
+----------------------
+
+Deployment
+~~~~~~~~~~
+
+All the components will be deployed using Kubernetes manifests through Salt
+inside a ``metalk8s-logging`` Namespace.
+
+As we already did for other components (e.g. Prometheus Operator), the
```
gdemonet: Why not use the Logging Operator from Banzai Cloud?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+We will use the same approach for Fluentd, using the `fluentd helm chart`_::
+
+  helm fetch -d charts --untar stable/fluentd
+  ./charts/render.py fluentd --namespace metalk8s-logging charts/fluentd.yaml \
+    charts/fluentd/ > salt/metalk8s/addons/fluentd/deployed/charts.sls
+
+.. _fluentd helm chart: https://hub.helm.sh/charts/stable/fluentd
+
+Container images must be added in the buildchain
```
gdemonet: Not sure we need this level of detail in the design doc, but will be highly appreciated in implementation tickets ;)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+``buildchain/buildchain/image.py`` and then replaced in
+``charts/loki-stack.yaml`` and ``charts/fluentd.yaml`` using the
+``__image__()`` macro to be replaced by jinja syntax when rendering the
+charts and then dynamically computed at run time.
+
+Fluent Bit
+""""""""""
+
+Fluent Bit will be deployed as a DaemonSet, because we need one collector on
+each control or workload plane node to be able to collect logs from both the
```
gdemonet: No need for distinction between control and workload plane.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+Kubernetes platform and the applications that run on top of it.
```
gdemonet: And all the system stuff running alongside K8s ;) (e.g. Salt minion)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+_______
+
+Since we're using filesystem to store Loki's data, we have basically 2 ways
+for having multiple Loki instances running at the same time, with the same
+set of data.
+
+Either we do like Prometheus and we store everything on every instance of
+Loki, but it means we raise the storage, RAM and CPU needs for each additionnal
+instance, either we use a `new experimental feature`_ from Loki 1.5.0 where
+Loki ingesters use a hash ring and talk between them to route the queries to
```
gdemonet, line 428: Then it's not about HA, but distributed storage, right? If I lose one, I lose data? Or is it doing some form of erasure coding?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+extra component.
+
+.. _new experimental feature: https://github.com/grafana/loki/blob/master/docs/operations/storage/filesystem.md#new-and-very-experimental-in-150-horizontal-scaling-of-the-filesystem-store
+
+Fluentd
+"""""""
+
+Fluentd will be deployed as a Deployment with constraints to schedule it only
+on infra nodes.
+By default we need at least 2 replicas (except on a single node platform) to be
```
gdemonet: How are Fluent Bits forwarding the logs in this case? To each available Fluentd? To the corresponding Service, hoping all logs will end up in Loki somehow?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+Configuration
+~~~~~~~~~~~~~
+
+Fluent Bit
+""""""""""
+
+Fluent Bit needs to be configured to scrape and handle properly journal and
+containers logs by default.
+Its configuration
```
gdemonet: ?

docs/developer/architecture/logs.rst
gdemonet: A fluentbit is tied to a single Node, should be easy for it to add the Node name everywhere, no?

docs/developer/architecture/logs.rst
gdemonet: Furthermore, these aren't always set. I would still try to get a value for `app.kubernetes.io/name` if available, with a placeholder if label was not set, because we have no other equivalent.

docs/developer/architecture/logs.rst
gdemonet: Definitely required, not only for Solutions (e.g. we add a component, we need a new parser for its log format, what do we do?).

Again, why not consider the Logging Operator? It defines all of these in CRs, making it easy to customize and extend.

gdemonet: Filters are rules to drop log events.
Parsers are basically regular expressions used to extract labels from log events.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+            Retry_Limit     False
+
+.. note::
+
+  Fluent Bit configuration is not a YAML file and looks more like an INI one.
+  CSC is not able to take care of these files and even if we add support for
+  this format, it will be hard to implement the merging with defaults mechanism
+  as blocks can not really be identified, because there is no unique name or ID.
+  Since we only need to provide to the user the possiblity to add a new output,
+  it should be sufficient as we don't need to edit what's deployed by MetalK8s.
```
gdemonet: What about switching between with and without Fluentd?

docs/developer/architecture/logs.rst
gdemonet: So we can't split Loki components into different K8s objects and scale them independently? What about autoscaling?

gdemonet: I don't think we support resizing PVs. Also, for the question of labels, it's really an implementation detail. The design would need to clarify the sizing rule, however (depending on retention + ingestion rate).

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+
+This module must provide necessary methods to show or edit the configuration
+in 2 different ways:
+
+   * simple
+   * advanced
+
+The ``simple`` mode will only display and allow to change some specific
+configuration, such as the retention or the number of replicas, and in an as
+simple as possible manner for the user.
```
gdemonet: "the simplest possible form"

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,444 @@ overview page.
...
+can be kept in only one place.
+
+This module must provide necessary methods to show or edit the configuration
+in 2 different ways:
+
+   * simple
+   * advanced
+
+The ``simple`` mode will only display and allow to change some specific
+configuration, such as the retention or the number of replicas, and in an as
```
gdemonet: I'm still worried about scaling Loki components independently from one another... Would this fit a "simple" mode?

gdemonet (changes requested):
First dump! Good job with this :)

We'll need a bit more details around the Fluentd question before committing on an implementation plan (honestly not sure if it should be optional).

docs/developer/architecture/logs.rst
alexandre-allard: Filters are also used to redirect the inputs in the right parsers, but yes that's basically the idea.

docs/developer/architecture/logs.rst
alexandre-allard: It's written, we need to define this formula.
We could have a basic formula which only make sure that every pods (or at least the daemonset ones & 1 per deployment) in the metalk8s-logging namespace is healthy, what do you think ?

docs/developer/architecture/logs.rst
alexandre-allard: Yes, that's my point, because here we don't really understand interactions when there is multiple Loki instances, for example, or do each collector write to one or multiple distributors ?

docs/developer/architecture/logs.rst
thomasdanan: So that is a very interesting feature and it should be part of the requirements (to be discussed during today's workshop)

docs/developer/architecture/logs.rst
alexandre-allard: Yes, that's "some processing".
Maybe it should be clearer.

docs/developer/architecture/logs.rst
gdemonet: Not sure if really needed in an Overview page, as it's not a critical component of the system. I'd rather have dedicated pages for the various "infra services" we deploy (auth, monitoring, logging).

docs/developer/architecture/logs.rst
alexandre-allard: That's not well written, it does not chunk or format anything, just checking that everything in the stream is valid, it means:
- The metric labels name are formally correct
- The configured max number of labels per metric is respected
- The configured max length of a label name and value is respected
- The timestamp is not older/newer than the configured min/max time range

docs/developer/architecture/logs.rst
alexandre-allard: Yes, it's just there for the reviewers to understand what will be put here.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, I did it on purpose, for consistency, we do it this way in most of the documentation.

docs/developer/architecture/logs.rst
alexandre-allard: Backend would probably be more appropriate.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, I didn't write it because, we say in the document that our focus is on resources consumption.
So, if we want something light, it often comes with less functionalities.
I could still precise it.

docs/developer/architecture/logs.rst
alexandre-allard: Looks like I didn't finish my sentence. :D

docs/developer/architecture/logs.rst
alexandre-allard: Well, I've considered it, but you can't do a simple architecture with fluentbit + loki (without fluentd).
Otherwise it looks like it's pretty good stuff and would ease many things.
But, if we consider the deployment of fluent as not optional, we could use it.

docs/developer/architecture/logs.rst
gdemonet: Suggestion from @xaf-scality was to also consider the whole pipeline: can we use alerting rules to compare the amount of logs seen by collectors with what gets ingested in Loki? Would help us see if something in the pipeline is broken, meaning we'd be losing logs.

docs/developer/architecture/logs.rst
gdemonet: But then the `this document` reference is global to the whole page... Doesn't make sense for this kind of link IMHO.

docs/developer/architecture/logs.rst
gdemonet: Would need empirical confirmation, but as @xaf-scality suggested, we may be able to configure fluentd as passthrough, reducing its impact on resources.

docs/developer/architecture/logs.rst
alexandre-allard: Fluentd has been removed from the design.

docs/developer/architecture/logs.rst
alexandre-allard: Not possible to split components using filesystem as storage backend, which means no autoscaling either because the monolithic pod is stateful thank to the ingester volume.

gdemonet: Can't split and horizontally scale - true
Can't autoscale - debatable: we may want to adjust `memory` and `cpu` consumption from the querier container (through vertical **autoscaling** :wink:)

docs/developer/architecture/logs.rst
alexandre-allard: It's the default behavior of Loki, depending on the configuration and the logging rate, a log could stay in ingester buffer for few minutes before being flushed to the storage backend, so if you do not query the ingester, you will not get the latest logs.
If the ingester is busy, the querier will wait `querier_config.query_timeout` time (1m by default) before timing out.
We can disable this mechanism or at least minimize it, with the option `quierer_config.query_ingesters_within`, which represents the maximum lookback beyond which queries are not sent to ingester (default to 0s, which means it always query ingesters, before falling back to storage).
Maybe we can set it to `-1` to disable it.

docs/developer/architecture/logs.rst
alexandre-allard: It does not seem to be possible, AFAICT you can only have 1 storage backend.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, it's reading the Kubernetes API to enrich the logs with metadata such as container labels.

docs/developer/architecture/logs.rst
alexandre-allard: removed

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,524 @@ overview page.
...
+   +----------------------+------------------+--------------------+
+   | promtail             | 1292             | 1925               |
+   +----------------------+------------------+--------------------+
+   | fluent-bit           | 1040             | 834                |
+   +----------------------+------------------+--------------------+
+   | fluent-bit + fluentd | 1447             | 1902               |
+   +----------------------+------------------+--------------------+
+
+We can see that the Fluent Bit + Loki architecture has the smallest impact
+on resources.
```
TeddyAndrieux: To me, we can also see that fluent-bit + fluentd seems to scale much better than fluent-bit alone (and fluent-bit always have impact on workload where fluentd could be on dedicated infra nodes)
Just saying, I don't say we should go for fluent-bit + fluentd but... interesting (may be worth to mention this somewhere)

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,524 @@ overview page.
...
+Logging Operator
+~~~~~~~~~~~~~~~~
+
+`Logging Operator`_ seemed to be a good candidate for the implementation
+we choose, offering the ability to deploy and configure Fluent Bit and Fluentd,
+but Fluentd is not optional and seems to have a central place as most of the
+parsing/filtering is done by this one, which means a bigger footprint on the
+hardware resources.
+
+.. _Logging Operator: https://github.com/banzaicloud/logging-operator
```
TeddyAndrieux: Minor: Maybe you should move this one just after `Fluent-bit + Fluentd + Loki` section

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,524 @@ overview page.
...
+All the components will be deployed using Kubernetes manifests through Salt
+inside a ``metalk8s-logging`` Namespace.
+
+Fluent Bit
+""""""""""
+
+Fluent Bit will be deployed as a DaemonSet, because we need one collector on
+each node to be able to collect logs from both the Kubernetes platform, the
+applications that run on top of it and all the system daemons running alongside
+(e.g. Salt minion).
```
TeddyAndrieux: Minor: IMO worth mentioning kubelet also here

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,524 @@ overview page.
...
+More details can be found on `Fluent Bit Operator repository`_ and manifest
+samples are
+`here <https://github.com/kubesphere/fluentbit-operator/tree/master/manifests/logging-stack>`_.
+
+.. _Fluent Bit Operator repository: https://github.com/kubesphere/fluentbit-operator
+
+Loki
+""""
+
+Loki's configuration is stored as a Secret, as for Fluent Bit, we need
```
TeddyAndrieux: Minor: Since we use fluent-bit operator it's not "as for Fluent Bit" I think

TeddyAndrieux (dismissed):
Cool, really comprehensive and detailed :+1: , just a few minor comments

docs/developer/architecture/logs-components-diagram.uml
```diff
@@ -0,0 +1,49 @@
+@startuml
+
+node "node1" {
+    journal_logs1 <-- [collector1] : read
```
gdemonet: I'm no expert on diagrams, so maybe worth asking what the common opinion is. But here's my understanding of what arrows convey as meaning: direction associated with the label.

In this example, I understand that there is a `read` action going on, and that it goes **from** `collector1` **to** `journal_logs1`. Does it mean that the information being read travels in this direction? Or does it mean that the action is exerted by `collector1` on the destination `container_logs1`?

My first understanding is that the arrow means the direction of information, and the label is describing what action represents this data transfer. Apparently, your approach is different, assuming the arrow direction is there to describe how the action is applied between components.

If we keep your approach (which is fine), maybe we can make it clearer by changing the labels into `reads from`/`writes to`, so that when reading the diagram we can "read the arrow" from one component to another as if it were a sentence.

gdemonet: What I can say, is that this diagram is already much better for understanding than the previous ASCII one :wink:

docs/developer/architecture/logs.rst
alexandre-allard: True, I can add one or two sentences explaining this.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, exempli gratia is not intented to be exhaustive, but maybe since Kubelet is a really important component, it could make sens to add it here. I'll do it.

docs/developer/architecture/logs.rst
alexandre-allard: Yes, good catch, it's a leftover from the previous version of the design.

docs/developer/architecture/logs.rst
gdemonet: Also, we're only comparing scaling in number of log streams here (which would, in theory, mostly scale with the number of nodes). We also need to consider scaling in ingestion rate (which can happen for uncontrolled reasons like a spike in a workload API usage).

Finally, even though you showed the overall CPU/RAM consumption, there is value in seeing which component consumes what (especially if we want to evaluate the impact of fluent-bit on each node).

This document should clearly state how the proposed solutions compare (maybe ignoring Promtail for limited configurability, which is explained below :ok_hand:), for future readers / users of MetalK8s to understand our choices and future strategy.

docs/developer/architecture/logs.rst
alexandre-allard: Why not.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,524 @@ overview page.
...
+storage will not be that much important.
+Moreover, it eases the deployment and maintenance since there is not an
+extra component.
+
+.. _new experimental feature: https://github.com/grafana/loki/blob/master/docs/operations/storage/filesystem.md#new-and-very-experimental-in-150-horizontal-scaling-of-the-filesystem-store
+
+.. todo::
+
+  define loki storage sizing depending on retention, ingestion rate,
+  average size of logs, number of streams, ...
```
gdemonet, line 443: Yes, but make this `todo` item more than just about storage: we need a sizing rule for CPU/RAM as well. This could be stated next to the numbers report you added above.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+
+ - FluentBit: Defines Fluent Bit instances and its associated config
+ - FluentBitConfig: Select input/filter/output plugins and generates the final
+   config into a Secret
+ - Input: Defines input config sections
+ - Filter: Defines filter config sections
+ - Output: Defines output config sections
+
+For example, if a user wants to forward all Kubernetes logs to an external log
+centralization system (e.g. Elasticisearch), he will need to define an
```
gdemonet:
```suggestion
centralization system (e.g. Elasticsearch), he will need to define an
```

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+
+.. _Fluent Bit Operator repository: https://github.com/kubesphere/fluentbit-operator
+
+Loki
+""""
+
+Loki's configuration is stored as a Secret.
+we need to expose few parameters to the user for customization
+(e.g. retention).
+Since we do not have Operator and CRs for Loki, we will use the CSC mechanism
```
gdemonet, line 507: But, "we do not have Operator" either for fluent-bit, today... Is there a good reason why:

- we need an Operator and CRs for fluent-bit that CSC cannot solve?
- we can simply rely on CSC for Loki configuration?

As explained offline, IMO the design document is here to explain design choices, not to detail the implementation plan. This plan may seem good right now and obsolete in a few months/years - OTOH, if we say "fluent-bit + loki has been chosen because it has a much lower minimal footprint (fixed cost) and allows better scaling in ingestion rate than other architectures", this has value even if new alternatives more powerful than fluent-bit arrived on the market.

It shows how we want to give priority to certain aspects of the software.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+``metalk8s-loki-config`` in the ``metalk8s-logging`` Namespace.
+
+A Salt module will be developed to manipulate this object, so the logic
+can be kept in only one place.
+
+This module must provide necessary methods to show or edit the configuration
+in 2 different ways:
+
+   * simple
+   * advanced
```
gdemonet: The format shown below is `advanced`? If yes, then let's not care about a `simple` mode at all, IMO. As @TeddyAndrieux proposed when we were discussing this on "basic alerting", the simplicity layer can be only implemented in the user interface. Whether the UI is browser-based or terminal-based will change the audience anyway, and as such, what's simple for one isn't for the other.

My suggestion is thus: build the sufficiently advanced configuration layer to get everything working, and see what's needed in the UIs. Of course, using real CRs instead of arbitrary text in ConfigMaps will help to avoid errors when UIs will consume/edit this configuration layer.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+either have 2 distinct dashboards or 2 log panels in the same.
+An example of what we want is `Loki dashboard`_.
+
+.. _Loki dashboard: https://github.com/grafana/loki/blob/master/production/loki-mixin/dashboard-loki-logs.json
+
+.. note::
+
+  The ``grafana_datasource: "1"`` and ``grafana_dashboard: "1"`` labels are
+  what is used by the Prometheus Operator to retrieve datasource and dashboard
+  for Grafana.
```
gdemonet: I guess it also needs to retrieve them from the same namespace as the Operator runs in (`metalk8s-monitoring`)? In any case, could be worth thinking about those the same way we'll think about Solution dashboards/datasources in the future.

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+
+Loki Volume Purge
```
gdemonet, line 667: One really strong message we received from field experts (cc @xaf-scality @bkettler) is that even though the volume is filling up at a dangerous rate, it should never get to the point where it's full.

Monitoring and alerting are paramount for an admin to make sure the system is stable, but a broken component or a DDoS attack over the week-end could result in ingestion increase, which shouldn't break the log system to a point where complex manual operations are required.

Depending on the cause, the logging pipeline could (should?) first pause automatically, and then:

- storage should be adjusted (as you said, by lowering retention or increasing capacity)
- outliers should be detected and silenced / removed from the pipeline (manually or automatically, depends on what we can do)
- ...other ideas?

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+
+Iteration 1
+"""""""""""
+
+The goal is to have a working log centralization system, with logs accessible
+from Grafana:
+
+* Deploy Fluent Bit and Loki
+* Deploy Grafana datasource & dashboard
+* Deploy Loki volumes purge mechanism
```
gdemonet: ? Which mechanism?

docs/developer/architecture/logs.rst
```diff
@@ -209,7 +730,19 @@ the workload properties.
 The Post Installation page is updated to indicate that persistent storage is
 needed for log centralization service.
 
-A new page should be added to explains how to operate the service.
+A new page should be added to explains how to operate the service and how to
```
gdemonet:
```suggestion
A new page should be added to explain how to operate the service and how to
```

docs/developer/architecture/logs.rst
```diff
@@ -197,6 +197,527 @@ overview page.
...
+Iteration 1
+"""""""""""
+
+The goal is to have a working log centralization system, with logs accessible
+from Grafana:
+
+* Deploy Fluent Bit and Loki
+* Deploy Grafana datasource & dashboard
+* Deploy Loki volumes purge mechanism
+* Customization of Loki with CSC mechanisms
```
gdemonet, line 708: IMO, first iteration should be minimal (just as Dex currently is). We gather the required features from other squads and focus on this. Future iterations will add fields to the available CSC(s).

gdemonet (dismissed):
Some details to polish, but indeed, a really great design doc :clap: :clap:!
I just wanted to make sure we don't do too much for the first iteration, which we'd really like to get as small as possible.

docs/developer/architecture/logs-components-diagram.uml
alexandre-allard: I'm not expert either, I asked myself for the direction of the arrow (my understanding is from where comes the action, e.g. a client querying a database is client -> db, but I know you can also represent that with client <-> db, because the client is asking and the db is returning the data), I wasn't sure how to do it, that's why I added the "read" for it to be clear.
I like the `reads from` / `writes to`, it's even clearer.

docs/developer/architecture/logs.rst
alexandre-allard: No no, it's Elastickysearch... :D

docs/developer/architecture/logs.rst
alexandre-allard: No mechanism now, it's a leftover.
It's more about having dedicated alert rules for Loki volume usage, I'll update this.
But maybe we still want to have something, depending on what we decide regarding your comment above (https://github.com/scality/metalk8s/pull/2668#discussion_r464871732)

docs/developer/architecture/logs.rst
alexandre-allard: I don't think we can go to a point where complex manual operations would be required if the storage is filled up, we just need to manually purge it (and probably restard loki pod).

First, for the pause in the log ingestion, I think it's already made this way in fluent-bit.
If an output plugin is unable to send data to a backend, it'll keep the data in memory and if the buffer is full, then it tells the input plugin to pause, see https://fluentbit.io/documentation/0.13/configuration/backpressure.html for more details.

For the storage resizing, we can't do much except providing some alerting (with prediction and also real time monitoring), documentation (through runbook for example) and good sizing rules.

For the outliers... I'm not sure how to achieve this automatically, it means we need to be able to analyze log flow and detect patterns... I'm not sure that's what we want in our simple logging stack.
To me, the pause + alert + documentation should be enough.
Maybe we can also provide some documentation to explain how to do basic filtering to be sure unwanted logs are dropped at collector level.

Otherwise, we could add an extra option in Loki configuration that would allow us to do purging based on size, which could be done with 2 options (to ensure we will not remove all logs if there is a crazy amount of logs):
- % of space used: we purge oldest logs if space is greater than a certain threshold 
- minimum days of retention: we purge logs older than minimum days of retention
It means, for example, you can have a retention of 2 weeks of logs, with a minimum of 3 days and if it's still not sufficient, fluent-bit will pause the flow.
I just think that if we go with automatic purge based on size, we need to be careful and trigger alerts each time we purge, otherwise operators will think they have 2 weeks of logs and will never be aware that there is a sizing issue on their platform.

docs/developer/architecture/logs.rst
alexandre-allard: I'm not sure they need to be in `metalk8s-monitoring` namespace, I think the Operator has cluster wide rights allowing it to watch/read ConfigMaps.
But I'm not super familiar with ClusterRole and stuff, so maybe I'm misunderstanding something.

docs/developer/architecture/logs.rst
alexandre-allard: CSC is not as powerful as an Operator (it could if we invest time on it, but from my understanding, we want to have real Operators when it's possible, which allow to watch and reconciliate in real time rather than having to launch salt state after each change).
For Loki, I don't think that's a problem to use CSC, since its configuration should not change that much and should only be edited by the MetalK8s cluster administrator.
But, for Fluent Bit, it's not the same, for example, we may want Solutions to be able to provide CRs to define their own filtering/parsing rules.
I agree, I should probably not talk about the implementation, but from my point of view (which is probably wrong), I think giving some explanations on how we may want to do that allows people to better understand the idea (and we're in implementation details section... what should I write here ?).

gdemonet: > CSC is not as powerful as an Operator (it could if we invest time on it, but from my understanding, we want to have real Operators when it's possible, which allow to watch and reconciliate in real time rather than having to launch salt state after each change).

That's granted :)

> For Loki, I don't think that's a problem to use CSC, since its configuration should not change that much and should only be edited by the MetalK8s cluster administrator.
> But, for Fluent Bit, it's not the same, for example, we may want Solutions to be able to provide CRs to define their own filtering/parsing rules.

Exactly. That's what we should see in the design.

> I agree, I should probably not talk about the implementation, but from my point of view (which is probably wrong), I think giving some explanations on how we may want to do that allows people to better understand the idea (and we're in implementation details section... what should I write here ?).

Of course, you're right to discuss implementation. What's most important in this document is not the implementation itself however, but what it implies: how it fits the requirements, why this one is chosen over another one, what are its shortcomings, what are the unknowns.

docs/developer/architecture/logs.rst
gdemonet: In this case, it would make sense to store logging-related dashboards/datasources in `metalk8s-logging`

alexandre-allard: Tested, it does not seem to work if outside of the `metalk8s-monitoring` namespace, at least for datasource (it may not be the case for dashboards, but I didn't find a clear answer, I should test it).

docs/developer/architecture/logs.rst
alexandre-allard: Finally added it back as it does not cost much to do (and it's already done :p).

thomasdanan (dismissed):
Excellent document !!!

- For the benchmark, next time it could be interesting to have several more data points to see how figures evolve with the load.
- I remember some discussion on how solution deployment could configure parsers or stuff like that, may be to consider in next req/design iteration

alexandre-allard:
@thomasdanan thanks for the feedback.
Ok for the first point, we will do this next time (actually we've way more points since it's extracted and computed from Prometheus data, but I wanted to keep it simple).
Regarding you second point, it'll be tackled as soon as we've the Fluent-bit operator ready, we didn't want to invest time on something that'll be removed soon.

thomasdanan (approved):

PR #2670, opened 2020-07-16
salt/tests: Add unit tests for `metalk8s_volumes` salt module

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

#2266 

**Summary**:

Add unit tests for `metalk8s_volumes` salt custom module, use a
dedicated yaml file containing input, arguments
and expected result

---

Refs: #2266

### DIFF
--- a/salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
@@ -1,0 +1,499 @@
+_volumes_details: &volumes_details
+  my-sparse-volume:
+    apiVersion: storage.metalk8s.scality.com/v1alpha1
+    kind: Volume
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/last-applied-configuration: '{"apiVersion":"storage.metalk8s.scality.com/v1alpha1","kind":"Volume","metadata":{"annotations":{},"name":"my-sparse-volume"},"spec":{"nodeName":"bootstrap","sparseLoopDevice":{"size":"1Gi"},"storageClassName":"metalk8s-prometheus","template":{"metadata":{"labels":{"app.kubernetes.io/name":"prometheus-operator-alertmanager"}}}}}
+
+'
+      creationTimestamp: '2020-06-12T13:33:44Z'
+      finalizers:
+      - storage.metalk8s.scality.com/volume-protection
+      generation: 2
+      name: my-sparse-volume
+      resourceVersion: '3106735'
+      selfLink: /apis/storage.metalk8s.scality.com/v1alpha1/volumes/my-sparse-volume
+      uid: f1d78810-3787-4ca4-b712-50a269e42560
+    spec:
+      nodeName: bootstrap
+      sparseLoopDevice:
+        size: 1Gi
+      storageClass:
+        allow_volume_expansion: null
+        allowed_topologies: null
+        api_version: null
+        kind: null
+        metadata:
+          annotations: null
+          cluster_name: null
+          creation_timestamp: '2020-06-12T13:30:34+00:00'
+          deletion_grace_period_seconds: null
+          deletion_timestamp: null
+          finalizers: null
+          generate_name: null
+          generation: null
+          initializers: null
+          labels:
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/part-of: metalk8s
+            heritage: salt
+            metalk8s.scality.com/version: 2.5.1-dev
+          managed_fields: null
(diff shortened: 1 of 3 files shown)


salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
```diff
@@ -0,0 +1,499 @@
...
+    pillar_volumes: *volumes_details
+    is_blkdev: False
+    result: False
+
+  ## Invalid volumes
+  # specified volume is not in the pillar
+  - name: unknown-volume
+    pillar_volumes: *volumes_details
+    raises: True
+    result: volume unknown-volume not found in pillar
```
gdemonet, line 244: Could be nice to use the same pattern to pass error messages (here we do `raises(bool)` + `result(str)`, in create we do `raise_msg(str)` + no result)

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
```diff
@@ -0,0 +1,499 @@
...
+    pillar_volumes: *volumes_details
+    ftruncate: False
+    raise_msg: "cannot create sparse file at .*: An error has occurred"
+
+  ## RAW BLOCK DEVICE volume
+  # create on raw block device (nothing to create)
+  - name: my-raw-block-device-volume
+    pillar_volumes: *volumes_details
+    # We do not need to create block device
+    raise_msg: block device /dev/sda1 does not exists
```
gdemonet:
```suggestion
    raise_msg: block device /dev/sda1 does not exists
```

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
```diff
@@ -0,0 +1,499 @@
...
+  ## Invalid volumes
+  # specified volume is not in the pillar
+  - name: unknown-volume
+    pillar_volumes: *volumes_details
+    raise_msg: volume unknown-volume not found in pillar
+
+  # empty pillar
+  - name: my-invalid-type-volume
+    pillar_volumes: *volumes_details
+    raise_msg: unsupported Volume type for Volume my-invalid-type-volume
```
gdemonet:
```suggestion
    raise_msg: unsupported Volume type for Volume my-invalid-type-volume
```

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
```diff
@@ -0,0 +1,499 @@
...
+  # sparse file does not exists
+  - name: my-sparse-volume
+    pillar_volumes: *volumes_details
+    remove_error: [2, "No such file or directory", "..."]
+
+  # error when trying to remove the sparse file
+  - name: my-sparse-volume
+    pillar_volumes: *volumes_details
+    remove_error: "An error has occurred during remove"
+    raise_msg: "An error has occurred during remove"
```
gdemonet, line 482: Not entirely certain from reading these example errors: is the `clean_up` method copying the error message from `os.remove` (or below, from `fcntl.ioctl`), or are we using a generic message which matches what's above? If the former, maybe it'd be nice to put some random word in the message, to make it obvious (e.g. "A wild spaghetti monster appeared!"), but if the latter, maybe we should show different messages between the actual error and what gets raised (sorry, didn't check the code).

gdemonet (dismissed):
Minor comments, otherwise looks good to me (as long as I have enough screen space to look at tests alongside the YAML file :wink:).

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
TeddyAndrieux: Right, but some other unit tests already merged get the same approach

- `raise_msg` when function tested do not return anything by default but may raise
- `result` + `raises` when function return something and may raise also

So ... maybe let's keep it for the moment

gdemonet: Ah OK, fine by me then :+1:

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
TeddyAndrieux: Right, right now I just put `An error has occurred` everywhere when I wanted to raise something but indeed it could be clearer we some random word. 
And indeed here we use the message raised by `os.remove` (and `fcntl.ioctl`)

(But here we have the `during remove` part that make sure that we really raise this one, but indeed you don't know if it's part of the code or no just by looking the yaml)

gdemonet (approved):

Issue #1999, opened 2019-11-05
Pre Checks


thomasdanan:
Generic pre Install checks prior to deploy metalK8s (servers, disks, partitions, os, network, etc…)

- Check Network configuration (netmask, interface ...)
- security ? (SELinux)
- memory size
- partitions: /var/log, /var/lib, swapoff, # core, firewall already there

Value: Avoid failing in easy to solve or to detect traps in order to increase usability and confidence during initial deployment phase.
To be executed by the customer, prior to run the full installation.

- [ ] https://github.com/scality/metalk8s/issues/241
- [ ] https://github.com/scality/metalk8s/issues/46

PR #2677, opened 2020-07-23
Make calico MTU configurable in bootstrap config

### LINKED ISSUE #1095: Handle networking MTUs
We currently hard-code the `veth` MTU to `1440` (the Calico self-hosted manifest default). However, this is a 'pessimistic' default, based on GCP deployment infrastructure.

Instead, since we don't enable IPIP by default, we should

- Ensure the MTU as configured in the CNI configuration file is equal to the MTU of the workload-plane interface on the node, I believe, and
- Set the MTU of the created IPIP tunnels to either a configurable value, or the MTU of the workload-plane interface minus the IPIP header size, i.e. 20 bytes

This would at least allow for 60bytes extra per packet on 'regular' ethernet networks, or 7560 bytes on networks configured with jumbo-frames.

TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**Context**: 

#1095 

**Summary**:

- Allow multiple CIDRs for workload plane and control plane networks
- Make calico MTU configurable through bootstrap config
- Check MTU defined is smaller than the one on the local workload network interface

New bootstrap config network layout (still backward compatible with old style)

```
apiVersion: metalk8s.scality.com/v1alpha3
networks:
  workloadPlane:
    cidr:
      - 10.200.10.0/24
      - 10.200.20.0/24
    mtu: 4242
  controlPlane:
    cidr: 10.200.30.0/24
```

---
TODO:

- [x] Add unit test for `get_mtu_from_ip`
- [x] Add unit test for `get_ip_from_cidrs`
- [x] Update documentation for BootstrapConfig changes
- [x] Update tests (and vagrant) to use the new BootstrapConfig style
- [x] Add documentation about MTU configuration in bootstrap config
- [x] Add changelog entry 
 
---

Fixes: #1095

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 - Enable `seccomp` support in `containerd`
   (Issue [#2259](https://github.com/scality/metalk8s/issues/2259),
   PR [#2369](https://github.com/scality/metalk8s/pull/2369))
+- [#1095](https://github.com/scality/metalk8s/issues/1095) - Ability to use
+  multiple CIDRs for control plane and workload plane networks and to specify
+  the workload plane MTU to compute the MTU used by Calico
+  (PR [#2677](https://github.com/scality/metalk8s/pull/2677))
 
 ### Enhancements
 - [#2674](https://github.com/scality/metalk8s/issues/2674) - Bump K8S version
--- a/docs/installation/advanced-guide.rst
+++ b/docs/installation/advanced-guide.rst
@@ -1,0 +1,36 @@
+Advanced guide
+==============
+
+.. _multiple CIDR network:
+
+Multiple CIDRs network
+----------------------
+
+In the Bootstrap Configuration it's possible to provide several CIDRs for a
+single network, it's needed when several nodes does not sit in the same
+network.
+
+.. code-block:: yaml
+
+     networks:
+       controlPlane:
+         cidr:
+           - 10.100.1.0/28
+           - 10.200.1.0/28
+       workloadPlane:
+         cidr:
+           - 10.100.2.0/28
+           - 10.200.2.0/28
+
+This kind of deployment needs good knowledge about networking, as each workload
+node needs to be able to communicate with all others, even those in a different
+workload CIDR.
+
+In this case :ref:`IP-in-IP encapsulation<enable IP-in-IP>` is likely needed.
+
+Some explanation can be found about this subject in
+`Calico documentation <https://docs.projectcalico.org/networking/vxlan-ipip>`_.
+
+.. todo::
+
+  - Explain more what is needed to not need IP-in-IP in that kind of scenario
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -35,11 +35,13 @@ Configuration
 
    .. code-block:: yaml
 
-      apiVersion: metalk8s.scality.com/v1alpha2
+      apiVersion: metalk8s.scality.com/v1alpha3
       kind: BootstrapConfiguration
       networks:
-        controlPlane: <CIDR-notation>
-        workloadPlane: <CIDR-notation>
+        controlPlane:
+          cidr: <CIDR-notation>
+        workloadPlane:
+          cidr: <CIDR-notation>
         pods: <CIDR-notation>
         services: <CIDR-notation>
       proxies:
@@ -60,11 +62,20 @@ notation for it's various subfields.
       These values specify the range of IP addresses that will be used at the
       host level for each member of the cluster.
 
+
+      .. info::
+
+        Several CIDRs can be provided if all nodes do not sit in the same
+        network. This is an :ref:`advanced configuration<multiple CIDR network>`
+        which we do not recommend for non-experts.
+
       .. code-block:: yaml
 
             networks:
-              controlPlane: 10.200.1.0/28
-              workloadPlane: 10.200.1.0/28
+              controlPlane:
+                cidr: 10.200.1.0/28
+              workloadPlane:
+                cidr: 10.200.1.0/28
 
       All nodes within the cluster **must** connect to both the control plane
       and workload plane networks. If the same network range is chosen for both
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -42,6 +42,7 @@ Configuration
           cidr: <CIDR-notation>
         workloadPlane:
           cidr: <CIDR-notation>
+          mtu: <network-MTU>
         pods: <CIDR-notation>
         services: <CIDR-notation>
       proxies:
@@ -69,6 +70,11 @@ notation for it's various subfields.
         network. This is an :ref:`advanced configuration<multiple CIDR network>`
         which we do not recommend for non-experts.
 
+      For ``workloadPlane`` entry an
+      `MTU <https://en.wikipedia.org/wiki/Maximum_transmission_unit>`_ can
+      also be provided, this MTU value should be the lowest MTU value accross
+      all the workload plane network. The default value for this MTU is 1460.
+
       .. code-block:: yaml
 
             networks:
@@ -76,6 +82,7 @@ notation for it's various subfields.
                 cidr: 10.200.1.0/28
               workloadPlane:
                 cidr: 10.200.1.0/28
+                mtu: 1500
 
       All nodes within the cluster **must** connect to both the control plane
       and workload plane networks. If the same network range is chosen for both
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -79,3 +79,30 @@ def get_oidc_service_ip():
     range.
     '''
     return _pick_nth_service_ip(OIDC_ADDRESS_NUMBER)
+
+
+def get_ip_from_cidrs(cidrs, current_ip=None):
+    '''
+    Return the first IP available
+
+    A current_ip can be given so that if the current_ip is already part of
+    one cidr we keep this IP (otherwise we return the first one)
+    '''
+    first_ip = None
+
+    for cidr in cidrs:
+        ip_addrs = __salt__["network.ip_addrs"](cidr=cidr)
+        if current_ip and current_ip in ip_addrs:
+            # Current IP still valid, return it
+            return current_ip
+
+        if ip_addrs and first_ip is None:
+            first_ip = ip_addrs[0]
+
+    if not first_ip:
+        raise CommandExecutionError(
+            "Unable to find an IP on this host in one of this cidr: {}"
+            .format(', '.join(cidrs))
+        )
+
+    return first_ip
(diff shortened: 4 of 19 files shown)


salt/_modules/metalk8s_network.py
```diff
@@ -79,3 +79,26 @@ def get_oidc_service_ip():
     range.
     '''
     return _pick_nth_service_ip(OIDC_ADDRESS_NUMBER)
+
+
+def get_mtu_from_ip(ip):
```
slaperche-scality, line 111: IP or CIDR?

Reading the [doc](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.network.html#salt.modules.network.ifacestartswith), I would say CIDR.

TeddyAndrieux: Not really the doc is wrong here it's not a cidr but an IP (maybe not full)

(And btw it's a bit bugged if you not specify the full IP because he consider 0 as wildcard)

```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether fa:16:3e:2b:fd:d5 brd ff:ff:ff:ff:ff:ff
    inet 10.200.2.69/16 brd 10.200.255.255 scope global dynamic eth0
       valid_lft 71176sec preferred_lft 71176sec
    inet6 fe80::f816:3eff:fe2b:fdd5/64 scope link 
       valid_lft forever preferred_lft forever
3: tunl0@NONE: <NOARP,UP,LOWER_UP> mtu 1440 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
    inet 10.233.132.64/32 brd 10.233.132.64 scope global tunl0
       valid_lft forever preferred_lft forever

[root@bootstrap ~]# salt-call network.ifacestartswith 10.200.0.0/16
local:
[root@bootstrap ~]# salt-call network.ifacestartswith 10.200
local:
    - tunl0
    - eth0
[root@bootstrap ~]# salt-call network.ifacestartswith 10.200.
local:
    - eth0
```

salt/metalk8s/kubernetes/cni/calico/configured.sls
```diff
@@ -3,6 +3,13 @@
 
 {%- set kubernetes_service_ip = salt.metalk8s_network.get_kubernetes_service_ip() %}
 
+{#- Check that service MTU configured is inferior to the local workload interface one #}
+{%- set workload_local_mtu = salt.metalk8s_network.get_mtu_from_ip(grains.metalk8s.workload_plane_ip) %}
+{%- if pillar['networks']['mtu'] > workload_local_mtu %}
+  {{ raise('Trying to configure CNI with ' ~ pillar['metalk8s']['networks']['mtu']
+           ~ 'MTU but local workload interface MTU is inferior ' ~ workload_local_mtu) }}
+{%- endif %}
```
gdemonet: Neat! Do you think we can add a similar check earlier in the bootstrap/`deploy_node` orchestrate?

Maybe consider adding these checks in `metalk8s.internal.preflight`?

TeddyAndrieux: Yes I could add check earlier (maybe not in the preflight because to me preflight should be removed at some point, since it does not bring any value)

I will try to find a better place earlier I agree it fail a bit late here (at least for bootstrap because for new node it will fail when rendering the "bring new node to highstate" so not really late I think)

salt/_pillar/metalk8s.py
```diff
@@ -70,6 +71,7 @@ def _load_networks(config_data):
         'workload_plane': networks_data['workloadPlane'],
         'pod': networks_data.get('pods', DEFAULT_POD_NETWORK),
         'service': networks_data.get('services', DEFAULT_SERVICE_NETWORK),
+        'mtu': networks_data.get('mtu', DEFAULT_MTU)
```
NicolasT: I don't think this is at the right 'location' in the config. It's not something *next to* the pod, service and other network CIDRs. It is *the MTU of the workload plane network*, so it should go in there. Indeed, that means bumping the version of the config format.

salt/metalk8s/kubernetes/cni/calico/configured.sls
```diff
@@ -3,6 +3,13 @@
 
 {%- set kubernetes_service_ip = salt.metalk8s_network.get_kubernetes_service_ip() %}
 
+{#- Check that service MTU configured is inferior to the local workload interface one #}
+{%- set workload_local_mtu = salt.metalk8s_network.get_mtu_from_ip(grains.metalk8s.workload_plane_ip) %}
+{%- if pillar['networks']['mtu'] > workload_local_mtu %}
+  {{ raise('Trying to configure CNI with ' ~ pillar['metalk8s']['networks']['mtu']
+           ~ 'MTU but local workload interface MTU is inferior ' ~ workload_local_mtu) }}
```
NicolasT: "smaller", not "inferior".

salt/_pillar/metalk8s.py
TeddyAndrieux: Right, as discussed, I will update it

salt/_modules/metalk8s_network.py
```diff
@@ -79,3 +79,26 @@ def get_oidc_service_ip():
...
+    if not ifaces:
+        raise CommandExecutionError(
+            'Unable to get interface for "{}"'.format(ip)
+        )
+
+    iface = ifaces[0]
+
+    if len(ifaces) > 1:
+        log.warning(
+            'Several interfaces has "%s" as IP, %s will be used: %s',
```
gdemonet:
```suggestion
            'Several interfaces match the IP "%s", %s will be used: %s',
```

salt/_pillar/metalk8s.py
```diff
@@ -15,6 +15,7 @@
 
 DEFAULT_POD_NETWORK = '10.233.0.0/16'
 DEFAULT_SERVICE_NETWORK = '10.96.0.0/12'
+DEFAULT_MTU = 1460
```
gdemonet, line 18: That's the default Calico uses based on GCP. Should we keep it (to smooth upgrades), or find one that better suits our use cases? (for on-prem, I don't what the lowest reasonable MTU we'd find, but maybe we could start by considering the ones on Scality Cloud - 1500 apparently)

tests/install/steps/test_expansion.py
```diff
@@ -90,15 +89,6 @@ def _check_node_status():
 # }}}
 # Helpers {{{
 
-def get_node_ip(node_name, ssh_config, bootstrap_config):
```
gdemonet, line 93: As discussed privately, this helper is here to get _another_ node's control plane IP, not the bootstrap one. So we can't just replace it with the `control_plane_ip` fixture, unfortunately :confused: 

Also, since we need it before installing Salt on this node, we cannot use the same approach (reading `grains`). Maybe we could still use Salt-SSH and the new `metalk8s_network.get_ip_from_cidrs` method, to avoid duplication of logic, but keeping the old one should also work, so that's your call :)

gdemonet (commented):
Some minor comments, but overall behaviour and user-facing features LGTM!

salt/metalk8s/kubernetes/cni/calico/configured.sls
gdemonet: Could be done later, and as you said, since we may want to do some refactoring around the pre-checks, would make sense to add as a task in #1999

tests/install/steps/test_expansion.py
TeddyAndrieux: A bit overkill to use salt-ssh to retrieve IP we will use for node (so for salt-ssh) to me.
Anyway I just consider that we have only 1 CIDR for control-plane network in our tests (which is acceptable I think, at least for the moment)

salt/_pillar/metalk8s.py
TeddyAndrieux: Don't really know, but to me it's not really part of this PR, but indeed it's a good question

gdemonet: Let's keep it for now, shouldn't hurt unless we start deploying in envs where this default value is too low.

gdemonet (dismissed):
LGTM!

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 - Enable `seccomp` support in `containerd`
   (Issue [#2259](https://github.com/scality/metalk8s/issues/2259),
   PR [#2369](https://github.com/scality/metalk8s/pull/2369))
+- [#1095](https://github.com/scality/metalk8s/issues/1095) - Ability to use
+  multiple CIDRs for control plane and workload plane networks and to configure
+  the MTU used by Calico (PR [#2677](https://github.com/scality/metalk8s/pull/2677))
```
gdemonet: Maybe specify "the workload plane MTU" as you did in docs?

gdemonet (dismissed):
Perfect! Thanks, I guess we just need to create some tracking tickets for open questions:
- how to get the MTU checks "standalone" so we can run them earlier
- what default MTU should we use
- what's the impact of having multiple CIDRs (manual + automated tests to consider)

No need to provide much details, just get a title, link to this PR, and the `state:question` label and we'll get back to them later.

NicolasT:
Something to keep in mind: when multiple CIDRs *are* used, you *almost* certainly require tunneling (IPIP/VXLAN) to be enabled! (Unless Calico is configured to peer with routers on the data path etc. which is rather unlikely in our deployments).

docs/installation/advanced-guide.rst
```diff
@@ -0,0 +1,36 @@
...
+           - 10.100.1.0/28
+           - 10.200.1.0/28
+       workloadPlane:
+         cidr:
+           - 10.100.2.0/28
+           - 10.200.2.0/28
+
+This kind of deployment need good knowledge about networking, as all workload
+node need to be able to communicate with all workload node even those that does
+not sit in the same CIDRs.
```
gdemonet:
```suggestion
This kind of deployment needs good knowledge about networking, as each workload
node needs to be able to communicate with all others, even those in a different
workload CIDR.
```

docs/installation/bootstrap.rst
```diff
@@ -60,11 +63,26 @@ notation for it's various subfields.
       These values specify the range of IP addresses that will be used at the
       host level for each member of the cluster.
 
+
+      .. warning::
```
gdemonet: Maybe let this just be an `info` admonition?

docs/installation/bootstrap.rst
```diff
@@ -60,11 +63,26 @@ notation for it's various subfields.
...
+
+        Several CIDR can be provided if all nodes does not sit in the same
+        network but it's an
+        :ref:`advanced configuration<multiple CIDR network>`.
```
```suggestion
        Several CIDRs can be provided if all nodes do not sit in the same
        network. This is an ref:`advanced procedure<multiple CIDR network>`
        which we do not recommend for non-experts.
```

gdemonet (changes requested):
Minor wording comments, and we should be good to go!

gdemonet (approved):

PR #2691, opened 2020-07-30
salt/tests: Add unit tests for `metalk8s_solutions` salt module

alexandre-allard:
**Component**: salt, tests

**Context**: 

**Summary**:
Add unit tests for `metalk8s_solutions` salt custom module, use a dedicated yaml file containing input, arguments and expected result

**Acceptance criteria**:
unittest are green and coverage for `metalk8s_solutions` module is 100%

---

See: #2266

### DIFF
--- a/salt/tests/unit/modules/files/test_metalk8s_solutions.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_solutions.yaml
@@ -1,0 +1,298 @@
+read_config:
+  # Ok - no solutions configuration file, create set to True
+  - create: True
+    result:
+      kind: SolutionsConfiguration
+      apiVersion: solutions.metalk8s.scality.com/v1alpha1
+      archives: []
+      active: {}
+  # Ok - existing configuration file
+  - config: |-
+      kind: SolutionsConfiguration
+      apiVersion: solutions.metalk8s.scality.com/v1alpha1
+      archives:
+        - /tmp/my-solution.iso
+      active:
+        my-solution: 1.0.0
+    result:
+      kind: SolutionsConfiguration
+      apiVersion: solutions.metalk8s.scality.com/v1alpha1
+      archives:
+        - /tmp/my-solution.iso
+      active:
+        my-solution: 1.0.0
+  # Nok - no solutions configuration file, create left to default (False)
+  - raises: True
+    result: 'Failed to load ".*": .*'
+  # Nok - no solutions configuration kind
+  - raises: True
+    config: |-
+      apiVersion: solutions.metalk8s.scality.com/v1alpha1
+    result: >-
+      Invalid `kind` in configuration \(None\), must be
+      "SolutionsConfiguration"
+  # Nok - invalid solutions configuration kind
+  - raises: True
+    config: |-
+      kind: InvalidSolutionsConfiguration
+      apiVersion: solutions.metalk8s.scality.com/v1alpha1
+    result: >-
+      Invalid `kind` in configuration \(InvalidSolutionsConfiguration\),
+      must be "SolutionsConfiguration"
+  # Nok - no solutions configuration apiVersion
+  - raises: True
+    config: |-
+      kind: SolutionsConfiguration
+    result: >-
+      Invalid `apiVersion` in configuration \(None\), must be one of: .*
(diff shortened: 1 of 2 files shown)


salt/tests/unit/modules/test_metalk8s_solutions.py
```diff
@@ -0,0 +1,338 @@
...
+    """
+    TestCase for `metalk8s_solutions` module
+    """
+    loader_module = metalk8s_solutions
+
+    def test_virtual_success(self):
+        """
+        Tests the return of `__virtual__` function, success
+        """
+        dict_patch = {'metalk8s.archive_info_from_iso': True}
```
TeddyAndrieux: Minor: To me, you should put a `MagicMock` instead of `True` since it should be callable in this salt dict (even if it's not important here)
```suggestion
        dict_patch = {'metalk8s.archive_info_from_iso': MagicMock()}
```

salt/tests/unit/modules/test_metalk8s_solutions.py
```diff
@@ -0,0 +1,338 @@
...
+        open_mock = mock_open(read_data=config)
+        handle = open_mock.return_value
+
+        def _open_mock(*_):
+            if config:
+                return handle
+            raise IOError(errno.ENOENT, "No such file or directory")
+
+        open_mock.return_value = None
+        open_mock.side_effect = _open_mock
```
TeddyAndrieux: To me it's a bit strange mock stuck, 
You should just set the IOError when empty config and that should be enough no ?

```suggestion
        open_mock = mock_open(read_data=config)
        if not config:
            open_mock.side_effect = IOError(errno.ENOENT, "No such file or directory")
```

salt/tests/unit/modules/test_metalk8s_solutions.py
```diff
@@ -0,0 +1,338 @@
...
+                raise Exception("Something bad happened! :/")
+            config = data
+
+        list_available_mock = MagicMock(return_value=available or {})
+        read_config_mock = MagicMock(return_value=config)
+        yaml_safe_dump_mock = MagicMock(side_effect=_yaml_safe_dump_mock)
+
+        with patch("metalk8s_solutions.list_available", list_available_mock), \
+                patch("metalk8s_solutions.read_config", read_config_mock), \
+                patch("__builtin__.open", mock_open()), \
```
TeddyAndrieux: To me you should mock open only in the module
```suggestion
                patch("metalk8s_solutions.open", mock_open()), \
```

salt/tests/unit/modules/files/test_metalk8s_solutions.yaml
```diff
@@ -0,0 +1,298 @@
...
+  # Ok - solution with no version provided (default to latest)
+  - solution: my-solution
+    available:
+      my-solution:
+        - version: 1.0.0
+    config:
+      active: {}
+    result:
+      active:
+        my-solution: latest
```
TeddyAndrieux, line 116: Wow, it looks like a wrong behavior from the module here, it's should be the latest available version at the moment we call this function, no ? (Not related to this PR anyway)

TeddyAndrieux (commented):
LGTM, with some minor comments (2.5 not reviewed yet, I'm on it)

salt/tests/unit/modules/test_metalk8s_solutions.py
alexandre-allard: Yeah, I agree, it's a bit over engineered for nothing... :D
I don't know what I had in mind at this moment.

salt/tests/unit/modules/test_metalk8s_solutions.py
alexandre-allard: True, it's a leftover.

salt/tests/unit/modules/files/test_metalk8s_solutions.yaml
alexandre-allard: Not really, it means we always want the latest version.
This is then computed in the Salt formula to pick up the latest available version.
It probably should not be used in production.

TeddyAndrieux (approved):

PR #2692, opened 2020-07-30
salt/tests: Add unit tests for block volumes from `metalk8s_volumes`

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

Metalk8s_volumes for development/2.5 +

#2266 

**Summary**:

Add unit tests for SparseLoopDeviceBlock volumes and RawBlockDeviceBlock
(disk, partition and LVM) volumes from `metalk8s_volumes` salt custom
module, also add unit tests for new `device_name` and `device_info`
functions

```
---------- coverage: platform linux2, python 2.7.18-final-0 ----------
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
salt/_modules/metalk8s_volumes.py     264      0   100%
```

---

Refs: #2266

### DIFF
--- a/salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
@@ -137,6 +137,105 @@ _volumes_details: &volumes_details
         lastUpdateTime: '2020-07-11T13:55:26Z'
         status: 'True'
         type: Ready
+  my-sparse-block-volume:
+    apiVersion: storage.metalk8s.scality.com/v1alpha1
+    kind: Volume
+    metadata:
+      name: my-sparse-block-volume
+      uid: 8474cda7-0dbe-40fc-9842-3cb0404a725a
+    spec:
+      nodeName: bootstrap
+      storageClass:
+        metadata:
+          name: metalk8s-monitoring
+        mount_options:
+        - rw
+        - discard
+        parameters:
+          fsType: ext4
+          mkfsOptions: '["-m", "0"]'
+        provisioner: kubernetes.io/no-provisioner
+        reclaim_policy: Retain
+        volume_binding_mode: WaitForFirstConsumer
+      storageClassName: metalk8s-monitoring
+      mode: Block
+      sparseLoopDevice:
+        size: 1Gi
+  my-raw-block-device-block-disk-volume:
+    apiVersion: storage.metalk8s.scality.com/v1alpha1
+    kind: Volume
+    metadata:
+      name: my-raw-block-device-block-disk-volume
+      uid: 9474cda7-0dbe-40fc-9842-3cb0404a725a
+    spec:
+      nodeName: bootstrap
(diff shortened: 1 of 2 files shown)


salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
```diff
@@ -260,13 +415,45 @@ create:
...
     pillar_volumes: *volumes_details
     # We do not need to create block device
     raise_msg: block device /dev/sda1 does not exists
 
+  ## RAW BLOCK DEVICE block disk volume
+  # create on raw block device (nothing to create)
+  - name: my-raw-block-device-block-disk-volume
+    pillar_volumes: *volumes_details
+    # We do not need to create block device
+    raise_msg: block device /dev/sdb does not exists
```
gdemonet, line 441: Not related to the test, but something from the original module (I just realized by reading your test cases :slightly_smiling_face:), this error message is strange, no? Shouldn't it say something like "can not create raw block devices"?

gdemonet (approved):
LGTM :ok_hand:

salt/tests/unit/modules/files/test_metalk8s_volumes.yaml
TeddyAndrieux: Hmm maybe, the thing is just that the user should not call this function for raw block device.

But indeed the message is not really clear about it, 
it's maybe more a question for @slaperche-scality

slaperche-scality: The reason behind `'block device {} does not exists'` is because if we ends up here, it means that `exists()` returned `False` (hence the `does not exists` error). Which can be a bit surprising when reading the code, but gives a good error message from Salt.

`"can not create raw block devices"` isn't really good, because we don't even try to create it.

gdemonet: Ah right, I get it. So the test for `create()` on block device should try with an existing one (which would succeed) and a missing one (which would fail, because we can't `create()` a block device).
OK so that could have been made a bit clearer in the test cases

TeddyAndrieux: Oh ok got it

> So the test for create() on block device should try with an existing one (which would succeed) and a missing one (which would fail, because we can't create() a block device).

No, it's not really this, calling `create` on raw block device will never succeed, it will always raise this error, because, I think, we do not call the `create` function when `exists` return True (which is always the case when the block device exists) but for test purpose we should maybe define an "un-existing" volume so that the test case is easier to understand

slaperche-scality: Yeah, you're not suppose to call `create` on a raw block device (sure you can, but as expected it will always fail because you can't create a raw block device out of thin air).
In the Salt state, it's guarded by a call to `exists`.

PR #2723, opened 2020-08-06
Deploy Fluent Bit

### LINKED ISSUE #2701: Deployment of  Fluent-Bit from Helm charts
Component: salt, charts, log, build

Why this is needed:
See #2668
This implementation may not be the final one, as we will probably change it afterwards to use the Fluent-Bit Operator, but we need it to have a functional logging pipeline for now.

What should be done:

Implementation proposal (strongly recommended):
As we already did for other add-ons (Prometheus, Dex, ...), we can use the charts provided by Fluent-Bit to ease the generation of Salt formulas, using the charts/renderer.py script:

```
  helm repo add loki https://grafana.github.io/loki/charts
  helm repo update
  helm fetch -d charts --untar loki/fluent-bit
  ./charts/render.py fluent-bit --namespace metalk8s-logging \
    charts/fluent-bit.yaml charts/fluent-bit/ \
    > salt/metalk8s/addons/fluent-bit/deployed/charts.sls
```

Test plan:
Ensure the fluent-bit pods are running (1 per node).

alexandre-allard:
**Component**: salt, build, log

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2701

### DIFF
--- a/charts/fluent-bit.yaml
+++ b/charts/fluent-bit.yaml
@@ -1,0 +1,129 @@
+---
+loki:
+  serviceName: loki  # Defaults to "${RELEASE}-loki" if not set
+  servicePort: 3100
+  serviceScheme: http
+  servicePath: /api/prom/push
+  # user: user
+  # password: pass
+  # the whole `config` block is not used as it does not allow us to add
+  # additional inputs. To workaround this, we're deploying our own
+  # ConfigMap instead.
+  # config:
+  #   port: 2020
+  #   tenantID: '""'
+  #   batchWait: 1
+  #   batchSize: 10240
+  #   loglevel: warn
+  #   lineFormat: json
+  #   k8sLoggingParser: "Off"
+  #   removeKeys:
+  #     - kubernetes
+  #     - stream
+  #   autoKubernetesLabels: false
+  #   labels: '{job="fluent-bit"}'
+  #   labelMap:
+  #     kubernetes:
+  #       namespace_name: namespace
+  #       labels:
+  #         app: app
+  #         release: release
+  #       host: node
+  #       container_name: container
+  #       pod_name: instance
+  #     stream: stream
+  # parsers: # Allow to define custom parsers. The key here is the same as the one in the [PARSER] section of parsers.conf file.
+  #  - Name: json
+  #    Format: json
+  #    Time_Key: time
+  #    Time_Format: "%d/%b/%Y:%H:%M:%S %z"
+
+  # extraOutputs: # Allow to define extra outputs in addition to the one automatically created
+  #   - Name: stdout
+  #     Format: json
+  #     json_date_format: time
+
+affinity: {}
+
+annotations: {}
+
+deploymentStrategy: RollingUpdate
+
+image:
+  repository: '__image__(fluent-bit-plugin-loki)'
+  tag: 1.5.0-amd64
+  pullPolicy: IfNotPresent
+
+nameOverride: fluent-bit
+
+## Node labels for pod assignment
+## ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+nodeSelector: {}
+
+## Pod Labels
+podLabels: {}
+
+podAnnotations:
+  prometheus.io/scrape: "true"
+  prometheus.io/port: "2020"
+  prometheus.io/path: /api/v1/metrics/prometheus
+
+## Assign a PriorityClassName to pods if set
+# priorityClassName:
+
+rbac:
+  create: true
+  pspEnabled: true
+
+resources:
+  limits:
+    memory: 100Mi
+  requests:
+    cpu: 100m
+    memory: 100Mi
+
+serviceAccount:
+  create: true
+  name:
+
+## Tolerations for pod assignment
+## ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+tolerations:
+- key: node-role.kubernetes.io/bootstrap
+  effect: NoSchedule
+  operator: Exists
+- key: node-role.kubernetes.io/etcd
+  effect: NoSchedule
+  operator: Exists
+- key: node-role.kubernetes.io/infra
+  effect: NoSchedule
+  operator: Exists
+- key: node-role.kubernetes.io/master
+  effect: NoSchedule
+  operator: Exists
+
+# Extra volumes to scrape logs from
+volumes:
+- name: varlog
+  hostPath:
+    path: /var/log
+# /run/log is used to fetch journal logs
+- name: runlog
+  hostPath:
+    path: /run/log
+
+volumeMounts:
+- name: varlog
+  mountPath: /var/log
+  readOnly: true
+- name: runlog
+  mountPath: /run/log
+  readOnly: true
+
+serviceMonitor:
+  enabled: true
+  interval: ""
+  additionalLabels:
+    release: prometheus-operator
+  annotations: {}
+  # scrapeTimeout: 10s
--- a/salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
+++ b/salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
@@ -1,0 +1,104 @@
+include:
+  - ...deployed.namespace
+
+Create fluent-bit ConfigMap:
+  metalk8s_kubernetes.object_present:
+    - manifest:
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: fluent-bit
+          namespace: metalk8s-logging
+          labels:
+            app: fluent-bit
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/name: fluent-bit
+            app.kubernetes.io/part-of: metalk8s
+            heritage: metalk8s
+            release: fluent-bit
+        data:
+          fluent-bit.conf: |-
+            [SERVICE]
+                HTTP_Server    On
+                HTTP_Listen    0.0.0.0
+                HTTP_PORT      2020
+                Flush          1
+                Daemon         Off
+                Log_Level      warn
+                Parsers_File   parsers.conf
+            [INPUT]
+                Name           tail
+                Tag            kube.*
+                Path           /var/log/containers/*.log
+                Parser         docker
+                DB             /run/fluent-bit/flb_kube.db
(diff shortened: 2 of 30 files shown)


charts/fluent-bit.yaml
```diff
@@ -0,0 +1,126 @@
...
+  labelMap:
+    kubernetes:
+      namespace_name: namespace
+      labels:
+        app: app
+        release: release
+      host: node
+      container_name: container
+      pod_name: instance
+    stream: stream
```
TeddyAndrieux: Since you do not use the config from the chart maybe it makes sense to comment this section and add a comment to say that we do not use this one from the chart as we cannot provide multiple inputs

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
```diff
@@ -0,0 +1,104 @@
...
+                Name           tail
+                Tag            kube.*
+                Path           /var/log/containers/*.log
+                Parser         docker
+                DB             /run/fluent-bit/flb_kube.db
+                Mem_Buf_Limit  5MB
+            [INPUT]
+                Name              systemd
+                Tag               host.*
+                DB                /run/fluent-bit/flb_journal.db
```
TeddyAndrieux: Minor: I know we do not really care, but look like indentation is not the same as the other block

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
```diff
@@ -0,0 +1,104 @@
...
+            [FILTER]
+                Name           kubernetes
+                Match          kube.*
+                Kube_URL       https://kubernetes.default.svc:443
+                Merge_Log On
+                K8S-Logging.Parser Off
+            [FILTER]
+                Name           modify
+                Match          host.*
+                Rename         HOSTNAME node
```
TeddyAndrieux: Not sure we should rename it to node because hostname may be different from node name :/
I would instead retrieve the node name of the current host here if it's doable (or add a hostname label on k8s logs but not sure it's easy to do as well) 
If it's not possible or to much work, then ignore this comment and let's keep hostname as node for the moment

tests/post/steps/test_logging.py
```diff
@@ -56,6 +73,60 @@ def _check_loki_ready():
...
+            "Failed to create logger Pod manifest template '{}': {}"
+        ).format(template_path, write_template.stderr)
+
+        pod_creation_ts = int(time.time())
+        manage_static_pod = host.run(
+            "salt-call state.single metalk8s.static_pod_managed "
+            'name=%s source=%s',
+            manifest_path,
+            template_path,
+        )
```
TeddyAndrieux: Why create a static pod and using salt ? Would be better to create the pod using kubectl (or python k8s) (with a label selector if you really want it to be on bootstrap node, but I think we do not really care here)

tests/post/steps/files/logger-pod.yaml.tpl
```diff
@@ -0,0 +1,20 @@
...
+  name: $name
+  namespace: metalk8s-logging
+spec:
+  tolerations:
+  - key: "node-role.kubernetes.io/bootstrap"
+    operator: "Equal"
+    effect: "NoSchedule"
+  - key: "node-role.kubernetes.io/infra"
+    operator: "Equal"
+    effect: "NoSchedule"
```
TeddyAndrieux: FYI: Since for the moment you use static pod you do not need tolerations

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
```diff
@@ -0,0 +1,104 @@
...
+            {
+              "kubernetes": {
+                "container_name": "container",
+                "host": "node",
+                "labels": {
+                  "app": "app",
+                  "release": "release"
+                },
+                "namespace_name": "namespace",
+                "pod_name": "instance"
```
TeddyAndrieux: It's a bit weird to have `instance` label being the pod name, why not name it `pod` ?

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
TeddyAndrieux: PS look like hostname is not really good, on my test env I got some logs with `localhost` and `localhost.localdomain` as node  (so likely HOSTNAME)
Looking on log it's seems these logs are not really relevant (maybe we should filter out logs with hostname `localhost` and `localhost.localdomain` )

Edit: Ok log with `localhost` and `localhost.localdomain` are only logs at the beginning before `systemd-hostnamed` update the hostname

tests/post/steps/test_logging.py
alexandre-allard: To not rely on kubectl, that's the only point, but I don't find a good reason for that.
I'll change it.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: This was defined this way in the ConfigMap generated with the helm chart, I've changed the configuration as little as possible.
But you're right, it's clearer with `pod`, I'll change it.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: Right, but even the configuration generated from the helm chart is not consistent with indentation. :D
I'll change it.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: It seems to be a little bit tricky to achieve that.
I can't find an easy way to retrieve the real hostname from the kubernetes logs or to retrieve the nodename from journal logs.
I've tried to use `$HOSTNAME` env var, but it's set to the container hostname... not really useful.
Plus, the helm chart does not allow to set up env vars based on pod specs... otherwise we could have done:
```
[...]
spec:
  containers:
    - env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
      [...]
```
And use the env var in fluent bit configuration.
Maybe we should just have 2 different labels `host` and `node`... or we should not use this chart.

tests/post/steps/test_logging.py
```diff
@@ -56,6 +72,49 @@ def _check_loki_ready():
...
+    manifest_path = str(MANIFESTS_PATH / "{}.yaml".format(pod_name))
+
+    with host.sudo():
+        write_manifest = utils.write_string(host, manifest_path, manifest)
+        assert write_manifest.rc == 0, (
+            "Failed to write logger Pod manifest '{}': {}"
+        ).format(manifest_path, write_manifest.stderr)
+
+        pod_creation_ts = int(time.time())
+        result = host.run('kubectl apply -f %s', manifest_path)
```
TeddyAndrieux: Wow, it's not really good here, you put the file iin the `/etc/kubernetes/manifest/` directory and you kubectl apply, you should only do the kubectl apply otherwise I think here you will create 2 pods: 
- one static pod `logger-test-bootstrap` on the bootstrap node (because of a YAML in the `/etc/kubernetes/manifest` directory)
- one pod `logger-test` on a "random" node (because of the kubectl apply)

IMHO you should use the same approach as in `tests/post/steps/test_dns.py` it's would be simpler

TeddyAndrieux (commented):
Just a comment about tests

tests/post/steps/test_logging.py
alexandre-allard: True, don't know what I had in mind, I may need some sleep. :D
Thanks for catching this.

salt/metalk8s/addons/logging/loki/deployed/services.sls
```diff
@@ -0,0 +1,29 @@
...
+            app: loki
+            app.kubernetes.io/name: loki
+            app.kubernetes.io/part-of: metalk8s
+        spec:
+          clusterIP: None
+          ports:
+            - name: http
+              port: 3100
+              protocol: TCP
+              targetPort: http
```
TeddyAndrieux: Does it works ? The only defined port for Loki is `http-metrics`

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
```diff
@@ -0,0 +1,109 @@
...
+                Remove         TRANSPORT
+                Remove         SYSLOG_FACILITY
+                Remove         BOOT_ID
+                Remove         UID
+                Remove         GID
+{%- for index in range(loki.spec.deployment.replicas) %}
+            [Output]
+                Name           loki
+                Match          *
+                Url            http://loki-{{ index }}:3100/api/prom/push
```
TeddyAndrieux: According to documentation, this endpoint is deprecated in favor of `loki/api/v1/push` but the new endpoint does not use the same POST format, fluent-bit do not handle the new format ?

https://grafana.com/docs/loki/latest/api/#post-apiprompush

salt/metalk8s/addons/logging/loki/deployed/services.sls
alexandre-allard: No, it does not work, I've tested with http-metrics and I made a mistake rewriting the sls.
We can even see that tests do not pass.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: As said in a previous comment, I've changed as less as possible the generated configmap, this was the endpoint defined by the helm chart.
I'll try with the new route and change it accordingly if it does work.

alexandre-allard: Looks like it works.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
TeddyAndrieux: Ok let's use `hostname` and `pod` for the moment, I updated the dashboard accordingly

TeddyAndrieux (approved):
LGTM

Issue #2724, opened 2020-08-06
NVME devices are incorrectly identified as partition


slaperche-scality:
**Component**:

operator, salt

**What happened**:

When trying to create a Block volume with an NVME device (such as `/dev/nvme0n1`), the creation fails with:

```
Status:
  Conditions:
    Last Transition Time:  2020-08-06T00:31:42Z
    Last Update Time:      2020-08-06T00:31:42Z
    Message:               Salt job 'PrepareVolume' failed with: Failed to prepare volume hd01-data1: error while trying to run `sgdisk --partition-guid 1:c16b9ed7-5f8d-4eba-8275-688875522fbc --change-name 1:hd01-data1 /dev/nvme0n`: Problem opening /dev/nvme0n for reading! Error is 2.
The specified file does not exist!.
    Reason:  CreationError
    Status:  False
    Type:    Ready
  Job:       PrepareVolume/20200806003107016653
Events:
  Type     Reason           Age   From               Message
  ----     ------           ----  ----               -------
  Normal   SaltCall         70s   volume-controller  volume provisioning step 1/2 started
  Normal   StateTransition  70s   volume-controller  volume phase transition from '' to 'Pending'
  Warning  SaltCall         35s   volume-controller  step 'volume provisioning (1/2)' failed
  Normal   StateTransition  35s   volume-controller  volume phase transition from 'Pending' to 'Failed'
```

The problem is that we wrongly identified `/dev/nvme0n1` as a partition and tried to use `/dev/nvme0n` as device (which doesn't exist).

**What was expected**:

`/dev/nvme0n1` is recognized as a device and the volume is correctly created.

**Steps to reproduce**

Try to create a volume with a device name ending with a number.

**Resolution proposal**:

Use a more precise regex (e.g. similar to the one used by `node-exporter`) instead of "ends with a number".

PR #2725, opened 2020-08-06
salt/volume: use a better regexp to recognize partitions

slaperche-scality:
**Component**:

salt

**Context**: 

The current regexp to recognize partitions was too simple and was treating all NVME devices as partitions.

**Summary**:

We replace use a new regexp inspired by the one used for `ignored-devices` in `node-exporter`.

**Acceptance criteria**: 

You can create Block volume on NVME devices.

---

Closes: #2724


gdemonet:
Now that we have unit tests in place for this module, could we also add a test case to check that NVME devices aren't interpreted as partition?

slaperche-scality:
> Now that we have unit tests in place for this module, could we also add a test case to check that NVME devices aren't interpreted as partition?

Done

gdemonet (approved):

Issue #2728, opened 2020-08-10
Failure to obtain device name from Salt results in error message stored in a Volume's `status.deviceName` field


gdemonet:
**Component**: storage-operator, salt

**What happened**:

From `kubectl describe volume ...`:

```
Device Name:             The minion function caused an exception: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1676, in _thread_return
    return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
  File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 12, in execute
    return func(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/modules/metalk8s_volumes.py", line 186, in device_name
    raise Exception('device `{}` not found'.format(path))
Exception: device `/dev/disk/by-uuid/dadda9e7-dc1f-4632-8450-61eb3bdf0db3` not found
```

After repairing the lost symlink (see #2726), the `deviceName` value was not reconciled, until a manual edition of the Volume object triggered it again.

**What was expected**:

The storage-operator shouldn't store this error message as a value for `status.deviceName`, but instead (as the related code suggests) should send an event with the error as a message.
This error should be seen as transient by the storage-operator, ensuring that fixing the underlying issue would be reconciled and the actual `deviceName` would end up being populated.

**Steps to reproduce**:

See reproducing steps for #2726 - once a loop devices symlink is lost after reboot, this error should appear in the Volume `status`.

**Resolution proposal** (optional):

The way this error gets retrieved (through SaltAPI) is probably due to the error being raised (to be checked). Instead of raising a plain `Exception` from `metalk8s_volumes.device_name`, try raising `salt.exceptions.CommandExecutionError` instead (which seems to be standard practice in Salt execution modules).

PR #2731, opened 2020-08-11
Fix device name resolution when the device doesn't exist

slaperche-scality:
**Component**:

salt, operator

**Context**: 

When the device name resolution fails because the given device doesn't exists, we wrongly interpret it as a success, thanks to the ambiguous Salt response.

**Summary**:

Instead of returning a plain string, return a dict with a success field and a result field.
See the commit message for more info on the different solutions that have been tested.

**Acceptance criteria**: 

- Delete the device symlink of the target volume `vol1`: `rm /dev/disk/by-uuid/<vol1-ID>`
- Edit the volume (e.g. change its size or path) to trigger a reconciliation: `kubectl edit volume <vol1-name>`
- Observe in the storage-operator logs that ① we have an error, ② we reschedule a reconciliation
- Recreate the device symlink: `ln -s <pv-devicepath> /dev/disk/by-uuid/<vol1-ID>`
- Observe that the reconciliation is a success on the next retry.

---

Closes: #2728

### DIFF
--- a/salt/tests/unit/modules/test_metalk8s_volumes.py
+++ b/salt/tests/unit/modules/test_metalk8s_volumes.py
@@ -438,13 +438,13 @@ class Metalk8sVolumesTestCase(TestCase, LoaderModuleMockMixin):
 
     @parameterized.expand([
         # Nominal case: device exists.
-        ([True], 'my-device'),
+        ('exists', [True], 'my-device'),
         # Error case: device doesn't exists.
-        ([False]*10, 'device `/dev/my-device` not found'),
+        ('missing', [False]*10, 'device `/dev/my-device` not found'),
         # Device is temporarily missing.
-        ([False, False, False, True], 'my-device'),
+        ('transient-missing', [False, False, False, True], 'my-device'),
     ])
-    def test_device_name(self, exist_values, result):
+    def test_device_name(self, _, exist_values, result):
         expected = {'success': any(exist_values), 'result': result}
         exists_mock = MagicMock(side_effect=exist_values)
         realpath_mock = MagicMock(side_effect=lambda path: path)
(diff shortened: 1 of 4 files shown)


storage-operator/pkg/salt/client.go
```diff
@@ -590,14 +590,27 @@ func parsePollAnswer(
...
-		if result, ok := results[0].(map[string]interface{}); ok {
-			if name, ok := result[nodeName].(string); ok {
-				return name, nil
+		if nodeResults, ok := results[0].(map[string]interface{}); ok {
+			if result, ok := nodeResults[nodeName].(map[string]interface{}); ok {
+				// Inspect "success" and "result".
+				is_success, has_success := result["success"].(bool)
+				output, has_result := result["result"].(string)
+				if !has_success || !has_result {
+					goto invalidFormat
```
gdemonet, line 604: TIL about this statement!

gdemonet (dismissed):
All good for me, thanks for tackling this!
The behaviour of Salt execution modules, few years after starting to read/write them, still is a mystery to me...

salt/tests/unit/modules/test_metalk8s_volumes.py
```diff
@@ -435,38 +435,25 @@ def test_clean_up(self, name, raise_msg=False, pillar_volumes=None,
...
     @parameterized.expand([
-        param('my-device'),
-        param(exists_return=False, raises=True, result='device `/dev/my-device` not found'),
-        param(exists_return=[False, False, False, True], result='my-device')
+        # Nominal case: device exists.
+        ([True], 'my-device'),
+        # Error case: device doesn't exists.
+        ([False]*10, 'device `/dev/my-device` not found'),
+        # Device is temporarily missing.
+        ([False, False, False, True], 'my-device'),
```
gdemonet: FWIW, you can use a string as first argument to each test case, it will be appended to the generated function name:

```python
    @parameterized.expand([
        ("exists", [True], 'my-device'),
        ("does not exist", [False]*10, 'device `/dev/my-device` not found'),
        ("temporarily missing", [False, False, False, True], 'my-device'),
    ])
    def test_device_name(self, _, exist_values, result):
        # ...
```

So instead of getting `pytest` output similar to:

```
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_0 PASSED [ 33%]
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_1 PASSED  [ 66%]
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_2 PASSED  [100%]
```

You'll have:


```
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_0_exists PASSED [ 33%]
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_1_does_not_exist PASSED  [ 66%]
salt/tests/unit/modules/test_metalk8s_volumes.py::Metalk8sVolumesTestCase::test_device_name_2_temporarily_missing PASSED  [100%]
```

gdemonet (dismissed):

gdemonet (dismissed):

salt/tests/unit/modules/test_metalk8s_volumes.py
slaperche-scality: Done

gdemonet (approved):

Issue #2704, opened 2020-08-03
Bootstrap may fail if epel repo enabled with some priority


TeddyAndrieux:
**Component**:

'salt'

**What happened**:

When running bootstrap with epel repo enable and has some yum priority https://wiki.centos.org/PackageManagement/Yum/Priorities set on it, salt from epel repository is installed (2015.x).

So bootstrap fail because some expected function not here (like `fopen`)

```
[CRITICAL] Rendering SLS 'base:metalk8s.salt.minion.local' failed: Jinja error: 'module' object has no attribute 'fopen'
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 364, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 2, in top-level template code
  File "/var/cache/salt/minion/extmods/modules/metalk8s.py", line 211, in get_archives
    info = archive_info_from_tree(archive)
  File "/var/cache/salt/minion/extmods/modules/metalk8s.py", line 158, in archive_info_from_tree
    with salt.utils.files.fopen(product_txt) as fd:
AttributeError: 'module' object has no attribute 'fopen'

; line 2

---
{%- from "metalk8s/map.jinja" import metalk8s with context %}
{%- set archives = salt.metalk8s.get_archives() %}    <======================

Configure salt minion for local mode:
  file.serialize:
    - name: /etc/salt/minion.d/99-metalk8s-local.conf
    - user: root
[...]
---
local:
    Data failed to compile:
----------
    Rendering SLS 'base:metalk8s.salt.minion.local' failed: Jinja error: 'module' object has no attribute 'fopen'
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 364, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
[...]

PR #2734, opened 2020-08-11
build,scripts,salt: Pin salt version in bootstrap and restore and set repo priority to 1

TeddyAndrieux:
**Component**:

'salt', 'build', 'scripts'

**Context**: 

#2704 

**Summary**:

In bootstrap and restore we install salt-minion on the node so we need
to pin the salt version we want to install and not just install the one
yum will find depending on various things (like available versions,
priority on repos, ...).
Also set the priority to 1 on all our repos so that our repositories has
priority over others if the `yum-plugin-priorities` package is
installed.
NOTE: If `yum-plugin-priorities` is installed and priority is set on one
repos then other repos with the same package will just be ignored
(e.g.: epel installed with priority=98 we cannot install `salt-3000.3`
as salt-3000.3 is excluded because a repo with some priority has the
package in a different version)

**Test**:

Tested manually with
```
[epel]
name=Extra Packages for Enterprise Linux 7 - $basearch
#baseurl=http://download.fedoraproject.org/pub/epel/7/$basearch
metalink=https://mirrors.fedoraproject.org/metalink?repo=epel-7&arch=$basearch
failovermethod=priority
enabled=1
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7
priority=9
```
And `yum-plugin-priorities` installed and it works well (even if salt-2015 is already installed before bootstrapping)

---

Fixes: #2704

### DIFF
--- a/salt/metalk8s/repo/redhat.sls
+++ b/salt/metalk8s/repo/redhat.sls
@@ -65,6 +65,9 @@ Configure {{ repo_name }} repository:
   {%- endif %}
     - repo_gpg_check: {{ repo_config.repo_gpg_check }}
     - enabled: {{ repo_config.enabled }}
+    # Set priority to 1, even if the `yum-plugin-priorities` is not installed,
+    # so that our repos has priority over others
+    - priority: 1
     # URL to the proxy server for this repository.
     # Set to '_none_' to disable the global proxy setting
     # for this repository.
--- a/buildchain/buildchain/iso.py
+++ b/buildchain/buildchain/iso.py
@@ -85,7 +85,10 @@ FILE_TREES : Tuple[helper.FileTree, ...] = (
                 task_name='bootstrap.sh',
                 source=constants.ROOT/'scripts'/'bootstrap.sh.in',
                 destination=constants.ISO_ROOT/'bootstrap.sh',
-                context={'VERSION': versions.VERSION},
+                context={
+                    'VERSION': versions.VERSION,
+                    'SALT_VERSION': versions.SALT_VERSION
+                },
                 file_dep=[versions.VERSION_FILE],
                 task_dep=['_iso_mkdir_root'],
             ),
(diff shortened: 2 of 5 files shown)


salt/metalk8s/repo/redhat.sls
```diff
@@ -65,6 +65,9 @@ Configure {{ repo_name }} repository:
   {%- endif %}
     - repo_gpg_check: {{ repo_config.repo_gpg_check }}
     - enabled: {{ repo_config.enabled }}
+    # Set priority to 1, even if the `yum-plugin-priorities` is not installed,
+    # so that our repos has priority over others
+    - priority: 1
```
gdemonet, line 70: Lower numbers mean higher priority?

gdemonet (approved):

salt/metalk8s/repo/redhat.sls
TeddyAndrieux: Yep and if for example epel has priority=4 and our one has priority=5 you cannot install salt 3000 (except if you disable the priority plugin)

gdemonet: ? So it only works with `priority=1`? I don't understand

TeddyAndrieux: No just you need higher priority than others repos that provide the same package or your package will be ignored 
(epel priority=4 provide salt-2015 so if our repo has priority=5 salt from our repo will just be ignored )
```
[root@test-salt-install ~]# yum install salt-minion-3000.3
Loaded plugins: fastestmirror, priorities
Loading mirror speeds from cached hostfile
 * base: mirroir.wptheme.fr
 * epel: epel.mirror.wearetriple.com
 * extras: centos.mirrors.proxad.net
 * updates: mirroir.wptheme.fr
236 packages excluded due to repository priority protections
No package salt-minion-3000.3 available.
Error: Nothing to do
```

gdemonet: Oh OK, you switched them in the first version of your comment so I thought there was a weird behaviour. OK makes sense!

Issue #2631, opened 2020-06-17
Configure the node-exporter to get the metrics for partitioned disk


ChengYanJin:
**Component**:

**Why this is needed**:
I would like to get the `IOPS/Throughput/Latency` for a specific `Volume`. But the metrics expose by Prometheus are the whole disk(device) before the partition.

After the job done by #2615 I could be able to get the device name from `Volume status.`

**What should be done**:
Configure the node-exporter to expose the metric for the partitioned disk.

**Implementation proposal** (strongly recommended):

A workaround could be:
1. `kubectl get -n metalk8s-monitoring daemonset`
2. Edit `kubectl edit -n metalk8s-monitoring daemonset prometheus-operator-prometheus-node-exporter`
3. ` --collector.diskstats.ignored-devices="^(ram|fd|(h|s|v|xv)d[a-z]|nvme\\d+n\\d+p)$":`

Update the regexp:
 ` --collector.diskstats.ignored-devices="^(ram|fd)\d+$"`

**Test plan**:
TBD

slaperche-scality:
We may also want to export loop device (for `SparseLoopDevice`, even though we're not supposed to use them in prod)

gdemonet:
Currently, `node-exporter` uses its default value for selection of devices to gather diskstats from. The default regexp for ignored devices is the following (as found [here](https://github.com/prometheus/node_exporter/blob/master/collector/diskstats_linux.go#L39)):

```
^(ram|loop|fd|(h|s|v|xv)d[a-z]|nvme\\d+n\\d+p)\\d+$
```

A solution could be to use `--collector.diskstats.ignored-devices="^(ram|fd|nvme\\d+n\\d+p)\\d+$"` as a command-line flag passed to `node-exporter` (in its DaemonSet PodTemplate).

One would need to make sure extending the set of collected devices will not break / make unreadable the Grafana charts that use this metric.

gdemonet:
So, I'm not sure how `node-exporter` works, but even after changing the regexp to `^(ram|fd)\\d+$` I have no stat for `loopX` devices :/
It seems there is no metric at all in `/proc/diskstats` about loop devices.

_Edit: actually, there are some stats, but all set to 0, and apparently filtered out from Grafana dashboards_

deeruattavar:
Above changes not working -   ^(ram|fd)\\d+$

Still I am unable to loop all partitions .. even I can see see some replication values for all disk partitions.

[...]

PR #2740, opened 2020-08-13
Collect I/O metrics from disk partitions and loop devices

gdemonet:
**Component**: node-exporter, tests

**Context**: See #2631 (we need metrics for all kinds of devices that can be used in Volumes)

**Summary**:

- Adapt `node-exporter`'s `collector.diskstats.ignored-devices` to only ignore RAM disks and floppy disks
- Add a test to ensure we can actually get metrics for a `SparseLoopDevice` Volume

**Acceptance criteria**: 

- tests pass
- metrics can be retrieved for all kinds of devices allowed in Volumes

_Note that those metrics are always 0 for loop devices on old kernels such as what comes with CentOS 7_

---

Closes: #2631

### DIFF
--- a/tests/post/features/monitoring.feature
+++ b/tests/post/features/monitoring.feature
@@ -43,3 +43,9 @@ Feature: Monitoring is up and running
         Given the Kubernetes API is available
         And we have 1 running pod labeled 'prometheus=prometheus-operator-prometheus' in namespace 'metalk8s-monitoring'
         Then the deployed Prometheus alert rules are the same as the default alert rules
+
+    Scenario: Volume metrics can be found based on device name
+        Given the Kubernetes API is available
+        And I have created a test Volume
+        And the Prometheus API is available
+        Then I can get I/O stats for this test Volume's device
--- a/tests/post/features/monitoring.feature
+++ b/tests/post/features/monitoring.feature
@@ -46,6 +46,6 @@ Feature: Monitoring is up and running
 
     Scenario: Volume metrics can be found based on device name
         Given the Kubernetes API is available
-        And I have created a test Volume
         And the Prometheus API is available
+        And a test Volume 'test-monitoring1' exists
         Then I can get I/O stats for this test Volume's device
--- a/tests/post/steps/test_monitoring.py
+++ b/tests/post/steps/test_monitoring.py
@@ -1,19 +1,26 @@
 import json
 import pathlib
+import random
+import string
 
 import kubernetes.client
+from kubernetes.client import CustomObjectsApi
 from kubernetes.client.rest import ApiException
 
 import pytest
 from pytest_bdd import scenario, given, then, parsers
+import testinfra
 
 from tests import utils
+from tests import kube_utils
 
 # Constants {{{
 
 ALERT_RULE_FILE_NAME = 'alerting_rules.json'
 ALERT_RULE_FILE_PATH = (pathlib.Path(__file__)/'..'/'..'/'..'/'..'
             /'tools'/'rule_extractor'/ALERT_RULE_FILE_NAME).resolve()
+
+NODE_EXPORTER_PORT = 9100
 
 # }}}
 # Scenarios {{{
@@ -55,6 +62,13 @@ def test_deployed_prometheus_rules(host):
     pass
 
 
+@scenario(
+    '../features/monitoring.feature',
+    'Volume metrics can be found based on device name')
+def test_volume_metrics(host):
+    pass
+
+
 # }}}
 # Given {{{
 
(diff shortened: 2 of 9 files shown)


tests/post/steps/test_volume.py
```diff
@@ -93,23 +91,27 @@
...
 def volume_client(k8s_apiclient, ssh_config):
-    return VolumeClient(CustomObjectsApi(api_client=k8s_apiclient), ssh_config)
+    return kube_utils.kube_utils.VolumeClient(
+        CustomObjectsApi(api_client=k8s_apiclient), ssh_config
+    )
 
 @pytest.fixture
 def pv_client(k8s_client):
-    return PersistentVolumeClient(k8s_client)
+    return kube_utils.Persistentkube_utils.VolumeClient(k8s_client)
```
gdemonet: that can't work :facepalm:

tests/post/features/monitoring.feature
```diff
@@ -42,4 +42,10 @@ Feature: Monitoring is up and running
...
         And we have 1 running pod labeled 'prometheus=prometheus-operator-prometheus' in namespace 'metalk8s-monitoring'
-        Then the deployed Prometheus alert rules are the same as the default alert rules
\ No newline at end of file
+        Then the deployed Prometheus alert rules are the same as the default alert rules
+
+    Scenario: Volume metrics can be found based on device name
+        Given the Kubernetes API is available
+        And I have created a test Volume
+        And the Prometheus API is available
+        Then I can get I/O stats for the test Volume device
```
slaperche-scality: To make the test faster/less flaky, I guess you could skip the creation of a brand new volume and try to get the stats from one of the two already existing volume (alertmanager's one or prometheus' one).

slaperche-scality (commented):
minor remark: I would have done the code moving/refactoring in a dedicated commit, to have a smaller commit introducing the new tests.

tests/post/features/monitoring.feature
gdemonet: Right... But I didn't want to rely on the fact that this volume _may_ be using a loop device (ideally, our `post` E2E tests are non intrusive, so we can run them to perform a sanity check against a "real" cluster).

gdemonet:
> I would have done the code moving/refactoring in a dedicated commit

Good point, will do that!

tests/post/features/monitoring.feature
slaperche-scality: ha yeah right, we want to check against a loop device specifically (since attaching a real partitioned volume would be even flakier…)

tests/post/steps/test_monitoring.py
```diff
@@ -81,6 +96,35 @@ def _check_object_exists():
...
+        assert volume['status'].get('deviceName') is not None, \
+            'Volume status.deviceName has not been reconciled'
+        return volume
+
+    yield utils.retry(
+        _check_volume_ready, times=30, wait=2,
+        name='waiting for Volume {} to be ready'.format(volume_name)
+    )
+
+    client.delete(volume_name, sync=True)
```
slaperche-scality: Can't we reuse the existing `@given(parsers.parse("a Volume '{name}' exist"))`
Looks very similar.

If not the step directly, at least the implementation (to reduce code duplication)

gdemonet: Can do!

gdemonet: Did something, but couldn't factorize everything (the test Volume doesn't have a static name, and the teardown is handled directly in the `given` fixture).

tests/post/steps/test_monitoring.py
```diff
@@ -81,6 +97,31 @@ def _check_object_exists():
...
+    def _check_volume_ready():
+        volume = _get_volume_and_check_status(client, volume_name, 'Available')
+        assert volume['status'].get('deviceName') is not None, \
+            'Volume status.deviceName has not been reconciled'
+        return volume
+
+    yield utils.retry(
+        _check_volume_ready, times=30, wait=2,
+        name='waiting for Volume {} to be ready'.format(volume_name)
+    )
```
slaperche-scality: Can't reuse `check_volume_status` for that one?

tests/post/steps/test_monitoring.py
slaperche-scality: Do we really need the random name?
We could use a simple naming scheme, like `test-monitoring-volumeN`?

tests/post/steps/test_monitoring.py
gdemonet: They are slightly different actually:
- this one checks if there's a device name - so it's a bit more than just checking the status phase
- `check_volume_status` is used in `test_volume` scenarii to manipulate the `context` fixture - we don't need it here

slaperche-scality: Ok fair enough.

tests/post/steps/test_monitoring.py
gdemonet: Most of the logic was factorized in the `VolumeClient` directly :slightly_smiling_face: but still not sharing the fixture/step because of how Volume tests rely on the `context` fixture.

slaperche-scality (dismissed):

gdemonet:
@slaperche-scality I had to rebase because of a bug in my test code. FYI, I was trying to resolve the real node name from what was defined in `spec.nodeName` of the Volume, but unlike what we need to do at creation time (i.e. replacing the "bootstrap" name in SSH config with the real node name we use), the Volume was already created hence this `nodeName` was already patched. So I could just use it as-is:

```python
    control_plane_ip = utils.get_grain(
        test_volume['spec']['nodeName'],
        'metalk8s:control_plane_ip'
    )
```

gdemonet:
> ```python
>     control_plane_ip = utils.get_grain(
>         test_volume['spec']['nodeName'],
>         'metalk8s:control_plane_ip'
>     )
> ```

Of course, this doesn't work either, because `utils.get_grain` expects a `host` object from testinfra... Need to think this through again.

gdemonet:
```python
    # Retrieve control-plane IP of another Node through Salt master, not
    # through testinfra, because the actual Node name may not match our
    # SSH config file
    node_name = test_volume['spec']['nodeName']
    command = [
        'salt',
        '--out json',
        node_name,
        'grains.get',
        'metalk8s:control_plane_ip'
    ]
    result = utils.run_salt_command(host, command, ssh_config)
    control_plane_ip = json.loads(result.stdout)[node_name]
```

This one should work :crossed_fingers:

slaperche-scality (approved):

Issue #2653, opened 2020-07-01
Admin user (in Dex) is not admin on Grafana


gdemonet:
**Component**: grafana

**What happened**:

When logging into Grafana, with the default `admin@metalk8s.invalid` static user in Dex, we get read-only access.

**What was expected**:

This user should have super-admin access.

**Steps to reproduce**: (affects all 2.5+ versions)

**Resolution proposal** (optional):

Not clear: see https://grafana.com/docs/grafana/latest/auth/generic-oauth/#role-mapping

Ebaneck:
It seems like the Dex static user does not support `groups`  and from upstream discussions, Dex will not support this feature for static users. Connector based settings like LDAP, SAML and others seem to support this.

```
2020-07-10T09:10:47.776792798Z stderr F time="2020-07-10T09:10:47Z" level=info msg="login successful: connector \"local\", username=\"admin\", preferred_username=\"\", email=\"admin@metalk8s.invalid\", groups=[]"
```

This means that we cannot use the role-mapping feature from Grafana since the user info endpoint has no entries for `groups`.

Alternatively, we could fix this by using the below hack. We should, however, remove the below settings when we implement role-mapping with non-static users
```
diff --git a/charts/prometheus-operator.yaml b/charts/prometheus-operator.yaml
index e9072631..14c2d24d 100644
--- a/charts/prometheus-operator.yaml
+++ b/charts/prometheus-operator.yaml
@@ -164,6 +164,8 @@ grafana:
       auth_url: '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/auth)'
       token_url:  '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/token)'
       api_url: '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oidc/userinfo)'
+    users:
+      auto_assign_org_role: "Admin"
 
```

gdemonet:
Oh no we can't deploy Grafana with this "hack". Instead, we could use role-mapping to map this special username (no need to rely on groups) to the Admin role.

Ebaneck:
Reading the Grafana role mapping docs, here is a fix proposal
```
role_attribute_path: contains(email, 'admin@metalk8s.invalid') && 'Admin'
```

Ebaneck:
Instead of simply hard-coding this value like proposed above, we could indeed obtain the value from a read service configuration.

Why?
[...]

PR #2743, opened 2020-08-17
Bind Dex static admin user to a Grafana Admin role automatically

Ebaneck:
**Component**:

'salt', 'prometheus-operator', 'charts', 'grafana'

**Context**: 

See: #2653 

**Summary**:

When logging into Grafana, with the default admin@metalk8s.invalid Dex static user, we should automatically get full access permissions in Grafana. This is possible since we now make use of the Grafana role attribute to map the Dex user to a Grafana Admin role.

**Acceptance criteria**: 

- We should be able to log in to Grafana with `admin@metalk8s.invalid` user possessing Admin role attributes in Grafana.

---

Closes: #2653


TeddyAndrieux (approved):
LGTM

PR #2748, opened 2020-08-20
[ui] Add unit tests used by new volume page

### LINKED ISSUE #2747: Add Unit Test Used by the New Volume Page
**Component**: UI, Unit test

**Why this is needed**:
Test the utility functions

**What should be done**:

Unit test for Utility Functions
- addMissingDataPoint() & jointDataPointBaseonTimeSeries() used by Metrics Chart to maually add the missing data points.
- getVolumeListData() used by VolumeListTable to retrive the data for volume table
Already covered:
- Delete a Non-deletable volume and check if UI prevents the users to do so.

**Test plan**:

ChengYanJin:
**Component**: ui, test

**Context**: Add Unit test for Utility Functions

**Summary**:
- `addMissingDataPoint()` used by Metrics Chart to maually add the missing data points.
- `getVolumeListData()` used by VolumeListTable to retrive the data for volume table

Remove the function `jointDataPointBaseonTimeSeries()` which merges the result of the query based on `instance` and `device` since the Prometheus query can help us handle the aggregation.

Already exist:
Delete a Non-deletable volume and check if UI prevents the users to do so.

==========
Note that: 
Update the storageclass from `metalk8s-prometheus` to `metalk8s` due to the change of PR #2755.

**Acceptance criteria**: 
Pass all the tests

---

Closes: #2747


ui/src/services/utils.js
```diff
@@ -303,8 +282,17 @@ export function addMissingDataPoint(
...
   sampleFrequency,
 ) {
-  if (!orginalValues || orginalValues.length === 0) {
-    return;
+  if (
+    !orginalValues ||
+    orginalValues.length === 0 ||
+    !startingTimeStamp ||
+    !sampleDuration ||
+    !sampleFrequency ||
```
MonPote, line 288: This can be removed since it's already handle by the next 2 lines.

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -216,7 +218,8 @@ const MetricGraphCard = (props) => {
     volumeThroughtReadData,
   );
   const volumeIOPSData = volumeIOPSWriteData?.concat(volumeIOPSReadData);
-
+  console.log('volumeIOPSData', volumeIOPSData);
+  console.log('volumeThroughputData', volumeThroughputData);
```
MonPote: This can be remove ^^

ui/src/ducks/app/monitoring.js
```diff
@@ -330,15 +334,20 @@ export function* fetchVolumeStats() {
...
-  // the queries for Throughput/Latency/IOPS
+  // the queries for Throughput/IOPS
   // rate calculates the per-second average rate of increase of the time series in the range vector.
-  const volumeThroughputReadQuery = `rate(node_disk_read_bytes_total{job="node-exporter"}[1m])`;
-  const volumeThroughputWriteQuery = `rate(node_disk_written_bytes_total{job="node-exporter"}[1m])`;
-  const volumeLatencyQuery = `rate(node_disk_io_time_seconds_total{job="node-exporter"}[1m])`;
-  const volumeIOPSReadQuery = `irate(node_disk_reads_completed_total{job="node-exporter"}[5m])`;
-  const volumeIOPSWriteQuery = `irate(node_disk_writes_completed_total{job="node-exporter"}[5m])`;
-  const volumeUsedQuery = 'kubelet_volume_stats_used_bytes';
+  // group the result of the query by instance and device (remove the filter {job="node-exporter"})
```
gdemonet, line 341: https://github.com/scality/metalk8s/pull/2757#discussion_r477327539

ui/src/ducks/app/monitoring.js
```diff
@@ -427,28 +436,40 @@ export function* fetchVolumeStats() {
...
 export function* fetchCurrentVolumeStats() {
   let volumeUsedCurrent = {};
+  let volumeCapacityCurrent = {};
   let volumeLatencyCurrent = {};
 
-  const volumeLatencyCurrentQuery = `rate(node_disk_io_time_seconds_total{job="node-exporter"}[1h])`;
+  const volumeLatencyCurrentQuery = `irate(node_disk_io_time_seconds_total{job="node-exporter"}[1h])`;
   const volumeUsedQuery = 'kubelet_volume_stats_used_bytes';
-
+  // Grafana - Used Space: kubelet_volume_stats_capacity_bytes - kubelet_volume_stats_available_bytes
```
gdemonet, line 443: IIUC, this comment relates to the `volumeUsedQuery` above - in general, we write comments above the related code, not under it

ui/src/ducks/app/monitoring.js
```diff
@@ -330,15 +334,20 @@ export function* fetchVolumeStats() {
...
+  // group the result of the query by instance and device (remove the filter {job="node-exporter"})
+  const volumeThroughputReadQuery = `sum(irate(node_disk_read_bytes_total[1m])) by (instance, device)`;
+  const volumeThroughputWriteQuery = `sum(irate(node_disk_written_bytes_total[1m])) by (instance, device)`;
+  const volumeIOPSReadQuery = `sum(irate(node_disk_reads_completed_total[5m])) by (instance, device)`;
+  const volumeIOPSWriteQuery = `sum(irate(node_disk_writes_completed_total[5m])) by (instance, device)`;
+
+  // the query for latency
+  // Disk latency is the time that it takes to complete a single I/O operation on a block device
+  // rate(node_disk_read_time_seconds_total[5m]) / rate(node_disk_reads_completed_total[5m]
+  const volumeLatencyQuery = `sum(irate(node_disk_io_time_seconds_total[1m])) by (instance, device)`;
```
gdemonet, line 350: Not sure I understand this query... The commented query makes more sense to me, where did you get the other one from?

ui/src/ducks/app/monitoring.js
```diff
@@ -427,28 +436,40 @@ export function* fetchVolumeStats() {
...
+
+  const volumeCapacityCurrentQueryResult = yield call(
+    queryPrometheus,
+    volumeCapacityQuery,
+  );
+  if (!volumeCapacityCurrentQueryResult.error) {
+    volumeCapacityCurrent = volumeCapacityCurrentQueryResult.data.result;
+  }
+
   const volumeLantencyCurrentResult = yield call(
```
gdemonet: nitpick: there's a typo here `s/Lantency/Latency`

ui/src/services/NodeVolumesUtils.js
```diff
@@ -223,21 +231,26 @@ export const getVolumeListData = createSelector(
...
 
-      const instanceIP = nodeList.find(
+      const instanceIP = nodeList?.find(
         (node) => node.name === volume?.spec?.nodeName,
       )?.internalIP;
 
+      const volumeCurrentLatency = volumeLatencyCurrent?.find(
+        (vLV) =>
+          vLV.metric.device === volume?.status?.deviceName &&
+          vLV.metric.instance === instanceIP + PORT_NUMBER_PROMETHEUS,
```
gdemonet: Shouldn't it be:
```suggestion
          vLV.metric.instance === `${instanceIP}:${PORT_NUMBER_PROMETHEUS}`,
```
?

ui/src/services/NodeVolumesUtils.js
```diff
@@ -223,21 +231,26 @@ export const getVolumeListData = createSelector(
...
-                100,
-            )
-          : intl.translate('unknown'),
+        usage:
+          volumeUsedCurrent && volumeCapacityCurrent
+            ? (
+                (volumeUsedCurrent?.value[1] / volumeCapacityCurrent.value[1]) *
+                100
+              ).toFixed(2)
+            : intl.translate('unknown'),
```
gdemonet: Unrelated to this change, but it'd make more sense to leave `undefined` as a value here and handle the choice of message and its translation in a display component.

ui/src/services/NodeVolumesUtils.js
```diff
@@ -252,12 +265,7 @@ export const getVolumeListData = createSelector(
...
           // for latency we need to query the volumeLatecyCurrent based on both `instance` and `deviceName`
-          volumeLatencyCurrent &&
-          volumeLatencyCurrent?.find(
-            (vLV) =>
-              vLV.metric.device === volume?.status?.deviceName &&
-              vLV.metric.instance === instanceIP + PORT_NUMBER_PROMETHEUS,
-          )
+          volumeCurrentLatency
             ? Math.round(
                 volumeLatencyCurrent?.find(
```
gdemonet: You already found the latency for this volume, no need to `.find` again, just re-use `volumeCurrentLatency`

ui/src/services/utils.test.js
```diff
@@ -91,14 +95,559 @@ it('test the sort with a custom sortBy', () => {
...
   expect(result).toEqual([
     { yanjin: '1Ki' },
     { yanjin: '1Mi' },
     { yanjin: '100Mi' },
-    { yanjin: '1Gi' }
+    { yanjin: '1Gi' },
   ]);
 });
+
+const originalValue_happy_path = [
```
gdemonet: nit: why mix camel-case with snake-case?

ui/src/services/utils.test.js
```diff
@@ -91,14 +95,559 @@ it('test the sort with a custom sortBy', () => {
...
+  [1597792971, '4899266560'],
+  [1597796571, '5053898752'],
+  [1597800171, '4997726208'],
+  [1597803771, '5150830592'],
+  [1597807371, '5064650752'],
+  [1597810971, '5216804864'],
+  [1597814571, '5148012544'],
+  [1597818171, '5303730176'],
+  [1597821771, '5234618368'],
+  [1597825371, '5384785920'],
```
gdemonet: I don't think it's relevant to use 130 datapoints for a unit-test. Just get the minimal amount needed, that will help make the test case more obvious.

ui/src/services/utils.test.js
```diff
@@ -91,14 +95,559 @@ it('test the sort with a custom sortBy', () => {
...
+];
+const startingTimeStamp = 1597220571;
```
gdemonet: For the same reason, you don't need "real" values in a unit test (unless they are compared with real values you cannot control. So, starting from timestamp 0 is OK, in my opinion, and makes it more obvious for a reader to understand what's happening.

ui/src/services/utils.test.js
```diff
@@ -91,14 +95,559 @@ it('test the sort with a custom sortBy', () => {
...
+
+it('should return all 0 when the original dataset is all 0', () => {
+  const result = addMissingDataPoint(
+    originalValue_all_0,
+    startingTimeStamp,
+    sampleDuration,
+    sampleFrequency,
+  );
+  expect(result).toEqual(originalValue_all_0);
+});
```
gdemonet: This is a duplicate of the above.

ui/src/services/NodeVolumesUtilsData.js
```diff
@@ -0,0 +1,2371 @@
+export const state_volume_filtered_by_node_master0 = {
```
gdemonet: How is this variable describing "volumes filtered by node master0"? It looks to me as a rather "normal" application state, with three nodes and volumes on each node...

ui/src/services/NodeVolumesUtils.test.js
```diff
@@ -420,3 +422,96 @@ it('should return error when called with failed Ready condition', () => {
...
+    hash: '',
+    key: 'u757uz',
+  },
+  match: {
+    path: '/volumes',
+    url: '/volumes',
+    isExact: false,
+    params: {},
+  },
+};
```
gdemonet: Do we really need all these props? It seems we only read the `location` part anyway...

On the side, it seems to me there are way too many things happening in this selector, and maintaining this code and associated tests will be a pain when the behaviour changes. We'll see.

ui/src/services/NodeVolumesUtils.js
```diff
@@ -168,13 +168,26 @@ export const getVolumeListData = createSelector(
...
+        return [];
+      } else {
+        // the node name exists, check if there is any volume belong to this node
+        nodeVolumes = volumes?.filter(
+          (volume) => volume.spec.nodeName === nodeFilter,
+        );
+        if (!nodeVolumes) {
+          return [];
+        }
+      }
```
gdemonet: That's excessively complex IMO, unless we really want to prevent the case where someone inputs a wrong node name both in the volume spec and in the URL... Otherwise, this would suffice:
```suggestion
      nodeVolumes = volumes?.filter(
        (volume) => volume.spec.nodeName === nodeFilter,
      );
```

ui/src/services/NodeVolumesUtils.js
```diff
@@ -168,13 +168,26 @@ export const getVolumeListData = createSelector(
...
-    let nodeVolumes = volumes;
-    // filter the volumes by the node name from URL
-    if (nodeName) {
-      nodeVolumes = volumes?.filter(
-        (volume) => volume.spec.nodeName === nodeName,
-      );
+    let nodeVolumes = null;
+
+    if (nodeFilter) {
+      // when there is node filter in URL, check if the node name is valid / exist
```
gdemonet: If this case is really important to consider, shouldn't we treat it more explicitly for the user? Otherwise, from a volume listing standpoint, it is the same to filter by a node which has no volume, or a non-existing node (since we just return an empty list).

ui/src/services/utils.js
```diff
@@ -285,7 +287,10 @@ export function addMissingDataPoint(
     orginalValues.length === 0 ||
     !startingTimeStamp ||
     !sampleDuration ||
-    !sampleFrequency
+    !sampleFrequency ||
+    // if the `sampleDuration` and `sampleFrequency` is not from the predefined value
+    !SAMPLE_DURATION_LIST.includes(sampleDuration) ||
+    !SAMPLE_FREQUENCY_LIST.includes(sampleFrequency)
```
gdemonet: Why add this behaviour? It only seems an artificial limitation to me (not showing an option in the visible UI doesn't mean your utility function can't do something).

gdemonet (changes requested):
Mostly comments around behaviour change and existing code, so you may relegate them to a debt ticket.

However, I think there's some effort to spend on:
- designing utility methods to limit their concern to single, simple tasks
- writing clear, minimal test cases, to make the tested behaviour obvious for a reader

Testing higher-order functions which interact with "real-life" datasets isn't really adapted to unit-testing (which is really about testing complex algorithmic behaviour), but could be worth writing integration tests for.

ui/src/ducks/app/monitoring.js
ChengYanJin: Yes! the commented query is the correct one actually. This is exactly the bug that #2757 fixed.

ui/src/services/NodeVolumesUtils.js
ChengYanJin: correct! It's more understandable!

ui/src/services/utils.js
ChengYanJin: For me, this function is **only** useful for the graph. And the choices of `sampleDuration` and `sampleFrequency` are quite limited (depends on the possible timespan). 
I add this check is to prevent the wrong inputs.

Does it make sense?

ui/src/services/NodeVolumesUtils.js
gdemonet: But it doesn't mean the same thing... So is `PORT_NUMBER_PROMETHEUS` actually a string containing `:9090`?

ui/src/services/utils.js
gdemonet: > For me, this function is **only** useful for the graph.

Mmmh. To me, this function is only **used** by the current charts we have, which only have some particular values as input. However, it could be useful for other charts using other sample sizes/steps, no?

> And the choices of sampleDuration and sampleFrequency are quite limited (depends on the possible timespan).
> I add this check is to prevent the wrong inputs.

I understand that we only give a user certain choices, through the selection of a timespan. But why would another choice be **wrong** for this utility function?

Let's take another example, coming from the real world. Imagine, a common tool called a screwdriver. Imagine you're building a screwdriver, so you can provide it to your users for, you guessed it... driving some screws. Now, let's say you only give 30mm, 40mm, and 50mm screws to your users, to use with your screwdriver. Does it mean the screwdriver **cannot** drive 60mm screws? Should the screwdriver break if it were given such a screw, by design?

ChengYanJin: Thanks for the explanation! 

I agree! Since it's a utility function we should not limit the choices. 

We can still check the properties `sampleFrequency` and `sampleDuration` if they're based on the choices we provided. But the checking should be at where we call the function.

ui/src/services/NodeVolumesUtils.js
ChengYanJin: yes.. originally the value of `PORT_NUMBER_PROMETHEUS` is `:9090` 
but yeah, it doesn't make any sense.

ui/src/services/NodeVolumesUtils.js
ChengYanJin: Absolutely right! 
Since we return an empty array for both cases, a node which has no volume and a non-existing node, I shouldn't implement such a complex logic here!

ui/src/ducks/app/monitoring.js
ChengYanJin: Fixed in #2757

ChengYanJin:
merged in PR #2779

Issue #2742, opened 2020-08-14
Replace various metalk8s-* storageclass by a single metalk8s one


alexandre-allard:
**Component**: salt, addons, docs

**Why this is needed**:
Today we have 1 metalk8s-prometheus (prometheus & alertmanager) and 1 metalk8s-loki (loki & fluent-bit) storageclass, but it does not make sense as the storageclass name must not be used to define the purpose of it.
Moreover, these class are exactly the same, so there is no point of keeping both.

**What should be done**:
Replace `metalk8s-prometheus` and `metalk8s-loki` storageclass everywhere by a `metalk8s` one.

**Implementation proposal** (strongly recommended):
Should be updated in examples, docs and salt

**Test plan**:
Tests should still work.

TeddyAndrieux:
NOTE: Right now in Loki chart you cannot provide any selector for volume so for the moment we consider all volume with storage class `metalk8s-loki` as Loki volume check this commit message https://github.com/scality/metalk8s/commit/97cd2fffce4a10c5019fa60b9e77eba93e6f6f51

PR #2755, opened 2020-08-24
Replace metalk8s-prometheus storageclass by a metalk8s one

alexandre-allard:
**Component**: salt, tests, docs

**Context**: See #2742 

**Summary**:
This PR only tackles Prometheus, another one will be done for Loki as it needs more than just renaming and it will only target `development/2.6`

**Acceptance criteria**:
Green build and tests

---

See: #2742

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -287,7 +287,6 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/prometheus-operator/deployed/namespace.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/deployed/',
             'service-configuration.sls'),
-    Path('salt/metalk8s/addons/prometheus-operator/deployed/storageclass.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/pre-downgrade.sls'),
     Path('salt/metalk8s/addons/prometheus-operator/pre-upgrade.sls'),
 
@@ -301,6 +300,8 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/addons/solutions/deployed/configmap.sls'),
     Path('salt/metalk8s/addons/solutions/deployed/init.sls'),
     Path('salt/metalk8s/addons/solutions/deployed/namespace.sls'),
+
+    Path('salt/metalk8s/addons/storageclass/deployed.sls'),
 
     Path('salt/metalk8s/addons/volumes/deployed.sls'),
     targets.TemplateFile(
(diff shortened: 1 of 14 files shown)


NicolasT:
Should you (or should you not, I didn't give it much thought) clean up the 'old' one as part of an upgrade? What about documentation?

alexandre-allard:
@NicolasT Good point for cleaning up the old storageclass, I'll do it.
The only part talking about the storageclass is in post-install which has been updated accordingly or are you talking about anything else ?

NicolasT:
Sorry, missed the docs change.

buildchain/buildchain/salt_tree.py
```diff
@@ -302,6 +301,8 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/addons/solutions/deployed/init.sls'),
     Path('salt/metalk8s/addons/solutions/deployed/namespace.sls'),
 
+    Path('salt/metalk8s/addons/storageclass.sls'),
```
TeddyAndrieux: To be consistent I would move it to `salt/metalk8s/addons/storageclass/deployed.sls`

salt/metalk8s/addons/prometheus-operator/post-upgrade.sls
```diff
@@ -2,3 +2,9 @@
 
 include:
   - .post-cleanup
+
+Delete old metalk8s-prometheus StorageClass:
+  metalk8s_kubernetes.object_absent:
+    - apiVersion: v1
+    - kind: StorageClass
+    - name: metalk8s-prometheus
```
TeddyAndrieux, line 10: If you remove a storageclass associated with some volumes does the volume still mounted and bound to the Pod (especially after a reboot ?)

I think we may want/need to replace `metalk8s-prometheus` by `metalk8s` on all volumes and all PVC, don't really know how our storage operator will react to that kind of change

slaperche-scality: > If you remove a storageclass associated with some volumes does the volume still mounted and bound to the Pod (especially after a reboot ?)

Good point.

I think we're good, because we don't care about the storage class once the Volume/PV are created (not sure about the ext pillar though).

To be confirmed by testing it.

alexandre-allard: I've already tested it, the volume will still be bound to the pod even if the storage class no longer exists (even across reboots).
We can change the storage class of a PV, but not a PVC.
We could maybe edit the PV and recreate the PVC, but I'm not sure that something we want to do as it works well without touching them.
As Sylvain said, we don't care about the storage class once the volume is created.

buildchain/buildchain/salt_tree.py
alexandre-allard: done

TeddyAndrieux (approved):

PR #2757, opened 2020-08-25
[ui] Bugfix the prometheus query to retrieve volume latency

### LINKED ISSUE #2756: Bugfix the Prometheus query to retrieve volume latency
**Component**:
ui, monitoring

**What happened**:
We didn't use the good query to retrieve the volume latency.

**What was expected**:
const volumeLatencyWriteQuery = `sum(irate(node_disk_write_time_seconds_total[5m])) by (instance, device) / sum(irate(node_disk_writes_completed_total[5m])) by (instance, device)`;
  const volumeLatencyReadQuery = `sum(irate(node_disk_read_time_seconds_total[5m])) by (instance, device) / sum(irate(node_disk_reads_completed_total[5m])) by (instance, device)`;

**Steps to reproduce**

ChengYanJin:
**Component**:  ui, monitoring

**Context**: 
The definition of disk latency is

>   Disk latency is the time that it takes to complete a single I/O operation on a block device

**Summary**:
  const volumeLatencyWriteQuery = 
`sum(irate(node_disk_write_time_seconds_total[5m]) / irate(node_disk_writes_completed_total[5m])) by (instance, device) * 1000000`;

  const volumeLatencyReadQuery = 
`sum(irate(node_disk_read_time_seconds_total[5m]) / irate(node_disk_reads_completed_total[5m])) by (instance, device) * 1000000`;
  
The latency in the volume table:
const volumeLatencyCurrentQuery = 
`irate(node_disk_io_time_seconds_total[1h]) * 1000000`;

 const volumeUsageQuery =
`kubelet_volume_stats_used_bytes / kubelet_volume_stats_capacity_bytes`;

**Acceptance criteria**: 

For loopdevice in Centos 7, the result of Prometheus is always **0**, so `query node_disk_writes_completed_total` and `node_disk_reads_completed_total` should return 0 which cause latency **NaN**, so no graph displayed.

 With warning:
```
WARN Infinite extent for field "read": [Infinity, -Infinity]
WARN Infinite extent for field "write": [Infinity, -Infinity]
```
**Please note that only the last 6 commits are belonging to this PR**

**Please give the approval in PR [#2779](https://github.com/scality/metalk8s/pull/2779)**

---

Closes: #2756


ui/src/services/NodeVolumesUtils.test.js
```diff
@@ -420,3 +422,96 @@ it('should return error when called with failed Ready condition', () => {
...
+  );
+  const volumelist_filtered_by_node_master0 = [
+    {
+      name: 'master-0-alertmanager',
+      node: 'master-0',
+      usage: '0.41',
+      status: 'link',
+      bound: 'Yes',
+      storageCapacity: '5Gi',
+      storageClass: 'metalk8s-prometheus',
```
alexandre-allard: All occurrences of `metalk8s-prometheus` storage class should probably be replaced by `metalk8s` according to the following PR #2755

ChengYanJin: Got it! Thanks for the sync up!

It will impact the Create volume E2E test!

ui/src/ducks/app/monitoring.js
```diff
@@ -330,22 +335,28 @@ export function* fetchVolumeStats() {
...
   // rate calculates the per-second average rate of increase of the time series in the range vector.
-  const volumeThroughputReadQuery = `rate(node_disk_read_bytes_total{job="node-exporter"}[1m])`;
-  const volumeThroughputWriteQuery = `rate(node_disk_written_bytes_total{job="node-exporter"}[1m])`;
-  const volumeLatencyQuery = `rate(node_disk_io_time_seconds_total{job="node-exporter"}[1m])`;
-  const volumeIOPSReadQuery = `irate(node_disk_reads_completed_total{job="node-exporter"}[5m])`;
-  const volumeIOPSWriteQuery = `irate(node_disk_writes_completed_total{job="node-exporter"}[5m])`;
-  const volumeUsedQuery = 'kubelet_volume_stats_used_bytes';
-
-  const volumeUsedQueryResult = yield call(
+  // group the result of the query by instance and device (remove the filter {job="node-exporter"})
```
slaperche-scality, line 343: the `(remove the filter {job="node-exporter"})` should be in a commit message, not in a comment here.

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -161,20 +156,26 @@ const MetricGraphCard = (props) => {
 
   // slot[0] => timestamp
   // slot[1] => value
-  const volumeUsageData = volumeUsedOperated?.map((slot) => {
+  const volumeUsageData = volumeUsageOperated?.map((slot) => {
     return {
       date: new Date(slot[0] * 1000), // convert from the RFC 3339 to time in JS
```
slaperche-scality: Instead of having a comment (which is not repeated l.176 and l.168), we could have a little function `dateFromRFC3339` that you could reuse everywhere it's needed.
And thanks to the explicit name, you won't need the comment.

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -161,20 +156,26 @@ const MetricGraphCard = (props) => {
...
+      y: slot[1] === null ? null : Math.round(slot[1] * 100),
     };
   });
 
-  const volumeLatencyData = volumeLatencyOperated?.map((slot) => {
+  const volumeLatencyWriteData = volumeLatencyWriteOperated?.map((slot) => {
     return {
       date: new Date(slot[0] * 1000),
-      y: slot[1] === null ? null : Math.round(slot[1] * 1000000),
+      write: slot[1] === null ? null : Math.round(slot[1] * 1000000),
```
slaperche-scality: ditto, here: instead of having a mysterious `1000000` value used for multiplication, we could hide the computation in a function `toGb` (or a better name :stuck_out_tongue: ) and reuse this function where it's needed.

ui/src/services/utils.js
```diff
@@ -280,7 +280,13 @@ export function addMissingDataPoint(
   sampleDuration,
   sampleFrequency,
 ) {
-  if (!orginalValues || orginalValues.length === 0) {
+  if (
```
slaperche-scality, line 283: could describe which edge cases are handled (in the commit message).

ui/src/containers/VolumePageContent.js
```diff
@@ -114,30 +114,30 @@ const VolumePageContent = (props) => {
   }
 
   const queryStartingTime = volumeStats?.queryStartingTime;
-  const volumeUsed = volumeStats?.volumeUsed?.filter(
+  const volumeUsed = volumeStats?.volumeUsed?.find(
     (vU) => vU.metric.persistentvolumeclaim === PVCName,
-  );
-  const volumeThroughputWrite = volumeStats?.volumeThroughputWrite?.filter(
+  )?.values;
+  const volumeThroughputWrite = volumeStats?.volumeThroughputWrite?.find(
```
slaperche-scality, line 119: why the change from `filter` to `find`?

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -139,8 +139,11 @@ const MetricGraphCard = (props) => {
   const volumeUsedOperated = operateMetricRawData(
     volumeMetricGraphData?.volumeUsed,
   );
-  const volumeLatencyOperated = operateMetricRawData(
-    volumeMetricGraphData?.volumeLatency,
+  const volumeLatencyWriteOperated = operateMetricRawData(
```
slaperche-scality, line 147: Commit message is not very explicit about what was fixed.

You should describe what was the problem with the current code and why the new code address the problem.

ui/src/containers/VolumePageContent.js
ChengYanJin: Since we add the aggregation in the Prometheus query, we have **only one** candidate in the array. 
So `find` is a little bit more efficient than `filter`. Because once it matches the value in findings, it will not check the remaining values in the array collection.

ui/src/components/VolumeMetricGraphCard.js
ChengYanJin: Thanks for the idea! 
The code is more understandable!

ui/src/services/utils.js
ChengYanJin: hi @slaperche-scality, actually it's more props checking not edge cases, I reword my commit message

ui/src/containers/VolumePageContent.js
slaperche-scality: Thanks for the explanation :+1: 
This is the kind of detail than can be explained in a commit message :slightly_smiling_face:

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -161,52 +161,59 @@ const MetricGraphCard = (props) => {
...
+
+  const volumeLatencyWriteData = volumeLatencyWriteOperated?.map((slot) => {
     return {
-      date: new Date(slot[0] * 1000), // convert from the RFC 3339 to time in JS
-      y:
-        slot[1] === null
-          ? null
-          : Math.round((slot[1] / volumeStorageCapacity) * 100),
+      date: fromUnixTimestampToDate(slot[0]),
+      write: slot[1] === null ? null : Math.round(toMicro(slot[1])),
```
slaperche-scality: I fact, I think we can ask Prometheus to do the conversion instead of calling `toMicro`.

For instance, using this query:

```javascript
  const volumeLatencyWriteQuery = `sum(irate(node_disk_write_time_seconds_total[5m]) / irate(node_disk_writes_completed_total[5m])) by (instance, device) * 1000000`;
```

To be tested, but I think it works (and if so, same logic can be used for `toMega`).
IMHO, we should let Prometheus do as much work as possible.

gdemonet: > we should let Prometheus do as much work as possible

Should we? Why add load on a critical server-side component when a user's computer is able to do the computation without too much effort? What about the change of scale from the UI (not in scope for now, but wouldn't be absurd)?

slaperche-scality: Because it's Prometheus purpose to deal with those data efficiently (it's a database), so I expect it to be more optimized for this kind of task than some JS code in the browser.
I don't think the multiplication is an heavy operation for Prometheus (otherwise I would be really worried).

If we don't want to load Prometheus, than we should simply fetch raw data and do everything (sampling, aggregation, …) with JS code, no?

> What about the change of scale from the UI (not in scope for now, but wouldn't be absurd)?

The data on the chart are already refreshed regularly, so we would simply update the query we use on the fly.
This won't require extra requests.

gdemonet: Thanks for the explanation, it makes sense indeed. More than a database (some of them are not built with the same purpose in mind), Prometheus is optimized for distributed queries with advanced computations such as the ones you mentioned, so let's use it :+1:!

PR #2760, opened 2020-08-26
Replace metalk8s-loki storageclass with metalk8s one

### LINKED ISSUE #2742: Replace various metalk8s-* storageclass by a single metalk8s one

alexandre-allard:
**Component**: salt, docs, charts

**Context**: See #2742

**Summary**:
* Update Loki helm chart to support selector for PVC (also opened an upstream PR https://github.com/grafana/loki/pull/2554)
* Replace the `metalk8s-loki` storageclass with `metalk8s` one
* Raise retries for logging tests.

**Acceptance criteria**: 

---

Closes: #2742


charts/loki/templates/statefulset.yaml
```diff
@@ -124,5 +124,9 @@ spec:
         requests:
           storage: {{ .Values.persistence.size | quote }}
       storageClassName: {{ .Values.persistence.storageClassName }}
+      {{- if .Values.persistence.selector }}
```
Ebaneck, line 127: Hopefully, we don't forget this local patch during Loki chart upgrades...

Ebaneck (approved):
LGTM

charts/loki/templates/statefulset.yaml
alexandre-allard: That's why I opened an upstream PR, I hope it'll be merged.

Ebaneck: Ha, I see the commit message now :+1:

PR #2763, opened 2020-08-27
Reconfigure MetalK8s authn to support external backends(Keycloak)

Ebaneck:
**Component**:

'salt', 'dex', 'charts', 'oidc'

**Context**: 

See: [ticket](https://scality.atlassian.net/browse/RINGX-18)

**Summary**:

- Reconfigure Dex to add support for any external Identity Provider
- Add documentation on how to edit the CSC that handles authn config params

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/charts/dex.yaml
+++ b/charts/dex.yaml
@@ -79,6 +79,7 @@ config:
     trustedPeers:
     - metalk8s-ui
     - grafana-ui
+    - external-oidc-ui
   - id: metalk8s-ui
     redirectURIs:
     - '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/oauth2/callback)'
@@ -89,6 +90,10 @@ config:
     redirectURIs:
     - '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth)'
     secret: "4lqK98NcsWG5qBRHJUqYM1"
+  - id: external-oidc-ui
+    name: '__var__(dex.spec.externalIDP.staticClient.name)'
+    redirectURIs: '__var_tojson_nospace__(dex.spec.externalIDP.staticClient.redirectURIs)'
+    secret: "ZXhhbXBsZS1hcHAtc2VjcmV0" #Secret cannot be put in a CSC ConfigMap
 
   enablePasswordDB: '__var__(dex.spec.localuserstore.enabled)'
 
--- a/salt/metalk8s/addons/dex/deployed/chart.sls
+++ b/salt/metalk8s/addons/dex/deployed/chart.sls
@@ -50,6 +50,7 @@ stringData:
       trustedPeers:
       - metalk8s-ui
       - grafana-ui
+      - external-oidc-ui
     - id: metalk8s-ui
       name: MetalK8s UI
       redirectURIs:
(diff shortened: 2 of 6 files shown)


charts/dex.yaml
```diff
@@ -89,6 +90,11 @@ config:
     redirectURIs:
     - '__escape__(https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth)'
     secret: "4lqK98NcsWG5qBRHJUqYM1"
+  - id: external-oidc-ui
+    name: '__var__(dex.spec.externalIDP.staticClient.name)'
+    redirectURIs:
+    - '__var__(dex.spec.externalIDP.staticClient.redirectURIs)'
```
TeddyAndrieux: To me `redirectURIs` is a list of URI so it's should NOT be in a list here 
```suggestion
    redirectURIs: '__var__(dex.spec.externalIDP.staticClient.redirectURIs)'
```

TeddyAndrieux: Also here since it's a list we want/need to put a `| tojson` filter to this one

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -60,6 +61,10 @@ stringData:
       redirectURIs:
       - "{% endraw -%}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%- raw %}"
       secret: 4lqK98NcsWG5qBRHJUqYM1
+    - id: external-oidc-ui
+      name: {% endraw -%}{{ dex.spec.externalIDP.staticClient.name }}{%- raw %}
+      redirectURIs:   {% endraw -%}{{ dex.spec.externalIDP.staticClient.redirectURIs | tojson }}{%- raw %}
```
TeddyAndrieux: Weird those triple space, but never mind it's not related to change of this PR

```suggestion
      redirectURIs: {% endraw -%}{{ dex.spec.externalIDP.staticClient.redirectURIs | tojson }}{%- raw %}
```

Ebaneck: Likely from the render script.
Let me fix it

Ebaneck: ~~Going to backport the changes maybe later!~~

salt/metalk8s/addons/dex/deployed/chart.sls
```diff
@@ -60,9 +61,13 @@ stringData:
...
       - "{% endraw -%}https://{{ grains.metalk8s.control_plane_ip }}:8443/grafana/login/generic_oauth{%- raw %}"
       secret: 4lqK98NcsWG5qBRHJUqYM1
+    - id: external-oidc-ui
+      name: {% endraw -%}{{ dex.spec.externalIDP.staticClient.name }}{%- raw %}
+      redirectURIs: {% endraw -%}{{ dex.spec.externalIDP.staticClient.redirectURIs | tojson }}{%- raw %}
+      secret: ZXhhbXBsZS1hcHAtc2VjcmV0
     enablePasswordDB: {% endraw -%}{{ dex.spec.localuserstore.enabled }}{%- raw %}
     staticPasswords:
-      {% endraw -%}{{ dex.spec.localuserstore.userlist | tojson }}{%- raw %}
+    {% endraw -%}{{ dex.spec.localuserstore.userlist | tojson }}{%- raw %}
```
TeddyAndrieux: Not sure it will work like this right now

gdemonet:
Two major comments (we can't merge as-is):

- This is adding a **client** to Dex, not a **connector**: it's like allowing another IdP (such as Keycloak) to federate Dex users, which is the opposite of what we want
- Adding clients is a good feature (in case someone wants to add another control-plane UI, e.g. what we'd do for Solutions), but the implementation is wrong IMO: this PR hardcodes a single client, to which we can only add redirect URIs - we should instead allow to configure an arbitrary amount of clients to Dex, and let these clients configure their own secrets

On the topic of secrets, I don't think we should keep hardcoded secrets for our own clients anyway (they are open-sourced, hence vulnerable). Let's take a step back and rethink this PR offline.

gdemonet:
Let's close this in favor of #2781 if that's OK for you @Ebaneck

Ebaneck:
Closing due to https://github.com/scality/metalk8s/pull/2763#issuecomment-688694805

PR #2764, opened 2020-08-28
[ui] Add delete volume for sparseloop device E2E test

### LINKED ISSUE #2762: Add Delete Volume E2E Test
**Component**: ui, tests

**Why this is needed**:
We should test the delete volume from UI is correctly done.

**What should be done**:
Add the delete volume E2E test

We should delete the `sparseloopdevice` which is just created in the `CreateVolume.feature`

ChengYanJin:
**Component**: ui, tests

**Context**: 
We want to add a delete volume E2E test for Sparseloop device. 

**Summary**:
After creating the Sparseloop device, we delete it. And check if this volume disappears from the volume list.

**Acceptance criteria**: 
Pass the E2E test

locally test with this command:
`./node_modules/cypress/bin/cypress run -b chrome -s cypress/integration/common/CreateDeleteVolume.feature --no-exit`

Note that please following code would not work:
`cy.get('[data-cy=volume_table]').contains(volumeName).should(...)`
According to [Only the last command is retried](https://docs.cypress.io/guides/core-concepts/retry-ability.html#Only-the-last-command-is-retried), the cy.get() will not retry. 

We should use only one `cy.contains()` instead of
```
  cy.contains('[data-cy=volume_table_name_cell]', volumeName, {
    timeout: 120000,
  }).should('not.exist');
```
---

Closes: #2762

### DIFF
--- a/ui/cypress/integration/common/CreateDeleteVolume.feature
+++ b/ui/cypress/integration/common/CreateDeleteVolume.feature
@@ -14,3 +14,6 @@ Feature: Create and delete volume scenario
     And the label of volume "test-volume-sparse" presents:
         | labelName  | kubernetest.io/name |
         | labelValue | test                |
+    When I click [Delete] button
+    And I confirm the deletion
+    Then the "test-volume-sparse" volume is removed from the list
--- a/ui/cypress/integration/common/CreateDeleteVolume.feature
+++ b/ui/cypress/integration/common/CreateDeleteVolume.feature
@@ -6,13 +6,13 @@ Feature: Create and delete volume scenario
         | name       | test-volume-sparse  |
         | type       | sparseLoopDevice    |
         | size       | 1 GiB               |
-        | labelName  | kubernetest.io/name |
+        | labelName  | kubernetes.io/name  |
         | labelValue | test                |
     And I click [Create] button
     Then I am redirected to the "test-volume-sparse" volume page
     And the volume "test-volume-sparse" becomes Ready
-    And the label of volume "test-volume-sparse" presents:
-        | labelName  | kubernetest.io/name |
+    And the labels the volume include:
+        | labelName  | kubernetes.io/name  |
         | labelValue | test                |
     When I click [Delete] button
     And I confirm the deletion
--- a/ui/cypress/integration/common/CreateDeleteVolume.feature
+++ b/ui/cypress/integration/common/CreateDeleteVolume.feature
@@ -1,0 +1,16 @@
+Feature: Create and delete volume scenario
+  Scenario: I can create a SparseLoopDevice volume
+    Given I am logged in
+    And I am on the volume creation page
+    When I fill out the volume creation form using:
+        | name       | test-volume-sparse  |
+        | type       | sparseLoopDevice    |
+        | size       | 1 GiB               |
+        | labelName  | kubernetest.io/name |
+        | labelValue | test                |
+    And I click [Create] button
+    Then I am redirected to the "test-volume-sparse" volume page
+    And the volume "test-volume-sparse" becomes Ready
+    And the label of volume "test-volume-sparse" presents:
+        | labelName  | kubernetest.io/name |
+        | labelValue | test                |
--- a/ui/cypress/integration/common/CreateVolume.feature
+++ b/ui/cypress/integration/common/CreateVolume.feature
@@ -1,6 +1,0 @@
-Feature: CreateVolume
- Scenario: CreateVolume with SparseLoopDevice
-     Given I log in
-     When I go to the volume page by click the volume icon in the sidebar
-     And I go to create volume page by click on Create a New Volume button
-     Then I fill out the create volume form with SparseLoopDevice volume type and check if the status is Ready
--- a/ui/cypress/integration/common/volume.js
+++ b/ui/cypress/integration/common/volume.js
@@ -1,16 +1,5 @@
 import { And, Then, When } from 'cypress-cucumber-preprocessor/steps';
 
-And(
-  'I go to the bootstrap node by click on the bootstrap row in the list',
-  () => {
-    cy.get('.sc-table-row').eq(1).click();
-  },
-);
-
-And('I choose the Volumes tag', () => {
-  cy.get('.sc-tabs-bar .sc-tabs-item-title').eq(1).click();
-});
-
 And('I am on the volume creation page', () => {
   cy.get('.sc-sidebar-item').eq(2).click();
   cy.get('[data-cy="create-volume-button"]').click();
--- a/ui/cypress/integration/common/volume.js
+++ b/ui/cypress/integration/common/volume.js
@@ -11,68 +11,98 @@ And('I choose the Volumes tag', () => {
   cy.get('.sc-tabs-bar .sc-tabs-item-title').eq(1).click();
 });
 
-And('I go to create volume page by click on Create a New Volume button', () => {
+And('I am on the volume creation page', () => {
+  cy.get('.sc-sidebar-item').eq(2).click();
   cy.get('[data-cy="create-volume-button"]').click();
 });
 
-When('I go to the volume page by click the volume icon in the sidebar', () => {
-  cy.get('.sc-sidebar-item').eq(2).click();
+When('I fill out the volume creation form using:', (dataTable) => {
+  cy.route('GET', '/api/kubernetes/api/v1/persistentvolumes').as(
+    'getPersistentVolumes',
+  );
+
+  // rowsHash: returns an object where each row corresponds to an entry (first column is the key, second column is the value).
+  const dataTableObject = dataTable.rowsHash();
+
+  const volumeName = dataTableObject.name;
+  const volumeType = dataTableObject.type;
+  const volumeSize = dataTableObject.size.split(' ')[0];
+  const volumeLabelName = dataTableObject.labelName;
+  const volumeLabelValue = dataTableObject.labelValue;
+
+  cy.get('input[name=name]').type(volumeName);
+  cy.get('input[name=labelName]').type(volumeLabelName);
+  cy.get('input[name=labelValue]').type(volumeLabelValue);
+  cy.get('[data-cy=add-volume-labels-button]').click();
+
+  // node name
+  cy.get('.sc-select').eq(0).click();
+  cy.get('.sc-select__menu').eq(0).click();
+
+  cy.get('.sc-select').eq(1).click();
+  cy.get('.sc-select__menu').find('[data-cy=storageClass-metalk8s]').click();
+
+  cy.get('.sc-select').eq(2).click();
+  cy.get('.sc-select__menu').find(`[data-cy="type-${volumeType}"]`).click();
+
+  cy.get('input[name=sizeInput]').type(volumeSize);
 });
 
-Then(
-  'I fill out the create volume form with SparseLoopDevice volume type and check if the status is Ready',
-  function () {
-    cy.route('GET', '/api/kubernetes/api/v1/persistentvolumes').as(
-      'getPersistentVolumes',
-    );
-    const volumeNameSparseLoopDevice = `volume-${new Date().getTime()}`;
-    const volumeCapacity = Cypress.env('volume_capacity');
-    const volume_label_name = Cypress.env('volume_label_name');
-    const volume_label_value = Cypress.env('volume_label_value');
+And('I click [Create] button', () => {
+  cy.get('[data-cy="submit-create-volume"]').click();
+});
 
-    cy.get('input[name=name]').type(volumeNameSparseLoopDevice);
+Then('I am redirected to the "test-volume-sparse" volume page', () => {
+  cy.location('pathname').should('eq', '/volumes/test-volume-sparse');
+});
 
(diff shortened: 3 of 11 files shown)


ui/cypress/integration/common/CreateVolume.feature
```diff
@@ -0,0 +1,6 @@
+Feature: CreateVolume
+ Scenario: CreateVolume with SparseLoopDevice
+     Given I log in
```
gdemonet: When using Gherkin syntax, we try to describe `Given`s as a _state we are in_, not as an action to perform.

```suggestion
     Given I am logged in
```

ui/cypress/integration/common/CreateVolume.feature
```diff
@@ -0,0 +1,6 @@
...
+     When I go to the volume page by click the volume icon in the sidebar
+     And I go to create volume page by click on Create a New Volume button
```
gdemonet: This scenario is testing the creation of a volume, not the navigation to this page (or is it? in which case, I'd argue we'd better split the scenario into one for navigation, and another for creation).

As such, we could sum it up by another `Given`:

```gherkin
    Given I am logged in
    And I am on the volume creation page
```

ui/cypress/integration/common/CreateVolume.feature
```diff
@@ -0,0 +1,6 @@
...
+     Then I fill out the create volume form with SparseLoopDevice volume type and check if the status is Ready
```
gdemonet: Here, we are doing more than a simple `Then`. Here's how BDD works:

- first, make sure you are in a **given** state - with `Given` steps
- then, perform some action(s) **which you want to test** - with `When` step(s)
- finally, check that you are in the **expected** state - with `Then` steps

So, in our case, the scenario could look like:

```gherkin
  Scenario: I can create a SparseLoopDevice volume
    Given I am logged in
    And I am on the volume creation page
    When I fill out the volume creation form using:
        | name | test-volume-sparse |
        | type | SparseLoopDevice   |
        | size | 1Gi                |
    And I click Create
    Then I am redirected to the volume list page
    And the volume "test-volume-sparse" becomes Ready
```

Note that for passing detailed info to the first `When`, we can rely on "data-tables" (see https://cucumber.io/docs/gherkin/reference/#data-tables, and https://github.com/TheBrainFamily/cypress-cucumber-preprocessor#data-table-parameters)

ui/src/components/VolumeListTable.js
```diff
@@ -282,6 +282,13 @@ function Table({ columns, data, nodeName, rowClicked, volumeName, theme }) {
                         <div>{intl.translate('unknown')}</div>
                       </Cell>
                     );
+                  } else if (cell.column.Header === 'Name') {
+                    // add data-cy for delete volume E2E test
+                    return (
+                      <Cell {...cellProps} data-cy="table_cell_volume_name">
```
gdemonet: IIUC, you're using this to count the number of rows. So, why not count `TableRow`s instead of `Cell`s?

ui/cypress/integration/common/CreateDeleteVolume.feature
```diff
@@ -4,3 +4,5 @@ Feature: CreateVolume
      When I go to the volume page by click the volume icon in the sidebar
      And I go to create volume page by click on Create a New Volume button
      Then I fill out the create volume form with SparseLoopDevice volume type and check if the status is Ready
+     And I delete the volume by click on the delete volume button and confirm the deletion
```
gdemonet: Following on the previous comments, here's how the rest of the scenario would look like:

```gherkin
    # [...]
    And the volume "test-volume-sparse" becomes Ready
    When I select the "test-volume-sparse" volume
    And I click Delete
    And I confirm the deletion
    Then no volume is selected
    And the "test-volume-sparse" volume is removed from the list
```

ui/cypress/integration/common/volume.js
```diff
@@ -68,11 +70,42 @@ Then(
       },
     );
 
+    // Check if the right side panel updates
```
gdemonet: That should be a single `Then` step

ui/cypress/integration/common/volume.js
```diff
@@ -68,11 +70,42 @@ Then(
...
+    cy.get('[data-cy="volume_detail_card_name"]').should(
+      'contain',
+      volumeNameSparseLoopDevice,
+    );
     // Check if all the labels are presents
```
gdemonet: This too

ui/cypress/integration/common/volume.js
```diff
@@ -68,11 +70,42 @@ Then(
...
       'contain',
       volume_label_value,
     );
   },
 );
+
+And(
+  'I delete the volume by click on the delete volume button and confirm the deletion',
+  () => {
+    cy.get('[data-cy="delete_volume_button"]').click({ force: true });
```
gdemonet: How does this work if "_the volume_" is not selected? What volume are we talking about?

ui/cypress/integration/common/volume.js
```diff
@@ -68,11 +70,42 @@ Then(
...
+    cy.get('[data-cy="confirm_deletion_button"]').click();
+  },
+);
+
+Then('I check the volume disappear from the volume list', () => {
```
gdemonet: You would be better off passing a specific volume name to all your steps, and checking it's not part of the list anymore. This check is too error-prone (only works when stars are aligned).

gdemonet (changes requested):
Some issues with how we're using BDD here. I can only recommend you spend some time reading through Cucumber documentation (start [here](https://cucumber.io/docs/guides/overview/)) to get a grasp of why and how BDD is done :)

ui/cypress/integration/common/CreateVolume.feature
ChengYanJin: It's good to know that we can rely on "data-tables" to pass the props!!! It's much easier to read😇

ui/cypress/integration/common/volume.js
ChengYanJin: Actually after creating the volume "test-volume-sparse", I already implement the action "Click on this volume" in the code. 

So we are already in the volume "test-volume-sparse" detail on the right-side panel. 

But I encountered an error return by cypress here that the `delete-volume-button` was covered by another element. That's why I add `.click({ force: true })` here.

ui/cypress/integration/common/volume.js
ChengYanJin: Yes. I agree. 
I didn't figure out a way to check that this row is not part of the list anymore. Maybe we need to traverse the returned JQuery object here.
```
  cy.waitUntil(
    () =>
      cy.get('[data-cy="volume_table_row"]').then(($el) => $el.length === 3),
    {
      errorMsg: `Volume test-volume-sparse is not deleted`,
      timeout: 120000, // waits up to 120000 ms, default to 5000
      interval: 5000, // performs the check every 5000 ms, default to 200
    },
  );
```

ui/cypress/integration/common/CreateDeleteVolume.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Create and delete volume scenario
+  Scenario: I can create a SparseLoopDevice volume
+    Given I am logged in
+    And I am on the volume creation page
+    When I fill out the volume creation form using:
+        | name       | test-volume-sparse  |
+        | type       | sparseLoopDevice    |
+        | size       | 1 GiB               |
+        | labelName  | kubernetest.io/name |
```
gdemonet: `kubernetest.io` :rofl:

ui/cypress/integration/common/volume.js
```diff
@@ -11,68 +11,98 @@ And('I choose the Volumes tag', () => {
...
+  cy.route('GET', '/api/kubernetes/api/v1/persistentvolumes').as(
+    'getPersistentVolumes',
+  );
+
+  // rowsHash: returns an object where each row corresponds to an entry (first column is the key, second column is the value).
+  const dataTableObject = dataTable.rowsHash();
+
+  const volumeName = dataTableObject.name;
+  const volumeType = dataTableObject.type;
+  const volumeSize = dataTableObject.size.split(' ')[0];
```
gdemonet, line 14: Shouldn't we also select the unit?

ui/cypress/integration/common/volume.js
```diff
@@ -11,68 +11,98 @@ And('I choose the Volumes tag', () => {
...
-      'contain',
-      volume_label_value,
-    );
-  },
-);
+And('I click [Create] button', () => {
+  cy.get('[data-cy="submit-create-volume"]').click();
+});
+
+Then('I am redirected to the "test-volume-sparse" volume page', () => {
```
gdemonet: This one would ideally allow arbitrary names (so it's reusable for other tests). [Reference](https://github.com/TheBrainFamily/cypress-cucumber-preprocessor#reusable-step-definitions)

```suggestion
Then('I am redirected to the "{string}" volume page', (volumeName) => {
```

ui/cypress/integration/common/volume.js
```diff
@@ -11,68 +11,98 @@ And('I choose the Volumes tag', () => {
...
+  cy.location('pathname').should('eq', '/volumes/test-volume-sparse');
+});
+
+And('the volume "test-volume-sparse" becomes Ready', () => {
```
gdemonet: Same here

ui/cypress/integration/common/CreateDeleteVolume.feature
```diff
@@ -0,0 +1,16 @@
...
+        | labelValue | test                |
+    And I click [Create] button
+    Then I am redirected to the "test-volume-sparse" volume page
+    And the volume "test-volume-sparse" becomes Ready
+    And the label of volume "test-volume-sparse" presents:
```
gdemonet:
```suggestion
    And the labels of volume "test-volume-sparse" include:
```

ui/cypress/integration/common/volume.js
```diff
@@ -11,68 +11,98 @@ And('I choose the Volumes tag', () => {
...
+  // fiter delete this volume, we should only have 3 default volume in CI.
+  cy.waitUntil(
+    () =>
+      cy.get('[data-cy="volume_table_row"]').then(($el) => $el.length === 3),
+    {
+      errorMsg: `Volume test-volume-sparse is not deleted`,
+      timeout: 120000, // waits up to 120000 ms, default to 5000
+      interval: 5000, // performs the check every 5000 ms, default to 200
+    },
+  );
```
gdemonet: You should be able to search through the found elements here:

```javascript
  cy.waitUntil(
    () =>
      cy.get('[data-cy="volume_table_row"]').then(($rows) => {
        $rows.forEach(($row) => {
          nameCell = // some query on $row children
          expect(nameCell).to.not.contain('test-volume-sparse')
        )
      }),
    {
      errorMsg: `Volume test-volume-sparse is not deleted`,
      timeout: 120000, // waits up to 120000 ms, default to 5000
      interval: 5000, // performs the check every 5000 ms, default to 200
    },
  );
```

But, more simply (to be tested), it seems the [`contains`](https://docs.cypress.io/api/commands/contains.html) command can search deep in the DOM, so this would probably be the simplest:

```suggestion
  cy.waitUntil(
    () =>
      cy.contains('[data-cy="volume_table_row"]', 'test-volume-sparse')
      	.should('not.exist'),
    {
      errorMsg: `Volume test-volume-sparse is not deleted`,
      timeout: 120000, // waits up to 120000 ms, default to 5000
      interval: 5000, // performs the check every 5000 ms, default to 200
    },
  );
```

gdemonet (changes requested):
Good job, this looks much more like BDD now! You will just need some time to get proficient with Cypress functionalities to do exactly what you want, and we'll have so many tests it's going to be awesome :partying_face:

ui/cypress/integration/common/volume.js
ChengYanJin: Definitely yes!!!

ui/cypress/integration/common/volume.js
ChengYanJin: Hello @gdemonet 
Note that: you cannot put assertions inside `checkFunction` (https://github.com/NoriSte/cypress-wait-until)
> In Cypress, a failing assertion forces the whole test to fail and you cannot restore it

I am afraid neither of them could work, so I didn't find other solutions than check there are 3 volume in the table

gdemonet: Ah, right, waiting is built-in many Cypress commands actually, so you don't need `waitUntil` at all: https://docs.cypress.io/api/commands/should.html#Wait-until-the-assertions-pass

ui/cypress/integration/common/volume.js
gdemonet: Let's track this one as debt :wink: we'll tackle it when we implement a larger set of tests with mocked backends!

gdemonet (approved):

ui/cypress/integration/common/volume.js
ChengYanJin: Hello @gdemonet,
Yes, IIUC, we need to have more `interaction tests` to check the reaction of form?

PR #2770, opened 2020-09-04
Add little procedure to upgrade MetalK8s platform with latest dev build

thomasdanan:
**Component**:

'doc'

**Context**: 

We have a standard procedure to upgrade to one MetalK8s version to another.
During the dev cycle, it may be interested to upgrade a cluster (which is already in latest version) to the latest dev build, on the same version.

**Summary**:

In this PR, we describe how to do this kind of upgrade

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/docs/developer/deploy/index.rst
+++ b/docs/developer/deploy/index.rst
@@ -1,0 +1,6 @@
+Deploy new MetalK8s image
+=========================
+
+.. toctree::
+
+   upgrade
--- a/docs/developer/deploy/upgrade.rst
+++ b/docs/developer/deploy/upgrade.rst
@@ -1,0 +1,60 @@
+Upgrade a Platform with the latest dev build
+============================================
+
+Prerequisites
+-------------
+- A Platform already installed with all pod up & running
+- A new metalk8s.iso image with a higher version
+
+Procedure
+---------
+
+If upgrading from a lower patch, minor or major version just follow the
+:doc:`standard upgrade procedure </operation/upgrade>`
+
+If upgrading from the same patch, minor and major version:
+
+1. Upload the new metalk8s.iso on the bootstrap node
+
+2. Locate how metalk8s.iso is mounted
+
+  .. parsed-literal::
+
+    grep metalk8s-|version| /etc/fstab
+    /root/metalk8s.iso		/srv/scality/metalk8s-|version|	iso9660	nofail,ro	0 0
+
+3. Unmount the current metalk8s.iso
+
+  .. parsed-literal::
+
+    umount /srv/scality/metalk8s-|version|
+
+4. Copy the new metalk8s.iso in place of the old one
+
+  .. parsed-literal::
+
+    cp metalk8s.iso /root/metalk8s.iso
+
+5. Mount the new metalk8s.iso
+
+  .. parsed-literal::
+
+    mount /root/metalk8s.iso /srv/scality/metalk8s-|version|
+
+6. Stop salt-master container on bootstrap node
+
+  .. parsed-literal::
+
+    crictl stop $(crictl ps -q --label io.kubernetes.container.name=salt-master --state Running)
+
+7. Provision the new metalk8s ISO content
+
+  .. parsed-literal::
+
+    /srv/scality/metalk8s-|version|/iso-manager.sh
+
+8. Upgrade the cluster
+
+  .. parsed-literal::
+
+    /srv/scality/metalk8s-|version|-dev/upgrade.sh --destination-version |version|
--- a/docs/developer/index.rst
+++ b/docs/developer/index.rst
@@ -6,5 +6,6 @@ Developer Guide
    architecture/index
    building/index
    running/index
+   deploy/index
    development/index
    solutions/index


docs/developer/deploy/upgrade.rst
```diff
@@ -0,0 +1,57 @@
...
+
+  .. code-block:: console
+
+    [root@zenkotda-bootstrap centos]# crictl ps | egrep 'ID|salt-master'
+    CONTAINER ID        IMAGE               CREATED             STATE               NAME                       ATTEMPT             POD ID
+    b9a78b65322c0       c09887cdd144d       20 hours ago        Running             salt-master                3                   66f3c727f5b53
+
+7. Stop the salt-master container
+
+  ``crictl stop b9a78b65322c0``
```
Ebaneck: I think you can combine steps 6 and 7 with something like

6. From the Bootstrap node, stop the salt-master container

``salt_master=$(crictl ps -q --label io.kubernetes.container.name=salt-master --state Running) | crictl stop $salt_master``

docs/developer/deploy/upgrade.rst
```diff
@@ -0,0 +1,57 @@
...
+
+8. Load the new metalk8s iso content
```
Ebaneck:
```suggestion
8. Provision the new MetalK8s ISO content
```
?

docs/developer/deploy/upgrade.rst
```diff
@@ -0,0 +1,57 @@
...
+2. Locate how metalk8s.iso is mounted
+
+  .. code-block:: console
+
+    [root@zenkotda-bootstrap centos]# cat /etc/fstab | grep metalk8s.iso
+    /root/metalk8s.iso		/srv/scality/metalk8s-2.6.0-dev	iso9660	nofail,ro	0 0
+
+3. Unmount the current metalk8s.iso
+
+  ``umount /srv/scality/metalk8s-2.6.0-dev``
```
Ebaneck: I am not sure if it is really important to place this in a sphinx parse-literal so that the version prefix is updated for next MetalK8s 2.7 doc or so 

I mean something like:
```
.. parsed-literal::

   umount /srv/scality/metalk8s-|version|

```

docs/developer/deploy/upgrade.rst
gdemonet: Proposed syntax is using a redirection (with the pipe `|`) which isn't doing anything here.
I think what Claude meant is more something like:

```shell
salt_master=$(crictl ps -q --label io.kubernetes.container.name=salt-master --state Running); crictl stop $salt_master
```

More simply (no need to assign a variable, unless we really want to check if it's empty, but that's not done here anyway):

```shell
crictl stop $(crictl ps -q --label io.kubernetes.container.name=salt-master --state Running)
```

docs/developer/deploy/upgrade.rst
```diff
@@ -0,0 +1,57 @@
...
+Prerequisites
+-------------
+- A Platform already installed with all pod up & running
+- A new metalk8s.iso image with a higher version
+
+Procedure
+---------
+
+If upgrading from a lower patch, minor or major version just follow the
+`standard upgrade procedure <https://metal-k8s.readthedocs.io/en/latest/operation/upgrade.html>`_.
```
gdemonet: This page lives in the same documentation, so we should use relative references instead of plain HTTP hyperlinks:

```suggestion
:doc:`standard upgrade procedure </operation/upgrade>`.
```

docs/developer/deploy/upgrade.rst
```diff
@@ -0,0 +1,57 @@
...
+If upgrading from the same patch, minor and major version:
+
+1. Upload the new metalk8s.iso on the bootstrap node
+
+2. Locate how metalk8s.iso is mounted
+
+  .. code-block:: console
+
+    [root@zenkotda-bootstrap centos]# cat /etc/fstab | grep metalk8s.iso
+    /root/metalk8s.iso		/srv/scality/metalk8s-2.6.0-dev	iso9660	nofail,ro	0 0
```
gdemonet: This assumes that the "current ISO" is called `metalk8s.iso`, which isn't necessary. A safer check could be `grep /srv/scality/metal8s-|version| /etc/fstab` (no need to `cat <file> | grep <pattern>`, `grep <pattern> <file>` is sufficient).

gdemonet:
Thanks for adding this little procedure @thomasdanan! This will surely help some developers trying to maintain a dev platform running (we could even consider adding it in our buildchain to safely upgrade a local Vagrant setup).

docs/developer/deploy/upgrade.rst
thomasdanan: done

docs/developer/deploy/upgrade.rst
thomasdanan: done

docs/developer/deploy/upgrade.rst
thomasdanan: done

docs/developer/deploy/upgrade.rst
thomasdanan: done

docs/developer/deploy/upgrade.rst
thomasdanan: done

docs/developer/deploy/index.rst
```diff
@@ -0,0 +1,6 @@
+Deploy new MetalK8s image
```
slaperche-scality, line 1: Typo in the commit message: `s/Mtelak8s/Metalk8s/`

thomasdanan: done

gdemonet (approved):
Thank you!

PR #2779, opened 2020-09-07
[UI] Bugfix the prometheus queries for volume latency (to be merged)

### LINKED ISSUE #2756: Bugfix the Prometheus query to retrieve volume latency

ChengYanJin:
**Component**:  ui, monitoring

**Context**: 
The definition of disk latency is

>   Disk latency is the time that it takes to complete a single I/O operation on a block device

**Summary**:
  const volumeLatencyWriteQuery = 
`sum(irate(node_disk_write_time_seconds_total[5m]) / irate(node_disk_writes_completed_total[5m])) by (instance, device) * 1000000`;

  const volumeLatencyReadQuery = 
`sum(irate(node_disk_read_time_seconds_total[5m]) / irate(node_disk_reads_completed_total[5m])) by (instance, device) * 1000000`;
  
The latency in the volume table:
const volumeLatencyCurrentQuery = 
`irate(node_disk_io_time_seconds_total[1h]) * 1000000`;

 const volumeUsageQuery =
`kubelet_volume_stats_used_bytes / kubelet_volume_stats_capacity_bytes`;

**Acceptance criteria**: 

For loopdevice in Centos 7, the result of Prometheus is always **0**, so `query node_disk_writes_completed_total` and `node_disk_reads_completed_total` should return 0 which cause latency **NaN**, so no graph displayed.

 With warning:
```
WARN Infinite extent for field "read": [Infinity, -Infinity]
WARN Infinite extent for field "write": [Infinity, -Infinity]
```
**Please note that only the last 6 commits are belonging to this PR**

---

Closes: #2756


ui/src/ducks/app/monitoring.js
```diff
@@ -336,18 +337,18 @@ export function* fetchVolumeStats() {
...
   const volumeThroughputWriteQuery = `sum(irate(node_disk_written_bytes_total[1m])) by (instance, device)`;
   const volumeIOPSReadQuery = `sum(irate(node_disk_reads_completed_total[5m])) by (instance, device)`;
   const volumeIOPSWriteQuery = `sum(irate(node_disk_writes_completed_total[5m])) by (instance, device)`;
 
-  // the query for latency
+  // the query for `latency`
   // Disk latency is the time that it takes to complete a single I/O operation on a block device
-  // rate(node_disk_read_time_seconds_total[5m]) / rate(node_disk_reads_completed_total[5m]
-  const volumeLatencyQuery = `sum(irate(node_disk_io_time_seconds_total[1m])) by (instance, device)`;
+  const volumeLatencyWriteQuery = `sum(irate(node_disk_write_time_seconds_total[5m]) / irate(node_disk_writes_completed_total[5m])) by (instance, device)`;
```
gdemonet: Are you certain we want to use `irate` here? I don't have a clear answer for you, so I suggest we keep it like this for now and open a ticket to check what this query (and others) mean exactly.

FYI, computing rates and means isn't an easy task (at all) and can be super counter-intuitive. Here are some valuable articles to get some insights on the topics:
- On `rate` vs. `irate`: https://utcc.utoronto.ca/~cks/space/blog/sysadmin/PrometheusRateVsIrate
- On the three Pythagorean means: https://towardsdatascience.com/on-average-youre-using-the-wrong-average-geometric-harmonic-means-in-data-analysis-2a703e21ea0

gdemonet (approved):
LGTM!

I'm just surprised by the last commit ["Update the test"](https://github.com/scality/metalk8s/commit/bc19256c45c77960fa3d5be5ac3b9b9381346612), which changes some test data but doesn't explain why it's needed or what it is actually changing :confused:

ui/src/ducks/app/monitoring.js
ChengYanJin: Hello @gdemonet ,
I also have an article to share: [Irate graphs are better graphs](https://www.robustperception.io/irate-graphs-are-better-graphs)

> The rate function takes a time series over a time range, and based on the first and last data points within that range (allowing for counter resets) calculates a per-second rate. 

> By contrast irate is an instant rate. It only looks at the last two points within the range passed to it and calculates a per-second rate.

By using irate we can see much briefer spikes and finer detail. 
So maybe the question is do we want to expose this detail in the graph?

PR #2785, opened 2020-09-09
Bugfix/2784 fix call to device name

### LINKED ISSUE #2784: Update device_name's caller
**Component**:

salt

**What happened**:

Trying to delete a volume in Block mode fails with the following traceback:

```python
  File "/usr/lib/python2.7/site-packages/salt/state .py", line 1981, in call
    **cdata['kwargs'])
  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1977, in wrapper
    return f(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/states/metalk8s_volumes.py", line 136, in removed
    if __salt__['metalk8s_volumes.is_cleaned_up'](name):
  Fi le "/var/cache/salt/minion/extmods/modules/metalk8s_volumes.py", line 145, in is_cleaned_up
    return _get_volume(name).is_cleaned_up
  File "/var/cache/salt/minion/extmods/modules/metalk8s_volumes.py", line 539, in _get_volume
    return RawBlockDeviceBlock(volume)
  File "/var/cache/salt/minion/extmods/mod ules/metalk8s_volumes.py", line 450, in __init__
    self._partition = self._get_partition(device_name(self.path))
  File "/var/cache/salt/minion/extmods/modules/metalk8s_volumes.py", line 516, in _get_partition
    match = re.search(part_re, device_name)
  File "/usr/lib64/python2.7/re.py", line 142, in sear ch
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer
```

This bug was introduced in 72324c2bdb3a38b4ced389855a2135801240fdd3: in order to deal with the perfectible™ error handling of Salt, we changed the return value of `metalk8s_volumes.device_name`.
We took care to update the code of the storage-operator accordingly, but we missed the internal usage of this module (the underlying Python function is also called from inside the module).

We totally missed that bug because:
- We don't have test on the deletion of a `rawBlockDevice` in Block mode (would need to attach disk to the test VM).
- We mock `device_name` in the unit tests, so even though we have 100% coverage it wasn't really meaningful…

**What was expected**:

The volume is deleted, no traceback.

**Steps to reproduce**

Create a `rawBlockDevice` volume in Block mode.
Delete the volume and see it fails…

**Resolution proposal**:

Update the functions that call `device_name` to handle the return value properly.

slaperche-scality:
**Component**:

salt

**Context**: 

Change in the return value of `device_name` broke some function relying on it.

**Summary**:

Add a helper around `device_name` that restore the old behavior.

**Acceptance criteria**: 

Tests are passing.

---

Closes: #2784

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -396,8 +396,9 @@ class SparseLoopDeviceBlock(SparseLoopDevice):
         # Partition 1 should be identified with the volume UUID.
         device = os.path.basename(self.path) + '1'
         try:
-            return device_name(self.persistent_path) == device
-        except: # Expected exception if the symlink doesn't exist.
+            return _device_name(self.persistent_path) == device
+        # Expected exception if the symlink doesn't exist.
+        except CommandExecutionError:
             return False
 
     def prepare(self, force=False):
@@ -447,7 +448,7 @@ class RawBlockDeviceBlock(RawBlockDevice):
         super(RawBlockDeviceBlock, self).__init__(volume)
         # Detect which kind of device we have: a real disk, only a partition or
         # an LVM volume.
-        self._partition = self._get_partition(device_name(self.path))
+        self._partition = self._get_partition(_device_name(self.path))
         if self._get_lvm_path() is not None:
             self._kind = DeviceType.LVM
         elif self._partition is not None:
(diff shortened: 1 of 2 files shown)


gdemonet (approved):
LGTM :ok_hand:

PR #2787, opened 2020-09-11
[UI] Fill the node information for node list table

### LINKED ISSUE #2772: [UI New Node Page] Fill node information in Node table
follow ##2767 Figure out the API needed for Node list
**Component**: ui, node

**Why this is needed**:
I want to see the basic information of each node. 

**What should be done**:
Fill the node table with
- Node Name
- Node Role
- Node  Status
- Node Health
- Node control plane IP
- Node workload plane IP

TBD

**Test plan**:
Unit test for the getNodeList() function

ChengYanJin:
**Component**: ui, nodes

**Context**: Fill the node list table

**Summary**:
- Add the Control Plane IP and Workload Plane IP under the Node Name
- Get the Roles by the labels `node-role.kubernetes.io/<role>`
- Be able to retrieve the alerts from alertmanager.
- But haven't implemented the filter to compute the health
- the color of the status depends on the condition and status
          // green for status.conditions['Ready'] == True and all other conditions are false
          // yellow for status.conditions['Ready'] == True and some other conditions are true
          // red for status.conditions['Ready'] == False
          // grey when there is no status.conditions
- Update the unit test in nodes

**Acceptance criteria**: 
image

---

Closes: #2772

### DIFF
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -103,6 +103,32 @@ const ActionContainer = styled.span`
   display: flex;
 `;
 
+const NodeNameText = styled.div`
+  font-size: ${fontSize.large};
+`;
+
+const IPText = styled.span`
+  font-size: ${fontSize.smaller};
+  padding-right: ${padding.small};
+  color: ${(props) => props.theme.brand.textSecondary};
+`;
+
+// the color of the status depends on the `Status` and `Condition` of the Node
+const StatusText = styled.div`
+  color: ${(props) => {
+    switch (props.textColor) {
+      case 'green':
+        return props.theme.brand.healthy;
+      case 'yellow':
+        return props.theme.brand.warning;
+      case 'red':
+        return props.theme.brand.critical;
+      default:
+        return props.theme.brand.textSecondary;
+    }
+  }};
+`;
+
 function GlobalFilter({
   preGlobalFilteredRows,
   globalFilter,
@@ -247,10 +273,17 @@ function Table({ columns, data, rowClicked, theme }) {
                       ...cell.column.cellStyle,
                     },
                   });
-                  if (
-                    cell.column.Header !== 'Name' &&
-                    cell.value === intl.translate('unknown')
-                  ) {
+                  if (cell.column.Header === 'Name') {
+                    return (
+                      <Cell {...cellProps}>
+                        <NodeNameText>{cell.value.name}</NodeNameText>
+                        <div>
+                          <IPText>CP: {cell.value.control_plane_ip}</IPText>
+                          <IPText>WP: {cell.value.workload_plane_ip}</IPText>
+                        </div>
+                      </Cell>
+                    );
+                  } else if (cell.value === intl.translate('unknown')) {
                     return (
                       <Cell {...cellProps}>
                         <div>{intl.translate('unknown')}</div>
@@ -270,19 +303,34 @@ function Table({ columns, data, rowClicked, theme }) {
 }
 
 const NodeListTable = (props) => {
+  const { nodeListData } = props;
+
   const theme = useSelector((state) => state.config.theme);
 
   const columns = React.useMemo(
     () => [
       {
-        Header: 'Name',
-        accessor: 'name',
-        width: 200,
+        Header: 'Health',
+        accessor: 'health',
+        cellStyle: { textAlign: 'center', width: '50px' },
+        Cell: (cellProps) => {
+          return (
+            <CircleStatus
+              className="fa fa-circle fa-2x"
+              status={cellProps.value}
+            />
+          );
+        },
       },
       {
-        Header: 'Role',
-        accessor: 'role',
-        width: 100,
+        Header: 'Name',
+        accessor: 'name',
+        cellStyle: { width: '180px' },
+      },
+      {
+        Header: 'Roles',
+        accessor: 'roles',
+        cellStyle: { width: '200px' },
       },
       {
         Header: 'Health',
@@ -298,6 +346,39 @@ const NodeListTable = (props) => {
         Header: 'Status',
         accessor: 'status',
         cellStyle: { textAlign: 'center', width: '100px' },
+        Cell: (cellProps) => {
+          const { status, conditions } = cellProps.value;
+          // green for status.conditions['Ready'] == True and all other conditions are false
+          // yellow for status.conditions['Ready'] == True and some other conditions are true
+          // red for status.conditions['Ready'] == False
+          // grey when there is no status.conditions
+          const otherConditions = conditions?.filter(
+            (cond) => cond !== 'ready',
+          );
+          if (status === 'ready' && otherConditions.length === 0) {
+            return (
+              <StatusText textColor="green">
+                {intl.translate('ready')}
+              </StatusText>
+            );
+          } else if (status === 'ready' && otherConditions.length !== 0) {
+            otherConditions.map((cond) => {
+              return <StatusText textColor="yellow">{cond}</StatusText>;
+            });
+          } else if (status !== 'ready') {
+            return (
+              <StatusText textColor="red">
+                {intl.translate('not_ready')}
+              </StatusText>
+            );
+          } else if (!otherConditions) {
+            return (
+              <StatusText textColor="gray">
+                {intl.translate('unknown')}
+              </StatusText>
+            );
+          }
+        },
       },
     ],
     [],
@@ -310,7 +391,7 @@ const NodeListTable = (props) => {
     <NodeListContainer>
       <Table
         columns={columns}
-        data={[]}
+        data={nodeListData}
         rowClicked={onClickRow}
         theme={theme}
         defaultPageSize={10}
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -342,9 +342,8 @@ const NodeListTable = (props) => {
           // yellow for status.conditions['Ready'] == True and some other conditions are true
           // red for status.conditions['Ready'] == False
           // grey when there is no status.conditions
-          const notReadyConditions = conditions?.filter(
-            (cond) => cond !== 'ready' ?? [],
-          );
+          const notReadyConditions =
+            conditions?.filter((cond) => cond !== 'Ready') ?? [];
           if (status === 'ready' && notReadyConditions.length === 0) {
             return (
               <StatusText textColor="green">
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -303,7 +303,7 @@ function Table({ columns, data, rowClicked, theme }) {
 }
 
 const NodeListTable = (props) => {
-  const { nodeListData } = props;
+  const { nodeTableData } = props;
 
   const theme = useSelector((state) => state.config.theme);
 
@@ -380,7 +380,7 @@ const NodeListTable = (props) => {
     <NodeListContainer>
       <Table
         columns={columns}
-        data={nodeListData}
+        data={nodeTableData}
         rowClicked={onClickRow}
         theme={theme}
         defaultPageSize={10}
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -338,20 +338,19 @@ const NodeListTable = (props) => {
         cellStyle: { textAlign: 'center', width: '100px' },
         Cell: (cellProps) => {
           const { status, conditions } = cellProps.value;
+          // the `conditions` include the other conditions that exclude "Ready".
           // green for status.conditions['Ready'] == True and all other conditions are false
           // yellow for status.conditions['Ready'] == True and some other conditions are true
           // red for status.conditions['Ready'] == False
           // grey when there is no status.conditions
-          const notReadyConditions =
-            conditions?.filter((cond) => cond !== 'Ready') ?? [];
-          if (status === 'ready' && notReadyConditions.length === 0) {
+          if (status === 'ready' && conditions.length === 0) {
             return (
               <StatusText textColor="green">
                 {intl.translate('ready')}
               </StatusText>
             );
-          } else if (status === 'ready' && notReadyConditions.length !== 0) {
-            notReadyConditions.map((cond) => {
+          } else if (status === 'ready' && conditions.length !== 0) {
+            conditions.map((cond) => {
               return <StatusText textColor="yellow">{cond}</StatusText>;
             });
           } else if (status !== 'ready') {
@@ -383,7 +382,6 @@ const NodeListTable = (props) => {
         data={nodeTableData}
         rowClicked={onClickRow}
         theme={theme}
-        defaultPageSize={10}
       />
     </NodeListContainer>
   );
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -342,17 +342,17 @@ const NodeListTable = (props) => {
           // yellow for status.conditions['Ready'] == True and some other conditions are true
           // red for status.conditions['Ready'] == False
           // grey when there is no status.conditions
-          const otherConditions = conditions?.filter(
-            (cond) => cond !== 'ready',
+          const notReadyConditions = conditions?.filter(
+            (cond) => cond !== 'ready' ?? [],
           );
-          if (status === 'ready' && otherConditions.length === 0) {
+          if (status === 'ready' && notReadyConditions.length === 0) {
             return (
               <StatusText textColor="green">
                 {intl.translate('ready')}
               </StatusText>
             );
-          } else if (status === 'ready' && otherConditions.length !== 0) {
-            otherConditions.map((cond) => {
+          } else if (status === 'ready' && notReadyConditions.length !== 0) {
+            notReadyConditions.map((cond) => {
               return <StatusText textColor="yellow">{cond}</StatusText>;
             });
           } else if (status !== 'ready') {
@@ -361,7 +361,7 @@ const NodeListTable = (props) => {
                 {intl.translate('not_ready')}
               </StatusText>
             );
-          } else if (!otherConditions) {
+          } else {
             return (
               <StatusText textColor="gray">
                 {intl.translate('unknown')}
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -310,19 +310,6 @@ const NodeListTable = (props) => {
   const columns = React.useMemo(
     () => [
       {
-        Header: 'Health',
-        accessor: 'health',
-        cellStyle: { textAlign: 'center', width: '50px' },
-        Cell: (cellProps) => {
-          return (
-            <CircleStatus
-              className="fa fa-circle fa-2x"
-              status={cellProps.value}
-            />
-          );
-        },
-      },
-      {
         Header: 'Name',
         accessor: 'name',
         cellStyle: { width: '180px' },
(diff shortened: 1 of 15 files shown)


ui/src/components/NodeListTable.js
```diff
@@ -270,34 +303,72 @@ function Table({ columns, data, rowClicked, theme }) {
...
         Header: 'Status',
         accessor: 'status',
         cellStyle: { textAlign: 'center', width: '100px' },
+        Cell: (cellProps) => {
+          const { status, conditions } = cellProps.value;
+          // green for status.conditions['Ready'] == True and all other conditions are false
+          // yellow for status.conditions['Ready'] == True and some other conditions are true
+          // red for status.conditions['Ready'] == False
+          // grey when there is no status.conditions
+          const otherConditions = conditions?.filter(
```
MonPote:
```javascript
const otherConditions = conditions?.filter((cond) => cond !== 'ready') ?? [];
```

ui/src/components/NodeListTable.js
```diff
@@ -270,34 +303,72 @@ function Table({ columns, data, rowClicked, theme }) {
...
+            otherConditions.map((cond) => {
+              return <StatusText textColor="yellow">{cond}</StatusText>;
+            });
+          } else if (status !== 'ready') {
+            return (
+              <StatusText textColor="red">
+                {intl.translate('not_ready')}
+              </StatusText>
+            );
+          } else if (!otherConditions) {
```
MonPote: If you do the preview change then it's should be `} else`

ui/src/services/NodeUtils.js
```diff
@@ -0,0 +1,31 @@
...
+            undefined,
+          workload_plane_ip:
+            (nodesIPsInterfaces[node.name] &&
+              nodesIPsInterfaces[node.name][METALK8S_WORKLOAD_PLANE_IP]) ||
+            undefined,
+        },
+        status: { status: node?.status, conditions: node?.conditions },
+        roles: node?.roles,
+      };
+    });
```
MonPote:
```javascript
    return nodes?.map((node) => {
      return {
        // IPs of Control Plane and Workload Plane are in the same Cell with Name
        name: {
          name: node?.name,
          control_plane_ip:
            (nodesIPsInterfaces[node.name] &&
              nodesIPsInterfaces[node.name][METALK8S_CONTROL_PLANE_IP]) ??
            undefined,
          workload_plane_ip:
            (nodesIPsInterfaces[node.name] &&
              nodesIPsInterfaces[node.name][METALK8S_WORKLOAD_PLANE_IP]) ??
            undefined,
        },
        status: { status: node?.status, conditions: node?.conditions },
        roles: node?.roles,
      };
    }) ?? [];
```

ui/src/ducks/app/nodes.js
```diff
@@ -439,6 +454,21 @@ export function* stopRefreshNodes() {
   );
 }
 
+export function* fetchNodesIPsInterface() {
+  // Action FETCH_NODES_IPS_INTERFACES is faster than SALT_AUTHENTICATION_SUCCESS, hence we get 401 Unauthorized here
+  // Manually authenticate salt API before get the IPs and Interfaces for Nodes
+  yield call(authenticateSaltApi);
```
gdemonet: Not sure that's the right way of waiting for some other action to complete (we always authenticate Salt API on login, so we shouldn't need to re-authenticate). Instead, your goal should be to model the behaviour of this specific action depending on the various cases:

- If we're authenticated to Salt API (`yield select(state => state.login.salt) !== null`), continue
- If we're not, wait for the authentication success/failure action

To manage this "waiting" behaviour, you can use some more advanced effect creators:

```javascript
const { success, failure, timeout } = yield race({
  success: take(SALT_AUTHENTICATION_SUCCESS),
  failure: take(SALT_AUTHENTICATION_FAILED),
  timeout: delay(3000),
});
```

See https://redux-saga.js.org/docs/advanced/RacingEffects.html for more info.

ui/src/ducks/app/nodes.js
```diff
@@ -145,6 +149,7 @@ const defaultState = {
   list: [],
   isRefreshing: false,
   isLoading: false,
+  nodesIPsInterfaces: {},
```
gdemonet: This key lives under `state.app.nodes` already, so maybe we could find a shorter name... Maybe simply `IPs`, or `IPsInfo`?

And one other thing, we don't want to keep unneeded info in the application state, so let's reduce the interface as an IP attribute. In the end, I think this value should look like this (and it could be built in the reducer from the Salt API response):

```javascript
{ IPsInfo: {
  bootstrap: {
    control_plane: [
      { ip: '10.0.1.42', interface: 'eth1'}
    ],
    workload_plane: [
      { ip: '10.100.0.2', interface: 'eth3'},
      { ip: '10.105.1.3', interface: 'eth4'}
    ],
  },
  'master-1': {...},
} }
```

ui/src/services/NodeUtils.js
```diff
@@ -0,0 +1,31 @@
...
+        name: {
+          name: node?.name,
+          control_plane_ip:
+            (nodesIPsInterfaces[node.name] &&
+              nodesIPsInterfaces[node.name][METALK8S_CONTROL_PLANE_IP]) ||
+            undefined,
+          workload_plane_ip:
+            (nodesIPsInterfaces[node.name] &&
+              nodesIPsInterfaces[node.name][METALK8S_WORKLOAD_PLANE_IP]) ||
+            undefined,
```
gdemonet: minor: Since you're checking if there is existing info about some node twice, would be better to store `nodesIPsInterfaces[node.name]` in a variable.

Also, small reminder that the "nullish coallescing operator" also works for bracket notation (cf. [the docs](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining)).

And a small nitpick: let's not use `snake_case` in some places while `camelCase` is the "standard" in this codebase. Consistency is important!

Wrapping it all together:

```javascript
const IPsInfo = nodesIPsInterfaces?.[node.name];
return {
  name: {
    name: node.name,
    controlPlaneIP: IPsInfo?.[METALK8S_CONTROL_PLANE_IP],
    workloadPlaneIP: IPsInfo?.[METALK8S_WORKLOAD_PLANE_IP],
  },
  // ...
}
```

ui/src/services/alertmanager/api.js
```diff
@@ -0,0 +1,11 @@
+import ApiClient from '../ApiClient';
+
+let alertmanagerApiClient = null;
+
+export function initialize(apiUrl) {
+  alertmanagerApiClient = new ApiClient({ apiUrl });
+}
+
+export function getAlerts() {
+  return alertmanagerApiClient.get('/api/v2/alerts');
```
gdemonet, line 10: Oh, small reminder that this `/v2` API is not stable yet. It's relatively OK to use it, because we will know when we update AlertManager, but maybe we should add some "stability" tests to check the API schema we rely on doesn't change without us noticing...

ui/src/ducks/app/nodes.js
```diff
@@ -252,6 +252,17 @@ export function* fetchNodes() {
...
+
+          // Store the name of conditions which the status are True in the array
+          // given the avaiable conditions (Ready, DiskPressure, MemoryPressure, PIDPressure, Network Unavailable, Unschedulable)
+          let conditions = [];
+          const activeCondition = node.status.conditions.find(
+            (cond) => cond.status === 'True',
+          );
+          if (activeCondition && activeCondition.types) {
+            conditions.push(activeCondition.type);
+          }
```
gdemonet: That's only giving one condition, if any, in the `conditions` array... I guess it should be:

```suggestion
          const conditions = node.status.conditions.filter(
            (cond) => cond.status === 'True'
          ).map((cond) => cond.type);
```

Or, using a single `reduce` call (which allows to also check if `type` is a valid attribute):

```javascript
const conditions = node.status.conditions.reduce(
  (acc, cond) => {
    if (cond.status === 'True' && cond?.type)
      acc.push(cond.type);
    return acc;
  },
  []
);
```

ui/src/ducks/app/nodes.js
```diff
@@ -252,6 +252,17 @@ export function* fetchNodes() {
             node.status.conditions.find(
               (conditon) => conditon.type === 'Ready',
             );
+
+          // Store the name of conditions which the status are True in the array
+          // given the avaiable conditions (Ready, DiskPressure, MemoryPressure, PIDPressure, Network Unavailable, Unschedulable)
```
gdemonet: Since we have all the information we need from the `Ready` condition, why do we bother with keeping it? If we look at how it's consumed in the nodes list, we're filtering it anyway (to get the `otherConditions`) :stuck_out_tongue_winking_eye:

ui/src/components/NodeListTable.js
gdemonet: In any case, it makes sense to always show this "unknown" status as a fallback (especially given how the `status` prop is computed).

ui/src/containers/NodePage.js
```diff
@@ -32,7 +41,10 @@ const NodePage = (props) => {
           ]}
         />
       </BreadcrumbContainer>
-      <NodePageContent nodes={nodes}></NodePageContent>
+      <NodePageContent
+        nodes={nodes}
+        nodeListData={nodeListData}
```
gdemonet: This is highly disturbing for a new-comer... We have `nodes`, which is supposedly a list of Node objects, and `nodeListData`, which is what? Some enriched list of Nodes with their associated alerts and IPs? Why not get only one of them?

ui/src/ducks/app/nodes.js
```diff
@@ -271,21 +271,26 @@ export function* fetchNodes() {
...
             status = API_STATUS_UNKNOWN;
           }
+
+          // the Roles of the Node should be the ones that are stored in the labels `node-role.kubernetes.io/<role-name>`
+          let nodeRolesLabels = [];
           const roleTaintMatched = roleTaintMap.find((item) => {
-            const nodeRoles = Object.keys(node.metadata.labels).filter((role) =>
+            nodeRolesLabels = Object.keys(node.metadata.labels).filter((role) =>
               role.includes(ROLE_PREFIX),
             );
```
gdemonet: Instead of filling `nodeRolesLabels` on each pass of `find` (which is highly inefficient), just define it when creating the `nodeRolesLabels` variable (which should be a `const` anyway).

Also, can you make sure we don't forget to remove `roleTaintMatched`, `roleTaintMap` and the corresponding role props (`bootstrap`, `infra`, `control_plane` and `workload_plane`) once this page replaces the old one?

ui/src/ducks/app/nodes.js
```diff
@@ -311,7 +316,7 @@ export function* fetchNodes() {
             workload_plane: roleTaintMatched && roleTaintMatched.workload_plane,
             bootstrap: roleTaintMatched && roleTaintMatched.bootstrap,
             infra: roleTaintMatched && roleTaintMatched.infra,
-            roles: rolesLabel.join(' / '),
+            roles: nodeRoles.join(' / '),
```
gdemonet, line 293: I'd rather make this an array, and let consumers display it as they wish. But since this prop is also consumed by the old page, I guess you can save this as debt.

ui/src/components/NodeListTable.js
```diff
@@ -342,17 +342,17 @@ const NodeListTable = (props) => {
           // yellow for status.conditions['Ready'] == True and some other conditions are true
           // red for status.conditions['Ready'] == False
           // grey when there is no status.conditions
-          const otherConditions = conditions?.filter(
-            (cond) => cond !== 'ready',
+          const notReadyConditions = conditions?.filter(
+            (cond) => cond !== 'ready' ?? [],
           );
```
gdemonet: Not like this, or you'll get an array of arrays :stuck_out_tongue: 

```suggestion
          const notReadyConditions = conditions?.filter(
            (cond) => cond !== 'ready',
          )  ?? [];
```

ui/src/containers/PrivateRoute.js
```diff
@@ -3,16 +3,22 @@ import { useSelector } from 'react-redux';
...
+    // Go to Dex Login Form if not authenticated
     userManager.signinRedirect({
       data: { path: window.location.pathname },
-    }); //Go to Dex Login Form if not authenticated
+    });
     return null;
-  } else {
+  } else if (isSaltAPIAuthenticated) {
+    // To make sure that Salt API is authenticated, before load the components
+    // Otherwise we may get 401 Unauthorized
```
gdemonet: That's a band-aid which is "covering" the issue but not fixing it. Actually, sagas which require Salt API shouldn't be called before we are authenticated. But components could very well be shown before that. And since most of our data doesn't come from Salt anyway, it makes a lot of sense to show it before we are authenticated to SaltAPI.

gdemonet (changes requested):
When reworking sagas and the Redux state, let's try to tidy things up as much as possible, otherwise debt will start piling up and it will soon become a mess for anyone not used to it :)

Otherwise, most of the behaviour change looks good to me :+1:

ui/src/containers/PrivateRoute.js
ChengYanJin: You're absolutely right! There is nothing stopping us load the component. 
We should use `yield race` to wait for Salt Authenticated success action. 
It's great to discover this effect creator. 

Thank you!

ui/src/ducks/app/nodes.js
ChengYanJin: What you mean is pass the `condition` object directly to the `<NodeListTable>` component instead of creating conditions array with ['Ready'...]?

True! We did the filter in NodeListTable anyway. 
But actually there is a lot of information in the condition field, such as lastHeartbeatTime, lastTransitionTime, reason, message... those data is not useful for the <NodeListTable> 

So maybe it's better to create an array with the name of the active conditions and pass it to <NodeListTable>?

What do you think?

gdemonet: Oh no that's not what I'm saying. I'm just suggesting that we could drop the `Ready` condition from this `conditions` array, because we don't look at it (since we have what we need in the `status` prop already) :slightly_smiling_face:

ui/src/containers/NodePage.js
ChengYanJin: Right! It is very confusing. 
So maybe I should rename `nodeListData` to `nodeTableData` which is used by the NodeListTable component. 

The two object `nodeTableData` and `nodes` have some differences and we need both in the RightsSidePanel.

ui/src/ducks/app/nodes.js
ChengYanJin: I will create a ticket, once we are done with the New Node Page, we should change the current temporary URL to the `/volumes` and remove the unused props as you mentioned before `roleTaintMatched` and `roleTaintMatched`...

ui/src/containers/NodePage.js
gdemonet: What difference? Can't we make sure we have all we need in a single array?

ChengYanJin: See for me, the `nodeTableData` should only include the data needed for the table, not more than that.
And the `nodes` is the object return by k8s with more information. 

Here actually we don't need to fetch the `nodes`. Since it's only needed in the health tabs. We should fetch it in the lower child component `<NodeRSP>`

ui/src/services/NodeUtils.js
```diff
@@ -0,0 +1,68 @@
...
+//   workload_plane: { ip: '10.100.0.2', interface: 'eth3'},
+// }
+export const nodesCPWPIPsInterface = (IPsInterfacesObject) => {
+  return {
+    controlPlane: {
+      ip: IPsInterfacesObject[METALK8S_CONTROL_PLANE_IP],
+      interface: Object.keys(IPsInterfacesObject[IP_INTERFACES]).find(
+        (en) =>
+          IPsInterfacesObject[IP_INTERFACES][en][0] ===
+          IPsInterfacesObject[METALK8S_CONTROL_PLANE_IP],
```
gdemonet: This kind of check assumes the IP we're looking for is returned first in `ip_interfaces`, which isn't robust... 

```suggestion
          IPsInterfacesObject[IP_INTERFACES][en].includes(
            IPsInterfacesObject[METALK8S_CONTROL_PLANE_IP]
          ),
```

ui/src/ducks/app/nodes.js
```diff
@@ -439,6 +474,39 @@ export function* stopRefreshNodes() {
...
+export function* fetchNodesIPsInterface() {
+  let result;
+  // Check if Salt API is already authenticated
+  // If not, wait for the CONNECT_SALT_API action.
+  const isSaltAPIAuthenticated = yield select(isSaltAPIAuthenticatedSelector);
+  if (isSaltAPIAuthenticated) {
+    result = yield call(ApiSalt.getNodesIPsInterfaces);
+  } else {
+    // eslint-disable-next-line no-unused-vars
+    const { success, failure, timeout } = yield race({
```
gdemonet, line 455: It's sad that we are missing proper error-handling for this :confused: I'll open a debt ticket to track this.

ui/src/ducks/app/nodes.js
```diff
@@ -246,21 +277,25 @@ export function* fetchNodes() {
             status = API_STATUS_UNKNOWN;
           }
 
+          // the Roles of the Node should be the ones that are stored in the labels `node-role.kubernetes.io/<role-name>`
+          let nodeRolesLabels = [];
           const roleTaintMatched = roleTaintMap.find((item) => {
-            const nodeRoles = Object.keys(node.metadata.labels).filter((role) =>
+            nodeRolesLabels = Object.keys(node.metadata.labels).filter((role) =>
               role.includes(ROLE_PREFIX),
             );
```
gdemonet: Quoting my previous comment:

> Instead of filling `nodeRolesLabels` on each pass of `find` (which is highly inefficient), just define it when creating the `nodeRolesLabels` variable (which should be a `const` anyway).

That would look like:

```javascript
// the Roles of the Node should be the ones that are stored in the labels `node-role.kubernetes.io/<role-name>`
const nodeRolesLabels = Object.keys(node.metadata.labels).filter(
  (label) => label.startsWith(ROLE_PREFIX),
);
const roleTaintMatched = roleTaintMap.find((item) => {
  return (
    // what you already have
  );
};
```

Note that I used `startsWith` instead of `includes`, which is more accurate (`includes` could match a `custom.node-role.kubernetes.io/toto` label, which we don't want)

gdemonet (changes requested):
Last few changes and we're good to go!

ui/src/ducks/app/nodes.js
ChengYanJin: Remove the `roleTaintMatched` and control_plane, workload_plane, bootstrap and infra from the returned Node object.
Since we can get the Roles from the Labels, I don't think we need those props anymore.

ui/src/ducks/app/nodes.test.js
```diff
@@ -267,10 +263,6 @@ describe('`fetchNodes` saga', () => {
 
     const expectedNode = {
       ...defaultNodeForState,
-      workload_plane: expected.wp,
-      control_plane: expected.cp,
-      bootstrap: expected.bootstrap,
-      infra: expected.infra,
```
gdemonet, line 272: Since they're not used, you can remove them from the `expected` object

gdemonet (approved):
Perfect! Thanks for taking care of the debt :smiley:

ChengYanJin:
> Perfect! Thanks for taking care of the debt 😃

I will. I think when tackle this debt ticket may deserve a workshop/discussion.

gdemonet:
> > Perfect! Thanks for taking care of the debt smiley
> 
> I will. I think when tackle this debt ticket may deserve a workshop/discussion.

Oh no I meant in this PR, no rush for the other one :wink:

PR #2799, opened 2020-09-16
docs: add loki sizing hints in post-install

### LINKED ISSUE #2687: Define the sizing rule for Loki volume
**Component**: loki, documentation

**Why this is needed**:
To provide a simple formula to help calculate the ideal size needed for the Loki volume.

**What should be done**:
Define a formula to calculate needed size for the Loki volume, based on retention, ingestion rate and number of streams.
Test plan for this could be:
- ignore the line-size, just measure its average on a real platform and assume all applications will lead to the same average size
- ignore the retention, just run the benchmarks with a 1 day retention and assume linear scaling (we have a single Loki instance, so that should be roughly true)
- define a synthetic benchmark, to get data points for multiple (ideally random) combinations of #(streams) x rate(ingestion)
- from the gathered data, try to build a linear regression model (with OLS)

**Implementation proposal** (strongly recommended): n/a

**Test plan**: n/a

alexandre-allard:
**Component**: docs

**Context**: 

**Summary**:
Define a simple Loki volume sizing rule (worst case).
The real sizing formula should be defined in each application documentation.

**Acceptance criteria**: 

---

Closes: #2687

### DIFF
--- a/docs/installation/post-install.rst
+++ b/docs/installation/post-install.rst
@@ -18,8 +18,8 @@ services <node-role-infra>`.
 
 To create the required *Volume* objects, write a YAML file with the following
 contents, replacing ``<node_name>`` with the name of the :term:`Node` on which
-to run Prometheus and AlertManager, and ``<device_path[2]>`` with the ``/dev``
-path for the partitions to use:
+to run Prometheus, AlertManager and Loki, and ``<device_path[3]>`` with the
+``/dev`` path for the partitions to use:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/installation/post-install.rst
```diff
@@ -84,6 +84,37 @@ For more details on the available options for storage management, see
...
+Loki volume sizing
+""""""""""""""""""
+
+Since the storage needs for logs greatly depends on the workload and the
+type of application that run on top of the MetalK8s cluster, you need to
+refer to the documentation provided by your applications to define the ideal
+size for the volume.
+
+We still provide some hints for the worst case, which is very unlikely.
+If your logs are gibberish, which makes them almost incompressible,
```
slaperche-scality: `a random stream of bytes` may be better than `gibberish`

NicolasT:
```suggestion
If the entropy of log messages is high, which makes them almost incompressible,
```

;)

alexandre-allard: You're right, fixed.

alexandre-allard: @NicolasT Oh yes, even better!

docs/installation/post-install.rst
```diff
@@ -84,6 +84,37 @@ For more details on the available options for storage management, see
...
+Regarding the MetalK8s cluster itself (internal services and system logs),
+**1Gb** per week of retention should be sufficient in most cases.
+
+.. warning::
+
+   When you calculate the storage needs, you must always add an extra week to
+   your actual retention, because of the current week of logs.
+
+   It's also recommended to add a security margin of +50% volume space, in case
+   of log burst.
```
gdemonet: Maybe we should also warn that there is no volume-based rotation of logs for now, so keeping a safety margin is really paramount?

alexandre-allard: Yes, right, better be safe than sorry.

docs/installation/post-install.rst
```diff
@@ -84,6 +84,40 @@ For more details on the available options for storage management, see
...
+
+Regarding the MetalK8s cluster itself (internal services and system logs),
+**1Gb** per week of retention should be sufficient in most cases.
+
+.. warning::
+
+   When you calculate the storage needs, you must always add an extra week to
+   your actual retention, because of the current week of logs.
+
+   Since there is no size-based purge mechanism, it's also recommended to add
```
gdemonet: nit: Do not use contractions in docs ;)

```suggestion
   Since there is no size-based purge mechanism, it is also recommended to add
```

gdemonet (dismissed):

docs/installation/post-install.rst
alexandre-allard: hmm, good to know, I was not aware of this.

alexandre-allard: done.

gdemonet (approved):

PR #2804, opened 2020-09-18
Expose plain Dex configuration in CSC ConfigMap

### LINKED ISSUE #2781: Expose plain Dex configuration in CSC ConfigMap
**Component**: salt, dex, service-config

**Why this is needed**:

We use Dex as an OIDC provider for K8s API (and some UIs). As such, it is likely that our users will want to configure for adding users, clients, connectors, of all sorts. Today, only a subset of Dex configuration is exposed through our persistent configuration mechanism called CSC (Cluster and Services Configuration).
The more features we want to add, the more complex our format and associated logic becomes, while still being cumbersome if some specifics are not supported.

**What should be done**:

To simplify our code and still have full support of Dex configuration options, we should expose plain Dex configuration in the CSC ConfigMap for Dex (`metalk8s-auth/metalk8s-dex-config`).
Simplifying user experience when interacting with this configuration file will be added later, through client-based means (e.g. in a UI, or through a `kubectl` plugin).

(to be detailed / validated)

- define new apiVersion to handle flat format for Dex configuration
- quietly convert old format to new in pre-upgrade (for Dex specifically, cf. nginx-ingress-controller issues we had), and reciprocally new to old in pre-downgrade
- document new format and adapt existing procedures for adding users
- add new procedures for adding clients and connectors (mostly referring to Dex docs)

**Test plan**:

(to do)

gdemonet:
**Component**: salt, dex, service-config

**Context**: See #2781 

**Summary**:

- change CSC format, described by new apiVersion `addons.metalk8s.scality.com/v1alpha2`
- adjust how "defaults" are loaded (now rely on `slsutil.renderer` to allow Jinja templating, choice of rendering engine being left to the source file) so we can store and update default clients and issuer URL
- adjust Grafana role mapping: all Dex static users are Admin

_Left to do_:
- [x] Document new format
- [x] Adjust existing procedures accordingly
- [x] Add procedure to configure an OIDC connector (useful for Keycloak) - kind of done, simply referred to Dex docs
- [x] Remove pre-downgrade logic, because 2.6 is not backward compatible

**Acceptance criteria**: 

- Upgrade still works, and Dex behaviour is unchanged, even with non-default configuration in Dex CSC

    - tested with a custom password and a GitHub connector, from 2.5.1 to 2.6.0

- Grafana role mapping still works by default, and additional Dex users are also Admin
- Tests pass (of course)
- Documented procedures for configuring Dex still work

---

Closes: #2781


salt/metalk8s/addons/logging/loki/deployed/chart.sls
```diff
@@ -184,7 +184,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: f6917e98336282f93b8cf80b99ca1e78f7adfd297b98c367c1cc420a1cb8d1ac
+        checksum/config: f9c7883a31a1ef8ef3264cfb5160f3cc015a91cf46c6687d7f2777a62af6936b
```
TeddyAndrieux: Not related to this PR but since we have CSC for loki configuration this one should be computed by salt.
And it's the same for alertmanager, prometheus and grafana I think (+ maybe others)

gdemonet: Good point, I'll add it in this PR if that's OK for you

gdemonet: So, Prometheus doesn't have a "config file" per-se, it's entirely managed through CRs by prometheus-operator, so I don't have something to compute a checksum from.
Did it for Grafana and Alertmanager though (even if the latter is also managed by prometheus-operator), and Loki of course.

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -306,27 +316,62 @@ For more details on Alert Rules, see the official
...
-      super admin account enabled at all times.
+   To continue using MetalK8s OIDC (especially for MetalK8s UI and Grafana)
+   in case of the loss of external identity providers, it is advised to
+   keep the static user store enabled.
 
-To add a new static user, perform the following operations:
+To disable (resp. enable) it, perform the following steps:
 
-   .. _Generate-password-hash:
+#. Set the ``enablePasswordDB`` configuration flag to ``true`` (resp.
```
alexandre-allard:
```suggestion
#. Set the ``enablePasswordDB`` configuration flag to ``false`` (resp.
```

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -335,180 +380,169 @@ To add a new static user, perform the following operations:
...
-
-   - Apply the ClusterRoleBinding configurations using:
+#. Find the entry for the selected user in the ``staticPasswords`` list, and
+   update its ``hash``:
 
    .. code-block:: shell
 
       root@bootstrap $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
-                         apply -f ClusterRoleBinding.yaml
+                         edit configmaps metalk8s-dex-config -n metalk8s-auth
```
alexandre-allard: To be consistent with the change you made above:
```suggestion
                         edit configmap metalk8s-dex-config -n metalk8s-auth
```

docs/operation/cluster_and_service_configuration.rst
gdemonet: :facepalm:

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -335,180 +380,169 @@ To add a new static user, perform the following operations:
...
-      entry you have specified.
+To define (or override) any configuration option, follow these steps:
+
+#. Add (or change) the corresponding field under the ``spec.config`` key of
+   the *metalk8s-auth/metalk8s-dex-config* ConfigMap:
 
    .. code-block:: shell
 
       root@bootstrap $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
                          edit configmaps metalk8s-dex-config -n metalk8s-auth
```
alexandre-allard: ditto

alexandre-allard:
Not tested, only found details in the documentation part, otherwise LGTM.

docs/operation/cluster_and_service_configuration.rst
gdemonet: Will do

docs/operation/cluster_and_service_configuration.rst
gdemonet: Done.

docs/operation/cluster_and_service_configuration.rst
gdemonet: Done.

docs/operation/cluster_and_service_configuration.rst
gdemonet: Done.

alexandre-allard (approved):

PR #2800, opened 2020-09-16
Deploy Calico CNI plugin for all node roles

alexandre-allard:
**Component**: salt

**Context**:
Calico is now needed on all nodes to be able to run fluent-bit.

**Summary**:
Add Calico CNI plugin on master & etcd roles

**Acceptance criteria**:

Tested, looks great, fluent-bit is now running on etcd/master nodes.

---

Closes: #2380

PR #2805, opened 2020-09-18
salt: remove loopback CNI plugin

alexandre-allard:
**Component**: salt

**Context**:
Since #2800, this SLS is no longer used anywhere except in node role, so let put everything that's inside it into the node role.

**Summary**:

**Acceptance criteria**:


salt/metalk8s/roles/node/init.sls
```diff
@@ -1,3 +1,7 @@
 include:
-  - metalk8s.roles.internal.node-without-calico
+  - metalk8s.node.grains
+  - metalk8s.kubernetes.cni.loopback
   - metalk8s.kubernetes.cni.calico
+  - metalk8s.kubernetes.kubelet
+  - metalk8s.kubernetes.apiserver-proxy
+  - metalk8s.internal.preflight
```
gdemonet: Why the re-ordering, compared to what was done before?

alexandre-allard: No real good reason, just to have both cni stuff grouped, since the ordering here does not seem to be important.

gdemonet: Mmh, OK. And do we still need to include `cni.loopback` before `cni.calico`?

alexandre-allard-scality: Do you mean, do we still need to include it or do we need to include it "before" ?

alexandre-allard: I've completely removed the loopback plugin.

salt/metalk8s/orchestrate/upgrade/post.sls
```diff
@@ -2,3 +2,4 @@
 
 include:
   - metalk8s.addons.prometheus-operator.post-upgrade
+  - metalk8s.kubernetes.cni.loopback.absent
```
TeddyAndrieux: No you cannot include it as is here, this state is run on the salt-master so if you do this it will just remove the cni in the salt master container (so likely do not remove anything)
So to me you need to create another orchestrate that run this on every minions
OR
(maybe better) Since we do not need this one once calico is configured as CNI, just include this absent in the calico state (with a comment to say we can remove it in 2.7)

TeddyAndrieux: Or maybe we want to remove all others CNI plugins when we deploy calico :thinking: I don't know if it can cause trouble to have different CNI plugins
@NicolasT @gdemonet Any idea ?
(Anyway it's not the goal of this PR)

NicolasT: Shouldn't cause 'trouble' since they're picked based on filename, hence the `10-` and `30-` (or something) prefix we use.

alexandre-allard: Yeah, I saw during my test that the file was not properly removed, then I forgot to change it.
It's now done.

TeddyAndrieux (approved):
LGTM

PR #2806, opened 2020-09-18
[UI] Add metrics tab in Node Page

ChengYanJin:
**Component**: UI

**Context**: 
Add metrics tab in the New Node Page, the default timespan is **Last 24 hours**

**Summary**:
Note that: 
- Miss the selection of timespan, by default is **Last 24 hours**
- Miss the auto-refresh for the graphs

(Will address them it in PR #2809 ).

**Update**:
- Make the Graph Container scrollable in the Metrics tab
- Set the width and height of the graph according to the `innerHeight` and `innerWidth` of the screen in order to leverage the space we have to display the charts

**Acceptance criteria**: 
image
Update with the line color in graphs with new design:
image


ui/src/ducks/app/monitoring.js
```diff
@@ -177,6 +208,8 @@ export const isCurrentVolumeStatsRefresh = (state) =>
   state.app.monitoring.volumeCurrentStats.isRefreshing;
 export const metricsTimeSpan = (state) =>
   state.app.monitoring.volumeStats.metricsTimeSpan;
+export const nodeMetricsTimeSpan = (state) =>
```
gdemonet: Since we're adding a `nodeMetricsTimeSpan`, would make sense to rename the one above `volumeMetricsTimeSpan`

ui/src/ducks/app/monitoring.js
```diff
@@ -521,6 +554,174 @@ export function* stopRefreshCurrentStats() {
...
+    sampleDuration = SAMPLE_DURATION_LAST_ONE_HOUR;
+    sampleFrequency = SAMPLE_FREQUENCY_LAST_ONE_HOUR;
+  }
+  const currentTime = new Date();
+  const currentTimeISO = currentTime.toISOString(); // To query Prometheus the date should follow `RFC3339` format
+  const startingTimestamp =
+    Math.round(currentTime.getTime() / 1000) - sampleDuration;
+  const startingTimeISO = new Date(startingTimestamp * 1000).toISOString();
+
+  const cpuUsageQuery = `(((count(count(node_cpu_seconds_total{instance=~"${instanceIP}:${PORT_NUMBER_PROMETHEUS}"}) by (cpu))) - avg(sum by (mode)(irate(node_cpu_seconds_total{mode='idle',instance=~"${instanceIP}:${PORT_NUMBER_PROMETHEUS}"}[5m])))) * 100) / count(count(node_cpu_seconds_total{instance=~"${instanceIP}:${PORT_NUMBER_PROMETHEUS}"}) by (cpu))`;
```
gdemonet: I was surprised that your queries could work with the Prometheus port, because we get these metrics from `node-exporter`... But indeed, this `PORT_NUMBER_PROMETHEUS` constant is just not using the right name, because `9100` **is** `node-exporter`'s port (Prom is listening on 9090).

ui/src/ducks/app/monitoring.js
```diff
@@ -521,6 +554,174 @@ export function* stopRefreshCurrentStats() {
...
+  const metrics = {
+    cpuUsage: cpuUsage,
+    systemLoad: systemLoad,
+    memory: memory,
+    iopsRead: iopsRead,
+    iopsWrite: iopsWrite,
+    controlPlaneNetworkBandwidthIn: controlPlaneNetworkBandwidthIn,
+    controlPlaneNetworkBandwidthOut: controlPlaneNetworkBandwidthOut,
+    workloadPlaneNetworkBandwidthIn: workloadPlaneNetworkBandwidthIn,
+    workloadPlaneNetworkBandwidthOut: workloadPlaneNetworkBandwidthOut,
```
gdemonet: Did you know the keys are not necessary in ES6 notation when the value is a variable of the same name? It's called ["shorthand property names"](https://alligator.io/js/object-property-shorthand-es6/#:~:text=New%20in%20JavaScript%20with%20ES6,simply%20pass%20the%20key%20name.)

ui/src/components/NodePageMetricsTab.js
```diff
@@ -1,8 +1,436 @@
...
+  SAMPLE_DURATION_LAST_TWENTY_FOUR_HOURS,
+  SAMPLE_DURATION_LAST_ONE_HOUR,
+  SAMPLE_FREQUENCY_LAST_SEVEN_DAYS,
+  SAMPLE_FREQUENCY_LAST_TWENTY_FOUR_HOURS,
+  SAMPLE_FREQUENCY_LAST_ONE_HOUR,
+} from '../constants';
+
+const GraphsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
```
gdemonet, line 30: Just a suggestion: wouldn't it be easier to use CSS Grid instead of a combination of Flexbox containers? Allows for a simpler gap size management, and uniform sizing (looks better when some of the charts are empty, for instance).

ui/src/components/NodePageMetricsTab.js
```diff
@@ -1,8 +1,436 @@
...
+    type: 'temporal',
+    axis: {
+      // Refer to all the available time format: https://github.com/d3/d3-time-format#locale_format
+      // format: '%m/%d %H:%M',
+      format: '%H:%M', // when the timespan is `Last 24 hours`
+      // Boolean value that determines whether the axis should include ticks.
+      ticks: true,
+      tickCount: 4,
+      labelAngle: -50,
+      labelColor: '#a8b5c1',
```
gdemonet: Shouldn't we have all colors in the theme? (same for chart lines)

ui/src/components/NodePageMetricsTab.js
```diff
@@ -87,98 +85,98 @@ const NodePageMetricsTab = (props) => {
...
-  const nodeControlPlaneNetworkBandwidthInData = controlPlaneNetworkBandwidthInOperated?.map(
-    (slot) => {
+  const cpuUsageData = nodeStatsOperated
+    ?.find((obj) => obj.metrics === 'cpuUsage')
+    ?.data?.map((slot) => {
+      return {
+        date: fromUnixTimestampToDate(slot[0]),
+        y: slot[1],
+      };
+    });
```
gdemonet: Weren't trying to reduce duplication? :stuck_out_tongue_winking_eye:

Instead of doing this mapping after building `nodeStatsOperated`, you could do it during construction:

```javascript
  const nodeStatsOperated = Object.keys(nodeStats).reduce((acc, metricName) => {
    const data = operateMetricRawData(nodeStats[metricName][0]?.values);
    acc[metricName] = data.map((slot) => ({
      date: fromUnixTimestampToDate(slot[0]),
      y: slot[1],
    }));
    return acc;
  }, {});
```

And then consume it like:

```javascript
<LineChart
  id={'node_cpu_usage_id'}
  data={nodeStatsOperated['cpuUsage']}
  {...rest}
/>
```

ui/src/components/NodePageMetricsTab.js
```diff
@@ -406,7 +404,7 @@ const NodePageMetricsTab = (props) => {
                 lineConfig={lineConfig}
               />
             ) : (
-              <NoDataGraphText>Loading…</NoDataGraphText>
+              <LoaderContainer size="small"></LoaderContainer>
```
gdemonet, line 329: Using a loader instead of some text isn't related to the "code deduplication" commit, would be better to create another one.

gdemonet (changes requested):
Some code review comments :wink: 

Also, I played with the page, and have a few comments:

- If my screen is too short in height, I can't see the bottom graphs **and I can't scroll**
- If my screen is too short in width, the graphs are going over the table (I'd rather have scrollbar and push them to the right)
- Whichever the size of my display or the level of zoom, I have (ugly, but that's Chrome's fault) scrollbars everywhere - this problem is true on all pages actually, am I the only one @scality/metalk8s?:

[](https://freeimage.host/i/2Agodv)

ui/src/components/NodePageMetricsTab.js
ChengYanJin: Hello @gdemonet 
Was trying to reduce the duplication here too. But we have different return keys when it comes to the graphs with multiple line (in/out, read/write)
Please see:
```
  const nodeWorkloadPlaneNetworkBandwidthOutData = nodeStatsOperated
    ?.find((obj) => obj.metrics === 'workloadPlaneNetworkBandwidthOut')
    ?.data?.map((slot) => {
      return {
        date: fromUnixTimestampToDate(slot[0]),
        out: slot[1],
        type: 'out',
      };
    });
```

gdemonet: Oh right, missed that one. There still are some ways to reduce this:

- each `.find((obj) => obj.metrics === someKey)` could be simplified if `nodeStatsOperated` was a map instead of an array:

```javascript
const nodeWorkloadPlaneNetworkBandwidthOutData = nodeStatsOperated
  ?.['workloadPlaneNetworkBandwidthOut']?.map((slot) => { ... })
```

- if you want to push it further, adding a `type` key can be done easily by inferring it from the `metricName`:

```javascript
// First variant, assumes a given format for `metricName`
const nodeStatsOperated = Object.keys(nodeStats).reduce((acc, metricName) => {
  const data = operateMetricRawData(nodeStats[metricName][0]?.values);

  let extra = {};
  if (metricName.endsWith('Out')) extra.type = 'out';
  else if (metricName.endsWith('In')) extra.type = 'in';
  else if (metricName.endsWith('Read')) extra.type = 'read';
  else if (metricName.endsWith('Write')) extra.type = 'write';

  acc[metricName] = data.map((slot) => ({
    date: fromUnixTimestampToDate(slot[0]),
    y: slot[1],
    ...extra
  }));
  return acc;
}, {});

// Second variant, stores the mapping between types and metric names
const typedMetrics = {
  iopsRead: 'read',
  iopsWrite: 'write',
  controlPlaneNetworkBandwidthIn: 'in',
  controlPlaneNetworkBandwidthOut: 'out',
  workloadPlaneNetworkBandwidthIn: 'in',
  workloadPlaneNetworkBandwidthOut: 'out',
};

const nodeStatsOperated = Object.keys(nodeStats).reduce((acc, metricName) => {
  const data = operateMetricRawData(nodeStats[metricName][0]?.values);

  let extra = {};
  const metricType = typedMetrics[metricName];
  if (metricType !== undefined) extra.type = metricType;

  acc[metricName] = data.map((slot) => ({
    date: fromUnixTimestampToDate(slot[0]),
    y: slot[1],
    ...extra
  }));
  return acc;
}, {});
```

ui/src/components/NodePageMetricsTab.js
```diff
@@ -87,98 +85,98 @@ const NodePageMetricsTab = (props) => {
...
+    ?.find((obj) => obj.metrics === 'memory')
+    ?.data?.map((slot) => {
+      return {
+        date: fromUnixTimestampToDate(slot[0]),
+        y: slot[1],
+      };
+    });
+
+  const nodeIOPSWriteData = nodeStatsOperated
+    ?.find((obj) => obj.metrics === 'iopsRead')
```
gdemonet: Looks like you switched them up, no?

```suggestion
    ?.find((obj) => obj.metrics === 'iopsWrite')
```

ui/src/components/NodePageMetricsTab.js
```diff
@@ -87,98 +85,98 @@ const NodePageMetricsTab = (props) => {
...
+    ?.data?.map((slot) => {
+      return {
+        date: fromUnixTimestampToDate(slot[0]),
+        write: slot[1],
+        type: 'write',
+      };
+    });
+
+  const nodeIOPSReadData = nodeStatsOperated
+    ?.find((obj) => obj.metrics === 'iopsWrite')
```
gdemonet:
```suggestion
    ?.find((obj) => obj.metrics === 'iopsRead')
```

ui/src/components/NodePageMetricsTab.js
ChengYanJin: Indeed, the color of the label should come from the `defaultTheme`. 
However, the color of the chart lines are exceptional color and we don't use them anywhere else yet. For me, it makes sense to hardcode the color today.

ui/src/components/NodePageMetricsTab.js
ChengYanJin: Actually I didn't see the big difference when come to use CSS Grid. For me, those two can both solve our issue. Maybe just we are more familiar with Flexbox and use it everywhere in our code.

When some of the charts are missing, we'll have an empty chart today. But we should display an error message which is missing today.

gdemonet: Won't fight on this topic, just suggesting you try Grid a bit more for building "grid layouts" (which is the case here). There's a reason they built it ;)

ui/src/components/NodePageMetricsTab.js
gdemonet: I'm not sure chart lines are "exceptional" in our UIs, since most of our resources will have some form of monitoring metrics. We can hardcode them today, sure, but it'd be a good idea to open a ticket in core-ui and think about it more.

ChengYanJin: Sure! I will open a ticket in core-ui. 
And actually, the colors we use in the graphs is still an open topic we haven't finalized yet. We still need to work with PO & UX.

ui/src/components/LinechartSpec.js
```diff
@@ -0,0 +1,40 @@
+export const yAxisUsauge = [
```
gdemonet: typo:

```suggestion
export const yAxisUsage = [
```

ui/src/components/NodePageMetricsTab.js
```diff
@@ -1,8 +1,325 @@
...
+import styled from 'styled-components';
+import {
+  fontSize,
+  padding,
+  fontWeight,
+} from '@scality/core-ui/dist/style/theme';
+import { LineChart, Loader } from '@scality/core-ui';
 import { TabContainer } from './CommonLayoutStyle';
+import {
+  yAxisUsauge,
```
gdemonet:
```suggestion
  yAxisUsage,
```

ui/src/components/VolumeMetricGraphCard.js
```diff
@@ -16,6 +16,7 @@ import {
   addMissingDataPoint,
   fromUnixTimestampToDate,
 } from '../services/utils';
+import { yAxisUsauge, yAxisWriteRead } from './LinechartSpec';
```
gdemonet:
```suggestion
import { yAxisUsage, yAxisWriteRead } from './LinechartSpec';
```

gdemonet (dismissed):
Just a small typo, should be good otherwise :)

gdemonet (approved):
:shipit:

Issue #2035, opened 2019-11-12
UI: improve "refresh" sagas


carlito-scality:
**Component**: UI

**Why this is needed**:
- In the UI, to refresh resources in a page we call "refresh" sagas that look like:
```
export function* refreshAlerts() {
  yield put(updateAlertsAction({ isRefreshing: true }));
  const resultAlerts = yield call(fetchAlerts);
  if (!resultAlerts.error) {
    yield delay(REFRESH_TIMEOUT);
    const isRefreshing = yield select(
      state => state.app.monitoring.alert.isRefreshing,
    );
    if (isRefreshing) {
      yield call(refreshAlerts);
    }
  }
}
```
and to stop the refresh, we call "stopRefresh" sagas that look like:
```
 yield put(updateAlertsAction({ isRefreshing: false }));
```
This "refresh" management is split in 2 different sagas, so it is hard to understand. Moreover, the "refresh" saga has to wait `REFRESH_TIMEOUT` seconds (since `delay` is a blocked call) to know if it has to relaunch the refresh based on a redux state `isRefreshing`. 

Suggested improvements (cf @gdemonet ):

```
export function* refreshAlerts() {
  while (true) {
    yield take(REFRESH_ALERTS);
    while (true) {
      yield fork(fetchAlerts);
      const {interrupt, requeue} = yield race({
        interrupt: take(STOP_REFRESH_ALERTS),
        requeue: delay(DELAY_TIMEOUT),
      }); 
      if (interrupt) { break; }
    }
  }
}
```
- if there is an error during `fetchAlerts`, we can just `yield put({type: STOP_REFRESH_ALERTS})`, to stop the refresh in the `fetchAlerts` saga.
- It would be even better to cancel "fetchAlerts" when "interrupt" occurs.
- Consequently, `state.app.monitoring.alert.isRefreshing` is only used to display refresh icon.

**What should be done**:

**Implementation proposal** (strongly recommended):

**Test plan**:

gdemonet:
Is this still relevant with React Query @JBWatenbergScality?

JBWatenbergScality:
Not really we should better have a debt ticket about switching the current implementation to react-query 🤔

PR #2809, opened 2020-09-22
[UI] Add timespan selection and auto refresh for metrics tab in node page

### LINKED ISSUE #2776: [UI New Node Page] Add metrics graph in Node Page
follows: #2765 

**Component**: ui, nodes

**What should be done**:
- [x] Add the graphs with the dafault timespan of 'Last 24hours'
- [x] the choices of timespan are last 24hours/ last 7 days/ last 1 hour ( by default is last 24 hours)
- [x] Update the URL with the query of current timespan
- [x] the graphs can be auto-refreshed

image
Add one more query param in the URL for the three timespan choices.

Last 7days: &from=now-7d
Last 24hours: &from=now-24h
Last 1 hour: &from=now-1h

**Test plan**:
TBD

ChengYanJin:
**Component**: ui, nodes

**Context**:  Add timespan selection and auto-refresh for metrics tab in the Node page

**Summary**:
- Add timespan selection in `metrics` tab, by default is Last 24 hours
- Add selected timespan in URL query
- Add auto-fresh of graphs every 1 min
- Keep the selected timespan when we switch back to metrics tab
- When we select a node from the list, we don't add the default the timespan(from=now-24h) in the URL. 
- The style of dropdown component is not the same as design. (Need to work on it in Core-ui)
- Launch a long-time running saga at the root of saga to watch the Refresh_NODESTATS action.

**Acceptance criteria**: 
image

image

---

Closes: #2776

### DIFF
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -13,6 +13,13 @@ import { useQuery } from '../services/utils';
 import { fontSize, padding } from '@scality/core-ui/dist/style/theme';
 import CircleStatus from './CircleStatus';
 import { Button } from '@scality/core-ui';
+import {
+  LAST_SEVEN_DAYS,
+  LAST_ONE_HOUR,
+  QUERY_LAST_SEVEN_DAYS,
+  QUERY_LAST_TWENTY_FOUR_HOURS,
+  QUERY_LAST_ONE_HOUR,
+} from '../constants';
 import { intl } from '../translations/IntlGlobalProvider';
 
 const NodeListContainer = styled.div`
@@ -312,8 +319,23 @@ const NodeListTable = (props) => {
   const history = useHistory();
   const location = useLocation();
   const { nodeTableData, selectedNodeName } = props;
-
   const theme = useSelector((state) => state.config.theme);
+
+  let queryTimeSpan;
+  const metricsTimeSpan = useSelector(
+    (state) => state.app.monitoring.nodeStats.metricsTimeSpan,
+  );
+
+  switch (metricsTimeSpan) {
+    case LAST_SEVEN_DAYS:
+      queryTimeSpan = QUERY_LAST_SEVEN_DAYS;
+      break;
+    case LAST_ONE_HOUR:
+      queryTimeSpan = QUERY_LAST_ONE_HOUR;
+      break;
+    default:
+      queryTimeSpan = QUERY_LAST_TWENTY_FOUR_HOURS;
+  }
 
   const columns = React.useMemo(
     () => [
@@ -390,8 +412,12 @@ const NodeListTable = (props) => {
       location.pathname.endsWith('volumes') ||
       location.pathname.endsWith('pods');
 
-    if (isTabSelected) {
-      // When switch between the nodes, keep the same tab selected
+    if (isTabSelected && location.pathname.endsWith('metrics')) {
+      // Keep the same timespan selected for metrics tab, when switch between the nodes
+      const currentTab = location?.pathname?.split('/')?.pop();
+      history.push(`/newNodes/${nodeName}/${currentTab}?from=${queryTimeSpan}`);
+    } else if (isTabSelected) {
+      // Keep the same tab selected when switch between the nodes
       const currentTab = location?.pathname?.split('/')?.pop();
       history.push(`/newNodes/${nodeName}/${currentTab}`);
     } else {
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -313,6 +313,7 @@ const NodeListTable = (props) => {
   const location = useLocation();
   const { nodeTableData, selectedNodeName } = props;
   const theme = useSelector((state) => state.config.theme);
+  const query = useQuery();
 
   const columns = React.useMemo(
     () => [
@@ -391,13 +392,18 @@ const NodeListTable = (props) => {
 
     if (isTabSelected) {
       const newPath = location.pathname.replace(
-        // eslint-disable-next-line no-useless-escape
-        /\/newNodes\/[^\/]*\//,
+        /\/newNodes\/[^/]*\//,
         `/newNodes/${nodeName}/`,
       );
-      history.push(newPath);
+      history.push({
+        pathname: newPath,
+        search: query.toString(),
+      });
     } else {
-      history.push({ pathname: `/newNodes/${nodeName}/health` });
+      history.push({
+        pathname: `/newNodes/${nodeName}/health`,
+        search: query.toString(),
+      });
     }
   };
 
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -13,13 +13,6 @@ import { useQuery } from '../services/utils';
 import { fontSize, padding } from '@scality/core-ui/dist/style/theme';
 import CircleStatus from './CircleStatus';
 import { Button } from '@scality/core-ui';
-import {
-  LAST_SEVEN_DAYS,
-  LAST_ONE_HOUR,
-  QUERY_LAST_SEVEN_DAYS,
-  QUERY_LAST_TWENTY_FOUR_HOURS,
-  QUERY_LAST_ONE_HOUR,
-} from '../constants';
 import { intl } from '../translations/IntlGlobalProvider';
 
 const NodeListContainer = styled.div`
@@ -321,22 +314,6 @@ const NodeListTable = (props) => {
   const { nodeTableData, selectedNodeName } = props;
   const theme = useSelector((state) => state.config.theme);
 
-  let queryTimeSpan;
-  const metricsTimeSpan = useSelector(
-    (state) => state.app.monitoring.nodeStats.metricsTimeSpan,
-  );
-
-  switch (metricsTimeSpan) {
-    case LAST_SEVEN_DAYS:
-      queryTimeSpan = QUERY_LAST_SEVEN_DAYS;
-      break;
-    case LAST_ONE_HOUR:
-      queryTimeSpan = QUERY_LAST_ONE_HOUR;
-      break;
-    default:
-      queryTimeSpan = QUERY_LAST_TWENTY_FOUR_HOURS;
-  }
-
   const columns = React.useMemo(
     () => [
       {
@@ -412,17 +389,15 @@ const NodeListTable = (props) => {
       location.pathname.endsWith('volumes') ||
       location.pathname.endsWith('pods');
 
-    if (isTabSelected && location.pathname.endsWith('metrics')) {
-      // Keep the same timespan selected for metrics tab, when switch between the nodes
-      const currentTab = location?.pathname?.split('/')?.pop();
-      history.push(`/newNodes/${nodeName}/${currentTab}?from=${queryTimeSpan}`);
-    } else if (isTabSelected) {
-      // Keep the same tab selected when switch between the nodes
-      const currentTab = location?.pathname?.split('/')?.pop();
-      history.push(`/newNodes/${nodeName}/${currentTab}`);
+    if (isTabSelected) {
+      const newPath = location.pathname.replace(
+        // eslint-disable-next-line no-useless-escape
+        /\/newNodes\/[^\/]*\//,
+        `/newNodes/${nodeName}/`,
+      );
+      history.push(newPath);
     } else {
-      // Set Health tab as default tab
-      history.push(`/newNodes/${nodeName}/health`);
+      history.push({ pathname: `/newNodes/${nodeName}/health` });
     }
   };
 
--- a/ui/src/components/NodePageMetricsTab.js
+++ b/ui/src/components/NodePageMetricsTab.js
@@ -168,8 +168,10 @@ const NodePageMetricsTab = (props) => {
     type: 'temporal',
     axis: {
       // Refer to all the available time format: https://github.com/d3/d3-time-format#locale_format
-      // format: '%m/%d %H:%M',
-      format: '%H:%M', // when the timespan is `Last 24 hours`
+      format:
+        metricsTimeSpan === (LAST_ONE_HOUR || LAST_TWENTY_FOUR_HOURS)
+          ? '%H:%M'
+          : '%m/%d',
       // Boolean value that determines whether the axis should include ticks.
       ticks: true,
       tickCount: 4,
--- a/ui/src/components/NodePageMetricsTab.js
+++ b/ui/src/components/NodePageMetricsTab.js
@@ -1,12 +1,16 @@
 import React from 'react';
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
 import styled from 'styled-components';
 import {
   fontSize,
   padding,
   fontWeight,
 } from '@scality/core-ui/dist/style/theme';
-import { LineChart, Loader } from '@scality/core-ui';
+import { LineChart, Loader, Dropdown } from '@scality/core-ui';
+import {
+  updateNodeStatsAction,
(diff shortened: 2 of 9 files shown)


ui/src/components/NodePageMetricsTab.js
```diff
@@ -212,8 +239,48 @@ const NodePageMetricsTab = (props) => {
   };
   const lineConfig = { strokeWidth: 1.5 };
 
+  const metricsTimeSpanItems = [
+    {
+      label: LAST_SEVEN_DAYS,
+      onClick: () => {
+        dispatch(updateNodeStatsAction({ metricsTimeSpan: LAST_SEVEN_DAYS }));
+        updateMetricsGraph();
```
gdemonet: Really more a question than a suggestion, but wouldn't it be "cleaner" to handle this at the saga level?

Instead of a single "refresh" saga which is going to send this "update metrics" action periodically, we could have it use a `race` effect, between `take(TIMESPAN_UPDATED)` and `delay(REFRESH_PERIOD)`, making sure we won't have too many call sites for the chart update... WDYT?

ui/src/components/NodeListTable.js
```diff
@@ -390,8 +412,12 @@ const NodeListTable = (props) => {
       location.pathname.endsWith('volumes') ||
       location.pathname.endsWith('pods');
 
-    if (isTabSelected) {
-      // When switch between the nodes, keep the same tab selected
+    if (isTabSelected && location.pathname.endsWith('metrics')) {
+      // Keep the same timespan selected for metrics tab, when switch between the nodes
+      const currentTab = location?.pathname?.split('/')?.pop();
+      history.push(`/newNodes/${nodeName}/${currentTab}?from=${queryTimeSpan}`);
```
gdemonet: This means we loose the other query parameters (like the table filters)... But I see it's not just related to this change, because we had this bug before apparently.

Wouldn't it be simpler to just keep the exact same URL and switch only the Node name part?
Something like:

```javascript
const newPath = location.pathname.replace(/\/newNodes\/[^\/]*\//, `/newNodes/${nodeName}/`);
history.push(newPath);
```

If you really need to sanitize the query string, I think you'd be able to by using `URLSearchParams`.

ui/src/components/NodePageMetricsTab.js
```diff
@@ -245,6 +251,9 @@ const NodePageMetricsTab = (props) => {
       onClick: () => {
         dispatch(updateNodeStatsAction({ metricsTimeSpan: LAST_SEVEN_DAYS }));
         updateMetricsGraph();
+        history.push(
+          `/newNodes/${selectedNodeName}/metrics?from=${QUERY_LAST_SEVEN_DAYS}`,
+        );
```
gdemonet: Ditto, this approach cleans off the other parts of the querystring, which is not desired behaviour IMO

ui/src/containers/NodePageRSP.js
```diff
@@ -53,7 +56,7 @@ const NodePageRSP = (props) => {
   // we should initialize the `metricsTimeSpan` in redux base on the URL query.
   const query = useQuery();
   const queryTimespan = query.get('from');
-  let metricsTimeSpan;
+  let metricsTimeSpan = LAST_TWENTY_FOUR_HOURS;
```
gdemonet: Why this change? The `default` case below seems sufficient

ui/src/ducks/app/monitoring.js
```diff
@@ -87,6 +89,9 @@ const defaultState = {
   },
   nodeStats: {
     metricsTimeSpan: LAST_TWENTY_FOUR_HOURS,
+    instanceIP: '',
+    controlPlaneInterface: '',
+    workloadPlaneInterface: '',
```
gdemonet, line 109: I think it makes sense to maintain these values in the state (in essence, this represents the "selected node" for which we'll pull metrics). However, doing this should allow us to only set them once, when we mount the metrics tab component (because we know which Node is selected), and never touch them again.

In practice, it means that:

- when mounting / rendering the component, you dispatch this once:

```javascript
      updateNodeStatsAction({
        metricsTimeSpan,
        instanceIP,
        controlPlaneInterface,
        workloadPlaneInterface,
      })
```

- the fetch saga reads from the state this information about the selected node and timespan, not from the payload, so you dispatch this at render, after the previous one:

```javascript
    fetchNodeStatsAction()
```

- the refresh saga doesn't care about the state, it's only periodically sending a `FETCH_NODE_STATS` action until it receives a halt (see later comment on how to write the refresh saga)

ui/src/ducks/app/monitoring.js
```diff
@@ -725,6 +743,33 @@ export function* fetchNodeStats({ payload }) {
...
+      },
+    });
+  }
+  yield delay(REFRESH_METRCIS_GRAPH);
+
+  const isRefreshing = yield select(isNodeStatsRefreshing);
+  if (isRefreshing) {
+    yield call(refreshNodeStats);
+  }
+}
```
gdemonet: This pattern for refresh sagas was already discussed with Carlito, back in the days, and we came to the conclusion that it had big drawbacks (you can loose a "STOP" action, or trigger the refresh loop multiple times in parallel, and overall, it's not easy to read). See #2035.

In practice, with the suggestion from a previous comment, that would look like:

```javascript
export function* refreshNodeStats() {
  while (true) {
    yield take(REFRESH_NODESTATS);
    while (true) {
      yield fork(fetchNodeStats);
      const {interrupt, update} = yield race({
        interrupt: take(STOP_REFRESH_NODESTATS),
        // If the refresh period expires before we receive a halt,
        // we can refresh the stats
        requeue: delay(DELAY_TIMEOUT),
        // NOTE: This action doesn't exist, but it should, so that
        // whenever we change one of the parameters for "fetchNodeStats",
        // it gets triggered again
        update: take(UPDATE_SELECTED_NODE),
      });
      if (interrupt) {
        yield cancel(fetchNodeStats);
        break;
      }
    }
  }
}
```

ui/src/containers/NodePageRSP.js
```diff
@@ -53,19 +54,38 @@ const NodePageRSP = (props) => {
...
+      case LAST_SEVEN_DAYS:
+        queryTimespan = QUERY_LAST_SEVEN_DAYS;
+        break;
+      case LAST_ONE_HOUR:
+        queryTimespan = QUERY_LAST_ONE_HOUR;
+        break;
+      default:
+        queryTimespan = QUERY_LAST_TWENTY_FOUR_HOURS;
+    }
   }
```
gdemonet: Would be nicer to use maps, especially if we start adding more options.

gdemonet (changes requested):
Just some code review (mostly Sagas again :wink:), I'll test it soon!

ui/src/ducks/app/monitoring.js
ChengYanJin: hello @gdemonet ,

I wanted to address this dept ticket for a long time! Let's work on it now!

I have one questions in my mind:
What does `yield take(REFRESH_NODESTATS);` do? In this case, I think we may never have the `REFRESH_NODESTATS` dispatched.

gdemonet: `take(SOME_EVENT)` is blocking until `SOME_EVENT` is seen.

So, the idea is to replace the `takeFirst` effect creator:

```javascript
while (true) {
  // wait for the start event
  yield take("START_REFRESH");
  // once we've seen this start event, we can enter the refresh loop
  while (true) {
    // trigger the data update
    yield fork(fetchData);
    // wait for the stop event for at most the refresh period
    const {interrupt} = yield race({
      interrupt: take("STOP_REFRESH"),
      requeue: delay(REFRESH_PERIOD),
    });
    if (interrupt) {
      // we received the stop event, so we get out of the refresh loop
      // and we will wait for the next start event
      break;
    }
  }
}
```

ui/src/containers/NodePageRSP.js
ChengYanJin: Actually, and we set the default value in the saga state. 
Maybe it doesn't hurt to set again here. 

``` javascript
  const query = useQuery();
  const nodeMetricsTimeSpan = useSelector(
    (state) => state.app.monitoring.nodeStats.metricsTimeSpan,
  );

  let metricsTimeSpan;
  const queryTimespan = query.get('from');

  if (queryTimespan) {
    // If timespan query specified in query string
    metricsTimeSpan = queryTimeSpansCodes?.find(
      (timespan) => timespan.label === queryTimespan,
    )?.value;
  } else {
    metricsTimeSpan = nodeMetricsTimeSpan;
  }
```


```javascript
nodeStats: {
    metricsTimeSpan: LAST_TWENTY_FOUR_HOURS,
    ...
}
```

ui/src/components/NodePageMetricsTab.js
ChengYanJin: Definitely right! 
Will dispatch `updateNodeStatsFetchArgumentAction` to the `race` everytime we change the timespan or node(information)

ui/src/components/NodeListTable.js
```diff
@@ -391,12 +390,14 @@ const NodeListTable = (props) => {
       location.pathname.endsWith('pods');
 
     if (isTabSelected) {
-      // When switch between the nodes, keep the same tab selected
-      const currentTab = location?.pathname?.split('/')?.pop();
-      history.push(`/newNodes/${nodeName}/${currentTab}`);
+      const newPath = location.pathname.replace(
+        // eslint-disable-next-line no-useless-escape
```
gdemonet: If there is a useless escape, maybe we should fix it?

ui/src/components/NodeListTable.js
```diff
@@ -391,12 +390,14 @@ const NodeListTable = (props) => {
...
+        /\/newNodes\/[^\/]*\//,
+        `/newNodes/${nodeName}/`,
+      );
+      history.push(newPath);
     } else {
-      // Set Health tab as default tab
-      history.push(`/newNodes/${nodeName}/health`);
+      history.push({ pathname: `/newNodes/${nodeName}/health` });
```
gdemonet: We'll loose URL search params here as well :wink: 
In any case, I think this URL parsing in the component is not the right approach, we should instead rely on react-router's `Route` matching, that'd be much cleaner and safer. Let's open a debt ticket for this, we'll try to refactor other pages at the same time.

ui/src/containers/NodePageMetricsTab.js
```diff
@@ -57,11 +68,22 @@ const LoaderContainer = styled(Loader)`
...
+  justify-content: flex-end;
+  padding-right: ${padding.base};
+`;
+
 const NodePageMetricsTab = (props) => {
   const { nodeStats } = props;
+  const dispatch = useDispatch();
   const theme = useSelector((state) => state.config.theme);
+  const history = useHistory();
+  const query = new URLSearchParams(history?.location?.search);
```
gdemonet: Is there a difference between this and the `useQuery` hook? Why do we need to rely on `useHistory().location` rather than `useLocation()` directly?

ui/src/containers/NodePageMetricsTab.js
```diff
@@ -212,8 +236,65 @@ const NodePageMetricsTab = (props) => {
...
+    }
+  };
+
+  const metricsTimeSpanItems = [
+    {
+      label: LAST_SEVEN_DAYS,
+      onClick: () => {
+        dispatch(
+          updateNodeStatsAction({ metricsTimeSpan: LAST_SEVEN_DAYS }),
+          updateNodeStatsFetchArgumentAction(),
```
gdemonet: Would be a bit cleaner to differenciate between the two:

- `UPDATE_NODESTATS` is used to update `app.monitoring.nodeStats.metrics`
- `UPDATE_NODESTATS_FETCH_ARG` is used to update the other keys used as fetch arguments

This way, we could:

- only dispatch one action, instead of two
- never forget to send the `UPDATE_FETCH_ARG` action when needed, because that's the only way to actually update them
- rely on the main `UPDATE_NODESTATS` if we had other sagas that need to react on some metrics update

ui/src/containers/NodePageMetricsTab.js
```diff
@@ -212,8 +236,65 @@ const NodePageMetricsTab = (props) => {
...
+      onClick: () => {
+        dispatch(
+          updateNodeStatsAction({ metricsTimeSpan: LAST_ONE_HOUR }),
+          updateNodeStatsFetchArgumentAction(),
+        );
+        writeUrlTimeSpan(LAST_ONE_HOUR);
+      },
+      selected: metricsTimeSpan === LAST_ONE_HOUR,
+    },
+  ];
```
gdemonet: Could be refactored:

```suggestion
  const metricsTimeSpanItems = [
    LAST_SEVEN_DAYS, LAST_TWENTY_FOUR_HOURS, LAST_ONE_HOUR
  ].map(
    option => ({
      label: option,
      onClick: () => {
        dispatch(
          updateNodeStatsFetchArgumentAction({ metricsTimeSpan: option }),
        );
        writeUrlTimeSpan(option);
      },
      selected: metricsTimeSpan === option,
    })
  );
```

ui/src/containers/NodePageMetricsTab.js
```diff
@@ -212,8 +236,65 @@ const NodePageMetricsTab = (props) => {
...
+
+  const metricsTimeSpanDropdownItems = metricsTimeSpanItems?.filter(
+    (mTS) => mTS.label !== metricsTimeSpan,
+  );
```
gdemonet, line 263: I'm surprised, does it mean we only have one item in the dropdown?

gdemonet (changes requested):
Mostly a change on how we handle updating the "fetch args"

ui/src/components/NodeListTable.js
gdemonet: #2838

ui/src/containers/NodePageMetricsTab.js
ChengYanJin: Nope. The items we have in the dropdown are all the options **exclude** the current selected`metricsTimeSpan`

ui/src/components/NodeListTable.js
ChengYanJin: You are right! Thanks for creating the debt ticket!
We're gonna lose the search params if we want to add filters for the table.

ui/src/components/NodeListTable.js
ChengYanJin: It's actually an [eslint-error](https://stackoverflow.com/a/47277197) because of the regex: Unnecessary escape character: \/ no-useless-escape 
But I think the regex is correct. That's why I disable it, maybe the message is not easy to understand.

slaperche-scality: As mentioned in the answer you linked, you don't need to escape the forward slash inside the character class e.g. you could write `/\/newNodes\/[^/]*\//` and have the same behavior:

```
> var nodeName = "foo";
undefined
> var str = "/newNodes/BAR/";
undefined
> str.replace(/\/newNodes\/[^\/]*\//, `/newNodes/${nodeName}/`); 
'/newNodes/foo/'
> str.replace(/\/newNodes\/[^/]*\//, `/newNodes/${nodeName}/`); 
'/newNodes/foo/'
```

Did you try that version?
Is ESLint happy with this regex? Or do you still have the warning?

ChengYanJin: I confirm `/\/newNodes\/[^/]*\//` is working correctly!

ui/src/containers/NodePageMetricsTab.js
gdemonet: :man_facepalming: Oopsie, I read that too fast! Thanks for the translation :)

gdemonet (approved):
LGTM!

PR #2810, opened 2020-09-22
ui/volumes: Store metrics timespan in URL

### LINKED ISSUE #2754: [UI New Volume Page]Add metrics timespan as query param in the URL
**Component**: ui

**Why this is needed**:
image

We would like to add the timespan for metrics graph in URL, so that when we share the URL we know the timespan has been chosen.

**What should be done**:
Add one more query param in the URL for the three timespan choices.

Last 7days: `&from=now-7d`
Last 24hours: `&from=now-24h`
Last 1 hour: `&from=now-1h`

We should update the URL in `onClick()`
```
  const metricsTimeSpanItems = [
    {
      label: LAST_SEVEN_DAYS,
      onClick: () => {
        dispatch(updateVolumeStatsAction({ metricsTimeSpan: LAST_SEVEN_DAYS }));
        updateMetricsGraph();
      },
      selected: metricsTimeSpan === LAST_SEVEN_DAYS,
    },
    {
      label: LAST_TWENTY_FOUR_HOURS,
      onClick: () => {
        dispatch(
          updateVolumeStatsAction({ metricsTimeSpan: LAST_TWENTY_FOUR_HOURS }),
        );
        updateMetricsGraph();
      },
      selected: metricsTimeSpan === LAST_TWENTY_FOUR_HOURS,
    },
    {
      label: LAST_ONE_HOUR,
      onClick: () => {
        dispatch(updateVolumeStatsAction({ metricsTimeSpan: LAST_ONE_HOUR }));
        updateMetricsGraph();
      },
      selected: metricsTimeSpan === LAST_ONE_HOUR,
    },
  ];
```

**Test plan**:
TBD

alexis-ld:
**Component**: ui

**Context**: 

We want to store the Volume metrics timespan in the URL as a param query to allow users to share the URL while specifying a timespan.

**Summary**:

On timespan selection we write the param query to the URL following the specs described in the issue : 
* Last 7days: `&from=now-7d`
* Last 24hours: `&from=now-24h`
* Last 1 hour: `&from=now-1h`

On metrics card component load, we check if a `from` param query is defined. If it is set we dispatch an Update Volume Stats redux action followed by a data fetch.

To handle URL query strings parsing and generation I used the `query-string` package ( https://www.npmjs.com/package/query-string ). I found this package to be reliable and well-maintained (over 12M weekly downloads), however I can still switch to a custom handmade URL query param system if needed.

**Acceptance criteria**: 
* Changing a timespan on the Volume metrics updates the URL with query params
* On page load if a valid query param is set, the timespan is changed accordingly
* If other query params are set they are preserved 

---

Closes: #2754

### DIFF
--- a/ui/src/components/VolumeMetricGraphCard.js
+++ b/ui/src/components/VolumeMetricGraphCard.js
@@ -105,7 +105,7 @@ const NoDataGraphText = styled.div`
 `;
 
 const MetricGraphCard = (props) => {
-  const { volumeCondition, volumeMetricGraphData } = props;
+  const { volumeCondition, volumeMetricGraphData, volumeName } = props;
   const dispatch = useDispatch();
   const history = useHistory();
   const query = new URLSearchParams(history?.location?.search);
@@ -144,14 +144,19 @@ const MetricGraphCard = (props) => {
     const urlTimeSpan = queryTimeSpansCodes.find(
       item => item.label === query.get('from')
     );
+
+    // If a time span is specified we apply it
+    // Else if a timespan has been set but is not in the URL yet (change of volume) we write it to the URL
     if (urlTimeSpan) {
       dispatch(updateVolumeStatsAction({ metricsTimeSpan: urlTimeSpan.value }));
       updateMetricsGraph();
+    } else if (metricsTimeSpan !== LAST_TWENTY_FOUR_HOURS && !urlTimeSpan) {
+      writeUrlTimeSpan(metricsTimeSpan);
     }
   }
   
-  // set the timespan if specified in url query
-  useEffect(handleUrlQuery, []);
+  // handle timespan in url query
+  useEffect(handleUrlQuery, [volumeName]);
 
   const updateMetricsGraph = () => dispatch(fetchVolumeStatsAction());
 
--- a/ui/src/components/VolumeMetricGraphCard.js
+++ b/ui/src/components/VolumeMetricGraphCard.js
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useHistory } from 'react-router';
 import { useDispatch, useSelector } from 'react-redux';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


gdemonet:
Hello @alexis-ld!

> To handle URL query strings parsing and generation I used the `query-string` package

I think we used [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) somewhere else in our code, wouldn't it be sufficient?

alexis-ld:
@gdemonet Indeed. I missed that. Updating ... :)

alexis-ld:
@gdemonet Updated and squashed :)

ChengYanJin (changes requested):
Hello @alexis-ld,

I found there are two issues related to this PR:

1. When we switch from one volume to another, we would like to keep the same timespan selected not only in the dropdown but also in the URL.
2. When I set the URL with `?from=now-1h` I don't see the graphs with data anymore. Could you please investigate what is the problem here? 

Thank you.

alexis-ld:
@ChengYanJin 

Thanks for the feedback. Updated 

1. URL Timespan is now preserved when changing the selected volume.

2. I investigated the data disappearing when loading the page on  -1h timespan.

This occurred because the VolumePage component was launching a metrics fetch on mount with basic params. Then quickly after that the VolumeMetricGraphCard was launching a fetch when seeing a valid URL query param. This created a race condition. On -1h the dataset being way smaller than the default 24h the requests resolved before the previously fired request resulting in random behavior.

The main reason behind this was the use of `takeEvery` on metrics related saga watchers. Since we want the metrics corresponding to the latest user action and not the ones that resolve first (race condition) I changed both metrics related saga watchers to `takeLatest` to avoid concurrency and it fixed everything nicely 😄  .

ChengYanJin (approved):
Thanks for investigating the issue and fixing it. 
Good job!

thomasdanan (approved):

PR #2812, opened 2020-09-24
salt: Do not remove existing taints when running mark-control-plane

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

In "mark-control-plane" state we add taints for bootstrap and infra,
but since "spec.taints" is a list in a node object, when we do a "basic"
update on these taints we remove all existing taints, which is wrong.

**Summary**:

Change the state so that we add some logic in Jinja to check existing
taints (and also do not add taints if the node already have the "node"
role).
We create a new Jinja file outside of the sls, so that the Jinja is
rendered right before executing it and not during sls rendering


gdemonet (approved):
LGTM

PR #2815, opened 2020-09-28
ui/volumes: Replace Prometheus alerts by AlertManager

### LINKED ISSUE #2793: [UI New Volume Page] Retrieve the alerts from AlertManager API
**Component**: ui, monitoring

**Why this is needed**:
Retrieve the alerts from Alert Manager 

**What should be done**:
Currently, we retrieve the alerts of volumes from Prometheus. 
To import the AlertManager service should be done in [PR#2787](https://github.com/scality/metalk8s/pull/2787)
```
export function getAlerts() {
  return alertmanagerApiClient.get('/api/v2/alerts');
}
```
**To be confirmed:**
Should we still filter the alerts for volumes base on the PVC?

Retrieve the alerts by the Alertmanager API- getAlerts() function

**Test plan**:
TBD

alexis-ld:
**Component**: ui, volumes, monitoring

**Context**: 

Currently the alerts from the volumes page come from the Prometheus API. We want to switch to use AlertManager's API instead.

**Summary**:

The AlertManager service was already implemented. I added a few Redux actions to replicate the current refresh behavior we had with Prometheus.

**Acceptance criteria**: 

* Alerts on the volume page are retrieved from AlertManager's API
* Alerts are displayed properly

---

Closes: #2793

### DIFF
--- a/ui/src/services/NodeVolumesUtils.js
+++ b/ui/src/services/NodeVolumesUtils.js
@@ -137,7 +137,7 @@ export const volumeGetError = (status) => {
 
 const getPVList = (state) => state?.app?.volumes?.pVList;
 const getPVCList = (state) => state?.app?.volumes?.pVCList;
-const getAlerts = (state) => state?.app?.monitoring?.alert;
+const getAlerts = (state) => state?.app?.alerts;
 const getNodes = (state) => state?.app?.nodes?.list;
 
 const getVolumeLatencyCurrent = (state) =>
--- a/ui/src/components/VolumeActiveAlertsCard.js
+++ b/ui/src/components/VolumeActiveAlertsCard.js
@@ -74,7 +74,7 @@ const ActiveAlertsCard = (props) => {
       name: alert.labels.alertname,
       severity: alert.labels.severity,
       alert_description: alert.annotations.message,
-      active_since: alert.activeAt,
+      active_since: alert.startsAt,
     };
   });
   // React Table for the volume list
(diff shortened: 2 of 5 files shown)


ChengYanJin (changes requested):
Hello @alexis-ld,

Thanks for working on this, there is one thing you may miss.
We actually compute the `Health` in the volume list table base on the alerts.

So we need to update  **getAlerts** Selector in getVolumeListData() too. 
Suggest change:
`const getAlerts = (state) => state?.app?.alerts;`

Thank you!

alexis-ld:
@ChengYanJin Indeed, I missed that :) thank you for the feedback.

Updated and squashed.

ui/src/services/NodeVolumesUtils.js
```diff
@@ -137,7 +137,7 @@ export const volumeGetError = (status) => {
 
 const getPVList = (state) => state?.app?.volumes?.pVList;
 const getPVCList = (state) => state?.app?.volumes?.pVCList;
-const getAlerts = (state) => state?.app?.monitoring?.alert;
+const getAlerts = (state) => state?.app?.alerts;
```
ChengYanJin, line 140: Sorry my mistake. 
```suggestion
const getAlerts = (state) => state?.app?.alerts?.list;
```

ChengYanJin, line 140: Hello @alexis-ld ,

Sorry, we still miss one thing that I forgot last time. 
We actually have a unit test for `getVolumeListData` function. It returns the data used by Volume Table. 
We may want to update the `stateApp` that used by the unit test.

```javascript
const stateApp = {
  app: {
    nodes: {...},
    pods: {...},
    volumes: {...},
    monitoring: {...},
    // we may want to add this.
    alerts: { list : [] }, 
}
```

I would suggest you run the unit test `npm run test` to see if the new changes will break the test or not.
Thank you!

ChengYanJin (changes requested):
Hello @alexis-ld ,

One last change then we are good to go! 
Thank you!

alexis-ld:
@ChengYanJin Adapted tests and squashed commits. :)

ChengYanJin (approved):
LGTM!

thomasdanan (approved):

PR #2819, opened 2020-09-30
[UI] Add the pod tab in new node page

### LINKED ISSUE #2778: [UI New Node Page] Add pod list in Node page
**Component**:

**Why this is needed**:

**What should be done**:
implement the pod list as designed

> A pod can be more than just one container, and each has a different status (when using kubectl, we see a READY column, in which we have the count of ready containers, e.g. 2/3)

 the status shows the information of ready containers, and color should be in yellow for quickly seeing not "100% running/available" values.

image
compare to what do we have today:
image

Update the new design:
05-03-node-v3-pods@2x

Status field: 
Current status of the pod plus the status of the containers.
 
The URL to acess the log of the pod:
`https://192.168.1.18:8443/grafana/d/a7e130cb82be229d6f3edbfd0a438001/logs?orgId=1&var-logs=Loki&var-logmetrics=Prometheus&var-metrics=Prometheus&var-podlogs=.*&var-systemlogs=.*&var-cluster=&var-node=All&var-instance=All&var-namespace=All&var-deployment=calico-kube-controllers&var-pod=kube-scheduler-master-1&var-container=All&var-hostname=All&var-unit=All&var-filter=`

`${config.api.url_grafana}/dashboard/db/logs?orgId=1&var-logs=Loki&var-logmetrics=Prometheus&var-metrics=Prometheus&var-podlogs=.*&var-systemlogs=.%2B&var-deployment=calico-kube-controllers&var-pod=${value}`

**Test plan**:
TBD

ChengYanJin:
**Component**: ui, nodes

**Context**:  Add pod tab in the New Node Page.

**Summary**:
- When we hover on the graph icon in the log column, we can see the URL at the button left. 
- When we click on the graph icon, we should be redirected to the `logs dashboard` in Grafana filer by the current pod. 
Note that, another tab in the browser should be open. 
- The color of the `Status` is either yellow or green, depending on if all the containers running.
- The format of the Age is displayed refer to the `kubectl get pod`.

**Acceptance criteria**: 
image

---

Closes: #2778

### DIFF
--- a/ui/src/services/PodUtils.js
+++ b/ui/src/services/PodUtils.js
@@ -1,0 +1,26 @@
+import { fromMilliSectoAge } from './utils.js';
+
+// Return the pod list for the Pod Tab in Node Page
+export const getPodsListData = (nodeName, pods) => {
+  const podsList = pods?.filter((pod) => pod.nodeName === nodeName);
+
+  return (
+    podsList?.map((pod) => {
+      const age = fromMilliSectoAge(new Date() - pod.startTime);
+
+      const numContainer = pod.containerStatuses.length;
+      const containerReady =
+        pod?.containerStatuses?.filter((pCS) => pCS.ready === true) ?? [];
+      return {
+        name: pod.name,
+        age: age,
+        namespace: pod.namespace,
+        status: {
+          status: `${pod.status} (${containerReady.length}/${numContainer})`,
+          isAllContainerRunning: containerReady.length === numContainer,
+        },
+        log: pod.name,
+      };
+    }) ?? []
+  );
+};
--- a/ui/src/services/utils.js
+++ b/ui/src/services/utils.js
@@ -329,3 +329,35 @@ export const fromUnixTimestampToDate = (date) => {
     return new Date(date * 1000);
   }
 };
+
+// Convert the timestamp from milliseconds to Age.
+// For example: 1h1s, 1d1h, 1d1s, 1d
+export const fromMilliSectoAge = (milliSecTime) => {
+  if (milliSecTime >= 1000) {
+    let day, hour, minute, second;
+    const age = [];
+
+    second = Math.floor(milliSecTime / 1000);
+    minute = Math.floor(second / 60);
+    hour = Math.floor(minute / 60);
+    day = Math.floor(hour / 24);
+
+    second = second % 60;
+    minute = minute % 60;
+    hour = hour % 24;
+
+    if (day > 0) {
+      age.push(day + 'd');
+    }
+    if (hour > 0) {
+      age.push(hour + 'h');
+    }
+    if (minute > 0) {
+      age.push(minute + 'm');
+    }
+    if (second > 0) {
+      age.push(second + 's');
+    }
+    return age.slice(0, 2).join('');
+  }
+};
--- a/ui/src/components/NodePagePodsTab.js
+++ b/ui/src/components/NodePagePodsTab.js
@@ -4,6 +4,7 @@ import { useTable } from 'react-table';
 import styled from 'styled-components';
 import { fontSize, padding } from '@scality/core-ui/dist/style/theme';
 import { TabContainer } from './CommonLayoutStyle';
+import { intl } from '../translations/IntlGlobalProvider';
 
 const PodTableContainer = styled.div`
   color: ${(props) => props.theme.brand.textPrimary};
@@ -106,7 +107,15 @@ function Table({ columns, data }) {
                     ...cell.column.cellStyle,
                   },
                 });
-                return <Cell {...cellProps}>{cell.render('Cell')}</Cell>;
+                if (cell.column.Header !== 'Name' && cell.value === undefined) {
+                  return (
+                    <Cell {...cellProps}>
+                      <div>{intl.translate('unknown')}</div>
+                    </Cell>
+                  );
+                } else {
+                  return <Cell {...cellProps}>{cell.render('Cell')}</Cell>;
+                }
               })}
             </TableRow>
           );
--- a/ui/src/components/NodePagePodsTab.js
+++ b/ui/src/components/NodePagePodsTab.js
@@ -1,8 +1,179 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
+import { useTable } from 'react-table';
+import styled from 'styled-components';
+import { fontSize, padding } from '@scality/core-ui/dist/style/theme';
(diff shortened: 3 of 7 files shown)


ui/src/services/PodUtils.js
```diff
@@ -0,0 +1,52 @@
...
+      name: pod.name,
+      age: age,
+      namespace: pod.namespace,
+      status: {
+        status: `${pod.status} (${containerReady.length}/${numContainer})`,
+        isAllContainerRunning: containerReady.length === numContainer,
+      },
+      log: pod.name,
+    };
+  });
```
MonPote: I guess you want the default value to be an array.
If `pods` is `null` or `undefined` you will return null but I believe you want an array.
So you might want to do a
```
}) ?? [];
```

ui/src/services/PodUtils.js
```diff
@@ -0,0 +1,52 @@
...
+      age = ageObject.hour + 'h';
+    } else if (ageObject.mintue > 0 && ageObject.second > 0) {
+      // start with `m`
+      age = ageObject.minute + 'm' + ageObject.second + 's';
+    } else if (ageObject.mintue > 0) {
+      age = ageObject.minute + 'm';
+    } else if (ageObject.second > 0) {
+      // start with `s`
+      age = ageObject.second + 's';
+    }
```
MonPote: I'm confident this logic should be extract and unit tested.
I'm not sure if this can be easily "simplify" so even though it does not look very smart. It's easily understandable so let's keep it that way.
Also maybe provide some example will help to quickly understand the purpose.

ui/src/services/PodUtils.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const numContainer = pod.containerStatuses.length;
+    const containerReady = pod?.containerStatuses?.filter(
+      (pCS) => pCS.ready === true,
+    );
```
MonPote:
```
) ?? [];
```
Or line `46` and `47` may crash.

ui/src/services/utils.js
```diff
@@ -329,3 +329,26 @@ export const fromUnixTimestampToDate = (date) => {
     return new Date(date * 1000);
   }
 };
+
+// Convert the timestamp in milliseconds to an object with day, hour, minute and second
+export const fromMilliSectoAge = (milliSecTime) => {
+  if (milliSecTime && milliSecTime > 0) {
```
MonPote: The first `milliSecTime` check is useless.

ui/src/services/utils.js
```diff
@@ -329,3 +329,26 @@ export const fromUnixTimestampToDate = (date) => {
...
+    minute = minute % 60;
+    hour = hour % 24;
+
+    return {
+      day: day,
+      hour: hour,
+      minute: minute,
+      second: second,
+    };
+  }
```
MonPote, line 362: What happen when `milliSecTime` is `null` or `undefined` ?
Do you chose to return `undefined` or you want to return something like this ?
```javascript
{
      day: 0,
      hour: 0,
      minute: 0,
      second: 0,
};
```
Or something else ?

ui/src/services/PodUtils.js
slaperche-scality: +1 for the unit test, that would catch:
- the typo (`s/mintue/minute/`) l.27 and l. 30
- the copy/paste error (`d` should be `h` l.22 and l.24)

And maybe other errors.

What if all fields are zero (unlikely but not impossible)? What will you return?

As for simplification: you could write that:

```javascript
  const age = [];
  
  if (ageObject.day > 0) {
    age.push(ageObject.day + 'd');
  }
  if (ageObject.hour > 0) {
    age.push(ageObject.hour + 'h');
  }
  if (ageObject.minute > 0) {
    age.push(ageObject.minute + 'm');
  }
  if (ageObject.second > 0) {
    age.push(ageObject.second + 's');
  }
  return age.slice(0, 2).join('');
```
=> Add every non-zero value and only take the two first.

ui/src/services/PodUtils.js
```diff
@@ -0,0 +1,52 @@
...
+    } else if (ageObject.day > 0) {
+      age = ageObject.day + 'd';
+    } else if (ageObject.hour > 0 && ageObject.minute > 0) {
+      // start with `h`
+      age = ageObject.hour + 'd' + ageObject.minute + 'm';
+    } else if (ageObject.hour > 0 && ageObject.second > 0) {
+      age = ageObject.hour + 'd' + ageObject.second + 's';
+    } else if (ageObject.hour > 0) {
+      age = ageObject.hour + 'h';
+    } else if (ageObject.mintue > 0 && ageObject.second > 0) {
```
gdemonet:
```suggestion
    } else if (ageObject.minute > 0 && ageObject.second > 0) {
```

ui/src/services/PodUtils.js
```diff
@@ -0,0 +1,52 @@
...
+      // start with `m`
+      age = ageObject.minute + 'm' + ageObject.second + 's';
+    } else if (ageObject.mintue > 0) {
```
gdemonet:
```suggestion
    } else if (ageObject.minute > 0) {
```

gdemonet (dismissed):
And that's why static type checking is useful ;)

MonPote:
> And that's why static type checking is useful ;)

Oh... I missed that and It's actually not the first time.
I'm beginning to reconsider if it will not bring more value than the cost to maintain it.
In this specific case the unit test would have catch it but if it was in other places static type checking may have a lot of sense.

ui/src/services/utils.js
ChengYanJin: Indeed!

ui/src/services/utils.js
ChengYanJin: I would return `undefined`. And let the pod table to address what should display. In this case, we should have an `unknown` displayed in the Age column.

Thanks to remind me!

ui/src/services/PodUtils.js
ChengYanJin: Simplified this `if/else if` by Sylvain's suggestion. 
Extract all the logic inside `fromMilliSectoAge` and added a unit test for this function.

ui/src/services/PodUtils.js
ChengYanJin: Thanks for this smart and neat solution!

MonPote (approved):

Issue #2794, opened 2020-09-14
Salt `metalk8s_kubernetes.replace_object` on clusterIP Service report false positif


TeddyAndrieux:
**Component**:

'salt', 'kubernetes'

**What happened**:

If we call `metalk8s_kubernetes.replace_object` with an "old_object" argument containing the current deployed object (it's automatically done by the salt state `metalk8s_kubernetes.object_present`) and we try to update the clusterIP we got a `The object was replaced` where it's not updated at all

**What was expected**:

An error because we cannot replace clusterIP it's immutable (user need to first remove this service manually)

**Steps to reproduce**

Create a simple Service with a clusterIP and try to replace just this "clusterIp" field with a `metalk8s_kubernetes.object_present` salt state.

**Resolution proposal** (optional):

In `metalk8s_kubernetes` salt module, check that we do not have any clusterIP (or an empty one) in the body before replacing it with the one in the "old_object".
So the request to apiServer will report the right error at the end

PR #2833, opened 2020-10-06
salt: Fix Kubernetes `replace_object` for Service with ClusterIP

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

#2794 

**Summary**:

Before this commit when we tried to replace a Service with `clusterIP`
the `clusterIP` were always replaced with the one from old_object, which
is wrong, if we provide a `clusterIP` in the input manifest we should
keep this one and not replace it with the one from old_object.
So that we let apiServer reject this call and not report a "Success"
where at the end `clusterIP` not get updated at all

---

Fixes: #2794

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -195,7 +195,9 @@ def _object_manipulation_function(action):
             if 'resourceVersion' in old_object['metadata']:
                 call_kwargs['body'].metadata.resourceVersion = \
                     old_object['metadata']['resourceVersion']
-            if obj.api_version == 'v1' and obj.kind == 'Service':
+            # Keep `cluster_ip` if not present in the body
+            if obj.api_version == 'v1' and obj.kind == 'Service' \
+                    and not call_kwargs['body'].spec.cluster_ip:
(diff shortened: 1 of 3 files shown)


salt/_modules/metalk8s_kubernetes.py
```diff
@@ -195,7 +195,10 @@ def method(manifest=None, name=None, kind=None, apiVersion=None,
             if 'resourceVersion' in old_object['metadata']:
                 call_kwargs['body'].metadata.resourceVersion = \
                     old_object['metadata']['resourceVersion']
-            if obj.api_version == 'v1' and obj.kind == 'Service':
+            # Keep `cluster_ip` if not present in the body
+            if obj.api_version == 'v1' and obj.kind == 'Service' \
+                    and not call_kwargs['body'].spec.cluster_ip:
+                print(call_kwargs['body'].spec.cluster_ip)
```
gdemonet: I think you can remove this `print`

TeddyAndrieux: Oh ! Shit ! Right

gdemonet (approved):

PR #2834, opened 2020-10-06
[UI] Add node overview tab in new node page

### LINKED ISSUE #2774: [UI New Node Page] Add Overview Tab in the New Node Page
**Component**: ui, nodes

**Why this is needed**:

**What should be done**:
Following the rework the design of Overview tab.

Moonshoot - User workflow

- Remove the [delete] & [drain] button 
- Replace last updated by created
- Remove Region & Zone
- Remove the health status bar
- Add taints
```
[root@zenkotda-bootstrap centos]# kubectl describe node bootstrap | grep -A 2 Taints
Taints:             node-role.kubernetes.io/bootstrap:NoSchedule
                    node-role.kubernetes.io/infra:NoSchedule
                    node.kubernetes.io/unschedulable:NoSchedule
```
- Display separately CP IP and WP IP

**We are waiting for the new design of Overview Tab in Node Page, will resume the work after the design**

**Test plan**:
TBD

ChengYanJin:
**Component**: ui, nodes

**Context**: 

**Summary**:
- There is an alert filter component missing. Will do it in Alert tab PR.
- The format of the creation time does not update yet. We need to first upgrade the react-intl to V3+ to bring the [FormattedDateParts](https://formatjs.io/docs/react-intl/components/#formatteddateparts) API available. Since there are a lot of breaking changes, will do it in another PR.

**Acceptance criteria**: 
image

Thanks for the review @thomasdanan 
Please find the update:
image
image

---

Closes: #2774

### DIFF
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import { useHistory } from 'react-router';
-import { useLocation } from 'react-router-dom';
+import { useLocation, useRouteMatch } from 'react-router-dom';
 import styled from 'styled-components';
 import {
   useTable,
@@ -305,6 +305,7 @@ const NodeListTable = (props) => {
   const { nodeTableData, selectedNodeName } = props;
   const theme = useSelector((state) => state.config.theme);
   const query = useQuery();
+  const { path } = useRouteMatch();
 
   const columns = React.useMemo(
     () => [
@@ -360,9 +361,10 @@ const NodeListTable = (props) => {
       location.pathname.endsWith('pods');
 
     if (isTabSelected) {
+      // TODO: Need to change the Regex when rename to /nodes
       const newPath = location.pathname.replace(
         /\/newNodes\/[^/]*\//,
-        `/newNodes/${nodeName}/`,
+        `${path}/${nodeName}/`,
       );
       history.push({
         pathname: newPath,
--- a/ui/src/components/NodeListTable.js
+++ b/ui/src/components/NodeListTable.js
@@ -118,19 +118,9 @@ const IPText = styled.span`
   color: ${(props) => props.theme.brand.textSecondary};
 `;
 
-// the color of the status depends on the `Status` and `Condition` of the Node
 const StatusText = styled.div`
   color: ${(props) => {
-    switch (props.textColor) {
-      case 'green':
-        return props.theme.brand.healthy;
-      case 'yellow':
-        return props.theme.brand.warning;
-      case 'red':
-        return props.theme.brand.critical;
-      default:
-        return props.theme.brand.textSecondary;
-    }
+    return props.textColor;
   }};
 `;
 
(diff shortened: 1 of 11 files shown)


thomasdanan (changes requested):
Hello @ChengYanJin, here are few general comments:

When selecting a tab, I see it should be in a "light" gray (compare to the other tabs which should be dark)
The background of the tab content is not as per it is in miro.
When the left nav bar is expanded, the separation between the left zone and the sight zone can be improved.

All the rest is perfect

thomasdanan (dismissed):

ui/src/containers/NodePageRSP.js
```diff
@@ -127,11 +138,19 @@ const NodePageRSP = (props) => {
 
   return (
     <NodePageRSPContainer>
-      <Tabs items={items} activeTabColor={theme.brand.primaryDark1}>
+      <Tabs items={items}>
         <Switch>
           <Route
-            path={`/newNodes/${selectedNodeName}/health`}
-            component={NodePageHealthTab}
+            path={`/newNodes/${selectedNodeName}/overview`}
```
MonPote: I think you should do something like this (nested route) instead of hardcoded.
https://reactrouter.com/web/example/nesting
That way, when you will get ride of the old /nodes, you don't need to change a single line code except.
```diff
- <PrivateRoute path="/newNodes" component={NodePage} />
+ <PrivateRoute path="/nodes" component={NodePage} />
```

In your case, it should be something like this :
```js
let { path, url } = useRouteMatch();
<Route
  path={`${path}/overview`}
  render={() => (
     <NodePageOverviewTab
       selectedNodeName={selectedNodeName}
       nodeTableData={nodeTableData}
       nodes={nodes}
       volumes={volumes}
       pods={pods}
    />
  )}
/>
```

ui/src/components/NodeListTable.js
```diff
@@ -343,37 +333,16 @@ const NodeListTable = (props) => {
...
-                {intl.translate('not_ready')}
-              </StatusText>
-            );
-          } else {
+          const { statusColor, computedStatus } = cellProps.value;
+          return computedStatus.map((status) => {
             return (
-              <StatusText textColor="gray">
-                {intl.translate('unknown')}
+              <StatusText textColor={statusColor}>
```
MonPote, line 345: You should have a key field (because of the map).
I wonder why react didn't show you a warning.

ui/src/containers/NodePageRSP.js
```diff
@@ -100,8 +111,8 @@ const NodePageRSP = (props) => {
   const items = [
     {
       selected: isHealthTabActive,
-      title: 'Health',
-      onClick: () => history.push(`/newNodes/${selectedNodeName}/health`),
+      title: 'Overview',
+      onClick: () => history.push(`/newNodes/${selectedNodeName}/overview`),
```
MonPote: See next comment.

MonPote (changes requested):
I have some `CORS` issue on my ptf so I didn't have the opportunity to test correctly this PR.
I've some comments on React Router, please have a look at it :)

ui/src/containers/NodePageRSP.js
ChengYanJin: Indeed! That's a smart move!

gdemonet: cf. #2838 :wink:

MonPote (dismissed):

MonPote (approved):

PR #2835, opened 2020-10-07
[UI] New volumes page

### LINKED ISSUE #2751: [UI New Volume Page] Add tabs for the Volume Right Side Panel
**Component**: UI

**Why this is needed**:
Since we want to implement tabs for the Node page, we should keep the same UX for Volume Page. 

**What should be done**:
- The current Right-side panel should be replaced with a `tab` component that includes `Overview, Alerts, Metrics and details`.
- The URL should be updated when a specific tab is selected:
 `volumes/<volume-name>/overview`
 `volumes/<volume-name>/alerts`
 `volumes/<volume-name>/metrics`
 `volumes/<volume-name>/details`
- When selecting a volume from the left side volume table, we have the Overview tab display in the right-side panel by default.
- When switching between different volumes, we want to keep the same tab selected (and the same time selector when available).

### overview tab design:
05-04-volumes_v9-overview@2x

When clicking on Critical, Warning, and Info on the right, the user is redirected to the Alert tab (which is filtered based on the Severity)
The usage bar color has to be changed according to this screenshot
The Node value is clickable and redirects the user to the Node Page
Also we should not implement Delete button as it is not yet supported
The Info severity is to be mapped with None severity when collecting alerts from AlertManager API

### Alerts tab design:
05-04-volumes_v9-Alerts@2x
- User can filter the list of active alerts on Severity criteria

Please note the following  **Update**:
- Remove the Info Severity
- When click on the alert tab, the URL should be `/newNodes/<node-name>/alerts`
- When we add the severity filer, the URL should be `/newNodes/<node-name>/alerts?severity=critical&severity=warning`
- We should be able to reuse the `alert filter component` since it should appear in Node Alert Tab too. 
image
- **Rename** the `Alert Description` column to `Description`

### Metrics tab design:
05-04-volumes_v9-performance@2x
Please note that: 
- To access the Advanced Metrics should be done in another ticket(#2750) 

``` javasript
          const config = useSelector((state) => state.config);

          <a
            href={`${config.api.url_grafana}/grafana/dashboard/db/kubernetes-persistent-volumes`}
            target="_blank"
            rel="noopener noreferrer"
          >
          </a>
```
- Ideally, when switching between the tabs, no reloading is needed.
- When switching to another Volume, the time span selection should remain the same

### Details tab design:
05-04-volumes_v9-details@2x

The content is retrieved from k8s API - [getClusterCustomObject method](https://user-images.githubusercontent.com/18453133/95017469-c44a6d00-0659-11eb-9a1d-01725ae543f3.png)

Suggestion:
```
export async function getVolumeObject(volumeName) {
  try {
    return await customObjects.getClusterCustomObject(
      'storage.metalk8s.scality.com',
      'v1alpha1',
      'volumes',
      volumeName,
    );
  } catch (error) {
    return { error };
  }
}
```

Please use the `tabs` component from [core-ui library](https://github.com/scality/core-ui)

alexis-ld:
**Component**:

ui, volumes

**Context**: 

We want to adapt the volumes page so it looks like the new node page (tab based).

**Summary**:

I refactored most of the Volume view. There were also small adaptations compared to the base view (alerts filtering in URL, ...) as described in #2751 .

Some components (ActiveAlertsFilters, ActiveAlertsCounter) were designed to be re-usable in other views.

EDIT : Missing tabs (metrics and details) were added in #2866 targeting this branch.

Screenshot 2020-10-15 at 15 41 13
Screenshot 2020-10-15 at 15 41 26
Screenshot 2020-10-15 at 15 41 46
Screenshot 2020-10-15 at 15 41 59

**Acceptance criteria**: 

* Volumes page right panel is separated in four tabs Overview, Alerts, Metrics, Details
* Total number of alerts is displayed in Alerts tab header
* It is possible to filter by alert severity on the Alert tab
* Clicking on an alert category in the overview tab opens the alerts tab and filters by the corresponding category
* When switching to another volume the selected tab is preserved
---

Closes: #2751

### DIFF
--- a/ui/src/components/ActiveAlertsCounter.js
+++ b/ui/src/components/ActiveAlertsCounter.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useHistory, useLocation } from 'react-router';
+import { padding, fontSize } from '@scality/core-ui/dist/style/theme';
+import { STATUS_WARNING, STATUS_CRITICAL } from '../constants.js';
+
+export const CountersWrapper = styled.div`
+  color: ${(props) => props.theme.brand.textPrimary};
+  display: flex;
+  justify-content: space-around;
+`;
+
+export const CounterWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  cursor: pointer;
+`;
+
+export const CounterValueWrapper = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+export const CounterTitle = styled.div`
+  padding: ${padding.smaller} 0px;
+  text-align: center;
+  font-size: ${fontSize.small};
+`;
+export const CounterValue = styled.div`
+  font-size: ${fontSize.larger};
+  padding-left: ${padding.smaller};
+`;
+
+export const CounterIcon = styled.i`
+  color: ${(props) => {
+    const theme = props.theme.brand;
+    let color = theme.textPrimary;
+
+    switch (props.status) {
+      case STATUS_WARNING:
+        color = theme.warning;
+        break;
+      case STATUS_CRITICAL:
+        color = theme.danger;
+        break;
+      default:
+        color = theme.textPrimary;
+    }
+    return color;
+  }};
+`;
+
+const ActiveAlertsCounter = (props) => {
+  const { alerts, baseLink } = props;
+  const history = useHistory();
+  const location = useLocation();
+
+  const getLink = (status) => {
+    const query = new URLSearchParams(location.search);
+    const existing = query.getAll('severity');
+    if (existing.indexOf(status) === -1) {
+      query.set('severity', status);
+    }
+    return `${baseLink}?${query.toString()}`;
+  };
+
+  const criticalCounter = alerts?.filter(
+    (item) => item?.labels?.severity === STATUS_CRITICAL,
+  ).length;
+  const warningCounter = alerts?.filter(
+    (item) => item?.labels?.severity === STATUS_WARNING,
+  ).length;
+
+  return (
+    <CountersWrapper>
+      <CounterWrapper onClick={() => history.push(getLink(STATUS_CRITICAL))}>
+        <CounterTitle>Critical</CounterTitle>
+        <CounterValueWrapper>
+          <CounterIcon
+            className="fas fa-times-circle"
+            status={STATUS_CRITICAL}
+          />
+          <CounterValue>{criticalCounter}</CounterValue>
+        </CounterValueWrapper>
+      </CounterWrapper>
+      <CounterWrapper onClick={() => history.push(getLink(STATUS_WARNING))}>
+        <CounterTitle>Warning</CounterTitle>
+        <CounterValueWrapper>
+          <CounterIcon
+            className="fas fa-exclamation-triangle"
+            status={STATUS_WARNING}
+          />
+          <CounterValue>{warningCounter}</CounterValue>
+        </CounterValueWrapper>
+      </CounterWrapper>
+    </CountersWrapper>
+  );
+};
+
+export default ActiveAlertsCounter;
--- a/ui/src/components/ActiveAlertsCounter.js
+++ b/ui/src/components/ActiveAlertsCounter.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { useHistory, useLocation } from 'react-router';
+import { useHistory, useLocation, useRouteMatch } from 'react-router';
 import { padding, fontSize } from '@scality/core-ui/dist/style/theme';
 import { STATUS_WARNING, STATUS_CRITICAL } from '../constants.js';
 
@@ -51,9 +51,10 @@ export const CounterIcon = styled.i`
 `;
 
 const ActiveAlertsCounter = (props) => {
-  const { criticalCounter, warningCounter, baseLink } = props;
+  const { criticalCounter, warningCounter } = props;
   const history = useHistory();
   const location = useLocation();
+  const match = useRouteMatch();
 
   const getLink = (status) => {
     const query = new URLSearchParams(location.search);
@@ -61,7 +62,7 @@ const ActiveAlertsCounter = (props) => {
     if (existing.indexOf(status) === -1) {
       query.set('severity', status);
     }
-    return `${baseLink}?${query.toString()}`;
+    return `${match.url.replace(/\/overview/, '/alerts')}?${query.toString()}`;
   };
 
   return (
--- a/ui/src/components/ActiveAlertsCounter.js
+++ b/ui/src/components/ActiveAlertsCounter.js
@@ -51,7 +51,7 @@ export const CounterIcon = styled.i`
 `;
 
 const ActiveAlertsCounter = (props) => {
-  const { alerts, baseLink } = props;
+  const { criticalCounter, warningCounter, baseLink } = props;
   const history = useHistory();
   const location = useLocation();
 
@@ -63,13 +63,6 @@ const ActiveAlertsCounter = (props) => {
     }
     return `${baseLink}?${query.toString()}`;
   };
-
-  const criticalCounter = alerts?.filter(
-    (item) => item?.labels?.severity === STATUS_CRITICAL,
-  ).length;
-  const warningCounter = alerts?.filter(
-    (item) => item?.labels?.severity === STATUS_WARNING,
-  ).length;
 
   return (
     <CountersWrapper>
--- a/ui/src/components/ActiveAlertsCounter.js
+++ b/ui/src/components/ActiveAlertsCounter.js
@@ -38,7 +38,7 @@ export const CounterIcon = styled.i`
 
     switch (props.status) {
       case STATUS_WARNING:
-        color = theme.warning;
+        color = theme.alert;
         break;
       case STATUS_CRITICAL:
         color = theme.danger;
--- a/ui/src/components/ActiveAlertsFilters.js
+++ b/ui/src/components/ActiveAlertsFilters.js
@@ -1,0 +1,88 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Dropdown } from '@scality/core-ui';
+import { useHistory } from 'react-router';
+import { useQuery } from '../services/utils';
+import { padding } from '@scality/core-ui/dist/style/theme';
+import { STATUS_WARNING, STATUS_CRITICAL } from '../constants.js';
+
+export const FilterIcon = styled.i`
+  color: ${(props) => {
+    const theme = props.theme.brand;
+    let color = theme.textPrimary;
+
+    switch (props.status) {
+      case STATUS_WARNING:
+        color = theme.warning;
+        break;
+      case STATUS_CRITICAL:
+        color = theme.danger;
+        break;
+      default:
+        color = theme.textPrimary;
+    }
+    return color;
+  }};
+  padding: ${padding.smaller};
+`;
+
+const ActiveAlertsFilter = (props) => {
+  const { baseLink } = props;
+  const history = useHistory();
+  const query = useQuery();
+  const selectedFilter = query.get('severity');
+
+  let items = [
+    {
+      label: 'All',
+      value: 'all',
+      onClick: () => {
+        query.delete('severity');
+        history.push(`${baseLink}?${query.toString()}`);
+      },
+      iconCode: '',
+    },
+    {
+      label: 'Critical',
+      value: STATUS_CRITICAL,
+      onClick: () => {
+        query.set('severity', STATUS_CRITICAL);
+        history.push(`${baseLink}?${query.toString()}`);
+      },
+      iconCode: 'fas fa-times-circle',
+    },
+    {
+      label: 'Warning',
+      value: STATUS_WARNING,
+      onClick: () => {
+        query.set('severity', STATUS_WARNING);
+        history.push(`${baseLink}?${query.toString()}`);
+      },
+      iconCode: 'fas fa-exclamation-triangle',
+    },
+  ];
+
+  const dropDownLabel = (() => {
+    if (!selectedFilter) return items[0].label;
+
+    const item = items.find((item) => item.value === selectedFilter);
+    return (
+      <span>
+        {item.value !== 'all' && (
+          <FilterIcon status={item.value} className={item.iconCode} />
+        )}
+        {item.label}
+      </span>
+    );
+  })();
+
+  if (selectedFilter) {
+    items = items.filter((item) => item.value !== selectedFilter);
+  } else {
+    items.splice(0, 1);
+  }
+
+  return <Dropdown items={items} text={dropDownLabel} size="small" />;
+};
+
+export default ActiveAlertsFilter;
--- a/ui/src/components/ActiveAlertsFilters.js
+++ b/ui/src/components/ActiveAlertsFilters.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import { Dropdown } from '@scality/core-ui';
-import { useHistory } from 'react-router';
+import { useHistory, useRouteMatch } from 'react-router';
 import { useQuery } from '../services/utils';
 import { padding } from '@scality/core-ui/dist/style/theme';
 import { STATUS_WARNING, STATUS_CRITICAL } from '../constants.js';
@@ -33,9 +33,9 @@ export const ActiveAlertsFilterWrapper = styled.div`
 `;
 
 const ActiveAlertsFilter = (props) => {
-  const { baseLink } = props;
   const history = useHistory();
   const query = useQuery();
+  const match = useRouteMatch();
   const selectedFilter = query.get('severity');
 
   let items = [
@@ -44,7 +44,7 @@ const ActiveAlertsFilter = (props) => {
       value: 'all',
       onClick: () => {
         query.delete('severity');
-        history.push(`${baseLink}?${query.toString()}`);
+        history.push(`${match.url}?${query.toString()}`);
       },
       iconCode: '',
     },
@@ -53,7 +53,7 @@ const ActiveAlertsFilter = (props) => {
       value: STATUS_CRITICAL,
       onClick: () => {
         query.set('severity', STATUS_CRITICAL);
-        history.push(`${baseLink}?${query.toString()}`);
+        history.push(`${match.url}?${query.toString()}`);
       },
       iconCode: 'fas fa-times-circle',
     },
@@ -62,7 +62,7 @@ const ActiveAlertsFilter = (props) => {
       value: STATUS_WARNING,
       onClick: () => {
         query.set('severity', STATUS_WARNING);
-        history.push(`${baseLink}?${query.toString()}`);
+        history.push(`${match.url}?${query.toString()}`);
       },
       iconCode: 'fas fa-exclamation-triangle',
     },
--- a/ui/src/components/ActiveAlertsFilters.js
+++ b/ui/src/components/ActiveAlertsFilters.js
@@ -26,6 +26,12 @@ export const FilterIcon = styled.i`
   padding: ${padding.smaller};
 `;
 
+export const ActiveAlertsFilterWrapper = styled.div`
+  .sc-dropdown > div {
+    background-color: ${(props) => props.theme.brand.info};
+  }
+`;
+
 const ActiveAlertsFilter = (props) => {
   const { baseLink } = props;
   const history = useHistory();
@@ -82,7 +88,11 @@ const ActiveAlertsFilter = (props) => {
     items.splice(0, 1);
   }
 
-  return <Dropdown items={items} text={dropDownLabel} size="small" />;
+  return (
+    <ActiveAlertsFilterWrapper>
+      <Dropdown items={items} text={dropDownLabel} size="small" />
+    </ActiveAlertsFilterWrapper>
+  );
 };
 
 export default ActiveAlertsFilter;
--- a/ui/src/components/CommonLayoutStyle.js
+++ b/ui/src/components/CommonLayoutStyle.js
@@ -64,3 +64,12 @@ export const TabTitle = styled.div`
   font-weight: ${fontWeight.bold};
   padding: 35px 0 ${padding.small} ${padding.large};
 `;
+
+export const TextBadge = styled.span`
+  background-color: ${(props) => props.theme.brand.base};
+  color: ${(props) => props.theme.brand.textPrimary};
+  padding: 2px ${padding.small};
+  border-radius: 4px;
+  font-size: ${fontSize.small}
+  font-weight: ${fontWeight.bold}
+`;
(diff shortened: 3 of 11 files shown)


ui/src/components/ActiveAlertsCounter.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useHistory, useLocation } from 'react-router';
+import {
+  padding,
+  fontSize
+} from '@scality/core-ui/dist/style/theme';
+
+import {
+    STATUS_WARNING,
```
ChengYanJin: Can we please make sure we have the same prettier setting? So that our codes have the same format. 

```javascript
{
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "all"
}
```

ui/src/components/VolumeDetailCard.js
```diff
@@ -195,6 +211,8 @@ const VolumeDetailCard = (props) => {
     <VolumeDetailCardContainer>
       <VolumeInformation>
         <VolumeNameTitle data-cy="volume_detail_card_name">
+          <CircleStatus className="fas fa-circle" status={health} />
```
ChengYanJin, line 230:
```suggestion
          <CircleStatus className="fas fa-circle fa-2x" status={health} />
```
According to the design, please use the bigger size of the icon.

ui/src/components/VolumeDetailCard.js
```diff
@@ -195,6 +211,8 @@ const VolumeDetailCard = (props) => {
...
+          &nbsp;
           {name}
```
ChengYanJin, line 231: the font size of the `Volume Name Title` should be ${fontSize.larger} according to the design

ui/src/components/ActiveAlertsFilters.js
```diff
@@ -0,0 +1,92 @@
...
+        const item = items.find(item => item.value === selectedFilter);
+        return <span>
+            {item.value !== 'all' && <FilterIcon status={item.value} className={item.iconCode} />}
+            {item.label}
+        </span>;
+    }) ();
+
+    return (
+        <Dropdown
+        items={items}
```
ChengYanJin: The items here should be all the available timespan **exclude** the current selection. So we need to filter the metricsTimeSpanItems and pass them to items.

ui/src/components/VolumeActiveAlertsCard.js
```diff
@@ -166,8 +182,17 @@ const ActiveAlertsCard = (props) => {
 
   return (
     <ActiveAlertsCardContainer>
-      <ActiveAlertsTitle>{intl.translate('active_alerts')}</ActiveAlertsTitle>
-      {PVCName && alertlist.length !== 0 ? (
+      <ActiveAlertsTitle>
+        <div>
+          {intl.translate('active_alerts')}
+          &nbsp;&nbsp;
+          <TextBadge>{alertlist?.length}</TextBadge>
```
ChengYanJin: After discussing with our UX, here we could remove the `<TextBadge>` since we have the same thing in the Alerts tab title. No need to duplicate. 

Same for the TextBadge in the Alert tab.

Thanks.

ui/src/components/ActiveAlertsFilters.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Dropdown } from '@scality/core-ui';
+
```
ChengYanJin, line 8: Please remove the empty line here, same for the rest. Thanks!

ui/src/components/ActiveAlertsFilters.js
```diff
@@ -0,0 +1,92 @@
...
+        return color;
+    }};
+    padding: ${padding.smaller}
+`;
+
+const ActiveAlertsFilter = (props) => {
+    const { baseLink } = props;
+    const history = useHistory();
+    const location = useLocation();
+    const query = new URLSearchParams(location.search);
```
ChengYanJin: FYI, we have `useQuery` hook. Would you think it's more efficient to `useLocation()` directly than `useHistory().location`?

```javascript
// A custom hook that builds on useLocation to parse the query string.
export const useQuery = () => {
  return new URLSearchParams(useLocation().search);
};
```

ui/src/components/VolumeDetailCard.js
```diff
@@ -117,6 +124,12 @@ const LabelValue = styled.span`
   color: ${(props) => props.theme.brand.textPrimary};
 `;
 
+const AlertsCounterContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+
 const VolumeDetailCard = (props) => {
```
ChengYanJin, line 146: Not directly related to this change, but can we prevent the text of  `Used by` and `Labels` from wrapping? since we have enough space after the Volume Usage Progress Bar.

ui/src/components/VolumeActiveAlertsCard.js
```diff
@@ -67,16 +76,22 @@ const NoActiveAlerts = styled.div`
 `;
 
 const ActiveAlertsCard = (props) => {
-  const { alertlist, PVCName } = props;
+  const { alertlist, PVCName, name } = props;
+  const match = useRouteMatch();
```
ChengYanJin: In order to get the selected volumes, we can rely on `react-router's` Route matching. So no need to pass `name` to component `<ActiveAlertsCard>`

We forgot to declare the parameter to VolumePage In Route, in this way would be easier to retrieve the name

Layout.js    
```javascript
<PrivateRoute path="/volumes/:name" component={VolumePage} />
```

ui/src/components/VolumeActiveAlertsCard.js
```diff
@@ -166,8 +182,17 @@ const ActiveAlertsCard = (props) => {
 
   return (
     <ActiveAlertsCardContainer>
-      <ActiveAlertsTitle>{intl.translate('active_alerts')}</ActiveAlertsTitle>
-      {PVCName && alertlist.length !== 0 ? (
+      <ActiveAlertsTitle>
+        <div>
+          {intl.translate('active_alerts')}
+          &nbsp;&nbsp;
```
ChengYanJin: Would be better to use the padding to be more accurate.

ui/src/components/VolumeListTable.js
```diff
@@ -400,12 +401,13 @@ const VolumeListTable = (props) => {
...
+    const parsedUrl = location.pathname.split('/');
 
     // there are two possiable URLs
     if (isAddNodeFilter) {
       history.push(`/volumes/${row.values.name}?node=${nodeName}`);
     } else {
-      history.push(`/volumes/${row.values.name}`);
+      history.push(`/volumes/${row.values.name}${parsedUrl.length === 4 ? `/${parsedUrl[3]}` : ''}${query.toString() && `?${query.toString()}`}`);
     }
   };
```
ChengYanJin: I would suggest not to parse the URL.

```javascript
    if (isTabSelected) {
      const newPath = location.pathname.replace(
        /\/volumes\/[^/]*\//,
        `/volumes/${volumeName}/`,
      );
      history.push({
        pathname: newPath,
        search: query.toString(),
      });
    } else {
      history.push({
        pathname: `/volumes/${volumeName}/overview`,
        search: query.toString(),
      });
    }
```

ui/src/containers/VolumePageContent.js
```diff
@@ -1,40 +1,72 @@
...
 
   const history = useHistory();
+  const location = useLocation();
+  const match = useRouteMatch();
+  const query = new URLSearchParams(location.search);
+
+  const theme = useSelector((state) => state.config.theme);
+
   const currentVolumeName =
     history?.location?.pathname?.split('/').slice(2)[0] || '';
```
ChengYanJin: Could retrieve the currentVolumeName by `react-router's` Route Matching.

ui/src/containers/VolumePageContent.js
```diff
@@ -131,39 +182,53 @@ const VolumePageContent = (props) => {
...
-              volumeListData?.find((vol) => vol.name === currentVolumeName)
-                .storageCapacity
-            }
-            health={
-              volumeListData?.find((vol) => vol.name === currentVolumeName)
-                .health
+          <Tabs activeColor={theme.brand.primary} items={tabsItems}>
+            {
+              isOverviewPage &&
+              <VolumeDetailCard
```
ChengYanJin: We may want to rename this component to `VolumeOverviewTab`

ui/src/containers/VolumePageContent.js
```diff
@@ -68,59 +100,78 @@ const VolumePageContent = (props) => {
...
+  //   const nodeName = volume?.spec?.nodeName;
+  //   const currentNode = nodes.find((node) => node.name === nodeName);
+  //   instance = `${currentNode?.internalIP}:${PORT_NODE_EXPORTER}`;
+  // } else {
+  //   instance = `${node?.internalIP}:${PORT_NODE_EXPORTER}`;
+  // }
+
+  // const queryStartingTime = volumeStats?.queryStartingTime;
+  // const volumeUsage = volumeStats?.volumeUsage?.find(
+  //   (vU) => vU.metric.persistentvolumeclaim === PVCName,
```
ChengYanJin: Please keep those codes from comments.

ui/src/containers/VolumePageContent.js
```diff
@@ -131,39 +182,53 @@ const VolumePageContent = (props) => {
...
+            {
+              isAlertsPage &&
+              <ActiveAlertsCard
+                name={currentVolumeName}
+                alertlist={alertlist}
+                PVCName={PVCName}
+              />
+              }
+          </Tabs>
+          {/* <ActiveAlertsCard
```
ChengYanJin: Please remove the code it's not used anymore.

ui/src/components/VolumeDetailCard.js
```diff
@@ -281,9 +299,23 @@ const VolumeDetailCard = (props) => {
...
+            <ActiveAlertsCounter
+              alerts={alertlist}
+              baseLink={`${match.url}/${name}/alerts`}
+            />
+          </AlertsCounterContainer>
+        }
         {condition === VOLUME_CONDITION_LINK && (
           <VolumeUsage>
-            <VolumeUsageTitle>Volume Usage</VolumeUsageTitle>
+            <VolumeSectionTitle>{intl.translate('volume_usage')}</VolumeSectionTitle>
```
ChengYanJin: The title should be just `Usage` according to the design.
And thanks for the translations.

ui/src/containers/VolumePageContent.js
```diff
@@ -131,39 +182,53 @@ const VolumePageContent = (props) => {
...
+                alertlist={alertlist}
+              />
             }
-            condition={currentVolume.status}
-            // the delete button inside the volume detail card should know that which volume is the first one
-            volumeListData={volumeListData}
-            pVList={pVList}
-          ></VolumeDetailCard>
-          <ActiveAlertsCard
+            {
```
ChengYanJin: We could have Switch inside the `<Tab>`, would you think it's easier to read?

```javascript
<Switch>
  <Route
    path={`/volumes/${volumeName}/overview`}
    component={VolumeOverviewTab}
   />
  ...
</Switch>
```

ChengYanJin (changes requested):
Thanks for the work! 
Here are a few things I would like to mention before the review: 

- Please make sure we have the same prettier setting.
- Please splits into commits with proper message. [Commit Best Practices](https://github.com/scality/metalk8s/blob/development/2.6/docs/developer/development/commit.rst). It will be easier for us to review the code.

ui/src/components/ActiveAlertsCounter.js
gdemonet: basically use the file at `ui/.prettierrc`

ui/src/components/ActiveAlertsFilters.js
alexis-ld: No, you are absolutely right, replacing with the hook in this component, thanks :) .

In ActiveAlertsCounter component though, I should likely keep the location item. I need independent URLSearchParams for link generation and thus I am creating those in functions. Rules of hooks prevent the use of useQuery() in a function :).
But definitely in ActiveAlertsFilers I am replacing by the hook. :)

ui/src/components/VolumeDetailCard.js
alexis-ld: Not sure... the values are still limited by the column splitting the panel in half. If we want to do that we should refactor the layout so the left column is spit in two part (one that is at Usage progress bar level, and one that is below), that can be tricky.
Do you want me to try that ? The wrapping is not so bad on wide screens.

ui/src/components/VolumeActiveAlertsCard.js
alexis-ld: Indeed :) .

Ended up with : 

```<PrivateRoute path="/volumes/:name?" component={VolumePage} />``` 

Without the ? the VolumePage component would not render if no name was specified in the URL.

File: ui/src/components/VolumeListTable.js (hunk 11 above)
Comment (alexis-ld) [reply ui/src/components/VolumeListTable.js]: Much cleaner indeed :) thanks

File: ui/src/containers/VolumePageContent.js (hunk 13 above)
Comment (alexis-ld) [reply ui/src/containers/VolumePageContent.js]: Also renamed VolumeAlertsTab. :)

File: ui/src/containers/VolumePageContent.js (hunk 14 above)
Comment (alexis-ld) [reply ui/src/containers/VolumePageContent.js]: The reason I commented this is because all this code will be re-enabled when I will add the metrics tab. I wanted to prevent this from being reviewed as "new code" since it was already there and it is a big chunk of code.

Should I delete it and re-add it later then ?

File: ui/src/containers/VolumePageContent.js (hunk 17 above)
Comment (alexis-ld) [reply ui/src/containers/VolumePageContent.js]: Honestly, I prefer the 
`isOverviewPage &&`
approach.

We have to define the isXXXXPage variables anyway to tell the Tabs component which one is selected so I think we should as well use these variables to decide what to render since these are already defined.
It avoid to create more wrappers with the Route Switch and also IMO is easier to read.

But I totally understand if you prefer the Switch and want me to follow the existing coding-style. Let me know if I should change ? :)

File: ui/src/containers/VolumePageContent.js (hunk 14 above)
Comment (ChengYanJin) [reply ui/src/containers/VolumePageContent.js]: IMO it makes sense to keep the code, even though we don't directly use them in this PR. 
It brings a lot of noise by commenting on it or removing it.

File: ui/src/components/VolumeDetailCard.js (hunk 8 above)
Comment (ChengYanJin) [reply ui/src/components/VolumeDetailCard.js:146]: From UX point of view, we would avoid to have "Overflow content" if there is enough space.
So, yeah, please refactor the layout.

File: ui/src/containers/VolumePageContent.js (hunk 17 above)
Comment (ChengYanJin) [reply ui/src/containers/VolumePageContent.js]: Yes. So let's follow the current coding-style ;)

File: ui/src/containers/VolumePageContent.js (hunk 14 above)
Comment (alexis-ld) [reply ui/src/containers/VolumePageContent.js]: I agree :) . The issue was that eslint was making the whole build fail for unused variables if I left the code in place without commenting it.

Comment (ChengYanJin) [reply ui/src/containers/VolumePageContent.js]: Indeed. we'll have ESLint error for the unused vars.
I would suggest you to locally disable the Rules with Inline Comments. And add a little comment explaining it's a temporary disable until refactoring is done ;)

```javascript
  // Temporarily disable the `no-unused-vars` lint until the refactoring is complete.
  // eslint-disable-next-line no-unused-vars
  const volumeMetricGraphData = {
    volumeUsage,
    volumeThroughputWrite,
    volumeThroughputRead,
    volumeLatencyWrite,
    volumeLatencyRead,
    volumeIOPSRead,
    volumeIOPSWrite,
    queryStartingTime,
  };
```

ui/src/components/ActiveAlertsCounter.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    return `${baseLink}?${query.toString()}`;
+  };
+
+  const criticalCounter = alerts?.filter(
+    (item) => item?.labels?.severity === STATUS_CRITICAL,
+  ).length;
+  const warningCounter = alerts?.filter(
+    (item) => item?.labels?.severity === STATUS_WARNING,
+  ).length;
```
ChengYanJin: Hi @alexis-ld,
Thanks for extracting those 2 components so we can share and reuse them. 

If I were you, I would have passed the number of `criticalCounter` and `warningCounter` to `ActiveAlertsCounter` instead of `alerts`.  Since what we really need is only the number ;)

ui/src/components/ActiveAlertsFilters.js
```diff
@@ -0,0 +1,98 @@
...
+`;
+
+export const ActiveAlertsFilterWrapper = styled.div`
+  .sc-dropdown > div {
+    background-color: ${(props) => props.theme.brand.info};
+  }
+`;
+
+const ActiveAlertsFilter = (props) => {
+  const { baseLink } = props;
```
ChengYanJin: I think it makes sense to remove the `baseLink` props since the URL should only change the query string in this component.

ui/src/components/CommonLayoutStyle.js
```diff
@@ -64,3 +64,12 @@ export const TabTitle = styled.div`
...
 `;
+
+export const TextBadge = styled.span`
+  background-color: ${(props) => props.theme.brand.info};
+  color: ${(props) => props.theme.brand.textPrimary};
+  padding: 2px ${padding.small};
+  border-radius: 4px;
+  font-size: ${fontSize.small}
+  font-weight: ${fontWeight.bold}
+`;
```
ChengYanJin, line 69: I would prefer to add  `margin-left: ${padding.smaller};` for TextBadge instead of `&nbsp;&nbsp;` in the tab title.

ui/src/containers/VolumePageContent.js
```diff
@@ -131,49 +184,57 @@ const VolumePageContent = (props) => {
...
-            PVCName={PVCName}
-          ></ActiveAlertsCard>
-          <MetricGraphCard
-            volumeName={currentVolumeName}
-            volumeMetricGraphData={volumeMetricGraphData}
-            // the volume condition compute base on the `status` and `bound/unbound`
-            volumeCondition={currentVolume.status}
-            // Hardcode the port number for prometheus metrics
-          ></MetricGraphCard>
+          <Tabs activeColor={theme.brand.primary} items={tabsItems}>
```
ChengYanJin, line 205: For the `<Tabs>` we get the feedback that the contrast of the unselected tab is not good, they are not very visible.

Please update with the new design: 
https://www.sketch.com/s/f7ae733f-222f-4335-b742-a86d25dda872/a/n1OpOz

Note that the Sketch is node page, but we want to apply to volume page as well.
The height of the tab title should be 40px according to the design.

Thanks!

ui/src/components/ActiveAlertsCounter.js
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+export const CounterIcon = styled.i`
+  color: ${(props) => {
+    const theme = props.theme.brand;
+    let color = theme.textPrimary;
+
+    switch (props.status) {
+      case STATUS_WARNING:
+        color = theme.warning;
```
ChengYanJin:
```suggestion
        color = theme.alert;
```

ui/src/containers/VolumePageContent.js
```diff
@@ -119,8 +154,26 @@ const VolumePageContent = (props) => {
     queryStartingTime,
   };
 
+  const isAlertsPage = location.pathname.endsWith('/alerts');
+  const isOverviewPage = location.pathname.endsWith('/overview') || !isAlertsPage;
```
ChengYanJin:
```suggestion
  const isOverviewPage = location.pathname.endsWith('/overview');
```

ui/src/components/VolumeOverviewTab.js
```diff
@@ -195,6 +210,8 @@ const VolumeDetailCard = (props) => {
     <VolumeDetailCardContainer>
       <VolumeInformation>
         <VolumeNameTitle data-cy="volume_detail_card_name">
+          <CircleStatus className="fas fa-circle fa-2x" status={health} />
```
ChengYanJin: Ha. If you use the component `<CircleStatus/>` the double size may not work.

You may want to cherry-pick [this commit](https://github.com/scality/metalk8s/pull/2868/commits/778b284d107e87614d1568af1ebd3b6225793f1a) (once it's merged) and properly use this component.

ui/src/components/VolumeOverviewTab.js
```diff
@@ -195,6 +210,8 @@ const VolumeDetailCard = (props) => {
...
+          &nbsp;
```
ChengYanJin: We may want to use padding to add space.

ui/src/components/VolumeOverviewTab.js
```diff
@@ -195,6 +210,8 @@ const VolumeDetailCard = (props) => {
...
           {name}
         </VolumeNameTitle>
         <InformationSpan>
```
ChengYanJin, line 233: You may want to reflector the layout to avoid the label from wrapping as we discussed last time?

ChengYanJin (changes requested):
Hello @alexis-ld ,

Thanks for the work!
Some comments about the overview and alerts tabs. I'll give my review of metrics and details review in another PR.

ui/src/components/ActiveAlertsCounter.js
```diff
@@ -0,0 +1,100 @@
...
+  const history = useHistory();
+  const location = useLocation();
+
+  const getLink = (status) => {
+    const query = new URLSearchParams(location.search);
+    const existing = query.getAll('severity');
+    if (existing.indexOf(status) === -1) {
+      query.set('severity', status);
+    }
+    return `${baseLink}?${query.toString()}`;
```
ChengYanJin: IMO, we could remove the baseLink in component `<ActiveAlertCounter/>` as well
What we need is too replace the `/overview` by `/alerts` what don't really need to know the pathname before.
```javascript
url = url.replace(/\/overview/, '/alerts');
```

```suggestion
    return `${url}?${query.toString()}`;
```

ui/src/components/VolumeOverviewTab.js
alexis-ld: I tried with your commit but it seems the CIRCLE_SIZE_... constants are not included in the commit. I removed the 2x for now it seems it is still displaying properly :) . Will update when your commit will be merged.

ui/src/components/VolumeOverviewTab.js
alexis-ld: I did that update in #2866 to avoid conflicting with the scrollable tab commit I did on the other PR.

alexis-ld:
@ChengYanJin I made all the changes, rebased from development 2.6 and fixed the conflicts :) .
Will fix the integration conflicts once we will have merged the missing pieces from #2866 to fix all conflicts at once.

ChengYanJin:
merged in PR #2866.

PR #2839, opened 2020-10-09
Migration to Python3

### LINKED ISSUE #2203: Migration to Python3
TBD

**Component**:

**Why this is needed**:

**What should be done**:

- Make salt custom modules, states, runners, pillars, ... works with python3
- Install Salt python3
- Update Salt unit test to work in python3

**Test plan**:

TeddyAndrieux:
**Component**:

'salt', 'build'

**Context**: 

#2203 

**Summary**:

- Install salt-master and salt-minion Python3 versions
- Make salt unit tests working in Python3
- Bump salt to `3002.2`
- Add an "ugly" orchestrate to migrate Salt minion
- Remove non-breaking space characters from Solutions manifests

**Acceptance criteria**: 

- [x] Working boostrap
- [x] Working expansions
- [x] Working upgrade
- [x] Working downgrade
- [x] Working solution install
- [x] Working bootstrap restore
- [x] Working unit tests in Python3
---

Fixes: #2203


tests/install/steps/test_expansion.py
```diff
@@ -37,16 +37,12 @@ def declare_node(
 
 @when(parsers.parse('we deploy the node "{node_name}"'))
 def deploy_node(host, ssh_config, version, node_name):
-    accept_ssh_key = [
```
slaperche-scality, line 40: Seems unrelated to this PR.

And if this is really useless, then the doc needs an update as well: https://metal-k8s.readthedocs.io/en/development-2.6/installation/expansion.html#id3

So maybe this should be done in a dedicated PR (and keep this one only about py3 migration)

salt/_modules/metalk8s.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import os.path
 import re
+import six
```
slaperche-scality: If we're moving to Python 3, why using the `six` compat layer?

Like:
> - replace `basestring` with `six.string_types`

Why not using `str` directly?

From https://docs.python.org/3.0/whatsnew/3.0.html:

> The builtin basestring abstract type was removed. Use str instead. The str and bytes types don’t have functionality enough in common to warrant a shared base class. The 2to3 tool (see below) replaces every occurrence of basestring with str.


---

Otherwise we're gonna have two migration PR?
2only -> 2-3 compat and then 2-3compat -> 3 only?

TeddyAndrieux: It's more because for the moment we cannot rewrite the salt module unit test to work in python3, and we will do it in the future.
Also, I didn't tested yet but for upgrade purpose, we will maybe need to have a module compatible with both py2 and py3

tests/install/steps/test_expansion.py
TeddyAndrieux: Yes you right, for the doc I think it's still good to have it as a pre-check for expansion.

Edit: Oh it will not work anyway in the doc since we need python3 installed to have a working salt-ssh, ok I will update the doc

TeddyAndrieux: I updated the documentation but I think I will keep it in this PR since it's changes needed for Python3 migration even if it's not directly related

docs/installation/expansion.rst
```diff
@@ -237,11 +237,25 @@ Open a terminal in the Salt Master container using
 #. Check that SSH access from the Salt Master to the new node is properly
    configured (see :ref:`Bootstrap SSH Provisioning`).
 
+   .. note::
+
+     You need to use `salt-ssh` as raw ssh since `salt-ssh` need Python3
+     installed to work properly (that will be installed by the following
+     salt orchestrate).
```
gdemonet: Not sure this note is relevant to our readers, but if it is, we should reformulate it a bit:

```
Salt SSH requires Python 3 to be installed on the remote host to run Salt functions.
It will be installed automatically when deploying the node, though you can send raw
shell commands before (using ``--raw-shell``) if needed.
```

(side-note: literal text needs double backticks **``** in RST (or the `:literal:` role), otherwise it's interpreted differently and we cannot guarantee how it will be displayed - see [the docs](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#interpreted-text))

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -6,13 +6,24 @@
...
+    - roster: kubernetes
+
 Deploy salt-minion on a new node:
   salt.state:
     - ssh: true
     - roster: kubernetes
     - tgt: {{ node_name }}
     - saltenv: metalk8s-{{ version }}
     - sls:
+      - metalk8s.salt.minion.dependencies
```
gdemonet: Shouldn't this be `include`d in `roles.minion` instead? I get that we need to run it first to avoid failing the installation of other packages, but can't we handle this ordering with `include`s?

tests/install/steps/test_expansion.py
gdemonet: Maybe you could replace this command with the one suggested in docs (using `--raw-shell`)? I think we added it here to make sure the test would fail early if there was any issue with the key, and also because it's good practice to do in tests exactly what we have in docs.

docs/installation/expansion.rst
TeddyAndrieux: Right, 
I think it's relevant because they cannot run directly `test.ping` as it was documented before, I will change it as you suggest

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Nop we can't because if we start the salt process without this `python36-rpm` installed once it come to salt-minion installation it will not compare the package version as expected (even if python36-rpm get installed during this state) because salt internal module get loaded only at the beginning of the state (I think).

Edit: Hmm ok I just did a try and it seems to work when I include it directly in the roles.minion :thinking: don't really know why my last try failed, I will change it

tests/install/steps/test_expansion.py
TeddyAndrieux: For me this one was here just because it was needed at the beginning to accept the ssh public key of the new node.
Anyway right I can add this check here like in the doc

salt/metalk8s/salt/minion/dependencies.sls
```diff
@@ -0,0 +1,17 @@
+# Salt need `python36-rpm` to do proper version comparaison and since we
+# use salt-ssh to install salt the first time, we want this rpm to be installed
+# before Salt-minion as salt-minion installation state will report a "failure"
+# because of version comparaison
```
gdemonet: minor: `s/comparaison/comparison/`

salt/metalk8s/orchestrate/migrate_salt.sls
```diff
@@ -0,0 +1,24 @@
+# This state is used for Salt migration:
+#  - for upgrade from 2.6.x to 2.7.x (migration from Python2 to Python3)
+#  - for downgrade from 2.7.x to 2.6.x (migration from Python3 to Python2)
+#
+# Salt migration cannot be done automatically by a salt state because of the
+# changement between Python2 and Python3
```
gdemonet: minor: `s/changement/change/`

gdemonet (dismissed):
Only really minor comments, otherwise LGTM! Will manually test as well

gdemonet (dismissed):

gdemonet (dismissed):

gdemonet (approved):

PR #2844, opened 2020-10-12
Expose Solution Operator config loading as a standalone Go module

### LINKED ISSUE #2837: Make Operator config loading a proper Go module for Solutions to import
**Component**: solutions

**Why this is needed**: To ease Solution Operators efforts to keep things in sync with our changes (in case this config file format evolves over time)

**What should be done**: Extract the loading code from `examples/metalk8s-solution-example/operator/pkg/config` into `go/operator-config` and make it a proper Go module.

**Test plan**: Import it from the example Operator.

gdemonet:
**Component**: go, solutions

**Context**:

Solution Operators are expected to read a config file on startup which lists the available image repositories for the Solution components. This config file format belongs to the Solution framework, and as such, it makes sense for said framework to expose an _importable_ package for the Operators to use.

**Summary**:

- Define a new Go module `github.com/scality/metalk8s/go/solution-operator`, including (for now) a single `config` package. Two methods are exposed for convenience:
	- `LoadConfiguration` which takes a `Reader` interface and returns a parsed and validated `OperatorConfig`,
	- and `LoadConfigurationFromFile` which takes a file path and passes its contents to `LoadConfiguration`
- Use this package in the `examples/metalk8s-solution-example/operator` example Solution Operator

**Acceptance criteria**: 

- The example Operator still works
- One can `go get` this new module and use the `config` package as expected

---

Closes: #2837


go/solution-operator/config/config_test.go
```diff
@@ -0,0 +1,174 @@
...
+						},
+					},
+				},
+			},
+			error: "",
+		},
+		"invalid-no-apiversion": {
+			filename: "invalid-no-apiversion.yaml",
+			result:   nil,
+			error:    "missing required field: apiVersion",
```
miniscruff: Still new to go but the way we have been testing errors in zenko-operator has been nice. Instead of using the string of the error, we have been creating error values and using github.com/pkg/errors wrap method to add extra values. Then errors.cause to check the root error. Avoids having to copy paste strings and allows you to change error messages without having to also change tests just for a string value.

miniscruff (dismissed):
Looking good to me, I might have an extra comment or two once I get to use it but as far as I can tell the API is simple and straightforward.

NicolasT:
Since `solution-operator` is not really an operator, did you consider `go/solution-operator-lib` instead as the package root? Could then also use `go/solution-operator-lib/pkg/config` as the package, so if there's a need for `go/solution-operator-lib/internal/...` at some point in time this can be trivially added.

go/solution-operator/config/config_test.go
NicolasT: For reference, https://blog.golang.org/go1.13-errors

gdemonet:
> Since `solution-operator` is not really an operator, did you consider `go/solution-operator-lib` instead as the package root? Could then also use `go/solution-operator-lib/pkg/config` as the package, so if there's a need for `go/solution-operator-lib/internal/...` at some point in time this can be trivially added.

Good points, will change these.

go/solution-operator/config/config_test.go
gdemonet: Thanks for the suggestion, I'll have a look!

gdemonet: So I'll go with the "plain" Go 1.13 `Unwraps` as suggested in Nicolas' link, which should achieve the same effect as `github.com/pkg/errors` without requiring any additional package.

examples/metalk8s-solution-example/operator/go.sum
```diff
@@ -42,13 +42,15 @@ github.com/Azure/go-autorest/autorest/date v0.2.0 h1:yW+Zlqf26583pE43KhfnhFcdmSW
 github.com/Azure/go-autorest/autorest/date v0.2.0/go.mod h1:vcORJHLJEh643/Ioh9+vPmf1Ij9AEBM5FuBIXLmIy0g=
 github.com/Azure/go-autorest/autorest/mocks v0.1.0/go.mod h1:OTyCOPRA2IgIlWxVYxBee2F5Gr4kF2zd2J5cFRaIDN0=
 github.com/Azure/go-autorest/autorest/mocks v0.2.0/go.mod h1:OTyCOPRA2IgIlWxVYxBee2F5Gr4kF2zd2J5cFRaIDN0=
+github.com/Azure/go-autorest/autorest/mocks v0.3.0 h1:qJumjCaCudz+OcqE9/XtEPfvtOjOmKaui4EOpFI6zZc=
```
slaperche-scality, line 45: Not sure why we have all those changes to the `go.sum` of `solution-example` (only additions, no removal).

gdemonet: I don't understand how `go.sum` works, TBH... I've removed the code from `pkg/config` and ran `go mod tidy` (I expected this would refresh both `go.mod` and `go.sum` based on the imports within the module), which complains about something entirely unrelated:

```
example-solution-operator imports
        sigs.k8s.io/controller-tools/pkg/crd/generator: module sigs.k8s.io/controller-tools@latest found (v0.4.0), but does not contain package sigs.k8s.io/controller-tools/pkg/crd/generator
```

And that's roughly all I did, so I'm not sure what's expected or what I should do.

gdemonet: OK I split the `go mod tidy` into its own commit, to make things clearer.

go/solution-operator/config/config_test.go
gdemonet: @miniscruff I explained a bit more the choices around error handling in [`9b6d7b5` (#2844)](https://github.com/scality/metalk8s/pull/2844/commits/9b6d7b5e834cc540c6765d5d387f46f9cdec6360), which doesn't really match what is done for Zenko Operator.
In any case, thanks for pointing out that improvements could be made, I learnt a great deal while playing with all this!

miniscruff (dismissed):
Looks good to me, and worked when I tried manually.

miniscruff (dismissed):
Code looks to be unchanged with the conflicts so still good to me, can run another manual test if you think I should.

alexandre-allard (approved):

PR #2850, opened 2020-10-13
salt: Make salt module compatible with Python3

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

We will migrate Salt to python3 in development/2.7 but first, let's make salt modules compatible with python2 and python3

**Summary**:

Make all custom salt modules, states, ... compatible with Python3 using
six:
- replace `basestring` with `six.string_types`
- replace `urlparse` with `six.moves.urllib.parse`
- do not try to decode string object
- do not convert string to unicode
- do not use `yum` Python API as it's only available with Python2
- encode and decode string object when using `base64`

NOTE: We do this change before migrating to Python3, so that future
downgrade will be simpler

---

Refs: #2203

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -11,6 +11,7 @@ import json
 import re
 import operator
 import os
+import six
 import time
 
 import logging
@@ -215,11 +216,9 @@ def device_info(name):
 # Volume {{{
 
 
-class Volume(object):
+@six.add_metaclass(abc.ABCMeta)
+class Volume():
     """Volume interface."""
-
-    # XXX: Will need to be updated when moving to Python 3
-    __metaclass__ = abc.ABCMeta
 
     def __init__(self, volume):
         self._volume = volume
@@ -295,7 +294,7 @@ class Volume(object):
                             .format(self.path))
         storage_class = self.get('spec.storageClass')
         # If we got a string that means the name wasn't replaced by the object.
-        if isinstance(storage_class, basestring):
+        if isinstance(storage_class, six.string_types):
             raise Exception('StorageClass {} not found'.format(storage_class))
         params = storage_class['parameters']
         # mkfs options, if any, are stored as JSON-encoded list.
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -5,6 +5,7 @@ Module for handling MetalK8s specific calls.
 import logging
 import os.path
 import re
+import six
 import socket
 import time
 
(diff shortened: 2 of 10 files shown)


salt/_modules/metalk8s_volumes.py
```diff
@@ -295,7 +296,7 @@ def prepare(self, force=False):
                             .format(self.path))
         storage_class = self.get('spec.storageClass')
         # If we got a string that means the name wasn't replaced by the object.
-        if isinstance(storage_class, basestring):
+        if isinstance(storage_class, six.string_types):
```
slaperche-scality, line 297: What about l.222-223 of this file?

```python
    # XXX: Will need to be updated when moving to Python 3
    __metaclass__ = abc.ABCMeta
```
:stuck_out_tongue: 

You'll probably want https://stackoverflow.com/a/35673504

TeddyAndrieux: Oh ! Right on my py3 test didn't faced any issue about this one. Anyway I will change this one

salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -2,8 +2,7 @@
 
 {#- When downgrading saltenv should be the newest version #}
 {%- set nodes_versions = pillar.metalk8s.nodes.values() | map(attribute='version') | list %}
-{%- do nodes_versions.sort(cmp=salt.pkg.version_cmp, reverse=True) %}
-{%- set expected = nodes_versions | first %}
+{%- set expected = salt.metalk8s.cmp_sorted(nodes_versions, cmp=salt.pkg.version_cmp, reverse=True) | first %}
```
gdemonet: Is this related to Python3 compat?

gdemonet (commented):
Didn't see any issue in the diff, but didn't check the rest of the code... Do you need me to test it with Python 3?

salt/metalk8s/orchestrate/downgrade/precheck.sls
gdemonet: Moreover, I don't have this `cmp_sorted` method available in `salt/_modules/metalk8s.py` :confused:

TeddyAndrieux: Oh shit, you right it's related to Python3 migration but not to Python3 compat, it shouldn't be in this PR. I will remove it

TeddyAndrieux:
> Didn't see any issue in the diff, but didn't check the rest of the code... Do you need me to test it with Python 3?

You can test if you want but I already did some upgrade and downgrade test from Py2 to Py3 Salt and it worked so ... should be ok :smiley:

gdemonet (approved):
> You can test if you want but I already did some upgrade and downgrade test from Py2 to Py3 Salt and it worked so ... should be ok :smiley: 

Then let's go! :shipit:

Issue #1750, opened 2019-09-21
Downgrade between two minor releases is not possible when etcd version change


sayf-eddine-scality:
**Component**:
etcd, downgrade

**What happened**:
Downgrade failed from 2.3 to 2.2 because etcd does not support downgrading the cluster from 3.3 to 3.2 if all the cluster members are on 3.3
Check: https://github.com/etcd-io/etcd/issues/10461

**What was expected**:
Downgrade to be successful

**Steps to reproduce**
Install a metalk8s 2.3 bootstrap node and downgrade to 2.2

**Resolution proposal** (optional):
TBD

gdemonet:
Given [this doc](https://github.com/etcd-io/etcd/blob/master/Documentation/upgrades/upgrade_3_3.md#downgrade), it's **impossible** to downgrade, unless we create some snapshot of data while having the `etcd` cluster in v3.2

How would we handle this? I guess downgrade is necessary only after a breaking upgrade, right? We would need to warn users during the 2.2 -> 2.3 upgrade, wouldn't we?

sayf-eddine-scality:
> Given [this doc](https://github.com/etcd-io/etcd/blob/master/Documentation/upgrades/upgrade_3_3.md#downgrade), it's **impossible** to downgrade, unless we create some snapshot of data while having the `etcd` cluster in v3.2
> 
> How would we handle this? I guess downgrade is necessary only after a breaking upgrade, right? We would need to warn users during the 2.2 -> 2.3 upgrade, wouldn't we?

the operations and the api are obviously backward incompatible, so so the only way is to backup the cluster before going through this path to be able to restore a 2.2 working cluster if things went south

Or (bit complicated), we can have an intermediate state where we only upgrade the etcd and keep the rest on 2.2 then upgrade to 2.3 thus if the upgrade fail we will have a working 3.3 etcd on a 2.2 cluster, not sure if this is totally supported thought

sayf-eddine-scality:
keeping this open due the possibility of having future versions of `etcd` non backward compatible.

gdemonet:
What should be the course of action? We disabled downgrade tests for now, but I guess we'd still want to support it in the long run, no?

sayf-eddine-scality:
yes I deactivated the downgrade test on the 2.3 branch but we need to find a definitive fix for this

PR #2858, opened 2020-10-14
Prevent downgrade when not possible

### LINKED ISSUE #2852: Prevent user from downgrading when it's not possible (e.g. etcd version change)
**Component**: salt

**Why this is needed**:
To prevent a user from launching a downgrade when it would break the cluster.

**What should be done**:
Add some pre-check in downgrade orchestration to prevent downgrade on some versions (e.g 2.6 to 2.5).

Enrich the currently existing `salt/metalk8s/orchestrate/downgrade/precheck.sls` with a check ensuring that downgrade is supported from the current version to the destination one.

**Test plan**:

alexandre-allard:
**Component**: salt, scripts, downgrade

**Context**: See #2852 :)

**Summary**:

**Acceptance criteria**: 

---

Closes: #2852

### DIFF
--- a/salt/metalk8s/defaults.yaml
+++ b/salt/metalk8s/defaults.yaml
@@ -1,6 +1,13 @@
 ---
 metalk8s:
   debug: false
+  downgrade:
+    # Define if the downgrade from this minor version is supported.
+    # It should be set to false if manual operations are needed
+    # (e.g. downgrade of etcd), prior to downgrading the cluster.
+    # The downgrade can still be forced setting `metalk8s.downgrade.enabled`
+    # to `True` in the pillar.
+    enabled: false
 
 kubernetes:
   cluster: kubernetes
--- a/salt/metalk8s/orchestrate/downgrade/precheck.sls
+++ b/salt/metalk8s/orchestrate/downgrade/precheck.sls
@@ -1,3 +1,5 @@
+{%- from "metalk8s/map.jinja" import metalk8s with context %}
+
 {%- set dest_version = pillar.metalk8s.cluster_version %}
 
 {#- When downgrading saltenv should be the newest version #}
@@ -13,6 +15,22 @@ Invalid saltenv "{{ saltenv }}" consider using "metalk8s-{{ expected }}":
 
 Correct saltenv "{{ saltenv }}" for downgrade to "{{ dest_version }}":
   test.succeed_without_changes
+
+{%- endif %}
+
+{%- set dest_minor_version, _ = (dest_version|string).rsplit('.', 1) %}
+{%- set current_minor_version, _ = (expected|string).rsplit('.', 1) %}
+
+{%- if not metalk8s.downgrade.enabled and dest_minor_version != current_minor_version %}
+
+{% set release_note = 'https://github.com/scality/metalk8s/releases/tag/'
+                       ~ current_minor_version ~ '.0' %}
+
+Cannot downgrade from {{ current_minor_version }} to {{ dest_minor_version }}:
+  test.fail_without_changes:
+    - comment: |-
+        Downgrade is not supported, because of etcd version change
+        (see {{ release_note }} for details).
 
 {%- endif %}
 
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -100,6 +100,8 @@ Sync module on salt-master:
   salt.runner:
     - name: saltutil.sync_all
     - saltenv: metalk8s-{{ dest_version }}
+    - require:
+      - salt: Execute the downgrade prechecks
 
 Deploy Kubernetes service config objects:
   salt.runner:
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -11,6 +11,8 @@ Execute the downgrade prechecks:
     - pillar:
         orchestrate:
           dest_version: {{ dest_version }}
+        metalk8s:
+          downgrade: {{ pillar.metalk8s.get('downgrade', {}) | tojson }}
 
 {%- set cp_nodes = salt.metalk8s.minions_by_role('master') | sort %}
 {%- set other_nodes = pillar.metalk8s.nodes.keys() | difference(cp_nodes) | sort %}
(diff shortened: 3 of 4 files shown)


salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -16,6 +18,13 @@ Correct saltenv "{{ saltenv }}" for downgrade to "{{ dest_version }}":
 
 {%- endif %}
 
+{%- if not defaults.downgrade_supported and not pillar.force_downgrade | default(False) %}
```
alexandre-allard: Maybe we want this pillar entry below the `metalk8s` one ? `pillar.metalk8s.force_downgrade` or even something like `pillar.metalk8s.downgrade.force` ?
Same for defaults, may be `downgrade.supported`, but I'm not sure the `downgrade` dict would be useful for anything else in the future.

salt/metalk8s/defaults.yaml
```diff
@@ -105,4 +105,11 @@ coredns:
 
 upgrade: false        # define if we're on an upgrade case
 
+# Define if the downgrade from this version is supported.
+# It should be set to false if manual operations are needed
+# (e.g. downgrade of etcd), prior to downgrading the cluster.
+# The downgrade can still be forced setting force_downgrade to True
+# in the pillar.
```
gdemonet: Good idea to make this overridable, but why not use the same name in both defaults and pillar, and merge them in `map.jinja`?

salt/metalk8s/orchestrate/downgrade/precheck.sls
gdemonet: I'd put everything under `metalk8s.downgrade`, either using both `.supported` and `.force`, or with a single key, maybe naming it `.enabled`?

salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -16,6 +18,13 @@ Correct saltenv "{{ saltenv }}" for downgrade to "{{ dest_version }}":
...
+
+Downgrade from current version is not supported:
```
gdemonet: Maybe make it obvious like:

```suggestion
Downgrade from current version ({{ expected }}) is not supported:
```

TeddyAndrieux: To me it's more `Downgrade from <current_version> to <dest_version> is not supported`
Maybe you can add the version here instead of "current version" :smile:

salt/metalk8s/orchestrate/downgrade/precheck.sls
TeddyAndrieux: (No strong opinion about the place for `force_downgrade` but maybe since it's only a pillar key that will be passed to orchestrate I would say to be consistent `metalk8s.orchestrate.force_downgrade` (or `metalk8s.orchestrate.downgrade.force` as you want))
Also here the check is wrong I think, because we can downgrade from 2.6.1 to 2.6.0 for example. IMO you should add a check about the destination version here

salt/metalk8s/defaults.yaml
```diff
@@ -105,4 +105,11 @@ coredns:
...
+downgrade_supported: false
```
TeddyAndrieux: Do we really want a pillar key to say downgrade is supported ? Because it's supported from 2.6.1 to 2.6.0 for example but will never be supported from 2.6.x to 2.5.x ... so to me we should just have a check about the version in the precheck

salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -16,6 +18,13 @@ Correct saltenv "{{ saltenv }}" for downgrade to "{{ dest_version }}":
...
+  test.fail_without_changes
```
TeddyAndrieux: To be clear I would add a clear comment here something like "etcd downgrade not supported blablabla" + Maybe a link to a github issue for this https://github.com/scality/metalk8s/issues/1750 (maybe we want to do some cleaning in the ticket (and talk about 2.6 instead of non-released 2.3) but no related to this PR)

salt/metalk8s/defaults.yaml
gdemonet: Good point

salt/metalk8s/orchestrate/downgrade/precheck.sls
gdemonet: I'd rather link to 2.6.0 release notes, where we will explain this limitation, but agreed, additional help would be very useful for users.

salt/metalk8s/orchestrate/downgrade/precheck.sls
```diff
@@ -16,6 +18,16 @@ Correct saltenv "{{ saltenv }}" for downgrade to "{{ dest_version }}":
 
 {%- endif %}
 
+{%- set dest_minor_version, _ = (dest_version|string).rsplit('.', 1) %}
+{%- set current_minor_version, _ = (expected|string).rsplit('.', 1) %}
+
+{%- if not metalk8s.downgrade.enabled and dest_minor_version != current_minor_version %}
+
+Downgrade from {{ current_minor_version }} to {{ dest_minor_version }} is not supported, because of etcd version change (see https://github.com/scality/metalk8s/releases/tag/2.6.0 for details):
+  test.fail_without_changes
```
gdemonet: Maybe use a simpler state ID and explain why in a comment:

```suggestion
{%- set relnotes = 'https://github.com/scality/metalk8s/releases/tag/'
                 ~ current_minor_version ~ '.O' %}
Cannot downgrade from {{ current_minor_version }} to {{ dest_minor_version }}:
  test.fail_without_changes:
    - comment: |-
        Downgrade is not supported, because of etcd version change
        (see {{ relnotes }} for details)
```

(note that we can likely always link to release notes of the first release of the current minor, this will prevent us from forgetting about it in future versions - the reason why this downgrade path is not supported may evolve in the future, but we'll remember if it does... :crossed_fingers:)

gdemonet (changes requested):
LGTM, just a small comment for the message to display in case the downgrade isn't supported.

salt/metalk8s/orchestrate/downgrade/precheck.sls
alexandre-allard: done.

gdemonet (approved):

TeddyAndrieux (approved):
Not tested but LGTM

PR #2868, opened 2020-10-16
[UI] Add alert tab in the new node page

### LINKED ISSUE #2775: [UI New Node Page] Add alerts table in Node Page
**Component**: ui, nodes

**Why this is needed**:
- We want to see all the active alerts related to this node.
- With the filter component to filter the alert base on severity.
 
**What should be done**:
- We should retrieve the alerts from `Alertmanager` and list all the alerts in the table

- When we add the severity filter the URL should update too.
`/newNodes/<node-name>/alerts?severity=<critical/warning>`
- Remove the info selector because we only have critical and warning for the moment for Node
- if there is no severity selected then no alert display. 
- When we come to the alert tab from the tab selection, by default display all the alerts(both critical and warning)
  the URL should be: `/newNodes/<node-name>/alerts?severity=critical,warning` or
  or `/newNodes/<node-name>/alerts?severity=critical&severity=warning`

**Update proposal**
image

Hardcode the alerts related to nodes in frontend.

Store all the selected severity in an array

const selectedSeverity = [ ];

const [isCriticalSelected, setIsCriticalSelected] = useState();
const [isWarningSelected, setIsWarningSelected] = useState();

**Test plan**:
TBD

ChengYanJin:
**Component**: ui, nodes

**Context**:  Add alerts for Node page.

**Summary**:
- Add active alert in Alert Tab
- Reuse the `<ActiveAlertFilter/>` component from @alexis-ld to filter the alerts base on severity.
- Reuse the `<ActiveAlertsCounter/>` component from @alexis-ld to display the alerts in Overview tab and be able to redirect to the Alert tab
- Update the theme.json base on the final dark-theme color-palette.

**Acceptance criteria**: 
image
image

---

Closes: #2775


thomasdanan (dismissed):
excellent !!!

If I had one thing to change, I would suggest improving the alert severity selector. We may want to tackle this improvement later on, in one other ticket.

ui/src/components/ActiveAlertsFilter.js
```diff
@@ -0,0 +1,102 @@
...
+    width: 140px;
+  }
+  .sc-dropdown > div {
+    background-color: ${(props) => props.theme.brand.info};
+    width: 140px;
+  }
+`;
+
+const ActiveAlertsFilter = (props) => {
+  const { baseLink } = props;
```
ChengYanJin, line 40: I think it makes sense to remove the `baseLink` since we change in this component is Query string.

PR #2875, opened 2020-10-19
eve: Create a single terraform worker (with ability to use snapshot)

### LINKED ISSUE #2842: Start using snapshot for release upgrade tests in the CI
**Component**:

'upgrade', 'tests'

**Why this is needed**:

Currently in the CI we get the ISO and do an install of the released version to then perform an upgrade to the version we try to test. 
Several issues:
- some flakies may happen during install of the released version (it's not what we want to test here so ....)
- It takes some time to install this released version on every post-merge
- for one branch we need to support install process of older versions in the CI which is not really good

**What should be done**:

Start using snapshot of working MetalK8s cluster in released versions.

E.g.: If we want to test upgrade from 2.5.2 to 2.6.0-dev
- Create some VM(s) using metalk8s-2.5.2 snapshot
- Wait for the cluster to be fully running and ready
- upgrade to 2.6.0-dev

For a first iteration:
- Create some snapshot manually for every versions we want to test in the CI
- Adapt (or create a new) terraform to spawn machine using those snapshot
- Wait for MetalK8s cluster to be fully ready and running
- Upgrade to <dest_version>
- Launch tests

NOTE: Since we rely on etcd we need to take different snapshot depending on the cluster size so let's follow this naming style:
```
metalk8s-<version>-<small_desc_of_env>-bootstrap
metalk8s-<version>-<small_desc_of_env>-node-x
```

For example, let's consider a 3 node etcd/master/workload 2.5.2 we will have 3 images
```
metalk8s-2.5.2-3etcdmasterworkload-bootstrap
metalk8s-2.5.2-3etcdmasterworkload-node-1
metalk8s-2.5.2-3etcdmasterworkload-node-2
```

TeddyAndrieux:
**Component**:

**Context**: 

#2842 

**Summary**:

Instead of creating several different worker with basically a copy/paste
for each different kind of environment we need using terraform, create a
single one.
Some update:
- Create 2 networks for Control and Workload plane
- Create Ports with static Ips for every machines
- Add a terraform variable to spawn from snapshot (and then restore a
  running MetalK8s cluster, will be used mainly for lifecycle tests)

**Acceptance criteria**:

- [x] Multi-node installation
- [x] RHEL single-node
- [ ] ~Bootstrap restore~ (#2886)
- [ ] ~Upgrade using snapshot~ (#2889)

---

TODO:

- [x] Handle RHEL in this terraform worker
- [x] Remove openstack rhel specific worker
- [x] Handle Cypress in this terraform worker
- [x] Remove openstack cypress specific worker
- [x] Use new terraform worker in CI (update `eve/main.yaml`)
- [x] Create a snapshot for single-node 2.5.1
- [x] Use snapshot in single-node upgrade for released versions

---

Fixes: #2842


tests/conftest.py
```diff
@@ -1,4 +1,5 @@
 import json
+import os.path
```
alexandre-allard: any reason for this import ?

eve/workers/openstack-terraform/terraform/scripts/network.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -xue -o pipefail
+
+IFACE=$1
+
+MAC_ADDR="$(cat "/sys/class/net/$IFACE/address")"
```
alexandre-allard: Useless use of cat & subshell:
```suggestion
read -r MAC_ADDR < "/sys/class/net/$IFACE/address"
```

eve/workers/openstack-terraform/terraform/nodes.tf
```diff
@@ -0,0 +1,314 @@
...
+      "sudo scripts/network.sh eth1",
+      "sudo scripts/network.sh eth2",
+      "sudo systemctl restart network"
+    ]
+  }
+
+  # Register RHSM if OS = rhel
+  provisioner "remote-exec" {
+    inline = [
+      "if [ '${var.os}' = 'rhel-7' ]; then sudo chmod +x scripts/rhsm-register.sh; fi",
```
alexandre-allard, line 257:
```suggestion
      "[ '${var.os}' = 'rhel-7' ] && sudo chmod +x scripts/rhsm-register.sh",
```

TeddyAndrieux: Yes but no because at the end the retcode will be `1` if we are not on rhel-7

tests/conftest.py
TeddyAndrieux: Good catch

eve/workers/openstack-terraform/terraform/scripts/network.sh
TeddyAndrieux: Right :+1:

alexandre-allard (approved):
Tested with a snapshot, it worked well. LGTM.

PR #2882, opened 2020-10-21
[UI New Node Page] Add detail tab in new node page (only for review)

ChengYanJin:
**Component**: ui, nodes

**Context**: Add detail tab in the Node Page

**Summary**:
Display JSON instead of YAML for the moment. Tried with libraries(js-yaml). Did not work as expected.

**Acceptance criteria**: 
Only for code review here, will not merge this PR.


slaperche-scality:
> Tried with libraries(js-yaml). Did not work as expected.

Would be nice to have a bit more info on that (what didn't work)

Issue #2876, opened 2020-10-19
Ability to sort list of Volumes in the Volume page


thomasdanan:
**Component**:

'ui'

**Why this is needed**:

In order to ease browsing and focus when looking at Volumes, it would be good to have the ability to sort Volumes per Name, usage, size, Health, Status and latency. 

**What should be done**:

The Volume Page UI should offer the ability to sort on all columns of the table (exclusive sorting: one column at a time). It should be possible to do ascending, descending sort. Because the nature of each info is different between one column or the other, the sorting algo will be different:

Name: Alphanumeric sorting, case-insensitive
usage: Numeric sorting
size: Numeric sorting
Health: Enum sorting: Critical, Warning, Healthy, Unknown
Status: Enum sorting: Link, Unlink and Exclamation
latency: Numeric sorting

By default, the table is sorted by Health (or usage if all Volumes are with the same Health TBC).

In terms of presentation, the user can click on a column to activate the sorting or to change the sorting order. When doing so, the column header is decorated with a little arrow in the right direction, depending on the sort order.
This sorting setting is part of the URL. If it is not part of the URL, the default sorting is applied.

**Implementation proposal** (strongly recommended):

**Test plan**:

PR #2884, opened 2020-10-22
[UI] Advanced metrics button

### LINKED ISSUE #2870: Volume Advanced Metrics should redirect to the right PVC Dashboard in Grafana
**Component**:

'ui', 'grafana'

**What happened**:

Currently, when clicking on advanced metrics from a Volume Metrics tab in the UI, you get redirected to the Kubernetes/PersitentVolume dashboard but not on the right PVC.

**What was expected**:

Actually, the UI should compute the right URK using the 2 following args: `&var-namespace=<namespace>&var-volume=<pvc>`

To get that info we should find the PVC which binds the PV as well as the namespace in which the PVC is.

Actually, if the Volume is not bound, the advanced Metric tab should be disabled (because no Persistent Volume dashboard exists for this Volume)

**Steps to reproduce**

alexis-ld:
**Component**:
ui, volumes

**Context**: 
The Advanced Metrics button for a volume was not opening the corresponding Dashboard on grafana.

**Summary**:
Forwarded Volume PVC Name and associated namespace to the component to generate the proper link.

**Acceptance criteria**: 
Advanced metrics button opens the dashboard corresponding to the selected volume.

---

Closes: #2870

### DIFF
--- a/ui/src/containers/VolumePageContent.js
+++ b/ui/src/containers/VolumePageContent.js
@@ -155,7 +155,7 @@ const VolumePageContent = (props) => {
   };
 
   const isAlertsPage = location.pathname.endsWith('/alerts');
-  const isOverviewPage = location.pathname.endsWith('/overview') || !isAlertsPage;
+  const isOverviewPage = location.pathname.endsWith('/overview');
   const isMetricsPage = location.pathname.endsWith('/metrics');
   const isDetailsPage = location.pathname.endsWith('/details');
 
--- a/ui/src/containers/VolumePageContent.js
+++ b/ui/src/containers/VolumePageContent.js
@@ -9,6 +9,7 @@ import VolumeListTable from '../components/VolumeListTable';
 import VolumeOverviewTab from '../components/VolumeOverviewTab';
 import VolumeAlertsTab from '../components/VolumeAlertsTab';
 import VolumeMetricsTab from '../components/VolumeMetricsTab';
+import VolumeDetailsTab from '../components/VolumeDetailsTab';
 import {
   SPARSE_LOOP_DEVICE,
   RAW_BLOCK_DEVICE,
@@ -77,6 +78,7 @@ const VolumePageContent = (props) => {
     pods,
     alerts,
     volumeStats,
+    currentVolumeObject,
   } = props;
 
   const history = useHistory();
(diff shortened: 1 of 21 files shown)


thomasdanan (dismissed):
works as expected

ui/src/services/NodeVolumesUtils.js
```diff
@@ -267,6 +267,7 @@ export const getVolumeListData = createSelector(
             ? Math.round(volumeCurrentLatency?.value[1]) + ' µs'
             : undefined,
         errorReason: volume?.status?.conditions[0]?.reason,
+        namespace: volumePVC?.metadata?.namespace,
```
ChengYanJin, line 270: Actually the NodeVolumeUtils should return the volume list only used by the Volume table. 

So I would rather not add a namespace field here, but get the **namespace** from the **PVC** in the `VolumePageContent.js`and pass it to `<VolumeMetricsTab/>`

The reason is that to have an isolated function to retrieve the data for VolumeTable. It will easier for us to do the test.

ui/src/containers/VolumePageContent.js
```diff
@@ -265,6 +265,8 @@ const VolumePageContent = (props) => {
                 render={() => (
                   <VolumeMetricsTab
                   volumeName={currentVolumeName}
+                  volumeNamespace={currentVolume?.namespace}
+                  volumePVCName={PVCName}
                   volumeMetricGraphData={volumeMetricGraphData}
                   // the volume condition compute base on the `status` and `bound/unbound`
                   volumeCondition={currentVolume.status}
```
ChengYanJin, line 272: Please remove the comment in L.273 `// Hardcode the port number for prometheus metrics`

ui/src/services/NodeVolumesUtils.js
```diff
@@ -280,6 +281,6 @@ export const getVolumeListData = createSelector(
       weights[STATUS_HEALTH] = 0;
 
       return weights[b.health] - weights[a.health];
-    })
+    });
```
ChengYanJin: I would also extract the sorting logic outside of this function. The responsibility of this function is only to retrieve the data for VolumeTable. Would that make sense to you?

But you can address it ticket #2876.

ChengYanJin (changes requested):

alexis-ld:
Closed in favor of #2887

PR #2885, opened 2020-10-22
precheck sysctl configuration

### LINKED ISSUE #2853: Add pre-checks to ensure various system configuration are OK
**Component**: salt

**Why this is needed**:
This is needed because sometimes configuration set during installation (e.g. sysctl) can be modified afterwards and we need a way to at least know that some parameters are not well set or even be able to apply them back.

**What should be done**:
Add a salt formula that checks that all system configuration needed by MetalK8s are well set.

**Implementation proposal** (strongly recommended): TBD

**Test plan**: TBD

alexandre-allard:
**Component**: salt

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2853


salt/metalk8s/internal/preflight/recommended.sls
```diff
@@ -4,7 +4,7 @@
 include:
   - metalk8s.repo
 
-{%- for pkg_name in kubeadm_preflight.mandatory.packages %}
+{%- for pkg_name in kubeadm_preflight.recommended.packages %}
```
TeddyAndrieux, line 7: Ohhh, good catch :+1:

salt/metalk8s/internal/preflight/check.sls
```diff
@@ -0,0 +1,29 @@
...
+{%- if status.error %}
+        Some sysctl parameters have been set during installation, but
+        their values are not taken in account, please check the sysctl
+        configuration to ensure MetalK8s one have precedence.
+        See above to find parameters with wrong values.
+    - result: False
+    - failhard: True
+{%- else %}
+        Success!
+{%- endif %}
```
TeddyAndrieux: Does it works ? For me, jinja is rendered at the same time as the one in `mandatory.sls` so you will not get up to date values
I think you need to use slot here (and btw slot will not be enough you need a custom module/state)
I would say a simple module to compare sysctl that just return True or false and then a slot on `result` in `test.configurable_test_state` (like we do in salt minion running state)

Also to me this state could be part of  `mandatory.sls` no ?

alexandre-allard: Yes, you're right, I just saw this when testing, I'm currently changing it.
I'm not sure it should be in `mandatory.sls`, it depends if we want it to be callable without doing any change which was the idea (so we can launch this on a regular basis through beacon/cron/whatever and send alert when it fails).

TeddyAndrieux: Right, yes it can be useful but to me, it's not the goal of a salt state if we want something like that it should be a module directly instead I think.
Maybe you should create a simple module `metalk8s_checks.py` with a function for sysctl that basically do exactly the same as this state and then in mandatory state you just do a call to this module, what do you think ? (and then in the futur we could imagine adding other checks in this modules if needed)

sayf-eddine-scality: slowly but steadily building a platformcheck clone ;p. btw did you consider having the mandatory sysctl settings in /etc/sysctl.conf directly ? that way they re not overwritten (I guess)

alexandre-allard: @TeddyAndrieux haha, that's what I'm currently doing. :)
@sayf-eddine-scality we could do that also, it works for sure.

alexandre-allard: This is the order the sysctl are read (it's like every following directories are merged, first definition of a specific filename overwrite the others, then configuration files are applied by ASCII order, with the last one, always being `/etc/sysctl.conf`):
```
/run/sysctl.d/*.conf
/etc/sysctl.d/*.conf
/usr/local/lib/sysctl.d/*.conf
/usr/lib/sysctl.d/*.conf
/lib/sysctl.d/*.conf
/etc/sysctl.conf
```
The last value defined overwrite all the previous one.

TeddyAndrieux: For me, we do not want to use `/etc/sysctl.conf` so that if one day we want to clean up a server it's really simpler to just remove our files  (cc @sayf-eddine-scality )

sayf-eddine-scality: I am just suggesting u chose.
That is not the first time we face such an issue, we simply don't have a lot of control over drop_in files. either
- we setup our own file which is recommended but we risk ppl or other tools to take precedence
- or we simply add it directly in the root file and maybe add `# added by metalk8s` to be able to clean it later.

salt/_states/metalk8s_checks.py
```diff
@@ -0,0 +1,35 @@
...
+from salt.exceptions import CheckError
+
+__virtualname__ = "metalk8s_checks"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def module_run(name, **kwargs):
```
gdemonet: Why wrap `module.run`? Since we have a custom state module and the corresponding execution module, why not just call it directly?

Something like:

```sls
Check sysctl overrides are properly applied:
  metalk8s_checks.sysctl:
    - params: {{ kubeadm_preflight.mandatory.sysctl_values | json }}
    - require:
      - cmd: Apply sysctl configuration files
```

And:

```python
def sysctl(name, params, **kwargs):
  ret = {
    "name": name,
    "changes": {},
    "result": True,
    "comment": "All sysctl settings are correctly configured."
  }
  error_msg = __salt__["metalk8s_checks.sysctl"](params, raises=False)
  if error_msg:
    ret["result"] = False
    ret["comment"] = error_msg
  return ret
```

gdemonet (commented):
most of it looks good, just a bit confused about why we want to wrap "module.run"

salt/_states/metalk8s_checks.py
alexandre-allard: It's what I did first, but it was to have something more generic.
To have a state able to run some "test" execution modules if we need to add some more tests later.
Also, wrapping module.run allow to specify multiple exec modules in one state, such as:
```
Check stuff:
  metalk8s_checks.module_run:
    - metalk8s_checks.sysctl:
        - params: {{ kubeadm_preflight.mandatory.sysctl_values | json }}
    - metalk8s_checks.some_other_tests:
        - some_args: 123
```
But maybe we don't need this and we prefer to define checks one by one.

salt/_states/metalk8s_checks.py
```diff
@@ -0,0 +1,35 @@
...
+    """Wrap `module.run` state, pretty print the error and fail gracefully if
+    the module raises a `CheckError` exception.
```
TeddyAndrieux: Minor: Mention also that this state should really do NOT changes anything on the host it's pure "checks"

salt/_modules/metalk8s_checks.py
```diff
@@ -0,0 +1,41 @@
...
+        raises (bool): the method will raise if there is any non-matching
+            sysctl value.
+    """
+    errors = []
+
+    for key, value in params.items():
+        current_value = __salt__["sysctl.get"](key)
+        if current_value != str(value):
+            errors.append(
+                "Incorrect value for '{0}', expecting '{1}' but "
```
TeddyAndrieux: Maybe just add the fact we talk about sysctl here even if it's quite clear with the value

```suggestion
                "Incorrect value for sysctl '{0}', expecting '{1}' but "
```

salt/metalk8s/internal/preflight/mandatory.sls
```diff
@@ -38,8 +38,22 @@ Set sysctl {{ item }} value to {{ value }}:
     - require:
       - kmod: Add the module br_netfilter to kernel
     {%- endif %}
+    - require_in:
+      - cmd: Apply sysctl configuration files
 {%- endfor %}
 
+{#- This is needed to ensure values loaded from files match in-memory ones. #}
```
TeddyAndrieux: Minor: We do not really care, but here it's a comment about YAML not jinja so to me it should be a YAML comment :smiley:

salt/tests/unit/modules/test_metalk8s_checks.py
```diff
@@ -0,0 +1,60 @@
...
+        Tests the return of `__virtual__` function
+        """
+        self.assertEqual(metalk8s_checks.__virtual__(), 'metalk8s_checks')
+
+    @utils.parameterized_from_cases(YAML_TESTS_CASES["sysctl"])
+    def test_sysctl(self, params, data, result, raises=False):
+        """
+        Tests the return of `sysctl` function
+        """
+        module_run_mock = MagicMock(side_effect=data.get)
```
TeddyAndrieux: Minor:

```suggestion
        sysctl_get_mock = MagicMock(side_effect=data.get)
```

TeddyAndrieux (commented):
LGTM just a few minor comments

salt/_states/metalk8s_checks.py
gdemonet: Oh. I see the idea, just wondering what benefit we have (outside of less YAML to write):
- if one module from the list fails (raises a `CheckError`), we'll only get output for this one
- the state ID for the `module.run` will not be specific, so we'll need to rely on the comment to understand what was being checked and what failed

If we really want something able to handle multiple checks (not sure it's worth the effort, TBH), we'll need to do a bit more than just wrapping `module.run` and catching the first `CheckError` that comes our way.
If we really want something able to handle any check, then maybe we don't need to wrap `module.run`, simply:

```python
def run(name, mod, **kwargs):
    ret = {
        "name": name,
        "changes": {},
        "result": True,
        "comment": "Check succeeded.",
    }

    try:
        comment = __salt__[mod](**kwargs)
    except KeyError:
        ret["result"] = False
        ret["comment"] = "Module `{}` does not exist.".format(mod)
    except CheckError as exc:
        ret["result"] = False
        ret["comment"] = str(exc)

    if comment:
        ret["comment"] = comment

    return ret
```

gdemonet: And if we really want to ensure this only runs known checks, we can restrict it with:

```python
def run(name, func, **kwargs):
    ret = { ... }

    try:
        comment = __salt__['metalk8s_checks.{}'.format(func)](**kwargs)
    except ...
```

TeddyAndrieux: Even if we only do one call per module.run at least here we have only one custom state to do all checks instead of writing a states + a module for each check we want to do. 
And btw a state for check does not really make sense at the end, a state should do some modification on the system and not only check stuff, we could imagine in the future having a "tool" that basically just run the module directly.

But no opinion about call state `module.run` vs call our module directly. Btw if we stay will `module.run` at least we can consider this state work the same as all other "classic module.run" states but it's not really an argument because anyway our states will be really simple so .... as you want

salt/tests/unit/modules/test_metalk8s_checks.py
alexandre-allard: done

salt/metalk8s/internal/preflight/mandatory.sls
alexandre-allard: done

salt/_modules/metalk8s_checks.py
alexandre-allard: done

salt/_states/metalk8s_checks.py
alexandre-allard: Done, I've also put a little explanation why we're using a state to do this.

salt/_states/metalk8s_checks.py
alexandre-allard: done

gdemonet (dismissed):

gdemonet (approved):

PR #2892, opened 2020-10-27
Documentation minor fixes

alexandre-allard:
**Component**: docs

**Context**: 
Some minor doc issues I have seen lately when going through it (for keycloak subject and debug the upgrade issues), there may be more, it makes me think that we should probably read the whole thing to see if it's still up to date, links work and every things we say in it are true. 

**Summary**:

**Acceptance criteria**:

### DIFF
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -9,11 +9,11 @@ Supported Versions
 ******************
 .. note::
 
-    MetalK8 supports upgrade **strictly** from one supported
-    minor version to another. For example:
+    MetalK8s supports upgrade of **at most** one minor version at a time.
+    For example:
 
-    - Upgrade from 2.0.x to 2.0.x
-    - Upgrade from 2.0.x to 2.1.x
+    - from 2.4.0 to 2.4.4
+    - from 2.4.0 to 2.5.1
 
     Please refer to the
     `release notes <https://github.com/scality/metalk8s/releases>`_ for more
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,7 @@
 import os
 import pathlib
 import sys
+import yaml
 
 # -- Path setup --------------------------------------------------------------
 
@@ -47,6 +48,11 @@ if versions.VERSION_SUFFIX == "-dev":
 else:
     release = versions.VERSION
 
+with open("../salt/metalk8s/defaults.yaml", "r") as fd:
+    salt_defaults = yaml.safe_load(fd)
+
+if not salt_defaults['metalk8s']['downgrade']['enabled']:
+    tags.add('downgrade_not_supported')
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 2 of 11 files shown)


docs/operation/upgrade.rst
```diff
@@ -9,8 +9,8 @@ Supported Versions
 ******************
 .. note::
 
-    MetalK8 supports upgrade **strictly** from one supported
-    minor version to another. For example:
+    MetalK8s supports upgrade **strictly** from one minor version to another.
+    For example:
 
     - Upgrade from 2.0.x to 2.0.x
     - Upgrade from 2.0.x to 2.1.x
```
TeddyAndrieux: Minor: Maybe it would make sense to change this especially since 2.0 and 2.1 will never be released (not sure it's needed to change it for every version but I think at least we should talk about released versions)

docs/operation/upgrade.rst
```diff
@@ -9,8 +9,8 @@ Supported Versions
 ******************
 .. note::
 
-    MetalK8 supports upgrade **strictly** from one supported
-    minor version to another. For example:
+    MetalK8s supports upgrade **strictly** from one minor version to another.
```
gdemonet: Not sure about this wording... @lucieleonard thoughts?

Maybe `from one minor version to the next`? Or `supports upgrade of **at most** one minor version at a time`?

alexandre-allard: I like the second one, I'll go for it.

docs/operation/upgrade.rst
alexandre-allard: Yes, why not, will change it.

gdemonet (dismissed):

docs/operation/upgrade.rst
```diff
@@ -9,11 +9,12 @@ Supported Versions
...
-    MetalK8 supports upgrade **strictly** from one supported
-    minor version to another. For example:
+    MetalK8s supports upgrade of **at most** one minor version at a time.
+    For example:
 
-    - Upgrade from 2.0.x to 2.0.x
-    - Upgrade from 2.0.x to 2.1.x
+    - ✓ Upgrade from 2.4.0 to 2.4.4
+    - ✓ Upgrade from 2.4.0 to 2.5.1
+    - ✗ Upgrade from 2.4.0 to 2.6.0
```
TeddyAndrieux: You have the exact same section in downgrade but maybe in another PR since we do not really care about downgrade for 2.6.0 (as you want)

gdemonet: Makes me think... In `docs/conf.py`, we could load `salt/metalk8s/defaults.yaml`, and read the `metalk8s.downgrade.enabled` value to define a special Sphinx tag. This way, we could show a warning in the downgrade procedure for appropriate versions.
Would likely warrant another PR (unless you feel like flakies are dissuasive :stuck_out_tongue_winking_eye:)

alexandre-allard: Done for downgrade section.
@gdemonet Great idea, did it here 53b881c7ca095f0c86d8e1090907018ebb5e1efe

gdemonet (approved):

PR #2899, opened 2020-10-28
[UI] Add integration test for sidebar and node page navigation

ChengYanJin:
**Component**: ui, tests

**Context**: 
We want to test the navigation of sidebar and node page

**Summary**:
Add integration test for the sidebar, nodelist and nodetabs in order to test the navigation.

**Acceptance criteria**: 
Pass all the tests.

**Some general questions**: 
- When it come to mock data, what dataset should we use? Should we use the real data from the real platform (6 nodes with *volumes)
- I have the impression that test the stubbed history is too close to the code implementation. Although it's a good approach to avoid the unexpected error from the new page since we only want to check if the URL is correct.

### DIFF
--- a/ui/cypress.json
+++ b/ui/cypress.json
@@ -1,6 +1,5 @@
 {
   "env": {
-    "target_url": "http://localhost:3000",
     "email": "admin@metalk8s.invalid",
     "username": "admin",
     "password": "password",
--- a/ui/cypress.json
+++ b/ui/cypress.json
@@ -15,5 +15,6 @@
     "mochaFile": "junit/cypress-junit.xml",
     "toConsole": true
   },
-  "chromeWebSecurity": false
+  "chromeWebSecurity": false,
+  "baseUrl": "http://localhost:3000"
 }
--- a/ui/cypress.json
+++ b/ui/cypress.json
@@ -15,5 +15,7 @@
     "toConsole": true
   },
   "chromeWebSecurity": false,
-  "baseUrl": "http://localhost:3000"
+  "baseUrl": "http://localhost:3000",
+  "viewportWidth": 1440,
+  "viewportHeight": 900
 }
--- a/ui/cypress.json
+++ b/ui/cypress.json
@@ -17,5 +17,6 @@
   "chromeWebSecurity": false,
   "baseUrl": "http://localhost:3000",
   "viewportWidth": 1440,
-  "viewportHeight": 900
+  "viewportHeight": 900,
+  "defaultCommandTimeout": 30000
 }
--- a/ui/cypress/integration/node/nodelist.spec.js
+++ b/ui/cypress/integration/node/nodelist.spec.js
@@ -25,7 +25,7 @@ describe('Node list', () => {
 
   it('brings me to the overview tab of master-0 Node', () => {
     cy.get('[data-cy="node_table_name_cell"]').contains('master-0').click();
-    cy.get('@historyPush').should('be.calledWith', {
+    cy.get('@historyPush').should('be.calledWithExactly', {
       pathname: '/nodes/master-0/overview',
       search: '',
     });
@@ -40,5 +40,26 @@ describe('Node list', () => {
       pathname: '/nodes/master-1/metrics',
       search: 'from=now-7d',
     });
+  });
+
+  it('brings me to create node page', () => {
+    cy.get('[data-cy="create_node_button"]').click();
+    cy.get('@historyPush').and('be.calledWithExactly', '/nodes/create');
+  });
+
+  it('updates url with the search ', () => {
+    cy.get('[data-cy="node_list_search"]').type('hello');
+    cy.get('@historyPush').and('be.calledWithExactly', '?search=hello');
+  });
+
+  it(`keeps warning severity for the alert while searching the node`, () => {
+    cy.visit('/nodes/master-0/alerts?severity=warning');
+    cy.stubHistory();
+    cy.get('[data-cy="node_list_search"]').type('hello');
+
+    cy.get('@historyPush').should(
+      'be.calledWithExactly',
+      '?severity=warning&search=hello',
+    );
   });
 });
--- a/ui/cypress/integration/node/nodelist.spec.js
+++ b/ui/cypress/integration/node/nodelist.spec.js
@@ -1,0 +1,64 @@
+beforeEach(() => {
+  cy.setupMocks();
+  cy.login();
+});
+
+// Navigation tests
+describe('Node list', () => {
+  beforeEach(() => {
+    // Visit is automatically prefixed with baseUrl
+    cy.visit('/nodes');
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+  });
+
+  // Test the navigation
+  it('brings me to the overview tab of the first Node', () => {
+    cy.fixture('kubernetes/nodes.json').then((nodes) => {
+      // It may break when we implement sorting for the nodes
+      cy.url().should(
+        'eq',
+        Cypress.config().baseUrl +
+          `/nodes/${nodes.items[0].metadata.name}/overview`,
+      );
+    });
+  });
+
+  it('brings me to the overview tab of master-0 Node', () => {
+    cy.get('[data-cy="node_table_name_cell"]').contains('master-0').click();
+    cy.get('@historyPush').should('be.calledWith', {
+      pathname: '/nodes/master-0/overview',
+      search: '',
+    });
+  });
+
+  it('brings me to another node with the same tab selected and queryString kept', () => {
+    cy.visit('/nodes/master-0/metrics');
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+    cy.get('.sc-tabs-item-content .sc-dropdown').click();
+    cy.get('[data-cy="Last 7 days"]').click();
+    cy.get('@historyPush').should('be.calledWith', {
+      search: 'from=now-7d',
+    });
+    cy.get('[data-cy="node_table_name_cell"]').contains('master-1').click();
+    // history.push({
+    //   pathname: newPath,
+    //   search: query.toString(),
+    // });
+    // For some reason, it has been seperated into 2 call.
+    cy.get('@historyPush').should('be.calledWith', {
+      search: 'from=now-7d',
+    });
+    cy.get('@historyPush').should('be.calledWith', {
+      pathname: '/nodes/master-1/metrics',
+      search: '',
+    });
+  });
+});
--- a/ui/cypress/integration/node/nodelist.spec.js
+++ b/ui/cypress/integration/node/nodelist.spec.js
@@ -8,11 +8,7 @@ describe('Node list', () => {
   beforeEach(() => {
     // Visit is automatically prefixed with baseUrl
     cy.visit('/nodes');
-    cy.window()
-      .its('__history__')
-      .then((history) => {
-        cy.stub(history, 'push').as('historyPush');
-      });
+    cy.stubHistory();
   });
 
   // Test the navigation
@@ -36,29 +32,13 @@ describe('Node list', () => {
   });
 
   it('brings me to another node with the same tab selected and queryString kept', () => {
-    cy.visit('/nodes/master-0/metrics');
-    cy.window()
-      .its('__history__')
-      .then((history) => {
-        cy.stub(history, 'push').as('historyPush');
-      });
-    cy.get('.sc-tabs-item-content .sc-dropdown').click();
-    cy.get('[data-cy="Last 7 days"]').click();
-    cy.get('@historyPush').should('be.calledWith', {
-      search: 'from=now-7d',
-    });
+    cy.visit('/nodes/master-0/metrics?from=now-7d');
+
+    cy.stubHistory();
     cy.get('[data-cy="node_table_name_cell"]').contains('master-1').click();
-    // history.push({
-    //   pathname: newPath,
-    //   search: query.toString(),
-    // });
-    // For some reason, it has been seperated into 2 call.
-    cy.get('@historyPush').should('be.calledWith', {
-      search: 'from=now-7d',
-    });
-    cy.get('@historyPush').should('be.calledWith', {
+    cy.get('@historyPush').should('be.calledOnce').and('be.calledWithExactly', {
       pathname: '/nodes/master-1/metrics',
-      search: '',
+      search: 'from=now-7d',
     });
   });
 });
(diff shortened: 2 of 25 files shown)


ui/cypress.json
```diff
@@ -15,5 +15,6 @@
     "mochaFile": "junit/cypress-junit.xml",
     "toConsole": true
   },
-  "chromeWebSecurity": false
+  "chromeWebSecurity": false,
+  "baseUrl": "http://localhost:3000"
```
gdemonet: Maybe we can get rid of `env.target_url` then?

ui/src/ducks/app/layout.test.js
```diff
@@ -1,32 +1,32 @@
...
+      type: SET_TOGGLE_SIDEBAR,
+    }),
   );
   expect(gen.next().value.type).toEqual('SELECT');
   expect(gen.next().done).toEqual(true);
 });
 
 it('should initToggleSideBar', () => {
-  localStorage.setItem(SIDEBAR_EXPENDED, 'true');
+  localStorage.setItem(SIDEBAR_EXPANDED, 'true');
```
gdemonet, line 22: Would make sense to squash this with the typo fix

ui/cypress/integration/node/nodelist.spec.js
```diff
@@ -0,0 +1,34 @@
+beforeEach(() => {
+  cy.setupMocks();
+  cy.login();
+});
+
+// Navigation tests
+describe('Node list', () => {
+  beforeEach(() => {
+    cy.server();
```
gdemonet: No need to re-run `cy.server`, that's already part of `setupMocks`

ui/cypress/integration/node/nodelist.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    cy.route('api/kubernetes/api/v1/nodes', 'fixture:kubernetes/nodes.json');
```
gdemonet: If this route is expected to be useful _in general_, then I'd suggest you define it in `cy.setupMocks`

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+  });
+
+  const tabs = ['alerts', 'metrics', 'volumes', 'pods', 'details'];
+  tabs.map((tab) => {
+    it(`brings me to the [${tab}] tab`, () => {
```
gdemonet: We shouldn't use square brackets in test names - if the test fails, the generated screenshot will have square brackets in its name and we cannot upload such files to artifacts today (discovered this when debugging CI issues).

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
+beforeEach(() => {
+  cy.setupMocks();
+  cy.login();
+});
+
+// Navigation tests
+describe('Overview tab in master-0 node page', () => {
```
gdemonet: That's not a generic "context" name - what you are testing is rather `Node page overview tab` (no need to specify a node name)

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+  });
+
+  const tabs = ['alerts', 'metrics', 'volumes', 'pods', 'details'];
+  tabs.map((tab) => {
+    it(`brings me to the [${tab}] tab`, () => {
+      cy.get('.sc-tabs-item')
+        .eq(tabs.indexOf(tab) + 1)
```
gdemonet: Can we define a `data-cy` attribute for those tabs?

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    cy.route('api/kubernetes/api/v1/nodes', 'fixture:kubernetes/nodes.json');
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+  });
+
+  const tabs = ['alerts', 'metrics', 'volumes', 'pods', 'details'];
+  tabs.map((tab) => {
```
gdemonet: This is not specific to the `overview` tab, so maybe it makes sense get this test in a more generic context (e.g. `Node page tabs`), and make sure that we can reach any tab from any other tab (and also check that clicking the current tab doesn't do anything).

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+});
+
+describe('Metrics tab in master-0 node page', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('api/kubernetes/api/v1/nodes', 'fixture:kubernetes/nodes.json');
+    cy.route(
+      'api/prometheus/api/v1/query?query=node_uname_info',
+      'fixture:prometheus/node-uname-info.json',
+    );
```
gdemonet: Same, this one makes sense to keep in the general `setupMocks`

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+});
+
+describe('Volumes tab in master-0 node page', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('api/kubernetes/api/v1/nodes', 'fixture:kubernetes/nodes.json');
+    cy.route(
+      '/api/kubernetes/apis/storage.metalk8s.scality.com/v1alpha1/volumes',
+      'fixture:kubernetes/volumes.json',
+    );
```
gdemonet: Ditto

gdemonet (commented):
Well done :+1: !

> When it come to mock data, what dataset should we use? Should we use the real data from the real platform (6 nodes with *volumes)

Yeah, let's try to get real-world data (hard-coded for now, we'll see how to get them from a real deployment, maybe from CI, to make sure our UI keeps working when APIs evolve).

> I have the impression that test the stubbed history is too close to the code implementation. Although it's a good approach to avoid the unexpected error from the new page since we only want to check if the URL is correct.

What would you rather do? IMHO, all we want is to make sure the "link" is correct. But unlike some elements which have an `href` we can check, most of our "links" are JS-only, so we need to catch the call made to `history.push` to see what their `href` would be.

ui/cypress.json
ChengYanJin: Well, I am not sure we want to remove it. The cost of removing it would be hardcode this `cy.visit(http://localhost:3000)` all the times.

ui/cypress/integration/node/nodetabs.spec.js
ChengYanJin: Thanks for letting me know this point!

ui/cypress.json
gdemonet: ?? Isn't the point of `baseUrl` to always prefix any path for `cy.visit`?
We would just need to replace any `cy.visit(Cypress.env('target_url'))` by `cy.visit('/')`.

ChengYanJin: Indeed!! ;)

ui/cypress/integration/node/nodelist.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
+    cy.get('.sc-tabs-item-content .sc-dropdown').click();
+    cy.get('[data-cy="Last 7 days"]').click();
+    cy.get('@historyPush').should('be.calledWith', {
+      search: 'from=now-7d',
+    });
```
gdemonet: That's likely not what you want. Here, since you're stubbing `history.push`, you won't have the right URL before switching to `master-1`. That's why your call doesn't contain the `search` part you're expecting.

Instead, you should directly call `cy.visit('/nodes/master-0/metrics?from=now-7d')`.

Also, side-note but quite important: a stub can store many calls for you to check after. When using `should('be.calledWith', args)`, you're only asserting it was called at least once with `args`. In your case, the call with `{ search: 'from=now-7d' }` comes from the click on `cy.get('[data-cy="Last 7 days"]')`, not from the switch to another node.

If we want to clean this test and make sure it does exactly what we want to, here's how we could do it:

```javascript
cy.visit('/nodes/master-0/metrics?from=now-7d');
cy.stubHistory(); // would be smart to create this command, since we seem to use it often
cy.contains('[data-cy="node_table_name_cell"]', 'master-1').click();
cy.get('@historyPush')
  .should('be.calledOnce')
  .and('be.calledWithExactly', {
    pathname: '/nodes/master-1/metrics',
    search: 'from=now-7d',
  });
```

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,173 @@
...
+    it(`brings me to the ${tab} tab`, () => {
+      cy.get(`[data-cy="${tab}_tab_node_page"]`).click();
+      cy.get('@historyPush').should('be.calledWith', `/nodes/master-0/${tab}`);
+    });
+  });
+
+  it('does not switch the tab', () => {
+    cy.get(`[data-cy="overview_tab_node_page"]`).click();
+    cy.get('@historyPush').should('not.be.called');
+  });
```
gdemonet: Could be factorized between source tabs:

```javascript
const allTabs = ['overview', 'alerts', 'metrics', 'volumes', 'pods', 'details'];
const tabSwitchingTestCases = (sourceTab) => {
  allTabs.filter((tab) => tab !== sourceTab).map((tab) => {
    it(`brings me to the ${tab} tab`, () => {
      cy.get(`[data-cy="${tab}_tab_node_page"]`).click();
      cy.get('@historyPush').should('be.calledWith', `/nodes/master-0/${tab}`);
    });
  });

  it('does not switch to the current tab', () => {
    cy.get(`[data-cy="${sourceTab}_tab_node_page"]`).click();
    cy.get('@historyPush').should('not.be.called');
  });
};

describe('Overview tab', () => {
  // beforeEach...

  tabSwitchingTestCases('overview');

  // more tests...
});

// more tabs...
```

ui/cypress/integration/node/nodetabs.spec.js
```diff
@@ -0,0 +1,173 @@
...
+
+  const alertSeverities = ['critical', 'warning'];
+  alertSeverities.map((severity) => {
+    it(`brings me to the alert tab prefiltered by ${severity} severity of master-0 node`, () => {
```
gdemonet:
```suggestion
    it(`brings me to the alert tab prefiltered by ${severity} severity`, () => {
```

ui/cypress/integration/sidebar.spec.js
```diff
@@ -0,0 +1,78 @@
...
+    cy.visit('/');
+  });
+
+  // TODO: Remove the interaction from the E2E test as much as possible
+  it('brings me to the monitoring page', () => {
+    cy.window()
+      .its('__history__')
+      .then((history) => {
+        cy.stub(history, 'push').as('historyPush');
+      });
```
gdemonet: Let's at least put this in `beforeEach` :slightly_smiling_face:

ui/cypress/integration/sidebar.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      .its('localStorage')
+      .then((store) => {
+        isSidebarExpanded = store.getItem('sidebar_expanded');
+      })
+      .then(() => {
+        // Cypress commands are asynchronous, so you cannot check a property value before the Cypress commands ran.
+        // use cy.then() callback to check the value.
+        expect(isSidebarExpanded).to.equal('false');
+      });
+  });
```
gdemonet: If this test checks that the sidebar `can expand`, I would expect `isSidebarExpanded` to be true, at the end.

Now, this test should not depend on "the default state of the app" (here, it's expanded by default), otherwise the test will fail once we change this default, while the feature still works.

Here's how I would rewrite those tests:

```javascript
  it('is expanded by default', () => {
    // assuming we have set the proper attributes
    // 'data-cy' and 'data-cy-state' in `sidebarConfig`
    cy.get('[data-cy="sidebar"]').should(
      'have.attr',
      'data-cy-state',
      'expanded',
    );

    cy.window()
      .its('localStorage')
      .invoke('getItem', 'sidebar_expanded')
      .should('equal', 'true');
  });

  it('can be collapsed', () => {
    cy.window()
      .its('localStorage')
      .invoke('setItem', 'sidebar_expanded', 'true');

    cy.get('[data-cy="sidebar"] > :first').click();

    cy.get('[data-cy="sidebar"]').should(
      'have.attr',
      'data-cy-state',
      'collapsed',
    );
    cy.window()
      .its('localStorage')
      .invoke('getItem', 'sidebar_expanded')
      .should('equal', 'false');
  });

  it('can be expanded', () => {
    cy.window()
      .its('localStorage')
      .invoke('setItem', 'sidebar_expanded', 'false');

    cy.get('[data-cy="sidebar"] > :first').click();

    cy.get('[data-cy="sidebar"]').should(
      'have.attr',
      'data-cy-state',
      'expanded',
    );
    cy.window()
      .its('localStorage')
      .invoke('getItem', 'sidebar_expanded')
      .should('equal', 'true');
  });

gdemonet (changes requested):
Few improvements to make on the tests, but that's starting to look really good! Making changes to core-ui or some navigation logic in the future should be done without worry that it breaks the user-facing behaviour of our UI :ok_hand: :sunglasses:!

ChengYanJin:
> Few improvements to make on the tests, but that's starting to look really good! Making changes to core-ui or some navigation logic in the future should be done without worry that it breaks the user-facing behaviour of our UI 👌 😎!

Indeed, we do spend some effort on implementing the navigation test, although we can test it manually for the moment. I would agree it's more for the future, our coming changes won't break the UI. 

I also read more about cy.stub(), have a better understanding now^^

Thanks for the review!

ui/cypress/integration/sidebar.spec.js
ChengYanJin: In the third case, we may not be able to get the default sidebar_expanded item in` localStorage` though.
Because if do nothing to the sidebar, we don't set it in localStorage. the default state would be in the redux store. 

We could check the data in the redux store, but I think to check the attr of .sc-sidebar the sidebar would be sufficient. Would you think so?

gdemonet: I guess you mean "in the first case", i.e. `it('is expanded by default')`?
I did see this when running the app with a clear localStorage (nothing set for `sidebar_expanded`), but somehow when running using Cypress it was set (and checking that it wasn't set failed...). Any idea why that would be the case?

gdemonet (approved):

PR #2900, opened 2020-10-28
[UI] Volumes Sorting

### LINKED ISSUE #2876: Ability to sort list of Volumes in the Volume page

alexis-ld:
**Component**:
ui, volumes

**Context**: 
We want to add sorting to the Volumes list.

**Summary**:
I implemented sorting capacities using react-table's custom sorting functions. I also pushed the sorting parameters in the URL as required in #2876 . If a sorting is set in the URL when loading the page it is applied if not is default back to health sorting.

**Acceptance criteria**: 
Clicking on a column name sorts the list based on the selected parameter.
The sorting for each columns follows the specifications described in #2876 .
If a sort( and optional desc) parameter i set in the URL it is applied on page load.

---

Closes: #2876

### DIFF
--- a/ui/src/components/VolumeListTable.js
+++ b/ui/src/components/VolumeListTable.js
@@ -130,7 +130,6 @@ const SortCaretWrapper = styled.span`
 `;
 
 const SortIncentive = styled.span`
-  padding-left: ${padding.smaller};
   position: absolute;
   display: none;
 `;
--- a/ui/src/components/VolumeListTable.js
+++ b/ui/src/components/VolumeListTable.js
@@ -20,15 +20,11 @@ import CircleStatus from './CircleStatus';
 import { Button, ProgressBar, Tooltip } from '@scality/core-ui';
 import { intl } from '../translations/IntlGlobalProvider';
 import {
-  STATUS_WARNING,
-  STATUS_CRITICAL,
-  STATUS_NONE,
-  STATUS_HEALTH,
   VOLUME_CONDITION_LINK,
   VOLUME_CONDITION_UNLINK,
   VOLUME_CONDITION_EXCLAMATION,
 } from '../constants';
-import { allSizeUnitsToBytes } from '../services/utils';
+import { allSizeUnitsToBytes, compareHealth } from '../services/utils';
 
 const VolumeListContainer = styled.div`
   color: ${(props) => props.theme.brand.textPrimary};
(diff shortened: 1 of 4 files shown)


thomasdanan (commented):
Works well.
Just noticed that some columns header are centered and some other are aligned on the left. When it is centered, activating the makes the header to move on the left, which is not great. Shouldn't we have all headers aligned on the left so that when we activate sorting, headers are not moving on the left?

Also, it is not super obvious that one can click on the header to sort. Could we make the sorting icon to appear when we mouse over the column header?

Since there is a default sorting (on Health), shouldn't we display the fact that the table is sorted on health when loading on the page?

alexis-ld:
@thomasdanan Just added a fix for the header moving on sort (I managed to keep their initial positioning but still not move them when the sorting caret appears). I also added the sort incentive on hover.

Working on the initial sorting caret...

thomasdanan:
> @thomasdanan Just added a fix for the header moving on sort (I managed to keep their initial positioning but still not move them when the sorting caret appears). I also added the sort incentive on hover.
> 
> Working on the initial sorting caret...

Thanks @alexis-ld, actually we had a discussion with @ChengYanJin and with our UX designer. We were saying that for table values, icon values should be center aligned, string values should be left-aligned and numerical values should be right-aligned. Should this rule apply to the header or should we have headers always left-aligned? we were thinking having header left-aligned would make it easier for you to fix the moving header issue.

alexis-ld:
@thomasdanan for now the code applies the same alignement on the header as the value. It looks a bit weird to have a centered value with a left header actually (the offset is quite important).

Anyway I fixed the moving header thing without moving the header to the left so we can keep it like that if that works for you.

alexis-ld:
I found a way to use react-table's default sorting without broking the URL sync. Should be all good now :) .

ui/src/components/VolumeListTable.js
```diff
@@ -115,6 +124,25 @@ const TooltipContent = styled.div`
   min-width: 60px;
 `;
 
+const SortCaretWrapper = styled.span`
+  padding-left: ${padding.smaller};
+  position: absolute;
+`;
+
+const SortIncentive = styled.span`
+  padding-left: ${padding.smaller};
```
ChengYanJin: I feel it's better to remove the padding-left, so when we click on the sort the caret would not jump.

ChengYanJin (changes requested):
It works!! 

I still encounter the infinite loop of URL by the following steps:
Sort by other columns => click on "Create Volume" => click on "Cancel" button. 

Can we try to set the initial sort in React-table, instead of initialing the data sorting in NodeVolumesUtils.js?

ChengYanJin (approved):
LGTM!

Issue #2726, opened 2020-08-10
Pod failed to mount SparseLoopDevice Volume after node reboot


thomasdanan:
**Component**:

'containers', 'kubernetes', 'volumes'

**What happened**:

When provisioning  a Volume using SparseLoopDevice mode, a pod consuming such volume may fail to start after a node reboot, because it is unable to mount the volume since its path on disk may not exist anymore:

`   Warning  FailedMount  90s (x23 over 32m)   kubelet, master-1  MountVolume.NewMounter initialization failed for volume "master-1-prometheus" : path "/dev/disk/by-uuid/183dd142-9c46-411f-b160-d99bbd03f6c9" does not exist`

**Workaround: restart salt-minion on all nodes where some Volumes were created using SparseLoopDevice mode**

**What was expected**:

As a persistent Volume, we expect its path on disk to be persistent so that the pod can mount it after a reboot and retrieve data that were stored under it.

**Steps to reproduce**

- Deploy MetalK8s cluster
- Provision a Volume using SparseLoopDevice mode, a PVC to bind it, and a pod to mount this PVC
- Wait for pod to start (and PVC to bind the volume's PV)
- Reboot the node on which the volume was provisioned / pod was scheduled
- Observe that the pod may stay in _Pending_ state because it is not able to find the device referenced by its PVC volume

**Resolution proposal** (optional):

NicolasT:
This can be worked around by applying the Salt state responsible for creating the loop devices. This is supposed to happen at node reboot (using `salt-minion`'s 'startup state execution' feature), however, this fails when `salt-master` is not (yet) available when the `salt-minion` service starts.

This is a nuisance, at least.

As discussed [on Slack](https://scality.slack.com/archives/CFXQQUDPT/p1605180081490900), I see two ways to 'fix' this:

- Implement a CSI driver which provisions and handles sparse files as volumes (an interesting project, but quite a bit of work)
- No longer make Salt responsible to set up the loop device, and instead rely on the system (`systemd`) to set these up and tear them down, e.g. at system boot.

Given the second approach, here's an outline of how this can be deployed:

- We can use a "[systemd template unit](https://fedoramagazine.org/systemd-template-unit-files/)" to manage the loop device(s). To deploy, we can drop the following file as `/etc/systemd/system/metalk8s-sparse-volume@.service`:

```
[Unit]
Description=Setup MetalK8s sparse loop device %I
RequiresMountsFor=/var/lib/metalk8s/storage/sparse

[...]

PR #2905, opened 2020-10-30
Launch multi node upgrade tests in ci

### LINKED ISSUE #1128: Run the upgrade tests in a "real" multinode environment
**Component**: tests

**Why this is needed**: From #1125, #1126, #1127 we have a good test suite for single node upgrade. Let's have this for a bigger cluster.

**What should be done**:

Expand the cluster to its [final form](https://impossiblehq.com/wp-content/uploads/2013/04/Final-Form.jpg):
* 5 nodes
  * 3 `etcd` nodes
  * 3 control-plane nodes
  * 2 workload-plane nodes

This means that we will need to have code for expansion procedure for every different version of expansion procedure.

**Test plan**: 

- [ ] We are able to run the upgrade suite with this cluster
- [ ] The CI triggers these expansions before the upgrade in its test suite.

TeddyAndrieux:
**Component**:

'build', 'tests', 'lifecycle'

**Context**: 

#1128 

**Summary**:

- Backport fb2b5a4
- Add a workaround for #2726 in snapshot upgrade stage
- Add 3 nodes snapshot upgrade tests in CI

---

Fixes: #1128

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -1617,7 +1617,26 @@ stages:
             TF_VAR_rhsm_username: "%(secret:rhel_ci_login)s"
             TF_VAR_rhsm_password: "%(secret:rhel_ci_password)s"
       # --- Wait for all pods to be running ---
-      - ShellCommand: *wait_pods_running_ssh
+      - ShellCommand:
+          <<: *wait_pods_running_ssh
+          # NOTE: Do not use `wait_pods_running_ssh` directly as we need to
+          # have a workaround to make sure all PVs are well recreated by
+          # storage operator
+          # Check https://github.com/scality/metalk8s/issues/2726
+          # As workaround if `wait_pods_status` failed just restart salt-minion
+          # everywhere and `wait_pods_status` again
+          command: >
+            git checkout "%(prop:branch)s" &&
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -1617,7 +1653,26 @@ stages:
...
+            scp -F ssh_config "%(prop:builddir)s/build/eve/wait_pods_status.sh" bootstrap:/tmp/ &&
+            ssh -F ssh_config bootstrap sudo bash -c "
+            /tmp/wait_pods_status.sh --sleep-time 5 --stabilization-time 30 --status Running
+            --retry 30 || (crictl exec -it
+            \$(sudo crictl ps -q --label io.kubernetes.pod.namespace=kube-system
+            --label io.kubernetes.container.name=salt-master --state Running)
+            salt '*' cmd.run_bg 'salt-call --local service.restart salt-minion' &&
+            sleep 10 &&
+            /tmp/wait_pods_status.sh --sleep-time 5 --stabilization-time 30 --status Running
+            --retry 30)"
```
alexandre-allard, line 1675: Shouldn't we always restart salt-minion, "just in case", instead of having this ?

alexandre-allard (approved):

eve/main.yml
TeddyAndrieux: Hmm right, we could, but since the logic is here now let's keep it for the moment :smile:

PR #2907, opened 2020-10-30
[UI] Bugfix the unused volume health status should be none

### LINKED ISSUE #2902: The unused/unbound volumes should not in healthy status.
**Component**: ui

**What happened**:
For the unused/unbound volume, it is in **healthy** state with green circle.
image

**What was expected**:
Since the volume is not bound, so we don't have PVC, then no monitoring. So we can't get any alerts, but it doesn't mean it's healthy.  

**Steps to reproduce**
Create a new volume.

Set **none** health state for the unused volumes (A circle without filled color)
image
Improvement:
We should give a message in the Alerts tab, saying that "No alert because the volume is unused"
image

ChengYanJin:
**Component**: ui, bugs

**Context**: 
Since the volume is not bound, so we don't have PVC, then no monitoring. So we can't get any alerts, but it doesn't mean it's healthy.

**Summary**:
- Set none health state for the unused volumes (A circle without filled color)
- For unbound volume, display "The volume is not bound yet" in the alerts tab

**Acceptance criteria**: 
image
image

---

Closes: #2902

### DIFF
--- a/ui/src/services/NodeVolumesUtils.js
+++ b/ui/src/services/NodeVolumesUtils.js
@@ -217,7 +217,7 @@ export const getVolumeListData = createSelector(
         // compute the volume health based on the severity of the alerts
         // critical => if there is at least one critical
         if (volumeAlerts.length) {
-          volumeHealth = volumeAlerts.find(
+          volumeHealth = volumeAlerts.some(
             (vol) => vol.labels.severity === STATUS_CRITICAL,
           )
             ? STATUS_CRITICAL
--- a/ui/src/services/NodeVolumesUtils.js
+++ b/ui/src/services/NodeVolumesUtils.js
@@ -212,23 +212,23 @@ export const getVolumeListData = createSelector(
           (alert) =>
             alert.labels.persistentvolumeclaim === volumePVC.metadata.name,
         );
-      } else {
-        volumeHealth = STATUS_NONE;
-      }
 
-      // THE RULES TO COMPUTE THE HEALTH
-      // compute the volume health based on the severity of the alerts
-      // critical => if there is at least one critical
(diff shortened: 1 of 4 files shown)


ui/src/services/NodeVolumesUtils.js
```diff
@@ -211,23 +211,23 @@ export const getVolumeListData = createSelector(
...
-        )
-          ? STATUS_CRITICAL
-          : STATUS_WARNING;
-      } else if (volumeComputedCondition === ('exclamation' || 'unlink')) {
-        volumeHealth = STATUS_NONE;
+        // THE RULES TO COMPUTE THE HEALTH
+        // compute the volume health based on the severity of the alerts
+        // critical => if there is at least one critical
+        if (volumeAlerts.length) {
+          volumeHealth = volumeAlerts.find(
```
slaperche-scality: nitpick: instead of using `find`, I would rather use [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

> The some() method tests whether at least one element in the array passes the test implemented by the provided function. It returns a Boolean value. 

Reason being: you don't want to **find** the first critical alert (i.e you don't care about the alert object itself), you only want to know if there are **some** critical alerts in the array.
=> it's more logical to use `some` than `find`.

ChengYanJin: Thanks Sylvain!! It definitely makes sense, we just need a Boolean to tell us if there is a critical alert.

thomasdanan (dismissed):
I am okay with the new behaviour

thomasdanan (approved):

PR #2912, opened 2020-11-03
salt: Add function to sort using a cmp function

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Migration to Python3

NOTE: Put this function in development/2.6 so that it can be used during downgrade from 2.7.x to 2.6.x and avoid some "hack" in the salt downgrade orchestrate

**Summary**:

In python3 `sorted` and `sort` function do not support `cmp` function
compare to python2 that does (but it's easily doable using
`functools.cmp_to_key`).
For Jinja in Salt states it's useful function so add a salt module to
do this `sorted` using a `cmp` function same as it was done in Python2

---

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -2,6 +2,7 @@
 '''
 Module for handling MetalK8s specific calls.
 '''
+import functools
 import logging
 import os.path
 import re
@@ -343,3 +344,14 @@ def format_slots(data):
             )
 
     return data
+
+
+def cmp_sorted(*args, **kwargs):
+    """Helper to sort a list using a function to compare (as `cmp` in Python2)
+
+    Useful when we want to sort a list in Jinja
+    """
+    if 'cmp' in kwargs:
+        kwargs['key'] = functools.cmp_to_key(kwargs.pop('cmp'))
+
+    return sorted(*args, **kwargs)
--- a/salt/tests/unit/modules/test_metalk8s.py
+++ b/salt/tests/unit/modules/test_metalk8s.py
@@ -307,3 +307,18 @@ class Metalk8sTestCase(TestCase, LoaderModuleMockMixin):
                     metalk8s.format_slots(data),
                     result
                 )
+
+    @parameterized.expand([
+        param([3, 5, 2, -4], [-4, 2, 3, 5]),
+        param([3, 5, 2, -4], [5, 3, 2, -4], reverse=True),
+        param([3, 5, 2, -4], [5, 3, 2, -4], cmp=lambda a, b: -1 if a > b else 1),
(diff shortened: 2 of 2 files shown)


salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -307,3 +307,18 @@ def test_format_slots(self, data, result, slots_returns=None,
...
+
+    @parameterized.expand([
+        param([3, 5, 2, -4], [-4, 2, 3, 5]),
+        param([3, 5, 2, -4], [5, 3, 2, -4], reverse=True),
+        param([3, 5, 2, -4], [5, 3, 2, -4], cmp=lambda a, b: -1 if a > b else 1),
+        param([3, 5, 2, -4], [2, 3, -4, 5], key=abs)
+    ])
+    def test_cmp_sorted(self, obj, result, **kwargs):
+        """
+        Tests the return of `format_slots` function
```
slaperche-scality: Wrong docstring.

salt/_modules/metalk8s.py
```diff
@@ -343,3 +344,14 @@ def format_slots(data):
             )
 
     return data
+
+
+def cmp_sorted(*args, **kwargs):
+    """Helper to sort a list using a function to compare (as `cmp` in Python2)
+
+    Usefull when we want to sort a list in Jinja
```
slaperche-scality: `Useful`

salt/_modules/metalk8s.py
```diff
@@ -343,3 +344,14 @@ def format_slots(data):
             )
 
     return data
+
+
+def cmp_sorted(*args, **kwargs):
```
slaperche-scality, line 349: I suppose at one point in time we will be able to remove this compat layer and fix the calling code.
Or are we gonna stay with those workaround forever?

TeddyAndrieux: For me, we will need it for a long time since I do not see another way to do some sorting using a specific function in Jinja (Python3).

(Except if we change the downgrade (and upgrade) logic, which is currently in discussion, but this function may be used for another purpose I think)

slaperche-scality: So why can't we use `key`?
At one point in the future, we will be full python3 and there will be no need for `cmp` no?

TeddyAndrieux: Because we can't easily in this context
The purpose for the downgrade is to sort node versions so you need to compare versions, if we want to use `key` instead of `cmp` we need to create dedicated objects for a version and store them... in the pillar (or have "something" (likely a function) to get a representation of a version that can be compared easily but in this case we cannot use the `pkg.version_cmp` from salt that may have some logic we want here), for me, it's not possible easily in salt context, especially for versions

gdemonet (approved):
Fine for me

PR #2913, opened 2020-11-03
[UI] Align Nodes and Volumes page design

### LINKED ISSUE #2895: Node and Volume pages should be better aligned in terms of design
**Component**: 'ui'

**Why this is needed**:

**The general goal of this ticket:**
The final goal is to have the **exactly the same** layout/size of the table/size of the right side panel, the paddings between Node Page and Volume Page. So when we switch over, we should feel only the contents change. 

While playing with the UI with the new Volume and Node pages, I noticed few inconsistencies that are worth to tackle in order to have a better UX: 

1. Node or Volume health status in overview tab are not of the same size in right panel
The size used in Volume overview tab is the one to use. But should be aligned with the following text
image

2. Health status is first col in volumes and third col in node status
Health should be in the first column in Node table, with Health header. Note that the **size** of the circle should be the same as volume table, a bit different from the design.
node-crea-05-03-node-v5-overview-tab-initial@2x

3. No Search field different background in Volume page
We should use background as in the following screenshot:
 Screenshot 2020-10-27 at 17 15 04

4. Margin between left table and right tabbed panel is different between vol and node page (prefer the vol page one)
We should use the margin as it is defined in the Volume page
Update: Should be 8px for both Volume and Node page

5. Size of the table are different between node and vol page
- We should have the same size of the table and right side panel for both Node Page and Volume Page. Since Volume Page has much more information, we should take the size of the volume table and apply it to the Node table. 
- We should have the volume table as small as possible. (We can reduce some width in volume table, since we have space)
image

6. Right margin (after tabbed panel) is different between node and vol page
We should use the same margin that is defined in Volume page. 

7. Create Node is just next to search field but Create Volume is at extreme right
Create Node button should be on the right but with proper margin on the right,

8. Active alert component is not identical between node and vol Overview tab
 In the Volume page, the alert component is with fixed size and doesn't move whatever content we put on the left. Should be the same size and same behaviour in Node page

9. Height of the node tabbed panel takes the whole page and is different from vol tabbed panel
Height of the tab should take all the space but with padding between the button side. 

10. Volume tab in Node tabbed panel is having lot of waste space (could use more margin or align differently)
According to the [Node Page - volume tab Design](https://www.sketch.com/s/f7ae733f-222f-4335-b742-a86d25dda872/a/wgkvaw) we should use more margin.

11. Create new Volume in Volume tab in Node tabbed panel is gray when in other pages, Create button is blue
All create buttons should be in blue

12. Size and colour of advanced metrics buttons are different between vol and node page
Volume page uses the correct one #434343

13. I think the width of the tables could be smaller
Please see 5th point.

14. table and search field alignments are different between node and volume pages
Should follow the design

15. Rename the "Create a New Volume" button to "Create Volume" above the volume table

16. The tabs should take all the space but with proper padding with the bottom. 

17. The Node table should have padding with the bottom, like the volume table.  
The table and right-side panel should align with each other.

18. The fontsize of Volume tabs is not correct, should be 14px.

19. Space between numerical value and unit. For exmaple: 5 GiB, 10 TiB

20. The Headers of the table should always align on the left.

Screenshot 2020-10-28 at 19 34 02

**What should be done**:

Some of the improvements may require some changes in core-ui, some other may require some factorization in metalk8s-ui, some are just little adjustments

**Note that:**
We should extract a common style for left side panel and right side panel, and put it in the `CommonLayoutStyle`.

**Test plan**:
TBD

alexis-ld:
**Component**:
ui, nodes, volumes

**Context**: 
We want to align designs for the Nodes and the Volumes page.

**Summary**:
Small tweaks, refactoring and CSS across both pages.

**Acceptance criteria**: 
Volumes and Nodes pages design are matching and switching from one to the other only changes the design.
All points in ticket #2895 are fixed.

---

Closes: #2895


ChengYanJin (changes requested):
Hello @alexis-ld ,

Thanks for the work! 

A general suggestion:
When we create a new branch, please make sure it's based on the latest `development/2.6` branch so that we have all the new features. 

Here are some of my early reviews whereas I am playing with it.
1. We may forget to consider No.15 (Rename the "Create a New Volume" button to "Create Volume" above the volume table)
2. We want to have the same margin-bottom for the left-side table and right-side panel ( and it should apply to **ALL** the tabs)
image
3. In the Node page Metrics tab, it seems like when I scroll down the entire panel goes down.
4. For non deployed node, we don't have the alert counter in the overview tab, neither the number in the Alert Tab.
5. When switch the node, can we try to avoid having the content of the overview tab moving? 
6. When I switch from node page to volume page, I still notice there is a slight shift of the search bar above the table. 
7. The style of ActiveAlertCounter on Node page should be the same as the volume page. I mean the **padding** between the title and the actual component.
8. The link of **node name** on Volume Overview tab is not correct. I've fixed it before, please **rebase** to the lastest development/2.6

Please come back to me if there is any point not clear.
Thanks!

thomasdanan:
> Hello @alexis-ld ,
> 
> Thanks for the work!
> 
> A general suggestion:
> When we create a new branch, please make sure it's based on the latest `development/2.6` branch so that we have all the new features.
> 
> Here are some of my early reviews whereas I am playing with it.
> 
> 1. We may forget to consider No.15 (Rename the "Create a New Volume" button to "Create Volume" above the volume table)
> 2. We want to have the same margin-bottom for the left-side table and right-side panel ( and it should apply to **ALL** the tabs)
>    image
> 3. In the Node page Metrics tab, it seems like when I scroll down the entire panel goes down.
> 4. For non deployed node, we don't have the alert counter in the overview tab, neither the number in the Alert Tab.
> 5. When switch the node, can we try to avoid having the content of the overview tab moving?
> 6. When I switch from node page to volume page, I still notice there is a slight shift of the search bar above the table.
> 7. The style of ActiveAlertCounter on Node page should be the same as the volume page. I mean the **padding** between the title and the actual component.
> 8. The link of **node name** on Volume Overview tab is not correct. I've fixed it before, please **rebase** to the lastest development/2.6
> 
> Please come back to me if there is any point not clear.
> Thanks!

To your points @ChengYanJin, and from what I saw with last commit
1. done
2. done
3. I am not sure to get your point: to me the tabs are staying always visible, only the tab content is going down. Indeed we may want to keep time selector and advanced metrics button always visible, we could address this in future PR
4. I think it is okay now
5. I guess you are talking about switching from non deployed node to deployed node?
6. I think it is okay now
7. I think it is okay now
8. seems okay to me with last commit

To me we are good enough to close this polish ticket. I didn't check the code thought ...

thomasdanan (dismissed):
thanks a lot !!!

ChengYanJin (dismissed):
Thanks for the improvements!!!

The code looks good for me! 
I pushed a commit to improve the regex to add a space for the storage capacity. Please let me know if it's good for you.

As @thomasdanan said,
We may want to keep time selector and advanced metrics button always visible in the tabs, we could address this in future PR. 

We can still improve on `Volume Page Metrics tab`, in order to fully use the empty space (Please refer to the Node Page Metrics tab, maybe we should generalize some common style). But can be done in future PR too.

alexis-ld:
@ChengYanJin Thank you very much for the regex. :)

thomasdanan (dismissed):

thomasdanan (approved):

PR #2915, opened 2020-11-05
[UI] : Volumes page - fix browser go back button

### LINKED ISSUE #2896: Browser backward button doesn't work when Volume page is displayed
**Component**:

'ui'

**What happened**:

When landing on the Volume page it is not possible to navigate backward, using the browser backward button.
Actually, the URL is not changed and consequently, we always stay on the Volume page

**What was expected**:

When Volume page is displayed and when the administrator is clicking on the browser backward button, he should be redirected to the previous URL in the browser history

**Steps to reproduce**

- DO: Navigate through different pages until you reach the Volume page
- DO: click on browser backward button
- OBSERVE: nothing happens
- EXPECT: browser should display previously visited page

alexis-ld:
**Component**:
ui,volumes

**Context**: 
Browser "go back" button was broken on Volumes view.

**Summary**:
Settings the URL queries and selected volume was pushing in history instead of replacing it breaking the go back action.

**Acceptance criteria**: 
Clicking on the "go back" button on Volumes page goes back to previous page.

---

Closes: #2896

### DIFF
--- a/ui/src/components/VolumeListTable.js
+++ b/ui/src/components/VolumeListTable.js
@@ -163,7 +163,7 @@ function GlobalFilter({
     } else {
       searchParams.set('search', value);
     }
-    history.replace(`?${searchParams.toString()}`);
+    history.push(`?${searchParams.toString()}`);
   }, 500);
 
   return (
--- a/ui/src/components/VolumeListTable.js
+++ b/ui/src/components/VolumeListTable.js
@@ -163,7 +163,7 @@ function GlobalFilter({
     } else {
       searchParams.set('search', value);
     }
-    history.push(`?${searchParams.toString()}`);
+    history.replace(`?${searchParams.toString()}`);
   }, 500);
 
   return (
@@ -290,7 +290,7 @@ function Table({
         query.delete('sort');
         query.delete('desc');
       }
-      history.push(`?${query.toString()}`);
+      history.replace(`?${query.toString()}`);
     }
   }, [sorted, desc, history, data.length]);
 
--- a/ui/src/containers/VolumePageContent.js
+++ b/ui/src/containers/VolumePageContent.js
@@ -181,22 +181,22 @@ const VolumePageContent = (props) => {
     {
       selected: isOverviewPage,
       title: intl.translate('overview'),
-      onClick: () => history.push(`${match.url}/overview${query.toString() && `?${query.toString()}`}`),
+      onClick: () => history.replace(`${match.url}/overview${query.toString() && `?${query.toString()}`}`),
     },
     {
       selected: isAlertsPage,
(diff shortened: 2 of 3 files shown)


ui/src/containers/VolumePageContent.js
```diff
@@ -181,22 +181,22 @@ const VolumePageContent = (props) => {
     {
       selected: isOverviewPage,
       title: intl.translate('overview'),
-      onClick: () => history.push(`${match.url}/overview${query.toString() && `?${query.toString()}`}`),
+      onClick: () => history.replace(`${match.url}/overview${query.toString() && `?${query.toString()}`}`),
```
ChengYanJin: We should keep using `history.push`

> The replace method is similar to push, but instead of adding a new location, it will replace the location at the current index. 

which we will lose the current location in history.

Same for all the tabs after.

ui/src/components/VolumeListTable.js
```diff
@@ -163,7 +163,7 @@ function GlobalFilter({
     } else {
       searchParams.set('search', value);
     }
-    history.push(`?${searchParams.toString()}`);
+    history.replace(`?${searchParams.toString()}`);
```
ChengYanJin: Here we should keep the `history.push` too ;)
Otherwise we will lose the history of search

ChengYanJin (changes requested):
IMO, we should only change to `history.repalce` in VolumePage. 
Other places we should keep `history.push`, otherwise we will lose the history.

alexis-ld:
@ChengYanJin Just updated based on your comments :) I was not sure if you wanted to have tab selection and search as part of the history. All good now.

thomasdanan (approved):
works as expected

ui/src/components/VolumeListTable.js
```diff
@@ -290,7 +290,7 @@ function Table({
         query.delete('sort');
         query.delete('desc');
       }
-      history.push(`?${query.toString()}`);
+      history.replace(`?${query.toString()}`);
```
ChengYanJin, line 293: I would say to keep `history.push` here too ;)
I guess you miss this change.

ChengYanJin (changes requested):

ui/src/components/VolumeListTable.js
alexis-ld: @ChengYanJin No, I didn't miss it :) replacing it here breaks the fix. This useEffect triggers on initial sort and leaving push makes the user clicks two time on "Go back" to go pack to the  previous page.

ChengYanJin: @alexis-ld Indeed. Sorry, I forgot it's in the useEffect. 
Giving my approval!

ChengYanJin (approved):
LGTM!

PR #2922, opened 2020-11-10
[UI] Fix volume sorting/ auto-select

### LINKED ISSUE #2920: [UI] Two sorting issues appear in volume list
**Component**: ui, volumes

**What happened**:
1. When we hit URL `/volumes`, it sometimes autoselects the **healthy volume** which appears on the second instead of the unhealthy volume. (The sorting is correct)
2. In some cases the URL automatically add `?sort=healthy` at the end. 
For example, `volumes/master-0-alertmanager/details?sort=health`
which should be `/volumes/master-0-alertmanager/details`
image

**What was expected**:
1. The first issue appear is because the calls to retrieve **volume health and status** is slower than the initialization of the volume list. So the first volume appears in the volume list at the beginning is not the unhealthy volume. 
Suggestion: We want to update the volume selection after the data is ready.
2. We should not add the default sort rule `?sort=health` in the URL 

**Steps to reproduce**

1. VolumePage.js
```javascript
  // If data has been retrieved and no volume is selected yet we select the first one
  useEffect(() => {
    if (volumeListData.length) {
      history.push({
        pathname: `/volumes/${volumeListData[0]?.name}/overview`,
        search: query.toString(),
      });
    }
  }, [volumeListData[0]?.name]);
```

2. VolumeListTable.js
```javascript
  useEffect(() => {
    // Creating a local query instance to avoid having it in the useEffect dependencies and creating infinite loops on search change
    const query = new URLSearchParams(window.location.search);
    const querySort = query.get('sort');
    const queryDesc = query.get('desc');
    if (data.length && (sorted !== querySort || desc !== queryDesc)) {
      if (sorted) {
        sorted ? query.set('sort', sorted) : query.delete('sort');
        desc ? query.set('desc', desc) : query.delete('desc');
        // Remove the default sorting `sort=health` from the query string
        if (sorted === 'health' && desc === false) {
          query.delete('sort');
          query.delete('desc');
        }
      } else if (!sorted && querySort) {
        query.delete('sort');
        query.delete('desc');
      }
      history.replace(`?${query.toString()}`);
    }
  }, [sorted, desc, history, data.length]);
```

alexis-ld:
**Component**:
ui,volumes

**Context**: 
There were some sorting issues on the Volumes list sorting (see #2920 )

- The auto-selected volume was not always the first one after health sorting
- The default sort sometime appeared in the URL

**Summary**:
Added a flag property to only auto-select when PVCs and Alerts are retrieved.
Removed URL sort when default sorted.

**Acceptance criteria**: 

The volumes list components waits for alerts data to be retrieved before auto-selecting the first volume.
The default sort does not appear in the URL anymore.

---

Closes: #2920

### DIFF
--- a/ui/src/containers/VolumePage.js
+++ b/ui/src/containers/VolumePage.js
@@ -92,7 +92,7 @@ const VolumePage = (props) => {
 
   // If data has been retrieved and no volume is selected yet we select the first one
   useEffect(() => {
-    if (volumeListData.length && !currentVolumeName) {
+    if (volumeListData[0]?.volumeAlertsRetrieved && !currentVolumeName) {
       history.replace({
         pathname: `/volumes/${volumeListData[0]?.name}/overview`,
         search: query.toString(),
--- a/ui/src/containers/VolumePage.js
+++ b/ui/src/containers/VolumePage.js
@@ -80,6 +80,12 @@ const VolumePage = (props) => {
     (state) => state.app.volumes.currentVolumeObject,
   );
   const pVList = useSelector((state) => state.app.volumes.pVList);
+
+  /*
+   ** The PVCs list is used to check when the alerts will be mapped to the corresponding volumes
+   ** in order to auto select the volume when all the data are there.
+   */
+  const pVCList = useSelector((state) => state?.app?.volumes?.pVCList);
   const alerts = useSelector((state) => state.app.alerts);
 
   const volumeStats = useSelector(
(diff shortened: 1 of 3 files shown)


ui/src/containers/VolumePage.js
```diff
@@ -92,7 +92,7 @@ const VolumePage = (props) => {
 
   // If data has been retrieved and no volume is selected yet we select the first one
   useEffect(() => {
-    if (volumeListData.length && !currentVolumeName) {
+    if (volumeListData[0]?.volumeAlertsRetrieved && !currentVolumeName) {
```
ChengYanJin:
```suggestion
    if (volumeListData.length && alerts.list.length && pVCList.length && !currentVolumeName) {
```

The `volumeAlertsRetrieved` is the flag to indicate if `PVC` and `alerts` are ready, add it for each item seem a bit redundant for me.

I would retrieve pVCList from VolumePage.  
```javascript
const pVCList = useSelector((state) => state?.app?.volumes?.pVCList);
```

You may want to add a little comment to quickly explain why we add this check/ what problem does it solve? 

Thanks.

ui/src/services/NodeVolumesUtils.js
```diff
@@ -244,6 +244,7 @@ export const getVolumeListData = createSelector(
       return {
         name: volume?.metadata?.name,
         node: volume?.spec?.nodeName,
+        volumeAlertsRetrieved :  alerts?.list && pVCList.length,
```
ChengYanJin: We can remove it ;)

ChengYanJin (changes requested):
Hello @alexis-ld ,
I see an improvement to do ;)

Thanks!

ui/src/containers/VolumePage.js
alexis-ld: Indeed :) thanks.

ChengYanJin (approved):
LGTM! 
Thanks!

thomasdanan (approved):

PR #2926, opened 2020-11-13
[UI] Nodes sorting

### LINKED ISSUE #2919: Ability to sort in MetalK8s UI Nodes table
**Component**:

'ui'

**Why this is needed**:

In the same way we can sort in Volumes table, it would be great to have the ability to sort in the Nodes table

**What should be done**:

Sorting should be available for all columns:

Name: Alphanumeric sorting, case-insensitive (based on Name String only)
Roles: Alphanumeric sorting
Health: Enum sorting: Critical, Warning, Healthy, Unknown
Status: Enum sorting: Ready, NotReady, Unknown

In terms of UI, it should be equivalent to what was done in Volumes table: https://github.com/scality/metalk8s/issues/2876

**Test plan**:

alexis-ld:
**Component**:
ui, nodes

**Context**: 
We want to be able to sort the Nodes table like the Volumes table.

**Summary**:
I used the same sorting logic (leveraging react-table's sorting abilities). 
I extracted the URL sync code to its custom hook and the Table sorting components to the Common layout file to avoid code redundancy between the two views.
Update:
- Introduce hook test libray 
- Bump React version to the latest v17.0.1

**Acceptance criteria**: 
Clicking on a column header on the Nodes table sorts the data according to what is specified in #2919 .
Loading the page with a sorting specified in the URL sorts the data accordingly.

---

Closes: #2919

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -2894,6 +2894,16 @@
         "loader-utils": "^1.2.3"
       }
     },
+    "@testing-library/react-hooks": {
+      "version": "3.4.2",
+      "resolved": "https://registry.npmjs.org/@testing-library/react-hooks/-/react-hooks-3.4.2.tgz",
+      "integrity": "sha512-RfPG0ckOzUIVeIqlOc1YztKgFW+ON8Y5xaSPbiBkfj9nMkkiLhLeBXT5icfPX65oJV/zCZu4z8EVnUc6GY9C5A==",
+      "dev": true,
+      "requires": {
+        "@babel/runtime": "^7.5.4",
+        "@types/testing-library__react-hooks": "^3.4.0"
+      }
+    },
     "@types/babel__core": {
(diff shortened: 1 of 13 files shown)


thomasdanan (dismissed):
works well for me ...

ChengYanJin (changes requested):
Smart move of `useTableSortURLSync` hook! Bravo! I suggest we add a unit test for this. 

BTW, since by default we sort the node list by health. Shouldn't we auto-select the first node base on the health? 

Thanks!

alexis-ld:
@ChengYanJin I added tests for the custom hook as well as the Node auto-select

ChengYanJin (approved):
LGTM!
Thanks for introducing the hook test library and bump React to the latest version ;)

thomasdanan (approved):
works as expected

PR #2930, opened 2020-11-16
salt/unit-test: Migrate to Python 3

gdemonet:
- removes `salttesting` dependency
- migrates away from deprecated `TestCase` methods
- migrates `Mock` objects call args checks


TeddyAndrieux (approved):
Green build so it's ok for me

PR #2936, opened 2020-11-19
Sparse loop device provisioning with systemd units

### LINKED ISSUE #2726: Pod failed to mount SparseLoopDevice Volume after node reboot

gdemonet:
**Component**: salt, storage

**Context**:

The previous approach was relying entirely on Salt (through our custom
`metalk8s_volumes` execution/state module) to manage provisioning and
cleanup of loop devices (on sparse files).
The issues arise when (re)booting a node:
- Salt minion has to execute a startup state, which may come in when
kube-apiserver is not (yet) available
- `losetup --find` calls, with the version available for CentOS 7, are
not atomic, so we may lose some devices if we have too many (re-running
the state manually is the current workaround)

**Summary**:

To fix these, we introduce two main changes:
- management of loop devices is delegated to systemd; using a unit
template file, we define a service per volume (passing the Volume's UUID
as a parameter), which will provision and attach the device on start,
and detach it on stop
- `losetup --find` invocations from these units is made sequential by
using an exclusive lockfile (this is not perfect, but avoids the need to
re-implement the command ourselves to include the fix from higher
versions of `losetup`)

Note that we cannot simply use `losetup --detach` in the unit template,
because sparse volumes may not always be discovered from the same path:
- either the volume is formatted, and we can find it in
/dev/disk/by-uuid/
- or it's kept raw, so we only have the UUID in the partition table, and
we can discover the device through /dev/disk/by-partuuid/

**Acceptance criteria**:

- CI keeps working (provisioning and cleanup of loop devices is tested)
- (manual) can reboot a node with many sparse loop volumes (10+) and have them all in `losetup -l` after

---

Closes: #2726

### DIFF
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -651,7 +651,8 @@ SALT_FILES : Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path('salt/metalk8s/volumes/init.sls'),
     Path('salt/metalk8s/volumes/prepared/init.sls'),
     Path('salt/metalk8s/volumes/prepared/installed.sls'),
-    Path('salt/metalk8s/volumes/provisioned/init.sls'),
+    Path('salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service'),
+    Path('salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py'),
     Path('salt/metalk8s/volumes/unprepared/init.sls'),
 
     Path('salt/_auth/kubernetes_rbac.py'),
--- a/salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
+++ b/salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
@@ -10,8 +10,3 @@ use_superseded:
 log_level_logfile: {{ 'debug' if debug else 'info' }}
 
 saltenv: {{ saltenv }}
-
-# Prepare volume at startup (required for loop devices persistency).
-startup_states: sls
-sls_list:
-  - /metalk8s/volumes/provisioned
--- a/salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
+++ b/salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Setup MetalK8s sparse loop volume %I
+RequiresMountsFor=/var/lib/metalk8s/storage/sparse
+AssertFileNotEmpty=/var/lib/metalk8s/storage/sparse/%i
+
+[Service]
+Type=oneshot
+ExecStart=/bin/flock --exclusive --wait 10 /var/lock/metalk8s-sparse-volume.lock \
+             /sbin/losetup --find --partscan "/var/lib/metalk8s/storage/sparse/%i"
+ExecStop=/usr/local/libexec/metalk8s-sparse-volume-cleanup "%i"
+TimeoutStartSec=10
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
--- a/salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
+++ b/salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
@@ -1,0 +1,183 @@
+#!/usr/bin/env python2
+"""Script for cleaning up MetalK8s sparse loop volumes.
+
+Handle discovery of the device through a Volume's UUID, and then detach the
+device using the right `ioctl`.
+"""
+
+from __future__ import print_function
+import argparse
+import errno
+import fcntl
+import os
+import stat
+import sys
+
+
+def parse_args(args=None):
+    """Parse command-line arguments."""
+    parser = argparse.ArgumentParser(
+        prog='metalk8s-sparse-volume-cleanup',
+        description='Cleanup MetalK8s sparse loop volumes',
+    )
+
+    parser.add_argument(
+        'volume_id',
+        help='UUID of the MetalK8s Volume object',
+    )
+
+    return parser.parse_args(args=args)
+
+
+SPARSE_FILES_DIR = '/var/lib/metalk8s/storage/sparse'
+
+
+def check_sparse_file(volume_id):
+    """Check if the sparse file exists for this volume ID."""
+    sparse_file_path = os.path.join(SPARSE_FILES_DIR, volume_id)
+    if not os.path.isfile(sparse_file_path):
+        raise Error(
+            "Sparse file {} does not exist for volume '{}'".format(
+                sparse_file_path, volume_id
+            ),
+            exit_code=errno.ENOENT
+        )
+
+
+def find_device(volume_id):
+    """Find the device provisioned for this volume ID.
+
+    SparseLoopDevice volumes are either:
+    - formatted, with the volume ID set in the file-system annotations
+    - raw, in which case the device is partitioned and the first partition
+      is labeled with the volume ID
+    """
+    for _get_device in [device_by_uuid, device_from_part_uuid]:
+        device_path = _get_device(volume_id)
+        if device_path is not None:
+            break
+    else:
+        raise Error(
+            "Device for volume '{}' was not found".format(volume_id),
+            exit_code=errno.ENOENT,
+        )
+
+    if not is_loop_device(device_path):
+        raise Error(
+            "{} (found for volume '{}') is not a loop device".format(
+                device_path, volume_id
+            ),
+            exit_code=errno.EINVAL,
+        )
+
+    return device_path
+
+
+# https://github.com/torvalds/linux/blob/master/include/uapi/linux/loop.h#L110
+LOOP_CLR_FD = 0x4C01
+
+
+def cleanup(volume_id):
+    """Clean up a sparse volume's loop device.
+
+    This function will bail out if a corresponding loop device cannot be
+    found.
+    """
+    device_path = find_device(volume_id)
+
+    print("Detaching loop device '{}' for volume '{}'".format(
+        device_path, volume_id
+    ))
+
+    device_handle = os.open(device_path, os.O_RDONLY)
+    try:
+        fcntl.ioctl(device_handle, LOOP_CLR_FD, 0)
+    except IOError as exn:
+        if exn.errno != errno.ENXIO:
+            raise Error(
+                "Unexpected error when trying to free device {}: {}".format(
+                    device_path, str(exn)
+                ),
+                exit_code=exn.errno,
+            )
+        print("Device already freed")
+    finally:
+        os.close(device_handle)
+
+    print("Loop device for volume '{}' was successfully detached".format(
+        volume_id
+    ))
+
+
+def main():
+    """Main "routine" of this script."""
+    args = parse_args()
+
+    check_sparse_file(args.volume_id)
+    cleanup(args.volume_id)
+
+
+# Helpers {{{
+# Error handling {{{
+
+def die(message, exit_code=1):
+    """Print a message to the standard error stream, and exit."""
+    print(message, file=sys.stderr)
+    sys.stderr.flush()
+    sys.exit(exit_code)
+
+
+class Error(Exception):
+    """Base-class for errors raised by methods from this module."""
+    def __init__(self, message, *args, exit_code=1):
+        super(Error, self).__init__(self, message, *args)
+        self.message = message
+        self.exit_code = exit_code
+
+
+# }}}
+# Loop device discovery {{{
+
+def device_by_uuid(volume_id):
+    """Find a device by file-system UUID."""
+    path_by_uuid = os.path.join("/dev/disk/by-uuid", volume_id)
+    if not os.path.exists(path_by_uuid):
+        return None
+
+    return os.path.realpath(path_by_uuid)
+
+
+def device_from_part_uuid(volume_id):
+    """Find a device from its first partition UUID."""
+    path_by_partuuid = os.path.join("/dev/disk/by-partuuid", volume_id)
+    if not os.path.exists(path_by_partuuid):
+        return None
+
+    partition_name = os.path.basename(os.path.realpath(path_by_partuuid))
+    device_link = os.path.join("/sys/class/block", partition_name, os.pardir)
+    device_name = os.path.basename(os.path.realpath(device_link))
+    return os.path.join("/dev", device_name)
+
+
+# }}}
+
+# https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/devices.txt#L191
+LOOP_MAJOR = 7
+
+
+def is_loop_device(path):
+    """Check if the provided path is a real loop device."""
+    device_stat = os.stat(path)
+    return stat.S_ISBLK(device_stat.st_mode) \
+        and (os.major(device_stat.st_rdev) == LOOP_MAJOR)
+
+
+# }}}
+
+if __name__ == '__main__':
+    try:
+        main()
+    except Error as exc:
+        die(exc.message, exit_code=exc.exit_code)
+    except Exception as exc:  # pylint: disable=broad-except
+        die('Unhandled exception when cleaning up volume: {!s}'.format(exc))
--- a/salt/metalk8s/volumes/prepared/init.sls
+++ b/salt/metalk8s/volumes/prepared/init.sls
@@ -3,22 +3,23 @@ include:
 
 {%- set volumes_to_create = [] %}
 {%- set all_volumes = pillar.metalk8s.volumes %}
-{%- set target_volume = pillar.get('volume') %}
+{%- set target_volume_name = pillar.get('volume') %}
 
 {#- If a volume is given we only take this one, otherwise we take them all. #}
-{%- if target_volume is not none %}
-  {%- if target_volume in all_volumes.keys() %}
+{%- if target_volume_name is not none %}
+  {%- set target_volume = all_volumes.get(target_volume_name) %}
+  {%- if target_volume is not none %}
     {%- do volumes_to_create.append(target_volume) %}
   {%- else %}
-Volume {{ target_volume }} not found in pillar:
+Volume {{ target_volume_name }} not found in pillar:
   test.configurable_test_state:
-    - name: {{ target_volume }}
+    - name: {{ target_volume_name }}
     - changes: False
     - result: False
-    - comment: Volume {{ target_volume }} not found in pillar
+    - comment: Volume {{ target_volume_name }} not found in pillar
   {%- endif %}
 {%- else %}
-    {%- do volumes_to_create.extend(all_volumes.keys()|list) %}
+    {%- do volumes_to_create.extend(all_volumes.values()|list) %}
 {%- endif %}
 
 Create the sparse file directory:
@@ -26,28 +27,50 @@ Create the sparse file directory:
     - name: /var/lib/metalk8s/storage/sparse/
     - makedirs: True
 
-{%- for volume in volumes_to_create %}
+Set up systemd template unit for sparse loop device provisioning:
+  file.managed:
+    - name: /etc/systemd/system/metalk8s-sparse-volume@.service
+    - source: salt://{{ slspath }}/files/metalk8s-sparse-volume@.service
+    - user: root
+    - group : root
+    - mode: 644
 
-Create backing storage for {{ volume }}:
+Install clean-up script:
+  file.managed:
+    - name: /usr/local/libexec/metalk8s-sparse-volume-cleanup
+    - source: salt://{{ slspath }}/files/sparse_volume_cleanup.py
+    - user: root
+    - group : root
+    - mode: 755
+
+{%- for volume in volumes_to_create %}
+  {%- set volume_name = volume.metadata.name %}
+
+Create backing storage for {{ volume_name }}:
   metalk8s_volumes.present:
-    - name: {{ volume }}
+    - name: {{ volume_name }}
     - require:
       - file: Create the sparse file directory
 
-Prepare backing storage for {{ volume }}:
+Prepare backing storage for {{ volume_name }}:
   metalk8s_volumes.prepared:
-    - name: {{ volume }}
+    - name: {{ volume_name }}
     - require:
       - metalk8s_package_manager: Install e2fsprogs
       - metalk8s_package_manager: Install xfsprogs
       - metalk8s_package_manager: Install gdisk
-      - metalk8s_volumes: Create backing storage for {{ volume }}
+      - metalk8s_volumes: Create backing storage for {{ volume_name }}
+  {%- if 'sparseLoopDevice' in volume.spec %}
 
-Provision backing storage for {{ volume }}:
-  metalk8s_volumes.provisioned:
-    - name: {{ volume }}
+Provision backing storage for {{ volume_name }}:
+  service.running:
+    - name: metalk8s-sparse-volume@{{ volume.metadata.uid }}
+    - enable: true
     - require:
-      - metalk8s_volumes: Prepare backing storage for {{ volume }}
+      - metalk8s_volumes: Prepare backing storage for {{ volume_name }}
+      - file: Set up systemd template unit for sparse loop device provisioning
+      - test: Ensure Python 2 is available
+  {%- endif %}
     - require_in:
       - module: Update pillar after volume provisioning
 
--- a/salt/metalk8s/volumes/prepared/installed.sls
+++ b/salt/metalk8s/volumes/prepared/installed.sls
@@ -17,3 +17,9 @@ Install gdisk:
   {{ pkg_installed('gdisk') }}
     - require:
       - test: Repositories configured
+
+# Needed by the sparse volume cleanup script
+Ensure Python 2 is available:
+  test.fail_without_changes:
+    - comment: Could not find a working Python 2 installation
+    - unless: /usr/bin/env python2 --version
--- a/salt/metalk8s/volumes/unprepared/init.sls
+++ b/salt/metalk8s/volumes/unprepared/init.sls
@@ -1,19 +1,28 @@
 {%- set volumes = pillar.metalk8s.volumes %}
-{%- set volume  = pillar.get('volume', '') %}
+{%- set volume_name  = pillar.get('volume', '') %}
+{%- set to_remove = volumes.get(volume_name) %}
 
+{%- if to_remove is not none %}
+  {%- if 'sparseLoopDevice' in to_remove.spec %}
+Disable systemd provisioning of loop device for {{ volume_name }}:
+  service.dead:
+    - name: metalk8s-sparse-volume@{{ to_remove.metadata.uid }}
+    - enable: false
+    - require_in:
+      - metalk8s_volumes: Clean up backing storage for {{ volume_name }}
 
-{%- if volume in volumes.keys() %}
-Clean up backing storage for {{ volume }}:
+  {%- endif %}
+Clean up backing storage for {{ volume_name }}:
   metalk8s_volumes.removed:
-    - name: {{ volume }}
+    - name: {{ volume_name }}
 {%- else %}
 
-{%- do salt.log.warning('Volume ' ~ volume ~ ' not found in pillar') -%}
+{%- do salt.log.warning('Volume ' ~ volume_name ~ ' not found in pillar') -%}
 
-Volume {{ volume }} not found in pillar:
+Volume {{ volume_name }} not found in pillar:
   test.configurable_test_state:
-    - name: {{ volume }}
+    - name: {{ volume_name }}
     - changes: False
     - result: True
-    - comment: Volume {{ volume }} not found in pillar
+    - comment: Volume {{ volume_name }} not found in pillar
 {%- endif %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   more robust about static pod restart and improve user experience
   (PR [#2928](https://github.com/scality/metalk8s/pull/2928))
 
+- [#2726](https://github.com/scality/metalk8s/issues/2726) - Ensure sparse loop
+  volumes are all provisioned on reboot
+  (PR [#2936](https://github.com/scality/metalk8s/pull/2936))
+
 ## Release 2.6.1 (in development)
 
 ## Release 2.6.0
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -62,39 +62,6 @@ def create(name):
     _get_volume(name).create()
 
 
-def is_provisioned(name):
-    """Check if the backing storage device is provisioned for the given volume.
-
-    Args:
-        name (str): volume name
-
-    Returns:
-        bool: True if the backing storage device is provisioned, otherwise False
-
-    CLI Example:
-
-    .. code-block:: bash
-
-        salt '<NODE_NAME>' metalk8s_volumes.is_provisioned example-volume
-    """
-    return _get_volume(name).is_provisioned
-
-
-def provision(name):
-    """Provision the backing storage device of the given volume.
-
-    Args:
-        name (str): volume name
-
-    CLI Example:
-
-    .. code-block:: bash
-
-        salt '<NODE_NAME>' metalk8s_volumes.provision example-volume
-    """
-    _get_volume(name).provision()
-
-
 def is_prepared(name):
     """Check if the given volume is prepared.
 
@@ -234,16 +201,6 @@ class Volume():
         return
 
     @abc.abstractproperty
-    def is_provisioned(self):  # pragma: no cover
-        """Check if the backing storage device is provisioned."""
-        return
-
-    @abc.abstractmethod
-    def provision(self):  # pragma: no cover
-        """Provision the backing storage device."""
-        return
-
-    @abc.abstractproperty
     def is_cleaned_up(self):  # pragma: no cover
         """Check if the backing storage device is cleaned up."""
         return
@@ -313,10 +270,6 @@ class Volume():
 
 
 class SparseLoopDevice(Volume):
-    # Recent losetup support `--nooverlap` but not the one shipped with
-    # CentOS 7.
-    PROVISIONING_COMMAND = ('losetup', '--find')
-
     @property
     def path(self):
         return '/var/lib/metalk8s/storage/sparse/{}'.format(self.uuid)
@@ -347,32 +300,16 @@ class SparseLoopDevice(Volume):
                 self.path, exn
             ))
 
-    @property
-    def is_provisioned(self):
-        # A sparse loop device is provisioned when a sparse file is associated
-        # to a loop device.
-        command = ' '.join(['losetup', '--associated', self.path])
-        pattern = r'\({}\)'.format(re.escape(self.path))
-        result  = _run_cmd(command)
-        return re.search(pattern, result['stdout']) is not None
-
-    def provision(self):
-        command = ' '.join(list(self.PROVISIONING_COMMAND) + [self.path])
-        return _run_cmd(command)
-
     def prepare(self, force=False):
         # We format a "normal" file, not a block device: we need force=True.
         super(SparseLoopDevice, self).prepare(force=True)
 
     @property
     def is_cleaned_up(self):
-        return not (self.is_provisioned or self.exists)
+        return not self.exists
 
     def clean_up(self):
-        LOOP_CLR_FD = 0x4C01  # From /usr/include/linux/loop.h
         try:
-            with _open_fd(self.persistent_path, os.O_RDONLY) as fd:
-                fcntl.ioctl(fd, LOOP_CLR_FD, 0)
             os.remove(self.path)
         except OSError as exn:
             if exn.errno != errno.ENOENT:
@@ -381,10 +318,6 @@ class SparseLoopDevice(Volume):
 
 
 class SparseLoopDeviceBlock(SparseLoopDevice):
-    # Recent losetup support `--nooverlap` but not the one shipped with
-    # CentOS 7.
-    PROVISIONING_COMMAND = ('losetup', '--find', '--partscan')
-
     @property
     def persistent_path(self):
         return '/dev/disk/by-partuuid/{}'.format(self.uuid)
@@ -419,13 +352,6 @@ class RawBlockDevice(Volume):
         raise Exception('block device {} does not exists'.format(
             self.path
         ))
-
-    @property
-    def is_provisioned(self):
-        return True  # Nothing to do so it's always True.
-
-    def provision(self):
-        return  # Nothing to do
 
     @property
     def path(self):
--- a/salt/_states/metalk8s_volumes.py
+++ b/salt/_states/metalk8s_volumes.py
@@ -51,43 +51,6 @@ def present(name):
     return ret
 
 
-def provisioned(name):
-    """Provision the given volume.
-
-    Args:
-        name (str): Volume name
-
-    Returns:
-        dict: state return value
-    """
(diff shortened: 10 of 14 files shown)


buildchain/buildchain/salt_tree.py
```diff
@@ -651,7 +651,8 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/volumes/init.sls'),
     Path('salt/metalk8s/volumes/prepared/init.sls'),
     Path('salt/metalk8s/volumes/prepared/installed.sls'),
-    Path('salt/metalk8s/volumes/provisioned/init.sls'),
+    Path('salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service'),
+    Path('salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up'),
```
NicolasT: I'd say `cleanup` but OK ;)

salt/metalk8s/salt/minion/files/minion-99-metalk8s.conf.j2
```diff
@@ -10,8 +10,3 @@ use_superseded:
 log_level_logfile: {{ 'debug' if debug else 'info' }}
 
 saltenv: {{ saltenv }}
-
-# Prepare volume at startup (required for loop devices persistency).
-startup_states: sls
-sls_list:
-  - /metalk8s/volumes/provisioned
```
NicolasT, line 17: :heart:

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Setup MetalK8s sparse loop device %I
+RequiresMountsFor=/var/lib/metalk8s/storage/sparse
```
NicolasT, line 3: We could add an `AssertFileNotEmpty=/var/lib/metalk8s/storage/sparse/%i` here as well, see https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Conditions%20and%20Asserts (I've been reading too many `systemd` man-pages recently...)

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,13 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/losetup --find --partscan "/var/lib/metalk8s/storage/sparse/%i"
+ExecStop=/usr/local/bin/metalk8s-volume-clean-up "%i"
```
NicolasT: I really think this should go in `/usr/local/bin/libexec`, not `/usr/local/bin`. It's not some executable supposed to be regularly executed by a system user.

Also, `cleanup` :wink:

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Setup MetalK8s sparse loop device %I
```
NicolasT:
```suggestion
Description=Setup MetalK8s sparse loop volume %I
```

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+    if len(sys.argv) != 2:
+        die("Invalid number of arguments: expected only one volume ID")
+
+    return sys.argv[1]
+
+
+def find_device(volume_id):
+    path_by_uuid = "/dev/disk/by-uuid/{}".format(volume_id)
+    if os.path.exists(path_by_uuid):
+        return path_by_uuid
```
NicolasT: Could actually read the symlink here to always return `/dev/loopN` from this function.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+    path_by_partuuid = "/dev/disk/by-partuuid/{}".format(volume_id)
+    if not os.path.exists(path_by_partuuid):
+        die("Volume '{}' backing device was not found".format(volume_id))
+
+    partition_name = os.path.basename(os.path.realpath(path_by_partuuid))
+    device_name = os.path.basename(
+        os.path.realpath("/sys/class/block/{}/..".format(partition_name))
+    )
+
+    return "/dev/{}".format(device_name)
```
NicolasT: Add an assertion to check this actually exists? It always should, but shouldn't harm...

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+
+
+def clean_up(device_path):
+    try:
+        subprocess.check_output(
+            ["losetup", "-d", device_path],
```
NicolasT: The original implementation in `salt/_modules/metalk8s_volumes.py` did the `losetup -d` in pure Python code. It could make sense to do the same instead of shelling out.

`losetup -d` implementation basically boils down to https://github.com/karelzak/util-linux/blob/a61955ba1f15bf0a252a03a9abe54c1e02828c2e/lib/loopdev.c#L1518

NicolasT: Note that, *iff* we need to have a wrapper for the loop device creation (locking...) as well, we could consider to (re-)implement this in pure Python as well.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+            stderr=subprocess.STDOUT,
+        )
+    except subprocess.CalledProcessError as exc:
+        die((
+            "Failed to remove device: "
+            "{} failed with exit code {} and output: {}"
+        ).format(exc.cmd, exc.returncode, exc.output))
+
+
+if __name__ == '__main__':
```
NicolasT:
```suggestion
def main():
```
and
```
if __name__ == '__main__':
    main()
```

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+    volume_id = get_volume_id()
+    device_path = find_device(volume_id)
+
+    print("Cleaning up backing device '{}' for volume '{}'".format(
+        device_path, volume_id
+    ))
+    clean_up(device_path)
+
+    print("Volume '{}' was successfully cleaned up".format(volume_id))
+    sys.exit(0)
```
NicolasT: Not needed.

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -3,51 +3,78 @@ include:
...
     - makedirs: True
 
+Set up systemd template unit for sparse loop device provisioning:
+  file.managed:
+    - name: /etc/systemd/system/metalk8s-sparse-volume@.service
+    - source: salt://{{ slspath }}/files/metalk8s-sparse-volume@.service
+    - user: root
+    - group : root
+    - mode: 644
+    - makedirs: True
```
NicolasT: No. If `/etc/systemd/system` doesn't exist then something is *really* wrong on the system and we shouldn't ignore that.

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -3,51 +3,78 @@ include:
...
+    - dir_mode: 755
+
+Install clean-up script:
+  file.managed:
+    - name: /usr/local/bin/metalk8s-volume-clean-up
```
NicolasT: `/usr/local/libexec`

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -3,51 +3,78 @@ include:
...
+    - source: salt://{{ slspath }}/files/metalk8s-volume-clean-up
+    - user: root
+    - group : root
+    - mode: 755
+    - makedirs: True
```
NicolasT: Meh. This should really exist on the system already.

salt/metalk8s/volumes/prepared/init.sls
```diff
@@ -3,51 +3,78 @@ include:
...
-    - name: {{ volume }}
+Provision backing storage for {{ volume_name }}:
+  service.running:
+    - name: metalk8s-sparse-volume@{{ volume.metadata.uid }}
+    - enable: true
     - require:
-      - metalk8s_volumes: Prepare backing storage for {{ volume }}
+      - metalk8s_volumes: Prepare backing storage for {{ volume_name }}
+    - watch:
+      - file: Set up systemd template unit for sparse loop device provisioning
```
NicolasT, line 71: Shouldn't this be a `require` instead of a `watch`? If it needs to be a `watch`, why exactly? We definitely don't want to 'restart' the service if for whatever reason we update the service template, because restarting will likely not be possible (since the loop device is in use by some K8s container).

NicolasT (changes requested):

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,13 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/losetup --find --partscan "/var/lib/metalk8s/storage/sparse/%i"
+ExecStop=/usr/local/bin/metalk8s-volume-clean-up "%i"
+TimeoutSec=60
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
```
slaperche-scality: missing newline at the end of the file (as GitHub pointed out ^^)

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
```
slaperche-scality: We should specified 2 or 3 (seeing the import future I guess it's 2, but better be explicit).

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+def get_volume_id():
+    if len(sys.argv) != 2:
+        die("Invalid number of arguments: expected only one volume ID")
+
+    return sys.argv[1]
+
+
+def find_device(volume_id):
+    path_by_uuid = "/dev/disk/by-uuid/{}".format(volume_id)
+    if os.path.exists(path_by_uuid):
```
slaperche-scality: Could even use `os.path.islink` (or Pathlib, but since it's Python 2, not really).

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+        return path_by_uuid
+
+    path_by_partuuid = "/dev/disk/by-partuuid/{}".format(volume_id)
+    if not os.path.exists(path_by_partuuid):
```
slaperche-scality: ditto

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
```diff
@@ -0,0 +1,61 @@
...
+        die("Volume '{}' backing device was not found".format(volume_id))
+
+    partition_name = os.path.basename(os.path.realpath(path_by_partuuid))
+    device_name = os.path.basename(
+        os.path.realpath("/sys/class/block/{}/..".format(partition_name))
+    )
```
slaperche-scality: I wonder if using regex to extract `loop[0-9]+` isn't simpler here, but maybe that way is more reliable.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Setup MetalK8s sparse loop device %I
+RequiresMountsFor=/var/lib/metalk8s/storage/sparse
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/losetup --find --partscan "/var/lib/metalk8s/storage/sparse/%i"
+ExecStop=/usr/local/bin/metalk8s-volume-clean-up "%i"
+TimeoutSec=60
```
NicolasT: Actually not sure we want/need such timeout.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
NicolasT: It's a link because that's how `udev` generally sets things up. However, it *could* (theoretically) be a proper block device. Why rule that out?

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
NicolasT: A regex would make assumptions about the naming, which is not guaranteed. Here all we assume is how sysfs is structured, which is kernel API (I believe...).

We *could* assert the final path we calculate is an actual block device (from `stat`), and a loop device (also from `stat`, though not sure how to figure out the major of `loop` except by parsing `/proc/devices`).

NicolasT: (Easy using `awk`: `awk '$2=="loop" {print $1}' /proc/devices`, and likely `7` on all systems, not dynamically registered, but still, hard-coding isn't great).

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
slaperche-scality: In that case, better keep the `os.path.exists` yeah.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
NicolasT: And somewhere make sure it's installed on the system. Sure, we can assume it is (we're running `salt-minion` after all), but still...

salt/metalk8s/volumes/prepared/init.sls
gdemonet: I thought `losetup --find` was idempotent. Using `watch` was indeed to reload the service in case the template changed (e.g. fixing a bug somewhere in that chain), but maybe we shouldn't do this at all.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Let's go for pure Python! Makes me think I have forgotten to remove the `losetup -d` from the `clean_up` method, in our module.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
TeddyAndrieux: > It's not some executable supposed to be regularly executed by a system user.

`/usr/local/sbin` then no ?

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
TeddyAndrieux: In metalk8s-2.7 we will likely go with Salt Python3 so we could likely assume we run with Python3 (but since it's not merged yet we may need to stick with Python2)

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: I do prefer the regex as well. It's not like we're trying to make this logic work for "any kind of device partition", we **know** it is supposed to be a loop device, so if we don't find `loop[0-9]+`, it's a good idea to bail out (if some other OS we support decides to break this behaviour, our code will explicitly tell us).

salt/metalk8s/volumes/prepared/init.sls
NicolasT: It's not:
```
[root@containerd ~]# losetup -l
[root@containerd ~]# losetup --find /root/disk
[root@containerd ~]# losetup -l
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE
/dev/loop0         0      0         0  0 /root/disk
[root@containerd ~]# losetup --find /root/disk
[root@containerd ~]# losetup -l
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE
/dev/loop0         0      0         0  0 /root/disk
/dev/loop1         0      0         0  0 /root/disk
[root@containerd ~]# losetup --find /root/disk
[root@containerd ~]# losetup -l
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE
/dev/loop0         0      0         0  0 /root/disk
/dev/loop1         0      0         0  0 /root/disk
/dev/loop2         0      0         0  0 /root/disk
```

Furthermore, a restart of the service would imply an `losetup -d` of the existing volume (unit is stopped), which we don't want to happen.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: OK, indeed it's almost an instantaneous operation.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
NicolasT: Nope, `sbin` is for executables to be used by a superuser. This script is only ever to be executed by `systemd`, not by any regular/interactive user, hence `libexec`.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
NicolasT: I do agree we want to assert it's a loop-device. However, we should not do so based on some name, but based on the major of the block-device matching the major assigned to the `loop` block-device driver.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Removed.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Added a check to verify that:
- the device path exists
- it's a block device
- its major is the "loop" major (set to 7, hardcoded, sorry @NicolasT)

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Kept `os.path.exists`, we are going to check what we find anyway.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Kept it.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Sticking with Python 2 for now, not exactly sure where we should check if it's installed though. In the service template? Or through Salt?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Done.

salt/metalk8s/volumes/prepared/init.sls
gdemonet: Used a `require` instead.

salt/metalk8s/volumes/prepared/init.sls
gdemonet: Removed.

salt/metalk8s/volumes/prepared/init.sls
gdemonet: Done.

salt/metalk8s/volumes/prepared/init.sls
gdemonet: Removed.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Removed.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Done.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Done (and removed the corresponding code from our custom module).

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Done (and also checking that this is a loop device)

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
gdemonet: Done.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Done.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Done.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Added.

buildchain/buildchain/salt_tree.py
gdemonet: Done.

salt/metalk8s/volumes/prepared/files/metalk8s-volume-clean-up
NicolasT: Salt. This is where maybe at some point these files shouldn't be dropped as-is by Salt, but packaged properly, which then allows to add a dependency on some Python version.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+from __future__ import print_function
+import argparse
+import errno
+import os
+import stat
+import sys
+
+
+SPARSE_FILES_DIR = '/var/lib/metalk8s/storage/sparse/'
+LOOP_MAJOR = 7
```
slaperche-scality: You could cite [Documentation/admin-guide/devices.txt](https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/devices.txt#L191) from linux source tree as reference.

e.g.:
```python
# See Documentation/admin-guide/devices.txt in linux source tree
 LOOP_MAJOR = 7 
```

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+            device_path, volume_id
+        ))
+
+    return device_path
+
+
+def is_loop_device(path):
+    device_stat = os.stat(path)
+    return stat.S_ISBLK(device_stat.st_mode) \
+        and (_major(device_stat.st_rdev) == LOOP_MAJOR)
```
slaperche-scality: Could use [`os.major(device_stat.st_rdev)`](https://docs.python.org/2.7/library/os.html#os.major) instead of coding your own `_major` :wink:

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+LOOP_MAJOR = 7
+
+
+def parse_args(args=None):
+    parser = argparse.ArgumentParser(
+        prog='metalk8s-sparse-volume-cleanup',
+        help='Cleanup MetalK8s sparse loop volumes',
+    )
+
+    parser.add_argument(
```
NicolasT: If this is required, why use a flag argument instead of a positional one?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+        '-i', '--uid',
```
NicolasT: It's not `uid`, but `uuid` (though again, not sure this should be a flag).

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+        required=True,
+        help='UID of the MetalK8s Volume object',
```
NicolasT: `UUID`

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+
+from __future__ import print_function
+import argparse
+import errno
+import os
+import stat
+import sys
+
+
+SPARSE_FILES_DIR = '/var/lib/metalk8s/storage/sparse/'
```
NicolasT: Can drop the last `/` (see below).

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+        help='UID of the MetalK8s Volume object',
+        dest='volume_id',
+    )
+
+    return parser.parse_args(args=args)
+
+
+def check_sparse_file(volume_id):
+    """Check if the sparse file exists for this volume ID."""
+    sparse_file_path = SPARSE_FILES_DIR + volume_id
```
NicolasT: `os.path.join` iso `+`

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+    except IOError as exn:
+        if exn.errno != errno.ENXIO:
+            die("Unexpected error when trying to free device {}: {}".format(
+                device_path, str(exn)
+            ))
+        print("Device already freed", file=sys.stderr)
+    finally:
+        fd.close()
+
+    print("Loop device for volume '{}' was successfully detached".format(
```
NicolasT: The `Device already freed` message would go to `stderr`. Is that consistent?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+def main():
+    args = parse_args()
+
+    check_sparse_file(args.volume_id)
+    cleanup(args.volume_id)
+
+
+# Helpers {{{
+def die(message, exit_code=1):
+    print(message, file=sys.stderr)
```
NicolasT: Does `sys.stderr` need a `flush` before `sys.exit`?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+    """Find the device provisioned for this volume ID.
+    
+    SparseLoopDevice volumes are either:
+    - formatted, with the volume ID set in the filesystem annotations,
+      hence discoverable under /dev/disk/by-uuid/
+    - raw, in which case the device is partitioned and the first partition
+      is labeled with the volume ID, hence discoverable under
+      /dev/disk/by-partuuid/
+    """
+    path_by_uuid = "/dev/disk/by-uuid/{}".format(volume_id)
```
NicolasT: `os.path.join`.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+    if os.path.exists(path_by_uuid):
+        device_path = os.path.realpath(path_by_uuid)
+    else:
+        path_by_partuuid = "/dev/disk/by-partuuid/{}".format(volume_id)
```
NicolasT: `os.path.join`

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+        if not os.path.exists(path_by_partuuid):
```
NicolasT: Now if we'd add a third type of device (discovery), this would become one more indented block, and so on. I'd put all 'discovery' code (per case) in a separate function, then call them one by one until one returns a non-`None` result (or doesn't raise, whatever), then fail if none of them return a 'good' value: cleaner code, better scoped, less nesting and easier to test.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+            die(
+                "Device for volume '{}' was not found".format(volume_id),
+                exit_code=errno.ENOENT,
+            )
+
+        partition_name = os.path.basename(os.path.realpath(path_by_partuuid))
+        device_name = os.path.basename(
+            os.path.realpath("/sys/class/block/{}/..".format(partition_name))
```
NicolasT: `os.path.join` and `os.path.pardir`

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
```diff
@@ -0,0 +1,128 @@
...
+        )
+        device_path = '/dev/{}'.format(match.group(0))
```
NicolasT: I may be missing something, but where does `match` come from? May need to add this file to the linted Python files in buildchain?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
NicolasT: Why not `os.major`?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Setup MetalK8s sparse loop volume %I
+RequiresMountsFor=/var/lib/metalk8s/storage/sparse
+AssertFileNotEmpty=/var/lib/metalk8s/storage/sparse/%i
+
+[Service]
+Type=oneshot
+ExecStart=/bin/flock --exclusive --wait 10 /var/lock/metalk8s-sparse-volume.lock \
+             -c "/sbin/losetup --find --partscan /var/lib/metalk8s/storage/sparse/%i"
```
NicolasT: Could not have `-c "..."`, and use the `sudo`-style invocation, no?

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
```diff
@@ -0,0 +1,14 @@
...
+ExecStop=/usr/local/bin/libexec/metalk8s-sparse-volume-cleanup --uid "%i"
```
NicolasT: `/usr/local/bin/libexec` ain't right.

salt/metalk8s/volumes/unprepared/init.sls
```diff
@@ -1,19 +1,27 @@
 {%- set volumes = pillar.metalk8s.volumes %}
-{%- set volume  = pillar.get('volume', '') %}
+{%- set volume_name  = pillar.get('volume', '') %}
+{%- set to_remove = volumes.get(volume_name) %}
 
+{%- if to_remove is not none %}
+  {%- if 'sparseLoopDevice' in to_remove.spec %}
+Disable systemd provisioning of loop device for {{ volume_name }}:
+  service.disabled:
```
NicolasT: Will this also 'stop' the service?

NicolasT (changes requested):
Some issues

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
slaperche-scality: `stderr` is usually unbuffered, but [it's not that clear in Python](https://stackoverflow.com/a/22478875) so we may want to flush to be safe (especially since the behavior is different between python 2 and 3, as pointed [here](https://bugs.python.org/msg149408)).

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service
gdemonet: Whoops.

salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume-cleanup
gdemonet: Sorry, remnants of the temporary implementation where I had an `action` (`provision` or `cleanup`) as positional. Anyway, will change.

salt/metalk8s/volumes/unprepared/init.sls
gdemonet: Good catch. I didn't realize it wasn't stopped, because I left the cleanup logic in the module :confused:

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
```diff
@@ -0,0 +1,158 @@
...
+import sys
+
+
+SPARSE_FILES_DIR = '/var/lib/metalk8s/storage/sparse'
+
+# https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/devices.txt#L191
+LOOP_MAJOR = 7
+
+# https://github.com/torvalds/linux/blob/master/include/uapi/linux/loop.h#L110
+LOOP_CLR_FD = 0x4C01
```
slaperche-scality: Since those constants are always used by a single function, maybe it's better to declare them near their usage.

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
```diff
@@ -0,0 +1,158 @@
...
+        if device_path is not None:
+            break
+    else:
+        die(
+            "Device for volume '{}' was not found".format(volume_id),
+            exit_code=errno.ENOENT,
+        )
+
+    if not is_loop_device(device_path):
+        die(
```
slaperche-scality: minor: instead of having a `die` function, maybe raising an exception (+ global try/catch around `main()`) would be clearer (more idiomatic at least).
This (global try/catch) would also handle the errors you forgot/ignored on purpose (failure of `os.stat` and `os.open` for instance).

Not sure it's worth it though, it's your call.

gdemonet: It's a good suggestion, and not too much work, I'll handle it.

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
gdemonet: Good point, I just store them at the top of my modules, out of habit :)

buildchain/buildchain/salt_tree.py
```diff
@@ -651,7 +651,8 @@ def _get_parts(self) -> Iterator[str]:
     Path('salt/metalk8s/volumes/init.sls'),
     Path('salt/metalk8s/volumes/prepared/init.sls'),
     Path('salt/metalk8s/volumes/prepared/installed.sls'),
-    Path('salt/metalk8s/volumes/provisioned/init.sls'),
+    Path('salt/metalk8s/volumes/prepared/files/metalk8s-sparse-volume@.service'),
```
slaperche-scality, line 654: About commit message: really good one :+1: 

On

> to include the fix from higher versions of `losetup`

Maybe add [a link to the fix](https://github.com/karelzak/util-linux/commit/f7e21185087c30c5f13bf92009238e316f8c1cf4) and/or mention in which version it was fixed (v2.25 IIRC), maybe when we migrate to CentOS 8 we can get rid of this workaround.

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
NicolasT: Hah, bike-shedding: I like them on top, actually.

buildchain/buildchain/salt_tree.py
gdemonet: Done!

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
gdemonet: Did something really basic, should do the trick :slightly_smiling_face:

salt/metalk8s/volumes/prepared/files/sparse_volume_cleanup.py
gdemonet: Ah sorry, missed your comment @NicolasT so already did the change. I think both sides have good arguments, but the stakes are too low for us to have this debate. So I won't change it anymore :stuck_out_tongue_winking_eye:

slaperche-scality (dismissed):

slaperche-scality (approved):

salt/metalk8s/volumes/prepared/installed.sls
```diff
@@ -17,3 +17,9 @@ Install gdisk:
   {{ pkg_installed('gdisk') }}
     - require:
       - test: Repositories configured
+
+# Needed by the sparse volume cleanup script
+Ensure Python 2 is available:
+  test.fail_without_changes:
```
NicolasT, line 23: Well, I would've used a `pkg_installed('python2')` or something along those lines here ;-) But sure, works for this particular case I guess.

NicolasT (approved):

PR #2938, opened 2020-11-20
[UI] Add Virtualise List for Volume Table

### LINKED ISSUE #2931: UI implement Virtualise list for Volume table
**Component**: ui, table

**Why this is needed**:
Since we have thousands of rows, we may consider using a technique known as “windowing”. It only renders a small subset of the rows at any given time and can dramatically reduce the time it takes to re-render the components as well as the number of DOM nodes created.

**What should be done**: Virtualize list

React-table, the table library we are using has already implemented this feature.
https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/virtualized-rows

Install dependencies. We’ll use **react-window** and also **react-virtualized-auto-sizer**.

```javascript
import { FixedSizeList as List } from "react-window"
import AutoSizer from "react-virtualized-auto-sizer"

<AutoSizer>
  {({ height, width }) => (
    <List
      className="List"
      height={height}
      itemCount={1000}
      itemSize={35}
      width={width}
    >
      {Row}
    </List>
  )}
</AutoSizer>
```
**Test plan**:
TBD

ChengYanJin:
**Component**: ui, tables

**Context**: 
With the use case of thousands of volumes, the table would meet performance issues when rendering them at once.
So instead of doing this, we just render a small subset of rows at a give time.

**Summary**:
Introduce _react-window_ and _react-virtualized-auto-sizer_  to realize the virtualize row. 
Adapt the interaction test

UI change: 
Replace the "Unknown" text in the **table** with a hyphen icon to reduce the noise that less important information brings.

**Acceptance criteria**: 
image
If we open the developer tool => Elements, we should see just a few of the `<tr/>` at the beginning, when we scroll down on the list, we can see the update of the list.
image

---

Closes: #2931


thomasdanan (dismissed):

thomasdanan (commented):
works well, just you have to mouse over very precisely on the hyphen to see the tooltip, may be the zone could be enlarged a bit

ChengYanJin:
> works well, just you have to mouse over very precisely on the hyphen to see the tooltip, may be the zone could be enlarged a bit

Done! Thanks for the feedback!

thomasdanan (dismissed):

ui/src/components/VolumeListTable.js
```diff
@@ -289,17 +263,86 @@ function Table({
     ?.isSortedDesc;
   useTableSortURLSync(sorted, desc, data);
 
+  const RenderRow = React.useCallback(
```
nicolas2bert, line 233: To lighten your component, you could move `RenderRow` to its own component.

ChengYanJin: Hello @nicolas2bert, 

Thanks for taking a look at this PR!
Just to be clear, you suggested me to move `RenderRow` to `VolumeListTable` component in my case?

nicolas2bert: I suggest you move `RenderRow` component out of `VolumeListTable` for clarity purposes.

ChengYanJin: OK. Actually `RenderRow` is a part of the `Table` component. And I render `Table` component in the `VolumeListTable`.

I guess the confusing part is I put the `Table` component inside `VolumeListTable` file, which should be separated. 

But since only `VolumeListTable` would use this Table for the moment, I did not generalize the `Table` yet.

nicolas2bert: Yes, moving `Table` component out of `VolumeListTable` would be another refacto.
I was talking about moving `RenderRow` out from `Table` and so being its own component.

nicolas2bert: The idea is that the component is already pretty heavy (and might need refacto later as you mention), let's try not to add too much code on top of it.

nicolas2bert: TableRow could even live in its own file to reduce the size of VolumeListTable.js

ChengYanJin: Indeed, thanks!

ui/src/components/TableRow.js
```diff
@@ -0,0 +1,98 @@
...
+
+const UnknownIcon = styled.i`
+  color: ${(props) => props.theme.brand.textSecondary};
+  // Increase the height so that the users don't need to hover precisely on the hyphen.
+  height: 30px;
+  padding-top: ${padding.base};
+`;
+
+const TableRow = (props) => {
+  const { row, style, rowClicked, volumeName, theme } = props;
```
JBWatenbergScality: I would rename `rowClicked` props to `onClick` to match standard event names

JBWatenbergScality: Instead of passing volumeName I would introduce an isSelected prop that reflect in a clearer way if the row is selected or not

JBWatenbergScality: In order to avoid props drilling of the theme you can rely on styled component which automatically map the theme as part of the props or use `useTheme` hook that retrieve current theme from styled-component internal context

ui/src/components/VolumeListTable.js
```diff
@@ -151,6 +100,13 @@ const TooltipContent = styled.div`
   min-width: 60px;
 `;
 
+const UnknownIcon = styled.i`
```
JBWatenbergScality: This component is duplicated in table row, please consider exporting it instead if it is necessary

ui/src/components/VolumeListTable.js
```diff
@@ -289,17 +246,31 @@ function Table({
     ?.isSortedDesc;
   useTableSortURLSync(sorted, desc, data);
 
+  const RenderRow = React.useCallback(
+    ({ index, style }) => {
+      const row = rows[index];
+      prepareRow(row);
+      // Extract the TableRow component outside of Table component
```
JBWatenbergScality: This comment can be deleted

ui/src/components/VolumeListTable.js
```diff
@@ -357,68 +334,48 @@ function Table({
...
                   textAlign: 'center',
                   background: theme.brand.primary,
+                  border: 'none',
                 }}
+                className="td"
               >
-                {intl.translate('no_volume_found')}
-              </td>
-            </HeadRow>
+                No Volume
```
JBWatenbergScality: A translation seems to be missing here, maybe `{intl.translate('no_volume_found')}`

JBWatenbergScality (approved):

PR #2943, opened 2020-11-25
calico: Update to Calico 3.17.0

TeddyAndrieux:
Deployment manifest updated based on upstream from
https://docs.projectcalico.org/v3.17/manifests/calico.yaml.

And Calico binaries sha256 updated from
https://github.com/projectcalico/cni-plugin/releases/tag/v3.17.0

---

### DIFF
--- a/packages/redhat/calico-cni-plugin.spec
+++ b/packages/redhat/calico-cni-plugin.spec
@@ -6,12 +6,12 @@
 
 %ifarch x86_64
 %global built_arch              amd64
-%global calico_sha256           baa67cb9fbceb2aebebd6b14e12d448c89dcf694f32b5f450ffd8b780fb3d18d
-%global calico_ipam_sha256      baa67cb9fbceb2aebebd6b14e12d448c89dcf694f32b5f450ffd8b780fb3d18d
+%global calico_sha256           9325d1ead70f2afb9222da64f0c7eaa1f3c5f68763241d6b07cff3a853b783f2
+%global calico_ipam_sha256      9325d1ead70f2afb9222da64f0c7eaa1f3c5f68763241d6b07cff3a853b783f2
 %endif
 
 Name:           calico-cni-plugin
-Version:        3.16.1
+Version:        3.17.0
 Release:        1%{?dist}
 Summary:        Calico CNI plugin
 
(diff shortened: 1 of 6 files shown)


packages/redhat/calico-cni-plugin.spec
```diff
@@ -6,12 +6,12 @@
 
 %ifarch x86_64
 %global built_arch              amd64
-%global calico_sha256           baa67cb9fbceb2aebebd6b14e12d448c89dcf694f32b5f450ffd8b780fb3d18d
-%global calico_ipam_sha256      baa67cb9fbceb2aebebd6b14e12d448c89dcf694f32b5f450ffd8b780fb3d18d
+%global calico_sha256           9325d1ead70f2afb9222da64f0c7eaa1f3c5f68763241d6b07cff3a853b783f2
+%global calico_ipam_sha256      9325d1ead70f2afb9222da64f0c7eaa1f3c5f68763241d6b07cff3a853b783f2
```
NicolasT, line 10: Same SHA256 for both now? That's surprising...

salt/metalk8s/kubernetes/cni/calico/deployed.sls
```diff
@@ -3787,6 +3762,10 @@ spec:
         #- name: cni-net-dir
         #  hostPath:
         #    path: /etc/cni/net.d
+        # Used to access CNI logs.
+        - name: cni-log-dir
+          hostPath:
+            path: /var/log/calico/cn
```
NicolasT: This seems odd. Not `cni`?

NicolasT: Also, should we create this on the host (using Salt)?

NicolasT (changes requested):

packages/redhat/calico-cni-plugin.spec
TeddyAndrieux: Yes and was already the case with 3.16.1

salt/metalk8s/kubernetes/cni/calico/deployed.sls
TeddyAndrieux: Ouh right typo on my side, good catch

Hmm I don't know :thinking: I need to test

TeddyAndrieux: We do not need to create the directory it's created automatically

NicolasT (dismissed):

NicolasT (approved):

PR #2948, opened 2020-11-27
Bump prometheus-operator version

alexandre-allard:
**Component**: charts, salt, build

**Context**:
Addon versions bump.

**Summary**:
Replace prometheus-operator which is deprecated by kube-prometheus-stack in version 12.2.3

**Acceptance criteria**: Currently testing if everything still OK.


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Release 2.7.0 (in development)
 ### Enhancements
+- Replace the prometheus-operator chart by the kube-prometheus-stack one and
+  bump the version to 12.2.3
+  (PR [#2948](https://github.com/scality/metalk8s/pull/2948))
```
gdemonet: Would be great to see which version of components are deployed with this chart (e.g. Grafana 7.2.1)... Maybe we can automate something, once we work on changelog from fragments.

gdemonet (commented):
I see a mention of Helm 3 in one of your commits... If we need it, then it's something we should at least document in the developer book, and likely also update in CI, I guess?

charts/render.py
```diff
@@ -72,7 +72,7 @@ def maybe_copy(doc, src, dest):
 
 
 def fixup_dict(doc):
-    if doc.get('heritage') == 'Tiller' or \
+    if doc.get('heritage') in ('Helm', 'Tiller') or \
             doc.get('app.kubernetes.io/managed-by') == 'Tiller':
```
TeddyAndrieux: I think you want to change this one as well

charts/render.py
```diff
@@ -279,10 +279,10 @@ def __call__(self, parser, args, values, option_string=None):
...
 
     template = subprocess.check_output([
-        'helm', 'template',
-        '--name', args.name,
+        'helm', 'template', args.name,
         '--namespace', args.namespace,
         '--values', args.values,
+        '--include-crds',
         args.path,
     ])
```
TeddyAndrieux, line 287: Should we do an if/else here to check if we are with helm2 or helm3 so that the script is compatible with helm2 as well (or we just expect all charts to be compatible with helm3, I don't know if helm3 has this kind of backward compatibility guaranteed)

alexandre-allard: On my system, both binaries are named `helm`, so I don't see how we can handle the two versions at the same time (except if we ask developers to have both, with specific paths or aliases, i.e.: `helm2` and `helm3`).
I think we may need to test if all charts are OK with helm 3, especially since helm 2 is now considered as deprecated.

TeddyAndrieux: We could easily have both,
But right if all charts are OK with helm3 we should get rid of helm2, I agree

charts/render.py
alexandre-allard: done

gdemonet (approved):

Issue #2949, opened 2020-11-27
The Environment list is broken in Solution UI when there is a Namespace in Terminating Status


ChengYanJin:
**Component**: ui, solution

**What happened**:
The Environment list does not display correctly when there is a namespace/environment on terminating status.
![image](https://user-images.githubusercontent.com/18453133/100442261-00251000-30a8-11eb-820c-33e181237165.png)
Got 404: https://10.200.4.124:8443/api/kubernetes/api/v1/namespaces/test/configmaps/metalk8s-environment

**What was expected**:
To be decided: 
**Do we consider to display the namespace on the list when it's in Terminating Status?** 
```
[root@metalk8sbootstrap centos]# kubectl get ns
NAME                  STATUS        AGE
default               Active        25h
kube-node-lease       Active        25h
kube-public           Active        25h
kube-system           Active        25h
metalk8s-auth         Active        25h
metalk8s-ingress      Active        25h
metalk8s-logging      Active        25h
metalk8s-monitoring   Active        25h
metalk8s-solutions    Active        25h
metalk8s-ui           Active        25h
test                  Terminating   19h
toto                  Active        19h
yanjin                Active        15h
yanjin-test           Terminating   20h
```

Why they are in terminating state? 
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    solutions.metalk8s.scality.com/environment-description: ""
  creationTimestamp: "2020-11-26T15:23:01Z"
  deletionTimestamp: "2020-11-26T15:27:14Z"
  labels:
    solutions.metalk8s.scality.com/environment: test
  name: test
  resourceVersion: "71785"
  selfLink: /api/v1/namespaces/test
  uid: 716f7b87-adf8-4d11-8b97-84c9ba862c53
spec:
  finalizers:
  - kubernetes
status:
  conditions:
  - lastTransitionTime: "2020-11-26T15:27:24Z"
    message: 'Discovery failed for some groups, 2 failing: unable to retrieve the
      complete list of server APIs: custom.metrics.k8s.io/v1beta1: the server is currently
      unable to handle the request, metrics.k8s.io/v1beta1: the server is currently
      unable to handle the request'
    reason: DiscoveryFailed
    status: "True"
    type: NamespaceDeletionDiscoveryFailure
  - lastTransitionTime: "2020-11-26T15:27:26Z"
    message: All legacy kube types successfully parsed
    reason: ParsedGroupVersions
    status: "False"
[...]

Issue #2941, opened 2020-11-25
Expansion fails if solution import is triggered in parallel


thomasdanan:
**Component**:

'salt'

**What happened**:

If we trigger cluster expansion by deploying a new node and in parallel import a new solution in the cluster, the node will fail to deploy. Actually, when importing a solution the `salt-master` is restarted to mount the new image and due to this the node deployment job is failing. 

**What was expected**:

We should not be able to trigger solution import if a node deployment going on. more generally speaking, all actions requiring a `salt-master` restart should be aborted if a node deployment is going on.

**Steps to reproduce**

Launch node deployment from the UI
Launch solution import right after

as a side note, after doing this, the UI was broken: the Node and Volume Pages were not displaying content anymore and some `salt-api` connection error was displayed in the console. The only way to get rid of this situation was to restart `salt-api` container.

**Resolution proposal** (optional):

PR #2954, opened 2020-12-01
[UI] Fix the UI bugs when there is environment under the Terminating status

ChengYanJin:
**Component**: UI, salt, solution

**Context**: 
The Environment list does not display correctly when there is a namespace/environment on terminating status.

**Summary**:
FIx the bug of #2949 

**Acceptance criteria**: 
image
image
- The environment list should be correctly displayed even there is env is under **Terminating** status, and we should not be able to delete this env again (The cursor is set to not-allowed)
- Add the missing translations and fix the typo of translation. 

---

Closes: #2949


ui/src/ducks/app/salt.js
```diff
@@ -178,6 +181,13 @@ export function* refreshJobStatus(job) {
       }),
     );
   }
+
+  // Check if the job jid from localStorage is exist. If not, we should remove it from localStorage
+  const jobNotFound = result?.return?.find(
+    (job) => Object.values(job)[0]['Error'] === ERROR_NO_JOB_JID_FOUND,
+  );
+  yield call(removeJobFromLocalStorage, Object.keys(jobNotFound)[0]);
```
gdemonet: Oh, I think that's not going to work well.

Try this:
- create a dummy `Node`
- click Deploy
- refresh the page, and look for the `printJob` call in your browser's network capture tool

Personally, until the job is finished, I get:

```json
{
  "return": [
    {
      "20201201120203793335": {
        "Result": {},
        "StartTime": "2020, Dec 01 12:02:03.793335",
        "Error": "Cannot contact returner or no job with this jid"
      }
    }
  ]
}
```

Which means the job would be cleared out, with this code.
However, if I query for a non-existing JID:

```json
{
  "return": [
    {
      "12345": {
        "Result": {},
        "StartTime": "",
        "Error": "Cannot contact returner or no job with this jid"
      }
    }
  ]
}
```

Maybe we should also check the `StartTime`...?

ui/src/services/salt/utils.js
```diff
@@ -85,9 +86,13 @@ export function getJobStatusFromPrintJob(result, jid) {
...
-    status.completed = true;
-    const returner = Object.values(job['Result'])[0].return;
-    status = { ...status, ...parseJobError(returner) };
+    if (job['Error'] === ERROR_NO_JOB_JID_FOUND) {
+      status.completed = true;
+    } else {
+      status.completed = true;
+      const returner = Object.values(job['Result'])[0].return;
+      status = { ...status, ...parseJobError(returner) };
+    }
```
gdemonet: Could be simpler to just check if there is _something_ in `job['Result']`.

ui/src/containers/SolutionList.js
```diff
@@ -193,6 +203,9 @@ const SolutionsList = (props) => {
                 dispatch(deleteEnvironmentAction(environment.name));
               }}
               inverted={true}
+              disabled={
+                environment.namespaces[0].status.phase === STATUS_TERMINATING
```
gdemonet: What happens if we have more than one namespace? Could be:

```javascript
environment.namespaces.some(ns => ns.status.phase === STATUS_TERMINATING)
```

ui/src/translations/fr.json
```diff
@@ -141,7 +141,8 @@
   "env_preparation": "Préparation de l’environnement",
   "env_preparation_success": "L'environnement {envName} est en cours de préparation.",
   "env_preparation_failed": "La préparation de l'environnement {envName} a échoué. {step} - {reason}",
-  "preparing_environemnt": "Préparer l'environnement",
+  "preparing_environment": "Préparer l'environnement",
```
gdemonet: If we have both "prepare" and "preparing", the right translation for "preparing" would be `Environnement en cours de préparation`.
What bugs me is that `env_preparation_success` already means "in progress", for some reason. Should we make sure all those translations are used appropriately?

ui/src/ducks/app/solutions.js
```diff
@@ -143,45 +143,64 @@ export function* prepareEnvironment(action) {
     console.error(`Environment '${envName}' is preparing`);
   }
 
-  const addSolutionToEnvironmentResult = yield call(
-    SolutionsApi.addSolutionToEnvironment,
-    envName,
-    solName,
-    solVersion,
+  // Need to check if Salt is under `deploy-node` before adding solution to environment
```
gdemonet: I don't think that's the issue with #2941 - "solution import" means running `./solutions.sh import --archive /path/to/solution.iso` from the Bootstrap node.
Adding a Solution to an environment only means performing K8s API calls (either from the UI directly, or from Salt master), but it should **not** trigger any Salt master restart.

So I don't think this check is needed at all.

ui/src/containers/SolutionList.js
```diff
@@ -180,7 +180,7 @@ const SolutionsList = (props) => {
             {isEnvironmentPreparing && (
               <LoaderContainer>
                 <Loader size="small"></Loader>
-                {intl.translate('preparing_environemnt', {
+                {intl.translate('preparing_environment', {
```
gdemonet, line 183: Should be in the same commit as the one changing the key in `src/translations/`.

ui/src/ducks/app/solutions.test.js
```diff
@@ -223,7 +223,9 @@ it('patch namespace with solution when preparing the evironment', () => {
   ];
   expect(gen.next().value.type).toEqual('SELECT');
   const clusterVersion = '2.6.0-dev';
-  expect(gen.next(existingEnv).value).toEqual(
+  expect(gen.next(existingEnv).value.type).toEqual('SELECT');
+  const jobs = [];
+  expect(gen.next(jobs).value).toEqual(
```
gdemonet: Updating a test should come in the same commit as the one changing the behaviour of the tested code, IMO.

gdemonet (changes requested):
Some things to change, but most importantly, I don't think we should implement any "locking" from the UI. If Salt cannot handle running two jobs in parallel, it should be responsible for saying so (or use some mechanism to queue one of them for later execution). UI cannot know everything (what if an incompatible job was triggered by CLI?), unfortunately.

ui/src/ducks/app/salt.js
ChengYanJin: Indeed, good catch! It will delete good jobs. 

I don't know how we can check the StartTime. Because let's say I executed Salt job from an instance, and switch to another instance. I will have a StartTime too. 

Maybe we should not remove any JID from localStorage?

ui/src/ducks/app/solutions.js
ChengYanJin: Absolutely right, I was confused with solution import and add a solution to an environment.
I will remove this checking...

ui/src/ducks/app/salt.js
gdemonet: > Because let's say I executed Salt job from an instance, and switch to another instance. I will have a StartTime too.

Are you sure? If the job ID doesn't exist, then I don't see how Salt can give you a `StartTime`...

slaperche-scality: @gdemonet That's weird that you get an error key with `Cannot contact returner or no job with this jid`, seems like something is wrong (the call failed or the ID really doesn't exists).
The presence of the `Error` key is also what we use in the operator:


```go
	// Check for "Error" field.
	if reason, found := info["Error"]; found {
		// Unknown Job ID: maybe the Salt server restarted or something like that.
		logger.Info("Salt job not found")
		return nil, fmt.Errorf("cannot get status for job %s: %v", jobID, reason)
	}
```
If your job is really in progress, then you'll get some info about it's status (and an key `Result` with an empty dict since it's not finished yet):

```go
	// Extract results.
	result, ok := info["Result"].(map[string]interface{})
	if !ok {
		return nil, fmt.Errorf("missing 'Result' key in %v", info)
	}
	// No result yet, the job is still running.
	if len(result) == 0 {
		logger.Info("Salt job is still running")
		return nil, nil
	}
```

At least for the states used by the storage-operator, I don't know if this holds for every Salt call.

gdemonet: OK, indeed there is something wrong somewhere (my bet is on Salt).

Tested the following:

- for a runner calling our `deploy_node` orchestrate, as the UI does

```
# salt-run --async state.orch metalk8s.orchestrate.deploy_node saltenv=metalk8s-2.6.0-dev pillar='{orchestrate: {node_name: worker-1}}'
[WARNING ] Running in asynchronous mode. Results of this execution may be collected by attaching to the master event bus or by examing the master job cache, if configured. This execution is running under tag salt/run/20201201163643992851
# salt-run jobs.print_job 20201201163643992851
20201201163643992851:
    ----------
    Error:
        Cannot contact returner or no job with this jid
    Result:
        ----------
    StartTime:
        2020, Dec 01 16:36:43.992851
[INFO    ] Runner completed: 20201201163704623108
```

- for a simple `salt` command, as the storage-operator does:

```
# salt --async 'worker-1' state.sls metalk8s.volumes saltenv=metalk8s-2.6.0-dev
Executed command with job ID: 20201201164240351912
# salt-run jobs.print_job 20201201164240351912
20201201164240351912:
    ----------
    Arguments:
        - metalk8s.volumes
        |_
          ----------
          __kwarg__:
              True
          saltenv:
              metalk8s-2.6.0-dev
    Function:
        state.sls
    Minions:
        - worker-1
    Result:
        ----------
    StartTime:
        2020, Dec 01 16:42:40.351912
    Target:
        worker-1
    Target-type:
        glob
    User:
        root
[INFO    ] Runner completed: 20201201164247350096
```

Of course, if I wait long enough, I get a full `Result` for both jobs, meaning there is no actual error in the first case outside the missing reporter.

ChengYanJin: > > Because let's say I executed Salt job from an instance, and switch to another instance. I will have a StartTime too.
> 
> Are you sure? If the job ID doesn't exist, then I don't see how Salt can give you a `StartTime`...

@gdemonet  I think if you try with the real job which generates the real timestamp (not a random number), you can get StartTime. 

So basically we should not remove any job with the error `Cannot contact returner or no job with this jid` right? ( It's not actually an error...) because it might be during the transition state.

ui/src/translations/fr.json
ChengYanJin: Thanks for the french translation of "preparing"!

Well, actually these two keywords hold the same meaning indeed, but they are used by different components.

`"env_preparation_success": "Environment {envName} is being prepared." `used by the notification 
`"preparing_environment": "Preparing the environment" `used by the loader in the env list when adding a solution which should be a shorter version.

so for me, it may not clear which one to use but make sure to have both.

ui/src/ducks/app/solutions.test.js
ChengYanJin: The test changes need to be removed since the behaviour changed is not necessary.

ui/src/ducks/app/salt.js
ChengYanJin: OK. As discussed during the meeting, I will drop this commit. 

Since the issue: Salt shares the Job JID because we store in localStorage will not happen in the production.

gdemonet (approved):

gdemonet:
Will need to wait for #2959

PR #2960, opened 2020-12-02
salt: Add missing requisite for image injection

gdemonet:
**Component**: salt

**Summary**:

The "Inject nginx image" state from `metalk8s.repo.installed` was not
requiring containerd/cri-tools to be available. In case the installation
of containerd failed, we should not try to run this state (because it
will always fail).

### DIFF
--- a/salt/metalk8s/repo/installed.sls
+++ b/salt/metalk8s/repo/installed.sls
@@ -17,12 +17,16 @@
 {%- set image_fullname = docker_repository ~ '/' ~ image_name ~ ':' ~ image_version %}
 
 include:
+  - metalk8s.container-engine.containerd
   - .configured
 
 Inject nginx image:
   containerd.image_managed:
     - name: {{ image_fullname }}
     - archive_path: {{ archives[saltenv].path }}/images/{{ image_name }}-{{ image_version }}.tar
+    - require:
+      - service: Start and enable containerd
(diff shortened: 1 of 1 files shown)


NicolasT:
Given this is a clear bugfix, should it not go in `development/2.6`? (maybe even 2.5)

salt/metalk8s/repo/installed.sls
```diff
@@ -17,12 +17,16 @@
 {%- set image_fullname = docker_repository ~ '/' ~ image_name ~ ':' ~ image_version %}
 
 include:
+  - metalk8s.container-engine.containerd
```
TeddyAndrieux, line 20: I do not really like this, it means if you configure repo (for example iso-manager) you will potentially upgrade (or downgrade) containerd as well.

I would suggest  to create a state to ensure containerd pkg is installed instead

NicolasT: You're supposed to use `iso-manager` of the *current* version anyway, so the version of `containerd` on said system should be the one we want already, meaning executing this state should not result in any changes, no?

TeddyAndrieux: Not exactly, for registry we have one file per version and we need to run the state for all version deployed (especially the "new version" state so that we have the file for the new registry)

TeddyAndrieux:
This issue is fixed in #3020

PR #2961, opened 2020-12-02
Bump ingress-nginx to 3.13.0

TeddyAndrieux:
**Component**:

'ingress-controller'

**Context**: 

Bump nginx-ingress version

**Summary**:

- Use the new helm3 chart for ingress-nginx version 3.12.0
- Adapt a bit the options file to works with the new charts
- Rename from `nginx-ingress` to `ingress-nginx`
- For upgrade, downgrade purpose: remove old ingress controller at the very end

---


gdemonet (dismissed):
Changes look good to me :+1: 

Just a comment / question (if we want to do it, it's fine to tackle in a follow-up PR IMO): in K8s 1.18, there is a new `IngressClass` resource type, which is supported in the updated chart we now use. Would it make sense to also update the `Ingress` objects we deploy to use this new API, and also make sure (once we have a 2.8 branch) to update their apiVersion to `/v1` (from `/v1beta1`)?

TeddyAndrieux:
> Just a comment / question (if we want to do it, it's fine to tackle in a follow-up PR IMO): in K8s 1.18, there is a new `IngressClass` resource type, which is supported in the updated chart we now use. Would it make sense to also update the `Ingress` objects we deploy to use this new API, and also make sure (once we have a 2.8 branch) to update their apiVersion to `/v1` (from `/v1beta1`)?

Sure but I think it's not possible with the current Python-Kubernetes we embed (and even with the last version available 1.17)

gdemonet (approved):

PR #2968, opened 2020-12-09
Prometheus retention customization

### LINKED ISSUE #2953: Make Prometheus retention period configurable
**Component**: monitoring, csc, salt

**Why this is needed**:
To allow user to change the retention period of Prometheus (which is 10 days by default).

**What should be done**:
Make it configurable through CSC ConfigMap `metalk8s-prometheus-config`.
We should also update documentation to explain how to customize retention.

Add a default in `salt/metalk8s/addons/prometheus-operator/config/prometheus.yaml`:
```
apiVersion: addons.metalk8s.scality.com
kind: PrometheusConfig
spec:
  config:
    retention: 10d
```
Then render `salt/metalk8s/addons/prometheus-operator/deployed/chart.sls` using this entry:
```
       release: prometheus-operator
   portName: web
   replicas: {% endraw -%}{{ prometheus.spec.deployment.replicas }}{%- raw %}
-  retention: 10d
+  retention: {% endraw -%}{{ prometheus.spec.config.retention }}{%- raw %}
   routePrefix: /
   ruleNamespaceSelector: {}
   ruleSelector:
```
The option will then be automatically configurable through CSC.

**Bonus:** we could also expose `retentionSize` (disabled by default) which is as the name suggests a retention based on the size, both options can be set at the same time. If we do so, we need to warn the user in the documentation that actually Prometheus does not take in account the WAL size, so the required size is greater than `retentionSize` (add at least 10% to be safe).

**Test plan**:

alexandre-allard:

**Context**: See #2953

**Summary**:

**Acceptance criteria**: 

---

Closes: #2953

### DIFF
--- a/docs/operation/cluster_and_service_configuration.rst
+++ b/docs/operation/cluster_and_service_configuration.rst
@@ -219,22 +219,84 @@ applied with Salt.
 Prometheus Configuration Customization
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Default configuration for Prometheus can be overridden by editing its
+Cluster and Service ConfigMap ``metalk8s-prometheus-config`` in namespace
+``metalk8s-monitoring`` under the key ``data.config.yaml``:
+
+.. code-block:: shell
+
+   root@bootstrap $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
+                      edit configmap -n metalk8s-monitoring \
+                      metalk8s-prometheus-config
+
+Change Retention Time
+"""""""""""""""""""""
+
+Prometheus is deployed with a retention based on time (10d).
+This value can be overriden:
+
+.. code-block:: yaml
+
+   ---
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: metalk8s-prometheus-config
+     namespace: metalk8s-monitoring
+   data:
+     config.yaml: |-
+       apiVersion: addons.metalk8s.scality.com
+       kind: PrometheusConfig
+       spec:
+         config:
+           retention_time: 30d
+
+.. note::
+
+   Supported time units are y, w, d, h, m s and ms
+   (years, weeks, days, hours, minutes, seconds and milliseconds).
+
+Set Retention Size
+""""""""""""""""""
+
+Prometheus is deployed with the size-based retention disabled.
+This functionality can be actived:
+
+.. code-block:: yaml
+
+   ---
+   apiVersion: v1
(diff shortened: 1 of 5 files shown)


salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
```diff
@@ -53429,7 +53429,7 @@ spec:
         - --path.sysfs=/host/sys
         - --path.rootfs=/host/root
         - --web.listen-address=$(HOST_IP):9100
-        - --collector.diskstats.ignored-devices=^(ram|fd|(h|s|v)d[a-z]|nvme\\d+n\\d+p)\\d+$
+        - --collector.diskstats.ignored-devices=^(ram|fd)\\d+$
```
alexandre-allard, line 53432: Ignore this line as this comes from the Prometheus stack version bump (I forgot to re-render after having changed a parameter in `charts/kube-prometheus-stack.yaml`)

alexandre-allard:
Do we want a changelog entry for this kind of stuff ?

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -219,22 +219,77 @@ applied with Salt.
...
+.. note::
+
+   Supported time units are y, w, d, h, m s and ms
+   (years, weeks, days, hours, minutes, seconds and milliseconds).
+
+Set Retention Size
+""""""""""""""""""
+
+Prometheus is deployed with the size-based retention disabled.
+This functionality can be actived:
```
gdemonet, line 263: From #2953:

> If we do so, we need to warn the user in the documentation that actually Prometheus does not take in account the WAL size, so the required size is greater than `retentionSize` (add at least 10% to be safe).

So we're missing a warning :wink:

salt/metalk8s/addons/prometheus-operator/config/prometheus.yaml
```diff
@@ -7,6 +7,9 @@ spec:
   # Configure the Prometheus Deployment
   deployment:
     replicas: 1
+  config:
+    retention_time: 10d
+    retention_size: 0
```
gdemonet: Since this file is `literal-include`d in our docs, would make sense to add a comment here saying that `0` means size-based retention is disabled.

gdemonet (changes requested):
Mostly about user-facing information. This PR would also need a changelog entry, since we're adding a feature (didn't see your comment, sorry).

gdemonet (dismissed):

gdemonet (approved):

Issue #2792, opened 2020-09-13
[UI New Volume Page] Handle the possible cases for volume list table 


ChengYanJin:
**Component**: ui

**Why this is needed**:
Currently, we didn't handle all the possible situations for New Volume Page.

**What should be done**:
- `/volumes?node=toto`
If node `toto` doesn't exist, the filter can be displayed but nothing matches in the table (same as if the node exists but no volume in this node)
We should dispaly "No volume found” in the volume table

- `/volumes/toto`
If volume `toto` doesn't exist, the right-side panel should show "volume toto not found"

- `/toto`
That's a 404 Not Found page.

- no volume at all return by the k8s volumes object
That's the [Empty State Page](https://github.com/scality/core-ui/issues/198)

**Implementation proposal** (strongly recommended):
Properly address the 4 situations above.

Please check react-table: [no-data-text](https://github.com/tannerlinsley/react-table/blob/v6/examples/custom-no-data-text/src/index.js)
```
 data={[]}
          noDataText="Oh Noes!"
          columns={[]....}

```

Note that: We should apply the same logic for the Node Page too.

**Test plan**:
TBD

ChengYanJin:
There is one point haven't address yet which is the 404 Not Found page.
because the design is not ready yet. 

I will close this PR and create another issue for the 404 page.

PR #2971, opened 2020-12-09
Improvement/2792 nodes volumes tables edge cases

alexis-ld:
**Component**: ui, nodes, volumes, emptystate

**Context**: 
This adresses the issues described in #2792 .

**Summary**:

- Both nodes and volumes table have appropriate message when no data matches the filters.
- If an non-existing volumes/node is specified in the URL the right panel displays and appropriate message
- If no nodes/volumes exist at all the page displays an "EmptyState" component that guides the user to create his first node/volume. (the component has dynamic link and icon to be re-used in both pages) : 

metalk8s-empty-demo

**Acceptance criteria**: 
The issues described in #2792 are fixed.

---

### DIFF
--- a/ui/src/containers/NodePageContent.js
+++ b/ui/src/containers/NodePageContent.js
@@ -45,7 +45,7 @@ const NodePageContent = (props) => {
   return (
     <PageContentContainer>
       {!nodeTableData.length && firstLoading ? (
-        <EmptyState label={'Node'} link="/nodes/create" />
+        <EmptyState label={'Node'} link="/nodes/create" icon="fa-server" />
       ) : (
         <Fragment>
           <LeftSideInstanceList>
--- a/ui/src/containers/NodePageContent.js
+++ b/ui/src/containers/NodePageContent.js
@@ -21,7 +21,7 @@ const NodePageContent = (props) => {
   const { nodeTableData, alerts, loading } = props;
   const { path } = useRouteMatch();
   const [defaultSelectNodeName, setDefaultSelectNodeName] = useState(null);
-  const [firstLoading, setFirstLoading] = useState(false);
+  const [isFirstLoadingDone, setIsFirstLoadingDone] = useState(false);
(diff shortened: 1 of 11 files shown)


ui/src/containers/NodePageContent.js
```diff
@@ -32,29 +44,37 @@ const NodePageContent = (props) => {
...
-          ></Route>
-          <Route
-            path={`${path}/:name`}
-            render={() => {
-              return <NodePageRSP nodeTableData={nodeTableData} />;
-            }}
-          ></Route>
-        </Switch>
-      </RightSidePanel>
+      {!nodeTableData.length && firstLoading ? (
```
ChengYanJin, line 47: I would rename `firstLoading` to `isFirstLoadingDone`

ChengYanJin: I would rename `firstLoading` to `isFirstLoadingDone`

ui/src/containers/VolumePage.js
```diff
@@ -76,6 +76,7 @@ const VolumePage = (props) => {
   const node = useSelector((state) => makeGetNodeFromUrl(state, props));
   const nodes = useSelector((state) => state.app.nodes.list);
   const volumes = useSelector((state) => state.app.volumes.list);
+  const volumesLoading = useSelector((state) => state.app.volumes.loading);
```
ChengYanJin: We can't arrive at the `EmptyState Page` in Volume Page, because we use `isLoading` property instead of `loading` ;)

Should be: 
```js
const volumesLoading = useSelector((state) => state.app.volumes.isLoading);
```

ChengYanJin (changes requested):
Thanks for the work!

Test by the integration test with mock backend, the Node Page works well. 
But not Volume Page, please address my comments.

ChengYanJin (approved):
LGTM!
The empty state is working for Volume list as well!

PR #2972, opened 2020-12-09
ci, tests: Ignore failed Pods if their equivalent is available

gdemonet:
**Component(s)**: ci, tests

**Context**:

Since K8s 1.18, we are seeing many occurences of Pods being rejected by kubelet and ending up in `NodeAffinity` state. This breaks the implementation of our `wait_pods_status.sh` script, which expects all scheduled Pods to be `Running`. It also would break our `logs should be accessible` E2E test if it were to run, since this test expects all scheduled Pods to be `Running` and have some logs.

**Summary**:

- Rework the stabilization script to only ensure Pods are not transitioning (if a Pod has failed, but remains in that state, it won't prevent this script from succeeding)
- Adjust the workaround for #2726 in `snapshot_upgrade` to use this new script and still ensure all Pods are `Running` before moving on with the upgrade
- Rework our sanity E2E tests to focus on MetalK8s-provided objects, still ignoring failed Pods if any

**Acceptance criteria**:

- Both pre-merge and post-merge pass
- CI runs are not affected by the `NodeAffinity` flaky anymore


alexandre-allard (approved):
LGTM and it seems that you got a green build, lucky guy! :)

PR #2976, opened 2020-12-11
Generate snapshot in the CI

### LINKED ISSUE #2906: Add eve stage to generate simple snapshots
**Component**:

'build', 'tests'

**Why this is needed**:

We use snapshot in post-merge to test upgrade from released versions, but for the moment we generate those snapshot manually

**What should be done**:

Create an eve stage (that will likely be only triggered manually) that create the snapshot needed for future upgrade in CI.

As a first step we can consider only 2 simple environments:
- Single node with a solution
- 3 nodes with a solution

The stage will deploy the VMs as expected to be used for future snapshot upgrade, it will also bootstrap and do expansions if needed + install the solution.

TeddyAndrieux:
**Component**:

'eve',

**Context**: 

#2906 

**Summary**:

- Make multi-node stage handling various number of nodes (up to 2 for the moment as we do not handle more in pytest)
- Add server IDs in terraform outputs so that we can use those IDs to directly act on server using `openstack` client
- Add a script to generate snapshot from terraform output using `openstack` client
- Add a stage to generate 3 nodes and single node snapshots

---

Fixes: #2906

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -1787,11 +1787,16 @@ stages:
           haltOnFailure: true
       # --- We do not need to Test promoted version ---
       # --- Upgrade to version N ---
-      - ShellCommand: *copy_iso_bootstrap_ssh
+      - ShellCommand:
+          <<: *copy_iso_bootstrap_ssh
+          env:
+            <<: *_env_copy_iso_bootstrap_ssh
+            DEST: "metalk8s-%(prop:metalk8s_version)s.iso"
       - ShellCommand:
           <<: *add_archive_ssh
           env:
             <<: *_env_add_archive_ssh
+            ISO_PATH: "metalk8s-%(prop:metalk8s_version)s.iso"
             PRODUCT_VERSION: "%(prop:product_promoted_version)s"
       - ShellCommand:
           name: Run upgrade from promoted version
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -562,6 +562,10 @@ models:
         OS_PASSWORD: "%(secret:scality_cloud_password)s"
         OS_TENANT_NAME: "%(secret:scality_cloud_tenant_name)s"
         TF_VAR_prefix: "%(prop:buildnumber)s-%(prop:stage_name)s"
+        TF_VAR_os: "%(prop:os:-centos-7)s"
+        TF_VAR_nodes_count: "%(prop:nodes_count:-0)s"
+        TF_VAR_rhsm_username: "%(secret:rhel_ci_login)s"
+        TF_VAR_rhsm_password: "%(secret:rhel_ci_password)s"
         TF_VAR_debug: "%(prop:metalk8s_debug:-false)s"
         MAX_RETRIES: "3"
       workdir: build/eve/workers/openstack-terraform/terraform/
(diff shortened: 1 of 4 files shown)


thomasdanan (approved):
Snapshot present when launching the eve stage ...

eve/main.yml
```diff
@@ -597,11 +601,20 @@ models:
...
         SSH_CONFIG: ssh_config
-        REPORT_OWNER: centos
-        REPORT_GROUP: centos
-      command: >
+        # NOTE: If REPORT_OWNER or REPORT_GROUP is empty auto detect from
+        # os property (which default to centos-7)
+        REPORT_OWNER: ""
+        REPORT_GROUP: ""
+      command: |-
+        HOSTS_LIST=${HOSTS_LIST:-bootstrap$(seq -s '' --format ' node-%%g' 1 %(prop:nodes_count:-0)s)}
```
gdemonet, line 703: (really minor) IIUC `-s` describes a separator, so why not use `seq -s ' ' --format 'node-%g' 1 $nodes_count` for clarity?

eve/main.yml
```diff
@@ -2073,45 +2086,44 @@ stages:
...
         TEST_SUITE: install
-        CLASS_NAME: multi node.centos7
+        CLASS_NAME: "multi node.%(prop:os:-centos-7)s"
         TEST_NAME: bootstrap restore
       simultaneous_builds: 20
     worker: *terraform_worker
     steps:
       - Git: *git_pull
+      - SetProperty:
+          # NOTE: Set property will be skipped if the value is already set,
```
gdemonet, line 2238: Always? Skipped if set by user or also by some "parent build stage"?

eve/main.yml
```diff
@@ -2042,7 +2055,7 @@ stages:
           name: Run installation scenarii on the bastion
           env:
             <<: *_env_bastion_tests
-            PYTEST_FILTERS: "install and ci and multinodes and not node2"
+            PYTEST_FILTERS: "install and ci and multinodes and not node$((%(prop:nodes_count)s + 1))"
```
gdemonet, line 2161: I guess we'll need to adjust these "tests", don't think they can handle more than 2 nodes.

eve/main.yml
```diff
@@ -589,6 +591,83 @@ models:
...
+        sudo "$METALK8S_MOUNTPOINT/solutions.sh"
+        import --archive "\$(readlink -f \"$SOLUTION_ARCHIVE\")"
+      workdir: build/eve/workers/openstack-terraform/terraform/
+      haltOnFailure: true
+  - ShellCommand: &activate_solution
+      doStepIf: "%(prop:install_solution:-false)s"
+      name: Activate Solution
+      env:
+        SOLUTION_NAME: "%(prop:solution_name:-example-solution)s"
+        SOLUTION_VERSION: "%(prop:solution_version:-1.0.0)s"
```
gdemonet, line 646: Can't we read these from the downloaded archive? This could serve as a minimal check, and would reduce the number of build props to set (using a Zenko archive as an example). The current approach needs at least 5 properties to be set:

  - `install_solution=true`
  - `solution_base_url=https://zenko-artifacts/2.3.4/` (or `solution_repo=https://zenko-artifacts/`)
  - `solution_archive=zenko-2.3.4.iso`
  - `solution_name=zenko`
  - `solution_version=2.3.4`

Reading the name and version from the ISO could reduce the required props to `install_solution`, `solution_base_url` and `solution_archive` (the two latter could be reduced to a single `solution_archive_url` from which we can extract the basename).

gdemonet (approved):
Some minor comments (which we can address later), overall LGTM :+1:

eve/main.yml
TeddyAndrieux: Just because I do not want to have "bootstrap " if we have no nodes :smile:

eve/main.yml
TeddyAndrieux: Hmm you right, I will do it, in another PR

eve/main.yml
TeddyAndrieux: Yes but we do not need more than 2 nodes for the moment and since it's only internal I think it's ok but you right if one day we need more we will need to update the expansion "test" (or change the way we do expansion in the CI :wink: )

eve/main.yml
TeddyAndrieux: Btw it's not the normal behavior of "SetProperty" but since I do a `value: "%(prop:nodes_count:-2)s"` yes it's always the case

Or yes you right it's not really "skipped" just we take the current value. I will update the comment message in another PR

eve/main.yml
gdemonet: It doesn't seem to be the behaviour of `seq`:

```bash
$ val=$(seq -s ' ' --format 'node-%g' 1 0)
$ echo "<start>$val<end>"
<start><end>
```

eve/main.yml
gdemonet: > since I do a value: "%(prop:nodes_count:-2)s" yes it's always the case

Ah right I missed this :+1: no rush for updating the comment :wink:

eve/main.yml
TeddyAndrieux: Yes exactly but  "bootstrap" is not part of this `seq` 
So if you do 
```
# echo -e "<start>bootstrap $(seq -s ' ' --format 'node-%g' 1 0)<end>"
<start>bootstrap <end>

gdemonet: Ahhh, right, :man_facepalming:

Issue #2964, opened 2020-12-04
Batch Volume Creation from MetalK8s UI


ChengYanJin:
**Component**: ui, volumes

**Why this is needed**:
We want to be able to batch create volume from UI.

**What should be done**:
![image](https://user-images.githubusercontent.com/18453133/101155661-b9936080-3627-11eb-93e1-5e9b1e4c7cf5.png)
![image](https://user-images.githubusercontent.com/18453133/101155678-bf894180-3627-11eb-9662-e1d0e247c7ab.png)
Please check the HD from Sketch:

https://www.sketch.com/s/b7c9bece-ad22-40e6-b2b5-410645289650/a/qP32vd
https://www.sketch.com/s/b7c9bece-ad22-40e6-b2b5-410645289650/a/L5Vox5

**Implementation proposal** (strongly recommended):
When we choose `SparseLoop` Device, we can have the size field for each of the volume instead of path.

The number selection component is missing in core-ui. I would suggest first implement in MetalK8s and generalize it to core-ui.

For the moment, I only see `Device Path` require question mark next to it with the following text as tooltip.
`A devicePath or a partition mount point. Example: /dev/sdb`

We apply the same logic as the Scality Cloud does. 
when selecting the multi-volume creation, by default we should have 1 volume to create.

We should dispatch the data in an array. 
```js
  [
    { name: "volume01'", devicePath: '/dev/sdb' },
    { name: "volume02'", devicePath: '/dev/sdc' },
  ];
```
**Test plan**:
We should implement the interaction test for the batch volume creation.

PR #2980, opened 2020-12-14
[UI] Batch volume creation

ChengYanJin:
**Component**: ui, volumes 

**Context**: 
We want to be able to batch create volume from UI.

**Summary**:
Note that:
For `RawBlockDevice`, the recommended path only happens with `'/dev/vda'`, we will see later if 
For `SparseLoopDevice`, the size of the batch volume are the same as the default one.

**Acceptance criteria**: 
1. The Default Volume Creation Form
image

2. Volume type `RawBlockDevice` batch creation 
image

3. Volume type `SparseLoopDevice` batch creation. (From big HD screen)
image

---

Closes: #2964


ChengYanJin:
Wrong target, will create a new PR.

Issue #213, opened 2018-07-26
Use one ConfigMap per Grafana dashboard


NicolasT:
See https://github.com/scality/metal-k8s/pull/203#issuecomment-408185576

NicolasT:
Ideally, `kube-prometheus` would support a selector for `ConfigMap`s, such that we can deploy dashboards *as part of the related role*, instead of stashing everything in `kube_prometheus`.

PR #2981, opened 2020-12-14
[UI] Volume Batch Creation

### LINKED ISSUE #2964: Batch Volume Creation from MetalK8s UI

ChengYanJin:
**Component**: ui, volumes 

**Context**: 
We want to be able to batch create volume from UI.

**Summary**:
Note that:
For `RawBlockDevice`, the recommended path only happens with `'/dev/vda'`, we will see later if 
For `SparseLoopDevice`, the size of the batch volume are the same as the default one.

**Acceptance criteria**: 
1. The Default Volume Creation Form
image

2. Volume type `RawBlockDevice` batch creation 
image

3. Volume type `SparseLoopDevice` batch creation. (From big HD screen)
image

---

Closes: #2964

### DIFF
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -5,7 +5,7 @@ import { Formik, Form } from 'formik';
 import * as yup from 'yup';
 import styled from 'styled-components';
 import Loader from '../components/Loader';
-import { Input, Button, Breadcrumb, Banner, Tooltip } from '@scality/core-ui';
+import { Input, Button, Banner, Tooltip, Checkbox } from '@scality/core-ui';
 import isEmpty from 'lodash.isempty';
 import {
   fetchStorageClassAction,
@@ -18,11 +18,6 @@ import {
   fontWeight,
 } from '@scality/core-ui/dist/style/theme';
 import { SPARSE_LOOP_DEVICE, RAW_BLOCK_DEVICE } from '../constants';
-import {
-  BreadcrumbContainer,
-  BreadcrumbLabel,
-  StyledLink,
-} from '../components/BreadcrumbStyle';
 import { sizeUnits, useQuery } from '../services/utils';
 import { intl } from '../translations/IntlGlobalProvider';
 
@@ -159,7 +154,7 @@ const CreateVolume = (props) => {
 
   const nodes = useSelector((state) => state.app.nodes.list);
   const storageClass = useSelector((state) => state.app.volumes.storageClass);
-  const theme = useSelector((state) => state.config.theme);
+
   const api = useSelector((state) => state.config.api);
 
   useEffect(() => {
@@ -252,23 +247,7 @@ const CreateVolume = (props) => {
   ) : (
     <PageContainer>
       <CreateVolumeFormContainer>
-        <BreadcrumbContainer>
-          <Breadcrumb
-            activeColor={theme.brand.secondary}
-            paths={[
-              <BreadcrumbLabel title={intl.translate('platform')}>
-                {intl.translate('platform')}
-              </BreadcrumbLabel>,
-              <StyledLink to="/volumes">
-                {intl.translate('volumes')}
-              </StyledLink>,
-              <BreadcrumbLabel>
-                {intl.translate('create_new_volume')}
-              </BreadcrumbLabel>,
-            ]}
-          />
-        </BreadcrumbContainer>
-
+        <TitlePage>Create New Volume</TitlePage>
         {isStorageClassExist ? null : (
           <Banner
             variant="warning"
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -1,7 +1,7 @@
 import React, { useState, useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router';
-import { Formik, Form } from 'formik';
+import { Formik, Form, FieldArray, useFormikContext, useField } from 'formik';
 import * as yup from 'yup';
 import styled from 'styled-components';
 import Loader from '../components/Loader';
@@ -18,7 +18,11 @@ import {
   fontWeight,
 } from '@scality/core-ui/dist/style/theme';
 import { SPARSE_LOOP_DEVICE, RAW_BLOCK_DEVICE } from '../constants';
-import { sizeUnits, useQuery } from '../services/utils';
+import {
+  sizeUnits,
+  useQuery,
+  linuxDrivesNamingIncrement,
+} from '../services/utils';
 import { intl } from '../translations/IntlGlobalProvider';
 
 // We might want to do a factorization later for
@@ -237,16 +241,55 @@ const CreateVolume = (props) => {
     sizeInput: '',
     labels: {},
     multiVolumeCreation: false,
-    // When the multi-volume creation mode is active, the default number is 1.
+    // When the multi-volume creation mode is active, the default/min number is 1.
     numberOfVolumes: 1,
-    // MultiVolumes should be the array of objects.
-    // For example:
-    // [
-    //   { name: "volume01'", devicePath: '/dev/sdb' },
-    //   { name: "volume02'", devicePath: '/dev/sdc' },
-    //   ...
-    // ];
-    multiVolumes: [],
+    volumes: [{ name: '', path: '' }],
+  };
+
+  // Set the dependend /recommand fields based on the default values of other fields (DevicePath / Size / SizeUnit) in Formik.
+  const RecommendNameField = (props) => {
+    const {
+      values: { name },
+      touched,
+      setFieldValue,
+    } = useFormikContext();
+    const [field, meta] = useField(props);
+    React.useEffect(() => {
+      if (name.trim() !== '' && touched.name) {
+        setFieldValue(props.name, `${name}${props.index + 1}`);
+      }
+    }, [name, touched.name, setFieldValue, props.name, props.index]);
+
+    return (
+      <>
+        <Input {...props} {...field} />
+        {!!meta.touched && !!meta.error && <div>{meta.error}</div>}
+      </>
+    );
+  };
+
+  const RecommendDevicePathField = (props) => {
+    const {
+      values: { path },
+      touched,
+      setFieldValue,
+    } = useFormikContext();
+    const [field, meta] = useField(props);
+    React.useEffect(() => {
+      if (path.trim() !== '' && touched.path) {
+        setFieldValue(
+          props.name,
+          linuxDrivesNamingIncrement(path, props.index + 1),
+        );
+      }
+    }, [path, touched.path, setFieldValue, props.name, props.index]);
+
+    return (
+      <>
+        <Input {...props} {...field} />
+        {!!meta.touched && !!meta.error && <div>{meta.error}</div>}
+      </>
+    );
   };
 
   const volumeNameRegex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
@@ -291,6 +334,14 @@ const CreateVolume = (props) => {
       then: yup.string(),
     }),
     labels: yup.object(),
+    volumes: yup.array().of(
+      yup.object().shape({
+        name: yup
+          .string()
+          .matches(volumeNameRegex, intl.translate('name_error'))
+          .required(''),
+      }),
+    ),
   });
 
   const isStorageClassExist = storageClassesName.length > 0;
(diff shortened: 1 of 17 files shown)


ui/src/services/utils.js
```diff
@@ -419,3 +419,45 @@ export const useDynamicChartSize = (container_id) => {
...
+ *
+ * @example
+ * const string = '/dev/vda'
+ *
+ * const nextDevicePath = linuxDrivesNamingIncrement(string)
+ */
+export const linuxDrivesNamingIncrement = (string, increment) => {
+  if (string.match(/^\/dev\/vd[a-z]/)) {
+    while (increment--) {
+      var lastChar = string.slice(-1);
```
JBWatenbergScality: I'm not yet sure about your styleguide conventions but I would always prefer usage of let or const for variable definitions, using const here will reflect clearly that this variable doesn't change elsewhere and will slightly simplify reading this code

ui/src/services/utils.js
```diff
@@ -419,3 +419,45 @@ export const useDynamicChartSize = (container_id) => {
...
+ * 18278	    vdzzz
+ *
+ * @param {string} string - The string that will be incremented
+ *
+ * @example
+ * const string = '/dev/vda'
+ *
+ * const nextDevicePath = linuxDrivesNamingIncrement(string)
+ */
+export const linuxDrivesNamingIncrement = (string, increment) => {
```
JBWatenbergScality: I would rename `string` to something more explicit such as `drivePath`

ui/src/services/utils.js
```diff
@@ -419,3 +419,45 @@ export const useDynamicChartSize = (container_id) => {
...
+ * 703	      vdaaa
+ * 704	      vdaab
+ * 18278	    vdzzz
+ *
+ * @param {string} string - The string that will be incremented
+ *
+ * @example
+ * const string = '/dev/vda'
+ *
+ * const nextDevicePath = linuxDrivesNamingIncrement(string)
```
JBWatenbergScality: This example lack increment parameter

ui/src/services/utils.js
JBWatenbergScality: Additionally for utils function like this one it seems important for me to add a unit test that covers the edge cases you want to support, but not sure if it is your traditional way of working so we can discuss about it.

ui/src/services/utils.js
```diff
@@ -419,3 +419,45 @@ export const useDynamicChartSize = (container_id) => {
 
   return [graphWidth, window.innerHeight / 6 - 30];
 };
+
+/**
+ * This function increments the string by 1
```
JBWatenbergScality: This description doesn't seems to be accurate

ui/src/containers/CreateVolume.js
```diff
@@ -237,16 +241,55 @@ const CreateVolume = (props) => {
...
+      }
+    }, [path, touched.path, setFieldValue, props.name, props.index]);
+
+    return (
+      <>
+        <Input {...props} {...field} />
+        {!!meta.touched && !!meta.error && <div>{meta.error}</div>}
+      </>
+    );
   };
```
JBWatenbergScality: As those two components only differs in the way they setFieldValue, they might be factorized.

ui/src/containers/CreateVolume.js
```diff
@@ -362,6 +413,34 @@ const CreateVolume = (props) => {
...
 
+              // Update the volume list base on the current input number
+              const setVolumeNumber = (field, arrayHelpers) => (
+                selectedObj,
+              ) => {
+                const inputVolNum = selectedObj.target.value;
+                const preVolNum = values.numberOfVolumes;
+
+                setFieldValue(field, inputVolNum);
+                var diff = preVolNum - inputVolNum;
```
JBWatenbergScality: I would use const instead to declare this variable

ui/src/containers/CreateVolume.js
```diff
@@ -362,6 +413,34 @@ const CreateVolume = (props) => {
...
+                if (diff > 0) {
+                  // REMOVE volume object from `values.volumes` base on the index
+                  for (var i = inputVolNum; i < preVolNum; i++) {
```
JBWatenbergScality: I would prefer use let to avoid any hoisting issues

ui/src/containers/CreateVolume.js
```diff
@@ -362,6 +413,34 @@ const CreateVolume = (props) => {
...
+                    arrayHelpers.remove(i);
+                  }
+                } else if (diff < 0) {
+                  // PUSH new volume object to `values.volumes`
+                  var absDiff = Math.abs(diff);
```
JBWatenbergScality: I would use const instead to declare this variable

ui/src/ducks/app/volumes.js
```diff
@@ -118,8 +118,8 @@ export const updateStorageClassAction = (payload) => {
   return { type: UPDATE_STORAGECLASS, payload };
 };
 
-export const createVolumeAction = (newVolume) => {
-  return { type: CREATE_VOLUMES, payload: { newVolume } };
+export const createVolumeAction = (newVolumes) => {
```
JBWatenbergScality: I would rename this to `createVolumesAction` to reflect the batch creation possibility

ui/src/ducks/app/volumes.js
```diff
@@ -205,97 +205,97 @@ export function* fetchStorageClass() {
...
+      }
     } else {
+      // We might want to change this behavior later
       yield put(
         addNotificationErrorAction({
-          title: intl.translate('volume_creation'),
-          message: intl.translate('volume_creation_failed', {
-            name: newVolume.name,
-          }),
+          title: 'Volume Form Error',
```
JBWatenbergScality, line 294: Translation seems to be missing here

ui/src/containers/CreateVolume.js
```diff
@@ -54,6 +55,8 @@ const ActionContainer = styled.div`
 const CreateVolumeLayout = styled.div`
   display: inline-block;
   margin-top: ${padding.base};
+  overflow-y: scroll;
```
JBWatenbergScality: Is it intended to display scrollbars by default even for non scrollable content ? You might prefer to use `auto` instead

ui/src/containers/CreateVolume.js
```diff
@@ -622,7 +640,22 @@ const CreateVolume = (props) => {
...
                         onChange={handleChange('path')}
-                        label={intl.translate('device_path')}
+                        label={
+                          <>
+                            {intl.translate('device_path')}
+                            <Tooltip
+                              placement="right"
+                              overlay={
+                                <div style={{ minWidth: '200px' }}>
+                                  A devicePath or a partition mount point.
```
JBWatenbergScality: I think there is a missing translation here

ui/src/containers/CreateVolume.js
```diff
@@ -290,9 +382,18 @@ const CreateVolume = (props) => {
             initialValues={initialValues}
             validationSchema={validationSchema}
             onSubmit={(values) => {
-              const newVolume = { ...values };
-              newVolume.size = `${values.sizeInput}${values.selectedUnit}`;
-              createVolume(newVolume);
+              const newVolumes = { ...values };
+              newVolumes.size = `${values.sizeInput}${values.selectedUnit}`;
+              if (newVolumes.multiVolumeCreation) {
```
JBWatenbergScality: This if could be transferred in formatDataForBatchVolumeCreation as it already partially exists there and just missing the else branch to return a one element array

ui/src/services/NodeVolumesUtils.js
```diff
@@ -276,3 +276,28 @@ export const getVolumeListData = createSelector(
     return nodeVolumes.sort((a, b) => compareHealth(b.health, a.health));
   },
 );
+
+export const formatDataForBatchVolumeCreation = (newVolumes) => {
```
JBWatenbergScality: I would add a simple unit test for the two branches of this function, also as mentioned in the usage of this function I would handle the `!multiVolumeCreation` case here and return a single entry array containing newVolumes, that would allow a simpler usage of the function as its ouput will always be of the same shape `(newVolumes: VolumeCreationFormState) => Volume[]` instead of `(newVolumes: VolumeCreationFormState) => Volume[] | void`

ui/src/services/utils.js
ChengYanJin: Indeed, it's a very bad name. I will rename it to `devicePath`.

thomasdanan (dismissed):
@ChengYanJin  I like the behavior.
I would change one little thing:
When you create multiple Volumes, the Device Path for each Volume is computed from the Global Device Path field. I think the first Volume should get what you input in the Global Device Path field (i.e. not increment the Device Path of the first Volume).

Also when playing with multiple Volume check box and the number of Volumes, you will observe strange behavior. Maybe some state not properly managed

I wonder if we should disable Global Name and Device Path global fields when creating multiple Volumes because if you override Name or Device Path and then you change the Global field, you will lose all your customizations (@JBWatenbergScality what do you think?)

Also for number increments, I would do 01, 02, 03 etc ... instead of 1, 2, 3 .... 

Thanks and well done

JBWatenbergScality:
@thomasdanan @ChengYanJin From my point of view once the name or path is touched we shouldn't change automatically their values to maintain the user intentions but should still let the ability for the user to change the first volume name and path. 

Also the ui might be a little bit misleading as the user is unsure if the default values will actually create a volume. To be precise mentioning `/dev/vda` by default and then creating one more volume will pre-complete the first entry with `/dev/vdb` in the current behavior but when submitting the form just `/dev/vdb` will be created.

The simplest interaction I could think of closed enough to the current implementation would be to remove the default name and path fields when checking `Create multiple volumes?` and compute the names of the volumes from the first one if subsequent one are not touched.

ui/src/services/utils.js
```diff
@@ -479,3 +479,23 @@ export const linuxDrivesNamingIncrement = (devicePath, increment) => {
     return '';
   }
 };
+
+/**
+ * This function formats the name base on the index
+ * @param {string} name - The name the default volume name
```
JBWatenbergScality: Remove redundant name `The name the default volume`

ui/src/services/utils.js
```diff
@@ -479,3 +479,23 @@ export const linuxDrivesNamingIncrement = (devicePath, increment) => {
...
+ * @param {number} index  - The number of index
+ *
+ * @example
+ * const name = 'volume-test'
+ *
+ * const nextDevicePath = linuxDrivesNamingIncrement(name, 1)
```
JBWatenbergScality: `const formatedName = formatBatchName(name, 1) `

ui/src/containers/CreateVolume.js
```diff
@@ -513,12 +633,108 @@ const CreateVolume = (props) => {
...
+                            >
+                              <span
+                                style={{
+                                  paddingRight: `${padding.base}`,
+                                }}
+                              >
+                                {intl.translate('number_volume_create')}
+                              </span>
+                              {/* TODO: Generalize the number input component to core-ui. */}
+                              <InputNumberComponentStyle
```
JBWatenbergScality, line 711: I think we should handle a max number of volumes to avoid dom freeze in case of the user mistakenly ask 10000 volumes, simply add a max prop here should be enough

ChengYanJin: Indeed, good point!
I am going to set the `max` number to`70` for the moment.

ChengYanJin:
Hello @thomasdanan @JBWatenbergScality ,

According to the feedback from you and our offline discussions. I've addressed the following issues. 

1.
> Thomas: When you create multiple Volumes, the Device Path for each Volume is computed from the Global Device Path field. I think the first Volume should get what you input in the Global Device Path field (i.e. not increment the Device Path of the first Volume).

Fix the bug of the Device path, the first volume should be the same as the Global field.

2.
> Thomas: Also when playing with multiple Volume check box and the number of Volumes, you will observe strange behavior. Maybe some state not properly managed

Fix the bug, will add interaction test of the form in the following PR.

3.
> Thomas: I wonder if we should disable Global Name and Device Path global fields when creating multiple Volumes because if you override Name or Device Path and then you change the Global field, you will lose all your customizations (@JBWatenbergScality what do you think?)

 In order to keep the user customization, edit the global value(name/path) should not affect the defaults pre-fill of the touched field.
Unless the users uncheck the box of multi-volume creation and re-check again, we only update the untouched field with the new global value.

4. 
> Thomas: Also for number increments, I would do 01, 02, 03 etc ... instead of 1, 2, 3 ....

Done.

5. Set the maximum number of the volume to create `70` to prevent the UI frozen. 
The UI should pop up the error message when the input number is bigger than 70 (Waiting for the [PR #213](https://github.com/scality/core-ui/pull/213) in core-ui to be landed)

Please let me know if you see any improvements/ bugs to be done. 
Thank you!

ui/src/containers/CreateVolume.js
```diff
@@ -421,6 +421,33 @@ const CreateVolume = (props) => {
...
+                  }
+                  if (
+                    touched &&
+                    touched.volumes &&
+                    touched?.volumes[i]?.path === true
+                  ) {
+                    continue;
+                  } else {
+                    setFieldValue(`volumes[${i}]path`, '');
+                  }
```
JBWatenbergScality: I think this can be simplified to : 
```js
if (!touched?.volumes[i]?.name) {
    setFieldValue(`volumes[${i}]name`, '');
}
if (!touched?.volumes[i]?.path) {
    setFieldValue(`volumes[${i}]path`, '');
}
```

ChengYanJin: Hi @JBWatenbergScality ,

Yes. Good point. But I guess we need to use optional chaining to make sure the Volumes array exists.
Please see:

```js
                  if (!touched?.volumes?.[i]?.name) {
                    setFieldValue(`volumes[${i}]name`, '');
                  }
                  if (!touched?.volumes?.[i]?.path) {
                    setFieldValue(`volumes[${i}]path`, '');
                  }
```

JBWatenbergScality (approved):

PR #2984, opened 2020-12-15
Ensure loop device management is accurately updated during upgrade/downgrade

### LINKED ISSUE #2982: Upgrading to 2.7 does not re-enable loop device provisioning
**Component**: salt, upgrade

**What happened**:

When implementing a solution for #2726 in #2936, we failed to account for upgrade scenarios, where existing SparseLoopDevice Volumes already provisioned will not have a proper systemd unit created.
Thus, after upgrade, any such Volume will fail to have its device provisioned on node reboot.

**What was expected**: Seamless upgrade, loop devices still provisioned on node startup.

**Steps to reproduce**: From a MetalK8s 2.6 cluster with existing SparseLoopDevice Volumes, upgrade to 2.7 and reboot.

- During upgrade, ensure the systemd unit files are created for existing Volumes
- During downgrade, remove the unit files (and re-provision the devices using the `metalk8s_volumes` module)

gdemonet:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #2982


salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -207,6 +207,41 @@ Check pillar before highstate:
...
+    units, introduced in 2.7 #}
+{%- if pillar.orchestrate.get('cleanup_loop_devices', False) %}
+  {%- if pillar.orchestrate.get('skip_draining', False) %}
+
+Cannot cleanup devices if skipping drain:
+  test.fail_without_changes: []
+
+  {%- else %}
+    {%- set loop_volume_ids = [] %}
+    {%- for volume in pillar.metalk8s.volumes %}
```
gdemonet: Ah, that can't work, this pillar is computed for a minion in particular... I guess the state should handle it on its own.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -70,7 +72,29 @@ Cordon the node:
   metalk8s_cordon.node_cordoned:
     - name: {{ node_name }}
 
-{%- if not pillar.orchestrate.get('skip_draining', False) %}
+{%- if not skip_draining or cleanup_loop_devices %}
+  {%- set run_drain = not skip_draining %}
```
TeddyAndrieux: Minor: It's not clear the first time you read this :smile: I would prefer to set `run_drain` at the beginning and use this variable directly in the if
```
{%- set run_drain = not pillar.orchestrate.get('skip_draining', False) %}
[...]
{%- if run_drain or cleanup_loop_devices %}
[...]
```

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -70,7 +72,29 @@ Cordon the node:
...
+                             | selectattr('spec.nodeName', 'equalto', node_name)
+                             | selectattr('spec.sparseLoopDevice', 'defined')
+                             | list
+    %}
+    {%- if volumes_to_clean %}
+      {%- do salt.log.warning(
+        'Forcing drain to clean up loop devices for the following Volumes: '
+        ~ volumes_to_clean | map(attribute='metadata.name') | join(', ')
+      ) %}
+      {%- set run_drain = True %}
```
TeddyAndrieux, line 91: Do we really want to force ?
Could be better/safer to just fail hard at this point, no ?

salt/metalk8s/volumes/cleanup.sls
```diff
@@ -0,0 +1,56 @@
...
+  {%- if not sparse_volumes %}
+
+Nothing to cleanup:
+  test.succeed_without_changes: []
+
+  {%- else %}
+    {%- for volume in sparse_volumes %}
+      {%- set volume_name = volume.metadata.name %}
+      {%- set volume_id = volume.metadata.uid %}
+      {%- if mode == 'upgrade' %}
```
TeddyAndrieux: Wow ! I do not really like this "mode" stuff, I would prefer to check the node version (which is the desired version) to know if we go to a 2.6.x or 2.7.x

salt/metalk8s/orchestrate/deploy_node.sls
gdemonet: If it's not clear then I'll change :+1:

(done in 780698a)

salt/metalk8s/orchestrate/deploy_node.sls
gdemonet: Not sure: if we fail hard, it means we can never upgrade a single-node cluster with sparse volumes (unless we remove the volumes before upgrade, which is cumbersome). Since such clusters only exist for dev purposes, I prefer to force the drain (and create a service interruption) rather than prevent the upgrade.

I guess another option could be to require some `metalk8s:upgrade:force_drain` pillar value or something along those lines, but I'm not sure it's worth the effort. WDYT @scality/metalk8s ?

salt/metalk8s/volumes/cleanup.sls
gdemonet: OK let me change that :+1:

(done in 780698a)

salt/metalk8s/orchestrate/deploy_node.sls
TeddyAndrieux: Hmm yes ok you right make sense.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -131,6 +154,28 @@ Wait minion available:
...
+  salt.state:
+    - tgt: {{ node_name }}
+    - saltenv: metalk8s-{{ version }}
+    - sls:
+      - metalk8s.volumes.cleanup
+    - require:
+      - salt: Wait minion available
+      - metalk8s_drain: Drain the node
+    - require_in:
+      - salt: Check pillar before etcd deployment
```
TeddyAndrieux: Hmm it means this orchestrate will fail if we are not on an etcd node

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -73,8 +73,60 @@ Upgrade salt-minion on {{ node }}:
...
+    - tgt: {{ node }}
+    - saltenv: {{ saltenv }} {#- Use current version of this formula #}
+    - sls:
+      - metalk8s.volumes.cleanup
+    - require:
+      - metalk8s_drain: Drain node {{ node }}
     - require_in:
       - salt: Deploy node {{ node }}
+
+  {%- endif %}
```
TeddyAndrieux, line 129: We only need this if dest_version < 2.7.0 maybe worth add an if otherwise, we will do some unnecessary cleanup when downgrading from 2.7.1 to 2.7.0 (even if I know the state will basically do nothing because systemd unit already exists)

salt/metalk8s/volumes/cleanup.sls
```diff
@@ -0,0 +1,56 @@
+# Clean already attached loop devices before applying highstate, to avoid
+# creating duplicates for a same sparse file.
+# This state can cleanup either "manually", for upgrade, or by disabling the
+# systemd units, for downgrade.
+{%- set target_version = pillar.metalk8s.nodes[grains.id].version %}
+
+include:
+  - .installed
```
TeddyAndrieux: Hmm do we really need to include this "installed" when going to version < 2.7.0 ? Because it's weird to me, we want to remove the systemd unit file but first, we ensure the unit file exists

salt/metalk8s/volumes/cleanup.sls
```diff
@@ -0,0 +1,56 @@
...
+Disable systemd unit for Volume {{ volume_name }}:
+  service.dead:
+    - name: metalk8s-sparse-volume@{{ volume_id }}
+    - enable: false
+    - require:
+      - test: Ensure Python 3 is available
+      - file: Set up systemd template unit for sparse loop device provisioning
+      - file: Install clean-up script
+
+      {%- endif %}
```
TeddyAndrieux: At the very end of the for loop we may want to remove the systemd unit file, so that nobody can manually enable back the service, no ?

salt/metalk8s/volumes/cleanup.sls
gdemonet: `.installed` only takes care of "dependencies": some packages for formatting devices, but most importantly, it ensures the cleanup script and the systemd unit **template** are available.

When going from 2.7 to 2.6, we only need the template, to make sure the service can be found (not sure how `service.dead` behaves if this file does not exist).
When going from 2.6 to 2.7, we need the cleanup script, which we call manually.

salt/metalk8s/volumes/cleanup.sls
gdemonet: Maybe, but I'm not sure why someone would want to enable this service manually unless they know what they're doing. And not sure what the issue would be, except that they may get duplicate loop devices (but then again, they would know, no?).

salt/metalk8s/volumes/cleanup.sls
TeddyAndrieux: > When going from 2.7 to 2.6, we only need the template, to make sure the service can be found (not sure how service.dead behaves if this file does not exist).

```
local:
----------
          ID: toto
    Function: service.dead
      Result: True
     Comment: The named service toto is not available
     Started: 13:39:24.705098
    Duration: 26.128 ms
     Changes:   

Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:  26.128 ms
```

I got your point but I think if we have the service running then we already have all the dependencies no ?
Anyway, as you want, it's not really important we can include this "installed" it does not hurt

salt/metalk8s/volumes/cleanup.sls
TeddyAndrieux: Hmm I don't know, 
I'm a user, I have some trouble on my platform, I check for "metalk8s" system services I see this one stopped, I start it :cry: :smile: 
Anyway it will happen only if a user do a downgrade and try to debug itself the cluster after this so ...I agree it's not critical and can be handled later

salt/metalk8s/orchestrate/downgrade/init.sls
gdemonet: Actually, this section is already included within an `if` block (which is used to migrate Salt + Python):

```sls
{%- if salt.pkg.version_cmp(dest_version, '2.7.0') == -1 %}
```

(https://github.com/scality/metalk8s/pull/2984/files/c8c911c4b96741a7987292c0f94caab77e4a6a21#diff-b9a4482d485ff3e2dcd700a24619346950f0db61ef5c07cb0fffe6c9541445c8R62)

TeddyAndrieux: Oh ! You right I didn't see :+1:

salt/metalk8s/volumes/cleanup.sls
gdemonet: I still need it when upgrading, so maybe I can only include `if target_version >= 2.7.0`?

salt/metalk8s/volumes/cleanup.sls
gdemonet: Since I'm reworking this state, I'll add the cleanup of this unit template.
But I'm not sure how you can find the disabled+stopped services:

```
$ # output after a downgrade from this branch to 2.6.0
$ systemctl list-units -a | grep metalk8s
  srv-scality-metalk8s\x2d2.6.0.mount          loaded    active   mounted   /srv/scality/metalk8s-2.6.0
  srv-scality-metalk8s\x2d2.7.0\x2ddev.mount   loaded    active   mounted   /srv/scality/metalk8s-2.7.0-dev
  var-lib-metalk8s-storage-sparse.mount        loaded    active   mounted   /var/lib/metalk8s/storage/sparse
  system-metalk8s\x2dsparse\x2dvolume.slice    loaded    active   active    system-metalk8s\x2dsparse\x2dvolume.slice
```

salt/metalk8s/volumes/cleanup.sls
TeddyAndrieux: Yes maybe the good approach, I think

salt/metalk8s/volumes/cleanup.sls
TeddyAndrieux: Hmm ok you right, you can have it with "systemctl list-unit-files" but yes ok you do not have the service, anyway would be cleaner to remove the unit file and it does not cost much

salt/metalk8s/orchestrate/deploy_node.sls
gdemonet: Added a check

salt/metalk8s/volumes/cleanup.sls
gdemonet: Done

salt/metalk8s/volumes/cleanup.sls
gdemonet: Done

TeddyAndrieux (commented):
LGTM :+1: 
(another PR will be needed to clean up dev/2.8 :) )

TeddyAndrieux (dismissed):
(Oups + approve) :smile:

gdemonet:
> (another PR will be needed to clean up dev/2.8 :) )

@TeddyAndrieux I'll take care of the clean-up in the integration branch instead, directly in the merge commit, if that's fine with you.

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):

PR #2986, opened 2020-12-16
buildchain, docs, salt, tests: update Python dependencies

NicolasT:
Ran `tox -e pip-compile` to bring everything up-to-date.

The `dependabot`-generated PR fails because for some reason, `importlib-metadata` is (attempted to be) installed but has no entry in `requirements.txt` (and hence no hash). Turns out this is due to `pluggy` now using `importlib.metadata` from the standard library when using Python >= 3.8, which is likely what @dependabot uses.

Hence, pinning the Python version used for `pip-compile` to Python 3.6, which is what RHEL/CentOS7 comes with (and what we run in CI).

The updated dependencies also bring a new version of `pylint` (which doesn't work with Python 3.9, hence also pinning to 3.6...). This introduces some new warnings, which are fixed in some related commits as well.

Finally, rename some unfortunately-named variables in a `tox.ini` script.

Closes: https://github.com/scality/metalk8s/pull/2897
Closes: https://github.com/scality/metalk8s/pull/2898
See: https://github.com/pytest-dev/pluggy/issues/222


slaperche-scality (commented):
LGTM, but a Mac user (@thomasdanan or @ChengYanJin) should run the `pip-compile` to also update `buildchain/requirements-Darwin.txt`.

NicolasT:
> LGTM, but a Mac user (@thomasdanan or @ChengYanJin) should run the `pip-compile` to also update `buildchain/requirements-Darwin.txt`.

@thomasdanan will do. I wonder whether there's a way to unify these two... The only difference appears to be including `pyinotify` on Linux and `macfsevents` on Darwin, both dependencies of `doit`. Could we pin some version of them in some other file with a platform-specific selector, and include that file with the `pip install -r` invocation?

slaperche-scality:
> I wonder whether there's a way to unify these two... The only difference appears to be including `pyinotify` on Linux and `macfsevents` on Darwin, both dependencies of `doit`. Could we pin some version of them in some other file with a platform-specific selector, and include that file with the `pip install -r` invocation?

I remember trying a lot of things back then, but none of them worked out, but I may have missed something ([environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) sounded like a good option, but didn't work with `pip-compile`)
The current solution is far from being convenient, clearly :/

NicolasT:
> > I wonder whether there's a way to unify these two... The only difference appears to be including `pyinotify` on Linux and `macfsevents` on Darwin, both dependencies of `doit`. Could we pin some version of them in some other file with a platform-specific selector, and include that file with the `pip install -r` invocation?
> 
> I remember trying a lot of things back then, but none of them worked out, but I may have missed something ([environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) sounded like a good option, but didn't work with `pip-compile`)
> The current solution is far from being convenient, clearly :/

Here's a suggestion:

- We create a `buildchain/constraints.txt` which constraints both `pyinotify` and `macfsevents` to a specific (current) version. From now on these are 'special' packages
- We create `buildchain/platform-requirements.txt` *manually* which includes both `pyinotify` and `macfsevents` at the version specified in the constraint file, including hashes and whatnot, and platform markers
- We add `-c constraints.txt` to `buildchain/requirements.in`
- We add a filter which runs after `pip-compile` (on `buildchain/requirements.in`) to remove both `pyinotify` and `macfsevents` from the resulting `requirements.txt`, and add `-r platform-requirements.txt` to the top

If I'm not mistaken, this then allows to run `pip install -r buildchain/requirements.txt` and get everything we want. Unless at some point one of the top-level dependencies adds some other platform-specific dependency, but :man_shrugging:

NicolasT:
For this PR, I'm going to revert the changes in `buildchain/requirements-*.txt` (and note so in the commit message), so this can go in (without updating the requirements for `doit`, both Linux and Darwin). Then later we can look into coding [the potential solution](https://github.com/scality/metalk8s/pull/2986#issuecomment-748019832).

@JBWatenbergScality tried updating on an OSX machine, but it turns out Python3.6 is not supported on Big Sur *sigh*

NicolasT:
@bert-e status

JBWatenbergScality (dismissed):

gdemonet (approved):
LGTM :ok_hand:

Issue #3001, opened 2020-12-24
UI - Enhance application typing - monitoring reducer


JBWatenbergScality:
**Component**:

UI, typing

**Why this is needed**:

This brings typesafety and enhanced developer experience to the project.

**What should be done**:

 Add @flow annotations in `ui/src/ducks/app/monitoring.js` and fix all the linting issues raised. Try not using `any` keyword.

**Test plan**:

Run flow cli to ensure typing is correctly interpreted, navigate through the code base in your IDE and see if all references to monitoring reducer are correctly typed

Issue #3000, opened 2020-12-24
UI - Enhance application typing - salt client


JBWatenbergScality:
**Component**:

UI, typing

**Why this is needed**:

This brings typesafety and enhanced developer experience to the project.

**What should be done**:

 Add @flow annotations in `ui/src/ducks/app/salt.js` and fix all the linting issues raised. Try not using `any` keyword and type salt responses precisely in `ui/src/services/salt/api.js`

**Test plan**:

Run flow cli to ensure typing is correctly interpreted, navigate through the code base in your IDE and see if all references to salt objects are correctly typed

Issue #2994, opened 2020-12-21
Alerts without a "message" annotation don't show meaningful details in MetalK8s UI


gdemonet:
> I have installed metalk8s and all seems to work fine, except for Alert UI. I don't see alert message for some of the alert like KubeCPUOvercommit. 
>
> [...]
>
> ```yaml
> - alert: KubeCPUOvercommit
>     annotations:
>       description: Cluster has overcommitted CPU resource requests for Pods and cannot tolerate node failure.
>       runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubecpuovercommit
>       summary: Cluster has overcommitted CPU resource requests.
> ```

_Originally posted by @n1603 in https://github.com/scality/metalk8s/discussions/2993_

We should likely handle all known alert annotations to display the most relevant information to our users when an alert fires.

n1603:
@gdemonet, Thanks, this is how I handled this. :  [link](https://github.com/n1603/metalk8s/blob/development/2.7/ui/src/containers/ClusterMonitoring.js#L173), please let me know your feedback. Similar change for VolumeAlertsTab.js and update on [monitoring test](https://github.com/n1603/metalk8s/blob/development/2.7/tests/post/steps/test_monitoring.py#L256).

NicolasT:
@n1603 Could you create a PR to contribute your fixes? @ChengYanJin left some notes on the process at https://github.com/scality/metalk8s/discussions/2993#discussioncomment-267200. Thanks!

n1603:
@NicolasT, thanks. if fix helps, I will create PR after going through the process.

n1603:
some issue with my pr branch squash not working as expected, will re-create.

n1603:
PR : https://github.com/scality/metalk8s/pull/3030

NicolasT:
I believe this is now closed by #3038 (#3030).

PR #3002, opened 2020-12-24
ui - enhance flow typing

### LINKED ISSUE #2999: UI - Enhance application typing - kubernetes client and reducers
**Component**:

UI, typing

**Why this is needed**:

This brings typesafety and enhanced developer experience to the project.

**What should be done**:

 - Add type definitions to `ui/src/services/k8s/api.js`
 - Add type definitions to `ui/src/ducks/reducer.js`
 - Type `useSelector` when it is being used with `useSelector<RootState>` or simply `useTypedSelector`

**Test plan**:

Run flow cli to ensure typing is correctly interpreted, navigate through the code base in your IDE and see if all references to kubernetes objects are correctly typed

JBWatenbergScality:
**Component**:

ui, typing

**Context**: 

This pull request is a first milestone to introduce better typings in the project. 
This brings type-safety and enhanced developer experience to the project.

As introducing strong typing everywhere in the project would require a few more pull requests the idea of this one is to focus on kubernetes client and reducers that make use of kubernetes client. However it also brings typings for few others reducers and introduce starting points to expand typing to other areas in the project.

Closes: #2999

See: #3000 #3001


ui/src/services/k8s/api.js
```diff
@@ -1,16 +1,60 @@
...
+    storageClassName: string
+  }
+}
+
+export type Metalk8sV1Alpha1Volume = {
+  apiVersion: 'storage.metalk8s.scality.com/v1alpha1',
+  kind: 'Volume',
+  metadata?: V1ObjectMeta,
+  spec: {
+    mode: 'Filesystem',
```
gdemonet: The mode can be either `Filesystem` or `Block`, actually. See [our docs](https://metal-k8s.readthedocs.io/en/latest/operation/volume_management/volume_creation_deletion_cli.html#volume-creation) or [the CRD](https://github.com/scality/metalk8s/blob/development/2.7/storage-operator/deploy/crds/storage.metalk8s.scality.com_volumes_crd.yaml#L43-L49)

ui/src/services/k8s/api.js
```diff
@@ -1,16 +1,60 @@
...
+    sparseLoopDevice?: {
+      size: string
+    },
+    rawBlockDevice?: {
+      devicePath: string
+    },
+    storageClassName: string
+  },
+  status: {
+    deviceName: string,
```
gdemonet: If this was not yet reconciled (e.g. because the Volume is `Pending`), this will be empty from the Operator's PoV, and absent in the API object.

ui/src/services/k8s/api.js
```diff
@@ -1,16 +1,60 @@
...
+    nodeName: string,
+    sparseLoopDevice?: {
+      size: string
+    },
+    rawBlockDevice?: {
+      devicePath: string
+    },
+    storageClassName: string
+  },
+  status: {
```
gdemonet: If you want to be exhaustive, there is also a `status.job` key (which is also optional, and its usage is internal to the Operator).

ui/src/ducks/app/volumes.js
```diff
@@ -138,42 +156,48 @@ export const refreshPersistentVolumesAction = () => {
...
   return { type: STOP_REFRESH_PERSISTENT_VOLUMES };
 };
 
-export const updateVolumesAction = (payload) => {
+export const updateVolumesAction = (payload: any) => {
   return { type: UPDATE_VOLUMES, payload };
 };
 
-export const fetchCurrentVolumeObjectAction = (volumeName) => {
+export const fetchCurrentVolumeObjectAction = (volumeName: any) => {
```
gdemonet: I guess we can safely assume this should be a `string`, no?

ui/src/ducks/app/volumes.js
```diff
@@ -138,42 +156,48 @@ export const refreshPersistentVolumesAction = () => {
...
 
 // Sagas
-export function* fetchVolumes() {
+export function* fetchVolumes(): Generator<{
+  body: {items: Metalk8sV1Alpha1Volume[]};
+} | {error: any, body: null}, {
+  body: {items: Metalk8sV1Alpha1Volume[]};
+} | {error: any, body: null}, {
+  body: {items: Metalk8sV1Alpha1Volume[]};
+} | {error: any, body: null}> {
```
gdemonet: I'm not sure that's a valid type for sagas. IIUC, the type signature for a coroutine is `Generator<Yield, Return, Next>`. Sagas only yield "effects" (in the sense of `redux-saga`), and they may be fed with `any`thing.
For this reason, there is a handy type definition in `flow-typed` for the `redux-saga` library:

```javascript
import type { Saga } from 'redux-saga';

// same as:
declare type Saga<T> = Generator<Effect, T, any>
```

See the definitions [here](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux-saga_v1.x.x/flow_v0.104.x-/redux-saga_v1.x.x.js)

ui/src/services/api.js
```diff
@@ -1,15 +1,69 @@
...
-let apiClient = null;
+let apiClient: ApiClient = null;
 
-export function initialize(apiUrl) {
+export function initialize(apiUrl: string) {
   apiClient = new ApiClient({ apiUrl });
 }
 
-export function fetchTheme() {
+export type Brand = {
```
gdemonet, line 10: OK for now, but IMHO this type (`Theme` as well) should live in `@scality/core-ui`.

ui/src/services/api.js
```diff
@@ -1,15 +1,69 @@
...
+export type Themes = {
+  "light": Theme,
+  "dark": Theme,
+  "custom": Theme
+}
+
+export type WrappedThemes = {
+  "theme": Themes,
+  "default": "dark"
+}
```
gdemonet: I'm not entirely sure if this is the intended type (maybe it's what the code expects today, didn't check). For me, the goal is more to have something like:

```javascript
export type WrappedThemes = {
  "theme": { [key: string]: Theme },
  "default": string
}
```

ui/src/ducks/app/nodes.js
```diff
@@ -161,6 +162,25 @@ const defaultState = {
   currentNodeObject: null,
 };
 
+export type NodesState = {
+  clusterVersion: string,
+  isRefreshing: boolean,
+  isLoading: boolean,
+  IPsInfo: any,
+  currentNodeObject: any,
```
gdemonet: Missing some `TODO` comments?

ui/src/ducks/app/pods.js
```diff
@@ -44,7 +64,7 @@ export function* fetchPods() {
             pod.status.containerStatuses && pod.status.containerStatuses.length
               ? pod.status.containerStatuses[0].restartCount
               : 0,
-          volumes: pod?.spec?.volumes?.map((volume) => ({
+          volumes: (pod?.spec?.volumes || []).map((volume) => ({
```
gdemonet: Shouldn't we use the nullish coallescing operator for that?

```javascript
pod?.spec?.volumes ?? []
```

ui/src/ducks/app/solutions.js
```diff
@@ -130,10 +137,10 @@ export function* createEnvironment(action) {
   yield call(fetchEnvironments);
 }
 
-export function* prepareEnvironment(action) {
+export function* prepareEnvironment(action: {payload: {envName: string, solName: string, solVersion: string }}): Generator<any, void, any> {
   const { envName, solName, solVersion } = action.payload;
 
-  const existingEnv = yield select((state) => state.app.solutions.environments);
+  const existingEnv: SolutionsApi.Environments = yield select((state) => state.app.solutions.environments);
```
gdemonet: I think that's a better approach to type effect returns (especially for `call`s) than trying to type the coroutine itself. May be worth generalizing to other sagas.

gdemonet: I'm a bit surprised though, don't we have a type mismatch? The annotation L.49 gives:

```javascript
  environments: SolutionsApi.Environment[],
```

But the type definition `SolutionsApi.Environments` describes a map (`{ [name: string]: Environment }`), not a list.

ui/src/services/alertmanager/api.js
```diff
@@ -1,11 +1,35 @@
+//@flow
 import ApiClient from '../ApiClient';
 
-let alertmanagerApiClient = null;
+let alertmanagerApiClient: ApiClient = null;
```
gdemonet: Shouldn't it be a `Maybe` type?

```javascript
let alertmanagerApiClient: ?ApiClient = null;
```

ui/src/services/alertmanager/api.js
```diff
@@ -1,11 +1,35 @@
...
-export function initialize(apiUrl) {
+export function initialize(apiUrl: string) {
   alertmanagerApiClient = new ApiClient({ apiUrl });
 }
 
-export function getAlerts() {
+export type PrometheusAlert = {
+  "annotations": {
+    [key: string]: string,
+  },
```
gdemonet: OK for now, but we'll need to type some of the expected keys (e.g. `description`, `message`, ...). We've been hit by a missing key recently (see #2994).

ui/src/ducks/app/alerts.js
```diff
@@ -35,7 +42,7 @@ export const fetchAlertsAlertmanagerAction = () => {
   return { type: FETCH_ALERTS_ALERTMANAGER };
 };
 
-export const updateAlertsAlertmanagerAction = (payload) => {
+export const updateAlertsAlertmanagerAction = (payload: $Shape<AlertsState>) => {
```
gdemonet: Oh, neat type 😄 , I guess we'll be able to use it for many "update" action creators.

ui/src/ducks/config.js
```diff
@@ -56,7 +59,26 @@ const defaultState = {
   themes: {}, // include light, dark and custom
 };
 
-export default function reducer(state = defaultState, action = {}) {
+export type ConfigState = {
+  language: string,
+  theme: Theme,
+  api: Config | null,
```
gdemonet: `?Config` would work here as well, IIUC

ui/src/ducks/config.js
```diff
@@ -100,48 +122,50 @@ export function fetchConfigAction() {
...
 
-export function setThemesAction(themes) {
+export function setThemesAction(themes: Themes) {
   return { type: SET_THEMES, payload: themes };
 }
 
 // Selectors
-export const languageSelector = (state) => state.config.language;
-export const apiConfigSelector = (state) => state.config.api;
+export const languageSelector = (state: RootState) => state.config.language;
```
gdemonet, line 171: Just a question: since `RootState` is properly typed, is flow able to infer the return type of these selectors?

ui/src/ducks/login.js
```diff
@@ -28,23 +36,19 @@ export default function reducer(state = defaultState, action = {}) {
...
 };
 
 // Sagas
-export function* authenticateSaltApi() {
-  const api = yield select(apiConfigSelector);
-  const user = yield select(state => state.oidc.user);
-  const result = yield call(ApiSalt.authenticate, user);
-  if (!result.error) {
+export function* authenticateSaltApi(): Generator<any, void, any> {
+  const api: Config | null = yield select(apiConfigSelector);
```
gdemonet: (ugh, even if flow was able to infer the return type of a selector, it can't guess the value injected by the coroutine's "owner"... some big limitation of sagas here)

ui/src/ducks/login.js
```diff
@@ -28,23 +36,19 @@ export default function reducer(state = defaultState, action = {}) {
...
+  const user: User = yield select((state: RootState) => state.oidc?.user);
+  const result: { error: any } | ApiSalt.SaltToken = yield call(ApiSalt.authenticate, user);
```
gdemonet: Could be handy to have type aliases for such "result" values:

```javascript
declare type Result<T> = { error: any } | T

// Or for API results under a `body` key
declare type APIResult<T> = Result<{ body: T }>
```

Here we'd have:

```javascript
const result: Result<Api.SaltToken> = yield call(...)
```

ui/src/ducks/reducer.js
```diff
@@ -29,4 +38,26 @@ const rootReducer = combineReducers({
...
+    salt: SaltState,
+    monitoring: MonitoringState,
+    alerts: AlertsState
+  }
+}
+
+export const useTypedSelector: <TSelected>(
+  selector: (state: RootState) => TSelected,
+  equalityFn?: (left: TSelected, right: TSelected) => boolean
+) => TSelected = useSelector;
```
gdemonet: Two comments:

- I think we have defined custom hooks in `src/services/utils.js`(though granted, it doesn't make much sense either) - IMHO, adding a plain `src/hooks.js` could make it easier to find in the future
- the purpose of this custom hook is not obvious at first glance, so I'd suggest we add a comment to explain it (and remind the reader that they should use it instead of the standard `useSelector` hook)

gdemonet (changes requested):
Super cool stuff, thanks for starting this heavy lifting, I'm confident this will increase productivity a lot!
Some comments and suggestions here and there 😉 , just adding a more general one here: are you running `Prettier`? We have a committed `.prettierrc` for you to use, the goal being to keep the code consistent without spending time on aesthetics in PR reviews 😜

ui/src/ducks/app/volumes.js
JBWatenbergScality: Thanks I was looking for those typings desperately 👍  👍

ui/src/ducks/config.js
JBWatenbergScality: Yes, flow is inferring the type of `languageSelector` from `state.config.language`

ui/src/ducks/login.js
JBWatenbergScality: Yes the simplest way I found to type complex saga with different return type over iterations is to type `Next` as any and  force the type of each of the variables 😞 . Even through this is not directly linked to redux but more to typing, IMO redux and redux saga are adding a lot of complexity for such simple tasks as data fetching, I prefer using libraries such as  `react-query` or `swr` that greatly ease data fetching and caching use cases and that are way more easier to type as in most cases the typing would be inferred from API clients typing.

ui/src/services/alertmanager/api.js
JBWatenbergScality: if they are always present in the alerts I can add typing for `description` and `summary`. I'm just unsure Prometheus alert manager will always send those keys as the documentation is not mentioning this.

ui-metalk8s-client-generator/README.md
```diff
@@ -0,0 +1,23 @@
+## Metalk8s client generator
+
+It generates a flow client based on `@kubernetes/client` custom objects client from kubernetes `CustomResourceDefinition` of metalk8s volumes.
```
gdemonet, line 3: Awesome :+1:

ChengYanJin (dismissed):
LGTM! 
Can't wait to get it to merge and benefit from the typing.

gdemonet (changes requested):
Could we squash the commits into logical units, now that reviewers approved? I'm not too fond of `fix stuff` commits, especially for fixing things introduced in the same PR.

ui/src/services/alertmanager/api.js
gdemonet: No they are optional indeed, so not sure if Flow can help in any way here.

ui-metalk8s-client-generator/src/generateClient.ts
```diff
@@ -55,6 +55,22 @@ export function generateClient(crdFile: string, destinationFile: string) {
...
+      return await customObjects.getClusterCustomObject(
+        '${crdSpec.spec.group}',
+        '${crdSpec.spec.version}',
+        '${crdSpec.spec.names.plural}',
+        ${singleName}Name,
+      );
+    } catch (error) {
+      return { error };
+    }
+  }
```
gdemonet: Maybe this part should be squashed with `ui: add a flow typed client generator` (it's more than just `ui: use generated metalk8s client and types`)

gdemonet (dismissed):

ChengYanJin (dismissed):

ChengYanJin (dismissed):

ChengYanJin (approved):

PR #3003, opened 2020-12-28
salt: Avoid duplicating static pod manifests

### LINKED ISSUE #2840: Salt master not deploy correctly because another salt master is already running.
**Component**: salt

**What happened**:
On a fresh install, `salt-master` is sometime not deployed correctly because another salt-master is already running a using the port.

```
[root@boostrap-test centos]# kubectl get pods -n kube-system
NAME                                              READY   STATUS             RESTARTS   AGE
apiserver-proxy-boostrap-test.novalocal           1/1     Running            0          78m
[...]
salt-master-boostrap-test.novalocal               1/2     CrashLoopBackOff   25         2m39s
storage-operator-7b54589795-j89p8                 1/1     Running            0          78m

[root@boostrap-test centos]# kubectl describe pods salt-master-boostrap-test.novalocal -n kube-system
[...]
  Type     Reason   Age                   From                              Message
  ----     ------   ----                  ----                              -------
  Normal   Pulled   46m (x11 over 78m)    kubelet, boostrap-test.novalocal  Container image "metalk8s-registry-from-config.invalid/metalk8s-2.6.0-dev/salt-master:3000.3-1" already present on machine
  Warning  BackOff  103s (x359 over 78m)  kubelet, boostrap-test.novalocal  Back-off restarting failed container
```

```
[root@boostrap-test centos]# kubectl logs salt-master-boostrap-test.novalocal salt-master -n kube-system
/usr/lib/python2.7/site-packages/salt/scripts.py:109: DeprecationWarning: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date.  Salt will drop support for Python 2.7 in the Sodium release or later.
[INFO    ] Setting up the Salt Master
[WARNING ] Unable to bind socket 10.200.4.17:4505, error: [Errno 98] Address already in use; Is there another salt-master running?
[INFO    ] The Salt Master is shut down
The salt master is shutdown. The ports are not available to bind
```

After some checking we figure out that this come from a rogue salt-master that escaped metalk8s tracking.
(container is still running, but not shown by kubectl)

```
[root@boostrap-test centos]# crictl ps -a
CONTAINER ID        IMAGE               CREATED              STATE               NAME                            ATTEMPT             POD ID
20e46184f3143       b0139962ff652       About a minute ago   Exited              salt-master                     28                  7252df6f59ddb
[...]
0b3f2a8cd512f       b0139962ff652       2 hours ago          Running             salt-api                        0                   7252df6f59ddb
cae6b3d0b08f4       b0139962ff652       2 hours ago          Running             salt-master                     0                   7252df6f59ddb
79e8c98f5a32d       f09fe80eb0e75       2 hours ago          Running             repositories                    0                   7039f0ae74cfb
```

Stoping and removing both `salt-master` containers solve the issue.

**What was expected**:
When you deployed a fresh bootstrap, `salt-master` should be deployed correctly.

**Steps to reproduce**
After some discussions with @slaperche-scality , this flaky can also happen when you `deploy/undeploy` a solution.

So the best way to reproduce this is to `deploy` and `undeploy` a (complex ?) solution.

It is maybe a bug when salt is restarting.

gdemonet:
When using `metalk8s.static_pod_managed`, we call `file.managed` behind
the scenes. This state does a lot of magic, including creating a
temporary file with the new contents before replacing the old file.
This temp file gets created **in the same directory** as the managed
file by default, so it gets picked up by `kubelet` as if it were
another static Pod to manage. If the replacement occurs too late,
`kubelet` may have already created another Pod for the temp file, and
may not be able to "remember" the old Pod, hence not cleaning it up.
This results in "rogue containers", which can create issues (e.g.
preventing new containers from binding some ports on the host).

This commit reimplements the 'file.managed' state in a minimal fashion,
to ensure the temporary file used for making an "atomic replace" is
ignored by kubelet. Note that it requires us to also reimplement the
'file.manage_file' execution function, since it always relies on the
existing "atomic copy" operation from `salt.utils.files.copyfile`.

Fixes: #2840

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -3,17 +3,20 @@
 Module for handling MetalK8s specific calls.
 '''
 import functools
+import itertools
 import logging
 import os.path
 import re
 import six
 import socket
+import tempfile
 import time
 
 from salt.pillar import get_pillar
 from salt.exceptions import CommandExecutionError
 import salt.utils.args
 import salt.utils.files
+from salt.utils.hashutils import get_hash
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


salt/_states/metalk8s.py
```diff
@@ -76,6 +76,7 @@ def static_pod_managed(name,
             context or {},
             config_digest=config_digest, metalk8s_version=metalk8s_version
         ),
+        tmp_dir=kwargs.pop("tmp_dir", "/tmp"),
```
slaperche-scality: Instead of hardcoding `/tmp`, you can probably use [`tempfile.gettempdir`](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) which would be more reliable.

gdemonet: Done.

gdemonet:
Received a comment from someone in PTO (😅 ), which points out that writing the temporary file to another directory (e.g. `/tmp`) may break the atomicity of replacing the original file (if `/tmp` is not on the same device as `/etc/kubernetes/manifests`, which is very likely for `tmpfs`).

Since `kubelet` ignores Pod manifests with filenames starting with a dot (see [here](https://github.com/kubernetes/kubernetes/blob/0a839c6c3b9489716e216b4fc9dd2f9c17948c16/pkg/kubelet/config/file.go#L162)), I'll change the approach to use this behaviour instead.

gdemonet:
After investigating what `file.managed` does behind the scenes (mostly calling out to the `file.manage_file` execution method), two things arose:

- `tmp_dir` would not have been used anyway, because this only serves for the `check_cmd` (which we don't use)
- given the log we get from `kubelet` when this error happens:

```
Unable to process watch event: can't process config file "/etc/kubernetes/manifests/repositories.yamlcsl_iuli": open /etc/kubernetes/manifests/repositories.yamlcsl_iuli: no such file or directory
```

It appears this file is created by the call to `salt.utils.files.mkstemp`, with `prefix` set to the destination filename and `dir` to the destination directory. This call is made from `salt.utils.files.copyfile`, and there is no way around this behaviour.

In conclusion, I'll need to write a custom module for managing these manifests as we want 😞 .

salt/_states/metalk8s.py
```diff
@@ -76,6 +77,7 @@ def static_pod_managed(name,
             context or {},
             config_digest=config_digest, metalk8s_version=metalk8s_version
         ),
+        tmp_dir=kwargs.pop("tmp_dir", tempfile.gettempdir()),
```
TeddyAndrieux: Minor, maybe deserve a small comment :smile:

gdemonet: This line is going away, I'm going to push the latest code so you can have a look :)

salt/_modules/metalk8s.py
```diff
@@ -355,3 +358,164 @@ def cmp_sorted(*args, **kwargs):
...
+    base_name = os.path.basename(dest)
+    dir_name = os.path.dirname(dest)
+
+    with open(source, mode='rb') as src_file:
+        contents = src_file.read()
+
+    with tempfile.NamedTemporaryFile(
+        prefix=tmp_prefix, dir=dir_name, delete=False,
+    ) as tmp_file:
+        tmp_file.write(contents)
```
NicolasT, line 391: You'll want to `fsync` `tmp_file`.

salt/_modules/metalk8s.py
```diff
@@ -355,3 +358,164 @@ def cmp_sorted(*args, **kwargs):
...
+
+    try:
+        os.replace(tmp_file.name, dest)
+    except OSError as exc:
```
NicolasT: No need to capture the exception (`as exc`).

salt/_modules/metalk8s.py
```diff
@@ -355,3 +358,164 @@ def cmp_sorted(*args, **kwargs):
...
+                    ret["changes"]["diff"] = exc.strerror
+
+    else:  # target file does not exist
+        ret["changes"]["diff"] = "New file"
+        real_name = name
+
+    if ret["changes"] and not __opts__["test"]:
+        # The file needs to be replaced
+        try:
+            _atomic_copy(source_filename, real_name)
```
NicolasT: Wouldn't it make sense for `_atomic_copy` *not* to have an arbitrary default for `tmp_prefix` and instead pass one explicitly from here?

salt/_modules/metalk8s.py
NicolasT, line 391: `manage_static_pod_manifest` sets some permissions on the final file, after the file has been written. However, this means we potentially/likely have a file with different permissions at some point in time, in which target data is written.

As such, I'd suggest to set the perms etc. on `tmp_file` before writing any data into it.

salt/_states/metalk8s.py
```diff
@@ -41,44 +52,96 @@ def static_pod_managed(name,
...
+        )
+    except Exception as exc:  # pylint: disable=broad-except
+        log.debug(traceback.format_exc())
+        return _error(ret, "Unable to get managed file: {}".format(exc))
+
+    if comment_:
+        return _error(ret, comment_)
+    else:
+        try:
+            return __salt__["metalk8s.manage_static_pod_manifest"](
```
NicolasT, line 133: Thinking of it, shouldn't this be called `static_pod_manifest_managed` similar to `file.managed`?

salt/_states/metalk8s.py
```diff
@@ -1,28 +1,39 @@
 """Custom states for MetalK8s."""
 
+import logging
+import tempfile
```
NicolasT: Is this used? Seems surprising in a state module.

NicolasT (changes requested):

NicolasT:
May want to update PR description with current commit message to avoid confusion.

salt/_states/metalk8s.py
TeddyAndrieux: This module is a "copy" of salt `file.manage_file` so for me the name is good here (especially since it's an execution module and not a state)

NicolasT: Argh good point!

salt/_modules/metalk8s.py
gdemonet: It's flushed on `.close()`, which the contextmanager takes care of, IIUC.

See:
- https://github.com/python/cpython/blob/master/Lib/tempfile.py#L427-L434
- https://github.com/python/cpython/blob/master/Lib/tempfile.py#L486-L491

But I can add a `tmp_file.flush()` if you think that's required.

salt/_modules/metalk8s.py
gdemonet: Sure, if we want it to be less specific about our use-case. I'll make this change.

salt/_modules/metalk8s.py
gdemonet: Right, good point, so I guess we only need the `file.check_perms` call if the source and target checksums match.

NicolasT: From the comments in the state implementation, I have the impression `ret` is passed to `check_perms` which is idempotent, so that can stay as-is. I'd set the file perms/owner to whatever is later passed to `file.check_perms` later on, at this point.

NicolasT: A `close` (and `flush`, for buffered IO) doesn't `fsync` data to disk.

salt/_states/metalk8s.py
gdemonet: I can rename this state function `static_pod_manifest_managed` (from `static_pod_managed`) if you want, but this line is about the execution function (which mirrors `file.manage_file` in principle).

salt/_states/metalk8s.py
gdemonet: No indeed, good catch.

salt/_modules/metalk8s.py
gdemonet: OK, adding `os.fdatasync(tmp_file.fileno())` in that case.

gdemonet: However, it means I need to compare the desired perms/owner with the existing file if there is one (otherwise `check_perms` would not add anything to `ret["changes"]`).

NicolasT: How so? `check_perms` sets the uid/gid/attrs to the desired ones, unless they're already set to the desired ones (in which cases there are no 'changes' reported). This is unconditionally called in `manage_static_pod_manifest` as using root/root/0600. Now, in this function *we are always creating a new file*, so there *are* 'changes' (new file created). We don't really need/want `check_perms` to 'add' more changes since we can as well create this new file using the attributes we want it to have anyway.

So, the 'changes' of the `manage_static_pod_manifest` call are either 'file created/replaced (using the content and attrs we want it to have)', or 'file not modified, but existing file attrs changed to desired values because someone changed them to something else for whatever reason', or 'no changes (file contents and attrs are what they should be)'.

NicolasT: You want `fsync`, not `fdatasync`, since with `fdatasync` the perms you set before are not guaranteed to be synced to disk.

gdemonet: Ha, my bad I didn't know the difference, thanks!

NicolasT (approved):

PR #3008, opened 2021-01-05
Init development/2.8

TeddyAndrieux:
**Component**:

'kubernetes', 'build'

**Context**: 

Initialization of the branch development/2.8

**Summary**:

- Bump MetalK8s version
- Bump Kubernetes version to 1.19.6
- Bump coredns version to 1.7.0
- Bump etcd version to 3.4.13-0
- Deactivate downgrade test for minor version in the CI (See: #1750)

---


eve/main.yml
```diff
@@ -753,7 +753,9 @@ stages:
           name: Trigger upgrade and downgrade test stages simultaneously
           stage_names:
             - single-node-upgrade-centos
-            - single-node-downgrade-centos
+            # NOTE: Deactivate minor downgrade on this branch
+            # See: https://github.com/scality/metalk8s/issues/1750
+            #- single-node-downgrade-centos
```
gdemonet, line 758: We upgrade from 3.4.3 to 3.4.13, shouldn't downgrade work for patch versions? (I tried looking around their docs, but there doesn't seem to be mentions of downgrade)

gdemonet (approved):

eve/main.yml
TeddyAndrieux: Hmm right, worth a try, I will merge this PR as is for the moment and I will check this :+1:

TeddyAndrieux: Btw I also forgot to disable the downgrade in salt in the defaults.yaml so anyway I will do another PR for downgrade (either to really disable it or to enable it back :smile: )

TeddyAndrieux: Patch version etcd downgrade is supported (at least from 3.4.13 to 3.4.3) so this commit get reverted in #3023

PR #3010, opened 2021-01-05
Clean up development/2.8

TeddyAndrieux:
**Component**:

'salt', 'build'

**Context**: 

Clean up no longer needed stuff in development/2.8

**Summary**:

NOTE: Upgrade and downgrade only support one minor version it means we can only upgrade in 2.8 from MetalK8s 2.7.0+ and downgrade from 2.8 to MetalK8s 2.7.0+

- Remove workaround for loop device (since restored snapshot can only be MetalK8s 2.7.0+)
- Remove upgrade/downgrade workaround for nginx-ingress renaming (MetalK8s 2.7.0+ already contains the renamed nginx-ingress)
- Remove salt migration orchestrate (MetalK8s 2.7.0+ already run with Salt Python 3)
- Clean up downgrade orchestrate (MetalK8s 2.7.0+ already run with Salt Python 3)

---


gdemonet (approved):
:ok_hand:

PR #3018, opened 2021-01-08
ci/ui-integration: Increase app wait timeout

gdemonet:
Before starting the Cypress test suite, we try to reach the sidecar
container running the application under test.
This timeout was set to 2 minutes, which sometimes fails (because things
are too slow; we can know this by seeing the Cypress suite succeeding
after such timeout). We bump it to 5 minutes.
We also ensure such timeouts will prevent the suite from being executed
further.


ChengYanJin (approved):
LGTM!

PR #3020, opened 2021-01-08
Ensure containerd is ready before importing container image

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Time to time importing of pause image fail because containerd just get upgraded and not fully restarted

```
         ID: Inject pause image
    Function: containerd.image_managed
        Name: k8s.gcr.io/pause:3.1
      Result: False
     Comment: Failed to import archive: time="2021-01-08T10:22:26+01:00" level=debug msg="unpacking 1 images"
     Started: 10:22:26.698077
    Duration: 168.411 ms
     Changes: 
```

**Summary**:

- Add a salt module to check for container engine to be ready (using `crictl version`)
- Make sure container engine is ready before importing pause image
- Make sure container engine is ready before importing nginx image

---

### DIFF
--- a/salt/metalk8s/container-engine/containerd/installed.sls
+++ b/salt/metalk8s/container-engine/containerd/installed.sls
@@ -10,6 +10,7 @@
 
 include:
   - metalk8s.repo
+  - .running
 
 {%- if grains['os_family'].lower() == 'redhat' %}
 Install container-selinux:
@@ -36,6 +37,8 @@ Install containerd:
       {%- if grains['os_family'].lower() == 'redhat' %}
       - metalk8s_package_manager: Install container-selinux
       {%- endif %}
+    - watch_in:
+      - service: Ensure containerd running
 
 Create containerd service drop-in:
   file.managed:
@@ -69,11 +72,15 @@ Create containerd service drop-in:
           HTTPS_PROXY: "{{ proxies.https }}"
           {%- endif %}
         {%- endif %}
+    - watch_in:
+      - service: Ensure containerd running
 
 Install and configure cri-tools:
   {{ pkg_installed('cri-tools') }}
     - require:
       - test: Repositories configured
+    - require_in:
+      - test: Ensure containerd is ready
   file.serialize:
     - name: /etc/crictl.yaml
     - dataset:
@@ -84,6 +91,8 @@ Install and configure cri-tools:
     - group: root
     - mode: '0644'
     - formatter: yaml
+    - require_in:
+      - test: Ensure containerd is ready
 
 Configure registry IP in containerd conf:
   file.managed:
@@ -95,3 +104,5 @@ Configure registry IP in containerd conf:
 
         [debug]
         level = "{{ 'debug' if metalk8s.debug else 'info' }}"
+    - watch_in:
+        - service: Ensure containerd running
--- a/salt/metalk8s/container-engine/containerd/running.sls
+++ b/salt/metalk8s/container-engine/containerd/running.sls
@@ -1,0 +1,21 @@
+# NOTE: This state just ensure containerd running and "ready"
+# To restart containerd you need to use a `watch_in` in another state 
+# that include this one
+# ```
+# - watch_in:
+#   - service: Ensure containerd running
+# ```
+
+Ensure containerd running:
+  service.running:
+    - name: containerd
+    - enable: True
+    - init_delay: 2
+
+Ensure containerd is ready:
+  test.configurable_test_state:
+    - changes: False
+    - result: __slot__:salt:cri.ready()
+    - comment: Ran 'cri.ready'
+    - require:
+      - service: Ensure containerd running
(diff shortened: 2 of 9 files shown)


gdemonet (dismissed):

salt/metalk8s/container-engine/containerd/running.sls
```diff
@@ -0,0 +1,21 @@
...
+# ```
+
+Ensure containerd running:
+  service.running:
+    - name: containerd
+    - enable: True
+    - init_delay: 2
+
+Ensure containerd is ready:
+  test.configurable_test_state:
```
NicolasT, line 16: Any reason `cri.ready` is not implemented as some module?

salt/metalk8s/repo/installed.sls
```diff
@@ -18,11 +18,14 @@
...
 include:
   - .configured
+  - metalk8s.container-engine.running
 
 Inject nginx image:
   containerd.image_managed:
     - name: {{ image_fullname }}
     - archive_path: {{ archives[saltenv].path }}/images/{{ image_name }}-{{ image_version }}.tar
+    - require:
+      - test: Container engine is ready
```
NicolasT: This doesn't exist if `kubelet.container_engine` is `false` (`test: No container engine to check` would).

salt/metalk8s/container-engine/containerd/running.sls
gdemonet: Would need to create a state module in addition to wrap this call. Using slots looks simpler to me :)

NicolasT (dismissed):
I believe there's a little bug (likely not something we'd ever run into, but still).

NicolasT:
Is this in any way related to https://github.com/scality/metalk8s/pull/2960 ?

salt/metalk8s/container-engine/containerd/running.sls
TeddyAndrieux: WDYM ? As a salt state module ?

For me, since it's something that does not change anything on the machine we do not want to have a salt state module for this.

salt/metalk8s/repo/installed.sls
TeddyAndrieux: Hmm you right if container engine is not define I think we should safely ignore the "container engine is ready". I will require the full sls

TeddyAndrieux:
> I believe there's a little bug (likely not something we'd ever run into, but still).

Ups, I forgot to add the new sls in the iso :smile: 

> Is this in any way related to #2960 ?

Yep I will close this one

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -69,11 +72,15 @@ Create containerd service drop-in:
           HTTPS_PROXY: "{{ proxies.https }}"
           {%- endif %}
         {%- endif %}
+    - watch_in:
+      - service: Ensure containerd running
```
NicolasT, line 76: Shouldn't this result in a `containerd` restart, and even more, a `systemctl daemon-reload`?

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -36,6 +37,8 @@ Install containerd:
       {%- if grains['os_family'].lower() == 'redhat' %}
       - metalk8s_package_manager: Install container-selinux
       {%- endif %}
+    - watch_in:
+      - service: Ensure containerd running
```
NicolasT, line 41: Similar as in other places, will the service be *re*started when the package is upgraded?

salt/metalk8s/container-engine/containerd/installed.sls
```diff
@@ -95,3 +104,5 @@ Configure registry IP in containerd conf:
 
         [debug]
         level = "{{ 'debug' if metalk8s.debug else 'info' }}"
+    - watch_in:
+        - service: Ensure containerd running
```
NicolasT, line 108: Shouldn't this cause a service restart?

salt/metalk8s/container-engine/containerd/installed.sls
TeddyAndrieux: Yes it does, and it's the case with this `watch_in`

salt/metalk8s/container-engine/containerd/installed.sls
TeddyAndrieux: Yes

salt/metalk8s/container-engine/containerd/installed.sls
TeddyAndrieux: It does, it's why I use `watch_in` and not `require_in`

When you do a `service.running` and you watch a state that do some changes then the service is restarted

gdemonet (approved):
LGTM

salt/metalk8s/container-engine/containerd/installed.sls
NicolasT: Thanks!

PR #3028, opened 2021-01-12
tests: Raise volume client retry delay to 4s

alexandre-allard:
**Component**: tests

**Context**: 
Some volume tests are flaky because we probably do not wait enough time.  

**Summary**:
We double the delay between each try (4s instead of 2s), so we wait 2 min in total.

**Acceptance criteria**: At least no flakies on volume tests. :)

### DIFF
--- a/tests/kube_utils.py
+++ b/tests/kube_utils.py
@@ -253,7 +253,7 @@ class Client(abc.ABC):
 class VolumeClient(Client):
     def __init__(self, k8s_client, ssh_config):
         super().__init__(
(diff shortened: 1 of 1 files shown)


gdemonet (approved):
There are some other `utils.retry` in `tests/post/steps/test_volume.py` as well, do they need to be bumped (not sure if I saw flakiness for these)?

alexandre-allard:
@gdemonet Can't remember if I saw other flakies (not related to the certificates tests) than this one.
Let's merge this and we will see if we need another PR for the rest.

Issue #3026, opened 2021-01-12
Embed checksum in ISO so its integrity can be checked


NicolasT:
We currently ship a `SHA256SUM` file with the ISO build artifact, which is a well-understood way to check the integrity of files, e.g., after downloading.

RedHat/Fedora install media, which are also ISOs (as software files or on burnt media) have an embedded checksum for the data segments in an otherwise unused sector of the ISO. This allows for the integrity of the (data on the) ISO to be validated without requiring an 'external' source of information (e.g., such `SHA256SUM` file). The `implantisomd5` and `checkisomd5` tools (from [isomd5sum](https://github.com/rhinstaller/isomd5sum)) can be used to achieve this.

Once we embed this in our ISOs, the `import-iso.sh` script could be extended to validate the integrity of a MetalK8s (or solution) ISO before continuing the import process.

NicolasT:
#3032

PR #3032, opened 2021-01-13
buildchain: embed ISO MD5 in the generated image

NicolasT:
The `isomd5sum` tools allow to embed a hash of data sectors found in an
ISO file in an otherwise unused section, hence allowing the integrity of
(the data sectors of) an ISO image to be checked. This is, e.g., also
done for RHEL/CentOS/Fedora ISOs (their integrity can be checked at boot
time).

This commit adds a call to `implantisomd5` at the end of the buildchain,
right after the ISO is created and before its SHA256 is calculated.

Given this, one can run `checkisomd5` on a resulting ISO file after
download to ensure it's not corrupted.

Fixes: #3026
See: https://github.com/scality/metalk8s/issues/3026
See: https://github.com/rhinstaller/isomd5sum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 
 - Bump etcd version to 3.4.13-0 (PR[#3008](https://github.com/scality/metalk8s/pull/3008))
 
+- [#3026](https://github.com/scality/metalk8s/issues/3026) - Embed a checksum
+  of the data contained in the ISO image inside the ISO so its integrity can be
+  ensured after download, next to or instead of checking the `SHA256SUM` using
+  `checkisomd5` (from [isomd5sum](https://github.com/rhinstaller/isomd5sum))
+  (PR [#3032](https://github.com/scality/metalk8s/pull/3032))
+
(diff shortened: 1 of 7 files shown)


slaperche-scality (dismissed):

gdemonet (commented):
Maybe worth adding a mention of this `isomd5sum` package in the developer docs (under `docs/developer/building/requirements.rst`)?

NicolasT:
As suggested (@gdemonet) I added the new requirement to the developer documentation. Also extended the Eve step which currently only tests integrity using the `SHA256SUM` file to also use `checkisomd5` if there's some sign the ISO contains the relevant metadata (`checkisomd5` errors out with the same exit code both when there's a checksum mismatch, as well as when no checksum is embedded, so no way to dispatch). The detection mechanism is... 'crude', but shouldn't result in false positives. It could be removed later once all ISOs we use in CI (ideally starting with 2.9?) contain the checksum.

gdemonet (dismissed):

slaperche-scality (dismissed):

gdemonet (approved):

Issue #3022, opened 2021-01-11
Salt-master pod (on bootstrap node) not starting if salt-minion failed to start


thomasdanan:
**Component**:

'salt', 'bootstrap'

**Why this is needed**:

If for some reason, salt-minion service failed to start on `bootstrap` node, the pod `salt-master` will fail to start with the following error:

`Warning  FailedMount  4s (x12 over 8m18s)  kubelet  MountVolume.SetUp failed for volume "run" : hostPath type check failed: /var/run/salt is not a directory
[root@zenkotda-bootstrap scality]# systemctl restart kubelet`

**What should be done**:

salt-master pod should be able to start even if salt-minion is not running (i.e. `/var/run/salt` is not available)

**Implementation proposal** (strongly recommended):

**Test plan**:

NicolasT:
#3041

NicolasT:
#3041

PR #3041, opened 2021-01-14
salt: Create `/var/run/salt` directory in salt-master manifests

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

#3022 

**Summary**:

Salt process create a `/var/run/salt` directory at startup this
directory is shared between salt-master container and salt-minion
running on the host so it's mounted in the container, since this
directory does not exists at system boot we need to use
`DirectoryOrCreate` so that salt-master does not need salt-minion to
start before being able to start

---

Fixes: #3022

### DIFF
--- a/salt/metalk8s/salt/master/files/salt-master-manifest.yaml.j2
+++ b/salt/metalk8s/salt/master/files/salt-master-manifest.yaml.j2
@@ -118,7 +118,7 @@ spec:
     - name: run
       hostPath:
         path: '/var/run/salt'
-        type: Directory
+        type: DirectoryOrCreate
     {%- for env, archive in archives.items() | sort(attribute='0') %}
     - name: states-{{ env | replace('.', '-') }}
       hostPath:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 
 - Bump etcd version to 3.4.13-0 (PR[#3008](https://github.com/scality/metalk8s/pull/3008))
 
+### Bug fixes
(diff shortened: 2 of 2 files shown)


gdemonet (approved):

gdemonet (commented):
Actually, just checked the `DirectoryOrCreate` docs:

> If nothing exists at the given path, an empty directory will be created there as needed with permission set to 0755, having the same group and ownership with Kubelet

Isn't this going to cause problems? Or we don't care because it's always `root:root` for both?

TeddyAndrieux:
> Isn't this going to cause problems? Or we don't care because it's always `root:root` for both?

Today both run as root so we don't really care

salt/metalk8s/salt/master/files/salt-master-manifest.yaml.j2
```diff
@@ -118,7 +118,7 @@ spec:
     - name: run
       hostPath:
         path: '/var/run/salt'
-        type: Directory
+        type: DirectoryOrCreate
```
NicolasT, line 121: Not super fond of that. Can't we/shouldn't we instead create the directory on the host as part of the Salt state which lays out the manifest?

NicolasT: Follow-up: `/var/run` is `tmpfs` so not persisted. Hence, `DirectoryOrCreate` *is* the right approach.

TeddyAndrieux: For this specific one, no, we can't since it's part of tmpfs `/var/run`

NicolasT (approved):

PR #3044, opened 2021-01-15
dependencies: Configure dependabot to autoupdates core-ui version

JBWatenbergScality:
**Component**:

ui, dependencies

**Context**: 

We would like core-ui version to be automatically updated in all the projects that makes usse of it in order to benefit from latest fixes and feature.
Configuring Dependabot to do so would enable us to automatically get pull requests on ui projects as soon as a new version of core ui is available, easing communication of new version availabe accross UI chapter and enabling each squad to easily test their UI with the latest core-ui version.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/ui"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "@scality/core-ui"
+


NicolasT:
How does this impact the 'default' Dependabot configuration? As in, with this file in place, will we still get notifications about other NPM packages used in `/ui`, and, more importantly, notifications about dependencies in some Python `requirements.txt`, Golang `go.sum` and whatnot?

JBWatenbergScality:
I just checked and github security updates and dependabot version updates are 2 different things with distinct configuration even if behind the hoods they both use dependabot as their core. Github security updates are still performed despite this version update configuration. 
I activated automatic security updates and version updates with a similar configuration on this repository https://github.com/JBWatenbergScality/dependabot-test and despite allowing version updates only on @scality/core-ui dependency I got another pull request that solve a security issue on axios : https://github.com/JBWatenbergScality/dependabot-test/pulls

NicolasT (approved):

PR #3049, opened 2021-01-18
docs: Explain how to create a ClusterRoleBinding on groups

### LINKED ISSUE #3043: Update Default Dex configuration to allow group fetching
**Component**:

'dex'

**Why this is needed**:

Currently, we cannot properly define a clusterrole binding between a group and a cluster role when dex is integrated with an external OIDC IDP. Actually, the default configuration of dex we deploy doesn't allow it. Also, the documentation we provide to guide the user when creating clusterrolebinding is focusing on user to cluster-role binding but one other important use case would be between group and cluster-role. We should add an example for such configuration.

**What should be done**:

- update default dex configuration to allow group fetch from external OIDC
- update the documentation to explain how to create a clusterrolebinding between group and cluster-role

**Test plan**:

alexandre-allard:
**Component**: docs

**Context**:
We need to be able to bind Keycloak group to role in K8s.

**Summary**:
Explain how to create a ClusterRoleBinding on a group.

**Acceptance criteria**: n/a

---

Closes: #3043

### DIFF
--- a/docs/operation/account_administration.rst
+++ b/docs/operation/account_administration.rst
@@ -158,9 +158,38 @@ these steps:
       root@bootstrap $ kubectl --kubeconfig /etc/kubernetes/admin.conf \
                          apply -f role_binding.yaml
 
+.. _bind-group-to-role:
+
+To bind one or more groups to an existing ClusterRole in all namespaces, follow
+these steps:
+
+#. Create a ClusterRoleBinding manifest (:file:`role_binding.yaml`) from the
+   following template:
+
+   .. code-block:: yaml
+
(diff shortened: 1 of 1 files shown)


gdemonet (approved):
LGTM. Just wondering if we thought about adding a prefix to OIDC users (using `--oidc-groups-prefix` and `--oidc-username-prefix`), which is recommended to avoid conflicts with other authentication methods.

alexandre-allard:
@gdemonet No, but that's a good idea, I'll create an issue to track this.

NicolasT:
We *must* use such prefix, otherwise there are some glaring security issues. That's a blocker (I though we were already having those...)

PR #3050, opened 2021-01-18
Properly fail if a conflicting package is installed and do not install conflicting packages

### LINKED ISSUE #2992: Install (rightfully) fails when Docker is installed, but this is hard to troubleshoot
When `docker` is installed on a host, the installation fails (and should fail). `docker` has a `Provides` for `runc` which we want to install, but it's not a 'good' version of `runc` (and `containerd`). At some later point in time, install will then fail for obscure reasons.

https://github.com/scality/metalk8s/pull/2960 could help to make the late error a bit more clear, but it's still difficult to link this (as a user) to `docker` being installed on the system.

As such, it likely makes sense to keep some kind of blacklist of 'known incompatible' packages, and ensuring these are not installed, either in a Salt state (somehow?), or in `bootstrap.sh` (though then the same checks would not be performed on non-bootstrap nodes...)

Known conflicting packages:
- `docker`
- `containerd.io`
- Alternative Salt versions?
- Also see https://github.com/scality/metalk8s/discussions/2989#discussioncomment-224790

Also suggested by @Baitanik in https://github.com/scality/metalk8s/discussions/2989#discussioncomment-222008 and reported earlier by @vrancurel.

TeddyAndrieux:
**Component**:

'salt', 'check', 'deployment'

**Context**: 

#2992 

**Summary**:

- Add an execution module to retrieve value from `map.jinja`
- Check for conflicting package installed during bootstrap and expansion
- Exclude conflicting packages when installing package in salt (e.g: we do not install `containerd.io` instead of `containerd` and `runc`) (NOTE that before this fix when running `containerd` SLS if `containerd.io` is available we install this one instead of `runc` and `containerd`)

**Acceptance criteria**: 

Bootstrap output when docker-ce installed on the host
```
# /srv/scality/metalk8s-2.8.0-dev/bootstrap.sh 
> Determine the OS... done [0s]
> Checking that BootstrapConfiguration is present... done [0s]
> Pre-minion system tests... done [0s]
> Configure internal repositories... done [1s]
> Check mandatory packages presence... done [4s]
> Disabling Salt minion service... done [0s]
> Stopping Salt minion service... done [0s]
> Installing mandatory packages... done [11s]
> Configuring Salt minion to run in local mode... done [6s]
> Ensure archive is available... done [2s]
> Checking local node... fail [2s]

Failure while running step 'Checking local node'

Command: check_local_node

Output:

<< BEGIN >>
Error running 'metalk8s_checks.node': Node bootstrap: Package containerd.io-1.4.3 conflicts with MetalK8s installation, please remove it.
Package docker-ce-20.10.2 conflicts with MetalK8s installation, please remove it.
<< END >>

This script will now exit
```

Expansion output when docker-ce installed on this new node
```
# crictl exec -it 8cf89df3f31ef salt-run state.sls metalk8s.orchestrate.deploy_node saltenv=metalk8s-2.8.0-dev pillar="{'orchestrate': {'node_name': 'node-1'}}"
[...]
bootstrap_master:
[...]
----------
          ID: Check node
    Function: metalk8s.saltutil_cmd
        Name: metalk8s_checks.node
      Result: False
     Comment: Running function metalk8s_checks.node failed on minions: node-1
     Started: 14:45:26.883648
    Duration: 3345.072 ms
     Changes:   
              ----------
              node-1:
                  ----------
                  retcode:
                      1
                  stderr:
                      Error running 'metalk8s_checks.node': Node node-1: Package containerd.io-1.4.3 conflicts with MetalK8s installation, please remove it.
                      Package docker-ce-20.10.2 conflicts with MetalK8s installation, please remove it.
                  stdout:

Summary for bootstrap_master
------------
Succeeded: 1 (changed=2)
Failed:    1
------------
Total states run:     2
Total run time:   5.698 s

```

---

Fixes: #2992

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -10,10 +10,13 @@ import re
 import six
 import socket
 import tempfile
+import textwrap
 import time
 
 from salt.pillar import get_pillar
 from salt.exceptions import CommandExecutionError
+import salt.loader
+import salt.template
 import salt.utils.args
 import salt.utils.files
 from salt.utils.hashutils import get_hash
@@ -557,3 +560,58 @@ def manage_static_pod_manifest(
     if source_filename:
         _clean_tmp(source_filename)
     return ret
+
+
+def get_from_map(value, saltenv=None):
+    """Get a value from map.jinja so that we have an up to date value
+    computed from defaults.yaml and pillar.
+
+    Basically the same as a `jinja.map_load` but with support for saltenv and
+    also hardcoded path to MetalK8s map.jinja.
+
+    Also add logic to retrieve the saltenv using version in the pillar.
+
+    Arguments:
+
+        value (str): Name of the value to retrieve
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        # Retrieve `metalk8s` merge between defaults.yaml and pillar using
+        # current node version as saltenv
+        salt '*' metalk8s.get_from_map metalk8s
+
+        # Retrieve `metalk8s` from a specific saltenv
+        salt '*' metalk8s.get_from_map meltak8s saltenv=my-salt-env
+    """
+    path = "metalk8s/map.jinja"
+    if not saltenv:
+        current_version = __pillar__.get('metalk8s', {}).get('nodes', {}).get(
+            __grains__['id'], {}).get('version')
+        if not current_version:
+            log.warning(
+                'Unable to retrieve current running version, fallback on "base"'
+            )
+            saltenv = 'base'
+        else:
+            saltenv = "metalk8s-{}".format(current_version)
+
+    tmplstr = textwrap.dedent(
+        """\
+        {{% from "{path}" import {value} with context %}}
+        {{{{ {value} | tojson }}}}
+        """.format(
+            path=path, value=value
+        )
+    )
+    return salt.template.compile_template(
+        ":string:",
+        salt.loader.render(__opts__, __salt__),
+        __opts__["renderer"],
+        __opts__["renderer_blacklist"],
+        __opts__["renderer_whitelist"],
+        input_data=tmplstr,
+        saltenv=saltenv
+    )
--- a/salt/_modules/metalk8s_checks.py
+++ b/salt/_modules/metalk8s_checks.py
@@ -12,6 +12,93 @@ def __virtual__():
     return __virtualname__
 
 
+def node(raises=True, **kwargs):
+    """Check if the current Salt-minion match some requirements so that it can
+    be used as a MetalK8s node.
+
+    Arguments:
+        raises (bool): the method will raise if there is any problem.
+    """
+    errors = []
+
+    # Run `packages` check
+    pkg_ret = __salt__['metalk8s_checks.packages'](raises=False, **kwargs)
+    if pkg_ret is not True:
+        errors.append(pkg_ret)
+
+    # Compute return of the function
+    if errors:
+        error_msg = 'Node {}: {}'.format(__grains__['id'], '\n'.join(errors))
+        if raises:
+            raise CheckError(error_msg)
+        return error_msg
+
+    return True
+
+
+def packages(conflicting_packages=None, raises=True, **kwargs):
+    """Check if some conflicting package are installed on the machine,
+    return a string (or raise if `raises` is set to `True`) with the list of
+    conflicting packages.
+
+    Arguments:
+        conflicting_packages (dict): override the list of package that conflict
+            with MetalK8s installation.
+        raises (bool): the method will raise if there is any conflicting
+            package.
+
+    Note: We have some logic in this function, so `conflicting_packages` could
+    be:
+    - a single string `<package_name>` which is the name of the conflicting
+      package (it means we conflict with all versions of this package)
+    - a list `[<package_name1>, <package_name2]` with all conflicting package
+      names (it means we conflict with all versions of those packages)
+    - a dict `{<package1_name>: <package1_versions>, <package2_name>: <package2_versions>}`
+      where `package_versions` could be
+      - `None` mean we conflicting with all versions of this package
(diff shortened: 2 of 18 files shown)


salt/_modules/metalk8s.py
```diff
@@ -557,3 +560,58 @@ def _clean_tmp(sfn):
...
+        # current node version as saltenv
+        salt '*' metalk8s.get_from_map metalk8s
+
+        # Retrieve `metalk8s` from a specific saltenv
+        salt '*' metalk8s.get_from_map meltak8s saltenv=my-salt-env
+    """
+    path = "metalk8s/map.jinja"
+    if not saltenv:
+        current_version = __pillar__.get('metalk8s', {}).get('nodes', {}).get(
+            __grains__['id'], {}).get('version')
```
gdemonet, line 592: minor: Could be more readable with:

```python
current_version = __salt__['pillar.get'](
    'metalk8s:nodes:{}:version'.format(__grains__['id'])
)
```

salt/_modules/metalk8s_checks.py
```diff
@@ -12,6 +12,81 @@ def __virtual__():
...
+    for package, version in conflicting_packages.items():
+        if isinstance(version, str):
+            version = [version]
+        if package in installed_packages:
+            conflicting_versions = set(
+                pkg_info['version']
+                for pkg_info in installed_packages[package]
+            )
+            if version:
+                conflicting_versions &= set(version)
```
gdemonet: minor: there's a method to do this, instead of constructing a new set (and also it's more readable than the overloaded bitwise-operator):

```python
conflicting_versions.intersection_update(version)
```

salt/_modules/metalk8s_checks.py
```diff
@@ -12,6 +12,81 @@ def __virtual__():
...
+
+
+def packages(conflicting_packages=None, raises=True, **kwargs):
+    """Check if some conflicting package are installed on the machine,
+    return a string (or raise if `raises` is set to `True`) with the list of
+    conflicting packages.
+
+    Arguments:
+        conflicting_packages (dict): override the list of package that conflict
+            with MetalK8s installation.
```
gdemonet, line 46: The format of this dict (which can also be a single string, or a list, apparently) would warrant some examples (especially since the values can be either `None`, a single version, or a list of versions) for the reader to understand.

salt/_modules/metalk8s_checks.py
```diff
@@ -12,6 +12,81 @@ def __virtual__():
...
+                pkg_info['version']
+                for pkg_info in installed_packages[package]
+            )
+            if version:
+                conflicting_versions &= set(version)
+
+            if conflicting_versions:
+                for ver in sorted(conflicting_versions):
+                    errors.append(
+                        "Package {}-{} conflict with MetalK8s installation, "
```
gdemonet:
```suggestion
                        "Package {}-{} conflicts with MetalK8s installation, "
```

gdemonet (changes requested):
Just one typo in the user-facing message, the rest is more minor/internal. Otherwise LGTM :ok_hand: !

salt/metalk8s/defaults.yaml
```diff
@@ -37,6 +37,14 @@ kubeadm_preflight:
...
 
 repo:
+  conflicting_packages:
+    # List of package that conflict with MetalK8s installation
+    # <package_name>: [<version1>, <version2>]
+    # Is version list is None then we consider all versions as conflicting
+    docker: null
+    docker-ce: null
+    containerd.io: null
+
```
NicolasT, line 47: Could you take a look at the list provided by @baitanik in https://github.com/scality/metalk8s/discussions/2989#discussioncomment-224790 and check whether this list should be extended accordingly?

TeddyAndrieux: Yes I could try with those packages, but for pip packages, it's a bit hard to check IMHO.
E.g.: Checking all dependencies of Salt installed by pip and not by system package... It's doable but not sure it's worth and honestly deserve another PR since it's really a different check to me

salt/_modules/metalk8s_checks.py
TeddyAndrieux: You have test for this :smile:, no, joking, you right I will add some examples

salt/_modules/metalk8s_checks.py
TeddyAndrieux: Ok right

salt/_modules/metalk8s.py
TeddyAndrieux: Personally, I prefer to use directly the `__pillar__` dict instead of calling a module just to retrieve the pillar value

salt/_modules/metalk8s_checks.py
TeddyAndrieux: Done

salt/_modules/metalk8s.py
gdemonet: OK fine by me.

salt/_modules/metalk8s_checks.py
```diff
@@ -12,6 +12,93 @@ def __virtual__():
...
+    be:
+    - a single string `<package_name>` which is the name of the conflicting
+      package (it means we conflict with all versions of this package)
+    - a list `[<package_name1>, <package_name2]` with all conflicting package
+      names (it means we conflict with all versions of those packages)
+    - a dict `{<package1_name>: <package1_versions>, <package2_name>: <package2_versions>}`
+      where `package_versions` could be
+      - `None` mean we conflicting with all versions of this package
+      - A string for a single conflicting version of this package
+      - A list of string for multiple conflicting versions of this package
```
gdemonet, line 60: :100:

gdemonet (approved):
Approving for these packages, I guess we can easily add some more (as the ones mentioned in https://github.com/scality/metalk8s/discussions/2989#discussioncomment-224790) later.

PR #3053, opened 2021-01-19
Certificates beacon no longer check apiserver URL

alexandre-allard:
**Component**:
salt

**Context**:
In kubeconfig beacon we check for the apiserver URL with a specific hardcoded port (6443) which is not working for some of the kubeconfigs (e.g. calico), thus triggering a certificate renewal every time the beacon run for this certificate (luckily the certificate is not renewed because there is too much days left).

**Summary**:
Make the apiserver check optional in the `metalk8s_kubeconfig.validate()` function and no longer check it in kubeconfig beacon.

**Acceptance criteria**:
to be tested

### DIFF
--- a/salt/_modules/metalk8s_kubeconfig.py
+++ b/salt/_modules/metalk8s_kubeconfig.py
@@ -16,7 +16,7 @@ def __virtual__():
 
 def validate(filename,
              expected_ca_data,
-             expected_api_server,
+             expected_api_server=None,
              expected_cn=None,
              days_remaining=90):
     """Validate a kubeconfig filename.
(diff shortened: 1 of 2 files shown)


salt/_modules/metalk8s_kubeconfig.py
```diff
@@ -16,7 +16,7 @@ def __virtual__():
 
 def validate(filename,
              expected_ca_data,
```
TeddyAndrieux: Not related to this PR but to me, this one should be optional as well so that a user can easily check a kubeconfig by running `salt-call metalk8s_kubeconfig.validate <my-kubeconf>`

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):

PR #3058, opened 2021-01-20
tools: Move ui-metalk8s-client-generator to tools/crd-client-generator-js and refactor it to be completely uncoupled with metalk8s

JBWatenbergScality:
**Component**:

tools, crd-client-generator

**Context**: 

Move ui-metalk8s-client-generator to tools/crd-client-generator-js.
I also introduced a new optional cli parameter `typePrefix` which was before hardcoded to Metalk8s, coupling the crd generator with metalk8s whereas it could be used for any other crd.


gdemonet (dismissed):

eve/main.yml
```diff
@@ -1336,6 +1336,49 @@ stages:
             STEP_NAME: unit_tests_ui
       - Upload: *upload_final_status_artifact
 
+  unit_tests_crd_client_generator:
```
gdemonet, line 1314: You need to add this stage to the stages triggered by `unit_tests`

eve/main.yml
```diff
@@ -1336,6 +1336,49 @@ stages:
...
+            <<: *_env_final_status_artifact_failed
+            <<: *_unit-test_crd_client_generator_junit_info
+            STEP_NAME: unit_tests_crd_client_generator
+      - ShellCommand: *setup_cache
+      - ShellCommand:
+          name: Install CRD client generator dependencies
+          workdir: build/tools/crd-client-generator-js
+          command: >
+            npm ci --prefer-offline --no-audit
+          haltOnFailure: true
```
gdemonet, line 1341: Could be done in the worker image, to cache those dependencies as much as possible

eve/main.yml
```diff
@@ -1336,6 +1336,49 @@ stages:
...
+    _metalk8s_internal_info:
+      junit_info: &_unit-test_crd_client_generator_junit_info
+        TEST_SUITE: unit test
+        TEST_NAME: crd client generator
+    worker:
+      type: kube_pod
+      path: eve/workers/pod-unit-tests/pod.yaml
+      images:
+        docker-unit-tests:
+          context: 'ui'
```
gdemonet: not sure it makes sense to use this build context

eve/main.yml
JBWatenbergScality: This is already done in the worker : https://github.com/scality/metalk8s/blob/development/2.8/eve/workers/pod-unit-tests/ui/Dockerfile#L35 but the build context was targeting `ui` instead of `tools/crd-client-generator-js` so wrong dependencies were cached.

tools/crd-client-generator-js/package.json
```diff
@@ -21,6 +21,7 @@
     "@types/jest": "^26.0.19",
     "babel-jest": "^26.6.3",
     "jest": "^26.6.3",
-    "ts-node": "^9.1.1"
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.3"
```
gdemonet, line 26: How is this new dependency related to the "move" commit?

tools/crd-client-generator-js/README.md
```diff
@@ -1,11 +1,17 @@
-## Metalk8s client generator
+## CRD client generator
 
-It generates a flow client based on `@kubernetes/client` custom objects client from kubernetes `CustomResourceDefinition` of metalk8s volumes.
+It generates a flow client based on `@kubernetes/client` custom objects client from a kubernetes `CustomResourceDefinition`.
```
gdemonet: minor: capitalize "Kubernetes"

eve/main.yml
gdemonet: Is this still required then?

tools/crd-client-generator-js/src/generateClient.ts
```diff
@@ -1,5 +1,6 @@
 const YAML = require('yaml');
 const fs = require('fs');
+import { version } from '@babel/core';
```
gdemonet: Automatic import? Don't think this one is used :/

tools/crd-client-generator-js/src/generateClient.ts
```diff
@@ -17,15 +18,29 @@ export function generateClient(crdFile: string, destinationFile: string, prefix?
...
-  const typePrefix = (prefix || '') + crdSpec.spec.version.replace(/^\w/, (c) => c.toUpperCase());
+  const generateClientHeader = `
+  //@flow
+  // /!\\ WARNING :  THIS FILE IS AUTOGENERATED FROM A KUBERNETES CUSTOM RESOURCE DEFINITION FILE. DO NOT CHANGE IT, use crd-client-generator-js to update it.
+  import { customObjects } from './api';
+  export type Result<T> = T | {error: any};
+  
+  `;
+
+  const versionsClients = versions.map(version => {
```
gdemonet: New block, but it looks like the code that follows was not indented

gdemonet (changes requested):
Some (late, sorry) comments

tools/crd-client-generator-js/package.json
JBWatenbergScality: This is not related to the move I found it was missing when reinstalling dependencies, I will move this to a separate commit

tools/crd-client-generator-js/src/generateClient.ts
```diff
@@ -1,59 +1,90 @@
...
+  const file = fs.readFileSync(crdFile, "utf8");
   const crdSpec = YAML.parse(file);
-  
+
   const generator = new Generator();
+
+  const versions =
+    crdSpec.apiVersion === "apiextensions.k8s.io/v1"
+      ? crdSpec.spec.versions.map((version) => version.name)
+      : [crdSpec.spec.version];
```
gdemonet: Sorry, just caught this: I don't think the `versions` field only exists in `v1`. It's the opposite: in `v1beta1`, there is **also** a `version` field, and both may be specified (in which case `version` **must** match `versions[0]`).

tools/crd-client-generator-js/src/generateClient.ts
```diff
@@ -1,59 +1,90 @@
...
+  const jsonSchemas =
+    crdSpec.apiVersion === "apiextensions.k8s.io/v1"
+      ? crdSpec.spec.versions.reduce(
+          (aggregatedJsonSchemas, version) => ({
+            ...aggregatedJsonSchemas,
+            [version.name]: version.schema.openAPIV3Schema,
+          }),
+          {}
+        )
+      : { [crdSpec.spec.version]: crdSpec.spec.validation.openAPIV3Schema };
```
gdemonet: Same here: `spec.validation` may describe a top-level schema for all versions, but it's an optional field, and `v1beta1` CRDs can use `spec.versions[*].schema` instead.

gdemonet (changes requested):
Oops, missed some important details last time (that's why we try not to mix commits which update logic with ones that only apply formatting changes, otherwise reviewers may not be careful enough :wink:).
For reference:
- [CRD v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#customresourcedefinition-v1beta1-apiextensions-k8s-io)
- [CRD v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#customresourcedefinition-v1-apiextensions-k8s-io)

Maybe you want to use `@kubernetes/client` types in your generator? :stuck_out_tongue_winking_eye:

JBWatenbergScality:
I should have thought about using `@kubernetes/client` types here before ! Thanks 👍   !

tools/crd-client-generator-js/src/generateClient.ts
```diff
@@ -1,59 +1,95 @@
...
+    ? { [crdSpec.spec.version]: crdSpec.spec.validation.openAPIV3Schema }
+    : (crdSpec.spec.versions || []).reduce(
+        (aggregatedJsonSchemas, version) => ({
+          ...aggregatedJsonSchemas,
+          [version.name]:
+            version.schema?.openAPIV3Schema ||
+            crdSpec.spec.validation.openAPIV3Schema,
+        }),
+        {}
+      );
```
gdemonet, line 43: :+1:

gdemonet (dismissed):

eve/main.yml
gdemonet: Seen offline: it's required since we can't cache `node_modules` in the image (as we need to mount the sources over it). Though the worker build should pre-install them, saving as much as possible in `~/.npm`, and this step tries to re-use this cache as much as possible with `--prefer-offline`.

ChengYanJin (approved):

PR #3066, opened 2021-01-25
Enforce salt Python linting

### LINKED ISSUE #2649: Add linting for all Python in Salt directory in CI
**Component**:

'salt', 'ci'

**Why this is needed**:

We do not have any linting check for python in Salt module, states, ... in the CI

**What should be done**:

Add python linting for Salt python stuff

pylint

TeddyAndrieux:
**Component**:

'salt', 'lint', 'python'

**Context**: 

#2649 

**Summary**:

- Fix python linting issue in all Salt modules
- Enable salt python linting in tox and doit, enforce it in CI build

---

Fixes: #2649


salt/_modules/metalk8s_drain.py
```diff
@@ -56,7 +56,7 @@ def __str__(self):
 
 class DrainTimeoutException(DrainException):
     '''Timeout-specific drain exception.'''
-    pass
+    pass  # pylint: disable=unnecessary-pass
```
NicolasT: Can't it be removed instead?

salt/_modules/metalk8s_drain.py
```diff
@@ -357,7 +357,7 @@ def run_drain(self, dry_run=False):
         try:
             pods = self.get_pods_for_eviction()
         except DrainException as exc:
-            raise CommandExecutionError(
+            raise CommandExecutionError(  # pylint: disable=raise-missing-from
```
NicolasT: Can't the `from` be added instead?

salt/_modules/metalk8s_monitoring.py
```diff
@@ -203,7 +203,7 @@ def _requests_alertmanager_api(route, method='GET', **kwargs):
                 "Malformed response returned from Alertmanager API: {!s}: {}"
                 .format(exc, response.text)
             )
-        raise CommandExecutionError(error)
+        raise CommandExecutionError(error)  # pylint: disable=raise-missing-from
```
NicolasT: Same?

TeddyAndrieux: Do we really want to raise the ValueError part of `response.json` ? 
Hmm maybe it make sense so that we know what the problem with the response content as it may not be obvious even if you see the response text
Ok fair I will update this one

salt/_modules/metalk8s_drain.py
TeddyAndrieux: We can but not sure it bring any value, you will just get the traceback to see from where it gets raised and it's will likely be on line 
```
        if failures:
            raise DrainException(_message_from_pods_dict(failures))
```
So ....

salt/_modules/metalk8s_drain.py
TeddyAndrieux: It could, but maybe we want to keep this one for code clarity, no strong opinion on this one, with a `pass` or without both ok for me

salt/_modules/metalk8s_drain.py
gdemonet: What if we later change the `DrainException` to be raised from another one? IMO, if `pylint` enforces it as a rule, a `raise Something from caught_exception` never hurts.

salt/_modules/metalk8s_drain.py
gdemonet: From what I've seen in many Python codebases, the `pass` is usually omitted when possible (like in this case, because there's a docstring).
Exception subclasses are often using this approach: see https://github.com/saltstack/salt/blob/master/salt/exceptions.py#L72-L117 for instance.
So IMHO, we should remove it and let pylint check that we remain consistent.

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
```diff
@@ -11,7 +11,7 @@ common_tests:
       a: b
     info_scope: null
     raises: True
-    result: "Invalid manifest: An error has occurred"
+    result: "Invalid manifest"
```
gdemonet, line 14: In tests, could be useful to check that we can find the original error in the traceback. Maybe to tackle later, as debt.

salt/_states/metalk8s.py
```diff
@@ -5,6 +5,7 @@
 import traceback
 import re
 
+from salt.exceptions import CommandExecutionError
```
gdemonet, line 8: :+1: (undefined vars are a really useful check :heart:)

salt/_renderers/metalk8s_kubernetes.py
```diff
@@ -38,7 +38,7 @@ def _step_name(manifest, absent=False):
     try:
         name = manifest['metadata']['name']
     except KeyError:
-        raise SaltRenderError('Object `metadata.name` must be set.')
+        raise SaltRenderError('Object `metadata.name` must be set.')  # pylint: disable=raise-missing-from
```
gdemonet: This one could have been kept (just to know whether the missing key is `metadata` or `name`)

salt/.pylintrc
```diff
@@ -0,0 +1,704 @@
...
+  do-not-assign-a-lambda-expression-use-a-def,
+  3rd-party-local-module-not-gated,
+  pep8-reserved-keywords,
+  str-format-in-logging,
+  import-outside-toplevel,
+  deprecated-method,
+  repr-flag-used-in-string,
+  keyword-arg-before-vararg,
+  incompatible-py3-code,
+  multiple-spaces-before-operator
```
gdemonet: Isn't this a lot to disable? Recommended by Salt? Some checks like `unused-variable`/`undefined-loop-variable` may be useful, don't you think?

gdemonet (changes requested):
Thanks for taking care of this, will be a good protection against simple mistakes! I just answered on some conversations, nothing really blocking though.

salt/.pylintrc
TeddyAndrieux: Yes maybe, btw this one just get copied from pylintrc from salt repo, but indeed we may want to tune it a bit (I remember I already did some minor changes, do not remember the exact ones)

salt/_renderers/metalk8s_kubernetes.py
TeddyAndrieux: Right, make sense

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
TeddyAndrieux: Hmm you right, I will try to do something about it

salt/_modules/metalk8s_drain.py
TeddyAndrieux: Ok fair I remove this one

salt/_modules/metalk8s_drain.py
TeddyAndrieux: Ok fair

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
TeddyAndrieux: Ok I think we will do it later I will open a ticket for this, basically you need to check the content of `__cause__` in the exception so will likely need to create a new `assertRaises...` to handle it (#3086)

gdemonet: Perfect, thanks!

salt/.pylintrc
gdemonet: OK let's keep it like this for now, and add some debt ticket to track that we'll want to include more rules :)

gdemonet (dismissed):

salt/.pylintrc
TeddyAndrieux: I will remove from ignore list (and fix our code if needed) for at least those 
```
whitespace-before-colon,
indentation-is-not-a-multiple-of-four-comment,
whitespace-before-colon,
indentation-is-not-a-multiple-of-four-comment,
undefined-loop-variable,
unexpected-indentation-comment,
continuation-line-indentation-is-not-a-multiple-of-four,
continuation-line-missing-indentation-or-outdented,
closing-bracket-does-not-match-indentation-of-opening-brackets-line,
closing-bracket-does-not-match-visual-indentation,
continuation-line-does-not-distinguish-itself-from-next-logical-line,
continuation-line-over-indented-for-hanging-indent,
continuation-line-over-indented-for-visual-indent,
continuation-line-under-indented-for-visual-indent,
visually-indented-line-with-same-indent-as-next-logical-line,
unaligned-for-hanging-indent,
block-comment-should-start-with-cardinal-space,
too-many-leading-hastag-for-block-comment,
pep8-reserved-keywords,
str-format-in-logging,
```
Tell me if you have others that you think we should not ignore (anyway we could remove then from ignore list  in the future if needed)

salt/_auth/kubernetes_rbac.py
```diff
@@ -177,7 +177,7 @@ def _load_kubeconfig(opts):
 
 
 @_check_auth_args
-def auth(username, token=None, **kwargs):
+def auth(username, token=None, **kwargs):  # pylint: disable=unused-argument
```
gdemonet: I'm not sure, but I think there is some convention to describe "sink" arguments (ones that you will not use) to prefix them with an underscore.
How does pylint behave if you write instead:
```python
def auth(username, token=None, **_kwargs):
```
?

gdemonet (dismissed):
Just a style question on some function signatures, where we decide not to use some arguments.
Would you prefer keeping the original signature, as is done today, or make it explicit that you discard some arguments with `(..., **_kwargs):`?
Anyhow, let's not block the merge for this :wink:

salt/_auth/kubernetes_rbac.py
TeddyAndrieux: For kwargs yes it's true we can do this. I will change this. 

Note that pylint point of view if you do a `def toto(_b=None)` for a `b` args that you do not use, pylint will report no error but then the signature of the function will be different

gdemonet (approved):

PR #3069, opened 2021-01-27
salt: Add a check about conflicting services for MetalK8s

TeddyAndrieux:
**Component**:

'pre-check', 'salt'

**Context**: 

#3067 

**Summary**:

If some service are started on the host where we want to deploy MetalK8s
the installation may not work properly (e.g.: firewalld)
Add a new function to check that those service are not started on the
host before deploying all the MetalK8s components.
NOTE: We do not automatically stop the service from the host since those
services may have been started for good reason, so just ask the user to
remove those packages

**Acceptance criteria**: 

Bootstrap output with firewalld started
```
# /srv/scality/metalk8s-2.8.0-dev/bootstrap.sh 
> Determine the OS... done [0s]
> Checking that BootstrapConfiguration is present... done [0s]
> Pre-minion system tests... done [0s]
> Configure internal repositories... done [0s]
> Check mandatory packages presence... done [3s]
> Disabling Salt minion service... done [0s]
> Stopping Salt minion service... done [0s]
> Installing mandatory packages... done [10s]
> Configuring Salt minion to run in local mode... done [5s]
> Ensure archive is available... done [2s]
> Checking local node... fail [2s]

Failure while running step 'Checking local node'

Command: check_local_node

Output:

<< BEGIN >>
Error running 'metalk8s_checks.node': Node bootstrap: Service firewalld conflicts with MetalK8s installation, please stop and disable it.
<< END >>

This script will now exit
```

Expansion with firewalld started on the new node
```
# crictl exec -it 758110e823084 salt-run state.sls metalk8s.orchestrate.deploy_node saltenv=metalk8s-2.8.0-dev pillar="{'orchestrate': {'node_name': 'node-1'}}"
[...]
bootstrap_master:
----------
[...]
----------
          ID: Check node
    Function: metalk8s.saltutil_cmd
        Name: metalk8s_checks.node
      Result: False
     Comment: Running function metalk8s_checks.node failed on minions: node-1
     Started: 17:36:42.667642
    Duration: 9560.307 ms
     Changes:   
              ----------
              node-1:
                  ----------
                  retcode:
                      1
                  stderr:
                      Error running 'metalk8s_checks.node': Node node-1: Service firewalld conflicts with MetalK8s installation, please stop and disable it.
                  stdout:

Summary for bootstrap_master
------------
Succeeded: 1 (changed=2)
Failed:    1
------------
Total states run:     2
Total run time:  22.045 s

```

---

Fixes: #3067

PR #3076, opened 2021-02-02
salt: Check Service CIDR has a route declared

gdemonet:
**Component**: salt, kubernetes, networking

**Context**:

When deploying some control plane services in the cluster (e.g. calico), we let such services interact with `kube-apiserver` using its Service IP (10.96.0.1 by default).
However, the resolution of such virtual IPs, emulated using `iptables` (through `kube-proxy`), requires at a minimum that a route exists for the Service CIDR (a default route also works).
We add a check to the `metalk8s_checks.node` helper, to have it run before bootstrap and expansion.

See: kubernetes/kubernetes#57534

**Summary**:

- First check implemented using `network.get_route` Salt execution method
- Re-worked to handle systems where kernel refuses network ranges (strict netlink validation)
- Removed initial implementation, simply read through the entire routing table

**Acceptance criteria**:

- CI keeps passing
- Installing on a system without a route set for the configured `networks:service` range should bail out (both in bootstrap and expansion), similarly to checks from #3050 or #3069

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@
   services already started on the machine before doing all the installation
   (PR [#3069](https://github.com/scality/metalk8s/pull/3069))
 
+- [kubernetes/kubernetes#57534](https://github.com/kubernetes/kubernetes/issues/57534) -
+  Check if a route exists for the Service IPs CIDR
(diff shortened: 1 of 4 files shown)


TeddyAndrieux (approved):
LGTM

PR #3081, opened 2021-02-03
Improve iso-manager script error handling

### LINKED ISSUE #2862: Invalid `archives` in BootstrapConfiguration should error
**Component**: salt

**What happened**:

If a value in the `archives` list in BootstrapConfiguration isn't either a file or a directory, we simply skip it.

**What was expected**:

Given this list is user-provided configuration, it makes sense for us to error out when we encounter invalid values.

**Steps to reproduce**:

Run `bootstrap.sh` with (only) an invalid key under the `archives` key of BootstrapConfiguration.
See #2861 for an example of logs.

- Make sure the `metalk8s.get_archives` execution function raises a `CommandExecutionError` if such a situation occurs, instead of skipping with a warning log
- Add a validation step in `bootstrap.sh` to fail early

### LINKED ISSUE #3079: iso-manager script should fail fast with clear error message when bad archive entries are detected in bootstrap.yaml
**Component**:

'iso-manager'

**Why this is needed**:

When upgrading to a new metalk8s version, we first need to import the new metalk8s.iso.
When doing so, the `iso-manager` script may fail with the following error:

``Rendering SLS 'metalk8s-2.7.0-dev:metalk8s.salt.master.installed' failed: Problem running salt function in Jinja template: Failed to run isoinfo: isoinfo: No such file or directory. Cannot open 'metalk8s.iso'. Cannot open SCSI driver.
isoinfo: No such file or directory. Unable to open metalk8s.iso; line 34``

Actually, this is because of the `bootstrap.yaml` file was having a bad entry (a relative path to an iso) in the archives section.
Don't really know how this entry was added to the config file but since it can be edited manually, it may happen that bad archive entries are present in the file.

**What should be done**:

In such a situation, the `iso-manager` script should abort immediately and provide the administrator with a clear error message for him to understand that the `bootstrap.yaml `contains some wrong archive entries and that it should be removed.
Ideally, the `iso-manager` script starts with listing the archive entries and make sure they are all valid (i.e. they point to existing iso file on disk). If one of them is not valid it failed and displays the error message.

**Test plan**:

### LINKED ISSUE #3082: iso-manager script should build and store metalk8s.iso absolute path in bootstrap.yaml file
**Component**:

'iso-manager', 'upgrade'

**Why this is needed**:

When doing an upgrade the first step is to mount the new metalk8s.iso using the iso-manager script using the following command: `/srv/scality/metalk8s-X.X.X/iso-manager.sh -a <path_to_iso>`.
One may use a relative path for the metalk8s.iso, which can lead to a situation where the `boostrap.yaml` file is updated with an invalid archive entry. (cf #3079). The solution script (which is doing something similar when importing a solution) is computing the absolute path from what is provided by the user.

**What should be done**:

As for the solution script, the `iso-manager` script should check and compute the absolute path for the metalk8s.iso arg that is passed by the administrator and store it as an absolute path in the `boostrap.yaml` file

**Test plan**:

alexandre-allard:
**Component**: scripts

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #3079
Closes: #2862

### DIFF
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -25,6 +25,8 @@ log = logging.getLogger(__name__)
 
 __virtualname__ = "metalk8s"
 
+BOOTSTRAP_CONFIG = "/etc/metalk8s/bootstrap.yaml"
+
 
 def __virtual__():
     return __virtualname__
@@ -146,6 +148,28 @@ def _get_archive_info(info):
     return {"version": _get_archive_version(info), "name": _get_archive_name(info)}
 
 
+def archive_info_from_product_txt(archive):
+    if os.path.isdir(archive):
+        info = archive_info_from_tree(archive)
+        info.update({
+            'iso': None,
+            'path': archive,
+        })
+    elif os.path.isfile(archive):
+        info = archive_info_from_iso(archive)
+        info.update({
+            'iso': archive,
+            'path': '/srv/scality/metalk8s-{0}'.format(info['version']),
+        })
+    else:
+        raise CommandExecutionError(
+            'Invalid archive path {0}, should be an iso or a directory.'
+            .format(archive)
+        )
+
+    return info
+
+
 def archive_info_from_tree(path):
     """Extract archive information from a directory
 
@@ -210,25 +234,8 @@ def get_archives(archives=None):
 
     res = {}
     for archive in archives:
-        if os.path.isdir(archive):
-            iso = None
-            info = archive_info_from_tree(archive)
-            path = archive
-            version = info["version"]
-        elif os.path.isfile(archive):
-            iso = archive
-            info = archive_info_from_iso(archive)
-            version = info["version"]
-            path = "/srv/scality/metalk8s-{0}".format(version)
-        else:
-            log.warning(
-                "Skip, invalid archive path %s, should be an iso or a " "directory.",
-                archive,
-            )
-            continue
-
-        info.update({"path": path, "iso": iso})
-        env_name = "metalk8s-{0}".format(version)
+        info = archive_info_from_product_txt(archive)
+        env_name = 'metalk8s-{0}'.format(info['version'])
 
         # Warn if we have 2 archives with the same version
         if env_name in res:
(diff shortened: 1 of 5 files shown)


scripts/iso-manager.sh
```diff
@@ -139,14 +80,29 @@ _set_env() {
...
+    if (( ${#not_existing[@]} )); then
+        echo "The following archives do not exist:" >&2
+        for archive in "${not_existing[@]}"; do
+            echo "- $archive" >&2
+        done
+        return 1
+    fi
+
+    return 0
 }
```
TeddyAndrieux: IMHO this check should be in salt so that we can also check this in bootstrap, also this remind me this ticket #2862

scripts/iso-manager.sh
```diff
@@ -157,21 +113,15 @@ _add_archives() {
     # local: ["/srv/scality/metalk8s-2.0.0/", "/tmp/metalk8s-2.1.0.iso"]
     # parsed archives:
     # ("/srv/scality/metalk8s-2.0.0/" "/tmp/metalk8s-2.1.0.iso")
-    IFS=" " read -r -a \
-        archives <<< "$(salt-call pillar.get metalk8s:archives \
-        --out txt | cut -d' ' -f2- | tr -d '[],')"
+    local -a archives=(${CONFIGURED_ARCHIVES[@]+"${CONFIGURED_ARCHIVES[@]}"})
```
TeddyAndrieux: I'm not sure I understand what it does exactly
```
${CONFIGURED_ARCHIVES[@]+"${CONFIGURED_ARCHIVES[@]}"}
```

alexandre-allard: Same as:
```
local -a archives
if (( ${#CONFIGURED_ARCHIVES[@]} )); then
    archives=("${CONFIGURED_ARCHIVES[@]}")
fi
```
I can change it if you feel it's not clear enough.

scripts/iso-manager.sh
alexandre-allard: I agree, but having it here allow to fail early before touching this file.
Maybe we need it here and in Salt too or we could also do like we did for Solutions, not using file.serialize and having a custom module adding/removing ISO from this file.

TeddyAndrieux: The salt module could take archive list as input (and default to pillar content)

Hmm yes maybe it make sense to have a dedicated Salt module to add iso to bootstrap file

scripts/iso-manager.sh
TeddyAndrieux: Ok, as you want now that I know this, I'm ok to keep this one line :smiley:

alexandre-allard: Anyway, this line no longer exist :'( it's in Salt module now.

salt/_modules/metalk8s.py
```diff
@@ -214,26 +239,8 @@ def get_archives(archives=None):
...
-            )
-            continue
-
-        info.update({'path': path, 'iso': iso})
-        env_name = 'metalk8s-{0}'.format(version)
+        info = archive_info_from_product_txt(archive)
+        env_name = 'metalk8s-{0}'.format(info['version'])
 
         # Warn if we have 2 archives with the same version
         if env_name in res:
```
TeddyAndrieux: I think we may want to raise in this case as well (maybe in another PR/ticket)

salt/_modules/metalk8s.py
```diff
@@ -615,3 +622,36 @@ def get_from_map(value, saltenv=None):
...
+    archive_info_from_product_txt(archive)
+
+    try:
+        with open(BOOTSTRAP_CONFIG, 'r') as fd:
+            config = yaml.safe_load(fd)
+    except IOError as exc:
+        msg = 'Failed to load bootstrap config file at "{}": {}'.format(
+            BOOTSTRAP_CONFIG, str(exc)
+        )
+        raise CommandExecutionError(message=msg)
```
TeddyAndrieux: Not link to this PR but in Python3 it's better to raise exception and use `from exc` to have both exceptions (check this https://www.python.org/dev/peps/pep-3134/#explicit-exception-chaining)

NOTE: I fix all other except in my pylint PR #3066 (e.g.: https://github.com/scality/metalk8s/pull/3066/commits/e4557955353e02b360fe71926dbf1760d5ce5cc1)

salt/_modules/metalk8s.py
```diff
@@ -615,3 +622,36 @@ def get_from_map(value, saltenv=None):
         input_data=tmplstr,
         saltenv=saltenv
     )
+
+
+def configure_archive(archive, removed=False):
```
TeddyAndrieux: Minor: to me it's more "remove" than "removed" no ?

salt/_modules/metalk8s.py
```diff
@@ -615,3 +622,36 @@ def get_from_map(value, saltenv=None):
...
+        raise CommandExecutionError(message=msg)
+
+    if removed:
+        try:
+            config['archives'].remove(archive)
+        except ValueError:
+            pass
+    else:
+        if archive not in config['archives']:
+            config['archives'].append(archive)
```
TeddyAndrieux: Minor: To me it could be cool to have a simple message output like `<archive> added` `<archive> removed` `<archive> already present` `<archive> already absent` something like that

TeddyAndrieux (commented):
Few comments otherwise LGTM (just unit tests are missing + likely a changelog entry)

salt/_modules/metalk8s.py
alexandre-allard: Yes, I was wondering if we wanted to do that, I can handle it here.

salt/_modules/metalk8s.py
alexandre-allard: Oh, didn't know that, I will change it.
Thanks for the tip.

TeddyAndrieux (dismissed):
LGTM

TeddyAndrieux (approved):

PR #3087, opened 2021-02-05
Support the minimal responsiveness for MetalK8s UI

### LINKED ISSUE #3009: MetalK8s UI should support the minimal responsiveness for tablet screen
**Component**: UI

**Why this is needed**: The Pages of MetalK8s UI should display correctly on tablet, laptop, control room screens.

the resolution we should support for the small screen is 1024 x 768

metalk8s-responsivness-issue-4
metalk8s-responsivness-issue-3
metalk8s-responsivness-issue-2
metalk8s-responsivness-issue-1

**What should be done**: Adjust the screen for the small tablet screen.

There are some ways to adjust the Node & Volume page for the tablet screen.
1) We don't separate the page into two parts, which means the table will take all the space and there is a scrollbar available to check the tabs on the right.
2) Still separate the screen, make the left side and right side both scrollable on the width. 
3) https://scality.invisionapp.com/console/share/MP1YPLLYZ9/497099980

We will choose the first solution to support the minimal responsiveness to at least align with Zenko-UI.

**Test plan**:

ChengYanJin:
**Component**: UI, responsiveness

**Context**: To support the 1024*768 screen. 

**Summary**:

Introduce responsive font:
https://docs.google.com/document/d/1NwOYmiCf4hFAVsCi3hJYhHhPQ2n-5_PxydoJJBimOSw/edit

Preview: 
https://user-images.githubusercontent.com/18453133/107057319-9cb80c80-67d3-11eb-8ec5-7e6c635dd3df.mov

**Acceptance criteria**: 

---

Closes: #3009

### DIFF
--- a/ui/src/components/NodePageOverviewTab.js
+++ b/ui/src/components/NodePageOverviewTab.js
@@ -19,7 +19,8 @@ import { intl } from '../translations/IntlGlobalProvider';
 
 const TabContentContainer = styled.div`
   overflow-y: auto;
-  height: 75vh;
+  // 100vh subtract the height of navbar and tab header
+  height: calc(100vh - 40px - 2.8rem);
 `;
 
 const InformationSpan = styled.div`
--- a/ui/src/components/NodePageOverviewTab.js
+++ b/ui/src/components/NodePageOverviewTab.js
@@ -19,13 +19,11 @@ import { intl } from '../translations/IntlGlobalProvider';
 
 const TabContentContainer = styled.div`
   overflow-y: auto;
-  height: 78vh;
+  height: 75vh;
 `;
 
 const InformationSpan = styled.div`
-  padding-bottom: ${padding.large};
-  padding-left: ${padding.large};
-  display: flex;
+  padding: 0 0 ${padding.large} ${padding.large};
 `;
 
 const InformationLabel = styled.span`
@@ -48,8 +46,6 @@ const NodeNameContainer = styled.div`
   padding: ${padding.large} 0 ${padding.larger} ${padding.large};
 `;
 
-const NodeNameStatusContainer = styled.div``;
-
 const NodeName = styled.span`
   font-size: ${fontSize.larger};
   padding-left: ${padding.smaller};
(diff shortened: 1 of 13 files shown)


JBWatenbergScality:
## **Visual review :**

I think we should try to get rid of those unneeded scrallbars : 
image


image


1. narrow down the spacing between the search field and the button
2. tabs should have a minimal width matching their content width (plus maybe some padding), when the screen becomes too small we can add a scrollbar or buttons at the beginning and end to scroll to next tabs (eg: https://material-ui.com/components/tabs/#scrollable-tabs)

image

Add a tooltip when the text is shrunk

image
We might want to add a little bit of padding right to the text to be sure long volume names stays distinct from the node name

thomasdanan (dismissed):
Really cool,
I tested it on iPadPro and 1250x600 resolutions

Thanks

ui/src/components/ConstrainedText.js
```diff
@@ -0,0 +1,28 @@
...
+  white-space: nowrap;
+`;
+
+const BlockTooltip = styled.div`
+  & > .sc-tooltip {
+    display: block;
+  }
+`;
+
+export function ConstrainedText({ text }: { text: string }): Node {
```
JBWatenbergScality: I was about to move this component to core ui, maybe we should do this before merging this PR

ui/src/components/NodePageOverviewTab.js
```diff
@@ -19,13 +19,11 @@ import { intl } from '../translations/IntlGlobalProvider';
 
 const TabContentContainer = styled.div`
   overflow-y: auto;
-  height: 78vh;
+  height: 75vh;
```
JBWatenbergScality: I think that we should compute the tab content here because likely `75vh` are not taking the full height of the page

ui/src/components/VolumeListTable.js
```diff
@@ -77,7 +77,7 @@ const VolumeListContainer = styled.div`
 // * table body
 const Body = styled.div`
   display: block;
-  height: calc(100vh - 176px);
+  height: calc(100vh - 162px);
```
JBWatenbergScality: Why `162px` ? Maybe use some constant here to describe how you computed it

ui/src/components/ConstrainedText.js
ChengYanJin: Sure. Make sense to move this component to core-ui.

JBWatenbergScality (approved):

PR #3098, opened 2021-02-09
[UI] Hide warning alerts when critical alert is triggered

### LINKED ISSUE #3077: MetalK8s UI should hide the warning alert when the critical one is triggered
**Component**: ui, alerts

**Why this is needed**: When the critical alert is triggered, we still keep the warning alert, meaning we could see two exactly the same alerts just the severity is different. 

**What should be done**:
We should hide the warning alerts on the front-end side as long as the critical alert is active.

A `custom hook` or a utility function that should apply to all the alert table or even be reused by xcore.

**Test plan**:
Unit test

ChengYanJin:
**Component**: ui, monitoring

**Context**: 
We want to hide the warning alert when the critical one is triggered from the UI.

**Summary**:
Use React function instead of a custom hook to be able to apply it on the saga level, instead of at the component level. (Open for discussion)

Out of the scope of the issue:
Add flow-typed to allow Jest tests verified by Flow.

**Acceptance criteria**: 

---

Closes: #3077

### DIFF
--- a/ui/src/services/utils.js
+++ b/ui/src/services/utils.js
@@ -485,3 +485,47 @@ export const linuxDrivesNamingIncrement = (devicePath, increment) => {
     return '';
   }
 };
+
+// Return boolean to tell if the two alerts are the same only with different severity
+const isSameAlertWithDiffSeverity = (
+  alert1: $PropertyType<PrometheusAlert, labels>,
+  alert2: $PropertyType<PrometheusAlert, labels>,
+): boolean => {
+  // Filtering out the `severity` property
+  function replacer(key, value) {
+    if (key === 'severity') {
+      return undefined;
+    }
+    return value;
+  }
+
+  return JSON.stringify(alert1, replacer) === JSON.stringify(alert2, replacer);
+};
+
+/*
+ ** This function removes the warning alert from the list when critical one is triggered
+ ** It should be called at saga level before storing the alerts to `redux-store`
+ ** or where we resolve the promise with `react-query`
+ */
+export const removeWarningAlerts = (
+  alerts: Array<PrometheusAlert>,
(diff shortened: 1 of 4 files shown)


ui/src/services/utils.js
```diff
@@ -485,3 +485,47 @@ export const linuxDrivesNamingIncrement = (devicePath, increment) => {
...
+          removeIndex.push(i);
+        } else if (alerts[j].labels.severity === STATUS_WARNING) {
+          removeIndex.push(j);
+        }
+      }
+    }
+  }
+
+  let removedWarningAlerts = alerts;
+  removeIndex.forEach((index) => removedWarningAlerts.splice(index, 1));
```
JBWatenbergScality: In order to avoid any possible side effect on initial alerts array I would make `removedWarningAlerts` a copy of alerts instead of a simple ref. 
```suggestion
  let removedWarningAlerts = [...alerts];
  removeIndex.forEach((index) => removedWarningAlerts.splice(index, 1));
```

ui/src/components/NodePartitionTable.js
```diff
@@ -116,7 +117,7 @@ const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
             return getNodePartitionsTableData(
               nodeFSUsageResult.data.result,
               nodeFSSizeResult.data.result,
-              alerts,
+              removeWarningAlerts(alerts),
```
JBWatenbergScality: I think it is not necessary to filter alerts here

ChengYanJin: Agree. Dropped this commit.

JBWatenbergScality (approved):

PR #3100, opened 2021-02-09
[UI] Bugfix no banner display for no prom volume provision

ChengYanJin:
**Component**: ui, bug

**Context**: 
When we deploy a fresh bootstrap node, before provisioning the volume
for prometheus, we can't see the banner to guide you to provision the
volume from the deployed UI (not from local dev environment)

**Summary**:
The root cause is how do we handle the prometheus error. The logic is
correct, we first check if there error from prometheus API, if no, we go
to check the deployment.

However, we can't tell the status of the HTTP request by just check if
`result.error.response` exist. It exists for all the out of 2XX
response. So we should check the response status instead of.

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/src/ducks/app/monitoring.js
+++ b/ui/src/ducks/app/monitoring.js
@@ -333,7 +333,9 @@ function getClusterQueryStatus(result) {
 }
 
 export function* handlePrometheusError(clusterHealth, result) {
-  if (result.error.response) {
+  // TODO:
+  // Type result argument to something similar to {error: {response?: AxiosResponse} } to simplify optional chaining use here.
+  if (result?.error?.response?.status < 500) {
     yield put(setPrometheusApiAvailable(true));
     clusterHealth.error = `Prometheus - ${result.error.response.statusText}`;
   } else {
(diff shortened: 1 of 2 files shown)


gdemonet (dismissed):

ui/src/ducks/app/monitoring.js
```diff
@@ -333,7 +333,7 @@ function getClusterQueryStatus(result) {
 }
 
 export function* handlePrometheusError(clusterHealth, result) {
-  if (result.error.response) {
+  if (result?.error?.response?.status < 500) {
```
JBWatenbergScality, line 338: I think you can now use typing to assume that if you get into this function result would always have an error property, then you can type `result` argument to something similar to `{error: {response?: AxiosResponse} }` and that may help you to simplify optional chaining use here.

ChengYanJin: Hi @JBWatenbergScality, agree! we should use typing to simplify the optional chaining here. 
However, enable flow typing in this monitoring reducer would introduce a lot of typing errors.
So I added TODO commit and will address later on this issue: https://github.com/scality/metalk8s/issues/3001
WDYT?

JBWatenbergScality: ok

JBWatenbergScality (approved):

PR #3101, opened 2021-02-09
Fix snapshot generation in CI

gdemonet:
**Component**: ci

**Context**: 2.7.1 release process - could not generate snapshots (and when it ended up working, they were not valid for post-merge tests, at least in single node)

**Summary**:

- Upgrade pip to latest before install `python-openstackclient`
- Add `metalk8s-` prefix to generated snapshots
- Untaint Bootstrap node before deploying a Solution operator
- Check that the Solution operator is ready after deployment

**Acceptance criteria**:

- Can generate snapshots
- Can use said snapshots in post-merge

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -715,6 +715,20 @@ models:
         --solution "$SOLUTION_NAME" --version "$SOLUTION_VERSION"
       workdir: build/eve/workers/openstack-terraform/terraform/
       haltOnFailure: true
+  - ShellCommand: &wait_for_solution_operator
+      doStepIf: "%(prop:install_solution:-false)s"
+      name: Wait for Solution Operator to become ready
+      env:
+        ENVIRONMENT_NAME: "%(prop:solution_env_name:-example-environment)s"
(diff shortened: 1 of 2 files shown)


TeddyAndrieux (dismissed):
Thank you ! 
LGTM

eve/main.yml
```diff
@@ -663,6 +663,17 @@ models:
...
       haltOnFailure: true
+  - ShellCommand: &untaint_bootstrap
+      doStepIf: "%(prop:install_solution:-false)s"
+      name: Untaint Bootstrap before deploying a Solution
+      env:
+        NODE_NAME: "bootstrap"
+      command: >
+        ssh -F ssh_config bootstrap
+        sudo kubectl --kubeconfig /etc/kubernetes/admin.conf
+        patch node "$NODE_NAME" --patch '{"spec": {"taints": []}}'
```
gdemonet: :facepalm: Gah this fails because the quotes are removed by the first shell, before the command is sent through SSH...

TeddyAndrieux (approved):

PR #3114, opened 2021-02-12
[UI] Use new core UI components

### LINKED ISSUE #3060: Replace some UI components with the existing ones in core-ui
**Component**: ui

**Why this is needed**:
We extracted/improved some common components in core-ui. 
Thanks to #3046 bump the core-ui version to 0.7.0, we should be able to benefit all the new core-ui features/components in MetalK8s UI.

**What should be done**:
- The health severity selector
image

- The number input component
image

- Enable the hoverable sidebar

**Test plan**:

alexis-ld:
**Component**:
ui, sidebar, inputs, alerts

**Context**: 

We want to use the new Core-UI components. #3060 

**Summary**:
* Replaced ActiveAlertsFilters by core-ui's HealthSelector.
* Made Sidebar hoverable.
* #3060 also mentions number inputs but those work without any adaptations

**Important** : For this to work we need this fix (https://github.com/scality/core-ui/pull/234) to be included in the core-ui version used by metalk8s.

**Acceptance criteria**: 
* Sidebar is now hoverable
* Number inputs are working
* Alerts are filtered through the Healthselector component

---

Closes: #3060

### DIFF
--- a/ui/src/components/ActiveAlertsFilters.js
+++ b/ui/src/components/ActiveAlertsFilters.js
@@ -1,36 +1,12 @@
 import React from 'react';
-import styled from 'styled-components';
-import { Dropdown } from '@scality/core-ui';
+import { Healthselector } from '@scality/core-ui';
 import { useHistory, useRouteMatch } from 'react-router';
 import { useQuery } from '../services/utils';
-import { padding } from '@scality/core-ui/dist/style/theme';
-import { STATUS_WARNING, STATUS_CRITICAL } from '../constants.js';
-
-export const FilterIcon = styled.i`
-  color: ${(props) => {
-    const theme = props.theme.brand;
-    let color = theme.textPrimary;
-
-    switch (props.status) {
-      case STATUS_WARNING:
-        color = theme.warning;
-        break;
-      case STATUS_CRITICAL:
-        color = theme.danger;
-        break;
-      default:
-        color = theme.textPrimary;
-    }
-    return color;
-  }};
-  padding: ${padding.smaller};
-`;
-
-export const ActiveAlertsFilterWrapper = styled.div`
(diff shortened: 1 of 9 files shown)


ui/src/components/ActiveAlertsFilters.js
```diff
@@ -41,65 +17,42 @@ const ActiveAlertsFilter = (props) => {
...
-      <Dropdown
-        items={items}
-        text={dropDownLabel}
-        size="small"
-        data-cy="alert_filter"
-      />
-    </ActiveAlertsFilterWrapper>
+    <div data-cy="alert_filter">
+      <Healthselector items={items} />
+    </div>
```
ChengYanJin: In my opinion, we should add the `data-cy` directly to `<Healthselector/>` so we don't need to adjust the code in the test.

ChengYanJin (changes requested):
Was mainly focus on checking the test before. 
Code review, small change request.

ui/src/components/ActiveAlertsFilters.js
alexis-ld: @ChengYanJin That's what I did initially but seems `<Healthselector/>` is directly exported from core-ui the custom HTML attributes such as `data-cy` are not forwarded, that's why the tests were failing. Do you have an other way around that ? :)

ChengYanJin: @alexis-ld  one thing we can do is to add the spread syntax to expand the propery of the component in core-ui.
```js
function Healthselector({ size, items, ...rest }: Props) {
  return (
    <HealthselectorContainer className="sc-healthselector" {...rest}>
}
```

ChengYanJin (approved):

PR #3116, opened 2021-02-15
ui: Fix salt may return false when fetching IP interfaces

JBWatenbergScality:
**Component**:

ui

**Context**: 

When a cluster goes into an incorrect state, salt API may sends `false` when requesting IP interfaces of nodes which then prevent the UI to display the nodes.

**Acceptance criteria**: 

The UI should display the nodes even if we can't fetch their IPs.

### DIFF
--- a/ui/src/ducks/app/nodes.js
+++ b/ui/src/ducks/app/nodes.js
@@ -241,10 +241,6 @@ export const stopRefreshNodesAction = () => {
   return { type: STOP_REFRESH_NODES };
 };
 
-export const fetchNodesIPsInterfaceAction = () => {
-  return { type: FETCH_NODES_IPS_INTERFACES };
-};
-
 export const updateNodesIPsInterfacesAction = (payload) => {
   return { type: UPDATE_NODES_IPS_INTERFACES, payload };
 };
@@ -261,7 +257,7 @@ export const updateNodeObjectAction = (payload) => {
 export const clusterVersionSelector = (state) => state.app.nodes.clusterVersion;
 export const nodesRefreshingSelector = (state) => state.app.nodes.isRefreshing;
 export const isSaltAPIAuthenticatedSelector = (state) => state.login.salt;
-
+const nodeListSelector = (state) => state.app.nodes.list;
 // Sagas
 export function* fetchClusterVersion() {
   const result = yield call(CoreApi.getKubeSystemNamespace);
@@ -341,6 +337,7 @@ export function* fetchNodes() {
   }
   yield delay(1000); // To make sure that the loader is visible for at least 1s
   yield put(updateNodesAction({ isLoading: false }));
+  yield call(fetchNodesIPsInterface);
   return result;
 }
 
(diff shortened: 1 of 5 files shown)


ui/src/services/salt/api.js
```diff
@@ -64,7 +78,17 @@ export async function prepareEnvironment(environment, version) {
...
+};
+export async function getNodesIPsInterfaces(): Promise<{
+  return: [{ [nodeName: string]: boolean | IPInterfaces }],
+}> {
+  if (!saltApiClient) {
+    throw new Error('Salt api client should be defined.');
+  }
   return saltApiClient.post('/', {
     client: 'local',
     tgt: '*',
```
gdemonet: Maybe we should make sure we only fetch grains from nodes we know about (from K8s)

gdemonet: Could be done with:

```javascript
export async function getNodesIPsInterfaces(nodeNames: string[]): Promise<{
  return: [{ [nodeName: string]: boolean | IPInterfaces }],
}> {
  if (!saltApiClient) {
    throw new Error('Salt api client should be defined.');
  }
  return saltApiClient.post('/', {
    client: 'local',
    tgt: nodeNames.join(','),
    tgt_type: 'list',
    fun: 'grains.item',
    arg: [
      'metalk8s:control_plane_ip',
      'metalk8s:workload_plane_ip',
      'ip_interfaces',
    ],
  });
}
```

ui/src/services/NodeUtils.js
```diff
@@ -141,7 +142,19 @@ export const getNodeListData = createSelector(
...
+export const nodesCPWPIPsInterface = (
+  IPsInterfacesObject: IPInterfaces | boolean,
+): {
+  control_plane: { ip: string, interface: string },
+  workload_plane: { ip: string, interface: string },
+} => {
+  if (!IPsInterfacesObject) {
+    return {
+      control_plane: { ip: '', interface: '' },
+      workload_plane: { ip: '', interface: '' },
```
ChengYanJin: Same here ;)

ui/src/services/NodeUtils.js
```diff
@@ -141,7 +142,19 @@ export const getNodeListData = createSelector(
 //   control_plane: { ip: '10.0.1.42', interface: 'eth1'}
 //   workload_plane: { ip: '10.100.0.2', interface: 'eth3'},
 // }
-export const nodesCPWPIPsInterface = (IPsInterfacesObject) => {
+export const nodesCPWPIPsInterface = (
+  IPsInterfacesObject: IPInterfaces | boolean,
+): {
+  control_plane: { ip: string, interface: string },
+  workload_plane: { ip: string, interface: string },
```
ChengYanJin: Maybe the example above confused you, what we return is camelCase object ^^
Let's update the example as well :) 

```suggestion
  controlPlane: { ip: string, interface: string },
  workloadPlane: { ip: string, interface: string },
```

NicolasT:
I'm a bit surprised we need to call into Salt to retrieve this information? Isn't it already available on `node` objects? If not, shouldn't we add some labels on `Node` objects (`metalk8s.scality.com/workload-plane-ip` and alike) and fill these in when we deploy (and upgrade) a `Node`? Then the UI can simply get this from K8s.

Any time the UI calls SaltAPI (except for cluster expansion), we should question whether there should be no other API.

ui/src/ducks/app/nodes.js
```diff
@@ -484,19 +484,27 @@ export function* stopRefreshNodes() {
 
 export function* fetchNodesIPsInterface() {
   let result;
+
+  // Call K8s API to get the existing node to fetch grains
+  yield call(fetchNodes);
```
JBWatenbergScality: I think fetchNodesIPsInterface should be called when fetching nodes is finished instead of re-triggering an API call to kubernetes API

ChengYanJin (approved):

PR #3125, opened 2021-02-17
Improve robustness of salt-master and upgrade

TeddyAndrieux:
**Component**:

'salt', 'lifecycle'

**Context**: 

Upgrade get a lot of flakies on environment that are a bit slow

**Summary**:

- Backport https://github.com/scality/metalk8s/commit/9ee0c533d4336edabc5832f2351240cd8995c99b
- Bump timeout waiting for containers after local container engine upgrade from 60s to 120s
- Backport https://github.com/scality/metalk8s/commit/73835bef88a38db53330cb825c0d69e543386603
- Increase Salt master `sock_pool_size` (from 1 to 15) and `worker_threads` (from 5 to 10)

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # CHANGELOG
 
 ## Release 2.7.2 (in development)
+### Enhancements
+- Improve Salt master and cluster upgrade stability in slow environments
+  (PR [#3125](https://github.com/scality/metalk8s/pull/3125))
+
 ### Bug fixes
 - Embed `pause` image version 3.2 instead of 3.1 needed for MetalK8s to work
   offline (needed by containerd version superior to 1.4.0)
--- a/salt/_modules/cri.py
+++ b/salt/_modules/cri.py
@@ -6,6 +6,7 @@ import re
 import logging
 import time
 
+from salt.exceptions import CommandExecutionError
 import salt.utils.json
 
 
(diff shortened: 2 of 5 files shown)


salt/metalk8s/salt/minion/running.sls
```diff
@@ -5,7 +5,7 @@ Restart salt-minion:
 Wait until salt-minion restarted:
   test.configurable_test_state:
     - changes: False
-    - result: __slot__:salt:test.sleep(10)
+    - result: __slot__:salt:test.sleep(20)
```
gdemonet: Never fond of these, bumping timeouts is cool, because it only has a positive impact in worse scenarios, but bumping a sleep also has a negative impact in better scenarios.
Are you sure that's all we can do for this?
What if we had the minion write/touch a file on startup (maybe there's already something in place?), and we waited for such a file to be updated (with a timeout :wink:)?

TeddyAndrieux: Removed, you right it's not really make sense to bump this one here. Would need more investigation

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # CHANGELOG
 
 ## Release 2.7.2 (in development)
+### Enhancements
+- Improve salt-master and upgrade robustness on environment that are a bit slow
```
gdemonet: Since it's customer-facing:

```suggestion
- Improve Salt master and cluster upgrade stability in slow environments
```

gdemonet (changes requested):
Code is OK, just an edit to make for the changelog entry (sorry :innocent:)

CHANGELOG.md
TeddyAndrieux: Done

gdemonet (approved):
:shipit:

PR #3141, opened 2021-02-22
eve: Properly handle multi pods in NodeAffinity in the same ns

TeddyAndrieux:
In snapshot upgrade we need to delete NodeAffinity pods but several pods
can be in the same namespace and then the `node_aff_pods` variable
contains multiple pods separated by the space so we need to quote it in
the if to avoid
```
-bash: line 3: [: too many arguments
```

Sees: 27adae15d


gdemonet (approved):

gdemonet:
FYI, I did this with Python on some other project, much simpler (and more readable :wink:)

PR #3149, opened 2021-02-24
 Set up the util functions for alert related UI components

### LINKED ISSUE #3147: Implement util functions for alerts related UI components
**Component**: ui, monitoring

**Why this is needed**:
https://docs.google.com/document/d/1QrKJep8tQ31go7GV1YRGtfXEzg8Z6fcexTWoVCR2gCY/edit

**What should be done**:

Put in place the following utils functions: 

```js
getActiveAlerts(selectors: {[labelName: string]: string}): Promise<AlertList>

getHistoryAlerts(selectors: {[labelName: string]: string}, startDate: string, endDate?: string = 'now in iso format'): Promise<AlertList>

checkActiveAlertProvider(): Promise<"HEALTHY" | "CRITICAL">

checkPastAlertProvider(): Promise<{status: "HEALTHY" | "CRITICAL", retention: duration in s}> 
 => query /config to get retention

getStatus(alertList: AlertList, activeOn: string): "HEALTHY" | "WARNING" | "CRITICAL"

```

**Test plan**:
unit tests

ChengYanJin:
**Component**: ui, monitoring

**Context**:  
We want to add some util functions that will be used by all the alert related component as listed here:
https://docs.google.com/document/d/1QrKJep8tQ31go7GV1YRGtfXEzg8Z6fcexTWoVCR2gCY/edit
those components only need to pass the filter to the util function and will get the filtered alert and health.

**Summary**:
Add util functions `filterAlerts`, `getHealthStatus`  in alertUtils and unit tests.
The next step would be to implement the global fetch alert and replace the current implementation with the util functions. 

**Acceptance criteria**: 
Pass the unit tests.

_Keep the commit history just for easier review, will squash all the commits and redo in the end._

---

Closes: #3147


ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+  filters: {
+    [labelName: string]: string | string[],
+    parents?: string[],
+    selectors?: string[],
+  },
+): boolean => {
+  return Object.getOwnPropertyNames(filters)
+    .map((key) => {
+      if (!filters[key] || !labels[key] || !isObjectHasProperty(labels, key))
+        return false;
```
JBWatenbergScality:
```suggestion
      if (!filters[key] || !labels[key]) {
        return false;
      }
```

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+// Sort the alerts base on the `severity`
+export const sortAlerts = (alerts: AlertList): AlertList => {
+  return alerts.sort(function (a, b) {
+    return compareHealth(a.labels.severity, b.labels.severity);
+  });
+};
+
+export const isObjectHasProperty = (obj: {}, property: string): boolean => {
+  return Object.prototype.hasOwnProperty.call(obj, property);
+};
```
JBWatenbergScality: I'm not sure this method is necessary. as label values can only be of type string, wouldn't it be easier to use `!!obj.property` ?

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return {
+      summary: alert.annotations.summary,
+      description: alert.annotations.description || alert.annotations.message,
+      startsAt: alert.startsAt,
+      endsAt: alert.endsAt || new Date().toISOString(),
+      severity: alert.labels.severity,
+      documentationUrl: alert.annotations.runbook_url,
+      labels: { ...alert.labels, parents, selectors },
+    };
```
JBWatenbergScality: In order to avoid introducing a new condition + variable declaration every time we need a new array, I would purpose the following implementation :

```suggestion
    return {
      summary: alert.annotations.summary,
      description: alert.annotations.description || alert.annotations.message,
      startsAt: alert.startsAt,
      endsAt: alert.endsAt || new Date().toISOString(),
      severity: alert.labels.severity,
      documentationUrl: alert.annotations.runbook_url,
      labels: Object.fromEntries(Object.entries(alert.labels).map(([key, value]) => {
        if (key === 'parents' || key === 'selectors') {
           return [key, value.split(',')];
        }
        return [key, value];
      })),
    };
```

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+            }) !== undefined
+          );
+        } else return filters[key].includes(labels[key]);
+      } else {
+        return labels[key].includes(filters[key]);
+      }
+    })
+    .reduce((boolA, boolB) => {
+      return boolA && boolB;
+    });
```
JBWatenbergScality: I would add an explicit initial value here to ease readability
```suggestion
}, true);
```

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+): boolean => {
+  return Object.getOwnPropertyNames(filters)
+    .map((key) => {
+      if (!filters[key] || !labels[key] || !isObjectHasProperty(labels, key))
+        return false;
+      if (Array.isArray(filters[key])) {
+        if (Array.isArray(labels[key])) {
+          return (
+            filters[key].find((val) => {
+              return labels[key].indexOf(val) >= 0;
```
JBWatenbergScality:
```suggestion
labels[key].includes(val)
```

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+
+// TODO: Call Loki Endpoint
+export const getHistoryAlerts = (
+  start: string, //TODO: maybe default last 7 days
+  end?: string = new Date().toISOString(),
+  filters: {
+    [labelName: string]: string | string[],
+    parents?: string[],
+    selectors?: string[],
+  },
```
JBWatenbergScality:
```
{
    [labelName: string]: string | string[],
    parents?: string[],
    selectors?: string[],
  }
```

This type could be extracted in order to avoid redefining it for each function

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,208 @@
...
+    return isAlertSelected(alert.labels, filters);
+  });
+};
+
+// check if the given time is DURING the start and end or OUT.
+export const firingTime = (
+  start: string,
+  end: string,
+  on: string,
+): 'OUT' | 'DURING' => {
```
JBWatenbergScality: You may want to represent the output of this function using a boolean type and rename your function to `dateIsBetween(date: string, start: string, end: string): boolean`

ui/src/services/alertUtils.js
ChengYanJin: Hi @JBWatenbergScality, 

Thanks for the proposal!
However, I noticed that `Object.fromEntries` requires Node 12.0.0. I would say avoid using this method if it's possible given our current env requirement is Node 10.16. WDYT?

> Running the platform UI locally
> Requirements
> Node.js, 10.16
https://metal-k8s.readthedocs.io/en/latest/developer/running/ui.html?highlight=UI%20locally#running-the-platform-ui-locally

JBWatenbergScality: This only impacts tests execution which can import a polyfill to workarround the issue. Browser support of Object.fromEntries covers all the browser we want to support I believe : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#browser_compatibility. Also node v10 will be end of life by the end of April (https://nodejs.org/en/about/releases/)  I believe we should study an upgrade soon.

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,186 @@
...
+    .reduce((boolA, boolB) => {
+      return boolA && boolB;
+    }, true);
+};
+
+// TODO: Call Loki Endpoint
+export const getHistoryAlerts = (
+  start: string, //TODO: maybe default last 7 days
+  end?: string = new Date().toISOString(),
+  filters: FilterLabels,
```
JBWatenbergScality: IMO required parameters should be defined before optional parameters, else when you actually used the function you are being forced to pass those optional parameters (to nullish values if you want to use the default values) in order to pass your required parameters
```suggestion
  filters: FilterLabels,
  start: string, //TODO: maybe default last 7 days
  end?: string = new Date().toISOString(),
```

ui/src/services/alertmanager/api.js
```diff
@@ -28,15 +29,54 @@ export type PrometheusAlert = {
...
+  id: string,
+  summary: string,
+  description: string,
+  startsAt: string,
+  endsAt: string,
+  severity: string,
+  documentationUrl?: string,
+  labels: AlertLabels,
+  originalAlert: Object,
+}[];
```
JBWatenbergScality: Nit: I would not define AlertList type in alertmanager as when fetching alert history we would also need to convert passed alerts to this format. I think it would be better to move it to alert utils.
<hr/>

IMO it could make sens to only type the `Alert` at the item level and not the array because in some places we may want to manipulate items instead of arrays
<hr/>

originalAlert is here of type PrometheusAlert, when alert history would be in place we can type it with an union of PrometheusAlert and the type of the passed Alert service provider

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,39 @@
...
+  lokiApiClient = new ApiClient({ apiUrl });
+}
+
+export function getAlertsLoki(start: string, end: string) {
+  if (!lokiApiClient) {
+    throw new Error('lokiApiClient should be defined');
+  }
+  // TODO: Hardcode for now
+  const query =
+    "{namespace='metalk8s-monitoring',container='alertmanager-logger'}";
```
JBWatenbergScality: As we want to fetch all the alerts cross namespaces I think we can drop this query. Alert filtering will be performed through our utils functions

ui/src/services/alertUtils.js
JBWatenbergScality: I think alert fetching and filtering should be separated as in the end we are going to fetch all the alerts one time for the whole application

ui/src/services/alertUtils.js
```diff
@@ -0,0 +1,186 @@
...
+): AlertList => {
```
JBWatenbergScality: It should return a time serie of Alert instead of an AlertList

ui/src/services/loki/api.js
ChengYanJin: According to @alexandre-allard-scality 
We will need to specify the container name. Otherwise, we'll end up with the logs of everything.

But yes, you are right, we don't need to specify namespace here.

gdemonet: Let's keep the namespace, just in case, there's no harm in being specific here. We're reading logs from a specific pod.

ChengYanJin: Hi, @gdemonet We were thinking about exporting the `getAlertsLoki` function through module federation so that other UIs could benefit from the alert fetching function. (@JBWatenbergScality can you confirm this?)

And then apply the filter `{namespace: 'metalk8s-monitoring'}` to scope the alerts for a specific application.
WDYT?

JBWatenbergScality: I confirm the idea would be to use the same function in others scopes so if this namespace filter in query parameters refrain us to do so I think we should get rid of it.

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,82 @@
...
+  },
+};
+
+// by default to get the last 7day alerts from Loki
+export function getLast7DaysAlerts(): LokiQueryResult {
+  if (!lokiApiClient) {
+    throw new Error('lokiApiClient should be defined');
+  }
+
+  const query = "{container='alertmanager-logger'}";
```
JBWatenbergScality: after discussing with @gdemonet we should have kept the namespace (it is for retrieving the log from a specific loki pod and has nothing to do with namespacing in alert manager which confused us)

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,82 @@
...
+      return resolve;
+    });
+}
+
+// customise the `start` and `end` time to retrieve the history alerts
+export function getAlertsLoki(start: string, end: string): LokiQueryResult {
+  if (!lokiApiClient) {
+    throw new Error('lokiApiClient should be defined');
+  }
+  const query = "{container='alertmanager-logger'}";
```
JBWatenbergScality: same here, maybe this query can be mutualized

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,82 @@
...
+
+  return lokiApiClient
+    .get(`/api/v1/query_range?start=${start}&end=${end}&query=${query}`)
+    .then((resolve) => {
+      if (resolve.error) {
+        throw resolve.error;
+      }
+      return resolve;
+    });
+}
```
JBWatenbergScality: In order to avoid code duplication, it should be implemented like so :
```suggestion
export function getLast7DaysAlerts(): LokiQueryResult {
 
  const start = new Date(
    new Date().getTime() - 7 * 24 * 60 * 60 * 1000,
  ).toISOString();
  const end = new Date().toISOString();

  return getAlertsLoki(start, end);
}
```

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,82 @@
...
+
+// customise the `start` and `end` time to retrieve the history alerts
+export function getAlertsLoki(start: string, end: string): LokiQueryResult {
```
JBWatenbergScality: It should return a promise, also please ensure flow was reporting this issue as it should have reported it.
```suggestion
export function getAlertsLoki(start: string, end: string): Promise<LokiQueryResult> {
```

ui/src/services/loki/api.js
```diff
@@ -0,0 +1,82 @@
...
+    resultType: 'matrix' | 'streams',
+    result: MetrixValue | StreamValue,
+    // Those statistics allow users to understand the amount of data processed and at which speed.
+    // https://grafana.com/docs/loki/latest/api/#Statistics
+    stats: {},
+  },
+};
+
+// by default to get the last 7day alerts from Loki
+export function getLast7DaysAlerts(): LokiQueryResult {
```
JBWatenbergScality:
```suggestion
export function getLast7DaysAlerts(): Promise<LokiQueryResult> {
```

ui/src/services/loki/api.js
ChengYanJin: OK. Thanks for the update!!

ui/src/services/alertUtils.js
```diff
@@ -181,3 +182,42 @@ export const getHealthStatus = (
...
+  });
+
+  return alerts.reduce((acc, current) => {
+    const x = acc.find((item) => item.id === current.id);
+    if (!x) {
+      return acc.concat([current]);
+    } else {
+      return acc;
+    }
+  }, []);
```
JBWatenbergScality: Maybe in order to simplify this we can index the alerts by their ids and then keep the values of this object.

```suggestion
  const alerts = streamValues.reduce((agg, value) => {
    const alert = JSON.parse(value.values[1]);
    return {
       ...agg,
       [alert.fingerprint]: {
          id: alert.fingerprint,
          summary: alert.annotations.summary,
          description: alert.annotations.description || alert.annotations.message,
          startsAt: alert.startsAt,
          endsAt: alert.endsAt || new Date().toISOString(),
          severity: alert.labels.severity,
          documentationUrl: alert.annotations.runbook_url,
          labels: {
            ...alert.labels,
            parents:
              (alert.annotations.parents && alert.annotations.parents.split(',')) ||
              [],
            selectors:
              (alert.annotations.selectors &&
                alert.annotations.selectors.split(',')) ||
              [],
       }
    },
      originalAlert: alert,
    };
  }, {});

  return Object.values(alerts);
```

ui/src/services/loki/api.js
```diff
@@ -22,44 +21,39 @@ type StreamValue = {
...
-  return lokiApiClient
-    .get(`/api/v1/query_range?start=${start}&end=${end}&query=${query}`)
-    .then((resolve) => {
-      if (resolve.error) {
-        throw resolve.error;
-      }
-      return resolve;
-    });
+  return getAlertsLoki(start, end).then((result) =>
+    removeWarningAlerts(formatHistoryAlerts(result.data.result)),
```
JBWatenbergScality: Maybe this can be done in getAlertsLoki in order to avoid having to do it in calling methods.

JBWatenbergScality (commented):
I think it is missing some tests for `formatHistoryAlerts`. The rest is ok for me 👍

JBWatenbergScality (approved):

PR #3120, opened 2021-02-16
Bump pause image to 3.2

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Invalid pause image version injected in containerd

**Summary**:

- Rewrite a bit the pause injection
- Bump pause image to 3.2

---

PR #3151, opened 2021-02-25
Restore CI in offline mode

alexandre-allard:
**Component**: ci

**Context**:
We found some bug lately (#3120), not catched by the CI, because we're no longer testing in offline environments since #1928.

**Summary**:
Restore the offline mode by configuring some security groups on Openstack and using a proxy for the YUM repositories.

**Acceptance criteria**:

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -597,6 +597,8 @@ models:
         TF_VAR_rhsm_username: "%(secret:rhel_ci_login)s"
         TF_VAR_rhsm_password: "%(secret:rhel_ci_password)s"
         TF_VAR_debug: "%(prop:metalk8s_debug:-false)s"
+        TF_VAR_offline: "%(prop:offline:-true)s"
+        TF_VAR_use_proxy: "%(prop:use_proxy:-true)s"
         MAX_RETRIES: "3"
       workdir: build/eve/workers/openstack-terraform/terraform/
       haltOnFailure: true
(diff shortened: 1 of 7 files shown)


eve/workers/openstack-terraform/terraform/common.tf
```diff
@@ -23,6 +23,21 @@ variable "nodes_count" {
...
 }
 
+variable "offline" {
+  type    = bool
+  default = true
+}
+
+variable "proxy_ip" {
+  type    = string
+  default = "10.100.4.67"
```
TeddyAndrieux: We cannot rely on ScalityCloud dns `proxy-cache` ? IP is mandatory for security group rules right ?

TeddyAndrieux (commented):
LGTM

It seems RHEL8 does not work offline (don't know why just build status)

eve/workers/openstack-terraform/terraform/common.tf
gdemonet: :+1: - the `dns_a_record_set` data source cannot resolve it, as you did for `subscription.rhsm.redhat.com`?

alexandre-allard: Yes, sure, we could totally do that.

eve/main.yml
```diff
@@ -1641,6 +1643,14 @@ stages:
...
           property: os
           value: "rhel-8"
+      - SetProperty:
+          name: Set offline property to false
+          property: offline
+          value: false
+      - SetProperty:
+          name: Set use_proxy property to false
+          property: use_proxy
+          value: false
```
gdemonet: Worth adding a comment + link to a ticket (not sure if we've created one yet) to remember why this is disabled and what we should do to re-enable it.

gdemonet (commented):
Except for a small comment on disabling offline for RHEL 8, I think this is good to merge :slightly_smiling_face: 
Just wondering, how does this work with already existing snapshots which were not configured to run offline? (I guess it still works, but worth checking)

gdemonet (approved):

gdemonet (dismissed):

gdemonet (approved):

PR #3154, opened 2021-02-26
UI make navbar options configurable and filter menu items according to user groups

### LINKED ISSUE #3159: shell-ui: make navbar options configurable
**Component**:

shell-ui

**Why this is needed**:

The navbar should display configurable menu options according to deployed solutions.

**What should be done**:

- [x] add an `options` entry in the shell ui config map + exposes this as an optional entry of the navbar
- [x] map those options to menu entries
- [x] filters the entries according to user OIDC group claim

**Test plan**:

unit test

JBWatenbergScality:
**Component**:

ui, shell-ui

**Context**: 

Menu items displayed in the navbar should be configurable and filtered according to user groups (as provided as part of OIDC claims).

**Acceptance criteria**: 

Unit tests

Closes : https://github.com/scality/metalk8s/issues/3159

### DIFF
--- a/shell-ui/src/NavBar.js
+++ b/shell-ui/src/NavBar.js
@@ -2,10 +2,17 @@
 import CoreUINavbar from '@scality/core-ui/dist/components/navbar/Navbar.component';
 import { useAuth } from 'oidc-react';
 import { useLayoutEffect, useState } from 'react';
-import type { Options, SolutionsNavbarProps } from './index';
+import type {
+  Options,
+  SolutionsNavbarProps,
+  TranslationAndGroups,
+} from './index';
 import type { Node } from 'react';
 import { logOut } from './auth/logout';
-import { options } from 'jest-runtime/build/cli/args';
+import {
+  isEntryAccessibleByTheUser,
+  normalizePath,
+} from './auth/permissionUtils';
 
 export const LoadingNavbar = (): Node => (
   <CoreUINavbar role="navigation" tabs={[{ title: 'loading' }]} />
@@ -37,7 +44,9 @@ const translateOptionsToMenu = (
             };
           } catch (e) {
             throw new Error(
-              `[navbar][config] Options path should be provided thanks to fully qualified urls such as {protocol}://{host}{path}?{queryParams} but got "${path}" in ${section}[${i}] `,
+              `[navbar][config] Invalid path specified in "options.${section}": "${path}" ` +
+                '(keys must be defined as fully qualified URLs, ' +
+                'such as "{protocol}://{host}{path}?{queryParams}")',
             );
           }
         },
@@ -48,7 +57,16 @@ const translateOptionsToMenu = (
 export const Navbar = ({ options }: { options: Options }): Node => {
   const auth = useAuth();
 
-  const tabs = translateOptionsToMenu(options, 'main');
+  const userGroups: string[] = auth.userData?.profile?.groups || [];
+
+  const tabs = translateOptionsToMenu(
+    options,
+    'main',
+    (path, translationAndGroup) => ({
+      link: <a href={path}>{translationAndGroup.en}</a>,
+    }),
+    userGroups,
+  );
 
   const rightActions = [
     {
@@ -56,7 +74,17 @@ export const Navbar = ({ options }: { options: Options }): Node => {
       text: auth.userData?.profile.name || '',
       icon: <i className="fas fa-user" />,
       items: [
-        ...translateOptionsToMenu(options, 'subLogin'),
+        ...translateOptionsToMenu(
+          options,
+          'subLogin',
+          (path, translationAndGroup) => ({
+            label: translationAndGroup.en,
+            onClick: () => {
+              location.href = path;
+            },
+          }),
+          userGroups,
+        ),
         {
           label: 'Log out',
           onClick: () => {
@@ -66,6 +94,7 @@ export const Navbar = ({ options }: { options: Options }): Node => {
       ],
     },
   ];
+
   return (
     <CoreUINavbar rightActions={rightActions} tabs={tabs} role="navigation" />
   );
--- a/shell-ui/src/NavBar.js
+++ b/shell-ui/src/NavBar.js
@@ -2,64 +2,61 @@
 import CoreUINavbar from '@scality/core-ui/dist/components/navbar/Navbar.component';
 import { useAuth } from 'oidc-react';
 import { useLayoutEffect, useState } from 'react';
-import type { SolutionsNavbarProps } from './index';
+import type { Options, SolutionsNavbarProps } from './index';
 import type { Node } from 'react';
 import { logOut } from './auth/logout';
+import { options } from 'jest-runtime/build/cli/args';
 
-export const LoadingNavbar = (): Node => <CoreUINavbar role="navigation" tabs={[{title: 'loading'}]} />;
+export const LoadingNavbar = (): Node => (
+  <CoreUINavbar role="navigation" tabs={[{ title: 'loading' }]} />
+);
 
-export const Navbar = ({
-  options,
-}: {
-  options: $PropertyType<SolutionsNavbarProps, 'options'>,
-}): Node => {
+const translateOptionsToMenu = (
+  options: Options,
+  section: 'main' | 'subLogin',
+  renderer: (
+    path: string,
+    translationAndGroup: TranslationAndGroups,
+  ) => { link: Node } | { label: string, onClick: () => void },
+  userGroups: string[],
+) => {
+  const normalizedLocation = normalizePath(location.href);
+  return (
+    Object.entries(options[section])
+      //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, TranslationAndGroups]
+      .filter((entry: [string, TranslationAndGroups]) =>
+        isEntryAccessibleByTheUser(entry, userGroups),
+      )
+      .map(
+        //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, TranslationAndGroups]
+        ([path, translationAndGroup]: [string, TranslationAndGroups], i) => {
+          try {
+            return {
+              ...renderer(path, translationAndGroup),
+              selected: normalizedLocation === normalizePath(path),
+            };
+          } catch (e) {
+            throw new Error(
+              `[navbar][config] Options path should be provided thanks to fully qualified urls such as {protocol}://{host}{path}?{queryParams} but got "${path}" in ${section}[${i}] `,
+            );
+          }
+        },
+      )
+  );
+};
+
+export const Navbar = ({ options }: { options: Options }): Node => {
   const auth = useAuth();
 
-  const tabs = [
-    {
-      selected: true,
-      title: 'Overview',
-      onClick: () => console.log('Overview'),
-    },
-  ];
+  const tabs = translateOptionsToMenu(options, 'main');
+
   const rightActions = [
-    {
-      type: 'dropdown',
-      text: 'FR',
-      icon: <i className="fas fa-globe" />,
-      items: [
-        {
-          label: 'English',
-          name: 'EN',
-          onClick: () => console.log('en'),
-        },
-      ],
-    },
-    {
-      type: 'dropdown',
-      icon: <i className="fas fa-question-circle" />,
-      items: [
-        { label: 'About', onClick: () => console.log('About clicked') },
-        {
-          label: 'Documentation',
-          onClick: () => console.log('Documentation clicked'),
-        },
-        {
-          label: 'Onboarding',
-          onClick: () => console.log('Onboarding clicked'),
-        },
-      ],
-    },
-    {
-      type: 'button',
-      icon: <i className="fas fa-sun" />,
-      onClick: () => console.log('Theme toggle clicked'),
-    },
     {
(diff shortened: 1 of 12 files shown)


shell-ui/src/NavBar.js
```diff
@@ -2,64 +2,95 @@
...
+);
 
-export const Navbar = ({
-  options,
-}: {
-  options: $PropertyType<SolutionsNavbarProps, 'options'>,
-}): Node => {
+const translateOptionsToMenu = (
+  options: Options,
+  entries: 'main' | 'subLogin',
```
gdemonet: Not sure about the "entries" argument name here. Given the allowed values, I imagine we're describing "groups of entries", so maybe use category or section instead?

shell-ui/src/NavBar.js
```diff
@@ -2,64 +2,95 @@
...
+      .filter((entry: [string, TranslationAndGroups]) =>
+        isEntryAccessibleByTheUser(entry, userGroups),
+      )
+      .map(
+        //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, TranslationAndGroups]
+        ([path, translationAndGroup]: [string, TranslationAndGroups], i) => {
+          try {
+            return {
+              ...renderer(path, translationAndGroup),
+              selected: normalizedLocation === normalizePath(path),
```
gdemonet: Could be interesting to warn somehow (likely in console) when not a single entry matches the current location.

shell-ui/src/Navbar.spec.js
```diff
@@ -62,7 +62,7 @@ describe('navbar', () => {
             />
```
ChengYanJin: Would be great to format the code^^

shell-ui/src/setupTests.js
```diff
@@ -2,6 +2,36 @@ import 'whatwg-fetch';
 import "regenerator-runtime/runtime"
```
ChengYanJin: Missing the auto-format here too.

shell-ui/src/NavBar.js
gdemonet: Also, from what I can see, normalized paths are simply `p.origin + p.pathname`. Isn't this going to break when a path is only linking to the root of a single-page app, and this app then manipulates the URL for virtual routes?

shell-ui/src/NavBar.js
```diff
@@ -2,64 +2,95 @@
...
+            };
+          } catch (e) {
+            throw new Error(
+              `[navbar][config] Options path should be provided thanks to fully qualified urls such as {protocol}://{host}{path}?{queryParams} but got "${path}" in ${entries}[${i}] `,
```
gdemonet: Three comments:

- isn't Prettier complaining about the line-length? I don't like scrolling horizontally :stuck_out_tongue_winking_eye: 
- wording can be improved:

```javascript
throw new Error(
  `[navbar][config] Invalid path specified in "options.${section}": "${path}" `
  + '(keys must be defined as fully qualified URLs, '
  + 'such as "{protocol}://{host}{path}?{queryParams}")'
)
```

- showing the index of an entry in an object isn't really useful, right? Or are JS objects ordered somehow?

shell-ui/src/NavBar.js
```diff
@@ -2,64 +2,95 @@
...
+          (path, translationAndGroup) => ({
+            label: translationAndGroup.en,
+            onClick: () => {
+              location.href = path;
+            },
+          }),
+          userGroups,
+        ),
         {
           label: 'Log out',
```
gdemonet, line 101: I'm wondering (that's likely just a design question): do we want the `Log out` entry to always be the last? I'd instinctively put it first, but maybe my instincts are wrong :sweat_smile:

shell-ui/src/Navbar.spec.js
```diff
@@ -123,12 +123,59 @@ describe('navbar', () => {
...
+                        "http://localhost:8082/test":{ "en": "Test", "fr": "Test" }
+                    },
+                    "subLogin": {}
+                })
+            }
+            />
+        )
+        //E
+        await waitForLoadingToFinish();
+        //V
```
gdemonet: Are those comments meaning something? Never seen them before

shell-ui/src/index.js
```diff
@@ -23,7 +28,7 @@ export type SolutionsNavbarProps = {
   'response-type'?: string,
   'redirect-url'?: string,
   'config-url'?: string,
-  options?: Options,
+  options?: string,
```
gdemonet, line 31: Maybe a stupid question: why are we moving from a typed object to some free-form string? Is it supposed to be easier when working with WebComponents?

shell-ui/src/auth/permissionUtils.js
```diff
@@ -0,0 +1,40 @@
+//@flow
+import type { Options, TranslationAndGroups } from '../index';
+
+export const isEntryAccessibleByTheUser = (
+  [path, translationAndGroup]: [string, TranslationAndGroups],
+  userGroups: string[],
+): boolean => {
+  return (
+    (translationAndGroup.groups &&
+      translationAndGroup.groups
```
gdemonet: Can't you use optional chaining?

shell-ui/src/auth/permissionUtils.js
```diff
@@ -0,0 +1,40 @@
...
+        .map((group) => userGroups.includes(group))
+        .reduce(
+          (aggregatedPredicates, current) => aggregatedPredicates && current,
+          true,
+        )) ||
+    !translationAndGroup.groups
+  );
```
gdemonet: Could be written more simply with `Array.prototype.every`:

```suggestion
  return (
    translationAndGroup.groups?.every(
      (group) => userGroups.includes(group)
    ) ?? true
  );
```

(one nice benefit of `every` is that it will automatically return `true` if your array is empty :wink:)

shell-ui/src/auth/permissionUtils.js
```diff
@@ -0,0 +1,40 @@
...
+export const getAccessiblePathsFromOptions = (
+  options: Options,
+  userGroups: string[],
+): string[] => {
+  return [...Object.entries(options.main), ...Object.entries(options.subLogin)]
+    //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, TranslationAndGroups]
+    .filter((entry: [string, TranslationAndGroups]) =>
+      isEntryAccessibleByTheUser(entry, userGroups),
+    )
+    .map(([path]) => path);
```
gdemonet: This is already a builtin:

```suggestion
    .flat();
```

shell-ui/src/index.js
```diff
@@ -10,11 +10,16 @@ import { QueryClient, QueryClientProvider, useQuery } from 'react-query';
 import { LoadingNavbar, Navbar } from './NavBar';
 import { UserDataListener } from './UserDataListener';
 import { logOut } from './auth/logout';
+import { getAccessiblePathsFromOptions, isPathAccessible } from './auth/permissionUtils';
```
gdemonet: Unused imports?

shell-ui/src/auth/permissionUtils.js
```diff
@@ -0,0 +1,40 @@
...
+        .map((group) => userGroups.includes(group))
+        .reduce(
+          (aggregatedPredicates, current) => aggregatedPredicates && current,
+          true,
+        )) ||
+    !translationAndGroup.groups
+  );
+};
+
+export const getAccessiblePathsFromOptions = (
```
gdemonet, line 14: Is this method ever used (outside of tests :sweat_smile:)?

shell-ui/src/setupTests.js
```diff
@@ -2,6 +2,36 @@ import 'whatwg-fetch';
...
+
+if (!Array.prototype.flatMap) {
+	Object.defineProperty(Array.prototype, 'flatMap', {
+		configurable: true,
+		value: function flatMap (callback) {
+			return Array.prototype.map.apply(this, arguments).flat();
+		},
+		writable: true
+	});
+}
```
gdemonet: Why do we need those polyfills? I don't even see those methods used in the code :/

shell-ui/src/auth/permissionUtils.js
gdemonet: Ah, I was reading through the commits, and it's used only for pre-fetching all accessible paths.

gdemonet (changes requested):
Some small code improvements, behaviour looks good to me :+1:

shell-ui/src/NavBar.js
JBWatenbergScality: Github put it at the end : 
image

@Cuervino do you have an opinion on this ?

shell-ui/src/Navbar.spec.js
JBWatenbergScality: I used them to segregate my tests in 3 steps : 
 - //S = //Setup, set all the tests dependencies
 - //E = //Exercise, act / perform the action which is being tested
 - //V = //Verify, assert on the result of the action

shell-ui/src/index.js
JBWatenbergScality: WebComponents attributes can sadly not accept object as attributes because the resulting xhtml markup would then be invalid. A common trick is to serialize the object as JSON to enable passing objects to web components 😢

shell-ui/src/setupTests.js
JBWatenbergScality: I was using flatMap at some point in my implementation and as tests runs using node 10 it was not available, I will clean this up 👍

shell-ui/src/NavBar.js
JBWatenbergScality: You are right, I think I will add a regex in options to let solutions configure which routes should be considered as selected and maybe use this implementation as the default one if no regex is provided.

shell-ui/src/auth/permissionUtils.spec.js
```diff
@@ -0,0 +1,114 @@
...
+  it('should normalize path', () => {
+    expect(normalizePath('http://fake/groupPath')).toEqual(
+      'http://fake/groupPath',
+    );
+    expect(normalizePath('http://fake/path/subPath?a=test#test')).toEqual(
+      'http://fake/path/subPath',
+    );
+  });
+
+  it('should normalize path throw if invalid url', () => {
```
ChengYanJin:
```suggestion
  it('should throw if the path is an invalid url', () => {
```

shell-ui/src/NavBar.js
gdemonet: Indeed, maybe assume that `path` is always a regex (not sure how to handle escapes though).

shell-ui/src/index.js
gdemonet: Oh :confused: OK

shell-ui/src/setupTests.js
gdemonet: Argh, but since this project (shell-ui) is separate from the main MetalK8s UI, why aren't we already using Node 14.x? :wink:

shell-ui/src/NavBar.js
Cuervino: I checked on few apps: on most of them, the "Log out" option is at the last position of the menu list.
Having it on the last position is also nice for avoiding "Accidental logouts" (harder to misclick when the "log out" is further from the first click on the icon for accessing the menu).
So for these 2 reasons (**Avoid Accidental logouts** and **Follow standards**), my opinion is we should keep it at the last position.

shell-ui/src/auth/permissionUtils.js
JBWatenbergScality: I can't use flat here because I actually want to destructure the first entry of Object entries to retrieve only the `path`.

gdemonet: Ah right, read too fast :+1:

shell-ui/src/NavBar.js
```diff
@@ -2,64 +2,101 @@
...
+        //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, TranslationAndGroups]
+        ([path, translationAndGroup]: [string, TranslationAndGroups], i) => {
+          try {
+            return {
+              ...renderer(path, translationAndGroup),
+              selected: translationAndGroup.activeIfMatches
+                ? new RegExp(translationAndGroup.activeIfMatches).test(
+                    location.href,
+                  )
+                : normalizedLocation === normalizePath(path),
```
gdemonet, line 49: I'm afraid this default isn't really useful, since it will never match on "sub-paths"...
Maybe `normalizedLocation.startsWith(normalizePath(path))` instead?

gdemonet (approved):
I'm OK with merging in this state :ok_hand: 

For answering "which UI is active at the moment?", it will likely become easier once we switch tables and get the Shell to include other UIs instead of the opposite.

shell-ui/src/NavBar.js
JBWatenbergScality: I also thought about switching to startsWith but IMO it is not a correct approach because we can for example have a menu entry for `/` and for `/alerts` on the same UI where this default will make both entries selected. I think an exact match is the strongest default to avoid duplicated selected entries that could mislead the user.

gdemonet: Mmh, right. So we'd need to sort by specificity of configured paths to know "which Ingress is catching the current page", I guess.

PR #3166, opened 2021-03-03
[UI] Apply util functions for the existing alert related components

ChengYanJin:
**Component**: ui, loki

**Context**: 
Apply the `filtersAlerts` to the existing alert/health related component.

**Summary**:
The component(NodePartitionsTable) fetches alerts by react-query could benefit from the useAlerts() hook. 
For the rest components, we still retrieve the alerts by selectors and then apply the util function. 

Set up Loki service to retrieve the alert from Loki API tested by Alexandre's platform.

**Acceptance criteria**: 
UI should work as before.

### DIFF
--- a/ui/src/components/NodePartitionTable.js
+++ b/ui/src/components/NodePartitionTable.js
@@ -10,8 +10,14 @@ import {
   queryNodeFSSize,
 } from '../services/prometheus/fetchMetrics';
 import CircleStatus from './CircleStatus';
-import { getAlerts } from '../services/alertmanager/api';
 import { getNodePartitionsTableData } from '../services/NodeVolumesUtils';
+import { useAlerts } from '../containers/AlertProvider';
+import {
+  NODE_FILESYSTEM_SPACE_FILLINGUP,
+  NODE_FILESYSTEM_ALMOST_OUTOF_SPACE,
+  NODE_FILESYSTEM_FILES_FILLINGUP,
+  NODE_FILESYSTEM_ALMOST_OUTOF_FILES,
+} from '../constants';
 import { intl } from '../translations/IntlGlobalProvider';
 
 const HeadRow = styled.tr`
@@ -98,6 +104,15 @@ const LoaderContainer = styled.div`
 
 const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
   const theme = useTheme();
+  const { data: alertNF } = useAlerts({
+    alertname: [
+      NODE_FILESYSTEM_SPACE_FILLINGUP,
+      NODE_FILESYSTEM_ALMOST_OUTOF_SPACE,
+      NODE_FILESYSTEM_FILES_FILLINGUP,
+      NODE_FILESYSTEM_ALMOST_OUTOF_FILES,
+    ],
+  });
+
   const { data: partitions, status } = useQuery(
     ['nodeDevices', instanceIP],
     useCallback(
@@ -105,8 +120,7 @@ const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
         Promise.all([
           queryNodeFSUsage(instanceIP),
           queryNodeFSSize(instanceIP),
-          getAlerts(),
-        ]).then(([nodeFSUsageResult, nodeFSSizeResult, alerts]) => {
+        ]).then(([nodeFSUsageResult, nodeFSSizeResult]) => {
           if (
             nodeFSUsageResult.status === 'success' &&
             nodeFSSizeResult.status === 'success' &&
@@ -116,11 +130,11 @@ const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
             return getNodePartitionsTableData(
               nodeFSUsageResult.data.result,
               nodeFSSizeResult.data.result,
-              alerts,
+              alertNF,
             );
           }
         }),
-      [instanceIP],
+      [alertNF, instanceIP],
     ),
   );
 
--- a/ui/src/components/NodePartitionTable.js
+++ b/ui/src/components/NodePartitionTable.js
@@ -17,6 +17,7 @@ import {
   NODE_FILESYSTEM_ALMOST_OUTOF_SPACE,
   NODE_FILESYSTEM_FILES_FILLINGUP,
   NODE_FILESYSTEM_ALMOST_OUTOF_FILES,
+  PORT_NODE_EXPORTER,
 } from '../constants';
 import { intl } from '../translations/IntlGlobalProvider';
 
@@ -104,15 +105,17 @@ const LoaderContainer = styled.div`
 
 const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
   const theme = useTheme();
-  const { data: alertNF } = useAlerts({
+  const alertList = useAlerts({
     alertname: [
       NODE_FILESYSTEM_SPACE_FILLINGUP,
       NODE_FILESYSTEM_ALMOST_OUTOF_SPACE,
       NODE_FILESYSTEM_FILES_FILLINGUP,
       NODE_FILESYSTEM_ALMOST_OUTOF_FILES,
     ],
+    instance: `${instanceIP}:${PORT_NODE_EXPORTER}`,
   });
 
+  const alertNF = alertList && alertList.alerts;
   const { data: partitions, status } = useQuery(
     ['nodeDevices', instanceIP],
     useCallback(
@@ -134,7 +137,10 @@ const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
             );
           }
         }),
-      [alertNF, instanceIP],
+      // disable the eslint checking because it requires `alertNF` to be in the dependency list,
+      // and since react is not performing a deep comparison, so useCallback would be recalled everytime useAlerts refetch the alerts.
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [instanceIP, JSON.stringify(alertNF)],
     ),
   );
 
--- a/ui/src/components/NodePartitionTable.test.js
+++ b/ui/src/components/NodePartitionTable.test.js
@@ -4,14 +4,18 @@ import { screen } from '@testing-library/react';
 import { setupServer } from 'msw/node';
 import { rest } from 'msw';
 import NodePartitionTable from './NodePartitionTable';
-import { waitForLoadingToFinish, render } from './__TEST__/util';
+import {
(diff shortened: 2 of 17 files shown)


ui/src/containers/AlertProvider.js
```diff
@@ -0,0 +1,32 @@
+import React, { createContext, useContext } from 'react';
```
JBWatenbergScality, line 2: It should be typed using flow

ui/src/containers/AlertProvider.js
```diff
@@ -0,0 +1,32 @@
...
+import { useQuery } from 'react-query';
+import { getAlerts } from '../services/alertmanager/api';
+import { filterAlerts } from '../services/alertUtils';
+
+const AlertContext = createContext(null);
+
+export function useAlerts(filters: FilterLabels) {
+  const query = useContext(AlertContext);
+  if (query.status === 'success') {
```
JBWatenbergScality: I think a null check on query is missing here and should throw an exception to indicate that this hook can only be used within an `AlertProvider`

ui/src/containers/AlertProvider.js
```diff
@@ -0,0 +1,32 @@
...
+    return {
+      ...query,
+      data: filterAlerts(query.data, filters),
```
JBWatenbergScality:
```suggestion
      alerts: filterAlerts(query.data, filters),
```

ui/src/containers/AlertProvider.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+  return query;
+}
+
+const AlertProvider = ({ children }) => {
+  const query = useQuery('activeAlerts', () =>
+    getAlerts().then((data) => {
+      return data;
+    }),
+  );
```
JBWatenbergScality: It should keep the data updated by refetching the alerts frequently (https://react-query.tanstack.com/examples/auto-refetching).

```suggestion
  const query = useQuery('activeAlerts', () => getAlerts(), {
      // Refetch the alerts every 10 second
      refetchInterval: 10000,
    });
```

ui/src/containers/AlertHistoryProvider.js
```diff
@@ -0,0 +1,32 @@
+import React, { createContext, useContext } from 'react';
```
JBWatenbergScality, line 2: It should be typed using flow

ui/src/containers/AlertHistoryProvider.js
```diff
@@ -0,0 +1,32 @@
...
+import { useQuery } from 'react-query';
+import { getLast7DaysAlerts } from '../services/loki/api';
+import { filterAlerts } from '../services/alertUtils';
+
+const AlertHistoryContext = createContext(null);
+
+export function useHistoryAlerts(filters?: FilterLabels) {
+  const query = useContext(AlertHistoryContext);
+  if (query.status === 'success') {
```
JBWatenbergScality: I think a null check on query is missing here and should throw an exception to indicate that this hook can only be used within an AlertHistoryProvider

ui/src/containers/AlertHistoryProvider.js
```diff
@@ -0,0 +1,32 @@
...
+    return {
+      ...query,
+      data: filterAlerts(query.data, filters),
```
JBWatenbergScality:
```suggestion
      alerts: filterAlerts(query.data, filters),
```

ui/src/containers/AlertHistoryProvider.js
```diff
@@ -0,0 +1,32 @@
...
+    };
+  }
+  return query;
+}
+
+const AlertHistoryProvider = ({ children }) => {
+  const query = useQuery('alertsHistory', () =>
+    getLast7DaysAlerts().then((data) => {
+      return data;
+    }),
```
JBWatenbergScality:
```suggestion
    getLast7DaysAlerts(),
```

ui/src/components/NodePartitionTable.js
```diff
@@ -98,15 +104,23 @@ const LoaderContainer = styled.div`
...
 const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
   const theme = useTheme();
+  const { data: alertNF } = useAlerts({
+    alertname: [
+      NODE_FILESYSTEM_SPACE_FILLINGUP,
+      NODE_FILESYSTEM_ALMOST_OUTOF_SPACE,
+      NODE_FILESYSTEM_FILES_FILLINGUP,
+      NODE_FILESYSTEM_ALMOST_OUTOF_FILES,
+    ],
+  });
```
JBWatenbergScality: I think a filter to retrieve the specific alerts of this instance might be missing

JBWatenbergScality: I'm not sure but I think we implemented alert filter so that it checks filters with an AND condition so maybe using alertname here with a list of possible values will never retrieved the expected value. I would opt for filtering all the alerts linked to the selected instance and then let `computePartitionHealth` do the filtering as it is already the case.

ui/src/components/NodePartitionTable.js
```diff
@@ -116,11 +130,11 @@ const NodePartitionTable = ({ instanceIP }: { instanceIP: string }) => {
             return getNodePartitionsTableData(
               nodeFSUsageResult.data.result,
               nodeFSSizeResult.data.result,
-              alerts,
+              alertNF,
             );
           }
         }),
-      [instanceIP],
+      [alertNF, instanceIP],
```
JBWatenbergScality: alertNF is a complex object so using it as a dependency of useCallback could make this callback executed at every render of the NodePartitionTable because react is not performing a deep comparison on the provided dependencies. Maybe you can `JSON.stringify(alertNF)`

ui/src/containers/NodePageRSP.js
```diff
@@ -121,13 +122,10 @@ const NodePageRSP = (props) => {
...
   // Filter alerts for the specific node, base on the InstaceIP and Alert Name
   const alerts = useSelector((state) => state.app.alerts.list);
-  const alertsNode =
-    alerts?.filter(
-      (alert) =>
-        NODE_ALERTS_GROUP.includes(alert?.labels?.alertname) &&
-        `${instanceIP}:${PORT_NODE_EXPORTER}` === alert?.labels?.instance,
-    ) ?? [];
-
+  const alertsNode = filterAlerts(alerts, {
```
JBWatenbergScality: I think alerts should be dropped from redux store to avoid duplicating the source of alerts for the application. So instead of calling filterAlerts here we should use useAlerts hook

ui/src/containers/VolumePageContent.js
```diff
@@ -102,10 +103,7 @@ const VolumePageContent = (props) => {
 
   // get the alert
   const alertlist =
-    PVCName &&
-    alerts?.list?.filter(
-      (alert) => alert.labels.persistentvolumeclaim === PVCName,
-    );
+    PVCName && filterAlerts(alerts.list, { persistentvolumeclaim: PVCName });
```
JBWatenbergScality: Similarly here we can useAlerts instead of retrieving the alerts from redux

JBWatenbergScality (changes requested):

ui/src/components/NodePartitionTable.js
ChengYanJin: Actually, the util function we implemented is OR conditions:
https://github.com/scality/metalk8s/blob/cd742b579be666ef6bf6f0df5f3cc40eac9cd070/ui/src/services/alertUtils.js#L136

ui/src/components/NodePartitionTable.js
ChengYanJin: Thanks for pointing me to this performance issue. 
Just for me to understand. In our case, the NodePartitionTable should re-render when `instanceIP` changes which should retrigger the useCallback re-run.

ui/src/components/NodePartitionTable.test.js
```diff
@@ -52,7 +53,12 @@ const server = setupServer(
               'https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-nodefilesystemalmostoutofspace',
             summary: 'Filesystem has less than 5% space left.',
           },
-          endsAt: '2021-01-29T07:40:05.358Z',
+          /*
+          We want to have an active alert, meaning the current time should be in between `startsAt` and `endsAt`.
+          Since we can't spyon the `activeOn` in getHealthStatus(). If we use ```endsAt: new Date().toISOString()```, there will be slightly bwtween the two current time.
```
JBWatenbergScality:
```suggestion
          Since we can't spy on the `activeOn` in getHealthStatus(). If we use ```endsAt: new Date().toISOString()```, there will be a slightly difference between the two current time.
```

ui/src/components/NodePartitionTable.test.js
```diff
@@ -52,7 +53,12 @@ const server = setupServer(
...
+          Hence, here we add one day to make sure the alert is active.
+          */
+          endsAt: new Date(new Date().getTime() + 86400000).toISOString(),
```
JBWatenbergScality:
```suggestion
          endsAt: new Date(new Date().getTime() + (1000 * 60 * 60 * 24)).toISOString(),
```

ui/src/components/NodePartitionTable.test.js
```diff
@@ -100,6 +108,7 @@ describe('the system partition table', () => {
     jest.useFakeTimers();
     initializeProm('http://192.168.1.18:8443/api/prometheus');
     initializeAM('http://192.168.1.18:8443/api/alertmanager');
+    initializeLoki('http://192.168.1.18:8443/api/loki');
```
JBWatenbergScality: Maybe we can extract this IP `192.168.1.18 ` to a constant and use it afterwards as `${FAKE_CONTROL_PLANE_IP}` in string templates. Also we can change the actual value to `fake.control.plane.ip.invalid` to make it clear it is a fake ip

ui/src/containers/NodePageRSP.js
```diff
@@ -35,7 +36,8 @@ import {
   NODE_ALERTS_GROUP,
   PORT_NODE_EXPORTER,
 } from '../constants';
-import { filterAlerts } from '../services/alertUtils';
+import { useAlerts } from '../containers/AlertProvider';
+//import { filterAlerts } from '../services/alertUtils';
```
JBWatenbergScality: can be removed

JBWatenbergScality (approved):

PR #3171, opened 2021-03-05
salt/tests: Add unit testing for Salt Jinja templates

### LINKED ISSUE #3073: Unit Test for templated SLS rendering
**Component**:

'salt'

**Why this is needed**:

When writing YAML templates (using Jinja) we don't have ways to make sure it will successfully render at runtime.
As such we faced some issues where the template was referencing undefined variables and we observed the issue quite late in the process.

**What should be done**:

We should add, as part of our pre-merge CI, some tests to render the template to make sure it is working fine.
The idea is not to compare the rendered YAML file with some YAML results but just to make sure the system can render it.

**Test plan**:

gdemonet:
**Component**: salt, tests

**Context**: See #3073

**Summary**:

This PR introduces unit tests for all Jinja templates under `salt/`.
The goal is to automatically ensure that rendering logic is valid and will not break at runtime.

At this stage, no coverage information is provided. Furthermore, only rendering is verified, and the output is simply discarded - a future PR may add optional validity checks for the output (e.g. Salt states as YAML).

**Acceptance criteria**:

- All templates are tested as part of `tox -e unit-tests`, and in CI

---

Closes: #3073

---

<details>
  <summary>Task list</summary>

- [x] Document unit testing strategy
- [x] Implement basic logic for automatically testing all templates
- [x] Implement logic for configuring rendering contexts
- [x] Progressively enable tests for all templates:
    - [x] metalk8s.addons
    - [x] metalk8s.archives
    - [x] metalk8s.beacon
    - [x] metalk8s.container-engine
    - [x] metalk8s.internal
    - [x] metalk8s.kubectl
    - [x] metalk8s.kubernetes
    - [x] metalk8s.node
    - [x] metalk8s.orchestrate
    - [x] metalk8s.reactor
    - [x] metalk8s.repo
    - [x] metalk8s.roles
    - [x] metalk8s.salt
    - [x] metalk8s.service-configuration
    - [x] metalk8s.solutions
    - [x] metalk8s.sreport
    - [x] metalk8s.utils
    - [x] metalk8s.volumes
</details>


salt/tests/unit/formulas/config.yaml
```diff
@@ -0,0 +1,582 @@
...
+  minion_state:
+    - ready
+  volumes:
+    - none
+  mode:
+    - minion
+
+metalk8s:
+  # Use the special `_skip` keyword to omit rendering of a directory or formula
+  _skip: true
```
alexandre-allard: I would rather have `_skip` set to `false` at the top level (or even as a default) and explicitly skip the things we don't want to/can't test, but maybe it's already what is planned when everything will be ready.

gdemonet: `_skip` is already set to `false` by default, but since I wanted to build this PR by progressively adding more formulas to the scope, one of the first commits "skips everything except `map.jinja`" and the following ones are iteratively "unskipping" parts of the `metalk8s/` tree.

The goal is to get rid of the `_skip` statements after this PR, so nothing is skipped.

alexandre-allard (approved):
Really cool stuff you did there.
I tried some common mistakes in Jinja and it works well.
LGTM.

PR #3182, opened 2021-03-09
salt: Do not consider all ISOs in `/srv/scality` as Solutions

### LINKED ISSUE #3128: MetalK8s considers all ISO mounts under `/srv/scality` to be 'solutions' and fails otherwise
Any mounted `iso9660` FS under `/srv/scality` at a path that doesn't start with `metalk8s-` is considered to be a 'solution', and if it's not (no `manifest.yaml` etc), related code bails out.

Of course other things could be mounted at this path, so no real reason to reject these I guess?

See https://github.com/scality/metalk8s/blob/5b3ca3a9b63f20b0a8a4fecbcd46ba5c322c860c/salt/_modules/metalk8s_solutions.py#L159

TeddyAndrieux:
**Component**:

'salt', 'solutions'

**Context**: 

#3128 

**Summary**:

Add more check about MetalK8s solutions ISOs, the ISOs mounted in
`/srv/scality` also need to have a `manifests.yaml` with some expected
keys like a good `kind` and `apiVersion`, otherwise the mounted ISO is
ignored and consider as a non-metalk8s mountpoint

---

Fixes: #3128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,10 @@
   Salt dependencies from "base" RHEL 7 repository
   (PR [#3083](https://github.com/scality/metalk8s/pull/3083))
 
+- [#3128](https://github.com/scality/metalk8s/issues/3128) - No longer assume ISOs
+  mounted under `/srv/scality` are Solutions
+  (PR [#3182](https://github.com/scality/metalk8s/pull/3182))
+
 ### Deprecations and Removals
 
 - [#3168](https://github.com/scality/metalk8s/issues/3168) - [UI] Remove the environment page
--- a/salt/_modules/metalk8s_solutions.py
+++ b/salt/_modules/metalk8s_solutions.py
@@ -156,26 +156,6 @@ def deactivate_solution(solution):
     return True
 
 
-def _is_solution_mount(mount_tuple):
-    """Return whether a mount is for a Solution archive.
-
-    Any ISO9660 mounted in `/srv/scality` that isn't for MetalK8s is considered
-    to be a Solution archive.
-    """
-    mountpoint, mount_info = mount_tuple
-
-    if not mountpoint.startswith("/srv/scality/"):
-        return False
-
-    if mountpoint.startswith("/srv/scality/metalk8s-"):
-        return False
-
-    if mount_info["fstype"] != "iso9660":
-        return False
-
-    return True
-
-
 SOLUTION_MANIFEST = "manifest.yaml"
 SOLUTION_MANIFEST_KIND = "Solution"
 SOLUTION_MANIFEST_APIVERSIONS = [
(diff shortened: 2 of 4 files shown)


salt/_modules/metalk8s_solutions.py
```diff
@@ -159,8 +159,8 @@ def deactivate_solution(solution):
 def _is_solution_mount(mount_tuple):
     """Return whether a mount is for a Solution archive.
 
-    Any ISO9660 mounted in `/srv/scality` that isn't for MetalK8s is considered
-    to be a Solution archive.
+    Any ISO9660 mounted in `/srv/scality` that isn't for MetalK8s and has a
+    `manifests.yaml` with expected style, is considered to be a Solution archive.
```
gdemonet: (nit)
```suggestion
    `manifest.yaml` in the expected format, is considered to be a Solution archive.
```

CHANGELOG.md
```diff
@@ -58,6 +58,10 @@
   Salt dependencies from "base" RHEL 7 repository
   (PR [#3083](https://github.com/scality/metalk8s/pull/3083))
 
+- [#3128](https://github.com/scality/metalk8s/issues/3128) - Do not consider all ISOs
+  mounted in `/srv/scality` as MetalK8s mountpoint
```
```suggestion
- [#3128](https://github.com/scality/metalk8s/issues/3128) - No longer assume ISOs
  mounted under `/srv/scality` are Solutions (now validating the presence and format
  of a `manifest.yaml` file)
```

(not sure if we want to specify the part about the manifest though, feel free to omit it)

salt/_modules/metalk8s_solutions.py
```diff
@@ -173,6 +173,16 @@ def _is_solution_mount(mount_tuple):
     if mount_info["fstype"] != "iso9660":
         return False
 
+    try:
+        read_solution_manifest(mountpoint)
+    except CommandExecutionError as exc:
```
gdemonet: Not sure we want to do this `try`/`except` in `_is_solution_mount`. Since we only use this method in `list_available`:

```python
solution_mounts = filter(_is_solution_mount, active_mounts.items())

for mountpoint, mount_info in solution_mounts.items():
    manifest, info = read_solution_manifest(mountpoint)
```

Maybe we could move this try/except to `list_available`, and keep the resulting manifest around if there is one. This will avoid calling `read_solution_manifest` twice.


```python
candidate_mounts = filter(_is_candidate_mount, active_mounts.items())

for mountpoint, mount_info in candidate_mounts.items():
	try:
	    manifest, info = read_solution_manifest(mountpoint)
	    log.info("Skipping ...")
	    continue

    # the rest is unchanged
```

gdemonet (changes requested):

CHANGELOG.md
TeddyAndrieux: Indeed to me we do not really care about "manifest.yaml" here but I will update the sentence

salt/_modules/metalk8s_solutions.py
TeddyAndrieux: Then the `_is_solution_mount` should be renamed.

To me, with this approach, we can easily validate that a mountpoint is a solution just by calling this `_is_solution_mount`

Your approach works well also just I don't really see the benefits of `_is_solution_mount` (or `_is_candidate_mount`) to be efficient we could simply do one loop that filter out and extract everything at the same time.

Basically remove the `filter` and the `_is_solution_mount` and do everything in the for loop in `list_available`

gdemonet: Yeah we could do this as well, I'm fine with getting rid of the `filter` call. It's just there to make it clear what objects we want to iterate on, and reduce the clutter inside the `for` loop.

TeddyAndrieux: Done

tox.ini
```diff
@@ -79,6 +79,7 @@ omit =
   salt/_modules/metalk8s_package_manager_apt.py
 
 [testenv:unit-tests]
+basepython = python3.6
```
gdemonet: I'm surprised about this, wasn't it included in #3163? [relevant commit](https://github.com/scality/metalk8s/pull/3163/commits/3c736a0df7a4294609f75d94b392a6ec3a775000)

gdemonet (dismissed):

tox.ini
TeddyAndrieux: Oh ! You right a bad `git commit -a` :smiley:

TeddyAndrieux: Anyway it will not be in the resulting commit in the queue so it's ok

gdemonet (approved):

Issue #2667, opened 2020-07-13
Put Prometheus & Alertmanager APIs behind an OIDC proxy


alexandre-allard:
**Component**:

**Why this is needed**:
For now, Prometheus and Alertmanager APIs are exposed without any authentication through an ingress.
It means that anyone having access to the control plane network can query these APIs.
Prometheus API is read-only (except when /admin routes are enabled) and Alertmanager one allow to put silence or push alerts in addition to the ability of reading configuration, alerts, ... there is nothing really critical, but we don't want unauthorized people or applications to access these data.

**What should be done**:
As suggested by @NicolasT in https://github.com/scality/metalk8s/issues/2636#issuecomment-647401938, we need to put an OIDC proxy in front of these 2 APIs, to only give access to granted users.

**Implementation proposal** (strongly recommended):

**Test plan**:

alexandre-allard:
And Loki API

PR #3191, opened 2021-03-10
Alert history

### LINKED ISSUE #3180: Alert history
**Component**: alerting, logging

**Why this is needed**:
In NextGen UI we are introducing the Global Health Component that shows real time entity Health but also intends to show entity Health over the last X days.
As for now, we do not have a way of persisting alerts, but this component needs to be able to retrieve past alerts no longer stored in Alertmanager.

See https://github.com/scality/ringx/blob/development/1.0/docs/developer/architecture/alert_history.rst for details

**What should be done**:
Implement a mechanism to persist all alerts in a database and then be able to query this database via an API to retrieve them.

We will use an Alertmanager webhook which will log every alert received to its standard output, this way fluent-bit will be able to retrieve them and forward them to Loki.
The alerts can then be retrieved querying Loki like any other log.

See https://github.com/scality/ringx/pull/44 for details

**Test plan**:
Add a test in post-install checking that we can retrieve alerts via Loki API (we should at least have the Watchdog alert).

alexandre-allard:
**Component**: images, build, salt

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #3180


docs/developer/architecture/alert-history.rst
```diff
@@ -0,0 +1,285 @@
...
+
+Loki also provides an API allowing us to expose and query these alerts using
+the LogQL language.
+
+Plus, since Loki is already part of the cluster, it saves us from having
+to install, manage and expose a new database.
+
+Using Loki, we also directly benefit from its retention and purge mechanisms,
+making the alerts history retention time automatically aligned with all
+other logs (14 days by default).
```
gdemonet, line 150: :trophy:

docs/developer/architecture/alert-history.rst
```diff
@@ -0,0 +1,285 @@
...
+
+   There is a drawback in using Loki, if at some point its volume is full
+   (because there is too much logs), we will not be able to store new alerts
+   anymore, especially since there is no size-based purge mechanism.
+
+   Another issue is, since we share the retention configuration with the other
+   logs, it is hard to ensure we will keep enough alert history.
+
+   As for now, there is no retention based on labels, streams, tenant or
+   whatever (on-going discussion `GH Loki #162`_).
```
gdemonet, line 162: We need to follow this closely, and maybe contribute. Having tenant-based retention could be really nice for separating different kinds of log streams (not only alerts, but also e.g. Salt events, audit logs, ...).

docs/developer/architecture/alert-history.rst
```diff
@@ -0,0 +1,285 @@
...
+containers.
+Timestamps of the logs must be extracted from the JSON ``timestamp`` key and
+only the JSON part of the log must be stored to make it easier to use by
+automatic tools.
+
+Expose Loki API
+~~~~~~~~~~~~~~~
+
+The Loki API must be reachable via the web UI, therefore it must be exposed
+through an ingress as it is already done for Prometheus or Alertmanager APIs.
```
gdemonet, line 250: We should also start thinking about hiding these behind an OIDC proxy or something, cc @thomasdanan

docs/developer/architecture/alert-history.rst
```diff
@@ -0,0 +1,285 @@
...
+~~~~~~~~~~~~~~~~~
+
+Alerts are already retrievable from the ``Logs`` dashboard, but it is not
+user friendly as the webhook pod name must be known by the user and
+metrics displayed are relative to the pod, not the alerts themselves.
+
+A dedicated Grafana dashboard with the alerts and metrics related to them will
+be added.
+
+.. note::
```
gdemonet:
```suggestion
.. todo::
```

images/metalk8s-alert-logger/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if err := json.NewDecoder(r.Body).Decode(&alerts); err != nil {
+		log.Printf("Unable to parse HTTP body: %s", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	for _, alert := range alerts.Alerts {
+		encoded_alert, _ := json.Marshal(alert)
+	        log.Println(string(encoded_alert))
```
gdemonet: Some tab was replaced with spaces apparently:

```suggestion
		log.Println(string(encoded_alert))
```

Hints that we should add this file to `doit lint:go` :wink:

salt/metalk8s/addons/alert-logger/deployed/deployment.sls
```diff
@@ -0,0 +1,53 @@
...
+kind: Deployment
+metadata:
+  name: metalk8s-alert-logger
+  namespace: metalk8s-monitoring
+  labels:
+    app: metalk8s-alert-logger
+    heritage: metalk8s
+    app.kubernetes.io/name: metalk8s-alert-logger
+    app.kubernetes.io/part-of: metalk8s
+    app.kubernetes.io/managed-by: salt
```
gdemonet, line 16: Those labels should also be added to the PodTemplate IMO (maybe not in the selector though)

salt/metalk8s/addons/alert-logger/deployed/deployment.sls
```diff
@@ -0,0 +1,53 @@
...
+spec:
```
gdemonet, line 17: I see no `spec.replicas`, so we will only ever get one of those loggers (except when the Deployment is doing some rolling update)? Would it make sense to make it configurable? IIRC, you said in the design that since AM will retry sending events to the webhook, I guess it's fine to temporarily lose an instance of this Pod, I just wanted to confirm.

salt/metalk8s/addons/alert-logger/deployed/deployment.sls
```diff
@@ -0,0 +1,53 @@
...
+  selector:
+    matchLabels:
+      app: metalk8s-alert-logger
```
gdemonet: This `app` label is "deprecated", we should rely on the `app.kubernetes.io/name` equivalent instead. Same in the Service selector.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
```diff
@@ -109,7 +113,9 @@ Create fluent-bit ConfigMap:
             }
           parsers.conf: |-
             [PARSER]
-                Name        docker
-                Format      json
+                Name        container
+                Format      regex
+                Regex       ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]+) (?<message>.+)$
```
gdemonet, line 118: I think this parser should always be used for `containerd`-formatted logs, no? And as such, likely this commit should be backported to the first version where Loki was introduced (2.6, I guess?)

tests/post/steps/test_logging.py
```diff
@@ -194,6 +194,28 @@ def _check_log_line_exists():
...
+        "query": '{app="metalk8s-alert-logger"}',
+    }
+
+    def _check_log_line_exists():
+        response = query_loki_api(k8s_client, query, route="query_range")
+        try:
+            result_data = response[0]["data"]["result"][0]["values"]
+        except (IndexError, KeyError):
+            result_data = []
+        assert result_data, "No cluster alerts found in Loki"
```
gdemonet: Don't we want to add an extra check to verify we can see the Watchdog alert there, as suggested in the design? This way, we can also ensure the format of such logs is stable over time.

gdemonet (changes requested):
Mostly peripheral comments, some things should change in this PR though:

- add the logger Go code to the linter
- adjust the K8s manifests to use standard labels
- adjust the test to check alert format (using the Watchdog "example")

docs/developer/architecture/alert-history.rst
alexandre-allard: Agree, there is already an issue opened for this (#2667) concerning Prometheus and Alertmanager and I've added a comment to not forget about Loki.

salt/metalk8s/addons/alert-logger/deployed/deployment.sls
alexandre-allard: Yes, there is only one instance.
It was done on purpose, I don't see the point of having multiple instances of this Pod (I may be wrong).
We can still make it configurable if we want, it should not be a problem.

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: Yes, I agree, we should probably backport it.
Will do it once this PR is merged.

tests/post/steps/test_logging.py
alexandre-allard: Oh, yes we could, I didn't make it specific to Watchdog just in case it would change at some point, to avoid having to change the test and because the purpose was only to ensure the alert history is working fine.
But, you're right, this way we could also ensure we log valid JSON objects, with specific fields, etc (it would avoid silently breaking the UI in case the object sent by Alertmanager changed due to an upgrade of it).
I will change it.

salt/metalk8s/addons/alert-logger/deployed/deployment.sls
alexandre-allard: Totally, I forgot to do it. :)

images/metalk8s-alert-logger/main.go
```diff
@@ -0,0 +1,45 @@
...
+	var alerts template.Data
+
+	if err := json.NewDecoder(r.Body).Decode(&alerts); err != nil {
+		log.Printf("Unable to parse HTTP body: %s", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	for _, alert := range alerts.Alerts {
+		encoded_alert, _ := json.Marshal(alert)
```
gdemonet: Maybe add some error handling here?

images/metalk8s-alert-logger/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.16.0-alpine AS builder
```
gdemonet: Go and Alpine versions should likely be exposed as build arguments, don't you think?

gdemonet (dismissed):
Some relatively minor comments, feel free to put them in a debt ticket.

images/metalk8s-alert-logger/Dockerfile
alexandre-allard: Hmm, yes, why not, I'm not sure if it will useful, but it does not cost anything to do it.
I'll change it.

images/metalk8s-alert-logger/main.go
alexandre-allard: Done

images/metalk8s-alert-logger/Dockerfile
alexandre-allard: Done

gdemonet (dismissed):

gdemonet (approved):
Don't forget about https://github.com/scality/metalk8s/pull/3191/files#r601234682 :wink:

salt/metalk8s/addons/logging/fluent-bit/deployed/configmap.sls
alexandre-allard: PR opened here #3276

PR #3194, opened 2021-03-11
UI : fix design discrepancies

### LINKED ISSUE #3121: Discrepancies between design and actual implementation in MetalK8s UI
**Component**: ui

**Why this is needed**:
Thanks to the review done by @Cuervino, we have a list of discrepancies to address.
https://miro.com/app/board/o9J_lWWXfu8=/?moveToWidget=3074457354525320056&cot=14

**What should be done**:
The discrepancies appear to the following pages:

### NodeVolumeTable: 
image
1. Volume's names to change to blue.
2. The link to the volume page should be on the name, not on the full line.
3. The last item is not fully visible, even when totally scrolling down. 
   Yanjin: "I believe it's Because of the height of the table"
4. When hovering, the tooltip should be displayed on top of the header, not under it. 
  Yanjin: "We have this issue for all the tables. So please check all the tables"
5. Add tooltip with "Advanced Monitoring" when hover on icons on `Logs` column.
image
6. When hovering on the header and being able to sort, the cursor should be the "action" one, not the "edit"
image

### NodeListTable:
7. Search bar styling: Add mangnifiying icon and text style to be changed
8. Change icon of Create Node to +
9. Avoid the style change for the selected node.

image
image

### NodeCreation:
https://www.sketch.com/s/f7ae733f-222f-4335-b742-a86d25dda872/a/zDGz9e

10. Field width to expand
11. Title to add 
12. The creation button is not visible
13. Update the checkbox with the new design
image
14. Add "*" when fields are mandatory (+sentence "All * are mandatory fields)
image

### Charts: 
~~15.  During the loading, we should display the layout of charts.~~
image
Already done.

16. When switching on the Node Cluster Average, display a small loader next to the switch.

(Maybe in a separate PR)
VolumeOverviewPage:
17. Timeformat to be aligned with XDM UI.
image
18. Scrollbar to style

To case the review, I would suggest is one commit to address one issue.

**Test plan**:
TBD

alexis-ld:
**Component**:
ui, nodes, volumes

**Context**:  We want to align the UI better to the design following specs in #3121 .

**Summary**:

* Fixed table tooltips top row display issue. 
* Small adjustments all over the nodes/volumes pages (cursor changes, icons, ...)
* Reworked the create node and create volume pages so they fit the design better (global alignment, smaller checkboxes, toggle instead of checkbox, display required fields, ...) (#3121 requested wider inputs in node creation page, it is done, however this has not been replicated on volume creation page as the layout is a bit different and sometime has two input next to each other..)
Screenshot 2021-03-10 at 19 02 46

* Following a discussions with @ChengYanJin I split the the VolumeListTable component in two, the layout / behavior was getting two different between the Volume page and the Node page use of this component. The new created component (NodePageVolumesTable) is much lighter and readable because I removed the need for conditional Search and Node column display. It also allowed to fix some display issue related to table size more easily.
* Added a loader next to the Node avg metrics toggle

As suggested in #3121  points 17 & 18 will come in another PR.

Screenshots:
NodeVolumeTab:
NodeVolumeTable

CreateVolume:
VolumeCreation

VolumePage:
Volumelist

---

Closes: #3121


Cuervino:
1. Nodes page, pods tab > the logs column is not aligned.
Moreover, the header of "Status" and "Namespace" columns are not aligned (left) with the content.
image

2. In the forms, the sentence "All * are mandatory fields" is in bold, it should be in regular font.

3. For the form "create new node", I still see issue 12 (The creation button is not visible), and there is a ghost slider (on Firefox).
image

4. Node page, volumes tab > the name are in blue, but should not be underlined (for instance, it's not the case in the Volumes page, overview tab)
5. Table on Volumes tab > too much aligned on the right (the sorting icon that is displayed on hover is outside of the table.
I would leave at least 8px before the end of the area.
image

6. Remark: The tooltip on hovering when content overflow could be displayed on a single line (especially that the content here is not very long).
The tooltip is displayed even when there is no truncation, I don't see the added value.
image

ui/src/components/TableRow.js
```diff
@@ -36,12 +35,11 @@ export const UnknownIcon = styled.i`
 `;
 
 const TableRow = (props) => {
-  const { row, style, onClickRow, isSelected } = props;
+  const { row, style, isSelected } = props;
 
   return (
     <TableRowStyle
       {...row.getRowProps({
-        onClick: () => onClickRow(row),
```
ChengYanJin: Would be cool to remove the highlight of the row, since we won't be able to select the row anymore. 
@Cuervino Can you confirm that please?

image

ui/src/components/VolumeListTable.js
```diff
@@ -96,6 +96,10 @@ const ActionContainer = styled.span`
   padding: ${padding.large} ${padding.base} 26px 20px;
 `;
 
+const NameLinkContaner = styled.div`
+  cursor: pointer;
```
ChengYanJin: 1. We should remove the underline of the volume name.
2. We should add a margin between the volume name and usage progressbar.

ui/src/components/NodePagePodsTab.js
```diff
@@ -184,13 +188,18 @@ const NodePagePodsTab = (props) => {
         cellStyle: { textAlign: 'center', width: '40px' },
         Cell: ({ value }) => {
           return (
-            <ExternalLink
-              href={`${config.api.url_grafana}/dashboard/db/logs?orgId=1&var-logs=Loki&var-logmetrics=Prometheus&var-metrics=Prometheus&var-podlogs=.*&var-systemlogs=.%2B&var-deployment=calico-kube-controllers&var-pod=${value}`}
-              target="_blank"
-              rel="noopener noreferrer"
+            <Tooltip
+              placement={'left'}
+              overlay={<TooltipContent>Advanced Monitoring</TooltipContent>}
```
ChengYanJin: would be great to add the translations

ui/src/components/VolumeListTable.js
```diff
@@ -126,28 +132,14 @@ function GlobalFilter({
   }, 500);
 
   return (
-    <input
+    <SearchInput
```
ChengYanJin, line 131: the font family of the search input should be `Lato`

ui/src/components/style/FormStyle.js
```diff
@@ -0,0 +1,44 @@
...
+      margin: 0px;
+    }
+
+    span {
+      vertical-align: baseline;
+    }
+  }
+`;
+
+export const ToggleWrapper = styled.div`
```
ChengYanJin: Would make sense to create a PR in core-ui for the `CheckboxWrapper` and `ToggleWrapper`

ui/src/containers/CreateVolume.js
```diff
@@ -150,12 +157,6 @@ const LabelsName = styled(LabelsValue)`
   color: ${(props) => props.theme.brand.textPrimary};
 `;
 
-const PageContainer = styled.div`
-  display: flex;
-  // To avoid the big empty space on the right side of page
-  justify-content: space-around;
-`;
-
```
ChengYanJin, line 158: We should keep this styling to put this creation form in the middle of the page to avoid the big empty space on the right side of the page.

ui/src/containers/CreateVolume.js
```diff
@@ -514,19 +516,23 @@ const CreateVolume = (props) => {
 
               return (
                 <Form>
+                  <FormSection>
+                    <RequiredText>All * are mandatory fields</RequiredText>
```
ChengYanJin: the text might need to be slightly changed to, please add translation as well.
`* Required fields`

ChengYanJin (changes requested):
Thanks for changing the target to `development/2.9`
And address some visual changes.

In addition: 
I think we haven't addressed this issue:
NodeCreation:12. The creation button is not visible
image
We can't scroll down to see the button at this moment.

ui/src/components/TableRow.js
Cuervino: Yes, indeed, here selecting the line doesn't have any added value. Moreover, the table in the "Pods" tab already has this "non-selectable lines" logic.

ui/src/components/style/FormStyle.js
alexis-ld: @ChengYanJin the other components in core-ui have a "size" prop with the usual variants (smaller, small, base, large, larger) . I don't think this is relevant for the checkbox of the toggle, it is unlikely we need a very larger checkbox anywhere... ^^ do you have any preference on how to implement this then ? a "isSmall" prop or something like that ?

ui/src/containers/CreateVolume.js
alexis-ld: Oh ok. On the design it was aligned left : 
Screenshot 2021-03-16 at 15 37 35

Will move it back to the center.

ChengYanJin: @alexis-ld 
thank you! this should be the correct one!
image

alexis-ld:
@ChengYanJin @Cuervino I pushed the changes.

The table scrollbar issue seems a bit more complex than expected and I will push the fix in another PR so we can move on with this one.

alexis-ld:
Also the toggle/tooltip/checkbox fixes have been moved to core-ui directly. So for this to work properly we will need to merge https://github.com/scality/core-ui/pull/272 first and use this version in metal :)

ui/src/components/TableRow.js
```diff
@@ -1,16 +1,20 @@
...
 import { intl } from '../translations/IntlGlobalProvider';
 
 const TableRowStyle = styled.div`
-  &:hover,
-  &:focus {
-    background-color: ${(props) => props.theme.brand.backgroundBluer};
-    outline: none;
-    cursor: pointer;
-  }
+  ${(props) =>
```
ChengYanJin: I feel by doing this will create complexity when we will extract the table component.
Basically, we will extract at least 2 table components. One simple: like the one in the volume tab.  Another one with more functionalities, maybe row selection + search, etc. 

So here, I would purpose to have another <TableRow/> which be used in the simple use case. 
I know we will duplicate the code for the moment, but that's OK.

ChengYanJin (changes requested):
Thanks for taking care of the change request from me and Valamir!

I see three remain issues:
1. Would be great to leverage the empty space be to used by the Volume Name, whose field usually has longer text. 
image
2. the loader of when active cluster average should be next to the text.
image
3. the documentation tooltip is not visible.
image

alexis-ld:
@ChengYanJin Should be all good now :) 

Also please note that since we moved the Custom Scrollbar rules to its own component in core-ui, these changes won't work before https://github.com/scality/core-ui/pull/274 is merged and this version of core-ui is used within Metal :)

ChengYanJin (approved):
LGTM!

Issue #2878, opened 2020-10-19
[UI] Add integration tests using Cypress


gdemonet:
**Component**: ui, tests

**Why this is needed**: We're missing a lot of coverage for UI behaviour, making new additions and refactoring harder over time.

**What should be done**:

- [x] Introduce Cypress integration tests (no deployment required, only mocks)
- [x] Test navigation scenarios (sidebar, selecting table entries, switching between tabs...)
- [ ] Test Node deployment scenarios (successful deployment, failed deployment, no Job found)
- [ ] Test Alerts scenarios (impact on table cells, Overview tab, Alerts tab)
- [ ] Test Creation scenarios (create Volume from Volume page, from Node page, test various error cases, ...)

Maybe some other to be added

**Implementation proposal** (strongly recommended):
Add a step to run Cypress tests in a UI-dedicated test stage (maybe under `unit-tests`?), and when applicable, add an E2E variant to these tests to run under `multiple-nodes-centos`.

**Test plan**: N/A

PR #3195, opened 2021-03-11
Fix node condition display bug in Metalk8s UI Node Page 

ChengYanJin:
**Component**: ui, nodes

**Context**: 
When the node is in the condition  `'DiskPressure' | 'MemoryPressure' | 'PIDPressure'| 'NetworkUnavailable' | 'Unschedulable'` there is a JS error that occurs. 

**Summary**:
The buggy code is `nodes.conditions.map((cond) => {}` which should be `node.conditions.map((cond) => {}`
so clearly a typo that could have been detected by flow-typing. 

So apart from fixing the bug, I also enabled the flow-typing in the `NodeUtils.js file`.

However, Flow doesn't really support explicitly specifying the parameters in a function call

> for more information: https://github.com/flow-typed/flow-typed/issues/2445#issuecomment-444278046
so we need to  override the typing of createSelector and create a generic one

```js
export const createTypedSelector: <T>(
  selector: (state: RootState) => any,
  selector: (state: RootState) => any,
  selector: (state: RootState) => any,
  selector: (state: RootState) => any,
  selectorsResult: (...result: any) => T,
) => T = createSelector;
```

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/src/ducks/app/nodes.js
+++ b/ui/src/ducks/app/nodes.js
@@ -1,3 +1,4 @@
+//@flow
 import {
   all,
   call,
--- a/ui/src/ducks/reducer.js
+++ b/ui/src/ducks/reducer.js
@@ -4,7 +4,8 @@ import { reducer as oidcReducer, UserState } from 'redux-oidc';
 
 import config from './config';
 import type { ConfigState } from './config';
-import nodes, { NodesState } from './app/nodes';
+import nodes from './app/nodes';
+import type { NodesState } from './app/nodes';
 import pods from './app/pods';
 import type { PodsState } from './app/pods';
 import volumes from './app/volumes';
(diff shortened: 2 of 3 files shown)


JBWatenbergScality (approved):
I think it should also be enforced by adding tests that  covers branches where the nodes are not ready. Maybe we should open an issue to track this missing test coverage.

ChengYanJin:
> I think it should also be enforced by adding tests that covers branches where the nodes are not ready. Maybe we should open an issue to track this missing test coverage.

Hello @JBWatenbergScality 
Yes. It would make sense to add integration tests(To test when the node are not in `Ready` condition) to this ticket.
https://github.com/scality/metalk8s/issues/2878

PR #3198, opened 2021-03-12
tests/post: Reduce some flakiness

gdemonet:
**docs, tests: Apply specific '.deployed' states**

When changing the CSC ConfigMap of a component we were often running
the whole `metalk8s.deployed` state.
We now document which state to run for each specific component, and
ensure the tests are following this documentation.

---

**tests: Clean-up Dex-related tests**

Mainly remove the `request_retry_session` fixture and replace it with
direct calls to `utils.request_retry_session` when needed, so one can
adjust the expected error codes and the number of retries depending on
the use-case.
Adding `502` and `504` to the list of expected error codes, and bumping
retries from 3 to 5, for all requests specifically targeted at Dex.

---

**tests: Wait for rollout of CSC changes to complete**

Not waiting was leading to some flaky issues where the expected changes
were not yet propagated to the tested service, hence breaking the test.

### DIFF
--- a/docs/operation/account_administration/user_authentication_and_identity_management.rst
+++ b/docs/operation/account_administration/user_authentication_and_identity_management.rst
@@ -96,12 +96,26 @@ the following steps from the bootstrap node.
 
 #. Apply your changes.
 
-   .. parsed-literal::
+   #. Update Dex deployment.
 
-      root\@bootstrap $ kubectl exec -n kube-system -c salt-master \\
-                         --kubeconfig /etc/kubernetes/admin.conf \\
-                         salt-master-bootstrap -- salt-run state.sls \\
-                         metalk8s.addons.dex.deployed saltenv=metalk8s-|version|
+      .. parsed-literal::
+
+         root\@bootstrap $ kubectl exec -n kube-system -c salt-master \\
+                           --kubeconfig /etc/kubernetes/admin.conf \\
+                           salt-master-bootstrap -- salt-run state.sls \\
+                           metalk8s.addons.dex.deployed \\
+                           saltenv=metalk8s-|version|
+
+   #. Update Grafana configuration (so this new user is recognized as a
+      Grafana admin user).
+
+      .. parsed-literal::
+
+         root\@bootstrap $ kubectl exec -n kube-system -c salt-master \\
+                           --kubeconfig /etc/kubernetes/admin.conf \\
+                           salt-master-bootstrap -- salt-run state.sls \\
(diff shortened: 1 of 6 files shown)


alexandre-allard (dismissed):

JBWatenbergScality (dismissed):

gdemonet:
Actually, @alexandre-allard-scality has a better solution: 876ebcc

The proposed approach here encodes in the tests the state which should be executed, whereas the other approach makes it part of "the feature", and thus assumes a user would run it.
I will cherry-pick his commit and also update the documentation in the relevant places.

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -610,7 +610,7 @@ To change the number of replicas, perform the following operations:
 
       root\@bootstrap $ kubectl exec --kubeconfig /etc/kubernetes/admin.conf \\
                          -n kube-system -c salt-master salt-master-bootstrap \\
-                         -- salt-run state.sls metalk8s.deployed \\
+                         -- salt-run state.sls metalk8s.logging.deployed \\
```
alexandre-allard: The `Replicas Count Customization` section is not only about Loki, but for all services using the CSC mechanism.

alexandre-allard (dismissed):

docs/operation/cluster_and_service_configuration.rst
gdemonet: Ah right, went too fast. Good catch :ok_hand:

gdemonet: Done

gdemonet:
Still one flaky occurrence to fix on the service-configuration test:

```
=================================== FAILURES ===================================
________________ test_static_user_change[paramiko://bootstrap] _________________
[...]
        assert auth_response.text is not None
        if should_fail:
>           assert auth_response.status_code == 200
E           assert 303 == 200
E            +  where 303 = <Response [303]>.status_code
```

This is likely due to a race condition:

- the test applies a service-configuration change to edit the password of our static user in Dex
- the test then attempts to login using the old password, and expects this to fail
- however, it succeeds, indicating that the login attempt occurred before the changes were fully propagated to Dex

This race issue directly comes from Kubernetes declarative approach: our states only ensure that the declaration of desired state is updated, not that the observed state has been reconciled to match this desired state declaration.

I will work around this by adding a `kubectl rollout status deploy/dex -n metalk8s-auth`.

alexandre-allard (approved):

PR #3206, opened 2021-03-16
UI :  make UI deployable on other path than root

### LINKED ISSUE #3019: metalk8s ui routing should adapt itself to the path on which we serve it in ingress configuration
**Component**:

ui: routing

**What happened**:

When changing the path on which the ui is served in ingress configuration the router lost it's context and can't route the user to the expected page.

**What was expected**:

When the ui is not served at the root path `/`, the client routing should be aware of this new routing context and base it's path on the path of the SPA entrypoint.

**Steps to reproduce**

 1. Patch ingress configuration to serve the ui on a path different than `/`
 2. Try reach the ui at `http://<ip>:8443/<some_new_path>`
 3. You then will saw a blank page as this route is not anymore handled by the `react-router`

**Resolution proposal**:

There are several possibilities available to solve this issue : 

1. Set the basename attribute of history to the new path where the ui is being served :
        -  We should probably get this path form the configuration of the ui (cm/metalk8s-ui) or directly form the ingress configuration (ingress/metalk8s-ui)
        - It will then impact a little but the user experience as the user will saw a loading state as long as the configuration is not resolved when trying to access the ui

2. Switch react-router implementation to use a HashRouter instead of a browser router. HashRouter will then compute the routes starting from the # sign in the path and it should then enable us to serve the ui from any paths

JBWatenbergScality:
**Component**:

ui

**Context**: 

When changing the path on which the ui is served in ingress configuration the router lost it's context and can't route the user to the expected page.

**Summary**:

This PR solves this by initiating browser history object with a basename coming from UI configuration. 

**Acceptance criteria**: 

integration tests

---

Closes: #3019


ui/src/services/api.js
```diff
@@ -54,6 +54,7 @@ export type Config = {
   flags?: [],
   url_navbar: string,
   url_navbar_config: string,
+  ui_base_path: string,
```
ChengYanJin, line 57: should we add it in the `config.json` as well?

ChengYanJin: should we add this `ui_base_path` in `config.json` as well?

salt/metalk8s/addons/ui/deployed/ingress.sls
```diff
@@ -90,3 +90,19 @@ spec:
         backend:
           serviceName: metalk8s-ui
           servicePort: 80
+      - path: /config.json
```
ChengYanJin, line 93: IIUC, here we want to expose those files to the root, so that can be used by other application. 
It's more a question, do you think we may have filename conflicts?

ui/src/ducks/app/nodes.test.js
```diff
@@ -12,13 +11,33 @@ import {
...
+//       },
+//       history: {
+//         history: {
+//           push: jest.fn()
+//         }
+//       }
+//     })
+//   }
+// }));
+
```
ChengYanJin: To be removed? ;)

ui/src/services/api.js
JBWatenbergScality: I added this configuration in a previous PR : https://github.com/scality/metalk8s/blob/improvement/enable-users-group-mapping-in-shell-ui-config/salt/metalk8s/addons/ui/deployed/ui.sls.in#L83

JBWatenbergScality: I don't think it is necessary to be added in local config.json nor cypress one for now as if the value is null it will by default assume that it is deployed on the root path. I will make this explicit by switching this type to a nullable string through

salt/metalk8s/addons/ui/deployed/ingress.sls
JBWatenbergScality: I agree and I think ultimately we should move those files under the same path of the UI. However right now we have strong dependencies on those files as discussed offline. Also moving the config file under the same path of the UI requires to review how the configuration is injected in the UI. Indeed it is this configuration that provides the path on which the UI is deployed, so the UI should be able to get it's configuration no matter the path on which it is deployed (likely by injecting it in index.html when deploying the UI or via another way).

ChengYanJin (approved):
LGTM!

ui/src/index.js
```diff
@@ -24,11 +26,37 @@ if (window.Cypress) window.__store__ = store;
...
+    if (window.Cypress) window.__history__ = history;
+  }, [basename]);
+ 
+  if (configStatus === 'error') {
+    return <>An error occurred, please try to refresh the page</>//Todo display an error page
+  }
+
+  if (configStatus === 'idle' || configStatus === 'loading') {
+    return <>{children}</>
+  }
```
gdemonet, line 50: What is this behaviour meaning? All `<Route/>`s will be rendered at once?

JBWatenbergScality: Not exactly the `<App />` component is the one responsible for fetching the configuration so it needs to be rendered firstly. However we can't render the router until the configuration is not fetched because we not yet have the base path. So we render children here, to render App, which triggers config fetching when the status is idle and then when it is successfully fetched we wrap the App around the router. Also in App component we only render child routes once the status is success so with the router wrapper being defined

gdemonet (approved):

PR #3210, opened 2021-03-19
docs: Add details on salt-ssh annotations

### LINKED ISSUE #2440: [docs]: Missing Salt-SSH annotations details on Node manifests creation
**Component**: docs

**What happened**:

In the Installation Guide, when [creating Nodes using the CLI](https://metal-k8s.readthedocs.io/en/latest/installation/expansion.html#adding-a-node-from-the-command-line), we get this example manifest:

```yaml
apiVersion: v1
kind: Node
metadata:
  name: <node_name>
  annotations:
    metalk8s.scality.com/ssh-key-path: /etc/metalk8s/pki/salt-bootstrap
    metalk8s.scality.com/ssh-host: <node control plane IP>
    metalk8s.scality.com/ssh-sudo: 'false'
  labels:
    metalk8s.scality.com/version: '2.6.0-dev'
    <role labels>
spec:
  taints: <taints>
```

And no further explanation regarding the Salt-SSH annotations (hence, no mention of the `ssh-user` annotation, and its default value being `root`).

**What was expected**:

A user would probably need to adjust these values depending on its environment. The documentation should make the intent of each annotation clear, and give a simple example for providing a non-root user.

alexandre-allard:
**Component**: docs

**Context**:

**Summary**:
Add missing `ssh-user` annotation and add a quick description of each annotation and the default value used if not provided.

**Acceptance criteria**: 

---

Closes: #2440

### DIFF
--- a/docs/installation/expansion.rst
+++ b/docs/installation/expansion.rst
@@ -132,11 +132,32 @@ following the template below:
        metalk8s.scality.com/ssh-key-path: /etc/metalk8s/pki/salt-bootstrap
        metalk8s.scality.com/ssh-host: <node control plane IP>
        metalk8s.scality.com/ssh-sudo: 'false'
+       metalk8s.scality.com/ssh-user: 'root'
      labels:
        metalk8s.scality.com/version: '|version|'
        <role labels>
    spec:
      taints: <taints>
+
+Annotations are used by Salt-SSH to connect to the node and deploy it.
+All annotations are prefixed with ``metalk8s.scality.com/``:
+
++--------------+--------------------------------------------------+-----------+
+| Annotation   | Description                                      | Default   |
++==============+==================================================+===========+
+| ssh-host     | Control plane IP of the node, must be accessible | None      |
+|              | over SSH from the Bootstrap node                 |           |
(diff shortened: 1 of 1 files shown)


docs/installation/expansion.rst
```diff
@@ -132,12 +132,33 @@ following the template below:
...
    spec:
      taints: <taints>
 
+Annotations are used by Salt-SSH to connect to the node and deploy it.
+All annotations are prefixed with ``metalk8s.scality.com/``:
+
++--------------+--------------------------------------------------+-----------+
+| Annotation   | Description                                      | Default   |
++==============+==================================================+===========+
+| ssh-host     | Control plane IP of the node, must be accessible | None      |
```
gdemonet, line 148: I think "accessible" in itself doesn't mean much, would be good to specify "accessible from the Bootstrap node"

gdemonet (commented):
Mostly LGTM, thanks for taking care of this ticket :+1:

docs/installation/expansion.rst
alexandre-allard: changed

gdemonet (approved):

Issue #3211, opened 2021-03-19
Handle `fs.may_detach_mounts` sysctl for containerd


NicolasT:
On EL7.4, a new sysctl, `fs.may_detach_mounts`, was added which should be enabled on hosts where container runtimes are being used (it's off by default for backwards compatibility). Docker sets this as part of the daemon startup process, `containerd` (and `cri-o`) don't. We've seen the netns cleanup issue resulting from this not being set in the past in various deployments.

Adding a `sysctl.d` drop-in to set the value as part of the `containerd` RPM we build should fix this.

See:
- https://access.redhat.com/solutions/5430091
- https://github.com/containerd/containerd/issues/3667#issuecomment-802373736
- https://github.com/cri-o/cri-o/pull/4210

and various other sources around the internet.

PR #3212, opened 2021-03-19
packages: Handle fs.may_detach_mounts sysctl for containerd

NicolasT:
On EL7.4, a new sysctl, `fs.may_detach_mounts`, was added which should
be enabled on hosts where container runtimes are being used (it's off
by default for backwards compatibility). Docker sets this as part of
the daemon startup process, `containerd` (and `cri-o`) don't. We've
seen the netns cleanup issue resulting from this not being set in the
past in various deployments.

Adding a `sysctl.d` drop-in to set the value as part of the containerd
RPM we build should fix this.

Fixes: https://github.com/scality/metalk8s/issues/3211
See: https://access.redhat.com/solutions/5430091
See: https://github.com/containerd/containerd/issues/3667#issuecomment-802373736
See: https://github.com/cri-o/cri-o/pull/4210

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -311,6 +311,7 @@ CONTAINERD_RPM = _rpm_package(
         Path('0001-Revert-commit-for-Windows-metrics.patch'),
         Path('containerd.service'),
         Path('containerd.toml'),
+        Path('60-containerd.conf'),
         Path('containerd-{}.tar.gz'.format(versions.CONTAINERD_VERSION)),
     ],
 )
--- a/buildchain/buildchain/versions.py
+++ b/buildchain/buildchain/versions.py
@@ -21,7 +21,7 @@ CALICO_VERSION     : str = '3.17.0'
 K8S_VERSION        : str = '1.18.16'
 SALT_VERSION       : str = '3002.5'
 CONTAINERD_VERSION : str = '1.4.3'
-CONTAINERD_RELEASE : str = '1.el7'
+CONTAINERD_RELEASE : str = '2.el7'
 
 def load_version_information() -> None:
     """Load version information from `VERSION`."""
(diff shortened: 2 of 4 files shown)


NicolasT:
@bert-e status

NicolasT:
Note for reviewers: in the 2.8 integration branch, I bumped `containerd` to `1.4.3-3` with a small change to only include the sysctl drop-in on EL7 builds, not EL8.

gdemonet (approved):
> Note for reviewers: in the 2.8 integration branch, I bumped `containerd` to `1.4.3-3` with a small change to only include the sysctl drop-in on EL7 builds, not EL8.

I'm curious, why not define this `%if 0%{?el7}` condition in the `1.4.3-2` version of the package, given this kernel parameter only exists in EL7 (even if we don't support EL8 in MetalK8s 2.7)?

NicolasT:
> > Note for reviewers: in the 2.8 integration branch, I bumped `containerd` to `1.4.3-3` with a small change to only include the sysctl drop-in on EL7 builds, not EL8.
> 
> I'm curious, why not define this `%if 0%{?el7}` condition in the `1.4.3-2` version of the package, given this kernel parameter only exists in EL7 (even if we don't support EL8 in MetalK8s 2.7)?

Hmh, good point, could've worked. At the same time, the spec file we have in 2.7 is not compatible with el8, so having an el7 check in there could give wrong impressions.

NicolasT:
@bert-e status

NicolasT:
@bert-e status

gdemonet (approved):

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

NicolasT:
@bert-e status

PR #3213, opened 2021-03-19
salt: Add retry when trying to lookup grains

alexandre-allard:
**Component**: salt

**Context**: 
Bootstrap orchestration sometimes fails when trying to retrieve `metalk8s:control_plane_ip` and `metalk8s:workload_plane_ip` grains due to the CI being very slow.

**Summary**:
Add a retry mechanism when getting the grains.
Add also a test state to bail out early with an explicit error message when we don't manage to retrieve grains after all retries.
Remove `metalk8s:workload_plane_ip` grain retrieval as we do not seem to use it anywhere.

**Acceptance criteria**:
This flaky does not occur anymore.

### DIFF
--- a/salt/metalk8s/orchestrate/bootstrap/init.sls
+++ b/salt/metalk8s/orchestrate/bootstrap/init.sls
@@ -1,23 +1,33 @@
 {# Because of the grain lookup below, the bootstrap minion *must* be available
    before invoking this SLS, otherwise rendering will fail #}
-{%- set bootstrap_control_plane_ip = salt.saltutil.cmd(
-        tgt=pillar.bootstrap_id,
-        fun='grains.get',
-        kwarg={
-            'key': 'metalk8s:control_plane_ip',
-        },
-    )[pillar.bootstrap_id]['ret']
-%}
 
-{%- set bootstrap_workload_plane_ip = salt.saltutil.cmd(
-        tgt=pillar.bootstrap_id,
-        fun='grains.get',
-        kwarg={
-            'key': 'metalk8s:workload_plane_ip',
-        },
-    )[pillar.bootstrap_id]['ret']
-%}
-
+{%- set max_try = 5 %}
+{%- set bootstrap_grains = {} %}
+{%- for _ in range(max_try) %}
+  {%- set res = salt.saltutil.cmd(
+          tgt=pillar.bootstrap_id,
+          fun='grains.get',
+          kwarg={
+              'key': 'metalk8s:control_plane_ip',
+          },
+      )
+  %}
+  {%- set control_plane_ip = res.get(pillar.bootstrap_id, {}).get('ret') %}
+  {%- if control_plane_ip %}
+    {%- do bootstrap_grains.update({'control_plane_ip': control_plane_ip}) %}
+    {%- break %}
+  {%- endif %}
+{%- endfor %}
+{%- if 'control_plane_ip' not in bootstrap_grains %}
+Get metalk8s:control_plane_ip grain:
+  test.fail_without_changes:
+    - comment: >-
+        Unable to get metalk8s:control_plane_ip grain from {{ pillar.bootstrap_id }}
+        after {{ max_try }} tries.
+    - require_in:
+      - salt: Deploy CA role on bootstrap minion
+    - failhard: True
+{%- endif %}
 
 {%- if 'metalk8s' in pillar
         and 'nodes' in pillar.metalk8s
@@ -42,13 +52,13 @@
             'volumes': None,
             'endpoints': {
                 'salt-master': {
-                    'ip': bootstrap_control_plane_ip,
+                    'ip': bootstrap_grains['control_plane_ip'],
                     'ports': {
                         'api': 4507,
                     },
                 },
                 'repositories': {
-                    'ip': bootstrap_control_plane_ip,
+                    'ip': bootstrap_grains['control_plane_ip'],
                     'ports': {
                         'http': 8080,
                     },


salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,33 @@
...
-        fun='grains.get',
-        kwarg={
-            'key': 'metalk8s:workload_plane_ip',
-        },
-    )[pillar.bootstrap_id]['ret']
-%}
 
+{%- set max_try = 5 %}
+{%- set bootstrap_grains = {} %}
+{%- for _ in range(1, max_try) %}
```
gdemonet:
```suggestion
{%- for _ in range(max_try) %}
```

otherwise your loop will only run `max_try - 1` times

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,33 @@
...
+              'key': 'metalk8s:control_plane_ip',
+          },
+      )
+  %}
+  {%- if pillar.bootstrap_id in res and 'ret' in res[pillar.bootstrap_id] %}
+    {%- do bootstrap_grains.update(
+        {'control_plane_ip': res[pillar.bootstrap_id]['ret']}
+    ) %}
+    {%- break %}
+  {%- endif %}
```
gdemonet: Python-style is usually rather "try to get something, and handle failure if it's not there" than "check if something is there, and get it if it is". This avoids duplicating lookups and race conditions (although there is no risk of such condition in your case).

```suggestion
  {%- set control_plane_ip = res.get(pillar.bootstrap_id, {}).get('ret') %}
  {%- if control_plane_ip is not none %}
    {%- do bootstrap_grains.update({'control_plane_ip': control_plane_ip}) %}
    {%- break %}
  {%- endif %}
```

gdemonet (dismissed):
LGTM :ok_hand: 
Could further improve this by creating a Jinja macro, for easy re-use in other places where some render-time logic needs to be retried (IMO, we should try to avoid spending too much time in rendering, but... we're kind of stuck here).

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,33 @@
...
+{%- endfor %}
+{%- if 'control_plane_ip' not in bootstrap_grains %}
+Get metalk8s:control_plane_ip grain:
+  test.fail_without_changes:
+    - comment: >-
+        Unable to get metalk8s:control_plane_ip grain from {{ pillar.bootstrap_id }}
+        after {{ max_try }} tries.
+    - require_in:
+      - salt: Deploy CA role on bootstrap minion
```
TeddyAndrieux, line 28: IMHO you should raise in jinja here instead of using a state, especially since I think if control plane ip not available it will raise in jinja when trying to render the pillar_data dict bellow

alexandre-allard: I tried to raise with jinja first, but the traceback is really ugly and not very helpful.

salt/metalk8s/orchestrate/bootstrap/init.sls
alexandre-allard: oh yes, true :facepalm:

salt/metalk8s/orchestrate/bootstrap/init.sls
gdemonet: Wouldn't it be better if templates never raised? Salt output is already not the easiest to parse, if we need to also extract messages from exception tracebacks... :sob:

gdemonet: ~You would still need to put all the rest of this formula in an `{% else %}` block, to avoid generating another traceback few lines below.~

_Edit: ah no, since `bootstrap_grains['control_plane_ip']` would simply return a `jinja2.Undefined` object, and execution would be skipped with your `require_in`. Forget about me :sweat_smile:_

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,32 @@
...
-        fun='grains.get',
-        kwarg={
-            'key': 'metalk8s:workload_plane_ip',
-        },
-    )[pillar.bootstrap_id]['ret']
-%}
 
+{%- set max_try = 5 %}
+{%- set bootstrap_grains = {} %}
+{%- for _ in range(max_try) %}
```
gdemonet, line 6: BTW, could be interesting to add some unit test cases for this logic (would need to customize the mocks to simulate "intermittent failures" though).

salt/metalk8s/orchestrate/bootstrap/init.sls
TeddyAndrieux: > Edit: ah no, since `bootstrap_grains['control_plane_ip']` would simply return a `jinja2.Undefined` object,

Really ? Didn't know.

Ok fair ! Maybe we want a `failhard` then. I think for that kind of check it make sense

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,33 @@
...
-%}
-
-{%- set bootstrap_workload_plane_ip = salt.saltutil.cmd(
-        tgt=pillar.bootstrap_id,
-        fun='grains.get',
-        kwarg={
-            'key': 'metalk8s:workload_plane_ip',
-        },
-    )[pillar.bootstrap_id]['ret']
-%}
```
TeddyAndrieux, line 19: Hmm don't know why we get it here in the first place

salt/metalk8s/orchestrate/bootstrap/init.sls
TeddyAndrieux: As discussed offline with @alexandre-allard-scality it will be handle in another PR since those tests do not exists in development/2.8

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -1,23 +1,33 @@
...
+  {%- endif %}
+{%- endfor %}
+{%- if 'control_plane_ip' not in bootstrap_grains %}
+Get metalk8s:control_plane_ip grain:
+  test.fail_without_changes:
+    - comment: >-
+        Unable to get metalk8s:control_plane_ip grain from {{ pillar.bootstrap_id }}
+        after {{ max_try }} tries.
+    - require_in:
+      - salt: Deploy CA role on bootstrap minion
```
TeddyAndrieux, line 28: Minor: This require_in could/should be removed since we use `failhard`

TeddyAndrieux (approved):
One minor comment but LGTM

PR #3013, opened 2021-01-05
Fix nested sosreport plugin options

ghost:
Dependency on option all removed from option describe for sosreport metalk8s plugin.

PR #3222, opened 2021-03-23
Enrich SOS report content

### LINKED ISSUE #3218: Enrich sos report content
**Component**:

**Why this is needed**:

When there is a deployment issue, it is hard to do some root cause analysis because some logs and other content are missing from the sos report.

**What should be done**:

Add to the sos report all logs, db dump, etc ... that is needed to do operation failure root cause analysis

**Test plan**:

alexandre-allard:
**Component**: sosreport

**Context**:
In our CI, we do not have enough data collected when a build fails, so it's hard to investigate flakies or other issue that might happen.

**Summary**:
* Backport of #3013
* Add Salt minion service logs
* Add Salt configuration directory (/etc/salt)
* Add a Prometheus snapshot
* Add kubectl top nodes & pods
* Add bootstrap and solutions configuration files
* Fix an issue in both containerd and metalk8s plugins that were using an undefined method.

Still need to enable it in the CI (especially enabling Prometheus admin API to be able to generate a snapshot).

**Acceptance criteria**: 
We can retrieve an sosreport with all elements listed above.

---

Closes: #3218

### DIFF
--- a/packages/common/metalk8s-sosreport/metalk8s.py
+++ b/packages/common/metalk8s-sosreport/metalk8s.py
@@ -1,7 +1,9 @@
 #! /bin/env python3
 
-from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from os import path
+
+import requests
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
 class metalk8s(Plugin, RedHatPlugin, UbuntuPlugin):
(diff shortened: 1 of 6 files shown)


packages/common/metalk8s-sosreport/metalk8s.py
```diff
@@ -72,39 +153,54 @@ def setup(self):
                 if self.get_option('all'):
                     kube_namespaced_cmd = '{0} {1} {2}'.format(kube_cmd, kube_get_cmd, kube_namespace)
 
-                    self.add_cmd_output('{} events'.format(kube_namespaced_cmd))
+                    for subcmd in ["events", "top pods"] + resources:
+                        self.add_cmd_output(
```
TeddyAndrieux, line 157: What the difference between `add_cmd_output` and `exec_cmd`, we should likely use the same everywhere

TeddyAndrieux (commented):
Just a question and comments (already said offline):
- changelog entry
- Add the command to render the prometheus chart in the commit message (just to be consistent)
- A bit of documentation, at least about prometheus snapshot generation

Otherwise it looks good to me

packages/common/metalk8s-sosreport/metalk8s.py
alexandre-allard: `exec_cmd` only executes a command and return the result for further processing, while `add_cmd_output` executes the command and stores it in a file in the sosreport archive.

TeddyAndrieux (approved):
LGTM

PR #3226, opened 2021-03-24
Improve buildchain verbosity and readability

### LINKED ISSUE #3217: Enrich MetalK8s buildchain verbosity/logs
**Component**:

'buildchain'

**Why this is needed**:

When MetalK8s build is failing, it is very hard to understand the cause of the failure because some logs are missing

**What should be done**:

Get all std out/err available
Have timers to measure the time spent in each phase

**Test plan**:

alexandre-allard:
**Component**: buildchain

**Context**:
In the buildchain, we are actually missing some output for debugging purposes and some output are not really readable.
We also do not have timers to see the time taken by each step.

**Summary**:
* Clean up the docker build//run output on failure (make it more readable)
* Add timer for each step

I still need to basically pipe every logs to a file.

**Acceptance criteria**: 

---

Closes: #3217

### DIFF
--- a/buildchain/buildchain/action.py
+++ b/buildchain/buildchain/action.py
@@ -1,0 +1,36 @@
+# coding: utf-8
+"""Implement custom actions used by doit tasks"""
+
+from pathlib import Path
+from typing import Any, Callable, List, Optional, Union
+
+import doit  # type: ignore
+
+
+class CmdActionOnFailure(doit.action.CmdAction):  # type: ignore
+    """Wrapper around CmdAction allowing to pass a callable run if action fails"""
+
+    def __init__(
+        self,
+        action: Union[str, List[Union[str, Path]], Callable[..., Optional[Any]]],
+        task: Optional[doit.task.Task] = None,
+        save_out: Optional[str] = None,
+        shell: bool = True,
+        encoding: str = "utf-8",
+        decode_error: str = "replace",
+        buffering: int = 0,
+        on_failure: Optional[Callable[..., None]] = None,
+        **pkwargs: Any
+    ):
+        super().__init__(
+            action, task, save_out, shell, encoding, decode_error, buffering, **pkwargs
+        )
+        self.on_failure = on_failure
+
+    def execute(
+        self, out: Any = None, err: Any = None
+    ) -> doit.exceptions.CatchedException:
+        failure = super().execute(out, err)
+        if failure and self.on_failure:
+            self.on_failure()
+        return failure
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -6,6 +6,7 @@ Expose Docker commands (build, run, …) as Python classes, using Docker API.
 The instantiated objects are callable and can be used as doit action directly.
 """
 
+import ast
 import copy
 import functools
 import os
@@ -79,7 +80,14 @@ def build_error_handler(build_error: BuildError) -> str:
 
 def container_error_handler(container_error: ContainerError) -> str:
     """String formatting exception handler for Docker API ContainerError."""
-    return "{}:\n{}".format(str(container_error), container_error.stderr)
+    try:
+        err = ast.literal_eval(str(container_error.stderr)).decode()
+    except (ValueError, SyntaxError):
+        err = str(container_error.stderr)
+    msg = "Command '{0}' in image '{1}' returned non-zero exit status {2}:\n{3}".format(
+        container_error.command, container_error.image, container_error.exit_status, err
+    )
+    return msg
 
 
 def task_error(
--- a/buildchain/buildchain/docker_command.py
+++ b/buildchain/buildchain/docker_command.py
@@ -63,11 +63,18 @@ def build_error_handler(build_error: BuildError) -> str:
         if "stream" in item:
             line = item["stream"]
         elif "status" in item:
-            line = "{}: {}/{}".format(
-                item["status"],
-                item["progressDetail"]["current"],
-                item["progressDetail"]["total"],
-            )
+            try:
+                line = "{0}: ".format(item["id"])
+            except KeyError:
+                line = ""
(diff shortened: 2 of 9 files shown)


buildchain/dodo.py
```diff
@@ -112,6 +120,11 @@ def _write_failure(self, result: Dict[str, Any]) -> None:
         self._write(result["exception"].get_msg())
         self._write("\n")
 
+    def _write_time_elapsed(self, task: doit.task.Task) -> None:
+        if task.actions:
+            time_elapsed = int(self.t_results[task.name].to_dict()["elapsed"])
```
gdemonet: `TaskResult.elapsed` should be of type `Optional[float]` already, so you're using the type cast as an implicit guard against `None` values? Or just to essentially truncate the float value to remove sub-second precision digits? Either way, could be better to be explicit, either with a comment or with more explicit code:

```python
# Use `to_dict` to evaluate the "elapsed" value
task_result = self.t_results[task.name].to_dict()
elapsed = task_result["elapsed"]
assert elapsed is not None, "Some runtime error message"
# `elapsed` is a float, but we strip sub-second digits
self._write(f" [{elapsed:.0f}s]\n")
```

buildchain/buildchain/action.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        self.on_failure = on_failure
+
+    def execute(
+        self, out: Any = None, err: Any = None
+    ) -> doit.exceptions.CatchedException:
+        failure = super().execute(out, err)
+        if failure and self.on_failure:
+            self.on_failure()
+        return failure
```
TeddyAndrieux, line 36: Maybe it's also something that can be part of upstream doit ? (just a question not said it should be done right now)

tox.ini
```diff
@@ -23,7 +23,7 @@ setenv =
 commands =
     bash -c ' \
         pip-compile \
-            --emit-index-url --emit-trusted-host --annotate --generate-hashes \
+            --emit-index-url --emit-trusted-host --annotate \
```
TeddyAndrieux, line 26: Why ? It does not work when using `git` with a hash ?

buildchain/buildchain/docker_command.py
```diff
@@ -63,11 +63,18 @@ def build_error_handler(build_error: BuildError) -> str:
...
+            except KeyError:
+                line = ""
+            line += item["status"]
+            try:
+                line += ": {0}/{1}\x1b[1K\r".format(
+                    item["progressDetail"]["current"],
+                    item["progressDetail"]["total"],
+                )
+            except KeyError:
+                line += "\n"
```
TeddyAndrieux, line 77: Not sure to understand why we only add a break line if we do not have progressDetail

TeddyAndrieux (dismissed):
Few minor comments/questions, otherwise LGTM

tox.ini
alexandre-allard: Yes

buildchain/buildchain/docker_command.py
alexandre-allard: If we display progressDetail, we want it to be behave like it's done on CLI output, re-writring previous progress line with the new one, otherwise we want to add a newline for the sake of readability (to avoid having everything on a single line).
To help you better understand, here is an example of what's in the logs, each line being an `item`:
```
{'status': 'Pulling from library/centos', 'id': 'sha256:5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121'}
{'status': 'Pulling fs layer', 'progressDetail': {}, 'id': '581e78aaf612'}
{'status': 'Downloading', 'progressDetail': {'current': 73603419, 'total': 75167150}, 'progress': '[================================================>  ]   73.6MB/75.17MB', 'id': '581e78aaf612'}
[...]
{'status': 'Downloading', 'progressDetail': {'current': 74680667, 'total': 75167150}, 'progress': '[=================================================> ]  74.68MB/75.17MB', 'id': '581e78aaf612'}
{'status': 'Verifying Checksum', 'progressDetail': {}, 'id': '581e78aaf612'}
{'status': 'Download complete', 'progressDetail': {}, 'id': '581e78aaf612'}
{'status': 'Extracting', 'progressDetail': {'current': 557056, 'total': 75167150}, 'progress': '[>                                                  ]  557.1kB/75.17MB', 'id': '581e78aaf612'}
[...]
{'status': 'Extracting', 'progressDetail': {'current': 69074944, 'total': 75167150}, 'progress': '[=============================================>     ]  69.07MB/75.17MB', 'id': '581e78aaf612'}
{'status': 'Pull complete', 'progressDetail': {}, 'id': '581e78aaf612'}
{'status': 'Digest: sha256:5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121'}
{'status': 'Status: Downloaded newer image for centos@sha256:5d4f4e6051c7cc10f2e712f9dc3f86a2bd67e457bced7ca52a71c243099c0121'}
```

alexandre-allard: We could also not display this progressDetail at all, since normally it's overwritten by the next line (e.g. 'Verifying Checksum'), it's just in case the download/extract would be stopped in the middle.

TeddyAndrieux: Oh ok, got it, thank you for the explanation :ok_hand:

buildchain/buildchain/action.py
alexandre-allard: Yes, maybe, but since anyone can easily overload this class, it's not an issue (the other fix was necessary since it touches a part that can't be overloaded).

TeddyAndrieux (approved):

buildchain/buildchain/action.py
TeddyAndrieux: Yes sure it's not needed

PR #3230, opened 2021-03-25
[UI] Fix the bugs of volume alertlist

ChengYanJin:
**Component**: UI, alerts

**Context**: 

**Summary**:
`alerts.data` would still get the entire alert list before the filter.  

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


JBWatenbergScality (approved):

thomasdanan:
@ChengYanJin I don't know in what exact version the bug was reported but we should fix it in the version in which it appears, so probably before 2.9.0

PR #3260, opened 2021-04-02
shell-ui: Share events and logic with consumers

JBWatenbergScality:
**Component**:

shell-ui

**Context**: 

In order to ease change over time on events name and shell ui shared logic, client apps should be able to import common logic.

**Summary**:

Add a library exposing event names and common logic to window global object. The shared object is namespaced to shell-ui and segregated per version.

**Acceptance criteria**: 

Integration tests

### DIFF
--- a/shell-ui/src/library.js
+++ b/shell-ui/src/library.js
@@ -1,0 +1,19 @@
+import {
+  AUTHENTICATED_EVENT,
+  LANGUAGE_CHANGED_EVENT,
+  THEME_CHANGED_EVENT,
+} from './events';
+import { version } from '../package.json';
+
+window.shellUINavbar = {
+  ///spread shellUI to keep all versions libraries
+  ...window.shellUINavbar,
+  [version]: {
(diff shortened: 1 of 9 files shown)


shell-ui/src/library.js
```diff
@@ -0,0 +1,19 @@
+import {
+  AUTHENTICATED_EVENT,
+  LANGUAGE_CHANGED_EVENT,
+  THEME_CHANGED_EVENT,
+} from './events';
+import { version } from '../package.json';
+
+window.shellUI = {
+  ///spread shellUI to keep all versions libraries
+  ...window.shellUI,
```
gdemonet: Are we expecting multiple "shellUI"s in a single window?

JBWatenbergScality: Ideally we don't want this but this could be the case in the futur : for example a micro app component could import shell-ui `1.0.0` and on the same page another component could rely on `1.1.0` ending up in the necessity to keep both libraries

gdemonet (dismissed):

gdemonet (approved):

PR #3261, opened 2021-04-06
repo: Disable repo gpg check for Kubernetes repo

TeddyAndrieux:
Due to a bug on Kubernetes side we cannot check the gpg key for the
repository
Suggested workaround is to disable this repo gpg check especially since
we use "trusted" https endpoints

Sees: https://cloud.google.com/compute/docs/troubleshooting/known-issues#keyexpired


gdemonet (approved):

TeddyAndrieux:
Not needed since it seems fixed upstream

PR #3268, opened 2021-04-06
Improvement/3088 overview metrics

### LINKED ISSUE #3088: MetalK8s UI Overview Page Metrics Panel
**Component**: ui

**Why this is needed**:

**What should be done**:
image

Questions to @thomasdanan 
1. We already have the metrics for each of the nodes, what **aggregation** we should do? **Sum** or **average** of all the nodes?
2. Which Grafana dashboard we should be redirected to when clicking on "Advanced Metrics"? To Node Details? 

As discussed with @Cuervino, to achieve better UX, during the loading state, we should display the skeleton of the charts, as least to keep the space for charts to avoid the whole layout changes.

For the read and write, we prefer to separate them into two quadrants
image

The color guidance of linechart:
image

1. CPU Usage - one line per node. 
The same promQL with node metric in Node Page.
When the number of nodes bigger than 4, use the variant of each color.

2. System Load -  one line per node.

3. Memory  - one line per node.

4. Replace `IOPS` with `Throughput` 
We will Separate `read` and `write` into two quadrants like the following example from Grafana.
image
The PromQL should be the aggregation of Throughput (read&write) on all devices for this node.
 
 We need to do aggregation for all the volumes on this node. 
 ```js
 const nodeThroughputReadQuery = `sum(sum(irate(node_disk_read_bytes_total[1m])) by (instance, device) * 0.000001) by(instance)`;
 const nodeThroughputWriteQuery = `sum(sum(irate(node_disk_written_bytes_total[1m])) by (instance, device) * 0.000001)by(instance)`;
```
 
Note that: 
1. We implemented a feature gate for the overview/dashboard page, you will need to add `flag:["dashboard"]` in the `config.json` to have the access to `/dashboard` 
2. Please use `react-query` to retrieve the metrics.

**Test plan**:
We should test the loading state, error state and success state.

### LINKED ISSUE #3345: Update the layout dashboard page in MetalK8s UI
**Component**: UI

**Why this is needed**:
Following the rebranding of MetalK8s UI, we should rebrand the dashboard page as well as the layout. 
Then work on the specific panels.

**What should be done**:
Please find the latest design here:
https://github.com/scality/metalk8s/issues/3088#issuecomment-827698089
Should also work on 1024*768 screen

**Test plan**:

### PR DESCRIPTION
**Pull request by alexis-ld**
**Component**: ui, dashboard

**Context**: 
We want to add a metrics panel in the dashboard page. More infos on #3088 

**Summary**:

https://user-images.githubusercontent.com/3480526/115054259-54f2e700-9ee0-11eb-9651-48b14986a085.mov

---

Closes: #3088 
Closes: #3345

### DIFF
--- a/ui/src/components/DashboardChartCpuUsage.js
+++ b/ui/src/components/DashboardChartCpuUsage.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import { LineChart, Loader } from '@scality/core-ui';
+import { yAxisUsage } from './LinechartSpec';
+import { GraphTitle, GraphWrapper } from './DashboardMetrics';
+import { formatNodesPromRangeForChart } from '../services/graphUtils';
+import { useQueries } from 'react-query';
+import { queryNodeCPUMetrics } from '../services/prometheus/fetchMetrics';
+import { useNodeAddressesSelector, useNodes } from '../hooks';
+import type { DashboardChartProps } from '../containers/DashboardPage';
+
+const DashboardChartCpuUsage = (props: DashboardChartProps) => {
+  const nodeAddresses = useNodeAddressesSelector(useNodes());
+
+  // Passing nodes table as a react-queries identifier so if a node is added/removed the data are refreshed
+  // Also it makes the data to auto-refresh based on the node refresh timeout that is already implemented
+  const cpuDataQuery = useQueries(
+    nodeAddresses.map((node) => {
+      return {
+        queryKey: ['nodeMetricsCPU', node.name, props.metricsTimeSpan],
+        queryFn: () =>
+          queryNodeCPUMetrics(node.internalIP, props.metricsTimeSpan),
+        ...props.reactQueryOptions,
+      };
+    }),
+  );
+
+  const isNodeLoading = cpuDataQuery.some((query) => query.isLoading);
+
+  return (
+    <GraphWrapper>
+      <GraphTitle>
+        <div>CPU Usage (%)</div>
+        {isNodeLoading && <Loader />}
+      </GraphTitle>
+      <LineChart
+        id={'dashboard_cpu_id'}
+        data={formatNodesPromRangeForChart(
+          cpuDataQuery.map((query) => query.data),
+          nodeAddresses,
+        )}
+        xAxis={props.xAxis}
+        yAxis={yAxisUsage}
+        color={props.perNodeColor}
+        width={props.graphWidth - 35}
+        height={props.graphHeight}
+        lineConfig={props.lineConfig}
+        tooltip={true}
+        tooltipConfig={props.perNodeTooltip}
+        tooltipTheme={'dark'}
+      />
+    </GraphWrapper>
+  );
+};
+
+export default DashboardChartCpuUsage;
--- a/ui/src/components/DashboardMetrics.js
+++ b/ui/src/components/DashboardMetrics.js
@@ -1,0 +1,193 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useQuery } from 'react-query';
+import { Button } from '@scality/core-ui';
+import { padding } from '@scality/core-ui/dist/style/theme';
+import { lighten, darken } from 'polished';
+import { intl } from '../translations/IntlGlobalProvider';
+
+import { fetchConfig } from '../services/api';
+import {
+  useTypedSelector,
+  useNodes,
+  useMetricsTimeSpan,
+  useNodeAddressesSelector,
+} from '../hooks';
+import {
+  lineColor1,
+  lineColor2,
+  lineColor3,
+  lineColor4,
+  SAMPLE_DURATION_LAST_SEVEN_DAYS,
+  REFRESH_METRICS_GRAPH,
+} from '../constants';
+import { getTooltipConfig } from './LinechartSpec';
+import {
+  GraphTitle as GraphTitleCommon,
+  GraphWrapper as GraphWrapperCommon,
+  PageSubtitle,
+} from '../components/style/CommonLayoutStyle';
+import { useDynamicChartSize } from '../services/utils';
+import { expressionFunction } from 'vega';
+import DashboardChartSystemLoad from './DashboardChartSystemLoad';
+import DashboardChartThroughput from './DashboardChartThroughput';
+import DashboardChartMemory from './DashboardChartMemory';
+import DashboardChartCpuUsage from './DashboardChartCpuUsage';
+
+// Custom formatter to display negative value as an absolute value in throughput chart
+expressionFunction('throughputFormatter', function (datum) {
+  return Math.abs(datum).toFixed(2);
+});
+
+const MetricsContainer = styled.div`
+  padding: 2px ${padding.smaller};
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+`;
+
+export const GraphWrapper = styled(GraphWrapperCommon)`
+  position: relative;
+  min-height: 90px;
+  padding: 2px 0px;
+
+  .sc-loader {
+    background: none;
+  }
+`;
+
+export const GraphTitle = styled(GraphTitleCommon)`
+  padding-top: 0px;
+  padding-left: ${padding.small};
+`;
+
+const GraphsWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  flex-grow: 1;
+`;
+
+const PanelActions = styled.div`
+  display: flex;
+  padding: ${padding.small};
+  align-items: center;
+  justify-content: space-between;
+`;
+
+// We use 4 main color from the palette and decline them (lighter/ darker) when we have more than 4 datasets
+export const colorRange = [
+  lineColor1,
+  lineColor2,
+  lineColor3,
+  lineColor4,
+  lighten(0.3, lineColor1),
+  lighten(0.3, lineColor2),
+  lighten(0.3, lineColor3),
+  lighten(0.3, lineColor4),
+  darken(0.2, lineColor1),
+  darken(0.2, lineColor2),
+  darken(0.2, lineColor3),
+  darken(0.2, lineColor4),
+];
+
+const DashboardMetrics = () => {
+  const theme = useTypedSelector((state) => state.config.theme);
+  const nodeAddresses = useNodeAddressesSelector(useNodes());
+  const [metricsTimeSpan] = useMetricsTimeSpan();
+
+  // Get dynamic chart size for 1 column, 4 rows
+  const [graphWidth, graphHeight] = useDynamicChartSize(
+    'dashboard-metrics-container',
+    1,
+    4,
+  );
+
+  const xAxis = {
+    field: 'date',
+    type: 'temporal',
+    axis: {
+      // Refer to all the available time format: https://github.com/d3/d3-time-format#locale_format
+      format:
+        metricsTimeSpan < SAMPLE_DURATION_LAST_SEVEN_DAYS
+          ? '%m/%d %H:%M'
+          : '%m/%d',
+      ticks: true,
+      tickCount: 4,
+      labelColor: theme.brand.textSecondary,
+      labelFlush: 20,
+    },
+    title: null,
+  };
+
+  const perNodeColor = {
+    field: 'type',
+    type: 'nominal',
+    scale: {
+      range: colorRange,
+    },
+    legend: null,
+  };
+
+  const perNodeTooltip = getTooltipConfig(
+    nodeAddresses.map((node) => ({
+      field: node.name.replace('.', '\\.'),
+      type: 'quantitative',
+      title: node.name,
+      format: '.2f',
+    })),
+  );
+
+  const lineConfig = { strokeWidth: 1.5 };
+
+  // App config, used to generated Advanced metrics button link
+  const configQuery = useQuery('appConfig', fetchConfig);
+
+  const chartsConfigProps = {
+    metricsTimeSpan: metricsTimeSpan,
+    xAxis: xAxis,
+    perNodeColor: perNodeColor,
+    graphWidth: graphWidth,
+    graphHeight: graphHeight,
+    lineConfig: lineConfig,
+    perNodeTooltip: perNodeTooltip,
+    reactQueryOptions: {
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      refetchInterval: REFRESH_METRICS_GRAPH,
+      refetchIntervalInBackground: true,
+    },
+  };
+
+  return (
+    <MetricsContainer id="dashboard-metrics-container">
+      <PanelActions>
+        <PageSubtitle>{intl.translate('metrics')}</PageSubtitle>
+        {configQuery.isSuccess && configQuery.data.url_grafana && (
+          <Button
+            text={intl.translate('advanced_metrics')}
+            variant={'buttonSecondary'}
+            icon={<i className="fas fa-external-link-alt" />}
+            size={'small'}
+            href={`${configQuery.data.url_grafana}/dashboard/db/nodes-detailed`}
+            target="_blank"
+            rel="noopener noreferrer"
+            data-cy="advanced_metrics_node_detailed"
+          />
+        )}
+      </PanelActions>
+      <GraphsWrapper>
+        {graphWidth !== 0 && (
+          <>
+            <DashboardChartCpuUsage {...chartsConfigProps} />
+            <DashboardChartMemory {...chartsConfigProps} />
+            <DashboardChartThroughput {...chartsConfigProps} />
+            <DashboardChartSystemLoad {...chartsConfigProps} />
+          </>
+        )}
+      </GraphsWrapper>
+    </MetricsContainer>
+  );
+};
+
+export default DashboardMetrics;
--- a/ui/src/containers/DashboardPage.js
+++ b/ui/src/containers/DashboardPage.js
@@ -1,31 +1,42 @@
 //@flow
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { useHistory } from 'react-router';
+import { UseQueryOptions } from 'react-query';
 import styled from 'styled-components';
+import DashboardMetrics from '../components/DashboardMetrics';
+import { padding } from '@scality/core-ui/dist/style/theme';
+import { Dropdown } from '@scality/core-ui';
+
+import {
+  LAST_SEVEN_DAYS,
+  LAST_TWENTY_FOUR_HOURS,
+  LAST_ONE_HOUR,
+  queryTimeSpansCodes,
+} from '../constants';
+
+import { useURLQuery } from '../services/utils';
 
 const DashboardGrid = styled.div`
   display: grid;
   gap: 8px;
   grid-template:
-    'health    health   network metrics' 160px
-    'inventory services network metrics' 160px
-    'alerts    services network metrics' auto
-    / 1fr 1fr 1fr 1fr;
-
-  @media (max-width: 1024px) {
-    grid-template:
-      'health    health' 160px
-      'inventory services' 160px
-      'alerts    services' minmax(160px, auto)
-      'network   metrics' auto
-      / 1fr 1fr;
-  }
+    'header header header header header' 40px
+    'health     health  health  health  health' 100px
+    'inventory  network network metrics metrics' auto
+    / 1fr 1fr 1fr 1fr 1fr;
 
   height: calc(100vh - 48px - 8px);
   margin: 0 8px 0 8px;
 
-  div {
-    background-color: ${(props) => props.theme.brand.primary};
+  > div {
+    background-color: ${(props) => {
+      return props.theme.brand.backgroundLevel2;
+    }};
     color: ${(props) => props.theme.brand.textPrimary};
+    padding: 2px ${padding.smaller};
+    .sc-vegachart svg {
+      background-color: inherit !important;
+    }
   }
   .health {
     grid-area: health;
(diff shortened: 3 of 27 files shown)

### CODE REVIEW

File: ui/src/components/DashboardMetrics.js [hunk 1]
```diff
@@ -0,0 +1,234 @@
...
+} from '../components/style/CommonLayoutStyle';
+import { useDynamicChartSize } from '../services/utils';
+
+// import styled from 'styled-components';
+
+const DashboardMetrics = () => {
+  // Making sure nodes list is loaded
+  useRefreshEffect(refreshNodesAction, stopRefreshNodesAction);
+  const nodes = useTypedSelector((state) => state.app.nodes.list);
+  const theme = useTypedSelector((state) => state.config.theme);
```
ChengYanJin: We would eventually get rid of redux.  So I suggest using `react-query` to achieve it.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,234 @@
...
+
+  const xAxis = {
+    field: 'date',
+    type: 'temporal',
+    axis: {
+      // Refer to all the available time format: https://github.com/d3/d3-time-format#locale_format
+      format:
+        metricsTimeSpan === (LAST_ONE_HOUR || LAST_TWENTY_FOUR_HOURS)
+          ? '%H:%M'
+          : '%m/%d',
```
ChengYanJin: From the screenshot, I see the x-axis is always the same date, not sure what's the timespan though. Please add the timespan selector with three choices.

ui/src/containers/DashboardPage.js
```diff
@@ -26,6 +28,10 @@ const DashboardGrid = styled.div`
   div {
     background-color: ${(props) => props.theme.brand.primary};
     color: ${(props) => props.theme.brand.textPrimary};
+    padding: ${padding.smaller};
+    .sc-vegachart svg {
+      background-color: inherit !important;
+    }
   }
```
ChengYanJin: Better go inside the `<DashboardMetrics/>`component. Since it's the detail about this component.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,234 @@
...
+            data={memoryDataQuery.data}
+            xAxis={xAxis}
+            yAxis={yAxisUsage}
+            color={perNodeColor}
+            width={graphWidth}
+            height={graphHeight}
+            lineConfig={lineConfig}
+            tooltip={true}
+            tooltipConfig={perNodeTooltip}
+            tooltipTheme={'dark'}
```
ChengYanJin: We added a custom tooltip for vega-tooltip. 
Once we release a new version of `core-ui`, you can benefit from it.
```tooltipTheme={'custom'}```

File: ui/src/components/DashboardMetrics.js (hunk 2 above)
Comment (alexis-ld) [reply ui/src/components/DashboardMetrics.js]: For now it's static on LAST_TWENTY_FOUR_HOURS. The design does not show a timespan selector so I figured it would be static.

@ChengYanJin @Cuervino there is not much space between the panel title and the Advanced metrics button on the design, where should I put this ?

File: ui/src/containers/DashboardPage.js (hunk 3 above)
Comment (alexis-ld) [reply ui/src/containers/DashboardPage.js]: @ChengYanJin I put it there because we will have charts in the Network panel and we will need the same background on those as well. :) Should I still move it and then copy the code in both components ?

**Comment by alexis-ld** [comment]
@ChengYanJin Switching to ready for review PR removing Draft.
Added Throughput chart, panel title and Advanced metrics button.
As we discussed I also removed the timespan selector since it will be handled at the top level of the dashboard;

Also it seems the overflowing elements at the bottom of the page are not visible. But as we discussed it does not matter since the layout will soon be updated.

File: ui/src/services/utils.js [hunk 5]
```diff
@@ -419,18 +419,21 @@ export const useTableSortURLSync = (sorted, desc, data, defaultSortKey) => {
...
-    ?.offsetWidth;
+export const useDynamicChartSize = (container_id, columns = 2, rows = 3) => {
+  const graphsContainerWidth =
+    document.getElementById(container_id)?.offsetWidth;
   const [graphWidth, setGraphWidth] = useState(0);
   useEffect(() => {
-    if (graphsContainerWidth) setGraphWidth(graphsContainerWidth / 2 - 50);
-  }, [graphsContainerWidth]);
+    if (graphsContainerWidth) {
+      if (columns === 1) setGraphWidth(graphsContainerWidth - 100);
```
ChengYanJin: I think the width of the charts should be bigger to fully use the space on the right.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,443 @@
...
+  left: 30px;
+  z-index: ${zIndex.overlay};
+`;
+
+const PanelActions = styled.div`
+  display: flex;
+  justify-content: space-between;
+`;
+
+// import styled from 'styled-components';
```
ChengYanJin: can be removed

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,443 @@
...
+    1,
+    4,
+  );
+
+  // We use 4 main color from the palette and decline them (lighter/ darker) when we have more than 4 datasets
+  const colorRange = [
+    '#245A83',
+    '#808080',
+    '#A04EC9',
+    '#C6B38A',
```
ChengYanJin: I would suggest we create a global constant for those colors and name them maybe `linePrimary`, `lineSecondary` since we don't have those colors in the color palette yet.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,443 @@
...
+      <GraphWrapper>
+        <GraphTitle>
+          <div>Memory</div>
+        </GraphTitle>
+        {memoryDataQuery.isLoading && (
+          <Loader size="massive" topPosition={graphHeight - 10} />
+        )}
+        <LineChart
+          id={'dashboard_memory_id'}
+          data={memoryDataQuery.isSuccess ? memoryDataQuery.data : noData}
```
ChengYanJin: are you sure we need this given
```
  useEffect(() => {
    if (!nodesQuery.isLoading && nodesQuery.isSuccess) {
      setNodes(nodesQuery.data);
    }
  }, [nodesQuery]);
  ```
anyway, we should display the loader if the data is an empty array, no?

File: ui/src/components/DashboardMetrics.js [hunk 9]
```diff
@@ -0,0 +1,443 @@
...
+          xAxis={xAxis}
+          yAxis={yAxisUsage}
+          color={memoryDataQuery.isSuccess ? perNodeColor : null}
```
ChengYanJin: not sure we need this either. what's the reason?

ui/src/services/utils.js
```diff
@@ -419,18 +419,21 @@ export const useTableSortURLSync = (sorted, desc, data, defaultSortKey) => {
...
+      else setGraphWidth(graphsContainerWidth / columns - 90);
```
ChengYanJin: and we suggest not to change the layout for the node metrics and volume metrics page.

ChengYanJin (changes requested):
As well as some issues I already sent you:
- Fix the small responsiveness issue in order to see the legends.
- Errors pop up in the console
  1. ERROR Error: invalid bounds
  2. vega-scenegraph.module.js:4680 Error: <svg> attribute viewBox: Expected number, "0 0 1.79769313486231…".


Please investigate the error, and let me know if you need any help.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,460 @@
...
+  const cpuDataQuery = useQuery(
+    ['dashboardMetricsCPU', nodes, metricsTimeSpan],
+    useCallback(
+      () =>
+        Promise.all(
+          nodes.map((node) =>
+            queryNodeCPUMetrics(node.internalIP, metricsTimeSpan),
+          ),
+        ).then((result) => formatNodesPromRangeForChart(result)),
+      [nodes, metricsTimeSpan, formatNodesPromRangeForChart],
```
ChengYanJin: what's the reason we put `formatNodesPromRangeForChart` as useCallback dependency?

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,460 @@
...
+          format: '.2f',
+        });
+      });
+      return res;
+    })(nodes),
+  );
+
+  const lineConfig = { strokeWidth: 1.5 };
+
+  const formatNodesPromRangeForChart = useCallback(
```
ChengYanJin: I suggest we add unit test for `formatNodesPromRangeForChart` and `formatNodesThroughputPromRangeForChart`

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,460 @@
...
+    type: 'nominal',
+    scale: {
+      range: getThroughputColorRange(nodes.length),
+    },
+    legend: null,
+  };
+
+  const perNodeTooltip = getTooltipConfig(
+    nodes.map((node) => ({
+      field: node.name,
```
ChengYanJin: The value in the tooltips is NaN :(

ui/src/hooks.js
```diff
@@ -10,3 +14,35 @@ export const useTypedSelector: <TSelected>(
...
   equalityFn?: (left: TSelected, right: TSelected) => boolean,
 ) => TSelected = useSelector;
+
+/**
+ * It retrieves the nodes data through react-queries
+ */
+export const useNodes = (): Object[] => {
+  const [nodes, setNodes] = useState([]);
+
+  const nodesQuery = useQuery(['nodesNames'], () =>
```
ChengYanJin:
```suggestion
  const nodesQuery = useQuery('nodesNames', () =>
```

ChengYanJin (changes requested):
1. I would not expect to reload the graph when I switch back to the dashboard page from the Node page for example. Some optimation needs to be done around `react-query`. 

2. We should have graph refresh every 1 min(REFRESH_METRCIS_GRAPH). Base on the current implementation I think we miss this functionality.

3. I still can reproduce the console error `ERROR Error: invalid bounds` by the following path:
Go to dashboard page => go to node page => go back to dashboard page.

ui/src/components/DashboardMetrics.js
alexis-ld: @ChengYanJin to avoid the warning: 
` React Hook useCallback has a missing dependency: 'formatNodesPromRangeForChart'. Either include it or remove the dependency array  react-hooks/exhaustive-deps`

React wants to know it needs to update the function if formatNodesPromRangeForChart changes (it won't happen) but it considers it's a good practice to include it.

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  GraphWrapper as GraphWrapperCommon,
+  PageSubtitle,
+} from '../components/style/CommonLayoutStyle';
+import {
+  useDynamicChartSize,
+  useQuery as useURLQuery,
+} from '../services/utils';
+import { expressionFunction } from 'vega';
+
+// Custom formatter to display negative value as absolut value in throughput chart
```
JBWatenbergScality:
```suggestion
// Custom formatter to display negative value as an absolute value in throughput chart
```

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  justify-content: space-around;
+  flex-grow: 1;
+`;
+
+const PanelActions = styled.div`
+  display: flex;
+  justify-content: space-between;
+`;
+
+export const formatNodesPromRangeForChart = (result, nodes) => {
```
JBWatenbergScality: We would like to enforce the typing in order to prevent some predictable errors. Could you please add it for the arguments and the return type ?

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+        type: node.name,
+        y: item[1],
+      }));
+    }
+    return acc.concat(temp);
+  }, []);
+  return reduced;
+};
+
+export const formatNodesThroughputPromRangeForChart = (result, nodes) => {
```
JBWatenbergScality: We would like to enforce the typing in order to prevent some predictable errors. Could you please add it for the arguments and the return type ?

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+};
+
+const DashboardMetrics = () => {
+  const theme = useTypedSelector((state) => state.config.theme);
+  const nodes = useNodes();
+  const query = useURLQuery();
+  const queryTimeSpan = query.get('from');
+
+  const [metricsTimeSpan, setMetricsTimeSpan] = useState(
+    LAST_TWENTY_FOUR_HOURS,
```
JBWatenbergScality: Later on we want the time span to be the same one across the application if the user change it. I believe we should introduce a context and a custom hook to retrieve the selected / set the chart time span so that component can use the same one globally available.

JBWatenbergScality: Once the time span mutualised in a context we can also move the query parma from behavior to that global context and avoid duplicating this behavior in all the components using charts

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+    lighten(0.3, lineColor2),
+    lighten(0.3, lineColor3),
+    lighten(0.3, lineColor4),
+    darken(0.2, lineColor1),
+    darken(0.2, lineColor2),
+    darken(0.2, lineColor3),
+    darken(0.2, lineColor4),
+  ];
+
+  const getThroughputColorRange = (size) => {
```
JBWatenbergScality: Please add some typing for the arguments and return type

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  };
+
+  const xAxis = {
+    field: 'date',
+    type: 'temporal',
+    axis: {
+      // Refer to all the available time format: https://github.com/d3/d3-time-format#locale_format
+      format:
+        metricsTimeSpan === LAST_TWENTY_FOUR_HOURS ||
+        metricsTimeSpan === LAST_ONE_HOUR
```
JBWatenbergScality: I suggest to switch this implementation to an interval base time selection as we may want to give more control to the user later on to select a specific time period

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+
+      return acc.concat(tempRead).concat(tempWrite);
+    }, []);
+    return reduced;
+  }
+
+  return null;
+};
+
+const DashboardMetrics = () => {
```
JBWatenbergScality, line 94: In order to simplify maintainability of this component I would suggest to extract all the graphs related functions to a specific component

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  // Passing nodes table as a react-queries identifier so if a node is added/removed the data are refreshed
+  // Also it makes the data to auto-refresh based on the node refresh timeout that is already implemented
+  const cpuDataQuery = useQuery(
+    ['dashboardMetricsCPU', nodes, metricsTimeSpan],
+    useCallback(
+      () =>
+        Promise.all(
+          nodes.map((node) =>
+            queryNodeCPUMetrics(node.internalIP, metricsTimeSpan),
+          ),
```
JBWatenbergScality: Those metrics are also loaded in the node page individually so I think it would make sense to use `useQueries` here (https://react-query.tanstack.com/reference/useQueries) in order to avoid re fetching the same data when switching to the node page and mutualise caching

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  );
+
+  const memoryDataQuery = useQuery(
+    ['dashboardMetricsMemory', nodes, metricsTimeSpan],
+    useCallback(
+      () =>
+        Promise.all(
+          nodes.map((node) =>
+            queryNodeMemoryMetrics(node.internalIP, metricsTimeSpan),
+          ),
```
JBWatenbergScality: Similarly to the CPU metrics it would makes sense to use `useQueries` here

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+  );
+
+  const loadDataQuery = useQuery(
+    ['dashboardMetricsLoad', nodes, metricsTimeSpan],
+    useCallback(
+      () =>
+        Promise.all(
+          nodes.map((node) =>
+            queryNodeLoadMetrics(node.internalIP, metricsTimeSpan),
+          ),
```
JBWatenbergScality: Similarly to the CPU metrics it would makes sense to use useQueries here

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,456 @@
...
+
+  // Sync url timespan to local timespan
+  useEffect(() => {
+    if (queryTimeSpan) {
+      let formatted = queryTimeSpansCodes.find(
+        (item) => item.label === queryTimeSpan,
+      );
+      if (formatted && formatted.value) setMetricsTimeSpan(formatted.value);
+    }
+  }, [queryTimeSpan]);
```
JBWatenbergScality: To move to the time span context when applying the previous comment

ui/src/containers/DashboardPage.js
```diff
@@ -42,20 +53,99 @@ const DashboardGrid = styled.div`
...
+    // Needed to avoid dynamically sized charts to expand infinitely on refresh
+    // more info here: https://www.w3.org/TR/css3-grid-layout/#min-size-auto
+    min-width: 0;
+    min-height: 0;
   }
 `;
 
 const DashboardPage = (props: {}) => {
+  const history = useHistory();
+  const query = useQuery();
```
JBWatenbergScality: nit: we might want to rename this hook in order to avoid confusion with react-query when reviewing the code

ui/src/containers/DashboardPage.js
```diff
@@ -42,20 +53,99 @@ const DashboardGrid = styled.div`
...
+    if (queryTimeSpan) {
+      let formatted = queryTimeSpansCodes.find(
+        (item) => item.label === queryTimeSpan,
+      );
+      if (formatted && formatted.value) setMetricsTimeSpan(formatted.value);
+    }
+  }, [queryTimeSpan]);
+
+  // Write the selected timespan in URL
+  const writeUrlTimeSpan = (timespan) => {
```
JBWatenbergScality: Please add some typing here

ui/src/hooks.js
```diff
@@ -10,3 +14,43 @@ export const useTypedSelector: <TSelected>(
   selector: (state: RootState) => TSelected,
   equalityFn?: (left: TSelected, right: TSelected) => boolean,
 ) => TSelected = useSelector;
+
+/**
+ * It retrieves the nodes data through react-queries
+ */
+export const useNodes = (): Object[] => {
```
JBWatenbergScality: We should avoid to use Object type as this is too generic.

ui/src/services/prometheus/api.js
```diff
@@ -69,7 +69,7 @@ export function queryPrometheusRange(
   end: string,
   step: number, // step (time in seconds between 2 metrics)
   query: string, // query (the actual PromQL expression)
-) {
+): function {
```
JBWatenbergScality: function type is too generic. In practice here it is returning a Promise< PrometheusQueryResult >

ui/src/services/utils.js
```diff
@@ -421,18 +421,21 @@ export const useTableSortURLSync = (sorted, desc, data, defaultSortKey) => {
...
- ** This calculates the width for rows of 2 charts
- ** Takes container id as a param and returns [ width, heigth ]
+ ** By default this calculates the width for 3 rows of 2 charts
+ ** Takes container id as a param and optionnally desired number of columns and row and returns [ width, heigth ]
  */
-export const useDynamicChartSize = (container_id) => {
-  const graphsContainerWidth = document.getElementById(container_id)
-    ?.offsetWidth;
-  const [graphWidth, setGraphWidth] = useState(0);
+export const useDynamicChartSize = (container_id, columns = 2, rows = 3) => {
```
JBWatenbergScality: Please add some typing here

ui/src/components/DashboardMetrics.test.js
```diff
@@ -0,0 +1,262 @@
+import {
```
ChengYanJin: Thanks for adding the test!
However, for me `formatNodesPromRangeForChart` is more a util function. 
`DashboardMetrics.test.js` would make me think it's the unit test for the entire component.
Hence, I would suggest moving the test to `graphUtil.test.js`

ui/src/components/DashboardChartCpuUsage.js
```diff
@@ -0,0 +1,68 @@
...
+type Props = {
+  metricsTimeSpan: string,
+  xAxis: Object,
+  perNodeColor: Object,
+  graphWidth: number,
+  graphHeight: number,
+  lineConfig: Object,
+  perNodeTooltip: Object,
+  reactQueryOptions: UseQueryOptions,
+};
```
ChengYanJin: Do we have exactly the same props for all the charts in the dashboard page? Do you think we can share it?

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,260 @@
...
+        type: node.name,
+        y: item[1],
+      }));
+    }
+    return acc.concat(temp);
+  }, []);
+  return reduced;
+};
+
+export const formatNodesThroughputPromRangeForChart = (
```
ChengYanJin: I would suggest extracting this util function to a dedicated file - `graphUtil.js` 
Same as `formatNodesPromRangeForChart`

ui/src/components/DashboardMetrics.js
```diff
@@ -0,0 +1,260 @@
...
+    xAxis: xAxis,
+    perNodeColor: perNodeColor,
+    graphWidth: graphWidth,
+    graphHeight: graphHeight,
+    lineConfig: lineConfig,
+    perNodeTooltip: perNodeTooltip,
+    reactQueryOptions: {
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      refetchInterval: 60000,
```
ChengYanJin: I suggest we use `REFRESH_METRCIS_GRAPH` from constant.js

ui/src/components/DashboardMetrics.test.js
```diff
@@ -0,0 +1,262 @@
...
+          ],
+        },
+      ],
+    },
+  },
+];
+
+const testFailedData = [
+  {
+    status: 'fail',
```
ChengYanJin:
```suggestion
    status: 'error',
```

ui/src/types.js
```diff
@@ -1,12 +1,12 @@
...
 export type APIResult<T> = Result<{ body: T }>;
 
 export type Nodes = Array<Node>;
 
 export type Node = {
-  internalIP: V1NodeAddress.address | undefined,
-  name: V1ObjectMeta.name | undefined,
-};
+  internalIP: V1NodeAddress.address,
+  name: V1ObjectMeta.name,
```
ChengYanJin: Not sure this is correct.
If you want to typing `instanceIP`, you can directly type `string`

ui/src/hooks.js
```diff
@@ -60,3 +61,55 @@ export const useNodes = (): Nodes => {
...
+  return (
+    <MetricsTimeSpanContext.Provider
+      value={{ metricsTimeSpan, setMetricsTimeSpan }}
+    >
+      {children}
+    </MetricsTimeSpanContext.Provider>
+  );
+};
+
+export const useMetricsTimeSpan = (): [
```
ChengYanJin, line 91: I am wondering if it's possible to add some unit tests for this hook?

ChengYanJin (changes requested):
Some small changes request, and please make sure to clean the commit in the end.
Also, in my opinion, we should avoid this reload as much as possible since we already fetch the metrics.
https://user-images.githubusercontent.com/18453133/120627386-e8b64d80-c463-11eb-8ee3-64fe9b6554b8.mov

For this point, one solution I see is not to render the graph until the graphWidth!==0.

ChengYanJin:
Last point: 
Please double with design, if the paddings of the `Advanced button` and the `graph title` are correct.

ui/src/services/graphUtils.test.js
```diff
@@ -0,0 +1,262 @@
...
+          ],
+        },
+      ],
+    },
+  },
+];
+
+const testFailedThroughputData = [
+  {
+    status: 'fail',
```
ChengYanJin:
```suggestion
    status: 'error',
```

ui/src/services/prometheus/fetchMetrics.js
```diff
@@ -30,3 +33,139 @@ export function queryNodeFSSize(
...
+  const currentTime = new Date();
+  const currentTimeISO = currentTime.toISOString(); // To query Prometheus the date should follow `RFC3339` format
+  const startingTimestamp = Math.round(currentTime.getTime() / 1000) - timeSpan;
+  const startingTimeISO = new Date(startingTimestamp * 1000).toISOString();
+
+  return { startingTimeISO, currentTimeISO, sampleFrequency };
+};
+
+export function queryNodeCPUMetrics(
+  instanceIP: $PropertyType<NodeAddress, 'internalIP'>,
```
ChengYanJin:
```suggestion
  instanceIP: string,
```

ui/src/services/prometheus/fetchMetrics.js
```diff
@@ -30,3 +33,139 @@ export function queryNodeFSSize(
...
+  )?.then((resolve) => {
+    if (resolve.error) {
+      throw resolve.error;
+    }
+    return resolve;
+  });
+}
+
+export function queryNodeMemoryMetrics(
+  instanceIP: $PropertyType<NodeAddress, 'internalIP'>,
```
ChengYanJin:
```suggestion
  instanceIP: string,
```

ui/src/hooks.js
```diff
@@ -10,3 +22,97 @@ export const useTypedSelector: <TSelected>(
...
+  useEffect(() => {
+    if (!nodesQuery.isLoading && nodesQuery.isSuccess) {
+      setNodes(nodesQuery.data);
+    }
+  }, [nodesQuery]);
+
+  return nodes;
+};
+
+export const useNodeAddressesSelector = (nodes: Nodes): NodesAddresses => {
```
ChengYanJin:
```suggestion
export const useNodeAddressesSelector = (nodes: Nodes): Array<{internalIP: string, name: string}>=> {
```
IMO, I would be easier to read if we directly type here.

ui/src/types.js
```diff
@@ -1,3 +1,13 @@
 //@flow
+import { V1Node } from '@kubernetes/client-node';
```
ChengYanJin: I think we can directly use `V1NodeList` to type the return of `useNodes()`
import { V1NodeList } from '@kubernetes/client-node';

ui/src/types.js
```diff
@@ -1,3 +1,13 @@
...
+
 export type Result<T> = { error: any } | T;
 export type APIResult<T> = Result<{ body: T }>;
+
+export type NodeAddress = {
+  internalIP: string,
+  name: string,
+};
+export type NodesAddresses = Array<NodeAddress>;
```
ChengYanJin: IMO, all the typing here in this file can be removed and type directly in various places.

ui/src/services/prometheus/fetchMetrics.js
```diff
@@ -30,3 +33,139 @@ export function queryNodeFSSize(
...
+  )?.then((resolve) => {
+    if (resolve.error) {
+      throw resolve.error;
+    }
+    return resolve;
+  });
+}
+
+export function queryNodeLoadMetrics(
+  instanceIP: $PropertyType<NodeAddress, 'internalIP'>,
```
ChengYanJin:
```suggestion
  instanceIP: string,
```

ui/src/services/graphUtils.test.js
```diff
@@ -0,0 +1,262 @@
...
+          metric: {
+            instance: '192.168.1.2:9100',
+          },
+          values: [],
+        },
+      ],
+    },
+  },
+  {
+    status: 'fail',
```
ChengYanJin:
```suggestion
    status: 'error',
```

ChengYanJin (commented):
My last change request ;) mainly about the typing. Thanks for addressing them.

ChengYanJin (approved):
LGTM! Not the scope for this PR, but we need to decide the time format we want to use in graphs.

PR #3275, opened 2021-04-07
UI: Expose hooks that retrieve the highest alerts for each entity

### LINKED ISSUE #3257: UI: Implements hooks to retrieve the health of entities(Volume/Node/Service/Network)
**Component**: UI

**Why this is needed**:
We need to display the health of Volume, Node, Service, and Network. 
This information would need to rely on Metalk8s UI to retrieve and export. 

**What should be done**:
Create methods/hooks for each of the entities. 

- [ ] useNodesAlert(): Alert?
- [ ] useVolumesAlert(): Alert?
- [ ] useServiceAlert(): Alert?
- [ ] useNetworkAlert(): Alert?
- [ ] getNodesCountQuery(): Query // type Query = {cacheKey, queryFn, options} from react-query
- [ ] getVolumesCountQuery(): Query // type Query = {cacheKey, queryFn, options} from react-query

```js
const useNodeStatus = (): alert =>  {
const nodeAlertsWarning = useAlerts({name: ["Nodes_warning","Nodes_critial"]})
	return nodeAlerts[0]
}

```

**Test plan**:

Hooks unit tests

ChengYanJin:
**Component**: ui, alerts

**Context**: 
Provide hooks to fetch the highest severity alerts for each entity to display the health.
```
 useNodesAlert(): Alert?
 useVolumesAlert(): Alert?
 useServiceAlert(): Alert?
 useNetworkAlert(): Alert?
 getNodesCountQuery(): Query // type Query = {cacheKey, queryFn, options} from react-query
 getVolumesCountQuery(): Query // type Query = {cacheKey, queryFn, options} from react-query
```
**Summary**:
```js
const nodesAlerts = useHighestSeverityAlerts(getNodesAlertName());
const volumesAlerts = useHighestSeverityAlerts(getVolumesAlertName());
const volumesAlerts = useHighestSeverityAlerts(getNetworksAlertName());
const volumesAlerts = useHighestSeverityAlerts(getServicesAlertName());
const coutNodes = getNodesCountQuery();
const coutVolumes = getVolumesCountQuery();
```

As discussed with @JBWatenbergScality, we will eventually remove the AlertProvider from Metalk8s UI.

**Acceptance criteria**: 

---

Closes: #3257

### DIFF
--- a/shell-ui/src/alerts/services/alertHooks.js
+++ b/shell-ui/src/alerts/services/alertHooks.js
@@ -27,10 +27,14 @@ export const getServicesAlertName = (): FilterLabels => {
 
 /**
  *
- * @param {FilterLabels} filterss
+ * @param {useContext} useContext
+ * @param {FilterLabels} filters
  * @returns An array of alerts with the highest severity
  */
-export const useHighestSeverityAlerts = (filters: FilterLabels): Alert[] => {
+export const useHighestSeverityAlerts = (
+  useContext: typeof useContext,
+  filters: FilterLabels,
+): Alert[] => {
   const query = useAlerts(useContext)(filters);
   const filteredAlerts = query && query.alerts;
 
--- a/shell-ui/src/alerts/services/alertHooks.js
+++ b/shell-ui/src/alerts/services/alertHooks.js
@@ -1,0 +1,40 @@
+//@flow
+import { useContext } from 'react';
+import { Query } from 'react-query';
+import { useAlerts } from '../AlertProvider';
+import type { FilterLabels, Alert } from './alertUtils';
+import { getHealthStatus } from './alertUtils';
+
+export const getNodesAlertSelectors = (): FilterLabels => {
+  return { alertname: ['NodeAtRisk', 'NodeDegraded'] };
+};
+
+export const getVolumesAlertName = (): FilterLabels => {
+  return { alertname: ['VolumeAtRisk', 'VolumeDegraded'] };
+};
+
+export const getNetworksAlertName = (): FilterLabels => {
+  return {
+    alertname: ['ControlPlaneNetworkDegraded', 'WorkloadPlaneNetworkDegraded'],
+  };
+};
+
+export const getServicesAlertName = (): FilterLabels => {
+  return {
+    alertname: ['PlatformServicesAtRisk', 'PlatformServicesDegraded'],
+  };
+};
+
+/**
+ *
+ * @param {FilterLabels} filterss
+ * @returns An array of alerts with the highest severity
+ */
+export const useHighestSeverityAlerts = (filters: FilterLabels): Alert[] => {
+  const query = useAlerts(useContext)(filters);
+  const filteredAlerts = query && query.alerts;
+
+  const health = getHealthStatus(filteredAlerts);
+  if (!filteredAlerts || !filteredAlerts.length) return [];
+  return filteredAlerts.filter((alert) => alert.severity === health);
+};
--- a/shell-ui/src/alerts/services/alertHooks.test.js
+++ b/shell-ui/src/alerts/services/alertHooks.test.js
@@ -1,0 +1,112 @@
+import '../library';
+import React, { createContext, useContext } from 'react';
+import { version } from '../../../package.json';
+import { setupServer } from 'msw/node';
+import { rest } from 'msw';
+import { renderHook } from '@testing-library/react-hooks';
+import { QueryClient, QueryClientProvider, useQuery } from 'react-query';
+import { AlertProvider } from '../AlertProvider.js';
+import { useHighestSeverityAlerts } from './alertHooks';
+import { afterAll, beforeAll, jest } from '@jest/globals';
+
+const testService = 'http://10.0.0.1/api/alertmanager';
+
+const server = setupServer(
+  rest.get(`${testService}/api/v2/alerts`, (req, res, ctx) => {
+    const alerts = [
+      {
+        annotations: {
+          description:
+            'Filesystem on /dev/vdc at 192.168.1.29:9100 has only 3.13% available space left.',
+          runbook_url:
+            'https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-nodefilesystemalmostoutofspace',
+          summary: 'Filesystem has less than 5% space left.',
+        },
+        endsAt: new Date(
+          new Date().getTime() + 1000 * 60 * 60 * 24,
+        ).toISOString(),
+        fingerprint: '37b2591ac3cdb320',
+        startsAt: '2021-01-25T09:12:05.358Z',
+        updatedAt: '2021-01-29T07:36:11.363Z',
+        generatorURL:
+          'http://prometheus-operator-prometheus.metalk8s-monitoring:9090/graph?g0.expr=%28node_filesystem_avail_bytes%7Bfstype%21%3D%22%22%2Cjob%3D%22node-exporter%22%7D+%2F+node_filesystem_size_bytes%7Bfstype%21%3D%22%22%2Cjob%3D%22node-exporter%22%7D+%2A+100+%3C+5+and+node_filesystem_readonly%7Bfstype%21%3D%22%22%2Cjob%3D%22node-exporter%22%7D+%3D%3D+0%29&g0.tab=1',
+        labels: {
+          alertname: 'VolumeDegraded',
+          severity: 'warning',
+        },
+      },
+      {
+        annotations: {
+          description:
+            'Filesystem on /dev/vdc at 192.168.1.29:9100 has only 3.13% available space left.',
+          runbook_url:
+            'https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-nodefilesystemalmostoutofspace',
+          summary: 'Filesystem has less than 5% space left.',
+        },
+        endsAt: new Date(
+          new Date().getTime() + 1000 * 60 * 60 * 24,
+        ).toISOString(),
+        fingerprint: '37b2591ac3cdb320',
+        startsAt: '2021-01-25T09:12:05.358Z',
+        updatedAt: '2021-01-29T07:36:11.363Z',
+        generatorURL:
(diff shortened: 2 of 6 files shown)


ui/src/services/alertHooks.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export const getServicesAlertName = (): string[] => {
+  return ['PlatformServicesAtRisk', 'PlatformServicesDegraded'];
+};
+
+/*
+The critical and warning alerts may active at the same time.
+how to use:
+  const nodesAlert = useHighestSeverityAlert({
```
JBWatenbergScality: Nit : You may want to use jsdoc notation for this kind of comments to benefit from it in IDE autocompletion 

The name of the functions doesn't match the implementation :
```suggestion
  const nodesAlert = useHighestSeverityAlerts({
```

ui/src/containers/AlertProvider.js
```diff
@@ -22,7 +22,8 @@ export function useAlerts(filters: FilterLabels) {
   return query;
 }
 
-const AlertProvider = ({ children }: any) => {
+// the `fakequery` is for testing purpose in order to inject alerts
+const AlertProvider = ({ children, fakequery }: any) => {
```
JBWatenbergScality:
```suggestion
import type {Node} from "react";
const AlertProvider = ({ children }: {children: Node}) => {
```

Instead of introducing a `fakeQuery` here I would go for adding an optional second parameter to useAlerts, this optional second parameter would be the alert context. Doing so in the test you won't need to render an AlertProvider wrapper anymore and could `createContext` initialised with your mocked query.

JBWatenbergScality: Another alternative (maybe better because in this way it won't impact the definition of useAlerts neither useHighestSeverityAlerts definition) could be to move the context to a separate module, mock with jest.mock and returning a context containing your fakeQuery

ui/src/services/alertHooks.js
```diff
@@ -0,0 +1,69 @@
...
+  return ['VolumeAtRisk', 'VolumeDegraded'];
+};
+
+export const getNetworksAlertName = (): string[] => {
+  return ['ControlPlaneNetworkDegraded', 'WorkloadPlaneNetworkDegraded'];
+};
+
+export const getServicesAlertName = (): string[] => {
+  return ['PlatformServicesAtRisk', 'PlatformServicesDegraded'];
+};
```
JBWatenbergScality: I think a better abstraction would be to return a set of selectors instead of the name, in this way if in the futur we need to narrow down the selection of alerts we won't need to change the definition of those functions.

 ```suggestion
export const getNodesAlertSelectors = (): FilterLabels => {
  return {name: ['NodeAtRisk', 'NodeDegraded']};
};
// ...
```

File: ui/src/services/alertHooks.js [hunk 4]
```diff
@@ -0,0 +1,69 @@
+//@flow
+import { useAlerts } from '../containers/AlertProvider';
+import { getNodes } from './k8s/core.js';
+import { getMetalk8sV1alpha1VolumeList } from './k8s/Metalk8sVolumeClient.generated';
+import type { FilterLabels, Alert } from './alertUtils';
+import { getHealthStatus } from './alertUtils';
+
+type Query = { cacheKey: string, queryFn: Function, options?: Object };
```
JBWatenbergScality: It would be better to reuse Query type from react query

ui/src/containers/AlertProvider.js
ChengYanJin: Sure. thanks for the suggestion. Will go for the second solution!

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+      fetch(`${k8sUrl}/api/v1/nodes`)
+        .then((r) => {
+          if (r.ok) {
+            return r.json();
+          }
+        })
+        .then((res) => {
+          return res.items.length;
+        }),
+    options: { refetchInterval: 10000 },
```
JBWatenbergScality: I think it is missing enabled options :

```suggestion
    options: { refetchInterval: 10000, enabled: token },
```

this will only triggers the query if the token is provided.

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+};
+
+export const getNodesCountQuery = (
+  k8sUrl: string,
+  token?: string | null,
+): typeof useQuery => {
+  if (!token) {
+    console.error('K8s API Not authenticated');
+    return;
+  }
```
JBWatenbergScality: IMO it would be better to get rid of this early return in order to have a consistant return type even when the token is not provided. Ensuring the token is provided to perform the request can be done through the `enabled` useQuery option

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+};
+
+export const getVolumesCountQuery = (
+  k8sUrl: string,
+  token?: string | null,
+): typeof useQuery => {
+  if (!token) {
+    console.error('K8s API Not authenticated');
+    return;
+  }
```
JBWatenbergScality: same as above

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+      })
+        .then((r) => {
+          if (r.ok) {
+            return r.json();
+          }
+        })
+        .then((res) => {
+          return res.items.length;
+        }),
+    options: { refetchInterval: 10000 },
```
JBWatenbergScality: same as above

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+      // $FlowFixMe
+      setToken(evt.detail.id_token);
+    };
+    navbarElement.addEventListener(AUTHENTICATED_EVENT, onAuthenticated);
+    return () =>
+      navbarElement.removeEventListener(AUTHENTICATED_EVENT, onAuthenticated);
+  }, []);
+
+  const queryVolumesResult = useQuery(getVolumesCountQuery(k8sUrl, token));
+  queryVolumesResult.volumesCount = queryVolumesResult.data;
```
JBWatenbergScality, line 56: in the current implementation `queryVolumesResult`  can be null, see bellow my comment about the `enabled` option.

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+      // $FlowFixMe
+      setToken(evt.detail.id_token);
+    };
+    navbarElement.addEventListener(AUTHENTICATED_EVENT, onAuthenticated);
+    return () =>
+      navbarElement.removeEventListener(AUTHENTICATED_EVENT, onAuthenticated);
+  }, []);
+
+  const queryNodesResult = useQuery(getNodesCountQuery(k8sUrl, token));
+  queryNodesResult.nodesCount = queryNodesResult.data;
```
JBWatenbergScality, line 28: in the current implementation `queryNodesResult `  can be null, see bellow my comment about the `enabled` option.

JBWatenbergScality (dismissed):

JBWatenbergScality (commented):
-

shell-ui/src/alerts/services/alertHooks.js
```diff
@@ -0,0 +1,40 @@
...
+  };
+};
+
+/**
+ *
+ * @param {FilterLabels} filterss
+ * @returns An array of alerts with the highest severity
+ */
+export const useHighestSeverityAlerts = (filters: FilterLabels): Alert[] => {
+  const query = useAlerts(useContext)(filters);
```
JBWatenbergScality, line 38: useContext should be injected

shell-ui/src/alerts/services/alertHooks.test.js
```diff
@@ -0,0 +1,112 @@
...
+  }),
+);
+
+describe('useHighestSeverityAlerts hook', () => {
+  jest.useFakeTimers();
+  beforeAll(() => server.listen({ onUnhandledRequest: 'error' }));
+  afterEach(() => server.resetHandlers());
+
+  const alertLibrary = window.shellUIAlerts[version];
+  console.log('window.shellUIAlerts', window.shellUIAlerts);
```
JBWatenbergScality: It could be removed

shell-ui/src/platform/service/k8s.js
ChengYanJin: Right, I've checked the doc, the `enabled` is a boolean, so maybe the following one is more correct
 ```options: { refetchInterval: 10000, enabled: token ? true : false },```

JBWatenbergScality: JS is automatically casting this to a boolean so  `token ? true : false` is strictly equivalent to `token`. Actually in `token ? true : false` you are also casting token as a boolean for the ternary condition

JBWatenbergScality:
> Cannot find module 'axios' from 'src/platform/service/k8s.js'

Axios is not available in shell-ui. I think we can simply use fetch as it is available in the browser we target

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+  token?: string | null,
+): typeof useQuery => {
+  if (!token) {
+    console.error('K8s API Not authenticated');
+    return;
+  }
+  return {
+    cacheKey: 'countNodes',
+    queryFn: () =>
+      fetch(`${k8sUrl}/api/v1/nodes`)
```
JBWatenbergScality: missing authorization header :
```
{
        headers: {
          authorization: `bearer ${token}`,
        },
      }
```

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,113 @@
...
+  if (!token) {
+    console.error('K8s API Not authenticated');
+    return;
+  }
+  return {
+    cacheKey: 'countVolumes',
+    queryFn: () =>
+      fetch(`${k8sUrl}/api/v1/persistentvolumes`, {
+        headers: {
+          authorisation: `bearer ${token}`,
```
JBWatenbergScality: it should be `authorization`

JBWatenbergScality (approved):

JBWatenbergScality:
I think you can merge this manually in my branch `improvement/shell-ui-alert-provider`

shell-ui/src/platform/service/k8s.js
```diff
@@ -0,0 +1,110 @@
...
+      })
+        .then((r) => {
+          if (r.ok) {
+            return r.json();
+          }
+        })
+        .then((res) => {
+          return res.items.length;
+        }),
+    options: { refetchInterval: 10000, enabled: token },
```
JBWatenbergScality: not sure about this options to be checked

shell-ui/src/platform/service/k8s.js
ChengYanJin: For some reason, the tests won't execute by `enabled: token`. So I have to switch back to `enabled: token ? true : false`

PR #3282, opened 2021-04-12
ci: Always check embedded MD5 in ISO

gdemonet:
Since #3032 (merged in 2.8), all our ISOs include an embedded MD5
checksum. We had an empirical check to know whether this checksum was
indeed present, since we also need to verify ISOs from (N-1) when
running upgrade/downgrade tests from version (N).

Now, in 2.9, all ISOs we expect should include this checksum. The
empirical check is no longer needed, and we remove it.
We also add a common `requirements.sh` to install the required package
for openstack workers which were only used to spawn other VMs through
Terraform, since these workers are responsible for validating the
integrity of retrieved artifacts.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -103,24 +103,10 @@ models:
             ssh eve@$IP
         END
   - ShellCommand: &check_iso_checksum
-      name: Check image with checksum
+      name: Check image with checksums
       command: |
         sha256sum -c SHA256SUM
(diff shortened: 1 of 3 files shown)


gdemonet:
Checksum validation works in both pre-merge and post-merge stages, I think we're good to go.

NicolasT (approved):

PR #3289, opened 2021-04-14
eve: Add label role node when untaint node

TeddyAndrieux:
Add the role label "node" to the node when untaint so that if we re-run
bootstrap we do not add taints again

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -671,7 +671,9 @@ models:
       command: |
         ssh -F ssh_config bootstrap <<ENDSSH
         sudo kubectl --kubeconfig /etc/kubernetes/admin.conf \
-        patch node "$NODE_NAME" --patch '{"spec": {"taints": []}}'
+        patch node "$NODE_NAME" --patch '{"spec": {"taints": []}}' &&
+        sudo kubectl --kubeconfig /etc/kubernetes/admin.conf \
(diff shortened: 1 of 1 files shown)


eve/main.yml
```diff
@@ -671,7 +671,9 @@ models:
       command: |
         ssh -F ssh_config bootstrap <<ENDSSH
         sudo kubectl --kubeconfig /etc/kubernetes/admin.conf \
-        patch node "$NODE_NAME" --patch '{"spec": {"taints": []}}'
+        patch node "$NODE_NAME" --patch '{"spec": {"taints": []}}' &&
+        sudo kubectl --kubeconfig /etc/kubernetes/admin.conf \
+        label node --overwrite  "$NODE_NAME" "node-role.kubernetes.io/node="
```
gdemonet, line 676: Why not do this within the patch itself?

```suggestion
        patch node "$NODE_NAME" --patch '{
          "metadata": {"labels": {"node-role.kubernetes.io/node": ""}},
          "spec": {"taints": []}
        }'
```

gdemonet (approved):
(still approving, since I believe this is really minor) :shipit:

eve/main.yml
TeddyAndrieux: Right, I could have done this in a single patch, does not really matter :smile:

PR #3299, opened 2021-04-19
salt: Add dedicated Ingress for MetalK8s docs

gdemonet:
Adding a `rewrite-target` annotation is needed when the base path is not
/, because NGINX does not know where to find the documentation static
files.

We now always expose MetalK8s docs behind /docs, optionally appending
the configured base path for MetalK8s UI (e.g. /docs/platform).
This will avoid strange conflicts with the behaviour of the current
Ingress when running MetalK8s at a non-root path, because mixing regular
expressions and exact matches does not play well with ingress-nginx.

See: #3254
See: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#use-regex

### DIFF
--- a/salt/metalk8s/addons/ui/deployed/ingress.sls
+++ b/salt/metalk8s/addons/ui/deployed/ingress.sls
@@ -108,3 +108,28 @@ spec:
           serviceName: metalk8s-ui
           servicePort: 80
 {% endfor %}
+---
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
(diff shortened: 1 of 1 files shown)


gdemonet:
Ideally, we would add some simple end-to-end tests checking that all exposed services continue working as desired when configuring the UI base path. We will need to postpone this task to a later improvement however, due to pressing matters.

JBWatenbergScality (approved):

PR #3308, opened 2021-04-22
Feature/3266 apply error pages

### LINKED ISSUE #3266: Apply 4XX & 5XX page in MetalK8s UI
**Component**: UI

**Why this is needed**:
Following the implementation of 4XX & 5XX https://github.com/scality/core-ui/issues/245
We should display the correspondent page when it's needed.

**What should be done**:
List the cases/roue that should display the 4XX & 5XX page:

Implement a wrapper to retrieve the current language from the navbar to set the lang prop in 4XX/5XX. This wrapper would potentially be needed for some other components as well.

4XX  Page: The path in the URL does not exist.

**Test plan**:
Add navigation test if 4XX & 5XX page are displayed correctly.

alexis-ld:
**Component**: ui, errorpages

**Context**:  #3266 
We want to add error pages to metalK8S.

**Summary**:
This adds :
* a 404 page to undefined routes.
* a 500 page when the config file is not found/valid
* a 500 page on Navbar fail
* an Auth error page when the user is not authenticated

---

Closes: #3266

### DIFF
--- a/ui/cypress/fixtures/config.json
+++ b/ui/cypress/fixtures/config.json
@@ -1,12 +1,13 @@
 {
-    "url": "/api/kubernetes",
-    "url_salt": "/api/salt",
-    "url_prometheus": "/api/prometheus",
-    "url_grafana": "/grafana",
-    "url_doc": "/docs",
-    "url_alertmanager": "/api/alertmanager",
-    "url_navbar": "/shell/solution-ui-navbar.1.0.0.js",
-    "url_alerts": "/shell/alerts.1.0.0.js",
-    "alerts_lib_version": "1.0.0",
-    "url_navbar_config": "/shell/config.json"
+  "url": "/api/kubernetes",
+  "url_salt": "/api/salt",
+  "url_prometheus": "/api/prometheus",
+  "url_grafana": "/grafana",
+  "url_doc": "/docs",
+  "url_alertmanager": "/api/alertmanager",
+  "url_navbar": "/shell/solution-ui-navbar.1.0.0.js",
+  "url_alerts": "/shell/alerts.1.0.0.js",
+  "alerts_lib_version": "1.0.0",
+  "url_navbar_config": "/shell/config.json",
+  "url_support": "https://support.com"
 }
--- a/ui/cypress/integration/error-pages.spec.js
+++ b/ui/cypress/integration/error-pages.spec.js
@@ -1,0 +1,95 @@
+describe('Error Pages Navigation', () => {
+  beforeEach(() => {
+    cy.setupMocks();
+    cy.login();
+  });
+
+  it('redirects me to 404 error page if route not found', () => {
+    cy.visit('/undefined-route-that-doesnt-exists');
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page404"]').contains(
+      'Error: this page does not exist',
+    );
+  });
+});
+
+describe('Error Pages Custom config', () => {
+  beforeEach(() => {
+    cy.setupMocks(null);
+    cy.login();
+  });
+
+  it('redirects me to 500 - fetch config results in internal error', () => {
+    cy.route2('GET', '/config.json', {
+      fixture: 'config.json',
+      statusCode: 500,
+    });
+    cy.visit('/');
+    cy.stubHistory();
+    cy.get('[data-cy="sc-error-page500"]').contains('Unexpected Error');
+  });
+
+  it('redirects me to 500 in case of wrong url_alerts config', () => {
+    let url_support = 'https://mysupportlink.com';
+    cy.fixture('config.json').then((config) => {
+      config.url_support = url_support;
+      // must be a fake url
+      // otherwise in CI nginx which is served with a try files rule, will return a html with successful response code
+      config.url_alerts = 'http://invalid/url';
+      cy.route2('GET', '/config.json', config);
+    });
+
+    // alerts page will throw an error if it has failed
+    // we must catch it on cypress
+    cy.on('uncaught:exception', () => false);
+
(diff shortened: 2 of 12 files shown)


ui/cypress/integration/error-pages.spec.js
```diff
@@ -0,0 +1,100 @@
...
+    cy.setupMocks();
+    cy.login();
+  });
+
+  it('brings me to 404 error page', () => {
+    cy.visit('/404');
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page404"]').should('have.length', 1);
+  });
```
ChengYanJin: IMO, can be removed since it's the same case as the next one.

ui/src/components/Navbar.js
```diff
@@ -187,19 +188,9 @@ function useLogoutEffect(
...
-      <p>Something went wrong:</p>
-      <pre>{error.message}</pre>
-    </div>
-  );
-}
-
 export function Navbar() {
   return (
-    <ErrorBoundary FallbackComponent={ErrorFallback}>
+    <ErrorBoundary fallbackRender={() => <Redirect to='/500'/>}>
```
ChengYanJin: We should directly render the `ErrorPage` instead of having route `/500` and redirecting to it.

ui/src/containers/Layout.js
```diff
@@ -160,6 +167,25 @@ const Layout = () => {
...
+                  exact
+                  path="/500"
+                  component={() => (
+                    <ErrorPage500
+                      data-cy="sc-error-page500"
+                      supportLink={url_support}
+                      locale={language}
+                    />
+                  )}
+                />
```
ChengYanJin: Should be removed

ui/cypress/integration/error-pages.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+  it('brings me to 500 error page', () => {
+    cy.visit('/500');
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page500"]').should('have.length', 1);
+  });
```
ChengYanJin: Should be removed since we don't need to have `/500` route

ui/cypress/integration/error-pages.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page404"]').should('have.length', 1);
+  });
+
+  it('redirects me to 404 error page if route not found', () => {
+    cy.visit('/undefined-route-that-doesnt-exists');
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page404"]').should('have.length', 1);
```
ChengYanJin: IMO, we should test if the text on the 404 page render correctly because that's what the users see and really matter. 

Same for all the error pages

ui/cypress/integration/error-pages.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    cy.visit('/500');
+    cy.stubHistory();
+
+    cy.get('[data-cy="sc-error-page500"]').should('have.length', 1);
+    cy.get('.sc-error-page500 * a')
+      .should('have.attr', 'href')
+      .and('include', url_support);
+  });
+});
```
ChengYanJin: For me, this should not be an independent scenario to test, should be an assertion in a scenario. For example, when we test the navbar failing scenario, we can check if the support link is rendered on the 500 error page.

ChengYanJin (changes requested):

ui/cypress/integration/error-pages.spec.js
alexis-ld: @ChengYanJin do you mean testing if the 404 text is rendered properly in there ?

I think that may be a bad idea because this text is directly into core-ui ErrorPage component, meaning if we update the text in core-ui it would make the metal tests fail.

ChengYanJin: I think we don't need to check the actual text, but add an `aria-label` for the text element in core-ui

ui/cypress/fixtures/config.json
```diff
@@ -1,12 +1,13 @@
...
+  "url_salt": "/api/salt",
+  "url_prometheus": "/api/prometheus",
+  "url_grafana": "/grafana",
+  "url_doc": "/docs",
+  "url_alertmanager": "/api/alertmanager",
+  "url_navbar": "/shell/solution-ui-navbar.1.0.0.js",
+  "url_alerts": "/shell/alerts.1.0.0.js",
+  "alerts_lib_version": "1.0.0",
+  "url_navbar_config": "/shell/config.json",
+  "url_support": "https://support.scality.com/hc/en-us"
```
ChengYanJin: don't see why we need a real link here. 
And also please add an empty entry in metalk8s UI `config.json`
`"url_support": ""`

gdemonet: Actually, if we could have something like `https://github.com/scality/metalk8s/discussions/new`, that'd be awesome! (not in the test, in the default config used when deploying MetalK8s)

ui/public/config.json
```diff
@@ -10,5 +10,5 @@
   "url_alerts": "/shell/alerts.1.0.0.js",
   "url_navbar_config": "/shell/config.json",
   "alerts_lib_version": "1.0.0",
-  "url_support": ""
+  "url_support": "https://github.com/scality/metalk8s/discussions/new"
```
gdemonet, line 13: Thanks! :+1:

ChengYanJin (dismissed):

ChengYanJin (dismissed):

ChengYanJin (approved):

PR #3310, opened 2021-04-22
ui: Remove sorting sync with url for the table in volume tab

### LINKED ISSUE #3307: UI: Sorting in Node page is not working
**Component**: UI

**What happened**:
The sorting of the volume table in `Node page` is not working.. 

**What was expected**:

**Steps to reproduce**

ChengYanJin:
**Component**: UI

**Context**: 
Given we have two tables with the ability to sort at the same page,
and we use the same query parameter `sort` for both tables, cause a conflict.

We decided to keep the sorting of the table in the volume tab, but it won't
sync with URL.

**Summary**:

**Acceptance criteria**: 

---

Closes: #3307

### DIFF
--- a/ui/src/services/utils.js
+++ b/ui/src/services/utils.js
@@ -375,11 +375,15 @@ export const formatSizeForDisplay = (value) => {
   return value.replace(/^(\d+)(\D+)$/, '$1 $2');
 };
 
-/*
- ** Custom hook that stores table sorting choice in the URL queries
- ** Defaults to health sorting (used on Nodes and Volumes tables)
+/**
+ * Custom hook that stores table sorting choice in the URL queries
+ *
+ * @param {string} sorted
+ * @param {boolean} desc
+ * @param {array} data
+ * @param {string} defaultSortKey default sorting key
  */
-export const useTableSortURLSync = (sorted, desc, data) => {
+export const useTableSortURLSync = (sorted, desc, data, defaultSortKey) => {
   const history = useHistory();
   const location = useLocation();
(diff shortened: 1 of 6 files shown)


ui/src/services/utils.js
```diff
@@ -375,11 +375,16 @@ export const formatSizeForDisplay = (value) => {
   return value.replace(/^(\d+)(\D+)$/, '$1 $2');
 };
 
-/*
- ** Custom hook that stores table sorting choice in the URL queries
- ** Defaults to health sorting (used on Nodes and Volumes tables)
+/**
+ * Custom hook that stores table sorting choice in the URL queries
+ * Defaults to health sorting (used on Nodes and Volumes tables)
```
gdemonet: This comment is not applicable anymore, since we now consider a single default sorting key, which is passed as an  argument to this hook.

ui/src/services/utils.js
```diff
@@ -375,11 +375,16 @@ export const formatSizeForDisplay = (value) => {
...
+ *
+ * @param {string} sorted
+ * @param {boolean} desc
+ * @param {array} data
+ * @param {string} key default sorting value
```
gdemonet: To make it obvious, renaming this argument to `defaultSortKey` or `defaultKey` would be nice. Also, the comment should likely say "default sorting key", not "value".

gdemonet (changes requested):
minor changes, but behaviour LGTM (not tested though)

gdemonet (approved):

PR #3312, opened 2021-04-22
csc, ui: Move navbar entries to user configuration

### LINKED ISSUE #3306: Do not store Shell UI navbar menus in the static defaults
**Component**: shell-ui

**Why this is needed**:

When updating the ShellUIConfig CSC, removing default menu entries is not possible, because of how we merge defaults with overrides.

**What should be done**:

Define default entries as part of the initial values in the overrides.

gdemonet:
These menu entries were previously configured from a merge between
static defaults and user overrides.
This doesn't play nice when a user wants to remove an entry.
To handle this properly, we now ensure the values are populated in the
initial CSC ConfigMap for ShellUIConfig.

Fixes: #3306

### DIFF
--- a/salt/metalk8s/addons/ui/deployed/ui-configuration.sls
+++ b/salt/metalk8s/addons/ui/deployed/ui-configuration.sls
@@ -41,6 +41,23 @@ metalk8s-ui-config ConfigMap already exist:
 {%- endif %}
 
 {%- if metalk8s_shell_ui_config is none %}
+  {%- set metalk8s_ui_defaults = salt.slsutil.renderer(
+          'salt://metalk8s/addons/ui/config/metalk8s-ui-config.yaml',
+          saltenv=saltenv
+      )
+  %}
+  {%- if metalk8s_ui_config is none %}
+    {%- set metalk8s_ui = metalk8s_ui_defaults %}
+  {%- else %}
+    {%- set metalk8s_ui = salt.metalk8s_service_configuration.get_service_conf(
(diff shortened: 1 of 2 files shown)


salt/metalk8s/addons/ui/deployed/ui-configuration.sls
```diff
@@ -54,7 +71,22 @@ Create metalk8s-shell-ui-config ConfigMap:
           config.yaml: |-
             apiVersion: addons.metalk8s.scality.com/v1alpha1
             kind: ShellUIConfig
-            spec: {}
```
JBWatenbergScality, line 57: I think we also need the user group mapping here else we can't remove a group as well

gdemonet: Debatable, I like that we can always assume all static users in Dex config will have the `metalk8s:admin` role (same as in Grafana config). If that's user-editable, then we can't update it when we change Dex config.

JBWatenbergScality (dismissed):

gdemonet:
Build consistently failing on Cypress tests, I need to investigate.

gdemonet:
Got it:

```
Error: [navbar][config] Invalid path specified in "options.main": "https://10.200.2.119:8443alerts" (keys must be defined as fully qualified URLs, such as "{protocol}://{host}{path}?{queryParams}")
```

Got mixed up when normalizing the paths :facepalm:

JBWatenbergScality (approved):

PR #3318, opened 2021-04-23
salt: Allow to provide featureGates for apiserver in bootstrap config

### LINKED ISSUE #3294: Allow configuration of `kube-apiserver` feature gates in BootstrapConfiguration
_TBD_

TeddyAndrieux:
Add ability to provide some feature gates for kube apiserver in
bootstrap config that will be persisted after upgrade/salt highstate ...

---

Fixes: #3294

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   Alertmanager are now stored in Loki database for persistence
   (PR[#3191](https://github.com/scality/metalk8s/pull/3191))
 
+- [#3294](https://github.com/scality/metalk8s/issues/3294) - Allow to manage
+  `kube-apiserver` feature gates from Bootstrap Configuration file
+  (PR[#3318](https://github.com/scality/metalk8s/pull/3318))
+
 ### Enhancements
 - Bump Kubernetes version to 1.20.6
   (PR[#3311](https://github.com/scality/metalk8s/pull/3311))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ### Features Added
 - [#3180](https://github.com/scality/metalk8s/issues/3180) - All alerts from
   Alertmanager are now stored in Loki database for persistence
-  (PR[#3191](https://github.com/scality/metalk8s/pull/))
+  (PR[#3191](https://github.com/scality/metalk8s/pull/3191))
 
 ### Enhancements
 - Bump Kubernetes version to 1.20.6
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -170,6 +170,7 @@ def ext_pillar(minion_id, pillar, bootstrap_config):  # pylint: disable=unused-a
             "networks": _load_networks(config),
             "metalk8s": metal_data,
             "proxies": config.get("proxies", {}),
+            "kubernetes": config.get("kubernetes", {}),
         }
 
         if not isinstance(metal_data["archives"], list):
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
 ### Features Added
 - [#3180](https://github.com/scality/metalk8s/issues/3180) - All alerts from
   Alertmanager are now stored in Loki database for persistence
-  (PR[#3191](https://github.com/scality/metalk8s/pull/))
+  (PR[#3191](https://github.com/scality/metalk8s/pull/3191))
```
alexandre-allard, line 7: Thanks

alexandre-allard (dismissed):

CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
...
+
+- [#3294](https://github.com/scality/metalk8s/issues/3294) - Allow to manage
+  Kube Api Server feature gates from Bootstrap Configuration file
```
```suggestion
  `kube-apiserver` feature gates from Bootstrap Configuration file
```

salt/_pillar/metalk8s.py
```diff
@@ -170,6 +170,7 @@ def ext_pillar(minion_id, pillar, bootstrap_config):  # pylint: disable=unused-a
             "networks": _load_networks(config),
             "metalk8s": metal_data,
             "proxies": config.get("proxies", {}),
+            "kubeApiServer": config.get("kubeApiServer", {}),
```
gdemonet: Can we do something more future-proof, like:

```yaml
kind: BootstrapConfiguration
# [...]
kubernetes:
  apiServer:
    featureGates:
      ephemeralContainers: "true"
```

So we can add fields easily to `kubernetes` and keep it well-contained (also set default values, if needed, and share values across core K8s components, if that's something).

gdemonet (changes requested):

salt/_pillar/metalk8s.py
TeddyAndrieux: Sure I will

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -26,6 +26,11 @@ include:
 {%- endif %}
 {%- set etcd_servers = etcd_servers | unique %}
 
+{%- set feature_gates = [] %}
+{%- for feature, value in pillar.kubeApiServer.get("featureGates", {}).items() %}
+{%-   do feature_gates.append(feature ~ "=" ~ value) %}
```
NicolasT, line 31: This code now expects the values to be strings (which they should be). However, one could easily put something like
```
kubernetes:
  apiServer:
    featureGates:
      ephemeralContainers: true
```
in the configuration file, which turns the value into a bool instead of a string, which would be stringified in Python to `True` (I believe?), and `kube-apiserver` may not like that as an argument to `--feature-gates=...`.

So, we should either cater for such user 'mistake', or bail out.

NicolasT (changes requested):

salt/_pillar/metalk8s.py
TeddyAndrieux: done

salt/metalk8s/kubernetes/apiserver/installed.sls
TeddyAndrieux: Oh wait, but it works even if it's "True" at the end

NicolasT:
This likely needs a documentation entry as well.

TeddyAndrieux:
> This likely needs a documentation entry as well.

Do we really want to document it ?

salt/metalk8s/kubernetes/apiserver/installed.sls
TeddyAndrieux: Anyway ok I will convert bool to "true" or "false"

NicolasT:
> > This likely needs a documentation entry as well.
> 
> Do we really want to document it ?

Yes since it's a feature. At least in the open source docs, maybe not in the product ones.

PR #3324, opened 2021-04-25
ui: Fix a race condition leading to an infinite authentication loop on firefox

JBWatenbergScality:
**Component**:

ui

**Context**: 

On firefox a render race condition post authentication was leading to an infinite authentication loop.

**Summary**:

The loop was due to an intermediary render of the navbar before the user being pushed to the redux store resulting in the logout effect being executed with an isAuthenticated flag without user which triggers a logout loop reloading the page, rendering the navbar which trigger a new auth loop and so on...

**Acceptance criteria**: 

The authentication should now work on firefox.

Note for later: we should execute our end to end tests against all the browser that we want to support.


ChengYanJin (approved):
LGTM! Thanks for the fix!

PR #3334, opened 2021-04-28
Bugfix/navbar entries order

JBWatenbergScality:
**Component**:

UI

**Context**: 

We want to provide a way for users to define a specific order of appearance for the navbar entries.

**Summary**:

Adding an optional `order` property to define the order of entries.

**Acceptance criteria**: 

Unit tests

### DIFF
--- a/shell-ui/src/navbar/NavBar.js
+++ b/shell-ui/src/navbar/NavBar.js
@@ -47,33 +47,30 @@ const translateOptionsToMenu = (
   userGroups: string[],
 ) => {
   const normalizedLocation = normalizePath(location.href);
-  return (
-    Object.entries(options[section])
-      //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, PathDescription]
-      .filter((entry: [string, PathDescription]) =>
-        isEntryAccessibleByTheUser(entry, userGroups),
-      )
-      .map(
-        //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, PathDescription]
-        ([path, pathDescription]: [string, PathDescription], i) => {
-          try {
-            return {
-              ...renderer(path, pathDescription),
-              selected: pathDescription.activeIfMatches
-                ? new RegExp(pathDescription.activeIfMatches).test(
-                    location.href,
-                  )
-                : normalizedLocation === normalizePath(path),
-            };
-          } catch (e) {
-            throw new Error(
-              `[navbar][config] Invalid path specified in "options.${section}": "${path}" ` +
-                '(keys must be defined as fully qualified URLs, ' +
-                'such as "{protocol}://{host}{path}?{queryParams}")',
-            );
-          }
-        },
-      )
+  const entries = Object.entries(options[section])
+    //$FlowIssue - flow typing for Object.entries incorrectly typing values as [string, mixed] instead of [string, PathDescription]
+    .filter((entry: [string, PathDescription]) =>
+      isEntryAccessibleByTheUser(entry, userGroups),
+    );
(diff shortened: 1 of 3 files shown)


shell-ui/src/navbar/NavBar.js
```diff
@@ -53,6 +53,7 @@ const translateOptionsToMenu = (
       .filter((entry: [string, PathDescription]) =>
         isEntryAccessibleByTheUser(entry, userGroups),
       )
+      .sort((entryA, entryB) => entryA.order < entryB ? 1 : -1)
```
alexanderchan-scality:
```suggestion
      .sort((entryA, entryB) => entryA.order < entryB.order ? 1 : -1)
```

shell-ui/src/navbar/NavBar.js
```diff
@@ -47,33 +47,30 @@ const translateOptionsToMenu = (
...
+          selected: pathDescription.activeIfMatches
+            ? new RegExp(pathDescription.activeIfMatches).test(location.href)
+            : normalizedLocation === normalizePath(path),
+        };
+      } catch (e) {
+        throw new Error(
+          `[navbar][config] Invalid path specified in "options.${section}": "${path}" ` +
+            '(keys must be defined as fully qualified URLs, ' +
+            'such as "{protocol}://{host}{path}?{queryParams}")',
+        );
```
gdemonet, line 71: Now realizing this `try`/`catch` is likely too wide for the error message it raises... Let's tackle this as debt.

shell-ui/src/navbar/Navbar.spec.js
```diff
@@ -335,5 +335,82 @@ describe('navbar', () => {
...
+      />,
+    );
+    //E
+    await waitForLoadingToFinish();
+    //V
+    const entries = [...screen.queryByText(/First/i).parentElement.parentElement.children].map(e => e.querySelector('a').innerHTML)
+    expect(entries).toStrictEqual(['First', 'Second', 'Third'])
+  });
+
+  it('should enqueue unordered entries', async () => {
```
gdemonet, line 377: How can this work with:

```javascript
entries.sort(([_, entryA], [__, entryB]) => (entryA.order - entryB.order));
```

??? :thinking: Javascript is really a mistery to me...

gdemonet (approved):

shell-ui/src/navbar/Navbar.spec.js
ChengYanJin: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

> The order of the array returned by Object.entries() does not depend on how an object is defined. If there is a need for certain ordering, then the array should be sorted first, like Object.entries(obj).sort((a, b) => b[0].localeCompare(a[0]));.

gdemonet: Oh, that was not my question actually @ChengYanJin (but it confirms that we should **really** not rely on default ordering).

My point was that we are sorting `entries` based on `entryA.order - entryB.order`... But we are able to handle entries for which `entryC.order === undefined`, and somehow they end up at the end (which this test is verifying). This part is really mystical to me.

gdemonet: OK, tested it a bit (see https://jsfiddle.net/Loxm9ck4/2/), it's actually broken, the `sort` logic just doesn't move elements when faced with such a case (the comparison function returns `NaN`). So I think your test is somewhat broken @JBWatenbergScality :confused:

JBWatenbergScality: (Typically something that would have been caught immediately with typescript)

Issue #3328, opened 2021-04-27
Bug in display when adding a new disk with long labels


lavigne958:
**Component**:

UI

**What happened**:
I clicked on `Volumes` => `+ create volume` => added a label that is longer than expected.
The label key is overlapping on the label value which makes it hard to read.

**What was expected**:
The list of labels is nicely displayed 

**Steps to reproduce**
Add a new volume and add very long label key

**Resolution proposal** (optional):

![image (2)](https://user-images.githubusercontent.com/15084254/116262363-c3556600-a778-11eb-829e-4f0ada404510.png)

ChengYanJin:
merged in #3338

Issue #3203, opened 2021-03-16
UI: Validate labels names for volumes


JBWatenbergScality:
**Component**:

UI

**What happened**:

When creating a volume using specials characters in the UI is authorised for labels but denied when submitting the query to kubernetes API.

Kubernetes API answered with this message : `Volume.storage.metalk8s.scality.com "test" is invalid: metadata.labels: Invalid value: "'world'": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')` but from the UI point of view we just see a generic error :

![image](https://user-images.githubusercontent.com/75977494/111287647-b86cc900-8643-11eb-8d55-784a0c45a0bb.png)

**What was expected**:

Field should be validated when they are being filled in or we should display the error message so that the user can understand that the issue comes from invalid characters.

**Steps to reproduce**

Create a volume and use a `'` as a label name or value.

ChengYanJin:
merged in #3338

PR #3338, opened 2021-04-30
ui: Bugfix several bug fixes around cluster version create volume 

### LINKED ISSUE #3303: ui: Volume creation : button add for labels should change from disable to enable status when a label value and key is provided
**Component**:

UI

**Why this is needed**:

From a ux point of view it is not enough evident that we have to click `add` button when having entered the label, leading to human error creating the volume without the label despite having filled it in.

image

**What should be done**:

We should change the status of the button once having provided a label key and value so that the user clearly see the required action.

**Test plan**:

Unit test

ChengYanJin:
**Component**: ui

**Context**: 
Bugfix:
1. When the size is `Unknown` no need to add a unit
2. Add the color of `create` button in create node & volume page to `buttonPrimary`
3. Fix the cluster version in create node page
4. Add about entry from the Navbar
5. Fix the display bug of labels in crate volume page
6. Add default `ui_base_path` in `config.json`
7.  Translations 
8. After deleting a volume, redirect to the first volume - overview tab
9. Fix the error when a new node is not ready, we can receive the string from ipInterfaceObject

Improvement:
10. Merge the label validation before creating a new volume and **improve the UX of creating volume**
- when the label key and value are valid, the `Add` button will become enabled, otherwise stay disable.

image

**Summary**:

**Acceptance criteria**: 

---

Closes: #3303  
Closes: #3328
Closes: #3203


ui/src/containers/CreateVolume.js
```diff
@@ -568,14 +570,25 @@ const CreateVolume = (props) => {
...
+                              variant={'buttonSecondary'}
+                            />
+                          ) : (
+                            <Button
+                              text={intl.translate('add')}
+                              type="button"
+                              onClick={addLabel}
+                              data-cy="add-volume-labels-button"
+                              outlined
+                            />
```
JBWatenbergScality: shouldn't we disable the button initially instead of just changing it's color ? Also I think we can remove the onClick handler in this case

ui/src/services/salt/api.js
```diff
@@ -68,10 +68,36 @@ export type IPInterfaces = {
...
+            "127.0.0.1",
+            "::1"
+          ],
+          "eth0": [
+            "10.200.6.201",
+            "fe80::f816:3eff:fec3:b710"
+          ]
+        }
+      },
+      "toto": "Minion did not return. [No response]\nThe minions may not have all finished running and any remaining minions will return upon completion. To look up the return data for this job later, run the following command:\n\nsalt-run jobs.lookup_jid 20210429184803777520"
```
JBWatenbergScality:
```suggestion
      "nodeName": "Minion did not return. [No response]\nThe minions may not have all finished running and any remaining minions will return upon completion. To look up the return data for this job later, run the following command:\n\nsalt-run jobs.lookup_jid 20210429184803777520"
```

ui/src/containers/CreateVolume.js
ChengYanJin: Done in https://github.com/scality/metalk8s/pull/3338/commits/7a961e20d30abe2e3ea76d4d7050fbe7bc9464c4 ;)

JBWatenbergScality (dismissed):

ui/src/containers/CreateVolume.js
```diff
@@ -435,13 +453,64 @@ const CreateVolume = (props) => {
                   }
                 }
               };
+              // TODO: Apply this function as `labelName` field validation function, for the moment we only have one error message.
```
JBWatenbergScality: Is this todo still valid ? I see various error message in the implementation.

ChengYanJin: OK. So in the end I switched to `Field-level Validation` instead of `Yup` schema which will display the error message separately for the `label prefix` and `labe name value`.
Please see: https://github.com/scality/metalk8s/pull/3338/commits/9917a17a96db87f73451d6d13ff38e6b7db8de77

ui/src/containers/CreateVolume.js
```diff
@@ -453,64 +455,40 @@ const CreateVolume = (props) => {
...
-              // TODO: Apply this function as `labelName` field validation function, for the moment we only have one error message.
-              const handleErrorLabel = () => {
-                const isLabelNameMatched = labelFullNameRegex.test(
-                  values.labelName,
-                );
-                const isLabelValueMatched = labelValueRegex.test(
-                  values.labelValue,
-                );
 
+              const validataLabelName = (value) => {
```
JBWatenbergScality, line 459: validateLabelName

ui/cypress/fixtures/kubernetes/storageclasses.json
```diff
@@ -0,0 +1,207 @@
+{
+  "apiVersion": "storage.k8s.io/v1",
+  "items": [
```
JBWatenbergScality, line 3: I would limit the storage classes only to the one that makes sense for our tests

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,70 @@
...
+
+// interaction test for creating volume
+describe('the create button in create volume page', () => {
+  it('should be enabled and the Add button should be disabled when no labels is specified', () => {
+    cy.get('input[name=name]').type(VOLUME_NAME);
+
+    cy.get('.sc-select').eq(0).click();
+    cy.get('.sc-select__menu').eq(0).click();
+    cy.get('.sc-select').eq(1).click();
+    cy.get('.sc-select__menu')
```
JBWatenbergScality: We should avoid dom selectors by classnames and tests this as user interact with. With the current implementation the most elegant solution is to look for elements with text matching /node*/i and click on the next `.sc-select`. Ideally we should add an aria role with listbox value on the dropdown component https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role + an aria-label with value "node"

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,70 @@
...
+      .find(`[data-cy=storageClass-${STORAGECLASS}]`)
```
JBWatenbergScality: We should select the element via the text presented to the user

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,70 @@
...
+      .click();
+    cy.get('.sc-select').eq(2).click();
+    cy.get('.sc-select__menu').find(`[data-cy="type-${VOLUME_TYPE}"]`).click();
+    cy.get('input[name=sizeInput]').type(VOLUME_SIZE);
+
+    cy.get('[data-cy="submit-create-volume"]').should('be.enabled');
```
JBWatenbergScality: use cy.queryByRole('button', {name: /create/i})

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    cy.get('[data-cy="add-volume-labels-button"]').should('be.disabled');
```
JBWatenbergScality: use cy.queryByRole('button', {name: /add/i})

ui/cypress/fixtures/kubernetes/storageclasses.json
ChengYanJin: OK. I was hesitating should I keep only one storage class. I will do so then.

ui/cypress/integration/volume/createvolume.spec.js
ChengYanJin: Just FYI in case you miss it :)
https://github.com/testing-library/cypress-testing-library/releases/tag/v6.0.0
so should be: `  cy.findByRole('button', { name: /create/i })`

ui/cypress/integration/volume/createvolume.spec.js
ChengYanJin: In the end, we got to keep the `data-cy`, the reason is:

>   /* we use `data-cy`is because the storageClass may display in both the select list and selected item, 
>      so we use unique selector to make sure to choose the item in the list. */

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+// interaction test for creating volume
+describe('the create button in create volume page', () => {
+  it('should be enabled and the Add button should be disabled when no labels is specified', () => {
+    cy.findByRole('button', { name: /create/i }).should('be.enabled');
+    cy.findByRole('button', { name: /add/i }).should('be.disabled');
+  });
+
+  it('should be disabled before adding the valid label', () => {
+    cy.get('input[name=labelName]').type(VOLUME_LABEL_NAME);
```
JBWatenbergScality: it could be replaced with findByRole('textbox', {name: /Enter label name/i})

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    cy.get('input[name=labelValue]').type(VOLUME_LABEL_VALUE);
```
JBWatenbergScality: it could be replaced with findByRole('textbox', {name: /Enter label value/i})

ui/cypress/integration/volume/createvolume.spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+    cy.findByRole('button', { name: /create/i }).should('be.disabled');
+    cy.findByRole('button', { name: /add/i }).should('be.enabled');
+  });
+
+  it('should be disabled as well as the Add button if labelName is not valid', () => {
+    cy.get('input[name=labelName]').type(VOLUME_LABEL_INVALID);
+    cy.get('input[name=labelValue]').type(VOLUME_LABEL_VALUE);
```
JBWatenbergScality: same here

ui/cypress/integration/volume/createvolume.spec.js
ChengYanJin: OK. Maybe` cy.findByPlaceholderText(/Enter label name/i).` would be even better!

JBWatenbergScality (approved):

PR #3346, opened 2021-05-03
docs: Add `release` label to prometheusRule example

### LINKED ISSUE #3293: [doc] PrometheusRules require specific labels
**Component**:

Documentation - Prometheus

**Why this is needed**:
Without it the alert rules are not taken by prometheus operator

**What should be done**:
Add documentation about these required labels.

> In order for the new alerting rule to be added by the prometheus operator the following labels must be set with the following values:

```yaml
// Required for prometheus discovery
- app:                    "prometheus-operator"
- app.kubernetes.io/name: "prometheus-operator"
- role:                   "alert-rules"
```

**Test plan**:
Add new alerting rule and ommit one of the above labels.

alexandre-allard:
**Component**: docs

**Context**: 
In the documentation we were missing the `release` label for PrometheusRules, so rules were not taken into account by Prometheus.

**Summary**:
Add the label in the example and a note to explain the labels are mandatory.

**Acceptance criteria**: n/a

---

Closes: #3293

### DIFF
--- a/docs/operation/cluster_and_service_configuration.rst
+++ b/docs/operation/cluster_and_service_configuration.rst
@@ -379,6 +379,7 @@ In order to add Alert rules, a new ``PrometheusRule`` manifest must be created.
      labels:
        app: prometheus-operator
        app.kubernetes.io/name: prometheus-operator
+       release: prometheus-operator
      name: <prometheus-rule-name>
      namespace: <namespace-name>
    spec:
@@ -393,6 +394,11 @@ In order to add Alert rules, a new ``PrometheusRule`` manifest must be created.
          for: 1h
          labels:
            severity: warning
+
+.. note::
+
+   All the labels in the example above are mandatory for Prometheus
+   to take the new rules into account.
 
 Then this manifest must be applied.
 


gdemonet (approved):

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -394,6 +395,11 @@ In order to add Alert rules, a new ``PrometheusRule`` manifest must be created.
          labels:
            severity: warning
 
+.. note::
+
+   All the labels in the example above are mandatory for Prometheus
+   to take the new rules into account.
```
NicolasT, line 401: That seems like an issue, no? The app deploys the rule, so `app.kubernetes.io/name` and `app` must be related to the app *creating* the rule, totally unrelated to `prometheus-operator`. Similarly, `release` should not be part of any (required) set of labels, and should not be `prometheus-release`.

In essence, we need to *change* the set of labels our `prometheus-operator` uses to filter out rules (and other objects) it should take into account, e.g., turn the selector into `metalk8s.scality.com/monitor: "true"` or something along those lines.

alexandre-allard: Can easily be overridden using `prometheus.prometheusSpec.ruleSelector` in the `charts/kube-prometheus-stack.yaml` file:
```
prometheus:
  prometheusSpec:
    ruleSelector:
      matchLabels:
        metalk8s.scality.com/monitor: "true"
```
And we should probably do the same for ServiceMonitor.

gdemonet: Indeed, that's what we recommend in #2199
This PR was more about making sure our docs are accurate about the current state of the product (although it's debatable whether it's a useful change :stuck_out_tongue_winking_eye:)

alexandre-allard: Oneline fix, does not cost much, at least now it's accurate even if it's ugly (which was already the case).

PR #3350, opened 2021-05-04
ui: Remove the fetchalert action & Fix the warning from console

### LINKED ISSUE #3301: Remove fetchAlertsAlertmanagerAction
**Component**:

ui

**Why this is needed**:

Alert fetching has moved to shell-ui

**What should be done**:

Remove fetchAlertsAlertmanagerAction

ChengYanJin:
**Component**:  UI

**Context**: 
There are some warnings regarding variable not tracked in dependency arrays.

**Summary**:
- fix the warnings in devtools
- remove the fetch  alert action 

**Acceptance criteria**: 

---

Closes: #3301


shell-ui/src/navbar/index.js
```diff
@@ -124,7 +124,7 @@ const SolutionsNavbar = ({
     default:
       return <LoadingNavbar logo={logos.dark || `/brand/assets/logo-dark.svg`} />;
     case 'error':
-      return ReactDOM.render(<ErrorPage500 data-cy='sc-error-page500'/>, document.body);
+      return <ErrorPage500 data-cy="sc-error-page500" />;
```
JBWatenbergScality: I believe this is not what we want either. With this behavior the navbar will be replaced with the error page but the rest of the. page will be displayed under the page. What we want instead at this level is raising an exception and catching it in the class SolutionsNavbarWebComponent where we would want to render the error page via ReactDOM.render on the first child element of the body I think.

JBWatenbergScality (dismissed):

JBWatenbergScality (dismissed):

JBWatenbergScality (approved):

PR #3352, opened 2021-05-04
Add custom metalk8s_network.routes execution module

### LINKED ISSUE #3349: Service CIDR pre-check fails if `blackhole` routes exist
**Component**: salt, networking

**What happened**:

Calico creates a "blackhole" route with the BIRD protocol for the CIDR of ClusterIPs assigned to a host, so that it can advertise it to its BGP peers.
Such routes are not properly recognized by Salt's `network.routes` execution method, which we use to check whether a route exists for the Service CIDR.
As such, re-running bootstrap on a node with such a "blackhole" route will fail with a traceback similar to the following:

```
Failure while running step 'Checking local node'

Command: check_local_node

Output:

<< BEGIN >>
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
ValueError: 'blackhole' does not appear to be an IPv4 or IPv6 network
  File "/usr/bin/salt-call", line 11, in <module>
    load_entry_point('salt==3002.6', 'console_scripts', 'salt-call')()
  File "/usr/lib/python3.6/site-packages/salt/scripts.py", line 449, in salt_call
    client.run()
  File "/usr/lib/python3.6/site-packages/salt/cli/call.py", line 58, in run
    caller.run()
  File "/usr/lib/python3.6/site-packages/salt/cli/caller.py", line 112, in run
    ret = self.call()
  File "/usr/lib/python3.6/site-packages/salt/cli/caller.py", line 220, in call
    self.opts, data, func, args, kwargs
  File "/usr/lib/python3.6/site-packages/salt/executors/direct_call.py", line 12, in execute
    return func(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/modules/metalk8s_checks.py", line 52, in node
    service_route_ret = route_exists(destination=service_cidr, raises=False)
  File "/var/cache/salt/minion/extmods/modules/metalk8s_checks.py", line 411, in route_exists
    all_routes = __salt__["network.routes"](family="inet")
  File "/usr/lib/python3.6/site-packages/salt/modules/network.py", line 1769, in routes
    routes_ = _ip_route_linux()
  File "/usr/lib/python3.6/site-packages/salt/modules/network.py", line 554, in _ip_route_linux
    address_mask = convert_cidr(comps[0])
  File "/usr/lib/python3.6/site-packages/salt/modules/network.py", line 1237, in convert_cidr
    cidr = calc_net(cidr)
  File "/usr/lib/python3.6/site-packages/salt/modules/network.py", line 1260, in calc_net
    return __utils__["network.calc_net"](ip_addr, netmask)
  File "/usr/lib/python3.6/site-packages/salt/utils/network.py", line 1109, in calc_net
    return str(ipaddress.ip_network(ipaddr, strict=False))
  File "/usr/lib64/python3.6/ipaddress.py", line 84, in ip_network
    address)
ValueError: 'blackhole' does not appear to be an IPv4 or IPv6 network
```

**What was expected**:

Since such routes are expected, our checks should not fail when encountering them.

**Steps to reproduce**:

Re-running bootstrap on a functional deployment should exhibit the same error.
(Note that for some reason, VMs in our internal cloud, used by our CI, do not show such routes, hence do not fail when re-running bootstrap).

Implement our own parsing of the kernel routing table, instead of relying on Salt (also open an upstream ticket to fix this behaviour), and ignore such `blackhole` routes.

alexandre-allard:
**Component**: salt, tests

**Context**:
We rely on the Salt `network.routes` execution module to retrieve the routes during pre-check to ensure a route exists for a specific destination (or at least a default route).
The issue is that this module is bugged and fails when there is a blackhole route defined, which is our case, since Calico defines one.

**Summary**:
We implement a basic version of the `network.routes` module, fixing the blackhole issue and only keeping what we really need.

**Acceptance criteria**:
Unit tests are OK and `salt-call metalk8s_network.routes` returns the list of routes without crashing when a blackhole route is defined and there is no `net-tools` package installed

---

Closes: #3349

### DIFF
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -167,3 +167,62 @@ def get_listening_processes():
         )
 
     return all_listen_connections
+
+
+def routes():
+    """
+    Return currently configured IPv4 routes from routing table
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' metalk8s_network.routes
+    """
+    ret = []
+    cmd = "ip -4 route show table main"
+    out = __salt__["cmd.run"](cmd)
+    for line in out.splitlines():
+        comps = line.split()
+
+        if comps[0] in ("unreachable", "blackhole"):
+            continue
+
+        if comps[0] == "default":
+            ip_interface = ""
+            if comps[3] == "dev":
+                ip_interface = comps[4]
+
+            ret.append(
+                {
+                    "addr_family": "inet",
+                    "destination": "0.0.0.0",
+                    "gateway": comps[2],
+                    "netmask": "0.0.0.0",
+                    "flags": "UG",
+                    "interface": ip_interface,
+                }
+            )
+        else:
+            try:
+                address_mask = __salt__["network.convert_cidr"](comps[0])
+            except ValueError:
+                log.warning("Unsupported route type or format: %s", line)
+                continue
+
+            ip_interface = ""
+            if comps[1] == "dev":
+                ip_interface = comps[2]
+
+            ret.append(
+                {
+                    "addr_family": "inet",
+                    "destination": address_mask["network"],
+                    "gateway": "0.0.0.0",
+                    "netmask": address_mask["netmask"],
+                    "flags": "U",
+                    "interface": ip_interface,
+                }
+            )
+
+    return ret
--- a/salt/tests/unit/modules/files/test_metalk8s_network.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_network.yaml
@@ -37,3 +37,49 @@ get_listening_processes:
         127.0.0.1:
           pid: 123
           name: likely-something
+
+routes:
+  # 0. Default route
+  - ip_route_output: |-
+      default via 10.200.0.1 dev eth0
+    result:
+      - &default_route
+        addr_family: inet
+        destination: 0.0.0.0
+        flags: UG
+        gateway: 10.200.0.1
+        interface: eth0
+        netmask: 0.0.0.0
+  # 1. A simple route
+  - ip_route_output: |-
+      10.200.0.0/16 dev eth0 proto kernel scope link src 10.200.2.41
+    result:
+      - &simple_route
+        addr_family: inet
+        destination: 10.200.0.0
+        flags: U
+        gateway: 0.0.0.0
+        interface: eth0
+        netmask: 255.255.0.0
+  # 2. A blackhole route
+  - ip_route_output: |-
+      blackhole 10.233.162.0/26 proto bird
+    result: []
+  # 3. Multiple routes
+  - ip_route_output: |-
+      default via 10.200.0.1 dev eth0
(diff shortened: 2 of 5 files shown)


salt/_modules/metalk8s_network.py
```diff
@@ -167,3 +167,58 @@ def get_listening_processes():
...
+                    "addr_family": "inet",
+                    "destination": "0.0.0.0",
+                    "gateway": comps[2],
+                    "netmask": "0.0.0.0",
+                    "flags": "UG",
+                    "interface": ip_interface,
+                }
+            )
+        else:
+            address_mask = __salt__["network.convert_cidr"](comps[0])
```
gdemonet: So, if for some reason we miss another special case (I don't know if we can be 100% certain that the first part of a line is always either `unreachable`, `blackhole`, `default`, or a valid CIDR), this will fail just as before with `blackhole`.

Wouldn't it make sense to catch errors here, and handle it with an error log instead of crashing? I mean, all we care about is whether a valid route is available, not checking the sanity (if there is such a thing) of the overall routing table.

salt/_modules/metalk8s_network.py
```diff
@@ -167,3 +167,57 @@ def get_listening_processes():
...
+                    "netmask": "0.0.0.0",
+                    "flags": "UG",
+                    "interface": ip_interface,
+                }
+            )
+        else:
+            try:
+                address_mask = __salt__["network.convert_cidr"](comps[0])
+            except ValueError:
+                continue
```
gdemonet, line 211: Maybe add a log, at least in `WARNING`?

TeddyAndrieux (dismissed):

gdemonet (dismissed):

alexandre-allard:
@gdemonet @TeddyAndrieux  Forgot to update the tests according to the changes requested... you can re-approve now :)

salt/tests/unit/modules/files/test_metalk8s_network.yaml
```diff
@@ -37,3 +37,49 @@ get_listening_processes:
...
+      blackhole 10.233.162.0/26 proto bird
+    result:
+      - *default_route
+      - *simple_route
+  # 4. No routes
+  - ip_route_output: ''
+    result: []
+  # 5. Unsupported type or format
+  - ip_route_output: |-
+      banana route via foo
```
gdemonet, line 81: :rofl:

gdemonet (approved):

salt/tests/unit/modules/files/test_metalk8s_network.yaml
TeddyAndrieux: :banana:

TeddyAndrieux (approved):

salt/tests/unit/modules/files/test_metalk8s_network.yaml
alexandre-allard: I've been greatly inspired by your tests. :)

Issue #3271, opened 2021-04-07
build: Building `metalk8s-ui` container may fail in concurrent mode


gdemonet:
**Component**: build

**What happened**:

Sometimes, we see errors when building with `./doit.sh -n 4` such as the following:

```
.  IMG BUILD      metalk8s-ui:2.8.1-dev
TaskError - taskid:_image_build:metalk8s-ui
PythonAction Error
Traceback (most recent call last):
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/.venv/lib64/python3.6/site-packages/docker/utils/build.py", line 96, in create_archive
    with open(full_path, 'rb') as f:
PermissionError: [Errno 13] Permission denied: '/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/_build/packages/debian/calico-cni-plugin-3.17.0/opt/cni/bin/calico-ipam'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/.venv/lib64/python3.6/site-packages/doit/action.py", line 437, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/buildchain/docker_command.py", line 102, in decorated_task
    task_func(*args, **kwargs)
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/buildchain/docker_command.py", line 130, in docker_build
    forcerm=True,
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/.venv/lib64/python3.6/site-packages/docker/models/images.py", line 279, in build
    resp = self.client.api.build(**kwargs)
  File "/home/eve/workspace/kube_pod-prod-metalk8s-backend-0/build/metalk8s-2.8/buildchain/.venv/lib64/python3.6/site-packages/docker/api/build.py", line 160, in build
    path, exclude=exclude, dockerfile=dockerfile, gzip=gzip
[...]

PR #3353, opened 2021-05-04
Allow defining an explicit `docker build` context instead of a directory

gdemonet:
**Component**: build

**Context**: See #3271

**Summary**:

- Fix an issue with dependency checking for `LocalImage` targets, and replace this check with a dynamic calculation of these dependencies.
- Introduce an `ExplicitContext` object to prepare a tar archive in memory before sending it as a build context to Docker client.

**Acceptance criteria**: Builds still pass, no flakiness is observed on `_image_build:metalk8s-ui`.

---

Closes: #3271

### DIFF
--- a/buildchain/buildchain/targets/local_image.py
+++ b/buildchain/buildchain/targets/local_image.py
@@ -17,65 +17,16 @@ import operator
 import os
 import re
 from pathlib import Path
-from typing import Any, Dict, List, Match, Optional, Union
-
-from doit.exceptions import TaskError # type: ignore
+from typing import Any, Dict, List, Optional, Union
 
 from buildchain import config
+from buildchain import constants
 from buildchain import coreutils
 from buildchain import docker_command
 from buildchain import types
 
+from . import base
 from . import image
-
-
-class DockerFileDep:
-    """A Dockerfile directive line creating external file dependencies."""
-
-    def __init__(
-        self,
-        line_no: int,
-        line_contents: str,
-        line_match: Match[str],
-        dockerfile_path: Path
-    ):
-        self.line_no = line_no
-        self.line_contents = line_contents
-        self.match = line_match
-        self.dockerfile_path = dockerfile_path
-
-    def expand_dep(self) -> List[Path]:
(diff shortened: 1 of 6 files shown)


buildchain/buildchain/targets/local_image.py
```diff
@@ -212,3 +159,19 @@ def _do_save(self) -> List[types.Action]:
...
+            return [dep]
+        # Directory or `.`
+        if dep.is_dir():
+            return list(coreutils.ls_files_rec(dep))
+        # Globs - `*` or specific e.g. `*.py`, `*.repo`
+        return [
+            sub_path
+            for sub_path in dep.parent.glob(dep.name)
+            if sub_path.is_file()
+        ]
```
alexandre-allard, line 218: What happens if I have the following:
```
COPY src/* dest/
```
And in `src`, I have multiple level of directories, e.g.:
```
src
├── foo
│   └── bar
│       └── baz
└── banana
```
We don't add `src/foo/bar/baz` as a dependency, no ?

Note: there is no such case in any of our Dockerfile
Note2: it was already the case before this PR

gdemonet: I didn't change the behaviour for parsing such glob patterns, so I don't really know, and didn't test. I would assume `pathlib.Path.glob` works roughly the same as Docker does when parsing those statements. If it doesn't, I'd suggest we open a tracking issue and tackle it in another PR :wink:

gdemonet: Just checked, and indeed, there is a difference between the two. I guess this can be easily improved, by doing something like:

```python
deps = []
for sub_path in dep.parent.glob(dep.name):
    if sub_path.is_file():
        deps.append(sub_path)
    elif sub_path.is_dir():
        deps.extend(coreutils.ls_files_rec(sub_path))
    else:
        # do something? or not?
        continue
return deps

alexandre-allard (approved):

buildchain/buildchain/targets/local_image.py
gdemonet: See #3359

PR #3356, opened 2021-05-06
Customize Prometheus label selector

### LINKED ISSUE #2199: Customize label selector for `prometheus-operator`-managed ServiceMonitors

alexandre-allard:
**Component**: charts, salt

**Context**:
Actually we use default label selector from kube-prometheus-stack charts `release: prometheus-operator` and `app: prometheus-operator` for Prometheus to be able to retrieve `ServiceMonitor` and `PrometheusRules` objects.
Those labels do not make real sense and should be replaced by something more meaningful.

**Summary**:
Replace the default selector by a new one `metalk8s.scality.com/monitor: ''`.

**Acceptance criteria**:
Monitoring still works as expected.

---

Closes: #2199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Release 2.10.0 (in development)
-## Enhancements
+### Enhancements
 
 - Bump Kubernetes version to 1.21.0
   (PR[#3354](https://github.com/scality/metalk8s/pull/3354))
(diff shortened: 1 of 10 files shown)


alexandre-allard-scality:
@bert-e wait

alexandre-allard:
Putting @scality/storage and @scality/object as reviewers so they are aware of this breaking change.
This change will only be done in metalk8s-2.10.0 and later versions

NicolasT:
For @scality/storage and @scality/object: you can *add* this new label next to the ones already in place, to be compatible with 'current' MetalK8s as well as 'future' MetalK8s. Then, the current labels can be removed over time, once 'current' MetalK8s is no longer supported (by your application).

TeddyAndrieux:
Worth a changelog entry no ?

CHANGELOG.md
```diff
@@ -1,18 +1,26 @@
 # CHANGELOG
 
 ## Release 2.10.0 (in development)
-## Enhancements
+### Enhancements
```
TeddyAndrieux, line 4: Oh :+1:

TeddyAndrieux (approved):

PR #3365, opened 2021-05-07
Sync kubelet with kubeadm (and cleanup)

### LINKED ISSUE #3355: Sync Kubelet config with kubeadm
**Component**:

'salt', 'kubernetes'

**Why this is needed**:

The kubelet config do not get updated since a long time we should keep this one in sync with what is done by kubeadm

**What should be done**:

- Add kubelet config "support" in `get-kubeadm-manifests` tool
- Update kubelet config to match (more or less) kubeadm one

TeddyAndrieux:
**Component**:

'salt', 'kubelet'

**Context**: 

Sync the kubelet config we deploy with what is done in kubeadm so that future Kubernetes version bump will be simpler

**Summary**:

- Add kubelet config in `get-kubeadm-manifest` tool
- Sync kubelet config

NOTE: The only real changes here in the kubelet config is `nodeStatusReportFrequency` that was set to `1m0s` and now is set to
`0s` (so we use the default which is `5m`)

See the commit message for more details

---

Fixes: #3355

### DIFF
--- a/salt/metalk8s/kubernetes/kubelet/standalone.sls
+++ b/salt/metalk8s/kubernetes/kubelet/standalone.sls
@@ -40,78 +40,52 @@ Create kubelet config file:
     - dir_mode: '0750'
     - formatter: yaml
     - dataset:
-        address: {{ grains['metalk8s']['control_plane_ip'] }}
-        kind: KubeletConfiguration
+      # kubeadm config {
         apiVersion: kubelet.config.k8s.io/v1beta1
-        staticPodPath: /etc/kubernetes/manifests
         authentication:
           anonymous:
             enabled: false
           webhook:
-            cacheTTL: 2m0s
+            cacheTTL: 0s
             enabled: true
           x509:
             clientCAFile: /etc/kubernetes/pki/ca.crt
         authorization:
           mode: Webhook
           webhook:
-            cacheAuthorizedTTL: 5m0s
-            cacheUnauthorizedTTL: 30s
-        cgroupDriver: cgroupfs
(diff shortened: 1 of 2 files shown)


salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -40,78 +40,52 @@ Create kubelet config file:
...
           mode: Webhook
           webhook:
-            cacheAuthorizedTTL: 5m0s
-            cacheUnauthorizedTTL: 30s
-        cgroupDriver: cgroupfs
-        cgroupsPerQOS: true
+            cacheAuthorizedTTL: 0s
+            cacheUnauthorizedTTL: 0s
+        # NOTE: we use cgroupfs for the moment
+        # cgroupDriver: systemd
```
TeddyAndrieux, line 59: This change will be done in another PR, 
I tried and it does not work as-is (maybe we need to do some changes on containerd and/or use a newer containerd version, not sure yet)

gdemonet: Do we have a ticket to track this topic? Could make sense, at least to publicly discuss the change and potential impact.

TeddyAndrieux: Not yet, but right, I can create one

TeddyAndrieux: #3366

alexandre-allard (approved):

PR #3370, opened 2021-05-10
Bump dex image version

### LINKED ISSUE #3279: Bump Dependencies for MetalK8s 2.10
**Component**:

'build'

**Why this is needed**:

We didn't update dependencies such as Prometheus, Grafana, etc ... since branch 2.8. We want to always embed the latest fix, security fixes, or new features and do it before releasing the first patch version of a specific branch.

**What should be done**:

Check and update all dependencies and make sure we don't break anything ...

**Test plan**:

alexandre-allard:
**Component**: salt, charts, build

**Context**: 
Update Dex addon

**Summary**:
We can not bump the chart version as it is no longer supported and the new chart (https://github.com/dexidp/helm-charts) does not offer yet the full capabilities that we need (it's not an exhaustive list, there may be more):

- no HTTPS configuration in Ingress
- not possible to disable HTTP in Service/Deployment
- not possible to override checksum/config annotation

**Acceptance criteria**: 
Dex is still working as expected

---

See: #3279


charts/dex.yaml
```diff
@@ -3,7 +3,7 @@ image: '__image__(dex)'
 # NOTE: We do not use the new Helm3 chart for the moment as it's not yet part
 # of stable repo, but since Dex 2.24 is affected by CVE (CVE-2020-15216) use
 # the newest 2.27 image even if we still use the deprecated chart
```
TeddyAndrieux: Nitpick you should update this comment as well
```suggestion
# the newest 2.28 image even if we still use the deprecated chart
```

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):

PR #3396, opened 2021-05-26
salt: Support `127.0.0.1` hostPort

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

Support `127.0.0.1` as hostPort

**Summary**:

Before this commit we only support hostPort on workloadPlane IP, this
commit also add `127.0.0.1` as supported hostPort

NOTE: It's needed for Kubernetes conformance tests for Kubernetes 1.20+

---

### DIFF
--- a/salt/metalk8s/kubernetes/cni/calico/configured.sls
+++ b/salt/metalk8s/kubernetes/cni/calico/configured.sls
@@ -63,7 +63,7 @@ Create CNI calico configuration file:
             snat: true
             capabilities:
               portMappings: true
-            conditionsV4: ["-d", "{{ grains.metalk8s.workload_plane_ip }}/32"]
+            conditionsV4: ["-d", "{{ grains.metalk8s.workload_plane_ip }}/32,127.0.0.1/32"]
           # Note: Calico upstream enables the `bandwidth` CNI plugin by default.
           # However, this plugin (executable) is not available in the CNI RPM
           # package we currently install. Hence, not enabling this functionality
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # CHANGELOG
 
 ## Release 2.9.3 (in development)
+## Enhancements
+- Allow hostPort on 127.0.0.1
+  (PR[#3396](https://github.com/scality/metalk8s/pull/3396))
 
 ## Release 2.9.2
 ### Bug fixes


alexandre-allard (approved):

salt/metalk8s/kubernetes/cni/calico/configured.sls
```diff
@@ -63,7 +63,7 @@ Create CNI calico configuration file:
             snat: true
             capabilities:
               portMappings: true
-            conditionsV4: ["-d", "{{ grains.metalk8s.workload_plane_ip }}/32"]
+            conditionsV4: ["-d", "{{ grains.metalk8s.workload_plane_ip }}/32,127.0.0.1/32"]
```
alexandre-allard, line 66: Shouldn't it be on the whole `127.0.0.0/8` network ?

TeddyAndrieux: Hmmm I'm not sure, to me here we only want localhost and not any other "loopback" addresses used for .... whatelse
I think but not sure

gdemonet: IIUC, what we want is "all host-local addresses except control plane", no?

TeddyAndrieux: To me, we want workload plane IP and that's it (just we add 127.0.0.1 as it's needed for conformance tests and does not hurt anyway)

alexandre-allard: I don't really know what we should do either and I'm not even sure that's really important.
We can merge your changes as that's only what's requested by the K8s conformance tests.
But I don't see any valid reason to restrict to only `127.0.0.1/32` instead of `127.0.0.0/8`

TeddyAndrieux: Agree :+1:,
Maybe we will want to make it configurable at some point also, but for the moment let's not bother about it :smiley:

NicolasT: Likely, yes.

NicolasT:
Others ran into this as well:
- https://github.com/cilium/cilium/issues/14287 
- https://github.com/kubernetes/kubernetes/pull/100522

salt/metalk8s/kubernetes/cni/calico/configured.sls
NicolasT: Though, maybe the current change is sufficient. It passes conformance, so it's supposed to be conformant. I'm not sure where it's said a K8s cluster *must* expose `NodePort`s on `127.0.0.0/8`. If this is the case, maybe we should widen things up further.

NicolasT:
For future reference, cfr. https://kubernetes.slack.com/archives/C09QYUH5W/p1622023243096200

Issue #2002, opened 2019-11-05
Bootstrap HA 


thomasdanan:
We want to have high availability of the bootstrap node as it is a single access point for all operations (through salt-master) and because this is from where the containers images are served + CA may be part of the bootstrap node as well.  An Active/Passive approach (especially for salt master) is probably acceptable

TeddyAndrieux:
POC Registry HA (it's only a part of "Bootstrap HA"):

1.Copy all ISOs on another host at the same place as on the bootstrap one (to match what is in the bootstrap config)
2. Mount ISO(s) `salt-call state.sls metalk8s.archives.mounted saltenv=<saltenv>`
3. Deploy repository `salt-call state.sls metalk8s.repo.installed saltenv=<saltenv>`
4. Reconfigure ALL containerd to have both registry endpoints and restart them (note: to do it with the salt state it needs some changes in the code to support "multiple endpoints" for the repository)

Containerd will automatically try both endpoints to pull images so "registry HA"

NicolasT:
> Containerd will automatically try both endpoints to pull images so "registry HA"

What happens if both registries are up, but one doesn't have all ISOs (yet), and 404s when `containerd` requests an image/layer? Will it try on the other addresses as well?

TeddyAndrieux:
Right, we discuss about it this morning (during our standup) and yes it works

First repo (2.10 ISOs not here yet): 
```
2021-05-21T12:52:56.084559995Z stdout F 10.100.6.146 - - [21/May/2021:12:52:56 +0000] "HEAD /v2/metalk8s-2.10.0-dev/kube-apiserver/manifests/v1.21.0?ns=metalk8s-registry-from-config.invalid HTTP/1.1" 404 0 "-" "containerd/1.4.3" "-"
```
Second repo (with the ISO mounted/configured):
```
2021-05-21T12:52:56.082589301Z stdout F 10.100.6.146 - - [21/May/2021:12:52:56 +0000] "HEAD /v2/metalk8s-2.10.0-dev/kube-apiserver/manifests/v1.21.0?ns=metalk8s-registry-from-config.invalid HTTP/1.1" 200 0 "-" "containerd/1.4.3" "-"
```

PR #3400, opened 2021-05-27
Procedure to deploy registry HA

TeddyAndrieux:
**Component**:

'salt', 'docs', 'tests'

**Context**: 

The first step of bootstrap HA (registry + repository HA)

**Summary**:

- Update the salt module to retrieve Kubernetes endpoint to retrieve the full list of all endpoints and not only a single one.
- Configure contained and repo with all repositories endpoints (HA registry and repo)
- Add tests for registry
- Add procedure to deploy registry container on a second node (HA registry)

---

See: #2002


docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
+Registry HA
+===========
+
+MetalK8s come with his own registry used to serve all images used by MetalK8s
+containers, so that it can run fully offline, this registry container sit on
+the Bootstrap node.
```
alexandre-allard:
```suggestion
To be able to run fully offline, MetalK8s comes with its own registry serving
all necessary images used by its containers.
This registry container sits on the Bootstrap node.
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Highly available registry allow to lose the Bootstrap node and still be able
+to schedule pods on nodes that need to pull some container images. So it
+means, lose of Bootstrap node should have no impact on the workload pods
+running in the cluster.
```
alexandre-allard:
```suggestion
With a highly available registry, container images are served by multiple nodes,
which means the Bootstrap node can be lost without impacting the cluster.
It allows pods to be scheduled, even if the needed images are not cached locally.
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+  functionnaly. Check this ticket for more informations
+  https://github.com/scality/metalk8s/issues/2002
+
+Prepare the node
+----------------
+
+In order to be able to schedule a repository pod on a node that will
+be used as registry, you need this node to be part of the MetalK8s
+cluster, no specific roles or taints is needed, this pods can be scheduled
+on any node.
```
alexandre-allard:
```suggestion
To configure a node to host a registry, a ``repository`` pod must be scheduled on it.
This node must be part of the MetalK8s cluster and no specific roles or taints are needed.
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+All ISOs listed in ``archives`` section of
+``/etc/metalk8s/bootstrap.yaml`` and of ``/etc/metalk8s/solutions.yaml``
+from the Bootstrap node must be on the exact same place on the node
+that will be used as registry.
```
alexandre-allard:
```suggestion
All ISOs listed in the ``archives`` section of
``/etc/metalk8s/bootstrap.yaml`` and ``/etc/metalk8s/solutions.yaml``
must be copied from the Bootstrap node to the target node at exactly the same location.
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+- Prepare all the MetalK8s ISOs
+
+  .. parsed-literal::
+
+    root@node-1 $ salt-call state.sls \\
+        metalk8s.archives.mounted \\
+        saltenv=metalk8s-|version|
+
+- If you have some solutions, prepare solutions ISOs
```
alexandre-allard:
```suggestion
- If you have some solutions, prepare the solutions ISOs
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+All ISOs listed in ``archives`` section of
+``/etc/metalk8s/bootstrap.yaml`` and of ``/etc/metalk8s/solutions.yaml``
+from the Bootstrap node must be on the exact same place on the node
+that will be used as registry.
+
+Deploy the registry
+-------------------
+
+Connect on the node where you want to deploy this registry and run the
```
alexandre-allard:
```suggestion
Connect to the node where you want to deploy the registry and run the
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+  .. parsed-literal::
+
+    root@node-1 $ salt-call state.sls \\
+        metalk8s.repo.installed \\
+        saltenv=metalk8s-|version|
+
+
+Reconfigure all nodes
+---------------------
```
alexandre-allard:
```suggestion
Reconfigure all the nodes
-------------------------
```
Or even (since we more specifically reconfigure containerd):
```suggestion
Reconfigure the container engines
---------------------------------
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+
+In order to have registry HA we need to reconfigure all containerd to
+use both registry as endpoint so that if one is down containerd
+automatically pull from the other registry.
+
+For this you need to run a state on all node using Salt Master
```
alexandre-allard:
```suggestion
From the Boostrap node, run:
```

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,78 @@
...
+        metalk8s.repo.installed \\
+        saltenv=metalk8s-|version|
+
+
+Reconfigure all nodes
+---------------------
+
+In order to have registry HA we need to reconfigure all containerd to
+use both registry as endpoint so that if one is down containerd
+automatically pull from the other registry.
```
alexandre-allard:
```suggestion
Containerd must be reconfigured to add the freshly deployed registry to its endpoints
and so it can still pull images in case the Bootstrap node's one is down.
```

alexandre-allard (commented):
First round, wording/typos/etc. :)

tests/post/steps/test_registry.py
```diff
@@ -0,0 +1,173 @@
...
+
+@when(parsers.parse("we pull metalk8s utils image from node '{node_name}'"))
+def pull_metalk8s_utils(host, context, ssh_config, utils_image, node_name):
+    if ssh_config:
+        node = testinfra.get_host(node_name, ssh_config=ssh_config)
+    else:
+        node = host
+
+    with node.sudo():
+        context["pull_ret"] = node.run("crictl pull {}".format(utils_image))
```
alexandre-allard, line 73: details, but could be slightly changed to handle any image and so be re-usable elsewhere.

docs/operation/registry_ha.rst
TeddyAndrieux: Right, let's go with contaienr engine

salt/tests/unit/modules/files/test_metalk8s_kubernetes_utils.yaml
```diff
@@ -165,25 +165,94 @@ get_service_endpoints:
...
+          http: 8080
 
-  # 2. Error when getting the Endpoint object
+  # 3. Error when getting the Endpoint object
   - obj: False
     raises: True
     result: "Unable to get kubernetes endpoints for my_service in namespace my_namespace"
 
-  # 3. Endpoint object does not exists
+  # 4. Endpoint object does not exists
```
alexandre-allard:
```suggestion
  # 4. Endpoint object does not exist
```

salt/tests/unit/modules/files/test_metalk8s_kubernetes_utils.yaml
```diff
@@ -165,25 +165,94 @@ get_service_endpoints:
...
   - obj: null
     raises: True
     result: "Unable to get kubernetes endpoints for my_service in namespace my_namespace"
 
-  # 4. Endpoint do not have any addresses
+  # 5. Endpoint do not have any addresses
```
alexandre-allard:
```suggestion
  # 5. Endpoint does not have any address
```

salt/metalk8s/repo/redhat.sls
```diff
@@ -41,29 +60,25 @@ Ensure yum versionlock plugin is enabled:
...
+                            repo_name ~ "-el$metalk8s_osmajorrelease") %}
+  {%- endfor %}
   {%- set gpg_keys = [] %}
   {%- for gpgkey in repo_config.gpgkeys %}
-    {%- do gpg_keys.append(repo_url ~ "/" ~ gpgkey) %}
+    {# NOTE: We only use the first repo url as gpg key since the keys are the same,
+             and we do not want to depends on both repos to import this key,
+       When installing a package and the gpg key of the package is not already imported
+       on the host, it will try to import **all** gpgkeys listed and fail if we are
+       unable to download one of them #}
```
alexandre-allard:
```suggestion
    {# NOTE: We only use the first repo URL for GPG key since they are the same,
       and we do not want to depend on both repos to import this key.
       When installing a package, if the package's GPG key is not already imported
       on the host, it will try to import **all** listed GPG keys and will fail if unable to
       download one of them #}
```

alexandre-allard: Well, but it means it the Bootstrap node is down, we can't install new packages ?
Can't we just pre isntall the GPG keys from all repositories on all nodes ?
Or even if the key is pre-installed, it tries to reach the one from the remote repository ?

alexandre-allard (commented):
LGTM.

tests/post/steps/test_registry.py
TeddyAndrieux: Hmm, sure but it could be done at the time we will need such stuff because anyway today it's in the "registry" file and I will not put it in conftest or elsewhere since we do not need it elsewhere for the moment.
The "problem" with a generic function is that you will need to provide the version of the image and here it's the version of MetalK8s itself and we do not want to hardcode the version in the test feature files, we could imagine a specific `when` for MetalK8s version image but... not sure it worth doing something now

alexandre-allard (dismissed):

salt/metalk8s/repo/redhat.sls
alexandre-allard: I'm not sure that's an issue for now, because packages are only needed when deploying stuff and the salt-master is not HA yet, so if the bootstrap node is down... we can't deploy anything anyway.

TeddyAndrieux: If the gpg key is already imported it's ok even if the "gpgkey" listed is not reachable. 
So to me, we could keep the bootstrap node only for the moment since we need bootstrap to deploy the node anyway, it means we will import those keys when deploying the node anyway.

But I see what you mean we could manually import those keys first and then we are good, yes sure the only "issue" is that I need to manually do all the "yum variable" replacements in Salt... and not sure it's worth doing it

alexandre-allard: Or we could get rid of this yum variable and hardcode the value in the URL, using this variable does not bring anything, except extra complexity.
But as said above, it can wait for the full bootstrap to be HA, I think.

TeddyAndrieux: Right, just if you "upgrade" from rhel7 to rhel8 you can keep repo definition but ... not sure we care

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,79 @@
...
+endpoints and so it can still pull images in case the Bootstrap node's one is
+down.
+
+From the Bootstrap node, run:
+
+.. parsed-literal::
+
+  root@bootstrap $ kubectl exec -n kube-system -c salt-master \\
+     --kubeconfig=/etc/kubernetes/admin.conf \\
+     salt-master-bootstrap -- salt '*' state.sls \\
```
NicolasT: This pod name is not fixed, so this isn't a copy-paste command, even though it looks like one. This needs to be explained.

salt/_pillar/metalk8s_endpoints.py
```diff
@@ -30,19 +30,28 @@ def ext_pillar(minion_id, pillar, kubeconfig):  # pylint: disable=unused-argumen
...
                     ](service, namespace, kubeconfig)
                 except CommandExecutionError as exc:
-                    service_endpoints = __utils__["pillar_utils.errors_to_dict"](
-                        str(exc)
-                    )
+                    errors.append(str(exc))
+
+                # NOTE: This is needed for downgrade as this pillar
+                # is used for downgrade
+                # To be removed in 2.11
```
NicolasT, line 47: Maybe file a ticket to do/track this.

NicolasT (dismissed):

docs/operation/registry_ha.rst
TeddyAndrieux: Agree, but it's not only here, everywhere in the documentation we do this ... which is an issue I agree

salt/_pillar/metalk8s_endpoints.py
TeddyAndrieux: I will

docs/operation/registry_ha.rst
NicolasT: That's a bug, cc @thomasdanan

TeddyAndrieux: I will replace this one at least

salt/_pillar/metalk8s_endpoints.py
TeddyAndrieux: Done #3413

salt/metalk8s/repo/redhat.sls
alexandre-allard: Yes, but if you upgrade, you will likely completely reinstall the node and you still need to change the variable value (it's easier, I agree, but not sure it worth it).
I think if someone upgrade, he will have some other troubles and will probably have to re-run an highstate anyway, fixing this at the time.

TeddyAndrieux: Yes totally agree

Anyway, I will **not** do this in this PR :smile:

docs/operation/registry_ha.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Reconfigure the container engines
+---------------------------------
+
+Containerd must be reconfigured to add the freshly deployed registry to its
+endpoints and so it can still pull images in case the Bootstrap node's one is
+down.
+
+From the Bootstrap node, run (replace ``<bootstrap_node_name>`` with the
+acutal Bootstrap node name):
```
alexandre-allard:
```suggestion
actual Bootstrap node name):
```

alexandre-allard (approved):

PR #3406, opened 2021-05-31
ci: Add shell-ui publish stage

### LINKED ISSUE #3269: publish a shell-ui docker image in registry.scality.com
**Component**:

CI

**Why this is needed**:

In order to enable solutions-ui using shell-ui components to be tested independently from metalk8s.

**What should be done**:

We should publish a shell-ui image in namespace metalk8s-dev when merging a branch and tag it to the version number in metalk8s namespace when releasing metalk8s.

alexandre-allard:
**Component**: ci

**Context**:
We need to provide shell-ui image from outside of MetalK8s to ease development of other product using it.

**Summary**:
Add a publish-images stage in the CI.
This stage is triggered during post-merge to publish latest development images.
It can also be triggered manually, especially during a release to push new stable images.
By default, images are pushed into the development registry, but you can force the production one by setting the `use_production_registry` property to `true`.

**Acceptance criteria**:
`publish` step is working as expected (see https://eve.devsca.com/github/scality/metalk8s/#/builders/1/builds/35056)

---

Closes: #3269

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -111,11 +111,11 @@ models:
   - Upload: &upload_artifacts
       name: Upload artifacts
       source: artifacts
-      alwaysRun: True
+      alwaysRun: true
   - Upload: &upload_report_artifacts
       name: Upload sosreport logs to artifacts
       source: sosreport
-      alwaysRun: True
+      alwaysRun: true
   - Upload: &upload_cypress_artifacts
       name: Upload Cypress and UI Junit folder
       source: ui-tests
@@ -509,7 +509,7 @@ models:
           Artifact Url: %(prop:artifacts_public_url)s
           Branch: %(prop:branch)s
           Commit: %(prop:revision)s
-      haltOnFailure: True
+      haltOnFailure: true
   - ShellCommand: &add_final_status_artifact_success
       <<: *add_final_status_artifact
       name: Add successful status to artifacts
@@ -525,7 +525,22 @@ models:
   - Upload: &upload_final_status_artifact
       name: Upload final status to artifacts
       source: build_status
-      alwaysRun: True
+      alwaysRun: true
+  - ShellCommand: &image_registry_login
+      name: Login to Harbor image registry
+      command: |
+        if [ "$USE_PRODUCTION_REGISTRY" = true ]; then
+            HARBOR_LOGIN='%(secret:harbor_prod_login)s'
+            HARBOR_PASSWORD='%(secret:harbor_prod_token)s'
+        else
+            HARBOR_LOGIN='%(secret:harbor_dev_login)s'
+            HARBOR_PASSWORD='%(secret:harbor_dev_token)s'
+        fi
+        docker login --username "$HARBOR_LOGIN" --password "$HARBOR_PASSWORD" "$REGISTRY_HOST"
+      env: &_env_image_registry
+        USE_PRODUCTION_REGISTRY: "%(prop:use_production_registry:-false)s"
+        REGISTRY_HOST: registry.scality.com
+      haltOnFailure: true
 
   # --- Re-usable steps related to terraform actions ---
   - ShellCommand: &terraform_install
@@ -806,7 +821,7 @@ stages:
       type: local
     steps:
       - ShellCommand: *add_final_status_artifact_failed
-      - SetProperty:
+      - SetProperty: &set_premerge_artifacts_name
           # Set premerge_artifacts_name to current artifacts name
           # as we are in premerge so we can use same step to retrieve ISO
           # in pre-merge and post-merge
@@ -815,9 +830,10 @@ stages:
           value: "%(prop:artifacts_name)s"
       - SetProperty: *set_premerge_url
       - TriggerStages:
-          name: Trigger build, docs and lint stages simultaneously
+          name: Trigger build, docs, test and lint stages simultaneously
           stage_names:
             - build
+            - build-shell-ui
             - docs
             - lint
             - unit_tests
@@ -826,7 +842,7 @@ stages:
       - SetPropertyFromCommand: *set_version_prop
       - SetPropertyFromCommand: *set_short_version_prop
       - TriggerStages:
-          name: Trigger multiple-nodes step with built ISO
+          name: Trigger single and multiple nodes install stages with built ISO
           stage_names:
             - single-node-install-rhel
             - multiple-nodes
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -2671,3 +2686,42 @@ stages:
...
+          workdir: build/shell-ui
+          haltOnFailure: true
+      - ShellCommand:
+          name: Publish shell-ui container image
+          command: |
+            if [ "$USE_PRODUCTION_REGISTRY" = true ]; then
+                PROJECT='%(secret:harbor_prod_project)s'
+                LATEST_STABLE_RELEASE=$(git tag | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n 1)
+                if [ "%(prop:shell_ui_version)s" = "$LATEST_STABLE_RELEASE" ]; then
+                    docker tag shell-ui:v%(prop:shell_ui_version)s shell-ui:stable
```
JBWatenbergScality: When using USE_PRODUCTION_REGISTRY we might want to tag the existing image from dev harbor project instead of tagging a freshly built one.

alexandre-allard: Since we're only pushing the development image in post-merge, it means during a release we would need to force a post-merge (or be sure nothing has been changed since the last post-merge ran) to publish a new production image.
It could work for this specific case, but I'm not sure that's a good thing to do as we're not really building the product from the  specific tag.
That said, I agree that's not perfect to rebuild a new image, which is not really tested (but it would also be the case if we take the one generated in post-merge).
It would probably be better to just publish the image used for integration tests, but it's not possible as it is done now.

eve/main.yml
```diff
@@ -2671,3 +2686,42 @@ stages:
...
+    steps:
+      - ShellCommand: *wait_for_docker
+      - ShellCommand: *image_registry_login
+      - Git: *git_pull
+      - SetPropertyFromCommand:
+          name: Set shell-ui version as a property
+          property: shell_ui_version
+          command: |
+            . ./VERSION
+            echo "$VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH$VERSION_SUFFIX"
```
gdemonet, line 2712: Isn't this already available as `prop:metalk8s_version`?

eve/main.yml
```diff
@@ -2671,3 +2686,42 @@ stages:
...
+      - ShellCommand:
+          name: Build shell-ui container image
+          command: docker build . --tag shell-ui:v%(prop:shell_ui_version)s
+          workdir: build/shell-ui
+          haltOnFailure: true
```
gdemonet, line 2717: Shouldn't we build (and save) this image during pre-merge instead? So if we want to add tests, we can.

eve/main.yml
```diff
@@ -856,13 +870,14 @@ stages:
...
       - TriggerStages:
-          name: Trigger post-merge lifecycle and solutions stages simultaneously
+          name: Trigger post-merge lifecycle, solutions, restore and publish stages simultaneously
           stage_names:
             - post-merge-solutions
             - lifecycle-dev-branch
             - lifecycle-patch-version
             - lifecycle-minor-version
             - bootstrap-restore
+            - publish-images
```
gdemonet, line 886: We wanted this step to be optional, so that a manual trigger of `post-merge` will not publish anything by default, only when triggered by the nightly schedule.

gdemonet (changes requested):

eve/main.yml
alexandre-allard: This property rely on the `product.txt` from pre-merge artifacts which may not be there depending from where this stage is triggered.
Anyway, I'll change this to achieve https://github.com/scality/metalk8s/pull/3406#discussion_r642814463

eve/main.yml
```diff
@@ -2671,3 +2692,95 @@ stages:
...
+                    docker tag shell-ui:v%(prop:metalk8s_version)s shell-ui:stable
+                fi
+            else
+                PROJECT='%(secret:harbor_dev_project)s'
+                LATEST_DEV_BRANCH=$(git branch -r --list "origin/development/[0-9].*[0-9]" | cut -d/ -f3 | sort -V | tail -n 1)
+                if [ "%(prop:branch)s" = "development/$LATEST_DEV_BRANCH" ]; then
+                    docker tag shell-ui:v%(prop:metalk8s_version)s shell-ui:latest
+                fi
+                docker tag shell-ui:v%(prop:metalk8s_version)s shell-ui:v%(prop:metalk8s_version)s-%(prop:commit_short_revision)s
+            fi
```
TeddyAndrieux: To me, we want to use a property instead of checking here

Like "is_latest" and "is_stable" and then we set `is_latest` in cron job for the latest development branch in the CI directly

alexandre-allard: Hm, yeah, why not, but so we need to change it every time we have a new branch...

TeddyAndrieux: Yes, but also it's more flexible; we could for example do not publish the latest dev branch if we think that this new branch is not good enough for the moment (even if it's only development)

eve/main.yml
```diff
@@ -2671,3 +2693,165 @@ stages:
...
+            else
+                PROJECT='%(secret:harbor_dev_project)s'
+            fi
+            echo "$REGISTRY_HOST/$PROJECT"
+          env: *_env_image_registry
+      - SetPropertyFromCommand:
+          name: Set is_dev property
+          property: is_dev
+          value: false
+          doStepIf: '%(prop:use_production_registry:-false)s'
```
gdemonet: This should be a `SetProperty` (not `SetPropertyFromCommand`)

alexandre-allard: Yup, I saw it, it's fixed.

eve/main.yml
alexandre-allard: We can do both now. :)

gdemonet (approved):

eve/main.yml
```diff
@@ -2671,3 +2693,166 @@ stages:
...
+          name: Push shell-ui image to the remote registry
+          command: |
+            if [ "$USE_PRODUCTION_REGISTRY" = true ]; then
+                PROJECT='%(secret:harbor_prod_project)s'
+            else
+                PROJECT='%(secret:harbor_dev_project)s'
+            fi
+            echo "The following $IMAGE_NAME tags will be pushed:"
+            docker images "$REGISTRY_HOST/$PROJECT/$IMAGE_NAME"
+            docker push "$REGISTRY_HOST/$PROJECT/$IMAGE_NAME"
```
gdemonet, line 2855: Don't we want:
```suggestion
            docker push --all-tags "$REGISTRY_HOST/$PROJECT/$IMAGE_NAME"
```
here?

alexandre-allard: No, because this option does not exist in this docker version and it's the default behavior.

PR #3410, opened 2021-06-02
UI: Add the ability to create LVM Logical Volume from Metalk8s UI

### LINKED ISSUE #3244: UI: Support LVMLogicalVolume volume in the metalk8s-ui
**Component**:

ui

**Why this is needed**:

`LVMLogicalVolume` support get added in https://github.com/scality/metalk8s/pull/3220 in storage operator

**What should be done**:

- [x] Properly listing `LVMLogicalVolume` volume in the UI (in #3316)
- [x] Ability to create that kind of volumes from the UI

NOTE: Documentation may need to be updated a bit to reflect this UI change

Volume object example: 
```
apiVersion: storage.metalk8s.scality.com/v1alpha1
kind: Volume
metadata:
  name: my_volume
spec:
  nodeName: my_node
  storageClassName: metalk8s
  lvmLogicalVolume:
    size: 10Gi
    vgName: my-vg
```

New fields in `lvmLogicalVolume`:
- `size` exactly the same as `sparseLoopDevice` size
- `vgName` the LVM Volume Group name to create the LVM LogicalVolume (Note: a VG could (and will) be used by several LVMLogicalVolume)

**Test plan**:

ChengYanJin:
**Component**: UI

**Context**:   Add the ability to create LVM Logical Volume from the UI 

**Summary**:
For the batch creation, we only implement the most simple case, in which we can only specify the Name between the volumes.
image

Add interaction test for this kind of volume.

Next step: 
We should display a message/notification when the volume is failed to create, for example, the Volume Group doesn't exist.
More generally speaking, we will organise a design workshop with @Cuervino and maybe @JBWatenbergScality @MonPote  to decide how to display the message when a failure popup.

**Acceptance criteria**: 

---

Closes: #3244

### DIFF
--- a/ui/src/containers/CreateVolume.js
+++ b/ui/src/containers/CreateVolume.js
@@ -582,6 +582,88 @@ const CreateVolume = (props) => {
                 };
               });
 
+              const sizeInputCombinedField = () => {
+                return (
+                  <SizeFieldContainer>
+                    <Input
+                      name="sizeInput"
+                      type="number"
+                      min="1"
+                      value={values.sizeInput}
+                      onChange={handleChange('sizeInput')}
+                      label={`${intl.translate('volume_size')}*`}
+                      error={touched.sizeInput && errors.sizeInput}
+                      onBlur={handleOnBlur}
+                    />
+                    <SizeUnitFieldSelectContainer>
+                      <Input
+                        clearable={false}
+                        type="select"
+                        options={optionsSizeUnits}
+                        noOptionsMessage={() => intl.translate('no_results')}
+                        name="selectedUnit"
+                        onChange={handleSelectChange('selectedUnit')}
+                        value={getSelectedObjectItem(
+                          optionsSizeUnits,
+                          values?.selectedUnit,
+                        )}
+                        error={touched.selectedUnit && errors.selectedUnit}
+                        onBlur={handleOnBlur}
+                      />
+                    </SizeUnitFieldSelectContainer>
+                  </SizeFieldContainer>
+                );
+              };
+
+              // render the fields specifically for different volume types
+              const renderVolumeTypeSpecificFields = (volumeType) => {
+                switch (volumeType) {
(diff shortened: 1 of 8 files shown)


ui/src/containers/CreateVolume.js
```diff
@@ -711,7 +732,7 @@ const CreateVolume = (props) => {
                           />
                         </SizeUnitFieldSelectContainer>
                       </SizeFieldContainer>
-                    ) : (
+                    ) : values.type === RAW_BLOCK_DEVICE ? (
```
JBWatenbergScality: Now that there is more than 2 options I suggest to refactor this to a switch case

ui/src/containers/CreateVolume.js
```diff
@@ -734,6 +755,49 @@ const CreateVolume = (props) => {
...
+                        </SizeFieldContainer>
+                        <Input
+                          name="vgName"
+                          value={values.vgName}
+                          onChange={handleChange('vgName')}
+                          label={`Volume Group Name*`}
+                          error={touched.vgName && errors.vgName}
+                          onBlur={handleOnBlur}
+                        />
+                      </>
```
JBWatenbergScality: Maybe we can extract the fields to a <LVMAdditonnalFields /> component ? It would then make sense to validate those additional fields in that component and just add an onChange handler prop to that component that just addup the fields value to the submitted object.

ChengYanJin: Hi @JBWatenbergScality 
I am not sure of the benefit to extract these fields. For me, it would add some complexity than keep everything centralized. ( but I am open to discuss).

However, I see a reusable combined field  `sizeInputCombinedField` which is used in both `SPARSE_LOOP_DEVICE` and `LVM_LOGICAL_VOLUME`
https://github.com/scality/metalk8s/pull/3410/commits/8cd3fda6d7454bab3bcacfc926a20f0cb592c808

JBWatenbergScality (approved):

PR #3415, opened 2021-06-08
Make Control Plane Ingress IP configurable

### LINKED ISSUE #2381: Control plane Ingress controller is tied to Bootstrap
**Component**: salt, ingress

**Problem**:

**Resolution proposal**:

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

Being able to choose which IP will be used for Ingress Control Plane, so that user can manage them self HA of this IP through all Control Plane nodes.

**Summary**:

- Define Control Plane Ingress IP and Endpoint at a single place
- Male this Control Plane Ingress IP configurable in the Bootstrap Configuration
- Some docs and tests for this

---

Sees: #2381 

--- 

NOTE: It's only the first part of Control Plane Ingress HA, as we also want to add an optional MetalLB deployment as part of MetalK8s, so that if it's possible in the user environment HA of Control Plane Ingress can be managed by MetalLB directly.


salt/tests/unit/modules/files/test_metalk8s_network.yaml
```diff
@@ -83,3 +83,47 @@ routes:
...
+
+get_control_plane_ingress_ep:
+  # 1. Nominal
+  - cp_ingress_ip_ret: 1.1.1.1
+    result: https://1.1.1.1:8443
+
+  # 2. Error when getting IP
+  - cp_ingress_ip_ret: "Error because of banana"
+    raises: true
+    result: "Error because of banana"
```
alexandre-allard, line 137: This error happens to me all the time, I'm glad it's tested ! :+1:

salt/_modules/metalk8s_network.py
```diff
@@ -226,3 +226,31 @@ def routes():
...
+
+    if not isinstance(mine_ret, dict) or bootstrap_id not in mine_ret:
+        raise CommandExecutionError(
+            "Unable to get {} Control Plane IP: {}".format(bootstrap_id, mine_ret)
+        )
+
+    return mine_ret[bootstrap_id]
+
+
+def get_control_plane_ingress_ep():
```
gdemonet: I'd rather use full words (`get_control_plane_ingress_endpoint`), especially since `ip` and `ep` are really close so less distinguishable.

CHANGELOG.md
```diff
@@ -60,6 +60,12 @@
   LVMLogicalVolume volume in the UI
   (PR[#3410](https://github.com/scality/metalk8s/pull/3410))
 
+- [#2381](https://github.com/scality/metalk8s/issues/2381)) - Allow
+  to configure the external IP used for Control Plane Ingress, in
+  order for user to manage them self High Availability of this IP accross
+  all Control Plane nodes
```
alexandre-allard:
```suggestion
- [#2381](https://github.com/scality/metalk8s/issues/2381)) - Allow
  the configuration of the Control Plane Ingress' external IP, so that a
  user can manage himself the high availability of this IP on all the
  nodes of the control plane
```

docs/installation/bootstrap.rst
```diff
@@ -80,6 +81,15 @@ notation for it's various subfields.
...
         which we do not recommend for non-experts.
 
+      For ``controlPlane`` entry an ``ingressIP`` can also be provided, this
+      IP is the one that will be used to connect to control plane components,
+      like MetalK8s-UI and all the monitoring stack. We suggest to use
+      a `Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ that
+      will sit on a working master Node. The default value for this
+      Ingress IP is the control plane IP of the Bootstrap node (which means
+      that if you lose the Bootstrap node, you no longer have access to any
+      control plane conponent).
```
alexandre-allard:
```suggestion
      For ``controlPlane`` entry, an ``ingressIP`` can also be provided. This
      IP is used to connect to all the control plane components, like MetalK8s-UI
      and the whole monitoring stack. We suggest using
      a `Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ that
      will sit on a working master Node. The default value for this
      Ingress IP is the control plane IP of the Bootstrap node (which means
      that if you lose the Bootstrap node, you no longer have access to any
      control plane component).
```

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -0,0 +1,43 @@
...
+   ``<bootstrap_node_name>`` with the actual Bootstrap Node name):
+
+   .. parsed-literal::
+
+      $ kubectl exec -n kube-system -c salt-master \\
+          --kubeconfig=/etc/kubernetes/admin.conf \\
+          salt-master-<bootstrap_node_name> -- salt-run \\
+          state.orchestrate \\
+          metalk8s.orchestrate.update-control-plane-ingress-ip \\
+          saltenv=metalk8s-|version|
```
alexandre-allard: Couldn't we programmatically retrieve the bootstrap node name using Salt instead of this ? It would be easier to just have to copy/paste the command.

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -0,0 +1,43 @@
+Changing the Control Plane Ingress IP
+=====================================
+
+#. On the bootstrap node, update the ``ingressIP`` field from
+   ``network.controlPlane`` in the Bootstrap configuration file.
```
alexandre-allard:
```suggestion
   ``networks.controlPlane`` in the Bootstrap configuration file.
```

docs/operation/changing_control_plane_ingress_ip.rst
TeddyAndrieux: Sure we could even do it with kubectl but ... to be consistent I used the same approach as other doc sections, but indeed, I agree we should make all Salt-master commands easy to copy/paste.

Something like this:
```
kubectl exec -n kube-system -c salt-master $(kubectl get pod -l "app.kubernetes.io/name=salt-master" --namespace=kube-system -o jsonpath='{.items[0].metadata.name}') -- blabla
```

Hmm, maybe I will change this one so at least we start something :smile:

alexandre-allard (dismissed):

salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls
```diff
@@ -31,6 +30,7 @@ Create Control-Plane Ingress server private key:
     'nginx-ingress-control-plane.metalk8s-ingress.svc',
     'nginx-ingress-control-plane.metalk8s-ingress.svc.cluster.local',
     grains.metalk8s.control_plane_ip,
```
gdemonet: Should we keep the local IP, since it was there only to represent the Service IP?

tests/post/steps/test_ui.py
```diff
@@ -13,11 +9,9 @@ def test_ui(host):
 
 
 @then("we can reach the UI")
-def reach_UI(host):
-    ip = utils.get_grain(host, "metalk8s:control_plane_ip")
-
+def reach_UI(host, control_plane_ingress_ep):
     response = requests.get(
-        "https://{ip}:8443".format(ip=ip),
+        control_plane_ingress_ep,
```
gdemonet, line 14: Realizing this should depend on the value computed by CSC, but to make the test more robust, we could simply find the right host + path from the Ingress object we're testing. Not related to this PR though, I'll open a debt ticket.

docs/installation/services.rst
```diff
@@ -12,13 +12,14 @@ and can be used for operating, extending and upgrading a MetalK8s cluster.
...
-.. code-block:: shell
+.. code-block:: console
 
-    root@bootstrap $ salt-call grains.get metalk8s:control_plane_ip
-    local:
-        <the control plane IP>
+    root@bootstrap $ kubectl --kubeconfig=/etc/kubernetes/admin.conf \
+      get svc -n metalk8s-ingress ingress-nginx-control-plane-controller \
+      -o=jsonpath='{.spec.externalIPs[0]}{"\n"}'
+    <the ingress control plane IP>
```
gdemonet, line 22: :cry:

CHANGELOG.md
```diff
@@ -60,6 +60,12 @@
   LVMLogicalVolume volume in the UI
   (PR[#3410](https://github.com/scality/metalk8s/pull/3410))
 
+- [#2381](https://github.com/scality/metalk8s/issues/2381)) - Allow
+  the configuration of the Control Plane Ingress' external IP, so that a
+  user can manage himself the high availability of this IP on all the
+  nodes of the control plane
```
gdemonet: Maybe reword to:

> Allow configuring the Control Plane Ingress' external IP, to enable high availability with failover of this (virtual) IP between control plane nodes. This failover is not managed by MetalK8s.

docs/installation/bootstrap.rst
```diff
@@ -47,6 +47,7 @@ Configuration
       networks:
         controlPlane:
           cidr: <CIDR-notation>
+          ingressIP: <IP-for-ingress>
```
gdemonet: I would suggest we use two levels for this:

```yaml
networks:
  controlPlane:
    ingress:
      ip: <IP-for-ingress-controller-svc>
```

So we can add `ingress.host` (or `ingress.fqdn` :shrug:) later, and use it to sign certs and also in OIDC URLs (issuer URL, redirect URLs...).

gdemonet: Also, maybe the field should state `ingressController`? Since it's not "one Ingress object" in particular... I don't know, maybe it's obvious given it's part of the `networks` section :shrug:

docs/installation/bootstrap.rst
```diff
@@ -80,6 +81,15 @@ notation for it's various subfields.
...
         which we do not recommend for non-experts.
 
+      For ``controlPlane`` entry, an ``ingressIP`` can also be provided. This
+      IP is used to connect to all the control plane components, like
+      MetalK8s-UI and the whole monitoring stack. We suggest using
+      a `Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ that
+      will sit on a working master Node. The default value for this
+      Ingress IP is the control plane IP of the Bootstrap node (which means
+      that if you lose the Bootstrap node, you no longer have access to any
+      control plane component).
```
gdemonet, line 93: ...through the Ingress controller. But some components also listen on host ports directly, no? Especially K8s API, which is useful to access in case of Bootstrap issues.

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -0,0 +1,44 @@
...
+
+#. Reconfigure Control Plane conponents:
+
+   .. parsed-literal::
+
+      $ kubectl exec -n kube-system -c salt-master \\
+          --kubeconfig=/etc/kubernetes/admin.conf \\
+          $(kubectl --kubeconfig=/etc/kubernetes/admin.conf get pod \\
+          -l "app.kubernetes.io/name=salt-master" \\
+          --namespace=kube-system -o jsonpath='{.items[0].metadata.name}')  \\
```
gdemonet, line 38: `-o name`?

gdemonet (changes requested):
Few minor changes, most important IMO is the configuration format to build as future-proof as possible.

salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls
TeddyAndrieux: It's not needed indeed, I will remove it

tests/post/steps/test_ui.py
TeddyAndrieux: Hmm right this path is configurable from CSC :thinking:, indeed it will need some update to take this into account, but I will not do it in this PR, so yes if you can create a debt ticket for it :heart:

docs/installation/services.rst
TeddyAndrieux: It's just to check that it's the IP we want, you do not like it ?

docs/installation/bootstrap.rst
TeddyAndrieux: To me it's not really linked to the controller itself, it's like the IP you use to reach ("Ingress into") Control Plane stuffs.

For the extra layer, I'm not 100% convinced that we should do it now, I see the benefits, and what you suggest may make sense but it's not needed for the moment (and maybe it will never be, I don't know) so it makes the configuration file a bit more complex for ... no reason (at least for the moment) and changing the configuration file layout because we support new features later make sense at the end.

Honestly no strong opinion here, I can change it

docs/installation/bootstrap.rst
TeddyAndrieux: Right, but if you lose bootstrap you do not have access to OIDC so access to Kubernetes APIServer (should) do not work as well (if we generated kubeconfig specific to each user :wink:)
But agree, I will replace "control plane component" with `OIDC and various Control Plane UIs`

docs/operation/changing_control_plane_ingress_ip.rst
TeddyAndrieux: Hmm right I could use it :thinking: 
It will have strange behavior if for whatever reason you have multiple salt-master running, maybe better to keep this one for the moment (so that this procedure may be compatible with future Bootstrap HA :smile: :wink: )

tests/post/steps/test_ui.py
gdemonet: Done: #3421

docs/installation/services.rst
gdemonet: Most of our commands are getting super verbose (same for running a command in the salt-master)... Not saying it's wrong :wink:, just that it'd be nice if we could wrap those calls in some `kubectl` plugin... :innocent:

docs/installation/bootstrap.rst
gdemonet: > it's not really linked to the controller itself, it's like the IP you use to reach ("Ingress into") Control Plane

OK, that makes sense indeed.

> it makes the configuration file a bit more complex for ... no reason (at least for the moment) and changing the configuration file layout because we support new features later make sense at the end.

IMHO, the added complexity today (need to break a word into two, add a newline character) is ridiculous compared to the amount of code + doc changes we need when changing the config file schema.

docs/operation/changing_control_plane_ingress_ip.rst
gdemonet: Haha good point :ok_hand: We'll have to make these commands simpler anyway, so if we change how we deploy things, we don't have to update it in hundreds of places across our docs.

docs/installation/services.rst
TeddyAndrieux: Yess I agree

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -0,0 +1,44 @@
...
+
+#. Check that the change is taken into account.
+
+   .. code-block:: console
+
+      $ salt-call metalk8s_network.get_control_plane_ingress_ip
+      local:
+          <my-new-ip>
+
+#. On the bootstrap node, reconfigure ApiServer:
```
gdemonet: (consistency)

```suggestion
#. On the Bootstrap node, reconfigure apiserver:
```

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -0,0 +1,44 @@
...
+
+   .. parsed-literal::
+
+     $ salt-call state.sls \\
+         metalk8s.kubernetes.apiserver \\
+         saltenv=metalk8s-|version|
+
+#. Reconfigure Control Plane conponents:
```
gdemonet:
```suggestion
#. Reconfigure Control Plane components:
```

gdemonet (dismissed):
(minor typos) :rocket:

gdemonet (approved):

PR #3417, opened 2021-06-09
docs: Add design document for Control Plane Ingress HA

### LINKED ISSUE #2381: Control plane Ingress controller is tied to Bootstrap

TeddyAndrieux:
Design for Control Plane Ingress HA

---

Sees: #2381

### DIFF
--- a/docs/developer/architecture/control-plane-ingress.rst
+++ b/docs/developer/architecture/control-plane-ingress.rst
@@ -1,0 +1,204 @@
+Control Plane Ingress
+=====================
+
+Context
+-------
+
+Initially, Control Plane Ingress Controller was deployed using a DaemonSet with
+a ClusterIP service using Bootstrap Control Plane IP as External IP and then
+configuring all Control Plane components using this “External IP” (like OIDC
+and various UIs).
+
+So it means, we can only reach those components using Bootstrap Control
+Plane IP, and if, for whatever reason, Bootstrap is temporary down you can no
+longer access any UIs and you need to reconfigure all components manually,
+change the IP used everywhere to another Control Plane node IP, in order to
+restore access to Control Plane components, or if the Bootstrap is down
+permanently you need to restore the Bootstrap node.
+
+Here, we want to solve this issue and make Control Plane components Highly
+Available, so if you lose one node, including Bootstrap Node, you can still
+access various UIs.
+NOTE: In this document, we do not talk about the High Availability of the
+components itself but really only to the access through the Ingress
+(e.g.: We do not want to solve salt-master HA here).
+
+User Stories
+------------
+
+MetalK8s and Grafana UIs HA
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+I have a multi-node MetalK8s cluster, with at least 3 Control Plane nodes
+if I lose one of the Control Plane nodes (including the Bootstrap one)
+I can still access and authenticate on the MetalK8s and Grafana UIs.
+
+Design Choices
+--------------
+
+To have a proper HA Control Plane Ingress we want to use a
+`Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ using
+`MetalLB <https://metallb.universe.tf/>`_ so that we can rely on layer2
+ARP requests when possible.
+
+But in some network, it may be not possible to use this method
+so we also let the possibility to not use MetalLB but instead just
+assign an External IP, provided by the user, that we expect to be a
+Virtual IP, and we do not manage on our side but it’s managed by the user
+using whatever mechanism to switch this IP between Control Plane nodes.
+
+To summarize 2 different deployments possible depending on the user
+environment:
+
+- Using VIP managed by Layer2 ARP with MetalLB (builtin MetalK8s)
+- Using a user-provided IP that should switch between Control
+  Plane nodes (managed by the user)
+
+NOTE: In those 2 approaches we want the user to provide the Control
+Plane Ingress IP he wants to use.
+
+Rejected Design Choices
+-----------------------
+
+Manage Virtual IP by MetalK8s
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Instead of using MetalLB to manage the Virtual IP we could have manage this
+Virtual IP directly in MetalK8s with KeepAliveD (or any other “HA tool”) this
+way we really controle the Virtual IP.
+
+This approach was rejected because it seems to do not provide any real
+advantage compare to use MetalLB directly that will manage this Virtual IP
+for us and may provide a bunch of other useful feature for the future.
+
+Rely on DNS resolution
+~~~~~~~~~~~~~~~~~~~~~~
+
+Instead of using a Virtual IP we could rely on DNS resolution use this
+FQDN to configure Control Plane components. In this case we let the user
+configure his DNS to resolve on an IP of one Control Plane node.
+
+WIth this approach we let the DNS server handle the “High Availability” of
+the Control Plane UIs, basically if we lose the node that resolve the DNS
+then we expect the DNS server to switch to another IP of a working Control
+Plane node.
+
+This approach was rejected because it’s not a real HA as we expect DNS server
+to have some intelligence which it likely not the case in most of user
+environments.
+
+Use Nodes Control Plane IPs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Instead of using a Virtual IP or a FQDN we could still rely on Control Plane
+IPs and configure all Control Plane components using either relative path
+either all node Control Plane IPs. So that we can reach all Control Plane
+components using any Control Plane IP.
+
+With this approach we do not need any specific infrastructure on the user
+environment but it means every time we add a new Control Plane node we need
+to re-configure every Control Plane component that use Control Plane IPs but
+we also need to re-configure every Control Plane component when we remove a
+Control Plane node.
+
+This approach was rejected because Control Plane components we deploy today
+in MetalK8s does not seems to all support relative path and it does not
+seems to be possible to have proper HA without re-configuring some Control
+Plane components when we lose a Control Plane node.
+
+Non-Goals
+---------
+
+These points may be addressed later (or not) but in this document, we focus on
+a first simple deployment that (should) fit in most of the user environments.
+
+- Manage Workload Plane Ingress HA
+- Manage BGP routers
+
+Implementation Details
+----------------------
+
+Control Plane Ingress IP
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to configure all Control Plane components we need a single IP as
+Control Plane Ingress, so we expect the user to provide this Control Plane
+Ingress IP in the Bootstrap configuration file.
+To have some backward compatibility this Ingress IP is only mandatory when
+you use MetalLB and will default to Bootstrap Control Plane IP if not
+(so that we have the same behavior as before).
+
+This Control Plane Ingress IP can be changed at any time just by editing
+the Bootstrap configuration file and follow a simple documented procedure
+with some Salt states to reconfigure every component that needs to be
+re-configured.
+
+NOTE: Changing this Control Plane Ingress IP means we need to reconfigure
+all Kubernetes APIServer since we use this Ingress IP as an OIDC provider.
+
+MetalLB Configuration
+~~~~~~~~~~~~~~~~~~~~~
+
+MetalLB is not deployed in every environment so it needs to be enabled
+from the Bootstrap configuration file, that’s why we have a new field about
+MetalLB in the Control Plane network section.
+
+Today, we only allow MetalLB using Layer2 so we do not need to make MetalLB
+configuration configurable, if MetalLB is enabled in Bootstrap configuration
+MetalK8s will deploy the following configuration for MetalLB:
+
+.. code-block:: yaml
+
+    address-pools:
+    - name: control-plane-ingress-ip
+      protocol: layer2
+      addresses:
+      - <control-plane ingress ip>/32
+      auto-assign: false
+
+Same as the Control Plane Ingress IP, we can switch from non-MetalLB to MetalLB
+(and the opposite) at any time just by following the same procedure.
+
+Deployment
+~~~~~~~~~~
+
+As for every other addon in MetalK8s, we will use the MetalLB helm chart and
+render this one using a specific “option” file. But this one will not be always
+deployed as we only want to deploy it when a specific key is set in the
+Bootstrap configuration file, so in the Salt pillar at the end.
+
+When we use MetalLB we do not want to use the same NGINX Ingress Controller
+deployments, since MetalLB will be the entry point in the Kubernetes cluster
+we do not need to use a DaemonSet running on every Control Plane nodes,
+instead, we will use a Deployment with 2 replicas.
+
+We also need to configure the Service for Ingress Controller differently
+depending on if we use MetalLB or not when we use it we want to use a
+LoadBalancer service, set the LoadBalancerIP to IngressIP provided by
+the user and set externalTrafficPolicy to Local. If we do not use MetalLB
+then we use ClusterIP Service with IngressIP provided by the user as External
+IPs.
+
+It means the deployment of NGINX Ingress Controller depends on some Salt
+pillar values, also since we want to be able to switch between MetalLB and
+non-MetalLB we need to make sure the Salt states that deploy NGINX Ingress
+Controller remove no-longer-needed objects (e.g.: if you switch from
+non-MetalLB to MetalLB you want to remove the DaemonSet for NGINX Ingress
+Controller).
+
+Documentation
+-------------
+
+- Describe all new Bootstrap configuration fields
+- Add a simple procedure to change the Control Plane Ingress IP and
+  reconfigure all Control Plane conponents that need to.
+
+Test Plan
+---------
+
+Add some End-to-End tests in the CI:
+
+- Use MetalLB and a VIP as Control Plane Ingress IP
+- Test failover of MetalLB VIP
+- Change Control Plane Ingress IP using documented procedure
+- Switch from non-MetalLB to MetalLB using documented procedure
--- a/docs/developer/architecture/index.rst
+++ b/docs/developer/architecture/index.rst
@@ -10,6 +10,7 @@ Architecture Documents
    centralized-cli
    ci
    configurations
+   control-plane-ingress
    deployment
    logs
    metalk8s-ui


docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+I can still access and authenticate on the MetalK8s and Grafana UIs.
+
+Design Choices
+--------------
+
+To have a proper HA Control Plane Ingress we want to use a
+`Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ using
+`MetalLB <https://metallb.universe.tf/>`_ so that we can rely on layer2
+ARP requests when possible or let MetalLB advertise IPs using BGP when
+the user router support it.
```
NicolasT: Do we plan to support BGP-based address management?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+
+NOTE: MetalLB will be used only for Control Plane Ingress and not
+for Workload Plane.
```
NicolasT: Never ever?

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: I don't know, it's more a question for @thomasdanan 
But to me, since it's built-in metallb it's only some field to fill in the metallb config it's easy to "support" it

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Hmm at least it's not part of this specific design, I can change a bit this sentence

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+- Using a user-provided IP that should switch between Control
+  Plane nodes (managed by the user)
+
+NOTE: In all those 3 approaches we want the user to provide the Control
+Plane Ingress IP he wants to use.
+
+Rejected Design Choices
+-----------------------
+
+TODO
```
NicolasT: Please do before merging :wink:

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+re-configured.
+
+NOTE: Changing this Control Plane Ingress IP means we need to reconfigure
+all Kubernetes APIServer since we use this Ingress IP as an OIDC provider.
+
+MetalLB Configuration
+~~~~~~~~~~~~~~~~~~~~~
+
+MetalLB is not deployed in every environment so it needs to be configurable
+from the Bootstrap configuration file, that’s why we have a new field about
```
NicolasT, line 143: How is this done? What's for configuration format? How is this backwards-compatible? How can it be made forward-compatible?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+we will use a simple Layer2 configuration for MetalLB so that if the user
+environment allows using Layer2 protocol of MetalLB you just have to enable
+MetalLB, provide an Ingress IP for the Control Plane and that it.
+
+Default MetalLB configuration:
+
+.. code-block:: yaml
+
+    address-pools:
+    - name: ingress-ip
```
NicolasT: Let's find a more descriptive name.

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+      addresses:
+      - <control-plane ingress ip>/32
+      auto-assign: false
+
+But, since BGP router configuration can be really different between various
+environments, we need to make sure this
+`MetalLB configuration <https://metallb.universe.tf/configuration/#bgp-configuration>`_
+is exposed, that’s why we also have the ability to override default MetalLB
+configuration from the Bootstrap configuration file in order to provide BGP
+router information.
```
NicolasT: This implies we'd support BGP-based environments. Should we, now? Could we, later?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+When we use MetalLB we do not want to use the same NGINX Ingress Controller
+deployments, since MetalLB will be the entry point in the Kubernetes cluster
+we do not need to use a DaemonSet running on every Control Plane nodes,
+instead, we will use a Deployment with 2 replicas.
+
+We also need to configure the Service for Ingress Controller differently
+depending on if we use MetalLB or not when we use it we want to use a
+LoadBalancer service and set the LoadBalancerIP to IngressIP provided by
+the user. If we do not use MetalLB then we use ClusterIP Service with IngressIP
+provided by the user as External IPs.
```
NicolasT: No mention of `externalTrafficPolicy`?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+non-MetalLB we need to make sure the Salt states that deploy NGINX Ingress
+Controller remove no-longer-needed objects (e.g.: if you switch from
+non-MetalLB to MetalLB you want to remove the DaemonSet for NGINX Ingress
+Controller).
+
+Documentation
+-------------
+
+- Describe all new Bootstrap configuration fields (with some links
+  to MetalLB documentation to configure MetalLB to use BGP protocol)
```
NicolasT: So we'd directly expose (a subset of) the MetalLB configuration in `BootstrapConfiguration`? Not sure that's wise. What if we want to switch MetalLB to something else later? If it really required to have access to all of MetalLBs tunables? Is that a feature we provide?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,164 @@
...
+- Add a simple procedure to change the Control Plane Ingress IP and
+  reconfigure all Control Plane conponents that need to.
+
+Test Plan
+---------
+
+Add some End-to-End tests in the CI:
+
+- Use MetalLB and a VIP as Control Plane Ingress IP
```
NicolasT, line 201: Likely needs a VIP failover test scenario as well.

NicolasT (dismissed):

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: sure it's the plan :wink:

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: You can check #3418, basically, I have a `metallb` key in `networks:controlPlane` and in this dict I have `enabled` (which default to `false`)  and an optional `config` field that get merge with a default MetalLB configuration
https://github.com/scality/metalk8s/blob/a406960d634d9f6b06b0e0c770d3b673e1ae473f/salt/_pillar/metalk8s.py#L94-L119

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Sure, any suggestion? 
Maybe `metalk8s-control-plane-entrypoint-ip` ?

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: We can say we will do this later, just it's really easy to expose this MetalLB functionally, but maybe we do not want to expose it for the moment and do not talk about it at all

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Right, I will add it (btw it's done in implementation PR)

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Depends if we want to support BGP or not, if we only support Layer2 mode then we do not need to expose MetalLB config and just add "something" to enable or not MetalLB (talking about MetaLLB directly in doc/bootstrap config or not)

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Hmmm, right but not sure it's really easy to do, I don't know if we have an easy way to know where this VIP sit at any point in time (except doing an ARP request to get mac address and then checking the host having the network interface with this mac address)
Anyway I agree, I will add it here

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: But as I answered in another comment below, if we do not support BGP we do not need to expose config and we can also hide "metallb" from Bootstrap config and documentation, and just put `manageVIP: true/false` (or something else) in the bootstrap config

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: (Added)

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+Initially, Control Plane Ingress Controller was deployed using a DaemonSet with
+a ClusterIP service using Bootstrap Control Plane IP as External IP and then
+configuring all Control Plane components using this “External IP” (like OIDC
+and various UIs).
+
+So it means, we can only reach those components using Bootstrap Control
+Plane IP, and if, for whatever reason, Bootstrap is down you can no longer
+access any UIs and you need to reconfigure all components manually, change the
+IP used everywhere to another Control Plane node IP, in order to restore
+access to Control Plane components.
```
gdemonet: Or restore the Bootstrap :wink:

docs/developer/architecture/control-plane-ingress.rst
gdemonet: But then we need to test it and make it configurable... If most of our customer base and target audience for ARTESCA is OK with the L2 approach, I think we should try to keep it constrained. Once we get a clear use-case to address, we can add it easily using your suggested approach :ok_hand:

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+--------------
+
+To have a proper HA Control Plane Ingress we want to use a
+`Virtual IP <https://en.wikipedia.org/wiki/Virtual_IP_address>`_ using
+`MetalLB <https://metallb.universe.tf/>`_ so that we can rely on layer2
+ARP requests when possible or let MetalLB advertise IPs using BGP when
+the user router support it.
+
+NOTE: In this document we talk about MetalLB used for Control Plane
+Ingress and not for Workload Plane.
```
gdemonet: Maybe remind why: we would likely need to go with "real" load-balancing, which makes the current implementation of MetalLB problematic (can only manage one VIP per Service, IIRC).

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+Instead of using a Virtual IP we could rely on DNS resolution use this
+FQDN to configure Control Plane components. In this case we let the user
+configure his DNS to resolve on an IP of one Control Plane node.
+
+WIth this approach we let the DNS server handle the “High Availability” of
+the Control Plane UIs, basically if we lose the node that resolve the DNS
+then we expect the DNS server to switch to another IP of a working Control
+Plane node.
+
+This approach was rejected because it’s not a real HA and we expect DNS server
```
gdemonet: > not a real HA

Meaning...? Something along the lines of "transparent for clients", which isn't the case for DNS-based failover?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+you use MetalLB and will default to Bootstrap Control Plane IP if not
+(so that we have the same behavior as before).
+
+This Control Plane Ingress IP can be changed at any time just by editing
+the Bootstrap configuration file and follow a simple documented procedure
+with some Salt states to reconfigure every component that needs to be
+re-configured.
+
+NOTE: Changing this Control Plane Ingress IP means we need to reconfigure
+all Kubernetes APIServer since we use this Ingress IP as an OIDC provider.
```
gdemonet, line 137: this will change with oauth2-proxy I think, but we'll need to reconfigure the proxy anyway :innocent:

docs/developer/architecture/control-plane-ingress.rst
gdemonet: `control-plane-ingress-ip`?

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+
+As for every other addon in MetalK8s, we will use the MetalLB helm chart and
+render this one using a specific “option” file. But this one will not be always
+deployed as we only want to deploy it when a specific key is set in the
+Bootstrap configuration file, so in the Salt pillar at the end.
+
+When we use MetalLB we do not want to use the same NGINX Ingress Controller
+deployments, since MetalLB will be the entry point in the Kubernetes cluster
+we do not need to use a DaemonSet running on every Control Plane nodes,
+instead, we will use a Deployment with 2 replicas.
```
gdemonet, line 173: Won't this cause unnecessary mesh networking? I guess it's not super critical for control plane, but I don't really see the benefit of switching to a Deployment.

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,210 @@
...
+
+We also need to configure the Service for Ingress Controller differently
+depending on if we use MetalLB or not when we use it we want to use a
+LoadBalancer service, set the LoadBalancerIP to IngressIP provided by
+the user and set externalTrafficPolicy to Local. If we do not use MetalLB
```
gdemonet, line 178: If we use `Local`, why don't we set up the Ingress controller pods to run locally to each node? What happens if no replica runs locally?

gdemonet (commented):
##### TL;DR

Generally OK :ok_hand: with the design.
Simply questioning the deployment strategy change for the Ingress controller, with MetalLB in place (maybe I misunderstood something).
Also suggesting that we relegate BGP support to a later (optional) iteration.

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: True, but it's only in case of bootstrap node down for ever, I will add it

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: It may be transparent if the DNS server is active and properly switch from an IP to another in case of failure, but as is it's a "False HA" since we do not have any HA we just move the "problem": "we do not need the IP to be HA but a specific DNS"

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Likely yes, but since we do not have any design for oauth2-proxy yet I do not mention it

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Yes I agree, I can remove it for the moment and stay on Layer2 only. 
Maybe @xaf-scality you have an opinion on this one as well ?

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Not sure it's worth talking about it here, it's really the design for "Control Plane Ingress"

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: It's less ressources, here we stick with 2 replicas for the full cluster (no matter the number of Control Plane nodes).
To me it's more like, "why should we use a DaemonSet when we can use a Deployment ?" 
I think @NicolasT also had other arguments

docs/developer/architecture/control-plane-ingress.rst
TeddyAndrieux: Hmm to me, when you use Local anyway the VIP will sit on one node that has some "pod" of the Service running locally,so it cannot happens

docs/developer/architecture/control-plane-ingress.rst
gdemonet: OK, then maybe promote this clear distinction to a "Goals vs. Non-goals" section? A note in the "Design Choices" seems too late to me.

docs/developer/architecture/control-plane-ingress.rst
gdemonet: Ah OK, I misunderstood this behaviour then.

docs/developer/architecture/control-plane-ingress.rst
gdemonet: OK OK, I would just vote for lowering the priority of this switch:

- it's an architecture change, so it has an impact on users willing to operate/monitor our stuff
- it's not a huge win (an idle Ingress controller doesn't take much resources)

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,204 @@
...
+
+This approach was rejected because Control Plane components we deploy today
+in MetalK8s does not seems to all support relative path and it does not
+seems to be possible to have proper HA without re-configuring some Control
+Plane components when we lose a Control Plane node.
+
+Non-Goals
+---------
+
+Those points may be addresses later (or not) but in this document, we focus on
```
gdemonet:
```suggestion
These points may be addressed later (or not) but in this document, we focus on
```

docs/developer/architecture/control-plane-ingress.rst
```diff
@@ -0,0 +1,204 @@
...
+with some Salt states to reconfigure every component that needs to be
+re-configured.
+
+NOTE: Changing this Control Plane Ingress IP means we need to reconfigure
+all Kubernetes APIServer since we use this Ingress IP as an OIDC provider.
+
+MetalLB Configuration
+~~~~~~~~~~~~~~~~~~~~~
+
+MetalLB is not deployed in every environment so it needs to be activatable
```
gdemonet:
```suggestion
MetalLB is not deployed in every environment so it needs to be enabled
```

gdemonet (dismissed):
(minor typos) :shipit:

gdemonet (approved):

Issue #2966, opened 2020-12-08
Vendoring of Helm charts should be automated with "buildchain" targets


gdemonet:
**Component**: build, charts

**Why this is needed**:

Everytime one wants to update a vendored Helm chart, a manual process is required to:
- find the source of the Helm chart, to further issue a `helm fetch $chart_src`
- re-apply patches to these charts in-tree
- re-render the right SLS from the chart and its options using the right command-line (which can be found in the last commit message updating said SLS...)

This is cumbersome and error-prone.

**What should be done**:

Add `doit` targets for:

- fetching and vendoring a Helm chart (with source + version stored in the `buildchain` package), applying patches as required (they would need to be managed differently than today, I guess)
- rendering the corresponding SLS formula(s)

gdemonet:
One approach to consider: instead of patching the vendored chart + applying some magic in our render script, try to "just" run `helm template` and use `kustomize` to modify the resulting YAML, and finally convert this intermediate YAML into a proper SLS.
Also, make sure we store all intermediate files in the `_build` directory for easier inspection.
If all this works, we could even consider **not** vendoring Helm charts at all.

gdemonet:
Relates to #1449

PR #3418, opened 2021-06-11
Control Plane Ingress IP managed by MetalLB

### LINKED ISSUE #2381: Control plane Ingress controller is tied to Bootstrap

TeddyAndrieux:
**Component**:

'salt', 'network'

**Context**: 

Allow MetalK8s to manage Control Plane Ingress IP with MetalLB

**Summary**:

- Ability to enabled MetalLB deployment from Bootstrap config
- If MetalLB enabled deploy MetalLB and Control Plane Ingress Controller using a Deployment and a LoadBalancer service
- If MetalLB disabled do not deploy MetalLB and deploy Control Plane Ingress Controller same way as before this PR
- Allow switching from MetalLB to non-MetaLB (and the opposite)

---

Fixes: #2381


docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -1,8 +1,16 @@
...
 
-#. On the bootstrap node, update the ``ingressIP`` field from
-   ``networks.controlPlane`` in the Bootstrap configuration file.
+This procedure allow to change the Control Plane Ingress IP and also to move
+from MetalLB configuration to non-MetalLB configuration and the opposite.
+
+.. note::
+
+  If you switch from MetalLB to non-MetalLB this procedure do **not** remove
+  MetalLB installation, it just do not use it for MetalK8s components
```
TeddyAndrieux: It's debatable btw

buildchain/buildchain/salt_tree.py
```diff
@@ -425,7 +425,10 @@ def _get_parts(self) -> Iterator[str]:
     Path("salt/metalk8s/addons/nginx-ingress-control-plane/certs/init.sls"),
     Path("salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls"),
     Path("salt/metalk8s/addons/nginx-ingress-control-plane/deployed/init.sls"),
-    Path("salt/metalk8s/addons/nginx-ingress-control-plane/deployed/chart.sls"),
+    Path(
+        "salt/metalk8s/addons/nginx-ingress-control-plane/deployed/",
+        "chart-daemonset.sls",
+    ),
```
gdemonet, line 432: Wondering if this is the right approach... Is the DaemonSet "rendered chart" really different from the Deployment one? Or is it mostly about the `ds` vs. `deploy` manifests? In which case we could simply "drop" this manifest when rendering, and keep it in a separate file, making the `chart.sls` formula only about common parts... WDYT?

salt/metalk8s/addons/metallb/deployed/namespace.sls
```diff
@@ -0,0 +1,10 @@
+#! metalk8s_kubernetes
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: metalk8s-metallb
```
gdemonet: Should we really tie the namespace name to the implementation choice?
I think we should try to follow the same principle as for other addons, so either `metalk8s-loadbalancing` or simply reusing `metalk8s-ingress`, since that's where we use such load balancing.

CHANGELOG.md
```diff
@@ -66,6 +66,11 @@
   nodes. This failover is not managed by MetalK8s.
   (PR[#3415](https://github.com/scality/metalk8s/pull/3415))
 
+- [#2381](https://github.com/scality/metalk8s/issues/2381)) - Allow
+  MetalK8s to manage Control Plane Ingress Virtual IP using MetalLB if
+  it possible in the user environment
+  (PR[#3418](https://github.com/scality/metalk8s/pull/3418))
+
```
gdemonet: Let's merge those two, because "failover is not managed by MetalK8s" is not true anymore :wink: 

```suggestion
  nodes (PR[#3415](https://github.com/scality/metalk8s/pull/3415)).
  If supported by the user environment, MetalK8s can manage fail-over
  of this virtual IP using [MetalLB](https://metallb.universe.tf/)
  (PR[#3418](https://github.com/scality/metalk8s/pull/3418)).

```

charts/ingress-nginx-control-plane-deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  metrics:
+    enabled: true
+    serviceMonitor:
+      enabled: true
+      additionalLabels:
+        metalk8s.scality.com/monitor: ''
+
+defaultBackend:
+  enabled: false
```
gdemonet: Why do we disable the default backend now?

docs/installation/bootstrap.rst
```diff
@@ -49,6 +49,8 @@ Configuration
           cidr: <CIDR-notation>
           ingress:
             ip: <IP-for-ingress>
+          metalLB:
+            enabled: true
```
```suggestion
            enabled: <boolean>
```

docs/installation/bootstrap.rst
```diff
@@ -92,6 +94,14 @@ notation for it's various subfields.
       that if you lose the Bootstrap node, you no longer have access to any
       control plane component).
 
+      This ``ip`` for ``ingress`` can be managed by MetalK8s directly if
+      it's possible in your environment, to do so we use
```
gdemonet, line 98: We may want to explain what "kind of environment" will support this (physical isolated networks, for instance), or at least known limitations (e.g. virtual networks with PortSecurity will not allow arbitrary ARP requests).

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -1,6 +1,14 @@
 Changing the Control Plane Ingress IP
 =====================================
 
+This procedure allow to change the Control Plane Ingress IP and also to move
+from MetalLB configuration to non-MetalLB configuration and the opposite.
```
gdemonet:
```suggestion
This procedure describes how to change the Control Plane Ingress IP, and
to enable (or disable) MetalLB management of this IP.
```

docs/operation/changing_control_plane_ingress_ip.rst
gdemonet: Fine by me. Let's wait for someone to report this as an issue before we do anything.

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -1,6 +1,14 @@
...
+
+.. note::
+
+  If you switch from MetalLB to non-MetalLB this procedure do **not** remove
+  MetalLB installation, it just do not use it for MetalK8s components
```
gdemonet:
```suggestion
  Disabling MetalLB using this procedure does **not** remove MetalLB,
  it simply disables its use for managing the ``LoadBalancer`` *Service*
  for MetalK8s Control Plane Ingress.
```

docs/operation/changing_control_plane_ingress_ip.rst
```diff
@@ -18,6 +26,15 @@ Changing the Control Plane Ingress IP
...
           <my-new-ip>
+      $ salt-call pillar.get networks:control_plane
+      local:
+        ----------
+        cidr:
+          - <control-plane-cidr>
+        ingress:
+          ip:
+            <my-new-ip>
+        metalLB: <my-new-metallb-config>
```
gdemonet: Shouldn't this reflect the same as what we set in BootstrapConfiguration?

```suggestion
        metalLB:
          enabled: <true | false>
```

gdemonet (commented):
Initial review, will go over the tests after lunch :meat_on_bone:

salt/metalk8s/defaults.yaml
```diff
@@ -84,9 +84,18 @@ networks:
...
+          Control plane MetalLB speaker metrics (only if MetalLB enabled)
+      control_plane_ip:7946:
+        expected: speaker
+        description: >-
+          Control plane MetalLB speaker (only if MetalLB enabled)
       ingress_control_plane_ip:8443:
         expected: kube-proxy
-        description: Control plane nginx ingress
+        description: >-
+          Control plane nginx ingress (only if MetalLB disabled)
```
gdemonet: Don't we always need this port to be open (even if MetalLB is enabled)?

tests/post/features/ingress.feature
```diff
@@ -19,12 +19,36 @@ Feature: Ingress
...
+        When we stop the node hosting the Control Plane Ingress VIP
+        Then the node hosting the Control Plane Ingress VIP changed
+        And we are able to login to Dex as 'admin@metalk8s.invalid' using password 'password'
+
     Scenario: Change Control Plane Ingress IP to node-1 IP
         Given the Kubernetes API is available
         And we are on a multi node cluster
         And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
-        When we update control plane ingress IP to node 'node-1' IP
+        When we disable MetalLB and set control plane ingress IP to node 'node-1' IP
```
gdemonet: Let's try not to do two actions in the same `when`. Also, I think this should be a `given`, in this case:

```gherkin
Given MetalLB is disabled
...
When we set control plane ingress IP to node 'node-1' IP
...
```

tests/post/features/ingress.feature
```diff
@@ -19,12 +19,36 @@ Feature: Ingress
...
         And we are able to login to Dex as 'admin@metalk8s.invalid' using password 'password'
+
+    Scenario: Change Control Plane Ingress IP to a new VIP
+        Given the Kubernetes API is available
+        And MetalLB is enabled
+        And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
+        # In our CI we need a private network to use this "VIP" feature,
+        # assume when Control Plane network is equal to "192.168.1.0/24" that we are
+        # on a private network
+        And the control plane cidr is equal to '192.168.1.0/24'
```
gdemonet: Mmh, not super fond of this approach, where the test scenario depends on the execution context :confused: 

Could we instead define:

- a command-line option for pytest to pass a VIP for control plane
- a `given` saying "a VIP for control plane is available", which checks if the VIP correctly belongs to the control plane CIDR (if needed, not sure), and stores its value in a fixture
- a `when` saying "we set control plane ingress IP to the available VIP", which uses the aforementioned fixture

?

tests/post/steps/test_ingress.py
```diff
@@ -125,6 +204,37 @@ def check_cp_ingress_node_ip(control_plane_ingress_ip, node_name, ssh_config):
...
+    )
+
+    assert events.items, "Unable to get event for Control Plane Ingress Service"
+
+    match = re.search(
+        r'announcing from node "(?P<node>.+)"',
+        sorted(events.items, key=lambda event: event.last_timestamp, reverse=True)[
+            0
+        ].message,
+    )
```
gdemonet: Is this robust enough? I'm wondering if we may get other events matching the field selectors that won't match this regular expression.

tests/post/steps/test_sanity.py
```diff
@@ -69,6 +69,33 @@ def test_statefulset_running(host):
...
+
+# }}}
+# Given {{{
+
+
+@given(parsers.parse("MetalLB is {state}"))
+def is_metalb_enabled(host, state):
+    expected = state == "enabled"
+    if expected != utils.get_pillar(host, "networks:control_plane:metalLB:enabled"):
+        pytest.skip("We skip as we run this test only if MetalLB {}".format(state))
```
gdemonet, line 96: Can't we have this `given` change the current configuration if it doesn't match what's expected (and restore after the test), instead of skipping?

I guess we could add a command-line flag to disable MetalLB tests if it's really required.

gdemonet (changes requested):
I think we can be more flexible on the tests, but overall approach LGTM :ok_hand:

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Today they are not really different (mainly Service ClusterIP vs LoadBalancer and Deployment vs DaemonSet) but at some point, it may have some logic in the helm chart that makes the resulting "chart" really different.

I agree that we could have something smarter here to do not duplicate things in different SLS files but to me, with the current status of helm chart management, we have in MetalK8s this approach is the simpler/better(? maybe) one

salt/metalk8s/addons/metallb/deployed/namespace.sls
TeddyAndrieux: Hmm indeed, you right I will rename it

CHANGELOG.md
TeddyAndrieux: It's still true, you can still use one VIP you manage your self you have both features, so to me, it deserves 2 changelog entry.
It make sense no ?

charts/ingress-nginx-control-plane-deployment.yaml
TeddyAndrieux: Hmmm likely for no good reason just copied the one from DaemonSet but indeed we can enable it

docs/installation/bootstrap.rst
TeddyAndrieux: No no no you must enable it !! It's mandatory (joke)
Sure I will change it

docs/installation/bootstrap.rst
TeddyAndrieux: Right, I will try to add something for this

docs/operation/changing_control_plane_ingress_ip.rst
TeddyAndrieux: Right I forgot to update this one (basically in older version of this PR MetalLB config was there also) I will update it

salt/metalk8s/defaults.yaml
TeddyAndrieux: ~Indeed good catch~

Edit: Oh no sorry it's not needed, host point of view, when we use MetalLB since it's managed by iptables directly so ... it needs to be open on various firewalls but we do not need to have this specific port free on the host.
But Indeed documentation point of view it's weird to have this, I will remove this "(only if MetalLB disabled)"

tests/post/features/ingress.feature
TeddyAndrieux: To me "given" was more "given environment is in this specific state" (basically checking the state of the environment but do not change anything) and "when" more like "when we do this specific action" ... "then we are in this state"
It's not right ?

The split is a bit difficult because you do not want to disable MetalLB and keep the ingress IP to a ... non-existing VIP. To me it's a single action "I switch from MetalLB to non-MetalLB using Node-1 IP" (maybe I should reword it a bit :thinking: )

tests/post/features/ingress.feature
TeddyAndrieux: > VIP correctly belongs to the control plane CIDR (if needed, not sure)

It's not strictly needed we just need the IP to be routed on the Control Plane nodes

But right I can likely make something better with a variable but it means we need to pass this variable to run this test where, with the current approach, we always run this test if we use MetalLB but ... ok I will do with a variable it cleaner

tests/post/steps/test_ingress.py
TeddyAndrieux: Since we filter on this specific service with only "reason=nodeAssigned" to me it's safe and I do not see any better approach for this (already tried relying directly on ARP but it's a mess since sometimes you may run these tests on the machine hosting the VIP currently, so you do not have the mac address since it's not resolved locally in arp table ... )

tests/post/steps/test_sanity.py
TeddyAndrieux: Yes we could

TeddyAndrieux: Oh no sorry I read too quickly, this one cannot change it's in `test_sanity` so to me here we only want to check the running pods according to the configuration (which is what we do here, the check is different between if you have MetalLB or not).

CHANGELOG.md
gdemonet: Mmh, maybe we can keep the two entries, but I would at least remove the "failover is not managed by MetalK8s" part from the first one.

buildchain/buildchain/salt_tree.py
gdemonet: OK, let's indeed reconsider later (linking this discussion to #2966).

charts/metallb/templates/speaker/psp.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if and .Values.psp.create .Values.speaker.psp.create -}}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
```
gdemonet, line 3: Something to keep in mind for the future (we likely want to open a ticket to track this, not only for MetalLB but for all charts we embed): PodSecurityPolicy is deprecated starting in K8S 1.21, and will be removed in 1.25. See https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

eve/main.yml
```diff
@@ -442,13 +446,15 @@ models:
         TEST_HOSTS_LIST: "bootstrap"
         PYTEST_FILTERS: "post and ci"
         BOOTSTRAP_BACKUP_ARCHIVE: ""
+        CONTROL_PLANE_INGRESS_VIP: "192.168.1.253"
```
gdemonet, line 449: Shouldn't it be:

```suggestion
        CONTROL_PLANE_INGRESS_VIP: "192.168.1.254"
```

, as defined in BootstrapConfig above?

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
```diff
@@ -311,6 +311,80 @@ replace_object:
...
+      spec:
+        type: LoadBalancer
+        clusterIP: "10.11.12.13"
+        healthCheckNodePort: "12345"
+
+  # Replace service object (special case: health_check_node_port set in the manifest
+  # and old_object, keep the manifest one)
+  # NOTE: This test just ensure we keep the one from input manifest and not
+  #       replace it with the one from old_object, but in practice apiServer
+  #       will reject this call since health_check_node_port is immutable field
```
gdemonet, line 352: Maybe our client code should prevent it, in that case?

tests/post/features/ingress.feature
gdemonet: > To me "given" was more "given environment is in this specific state" (basically checking the state of the environment but do not change anything) and "when" more like "when we do this specific action" ... "then we are in this state"
> It's not right ?

In my mind, there's nothing preventing a "given" from performing an action (e.g. "Given I have a sample pod running" will schedule such a pod). The idea behind the distinction between "given" and "when" is to describe what is a "pre-condition" for some feature/operation to work (given), and what is "the feature/operation under test" (when).

> The split is a bit difficult because you do not want to disable MetalLB and keep the ingress IP to a ... non-existing VIP. To me it's a single action "I switch from MetalLB to non-MetalLB using Node-1 IP" (maybe I should reword it a bit :thinking:)

IMO, the "operation under test" here is really about **manual change of the Control Plane Ingress IP**, not **disabling MetalLB**.
If we agree on this, I would define a "given" that ensures (potentially applying changes) we have disabled MetalLB and we don't specify any VIP for Control Plane Ingress (leaving it to the default behaviour). Then, we are in the same situation to run the scenario as before this PR, so the "when" can stay the same. Finally, the "given" will handle a _teardown_ step once the scenario completes, restoring the configuration to what it was before.

Now, **iff** we want to also test the **disable MetalLB** feature, then it likely deserves another scenario, with other checks (like verifying that MetalLB is still deployed, but not doing anything, and that the old VIP is no longer reachable).

tests/post/features/ingress.feature
```diff
@@ -19,12 +19,32 @@ Feature: Ingress
...
         And we wait for the rollout of 'daemonset/ingress-nginx-control-plane-controller' in namespace 'metalk8s-ingress' to complete
         And we wait for the rollout of 'deploy/dex' in namespace 'metalk8s-auth' to complete
         Then the control plane ingress IP is equal to node 'node-1' IP
         And we are able to login to Dex as 'admin@metalk8s.invalid' using password 'password'
+
+    Scenario: Change Control Plane Ingress IP to a new VIP
+        Given the Kubernetes API is available
+        And a VIP for Control Plane Ingress is available
+        And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
+        When we enable MetalLB and set control plane ingress IP to '{new_cp_ingress_vip}'
```
gdemonet, line 46: Same here IMO, are we testing the **VIP change while MetalLB is enabled**, or **enabling MetalLB**? We shouldn't do both in the same scenario.

tests/post/steps/test_ingress.py
gdemonet: > Since we filter on this specific service with only "reason=nodeAssigned" to me it's safe

OK :ok_hand: 

I guess we could have been slightly more conservative with e.g.

```suggestion
    match = None
    for event in sorted(events.items, key=lambda event: event.last_timestamp, reverse=True):
        match = re.search(r'announcing from node "(?P<node>.+)"', event.message)
        if match is not None:
            break
```

tests/post/steps/test_sanity.py
gdemonet: Indeed, good point, for these scenarios we shouldn't change things (we could document these tests for open-source users wanting to validate their deployment, akin to [`helm test`](https://helm.sh/docs/topics/chart_tests/)).

eve/main.yml
TeddyAndrieux: Nop, it's the goal here, it's to change the VIP to something else

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
TeddyAndrieux: Hmm no I prefer that we get APIServer response so that if one day this constraint get removed from APIServer we are good

tests/post/features/ingress.feature
TeddyAndrieux: It does not make sense to just "enable MetalLB" you **must** provide a VIP when you enable MetalLB

CHANGELOG.md
TeddyAndrieux: Hmm right, ok I didn't read the suggestion you put (my bad) it right, I can do something like you suggest

tests/post/steps/test_ingress.py
TeddyAndrieux: Right, we can do this also, not sure it brings any value today but ... it does not cost much so I will do

tests/post/features/ingress.feature
TeddyAndrieux: I see your point we should have a test that "Given MetalLB is disabled then we switch to MetalLB using this IP" and a test "Given MetalLB is enabled then we change the VIP"
But to me it does not bring any value since all of this is already tested as part of other tests (including teardown of tests) so it will just make the test last longer for ... no reason

charts/metallb/templates/speaker/psp.yaml
gdemonet: Opened #3437

eve/main.yml
gdemonet: I see, OK.

salt/tests/unit/modules/files/test_metalk8s_kubernetes.yaml
gdemonet: Makes sense

tests/post/features/ingress.feature
gdemonet: OK, agreed, the "VIP change while MetalLB is enabled" is tested in the teardown of this test (that's only true because we start the test suite from a state where MetalLB is enabled, but let's admit that's enough for now).

I think this scenario should be slightly updated to reflect the intent better:

```diff
-    Scenario: Change Control Plane Ingress IP to a new VIP
+    Scenario: Enable MetalLB
         Given the Kubernetes API is available
+        And MetalLB is disabled
         And a VIP for Control Plane Ingress is available
         And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
         When we enable MetalLB and set control plane ingress IP to '{new_cp_ingress_vip}'
         And we wait for the rollout of 'deploy/metallb-controller' in namespace 'metalk8s-loadbalancing' to complete
         And we wait for the rollout of 'daemonset/metallb-speaker' in namespace 'metalk8s-loadbalancing' to complete
         And we wait for the rollout of 'deploy/ingress-nginx-control-plane-controller' in namespace 'metalk8s-ingress' to complete
         And we wait for the rollout of 'deploy/dex' in namespace 'metalk8s-auth' to complete
         Then the control plane ingress IP is equal to '{new_cp_ingress_vip}'
         And we are able to login to Dex as 'admin@metalk8s.invalid' using password 'password'
```

TeddyAndrieux: I can do this sure, the test will then really be "Enable MetalLB" I totally agree but it will last longer (which is not really an issue I think) and also if we do what you suggest we will have those tests covered in ingress

If MetalLB enabled initially:
1- Switch to non-MetalLB + Change CP Ingress IP + (teardown) switch to MetalLB
2- Switch to non-MetalLB + Switch to MetalLB + (teardown) change MetalLB IP
If MetalLB disabled initially:
1- Change CP Ingress IP + (teardown) change again CP Ingress IP
2- Switch to MetalLB + (teardown) switch back to non MetalLB

Where with my current approach:
 
If MetalLB enabled initially:
1- Switch to non-MetalLB + Change CP Ingress IP + (teardown) switch to MetalLB
2- Change MetalLB IP + (teardown) change MetalLB IP
If MetalLB disabled initially:
1- Change CP Ingress IP + (teardown) change again CP Ingress IP
2- Switch to MetalLB + (teardown) switch back to non MetalLB

So ... the list of what is tested is the same at the end but I agree you approach is likely cleaner, I will change it :+1:

gdemonet (approved):
Awesome, thanks again @TeddyAndrieux :rocket:!

PR #3443, opened 2021-07-09
shell: Add defaultDexConnector option to automatically trigger login on a predefined Dex connector

JBWatenbergScality:
**Component**:

shell

**Context**: 

Dex is providing the possibility to use several connectors to federate OIDC providers. However under some conditions we may need to force the usage of a specific connector.

**Summary**:

This PR add a `defaultDexConnector` option to enable this.

This option can be added to the auth section of micro application `runtime-app-configuration`.
It intercept the `authorization_endpoint` of the oidc web finger and append it the `connector_id` query parameter.

The connector choice can be forced by adding the `displayLoginChoice` query parameter to the current URL.
This enable to maintain UI login possible when the default connector is unavailable.

Additionaly as Dex is not supporting `end_session_endpoint` the logout is reviewed to clean the cookies of the
domain and clearing the localStorage entries of oidc client when `providerLogout` is set to true.

**Acceptance criteria**: 

We shall be able to login directly using the default dex connector when the property is defined

### DIFF
--- a/shell-ui/src/auth/AuthProvider.js
+++ b/shell-ui/src/auth/AuthProvider.js
@@ -116,28 +116,45 @@ export function useLogOut() {
   let auth;
   try {
     auth = useOauth2Auth();
-  } catch(e) {
+  } catch (e) {
     //If an exception is raised here it is likely because the app is not using OIDC auth kind, so we can ignore this
-    console.log('Failed to retrieve auth informations for OIDC auth kind', e)
+    console.log('Failed to retrieve auth informations for OIDC auth kind', e);
   }
 
-  return {logOut: useCallback(() => {
-    if (!authConfig) {
-      return;
-    }
-  
-    if (authConfig.kind === 'OAuth2Proxy') {
-      throw new Error('OAuth2Proxy authentication kind is not yet supported');
-    }
-    if (auth && auth.userManager) {
-      auth.userManager.revokeAccessToken();
-      if (authConfig.providerLogout) {
-        auth.userManager.signoutRedirect();
-      } else {
-        auth.userManager.removeUser();
-        location.reload();
+  return {
+    logOut: useCallback(() => {
+      if (!authConfig) {
+        return;
       }
-    }
-  }, [JSON.stringify(authConfig), auth])}
-  
+
+      if (authConfig.kind === 'OAuth2Proxy') {
+        throw new Error('OAuth2Proxy authentication kind is not yet supported');
+      }
+      if (auth && auth.userManager) {
+        auth.userManager.revokeAccessToken();
+        if (authConfig.providerLogout) {
+          auth.userManager.signoutRedirect().catch((e) => {
+            if (e.message === 'no end session endpoint') {
+              console.log(
+                "OIDC provider doesn't support end session endpoint, fallback to clearing document cookies",
+              );
+              document.cookie.split(';').forEach(function (c) {
+                document.cookie =
+                  c.trim().split('=')[0] +
+                  '=;' +
+                  'expires=Thu, 01 Jan 1970 00:00:00 UTC;';
+              });
+            } else {
+              console.error(e);
+            }
+            auth.userManager.removeUser();
+            location.reload();
+          });
+        } else {
+          auth.userManager.removeUser();
+          location.reload();
+        }
+      }
+    }, [JSON.stringify(authConfig), auth]),
+  };
 }
--- a/shell-ui/src/auth/AuthProvider.js
+++ b/shell-ui/src/auth/AuthProvider.js
@@ -9,7 +9,7 @@ import {
 } from 'oidc-react';
 import { useShellConfig } from '../initFederation/ShellConfigProvider';
 import { getUserGroups } from '../navbar/auth/permissionUtils';
-import { WebStorageStateStore } from 'oidc-client';
+import { MetadataService, WebStorageStateStore } from 'oidc-client';
 
 export function AuthProvider({ children }: { children: Node }) {
   const { authConfig } = useAuthConfig();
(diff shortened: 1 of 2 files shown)


shell-ui/src/auth/AuthProvider.js
```diff
@@ -39,6 +57,9 @@ function OAuth2AuthProvider({ children }: { children: Node }) {
     loadUserInfo: true,
     automaticSilentRenew: true,
     monitorSession: false,
+    MetadataServiceCtor: authConfig.defaultDexConnector
```
ChengYanJin, line 60: I am wondering should we add this `defaultDexConnector` prop to type `OIDCConfig`?

shell-ui/src/auth/AuthProvider.js
```diff
@@ -39,6 +57,9 @@ function OAuth2AuthProvider({ children }: { children: Node }) {
...
+      ? defaultDexConnectorMetadataService(authConfig.defaultDexConnector)
```
ChengYanJin, line 61: It's more a question: 
If we want to use Dex to authenticate the users against a connector-based interface, how should we set the `defaultDexConnector` in `runtime-app-configuration`?

shell-ui/src/auth/AuthProvider.js
```diff
@@ -95,28 +116,45 @@ export function useLogOut() {
...
+            if (e.message === 'no end session endpoint') {
+              console.log(
+                "Oidc provider doesn't support end session endpoint, fallback to clearing document cookies",
+              );
+              document.cookie.split(';').forEach(function (c) {
+                document.cookie =
+                  c.trim().split('=')[0] +
+                  '=;' +
+                  'expires=Thu, 01 Jan 1970 00:00:00 UTC;';
+              });
```
gdemonet, line 146: What's that? We split the cookie and assign a new value for each? How does it work? Is it that assigning a string to `document.cookie` is actually interpreted as an instruction, rather than describing the whole "cookies state"?

shell-ui/src/auth/AuthProvider.js
```diff
@@ -95,28 +116,45 @@ export function useLogOut() {
...
+      if (authConfig.kind === 'OAuth2Proxy') {
+        throw new Error('OAuth2Proxy authentication kind is not yet supported');
+      }
+      if (auth && auth.userManager) {
+        auth.userManager.revokeAccessToken();
+        if (authConfig.providerLogout) {
+          auth.userManager.signoutRedirect().catch((e) => {
+            if (e.message === 'no end session endpoint') {
+              console.log(
+                "Oidc provider doesn't support end session endpoint, fallback to clearing document cookies",
```
gdemonet:
```suggestion
                "OIDC provider doesn't support end session endpoint, fallback to clearing document cookies",
```

shell-ui/src/auth/AuthProvider.js
```diff
@@ -95,28 +116,45 @@ export function useLogOut() {
...
+                  c.trim().split('=')[0] +
+                  '=;' +
+                  'expires=Thu, 01 Jan 1970 00:00:00 UTC;';
+              });
+            } else {
+              console.error(e);
+            }
+            auth.userManager.removeUser();
+            location.reload();
+          });
```
gdemonet: This whole `.catch` block doesn't seem related to the change in this PR (which is about a pre-defined Dex connector). Would be nice to at least use a different commit :wink:

shell-ui/src/auth/AuthProvider.js
gdemonet: Didn't we start documenting all this rework of shell-ui? Would be useful IMO, both now for reviewers catching up with the progress, and in the future for maintenance / onboarding.

gdemonet (changes requested):
No need to tackle documentation as part of this PR :wink:

shell-ui/src/auth/AuthProvider.js
JBWatenbergScality: Yes using `document.cookie` we can only set or update a single cookie at a time (https://developer.mozilla.org/en-US/docs/web/api/document/cookie#write_a_new_cookie). What I'm trying to achieve here is a hack to logout the users from the federated OIDC Provider behind Dex because Dex is not supporting the end session endpoint and then the session on the original OIDC provider can't be closed in any other way 😕 .

shell-ui/src/auth/AuthProvider.js
JBWatenbergScality: Done in db1812c1d0d9a5d1e21a20caefd003b31b9a1a75

shell-ui/src/auth/AuthProvider.js
JBWatenbergScality: `defaultDexConnector` should take the connector id as defined in dex configuration.
For example given the following Dex connectors configuration :
```
       connectors:
            - type: oidc
              id: keycloak
              name: Keycloak
              config:
                   //....
```

Providing `keycloak` as `defaultDexConnector` in `runtime-app-configuration` will automatically redirect the user to this connector.

JBWatenbergScality: I've opened #3476 regarding the documentation and will address it as soon as possible

shell-ui/src/auth/AuthProvider.js
JBWatenbergScality: Done in e6566809c79b0f9ea3cb11fa81d856096bb04d06

shell-ui/src/auth/AuthProvider.js
JBWatenbergScality: Done in e6566809c79b0f9ea3cb11fa81d856096bb04d06

gdemonet (approved):

PR #3447, opened 2021-07-13
Upgrade: Best effort drain when upgrading single node cluster

### LINKED ISSUE #3445: Single node "best-effort drain" during upgrade
**Component**:

'lifecycle'

**What happened**:

Due to a "bug" (:question:) in kubelet when upgrading a single node cluster with a bunch of Pods running, upgrade fail because it takes too much time to restart static Pods (like apiserver).

See: https://github.com/kubernetes/kubernetes/issues/103658

**What was expected**:

Single node upgrade to work properly

In order to avoid that kind of issue let's drain the node during the upgrade even if it's not really needed since it's .... single-node cluster so anyway every services may/will have downtime during the upgrade process.

Draining may not be possible on a single node as you do not have any other node to schedule needed Pods (e.g.: If you have some PodDisruptionBudget configured in the cluster).

Add a "best-effort" drain used for single-node drain during the upgrade, this drain just cordon the node and evict all Pods possible, and do not retry/fail if one Pod cannot be evicted just continue the "classic upgrade process"

TeddyAndrieux:
**Component**:

'lifecycle'

**Context**: 

See: #3445 

**Summary**:

When upgrading a cluster with a single control plane node, first do a "best-effort" drain that will evict all Pods but does not retry if some Pods cannot be evicted instead continue the "classic" upgrade process

NOTE: We do not explicitly uncordon the node at the end, as it's automatically handled by the "deploy_node" orchestrate

---

Fixes: #3445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,12 @@
   aggregation of solutions UIs
   (PR[#3414](https://github.com/scality/metalk8s/pull/3414))
 
+### Bug fixes
+- [#3445](https://github.com/scality/metalk8s/issues/3445) - Avoid
+  kube-apiserver timeout during single node cluster upgrade when a
+  lot of pod ran on the node
+  (PR[#3447](https://github.com/scality/metalk8s/pull/3447))
+
 ### Breaking changes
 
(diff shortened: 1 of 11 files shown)


sayf-eddine-scality (approved):
lgtm

PR #3459, opened 2021-07-20
ci: Fix mismatch in "publish" stage names

gdemonet:
See: #3406

NOTE: we can't use it anyway, but this mismatch currently breaks all post-merge runs.


alexandre-allard (approved):
:facepalm:

PR #3465, opened 2021-07-29
Improvement: 3093 Dashboard Services panel

### LINKED ISSUE #3093: MetalK8s UI Overview page Services Panel
**Component**: UI, monitoring

**Why this is needed**:
We want to display the health of the grouped services.

**What should be done**:

image

~~We would require the final design from @Cuervino @thomasdanan about the interaction, more specifically, when hovering on the status what should we display on the tooltip.~~
~~-  We currently blocked by the group alert feature. (so this ticket is only about UI layout)~~
(Edited: Not anymore blocked)
~~- This component should include a layout property and stay independent to be reused by other applications at runtime.~~
(Edited: We don't require this anymore since we will import the alerthooks from shell UI through the federated hook)

Interaction design:
https://www.figma.com/proto/N2jOAK44dTSiBclDjXzz6a/Next_Gen_Design?node-id=301%3A116&viewport=1035%2C-840%2C0.23489561676979065&scaling=min-zoom&page-id=180%3A3495

We need to restyle the tooltip in core-ui to meet the interaction design. 

**Test plan**:
TBD

alexis-ld:
**Component**: dashboard

**Context**:  #3093 

**Summary**: Adds a Services panel to the dashboard

https://user-images.githubusercontent.com/3480526/127501863-1663e0bd-936c-4812-8854-4f1b2bc2557b.mov

Closes: #3093

### DIFF
--- a/ui/src/components/DashboardServices.js
+++ b/ui/src/components/DashboardServices.js
@@ -1,0 +1,322 @@
+// @flow
+import React from 'react';
+import styled from 'styled-components';
+import { Link } from 'react-router-dom';
+import { useIntl } from 'react-intl';
+import { Tooltip } from '@scality/core-ui';
+import {
+  spacing,
+  fontSize,
+  fontWeight,
+} from '@scality/core-ui/dist/style/theme';
+
+import type { Alert } from '../services/alertUtils';
+import { PageSubtitle } from '../components/style/CommonLayoutStyle';
+import {
+  useAlertLibrary,
+  useHighestSeverityAlerts,
+  highestAlertToStatus,
+} from '../containers/AlertProvider';
+import CircleStatus from './CircleStatus';
+import {
+  STATUS_WARNING,
+  STATUS_CRITICAL,
+  STATUS_SUCCESS,
+  STATUS_HEALTH,
+} from '../constants.js';
+
+const ServiceItems = styled.div`
+  display: flex;
+  flex-direction: column;
+  padding: ${spacing.sp4};
+`;
+
+const ServiceItemLabelWrapper = styled.div`
+  display: flex;
+  align-items: baseline;
+`;
+
+const ServiceItemLabel = styled.div`
+  margin-left: ${spacing.sp8};
+`;
+
+const ServiceItemElement = styled.div`
+  padding: ${spacing.sp4};
+`;
+
+const ClickableServiceItemElement = styled(ServiceItemElement)`
+  width: 100%;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  :hover {
+    background-color: ${(props) => props.theme.highlight};
+  }
+`;
+
+const NonHealthyServiceItemElement = styled.div`
+  cursor: pointer;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+
+  a {
+    text-decoration: none;
+    color: inherit;
+    width: 100%;
+    display: flex;
+    align-items: center;
+  }
+`;
+
+const NonHealthyPopUp = styled.div`
+  display: flex;
+  flex-direction: column;
+  font-size: ${fontSize.base};
+
+  label {
+    width: 25%;
+    margin-right: ${spacing.sp8};
+    color: ${(props) => props.theme.textSecondary};
+    text-align: right;
+  }
+`;
+
+const NonHealthyPopUpTitle = styled.div`
+  font-weight: ${fontWeight.bold}
+  text-align: center;
+`;
+
+const NonHealthyPopUpItem = styled.div`
+  width: 100%;
+  display: flex;
+  margin: ${spacing.sp4} ${spacing.sp14};
+`;
+
+const NonHealthyPopUpSeverity = styled.div`
+  color: ${(props) => {
+    const theme = props.theme;
+    let color;
+
+    switch (props.status) {
+      case STATUS_HEALTH:
+      case STATUS_SUCCESS:
+        color = theme.statusHealthy;
+        break;
+      case STATUS_WARNING:
+        color = theme.statusWarning;
+        break;
+      case STATUS_CRITICAL:
+        color = theme.statusCritical;
+        break;
+      default:
+        color = theme.statusHealthy;
+    }
+    return color;
+  }};
+`;
+
+const ClickableIcon = styled.i`
+  self-align: flex-end;
+`;
+
+const ServiceItem = ({
+  label,
+  status,
+  alerts,
+}: {
+  label: string,
+  status: string,
+  alerts: Alert[],
+}) => {
+  let date = null;
+
+  const getStartDate = (alerts: Alert[]) => {
+    // Check we only have alerts and set startsAt properties
+    const filtered = alerts.filter((item) => item && item.startsAt);
+    // Sort to make sure we have oldest alert first
+    filtered.sort((a, b) => new Date(a.startsAt) - new Date(b.startsAt));
+
+    if (filtered[0] && filtered[0].startsAt) return filtered[0].startsAt;
+    return '';
+  };
+
+  if (alerts.length) date = new Date(getStartDate(alerts));
+
+  if (!alerts.length && status === STATUS_HEALTH)
+    return (
+      <ServiceItemElement aria-label={label}>
+        <ServiceItemLabelWrapper>
+          <CircleStatus status={status} />
+          <ServiceItemLabel>{label}</ServiceItemLabel>
+        </ServiceItemLabelWrapper>
+      </ServiceItemElement>
+    );
+  else
+    return (
+      <NonHealthyServiceItemElement>
+        <Tooltip
+          // Right placement to avoid Z index issues between left sidebar and tooltip or out of screen tooltip displays
+          placement="top"
+          overlayStyle={{
+            // em sizing to handle font-size change on large screens
+            width: '20em',
+            height: '100%',
+          }}
+          overlay={
+            <NonHealthyPopUp>
+              <NonHealthyPopUpTitle>View details</NonHealthyPopUpTitle>
+              <NonHealthyPopUpItem>
+                <label>Severity</label>
+                <NonHealthyPopUpSeverity status={status}>
+                  {status}
+                </NonHealthyPopUpSeverity>
+              </NonHealthyPopUpItem>
+              {
+                // Checking date validity
+                date instanceof Date && !isNaN(date) && (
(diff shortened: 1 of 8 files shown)


ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+        break;
+      case STATUS_CRITICAL:
+        color = theme.statusCritical;
+        break;
+      default:
+        color = theme.statusHealthy;
+    }
+    return color;
+  }};
+`;
```
ChengYanJin: could be removed

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+            height: '100%',
+          }}
+          overlay={
+            <NonHealthyPopUp>
+              <NonHealthyPopUpTitle>View details</NonHealthyPopUpTitle>
+              <NonHealthyPopUpItem>
+                <label>Severity</label>
+                <NonHealthyPopUpSeverity status={status}>
+                  {status}
+                </NonHealthyPopUpSeverity>
```
ChengYanJin: Could benefit from the New `Text` component  in core-ui.  You may want to check the `statusText` in storybook

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+          <CircleStatus status={status} />
+          <ServiceItemLabel>{label}</ServiceItemLabel>
+        </ServiceItemLabelWrapper>
+      </ServiceItemElement>
+    );
+  else
+    return (
+      <NonHealthyServiceItemElement>
+        <Tooltip
+          // Right placement to avoid Z index issues between left sidebar and tooltip or out of screen tooltip displays
```
ChengYanJin, line 123: can be removed since we already fixed the issue ;)

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+  self-align: flex-end;
+`;
+
+const ServiceItem = ({
+  label,
+  status,
+  alerts,
+}: {
+  label: string,
+  status: string,
```
ChengYanJin: I believe the type of status can be more specific. 

We have already typed the Status in `AlertProvider`
> export type Status = 'healthy' | 'warning' | 'critical';

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+  alerts: Alert[],
+}) => {
+  let date = null;
+
+  const getStartDate = (alerts: Alert[]) => {
+    // Check we only have alerts and set startsAt properties
+    const filtered = alerts.filter((item) => item && item.startsAt);
+    // Sort to make sure we have oldest alert first
+    filtered.sort((a, b) => new Date(a.startsAt) - new Date(b.startsAt));
```
ChengYanJin: it's good thinking, however, I don't think it's necessary ;)
In theory, at the current time, the same alert should not be triggered more than once.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+      </NonHealthyServiceItemElement>
+    );
+};
+
+const DashboardServices = () => {
+  const intl = useIntl();
+  const alertsLibrary = useAlertLibrary();
+
+  // K8s Master
+  const k8sAlerts = useHighestSeverityAlerts(
```
ChengYanJin: In this specific case, we only gather two alerts: "K8sMasterServicesAtRisk" and "K8sMasterServicesDegraded"
and then we filter out the lower severity, so in the end, we should have at most one alert or an empty array. 
```suggestion
  const k8sHighestSeverityAlert = useHighestSeverityAlerts(
```

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+`;
+
+const ServiceItem = ({
+  label,
+  status,
+  alerts,
+}: {
+  label: string,
+  status: string,
+  alerts: Alert[],
```
ChengYanJin, line 106: IMO, we should have at most one alert, please see the explanation below :)
suggestion:
`alert: Alert,`

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+          // Right placement to avoid Z index issues between left sidebar and tooltip or out of screen tooltip displays
+          placement="top"
+          overlayStyle={{
+            // em sizing to handle font-size change on large screens
+            width: '20em',
+            height: '100%',
+          }}
+          overlay={
+            <NonHealthyPopUp>
+              <NonHealthyPopUpTitle>View details</NonHealthyPopUpTitle>
```
ChengYanJin: I think we should translate this as well.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+              <NonHealthyPopUpItem>
+                <label>Severity</label>
```
ChengYanJin: missing translation.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+                <NonHealthyPopUpSeverity status={status}>
+                  {status}
+                </NonHealthyPopUpSeverity>
+              </NonHealthyPopUpItem>
+              {
+                // Checking date validity
+                date instanceof Date && !isNaN(date) && (
+                  <NonHealthyPopUpItem>
+                    <label>Start</label>
```
ChengYanJin: ideally, we should translate start as well.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+                    <div>
+                      {`${date.toISOString().split('T')[0]}
+              ${date.toISOString().split('T')[1].slice(0, 8)}`}
```
ChengYanJin: I think second is not necessary.
so you could benefit from`formatDateToMid1` in utils.js

ui/src/components/__TEST__/util.js
```diff
@@ -58,15 +62,17 @@ const AllTheProviders = ({ children }) => {
...
-    <IntlProvider locale="en" messages={translations_en}>
-      <Provider store={store}>
-        <QueryClientProvider client={queryClient}>
-          <AlertProvider>
-            <ThemeProvider theme={theme}>{children}</ThemeProvider>
-          </AlertProvider>
-        </QueryClientProvider>
-      </Provider>
-    </IntlProvider>
+    <Router history={history}>
```
ChengYanJin, line 65: can you please explain to me why we need this Wrapper?

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,322 @@
...
+} from '@scality/core-ui/dist/style/theme';
+
+import type { Alert } from '../services/alertUtils';
+import { PageSubtitle } from '../components/style/CommonLayoutStyle';
+import {
+  useAlertLibrary,
+  useHighestSeverityAlerts,
+  highestAlertToStatus,
+} from '../containers/AlertProvider';
+import CircleStatus from './CircleStatus';
```
ChengYanJin, line 21: I thought we already move it core-ui, but apparently not.. ;) 
Not the scope for this PR will do it in another PR.

ChengYanJin (changes requested):

ui/src/components/__TEST__/util.js
alexis-ld: @ChengYanJin , because we use react-router-dom's `Link` in the component, without the Router wrapper this makes the test suite fail because react-router component cannot be rendered outside a Router context.

Here we create an in-memory fake history to be provided to the test suite.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,295 @@
...
+              <NonHealthyPopUpTitle>
+                {intl.formatMessage({ id: 'view_details' })}
+              </NonHealthyPopUpTitle>
+              <NonHealthyPopUpItem>
+                <label>{intl.formatMessage({ id: 'severity' })}</label>
+                <StatusText status={status}>{status}</StatusText>
+              </NonHealthyPopUpItem>
+              {
+                // Checking date validity
+                date && (
```
ChengYanJin:
```suggestion
                alert[0].startsAt && (
```
if we want to make sure we have a startsAt property, we can check here.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,295 @@
...
+  const alertsLibrary = useAlertLibrary();
+
+  // K8s Master
+  const k8sHighestSeverityAlert = useHighestSeverityAlerts(
+    alertsLibrary.getK8SMasterAlertSelectors(),
+  );
+  const k8sStatus = highestAlertToStatus(k8sHighestSeverityAlert);
+
+  // Bootstrap
+  const bootstrapAlerts = useHighestSeverityAlerts(
```
ChengYanJin:
```suggestion
  const bootstrapHighestSeverityAlert = useHighestSeverityAlerts(
```
same for the rest...

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,295 @@
...
+  label: string,
+  status: Status,
+  alerts: Alert[],
+}) => {
+  let date = null;
+  const intl = useIntl();
+
+  const getStartDate = (alerts: Alert[]) => {
+    // Check we only have alerts and set startsAt properties
+    const filtered = alerts.filter((item) => item && item.startsAt);
```
ChengYanJin: I think the filter is not necessary here, since what you want is only a check to make sure we do have the startsAt property.

ui/src/components/DashboardServices.js
```diff
@@ -0,0 +1,295 @@
...
+  if (!alerts.length && status === STATUS_HEALTH)
+    return (
+      <ServiceItemElement aria-label={label}>
+        <ServiceItemLabelWrapper>
+          <CircleStatus status={status} />
+          <ServiceItemLabel>{label}</ServiceItemLabel>
+        </ServiceItemLabelWrapper>
+      </ServiceItemElement>
+    );
+  else
```
ChengYanJin, line 119: if you want to be super strict,  but I think it's not necessary.. 
```suggestion
  else if(alerts.length === 1  && status !== STATUS_HEALTH)
```
and in the end, if we are not in any of the two scenarios, 
else{
 console.error(` The Service ${label} status has encountered an expected scenario.`)
}

ChengYanJin (approved):
LGTM!

PR #3469, opened 2021-08-02
shell: Improve module federation performance

JBWatenbergScality:
**Component**:

Shell-UI

**Context**: 

When switching between two federated apps in the navbar a loading screen is currently displayed.

image

**Summary**:

Around half of the time of this loading time is dedicated to loading remoteEntry chunk of the federated app. In order to improve navigation performance and reduce loading screen time, in this PR we prefetch remoteEntry chunks of apps for which we display a link in the navbar. The preloading is done in parallel of painting the current page and hence the impact on the app first paint time is negligible.

**Acceptance criteria**: 

Loading screens while browsing federated apps are displayed for a shorter time.
In a dev env, where chunks are not optimised for production, we can see a 2x improvement on time spent on loading screen compared to the previous performance test using production optimised chunks. 

image

### DIFF
--- a/shell-ui/src/FederatedApp.jsx
+++ b/shell-ui/src/FederatedApp.jsx
@@ -8,6 +8,7 @@ import {
   lazy,
   Suspense,
   type Node,
+  useRef,
   useMemo,
 } from 'react';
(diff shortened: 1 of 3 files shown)


ChengYanJin (approved):
LGTM!

Issue #2829, opened 2020-10-05
Default "cluster" selected in Grafana etcd dashboard is incorrect (loki-headless)


alexandre-allard:
**Component**: salt, prometheus-operator, grafana

**What happened**:
When you go to the etcd dashboard in Grafana, there is a dropdown list to select the etcd cluster, the first element selected by default is `loki-headless` which obviously is not an etcd cluster, so all the charts are empty.
You have to click on the list and select the right cluster `kube-etcd` to make it works.

![Selection_004](https://user-images.githubusercontent.com/31347867/95067627-66269400-0704-11eb-9ecd-8c7711913231.png)

**What was expected**:
When I go to this page, `kube-etcd` is already selected and I don't have non-existing cluster in the list.

**Steps to reproduce**:
Install a 2.6.x, then go to the Grafana etcd dashboard.

**Resolution proposal** (optional):
The issue comes from the query used to get the list of etcd cluster which is:
```
label_values(etcd_server_has_leader, job)
```
The problem is the loki `/metrics` endpoint also exposes an `etcd_server_has_leader` metric.
So, we need to tune the query to only retrieve etcd clusters.
For this, we need to extract the etcd dashboard from prometheus-operator rendered chart as it's already done for some objects we customize (e.g. Prometheus rules).

gdemonet:
This issue relates to https://github.com/grafana/loki/issues/3262

In MetalK8s 2.10+, we introduced a new version of this etcd dashboard, which hides the `cluster` template variable. The aforementioned issue makes the dashboard unusable (because `loki-headless` is selected by default).

We have three options:

- we can drop these metrics from the recording rule
- we can filter out the "loki-headless" cluster from the Grafana dashboard, and re-enable the `cluster` variable
- we can hardcode the "kube-etcd" cluster in the Grafana dashboard

PR #3491, opened 2021-08-12
Fix etcd dashboard (empty stats, missing "cluster" variable)

gdemonet:
**Component**: grafana, charts

**Context**: #2829 (this PR addresses https://github.com/scality/metalk8s/issues/2829#issuecomment-896856442 specifically, and targets 2.10+ only)

**Summary**:

- Refactor vendoring of dashboards (move to `salt/metalk8s/addons` instead of `charts`)
- Refactor patching of dashboards embedded in charts and extract UID checks to `tests/post`
- Fix the "etcd" dashboard, by filtering out "loki" metrics, and pre-selecting "kube-etcd"

**Acceptance criteria**:

- "etcd" dashboard displays metrics for the "kube-etcd" cluster on first load
- "etcd" dashboard allows changing the selected cluster to watch other deployed and monitored etcd clusters
- "etcd" dashboard never shows the "loki-headless" cluster in the available options

---

Closes: #2829


salt/metalk8s/addons/logging/deployed/dashboards.sls
```diff
@@ -0,0 +1,21 @@
...
+          'name': 'logs',
+          'title': 'Logs',
+          'tags': ['logging'],
+      },
+      {
+          'name': 'loki',
+          'title': 'Loki',
+          'tags': ['logging'],
+      },
+]) }}
```
TeddyAndrieux, line 21: I'm ok with the move of the dashboard closer to the SLS since it does not come from the chart, but any reason to regroup all those logging dashboards in a single sls ?
Just wondering if we may want in some case to deploy/upgrade only a part of the logging stack, but right, today I do not see any reason to do so :thinking:

charts/grafana_dashboard_patches.json
```diff
@@ -0,0 +1,17 @@
...
+    "uid": "3e97d1d02672cdd0861f4c97c64f89b2"
+  },
+  "USE Method / Node": {
+    "uid": "fac67cfbe174d3ef53eb473d73d9212f"
+  },
+  "Nodes": {
+    "uid": "fa49a4706d07a042595b664c87fb33ea"
+  },
+  "Prometheus / Overview": {
+    "uid": "prometheus-overview"
```
TeddyAndrieux: Not link to this PR, but wondering if we shouldn't put "explicit" UID for all dashboards

TeddyAndrieux (approved):
LGTM

salt/metalk8s/addons/logging/deployed/dashboards.sls
gdemonet: It was mostly to keep things simple, but I could have kept the previous separation, you're right. Just thought that we would always need to keep the full `metalk8s.addons.logging` in sync (since fluent-bit and loki need to talk to each other).

charts/grafana_dashboard_patches.json
gdemonet: Yeah, wondered about the same thing. It could simplify the UI links, at least :)

PR #3499, opened 2021-08-16
salt: Fix shell-ui configuration upgrade

### LINKED ISSUE #3496: Upgrade overwrites some sections of Shell UI configuration
**Component**: salt, shell-ui, csc

**What happened**: On a platform where Shell UI was re-configured to include more UIs and map users to new groups, an upgrade of MetalK8s triggered a reset for some sections of this configuration.

**What was expected**: Upgrade should maintain user configurations.

**Steps to reproduce**: Add a role in the `spec.userGroupsMapping[admin@metalk8s.invalid]` key of the `metalk8s-ui/shell-ui-config` ConfigMap, and trigger an upgrade. The added role is removed.

**Resolution proposal** (optional): Implement a smarter merge algorithm when combining user-provided config with embedded defaults.

JBWatenbergScality:
**Component**:

salt, shell-ui

**Context**: 

When upgrading from 2.9.x to 2.10.x we lost patches to OIDC and users groups mapping configuration. 

**Summary**:

This solves it by copying the old configuration to the new CMs.

**Acceptance criteria**: 

Tested an upgrade from 2.9 with patched configuration to 2.10.3-dev and the configuration was correctly kept.

Closes: #3496

### DIFF
--- a/salt/metalk8s/addons/ui/deployed/ui-configuration.sls
+++ b/salt/metalk8s/addons/ui/deployed/ui-configuration.sls
@@ -67,8 +67,40 @@ Create metalk8s-ui-config ConfigMap:
 
 {%- else %}
 
+  {%- if metalk8s_shell_ui_config is not none %}
+    {%- set config_data = metalk8s_shell_ui_config.data['config.yaml'] | load_yaml %}
+
+    {%- if config_data.apiVersion == 'addons.metalk8s.scality.com/v1alpha1' and 'oidc' in config_data.spec %}
+
+Create metalk8s-ui-config ConfigMap:
+  metalk8s_kubernetes.object_present:
+    - manifest:
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: metalk8s-ui-config
+          namespace: metalk8s-ui
+        data:
+          config.yaml: |-
+            apiVersion: addons.metalk8s.scality.com/v1alpha2
+            kind: UIConfig
+            spec:
+              kind: "OIDC"
+              {{ config_data.spec.oidc | yaml(False) | indent(14) }}
+
+    {%- else %}
+
 metalk8s-ui-config ConfigMap already exist:
   test.succeed_without_changes: []
+
+    {%- endif %}
+
+  {%- else %}
+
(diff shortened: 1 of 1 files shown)


salt/metalk8s/addons/ui/deployed/ui-configuration.sls
```diff
@@ -67,9 +67,41 @@ Create metalk8s-ui-config ConfigMap:
...
+
+    {%- else %}
+
+metalk8s-ui-config ConfigMap already exist:
+  test.succeed_without_changes: []
+
+    {%- endif %}
+
+  {%- else %}
+
```
gdemonet, line 99: Not sure to understand this part:

- we have both `metalk8s-ui-config` and `metalk8s-shell-ui-config` already existing
- we want to copy over the OIDC settings from the latter to the former, in case the `shell-ui` config is in the old format

But if that's true, we're overwriting the previous `metalk8s-ui-config`, without checking its own apiVersion or previous contents (what if we changed the basePath?)

salt/metalk8s/addons/ui/deployed/ui-configuration.sls
```diff
@@ -252,6 +284,30 @@ Convert old Metalk8s Shell UI ServiceConfiguration to new format:
...
+                {{ config_data.spec.favicon | yaml(False) | indent(16) }}
+              {%- endif %}
+              {%- if 'canChangeLanguage' in config_data.spec %}
+              canChangeLanguage:
+                {{ config_data.spec.canChangeLanguage | yaml(False) | indent(16) }}
+              {%- endif %}
+              {%- if 'canChangeTheme' in config_data.spec %}
+              canChangeTheme:
+                {{ config_data.spec.canChangeTheme | yaml(False) | indent(16) }}
+              {%- endif %}
```
gdemonet, line 310: That's really too verbose, don't you think? Would be good to reimplement the `service_config_updated` state module we built in another project (:wink:)...

gdemonet (changes requested):
Can we add some rendering test cases, to clarify the expected behaviour?

Issue #3487, opened 2021-08-09
Bump Kubernetes to 1.22.X impossible with current python-kubernetes version


TeddyAndrieux:
**Component**:

'build'

**What happened**:

Currently, our salt module to interact with APIServer use python-kubernetes but python-kubernetes is a bit late on nowadays Kubernetes version (still at Kubernetes 1.18 (in beta) when creating this ticket, Kubernetes 1.22.0 just came out).

https://github.com/kubernetes-client/python#compatibility

Due to this old python-kubernetes version some APIs/Models are not available, like `Ingress` as part of `networking.k8s.io/v1` API and since Kubernetes 1.22.0 `Ingress` Beta APIs get removed so with the current Python-kubernetes version we cannot manage `Ingress` objects when calling APIServer 1.22.0+.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22

**What was expected**:

Our salt module interacting with APIServer should be compatible with the Kubernetes version embedded as part of MetalK8s.

**Resolution proposal** (optional):

TBD
Several solutions possible (non ordered and maybe not exhaustive):

1. Wait for a new python-kubernetes version compatible with 1.22.0 (and likely contribute to python-kubernetes upstream to help catch up on upstream Kubernetes versions)
2. Use "pure HTTP query" in our salt module (basically get rid of python-kubernetes and do the call directly to APIServer with the right header, body, ...) that will potentially need to be maintained for future Kubernetes versions
3. Use kubectl in our salt module so that we do not have anything to maintain except if kubectl interface changes (that should **NOT** happen a lot)

TeddyAndrieux:
Solution 3. seems like the best choice IMHO

PR #3506, opened 2021-08-20
docs: Add design docs for "interaction with k8s from salt"

TeddyAndrieux:
Refs: #3487

### DIFF
--- a/docs/developer/architecture/interaction-with-k8s-from-salt.rst
+++ b/docs/developer/architecture/interaction-with-k8s-from-salt.rst
@@ -1,0 +1,140 @@
+Interaction with Kubernetes API from Salt
+=========================================
+
+Context
+-------
+
+In MetalK8s we need to interact with Kubernetes API from Salt in order to
+create and edit various objects. Since the Salt Kubernetes module is not
+flexible enough (at least today) we decided to build our own Salt module that
+will cover all we need in MetalK8s. This Salt module relies on
+`python-kubernetes`_ lib so that we have a proper python class for every
+Kubernetes object and we can validate the object content without any call to
+the actual Kubernetes API.
+
+But `python-kubernetes`_ lib is a bit late on Kubernetes versions (at the
+time of writing, the latest `python-kubernetes`_ is for k8s-1.18 and k8s
+1.22.1 just came out) so it means that some Kubernetes objects from the
+latest versions cannot be managed by this python lib.
+
+Here, we want to solve this issue and make sure when a new Kubernetes
+version comes out that we can easily support interaction with all the
+Kubernetes API from Salt.
+
+Design Choices
+--------------
+
+Instead of relying on model and API object from `python-kubernetes`_, which
+is the representation from ``openapi`` of a specific Kubernetes version, we
+can directly use `python-base`_ which is what is used in the back by
+`python-kubernetes`_.
+
+This `python-base`_ has a ``DynamicClient`` that can be used to manage any
+Kubernetes objects no matter the Kubernetes version. It’s really more
+flexible it’s mean that we do not validate the manifest we send to
+Kubernetes API, but the manifest can be validated using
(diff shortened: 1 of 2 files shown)


docs/developer/architecture/interaction-with-k8s-from-salt.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Contribute to python-kubernetes Lib
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We could contribute to python-kubernetes in order to make it compatible with
+the latest versions of Kubernetes.
+
+This approach was rejected because python-kubernetes start to be really late
+on Kubernetes versions, and this work, likely, needs to be done for every new
+Kubernetes version.
```
NicolasT: This would still be my preference. Yes, it adds a bit of work, but that's work we could/should take on/help out with upstream, for everyone's benefit. Resorting to invoking `kubectl` from a Python app because the upstream Python client bindings struggle to keep up with new releases of the server doesn't feel right to me.

docs/developer/architecture/interaction-with-k8s-from-salt.rst
```diff
@@ -0,0 +1,96 @@
...
+Direct HTTP Query to Kubernetes API from Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We could build a Salt module that directly interacts with Kubernetes API the
+way we want so that we do not rely on python-kubernetes but only implement the
+interactions we need.
+
+This approach was rejected because it means this Salt module needs to be
+maintained for every new Kubernetes version and for every API/object we want
+to support.
```
NicolasT, line 85: Why's that? Given a K8s object (any, really) you can construct the URL to PUT/POST to, or GET from. The API is very regular. There should be no need to make any changes to such 'generic' client infrastructure.

It's quite likely the upstream Python K8s client already has everything you need: where it falls short right now is the models/classes representing all kinds of K8s objects (generated from OpenAPI specs). Underlying, there's the functionality to call into the HTTP API which we could re-use, generically, for any kind of object (since we no longer rely on the models/classes to be kept in sync).

We loose all client-side validation, of course.

NicolasT: https://github.com/kubernetes-client/python-base/tree/master/dynamic (H/t @gdemonet)

TeddyAndrieux: You're right, it would be a better approach, I will check this and update this PR

docs/developer/architecture/interaction-with-k8s-from-salt.rst
```diff
@@ -0,0 +1,125 @@
...
+flexible it’s mean that we do not validate the manifest we send to
+Kubernetes API, but the manifest can be validated using
+`kubernetes-validate`_ lib.
+
+This `kubernetes-validate`_ lib can have the same issue as
+`python-kubernetes`_ as it needs to be generated for every Kubernetes
+version, but it seems a bit more up-to-date and this validation is not
+strictly needed to create the objects, so if a specific version of
+`kubernetes-validate`_ is not yet released we could just remove the
+validation waiting for a new version to be released.
```
gdemonet, line 43: What about using [`dryRun`](https://kubernetes.io/docs/reference/kubernetes-api/common-parameters/common-parameters/#dryRun) as a means for validation, which would solve the update problem?

gdemonet (dismissed):
Outside the validation topic, LGTM :ok_hand:

docs/developer/architecture/interaction-with-k8s-from-salt.rst
TeddyAndrieux: Right, I will mention the 2 options here

gdemonet (approved):

PR #3510, opened 2021-08-26
Re-write Kubernetes Salt module to use `DynamicClient`

### LINKED ISSUE #3487: Bump Kubernetes to 1.22.X impossible with current python-kubernetes version

TeddyAndrieux:
**Component**:

'salt', 'tests'

**Context**: 

#3487 

**Summary**:

- Rewrite salt `metalk8s_kubernetes` execution module to use `DynamicClient`
- Remove salt `kubernetes` utils module as it's no longer needed
- Update e2e tests to use `DynamicClient` as well

---

Fixes: #3487


gdemonet (approved):

gdemonet:
Thanks for the heavy lifting :muscle:

PR #3513, opened 2021-08-31
storage-operator: Re generate CRD for Volume using "v1" API

TeddyAndrieux:
Since "v1beta1" of CRD is deprecated (and removed in Kubernetes 1.22)
let's generate Volume CRD using "v1"

NOTE: Since fields are pruned with this CRD version we need to disable
pruning on `.spec.template.metadata` since `operator-sdk` generate this
one with no properties

See: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

### DIFF
--- a/buildchain/buildchain/constants.py
+++ b/buildchain/buildchain/constants.py
@@ -126,5 +126,10 @@ STORAGE_OPERATOR_SOURCES: FrozenSet[Path] = frozenset(
 )
 GO_SOURCES: FrozenSet[Path] = frozenset(filepath for filepath in ROOT.rglob("*.go"))
 
+OPERATOR_SDK_GENERATE_CMDS: List[List[str]] = [
+    [config.ExtCommand.OPERATOR_SDK.value, "generate", "k8s"],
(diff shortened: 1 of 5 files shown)


buildchain/buildchain/constants.py
```diff
@@ -126,5 +126,14 @@ def git_ref() -> Optional[str]:
 )
 GO_SOURCES: FrozenSet[Path] = frozenset(filepath for filepath in ROOT.rglob("*.go"))
 
+OPERATOR_SDK_GENERATE_CMDS: List[List[str]] = [
+    [
+        config.ExtCommand.OPERATOR_SDK.value,
+        "generate",
+        "k8s",
+    ],
```
gdemonet: nit(style): why is this command split into multiple lines, but not the following one?

gdemonet (dismissed):

buildchain/buildchain/constants.py
TeddyAndrieux: For .... no reason :smile:

TeddyAndrieux: Fixed

gdemonet (approved):

PR #3514, opened 2021-08-31
salt: Use  `v1` APIVersion instead of `v1beta1` for `Ingress` objects

TeddyAndrieux:
Since `apiVersion` `networking.k8s.io/v1beta1` for Ingress object is deprecated and removed in Kubernetes 1.22 migrate all Ingress object to `networking.k8s.io/v1`

### DIFF
--- a/charts/render.py
+++ b/charts/render.py
@@ -328,6 +328,13 @@ def main():
     parser.add_argument("path", help="Path to the chart directory")
     args = parser.parse_args()
 
+    api_versions = [
+        # Used by APIService (available since Kubernetes 1.10)
+        "apiregistration.k8s.io/v1",
+        # Available since Kubernetes 1.19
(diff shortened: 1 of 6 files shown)


charts/render.py
```diff
@@ -328,6 +328,13 @@ def __call__(self, parser, args, values, option_string=None):
     parser.add_argument("path", help="Path to the chart directory")
     args = parser.parse_args()
 
+    api_versions = [
+        # Used by APIService (available since Kubernetes 1.10)
+        "api.registration.k8s.io",
```
gdemonet: Here it's written `api.registration`, whilst in the previous approach it was `apiregistration`. Which one is correct?

TeddyAndrieux: Oh right it's a typo, good catch

gdemonet (approved):

PR #3521, opened 2021-09-02
UI - Bugfixes multi nodes - high availability

JBWatenbergScality:
**Component**:

UI

**Context**: 

When a nodes is unavailable the UI reported wrong / incomplete information.
 - All the volumes where listed for all the nodes instead of filtering only the volumes of the selected node in the volumes tab of node detail pane
 - The deploy button was displayed when a node is unavailable but has already been deployed
 - An incorrect node status (unknown) was reported in the node table when a node is unavailable

### DIFF
--- a/ui/src/containers/NodePageRSP.js
+++ b/ui/src/containers/NodePageRSP.js
@@ -123,10 +123,13 @@ const NodePageRSP = (props) => {
 
   const alertList = useAlerts({
     alertname: NODE_ALERTS_GROUP,
-    instance: `${instanceIP}:${PORT_NODE_EXPORTER}`,
   });
 
-  const alertsNode = (alertList && alertList.alerts) || [];
+  const alertsNode = ((alertList && alertList.alerts) || []).filter(
+    (alert) =>
+      alert.labels.instance === `${instanceIP}:${PORT_NODE_EXPORTER}` ||
+      alert.labels.node === name,
+  );
 
   const isHealthTabActive = location.pathname.endsWith('/overview');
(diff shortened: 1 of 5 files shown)


ui/src/containers/NodePageRSP.js
```diff
@@ -123,10 +123,13 @@ const NodePageRSP = (props) => {
...
     alertname: NODE_ALERTS_GROUP,
-    instance: `${instanceIP}:${PORT_NODE_EXPORTER}`,
   });
 
-  const alertsNode = (alertList && alertList.alerts) || [];
+  const alertsNode = ((alertList && alertList.alerts) || []).filter(
+    (alert) =>
+      alert.labels.instance === `${instanceIP}:${PORT_NODE_EXPORTER}` ||
+      alert.labels.node === name,
+  );
```
ChengYanJin, line 132: I am wondering why can't we add this filter within useAlerts() ?
```js
  const alertList = useAlerts({
    alertname: NODE_ALERTS_GROUP,
    instance: `${instanceIP}:${PORT_NODE_EXPORTER}`,
   node: name
  });
  ```

JBWatenbergScality: That would make an AND filter and then exclude the Alerts with label instance: `${instanceIP}:${PORT_NODE_EXPORTER}` if the node label is not defined.

ChengYanJin: Oh. Yes, that's right.. I miss the OR logic in the filter.. 
Look good to me then!

ChengYanJin (approved):

PR #3501, opened 2021-08-19
UI: Fetch alert history

JBWatenbergScality:
**Component**:

ui

**Context**: 

In order to display the Global health component we have to provide a way for the UI to retrieve alert history.

**Summary**:

This PR improve the behavior of `useHistoryAlert` to take in account platform downTimes.

**Acceptance criteria**: 

Unit testing

PR #3522, opened 2021-09-02
ARTESCA-1878: Dashboard Network Plane Health

### LINKED ISSUE #3511: MetalK8s UI Dashboard Network Plane Health
**Component**: UI, monitoring

**Why this is needed**:
To display the health of **Control Plane Network** and **Workload Plane Network**. 
And being able to bring the users to the **Alert** Page if the health status is NOT healthy.

**What should be done**:
Graphic design:
image

The interaction design should be the same as this:
https://www.figma.com/proto/N2jOAK44dTSiBclDjXzz6a/Next_Gen_Design?node-id=301%3A116&viewport=1035%2C-840%2C0.23489561676979065&scaling=min-zoom&page-id=180%3A3495

Technical proposal:
We haven't deploy Alert _ControlPlaneNetworkDegraded_ and _WorkloadPlaneNetworkDegraded_, only _NetworkDegraded_ is avaiable.  So baically we use the same alert(NetworkDegraded) to compute the health of control plane network and workload plane network.

So I suggest we update the **getNetworksAlertSelectors** in shell-ui as shown below:
```
export const getNetworksAlertSelectors = (): FilterLabels => {
  return {
    alertname: ['NetworkDegraded'],
  };
};
```
And retrieve the selector in MetalK8s UI as it's done for other selectors. 

**Test plan**:
Unit test

theocerutti:
screen
Screenshot from 2021-09-02 15-38-13

Closes: #3511

### DIFF
--- a/ui/src/components/DashboardPlane.js
+++ b/ui/src/components/DashboardPlane.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+import styled from 'styled-components';
+import {
+  useAlertLibrary,
+  useHighestSeverityAlerts,
+  highestAlertToStatus,
+} from '../containers/AlertProvider';
+import { PageSubtitle } from './style/CommonLayoutStyle';
+import { PanelActions, NetworkContainer } from './DashboardNetwork';
+import HealthItem from './HealthItem.js';
+import { spacing } from '@scality/core-ui/dist/style/theme';
+
+const PlanesContainer = styled.div`
+  padding-left: ${spacing.sp8};
+  display: flex;
+  flex-direction: row;
+`;
+
+const PlaneContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  margin-right: ${spacing.sp40};
+`;
+
+const DashboardPlane = () => {
+  const intl = useIntl();
+  const alertsLibrary = useAlertLibrary();
+
+  const planesHighestSecurityAlert = useHighestSeverityAlerts(
+    alertsLibrary.getNetworksAlertSelectors(),
+  );
+  const planesStatus = highestAlertToStatus(planesHighestSecurityAlert);
+
+  return (
+    <NetworkContainer>
+      <PanelActions>
+        <PageSubtitle>Planes</PageSubtitle>
+      </PanelActions>
+      <PlanesContainer>
+        <PlaneContainer>
+          <HealthItem
+            label={intl.formatMessage({ id: 'control_plane' })}
+            status={planesStatus}
+            alerts={planesHighestSecurityAlert}
+            showArrow={false}
+          />
+        </PlaneContainer>
+        <PlaneContainer>
+          <HealthItem
+            label={intl.formatMessage({ id: 'workload_plane' })}
+            status={planesStatus}
+            alerts={planesHighestSecurityAlert}
+            showArrow={false}
+          />
+        </PlaneContainer>
+      </PlanesContainer>
+    </NetworkContainer>
+  );
+};
+
+export default DashboardPlane;
--- a/ui/src/components/DashboardPlane.test.js
+++ b/ui/src/components/DashboardPlane.test.js
@@ -1,7 +1,7 @@
 //@flow
 import React from 'react';
 import { screen } from '@testing-library/react';
-import DashboardServices from './DashboardServices';
+import DashboardPlane from './DashboardPlane';
 import { render } from './__TEST__/util';
 import type { Alert } from '../services/alertUtils';
 import { useHighestSeverityAlerts } from '../containers/AlertProvider';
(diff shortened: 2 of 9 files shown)


ui/src/components/HealthItem.js
```diff
@@ -0,0 +1,166 @@
...
+    );
+  else
+    return (
+      <NonHealthyServiceItemElement>
+        <Tooltip
+          // Right placement to avoid Z index issues between left sidebar and tooltip or out of screen tooltip displays
+          placement="top"
+          overlayStyle={{
+            // em sizing to handle font-size change on large screens
+            width: '20em',
```
ChengYanJin: In order to handle responsiveness font, we use `rem` instead of `em` :)

ui/src/translations/fr.json
```diff
@@ -137,6 +137,8 @@
   "used_by": "Utilisé Par",
   "backend_disk": "Backend Disk",
   "metrics": "Métrique",
+  "network": "Réseau",
+  "planes": "Plans",
```
ChengYanJin: IMO, we don't need to translate the `Planes`.

ui/src/components/DashboardPlane.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const planesHighestSecurityAlert = useHighestSeverityAlerts(
+    alertsLibrary.getNetworksAlertSelectors(),
+  );
+  const planesStatus = highestAlertToStatus(planesHighestSecurityAlert);
+
+  return (
+    <NetworkContainer>
+      <PanelActions>
+        <PageSubtitle>{intl.formatMessage({ id: 'planes' })}</PageSubtitle>
```
ChengYanJin: We should not translate this `planes`.

ui/src/components/DashboardPlane.test.js
```diff
@@ -0,0 +1,96 @@
...
+      isFederated: true,
+      view: { path: '/alerts' },
+    },
+  ],
+}));
+
+const NB_ITEMS = 2;
+
+describe("the dashboard network's plane panel", () => {
+  test("displays the network's plane panel and display 2 green statuses when no alerts are present", async () => {
```
ChengYanJin, line 66:
```suggestion
  test("displays 2 green statuses when no alerts are present", async () => {
```

ChengYanJin (commented):
Apart from the comment, I think we should have `StartsAt` field in the tooltip when the status is NOT healthy. 

I am not sure if we should display the arrow, I will let @Cuervino answer the question ;)

ui/src/components/DashboardPlane.test.js
```diff
@@ -39,17 +39,34 @@ jest.mock('../containers/AlertProvider', () => ({
   },
 }));
 
+jest.mock('../containers/ConfigProvider', () => ({
```
ChengYanJin, line 42: I don't think the work is related to the commit message :(

And please refer to the best practice of commit messages. 

https://metal-k8s.readthedocs.io/en/latest/developer/development/best_practices/commit.html?highlight=commit#subject

ui/src/components/HealthItem.js
```diff
@@ -132,15 +143,21 @@ const HealthItem = ({
             </NonHealthyPopUp>
           }
         >
-          <Link to="/alerts" data-testid="alert-link">
+          <div
+            onClick={() => {
+              history.push('/alerts');
```
ChengYanJin: could be removed, the openLink() already take care of this. 
```suggestion

```

ui/src/components/HealthItem.js
```diff
@@ -0,0 +1,166 @@
...
+      </ServiceItemElement>
+    );
+  else
+    return (
+      <NonHealthyServiceItemElement>
+        <Tooltip
+          // Right placement to avoid Z index issues between left sidebar and tooltip or out of screen tooltip displays
+          placement="top"
+          overlayStyle={{
+            // em sizing to handle font-size change on large screens
```
ChengYanJin: the comment could be removed
```suggestion
```

ui/src/components/DashboardPlane.js
```diff
@@ -44,13 +43,15 @@ const DashboardPlane = () => {
             label={intl.formatMessage({ id: 'control_plane' })}
             status={planesStatus}
             alerts={planesHighestSecurityAlert}
+            showArrow={false}
```
ChengYanJin, line 46: I checked with @Cuervino, so I think you were right, let's remove the arrow here.

ChengYanJin (changes requested):

ui/src/components/HealthItem.js
JulienMeziere: With the openLink() only, the "alert" page gets selected in the navbar but the page doesn't show up.
Before what we did, a Link component was used and it was showing the alert page but not selecting "alert" in the navbar.
Is there something I am missing to navigate using openLink ?

ChengYanJin: Please use `history.replace` instead of `history.push`

And clean up the commits with the best practice of the commit message. 

Then we are good to go.

Thanks!

JBWatenbergScality: To be precise the `history.replace` should be called after `openLink`.
This is necessary because openLink triggers a change on `shell-ui` history which renders the same federated component coming from Metalk8s (/ is matched for /alerts) then Metalk8s federated UI doesn't detect any changes in it's history and doesn't rerender.
Using `history.replace` before calling openLink would replace the current browsed page history. Calling it after would replace the change performed by `openLink`, so you should be browsed to the Alert page and pressing once the back button should bring you back to the overview page.

ui/webpack.common.js
```diff
@@ -29,7 +29,7 @@ module.exports = (env) => ({
     hints: 'error',
     // ~732 KiB for production
     // ~4.06 MiB for development because flow increase the size of assets.
-    maxAssetSize: process.env.NODE_ENV === 'production' ? 750000 : 5000000,
+    maxAssetSize: process.env.NODE_ENV === 'production' ? 750000 : 100000000,
```
ChengYanJin: I suggest dropping this change because we have already handled it on our side.

ChengYanJin: and you may want to fixup this commit with the first one

ChengYanJin (dismissed):

ChengYanJin (approved):

JBWatenbergScality:
Cherry-picked in #3501

Issue #3502, opened 2021-08-19
Control Plane Ingress IP management with MetalLB L2 does not work (by design) with multiple L2 segments for control plane network


TeddyAndrieux:
**Component**:

'metallb', 'network'

**What happened**:

Currently, MetalLB L2 does not support nodes in different L2 segments see https://github.com/metallb/metallb/issues/511

But even if MetalLB does current design in MetalK8s for control plane ingress with MetalLB cannot work as-is if we have control plane node that sits in different L2 segments, since we define a single VIP that, obviously, cannot be part of the various L2 segments we define for the control plane.

**What was expected**:

To be discussed/design

Several possibilities
- Forbid MetalLB L2 management when we have multiple L2 segments and let the user manage the (V)IP
- Forbid MetalLB L2 management when we have multiple L2 segments and only support MetalLB BGP in this case
- Allow defining multi VIP for control plane ingress, so that if you have multiple L2 segments you can define a VIP per L2 segments (if you want to) (NOTE: to do this approach we first need the MetaLB issue mentioned above to be fixed otherwise it will not work)

**Resolution proposal** (optional):

PR #3533, opened 2021-09-14
salt: Ensure we do not have multi cidr for Control Plane with MetalLB

TeddyAndrieux:
Today we do not support multiple subnets for the Control Plane network when
MetalLB is enabled, so this commit check that it's not the case

See: #3502

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
 - Fix UI issues in multi nodes environment when a node
   is unavailable (PR[#3521](https://github.com/scality/metalk8s/pull/3521))
 
+## Bug fixes
+
+- Enforce a single subnet for control plane when using a
+  MetalLB-managed VIP for Ingress
+  (PR [#3533](https://github.com/scality/metalk8s/pull/3533))
+
 ## Release 2.10.2
 ### Bug fixes
 - Fix the link to documentation from the UI navigation bar
--- a/salt/_pillar/metalk8s.py
+++ b/salt/_pillar/metalk8s.py
@@ -94,13 +94,18 @@ def _load_networks(config_data):
     # MetalLB disabled by default
     networks_data["controlPlane"].setdefault("metalLB", {}).setdefault("enabled", False)
 
-    if networks_data["controlPlane"]["metalLB"]["enabled"] and not networks_data[
-        "controlPlane"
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,13 @@
 - Fix UI issues in multi nodes environment when a node
   is unavailable (PR[#3521](https://github.com/scality/metalk8s/pull/3521))
 
+## Bug fixes
+
+- Manage Control Plane Ingress Virtual IP with MetalLB is not
+  supported when you have multiple subnet in Control Plane
+  network, now it's properly checked
```
TeddyAndrieux: Not 100% convinced about the sentence :smile: :thinking:

gdemonet: We usually document the change rather than the issue in this document, so maybe:

```suggestion
- Enforce a single subnet for control plane when using a
  MetalLB-managed VIP for Ingress
```

And maybe add a link to the relevant issue and/or discussions you had with the MetalLB guys when investigating this limitation.

salt/_pillar/metalk8s.py
```diff
@@ -94,13 +94,19 @@ def _load_networks(config_data):
...
+        if not networks_data["controlPlane"].get("ingress", {}).get("ip"):
+            errors.append(
+                "'ip' for 'ingress' in 'controlPlane' network is mandatory when "
+                "'metalLB' is enabled"
+            )
+        if len(networks_data["controlPlane"]["cidr"]) > 1:
+            errors.append(
+                "Control Plane Ingress Virtual IP management with MetalLB is "
+                "currently not supported when you have multiple subnet for "
+                "Control Plane network, see "
```
gdemonet: Let's keep the focus on what needs to be done, rather than what cannot be done:

```suggestion
                "Enabling 'metalLB' requires a single 'cidr' in "
                "'controlPlane' network, see "
```

gdemonet (commented):
Thanks for taking care of this :+1:

CHANGELOG.md
TeddyAndrieux: Ok fair,
For the link I'm not sure it's needed, we have the link to the PR that have a link to the ticket that talk about it, not sure we want to put the link to the ticket directly in the changelog

alexandre-allard (approved):

PR #3540, opened 2021-09-17
Define a new tool for generating alert trees, and use it for MetalK8s

gdemonet:
**Component**: monitoring, build, salt

**Context**: In the Overview page, we will require higher-level logical alerts to display the health of each "service category".

**Summary**:

- [x] Build a Python library for creating alert hierarchies and manipulating them
- [x] Use this library to represent all required alerts for MetalK8s
- [x] Generate the rules at build-time
- [x] Deploy the generated rules
- [x] Document the library in the developer guide

**Acceptance criteria**: 

- Overview page works after a plain bootstrap
- Other projects can use this library to build their own tree

---

#### `tox -e alert-tree -- show`

image

#### `tox -e alert-tree -- gen-rule`

image

---

Heavily inspired by an internal tool developed by @sayf-eddine-scality


buildchain/buildchain/config.py
```diff
@@ -63,6 +63,7 @@ class ExtCommand(enum.Enum):
     VAGRANT = os.getenv("VAGRANT_BIN", "vagrant")
     OPERATOR_SDK = os.getenv("OPERATOR_SDK_BIN", "operator-sdk")
     GOFMT = os.getenv("GOFMT_BIN", "gofmt")
+    TOX = os.getenv("TOX_BIN", "tox")
```
gdemonet, line 66: If you wonder why I added this check, I started by using `tox` for rendering the rules manifest, but it was strange to add a requirement for both `tox` and `poetry` to be available for building the product, just for this small manifest. So I went with Docker instead :roll_eyes:

docs/developer/tools/lib-alert-tree/examples/example.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Existing alerts are defined with their name and labels only.
+FOO_WARNING = Existing("FooServiceTooManyErrors", severity="warning")
+FOO_CRITICAL = Existing("FooServiceTooManyErrors", severity="critical")
+
+# There are shortcuts for defining the severity.
+BAR_BAZ_WARNING = Existing.warning("BarAlmostOutOfSpace", bar="baz")
+BAR_BAZ_CRITICAL = Existing.critical("BarAlmostOutOfSpace", bar="baz")
+
+# Derived alerts are built from a list of children and a relationship type (the default
```
sayf-eddine-scality: imo it is gonna be hard to undertand the code with such implicit relationship

docs/developer/tools/lib-alert-tree/index.rst
```diff
@@ -0,0 +1,53 @@
+``lib_alert_tree`` Python library
+=================================
+
+This Python library serves as common grounds when composing Prometheus alerts
```
sayf-eddine-scality:
```suggestion
This Python library serves as a common ground when composing Prometheus alerts
```

docs/developer/tools/lib-alert-tree/install.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. important::
+
+   This requires ``pip`` version 19 or higher to handle PEP 517 packages as
+   generated by Poetry.
+
+.. code:: console
+
+   $ python3 -m pip install --use-pep517 \
+       "git+https://github.com/scality/metalk8s.git@feature/lib-alert-tree#egg=lib_alert_tree&subdirectory=tools/lib-alert-tree"
```
sayf-eddine-scality: you will change the branch later? :D

docs/developer/tools/lib-alert-tree/usage.rst
```diff
@@ -0,0 +1,52 @@
...
+   $ python -m example gen-rule \
+       --name example.rules --namespace example \
+       --out example_rule.yaml
+
+The generated manifest should look like
+:download:`this one <examples/example_rule.yaml>`. Note that only
+``DerivedAlert`` instances are part of this manifest (exiting alerts should not
+need any additional rules, and are only referred to in the derived ``expr``).
+
+Use the ``-h`` / ``--help`` option to print a help message and discover more
```
alexandre-allard: There is no such `-h` option, only `--help`.

alexandre-allard (commented):
LGTM, nothing much to say, I've tested it and it's really cool.
I'm just wondering, you've changed some parent alert names (e.g. K8sMasterServicesDegraded > KubernetesControlPlaneDegraded), I don't know how these alerts are consumed UI-wise, etc. is it OK ? Or I assume you'll do a following PR on ARTESCA side to adapt that (at leas to remove the Prometheus rules deployment).
I don't approve yet since you already approved and you may want to edit some stuff (regarding the comments) prior to merge it.

docs/developer/tools/lib-alert-tree/examples/example.py
gdemonet: How is it implicit? I feel it to be simpler to understand when you link all children from a parent, rather than linking the parent to all children individually. What would you change?

Or do you mean the default being set to "ANY"?

gdemonet:
> LGTM, nothing much to say, I've tested it and it's really cool.

:heart:

> I'm just wondering, you've changed some parent alert names (e.g. K8sMasterServicesDegraded > KubernetesControlPlaneDegraded), I don't know how these alerts are consumed UI-wise, etc. is it OK ?

Good catch, I forgot about that, indeed I'll likely need to change a few things in MetalK8s UI.

> Or I assume you'll do a following PR on ARTESCA side to adapt that (at leas to remove the Prometheus rules deployment).

Indeed, one to remove the previous MetalK8s rules and maybe add some for ARTESCA services on top of those.

> I don't approve yet since you already approved and you may want to edit some stuff (regarding the comments) prior to merge it.

Good catch as well, I removed my `/approve`.

docs/developer/tools/lib-alert-tree/examples/example.py
sayf-eddine-scality: yes the ANY part

gdemonet: OK I'll set it back to being a mandatory argument.

gdemonet: Done in 542b283.

docs/developer/tools/lib-alert-tree/index.rst
gdemonet: Done

docs/developer/tools/lib-alert-tree/install.rst
gdemonet: Done (removed the branch specification entirely, this way the command defaults to the default branch, I think... anyway, not really recommended to use this unless you know what you're doing)

docs/developer/tools/lib-alert-tree/usage.rst
gdemonet: Fixed

gdemonet:
@alexandre-allard-scality I updated Shell UI's library for showing alerts in 89f5d12 (cc @JBWatenbergScality if you want to check I didn't miss anything).

shell-ui/src/alerts/alertHooks.js
```diff
@@ -6,7 +6,7 @@ import type { FilterLabels, Alert } from './services/alertUtils';
 import { getHealthStatus, filterAlerts } from './services/alertUtils';
 
 export const getPlatformAlertSelectors = (): FilterLabels => {
-  return { alertname: ['PlatformAtRisk', 'PlatformDegraded'] };
+  return { alertname: ['ClusterAtRisk', 'ClusterDegraded'] };
```
JBWatenbergScality, line 9: I'm wondering wether we should not include both the old alert names and the new ones in this filter. Because when upgrading we could loose historical alerts in the global health component. But maybe this is a none existing issue if no one is using those alerts in prod yet.

JBWatenbergScality:
I think an update of `alert-grouping.rst` documentation is missing with the new alert names

shell-ui/src/alerts/alertHooks.js
gdemonet: This lib hasn't been released yet, right? So I'm not sure how upgrade is a concern.
Some deployments may have alert rules matching the old names in place, and we may not remove them, but the lib won't care and won't fetch them.

JBWatenbergScality: The lib is not yet released indeed, however we already have packaged those alerts in products that are released and my concern is that for such installations an upgrade to Metalk8s 2.11 bundling the `dashboard page` would reflect a wrong history of alerts (pre upgrade some `PlatformAtRisk` or `PlatformDegraded` could be raised and then post upgrade we will watch `ClusterAtRisk` and `ClusterDegraded`) which will then mislead the user experience

gdemonet: Mmh, I consider it a feature addition, the previous alerts were not part of MetalK8s offering, so I don't think we should care.

JBWatenbergScality (approved):
Documentation will be done later on. Approving for now as discussed.

PR #3552, opened 2021-09-29
salt: Make prometheus persistent volume claim size configurable through CSC

JBWatenbergScality:
**Component**:
 
salt

**Context**: 

Add a prometheus.spec.config.storage_size in metalk8s-prometheus-config config map.

### DIFF
--- a/docs/operation/cluster_and_service_configuration.rst
+++ b/docs/operation/cluster_and_service_configuration.rst
@@ -291,6 +291,40 @@ This value can be overriden:
 
 Then :ref:`apply the configuration<csc-prometheus-apply-cfg>`.
 
+Change Storage Size
+"""""""""""""""""""
+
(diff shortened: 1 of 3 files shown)


docs/operation/cluster_and_service_configuration.rst
```diff
@@ -291,6 +291,40 @@ This value can be overriden:
...
+   metadata:
+     name: metalk8s-prometheus-config
+     namespace: metalk8s-monitoring
+   data:
+     config.yaml: |-
+       apiVersion: addons.metalk8s.scality.com
+       kind: PrometheusConfig
+       spec:
+         config:
+           storage_size: 30Gi
```
NicolasT, line 317: FWIW, K8s-style schemas use `storageSize`, not `storage_size`. Also, `storageCapacity` or similar is likely a better term.

PR #3554, opened 2021-10-01
salt: Add kubelet_scrape_timeout to configure kubelet service monitor scrape timeout

JBWatenbergScality:
**Component**:

salt

**Context**: 

When using a bunch of sparseloopdevice kubelet metrics endpoint may spend
a lot of time to retrieve volumes metrics. In such situation the default
10s scrape timeout of Prometheus is not enough as the metrics endpoint may
take longer to answer.

To avoid losing data in such situations, we enabled customization of kubelet
scarpe timeout.

### DIFF
--- a/docs/operation/cluster_and_service_configuration.rst
+++ b/docs/operation/cluster_and_service_configuration.rst
@@ -328,6 +328,37 @@ Both size and time based retentions can be activated at the same time.
 
 Then :ref:`apply the configuration<csc-prometheus-apply-cfg>`.
 
+Set Kubelet metrics scrape tiemout
+""""""""""""""""""""""""""""""""""
+
+In some cases (e.g. when using a lot of sparse loop devices), the kubelet
+metrics endpoint can be very slow to answer and the Prometheus' default 10s
+scrape timeout may not be sufficient.
+To avoid timeouts and thus losing metrics, you can customize the scrape
+timeout as follows:
+
+
+.. code-block:: yaml
+
+   ---
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: metalk8s-prometheus-config
+     namespace: metalk8s-monitoring
+   data:
+     config.yaml: |-
+       apiVersion: addons.metalk8s.scality.com
+       kind: PrometheusConfig
+       spec:
+         config:
+           serviceMonitor:
+             kubelet:
+               scrapeTimeout: 30s
+
+Then :ref:`apply the configuration<csc-prometheus-apply-cfg>`.
+
+
 Predefined Alert Rules Customization
 """"""""""""""""""""""""""""""""""""
 
--- a/charts/kube-prometheus-stack.yaml
+++ b/charts/kube-prometheus-stack.yaml
@@ -244,3 +244,7 @@ kubeControllerManager:
   serviceMonitor:
     https: true
     insecureSkipVerify: true
+
+kubelet:
+  serviceMonitor:
+    scrapeTimeout: "__var__(prometheus.spec.config.serviceMonitor.kubelet.scrapeTimeout)"
--- a/salt/metalk8s/addons/prometheus-operator/config/prometheus.yaml
+++ b/salt/metalk8s/addons/prometheus-operator/config/prometheus.yaml
@@ -11,6 +11,9 @@ spec:
     retention_time: "10d"
     retention_size: "0"  # "0" to disable size-based retention
     enable_admin_api: false
+    serviceMonitor:
+      kubelet:
+        scrapeTimeout: 10s
   rules:
     node_exporter:
       node_filesystem_space_filling_up:
--- a/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
@@ -60866,6 +60866,7 @@ spec:
       - __metrics_path__
       targetLabel: metrics_path
     scheme: https
+    scrapeTimeout: {% endraw -%}{{ prometheus.spec.config.serviceMonitor.kubelet.scrapeTimeout }}{%- raw %}
     tlsConfig:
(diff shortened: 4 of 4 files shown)


charts/render.py
```diff
@@ -254,6 +254,13 @@ def patch_grafana_dashboards(manifest):
     return manifest
 
 
+def patch_kubelet_service_monitor(manifest):
+    manifest["spec"]["endpoints"][0][
+        "scrapeTimeout"
+    ] = "__var__(prometheus.spec.config.kubelet_scrape_timeout)"
+    return manifest
```
alexandre-allard: You should rather change the `charts/kube-prometheus-stack.yaml` values:
```
kubelet:
  serviceMonitor:
    scrapeTimeout: "__var__(prometheus.spec.config.kubelet_scrape_timeout)"
```
And let the script handle that for you, without even touching it.

alexandre-allard (changes requested):

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -324,10 +324,41 @@ This functionality can be actived:
...
+       apiVersion: addons.metalk8s.scality.com
+       kind: PrometheusConfig
+       spec:
+         config:
+          kubelet_scrape_timeout: 30s
 
 Then :ref:`apply the configuration<csc-prometheus-apply-cfg>`.
 
+Kubelet metrics scrape tiemout and both size and time based retentions can be
+activated at the same time.
```
alexandre-allard: For me this information is only relevant for the retention, because it could be mutually exclusive, you should let it where it was.

alexandre-allard:
To apply the changes, the Prometheus manifest must be re-generated with the following command:
```
./charts/render.py prometheus-operator \
       charts/kube-prometheus-stack.yaml \
       charts/kube-prometheus-stack/ \
       --namespace metalk8s-monitoring \
       --service-config grafana \
       metalk8s-grafana-config \
       metalk8s/addons/prometheus-operator/config/grafana.yaml \
       metalk8s-monitoring \
       --service-config prometheus \
       metalk8s-prometheus-config \
       metalk8s/addons/prometheus-operator/config/prometheus.yaml \
       metalk8s-monitoring \
       --service-config alertmanager \
       metalk8s-alertmanager-config \
       metalk8s/addons/prometheus-operator/config/alertmanager.yaml \
       metalk8s-monitoring \
       --service-config dex \
       metalk8s-dex-config \
       metalk8s/addons/dex/config/dex.yaml.j2 metalk8s-auth \
       --drop-prometheus-rules charts/drop-prometheus-rules.yaml \
       > salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
```

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -324,10 +324,41 @@ This functionality can be actived:
...
+Set Kubelet metrics scrape tiemout
+""""""""""""""""""""""""""""""""""
+
+When using a bunch of sparseloopdevice kubelet metrics endpoint may spend
+a lot of time to retrieve volumes metrics. In such situation the default
+10s scrape timeout of Prometheus is not enough as the metrics endpoint may
+take longer to answer.
+
+To avoid losing data in such situations, we enabled customization of kubelet
+scarpe timeout, to customize it use the following configuration:
```
alexandre-allard:
```suggestion
In some cases (e.g. when using a lot of sparse loop devices), the kubelet metrics
endpoint can be very slow to answer and the Prometheus' default 10s scrape
timeout may not be sufficient. 

To avoid timeouts and thus losing metrics, you can customize the scrape timeout as follows:
```

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -324,10 +324,41 @@ This functionality can be actived:
...
+   metadata:
+     name: metalk8s-prometheus-config
+     namespace: metalk8s-monitoring
+   data:
+     config.yaml: |-
+       apiVersion: addons.metalk8s.scality.com
+       kind: PrometheusConfig
+       spec:
+         config:
+          kubelet_scrape_timeout: 30s
```
alexandre-allard: bad indentation:
```suggestion
           kubelet_scrape_timeout: 30s
```

alexandre-allard: I'm wondering if we should rather add an extra level for service monitor, rather than putting it like this, e.g.:
```
spec:
  config:
    serviceMonitor:
      kubelet:
        scrapeTimeout: 30s
```
Or even:
```
spec:
  config:
    kubelet:
      scrapeTimeout: 30s
```

charts/render.py
JBWatenbergScality: 👍  for some reason I hadn't found it configurable in the chart last week but in fact it is (https://github.com/scality/metalk8s/blob/development/2.11/charts/kube-prometheus-stack/templates/exporters/kubelet/servicemonitor.yaml#L45) so yes I can change it to what you suggested, good catch !!

JBWatenbergScality: Done in 9ca67a0442b2eb1dee63f9b5bed84eda288c9e03

docs/operation/cluster_and_service_configuration.rst
JBWatenbergScality: Done in efb76c1cae80a2eda688435e521bea325ee75e98

docs/operation/cluster_and_service_configuration.rst
JBWatenbergScality: Done in efb76c1cae80a2eda688435e521bea325ee75e98

docs/operation/cluster_and_service_configuration.rst
JBWatenbergScality: Done in efb76c1cae80a2eda688435e521bea325ee75e98

alexandre-allard (approved):
LGTM

PR #3560, opened 2021-10-08
lib-alert-tree: Convert children annotation from coma separated promql to JSON Path.

JBWatenbergScality:
**Component**:

lib-alert-tree

**Context**: 

In order to ease parsing and alert tree building by consumers we decided to switch
from a non standard coma separated set of promql queries defining the child alerts
to a well defined standard JSON Path.


tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -156,8 +161,9 @@ def build_tree(self, parent=None, tree=None):
 
     @property
     def alert_rule(self):
+        json_path_filters = " || ".join(map(lambda child: child.json_path, self.children))
```
gdemonet: This depends on the selected `Relationship`, so we should implement the "JSONPath query" in there.

tools/lib-alert-tree/lib_alert_tree/prometheus.py
```diff
@@ -86,6 +98,12 @@ def child_id(self):
         """A short representation of this alert, for use in annotations."""
         return f"{self.name}{{{self.format_labels()}}}"
 
+    @property
+    def child_json_path(self):
+        """A set of JSON Path filters representing this alert, for use in annotations."""
```
gdemonet: Technically not a set, rather an expression (since they are `&&`-ed together).

Also, "use in annotations" is not really describing the intent. Maybe reword to:

```suggestion
        """A JSONPath filter expression for selecting this alert as a child.
        
        This expression will be combined into a full JSONPath query for retrieving
        all children of a derived alert, exposed in an annotation for consumption
        by clients (such as UIs).
        """
```

tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -156,8 +161,9 @@ def build_tree(self, parent=None, tree=None):
...
         self.params.setdefault("annotations", {}).update(
-            {"children": ", ".join(map(str, self.children))}
+            {"children": f"$[?({json_path_filters})]" }
```
gdemonet: Wondering if we couldn't "simply" add a key for this, since the other one provides a more user-readable syntax :thinking:

gdemonet (changes requested):

tools/lib-alert-tree/lib_alert_tree/prometheus.py
gdemonet: (granted, it's almost the same docstring as just above for `child_id`)

tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -116,6 +121,15 @@ def build_query(self, children, group_by=None):
...
         return f"{prefix}({children_query}) >= 1"
 
+    def build_json_path(self, children):
+        """Build a JSON Path to retrieve alert children."""
+        if self == self.ALL:
+            join_by = " && "
+        elif self == self.ANY:
+            join_by = " || "
+        json_path_filters = join_by.join(map(lambda child: child.json_path, children))
+        return f"$[?({json_path_filters})]"
```
gdemonet: Now that's implement here, I realize we're missing the "group by" logic that we have in the `Relationship.build_query` method.

The idea behind this logic is the following:

- the child query is used to filter the domain of alerts
- the `group_by` argument then decomposes the filtered domain into units which will become alerts

In practice:

- say you have a `Parent` alert with a single child `Child{kind='example'}`, and its `group_by` constructor argument set to `["name"]`
- assume the `Child` alert has two labels, `kind` and `name`
- when building the subtree for a `Parent` instance, you'll want to also filter the `Child` alerts on `name`

This means we should propagate the `group_by` attribute to `prometheus.AlertRule` objects, so we can render it in another annotation and as part of the JSONPath query.

tools/lib-alert-tree/lib_alert_tree/prometheus.py
```diff
@@ -86,6 +98,16 @@ def child_id(self):
         """A short representation of this alert, for use in annotations."""
         return f"{self.name}{{{self.format_labels()}}}"
 
+    @property
+    def child_json_path(self):
+        """A JSONPath filter expression for selecting this alert as a child.
+
+        This expression will be combined into a full JSONPath query for retrieving
+        all children of a derived alert, exposed in an annotation for consumption
+        by clients (such as UIs)."""
```
gdemonet: [PEP 257 - Docstring Conventions](https://www.python.org/dev/peps/pep-0257/) recommends the following style:

```suggestion
        by clients (such as UIs).
        """
```

tools/lib-alert-tree/tests/test_models.py
```diff
@@ -101,12 +101,22 @@ def test_all():
             models.Relationship.ALL.build_query([models.ExistingAlert("test")])
             == "(ALERTS{alertname='test', alertstate='firing'}) >= 1"
         )
+        assert (
+            models.Relationship.ALL.build_json_path([models.ExistingAlert("test")])
+            == "$[?((@.labels.alertname === 'test'))]"
+        )
```
gdemonet: Could be interesting to refactor these tests with `pytest.mark.parametrize`, but it's not urgent. Will open a debt ticket.

gdemonet (changes requested):

tools/lib-alert-tree/lib_alert_tree/models.py
JBWatenbergScality: done in d28eb95

JBWatenbergScality:
Regarding parametrised tests they weren't collected by pytest despite the `@staticmethod` annotation so I had to drop the class and I renamed the 2 tests to make it explicit that they are testing the Relationship behavior.

gdemonet (approved):

tools/lib-alert-tree/tests/test_models.py
```diff
@@ -91,52 +91,89 @@ def test_existing_alert_severity_check():
...
+@pytest.mark.parametrize(
+    "relationship,children,expected_promql,expected_jsonpath",
+    [
+        (
+            models.Relationship.ALL,
+            [models.ExistingAlert("test1"), models.ExistingAlert("test2")],
+            "sum by (instance) (ALERTS{alertname='test1', alertstate='firing'} and "
+            "ALERTS{alertname='test2', alertstate='firing'}) >= 1",
+            "$[?(((@.labels.alertname === 'test1') && "
+            "(@.labels.alertname === 'test2')) && "
```
gdemonet: Actually, not sure how this filter expression can work, since no alert can honor `(@.labels.alertname === 'test1') && (@.labels.alertname === 'test2')`

gdemonet (changes requested):

tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -112,10 +117,26 @@ def build_query(self, children, group_by=None):
...
 
-        prefix = f"sum by ({', '.join(group_by)}) " if group_by else ""
+        prefix = f"sum by ({', '.join(group_by)}) " if group_by else "sum"
         children_query = join_by.join(child.query for child in children)
         return f"{prefix}({children_query}) >= 1"
 
+    @staticmethod
+    def build_json_path(children, group_by=None):
+        """Build a JSON Path to retrieve alert children."""
+        json_path_filters = " || ".join(map(lambda child: child.json_path, children))
```
gdemonet: `map(lambda x: ..., iterable)` is not idiomatic Python, we usually prefer to use generator expressions:

```suggestion
        json_path_filters = " || ".join(child.json_path for child in children)
```

tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -112,10 +117,26 @@ def build_query(self, children, group_by=None):
...
+
+        if group_by:
+            group_filters = " && ".join(
+                map(
+                    lambda key: f"@.labels.{key} === '{{{{ $labels.{key} }}}}'",
+                    group_by,
+                )
+            )
```
gdemonet: Same here:

```suggestion
            group_filters = " && ".join(
                f"@.labels.{key} === '{{{{ $labels.{key} }}}}'"
                for key in group_by
            )
```

tools/lib-alert-tree/lib_alert_tree/models.py
```diff
@@ -157,7 +178,12 @@ def build_tree(self, parent=None, tree=None):
     @property
     def alert_rule(self):
         self.params.setdefault("annotations", {}).update(
-            {"children": ", ".join(map(str, self.children))}
+            {
+                "children": ", ".join(map(str, self.children)),
```
gdemonet, line 179: Note: here, we use map, because we only apply a pre-existing function. It could feel more verbose written as `str(child) for child in self.children`, so I personally like to use `map(str, self.children)` instead. Matter of preference, I'd say.

gdemonet (changes requested):

gdemonet (approved):

PR #3561, opened 2021-10-08
docs/developer: Add design about Monitoring stack

TeddyAndrieux:
**Component**:

'docs', 'monitoring'

**Context**: 

Design document for monitoring stack with proper HA for Prometheus

**Summary**:

Add design document for monitoring stack already in place + document
new option to have proper HA for prometheus (using Thanos querier)

---

### DIFF
--- a/docs/developer/architecture/monitoring.rst
+++ b/docs/developer/architecture/monitoring.rst
@@ -1,11 +1,158 @@
 Monitoring
 ==========
 
-This document describes the monitoring features included in MetalK8s.
+Context
+-------
+
+In a MetalK8s cluster, almost each component (including applications running
+on top of it) generates metrics that allow to monitor its health or expose
+internal information.
+
+These metrics can be scrapped and then displayed to the end user in a fashion
+way to give him an easily readable overview of the cluster and applications
+health and status.
+
+Moreover, the metrics can also be used to send alerts depending on their
+values, allowing to quickly see if there is any issue or strange behavior
+on a component.
+
+Here, we want to describe the monitoring stack deployed in MetalK8s.
+
+Goals
+-----
+
+- Highly available metric history
+- Trigger alerts based on metric values
+- Visualization of metrics through charts
+- Easily scrape new metrics (from new MetalK8s pods or from workload pods
+  running on top of MetalK8s)
+- Configuration of alert routing and some existing alerts
+- Possibility to add extra alerts
+
+Design Design Choices
+---------------------
+
+In order to scrape metrics, trigger alerts and visualize metrics history we
+decided to use Prometheus, AlertManager and Grafana since it is the
+de facto standard for Kubernetes monitoring and is well integrated with
+almost any application running on top of Kubernetes.
+
+We use `kube-prometheus <https://github.com/prometheus-operator/kube-prometheus>`__
+which embeds the prometheus-operator that deploys the various components but
+also allows to easily monitor new metrics, add new Grafana dashboards and
+datasources.
 
 .. todo::
 
-   Describe the monitoring stack (`#1075`_), include quick explanation in
-   quickstart guide.
+  Add explanation about all components part of Prometheus stack and explain how
+  each of them interact between them and with the end users (alerts, graphs,
+  ...)
 
-.. _`#1075`: https://github.com/scality/metalk8s/issues/1075
+But Prometheus does not work as a cluster which means that to have proper HA we
+need to duplicate information on several Prometheus instances.
+
+Either each Prometheus instance monitors only a part of all nodes (with some
+overlaps in order to have HA), but then to visualize information we need to
+reach the right Prometheus instance depending on the node we want to see.
+
+Either each Prometheus instance monitors all nodes, but even in this case
+if for whatever reason one Prometheus is down for a short time we will have a
+"blank" in the resulting graphs while the information is available in the other
+Prometheus instances.
+
+To solve this we decided to use `Thanos <https://github.com/thanos-io/thanos>`__,
+so that we have a Thanos side-car on every Prometheus instances and a global
+Thanos querier that will query all the Prometheus instances in order to get
+metrics asked. And then it will deduplicate series, that come from different
+Prometheus instances, by merging all of them in a single time series.
+Check `the Thanos querier documentation <https://thanos.io/tip/components/query.md/#global-view>`__
+if you want more information about how it works.
+
+Thanos queries are exactly the same as Prometheus ones they both use PromQL.
+
+For the moment we only consider Thanos as a querier and we will not
+configure any long term retention, but it's a Thanos capability that can
+be interesting in the future.
+
+Rejected Choices
+----------------
+
+Federated Prometheus
+~~~~~~~~~~~~~~~~~~~~
+
+Prometheus can federate other Prometheus instances, it basically fetches all
+metrics from other Prometheus instances instead of getting them directly
+from Pod metric endpoints.
+
+It allows to have all metrics available in a single point, with a small
+network impact.
+
+This approach was rejected because it means we need to store all metrics in
+yet another Prometheus instance and it does not properly scale
+if we really want all metrics in this federated Prometheus instance.
+
+Cortex
+~~~~~~
+
+`Cortex <https://cortexmetrics.io/>`__ allows to have highly available
+metrics history, all Prometheus instances just push all metrics to
+Cortex.
+
+Cortex works almost the same way as a Federated Prometheus except that
+it runs as a cluster, so it is more scalable since you can have
+several Cortex replicas that only own a part of the metrics.
+The Cortex Querier will then retrieve the metrics from the right Cortex
+instance.
+
+This approach was rejected because it needs a dedicated DB to store the
+Cortex hash and it consumes more storage, it is a bit excessive for
+a single Kubernetes cluster (which is the intend of MetalK8s) and
+Cortex cannot be used as a querier only.
+
+Implementations Details
+-----------------------
+
+In order to deploy Thanos Prometheus side-car and deploy the Thanos querier
+that interact with those side-cars we need to add new Salt states in
+MetalK8s.
+
+Those salt states come from
+`Kubernetes Prometheus stack helm chart <https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack>`__
+and `Thanos helm chart <https://github.com/banzaicloud/banzai-charts/tree/master/thanos>`__
+as described in `the Thanos readme <https://github.com/banzaicloud/banzai-charts/tree/master/thanos#install-prometheus-operator>`__,
+then those helm charts are rendered to salt states using the ``render.py``
+script from ``charts`` directory.
+
+Thanos will only be used as a querier for the moment so we do not need any
+specific configuration for it.
+
+Create a Grafana datasource for Thanos querier and use it for all dashboards
+that need to query Prometheus metrics.
+
+Expose Thanos querier through the `metalk8s-ui-proxies-http` Ingress object to
+be able to query Prometheus metrics from the MetalK8s UI.
+
+Documentation
+-------------
+
+In the Installation Guide:
+
+* Document how to access Grafana
+
+In the Operational Guide:
+
+* Document how this monitoring stack works
+* Document what are the default alerts deployed by MetalK8s
+* Document how to perform a Prometheus snapshot
+* Document how to configure Prometheus through CSC
+
+Test Plan
+---------
+
+Add test scenarios for monitoring stack using pytest-bdd framework to
+ensure the correct behavior of this feature.
+
+* Ensure that every deployed Pods are properly running after installation
+* Ensure that metrics are properly scrapped on every Prometheus instances
+* Ensure that alerts are properly raised
+* Ensure that we can query metrics from Thanos Querier


xaf-scality (dismissed):

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+
+Create a Grafana datasource for this Thanos querier and remove the Prometheus
+one.
+
+Update the Ingress object used to reach Prometheus API so that it goes to
+Thanos querier instead.
+
+We may also want to change the path in this Ingress as it mention Prometheus,
+then we also need to update the MetalK8s UI in order to use the new path to
+retrieve metrics.
```
alexandre-allard: Not sure it should appears as is, it's more a tasks list.
It should either be reformulated and/or go to an "Iterations" section where you want to put "actions" to do in order to apply the design.

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
 ==========
 
-This document describes the monitoring features included in MetalK8s.
+Context
+-------
 
-.. todo::
+MetalK8s value is to provide, out of the box, some services in order to ease
+the monitoring and operation of the platform as well as workloads running on
+top of it.
```
alexandre-allard: I saw what you did there. :)
To me, the context should quickly explain what are the monitoring principles and what it is used for, something like that (can/should be reworded, it's just an idea):
```
In a MetalK8s cluster, almost each component (including applications running on top of it)
generates metrics that allow to monitor its health or expose internal information.
These metrics can be scrapped and then displayed to the end user in a fashion
way to give him an easily readable overview of the cluster and applications health and status.
Moreover, the metrics can also be used to send alerts depending on their values,
allowing to quickly see if there is any issue or strange behavior on a component.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+Prometheus can federate other Prometheus instances, it basically fetch all
+metrics from another Prometheus instance instead of getting them directly from
+metric endpoints of the Pods.
+
+It allow to have all metrics available in a single point with a smaller network
+impact.
+
+This approach was rejected because it means we need to store all metrics in
+yet another Prometheus instance and it does not allow to properly scale
+Prometheus if we really want all metrics in this federate Prometheus instance.
```
alexandre-allard:
```suggestion
Prometheus can federate other Prometheus instances, it basically fetches all
metrics from other Prometheus instances instead of getting them directly
from Pod metric endpoints.

It allows to have all metrics available in a single point, with a small network
impact.

This approach was rejected because it means we need to store all metrics in
yet another Prometheus instance and it does not properly scale
if we really want all metrics in this federated Prometheus instance.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+
+Cortex
+~~~~~~
+
+`Cortex <https://cortexmetrics.io/>`__ allow to have Highly available
```
alexandre-allard:
```suggestion
`Cortex <https://cortexmetrics.io/>`__ allows to have highly available
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+To solve this we decided to use `Thanos <https://github.com/thanos-io/thanos>`__,
+so that we have a Thanos side-car on every Prometheus instances and a global
+Thanos querier that will query all the Prometheus instances in order to get
+metrics asked.
+
+Thanos queries are exactly the same as Prometheus ones they both use PromQL.
+
+For the moment we only consider Thanos as a querier and we will not
+configure any long term retention, but it's a Thanos capability that can
+be interesting in the future.
```
alexandre-allard: To me, this design does not perfectly reflect the monitoring stack and is more focused on the HA part of it.
I know that's what you need to work on, for Thanos, but maybe we could put more effort on it or maybe having a dedicated document for monitoring HA and leave the monitoring design doc empty for now would be a way to go, to avoid losing too much time on it.

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+
+`Cortex <https://cortexmetrics.io/>`__ allow to have Highly available
+metrics history, all Prometheus instances just push all metrics to
+Cortex.
+
+Cortex work almost the same way as a Federated Prometheus except that
+it work as a cluster so it's really more scalable as you can have
+several Cortex replicas that own only a part of all the metrics and
+then the Cortex Querier just retrieve the metric to the right Cortex
+instance.
```
alexandre-allard:
```suggestion
Cortex works almost the same way as a Federated Prometheus except that
it runs as a cluster, so it is more scalable since you can have
several Cortex replicas that only own a part of the metrics.
The Cortex Querier will then retrieve the metrics from the right Cortex
instance.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+
+This approach was rejected as it need's a dedicated DB to store the
+Cortex hash and it's consume more storage, it's a bit overkill for
+a single Kubernetes cluster (which is the intend of MetalK8s) and
+Cortex cannot be used as a Querier only.
```
alexandre-allard:
```suggestion
This approach was rejected because it needs a dedicated DB to store the
Cortex hash and it consumes more storage, it is a bit excessive for
a single Kubernetes cluster (which is the intend of MetalK8s) and
Cortex cannot be used as a querier only.
```

alexandre-allard (commented):
Few comments.

docs/developer/architecture/monitoring.rst
TeddyAndrieux: :shushing_face: 
Ok you suggestion looks good yeah :+1:

docs/developer/architecture/monitoring.rst
TeddyAndrieux: Hmm I agree that it does not totally describe the monitoring stack but to me this "monitoring HA" stuff should be part of the "monitoring design" at the end, maybe I could just add some todo to properly describe the full monitoring stack.

But I don't know what we can add about it here ? The design choice is only about "prometheus stack" we do not do anything special about this I think, it's "classic" "prometheus operator deployment", no ?

docs/developer/architecture/monitoring.rst
TeddyAndrieux: Hmm then I'm not sure what should we put in "Implementatios Details" here ?

docs/developer/architecture/monitoring.rst
alexandre-allard: Yes, the HA is part of the whole design, I agree, so maybe just put todo as you suggests is good enough for now.
It's pretty classic stuff, but we should still describe each element we want (monitoring, alerting, charts, ...) and how each of these features are provided (e.g. alerting is done with Prometheus through alerting rules on metrics and Alertmanager takes care of dedup and routes message, etc.), how they behaves, communicates, ...

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+Implementations Details
+-----------------------
+
+Create salt states rendered from
+`Kubernetes Prometheus stack helm chart <https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack>`__
+and `Thanos helm chart <https://github.com/banzaicloud/banzai-charts/tree/master/thanos>`__
+as described in `the Thanos readme <https://github.com/banzaicloud/banzai-charts/tree/master/thanos#install-prometheus-operator>`__.
+
+Thanos will only be used as a querier for the moment so we do not need any
+specific configuration for this Thanos querier.
```
alexandre-allard:
```suggestion
Thanos will only be used as a querier for the moment so we do not need any
specific configuration for it.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,112 @@
...
+a single Kubernetes cluster (which is the intend of MetalK8s) and
+Cortex cannot be used as a Querier only.
+
+Implementations Details
+-----------------------
+
+Create salt states rendered from
+`Kubernetes Prometheus stack helm chart <https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack>`__
+and `Thanos helm chart <https://github.com/banzaicloud/banzai-charts/tree/master/thanos>`__
+as described in `the Thanos readme <https://github.com/banzaicloud/banzai-charts/tree/master/thanos#install-prometheus-operator>`__.
```
alexandre-allard: It should be explained a bit, at least what is it for ? Maybe, how do you render these Salt states ?

docs/developer/architecture/monitoring.rst
TeddyAndrieux: Ok you right, fair I will add a todo for this

docs/developer/architecture/monitoring.rst
TeddyAndrieux: Sure

TeddyAndrieux: done

docs/developer/architecture/monitoring.rst
TeddyAndrieux: done

docs/developer/architecture/monitoring.rst
alexandre-allard: I don't really know and maybe I'm wrong, just to me it's strange to have implementation details for HA that rely on changing stuff from the monitoring stack implementation which is not yet described.
And it will be weird when the whole design will be there (if it is someday).
Maybe something that does not talk about the current implementation:
```suggestion
Create a Grafana datasource for Thanos querier and use it for all dashboards that
need to query Prometheus metrics.

Expose Thanos querier through the `metalk8s-ui-proxies-http` Ingress object to
be able to query Prometheus metrics from outside the cluster.

Use Thanos querier to query Prometheus metrics from the MetalK8s UI.
```
But you can also probably leave it as is and at some point we will rework the whole document...

TeddyAndrieux: Right ok I see your point, I agree I will make it more "generic" (not only mention what need to be done today) and add a todo as well here to add implementation details about "prometheus stack"

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
+--------------
+
+In order to scrape metrics, trigger alerts and visualize metrics history we
+decided to use Prometheus, AlertManager and Grafana, it's really common choice
+in Kubernetes world.
+
+We use `kube-prometheus <https://github.com/prometheus-operator/kube-prometheus>`__
+which embed the prometheus-operator that deploy the various components but also
+allow to easily monitor new metrics, add new Grafana dashboards, add new
+Grafana datasources.
```
alexandre-allard:
```suggestion
which embeds the prometheus-operator that deploys the various components but also
allow to easily monitor new metrics, add new Grafana dashboards and datasources.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
+- Easy way to visualize metrics history
+- Easy way to add new metrics (from new MetalK8s pods or from workload pods
+  running on top of MetalK8s)
+
+Design Choices
+--------------
+
+In order to scrape metrics, trigger alerts and visualize metrics history we
+decided to use Prometheus, AlertManager and Grafana, it's really common choice
+in Kubernetes world.
```
alexandre-allard: We can't choose something only because it's a common choice. :) At least we must say it's because it's already integrated with a lot of things which makes it easier to integrate in the stack or interact with:
```suggestion
decided to use Prometheus, AlertManager and Grafana since it is the 
de facto standard for Kubernetes monitoring and is well integrated with
almost any application running on top of Kubernetes.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
+  Add explanation about all components part of Prometheus stack and explain how
+  each of them interact between them and with the end users (alerts, graphs,
+  ...)
+
+But Prometheus does not work as a cluster which means that to have proper HA we
+need to duplicate information on several Prometheus instance.
+
+Either each Prometheus instances monitor only a part of all nodes (with some
+overlap in order to have HA), but then to visualize information we need to
+reach the right Prometheus instance depending on the node we want to see.
```
alexandre-allard:
```suggestion
Either each Prometheus instance monitors only a part of all nodes (with some
overlaps in order to have HA), but then to visualize information we need to
reach the right Prometheus instance depending on the node we want to see.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
+
+Either each Prometheus instances monitor all nodes, but even in this case
+if for whatever reason one Prometheus is down for a short time we will have a
+"blank" in the resulting graphs while the information is available in the other
+Prometheus instance.
```
alexandre-allard:
```suggestion
Either each Prometheus instance monitors all nodes, but even in this case
if for whatever reason one Prometheus is down for a short time we will have a
"blank" in the resulting graphs while the information is available in the other
Prometheus instances.
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
+
+To solve this we decided to use `Thanos <https://github.com/thanos-io/thanos>`__,
+so that we have a Thanos side-car on every Prometheus instances and a global
+Thanos querier that will query all the Prometheus instances in order to get
+metrics asked.
```
alexandre-allard: Does Thanos aggregate metrics and do an average (ex: I have 3 instances, that return 3, 4 and 5 for a specific metric, then Thanos returns 4) or it only tries to fetch the metrics from the first instance that have them and consolidates them ?
Maybe it's behavior could be explained a bit.

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,128 @@
...
 .. todo::
 
-   Describe the monitoring stack (`#1075`_), include quick explanation in
-   quickstart guide.
+  Add explanation about all components part of Prometheus stack and explain how
+  each of them interact between them and with the end users (alerts, graphs,
+  ...)
+
+But Prometheus does not work as a cluster which means that to have proper HA we
+need to duplicate information on several Prometheus instance.
```
alexandre-allard:
```suggestion
But Prometheus does not work as a cluster which means that to have proper HA we
need to duplicate information on several Prometheus instances.
```

alexandre-allard (commented):
2nd round, mostly typos.

docs/developer/architecture/monitoring.rst
alexandre-allard: You could also add an "Iterations" section where you could left the first iteration empty with a todo (the one for what we have as of today) and add a second one for the HA where you could put all the actions needed to achieve the HA.

docs/developer/architecture/monitoring.rst
TeddyAndrieux: True but it's already integrated with a lot of things because it's a common choice :stuck_out_tongue: 
No you right I will update

docs/developer/architecture/monitoring.rst
TeddyAndrieux: I don't know, I need to check but normally you should have the same value on every prometheus instances for a specific metric at a specific time

alexandre-allard: Not necessarily, well not with the exact same timestamp for sure, but Prometheus instances do not scrape the metrics at the very same moment, so the metrics for a time range can slightly diverge from an instance to another.

alexandre-allard: Ok, I found my answer in Thanos documentation.
```
The query layer can deduplicate series that were collected
from high-availability pairs of data sources such as Prometheus.
A fixed single or multiple replica labels must be chosen for the
entire cluster and can then be passed to query nodes on startup.

Two or more series that are only distinguished by the given replica
label, will be merged into a single time series. This also hides gaps
in collection of a single data source.
```
So, it basically merges the timeseries of all instances into a bigger timeseries.
Maybe, just add a sentence explaining this behavior.

TeddyAndrieux: Yes right I saw this as well, I will add it and also add a link to Thanos documentation, thank you :+1:

TeddyAndrieux: Done

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,132 @@
...
+
+Goals
+-----
+
+- History on metrics
+- Trigger alerts based on some "expression" on metric values
+- Metrics history highly available
+- Easy way to visualize metrics history
+- Easy way to add new metrics (from new MetalK8s pods or from workload pods
+  running on top of MetalK8s)
```
alexandre-allard:
```suggestion
- Highly available metric history
- Trigger alerts based on metric values
- Visualization of metrics through charts
- Easily scrape new metrics (from new MetalK8s pods or from workload pods
  running on top of MetalK8s)
- Configuration of alert routing and some existing alerts
- Possibility to add extra alerts
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,133 @@
...
+Check `the Thanos querier documentation <https://thanos.io/tip/components/query.md/#global-view>`__
+if you want more information about how it works.
+
+Thanos queries are exactly the same as Prometheus ones they both use PromQL.
+
+For the moment we only consider Thanos as a querier and we will not
+configure any long term retention, but it's a Thanos capability that can
+be interesting in the future.
+
+Rejected Choices
```
alexandre-allard, line 77:
```suggestion
Rejected Design Choices
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,133 @@
...
+--------------
+
+In order to scrape metrics, trigger alerts and visualize metrics history we
+decided to use Prometheus, AlertManager and Grafana since it is the
+de facto standard for Kubernetes monitoring and is well integrated with
+almost any application running on top of Kubernetes.
+
+We use `kube-prometheus <https://github.com/prometheus-operator/kube-prometheus>`__
+which embeds the prometheus-operator that deploys the various components but
+also allow to easily monitor new metrics, add new Grafana dashboards and
```
alexandre-allard:
```suggestion
also allows to easily monitor new metrics, add new Grafana dashboards and
```

alexandre-allard (commented):
It may miss a `Test Plan` and a `Documentation` sections, the actual tests we have for Prometheus should still work since we're using the ingress, but still we should at least write the cases related to Thanos (e.g. ability to retrieve some metrics from an API call) and leave a TODO for the others.
I'm also wondering if we should have tests specific to Thanos, at least check the querier is here in `Scenario: Expected Pods` and, up and running in `Scenario: Monitored components statuses`.
For the documentation... I don't see anything specific to Thanos... so maybe it's OK to omit it for now.

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,158 @@
...
+
+In the Installation Guide:
+
+* Document how to access Grafana
+
+In the Operational Guide:
+
+* Document how this monitoring stack works
+* Document what are all the defaults alerts deployed by MetalK8s
+* Document how to performs a Prometheus snapshot
```
alexandre-allard:
```suggestion
* Document what are the default alerts deployed by MetalK8s
* Document how to perform a Prometheus snapshot
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,158 @@
...
+* Document how to increase the number of Prometheus instances
```
alexandre-allard: If you want to be exhaustive, there is also documentation about Prometheus retention by size, by age and other stuff, so instead of listing what can be configured, maybe just say:
```suggestion
* Document how to configure Prometheus through CSC
```

docs/developer/architecture/monitoring.rst
```diff
@@ -1,11 +1,158 @@
...
+
+Test Plan
+---------
+
+Add test scenario for monitoring stack using pytest-bdd framework to
```
alexandre-allard:
```suggestion
Add test scenarios for monitoring stack using pytest-bdd framework to
```

alexandre-allard (commented):
Nitpicking. :)

alexandre-allard (dismissed):

alexandre-allard (approved):

Issue #3564, opened 2021-10-11
salt-ssh cannot load pillar if a volume already exists on this specific node


TeddyAndrieux:
**Component**:

'salt'

**What happened**:

If you try to run a salt-ssh state on a node that has some volumes deployed on it (it's the case when you try to restore bootstrap node for example) it fails

```console
TypeError encountered executing state.sls: Object of type 'datetime' is not JSON serializable
```

It's due to a `time` in storage classes `metadata:managed_fields`.

**What was expected**:

Working salt-ssh even if you have some volumes

**Steps to reproduce**

- Have a multi-node cluster with some volumes
- Try to run a state with salt-ssh on a node that has some volumes
- :boom: 

**Resolution proposal** (optional):

As we do for other timestamps In storage classes, convert this `time` to str in our custom pillar

---

NOTE: It's an issue on MetalK8s 2.10 but may not be an issue with the latest 2.11 as we changed the way to interact with Kubernetes API

PR #3566, opened 2021-10-12
salt: Convert `datetime` from `managedFields` from storage classes in pillar

TeddyAndrieux:
**Component**:

'salt'

**Context**: 

#3564 

**Summary**:

We store volumes and storage classes in the salt pillar that contains
some `datetime` python objects and `datetime` objects cannot be rendered
to json using "classic" `json.dump`.
This means that salt-ssh command that relies on the pillar (like salt
states) crash if some volume is deployed on the target node

---

Fixes: #3564

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 ## Release 2.10.4 (in development)
+## Bug fixes
+
+- [#3564](https://github.com/scality/metalk8s/issues/3564) - Fix a bug that
+  prevents running salt states using salt-ssh if the target node has some
+  MetalK8s volumes
+  (PR[#3566](https://github.com/scality/metalk8s/pull/3566))
+
 ## Release 2.10.3
 ### Enhancements
 
--- a/salt/_pillar/metalk8s_nodes.py
+++ b/salt/_pillar/metalk8s_nodes.py
@@ -84,6 +84,9 @@ def get_storage_classes(kubeconfig=None):
         storageclass["metadata"]["deletion_timestamp"] = iso_timestamp_converter(
             storageclass["metadata"]["deletion_timestamp"]
         )
+        for managed_field in storageclass["metadata"]["managed_fields"]:
+            managed_field["time"] = iso_timestamp_converter(managed_field["time"])
+
         storage_classes[storageclass["metadata"]["name"]] = storageclass
 
     return storage_classes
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # CHANGELOG
 ## Release 2.10.4 (in development)
+## Bug fixes
+
+- [#3564](https://github.com/scality/metalk8s/issues/3564) - Fix a bug that
+  prevent running salt state using salt-ssh if the target node has some
```
alexandre-allard:
```suggestion
  prevents running salt states using salt-ssh if the target node has some
```

TeddyAndrieux: Too fast !!!

tests/post/steps/test_salt_ssh.py
```diff
@@ -0,0 +1,43 @@
...
+
+@scenario(
+    "../features/salt_ssh.feature", "salt-ssh state work on every non-bootstrap nodes"
+)
+def test_salt_ssh_state(host):
+    pass
+
+
+@then(
+    parsers.parse("we are able to run '{state}' using salt-ssh on non-bootstrap nodes")
```
alexandre-allard: Could be a variable, but... hey, for now I think it will be enough. So, feel free to ignore.
```suggestion
    parsers.parse("we are able to run '{state}' using salt-ssh on non-{role} nodes")
```

alexandre-allard (dismissed):

tests/post/steps/test_salt_ssh.py
TeddyAndrieux: I will do it since build just started

alexandre-allard (approved):

salt/_pillar/metalk8s_nodes.py
```diff
@@ -84,6 +84,9 @@ def get_storage_classes(kubeconfig=None):
         storageclass["metadata"]["deletion_timestamp"] = iso_timestamp_converter(
             storageclass["metadata"]["deletion_timestamp"]
         )
+        for managed_field in storageclass["metadata"]["managed_fields"]:
```
gdemonet, line 87: IMO, we don't care about those when dumping "embedded" objects, but :shrug:

TeddyAndrieux: Yes I agree but since anyway this "convert" is needed only for 2.10 and we already have those in the previous 2.10 ... I kept them

PR #3569, opened 2021-10-19
MetalK8s UI: Fix wrong status in global health component

ChengYanJin:
**Component**: UI, dashboard

**Context**: 
In order to retrieve history alerts from LOKI Endpoint, we have to execute several queries in sequence, which causes the performance issue. 

So we switch to query Prometheus Endpoint to display the status of the Global Health Component. 

**Summary**:
Format the storage capacity in the PV list return by K8s Endpoint, to make sure the size has the correct unit.

We don't have data for the query we use to retrieve the number of ClusterAtRisk and ClusterDegraded, which cause the unavailable segment.

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


ui/src/components/DashboardGlobalHealth.js
```diff
@@ -19,7 +19,9 @@ import { useStartingTimeStamp } from '../containers/StartTimeProvider';
 import LoaderComponent from '@scality/core-ui/dist/components/loader/Loader.component';
 import CircleStatus, { StatusIcon } from './CircleStatus';
 import GlobalHealthBarComponent from '@scality/core-ui/dist/components/globalhealthbar/GlobalHealthBar.component';
-import { useHistoryAlerts } from '../containers/AlertHistoryProvider';
+import { getClusterAlertSegement } from '../services/platformlibrary/metrics';
```
JBWatenbergScality: Small typos and I would rename it as follows
```suggestion
import { getClusterAlertSegmentQuery } from '../services/platformlibrary/metrics';
```

ui/src/services/loki/api.js
```diff
@@ -40,33 +41,85 @@ export function getLast7DaysAlerts(): Promise<Alert[]> {
...
+    LOKI_RE_NOTIFICATION_INTERVAL
+  ) {
+    start = new Date(
+      new Date(end).getTime() - LOKI_RE_NOTIFICATION_INTERVAL,
+    ).toISOString();
   }
 
   return lokiApiClient
     .get(
       //We set limit to 1000 because the default number of lines retrievable is 100 which is
```
JBWatenbergScality: Upgrade this comment, `4999 because it is the max retrievable in one call`

JBWatenbergScality (dismissed):

ui/src/ducks/app/volumes.js
```diff
@@ -250,6 +251,13 @@ export function* fetchPersistentVolumes(): Generator<
...
   const result = yield call(VolumesApi.getPersistentVolumes);
   if (!result.error) {
+    //the storage capacity in the spec may not contain the reasonable/correct unit size
+    result &&
+      result.body &&
+      result.body.items &&
+      result.body.items.forEach((item) => {
+        bytesToSize(allSizeUnitsToBytes(item.spec.capacity.storage));
+      });
     yield put(setPersistentVolumesAction(result.body.items ?? []));
```
JBWatenbergScality:
```suggestion
    const pvs = result?.body?.items?.map((item) => {
        item.spec.capacity.storage = bytesToSize(allSizeUnitsToBytes(item.spec.capacity.storage));
return item;
      });
    yield put(setPersistentVolumesAction(pvs ?? []));
```

JBWatenbergScality (approved):

PR #3572, opened 2021-10-21
docs: Add section in logs arch doc about Loki HA

alexandre-allard:


docs/developer/architecture/logs.rst
```diff
@@ -442,6 +442,41 @@ extra component.
   define loki storage sizing depending on retention, ingestion rate,
   average size of logs, number of streams, ...
 
+Loki HA
+_______
+
+It is possible to provide a member list to Loki through its configuration,
+so that it is aware of the other Loki instances.
```
JBWatenbergScality:
```suggestion
so that it can discover the other Loki instances.
```

docs/developer/architecture/logs.rst
```diff
@@ -442,6 +442,41 @@ extra component.
...
+This way, it allows any instance to replicate, on some or all the other
+instances (depending on the replication factor), every logs it receives.
+This configuration will be dynamically generated during installation depending
```
JBWatenbergScality: Can't we try to make it dynamic ? If I recall correctly it was possible to configure loki member list through a service discovery mechanism (I remember seeing a mention to etcd, so maybe kubernetes based service discovery is achievable, worth checking IMO).

alexandre-allard: Yep, looks like it is possible to pass a headless service.
I'll give it a try and change that accordingly.

docs/developer/architecture/logs.rst
```diff
@@ -442,6 +442,48 @@ extra component.
...
+quantity of logs of the platform), it will be able to catch up what is still
+in other ingesters memory and not written yet to the storage backend.
+
+To solve this problem, we either need to be able to share this storage
+between all the instances (e.g. NFS), but it is probably a bad idea, either we
+need an extra component in front of the queriers, which takes care of querying
+each instance, deduplicating the entries and returning them.
+We have already considered the Loki Query Frontend, but its goal is only
+to split queries, dispatch the load on the multiple instances and do some
+caching.
```
gdemonet, line 485: Which could come in handy if we want to implement sharding

gdemonet (approved):

PR #3576, opened 2021-10-25
fix(CircleStatus): Fix rendering issue for CircleStatus

hug33k:
This PR fixes rendering issue for CircleStatus by avoiding non-necessary re-renders

**Component**:

`CircleStatus`, `StatusIcon`

**Context**: 

When `Icons` from Core-UI are implemented into `CircleStatus` component, they may re-render periodically even when an update was not required (e.g.: Status not changing).

**Summary**:

By adding a prop update verification into `CircleStatus`, component will now only re-render when status is changing, in order to avoid non-necessary re-renders.

Also, `StatusIcon` component is now in a dedicated file since it is now no longer used by `CircleStatus`.

**Acceptance criteria**: 

Health icons in Core-UI ([See related PR](https://github.com/scality/core-ui/pull/378)) should be merged and available first, in order to have access to the icons used by `CircleStatus`.

---

See: ARTESCA-2152

### DIFF
--- a/ui/src/components/CircleStatus.js
+++ b/ui/src/components/CircleStatus.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import styled from 'styled-components';
+import { Icon } from '@scality/core-ui';
 
 import {
   STATUS_WARNING,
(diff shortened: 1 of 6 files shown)


ui/src/components/CircleStatus.js
```diff
@@ -11,74 +11,45 @@ import {
...
+        color: "infoPrimary"
+      }
+  }
+}
 
 class CircleStatus extends React.Component {
+  shouldComponentUpdate({ status }) {
+    return this.props.status !== status;
+  }
+
```
JBWatenbergScality: Couldn't we just refactor this to a function component and check if we still have the refresh issue ? If so I think we can play with the component key instead of hacking around with the shouldComponentUpdate lifecycle.

ui/src/components/StatusIcon.js
```diff
@@ -0,0 +1,40 @@
...
+  STATUS_CRITICAL,
+  STATUS_SUCCESS,
+  STATUS_NONE,
+  STATUS_HEALTH,
+} from '../constants.js';
+
+const StatusIcon = styled.i`
+  color: ${(props) => {
+    const theme = props.theme;
+    console.log(theme);
```
JBWatenbergScality:
```suggestion
```

JBWatenbergScality (changes requested):

hug33k:
@JBWatenbergScality I updated support for `aria-label` in `CircleStatus` and also tests related to that prop

ChengYanJin:
hello @hug33k,
The name of the branch is not valid..
You'll have to name it `bugfix/circle-status-rendering-issue` and create another pull request.

PR #3577, opened 2021-10-25
Loki HA

alexandre-allard:
**Component**: salt, logging

**Context**:
As for now, if we lose a Loki instance, fluent-bit stay stuck, waiting for the Loki to come back, meaning if we lose a node, we can't ingest log anymore.

**Summary**:
Use memberlist to spread the Loki hash ring and allow distributor, ingester & querier to communicate for replication.
Fluent-bit is now talking to the Loki service, so it can always push its logs even if a Loki instance is down.

**Acceptance criteria**:


charts/render.py
```diff
@@ -324,6 +381,16 @@ def __call__(self, parser, args, values, option_string=None):
...
+    parser.add_argument(
+        "--patch",
+        action="append",
+        default=[],
+        help=(
+            "Patch the resulting manifest. Format is "
+            "'kind,namespace,name,path,value' where `path` is colon separated "
+            "(e.g. 'metadata:name') and `value` is a JSON object."
+        ),
+    )
```
TeddyAndrieux, line 393: IMHO since we vendor the chart we should instead update the helm chart for that kind of "patch".
Especially here where it is something that could be useful for other people upstream, so to me we should patch our vendored chart and, at least, open a ticket upstream to add this ability in the upstream helm chart (and maybe also do the PR upstream to support it)

salt/metalk8s/addons/logging/loki/deployed/loki-configuration-secret.sls
```diff
@@ -8,6 +8,10 @@
     )
 %}
 
+{%- do loki.spec.config.ingester.lifecycler.ring.update(
+  {"replication_factor": (loki.spec.deployment.replicas / 2) | int + 1}
+) %}
```
TeddyAndrieux: Maybe we do not want to override it if the user sets it from CSC no ?
(to me we should remove it from the default YAML and compute it here if it's not part of the resulting merge of default + CM)

alexandre-allard: Hmm, I don't know if we want to let the user change this value... but OK, I've done as you suggested.

charts/render.py
alexandre-allard: I agree for the upstream ticket, but I'm not sure we should patch our helm charts, because if nothing is done upstream, then we bump this chart version, we will probably forget to re-apply the patches and lose the changes.

alexandre-allard: Reported the request upstream: https://github.com/grafana/helm-charts/issues/759
I'll try to open a PR to tackle it soon.

TeddyAndrieux (dismissed):
LGTM, just we may want to remove "old" loki-X services during upgrade

salt/metalk8s/addons/logging/loki/post-upgrade.sls
```diff
@@ -0,0 +1,17 @@
...
+%}
+
+{%- for index in range(loki.spec.deployment.replicas) %}
+Delete old loki-{{ index }} service object:
+  metalk8s_kubernetes.object_absent:
+    - apiVersion: v1
+    - kind: Service
+    - name: loki-{{ index }}
+    - namespace: metalk8s-logging
+{%- endfor %}
```
TeddyAndrieux, line 17: :+1:

TeddyAndrieux (approved):

charts/render.py
gdemonet: One approach for patching Helm charts I've seen is to actually commit some "patch" files, and have the buildchain check that said patches are applied (and applies them on user request). For instance, that's how Zenko team modified the MongoDB chart from Bitnami (see the `Makefile` and `patches` directory): https://github.com/scality/Zenko/tree/development/2.1/solution-base/mongodb
Keeping the patches in a command-line invocation of some render script, which we try to keep up-to-date in commit messages, is painfully sub-optimal.

TeddyAndrieux: Agree, but I would say that it's painful to retrieve this command-line even without those patches :wink:

alexandre-allard: Yes, not related to this PRs, but we could either go with this patch thing or we could also put the render commands somewhere, then fetch and render at build time instead of having to commit these charts.

PR #3579, opened 2021-10-26
Manage `podAntiAffinity` and number of replicas for CoreDNS

### LINKED ISSUE #3574: Expose `anti-affinity` for coredns deployment
**Component**:

'kubernetes', 'dns'

**Why this is needed**:

Today we deploy coredns with 2 replicas and no affinity which means that all coredns may sit on the same node, which mean no HA for the Kubernetes DNS which likely has an impact on the workload running in the cluster.

**What should be done**:

In order to have HA of coredns we need to make sure the coredns replica sits in different "failure zone". 
But this cannot be done automatically and need to be input from the user, so we need to expose a way to change this.

We likely want to also expose the number of coredns replicas.

Add an entry in the bootstrap config for coredns so that user can input:
- the number of replicas for coredns (default to 2)
- the topologyKey to use for anti-affinity (soft and hard) of coredns Pods
  suggested default (TBD) (soft anti-affinity on hostname):
  ```yaml
   podAntiAffinity:
     preferredDuringSchedulingIgnoredDuringExecution:
     - weight: 100
       podAffinityTerm:
         labelSelector:
           matchLabels:
             k8s-app: kube-dns
         namespaces:
         - kube-system
         topologyKey: kubernetes.io/hostname
  ```

Since coredns pod will not be re-scheduled during the life of the platform in order to be sure our "soft" anti-affinity gets applied when it's possible we will need to trigger a "rollout restart" of the coredns when we expand MetalK8s cluster with a new master node.

Check kubernetes docs for more information about affinity and anti-affinity https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity

TeddyAndrieux:
**Component**:

'dns'

**Context**: 

See: #3574 

**Summary**:

Add entry in the bootstrap configuration file to manage the number of replicas to deploy for CoreDNS and the `podAntiAffinity`.

So that users can set that anti-affinity according to the platform and failure scenario he want to support

---

Fixes: #3574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@
   (PR[#3540](https://github.com/scality/metalk8s/pull/3540))
 
 - [#3574](https://github.com/scality/metalk8s/issues/3574) - Allow to manage
-  soft and hard `podAntiAffinity` for `CoreDNS` from Bootstrap configuration
-  file, with a default soft anti-affinity on hostname, so that if it's
-  possible each `CoreDNS` pods will sit on different infra node
+  number of replicas and, soft and hard `podAntiAffinity` for `CoreDNS`
+  from Bootstrap configuration file, with a default soft anti-affinity on
+  hostname, so that if it's possible each `CoreDNS` pods will sit on different infra node
   (PR[#3579](https://github.com/scality/metalk8s/pull/3579))
 
 ### Removals
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
   when observing the cluster state (used in the UI Dashboard page)
   (PR[#3540](https://github.com/scality/metalk8s/pull/3540))
 
+- [#3574](https://github.com/scality/metalk8s/issues/3574) - Allow to manage
+  soft and hard `podAntiAffinity` for `CoreDNS` from Bootstrap configuration
+  file, with a default soft anti-affinity on hostname, so that if it's
+  possible each `CoreDNS` pods will sit on different infra node
+  (PR[#3579](https://github.com/scality/metalk8s/pull/3579))
+
 ### Removals
 
 - Removed the PDF support for documentation, replaced it with the HTML output
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -71,6 +71,7 @@ Configuration
           featureGates:
             <feature_gate_name>: True
         coreDNS:
+          replicas: 2
           podAntiAffinity:
             hard: []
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
   when observing the cluster state (used in the UI Dashboard page)
   (PR[#3540](https://github.com/scality/metalk8s/pull/3540))
 
+- [#3574](https://github.com/scality/metalk8s/issues/3574) - Allow to manage
+  number of replicas and, soft and hard `podAntiAffinity` for `CoreDNS`
+  from Bootstrap configuration file, with a default soft anti-affinity on
```
gdemonet, line 11: Why do we want this to live in Bootstrap config? Is CoreDNS required for the control plane to function, or would a CSC ConfigMap (yeah I know, we want to kill CSC) work?

TeddyAndrieux: CSC could work as well for this, but ... yeah first we want to kill it, and also to me it's simpler to provide it at "installation" than doing installation then edit a configmap then apply this configmap changes.

And btw to me (I may be wrong), I consider bootstrap config file as configuration for Core stuff (so CoreDNS is part of it) and CSC as "services stuff".
Bootstrap config should likely change a bit and we should have it in Kubernetes (and likely do no longer call it bootstrap).

Anyway, no strong strong opinion, to me today that kind of config could be in CSC or bootstrap, it does not really matter, just on my side I like the fact that I can configure this at installation and not .... you know deploy something I do not want to then change it so that it matches what I want

docs/installation/bootstrap.rst
```diff
@@ -159,10 +165,28 @@ the bootstrap script is executed, those ISOs are automatically mounted and the
...
-``kubernetes.apiServer.featureGates`` mapping.
+Kubernetes `Feature Gates`_ to enable or disable and if you are ok with
+defaults kubernetes configuration.
+
+  If you need to enable or disable specific features for ``kube-apiserver``
+  configure the corresponding entries in the
+  ``kubernetes.apiServer.featureGates`` mapping.
+
+  If you want to override the default ``coreDNS`` podAntiAffinity or number of
+  replicas, by default MetalK8s deploy 2 replicas and use soft podAntiAffinity
```
alexandre-allard-scality, line 176:
```suggestion
  replicas. By default, MetalK8s deploy 2 replicas and use soft podAntiAffinity
```

alexandre-allard (dismissed):

alexandre-allard (approved):

PR #3584, opened 2021-10-28
salt: fix backup replication job name

alexandre-allard:
**Component**: salt, backup

**Context**:
We were using the nodeName in the Job name but the nodeName can be very long, meaning the Job name could exceed the limit of 63 characters, thus making it impossible to run.

**Summary**:
We now generate a random ID of 15 chars.
It also allows to keep the previous backup Job a little longer (1 day), to be able to retrieve logs from the previous replication run without any risk of conflict if a new one had to be scheduled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@
 - Bump Kubernetes version to 1.21.6
   (PR[#3583](https://github.com/scality/metalk8s/pull/3583))
 
+## Bug fixes
+
+- Fix the backup replication Job name which was including the node name,
+  so that he could exceed the limit of 63 characters.
+  (PR[#3584](https://github.com/scality/metalk8s/pull/3584))
+
+- Fluent-bit instances stayed stuck when a Loki instance was down, blocking
+  the whole logging pipeline. It is now fixed as we configure fluent-bit to
+  talk with Loki's service and use memberlist to keep the Loki instances
+  replicated.
+  (PR[#3557](https://github.com/scality/metalk8s/pull/3557))
+
 ## Release 2.10.4
 ## Bug fixes
 
--- a/salt/metalk8s/orchestrate/backup/replication.sls
+++ b/salt/metalk8s/orchestrate/backup/replication.sls
@@ -10,6 +10,7 @@ Schedule backup replication Job on {{ node }}:
     - name: salt://{{ slspath }}/files/job.yaml.j2
     - template: jinja
     - defaults:
+        name: backup-replication-{{ salt.random.get_str(length=15) | lower }}
         node: {{ node }}
         image: {{ image }}
 
--- a/salt/tests/unit/formulas/config.yaml
+++ b/salt/tests/unit/formulas/config.yaml
@@ -298,6 +298,7 @@ metalk8s:
           _cases:
             "Create job manifest for a node":
               extra_context:
+                name: replication-backup-012345689abcde
                 node: master-1
                 image: registry/some-image-name:tag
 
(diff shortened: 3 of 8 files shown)


salt/metalk8s/orchestrate/backup/replication.sls
```diff
@@ -10,6 +10,7 @@ Schedule backup replication Job on {{ node }}:
     - name: salt://{{ slspath }}/files/job.yaml.j2
     - template: jinja
     - defaults:
+        name: backup-replication-{{ salt.random.get_str(length=15) }}
```
TeddyAndrieux: Apparently, you also need to add it to formula unit tests

alexandre-allard: Yep, true

TeddyAndrieux (dismissed):

salt/metalk8s/orchestrate/backup/replication.sls
```diff
@@ -10,6 +10,7 @@ Schedule backup replication Job on {{ node }}:
     - name: salt://{{ slspath }}/files/job.yaml.j2
     - template: jinja
     - defaults:
+        name: backup-replication-{{ salt.random.get_str(length=15, uppercase=False, punctuation=False) }}
```
TeddyAndrieux: those "uppsercase" and "punctuation" are only available in Salt 3004, to me you should use `lower` jinja filter should be ok (need also to check that in salt 3002.7 we do not have punctuation in the random str

salt/tests/unit/formulas/config.yaml
```diff
@@ -298,6 +298,7 @@ metalk8s:
           _cases:
             "Create job manifest for a node":
               extra_context:
+                name: replication-backup-012345689ABCDE
```
TeddyAndrieux: nitpick it should be lowercase since uppercase is not supported :smile:

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@
...
   (PR[#3583](https://github.com/scality/metalk8s/pull/3583))
 
+## Bug fixes
+
+- Fix the backup replication Job name which was including the node name,
+  so that he could exceed the limit of 63 characters.
+  (PR[#3584](https://github.com/scality/metalk8s/pull/3584))
+
+- Fluent-bit instances stay stuck when a Loki instance is down, blocking
+  the whole logging pipeline.
```
TeddyAndrieux: nitpick to me it should be re-phrase a bit because here it does not explicitly mention that it's fix now

alexandre-allard: It's in the bugfix list, of course it's fixed. :)
I'll update it.

salt/metalk8s/orchestrate/backup/replication.sls
alexandre-allard: True, should have changed the documentation version :facepalm: 
Apparently there is only aphanum chars in 3002.7, so I'll go with a simple lower Jinja filter.

TeddyAndrieux (approved):

salt/metalk8s/orchestrate/backup/replication.sls
```diff
@@ -10,6 +10,7 @@ Schedule backup replication Job on {{ node }}:
     - name: salt://{{ slspath }}/files/job.yaml.j2
     - template: jinja
     - defaults:
+        name: backup-replication-{{ salt.random.get_str(length=15) | lower }}
```
NicolasT, line 13: Instead of calculating a random name here, can't you use K8s' built-in `generateName` feature?

TeddyAndrieux: Oh!!! Didn't know about this feature, but it's exactly what we want indeed :laughing:

TeddyAndrieux: Ok I will change it in another PR after

alexandre-allard: Oh, didn't know either, thanks for the tip.

PR #3606, opened 2021-11-18
ci: Adapt tests for the new MetalK8s versioning

TeddyAndrieux:
**Component**:

ci

**Context**: 

Adapt CI tests for the new MetalK8s versioning

**Summary**:

- Enable back the major version downgrade for this major version (as today it's supported)
- Change development branch tests to test only the major version upgrade/downgrade
- Add major version upgrade/downgrade test using snapshots

---

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -759,8 +759,8 @@ models:
       name: clone previous version branch
       command: >
           git clone "%(prop:repository)s"
-          --branch "development/%(prop:product_version_prev)s"
-          metalk8s-"%(prop:product_version_prev)s"
+          --branch "%(prop:prev_branch)s"
+          metalk8s-prev
       haltOnFailure: true
 
   - ShellCommand: &generate_report_over_ssh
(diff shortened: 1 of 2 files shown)


eve/main.yml
```diff
@@ -1924,27 +1925,32 @@ stages:
           command: >
               major=$(echo "%(prop:product_version)s" | cut -d'.' -f1) &&
               minor=$(echo "%(prop:product_version)s" | cut -d'.' -f2) &&
-              git tag --sort=taggerdate --list "$major.$(( $minor-1 )).*" | tail -1
+              git tag --sort=taggerdate --list "$major.$(( $minor-1 )).*" | grep -v '\-' | tail -1
```
alexandre-allard, line 1928: Just so you know, you can also use `--` option to provide a pattern starting with a `-`:
```suggestion
              git tag --sort=taggerdate --list "$major.$(( $minor-1 )).*" | grep -v -- - | tail -1
```

alexandre-allard (approved):
Not tested, but LGTM.

eve/main.yml
TeddyAndrieux: Ohh ok good to know :+1:, but I will not do it here right now (do not want to wait for another build) :smile:

PR #3620, opened 2021-11-30
salt,charts: Grafana sidecar picks up dashboards in all namespaces

### LINKED ISSUE #3618: Pick up dashboards (as ConfigMaps) from all namespaces
**Component**: charts,salt

**Why this is needed**: Currently, the deployed Grafana pod is configured to pick up any ConfigMap with the `grafana_dashboard` label available in its own namespace, i.e. `metalk8s-monitoring`. 

For external solutions, we want:

* to be able to deploy dashboard ConfigMaps with the `grafana_dashboard` label in a solution's namespace and have them show up in the `metalk8s-monitoring` Grafana.
* to be able to control which folders a solution's dashboards are stored into

**What should be done**: modify chart values for grafana, re-render the salt state.

**Test plan**: E2E confirmation to be provided.

jbertran:
**Component**: salt,charts

**Context**: Allowing solutions to expose dashboards that get picked up by metalk8s' Grafana deployment.

---

Closes: #3618

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,12 @@
 
 - Handle 401 unauthorized error in MetalK8s UI
   (PR[#3640](https://github.com/scality/metalk8s/pull/3640))
+  
+- [#3618](https://github.com/scality/metalk8s/issues/3618) Detect Grafana
+  dashboard ConfigMaps in any namespace rather than just `metalk8s-monitoring`,
+  and enable Grafana folder generation from dashboard file structure (PR
+  [#3620](https://github.com/scality/metalk8s/pull/3620))
+
 ## Bug fixes
 
 - [#3601](https://github.com/scality/metalk8s/issues/3601) - Marks
--- a/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
@@ -25022,13 +25022,12 @@ data:
     providers:
     - name: 'sidecarProvider'
       orgId: 1
-      folder: ''
       type: file
       disableDeletion: false
       allowUiUpdates: false
       updateIntervalSeconds: 30
       options:
-        foldersFromFilesStructure: false
+        foldersFromFilesStructure: true
         path: /tmp/dashboards
 kind: ConfigMap
 metadata:
@@ -61885,7 +61884,7 @@ spec:
           apiVersion="v1", namespace="metalk8s-monitoring", name="prometheus-operator-grafana",
           path="data:grafana.ini")
         checksum/dashboards-json-config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
-        checksum/sc-dashboard-provider-config: 1f57959a5b236ca002fed3fe494a562bc462eb0db530ba1500664546d01cb395
+        checksum/sc-dashboard-provider-config: 4b86434bc5d3344f34c6418fe3e431b5dbd5ffb26f52b48a8325c63fff74a1f8
         checksum/secret: 05ea72d91e6ac99b8da34f36157887ab504aba30dde8e63b1d06b3527f889a6e
       labels:
         app.kubernetes.io/instance: prometheus-operator
@@ -61902,6 +61901,10 @@ spec:
           value: /tmp/dashboards
         - name: RESOURCE
           value: both
+        - name: NAMESPACE
+          value: ALL
+        - name: FOLDER_ANNOTATION
+          value: metalk8s.scality.com/grafana-folder-name
         image: {% endraw -%}{{ build_image_name("k8s-sidecar", False) }}{%- raw %}:1.14.2
         imagePullPolicy: IfNotPresent
         name: grafana-sc-dashboard
--- a/tests/post/features/monitoring.feature
+++ b/tests/post/features/monitoring.feature
@@ -62,3 +62,12 @@ Feature: Monitoring is up and running
         And the Grafana API is available
         When we put a datasource as ConfigMap named 'test-new-ds' in namespace 'metalk8s-monitoring'
         Then we have a datasource named 'test-new-ds'
+
+    Scenario: Inserting a new ConfigMap provides a new dashboard
+        Given the Grafana API is available
+        And the 'zenko' namespace exists
+        And we have a dashboard ready with a specific UID
+        When we put the dashboard as a ConfigMap in namespace 'zenko' with folder annotation value 'ZENKO-PLEASE'
+        Then we have the dashboard in folder 'ZENKO-PLEASE' in Grafana
+        Then on deletion of the dashboard's ConfigMap from namespace 'zenko'
+        Then the dashboard is no longer in folder 'ZENKO-PLEASE' in Grafana
--- a/tests/post/steps/test_monitoring.py
+++ b/tests/post/steps/test_monitoring.py
@@ -1,15 +1,13 @@
 import json
 import operator
 import pathlib
-import random
-import string
+import uuid
 import yaml
 
 from kubernetes.client.rest import ApiException
 
 import pytest
 from pytest_bdd import scenario, given, when, then, parsers
-import testinfra
 
 from tests import utils
 from tests import kube_utils
@@ -92,8 +90,16 @@ def test_configmap_add_datasource(host, teardown):
     pass
 
 
+@scenario(
+    "../features/monitoring.feature",
+    "Inserting a new ConfigMap provides a new dashboard",
+)
+def test_configmap_add_dashboards(host):
+    pass
+
+
 # }}}
-# Fixture {{{
+# Fixtures {{{
 
 
 @pytest.fixture(scope="function")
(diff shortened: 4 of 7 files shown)


charts/kube-prometheus-stack/charts/grafana/values.yaml
```diff
@@ -630,12 +630,12 @@ sidecar:
...
     # Otherwise the namespace in which the sidecar is running will be used.
     # It's also possible to specify ALL to search in all namespaces
-    searchNamespace: null
+    searchNamespace: ALL
     # search in configmap, secret or both
     resource: both
     # If specified, the sidecar will look for annotation with this name to create folder and put graph here.
     # You can use this parameter together with `provider.foldersFromFilesStructure`to annotate configmaps and create folder structure.
-    folderAnnotation: null
+    folderAnnotation: metalk8s.scality.com/grafana-folder-name
```
gdemonet: Is the `-name` suffix needed? I mean, do we expect some other annotations regarding Grafana folders?

jbertran: We don't, it just seems better to me to explain what impact this annotation has on the folder. I can change it if you'd like.

salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
```diff
@@ -59893,7 +59896,7 @@ spec:
       expr: (kube_node_spec_taint{job="kube-state-metrics",key="node.kubernetes.io/unreachable",effect="NoSchedule"}
         unless ignoring(key,value) kube_node_spec_taint{job="kube-state-metrics",key=~"ToBeDeletedByClusterAutoscaler|cloud.google.com/impending-node-termination|aws-node-termination-handler/spot-itn"})
         == 1
-      for: 5m
+      for: 15m
```
jbertran: Note: I can't figure out where this comes from - it doesn't look like that's from my diff, perhaps a value change that occurred without regeneration?

lavigne958: yes it is part of your commit, perhaps a small typo ??
https://github.com/scality/metalk8s/commit/88030820f89ecbdeeefce932b0ad12142c479d54

NicolasT: I'll take a look.

NicolasT: Looks like re-rendering the chart requires some 'special care': see the commit message of https://github.com/scality/metalk8s/commit/1c88876d2100baf04c898232f30531ab62f8f3c8

NicolasT (dismissed):
Chart needs to be re-rendered with the 'right' command (see other comment).

charts/kube-prometheus-stack/charts/grafana/values.yaml
```diff
@@ -630,12 +630,12 @@ sidecar:
     # If specified, the sidecar will search for dashboard config-maps inside this namespace.
```
TeddyAndrieux: Any reason to edit the grafana helm value file and not the `charts/kube-prometheus-stack.yaml` helm option file directly ?

TeddyAndrieux (dismissed):

charts/kube-prometheus-stack/charts/grafana/values.yaml
NicolasT: Heh, good point, indeed one isn't supposed to make changes to the vendored files unless really necessary.

jbertran: Oh. That's just my weak knowledge of Helm. I saw `Values`, so I edited `Values` :sweat_smile:

NicolasT (commented):
Couple of other things:
- Needs CHANGELOG entry
- Likely needs some small documentation in the dev guide
- May benefit from a test?

jbertran:
> Couple of other things:
> 
> * Needs CHANGELOG entry
> 
> * Likely needs some small documentation in the dev guide
> 
> * May benefit from a test?

Changelog, docs update: done.
Tests: since the effect of this change is local to a generated `chart.sls` where the diff does not depend on Jinja to be rendered, I don't know what to test here.

gdemonet:
@jbertran For the test, I guess we'd build something E2E, creating a ConfigMap in some namespace, and check that Grafana reflects these changes. We can see how to implement it together, if you want.

jbertran:
> @jbertran For the test, I guess we'd build something E2E, creating a ConfigMap in some namespace, and check that Grafana reflects these changes. We can see how to implement it together, if you want.

It feels a bit like we're testing that various third party components work as advertised (basically `k8s-sidecar` provided directory structure and Grafana picking it up), but I can do E2E for sure - it's pytest so I'm not too lost there. I'll set up a bunch of namespaces and configmaps, push them and verify dashboards _and_ folders present. Does that sound sufficient?

NicolasT:
> It feels a bit like we're testing that various third party components work as advertised (basically k8s-sidecar provided directory structure and Grafana picking it up)

Well, yes indeed. If this functionality, no matter how it's provided, is part of the expected functionality of the project, it should be tested (granted, there's tests lacking in many places).
It's not really about "testing third parties", rather ensuring that, e.g., if upstream changes how they do things, and we update the chart(s), then we detect this change and must make relevant changes in MetalK8s to cope with it, ideally continuing to expose the feature as it is/was, otherwise changing the feature (kinda breaking change).

Charts and `values.yaml` have been... unstable APIs in the past, so we really shouldn't rely on upstream not 'breaking' things (sometimes involuntary).

CHANGELOG.md
```diff
@@ -69,6 +69,11 @@
   highly-available when we have multiple Prometheus instances
   (PR[#3573](https://github.com/scality/metalk8s/pull/3573))
 
+- [#3618](https://github.com/scality/metalk8s/issues/3618) Detect Grafana
+  dashboard ConfigMaps in any namespace rather than just `metalk8s-monitoring`,
+  and enable Grafana folder generation from dashboard file structure (PR
+  [#3620](https://github.com/scality/metalk8s/pull/3620/commits))
```
TeddyAndrieux:
```suggestion
  [#3620](https://github.com/scality/metalk8s/pull/3620))
```

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +129,169 @@ def _check_object_exists():
...
+    namespaces = client.get().items
+    try:
+        next(
+            filter(lambda ns: ns.get("metadata", {}).get("name") == ns_name, namespaces)
+        )
+    except StopIteration:
+        try:
+            client.create({"metadata": {"name": ns_name}})
+        except ApiException as err:
+            raise AssertionError(f"Unable to create namespace: {err}")
```
TeddyAndrieux: Why iterate on all namespaces ?
You can retrieve only one (and create it if it does not exist)
Something like that
```suggestion
    try:
      client.get(name=ns_name)
    except ApiException as err:
      if err.status == 404:
        client.create({"metadata": {"name": ns_name}})
```

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +129,169 @@ def _check_object_exists():
...
+        raise AssertionError(f"no folder named '{folder_name}'")
+    dashboards = grafana_api.get_dashboards(folder_id=folder["id"])
+    try:
+        next(filter(lambda dash: dash["title"] == dash_name, dashboards))
+    except StopIteration:
+        raise AssertionError(
+            f"found no dashboard named '{dash_name}' in '{folder_name}'"
+        )
+    else:
+        return True
```
TeddyAndrieux: Maybe to make it simpler we could simply use the uid directly since we can set this one manually in the dashboard.

To make this test really "realistic" we could generate an `uuid` at the beginning of the test (put it in the context) and then we create a dashboard using this `uuid` as uid

tests/post/features/monitoring.feature
```diff
@@ -56,3 +56,16 @@ Feature: Monitoring is up and running
...
+        And the 'zenko' namespace exists
+        When we put a dashboard as ConfigMap named 'zk-dash' in namespace 'zenko' with folder annotation value 'ZENKO-PLEASE'
+        Then we have a dashboard named 'zk-dash' in folder 'ZENKO-PLEASE'
+
+    Scenario: Removing a ConfigMap removes its dasboard
+        Given the Grafana API is available
+        And the 'zenko' namespace exists
+        And the 'zk-dash' dashboard exists in folder 'ZENKO-PLEASE'
+        When we delete the 'zk-dash' ConfigMap in namespace 'zenko'
+        Then we have no dashboard named 'zk-dash' in folder 'ZENKO-PLEASE'
```
TeddyAndrieux: Personally do not really like the fact that this test need the previous one to succeed.

To me we could have a single test with 2 "When - Then" (but maybe it's not a good practice I don't know)
Something like that 
```
Scenario: Inserting a new ConfigMap provides a new dashboard
    Given the Grafana API is available
    And the 'zenko' namespace exists
    When we put a dashboard as ConfigMap named 'zk-dash' in namespace 'zenko' with folder annotation value 'ZENKO-PLEASE'
    Then we have a dashboard named 'zk-dash' in folder 'ZENKO-PLEASE'
    When we delete the 'zk-dash' ConfigMap in namespace 'zenko'
    Then we have no dashboard named 'zk-dash' in folder 'ZENKO-PLEASE'
```

TeddyAndrieux (dismissed):
LGTM just comment on test :smile:

tests/post/features/monitoring.feature
jbertran: How about I switch this in the second test?
```         
- And the 'zk-dash' dashboard exists in folder 'ZENKO-PLEASE'
+ And the 'zk-dash' ConfigMap exists and deploys the 'zk-dash' dashboards in folder 'ZENKO-PLEASE'
```
This way we can deploy the ConfigMap if it's not available.

The goal of the second one is to make the first one's insertion of a strictly test-related ConfigMap (maybe there's an easy-to-use cleanup hook? I didn't find it :slightly_frowning_face: )

TeddyAndrieux: But with the suggestion, I send to you the CM will be deleted at the end of the test also.

If you just want some clean up at the end of the test it's doable with some teardown  (you can check `tests/post/steps/test_ingress.py` we do some simple teardown)

francoisferrand (dismissed):

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +134,152 @@ def _check_object_exists():
...
+
+    cm_manifest["metadata"]["annotations"] = {
+        "metalk8s.scality.com/grafana-folder-name": cm_folder,
+    }
+    cm_manifest["metadata"]["namespace"] = namespace
+
+    try:
+        next(
+            filter(lambda cm: cm.get("metadata", {}).get("name") == cm_name, configmaps)
+        )
```
gdemonet: Why do we get the whole list if we only care about one?

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +134,152 @@ def _check_object_exists():
     utils.retry(_check_object_exists, times=20, wait=3)
 
 
+@given(parsers.parse("the '{ns_name}' namespace exists"))
+def namespace_exists(k8s_client, ns_name):
+    client = k8s_client.resources.get(api_version="v1", kind="Namespace")
+    try:
+        client.get(name=ns_name)
+    except ApiException:
```
gdemonet: This may need more specificity: we should only attempt to POST if GET returns a 404. Multiple issues may happen otherwise, such as an attempt to create after the GET returned a temporary 500, which will fail even if the API is back in a good shape.

gdemonet: Also, could be nice to wrap these in retries, just in case... Well I guess we'll do it if we see the tests failing.

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +134,152 @@ def _check_object_exists():
...
+            client.delete(name=cm_name)
+        except ApiException as err:
+            raise AssertionError(f"Unable to delete existing ConfigMap: {err}")
+    except StopIteration:
+        pass
+    finally:
+        try:
+            client._create(cm_manifest)
+        except ApiException as err:
+            raise AssertionError(f"Unable to push ConfigMap: {err}")
```
gdemonet: Why do we use a `finally:` block here? If the API is in a bad shape for some reason and refuses the deletion (or even the retrieval, which isn't wrapped in `try/except` block for now), we'll still try to create the ConfigMap.

tests/post/steps/test_monitoring.py
```diff
@@ -119,10 +134,152 @@ def _check_object_exists():
...
+    cm_manifest = context["grafana-cm-manifest"]
+    cm_name = cm_manifest.get("metadata", {}).get("name")
+    try:
+        client.delete(cm_name)
+    except ApiException as err:
+        raise AssertionError(f"failed to delete dashboard '{cm_name}': {err}")
+
+
+@then(parsers.parse("the dashboard is no longer in folder '{folder_name}' in Grafana"))
+def check_no_dash_in_folder(context, gf_folder_name, grafana_api):
```
gdemonet: Should be:

```suggestion
@then(parsers.parse("the dashboard is no longer in folder '{gf_folder_name}' in Grafana"))
def check_no_dash_in_folder(context, gf_folder_name, grafana_api):
```

I imagine?

gdemonet (changes requested):
A few things to change w.r.t. API error-handling, but LGTM otherwise! Thank you :+1:

tests/post/steps/test_monitoring.py
jbertran: Got lost in the control flow :sweat: fixed in `c60bea50e`

tests/post/steps/test_monitoring.py
jbertran: fixed in `c60bea50e` - leaving the retries out for now

tests/post/steps/test_monitoring.py
jbertran: fixed in `c60bea50e`

tests/post/steps/test_monitoring.py
jbertran: Oops, fixed in `c60bea50e`

gdemonet (commented):
LGTM :ok_hand: I'll approve once squashed!

jbertran:
@gdemonet done

gdemonet (approved):

tests/post/steps/test_monitoring.py
```diff
@@ -148,6 +153,98 @@ def _check_object_exists():
...
+            raise AssertionError(f"Unable to create namespace: {exc}")
+        try:
+            client.create({"metadata": {"name": ns_name}})
+        except ApiException as err:
+            raise AssertionError(f"Unable to create namespace: {err}")
+
+
+@given(parsers.parse("we have a dashboard ready with a specific UID"))
+def dashboard_ready(context):
+    cm_uid = str(uuid.uuid4())
```
gdemonet, line 173: Missing import for this module

gdemonet (changes requested):

tests/post/steps/test_monitoring.py
jbertran: :man_facepalming: conflict resolution gone wrong

gdemonet (dismissed):

lavigne958 (dismissed):

lavigne958 (approved):

gdemonet (approved):

PR #3624, opened 2021-12-02
Marks `pause` image as pod infra container image in kubelet (and only use image from embedded repos

### LINKED ISSUE #3601: "pause" image might be removed when a node goes in "DiskPressure"
**Component**:

'kubernetes'

**What happened**:

If for whatever reason a node goes in "DiskPressure" kubelet will remove the "pause" image, which means if the node does not have internet access, new containers will not be able to get scheduled on this node

**What was expected**:

Pause image should **NOT** get removed

See: https://github.com/kubernetes/kubernetes/issues/81756
We have to set `--pod-infra-container-image` for kubelet

TeddyAndrieux:
We mark the `pause` image used by containerD as `pod infra container
image` for kubelet, so that kubelet does not remove this image.

We also put `pause` and `nginx` images (used by the registry pod) in the embedded registry,
so that we are sure that the node does not fetch this image from the internet.

NOTE: Since we fetch the pause image from the embedded registry we no
longer need to import it manually with salt on every node but only for
the registry node.

---

Fixes: #3601


CHANGELOG.md
```diff
@@ -69,6 +69,13 @@
   highly-available when we have multiple Prometheus instances
   (PR[#3573](https://github.com/scality/metalk8s/pull/3573))
 
+## Bug fixes
+
+- [#3601](https://github.com/scality/metalk8s/issues/3601) - Mark
+  the `pause` image used by ContainerD as `pod infra container image`
```
NicolasT:
```suggestion
  the `pause` image used by `containerd` as `pod infra container image`
```

salt/_modules/containerd.py
```diff
@@ -26,6 +26,7 @@ def load_cri_image(path):
         Path of the Docker image archive to load
     """
     log.info('Importing image from "%s" into CRI cache', path)
-    return __salt__["cmd.run_all"](
-        'ctr --debug -n k8s.io image import "{0}"'.format(path)
-    )
+    cmd = 'ctr --debug -n k8s.io image import "{}"'.format(path)
+    if fullname:
+        cmd += ' --index-name "{}"'.format(fullname)
```
NicolasT, line 31: Why is this needed whilst it wasn't before?

TeddyAndrieux: It's to import an index name for the image, it was not needed before because we used the repo in the "repositories" file in the tar archive
```json
# cat repositories
{"k8s.gcr.io/pause":{"3.2":"a38657cdc544bd8c100ef952fc60e31709509b0bc4f31804ab107fe7fd0e6f4a"}}
```

So several options:
1- At build time we put the right repo in the tar for the image (which mean we "compute" the repository name at build time, where today it's at run time in salt)
2- At run time we create the tar file containing the right `repositories` file with the repo (it's a bit of work...)
3- We use `--index-name` so that in the cache we have an index on the pause image

I choose the last option as it was the simplest to do, but we can still change this if we need to

NicolasT: OK, thanks for explaining. Maybe consider putting such intel in commit message in the future :)

NicolasT (approved):

salt/_modules/containerd.py
TeddyAndrieux: Right

PR #3625, opened 2021-12-02
Improvement/remove health in selector

gaudiauj:
**Component**:

active alerts filter

**Context**: 

there is no reason to have an healthy filter in the alert therefore we remove it from the dropdown

**Acceptance criteria**: 

healthy shouldn't be visible on the dropdown.


ChengYanJin:
Update: We decided to rework the health selector in Core-ui with the lastest design, so this PR is on hold.

PR #3627, opened 2021-12-06
Improvement/user react query in volume page

gaudiauj:
**Component**:

volumes pages

**Context**: 

In order to have a unified experience arround all the scality product, it was decided to use react-query rather than redux / redux saga. 

**Summary**:

This PR is the begining of using react-query in the UI of metalk8s. 
The first step is to migrate all the api call from redux-saga to react-query but still use redux as the state manager.

This PR aim to change the refresh call in the volume page. 

The next steps are : 

- add integration test. 
- migrate the other call to react-query
- get the data directly from react-query without using redux

**Acceptance criteria**: 

same comportement than before


ui/src/containers/AlertProvider.js
```diff
@@ -43,13 +44,18 @@ const InternalAlertProvider = ({
     (state) => state.config.api.url_alertmanager,
   );
 
+  const queryClient = new QueryClient();
+
   return (
-    <FederatedComponent
+    <QueryClientProvider client={queryClient}>
```
JBWatenbergScality: We are using QueryClientProvider of shell ui in order to share cache with other UIs. I think the pb you met is likely due to the update not being performed on shell-ui side.

ui/src/ducks/app/volumes.hooks.js
```diff
@@ -0,0 +1,28 @@
...
+export function useRefreshVolume() {
+  const dispatch = useDispatch();
+  const { data } = useQuery(
+    ['volumes'],
+    Metalk8sVolumesApi.getMetalk8sV1alpha1VolumeList,
+    {
+      select: (data) => data.body?.items,
+      staleTime: fiveSecondsInMs,
+      refetchInterval: TenSecondsInMs,
+    },
```
JBWatenbergScality: I think we already have a query to retrieve the volumes in shell-ui platform library. This query should reuse the same  query to mutualise the cache.

ui/src/ducks/app/volumes.hooks.js
```diff
@@ -0,0 +1,28 @@
...
+  );
+
+  useEffect(() => {
+    if (data) {
+      dispatch(setVolumesAction(data));
+    }
+  }, [data, dispatch]);
```
JBWatenbergScality: I think we can try to drop redux here and then I would question the added value of this custom hook. To me it would make sense to have a `useVolumes()` hook but then it's returned data should match the target format (including usage from prometheus and so on).

ui/src/ducks/app/volumes.hooks.js
```diff
@@ -20,9 +20,34 @@ export function useRefreshVolume() {
...
   useEffect(() => {
     if (data) {
       dispatch(setVolumesAction(data));
     }
   }, [data, dispatch]);
+
+  return result;
+}
+
+export function useFetchCurrentVolumeObject(volumeName: string) {
```
JBWatenbergScality: I'm not sure refactoring using "simple" Fetch hooks is the right approach. I would more see what is the expected data output, what is the input and then build a hook abstracting away fetching, data conversion and formatting. Kind of what I tried to initialize with the `useVolumesWithAlerts` hook.

JBWatenbergScality (changes requested):

ui/package-lock.json
```diff
@@ -17885,9 +17885,9 @@
       "integrity": "sha512-fBASbA6LnOU9dOU2eW7aQ8xmYBSXUIWr+UmF9b1efZBazGNO+rcXT/icdKnYm2pTwcRylVUYwW7H1PHfLekVzA=="
     },
     "react-query": {
-      "version": "3.33.7",
-      "resolved": "https://registry.npmjs.org/react-query/-/react-query-3.33.7.tgz",
-      "integrity": "sha512-uL/1ZatxcRyvfhbNPeuOlLVbcdMQKzQ2t+SQLJMrMM0plfa/Al3buaPbGsbsmIwvyWeI4CVkkSBKzQIkU4p9MQ==",
+      "version": "3.34.0",
```
ChengYanJin, line 17893: Likely there is a change in ui/package.json too?

Oh. I see it in the previous commits. 
Would ease the review process if we can have clean commits ;)

ui/src/ducks/app/monitoring.hooks.js
```diff
@@ -1,48 +1,61 @@
...
+import { prometheusKey } from '../../services/prometheus/prometheus.key';
 import { REFRESH_METRICS_GRAPH } from '../../constants';
 import { useDispatch } from 'react-redux';
 import { updateCurrentVolumeStatsAction } from './monitoring';
 
-export function useFetchCurrentVolumeStats() {
-  const dispatch = useDispatch();
-
+export function getVolumeLatencyCurrentQueryOption() {
   const volumeLatencyCurrentQuery = `irate(node_disk_io_time_seconds_total[1h]) * 1000000`;
```
ChengYanJin: (Maybe not directly related to your PR)

I think the correct PromQL query would be `irate(node_disk_io_time_seconds_total[1h])`
The reason we add  ***1000000** is that we hardcode **ms** as the unit of the latency. 

It's a very bad practice IMO, the unit should be added based on the value.

ui/src/services/k8s/volumes.key.js
```diff
@@ -0,0 +1,5 @@
+export const volumesKey = {
+  all: ['volumes', 'list'],
+  volumeObject: ['volumes', 'object'],
+  persitant: ['volumes', 'Persistant'],
```
ChengYanJin: persistentVolumes:  [''persistentVolumes"]

ui/src/ducks/app/monitoring.hooks.js
```diff
@@ -1,48 +1,61 @@
...
 import { queryPrometheus } from '../../services/prometheus/api';
+import { prometheusKey } from '../../services/prometheus/prometheus.key';
 import { REFRESH_METRICS_GRAPH } from '../../constants';
 import { useDispatch } from 'react-redux';
 import { updateCurrentVolumeStatsAction } from './monitoring';
 
-export function useFetchCurrentVolumeStats() {
-  const dispatch = useDispatch();
-
+export function getVolumeLatencyCurrentQueryOption() {
```
ChengYanJin: I don't think it makes sense to keep the functions that return the query option under the `/ducks/` folder.
  
We probably should put them under `/services/queries/volumes`
WDYT @JBWatenbergScality @MonPote ?

ui/src/services/k8s/core.key.js
```diff
@@ -0,0 +1,3 @@
+export const nodeKey = {
```
ChengYanJin, line 1: The keys should live next to the queries too.

ui/src/ducks/app/nodes.hook.js
```diff
@@ -0,0 +1,97 @@
+import { useEffect } from 'react';
```
ChengYanJin: I am not convinced to put the hook under `/src/ducks/`

ChengYanJin (dismissed):
Some comments about codebase structure.
We tend to put everything related to redux under `/ducks` folder, that's why I am not convinced to add queries and hooks here.

ui/package-lock.json
ChengYanJin: I guess we miss the change in `ui/package.json` ;)

And since we bump the version of React Query in shell UI, we have to check if it's still compatible with other federated UIs. Because the other federated UI won't have the same version with shell after we merge your PR.

ui/src/containers/VolumePage.js
```diff
@@ -31,27 +26,16 @@ const VolumePage = (props) => {
...
-  useRefreshEffect(
-    refreshPersistentVolumesAction,
-    stopRefreshPersistentVolumesAction,
-  );
-  useRefreshEffect(
-    refreshCurrentVolumeStatsAction,
-    stopRefreshCurrentVolumeStatsAction,
-  );
+  useFetchCurrentVolumeObject(currentVolumeName);
+  useRefreshNodes();
```
JBWatenbergScality, line 30: Can we drop `dispatch(fetchNodesAction());` below ?
Can we `const nodes = useRefreshNodes();` and drop `const nodes = useTypedSelector((state) => state.app.nodes.list);` below ?
I wonder the same for the other resources.

ui/src/hooks/nodes.js
```diff
@@ -0,0 +1,106 @@
...
+  useEffect(() => {
+    if (data) {
+      dispatch(updateNodesAction({ list: data }));
+
+      // To make sure that the loader is visible for at least 1s
+      setTimeout(function () {
+        // this dispatch should be migrated to react-query
+        // however it should be done when the nodes pages is migrated
+        dispatch({ type: FETCH_NODES_IPS_INTERFACES });
+      }, 1000);
```
JBWatenbergScality: This timeout should be cleaned up to avoid memory leaks.

JBWatenbergScality (changes requested):

ui/src/containers/VolumePage.js
gaudiauj: we have actually an other tickets to handle the removal of all redux function

ui/src/hooks/nodes.js
```diff
@@ -0,0 +1,106 @@
...
+      dispatch({ type: FETCH_NODES_IPS_INTERFACES });
+    }
+    if (data) {
+      dispatch(updateNodesAction({ list: data }));
+
+      // To make sure that the loader is visible for at least 1s
+      setTimeout(dispatchIpInterfaces, 1000);
+    }
+    dispatch(updateNodesAction({ isLoading }));
+    return () => clearTimeout(dispatchIpInterfaces);
```
JBWatenbergScality:
```suggestion
    let timer = false;
    if (data) {
      dispatch(updateNodesAction({ list: data }));

      // To make sure that the loader is visible for at least 1s
      const timer = setTimeout(() => {
          dispatch({ type: FETCH_NODES_IPS_INTERFACES });
      }, 1000);
    }
    dispatch(updateNodesAction({ isLoading }));
    return () => {
        if (timer) {
            clearTimeout(timer);
        }
     }
```

JBWatenbergScality: https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout

JBWatenbergScality (dismissed):

JBWatenbergScality (dismissed):

ChengYanJin (approved):

Issue #3341, opened 2021-04-30
storage operator cannot grab the status of a new formatted raw device


thomasdanan:
**Component**:

'storage-operator'

**What happened**:

For some reason sometimes the storage operator cannot grab the metadata of a newly formatted raw block device (UUID, ...)
There is a failed message in the volume objects for phase 2/2
if you do a `blkid` you will see the UUID of the raw block device.

```
blkid
     /dev/nvme4n1: UUID="6eac9fce-98f9-4cb0-8bbb-fc5b95011603" TYPE="ext4"
```

if you do a `lsblk -f` the line for this raw device is empty.

```
lsblk -f
nvme3n1                        ext4                       99e97bee-f980-40f0-a57e-3b4ed6a48c71
nvme4n1
nvme5n1
nvme6n1                        ext4                       3a480516-edef-4f2c-aff9-7485ec601fbf
```

**What was expected**:
I fixed it by rebooting the server but this should be something else to "rescan properly" devices

**Steps to reproduce**
When you have this issue you can reproduce it easily with a simple `mkfs.ext4` on the raw device to see if you fix it or not.

**Resolution proposal** (optional):

NickSayer-scality:
I've spent some time trying to reproduce this unaligned state on a LAB DL325 with no success (looping over wipes and mkfs). Not able to put the system in a similar state. As Xavier never encountered this behaviour before and on other hardware. Will try to focus tests on the same HW than the one used by Xavier. As this point, we suspect it might be HW related.
Will update after this additional testing.

NicolasT:
Closing since ARTESCA-855 was closed. Can be re-opened if we ever observe this again.

TeddyAndrieux:
In order to solve this error where a `blkid` report the UUID of a drive but not `lsblk` nor list on `/dev/disk/by-uuid` we need to refresh disk information using `udevadm trigger --action=changes`

PR #3630, opened 2021-12-07
salt: Add a udevadm trigger when `persistentPath` does not exists

TeddyAndrieux:
In some cases, a volume may be properly created but the path symlink (
`/dev/disks/by-XXX/`) does not exist because udevadm did not get
triggered on this device for... whatever reason.

To avoid that kind of issue we now trigger udevadm manually in the salt
execution module when the `persistentPath` does not exist and the
volume is actually "prepared".

Fixes: #3341

### DIFF
--- a/salt/_modules/metalk8s_volumes.py
+++ b/salt/_modules/metalk8s_volumes.py
@@ -213,6 +213,25 @@ class Volume:
 
     def device_info(self):
         """Return size and path of the underlying block device"""
+        if not os.path.exists(self.persistent_path) and self.is_prepared:
+            if __salt__["file.is_blkdev"](self.path):
+                filter = self.path
+            else:
+                # It's not a blkdev so We consider it's a sparse loop and since
+                # we do not have any easy way to get the device name of a specific sparse
+                # loop device trigger it on every loop device
+                filter = "--sysname-match='loop*'"
+            # Trigger udev refresh
+            _run_cmd(
+                f"udevadm trigger --action=change --type=devices --subsystem-match=block {filter}"
+            )
+            _run_cmd(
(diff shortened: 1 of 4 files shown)


salt/_modules/metalk8s_volumes.py
```diff
@@ -213,6 +213,25 @@ def clean_up(self):  # pragma: no cover
 
     def device_info(self):
         """Return size and path of the underlying block device"""
+        # NOTE: The ONLY case where `self.path` is not a "blkdev" is when we are using SparseLoop,
+        # which shouldn't be used in production so we do not consider those here (as we do not have
+        # any easy way to get the actual device path without the `/dev/disks/by-uuid` symlink)
```
TeddyAndrieux: Does it seem acceptable to you ?

gdemonet: We could find out the loop device associated with our sparse file using `losetup -j {self.path}`. Also, why do we scope the `udevadm trigger` to a single device? IIRC, if no device is specified, this command will just go through "its backlog of events".

TeddyAndrieux: Agree that with losestup we can find it but a sparse file could be exposed as several devices so you may have several lines for it (even if I'm not sure it work well if we are in this case today)
We could trigger an udevadm on every device yes but I'm not sure about the impact on the device already mounted, right now I'm not fine running `udevadm trigger` on every device here as it will be triggered at any point in time when the platform is running

TeddyAndrieux: As discussed offline let's do a simple global trigger on all loop when we have a problem with a sparse file

gdemonet (approved):

PR #3638, opened 2021-12-10
Improvement/table v2

gaudiauj:
**Component**:

pod and volumes tab in node page

**Context**: 

We want to migrate to the new table from core-ui

**Summary**:

pod and volume table were migrating to the core-ui table

**Acceptance criteria**: 

Same as before.


ui/src/components/NodePageVolumesTable.js
```diff
@@ -96,231 +28,17 @@ const ActionContainer = styled.span`
...
+  const { nodeName, volumeListData } = props;
   const history = useHistory();
   const theme = useTheme();
   const intl = useIntl();
+
+  const NameLinkContaner = styled.div`
+    cursor: pointer;
+    padding-right: ${padding.small};
+    color: ${theme.selectedActive};
+  `;
```
ChengYanJin, line 45: The definition of the style should be outside of the VolumeListTable component.  
 
Not really a scope of this PR, but I suggest to add the link in the core-ui.

ui/src/components/NodePagePodsTab.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components';
 import { Tooltip } from '@scality/core-ui';
-import Table from '@scality/core-ui/dist/components/tablev2/Tablev2.component';
+import { Table } from '@scality/core-ui/dist/next';
```
ChengYanJin, line 6: this commit could be fixup with the two previous ones.

ChengYanJin (commented):
A small comment about the commit message. 
In MetalK8s we tend to put the component name before the colon. 
Example:

> UI: Replace the volume tab table with the single selectable table

ui/src/components/NodePagePodsTab.js
```diff
@@ -22,45 +22,7 @@ const PodTableContainer = styled.div`
   padding: ${padding.large};
   font-family: 'Lato';
   font-size: ${fontSize.base};
```
ChengYanJin: I think the style for the table is already defined within the table component. We should be able to get rid of it.

ui/src/components/NodePageVolumesTable.js
```diff
@@ -1,88 +1,20 @@
...
-  TableHeader,
-} from './style/CommonLayoutStyle';
+import { formatSizeForDisplay } from '../services/utils';
 import { UnknownIcon, TooltipContent } from './TableRow';
+import { Button } from '@scality/core-ui/dist/next';
 
 const VolumeListContainer = styled.div`
   color: ${(props) => props.theme.textPrimary};
   font-family: 'Lato';
   font-size: ${fontSize.base};
```
ChengYanJin: Same here, we should be able to get rid of it.

ChengYanJin:
When I run the UI locally, I see two issues: 

Volume tab: 
The tooltip of the Usage in the first line of the table is not visible :(

image

Pod tab:
We probably miss to define the `cellStyle` for the name colomn :(
image

gaudiauj:
> Volume tab: The tooltip of the Usage in the first line of the table is not visible :(
> 
> image

i will open a PR to fix this directly in core ui : https://github.com/scality/core-ui/blob/development/1.0/src/lib/components/tablev2/SingleSelectableContent.js#L109

ui/cypress/fixtures/kubernetes/pods.json
```diff
@@ -0,0 +1,874 @@
+{
+    "kind": "PodList",
+    "apiVersion": "v1",
+    "metadata": {
+        "selfLink": "/api/v1/pods",
+        "resourceVersion": "27305765"
+    },
+    "items": [
+        {
```
ChengYanJin: I am wondering can we have just an empty array for the item?

ui/src/components/NodePagePodsTab.js
```diff
@@ -156,11 +57,34 @@ const NodePagePodsTab = (props) => {
       {
         Header: 'Name',
         accessor: 'name',
+        cellStyle: {
+          flexGrow: '4',
+          width: '100%',
+          maxWidth: '350px',
+          minWidth: '200px',
```
ChengYanJin: We should use `rem` as the unit instead of `px`. 
16px = 1rem
Please make sure to check the display on the small screen as well ;)
image

ui/src/components/NodePageVolumesTable.js
```diff
@@ -391,11 +113,16 @@ const VolumeListTable = (props) => {
         Header: 'Size',
         accessor: 'storageCapacity',
         cellStyle: {
-          textAlign: 'right',
```
ChengYanJin: Please keep this

ui/src/components/NodePageVolumesTable.js
```diff
@@ -457,13 +184,19 @@ const VolumeListTable = (props) => {
...
+          weights[VOLUME_CONDITION_UNLINK] = 1;
+          weights[VOLUME_CONDITION_EXCLAMATION] = 0;
+          return weights[row1?.values?.status] - weights[row2?.values?.status];
+        },
       },
       {
         Header: 'Latency',
         accessor: 'latency',
         cellStyle: {
-          textAlign: 'right',
```
ChengYanJin: Please keep the right alignment.

ChengYanJin (changes requested):
Besides some comments below, I have two general comments.

1. Do you think there is a way for us to avoid having a double scrollbar here?
And to always keep the table header on top when I scroll down?
image


2. Not related to the code itself, please find the best practice for the commit message here:
usually before the colon, we tend to specify the topic: usually the name of the affected component (ui, build, docs, etc.)
https://metal-k8s.readthedocs.io/en/latest/developer/development/best_practices/commit.html?highlight=commit#subject


Thanks for addressing them! ;)

ui/cypress/fixtures/kubernetes/pods.json
```diff
@@ -6,869 +6,5 @@
         "resourceVersion": "27305765"
     },
     "items": [
-        {
```
ChengYanJin: Can you please fix up this comment with https://github.com/scality/metalk8s/pull/3638/commits/a13015e24a65e5ee899bcca5602ed9949951d1d9?

I prefer not to have unnecessary comments in git history...

ui/src/components/NodePageVolumesTable.js
```diff
@@ -196,7 +197,7 @@ const VolumeListTable = React.memo((props) => {
         Header: 'Latency',
         accessor: 'latency',
         cellStyle: {
-          textAlign: 'center',
+          textAlign: 'right',
```
ChengYanJin: I prefer we fix up this comment with the previous one where we did the change

ui/package.json
```diff
@@ -9,7 +9,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.15.3",
     "@fortawesome/react-fontawesome": "^0.1.14",
     "@kubernetes/client-node": "github:scality/kubernetes-client-javascript.git#browser-0.10.2-63-g579d066",
-    "@scality/core-ui": "github:scality/core-ui.git#17f9913ea6a068753b71d60711740c75c0232af7",
+    "@scality/core-ui": "github:scality/core-ui.git#v0.26.0",
```
ChengYanJin: Let's just put the `package.json` and `package-lock.json` change in a separate comment.

The idea is to have a rather clean history.. thanks!

ChengYanJin (dismissed):
LGTM!

ChengYanJin (approved):

PR #3640, opened 2021-12-14
Handle 401 unauthorized error in the UI

ChengYanJin:
**Component**: UI

**Context**: 
The aim of this PR is to handle correctly the 401 unauthorize error on the UI side.

**Summary**:
Add Error Reducer in order to dispatch authErrorAction.
We throw out AuthError when we get unauthorized status from Salt API and K8s API, and catch it at the top of reducer.
In the PrivateRoute, we display the `<ErrorPage401/>` if isAuthError is true in the redux-store. 

**Acceptance criteria**: 
EN:
image
FR:
image

Core-ui v0.25.0 includes this 401 unauthorized error page.

### DIFF
--- a/shell-ui/package.json
+++ b/shell-ui/package.json
@@ -47,7 +47,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "dependencies": {
-    "@scality/core-ui": "github:scality/core-ui.git#v0.24.2",
+    "@scality/core-ui": "github:scality/core-ui.git#v0.25.0",
     "@scality/module-federation": "github:scality/module-federation.git#1.0.0",
     "oidc-client": "^1.11.3",
     "oidc-react": "^1.1.5",
--- a/ui/src/ducks/app/authError.js
+++ b/ui/src/ducks/app/authError.js
@@ -1,0 +1,29 @@
+//@flow
+export type AuthErrorState = {
+  isAuthError: boolean,
+};
+
+type AuthErrorAction = {
+  type: 'AUTH_ERROR',
+};
+
+const defaultState: AuthErrorState = {
+  isAuthError: false,
+};
+
+export default function reducer(
+  state: AuthErrorState = defaultState,
+  action: AuthErrorAction,
+) {
+  switch (action.type) {
+    case 'AUTH_ERROR': {
+      return { ...state, isAuthError: true };
+    }
+    default:
+      return { ...state };
+  }
+}
+
+export function authErrorAction() {
+  return { type: 'AUTH_ERROR' };
+}
--- a/ui/src/services/salt/api.js
+++ b/ui/src/services/salt/api.js
@@ -1,6 +1,7 @@
 //@flow
 import { User } from 'oidc-client';
 import ApiClient from '../ApiClient';
+import { handleUnAuthorizedError } from '../errorhandler';
 
 let saltApiClient = null;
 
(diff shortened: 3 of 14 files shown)


Cuervino:
@scality/doc Could you please review the wording? 
I'm not sure about the "supplied".

shell-ui/package.json
```diff
@@ -47,7 +47,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "dependencies": {
-    "@scality/core-ui": "github:scality/core-ui.git#v0.22.2.1",
+    "@scality/core-ui": "github:scality/core-ui.git#695c63c",
```
gaudiauj: do we really want to keep the commit hash rather than deploying a new version ?

ui/src/ducks/app/authError.js
```diff
@@ -0,0 +1,29 @@
...
+  action: AuthErrorAction,
+) {
+  switch (action.type) {
+    case 'AUTH_ERROR': {
+      return { ...state, isAuthError: true };
+    }
+    default:
+      return { ...state };
+  }
+}
```
gaudiauj, line 25: i'm not familiar with the way it work but what i see from the code, there is no way to remove the isAuthError. And we check in `privateRoute.js` if isAuthError is true. Are we sure that if the user relogin the isAuthError is reset ?

gaudiauj (dismissed):
just couple of questions, but besides that it looks good. 

approved

shell-ui/package.json
ChengYanJin: Yes. You are right. We should tag a new version for the core-ui. 
I am just waiting a bit to see if I can get a review from @scality/doc about the wording in the 401 unauthorized page.

ui/src/services/salt/api.js
```diff
@@ -113,4 +128,9 @@ export async function getNodesIPsInterfaces(
       'ip_interfaces',
     ],
   });
+  if (result.error) {
+    handleUnAuthorizedError({ error: result.error });
```
JBWatenbergScality: I think it may need to be returned

JBWatenbergScality (commented):
Other than this comment and Jean's comment LGTM

ui/src/services/salt/api.js
```diff
@@ -50,17 +57,27 @@ export async function deployNode(node: string, version: string) {
...
     throw new Error('Salt api client should be defined.');
   }
-  return saltApiClient.post('/', {
+  const result = saltApiClient.post('/', {
     client: 'runner',
     fun: 'jobs.print_job',
     arg: [jid],
   });
+  if (result.error) {
+    handleUnAuthorizedError({ error: result.error });
```
JBWatenbergScality: here too

ui/src/ducks/app/authError.js
ChengYanJin: Hello @gaudiauj ,
It's a very good concern! but thanks to the authentication workflow, the isAuthError should be reset if we re-login again.

Once the user triggers logout action, the page will be redirected to the IDP, either dex or Keycloak. 

Given the page is fully reloaded, the redux store will be re-initiated as well.

cathydossantospinto:
@ChengYanJin Here's my suggestion for the French part.
 
```
! Accès refusé
Les identifiants fournis ne vous permettent pas de consulter cette page.
Vous pouvez contacter le support pour signaler le problème.
```
@lucieleonard Feel free to suggest something better if you can 😛 

@John-VanHook will review the English part 😉

John-VanHook:
The content of the dialog box is not wrong. You could keep it as is. But I suggest this is better: 

You don’t have permission to view this page using the credentials **you have** supplied. 

You **can** contact `support` to report this issue.

ChengYanJin:
ping

JBWatenbergScality (approved):

PR #3641, opened 2021-12-15
salt: Skip "Pending" pods when draining a node

TeddyAndrieux:
In most cases, it does not make sense to try to evict a Pod which is
Pending as this one is not yet running and it also can cause trouble
during MetalK8s expansions as some Pod may be Pending on a node that is
not yet deployed (so the Pending pod cannot get evicted).

Let's skip "Pending" pods by default when we drain a node in
`deploy_node` orchestrate (used during expansions and upgrade,
downgrade)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 ## Release 2.10.7 (in development)
+## Enhancements
+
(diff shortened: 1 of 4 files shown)


gdemonet (approved):
:shipit:

PR #3642, opened 2021-12-15
shell: Wrap federated routes with a router

JBWatenbergScality:
**Component**:

shell

**Context**: 

Before this modification if a federated app embed more than one route under the same path the navigation from shell-ui was broken as the federated app history didn't receive updates from navigation performed in the shell


ChengYanJin (approved):
LGTM!

PR #3643, opened 2021-12-16
ui: fix the alert not displaying correctly when the name is too long

gaudiauj:
**Component**:

alert page

**Context**: 
Very long name are not correctly displayed in the alert table

Before
image-2021-10-28-09-05-30-680

After
Capture d’écran du 2021-12-16 11-56-39

**Summary**:
add ellipsis

### DIFF
--- a/ui/src/containers/AlertPage.js
+++ b/ui/src/containers/AlertPage.js
@@ -16,11 +16,7 @@ import { useAlerts } from './AlertProvider';
 import StatusIcon from '../components/StatusIcon';
 import CircleStatus from '../components/CircleStatus';
 import { TextBadge } from '../components/style/CommonLayoutStyle';
-import {
-  STATUS_WARNING,
-  STATUS_CRITICAL,
-  STATUS_HEALTH,
-} from '../constants';
+import { STATUS_WARNING, STATUS_CRITICAL, STATUS_HEALTH } from '../constants';
 import {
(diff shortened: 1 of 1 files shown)


ui/src/containers/AlertPage.js
```diff
@@ -453,7 +443,11 @@ export default function AlertPage() {
       {
         Header: 'Name',
         accessor: 'labels.alertname',
-        cellStyle: { width: '300px' },
+        cellStyle: {
+          width: '300px',
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+        },
```
JBWatenbergScality, line 450: I think we should still try to use the new table because this fix is not fully solving the problem. The user won't be able to see the full alert name, I think the right implementation should be, if it is ellipsis then hovering it should display a tooltip with the complete alertName. IIRC this is the default behaviour we implemented in the new table component.

gaudiauj: For what i can see in core-UI storybook it's not the default behavior for the table, we will have to code it anyway. But i can definitly migrate to the new table. I just think it's 2 differents topics.

gaudiauj:
discussed with @JBWatenbergScality . We will make sure that the table from core-ui handle correctly text-overflow and then migrate this table to core-ui table.

Closing for now.

PR #3645, opened 2021-12-17
Refactor UI Node Metric Charts by using custom chart hooks

ChengYanJin:
**Component**: UI

**Context**: 
All the charts in the UI use the custom chart hook to retrieve the series, but the charts on Node Metrics Tab, which was a technical debt. 

**Summary**:
The aim of this PR is to refactor the component `<MetricChart/>` and `<MetricSymmetricalChart/>` to use the `useChartSeries`and `useSymetricalChartSeries` to retrieve all the props for LinetemporalChart.

**Acceptance criteria**: 
The Node Metrics Charts should behave the same as before.

### DIFF
--- a/ui/src/containers/NodePageMetricsTab.js
+++ b/ui/src/containers/NodePageMetricsTab.js
@@ -10,6 +10,7 @@ import { useIntl } from 'react-intl';
 import { SyncedCursorCharts } from '@scality/core-ui/dist/next';
 import { UNIT_RANGE_BS } from '@scality/core-ui/dist/components/linetemporalchart/LineTemporalChart.component';
 import { updateNodeStatsFetchArgumentAction } from '../ducks/app/monitoring';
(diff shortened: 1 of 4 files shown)


ui/src/containers/NodePageMetricsTab.js
```diff
@@ -104,7 +104,7 @@ const NodePageMetricsTab = ({
   instanceIP: string,
   controlPlaneInterface: string,
   workloadPlaneInterface: string,
-  nodesIPsInfo: [],
+  nodesIPsInfo: Object,
```
JBWatenbergScality: I think we already have a type for that, I would say we should use it.

ChengYanJin: Yes. that's totally right!

JBWatenbergScality (approved):

PR #3651, opened 2021-12-30
bugfix: fix display of volume usage on newly created volumes

gaudiauj:
**Component**:

Volume overview

**Context**: 

fix the following errors : 
image (9)
image (10)

### DIFF
--- a/ui/src/components/VolumeOverviewTab.js
+++ b/ui/src/components/VolumeOverviewTab.js
@@ -19,7 +19,7 @@ import {
   STATUS_WARNING,
   LVM_LOGICAL_VOLUME,
 } from '../constants';
-import { Button, Modal, ProgressBar, Loader } from '@scality/core-ui';
+import { Button, Modal, ProgressBar } from '@scality/core-ui';
 import { useIntl } from 'react-intl';
 import { VolumeTab } from './style/CommonLayoutStyle';
 import { formatSizeForDisplay } from '../services/utils';
@@ -112,9 +112,6 @@ const ModalBody = styled.div`
 
 const CancelButton = styled(Button)`
   margin-right: ${padding.small};
-`;
-const LoaderContainer = styled(Loader)`
-  padding-left: ${padding.small};
 `;
 
 const LabelName = styled.span`
@@ -246,7 +243,7 @@ const VolumeDetailCard = (props) => {
               {intl.formatMessage({ id: 'size' })}
             </InformationLabel>
             <InformationValue data-cy="volume_size_value">
-              {storageCapacity || intl.formatMessage({id: 'unknown'})}
+              {storageCapacity || intl.formatMessage({ id: 'unknown' })}
             </InformationValue>
           </InformationSpan>
           <InformationSpan>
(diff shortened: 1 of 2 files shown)


ui/src/services/utils.js
```diff
@@ -372,9 +372,14 @@ export const compareHealth = (status1, status2) => {
...
-    return value.replace(/^(\d+)(\D+)$/, '$1 $2') + 'B';
-  else return value;
+  if (value && value.match(/^(\d+)(\D+)$/)) {
+    const newValue = value.replace(/^(\d+)(\D+)$/, '$1 $2');
+    if (newValue.match(/.+(B$)/i)) {
+      return newValue;
+    } else {
+      return newValue + 'B';
+    }
+  } else return value;
```
ChengYanJin: It could fix the issue, but if it was me, I would get further to see why I have already had the unit B in the raw value. 

And I found the data `storageCapacity` is from `NodeVolumeUtile.js`

```
storageCapacity: volumeCapacityCurrent?.value[1]
    ? bytesToSize(volumeCapacityCurrent?.value[1])
    : formatSizeForDisplay(volumePV?.spec?.capacity?.storage),
```

So I would remove formatSizeForDisplay. and I think with this fix we don't need to have the regex checking.
```
storageCapacity: volumeCapacityCurrent?.value[1]
    ? bytesToSize(volumeCapacityCurrent?.value[1])
    : volumePV?.spec?.capacity?.storage,
```

ui/src/components/VolumeOverviewTab.js
```diff
@@ -360,31 +360,32 @@ const VolumeDetailCard = (props) => {
...
-            </VolumeUsage>
-          )}
+          {volumeUsagePercentage !== undefined &&
+            condition === VOLUME_CONDITION_LINK && (
+              <VolumeUsage>
+                <VolumeSectionTitle>
+                  {intl.formatMessage({ id: 'usage' })}
+                </VolumeSectionTitle>
+                {volumeUsagePercentage !==
+                intl.formatMessage({ id: 'unknown' }) ? (
```
ChengYanJin: Not really related to this specific bug fix, but I think we can remove this checking.

ChengYanJin (changes requested):
Please address the conflicts within your PR as well, thanks!
You can follow the bert-e instructions just before this comment.

ui/src/components/VolumeOverviewTab.js
gaudiauj: I trust you on this one, i'm not 100% sure that `volumeUsagePercentage`  has never this value

ui/src/services/utils.js
gaudiauj: :thinking:  

Since the storage capacity is used in other components like the overviewTab. I would rather remove formatSizeForDisplay from here : https://github.com/scality/metalk8s/blob/development/2.11/ui/src/components/VolumeListTable.js#L420 . WDYT ? 

I agree that this fix doesn't solve the root issue, but since the fix was needed this week. I was more comfortable doing that because  i was sure it doesn't have side effect on other components.

ChengYanJin: Right! OK. Let's remove the formatSizeForDisplay in the VolumeListTable then.

ui/src/components/VolumeOverviewTab.js
ChengYanJin: I think it should be OK.. We got the `volumeUsagePercentage` from the `usage` which we compute here:

NodeVolumesUtil.js
```
        usage:
          volumeUsedCurrent && volumeCapacityCurrent
            ? (
                (volumeUsedCurrent?.value[1] / volumeCapacityCurrent.value[1]) *
                100
              ).toFixed(2)
            : undefined,
 ```

ChengYanJin (approved):

PR #3657, opened 2022-01-04
Fix upgrade from MetalK8s 2.10.x to 2.11

TeddyAndrieux:
Since the LabelSelector for the kube-state-metrics deployment changed in
8be15bd810d434bc61f762023b4afe5625695c64, we cannot replace the
kube-state-metrics deployment object if the old one exists, that's why
we ensure old kube-state-metrics deployment get removed before deploying
the new one

### DIFF
--- a/salt/metalk8s/addons/prometheus-operator/deployed/init.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/init.sls
@@ -8,3 +8,31 @@ include:
   - .chart
   - .prometheus-rules
   - .thanos-chart
+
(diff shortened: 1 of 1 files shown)


alexandre-allard (approved):

gdemonet (approved):
I guess this will cause a small hole in metrics when no k-s-m pod is available anymore? Anyway, looks acceptable to me.

TeddyAndrieux:
> I guess this will cause a small hole in metrics when no k-s-m pod is available anymore? Anyway, looks acceptable to me.

Yes likely but it's right before applying the new chart so yes it should be short

PR #3660, opened 2022-01-10
ui: Exclude the NSFS filesystem from the system partitions

ChengYanJin:
**Component**: UI

**Context**: 
Exclude the NSFS filesystem from the system partitions

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


ChengYanJin:
Will add the NSFS in the ignored-fs-types in node-exporter instead of adding the filter on the UI side.

PR #3664, opened 2022-01-14
packages: Make our sosreport package compatible with new sos version

### LINKED ISSUE #3387: metalk8s-sosreport package is not compatible with latest sos 4.0.0+ versions
**Component**: packages

**What happened**:
In the CI, we are unable to successfully run the sosreport step for RHEL 8

**What was expected**:
Report is generated successfully.

**Steps to reproduce**:
Launch a build and wait for the single-node-install rhel stage to fail.

First, we can fix the sos version needed to an older version, but in a near future we may want to be compatible with sos 4.0.0+ versions.

TeddyAndrieux:
Since sos layout changed in 4.0 and CentOS 7 does not have this sos
version available by default, so we need some logic in the packages
in order to be compatible with sos > 4.0 and sos < 4.0.

We enforce sos < 4.0 for CentOS/RHEL 7 and sos >= 4.0 for RHEL8

Fixes: #3387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,10 @@
   and enable Grafana folder generation from dashboard file structure (PR
   [#3620](https://github.com/scality/metalk8s/pull/3620))
 
+- [#3387](https://github.com/scality/metalk8s/issues/3387) - Make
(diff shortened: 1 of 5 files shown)


alexandre-allard (approved):
LGTM

PR #3668, opened 2022-01-18
charts,salt: Set UID for grafana datasources

TeddyAndrieux:
Since in newer grafana version we need to use the datasource UID to
reference it, we explicitly set the datasource UID from ConfigMap so
that it's easier to reference

### DIFF
--- a/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
@@ -25089,13 +25089,14 @@ data:
   datasource.yaml: |-
     apiVersion: 1
     datasources:
-    - name: Prometheus
-      type: prometheus
-      url: http://thanos-query-http:10902/
-      access: proxy
+    - access: proxy
       isDefault: true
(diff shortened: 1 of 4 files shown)


salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
```diff
@@ -25089,13 +25089,12 @@ data:
...
+    - access: proxy
+      isDefault: true
+      name: Prometheus
       type: prometheus
+      uid: metalk8s-prometheus
       url: http://thanos-query-http:10902/
-      access: proxy
-      isDefault: true
-      jsonData:
-        timeInterval: 30s
```
gdemonet: This was lost :confused:

TeddyAndrieux: Hmm I thought it was the default from grafana but it's not apparently, I will add it back :+1:

gdemonet (approved):

francoisferrand (approved):

PR #3669, opened 2022-01-19
improvement: use tableV2 in the alert page

gaudiauj:
**Component**:

AlertPage 

**Context**: 

fix bad formating of alerts by using new table

### DIFF
--- a/ui/src/containers/AlertPage.js
+++ b/ui/src/containers/AlertPage.js
@@ -1,39 +1,25 @@
 //@flow
 import React, { useMemo } from 'react';
 import styled from 'styled-components';
-import { useHistory } from 'react-router';
-import { useLocation } from 'react-router-dom';
-import {
-  useTable,
-  useSortBy,
-  useAsyncDebounce,
-  useFilters,
-  useGlobalFilter,
-} from 'react-table';
-import { EmptyTable, SearchInput, StatusWrapper } from '@scality/core-ui';
-import { padding, fontSize } from '@scality/core-ui/dist/style/theme';
+import { StatusWrapper, ConstrainedText } from '@scality/core-ui';
+import { Table } from '@scality/core-ui/dist/next';
+import { padding, fontSize, spacing } from '@scality/core-ui/dist/style/theme';
 import { useAlerts } from './AlertProvider';
 import StatusIcon from '../components/StatusIcon';
-import CircleStatus from '../components/CircleStatus';
 import { TextBadge } from '../components/style/CommonLayoutStyle';
-import {
-  STATUS_WARNING,
-  STATUS_CRITICAL,
-  STATUS_HEALTH,
-} from '../constants';
-import {
-  compareHealth,
-  useURLQuery,
-  useTableSortURLSync,
-  formatDateToMid1,
-} from '../services/utils';
+import { STATUS_WARNING, STATUS_CRITICAL, STATUS_HEALTH } from '../constants';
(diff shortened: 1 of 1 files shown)


ui/src/containers/AlertPage.js
```diff
@@ -89,18 +75,27 @@ const AlertStatusIcon = styled.div`
...
+  return a.every((alertData) =>
+    b.find(
+      (alert) =>
+        alert.id === alertData.id && alert.severity === alertData.severity,
+    ),
+  );
+};
+
+const getAlertStatus = (critical, warning) =>
+  critical > 0 ? STATUS_CRITICAL : warning > 0 ? STATUS_WARNING : STATUS_HEALTH;
```
ChengYanJin: We have `id` for each alert, maybe we can benefit from it?

ui/src/containers/AlertPage.js
```diff
@@ -179,260 +158,81 @@ export const TableHeader = styled.th`
...
+    const DEFAULT_SORTING_KEY = 'severity';
+
+    return (
+      <Table
+        columns={columns}
+        data={data}
+        defaultSortingKey={DEFAULT_SORTING_KEY}
+        sortTypes={sortTypes}
+      >
+        <div style={{ margin: '16px 0' }}>
```
ChengYanJin: Please use {spacing} from core-ui.

ui/src/containers/AlertPage.js
```diff
@@ -179,260 +158,81 @@ export const TableHeader = styled.th`
...
+);
 
 export default function AlertPage() {
   const alerts = useAlerts({});
+
   const leafAlerts = useMemo(
     () => alerts?.alerts.filter((alert) => !alert.labels.children) || [],
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [JSON.stringify(alerts?.alerts)],
+    [alerts?.alerts],
```
ChengYanJin: I think we should keep the previous code.

React uses “referential equality” for object and array within the dependencies array.
so even we have the same content of alerts, it will still re-run the effect. That's why we stringify. 

A pretty good article to read about.
https://www.benmvp.com/blog/object-array-dependencies-react-useEffect-hook/

ChengYanJin (changes requested):

ui/src/containers/AlertPage.js
gaudiauj: it's actually the number of alerts, i changed so it's more easy to understand

ui/src/containers/AlertPage.js
```diff
@@ -179,259 +158,80 @@ export const TableHeader = styled.th`
...
+    return isEqual(a.columns, b.columns) && isEqualAlert(a.data, b.data);
+  },
+);
 
 export default function AlertPage() {
   const alerts = useAlerts({});
+
   const leafAlerts = useMemo(
     () => alerts?.alerts.filter((alert) => !alert.labels.children) || [],
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ChengYanJin, line 434: You will need to keep this line, otherwise Eslint will complain about about the missing dependency :(

ChengYanJin (dismissed):

ui/src/containers/AlertPage.js
gaudiauj: oh my bad will fix this :)

ChengYanJin (approved):

PR #3675, opened 2022-01-21
salt: Do not fail if CP Ingress IP not set

TeddyAndrieux:
Do not fail if the Control Plane Ingress section exists in the
Bootstrap configuration file, but Ingress IP is not set.

### DIFF
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -229,7 +229,7 @@ def routes():
 
 
 def get_control_plane_ingress_ip():
-    if "ingress" in __pillar__["networks"]["control_plane"]:
+    if __pillar__["networks"]["control_plane"].get("ingress", {}).get("ip"):
         return __pillar__["networks"]["control_plane"]["ingress"]["ip"]
 
     # Use Bootstrap Control Plane IP as Ingress Control plane IP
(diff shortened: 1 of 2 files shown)


JBWatenbergScality (approved):

salt/_modules/metalk8s_network.py
```diff
@@ -229,7 +229,7 @@ def routes():
 
 
 def get_control_plane_ingress_ip():
-    if "ingress" in __pillar__["networks"]["control_plane"]:
+    if __pillar__["networks"]["control_plane"].get("ingress", {}).get("ip"):
         return __pillar__["networks"]["control_plane"]["ingress"]["ip"]
```
gdemonet, line 233: Could be simplified with a try/except:

```suggestion
    try:
        return __pillar__["networks"]["control_plane"]["ingress"]["ip"]
    except KeyError:
        pass
```

gdemonet (approved):
Just a nitpick, feel free to discard it :)

salt/_modules/metalk8s_network.py
TeddyAndrieux: Hmm not exactly the same for example if you set the IP to `None` but right we may not want to support this :smiley:

PR #3681, opened 2022-01-25
Add datasource for each Loki instance

alexandre-allard:
**Component**: salt,logs

**Context**: 

**Summary**:

**Acceptance criteria**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 ## Release 2.11.1 (in development)
 ### Enhancements
 
+- Add each Loki instance as datasources in Grafana dashboards
+  (PR[#3681](https://github.com/scality/metalk8s/pull/3681))
+
 - Bump Grafana image version to 8.3.4-ubuntu
   (PR[#3684](https://github.com/scality/metalk8s/pull/3684))
+
 ## Release 2.11.0
 ### Additions
 
--- a/salt/metalk8s/addons/logging/loki/deployed/datasource.sls
+++ b/salt/metalk8s/addons/logging/loki/deployed/datasource.sls
@@ -1,7 +1,16 @@
+{%- set loki_defaults = salt.slsutil.renderer(
+        'salt://metalk8s/addons/logging/loki/config/loki.yaml', saltenv=saltenv
+    )
+%}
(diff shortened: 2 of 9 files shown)


salt/metalk8s/addons/logging/loki/deployed/datasource.sls
```diff
@@ -24,3 +33,11 @@ Deploy ConfigMap for Loki datasource:
...
               url: http://loki.metalk8s-logging.svc:3100/
               version: 1
+{%- for index in range(loki.spec.deployment.replicas) %}
+            - name: Loki-{{ index }}
+              uid: metalk8s-loki-{{ index }}
+              type: loki
+              access: proxy
+              url: http://loki-{{ index }}.metalk8s-logging.svc:3100/
+              version: 1
+{%- endfor %}
```
TeddyAndrieux, line 43: Wondering if we can have one "default" datasource per type or if it's global, maybe we want to set the "Loki" datasource as default ?

TeddyAndrieux (commented):
LGTM

Can you just add a changelog, and maybe we want to add a simple test just to be sure the "loki-x" service properly redirect to Loki
(I think about maybe one day a label in the Loki statefulset that change so the extra service we deploy does not get updated and we do not catch it)

salt/metalk8s/addons/logging/loki/deployed/datasource.sls
gdemonet: Not possible, we can only set one default per org

TeddyAndrieux: :cry:

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@
 
 ### Enhancements
 
+- Add each Loki instance as datasources in Grafana dashboards
+  (PR[#3681](https://github.com/scality/metalk8s/pull/3681))
+
```
TeddyAndrieux: You will need to rebase since this should be in the 2.11.1 section (yet another issue with changelog as a simple file)

TeddyAndrieux (dismissed):
Ok for me !

TeddyAndrieux (approved):

PR #3687, opened 2022-02-03
alert-tree: Add missing "group_by" on Node alerts

gdemonet:
We forgot these when migrating from the previous definitions, which
leads to empty labels and incorrect grouping (only one alert can fire,
even if multiple nodes / partitions are affected).


ChengYanJin (dismissed):
LGTM!

ChengYanJin (approved):

PR #3688, opened 2022-02-04
salt,dex: Allow to disable Dex deployment and configure another IDP

alexandre-allard:
**Component**: salt, dex

**Context**:
We want to be able to not deploy Dex and rather rely on an external IDP.

**Summary**:

To disable Dex deployment at installation,
you must add the following to the boostrap config:

```
addons:
  dex:
    enabled: false
```

**Acceptance criteria**: 
We can use another IDP than Dex (e.g. Keycloak) and everything work fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # CHANGELOG
 ## Release 123.0.0 (in development)
 
+### Additions
+
+- Add ability to deploy MetalK8s without Dex and possibility to configure
+  your own IDP for kube-apiserver, Grafana and MetalK8s UI
+  (PR[#3688](https://github.com/scality/metalk8s/pull/3688))
+
 ### Enhancements
 
 - Bump Kubernetes version to
--- a/buildchain/buildchain/salt_tree.py
+++ b/buildchain/buildchain/salt_tree.py
@@ -346,7 +346,7 @@ SALT_FILES: Tuple[Union[Path, targets.AtomicTarget], ...] = (
     Path("salt/metalk8s/addons/prometheus-operator/post-downgrade.sls"),
     Path("salt/metalk8s/addons/prometheus-operator/post-upgrade.sls"),
     Path("salt/metalk8s/addons/prometheus-operator/config/alertmanager.yaml"),
-    Path("salt/metalk8s/addons/prometheus-operator/config/grafana.yaml"),
+    Path("salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2"),
     Path("salt/metalk8s/addons/prometheus-operator/config/prometheus.yaml"),
     Path(
         "salt/metalk8s/addons/prometheus-operator/deployed/",
@@ -359,6 +359,7 @@ SALT_FILES: Tuple[Union[Path, targets.AtomicTarget], ...] = (
         "salt/metalk8s/addons/prometheus-operator/deployed/files/",
         "node-exporter-full.json",
     ),
+    Path("salt/metalk8s/addons/prometheus-operator/deployed/grafana-ini-configmap.sls"),
     Path("salt/metalk8s/addons/prometheus-operator/deployed/init.sls"),
     Path("salt/metalk8s/addons/prometheus-operator/deployed/namespace.sls"),
     Path("salt/metalk8s/addons/prometheus-operator/deployed/prometheus-rules.sls"),
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -73,6 +73,9 @@ Configuration
         minion: <hostname-of-the-bootstrap-node>
       archives:
         - <path-to-metalk8s-iso>
+      addons:
+        dex:
+          enabled: True
       kubernetes:
         apiServer:
           featureGates:
@@ -187,6 +190,14 @@ For example;
 The ``archives`` field is a list of absolute paths to MetalK8s ISO files. When
 the bootstrap script is executed, those ISOs are automatically mounted and the
 system is configured to re-mount them automatically after a reboot.
+
+The ``addons`` field can be omitted if you do not have any specific addons
+to configure.
+
+  If you need to disable deployment of ``dex`` as default OIDC used by
+  MetalK8s you can disable it by setting ``addons.dex.enabled`` to ``false``.
+  If ``dex`` is disabled you will not be able to use the MetalK8s UI and
+  Grafana.
 
 The ``kubernetes`` field can be omitted if you do not have any specific
 Kubernetes `Feature Gates`_ to enable or disable and if you are ok with
--- a/docs/installation/bootstrap.rst
+++ b/docs/installation/bootstrap.rst
@@ -78,6 +78,7 @@ Configuration
           enabled: True
       kubernetes:
         apiServer:
+          oidc: {}
           featureGates:
             <feature_gate_name>: True
         controllerManager:
@@ -207,6 +208,20 @@ defaults kubernetes configuration.
   configure the corresponding entries in the
   ``kubernetes.apiServer.featureGates`` mapping.
 
+  If ``dex`` is enabled, it will be used as ``oidc`` for ``kube-apiserver``
+  but you can use a `specific OpenID for kube-apiserver`_, to do so:
+
+    .. code-block:: yaml
+
+      kubernetes:
+        apiServer:
+          oidc:
+            issuerURL: <OIDC issuer URL>
+            clientID: <Client ID>
+            CAFile: <Certificate Authority certificate file>
+            usernameClaim: <Username Claim>
+            groupsClaim: <Groups Claim>
+
   If you want to override the default ``coreDNS`` podAntiAffinity or number of
   replicas, by default MetalK8s deploy 2 replicas and use soft podAntiAffinity
   on hostname so that if it's possible ``coreDNS`` pods will be spread on
@@ -230,6 +245,7 @@ defaults kubernetes configuration.
   disabled (default to ``500``)
 
 .. _Feature Gates: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+.. _specific OpenID for kube-apiserver: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens
 
 .. _Bootstrap SSH Provisioning:
 
--- a/salt/metalk8s/addons/dex/deployed/init.sls
+++ b/salt/metalk8s/addons/dex/deployed/init.sls
@@ -9,6 +9,8 @@
 # * chart                  -> charts used to deploy Dex
 # * clusterrolebinding     -> binds dex static user to cluster admin
 
+{%- if pillar.addons.dex.enabled %}
+
 include:
 - .namespace
 - .tls-secret
@@ -18,3 +20,10 @@ include:
 - .secret
 - .chart
 - .clusterrolebinding
+
+{%- else %}
+
+Dex is disabled, nothing to deploy:
+  test.nop
+
+{%- endif %}
--- a/salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
+++ b/salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
@@ -1,0 +1,44 @@
+#!jinja|yaml
+
+{%- if pillar.addons.dex.enabled %}
+  {%- set dex_defaults = salt.slsutil.renderer('salt://metalk8s/addons/dex/config/dex.yaml.j2', saltenv=saltenv) %}
+  {%- set dex = salt.metalk8s_service_configuration.get_service_conf('metalk8s-auth', 'metalk8s-dex-config', dex_defaults) %}
+{%- endif %}
+{%- set control_plane_ingress_endpoint = salt.metalk8s_network.get_control_plane_ingress_endpoint() %}
+
+# Configuration of the Grafana service
+apiVersion: addons.metalk8s.scality.com
+kind: GrafanaConfig
+spec:
+  # Configure the Grafana Deployment
+  deployment:
+    replicas: 1
+  config:
+    grafana.ini:
+      analytics:
+        check_for_updates: false
+        reporting_enabled: false
+      paths:
+        data: /var/lib/grafana/
+        logs: /var/log/grafana
+        plugins: /var/lib/grafana/plugins
+        provisioning: /etc/grafana/provisioning
+      log:
+        mode: console
+      server:
+        root_url: "{{ control_plane_ingress_endpoint }}/grafana"
+      auth:
+        disable_login_form: true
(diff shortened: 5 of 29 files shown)


salt/tests/unit/modules/files/test_metalk8s_service_configuration.yaml
```diff
@@ -64,6 +64,24 @@ get_service_config:
           containers:
             - name: my_new_container
 
+  #. Success: retrieved ConfigMap manifest is empty
```
gdemonet, line 67: That's quite the functional change, any idea about the impact? Are we properly handling this case in formulas?

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
```diff
@@ -0,0 +1,45 @@
...
+        reporting_enabled: falseœ
+      paths:
+        data: /var/lib/grafana/
+        logs: /var/log/grafana
+        plugins: /var/lib/grafana/plugins
+        provisioning: /etc/grafana/provisioning
+      log:
+        mode: console
+      grafana_net:
+        url: https://grafana.net
```
gdemonet: Let's drop this, we really don't need it IMO: https://grafana.com/docs/grafana/latest/administration/configuration/#grafana_net (they don't even explain what it's for)

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
```diff
@@ -0,0 +1,45 @@
...
+kind: GrafanaConfig
+spec:
+  # Configure the Grafana Deployment
+  deployment:
+    replicas: 1
+  config:
+    grafana.ini:
+      analytics:
+        check_for_updates: true
+        reporting_enabled: falseœ
```
gdemonet: That last character `œ` shouldn't be here

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
```diff
@@ -0,0 +1,45 @@
...
+apiVersion: addons.metalk8s.scality.com
+kind: GrafanaConfig
+spec:
+  # Configure the Grafana Deployment
+  deployment:
+    replicas: 1
+  config:
+    grafana.ini:
+      analytics:
+        check_for_updates: true
```
gdemonet: Why is it now `true`?

salt/metalk8s/addons/dex/deployed/init.sls
```diff
@@ -9,6 +9,8 @@
 # * chart                  -> charts used to deploy Dex
 # * clusterrolebinding     -> binds dex static user to cluster admin
 
+{%- if pillar.addons.dex.enabled %}
```
TeddyAndrieux, line 12: Nitpick: To me, it could be done in `metalk8s.deployed` state directly so that you can manually deploy Dex running the state but it's not deployed by bootstrap, upgrade, ...

salt/metalk8s/addons/ui/deployed/init.sls
```diff
@@ -1,6 +1,15 @@
+{%- if pillar.addons.dex.enabled %}
```
TeddyAndrieux: ditto, it could be in `metalk8s.deployed` state

salt/metalk8s/addons/prometheus-operator/deployed/grafana-ini-configmap.sls
```diff
@@ -0,0 +1,30 @@
...
+          name: prometheus-operator-grafana
+          namespace: metalk8s-monitoring
+        data:
+          grafana.ini: |-
+{%- for key, value in grafana.spec.config["grafana.ini"].items() %}
+            [{{ key }}]
+  {%- for element, element_value in value.items() %}
+            {{ element }} = {{ element_value }}
+  {%- endfor %}
+{%- endfor %}
```
TeddyAndrieux, line 30: Uh :confused: this is really needed ? A `| yaml(False) | indent(12)` does not work ?

salt/tests/unit/formulas/data/base_pillar.yaml
```diff
@@ -187,3 +187,4 @@ kubernetes:
 addons:
   dex:
     enabled: True
+oidc: {}
```
TeddyAndrieux: Maybe worth adding a unit test with an OIDC in pillar content (just to ensure that we do not break the salt state rendering)

buildchain/buildchain/salt_tree.py
```diff
@@ -351,6 +351,7 @@ def task(self) -> types.TaskDict:
         "alertmanager-configuration-secret.sls",
     ),
     Path("salt/metalk8s/addons/prometheus-operator/deployed/chart.sls"),
+    Path("salt/metalk8s/addons/prometheus-operator/deployed/chart-without-grafana.sls"),
```
TeddyAndrieux: TBH: I'm not sure we need to do this... it means yet another state to maintain (and test, ...),

Since the grafana config can be managed by the user himself, we shouldn't care about it and always deploy grafana. (what happened if we deploy grafana without `auth` section ?)

And maybe it should be the same for the metalk8s-ui depending on how the pod behaves when you do not have OIDC (crash loop or just the web UI that show an error ?)

TeddyAndrieux (commented):
Overall LGTM but I think we will need to add some E2E tests maybe just a really simple one for the moment
Like just a simple install with no OIDC for the moment, we put it in post-merge so that we can enrich it in the future with a real scenario (deploy with no OIDC, then set up another OIDC then re-configure the product to use this new OIDC)

buildchain/buildchain/salt_tree.py
alexandre-allard: Everything works if we still deploy Grafana and MetalK8s UI, I only did this because we said we don't want to deploy them without an OIDC in the meeting we did.

salt/metalk8s/addons/prometheus-operator/deployed/grafana-ini-configmap.sls
alexandre-allard: It's not YAML, but INI configuration file.

alexandre-allard: But we could probably add a Jinja filter to format INI configurations.

salt/metalk8s/addons/dex/deployed/init.sls
alexandre-allard: Well, it's what I did in the beginning, but the goal was to be sure we can't force or mistakenly deploy it.
For example, in the MetalK8s cluster and services configurations documentation, we explicitly say to run `metalk8s.addons.dex.deployed` after a change to its related CSC ConfigMap.
And we do it in the ARTESCA installer for both Keycloak and MetalK8s configuration//deployment.

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Ok, yes I see, but now to me, we shouldn't deploy Dex as it will not be used, but deploy Grafana and MetalK8s UI even if it's still not usable for the moment... it's ok and easier to maintain

Maybe it should be re-discussed a bit with @scality/metalk8s 

I just do not see a real value of not deploying them. Especially for Grafana where we will need to "maintain" a new salt state (and ensure this one works properly)

salt/metalk8s/addons/prometheus-operator/deployed/grafana-ini-configmap.sls
TeddyAndrieux: Oh you're right :cry: 
Ok no I think it's good like that then

salt/metalk8s/addons/dex/deployed/init.sls
TeddyAndrieux: Right ok make sense

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
```diff
@@ -0,0 +1,45 @@
...
+        oauth_auto_login: true
+      auth.generic_oauth:
+        api_url: "{{ control_plane_ingress_endpoint }}/oidc/userinfo"
+        auth_url: "{{ control_plane_ingress_endpoint }}/oidc/auth"
+        client_id: grafana-ui
+        client_secret: 4lqK98NcsWG5qBRHJUqYM1
+        enabled: true
+        role_attribute_path: contains(`{{ dex.spec.config.staticPasswords | map(attribute='email') | list | tojson }}`, email) && 'Admin'
+        scopes: openid profile email groups
+        tls_skip_verify_insecure: true
```
gdemonet, line 42: Oh really? :disappointed: (not something you need to take care of, but we'll have to open a ticket)

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -96,15 +109,15 @@ Create kube-apiserver Pod manifest:
...
-          - --oidc-ca-file=/etc/metalk8s/pki/nginx-ingress/ca.crt
-          - --oidc-username-claim=email
-          - --oidc-groups-claim=groups
-          {%- endif %}
+          {%- if oidc_config %}
+          - --oidc-issuer-url={{ oidc_config.issuerURL }}
+          - --oidc-client-id={{ oidc_config.clientID }}
+          - --oidc-ca-file={{ oidc_config.CAFile }}
+          - --oidc-username-claim={{ oidc_config.usernameClaim }}
+          - --oidc-groups-claim={{ oidc_config.groupsClaim }}
```
gdemonet, line 117: Since these are k-a settings only, wouldn't it make more sense to define in `pillar.kubernetes` instead?

buildchain/buildchain/salt_tree.py
gdemonet: Good point @TeddyAndrieux, we could just deploy and expect things not to work (on client-side)... For the UX aspect, however, I'd like to review how it looks like when we try to load these UIs in the browser.

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
alexandre-allard: Ha! It was just to see if you really review the PR!

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
alexandre-allard: That's a good question, someone has hacked my github account and changed it I presume...

salt/tests/unit/formulas/data/base_pillar.yaml
alexandre-allard: done

salt/metalk8s/kubernetes/apiserver/installed.sls
alexandre-allard: done

salt/tests/unit/modules/files/test_metalk8s_service_configuration.yaml
TeddyAndrieux: I think, we do, since we never catch the exception today so we always expect the ConfigMap to exists

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
TeddyAndrieux: Done

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Already discussed offline on Slack and fixed in this PR + a ticket has been opened for MetalK8s UI

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
```diff
@@ -0,0 +1,44 @@
...
+        data: /var/lib/grafana/
+        logs: /var/log/grafana
+        plugins: /var/lib/grafana/plugins
+        provisioning: /etc/grafana/provisioning
+      log:
+        mode: console
+      server:
+        root_url: "{{ control_plane_ingress_endpoint }}/grafana"
+      auth:
+        disable_login_form: true
```
gdemonet, line 31: That's worth documenting in the changelog, no?

docs/installation/bootstrap.rst
```diff
@@ -207,6 +208,20 @@ defaults kubernetes configuration.
   configure the corresponding entries in the
   ``kubernetes.apiServer.featureGates`` mapping.
 
+  If ``dex`` is enabled, it will be used as ``oidc`` for ``kube-apiserver``
+  but you can use a `specific OpenID for kube-apiserver`_, to do so:
```
gdemonet, line 212: Maybe mention this in the paragraph above about disabling dex, no?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # CHANGELOG
 ## Release 123.0.0 (in development)
 
+### Additions
+
+- Add ability to deploy MetalK8s without Dex and possibility to configure
+  our own IDP for kube-apiserver, Grafana and MetalK8s UI
```
```suggestion
  your own IDP for kube-apiserver, Grafana and MetalK8s UI
```

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
TeddyAndrieux: To me no, it does not change anything compared to the current behavior since anyway when we used dex `oauth_auto_login` was set to True so we never reached the Grafana login form

docs/installation/bootstrap.rst
TeddyAndrieux: Hmm right I will mention apiserver as well

salt/metalk8s/addons/prometheus-operator/config/grafana.yaml.j2
gdemonet: Oh good point :+1:

docs/installation/bootstrap.rst
TeddyAndrieux: Btw I'm not sure it really makes sense to put this in the dex section,
Because if you disable Dex there is no impact on apiserver itself, except that it will not use Dex but it's obvious. To me what we want to reflect is that oidc for apiserver is configurable and default to dex when it's deployed. 
WDYT ?

gdemonet: Ah right, we only have Grafana / MetalK8s UI if we use Dex, today at least... So indeed, no need to add anything to the addons paragraph, until we manage these.

gdemonet (approved):

TeddyAndrieux (approved):
~/bypass_author_approval~ (it does not work in review comment apparently)

PR #3702, opened 2022-02-14
charts,salt,build: Bump Prometheus Adapter chart to 3.0.2

TeddyAndrieux:
Bump the Prometheus Adapter chart to 3.0.2 and also bump the
prometheus-adapter image to v0.9.1.

NOTE: The label selector for the Deployment changed we need to
remove the deployment before doing the upgrade or the downgrade.


salt/metalk8s/addons/prometheus-adapter/deployed/init.sls
```diff
@@ -1,3 +1,31 @@
...
+    - kind: Deployment
+    - name: prometheus-adapter
+    - namespace: metalk8s-monitoring
+    - wait:
+        attempts: 30
+        sleep: 10
+    - require_in:
+      - sls: metalk8s.addons.prometheus-adapter.deployed.chart
+
+{%- endif %}
```
alexandre-allard, line 31: I think the change here and for downgrade should be in a dedicated commit so we can easily do a git revert in 124.0.
Moreover, I was thinking we should create an issue to start tracking this kind of things, so when we create the next branch, we just have to look at the list of things to remove instead of searching/grepping the code and potentially miss some stuff.

TeddyAndrieux: You're right, it's kind of part of the "bump" in a sense but agree that for revert it makes sense to move it to another commit, fair I will do it

alexandre-allard (approved):

PR #3703, opened 2022-02-15
ci: Fix the final status artifact step name

alexandre-allard:


eve/main.yml
```diff
@@ -1868,7 +1868,7 @@ stages:
           env:
             <<: *_env_final_status_artifact_success
             <<: *_install_single-node_junit_info
-            STEP_NAME: single-node-install-rhel
```
TeddyAndrieux, line 1871: Oh ! It was broken even if we tried to run with centos-7 then :smile:

TeddyAndrieux (approved):

gdemonet (approved):

Issue #2060, opened 2019-11-20
Migrate stored version of K8s objects after upgrade or downgrade


gdemonet:
**Component**: kubernetes

**Why this is needed**:

> Today it is possible to create API objects (e.g., HPAs) in one version of Kubernetes, go through multiple upgrade cycles without touching those objects, and eventually arrive at a version of Kubernetes that can’t interpret the stored resource and crashes. See k8s.io/pr/52185.

See:

- [Storage migration KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/0030-storage-migration.md)
- [`StorageVersionHash` KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/35-storage-version-hash.md)
- [A migration controller](https://github.com/kubernetes-sigs/kube-storage-version-migrator) that implements the [automatic storage migration KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/storage-migration-auto-trigger.md)

**What should be done**: Ensure all objects are stored in the "latest" version before and after any upgrade or downgrade.

**Implementation proposal** (strongly recommended):

Since we control the upgrade/downgrade of the cluster ourselves, it may not make sense to have an automatic migration in place.
For the logic, get inspiration from the linked KEPs and controller. If `StorageVersionHash` is not available (before K8s 1.14), use the "old" approach to "get" + "replace" all objects.

**Test plan**:

- Have an automatic check that validates the migration tooling independently (create some object in an old version, run migration, check new version) - _Note: would make sense to wait for `StorageVersionHash` to be available before implementing this_
- Watch an object's stored apiVersion during an upgrade (observe two version switches, if possible)

NicolasT:
https://github.com/kubernetes-sigs/kube-storage-version-migrator could be of use to get this done.

PR #3704, opened 2022-02-15
charts,salt,tests: Make Ingress controller watch Ingress without class

TeddyAndrieux:
The Workload Plane Ingress is the default Ingress controller that should
be used for Ingress object, so let's make this Ingress controller watch
Ingress object with no class.

NOTE: It shouldn't be needed since we define a default IngressClass
object, but during upgrade, the Ingress object created without Class
in MetalK8s 2.10.x are ignored after upgrade with a message that the
Ingress objects do not have any Class.

### DIFF
--- a/tests/post/features/ingress.feature
+++ b/tests/post/features/ingress.feature
@@ -12,6 +12,27 @@ Feature: Ingress
         When we perform an HTTPS request on port 443 on a workload-plane IP
         Then the server returns 404 'Not Found'
 
+    Scenario: Create new Ingress object (without class)
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
+        When we create a 'metalk8s-test-1' Ingress on path '/_metalk8s-test-1' on 'repositories' service on 'http' in 'kube-system' namespace
+        And we perform an HTTPS request on path '/_metalk8s-test-1' on port 443 on a workload-plane IP
+        Then the server returns 200 'OK'
+
+    Scenario: Create new Ingress object (nginx class)
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=ingress-nginx' are 'Ready'
+        When we create a 'metalk8s-test-2' Ingress with class 'nginx' on path '/_metalk8s-test-2' on 'repositories' service on 'http' in 'kube-system' namespace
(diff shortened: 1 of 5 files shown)


gdemonet:
Could be linked with their "stored apiVersion" in etcd... which relates to #2060

TeddyAndrieux:
> Could be linked with their "stored apiVersion" in etcd... which relates to #2060

I'm not sure, maybe, because the controller can handle the Ingress it's just that apparently the `IngressClass` with "default" annotation, just set the IngressClassName on the Ingress object on creation so if the Ingress exists before the `IngressClass` then the Class will not be set.

tests/post/features/ingress.feature
```diff
@@ -12,6 +12,13 @@ Feature: Ingress
         When we perform an HTTPS request on port 443 on a workload-plane IP
         Then the server returns 404 'Not Found'
 
+    Scenario: Create new Ingress object
```
gdemonet: Maybe note that:

```suggestion
    Scenario: Create new Ingress object (with default class)
```

Not sure if we want to add tests for specifying the class (hence, with the default `nginx`, but also with the `nginx-control-plane` and some unknown value)...

gdemonet:
> I'm not sure, maybe, because the controller can handle the Ingress it's just that apparently the `IngressClass` with "default" annotation, just set the IngressClassName on the Ingress object on creation so if the Ingress exists before the `IngressClass` then the Class will not be set.

You're right, I think the issue is rather that "old" Ingress objects are not updated with an `ingressClassName` when migrated, we would need to "re-create" them, as suggested in the [ingress-nginx FAQ](https://kubernetes.github.io/ingress-nginx/#i-have-only-one-ingress-controller-in-my-cluster-what-should-i-do). Anyway, this PR implements the alternative solution, so LGTM!

tests/post/features/ingress.feature
TeddyAndrieux: Done, I did not test the control-plane Ingress creation since we already create some of them as part of MetalK8s so it's already tested and this PR is not really about Control Plane Ingress (we can easily add more in the future if we want to)

gdemonet (dismissed):
:100:  :shipit:

gdemonet (approved):
(not sure what changed, you just added a `time.sleep(5)` after the Ingress creation?)

TeddyAndrieux:
> (not sure what changed, you just added a `time.sleep(5)` after the Ingress creation?)

Yes and also added missing "teardown" on tests

PR #3711, opened 2022-02-21
build,packages: Bump containerd to 1.6.0

TeddyAndrieux:
The pause image has been bumped accordingly to 3.6


gdemonet (approved):
Buildchain changes LGTM, I'll spend more time looking at the release notes in details (nothing seems problematic at first sight, but I want to investigate a bit more, especially around the added metrics and if/how we consume them).

PR #3712, opened 2022-02-21
build,packages,salt: Update Calico to 3.22.0 

TeddyAndrieux:
NOTE: cni-plugin repository from Calico does no longer exist now the CNI
plugin is part of the main calico repository.
That's why we need to extract only the CNI plugin from this repository
when building the Calico CNI plugin package.


gdemonet (dismissed):
Buildchain changes LGTM, I'll spend more time looking at the release notes in details (nothing seems problematic at first sight, but I want to investigate a bit more).

gdemonet (approved):

PR #3715, opened 2022-02-25
ARTESCA-2606 - fix ui crashed when alert manager is not available

MonPote:


JBWatenbergScality (approved):

ui/src/hooks.js
```diff
@@ -101,7 +101,7 @@ export const useVolumesWithAlerts = (nodeName?: string) => {
   //This forces alerts to have been fetched at least once (watchdog alert should be present)
   // before rendering volume list
   // TODO enhance this using useAlerts status
-  if (alerts.length === 0) return [];
+  if (!alerts || alerts.length === 0) return [];
```
gdemonet, line 104: Can't we use null coalescing for this?

```suggestion
  if (alerts?.length ?? 0 === 0) return [];
```

Or just rely on the "falsiness" of `0`, as is done in the other change below:

```suggestion
  if (!alerts?.length) return [];
```

ui/src/services/alertUtils.js
```diff
@@ -175,7 +175,7 @@ export const getHealthStatus = (
   alerts: Alert[],
   activeOn?: string = new Date().toISOString(),
 ): Health => {
-  if (!alerts.length) return STATUS_HEALTH;
+  if (!alerts || !alerts.length) return STATUS_HEALTH;
```
gdemonet, line 178:
```suggestion
  if (!alerts?.length) return STATUS_HEALTH;
```

ui/src/hooks.js
MonPote: We can but I have the feeling that it's a bit difficult to read.
Especially in this case : `alerts?.length ?? 0 === 0`.
```diff
-  if (!alerts || alerts.length === 0) return [];
+  if (alerts?.length ?? 0 === 0) return [];
```

I choose to just add `!alerts ||` for consistency but if you want I can change at least : 
```diff
- if (!alerts || !alerts.length) return STATUS_HEALTH;
+ if (!alerts?.length) return STATUS_HEALTH;
```

PR #3723, opened 2022-03-09
ARTESCA-3564 Add a precheck about containerd fstype

sayf-eddine-scality:
**Component**: Salt

**Context**:  Check if containerd folder is on an XFS, then it should have ftype=1

**Summary**:

**Acceptance criteria**: 

---

Closes: #ARTESCA-3564

### DIFF
--- a/salt/_modules/metalk8s_checks.py
+++ b/salt/_modules/metalk8s_checks.py
@@ -4,9 +4,10 @@ Execution module for handling MetalK8s checks.
 """
 
 import ipaddress
+import os
 import re
 
-from salt.exceptions import CheckError
+from salt.exceptions import CheckError, CommandExecutionError
 
 __virtualname__ = "metalk8s_checks"
 
@@ -43,6 +44,13 @@ def node(raises=True, **kwargs):
     ports_ret = __salt__["metalk8s_checks.ports"](raises=False, **kwargs)
     if ports_ret is not True:
         errors.append(ports_ret)
+
+    # Run `containerd filesystem` checks
+    fs_checks = __salt__["metalk8s_checks.containerd_filesystem"](
+        raises=False, **kwargs
+    )
+    if fs_checks is not True:
+        errors.append(fs_checks)
 
     # Run `route_exists` check for the Service Cluster IPs
     service_cidr = kwargs.pop(
@@ -180,7 +188,7 @@ def ports(
     raises=True,
     listening_process_per_role=None,
     roles=None,
-    **kwargs
+    **kwargs,
 ):
     """Check if an unexpected process already listening on a port on the machine,
     return a string (or raise if `raises` is set to `True`) with the list of
@@ -431,6 +439,35 @@ def route_exists(destination, raises=True):
     if error and raises:
         raise CheckError(error)
 
+    return error or True
+
+
+def containerd_filesystem(raises=True, **_kwargs):
+    error = None
+    folder = "/var/lib/containerd"
+    while not os.path.exists(folder):
+        folder = os.path.dirname(folder)
+    res = __salt__["cmd.run_all"](f"df --output=source,fstype {folder}")
+    if res["retcode"] != 0:
+        error = f"Cannot check filesystem of {folder}, ({res['stderr']})"
+    else:
+        device = res["stdout"].split()[2]
+        fstype = res["stdout"].split()[3]
+        # We use the exact output to prevent false positives
+        if fstype != "xfs":
+            # Not an XFS filesystem, no need to check
+            return True
+        try:
+            infos = __salt__["xfs.info"](device)
+            ftype = infos["naming"]["ftype"]
+            if ftype != "1":
+                error = (
+                    f"Containerd XFS filesystem ({folder}) has ftype={ftype} expected 1"
+                )
+        except CommandExecutionError as ex:
+            error = f"Error checking ftype value from {folder}, ({ex})"
+    if error and raises:
+        raise CheckError(error)
     return error or True
 
 
--- a/salt/tests/unit/modules/files/test_metalk8s_checks.yaml
+++ b/salt/tests/unit/modules/files/test_metalk8s_checks.yaml
@@ -5,6 +5,8 @@ node:
     services_ret: True
     ports_ret: True
     route_exists_ret: True
+    fs_check_ret: True
+
     pillar:
       networks:
         service: 10.96.0.0/12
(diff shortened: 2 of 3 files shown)


salt/_modules/metalk8s_checks.py
```diff
@@ -44,6 +45,13 @@ def node(raises=True, **kwargs):
     if ports_ret is not True:
         errors.append(ports_ret)
 
+    # Run `containerd filesystem checks`
```
TeddyAndrieux:
```suggestion
    # Run `containerd filesystem` checks
```

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,23 @@ def route_exists(destination, raises=True):
...
+    folder = folder or "/var/lib/containerd"
+    while not os.path.exists(folder):
+        folder = os.path.dirname(folder)
+    try:
+        infos = __salt__["xfs.info"](folder)
+        ftype = infos["naming"]["ftype"]
+        if ftype != "1":
+            error = f"XFS filesystem has ftype={ftype} expected 1"
+    except CommandExecutionError:
+        error = f"{folder} is not on an XFS"
```
TeddyAndrieux: It's not an error just we want to check if it's xfs then `ftype` should be set to `1` but it could be something else than xfs it's ok also. 
Maybe we should enforce ext4 but for the moment I think if it's not XFS then it's ok **but** IMHO we should really check if it's xfs and not rely on "xfs.info" failure

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -64,6 +64,7 @@ Check node:
         # Sees: https://github.com/saltstack/salt/issues/28503
         service_cidr: {{ pillar.networks.service }}
         roles: {{ pillar.metalk8s.nodes[node_name].roles }}
+        folder: "/var/lib/containerd"
```
TeddyAndrieux: Hmm to me you should keep the default from the module do not need to override it here

TeddyAndrieux (dismissed):

TeddyAndrieux:
You will also need to add unit test for it (`salt/tests/unit`)

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,31 @@ def route_exists(destination, raises=True):
...
+        if "Type\nxfs" not in res['stdout']:
+            # Not an XFS filesystem, no need to check
+            return True
+        try:
+            infos = __salt__["xfs.info"](folder)
+            ftype = infos["naming"]["ftype"]
+            if ftype != "1":
+                error = f"XFS filesystem has ftype={ftype} expected 1"
+        except CommandExecutionError:
+            error = f"Error checking ftype value from {folder}"
```
TeddyAndrieux: Can you just add the exception message in the error

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,31 @@ def route_exists(destination, raises=True):
...
+    else:
+        # We use the exact output to prevent false positives
+        if "Type\nxfs" not in res['stdout']:
+            # Not an XFS filesystem, no need to check
+            return True
+        try:
+            infos = __salt__["xfs.info"](folder)
+            ftype = infos["naming"]["ftype"]
+            if ftype != "1":
+                error = f"XFS filesystem has ftype={ftype} expected 1"
```
TeddyAndrieux: Can you add the "folder" in the error message
Something like that
```suggestion
                error = f"'Containerd XFS filesystem ({folder}) has ftype={ftype} expected 1"
```

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,33 @@ def route_exists(destination, raises=True):
...
 
 
+def containerd_filesystem(folder=None, raises=True, **kwargs):
+    error = None
+    folder = folder or "/var/lib/containerd"
+    while not os.path.exists(folder):
+        folder = os.path.dirname(folder)
+    res = __salt__["cmd.run_all"](f"df --output=fstype {folder}")
+    if res["retcode"] != 0:
+        error = f"Cannot check filesystem type of {folder}"
```
TeddyAndrieux: I think it worth adding at least stderr in the error

salt/tests/unit/modules/files/test_metalk8s_checks.yaml
```diff
@@ -54,7 +56,13 @@ node:
...
       https://github.com/kubernetes/kubernetes/issues/57534#issuecomment-527653412).
 
-  # 7. Error: Combined errors
+  # 7. Error: fs_check
+  - <<: *node_nominal
+    fs_check_ret: "Containerd XFS filesystem (/var/lib/containerd) has ftype=2 expected 1"
+    expect_raise: True
+    result: 'Node my_node_1: Containerd XFS filesystem \(/var/lib/containerd\) has ftype=2 expected 1'
+
+  # 8. Error: Combined errors
```
TeddyAndrieux, line 65: You should add an fs_check error in this one as well so that we ensure that we properly get all messages at once when you have several issues

salt/tests/unit/modules/files/test_metalk8s_checks.yaml
```diff
@@ -601,3 +609,44 @@ route_exists:
     error: >-
       A route was found for 10.96.0.0, but it does not match the full
       destination network 10.96.0.0/12
+
+containerd_filesystem:
+  # 1. Success Not XFS
+  - run_all_df_ret:
+      stdout: "Type\next4"
```
TeddyAndrieux: nitpick: Maybe you can put the full string not using `\n` so that we easily see that it's a df output

```suggestion
      stdout: |-
        Type
        ext4
```

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,33 @@ def route_exists(destination, raises=True):
     return error or True
 
 
+def containerd_filesystem(folder=None, raises=True, **kwargs):
+    error = None
+    folder = folder or "/var/lib/containerd"
```
TeddyAndrieux: TBH I don't think we really need this folder to be configurable, but as you wish

salt/tests/unit/modules/test_metalk8s_checks.py
```diff
@@ -180,6 +183,40 @@ def test_sysctl(self, params, data, result, raises=False):
...
+            )
+        else:
+            xfs_info_mock.return_value = xfs_info_ret
+
+        patch_dict = {
+            "cmd.run_all": run_all_df_mock,
+            "xfs.info": xfs_info_mock
+        }
+
+        with patch.dict(metalk8s_checks.__salt__, patch_dict):
```
TeddyAndrieux: I think you want to mock and test the `os.path.exists` otherwise I think if `/var/lib/containerd` (and `/foo/bar`) directory exists when you run the test you do not get 100% coverage
Which is sad that test coverage depends on the environment where you run the test :smile:

TeddyAndrieux (dismissed):

salt/_modules/metalk8s_checks.py
```diff
@@ -434,6 +442,33 @@ def route_exists(destination, raises=True):
     return error or True
 
 
+def containerd_filesystem(raises=True, **kwargs):
```
TeddyAndrieux:
```suggestion
def containerd_filesystem(raises=True, **_kwargs):
```
To have the argument but ignored

TeddyAndrieux (approved):

PR #3730, opened 2022-03-18
ARTESCA-3572 Allow the removal of Metalk8s archives in iso-manager

sayf-eddine-scality:
**Component**:

**Context**: 
Allow the removal of the iso
**Summary**:
```
./iso-manager.sh --rm-archive /root/metalk8s-2.10.7.iso
```
**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   `metalk8s-utils` container
   (PR[#3773](https://github.com/scality/metalk8s/pull/3773))
 
+- Add a `-r`/`--rm-archive` option to the `iso-manager.sh` script, allowing to
+  remove MetalK8s ISOs from a cluster
+  (PR[#3730](https://github.com/scality/metalk8s/pull/3730))
+
 ### Removals
 
 - The `Statefulsets` Grafana dashboard has been removed
@@ -27,6 +31,13 @@
 - Remove the `calico-cni-plugin` RPM package and rely instead on
   the `calico-cni` container to deploy the CNI binaries on the host
   (PR[#3793](https://github.com/scality/metalk8s/pull/3793))
+
+### Deprecations
+
+- The long name `--archive` for the "add" option to the `iso-manager.sh` script
+  is deprecated in favor of `--add-archive` (shorthand `-a` is still valid).
+  This deprecated option will be removed in MetalK8s 124.0.0
+  (PR[#3730](https://github.com/scality/metalk8s/pull/3730))
 
 ### Enhancements
 
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -2183,6 +2183,26 @@ stages:
             # time to stabilize (e.g.: Rolling update of some DaemonSet that may take
             # some times, especially in multi node context)
             STABILIZATION_TIME: "240"
+      # --- Remove version N-1 ---
+      - SetPropertyFromCommand:
+          name: Save previous ISO path as a property
+          property: previous_iso_path
+          command: >
+            ssh -F ssh_config bootstrap
+            "sudo salt-call slsutil.renderer --out=txt
+            string=\"{{ salt.metalk8s.get_archives()['metalk8s-%(prop:product_promoted_version)s']['iso'] }}\"
+            | cut -c 8-"
+          workdir: *terraform_workdir
+          haltOnFailure: true
+      - ShellCommand:
+          name: Remove previous ISO before running tests
+          command: >
+            ssh -F ssh_config bootstrap "
+            sudo bash
+            /srv/scality/metalk8s-%(prop:metalk8s_version)s/iso-manager.sh
+            --rm-archive \"\$(readlink -f '%(prop:previous_iso_path)s')\" --verbose"
+          workdir: *terraform_workdir
+          haltOnFailure: true
       # --- Test version N ---
       - ShellCommand: *git_pull_ssh
       - ShellCommand:
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -149,6 +149,14 @@ def _get_archive_info(info):
 
 
 def archive_info_from_product_txt(archive):
+    """Extract archive information from the 'product.txt' file in an archive.
+
+    Will ensure that the information describes a valid MetalK8s archive (including a
+    version).
+
+    Arguments:
+        archive (str): path to an ISO or a directory
+    """
     if os.path.isdir(archive):
         info = archive_info_from_tree(archive)
         info.update(
@@ -168,6 +176,18 @@ def archive_info_from_product_txt(archive):
     else:
         raise CommandExecutionError(
             "Invalid archive path {0}, should be an iso or a directory.".format(archive)
+        )
+
+    if info["name"] != "MetalK8s":
+        raise CommandExecutionError(
+            "Invalid archive '{}', 'NAME' should be 'MetalK8s', found '{}'.".format(
+                archive, info["name"]
+            )
+        )
+
+    if not info["version"]:
+        raise CommandExecutionError(
+            "Invalid archive '{}', 'VERSION' must be provided.".format(archive)
         )
 
     return info
@@ -216,6 +236,32 @@ def archive_info_from_iso(path):
         )
 
     return _get_archive_info(result["stdout"])
+
+
+def get_mounted_archives():
+    """List the MetalK8s ISOs mounted under /srv/scality."""
+    archives = {}
+    active_mounts = __salt__["mount.active"]()
+
+    for mountpoint, mount_info in active_mounts.items():
+        if not mountpoint.startswith("/srv/scality/metalk8s-"):
+            continue
+
+        archive = mount_info["alt_device"]
+        try:
+            archive_info = archive_info_from_product_txt(archive)
+        except CommandExecutionError as exc:
+            log.info(
+                "Skipping %s (mounted at %s): not a valid MetalK8s archive - %s",
+                archive,
+                mountpoint,
+                exc,
+            )
+            continue
+
+        env_name = "metalk8s-{0}".format(archive_info["version"])
+        archives[env_name] = archive_info
+    return archives
 
 
 def get_archives(archives=None):
@@ -669,8 +715,10 @@ def _write_bootstrap_config(config):
 
 def configure_archive(archive, remove=False):
     """Add (or remove) a MetalK8s archive in the bootstrap config file."""
-    # raise if the archive does not exist or is invalid
-    archive_info_from_product_txt(archive)
+    if not remove:
+        # raise if the archive does not exist or is invalid
+        archive_info_from_product_txt(archive)
+
     config = _read_bootstrap_config()
 
     if remove:
--- a/salt/metalk8s/archives/mounted.sls
+++ b/salt/metalk8s/archives/mounted.sls
@@ -1,27 +1,14 @@
 include:
   - metalk8s.salt.minion.local
 
-{%- for _, archive in salt.metalk8s.get_archives().items() %}
-  {%- if archive.iso %}
-    {%- if not archive.version %}
+{%- for archive in salt.metalk8s.get_archives().values() | selectattr("iso") %}
 
-Archive {{ archive.iso }} available:
-  test.configurable_test_state:
-  - changes: false
-  - result: false
-  - comment: >
-      Unable to retrieve archive version from given archive file
-      {{ archive.iso }}, assumed to be an ISO archive containing a
-      "product.txt" file.
-
-    {%- else %}
-
-Archive path {{ archive.path }} exists:
+Create archive mountpoint '{{ archive.path }}':
   file.directory:
   - name: {{ archive.path }}
   - makedirs: true
 
-Archive {{ archive.iso }} available at {{ archive.path }}:
+Mount '{{ archive.iso }}' at '{{ archive.path }}':
   mount.mounted:
   - name: {{ archive.path }}
   - device: {{ archive.iso }}
@@ -34,23 +21,6 @@ Archive {{ archive.iso }} available at {{ archive.path }}:
   - match_on:
     - name
   - require:
-    - file: Archive path {{ archive.path }} exists
-  - require_in:
-    - file: Assert '{{ archive.path }}/product.txt' exists
-
-    {%- endif %}
-
-  {%- endif %}
-
-Assert '{{ archive.path }}/product.txt' exists:
-  file.exists:
-  - name: {{ archive.path }}/product.txt
-
-Assert archive '{{ archive.path }}' is MetalK8s:
-  cmd.run:
-  - name: grep '^NAME=MetalK8s$' {{ archive.path }}/product.txt > /dev/null
-  - stateful: true
-  - require:
-    - file: Assert '{{ archive.path }}/product.txt' exists
+    - file: Create archive mountpoint '{{ archive.path }}'
 
 {%- endfor %}
--- a/salt/metalk8s/archives/unmounted.sls
+++ b/salt/metalk8s/archives/unmounted.sls
@@ -1,0 +1,43 @@
+{# Remove mounted ISOs which were removed from configuration #}
+{%- set desired = salt.metalk8s.get_archives() %}
+{%- set available = salt.metalk8s.get_mounted_archives() %}
+{%- set ns = namespace(found_archives_to_remove=false) %}
+
+{%- for env, archive in available.items() %}
+  {%- if env not in desired %}
+    {%- set ns.found_archives_to_remove = true %}
+
+Unmount '{{ archive.iso }}' from '{{ archive.path }}':
+  mount.unmounted:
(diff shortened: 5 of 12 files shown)


scripts/iso-manager.sh
```diff
@@ -84,6 +90,12 @@ _add_archives() {
     done
 }
 
+_remove_archives() {
+    for archive; do
+        $SALT_CALL metalk8s.configure_archive "$archive" remove=True
```
TeddyAndrieux, line 106: Not really due to this PR but it's really sad that even when you try to remove an archive, we need this one to be there and available 

IMO this line in the module shouldn't be there when we remove 
```python
    # raise if the archive does not exist or is invalid
    archive_info_from_product_txt(archive)
```

salt/_modules/metalk8s.py
```diff
@@ -218,6 +218,34 @@ def archive_info_from_iso(path):
     return _get_archive_info(result["stdout"])
 
 
+def get_configured_archives():
+    """List the already configured metalk8s archives"""
+    archives = {}
+    base_mount = "/srv/scality"
+    for path in os.listdir(base_mount):
```
TeddyAndrieux: Why do we check all directories, IMHO we should just check for mountpoint

salt/_modules/metalk8s.py
```diff
@@ -218,6 +218,34 @@ def archive_info_from_iso(path):
...
+        if not path.startswith("metalk8s-"):
+            continue
+        path = os.path.join(base_mount, path)
+        archive = archive_info_from_product_txt(path)
+        # Check if the archive is really configured or just mounted
+        env_name = "metalk8s-{0}".format(archive["version"])
+        if not os.path.exists(f"/var/lib/metalk8s/repositories/conf.d/99-{env_name}-registry.inc"):
+            continue
```
TeddyAndrieux: I'm not sure we want to check this cause if for whatever reason this file does not exist we will not umount the ISO, to me we manage everything in `/srv/scality/` so if there is "something" mounted in this directory that we don't know we just umount it.

Especially since we first reconfigure the registry before actually umount the ISO it means we will never umount ISOs since when we run the state to umount the ISOs the registry file is already deleted

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,10 @@
+{% set removable_archives = salt.metalk8s.get_removable_archives() %}
+{% for env, archive in removable_archives.items() %}
+{% if archive.iso %}
+Archive {{ archive.iso }} absent at {{ archive.path }}:
+  mount.unmounted:
+    - name: {{ archive.path }}
+    - device: {{ archive.iso }}
+    - persist: true
```
TeddyAndrieux, line 14: It properly clean the fstab entry ? Cool :+1:

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,10 @@
+{% set removable_archives = salt.metalk8s.get_removable_archives() %}
+{% for env, archive in removable_archives.items() %}
+{% if archive.iso %}
+Archive {{ archive.iso }} absent at {{ archive.path }}:
```
TeddyAndrieux: I know it comes from the state name from `mounted.sls` but I suggest putting a name a bit clearer
```suggestion
Unmount archive {{ archive.iso }} from {{ archive.path }}:
```

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,10 @@
...
+  mount.unmounted:
+    - name: {{ archive.path }}
+    - device: {{ archive.iso }}
+    - persist: true
+{% endif %}
```
TeddyAndrieux: IMO we also want to remove the directory when we umount, only if the directory is empty

Something like that (not tested)
```yaml
Cleanup {{ archive.path }} directory:
  file.absent:
    - name: {{ archive.path }}
    - unless: file.path_exists_glob("{{ archive.path }}/*")
```

salt/metalk8s/repo/configured.sls
```diff
@@ -48,3 +48,10 @@ Generate container registry configuration:
     - backup: false
     - defaults:
         archives: {{ archives }}
+
+{% set removable_archives = salt.metalk8s.get_removable_archives() %}
+{% for env in removable_archives %}
+Remove config for env {{ env }}:
+    file.absent:
+      - name: {{ salt.file.join(repo.config.directory, '99-' ~ env ~ '-registry.inc') }}
+{% endfor %}
```
TeddyAndrieux: Again, to me we manage the full `repo.config.directory` so if we don't know a file in this directory we just remove it, since it's a bit complicated with solutions I suggest just checking all `99-metalk8s-*` we remove all files that are not expected

TeddyAndrieux (commented):
Few comments (if you add salt module you also need to add unit tests :wink: )

salt/metalk8s/repo/configured.sls
gdemonet: :thinking: what if the Solution is called `metalk8s-solution-example`..? I'll go with a regex such as `99-metalk8s-[0-9].*`, should be enough.

Also, this state is a bit strange, we're not really managing all the includes, only adding one for the current version :confused:... I'll leave it like that but still confusing IMHO

gdemonet: :thinking: what if the Solution is called `metalk8s-solution-example`..? I'll go with a regex such as `99-metalk8s-[0-9].*`, should be enough.

Also, this state is a bit strange, we're not really managing all the includes, only adding one for the current version :confused:... I'll leave it like that but still confusing IMHO

TeddyAndrieux: Maybe solutions shouldn't have the "99-" prefix but another one, or yes just a simple regex

> Also, this state is a bit strange, we're not really managing all the includes, only adding one for the current version confused... I'll leave it like that but still confusing IMHO

That is my point, to me, we may want to change this state so that it basically "computes" the full directory content

CHANGELOG.md
```diff
@@ -24,6 +28,13 @@
   the `metalk8s-utils` container
   (PR[#3773](https://github.com/scality/metalk8s/pull/3773))
 
+### Deprecations
+
+- The long name `--archive` for the "add" option to the `iso-manager.sh` script
+  is deprecated in favor of `--add-archive` (shorthand `-a` is still valid).
+  This deprecated option will be removed in MetalK8s 123.1.0
```
TeddyAndrieux: To me, it should be removed in 124.0.0 cause you can upgrade from 2.11 to 123.1 (which means it gets removed directly with no "deprecations")

eve/main.yml
```diff
@@ -2183,6 +2183,19 @@ stages:
             # time to stabilize (e.g.: Rolling update of some DaemonSet that may take
             # some times, especially in multi node context)
             STABILIZATION_TIME: "240"
+      # --- Remove version N-1 ---
+      - ShellCommand:
+          name: Remove previous ISO before running tests
+          command: >
+            ISO_PATH=$(
+            [[%(prop:nodes_count:-0)s -eq 0]] && echo "metalk8s.iso"
+            || echo "/archives/metalk8s.iso") &&
```
TeddyAndrieux: :cry: 
Can we get the path using salt instead ?
With `metalk8s.get_archives` module and `%(prop:product_promoted_version)s`
Something like that 
```console
$ salt-call slsutil.renderer --out=txt string="{{ salt.metalk8s.get_archives()['metalk8s-%(prop:product_promoted_version)s']['iso'] }}" | cut -c 8-
/archives/metalk8s.iso
```

salt/_modules/metalk8s.py
```diff
@@ -218,6 +238,33 @@ def archive_info_from_iso(path):
...
+        if not mountpoint.startswith("/srv/scality/metalk8s-"):
+            continue
+
+        archive = mount_info["alt_device"]
+        try:
+            archive_info = archive_info_from_product_txt(archive)
+        except CommandExecutionError as exc:
+            log.info(
+                "Skipping %s (mounted at %s): not a valid MetalK8s archive"
+                "(failed to read/parse 'product.txt' contents): %s",
```
TeddyAndrieux: nitpick: Not sure it is worth mentioning "failed to read/parse" since we put the exception message, especially since you added a check on name and version (in those cases the read and parse of the product.txt worked well just not a valid archive)

salt/metalk8s/archives/mounted.sls
```diff
@@ -1,27 +1,14 @@
 include:
   - metalk8s.salt.minion.local
 
-{%- for _, archive in salt.metalk8s.get_archives().items() %}
-  {%- if archive.iso %}
-    {%- if not archive.version %}
+{%- for archive in salt.metalk8s.get_archives().values() | selectattr("iso") %}
```
TeddyAndrieux, line 4: :+1:

salt/metalk8s/archives/unmounted.sls
TeddyAndrieux: Did you check this btw ?

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,31 @@
...
+    - name: {{ archive.path }}
+    - device: {{ archive.iso }}
+    - persist: true
+
+Remove archive mountpoint '{{ archive.path }}':
+  file.absent:
+    - name: {{ archive.path }}
+    - unless:
+      - fun: file.path_exists_glob
+        path: {{ archive.path }}/*
```
TeddyAndrieux, line 21: So in this case we just silently ignore it and we do not remove the directory ? :thinking: 
Ok fine

scripts/iso-manager.sh
```diff
@@ -48,6 +57,13 @@ while (( "$#" )); do
   esac
 done
 
+if [ -z ${ARCHIVE_DEPRECATED_OPTION} ]; then
+  echo "Warning: Usage of the '--archive' option is deprecated, and will be removed" \
+       "in the next MetalK8s minor version, 123.1.0. Please make sure to update your" \
```
TeddyAndrieux: ditto: 124.0.0

scripts/iso-manager.sh
```diff
@@ -13,18 +13,27 @@ SALTENV=""
 _usage() {
     echo "iso-manager.sh [options]"
     echo "Options:"
-    echo "-a/--archive <archive path>:     Path to archive folder or ISO"
-    echo "-l/--log-file <logfile_path>:    Path to log file"
-    echo "-v/--verbose:                    Run in verbose mode"
-    echo "-d/--dry-run:                    Run actions in dry run mode"
+    echo "-a/--add-archive <archive path>:     Path to an archive folder or ISO to add"
+    echo "    [DEPRECATED] You may also use --archive as an equivalent to the above."
+    echo "                 This option name will be removed in MetalK8s 123.1.0"
```
TeddyAndrieux: ditto: 124.0.0

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,31 @@
+{# Remove mounted ISOs which were removed from configuration #}
```
TeddyAndrieux, line 1: I would have preferred a single state `available` (same as what we have for solutions) so that in the future we could have a single state to manage all archives (metalk8s + solutions)

But not really mandatory

salt/tests/unit/modules/files/test_metalk8s.yaml
```diff
@@ -67,6 +67,45 @@ get_archives:
     raises: True
     result: "Invalid archives: list or string expected, got .*"
 
+get_mounted_archives:
+  # 0. Nothing mounted at all
+  - mounts: {}
+    logs: []
+    result: {}
+  # 1. Nothing mounted under /srv/scality/metalk8s-*
```
TeddyAndrieux, line 75: Maybe just add another that looks like a solution that will be in `/srv/scality/`

I would have also added a "complex" test with at least 4 mount points 2 valid MetalK8s 1 invalid (skipped with logs) and one "other" mount just ignored because not under `/srv/scality/metalk8s-*`

scripts/iso-manager.sh
```diff
@@ -84,6 +100,12 @@ _add_archives() {
     done
 }
 
+_remove_archives() {
+    for archive; do
+        $SALT_CALL metalk8s.configure_archive "$archive" remove=True
```
TeddyAndrieux, line 106: Not linked to this PR, but from what I see even if you use `--dry-run` we will still update the bootstrap config

TeddyAndrieux (commented):
Overall LGTM

CHANGELOG.md
gdemonet: Good point, I didn't remember we changed the "one minor at a time" to "one major at a time" :sweat_smile:

eve/main.yml
gdemonet: Will be simpler to maintain indeed, good idea :ok_hand:

salt/_modules/metalk8s.py
gdemonet: True, will change

salt/metalk8s/archives/unmounted.sls
gdemonet: That was your suggestion, I just implemented it :innocent: But it makes sense to tell the user about it, I'll add a state for this.

salt/metalk8s/archives/unmounted.sls
gdemonet: I thought about it, but right now we can't, because we want to first mount new ISOs, reconfigure the repo container, then unmount old ISOs (otherwise we may have issues with it, I think).

salt/metalk8s/archives/unmounted.sls
gdemonet: Yep, just checked (also looked at `/etc/fstab`, to be sure :innocent:):

```
          ID: Unmount '/srv/scality/releases/metalk8s-2.11.7.iso' from '/srv/scality/metalk8s-2.11.7'
    Function: mount.unmounted
        Name: /srv/scality/metalk8s-2.11.7
      Result: True
     Comment: Target was successfully unmounted. Removed target from fstab
     Started: 10:09:59.285132
    Duration: 262.827 ms
     Changes:   
              ----------
              persist:
                  purged
              umount:
                  True
```

salt/metalk8s/archives/unmounted.sls
TeddyAndrieux: Right make sense

scripts/iso-manager.sh
gdemonet: Added [`e2a95a2` (#3730)](https://github.com/scality/metalk8s/pull/3730/commits/e2a95a2aea8bb51a163e83bcf3a9b4c7ef6b4033) to handle this, but read the commit message, the behaviour will not be ideal.

TeddyAndrieux: I'm not sure it's worth, I personally prefer the behavior we have today :confused:

gdemonet: :cry:

gdemonet: OK just removed it then, we'll have to think of a better way to manage this (or just close our eyes until we remove Salt :laughing:)

TeddyAndrieux: Alt Text

salt/metalk8s/archives/unmounted.sls
TeddyAndrieux: Oh, indeed :smile:

salt/metalk8s/archives/unmounted.sls
```diff
@@ -0,0 +1,43 @@
...
+Remove archive mountpoint '{{ archive.path }}':
+  file.absent:
+    - name: {{ archive.path }}
+    - unless:
+      - fun: file.path_exists_glob
+        path: {{ archive.path }}/*
+    - require:
+      - mount: Unmount '{{ archive.iso }}' from '{{ archive.path }}'
+
+Mountpoint '{{ archive.path }}' cannot be removed:
```
TeddyAndrieux: I would prefer a title that does not look like "an error" cause right now even if there is nothing in the directory in the salt output we will have "Mountpoint '{{ archive.path }}' cannot be removed" in the output.
(if `onlyif` is False the state is not executed but you still have it in the output with a comment saying "not executed because of onlyif")

I suggest 
```suggestion
Ensure '{{ archive.path }}' mountpoint is empty
```

scripts/iso-manager.sh
TeddyAndrieux: :eye: :smile:

gdemonet: ooops

salt/metalk8s/archives/unmounted.sls
gdemonet: OK will do that. But to me, it's rather an issue with how the default output shows everything, even when skipped because of requisites.

TeddyAndrieux (approved):

PR #3737, opened 2022-03-29
Allow to setup Loki `resources` from CSC

TeddyAndrieux:
Since Loki has no way to limit the "in-memory" total chunk size in some
cases we may want to put a hard limit on the memory for the Loki pod in
order to avoid Loki to consume all the memory and make the server crash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # CHANGELOG
 ## Release 2.11.5 (in development)
+### Enhancements
 
+- Make Loki pod resources configurable
+  (PR[#3737](https://github.com/scality/metalk8s/pull/3737))
 ## Release 2.11.4
 ### Bug fixes
 
--- a/docs/operation/cluster_and_service_configuration.rst
+++ b/docs/operation/cluster_and_service_configuration.rst
@@ -690,6 +690,36 @@ configuration, see the official `Loki documentation`_.
 
 .. _Loki documentation: https://grafana.com/docs/loki/latest/configuration/
 
+Add Loki memory limit
+"""""""""""""""""""""
+
+Loki consumes some memory to store chunks before they get written to disks.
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # CHANGELOG
 ## Release 2.11.5 (in development)
+### Enhancements
 
+- Make Loki pod resources configurable
```
gdemonet, line 5: Could be good to add a doc section showing how this can be used to limit the memory, if you have time

gdemonet (dismissed):

CHANGELOG.md
TeddyAndrieux: Right, ok

TeddyAndrieux: Done

docs/operation/cluster_and_service_configuration.rst
```diff
@@ -690,6 +690,34 @@ configuration, see the official `Loki documentation`_.
...
+Add Loki memory limit
+"""""""""""""""""""""
+
+Loki consume some memory to store chunk before they get written to disks,
+so his memory consumption really depend on the usage that's why we do not
+set any limit by default.
+
+But if, for whatever reason, Loki is unable to write on the disk, Loki will
+keep getting all logs in memory, to avoid issue where Loki take too much
+memory from the server, you can set a limit to this Pod.
```
gdemonet:
```suggestion
Loki consumes some memory to store chunks before they get written to disks.
Its memory consumption really depends on the usage, which is why we do not
set any limit by default.

However, if Loki is unable to write to the disk for any reason, it will
continue keeping logs in memory, leading to large memory consumption until
the issue is resolved. To prevent Loki from taking too much from the host,
potentially leading to starvation, you can define a resource limit on the
Pod.
```

gdemonet (approved):

PR #3741, opened 2022-03-30
scripts,salt: Do not reconfigure repository in local mode

TeddyAndrieux:
Since the salt state to configure the repository relies on some variable from
pillar available in salt-master to manage the solution images, if we run
the repository state in local mode the solution image will no longer be
available in the registry

To avoid that kind of issue during the upgrade we only run the salt
master state in local mode


gdemonet (approved):
I assume you tested the upgrade with those changes, code LGTM :ok_hand:

TeddyAndrieux:
Indeed I did

PR #3749, opened 2022-04-20
docs: Update entrypoint so that we use the right user to run tox

TeddyAndrieux:
In order to build the documentation, we need to use the right user,
otherwise, git throw an error.
That's why in the build entry point we create a new user with the right IDs
if we need to

### DIFF
--- a/docs/entrypoint.sh
+++ b/docs/entrypoint.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
 
 BUILD_DIR=${BUILD_DIR:-/usr/src/metalk8s/docs/_build}
+TARGET_UID=${TARGET_UID:-$(id -u)}
+TARGET_GID=${TARGET_GID:-$(id -g)}
 
-change_build_ownership() {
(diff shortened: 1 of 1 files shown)


JBWatenbergScality (dismissed):

docs/Dockerfile
```diff
@@ -34,4 +34,4 @@ COPY tox.ini .
 COPY docs/requirements.txt docs/
 
 RUN tox --workdir /tmp/tox --notest -e docs && \
-    rm -rf ~/.cache/pip
+    rm -rf ~/.cache/pip /tmp/tox
```
gdemonet: If you remove the workdir, the whole `RUN` instruction here is pointless (it's used to pre-install the required venv)

gdemonet (changes requested):

docs/Dockerfile
TeddyAndrieux: Hmm ok I see, then I will just change the ownership on the directory

gdemonet (approved):

PR #3751, opened 2022-04-25
Bugfix: retrieve cluster version in node list component so that it is available when deploying 

JBWatenbergScality:
**Component**:

UI

**Context**: 

Deploying a node from the UI was not operating normally.

**Summary**:

Redux state was not initiated with cluster version and the node name was not sent when deploying the node.

### DIFF
--- a/ui/src/containers/NodeList.js
+++ b/ui/src/containers/NodeList.js
@@ -1,6 +1,6 @@
 //@flow
 
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { useHistory } from 'react-router';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


ui/src/containers/NodeList.js
```diff
@@ -1,14 +1,16 @@
...
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { useHistory } from 'react-router';
 import styled from 'styled-components';
 import { Table, Loader } from '@scality/core-ui';
 import { Button } from '@scality/core-ui/dist/next';
 import { padding } from '@scality/core-ui/dist/style/theme';
 import {
+  clearCreateNodeErrorAction,
```
gdemonet: unused import?

ChengYanJin (approved):

PR #3752, opened 2022-04-26
charts,salt: Allow to set Control Plane Ingress IP to an external IP

TeddyAndrieux:
In some context, the IP used to reach the Control Plane Ingress (MetalK8s
UI) is not an IP available on the node, it could be a Load balancer IP
or a NATed IP, so we need to be able to configure OIDC with that kind of
IP

That's why we now use the Control Plane Ingress IP defined in the
bootstrap config and all master Control Plane node IP as `externalIPs`
for the Control Plane Ingress service, so that we can reach the Control
Plane Ingress with any of those IPs (but only the Control Plane Ingress
Ip will be used as redirect IP for OIDC)

### DIFF
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -257,3 +257,25 @@ def get_control_plane_ingress_endpoint():
     return "https://{}:8443".format(
         __salt__["metalk8s_network.get_control_plane_ingress_ip"]()
     )
+
+
+def get_control_plane_ingress_external_ips():
+    """Get all Control Plane Ingress external IPs
+
+    NOTE: Only the first one will be used as redirect IP for oidc
+    """
+    master_nodes = __salt__["metalk8s.minions_by_role"]("master")
+
+    # This function only run on master
+    mine_ret = __salt__["saltutil.runner"](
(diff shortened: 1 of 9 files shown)


salt/_modules/metalk8s_network.py
```diff
@@ -257,3 +257,25 @@ def get_control_plane_ingress_endpoint():
...
+        "mine.get", tgt=",".join(master_nodes), tgt_type="list", fun="control_plane_ip"
+    )
+
+    if not isinstance(mine_ret, dict):
+        raise CommandExecutionError(
+            f"Unable to get master Control Plane IPs: {mine_ret}"
+        )
+
+    return [__salt__["metalk8s_network.get_control_plane_ingress_ip"]()] + sorted(
+        list(mine_ret.values())
```
gdemonet, line 280: FYI: no need to cast this into a list, `sorted` can take any iterable

gdemonet (approved):

salt/_modules/metalk8s_network.py
TeddyAndrieux: Oh right, I added the `sorted` at the end and you cannot concat list and dict values that's why :smile: 
I will merge it as is to not wait another couple of builds

PR #3755, opened 2022-05-04
salt,docs,tests: Allow to change portmap IPs

TeddyAndrieux:
In order to be able to expose the Workload Plane Ingress on another IP
than the Workload Plane IP of each node, we allow setting the portmap
CIDRs from the bootstrap configuration file

### DIFF
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -257,3 +257,22 @@ def get_control_plane_ingress_endpoint():
     return "https://{}:8443".format(
         __salt__["metalk8s_network.get_control_plane_ingress_ip"]()
     )
+
+
+def get_portmap_ips(as_cidr=False):
+    portmap_ips = set()
+    portmap_cidrs = __salt__["pillar.get"]("networks:portmap:cidr")
+
+    if portmap_cidrs:
+        if not isinstance(portmap_cidrs, list):
+            portmap_cidrs = [portmap_cidrs]
+        for cidr in portmap_cidrs:
+            portmap_ips.update(__salt__["network.ip_addrs"](cidr=cidr))
+    else:
+        portmap_ips = {__grains__["metalk8s"]["workload_plane_ip"]}
+
+    portmap_ips.add("127.0.0.1")
+
+    if as_cidr:
+        return [f"{ip}/32" for ip in sorted(portmap_ips)]
+    return sorted(portmap_ips)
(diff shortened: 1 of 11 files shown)


salt/_modules/metalk8s_network.py
```diff
@@ -257,3 +257,24 @@ def get_control_plane_ingress_endpoint():
...
+    portmap_ips = set()
+    portmap_cidrs = __salt__["pillar.get"]("networks:portmap:cidr")
+
+    if portmap_cidrs:
+        if not isinstance(portmap_cidrs, list):
+            portmap_cidrs = [portmap_cidrs]
+        for cidr in portmap_cidrs:
+            portmap_ips = portmap_ips.union(
+                set(__salt__["network.ip_addrs"](cidr=cidr))
+            )
```
gdemonet: Sets are mutable objects, for each logical operation on them (such as `union`, `intersection` or `difference`), there's an `_update` counterpart (in the case of `union`, the counterpart is just `update`, because it's quite easy to understand as is).

```suggestion
            portmap_ips.update(
                __salt__["network.ip_addrs"](cidr=cidr)
            )
```

tests/post/steps/test_ingress.py
```diff
@@ -283,6 +324,36 @@ def server_does_not_respond(host, context):
...
+    server_returns(host=host, context=context, status_code=status_code, reason=reason)
+
+
+@then(
+    parsers.re(
+        r"an (?P<protocol>HTTPS?) request (?:on path '(?P<path>[^']+)' )?"
+        r"on port (?P<port>\d+) on a (?P<plane>.*) IP should not return"
+    ),
+)
+def server_request_returns(host, context, protocol, port, plane, path):
```
gdemonet: Same function name as above, I guess that's OK for pytest-bdd, but it may cause issues with static analysis (if we were to do that :sweat_smile:)

```suggestion
def server_request_does_not_return(host, context, protocol, port, plane, path):
```

gdemonet (commented):
Two minor comments, otherwise LGTM, thank you!

salt/_modules/metalk8s_network.py
TeddyAndrieux: OH well, TIL, there is update for set, great I will change this thank you :+1:

tests/post/steps/test_ingress.py
TeddyAndrieux: good catch just copy/paste

gdemonet (approved):

PR #3772, opened 2022-05-18
build: Do no longer check the nginx image digest

TeddyAndrieux:
The nginx image digest is quite often republished with the same tag (but
a different digest), to avoid breaking the build when it happens, let's
just not check the digest for this specific image


gdemonet (approved):
Sad, but OK

PR #3773, opened 2022-05-18
build: Replace CentOS 7 base images with Rocky Linux

gdemonet:
**Component**: build

**Context**: There are some CVEs (notably [CVE-2018-25032](https://nvd.nist.gov/vuln/detail/CVE-2018-25032)) for which a fix is only available in Red Hat 8 repositories, so we want to take a safer bet by running on a more recent base OS

**Summary**: The base image for `metalk8s-utils` and `salt-master` is changed from `centos:7` to `rockylinux:8.5`.

**Acceptance criteria**: Everything keeps working as before.

### DIFF
--- a/images/metalk8s-utils/Dockerfile
+++ b/images/metalk8s-utils/Dockerfile
@@ -53,7 +53,9 @@ LABEL maintainer="squad-metalk8s@scality.com" \
       com.scality.metalk8s.version="$METALK8S_VERSION"
 
 # Final layers, installing tooling
-RUN dnf install -y epel-release && \
+RUN dnf clean expire-cache && \
+    dnf update -y && \
+    dnf install -y epel-release && \
     dnf install -y \
         bash-completion \
         bind-utils \
--- a/images/metalk8s-utils/Dockerfile
+++ b/images/metalk8s-utils/Dockerfile
@@ -1,6 +1,6 @@
 # SHA256 digest of the base image
 ARG BASE_IMAGE_SHA256
-ARG BASE_IMAGE=docker.io/centos
+ARG BASE_IMAGE=docker.io/rockylinux
 
 FROM $BASE_IMAGE@sha256:$BASE_IMAGE_SHA256
 
@@ -8,6 +8,8 @@ FROM $BASE_IMAGE@sha256:$BASE_IMAGE_SHA256
 ARG KUBERNETES_VERSION
 # Salt version
 ARG SALT_VERSION
+# Etcd version
+ARG ETCD_VERSION
 
 COPY configure-repos.sh /
 RUN /configure-repos.sh $SALT_VERSION && rm /configure-repos.sh
@@ -22,7 +24,7 @@ ARG VERSION
 ARG METALK8S_VERSION
 
 # These contain BUILD_DATE so should come 'late' for layer caching
-LABEL maintainer="moonshot-platform@scality.com" \
+LABEL maintainer="squad-metalk8s@scality.com" \
       # http://label-schema.org/rc1/
       org.label-schema.build-date="$BUILD_DATE" \
       org.label-schema.name="metalk8s-utils" \
@@ -35,7 +37,7 @@ LABEL maintainer="moonshot-platform@scality.com" \
       org.label-schema.schema-version="1.0" \
       # https://github.com/opencontainers/image-spec/blob/master/annotations.md
       org.opencontainers.image.created="$BUILD_DATE" \
-      org.opencontainers.image.authors="moonshot-platform@scality.com" \
+      org.opencontainers.image.authors="squad-metalk8s@scality.com" \
       org.opencontainers.image.url="https://github.com/scality/metalk8s/" \
       org.opencontainers.image.source="https://github.com/scality/metalk8s.git" \
       org.opencontainers.image.version="$VERSION" \
@@ -51,10 +53,9 @@ LABEL maintainer="moonshot-platform@scality.com" \
       com.scality.metalk8s.version="$METALK8S_VERSION"
 
 # Final layers, installing tooling
-RUN yum install -y epel-release && \
-    yum install -y --setopt=skip_missing_names_on_install=False \
+RUN dnf install -y epel-release && \
+    dnf install -y \
         bash-completion \
-        bash-completion-extras \
         bind-utils \
         bzip2 \
         conntrack-tools \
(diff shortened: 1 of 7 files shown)


images/metalk8s-utils/Dockerfile
```diff
@@ -51,32 +53,30 @@ LABEL maintainer="moonshot-platform@scality.com" \
       com.scality.metalk8s.version="$METALK8S_VERSION"
 
 # Final layers, installing tooling
-RUN yum install -y epel-release && \
-    yum install -y --setopt=skip_missing_names_on_install=False \
+RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
```
TeddyAndrieux: Why this change ? You cannot use `dnf install -y epel-release` directly instead ?

images/metalk8s-utils/Dockerfile
```diff
@@ -109,6 +109,17 @@ RUN yum install -y epel-release && \
         wireshark \
         xfsprogs \
         && \
-    yum clean all
+    dnf update -y && \
```
TeddyAndrieux: The update should be done before otherwise, we might update packages we just installed with a "pin" version (like kubectl here for example)

images/metalk8s-utils/configure-repos.sh
```diff
@@ -2,13 +2,12 @@
...
 
-CENTOS_VERSION=$1
-SALT_VERSION=$2
+SALT_VERSION=$1
 
 cat > /etc/yum.repos.d/kubernetes.repo << EOF
 [kubernetes]
 name=Kubernetes
-baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el$CENTOS_VERSION-x86_64
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
```
TeddyAndrieux, line 10: Oh, sad, el8 packages are also available on `kubernetes-el7...` make sense :upside_down_face:

images/salt-master/Dockerfile
```diff
@@ -13,20 +15,21 @@ baseurl=https://repo.saltstack.com/py3/redhat/\$releasever/\$basearch/archive/%s
...
  && pip3 install "git+https://github.com/kubernetes-client/python.git@cef5e9bd10a6d5ca4d9c83da46ccfe2114cdaaf8#egg=kubernetes" \
- && yum remove -y git \
+ && dnf remove -y git \
  && pip3 uninstall -y \
      requests \
      urllib3 \
- && yum install -y salt-master salt-api salt-ssh openssh-clients \
- && yum clean all
+ && dnf install -y salt-master salt-api salt-ssh openssh-clients \
+ && dnf update -y \
```
TeddyAndrieux: ditto, this one should be done at the very beginning

images/salt-master/Dockerfile
```diff
@@ -13,20 +15,21 @@ baseurl=https://repo.saltstack.com/py3/redhat/\$releasever/\$basearch/archive/%s
...
- && yum install -y python3-pip \
+ && rpm --import https://repo.saltstack.com/py3/redhat/8/x86_64/archive/${SALT_VERSION}/SALTSTACK-GPG-KEY.pub \
+ && dnf clean expire-cache \
+ && dnf install -y epel-release \
+ && dnf install -y python3-pip \
  && pip3 install pip==20.1 \
  && pip3 install "etcd3 != 0.11.0" \
- && yum install -y git \
+ && dnf install -y git \
  && pip3 install "git+https://github.com/kubernetes-client/python.git@cef5e9bd10a6d5ca4d9c83da46ccfe2114cdaaf8#egg=kubernetes" \
```
TeddyAndrieux, line 26: (not necessarily part of this PR, but I think we can now move to a "real" version of python-kubernetes since we do not really rely on it)

TeddyAndrieux (commented):
Overall LGTM, we will see if it pass the CI

Wondering if there is a reason to pick RockyLinux instead of Ubi for example ? (just asking I have no strong opinion on this choice)

images/metalk8s-utils/Dockerfile
gdemonet: Ah, that's a remnant from when I tried on UBI before realizing it was missing a whole bunch of other packages. Will change

images/metalk8s-utils/Dockerfile
gdemonet: Right, makes sense

images/metalk8s-utils/configure-repos.sh
gdemonet: Yeah... :shrug:

images/salt-master/Dockerfile
gdemonet: Will open a ticket then!

gdemonet:
> Wondering if there is a reason to pick RockyLinux instead of Ubi for example ? (just asking I have no strong opinion on this choice)

For `metalk8s-utils`, UBI was really missing too many packages (it doesn't have most core repos), so it was going to be a pain to download all the missing ones from various places, Rocky made more sense to me.
For `salt-master`, well, since we already have one base image in the ISO, I decided to use the same one to avoid increasing the storage consumption.

salt/metalk8s/map.jinja
```diff
@@ -36,6 +36,14 @@
...
       },
     },
+    'default': {
+      'default': {
+        'kubernetes': versions.kubernetes,
+        'repo': {
+          'images': versions.images,
+        },
+      },
+    },
```
TeddyAndrieux: Instead of doing this, cannot we set the grains directly in the container, so that it has the same behavior as what we do for salt-minion

Basically, just create the `/etc/salt/grains` file like that 
```yaml
os_family: RedHat
os: CentOS
```
(And add a comment to make sure we remove it once we move to salt 3004)

gdemonet: I'll give it a try

images/salt-master/Dockerfile
```diff
@@ -1,32 +1,38 @@
...
- && yum install -y epel-release \
- && yum install -y python3-pip \
+ && rpm --import https://repo.saltstack.com/py3/redhat/8/x86_64/archive/${SALT_VERSION}/SALTSTACK-GPG-KEY.pub \
+ && dnf clean expire-cache \
+ && dnf update -y \
+ && dnf install -y glibc-all-langpacks langpacks-en \
+ && dnf install -y epel-release \
+ && dnf install -y python3-pip \
  && pip3 install pip==20.1 \
  && pip3 install "etcd3 != 0.11.0" \
```
TeddyAndrieux: nitpick: It's not consistent :cry: 
```suggestion
 && pip3 install "pip == 20.1" \
 && pip3 install "etcd3 != 0.11.0" \
```

TeddyAndrieux (approved):
:shipit:

PR #3785, opened 2022-06-02
tests: Use a specific cache file for every k8s_client

TeddyAndrieux:
Since the k8s_client cached information may change between runs, for
For example when we run a test before the upgrade and after, we need to reload
the cache


gdemonet (approved):
Didn't know that was a problem, did you see some side-effects that could have been explained by this cache?

PR #3793, opened 2022-06-14
salt,build,packages: No longer use/build `calico-cni-plugin` RPM

TeddyAndrieux:
Instead of relying on Salt to install the Calico CNI binaries and manage
the CNI config file, we use the calico CNI image with some host path
mounted to drop those binaries and write the CNI config file.

NOTE: This is how Calico upstream is used to manage those CNI binaries
and config


CHANGELOG.md
```diff
@@ -24,6 +24,10 @@
   the `metalk8s-utils` container
   (PR[#3773](https://github.com/scality/metalk8s/pull/3773))
 
+- Remove the `calico-cni-plugin` RPM packages and instead relies on
```
```suggestion
- Remove the `calico-cni-plugin` RPM package and rely instead on
```

salt/metalk8s/kubernetes/cni/calico/deployed.sls
```diff
@@ -4321,75 +4323,72 @@ spec:
         # This container performs upgrade from host-local IPAM to calico-ipam.
         # It can be deleted if this is a fresh installation, or if you have already
         # upgraded to use calico-ipam.
```
gdemonet, line 4325: This comment seems to indicate that's only needed in a context where we had host-local IPAM configured, but that's never been the case for us, right? Can't we just remove it?

salt/metalk8s/kubernetes/cni/calico/deployed.sls
```diff
@@ -4474,15 +4470,8 @@ spec:
...
               value: "false"
-            # Set Felix logging to "info"
-            - name: FELIX_LOGSEVERITYSCREEN
-              value: "info"
             - name: FELIX_HEALTHENABLED
               value: "true"
-            # Note: We do not want to report about outgoing connections
-            #       in Metalk8s
-            - name: FELIX_USAGEREPORTINGENABLED
-              value: "false"
```
gdemonet: I think we still want this to be disabled, MetalK8s should not attempt to send information to the outside.

salt/tests/unit/modules/test_cri.py
```diff
@@ -276,3 +286,34 @@ def test_ready(self, retcode, result):
...
+
+        salt_dict = {"cmd.run_all": MagicMock(side_effect=cmd_run_mock)}
+
+        with patch.dict(cri.__salt__, salt_dict):
+            if raises:
+                self.assertRaisesRegex(
+                    CommandExecutionError,
+                    result,
+                    cri.stop_pod,
+                    {"my.label": "ABCD"},
```
gdemonet: Since we can't really test the effect of passing labels to the command, can we at least make sure that the `cmd.run_all` mock receives the expected command lines?

Something like this for `crictl pods`:

```python
# after the call was made
self.assertEqual(
    cmd_run_mock.call_args_list[0][0],
    "crictl pods --quiet --selector=my.label=ABCD"
)
```

Could also add something for `crictl stopp` (comparing with the IDs returned by `crictl pods`), but not necessary I guess.

salt/metalk8s/kubernetes/cni/calico/post-upgrade.sls
```diff
@@ -0,0 +1,21 @@
...
+          k8s-app: calico-node
+    - onchanges:
+      - pkg: Ensure calico-cni-plugin is removed
+
+Make sure calico-node is up and ready:
+  test.configurable_test_state:
+    - changes: False
+    - result: __slot__:salt:cri.wait_container(name="calico-node", state="running")
+    - require:
+      - module: Restart calico-node
```
gdemonet, line 21: I'm curious, do we really have to do this node-by-node? Couldn't we "just" apply `pkg.removed calico-cni-plugin` to all our nodes, and only then issue a restart on the DaemonSet?

gdemonet (changes requested):

salt/metalk8s/kubernetes/cni/calico/deployed.sls
TeddyAndrieux: Today, yes we could but maybe in the future this `calico-ipam -upgrade` command will do something "useful", I'm not sure it's worth removing it, it just add some "manual patch" to do on the calico upstream manifest

salt/metalk8s/kubernetes/cni/calico/deployed.sls
TeddyAndrieux: Right, I will add it back

salt/metalk8s/kubernetes/cni/calico/post-upgrade.sls
TeddyAndrieux: It's not "needed" but this way, since we have "require" in the orchestrate, we will stop the process if one calico node is not working after the package deletion

salt/tests/unit/modules/test_cri.py
TeddyAndrieux: Right I can add it

salt/_modules/cri.py
```diff
@@ -219,3 +219,35 @@ def ready(timeout=10):
...
+    selector = ",".join([f"{key}={value}" for key, value in labels.items()])
+
+    pod_ids_out = __salt__["cmd.run_all"](f"crictl pods --quiet --label={selector}")
+    if pod_ids_out["retcode"] != 0:
+        raise CommandExecutionError(
+            f"Unable to get pods with labels {selector}:\n"
+            f"STDERR: {pod_ids_out['stderr']}\nSTDOUT: {pod_ids_out['stdout']}"
+        )
+
+    pod_ids = pod_ids_out["stdout"]
```
gdemonet, line 241: Just realized that this is multiline, we'll have issues when passing this to `cmd.run_all`, no?

I just tested it manually:

```
# salt-call cmd.run_all "crictl stopp 9f93d9604d07d865ad74642a5a3ed2ce87fd605f5791e9b7ffbecad300219992
> a92c256f8baa99e57c65c9b730825b7a904d060cd88a0487d466a9e07f2eb1bc"
[ERROR   ] Command 'crictl' failed with return code: 127
[ERROR   ] stdout: Stopped sandbox 9f93d9604d07d865ad74642a5a3ed2ce87fd605f5791e9b7ffbecad300219992
[ERROR   ] stderr: /bin/bash: line 1: a92c256f8baa99e57c65c9b730825b7a904d060cd88a0487d466a9e07f2eb1bc: command not found
[ERROR   ] retcode: 127
local:
    ----------
    pid:
        27435
    retcode:
        127
    stderr:
        /bin/bash: line 1: a92c256f8baa99e57c65c9b730825b7a904d060cd88a0487d466a9e07f2eb1bc: command not found
    stdout:
        Stopped sandbox 9f93d9604d07d865ad74642a5a3ed2ce87fd605f5791e9b7ffbecad300219992
```

TeddyAndrieux: No, it works well, you do not have the bash interpretation here
```console
# salt-call cri.stop_pod labels="{'app.kubernetes.io/instance': 'ingress-nginx'}"
local:
    Stopped sandbox c99388e45b37ff00d76e19bc4a9bec27fbf717dfa79f99a304e40221d88d8f19
    Stopped sandbox 2727aef315f60006937d249ff47534f30e669c5746816e07244687ae8dadc7e8
```

If you want to do the same with bash it will more look like this (that work as well)
```console
# crictl stopp 8fca1769df6dcd8a29c65854ac60b95c104d9673b00c95f3023c0e0a6498907f \
> fdd31265ec3b3e9ad8e0b3f5c91a6fe4fdc82b196364c76b14138f2e1bb87828
Stopped sandbox 8fca1769df6dcd8a29c65854ac60b95c104d9673b00c95f3023c0e0a6498907f
Stopped sandbox fdd31265ec3b3e9ad8e0b3f5c91a6fe4fdc82b196364c76b14138f2e1bb87828
# salt-call cmd.run_all "crictl stopp 2be2f438ed2b16c1588cbe8e0c34ecb55cc23813af919cb71028b6b44c2bba19 \
> 06969ef0112cd2c8d8f3a854a5938ec8715a89604586f9ca8ff2e3b4ebc9e400"
local:
    ----------
    pid:
        13339
    retcode:
        0
    stderr:
    stdout:
        Stopped sandbox 2be2f438ed2b16c1588cbe8e0c34ecb55cc23813af919cb71028b6b44c2bba19
        Stopped sandbox 06969ef0112cd2c8d8f3a854a5938ec8715a89604586f9ca8ff2e3b4ebc9e400
```

gdemonet: OK, I'm surprised it works but I trust you :+1: (it's not really a Bash interpretation problem IMO, in your example you're escaping the newline while I explicitly send it to `salt-call`, which should be roughly the same as doing `f"crictl stopp {pod_ids_out['stdout']}"`, unless I misunderstood something.. maybe it's just some other Salt magic :shrug: not worth investigating though, if it works then it's fine for me)

gdemonet (approved):

PR #3804, opened 2022-07-01
Bump core UI to v0.39.2

ognyan-kostadinov:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


ChengYanJin (dismissed):
LGTM!

shell-ui/src/initFederation/UIListProvider.js
```diff
@@ -1,16 +1,15 @@
 //@flow
-import { createContext, useContext, type Node, useCallback } from 'react';
+import { createContext, useContext, type Node } from 'react';
 import { useQuery } from 'react-query';
-import Loader from '@scality/core-ui/dist/components/loader/Loader.component';
-import ErrorPage500 from '@scality/core-ui/dist/components/error-pages/ErrorPage500.component';
+import { Loader, ErrorPage500 } from '@scality/core-ui';
```
JBWatenbergScality: To minimise shell ui bundle size maybe we should stick to only importing loader and erropage components. 
E,g zenko ui is not using vega but if we change the imports here the bundle will contain vega which is a very large library so as long as core ui can't be tree shaked I would advise using direct imports in shell ui

JBWatenbergScality (approved):

PR #3818, opened 2022-07-08
Wait for apiserver to be updated or restart kubelet on timeout

gdemonet:
This will ensure that if kubelet gets stuck and doesn't restart k-a after an update of its manifest, we can detect the situation and restart kubelet.

### DIFF
--- a/salt/_modules/cri.py
+++ b/salt/_modules/cri.py
@@ -2,8 +2,8 @@
 Various functions to interact with a CRI daemon (through :program:`crictl`).
 """
 
-import re
 import logging
+import re
 import time
 
 from salt.exceptions import CommandExecutionError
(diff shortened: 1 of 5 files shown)


salt/_modules/cri.py
```diff
@@ -229,25 +229,129 @@ def stop_pod(labels):
...
+        Whether to raise if the timeout period is exceeded (otherwise, return False)
+    """
+    start_time = time.time()
+
+    while time.time() - start_time < timeout:
+        current_id = get_pod_id(name=name, state=state, ignore_not_found=True)
+        if current_id and current_id != last_id:
+            return True
+        remaining = timeout + start_time - time.time()
+        if remaining < sleep:  # Don't sleep if we know it's going to time out
```
TeddyAndrieux, line 347: Habile bill

salt/_modules/cri.py
```diff
@@ -229,25 +229,129 @@ def stop_pod(labels):
...
+            break
+        time.sleep(sleep)
+
+    if raise_on_timeout:
+        verb = "updated" if last_id else "created"
         raise CommandExecutionError(
-            f"Unable to stop pods with labels {selector}:\n"
-            f"IDS: {pod_ids}\nSTDERR: {out['stderr']}\nSTDOUT: {out['stdout']}"
+            f"Pod {name} was not {verb} after {(time.time() - start_time):.0f} seconds"
```
TeddyAndrieux, line 354: Any reason to not use timeout variable directly ?

TeddyAndrieux (approved):
LGTM

salt/_modules/cri.py
gdemonet: We may exit earlier in case `timeout` is not a multiple of `sleep` :innocent:

PR #3822, opened 2022-07-13
operator: Init MetalK8s Operator

TeddyAndrieux:
For reviewer: You should review commits one by one as some of the commits is just "some code" generated by `operator-sdk`

---

- Init operator and ClusterConfig CRD :heavy_check_mark: 
- Operator image build :heavy_check_mark:
- Operator manifests builds :heavy_check_mark: 
- Operator go linting in CI :heavy_check_mark: 
- Operator go codegen in doit :heavy_check_mark: 
- Operator go test in CI :heavy_check_mark: 
- Operator deployment :heavy_check_mark:
- Create a default ClusterConfig object at operator startup (and delete any other ClusterConfig objects) :heavy_check_mark:
- Add a test to ensure the default ClusterConfig get recreated and "extra" ClusterConfig get deleted :heavy_check_mark: 
- Add changelog entry :heavy_check_mark: 

NOTE: We do not add any Webhook to ensure that you cannot create/delete ClusterConfig objects in this PR as it needs certificates and we may want to use Cert-manager to manage those certs, so it will be done in another PR


operator/PROJECT
```diff
@@ -0,0 +1,9 @@
+domain: metalk8s.scality.com
+layout:
+- go.kubebuilder.io/v3
+plugins:
+  manifests.sdk.operatorframework.io/v2: {}
+  scorecard.sdk.operatorframework.io/v2: {}
+projectName: operator
+repo: github.com/metalk8s/operator
```
JBWatenbergScality: nit: Shall not we simply use `github.com/scality/metalk8s` instead of `github.com/metalk8s/operator` as we don't really own the repo `github.com/metalk8s/operator` and I think ideally this info should match the repo holding the source code of the operator as a best practice (some tooling for doc may rely on this info to generate links or so...)

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
 
 	// TODO(user): your logic here
 
-	return ctrl.Result{}, nil
+	return endReconciliation(nil)
+}
+
+// Trigger a reschedule after a short delay.
+func delayedRequeue(err error) (ctrl.Result, error) {
+	delay := 10 * time.Second
```
JBWatenbergScality, line 101: nit: shall we instead of a fixed retry time use an exponential backoff approach to avoid overload the API server with failures when something wrong is already happening ?

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
+	return ctrl.Result{Requeue: err == nil, RequeueAfter: delay}, err
+}
+
+// Trigger a reschedule as soon as possible.
+func requeue(err error) (ctrl.Result, error) {
+	return ctrl.Result{Requeue: err == nil}, err
+}
```
JBWatenbergScality: nit: maybe those methods should try to update the status of the CR to notify about a failure ?

tests/post/features/operator.feature
```diff
@@ -0,0 +1,13 @@
...
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=metalk8s-operator' are 'Ready'
+        When we create an extra 'test-cc' ClusterConfig
+        Then the 'test-cc' ClusterConfig get deleted
+
+    Scenario: Deletion of the default ClusterConfig
+        Given the Kubernetes API is available
+        And pods with label 'app.kubernetes.io/name=metalk8s-operator' are 'Ready'
+        When we delete the 'default' ClusterConfig
+        Then the 'default' ClusterConfig get automatically recreated
```
JBWatenbergScality: nit: With this approach we are constraining the use of Metalk8s operator only with Metalk8s (which is likely fine for now) however I think we might want to use it to install adds-on easily on other K8s distributions, to be discussed. 
The test is plenty valid for Metalk8s context through, but I let the comment here as a base to discuss.

operator/PROJECT
TeddyAndrieux: Hmm, I'm not sure, cause we have multiple operators in the same repo and the "root" for the operator is this directory. And to me this repo is the "URL" to the operator go I think, no ?

I don't know about the best practice and not have any strong opinion here, just today we used the same approach for the storage-operator (`repo: github.com/scality/metalk8s/storage-operator`)
Any opinion on this @gdemonet ?

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Yes could make sense, for "usual" operator, I agree, but here we talk about an operator that can handle only one single object so ... I'm not sure it bring a lot of value here

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Yes, you're right maybe if we have an error we want to put it in the object status :thinking:

tests/post/features/operator.feature
TeddyAndrieux: I'm not sure to get your point here.

I agree that such operator could be used outside of "metalk8s core" at some point, but I don't get why this approach could be an issue for this ?

NOTE: This test will be updated at some point when we will have a proper Webhook to "forbib" object creation and deletion

operator/PROJECT
gdemonet: Yes I think we should use the "real" URL, so none of `github.com/scality/metalk8s` or `github.com/metalk8s/operator` are valid :innocent: 

```suggestion
repo: github.com/scality/metalk8s/operator
```

operator/controllers/clusterconfig_controller.go
gdemonet: This shouldn't be part of the "requeue" methods. Deciding whether to show an error in the managed object is part of the reconciliation logic of a specific controller, requeuing is just sending information to the controller manager.

tests/post/features/operator.feature
gdemonet: Are you suggesting we shouldn't use a `ClusterConfig` object when running in other K8s distributions @JBWatenbergScality? It's debatable, and adding support for this could be done later IMO.

operator/PROJECT
TeddyAndrieux: Oh ! Good catch indeed I forgot the `scality`, but I don't see why `github.com/scality/metalk8s` is not valid

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: No strong opinion here, it could be an option but indeed we may have some cases where we just want to requeue and not update the object

operator/PROJECT
TeddyAndrieux: Fixed

operator/api/v1alpha1/clusterconfig_types.go
```diff
@@ -23,13 +23,17 @@ import (
...
 // ClusterConfigSpec defines the desired state of ClusterConfig
 type ClusterConfigSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Foo is an example field of ClusterConfig. Edit clusterconfig_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// Information about the Workload Plane.
+	// +optional
+	WorkloadPlane WorkloadPlaneSpec `json:"workloadPlane,omitempty"`
```
gdemonet, line 36: :thinking:

We have received requests for handling more than a single "workload plane network", I'm wondering if the assumption we took about the network layout (1 control plane + 1 workload plane) wasn't a bit too restrictive... I guess we'd still need to special-case the control plane, but I think we'll end up changing this assumption to something more generic (i.e. managing an arbitrary number of networks, for which we'll have ingress controllers, and out of which a single one can bear the control plane services).

It's likely too soon to jump on this, but I bet we'll have to rework the CR structure to handle this differently.

buildchain/buildchain/image.py
```diff
@@ -280,6 +280,19 @@ def _local_image(name: str, **kwargs: Any) -> targets.LocalImage:
...
+        name="metalk8s-operator",
+        dockerfile=constants.METALK8S_OPERATOR_ROOT / "Dockerfile",
+        build_context=constants.METALK8S_OPERATOR_ROOT,
+        build_args={
+            "BUILD_DATE": datetime.datetime.now(datetime.timezone.utc)
+            .astimezone()
+            .isoformat(),
+            "VCS_REF": constants.GIT_REF or "<unknown>",
+            "METALK8S_VERSION": versions.VERSION,
+            "VERSION": "latest",
```
gdemonet: Why are we doing this? Couldn't we just tag our operators with the project version? (I see we actually use the version in the image tag, but this label should reflect the same value I think)

operator/deploy/manifests.yaml
```diff
@@ -0,0 +1,227 @@
...
+  name: metalk8s-operator-manager-role
+subjects:
+- kind: ServiceAccount
+  name: metalk8s-operator-controller-manager
+  namespace: kube-system
+---
+apiVersion: v1
+data:
+  controller_manager_config.yaml: "apiVersion: controller-runtime.sigs.k8s.io/v1alpha1\nkind: ControllerManagerConfig\nhealth:\n  healthProbeBindAddress: :8081\nmetrics:\n  bindAddress: 127.0.0.1:8080\nwebhook:\n  port: 9443\nleaderElection:\n  leaderElect: true\n  resourceName: 02367e3e.metalk8s.scality.com\n#   leaderElectionReleaseOnCancel defines if the leader should step down volume \n#   when the Manager ends. This requires the binary to immediately end when the\n#   Manager is stopped, otherwise, this setting is unsafe. Setting this significantly\n#   speeds up voluntary leader transitions as the new leader don't have to wait\n#   LeaseDuration time first.\n#   In the default scaffold provided, the program ends immediately after \n#   the manager stops, so would be fine to enable this option. However, \n#   if you are doing or is intended to do any operation such as perform cleanups \n#   after the manager stops then its usage might be unsafe.\n#   leaderElectionReleaseOnCancel: true\n"
```
gdemonet, line 159: Not super readable :disappointed:

operator/config/metalk8s/deploy_patch.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: not-important
```
gdemonet, line 4: Do we still need it if it's not relevant?

operator/controllers/clusterconfig_controller.go
```diff
@@ -33,6 +37,13 @@ type ClusterConfigReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+// ClusterConfig name to manage, since we only support one ClusterConfig
+// the object is created by the operator and it's the only one that will be managed
+var (
+	log          = logf.Log.WithName("clusterconfig-controller")
+	instanceName = "default"
```
gdemonet: (nit) Not sure about the name, I'd rather use "main" :thinking: WDYT? Also, isn't it a `const` rather than a `var`?

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
 	// TODO(user): your logic here
 
-	return ctrl.Result{}, nil
+	return endReconciliation(nil)
+}
+
+// Trigger a reschedule after a short delay.
+func delayedRequeue(err error) (ctrl.Result, error) {
+	delay := 10 * time.Second
+	return ctrl.Result{Requeue: err == nil, RequeueAfter: delay}, err
```
gdemonet: Just for my own understanding, if `err != nil`, does it mean we will still `RequeueAfter`?

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
+	}
+
+	if instance.Name != instanceName {
+		if err := r.Delete(ctx, instance); err != nil {
+			reqLogger.Error(
+				err, "cannot delete extra ClusterConfig: requeue",
+			)
+			return delayedRequeue(err)
+		}
+		reqLogger.Info("deleting extra ClusterConfig, consider updating the default one")
```
gdemonet: Would be better to print the actual `instanceName` in this message I believe

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	client := mgr.GetClient()
+
+	instance := &metalk8sscalitycomv1alpha1.ClusterConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: instanceName},
+		Spec:       metalk8sscalitycomv1alpha1.ClusterConfigSpec{},
+	}
+
+	err := client.Create(context.Background(), instance)
```
gdemonet: Is it safe to use such a context when attempting to interact with K8s API? I guess we'd end up getting a timeout from k-a or the network layer if something is really wrong?

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
+func endReconciliation(err error) (ctrl.Result, error) {
+	return ctrl.Result{}, err
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	client := mgr.GetClient()
+
+	instance := &metalk8sscalitycomv1alpha1.ClusterConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: instanceName},
```
gdemonet: Could be worth adding a finalizer, just to prevent mistakes (which could be quite troublesome if there were a lot of customizations in the object)

operator/controllers/clusterconfig_controller.go
```diff
@@ -47,15 +58,72 @@ type ClusterConfigReconciler struct {
...
+	defer reqLogger.Info("reconciling ClusterConfig: STOP")
+
+	instance := &metalk8sscalitycomv1alpha1.ClusterConfig{}
+	err := r.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			if req.Name == instanceName {
+				panic(fmt.Errorf(
+					"%s ClusterConfig object should not be deleted", req.Name,
+				))
```
gdemonet, line 74: Worth adding a comment about the operator behavior here (either directly in the message, or at least in the code): it's not obvious that the panic will lead to a restart (because, K8s) followed by a re-creation of the main ClusterConfig on startup.

operator/.gitignore
```diff
@@ -0,0 +1,25 @@
...
+
+# Test binary, build with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Kubernetes Generated files - skip generated files, except for vendored files
+
+!vendor/**/zz_generated.*
```
gdemonet, line 19: ? I don't understand, this line should imply we ignore generated files, but they are still part of some following commits :thinking:

operator/api/v1alpha1/clusterconfig_types.go
TeddyAndrieux: I'm not sure, to me we will keep 1 network for Control Plane and 1 network for Workload Plane, but we may expose Ingresses on other networks/IPs.

Why we would need to change the CR layout, to me with the same layout as Bootstrap config we are fine

buildchain/buildchain/image.py
TeddyAndrieux: TBH I don't really know, I did the same as for the storage operator, I think you should ask it to @gdemonet who did it for storage operator https://github.com/scality/metalk8s/commit/fdeae9674719c683e4955de7f081da6087fc32bc#diff-cdf9db2f127c381762a567b95b0e86870b5ef70f25d5e38750d5294bc3597044 :stuck_out_tongue_winking_eye:

operator/deploy/manifests.yaml
TeddyAndrieux: It's auto-generated, but yes indeed

operator/config/metalk8s/deploy_patch.yaml
TeddyAndrieux: Unfortunately yes, `name` is mandatory but not used

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Right, I will change it

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: I did not tested this, this function comes from the storage-operator, from the documentation if `Requeue` is False he will not be requeued so I think no but I will do a test

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Hmm yes indeed, I will fix it

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: This function is only called at operator startup, but yes we might be able to use another context, I will see

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: I'm not sure.

We could add a finalizer that we never remove but the fact is AFAIK we cannot "cancel" a deletion, so even if we have a finalizer, if the user deletes the object, it will be marked as deleted and will stay in this state

So ... I don't know, and to me, I see this as, a user deleting the ClusterConfig object so we basically "reset" to the default config (today the default config is empty but maybe at some point we will have some "good" default), make sense no ?

And at some point, we will have webhook to forbid it

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Ok I will add a comment

operator/.gitignore
TeddyAndrieux: This gitignore come from operator-sdk.

To me it's the opposite `!`, we do not ignore the `zz_generated.*` files.

I think it's either a leftover in operator-sdk because they usually git ignore the `vendor` directory, either just here to ensure we do not git ignore those specific files

buildchain/buildchain/image.py
gdemonet: :trollface:

gdemonet: OK, let's not change storage-operator right now, but can you change this line to:

```suggestion
            "VERSION": versions.VERSION,
```

please?

operator/.gitignore
gdemonet: Right, that's strange, the comment is quite misleading.

operator/controllers/clusterconfig_controller.go
gdemonet: I think we can cancel a deletion by removing `metadata.deletionTimestamp` and `metadata.deletionGracePeriodSeconds`, but yeah OK, since we will add the webhook I guess that's fine. And since we don't want to handle deletion, adding a finalizer doesn't make much sense.

For "resetting to defaults", I think updating the object to have `spec: null` could be enough, no need to delete.

operator/api/v1alpha1/clusterconfig_types.go
gdemonet: My point is that the assumptions about what services should use which network could be wrong (e.g. don't use the public-facing workload plane for the overlay network, or don't expose the control-plane-ingress-controller on the control plane network), and this will likely become easier to handle if we don't encode such assumptions in the CR structure.

The fact that BootstrapConfiguration also has these assumptions encoded would also become a problem, IMO.

Anyway, let's not jump to conclusions on this topic, we'll try to think about it more in the future. And changes to the CR schema are possible, so not much to worry about.

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: AFAIK, we cannot remove those fields.

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: I will let it like that for the moment and we will see when we will add more logic to the operator if we want to update the status or not

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: So I will let it like that for the moment and we will see later if we may do something a bit smarter to compute the delay

buildchain/buildchain/image.py
TeddyAndrieux: Sure I will do it

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: So I did some tests and research on the internet, and it appears that if there is an error in "Reconcile" return, it will be automatically requeued no matter the "Result" content AND the Controller automatically handle the retry backoff.

So I will update a bit the functions

buildchain/buildchain/image.py
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: Done

operator/controllers/clusterconfig_controller.go
TeddyAndrieux: I did some research and the controller-runtime is automatically implementing a retry back-off for us

tests/post/features/operator.feature
JBWatenbergScality: I think we can keep it like this for now. 
For the long term goal I was thinking about using the same ClusterConfig but without the "core" part, hence in such cases we should be able to delete the resource. But anyway let's start with this implementation and iterate over it later on

gdemonet (dismissed):
LGTM :ok_hand:

gdemonet (approved):

PR #3824, opened 2022-07-18
eve: Increase a bit the timeout for Multi node tests

TeddyAndrieux:
Some tests may take a bit of time (like Ingress tests), bump a bit the
timeout


gdemonet (approved):
Right, saw that but forgot to push a PR. Thanks!

PR #3838, opened 2022-07-29
ui: Fixed the inverted data series after switching between multiple nodes

ognyan-kostadinov:
**Component**:
NodeListTable
platformlibrary/metrics.js
hooks.js

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/src/hooks.js
+++ b/ui/src/hooks.js
@@ -190,7 +190,18 @@ export const useChartSeries = ({
   );
 
   const isLoading = queries.find((query) => query.isLoading);
-  const queriesData = queries.map((query) => query.data);
+  const queriesData = queries
+    .map((query) => {
+      return query.data;
+    })
+    /* useQueries is running the requests in paralel and given that
+     * in transformPrometheusDataToSeries (which is a generic function used by multiple charts)
+     * we make an assumption on the order of responses
+     * then we need to make sure that the average query is the second one in the array
+     * That is achieved by giving a key param to the response object (e.g. 'cpuUsage' and 'cpuUsageAvg')
+     * and sorting the array alphanumerically on its 'key' property
+     */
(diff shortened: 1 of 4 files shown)


ChengYanJin (approved):

ui/src/hooks.js
```diff
@@ -190,7 +190,18 @@ export const useChartSeries = ({
...
+      return query.data;
+    })
+    /* useQueries is running the requests in paralel and given that
+     * in transformPrometheusDataToSeries (which is a generic function used by multiple charts)
+     * we make an assumption on the order of responses
+     * then we need to make sure that the average query is the second one in the array
+     * That is achieved by giving a key param to the response object (e.g. 'cpuUsage' and 'cpuUsageAvg')
+     * and sorting the array alphanumerically on its 'key' property
+     */
+    .sort((query1, query2) => (query1.key > query2.key ? 1 : -1));
```
JBWatenbergScality, line 204: Instead of introducing this _hacky_ way of fixing a non ideal logic implementation (transformPrometheusDataToSeries being based on array order) I would suggest trying to see if we can migrate the data structure to a well defined object (example: `{nodes: {<NODE_NAME>: <node_data>}, avg: <avg_data>}`) This approach would benefit a lot to readability and code maintainability IMO.

JBWatenbergScality: Then we could wrap data structures transformation into a custom hook `useChartQueries` or something like that takes queries as args and prepare all the data for the charts to go further.

ognyan-kostadinov: Yes, it would've been great given that `useQueries` returned a keyed object (based on queryKey or whatever) or if they labeled in any way which response corresponds to which request. It is possible that they intended to have the order of array elements to be the same as the ones of requests (because it's always the case on first load), but apparently on the second one they flip

PR #3841, opened 2022-08-01
ui: Make sure that K8S client is reinitialised when the access token is renewed

JBWatenbergScality:
**Component**:

ui

**Context**: 

K8S client may not be using the latest available token and hence the UI may display authorisation errors when K8S client is using an expired token to retrieve informations.

**Summary**:

In this PR I make sure that anytime the token change we reinit K8S client with the newly negotiated one.


ChengYanJin (approved):
LGTM!

PR #3848, opened 2022-08-08
salt,docs: Allow to change MetalK8s UI replicas and affinity

TeddyAndrieux:
We also bump the default number of replicas for MetalK8s UI to 2

### DIFF
--- a/salt/metalk8s/addons/ui/config/metalk8s-ui-config.yaml.j2
+++ b/salt/metalk8s/addons/ui/config/metalk8s-ui-config.yaml.j2
@@ -4,7 +4,20 @@
 apiVersion: addons.metalk8s.scality.com/v1alpha2
 kind: UIConfig
 spec:
+  # Deployment configuration
+  deployment:
+    replicas: 2
+    affinity:
+      podAntiAffinity:
+        soft:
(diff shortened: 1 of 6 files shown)


salt/metalk8s/addons/ui/config/metalk8s-ui-config.yaml.j2
```diff
@@ -4,7 +4,20 @@
...
+  deployment:
+    replicas: 2
+    affinity:
+      podAntiAffinity:
+        soft:
+          - topologyKey: kubernetes.io/hostname
+        #   - topologyKey: my.second.important/topologyKey
+        #     weight: 42
+        # hard:
+        #   - topologyKey: kubernetes.io/hostname
```
JBWatenbergScality, line 17: nit: why do we keep those commented lines ?

JBWatenbergScality (approved):

salt/metalk8s/addons/ui/config/metalk8s-ui-config.yaml.j2
TeddyAndrieux: It's because we take this file in the documentation so that from the documentation the user can see those "example"

PR #3851, opened 2022-08-09
operator: Do not override kube-system namespace

TeddyAndrieux:
Since we deploy the operator on the kube-system namespace we do not need
to create this one, especially since it removes the annotations we set on
this namespace

### DIFF
--- a/operator/config/metalk8s/delete_ns.yaml
+++ b/operator/config/metalk8s/delete_ns.yaml
@@ -1,0 +1,5 @@
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
--- a/operator/config/metalk8s/kustomization.yaml
+++ b/operator/config/metalk8s/kustomization.yaml
@@ -16,3 +16,6 @@ patches:
 - path: deploy_patch.yaml
   target:
     kind: Deployment
+
+patchesStrategicMerge:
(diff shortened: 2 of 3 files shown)


operator/config/metalk8s/delete_ns.yaml
```diff
@@ -0,0 +1,5 @@
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
```
gdemonet, line 5: From your commit message, I understand why we don't want to apply the `kube-system` *Namespace* manifest, but what about this one? Did we create it before? Or is it just part of the "default config" which we include and translated to `kube-system` somehow?

gdemonet (approved):
Approving, the comment is just a question :ok_hand:

operator/config/metalk8s/delete_ns.yaml
TeddyAndrieux: > is it just part of the "default config" which we include and translated to kube-system somehow?

Yes, by default with operator SDK we have this manifest 
```yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    control-plane: controller-manager
  name: system
```

That we update with kustomize, but to remove it we need to refer to it using the "original" name

PR #3856, opened 2022-08-19
operator: Add Keepalived Controller

TeddyAndrieux:
- Update ClusterConfig CRD
- Add some utils to manage Objects
- Add Go variable in operator with MetalK8s version
- Add skeleton for "sub reconciler"
- Handle MetalK8s VIP Namespace creation/update/deletion
- Handle MetalK8s VIP High-Level ConfigMap creation/update/deletion
- Handle MetalK8s VIP DaemonSet creation/update/deletion
- Add (and update) some status on ClusterConfig CR
- Add some events on ClusterConfig CR

NOTE: For reviewers, you should review commits one by one

NOTE: Some features will be handled later (in other PRs)
- Spread constraints on the VIPs
- CNI reconfiguration (likely once we will have migrated to Tigera operator, to manage the CNI)
- Ingress server certificate "reconfiguration"

---

NOTE: This PR is a work in progress, still some stuff todo

- Add Go unit tests => in progress
- Add tests about VIPs management
- Add documentation
- Add changelog entry


operator/pkg/controller/utils/reconcile.go
```diff
@@ -0,0 +1,70 @@
...
+	return newReconcilerResult(endReconcile, nil)
+}
+
+// Return wether or not the main Reconcile function should return
+func (r ReconcilerResult) ShouldReturn() bool {
+	return r.status != nothingToDo
+}
+
+// Get the Return for the main Reconcile function
+func (r ReconcilerResult) GetReturn() (ctrl.Result, error) {
```
gdemonet: nit: should likely be called `GetResult`, since that's the type name of the method's return value.

operator/pkg/controller/utils/reconcile.go
```diff
@@ -0,0 +1,70 @@
...
+func Requeue(err error) ReconcilerResult {
+	return newReconcilerResult(requeue, err)
+}
+
+// A delayed requeue is needed
+func DelayedRequeue() ReconcilerResult {
+	return newReconcilerResult(delayRequeue, nil)
+}
+
+// End the Reconciliation
```
gdemonet: nit: no need for a capital R :wink:

operator/pkg/controller/utils/object.go
```diff
@@ -0,0 +1,160 @@
...
+const (
+	labelManagedByName  = "app.kubernetes.io/managed-by"
+	labelManagedByValue = "metalk8s-operator"
+
+	labelPartOfName  = "app.kubernetes.io/part-of"
+	labelPartOfValue = "metalk8s"
+
+	labelComponentName = "app.kubernetes.io/component"
+
+	labelVersionName = "metalk8s.scality.com/version"
```
gdemonet: There is a standard label key for this as well (`app.kubernetes.io/version`), IMO we should use it in addition to our own

operator/pkg/controller/utils/object.go
```diff
@@ -0,0 +1,160 @@
...
+	Client client.Client
+	Scheme *runtime.Scheme
+	Logger logr.Logger
+
+	Component string
+
+	labels map[string]string
+}
+
+func NewObjectHandler(instance *metalk8sv1alpha1.ClusterConfig, client client.Client, scheme *runtime.Scheme, logger logr.Logger, component string) *ObjectHandler {
```
gdemonet, line 64: Shouldn't we always document public functions (and types, I guess)?
In this specific case, I'm not sure what `component` is used for, and I only know what an ObjectHandler does because I read the commit message :)

operator/pkg/controller/virtualip/hl_config.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return nil
+}
+
+func (c *HLConfig) isSupportedApiVersion() bool {
+	for i := range supportedApiVersion {
+		if c.ApiVersion == supportedApiVersion[i] {
+			return true
+		}
+	}
```
gdemonet: Why the index-based loop here, while we usually just loop through items?

operator/pkg/controller/virtualip/controller.go
```diff
@@ -69,7 +83,213 @@ func (r *VirtualIPReconciler) Reconcile(ctx context.Context) utils.ReconcilerRes
...
+	if err := r.Client.List(ctx, &r.nodeList); err != nil {
+		return utils.Requeue(err)
+	}
+
+	objsToDelete := []client.Object{}
+	objsToUpdate := []client.Object{}
+
+	poolsName := []string{}
+
+	for pName := range pools {
```
gdemonet, line 140: Shouldn't this be `for _, pName := range pools {` ?

operator/pkg/controller/virtualip/controller.go
```diff
@@ -69,7 +83,213 @@ func (r *VirtualIPReconciler) Reconcile(ctx context.Context) utils.ReconcilerRes
...
 
+	// Retrieve all nodes
+	if err := r.Client.List(ctx, &r.nodeList); err != nil {
+		return utils.Requeue(err)
+	}
+
+	objsToDelete := []client.Object{}
+	objsToUpdate := []client.Object{}
+
+	poolsName := []string{}
```
gdemonet: nit: `poolNames` :wink:

operator/pkg/controller/virtualip/controller.go
```diff
@@ -0,0 +1,39 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	metalk8sv1alpha1 "github.com/scality/metalk8s/operator/api/v1alpha1"
+	"github.com/scality/metalk8s/operator/pkg/controller/utils"
+)
+
+type VirtualIPReconciler struct {
+	utils.ObjectHandler
+}
```
gdemonet: This is black magic to me, do you have some reading material I could get my hands on to understand this better? IIUC, by defining this "anonymous" field in a struct, the struct inherits the public methods from its child...? e.g. the `CreateOrUpdateOrDelete` method from the commit before?

operator/pkg/controller/virtualip/controller.go
```diff
@@ -15,6 +19,9 @@ import (
 
 type VirtualIPReconciler struct {
 	utils.ObjectHandler
+
+	usedVRID map[int]bool
```
gdemonet, line 30: Could be worth adding a comment here, explaining that we need this cache to ensure we don't reuse VRIDs, even for different pools

operator/pkg/controller/virtualip/controller.go
```diff
@@ -293,3 +309,80 @@ func (r *VirtualIPReconciler) mutateConfigMap(obj *corev1.ConfigMap) error {
...
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: r.getObjName(poolName),
+					},
+					DefaultMode: &defaultMode,
+				},
+			},
+		},
+	}
+
+	// If we have more than 1 container we clean up the containers list
```
gdemonet, line 432: Guess that's fine for now, but why do we need to do this? Can't we just select by name? I'm thinking we might want to let admins add some sidecar, but not sure... Anyway, that's nitpicking, no need to change anything until the need comes up.

operator/api/v1alpha1/clusterconfig_types.go
```diff
@@ -84,6 +87,57 @@ type ClusterConfig struct {
...
+func (c *ClusterConfig) SetCondition(
+	kind string,
+	status metav1.ConditionStatus,
+	reason string,
+	message string,
+) {
+	condition := metav1.Condition{
+		Type:               kind,
+		Status:             status,
+		ObservedGeneration: c.Generation,
```
gdemonet, line 110: Oh! Is this "standard", i.e. always part of `metav1.Condition`? I'm thinking we may have "over-engineered" https://github.com/scality/zenko-operator/pull/328 (cc @williamlardier) while having this available in each condition allows us to know which one was updated after seeing the CR update... :thinking: Anyway, the current approach should work as well :innocent: but it's good to know!

operator/pkg/controller/virtualip/controller.go
```diff
@@ -157,17 +202,32 @@ func (r *VirtualIPReconciler) Reconcile(ctx context.Context) utils.ReconcilerRes
...
+			metav1.ConditionFalse,
+			"ObjectUpdateInProgress",
+			"Creation/Update of various objects in progress",
+		)
 		return utils.EndReconciliation()
 	}
 
 	// TODO: Reconfigure CNI + re-generate the Ingress server certificate
 
+	r.updateCondition(metav1.ConditionTrue, "Configured", "All objects properly configured")
```
gdemonet, line 228: We stop here? What about checking that the keepalived DaemonSets are actually ready? I don't think it's actually needed atm, and it's even probably not as useful as one might think (we are more interested in knowing the status of the VIPs themselves than that of the Pods), but just wanted to make that clear... It's fine if we say that for now, our status only cares about "K8s objects being properly configured", and once we add monitoring (with our sidecar "API"), we'll add more info about the availability of each VIP.

gdemonet (commented):
As usual, really good stuff :100: 

I was a bit confused at first by the "magic update/create/delete with a generic mutate handler", but it does make the code less redundant and keeps things simpler when considering a single object type to mutate.

Otherwise, just had a few questions and requests here and there, but nothing blocking or of significant impact :wink:

operator/api/v1alpha1/clusterconfig_types.go
ghost: Hey, yes we thought about that for Zenko as well, but we wanted to have it as a standalone value, similar to the deployments, so we can easily check its value regardless of the conditions. But maybe this can be improved, depending on what we want to achieve at the product level.

operator/pkg/controller/virtualip/controller.go
```diff
@@ -0,0 +1,474 @@
...
+		},
+	}
+
+	// If we have more than 1 container we clean up the containers list
+	if len(obj.Spec.Template.Spec.Containers) != 1 {
+		obj.Spec.Template.Spec.Containers = []corev1.Container{{}}
+	}
+	container := &obj.Spec.Template.Spec.Containers[0]
+
+	container.Name = "keepalived"
```
JBWatenbergScality, line 438: I'm wondering wether we should abstract away this controller from keepalived and not take assumption that we are relying on it for the implementation. We could introduce an intermediate layer "VirtualIpProvider" or something similar which could then be implemented using keepalived, metallb or whatever else if a new virtualip implementation suits better our needs later on, this would ease keepalived replacement

operator/pkg/controller/utils/reconcile.go
TeddyAndrieux: True but this function return the Result + error, but yes maybe GetResult is clearer, I will change it

operator/pkg/controller/utils/object.go
TeddyAndrieux: Right, that makes sense, I will add it.

Edit: Hmm I think we want to keep the `app.kubernetes.io/version` for the version of the component itself (so in our case for keepalived since we use the MetalK8s it will be the same but, it's not always the same)
But still I will add it just it will not always match the MetalK8s version

operator/pkg/controller/utils/object.go
TeddyAndrieux: Ok, make sense

operator/pkg/controller/virtualip/hl_config.go
TeddyAndrieux: TBH, no good reason, I will change it

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: I do not have any specific link for it, but yes exactly, it's like inheritance

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: No, because `pools` is a map and here I iterate on the map "keys" (and I do not need any information from the "values")

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: Fair

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: We could use the name to retrieve the right container, but, to me, we manage this DaemonSet so the user is not supposed to change it, if we want to support sideCars we will want to expose it in our ClusterConfig

operator/api/v1alpha1/clusterconfig_types.go
TeddyAndrieux: > Oh! Is this "standard", i.e. always part of `metav1.Condition`?

Indeed, and from what I see (but I did not tested myself) the `kubectl wait --for=condition` honor it by default which means that it will wait for the `ObservedGeneration` to match the object `Generation`

ghost: cc @francoisferrand do you think this could be helpful in a Zenko context?

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: I was wondering about it as well, it was simple to only handle "configuration" but I think I will wait for the DaemonSet pods to be ready as well

operator/pkg/controller/virtualip/controller.go
TeddyAndrieux: Hmm yes, we can do this, not sure it's really worth but ... may make sense.

Edit: The fact is that even the "HLConfig" is quite linked to keepalived, we also need a "global" knowledge of all the Virtual Router IDs used.
If one day we change the "Virtual IP Provider" to me the logic will be quite different, I'm not sure it's really worth building an intermediate layer today, that will be over-engineer to be able to support potential other providers (or that will need some tuning in the future to support "this" other provider).
If we change the "Virtual IP Provider", to me, the only logic we can keep from this "controller" is the namespace "management" and .... that's pretty much it, all the re-usable logic is part of the "ObjectHandler" object to me

TeddyAndrieux:
Close this PR as it get reworked and merged by #3864

PR #3859, opened 2022-08-29
ui: Update the dev environment of authentication following Dex'removal

ChengYanJin:
**Component**:  UI

**Context**: 
Give the fact of we removed Dex and use Keyclock as Authentication provider, we have to update the auth configuration in runtime-app-configuration of each microApp. 

**Summary**:

**Acceptance criteria**:


ChengYanJin:
We still use Dex if we run MetalK8s alone.

PR #3864, opened 2022-09-07
operator: Add Keepalived Controller (v2)

TeddyAndrieux:
(Second version of #3856, this time we have a dedicated CRD for VirtualIPPool)

- Move ClusterConfig controller logic to `pkg` to ease "muli controller" approach
- Add some utils to manage Objects
- Add Go variable in operator with MetalK8s version
- Add the new `VirtualIPPool` API and resource
- Handle High-Level ConfigMap and DaemonSet creation/update/deletion from VirtualIPPool controller
- Handle MetalK8s VIP Namespace creation/update/deletion from ClusterConfig controller
- Handle `VirtualIPPool` creation/update/deletion from ClusterConfig controller
- Add (and update) some status on ClusterConfig and VirtualIPPool CRs
- Add some events on ClusterConfig and VirtualIPPool CRs
- Add end-to-end tests about VIPs management
- Add documentation about VIPs management

NOTE: For reviewers, you should review commits one by one

NOTE: Some features will be handled later (in other PRs)
- Spread constraints on the VIPs
- CNI reconfiguration (likely once we will have migrated to Tigera operator, to manage the CNI)
- Ingress server certificate "reconfiguration"


operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
```diff
@@ -37,10 +37,146 @@ spec:
...
+                items:
+                  description: "Condition contains details for one aspect of the current
+                    state of this API Resource. --- This struct is intended for direct
+                    use as an array at the field path .status.conditions.  For example,
+                    type FooStatus struct{ // Represents the observations of a foo's
+                    current state. // Known .status.conditions.type are: \"Available\",
+                    \"Progressing\", and \"Degraded\" // +patchMergeKey=type // +patchStrategy=merge
+                    // +listType=map // +listMapKey=type Conditions []metav1.Condition
+                    `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\"
+                    protobuf:\"bytes,1,rep,name=conditions\"` \n // other fields }"
```
JBWatenbergScality: Maybe we should review a bit this description to at least drop the `foo` references

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
```diff
@@ -37,10 +37,146 @@ spec:
...
+                      - "True"
+                      - "False"
+                      - Unknown
+                      type: string
+                    type:
+                      description: type of condition in CamelCase or in foo.example.com/CamelCase.
+                        --- Many .condition.type values are consistent across resources
+                        like Available, but because arbitrary conditions can be useful
+                        (see .node.status.conditions), the ability to deconflict is
+                        important. The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
```
JBWatenbergScality, line 163: same here

operator/config/crd/bases/metalk8s.scality.com_virtualippools.yaml
```diff
@@ -0,0 +1,170 @@
...
+    controller-gen.kubebuilder.io/version: v0.9.0
+  creationTimestamp: null
+  name: virtualippools.metalk8s.scality.com
+spec:
+  group: metalk8s.scality.com
+  names:
+    kind: VirtualIPPool
+    listKind: VirtualIPPoolList
+    plural: virtualippools
+    singular: virtualippool
```
JBWatenbergScality, line 17: nit: as this resource name starts to be a bit long I would suggest give it a `shortName` such as `vipp` or `vpp` 👨‍🎨

operator/pkg/controller/clusterconfig/virtualippool.go
```diff
@@ -0,0 +1,198 @@
...
+	// Reconcile the VIP namespace
+	namespaceInstance := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: vipNamespaceName}}
+
+	if len(pools) == 0 {
+		// The namespace shouldn't exists
+		changed, err := handler.CreateOrUpdateOrDelete(ctx, nil, []client.Object{namespaceInstance}, nil)
+		if err != nil {
+			r.setVIPConfiguredCondition(
+				metav1.ConditionFalse,
+				"NamespaceDeletionError",
```
JBWatenbergScality, line 38: I would suggest using an enum for the various conditions in the types definition (hence it can be documented in the CRD as well) and will uses error handling for clients such as the UI

operator/pkg/controller/clusterconfig/virtualippool.go
```diff
@@ -0,0 +1,198 @@
...
+		r.setVIPConfiguredCondition(
+			metav1.ConditionUnknown,
+			"NamespaceUpdateError",
+			err.Error(),
+		)
+		return utils.NeedRequeue(err)
+	}
+	if changed {
+		r.setVIPConfiguredCondition(
+			metav1.ConditionFalse,
```
JBWatenbergScality, line 70: Shall we have an "updating" condition. I'm thinking about use cases where we would want to display a status in the UI when the definition just has been changed. Reading the condition would be easy and would enable us to display the updating status in the UI

operator/pkg/controller/clusterconfig/virtualippool.go
```diff
@@ -0,0 +1,198 @@
...
+			return utils.NeedRequeue(err)
+		}
+
+		poolReady := pool.GetReadyCondition()
+
+		// If ObservedGeneration do not match the Generation then we still need to wait
+		if poolReady != nil && poolReady.ObservedGeneration != pool.Generation {
+			r.setVIPReadyCondition(
+				metav1.ConditionFalse,
+				"PoolNotUpdateToDate",
```
JBWatenbergScality:
```suggestion
				"PoolNotUpToDate",
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,425 @@
...
+			err = nil
+		}
+		r.setAvailableCondition(status, reason, message)
+		r.setReadyCondition(status, reason, message)
+		return utils.Requeue(err)
+	}
+
+	// If ObservedGeneration do not match the Generation then we still need to wait
+	if ds.Status.ObservedGeneration != ds.Generation {
+		reason := "DaemonSetNotUpdateToDate"
```
JBWatenbergScality:
```suggestion
		reason := "DaemonSetNotUpToDate"
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,425 @@
...
+// Return a not used Virtual Router ID and mark it as used
+func (r *VirtualIPPoolReconciler) getFreeVRID() int {
+	for i := 1; i <= 255; i++ {
+		if !r.usedVRID[i] {
+			r.usedVRID[i] = true
+			return i
+		}
+	}
+	// There is no more Virtual Router ID free
+	return -1
```
JBWatenbergScality, line 248: I wonder if using the return tuple (err, result) pattern wouldn't be more idiomatic in go

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	// Retrieve all configured Virtual Router IDs
+	r.usedVRID = make(map[int]bool, 255)
+	pools := metalk8sscalitycomv1alpha1.VirtualIPPoolList{}
+	if err := r.client.List(ctx, &pools); err != nil {
+		return utils.Requeue(err)
+	}
+	for _, pool := range pools.Items {
+		r.retrieveVRID(ctx, pool)
```
JBWatenbergScality: Shall we handle the possible error here ?

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,425 @@
...
+		return nodes
+	}
+
+	// Spread the IPs on every nodes
+	// NOTE: We take the "Ceil" since, in case where we have a number of
+	// IPs that is not a multiple of nodes, the last IPs may sit on any of
+	// those nodes
+	number := math.Ceil(float64(len(r.instance.Spec.Addresses)) / float64(len(nodes)))
+	for node := range nodes {
+		nodes[node] = int(number)
```
JBWatenbergScality: Instead of casting to int shouldn't we round ?

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,425 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&metalk8sscalitycomv1alpha1.VirtualIPPool{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&appsv1.DaemonSet{}).
+		Complete(reconciler)
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *VirtualIPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
JBWatenbergScality, line 77: nit: I'm not a big fan of naming vars with such short name, reviewing it is not sufficiently explicit in my mind
```suggestion
func (virtualIPPoolReconciler *VirtualIPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```

operator/pkg/controller/virtualippool/controller.go
JBWatenbergScality: If I understand correctly the sequence here `r.retrieveVRID` is mutating `r.usedVRID` so when we use it we have to be aware of the exact sequence we need to follow. IMO we should try to avoid this imperative paradigm. Maybe we can use an intermediate facade 🤔

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
TeddyAndrieux: Hmm right, I will update it

operator/config/crd/bases/metalk8s.scality.com_virtualippools.yaml
TeddyAndrieux: Oh yes, I was thinking about it as well but I forgot, I will do it

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: For VirtualIPPool object, yes we can but for ClusterConfig, I was thinking about only one "global" condition `Ready` and then the SubReconciler can manage the Conditions they want, but ... :thinking:  I can still do an enum, it just means that if a SubReconciler add a Condition they need to update the ENUM.

Ok fine, I will do it

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: Hmm, the fact is that some "computation" + all object push are done before actually reaching this point where we put this Condition, so I'm not sure an "Updating" condition will make much sense.

To me, for the use case you describe, you should act the same as a `kubectl wait --for=condition` so you basically show "updating status" in the UI until the `Ready` Condition `ObservedGeneration` match the actual object `Generation` and the Status of the Condition is `True`, I'm not sure we need a dedicated Condition for that, especially since by default, nowadays, the condition contains the `ObservedGeneration`

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: Oh :smile:

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Copy/pasta

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Maybe, but not sure it's worth it here, it will make the usage of the function a bit more complex and it's really a case that should not happen (a user that wants to use more than 255 VIPs at the same time :confused: )

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: > Shall we handle the possible error here ?

Indeed, I forgot about it, I will add it

TeddyAndrieux: I agree it's a bit weird, but I do not see any "clean" solution to do it

> If I understand correctly the sequence here `r.retrieveVRID` is mutating `r.usedVRID`

Yes

> so when we use it we have to be aware of the exact sequence we need to follow

WDYM ?

To explain a bit the behavior here, we retrieve all the defined pools, we iter on those pools to retrieve the "potential" already defined VRID to "cache" them, so that if we need to pick a new VRID we pick one that is not already used

> Maybe we can use an intermediate facade

I'm not sure to understand how it could work, except if we have "something" that basically manages all the VirtualIPPools at once, but it makes things complicated and not really good as well IMHO.

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: We already `Ceil`, not sure `round` is better than casting to `int` in this case, but we could cast to `int` only once when "computing" `number`, I will change it

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: I agree, but it's kind of "standard" in Go from what I saw

operator/pkg/controller/clusterconfig/virtualippool.go
JBWatenbergScality: Not sure it is possible in Golang but can an Enum extend another one ?

operator/pkg/controller/clusterconfig/virtualippool.go
JBWatenbergScality: Ok fair 👍

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
TeddyAndrieux: Done

operator/config/crd/bases/metalk8s.scality.com_virtualippools.yaml
TeddyAndrieux: Done

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
TeddyAndrieux: Done

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: So I did not define a dedicated Type for it, since it was a mess then because we couldn't have used the `metav1.Condition` type directly (or we should have built a function to convert from one to the other). But I defined a `const` close to the CRD per conditions

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: Done

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Done

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: > Shall we handle the possible error here ?

Done

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: I moved the cast to `int` to do it only once

operator/pkg/controller/virtualippool/controller.go
gdemonet: Indeed, it seems that the short name indicates the object to which the current method applies. In storage-operator, we replaced it with `self`, similar to what we use in Python. Not sure which is best, but I think it's useful to differenciate this "method holder" from other variables.

operator/api/v1alpha1/virtualippool_types.go
```diff
@@ -17,19 +17,40 @@ limitations under the License.
...
-	// Foo is an example field of VirtualIPPool. Edit virtualippool_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// Node Selector to deploy the Virtual IPs manager
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+	// Tolerations to deploy the Virtual IPs manager
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// Spread constraints for the Virtual IPs
+	// NOTE: Not supported yet
```
gdemonet, line 48: Do we have a ticket to track this feature?

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -57,7 +75,177 @@ func (r *VirtualIPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if changed {
+		return utils.EndReconciliation()
+	}
+
 	// TODO(user): your logic here
 
 	return utils.EndReconciliation()
 }
+
+func (r *VirtualIPPoolReconciler) retrieveVRID(ctx context.Context, pool metalk8sscalitycomv1alpha1.VirtualIPPool) error {
```
gdemonet: nit: the name is slightly misleading, we could expect to get a VRID as a result, but it's actually just storing them in "the cache"... Maybe `cacheUsedVRIDs` instead?

gdemonet: Also, a quick comment explaining why we use this "cache" pattern could avoid questions in the future, or eager attempts to simplify it (IIUC, we need it as a consequence of the "magic create/update/delete with a `mutate` handler" pattern)

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -57,7 +75,177 @@ func (r *VirtualIPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	for node, nb := range nodes {
+		for i := 0; i < nb; i++ {
+			for index := range desired.Addresses {
+				if desired.Addresses[index].Node == "" {
+					desired.Addresses[index].Node = node
+					break
+				}
+			}
+		}
+	}
```
gdemonet: Isn't this loop slightly wrong? Imagine we have 3 nodes, and 4 VIPs to spread. `Ceil(4/3) = 2`, so we'll have `nb = 2`. Which means, IIUC, that we'll assign 2 VIPs to the first node, 2 more to the second node, and none to the third :confused:

Is there any use case where we'd have a different `nb` per node selected in a pool? I don't remember any...
If not, I'd suggest we refactor the `getNodesList` method to only return a list of candidate node names, and compute the `number` (max per node) only once, in this `mutate` handler. Then the loops would be switched around:

```
for i := 0; i < maxAddressesPerNode; i++ {
    for _, node := range nodes {
        // assign one address
    }
}
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -57,7 +75,177 @@ func (r *VirtualIPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	r.usedVRID = make(map[int]bool, 255)
+	pools := metalk8sscalitycomv1alpha1.VirtualIPPoolList{}
+	if err := r.client.List(ctx, &pools); err != nil {
+		return utils.Requeue(err)
+	}
+	for _, pool := range pools.Items {
+		if err := r.retrieveVRID(ctx, pool); err != nil {
+			return utils.Requeue(err)
+		}
+	}
```
gdemonet: An idea comes to mind, but let's keep it for another PR.

I'm actually concerned by this kind of logic within a Reconcile pass: we should be reconciling a single instance, without having to retrieve all instances again. I think we should be fine because the `client` already has a cache built in, so there wouldn't be an extra call sent to K8s API (neither for the "list pools" nor for the "get configmap").
Still, it would be cleaner (IMO) to have each Pool reconciliation update the cache (about the VRIDs it's using) without having to recreate it on each pass for each Pool.

operator/pkg/controller/clusterconfig/virtualippool.go
```diff
@@ -0,0 +1,197 @@
...
+			ObjectMeta: r.getPoolMeta(name),
+		}
+		if err := r.client.Get(ctx, client.ObjectKeyFromObject(pool), pool); err != nil {
+			status := metav1.ConditionUnknown
+			reason := "PoolRetrievingError"
+			message := err.Error()
+			if errors.IsNotFound(err) {
+				status = metav1.ConditionFalse
+				reason = "PoolNotCreated"
+				message = fmt.Sprintf("The VirtualIPPool %s does not exists yet", pool.GetName())
```
gdemonet: Since it's user-facing:

```suggestion
				message = fmt.Sprintf("The VirtualIPPool %s does not exist yet", pool.GetName())
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -0,0 +1,428 @@
...
+	// Wait for DaemonSet replicas to be Ready
+	ds := r.instance.GetDaemonSet()
+	if err := r.client.Get(ctx, client.ObjectKeyFromObject(ds), ds); err != nil {
+		status := metav1.ConditionUnknown
+		reason := "DaemonSetRetrievingError"
+		message := err.Error()
+		if errors.IsNotFound(err) {
+			status = metav1.ConditionFalse
+			reason = "DaemonSetNotCreated"
+			message = "The DaemonSet does not exists yet"
```
gdemonet:
```suggestion
			message = "The DaemonSet does not exist yet"
```

operator/api/v1alpha1/virtualippool_types.go
```diff
@@ -22,7 +22,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const configuredConditionName = "Configured"
+const (
+	configuredConditionName = "Configured"
+	availableConditionName  = "Available"
+	readyConditionName      = "Ready"
```
gdemonet: Not sure if that's the right place, or if we should add it to the `description` in OpenAPI spec, but it would be nice to explain each condition type and their nuances (e.g. Available = at least one up, Ready = all up).

operator/pkg/controller/utils/reconcile_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func assertRequeue(result ctrl.Result, err error) {
+	Expect(err).ToNot(HaveOccurred())
+	Expect(result.Requeue).To(BeTrue())
+	Expect(result.RequeueAfter).To(BeZero())
+}
+
+func assertRequeueError(myErr error, result ctrl.Result, err error) {
+	Expect(err).To(HaveOccurred())
+	Expect(err.Error()).To(Equal("An ErRoR !"))
```
gdemonet:
```suggestion
	Expect(err.Error()).To(Equal(myErr))
```

gdemonet (changes requested):
Just a few minor changes, except for the spreading algorithm which I believe is wrong atm, otherwise LGTM :slightly_smiling_face:

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Indeed, no I will create it and the one about CNI + certs as well

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: > IIUC, we need it as a consequence of the "magic create/update/delete with a `mutate` handler" pattern

Not really, we need it, because we do not want to use the same VRID as another VirtualIPPool VIP so we retrieve all the VRID from all the VirtualIPPools

I will rename it and add a comment

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: > Isn't this loop slightly wrong? Imagine we have 3 nodes, and 4 VIPs to spread. `Ceil(4/3) = 2`, so we'll have `nb = 2`. Which means, IIUC, that we'll assign 2 VIPs to the first node, 2 more to the second node, and none to the third

:thinking: hmm indeed, ok I will fix it

> Is there any use case where we'd have a different `nb` per node selected in a pool? I don't remember any...
> If not, I'd suggest we refactor the `getNodesList` method to only return a list of candidate node names, and compute the `number` (max per node) only once, in this `mutate` handler.

The fact is that we do not want to change the address if it's already assigned to one node, see the loop above 
```go
		if addr.Node != "" {
			// The node is already defined
			if nodes[addr.Node] == 0 {
				// This node is not available, remove it
				addr.Node = ""
			} else {
				nodes[addr.Node]--
			}
		}
```

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: I agree that it's not ideal but I don't see any better approach.

> Still, it would be cleaner (IMO) to have each Pool reconciliation update the cache (about the VRIDs it's using) without having to recreate it on each pass for each Pool.

It will not work well, it means the first VirtualIPPool to get reconciled may take one VRID already taken by another Pool, or we need to load all the pools at the startup of the controller but not ideal as well.

TBH, we will never have tonnes of VirtualIPPools at once so I don't think it's really a problem, even if I agree it's not ideal.

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Ok I will add some explanation

operator/pkg/controller/utils/reconcile_test.go
TeddyAndrieux: Indeed, I can do something like that

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Done (already squashed for this one)

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Done https://github.com/scality/metalk8s/pull/3864/commits/2c7edac0408aea18fe3b41184e70085045c56ccb

operator/pkg/controller/clusterconfig/virtualippool.go
TeddyAndrieux: Done

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Done

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Done https://github.com/scality/metalk8s/pull/3864/commits/d79b19c49a776c53abf77dcc3e68fa858351a664

operator/pkg/controller/utils/reconcile_test.go
TeddyAndrieux: Done

operator/pkg/controller/virtualippool/controller.go
gdemonet: Yes that part I got, but the fact that we need it attached to the reconciler as a struct field is only because of the `mutate`, right? Because we refresh this cache on every `Reconcile` call, no? If we don't, then we may be missing invalidation (when we remove a pool / an address from a pool).

TeddyAndrieux: Oh yes indeed, it could be just a variable in the reconcile function and then we add a "lambda" function so that we give this argument to the mutate, but it will be almost the same. And indeed it's re-computed at every reconcile loop

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Done

gdemonet (approved):
:100: :shipit:

operator/api/v1alpha1/virtualippool_types.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Set Ready Condition
+func (v *VirtualIPPool) SetReadyCondition(status metav1.ConditionStatus, reason string, message string) {
+	v.SetCondition(readyConditionName, status, reason, message)
+}
+
+// Get Ready Condition
+func (v *VirtualIPPool) GetReadyCondition() *Condition {
+	return v.GetCondition(readyConditionName)
+}
```
vietnscality, line 131: just a minor note: randomly i saw this PR, not understand much but you can remove the comment, the function/method name are clean enough so self-documented

TeddyAndrieux: Agree, but it's good practice in Golang to have a comment for every "exported" method (I'm not sure we do have a comment for all of them here... but ideally we should)

`golint` would complain if we do not have such comments (we should add `golint` validation in the CI btw)

PR #3866, opened 2022-09-13
build: Add codegen for addons charts rendering

TeddyAndrieux:
Allow to use doit to generate the charts SLS, and also validate the generated SLS content in the CI

```console
$ ./doit.sh list --all codegen
codegen                               Run the code generation tools.
codegen:chart_dex                     Generate the SLS file for Dex using the chart render script.
codegen:chart_fluent-bit              Generate the SLS file for fluent-bit using the chart render script.
codegen:chart_ingress-nginx           Generate the SLS file for NGINX Ingress using the chart render script.
codegen:chart_kube-prometheus-stack   Generate the SLS file for Kube Prometheus Stack using the chart render script.
codegen:chart_loki                    Generate the SLS file for Loki using the chart render script.
codegen:chart_metallb                 Generate the SLS file for MetalLB using the chart render script.
codegen:chart_prometheus-adapter      Generate the SLS file for Prometheus Adapter using the chart render script.
codegen:chart_thanos                  Generate the SLS file for Thanos using the chart render script.
codegen:metalk8s_operator             Generate Go code for the MetalK8s Operator using the Operator SDK Makefile.
codegen:storage_operator              Generate Go code for the Storage Operator using the Operator SDK Makefile.
```

---

NOTE: This is not the best approach the `charts/render.py` script should be migrated in the buildchain (or we should totally change the way to manage "addons" but ... you know :smile: ), but at least it's better than having the command in commit message :)

### DIFF
--- a/buildchain/buildchain/codegen.py
+++ b/buildchain/buildchain/codegen.py
@@ -14,6 +14,21 @@ from buildchain import types
 from buildchain import utils
 
 
+def get_task_information() -> types.TaskDict:
+    """Retrieve all the task information from codegen"""
+    result: types.TaskDict = {
+        "actions": [],
+        "task_dep": [],
+        "file_dep": [],
+    }
+    for task_fun in CODEGEN:
+        task = task_fun()
+        for key, value in result.items():
+            value.extend(task.get(key, []))
+
+    return result
+
+
 def task_codegen() -> Iterator[types.TaskDict]:
     """Run the code generation tools."""
     for create_codegen_task in CODEGEN:
--- a/buildchain/buildchain/codegen.py
+++ b/buildchain/buildchain/codegen.py
@@ -318,6 +318,34 @@ def codegen_chart_prometheus_adapter() -> types.TaskDict:
     }
 
 
+def codegen_chart_thanos() -> types.TaskDict:
+    """Generate the SLS file for Thanos using the chart render script."""
+    target_sls = (
+        constants.ROOT
+        / "salt/metalk8s/addons/prometheus-operator/deployed/thanos-chart.sls"
+    )
+    chart_dir = constants.CHART_ROOT / "thanos"
+    value_file = constants.CHART_ROOT / "thanos.yaml"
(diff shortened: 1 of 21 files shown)


buildchain/buildchain/codegen.py
```diff
@@ -14,6 +14,21 @@
 from buildchain import utils
 
 
+def get_task_information() -> types.TaskDict:
+    """Retrieve all the task information from codegen"""
+    result = {
+        "actions": [],
+        "task_dep": [],
+        "file_dep": [],
+    }  # type: types.TaskDict
```
gdemonet: To stay consistent with other type annotations in the `buildchain` package:

```suggestion
    result: types.TaskDict = {
        "actions": [],
        "task_dep": [],
        "file_dep": [],
    }
```

This also should remove the need for having "TaskDict" in our `.pylint-dict`

buildchain/static-container-registry/test.sh
```diff
@@ -21,31 +21,31 @@ TEST_SKOPEO=${TEST_SKOPEO:-1}
 
 if [ "$TEST_DOCKER" -eq 1 ]; then
 test_docker() {
-        for image in ${AVAILABLE_IMAGES[*]}; do
+        for image in "${AVAILABLE_IMAGES[@]}"; do
```
gdemonet, line 24: Shouldn't we rather ignore these files when running shellcheck? They come from another repository (which I don't remember how we imported though), we shouldn't really care IMO.

gdemonet (commented):
Great first step, thank you!!

buildchain/static-container-registry/test.sh
TeddyAndrieux: Indeed, maybe we should

buildchain/buildchain/codegen.py
TeddyAndrieux: right

gdemonet (approved):

PR #3869, opened 2022-09-16
ARTESCA-5471 Upgrade Core-UI and use Icon and colors properly

ghivert:
**Component**:

A lot.

**Context**: 

Updating Icons and Colors


ui/src/containers/NodeList.js
```diff
@@ -0,0 +1,192 @@
+//@flow
```
ChengYanJin: In fact, we've already removed this file in development/124.0  - it's not used anymore.

JBWatenbergScality (approved):

PR #3877, opened 2022-09-26
salt,tests: Allow forcing the creation of an LV

gdemonet:
In case a formatted LV was removed without wiping its FS signature, creation from storage-operator will fail. We add support for an annotation (metalk8s.scality.com/force-lvcreate) to force the creation of the LV, at the risk of also wiping the previous FS contents.

### DIFF
--- a/tests/post/features/volume.feature
+++ b/tests/post/features/volume.feature
@@ -269,3 +269,38 @@ Feature: Volume management
         Then the Volume 'test-volume12-lvmlv' is 'Available'
         And the PersistentVolume 'test-volume12-lvmlv' has size '10Gi'
         And the device '/dev/test-vg-12/test-volume12-lvmlv' exists
+
+    Scenario: Test volume re-creation (lvmLogicalVolume force-lvcreate)
+        # Skip test on CentOS/RHEL 7, because lvm still creates when signatures are found
+        Given the grain 'osmajorrelease' is not 7
+        And the Kubernetes API is available
+        And a LVM VG 'test-vg-13' exists
+        And a LVM LV 'test-lv-13' in VG 'test-vg-13' was created, formatted, then removed
+        When I create the following Volume:
+            apiVersion: storage.metalk8s.scality.com/v1alpha1
+            kind: Volume
+            metadata:
+              name: test-lv-13
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s
+              lvmLogicalVolume:
+                vgName: test-vg-13
+                size: 1Gi
+        Then the Volume 'test-lv-13' is 'Failed' with code 'CreationError' and message matches 'signature detected on /dev/test-vg-13/test-lv-13'
+        When I delete the Volume 'test-lv-13'
+        Then the Volume 'test-lv-13' does not exist
+        When I create the following Volume:
+            apiVersion: storage.metalk8s.scality.com/v1alpha1
+            kind: Volume
+            metadata:
+              name: test-lv-13
+              annotations:
+                metalk8s.scality.com/force-lvcreate: ""
+            spec:
+              nodeName: bootstrap
+              storageClassName: metalk8s
+              lvmLogicalVolume:
+                vgName: test-vg-13
+                size: 1Gi
+        Then the Volume 'test-lv-13' is 'Available'
--- a/tests/post/steps/test_volume.py
+++ b/tests/post/steps/test_volume.py
@@ -125,6 +125,14 @@ def test_volume_creation_lvmlv_block(host, teardown):
     pass
 
 
+@scenario(
+    "../features/volume.feature",
+    "Test volume re-creation (lvmLogicalVolume force-lvcreate)",
+)
+def test_volume_recreation_lvm_force(host, teardown):
+    pass
+
+
 # }}}
 # Given {{{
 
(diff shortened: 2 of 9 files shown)


ui/src/containers/CreateVolume.js
```diff
@@ -666,10 +672,34 @@ const CreateVolume = (props) => {
...
+                          label={`${intl.formatMessage({ id: "vg_name" })}*`}
                           error={touched.vgName && errors.vgName}
                           onBlur={handleOnBlur}
                         />
+                        <Input
+                          name="forceLVCreate"
+                          value={values.forceLVCreate}
+                          onChange={handleChange('forceLVCreate')}
+                          label={
+                            <>
```
JBWatenbergScality: I think this should be a toggle.

ui/src/containers/CreateVolume.js
```diff
@@ -666,10 +672,34 @@ const CreateVolume = (props) => {
...
+                                placement="right"
+                                overlay={
+                                  <div style={{ minWidth: '200px' }}>
+                                    {intl.formatMessage({
+                                      id: 'force_lvcreate_explanation',
+                                    })}
+                                  </div>
+                                }
+                              >
+                                <InputWarning className="fas fa-triangle-exclamation"></InputWarning>
```
JBWatenbergScality: nit: `fa-triangle-exclamation` icon is being deprecated in favor of `fa-exclamation-circle`

ui/src/containers/CreateVolume.js
```diff
@@ -666,10 +672,34 @@ const CreateVolume = (props) => {
...
+                          name="forceLVCreate"
+                          value={values.forceLVCreate}
+                          onChange={handleChange('forceLVCreate')}
+                          label={
+                            <>
+                              {`${intl.formatMessage({ id: "force_lvcreate" })}`}
+                              <Tooltip
+                                placement="right"
+                                overlay={
+                                  <div style={{ minWidth: '200px' }}>
```
JBWatenbergScality: nit: the new form component is going to handle this, so you don't have to care about it
```suggestion
<div style={{ minWidth: '80rem' }}>
```

ui/src/containers/CreateVolume.js
```diff
@@ -208,6 +208,10 @@ const InputQuestionMark = styled.i`
   color: ${(props) => props.theme.infoPrimary};
 `;
 
+const InputWarning = styled(InputQuestionMark)`
```
JBWatenbergScality: nit: `InputWarning` implies that it would be an input component simply with color changing to `warning`. However here it is the Icon that change. Moreover we have a new `Icon` component in core-ui that should be used for this purpose. `<Icon color='statusWarning' name='Exclamation-circle' />`

tests/post/features/volume.feature
```diff
@@ -269,3 +269,36 @@ Feature: Volume management
         Then the Volume 'test-volume12-lvmlv' is 'Available'
         And the PersistentVolume 'test-volume12-lvmlv' has size '10Gi'
         And the device '/dev/test-vg-12/test-volume12-lvmlv' exists
+
+    Scenario: Test volume re-creation (lvmLogicalVolume force-lvcreate)
```
TeddyAndrieux, line 273: From what I see you forgot to add this scenario in the test file, so this test is not run

tests/post/steps/test_volume.py
```diff
@@ -206,6 +206,28 @@ def storage_class_exist(name, sc_client):
...
+        host.run_test(f"mkfs.ext4 /dev/{vg_name}/{name}")
+        host.run_test(f"lvremove -y /dev/{vg_name}/{name}")
+
+    yield
+
+    with host.sudo():
+        lv_exists = host.run(f"lvdisplay /dev/{vg_name}/{name}")
+        if lv_exists.rc != 0:
+            # LV does not exist, attempt to recreate for wiping the signature
+            host.run_test(f"lvcreate -L 1G -n {name} -y {vg_name}")
```
TeddyAndrieux, line 235: The wipefs is missing here I think

gdemonet: Oh yes, I need an `else` for this (because the `lvcreate -y` will handle the wipe)

gdemonet: (actually will do it everytime, wipefs doesn't care if there is actually something to wipe)

tests/post/features/volume.feature
TeddyAndrieux, line 273: From the CI build it seems you got a failure on this test in 3-nodes (flaky ?, weird, I don't see how this test could be flaky and the one for the single node succeeded :confused: )

gdemonet: Yeah I need to debug this, haven't had time yet

gdemonet: For reference: this was caused by an older version of LVM in CentOS 7, which doesn't prevent LV creation even if signatures were detected and not wiped. Added a note about this in the CHANGELOG, and skipped the test on CentOS 7 / RHEL 7 (actually, just testing the `osmajorrelease` grain, but we only support RedHat-family OSes, so should be enough).

tests/post/features/volume.feature
```diff
@@ -269,3 +269,37 @@ Feature: Volume management
...
+        Given the grain 'osrelease' is not '7'
```
TeddyAndrieux: Worth a comment I think here

gdemonet: Will do!

gdemonet: Argh, I'm stupid (`s/osrelease/osmajorrelease/`)

gdemonet: done

TeddyAndrieux (approved):
LGTM

PR #3884, opened 2022-09-30
build,salt: Upgrade Calico to 3.24.1

TeddyAndrieux:
Images bumped with:
  docker.io/calico/node:v3.24.1
  docker.io/calico/kube-controllers:v3.24.1
  docker.io/calico/cni:v3.24.1

Manifest updated based on upstream from:
  https://github.com/projectcalico/calico/blob/v3.24.1/manifests/calico.yaml

### DIFF
--- a/salt/metalk8s/kubernetes/cni/calico/deployed.sls
+++ b/salt/metalk8s/kubernetes/cni/calico/deployed.sls
@@ -8,6 +8,38 @@
 # Various changes to the original are made, based on how we deploy Calico (and
 # its CNI plugins etc.) within MetalK8s.
 
+# It comes from: https://github.com/projectcalico/calico/blob/v3.24.1/manifests/calico.yaml
+
+---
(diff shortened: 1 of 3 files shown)


salt/metalk8s/kubernetes/cni/calico/deployed.sls
```diff
@@ -4389,12 +4477,38 @@ spec:
               name: cni-net-dir
           securityContext:
             privileged: true
+        # This init container mounts the necessary filesystems needed by the BPF data plane
+        # i.e. bpf at /sys/fs/bpf and cgroup2 at /run/calico/cgroup. Calico-node initialisation is executed
+        # in best effort fashion, i.e. no failure for errors, to not disrupt pod creation in iptable mode.
```
gdemonet, line 4482: Wondering if we really should include it, but I guess it's not going to cause issues (just a slightly longer startup time)...

gdemonet (approved):

PR #3901, opened 2022-10-19
Integrated Layout2 component

ognyan-kostadinov:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


shell-ui/src/navbar/NavBar.js
```diff
@@ -316,13 +317,18 @@ export const Navbar = ({
...
-      <CoreUINavbar
-        logo={<Logo src={logo} alt="logo" />}
-        rightActions={rightActions}
-        tabs={tabs}
-        role="navigation"
-      />
-      {children}
+      <Layout2
+        headerNavigation={
+          <CoreUINavbar
```
ChengYanJin, line 322: I think the navbarHeight in core-ui is not the same as Figma Desgin,should be 3rem if I am not mistaken.
`export const navbarHeight = '3.357rem';`

ui/src/containers/DashboardPage.js
```diff
@@ -9,13 +9,14 @@ import { padding, spacing } from '@scality/core-ui/dist/style/theme';
 import { SyncedCursorCharts } from '@scality/core-ui/dist/components/vegachartv2/SyncedCursorCharts';
 import TimespanSelector from './TimespanSelector';
 import DashboardNetwork from '../components/DashboardNetwork';
+import { AppContainer } from '@scality/core-ui';
 
 const DashboardGrid = styled.div`
   display: grid;
   gap: 8px;
```
ChengYanJin: I think the gap:8px is no longer valid. I guess it should be ${spacing.r1} base on Figma's design.

We defined the `height: calc(100vh - 3.357rem - 8px);` in the DashboardGrid, likely need to be updated as weel.

ui/src/containers/Layout.js
```diff
@@ -129,14 +122,14 @@ const Layout = () => {
   };
 
   return (
-    <PatchedCoreUiLayout
-      sidebar={
+    <AppContainer
```
ChengYanJin, line 125: it's not related to this specific change, but I've noticed that there is a global scroll bar in all the pages. 
We should likely check the reason and ideally remove it.

ui/src/containers/AlertPage.js
```diff
@@ -282,14 +284,18 @@ export default function AlertPage() {
 
   return (
     <AlertPageContainer>
```
ChengYanJin: I am just wondering if we could use our `<AppContainer/>` instead of the legacy `<AlertPageContainer/>` ?

ChengYanJin: I notice there is horizontal scrollbar in the alert page, probably it's the width is not correctly set. Let's check that ;)

ui/src/components/style/CommonLayoutStyle.js
```diff
@@ -21,19 +21,13 @@ export const CenteredPageContainer = styled.div`
 `;
 
 export const LeftSideInstanceList = styled.div`
-  flex-direction: column;
-  min-height: 696px;
-  width: 49%;
-  margin-bottom: 20px;
-  min-width: 49%;
+  flex: 1;
```
ChengYanJin, line 16: I am just wondering why couldn't we set this `flex:1` in the `<AppContainer.MainContent/>` level?

ui/src/components/DashboardGlobalHealth.js
```diff
@@ -78,89 +63,100 @@ const DashboardGlobalHealth = () => {
...
-      <div className="datacenter">
-        <SpacedBox ml={12} mr={12}>
-          <PlatformStatusIcon>
-            <StatusWrapper status={platformStatus}>
-              <StatusIcon status={platformStatus} name="Datacenter" />
-            </StatusWrapper>
-          </PlatformStatusIcon>
-        </SpacedBox>
+    <AppContainer.OverallSummary>
+      <GlobalHealthContainer>
```
ChengYanJin: We could simplify the GlobalHealthContainer by using the Stack component.

```js
<Stack gap="r32" withSeparators >
  <Box flex={0.4} />
  <Box flex={0.4} />
  <Box flex={0.4} />
</Stack>
```

You can also find the example in core-ui stories:
https://scality.github.io/core-ui/?path=/docs/components-navigation-layout--layout-2-simplest

shell-ui/src/navbar/NavBar.js
ognyan-kostadinov: You're right, it's fixed now

ui/src/containers/DashboardPage.js
ognyan-kostadinov: Got rid of the height property completely

ui/src/containers/Layout.js
ognyan-kostadinov: Fixed

ui/src/containers/AlertPage.js
ognyan-kostadinov: Yep, fixed

ognyan-kostadinov: I can't reproduce that, but I think it should be fixed along the changes which I pushed

ui/src/components/style/CommonLayoutStyle.js
ognyan-kostadinov: Got rid of the nested component which gave `flex: 1` in core-ui and thus had to use it here. The reason I got rid of it is because I needed to pass some extra props, but the extra (`rest`) params were passed to the FillAvailableFlexBox component. As I warned earlier - it's bad design practice to go along this "one-size-fits-all" approach, which gives developer convenience for some use-cases at the expense of preventing further customization

ui/src/components/DashboardGlobalHealth.js
ognyan-kostadinov: Nope, the Stack component doesn't allow to set separator width. And I don't see the point of a separate component which does such a trivial action. Apparently we would need to extend it endlessly (more PRs, releases of core-ui and so on) just to have some thin meaningless layer

JBWatenbergScality: > doesn't allow to set separator width

I don't think the separator width should be different than the one provided by the stack component. The whole idea of using the stack component is to unify it with others separators. 
The idea behind core-ui is to enforce a set of design rules so that such details don't need to be re-implemented in the app, and all the UIs got the same visual aspect. Even through this looks like an unsignificant details I would strongly push to use Stack separators and don't re-invent the wheel here. 
To me we don't need any changes on core-ui to add this separator. Through I will let @Cuervino confirm that the separation here should be the same as the other ones.

ognyan-kostadinov: Okay, given that you're strongly pushing for that option, can we make sure that the component actually fulfills the desired design? Right now it adds a whitespace (the width of which is font-size dependent), I see there's some min-height and min-width set in rem. Given that it's a separator, I don't see why we would want it to scale. The other problem is that the separator is outside the individual flex items, thus making it almost impossible to adjust its position in regards to other visual elements (you can go with flex values being something like 1.98 or whatever, but you see how absurd that is). And this situation and this screen are perfect example how you think you've made one component which serves all needs and in reality even in our existing pages - it doesn't. So that's why I suggested a compromise - at least add some mechanisms to extend the current behavior when and if needed. A sane default which covers most of the cases, and some props to tackle the more peculiar ones. Even in the epic-react all reusable components were built with extension in mind, every subcomponent and handler could be optionally replaced with a custom one

Here are two screenshots - one of my implementation and one of using Stack

platform-dashboard-current

platform-dashboard-with-stack-component

Cuervino: My 2 cents:

- About **responsive/scalable**: agree, it's really not needed to have them scaling.
- I won't push too much for **alignement between separators and the sections below**, as the content of the Overall summary is not final, we might have new areas in the futur (activities, or something else). In fact, these separators are not really needed, and there are not used in others places of the UIs.

Cuervino: Separator can be remoeved as long as there is enough spacing between the areas in the Overall_summary.

ChengYanJin:
Just adding some notes here in case we forget:
- Remove  `<PageContainer/>` in `VolumePage.js` and `NodePage.js` and potentially in `CommonLayoutStyle.js`
- In `VolumePageContent.js`, the icon name was wrongly spelled, which cause the UI cashed. 
```js
  if (!volumeListData.length && isFirstLoadingDone) {
    return (
      <EmptyState
        label={'Volume'}
        link="/volumes/createVolume"
        icon="Node-pdf"
        history={history}
      />
    );
  }
```

ui/src/containers/AlertPage.js
```diff
@@ -281,15 +274,19 @@ export default function AlertPage() {
...
+      <AppContainer.OverallSummary>
+        <AlertPageHeader
+          activeAlerts={leafAlerts.length}
+          critical={criticalAlerts.length}
+          warning={wariningAlerts.length}
+        />
+      </AppContainer.OverallSummary>
+      <AppContainer.MainContent hasPadding>
+        <AlertContent>
+          <ActiveAlertTab data={leafAlerts} columns={columns} />
```
ChengYanJin: I checked the reason we have the wrong non-selected color for the table row is because we didn't set correctly the `backgroundVariant` in the table.

```js
        <Table.SingleSelectableContent
          rowHeight="h48"
          separationLineVariant="backgroundLevel1"
          backgroundVariant="backgroundLevel3" // was wrongly set to backgroundLevel2 before
          customItemKey={(index, data) => {
            return data[index].id;
          }}
        />
```

ui/src/containers/AlertPage.js
```diff
@@ -163,7 +157,6 @@ export const TableHeader = styled.th`
 
 const AlertContent = styled.div`
```
ChengYanJin: I think we can eventually remove the <AlertContent>, the only issue we have today is the table doesn't take the 100% width of it's parents. I will push a fix in TableV2 core-ui.

BTW, Just to clear the code a bit, we can remove SortCaretWrapper, SortIncentive, TableHeader Style.

ChengYanJin (approved):
LGTM !!

PR #3921, opened 2022-11-24
Bump Loki helm chart to 3.4.2 and Loki image to 2.7.0

TeddyAndrieux:


charts/render.py
```diff
@@ -420,6 +420,7 @@ def __call__(self, parser, args, values, option_string=None):
         "networking.k8s.io/v1/Ingress",
         # Used by ServiceMonitor and other monitoring objects
         "monitoring.coreos.com/v1",
+        "monitoring.coreos.com/v1/ServiceMonitor",
```
gdemonet, line 423: Sad that the `Capabilities.APIVersions` can't deduce from the above that ServiceMonitor is indeed available 😕 I guess it's mostly working for running K8s clusters

JBWatenbergScality (approved):
LGTM

salt/metalk8s/addons/logging/loki/deployed/chart.sls
```diff
@@ -281,26 +225,36 @@ apiVersion: monitoring.coreos.com/v1
...
-    - metalk8s-logging
+  - path: /metrics
+    port: http-metrics
+    relabelings:
+    - replacement: metalk8s-logging/$1
+      sourceLabels:
+      - job
+      targetLabel: job
+    - replacement: loki
+      targetLabel: cluster
```
gdemonet, line 248: Oh, that's nice 👍 we'll stop seeing Loki in the etcd dashboard :)

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -111,6 +111,27 @@ Deploy Kubernetes service config objects:
   - require_in:
     - salt: Deploy Kubernetes objects
 
+{#- Due to a change of Loki StatefulSet labelSelector in 124.1.0, which is immutable field
+    Manually delete the Loki StatefulSet object if dest_version < 124.1.0
+    NOTE: This logic can be removed in `development/126.0` #}
+{%- if salt.pkg.version_cmp(dest_version, '124.1.0') == -1 %}
```
gdemonet, line 117: I guess we could also check if the current version of the statefulset is higher or equal to 124.1 (in case we are re-running the orchestrate), but it's fine, downgrade doesn't need to be super optimized IMO

gdemonet (approved):
LGTM :shipit:

gdemonet:
(lol, too late)

charts/render.py
TeddyAndrieux: Yes exactly, this is just because we use `helm template` with no cluster

salt/metalk8s/orchestrate/downgrade/init.sls
TeddyAndrieux: Yes we could, but anyway, the state is idempotent so it will just do nothing since the object does not exist

PR #3923, opened 2022-11-28
salt: Add a workaround CronJob to delete old Loki chunks

TeddyAndrieux:
Waiting for a proper fix, add a simple CronJob to delete old Loki chunks

### DIFF
--- a/salt/metalk8s/addons/logging/loki/deployed/init.sls
+++ b/salt/metalk8s/addons/logging/loki/deployed/init.sls
@@ -5,6 +5,7 @@ include:
   - .chart
   - .service
   - .datasource
+  - .workaround-job-dep
 
 {#- Due to a change of Loki StatefulSet labelSelector in 124.1.0, which is immutable field
     Manually delete the Loki StatefulSet object if it's an old one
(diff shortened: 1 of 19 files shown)


salt/metalk8s/addons/logging/loki/deployed/init.sls
```diff
@@ -33,3 +34,98 @@ Delete the old Loki StatefulSet:
...
+{%- set loki = salt.metalk8s_service_configuration.get_service_conf('metalk8s-logging', 'metalk8s-loki-config', loki_defaults) %}
+
+{#- Let's consider we only support retention_period and index period in hours for the moment #}
+{%- if loki.spec.config.table_manager.retention_period[-1] != "h" %}
+
+Invalid Loki table manager retention_period, should be in hours:
+  test.fail_without_changes: []
+
+{#- NOTE: Consider the last index is the one currently used #}
+{%- elif loki.spec.config.schema_config.configs[-1].index.period[-1] != "h" %}
```
JBWatenbergScality: ideally we should look for the last config entry using boltdb, but likely ok for now (to. enable us use whatever settings we want for boltdb-shipper which will be the new latest entry)

salt/metalk8s/addons/logging/loki/deployed/pre-upgrade.sls
```diff
@@ -0,0 +1,101 @@
...
+    {#- Let's remove file older than retention time + index time + 3 days #}
+    {%- set max_age = ((
+              (loki.spec.config.table_manager.retention_period[:-1] | int) / 24
+              + (loki.spec.config.schema_config.configs[-1].index.period[:-1] | int) / 24
+            )| int) + 3 %}
+
+Create Loki Cleaner Workaround Job:
+  metalk8s_kubernetes.object_present:
+    - manifest:
+        apiVersion: batch/v1
```
JBWatenbergScality, line 33: The job definition is duplicated, can we easily include it ?

JBWatenbergScality: similar question for the max age computation logic that we could try to factorize

JBWatenbergScality (dismissed):
Approving with few comments, feel free to address them if you can/consider them relevant

salt/metalk8s/addons/logging/loki/deployed/init.sls
TeddyAndrieux: ~WDYM ? We already use the latest entry of the array.
Or do you mean we should check for the `from` date and take the last one ? (I agree but it's a bit of work just for a temporary workaround)~

Edit: Oh ok I miss read the first time, I see I should check for the last one having `boltdb`, got it. I will do it

salt/metalk8s/addons/logging/loki/deployed/pre-upgrade.sls
TeddyAndrieux: It's not exactly the same since the Loki data dir changed in the latest helm-chart `/data/loki => /var/loki`, but yes it could be factorized using a macro if we want to.

I just think about the fact that since we support 1 major upgrade at a time in MetalK8s, it means we need to keep this workaround until `development/126.0` so in 125.x the Loki definition may change again and may need some update on the `CronJob` but not on the pre-upgrade `Job` since this one will always run on older versions, and for the `max_age` to me the computation might be different between pre-upgrade and the `CronJob` once we will have the boltdb-shipper (but right we could go with a macro for the moment and change it later if it's needed also)

No strong opinion here

salt/metalk8s/addons/logging/loki/deployed/init.sls
TeddyAndrieux: Done

salt/metalk8s/addons/logging/loki/deployed/pre-upgrade.sls
TeddyAndrieux: Done

gdemonet (approved):

PR #3953, opened 2023-01-03
shell-ui: Fix the twice logout issue

ChengYanJin:
**Component**: shell-ui

**Context**: 
In some cases, we have to click on logout twice in order to logout.

**Summary**:
- To fix this issue, we introduced a query force to remove the user from the local storage as long as the token is expired.
- Add an event listener to local storage to force the location reload when the user has been removed from the local stroage.

**Acceptance criteria**: 
The logout should function better now!!!


JBWatenbergScality (approved):
LGTM !!!!

PR #3954, opened 2023-01-03
Add promotion GitHub actions workflow

TeddyAndrieux:
The workflow is triggered on a tag push to promote the artifact and publish the shell-ui image on Scality registry.

NOTE: There is no release workflow for the moment


JBWatenbergScality (approved):
LGTM

PR #3973, opened 2023-01-17
Bump Kubernetes to 1.25.5 and update manifest accordingly

TeddyAndrieux:
- `PodSecurityPolicy` API gets removed in 1.25, so make sure not to use them
- `kube-proxy` is now distroless, so rework the way to build the config so that it use an `initContainer` instead of a `sed` at `kube-proxy` container startup
- Bump Kubernetes to `1.25.5` and update manifests (based on kubeadm)
- Bump etcd to `3.5.6` and update manifest (based on kubeadm)
- Bump CoreDNS to `1.9.3` and update manifest (based on kubeadm)


buildchain/buildchain/constants.py
```diff
@@ -17,16 +17,16 @@
 CMD_WIDTH: int = 14
 
 # URLs of the main container repositories.
+K8S_REPOSITORY: str = "registry.k8s.io"
 BITNAMI_REPOSITORY: str = "docker.io/bitnami"
 CALICO_REPOSITORY: str = "docker.io/calico"
-COREDNS_REPOSITORY: str = "k8s.gcr.io/coredns"
+COREDNS_REPOSITORY: str = f"{ K8S_REPOSITORY }/coredns"
```
gdemonet: Did Black run on this file? Or did the rules change recently? IIRC, we would not have had spaces inside the curly braces, but I may be wrong.

gdemonet (dismissed):

buildchain/buildchain/constants.py
TeddyAndrieux: Yes black has ran on it, but apparently he don't care space or no space ....

But indeed normally there is no space, I will change it

gdemonet (approved):

PR #3974, opened 2023-01-19
.github: Add snapshot generation on promotion

TeddyAndrieux:


.github/workflows/e2e-tests.yaml
```diff
@@ -111,6 +121,27 @@ jobs:
...
+      - name: Get full MetalK8s version
+        if: inputs.generate-snapshots
+        run: |
+          source product.txt
+          echo "METALK8S_VERSION=$VERSION" >> $GITHUB_ENV
+      - name: Generate snapshot
+        if: inputs.generate-snapshots
+        uses: ./.github/actions/generate-snapshots
+        with:
+          snapshot-name: "metalk8s-${{ env.METALK8S_VERSION }}-rocky-8-${{ env.NAME }}-simple environment"
```
gdemonet: Do you remember why we introduced such complicated names? The suffix seems really redundant IMO (`simple environment` when we run `single-node`, and `3 nodes` when we run `multi-nodes`), wondering if we couldn't just give them a single "architecture" like in ARTESCA.

TeddyAndrieux: The idea was to have a lot more snapshots in the future, but right I will keep it simple now that we are (kind of) refactoring it

gdemonet (approved):
LGTM, I guess we could have factorized further and have a workflow handling a single spawn+install+tests+snapshot, using it twice (one for single-node, one for multi-nodes), but that's not critical I think

PR #3975, opened 2023-01-25
.github: Add Nightly workflow with snapshot upgrade tests

TeddyAndrieux:


.github/workflows/lifecycle-promoted.yaml
```diff
@@ -0,0 +1,218 @@
...
+          OVH_REGION: ${{ secrets.OVH_REGION }}
+      - name: Spawn cluster with Terraform
+        uses: ./.github/actions/spawn-cluster
+        with:
+          # Parameters
+          CLOUD: ${{ env.CLOUD }}
+          OS: ${{ env.OS }}
+          NAME: ${{ env.NAME }}
+          NODES_COUNT: ${{ matrix.nodes-count }}
+          RESTORE_ENV: ${{ inputs.promoted-version }}-${{ env.OS }}-${{ matrix.NAME }}
```
gdemonet: Shouldn't this show `matrix.name` (lowercase) i.s.o `matrix.NAME`?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+          curl --fail -LO -u ${{ secrets.ARTIFACTS_USER }}:${{ secrets.ARTIFACTS_PASSWORD }}
+          ${{ steps.artifacts.outputs.link }}/product.txt
+      - name: Retrieve current version
+        id: get-version
+        run: cat product.txt >> $GITHUB_OUTPUT
+
+  compute-lifecycle-matrix:
+    runs-on: ubuntu-22.04
+    needs:
+      - retrieve-info
```
gdemonet, line 41: Why do we split `retrieve-info` from this `compute-lifecycle-matrix` job?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  retrieve-info:
+    runs-on: ubuntu-22.04
+    outputs:
+      artifacts-link: ${{ steps.artifacts.outputs.link }}
+      version: ${{ steps.get-version.outputs.version }}
```
gdemonet: Are outputs case-insensitive? Because IIRC, `product.txt` variables are all uppercase

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Fetch tags
+        run: git fetch --tags
+      - name: Retrieve current version
+        run: |
+          echo "major=$(echo "$VERSION" | cut -d'.' -f1)" >> $GITHUB_ENV
+          echo "minor=$(echo "$VERSION" | cut -d'.' -f2)" >> $GITHUB_ENV
+          echo "patch=$(echo "$VERSION" | cut -d'.' -f3)" >> $GITHUB_ENV
```
gdemonet: Why do we need to go this convoluted route (fetching product.txt from artifacts, getting its VERSION, and parse it into components) while we have the `VERSION` file at hand?
Also, if I'm not mistaken, the parsing will always have `-dev` in the resulting `patch` (since we are running the nightly on development branches).

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+      # Every element of the matrix will have
+      # - `type` which is the type of lifecycle (major, minor, patch)
+      # - `version` the promoted version
+      lifecycle-matrix: ${{ steps.get-matrix.outputs.matrix }}
+      artifacts-link: ${{ needs.retrieve-info.outputs.artifacts-link }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Fetch tags
+        run: git fetch --tags
```
gdemonet, line 56: IIUC, this could be done with https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches in the `actions/checkout` step above

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+        id: get-matrix
+        run: |
+          matrix="["
+          for type in patch minor major; do
+            var_name="${type}_version"
+            if [ -n "${!var_name}" ]; then
+              if [ "${#matrix}" -gt "1" ]; then
+                matrix+=","
+              fi
+              matrix+="{'type': '$type', 'version': '${!var_name}'}"
```
gdemonet: That's not valid JSON (single quotes instead of double quotes), are we sure the `fromJSON` works fine with it?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+  lifecycle-promoted:
+    uses: ./.github/workflows/lifecycle-promoted.yaml
+    secrets: inherit
+    needs:
+      - compute-lifecycle-matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJSON(needs.compute-lifecycle-matrix.outputs.lifecycle-matrix) }}
```
gdemonet, line 94: Cool idea, btw :100:

.github/workflows/lifecycle-promoted.yaml
TeddyAndrieux: Hmm indeed, but it works like that, weird, maybe not case sensitive :thinking: 

Anyway I will change it

.github/workflows/nightly.yaml
TeddyAndrieux: Cause `retrieve-info` will be used by some other Nightly job (in another PR)

.github/workflows/nightly.yaml
TeddyAndrieux: Apparently :smile: https://docs.github.com/en/actions/learn-github-actions/variables#naming-conventions-for-configuration-variables

But I will change it, so that it's clearer

.github/workflows/nightly.yaml
TeddyAndrieux: > Why do we need to go this convoluted route (fetching product.txt from artifacts, getting its VERSION, and parse it into components) while we have the `VERSION` file at hand?

We could, I'm using `product.txt` so that we can basically use any artifacts to run it, but here indeed we will likely always be linked to the branch where we are running

> Also, if I'm not mistaken, the parsing will always have `-dev` in the resulting `patch` (since we are running the nightly on development branches).

Right, I will change it (it works when patch > 1 since `$ "1-dev" - 1))` give `0` but when it will be `0-dev` it will not work properly)

I will switch to use the repository version file

.github/workflows/nightly.yaml
TeddyAndrieux: Yes I saw it, but it will retrieve all branches as well, here we only want to have tags and it's not possible with checkout action

.github/workflows/nightly.yaml
TeddyAndrieux: Yes it works https://github.com/scality/metalk8s/actions/runs/3998482145
But indeed it should be double quotes, I will change it

gdemonet: Magic... :magic_wand:

gdemonet (approved):
:shipit:

docs/Dockerfile
```diff
@@ -15,7 +15,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         make \
         plantuml \
         graphviz \
-        python3-buildbot-worker \
```
gdemonet, line 18: :sweat_smile:

PR #3978, opened 2023-01-27
.github: Fix CronsJob workflow by adding Git ACCESS Token

TeddyAndrieux:


.github/workflows/crons.yaml
```diff
@@ -22,4 +23,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Run ${{ matrix.name }}
+        env:
+          GH_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
```
gdemonet: I guess you checked, but they use `GITHUB_TOKEN` instead of `GH_TOKEN` in RING

gdemonet (dismissed):

.github/workflows/crons.yaml
TeddyAndrieux: (I did not tested) But yes I saw it but from the workflow output the suggestion is on `GH_TOKEN` https://github.com/scality/metalk8s/actions/runs/4020367062/jobs/6908217420

Maybe both works, but I will go for `GITHUB_TOKEN`

gdemonet (approved):

PR #3979, opened 2023-01-27
salt: Add retry on DynamicClient creation to avoid flaky

TeddyAndrieux:


gdemonet (approved):
Oh, wouldn't expect this to fail, good catch! How did you see it, btw? Never caught logs about it, personally...

TeddyAndrieux:
> Oh, wouldn't expect this to fail, good catch! How did you see it, btw? Never caught logs about it, personally...

It's the famous 
```
Failed to load ext_pillar metalk8s_endpoints: HTTPSConnectionPool(host='127.0.0.1', port=7443): Max retries exceeded with url: /version (Caused by ProtocolError('Connection aborted.', OSError(0, 'Error')))
```

It was not super easy to catch and reproduce but was able to do it with some loop on applying states (and with this fix it did not fail once while running during the whole night)

PR #3981, opened 2023-01-27
.github: Add missing lifecyle tests in nightly

TeddyAndrieux:


.github/workflows/downgrade-test.yaml
```diff
@@ -0,0 +1,190 @@
...
+          ARTIFACTS_URL: ${{ inputs.to-artifacts-url }}
+      - name: Retrieve target product.txt from artifacts
+        run: >
+          curl --fail -LO -u ${{ secrets.ARTIFACTS_USER }}:${{ secrets.ARTIFACTS_PASSWORD }}
+          ${{ inputs.to-artifacts-url }}/product.txt
+      - name: Save target version and git commit in environment
+        run: |
+          source product.txt
+          echo "target_version=$VERSION" >> $GITHUB_ENV
+          echo "target_git_commit=$GIT" >>$GITHUB_ENV
```
gdemonet, line 120: `$GIT` isn't a commit SHA1, but the output of `git describe`, I didn't know we could just check out this value (TIL)... But I guess you tested that :wink:

.github/workflows/lifecycle-dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+        id: get-url
+        run: |
+          search_url="https://artifacts.scality.net/search/last_success/commit/github/scality/metalk8s/Pre-merge/$dev_commit"
+          artifacts_name=$(curl --fail "$search_url" -u "${{ secrets.ARTIFACTS_USER }}:${{ secrets.ARTIFACTS_PASSWORD }}")
+          if [ -n "$artifacts_name" ]; then
+            echo "artifacts_url=https://artifacts.scality.net/builds/$artifacts_name" >> $GITHUB_OUTPUT
+          else
+            echo "Did not find artifacts for branch ${{ inputs.dev-branch }} - commit $dev_commit." >&2
+            exit 1
+          fi
```
gdemonet: A simpler approach would be to checkout `inputs.dev-branch`, then use the `scality/action-artifacts` action to find the artifacts link

gdemonet (dismissed):
LGTM, the suggestion really isn't necessary

.github/workflows/downgrade-test.yaml
TeddyAndrieux: Indeed but it works :smiley:

.github/workflows/lifecycle-dev.yaml
TeddyAndrieux: Right, I will change this

gdemonet (approved):

PR #3985, opened 2023-01-31
.github: Add single node specific nightly jobs

TeddyAndrieux:


.github/actions/destroy-cluster/action.yaml
```diff
@@ -16,15 +16,18 @@ inputs:
   # Tooling configuration
   TERRAFORM_VERSION:
     description: "Terraform version to use"
-    required: true
+    required: false
+    default: "1.3.6"
```
gdemonet, line 20: Mmmh, I don't really agree with this commit, it seems easier to me to have a single source of truth (in the calling workflow) instead of spreading these defaults in action definitions... Why did you change it?

gdemonet (approved):
You can ignore my comment about the defaults, I think a good solution would rather be to use configuration variables  (see https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository)

.github/actions/destroy-cluster/action.yaml
TeddyAndrieux: A single source of truth in (today) 5 different workflows ? 

To me this version shouldn't be configurable, it's hardcoded somewhere and that's it
With this PR, it's only in 2 different actions (we could indeed move it to a single place, I don't think a github variable is a good idea because we may change this value depending on the branch)

PR #3988, opened 2023-02-02
.github: Add multi nodes specific nightly jobs 

TeddyAndrieux:


.github/actions/spawn-cluster/action.yaml
```diff
@@ -60,11 +60,16 @@ inputs:
     description: Specific arguments to add to the apply command
     required: false
     default: ""
+  skip-prepare:
+    description: Skip the preparation of the worker
+    required: false
+    default: "false"
```
gdemonet, line 66: Would be nicer to use a boolean for this, IMO

.github/workflows/multi-node-test.yaml
```diff
@@ -299,16 +342,9 @@ jobs:
       ## }}
 
       # Generate snapshots (if enabled) {{{
-      - name: Retrieve product.txt from artifacts
-        if: inputs.generate-snapshots
-        run: >
-          curl --fail -LO -u ${{ secrets.ARTIFACTS_USER }}:${{ secrets.ARTIFACTS_PASSWORD }}
-          ${{ inputs.artifacts-url }}/product.txt
       - name: Get full MetalK8s version and total number of nodes
```
gdemonet: (nit) I guess this can be renamed now

gdemonet (dismissed):
Only nitpicks, LGTM otherwise :100:

.github/actions/spawn-cluster/action.yaml
TeddyAndrieux: To me, we cannot have such type in composite actions, it will always be casted to string

.github/workflows/multi-node-test.yaml
TeddyAndrieux: Right, I will do it now

.github/actions/spawn-cluster/action.yaml
gdemonet: You're right, sorry. That's really crappy though

gdemonet (approved):

PR #3995, opened 2023-02-09
Add useNavbar hook and navbarUpdaterComponents support

JBWatenbergScality:
**Component**:

Shell-UI

**Context**: 

Some MicroApps may want to update the navbar entries (e.g.: In the future we would want MetalK8s UI to be able to display the number of currently raised alerts next to the alerts entries, but also some MicroApps rely on finer grained access control than user groups and would want to remove/add entries to the navbar based on this finer access control, a last use case would be to enable custom features provided by those MicroApps).

In order to achieve this we need to introduce a _federable_ hook, so called `useNavbar`, providing access to the navbar configuration. It would list detected navbar entries and enable MicroApps to customise them. 
This additional hook is also introduced with a new `NavbarUpdaterComponents` notion that can optionally be configured in MicroApps build time web fingers configuration (`{microAppEndpoint}/.well-known/micro-app-configuration`). These `NavbarUpdaterComponents` are headless (aka invisible, they render nothing) components that ShellUI would federate from the micro apps once it got fully initiated. Those components can then use above `useNavbar` hook federated from Shell UI to update the navbar.

### DIFF
--- a/shell-ui/package-lock.json
+++ b/shell-ui/package-lock.json
@@ -1634,7 +1634,6 @@
     },
     "node_modules/@fortawesome/fontawesome-free": {
       "version": "5.15.3",
-      "dev": true,
       "hasInstallScript": true,
       "license": "(CC-BY-4.0 AND OFL-1.1 AND MIT)",
       "engines": {
@@ -20464,8 +20463,7 @@
       "version": "0.2.36"
     },
     "@fortawesome/fontawesome-free": {
-      "version": "5.15.3",
-      "dev": true
+      "version": "5.15.3"
     },
     "@fortawesome/fontawesome-svg-core": {
       "version": "1.2.36",
@@ -21102,7 +21100,8 @@
     },
     "@mdx-js/react": {
       "version": "1.6.22",
-      "dev": true
+      "dev": true,
+      "requires": {}
     },
     "@mdx-js/util": {
       "version": "1.6.22",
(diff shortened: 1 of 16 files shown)


gdemonet:
Hey @JBWatenbergScality 

I haven't had much time to look at your PR in details, but as explained in another discussion, I am a bit confused about the chosen approach.

IIUC, the current approach can be summed up as:

- ShellUI owns the navbar component, builds a bunch of entries and loads various micro apps depending on its configuration and the various micro app webfingers
- It exposes a hook for micro apps to consume (akin to the authn provider) for updating the navbar
- Each micro app now includes a headless component, (always?) rendered by ShellUI so they can use the hook and dynamically update the navbar

I understand the *why* (each micro app knows about the entries it wants to show in the navbar better than ShellUI), but I don't get the *how*. ShellUI owns the navbar, has a configuration to define some links in there, but somehow each individual micro app can override this? What if micro apps disagree?

Random example: two micro apps want to manage a main link called "Details". How do they coordinate? The current hooks expose `setMainLinks(links: Link[]): void` and `getLinks(): NavbarLinks`. These micro apps then need to first get the current state through `getLinks`, compute their changes (adding or updating the found "Details" link), and send them with `setMainLinks`. But nothing prevents them from continuously overwriting the changes from their counterpart.

IMO, the benefit of ShellUI is to manage the orchestration of these micro apps without them having to coordinate in any way. I don't know if we have feasibility constraints, but in my mind, we should try to build something where:

- ShellUI still owns the navbar, and still manages micro apps by looking at its configuration and the micro app webfingers
- Each micro app exposes, in addition to the main renderable, a "navbar library" (TBH I don't know if it should be a component, a set of components, hooks, or whatnot) which has multiple purposes:
    - It has a `registerLinks(): {[key: string]: LinkSpec}` method, which ShellUI calls to get placeholders for the links this micro apps wants to show (optionally, they may not be displayed immediately, nor for every user) - at this point, conflicts can be identified early, because two micro apps attempting to register the "same link" will be rejected
    - It has a `LinkManager` component (I guess, since we don't really have reducers nowadays), which role is to update the registered links for a single micro app as it receives state updates
    - It can include a `StateManager` (would also need its own registering mechanism), if the micro app needs to have a state continuously updated even when the main renderable is not loaded
- Then, ShellUI defines a state (context?) for each micro app, mounts the StateManager if defined, mounts the LinkManager by giving it this state, and exposes a hook for the corresponding micro app to update the state when mounted

I'd be happy to discuss this further, if I wasn't clear enough :wink:
Also, I won't block this PR, we can always move to another strategy in the future, just sharing my concerns with the current approach.
We could also settle for a middle-ground, where the hooks exposed to micro apps are scoped, and no conflict can occur (i.e. each link is owned by a single app, and only this app can change it)

ChengYanJin (approved):
LGTM !

PR #4000, opened 2023-02-24
Move Control Plane Ingress VIP management to MetalK8s Operator instead of MetalLB

TeddyAndrieux:
- Fix some flaky tests
- Rework the Sub reconciler handling in MetalK8s Operator
- Rework VirtualIPPool to be less "WP Ingress linked"
- Move Control Plane Ingress IP handling in MetalK8s Operator
- Properly handle upgrade and migration from MetalLB
- Remove everything linked to MetalLB


tests/post/steps/test_logging.py
```diff
@@ -191,7 +191,7 @@ def _check_log_line_exists():
     utils.retry(
         _check_log_line_exists,
         times=60,
-        wait=5,
+        wait=10,
```
gdemonet, line 194: https://youtu.be/h1cEaob7Iew

operator/pkg/controller/clusterconfig/controller.go
```diff
@@ -11,20 +12,19 @@ import (
...
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/go-logr/logr"
 	metalk8sscalitycomv1alpha1 "github.com/scality/metalk8s/operator/api/v1alpha1"
+	"github.com/scality/metalk8s/operator/pkg/controller/clusterconfig/workloadplane"
 	"github.com/scality/metalk8s/operator/pkg/controller/utils"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
gdemonet: (nit) We should stick to one single way of ordering imports, these diffs are confusing for no good reason. IIUC, `gofmt` sorts alphabetically, and `goimports` groups them into (stdlib) - (external) - (current pkg). I personally prefer the latter. But in any case, this re-ordering doesn't make much sense to me (`ctrl` is above but `client` gets moved down here, while they come from the same package).

operator/pkg/controller/clusterconfig/controller.go
```diff
@@ -114,26 +116,58 @@ func (r *ClusterConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-			r.instance.SetReadyCondition(metav1.ConditionFalse, "Reconciling", "Reconciliation in progress")
-			return result.GetResult()
-		}
+	defer r.client.Status().Update(ctx, instance)
+
+	subReconcilers := map[string]interface {
+		Load(*utils.ObjectHandler, logr.Logger, *metalk8sscalitycomv1alpha1.ClusterConfig)
+		GetLogger() logr.Logger
+		Reconcile(context.Context) utils.SubReconcilerResult
+	}{
```
gdemonet: Is it good practice to define the interface in the middle of a function? I'd rather have something easy for other modules to import and rely on when implementing sub-reconcilers, that'd make things more obvious when changing the interface.

```go
type SubReconciler interface {
	Load(*utils.ObjectHandler, logr.Logger, *metalk8sscalitycomv1alpha1.ClusterConfig)
	GetLogger() logr.Logger
	Reconcile(context.Context) utils.SubReconcilerResult
}
```

---

_EDIT_: So I just realized we can't tell the compiler to check that a type implements a given interface, so all we'd get is an error wherever we try to use a value as this interface. Great 😮‍💨 The only benefit of naming is the error message is clearer: instead of getting `interface{Load(...), GetLogger(...), Reconcile(...)}` we'll have for instance `does not implement SubReconciler (missing method Load)`

A way to workaround this (and bubble-up the type checking in the module defining the interface), would be to cheat and try and assign a variable with this interface as a type annotation. Sad...

operator/api/v1alpha1/clusterconfig_types_test.go
```diff
@@ -9,7 +11,7 @@ import (
 var _ = Describe("ClusterConfig", func() {
 	Describe("GetSetCondition", func() {
 		It("can add and get a condition", func() {
-			now := metav1.Now()
+			now := metav1.Now().Add(-time.Millisecond)
```
JBWatenbergScality, line 16: Just wondering can't we just freeze the time for the tests so that the lastTransitionTime value becomes predictable and can be exactly matched in the tests below.

JBWatenbergScality: Also the -1 millisecond logic is a bit duplicated al over the tests.

operator/api/v1alpha1/clusterconfig_types.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	vIPConfiguredConditionName = "VirtualIPPool" + configuredConditionName
-	vIPReadyConditionName      = "VirtualIPPool" + readyConditionName
+	wPVIPConfiguredConditionName = "WorkloadPlaneVirtualIPPool" + configuredConditionName
```
JBWatenbergScality, line 26: Note to self: how do we handle upgrades as this changed ? Shouldn't we bump the CR version as it is "breaking" ?

operator/api/v1alpha1/virtualippool_types.go
```diff
@@ -30,6 +30,31 @@ type SpreadConstraintSpec struct {
...
+	// +kubebuilder:validation:Enum={"HTTP", "HTTPS"}
+	Scheme string `json:"scheme"`
+	// The port to do the HTTP request
+	// +optional
+	// +kubebuilder:default=443
+	Port int `json:"port"`
+	// Path for the HTTP request
+	// +optional
+	Path string `json:"path"`
+}
```
JBWatenbergScality: Note to self: For HTTPS do we need an additional ca bundle link ?

operator/pkg/controller/clusterconfig/controller.go
```diff
@@ -114,26 +116,58 @@ func (r *ClusterConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+			if res := subrec.Reconcile(ctx); res.ShouldReturn() {
+				mutex.Lock()
+				if res.IsSuperior(result) {
+					result = res
+				}
+				mutex.Unlock()
+			}
+		}(key)
+	}
```
gdemonet: LGTM, but would be nice to have this piece of logic (orchestrating subreconcilers) in a separate function/package, so we can test it easily.

operator/pkg/controller/clusterconfig/workloadplane/virtualippool.go
```diff
@@ -14,24 +14,37 @@ import (
 	"github.com/scality/metalk8s/operator/pkg/controller/utils"
 )
 
+type VirtualIPPoolReconciler struct {
+	handler *utils.ObjectHandler
+	logger  logr.Logger
+
+	instance *metalk8sscalitycomv1alpha1.ClusterConfig
```
gdemonet, line 21: If each reconciler holds a reference to this instance, and they all run concurrently, won't we have data races? Seems fishy to me, wondering if they shouldn't rather receive a copy 😕

operator/pkg/controller/utils/reconcile_test.go
```diff
@@ -48,6 +48,64 @@ var _ = Describe("SubReconciler tests", func() {
...
+			r := utils.NothingToDo()
+
+			Expect(r.IsSuperior(utils.NeedRequeue(nil))).To(BeFalse())
+		})
+
+		It("is not superior to EndReconciliation", func() {
+			r := utils.NothingToDo()
+
+			Expect(r.IsSuperior(utils.NeedEndReconciliation())).To(BeFalse())
+		})
```
gdemonet: That's quite verbose and not ideal to read, could we consider refactoring with [`DescribeTable`](https://onsi.github.io/ginkgo/#table-specs)?

gdemonet (commented):
(first batch, only read the first 3 commits)

operator/api/v1alpha1/clusterconfig_types.go
```diff
@@ -55,11 +100,15 @@ type ClusterConfigStatus struct {
...
 	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+
+	// Control Plane Information
+	ControlPlane ControlPlaneStatus `json:"controlPlane,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:scope=Cluster,shortName=cc
+//+kubebuilder:printcolumn:name="Control-Plane-Url",type="string",JSONPath=".status.controlPlane.ingress.endpoint",description="The URL to reach the Control Plane Ingress"
```
gdemonet, line 113: ~Let's show "Control Plane URL" as a name, as this is displayed to users~

Seen in peer review, we want to keep dashes so it's obvious in `kubectl` output that it's a single column.

tests/post/steps/test_logging.py
TeddyAndrieux: It's not linked to this PR but was to avoid some flakies I got when testing in CI

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
```diff
@@ -37,6 +42,33 @@ spec:
           spec:
             description: ClusterConfigSpec defines the desired state of ClusterConfig
             properties:
+              controlPlane:
+                description: Information about the Workload Plane.
```
gdemonet: (this was seen in peer-review)

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: Right, normally I let `goimports` handle them but maybe there is some piece of the code that get re-formated with `gofmt`

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: Indeed, ok yes I can define the interface explicitly somewhere else

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: You mean the full "for" loop, yes ok we could

operator/pkg/controller/clusterconfig/workloadplane/virtualippool.go
TeddyAndrieux: Each reconciler should only "manage" his own part of the `status` so we shouldn't have any issue.

The issue about copying it is that each sub reconcile may edit the `status` and add the end (with `defer` from the "main" controller) we actually push the changes, so if we do a copy we will have to "reconcile" the `status` content and it might be a mess (especially if we consider that 2 different sub reconcile may edit the same status)

operator/pkg/controller/utils/reconcile_test.go
TeddyAndrieux: Yes likely (not sure it will be easier to read but at least less code duplication)

operator/api/v1alpha1/virtualippool_types.go
```diff
@@ -30,6 +30,31 @@ type SpreadConstraintSpec struct {
 	TopologyKey string `json:"topologyKey"`
 }
 
+type HttpGetSpec struct {
+	// The host name to do the HTTP request
+	// (default to keepalived Pod IP)
+	// +optional
+	Host string `json:"host,omitempty"`
```
TeddyAndrieux: ipaddress

operator/api/v1alpha1/clusterconfig_types.go
```diff
@@ -39,11 +68,27 @@ type ClusterConfigSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// Information about the Workload Plane.
```
TeddyAndrieux: copy pasta

operator/pkg/controller/clusterconfig/controlplane/ingress.go
```diff
@@ -0,0 +1,273 @@
...
+// Reconcile the Control Plane Ingress
+func (r *IngressReconciler) Reconcile(ctx context.Context) utils.SubReconcilerResult {
+	// Reconcile the Ingress namespace
+	namespaceInstance := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespaceName}}
+
+	// Create or update the namespace if needed
+	// NOTE: We treat the namespace alone has nothing else can be created if this one
+	// do not exist
+	// TODO: Use CreateOrUpdateOrDelete, It's not done yet as we do not want to Own the namespace
+	// since we not yet manage it entirely, just ensure that it exist for the moment
```
TeddyAndrieux: Consider also wp ingress in the same namespace

images/metalk8s-keepalived/generate-config.py
```diff
@@ -14,6 +15,12 @@
 
 
 def get_interface_from_ip(ip):
+    # Check if the IP does not already sit on an interface (if yes then use it)
+    for interface in netifaces.interfaces():
+        for link in netifaces.ifaddresses(interface).get(netifaces.AF_INET) or []:
+            if link.get("addr") == ip:
+                return interface
+
     # NOTE: We do not have any easy way (without an external lib)
```
gdemonet, line 24: We included said external lib, should we still keep the previous logic? What about the bug with loopback you mentioned?

buildchain/buildchain/salt_tree.py
```diff
@@ -352,14 +352,15 @@ def task(self) -> types.TaskDict:
     targets.TemplateFile(
         task_name="metalk8s-operator.sls",
         source=constants.ROOT.joinpath(
-            "salt/metalk8s/addons/metalk8s-operator/deployed.sls.in"
+            "salt/metalk8s/addons/metalk8s-operator/deployed/chart.sls.in"
         ),
         destination=constants.ISO_ROOT.joinpath(
-            "salt/metalk8s/addons/metalk8s-operator/deployed.sls"
+            "salt/metalk8s/addons/metalk8s-operator/deployed/chart.sls"
```
gdemonet: Why would we call this "chart.sls", it's not rendered from a Helm chart, and won't ever be, right? I understand we want to split, but maybe it could make more sense to give it a relevant name, such as "bundle.sls" or "manifests.sls".

operator/api/v1alpha1/clusterconfig_types_test.go
TeddyAndrieux: As discussed yesterday, we might be able to build something smarter by mocking the `metav1.Now()` but we will not invest time on it in this PR

operator/api/v1alpha1/clusterconfig_types.go
TeddyAndrieux: Already discussed yesterday (It's only a condition name change that (AFAIK) is not used, so let's just rename it)

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Maybe something to consider in the future, but not for the moment

salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls
```diff
@@ -30,9 +30,13 @@ Create Control-Plane Ingress server private key:
     'nginx-ingress-control-plane.metalk8s-ingress',
     'nginx-ingress-control-plane.metalk8s-ingress.svc',
     'nginx-ingress-control-plane.metalk8s-ingress.svc.' ~ coredns.cluster_domain,
-    salt.metalk8s_network.get_control_plane_ingress_ip(),
 ] %}
 
+{%- set cp_ingress_ip = salt.metalk8s_network.get_control_plane_ingress_ip() %}
+{%- if salt.metalk8s_network.get_control_plane_ingress_ip() %}
```
gdemonet: this could just be

```sls
{%- if cp_ingress_ip %}
```

tests/post/steps/test_ingress.py
```diff
@@ -655,35 +729,23 @@ def check_cp_ingress_ip(context, control_plane_ingress_ip, ip):
...
-        namespace="metalk8s-ingress",
-        field_selector=",".join(field_selectors),
+def get_node_hosting_cp_ingress_vip(host, ssh_config, vip):
+    node_ips = json.loads(
+        utils.run_salt_command(
+            host,
+            ["salt", "--static", "--out=json", "'*'", "network.ip_addrs"],
+            ssh_config,
+        ).stdout
     )
```
gdemonet: Worth sharing in a utility function, since we have the same call for workload plane

.github/actions/enable-cp-ingress-managed-vip/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+      uses: ./.github/actions/run-command-ssh
+      with:
+        COMMAND: >
+          sudo salt-call state.sls metalk8s.kubernetes.apiserver
+    - name: Reconfigure Control Plane components
+      uses: ./.github/actions/run-command-ssh
+      with:
+        COMMAND: |
+          SALT_MASTER=\$(sudo crictl ps --label="io.kubernetes.container.name=salt-master" -q)
+          sudo crictl exec \$SALT_MASTER salt-run state.sls metalk8s.orchestrate.update-control-plane-ingress-ip saltenv=metalk8s-$METALK8S_VERSION
```
gdemonet: Depending on this `METALK8S_VERSION` env variable to be available is not ideal IMO, an input would make things clearer

gdemonet (commented):
Last few comments, but LGTM overall 👌

operator/api/v1alpha1/clusterconfig_types.go
TeddyAndrieux: Yes, the current output
```console
$ kubectl get cc
NAME   CONTROL-PLANE-URL
main   https://192.168.20.254:8443
```

images/metalk8s-keepalived/generate-config.py
TeddyAndrieux: This is the logic to handle the "loopback" bug, we see if there is an interface that already has the VIP.

AFAIK, this lib cannot be used to get the route.

The logic is if the VIP already sits on an interface (it usually shouldn't happen but it could if for whatever reason keepavlied get killed) we keep it, otherwise we "search" it based on the routes

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Yes right, ok I'm fine about `manifests.sls`

salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls
TeddyAndrieux: Right, good catch

.github/actions/enable-cp-ingress-managed-vip/action.yaml
TeddyAndrieux: Yes but would always need input (that might be no longer needed at some point once (/if) we will have the full Ingress management in the operator) but right ok let's move it as input

TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/4132813829c2f2aa465abfce7eeb247569111f11

tests/post/steps/test_ingress.py
TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/4132813829c2f2aa465abfce7eeb247569111f11

salt/metalk8s/addons/nginx-ingress-control-plane/certs/server.sls
TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/4132813829c2f2aa465abfce7eeb247569111f11

buildchain/buildchain/salt_tree.py
TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/4132813829c2f2aa465abfce7eeb247569111f11

operator/config/crd/bases/metalk8s.scality.com_clusterconfigs.yaml
TeddyAndrieux: Fixed in https://github.com/scality/metalk8s/pull/4000/commits/827fb5dff969e89c7e4c129e69421e319871fe69

operator/pkg/controller/clusterconfig/controlplane/ingress.go
TeddyAndrieux: Fixed in https://github.com/scality/metalk8s/pull/4000/commits/827fb5dff969e89c7e4c129e69421e319871fe69

operator/api/v1alpha1/clusterconfig_types.go
TeddyAndrieux: Fixed in https://github.com/scality/metalk8s/pull/4000/commits/827fb5dff969e89c7e4c129e69421e319871fe69

operator/api/v1alpha1/virtualippool_types.go
TeddyAndrieux: Fixed in https://github.com/scality/metalk8s/pull/4000/commits/0166e262ae779c9e8bcd2104c0f6e798cfbab224

operator/pkg/controller/utils/reconcile_test.go
TeddyAndrieux: I did not use DescribeTable since I did not find any way to have it clearer than now, but I moved to a single `It` per SubReconcilerResult type for all "superior"

See https://github.com/scality/metalk8s/pull/4000/commits/2fbab1f825034ac1e3c199008e010affe5468b0a

operator/pkg/controller/clusterconfig/workloadplane/virtualippool.go
TeddyAndrieux: So I added a Mutex logic to handle Conditions for ClusterConfig

See https://github.com/scality/metalk8s/pull/4000/commits/2fbab1f825034ac1e3c199008e010affe5468b0a

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/2fbab1f825034ac1e3c199008e010affe5468b0a

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: Done in https://github.com/scality/metalk8s/pull/4000/commits/2fbab1f825034ac1e3c199008e010affe5468b0a

operator/pkg/controller/clusterconfig/controller.go
TeddyAndrieux: I changed the imports order (and explicitly used `goimports` in my vscode :smiley: )

operator/pkg/controller/utils/reconcile_test.go
```diff
@@ -160,27 +116,12 @@ var _ = Describe("SubReconciler tests", func() {
 			assertEndReconciliation(r.GetResult())
 		})
 
-		It("is superior to NothingToDo", func() {
+		It("is only superior to Everything", func() {
```
gdemonet: nit (maybe change it when squashing):

```diff
-		It("is only superior to Everything", func() {
+		It("is superior to everything", func() {
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -303,6 +297,22 @@ func (r *VirtualIPPoolReconciler) mutateConfigMap(obj *corev1.ConfigMap) error {
...
+		if ip == "" {
+			ip = "{NODE_IP}"
+		}
+		desired.Healthcheck = fmt.Sprintf(
+			"%s://%s:%d%s",
+			strings.ToLower(healthInfo.Scheme),
+			ip,
+			healthInfo.Port,
+			healthInfo.Path,
+		)
```
gdemonet: Since `healthInfo.Path` is still a free-form string, we should be more careful... IMO, using [`net/url`](https://pkg.go.dev/net/url) could be a better approach, since we have all the parts we need to build a `URL` instance. The only nuance would be about using the `{NODE_IP}` placeholder, but we could work around this by using a reserved hostname, for instance, and replacing it with the desired environment variable in our Jinja template.

gdemonet (commented):
Still unsure about our risk of script injection :/ Although we don't really document how to use it, nor expose this functionality to users, so I guess it's mostly safe. Could be reworked in a follow-up ticket, TBH.

operator/pkg/controller/virtualippool/controller.go
TeddyAndrieux: Right ok

TeddyAndrieux: Fixed in https://github.com/scality/metalk8s/pull/4000/commits/be84bb1cd7b81a1695c898f4426b1891a415c3b9

gdemonet (dismissed):
LGTM! :shipit:

.github/actions/enable-cp-ingress-managed-vip/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+      uses: ./.github/actions/run-command-ssh
+      with:
+        COMMAND: >
+          sudo salt-call state.sls metalk8s.kubernetes.apiserver
+    - name: Reconfigure Control Plane components
+      uses: ./.github/actions/run-command-ssh
+      with:
+        COMMAND: |
+          SALT_MASTER=\$(sudo crictl ps --label="io.kubernetes.container.name=salt-master" -q)
+          sudo crictl exec \$SALT_MASTER salt-run state.sls metalk8s.orchestrate.update-control-plane-ingress-ip saltenv=metalk8s-${{ inputs.metalk8s-version }}
```
JBWatenbergScality, line 42: Nit: Just for my own information, is there a specific reason why we perform this through crictl instead of a simple kubectl exec ?

JBWatenbergScality (dismissed):

.github/actions/enable-cp-ingress-managed-vip/action.yaml
gdemonet: The orchestrate may restart kube-apiserver :wink:

TeddyAndrieux: The initial reason was because this orchestrate may restart the APIServer (and then break the kubectl command) but since nowadays we actually do the "local" APIServer first we shouldn't have any issue

JBWatenbergScality (approved):

gdemonet (approved):

PR #4017, opened 2023-03-07
Bump kube-prometheus-stack to 45.5.0

gdemonet:
Update the charts with:

```
helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
helm repo update
rm -rf charts/kube-prometheus-stack
helm fetch -d charts --untar prometheus-community/kube-prometheus-stack
```

Also bump a number of components:
- Prometheus to 2.42.0
- Thanos to 0.30.2
- Grafana to 9.3.8
- kiwigrid/k8s-sidecar to 1.22.3
- kube-state-metrics to 2.8.0
- node-exporter to 1.5.0
- prometheus-operator to 0.63.0

The chart.sls was re-rendered with:

```
./doit.sh codegen:chart_kube-prometheus-stack
```

Since we bumped Thanos, we also re-render its own chart (without
updating, since it did not change since the last update) with:

```
./doit.sh codegen:chart_thanos
```

Important note: Alertmanager configuration was updated by hand,
and a note was added to try and remind maintainers of doing it in the
future. This should help making the "InfoInhibitor" alert more useful.


charts/kube-prometheus-stack.yaml
```diff
@@ -12,7 +12,8 @@ commonLabels:
 alertmanager:
   alertmanagerSpec:
     image:
-      repository: '__image__(alertmanager)'
+      registry: '__var__(repo.registry_endpoint)'
+      repository: '__image_no_reg__(alertmanager)'
```
TeddyAndrieux, line 20: Sad, but ok, yes it's needed

But cannot we set this registry only once in `global` ?

charts/kube-prometheus-stack/README.md
```diff
@@ -80,6 +80,174 @@ _See [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) for command documen
...
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.59.1/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.59.1/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.59.1/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.59.1/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml
+```
+
+Starting from prometheus-node-exporter version 4.0.0, the `node exporter` chart is using the [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). Therefore you have to delete the daemonset before you upgrade.
```
TeddyAndrieux, line 200: :eyes: 

We need to handle it

TeddyAndrieux (commented):
Really really hard to review

Just by looking at the changes on the salt side it looks good to me (except for the upgrade handling)

CHANGELOG.md
```diff
@@ -56,13 +56,55 @@
   [v2.35.3](https://github.com/dexidp/dex/releases/tag/v2.35.3)
   (PR[#4012](https://github.com/scality/metalk8s/pull/4012))
 
+- Bump kube-prometheus-stack chart version to
+  [45.5.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-prometheus-stack-45.5.0)
+  (PR[#4017](https://github.com/scality/metalk8s/pull/4011))
```
TeddyAndrieux: typo: 
```suggestion
  (PR[#4017](https://github.com/scality/metalk8s/pull/4017))
```

CHANGELOG.md
```diff
@@ -56,13 +56,55 @@
...
 - Fix a bug in Workload Plane Ingress Virtual IPs that make the DaemonSet Pod
   goes in CrashLoopBackOff because keepalived wrongly pick the loopback
   interface instead of the actual interface
   (PR[#4000](https://github.com/scality/metalk8s/pull/4000))
 
+### Removals
+
+- Some metrics were removed when bumping kube-prometheus-stack chart to
+  [45.5.0](https://github.com/prometheus-community/helm-charts/releases/tag/kube-prometheus-stack-45.5.0)
+  (PR[#4017](https://github.com/scality/metalk8s/pull/4011)):
```
TeddyAndrieux: ditto

charts/kube-prometheus-stack/README.md
gdemonet: Oh yes good catch, didn't check upgrade notes 🤦

charts/kube-prometheus-stack.yaml
gdemonet: Right will do, found out too late about this one 😇

gdemonet: Argh no I can't do that because the kube-state-metrics subchart doesn't use this value correctly:

```
image: "{{ .Values.global.imageRegistry | default .Values.image.repository }}:{{ .Values.image.tag | ...
```

:sob: 

I'll submit a PR over there to have it fixed

charts/kube-prometheus-stack/values.yaml
```diff
@@ -897,7 +1057,31 @@ kubelet:
...
+        action: drop
+        regex: 'container_(file_descriptors|tasks_state|threads_max)'
+      # Drop container spec metrics that overlap with kube-state-metrics.
+      - sourceLabels: [__name__]
+        action: drop
+        regex: 'container_spec.*'
+      # Drop cgroup metrics with no pod.
+      - sourceLabels: [id, pod]
+        action: drop
+        regex: '.+;'
```
gdemonet, line 1084: This causes the current prometheus-adapter `resourceRules.(cpu|memory).nodeQuery` to not hit anything, because we dropped the metrics for nodes (`pod="", id="/"`).

It appears the goal is to now rely on more efficient node-exporter metrics (see https://github.com/kubernetes-sigs/prometheus-adapter/pull/516 and https://github.com/prometheus-community/helm-charts/pull/2827), but we're now hitting an issue with labels: our node-exporter metrics don't have a "node" label, and that's not good when trying to map to `/metrics.k8s.io/v1beta1/nodes`.

We will need to fix this label issue (TBH, would be much simpler to explore and query, e.g. from our UI), but I'm not sure how involved this will be. For now, I'm considering a temporary workaround by keeping these metrics around, with a follow-up ticket.

gdemonet: Here's the follow-up https://github.com/scality/metalk8s/issues/4018

gdemonet: And the workaround: https://github.com/scality/metalk8s/pull/4017/commits/97ce7c66bedb148d381a9f1011b4afdfc70b6106

salt/metalk8s/addons/prometheus-operator/pre-upgrade.sls
```diff
@@ -0,0 +1,29 @@
...
+        - kind: StatefulSet
+        - name: prometheus-operator-prometheus-node-exporter
+        - namespace: metalk8s-monitoring
+
+{%- else %}
+
+Nothing to do before upgrading for prometheus-operator:
+  test.nop: []
+
+{%- endif %}
```
TeddyAndrieux: Usually I prefer to do it directly in the `init.sls` of the deployed states (so that we only remove it when we are actually deploying the new one)

And we also need to do it for downgrade btw
(https://github.com/scality/metalk8s/commit/f30023586692d4672dc5d2a4c0d12c7cad987bab)

gdemonet: Ah yes, downgrade 😬 On it

gdemonet: Done

TeddyAndrieux (approved):
:shipit:

salt/metalk8s/addons/prometheus-operator/deployed/init.sls
```diff
@@ -8,3 +8,28 @@ include:
...
+        node_exporter_sts["metadata"]["labels"]["metalk8s.scality.com/version"],
+        "125.0.0"
+    ) == -1 %}
+
+Delete old node-exporter StatefulSet:
+    metalk8s_kubernetes.object_absent:
+        - apiVersion: apps/v1
+        - kind: StatefulSet
+        - name: prometheus-operator-prometheus-node-exporter
+        - namespace: metalk8s-monitoring
```
TeddyAndrieux, line 33: Oh we should `wait` for deletion here and `require_in` the `sls` otherwise it will fail on the first upgrade run

TeddyAndrieux (changes requested):

salt/metalk8s/addons/prometheus-operator/deployed/init.sls
gdemonet: Good catch, eagle eyes :eagle: !

TeddyAndrieux (approved):

PR #4019, opened 2023-03-09
salt,docs,tests: Allow to disable CoreDNS host forwarding

TeddyAndrieux:


docs/installation/bootstrap.rst
```diff
@@ -206,6 +207,15 @@ defaults kubernetes configuration.
...
             groupsClaim: <Groups Claim>
 
+  From ``coreDNS`` section you can disable the ``hostForward`` so that the DNS
+  request from the Pod inside Kubernetes are not forwarded to the configured
+  nameservers from the host.
+
+    .. note::
+
+      It means pod running in Kubernetes will not be able to resolve any name
+      that are not in Kubernetes.
```
gdemonet, line 217: There still is a way to make it work with `ExternalName` Service objects 😉

salt/metalk8s/kubernetes/coredns/files/coredns-deployment.yaml.j2
```diff
@@ -17,6 +17,11 @@ spec:
       {{ label_selector | yaml(False) | indent(8) }}
   template:
     metadata:
+      annotations:
+        # NOTE: Add annotation for config checksum, so that Pod get restarted on
+        # ConfigMap change
+        checksum/config: __slot__:salt:metalk8s_kubernetes.get_object_digest(kind="ConfigMap",
+          apiVersion="v1", namespace="kube-system", name="coredns", path="data")
```
gdemonet, line 24: 💯

salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -8,6 +8,12 @@
 include:
   - .running
 
+Ensure resolv config file exists:
+  file.managed:
+    - name: {{ kubelet.config.resolvConf }}
+    - create: true
+    - replace: false
```
gdemonet, line 15: Can you explain this part? I see we're defining it to be `/etc/resolv.conf`, and we're passing it to kubelet, but not sure what it does... And also is it safe to just create an empty one? Like, how would this interact with the rest of the system services?

gdemonet (dismissed):
LGTM, would just like a bit more details (for posterity) about kubelet needing a resolv.conf :)

docs/installation/bootstrap.rst
TeddyAndrieux: If it's in a Service as `ExternalName` then it's in Kubernetes :smiley:

salt/metalk8s/kubernetes/kubelet/standalone.sls
TeddyAndrieux: It just creates an empty `resolv.conf` because Kubelet needs a file otherwise it fails, and for my testing it works well with an empty file

gdemonet: OK, let's keep the conversation open for future readers, but I trust your testing 😉

gdemonet (approved):

PR #4020, opened 2023-03-10
salt: Fix deletion of old node-exporter DaemonSet

TeddyAndrieux:
(Typo: it's not a StatefulSet but a DaemonSet)


gdemonet (dismissed):
Argh sorry 😔

gdemonet (approved):

PR #4022, opened 2023-03-13
ARTESCA-7133 exit using die in _save_cp

eg-ayoub:


TeddyAndrieux (commented):
Just missing a changelog entry

PR #4032, opened 2023-03-23
ARTESCA-7317: enable webhook on deployment

lamphamsy:
**Component**: 'salt'

**Context**:
To enable webhook on deployment. 

**Summary**:
Add volume to use secret to configure webhook in a deployment. This volume is configured when solution.sh is executed with the flag `-w` or `--webhook`.

**Acceptance criteria**:

### DIFF
--- a/salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml.j2
+++ b/salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml.j2
@@ -49,7 +49,18 @@ spec:
           volumeMounts:
             - name: operator-config
               mountPath: /etc/config
+{%- if webhook_enabled is defined and webhook_enabled %}
+            - name: cert
+              mountPath: /cert
+              readOnly: true
+{%- endif %}
       volumes:
         - name: operator-config
           configMap:
             name: {{ solution }}-operator-config
+{%- if webhook_enabled is defined and webhook_enabled %}
+        - name: cert
+          secret:
+            defaultMode: 420
+            secretName: {{ solution }}-cert
+{%- endif %}
--- a/salt/tests/unit/formulas/config.yaml
+++ b/salt/tests/unit/formulas/config.yaml
@@ -753,6 +753,15 @@ metalk8s:
                   image_name: example-operator
                   image_tag: "1.2.3"
 
+              "Example Solution v1.2.3 with webhook":
+                extra_context:
+                  <<: *base_context_solution_operator_files
+                  repository: >-
+                    metalk8s-registry-from-config.invalid/example-solution-1.2.3
+                  image_name: example-operator-webhook
+                  image_tag: "1.2.3"
+                  webhook_enabled: true
+
           role_binding.yaml.j2:
             _cases:
               "Example Solution v1.2.3 (see ../../prepare-environment.sls)":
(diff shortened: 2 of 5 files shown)


lavigne958 (commented):
:eyes: :+1:

scripts/solutions.sh
```diff
@@ -437,7 +443,8 @@ add_solution() {
     local -ra pillar=(
         "{"
         "  'orchestrate': {"
-        "    'env_name': '$NAME'"
+        "    'env_name': '$NAME',"
+        "    'webhook_enabled': '$WEBHOOK_ENABLED'"
```
TeddyAndrieux: I think you want to remove the quote otherwise it will always be "true" (since `bool('false') == True`)

```suggestion
        "    'webhook_enabled': $WEBHOOK_ENABLED"
```

salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml.j2
```diff
@@ -49,7 +49,18 @@ spec:
...
       volumes:
         - name: operator-config
           configMap:
             name: {{ solution }}-operator-config
+{%- if webhook_enabled is defined and webhook_enabled %}
+        - name: cert
+          secret:
+            defaultMode: 420
+            secretName: {{ solution }}-cert
+{%- endif %}
```
TeddyAndrieux, line 66: nitpick the `is defined` is not needed here since it will always be defined but fine
```suggestion
{%- if webhook_enabled %}
            - name: cert
              mountPath: /cert
              readOnly: true
{%- endif %}
      volumes:
        - name: operator-config
          configMap:
            name: {{ solution }}-operator-config
{%- if webhook_enabled %}
        - name: cert
          secret:
            defaultMode: 420
            secretName: {{ solution }}-cert
{%- endif %}
```

salt/tests/unit/formulas/config.yaml
```diff
@@ -753,6 +753,15 @@ metalk8s:
...
                   image_tag: "1.2.3"
 
+              "Example Solution v1.2.3 with webhook":
+                extra_context:
+                  <<: *base_context_solution_operator_files
+                  repository: >-
+                    metalk8s-registry-from-config.invalid/example-solution-1.2.3
+                  image_name: example-operator-webhook
+                  image_tag: "1.2.3"
+                  webhook_enabled: 'true'
```
TeddyAndrieux: Here as well it shouldn't be quoted

```suggestion
                  webhook_enabled: true
```

TeddyAndrieux (commented):
LGTM except for the quote on boolean (and if you could just add a line about it in `CHANGELOG.md` in `Additions`)

scripts/solutions.sh
lamphamsy: indeed, the quote should be removed. However, it's weird that when I tested that the process works correctly even with the quote.

salt/metalk8s/orchestrate/solutions/files/operator/deployment.yaml.j2
lamphamsy: as in the unit tests `formulas`, this variable is not defined. So I add this check to cover the case.

salt/tests/unit/formulas/config.yaml
lamphamsy: it's updated. Thanks

TeddyAndrieux (approved):

PR #4038, opened 2023-04-06
Add behaviour to remove older backups

eg-ayoub:


salt/metalk8s/orchestrate/backup/files/job.yaml.j2
```diff
@@ -22,7 +22,10 @@ spec:
       containers:
       - name: backup-replication
         image: {{ image }}
-        command:
+        command: ["/bin/sh", "-c"]
+        args:
+          - ls -t -1 /var/lib/metalk8s/backups/*.tar.gz | tail -n+6 | xargs rm -f
```
eg-ayoub: could this be in the pillar ? I can't find it

TeddyAndrieux: You mean the `path` ? 
Here you are in a Pod and as you can see below
```yaml
        volumeMounts:
        - name: backups
          mountPath: /backups
```
The backups are mounted in `/backups` in the Pod

eg-ayoub: that makes it a lot easier, thanks !

salt/metalk8s/orchestrate/backup/files/job.yaml.j2
```diff
@@ -22,7 +22,10 @@ spec:
       containers:
       - name: backup-replication
         image: {{ image }}
-        command:
+        command: ["/bin/sh", "-c"]
+        args:
+          - ls -t -1 /backups/*.tar.gz | tail -n+6 | xargs rm -f
+          - ;
```
TeddyAndrieux: I would prefer to do this removal after the `wget`

minor: And I would add the `--verbose` to the `xargs` so that we can actually see from the pod logs what get removed

tests/post/features/backup.feature
```diff
@@ -0,0 +1,5 @@
+@post @ci @local @backup
+Feature Backup
```
TeddyAndrieux: Missing a `:` here

```suggestion
Feature: Backup
```

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,43 @@
...
+@then(parsers.parse("we have at most {most_unt} backups on each node"))
+def check_backup_archive_count(host, version, ssh_config, most_count):
+    command = [
+        "salt",
+        "--static",
+        "--out=json",
+        "'*'",
+        "cmd.run",
+        "ls -1 /var/lib/metalk8s/backups/*.tar.gz | wc -l",
+        f"saltenv=metalk8s-{version}",
```
TeddyAndrieux: You do not need `saltenv` here

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,43 @@
...
+    ]
+
+    backup_counts = json.loads(utils.run_salt_command(host, command, ssh_config))
+
+    for name, count in backup_counts.items():
+        assert int(count) <= int(most_count), (
```
TeddyAndrieux: check the retcode first please

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+# Then
+@then(parsers.parse("we have at most {most_unt} backups on each node"))
+def check_backup_archive_count(host, version, ssh_config, most_count):
+    command = [
+        "salt",
+        "--static",
+        "--out=json",
+        "'*'",
```
TeddyAndrieux: You should run it only on "master" nodes

salt/metalk8s/orchestrate/backup/files/job.yaml.j2
```diff
@@ -22,7 +22,10 @@ spec:
       containers:
       - name: backup-replication
         image: {{ image }}
-        command:
+        command: ["/bin/sh", "-c"]
+        args:
+          - ls -t -1 /backups/*.tar.gz | tail -n+6 | xargs rm -f
```
TeddyAndrieux: Since you use `sh -c` you should have the full command as a single argument

salt/metalk8s/orchestrate/backup/files/job.yaml.j2
```diff
@@ -22,15 +22,15 @@ spec:
...
-          - https://backup/
+        command: ["/bin/sh", "-c"]
+        args:
+          - >-
+            wget --accept="*.tar.gz"
+            --no-host-directories --mirror
+            --ca-certificate=/certificate/ca.crt
+            --user=$(BACKUP_USERNAME)
+            --password=$(BACKUP_PASSWORD)
+            https://backup/; ls -t -1 /backups/*.tar.gz | tail -n+6 | xargs -t rm -f
```
TeddyAndrieux: What happens if the wget fail ? It will silently continue and report success on the job no ?

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+    master_nodes = yaml.load(
+        utils.run_salt_command(host, get_master_nodes_command, ssh_config).stdout,
+        Loader=yaml.Loader,
+    )
+    assert len(master_nodes) > 0, "No master node found"
+
+    if len(master_nodes) == 1:
+        master_nodes = master_nodes[0]
+    else:
+        master_nodes = f"-L {','.join(master_nodes)}"
```
TeddyAndrieux: To me you do not need to make this distinction a `salt -L <x>` with only one minion works as well

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+        "salt-run",
+        "--out yaml",
+        "salt.cmd",
+        "metalk8s.minions_by_role",
+        "master",
+        "with_pillar=True",
+    ]
+    master_nodes = yaml.load(
+        utils.run_salt_command(host, get_master_nodes_command, ssh_config).stdout,
+        Loader=yaml.Loader,
```
TeddyAndrieux: This is needed ?

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+def check_backup_archive_count(host, version, ssh_config, count):
+    # using yaml becasue json output is not always correctly formatted
+    get_master_nodes_command = [
+        "salt-run",
+        "--out yaml",
+        "salt.cmd",
+        "metalk8s.minions_by_role",
+        "master",
+        "with_pillar=True",
+    ]
```
TeddyAndrieux: I would suggest retrieving the list of masters from kubernetes directly, it's a simple `get` with a selector

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+    assert len(master_nodes) > 0, "No master node found"
+
+    if len(master_nodes) == 1:
+        master_nodes = master_nodes[0]
+    else:
+        master_nodes = f"-L {','.join(master_nodes)}"
+
+    command = [
+        "salt",
+        "--out yaml",
```
TeddyAndrieux: You should use json instead to me

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+    iso_root = request.config.getoption("--iso-root")
+
+    for _ in range(int(times)):
+        with host.sudo():
+            res = host.run("%s/backup.sh", str(iso_root))
+            assert res.rc == 0, res.stdout
+
+
+# Then
+@then(parsers.parse("we have {count} backups on each node"))
```
TeddyAndrieux: nit: You can cast it to int directly using `converters` args

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+        master_nodes = master_nodes[0]
+    else:
+        master_nodes = f"-L {','.join(master_nodes)}"
+
+    command = [
+        "salt",
+        "--out yaml",
+        master_nodes,
+        "cmd.run_all",
+        '"find /var/lib/metalk8s/backups -name "*.tar.gz" | wc -l"',
```
TeddyAndrieux: I would suggest using `file.find` and counting the number of files in Python here instead

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,62 @@
...
+    for _ in range(int(times)):
+        with host.sudo():
+            res = host.run("%s/backup.sh", str(iso_root))
+            assert res.rc == 0, res.stdout
+
+
+# Then
+@then(parsers.parse("we have {count} backups on each node"))
+def check_backup_archive_count(host, version, ssh_config, count):
+    # using yaml becasue json output is not always correctly formatted
```
TeddyAndrieux: Really ? We use json in a lot of places in the tests already

tests/post/features/backup.feature
```diff
@@ -0,0 +1,5 @@
+@post @ci @local @backup
+Feature: Backup
+    Scenario: Backup multiple times
+        When we run the backup script 6 times
+        Then we have 5 backups on each node
```
TeddyAndrieux: I think there is an issue somewhere in the test (not sure where) because, as is, it cannot work in multi node to me, since the Job to replicate and clean are asynchronous and you do not wait for them to complete here

salt/metalk8s/orchestrate/backup/files/job.yaml.j2
eg-ayoub: both commands can silently fail here,
that is probably why we didn't catch it before
can I add set -e and pipefail ?

tests/post/steps/test_backup.py
eg-ayoub: I put a comment about the yaml output in the beginning, 
salt json outputs a single json object for each targetted nod. so output looks like:
`
{
  "attr1": 0 
}
{
  "attr1" 0
}
`

which was too much trouble to parse manually,
since yaml is 'stackable' I preferred to use it

tests/post/steps/test_backup.py
eg-ayoub: ditto here

salt/metalk8s/orchestrate/backup/files/job.yaml.j2
TeddyAndrieux: Hmm yes we could likely, but otherwise just a `&&` instead of `;`

tests/post/features/backup.feature
eg-ayoub: I think it may be enough to add a simple wait here maybe 1 minute or two

tests/post/steps/test_backup.py
eg-ayoub: for some reason it just would not work when trying live with json, 
I wrote another comment with what the output was like

tests/post/features/backup.feature
TeddyAndrieux: Why ? It's pretty simple to wait for Job to succeed (and it's a better approach that arbitrary waiting time)

tests/post/steps/test_backup.py
TeddyAndrieux: You just have to use `--static` to avoid this

> -s, --static
By default as of version 0.9.8 the salt command returns data to the console as it is received from minions, but previous releases would return data only after all data was received. Use the static option to only return the data with a hard timeout and after all minions have returned. Without the static option, you will get a separate JSON string per minion which makes JSON output invalid as a whole.

tests/post/steps/test_backup.py
eg-ayoub: but this needs to run on all nodes

TeddyAndrieux: Yes, but I don't see the issue, you can run a `file.find` on all nodes and then you iter on the dict like you do bellow

tests/post/steps/test_backup.py
```diff
@@ -0,0 +1,61 @@
...
+            res = host.run("%s/backup.sh", str(iso_root))
+            assert res.rc == 0, res.stdout
+
+    with host.sudo():
+        res = host.run(
+            "kubectl --kubeconfig=/etc/kubernetes/admin.conf "
+            "wait --for=condition=complete --timeout=60s "
+            "job -l app.kubernetes.io/name=backup-replication --namespace=kube-system"
+        )
+        assert res.rc == 0, res.stdout
```
TeddyAndrieux: I would prefer to move this to another "when" so that the process described in the feature is closer to what a user will do "when I run the backup 6 times and wait for the replication job to complete then I only have 5 backups"

TeddyAndrieux (commented):
Just a last comment otherwise LGTM

tests/conftest.py
```diff
@@ -361,6 +361,17 @@ def wait_rollout_status(host, resource, namespace):
...
 
+@when(parsers.parse("we wait {seconds} seconds for the {job} job to complete"))
+def wait_for_job(host, seconds, job):
+    with host.sudo():
+        res = host.run(
+            "kubectl --kubeconfig=/etc/kubernetes/admin.conf "
+            f"wait --for=condition=complete --timeout={seconds}s "
+            f"job -l app.kubernetes.io/name={job} -A"
+        )
+        assert res.rc == 0, res.stdout
```
TeddyAndrieux, line 372: Cool a generic one :+1:

TeddyAndrieux (approved):

PR #4053, opened 2023-05-15
ARTESCA-4416 put OOMKill alert in alert tree

aprucolimartins:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/tools/lib-alert-tree/lib_alert_tree/kubernetes.py
+++ b/tools/lib-alert-tree/lib_alert_tree/kubernetes.py
@@ -3,6 +3,26 @@
 from . import models
 
 
+def container_alerts(name=".*", pod=".*", namespace="default"):
+    """Common alerts for Containers."""
+    return [
+        models.ExistingAlert(
+            "KubeContainerOOMKilled",
+            severity="warning",
+            namespace=namespace,
+            pod=pod,
+            container=name,
+        ),
+        models.ExistingAlert(
+            "KubeContainerOOMKillSurge",
+            severity="critical",
+            namespace=namespace,
+            pod=pod,
+            container=name,
+        ),
+    ]
+
+
 def pod_alerts(name, severity="warning", namespace="default"):
     """Common alerts for Pods."""
     return [
@@ -10,7 +30,7 @@ def pod_alerts(name, severity="warning", namespace="default"):
             alertname, severity=severity, namespace=namespace, pod=name
         )
         for alertname in ["KubePodNotReady", "KubePodCrashLooping"]
-    ]
+    ] + container_alerts(".*", pod=name, namespace=namespace)
 
 
 def deployment_alerts(name, severity="warning", namespace="default"):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
   Changes metalk8s-sosreport package version to add the `patch` digit
   (PR[#4056](https://github.com/scality/metalk8s/pull/4056))
 
+### Enhancements
+
+- Add Out Of Memory alerts for containers
+  (PR[#4042](https://github.com/scality/metalk8s/pull/4042))
+
 ## Release 124.1.5
 
 ### Enhancements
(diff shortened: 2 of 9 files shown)


ezekiel-alexrod (dismissed):

gdemonet (commented):
When modifying these "composite alerts", a few things are automatically generated (for instance, `salt/metalk8s/addons/alert-tree/files/prometheus_rule.yaml` in the ISO). However, there is an end-to-end test (scenario here: https://github.com/scality/metalk8s/blob/development/125.0/tests/post/features/monitoring.feature#L45-L48 and code here: https://github.com/scality/metalk8s/blob/development/125.0/tests/post/steps/test_monitoring.py#L497), which requires a manual update for the list of deployed rules (since these are used for generating documentation).

To do this, you'll need a running platform with your custom alerts deployed (you can use an existing one and just apply the generated rules by hand), then run the script in `tools/rule_extractor/rule_extractor.py` (it needs to reach Prom API), and commit the output files `alerting_rules.json` and `rules.json` with your PR.

It's cumbersome, but today that's how it works :/ (maybe we should add some dev docs...)

---

Edit: that's why the docs build fails at the moment with the cryptic error

```
Warning, treated as error:
<unknown>:47:undefined label: simple-alert-kubecontaineroomkilled-warning
```

tools/lib-alert-tree/lib_alert_tree/kubernetes.py
```diff
@@ -3,14 +3,24 @@
 from . import models
 
 
+def container_alerts(name, severity="warning", namespace="default"):
+    """Common alerts for Containers."""
+    return [
+        models.ExistingAlert(
+            alertname, severity=severity, namespace=namespace, statefulset=name
```
TeddyAndrieux: It should be container here
```suggestion
            alertname, severity=severity, namespace=namespace, container=name
```

And in order to be able to use it in `pod_alerts` you will need to support both `container` and `pod`

tools/lib-alert-tree/lib_alert_tree/kubernetes.py
```diff
@@ -3,14 +3,24 @@
 from . import models
 
 
+def container_alerts(name, severity="warning", namespace="default"):
+    """Common alerts for Containers."""
+    return [
+        models.ExistingAlert(
+            alertname, severity=severity, namespace=namespace, container=name
+        )
```
TeddyAndrieux: You need to support pod name here as well as container name otherwise the call from `pod_alerts` will not give what you expected, (container name is not necessarily equal to the pod name)

aprucolimartins: Done !!!

tools/lib-alert-tree/lib_alert_tree/kubernetes.py
```diff
@@ -3,14 +3,24 @@
 from . import models
 
 
+def container_alerts(name, container=".*", severity="warning", namespace="default"):
```
TeddyAndrieux: I would set this the otherway around :) 
name for container and another arg for the pod name

aprucolimartins: I'll do that, but I prefer that way because I think more readable.

TeddyAndrieux: Why do you think it's more readable ?
To me it's less readable, here you have a function `container_alerts` with a `name` argument that is not the container name but a pod name, so to me, it's not super clear

tools/lib-alert-tree/lib_alert_tree/kubernetes.py
```diff
@@ -3,14 +3,24 @@
 from . import models
 
 
+def container_alerts(name, pod=".*", severity="warning", namespace="default"):
```
TeddyAndrieux: I suggest you put a default on `name` as well so that (I agree) the call in `pod_alerts` will be more readable

```suggestion
def container_alerts(name=".*", pod=".*", severity="warning", namespace="default"):
```

ezekiel-alexrod (approved):

PR #4118, opened 2023-08-03
operator: Add liveness probe to keepalived

ezekiel-alexrod:
**Component**: operator

**Context**: Keepalived maps its configuration upon interface names. When VIPs are managed by keepalived and an interface which has VIP changes name or goes down, keepalived does not sync.  Therefore, the configuration is not correct and must be regenerated. The config is set up at pod startup. 

**Summary**: Adding a liveness probe in Kubernetes to restart the keepalived pod

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,9 @@
   [1.19](https://golang.org/doc/devel/release.html#go1.19.minor)
   (PR[#4110](https://github.com/scality/metalk8s/pull/4110))
 
+- Add liveness probe to `keepalived` pod 
+  (PR[#4118](https://github.com/scality/metalk8s/pull/4118))
+
 ## Release 125.0.6 (In development)
 
 
--- a/images/metalk8s-keepalived/liveness-probe.sh
+++ b/images/metalk8s-keepalived/liveness-probe.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+set -xue -o pipefail
+
+# Simple check to ensure that the config does not need to be updated
+/generate-config.py --input /etc/keepalived/keepalived-input.yaml --template /etc/keepalived/keepalived.conf.j2 | diff -qw /etc/keepalived/keepalived.conf -
--- a/operator/pkg/controller/virtualippool/controller.go
+++ b/operator/pkg/controller/virtualippool/controller.go
@@ -468,6 +468,20 @@ func (r *VirtualIPPoolReconciler) mutateDaemonSet(obj *appsv1.DaemonSet) error {
 			},
 		},
 	}
+	if container.LivenessProbe == nil {
+		container.LivenessProbe = &corev1.Probe{}
+	}
+	container.LivenessProbe.ProbeHandler = corev1.ProbeHandler{
+		Exec: &corev1.ExecAction{
+			Command: []string{
+				"/liveness-probe.sh",
+			},
+		},
+	}
+	container.LivenessProbe.InitialDelaySeconds = 10
+	container.LivenessProbe.PeriodSeconds = 30
+	container.LivenessProbe.TimeoutSeconds = 10
+	container.LivenessProbe.FailureThreshold = 1
 
 	return nil
 }
--- a/images/metalk8s-keepalived/Dockerfile
+++ b/images/metalk8s-keepalived/Dockerfile
@@ -39,6 +39,7 @@ COPY --chown=keepalived:keepalived keepalived.conf.j2 /etc/keepalived/
 COPY --chown=keepalived:keepalived check-get.sh /etc/keepalived/
 
 COPY --chown=keepalived:keepalived generate-config.py /
+COPY --chown=keepalived:keepalived liveness-probe.sh /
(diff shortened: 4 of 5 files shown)


operator/pkg/controller/virtualippool/controller.go
```diff
@@ -468,6 +468,19 @@ func (r *VirtualIPPoolReconciler) mutateDaemonSet(obj *appsv1.DaemonSet) error {
 			},
 		},
 	}
+	container.LivenessProbe = &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			Exec: &corev1.ExecAction{
+				Command: []string{
+					"/bin/bash",
+					"-c",
+					"/generate-config.py --input /etc/keepalived/keepalived-input.yaml --template /etc/keepalived/keepalived.conf.j2 --output /etc/keepalived/keepalived.conf.liveness && diff -w /etc/keepalived/keepalived.conf /etc/keepalived/keepalived.conf.liveness",
```
TeddyAndrieux: Can we add a script `liveness.sh` (or something like that) directly in the image since the logic to create this config file is there
(So that we could also easily enrich it in the future)

images/metalk8s-keepalived/liveness-probe.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+/generate-config.py --input /etc/keepalived/keepalived-input.yaml --template /etc/keepalived/keepalived.conf.j2 --output /etc/keepalived/keepalived.conf.liveness && diff -w /etc/keepalived/keepalived.conf /etc/keepalived/keepalived.conf.liveness
```
TeddyAndrieux: Why `diff -w` ? I would go with a `-q` instead no ? And to me there is no need to actually create the configuration file just give it as input to diff

If we could also have a simple comment about this check

I would change it like that 

```suggestion
#!/bin/bash

set -xue -o pipefail

# Simple check to ensure that the config does not need to be updated
/generate-config.py --input /etc/keepalived/keepalived-input.yaml --template /etc/keepalived/keepalived.conf.j2 | diff -q /etc/keepalived/keepalived.conf -
```

operator/pkg/controller/virtualippool/controller.go
```diff
@@ -468,6 +468,17 @@ func (r *VirtualIPPoolReconciler) mutateDaemonSet(obj *appsv1.DaemonSet) error {
...
 	}
+	container.LivenessProbe = &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			Exec: &corev1.ExecAction{
+				Command: []string{
+					"/liveness-probe.sh",
+				},
+			},
+		},
+		TimeoutSeconds:   5,
```
TeddyAndrieux: I would put a `periodSeconds` to `30` and maybe `10` as timeout instead

TeddyAndrieux (changes requested):
Just two last comments + could you add a simple changelog entry about this :)

images/metalk8s-keepalived/liveness-probe.sh
ezekiel-alexrod: `diff -w` is for not considering the changes about whitespaces which I had when regenerating the config file
I can put both : `diff -qw`

TeddyAndrieux: Oh indeed, I read the beginning of diff helper and ....

```console
$ LANG= diff --help | grep "\-w" 
  -W, --width=NUM               output at most NUM (default 130) print columns
  -w, --ignore-all-space          ignore all white space
$ LANG=fr_FR.utf8 diff --help | grep "\-w"
  -w, --width=N                Limiter la sortie à au plus N colonnes imprimées (130 par défaut)
  -w, --ignore-all-space        Ignorer tout blanc d'espacement
```
NOTE: The double `-w` lowercase in FR ... 

Ok let's put both

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 
 ### Enhancements
 
+- Add liveness probe to `keepalived` pod 
+  (PR[#4118](https://github.com/scality/metalk8s/pull/4118))
```
TeddyAndrieux: Nit: Can you just move this at the end of `Enhancements` list :)

ezekiel-alexrod: Oopsie

CHANGELOG.md
```diff
@@ -90,6 +90,9 @@
   [1.19](https://golang.org/doc/devel/release.html#go1.19.minor)
   (PR[#4110](https://github.com/scality/metalk8s/pull/4110))
 
+  - Add liveness probe to `keepalived` pod
```
TeddyAndrieux: Indentation does not looks good :)

ezekiel-alexrod: omg, I'm so off today :(
My bad and thanks

TeddyAndrieux (commented):
Good to go for squash :)

TeddyAndrieux (approved):
Good job !

PR #4127, opened 2023-08-18
shell-ui: Filter out the non active alerts

ChengYanJin:
**Component**: Shell UI

**Context**: 
Following the act of silencing an alert, the alert's status would transition to suppressed. Nonetheless, our aim is to fetch solely the active alerts. Therefore, we must sort out and exclude the alerts that are not currently active.

**Summary**:

**Acceptance criteria**: 
On the Alert Page, only the active alerts should be visible.


JBWatenbergScality (dismissed):
LGTM !!!

JBWatenbergScality (approved):
notifications-gif

PR #4130, opened 2023-08-25
Ui: Display the infra role for node creation only in pure MetalK8s scenario

ChengYanJin:
**Component**: UI

**Context**: 
Notification Center: 
We want to have the same height for each of item in Notification Center.

Create Node Form: 
Infra node only make sense in pure Metalk8s scenario

**Summary**:

**Acceptance criteria**:


MonPote (approved):

MonPote:
We may want to find a better way to do this.

PR #4133, opened 2023-08-30
shell-ui: Show the Notification Center only for platform admin

ChengYanJin:
**Component**: Shell UI

**Context**:  Show the Notification Center only for platform admin

**Summary**: We just need to hide the Notification Center Component from Navbar. But the logic has been implemented in Version, Alert and License. there is no need to change.

**Acceptance criteria**:

### DIFF
--- a/shell-ui/src/navbar/NavBar.tsx
+++ b/shell-ui/src/navbar/NavBar.tsx
@@ -12,7 +12,7 @@ import { useTheme } from 'styled-components';
 import { useLanguage } from './lang';
 import { useThemeName } from './theme';
 import { useIntl } from 'react-intl';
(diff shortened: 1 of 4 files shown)


shell-ui/src/navbar/NavBar.tsx
```diff
@@ -347,6 +356,9 @@ export const Navbar = ({
     },
   ];
 
+  if (isPlatformAdmin(userData)) {
+    rightTabs.splice(secondaryTabs.length - 1, 0, notificationTab);
```
hervedombya, line 360: why not doing something like `...(isPlatformAdmin(userData) ? [notificationTab] : [])` rather than mutating the array ?

hervedombya (approved):

PR #4134, opened 2023-08-30
salt: add allocated resources for OS and kubelet

ezekiel-alexrod:
**Component**:

salt

**Context**: 

To avoid exhausting memory and CPU resources on the server, we want to set a minimum amount of memory and CPU for the operating system and kubelet.

We will follow this document [Allocatable resources](https://learnk8s.io/allocatable-resources) and especially Google recommandation as they are the state of art of kubernetes.

We will add some “memory and CPU” because we are using salt and mdadm and they are vital for the cluster.

**Summary**:

I allow myself to add this link as it is the 
[Official documentation of Google](https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes#memory_and_cpu_reservations)

Added `kubeReserved` and `systemReserved` to `KubeletConfiguration` following the Google's recommandations.

### DIFF
--- a/salt/_modules/metalk8s_os.py
+++ b/salt/_modules/metalk8s_os.py
@@ -1,0 +1,59 @@
+"""
+MetalK8s OS module
+"""
+
+__virtualname__ = "metalk8s_os"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def get_kubereserved():
+    """Get kubeReserved memory and cpu allocations following Google's GKE recommandations.
+
+    For CPU resources, GKE reserves the following:
+    - 6% of the first core
+    - 1% of the next core (up to 2 cores)
+    - 0.5% of the next 2 cores (up to 4 cores)
+    - 0.25% of any cores above 4 cores
+
+    For memory resources, GKE reserves the following:
+    - 255 MiB of memory for machines with less than 1 GB of memory
+    - 25% of the first 4GB of memory
+    - 20% of the next 4GB of memory (up to 8GB)
+    - 10% of the next 8GB of memory (up to 16GB)
+    - 6% of the next 112GB of memory (up to 128GB)
+    - 2% of any memory above 128GB
+
+
+    https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes#memory_and_cpu_reservations
+    https://learnk8s.io/allocatable-resources
+    """
+
+    os_cpu = __grains__["num_cpus"]
+
+    core_1 = 0.06
+    cores_2 = 0.01 * max([(min([os_cpu, 2]) - 1), 0])
+    cores_4 = 0.005 * max([(min([os_cpu, 4]) - 2), 0])
+    cores_above_4 = 0.0025 * max([(os_cpu - 4), 0])
+
+    kube_cpu = round((core_1 + cores_2 + cores_4 + cores_above_4) * 1000)
+
+    os_memory = __grains__["mem_total"]
+
+    gb_4 = 4 * 1024
+    gb_8 = 8 * 1024
+    gb_16 = 16 * 1024
+    gb_128 = 128 * 1024
+
+    memory_4 = 0.25 * min([os_memory, gb_4])
+    memory_8 = 0.2 * max([(min([os_memory, gb_8]) - gb_4), 0])
+    memory_16 = 0.1 * max([(min([os_memory, gb_16]) - gb_8), 0])
+    memory_128 = 0.06 * max([(min([os_memory, gb_128]) - gb_16), 0])
+    memory_above_128 = 0.02 * max([(os_memory - gb_128), 0])
+
+    kube_memory = round(memory_4 + memory_8 + memory_16 + memory_128 + memory_above_128)
+
+    # CPU is in millicores, memory is in MiB
+    return {"cpu": f"{kube_cpu}m", "memory": f"{kube_memory}Mi"}
--- a/salt/metalk8s/kubernetes/kubelet/standalone.sls
+++ b/salt/metalk8s/kubernetes/kubelet/standalone.sls
@@ -103,6 +103,10 @@ Create kubelet config file:
         {%- if pillar.get("kubernetes:kubelet:config:maxPods") %}
         maxPods: {{ pillar.kubernetes.kubelet.config.maxPods }}
         {%- endif %}
+        systemReserved:
+          cpu: 200m
+          memory: 200Mi
+        kubeReserved: {{ salt.metalk8s_os.get_kubereserved() | tojson }}
 {%- for key, value in kubelet.config.items() %}
         {{ key }}: {{ value }}
 {%- endfor %}
--- a/salt/tests/unit/formulas/fixtures/salt.py
+++ b/salt/tests/unit/formulas/fixtures/salt.py
@@ -26,6 +26,7 @@ DEFAULT_CONFIG = {
     "file_client": "remote",
 }
 
+
 # The "public methods" are dynamically added by the `register` decorator
 # pylint: disable=too-few-public-methods
 class SaltMock:
@@ -186,6 +187,7 @@ def register_basic(func_name: str) -> Callable[[MockFunc], MockFunc]:
 
 # }}}
 # Mock definitions {{{
+
 
 # Data-driven mocks {{{
 @register("config.get")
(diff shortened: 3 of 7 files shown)


salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -5,6 +5,36 @@
 
 {%- set cluster_dns_ip = salt.metalk8s_network.get_cluster_dns_ip() %}
 
+{%- set total_memory = (salt['status.meminfo']()['MemTotal']['value'] | int) / 1024 %}
+{%- set total_cpu_cores = salt['status.cpuinfo']()['cpu cores'] | int %}
+
+# Google recommandations about OS and kubelet allocated resources
+# https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes#memory_and_cpu_reservations 
+# https://learnk8s.io/allocatable-resources 
+# {
```
TeddyAndrieux: Nit let's move it as jinja comments 

```suggestion
{#- Google recommandations about OS and kubelet allocated resources
  https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes#memory_and_cpu_reservations 
  https://learnk8s.io/allocatable-resources 
#}
```

TeddyAndrieux: I would also add in the comment the "calculation done" (just in case the links are no longer available and to ease the understanding of the below calculs

salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -5,6 +5,36 @@
...
+{%- set gb_16 = 16384 %}
+{%- set gb_128 = 131072 %}
+
+{%- set memory_4 = 0.25 * ([total_memory, gb_4] | min) %}
+{%- set memory_8 = 0.2 * [([total_memory, gb_8] | min) - gb_4, 0] | max %}
+{%- set memory_16 = 0.1 * [([total_memory, gb_16] | min) - gb_8, 0] | max %}
+{%- set memory_128 = 0.06 * [([total_memory, gb_128] | min) - gb_16, 0] | max %}
+{%- set memory_above_128 = 0.02 * [total_memory - gb_128, 0] | max %}
+
+{%- set reserved_kube_memory = (memory_4 + memory_8 + memory_16 + memory_128 + memory_above_128) | round | int %}
```
TeddyAndrieux: After reflection, I would move this calculation to a dedicated python module so that we could also add some basic unit tests to be sure we do not put really wrong values, something like `metalk8s_os.py` or something like that

salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -103,6 +133,12 @@ Create kubelet config file:
         {%- if pillar.get("kubernetes:kubelet:config:maxPods") %}
         maxPods: {{ pillar.kubernetes.kubelet.config.maxPods }}
         {%- endif %}
+        systemReserved:
+          cpu: 100m
+          memory: 100Mi
```
TeddyAndrieux: Since we also have salt running on the system, I suggest to double it

TeddyAndrieux (changes requested):
If you may also add a changelog entry about it

salt/metalk8s/kubernetes/kubelet/standalone.sls
ezekiel-alexrod: Ok, fine. Wouldn't it make sense to put it in the already existent `metalk8s_kubeconfig` since it's some values for this file ?

TeddyAndrieux: Hmm no it's not, `metalk8s_kubeconfig` is for kubeconfig files here you are updating the kubelet config, it's not the same thing

kubeconfig is a config file to talk to kube APIServer

salt/_modules/metalk8s_os.py
```diff
@@ -0,0 +1,59 @@
...
+    - 10% of the next 8GB of memory (up to 16GB)
+    - 6% of the next 112GB of memory (up to 128GB)
+    - 2% of any memory above 128GB
+
+
+    https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes#memory_and_cpu_reservations
+    https://learnk8s.io/allocatable-resources
+    """
+
+    os_cpu = __salt__["status.nproc"]()
```
TeddyAndrieux: I would use the grain instead so 

```suggestion
    os_cpu = __grains__["num_cpus"]
```

salt/_modules/metalk8s_os.py
```diff
@@ -0,0 +1,59 @@
...
+
+    core_1 = 0.06
+    cores_2 = 0.01 * max([(min([os_cpu, 2]) - 1), 0])
+    cores_4 = 0.005 * max([(min([os_cpu, 4]) - 2), 0])
+    cores_above_4 = 0.0025 * max([(os_cpu - 4), 0])
+
+    kube_cpu = round((core_1 + cores_2 + cores_4 + cores_above_4) * 1000)
+
+    os_memory = int(__salt__["status.meminfo"]()["MemTotal"]["value"]) / 1024
```
TeddyAndrieux: Ditto let's use grains as well here (note grains is already in MB)

```suggestion
    os_memory = __grains__["mem_total"]
```

salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -103,6 +105,12 @@ Create kubelet config file:
         {%- if pillar.get("kubernetes:kubelet:config:maxPods") %}
         maxPods: {{ pillar.kubernetes.kubelet.config.maxPods }}
         {%- endif %}
+        systemReserved:
+          cpu: 200m
+          memory: 200Mi
+        kubeReserved:
+          cpu: {{ kubereserved.cpu }}m
+          memory: {{ kubereserved.memory }}Mi
```
TeddyAndrieux: To me the, `m` and `Mi` should be part of the return of the module function

You can even rewrite it to 

```yaml
kubeReserved: {{ salt.metalk8s_os.get_kubereserved() | tojson }}
```

salt/_modules/metalk8s_os.py
```diff
@@ -0,0 +1,59 @@
...
+
+    gb_4 = 4096
+    gb_8 = 8192
+    gb_16 = 16384
+    gb_128 = 131072
```
TeddyAndrieux: nitpick: Just for readability

```suggestion
    gb_4 = 4 * 1024
    gb_8 = 8 * 1024
    gb_16 = 16 * 1024
    gb_128 = 128 * 1024
```

salt/_modules/metalk8s_os.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+__virtualname__ = "metalk8s_os"
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def get_kubereserved() -> dict:
```
TeddyAndrieux: nit: Let's not type here we do not use type anywhere else in salt modules

```suggestion
def get_kubereserved():
```

salt/tests/unit/formulas/fixtures/salt.py
```diff
@@ -26,6 +26,7 @@
     "file_client": "remote",
 }
 
+
```
TeddyAndrieux, line 29: I think those extra lines can be removed :)

salt/tests/unit/formulas/fixtures/salt.py
```diff
@@ -187,6 +188,7 @@ def register_basic(func_name: str) -> Callable[[MockFunc], MockFunc]:
 # }}}
 # Mock definitions {{{
 
+
```
TeddyAndrieux, line 191: ditto

TeddyAndrieux (changes requested):

salt/tests/unit/formulas/fixtures/salt.py
ezekiel-alexrod: They were introduced when running `black-formatter` 
Do I remove those then ?

TeddyAndrieux: Yes, I think they were added just because you do not use the exact same black version that the one we have in the CI, so let's just remove them since you do not touch the file in your PR

salt/tests/unit/modules/files/test_metalk8s_os.yaml
```diff
@@ -0,0 +1,67 @@
...
+    result: 
+      cpu: 710m
+      memory: 17408Mi
+  # 09 - 512 cores and 1024 GB of RAM
+  - num_cpus: 512
+    mem_total: 1048576
+    result: 
+      cpu: 1350m
+      memory: 27894Mi
+  # 10 - 3 cores and 14.2 GB of RAM
```
TeddyAndrieux, line 62: Nice :smile:

TeddyAndrieux (approved):
LGTM good job :+1:

PR #4135, opened 2023-08-31
ARTESCA-6324 // pod status not well reflected

hervedombya:
NodePagePodsTab test

**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER

### DIFF
--- a/ui/src/components/NodePagePodsTab.spec.tsx
+++ b/ui/src/components/NodePagePodsTab.spec.tsx
@@ -1,0 +1,124 @@
+import { brand } from '@scality/core-ui/dist/style/theme';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { QueryClient, QueryClientProvider } from 'react-query';
+import { useSelector } from 'react-redux';
+import { ThemeProvider } from 'styled-components';
+import { mockOffsetSize } from '../tests/mocks/util';
+import translations_en from '../translations/en.json';
+import NodePagePodsTab from './NodePagePodsTab';
+
+const mockUseSelector = useSelector as jest.Mock;
+
+const mockResult = (numContainerRunning: number, status = 'Running') => ({
+  name: 'test-pod',
+  age: '4d2h',
+  namespace: 'xcore',
+  status: {
+    status,
+    numContainer: 3,
+    numContainerRunning,
+  },
+  log: 'test',
+});
+
+const statusHealthyRGB = 'rgb(10, 173, 166)';
+const statusWarningRGB = 'rgb(248, 243, 43)';
+const statusCriticalRGB = 'rgb(232, 72, 85)';
+
+const selectors = {
+  podName: () => screen.getByText('test-pod'),
+  podAge: () => screen.getByText('4d2h'),
+  podNamespace: () => screen.getByText('xcore'),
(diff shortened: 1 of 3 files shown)


ui/src/components/NodePagePodsTab.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  name: 'test-pod',
+  age: '4d2h',
+  namespace: 'xcore',
+  status: {
+    status,
+    numContainer: 3,
+    numContainerRunning,
+  },
+  log: 'test',
+});
```
ChengYanJin, line 24: I would recommend utilizing msw for mocking purposes if it's possible.

I do realise that we still reply a lot on redux. but when we implement the test, we may forgot about the technical, we just want to mock the api and check how the UI is display.

You will need to mock the response of /api/kubernetes/api/v1/pods/
Example of Pod running but one container not ready.

```json
{
  "apiVersion": "v1",
  "kind": "Pod",
  "metadata": {
    "name": "example-pod"
  },
  "status": {
    "phase": "Running",
    "containerStatuses": [
      {
        "name": "container-1",
        "state": {
          "running": {
            "startedAt": "2023-09-01T10:00:00Z"
          }
        },
        "ready": true
      },
      {
        "name": "container-2",
        "state": {
          "waiting": {
            "reason": "ImagePullBackOff",
            "message": "Back-off pulling image \"nginx\""
          }
        },
        "ready": false
      }
    ]
  }
}
```
example of pod is not running:
```json
{
  "apiVersion": "v1",
  "kind": "Pod",
  "metadata": {
    "name": "example-pod"
  },
  "status": {
    "phase": "Failed",
    "reason": "CrashLoopBackOff",
    "message": "Container 'my-container' exited with error code 1",
    "containerStatuses": [
      {
        "name": "my-container",
        "state": {
          "terminated": {
            "exitCode": 1,
            "reason": "Error",
            "startedAt": "2023-09-01T10:00:00Z",
            "finishedAt": "2023-09-01T10:05:00Z"
          }
        },
        "ready": false
      }
    ]
  }
}
```

ui/src/components/NodePagePodsTab.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  beforeEach(() => {
+    mockUseSelector.mockImplementation((selector) =>
+      selector({
+        config: {
+          api: {
+            url_grafana: 'http://mock.grafana.url',
+          },
+        },
+      }),
+    );
```
ChengYanJin, line 73: I think that we can mocking through msw. This way, if we ever transition from Redux to React-Query, we won't have to rewrite this test.

ChengYanJin (commented):
The fix LGTM. but the test I would suggest to implement using msw.

ui/src/components/NodePagePodsTab.spec.tsx
ChengYanJin: Due to the current implementation of the component not being conducive to unit testing, it's necessary to render the entire Node page for using msw to mock API responses. 😢

In light of this, we should consider refactoring the component to make testing more straightforward.

ChengYanJin (approved):
It's sad that we'll leave some technical debt within the test.

but I think without a refactoring of the component is not easy to test. We will do the refactoring while we migrating redux to react-query.

PR #4153, opened 2023-09-18
shell-ui: ARTESCA-9128 // ToastProvider

hervedombya:
/**
 * Hook that handles the completion of multiple mutations and displays a toast message.
 * @param mainMutation The main mutation to execute.
 * @param dependantMutations Optional array of dependant mutations to execute.
 * @param messageDescriptionBuilder Function that builds the message to display in the toast.
 * @param toastStyles Optional styles for the toast.
 * @param onPrimarySuccess Optional callback to execute when the primary mutation succeeds.
 */

### DIFF
--- a/shell-ui/src/toast/useMutationsHandler.ts
+++ b/shell-ui/src/toast/useMutationsHandler.ts
@@ -1,0 +1,96 @@
+import { ReactNode, useCallback, useEffect } from 'react';
+import { UseMutationResult } from 'react-query';
+import { useToast } from './useToast';
+
+export type MutationConfig<Data, Variables> = {
+  mutation: UseMutationResult<Data, unknown, Variables, unknown>;
+  name: string;
+};
+
+type DescriptionBuilder<Data> = {
+  data?: Data;
+  error?: unknown;
+  name: string;
+};
+
+type MutationsHandlerProps<Data, Variables> = {
+  mainMutation: MutationConfig<Data, Variables>;
+  dependantMutations?: MutationConfig<Data, Variables>[];
+  messageDescriptionBuilder: (
+    successMutations: DescriptionBuilder<Data>[],
+    errorMutations: DescriptionBuilder<Data>[],
+  ) => ReactNode;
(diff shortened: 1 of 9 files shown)


shell-ui/src/toast/useMutationsHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+
+type MutationsHandlerProps<Data, Variables> = {
+  mainMutation: MainMutationConfig<Data, Variables>;
+  dependantMutations?: DependantMutationConfig<Data, Variables>[];
+  messageDescriptionBuilder: (
+    successMutations: DescriptionBuilder<Data>[],
+    errorMutations: DescriptionBuilder<Data>[],
+  ) => ReactNode;
+  toastStyles?: React.CSSProperties;
+  onPrimarySuccess?: () => void;
```
ChengYanJin: maybe we should rename this to mainMutation too ;) 
```suggestion
  onMainMutationSuccess?: () => void;
```

shell-ui/src/toast/useMutationsHandler.ts
```diff
@@ -0,0 +1,108 @@
+import { ReactNode, useCallback, useEffect } from 'react';
+import { UseMutationResult } from 'react-query';
+import { useToast } from './useToast';
+
+export type MainMutationConfig<Data, Variables> = {
+  mutation: UseMutationResult<Data, unknown, Variables, unknown>;
+  name: string;
+  isPrimary: boolean;
```
ChengYanJin: I believe we could remove this `isPrimary` prop.

shell-ui/src/toast/useMutationsHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+type DependantMutationConfig<Data, Variables> = Omit<
+  MainMutationConfig<Data, Variables>,
+  'isPrimary'
+>;
```
ChengYanJin: So basically the type is the same is MainMutation, we could extract it .

PR #4200, opened 2023-11-27
salt: Fix the `get` method for `maxPods` value

ezekiel-alexrod:


salt/metalk8s/kubernetes/kubelet/standalone.sls
```diff
@@ -100,8 +100,8 @@ Create kubelet config file:
         address: {{ grains['metalk8s']['control_plane_ip'] }}
         rotateCertificates: false
         port: 10250
-        {%- if pillar.get("kubernetes:kubelet:config:maxPods") %}
-        maxPods: {{ pillar.kubernetes.kubelet.config.maxPods }}
+        {%- if salt.pillar.get("kubernetes:kubelet:config:maxPods") %}
+        maxPods: {{ salt.pillar.kubernetes.kubelet.config.maxPods }}
```
TeddyAndrieux: No here you need to keep just `pillar` otherwise here it means it will try to call the salt module named `pillar.kubernetes....` which will not work)

TeddyAndrieux (approved):

PR #4217, opened 2023-12-28
Check if pod/svc and node CIDRs overlap

eg-ayoub:


salt/_modules/metalk8s_checks.py
```diff
@@ -16,6 +16,47 @@ def __virtual__():
     return __virtualname__
 
 
+def bootstrap_file():
```
TeddyAndrieux, line 19: I would prefer to check using the value from the pillar directly so that you will have the defaults & co and do not have to manage the potential various version of bootstrap file

salt/_modules/metalk8s_checks.py
```diff
@@ -16,6 +16,47 @@ def __virtual__():
...
+    if pods_cidr is not None:
+        object_networks.append(("pods", ipaddress.ip_network(pods_cidr)))
+    svc_cidr = config.get("networks", {}).get("services")
+    if svc_cidr is not None:
+        object_networks.append(("services", ipaddress.ip_network(svc_cidr)))
+
+    for o in object_networks:
+        for n in node_networks:
+            if o[1].overlaps(n[1]):
+                errors.append(f" - {o[0]} and {n[0]}")
```
TeddyAndrieux: We should also check that SVC and POD CIDRs do not conflict between them

salt/_modules/metalk8s_checks.py
eg-ayoub: I can't access it in the pillar at this early step of execution,
I can maybe move this forward right before the bootstrap states ?

TeddyAndrieux: Hmm are you sure ?
This step runs after the "configure local salt minion" so the pillar should work here already to me

eg-ayoub: you were right about this, I made the move to using pillar

salt/_modules/metalk8s_checks.py
```diff
@@ -16,6 +16,60 @@ def __virtual__():
...
+    ]
+
+    # These should always be defined in the pillar
+    # They are each a CIDR address
+    object_networks = [
+        ("pods", ipaddress.ip_network(networks.get("pod"))),
+        ("services", ipaddress.ip_network(networks.get("service"))),
+    ]
+
+    if len(object_networks) > 1 and object_networks[0][1].overlaps(
```
TeddyAndrieux: I don't see how len of `object_networks` could be different than 2

salt/_modules/metalk8s_checks.py
```diff
@@ -16,6 +16,60 @@ def __virtual__():
...
+        object_networks[1][1]
+    ):
+        errors.append(" - pods and services")
```
TeddyAndrieux: Could be cool to have the actual CIDR in the error message (also why "Control Plane" and "Workload Plane" have capital and not "pods" and "services")

salt/_modules/metalk8s_checks.py
```diff
@@ -16,6 +16,60 @@ def __virtual__():
...
+
+    for o in object_networks:
+        for n in node_networks:
+            if any(o[1].overlaps(net) for net in n[1]):
+                errors.append(f" - {o[0]} and {n[0]}")
```
TeddyAndrieux, line 58: ditto: could be cool to have the actual CIDR that conflicts

salt/tests/unit/modules/files/test_metalk8s_checks.yaml
```diff
@@ -1,3 +1,66 @@
+bootstrap_file:
+  # 1. Success: Nominal
+  - &bstrapf_nominal
+    cp_cidrs: [10.0.0.0/16]
+    wp_cidrs: [10.0.0.0/16]
```
TeddyAndrieux: let's consider we have 2 distinct network (I mean CP != WP)  as nominal

salt/_modules/metalk8s_checks.py
eg-ayoub: this was a remnant from the first draft - im removing it

salt/tests/unit/modules/files/test_metalk8s_checks.yaml
```diff
@@ -1,3 +1,64 @@
+bootstrap_file:
+  # 1. Success: Nominal
+  - &bstrapf_nominal
+    cp_cidrs: [10.1.0.0/16]
+    wp_cidrs: [10.2.0.0/16]
+    pods_cidr: 10.233.0.0/16
+    svc_cidr: 10.96.0.0/12
+
+  # 2. Fail: pods intersect fully
```
TeddyAndrieux, line 9: Can we have at least one "not obvious" conflict like `10.96.0.0/12` and `10.100.0.0/16` because here we only have conflicts where there are strictly equals

TeddyAndrieux (approved):

PR #4234, opened 2024-02-06
charts,salt,build: Bump kube-prometheus-stack chart to 56.6.2

ezekiel-alexrod:
Bump the kube-prometheus-stack chart to 56.6.2 and also bump grafana image to 10.3.1

Upgrade the kube-prometheus-stack chart using:
```
rm -rf charts/kube-prometheus-stack
helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
helm repo update
helm fetch -d charts --untar prometheus-community/kube-prometheus-stack
```

Re-render prometheus-operator salt state using:
```
./doit.sh codegen:chart_kube-prometheus-stack
```


salt/metalk8s/addons/prometheus-operator/deployed/chart.sls
```diff
@@ -50376,15 +51607,15 @@ metadata:
...
+    app.kubernetes.io/version: 10.3.1
+    helm.sh/chart: grafana-7.3.0
     heritage: metalk8s
   name: prometheus-operator-grafana
   namespace: metalk8s-monitoring
 spec:
   ingressClassName: nginx-control-plane
   rules:
-  - host: null
+  - host: 'null'
```
TeddyAndrieux, line 51618: I think this will cause trouble, I see they changed the template to use `| quote` filter so we should change this so that it's an empty string

TeddyAndrieux (commented):
From salt changes it seems there is some other images that also need to be bumped including Thanos, so I suggest to also bump Thanos helm chart

You will also need to include the new extract of alerts rules since there is quite some changes there

TeddyAndrieux (approved):

PR #4250, opened 2024-02-29
:lock::arrow_up: bump thanos version to 0.34.1

m4nch0t:
**Component**:

Thanos

**Context**: 

CVE

**Summary**:

FIx CVE, according to :
https://github.com/thanos-io/thanos/pull/7078

**Acceptance criteria**:


charts/kube-prometheus-stack/values.yaml
```diff
@@ -2712,7 +2712,7 @@ prometheusOperator:
   thanosImage:
     registry: quay.io
     repository: thanos/thanos
-    tag: v0.34.0
+    tag: v0.34.1
```
TeddyAndrieux, line 2800: This file come from the helm chart so it shouldn't be updated, you need to update the full charts using the `helm` commands

Basically we do not update by hand any files in `charts/<name>/*` (just the options file so `charts/<name>.yaml`)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Enhancements
 
+- Bump thanos version to
+  [0.34.1](https://github.com/thanos-io/thanos/releases/tag/v0.34.1)
+  (PR[#4250](https://github.com/scality/metalk8s/pull/4250))
```
TeddyAndrieux: Since there is already a mention of thanos in this `127.0.0` I would prefer you update this one, it's a bit weird to have 2 bump of the same image in the changelog of one version

charts/kube-prometheus-stack/values.yaml
m4nch0t: Yes, it's only a test to check if CI failed as my newly devcontainer with go.
I've supposed that my changes with kube-prometheus-stack breaks everyhing :/
You said that you fixed that in other commit, so I've rebased and will push it

CHANGELOG.md
m4nch0t: It was weird for me too, but I don't know if I could mix multiple PR in changelog.

PR #4311, opened 2024-04-12
salt: Add the `grains_cache` option to Salt minions 

ezekiel-alexrod:
Force the sync of the custom set grains. Relates to https://github.com/saltstack/salt/issues/54331


TeddyAndrieux (commented):
If you may just add a changelog entry and then LGTM

TeddyAndrieux (approved):

PR #4313, opened 2024-04-16
Hide diff when creating a Secret object from salt state

eg-ayoub:


salt/_states/metalk8s_kubernetes.py
```diff
@@ -159,7 +159,11 @@ def object_present(name, manifest=None, **kwargs):
         name=name_arg, manifest=manifest, old_object=obj, saltenv=__env__, **kwargs
     )
     diff = __utils__["dictdiffer.recursive_diff"](obj, new)
-    ret["changes"] = diff.diffs
+    # Anonymize diff when object has `kind: Secret`
+    if manifest and manifest.get("kind") == "Secret":
```
TeddyAndrieux: I would use `new` directly instead of `manifest`, normally `new` is always filled

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Enhancements
 
+- Anonymize diff when updating a Kubernetes Secret object from salt state
+  (PR[#4313](https://github.com/scality/metalk8s/pull/4313))
```
TeddyAndrieux: If you may just move it at the end of Enhancements (I think it's better to have the bumps first)

TeddyAndrieux (approved):

PR #4317, opened 2024-04-18
salt,tests: Handle a 409 conflict error when using `metalk8s_kubernetes.object_present`

ezekiel-alexrod:
Handle a 409 conflict error when using `metalk8s_kubernetes.object_present` Salt state due to another component modifying the wanted object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Release 127.0.2 (in development)
 
+### Enhancements
+
+- Handle a 409 conflict error when using
+  `metalk8s_kubernetes.object_present` Salt state
+  due to another component modifying the wanted object
+  (PR[#4317](https://github.com/scality/metalk8s/pull/4317))
+
 ## Release 127.0.1
 
 ### Enhancements
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -59,6 +59,12 @@ def _handle_error(exception, action):
         and exception.status == 404
     ):
         return None
+    elif (
+        action == "replace"
+        and isinstance(exception, ApiException)
+        and exception.status == 409
+    ):
+        raise CommandExecutionError("409 Conflict") from exception
     else:
         raise CommandExecutionError(base_msg) from exception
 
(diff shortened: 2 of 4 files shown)


salt/_states/metalk8s_kubernetes.py
```diff
@@ -158,6 +159,25 @@ def object_present(name, manifest=None, **kwargs):
     new = __salt__["metalk8s_kubernetes.replace_object"](
         name=name_arg, manifest=manifest, old_object=obj, saltenv=__env__, **kwargs
     )
+    # Handle the conflict error by getting the most up to date object
+    # and retrying
+    if new == "409":
+        for _ in range(5):
```
TeddyAndrieux: I would prefer to have this loop on the whole state, because let's imagine the objects get deleted and not just updated (I think there is no real case like that, but I still prefer to run the whole state logic on each loop)

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -226,6 +226,9 @@ def method(
         try:
             result = method_func(**call_kwargs)
         except (ApiException, HTTPError) as exc:
+            # Handle 409 conflict error for replace action, send it back to the state
+            if exc.status == 409:
+                return "409"
```
TeddyAndrieux: From what I see HTTPError does not have a `status` field

I would prefer to have this logic in `handle_error` directly and check that the exception is `ApiException`

And also I do not really like the simple return of a string, I would prefer to raise a classic salt exception and catch it based on the message in the state, something like that (not tested)

``` python
# in the module
raise CommandExecutionError("409 Conflict") from exception

# in the state
try:
    new = <classic replace call ...>
    if exc.message == "409 Conflict":
        # LET'S RETRY !!
        continue
    raise
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Release 127.0.2 (in development)
 
+- Handle a 409 conflict error when using
```
TeddyAndrieux: Can you add the `Enhancements` section

salt/_states/metalk8s_kubernetes.py
```diff
@@ -104,66 +106,81 @@ def object_present(name, manifest=None, **kwargs):
...
+            if exc.message == "409 Conflict":
+                # Retry the state
+                continue
+            raise
+
+        diff = __utils__["dictdiffer.recursive_diff"](obj, new)
+        ret["changes"] = diff.diffs
+        ret["comment"] = "The object was replaced"
 
         return ret
```
TeddyAndrieux, line 184: You also have to return/raise something in case all the 5 retry fail

TeddyAndrieux (approved):
LGTM

PR #4336, opened 2024-06-06
kubelet: add `containerLogMaxSize` to the configuration

ezekiel-alexrod:
And bumping the default from `10Mi` to `50Mi`


TeddyAndrieux (approved):
It could be configurable at some point but I think it's still small enough to be the default so let's go !

PR #4427, opened 2024-09-19
Spread metalk8s nightly runs for each branch

aprucolimartins:
Spread metalk8s nightly runs for each branch.

### DIFF
--- a/.github/workflows/crons.yaml
+++ b/.github/workflows/crons.yaml
@@ -4,7 +4,11 @@ on:
   # NOTE: Schedule is only run on the latest commit of the default branch,
   # so we need to schedule Cron Jobs for all branches here
   schedule:
-    - cron: "1 0 * * 1-5"
+    # Run V 0 nightly every weekday at 23:00
+    - cron: "0 23 * * 1-5"
+
+    # Run V -1 nightly every weekday at 00:00
+    - cron: "0 0 * * 1-5"
 
 jobs:
   crons:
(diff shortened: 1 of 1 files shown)


.github/workflows/crons.yaml
```diff
@@ -4,19 +4,25 @@ on:
   # NOTE: Schedule is only run on the latest commit of the default branch,
   # so we need to schedule Cron Jobs for all branches here
   schedule:
-    - cron: "1 0 * * 1-5"
+    # Run 129.0 nightly every weekday at 23:00
+    - cron: "0 23 * * 1-5"
+
+    # Run 128.0 nightly every weekday at 00:00
+    - cron: "0 0 * * 1-5"
```
TeddyAndrieux: Let's call it "branch 1" and "branch 2" instead so that we do not have to update it every time we bump

.github/workflows/crons.yaml
```diff
@@ -4,19 +4,25 @@ on:
...
 jobs:
   crons:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        include:
+        jobs:
           - name: "Nightly for 129.0"
+            cron: "0 23 * * 1-5"
```
TeddyAndrieux, line 21: It's not "auto magic" you have to handle the `cron` in the `steps` :)

.github/workflows/crons.yaml
```diff
@@ -4,19 +4,25 @@ on:
...
+    - cron: "0 0 * * 1-5"
 
 jobs:
   crons:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        include:
+        jobs:
```
TeddyAndrieux: You can keep `include` here, so that you avoid an extra layer "jobs"

TeddyAndrieux (changes requested):

.github/workflows/crons.yaml
aprucolimartins: I put V 0, V -1, ... What do you think?

.github/workflows/crons.yaml
aprucolimartins: Oh Merlin!!!

.github/workflows/crons.yaml
aprucolimartins: Thanks for the tip :)

.github/workflows/crons.yaml
TeddyAndrieux: Yep it's good to me

TeddyAndrieux (approved):

PR #4463, opened 2024-10-28
buildchain: test pulling image from cgr

eg-ayoub:
test1


.github/workflows/pre-merge.yaml
```diff
@@ -24,6 +24,10 @@ on:
       - "w/**"
       - "q/*/**"
 
+permissions:
+  contents: read
+  id-token: write
+
```
eg-ayoub, line 30: remove

PR #4489, opened 2024-11-29
Add a new workflow to create a dev branch

eg-ayoub:


.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+          echo "NEW_VERSION_MINOR=$NEW_VERSION_MINOR" >> $GITHUB_ENV
+      # calculate new branch name
+      - name: Calculate new branch name
+        run: |
+          NEW_BRANCH_NAME="development/${{ env.NEW_SHORT_VERSION }}"
+          INIT_BRANCH_NAME="feature/init-dev-${{ env.NEW_SHORT_VERSION }}"
+
+          echo "NEW_BRANCH_NAME=$NEW_BRANCH_NAME" >> $GITHUB_ENV
+          echo "INIT_BRANCH_NAME=$INIT_BRANCH_NAME" >> $GITHUB_ENV
+      # create development branch
```
TeddyAndrieux, line 63: Let's ensure the NEW_BRANCH does not exist already (like there is `development/129.0` you try to trigger create new major from `development/128.0` (I know that the push (or checkout -b) should fail but it's better to have a proper error message and no try to push to see that we have an issue)

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+          git checkout -b "${{ env.INIT_BRANCH_NAME }}"
+      # find mentions of new version: warn user
+      - name: Look for important messages
+        id: warn_user
+        run: |
+          link="https://github.com/${{ github.repository }}/blob/${{ github.ref_name }}/"
+          {
+            echo 'COMMENT<<EOF'
+            echo "# Mentions of ${{ env.NEW_SHORT_VERSION }} in ${{ github.ref_name }}"
+            git grep -rHn "${{ env.NEW_SHORT_VERSION }}" | sed -E "s;(.*):([0-9]+):.*;$link\1#L\2;"
```
TeddyAndrieux: Add `-F` to avoid the `.` matching anything

And to me instead of this quite ugly sed :smile: we could do a simple `cut` like `| cut -d':' -f1,2`

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+        run: |
+          git checkout -b "${{ env.INIT_BRANCH_NAME }}"
+      # find mentions of new version: warn user
+      - name: Look for important messages
+        id: warn_user
+        run: |
+          link="https://github.com/${{ github.repository }}/blob/${{ github.ref_name }}/"
+          {
+            echo 'COMMENT<<EOF'
+            echo "# Mentions of ${{ env.NEW_SHORT_VERSION }} in ${{ github.ref_name }}"
```
TeddyAndrieux: To me it would be better to link on the new development branch (since it's on this branch we will want to update it)

```suggestion
          link="https://github.com/${{ github.repository }}/blob/${{ env.NEW_BRANCH_NAME }}/"
          {
            echo 'COMMENT<<EOF'
            echo "# Mentions of ${{ env.NEW_SHORT_VERSION }} in ${{ env.NEW_BRANCH_NAME }}"
```

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+          fetch-depth: 0
+          token: ${{ steps.app-token.outputs.token }}
+      # get version
+      - name: Get current version
+        run: |
+          source VERSION
+
+          SHORT_VERSION="$VERSION_MAJOR.$VERSION_MINOR"
+
+          echo "SHORT_VERSION=$SHORT_VERSION" >> $GITHUB_ENV
```
TeddyAndrieux: This does not seems used

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+      - name: add new entry to changelog
+        run: |
+          sed -i "s/# CHANGELOG/# CHANGELOG\n\n## Release ${{ env.NEW_SHORT_VERSION }}.0 (in development)/" CHANGELOG.md
+      # edit cron jobs
+      - name: bump version in cron jobs
+        run: |
+          current=$(grep current .github/workflows/crons.yaml | sed -E "s/\s*# current=//")
+          old=$(grep old .github/workflows/crons.yaml | sed -E "s/\s*# old=//")
+          sed -i "s/$current/${{ env.NEW_SHORT_VERSION }}/g" .github/workflows/crons.yaml
+          sed -i "s/$old/$current/g" .github/workflows/crons.yaml
```
TeddyAndrieux, line 111: Ok fine it's not really beautiful but it's O K

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,142 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr = await github.rest.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: "${{ env.INIT_BRANCH_NAME }}",
+              base: "${{ env.NEW_BRANCH_NAME }}",
+              title: "Initialize ${{ env.NEW_BRANCH_NAME }} branch"
+            });
```
TeddyAndrieux, line 131: You also need to add a comment with `/approve`

.github/workflows/create-dev-branch.yaml
eg-ayoub: ok about the `-F` but the cut doesn't help me with making links from the output https://github.com/scality/gha-playground/pull/33#issuecomment-2507556059 like this

TeddyAndrieux: Oh yes right you also want to replace the `:` with `#` I see could be cut + sed then but indeed a single sed is better then

.github/workflows/create-dev-branch.yaml
eg-ayoub: done in next step

.github/workflows/create-dev-branch.yaml
```diff
@@ -0,0 +1,146 @@
...
+          NEW_BRANCH_NAME="development/${{ env.NEW_SHORT_VERSION }}"
+          INIT_BRANCH_NAME="feature/init-dev-${{ env.NEW_SHORT_VERSION }}"
+
+          echo "NEW_BRANCH_NAME=$NEW_BRANCH_NAME" >> $GITHUB_ENV
+          echo "INIT_BRANCH_NAME=$INIT_BRANCH_NAME" >> $GITHUB_ENV
+      # create development branch
+      - name: Create new branch
+        run: |
+          git ls-remote --exit-code --heads origin refs/heads/${{ env.NEW_BRANCH_NAME }}
+          if [[ $? = 0 ]]; then
```
TeddyAndrieux: nit

```suggestion
          if git ls-remote --exit-code --heads origin refs/heads/${{ env.NEW_BRANCH_NAME }}; then
```

TeddyAndrieux (approved):
LGTM good job !

PR #4508, opened 2025-01-21
salt: re-generate manifests after renew (scheduler, controller manager)

eg-ayoub:


salt/metalk8s/defaults.yaml
```diff
@@ -296,6 +296,7 @@ certificates:
         renew:
           sls:
             - metalk8s.backup.certs.server
+            - metalk8s.backup.deployed
```
eg-ayoub: may be too broad, should I just regenerate secret-tls and the deployment ?

TeddyAndrieux: No but the issue here is that the `deployed` SLS needs to be run from the salt master and not from the minions, so this will not work you have to use `post.orch` like for ingress controllers

Btw for control-plane and workload-plane ingresses the renewable should also restart the pods and it's not the case today.
And I feel it's the same for salt-master and apiserver

eg-ayoub: - [x] backup
- [x] ingresses
- [x] salt-master
- [ ] apiserver ( already set to install ? )
- [x] dex

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # CHANGELOG
 
-## Release 128.0.1 (in development)
+## Release 128.0.2
```
TeddyAndrieux: In dev is missing

So the release workflow did not worked properly ?

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # CHANGELOG
 
-## Release 128.0.1 (in development)
+## Release 128.0.3 (in development)
```
TeddyAndrieux, line 3: Hmm so there is nothing in 128.0.2 :cry:

eg-ayoub: so it seems after release 128.0.1 the changelog was not changed.
the workflow (which I used to release 128.0.2) uses sed so it expects the last version to be correct.
since it didn't find `128.0.2 in development` in the changelog, it wasn't changed. 
I found that as a result the changes for 128.0.2 had their logs in the 128.0.1 mistakenly.
I will add a new commit to repair this.

TeddyAndrieux (approved):

eg-ayoub:
ping

PR #4514, opened 2025-02-07
salt: upgrade apiserver on bootstrap first

eg-ayoub:


salt/metalk8s/orchestrate/apiserver.sls
```diff
@@ -1,7 +1,14 @@
 {%- set dest_version = pillar.metalk8s.cluster_version %}
 {%- set master_nodes = salt.metalk8s.minions_by_role('master') %}
+{%- set bootstrap_nodes = salt.metalk8s.minions_by_role('bootstrap') %}
```
TeddyAndrieux: Can you sort those 2 so that we are sure the order is always the same between 2 runs of the upgrade

TeddyAndrieux (approved):

PR #4575, opened 2025-04-16
Improvement/bump containerd 1.6.38

eg-ayoub:


packages/redhat/8/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN useradd -m build
 
 ARG SALT_VERSION
 ARG K8S_SHORT_VERSION
-ARG GO_VERSION=1.19.13
+ARG GO_VERSION=1.24.2
```
TeddyAndrieux: Let's only move to latest 1.20 since it's the version used by this containerd version

TeddyAndrieux (approved):

PR #4590, opened 2025-04-29
salt: add a hash annotation to force restart backup server on certificate renewal

ezekiel-alexrod:


eg-ayoub (dismissed):

salt/metalk8s/backup/deployed/deployment.sls
```diff
@@ -18,6 +18,9 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        checksum/config: __slot__:salt:metalk8s_kubernetes.get_object_digest(kind="ConfigMap",
+          apiVersion="v1", namespace="kube-system", name="backup-ca-cert", path="data")
```
TeddyAndrieux: The ca cert will not be renewed quite often here we should focus on the tls cert so the secret `backup-tls` on `backup.crt` only (because we don't want to be able to retrieve the private key from the annotation for security reason :) )

vdaviot (dismissed):

TeddyAndrieux (approved):

rdebay-scality (approved):

PR #4618, opened 2025-07-03
rework micro app routing

MonPote:
**Context**: 

The goal of this PR is to prepare the ability to load several versions of the same micro app and be able to load micro app from a micro app.
To achieve that, we need to  : 
- Be able to rely only on the discoveryUrl to render the Route instead of mixing it with the value of the config.json.
- Rework the way we render the micro app's `Routes`
 (see: `InternalRouter` in `shell-ui/src/FederatedApp.tsx`)

In the previous implementation we use `useDiscoveredViews` to render the Routes and render the navbar entries.
This was convenient because we reuse the hooks for the rendering the `Navbar` (see: `useFederatedNavbarEntries`) and rendering the routes (DRY) but also cause some unnecessary coupling.
The route need to be exposed in the navbar in order to be accessible in the UI.

In a close futur, we want to be able to render some FederatedComponent without it to be exposed in the navbar.
For example, we might want to switch micro app with a selector instead of click on the navbar.

Here are some notable change made : 
- we keep `useDiscoveredViews` for the navbar entries
- we create `useFederatedRoutes` to compute the federatedRoutes
- we add the url from where the micro app has been fetch in the metadata (cf: `EnrichedBuildtimeWebFinger`)in order to be able to find the right micro app config since we can now have several time the same app.

Those changes simplify a lot the `FederatedApp.tsx` however we lost the capability to render only `Routes` which the user have access because this information is only present on the config.json. It means that we rely on the API to filter what the user can or not do (it's already the case). The Navbar will still render only the correct entries though.

I've also add a checking to warn the user when `appHistoryBasePath` is not correct.

### DIFF
--- a/shell-ui/src/initFederation/ConfigurationProviders.tsx
+++ b/shell-ui/src/initFederation/ConfigurationProviders.tsx
@@ -353,6 +353,14 @@ export function useFederatedRoutes(): FederatedRoute[] {
   const { retrieveDeployedApps } = useDeployedAppsRetriever();
   const deployedApps = retrieveDeployedApps();
 
+  deployedApps.forEach((app) => {
+    if (app.appHistoryBasePath.endsWith('/')) {
+      throw new Error(
+        `appHistoryBasePath of app ${app.name} ends with a /, this is not allowed`,
+      );
+    }
+  });
+
   const federatedRoutes = deployedApps.flatMap((app) => {
     const appBuildConfig = retrieveConfiguration<'build'>({
       configType: 'build',
--- a/shell-ui/src/initFederation/ConfigurationProviders.tsx
+++ b/shell-ui/src/initFederation/ConfigurationProviders.tsx
@@ -64,11 +64,21 @@ export type BuildtimeWebFinger = {
     instanceNameAdapter?: FederatedModuleInfo;
   };
 };
+
+export type EnrichedBuildtimeWebFinger = BuildtimeWebFinger & {
+  metadata: {
+    // This information is not present in the original webFinger, it's filled by the ConfigurationProvider
+    url: string;
+  };
+};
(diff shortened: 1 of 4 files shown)


shell-ui/src/initFederation/ConfigurationProviders.tsx
```diff
@@ -318,9 +341,52 @@ export function useDiscoveredViews(): ViewDefinition[] {
...
+  const { retrieveConfiguration } = useConfigRetriever();
+  const { retrieveDeployedApps } = useDeployedAppsRetriever();
+  const deployedApps = retrieveDeployedApps();
+
+  deployedApps.forEach((app) => {
+    if (app.appHistoryBasePath.endsWith('/')) {
+      throw new Error(
+        `appHistoryBasePath of app ${app.name} ends with a /, this is not allowed`,
+      );
+    }
```
JBWatenbergScality: Do we really want to run this on each render of the component holding this hook ? I think we can at least move a good part of this hook content into a useMemo

shell-ui/src/initFederation/ConfigurationProviders.tsx
```diff
@@ -318,9 +341,52 @@ export function useDiscoveredViews(): ViewDefinition[] {
...
+  const federatedRoutes = deployedApps.flatMap((app) => {
+    const appBuildConfig = retrieveConfiguration<'build'>({
+      configType: 'build',
+      name: app.name,
+      url: app.url,
+    });
+
+    if (!appBuildConfig) {
+      return [];
+    }
```
JBWatenbergScality: nit: we are iterating twice on deployedApps within the same scope, can we iterate only once on this collection?

JBWatenbergScality (approved):

PR #4621, opened 2025-07-10
sos: add DNS troubleshooting outputs

ezekiel-alexrod:


packages/common/metalk8s-sosreport/metalk8s.py
```diff
@@ -135,6 +134,15 @@ def _setup_common(self):
         for service in services:
             self.add_journal(units=service)
 
+        # Add DNS troubleshooting outputs
```
TeddyAndrieux, line 137: Maybe just a quick comment to explain why we use bash -c :)

TeddyAndrieux (approved):

PR #4636, opened 2025-08-04
use ui-operator for shell-ui

MonPote:
**Component**:

**Context**: 

**Summary**:

**Acceptance criteria**: 

---

Closes: #ISSUE_NUMBER


salt/_pillar/metalk8s.py
```diff
@@ -168,6 +168,7 @@ def _load_addons(config_data):
     addons_data.setdefault("dex", {}).setdefault("enabled", True)
     addons_data.setdefault("loki", {}).setdefault("enabled", True)
     addons_data.setdefault("fluent-bit", {}).setdefault("enabled", True)
+    addons_data.setdefault("shell-ui", {}).setdefault("enabled", True)
```
TeddyAndrieux: You never used this flag from what I see so you should remove it no ?

salt/metalk8s/addons/shell-ui/deployed/ui-operator-cr.sls
```diff
@@ -0,0 +1,77 @@
...
+          navbar:
+            main:
+              - internal:
+                  kind: metalk8s-ui
+                  view: platform
+                  groups: [metalk8s:admin]
+              - internal:
+                kind: metalk8s-ui
+                view: alerts
+                groups: [metalk8s:admin]
```
TeddyAndrieux: To me you want/need and if/else here so that you create those if they do not exists otherwise you only patch the image field (and most likely redirectUrl as well since you retrieve it with salt)

Otherwise it means during upgrade you will override every tuning you did on the objects

salt/_pillar/metalk8s.py
MonPote: Indeed thx, I will remove it.

TeddyAndrieux (dismissed):
Salt part LGTM for the UI related changes I cannot say anything

TeddyAndrieux (approved):

PR #4661, opened 2025-09-24
Refactor chart components to utilize LineTimeSerieChart

JeanMarcMilletScality:
The goal of this PR is to replace line chart with new core-ui version
Include only the "simple" charts 
The transformation data were not changed as they are also use by quantile chart

Next step: refactor the quantile chart and the data transformation fn to reflect new types for line chart

~~Still a draft as a new release of core-ui is needed~~

### DIFF
--- a/ui/src/components/MetricChart.tsx
+++ b/ui/src/components/MetricChart.tsx
@@ -1,14 +1,20 @@
-import { useCallback } from 'react';
+import React, { useCallback, useMemo } from 'react';
 import type { UseQueryOptions } from 'react-query';
 import 'react-query';
 import {
   LineTimeSerieChart,
+  useChartId,
   useMetricsTimeSpan,
 } from '@scality/core-ui/dist/next';
 import { convertPrometheusResultToSerieWithAverage } from '../services/graphUtils';
-import { HEIGHT_DEFAULT_CHART, NODE_SYNC_ID } from '../constants';
+import {
+  CLUSTER_AVERAGE,
+  HEIGHT_DEFAULT_CHART,
+  NODE_SYNC_ID,
+} from '../constants';
 import { useChartSeries } from '../hooks';
 import { TimeSpanProps } from '../services/platformlibrary/metrics';
+import { useChartLegendRegistration } from '../hooks/useChartLegendRegistration';
 
 const MetricChart = ({
   title,
@@ -38,6 +44,15 @@ const MetricChart = ({
     label: string;
   }[];
 }) => {
+  console.log(
+    'DEBUG MetricChart',
+    title,
+    yAxisType,
+    nodeName,
+    instanceIP,
+    showAvg,
+  );
+  const chartId = useChartId();
   const { interval, duration } = useMetricsTimeSpan();
   const { isLoading, series, startingTimeStamp } = useChartSeries({
     getQueries: useCallback(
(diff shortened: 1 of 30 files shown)


ChengYanJin (dismissed):
LGTM !

ui/src/components/MetricChart.tsx
```diff
@@ -38,6 +44,15 @@ const MetricChart = ({
...
 }) => {
+  console.log(
+    'DEBUG MetricChart',
+    title,
+    yAxisType,
+    nodeName,
+    instanceIP,
+    showAvg,
+  );
+  const chartId = useChartId();
```
ChengYanJin: oupsi

ui/src/hooks/useChartLegendRegistration.ts
```diff
@@ -0,0 +1,61 @@
+import { useEffect } from 'react';
+import { useChartLegend } from '@scality/core-ui/dist/components/chartlegend/ChartLegendWrapper';
+import { Serie } from '@scality/core-ui/dist/components/linetimeseriechart/linetimeseriechart.component';
+
+interface SymmetricalSeries {
+  above: Serie[];
+  below: Serie[];
+}
```
ChengYanJin: Could we use type here ?

ui/src/hooks/useChartLegendRegistration.ts
```diff
@@ -0,0 +1,62 @@
...
+      series: Serie[] | null;
+      isSymmetrical: false;
+    }
+  | {
+      series: SymmetricalSeries | null;
+      isSymmetrical: true;
+    }
+);
+
+export const useChartLegendRegistration = ({
```
ChengYanJin, line 23: Should we add some test for this hook ? 
Can be done in the next PR.

ui/src/hooks/useChartLegendRegistration.ts
```diff
@@ -0,0 +1,62 @@
...
+        symmetricalSeries.above?.length > 0 ||
+        symmetricalSeries.below?.length > 0
+      ) {
+        const aboveNames =
+          symmetricalSeries.above?.map((s) => s.resource) || [];
+        const belowNames =
+          symmetricalSeries.below?.map((s) => s.resource) || [];
+        seriesNames = [...aboveNames, ...belowNames];
+      }
+    } else if (isSymmetrical === false) {
```
ChengYanJin, line 49: Nit: I think the else if might be redundant condition check ? could be directly else ?

ui/src/hooks/useChartLegendRegistration.ts
```diff
@@ -0,0 +1,62 @@
+import { useEffect } from 'react';
+import { useChartLegend } from '@scality/core-ui/dist/components/chartlegend/ChartLegendWrapper';
+import { Serie } from '@scality/core-ui/dist/components/linetimeseriechart/linetimeseriechart.component';
+
+interface SymmetricalSeries {
+  above: Serie[];
+  below: Serie[];
+}
+type ChartLegendRegistation = {
```
ChengYanJin: Small typo in the type name `Registration` ;)

ui/src/hooks/useChartLegendRegistration.ts
JeanMarcMilletScality: Yes I think so too, but I had issue with type inference. TS was not happy with the else

ChengYanJin:
```
else {
      const arrayedSeries = series as Serie[];
      if (arrayedSeries && arrayedSeries.length > 0) {
        seriesNames = arrayedSeries.map((s) => s.resource);
      }
    }
```

ChengYanJin (approved):

PR #4750, opened 2026-01-15
Bump `k8s.io` go libraries to `0.33.7` and go version to `1.25.5`

ezekiel-alexrod:


ezekiel-alexrod:
Closing for a couple of reasons:
- the branch is misnamed and does not reflect the content of the PR
- the order to bump kubernetes in the project is not optimal

Hence, this PR will be renewed when appropriate

PR #4792, opened 2026-02-23
charts,salt: bump Loki chart version to 6.53.0 and image to 3.6.5

ezekiel-alexrod:
Closes: MK8S-101


salt/metalk8s/addons/logging/loki/deployed/chart.sls
```diff
@@ -175,6 +175,8 @@ spec:
       annotations:
         checksum/config: __slot__:salt:metalk8s_kubernetes.get_object_digest(kind="Secret",
           apiVersion="v1", namespace="metalk8s-logging", name="loki", path="data:config.yaml")
+        kubectl.kubernetes.io/default-container: loki
+        storage/size: 10Gi
```
TeddyAndrieux, line 179: Did you check what the usage of this annotation ?

.gitignore
```diff
@@ -33,3 +33,7 @@ junit/
 /_build/
 
 vagrant_config.rb
+logs_57921269207
+logs_58008460735
+logs_58031326724
+logs_58249689133
```
TeddyAndrieux: 🤔

ezekiel-alexrod: oops

salt/metalk8s/addons/logging/loki/deployed/chart.sls
ezekiel-alexrod: What should I check ? I just did `./doit.sh codegen`

I went to the loki repo and it says https://github.com/grafana/loki/blob/v3.6.5/production/helm/loki/templates/single-binary/statefulset.yaml#L46
```
storage/size: {{ .Values.singleBinary.persistence.size | quote }}
```

TeddyAndrieux: Yes ok I was just thinking if Loki would do something about the annotation but since it's not an operator but a "simple" sts he don't I think

Ok fine

salt/metalk8s/addons/logging/loki/deployed/chart.sls
```diff
@@ -257,9 +237,37 @@ spec:
...
+          value: /rules
+        - name: RESOURCE
+          value: both
+        - name: WATCH_SERVER_TIMEOUT
+          value: '60'
+        - name: WATCH_CLIENT_TIMEOUT
+          value: '60'
+        - name: LOG_LEVEL
+          value: INFO
+        image: docker.io/kiwigrid/k8s-sidecar:1.30.9
```
TeddyAndrieux: No linked to this PR but it's not good, we should use build_image_name here like for loki image

It mean since we have this image the pod cannot work on offline deployment (weird that we do not catch it in pre-merge tests)

ezekiel-alexrod: Should I make the changes now? 
Why isn't it catched by the CI ?

TeddyAndrieux: > Should I make the changes now?

To me yes it could be good otherwise we will forget about it

> Why isn't it catched by the CI ?

Just checked and we are not offline in ci test here :confused:

TeddyAndrieux (approved):

anthony-treuillier-scality (approved):

PR #4793, opened 2026-02-24
MK8S-139 : Extend the authentication section to PrometheusConfig and AlertManagerConfig

ChengYanJin:
**Component**: Salt

**Context**:  
The `PrometheusConfig` CRD will be extended with a new `enable_oidc_authentication` field and an `oidc` sub-configuration block.

This allows operators to declaratively configure OIDC authentication for the Prometheus service alongside the existing configuration.

The same apply for `AlertManagerConfig`.

**Summary**: 
This PR only extends the configuration. The next PR will be using Salt to deploy the required resources based on the values defined here.

**Acceptance criteria**: 

Ref: https://github.com/scality/citadel/pull/306

### DIFF
--- a/salt/metalk8s/addons/prometheus-operator/config/alertmanager.yaml
+++ b/salt/metalk8s/addons/prometheus-operator/config/alertmanager.yaml
@@ -13,6 +13,13 @@ spec:
   # Configure the Alertmanager Deployment
   deployment:
     replicas: 1
+  config:
+    enable_oidc_authentication: false
+    oidc:
+      issuer: "" # OIDC provider URL
+      audience: "" # Expected audience claim in the token
+      groupsClaim: "" # JWT claim name that carries user groups/roles
+      authorizedGroups: [] # Groups/roles allowed to access Alertmanager
   notification:
     config:
       global:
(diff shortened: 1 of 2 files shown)


salt/metalk8s/addons/prometheus-operator/config/alertmanager.yaml
```diff
@@ -13,6 +13,14 @@ spec:
   # Configure the Alertmanager Deployment
   deployment:
     replicas: 1
+  config:
+    enable_oidc_authentication: false
+    oidc:
+      issuer: "" # OIDC provider URL
+      audience: "" # Expected audience claim in the token
+      ca: "" # Path to CA certificate used to verify the issuer TLS certificate
```
JBWatenbergScality: I'm not sure about this one. Maybe it should remains internal

JBWatenbergScality: The rational is that user cannnot "mount" anything in current config here. And we will limit to issuer exposed on control plane. Metalk8s has ownership of the CA for control plane, so this can be a constraint and an internal detail

ChengYanJin: Yes, indeed, we will use a sidecar to discover the CA secret and mount it.

JBWatenbergScality (approved):

JeanMarcMilletScality (approved):

PR #4795, opened 2026-02-25
MK8S-140 - Salt deploy oauth2-proxy

ChengYanJin:
**Component**: Salt

## Context

The `PrometheusConfig` and `AlertManagerConfig` CRDs have been extended (in a previous PR) with an `enable_oidc_authentication` flag and an `oidc` configuration block.

This PR deploys the Salt resources needed to support OIDC authentication for Prometheus and Alertmanager.

### Stable proxy services (always created)

Instead of having the UI state conditionally point to different backends, this PR introduces two stable `ExternalName` services — always created by the prometheus-operator Salt state — that act as a routing layer:

| Service | OIDC enabled | OIDC disabled |
|---|---|---|
| `prometheus-proxy` | → `oauth2-proxy-prometheus` | → `thanos-query-http` |
| `alertmanager-proxy` | → `oauth2-proxy-alertmanager` | → `prometheus-operator-alertmanager` |

The UI Salt state always points to `prometheus-proxy` and `alertmanager-proxy`. **Toggling OIDC on/off only requires re-applying the prometheus-operator Salt state** — the UI state never needs to change.

### oauth2-proxy deployment (conditional)

When OIDC is enabled, a `oauth2-proxy` Deployment and Service are created for each of Prometheus and Alertmanager.

### RBAC (`oidc-proxy-rbac.sls`)

ServiceAccounts, Roles, and RoleBindings that allow the oauth2-proxy pods to read the CA certificate Secret from the configured namespace.

### Configurable CA certificate

A `caSecret` field has been added to the OIDC config in `prometheus.yaml` and `alertmanager.yaml`, so operators can point to any Kubernetes Secret containing the CA certificate. Defaults to the control plane ingress certificate in `metalk8s-ingress`.

## Future Work

- [ ] SLS Template Rendering Tests (Unit Tests)
- [x] Add oauth2-proxy image to MetalK8s build chain
- [ ] Restart script
- [ ] UI update to include auth token
- [ ] Documentation

## Acceptance Criteria

Ref: https://github.com/scality/citadel/pull/306

### DIFF
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -1,4 +1,5 @@
 {%- from "metalk8s/repo/macro.sls" import build_image_name with context %}
+{%- from "metalk8s/map.jinja" import coredns with context %}
 
 {%- set alertmanager_defaults = salt.slsutil.renderer(
         'salt://metalk8s/addons/prometheus-operator/config/alertmanager.yaml',
@@ -126,3 +127,25 @@ Ensure oauth2-proxy-alertmanager Service does not exist:
     - apiVersion: v1
 
 {%- endif %}
+
+Create alertmanager-proxy Service:
+  metalk8s_kubernetes.object_present:
+    - manifest:
+        apiVersion: v1
+        kind: Service
+        metadata:
+          name: alertmanager-proxy
+          namespace: metalk8s-monitoring
+          labels:
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/part-of: metalk8s
+            heritage: metalk8s
+        spec:
+          type: ExternalName
+          {%- if alertmanager_oidc_enabled %}
+          externalName: oauth2-proxy-alertmanager.metalk8s-monitoring.svc.{{ coredns.cluster_domain }}
+          {%- else %}
+          externalName: prometheus-operator-alertmanager.metalk8s-monitoring.svc.{{ coredns.cluster_domain }}
+          {%- endif %}
+          ports:
+          - port: 9093
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -42,7 +42,7 @@ Create oauth2-proxy-alertmanager Deployment:
               labels:
                 app: oauth2-proxy-alertmanager
             spec:
-              serviceAccountName: oidc-proxy
+              serviceAccountName: oidc-proxy-alertmanager
               {%- if ca_configured %}
               initContainers:
               - name: k8s-sidecar
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -86,9 +86,9 @@ Create oauth2-proxy-alertmanager Deployment:
                 {%- if ca_configured %}
                 - --provider-ca-file=/tmp/secrets/{{ ca_file }}
                 {%- endif %}
-                - --http-address=0.0.0.0:4180
+                - --http-address=0.0.0.0:9093
                 ports:
-                - containerPort: 4180
+                - containerPort: 9093
                 volumeMounts:
                 - name: secrets-volume
                   mountPath: /tmp/secrets
@@ -111,7 +111,7 @@ Create oauth2-proxy-alertmanager Service:
           selector:
             app: oauth2-proxy-alertmanager
           ports:
-          - port: 4180
+          - port: 9093
 
 {%- else %}
 
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -14,7 +14,11 @@
 {%- set alertmanager_oidc_enabled = alertmanager.spec.get('config', {}).get('enable_oidc_authentication', False) %}
 {%- set alertmanager_oidc = alertmanager.spec.get('config', {}).get('oidc', {}) %}
 
-{%- set ingress_ca_file = 'namespace_metalk8s-ingress.secret_ingress-control-plane-default-certificate.tls.crt' %}
+{%- set alertmanager_oidc_ca = alertmanager_oidc.get('caSecret', {}) %}
+{%- set ca_namespace = alertmanager_oidc_ca.get('namespace', '') %}
+{%- set ca_name = alertmanager_oidc_ca.get('name', '') %}
+{%- set ca_configured = ca_namespace and ca_name %}
+{%- set ca_file = 'namespace_' ~ ca_namespace ~ '.secret_' ~ ca_name ~ '.tls.crt' %}
 
 {%- if alertmanager_oidc_enabled %}
 
@@ -39,6 +43,7 @@ Create oauth2-proxy-alertmanager Deployment:
                 app: oauth2-proxy-alertmanager
             spec:
               serviceAccountName: oidc-proxy
+              {%- if ca_configured %}
               initContainers:
               - name: k8s-sidecar
                 image: {{ build_image_name("k8s-sidecar") }}
@@ -46,11 +51,11 @@ Create oauth2-proxy-alertmanager Deployment:
                 restartPolicy: Always
                 env:
                 - name: LABEL
-                  value: metalk8s.scality.com/version
+                  value: metalk8s.scality.com/oidc-ca
                 - name: FOLDER
                   value: /tmp/secrets
                 - name: NAMESPACE
-                  value: metalk8s-ingress
+                  value: {{ ca_namespace }}
                 - name: RESOURCE
                   value: secret
                 - name: UNIQUE_FILENAMES
@@ -58,6 +63,7 @@ Create oauth2-proxy-alertmanager Deployment:
                 volumeMounts:
                 - name: secrets-volume
                   mountPath: /tmp/secrets
+              {%- endif %}
               containers:
               - name: oauth2-proxy
                 image: {{ build_image_name("oauth2-proxy") }}
@@ -74,7 +80,9 @@ Create oauth2-proxy-alertmanager Deployment:
                 {%- for group in alertmanager_oidc.get('authorizedGroups', []) %}
                 - --allowed-group={{ group }}
                 {%- endfor %}
-                - --provider-ca-file=/tmp/secrets/{{ ingress_ca_file }}
+                {%- if ca_configured %}
+                - --provider-ca-file=/tmp/secrets/{{ ca_file }}
+                {%- endif %}
                 - --http-address=0.0.0.0:4180
                 ports:
                 - containerPort: 4180
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -28,9 +28,6 @@ Create oauth2-proxy-alertmanager Deployment:
           namespace: metalk8s-monitoring
           labels:
             app: oauth2-proxy-alertmanager
-            app.kubernetes.io/managed-by: salt
-            app.kubernetes.io/part-of: metalk8s
-            heritage: metalk8s
         spec:
           replicas: 1
           selector:
@@ -99,9 +96,6 @@ Create oauth2-proxy-alertmanager Service:
           namespace: metalk8s-monitoring
           labels:
             app: oauth2-proxy-alertmanager
-            app.kubernetes.io/managed-by: salt
-            app.kubernetes.io/part-of: metalk8s
-            heritage: metalk8s
         spec:
           selector:
             app: oauth2-proxy-alertmanager
--- a/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
+++ b/salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
@@ -15,9 +15,9 @@
 {%- set alertmanager_oidc = alertmanager.spec.get('config', {}).get('oidc', {}) %}
 
 {%- set alertmanager_oidc_ca = alertmanager_oidc.get('caSecret', {}) %}
-{%- set ca_namespace = alertmanager_oidc_ca.get('namespace', '') %}
-{%- set ca_name = alertmanager_oidc_ca.get('name', '') %}
-{%- set ca_configured = ca_namespace and ca_name %}
+{%- set ca_namespace = alertmanager_oidc_ca.get('namespace', 'metalk8s-ingress') %}
+{%- set ca_name = alertmanager_oidc_ca.get('name', 'ingress-control-plane-default-certificate') %}
+
 {%- set ca_file = 'namespace_' ~ ca_namespace ~ '.secret_' ~ ca_name ~ '.tls.crt' %}
 
 {%- if alertmanager_oidc_enabled %}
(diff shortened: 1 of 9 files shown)


salt/metalk8s/addons/ui/deployed/ingress.sls
```diff
@@ -77,16 +103,28 @@ spec:
...
           service:
+            {%- if prometheus_oidc_enabled %}
+            name: oauth2-proxy
+            port:
+              number: 4180
+            {%- else %}
             name: thanos-api
             port:
               number: 10902
+            {%- endif %}
```
TeddyAndrieux: Do you really need to update Ingress, cannot we handle it directly in the service, I mean keep Ingress as today and change the service content depending if oidc is enabled or not ?

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+              app: oauth2-proxy-alertmanager
+          template:
+            metadata:
+              labels:
+                app: oauth2-proxy-alertmanager
+            spec:
+              serviceAccountName: oidc-proxy
+              initContainers:
+              - name: k8s-sidecar
+                image: {{ repo.registry_endpoint }}/{{ saltenv }}/k8s-sidecar:1.28.0
```
TeddyAndrieux: Use build_image_name instead here so that you do not have to set the version, it will automatically retrieve the one we embed in Metal


```suggestion
                image: {{ build_image_name("k8s-sidecar") }}
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+                - name: RESOURCE
+                  value: secret
+                - name: UNIQUE_FILENAMES
+                  value: "true"
+                volumeMounts:
+                - name: secrets-volume
+                  mountPath: /tmp/secrets
+              containers:
+              - name: oauth2-proxy
+                image: {{ repo.registry_endpoint }}/{{ saltenv }}/oauth2-proxy/oauth2-proxy:v7.6.0
```
TeddyAndrieux: ditto,, btw you have to embed the image in the product by puting it in `buildchain/buildchain/versions.py`

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: oauth2-proxy-alertmanager
+          namespace: metalk8s-monitoring
+          labels:
+            app: oauth2-proxy-alertmanager
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/part-of: metalk8s
+            heritage: metalk8s
```
TeddyAndrieux: No need to specify those, we already set some defaults in the salt module

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+                  value: /tmp/secrets
+                - name: NAMESPACE
+                  value: metalk8s-ingress
+                - name: RESOURCE
+                  value: secret
+                - name: UNIQUE_FILENAMES
+                  value: "true"
+                volumeMounts:
+                - name: secrets-volume
+                  mountPath: /tmp/secrets
```
TeddyAndrieux, line 65: So you give access to all secrets of the cluster with this label to this Pod to handle CA change ? 

It's not really good IMO

Because of current approach I agree that we do not have real better option to handle it, so ...maybe better to put a specific label on this ingress cert and only retrieve this one

It's linked to my bellow comment about CA file but maybe the best option right now is I suggest to make the CA file configurable in the alertmanager config like
```yaml
config:
  oidc:
    caSecret:
      namespace: <secret namespace>
      name: <secret name>
```

The pointed secret would need to have a label we choose like `metalk8s.scality.com/oidc-ca`

Then here we would do something like that

```yaml
              initContainers:
              - name: k8s-sidecar
                [...]
                env:
                - name: LABEL
                  value: metalk8s.scality.com/oidc-ca
                [...]
              containers:
              - name: oauth2-proxy
                [...]
                args:
                [...]
                - --provider-ca-file=/tmp/secrets/namespace_{{ alertmanager_oidc.caSecret.namespace }}.secret_{{ alertmanager_oidc.caSecret.name }}.tls.crt
```

It's just an idea but I feel it's a bit cleaner

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+                - --cookie-secret=MDEyMzQ1Njc4OWFiY2RlZmdoaWprbG1ub3BxcnN0dXY=
+                - --client-secret=unused-but-required
+                - --skip-jwt-bearer-tokens=true
+                - --email-domain=*
+                - --upstream=http://prometheus-operator-alertmanager.metalk8s-monitoring.svc:9093
+                - --oidc-groups-claim={{ alertmanager_oidc.get('groupsClaim', 'roles') }}
+                {%- for group in alertmanager_oidc.get('authorizedGroups', []) %}
+                - --allowed-group={{ group }}
+                {%- endfor %}
+                - --provider-ca-file=/tmp/secrets/{{ ingress_ca_file }}
```
TeddyAndrieux: It's a bit weird to hardcode the secret when the issuer is configurable, I know in our case the issuer will be exposed behind our ingress but since it's configurable it may not

But I don't see clean solution right now, maybe we should put in place cert-manager trust-manager to handle that kind of case

I think we can keep this approach for now

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,127 @@
...
+                - name: secrets-volume
+                  mountPath: /tmp/secrets
+              containers:
+              - name: oauth2-proxy
+                image: {{ repo.registry_endpoint }}/{{ saltenv }}/oauth2-proxy/oauth2-proxy:v7.6.0
+                args:
+                - --provider=oidc
+                - --oidc-issuer-url={{ alertmanager_oidc.get('issuer', '') }}
+                - --client-id={{ alertmanager_oidc.get('audience', '') }}
+                - --cookie-secret=MDEyMzQ1Njc4OWFiY2RlZmdoaWprbG1ub3BxcnN0dXY=
```
TeddyAndrieux, line 76: Shouldn't we generate a new one randomly on every install ?

ChengYanJin: `--cookie-secret` is a required parameter for oauth2-proxy, but in this setup it is never actually used.
Since `--skip-jwt-bearer-tokens=true` is enabled, authentication relies entirely on JWT Bearer tokens in the Authorization header, meaning no cookies are issued or validated.

As a result, I simply set a static hardcoded valid value. I also added some explanatory comments in 96ec16d.

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
ChengYanJin: Maybe we can have a discussion around this :)

salt/metalk8s/addons/ui/deployed/ingress.sls
ChengYanJin: 124492fe1

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
ChengYanJin: aa8504dae

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
ChengYanJin: aa8504dae
yes I was planing to do it in the coming PR ;)

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
ChengYanJin: 56d6aa318

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-prometheus.sls
```diff
@@ -0,0 +1,132 @@
...
+
+{%- if prometheus_oidc_enabled %}
+
+Create oauth2-proxy Deployment:
+  metalk8s_kubernetes.object_present:
+    - manifest:
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: oauth2-proxy
```
TeddyAndrieux: To be consistent maybe this one should be named `oauth2-proxy-prometheus`

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-rbac.sls
```diff
@@ -0,0 +1,105 @@
...
+
+{%- set prometheus_oidc_enabled = prometheus.spec.config.get('enable_oidc_authentication', False) %}
+{%- set alertmanager_oidc_enabled = alertmanager.spec.get('config', {}).get('enable_oidc_authentication', False) %}
+
+{%- set oidc_enabled = prometheus_oidc_enabled or alertmanager_oidc_enabled %}
+
+{%- set oidc_ca_namespace = prometheus.spec.config.get('oidc', {}).get('caSecret', {}).get('namespace', '') %}
+{%- if not oidc_ca_namespace %}
+  {%- set oidc_ca_namespace = alertmanager.spec.get('config', {}).get('oidc', {}).get('caSecret', {}).get('namespace', '') %}
+{%- endif %}
```
TeddyAndrieux: nit: But I think those 2 CAs might be different so I think you want to always create 2 Role and RoleBinding (even if they are the same it does not hurt to me)

salt/metalk8s/addons/ui/deployed/dependencies.sls
```diff
@@ -51,10 +77,18 @@ metadata:
...
+  {%- else %}
   externalName: thanos-query-http.metalk8s-monitoring.svc.{{ coredns.cluster_domain }}
+  {%- endif %}
   ports:
     - name: http
+      {%- if prometheus_oidc_enabled %}
+      port: 4180
+      {%- else %}
       port: 10902
+      {%- endif %}
```
TeddyAndrieux: Just a question, but do we have a good reason to use 4180 for oauth2-proxy instead of using the same port as the backend one ?

Just asking I have no strong opinion here both ok for me

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
TeddyAndrieux: Sure but let's keep this for later on I don't think we can have a quick clean solution so let's keep your current approach for now it's fine :+1:

salt/metalk8s/addons/ui/deployed/dependencies.sls
ChengYanJin: Good question. There's no strong technical reason — it's just that 4180 is oauth2-proxy's default port, so it was left as-is. But I will make the change to configure oauth2-proxy to listen on 10902 as well.

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,132 @@
...
+        'metalk8s-monitoring', 'metalk8s-alertmanager-config', alertmanager_defaults
+    )
+%}
+
+{%- set alertmanager_oidc_enabled = alertmanager.spec.get('config', {}).get('enable_oidc_authentication', False) %}
+{%- set alertmanager_oidc = alertmanager.spec.get('config', {}).get('oidc', {}) %}
+
+{%- set alertmanager_oidc_ca = alertmanager_oidc.get('caSecret', {}) %}
+{%- set ca_namespace = alertmanager_oidc_ca.get('namespace', '') %}
+{%- set ca_name = alertmanager_oidc_ca.get('name', '') %}
```
JBWatenbergScality:
```suggestion
{%- set ca_namespace = alertmanager_oidc_ca.get('namespace', 'metalk8s-ingress') %}
{%- set ca_name = alertmanager_oidc_ca.get('name', 'ingress-control-plane-default-certificate') %}
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,132 @@
...
+{%- set ca_configured = ca_namespace and ca_name %}
```
JBWatenbergScality:
```suggestion
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,132 @@
...
+          selector:
+            matchLabels:
+              app: oauth2-proxy-alertmanager
+          template:
+            metadata:
+              labels:
+                app: oauth2-proxy-alertmanager
+            spec:
+              serviceAccountName: oidc-proxy-alertmanager
+              {%- if ca_configured %}
```
JBWatenbergScality:
```suggestion
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,132 @@
...
+                - name: NAMESPACE
+                  value: {{ ca_namespace }}
+                - name: RESOURCE
+                  value: secret
+                - name: UNIQUE_FILENAMES
+                  value: "true"
+                volumeMounts:
+                - name: secrets-volume
+                  mountPath: /tmp/secrets
+              {%- endif %}
```
JBWatenbergScality:
```suggestion
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -0,0 +1,132 @@
...
+                - --skip-jwt-bearer-tokens=true
+                - --email-domain=*
+                - --upstream=http://prometheus-operator-alertmanager.metalk8s-monitoring.svc:9093
+                - --oidc-groups-claim={{ alertmanager_oidc.get('groupsClaim', 'roles') }}
+                {%- for group in alertmanager_oidc.get('authorizedGroups', []) %}
+                - --allowed-group={{ group }}
+                {%- endfor %}
+                {%- if ca_configured %}
+                - --provider-ca-file=/tmp/secrets/{{ ca_file }}
+                {%- endif %}
```
JBWatenbergScality:
```suggestion
                - --provider-ca-file=/tmp/secrets/{{ ca_file }}
```

salt/metalk8s/addons/ui/deployed/dependencies.sls
```diff
@@ -2,6 +2,32 @@
 
 {%- from "metalk8s/map.jinja" import coredns with context %}
 
+{%- set prometheus_defaults = salt.slsutil.renderer(
```
JBWatenbergScality: Maybe we can create a new svc in monitoring scope and change the underlying used svc there. This way we don't need to re-apply UI salt states when modifying prom/am config

JBWatenbergScality (approved):
In a next PR we will need label `metalk8s.scality.com/oidc-ca: "true"` to be set on secret as we introduce a selector on that in the sidecar

TeddyAndrieux (approved):
LGTM

PR #4802, opened 2026-03-02
scripts: Avoid silent skip of part of salt orchestrate execution

TeddyAndrieux:
This pull request refactors how commands are executed inside containers across all shell scripts, switching from using `crictl exec` to directly using `runc exec`. This change is intended to improve reliability, especially during upgrade or maintenance operations when containerd or kubelet may be restarted, which could cause `crictl` to fail silently. 

Additionally, the corresponding Python module and tests for `crictl exec` have been removed, and a bug fix is documented in the changelog.


anthony-treuillier-scality (approved):
LGTM

vdaviot (approved):

PR #4816, opened 2026-03-12
tools: add script to upgrade operator-sdk

ezekiel-alexrod:
### Summary

A generic, config-driven tool that automates operator-sdk upgrades by scaffolding
fresh projects and applying customizations via GNU patch files. Designed to be
reusable across repositories.

Closes: MK8S-190

### Execution flow

```mermaid
flowchart TD
    A["Load config.yaml"] --> B["Confirm upgrade (--yes to skip)"]
    B --> C["Install operator-sdk (cached)"]
    C --> D["Phase 1: Backup operator/ → operator.bak/"]
    D --> E["Phase 2: Scaffold fresh project"]
    E --> E1["operator-sdk init + create api"]
    E1 --> E2["Delete unwanted files (delete config)"]
    E2 --> F["Phase 2b: Detect latest versions"]
    F --> F1["operator-sdk: GitHub releases/latest"]
    F --> F2["Go toolchain: go.dev latest patch"]
    F --> F3["k8s.io libs: module proxy latest patch"]
    F1 & F2 & F3 --> G["Reconcile: compare detected vs pinned"]
    G --> H["Phase 3: Restore custom code (raw_copy)"]
    H --> I["Phase 4: Apply patch files"]
    I --> J["Substitute __GOTOOLCHAIN__ placeholder"]
    J --> K["Phase 5: go get k8s.io + go mod tidy"]
    K --> L["make manifests generate"]
    L --> M["make fmt vet"]
    M --> N["Extra commands (make metalk8s)"]
    N --> O["Upgrade complete"]
```

### Usage

```bash
python3 tools/upgrade-operator-sdk/upgrade.py \
    --operator-dir operator \
    tools/upgrade-operator-sdk/operator

python3 tools/upgrade-operator-sdk/upgrade.py \
    --operator-dir storage-operator \
    tools/upgrade-operator-sdk/storage-operator
```

| Option | Description |
|---|---|
| `--operator-dir` | Path to the operator project directory (required) |
| `--skip-backup` | Reuse an existing `.bak` directory |
| `--clean-tools` | Remove tool cache after upgrade |
| `--yes, -y` | Skip the confirmation prompt |

| Env variable | Description |
|---|---|
| `GITHUB_TOKEN` | Optional. Raises GitHub API rate limit (60 → 5000 req/hour) |

Prerequisites: `go`, `curl`, `patch`, `pyyaml`

### Configuration

Each operator has a config directory with `config.yaml` and `patches/`:

```
tools/upgrade-operator-sdk/
  upgrade.py
  operator/
    config.yaml
    patches/
      Dockerfile.patch          # extra COPY dirs, ldflags, LABEL block
      Makefile.patch            # GOTOOLCHAIN, metalk8s make target
      clusterconfig_types.patch # CRD type definition
      clusterconfig_controller.patch
      virtualippool_types.patch
      virtualippool_controller.patch
  storage-operator/
    config.yaml
    patches/
      Dockerfile.patch
      Makefile.patch
      volume_types.patch
      volume_controller.patch
```

#### config.yaml fields

| Field | Required | Description |
|---|---|---|
| `operator_sdk_version` | yes | Target operator-sdk release (e.g. `v1.42.1`) |
| `repo` | yes | Go module path for `operator-sdk init --repo` |
| `domain` | yes | CRD domain for `operator-sdk init --domain` |
| `apis` | yes | List of CRDs with `group`, `version`, `kind`, `namespaced` |
| `go_toolchain` | no | Pin Go toolchain (e.g. `go1.24.13`). If absent, detected from go.dev |
| `k8s_libs` | no | Pin k8s.io libs (e.g. `v0.33.10`). If absent, detected from module proxy |
| `raw_copy` | no | Dirs/files to copy from backup (purely custom, no scaffold equivalent) |
| `delete` | no | Dirs/files to remove from scaffold after generation |
| `extra_commands` | no | Commands to run after `make fmt vet` (e.g. `["make", "metalk8s"]`) |

### Version reconciliation (CI-friendly)

After scaffolding, the tool queries go.dev, the Go module proxy, and GitHub to detect the latest available versions. It compares them with the YAML pins:

| Situation | Behavior |
|---|---|
| No pin in YAML | Use detected version for this run (YAML not modified) |
| Pin matches detected | All good |
| Pin is older than detected | Warning with newer version available, keeps pinned value |
| Pin is newer than detected | Warning, uses detected value |

Zero interactive input during reconciliation -- safe for CI.

### Design decisions

- **Patch files over programmatic modifications**: all scaffold customizations are plain `diff -u` files, readable and editable without understanding the Python code
- **`raw_copy` for purely custom code**: only dirs/files with no scaffold equivalent are copied from backup; everything else uses patches
- **`delete` config field**: configurable removal of scaffold files (`.devcontainer`, `.github`, incompatible test stubs)
- **`--operator-dir` as CLI arg**: not in the YAML, so the tool works from any directory
- **No auto-pin**: the tool warns about newer versions but never modifies the config file
- **`__GOTOOLCHAIN__` placeholder**: only runtime placeholder, detected dynamically (latest Go patch from go.dev) and substituted in the Makefile
- **Jinja expressions hardcoded in patches**: each operator's `Makefile.patch` contains its own `build_image_name()` call directly, no placeholder needed
- **`--forward` flag on `patch`**: prevents false "reversed patch" detection on patches that remove most of a file's content


scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+    + _GITHUB_REPO_OPERATOR_SDK
+    + "/{version}/go.mod"
+)
+_URL_OPERATOR_SDK_DOWNLOAD: Final = (
+    "https://github.com/"
+    + _GITHUB_REPO_OPERATOR_SDK
+    + "/releases/download/{version}/operator-sdk_{goos}_{goarch}"
+)
+_URL_GO_RELEASES: Final = "https://go.dev/dl/?mode=json&include=all"
+_URL_GOLANGCI_INSTALL: Final = (
```
vdaviot: `_URL_GOLANGCI_LINT_INSTALL`

scripts/pyproject.toml
```diff
@@ -0,0 +1,39 @@
+# Linting and formatting configuration for scripts/upgrade-operator-sdk.py
+# Run from the scripts/ directory:
+#   python3 -m black upgrade-operator-sdk.py
+#   python3 -m ruff check upgrade-operator-sdk.py
+#   python3 -m mypy upgrade-operator-sdk.py
+
+[tool.black]
+line-length = 88
+target-version = ["py39"]
```
TeddyAndrieux: any reason to not use a non-EOL version ?

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+#   Root files:  .dockerignore  .gitignore  go.mod  go.sum  Makefile  PROJECT
+#                README.md
+#   Generated:   *zz_generated*  internal/controller/suite_test.go
+# ---------------------------------------------------------------------------
+
+_SCAFFOLD_ONLY_DIRS: Final[frozenset[str]] = frozenset(
+    {
+        ".devcontainer",
+        ".github",
+        "bin",
```
TeddyAndrieux: This directory is in gitignore so we don't care about this one

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+        "cmd",
+        "config",
```
TeddyAndrieux: We have a metalk8s directory in this one that we want to keep

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+    {
+        ".dockerignore",
+        ".gitignore",
+        "go.mod",
+        "go.sum",
+        "Makefile",
+        "PROJECT",
+        "README.md",
+        # test coverage artefact — regenerated by `make test`, not custom code
+        "cover.out",
```
TeddyAndrieux: Those are not in git as well so to me we don't care

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+# scaffold addition that needs classifying in the sets above.
+_KNOWN_SCAFFOLD_ROOTS: Final[frozenset[str]] = frozenset(
+    _SCAFFOLD_ONLY_DIRS
+    | {f.split("/")[0] for f in _SCAFFOLD_ONLY_FILES}
+    | {
+        "api",
+        "hack",
+        "internal",  # created by `create api`
+        # root files scaffold creates that we intentionally overwrite from backup
+        ".golangci.yml",
```
TeddyAndrieux: You sure ? To me we could keep the one generated by operator-sdk here no ?

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1063 @@
...
+        "Dockerfile",
```
TeddyAndrieux: Here ditto to me we want to keep the one from operator sdk and just apply our change on top of it

I think we should work with patch file for those and in case the patch has conflict user has to solve it by hand

Basically to me the tool should work as follow

- Do operator-sdk scalfolding
- For a list of known file and directory we ensure it does not exists in the target and copy it from the backup (e.g.: `pkg`, `version`, `config/metalk8s` folder here)
- Apply patch for every files that we update
  - `Makefile`
  - `Dockerfile`
  - internal controller go files
  - ...
- A `go mod tidy`
- Re-render using `make generate` `make manifests` (and `make metalk8s` most likely)

And I think we are good and it make the tool really more generic

Here with the current implementation since we replace some files in place without checking if the content has changed we will for sure miss some future changes
I think about `main.go` and `Dockerfile` for example

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: Already handled: `_should_merge()` has an explicit exception for `config/metalk8s/` — it is always restored from backup even though `config/ `as a whole is scaffold-only. See the first check in the function:

```python
def _should_merge(rel: str) -> bool:
    if rel.startswith("config/metalk8s/"):
        return True
```
I've also clarified this in the merge policy comment block at the top of the file, which now reads `config (except config/metalk8s/`).

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: Thanks for the detailed feedback — I've reworked the approach based on your comments. Here's how it works now:
**The script now keeps scaffold versions for `Dockerfile` and `.golangci.yml`** (as well as `Makefile`, `go.mod`, `cmd/`, `config/`, etc.) and applies our customizations programmatically on top, rather than restoring from backup.
The flow is essentially what you described:
1. `operator-sdk init` + `create api` (fresh scaffold)
2. Copy purely custom directories from backup (`pkg/`, `version/`, `config/metalk8s/`, `api/`, `hack/`, `internal/controller/*.go`)
3. Apply programmatic adaptations on scaffold files:
   - **Makefile**: `GOTOOLCHAIN` block + golangci-lint version + `metalk8s` target (same pattern as before)
   - **Dockerfile**: align `FROM golang:X.Y`, insert extra `COPY` lines for our custom dirs, add `ldflags` for version injection, append the Scality `LABEL` block — all driven by a `DockerfilePatch` dataclass per operator
   - **Source files**: idempotent `SourceFix` regex patches (e.g. `ioutil` → `os`)
4. `.golangci.yml`: scaffold version used as-is (the v1→v2 format change makes migrating our old config fragile; future v2→v2+ upgrades get the new scaffold config automatically)
5. `go mod tidy` → `make generate manifests` → `make build` → `make metalk8s`

I chose programmatic adaptations over patch files because:
- Our Dockerfile changes are **additive** (extra `COPY`, `LABEL` block) not structural — regex insertions are deterministic and won't conflict
- Patch files would need to be regenerated/maintained for each SDK version and would silently fail on structural scaffold changes
- The `diff -r operator.bak/ operator/` recommended at the end still lets the user review everything
This way, any upstream Dockerfile improvement (new base image, security fix, build pattern change) is picked up automatically, and our customizations are applied cleanly on top.

TeddyAndrieux: I really feel it will be better to have a set of classic GNU patch file next to the script that would be applied with simple `patch` commands

For future it will be way easier to generate new patch file and to use the tool without reading the full code and understand everything


> I chose programmatic adaptations over patch files because:
> 
>     * Our Dockerfile changes are **additive** (extra `COPY`, `LABEL` block) not structural — regex insertions are deterministic and won't conflict

We also change the command to build so not really

>     * Patch files would need to be regenerated/maintained for each SDK version and would silently fail on structural scaffold changes

Why ? I mean if there is no changes the patchfile will apply without any changes so it's fine, and if there is changes somewhere not touched by the patchfile it will also properly apply without issue

>     * The `diff -r operator.bak/ operator/` recommended at the end still lets the user review everything

I'm not sure it's really needed it will basically be the same output has a `git diff`

>       This way, any upstream Dockerfile improvement (new base image, security fix, build pattern change) is picked up automatically, and our customizations are applied cleanly on top.

To me it does the same with patchfile and it's easier to use you do not have to know the tool by itself

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+        "go.mod",
+        "go.sum",
+        "Makefile",
+        "PROJECT",
+        "README.md",
+        # scaffold-generated test setup (version-specific, not custom code)
+        "internal/controller/suite_test.go",
+    }
+)
```
TeddyAndrieux: Why do we still needs those ? We always keep everything from operator-sdk scalfolding and only apply "our stuff" on top of it

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    parser = argparse.ArgumentParser(
+        description="Upgrade operator-sdk projects by scaffolding fresh and "
+        "merging custom code from backup.",
+    )
+    parser.add_argument(
+        "--operator-only", action="store_true", help="Only process operator/"
+    )
+    parser.add_argument(
+        "--storage-only", action="store_true", help="Only process storage-operator/"
+    )
```
TeddyAndrieux: I would prefer we do not have anything "specific" to those and instead just specify
- an "operator directory"
- a "patch directory"
- a file (I think a yaml could be good) containing the information you have in `OPERATORS`

And so in our case we will run the tools twice one for operator and one for storage-operator

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    if args.operator_only:
+        targets = ["operator"]
+    elif args.storage_only:
+        targets = ["storage-operator"]
+    else:
+        targets = ["operator", "storage-operator"]
+
+    _check_prerequisites()
+
+    global versions
```
TeddyAndrieux: 😢

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+            # Go 1.24+: go vet rejects non-constant format strings in fmt.Errorf.
+            # Remove once the backup no longer contains this pattern
+            # (i.e., after this script has run at least once from v1.37.0).
+            SourceFix(
+                path="pkg/controller/clusterconfig/controlplane/ingress.go",
+                old=r"fmt\.Errorf\(([a-zA-Z_]\w*)\)",
+                new=r'fmt.Errorf("%s", \1)',
+                regex=True,
+            ),
+        ),
```
TeddyAndrieux: Why do we need it ? I mean those kind of step that are needed to bump go should be done by hand by the people doing the upgrade of operator sdk no need to be included in the tooling

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+                path="internal/controller/volume_controller.go",
+                old='"io/ioutil"',
+                new='"os"',
+            ),
+            SourceFix(
+                path="internal/controller/volume_controller.go",
+                old="ioutil.ReadFile",
+                new="os.ReadFile",
+            ),
+        ),
```
TeddyAndrieux: ditto

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+        raise RuntimeError("versions not initialised — call detect_versions() first")
+    return {
+        "PATH": f"{TOOLS_BIN}:{os.environ.get('PATH', '')}",
+        "GOTOOLCHAIN": versions.go_toolchain,
+    }
+
+
+def run(
+    cmd: list[str],
+    *,
```
TeddyAndrieux: TIL, didn't know that kind of hack

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+
+
+def _detect_golangci_lint_version() -> str:
+    log_info("Querying GitHub for latest golangci-lint release...")
+    ver = _fetch_latest_github_release(_GITHUB_REPO_GOLANGCI_LINT)
+    log_info(f"  golangci-lint:      {ver}")
+    return ver
+
+
+def detect_versions() -> VersionInfo:
```
TeddyAndrieux: I think we should provide a way to enforce the versions (I think from environment variable could be good)

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    k8s_version = _latest_k8s_patch(base_k8s)
+
+    return cr_version, k8s_version
+
+
+def _detect_golangci_lint_version() -> str:
+    log_info("Querying GitHub for latest golangci-lint release...")
+    ver = _fetch_latest_github_release(_GITHUB_REPO_GOLANGCI_LINT)
+    log_info(f"  golangci-lint:      {ver}")
+    return ver
```
TeddyAndrieux: Not sure it's needed I think if we keep the one from operator-sdk it's good enough

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    cr_gomod = _fetch_text(_URL_CONTROLLER_RUNTIME_GOMOD.format(version=cr_version))
+    m_k8s = re.search(_PAT_K8S_API_IN_GOMOD, cr_gomod)
+    if not m_k8s:
+        die("Failed to parse k8s.io/api version from controller-runtime go.mod")
+    base_k8s = m_k8s.group(1)
+
+    # Bump to the latest patch of that major.minor.
+    k8s_version = _latest_k8s_patch(base_k8s)
+
+    return cr_version, k8s_version
```
TeddyAndrieux: Why do we need those ? To me if we keep the one from operator-sdk it's good enough

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    )
+
+    for api in spec.apis:
+        _create_api(op_dir, sdk, api)
+
+    # Remove scaffold additions we don't want in the project.
+    devcontainer = op_dir / ".devcontainer"
+    if devcontainer.exists():
+        shutil.rmtree(devcontainer)
+        log_info("Removed .devcontainer/ (not needed)")
```
TeddyAndrieux: I suggest to have a list for this in the `OPERATORS` (/ yaml input cf my previous comment)

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+
+    # Warn about any scaffold-generated root entries not yet in our policy.
+    _check_scaffold_completeness(op_dir)
```
TeddyAndrieux: I'm not sure it's needed it could be checked with a simple `git status` not sure we want to maintain such list

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    log_info("Scaffold complete")
+
+
+def _create_api(op_dir: Path, sdk: str, api: ApiDef) -> None:
+    # Build the common trailing arguments once to avoid duplication.
+    tail = ["--version", api.version, "--kind", api.kind, "--resource", "--controller"]
+
+    result = run(
+        [sdk, "create", "api", "--group", api.group, *tail], cwd=op_dir, check=False
+    )
```
TeddyAndrieux: You are missing handling of scope here (namespaced or not)

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+
+    if api.group:
+        die(f"Failed to create API {api.kind}")
+
+    # operator-sdk may reject an empty group; retry with a placeholder and then
+    # scrub it from PROJECT so the CRD group stays empty.
+    log_warn(f"Empty group rejected for {api.kind}, retrying with placeholder")
+    run([sdk, "create", "api", "--group", "metalk8s", *tail], cwd=op_dir)
+    file_regex_replace(op_dir / "PROJECT", _PAT_PROJECT_GROUP_LINE, "")
+    log_info("Patched PROJECT: removed placeholder group")
```
TeddyAndrieux: What, I don't get why we would need this logic ?

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+        if not src.is_file():
+            continue
+        rel = src.relative_to(bak).as_posix()
+        if not _should_merge(rel):
+            continue
+        dst = op_dir / rel
+        dst.parent.mkdir(parents=True, exist_ok=True)
+        # shutil.copy (not copy2): gives merged files current timestamps so
+        # that make does not mistake them for stale relative to generated artefacts.
+        shutil.copy(src, dst)
```
TeddyAndrieux: To me if the file already exists in the target we want to print an error and maybe show a diff between the two

And in this situation to me the process should be to update the file in the bak to match what we want at the end and then re-run this tooling

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+# ===================================================================
+
+
+def merge_backup(spec: OperatorSpec) -> None:
+    log_step(f"Phase 3: Merging custom code for {spec.name}")
+    op_dir = spec.op_dir
+    bak = spec.bak_dir
+
+    merged: list[str] = []
+    for src in sorted(bak.rglob("*")):
```
TeddyAndrieux: I would prefer to have a list of file and/or directory in the OPERATORS (/yaml input file) to list all files that should be retrieved

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+        text = text.replace("__IMAGE__", jinja_image)
+        text = re.sub(
+            _PAT_MAKEFILE_GOLANGCI_VERSION,
+            f"GOLANGCI_LINT_VERSION ?= {versions.golangci_lint}",
+            text,
+            flags=re.MULTILINE,
+        )
+        makefile.write_text(text, encoding=_ENCODING)
+
+    log_info("Version substitutions applied")
```
TeddyAndrieux: IMHO it shouldn't be needed, either we take versions from operator-sdk either we just have to update our patch files to match the changes we need to do

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    )
+    if updated != text:
+        log_info(f"Applied fix: {Path(fix.path).name}")
+        path.write_text(updated, encoding=_ENCODING)
+
+
+def _apply_source_fixes(spec: OperatorSpec) -> None:
+    """Apply the operator-specific source fixes declared in spec.fixes."""
+    for fix in spec.fixes:
+        _apply_source_fix(spec.op_dir, fix)
```
TeddyAndrieux: ditto shouldn't be needed to me

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    where the inner struct registered with the manager does not expose Reconcile(),
+    so these stubs must be removed.  This applies to every operator-sdk upgrade.
+    """
+    ctrl_dir = op_dir / "internal" / "controller"
+    if not ctrl_dir.exists():
+        return
+    for test_file in ctrl_dir.glob("*_controller_test.go"):
+        if ".Reconcile(" in test_file.read_text(encoding=_ENCODING):
+            log_info(f"Removing incompatible scaffold test: {test_file.name}")
+            test_file.unlink()
```
TeddyAndrieux: Should be handled with patch file to me

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+    op_dir = spec.op_dir
+    bin_dir = op_dir / "bin"
+    if bin_dir.exists():
+        shutil.rmtree(bin_dir)
+
+    # Explicitly pin k8s.io libs to the latest compatible patch before tidy,
+    # so `go mod tidy` does not silently keep an older patch from the scaffold.
+    k8s_get_args = [f"{lib}@{versions.k8s_libs}" for lib in _K8S_LIBS]
+    log_info(f"Bumping k8s.io libs to {versions.k8s_libs}...")
+    run(["go", "get", *k8s_get_args], cwd=op_dir)
```
TeddyAndrieux: I would suggest to not do it (at least not in this tooling if we want to use another version than the one from operator-sdk it should be handled with a patch to me

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+
+    steps = [
+        ("go mod tidy...", ["go", "mod", "tidy"]),
+        ("make manifests generate...", ["make", "manifests", "generate"]),
+        ("make fmt vet...", ["make", "fmt", "vet"]),
+        ("make build...", ["make", "build"]),
```
TeddyAndrieux: Not sure the make build is really relevent here

scripts/upgrade-operator-sdk.py
```diff
@@ -0,0 +1,1081 @@
...
+        ("make metalk8s...", ["make", "metalk8s"]),
```
TeddyAndrieux: (I feel this tool could be used in some other repos so I suggest to add this specific command in the `spec` like `extraGenerateCommands`

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: I mean, why wouldn't we want the latest versions of operator-sdk and golang?

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: Ok, so the original ticket I got from bumping the k8s.io go libs will be manual then? Even when the latest patch version is retrocompatible?

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: ditto about the retrieving the latest version

scripts/upgrade-operator-sdk.py
TeddyAndrieux: To me we would want to run this tool in the CI to ensure there is not someone that did some modification in the code and not properly reported it in the patch files (so that we do not mess in a future update or operator-sdk)

Basically like the other codegen for charts & co

scripts/upgrade-operator-sdk.py
TeddyAndrieux: If we want to manage the k8s go lib it has to match the kube version we embed and not always take the last one

But I'm not sure it's needed for now we would stay with operator-sdk one for now, but right being able to pin the k8s go lib we want might be valuable

scripts/upgrade-operator-sdk.py
TeddyAndrieux: To me for golang we should stay with the one from operator-sdk, I do not see why we would want a more recent one instead of waiting for a new operator-sdk release

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: yes of course, the latest patch of the minor...
FYI, if I do what you think, the k8s.io go libs will be in v0.33.1, whereas v0.33.9 is out and compatible.

scripts/upgrade-operator-sdk.py
ezekiel-alexrod: CVEs? minor performance enhancements?

Also, I talked about the latest patch of the minor compatible
For example: go 1.25.0 -> go 1.25.8

scripts/upgrade-operator-sdk/operator/config.yaml
```diff
@@ -0,0 +1,28 @@
+repo: github.com/scality/metalk8s/operator
+domain: metalk8s.scality.com
+operator_dir: operator
```
TeddyAndrieux: Nit: I think it will be simpler to use if this one is a CLI arg so that you do not need to be in a specific directory to be able to run the tool

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+    dockerfile = op_dir / "Dockerfile"
+    if dockerfile.exists():
+        text = dockerfile.read_text(encoding=_ENCODING)
+        text = re.sub(
+            _PAT_DOCKERFILE_FROM_GOLANG,
+            f"FROM golang:{cfg['go_major_minor']}",
+            text,
+        )
+        dockerfile.write_text(text, encoding=_ENCODING)
```
TeddyAndrieux: As discussed to me this one is not needed

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+try:
+    import yaml
+except ImportError:
+    print("pyyaml is required: pip install pyyaml", file=sys.stderr)
+    sys.exit(1)
+
+# ---------------------------------------------------------------------------
+# Paths
+# ---------------------------------------------------------------------------
+REPO_ROOT: Final = Path(__file__).resolve().parent.parent.parent
```
TeddyAndrieux: Ok I see but I would prefer paths are inputs and not deducted like this

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+def load_config(config_dir: str) -> dict[str, Any]:
+    """Load and validate the operator config from a directory.
+
+    The directory must contain ``config.yaml`` and may contain a
+    ``patches/`` subdirectory with GNU unified diff files.
+
+    If *config_dir* is a plain name (no path separators), it is resolved
+    relative to the script's own directory — so ``operator`` finds
+    ``scripts/upgrade-operator-sdk/operator/``.  Otherwise it is used
```
TeddyAndrieux: That's weird behavior no ? It could be relative and in the current directory you would not have any `/` then

I think we should apply KISS principle especially for tools like that that will most likely not be used everyday so the simpler they are the best

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+    with config_file.open(encoding=_ENCODING) as f:
+        cfg: dict[str, Any] = yaml.safe_load(f)
+
+    required = (
+        "repo",
+        "domain",
+        "operator_dir",
+        "apis",
+        "operator_sdk_version",
+        "go_toolchain",
```
TeddyAndrieux: Nit: IMHO go_toolchain should be optional, in some other repo we may not want to set those

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+    cfg.setdefault("backup_paths", [])
+    cfg.setdefault("extra_commands", [])
+    cfg.setdefault("k8s_libs", "")
+    cfg["operator_dir"] = REPO_ROOT / cfg["operator_dir"]
+    cfg["patches_dir"] = d / "patches"
+    cfg["backup_dir"] = REPO_ROOT / f"{cfg['operator_dir'].name}.bak"
+
+    tc = cfg["go_toolchain"]
+    m = re.match(r"go(\d+\.\d+)", tc)
+    cfg["go_major_minor"] = m.group(1) if m else tc.lstrip("go")
```
TeddyAndrieux: So you update the dockerfile based on go toolchain, it's a bit weird to me (anyway to me we shouldn't update the base image of the dockerfile here

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+    count = 0
+    conflicts: list[str] = []
+    for rel_path in cfg["backup_paths"]:
+        src = bak / rel_path
+        dst = op_dir / rel_path
+
+        if not src.exists():
+            log_warn(f"  {rel_path} not found in backup, skipping")
+            continue
```
TeddyAndrieux: I think we should fail in this case no ?

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+        is_dir = rel_path.endswith("/")
```
TeddyAndrieux: Hmm I think you should instead test the file ("compute" the actual path and us `.is_dir()` directly)

tools/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+        if not src.exists():
+            log_warn(f"  {rel_path} not found in backup, skipping")
+            continue
+
+        is_dir = rel_path.endswith("/")
+
+        if is_dir:
+            if dst.exists():
+                shutil.rmtree(dst)
```
TeddyAndrieux: I'm not agree with this, it means we will potentially silently ignore new files in a directory we "manage" I prefer we error in this case

scripts/upgrade-operator-sdk/operator/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+    namespaced: false
+  - version: v1alpha1
+    kind: VirtualIPPool
+    namespaced: true
+
+backup_paths:
+  - pkg/
+  - version/
+  - config/metalk8s/
+  - api/
```
TeddyAndrieux: I would prefer we have a patch for the `_types.go` files so that we support potential updates on those by operator-sdk

scripts/upgrade-operator-sdk/operator/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - hack/
```
TeddyAndrieux: Why ? Those are pure operator-sdk content to me

scripts/upgrade-operator-sdk/operator/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - internal/controller/
```
TeddyAndrieux: ditto here to me we should rely on patch files

scripts/upgrade-operator-sdk/operator/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+apis:
+  - version: v1alpha1
+    kind: ClusterConfig
+    namespaced: false
+  - version: v1alpha1
+    kind: VirtualIPPool
+    namespaced: true
+
+backup_paths:
```
TeddyAndrieux: Nit: Not sure it's super clear what this field do without the script context, maybe something like `raw_copy` or something like that

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+
+        if is_dir:
+            if dst.exists():
+                shutil.rmtree(dst)
+            shutil.copytree(
+                src,
+                dst,
+                ignore=shutil.ignore_patterns("*zz_generated*"),
```
TeddyAndrieux: It shouldn't be needed to me

scripts/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+            if dst.exists():
+                log_error(f"  {rel_path} exists in both scaffold and backup")
+                result = subprocess.run(
+                    ["diff", "-u", str(dst), str(src)],
+                    capture_output=True,
+                    text=True,
+                )
+                if result.stdout:
+                    print(result.stdout)
+                conflicts.append(rel_path)
```
TeddyAndrieux: 👍 I think we want the same for directories

tools/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,588 @@
...
+            f"FROM golang:{cfg['go_major_minor']}",
+            text,
+        )
+        dockerfile.write_text(text, encoding=_ENCODING)
+
+    makefile = op_dir / "Makefile"
+    if makefile.exists():
+        text = makefile.read_text(encoding=_ENCODING)
+        text = text.replace("__GOTOOLCHAIN__", cfg["go_toolchain"])
+        text = text.replace("__IMAGE__", cfg.get("image_placeholder", ""))
```
TeddyAndrieux: Why do we need this ? It could be part of the patch

scripts/upgrade-operator-sdk.py
TeddyAndrieux: As discussed I'm fine with keeping k8s go lib

scripts/upgrade-operator-sdk.py
TeddyAndrieux: We have no issue about CVE for go version since we build in the docker image that already take the latest patch

TeddyAndrieux (commented):
I would suggest also to move this to `tools` it has nothing to do with `scripts` (this directory is for scripts that will be used by the end user not by the developers)

tools/upgrade-operator-sdk/upgrade.py
ezekiel-alexrod: `__IMAGE__` is already operator-specific (each operator has its own Makefile.patch) so we could hardcode it. However, `__GOTOOLCHAIN__` is detected dynamically at runtime (latest Go patch from go.dev based on the scaffold's go.mod) and pinned in `config.yaml` -- hardcoding it in the patch would require editing the patch file on every upgrade instead of just updating the YAML pin.

Happy to hardcode `__IMAGE__` in the patches if you prefer (it would remove `_substitute_placeholders()` for that part), but I'd keep `__GOTOOLCHAIN__` as a placeholder since its value changes with each operator-sdk release.

tools/upgrade-operator-sdk/operator/patches/clusterconfig_controller.patch
```diff
@@ -0,0 +1,73 @@
+--- a/internal/controller/clusterconfig_controller.go
++++ b/internal/controller/clusterconfig_controller.go
+@@ -1,5 +1,5 @@
+ /*
+-Copyright 2026.
++Copyright 2022.
```
TeddyAndrieux: I think we can remove it from the patch

tools/upgrade-operator-sdk/operator/patches/clusterconfig_controller.patch
```diff
@@ -0,0 +1,73 @@
...
+@@ -33,9 +29,13 @@
+ 	Scheme *runtime.Scheme
+ }
+ 
+-// +kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs,verbs=get;list;watch;create;update;patch;delete
+-// +kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs/status,verbs=get;update;patch
+-// +kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs/finalizers,verbs=update
++//+kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs,verbs=get;list;watch;create;update;patch;delete
++//+kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs/status,verbs=get;update;patch
++//+kubebuilder:rbac:groups=metalk8s.scality.com,resources=clusterconfigs/finalizers,verbs=update
```
TeddyAndrieux: I remember those comments was needed by golangcilint at some point I feel it should no longer needed I would prefer to have patch as simple as possible

tools/upgrade-operator-sdk/operator/patches/clusterconfig_controller_test.patch
TeddyAndrieux, line 1: I think we may want an entry in the config to delete some files like this one and devcontainer and github for this repo

tools/upgrade-operator-sdk/upgrade.py
TeddyAndrieux: Yes agree to remove image one and just keep the toolchain one

tools/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,790 @@
...
+        pinned = cfg.get(key, "")
+
+        if not pinned:
+            log_info(f"  {key}: {found} (detected, auto-pinning)")
+            cfg[key] = found
+            auto_pins[key] = found
+        elif found == pinned:
+            log_info(f"  {key}: {pinned} (up to date)")
+        elif found > pinned:  # lexicographic, works for semver
+            log_warn(f"  {key}: pinned {pinned}, " f"newer {found} available")
```
TeddyAndrieux:
```suggestion
            log_warn(f"  {key}: pinned {pinned}, newer {found} available")
```

tools/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,790 @@
...
+            cfg[key] = pinned
+        else:
+            log_warn(f"  {key}: pinned {pinned} > detected {found}, " "using detected")
```
TeddyAndrieux:
```suggestion
            log_warn(f"  {key}: pinned {pinned} > detected {found}, using detected")
```

tools/upgrade-operator-sdk/upgrade.py
```diff
@@ -0,0 +1,790 @@
...
+            cfg[key] = found
+
+    if auto_pins:
+        _update_yaml(cfg["config_file"], auto_pins)
```
TeddyAndrieux: Do we really want to auto update the yaml ?

tools/upgrade-operator-sdk/upgrade.py
TeddyAndrieux: It does not seems to be resolved

Ideally I would say we do the `diff` and if they are identical we just skip and continue if not we raise like for file (same diff logic might be done for file as well) -> this way if you re-run and some part has already been copied it will still work

tools/upgrade-operator-sdk/upgrade.py
ezekiel-alexrod: why not? it pins the versions in the yaml for the future, just in the case where the pins don't exist in the yaml
If it does, it warns the user and don't update the yaml

TeddyAndrieux: To me we might have case where we don't want to pin if we don't care about the version, (not for metal but thinking about others projects)

TeddyAndrieux (approved):

vdaviot (approved):

scripts/upgrade-operator-sdk.py
vdaviot: I cannot cancel this comment, as the file does not exist anymore

PR #4822, opened 2026-03-17
MK8S-184 - Add restart script to react when control plane ingress change

ChengYanJin:
**Component**:  Salt

**Context**: 
 This PR adds an automatic restart mechanism for the oauth2-proxy deployments when the CA certificate secret changes.

When OIDC authentication is enabled, the oauth2-proxy pods use a k8s-sidecar to sync the CA certificate from a Kubernetes secret. However, when the CA secret is updated (e.g., certificate rotation), the oauth2-proxy process keeps using the old CA loaded in memory, causing authentication failures.

**Summary**:
  Leverage the k8s-sidecar's SCRIPT feature to run a Python script whenever the CA secret changes. The script:
  1. Computes a SHA256 hash of the CA files on disk
  2. Compares it with the previously stored hash
  3. If changed, triggers a rolling restart of the deployment via the Kubernetes API (same as kubectl rollout restart)
  4. On initial startup, saves the hash and skips restart to avoid a restart loop

**Acceptance criteria**: 

When the CA secret data changes, the oauth2-proxy pods (prometheus and alertmanager) are automatically restarted via a rolling restart.

---

[Closes: #[MK8S-184](https://scality.atlassian.net/browse/MK8S-184)]

[MK8S-184]: https://scality.atlassian.net/browse/MK8S-184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


salt/metalk8s/addons/nginx-ingress-control-plane/deployed/tls-secret.sls
```diff
@@ -7,6 +7,8 @@ kind: Secret
 metadata:
   name: ingress-control-plane-default-certificate
   namespace: metalk8s-ingress
+  labels:
+    metalk8s.scality.com/oidc-ca: "true"
```
MonPote, line 11:
```suggestion
    metalk8s.scality.com/cp-oidc-ca: "true"
```

`cp-oidc-ca` might be a more explicit label to differentiate from the workload plane.

salt/tests/unit/formulas/fixtures/salt.py
```diff
@@ -412,6 +412,17 @@ def slsutil_renderer(salt_mock: SaltMock, source: str, **_kwargs: Any) -> Any:
 register_basic("file.find")(MagicMock(return_value=[]))
 register_basic("file.join")(lambda *args: "/".join(args))
 register_basic("file.read")(MagicMock(return_value="<file contents>"))
+
+
+@register_basic("cp.get_file_str")
+def cp_get_file_str(source: str, **_kwargs: Any) -> str:
```
MonPote: Do you think something like 
`@register_basic("cp.get_file_str")(MagicMock(return_value="toto"))` is enough ?

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-restart-script.sls
```diff
@@ -0,0 +1,61 @@
...
+          namespace: metalk8s-monitoring
+          labels:
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/part-of: metalk8s
+            heritage: metalk8s
+        data:
+          restart-on-ca-change.py: |-
+{{ script_content | indent(12, first=True) }}
+          restart-on-ca-change.sh: |
+            python3 /scripts/restart-on-ca-change.py
```
TeddyAndrieux: Do we really need this ?

To me it should work with your current shebang and maybe put mode when mounting it
```yaml
configMap:
  name: oidc-proxy-restart-script
  defaultMode: 0555
```

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-restart-script.sls
```diff
@@ -0,0 +1,61 @@
...
+        metadata:
+          name: oidc-proxy-restart-script
+          namespace: metalk8s-monitoring
+          labels:
+            app.kubernetes.io/managed-by: salt
+            app.kubernetes.io/part-of: metalk8s
+            heritage: metalk8s
+        data:
+          restart-on-ca-change.py: |-
+{{ script_content | indent(12, first=True) }}
```
TeddyAndrieux: Nit: For readability


```suggestion
            {{ script_content | indent(12) }}
```

salt/tests/unit/formulas/fixtures/salt.py
```diff
@@ -412,6 +412,9 @@ def slsutil_renderer(salt_mock: SaltMock, source: str, **_kwargs: Any) -> Any:
 register_basic("file.find")(MagicMock(return_value=[]))
 register_basic("file.join")(lambda *args: "/".join(args))
 register_basic("file.read")(MagicMock(return_value="<file contents>"))
+
+
```
TeddyAndrieux: Any reason to add those 2 ? :)

salt/metalk8s/addons/prometheus-operator/deployed/files/restart-on-ca-change.py
```diff
@@ -0,0 +1,100 @@
...
+        sys.exit(1)
+
+    # Persist hash only after successful restart
+    with open(HASH_FILE, "w") as f:
+        f.write(current_hash)
+    print(f"Rolling restart triggered for {deployment}")
+
+
+if __name__ == "__main__":
+    main()
```
TeddyAndrieux, line 87: I think it could be a bit more "recent python style" and a bit of small rework

Btw we know the ca file name so why do we need to iter over all files ? We could simply check the right file that could be gave has input to the script like the namespace and deployment name CA_DIR should also be an input to the script to me

So I suggest something like that (to be tested)

```python
#!/usr/bin/env python3
import hashlib
import json
import os
import sys
from datetime import datetime, timezone
from pathlib import Path

import requests

SA_TOKEN = Path("/var/run/secrets/kubernetes.io/serviceaccount/token")
SA_CA = Path("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")

K8S_API = "https://kubernetes.default.svc"
HASH_FILE_NAME = ".ca-hash-previous"


def hash_file(file_path: Path) -> str:
    h = hashlib.sha256()
    h.update(file_path.read_bytes())
    return h.hexdigest()


def trigger_restart(namespace: str, deployment: str) -> None:
    token = SA_TOKEN.read_text()
    timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
    body = {
        "spec": {
            "template": {
                "metadata": {
                    "annotations": {
                        "kubectl.kubernetes.io/restartedAt": timestamp,
                    }
                }
            }
        }
    }
    url = f"{K8S_API}/apis/apps/v1/namespaces/{namespace}/deployments/{deployment}"
    response = requests.patch(
        url,
        json=body,
        headers={
            "Authorization": f"Bearer {token}",
            "Content-Type": "application/strategic-merge-patch+json",
        },
        verify=str(SA_CA),
    )
    response.raise_for_status()


def main():
    ca_dir = Path(os.environ["CA_DIR"])
    ca_file = ca_dir / os.environ["CA_FILE_NAME"]

    hash_file = ca_dir / HASH_FILE_NAME

    if not ca_file.exists():
        print(f"CA file {ca_file} does not exist, skipping")
        return

    current_hash = hash_file(ca_file)

    if not hash_file.exists():
        hash_file.write_text(current_hash)
        print("Initial CA load, skipping restart")
        return

    previous_hash = hash_file.read_text().strip()

    if current_hash == previous_hash:
        return

    namespace = os.environ["POD_NAMESPACE"]
    deployment = os.environ["DEPLOYMENT_NAME"]

    try:
        trigger_restart(namespace, deployment)
    except requests.RequestException as e:
        print(
            f"Failed to trigger restart for {deployment}: {e}",
            file=sys.stderr,
        )
        sys.exit(1)

    # Persist hash only after successful restart
    hash_file.write_text(current_hash)
    print(f"Rolling restart triggered for {deployment}")


if __name__ == "__main__":
    main()
```

Btw since there is a single file the script could simply copy the file and check if the file are identical or not it could be as simple as that also

salt/metalk8s/addons/prometheus-operator/deployed/oidc-proxy-alertmanager.sls
```diff
@@ -60,9 +60,18 @@ Create oauth2-proxy-alertmanager Deployment:
                   value: secret
                 - name: UNIQUE_FILENAMES
                   value: "true"
+                - name: SCRIPT
+                  value: /scripts/restart-on-ca-change.sh
+                - name: POD_NAMESPACE
+                  value: metalk8s-monitoring
+                - name: DEPLOYMENT_NAME
+                  value: oauth2-proxy-alertmanager
```
TeddyAndrieux: Why one is named `DEPLOYMENT_` and the other `POD_` maybe better to named both `DEPLOYMENT_`

salt/tests/unit/scripts/test_restart_on_ca_change.py
TeddyAndrieux, line 1: Can we move this one to a `scripts` folder in `unit` directory maybe

TeddyAndrieux (commented):
I do not really like using script in configmap like that 😕 but I don't have any quick easy way to handle it so ... let's go like this for now

TeddyAndrieux (approved):
LGTM

MonPote (approved):
LGTM

PR #4827, opened 2026-03-18
MK8S-197 - UI update to include auth token

ChengYanJin:


shell-ui/src/alerts/AlertProvider.tsx
```diff
@@ -18,13 +19,18 @@ export default function AlertProvider({
...
-    // refetch the alerts every 10 seconds
-    refetchInterval: 30000,
-    // TODO manage this refresh interval globally
-    // avoid stucking at the hard loading state before alertmanager is ready
-    initialData: [],
-  });
+  const { userData } = useAuth();
+  const query = useQuery(
+    ['activeAlerts', userData?.token],
+    () => getAlerts(alertManagerUrl, userData?.token),
```
MonPote: Shouldn't we use the `getToken` function  like this ?
https://github.com/scality/metalk8s/blob/9d31df275e0caa069bf07c12e018e2c16751d631/ui/src/hooks/nodes.ts#L134

shell-ui/src/alerts/services/alertManager.ts
```diff
@@ -29,8 +29,10 @@ export type AlertLabels = {
   selectors?: string[];
   [labelName: string]: string;
 };
-export function getAlerts(alertManagerUrl: string) {
-  return fetch(alertManagerUrl + '/api/v2/alerts')
+export function getAlerts(alertManagerUrl: string, token?: string) {
```
MonPote, line 30: I think we should always have a token in our case.
Even if the prometheus is not hide behind the proxy.

ui/src/components/DashboardAlerts.tsx
```diff
@@ -61,25 +61,45 @@ const DashboardAlerts = () => {
...
-          {intl.formatMessage({
-            id: 'platform_active_alerts',
-          })}
-        </Text>
-        <TextBadge
-          variant="infoPrimary"
-          data-testid="all-alert-badge"
-          text={totalAlerts}
-        />
+      <div style={{ display: 'flex', alignItems: 'center', gap: spacing.r8 }}>
```
MonPote: nit: Maybe we should we Box. (not important at all)

ui/src/containers/PrometheusAuthProvider.tsx
```diff
@@ -0,0 +1,23 @@
+import { ReactNode, useEffect, useState } from 'react';
```
MonPote: I think this is not necessary, instead you should lazy load the token at the resquest`ui/src/services/prometheus/api.ts` 

Something like this : 

```typescript
export function getAlerts(token: string) {
  if (prometheusApiClient) {
    return prometheusApiClient.get('/api/v1/alerts', { headers: { Authorization: `Bearer ${token}` } });
  }
}
```

```typescript
const query = useQuery('activeAlerts', async () => getAlerts(alertManagerUrl, await getToken()), {
    // refetch the alerts every 10 seconds
    refetchInterval: 30000,
    // TODO manage this refresh interval globally
    // avoid stucking at the hard loading state before alertmanager is ready
    initialData: [],
  });
```

MonPote (changes requested):

ui/src/containers/PrometheusAuthProvider.tsx
ChengYanJin: Unfortunately, we still have saga in MetalK8s UI. The Prometheus client is initialized in the setApiConfig saga when config loads — at that point, the user isn't authenticated yet, so there's no token available. 

Since we plan to remove the ducks soon, I'd prefer not to modify the saga now just to move the token logic there. Once we remove the ducks, we can revisit this.

shell-ui/src/alerts/AlertProvider.tsx
ChengYanJin: Just wondering any reason it's better to use `getToken()` here ? 🤔 

I think using `userData.token` is fine here. The query refetches every 30s, and when a silent token renewal happens, React state updates, which changes the query key and triggers a refetch with the new token.

ui/src/containers/PrometheusAuthProvider.tsx
MonPote: Hmm ok, can you write a comment about the intent of this and say that it's temporary.

MonPote (dismissed):
LGTM

shell-ui/src/alerts/services/alertManager.ts
ChengYanJin: the token could be null, so my original change is better ^^

shell-ui/src/alerts/services/alertManager.ts
```diff
@@ -29,8 +29,10 @@ export type AlertLabels = {
   selectors?: string[];
   [labelName: string]: string;
 };
-export function getAlerts(alertManagerUrl: string) {
-  return fetch(alertManagerUrl + '/api/v2/alerts')
+export function getAlerts(alertManagerUrl: string, token: string) {
+  return fetch(alertManagerUrl + '/api/v2/alerts', {
+    headers: { Authorization: `Bearer ${token}` },
```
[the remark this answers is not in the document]
ChengYanJin: Thank you claude, checkActiveAlertProvider is not used anymore, I will remove it.

ui/src/containers/PrometheusAuthProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children,
+}: {
+  children: ReactNode;
+}) {
+  const { userData } = useAuth();
+  const token = userData?.token;
+
+  useEffect(() => {
+    if (token) {
+      setHeaders({ Authorization: `Bearer ${token}` });
```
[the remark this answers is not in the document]
ChengYanJin: Thank you @claude[agent], I will remove the getAlerts function

ui/src/containers/PrometheusAuthProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default function PrometheusAuthProvider({
+  children,
+}: {
+  children: ReactNode;
+}) {
+  const { userData } = useAuth();
+  const token = userData?.token;
+
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
ChengYanJin: Using useLayoutEffect intentionally to avoid a race condition: children's useEffect (react-query) fires before parent's useEffect, so the first Prometheus request would go without auth headers. useLayoutEffect guarantees headers are set before any useEffect runs. The paint-blocking cost is negligible (just a property assignment).

JBWatenbergScality (approved):

MonPote (approved):

PR #4834, opened 2026-03-23
chart, salt: Add annotations to fluent-bit pod

eg-ayoub:
What/How: Add annotations containing configmap and secret to fluent-bit pod
Why: so it is restarted whenever these objects are updated.

MK8S-203

### DIFF
--- a/charts/fluent-bit.yaml
+++ b/charts/fluent-bit.yaml
@@ -5,9 +5,8 @@ image:
 existingConfigMap: fluent-bit
 
 podAnnotations:
-  prometheus.io/scrape: "true"
-  prometheus.io/port: "2020"
-  prometheus.io/path: /api/v1/metrics/prometheus
+  checksum/config: '__slot__:salt:metalk8s_kubernetes.get_object_digest(kind="ConfigMap", apiVersion="v1", namespace="metalk8s-logging", name="fluent-bit", path="data:fluent-bit.conf")'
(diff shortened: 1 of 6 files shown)


charts/fluent-bit.yaml
```diff
@@ -8,6 +8,8 @@ podAnnotations:
   prometheus.io/scrape: "true"
   prometheus.io/port: "2020"
   prometheus.io/path: /api/v1/metrics/prometheus
```
TeddyAndrieux: Interesting, not linked to this PR, but I don't remembered about those, to me it's not longer needed since we have a service monitor

TeddyAndrieux (dismissed):

charts/fluent-bit.yaml
eg-ayoub: done

TeddyAndrieux (dismissed):

anthony-treuillier-scality (dismissed):

salt/_modules/metalk8s_kubernetes.py
```diff
@@ -410,7 +410,9 @@ def list_objects(
     return result.to_dict()["items"]
 
 
-def get_object_digest(path=None, checksum="sha256", *args, **kwargs):
+def get_object_digest(
+    ignore_not_found=False, path=None, checksum="sha256", *args, **kwargs
```
TeddyAndrieux: I think it would be better to keep current arg ordering and put the ignore not found at the end

TeddyAndrieux (approved):

ezekiel-alexrod (approved):

PR #4848, opened 2026-03-27
improve instance name ux

MonPote:
## Context / Intent
Follow-up of this PR : https://github.com/scality/artesca/pull/4945

This PR is about the integration of the instance name edit in the UI.
The user should be able to : 
- see the instance name's name.
- edit the instance name.

This first iteration consider that everything is synchronous, we will manage the async part in a futur PR  

## Change since last review 

- ux: change the input error behavior (remove toast, inline error and banner for api error)
- refactor: move the mutation directly in `EditableDeploymentName`
- pr: link to this PR https://github.com/scality/artesca/pull/4953

Happy Path : 

https://github.com/user-attachments/assets/6b6e3eb0-3edb-42ff-8c8d-387bc04b42d3

`main` instancename CR is not availaible : 
not main available

Error input management : 

https://github.com/user-attachments/assets/f330d7e7-b981-4270-94dd-15efa834a88d

API error management : 

https://github.com/user-attachments/assets/9726d1ff-4e60-4093-92fd-7b4a7fd46a43

## How to test ?

The easiest way to to test it will the latest version of artesca (dev/4)
And use the launcher with the command : 
```
npm run v2 -- artesca metalk8s
```

## Next ?
artesca: activate `canChangeInstanceName` to true in artesca (when this is ready)

ref: https://docs.google.com/document/d/1tuDBOIulVYzQQqElrqouh9s-BDS7_6eSSCjfFGVo_mg/edit?tab=t.clq3575svngm#heading=h.g1vuioutall
core-ui spec: https://github.com/scality/core-ui/pull/1052

### DIFF
--- a/shell-ui/src/navbar/EditableDeploymentName.tsx
+++ b/shell-ui/src/navbar/EditableDeploymentName.tsx
@@ -1,0 +1,212 @@
+import { useEffect, useRef, useState } from 'react';
+import styled from 'styled-components';
+import { Modal } from '@scality/core-ui/dist/components/modal/Modal.component';
+import { Text } from '@scality/core-ui/dist/components/text/Text.component';
+import { Stack } from '@scality/core-ui/dist/spacing';
+import { Tooltip } from '@scality/core-ui/dist/components/tooltip/Tooltip.component';
+import { InfoMessage } from '@scality/core-ui/dist/components/infomessage/InfoMessage.component';
+import { SecondaryText } from '@scality/core-ui/dist/components/text/Text.component';
+import { Loader } from '@scality/core-ui/dist/components/loader/Loader.component';
+import { Button } from '@scality/core-ui/dist/components/buttonv2/Buttonv2.component';
+import { Box } from '@scality/core-ui/dist/components/box/Box';
+
+import { spacing } from '@scality/core-ui/dist/spacing';
+
+const ModalDivider = styled.hr`
+  border: none;
+  border-top: 1px solid ${(props) => props.theme.backgroundLevel3};
+  margin: 0;
+`;
+
+const InlineLoaderWrapper = styled.span`
+  display: inline-flex;
+  align-items: center;
+  svg {
+    width: 1.25em;
+    height: 1.25em;
+  }
+`;
+
+const NameInput = styled.input`
+  font-size: 1rem;
+  font-family: 'Lato';
+  color: ${(props) => props.theme.textPrimary};
+  background: ${(props) => props.theme.backgroundLevel2};
+  border: 1px solid ${(props) => props.theme.infoPrimary};
+  border-radius: ${spacing.r4};
+  padding: ${spacing.r4} ${spacing.r8};
+  outline: none;
+  min-width: 180px;
+
+  &:focus {
+    border-color: ${(props) => props.theme.selectedActive};
+    box-shadow: 0 0 0 2px ${(props) => props.theme.selectedActive}33;
+  }
+`;
+
+const KeyValueGrid = styled.dl`
+  display: grid;
+  grid-template-columns: max-content 1fr;
+  gap: ${spacing.r8} ${spacing.r16};
+  margin: 0;
+`;
+
+const KeyLabel = styled.dt`
+  color: ${(props) => props.theme.textSecondary};
+  margin: 0;
+`;
+
+const KeyValue = styled.dd`
+  color: ${(props) => props.theme.textPrimary};
+  margin: 0;
+`;
+
+/** Pill around deployment name: typography + interactive hover chrome */
+const NameTrigger = styled.span`
+  display: inline-flex;
+  align-items: center;
+  gap: ${spacing.r4};
+  padding: ${spacing.r4} ${spacing.r8};
+  border-radius: ${spacing.r16};
+  border: 1px solid transparent;
+  background: ${(props) => props.theme.backgroundLevel3};
+  font-size: 1rem;
+  font-family: 'Lato';
+  color: ${(props) => props.theme.textPrimary};
+  white-space: nowrap;
+  cursor: pointer;
+  transition: border-color 0.15s ease, background 0.15s ease, opacity 0.15s ease;
+
+  &[data-disabled='true'] {
+    opacity: 0.6;
+  }
+
+  &:not([data-disabled='true']):hover {
+    border-color: ${(props) => props.theme.infoPrimary};
+    background: ${(props) => props.theme.highlight};
+  }
+`;
+
+export function EditableDeploymentName({
+  name,
+  isPropagating,
+  onChange,
+}: {
+  name: string;
+  isPropagating: boolean;
+  onChange: (newName: string) => void;
+}) {
+  const [isEditing, setIsEditing] = useState(false);
+  const [pendingName, setPendingName] = useState(name);
+  const [modalOpen, setModalOpen] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (isEditing) {
+      setPendingName(name);
+      inputRef.current?.focus();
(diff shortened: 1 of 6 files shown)


Cuervino:
"Renaming it early helps distinguish this deployment from others in multi-deployment environments."
is better than the "running in parallel" sentence.

MonPote:
> "Renaming it early helps distinguish this deployment from others in multi-deployment environments." is better than the "running in parallel" sentence.

Oh yes indeed, I forgot this change.

Cuervino (changes requested):
Looks rather ok from the UX point of view

Not fan of the toast for the ux rules when format issue, will be much better inline.
Example here (with the small backgroundLevel1 with padding, providing visual separation from the navbar line): 
[https://jubilant-adventure-5vmkqwo.pages.github.io/?path=/docs/templates-deployment-name-edit--format-validation](https://jubilant-adventure-5vmkqwo.pages.github.io/?path=/docs/templates-deployment-name-edit--format-validation)

why the "," are authorized? Should be "- _ ."

MonPote:
> Example here (with the small backgroundLevel1 with padding, providing visual separation from the navbar line):

`,` is not allowed. It's indeed not clear. The "," is here to seperate the `-` and `_`.

Your formulation `Name must be alphanumeric and can only contain dots, dashes, and underscores.` removed this doubt.

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -1,18 +1,16 @@
...
-import { ErrorPage500 } from '@scality/core-ui/dist/components/error-pages/ErrorPage500.component';
-import { Text } from '@scality/core-ui/dist/components/text/Text.component';
-import { UserData, useAuth } from '../auth/AuthProvider';
+import { EditableDeploymentName } from './EditableDeploymentName';
+import { Icon } from '@scality/core-ui/dist/components/icon/Icon.component';
+import { Loader } from '@scality/core-ui/dist/components/loader/Loader.component';
+import { Tooltip } from '@scality/core-ui/dist/components/tooltip/Tooltip.component';
+import { useToast } from '@scality/core-ui/dist/components/toast/ToastProvider';
 
 const InstanceNameContext = createContext<{
```
damiengillesscality: We should not need a context for this feature, it just saves the instance name which is in react-query state

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -69,54 +61,123 @@ export const useInstanceNameAdapter = () => {
...
   const { data, status } = useQuery({
     queryKey: ['instanceName'],
-    queryFn: () =>
-      moduleExports[instanceNameAdapter?.module ?? ''].getInstanceName(
-        userData,
-      ),
+    queryFn: async () =>
+      moduleExports[instanceNameAdapter?.module ?? ''].getInstanceName(userData, runtimeAppConfiguration),
     onSuccess: (data) => {
       setInstanceName(data);
```
damiengillesscality: The patern is not good just put this useQuery in a hook and use the hook wherever you need data accessing it with the return value of useQuery function

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -69,54 +61,123 @@ export const useInstanceNameAdapter = () => {
...
+    enabled: !!runtimeAppConfiguration,
   });
 
   const mutation = useMutation({
-    mutationFn: ({ value }: { value: string }) => {
-      return moduleExports[instanceNameAdapter?.module ?? ''].setInstanceName(
-        userData,
-        value,
-      );
+    mutationFn: async ({ value }: { value: string }) => {
```
damiengillesscality: It should do an [optimistic update](https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates), this also help to use react query state instead of the context

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -69,54 +61,123 @@ export const useInstanceNameAdapter = () => {
...
+      return moduleExports[instanceNameAdapter?.module ?? ''].setInstanceName(userData, value, runtimeAppConfiguration);
+    },
+    onSuccess: () => {
+      queryClient.invalidateQueries(['instanceName']);
```
damiengillesscality: invalidation could be in onSettled and be removed from onSuccess/error

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -69,54 +61,123 @@ export const useInstanceNameAdapter = () => {
...
+    },
+    onError: (error) => {
+      let errorMessage = 'An error occurred while updating the deployment name';
+      if (error instanceof Error) {
+        errorMessage = error.message;
+      }
```
damiengillesscality: NIT: could be just one line on 154 error.message ?? DEFAULT_UPDATE_ERR_MESSAGE. Putting the constant at the top at the same time

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -69,54 +61,123 @@ export const useInstanceNameAdapter = () => {
...
+    onError: (error) => {
+      let errorMessage = 'An error occurred while loading the deployment name';
+      if (error instanceof Error) {
+        errorMessage = error.message;
+      }
+      showToast({
+        open: true,
+        status: 'error',
+        message: errorMessage,
+      });
```
damiengillesscality: NIT: could be just one line on 154 error.message ?? DEFAULT_LOAD_ERR_MESSAGE. Putting the constant at the top at the same time

shell-ui/src/navbar/EditableDeploymentName.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  &:not([data-disabled='true']):hover {
+    border-color: ${(props) => props.theme.infoPrimary};
+    background: ${(props) => props.theme.highlight};
+  }
+`;
+
+export function EditableDeploymentName({
+  name,
+  isPropagating,
```
damiengillesscality: NIT: perharps a isLoading or isMutationLoading would be a better name

shell-ui/src/navbar/InstanceName.tsx
MonPote: You are right, it's because of legacy code. I will remove it in an other PR

shell-ui/src/navbar/InstanceName.tsx
MonPote: I agree, I will remove this part. It's actually a bit useless.
I will however remove the context in an other PR. Since this PR is starting to big pretty big (in reworking)

shell-ui/src/navbar/InstanceName.tsx
MonPote: I've simplify this part. So we don't really need this anymore.

ChengYanJin (commented):
-

shell-ui/src/navbar/InstanceName.tsx
```diff
@@ -1,50 +1,23 @@
...
-      'useSetInstanceName must be used within a InstanceNameProvider',
-    );
-  }
-  return context.setInstanceName;
+export const useInstanceName = (): string => {
+  const { data } = useQuery({
+    queryKey: [INSTANCE_NAME_QUERY_KEY],
+    enabled: false, // Don't refetch, just read from cache
+  });
+  return (data as string) ?? '';
```
ChengYanJin:
```suggestion
  export const useInstanceName = (): string | undefined => {
    const queryClient = useQueryClient();
    return queryClient.getQueryData<string>([INSTANCE_NAME_QUERY_KEY]);
  };
```

ChengYanJin (dismissed):

ChengYanJin (approved):

Cuervino (approved):
LGTM

PR #4875, opened 2026-04-10
Share GlobalHealthBar via Module Federation

JeanMarcMilletScality:
## Context
The new overview page in artesca has a Platform status section showing the global health bar

## Summary
Extract the global health bar from the dashboard and move it in its own component
Use the new component in dashboard
Wrap the new component with its providers before sharing it via module federation

### DIFF
--- a/ui/src/components/PlatformGlobalHealthBarFederated.tsx
+++ b/ui/src/components/PlatformGlobalHealthBarFederated.tsx
@@ -30,7 +30,7 @@ export default function PlatformGlobalHealthBarFederated({
 
   /**
    * Initialize the Prometheus client and set the authorization header if the token is available
-   * The initialization of Prometheus client is neeeded here as it is shared with Module Federation to another ui
+   * The initialization of Prometheus client is needed here as it is shared with Module Federation to another ui
    * The prometheus client could not be initialized in the parent component rendering it.
    */
   useLayoutEffect(() => {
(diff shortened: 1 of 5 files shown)


ui/src/components/PlatformGlobalHealthBarFederated.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { useAuth } = useShellHooks();
+  const { userData } = useAuth();
+  const token = userData?.token;
+
+  useLayoutEffect(() => {
+    if (token) {
+      initializePrometheus(prometheusUrl);
+      setHeaders({ Authorization: `Bearer ${token}` });
+    }
+  }, [prometheusUrl, token]);
```
damiengillesscality, line 41: - Since this component displays null when no token is available we should probably make token a property of the component and do conditional rendering, that would allow us to dodge some ifs.
- Why is it a useLayoutEffect and not a useEffect ?
- I don't see initializePrometheus defined anywhere in the code, there is function that looks like it with a different name tho.
- From where will prometheusUrl come from ? Does it need to initialise the api so deep into the code ? This part can probably be bubled up or we will forget it and reinitialise it at different places causing more redraws.

damiengillesscality (changes requested):

ui/src/components/PlatformGlobalHealthBarFederated.tsx
JeanMarcMilletScality: initializePrometheus is an alias for initialize from prometheus api: 
`import { initialize as initializePrometheus, setHeaders } from '../services/prometheus/api';`

The component will be used in artesca, if the prometheus is not initialized the query will throw an error

damiengillesscality (dismissed):

ChengYanJin (approved):

damiengillesscality (approved):

PR #4898, opened 2026-04-30
improvement: Bump Salt to 3006.26

### LINKED ISSUE #3436: Upgrade Salt to 3003.1
**Component**:

'salt'

**Why this is needed**:

To follow new Salt versions

**What should be done**:

Upgrade Salt to the latest version (3003.1 when this ticket get opened)

This branch https://github.com/scality/metalk8s/compare/improvement/salt-3003

---

May need some more investigation,
An issue with Salt-API for the storage operator, I don't know if it's a bug in the new Salt version or if it's just something that changed in Salt-API (did not see anything related to this in Salt changelog).

When the storage operator tries to get the device name of a device the salt API return is wrong.

Manual tests output: 

Salt API (existing device):
``` console
$ curl -Ski -b cookies.txt https://192.168.1.100:4507/ -H "Accept: application/x-yaml" -H "Content-type: application/json" -d '[{"client": "local", "tgt": "bootstrap", "fun": "metalk8s_volumes.device_name", "arg": ["/dev/loop1"]}]'
HTTP/1.1 200 OK
Content-Type: application/x-yaml
Server: CherryPy/5.6.0
Date: Wed, 30 Jun 2021 15:04:53 GMT
Allow: GET, HEAD, POST
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: GET, POST
Access-Control-Allow-Credentials: true
Cache-Control: private
Vary: Accept-Encoding
Content-Length: 27
Set-Cookie: session_id=23be7c9b8690d47c3cc695388c337fcadcae546e; expires=Thu, 01 Jul 2021 01:04:53 GMT; Path=/

return:
- bootstrap: false
```

Salt call (existing device)
```console
$ salt-call metalk8s_volumes.device_name /dev/loop1
local:
    ----------
    result:
        loop1
    success:
        True
```

Salt call (device that does not exist)
```console
$ salt-call metalk8s_volumes.device_name /dev/invalid-path
local:
    ----------
    result:
        device `/dev/invalid-path` not found
    success:
        False
```

TeddyAndrieux:
### Salt 3006.26 upgrade
- Bump Salt to 3006.26 (master image now uses the onedir-based RPM)
- Drop the `m2crypto` package and the `m2crypto` Salt state — Salt 3006 ships `cryptography`-backed `x509_v2`, enabled via `features: { x509_v2: true }` on master and minion configs
- Remove deprecated `verbose` parameter from `x509` module calls
- Remove no-longer-needed `use_superseded` on `module.run` states
- Replace deprecated `module.wait` + `watch` pattern with `module.run` + `onchanges`
- Drop the `module.run` Salt state where it has no remaining purpose

### Salt master/minion compatibility during upgrade
- Add `minimum_auth_version: 2` on the master to accept v2-protocol minions during the rolling upgrade (to be removed in development/135)
- Add `reload_modules: False` override on the salt-minion package install to skip the post-install `module_refresh` that fails when the running 3002 minion's files are replaced by the onedir layout
- Add a dedicated salt-ssh `ssh_pre_flight` script (`ssh-preflight.sh`) that installs `python3.12` and switches the `python3` alternative on each target — Salt 3006 thin requires Python >= 3.7, RHEL/Rocky 8 ships 3.6
- Switch the bootstrap script to install `python3.12` and set the `python3` alternative
- Add `dnf` to the salt-master image so the `yumpkg` Salt module loads (its `__virtual__` rejects `microdnf`-only environments and would otherwise disable `pkg.*` including `pkg.version_cmp`)

### Certificate authority handling (upgrade path from old `x509`/m2crypto-generated certs)
- Add a shared `preserved_ski` Jinja macro that pins `subjectKeyIdentifier` on `x509.certificate_managed` to the value of the existing CA cert (or `hash` on first install) — prevents `x509_v2` from regenerating CAs with a new SKI that would invalidate every leaf cert's AKI, since `cryptography` and `m2crypto` compute the SKI differently for the same public key. Critical when upgrading clusters whose CAs were originally generated by the old `x509` (m2crypto) module.
- Apply the macro to all six CAs: kubernetes, etcd, front-proxy, dex, nginx-ingress, backup-server

### Salt orchestration fixes
- Replace `state.orchestrate_single` with `state.orchestrate` + a dedicated `patch_kubesystem_namespace.sls` — the `_single` runner has long tripped a `ReferenceError: weakly-referenced object no longer exists` on the post-state event firing, but in 3006 it now propagates as a non-zero exit code (was previously swallowed silently)
- Use a dedicated state to patch the kube-system namespace cluster-version annotation
- Create the salt-master kubeconfig after the highstate completes, not during it (kube-apiserver isn't ready yet during the bootstrap highstate)
- Fix invalid `require_in` in `deploy_node` orchestrate
- Add missing dependency on deploy core component objects in the upgrade orchestrate

### Python and OS toolchain
- Drop Python 3.6 support across the codebase: remove `python36-rpm`, `python36-pyOpenSSL`, `python36-psutil` package dependencies
- Drop the Rocky Linux grains workaround that was needed for older Salt versions
- Drop the no-longer-needed RedHat 7 logic in Salt states
- Bump the buildchain to require Python 3.10+
- Bump `lib-alert-tree` to Python 3.10
- Bump devcontainer Python to 3.10.13; drop Python 2.7 and Python 3.6 from the devcontainer
- Bump Python dependencies for tests, docs, salt tests
- Remove the `virtualenv < 20.22.0` pin in tox.ini (was needed for Python 3.6)

### Build / CI / lint
- Fix doit >= 0.37 compatibility: doit 0.37 dropped `cloudpickle` and uses stdlib `pickle`, which rejects local closures — refactor `on_failure` and `title_*` helpers in `buildchain/buildchain/iso.py` and `buildchain/buildchain/utils.py` to module-level functions (using `functools.partial` where they capture an argument)
- Fix the docs build: change `git config --global` to `git config --system` in `docs/entrypoint.sh` so the tempuser the build sudo's to inherits the `safe.directory` setting and `git describe` no longer fails (which had silently set `release = None` and broken Sphinx)
- Move docs / tests / salt-tests / buildchain `pip-compile` invocations from `tox.ini` to pre-commit hooks
- Bump black to 26.3.1 (and apply formatting)
- Bump pylint to v4.0.5 with associated dependencies
- Bump mypy to v1.20.2 with associated dependencies; add a mypy config file
- Bump yamllint to v1.38.0

### Patch

- A custom patch has been applied to make salt-ssh work with python3.12 (see https://github.com/saltstack/salt/issues/68080)

Fixes: #3436 MK8S-251


salt/_states/metalk8s.py
eg-ayoub: is it worth investigating if this is still buggy ?

tests/post/features/solutions.feature
```diff
@@ -42,12 +41,12 @@ Feature: Solutions
         When we deploy Solution 'example-solution' in environment 'example-environment' with version '1.1.0'
         Then we have 1 running pod labeled 'app=example-solution-operator' in namespace 'example-environment'
         When we import a Solution archive '/var/tmp/example-solution-next.iso'
```
eg-ayoub, line 43: maybe we can rename here also for clarity

tests/post/steps/test_solutions.py
```diff
@@ -197,12 +202,10 @@ def is_available_solution(host, request, name, solution_version):
     assert name in available_solution
     for k, v in available_solution.items():
         if k == name:
-            for item in v:
-                assert (
-                    item.get("version", None) == solution_version
-                ), "Expected solution version {}: got {}".format(
-                    solution_version, item.get("version", None)
-                )
+            available_versions = [item.get("version", None) for item in v]
```
eg-ayoub: this syntax feels like it may have false positives if solution_version is None and we have None in available_versions

images/salt-master/Dockerfile
```diff
@@ -8,26 +8,20 @@ FROM ${BASE_IMAGE}@sha256:${BASE_IMAGE_SHA256}
...
+ # its `__virtual__` rejects load if neither dnf/dnf5/tdnf/yum is on PATH;
+ # microdnf is not recognized.
+ && microdnf install -y \
+      salt-master-${SALT_VERSION} salt-api-${SALT_VERSION} salt-ssh-${SALT_VERSION} \
+      openssh-clients procps-ng \
+      dnf \
+      glibc-all-langpacks langpacks-en \
+ && salt-pip install --no-cache-dir --upgrade pip \
+ # Pin protobuff version since newer versions does not work with etcd3
+ && salt-pip install --no-cache-dir "protobuf <= 3.20.1" "etcd3 != 0.11.0" "kubernetes ~= 33.1.0" \
```
eg-ayoub: we should probably update some kind of internal doc somewhere to remember to always bump kubernetes here, 
or add a check somewhere. 
I see us potentially forgetting this quite often.

salt/metalk8s/addons/dex/certs/server.sls
```diff
@@ -13,7 +13,7 @@ include:
 Create Dex server private key:
   x509.private_key_managed:
     - name: {{ private_key_path }}
-    - bits: 4096
+    - keysize: 4096
```
eg-ayoub, line 13: this works in both salt versions ?

tests/post/features/solutions.feature
TeddyAndrieux: Rename what ? The ISO file ?

eg-ayoub (dismissed):
Is it possible to push this a little further and require 3.12 in tox and the devcontainer as well ?

tests/post/features/solutions.feature
eg-ayoub: yes since we removed all instances of -next in the other steps

tests/post/steps/test_solutions.py
TeddyAndrieux: `solution_version` is the arg from scenario it cannot be None to me but we could add an assert for it right

Btw just saw also that the top for loop does not make any sense we could directly get the dictionary value, I will update it

salt/metalk8s/addons/dex/certs/server.sls
TeddyAndrieux: No but we don't care since we have an `unless` and we never re-create key if it already exists so we fine to only put the "new" behavior

images/salt-master/Dockerfile
TeddyAndrieux: Agree, ideally it should be handled by renovate but not linked to this PR

tests/post/steps/test_solutions.py
TeddyAndrieux: done

tests/post/features/solutions.feature
TeddyAndrieux: yes but this is the actual name of the ISO file, we could suffix the ISO with the version of it but it might be a bit harder to handle it in the github action that retrieve it

eg-ayoub: makes sense, I'm ok to keep it

salt/metalk8s/defaults.yaml
```diff
@@ -68,13 +68,13 @@ networks:
...
-        expected: salt-master
+        expected: "/opt/saltstack/"
         description: Salt master publisher
       control_plane_ip:4506:
-        expected: salt-master
+        expected: "/opt/saltstack/"
         description: Salt master request server
       control_plane_ip:4507:
-        expected: salt-api
+        expected: "/opt/saltstack/"
```
eg-ayoub, line 77: this is because of the way pgrep interacts with the new salt processes,
maybe in the future we can improve this check

ezekiel-alexrod (approved):
LGTM

eg-ayoub (approved):
looks good to me, maybe we can open a ticket for the psutil thing as it's out of scope

salt/metalk8s/defaults.yaml
TeddyAndrieux: Agree that it could be better if we get the full process but I feel for that kind of check it's sufficient, anyway let's tackle it in another PR if needed

PR #4900, opened 2026-05-04
security: disable anonymous auth on kube-apiserver (MK8S-187)

g-carre:
**Component**: salt, tests

**Context**:

A security assessment flagged that the ARTESCA Kubernetes cluster permits anonymous access to the API server. With kube-apiserver running on its default `--anonymous-auth=true`, unauthenticated callers can reach endpoints bound to the kubeadm-managed `system:public-info-viewer` ClusterRoleBinding (e.g. `/version`, `/healthz`) and disclose Kubernetes / Go versions usable for further attack planning.

Reproducer (from the advisory): `curl -k https://<INSTANCE_IP>:8443/api/kubernetes/version` returns a JSON body with `gitVersion` / `goVersion` instead of `401`.

**Summary**:

- Configure kube-apiserver with an `AuthenticationConfiguration` (`/etc/kubernetes/authentication-config.yaml`, generated from a new sibling salt state mirroring `cryptconfig.sls`) that scopes anonymous access to `/livez`, `/readyz`, `/healthz` only. `--anonymous-auth=false` was tried first but is incompatible with kubelet `httpGet` probes (which cannot carry credentials) and would put the apiserver in a startup-probe crash-loop after ~250s; it is also mutually exclusive with the `AuthenticationConfiguration.anonymous` block, so it is dropped.
- **Migrate the OIDC issuer config from `--oidc-*` CLI flags into `AuthenticationConfiguration.jwt`.** K8s 1.32 rejects the two mechanisms together (`pkg/kubeapiserver/options/authentication.go`: *"authentication-config file and oidc-* flags are mutually exclusive"*). A first apiserver start can succeed before the control-plane Ingress endpoint is known (no `--oidc-*` flags emitted), but the subsequent **Reconfigure control plane Ingress** pass populates `oidc_config`, re-renders the manifest with both flag sets, and the apiserver crash-loops on startup. The same five fields the legacy code consumed (`issuerURL`, `clientID`, `CAFile`, `usernameClaim`, `groupsClaim`) feed the new `jwt` issuer block. The CA PEM is read from the salt mine entry `ingress_ca_b64` for the default Dex / Ingress path (avoiding any state-graph ordering hazard), with a `salt['file.read']` fallback for a pillar-supplied OIDC override pointing at a different CA path. When the username claim is `email`, a `claimValidationRules` CEL expression `claims.?email_verified.orValue(true)` is added to reproduce the implicit `email_verified == true` guard that `--oidc-username-claim=email` used to apply automatically. ARTESCA configures Keycloak via the same OIDC pillar override (`installer/artesca_installer/salt/artesca/base/keycloak/configured.sls`), so its reconfiguration path is unaffected.
- The new mechanism relies on the `AnonymousAuthConfigurableEndpoints` and `StructuredAuthenticationConfiguration` feature gates, both **beta and on-by-default in Kubernetes 1.32** (per `pkg/features/versioned_kube_features.go` in `release-1.32`). No `--feature-gates` flag needed for the version metalk8s pins.
- Authenticate the salt `http.wait_for_successful_query` probes that hit `/healthz` on the apiserver (one direct, six through the local apiserver-proxy `:7443` raw stream) using the existing `apiserver-kubelet-client.{crt,key}` (CN `kube-apiserver-kubelet-client`, group `system:masters`). Defence-in-depth: those probes still succeed even though `/healthz` is now anonymous-allowed, and the cert pattern matches what the etcd state already does.
- Expose `apiserver-kubelet-client.key` as `certificates.client.files['apiserver-kubelet'].key` in `salt/metalk8s/defaults.yaml` and switch the existing literal references in `apiserver/installed.sls` and `apiserver/certs/kubelet-client.sls` to it, to keep the path in a single place.
- Tests in `tests/post/features/authentication.feature`:
  - Original advisory regression scenario: anonymous `GET /api/kubernetes/version` → `401 Unauthorized`. Still passes — `/version` is not in the allowed anonymous list.
  - New scenario outline: anonymous `GET /api/kubernetes/{livez,readyz,healthz}` via the control-plane Ingress → `200 ok`.
  - New scenario outline: authenticated `GET /{livez,readyz,healthz}` directly on the API server (admin client cert pulled from the kubeconfig) → `200 ok`.

**Acceptance criteria**:

- [x] `kubectl`, salt, kubelet → apiserver, and the bootstrap/post-install flow continue to work end-to-end (single-node and multi-node e2e green).
- [x] On a deployed cluster:
  - [ ] `curl -k https://<INSTANCE_IP>:8443/api/kubernetes/version` returns `401`, body `Unauthorized`.
  - [ ] `curl -k https://<INSTANCE_IP>:8443/api/kubernetes/livez` (and `/readyz`, `/healthz`) returns `200`, body `ok`.
- [x] CI green, including the three `kube-apiserver …` scenarios in `tests/post/features/authentication.feature`.
- [x] Login via the configured OIDC IDP (Dex by default; Keycloak under ARTESCA) keeps working — verified by Dex BDD scenario and ARTESCA SIT.

**Follow-ups**:

- [MK8S-258](https://scality.atlassian.net/browse/MK8S-258): bump `apiVersion` from `apiserver.config.k8s.io/v1beta1` to `v1` once metalk8s pins Kubernetes ≥ 1.34 (no urgency — v1beta1 still served through 1.36+).

---

See: MK8S-187

[MK8S-258]: https://scality.atlassian.net/browse/MK8S-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -85,6 +78,7 @@ Create kube-apiserver Pod manifest:
           - kube-apiserver
           - --advertise-address={{ host }}
           - --allow-privileged=true
+          - --authentication-config={{ authn_config_path }}
```
TeddyAndrieux: Could you move it after the `# }` it would ease future upgrade since we use flags order from kubeadm manifest

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -192,6 +180,9 @@ Make sure kube-apiserver container is up and ready:
     - name: https://{{ host }}:6443/healthz
     - verify_ssl: True
     - ca_bundle: /etc/kubernetes/pki/ca.crt
+    - cert:
+      - {{ certificates.client.files['apiserver-kubelet'].path }}
+      - {{ certificates.client.files['apiserver-kubelet'].key }}
```
TeddyAndrieux: I don't get it, is it need since `/healthz` is still not authentified ?

salt/metalk8s/kubernetes/apiserver/installed.sls
```diff
@@ -5,6 +5,7 @@
 {%- from "metalk8s/map.jinja" import networks with context %}
 
 {%- set encryption_k8s_path = "/etc/kubernetes/encryption.conf" %}
+{%- set authn_config_path = "/etc/kubernetes/authentication-config.yaml" %}
```
TeddyAndrieux: If this path is needed in several places I suggest to move it to `defaults.yaml`

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
```diff
@@ -0,0 +1,113 @@
...
+#     exclusive with --authentication-config in Kubernetes 1.32+
+#     (pkg/kubeapiserver/options/authentication.go: "authentication-config
+#     file and oidc-* flags are mutually exclusive").
+#
+# Relies on the AnonymousAuthConfigurableEndpoints and
+# StructuredAuthenticationConfiguration feature gates, both beta and
+# on-by-default in Kubernetes 1.32+.
+
+include:
+  - .installed
```
TeddyAndrieux, line 21: Any reason to do it this way instead of the otherway around ? (including `authconfig` in `installed.sls`)

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
```diff
@@ -0,0 +1,113 @@
...
+{%- endif %}
+
+Create kube-apiserver authentication configuration:
+  file.serialize:
+    - name: {{ authn_config_path }}
+    - mode: '0600'
+    - user: root
+    - group: root
+    - makedirs: True
+    - dataset: {{ authn_config | tojson }}
```
TeddyAndrieux: I think it would be eaiser to read like this (not tested)


```suggestion
    - dataset:
        apiVersion: apiserver.config.k8s.io/v1beta1
        kind: AuthenticationConfiguration
        anonymous:
          enabled: true
          conditions:
          - path: /livez
          - path: /readyz
          - path: /healthz
        {%- if oidc_config and ca_pem %}
        jwt:
        - issuer:
            url: {{ oidc_config.issuerURL }}
            audiences:
            - {{ oidc_config.clientID }}
            certificateAuthority: {{ ca_pem }}
          claimMappings:
            username:
              claim: {{ oidc_config.usernameClaim }}
              prefix: "oidc:"
            groups:
              claim: {{ oidc_config.groupsClaim }}
              prefix: "oidc:"
          {%- if oidc_config.usernameClaim == 'email' %}
          claimValidationRules:
          - expression: "claims.?email_verified.orValue(true) == true"
            message: "email_verified claim must be true when set"
          {%- endif %}       
        {%- endif %}
```

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
```diff
@@ -0,0 +1,113 @@
...
+    "issuerURL": salt.metalk8s_network.get_control_plane_ingress_endpoint() ~ "/oidc",
+    "clientID": "oidc-auth-client",
+    "CAFile": "/etc/metalk8s/pki/nginx-ingress/ca.crt",
+    "usernameClaim": "email",
+    "groupsClaim": "groups",
+  }) %}
+{%- endif %}
+
+{#- AuthenticationConfiguration's `jwt[].issuer.certificateAuthority` field
+    expects PEM content inline, not a file path. For the default Dex case the
```
TeddyAndrieux: That's sad, it mean config has to be re-generated everytime the cert change, but ok fine

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
g-carre: That is reasonable

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
g-carre: @TeddyAndrieux cryptconfig is doing the same no ? 
Should I include it in install as well if I'm doing so for authnconfig

g-carre: As discussed leaving this unchanged.

salt/metalk8s/kubernetes/apiserver/authnconfig.sls
```diff
@@ -0,0 +1,102 @@
...
+{%- if oidc_config %}
+{%-   set ingress_ca_path = '/etc/metalk8s/pki/nginx-ingress/ca.crt' %}
+{%-   if oidc_config.get('CAFile') == ingress_ca_path %}
+{%-     set ingress_ca_mine = salt['mine.get'](pillar.metalk8s.ca.minion, 'ingress_ca_b64') %}
+{%-     if ingress_ca_mine %}
+{%-       set ca_pem = salt['hashutil.base64_b64decode'](ingress_ca_mine[pillar.metalk8s.ca.minion]) %}
+{%-     endif %}
+{%-   elif oidc_config.get('CAFile') %}
+{%-     set ca_pem = salt['file.read'](oidc_config.CAFile) %}
+{%-   endif %}
```
TeddyAndrieux: I'm not sure to get why we do this distinction ? I mean in both cases the cert is available on the host why not using it ?

g-carre: In the second case it's available on the host only after adversised.sls has been completed and this may not be the case at render time. So we basicaly read it from the mine like advertised is doing to create the file whithout having to rely on a "require:" that would slow things. 

In the case of oidc being already present we can assume that the file could be present but we do not read from the mine as the file may have been modified by users (unlikely without using salt but possible)

I have interleaved the 2 logic to make it more readable.

TeddyAndrieux (dismissed):

TeddyAndrieux (approved):

ezekiel-alexrod (approved):

PR #4998, opened 2026-06-22
tests: Retry ingress requests and dump cluster state on failure

g-carre:
## Problem

Post-upgrade ingress e2e test flakes with a one-shot connection failure (seen on the 3-node major snapshot-upgrade nightly, run 27829624212 — the upgrade itself succeeded):

```
tests/post/steps/test_ingress.py: AssertionError
  Unable to reach ingress on 'http://192.168.2.202:80/': ... [Errno 111] Connection refused
```

The request-and-assert steps did a **single `requests.get` per node IP with no retry**. The scenario only gates on `app.kubernetes.io/name=ingress-nginx` pods being `Ready`, which does not guarantee each node's host-port `:80`/`:443` is already accepting connections. During a node-by-node upgrade the workload-plane `ingress-nginx-controller` DaemonSet rolls (and calico re-establishes — kubelet shows `dial tcp 10.96.0.1:443: i/o timeout`), so a single GET can hit "connection refused" on a node whose controller is not serving yet and fail the whole test. At sosreport time all six controllers were `1/1 Running`, confirming it is transient.

## Changes (`tests/post/steps/test_ingress.py`)

- Add a shared `_assert_request_returns` helper that retries the GET + status/reason assertions using the existing `utils.retry` budget (`24 x 5s = 2 min`, matching other steps in this module), and use it from `server_request_returns` and `server_request_returns_multiple_ips`.
- **On final failure, dump live cluster ingress state** to the test output to complement the sosreport (which is collected later and can miss the transient state): ingress-nginx pods / daemonsets / services / endpoints, the calico-node daemonset, and recent ingress events. Collection is best-effort (never raises).
- Negative steps (`server_does_not_respond`, `*does_not_return`, `*not_returns*`) are left one-shot so they keep failing fast.

## Testing

Python unit/lint runs in CI (pre-commit needs python3.6, unavailable locally); the file compiles and the change reuses the module's existing `utils.retry` pattern. Functional validation is the e2e suite itself.

Issue: MK8S-331

### DIFF
--- a/tests/post/steps/test_ingress.py
+++ b/tests/post/steps/test_ingress.py
@@ -17,6 +17,93 @@ MAIN_CC_NAME = "main"
 OPERATOR_DEPLOYMENT = "metalk8s-operator-controller-manager"
 OPERATOR_NAMESPACE = "kube-system"
 
+# Grain holding the IP to reach, per plane the request targets.
+PLANE_GRAINS = {
+    "workload-plane": "metalk8s:workload_plane_ip",
+    "control-plane": "metalk8s:control_plane_ip",
+}
+
+
+def _format_endpoint(protocol, ip, port, path):
+    return "{proto}://{ip}:{port}/{path}".format(
+        proto=protocol.lower(), ip=ip, port=port, path=path or ""
+    )
+
+
+def _resolve_endpoint(host, protocol, port, plane, path):
+    """Build the endpoint URL to reach `plane`'s Ingress, from its grain IP."""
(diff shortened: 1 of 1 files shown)


g-carre:
Superseded by #5007, which fixes the root cause (wait for the ingress DaemonSet — including terminating old-revision pods — to be ready before asserting) instead of retrying the request, while keeping the cluster-state-on-failure diagnostics from here. Closing in favor of #5007.

PR #5014, opened 2026-06-30
Add support for Rocky 9

TeddyAndrieux:
**Component**: build, salt, containers, tests, docs

**Context**:

MetalK8s only supports Rocky/RHEL 8 so far. This adds Rocky 9 support so we
can default to it and drop Rocky 8 in a future release.

**Summary**:

- Add Rocky/RHEL 9 build support: `redhat-9` package builder image,
  repositories and packages
- Make the Salt states and install scripts OS-version agnostic: drop the
  hardcoded RHEL-8 special cases, build GPG keys and repo paths dynamically
  from `osmajorrelease`
- Move the `salt-master`, `metalk8s-utils` and `get-kubeadm-manifests` images
  to a Rocky 9 minimal base
- Default the e2e tests to Rocky 9 (single-node, multi-node, spawn-cluster) and
  add Rocky 9 to the nightly matrix. Upgrade/downgrade lifecycle tests stay on
  Rocky 8 on purpose, since they start from the previous version which does not
  support Rocky 9 yet
- Clean up the legacy forced `os`/`os_family` grains workaround during upgrade
- Remove the remaining CentOS mentions

**Acceptance criteria**:

- CI is green on both Rocky 8 and Rocky 9 (install + nightly)
- Formula unit tests pass with the Rocky 9 default

---

Issue: MK8S-337


eg-ayoub (dismissed):

docs/installation/setup.rst
```diff
@@ -12,7 +11,7 @@ Prerequisites
 .. _SaltStack: https://www.saltstack.com
 .. _Puppet: https://puppet.com
 
-MetalK8s_ clusters require machines running CentOS_\/RHEL_/Rocky_ 8 as
+MetalK8s_ clusters require machines running RHEL_/Rocky_ 8 and 9 as
```
m4nch0t: and --> or ?

TeddyAndrieux: Right

packages/redhat/9/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Equal to rockylinux:9.7
```
m4nch0t: why we don't use 9.8?

TeddyAndrieux: No good reason it's was just not available when this commit was initially written, I will change it

salt/metalk8s/defaults.yaml
```diff
@@ -23,7 +23,8 @@ kubeadm_preflight:
       - tar
       - util-linux            # provides nsenter, mount
       - iproute               # provides ip
-      - iptables              # provides iptables
+      - iptables-nft          # provides iptables
+      - socat                 # provides socat
```
m4nch0t, line 27: Why do we want socat ?

salt/tests/unit/formulas/config.yaml
```diff
@@ -983,6 +987,10 @@ metalk8s:
           os: Rocky/8
         "RHEL 8":
           os: RedHat/8
+        "Rocky 9":
```
g-carre, line 990: Should we remove all CentOS support

salt/tests/unit/formulas/config.yaml
```diff
@@ -22,6 +22,10 @@ metalk8s:
         os: Rocky/8
       "RHEL 8":
         os: RedHat/8
+      "Rocky 9":
```
g-carre, line 25: Should we remove Centos support ?

salt/tests/unit/formulas/config.yaml
TeddyAndrieux: We should but as of "old" MetalK8s version relying on salt 3002.9 that does not support Rocky 8 we had a workaround forcing grains to "CentOS" so in 134 we have to make sure we still support those grains otherwise we will have issue during upgrade (since people in 133 on Rocky 8 has those CentOS grains)

salt/metalk8s/defaults.yaml
TeddyAndrieux: It's useful to be able to run `kubectl port-forward` and originally it was there has a requirement by kubeadm

salt/metalk8s/salt/master/files/ssh-preflight.sh
```diff
@@ -17,5 +17,10 @@ set -eu
 
 SUDO="${1:-}"
 
+# shellcheck disable=SC1091
+. /etc/os-release
+
 $SUDO yum install -y python3.12
-$SUDO alternatives --set python3 /usr/bin/python3.12
+if [ "${VERSION_ID%%.*}" -eq 8 ]; then
```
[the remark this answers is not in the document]
g-carre: Are we skipping the alternate here only because we want python3 -> 3.9 for salt and we will explicitly use python3.12 for other needs ? 
We may need to add a comment for that.

m4nch0t (approved):

eg-ayoub (approved):

salt/metalk8s/salt/master/files/ssh-preflight.sh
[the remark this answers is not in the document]
TeddyAndrieux: In Rocky 9 you cannot setup any "alternate" for python3 it's python3 == python3.9 no choice

g-carre (approved):

PR #5017, opened 2026-07-02
chore: bump salt to 3006.27 and default python to 3.11.15

TeddyAndrieux:
We also remove no longer needed patch


ezekiel-alexrod (dismissed):

g-carre (dismissed):

ezekiel-alexrod (approved):
Let's goooo 🥳

salt/_runners/metalk8s_saltutil.py
```diff
@@ -43,23 +43,37 @@ def accept_minion(minion):
...
 
 
-def wait_minions(tgt="*", retry=10):
+def wait_minions(tgt="*", retry=10, expected_salt_version=None):
     client = salt.client.get_local_client(__opts__["conf_file"])
 
     minions = None
 
+    def minion_ready(status):
+        if expected_salt_version is None:
```
g-carre, line 52: Not a fan of  `expected_salt_version` being a `global` down the path of condition_reached and minion_ready.

g-carre (approved):

salt/_runners/metalk8s_saltutil.py
g-carre: mitigated by the fact that it is never changed

PR #5022, opened 2026-07-03
improvement(salt): wait_apiserver can verify RBAC bootstrap completion

g-carre:
## Context

Addresses **MK8S-263**. The `metalk8s.wait_apiserver` helper — and the `/healthz` polls used across the Salt orchestrates — considered the apiserver ready as soon as it answered a basic probe, *before* the `poststarthook/rbac/bootstrap-roles` hook had reconciled the default ClusterRoles. Any RBAC-dependent step run right after was exposed to a race, amplified on Kubernetes 1.33+ (kubeadm KEP-4471 local-apiserver-first join + the slow `rbac/bootstrap-roles` hook, kubernetes/kubernetes#97119). MetalK8s itself hits this (bootstrap Node-object config; uncordon on node deploy).

## What changed

**1. `metalk8s.wait_apiserver` gains a `verify_rbac` option** (`salt/_modules/metalk8s.py`)
- When `verify_rbac=True`, the apiserver is only considered ready once the `cluster-admin` ClusterRole can be retrieved — the practical indicator that bootstrap RBAC is in place, mirroring `kubectl get clusterrole cluster-admin`. Checked on every retry, reusing the existing `metalk8s_kubernetes.get_object` binding (no new deps, no raw HTTP).
- Any error reaching the apiserver during that check — including raw client/connection errors **not** wrapped as `CommandExecutionError` (e.g. during resource discovery) — is treated as "not ready yet", so the retry loop keeps polling instead of aborting.

**2. It raises on timeout instead of returning `False`**
- `module.run` only fails a state when the called function *raises* — on Salt 3002 a `False` return is recorded as a change with `result: True`, so returning `False` would let an orchestrate proceed as if the apiserver were ready. `wait_apiserver` now raises `CommandExecutionError` after exhausting `retry` attempts (same pattern as `metalk8s_etcd.check_etcd_health`). Uniform across Salt 3002/3006/3008: ready → returns `True`; not ready within budget → raises → the state fails → the orchestrate aborts.

**3. Gate the RBAC-dependent orchestrate steps on RBAC readiness** (`bootstrap/init.sls`, `deploy_node.sls`)
- **deploy_node**: the pre-uncordon `/healthz` gate → `module.run: metalk8s.wait_apiserver` with `verify_rbac: true` (existing cluster, salt-master kubeconfig already present).
- **bootstrap**: a *new* `Wait for RBAC bootstrap to complete` gate is added **after** "Deploy kubeconfig to bootstrap" and required by "Configure bootstrap Node object". The pre-kubeconfig `/healthz` gate is **kept as-is** — the salt-master kubeconfig does not exist there yet (it is intentionally skipped during the bootstrap highstate, see `salt/master/init.sls`), and `wait_apiserver`'s `ping` needs a kubeconfig; `/healthz` is an unauthenticated probe that does not.

**4. Unit tests** (`test_metalk8s.py`)
- `verify_rbac` branch: cluster-admin present → ready; absent → raises; appears after retries → ready; apiserver down → raises (RBAC never checked); RBAC check raising `CommandExecutionError` / a raw `RuntimeError` → swallowed → raises on timeout.
- `test_wait_apiserver` updated so the timeout cases assert the raise.

## Acceptance criteria (MK8S-263)

- [x] `wait_apiserver()` accepts `verify_rbac` and waits for bootstrap-roles completion when True
- [x] Documentation explains when to use `verify_rbac=True` (docstring)
- [x] Unit test covers the new branch

## Validation

Local: `pytest -k wait_apiserver` (12 passed), full `test_metalk8s.py`, black (pinned 22.8.0), salt-lint, and the formula render harness for both orchestrates — all clean. These are unit/render checks only; the orchestrate behavior change (bootstrap + node join) needs a nightly/e2e on the target Salt version for definitive confirmation.

Issue: MK8S-263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
   probe, which returns "ok" before the default ClusterRoles are reconciled,
   leading to intermittent failures — a race amplified on Kubernetes 1.33+
   (kubeadm KEP-4471)
-  (PR[#XXXX](https://github.com/scality/metalk8s/pull/XXXX))
+  (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 ### Enhancements
 
@@ -20,7 +20,7 @@
   hook has reconciled the default ClusterRoles (checked via the `cluster-admin`
   ClusterRole). This lets callers running RBAC-dependent operations right
   after the wait avoid a race that is amplified on Kubernetes 1.33+
-  (PR[#XXXX](https://github.com/scality/metalk8s/pull/XXXX))
+  (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 ## Release 133.0.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 - Bootstrap and node deployment now gate their RBAC-dependent steps
   (configuring the bootstrap Node object; uncordoning a freshly deployed node)
-  on `/readyz` rather than `/healthz`.  Waiting for the infromers to sync,
+  on `/readyz` rather than `/healthz`.  Waiting for the informers to sync,
   as `informer-sync` protects a freshly started apiserver. Preventive: both
   steps run as `system:masters` and cannot hit this race themselves
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,12 +18,7 @@
   (configuring the bootstrap Node object; uncordoning a freshly deployed node)
   on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
   read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
-  answers "ok" before the default ClusterRoles are reconciled. This is
-  preventive hardening rather than a fix for observed failures: both steps
-  authenticate as `system:masters`, which is authorized before RBAC is
-  consulted, and MetalK8s nightlies show no such failure signature — the
-  observed failures of this race are on the consumer side (ARTESCA move-cluster,
-  ARTESCA-17600)
+  answers "ok" before the default ClusterRoles are reconciled. 
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 - `metalk8s.wait_apiserver` accepts a `verify_rbac` option applying the same
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,18 +16,16 @@
 
 - Bootstrap and node deployment now gate their RBAC-dependent steps
   (configuring the bootstrap Node object; uncordoning a freshly deployed node)
-  on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
-  read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
-  answers "ok" before the default ClusterRoles are reconciled. Preventive: both
+  on `/readyz` rather than `/healthz`.  Waiting for the infromers to sync,
+  as `informer-sync` protects a freshly started apiserver. Preventive: both
   steps run as `system:masters` and cannot hit this race themselves
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 - `metalk8s.wait_apiserver` accepts a `verify_rbac` option applying the same
   check, for callers that run RBAC-dependent operations immediately after the
   wait. The probe is unauthenticated, so unlike a check on the `cluster-admin`
-  ClusterRole it is credential-independent, and it reports the hook itself
-  rather than the presence of an object that persists in etcd across apiserver
-  restarts
+  ClusterRole it is credential-independent, and it requires the aggregate
+  `/readyz` rather than one hook's own endpoint, which would skip `informer-sync`
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 ## Release 133.0.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 ## Release 133.0.12 (in development)
 
+### Bug Fixes
+
+- Bootstrap and node deployment now wait for the apiserver RBAC bootstrap
+  (`poststarthook/rbac/bootstrap-roles`) to complete before running
+  RBAC-dependent steps (configuring the bootstrap Node object; uncordoning a
+  freshly deployed node). Previously these gated only on a `/healthz` liveness
+  probe, which returns "ok" before the default ClusterRoles are reconciled,
+  leading to intermittent failures — a race amplified on Kubernetes 1.33+
+  (kubeadm KEP-4471)
+  (PR[#XXXX](https://github.com/scality/metalk8s/pull/XXXX))
+
+### Enhancements
+
+- `metalk8s.wait_apiserver` now accepts a `verify_rbac` option: when set, the
+  apiserver is only considered ready once the `rbac/bootstrap-roles` post-start
+  hook has reconciled the default ClusterRoles (checked via the `cluster-admin`
+  ClusterRole). This lets callers running RBAC-dependent operations right
+  after the wait avoid a race that is amplified on Kubernetes 1.33+
+  (PR[#XXXX](https://github.com/scality/metalk8s/pull/XXXX))
+
 ## Release 133.0.11
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,22 +4,34 @@
 
 ### Bug Fixes
 
-- Bootstrap and node deployment now wait for the apiserver RBAC bootstrap
-  (`poststarthook/rbac/bootstrap-roles`) to complete before running
-  RBAC-dependent steps (configuring the bootstrap Node object; uncordoning a
-  freshly deployed node). Previously these gated only on a `/healthz` liveness
-  probe, which returns "ok" before the default ClusterRoles are reconciled,
-  leading to intermittent failures — a race amplified on Kubernetes 1.33+
-  (kubeadm KEP-4471)
+- `metalk8s.wait_apiserver` now raises instead of returning `False` when the
+  apiserver is still not ready after the last attempt. On Salt 3002 a `False`
+  return from `module.run` is recorded as a change with `result: True`, so every
+  caller using it as a gate carried on as if the apiserver were ready once the
+  wait timed out. Note this changes the contract for any caller testing the
+  boolean return
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 ### Enhancements
 
-- `metalk8s.wait_apiserver` now accepts a `verify_rbac` option: when set, the
-  apiserver is only considered ready once the `rbac/bootstrap-roles` post-start
-  hook has reconciled the default ClusterRoles (checked via the `cluster-admin`
-  ClusterRole). This lets callers running RBAC-dependent operations right
-  after the wait avoid a race that is amplified on Kubernetes 1.33+
+- Bootstrap and node deployment now gate their RBAC-dependent steps
+  (configuring the bootstrap Node object; uncordoning a freshly deployed node)
+  on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
+  read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
+  answers "ok" before the default ClusterRoles are reconciled. This is
+  preventive hardening rather than a fix for observed failures: both steps
+  authenticate as `system:masters`, which is authorized before RBAC is
+  consulted, and MetalK8s nightlies show no such failure signature — the
+  observed failures of this race are on the consumer side (ARTESCA move-cluster,
+  ARTESCA-17600)
+  (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
+
+- `metalk8s.wait_apiserver` accepts a `verify_rbac` option applying the same
+  check, for callers that run RBAC-dependent operations immediately after the
+  wait. The probe is unauthenticated, so unlike a check on the `cluster-admin`
+  ClusterRole it is credential-independent, and it reports the hook itself
+  rather than the presence of an object that persists in etcd across apiserver
+  restarts
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 ## Release 133.0.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,8 @@
   (configuring the bootstrap Node object; uncordoning a freshly deployed node)
   on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
   read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
-  answers "ok" before the default ClusterRoles are reconciled. 
+  answers "ok" before the default ClusterRoles are reconciled. Preventive: both
+  steps run as `system:masters` and cannot hit this race themselves
   (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
 
 - `metalk8s.wait_apiserver` accepts a `verify_rbac` option applying the same
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -74,7 +74,7 @@ def _rbac_bootstrap_complete(apiserver_url=APISERVER_PROXY_URL):
     Any error reaching the apiserver is treated as "not ready yet" so the
     caller keeps retrying rather than aborting.
     """
-    url = "{}/readyz?verbose".format(apiserver_url.rstrip("/"))
+    url = f"{apiserver_url.rstrip('/')}/readyz?verbose"
     try:
         result = __salt__["http.query"](url, verify_ssl=False, status=True)
     except Exception as exc:  # pylint: disable=broad-except
@@ -182,7 +182,7 @@ def wait_apiserver(
         attempts += 1
 
     if not status:
-        message = "{} after {} attempts".format(reason, retry)
+        message = f"{reason} after {retry} attempts"
         log.error(message)
         raise CommandExecutionError(message)
 
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -57,7 +57,9 @@ def _rbac_bootstrap_complete(apiserver_url=APISERVER_PROXY_URL):
     """
     url = f"{apiserver_url.rstrip('/')}/readyz?verbose"
     try:
-        result = __salt__["http.query"](url, verify_ssl=False, status=True)
+        result = __salt__["http.query"](
+            url, verify_ssl=False, status=True, raise_error=False, timeout=10
+        )
     except Exception as exc:  # pylint: disable=broad-except
         log.debug("RBAC bootstrap check failed, treating as not ready: %s", exc)
         return False
--- a/salt/_modules/metalk8s.py
+++ b/salt/_modules/metalk8s.py
@@ -41,36 +41,16 @@ def __virtual__():
 
 
 def _rbac_bootstrap_complete(apiserver_url=APISERVER_PROXY_URL):
-    """Return True once the apiserver RBAC bootstrap has reconciled the
-    default ClusterRoles.
+    """Return True once ``/readyz`` reports overall readiness *and* its verbose
+    output shows the ``poststarthook/rbac/bootstrap-roles`` hook complete.
 
-    The ``poststarthook/rbac/bootstrap-roles`` hook (re)creates the default
-    ClusterRoles right after the apiserver starts answering requests. Until it
-    completes, every RBAC-dependent call answers ``Forbidden: RBAC: clusterrole
-    ... "cluster-admin" not found``.
-
-    Requires *both* that ``/readyz`` reports overall readiness (HTTP 200) and
-    that its verbose output shows this hook complete. Matching only the hook's
-    line would return ready while other readiness checks are still failing,
-    which is a weaker guarantee than the ``/healthz`` polls this replaces
-    already gave, so the aggregated status is kept and the hook checked on top.
-
-    Two properties this has and an object check does not (see MK8S-263):
-
-    - It is *credential-independent*. Checking that the ``cluster-admin``
-      ClusterRole can be retrieved is authorized before RBAC is consulted when
-      run as ``system:masters``, so it cannot observe the condition that makes
-      another identity fail. ``/readyz`` reports the hook itself, whoever asks.
-    - It is *per-instance and reports completion*. ``cluster-admin`` persists in
-      etcd, so its presence proves nothing about a freshly started apiserver
-      still reconciling RBAC -- and it is the first entry in
-      ``bootstrappolicy.ClusterRoles()``, so even on a first bootstrap it only
-      proves the hook started.
+    Until that hook completes, every RBAC-dependent call answers ``Forbidden:
+    RBAC: clusterrole ... "cluster-admin" not found``.
 
     ``/readyz`` is in the apiserver's ``--authorization-always-allow-paths``, so
-    this needs no kubeconfig and no Kubernetes client. Note that the per-check
-    endpoint ``/readyz/poststarthook/rbac/bootstrap-roles`` is *not* in that
-    list, hence the ``?verbose`` form here rather than the narrower path.
+    this needs no kubeconfig and no Kubernetes client. The per-check endpoint
+    ``/readyz/poststarthook/rbac/bootstrap-roles`` is *not* in that list, hence
+    the ``?verbose`` form rather than the narrower path.
 
     Any error reaching the apiserver is treated as "not ready yet" so the
     caller keeps retrying rather than aborting.
@@ -83,8 +63,6 @@ def _rbac_bootstrap_complete(apiserver_url=APISERVER_PROXY_URL):
         return False
 
     if result.get("error"):
-        # Expected while the apiserver is not listening yet (mid-restart, or
-        # still starting): keep polling.
         log.debug(
             "RBAC bootstrap check could not reach %s, treating as not ready: %s",
             url,
@@ -98,9 +76,6 @@ def _rbac_bootstrap_complete(apiserver_url=APISERVER_PROXY_URL):
         return True
 
     if status in (401, 403, 404):
-        # Not the race: the endpoint is unreachable for a reason polling will
-        # not fix (unexpected apiserver flags, or a version without the hook).
-        # Warn rather than silently burning the whole retry budget.
         log.warning(
             "RBAC bootstrap check got HTTP %s from %s; expected `%s` in the "
             "response body. Check the apiserver's "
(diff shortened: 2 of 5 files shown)


salt/_modules/metalk8s.py
```diff
@@ -32,17 +32,70 @@ def __virtual__():
...
+            __salt__["metalk8s_kubernetes.get_object"](
+                kind="ClusterRole",
+                apiVersion="rbac.authorization.k8s.io/v1",
+                name="cluster-admin",
+                **kwargs,
+            )
+            is not None
+        )
+    except CommandExecutionError:
+        return False
```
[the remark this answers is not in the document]
g-carre: Good catch — you're right that the docstring over-promised relative to the narrow catch. `get_object` only wraps `ApiException`/`HTTPError` from the actual call; the `get_kubeconfig`/`get_client`/resource-discovery steps can raise raw connection errors during a flapping apiserver that never become `CommandExecutionError`. Broadened to `except Exception` (with a debug log, matching the convention in `metalk8s_etcd.py`/`metalk8s_kubeconfig.py`) and added a non-`CommandExecutionError` test case. Fixed in a41563588.

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+### Bug Fixes
+
+- Bootstrap and node deployment now wait for the apiserver RBAC bootstrap
+  (`poststarthook/rbac/bootstrap-roles`) to complete before running
+  RBAC-dependent steps (configuring the bootstrap Node object; uncordoning a
+  freshly deployed node). Previously these gated only on a `/healthz` liveness
+  probe, which returns "ok" before the default ClusterRoles are reconciled,
+  leading to intermittent failures — a race amplified on Kubernetes 1.33+
+  (kubeadm KEP-4471)
+  (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
```
ezekiel-alexrod: The observed failures are on the ARTESCA move-cluster side (ARTESCA-17600: 403 for `kubernetes-admin`), not on these two steps: they run with the salt-master kubeconfig (`system:masters`), authorized before RBAC is consulted, and our nightlies show no such signature on MetalK8s runs. Suggest presenting the gates as preventive hardening rather than a fix for observed failures, so future debugging of an uncordon issue is not sent down the wrong path. Happy to reword directly.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -313,19 +313,18 @@ Run the highstate:
...
-  - name: https://127.0.0.1:7443/healthz
-  - match: 'ok'
-  - status: 200
-  - verify_ssl: false
-  - request_interval: 1
+  module.run:
+    - metalk8s.wait_apiserver:
+      - retry: 60
+      - interval: 5
+      - verify_rbac: true
```
ezekiel-alexrod: Two concerns here: this pulls kubeconfig, kubernetes client and the new raise semantics into node-deploy with unit/render coverage only, and since the salt-master kubeconfig is `system:masters` while `cluster-admin` persists in etcd, the check passes instantly on an apiserver restart or control-plane join, exactly the deploy_node scenario. The pattern this hunk removed gives a stronger guarantee with zero new machinery:

```yaml
Wait for API server to be available:
  http.wait_for_successful_query:
  - name: https://127.0.0.1:7443/readyz?verbose
  - match: 'poststarthook/rbac/bootstrap-roles ok'
  - status: 200
  - verify_ssl: false
  - request_interval: 5
```

Same for the new bootstrap gate (as plain HTTP it would not need to wait for the kubeconfig deploy). Either way, let's run a nightly on this branch before merging.

salt/_modules/metalk8s.py
```diff
@@ -32,21 +32,90 @@ def __virtual__():
...
+    the presence of the ``cluster-admin`` ClusterRole as the practical
+    indicator that this hook has completed, mirroring
+    ``kubectl get clusterrole cluster-admin``.
+
+    Any error reaching the apiserver is treated as "not ready yet" so the
+    caller keeps retrying rather than aborting.
+    """
+    try:
+        return (
+            __salt__["metalk8s_kubernetes.get_object"](
```
ezekiel-alexrod: The guarantee here is credential-dependent: the motivating failure (ARTESCA-17600) is a 403 for `kubernetes-admin`, but with the default salt-master kubeconfig (`system:masters`, authorized before RBAC) and `cluster-admin` persisting in etcd, this check passes instantly on an apiserver restart while `kubernetes-admin` still gets 403s. It only becomes a faithful probe when run with the kubeconfig the consumer's subsequent operations will use. Suggest documenting that in the docstring, or polling `/readyz?verbose` for `poststarthook/rbac/bootstrap-roles ok` (credential-independent); I updated MK8S-263 accordingly, it wrongly listed both options as equivalent.

salt/_modules/metalk8s.py
```diff
@@ -32,21 +32,90 @@ def __virtual__():
...
+                kind="ClusterRole",
+                apiVersion="rbac.authorization.k8s.io/v1",
+                name="cluster-admin",
+                **kwargs,
+            )
+            is not None
+        )
+    except Exception as exc:  # pylint: disable=broad-except
```
ezekiel-alexrod, line 63: Right call for transient errors, but permanent failures (bad kubeconfig, expired certs) are also swallowed at debug level and burn the full 5 minutes (retry=60) before the generic raise, pointing the operator at the apiserver instead of the credential. Suggest a `log.warning` here and a distinct raise message when `verify_rbac=True`, like "apiserver responded but RBAC bootstrap did not complete".

salt/_modules/metalk8s.py
```diff
@@ -32,21 +32,90 @@ def __virtual__():
...
+        salt-call metalk8s.wait_apiserver retry=30 interval=5 verify_rbac=True
     """
-    status = __salt__["metalk8s_kubernetes.ping"](**kwargs)
+
+    def _ready():
+        if not __salt__["metalk8s_kubernetes.ping"](**kwargs):
+            return False
+        if verify_rbac:
+            return _rbac_bootstrap_complete(**kwargs)
+        return True
```
ezekiel-alexrod: For the record: with `verify_rbac=True` the ping is redundant (a successful `get_object` proves the apiserver answers, and failures degrade to False anyway), doubling the per-iteration cost. Also `get_client` retries construction internally (up to 7 attempts with 5s sleeps), so a flapping apiserver can add ~30s per iteration beyond `retry x interval`; pre-existing with ping, doubled here, worth a docstring note.

salt/_modules/metalk8s.py
```diff
@@ -32,21 +32,90 @@ def __virtual__():
...
+    """
+    try:
+        return (
+            __salt__["metalk8s_kubernetes.get_object"](
+                kind="ClusterRole",
+                apiVersion="rbac.authorization.k8s.io/v1",
+                name="cluster-admin",
+                **kwargs,
+            )
+            is not None
```
ezekiel-alexrod: nit, if we keep the object check: `metalk8s_kubernetes.object_exists` already encodes `get_object(...) is not None`:

```python
return __salt__["metalk8s_kubernetes.object_exists"](
    kind="ClusterRole",
    apiVersion="rbac.authorization.k8s.io/v1",
    name="cluster-admin",
    **kwargs,
)
```

salt/_modules/metalk8s.py
```diff
@@ -32,21 +32,90 @@ def __virtual__():
...
         time.sleep(interval)
-        status = __salt__["metalk8s_kubernetes.ping"](**kwargs)
+        status = _ready()
         attempts += 1
 
     if not status:
         log.error("Kubernetes apiserver failed to respond after %d attempts", retry)
+        raise CommandExecutionError(
+            f"Kubernetes apiserver not ready after {retry} attempts"
+        )
```
ezekiel-alexrod: Right for `module.run`, and the docstring rationale is solid. Just noting it changes the contract for any external caller testing the boolean return, default path included; keep in mind for ARTESCA-17600, the first candidate consumer.

salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -80,7 +83,97 @@ def test_wait_apiserver(self, retry, status, result):
...
+        """
+        kubernetes_ping_mock = MagicMock(return_value=ping)
+
+        get_object_mock = MagicMock()
+        if isinstance(cluster_admin, list):
+            get_object_mock.side_effect = cluster_admin
+        elif isinstance(cluster_admin, Exception):
+            get_object_mock.side_effect = cluster_admin
+        else:
+            get_object_mock.return_value = cluster_admin
```
ezekiel-alexrod: nit: `side_effect` accepts both an iterable and an exception instance, so the two branches can merge. Committable from the suggestion:

```suggestion
        if isinstance(cluster_admin, (list, Exception)):
            get_object_mock.side_effect = cluster_admin
        else:
            get_object_mock.return_value = cluster_admin
```

salt/_modules/metalk8s.py
eg-ayoub: this is already reflected in the test.

CHANGELOG.md
g-carre: Agreed, and reworded. The gates moved to *Enhancements* and now say explicitly that both steps authenticate as `system:masters` (authorized before RBAC is consulted) and that MetalK8s nightlies show no such signature — the observed failures of this race are ARTESCA move-cluster / ARTESCA-17600.

I also split the entry: the raise-on-timeout stayed under *Bug Fixes*, since that one does change observed behaviour — every `wait_apiserver` gate carried on as if ready when the wait timed out. Fixed in 6d4a832e6.

salt/metalk8s/orchestrate/deploy_node.sls
g-carre: Taken, both here and for the bootstrap gate — no kubeconfig, client, or raise semantics in node-deploy any more.

One deviation from the snippet: I dropped `status: 200` and match the hook's line alone.

```yaml
Wait for API server to be available:
  http.wait_for_successful_query:
  - name: https://127.0.0.1:7443/readyz?verbose
  - match: 'poststarthook/rbac/bootstrap-roles ok'
  - verify_ssl: false
  - request_interval: 5
```

With `status: 200` the gate waits for *aggregate* readiness: `/readyz` is a 500 while any check is failing, so an unrelated one — `informer-sync`, say — would hold the uncordon back even once the ClusterRoles are in place. Matching the line alone gives per-check semantics over the always-allowed path. Pinned by `READYZ_OTHER_CHECK_FAILING` in the tests: HTTP 500, RBAC line `ok`, gate must open.

The bootstrap side simplified more than a like-for-like swap. `/readyz` is unauthenticated like `/healthz`, so I upgraded the existing gate in place and removed the separate post-kubeconfig gate — it no longer has to sit after `Deploy kubeconfig to bootstrap`, and the ordering comment I'd added is gone. Fixed in 6d4a832e6.

Agreed on the nightly before merging.

salt/_modules/metalk8s.py
g-carre: Fully taken, and thanks for updating MK8S-263 — the object check is removed rather than documented. `verify_rbac` now reads the hook's line out of `/readyz?verbose`, which is credential-independent and reports the hook itself rather than an object that persists in etcd across restarts. Both of your points are in the docstring so the next reader doesn't re-derive them.

One mechanical note on why `?verbose` rather than the narrower `/readyz/poststarthook/rbac/bootstrap-roles`: only `/healthz`, `/livez` and `/readyz` are in `--authorization-always-allow-paths`, so the per-check subpath needs credentials. That's why the ARTESCA gate can poll it with `admin.conf` while an unauthenticated Salt state can't — worth having on the record so the two repos' differing choices read as deliberate rather than divergent. Fixed in 6d4a832e6.

salt/_modules/metalk8s.py
g-carre: Good point — a credential problem pointing the operator at the apiserver for five minutes is the wrong failure mode. Now:

- `log.warning` on 401/403/404, naming `--authorization-always-allow-paths` and `--anonymous-auth` as what to check, since polling won't fix either.
- `log.debug` kept for connection errors and a pending hook, which are the expected transient states.
- Distinct raise: *"Kubernetes apiserver responded but its RBAC bootstrap (poststarthook/rbac/bootstrap-roles) did not complete after N attempts"* vs the existing *"failed to respond"*. There's a test asserting the two don't get conflated.

Worth noting the unauthenticated probe removes most of this class anyway — no kubeconfig or certs involved in the RBAC path now. Fixed in 6d4a832e6.

salt/_modules/metalk8s.py
g-carre: The redundancy dissolves with the move to `/readyz`, but for a reason worth stating rather than leaving implicit: the probe is unauthenticated, so it proves nothing about the authenticated path — `ping` and the RBAC check are now complementary. That's in the docstring.

Your `get_client` observation stands on its own and I've documented it: client construction retries internally (up to 7 attempts, 5s apart), so against a flapping apiserver one iteration can take far longer than `interval` and the total wait can exceed `retry * interval`. Fixed in 6d4a832e6.

salt/_modules/metalk8s.py
g-carre: Moot now that the object check is gone — but noted for future reference, I hadn't spotted that `object_exists` already encodes it.

salt/_modules/metalk8s.py
g-carre: Agreed, and now called out in the CHANGELOG entry so it isn't discovered by a caller testing the boolean.

On ARTESCA as first consumer: #5389 ended up with its own `wait_apiserver_rbac_ready` polling the per-check endpoint with `admin.conf`, so once both gates here use the HTTP state, `verify_rbac` has no in-tree consumer. Converging ARTESCA onto the shared helper is tracked as the second part of ARTESCA-17965 — the follow-up PR against `development/4` that gates the `Reconfigure APIServer` / `Reconfigure etcd` boundary. It is deliberately held until a MetalK8s ISO bump carries this PR, so it can call the shared helper directly instead of migrating to it later. Now that both sides read the same `/readyz` signal, that's where the two implementations become one.

salt/tests/unit/modules/test_metalk8s.py
g-carre: Applied — `isinstance(readyz, (list, Exception))`, one branch. The parameter set changed shape with the move to `/readyz` (fixtures per startup stage rather than ClusterRole values), but the merged branch carried over. Fixed in 6d4a832e6.

g-carre:
Thanks — this reshaped the PR substantially. Pushed as 6d4a832e6. Summary of what changed:

The `cluster-admin` object check is gone. Both orchestrate gates are back to `http.wait_for_successful_query`, now on `/readyz?verbose` matching the hook's own line, and `verify_rbac` reads the same signal — so no kubeconfig, no Kubernetes client, and no raise semantics in node-deploy. That also let the bootstrap gate collapse: since `/readyz` is unauthenticated, I upgraded the existing pre-kubeconfig gate in place and deleted the separate post-kubeconfig one, so the ordering workaround is gone entirely.

One deliberate deviation from your snippet, in the reply on `deploy_node.sls`: I dropped `status: 200` so the gate matches the hook's line alone, otherwise an unrelated failing readiness check can hold an RBAC-dependent step back.

I also updated the CHANGELOG framing per your first comment, and split it: the raise-on-timeout is the Bug Fix (it changes observed behaviour), the gates are Enhancements (preventive).

Verification: 106 tests pass in `test_metalk8s.py` (15 `wait_apiserver` cases, up from 12), black 22.8.0 and salt-lint 0.9.2 clean. Two caveats — python3.6 is not installable on my machine, so I ran the suite against salt 3006 with a local `salt.ext.six` shim, and the `pre-commit` hook could not run for the same reason (checks run by hand instead). That leaves the **formula render harness** (needs jinja2 <3 for `jinja2.ext.with_`) and **salt 3002 behaviour** unverified locally, resting on CI. Agreed on the nightly before merge; will dispatch and report back here.

Separately: worth documenting the split — unauthenticated HTTP state for states gating on RBAC readiness alone, `wait_apiserver(verify_rbac=True)` when the caller also needs the authenticated path verified? Happy to add a note to the module docstring or the salt-states README if you think it earns its place.

salt/metalk8s/orchestrate/deploy_node.sls
g-carre: Correcting myself here: **you were right to keep `status: 200`, and I've restored it** in 1918b583e.

My argument was that requiring 200 lets an unrelated failing check hold an RBAC-dependent step back. That part is true, but I missed the more important half: the `/healthz` gates I was replacing *already* required 200, so dropping it did not trade one guarantee for another — it removed one callers already had. Matching the hook's line alone lets the orchestrate proceed while `/readyz` is still a 500 on, say, `informer-sync`. Your snippet as written is strictly stronger than the previous gate; mine was weaker.

Both gates now require HTTP 200 *and* the hook's line, and `verify_rbac` does the same so the state and the helper cannot diverge. The unit test that asserted the old behaviour (500 + hook `ok` -> ready) is flipped, with one added case for hook-done-then-fully-ready.

What prompted the recheck: the nightly on this branch (run 31478015126, on dd45aaccc) came back with `test_expose_wp_ingress_on_vips` failing in both multi-node jobs — a TCP connect to the WP ingress VIP `192.168.2.200:80`. The `development/133.0` baseline of 2026-08-05 ran the same test files with **137 passed, 1 skipped, 0 failed**, so that one is not pre-existing, and proceeding earlier than the old gate is a plausible route to downstream flakiness like this.

I am not claiming causation: our gate logged `Result: Clean` in 8.765 ms, so the hook was already complete that early and `/healthz` would very likely have been 200 too — meaning the timing may not have differed at all in that run, and VIP/MetalLB timing is a well-known flake source. One sample either way. Restoring `status: 200` removes it as a variable, which seems worth more than the argument I was making for dropping it.

For the record on the other two nightly failures, neither looks related: the downgrade failed on `Apply cert-manager.io/v1/Certificate .../disk-management-agent-metrics-certs` with `failed calling webhook "webhook.cert-manager.io"` (and the same job failed on the 2026-07-29 baseline), and there are zero `cluster-admin" not found` occurrences anywhere in the run. Also worth noting `wait_apiserver` now has no in-tree caller at all, so the raise-on-timeout change could not have influenced any of it.

One genuinely useful thing the nightly did prove: `/readyz?verbose` is anonymously readable in practice — the gate would otherwise have spun for its full budget instead of returning in 9 ms.

Re-dispatching the nightly on 1918b583e.

CHANGELOG.md
```diff
@@ -2,6 +2,38 @@
...
+- Bootstrap and node deployment now gate their RBAC-dependent steps
+  (configuring the bootstrap Node object; uncordoning a freshly deployed node)
+  on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
+  read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
+  answers "ok" before the default ClusterRoles are reconciled. This is
+  preventive hardening rather than a fix for observed failures: both steps
+  authenticate as `system:masters`, which is authorized before RBAC is
+  consulted, and MetalK8s nightlies show no such failure signature — the
+  observed failures of this race are on the consumer side (ARTESCA move-cluster,
+  ARTESCA-17600)
```
g-carre:
```suggestion
  answers "ok" before the default ClusterRoles are reconciled. 
```

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -312,13 +312,15 @@ Run the highstate:
       - metalk8s_cordon: Cordon the node
       - salt: Check pillar before highstate
 
+# `/readyz` is unauthenticated, being in --authorization-always-allow-paths.
```
g-carre:
```suggestion
```

salt/metalk8s/orchestrate/bootstrap/init.sls
```diff
@@ -117,13 +117,17 @@ Bring bootstrap minion to highstate:
     - salt: Refresh bootstrap minion grains
     - salt: Deploy CA role on bootstrap minion
 
+# `/readyz` is unauthenticated (it is in --authorization-always-allow-paths), so
+# this runs before the salt-master kubeconfig exists -- that is skipped during
+# the bootstrap highstate, see salt/master/init.sls.
```
g-carre:
```suggestion
```

CHANGELOG.md
```diff
@@ -2,6 +2,34 @@
...
+  boolean return
+  (PR[#5022](https://github.com/scality/metalk8s/pull/5022))
+
+### Enhancements
+
+- Bootstrap and node deployment now gate their RBAC-dependent steps
+  (configuring the bootstrap Node object; uncordoning a freshly deployed node)
+  on the apiserver's `poststarthook/rbac/bootstrap-roles` hook having completed,
+  read from `/readyz?verbose`, rather than on a `/healthz` liveness probe that
+  answers "ok" before the default ClusterRoles are reconciled. Preventive: both
```
ezekiel-alexrod: I went back to MK8S-263 and the upstream refs before writing this, since it's the ticket's whole premise.

Two things don't hold. `poststarthook/*` checks are registered into healthz, livez and readyz alike (`apiserver/pkg/server/hooks.go:109` into `healthz.go:97-104`, and the apiserver's own path test lists `/healthz/poststarthook/...`), and [kubernetes#97119](https://github.com/kubernetes/kubernetes/issues/97119) says the same: the server won't become healthy or ready until every post-start hook completes. So the old gate, with `status: 200`, already waited for bootstrap-roles. The observed failures never went through it either: the wait that let the cordon slip only checked the CRI sandbox through crictl, and passed in 0s.

That said, moving to `/readyz` is still the right call, just for a different reason. `informer-sync` and `shutdown` are the only readyz-only checks, and healthz stops at `poststarthook/generic-apiserver-start-informers`, which starts the informers without waiting for them to sync. The RBAC authorizer answers from that cache, so `informer-sync` is the check that actually protects a freshly started apiserver.

So I'd keep the gate exactly as it is, and reword this entry around informer sync rather than the hook. It matters for consumers as well: a gate polling `/readyz/poststarthook/rbac/bootstrap-roles` on its own is weaker than the aggregate `/readyz`, because it skips `informer-sync`. Worth carrying into the follow-up on the consumer side.

One small correction while I'm here: MK8S-263 says `:7443/healthz` is the apiserver-proxy's own healthz. It isn't. The `stream` block passes 7443 straight to `:6443`, and nginx's own healthz sits on `:7080` (`apiserver-proxy.conf.j2`).

Sorry for the back and forth. I'm learning a bit more while reading and diving into the topic.

salt/_modules/metalk8s.py
```diff
@@ -27,26 +27,141 @@
...
+    this needs no kubeconfig and no Kubernetes client. The per-check endpoint
+    ``/readyz/poststarthook/rbac/bootstrap-roles`` is *not* in that list, hence
+    the ``?verbose`` form rather than the narrower path.
+
+    Any error reaching the apiserver is treated as "not ready yet" so the
+    caller keeps retrying rather than aborting.
+    """
+    url = f"{apiserver_url.rstrip('/')}/readyz?verbose"
+    try:
+        result = __salt__["http.query"](url, verify_ssl=False, status=True)
```
ezekiel-alexrod: `http.query` with no `timeout` falls back to `http_request_timeout`, which is 3600s (`salt/utils/http.py:540-543`, default in `salt/config/__init__.py:1223`). And the state loop only checks the clock between two queries (`salt/states/http.py:213-227`), so one hung request blows straight through `wait_for: 420` and `300`. Against an apiserver that accepts the connection and never answers, the gate hangs for an hour.

A `timeout=10` here, and `- timeout: 10` in both gates, keeps the budgets honest.

salt/_modules/metalk8s.py
```diff
@@ -27,26 +27,141 @@
...
+            result["error"],
+        )
+        return False
+
+    body = result.get("body") or ""
+    status = result.get("status")
+    if status == 200 and RBAC_BOOTSTRAP_MARKER in body:
+        return True
+
+    if status in (401, 403, 404):
```
ezekiel-alexrod, line 80: This branch can't fire today, which is a shame because it's the operator-facing one. The default backend is tornado (`salt/utils/http.py:197`) with `raise_error=True` (line 172), so any non-2xx comes back as `{"status": code, "error": ...}` with no body (lines 628-631), and the `result.get("error")` check on line 65 returns first.

Passing `raise_error=False` to `http.query` fixes it: the normal path fills in both `status` (lines 716-717) and `body` (line 660), so this warning and the debug log below finally say something useful.

salt/tests/unit/modules/test_metalk8s.py
```diff
@@ -80,7 +83,131 @@ def test_wait_apiserver(self, retry, status, result):
...
+                )
+
+    # What `/readyz?verbose` answers, per stage of the apiserver's startup.
+    READYZ_RECONCILED = {
+        "status": 200,
+        "body": "[+]etcd ok\n[+]poststarthook/rbac/bootstrap-roles ok\nreadyz check passed\n",
+    }
+    # The hook has not completed: the endpoint is a 500, and the marker is
+    # absent. This is the race the gate exists for.
+    READYZ_PENDING = {
```
ezekiel-alexrod, line 104: Related to the `http.query` note: these fixtures build `{"status": 500, "body": ...}`, a shape salt never returns for a non-2xx. `READYZ_UNREACHABLE` is actually the only realistic one.

So `READYZ_FORBIDDEN` does reach "not ready", but through line 75 instead of the `error` check on line 65, which means it passes for the wrong reason. Once the call passes `raise_error=False`, every fixture here becomes accurate as written.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -314,11 +314,12 @@ Run the highstate:
 
 Wait for API server to be available:
```
ezekiel-alexrod, line 315: This gate has no `require`, so it only runs after `Run the highstate` thanks to `state_auto_order` being on by default (`salt/config/__init__.py:1145`). The bootstrap one has `- require: - salt: Bring bootstrap minion to highstate`.

Not a bug today, but a gate that runs early passes instantly and still reports success, so the same line here would make it safe against a future reordering. I don't think we will reorder it soon, and the missing `require` was here before, but I guess that is still a good practice.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -314,11 +314,12 @@ Run the highstate:
...
   http.wait_for_successful_query:
-  - name: https://127.0.0.1:7443/healthz
-  - match: 'ok'
+  - name: https://127.0.0.1:7443/readyz?verbose
+  - match: 'poststarthook/rbac/bootstrap-roles ok'
   - status: 200
   - verify_ssl: false
-  - request_interval: 1
+  - request_interval: 5
+  - wait_for: 300
```
ezekiel-alexrod: nit: the budgets are the other way round from the rationale. Bootstrap gets 420s at 1s intervals, and the node join, the slow KEP-4471 case the docstring names, gets 300s at 5s. Worth aligning them, or saying why they differ.

salt/_modules/metalk8s.py
g-carre: Adding timout 10 here - ok

salt/metalk8s/orchestrate/deploy_node.sls
g-carre: Align tk 420 with time out 10

salt/_modules/metalk8s.py
g-carre: Also set to raise_error = False as per your comment on line 80

salt/_modules/metalk8s.py
g-carre: passing raise_error = False on line 60

salt/metalk8s/orchestrate/deploy_node.sls
g-carre: added require on High state

salt/tests/unit/modules/test_metalk8s.py
g-carre: raise_error = False added to http call

CHANGELOG.md
g-carre: Learning there too.
Got the point on informer-synch and will update for it. 

The comment on 7443/healthz is the apiserver-proxy's own healthz vs ngnix one is still unclear to me. I ll check the ticket again and code more in detail but that should not block the PR IMHO

ezekiel-alexrod (approved):
LGTM

eg-ayoub (approved):

PR #5068, opened 2026-07-30
Build boot cache images for control plane and worker nodes

ezekiel-alexrod:
**Component**: build, containers

**Context**:

A node that loses its container images cannot pull them back when the registry
holding them runs on the cluster itself: a pruned image store, a reinstall or a
cold start with no external registry leaves the node unable to run the very
workloads that would let it pull again.

MK8S-153: ship the images a node needs to boot as archives inside a container
image, so provisioning can restore them into containerd with no registry and no
working control plane.

**Summary**:

- `BOOT_CACHE_VARIANTS` in `image.py` lists what each variant caches, and
  everything else follows from it: the build context, the image, the archive.
  A name that is not pulled fails the build immediately, with a message saying
  what to fix.
- Two variants ship. `metalk8s-boot-cache-control-plane` carries `pause`, etcd,
  the API server, the controller manager, the scheduler, CoreDNS, `kube-proxy`
  and Calico. `metalk8s-boot-cache-worker` carries `pause` and
  `file-reflector`.
- Both carry `pause` because containerd pins it as its sandbox image
  (`pinned_images.sandbox` in `containerd/installed.sls`): without it the
  kubelet creates no sandbox, so a control plane node would not even start its
  static pods.
- `file-reflector` v0.2.0 is pulled for the first time here. It is the agent
  the registry operator runs to write containerd's mirror configuration on each
  node.
- The archives are converted with skopeo from the layers already pulled for the
  ISO, so nothing is downloaded twice, and they are staged under
  `_build/boot-cache/<variant>`, out of the ISO. The staging directory is
  emptied before each assembly: all of it ends up in the image, so an archive
  left over from a version bump must not survive.
- Each archive carries the name the kubelet asks for, the one the cluster
  registry serves, since a cached archive is imported as is, with no
  re-tagging, unlike the ones Salt imports through `--index-name`.
- Each image is saved both as layers, for the registry, and as an archive, for
  a node that has no registry yet: the two formats `pause` and `nginx` already
  use. `LocalImage` gained an `archive_reference` option for that.
- First unit tests for the buildchain (`buildchain/tests`), with a
  `unit-tests-buildchain` tox environment and a pre-merge job scoped by path.
  The suite is parametrised per variant, so declaring a new one covers it.

The registry variant stays declared but commented out: its three images are not
carried by the buildchain, and the stack that runs them is not part of MetalK8s
core yet. Tracked in MK8S-380, blocked by MK8S-168.

**Acceptance criteria**:

- `tox -e unit-tests-buildchain` passes (29 tests).
- `./doit.sh images` produces, for each variant,
  `_build/root/images/metalk8s-boot-cache-<variant>/<version>/` (layers) and
  `_build/root/images/metalk8s-boot-cache-<variant>-<version>.tar` (archive),
  and `_build/boot-cache/` stays out of the ISO.
- `tar tf` on an archive shows one `*.tar` per image listed for the variant,
  each named after the reference the node resolves.

Cost on the ISO: about 386 MB of layers plus a 1.1 GB archive for the control
plane variant, the deduplicated layers being shared with the images already
shipped. The worker variant adds `file-reflector` and reuses `pause`.

### DIFF
--- a/.changes/unreleased/Enhanced-20260820-140410-488064253.yaml
+++ b/.changes/unreleased/Enhanced-20260820-140410-488064253.yaml
@@ -1,0 +1,17 @@
+kind: Enhanced
+body: |-
+    Ship boot cache images on the ISO, each carrying an archive of the container images
+    a node of that role needs to boot: `metalk8s-boot-cache-control-plane` (`pause`,
+    etcd, `kube-apiserver`, `kube-controller-manager`, `kube-scheduler`, CoreDNS,
+    `kube-proxy` and Calico) and `metalk8s-boot-cache-worker` (`pause` and
+    `file-reflector`). Both carry `pause`, which containerd pins as its sandbox image:
+    without it the kubelet starts no Pod at all, not even a static one. A node can
+    restore them into containerd without reaching the registry, which breaks the
+    circular dependency between a node and the registry running on the cluster it
+    belongs to. Each archive is named after the image the kubelet asks for, so it is
+    imported as is, with no re-tagging. Both images are available as layers, for the
+    registry, and as an archive, for a node with no registry yet.
+time: 2026-08-20T14:04:10.488064253Z
+custom:
+    CommentId: "5356968643"
+    Pull: "5068"
--- a/.changes/unreleased/Enhanced-20260820-140434-10652025.yaml
+++ b/.changes/unreleased/Enhanced-20260820-140434-10652025.yaml
@@ -1,0 +1,10 @@
+kind: Enhanced
+body: |-
+    Pull the [file-reflector](https://github.com/scality/file-reflector) image
+    ([v0.2.0](https://github.com/scality/file-reflector/releases/tag/v0.2.0)), the agent
+    that writes containerd's mirror configuration on each node, so it can be cached on
+    workers.
+time: 2026-08-20T14:04:34.010652025Z
+custom:
+    CommentId: "5356969664"
+    Pull: "5068"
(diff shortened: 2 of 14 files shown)


eg-ayoub (dismissed):
I'm guessing the agent to use these image caches is implemented elsewhere, LGTM

m4nch0t (dismissed):

ezekiel-alexrod:
/changie Enhanced
Ship boot cache images on the ISO, each carrying an archive of the container images
a node of that role needs to boot: `metalk8s-boot-cache-control-plane` (`pause`,
etcd, `kube-apiserver`, `kube-controller-manager`, `kube-scheduler`, CoreDNS,
`kube-proxy` and Calico) and `metalk8s-boot-cache-worker` (`pause` and
`file-reflector`). Both carry `pause`, which containerd pins as its sandbox image:
without it the kubelet starts no Pod at all, not even a static one. A node can
restore them into containerd without reaching the registry, which breaks the
circular dependency between a node and the registry running on the cluster it
belongs to. Each archive is named after the image the kubelet asks for, so it is
imported as is, with no re-tagging. Both images are available as layers, for the
registry, and as an archive, for a node with no registry yet.

ezekiel-alexrod:
/changie Enhanced
Pull the [file-reflector](https://github.com/scality/file-reflector) image
([v0.2.0](https://github.com/scality/file-reflector/releases/tag/v0.2.0)), the agent
that writes containerd's mirror configuration on each node, so it can be cached on
workers.

anthony-treuillier-scality (approved):

g-carre (approved):

PR #5082, opened 2026-08-12
fix(salt): make the apiserver liveness probe reach the apiserver (MK8S-386)

g-carre:
## Problem

`metalk8s_kubernetes.ping` is the liveness probe for the apiserver, and it could report
healthy while the apiserver was down.

It read `DynamicClient.version`, which the dynamic client answers from an on-disk discovery
cache — `osrcp-<md5(host)>.json` under `tempfile.gettempdir()`, with no expiry, refreshed
only when the client library version changes. Both reads behind it are guarded:

- `_load_server_info`: `if not self._cache.get('version'):` → only then `GET /version`
- `parse_api_groups`: `if not self._cache.get('resources') or update:` → only then `GET /apis`

So on any node where an earlier call had succeeded, `/tmp` holds a warm cache and building
the client issues **no HTTP request at all**. `ping` returned `True` regardless of the
apiserver's state.

`metalk8s.wait_apiserver` uses `ping` as its base check, so without `verify_rbac` it could
return `True` immediately against a dead apiserver — the opposite of its purpose.

## Evidence

Reproduced against the pinned **kubernetes 30.1.0** on a local kind cluster (v1.33.1),
mirroring `get_client()`:

```
[warm-up]        DynamicClient OK  version=v1.33.1  (0.0s)
[warm-up]        cache /tmp/osrcp-d7945f352d0a0426bb6d4bd4490050f7.json
[warm-up]          version key: present   resources: populated (22 group-versions)

docker stop <control-plane>;  curl -sk --max-time 8 https://127.0.0.1:43729/version  ->  000

[apiserver-down] DynamicClient OK  version=v1.33.1  (0.0s)
```

Worth knowing for anyone re-checking this: hand-seeding the cache does **not** reproduce it.
An empty `resources` entry makes the second guard falsy, discovery goes to the network, the
client construction fails, and the bug looks absent. The cache has to be warmed by a real
successful call.

## Fix

Issue a `GET /version` explicitly in `ping`. `get_version_info` keeps its cheap cached
behaviour for callers that only want the version string — the fix belongs in the probe, not
in the version getter.

`verify_rbac` was never affected: it queries `/readyz?verbose` through `http.query`, with no
client library and no cache.

## Test plan

`test_ping` now asserts the probe issues `GET /version` and never touches `.version`.

Both of its cases fail against the previous implementation, so a revert is caught — verified
by reverting only `_modules/metalk8s_kubernetes_utils.py` and keeping the test:

- `test_ping_0` fails on `request.assert_called_once_with("get", "/version")` — the old code
  issues no request
- `test_ping_1` fails with `False != True` — the old code reports healthy while the
  connection is broken

With the fix, 146 pass across `test_metalk8s_kubernetes_utils.py` and `test_metalk8s.py`;
black 22.8.0 and pylint 2.13.9 (`salt/.pylintrc`) clean.

Local caveats, since `tox`/`pre-commit` pin python3.6 which is not installable on my machine:
the suite was run against salt 3006 + kubernetes 30.1.0 with a `salt.ext.six` shim, and the
checks were run at their pinned versions by hand. The formula render harness needs jinja2 <3
and could not run. CI is authoritative for all three.

The library-level caching proof is deliberately **not** in the unit suite — it needs a real
apiserver and a warm cache file, which tests the `kubernetes` package rather than MetalK8s.
It is recorded in MK8S-386 instead.

## Not in this PR

The probe still has no request timeout, so a hung apiserver can stall it beyond
`get_client`'s internal retries (7 attempts, 5s apart). Left alone deliberately to keep this
change to the caching defect.

## Note for reviewers

Based on `bugfix/MK8S-263-wait-apiserver-verify-rbac` (#5022) rather than
`development/133.0`, so the diff here is just these three files and a nightly on this branch
exercises both changes together. Needs retargeting to `development/133.0` once #5022 merges.

Issue: MK8S-386

### DIFF
--- a/salt/_modules/metalk8s_kubernetes_utils.py
+++ b/salt/_modules/metalk8s_kubernetes_utils.py
@@ -98,9 +98,12 @@ def ping(**kwargs):
     CLI Example:
         salt '*' metalk8s_kubernetes.ping
     """
+    kubeconfig, context = get_kubeconfig(**kwargs)
+
     try:
-        get_version_info(**kwargs)
-    except CommandExecutionError:
+        client = __utils__["metalk8s_kubernetes.get_client"](kubeconfig, context)
+        client.request("get", "/version")
+    except (ApiException, HTTPError):
         return False
     return True
 
--- a/salt/_modules/metalk8s_kubernetes_utils.py
+++ b/salt/_modules/metalk8s_kubernetes_utils.py
@@ -24,6 +24,8 @@ except ImportError:
 
 __virtualname__ = "metalk8s_kubernetes"
 
+PING_REQUEST_TIMEOUT = 10
+
 
 def __virtual__():
     if MISSING_DEPS:
@@ -95,6 +97,9 @@ def ping(**kwargs):
 
     Returns True if a request could be made, False otherwise.
 
+    The response body is discarded: this is a liveness check, not a version
+    read.
+
     CLI Example:
         salt '*' metalk8s_kubernetes.ping
     """
@@ -102,7 +107,12 @@ def ping(**kwargs):
 
     try:
         client = __utils__["metalk8s_kubernetes.get_client"](kubeconfig, context)
-        client.request("get", "/version")
+        client.request(
+            "get",
+            "/version",
+            serialize=False,
+            _request_timeout=PING_REQUEST_TIMEOUT,
+        )
     except (ApiException, HTTPError):
         return False
     return True
--- a/salt/tests/unit/modules/test_metalk8s_kubernetes_utils.py
+++ b/salt/tests/unit/modules/test_metalk8s_kubernetes_utils.py
@@ -137,23 +137,32 @@ class Metalk8sKubernetesUtilsTestCase(TestCase, mixins.LoaderModuleMockMixin):
     def test_ping(
         self,
         result,
-        get_version_info_error=False,
+        k8s_connection_raise=False,
     ):
         """
         Tests the return of `ping` function
         """
-        version_info_mock = MagicMock()
-        if get_version_info_error:
-            version_info_mock.side_effect = CommandExecutionError(
-                "Failed to get version info"
-            )
+        kubeconfig_mock = MagicMock(return_value=("my_kubeconfig.conf", "my_context"))
+
+        dynamic_client_mock = MagicMock()
+        if k8s_connection_raise:
+            dynamic_client_mock.request.side_effect = HTTPError("Failed to connect")
+
+        get_client_mock = MagicMock(return_value=dynamic_client_mock)
+
         with patch.object(
-            metalk8s_kubernetes_utils, "get_version_info", version_info_mock
+            metalk8s_kubernetes_utils, "get_kubeconfig", kubeconfig_mock
+        ), patch.dict(
+            metalk8s_kubernetes_utils.__utils__,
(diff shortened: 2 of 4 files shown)


salt/_modules/metalk8s_kubernetes_utils.py
```diff
@@ -98,9 +98,12 @@ def ping(**kwargs):
     CLI Example:
         salt '*' metalk8s_kubernetes.ping
     """
+    kubeconfig, context = get_kubeconfig(**kwargs)
+
     try:
-        get_version_info(**kwargs)
-    except CommandExecutionError:
+        client = __utils__["metalk8s_kubernetes.get_client"](kubeconfig, context)
+        client.request("get", "/version")
```
ezekiel-alexrod: Careful, this raises against a *healthy* apiserver. `meta_request` defaults `serializer` to `ResourceInstance` and only catches `ValueError` (kubernetes 30.1.0, `dynamic/client.py:53` and `:63`), `ResourceInstance.__init__` opens with `kind = instance['kind']` (`dynamic/resource.py:291`), and `/version` has no `kind` (`apimachinery/pkg/version/types.go:22`). So a successful 200 gives `KeyError: 'kind'`, which line 106 doesn't catch. I reproduced it on the pinned 30.1.0.

`serialize=False` fixes it and says out loud that the body is discarded. I'd pass `_request_timeout=` too (`dynamic/client.py:286`): the old warm-cache `ping` issued no request at all, and without a timeout urllib3 never gives up.

salt/tests/unit/modules/test_metalk8s_kubernetes_utils.py
```diff
@@ -137,24 +137,33 @@ def test_get_version_info(
...
         Tests the return of `ping` function
         """
-        version_info_mock = MagicMock()
-        if get_version_info_error:
-            version_info_mock.side_effect = CommandExecutionError(
-                "Failed to get version info"
-            )
+        kubeconfig_mock = MagicMock(return_value=("my_kubeconfig.conf", "my_context"))
+
+        dynamic_client_mock = MagicMock()
```
ezekiel-alexrod, line 151: This mock is why the `KeyError` above ships green: replacing the whole client means `meta_request` and `ResourceInstance` never run. Driving the real `request` against a stubbed `call_api` holding the `/version` JSON is what tells the fix apart from the break.

Two small ones. Line 165 asserts nothing: `m.version["kubernetes"]` records the call on the child mock, so `assert_not_called` passes (I checked). Use `assert not dynamic_client_mock.version.mock_calls`. And `get_client` raising isn't covered, which is the common real failure, the way `test_get_version_info` does it on line 117.

g-carre: Thanks, added get_client raise. 
The test failed without the change  serialized=False in the utils file. And passes with it.

salt/_modules/metalk8s_kubernetes_utils.py
g-carre: Fair and need test correction which as been done too.
The call now ignore the body with serialize=False and add a time out of 10s

ezekiel-alexrod (dismissed):
LGTM

eg-ayoub (dismissed):

ezekiel-alexrod (approved):

eg-ayoub (approved):

PR #5086, opened 2026-08-13
feat(shell-ui): ARTESCA-17948 push MCP notifications via a ToolContext callback

PierreAlexandreDelgado:
## Context

WebMCP is strictly request/response and has no task support: a tool cannot write into the chat once its `execute()` has returned. A tool whose work finishes in the **background** (the async ARTESCA `downloadArtescaUpdate` / `launchArtescaUpdate`, but this is app-agnostic) therefore can't report its own outcome.

An earlier cut solved this with a **push**: a `notify` callback on `ToolContext` that posted `MCP_NOTIFICATION` to the Guardian iframe. That is replaced here by the **MCP ext-tasks** pattern — the tool returns a `taskId` and Guardian **polls** for the outcome — because WebMCP/MCP have no task transport wired through shell-ui, and polling a typed status tool is easier to reason about (and to test) than a fire-and-forget push.

## What this does

shell-ui becomes the **host aggregator** for background tasks: it collects every app's tasks and exposes one tool Guardian can poll.

- **`mcp/MCPRegistrar.tsx`**
  - Module-level ordered list `hostTasks: HostTask[]` — shared across every app's registrar instance, task 1 stays before task 2.
  - The `execute` wrapper now **`await`s** the tool result (previously it inspected the unresolved Promise). When a tool returns an object with a `taskId` **and** its descriptor exposes an arg-less `getTaskStatus()`, the wrapper registers `{ taskId, createdAt, getStatus }` in `hostTasks` (dedup by id, order preserved).
  - **`useHostGetTaskStatusTool()`** registers **one** host tool `getTaskStatus` **once** (on the shell's lifetime, not per micro-app). It routes a polled `taskId` to the owning tool's `getStatus()`, **stamps the `taskId`** back onto the result, and **evicts** a settled task after its `ttlMs` (guarded by an `evicting` flag so repeated polls of a settled task don't queue a redundant timer each time). Unknown id → `{ status: 'cancelled' }`. `readOnlyHint: true` — silent, safe to poll.
- **`mcp/types.ts`** — `ToolContext.notify` **removed** (the push channel is gone); the contract is now `getToken` / `userData` / `selfConfiguration` / `queryClient`.
- **`guardian/GuardianContext.tsx`** — the `notify` callback / `MCP_NOTIFICATION` emitter and `useGuardianNotify()` are removed; the context keeps the Guardian `iframeRef` + `guardianOrigin` (still needed by the discovery/call relay).
- **`guardian/guardianPostMessageHooks.ts` / `GuardianDrawer.tsx`** — unchanged relay (`MCP_DISCOVERY` / `MCP_CALL` / `MCP_ACK` / `MCP_RESPONSE`); the tasks polling rides the existing `MCP_CALL` / `MCP_RESPONSE` path.

Any app that returns a `{ taskId }` and exposes a descriptor `getTaskStatus` shows up automatically — nothing here is ARTESCA-specific.

## Follow-up (not in this PR)

`ToolContext` is still declared twice — here and in `artesca/@artesca/ui/src/mcpTools/types.ts` — and they diverge (artesca's has no `queryClient`). shell-ui could expose the type over module federation so micro-apps import it instead of re-declaring it.

## Related

- artesca **#5394** — the update tools return a `working` Task with a `taskId` and expose the arg-less descriptor `getTaskStatus()`; the old `getArtescaUpdateStatus` tool and `notify` are dropped.
- guardian-ui **#339** — owns each `taskId`, polls this host `getTaskStatus` @5 s, shows a tasks bar, and posts the completion into the chat.

## Testing

- `npx tsc --noEmit` — clean
- `npx jest src/mcp` — **14/14** pass (registrar: createTools/legacy registration, duplicate-name skip, AbortSignal cleanup, modelContext feature detection)
- Biome: no new violations


damiengillesscality (dismissed):

shell-ui/src/mcp/MCPRegistrar.tsx
```diff
@@ -133,6 +180,72 @@ export const _InternalMCPRegistrar = ({
   return null;
 };
```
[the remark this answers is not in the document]
PierreAlexandreDelgado: Fixed in b4a797e33 — added 10 tests for the task aggregator in `MCPRegistrar.test.tsx`:

- a tool returning `{ taskId }` lands in `hostTasks` and the host `getTaskStatus` routes the poll to that tool's arg-less `getTaskStatus()`, stamping the id back on
- per-task routing across two tools (each id hits only its owner)
- a tool returning a taskId but exposing no `getTaskStatus` is not tracked
- dedup by taskId (first owner keeps the task)
- unknown id → `cancelled` with `ttlMs: 0`
- eviction after `ttlMs` (alive at ttl-1ms, `cancelled` at ttl) with fake timers
- the `evicting` guard: 3 polls of a settled task arm exactly 1 timer (`jest.getTimerCount() === 1`) — verified this one fails if the `!entry.evicting` condition is removed
- `working`/`input_required` tasks are never evicted
- the host tool stays registered and its tasks pollable after the app registrar unmounts

`useHostGetTaskStatusTool` and a `_resetHostTasks()` helper are now exported so the module-level list can be mounted/reset per test. 24/24 tests pass, `tsc --noEmit` clean.

shell-ui/src/mcp/MCPRegistrar.tsx
```diff
@@ -110,13 +149,27 @@ export const _InternalMCPRegistrar = ({
...
                 ...(params as Record<string, unknown>),
                 ...(injectedContext && { context: injectedContext }),
               },
               client,
             );
+
+            // A background op just started: it returned a taskId AND this tool can
+            // report its status. Register it in the shared host list (dedupe by id,
+            // preserve order) so the single host `getTaskStatus` tool can route to it.
+            if (ret instanceof Object && 'taskId' in ret && tool.getTaskStatus) {
```
damiengillesscality: Would have been nice to have a TaskResultClass with props taskId and createdAt and to check `instanceof TaskResultClass` would be cleaner here and bellow imho

shell-ui/src/mcp/MCPRegistrar.tsx
```diff
@@ -133,6 +186,73 @@ export const _InternalMCPRegistrar = ({
   return null;
 };
 
+// The ONE host tool Guardian polls. Registered ONCE (globally, on the shell's lifetime
+// — not per micro-app) so it never gets unregistered while an app is still mounted.
+// It routes the polled taskId to whichever tool owns it (via the shared hostTasks
+// list), stamps the id back onto the result, and evicts a settled task after its ttlMs.
+// Exported (as well as used by MCPRegistrar below) so tests can mount it standalone.
+export function useHostGetTaskStatusTool() {
```
damiengillesscality: can be in it's own file

shell-ui/src/mcp/MCPRegistrar.tsx
```diff
@@ -133,6 +186,73 @@ export const _InternalMCPRegistrar = ({
...
+        },
+        annotations: { readOnlyHint: true }, // silent, safe to poll
+        execute: async (params: unknown) => {
+          const { taskId } = (params as { taskId: string }) ?? { taskId: '' };
+          const now = new Date().toISOString();
+          const entry = hostTasks.find((h) => h.taskId === taskId);
+          if (!entry) {
+            return {
+              taskId,
+              status: 'cancelled' as TaskStatus,
```
damiengillesscality: TaskStatusReport typing is not enough here we need this casting ?

PierreAlexandreDelgado:
/changie Added Report long-running WebMCP tool operations as background tasks: a tool hands back a task id, and Shell UI exposes a single host-wide getTaskStatus tool that routes each poll to the tool owning that task, so it stays pollable from any federated UI even after the app that started it unmounts.

PierreAlexandreDelgado:
/changie Added Support background tasks for long-running WebMCP tool operations, pollable from any UI through a single host-wide getTaskStatus tool.

shell-ui/src/mcp/MCPRegistrar.tsx
PierreAlexandreDelgado: Extracted and named — `TaskHandle` + `isTaskHandle()` in the new `shell-ui/src/mcp/tasks.ts`, so the call site is now:

```ts
if (tool.getTaskStatus && isTaskHandle(ret)) {
  registerHostTask(ret, () => tool.getTaskStatus!());
}
```

Went with a type + type guard rather than a class, because `instanceof` can't work across the federation boundary: these tools come from other apps' bundles and shell-ui's `exposes` map (`rspack.config.ts`) doesn't publish `./mcp/*` to remotes, so each side would carry its own copy of the class and `instanceof` would always be false — passing locally, failing in production. The object is also flattened to plain JSON crossing the WebMCP/relay wire.

The guard covers the same ground and tightened the check on the way: `taskId` must now actually be a string rather than merely present, and the `as Task` cast is gone.

shell-ui/src/mcp/MCPRegistrar.tsx
PierreAlexandreDelgado: Done — the whole aggregator (task model, the shared `hostTasks` registry, `registerHostTask`, and `useHostGetTaskStatusTool`) now lives in `shell-ui/src/mcp/tasks.ts`. `MCPRegistrar.tsx` is back to only registering a federated app's tools, ~110 lines lighter.

shell-ui/src/mcp/MCPRegistrar.tsx
PierreAlexandreDelgado: Good catch — no, it wasn't needed. The cast was only there because the object literal widened `status` to `string`; annotating the return type is what checks the literal against the union, so it's gone:

```ts
const unknownTask = (taskId: string): HostTaskStatus => {
  const now = new Date().toISOString();
  return {
    taskId,
    status: 'cancelled',   // checked against TaskStatus by the annotation
    ...
  };
};
```

I also gave the poll path a real return type (`Promise<HostTaskStatus>` = report + stamped taskId), which the previously inferred signature was silently dropping `result`/`error` from.

shell-ui/src/mcp/tasks.ts
```diff
@@ -0,0 +1,164 @@
...
+          properties: { taskId: { type: 'string' } },
+          required: ['taskId'],
+        },
+        annotations: { readOnlyHint: true }, // silent, safe to poll
+        execute: async (params: unknown): Promise<HostTaskStatus> => {
+          const { taskId } = (params as { taskId: string }) ?? { taskId: '' };
+          const entry = hostTasks.find((h) => h.taskId === taskId);
+          if (!entry) return unknownTask(taskId);
+
+          const task: HostTaskStatus = { taskId, ...(await entry.getStatus()) };
```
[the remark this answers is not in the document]
PierreAlexandreDelgado: Fixed. The rejection now becomes a typed `failed` task instead of escaping:

```ts
const report = await entry.getStatus().catch((error) => failedReport(entry, error));
```

`failedReport` carries the error message through as `statusMessage`/`error` and sets `ttlMs: 0`, so the existing settle path evicts the broken entry immediately rather than re-calling a throwing tool on every poll.

Covered by two tests: one asserting a throwing tool's poll resolves as `failed` while a sibling task keeps reporting normally, and one asserting the broken entry is evicted and not retried.

damiengillesscality (dismissed):

shell-ui/src/mcp/tasks.ts
```diff
@@ -0,0 +1,183 @@
...
+    );
+    if (existingNames.has('getTaskStatus')) return;
+
+    const controller = new AbortController();
+    modelContext.registerTool(
+      {
+        name: 'getTaskStatus',
+        description:
+          'Read-only. Poll a background task by its taskId to see whether it is still working, or has ' +
+          'completed/failed. Works across every app. Returns an MCP task: status is "working" until it ' +
```
JBWatenbergScality: nit: does the LLM needs to know anything about the Micro-APP context ? I think it can be removed from the prompt
```suggestion
          'completed/failed. Returns an MCP task: status is "working" until it ' +
```

JBWatenbergScality (dismissed):

JBWatenbergScality (approved):

damiengillesscality (approved):

PR #5088, opened 2026-08-13
salt: raise the master's gather_job_timeout to 300s (MK8S-388)

g-carre:
Fixes **MK8S-383** (item 1) and **MK8S-388** single root
cause.

## Problem

While a `salt.state` runs, the master polls the minion with `saltutil.find_job`. If one probe
goes unanswered within `gather_job_timeout`, Salt decides the minion is no longer running the
job and returns a bare `False`. `saltmod.state` then logs `Output from salt state not highstate`
and reports `Run failed on minions: <node>`.

**The state has not failed — Salt has given up on it.** Every requisite of that state then
cascades, so the orchestrate aborts and `upgrade.sh` fails.

Salt's default is **10s**, which is too small for exactly the states that disrupt the node being
polled:

| State | What it does to its own target |
| --- | --- |
| `Deploy apiserver <node>` (`orchestrate/apiserver.sls`) | replaces the apiserver and its proxy — MK8S-383 |
| `Run the highstate` (`orchestrate/deploy_node.sls`) | replaces kubelet, containerd and the control plane static pods — MK8S-388 |

MK8S-383 has the mechanism proved from a salt-master log: node-1 answered every probe in 60–70 ms,
missed one, and replied **19.9s late** — 9.7s past the deadline. The state was still progressing,
confirmed by the static-pod swap 55s after the orchestrate had already declared failure.
MK8S-388 has six occurrences of the same fingerprint over eight months on a production
single-node platform, and one of those aborted upgrades is the entry condition of a
three-week customer escalation.

## Why this is fixed on the master and not per state

A state cannot set this knob. `saltmod.state` builds

```python
cmd_kw = {"arg": [], "kwarg": {}, "ret": ret, "timeout": timeout}
```

and then adds only `queue`, `concurrent`, `batch`, `subset` and `failhard`. There is no
`gather_job_timeout` parameter and unknown state arguments are not forwarded, so
`- gather_job_timeout: 300` on a `salt.state` block would be **silently ineffective**.

The plumbing below it would honour one — `modules/saltutil._exec` does `cmd_kwargs.update(kwargs)`
and `LocalClient.get_iter_returns` reads
`kwargs.get("gather_job_timeout", self.opts["gather_job_timeout"])` (`client/__init__.py:1088`) —
but `saltmod.state` never sends it. Read from the Salt **3002.9** source rather than assumed;
both tickets flagged this as the open question.

### The neighbouring `- timeout:` values do not help either

Worth stating, because it is the natural thing to reach for. In `get_iter_returns`:

```python
timeout_at = time.time() + timeout                  # the FIRST window only
...
if time.time() > timeout_at and minions_running:
    timeout_at = time.time() + gather_job_timeout   # every window after that
```

A state's `- timeout:` extends only the initial wait; every subsequent keepalive round is bounded
by `gather_job_timeout`. A state lasting minutes spends nearly all of it in that regime — which is
also why raising `Reconfigure salt-minion` from 200s to 300s on `development/134.0` could not have
helped this failure, and why adding `- timeout: 600` to `Run the highstate` would look plausible
and change nothing.

This is independently corroborated by the MK8S-383 timeline: the first `find_job` came 20s after
the job was published (the master's `timeout: 20`), then every 10s (`gather_job_timeout`), and the
give-up landed exactly 10s after the unanswered probe.

## Fix

`gather_job_timeout: 300` in the salt-master configuration, pillar-overridable through
`salt:master:gather_job_timeout`, in the same shape as its neighbours `timeout` and
`worker_threads`.

300s is the same order as the `wait_for` budgets already used for apiserver readiness and leaves
ample margin over the observed 19.9s gap. The cost is one-sided: a genuinely unresponsive minion
takes up to 300s longer to be declared gone, during operations that already run for tens of
minutes.

Being on the master, it also covers the deferred gap recorded on MK8S-326 — `Sync <node> minion`
and `Refresh <node> grains` in `orchestrate/apiserver.sls` having no `retry`, so a transient
"No minions responded" fails hard.

## Why no per-state `retry`

Considered and deliberately omitted. When Salt gives up, the minion is **still running the state
and still holds its lock**, so an immediate retry returns `the function state.sls is running as
PID …`. Unless the interval outlasts the remaining run — minutes, for a highstate — the attempts
are consumed without the state ever re-running, which is worse than no retry at all. The master
setting has no such interaction. Happy to add retries as well if a reviewer prefers belt and
braces.

## Test plan

The render harness renders `master-99-metalk8s.conf.j2` with an **explicit** `extra_context`
listing every variable, so the fixture gains the new key — without it the render test fails on an
undefined variable. Verified both directions locally with `StrictUndefined`: it renders with the
key, and raises `'gather_job_timeout' is undefined` without it. The rendered output parses as YAML
and `config.yaml` still parses.

Local caveat: `tox`/`pre-commit` pin `python3.6`, which is not installable on my machine, and the
formula render harness needs `jinja2 <3` for `jinja2.ext.with_`. The template itself uses no
jinja2-2-only syntax, so I rendered it with jinja2 3.x directly. **CI is authoritative for the
render harness.**

The real check is a nightly: `snapshot-upgrade` jobs should stop failing with
`Run failed on minions: <node>` / `Output from salt state not highstate`. Both tickets note the
missed keepalive is impractical to inject deterministically, so this is a statistical result
rather than a single green run.

## Not in this PR

* **MK8S-383 item 2** — nothing waits for a master's apiserver to be serving again before the
  next one is replaced. That is an availability concern rather than the abort fixed here, and it
  needs design: the orchestrate runs on the salt-master, so `127.0.0.1:7443` checks the master's
  own proxy rather than the node just reconfigured. Left on the ticket.
* Per-state `retry`, for the reason above.

Issue: MK8S-388

### DIFF
--- a/salt/metalk8s/salt/master/configured.sls
+++ b/salt/metalk8s/salt/master/configured.sls
@@ -22,6 +22,7 @@ Configure salt master:
         saltenv: "{{ saltenv }}"
         worker_threads: {{ salt.pillar.get("salt:master:worker_threads", default=12) }}
         timeout: {{ salt.pillar.get("salt:master:timeout", default=20) }}
+        gather_job_timeout: {{ salt.pillar.get("salt:master:gather_job_timeout", default=300) }}
 
 Configure salt master roots paths:
   file.serialize:
--- a/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
+++ b/salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
@@ -3,6 +3,7 @@ interface: {{ salt_ip }}
 log_level: {{ 'debug' if debug else 'info' }}
 
 timeout: {{ timeout }}
+gather_job_timeout: {{ gather_job_timeout }}
 sock_pool_size: 15
 worker_threads: {{ worker_threads}}
 
--- a/salt/tests/unit/formulas/config.yaml
+++ b/salt/tests/unit/formulas/config.yaml
@@ -1023,6 +1023,7 @@ metalk8s:
                 saltenv: metalk8s-2.7.1
                 worker_threads: 58
                 timeout: 42
+                gather_job_timeout: 84
 
         salt-master-manifest.yaml.j2:
           _cases:


salt/metalk8s/salt/master/configured.sls
```diff
@@ -22,6 +22,7 @@ Configure salt master:
         saltenv: "{{ saltenv }}"
         worker_threads: {{ salt.pillar.get("salt:master:worker_threads", default=12) }}
         timeout: {{ salt.pillar.get("salt:master:timeout", default=20) }}
+        gather_job_timeout: {{ salt.pillar.get("salt:master:gather_job_timeout", default=300) }}
```
ezekiel-alexrod, line 25: The mechanism checks out, `client/__init__.py:1088` says what you say it says.

One correction on "a state cannot set this knob". Line 1280 re-arms `minion_timeouts[minion] = time.time() + timeout` on every keepalive reply, and the give-up at 1286-1296 needs that per-minion deadline passed too. So `- timeout: 300` on `Run the highstate` does buy a 300s grace after the last answered probe, scoped to one state. The master's 20s and 10s are close enough that the MK8S-383 timeline can't tell the two knobs apart.

The cost isn't one-sided either, since any sync call now blocks `timeout + 300` when a target never answers at all. `_runners/metalk8s_saltutil.py:59` polls at `timeout=2` inside its own `retry=10`, so one call goes from about 120s to about 50 minutes. `deploy_node.sls:14` calls `manage.up` at render time and it forwards both knobs (`runners/manage.py:207`), so one silent node stalls 320s before any log line. `scripts/upgrade.sh.in:99` and the UI's `client: 'local'` behind nginx's 60s default (`addons/ui/deployed/ingress.sls:27`) are on the same list.

Since `cmd` takes a per-call value, I'd keep the master at 60 to 120 and put `- timeout:` on the two states that need the long grace.

salt/metalk8s/salt/master/files/master-99-metalk8s.conf.j2
```diff
@@ -3,6 +3,7 @@ interface: {{ salt_ip }}
 log_level: {{ 'debug' if debug else 'info' }}
 
 timeout: {{ timeout }}
+gather_job_timeout: {{ gather_job_timeout }}
```
ezekiel-alexrod, line 6: Two bits of paperwork. `CHANGELOG.md:3` is empty for 133.0.14, and this one changes how long the master waits on any silent minion, so it earns a line (133.0.10 recorded the MK8S-326 gate). And `docs/installation/bootstrap.rst:97-100` documents `worker_threads` and `timeout` under `salt: master:`, with the prose at line 259, so `gather_job_timeout` belongs in both.

MK8S-388 also asks for a comment recording why the value was chosen, in the `.sls` next to the pillar default, the way `Reconfigure salt-minion` does.

g-carre:
Closing this: the approach is wrong, and @ezekiel-alexrod's review is why.

**The correction.** I had concluded that no state-level knob could extend the keepalive grace, because `timeout_at` is re-armed with `gather_job_timeout` (`client/__init__.py:1216` on 3002.9) and `saltmod.state` builds a fixed `cmd_kw` that cannot forward it. Both facts are true, but the inference was wrong: `get_iter_returns` has **two independent deadlines**, and give-up requires both.

```python
1198   minion_timeouts[id_] = time.time() + timeout            # armed with `timeout`
1216   timeout_at = time.time() + gather_job_timeout           # re-armed with gjt
1280   minion_timeouts[raw["data"]["id"]] = time.time() + timeout   # re-armed per reply
1289   done = (now > timeout_at) and not minions_running
1292     for id_ in minions - found:
1293         if now < minion_timeouts[id_]: done = False; break
```

I traced only `timeout_at`. The per-minion deadline is armed at 1198 and re-armed at 1280 with `timeout` — which `saltmod.state:252` and `saltmod.function:500` **do** forward. So `- timeout: 300` on a state buys a ~300s grace measured from the last answered `find_job` probe, scoped to that state. Exactly as the review says.

**Verified on both Salt versions in use**, since 133.0 and 134.0 differ here:

| | 3002.9 (133.0) | 3006.27 (134.0) |
|---|---|---|
| give-up block | 1284-1297 | 1350-1363 — **byte-identical** (`diff` clean) |
| `minion_timeouts` armed/re-armed with `timeout` | 1198 / 1280 | 1264 / 1346 |
| `timeout_at` re-armed with `gather_job_timeout` | 1216 | 1282 |
| `saltmod` forwards `timeout`, not `gather_job_timeout` | `cmd_kw` 252 / 500 | `cmd_kw` 310 / 602, `gather_job_timeout` appears 0 times |
| `manage.up` forwards both knobs | 207 | 187-207 |

**The cost claims check out too**, all references exact:

- `_runners/metalk8s_saltutil.py:59` — `client.cmd(tgt, "test.ping", timeout=2)` inside `retry=10` (line 45). With `gjt=300` a never-answering target gives up at ~`timeout + gjt` = 302s per attempt, so ~120s becomes ~50 minutes.
- `deploy_node.sls:14` — `manage.up` at **render** time, and `runners/manage.py:207` forwards both knobs, so one silent node stalls ~320s before any log line.
- `scripts/upgrade.sh.in:99` — `retry 10 10 ... salt '*' test.ping`, no explicit timeout.
- The UI's `client: 'local'`: no `proxy-read-timeout` is set anywhere in the UI or nginx-ingress addons, so nginx's 60s default does apply.

**One point beyond the review, which is why I'm dropping the master knob entirely rather than lowering it to 60–120.** Once a generous per-state `timeout` is in place, `gather_job_timeout` is only the *probe cadence*: a missed probe no longer causes give-up, because the per-minion deadline holds the loop open. `gjt=10` simply re-probes every 10s and gets *more* chances to observe the minion answering. Raising it master-wide buys nothing for this failure mode and mildly reduces probe frequency, while paying the cost on every path above.

**Where the work goes.** MK8S-388 is being closed and folded into MK8S-383 — the two states share one root cause and one fix, and splitting them across two tickets and two PRs was the reason this landed as a master-wide change in the first place. A replacement PR will add `- timeout:` to `Deploy apiserver {{ node }}` (`orchestrate/apiserver.sls`) and `Run the highstate` (`orchestrate/deploy_node.sls`), each with the NOTE comment MK8S-383 asks for, in the style of `Reconfigure salt-minion`.

Thanks for the catch — this would have shipped a 50-minute `wait_minions`.

PR #5089, opened 2026-08-17
tests: gate WP ingress VIP assertions on the CNI and on the VIP answering

g-carre:
**Component**: tests

**Context**:

The same `Connection refused` on a WP ingress VIP has now been seen with **three different causes**. Only the first was fixed by #5007; this PR covers the other two.

| # | Evidence | Cause |
|---|---|---|
| 1 | original MK8S-331 | old-revision ingress pod still `Terminating`, holding the node host-port — **fixed by #5007** |
| 2 | run [31496464787](https://github.com/scality/metalk8s/actions/runs/31496464787), [31478015126](https://github.com/scality/metalk8s/actions/runs/31478015126) | all pods Ready and calico healthy, but VIP *assignment* and *serving* diverged: `.200` answered 200 while `.201` was refused in the same loop, pods 62s/39s/15s old |
| 3 | run [32015593020](https://github.com/scality/metalk8s/actions/runs/32015593020/job/95348119159) | `calico-node` unavailable, so the replacement ingress pod never got an IP |

Cause 3, from `_ingress_debug_on_failure`:

```
calico-node                DESIRED 1  READY 0  UP-TO-DATE 1  AVAILABLE 0
ingress-nginx-controller   DESIRED 1  READY 0                AVAILABLE 0
ingress-nginx-controller-2x2f8   0/1  ContainerCreating  2s  <none>
ingress-nginx-controller endpoints: <empty>
```

Every wait the scenario performs passed. `rollout_restart` waits for in-flight rollouts of its own resource and the DaemonSet-ready step confirms the ingress DaemonSet is settled — but nothing looked at the layer underneath. An ingress pod can be scheduled and up-to-date and still never serve, because it has no network.

**Summary**:

Two commits.

**1. `cd8010d2e`** — one line in `ingress.feature`: `Workload Plane Ingress VIPs multiple pools` waits for the `calico-node` rollout after the ClusterConfig update, as its sibling `... reconfiguration` scenario already does (added by f2e18d79d). This orders the calico rollout *before* the ingress rollout-restart steps.

**2. `d09d18039`** — moves the guarantee to where the assertions are, in `tests/post/steps/test_ingress.py`:

- **Wait for the CNI** before asserting reachability, in both the positive and the negative multiple-IPs step. The negative one is a *correctness* fix, not a flakiness one: `should not return` passes for the wrong reason when the CNI is down, because everything is refused.
- **Wait for each VIP to accept a connection** before asserting on it. Checks reachability *only* — never the status code or reason — so it cannot mask a server answering the wrong thing. This is a readiness wait before asserting, not a retry of the assertion, so the negative step does not get it and keeps failing fast. That was this ticket's explicit constraint and the reason #4998 was rejected.
- **Report more on failure**: calico-node pods, recent `kube-system` events, and a per-VIP owner section pairing each VIP with the node holding it and that node's WP ingress pod.

Budgets follow the conventions already in the file: 24 × 5s for the CNI (a rollout), 12 × 5s per VIP (matching the existing VIP waits).

**Why the feature-file line stays**: it is not redundant with the assertion-time gate. It orders the calico rollout ahead of the ingress rollout-restart steps, which run well before any assertion and would otherwise churn or time out against a half-rolled CNI. The new gate protects the assertions; the feature line protects the rollout.

**Worth a reviewer's eye**: `utils.retry` reports an exhausted budget through `pytest.fail`, and `Failed` derives **straight from `BaseException`** — not `AssertionError`, not even `Exception`. So neither the existing `except AssertionError` nor a broader `except Exception` would have caught it, and a timed-out wait would have produced no cluster dump, precisely when it is most useful. Both new waits now sit inside the diagnostics context manager and it catches `pytest.fail.Exception` too. Verified against the pinned `pytest==4.3.1`, not assumed:

```
pytest 4.3.1 · pytest.fail.Exception = <class 'Failed'>
is AssertionError: False · is Exception: False · is BaseException: True
```

**Acceptance criteria**:

- [x] Reachability assertions gated on `calico-node` being available, in a shared step no scenario can omit (MK8S-331, "CNI gating").
- [x] The multi-pool scenario waits for the calico rollout, as its sibling does (MK8S-331, "Missing sibling step").
- [x] Each VIP gets a bounded wait-until-answering before its assertion; negative scenarios keep failing fast (MK8S-331, "Assigned is not serving").
- [x] `_ingress_debug` answers "why is the CNI down" and "which pod should have served this VIP" (MK8S-331, "Diagnostics").
- [ ] `single-node-e2e` and `multi-node-e2e` pass.

**Checks run locally**: `black` 22.8.0 clean; `pyflakes` clean apart from a pre-existing unused `re` import on the base, left alone. No pylint hook covers `tests/**` (only `buildchain`, `salt/_*`, `salt/tests/unit/formulas`, `packages/common/metalk8s-sosreport`), so `black` is the only formatter that applies. Full pre-commit cannot run locally — it pins python3.6.

**Not addressed here, and worth a look**: `salt/metalk8s/kubernetes/cni/calico/deployed.sls` sets `IP_AUTODETECTION_METHOD: cidr=<workload plane CIDRs>`, and the WP ingress VIPs (`192.168.2.200-204`) sit inside the WP CIDR (`192.168.2.0/24`). Calico may therefore be able to adopt a *VIP* as a node's BGP address and lose it when the VIP moves — which would explain why a calico wait is needed after every VIP-pool change at all. **Unverified**: I have no calico-node logs confirming which address it picked. If it is real, it is a product defect and the waits in this PR only hide it. Flagged in MK8S-331 as out of scope with its own ticket to follow.

---

See: MK8S-331

### DIFF
--- a/tests/post/features/ingress.feature
+++ b/tests/post/features/ingress.feature
@@ -96,6 +96,7 @@ Feature: Ingress
         And the '{wp_ingress_first_pool}' IPs are spread on nodes
         When we update the ClusterConfig to add 'test-vip-3-pool-2' Workload Plane pool with IPs '{wp_ingress_second_pool}'
         And we wait for the ClusterConfig to be 'Ready'
+        And we wait for the rollout of 'daemonset/calico-node' in namespace 'kube-system' to complete
         And we trigger a rollout restart of 'daemonset/ingress-nginx-controller' in namespace 'metalk8s-ingress'
         And we wait for the rollout of 'daemonset/ingress-nginx-controller' in namespace 'metalk8s-ingress' to complete
         Then the DaemonSet 'ingress-nginx-controller' in the 'metalk8s-ingress' namespace has all desired Pods ready
--- a/tests/post/steps/test_ingress.py
+++ b/tests/post/steps/test_ingress.py
@@ -20,13 +20,15 @@ OPERATOR_DEPLOYMENT = "metalk8s-operator-controller-manager"
 OPERATOR_NAMESPACE = "kube-system"
 
 
-def _ingress_debug(host):
+def _ingress_debug(host, ssh_config=None, ips=None):
     """Collect live cluster-side ingress diagnostics.
 
     The sosreport is collected after the whole step finishes and may miss the
     transient state (a controller still rolling, missing endpoints, calico not
     ready, ...) that caused a request to fail, so we dump the relevant cluster
     state at failure time to complement it. Best-effort: never raises.
+
+    When `ssh_config` and `ips` are given, also report which node holds each VIP.
     """
     kubectl = "kubectl --kubeconfig=/etc/kubernetes/admin.conf"
     commands = (
@@ -42,8 +44,16 @@ def _ingress_debug(host):
             f"{kubectl} -n kube-system get daemonset calico-node -o wide",
         ),
         (
+            "calico-node pods",
+            f"{kubectl} -n kube-system get pods -l k8s-app=calico-node -o wide",
+        ),
+        (
             "recent ingress events",
             f"{kubectl} -n metalk8s-ingress get events --sort-by=.lastTimestamp",
+        ),
+        (
+            "recent kube-system events",
+            f"{kubectl} -n kube-system get events --sort-by=.lastTimestamp | tail -30",
         ),
     )
     sections = []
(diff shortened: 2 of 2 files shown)


g-carre:
Abandoned in favour of #5093.

This PR gates the assertions; #5093 fixes the cause. The root cause turned out to be that `re_configure_portmap` rewrites the portmap `conditionsV4` of the calico CNI configuration, which reaches a node only when its `calico-node` Pod restarts and `install-cni` rewrites `/etc/cni/net.d/10-calico.conflist` — while an Ingress Pod's hostPort DNAT rules are programmed by the portmap plugin at sandbox creation, from whatever conflist exists then. An Ingress Pod recreated before its node caught up keeps rules that do not cover the new VIPs, and they stay that way for the life of that Pod.

So the gates here cannot catch it: the CNI gate passes (calico had finished rolling 50s before the failure, `3/3` available), the per-VIP wait-until-answering cannot succeed because the VIP will never start answering, and the feature-file line lands on *multiple pools* while the failure was in *Expose Workload Plane Ingress on some VIPs*.

#5093 waits for the `calico-node` rollout inside `re_configure_portmap` instead, so the Ingress is never restarted over a stale conflist — one wait covering all four scenarios, the `wp_ingress_vip_setup` Given step and both teardowns. Two nightlies, four multi-node job instances, no recurrence.

Two things here are worth keeping and are noted on MK8S-331 rather than lost:

- the CNI gate on the **negative** step is a correctness fix, not a flake fix: `should not return` passes vacuously when the CNI is down, since everything is refused;
- `except (AssertionError, pytest.fail.Exception)` — `utils.retry` fails through `pytest.fail`, and `Failed` derives from `BaseException`, so the current catch misses it and a timed-out wait produces no cluster dump.

PR #5090, opened 2026-08-17
Per-node upgrade aborts on transient API errors and mislabels the node (MK8S-370)

ezekiel-alexrod:
**Component**: salt

**Context**:

Two defects in the same code path (`metalk8s.orchestrate.upgrade` -> `deploy_node`), both surfaced by the same incident: the node upgrade gave up too easily, and when it gave up it left the node claiming it had succeeded.

A pod eviction rejected with a 5xx status raised right away, so a single `etcdserver: request timed out` aborted `Drain the node`, then `Deploy node <node>`, then `Upgrading all nodes one by one`, then `upgrade.sh`. That failure mode is worst during an upgrade, since the upgrade restarts etcd and the API servers itself. In the reported incident the etcd cluster was loaded, not broken: 166 `apply request took too long` entries that day, peaking above 5 s, then 9 the next day and 2 the day after, so it recovered on its own. The aborted upgrade is what left the cluster in the mixed state behind the whole deadlock.

The node version label was also set before cordon, drain and highstate, so the node that failed kept advertising the destination version although its highstate never ran. It looked exactly like the two nodes that did complete.

**Summary**:

Six commits, each standing on its own.

1. `test(formulas)`: the rendering test harness popped `_subcases` from the parsed configuration, which is shared between tests. The first test asking for the cases of a template got them all, every test after that got the parent cases only, and nothing failed. Coverage of the sub-cases silently disappeared for any template another test had already read, `containerd/installed.sls` for one. This comes first because the tests added later would otherwise degrade the same way, without failing.

2. `fix(salt)`, eviction: evictions rejected with 500, 502, 503 or 504 are retried like a 429, up to 12 consecutive errors on the same pod, so about a minute at the 5 s interval `kubectl` uses. Past that the drain fails and names the pod and the API message. The list of remaining pods was already there before this change, what's new is the retry and the reason. Forbidden and malformed requests still fail immediately. `Drain the node` also carries a `retry` for the failures that happen outside the eviction loop, with its timeout split across the attempts, so a drain that can never succeed (an unsatisfiable `PodDisruptionBudget`) still gives up in about the time a single drain used to take. Without that split, `retry: 4` on a one hour drain would have meant four hours before `upgrade.sh` reports anything.

3. `fix(salt)`, node selection: the version label cannot simply move after the highstate, which is what I first tried. It feeds `pillar.metalk8s.nodes[node].version` (`salt/_pillar/metalk8s_nodes.py`), which selects the `saltenv` used by every step of `deploy_node.sls`, the highstate included (`salt/metalk8s/orchestrate/deploy_node.sls:8`). Moving it would have made the node re-run the old version's highstate, so the upgrade would have done nothing.

   So a node now carries two annotations. `metalk8s.scality.com/version-in-progress` is set when its version is bumped and removed once its deployment succeeded, so a node whose deployment never finished is visible, and is always deployed again whatever its label says. `metalk8s.scality.com/version-applied` records the last version a node completed, and that is what both orchestrates compare the destination to. It is trusted only when it is not ahead of the label, since a run by an older version of these orchestrates maintains the label but not the annotation. A node with neither annotation falls back to the label, which is what both orchestrates always did, so nothing changes on a cluster upgraded from an older platform.

   A node whose recorded version already reads the destination is skipped, so resuming an interrupted upgrade only works the nodes that need it. That skip fires on the annotation, never on the label alone: a label bumped before a deployment that then failed says nothing about what the node runs, which is the whole defect here. Config drift is out of this path, `deploy_node` can be run per node by hand for that. Commit 6 settles where the two annotations are written.

   Selecting on what a node ran makes the skip branch reachable in more situations, which uncovered an older bug. The per-node chain waited on `Deploy node {{ loop.previtem }}`, and a skipped node declares no state, so the node right after a skipped one failed with `The following requisites were not found` and every node after it was left behind. The chain now waits on the last node actually deployed.

4. `fix(salt)`, prechecks: a node whose deployment did not complete can only be recovered by resuming the same upgrade. Bringing it to another version would mean downgrading it, which the upgrade orchestrate cannot do. Downgrade has its own script and its own gate, and not every consumer of this project exposes it. The upgrade prechecks now refuse a destination that is not the version an interrupted upgrade was heading to, and name the nodes concerned, instead of silently leaving the cluster split across two versions. The downgrade prechecks are left alone on purpose: that runner comes from the destination saltenv, so a downgrade to any released version runs the code that predates this check.

5. `docs(upgrade)`: how to resume an interrupted upgrade, and why pointing `upgrade.sh` at another version is refused.

6. `fix(salt)`, where the annotations are written: on review, both moved into `deploy_node`. Every path that deploys a node goes through that orchestrate, an upgrade and a downgrade as much as an expansion, a hostname change, or a deployment started from the UI. Written from the upgrade orchestrate, they recorded nothing on any of those other paths, so a node added after the fact carried no `version-applied` at all and the prechecks had only its label to go on. The marker is now the first state of `deploy_node`, and its last state clears it, after every state that orchestrate ends on. Which states those are depends on the roles of the node, so the list is built once and shared by the state that writes and the one that explains its failure. The upgrade orchestrate still sets the marker along with the version label, so a run interrupted between that patch and the deployment leaves the node flagged all the same. The downgrade orchestrate keeps its own two states, since it runs the `deploy_node` of its destination, an older version that knows nothing about these annotations. A comment there says when they can go.

**Acceptance criteria**:

- A simulated `500 etcdserver: request timed out` during eviction is retried and the drain eventually succeeds, while non-retryable errors still fail the state. Covered in `salt/tests/unit/modules/test_metalk8s_drain.py`: new `evict_pod` cases for 500 with a body, 503 without a readable body, 403 and 422, a drain case where the server errors stop after two attempts, a `best_effort` case, and one where they never stop and the drain fails naming the pod.
- A short API server outage during `Upgrading all nodes one by one` no longer aborts the upgrade: retried inside `evict_pods` first, then by the `Drain the node` state.
- A node whose highstate fails during an upgrade does not advertise the target version as completed: it keeps the `version-in-progress` annotation and its `version-applied` stays on the version it really ran.
- Re-running the upgrade deploys the nodes that did not complete it, and skips the ones that did. The rendering tests now assert which nodes each orchestrate deploys, in both directions and for every history a node can have, rather than only that the template renders. One case exists for the trap: a node whose label alone reads the destination is deployed, since only the annotation proves a node ran a version.
- A node added by an expansion records the version it ran, which nothing did before. A rendering test asserts both annotation states on every `deploy_node` case, and the requisite test now walks that template too, since the states those two wait on depend on the roles of the node.
- Existing upgrade and downgrade e2e scenarios unchanged. A first upgrade sees no annotation anywhere and behaves exactly as before, and a downgrade runs the orchestrate of its destination version.

**Known limits**:

- If clearing the marker fails on a node that is otherwise fine, and the next run points `upgrade.sh` at an older version, that node is deployed to the older version. It takes a failed bookkeeping write plus a destination the operator should not be using, and a marked node is by definition in an unconfirmed state. The `onfail` state tells the operator how to clear the marker by hand.
- A urllib3 `HTTPError`, a connection dropped while the API server restarts for instance, is still fatal for a single drain attempt. The 5xx statuses are the ones the incident showed, and the state-level retry covers the rest.
- After a fresh install the bootstrap node has no `version-applied`: `bootstrap.sh` does not go through `deploy_node`, so that node falls back to its label until its first upgrade, exactly as every node did before this PR.
- On a single node cluster this change is a no-op for the drain, since `upgrade/init.sls` forces `skip_draining: True` there. The drain paths were exercised on three nodes.

**Verified locally**: the Salt unit suite, 1246 tests with 100% coverage on `salt/_modules` and `salt/_runners`, plus black, pylint (spelling included), mypy `--strict`, salt-lint and yamllint at the pins of this branch and at the pins of `development/134.0` for the merge-up. Two behaviours were checked against a real Salt run rather than read off the source: `attempts` counts total executions, and an `onfail` state fires on a requisite failure too, which is why the new explanation state also requires `Deploy node`.

**Verified on a cluster**: three nodes on AWS (Rocky 8.6), installed in 133.0.13, upgraded to a 133.0.14-dev ISO built from this branch.

The drain, before and after, under the same injected 500 on evictions:

```
133.0.13:      Failed to evict pod "victim-..." in namespace "drain-test"      (no retry, drain aborted)
133.0.14-dev:  Failed to evict pod "victim-..." in namespace "drain-test":
               Internal error occurred: failed calling webhook ..., retrying   (x11)
               ... gave up after 12 consecutive server errors
133.0.14-dev:  same, injection removed mid-drain after 5 retries
               -> Eviction complete.  (11 pods evicted)
```

The injection is a `ValidatingWebhookConfiguration` on `pods/eviction` with `failurePolicy: Fail` pointing at a service that does not exist, scoped to a labelled namespace. Every eviction there gets a 500, and deleting the webhook mid-drain is what proves the recovery.

The split budget shows up as expected. With `-D 240` the drain fails on `Drain did not complete within 60 seconds`, so four attempts of 60 s rather than one of 240 s.

An upgrade interrupted by a machine reboot, which is the case this PR is about:

```
node-1  label=133.0.14-dev  version-in-progress=133.0.14-dev  version-applied=-
node-2  label=133.0.13      -                                 -
node-3  label=133.0.13      -                                 -
```

Before this change that node advertised the destination version and nothing else, so it looked exactly like a node that had completed. Resuming the same upgrade redeployed it rather than skipping it, cleared its marker and wrote its applied version. All three nodes ended at `version-applied=133.0.14-dev`.

Pointing the upgrade at another version while that node was still marked:

```
An upgrade to 133.0.14-dev did not complete on ip-10-160-112-29..., resume it instead
of upgrading to 133.0.13. Upgrading to another version would leave those nodes behind,
and bringing them to 133.0.13 would be a downgrade
```

On a node whose deployment really failed, the explanation state stayed silent, as intended: `One or more requisite failed: ... Deploy node ..., ... Mark node ... as running`. On the healthy nodes it reported `State was not run because onfail req did not change`.

A skipped node no longer strands the ones after it. With the first node recorded as having completed 999.0.0, the run printed `Skip node ... already in 999.0.0 newer than 133.0.14-dev`, produced zero `requisites were not found`, and carried on with the next node.

Re-running the upgrade orchestrate on an already upgraded cluster redeployed all three nodes, 26 states succeeded and 0 failed. That was the behaviour at the time of the run. Two things changed on review: the selection now skips a node that recorded the destination, so the same re-run would skip all three, and the annotations are written by `deploy_node` rather than by the upgrade orchestrate. Neither touches what the cluster run proved, the interruption, the resume and the refusal, and the values observed on the nodes are the same ones. The new selection and the new write path are covered by the rendering cases.

One case fell out of the run by itself. A node whose recorded version was ahead of its label got redeployed rather than skipped, which is the stale annotation guard doing its job.

---

See: MK8S-370

### DIFF
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -67,3 +67,27 @@ Upgrade
 
       The version prefix metalk8s-**X.Y.Z** must be the *new* MetalK8s version
       you want to upgrade to.
+
+Resuming an Interrupted Upgrade
+*******************************
+
+Nodes are upgraded one at a time, and each one carries the
+``metalk8s.scality.com/version-in-progress`` annotation for as long as its
+deployment runs. A node that fails mid-upgrade keeps that annotation, and
+the last version it completed stays recorded in
+``metalk8s.scality.com/version-applied``.
+
+To resume, run ``upgrade.sh`` again with the same version. Only the nodes
+that need it are deployed again: those still carrying the in-progress
+annotation, and those that never recorded the destination version. A node
+whose ``metalk8s.scality.com/version-applied`` already reads the destination
+is left alone. For as long as a node carries the in-progress annotation, the
+upgrade prechecks refuse any other destination, since moving that node to
+another version would mean downgrading it.
+
+.. important::
+
+   Do not point ``upgrade.sh`` at a version older than the one the cluster
+   runs. A node that completed a newer version is skipped, not moved back.
+   Use ``downgrade.sh`` to go back to an older version, as described in
+   :doc:`/operation/downgrade`.
--- a/docs/operation/upgrade.rst
+++ b/docs/operation/upgrade.rst
@@ -75,7 +75,9 @@ Nodes are upgraded one at a time, and each one carries the
 ``metalk8s.scality.com/version-in-progress`` annotation for as long as its
 deployment runs. A node that fails mid-upgrade keeps that annotation, and
 the last version it completed stays recorded in
-``metalk8s.scality.com/version-applied``.
+``metalk8s.scality.com/version-applied``. Both are maintained by the node
+deployment itself, so they also follow a node added through
+:doc:`/installation/expansion`.
 
 To resume, run ``upgrade.sh`` again with the same version. Only the nodes
 that need it are deployed again: those still carrying the in-progress
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -146,6 +146,18 @@ Cordon the node:
 
 {%- if run_drain %}
 
+{#- The drain talks to the API server for every pod it evicts, and an upgrade
+    restarts etcd and the API servers, so a whole drain can fail on a transient
+    error. Retry instead of aborting the node deployment, and split the drain
+    budget across the attempts: a drain that can never succeed (an unsatisfiable
+    disruption budget) then still gives up in about the time a single drain used
+    to take. Budgets of a few seconds round up to one second per attempt, since a
+    zero timeout would silently mean the 3600 second default of the `metalk8s_drain`
+    module, which is also where the 3600 below comes from. #}
+{%- set drain_attempts = 4 %}
+{%- set drain_budget = pillar.orchestrate.get("drain_timeout") | int or 3600 %}
+{%- set drain_attempt_timeout = [(drain_budget / drain_attempts) | int, 1] | max %}
+
 Drain the node:
   metalk8s_drain.node_drained:
     - name: {{ node_name }}
@@ -153,9 +165,11 @@ Drain the node:
     - ignore_pending: True
     - delete_local_data: True
     - force: True
-    {%- if pillar.orchestrate.get("drain_timeout") %}
-    - timeout: {{ pillar.orchestrate.drain_timeout }}
-    {%- endif %}
+    - timeout: {{ drain_attempt_timeout }}
+    - retry:
+        attempts: {{ drain_attempts }}
+        interval: 30
+        splay: 10
     - require:
       - metalk8s_cordon: Cordon the node
     - require_in:
--- a/salt/metalk8s/orchestrate/deploy_node.sls
+++ b/salt/metalk8s/orchestrate/deploy_node.sls
@@ -11,6 +11,21 @@
 
 {%- set roles = pillar.get('metalk8s', {}).get('nodes', {}).get(node_name, {}).get('roles', []) %}
 
+{#- Every deployment of a node goes through here, so this is where the two version
+    annotations are maintained. The version label is set by whoever asked for this
+    deployment, since it selects the saltenv, and it says what the node was asked to
+    run. These say what it actually runs (MK8S-370). #}
+Mark node {{ node_name }} as being deployed in {{ version }}:
+  metalk8s_kubernetes.object_updated:
+    - name: {{ node_name }}
+    - kind: Node
+    - apiVersion: v1
+    - patch:
+        metadata:
+          annotations:
+            metalk8s.scality.com/version-in-progress: "{{ version }}"
+    - order: 1
+
 {%- if node_name not in salt.saltutil.runner('manage.up') %}
 # Salt-ssh need python3 to be installed on the destination host, so install it
 # manually using raw ssh
@@ -398,3 +413,62 @@ Update NPD workload plane peers:
     - saltenv: {{ saltenv }}
     - require:
       - metalk8s_cordon: Uncordon the node
+
+{#- Every state this orchestrate ends on, so the node is only recorded as running
+    this version once all of them succeeded. Kept in one place, since the state
+    below and the one explaining its failure must wait on the very same set. #}
+{%- set deployment_done = [
+    "salt: Update NPD workload plane peers",
+    "salt: Kill kube-controller-manager on all master nodes",
+] %}
+{%- if 'infra' in roles and 'infra' not in skip_roles %}
+  {%- do deployment_done.append("module: Restart CoreDNS pods") %}
+{%- endif %}
+{%- if 'master' in roles and 'master' not in skip_roles %}
+  {%- do deployment_done.append("salt: Reconfigure Control Plane Ingress") %}
+{%- endif %}
+
+Mark node {{ node_name }} as running {{ version }}:
+  metalk8s_kubernetes.object_updated:
+    - name: {{ node_name }}
+    - kind: Node
+    - apiVersion: v1
+    - patch:
+        metadata:
+          annotations:
+            metalk8s.scality.com/version-in-progress: null
+            metalk8s.scality.com/version-applied: "{{ version }}"
+    {#- This deployment just restarted the API server on this node, so give this
+        write a few tries before it fails the whole thing #}
+    - retry:
+        attempts: 5
+        interval: 30
+    - require:
+      {%- for state_id in deployment_done %}
+      - {{ state_id }}
+      {%- endfor %}
+
+{#- Clearing the marker is the last step, and it is the only one that can fail on
+    a node that is otherwise deployed. Say so, rather than leaving the operator with
+    a bare API error on the very last state. The `require` matters: without it this
+    state also fires when the deployment itself failed, and would then claim a broken
+    node is fine. #}
+Explain the stale marker on {{ node_name }}:
+  test.configurable_test_state:
+    - name: {{ node_name }}
+    - changes: False
+    - result: True
+    - comment: >-
+        Node {{ node_name }} was deployed in {{ version }}, only the
+        metalk8s.scality.com/version-in-progress annotation could not be cleared.
+        The node itself is fine. Deploy it again to record the version, or record
+        it by hand with "kubectl annotate node {{ node_name }}
+        metalk8s.scality.com/version-in-progress-
+        metalk8s.scality.com/version-applied={{ version }} --overwrite". Until one
+        of those, an upgrade to another version is refused.
+    - onfail:
+      - metalk8s_kubernetes: Mark node {{ node_name }} as running {{ version }}
+    - require:
+      {%- for state_id in deployment_done %}
+      - {{ state_id }}
+      {%- endfor %}
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -127,6 +127,12 @@ Deploy node {{ node }}:
     - require_in:
       - salt: Downgrade etcd cluster
 
+{#- During a downgrade, `deploy_node` comes from the destination version, which is
+    older than this one. That older `deploy_node` does not write these annotations,
+    so the two states below write them here instead. The upgrade orchestrate has no
+    such need: it always runs the newer `deploy_node`. Delete these two states once
+    every version we can downgrade to writes the annotations itself, so from the next
+    minor on. #}
 Mark node {{ node }} as running {{ dest_version }}:
   metalk8s_kubernetes.object_updated:
     - name: {{ node }}
--- a/salt/metalk8s/orchestrate/downgrade/init.sls
+++ b/salt/metalk8s/orchestrate/downgrade/init.sls
@@ -14,16 +14,60 @@ Execute the downgrade prechecks:
 {%- set cp_nodes = salt.metalk8s.minions_by_role('master') | sort %}
 {%- set other_nodes = pillar.metalk8s.nodes.keys() | difference(cp_nodes) | sort %}
 
+{#- Nodes are deployed one by one, each waiting on the previous one. A skipped node
+    declares no state, so the chain must remember the last node actually deployed,
+    otherwise the next one waits on a state that does not exist and Salt fails it
+    with "The following requisites were not found". #}
+{%- set deployed = namespace(previous=None) %}
+
 {%- for node in other_nodes + cp_nodes %}
 
-  {%- set node_version = pillar.metalk8s.nodes[node].version|string %}
+  {#- The version label is set before a node is deployed, since it selects the
+      saltenv, so it says what the node was asked to run, not what it runs. The
+      `version-applied` annotation records the last version a node completed, and
+      that is what decides whether it still needs this one.
+      NOTE: this orchestrate runs from the destination saltenv, so the annotation is
+      only there when downgrading to a version that knows about it. Otherwise the
+      fallback below keeps the behaviour this orchestrate always had. #}
+  {%- set node_label = pillar.metalk8s.nodes[node].version|string %}
+  {%- set node_applied = pillar.metalk8s.nodes[node].get('version_applied') %}
+  {%- set node_in_progress = pillar.metalk8s.nodes[node].get('version_in_progress') %}
+  {#- Of a node that did finish its last deployment, keep the higher of the annotation
+      and the label. Every version of this orchestrate maintains the label, only the
+      ones that know about the annotation maintain it, so the two can disagree. A
+      downgrade must not skip a node that might still run the higher of the two, which
+      is the opposite of what the upgrade orchestrate keeps. #}
+  {%- if node_applied
+      and salt.pkg.version_cmp(node_applied|string, node_label) in (0, 1) %}
+    {%- set node_version = node_applied|string %}
+  {%- else %}
+    {%- set node_version = node_label %}
+  {%- endif %}
   {%- set version_cmp = salt.pkg.version_cmp(dest_version, node_version) %}
+  {#- A node that completed the destination is left alone, so resuming an
+      interrupted downgrade only works the nodes that need it. This leans on the
+      annotation: `node_version` only equals `node_applied` when that annotation
+      is the one we trust, and the label alone never proved that a node ran the
+      version it advertises (MK8S-370). #}
+  {%- set completed_dest = node_applied is not none
+                           and node_version == node_applied|string
+                           and node_version == dest_version %}
   {#- If dest_version = 2.1.0 and node_version = 2.1.0-dev, version_cmp = 0
       but we should not downgrade this node #}
-  {%- if version_cmp == 1
-      or (version_cmp == 0 and dest_version != node_version and '-' not in dest_version) %}
+  {#- A node still carrying the in-progress marker is another matter: its
+      deployment never finished, so neither its label nor its recorded version
+      says what it runs, and it gets deployed rather than trusted. #}
+  {%- if node_in_progress is none
+      and (version_cmp == 1
+           or completed_dest
+           or (version_cmp == 0 and dest_version != node_version and '-' not in dest_version)) %}
 
-Skip node {{ node }}, already in {{ node_version }} older than {{ dest_version }}:
+  {%- if completed_dest %}
+    {%- set skip_reason = "already completed " ~ dest_version %}
+  {%- else %}
+    {%- set skip_reason = "already in " ~ node_version ~ " older than " ~ dest_version %}
+  {%- endif %}
+Skip node {{ node }}, {{ skip_reason }}:
   test.succeed_without_changes
 
   {%- else %}
@@ -37,10 +81,13 @@ Wait for API server to be available on {{ node }}:
   - request_interval: 1
   - require:
     - salt: Execute the downgrade prechecks
-  {%- if loop.previtem is defined %}
-    - salt: Deploy node {{ loop.previtem }}
+  {%- if deployed.previous %}
+    - salt: Deploy node {{ deployed.previous }}
   {%- endif %}
 
+{#- The version label selects the saltenv used to deploy the node, so it has to
+    be set before the deployment. The annotation records that the node does not
+    run this version yet, and is removed once the deployment succeeded. #}
 Set node {{ node }} version to {{ dest_version }}:
   metalk8s_kubernetes.object_updated:
     - name: {{ node }}
@@ -50,6 +97,8 @@ Set node {{ node }} version to {{ dest_version }}:
         metadata:
           labels:
             metalk8s.scality.com/version: "{{ dest_version }}"
+          annotations:
+            metalk8s.scality.com/version-in-progress: "{{ dest_version }}"
     - require:
       - http: Wait for API server to be available on {{ node }}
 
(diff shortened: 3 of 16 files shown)


salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -17,11 +17,16 @@ Execute the upgrade prechecks:
...
+  {%- set version_in_progress = pillar.metalk8s.nodes[node].get('version_in_progress') %}
   {%- set version_cmp = salt.pkg.version_cmp(dest_version, node_version) %}
   {#- If dest_version = 2.1.0-dev and node_version = 2.1.0, version_cmp = 0
       but we should not upgrade this node #}
-  {%- if version_cmp == -1
-      or (version_cmp == 0 and dest_version != node_version and '-' not in node_version) %}
+  {#- A node whose deployment to `dest_version` did not complete advertises a
+      version it does not run yet, so never skip it. A marker left over from
+      another version is ignored here, it must not turn a skip into a deploy. #}
+  {%- if version_in_progress != dest_version
```
g-carre: This guard cannot change any outcome, so the version-label half of MK8S-370 has no effect. The reasoning is worth spelling out because it hinges on an invariant that lives in a different file.

**What the guard is trying to do.** A node whose deploy to `dest_version` was interrupted advertises a version it is not actually running, so it must never be skipped. The new conjunct says: if the in-progress marker equals the destination, fall through to the deploy branch.

**Why it never fires.** The marker and the label are written by a *single* patch, `upgrade/init.sls:85-87`:

```yaml
    - patch:
        metadata:
          labels:
            metalk8s.scality.com/version: "{{ dest_version }}"
          annotations:
            metalk8s.scality.com/version-in-progress: "{{ dest_version }}"
```

One `object_updated`, so both keys land together or neither does. That gives an invariant: **the annotation is either absent, or equal to the label.** (`downgrade/init.sls:60-62` is the same shape. The only other writer of that label is `mark-control-plane/files/bootstrap_node_update.yaml.j2.in:39`, which sets the label alone — see below.)

Now walk the condition with `version_in_progress == dest_version`. The invariant means `node_version == dest_version` too, so:

- `version_cmp = salt.pkg.version_cmp(dest_version, node_version)` = `version_cmp(V, V)` = **0**, so `version_cmp == -1` is False.
- the second branch needs `version_cmp == 0 and dest_version != node_version`, and `dest_version != node_version` is **False**.

So the parenthesised expression is False on its own, the whole `if` was *already* False, and the node was already taking the `else` (deploy) branch. Adding `version_in_progress != dest_version` in front of an expression that is already False cannot turn a skip into a deploy.

**The case that does bite is the one excluded on purpose.** The comment above says a marker from another version "must not turn a skip into a deploy" — but that is the only state in which the marker carries information the label does not:

1. cluster at 2.8.0, `upgrade 2.9.0` → patch sets label **and** marker to 2.9.0, then `Deploy node` fails. The node still runs 2.8.0.
2. operator retargets: `upgrade 2.8.5`.
3. `version_cmp(2.8.5, 2.9.0) == -1` → skip branch, and the marker `2.9.0 != 2.8.5` so the new conjunct passes it straight through.
4. The node is skipped, silently, and stays on **2.8.0** — indistinguishable from before this PR.

That is the mislabelled-node scenario MK8S-370 describes, and it is untouched.

**What would actually close it.** The marker is only useful if it can disagree with the label, which means either writing it *before* the label (two patches: marker at the start of the node's deploy, label only once the deploy returns), or making the skip decision consult the marker independently of `version_cmp` — e.g. skip only when the label matches the destination *and* no marker is present, regardless of which version the marker names. Both are bigger than a conjunct, which is presumably why this shape was reached for; but as written the code and its comment describe a behaviour it does not have.

**One knock-on for the tests.** The fixtures added for this in `salt/tests/unit/formulas/config.yaml` (around `:598-612` and `:917-931`) set `version: 2.7.1` alongside `version_in_progress: 2.7.3` — label and marker disagreeing, which the single-patch invariant above makes unreachable. And `test_formulas.py` only asserts that the template renders without raising; it never inspects the rendered output, so the `# the node is skipped as usual` comments in those fixtures are not actually checked by anything. Both together are why this reads as covered.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -146,16 +146,27 @@ Cordon the node:
...
   metalk8s_drain.node_drained:
     - name: {{ node_name }}
     - ignore_daemonset: True
     - ignore_pending: True
     - delete_local_data: True
     - force: True
-    {%- if pillar.orchestrate.get("drain_timeout") %}
-    - timeout: {{ pillar.orchestrate.drain_timeout }}
-    {%- endif %}
+    - timeout: {{ (drain_budget / drain_attempts) | int }}
```
g-carre: Dividing the user's `drain_timeout` by the attempt count changes what a documented option means, and two input ranges behave badly. Worth walking because the failure is in the interaction between this line and a default deep in the module.

**What changed.** `drain_timeout` used to be the budget for *the* drain. It is now the budget for *one attempt* of up to five (see the separate note on `attempts`). Same key, same docs, different unit.

**Range 1 — small-but-sane values silently become unachievable.** The repo's own render fixture uses 60 (`salt/tests/unit/formulas/config.yaml:490`). `60 / 4 | int` = **15s per attempt**. A drain that legitimately needs 40s of PodDisruptionBudget waiting — perfectly normal, since a 429 is answered by "sleep 5s and retry until the drain timeout" — can now never succeed on any attempt, no matter how many it gets. Total wall clock meanwhile grows to about 5×15s + 4×(30-40s) ≈ 200s, so the operator waits *longer* to be told it failed.

**Range 2 — values 1-3 invert into hours.** Jinja integer division truncates:

```
{{ (3 / 4) | int }}   ->   0
```

and a zero timeout is not zero, because `metalk8s_drain.py:208` treats it as "unset":

```python
# Default timeout is 3600 seconds
self._timeout = timeout or 3600
```

So `--drain-timeout 3` yields `timeout=0` → `0 or 3600` → **3600s per attempt**, five times over. A deliberately tiny drain timeout, the way you would express "do not wait around", becomes roughly five hours. The same `or` idiom is used one line up in this file (`| int or 3600`), which is correct there because the input really is absent — it is only unsafe once a truncating division can manufacture a 0.

**Suggestions**, in rough order of preference:

1. Leave `- timeout:` as the user's value and let the attempt count multiply the worst case, documenting the new worst case in the CHANGELOG. Simplest, and preserves the option's meaning.
2. Keep the division but floor it, e.g. `{{ [(drain_budget / drain_attempts) | int, 30] | max }}`, so it can never reach the `or 3600` trapdoor.
3. Derive the per-attempt value only when `drain_timeout` is unset, and honour an explicit value as-is — the option then keeps its documented meaning for anyone who sets it.

Either way, changing the interpretation of a user-facing option deserves a CHANGELOG line; the current entry describes the retry but not the change of unit.

salt/metalk8s/orchestrate/deploy_node.sls
```diff
@@ -146,16 +146,27 @@ Cordon the node:
...
+    - retry:
+        attempts: {{ drain_attempts }}
```
g-carre, line 185: `attempts: 4` runs this state **five** times, not four, so the budget arithmetic in the comment above is off by one attempt and the drain can take ~77 min where the comment promises ~60.

ezekiel-alexrod: Checked this one before answering, because the meaning of `attempts` differs between tools.

On the Salt this branch ships (3002.9), it's the total number of runs. `State.call` starts at `retries=1` (`salt/state.py:2027`) and re-runs only while `attempts > retries` (`state.py:2232`), so `attempts: 4` gives four executions.
Ran it rather than trusted the read:
```
salt-call --local state.single test.fail_without_changes name=drain-retry-check \
  retry='{"attempts": 4, "interval": 0}'

Attempt 1: Returned a result of "False", with the following comment: "Failure!"
Attempt 2: Returned a result of "False", with the following comment: "Failure!"
Attempt 3: Returned a result of "False", with the following comment: "Failure!"
Failure!
```

Three retries plus the first run. Same count on 3006.27, the Salt `development/134.0` ships, so the merge-up doesn't change it. The `while attempts >= retries` loop in 3006 (`state.py:2206`) is the parallel path, and this state isn't parallel.

Worst case is 4 × 900 s of draining plus three waits of 30 s with up to 10 s of splay, so about 62 min against the 60 min a single drain took before.

m4nch0t (dismissed):

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -98,6 +108,23 @@ Deploy node {{ node }}:
...
+    - apiVersion: v1
+    - patch:
+        metadata:
+          annotations:
+            metalk8s.scality.com/version-in-progress: null
+    {#- The node deployment just restarted the API server on this node, so do not
+        fail the whole upgrade on a transient error while clearing the marker #}
+    - retry:
+        attempts: 5
+        interval: 30
```
g-carre: We should have a clear message in the logs mentionning that API server did not respond in time to perfrom "clear the marker" action. We failed the upgrade but it could well be ok a few moments later.  "Retry the upgrade on this node with cmd xyz "

salt/metalk8s/orchestrate/upgrade/init.sls
g-carre: If I understand the cases: 

We can have a genuine failed upgrade from the Start of `Deploy node` to the end of `Mark node {{ node }} as running {{ dest_version }}` 

We could also have a false failure during `Mark node {{ node }} as running {{ dest_version }}:` if the API server fails to respond in the retry budget.

It seems this condition is covering both cases if the user retries to update with the same version as before. 

However there is a narrow faulty case IMHO if the user retries with an older version. 
For ex. 4.2.3 -> 4.3 fails we end with node_version = 4.3 ,  version_in_progress = 4.3 
Now the user retry with 4.2.4 and dest_version=4.2.4 version_cmp  =  -1  and we skip the upgrade. 

Do we want to explicitly limit the user to retry with exactly the same version or higher ones ? in which case we should be clear in the documentation about it. Or do we need to refine the and the test but in this case there may be a lot more to clear-up. 

After a second though other node may be upgraded ok and I believed that we  do not support downgrade so upgrade only it is and should be simpler. 

But I would like to discuss more why is orchestrate/downgrade present then ?

ezekiel-alexrod:
/changie Fix
A node drain no longer aborts the upgrade on a transient API server or etcd error.
An eviction rejected with a 5xx status, typically `etcdserver: request timed out`
while the upgrade restarts etcd and the API servers, is now retried for about a
minute per pod, and the drain state is retried too, its budget split across the
attempts so a drain that can never succeed still gives up in about the time it
used to take. A pod that keeps failing still fails the drain, naming the pod and
the API error.

ezekiel-alexrod:
/changie Fix
A node that fails during an upgrade or a downgrade no longer advertises the
destination version as if it were running it. The version label is still set
before the deployment, since it selects the saltenv, but the node now also
carries a `metalk8s.scality.com/version-in-progress` annotation until that
deployment succeeds, plus a `metalk8s.scality.com/version-applied` annotation
recording the last version it completed. Both orchestrates select nodes on that
recorded version rather than on the label, and the upgrade prechecks refuse a
destination other than the one an interrupted upgrade was heading to.

salt/metalk8s/orchestrate/upgrade/init.sls
ezekiel-alexrod: Right, and the narrow case was real. Fixed, but not as I first tried: comparing the marker to the destination was wrong one way, ignoring it was wrong the other, since an older `upgrade.sh` defaults to its own version and would roll a node back with none of the downgrade safeguards.

A node now records what it completed in `metalk8s.scality.com/version-applied`, written in the same patch that clears the marker, and both orchestrates compare the destination to that rather than to the label. Your example deploys the node: the label says 4.3, what it ran says 4.2.3. A node that really completed 4.3 is still skipped.

Your "same version only" option is enforced now, not just documented. While a node is marked, the prechecks refuse another destination and name it: `A version change to 133.0.14-dev did not complete on <node>, resume it rather than upgrading to 133.0.13`. Resuming redeploys every node, the highstate being idempotent.

On your last question, this project does support downgrade: `docs/operation/downgrade.rst` documents it with the same one-major-at-a-time limit as upgrade, and `scripts/downgrade.sh.in` ships it, even if not every consumer exposes it. So both orchestrates stay. One caveat: the downgrade runner comes from the destination saltenv, so going back below this change ignores the annotations.

salt/metalk8s/orchestrate/upgrade/init.sls
ezekiel-alexrod: Added:

```
Node <node> was upgraded to <version>, only the metalk8s.scality.com/version-in-progress
annotation could not be cleared. The node itself is fine. Run upgrade.sh again to redeploy
it, or record it by hand with "kubectl annotate node <node> ... --overwrite". Until one of
those, an upgrade to another version is refused.
```

One catch: it needs `require: Deploy node` next to the `onfail`. A state whose requisite failed counts as failed, so without it the message also fires when the deployment itself failed, and tells you a broken node is fine. Checked on a real Salt run, then on a cluster: the failed node stayed at `One or more requisite failed`, the healthy ones at `State was not run because onfail req did not change`.

The run still ends failed, on purpose, and the other nodes keep going.

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -14,14 +14,43 @@ Execute the upgrade prechecks:
...
+  {%- if node_applied
+      and salt.pkg.version_cmp(node_applied|string, node_label) in (-1, 0) %}
+    {%- set node_version = node_applied|string %}
+  {%- else %}
+    {%- set node_version = node_label %}
+  {%- endif %}
+  {#- A node still carrying the in-progress marker has a deployment that never
+      finished, so neither its label nor its recorded version says what it runs.
+      Deploy it rather than trust either. #}
+  {%- set never_skip = node_in_progress is not none %}
```
g-carre: using never skip is overkill and lead to a double negation line 51, it would be better to just do node_in_progress is none there. Same apply for downgrade.

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -14,14 +14,43 @@ Execute the upgrade prechecks:
...
   {%- set version_cmp = salt.pkg.version_cmp(dest_version, node_version) %}
   {#- If dest_version = 2.1.0-dev and node_version = 2.1.0, version_cmp = 0
       but we should not upgrade this node #}
-  {%- if version_cmp == -1
-      or (version_cmp == 0 and dest_version != node_version and '-' not in node_version) %}
+  {#- A node that already completed this very version is deployed again, the
+      highstate is idempotent and re-running is how a drifted node gets fixed. #}
+  {%- if not never_skip
```
g-carre:
```suggestion
  {%- if node_in_progress is none
```

salt/metalk8s/orchestrate/upgrade/init.sls
g-carre: LGTM

g-carre (dismissed):

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -14,14 +14,42 @@ Execute the upgrade prechecks:
...
-  {%- if version_cmp == -1
-      or (version_cmp == 0 and dest_version != node_version and '-' not in node_version) %}
+  {#- A node that already completed this very version is deployed again, the
+      highstate is idempotent and re-running is how a drifted node gets fixed.
+      A node still carrying the in-progress marker is another matter: its
+      deployment never finished, so neither its label nor its recorded version
+      says what it runs, and it gets deployed rather than trusted. #}
+  {%- if node_in_progress is none
+      and (version_cmp == -1
+           or (version_cmp == 0 and dest_version != node_version and '-' not in node_version)) %}
```
TeddyAndrieux, line 59: now that you added the version applied to me we could skip upgrade if we are already in applied version (it was not the case before because we got no way to be sure that previous upgrade completed but since you added an annotation for it I think it's good now)

docs/operation/upgrade.rst
```diff
@@ -67,3 +67,26 @@ Upgrade
...
+Nodes are upgraded one at a time, and each one carries the
+``metalk8s.scality.com/version-in-progress`` annotation for as long as its
+deployment runs. A node that fails mid-upgrade keeps that annotation, and
+the last version it completed stays recorded in
+``metalk8s.scality.com/version-applied``.
+
+To resume, run ``upgrade.sh`` again with the same version. Nodes that still
+carry the in-progress annotation are deployed again, and so are the nodes that
+already completed the destination version, since applying their highstate again
+is how a node that drifted gets fixed. For as long as a node carries the
```
TeddyAndrieux: > and so are the nodes that already completed the destination version, since applying their highstate again is how a node that drifted gets fixed.

IMHO it's not needed, config drift can be fixed by hand calling directly deploy_node orchestrate for each node if needed

salt/metalk8s/orchestrate/upgrade/init.sls
```diff
@@ -98,6 +131,48 @@ Deploy node {{ node }}:
...
+Mark node {{ node }} as running {{ dest_version }}:
+  metalk8s_kubernetes.object_updated:
+    - name: {{ node }}
+    - kind: Node
+    - apiVersion: v1
+    - patch:
+        metadata:
+          annotations:
+            metalk8s.scality.com/version-in-progress: null
+            metalk8s.scality.com/version-applied: "{{ dest_version }}"
```
TeddyAndrieux: I think the version-applied and in-progress annotations could be managed by the deploy_node orchestrate (so that it's set also on node deployment as well, and those annotations would actually always show the "running" version)

NOTE: For downgrade we have to keep it in the downgrade orchestrate for a minor since we rely on previous `deploy_node` orchestrate

salt/metalk8s/orchestrate/upgrade/init.sls
ezekiel-alexrod: Done in ddfd35487

docs/operation/upgrade.rst
ezekiel-alexrod: Done in d22f278ae

salt/metalk8s/orchestrate/upgrade/init.sls
ezekiel-alexrod: Done in 3b51ff661

g-carre (dismissed):

salt/metalk8s/orchestrate/downgrade/init.sls
```diff
@@ -78,6 +127,48 @@ Deploy node {{ node }}:
...
+        metalk8s.scality.com/version-in-progress annotation could not be cleared.
+        The node itself is fine. Run downgrade.sh again to redeploy it and record
+        the version, or record it by hand with "kubectl annotate node
+        {{ node }} metalk8s.scality.com/version-in-progress-
+        metalk8s.scality.com/version-applied={{ dest_version }} --overwrite". Until
+        one of those, an upgrade to another version is refused.
+    - onfail:
+      - metalk8s_kubernetes: Mark node {{ node }} as running {{ dest_version }}
+    - require:
+      - salt: Deploy node {{ node }}
```
TeddyAndrieux, line 175: Can we have a comment here mentionning that this part is only needed until next minor (since during downgrade we use target "deploy_node" orchestrate)

ezekiel-alexrod: Done in 4b70af5d2

TeddyAndrieux (approved):

g-carre (approved):

PR #5093, opened 2026-08-19
tests(ingress): wait for calico before restarting the Ingress DaemonSet (MK8S-331)

g-carre:
Refs: [MK8S-331](https://scality.atlassian.net/browse/MK8S-331)

## Problem

The Workload Plane Ingress VIP scenarios restart the `ingress-nginx-controller` DaemonSet without waiting for the `calico-node` rollout that `re_configure_portmap` has just triggered. That is a race, and losing it makes a VIP unreachable:

* `re_configure_portmap` rewrites the portmap `conditionsV4` of the calico CNI configuration (`salt/metalk8s/kubernetes/cni/calico/deployed.sls`), where `metalk8s_network.get_portmap_ips(as_cidr=True)` lists every VIP pool address as an explicit `/32`.
* That configuration reaches a node only when its `calico-node` Pod restarts and `install-cni` rewrites `/etc/cni/net.d/10-calico.conflist`.
* The hostPort DNAT rules of an `ingress-nginx-controller` Pod are programmed by the portmap CNI plugin **at sandbox creation**, baking in the conditions of the conflist present at that instant.

So an Ingress Pod recreated before its node's conflist is refreshed keeps DNAT rules built from the previous configuration, which do not cover the new VIPs. keepalived then assigns a VIP that the node has no rule to forward, and the request is refused while the DaemonSet is fully rolled, every Pod is Ready and calico is healthy — the shape this flake has always had.

Two of the four scenarios that reconfigure the portmap CIDRs already wait for the `calico-node` rollout before restarting the Ingress (`f2e18d79d`); those two have never flaked, and the two without the wait are the two that have.

## Fix

Wait for the `calico-node` rollout inside `re_configure_portmap`, immediately after the state that triggers it, rather than adding a step to the scenarios that lack one. It is the function that changes the CNI configuration, so a single wait covers all of its callers — the two scenarios, the `wp_ingress_vip_setup` Given step and both teardowns — and a newly added scenario cannot omit it.

## Diagnostics

`_ingress_debug` could not see the layer this failure lives in: a WP Ingress VIP is assigned by the keepalived Pods of the VirtualIPPools, which the operator creates in `metalk8s-vips` (`vipNamespaceName`, `operator/pkg/controller/clusterconfig/workloadplane/virtualippool.go`), while the dump covered only `metalk8s-ingress` and the `calico-node` DaemonSet. The sosreport does not fill the gap either — it is collected after the suite, once the pools are torn down and the transient Pods are gone.

Collected now on any ingress assertion failure:

* `metalk8s-vips` — `virtualippool`, `daemonset`, `pods -o wide`, events, and keepalived logs with timestamps.
* `controllerrevisions` with creation timestamps, and `rollout history`, so DaemonSet churn is explicit rather than inferred from Pod ages.
* `calico-node` Pods next to the existing DaemonSet line, and calico-filtered `kube-system` events.
* Events with `LAST:.lastTimestamp` instead of the relative `AGE` column, next to a `collection time (UTC)` header, so every section lands on one timeline.

And, for the IP that failed:

* The owner node, resolved with `get_node_ips()` — the salt `network.ip_addrs` call `check_vips_spreading` already uses. When no node holds the IP, every node's addresses are printed instead.
* That node's Pods in both `metalk8s-ingress` and `metalk8s-vips`.
* Probes on that node, through the salt-master: its addresses, a local request on the VIP, the CNI configuration files with their mtime, the calico conflist, the NAT rules naming the VIP, and the `CNI-HOSTPORT-DNAT` chain.

Every command is best-effort and wrapped, so a failing probe appends a section instead of masking the assertion, and the whole dump runs only after an assertion has failed — the negative `should not return` scenarios keep failing fast.

`perform_request` records `context["request_ip"]` so the single-IP steps get the per-VIP sections too; `ssh_config` is threaded through the request-assertion steps.

## Validation

Two nightlies on this branch, four multi-node job instances, no recurrence — `test_ingress.py` 14/14 in each. Three of the four instances failed before the fix. The post suite on `bootstrap-restore` goes from 5269s to ~5850s, the cost of the added `calico-node` waits.

## Checks

* `black` 22.8.0, the pinned pre-commit rev: unchanged.
* `pyflakes`: only the pre-existing `'re' imported but unused`, identical to the base commit.

Generated with [Claude Code](https://claude.com/claude-code)

[MK8S-331]: https://scality.atlassian.net/browse/MK8S-331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
+++ b/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
@@ -1,5 +1,5 @@
 kind: Enhanced
-body: Add debug information on ingress test failures and make the test more tolerant by waiting in calico readyness
+body: Add debug information on ingress test failures and make the test more tolerant by waiting for calico readiness
 time: 2026-08-20T10:03:11.4150515Z
 custom:
     CommentId: "5354398477"
--- a/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
+++ b/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
@@ -1,6 +1,0 @@
-kind: Enhanced
-body: Add debug information on ingress test failures and make the test more tolerant by waiting for calico readiness
-time: 2026-08-20T10:03:11.4150515Z
-custom:
-    CommentId: "5354398477"
-    Pull: "5093"
--- a/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
+++ b/.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
@@ -1,0 +1,6 @@
+kind: Enhanced
+body: Add debug information on ingress test failures and make the test more tolerant by waiting in calico readyness
+time: 2026-08-20T10:03:11.4150515Z
+custom:
+    CommentId: "5354398477"
+    Pull: "5093"
--- a/tests/post/steps/test_ingress.py
+++ b/tests/post/steps/test_ingress.py
@@ -194,11 +194,11 @@ def _vip_debug(host, ssh_config, vip):
         ),
     )
     for title, command in probes:
-        sections.append(
-            _section(
-                f"'{owner}': {title}", _salt_cmd_run(host, ssh_config, owner, command)
-            )
-        )
+        try:
+            output = _salt_cmd_run(host, ssh_config, owner, command)
+        except Exception as exc:  # pylint: disable=broad-except
+            output = f"failed to collect: {exc}"
+        sections.append(_section(f"'{owner}': {title}", output))
 
     return sections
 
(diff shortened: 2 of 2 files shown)


g-carre:
/changie test add debug information on ingress test failures and make the test more tolerant by waiting in calico readyness

g-carre:
/changie Enhanced Add debug information on ingress test failures and make the test more tolerant by waiting in calico readyness

.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Enhanced
+body: Add debug information on ingress test failures and make the test more tolerant by waiting in calico readyness
```
g-carre:
```suggestion
body: Add debug information on ingress test failures and make the test more tolerant by waiting for calico readiness
```

tests/post/steps/test_ingress.py
```diff
@@ -870,6 +1066,18 @@ def re_configure_portmap(host, version, ssh_config, context=None):
...
+    # configuration, which reaches a node only when its `calico-node` Pod
+    # restarts and its `install-cni` container rewrites
+    # `/etc/cni/net.d/10-calico.conflist`. The hostPort DNAT rules of an
+    # `ingress-nginx-controller` Pod are programmed by the portmap plugin at
+    # sandbox creation, baking in the conditions of the conflist present at
+    # that instant: a Pod recreated before its node caught up keeps DNAT rules
+    # that do not cover the new VIPs, and the VIP is then assigned by
+    # keepalived but refuses connections. Wait here rather than in each caller,
+    # so no scenario can restart the Ingress over a stale conflist.
+    wait_rollout_status(host, "daemonset/calico-node", "kube-system")
```
ezekiel-alexrod, line 1086: Worth saying in the comment why this wait has something to wait for: the DaemonSet carries a `checksum/config` annotation built from the ConfigMap (`cni/calico/deployed.sls:7076`), so rewriting the ConfigMap changes the pod template and starts a rollout. As it reads now, one could think that editing a ConfigMap rolls a DaemonSet by itself.

Other point: `ingress.feature:52` and `:81` now wait twice, their `rollout status` returns at once. Dropping both steps keeps the wait in one place, and the two other scenarios stop looking like they are missing a step.

tests/post/steps/test_ingress.py
```diff
@@ -19,50 +19,221 @@
...
+            "portmap DNAT chain",
+            "iptables -t nat -S CNI-HOSTPORT-DNAT || echo none",
         ),
     )
+    for title, command in probes:
+        sections.append(
+            _section(
+                f"'{owner}': {title}", _salt_cmd_run(host, ssh_config, owner, command)
+            )
+        )
```
ezekiel-alexrod: The only `try` is around the whole `_vip_debug` call, so the first probe that raises takes the following ones with it. Lose the addresses probe and you also lose the conflist and the NAT rules, which are the two sections this dump exists for.

```suggestion
    for title, command in probes:
        try:
            output = _salt_cmd_run(host, ssh_config, owner, command)
        except Exception as exc:  # pylint: disable=broad-except
            output = f"failed to collect: {exc}"
        sections.append(_section(f"'{owner}': {title}", output))
```

Applied locally, `py_compile` clean, longest line 68 chars so black leaves it alone.

tests/post/steps/test_ingress.py
g-carre: I would keep ingress.feature:52 and :81 untouched they are redundant but harmless and could be handy if we need to change again. I am expacting to whole test to disapear or at least be heavily refactored once we replace ngninx ingress controller.

tests/post/steps/test_ingress.py
g-carre: done

.changes/unreleased/Enhanced-20260820-100311-415051500.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Enhanced
+body: Add debug information on ingress test failures and make the test more tolerant by waiting for calico readiness
+time: 2026-08-20T10:03:11.4150515Z
+custom:
+    CommentId: "5354398477"
+    Pull: "5093"
```
TeddyAndrieux: To me we shouldn't have changelog entry for a CI test fix only

g-carre: removed

TeddyAndrieux (approved):

ezekiel-alexrod (approved):

PR #5096, opened 2026-08-20
improvement(salt): gate each master's apiserver upgrade on the previous one (MK8S-383)

g-carre:
Refs: [MK8S-383](https://scality.atlassian.net/browse/MK8S-383) — Item 2, and the `retry` sub-item inherited from MK8S-326.

Item 1 of the ticket is #5092. This is the remainder, kept separate so the `require`-chain rewiring is bisectable on its own.

## Problem

`orchestrate/apiserver.sls` serialises the masters through a `require` chain, but nothing checks that a node's apiserver is serving again once `Deploy apiserver <node>` returns. On a multi-master cluster the orchestrate can therefore start replacing the next master's apiserver while the previous one is still starting, reducing control plane availability more than intended.

Separately, `Sync <node> minion` and `Refresh <node> grains` are the only states in the file with no tolerance at all, so a transient "No minions responded" on either fails the upgrade outright.

## Changes

**A readiness gate per node**, with the next node chained on the gate rather than on the deploy — so the last node is gated too, and the orchestrate no longer reports success while the final apiserver is still starting.

**The gate queries the node's own apiserver on `:6443`, not `127.0.0.1:7443`.** This is the part worth reviewing: `apiserver-proxy.conf.j2` builds its upstream from *every* master with `least_conn`, weighting the local one 100 against 1 — preferred, not pinned. So a proxy-routed check fails over to another apiserver and reports ready while the node under upgrade is still down, which is exactly the case this gate exists to catch.

For the same reason **`metalk8s.wait_apiserver` is not the right tool here**, despite the ticket nominating it as Item 2's natural consumer:

* its liveness leg calls `metalk8s_kubernetes.ping`, which uses the salt-master kubeconfig, and `salt/metalk8s/salt/master/kubeconfig.sls:16` sets `apiserver: "https://127.0.0.1:7443"`;
* its RBAC leg defaults to `APISERVER_PROXY_URL = "https://127.0.0.1:7443"` (`salt/_modules/metalk8s.py:32`).

Both answer "is *an* apiserver ready?", which is right for its bootstrap and `deploy_node` gates and wrong for "is *this* node's apiserver serving?". Pointing `apiserver_url` at the node would fix only the RBAC leg. So the helper from MK8S-263 (#5022) still has no in-tree caller after this PR — worth recording, since the ticket expected otherwise.

The gate reuses the #5022 pattern otherwise: `match` and `status` together, so it requires overall readiness *and* the RBAC bootstrap hook. `wait_for` is stated rather than left to its 300s default — on the upgrade recorded in MK8S-383 the static pod was only swapped some 2.5 minutes into the deploy, so 420 matches the existing gates and leaves margin.

**Control plane IP from the runner.** Rendering happens on the master, where `mine.get` as an execution module returns nothing, so this uses `salt.saltutil.runner('mine.get', ...)` as `orchestrate/register_etcd.sls` does. A node missing from the mine raises at render time with an actionable message rather than failing obscurely later; `upgrade.sh` runs `flush_and_refresh_mine` two steps before `upgrade_apiservers`, so it should not normally trigger.

**`- retry: attempts: 5`** on `Sync <node> minion` and `Refresh <node> grains`, matching `Check pillar on <node>` just below.

## Verification

`tox -e unit-tests` pins python3.6 and cannot run locally, so I rendered the template directly with jinja2 against a mocked master context (3 masters, bootstrap first) and parsed the result as YAML. It produces 15 states in the expected order, with `Deploy apiserver master-1` requiring `http: Wait for apiserver bootstrap to be serving` and so on down the chain, and a gate after the last node. `apiserver.sls` already has `mode: master` cases in `salt/tests/unit/formulas/config.yaml` (single-node and compact), so CI covers the render; the raise branch is not covered, as the fixture's `mine.get` mock always returns a value.

One note on the mock: it keys its return on the raw `tgt` string, so the lookup is done per node rather than batched with `tgt_type: list` — which also matches the existing `register_etcd.sls` idiom exactly and needs no fixture change.

## Interaction with #5092

Both touch the `Deploy apiserver <node>` block — #5092 adds `- timeout: 300` above `- require:`, this changes the `require` target and appends a state after the block. They should merge cleanly in either order; if not, the resolution is to keep both.

## Not covered here

The nightly acceptance criterion on the ticket is a rate and needs several runs.

[MK8S-383]: https://scality.atlassian.net/browse/MK8S-383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changes/unreleased/Fix-20260820-155813-846585830.yaml
+++ b/.changes/unreleased/Fix-20260820-155813-846585830.yaml
@@ -1,12 +1,10 @@
 kind: Fix
 body: |-
-    Upgrading the apiservers now waits for each master's apiserver to be serving
-    again before replacing the next one. The orchestrate serialised the masters but
-    checked nothing after a node's deploy returned, so on a multi-master cluster it
-    could start replacing the next apiserver while the previous one was still
-    starting, reducing control plane availability more than intended. Transient
-    "No minions responded" on the grain sync and refresh steps is now retried
-    rather than failing the upgrade outright
+    Upgrading the apiservers now waits, from the salt-master, for each master's
+    apiserver to answer `/readyz` on its own control plane IP before replacing the
+    next one, on top of the local `/healthz` check the deploy already runs.
+    Transient "No minions responded" on the grain sync and refresh steps is now
+    retried rather than failing the upgrade outright
 time: 2026-08-20T15:58:13.84658583Z
 custom:
     CommentId: "5358426857"
--- a/.changes/unreleased/Fix-20260820-155813-846585830.yaml
+++ b/.changes/unreleased/Fix-20260820-155813-846585830.yaml
@@ -1,0 +1,13 @@
+kind: Fix
+body: |-
+    Upgrading the apiservers now waits for each master's apiserver to be serving
+    again before replacing the next one. The orchestrate serialised the masters but
+    checked nothing after a node's deploy returned, so on a multi-master cluster it
+    could start replacing the next apiserver while the previous one was still
+    starting, reducing control plane availability more than intended. Transient
+    "No minions responded" on the grain sync and refresh steps is now retried
+    rather than failing the upgrade outright
+time: 2026-08-20T15:58:13.84658583Z
+custom:
+    CommentId: "5358426857"
+    Pull: "5096"
--- a/salt/metalk8s/orchestrate/apiserver.sls
+++ b/salt/metalk8s/orchestrate/apiserver.sls
@@ -81,9 +81,9 @@ Wait for apiserver {{ node }} to be serving:
       only weights the local one, so it fails over to another apiserver and
       would report ready while this node is still starting. `match` and
       `status` are both checked, so this requires overall readiness and the
-      RBAC bootstrap hook, as the gates added by MK8S-263 do. `wait_for` is
-      stated because it otherwise defaults to 300s: on the upgrade recorded in
-      MK8S-383 the static pod was only swapped ~2.5min into the deploy #}
+      RBAC bootstrap hook.
+      `wait_for` is set to 420s instead of default 300s: on some upgrades
+      the static pod may be only swapped ~2.5min into the deploy #}
   - name: https://{{ node_ip }}:6443/readyz?verbose
   - match: 'poststarthook/rbac/bootstrap-roles ok'
   - status: 200
--- a/salt/metalk8s/orchestrate/apiserver.sls
+++ b/salt/metalk8s/orchestrate/apiserver.sls
@@ -77,20 +77,22 @@ Deploy apiserver {{ node }} to {{ dest_version }}:
 Wait for apiserver {{ node }} to be serving:
   http.wait_for_successful_query:
   {#- Query this node's apiserver directly rather than through
+  {#- Query this node's apiserver directly rather than through
       `127.0.0.1:7443`: the apiserver-proxy upstream lists every master and
       only weights the local one, so it fails over to another apiserver and
-      would report ready while this node is still starting. `match` and
-      `status` are both checked, so this requires overall readiness and the
-      RBAC bootstrap hook.
-      `wait_for` is set to 420s instead of default 300s: on some upgrades
-      the static pod may be only swapped ~2.5min into the deploy #}
+      would report ready while this node is still starting.
+      `/readyz` rather than `/healthz` for its `informer-sync` check: the
+      apiserver only reports ready once its caches are loaded, so the next
+      RBAC call does not race it. The deploy already waits for `/healthz`
+      from the node itself (`kubernetes/apiserver/installed.sls:179`); what
+      this gate adds is a check run from the salt-master, and one after the
+      last master too. #}
   - name: https://{{ node_ip }}:6443/readyz?verbose
   - match: 'poststarthook/rbac/bootstrap-roles ok'
   - status: 200
   - verify_ssl: true
   - ca_bundle: /etc/kubernetes/pki/ca.crt
   - request_interval: 5
-  - wait_for: 420
   - timeout: 10
   - require:
     - salt: Deploy apiserver {{ node }} to {{ dest_version }}
--- a/salt/metalk8s/orchestrate/apiserver.sls
+++ b/salt/metalk8s/orchestrate/apiserver.sls
@@ -10,18 +10,38 @@
 
 {%- for node in master_nodes %}
 
+{#- The readiness gate below queries the node's apiserver directly, so we need
+    its control plane IP. Rendering happens on the master, where `mine.get` as
+    an execution module returns nothing, hence the runner (as in
+    `orchestrate/register_etcd.sls`) #}
+{%- set node_ip = salt.saltutil.runner(
+        'mine.get', tgt=node, fun='control_plane_ip'
+    ).get(node)
+%}
+{%- if not node_ip %}
+{{ raise(
+       "No control plane IP for '" ~ node ~ "' in the Salt mine, needed to "
+       ~ "check its apiserver is serving again before moving on to the next "
+       ~ "master. Run `salt '" ~ node ~ "' mine.update` and retry."
+   ) }}
+{%- endif %}
+
 Sync {{ node }} minion:
   salt.function:
     - name: saltutil.sync_all
     - tgt: {{ node }}
     - kwarg:
         saltenv: metalk8s-{{ dest_version }}
+    - retry:
+        attempts: 5
 
 Refresh {{ node }} grains:
   salt.function:
     - name: saltutil.refresh_grains
     - tgt: {{ node }}
     - timeout: 120
+    - retry:
+        attempts: 5
     - require:
       - salt: Sync {{ node }} minion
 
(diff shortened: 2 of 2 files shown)


g-carre:
/changie Fix
Upgrading the apiservers now waits for each master's apiserver to be serving
again before replacing the next one. The orchestrate serialised the masters but
checked nothing after a node's deploy returned, so on a multi-master cluster it
could start replacing the next apiserver while the previous one was still
starting, reducing control plane availability more than intended. Transient
"No minions responded" on the grain sync and refresh steps is now retried
rather than failing the upgrade outright

salt/metalk8s/orchestrate/apiserver.sls
```diff
@@ -51,7 +71,27 @@ Deploy apiserver {{ node }} to {{ dest_version }}:
...
+Wait for apiserver {{ node }} to be serving:
+  http.wait_for_successful_query:
+  {#- Query this node's apiserver directly rather than through
+      `127.0.0.1:7443`: the apiserver-proxy upstream lists every master and
+      only weights the local one, so it fails over to another apiserver and
+      would report ready while this node is still starting. `match` and
+      `status` are both checked, so this requires overall readiness and the
+      RBAC bootstrap hook, as the gates added by MK8S-263 do. `wait_for` is
+      stated because it otherwise defaults to 300s: on the upgrade recorded in
+      MK8S-383 the static pod was only swapped ~2.5min into the deploy #}
```
g-carre:
```suggestion
      `127.0.0.1:7443`: the apiserver-proxy upstream lists every master and
      only weights the local one, so it fails over to another apiserver and
      would report ready while this node is still starting. `match` and
      `status` are both checked, so this requires overall readiness and the
      RBAC bootstrap hook.
      `wait_for` is set to 420s instead of default 300s: on some upgrades
      the static pod may be only swapped ~2.5min into the deploy #}
```

salt/metalk8s/orchestrate/apiserver.sls
```diff
@@ -51,7 +71,27 @@ Deploy apiserver {{ node }} to {{ dest_version }}:
...
+      `status` are both checked, so this requires overall readiness and the
+      RBAC bootstrap hook.
+      `wait_for` is set to 420s instead of default 300s: on some upgrades
+      the static pod may be only swapped ~2.5min into the deploy #}
+  - name: https://{{ node_ip }}:6443/readyz?verbose
+  - match: 'poststarthook/rbac/bootstrap-roles ok'
+  - status: 200
+  - verify_ssl: false
+  - request_interval: 5
+  - wait_for: 420
```
ezekiel-alexrod: MK8S-263's description settles the readyz versus healthz part: post-start hooks register in `/livez`, `/healthz` and `/readyz` alike, so `/healthz` already covers `poststarthook/rbac/bootstrap-roles`. And the deploy already polls `https://<control plane IP>:6443/healthz` from the node, unconditionally (`apiserver/installed.sls:179`). The only readyz delta is `informer-sync`, which protects RBAC-dependent callers, and what follows this gate is a `salt.state` authenticating as `system:masters`.

What the gate does add: the check is made from the salt-master rather than from the node, and the last node is gated, which the require chain never did. Worth saying that instead. `wait_for: 420` needs another basis too, since the deploy only returns once healthz went green within its own 300s default.

.changes/unreleased/Fix-20260820-155813-846585830.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Fix
+body: |-
+    Upgrading the apiservers now waits for each master's apiserver to be serving
+    again before replacing the next one. The orchestrate serialised the masters but
+    checked nothing after a node's deploy returned, so on a multi-master cluster it
+    could start replacing the next apiserver while the previous one was still
+    starting, reducing control plane availability more than intended. Transient
+    "No minions responded" on the grain sync and refresh steps is now retried
+    rather than failing the upgrade outright
```
ezekiel-alexrod: The deploy already waits for this node's `/healthz` on `:6443` (`apiserver/installed.sls:179`), so "checked nothing after a node's deploy returned" announces a bug that was not there. Wording that keeps the change without the claim:

```suggestion
    Upgrading the apiservers now waits, from the salt-master, for each master's
    apiserver to answer `/readyz` on its own control plane IP before replacing the
    next one, on top of the local `/healthz` check the deploy already runs.
    Transient "No minions responded" on the grain sync and refresh steps is now
    retried rather than failing the upgrade outright
```

salt/metalk8s/orchestrate/apiserver.sls
```diff
@@ -51,7 +71,27 @@ Deploy apiserver {{ node }} to {{ dest_version }}:
...
+      only weights the local one, so it fails over to another apiserver and
+      would report ready while this node is still starting. `match` and
+      `status` are both checked, so this requires overall readiness and the
+      RBAC bootstrap hook.
+      `wait_for` is set to 420s instead of default 300s: on some upgrades
+      the static pod may be only swapped ~2.5min into the deploy #}
+  - name: https://{{ node_ip }}:6443/readyz?verbose
+  - match: 'poststarthook/rbac/bootstrap-roles ok'
+  - status: 200
+  - verify_ssl: false
```
ezekiel-alexrod: Unlike the `127.0.0.1:7443` gates this was copied from, the query crosses the control plane network, so the cert is worth checking. `/etc/kubernetes/pki` is mounted read only in the salt-master container and the node's control plane IP is in the cert SANs (`apiserver/certs/server.sls:33`), which is what `installed.sls:180` already relies on.

```suggestion
  - verify_ssl: true
  - ca_bundle: /etc/kubernetes/pki/ca.crt
```

TeddyAndrieux (dismissed):
Maybe the check of `/readyz` could be done in `installed` salt state of apiserver, but I'm fine about doing it here in the orchestrate

salt/metalk8s/orchestrate/apiserver.sls
ezekiel-alexrod: Shorter version of what I meant, plus an easier way out.

On `/readyz` vs `/healthz`: both endpoints report the same post-start hooks, so `poststarthook/rbac/bootstrap-roles` is already in `/healthz`, and the deploy already waits for `/healthz` on that node on every run (`kubernetes/apiserver/installed.sls:179`). The one thing `/readyz` adds is `informer-sync`, meaning the apiserver has finished loading its caches and can answer RBAC calls right away. That is the reason worth writing in the comment, the rbac hook one does not hold.

Two things this gate does add, and the comment does not mention: the query runs from the salt-master instead of from the node, and the last master is waited on too, which the `require` chain never did.

About `wait_for: 420`: the deploy only returns once `/healthz` is green, within its own 300s default. So when this state starts, the apiserver has already answered ok and only `informer-sync` is left. The default is plenty.

```suggestion
  {#- Query this node's apiserver directly rather than through
      `127.0.0.1:7443`: the apiserver-proxy upstream lists every master and
      only weights the local one, so it fails over to another apiserver and
      would report ready while this node is still starting.
      `/readyz` rather than `/healthz` for its `informer-sync` check: the
      apiserver only reports ready once its caches are loaded, so the next
      RBAC call does not race it. The deploy already waits for `/healthz`
      from the node itself (`kubernetes/apiserver/installed.sls:179`); what
      this gate adds is a check run from the salt-master, and one after the
      last master too. #}
  - name: https://{{ node_ip }}:6443/readyz?verbose
  - match: 'poststarthook/rbac/bootstrap-roles ok'
  - status: 200
  - verify_ssl: true
  - ca_bundle: /etc/kubernetes/pki/ca.crt
  - request_interval: 5
```

Teddy's idea is a good one, and it is close to a one word change: at `installed.sls:180`, poll `/readyz` instead of `/healthz`. Plain `/readyz` answers `ok` like `/healthz` does, so the rest of that state stays as it is. Then this file loses the new `Wait for apiserver ... to be serving` state, the `mine.get` block at the top with its raise, and the `require` goes back to the previous node's deploy. Same guarantee, no dependency on the mine, and install, expansion and downgrade get it as well. The cost is testing: `installed.sls` runs on every master highstate, so that version needs an install and an expansion run behind it, not just an upgrade.

Your call, and no problem if you keep it here. Teddy approved and is fine either way, so take the suggestion above and ship it if you prefer.

salt/metalk8s/orchestrate/apiserver.sls
```diff
@@ -51,7 +71,30 @@ Deploy apiserver {{ node }} to {{ dest_version }}:
     - require:
       - salt: Check pillar on {{ node }}
   {%- if loop.previtem is defined %}
-      - salt: Deploy apiserver {{ loop.previtem }} to {{ dest_version }}
+      - http: Wait for apiserver {{ loop.previtem }} to be serving
   {%- endif %}
 
+Wait for apiserver {{ node }} to be serving:
+  http.wait_for_successful_query:
+  {#- Query this node's apiserver directly rather than through
```
g-carre:
```suggestion

```

TeddyAndrieux (approved):

ezekiel-alexrod (approved):

PR #5103, opened 2026-08-25
feat(buildchain): fetch containerd-image-preload from its release (MK8S-165)

ezekiel-alexrod:
**Component**: build

**Context**:

The image-cache repository publishes `containerd-image-preload` as an RPM, one per RedHat release, attached to each of its GitHub releases. This PR puts that package in the `scality` repository of the ISO, so a node can install it from the local repository like any other MetalK8s package. Installing it and enabling its timer is MK8S-395, not this PR.

The buildchain had no path for a finished RPM coming from outside a yum repository. It knows how to build one from a spec we own (`RPM_TO_BUILD`, as for `metalk8s-sosreport`), and how to let dnf download one from a repository we configure (`REDHAT_PACKAGES_TO_DOWNLOAD`). The URL support in `targets/package.py` fetches the sources of a SRPM, not a package.

**Summary**:

`PrebuiltRPMPackage` (`buildchain/buildchain/targets/package.py`) downloads one asset of a GitHub release, checks it against a pinned sha256, and writes it only if the digest matches. Three things it does beyond fetching bytes:

- its up-to-date check hashes the package on disk rather than remembering the pin, so an incremental build stays off the network, while a bumped pin, a truncated download or an edited package all trigger a new one
- the package is streamed next to its target and moved in one step, so a failed download leaves neither a file nor a directory behind, and nothing untrusted ever sits at the target path
- what a previous pin or a killed build left in the repository directory is removed, since the metadata indexes every RPM found there and `mkisofs` copies the directory as it is

`buildchain/buildchain/packaging.py` declares the package for RedHat 8 and 9 under a new `RPM_TO_FETCH` mapping, and a new `_fetch_packages` task group runs the downloads. A package lands in the repository it is declared under, and that repository's metadata task depends on it, so `createrepo` always runs after the download.

`buildchain/buildchain/versions.py` pins the release tag, the RPM version derived from it, and one digest per RedHat release. The tag is `v0.1.0-alpha.1`, the first release image-cache cut. GitHub renames an asset when its name holds a character it does not accept, so the RPM built as `0.1.0~alpha.1` is served as `0.1.0.alpha.1`: the download asks for that name, and the package keeps its own once on disk. The package is also declared as a `PackageVersion`, which is what `pkg_installed` reads to pin a version, and what keeps dnf from looking for the package in a repository that does not carry it. `BUMPING.md` gains a section listing the four values a bump has to move together.

**Acceptance criteria**:

- the `scality` repository of the ISO carries `containerd-image-preload` for RedHat 8 and 9
- the version and the digest are pinned in `versions.py`, so two builds of the same MetalK8s version ship the same package
- a failed or missing download stops the build with a message naming the release it expected
- `tox -e unit-tests-buildchain` covers the download, the digest check, the leftover cleanup, the failure messages and the wiring

**The ISO ships a prerelease package**:

image-cache cut `v0.1.0-alpha.1` on 2026-09-01, and that is what this PR pins. The package is an alpha, so the pin has to move to the GA tag once image-cache cuts it, which is one line in `versions.py` plus the two digests. `test_digests_are_pinned` fails on a bump that moves the tag and forgets a digest.

**Run locally**:

`tox -e unit-tests-buildchain`, 73 passed. Every pre-commit hook covering these files: black 26.3.1, pylint 4.0.5 at 10.00/10, mypy 1.20.2.

The download was exercised against the real release, for both RedHat releases. It fetches the two packages, their digests match the pins, the second run is skipped as up to date, and a package left by another pin next to the target sends the task back to the download, which removes it. The unit tests reach the failure paths over `file://`: a wrong digest, a missing asset, a broken connection, a pin that is not a digest, a partial file from a killed build. Not run: a full `./doit.sh`, which builds an ISO this machine has no toolchain for.

**Two things a reviewer should weigh**:

**The package is declared to Salt before anything installs it.** `versions.json` feeds `repo.packages`, and `metalk8s.repo.redhat` runs `check_pkg_availability` on every node, which is a `yum install --setopt tsflags=test` per declared package against the ISO repositories only. So this PR makes every node resolve `containerd-image-preload` and its dependencies at repo configuration time, before MK8S-395 installs anything. The RPM requires `containerd`, and the `containerd.io` package the ISO already ships provides it, checked on the repodata of the 133.0.13 ISO. A future version of the package that needs more has to declare it in `versions.PACKAGES`, otherwise the state fails cluster-wide, and the comment on `_rpm_package_containerd_image_preload` says so.

**`rpm_package_dir` exists because a package has to know where it lands before the repository object exists.** The repository needs the package paths for its metadata dependencies, and the package needs the repository directory, so one of the two has to derive the layout on its own. `RPMRepository.package_dir` now returns the same directory, and a test asserts both that the package lands there and that the directory sits under the repository root the metadata is built from.

**Out of scope**:

- installing the package and enabling the timer (MK8S-395)
- cutting the first image-cache release, and making the repository public (MK8S-163)
- the same leftover problem for the packages the buildchain builds itself: bumping `metalk8s-sosreport` leaves the previous RPM in the repository directory, and `RPMRepository.build_packages` does not clean it either. Untouched here, say the word if you want it fixed in this PR.
- folding the three URL fetches of the buildchain (this one, `salt_tree._download_ui_operator_crds`, `Package._get_sources`) into one helper with a timeout and a single error format. Worth doing, and it would touch code this ticket has no reason to move.
- declaring the fetched package's dependencies for the ISO download list. `RPMPackage` gets them from the spec it builds, which a finished RPM cannot offer, and declaring them by name would have dnf download an unpinned copy next to the pinned one. A test asserts the one dependency this package has is already shipped.

**Also in here**: four lines of pasted code sitting inside the module docstring of `targets/package.py`, deleted while the file was open. `coreutils.sha256_of` comes out of `sha256sum`, which held the third copy of the same chunked hashing loop.

---

See: MK8S-165

### DIFF
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -33,7 +33,6 @@ Overview;
 
 from pathlib import Path
 from typing import (
-    Callable,
     Dict,
     FrozenSet,
     Iterator,
@@ -373,53 +372,45 @@ def _rpm_repository_basename(releasever: str) -> str:
 
 
 def _prebuilt_rpm_package(
-    name: str,
-    releasever: str,
-    repo: str,
-    source: str,
-    tag: str,
-    digest: str,
-    arch: str,
+    prebuilt: versions.PrebuiltRPM, releasever: str
 ) -> targets.PrebuiltRPMPackage:
     """Return a prebuilt RPM package object.
 
     Arguments:
-        name:       package name
+        prebuilt:   declaration of the package, from `versions.py`
         releasever: RedHat release the package is built for
-        repo:       name of the repository the package lands in
-        source:     GitHub repository publishing the package
-        tag:        tag of the release carrying the package
-        digest:     expected SHA256 digest of the package
-        arch:       package architecture
     """
     try:
-        pkg_info = versions.REDHAT_PACKAGES_MAP[releasever][name]
+        digest = prebuilt.sha256[releasever]
     except KeyError as exc:
         raise ValueError(
-            f'Missing version for package "{name}" for release "{releasever}"'
+            f'Missing digest for package "{prebuilt.name}" '
+            f'for release "{releasever}"'
         ) from exc
 
+    pkg_info = prebuilt.package_version(releasever)
     if not pkg_info.full_version:
         raise ValueError(
-            f'Package "{name}" for release "{releasever}" needs a pinned version'
+            f'Package "{prebuilt.name}" for release "{releasever}" '
+            f"needs a pinned version"
         )
 
     return targets.PrebuiltRPMPackage(
         basename=f"_fetch_redhat_{releasever}_packages",
-        name=name,
+        name=prebuilt.name,
         full_version=pkg_info.full_version,
-        arch=arch,
-        repository=source,
-        tag=tag,
+        arch=prebuilt.arch,
+        repository=prebuilt.source,
+        tag=prebuilt.tag,
         digest=digest,
-        destination_dir=targets.rpm_package_dir(repo, releasever),
+        destination_dir=targets.rpm_package_dir(prebuilt.repo, releasever),
         # The package lands in a directory the repository creates. Waiting for
         # it also settles the order `doit clean` walks the two in: it cleans a
         # task before the tasks it depends on, so the package leaves before
         # anyone tries to remove the directory holding it.
         task_dep=[
             f"{_rpm_repository_basename(releasever)}:"
-            f"{repo}/{targets.MKDIR_ARCH_TASK_NAME}"
+            f"{prebuilt.repo}/{targets.MKDIR_ARCH_TASK_NAME}"
         ],
     )
 
@@ -473,55 +464,28 @@ RPM_TO_BUILD: Dict[str, Dict[str, Tuple[targets.RPMPackage, ...]]] = {
 }
 
 
-def _rpm_package_containerd_image_preload(
-    repo: str, releasever: str
-) -> targets.PrebuiltRPMPackage:
-    """`containerd-image-preload` RPM, from the image-cache release.
-
-    The package requires `containerd`, provided by the `containerd.io` package
-    the ISO already carries, so nothing else has to be downloaded for it. A
-    future version that needs more has to declare it in `versions.PACKAGES`:
-    the availability check the Salt states run on every node resolves the
-    dependencies of every declared package against the ISO repositories only.
-    """
-    try:
-        digest = versions.IMAGE_CACHE_RPM_SHA256[releasever]
-    except KeyError as exc:
-        raise ValueError(
-            f'Missing digest for the image-cache RPM for release "{releasever}"'
-        ) from exc
-
-    return _prebuilt_rpm_package(
-        name="containerd-image-preload",
-        releasever=releasever,
-        repo=repo,
-        source=versions.IMAGE_CACHE_REPOSITORY,
-        tag=versions.IMAGE_CACHE_TAG,
-        digest=digest,
-        arch="noarch",
-    )
-
-
 def _packages_to_fetch(
-    repo: str, *factories: Callable[[str, str], targets.PrebuiltRPMPackage]
+    prebuilts: Sequence[versions.PrebuiltRPM],
 ) -> Dict[str, Dict[str, Tuple[targets.PrebuiltRPMPackage, ...]]]:
-    """The packages one repository fetches, one entry per RedHat release.
+    """The prebuilt packages, grouped the way the repositories read them.
 
-    The repository is named once here: it is both the key the packages are
-    declared under and the directory they land in, and a package that lands
-    somewhere else than where it is declared never reaches the ISO.
+    A package is declared under the repository it lands in, since that is
+    where the repository metadata has to find it, and both come from the
+    `repo` of the declaration.
     """
-    return {
-        repo: {
-            releasever: tuple(factory(repo, releasever) for factory in factories)
-            for releasever in versions.REDHAT_PACKAGES
-        }
-    }
+    grouped: Dict[str, Dict[str, Tuple[targets.PrebuiltRPMPackage, ...]]] = {}
+    for prebuilt in prebuilts:
+        per_release = grouped.setdefault(prebuilt.repo, {})
+        for releasever in versions.REDHAT_PACKAGES:
+            per_release[releasever] = per_release.get(releasever, ()) + (
+                _prebuilt_rpm_package(prebuilt, releasever),
+            )
+    return grouped
 
 
 # Packages to fetch from a release instead of building them, per repository.
 RPM_TO_FETCH: Dict[str, Dict[str, Tuple[targets.PrebuiltRPMPackage, ...]]] = (
-    _packages_to_fetch("scality", _rpm_package_containerd_image_preload)
+    _packages_to_fetch(versions.PREBUILT_RPMS)
 )
 
 
--- a/buildchain/buildchain/packaging.py
+++ b/buildchain/buildchain/packaging.py
@@ -6,6 +6,7 @@
 This modules provides several services:
 - build a unique container image for all the build tasks
 - downloading packages and repositories
+- downloading prebuilt packages attached to a GitHub release
 - building local packages from sources
 - building local repositories from local packages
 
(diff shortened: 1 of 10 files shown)


ezekiel-alexrod:
/changie Added The ISO now carries the containerd-image-preload package in its scality repository, for RedHat 8 and 9.

buildchain/buildchain/packaging.py
```diff
@@ -358,15 +472,57 @@ def _rpm_package_metalk8s_sosreport(releasever: str) -> targets.RPMPackage:
 }
 
 
-_RPM_TO_BUILD_PKG_NAMES: Dict[str, List[str]] = _list_packages_to_build(RPM_TO_BUILD)
+def _rpm_package_containerd_image_preload(
```
g-carre: RPM _TO_FETCH is based on same logic than RPM_TO_BUILD but this seems overly complex to read and since we do not have to deal with sources to build we could have a more basic "flat" table like : 

```# versions.py
class PrebuiltRPM(NamedTuple):
    name: str
    repo: str                 # ISO repository it lands in
    source: str               # GitHub repository publishing it
    tag: str
    release: str
    arch: str
    sha256: Dict[str, str]    # per releasever

    @property
    def version(self) -> str:
        return self.tag.removeprefix("v").replace("-", "~")

PREBUILT_RPMS = (
    PrebuiltRPM(
        name="containerd-image-preload",
        repo="scality",
        source="scality/image-cache",
        tag="v0.1.0-alpha.1",
        release="1",
        arch="noarch",
        sha256={"8": "b374b74d...", "9": "bc205c21..."},
    ),
)
```
then here  something that create RPM_TO_FETCH without having to rely on this. That being said the suggestion is not that simple to implement in the current code stucture. Maybe keeping a look alike consitency is better./

The other draw back I see right now is  I do not understand why we need to define "scality" in  releasever: (_rpm_package_containerd_image_preload("scality", releasever) it should be self defined in version

ezekiel-alexrod: Fixed the double `"scality"`, you're right it was noise. `_packages_to_fetch` takes the name once now. The `repo` parameter on the factory stays: a `PrebuiltRPMPackage` needs its `destination_dir` at construction, where a built one gets its path from the repository that owns it.

I implemented the flat table too, last commit. The win isn't readability. It's that the two `PackageVersion` literals in `versions.py` restated what the `IMAGE_CACHE_*` constants already said, so a bump had to move both and nothing caught a half-done one. The pins are generated from the tag now, and the rendered package list came out identical, I diffed it.

The nested shape stays though. `_list_package_names` is generic over both tables (packaging.py:516), and the repositories read `RPM_TO_FETCH["scality"]["8"]` directly (packaging.py:543). It just gets built from one declaration instead of several.

Say the word if you'd rather have that commit in its own ticket.

buildchain/buildchain/versions.py
```diff
@@ -450,23 +529,25 @@ def rpm_full_name(self) -> str:
...
         PackageVersion(name="xfsprogs"),
     ),
-    "redhat": {
-        "8": (
-            PackageVersion(name="iptables-ebtables", override="iptables-nft"),
-            PackageVersion(
-                name="metalk8s-sosreport",
-                version=NONSUFFIXED_VERSION,
-                release=f"{SOSREPORT_RELEASE}.el8",
+    "redhat": _with_prebuilt_rpms(
```
g-carre: _with_prebuilt_rpms is not very human friendly IMHO.

Can we just do something like: 
PACKAGES: Dict[str, Any] = { ... } as before. 

Then add after  
```
for version in PACKAGES["redhat"].keys() 
    PACKAGES["redhat"][version]+=tuple(prebuilt.package_version(version) for prebuilt in PREBUILT_RPMS)
```

ezekiel-alexrod: Done, the helper is gone and it's a loop right after `PACKAGES` now.

One catch: the loop variable can't be `version` or `releasever`, both shadow something else in the file and pylint fails on it. It's `rh_release`.

buildchain/buildchain/versions.py
```diff
@@ -413,6 +413,69 @@ def rpm_full_name(self) -> str:
...
+    # package the ISO already carries, so nothing else has to be downloaded
+    # for it. A future version that needs more has to be declared in
+    # `PACKAGES`: the availability check the Salt states run on every node
+    # resolves the dependencies of every declared package against the ISO
+    # repositories alone.
+    PrebuiltRPM(
+        name="containerd-image-preload",
+        repo="scality",
+        source="scality/image-cache",
+        tag="v0.1.0-alpha.1",
```
TeddyAndrieux, line 475: We will want to switch to a "real" release before doing an actual release of 134 but fine about alpha for now

buildchain/buildchain/versions.py
```diff
@@ -413,6 +413,69 @@ def rpm_full_name(self) -> str:
...
+        release="1",
```
TeddyAndrieux: TBH release=1 could be a default especially for our usage I think we will never change this and only bump the version (tag)

buildchain/buildchain/versions.py
```diff
@@ -413,6 +413,69 @@ def rpm_full_name(self) -> str:
...
+        arch="noarch",
```
TeddyAndrieux: ditto noarch could be a default

buildchain/buildchain/targets/package.py
```diff
@@ -339,6 +345,224 @@ def _get_buildsrpm_mounts(self, srpm_dir: Path) -> List[types.Mount]:
...
 
+class PrebuiltRPMPackage(base.AtomicTarget):
+    """A RPM built elsewhere and attached to a GitHub release.
+
+    Nothing is built here, unlike `RPMPackage`: the release carries one RPM per
+    RedHat release, and we only download the one we pinned and check it against
+    its expected digest before it lands in a repository.
+    """
+
+    GITHUB_RELEASES_URL = "https://github.com"
```
TeddyAndrieux, line 362: Ok so PrebuiltRPMPackage is github release specific, fine for now but maybe it the future we will want to support other place to download RPM (no need to do anything in this PR about it)

buildchain/buildchain/versions.py
```diff
@@ -413,6 +413,69 @@ def rpm_full_name(self) -> str:
...
+        )
+
+
+PREBUILT_RPMS: Tuple[PrebuiltRPM, ...] = (
+    # The package requires `containerd`, provided by the `containerd.io`
+    # package the ISO already carries, so nothing else has to be downloaded
+    # for it. A future version that needs more has to be declared in
+    # `PACKAGES`: the availability check the Salt states run on every node
+    # resolves the dependencies of every declared package against the ISO
+    # repositories alone.
```
TeddyAndrieux, line 464: To me we could automatically add dependencies like we do for package we build

I'm fine about no doing it in this PR since the list of dependencies is really small (only containerd) but if we start having more "prebuilt rpms" I think it would be valuable

TeddyAndrieux (dismissed):
A lot of logic "just" to download a RPM but it's future proof and allow for adding new prebuilt RPM in the future so LGTM

Just some minor comments

buildchain/buildchain/versions.py
ezekiel-alexrod: Agreed, an alpha can't ship in a real 134. I'll open a follow-up so the pin doesn't get forgotten between now and then.

buildchain/buildchain/versions.py
ezekiel-alexrod: Done, `release` defaults to `1`.

buildchain/buildchain/versions.py
ezekiel-alexrod: Done too. The declaration is down to name, repo, source, tag and the digests now.

buildchain/buildchain/targets/package.py
ezekiel-alexrod: Right, and I put that in the class docstring so it's not lost: what's tied to GitHub is the URL and the name the release serves the asset under. The digest check, the atomic write, the up-to-date check and the cleanup would be shared with another source.

buildchain/buildchain/versions.py
ezekiel-alexrod: Left a note above the declaration. Both ways cost something: by name, `dnf` pulls an unpinned copy, and reading them off the RPM makes the download list depend on a file we haven't fetched yet. Not worth it for one dependency that's already on the ISO, worth it on the second prebuilt package.

buildchain/buildchain/versions.py
ezekiel-alexrod: Opened it: https://scality.atlassian.net/browse/MK8S-408

buildchain/buildchain/versions.py
ezekiel-alexrod: Tracked in https://scality.atlassian.net/browse/MK8S-409, with the two options and what each one costs.

TeddyAndrieux (approved):

g-carre (approved):
Good

PR #5106, opened 2026-08-27
UI: keep every page usable in a 768px content box (ARTESCA-17790)

JeanMarcMilletScality:
**TL;DR** — Opening the shell's side drawer leaves the UI a 768px content box, which it used to clip and overflow rather than adapt to; the dashboard, the tables, the forms and the detail panels now reflow to the width they actually get.

Before/After screenshots at 712px:
Alerts page:
BEFORE:
image

AFTER:
image

Dashboard page:
BEFORE:
image

AFTER:
image

Node page
BEFORE:
image

AFTER:
image

volume page
BEFORE:
image

AFTER:

image

### Context / Why

Part of ARTESCA-17790: keeping this UI usable while it is embedded beside a 500px side drawer that the hosting shell can open. core-ui has already shipped the responsive abilities this needs, so this PR is `ui` + `shell-ui` adopting them and fixing the layouts no component prop can reach.

### Approach

**Why none of this is a media query.**

```
browser window — window.innerWidth stays 1268px throughout
┌────────────────────────────────────────────┬──────────────────┐
│   shell content column          768px      │   side drawer    │
│   ┌──────────────────────────────────────┐ │      500px       │
│   │   the hosted UI renders in here      │ │                  │
│   └──────────────────────────────────────┘ │                  │
└────────────────────────────────────────────┴──────────────────┘
```

The drawer is a flex *sibling* of the content column, not an overlay, so it takes width away from the app without the viewport ever changing. `@media (max-width: 768px)` therefore never fires in the one situation this PR exists for. Every query added here is `@container responsive (max-width: …)`, and the two `TwoPanelLayout`s get `container="both"` so the queries inside their panels have a container to resolve against — without it they are inert, which is why the node and volume panels looked untouched by earlier work.

**Two defects that were one bug.**

```tsx
// resolves through the Module Federation share scope → the host's copy of core-ui
import { Form } from '@scality/core-ui';

// a subpath is not shared → this app's own copy
import { Input } from '@scality/core-ui/dist/next';
```

Both rspack configs spread every dependency into the `shared` scope, so the bare specifier resolves to whichever copy the share scope hands out, while any subpath (`dist/index`, `dist/next`) resolves to the app's own. A `Form` reached one way and an `Input` reached the other are two different module instances, and anything held in module state stops connecting between them. It had already broken twice, silently and with no failing test:

| What a user saw | The module state that split |
| --- | --- |
| Scrollbars lost the core-ui skin after visiting any tab containing a `Select` | `createGlobalStyle` keys its rule group by a hash of the CSS and rebuilds it from the unmounting copy's own instances, so unmounting a `Select` wiped rules the other copy had inserted |
| `Form responsive` had no effect on the fields — labels shrank, inputs did not | `Form` passes `responsive`, `error` and `disabled` down through a React context, and `createContext` runs once per copy, so fields imported from `dist/next` never saw the Form's |

The fix in both cases is the import site, not the component: anything that has to talk to a sibling comes from the same entry point. The `dist/next` surface is only reachable by subpath, so for these files that entry point is `dist/index`.

### Review focus

- 🟡 `ui/src/FederableApp.tsx` › `ScrollbarWrapper` import, plus the same `dist/index` switch in `CreateVolume.tsx`, `NodeCreateForm.tsx` and `ConfigureAlerting.tsx` — these look like gratuitous import churn and are load-bearing. Reverting any one of them to the bare specifier silently reintroduces a row of the table above; nothing errors and no test fails.
- 🟡 `ui/src/containers/AlertPage.tsx` › `HeaderSegment` — the three header segments were fixed 250px boxes, each followed by a 250px absolutely-positioned `SeperationLine` whose right border happened to land on the segment's edge. They are now equal flex segments and the divider comes from `Stack`'s own `withSeparators`, so it is a 1px rule on the theme's `border` token where it used to be 2px of `backgroundLevel1` — the change in weight is the thing to eyeball.
- ⚪ `ui/src/components/style/CommonLayoutStyle.ts` › `ActiveAlertWrapper` — shared by the volume *and* the node overview. It is now `box-sizing: border-box`, so it renders 200px wide instead of 236px and the two counters sit ~36px closer together on the node page too.

### How to test

1. Open the UI in a ~1268px browser window and open the shell's side drawer. Everything below is at the 768px content box — the browser window is never resized.
2. **Dashboard** — the five-column row restacks twice on the way down (inventory beside the two chart panels, then all three stacked). No horizontal scrollbar at any point, and each panel scrolls within itself instead of the page scrolling as a whole.
3. **Nodes** and **Volumes** — columns drop as the list panel narrows and can be brought back from the table's own reveal control; the create button collapses to its icon. Open a volume and stay on Overview: the name, the labels and the Active alerts box all stay inside the panel, and the title holds its place while the card scrolls under it.
4. **Nodes → a node → Metrics tab → Alerts tab → back to Metrics** — the scrollbars keep the core-ui skin. Before, they reverted to the browser's default on the way back.
5. **Create Volume**, **Create Node**, **Configure Alerting** — labels *and* inputs shrink with the form. On any volume list, a volume that was never mounted shows `N/A` in the Usage column instead of a half-full bar labelled `undefined%`.

### Follow-up

- The dashboard's header banner still overflows at 768px. It needs a design decision rather than a layout fix, so it is deliberately left alone here.
- The notification popover keeps a fixed `25rem` width. Nothing between the content column and a navbar action declares a query container, so it has no way to bound itself by the column: `max-width: 100%` would resolve against the bell button, and viewport units measure the box the drawer does not change. It clears the 768px content box with ~400px to spare, so this binds only on a much wider drawer.

### References

- [ARTESCA-17790](https://scality.atlassian.net/browse/ARTESCA-17790) — epic: making the UI responsive to the content box it is given.

<details>
<summary>What changed</summary>

**Dependency** — `@scality/core-ui` 0.227.0 → 0.230.0 in both `ui` and `shell-ui`.

**shell-ui** — the notification center was positioned by arithmetic on `window.innerWidth`, which put it underneath the drawer as soon as the drawer opened; it is now anchored to the bell's right edge and needs no measurement. Its rows ellipsize instead of pushing the unread mark out.

**Tables** (nodes, volumes, pods, partitions, alerts) — every column gets a `minWidth` floor, the columns that can be spared get a `dropAt` width in the order they can be spared, tables get `revealDroppedColumns`, and the primary action buttons get `iconOnly`.

**Dashboard** — declares its own query container (it sits outside `TwoPanelLayout`, so nothing above it opted in), restacks in two steps, and has a single scroll owner instead of up to three nested scrollbars. The Grafana link moved out of the Metrics panel into the page's context bar next to the timespan selector, since both controls act on both panels.

**Detail panels** — overview rows stack at one shared width rather than each wrapping on its own content; the volume overview's title is pinned while its card scrolls, and its labels, title row and alerts box are all allowed to shrink.

**Usage column** — `ProgressBar`'s `percentage` defaults to 50 when it is not a number, so an unknown usage rendered as a half-full bar labelled `undefined%`, indistinguishable from a real reading. It is now an empty bar labelled `N/A`, in a `UsageProgressBar` shared by all four tables and covered by its own test.

**Deliberately out of scope** — the dashboard header banner (see Follow-up).

</details>

[ARTESCA-17790]: https://scality.atlassian.net/browse/ARTESCA-17790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell-ui/src/navbar/NotificationCenter.tsx
+++ b/shell-ui/src/navbar/NotificationCenter.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 
 import { useSelect } from 'downshift';
 import styled, { useTheme } from 'styled-components';
@@ -14,26 +14,16 @@ import { SleepingNotificationBell } from './SleepingNotificationBell';
 import { useNotificationCenter } from '../useNotificationCenter';
 import { ConstrainedText } from '@scality/core-ui/dist/components/constrainedtext/Constrainedtext.component';
 
-const NotificationMenu = styled.ul<{
-  $buttonBoundingRect: DOMRect;
-}>`
+/* Anchored to the bell's right edge rather than positioned against the viewport.
+   The navbar lives inside the flex-1 content column, which the Guardian drawer
+   narrows without changing window.innerWidth — so any viewport-derived offset
+   lets this menu extend underneath the drawer. Anchoring right keeps it inside
+   the column at any width, and needs no measurement to stay correct when the
+   drawer opens (which resizes the column live, without a reload). */
+const NotificationMenu = styled.ul`
   position: absolute;
-  width: 25vw;
-  /* TO MAKE SURE THE LIST IS CENTERED ON THE BELL BUTTON */
-  left: ${(props) => {
-    const notificationCenterWidth = 0.25 * window.innerWidth;
-    const leftRelativeToButton = -notificationCenterWidth / 2 + props.$buttonBoundingRect.width / 2;
-    const absoluteNotificationCenterX = props.$buttonBoundingRect.x + leftRelativeToButton;
-    if (absoluteNotificationCenterX + notificationCenterWidth > window.innerWidth) {
-      // |<------------------window.innerWidth------------------------------------>|<--offset-->|
-      // |<------------------button x------------------------------>|buttonwidth|---------------|
-      // |---------------------------------|<-LeftRelativeToButton->|---------------------------|
-      // |<--absoluteNotificationCenterX-->|<--notificationCenterWidth------------------------->|
-      const offset = absoluteNotificationCenterX + notificationCenterWidth - window.innerWidth;
-      return leftRelativeToButton - offset;
-    }
-    return leftRelativeToButton;
-  }}px;
+  right: 0;
+  width: 25rem;
   overflow: auto;
   list-style: none;
   padding: 0;
@@ -74,7 +64,6 @@ const NotificationCenterHeader = styled.div`
 
 const NotificationCenter = () => {
   const { notifications, readAllNotifications } = useNotificationCenter();
-  const [buttonBoundingRect, setButtonBoundingRect] = useState<DOMRect>(new DOMRect());
   const theme = useTheme();
   const { isOpen, getToggleButtonProps, getMenuProps, highlightedIndex, getItemProps } = useSelect({
     items: notifications,
@@ -98,11 +87,6 @@ const NotificationCenter = () => {
       style={{ position: 'relative'}}
       {...getToggleButtonProps({
         'aria-label': 'Notification Center',
-        ref: (e: HTMLDivElement) => {
-          if (e && (buttonBoundingRect.x === 0 || buttonBoundingRect.width === 0)) {
-            setButtonBoundingRect(e.getBoundingClientRect());
-          }
-        },
       })}
     >
       <Button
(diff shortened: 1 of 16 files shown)


JeanMarcMilletScality:
/changie Enhanced The UI pages now adapt to the width of the area they are given, and stay usable when that area is narrowed.

JBWatenbergScality (approved):

MonPote (approved):

PR #5109, opened 2026-09-01
feat(salt): install and enable the containerd image preload timer (MK8S-395)

ezekiel-alexrod:
**Component**: salt, tests

**Context**:

MK8S-165 (#5103) puts the `containerd-image-preload` RPM in the `scality` repository of the ISO. This PR is the salt side: install it on every node and start importing.

The package ships its systemd units disabled. That is deliberate, the spec says enabling the timer belongs to the provisioning layer, so nothing imports anything until we enable it here.

**Summary**:

A new `metalk8s.image-cache` formula installs the package with `pkg_installed`, like every other package MetalK8s owns, and runs `containerd-image-preload.timer` with `enable: True`. The `node` and `bootstrap` roles both include it, which covers every node that carries a container engine.

Two requisites carry the ordering the units cannot express on their own. The package state requires `Install containerd`, since the RPM requires `containerd` and the `containerd.io` package we ship is what provides that name. The timer state requires `Ensure containerd is ready`, because the timer fires as soon as it starts (`OnBootSec=0`) and the import goes into a running containerd. A node with no containerd renders a no-op rather than pulling the engine in.

The two new files are declared in `buildchain/buildchain/salt_tree.py`, otherwise they never reach the ISO.

Tests come in two layers. A formula test reads the requisites back from the rendered state, so a future edit that drops one fails the build rather than the cluster. It also covers the no-containerd branch, and fails if no rendering case exercises the timer at all. Two sanity scenarios then check the real thing on a live cluster: a package installed on every node, and a systemd unit enabled and running on every node. Both target the Kubernetes nodes by name, so a silent minion fails the check instead of dropping out of the results.

**Acceptance criteria**:

- every node in a fresh cluster has the package installed, the timer enabled and running
- an archive dropped in `/var/lib/image-cache` is imported into containerd's `k8s.io` namespace on the next tick
- removing an image with `crictl rmi` and waiting one tick brings it back
- the package and its timer are covered by the deployment sanity checks
- unit tests on the requisites

**This PR has to merge after #5103**:

The package only exists in the local repository once #5103 lands. Merged before it, `Install containerd image preload` fails with "No package containerd-image-preload available" and takes the whole highstate down. `pkg_installed` also reads `repo.packages` for the version to pin, and that entry comes from `versions.py` in #5103, so merged early the package would install unpinned.

**Run locally**:

311 formula rendering tests, 32 buildchain tests, and every pre-commit hook covering these files: black 26.3.1, pylint 4.0.5 at 10.00/10, mypy 1.20.2, salt-lint 0.9.2, yamllint 1.38.0.

`Provides: containerd` was checked against the RPM header of `containerd.io-2.2.2-1.el8` as built here, which lists `containerd`, `containerd.io` and `containerd.io(x86-64)`. The same check was made on the 133.0.13 repodata in #5103.

Not run: the sanity scenarios and the acceptance criteria that need a cluster, and a full ISO build.

**Three things a reviewer should weigh**:

**The formula includes what it depends on, even where the include is transitive.** `metalk8s.repo` and `metalk8s.container-engine.containerd.running` both come in through `containerd.installed`, so listing them adds nothing to the compiled highstate today. They are there so that a cleanup of the containerd includes breaks a test rather than silently unresolving `Repositories configured` or `Ensure containerd is ready`. Happy to drop them if you would rather keep the include list minimal.

**The bootstrap role includes the formula although it also reaches it through `master` and then `node`.** The same duplication already exists for `metalk8s.kubernetes.kubelet`, and the bootstrap node keeping the preload does not depend on which other roles it carries. It does mean a future change to which roles get the preload has to touch two files.

**The guard tests `container_engine == 'containerd'` rather than dispatching by name.** The preload script only knows `ctr`, so there is nothing to render for another engine. The other formulas dispatch with `.{{ container_engine }}`, which does not fit a formula that has one implementation.

**Out of scope**:

- filling the cache directory, covered by MK8S-157, MK8S-394 and MK8S-166
- getting the RPM into the ISO, covered by MK8S-165
- the cache directory and the platform, which come from `/etc/sysconfig/containerd-image-preload` as `%config(noreplace)`. The defaults are `/var/lib/image-cache` and `linux/amd64`, and MetalK8s needs neither changed. If it ever does, that file is the place, not the units.
- a node carrying only the `ca` role would fail the new sanity scenarios, since it never gets the formula. Such a node has no kubelet either, and the CA minion is the bootstrap node in every supported layout.
- the systemd unit scenario runs two cluster wide salt calls per unit, one for `service.enabled` and one for `service.status`. One `cmd.run` would halve that, at the cost of parsing text instead of booleans.

---

See: MK8S-395

### DIFF
--- a/salt/metalk8s/image-cache/installed.sls
+++ b/salt/metalk8s/image-cache/installed.sls
@@ -1,0 +1,36 @@
+{%- from "metalk8s/macro.sls" import pkg_installed with context %}
+{%- from "metalk8s/map.jinja" import kubelet with context %}
+
+{%- if kubelet.container_engine == 'containerd' %}
+
+include:
+  - metalk8s.repo
+  - metalk8s.container-engine.containerd.installed
+  - metalk8s.container-engine.containerd.running
+
+Install containerd image preload:
+  {{ pkg_installed('containerd-image-preload') }}
+    - require:
+      - test: Repositories configured
+      # NOTE: The package requires `containerd`, a virtual name the
+      # `containerd.io` package we ship provides. Ordering the two keeps a
+      # fresh node from resolving that dependency to another provider.
+      - metalk8s_package_manager: Install containerd
+
+Ensure containerd image preload timer running:
+  service.running:
+    - name: containerd-image-preload.timer
+    - enable: True
+    - require:
+      - metalk8s_package_manager: Install containerd image preload
+      # NOTE: The timer fires as soon as it starts (`OnBootSec=0`), then
+      # every 10 minutes, and the service imports into a running containerd,
+      # so wait for the engine rather than let the first run fail.
+      - test: Ensure containerd is ready
+
+{%- else %}
+
+No containerd to preload images into:
+  test.succeed_without_changes
+
+{%- endif %}
--- a/salt/metalk8s/image-cache/installed.sls
+++ b/salt/metalk8s/image-cache/installed.sls
@@ -4,9 +4,7 @@
 {%- if kubelet.container_engine == 'containerd' %}
 
 include:
-  - metalk8s.repo
   - metalk8s.container-engine.containerd.installed
-  - metalk8s.container-engine.containerd.running
 
 Install containerd image preload:
   {{ pkg_installed('containerd-image-preload') }}
--- a/salt/metalk8s/roles/bootstrap/init.sls
+++ b/salt/metalk8s/roles/bootstrap/init.sls
@@ -4,3 +4,4 @@ include:
   - metalk8s.salt.master
   - metalk8s.utils
   - metalk8s.backup.configured
+  - metalk8s.image-cache
--- a/salt/metalk8s/roles/bootstrap/init.sls
+++ b/salt/metalk8s/roles/bootstrap/init.sls
@@ -4,4 +4,3 @@ include:
   - metalk8s.salt.master
   - metalk8s.utils
   - metalk8s.backup.configured
-  - metalk8s.image-cache
--- a/tests/post/steps/test_sanity.py
+++ b/tests/post/steps/test_sanity.py
@@ -1,3 +1,5 @@
+import json
+
 import kubernetes.client
 from kubernetes.client.rest import ApiException
 import pytest
@@ -36,6 +38,18 @@ def test_deployment_running(host):
 
 @scenario("../features/sanity.feature", "DaemonSet has desired Pods ready")
 def test_daemonset_running(host):
+    pass
+
+
+@scenario("../features/sanity.feature", "Package is installed on every node")
+def test_package_installed(host):
+    pass
+
+
+@scenario(
+    "../features/sanity.feature", "Systemd unit is enabled and running on every node"
+)
+def test_systemd_unit_running(host):
     pass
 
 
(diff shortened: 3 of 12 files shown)


ezekiel-alexrod:
/changie Added
Every node now installs the `containerd-image-preload` package and runs its timer. The timer imports the image archives found under `/var/lib/image-cache` into containerd, once at boot and then every 10 minutes, so a node that lost images from its content store gets them back without reaching a registry. Nothing fills that directory yet, so the timer is a no-op on a fresh cluster until the cache provisioning lands.

salt/metalk8s/image-cache/installed.sls
```diff
@@ -0,0 +1,36 @@
+{%- from "metalk8s/macro.sls" import pkg_installed with context %}
+{%- from "metalk8s/map.jinja" import kubelet with context %}
+
+{%- if kubelet.container_engine == 'containerd' %}
+
+include:
+  - metalk8s.repo
+  - metalk8s.container-engine.containerd.installed
+  - metalk8s.container-engine.containerd.running
```
TeddyAndrieux: nit: installed already include running so you could only include `installed` here

salt/metalk8s/roles/bootstrap/init.sls
```diff
@@ -4,3 +4,4 @@ include:
   - metalk8s.salt.master
   - metalk8s.utils
   - metalk8s.backup.configured
+  - metalk8s.image-cache
```
TeddyAndrieux: It's not needed to include it here since it's in "node" role all nodes have the "node" role

tests/post/steps/test_sanity.py
```diff
@@ -220,4 +234,53 @@ def _wait_for_stateful_set():
...
+    versions = _salt_on_every_node(host, ssh_config, k8s_client, "pkg.version", name)
+
+    # An absent package gives an empty string, and a minion that failed to run the
+    # function gives its error message, so only take a version number for an answer.
+    missing = sorted(
+        node
+        for node, version in versions.items()
+        if not isinstance(version, str) or not version[:1].isdigit()
+    )
+    assert not missing, "'{}' is not installed on {}".format(name, ", ".join(missing))
```
TeddyAndrieux: Any reason to not use f-string ?

Weird that pylint does not complain (I think it's part of pylintrc for test)

tests/post/steps/test_sanity.py
```diff
@@ -220,4 +234,53 @@ def _wait_for_stateful_set():
...
+    for description, function in [
+        ("enabled", "service.enabled"),
+        ("running", "service.status"),
+    ]:
+        results = _salt_on_every_node(host, ssh_config, k8s_client, function, name)
+
+        failed = sorted(node for node, result in results.items() if result is not True)
+        assert not failed, "'{}' is not {} on {}".format(
+            name, description, ", ".join(failed)
+        )
```
TeddyAndrieux: ditto ?

tests/post/steps/test_sanity.py
```diff
@@ -220,4 +234,53 @@ def _wait_for_stateful_set():
...
+        node.metadata.name
+        for node in k8s_client.resources.get(api_version="v1", kind="Node").get().items
+    )
+    assert nodes, "no Kubernetes node to check"
+
+    command = ["salt", "--static", "--out=json", "-L", ",".join(nodes), function, *args]
+    results = json.loads(utils.run_salt_command(host, command, ssh_config).stdout)
+
+    silent = sorted(set(nodes) - set(results))
+    assert not silent, "no answer from {}".format(", ".join(silent))
```
TeddyAndrieux: ditto

TeddyAndrieux (approved):

g-carre (approved):

PR #5114, opened 2026-09-02
fix(salt): stop the workload plane probe from resetting the Calico BGP mesh

TeddyAndrieux:
**Component**: salt, tests

**Context**:

The workload plane monitor TCP-probes port 179 on every peer every 30 seconds.
BIRD takes an incoming connection from a configured mesh peer for a new session
and resets the established one, so every node tears down all of its BGP mesh
sessions once per window, one reset every 15 seconds per session on average.

Every inter-node pod-to-pod flow pays for it permanently: measured 0.58% UDP
loss against 0% without the probe, and single-stream TCP throughput down from
1.63 to 1.14 Gbit/s, the peer route being withdrawn for about 40 ms per reset.
It also floods the calico-node logs, 23 MB per node per day, and makes BGP
session state useless as a signal. Nothing detected it: no test covers BGP
stability, and the check itself is green, only its side effect is harmful.

**Summary**:

The probe now resolves its peers through the existing headless
`node-problem-detector` Service and connects to their exporter port instead,
where a TCP connect is inert. It still tests the same property, inter-node
pod-to-pod communication over the real data path, with no new component and no
new Kubernetes object -- it even removes the `npd-wp-peers` ConfigMap and the
Salt mine peer list it was built from, deleted on upgrade by a `post-upgrade`
state. The condition is renamed `PodNetworkUnavailable`, which is what the probe
actually measures.

A CI guard comes with it, since nothing covered BGP stability: watch the
calico-node logs for ten minutes and fail on any session reset.

---

Fixes: ARTESCA-18071

### DIFF
--- a/salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
+++ b/salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
@@ -1,27 +1,7 @@
 #!jinja | metalk8s_kubernetes
 
-{#- Mine can hold stale entries after node removal; keep only nodes still in the pillar #}
-{%- set wp_mine = salt.saltutil.runner('mine.get', tgt='*', fun='workload_plane_ip') %}
-{%- set peers = {} %}
-{%- for node in pillar.metalk8s.nodes.keys() | sort %}
-  {%- if node in wp_mine %}
-    {%- do peers.update({node: wp_mine[node]}) %}
-  {%- endif %}
-{%- endfor %}
-{#- An empty map is safe: the probe treats "no peers" as healthy #}
+{%- from "metalk8s/map.jinja" import coredns with context %}
 
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: npd-wp-peers
-  namespace: metalk8s-monitoring
-data:
-  peers: |
-    # nodeName   WP IP
-{%- for name, ip in peers | dictsort %}
-    {{ name }} {{ ip }}
-{%- endfor %}
----
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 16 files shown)


TeddyAndrieux:
/changie Fix
The node-problem-detector workload plane check no longer resets every node's Calico
BGP sessions, which degraded inter-node pod traffic. The condition it reports is
renamed PodNetworkUnavailable.

salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
```diff
@@ -32,29 +12,43 @@ data:
...
     # Config sanity -> Unknown (exit 2): never risk a wrong verdict
-    [ -n "$SELF" ]       || { echo "NODE_NAME not set (cannot exclude self)"; exit 2; }
-    [ -r "$PEERS_FILE" ] || { echo "peers file $PEERS_FILE not readable";      exit 2; }
+    [ -n "$SELF" ] || { echo "POD_IP not set (cannot exclude self)"; exit 2; }
+
+    addrs=""
+    attempt=0
+    while [ "$attempt" -lt "$ATTEMPTS" ]; do
+      attempt=$((attempt+1))
+      addrs=$(timeout 2 getent hosts "$SERVICE" | sed 's/[[:space:]].*//' | tr '\n' ' ')
```
ezekiel-alexrod, line 34: The second lookup always overwrites the first, even when it comes back empty, so a good answer gets thrown away.

With an empty cache we then exit 2 instead of 0, and nothing gets written to the cache either, so the next run has nothing to fall back on. That is the state right after NPD restarts.

Keeping the last answer that was not empty is enough. I tried it.

```suggestion
      ans=$(timeout 2 getent hosts "$SERVICE" | sed 's/[[:space:]].*//' | tr '\n' ' ')
      # A failed retry must not discard a usable answer
      [ -n "$ans" ] && addrs=$ans
```

Adjust the suggestion as you wish, the statement prevails.

tests/post/steps/test_network.py
```diff
@@ -254,6 +264,40 @@ def nodeport_service_request_does_not_respond(k8s_client, host, svc_name, plane)
...
+        "no BGP session is reset in the calico-node logs over the next "
+        "{minutes:d} minutes"
+    )
+)
+def no_bgp_session_reset(k8s_client, minutes):
+    window = minutes * 60
+
+    # Anything connecting to BIRD's port from a mesh peer makes it reset the
+    # session, which withdraws the peer routes and drops inter-node pod traffic
+    time.sleep(window)
```
ezekiel-alexrod, line 283: This sleeps ten minutes on every multi-node run, for one check.

`since_seconds` already looks back ten minutes on its own, so the sleep only helps when the cluster is younger than that and the first sessions are still in the logs. Ten minutes on every run is a lot to pay for that. Or am I missing something?

tests/post/steps/test_network.py
```diff
@@ -1,11 +1,16 @@
...
 import requests
 
+import kubernetes.client
 import pytest
 from pytest_bdd import given, scenario, when, then, parsers
 
 from tests import utils, kube_utils
 
+# BIRD logs this when a BGP session goes down and starts over
+BGP_RESET_LOG = "State changed to start"
```
ezekiel-alexrod, line 13: BIRD also prints this when it opens a session for the first time, not only when one drops. So if a calico-node pod restarts during the ten minutes, the test fails and says the sessions were reset.

We could skip pods whose container started inside the window. Or just say "reset or restarted" in the message, so whoever hits it knows where to look.

salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
```diff
@@ -32,29 +12,50 @@ data:
...
+      # An answer without our own IP is an incomplete endpoint list, retry
+      case " $addrs " in *" $SELF "*) break ;; esac
+      [ "$attempt" -lt "$ATTEMPTS" ] && sleep 1
+    done
+
+    if [ -n "$addrs" ]; then
+      { echo "$addrs" > "$CACHE"; } 2>/dev/null || true
```
ezekiel-alexrod: Sometimes DNS gives back our own IP and nothing else. The loop above stops as soon as it finds `$SELF`, so a list of one is enough for it, and after a rollout the endpoints do come back one pod at a time.

If that list gets cached, `checked` stays at 0. Every run after that says "no peers to check" and exits 0, even on a node that is really cut off, for as long as DNS is down. I tried it with a fake resolver.

Simplest is to only cache a list that has someone else in it. A single-node cluster keeps working the same way, it just writes nothing.

tests/post/steps/test_network.py
TeddyAndrieux: The point is that there is quite some tests that restart calico so if we do not wait we will not have a lot of logs but maybe 10 minutes is too much, I think 2 minutes my be enough to catch that kind of issue so yes let's reduce it

salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
TeddyAndrieux: Fair

salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
TeddyAndrieux: Done

tests/post/steps/test_network.py
TeddyAndrieux: Done

tests/post/steps/test_network.py
TeddyAndrieux: Right, done we now check for calico pod restart

salt/metalk8s/addons/node-problem-detector/deployed/wp-monitor.sls
TeddyAndrieux: Done

ezekiel-alexrod (approved):

g-carre (approved):

PR #5118, opened 2026-09-03
fix(ui): tell the node and volume tables when they are loading (ARTESCA-5650)

Cuervino:
The Nodes and Volumes pages render their table with an empty list while the first fetch is still running, and never tell the table that it is loading. core-ui reads a missing status as "success", so an empty list immediately renders the "No nodes found" / "No volumes found" placeholder. It stays on screen for the API round trip plus the deliberate one second delay in the nodes and volumes sagas, on every visit to the page.

The Redux store already carries the flag, it just was not passed down. Both tables now receive it and show the loading placeholder instead, and the "no results" message is only reached once the fetch is over.

Verified against a live platform with the MetalK8s UI running locally behind the shell.

Left out on purpose: the `isFirstLoadingDone` guard in both containers is still the fragile `usePrevious` heuristic, which belongs with the react-query migration (ARTESCA-2771).

### DIFF
--- a/.changes/unreleased/Fix-20260903-124757-486517410.yaml
+++ b/.changes/unreleased/Fix-20260903-124757-486517410.yaml
@@ -1,0 +1,6 @@
+kind: Fix
+body: The Nodes and Volumes pages no longer show "No nodes found" / "No volumes found" while the list is still loading
+time: 2026-09-03T12:47:57.48651741Z
+custom:
+    CommentId: "5525994634"
+    Pull: "5118"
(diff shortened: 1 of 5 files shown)


Cuervino:
/changie Fix The Nodes and Volumes pages no longer show "No nodes found" / "No volumes found" while the list is still loading

JBWatenbergScality (approved):

MonPote (approved):
LGTM

PR #5120, opened 2026-09-04
fix(ui): resolve plane interfaces live instead of from a stale grain

JBWatenbergScality:
## TL;DR

The Platform dashboard's **ControlPlane / WorkloadPlane Bandwidth** charts (and the same two panels on the node page) are permanently blank on any cluster whose plane IPs sit on a bond or a VLAN. The UI resolves the interface name from the `ip_interfaces` grain, which is a snapshot taken when the minion daemon started — before those interfaces existed. This resolves the interface live instead.

Fixes MK8S-411.

## Context

The UI asks Salt which NIC carries the control-plane / workload-plane IP, then filters `node_network_*` on that device name. On the affected cluster the grain answers:

```json
{"lo":["127.0.0.1"],"ens21f0":[],"eno1":[],"ens21f1":[],
 "eno2":[],"ens20f0":[],"ens20f1":[],"usb0":[]}
```

while a live `network.interfaces` call on the same minion, at the same second, answers `vlan-cp=192.168.17.1`, `vlan-wp=192.168.18.1` (plus `bond-artesca`, `tunl0` and the `cali*` devices). The grain holds exactly the set of interfaces that exist before any network configuration runs, with no addresses — the picture salt-minion saw at boot.

So the IP matched nothing, `interface` fell back to `''`, and the dashboard sent:

```
avg(irate(node_network_receive_bytes_total{device=~""}[5m])) by (instance,device)
```

`device=~""` matches only series with an empty `device` label. Prometheus returns **200 with an empty result**, so the panel renders blank with no error anywhere.

Why the grain goes stale: Salt computes grains once per minion-daemon start and holds them for the process lifetime; MetalK8s additionally sets `grains_cache: true`. Nothing refreshes them outside orchestration (`orchestrate/deploy_node.sls`, `bootstrap/init.sls`, `node/grains.sls` — whose own comment notes "Since we enable grains cache, the cache is not refreshed automatically"). The charts are therefore **correct right after install** and break at the **first node reboot**. Clusters whose plane IPs land on a NIC that already exists at minion start never show this.

## What changed

**`salt/_modules/metalk8s_network.py`** — new `get_planes_interfaces()`:

```python
{"control_plane":  {"ip": "10.200.0.42", "interface": "eth1"},
 "workload_plane": {"ip": "10.100.0.42", "interface": "eth3"}}
```

It resolves the interface with `network.ifacestartswith`, the same live primitive `get_mtu_from_ip` already uses. `interface` is `None` when nothing holds the IP, so callers can tell "unresolved" from a real name. This follows what the rest of the module already does — `get_ip_from_cidrs` uses live `network.ip_addrs` — the UI was the only place reading a volatile core grain.

**`ui/src/services/salt/api.ts`** — calls the new function instead of `grains.item`; `IPInterfaces` becomes `PlanesInterfaces`.

**`ui/src/services/NodeUtils.ts`** — `nodesCPWPIPsInterface` becomes a plain adapter; the IP-to-interface matching moves to the minion. Same output contract, so nothing downstream changes.

**`ui/src/services/graphUtils.ts`** — `getNodesInterfacesString` drops unresolved interfaces. The dashboard query is guarded by `enabled: !!devices?.length`, which never fired because `devices` was `['', '', '']` — non-empty. Filtering makes that guard work, so an unresolved interface now skips the query rather than sending one that cannot match.

## Review focus

1. **`get_planes_interfaces` returning `None` rather than raising** (`metalk8s_network.py`) — deliberate: one unresolvable plane shouldn't take down the other, and the UI wants to render an empty state, not an error. `get_mtu_from_ip` raises instead, so this is an intentional divergence from the neighbouring function.
2. **`''` kept as the UI's "unresolved" value** (`NodeUtils.ts`) — the node-page queries already guard with `enabled: planeInterface !== ''`. Switching to `undefined`/`null` would silently re-enable them with `device="undefined"`. Worth confirming you agree the contract should stay `''`.
3. **Prefix-matching semantics of `network.ifacestartswith`** — it matches on address prefix, so `10.200.0.4` would also match `10.200.0.42`. Inherited from `get_mtu_from_ip`; full plane IPs make a collision unlikely, but flagging it explicitly.
4. **Upgrade ordering** — during an upgrade a node may briefly run an older minion without the new module. The call then errors, `nodesCPWPIPsInterface` takes its existing string/false branch, and the chart shows empty — same as today, no crash.

## Testing

Salt, `salt/tests/unit/modules/test_metalk8s_network.py` — 5 new cases: a dedicated interface per plane, one interface carrying both, no interface holding the IP, several matches (first wins), and plane IPs missing from the grains.

```
5 passed
_modules/metalk8s_network.py   146 stmts   0 miss   100%
```

UI — new `src/services/NodeUtils.test.ts` (3 cases) and 2 new cases in `graphUtils.test.ts` covering the unresolved-interface filter.

```
Test Suites: 25 passed, 1 skipped
Tests:       227 passed
tsc --noEmit: clean
```

One pre-existing failure, `test_get_kubernetes_service_ip_raise_1_10_0_0_0_32`, reproduces on the base branch in my local Python 3.9.6 env and is unrelated.

### Verified against a live affected cluster

Without deploying anything, by calling the exact primitive the new function uses:

| Check | Result |
|---|---|
| `network.ifacestartswith` on each plane IP | `["vlan-cp"]` / `["vlan-wp"]` on all 3 nodes — one match each |
| `device=~""` (current behaviour) | `success`, **0 series** |
| `device=~"vlan-cp\|vlan-wp"` (after fix) | `success`, **6 series**, e.g. `vlan-cp = 350153 B/s` |

## Screenshots

Before — dashboard panels blank, legend present, no error:

After:

## Follow-ups (not in this PR)

- An unresolved interface still renders as an empty chart rather than saying so. A visible "interface not detected" state would be the honest fix.
- The staleness itself is untouched here. `grains_refresh_every` in the minion config is the only setting that refreshes an already-running daemon; ordering salt-minion `After=network-online.target` would fix the boot-time capture. Worth its own ticket.
- Ops workaround meanwhile: `salt '*' saltutil.refresh_grains`.

### DIFF
--- a/.changes/unreleased/Fix-20260904-081500-411000000.yaml
+++ b/.changes/unreleased/Fix-20260904-081500-411000000.yaml
@@ -1,0 +1,14 @@
+kind: Fix
+body: |
+  The Platform dashboard and node page bandwidth charts are no longer empty on
+  nodes whose control-plane or workload-plane IP sits on a bond or a VLAN. The
+  interface carrying each plane IP is now resolved live on the minion instead of
+  being read from the `ip_interfaces` grain, which is only computed when the
+  minion daemon starts -- before those interfaces exist -- and is not refreshed
+  afterwards. The charts used to query `node_network_receive_bytes_total` with an
+  empty device matcher, which no series can match, so Prometheus answered with an
+  empty result and the panel rendered blank without any error
+time: 2026-09-04T08:15:00.411000000Z
+custom:
+  Pull: 5120
+  CommentId: cli
--- a/salt/_modules/metalk8s_network.py
+++ b/salt/_modules/metalk8s_network.py
@@ -126,6 +126,61 @@ def get_mtu_from_ip(ip):
     return int(__salt__["file.read"](f"/sys/class/net/{iface}/mtu"))
 
 
+def _get_iface_from_ip(ip):
+    """
+    Return the name of the first interface holding the given IP, or None
+
+    The interface is resolved live, on purpose: the `ip_interfaces` grain is
+    only computed when the minion daemon starts, which may happen before the
+    interface holding the IP exists (bond, VLAN, ...), and grains are not
+    refreshed automatically afterwards.
+    """
+    if not ip:
+        return None
+
+    ifaces = __salt__["network.ifacestartswith"](ip)
+
+    if not ifaces:
+        log.warning('Unable to get interface for "%s"', ip)
+        return None
+
+    iface = ifaces[0]
+
+    if len(ifaces) > 1:
+        log.warning(
+            'Several interfaces match the IP "%s", %s will be used: %s',
+            ip,
+            iface,
+            ", ".join(ifaces),
+        )
+
+    return iface
+
+
+def get_planes_interfaces():
+    """
+    Return the IP and the interface holding it for each plane
+
+    .. code-block:: python
+
+        {
+            "control_plane": {"ip": "10.200.0.42", "interface": "eth1"},
+            "workload_plane": {"ip": "10.100.0.42", "interface": "eth3"},
+        }
+
+    An `interface` is None when no interface holds the plane IP, so that
+    callers can tell "not resolved" apart from a real interface name.
+    """
+    metalk8s_grains = __grains__.get("metalk8s") or {}
+
+    result = {}
+    for plane in ["control_plane", "workload_plane"]:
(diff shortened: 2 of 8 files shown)


TeddyAndrieux (approved):
Salt part LGTM (I let someone from Pod UI validate the UI part)

PR #5122, opened 2026-09-08
feat(salt): provision the boot cache on the bootstrap node (MK8S-157)

ezekiel-alexrod:
**Component**: salt, tests

**Context**:

MK8S-395 (#5109) installs `containerd-image-preload` and starts its timer on every node, and it is merged. Nothing fills the directory it imports from, so on a fresh cluster the timer is still a no-op. This PR fills it on the bootstrap node.

The bootstrap node is the hard case. There is no registry to pull from, and no Kubernetes to run the image cache agent, so the only source is the ISO. It ships `metalk8s-boot-cache-control-plane-<version>.tar` (MK8S-153, #5068), a carrier image built `FROM scratch` with a single `COPY`, holding one archive per container image a control plane node needs to boot.

**Summary**:

Three commits, in the order they build on each other.

`metalk8s_image_cache.provision` reads the archives out of the docker archive and writes them into the cache directory. The headers are walked before anything is written, and the layer is opened for random access rather than as a stream: an image the module refuses is refused whole, and a second run seeks over the payload instead of reading several hundred megabytes off the ISO to throw them away. Each archive is written under a hidden name, flushed, fsynced, then renamed, since the timer may fire at any moment and the node is mid-install. A new `metalk8s_image_cache.provisioned` state wraps it so that `test=True` reports what a run would write, and so the `changes` it returns are accurate.

Salt then renders `/etc/sysconfig/containerd-image-preload` from the same pillar values the states use. Without that, the cache directory is two independent values that drift: a changed pillar would have Salt populate one directory while the timer kept importing from the other, both reporting success. The package ships the file as `%config(noreplace)`, so RPM leaves it alone.

`metalk8s.image-cache.provisioned` extracts, then restarts the preload service once. The service is `Type=oneshot` and its timer fires at boot, then every 10 minutes. Enabling the timer keeps the cache imported over the life of the node, but at bootstrap it has most likely fired already, back when there was nothing to import, and waiting for the next firing would let the kubelet start with no images. A restart rather than a start, because systemd merges a start request into a start job already in flight: if the timer fired while the archives were still being written, the request would come back happy from a run that had listed the directory too early.

The ordering against the kubelet is declared in `roles/bootstrap/local.sls`. Not in the formula, because an `_in` requisite only resolves against a state in the same run, so declaring it there would make `state.sls metalk8s.image-cache.provisioned` fail to compile on its own. And not in `init.sls`, because that is not where the kubelet first starts: `bootstrap.sh` applies `metalk8s.roles.internal.early-stage-bootstrap`, which reaches `local.sls` and `metalk8s.kubernetes.kubelet.standalone` long before the highstate. `init.sls` includes `local.sls`, so the highstate is covered too.

Tests come in two layers. Twenty five unit tests on the module, built around tar archives fabricated in the test, covering the nominal extraction, idempotence, a truncated archive, the dry run, a failed write, and every archive this module refuses. Then four formula tests that read the requisites back from the rendered states, so an edit that drops the gate, the package configuration or the mount fails the build rather than the cluster.

**Acceptance criteria**:

- a bootstrap on a node with no registry reachable ends with the archives in `/var/lib/image-cache` and the images visible in containerd's `k8s.io` namespace
- the state is idempotent: a second run changes nothing and copies nothing
- an interrupted extraction leaves no partial archive under its final name
- unit tests on the module and on the requisites
- changelog entry

**Why not `ctr images import` and `ctr images mount`**:

The ticket asked for import, mount, copy, unmount. It does work, I checked it against containerd, but it is expensive on the node that can least afford it. The archive on the ISO is not compressed, so the import writes about 1.1 GB into the content store, and the unpack it performs by default writes about as much again into the snapshotter. Neither is given back by `ctr images rm`: the namespace still refuses to be deleted afterwards, with `must be empty, but it still has blobs, snapshots on "overlayfs"`. Containerd's garbage collector reclaims them whenever it next runs.

Reading the archive costs one pass and leaves nothing behind. It also drops the temporary mount that would have had to be unmounted even when the copy fails, which Salt has no clean way to guarantee.

One thing the mount approach had going for it: `ctr images mount` pulls the image itself when it is not local, so MK8S-157 and MK8S-394 would have shared a state, with only the source changing. A joining node reaches the ISO over HTTP from the `repositories` pod, which `cp.cache_file` accepts, so the same extraction should cover the join too. I have not proven that, it belongs to MK8S-394.

**Run locally**:

The whole salt unit suite on each of the three commits: 1277, 1278 and 1281 tests, coverage on `salt/_modules` and `salt/_runners` at 100% every time. Every pre-commit hook over the range: black 26.3.1, pylint 4.0.5 at 10.00/10, mypy 1.20.2, salt-lint 0.9.2, yamllint 1.38.0.

The shape the module relies on was checked rather than assumed. An image built the way the buildchain builds it, `FROM scratch` plus one `COPY`, exported with `skopeo copy` to `docker-archive:`, gives a single layer named by `manifest.json[0].Layers[0]`, uncompressed inside the archive, holding the `images/*.tar` entries. The import and mount numbers above come from the same image against a local containerd.

Not run: a real bootstrap. Now that #5103 and #5109 are merged, an ISO built from `development/134.0` carries the package and the state that installs it, so the e2e on this PR is the first run that can actually exercise the whole path.

**Three things a reviewer should weigh**:

**Idempotence is a size check, not a digest.** An archive already in the cache with the size the image announces is skipped. Same size with different content gets through, and the import that follows is what fails loudly. The agent does better, with a digest and a file list, because it handles updates. The cold path writes once. A digest here would mean reading both the cached file and the layer in full on every highstate.

**The module refuses more than it skips.** Only directory entries are ignored. A hard link, a symlink, two archives sharing a base name, a member with no file name, a second layer, a second image, an archive carrying nothing at all: each one fails with a message naming what it found, before a single byte is written. Half a cache reported as provisioned would release the kubelet on a node that cannot pull, and the preload script globs the directory whatever the state reported, so failing whole is worth more than being lenient. The single layer and the single image are assertions on how the boot cache image is built: change that Dockerfile and this state tells you.

**The gate is in the bootstrap role, so MK8S-394 has to add its own.** Keeping the ordering out of the formula is what leaves the formula applicable on its own, but it does mean the node role will need the same few lines rather than inheriting them. The alternative was to include the whole kubelet formula from the image cache formula, which would have made a standalone provisioning run install and start the kubelet.

**Out of scope**:

- filling the cache when a node joins, covered by MK8S-394
- upgrades, covered by MK8S-167
- the worker and registry variants, and picking a variant from the roles a node holds, covered by MK8S-380 and MK8S-394. Only the control plane variant is provisioned here.
- the agent and the `ImageCache` resources, covered by MK8S-166 and MK8S-396
- the preload script aborts its import loop on the first archive it cannot read, because of `set -euo pipefail`, so one bad archive keeps the ones listed after it out of containerd. Worth its own ticket against the package rather than a change here.
- the changelog entry #5109 left under `.changes/unreleased/` ends with "Nothing fills that directory yet". Both ship in the same release, so that sentence stops being true the day this merges. Its comment can no longer regenerate it, #5109 being merged, so the fragment has to be edited by hand. Say the word and I will do it here.

---

See: MK8S-157


ezekiel-alexrod:
/changie Added The bootstrap node now fills its image cache from the boot cache image shipped on the ISO, before the kubelet starts. The archives it carries are extracted into `/var/lib/image-cache` and imported into containerd right away, so the first node can bring up its static pods with no registry to pull from. Only the control plane variant is provisioned, and only at bootstrap: a node joining the cluster still gets nothing.

.changes/unreleased/Added-20260901-135539-931914026.yaml
```diff
@@ -1,5 +1,5 @@
 kind: Added
-body: Every node now installs the `containerd-image-preload` package and runs its timer. The timer imports the image archives found under `/var/lib/image-cache` into containerd, once at boot and then every 10 minutes, so a node that lost images from its content store gets them back without reaching a registry. Nothing fills that directory yet, so the timer is a no-op on a fresh cluster until the cache provisioning lands.
+body: Every node now installs the `containerd-image-preload` package and runs its timer. The timer imports the image archives found under `/var/lib/image-cache` into containerd, once at boot and then every 10 minutes, so a node that lost images from its content store gets them back without reaching a registry.
```
g-carre, line 2: Note for later: we should avoid putting temporary statements in changelog

g-carre (approved):
LGTM, would need a second pair of 👁️

anthony-treuillier-scality (approved):

.changes/unreleased/Added-20260901-135539-931914026.yaml
ezekiel-alexrod: Agreed. The sentence from #5109 is already gone, in 5ac2b35c9 on this branch.

My own entry has the same problem: it ends on "a node joining the cluster still gets nothing", which stops being true when MK8S-394 lands. I'd rather not push that here, since new commits dismiss the approvals for one clause.

The fragments stay editable until the 134.0 batch, and check-changelog counts a modification as much as an addition, so MK8S-394 can drop the clause in the same PR that makes it false. That's where I'll do it.

PR #5126, opened 2026-09-10
bugfix(salt): gate the apiserver upgrade on /readyz, and tolerate transient API failures (MK8S-420, MK8S-419)

g-carre:
**Component**: salt

**Context**:

Two orchestrates aborted because a Kubernetes API call was made once, around etcd churn. Sosreport analysis turned the first one from "the API server was unavailable, tolerate it" into a specific, preventable defect, so this PR leads with the prevention.

| | Ticket | Seen in |
| --- | --- | --- |
| `upgrade.sh` gate accepts a `/healthz ok` returned **during the API server's graceful shutdown** | MK8S-420 | [34178644526](https://github.com/scality/metalk8s/actions/runs/34178644526) |
| `object_present` aborts on `500 etcdserver: leader changed` mid-orchestrate | MK8S-419 | [34330756346](https://github.com/scality/metalk8s/actions/runs/34330756346/attempts/1) |

**Summary**: three commits, in dependency order. Full reasoning is in the commit messages.

1. **`/healthz` → `/readyz` in `upgrade.sh` (MK8S-420)** — the main fix. The sosreport shows the gate's single `ok` landing **three seconds into the API server's graceful shutdown**, which is exactly what `/healthz` is specified to do: while shutting down, kube-apiserver keeps `/healthz` and `/livez` succeeding and fails `/readyz`, so load balancers drain it. `/readyz` would have refused and the gate would have waited.
2. **Transport retry (MK8S-420)** — defence in depth. A gate constrains one instant; `list_objects` runs at orchestrate render time and has callers with no gate at all.
3. **Transient-status retry (MK8S-419)** — no gate can help here: `etcdserver: leader changed` arrives mid-orchestrate, arbitrarily far from any gate.

5xx and dropped connections get different budgets *and* different safety rules: a 5xx means the server did not apply the request, so any action may be repeated (5 × 2s); a dropped connection leaves that unknowable, so it is retried for reads only (12 × 5s). 404/409/4xx behaviour is unchanged.

**Verification**:

- `salt/tests/unit`: **850 passed**, `salt/_modules` at **100%** coverage (`--cov-fail-under=100` satisfied), run on python3.9 against the pinned deps. The only failures are 31 in `test_metalk8s_etcd.py`, identical on the base commit — local venv lacks `etcd3`.
- `shellcheck` 0.11.0 clean on `scripts/upgrade.sh.in`, matching `tox -e lint-shell`.
- `black` 22.8.0 clean. Full pre-commit cannot run here (pins python3.6), so saltpylint is unexercised locally.

**Superseded reading**: an earlier revision of this PR and of MK8S-420 said the gate "passed and the API server was gone again 29 seconds later", attributing it to the API server flapping under kubelet liveness restarts. Both were wrong. The 29s was the gate *waiting*; the sosreport shows one container generation for the whole run, no second generation, and no liveness failure or `Killing container` for kube-apiserver anywhere in the kubelet journal.

**Not in scope**:

- `orchestrate/etcd.sls` reported success at 02:25:58 while the kubelet only started the new etcd pod at 02:26:25 — `Check etcd cluster health` passed against the etcd being replaced. Gating on the new static pod actually running would close that 27s window rather than race it. Larger change, follow-up on MK8S-420.
- Changelog: none yet, deliberately. The `/changie` entry was removed while this is a draft — see MK8S-421. To be added before marking ready for review.

---

See: MK8S-420, MK8S-419

### DIFF
--- a/salt/_modules/metalk8s_kubernetes.py
+++ b/salt/_modules/metalk8s_kubernetes.py
@@ -43,6 +43,12 @@ __virtualname__ = "metalk8s_kubernetes"
 TRANSPORT_RETRY_ATTEMPTS = 12
 TRANSPORT_RETRY_INTERVAL = 5
 
+# The API server answered and asked us to come back later, an etcd leader
+# election typically. It is up, so this resolves in seconds rather than minutes.
+RETRIABLE_API_STATUSES = (500, 502, 503, 504)
+API_STATUS_RETRY_ATTEMPTS = 5
+API_STATUS_RETRY_INTERVAL = 2
+
 
 def __virtual__():
     if MISSING_DEPS:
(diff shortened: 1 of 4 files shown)


g-carre:
/changie Fix
An upgrade or a Control Plane Ingress reconfiguration no longer aborts when the
API server is briefly unavailable, as it is while etcd restarts or elects a new
leader. A request rejected with `etcdserver: leader changed` is now retried
instead of failing its state, and a connection dropped while the API server
restarts is retried on reads, so compiling the node pillar no longer ends the
whole orchestrate before a single state has run.

PR #5135, opened 2026-09-21
feat(salt): fill the image cache when a node joins the cluster (MK8S-394)

ezekiel-alexrod:
**Component**: salt, tests, build

**Context**:

MK8S-157 (#5122) fills the image cache on the bootstrap node, from the ISO that node mounts. Nothing mounts the ISO on a node that joins the cluster, so that node gets nothing. This PR fills its cache from the cluster registry instead, before its kubelet starts.

The boot cache image is only a carrier. It is built `FROM scratch` with one `COPY`, and its single layer holds one docker archive per image the node needs to boot.

**Summary**:

Nine commits. Each one explains its own reasoning, so this is only the shape.

A new function in `metalk8s_image_cache` fetches the manifest and the layer from the registry with `ctr content fetch-*`, then writes the archives into `/var/lib/image-cache`. One marker file per variant records what was written, so a second run answers from the disk alone. The `pulled` state and `metalk8s.image-cache.pulled` wire it up: every node pulls the worker variant, and a control plane node pulls the control plane one on top. `roles/node/init.sls` holds `Ensure kubelet running` until the cache is filled.

Importing the archives into containerd is not this PR. The `containerd-image-preload` service does it, into the `k8s.io` namespace (MK8S-154). This PR fills the directory it reads, then triggers it once.

**Acceptance criteria**:

- archives in `/var/lib/image-cache` and images in `k8s.io` before the kubelet starts: done, by the ordering gate and the sanity scenario
- the variants match the node roles: done, a formula test ties each render case to the set of variants it expects
- redeploying a node does not pull again what it already has: done, with no network call at all
- a registry that does not answer names the endpoint it tried: done, checked against a local registry
- unit tests, plus coverage in the node deployment tests: partly, see "Not verified"
- changelog entry: to post as a `/changie` comment once this PR is open

**Decisions worth knowing before you read the diff**:

- Not the `ctr images pull` the ticket asks for. The carrier is not the payload, so pulling it puts no usable image in containerd and the archives still need extracting. It also leaves about 2.2 GB behind that `ctr images rm` does not free, measured in MK8S-157.
- Not the ISO. The MK8S-157 PR body guessed a joining node could download the ISO `.tar` over HTTP. It cannot: the images sit outside the nginx docroot, and `image_path: '/images/'` in `repo/installed.sls` is a path on the ISO, not a URL. So `image-cache.provisioned` cannot be reused either, it includes `archives.mounted`.
- The disk is read before the registry. The gate runs on every highstate, not only on join. Calling the registry every time would make the kubelet depend on pods that need kubelets, and a broken cluster could no longer be fixed by a highstate.
- The gate names one state, not the whole SLS. `metalk8s.image-cache.pulled` has a branch that fails when the nodes pillar cannot be read, and that failure says nothing about the archives a node already holds. Gating on the SLS would leave a highstate unable to restart a kubelet that is down on a cluster whose apiserver is unreachable. A failed pull does still hold the kubelet back, which is the point.
- The gate has its own state ID. The bootstrap node applies both `roles/bootstrap/local.sls` and `roles/node`, and Salt needs unique IDs. The same ID in both fails the whole render. A test renders both files and compares.
- No plain HTTP flag. `--hosts-dir` is enough: the scheme comes from the host entry containerd writes. Measured, `--plain-http` alone fails on DNS and adding it changes nothing.

**Run locally**:

1322 Salt tests at 100% coverage, 77 buildchain tests, every pre-commit hook, and each commit green on its own. The module also ran against a local registry with `ctr` 2.2.5, the version the nodes pin: it filled an empty cache, did nothing on a full one, restored an archive deleted by hand, and named the endpoint when the image was missing.

**Out of scope**: upgrade (MK8S-167), the registry variant (MK8S-380, blocked by MK8S-168), and the agent (MK8S-166).

**Known limits, each worth a ticket**:

- `_ctr` has no timeout. A registry that connects then goes quiet blocks the read. The orchestrate timeout limits it in practice.
- disk use. Every archive exists twice before the renames, and nothing deletes the previous version, so an upgrade needs about three copies.
- the sanity step calls `file.find` with no max depth, so a `*.tar` in an agent subdirectory would pass.

**Not verified**:

- the `pulled` state has no unit test. No state in this repo has one.

---

See: #5122
Closes: MK8S-394


ezekiel-alexrod:
/changie Added
A node joining the cluster now fills its image cache from the registry, before its kubelet starts. It pulls the boot cache images its roles call for, extracts the archives into `/var/lib/image-cache` and imports them into containerd right away, so it holds the images it needs to boot without depending on a registry later on. Every node takes the worker variant, and a control plane node takes the control plane variant as well. A node that already holds the archives settles from its own disk, with no call to the registry.

salt/metalk8s/roles/node/init.sls
```diff
@@ -5,3 +5,28 @@ include:
   - metalk8s.internal.preflight
   - metalk8s.beacon.certificates
   - metalk8s.image-cache
+  - metalk8s.image-cache.pulled
```
TeddyAndrieux, line 8: IMHO we shouldn't do this here because if we do it here it means it will be run on every highstate (including upgrade) to me we only have to run it on first join and that's it, for upgrades it's handled by the agent

And I just saw that it's the same for bootstrap which is wrong as well to me, it basically mean the agent is pretty useless to me for bootstrap it's the same it should only be some commands in the script and that's it

salt/metalk8s/image-cache/pulled.sls
TeddyAndrieux, line 1: I don't get it, to me you always get the "worker" one and that's it then it's the agent that retrieve the other ones based on the roles

salt/_modules/metalk8s_image_cache.py
```diff
@@ -157,6 +187,330 @@ def _write(layer, member, target):
             os.remove(tmp)
 
 
+@contextlib.contextmanager
+def _ctr(args):
+    """Run a :command:`ctr` command, yielding its standard output as a stream.
+
+    The blob of a boot cache image weighs about a gigabyte, so it is read as it
+    arrives rather than captured: nothing here ever holds the whole payload.
+    """
```
TeddyAndrieux, line 196: As discussed on slack, I'm not sure to get the usage for this ? To me we should use simple `ctr` commands to import, mount and cp the images and that's it (+ a cleanup at the end)

